PR #2027, opened 2019-01-21
[1.11] Increase max grpc message size

mrunalp:
We hit a limit when there are a lot of containers running.

Note: For the newer branches I will open PRs with this as a configuration option.

@giuseppe @runcom @rhatdan ptal.

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -519,7 +519,13 @@ func main() {
 			logrus.Fatalf("failed to listen: %v", err)
 		}
 
-		s := grpc.NewServer()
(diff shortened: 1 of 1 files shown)


rjhowe:
We set this max to 8MB for the following why was 16MB set for cri-o?
- <https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/remote/docker_server.go#L31>
- <https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/remote/utils.go#L29>

https://github.com/kubernetes/kubernetes/pull/63977

PR #2028, opened 2019-01-21
 pkg/storage/image: wire in blocking registries feature

runcom:
**- What I did**

add blocking registries feature mainly, we'll need this PR for 1.12/1.13 as well

**- How I did it**

vim

**- How to verify it**

tests (about to add one)

### DIFF
--- a/pkg/storage/image.go
+++ b/pkg/storage/image.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 	"path"
 	"strings"
@@ -553,9 +554,16 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
 		}
 		return nil, err
 	}
+	registry, err := sysregistriesv2.FindRegistry(nil, imageName)
+	if err != nil {
+		return nil, err
+	}
+	if registry != nil && registry.Blocked {
+		return nil, fmt.Errorf("cannot resolve names from registry %q because it's blocked", imageName)
+	}
 	domain, remainder := splitDockerDomain(imageName)
+	// this means the image is already fully qualified
 	if domain != "" {
-		// this means the image is already fully qualified
 		return []string{imageName}, nil
 	}
 	// we got an unqualified image here, we can't go ahead w/o registries configured
@@ -598,6 +606,7 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
 		imageCache:            make(map[string]imageCacheItem),
 		ctx:                   ctx,
 	}
+
 	if len(registries) != 0 {
 		seenRegistries := make(map[string]bool, len(registries))
 		cleanRegistries := []string{}
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -68,9 +68,6 @@ STREAM_PORT=${STREAM_PORT:-10010}
 
 TESTDIR=$(mktemp -d)
 
-# podman pull needs a configuration file for shortname pulls
-export REGISTRIES_CONFIG_PATH="$INTEGRATION_ROOT/registries.conf"
-
 # Setup default hooks dir
 HOOKSDIR=$TESTDIR/hooks
 mkdir ${HOOKSDIR}


runcom:
I'm adding a test for this stuff

vrothberg:
As I wasn't very clear, this API doesn't exist yet :)

pkg/storage/image.go
```diff
@@ -556,6 +558,9 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
 	domain, remainder := splitDockerDomain(imageName)
 	if domain != "" {
 		// this means the image is already fully qualified
+		if svc.blockedRegistries[domain] {
```
vrothberg: An alternative to that would be `sysregistriesv2.FindRegistry(...)` and check for reg.Blocked (nil == not blocked).

vrothberg:
> LGTM
> 
> I think a `sysregistriesv2.IsBlocked(ref string) bool` would be nice for such tasks. It has an internal cache already. (Cc @mtrmac)

I judged too quickly, sorry @runcom. I suggest using `FindRegistry(...)` instead. Once, we migrate to the new format of syregistriesv2, we actually need to use that to transparently use prefixes.

pkg/storage/image.go
mtrmac: Yes please; use the API to do lookups (e.g. to allow the package to optimize them in the future).

pkg/storage/image.go
```diff
@@ -598,6 +603,25 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
...
+	}
+	blockedRegistries := make(map[string]bool)
+	unqualifiedSearchRegistries := []string{}
+	for _, r := range sysregistries {
+		if r.Blocked {
+			blockedRegistries[r.URL] = true
+			continue
+		}
+		if r.Search {
+			unqualifiedSearchRegistries = append(unqualifiedSearchRegistries, r.URL)
```
mtrmac: It’s not obviously true that `GetRegistries` returns search entries in the user-intended order. Why is this an integral part of the “blocked” feature?

pkg/storage/image.go
mtrmac: `FindRegistry`

runcom:
@vrothberg @mtrmac updated ptal

pkg/storage/image.go
```diff
@@ -555,6 +556,13 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
 	}
 	domain, remainder := splitDockerDomain(imageName)
 	if domain != "" {
+		registry, err := sysregistriesv2.FindRegistry(&types.SystemContext{}, domain)
```
mtrmac: (You can use `nil` instead of `&types.SystemContext{}`.)

Is it consistent to use an empty context here, while `GetImageService` makes it a variable?

(IIRC the only caller calls `GetImageService` with `nil`, so it may not matter.)

pkg/storage/image.go
```diff
@@ -555,6 +556,13 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		if registry.Blocked {
+			return nil, fmt.Errorf("cannot resolve names from registry %q because it's blocked", domain)
+		}
 		// this means the image is already fully qualified
 		return []string{imageName}, nil
```
mtrmac: Shouldn’t the non-qualified search code support the “blocked” flag as well?

runcom: yup forgot

runcom: done, basically filtering out the additional registries on startup

pkg/storage/image.go
```diff
@@ -616,6 +625,9 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
 			return nil, err
 		}
 		for _, r := range systemRegistries {
+			if r.Blocked {
+				continue
+			}
```
mtrmac: The other branch of the if (`cleanRegistries` etc.) does not need the Blocked flag?

pkg/storage/image.go
```diff
@@ -555,6 +556,13 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
...
 	domain, remainder := splitDockerDomain(imageName)
 	if domain != "" {
+		registry, err := sysregistriesv2.FindRegistry(nil, domain)
+		if err != nil {
+			return nil, err
+		}
+		if registry != nil && registry.Blocked {
+			return nil, fmt.Errorf("cannot resolve names from registry %q because it's blocked", domain)
+		}
 		// this means the image is already fully qualified
```
mtrmac: This comment logically belongs at top of this `if domain != ""` branch, above the added code.

mtrmac: Could you also add the comment into the domain == ""` case to say that `registry.Blocked` has already been processed earlier?

Although…, (@vrothberg ?), is it logically correct to do `FindRegistry`+`.Blocked` on hostnames only, and not on the full repo paths? That would break the “check `.Blocked` on initialization” optimization — and BTW `FindRegistry` above, should in that case include all of `imageName`, not just `domain`.

vrothberg: Good catch, @mtrmac! FindRegistries(...) must be called with `imageName`. At the moment, I might still just work as no `registries.conf` is using the new format yet but it would be a bug once the new format is used.

pkg/storage/image.go
vrothberg: @mtrmac a feeling inside me grows that the blocked semantics should move into c/image/docker to avoid the pitfalls (and the redundant calls to `FindRegistries(...)` + boilerplate across users). Let's all put our heads together this week at DevConf and check if there are similar cases to make sysregistriesv2 easier to use.

pkg/storage/image.go
runcom: is it working _now_ however if I pass `imageName`?

vrothberg: Absolutely (see https://github.com/containers/image/blob/master/pkg/sysregistriesv2/system_registries_v2_test.go#L79).

rhatdan:
@mtrmac @vrothberg PTAL, passing all tests now.

vrothberg:
LGTM (given the blocked-semantic will at some time move into c/image/docker) :+1:

pkg/storage/image.go
```diff
@@ -598,13 +606,21 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
...
 	}
+
 	if len(registries) != 0 {
 		seenRegistries := make(map[string]bool, len(registries))
 		cleanRegistries := []string{}
 		for _, r := range registries {
 			if seenRegistries[r] {
 				continue
 			}
+			registry, err := sysregistriesv2.FindRegistry(nil, r)
```
mtrmac: I can’t concentrate to review this in detail now, but shouldn’t the “search registries” case also use the full image name? (i.e. it would have to happen in the search evaluation phase, not here)

runcom: that's already covered in `ResolveNames` then right?

runcom: I'm removing this bits from here

mrunalp:
Is this ready?

pkg/storage/image.go
mtrmac: `ResolveNames` does not currently call `FindRegistry` on the full path, i.e. the value `path.Join(r, rem)`

pkg/storage/image.go
mtrmac: If `ResolveNames` is taught to check the `Blocked` flag for the complete image name created by combining a search registry with an unqualified input, this should be unnecessary.

TomSweeneyRedHat:
LGTM

pkg/storage/image.go
```diff
@@ -553,9 +554,16 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
 		}
 		return nil, err
 	}
+	registry, err := sysregistriesv2.FindRegistry(nil, imageName)
+	if err != nil {
+		return nil, err
+	}
+	if registry != nil && registry.Blocked {
+		return nil, fmt.Errorf("cannot resolve names from registry %q because it's blocked", imageName)
+	}
```
mtrmac: No, this calls `FindRegistry` with an unqualified value.

In the `if domain != ""` case, yes, do that (like, a previous version of this PR did).

In the other case, _within_ the `for _, r := range svc.unqualifiedSearchRegistries` loop, make the same check on the `path.Join(r, rem)` value.

rhatdan:
@runcom Any movement on this one?

pkg/storage/image.go
mrunalp: @vrothberg Any chance you can make these changes and push to Antonio's branch? or open a new PR carrying all the changes? Thanks!

vrothberg: @mrunalp, I opened https://github.com/kubernetes-sigs/cri-o/pull/2059. Feel free to close this one here.

mrunalp:
Closing in favor of #2059

PR #2032, opened 2019-01-23
Validate system registries configuration on start

saschagrunert:
This PR adds another runtime validation for the `/etc/containers/registries.conf`. The benefit is now that the user gets early feedback on cri-o start (for example if the user has mixed up the v1 and v2 syntax).

Currently this would only considered on image pull (which could be too late).

Beside that I have a question: Is it intended that we switch to sysregistriesv2 for `lib/config.go`? It should be possible since there is a backwards compatibility from v2 to v1, but would imply breaking changes within the configuration syntax of cri-o.

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/pkg/sysregistries"
+	"github.com/containers/image/pkg/sysregistriesv2"
 	"github.com/containers/image/types"
 	"github.com/containers/libpod/pkg/spec"
 	"github.com/containers/storage"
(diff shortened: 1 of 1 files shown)


lib/config.go
```diff
@@ -514,6 +515,11 @@ version 1.14, this option will no longer exist.
 			logrus.Debugf("found valid runtime '%s' for runtime_path '%s'\n",
 				runtime, handler.RuntimePath)
 		}
+
+		// Validate the system registries configuration
+		if _, err := sysregistriesv2.GetRegistries(nil); err != nil {
+			return fmt.Errorf("invalid /etc/containers/registries.conf: %q", err)
+		}
```
TomSweeneyRedHat: The code looks fine to me, my only question is if it is valid to pass 'nil' to GetRegistries here?  I know we do so for the tests, but thought we'd need a systemcontext for non tests.  @vrothberg WDYT?

runcom: that should be totally fine, we're passing `nil` pretty much anywhere else afaict (see #2028)

mrunalp:
@vrothberg @runcom Is this good to merge?

vrothberg:
> @vrothberg @runcom Is this good to merge?

Yes, LGTM. Thanks, @saschagrunert !

PR #2039, opened 2019-01-29
fixup vndr cleanup

runcom:
Signed-off-by: Antonio Murdaca <runcom@linux.com>

**- What I did**

- added a `Makefile` target to whitelist `ginkgo` and possibly others in the future when running `vndr`
- ran `vndr` to clean up state as of today
- added docs to contributing.md for contributors

**- How I did it**

vim

**- How to verify it**

check it out + trust the CI


saschagrunert:
Awesome, thanks for cleaning it up and making clear how to vendor the dependencies. :+1: 

Do we want to add an additional step to travis where we failure on vendor directory diffs?

runcom:
> Do we want to add an additional step to travis where we failure on vendor directory diffs?

I'll tackle that as a follow up, let's bring back master to a clean state :smile:

runcom:
@mrunalp @vrothberg @sboeuf  ptal

runcom:
also, don't worry about the 399k deletion, that's good anyway, tests pass so we're good. I'll follow up with a vendor check + additional cleanup

CONTRIBUTING.md
```diff
@@ -71,6 +71,23 @@ when the PR is merged.
...
 Some maintainers add themselves to [`CODEOWNERS`](.github/CODEOWNERS) to manage their [review notifications][code-owners], but those entries have no governance significance.
 
+### Dependencies management
+
+Make sure [`vndr`](https://github.com/LK4D4/vndr) is installed.
+
+In order to add a new dependency to this project:
+
+- add a new line to `vendor.conf` according to `vndr` rules (e.g. `github.com/pkg/errors master`)
+- run `make PKG=github.com/pkg/errors vendor`
```
vrothberg: I fear the `PKG` variable will be seen as an invite to update only **some** dependencies but not **all**.

I prefer to not expose the variable and always run `make vendor` which runs `vndr -whitelist` on the entire `vendor.conf`. In theory, only the altered dependencies will be updated (assuming non are pulling from a branch). If  others are updated as well, then there's something wrong in the `vendor.conf` but we're more likely to catch those.

vrothberg: Sounds great. I am cool with having this a temporary fix!

PR #2042, opened 2019-01-29
Fix apparmor related integration tests

### LINKED ISSUE #2040: Apparmor integration tests
During the port of the integration tests to other distributions I found out that the apparmor related tests (`test/apparmor.bats`) are skipped and not working anymore. Is it intended to enable this later on Fedora and RHEL? I could have a look into this in some time.

saschagrunert:
Hey, this PR fixes the apparmor related integration tests (currently disabled within the pipeline). I verified that it works with openSUSE Kubic. The main test appraoch changed a bit since it now uses the `apparmor_profile` field directly within the `security_context`. The plan is to run these tests later on a regular basis, too.

Closes #2040

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ RUN apt-get update && apt-get install -y \
     iptables \
     pkg-config \
     libaio-dev \
+    libapparmor-dev \
     libcap-dev \
     libfuse-dev \
     libnet-dev \
@@ -67,7 +68,7 @@ RUN set -x \
 	&& cd "$GOPATH/src/github.com/opencontainers/runc" \
 	&& git fetch origin --tags \
 	&& git checkout -q "$RUNC_COMMIT" \
-	&& make static BUILDTAGS="seccomp selinux" \
+	&& make static BUILDTAGS="seccomp selinux apparmor" \
 	&& cp runc /usr/bin/runc \
 	&& rm -rf "$GOPATH"
 
--- a/test/apparmor.bats
+++ b/test/apparmor.bats
@@ -17,7 +17,7 @@ function teardown() {
 
     start_crio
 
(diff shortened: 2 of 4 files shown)


mrunalp:
Any chance this can be tested on Ubuntu as well? :)

saschagrunert:
> Any chance this can be tested on Ubuntu as well? :)

I think so and will evaluate it asap.

vrothberg:
Thanks, @saschagrunert :pray: 

> Any chance this can be tested on Ubuntu as well? :)

That would be a great test vector! Podman CI just passed with enabled AppArmor on Ubuntu :+1:

saschagrunert:
@vrothberg @mrunalp It seems that travis [disabled apparmor completely](https://github.com/travis-ci/travis-cookbooks/blob/5a634d9e3cba098f817686c3fc0cfe207e88bb54/cookbooks/travis_docker/files/default/etc/default/grub.d/99-travis-settings.cfg#L3) from its ubuntu vms. I verified locally that it works on xenial. Are you able to setup an ubuntu on your Jenkins instance? I mean the (docker based) integration tests take quite a while on travis so we could move them elsewhere, too.

mrunalp:
@saschagrunert Yeah that is on our list but we won't be able to get to it soon :/

vrothberg:
> Okay, so I reverted my last change on the Makefile to make it work again. I'm not sure which test scenarios should be supported for ubuntu. I would assume that everything which currently runs on Fedora/RHEL, correct? For that the `localintegration` job can stay as it currently is.

That sounds good to me.

> Are you planning to add ubuntu to the ansible config or are you considering another way? I'm currently working on a terraform/libvirt appraoch which could be used for ubuntu, too.

I don't know but I think that a consistent way of provisioning would be good. @saschagrunert, CRI-O uses libpod's AppArmor package and libpod's CI has Ubuntu VMs with AppArmor enabled. To increase the testing vector a now, we could add integration tests to libpod. Feel free to open a PR if you have time. We just need to make sure that those are only executed when AppArmor is enabled.

saschagrunert:
Sounds good, for now I would need my proposed changes for testing purposes on openSUSE. I will have a look into libpods testing nevertheless, because we want to test this for openSUSE (including apparmor) too.

PR #2043, opened 2019-01-29
[1.12] Add container description to conmon

mrunalp:
Add a description of the form k8s namespaces/pod/container
for the conmon scope to make it easier to associate container IDs
to k8s objects for debugging.

@giuseppe @vrothberg @umohnani8 @runcom ptal.

### DIFF
--- a/oci/container.go
+++ b/oci/container.go
@@ -14,6 +14,7 @@ import (
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"k8s.io/apimachinery/pkg/fields"
 	pb "k8s.io/kubernetes/pkg/kubelet/apis/cri/runtime/v1alpha2"
+	"k8s.io/kubernetes/pkg/kubelet/types"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


oci/container.go
```diff
@@ -305,3 +306,8 @@ func (c *Container) SetCreated() {
 func (c *Container) Created() bool {
 	return c.created
 }
+
+// Description returns a description for the container
+func (c *Container) Description() string {
+	return fmt.Sprintf("%s/%s/%s", c.Labels()[types.KubernetesPodNamespaceLabel], c.Labels()[types.KubernetesPodNameLabel], c.Labels()[types.KubernetesContainerNameLabel])
```
vrothberg: Is that safe or do we need `path.Join(....)` instead?

umohnani8:
LGTM

oci/container.go
rhatdan: Well if this is not a path, then it potentially could cause issues.  I guess only difference was on Windows it would pring %s\%s\%s

rhatdan:
LGTM
Not sure if these needs to go through full tests.

oci/container.go
mrunalp: This is only a description string that shows up when one does crio-conmon-id.scope. That makes it easier to debug.

PR #2045, opened 2019-01-30
Fix network test failure

saschagrunert:
Hey, this fix makes the network test a bit more robust. I encountered network names like `last_reserved_ip.0` which will cause the test to fail but should not harm the overall test result.


TomSweeneyRedHat:
LGTM

PR #2047, opened 2019-01-31
[1.12] Move config structs to pkg/config

umohnani8:
Need to use these structs in the CRD and controller
for container runtime config. Moving the structs to
a separate package to reduce dependencies.


umohnani8:
@mrunalp PTAL

umohnani8:
Tests are green. @mrunalp @runcom @rhatdan PTAL

PR #2057, opened 2019-02-06
Added traces to CRI API pertaining to containers and sandboxes

haircommander:
Sandbox and Container CRI commands now log debugging information about the infra-container or container they've just operated upon.

Including adding function in util that translates labels in a sandbox run/container create request into a description. This allows the user to see requests to create a container even if it's failed.

Is currently a WIP because I'm looking for feedback on locations of the logging (for most commands once it's succeeded because there is no information about which container until it's been queried from the server, except for requests with a config), and because there is a single line of code duplication (formatting the description) that I am looking to get rid of somehow.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -470,6 +470,7 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 		recordError(operation, err)
 	}()
 	logrus.Debugf("CreateContainerRequest %+v", req)
+	logrus.Infof("Attempting to create container: %s", translateLabelsToDescription(req.GetConfig().GetLabels()))
 
 	s.updateLock.RLock()
 	defer s.updateLock.RUnlock()
@@ -554,6 +555,7 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 
 	container.SetCreated()
 
+	logrus.Infof("Created container: %s", container.Description())
 	resp := &pb.CreateContainerResponse{
 		ContainerId: containerID,
 	}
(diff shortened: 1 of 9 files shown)


haircommander:
@mrunalp PTAL

mrunalp:
@haircommander Can you sign the CLA?

server/container_create.go
```diff
@@ -470,6 +470,7 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 		recordError(operation, err)
 	}()
 	logrus.Debugf("CreateContainerRequest %+v", req)
+	logrus.Debugf("Attempting to create container: %s", translateLabelsToDescription(req.GetConfig().GetLabels()))
```
mrunalp: We need the traces to be always visible so use a log level that is visible by default. I think Infof should be visible without turning on debug.

haircommander: @mrunalp Infof still needs --log-level info when calling crio. Now, it's easier to see the traces if they're info and differentiated from all of the debugging info, so I can leave it. but it won't be default.

server/container_status.go
```diff
@@ -107,6 +107,7 @@ func (s *Server) ContainerStatus(ctx context.Context, req *pb.ContainerStatusReq
 		}
 	}
 
+	logrus.Infof("Found status for pod sandbox: %s", c.Description())
```
mrunalp: I think we can remove this one.

server/sandbox_status.go
```diff
@@ -62,6 +62,7 @@ func (s *Server) PodSandboxStatus(ctx context.Context, req *pb.PodSandboxStatusR
 		}
 	}
 
+	logrus.Infof("Found status for pod sandbox with infra container: %s", podInfraContainer.Description())
```
mrunalp: We can remove this one as well.

mrunalp:
Looks fine beside the 2 comments :+1:

mrunalp:
@haircommander can you port this to 1.13 and master as well?

PR #2059, opened 2019-02-07
Blocking

vrothberg:
Kicking https://github.com/kubernetes-sigs/cri-o/pull/2028 over the finish line.

### DIFF
--- a/pkg/storage/image.go
+++ b/pkg/storage/image.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 	"path"
 	"strings"
@@ -553,9 +554,17 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
 		}
 		return nil, err
 	}
+
 	domain, remainder := splitDockerDomain(imageName)
 	if domain != "" {
 		// this means the image is already fully qualified
+		registry, err := sysregistriesv2.FindRegistry(nil, imageName)
+		if err != nil {
+			return nil, err
+		}
+		if registry != nil && registry.Blocked {
+			return nil, fmt.Errorf("cannot use %q because it's blocked", imageName)
+		}
 		return []string{imageName}, nil
 	}
 	// we got an unqualified image here, we can't go ahead w/o registries configured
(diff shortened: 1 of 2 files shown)


vrothberg:
@mtrmac @mrunalp PTAL

pkg/storage/image.go
```diff
@@ -553,9 +554,17 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
...
 	domain, remainder := splitDockerDomain(imageName)
+	// this means the image is already fully qualified
 	if domain != "" {
-		// this means the image is already fully qualified
+		registry, err := sysregistriesv2.FindRegistry(nil, imageName)
+		if err != nil {
+			return nil, err
+		}
+		if registry != nil && registry.Blocked {
+			return nil, fmt.Errorf("cannot resolve names from registry %q because it's blocked", imageName)
```
mtrmac: (Non-blocking nit: `imageName` is not a registry.  “cannot use %q because …”?)

pkg/storage/image.go
```diff
@@ -572,7 +581,15 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
...
 		}
-		images = append(images, path.Join(r, rem))
+		image := path.Join(r, rem)
+		registry, err := sysregistriesv2.FindRegistry(nil, image)
+		if err != nil {
+			return nil, err
+		}
+		if registry != nil && registry.Blocked {
+			continue
+		}
```
mtrmac: One more thing… the comment above the function says
> Will only return an empty slice if err != nil.

and it seems that various callers expect this to be true.  So, after the loop, if `images` ends up empty, this should return an error (“all search locations for $remainder are blocked”) or something like that.

pkg/storage/image.go
```diff
@@ -553,9 +554,17 @@ func (svc *imageService) ResolveNames(imageName string) ([]string, error) {
 		}
 		return nil, err
 	}
+
 	domain, remainder := splitDockerDomain(imageName)
+	// this means the image is already fully qualified
```
mtrmac: (Absolutely non-blocking: The old location of this comment inside the `if` was a bit clearer.)

pkg/storage/image.go
vrothberg: done

pkg/storage/image.go
vrothberg: done

pkg/storage/image.go
vrothberg: done

mtrmac:
LGTM, for the record.

mrunalp:
Can we open a PR to release-1.12 for this?

mrunalp:
Rebasing to master should fix the tests here.

vrothberg:
@mrunalp do I have to repush to kick-off the kata job?

vrothberg:
@mrunalp, it's finally green \o/

PR #2060, opened 2019-02-07
journald log driver

haircommander:
Laid groundwork for journald log driver, including:
Changed the log-path to allow for a colon delimited <DRIVER>:<PATH>, with two options for the <DRIVER>:
    kubernetes-log-file: Write output to <PATH> logs in kubernetes .log format
    journald: Write output to journald, and ignore <PATH>
if there is no colon in the log-path, the driver will be kubernetes-log-file to maintain backwards compatibility

Made changes where the log path was assumed to be set, as that assumption is no longer correct. For instance, we don't attempt to open the log_fd, nor attempt to call write_k8s_log

Adding PR here to allow for CI to run against it (per @giuseppe 's suggestion). As of now, this is just a conmon change, and not a change with CRI-O, so the functionality isn't wired all together yet.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -19,6 +19,7 @@
 #include <sys/ioctl.h>
 #include <termios.h>
 #include <syslog.h>
+#include <systemd/sd-journal.h>
 #include <unistd.h>
 #include <inttypes.h>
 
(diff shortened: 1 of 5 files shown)


conmon/conmon.c
```diff
@@ -260,6 +262,49 @@ static log_level_t parse_level(char *level_name)
...
+ * in the case of no colon, the driver will be kubernetes-log-file,
+ * in the case the log driver is 'journald', the <PATH_NAME> is ignored.
+ * exits with error if <DRIVER_NAME> isn't 'journald' or 'kubernetes-log-file'
+ */
+void parse_log_path(char *log_config) {
+  char * driver = strtok(log_config, ":");
+  char * path = strtok(NULL, ":");
+  // If no : was found, driver is the log path, and the driver is
+  // kubernetes-log-file, set variables appropriately
+  if (path == NULL) {
```
giuseppe: this function should follow the same indentation as the rest of the file

conmon/conmon.c
```diff
@@ -740,9 +790,19 @@ static bool read_stdio(int fd, stdpipe_t pipe, gboolean *eof)
 		nwarnf("stdio_input read failed %s", strerror(errno));
 		return false;
 	} else {
-		if (write_k8s_log(log_fd, pipe, buf, num_read) < 0) {
-			nwarn("write_k8s_log failed");
-			return G_SOURCE_CONTINUE;
+		if (g_log_driver == JOURNALD) {
```
giuseppe: here we can use a `switch (g_log_driver)` where the `default` case exits with an error

haircommander:
closed in favor of https://github.com/kubernetes-sigs/cri-o/pull/2078

PR #1943, opened 2018-12-04
lib/container_server: Deprecate implicit hook directories

wking:
The previous approach was breaking when [the default hook directory was missing][1]:

```
crio[15828]: level=debug msg="reading hooks from /usr/share/containers/oci/hooks.d"
crio[15828]: level=debug msg="added hook /usr/share/containers/oci/hooks.d/test.json"
crio[15828]: level=debug msg="reading hooks from /etc/containers/oci/hooks.d"
crio[15828]: level=warning msg="failed to load hooks: open /etc/containers/oci/hooks.d: no such file or directory"
```

With this commit, I've deprecated the implicit defaults in favor of having users explicitly configure their hook directories (either via `--hooks-dir` on the command line or `hooks_dir` in their config file).  This is similar to the recent containers/libpod@a4b483c8 (containers/libpod#1920).

Different from the libpod implementation is that CRI-O is only loading the hook manager at this point, so we don't know if the hook directories actually have hooks that will apply to future container (or even if they have hooks at all).  I've just warned about their existence if the directories exist, to help poke folks into migration to explicitly-configured directories.

Another difference is that we get a single manager, so we need to ensure the default directories will work before feeding them in.  I'm now `stat`'ing them to check, which will address #1942.  But that's racy, since the directory could exist at stat time but be gone by the time the manager tries to load it.  I'm ok with the race, because that entire code path is deprecated anyway.

There's no race for the "user explicitly asked me to load `/whatever` and now it's gone" case.  Now that the default case is pre-checking, I've tightened 50d49939 (#1517)'s load-failure warning down to a fatal error.

I've dropped the trailing `path` from the old, hidden `--hooks-dir-path`, because I think `dir`(ectory) is already enough context for "we expect a path argument".  I consider this name change non-breaking because the old form was undocumented.  I've also changed `hooks_dir_path` to `hooks_dir` and changed it from a `string` to a `[]string`.  That is potentially a breaking change because `hooks_dir_path` became publicly documented in 50539c31 (#1749).  But it's only been a few months since it landed, and making that setting public was only one of a few changes it had made.  So I'm reasonably optimistic that there aren't (m)any users explicitly setting `hooks_dir_path` yet, and haven't gone through the trouble of continuing to support the old setting.

Fixes #1942

[1]: https://github.com/kubernetes-sigs/cri-o/issues/1942#issue-387401211

PR #2061, opened 2019-02-07
Backport hooks fix to release-1.12

QiWang19:
Signed-off-by: W. Trevor King <wking@tremily.us>

**- How I did it**
Add same changes in #1943 to release-1.12

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -178,8 +178,9 @@ default_capabilities = [
 default_sysctls = [
 {{ range $sysctl := .DefaultSysctls}}{{ printf "\t%q, \n" $sysctl}}{{ end }}]
 
-# Path to the OCI hooks directory for automatically executed hooks.
-hooks_dir_path = "{{ .HooksDirPath }}"
+# Path to OCI hooks directory for automatically executed hooks.
+hooks_dir = [
+{{ range $hooksDir := .HooksDir }}{{ printf "\t%q,\n" $hooksDir}}{{ end }}]
 
 # List of default mounts for each container. **Deprecated:** this option will
 # be removed in future versions in favor of default_mounts_file.
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -15,7 +15,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containers/libpod/pkg/hooks"
 	_ "github.com/containers/libpod/pkg/hooks/0.1.0"
 	"github.com/containers/storage/pkg/reexec"
 	units "github.com/docker/go-units"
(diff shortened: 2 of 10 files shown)


mrunalp:
Can you try git cherry-pick for the commit? That way we can retain and easily tell that the commit has been backported in the future.

QiWang19:
I run `git cherry-pick 461af2814b4cd0b879f25d15cd977c507470aacd`
Did I do it right?

mrunalp:
The commit title and author appear to have changed.

mrunalp:
@QiWang19 Can you `git commit --amend` and match the original commit title and message? You can add your own Signed-off-by to it.

```
lib/container_server: Deprecate implicit hook directories

The previous approach was breaking when the default hook directory was
missing [1]:

  crio[15828]: level=debug msg="reading hooks from /usr/share/containers/oci/hooks.d"
  crio[15828]: level=debug msg="added hook /usr/share/containers/oci/hooks.d/test.json"
  crio[15828]: level=debug msg="reading hooks from /etc/containers/oci/hooks.d"
  crio[15828]: level=warning msg="failed to load hooks: open /etc/containers/oci/hooks.d: no such file or directory"

With this commit, I've deprecated the implicit defaults in favor of
having users explicitly configure their hook directories (either via
--hooks-dir on the command line or hooks_dir in their config file).
This is similar to the recent containers/libpod@a4b483c
(libpod/container_internal: Deprecate implicit hook directories,
2018-12-02, containers/libpod#1920).

Different from the libpod implementation is that CRI-O is only loading
the hook manager at this point, so we don't know if the hook
directories actually have hooks that will apply to future container
(or even if they have hooks at all).  I've just warned about their
existence if the directories exist, to help poke folks into migrating
to explicitly-configured directories.

Another difference is that we get a single manager, so we need to
ensure the default directories will work before feeding them in.  I'm
now stat'ing them to check, which will address [1].  But that's racy,
since the directory could exist at stat time but be gone by the time
the manager tries to load it.  I'm ok with the race, because that
entire code path is deprecated anyway.

There's no race for the "user explicitly asked me to load /whatever
and now it's gone" case.  Now that the default case is pre-checking,
I've tightened 50d4993 (lib: Use libpod's hooks package,
2018-04-07, #1517)'s load-failure warning down to a fatal error.

I've dropped the trailing "path" from the old, hidden
--hooks-dir-path, because I think "dir(ectory)" is already enough
context for "we expect a path argument".  I consider this name change
non-breaking because the old form was undocumented.  I've also changed
hooks_dir_path to hooks_dir and changed it from a string to a
[]string.  That is potentially a breaking change because
hooks_dir_path became publicly documented in 50539c3 (crio.conf(5):
update manpage to the latest state, 2018-08-22, #1749).  But it's only
been a few months since it landed, and making that setting public was
only one of a few changes it had made.  So I'm reasonably optimistic
that there aren't (m)any users explicitly setting hooks_dir_path yet,
and haven't gone through the trouble of continuing to support the old
setting.

[1]: #1942 (comment)

Signed-off-by: W. Trevor King <wking@tremily.us>
Signed-off-by: Signed-off-by: Qi Wang <qiwan@redhat.com>

```

PR #2062, opened 2019-02-08
conmon: allow to specify additional arguments to the OCI runtime

giuseppe:
we can specify multiple arguments that are passed down to the OCI
runtime, before the command name:

--runtime-arg --debug

will be used as:

RUNTIME --debug COMMAND ...

Needed by: https://github.com/containers/libpod/issues/2285

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -92,6 +92,7 @@ static char *opt_exec_process_spec = NULL;
 static gboolean opt_exec = FALSE;
 static char *opt_restore_path = NULL;
 static gchar **opt_restore_args = NULL;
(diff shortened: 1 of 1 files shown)


vrothberg:
LGTM, very elegant!

mheon:
LGTM, will be helpful for libpod work

mrunalp:
We need to track the upstream commit that broke e2e. Most of them are failing with a similar issue.

mrunalp:
Rebasing to master should fix the tests here.

mrunalp:
I think rebase to master should work now after latest CI fixes.

PR #2077, opened 2019-02-19
Increase max grpc message size

mrunalp:

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -454,7 +454,12 @@ func main() {
 			logrus.Fatalf("failed to listen: %v", err)
 		}
 
-		s := grpc.NewServer()
+		// The default grpc msg size is 4MB.
+		// We bump it up to 16MB for allowing more containers.
+		maxMsgSize := 16 * 1024 * 1024
+		s := grpc.NewServer(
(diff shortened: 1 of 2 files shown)


cmd/crio/main.go
```diff
@@ -454,7 +454,12 @@ func main() {
 			logrus.Fatalf("failed to listen: %v", err)
 		}
 
-		s := grpc.NewServer()
+		// The default grpc msg size is 4MB.
+		// We bump it up to 16MB for allowing more containers.
+		maxMsgSize := 16 * 1024 * 1024
+		s := grpc.NewServer(
+			grpc.MaxMsgSize(maxMsgSize),
+		)
```
TomSweeneyRedHat: is this anything that someone would want to toggle via crio.conf?

TomSweeneyRedHat:
Changes LGTM, but tests aren't hip

cmd/crio/main.go
mrunalp: We added this in newer branches with crio.conf tunables. We don't want to support that for 1.9.

giuseppe:
/test e2e_fedora
/test e2e_rhel

LGTM

PR #2078, opened 2019-02-19
[1.13] journald log driver

haircommander:
Allowed for user to flag --log-to-journal to log container output to journald as well as kubernetes logs (in conmon).

(this is just a CI requirements test, still need to wire it up in cri-o)

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/conmon/utils.c
+++ b/conmon/utils.c
@@ -1,0 +1,28 @@
+#include "utils.h"
+#include <string.h>
+
+/* Set the log level for this call. log level defaults to warning.
+   parse the string value of level_name to the appropriate log_level_t enum value
+*/
+void set_conmon_logs(char *level_name, char *cid_, gboolean syslog_)
+{
+	cid = cid_;
+	use_syslog = syslog_;
+	// log_level is initialized as Warning, no need to set anything
+	if (level_name == NULL)
+		return;
+	if (!strcmp(level_name, "error") || !strcmp(level_name, "fatal") || !strcmp(level_name, "panic")) {
+		log_level = EXIT_LEVEL;
+		return;
+	} else if (!strcmp(level_name, "warn") || !strcmp(level_name, "warning")) {
+		log_level = WARN_LEVEL;
+		return;
+	} else if (!strcmp(level_name, "info")) {
+		log_level = INFO_LEVEL;
+		return;
+	} else if (!strcmp(level_name, "debug")) {
+		log_level = DEBUG_LEVEL;
+		return;
+	}
+	nexitf("No such log level %s", level_name);
+}
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -25,6 +25,7 @@ crio
 [--log=[value]]
 [--log-format value]
 [--log-level value]
+[--log-journald]
 [--metrics-port value]
 [--pause-command=[value]]
 [--pause-image=[value]]
@@ -117,6 +118,8 @@ If `hooks_dir` is unset, CRI-O will currently default to `/usr/share/containers/
 **--log-level**="": log crio messages above specified level: debug, info, warn, error (default), fatal or panic
 
 **--log-size-max**="": Maximum log size in bytes for a container (default: -1 (no limit)). If it is positive, it must be >= 8192 (to match/exceed conmon read buffer).
+
+**--log-journald**: log to systemd journal in addition to the kubernetes log specified with **--log**
 
 **--metrics-port**="": Port for the metrics endpoint (default: 9090)
 
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -279,6 +279,7 @@ function teardown() {
 	stop_crio
 }
 
+
 @test "ctr logging" {
 	start_crio
 	run crictl runp "$TESTDATA"/sandbox_config.json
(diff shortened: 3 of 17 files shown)


haircommander:
@mrunalp @sboeuf I think the Kata tests need to include libsystemd (or libsystemd-dev if Ubuntu based) to pass. I can open a PR but I'm not sure where those tests run from. Is it https://github.com/kata-containers/agent ?

haircommander:
@giuseppe I added conditional compilation of the journald content to get around systems that aren't compatible with systemd (or ones where it's too difficult to pin down what the package is actually called). Do you think it would be better to have compile a dummy sd_journal_print() instead of having the #ifdefs in the code? I'm not married to either solution.

mrunalp:
@chavafg Can you help with kata integration for this?

giuseppe:
any reason why this is for release-1.13 and not master?

The code LGTM

haircommander:
It was requested for OCP 4.1 which, as I understand, will be running 1.13. I've been prioritising here and will also add to master once code here is approved

haircommander:
Huzzah happy tests! @mrunalp @umohnani8 PTAL :)

giuseppe:
> It was requested for OCP 4.1 which, as I understand, will be running 1.13. I've been prioritising here and will also add to master once code here is approved

it should really not make any difference, but could you already open it so we can we sure tests pass on master too?

umohnani8:
LGTM

mrunalp:
Let us add some bats tests for this as well. The test can be similar to the logging tests that we have today but can also read from the journal to check that the logs are written correctly to the journal.

haircommander:
@mrunalp okay I will try. I think they will have to be optional tests that run only if the system has systemd installed (similar to how conmon is compiled). I will see what I can do.

mrunalp:
@haircommander Yes, that should work :+1:

docs/crio.8.md
```diff
@@ -25,6 +25,7 @@ crio
 [--log=[value]]
 [--log-format value]
 [--log-level value]
+[--log-to-journald]
```
sboeuf: Seems a bit long, I would use `log-journald` or simply `journald` to enable logs to be sent to the journal.

haircommander: I agree, I'm pretty iffy about the verbosity too.

haircommander: opted for log-journald because it seems clear but shorter!

docs/crio.8.md
```diff
@@ -25,6 +25,7 @@ crio
 [--log=[value]]
 [--log-format value]
 [--log-level value]
+[--log-journald]
```
rhatdan: Docker Daemon calls this the log-driver  I would go with that, and jsut have kubernetes, journald as the two values.
```
--log-driver="json-file|syslog|journald|gelf|fluentd|awslogs|splunk|etwlogs|gcplogs|none"
         Default driver for container logs. Default is json-file.
         Warning: docker logs command works only for json-file logging driver.
```

haircommander: @rhatdan the language in the card: https://jira.coreos.com/browse/RUN-258 says "as well as". I'm not sure the exact use case, but it could be that someone wants it logged in both. Further, there are a lot of assumptions in conmon that there is a log-path to write to as well. Originally, my PR had them exclusive, but in the interest of simplicity and requirement fulfilment, went for this.
WDYT?

rhatdan: OK Works for me. If this is what @mrunalp wants.

mrunalp: @rhatdan @haircommander For CRI-O, we always need to write to the k8s log file. However, that may not be true for podman. So, I think we can go with your suggestion of using --log-driver and make it accept multiple options. For CRI-O, we will always pass k8s CRI option. So, something like `--log-driver=cri-file --log-driver=journal` or `--log-driver=cri-file,journal`. For podman, it could be just `--log-driver=journal` or any other drivers that we choose to support.

haircommander: Roger roger I'll look into it

haircommander:
alright, now journald or cri-file can be chosen, as well as both at the same time. I left the crio flag --log-journald, as we always want to write to the cri file. @mrunalp @rhatdan @giuseppe @sboeuf PTAL

docs/crio.8.md
giuseppe: I think the best would be a single `--log-path` option, as:

 `--log-path=journal:`
`--log-path=k8s:/path/to/the/file`
`--log-path=/path/to/the/file` (default to k8s)

so if in the future we want to add more logging drivers we can simply specify a new prefix, and at the same time keep the ability to log to multiple destinations.

haircommander:
OK folks, I kinda went wild here following the advice. Container logging is now in a separate file, as well as some utility functions that are needed in both conmon.c and ctr_logging.c . There's a set of interface functions for conmon call to log any container output into any/all of the logging drivers. Eventually, this will allow for more flexibility with container logging. I will look into adding JSON as a proof of concept of that as well. 
following @giuseppe 's advice, the flag in conmon is now --log-path <driver>:<path> or --log-path <path> which defaults to k8s format. I like this, because it will allow us to write to JSON and k8s at the same time, to different files, with a clear CLI. However, I left the flag in CRI-O as --log-journald, as we always want to write to k8s. When I eventually port this to podman, I will likely follow conmon more closely in CLI. 

PTAL :) @rhatdan @mrunalp

conmon/utils.c
```diff
@@ -0,0 +1,28 @@
+#include "utils.h"
+
+/* Set the log level for this call. log level defaults to warning.
+   parse the string value of level_name to the appropriate log_level_t enum value
+*/
+void set_conmon_logs(char *level_name, char *cid_, gboolean syslog_)
+{
+	// TODO is this safe? strcpy?
```
haircommander: @giuseppe Considering the cid is always passed from main as the global bool, does the utils library need to copy it, or can it just use the same pointer? I worry about the CID being passed from a different variable down the line, and some funky seg fault action happening, but as written that won't happen.

test/ctr.bats
```diff
@@ -240,6 +242,54 @@ function teardown() {
...
+	echo "$output"
+	[ "$status" -eq 0 ]
+	run wait_until_exit "$ctr_id"
+	[ "$status" -eq 0 ]
+	run crictl rm "$ctr_id"
+	echo "$output"
+	[ "$status" -eq 0 ]
+
+    # priority of 5 is LOG_NOTICE
+	journalctl -t conmon -p 5 | grep -E "here is some output$"
```
mrunalp: How do we distinguish from all the pods/container running on the system?

haircommander: hm yeah lemme try to put some uniqueness into this

haircommander: fixed!

giuseppe:
LGTM

haircommander:
@mrunalp  @rhatdan PTAL

rhatdan:
LGTM
But @mrunalp Needs final say.

haircommander:
@mrunalp can now be filtered by ID like:
journalctl MESSAGE_ID=<ctr_id> :) PTAL!

haircommander:
@mrunalp PTAL when you get a chance :)

test/ctr.bats
```diff
@@ -283,6 +284,60 @@ function teardown() {
...
 }
 
+@test "ctr journald logging" {
+    # ensure we have journald logging capability
+	enabled=$(check_journald)
+	if [[ "$enabled" -ne 0 ]]; then
+		skip "journald not enabled"
+	fi
+
+	start_crio_journald
```
mrunalp: nit: the indentation is off in this file.

haircommander:
currently working on some small fixes in master, which I'd like to integrate into here when they're hammered out:
https://github.com/kubernetes-sigs/cri-o/pull/2171 
/hold

haircommander:
/hold cancel

updated to have newest journald content :)

mrunalp:
We don't need this for 1.13 anymore. Closing it for now.

PR #2083, opened 2019-02-21
server: delete the container if it cannot be restored

giuseppe:
If a pod or a container could not be restored, then delete them
from the storage.

Since the container was present in the storage and CRI-O didn't report
it to the the Kubelet, we would have failed the request from the
Kubelet to re-create it over and over again.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1677509

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -153,12 +153,14 @@ func (s *Server) restore() {
 	}
 	pods := map[string]*storage.RuntimeContainerMetadata{}
 	podContainers := map[string]*storage.RuntimeContainerMetadata{}
+	names := map[string][]string{}
 	for _, container := range containers {
 		metadata, err2 := s.StorageRuntimeServer().GetContainerMetadata(container.ID)
 		if err2 != nil {
 			logrus.Warnf("error parsing metadata for %s: %v, ignoring", container.ID, err2)
 			continue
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM, assuming happy tests that don't look happy atm.

mrunalp:
We can probably add some bats tests for this.

mrunalp:
We have restore related tests in restore.bats.

server/server.go
```diff
@@ -153,12 +153,14 @@ func (s *Server) restore() {
 	}
 	pods := map[string]*storage.RuntimeContainerMetadata{}
 	podContainers := map[string]*storage.RuntimeContainerMetadata{}
+	names := map[string][]string{}
```
sboeuf: Nit but I'd use this instead:
```go
names := make(map[string][]string)
```

giuseppe: how much is it important for you? :-)  I am fine either ways, but honestly I'd just avoid the CI circus if possible

sboeuf: It's fine, that's just a nit :)

mrunalp:
@giuseppe We need backports for this for the bugzilla. Thanks!

PR #2090, opened 2019-02-27
Always set gid if returned from container user files

mrunalp:
Image user could have a gid that we were skipping.


mrunalp:
@giuseppe @rhatdan ptal. We will cover tests for this as part of @umohnani8 PR.

giuseppe:
LGTM

PR #2100, opened 2019-03-04
test: Switch to golang 1.12

mrunalp:
We can handle travis in a follow-on. 


mrunalp:
@sboeuf No, not any particular fix. We just like to keep master tracking latest golang so we catch issues earlier and fix them if required.

sboeuf:
@mrunalp that makes sense, I was just curious :)

Issue #2101, opened 2019-03-04
Return IPv6 address for hostNetwork pods


mrunalp:
The current code returns hostIp for hostNetwork pods and that uses k8s function which prioritizes IPv4 over IPv6. So, the result is that hostNetwork pods end up with IPv4 addresses in a IPv6 environment.

PR #2102, opened 2019-03-05
Add support for host ip configuration

mrunalp:
This is one possible solution to #2101 


giuseppe:
this is good to have in any case, so LGTM

PR #2103, opened 2019-03-05
Add support for host ip configuration

### LINKED ISSUE #2101: Return IPv6 address for hostNetwork pods

mrunalp:
Closes #2101 

Port of #2102 to master. @umohnani8 @rhatdan ptal.


umohnani8:
LGTM

PR #2106, opened 2019-03-07
conmon: close the stdin on EOF if not tty

giuseppe:
if we receive EOF when reading from stdin and we are not using a tty,
ignore opt_leave_stdin_open and close immediately the master stdin
fd.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1624041

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -935,7 +935,7 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
 	/* End of input */
 	conn_sock_shutdown(sock, SHUT_RD);
 	if (masterfd_stdin >= 0 && opt_stdin) {
-		if (!opt_leave_stdin_open) {
+		if (!opt_leave_stdin_open || !opt_terminal) {
 			close(masterfd_stdin);
 			masterfd_stdin = -1;
 		} else {


rhatdan:
LGTM
@mrunalp @umohnani8 PTAL

haircommander:
LGTM

mrunalp:
This will end up closing the stdin of the command. The behavior we are trying to match is that we detach without closing stdin of the container.

giuseppe:
> This will end up closing the stdin of the command. The behavior we are trying to match is that we detach without closing stdin of the container.

yes, it conflicts with the option name although is there any usage in keeping the file open if it cannot be used anymore (as it is possible when we have a tty)?  If we leave it open, the other end in the container will never be notified of the EOF.

mrunalp:
@giuseppe Yeah, I agree with your assessment that it doesn't make sense to leave it open. Let me check if we aren't relying on this behavior and if we can get the upstream test disabled. If we reach agreement on that, then we can merge this which makes much more sense.

haircommander:
@mrunalp @giuseppe where are we on this?

rhatdan:
Seems to be passing all tests.  LGTM

mrunalp:
We can't merge this till we agree and settle on upstream behavior.

rhatdan:
What is the decission, and are you talking about something in the kubernetes world?

mrunalp:
@rhatdan Yes, see - https://bugzilla.redhat.com/show_bug.cgi?id=1624041.

mrunalp:
We had a call to discuss this and we won't be going this way. Instead, we want to detach in CRI-O when leave-stdin-open is true, tty is false and we get an EOF.

giuseppe:
yes, going to close this

PR #2110, opened 2019-03-08
lib/container_server: Split locale at the first dot, etc.

wking:
This ports containers/libpod@69cb8639 (containers/libpod#2550) to CRI-O.

### DIFF
--- a/lib/container_server.go
+++ b/lib/container_server.go
@@ -30,8 +30,8 @@ import (
 )
 
 var (
-	// localeToLanguage maps from locale values to language tags.
-	localeToLanguage = map[string]string{
+	// localeToLanguageMap maps from locale values to language tags.
+	localeToLanguageMap = map[string]string{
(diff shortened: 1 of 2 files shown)


mrunalp:
@vrothberg ptal

giuseppe:
LGTM

vrothberg:
LGTM although I would prefer libpod to expose that as a public API to avoid copying code.

wking:
> LGTM although I would prefer libpod to expose that as a public API to avoid copying code.

Sounds good to me, although isn't the plan to consume higher-level libpod APIs?  I haven't been tracking the progress of that unification.

vrothberg:
> > LGTM although I would prefer libpod to expose that as a public API to avoid copying code.
> 
> Sounds good to me, although isn't the plan to consume higher-level libpod APIs? I haven't been tracking the progress of that unification.

That's correct. I hope we can start the migration soon :)

PR #926, opened 2017-09-19
Add `kpod pause` and `kpod unpause`

TomSweeneyRedHat:
Implement the ability to pause and unpause running containers.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Taking over for @rhatdan to finish this up so he can chase down other pressing issues.  This takes over for d `kpod pause` and `kpod unpause` #923

PR #2111, opened 2019-03-08
lib/stop: Handle ContainerStateStopped with a switch case

wking:
The old approach with the `if` block inside the `default` dates back to 9db7cf13 (#926), but there's no need for the extra level of nesting.  With this commit, we use the `switch` cases to handle that distinction.


wking:
I've pushed f9d8ff5c -> 05c0042b to fix [the gofmt issue][1].

[1]: https://travis-ci.org/kubernetes-sigs/cri-o/jobs/503389873#L1099

PR #2115, opened 2019-03-08
Allow configuring credentials used for pulling the pause image

mtrmac:
**- What I did**

Added support for supplying credentials necessary for pulling the pause image.

**- How I did it**

Added a `pause_image_auth_file` config option, and a corresponding `--pause-image-auth-file` CLI option.

They are disabled by default, preserving the current behavior. Users have to opt in by explicitly pointing at a config file.

**- How to verify it**

This adds GoMock-based tests for that small piece of code in `runtimeService` (but not an end-to-end test, AFAICS there isn’t any existing test for credential handling either).

Ultimately, run with `--pause-image-auth-file` in the CLI, or the `pause_image_auth_file` option (I didn’t do that yet.)

New config and CLI options allow supplying credentials necessary for pulling the pause image.

### DIFF
--- a/pkg/storage/runtime.go
+++ b/pkg/storage/runtime.go
@@ -6,14 +6,13 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/containers/image/copy"
 	istorage "github.com/containers/image/storage"
 	"github.com/containers/image/transports/alltransports"
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
 	cstorage "github.com/containers/storage"
 	"github.com/containers/storage/pkg/idtools"
-	"github.com/opencontainers/image-spec/specs-go/v1"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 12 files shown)


rhatdan:
LGTM
/retest
@mrunalp @vrothberg PTAL

pkg/storage/runtime.go
```diff
@@ -330,12 +337,12 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
...
+	return r.createContainerOrPodSandbox(systemContext, podName, podID, imageName, imageID, containerName, podID, metadataName, uid, namespace, attempt, "", idMappings)
 }
 
-func (r *runtimeService) CreateContainer(systemContext *types.SystemContext, podName, podID, imageName, imageID, containerName, containerID, metadataName string, attempt uint32, mountLabel string, idMappings *idtools.IDMappings, copyOptions *copy.Options) (ContainerInfo, error) {
-	return r.createContainerOrPodSandbox(systemContext, podName, podID, imageName, imageID, containerName, containerID, metadataName, "", "", attempt, mountLabel, idMappings, copyOptions)
+func (r *runtimeService) CreateContainer(systemContext *types.SystemContext, podName, podID, imageName, imageID, containerName, containerID, metadataName string, attempt uint32, mountLabel string, idMappings *idtools.IDMappings) (ContainerInfo, error) {
+	return r.createContainerOrPodSandbox(systemContext, podName, podID, imageName, imageID, containerName, containerID, metadataName, "", "", attempt, mountLabel, idMappings)
```
runcom: we shouldn't need the pause image auth when creating containers. The sandbox is always created first by the CRI

runcom:
> LGTM
> /retest
> @mrunalp @vrothberg PTAL

This is still WIP as per @mtrmac comment above

pkg/storage/runtime.go
mtrmac: Right, and this code does not require the `CreateContainer` callers to provide any auth of the sort.

On the contrary, it removes a `copy.Options{}` parameter (that could only have been used to provide the pause image auth, and you’re right that it would never have been used), because the only caller of this function was passing `nil`.

mtrmac:
Updated to actually pass the path along (oops), and added GoMock-based tests for that small piece of code in `runtimeService` (but not an end-to-end test, AFAICS there isn’t any existing test for credential handling either).

mrunalp:
We need this in `release-1.12` first for a release but I can review/test from here.

mtrmac:
1.12: https://github.com/kubernetes-sigs/cri-o/pull/2118
1.13: https://github.com/kubernetes-sigs/cri-o/pull/2119

Issue #2121, opened 2019-03-11
CRI-O requires podman-like "cni_default_network" support


sysrich:
**Description**

This ticket started it's life in a discussion at https://github.com/containers/libpod/issues/2370

In short

- the CNI ecosystem is imperfect, and the assumption that all configurations in /etc/cni/net.d will be compatible with each other is not true (eg. flannel not working alongside a basic podman bridge)
- any container runtime should therefore have some way of mitigating those issues
- podman achieves this with its `cni_default_network` configuration parameter, allow podman to be 'locked' to a specific CNI configuration, ignoring what might be placed in /etc/cni/net.d by other CNI solutions (assumedly to be used by other parallel container runtimes)
- It would be highly advantageous if CRI-O had similar functionality

vrothberg:
Thanks a lot for moving the issue over here, @sysrich!

We can get some inspiration from https://github.com/containers/libpod/commit/4f699db8dad0 to fix the issue in CRI-O.

haircommander:
Working on it! (I'd assign myself but I don't have write access to the repo)

weteamsteve:
has there been any updates to this? currently having problems with Kubernetes and Podman working together

PR #2041, opened 2019-01-29
[WIP] Sync vendor and add validation

runcom:
Signed-off-by: Antonio Murdaca <runcom@linux.com>

**- What I did**

- [x] sync up the vendor tree
- [ ] pin versions in `vendor.conf` to have deterministic vndr runs
- [ ] removed the `PKG` env from `make vendor`; we want to run them all to have determinism
- [ ] add validation to the vendor tree in the CI

**- How I did it**

**- How to verify it**

**- Description for the changelog**

PR #2123, opened 2019-03-11
[master] Add option to change CNI network default

haircommander:
**- What I did**
Began work on changing the network configurations from using every config file in the config dir to only ones specified.
**- How I did it**
Vendored in changes to ocicni and cni to allow for adding default network config to InitCNI

### DIFF
--- a/server/sandbox_network.go
+++ b/server/sandbox_network.go
@@ -3,7 +3,6 @@ package server
 import (
 	"fmt"
 	"net"
-	"strings"
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	cnicurrent "github.com/containernetworking/cni/pkg/types/current"
@@ -34,23 +33,17 @@ func (s *Server) networkStart(sb *sandbox.Sandbox) (podIP string, result cnitype
 		return
 	}
 
-	tmp, err := s.netPlugin.GetPodNetworkStatus(podNetwork)
+	results, err := s.netPlugin.GetPodNetworkStatus(podNetwork)
 	if err != nil {
 		err = fmt.Errorf("failed to get network status for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
 		return
 	}
 
-	// only one cnitypes.Result is returned since newPodNetwork sets Networks list empty
-	result = tmp[0]
-	logrus.Debugf("CNI setup result: %v", result)
-
-	network, err := cnicurrent.GetResult(result)
+	podIP, result, err = s.getSandboxIPAndResult(results)
 	if err != nil {
-		err = fmt.Errorf("failed to get network JSON for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
+		err = fmt.Errorf("failed to get ip for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
 		return
 	}
-
-	podIP = strings.Split(network.IPs[0].Address.String(), "/")[0]
 
 	if len(sb.PortMappings()) > 0 {
 		ip := net.ParseIP(podIP)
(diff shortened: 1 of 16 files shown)


haircommander:
Adding WIP tag because: this patch currently doesn't actually work. Further, there should be some discussion on the default config name. Lastly, the documentation has to be added if this change is breaking. I just put it here to run against CI to see if any of my changes thus far break anything.

haircommander:
This PR will also:
fix #2121
unblock #2041

haircommander:
PTAL @rhatdan @umohnani8 @mrunalp @sboeuf

rhatdan:
/approve
LGTM

server/utils.go
```diff
@@ -104,6 +104,9 @@ func newPodNetwork(sb *sandbox.Sandbox) ocicni.PodNetwork {
 		Namespace: sb.Namespace(),
 		ID:        sb.ID(),
 		NetNS:     sb.NetNsPath(),
+		//PortMappings: sb.PortMappings(),
```
umohnani8: Is this commented out because we don't have support for it yet?

haircommander: Yes along with us not using ocicni for port bindings but rather a k8s library. Though I think I meant to remove that comment

umohnani8:
@haircommander needs rebase

haircommander:
so many rebases! thanks @umohnani8, working on it

haircommander:
blocked on #2140

haircommander:
Okay happy green tests
@mrunalp @umohnani8 PTAL

server/sandbox_run_linux.go
```diff
@@ -587,15 +587,17 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
...
-		ip, result, err = s.networkStart(sb)
+		ip, results, err = s.networkStart(sb)
 		if err != nil {
 			return nil, err
 		}
-		if result != nil {
-			g.AddAnnotation(annotations.CNIResult, result.String())
+		if len(results) > 0 {
+			for _, result := range results {
+				g.AddAnnotation(annotations.CNIResult, result.String())
```
mrunalp: This is overwriting the value with the last one being the winner.

server/sandbox_network.go
```diff
@@ -81,17 +76,32 @@ func (s *Server) getSandboxIP(sb *sandbox.Sandbox) (string, error) {
...
 	if err != nil {
 		return "", fmt.Errorf("failed to get network status for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
 	}
+	return s.getSandboxIPsFromResult(results)
+}
 
-	res, err := cnicurrent.GetResult(result[0])
-	if err != nil {
-		return "", fmt.Errorf("failed to get network JSON for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
+func (s *Server) getSandboxIPsFromResult(results []cnitypes.Result) (string, error) {
```
mrunalp: Are we guaranteed that we will only get one IP address? If so, how do we know from CNI config which one is the primary?

mrunalp:
@dcbw Do you mind reviewing this?

server/sandbox_network.go
haircommander: I opted to fail if there was more than one IP found in a config, because it seemed like we can't support multiple.

server/sandbox_run_linux.go
haircommander: Should we, rather than getting all results, just choose the first, and base everything off of it? I can't think of a better way of deciding which config is the correct one to use, both for the annotation and ip

server/sandbox_network.go
mrunalp: From this PRs perspective we should be picking the default network but I want to make sure we don't break multus cc: @dcbw

rhatdan:
@haircommander Is this still something you are working on?

haircommander:
@rhatdan it was backlogged but I'd like to continue working on it.

saschagrunert:
@haircommander this needs a rebase :)

rhatdan:
@haircommander Needs a rebase.

haircommander:
on my todo, which has gotten much longer with all of these pings :smile:

rhatdan:
@giuseppe Do you have a chance to take this one over?

saschagrunert:
If @giuseppe has no time I'm keen to take this one as well.

rhatdan:
@saschagrunert Take it.  I am sure @giuseppe Has plenty of other things to do?  I always have pet projects for him.  :^) (User namespace...)

haircommander:
closing in favor of #3452

PR #2127, opened 2019-03-14
Update error handling paths for sandbox add and removal

### LINKED ISSUE #1987: Index out of bounds access with invalid `selinuxLabel`
If I call the function `LoadSandbox()` with a container manifest which contains a wrong `selinuxLabel` like 
```json
"process": {
	"selinuxLabel": "wrong"
}
```
 the result will be an out of bounds access. In `vendor/github.com/opencontainers/selinux/go-selinux/selinux_linux.go:403`.

The main issue is the call to `label.DupSecOpt()`
https://github.com/kubernetes-sigs/cri-o/blob/95ad21041c8a9f337673f97ecca7709daca8b8b2/lib/container_server.go#L356

which builds up the following call stack:
```shell
  Test Panicked
  runtime error: index out of range
  /usr/lib64/go/1.11/src/runtime/panic.go:513

  Full Stack Trace
        /usr/lib64/go/1.11/src/runtime/panic.go:513 +0x1b9
  github.com/kubernetes-sigs/cri-o/vendor/github.com/opencontainers/selinux/go-selinux.NewContext(0xc0000c5198, 0x5, 0x16)
        vendor/github.com/opencontainers/selinux/go-selinux/selinux_linux.go:403 +0x2b0
  github.com/kubernetes-sigs/cri-o/vendor/github.com/opencontainers/selinux/go-selinux.DupSecOpt(0xc0000c5198, 0x5, 0x0, 0x0, 0xc000376aa0)
        vendor/github.com/opencontainers/selinux/go-selinux/selinux_linux.go:710 +0x42
  github.com/kubernetes-sigs/cri-o/vendor/github.com/opencontainers/selinux/go-selinux/label.DupSecOpt(0xc0000c5198, 0x5, 0x8, 0x176ac40, 0xc0003b2840)
        vendor/github.com/opencontainers/selinux/go-selinux/label/label_selinux.go:189 +0x35
  github.com/kubernetes-sigs/cri-o/lib.(*ContainerServer).LoadSandbox(0xc000376460, 0x181c6ac, 0x2, 0x0, 0x0)
        lib/container_server.go:359 +0x4a8
```
If we track down the stack we can see that the out of bounds access happens in L403 here:
https://github.com/kubernetes-sigs/cri-o/blob/95ad21041c8a9f337673f97ecca7709daca8b8b2/vendor/github.com/opencontainers/selinux/go-selinux/selinux_linux.go#L396-L410

I did not create a PR yet since I wanted to discuss with you how to address this issue: In opencontainers/selinux or cri-o? My suggestion would be to fix it directly in selinux-go.

saschagrunert:
Hey, this PR updates all related error paths to actually return the new introduced error possibilities from selinux-go 1.2. Beside this the crashing unit test can now be enabled.

Fixes #1987

### DIFF
--- a/lib/container_server.go
+++ b/lib/container_server.go
@@ -311,7 +311,9 @@ func (c *ContainerServer) Update() error {
 		}
 		sb.RemoveInfraContainer()
 		c.ReleasePodName(sb.Name())
-		c.RemoveSandbox(sb.ID())
+		if err := c.RemoveSandbox(sb.ID()); err != nil {
+			logrus.Warnf("failed to remove sandbox ID %s: %v", sb.ID(), err)
+		}
 		if err = c.podIDIndex.Delete(sb.ID()); err != nil {
 			return err
 		}
@@ -417,11 +419,15 @@ func (c *ContainerServer) LoadSandbox(id string) error {
 		}
 	}
 
-	c.AddSandbox(sb)
+	if err := c.AddSandbox(sb); err != nil {
+		return err
+	}
 
 	defer func() {
 		if err != nil {
-			c.RemoveSandbox(sb.ID())
+			if err := c.RemoveSandbox(sb.ID()); err != nil {
+				logrus.Warnf("could not remove sandbox ID %s: %v", sb.ID(), err)
+			}
 		}
 	}()
 
@@ -741,12 +747,12 @@ func (c *ContainerServer) ListContainers(filters ...func(*oci.Container) bool) (
 }
 
 // AddSandbox adds a sandbox to the sandbox state store
-func (c *ContainerServer) AddSandbox(sb *sandbox.Sandbox) {
+func (c *ContainerServer) AddSandbox(sb *sandbox.Sandbox) error {
 	c.state.sandboxes.Add(sb.ID(), sb)
 
 	c.stateLock.Lock()
-	c.addSandboxPlatform(sb)
-	c.stateLock.Unlock()
+	defer c.stateLock.Unlock()
+	return c.addSandboxPlatform(sb)
 }
 
 // GetSandbox returns a sandbox by its ID
@@ -769,17 +775,20 @@ func (c *ContainerServer) HasSandbox(id string) bool {
 }
 
 // RemoveSandbox removes a sandbox from the state store
-func (c *ContainerServer) RemoveSandbox(id string) {
+func (c *ContainerServer) RemoveSandbox(id string) error {
 	sb := c.state.sandboxes.Get(id)
 	if sb == nil {
-		return
+		return nil
 	}
 
 	c.stateLock.Lock()
-	c.removeSandboxPlatform(sb)
-	c.stateLock.Unlock()
+	defer c.stateLock.Unlock()
+	if err := c.removeSandboxPlatform(sb); err != nil {
+		return err
+	}
 
 	c.state.sandboxes.Delete(id)
+	return nil
 }
 
 // ListSandboxes lists all sandboxes in the state store
--- a/lib/container_server_linux.go
+++ b/lib/container_server_linux.go
@@ -8,23 +8,31 @@ import (
 	"github.com/opencontainers/selinux/go-selinux/label"
 )
 
-func (c *ContainerServer) addSandboxPlatform(sb *sandbox.Sandbox) {
-	// NewContext() always returns nil, so we can safely ignore the error
-	ctx, _ := selinux.NewContext(sb.ProcessLabel())
-	c.state.processLevels[ctx["level"]]++
+func (c *ContainerServer) addSandboxPlatform(sb *sandbox.Sandbox) error {
+	context, err := selinux.NewContext(sb.ProcessLabel())
+	if err != nil {
+		return err
+	}
+	c.state.processLevels[context["level"]]++
+	return nil
 }
 
-func (c *ContainerServer) removeSandboxPlatform(sb *sandbox.Sandbox) {
+func (c *ContainerServer) removeSandboxPlatform(sb *sandbox.Sandbox) error {
 	processLabel := sb.ProcessLabel()
-	// NewContext() always returns nil, so we can safely ignore the error
-	ctx, _ := selinux.NewContext(processLabel)
-	level := ctx["level"]
+	context, err := selinux.NewContext(processLabel)
+	if err != nil {
+		return err
+	}
+	level := context["level"]
 	pl, ok := c.state.processLevels[level]
 	if ok {
 		c.state.processLevels[level] = pl - 1
 		if c.state.processLevels[level] == 0 {
-			label.ReleaseLabel(processLabel)
-			delete(c.state.processLevels, level)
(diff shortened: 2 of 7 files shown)


lib/container_server.go
```diff
@@ -302,7 +302,9 @@ func (c *ContainerServer) Update() error {
 		}
 		sb.RemoveInfraContainer()
 		c.ReleasePodName(sb.Name())
-		c.RemoveSandbox(sb.ID())
+		if err := c.RemoveSandbox(sb.ID()); err != nil {
```
giuseppe: on an error now we skip `c.podIDIndex.Delete(sb.ID())`.  Is it the desired effect?

lib/container_server.go
```diff
@@ -728,12 +739,15 @@ func (c *ContainerServer) ListContainers(filters ...func(*oci.Container) bool) (
...
 
 // AddSandbox adds a sandbox to the sandbox state store
-func (c *ContainerServer) AddSandbox(sb *sandbox.Sandbox) {
+func (c *ContainerServer) AddSandbox(sb *sandbox.Sandbox) error {
 	c.state.sandboxes.Add(sb.ID(), sb)
 
 	c.stateLock.Lock()
-	c.addSandboxPlatform(sb)
+	if err := c.addSandboxPlatform(sb); err != nil {
+		return err
```
giuseppe: we need to unlock `c.stateLock`

lib/container_server.go
```diff
@@ -756,17 +770,20 @@ func (c *ContainerServer) HasSandbox(id string) bool {
...
 	sb := c.state.sandboxes.Get(id)
 	if sb == nil {
-		return
+		return nil
 	}
 
 	c.stateLock.Lock()
-	c.removeSandboxPlatform(sb)
+	if err := c.removeSandboxPlatform(sb); err != nil {
+		return err
```
giuseppe: same issue here

server/sandbox_remove.go
```diff
@@ -98,7 +98,9 @@ func (s *Server) RemovePodSandbox(ctx context.Context, req *pb.RemovePodSandboxR
 	}
 
 	s.ReleasePodName(sb.Name())
-	s.removeSandbox(sb.ID())
+	if err := s.removeSandbox(sb.ID()); err != nil {
+		return nil, err
```
giuseppe: same here.  Do we want to skip `s.PodIDIndex().Delete(sb.ID())`c on an error?

lib/container_server.go
saschagrunert: No, switched to an debug warning if the removal fails.

lib/container_server.go
saschagrunert: I added it as `defer`.

lib/container_server.go
saschagrunert: I added it as `defer`.

server/sandbox_remove.go
saschagrunert: I think so, switched to a log warning instead.

lib/container_server.go
rhatdan: I am a little worried about adding defers,  Wern't these causing performance issues? Or was that in libpod.

rhatdan:
I would rather fix up https://github.com/kubernetes-sigs/cri-o/pull/2096 and get that merged then continuing to manage SELinux labels externally.

lib/container_server.go
saschagrunert: Defer is causing a significant higher performance cost than not using it, yes. If this is a problem I can change it again.

saschagrunert: But we use it for other methods of the `ContainerServer` too, so it should not be that harmful in our case.

giuseppe: @rhatdan was the issue that we were holding the lock more than it was necessary?  I'd expect most of the cost to come from there

lib/container_server.go
```diff
@@ -307,7 +307,9 @@ func (c *ContainerServer) Update() error {
 		}
 		sb.RemoveInfraContainer()
 		c.ReleasePodName(sb.Name())
-		c.RemoveSandbox(sb.ID())
+		if err := c.RemoveSandbox(sb.ID()); err != nil {
+			logrus.Warnf("failed to remove sandbox: %v", err)
```
rhatdan: Should this include the sb.ID()?

lib/container_server.go
```diff
@@ -413,11 +415,15 @@ func (c *ContainerServer) LoadSandbox(id string) error {
...
-	c.AddSandbox(sb)
+	if err := c.AddSandbox(sb); err != nil {
+		return err
+	}
 
 	defer func() {
 		if err != nil {
-			c.RemoveSandbox(sb.ID())
+			if err := c.RemoveSandbox(sb.ID()); err != nil {
+				logrus.Warnf("could not remove sandbox: %v", err)
```
rhatdan: Include sb.ID()

lib/container_server.go
```diff
@@ -737,12 +743,15 @@ func (c *ContainerServer) ListContainers(filters ...func(*oci.Container) bool) (
...
 // AddSandbox adds a sandbox to the sandbox state store
-func (c *ContainerServer) AddSandbox(sb *sandbox.Sandbox) {
+func (c *ContainerServer) AddSandbox(sb *sandbox.Sandbox) error {
 	c.state.sandboxes.Add(sb.ID(), sb)
 
 	c.stateLock.Lock()
-	c.addSandboxPlatform(sb)
-	c.stateLock.Unlock()
+	defer c.stateLock.Unlock()
+	if err := c.addSandboxPlatform(sb); err != nil {
```
rhatdan: Just return c.addSandboxPlatform(sb)

lib/container_server.go
saschagrunert: Yes. Changed id.

lib/container_server_linux.go
```diff
@@ -8,23 +8,31 @@ import (
...
+	level := context["level"]
 	pl, ok := c.state.processLevels[level]
 	if ok {
 		c.state.processLevels[level] = pl - 1
 		if c.state.processLevels[level] == 0 {
-			label.ReleaseLabel(processLabel)
+			if err := label.ReleaseLabel(processLabel); err != nil {
+				return err
+			}
 			delete(c.state.processLevels, level)
```
giuseppe: we will miss the `delete` here on an error.  Could you move it before the ReleaseLabel and use a defer?

saschagrunert: Sure, thanks for the hint!

saschagrunert:
> @saschagrunert PR looks good to me. Could you add a commit message, as I think it's important to understand what this commit is doing by simply having a look at `git log`.

Yes, I updated the commit message to contain a more detailed description.

sboeuf:
Thanks @saschagrunert

umohnani8:
@saschagrunert please rebase to pick up CI fixes

haircommander:
LGTM

PR #2128, opened 2019-03-14
Allow device mounting to work in privileged mode

haircommander:
Before, device mounting was disabled when a container was privileged, because there was no clear way to make distinctions between mounts on the host and directories in the container. Resolve this by decreeing a privileged container cannot mount a device that already exists on the host. This allows privileged container mounts to work in other cases.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1614734

Allow device mounting in a privileged container.


mrunalp:
LGTM

mrunalp:
@giuseppe @umohnani8 @rhatdan ptal

umohnani8:
LGTM

PR #2129, opened 2019-03-14
[1.13] Fix detach non tty

mrunalp:
Detach automatically on stdin EOF for a non-tty attach when stdin Once is not set.


mrunalp:
@giuseppe @rhatdan ptal

PR #2132, opened 2019-03-15
Improve pkg/storage/runtime unit tests

### LINKED ISSUE #2117: pkg/storage/runtime_test.go is mostly wrong
**Description**

The tests in pkg/storage/runtime_test.go use `imageName` as a name of an image to be pulled. Because uppercase letters are invalid in Docker references (and therefore also `containers-storage:` references), the test exercise code paths that don’t match the real world much.

1. `s/imageName/imagename/g` in `pkg/storage/runtime_test.go`
2. `go test ./pkg/storage`

15 failures

Tests succeeding, same as after `s/imageName/imageName2/g` in the file.

HEAD as of commit a226a57727a7d88aa6ab9f958317003abd4d86fd .

saschagrunert:
This PR should fix the mentioned issue.

Closes #2117

### DIFF
--- a/pkg/storage/runtime_test.go
+++ b/pkg/storage/runtime_test.go
@@ -32,14 +32,16 @@ var _ = t.Describe("Runtime", func() {
 		gomock.InOrder(
 			imageServerMock.EXPECT().GetStore().Return(storeMock),
 			storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
-			imageServerMock.EXPECT().GetStore().Return(storeMock),
 			storeMock.EXPECT().GraphOptions().Return([]string{}),
 			storeMock.EXPECT().GraphDriverName().Return(""),
 			storeMock.EXPECT().GraphRoot().Return(""),
 			storeMock.EXPECT().RunRoot().Return(""),
 			imageServerMock.EXPECT().GetStore().Return(storeMock),
-			storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
-			storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
+			storeMock.EXPECT().Image(gomock.Any()).
+				Return(&cs.Image{
+					ID:    "123",
+					Names: []string{"imagename"},
+				}, nil).Times(2),
 			storeMock.EXPECT().ImageBigData(gomock.Any(), gomock.Any()).
 				Return(testManifest, nil),
 			storeMock.EXPECT().ListImageBigData(gomock.Any()).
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM assuming happy tests

pkg/storage/runtime_test.go
```diff
@@ -32,14 +32,16 @@ var _ = t.Describe("Runtime", func() {
 		gomock.InOrder(
 			imageServerMock.EXPECT().GetStore().Return(storeMock),
 			storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
```
mtrmac: AFAICS this corresponds to https://github.com/openSUSE/cri-o/blob/e758da45bd7365196e7148425a6405830ebbdfc9/vendor/github.com/containers/image/storage/storage_transport.go#L151 , which tries to interpret the name as an image ID, and that should fail with `(nil, cstorage.ErrImageUnknown)`.

pkg/storage/runtime_test.go
```diff
@@ -32,14 +32,16 @@ var _ = t.Describe("Runtime", func() {
...
 			storeMock.EXPECT().GraphRoot().Return(""),
 			storeMock.EXPECT().RunRoot().Return(""),
 			imageServerMock.EXPECT().GetStore().Return(storeMock),
-			storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
-			storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
+			storeMock.EXPECT().Image(gomock.Any()).
+				Return(&cs.Image{
+					ID:    "123",
+					Names: []string{"imagename"},
+				}, nil).Times(2),
```
mtrmac: (Looks right to me.)

pkg/storage/runtime_test.go
```diff
@@ -574,6 +576,77 @@ var _ = t.Describe("Runtime", func() {
...
+					imageServerMock.EXPECT().GetStore().Return(storeMock),
+					storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
+					imageServerMock.EXPECT().GetStore().Return(storeMock),
+					storeMock.EXPECT().GraphOptions().Return([]string{}),
+					storeMock.EXPECT().GraphDriverName().Return(""),
+					storeMock.EXPECT().GraphRoot().Return(""),
+					storeMock.EXPECT().RunRoot().Return(""),
+					imageServerMock.EXPECT().GetStore().Return(storeMock),
+					storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
+					storeMock.EXPECT().Image(gomock.Any()).Return(&cs.Image{}, nil),
```
mtrmac: This corresponds to https://github.com/openSUSE/cri-o/blob/e758da45bd7365196e7148425a6405830ebbdfc9/pkg/storage/runtime.go#L219 , doesn’t it? The code should never get that far with a camel-case name, and should _not_ succeed AFAICS, because all of
- https://github.com/openSUSE/cri-o/blob/e758da45bd7365196e7148425a6405830ebbdfc9/pkg/storage/runtime.go#L167
- https://github.com/openSUSE/cri-o/blob/e758da45bd7365196e7148425a6405830ebbdfc9/pkg/storage/runtime.go#L170
- https://github.com/openSUSE/cri-o/blob/e758da45bd7365196e7148425a6405830ebbdfc9/pkg/storage/runtime.go#L176

fails with an error.

saschagrunert: Yeah I think the test case does not represent a real world-scenario... I think it would be acceptable to remove it completely, WDYT?

mtrmac: I agree.

saschagrunert: Will be done in https://github.com/kubernetes-sigs/cri-o/pull/2132

PR #2137, opened 2019-03-18
Update lint to use golangci-lint

saschagrunert:
This PR switches from the deprecated gometalinter to golangci-lint, which should provide additional (fast) linting possiblities for the future.

The currently disabled linters could be enabled later on (including their fixes).


mrunalp:
Thanks for the PR :)

PR #2138, opened 2019-03-18
Remove camel-case imageName runtime test

saschagrunert:
As a follow up of #2132 we decided to remove the camel-case test because it does not match any real world scenario. Beside this, a small return value by the mock is fixed.

This PR should decrease the unit test coverage a bit.


haircommander:
LGTM

haircommander:
@rhatdan needs /approve too

mtrmac:
Thanks!

PR #2154, opened 2019-03-20
conmon: Adjust OOM score to low value

mrunalp:
We don't want conmon to be killed.

@giuseppe @rhatdan @umohnani8 ptal.

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -102,6 +102,7 @@ static GOptionEntry opt_entries[] = {
 	{NULL}};
 
 #define CGROUP_ROOT "/sys/fs/cgroup"
+#define OOM_SCORE "-999"
 
 static ssize_t write_all(int fd, const void *buf, size_t count)
 {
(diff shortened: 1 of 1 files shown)


mrunalp:
@giuseppe Do we want to skip this setting oom_score_adj for rootless and/or podman?

rhatdan:
I would figure this will break if you run it as a non privileged user.

rhatdan:
I would figure this requires CAP_RESOURCE to be allowed.

umohnani8:
LGTM

giuseppe:
We could probably just ignore errors or print a warning. It shouldn't be a fatal error if we fail to set this value

mrunalp:
Updated to not fail on errors.

haircommander:
LGTM

haircommander:
Update: this is totally fatal if we are not root, and we may need to get a new version of conmon into libpod. @mrunalp @giuseppe any advise on a fix? I get 
```
failed to write to /proc/self/oom_score_adj: Permission denied
``` for everything i do as rootless

giuseppe:
Does that also stop common?  I will take a look

haircommander:
oh nevermind,  it is just a warning

haircommander:
though, the error message is concerning from a podman perspective. It may be worth ignoring the error all together, unless maybe a user says to force adjust oom score or something?

PR #2160, opened 2019-03-22
conmon: add --exit-command-namespace

giuseppe:
the specified fd's that point to namespaces are joined before calling
the exit command.

This is useful for Podman when --uidmap is used.  We need to do the
cleanup on the parent namespace, while the container (and conmon) are
running in a separate mount namespace.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -22,6 +22,8 @@
 #include <termios.h>
 #include <unistd.h>
 #include <inttypes.h>
+#include <sched.h>
+#include <sys/ioctl.h>
 
 #if __STDC_VERSION__ >= 199901L
 /* C99 or later */
@@ -34,6 +36,12 @@
 
 #include "cmsg.h"
 #include "config.h"
+
+
+#ifndef NS_GET_USERNS
+#define NSIO 0xb7
+#define NS_GET_NSTYPE _IO(NSIO, 0x3)
+#endif
 
 static volatile pid_t container_pid = -1;
 static volatile pid_t create_pid = -1;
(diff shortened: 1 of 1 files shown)


rhatdan:
Won't this be a race condition?  If the processes running the namespace exit before the cleanup program executes, the namespace is gone?

giuseppe:
the namespace is kept alive by the open fd, even if there are no processes in it anymore.

Also, in podman we use this feature to refer to the parent namespace

conmon/conmon.c
```diff
@@ -1397,6 +1408,25 @@ int main(int argc, char *argv[])
...
+			int type, ns;
+			char *endptr = NULL;
+
+			errno = 0;
+			ns = strtol(opt_exit_namespace_args[fd], &endptr, 10);
+			if (errno != 0 || *endptr != '\0')
+				nexitf("unable to parse %s", opt_exit_namespace_args[fd]);
+
+			type = ioctl(ns, NS_GET_NSTYPE);
+			if (unshare(type) < 0)
```
mrunalp: Why are we calling unshare if we want to join the namespace? Shouldn't just setns be enough? In either case, I think adding some comments here will help :)

mrunalp: Should this just be restricted to the mount namespace to keep the scope narrow for the specific problem we are trying to fix?

giuseppe: without the unshare I get `EINVAL` from the setns.  This is the same pattern we are using in Podman.

I can restrict it to the mount namespace, although I preferred to make it more generic in case we will need in future to join some different namespace.

giuseppe:
In any case, let's hold for now, I'd still prefer we don't need this "hack" for dealing with user namespaces.

If we can get https://github.com/containers/libpod/pull/2730, then it will be easier to use/interact with containers not having root mapped into the namespace.

/hold

PR #2163, opened 2019-03-24
[1.13] conmon: Adjust OOM score to low value

mrunalp:
We don't want conmon to be killed


giuseppe:
LGTM

PR #2172, opened 2019-03-26
Update toml decode errors to be more precise

saschagrunert:
Hey, this is just a small change to increase the customer satisfaction and abort with a more meaningful error in case of configuration issues.

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -327,7 +327,7 @@ func (c *Config) UpdateFromFile(path string) error {
 
 	_, err = toml.Decode(string(data), t)
 	if err != nil {
-		return err
+		return fmt.Errorf("unable to decode configuration %v: %v", path, err)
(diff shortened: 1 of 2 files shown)


haircommander:
lgtm!

lib/config.go
```diff
@@ -327,7 +327,7 @@ func (c *Config) UpdateFromFile(path string) error {
 
 	_, err = toml.Decode(string(data), t)
 	if err != nil {
-		return err
+		return fmt.Errorf("unable to decode configuration %v: %v", path, err)
```
giuseppe: could we use `errors.Wrapf`?

saschagrunert: Oh yeah sure we could! Unfortunately it has been merged but I will change it if I come there again. :)

PR #2173, opened 2019-03-26
[1.11] Pick up correct gid

umohnani8:
When setting the gid, we were not checking if RunAsGroup
from k8s was nil or not. We were simply converting that
to an int and nil was becoming 0. Hence all the containers
were being run with gid root regardless of their uid.
Adding that check here fixes it and we get the correct gid.

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -1176,6 +1176,69 @@ function teardown() {
 	stop_crio
 }
 
+@test "ctr with run_as_username set to redis should get 101 as the gid for redis:alpine" {
+	start_crio
+	run crictl runp "$TESTDATA"/sandbox_config.json
+	echo "$output"
+	[ "$status" -eq 0 ]
+	pod_id="$output"
+
+	newconfig=$(cat "$TESTDATA"/container_redis.json | python -c 'import json,sys;obj=json.load(sys.stdin);obj["linux"]["security_context"]["run_as_username"] = "redis"; json.dump(obj, sys.stdout)')
+	echo "$newconfig" > "$TESTDIR"/container_user.json
+
+	run crictl create "$pod_id" "$TESTDIR"/container_user.json "$TESTDATA"/sandbox_config.json
+	echo "$output"
+	[ "$status" -eq 0 ]
+	ctr_id="$output"
+	run crictl start "$ctr_id"
+	[ "$status" -eq 0 ]
+
+	run crictl exec --sync "$ctr_id" id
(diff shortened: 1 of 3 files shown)


umohnani8:
@mrunalp @haircommander @rhatdan PTAL

haircommander:
good find; LGTM, is this something we should be porting to other branches?

umohnani8:
@haircommander no, 1.12 and above have the fix in place already. @mrunalp should we prot to 1.10 and 1.9? They don't have the nil check.

mrunalp:
@umohnani8 Yes, we should backport to 1.9 and 1.10

mrunalp:
The newly added integration test is failing in CI.

mrunalp:
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_crio_integration_fedora/1741/s3/download/gathered/testout.txt

test/ctr.bats
```diff
@@ -1176,6 +1176,39 @@ function teardown() {
 	stop_crio
 }
 
+@test "ctr with run_as_user set to 100 should 101 as the gid for redis:alpine" {
```
mrunalp: nit: should use 101

umohnani8: the uid for redis is 100, but gid is 101.

mrunalp: I meant the text: "ctr with run_as_user set to 100 should use 101 as the gid .." :)

umohnani8: oh sorry, yeah fixed that :)

mrunalp:
@umohnani8 we should port the test to other branches.

umohnani8:
Not sure why it wasn't liking the way I was setting run_as_user in one of the configs and writing to a new config, so I added another json with that specification instead. Hopefully tests will pass now.

umohnani8:
@sboeuf kata is consistently failing :/

sboeuf:
@umohnani8 well this PR introduces 2 new tests that Kata cannot honor, that's why it's failing...
@chavafg can we fix this by adding those to the list of tests to skip?

chavafg:
Sure, I'll open a PR in kata side and let you know when we can restart the kata job

chavafg:
please restart kata job. Should pass now.

mrunalp:
@chavafg Thanks for the quick fix :)

mrunalp:
@umohnani8 We need to cut a new release for 1.11.

PR #2175, opened 2019-03-27
Add RuntimeHandler.RuntimeRoot

saschagrunert:
Hey there, the main intention of this PR is to resolve this TODO:

https://github.com/kubernetes-sigs/cri-o/blob/7a9445829c6165d62a75bcdab1843b264bb01f37/oci/oci_linux.go#L75-L80

My approach would be to introduce a RuntimeRoot for the RuntimeHandler and pass it around where it is necessary. This would also help in unit testing, since we now could use a temporarily run root directory for runc where we are privileged.

WDYT?

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -214,6 +214,7 @@ manage_network_ns_lifecycle = {{ .ManageNetworkNSLifecycle }}
 [crio.runtime.runtimes.{{ $runtime_name }}]
 runtime_path = "{{ $runtime_handler.RuntimePath }}"
 runtime_type = "{{ $runtime_handler.RuntimeType }}"
+runtime_root = "{{ $runtime_handler.RuntimeRoot }}"
 {{ end }}
 
 # The crio.image table contains settings pertaining to the management of OCI images.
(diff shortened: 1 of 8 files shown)


sboeuf:
@saschagrunert @mrunalp I'm not clear on the need to differentiate this root path between `runc` and `kata-runtime` here. The factory can be initialized on a global path chosen by CRI-O, I don't think this needs to be specific to each runtime.
Also, I am not clear why CRI-O is calling into the libcontainer factory since runc should be responsible for this.
I think I'm looking for clarification here :slightly_smiling_face:

saschagrunert:
> @saschagrunert @mrunalp I'm not clear on the need to differentiate this root path between `runc` and `kata-runtime` here. The factory can be initialized on a global path chosen by CRI-O, I don't think this needs to be specific to each runtime.

I'm not sure, since other runtimes (I know there is currently no real use case for that) could use different root directories which may clash with that one of runc. This is only hypothetical, so I would also be fine in just using a global config for all runtimes here. WDYT?

> Also, I am not clear why CRI-O is calling into the libcontainer factory since runc should be responsible for this.
> I think I'm looking for clarification here

I don't think that the current runc API provides what we would need here, but I guess with an additional API we could get rid of this code part completely.

sboeuf:
After clarification from @mrunalp during the weekly CRI-O meeting, I realized the reason for this code and I think having this handled from `RuntimeHandler.RuntimeRoot` sounds good :+1:

umohnani8:
LGTM

haircommander:
LGTM

saschagrunert:

PR #2176, opened 2019-03-27
Enhance config validation

saschagrunert:
This commit updates to config validation on cri-o startup to work for:

- `RuntimeConfig.HooksDir`
- `RuntimeConfig.Conmon`
- `NetworkConfig.NetworkDir`
- `NetworkConfig.PluginDir`

All unit tests have been updated as well, whereas the returned error now
indicates that the issue happened on "config validation".

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -2,7 +2,6 @@ package lib
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -17,6 +16,7 @@ import (
 	cstorage "github.com/containers/storage"
 	units "github.com/docker/go-units"
 	"github.com/kubernetes-sigs/cri-o/oci"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


lib/config.go
```diff
@@ -475,6 +475,37 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
 		if _, err := sysregistriesv2.GetRegistries(nil); err != nil {
 			return fmt.Errorf("invalid /etc/containers/registries.conf: %q", err)
 		}
+
+		for _, hooksDir := range c.HooksDir {
+			if _, err := os.Stat(hooksDir); err != nil {
+				return fmt.Errorf("invalid hooks_dir entry '%s'", hooksDir)
```
vrothberg: Can we wrap `err`? Otherwise, we don't know which kind of error we're having (e.g., no file/dir or a permission one etc.).

lib/config.go
```diff
@@ -475,6 +475,37 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
...
+			}
+
+		}
+
+		if _, err := os.Stat(c.Conmon); err != nil {
+			return fmt.Errorf("invalid conmon path '%s'", c.Conmon)
```
vrothberg: Same here: let's wrap `err`.

lib/config.go
```diff
@@ -475,6 +475,37 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
...
+}
+
+// Validate is the main entry point for network configuration validation.
+// The parameter `onExecution` specifies if the validation should include
+// execution checks. It returns an `error` on validation failure, otherwise
+// `nil`.
+func (c *NetworkConfig) Validate(onExecution bool) error {
+	if onExecution {
+		if _, err := os.Stat(c.NetworkDir); err != nil {
+			return fmt.Errorf("unavailable network_dir '%s'", c.NetworkDir)
```
vrothberg: And here.

lib/config.go
```diff
@@ -475,6 +475,37 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
...
+		}
+
+		for _, pluginDir := range c.PluginDir {
+			if _, err := os.Stat(pluginDir); err != nil {
+				return fmt.Errorf("unavailable plugin_dir entry '%s'", pluginDir)
```
vrothberg: And here.

lib/config.go
saschagrunert: Sure, I adapt it as suggested.

lib/config.go
```diff
@@ -475,6 +475,36 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
...
+}
+
+// Validate is the main entry point for network configuration validation.
+// The parameter `onExecution` specifies if the validation should include
+// execution checks. It returns an `error` on validation failure, otherwise
+// `nil`.
+func (c *NetworkConfig) Validate(onExecution bool) error {
+	if onExecution {
+		if _, err := os.Stat(c.NetworkDir); err != nil {
+			return errors.Wrapf(err, "unavailable network_dir")
```
vrothberg: Can we make the messages consistent with the errors above? I prefer "invalid" as "unavailable" sounds like a service or process.

Apologies for missing that earlier.

saschagrunert: Sure, I changed it.

haircommander:
LGTM

PR #2190, opened 2019-03-28
Log oom_handling_score failure to debug

haircommander:
In testing the newest conmon with podman, the g_print for oom_handling_score failing to open the fd printed every time podman was run as rootless. Fix this by changing it to ndebugf, so a normal user won't encounter the failure.

Also move the handling higher so the log level is set already

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -955,19 +955,6 @@ int main(int argc, char *argv[])
 	int fds[2];
 	int oom_score_fd = -1;
 
-	oom_score_fd = open("/proc/self/oom_score_adj", O_WRONLY);
-	if (oom_score_fd < 0) {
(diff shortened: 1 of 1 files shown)


haircommander:
@giuseppe @mrunalp PTAL 
I'm unsure about moving:
** 	main_loop = g_main_loop_new(NULL, FALSE);**
but I wasn't sure if the main loop should wait to be set until after oom was configured.

giuseppe:
the oom is per process, but I think it is still safer as you did!

I am fine with the change as it is, let's wait for @mrunalp to weight in

TomSweeneyRedHat:
LGTM assuming happy tests

PR #2192, opened 2019-03-29
[1.10] Pick up correct gid

umohnani8:
When setting the gid, we were not checking if RunAsGroup
from k8s was nil or not. We were simply converting that
to an int and nil was becoming 0. Hence all the containers
were being run with gid root regardless of their uid.
Adding that check here fixes it and we get the correct gid.

backport https://github.com/kubernetes-sigs/cri-o/pull/2173

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -468,7 +468,11 @@ func setupContainerUser(specgen *generate.Generator, rootfs string, sc *pb.Linux
 		if sc.GetRunAsUser() == nil && sc.GetRunAsUsername() == "" && sc.GetRunAsGroup() != nil {
 			return fmt.Errorf("RunAsGroup should be specified only with RunAsUser or RunAsUsername")
 		}
-		groupstr := strconv.FormatInt(sc.GetRunAsGroup().GetValue(), 10)
+
+		var groupstr string
+		if sc.GetRunAsGroup() != nil {
+			groupstr = strconv.FormatInt(sc.GetRunAsGroup().GetValue(), 10)
+		}
 		if groupstr != "" {
 			containerUser = containerUser + ":" + groupstr
 		}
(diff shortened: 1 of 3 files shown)


umohnani8:
@mrunalp PTAL

mrunalp:
kata is not working on 1.10

umohnani8:
@sboeuf kata is down, probably need to disable the tests for 1.10 as well?

sboeuf:
@umohnani8 let me pull @chavafg here because I thought the Kata CI was only testing CRI-O master.
@chavafg is it expected to have the CI running on the 1.10 branch?

mrunalp:
It is working fine with master but 1.10 is failing consistently. Even 1.11 and 1.12 are fine. I am okay if we don't want to invest more time in 1.10 :)

chavafg:
yeah, I remember another PR for 1.10 last week which also failed, so it seems that kata is currently not working with 1.10.
I see it was an exec failure:
```
13:58:40 # time="2019-03-29 19:58:16.376764079Z" level=debug msg="ExecSyncRequest &ExecSyncRequest{ContainerId:e0e94e01dd4ad1857e5705c0a509123e21168aca157b2883c1e18ae8683cbc31,Cmd:[sleep 3],Timeout:1,}" 
13:58:40 # time="2019-03-29 19:58:17.590045340Z" level=debug msg="Received container exit code: -1, message: command timed out" 
13:58:40 # time="2019-03-29T19:58:17Z" level=fatal msg="execing command in container synchronously failed: rpc error: code = Unknown desc = command error: command timed out, stdout: , stderr: , exit code -1" 
```

umohnani8:
@mrunalp can we merge this?

mrunalp:
LGTM

PR #2197, opened 2019-04-01
userns: drop intermediate mount namespace

giuseppe:
Simplify the implementation for user namespaces by not using an
intermediate mount namespace.  For doing it, we need to relax the
permissions on the parent directories and allow browsing
them. Containers that are running without a user namespace, will still
maintain mode 0700 on their directory.

**- What I did**
dropped intermediate mount namespace

**- How to verify it**
when using user namespaces, containers mount are accessible from the host/

when using user namespaces, we no longer use an intermediate mount namespace

### DIFF
--- a/lib/sandbox/sandbox.go
+++ b/lib/sandbox/sandbox.go
@@ -337,10 +337,6 @@ func (s *Sandbox) GetContainer(name string) *oci.Container {
 // RemoveContainer deletes a container from the sandbox
 func (s *Sandbox) RemoveContainer(c *oci.Container) {
 	s.containers.Delete(c.Name())
-	intermediateMountPoint := c.IntermediateMountPoint()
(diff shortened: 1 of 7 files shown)


giuseppe:
@mrunalp this should solve the issue we have seen with the shared SHM

mrunalp:
Rebase needed for the tests to pass.

mrunalp:
Rebase to master will fix integration_rhel

giuseppe:
tests are green!

rhatdan:
LGTM

haircommander:
LGTM

haircommander:
it doesn't seem right I have more permissions than you @rhatdan ...

haircommander:
tide is sleeping so I'm taking its place

PR #2203, opened 2019-04-03
test: Increase number of inotify user watches

mrunalp:
Our tests fail without increasing this limit.

@giuseppe ptal

### DIFF
--- a/contrib/test/integration/e2e.yml
+++ b/contrib/test/integration/e2e.yml
@@ -3,6 +3,9 @@
 - name: include base e2e setup
   include: "e2e-base.yml"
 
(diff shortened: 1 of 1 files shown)


giuseppe:
thanks, LGTM

I am worried now if we are leaking inotify watchers fds, either in CRI-O or related to the new patch for the Kubelet.

TomSweeneyRedHat:
LGTM

giuseppe:
> I am worried now if we are leaking inotify watchers fds, either in CRI-O or related to the new patch for the Kubelet.

FYI: I am testing inotify leaks but cannot see anything wrong

PR #2182, opened 2019-03-28
[1.9] Fix possible out of bounds access during log parsing

saschagrunert:
This PR backports #2180

PR #2206, opened 2019-04-03
Add go-md2man to repository

saschagrunert:
Hey, to avoid changing dependencies which could break our builds (like the travis build in #2182) I would go the way to vendor necessary test dependencies directly into the repository.

I did this for go-md2man within this PR and cleaned up the Makefile a bit. If you're fine with that I would do the same for the rest of the external dependencies.


Makefile
```diff
@@ -155,21 +155,35 @@ dbuild: crioimage
...
+	$(call go-build,mockgen,./vendor/github.com/golang/mock/mockgen)
 
 vendor: .install.vndr
 	$(GOPATH)/bin/vndr \
-		-whitelist "github.com/onsi/ginkgo" \
-		-whitelist "github.com/golang/mock" \
+		-whitelist github.com/onsi/ginkgo \
+		-whitelist github.com/golang/mock \
+		-whitelist github.com/cpuguy83/go-md2man \
+		-whitelist github.com/russross/blackfriday \
```
TomSweeneyRedHat: are you using blackfriday with this PR?  Or is this something you need in place for later on?

saschagrunert: Yes we need it as direct dependency of `go-md2man` and have to whitelist it there to avoid removal during running `make vendor`.

TomSweeneyRedHat: OK, maybe we're calling/using go-md2man in Podman and Buildah differently, but we're not pulling in blackfriday in either of those projects that I can see.

TomSweeneyRedHat:
LGTM, assuming happy tests

haircommander:
@saschagrunert rebasing should fix CI issues :)

haircommander:
Tide is not cooperating. Merging manually

PR #2215, opened 2019-04-03
rootless: fix rootless mode

giuseppe:
a couple of issues that prevented to work in rootless mode.  More details in each patch

### DIFF
--- a/oci/runtime_oci.go
+++ b/oci/runtime_oci.go
@@ -124,7 +124,9 @@ func (r *runtimeOCI) CreateContainer(c *Container, cgroupParent string) (err err
 	// 0, 1 and 2 are stdin, stdout and stderr
 	cmd.Env = append(r.conmonEnv, fmt.Sprintf("_OCI_SYNCPIPE=%d", 3))
 	cmd.Env = append(cmd.Env, fmt.Sprintf("_OCI_STARTPIPE=%d", 4))
-	cmd.Env = append(cmd.Env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", os.Getenv("XDG_RUNTIME_DIR")))
+	if v, found := os.LookupEnv("XDG_RUNTIME_DIR"); found {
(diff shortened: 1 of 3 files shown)


oci/runtime_oci.go
```diff
@@ -280,6 +280,7 @@ func (r *runtimeOCI) ExecContainer(c *Container, cmd []string, stdin io.Reader,
 	args = append(args, "--process", processFile.Name())
 	args = append(args, c.ID())
 	execCmd := exec.Command(r.path, args...)
+	execCmd.Env = append(execCmd.Env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", os.Getenv("XDG_RUNTIME_DIR")))
```
rhatdan: What happens if XDG_RUNTIME_DIR is not defined.  Will commands work with it set to 
""?

rhatdan:
Could you try this out without XDG_RUNTIME_DIR being set.

oci/runtime_oci.go
giuseppe: it depends on the program whether they are checking for the variable existence or if it has a value.  Usually it is checked whether it has a value != "".  `os.Getenv` considers them the same, `os.LookupEnv` must be used to check if it is set.

runc when running in a namespace expects this to be set otherwise it will use `/run/runc` which is not accessible to the rootless user.

giuseppe: changed to use `LookupEnv`

rhatdan:
LGTM

rhatdan:
Not sure what is going on with Travis?

haircommander:
@rhatdan I'm willing to bet it has to do with the move

umohnani8:
LGTM

giuseppe:
kubernetes requires go 1.12.  Does it still make sense to use 1.11 and 1.10?  Just thinking if we can simplify the Travis CI somehow

haircommander:
@giuseppe rebasing to master should fix integration_rhel :)

PR #2223, opened 2019-04-05
test: Fix oom test

mrunalp:
We need to bump up the memory limit as the value we are writing isn't enough
with recent runc changes.


saschagrunert:
LGTM

haircommander:
tide is stalling for some reason. We can address that later. manually merging

PR #2225, opened 2019-04-07
Refactor: use idtools.ParseIDMap instead of bundling own version

isimluk:
ParseIDMap function was extracted to idtools in

    https://github.com/containers/storage/pull/236

it is already used in containers/storage and buildah and libpod. Cri-o should
use the very same code as the others.

Signed-off-by: Šimon Lukašík <isimluk@fedoraproject.org>


mrunalp:
@giuseppe ptal.

PR #2237, opened 2019-04-09
Add config interface nil check

saschagrunert:
This commit fixes a nil pointer dereference when creating a new
`ContainerServer` and providing `nil` as `ConfigIface`. Should no
big issue right now, but seems more safe to me.


umohnani8:
LGTM

haircommander:
LGTM

PR #2249, opened 2019-04-11
Re-sync github.com/containers/image

nalind:
Sync the vendored copy of github.com/containers/image to match the version intended in vendor.conf.


mtrmac:
LGTM.

PR #2252, opened 2019-04-13
Fix Makefile targets for sudo

saschagrunert:
This commit fixes the binary build targets when running
`sudo make`, like `sudo make install`.


umohnani8:
LGTM

mrunalp:
@chavafg kata jobs seems to be enqueued for a long time.

chavafg:
Seems like our VM agents got stuck, but it should be fine now. Thanks for reporting.

PR #2260, opened 2019-04-16
Remove unnecessary nolints

saschagrunert:
Since we vendored golangci lint within the repository we can safely
remove source code based nolints which would have no effect.

### DIFF
--- a/lib/wait.go
+++ b/lib/wait.go
@@ -21,12 +21,7 @@ func (c *ContainerServer) ContainerWait(container string) (int32, error) {
 
 	err = wait.PollImmediateInfinite(1,
 		func() (bool, error) {
-			if !isStopped(c, ctr) {
-				return false, nil
-			} else { // nolint
-				return true, nil // nolint
-			} // nolint
-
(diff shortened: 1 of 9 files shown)


lib/wait.go
```diff
@@ -23,10 +23,8 @@ func (c *ContainerServer) ContainerWait(container string) (int32, error) {
 		func() (bool, error) {
 			if !isStopped(c, ctr) {
 				return false, nil
-			} else { // nolint
-				return true, nil // nolint
-			} // nolint
-
+			}
+			return true, nil
```
saschagrunert: @baude I wasn't sure about this one above. Do you know what the main reason for the else construct is there?

haircommander: couldn't one just return isStopped(c, ctr), nil ?

haircommander:
one comment otherwise LGTM

lib/wait.go
saschagrunert: Yeah, that should be possible too. I will change it.

PR #2261, opened 2019-04-16
Revert "Add RuntimeHandler.RuntimeRoot"

giuseppe:
This reverts commit 296d80b6ac61b427583e1cfe06e74da73e7e875b.

I've noticed there are still places where we cannot set the runtime root in the vendored code, so CRI-O breaks unless we set the runtime to the default setting.

I think we should first ensure the vendored libraries allow to change the runtime root.

Tracing the execs:
```
runc             13862  13201    0 /usr/local/bin/runc --root /run/runc start c0d2d88dfd487a8fadaf4c8ce1550641415f910b922f336ee1afba433db7c755                                               
sleep            13841  13835    0 /bin/sleep 9999
runc             13863  13201    0 /usr/local/bin/runc --root /run/runc state c0d2d88dfd487a8fadaf4c8ce1550641415f910b922f336ee1afba433db7c755                                               
runc             13877  13873    0 /usr/local/bin/runc state c0d2d88dfd487a8fadaf4c8ce1550641415f910b922f336ee1afba433db7c755   
```

the last state is not specifying the runtime root

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -214,7 +214,6 @@ manage_network_ns_lifecycle = {{ .ManageNetworkNSLifecycle }}
 [crio.runtime.runtimes.{{ $runtime_name }}]
 runtime_path = "{{ $runtime_handler.RuntimePath }}"
 runtime_type = "{{ $runtime_handler.RuntimeType }}"
-runtime_root = "{{ $runtime_handler.RuntimeRoot }}"
 {{ end }}
 
 # The crio.image table contains settings pertaining to the management of OCI images.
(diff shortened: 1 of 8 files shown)


saschagrunert:
> let me debug more.. hopefully it is something we can fix easily :-)

Thanks so much, feel free to reach me out if I can do anything to support you. :pray:

giuseppe:
ok so the issue seems to be only in the tests suite, where we call directly the RUNTIME, so not worth to revert the patch as the tests can be fixed separately.  Going to close this.  Sorry for the noise.

giuseppe:
ah!  There is one case where it depends from CRI-O, I'll open a separate PR

giuseppe:
PR here: https://github.com/cri-o/cri-o/pull/2262

PR #2263, opened 2019-04-16
[1.13] Add min memory limit check to sandbox_run_linux.go

umohnani8:
The kubelet takes the memory limits of the containers running
in a pod and set is as the memory limit of the pod.
We were enforcing this minimum memory limit check during the
container creation time but not during the sandbox creation
time. Now we take the pod slice path and read the memory file
and double check that it is above the minimum memory limit.
Doing this also returns a correct error message when the memory
limit is too low, instead of the vague "signal killed" message.

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cri-o/cri-o/lib/sandbox"
 	"github.com/cri-o/cri-o/oci"
 	"github.com/cri-o/cri-o/pkg/annotations"
+	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
@@ -30,6 +31,8 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/leaky"
 	"k8s.io/kubernetes/pkg/kubelet/types"
 )
+
+const cgroupMemorySubsystemMountPath = "/sys/fs/cgroup/memory"
 
 func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest) (resp *pb.RunPodSandboxResponse, err error) {
 	const operation = "run_pod_sandbox"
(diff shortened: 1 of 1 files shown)


umohnani8:
@mrunalp @rhatdan @haircommander PTAL

umohnani8:
Will pick into 1.14 and master once this is in.

server/sandbox_run_linux.go
```diff
@@ -385,6 +388,28 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error expanding systemd slice path for %q", cgroupParent)
+	}
+	// read in the memory limit from the memory.limit_in_bytes file
+	fileData, err := ioutil.ReadFile(filepath.Join(cgroupMemoryPath, slicePath, "memory.limit_in_bytes"))
+	if err != nil {
+		return nil, errors.Wrapf(err, "error reading memory.limit_in_bytes file for slice %q", cgroupParent)
+	}
+	// strip off the newline character and convert it to an int
+	strMemory := string(fileData[:len(fileData)-1])
```
haircommander: If fileData is empty, but exists, does ReadFile succeed? if so, this will segfault. super odd case, but could save a headache later

haircommander:
small nit, otherwise LGTM

server/sandbox_run_linux.go
```diff
@@ -385,6 +388,28 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 	}
 	g.AddAnnotation(annotations.CgroupParent, cgroupParent)
 
+	// check memory limit is greater than the minimum memory limit of 4Mb
```
mrunalp: Can we move this up under the `if s.config.CgroupManager == oci.SystemdCgroupsManager {` code block?
This won't work for cgroupfs.

umohnani8: right, my bad. Fixed

server/sandbox_run_linux.go
umohnani8: I am pretty sure it works fine and you end up with an empty byte array, but will look into it further.

haircommander: sorry I wasn't clear, that's what I'm worried about. wouldn't fileData[:len(fileData)-1] try to make an array with fileData[:-1] and then segfault?

server/sandbox_run_linux.go
```diff
@@ -375,6 +378,30 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
...
+			if err != nil {
+				return nil, errors.Wrapf(err, "error expanding systemd slice path for %q", cgroupParent)
+			}
+			// read in the memory limit from the memory.limit_in_bytes file
+			fileData, err := ioutil.ReadFile(filepath.Join(cgroupMemoryPath, slicePath, "memory.limit_in_bytes"))
+			if err != nil {
+				return nil, errors.Wrapf(err, "error reading memory.limit_in_bytes file for slice %q", cgroupParent)
+			}
+			// strip off the newline character and convert it to an int
+			strMemory := string(fileData[:len(fileData)-1])
```
mrunalp: use `string.TrimSpace` here.

server/sandbox_run_linux.go
```diff
@@ -375,6 +378,30 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
...
+			if strMemory != "" {
+				memoryLimit, err := strconv.Atoi(strMemory)
+				if err != nil {
+					return nil, errors.Wrapf(err, "error converting from string to int %q", strMemory)
```
mrunalp: nit: need more context here that we are hitting this while converting cgroups value

server/sandbox_run_linux.go
```diff
@@ -31,6 +32,8 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/types"
 )
 
+const cgroupMemoryPath = "/sys/fs/cgroup/memory"
```
mrunalp: nit: This could be better named as `cgroupMemorySubsystemMountPath`

server/sandbox_run_linux.go
```diff
@@ -375,6 +378,30 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 			}
 			g.SetLinuxCgroupsPath(cgPath + ":" + "crio" + ":" + id)
 			cgroupParent = cgPath
+
+			// check memory limit is greater than the minimum memory limit of 4Mb
+			// expand the cgroup slice path
+			slicePath, err := systemd.ExpandSlice(cgroupParent)
```
giuseppe: how would this work for the cgroupfs backend?

server/sandbox_run_linux.go
umohnani8: fixed.

server/sandbox_run_linux.go
umohnani8: got it, fixed now. Using strings.TrimRight instead

server/sandbox_run_linux.go
umohnani8: Fixed

server/sandbox_run_linux.go
umohnani8: Fixed.

server/sandbox_run_linux.go
umohnani8: It won't, that is why it is executed under the systemd block only.

giuseppe:
LGTM

umohnani8:
Tests are green. @rhatdan @giuseppe @haircommander can one of you give a lgtm please.

PR #2266, opened 2019-04-17
Add min memory limit check to sandbox_run_linux.go

umohnani8:
The kubelet takes the memory limits of the containers running
in a pod and set is as the memory limit of the pod.
We were enforcing this minimum memory limit check during the
container creation time but not during the sandbox creation
time. Now we take the pod slice path and read the memory file
and double check that it is above the minimum memory limit.
Doing this also returns a correct error message when the memory
limit is too low, instead of the vague "signal killed" message.


umohnani8:
cherry picking https://github.com/cri-o/cri-o/pull/2263 into master.
@mrunalp @rhatdan @giuseppe @haircommander PTAL

haircommander:
LGTM

rhatdan:
LGTM

PR #2268, opened 2019-04-18
Add infra container check for pod sandbox

saschagrunert:
On improper API usage it is possible to setup a sandbox without any infra container. This commit adds an additional check for that case.


haircommander:
I am left wondering if this is the only place this assumption is made..
change LGTM

saschagrunert:
> I am left wondering if this is the only place this assumption is made..

Hm, I saw this one as part of the test implementation for the server, not within the master yet. I saw no other occurrences of the same issue but will definitely look forward into it. :innocent:

PR #2269, opened 2019-04-18
Add nil check for image status size

saschagrunert:
Theoretically it is possible to retrieve an image status size of zero, which would lead into a crash on dereferencing the `status.Size` pointer. This PR fixes this issue.


rhatdan:
LGTM

PR #2270, opened 2019-04-18
Add static crio binary build for x86-64 (glibc and musl)

### LINKED ISSUE #2236: crio static build
we should support a static build for CRI-O so that we don't need to carry dependencies and make it easier to copy the crio binary.

More information here: https://github.com/rootless-containers/usernetes/issues/19

saschagrunert:
This commit mainly adds a nix package for having a fully reproducible static build of CRI-O for glibc and musl libc.

- only x86-64 targets are supported for now
- the binaries are stripped
- CI builds runs within a prepared container image on travis
- the binaries are integration tested on travis as well

To test it locally without installed nix but inside a container runtime, simply run:

```
make build-static NIX_IMAGE=<IMAGE>
```

The resulting binaries should now be available within:
- `bin/crio-x86_64-static-glibc`
- `bin/crio-x86_64-static-musl`

To test it within your local nix installation, execute from the root of the git repository:

```
nix-build nix
```

The resulting binary should be now reside in:
- `result-bin/bin/crio-x86_64-static-glibc`
- `result-2-bin/bin/crio-x86_64-static-musl`

Closes #2236

The travis job has to be adapted afterwards to archive the binary and store them as tag assets. We can also decide what else has to be part of the resulting archive, like man pages and the configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ coverprofile
 .gdb_history
 *gomock_reflect*
 lib/state.json
+result-*bin
 
 Vagrantfile
 .vagrant/
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ services:
   - docker
 
 before_install:
-  - if [ "${TRAVIS_OS_NAME}" = linux ]; then make vendor; fi
-  - if [ "${TRAVIS_OS_NAME}" = linux ]; then ./hack/tree_status.sh; fi
   - if [ "${TRAVIS_OS_NAME}" = linux ]; then sudo apt-get -qq update; fi
   - if [ "${TRAVIS_OS_NAME}" = linux ]; then sudo apt-get -qq install btrfs-tools libdevmapper-dev libgpgme11-dev libseccomp-dev; fi
   - if [ "${TRAVIS_OS_NAME}" = linux ]; then sudo apt-get -qq install autoconf automake bison clang-format-3.9 e2fslibs-dev libfuse-dev libtool liblzma-dev gettext libsystemd-dev; fi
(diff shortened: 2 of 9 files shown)


rhatdan:
@giuseppe PTAL 
I think we need to get @mrunalp approval.  I think we want to strip out as much as possible.

giuseppe:
I am seeing this error:

```
$ sudo make build-static NIX_IMAGE=saschagrunert/crionix
podman run -it -v /nix:/crio saschagrunert/crionix sh -c \
	"nix-build crio --argstr revision 4e944052d68fc32ca56637262ea6bed27a3efdd8 && \
	cp result-bin/bin/crio crio"
Error: error checking path "/nix": stat /nix: no such file or directory
make: *** [Makefile:121: build-static] Error 125
```

saschagrunert:
> I am seeing this error:
> 
> ```
> $ sudo make build-static NIX_IMAGE=saschagrunert/crionix
> podman run -it -v /nix:/crio saschagrunert/crionix sh -c \
> 	"nix-build crio --argstr revision 4e944052d68fc32ca56637262ea6bed27a3efdd8 && \
> 	cp result-bin/bin/crio crio"
> Error: error checking path "/nix": stat /nix: no such file or directory
> make: *** [Makefile:121: build-static] Error 125
> ```

Yeah running it as sudo will change the current working directory. I will fix that with the next push. :)

saschagrunert:
Nevertheless, @giuseppe  I currently have no PR revision support baked in (so the git commit has to be available at github.com/cri-o/cri-o). If you still want to try it out until I've fixed the PR and sudo stuff please execute something like (please bear with me for the docker command):

```
> make build-static CONTAINER_RUNTIME=docker NIX_IMAGE=saschagrunert/crionix COMMIT_NO=6b273b1402b74be3378a493dfbecce18dec59094
docker run -it -v /home/sascha/go/src/github.com/cri-o/cri-o/nix:/crio saschagrunert/crionix sh -c \
        "nix-build crio --argstr revision 6b273b1402b74be3378a493dfbecce18dec59094 && \
        cp result-bin/bin/crio crio"
/nix/store/az2wlnfyyr8k444jgi8dq5vgm60zdq53-cri-o-6b273b1402b74be3378a493dfbecce18dec59094-bin
> file nix/crio
nix/crio: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=vA9AMEPM3fGDDdxnNFgG/sqF4WOpyxcJibFdbGl8Z/hoRojRmNy-edXwCeWu2U/SydJEnh4G3xUv5hzZuCj, stripped
```

It would be awesome if you can tell me if the resulting binary fits your needs. :) The main integration into the CI as well as the "package contents" are totally free to discuss here.

rhatdan:
@openshift-ci-robot s/docker/podman/g

giuseppe:
I've tested it locally and it works

LGTM

saschagrunert:
> I've tested it locally and it works

That's great to hear. I will change the build process that it uses the current repository instead of downloading the sources again from GitHub. The issue when running with `sudo` should now be fixed, too.

mrunalp:
> Do we want to run our tests against the binary? If yes, which?

We can add an job for it to run integration or e2e.

> What should be the feature set and contents of the package?
(Is excluding devmapper for now okay?)

That is fine for now.

> Where to put the build artifacts permanently?

I am thinking that if this binary is fairly portable, we can make it part of the releases. However, we need to
make sure it passes tests. Also, one more question I have is whether we can use static glibc instead of musl?

> Where to put the pre-built container image?

We have a cri-o account in quay.io. We can put it there. I will get the creds.

saschagrunert:
> > Do we want to run our tests against the binary? If yes, which?
> 
> We can add an job for it to run integration or e2e.

I added (as a separate commit) another integration test job to travis to run them with the static binary. We may come up with additional travis resources since the integration tests are currently blocking most of the time within the travis pipeline.

> Also, one more question I have is whether we can use static glibc instead of musl?

Yes, this will also reduce the complexity and build time a lot, since mainly all nix packages are pre-built for glibc. On the other side it will make the binary less portable, since parts of glibc will be always rely on the shared object at runtime:

```
# net
/nix/store/0y7jmqnj48ikjh37n3dl9kqw9hnn68nq-binutils-2.31.1/bin/ld: $WORK/b048/_x005.o: in function `_cgo_7e1b3c2abc8d_C2func_getaddrinfo':
/tmp/go-build/cgo-gcc-prolog:57: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
...
```

It is also not possible any more to link gpgme statically, because it tries to link against a function which is also just dynamically available within glibc.

I added the glibc change as a separate commit so it would be easy to revert. Which way do you prefer?

mrunalp:
@saschagrunert Thanks! Maybe we can support both flavors of static binaries with glibc and musl and default to musl? wdyt?

saschagrunert:
> @saschagrunert Thanks! Maybe we can support both flavors of static binaries with glibc and musl and default to musl? wdyt?

Yes that sounds good, I will prepare everything needed for that.

saschagrunert:
I'm done so far and took my freedom to give the [travis stages names for better identification](https://travis-ci.com/cri-o/cri-o/builds/109422385). Integration testing of both binaries works fine. I updated the image on [saschagrunert/crionix](https://cloud.docker.com/u/saschagrunert/repository/docker/saschagrunert/crionix), which now builds both (glibc and musl) binaries on travis.

The binaries generally look fine:
```
> sudo make build-static NIX_IMAGE=saschagrunert/crionix
...
> file bin/*
bin/crio-x86_64-static-glibc: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, for GNU/Linux 2.6.32, Go BuildID=w31mN4vhW4JGkl2JQW6R/X8AWjQV4_QyMt6sJhvvZ/CtgK5AZ9gNx4rEV_9EAT/_9GlIkOEsPkGHXCQvFUT, stripped
bin/crio-x86_64-static-musl:  ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=-HahBPaUHpzhzED6Bkaz/HkFxnTaOwqzcyAVOi4Wc/z6viFIUJMovYo5tjsoAc/AiubSy--8_QsVkOiW2nf, stripped
```

@mrunalp Do we have an S3 account where we can put the artifacts from travis? Another option would be to store them only on tags or completely manually:
- https://docs.travis-ci.com/user/deployment/releases
- https://docs.travis-ci.com/user/uploading-artifacts

mrunalp:
@saschagrunert I was thinking of uploading them when we cut releases - https://github.com/cri-o/cri-o/releases :) Thanks!

mrunalp:
Test is passing but result collection is hanging; giving it another try.

PR #2272, opened 2019-04-18
Fix coverity scan problem

rhatdan:
Coverity is complaining about a leaded file descriptor, since
we never close dummyfd.

This change will close the dummyfd when the program exits.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -848,7 +848,7 @@ static char *setup_attach_socket(void)
 	return attach_symlink_dir_path;
 }
 
-static void setup_terminal_control_fifo()
+static int setup_terminal_control_fifo()
 {
 	_cleanup_free_ char *ctl_fifo_path = g_build_filename(opt_bundle_path, "ctl", NULL);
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe PTAL

conmon/conmon.c
```diff
@@ -869,10 +869,10 @@ static void setup_terminal_control_fifo()
 	int dummyfd = open(ctl_fifo_path, O_WRONLY | O_CLOEXEC);
```
haircommander: why not just set this as _cleanup_close_ ? Does it need to be in the scope of main to work?

rhatdan: I believe so.
I will let Giuseppe judge, I think this has to be open for the entire run of conmon and only closed when it exits, but coverity sees this as a leak.

giuseppe: yes, it needs to be in the scope of main, otherwise it will be closed as soon as we leave `setup_terminal_control_fifo`

rhatdan:
@haircommander @umohnani8 @mrunalp PTAL

PR #2274, opened 2019-04-22
Vendor update for runtime-tools

umohnani8:
Pick up changes made to make setting ENV more efficient

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -68,8 +68,8 @@ require (
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/opencontainers/image-spec v1.0.1
 	github.com/opencontainers/runc v1.0.0-rc7
-	github.com/opencontainers/runtime-spec v0.0.0-20180909173843-eba862dc2470
-	github.com/opencontainers/runtime-tools v0.6.0
+	github.com/opencontainers/runtime-spec v0.1.2-0.20190408193819-a1b50f621a48
+	github.com/opencontainers/runtime-tools v0.3.1-0.20190418135848-095789df6c2b
 	github.com/opencontainers/selinux v1.2.1
 	github.com/openshift/imagebuilder v0.0.0-20190308124740-705fe9255c57 // indirect
 	github.com/opentracing/opentracing-go v0.0.0-20190218023034-25a84ff92183 // indirect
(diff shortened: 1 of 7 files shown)


umohnani8:
@mrunalp @vrothberg PTAL

umohnani8:
@saschagrunert PTAL.

go.mod
```diff
@@ -68,8 +68,8 @@ require (
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/opencontainers/image-spec v1.0.1
 	github.com/opencontainers/runc v1.0.0-rc7
-	github.com/opencontainers/runtime-spec v0.0.0-20180909173843-eba862dc2470
-	github.com/opencontainers/runtime-tools v0.6.0
+	github.com/opencontainers/runtime-spec v0.1.2-0.20190408193819-a1b50f621a48
+	github.com/opencontainers/runtime-tools v0.3.1-0.20190418135848-095789df6c2b
```
rhatdan: Why are you vendoring an older runtime-tools?  Used to have v0.6.0 now you have v0.3.1?

umohnani8: go mod is automatically generating that. I am not sure why but the changes are picked up in the vendor. @vrothberg do you know?

saschagrunert: To discover the latest tag go calls currently

```
> git describe --first-parent --always --abbrev=0 --match "v[0-9]*.[0-9]*.[0-9]*" --tags
```

The argument `--first-parent` avoids collecting tags which are not done directly at the master branch.

vrothberg: > go mod is automatically generating that. I am not sure why but the changes are picked up in the vendor. @vrothberg do you know?

To me, it seems that runtimes-tools has been pinned to a commit as `go mod` does not downgrade unless explicitly instructed to do so.

A `go get -u github.com/opencontainers/runtime-tools` updates to the latest version (i.e., `v0.9.0`). Same for the spec for getting `v1.0.1`.

umohnani8: @vrothberg I need to pick up the most recent commit in master. Setting it to the latest version would pick up v0.9.0, which would not have the fixes we want to vendor.

vrothberg: In that case, everything seems fine. @saschagrunert's comments explain why go-mod chooses the other prefix that we can safely ignore as the commit at the end is the important part.

saschagrunert: I decided to follow up on this and proposed a change in golang (https://github.com/golang/go/issues/31673), let's see if it can be fixed.

Issue #2178, opened 2019-03-27
Add support for injecting init


haircommander:
If a container is not in the pod's pid namespace, the processes can get orphaned and signals can get missed. This is a tracker issue for allowing the user to inject an init binary (like: https://github.com/Yelp/dumb-init or https://github.com/krallin/tini) into any container that doesn't share the pod's pid namespace.

giuseppe:
Podman by default we use catatonit when `--init` is specified

mrunalp:
One requirement that we have here is making sure that the logs are streamed correctly so we get the logs from the container process spawned by the init. We want to avoid making changes to conmon. From what we discussed in last week's CRI-O meeting, https://github.com/Yelp/dumb-init supports that. Does `catatonit` also support that?

giuseppe:
I guess so, it doesn't affect the std streams.  I forgot the reasons why we picked catatoinit and not considered dumb-init.  @vrothberg do you remember the details? :-)

vrothberg:
[catatonit](https://github.com/openSUSE/catatonit) does a better job at signal handling than others. @cyphar is the author of catatonit and may add some details.

vrothberg:
BTW, the init-logic from libpod can be found here: https://github.com/containers/libpod/blob/e7a2eecf5f3975edfb92cd2cacff0d34ef45f808/pkg/spec/createconfig.go#L153

haircommander:
I am pretty sure runc now creates a pid 1 in a private pid namespace for a container. As such, I'm closing this

PaulFurtado:
@haircommander do you have a pointer to where runc started doing that? I'm testing the static builds for cri-o 1.21 and neither runc nor crun appear do that.

cri-o is pretty unusable without that feature. That said, we do wrap the runc binary on our servers in order to inject additional settings into config.json, so we can easily workaround this, I'm just really surprised that this isn't a solved problem yet, so I'm wondering if I'm missing some setting somewhere or something. Thanks!

haircommander:
so I am pretty sure I was wrong about that, and I can't remember why I was under that impression. That said, I don't think CRI-O can unconditionally inject an init in a private pid namespace--there are cases like a systemd container that would be broken if that was done. We could entertain having an allowed_annotation that allows a user to opt-into that behavior for certain pods, but that wouldn't be much different than a user specifying a different entrypoint (and adding the init binary into the image). In other words: I don't think we can do it automatically, so the current manual steps may be the best route

[...]

PR #2275, opened 2019-04-22
[master] Add option to inject init binary

haircommander:
When a container is run with a private PID namespace, we want the user to be able to define an init binary and run their containers with it.
To do this:
	Add --init and --init-path as options in cri-o CLI and config files. The defaults are false and /usr/libexec/crio/catatonit
	Add a mount '/dev/init' that ro bind mounts the init binary the user specified if we are in a private PIDNS, and prepend '/dev/init --' to the process args
	Add bats test to verify functionality
	Add a install_catatonit.sh that pulls catatonit binary for testing with with bats.

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -166,6 +166,12 @@ default_mounts = [
 # Maximum number of processes allowed in a container.
 pids_limit = {{ .PidsLimit }}
 
+# Path to init binary to run inside containers that forwards signals and reaps processes
+# if the container has a private pid namespace
+# If the path is empty, init will not be used.
+# Note this binary will be bind mounted in /sbin/crioinit and run followed by '--', the standard for container init processes
+init = "{{ .Init }}"
+
 # Maximum sized allowed for the container log file. Negative numbers indicate
 # that no size limit is imposed. If it is positive, it must be >= 8192 to
 # match/exceed conmon's read buffer. The file is truncated and re-opened so the
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -163,6 +163,9 @@ func mergeConfig(config *server.Config, ctx *cli.Context) error {
 	if ctx.GlobalIsSet("pids-limit") {
 		config.PidsLimit = ctx.GlobalInt64("pids-limit")
 	}
+	if ctx.GlobalIsSet("init") {
+		config.Init = ctx.GlobalString("init")
+	}
 	if ctx.GlobalIsSet("log-size-max") {
 		config.LogSizeMax = ctx.GlobalInt64("log-size-max")
 	}
@@ -395,6 +398,10 @@ func main() {
 			Name:  "pids-limit",
 			Value: lib.DefaultPidsLimit,
 			Usage: "maximum number of processes allowed in a container",
+		},
+		cli.StringFlag{
+			Name:  "init",
+			Usage: fmt.Sprintf("Path to an init binary that forwards signals and reaps processes inside containers if the container has a private pid namespace (default: %s)", defConf.Init),
 		},
 		cli.Int64Flag{
 			Name:  "log-size-max",
--- a/lib/container_server.go
+++ b/lib/container_server.go
@@ -139,7 +139,7 @@ func New(ctx context.Context, configIface ConfigIface) (*ContainerServer, error)
 
 	storageRuntimeService := storage.GetRuntimeService(ctx, imageService, config.PauseImage, config.PauseImageAuthFile)
 
-	runtime, err := oci.New(config.DefaultRuntime, config.Runtimes, config.Conmon, config.ConmonEnv, config.CgroupManager, config.ContainerExitsDir, config.ContainerAttachSocketDir, config.LogSizeMax, config.LogToJournald, config.NoPivot, config.CtrStopTimeout)
+	runtime, err := oci.New(config.DefaultRuntime, config.Runtimes, config.Conmon, config.ConmonEnv, config.CgroupManager, config.ContainerExitsDir, config.ContainerAttachSocketDir, config.LogSizeMax, config.LogToJournald, config.NoPivot, config.CtrStopTimeout, config.Init)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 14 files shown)


haircommander:
Fixes: #2178

haircommander:
WIP because there are some cases we don't set process args but may want to set the mount. I'll need some direction there. They're labeled // TODO FIXME, and only exist in the container server restore case

server/container_create_linux.go
```diff
@@ -33,9 +33,14 @@ import (
...
-// minMemoryLimit is the minimum memory that must be set for a container.
-// A lower value would result in the container failing to start.
-const minMemoryLimit = 4194304
+const (
+	// minMemoryLimit is the minimum memory that must be set for a container.
+	// A lower value would result in the container failing to start.
+	minMemoryLimit = 4194304
+	// defaultInitLocation describes where an init binary should be
+	// mounted into the container
+	defaultInitLocation = "/dev/init"
```
mrunalp: I think /var/ may be more suitable. cc: @rhatdan

test/helpers.bash
```diff
@@ -54,6 +54,8 @@ COPYIMG_BINARY=${COPYIMG_BINARY:-${CRIO_ROOT}/cri-o/test/copyimg/copyimg}
 ARTIFACTS_PATH=${ARTIFACTS_PATH:-${CRIO_ROOT}/cri-o/.artifacts}
 # Path of the checkseccomp binary.
 CHECKSECCOMP_BINARY=${CHECKSECCOMP_BINARY:-${CRIO_ROOT}/cri-o/test/checkseccomp/checkseccomp}
+# Path of the bin2img binary.
```
mrunalp: We should call it INIT_BINARY.

server/container_create_linux.go
mrunalp: Re-reading the FHS, `/sbin/init` is the better location for this.

haircommander: hm, I was conforming with podman on this one. Also, for the majority of the mounts we use we mount in /dev, so isn't it idiomatic to do that here as well?

cmd/crio/config.go
```diff
@@ -166,6 +166,14 @@ default_mounts = [
 # Maximum number of processes allowed in a container.
 pids_limit = {{ .PidsLimit }}
 
+# Run an init inside containers that forwards signals and reaps processes
```
mrunalp: Can we condense these to just one setting: `init`? If empty string, then it does nothing, else use the binary.

server/container_create_linux.go
mrunalp: Seems wonky to me to mount a binary under /dev. Let's hear what @rhatdan thinks.

rhatdan: Yes /dev/init does seem strange but I believe the location was created based on what Docker was doing.

haircommander: so what's our consensus?

mrunalp: I vote something under `/sbin` but not overwrite `init` so maybe `/sbin/crioinit`.

haircommander: SGTM, @rhatdan ?

cmd/crio/config.go
haircommander: yeah I like that better

rhatdan:
Fine with me, but you would need to create the subdir.  Another option would be /crioinit.
In either of these cases you need to guarantee that their is an Inode to mount on top of, which I guess runc takes care of.

mrunalp:
runc should take care of creating the missing directories for a bind mount.

haircommander:
These issues are legitimate failures. Gotta fix something...

haircommander:
comments addressed and happy tests. PTAL when you get a chance @saschagrunert @mrunalp @umohnani8 @rhatdan

cmd/crio/main.go
```diff
@@ -396,6 +399,10 @@ func main() {
 			Value: lib.DefaultPidsLimit,
 			Usage: "maximum number of processes allowed in a container",
 		},
+		cli.StringFlag{
+			Name:  "init",
+			Usage: fmt.Sprintf("Path to an init binary that forwards signals and reaps processes inside containers if the container has a private pid namespace (default: %s)", defConf.Init),
```
saschagrunert: Please add this documentation to the man pages in `docs/*.md` too.

lib/container_server.go
```diff
@@ -451,6 +451,7 @@ func (c *ContainerServer) LoadSandbox(id string) error {
 		return err
 	}
 
+	// TODO FIXME We currently don't set anything up for the init binary here.
```
saschagrunert: Do you have a suggestion what needs to be done here? I would suggest to open a follow-up issue instead of leaving `TODO`s in the source code. :innocent:

server/container_create_linux.go
```diff
@@ -33,9 +33,14 @@ import (
...
-// minMemoryLimit is the minimum memory that must be set for a container.
-// A lower value would result in the container failing to start.
-const minMemoryLimit = 4194304
+const (
+	// minMemoryLimit is the minimum memory that must be set for a container.
+	// A lower value would result in the container failing to start.
+	minMemoryLimit = 4194304
+	// defaultInitLocation describes where an init binary should be
+	// mounted into the container
+	defaultInitLocation = "/sbin/crioinit"
```
saschagrunert: Do you think it would be good to put this into the lib config?

test/catatonit/install_catatonit.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+BASE_PATH="$1"
+CATATONIT_PATH="${BASE_PATH}/catatonit"
+CATATONIT_VERSION="v0.1.3"
```
saschagrunert:
```suggestion
CATATONIT_VERSION="v0.1.4"
```

rhatdan:
LGTM
But I have a couple of questions.  You have two TODOs, that talk about configuring init, what needs to be configured and what do you want to do?

rhatdan:
@haircommander Rebase?

cgwalters:
I don't think this can compatibly be enabled for all containers.  People might be using a custom pid1 in their container today.

Rather this should be driven by kube pod spec, something as simple as `init: true` right?  It's surprising there isn't a kube enhancement proposal for this yet.

What's really tempting is to patch `bash` to detect if it's being run in kube and automatically run as init...

PR #2277, opened 2019-04-23
Remove gofmt Makefile target

saschagrunert:
golangci-lint already lints with gofmt so we should avoid checking twice.


TomSweeneyRedHat:
LGTM
But the tests aren't happy.  I can't imagine it's related to these changes.

haircommander:
@vrothberg it seems kata is down; every other PR we have in flight is also failing. @chavafg PTAL

chavafg:
can someone please restart the kata job?

PR #2278, opened 2019-04-23
Remove ostree dependency

saschagrunert:
Since ostree is not needed any more we can remove the installation from our CI to gain a general speedup. Beside this, the script `hack/ostree_tag.sh` does not seem to be needed any more.


rhatdan:
LGTM

PR #2279, opened 2019-04-23
Fix testunit-bin Makefile target

saschagrunert:
The `PACKAGES` variable has been removed from the Makefile which breaks the `testunit-bin` target, which should be fixed now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,7 @@ testunit: mockgen ${GINKGO}
 
 testunit-bin:
 	mkdir -p ${TESTBIN_PATH}
-	for PACKAGE in ${PACKAGES}; do \
+	for PACKAGE in `$(GO) list ./...`; do \
 		go test $$PACKAGE \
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

umohnani8:
@sboeuf @chavafg kata seems to be down. It is failing on multiple PRs, please take a look.

chavafg:
Hi, yes, kata is broken using devicemapper. People already looking at the issue, but not sure if we will have a fix today. 
@sboeuf options that come to my mind to workaround the issue while it gets fixed are to:
1. merge the revert I opened at https://github.com/kata-containers/runtime/pull/1577 or
2. change the storage driver to overlay.

sboeuf:
@chavafg I'd say let's change the storage to overlay, only as a temporary measure. I hope @bergwolf can fix this soon, so that we can re-enable devicemapper testing here.

PR #2280, opened 2019-04-23
crio: set correct exit code on signal

giuseppe:
if we exit because of a signal, set the exit code to 128+signal.

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
 	"time"
 
 	_ "github.com/containers/libpod/pkg/hooks/0.1.0"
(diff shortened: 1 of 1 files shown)


cmd/crio/main.go
```diff
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
```
wking: ew, dead-end dependency ;).  Can we use `golang.org/x/sys/unix` behind the appropriate build tags?  I dunno what Windows expects for this (if CRI-O even builds for Windows?).

giuseppe: I can try that, I assumed it was safe to use "syscall" as we are already using it in few different places  (and I am not sure how it is supposed to work on Windows)

PR #2282, opened 2019-04-24
Add openpgp_tag.sh as fallback if no gpgme is available

saschagrunert:
Hey, I'm not sure if we want to support this but we can build CRI-O without `gpgme-devel`, `libassuan-devel` and `libgpg-error-devel`.


rhatdan:
LGTM

haircommander:
LGTM

PR #2283, opened 2019-04-24
Remove `go build -i` flag

saschagrunert:
The repository should be completely self containing so that the `-i` flag is not needed any more.


haircommander:
LGTM

PR #2284, opened 2019-04-24
Update PluginDir to be created if not existing

saschagrunert:
Kubernetes DeamonSets should be able to add CNI plugins after the start
of CRI-O. But currently, CRI-O fails to start if the PluginDir is not
available. This is now solved by simply creating the directory on
configuration validation.

Furthermore, it will be now validated if the specified PluginDirs are
real directories (and not a files for example).

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -507,7 +507,7 @@ func (c *NetworkConfig) Validate(onExecution bool) error {
 		}
 
 		for _, pluginDir := range c.PluginDir {
-			if _, err := os.Stat(pluginDir); err != nil {
+			if err := os.MkdirAll(pluginDir, 0755); err != nil {
 				return errors.Wrapf(err, "invalid plugin_dir entry")
 			}
 		}
--- a/server/config_test.go
+++ b/server/config_test.go
@@ -3,6 +3,7 @@ package server
 import (
 	"io/ioutil"
 	"os"
+	"path"
 	"testing"
 
 	"github.com/cri-o/cri-o/lib"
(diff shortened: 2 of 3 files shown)


TomSweeneyRedHat:
LGTM assuming happy tests

server/config_test.go
```diff
@@ -113,7 +113,8 @@ func TestConfigValidateDefaultSuccessOnExecution(t *testing.T) {
 	defaultConfig.Runtimes["runc"] = oci.RuntimeHandler{RuntimePath: validPath}
 	defaultConfig.Conmon = validPath
 	defaultConfig.NetworkConfig.NetworkDir = validPath
-	defaultConfig.NetworkConfig.PluginDir = []string{validPath}
+	defaultConfig.NetworkConfig.PluginDir = []string{"cni-test"}
```
vrothberg: Can we make this a temp directory? `cni-test` looks very generic, so there is a chance we delete local data.

saschagrunert: Yep sure, I will change it.

lib/config.go
```diff
@@ -507,7 +507,7 @@ func (c *NetworkConfig) Validate(onExecution bool) error {
 		}
 
 		for _, pluginDir := range c.PluginDir {
-			if _, err := os.Stat(pluginDir); err != nil {
+			if err := os.MkdirAll(pluginDir, 0755); err != nil {
```
umohnani8: should we only create it if the pluginDir doesn't exist?

umohnani8:
One question, otherwise LGTM

lib/config.go
saschagrunert: I would keep it: If `pluginDir` is already a directory, `MkdirAll()` does nothing and returns `nil`. If it’s a file, it errors which we had not before. :)

PR #2285, opened 2019-04-24
[1.13] Fix logic of server.restore()

umohnani8:
When cri-o is restarted, we were not returning an error if
an error ocurred when reading container state from disk. So
if the state.json file was corrupted or deleted in the meantime
the containers would be reporting an UNKOWN state and default
values for started time, created time, etc.
We now check for that and if a pod can't be restored, we delete the pod
as well as any containers associated with it. We also release the pod
and container names so that they can be used again.
We do the same for any containers that can't be restored.

Should fix https://bugzilla.redhat.com/show_bug.cgi?id=1702655

### DIFF
--- a/lib/container_server.go
+++ b/lib/container_server.go
@@ -451,7 +451,9 @@ func (c *ContainerServer) LoadSandbox(id string) error {
 		}
 	}
 
-	c.ContainerStateFromDisk(scontainer)
+	if err := c.ContainerStateFromDisk(scontainer); err != nil {
+		return fmt.Errorf("error reading sandbox state from disk %q: %v", scontainer.ID(), err)
+	}
 	sb.SetCreated()
 
 	if err = label.ReserveLabel(processLabel); err != nil {
@@ -566,7 +568,9 @@ func (c *ContainerServer) LoadContainer(id string) error {
 	spp := m.Annotations[annotations.SeccompProfilePath]
 	ctr.SetSeccompProfilePath(spp)
 
-	c.ContainerStateFromDisk(ctr)
+	if err := c.ContainerStateFromDisk(ctr); err != nil {
+		return fmt.Errorf("error reading container state from disk %q: %v", ctr.ID(), err)
+	}
 	ctr.SetCreated()
 
 	c.AddContainer(ctr)
@@ -583,9 +587,9 @@ func (c *ContainerServer) ContainerStateFromDisk(ctr *oci.Container) error {
 	if err := ctr.FromDisk(); err != nil {
 		return err
 	}
-	// ignore errors, this is a best effort to have up-to-date info about
-	// a given container before its state gets stored
-	c.runtime.UpdateStatus(ctr)
+	if err := c.runtime.UpdateStatus(ctr); err != nil {
+		return err
+	}
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


umohnani8:
@mrunalp @rhatdan @haircommander PTAL

mrunalp:
Looks fine to me. @giuseppe ptal.

lib/container_server.go
```diff
@@ -585,7 +589,9 @@ func (c *ContainerServer) ContainerStateFromDisk(ctr *oci.Container) error {
 	}
 	// ignore errors, this is a best effort to have up-to-date info about
```
haircommander: Comment is out of date

lib/container_server.go
```diff
@@ -595,7 +601,9 @@ func (c *ContainerServer) ContainerStateFromDisk(ctr *oci.Container) error {
 func (c *ContainerServer) ContainerStateToDisk(ctr *oci.Container) error {
 	// ignore errors, this is a best effort to have up-to-date info about
 	// a given container before its state gets stored
-	c.Runtime().UpdateStatus(ctr)
+	if err := c.Runtime().UpdateStatus(ctr); err != nil {
```
haircommander: Same here

server/server.go
```diff
@@ -167,24 +168,57 @@ func (s *Server) restore() {
...
 	}
+
+	// Go through all the containers and check if it can be restored. If an error occurs, delete the conainer and
+	// release the name associated with you.
 	for containerID := range podContainers {
 		if err := s.LoadContainer(containerID); err != nil {
 			logrus.Warnf("could not restore container %s: %v", containerID, err)
 			for _, n := range names[containerID] {
 				s.Store().DeleteContainer(n)
+				// Release the container name
```
haircommander: Should this also be an error for the pod or are we okay with partially restored pods?

haircommander:
Good find couple of nits then LGTM

server/server.go
umohnani8: This is deleting the pod completely, we won't end up with partially restored pods.

lib/container_server.go
umohnani8: fixed! thanks :)

lib/container_server.go
umohnani8: fixed!

mrunalp:
Let us test this on a cluster by restarting a node a few times and making sure everything is fine.

saschagrunert:
Might the integration test issue be related to this change? I dont't think so, but It is always test case 86 which fails: https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/2285/test_pull_request_crio_integration_fedora/2465/artifacts/testout.txt

/test integration_fedora

saschagrunert:
@umohnani8 Can you please rebase on top of the latest master?

umohnani8:
@saschagrunert this is to 1.13, can't rebase on top of master without pulling in a lot of things. Is there a fix there that I can cherry pick?

umohnani8:
Is it this? https://github.com/cri-o/cri-o/commit/9c4c00622b0aa1fd2677b789b42ede837f18a21f
Because the error says it can't find the cni plugin

saschagrunert:
> @saschagrunert this is to 1.13, can't rebase on top of master without pulling in a lot of things. Is there a fix there that I can cherry pick?

Ah my fault, I did not see this.


> Is it this? [9c4c006](https://github.com/cri-o/cri-o/commit/9c4c00622b0aa1fd2677b789b42ede837f18a21f)
> Because the error says it can't find the cni plugin

I don't think so, this patch has gone into the master.

I'm not sure what is going on here but either the integration tests are broken for the 1.13 release branch or it is related to this change. I will have a closer look...

umohnani8:
@saschagrunert yeah from the test output, it is failing to add a cni plugin. I can't seem to find why that is happening. The rhel integrations tests pass fine, just an issue with fedora.

saschagrunert:
> @saschagrunert yeah from the test output, it is failing to add a cni plugin. I can't seem to find why that is happening. The rhel integrations tests pass fine, just an issue with fedora.

Yep, but it's not related to this PR. I will check if we can fix this in #2289

saschagrunert:
#2289 should fix the integration test run.

umohnani8:
thanks a bunch @saschagrunert !

umohnani8:
rebased

umohnani8:
Did some more testing with rebooting the node and all, and the pods were either restored or destroyed and recreated when an issue occurred properly. This should be ready to merge.
/hold cancel

umohnani8:
@mrunalp needd approve and lgtm labels please

PR #2286, opened 2019-04-25
Add unit test inject files

saschagrunert:
This commit introduces a new build tag `test` to separate the testing API from the user API. All test targets have been adapted to use this tag, whereas the main binaries are excluded. This make some mentions that "this API function should not be used" obsolete and provides us more flexibility for future mocking approaches.

All methods I added here will be used by unit tests later on.


rhatdan:
@mrunalp @umohnani8 @cevich PTAL

haircommander:
LGTM

PR #2289, opened 2019-04-25
[1.13] Fix integration tests for missing cni_plugin_helper.bash

saschagrunert:
This is a trivial test if the release-1.13 pipeline runs as intended whether to check if we need to apply a fix.

Relates to: #2285

---

Edit: Found the problem, applied a quick fix and changed the commit message:

```
Some test machines do not have the `cni_plugin_helper.bash` installed
when running the integration tests. This commit fixes it with a manual
installation before running the actual suite.
```

### DIFF
--- a/test/test_runner.sh
+++ b/test/test_runner.sh
@@ -30,5 +30,8 @@ function execute() {
 # Tests to run. Defaults to all.
 TESTS=${@:-.}
 
+# Copy the cni helper
+cp cni_plugin_helper.bash /opt/cni/bin
+
 # Run the tests.
 execute time bats --tap $TESTS


saschagrunert:
I would assume that the removal of `cni_plugin_helper.bash` from the master branch has influence on the configuration of the test machine. Could it be related to some asynchronously executed ansible states? I will go for a manual installation of `/opt/cni/bin/cni_plugin_helper.bash` in `test_runner.sh`. Other branches might be affected as well.

mrunalp:
Maybe an ami job ran which picks up changes from setup.

PR #2290, opened 2019-04-25
Add server config interface

saschagrunert:
This server configuration interface is the equivalent implementation
to the library configuration interface, except that it provides access
to the internal config via `GetLibConfigIface()`. This change is needed
especially for test abstraction and mocking purposes. The necessary
mocks are included in this change as well.


TomSweeneyRedHat:
LGTM assuming happy tests.

PR #2002, opened 2019-01-04
WIP: Add server unit tests

saschagrunert:
This PR adds unit tests to the `server` package.

PR #2293, opened 2019-04-26
Add server test suite and initial cases

saschagrunert:
This PR is a spin-off from #2002 to split it up into more review-able pieces. It mainly does:

- Add server test suite
- Add the `server/server.go` tests
- Add ginkgo table extension to vendor folder
- Add new mocked interfaces
- Add certificate test data
- Update streamService to be public

I know that "making things public for testing" is not a best practice, but this should really not harm, because the only way to access it is hidden via `server/server_test_inject.go`.


server/server_test.go
```diff
@@ -0,0 +1,339 @@
...
+		})
+	})
+
+	t.Describe("StopStreamServer", func() {
+		// Prepare the sut
+		BeforeEach(setupSUT)
+
+		It("should succeed", func() {
+			// Given
+			/* gomock.InOrder(
```
mrunalp: Why is the commented out? Can we remove it if it is not needed?

saschagrunert: Yes, I will remove it.

PR #2300, opened 2019-04-29
Fix hack/openpgp_tag.sh on older distributions

saschagrunert:
Some distributions do not ship the `pkg-config` files for gpgme. This
commit uses the `gpgme-config` tool to check for its availability, which
should work for every distribution.


umohnani8:
LGTM

PR #2306, opened 2019-04-29
Update runc test dependency to v1.0.0-rc8

saschagrunert:
Hey, if there are no concerns I would like to just use the latest static runc binary from GitHub for integration testing.


saschagrunert:
I would like to leave the ami updates out of scope for now. We have to follow up on this with the help of someone who has direct access to the machines.

rhatdan:
LGTM
@umohnani8 @haircommander @mrunalp @giuseppe PTAL

PR #2309, opened 2019-04-30
Vendor Kubernetes v1.15.0-alpha.2

saschagrunert:
This PR vendors the latest Kubernetes alpha release to be tested against our pipeline. The whole vendoring part changed a lots in Kubernetes so I guess it would be good if we adopt early to avoid release issues with 1.15.0.


go.mod
```diff
@@ -98,16 +92,36 @@ require (
...
+	k8s.io/kubernetes v0.0.0
+	k8s.io/utils v0.0.0-20190221042446-c2654d5206da
 )
 
-replace cloud.google.com/go => cloud.google.com/go v0.34.0
-
-module github.com/cri-o/cri-o
+replace (
+	cloud.google.com/go => cloud.google.com/go v0.34.0
+	github.com/opencontainers/runtime-spec => github.com/opencontainers/runtime-spec v0.1.2-0.20190408193819-a1b50f621a48
```
vrothberg: I am not sure those need to be replaced. Can't we keep them in the require section only?

saschagrunert:
> Thanks a lot for the change. I am quite shocked what K8s upstream is doing with the ./staging tree. It's not obvious to me why they chose this path as it seems to partially defeat the purpose of go mod.

I thought the same and am currently trying to reach out the responsible upstream persons to work on an improvement. I guess until this discussion is not done we should leave this in WIP. I mean I get the idea of having a clean appraoch (also in relation to the [hack/update-vendor.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/update-vendor.sh) script), but it should be easier to depend in k8s.io/*.

go.mod
saschagrunert: Unfortunaetly not, because k8s directly uses v1.0.0 which is too old for our use case: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L121

saschagrunert:
I had a chat with Jordan Liggit about getting rid of the `k8s.io/kubernetes` import and only rely on the staging repositories (like `k8s.io/cri-api`). Unfortunately this does not work in our case because we rely on imports like `kubernetes.io/kubernetes/pkg/kubelet/types`.

saschagrunert:
I would like to stick to our current approach, PTAL.

PR #2311, opened 2019-04-30
[1.13] Vendor containers/storage

umohnani8:
Pick up changes to overlay for recreating symlinks.


umohnani8:
@mrunalp @rhatdan PTAL

haircommander:
is this an official vendor or did you just change the code? will this mess with `make vendor` calls later? @umohnani8

umohnani8:
@haircommander it is an official vendor. We don't have make vendor in 1.13, so I used the normal vndr tool.

haircommander:
hm wondering why vendor.conf isn't changed too, but if it ends up being a problem we can fix later

/lgtm

umohnani8:
@haircommander because I updated the branch we already had in vendor.conf, so just had to run vndr to pull in the changes

PR #2316, opened 2019-05-02
Add server image unit tests

saschagrunert:
This PR adds all image related unit tests to the server implementation, whereas the coverage should be 100%.


umohnani8:
LGTM

PR #2318, opened 2019-05-02
[1.14] Fix logic of server.restore()

umohnani8:
When cri-o is restarted, we were not returning an error if
an error ocurred when reading container state from disk. So
if the state.json file was corrupted or deleted in the meantime
the containers would be reporting an UNKOWN state and default
values for started time, created time, etc.
We now check for that and if a pod can't be restored, we delete the pod
as well as any containers associated with it. We also release the pod
and container names so that they can be used again.
We do the same for any containers that can't be restored.

### DIFF
--- a/lib/container_server.go
+++ b/lib/container_server.go
@@ -467,7 +467,9 @@ func (c *ContainerServer) LoadSandbox(id string) error {
 		}
 	}
 
-	c.ContainerStateFromDisk(scontainer)
+	if err := c.ContainerStateFromDisk(scontainer); err != nil {
+		return fmt.Errorf("error reading sandbox state from disk %q: %v", scontainer.ID(), err)
+	}
 	sb.SetCreated()
 
 	if err = label.ReserveLabel(processLabel); err != nil {
@@ -582,7 +584,9 @@ func (c *ContainerServer) LoadContainer(id string) error {
 	spp := m.Annotations[annotations.SeccompProfilePath]
 	ctr.SetSeccompProfilePath(spp)
 
-	c.ContainerStateFromDisk(ctr)
+	if err := c.ContainerStateFromDisk(ctr); err != nil {
+		return fmt.Errorf("error reading container state from disk %q: %v", ctr.ID(), err)
+	}
 	ctr.SetCreated()
 
 	c.AddContainer(ctr)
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

lib/container_server.go
```diff
@@ -467,7 +467,9 @@ func (c *ContainerServer) LoadSandbox(id string) error {
 		}
 	}
 
-	c.ContainerStateFromDisk(scontainer)
+	if err := c.ContainerStateFromDisk(scontainer); err != nil {
```
saschagrunert: These changes have direct influence on the current unit tests, which will now fail because the `state.json` is not available. :-/

umohnani8: Is there a way to add state.json files for every sandbox and container created in the test framework?

saschagrunert: Yes, this should work. We could create a new test function in `lib/suite_test` like this:

```go
func createDummyState() {
	ioutil.WriteFile("state.json", []byte("{}"), 0644)
}
```

Then we could use this function for all seven failing tests to succeed. More should not be needed, since every test already cleans up via `removeState()`.

server/server.go
```diff
@@ -149,38 +149,72 @@ func (s *Server) restore() {
...
+			s.Store().DeleteContainer(n)
+			// Release the infra container name and the pod name for future use
+			if strings.Contains(n, infraName) {
+				s.ReleaseContainerName(n)
+			} else {
+				s.ReleasePodName(n)
 			}
+
 		}
+		// Go through the containers and delete any container that was under the deleted pod
```
TomSweeneyRedHat: nit of a nit, change only if you've other changes:  s/was under/were under/

server/server.go
```diff
@@ -149,38 +149,72 @@ func (s *Server) restore() {
...
+					// Release the container name for future use
+					s.ReleaseContainerName(n)
+				}
+			}
+		}
+		// Add the pod id to the list of deletedPods so we don't try to restore IPs for it later on
+		deletedPods[sbID] = true
 	}
+
+	// Go through all the containers and check if it can be restored. If an error occurs, delete the conainer and
```
TomSweeneyRedHat: nother nit, s/conainer/container/

server/server.go
```diff
@@ -149,38 +149,72 @@ func (s *Server) restore() {
...
+	// release the name associated with you.
```
TomSweeneyRedHat: nit s/you/it/

TomSweeneyRedHat:
@umohnani8 got some test unhappiness

umohnani8:
Tests are green. @mrunalp @rhatdan PTAL

PR #2325, opened 2019-05-02
[1.13] Add username and homedir to generated password

rhatdan:
When generating a an entry in the /etc/passwd file, we
need to also add username and homedir, so that the user
can properly login.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -273,6 +273,17 @@ func setupContainerUser(specgen *generate.Generator, rootfs, mountLabel, ctrRunD
 		return fmt.Errorf("user group is specified without user or username")
 	}
 	imageUser := ""
+	homedir := ""
+	for _, env := range specgen.Config.Process.Env {
+		if strings.HasPrefix(env, "HOME=") {
+			homedir = strings.TrimPrefix(env, "HOME=")
+			break
+		}
+	}
+	if homedir == "" {
+		homedir = specgen.Config.Process.Cwd
+	}
+
 	if imageConfig != nil {
 		imageUser = imageConfig.Config.User
 	}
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -256,9 +268,14 @@ func GeneratePasswd(uid, gid uint32, rootfs, rundir string) (string, error) {
...
-	pwd := fmt.Sprintf("%s%d:x:%d:%d:container user:%s:/bin/sh\n", orig, uid, uid, gid, "/")
-	if err := ioutil.WriteFile(passwdFile, []byte(pwd), 0644); err != nil {
+	if username == "" {
+		username = "default"
+	}
+	if homedir == "" {
+		homedir = "/tmp"
+	}
+	pwd := fmt.Sprintf("%s%s:x:%d:%d:%s user:%s:/sbin/nologin\n", orig, username, uid, gid, username, homedir)
+	if err := ioutil.WriteFile(passwdFile, []byte(pwd), info.Mode()); err != nil {
```
mrunalp: We need to chown the file to the original file's user.group.

utils/utils.go
```diff
@@ -248,6 +248,18 @@ func GeneratePasswd(uid, gid uint32, rootfs, rundir string) (string, error) {
...
+		if os.IsNotExist(err) {
+			return "", nil
+		}
+		return "", errors.Wrapf(err, "unable to stat passwd file %s", originPasswdFile)
+	}
+	// Check if passwd file is world writable
+	if info.Mode().Perm()&(0022) != 0 {
+		return "", nil
+	}
+
```
mrunalp: We need another check for skipping for u+w. So, if uid matches the file owner's uid, then we can skip.

utils/utils.go
```diff
@@ -248,6 +248,24 @@ func GeneratePasswd(uid, gid uint32, rootfs, rundir string) (string, error) {
...
+		if os.IsNotExist(err) {
+			return "", nil
+		}
+		return "", errors.Wrapf(err, "unable to stat passwd file %s", originPasswdFile)
+	}
+	// Check if passwd file is world writable
+	if info.Mode().Perm()&(0022) != 0 {
+		return "", nil
+	}
+	passwdUid := info.Sys().(*syscall.Stat_t).Uid
```
mrunalp: We are missing `syscall` import.

mrunalp:
compilation failed:
```
# github.com/cri-o/cri-o/utils
utils/utils.go:262:28: undefined: syscall
utils/utils.go:263:28: undefined: syscall
make: *** [bin/crio] Error 2
```
otherwise LGTM

server/container_create.go
```diff
@@ -292,23 +303,33 @@ func setupContainerUser(specgen *generate.Generator, rootfs, mountLabel, ctrRunD
...
-	// verify uid exists in containers /etc/passwd, else generate a passwd with the user entry
-	passwdPath, err := utils.GeneratePasswd(uid, gid, rootfs, ctrRunDir)
-	if err != nil {
-		return err
+	genPasswd := true
+	for _, mount := range specgen.Config.Mounts {
+		if mount.Destination == "/etc" ||
+			mount.Destination == "/etc/" ||
+			mount.Destination == "/etc/passwd" {
+			genPasswd = false
```
mrunalp: We can add a break here when we find one of these to avoid looking at the rest.

utils/utils.go
mrunalp: wait - one more question - why aren't we appending to the original file's contents?

server/container_create.go
rhatdan: Fixed

utils/utils.go
rhatdan: Added

utils/utils.go
rhatdan: We are, that is what is in the orig variable.

mrunalp: :+1:

PR #2335, opened 2019-05-06
Update dependencies

saschagrunert:
Hi there, I'm playing around with a workflow for dependency update automation. To achieve (semi-)automatic updates, I utilize [Dependabot](https://dependabot.com/) and a [custom vendoring automatism in CircleCI](https://github.com/openSUSE/cri-o/commit/9601364669f56fde545f6e67783205e02cee7875) within the openSUSE/cri-o fork.

In the end it works like this:

1. Dependabot checks for updaes and submits PRs to the openSUSE/cri-o `module-update` branch
2. Since Dependabot supports no go vendoring yet, we do the vendoring within CircleCI and push into the  branch
3. We wait for the CI to succeed and intervene manually on any kind of failure
4. We tell Dependabot to merge the PR into `module-update`
5. I cherry-pick from  time to time commits from `module-update` into a staging branch `module-update-staging` and submit PRs to cri-o/cri-o

If Dependabot supports vendoring in the future, then the hack at `2.` would not be needed any more. What do you think about this?


vrothberg:
> What do you think about this?

I like it and the per-dependency commits are _really nice_. We should not rely on automation-only for sure, but this looks like a great tool to assist in the dependency management. Maybe we can create the update branch here in CRI-O instead so that all maintainers have access to it?

@mrunalp @umohnani8 @giuseppe @haircommander @rhatdan WDYT?

giuseppe:
> I like it and the per-dependency commits are _really nice_. We should not rely on automation-only for sure, but this looks like a great tool to assist in the dependency management. Maybe we can create the update branch here in CRI-O instead so that all maintainers have access to it?

that looks very useful.  What is the way to trigger it?  Will we run it periodically or before a new release?

saschagrunert:
@chavafg My last commit breaks kata because of the necessary systemd development files coming from the updated libpod dependency. I will fix that in a follow up PR within the kata test repository. For now I will force push this commit away.

References https://github.com/containers/libpod/issues/3079

mrunalp:
@saschagrunert I like the idea :+1:. We can aim for a weekly update of dependencies.

vendor/modules.txt
```diff
@@ -3,7 +3,7 @@ github.com/Azure/go-ansiterm/winterm
 github.com/Azure/go-ansiterm
 # github.com/BurntSushi/toml v0.3.1
 github.com/BurntSushi/toml
-# github.com/Microsoft/go-winio v0.4.11
+# github.com/Microsoft/go-winio v0.4.12
```
mrunalp: This commit doesn't have any code updates. Looks like the dep is commented out?

saschagrunert: The code which has changed within `go-winio` from v0.4.11 to v0.4.12 is not used by CRI-O and therefore not vendored or updated. The comment just indicates the section within the `modules.txt` file, whereas the lines below summarize the import paths:

```
github.com/Microsoft/go-winio
github.com/Microsoft/go-winio/archive/tar
github.com/Microsoft/go-winio/backuptar
```

So this should be fine :)

saschagrunert:
> @saschagrunert I like the idea . We can aim for a weekly update of dependencies.

Sounds good :+1:. So let's aim for a weekly update and I will collect all single updates within the fork (for now).

PR #2340, opened 2019-05-08
[1.14] Add in-place upgrade

haircommander:
Support in-place upgrades in CRI-O

Add functionality to version.go to write, read and compare a version.json file.
If the CRI-O version is sufficiently old (differs by a minor version), then we wipe storage to allow kubelet to repopulate images and containers in a (potentially) new format.
Only do this if --clear-storage-on-update configuration is set, to not surprise anyone by destroying their storage :)

note: once this PR lands, I will add a PR in 1.13 to lay down the version file at startup (though the case of not having a version is already handled)

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -46,7 +46,7 @@ file_locking = {{ .FileLocking }}
 file_locking_path = "{{ .FileLockingPath }}"
 
 # Whether CRI-O should wipe storage after a CRI-O upgrade happens
-auto_upgrade = {{ .AutoUpgrade }}
+clear_storage_on_upgrade = {{ .ClearStorageOnUpgrade }}
 
 
 # The crio.api table contains settings for the kubelet/gRPC interface.
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -45,6 +45,9 @@ file_locking = {{ .FileLocking }}
 # Path to the lock file.
 file_locking_path = "{{ .FileLockingPath }}"
 
+# Whether CRI-O should wipe storage after a CRI-O upgrade happens
+auto_upgrade = {{ .AutoUpgrade }}
+
 
 # The crio.api table contains settings for the kubelet/gRPC interface.
 [crio.api]
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -43,7 +43,7 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 **file_locking_path**="/runc/crio.lock"
   Path to the lock file.
 
-**auto_upgrade**=false
+**clear_storage_on_upgrade**=false
   If set to true, CRI-O will wipe storage after detecting a new version, to allow for an automatic upgrade.
 
 
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -43,6 +43,9 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 **file_locking_path**="/runc/crio.lock"
   Path to the lock file.
 
+**auto_upgrade**=false
+  If set to true, CRI-O will wipe storage after detecting a new version, to allow for an automatic upgrade.
+
 
 ## CRIO.API TABLE
 The `crio.api` table contains settings for the kubelet/gRPC interface.
--- a/server/server.go
+++ b/server/server.go
@@ -141,6 +141,23 @@ func (s *Server) getPortForward(req *pb.PortForwardRequest) (*pb.PortForwardResp
 	return s.stream.streamServer.GetPortForward(req)
 }
 
+func (s *Server) wipeStorage() {
+	containers, err := s.Store().Containers()
+	if err != nil && !os.IsNotExist(errors.Cause(err)) {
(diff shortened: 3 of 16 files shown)


haircommander:
@mrunalp @umohnani8 PTAL

cmd/crio/config.go
```diff
@@ -45,6 +45,9 @@ file_locking = {{ .FileLocking }}
 # Path to the lock file.
 file_locking_path = "{{ .FileLockingPath }}"
 
+# Whether CRI-O should wipe storage after a CRI-O upgrade happens
+auto_upgrade = {{ .AutoUpgrade }}
```
mrunalp: I feel like `auto_upgrade` may not be the best choice for the option name here.
How about something like `clear_storage_on_upgrade` ?

version/version.go
```diff
@@ -1,4 +1,119 @@
...
+	SubMinor  string `json:"sub_minor"`
+	Dev       bool   `json:"dev,omitempty"`
+	GitCommit string `json:"git_commit,omitempty"`
+}
+
+// WriteVersionFile writes the version information to a given file
+// file is the location of the old version file
+// gitCommit is the current git commit version. It will be added to the file
+// to aid in debugging, but will not be used to compare versions
+func WriteVersionFile(file string, gitCommit string) error {
```
mrunalp: Any chance we could make sure of `https://github.com/blang/semver` that we vendor already?

version/version.go
```diff
@@ -1,4 +1,119 @@
...
+	currentVersion := parseVersionConstant(gitCommit)
+	jsonVersion, err := json.Marshal(&currentVersion)
+	// This should never happen
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(file, jsonVersion, 0644)
```
mrunalp: Let's write this file using https://github.com/google/renameio. It ensures atomic writes as well as fsync.

version/version.go
haircommander: upon looking, yes we can do most of this with semver, but we still need to implement the comparison logic (semver doesn't have the option to ignore Patch, Pre, Build... as we need). :+1:

version/version.go
haircommander: ooh fun new package

cmd/crio/config.go
haircommander: kind of a mouthful but I can't think of anything better. sounds good

server/server.go
```diff
@@ -344,6 +344,18 @@ func New(ctx context.Context, config *Config) (*Server, error) {
...
 	}
 
+	if config.ClearStorageOnUpgrade {
+		upgrade, err := version.ShouldCrioUpgrade(CrioVersionPath)
+		if err != nil {
+			logrus.Errorf(err.Error())
+		}
+		if upgrade {
+			logrus.Debugf("New version detected, upgrading")
+			if err = s.Store().Wipe(); err != nil {
```
mrunalp: We need to ensure is that we don't race against any podman containers. (We can test that out.)

saschagrunert:
I'm just wondering: What is the main reason to not target the master branch for this feature?

server/server.go
haircommander: Updated to only wipe crio containers/pods :)

mrunalp:
@saschagrunert We will port it to master but want to start with release branches that we can test in a realistic scenario first.

server/server.go
```diff
@@ -141,6 +141,23 @@ func (s *Server) getPortForward(req *pb.PortForwardRequest) (*pb.PortForwardResp
...
+		logrus.Warnf("could not read containers and sandboxes: %v", err)
+	}
+	for i := range containers {
+		ctr := &containers[i]
+		_, err := s.StorageRuntimeServer().GetContainerMetadata(ctr.ID)
+		if err != nil {
+			continue
+		}
+		for _, name := range ctr.Names {
+			s.Store().DeleteContainer(name)
```
rhatdan: Should be logging failures here.

rhatdan: You are only deleting containers here, what about container images?

docs/crio.conf.5.md
```diff
@@ -43,6 +43,9 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 **file_locking_path**="/runc/crio.lock"
   Path to the lock file.
 
+**clear_storage_on_upgrade**=false
+  If set to true, CRI-O will wipe storage after detecting a new version, to allow for an automatic upgrade.
```
rhatdan: Are we only talking about containers/storage here and not container images?  Then we should fix this message.

server/server.go
haircommander: @rhatdan I am unsure actually.
On the one hand, I can't seem to find a way to only wipe cri-o's image store like I did containers. Plus, having old images on the system won't break kubelet if they have updated, kubelet will just pull new ones.
On the other hand, those images won't be managed by kubelet and thus stick around unnecessarily. So in the interest of cleaning up after ourselves, I think it'd be better to delete them.

I am going to look into trying to clean only cri-o's images, but ultimately I think cleaning the images is the best option. cc @mrunalp

haircommander:
closing in favor of #2397

PR #2344, opened 2019-05-09
Add gocritic appendCombine linter and fixes

saschagrunert:
This commit fixes issues related to the `appendCombine` linter, which
detects `append` chains to the same slice that can be done in a single
`append` call.


TomSweeneyRedHat:
LGTM assuming happy tests.

PR #2349, opened 2019-05-09
Add config hot reload feature (`log_level`)

saschagrunert:
Hey there :wave:, this is an idea I had recently, so I want to put it under discussion here.

I think it would be nice to reload the CRI-O configuration files (`/etc/crio/crio.conf`, `/etc/containers/*`) without restarting the application. This is surely a feature where it has to be decided per option if an application restart is sensible, so I would add an extra documentation part for every "hot reloadable" option.

For a small test I implemented the feature for the `log_level` option within this PR. It actually works by sending `SIGHUP` to the `crio` process:

```
...
DEBU[2019-05-09 14:04:01.433231541+02:00] successfully registered config file watcher
...
DEBU[2019-05-09 14:04:11.294262201+02:00] reloading configuration
INFO[2019-05-09 14:04:11.298918608+02:00] set config log level to "error"
INFO[2019-05-09 14:04:17.583173259+02:00] set config log level to "info"
INFO[2019-05-09 14:04:24.454580923+02:00] set config log level to "debug"
...
```

What do you think? Do you think this is something where we should follow up?

---

**Edit:** Implemented it via the handling of `SIGHUP` and adapted documentation and tests. PTAL

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -17,6 +17,11 @@ var commentedConfigTemplate = template.Must(template.New("config").Parse(`
 #
 # Please refer to crio.conf(5) for details of all configuration options.
 
+# CRI-O supports partial configuration reload during runtime, which can be
+# done by sending SIGHUP to the running process. Currently supported options
+# are explicitly mentioned with: 'This option supports live configuration
+# reload'.
+
 # CRI-O reads its storage defaults from the containers-storage.conf(5) file
 # located at /etc/containers/storage.conf. Modify this storage configuration if
 # you want to change the system's defaults. If you want to modify storage just
@@ -185,7 +190,8 @@ container_attach_socket_dir = "{{ .ContainerAttachSocketDir }}"
 read_only = {{ .ReadOnly }}
 
 # Changes the verbosity of the logs based on the level it is set to. Options
-# are fatal, panic, error, warn, info, and debug.
+# are fatal, panic, error, warn, info, and debug. This option supports live
+# configuration reload.
 log_level = "{{ .LogLevel }}"
 
 # The default log directory where all logs will go unless directly specified by the kubelet
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -8,6 +8,8 @@ crio.conf - configuration file of the CRI-O OCI Kubernetes Container Runtime dae
 # DESCRIPTION
 The CRI-O configuration file specifies all of the available configuration options and command-line flags for the [crio(8) OCI Kubernetes Container Runtime daemon][crio], but in a TOML format that can be more easily modified and versioned.
 
+CRI-O supports partial configuration reload during runtime, which can be done by sending SIGHUP to the running process. Currently supported options are explicitly marked with 'This option supports live configuration reload'.
+
 The default crio.conf is located at /etc/crio/crio.conf.
 
 # FORMAT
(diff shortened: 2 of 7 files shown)


mrunalp:
:+1: to the idea. I think that all options can't be supported (like changing the storage), so what about using a more manual approach of using a SIGHUP? The admin would then be responsible for updating the configuration files and sending a SIGHUP to reload the configuration.

saschagrunert:
Implemented it via the handling of SIGHUP and adapted documentation and tests. PTAL

rhatdan:
LGTM

saschagrunert:

server/config.go
```diff
@@ -190,3 +191,41 @@ func (c *Config) Validate(onExecution bool) error {
...
+	}
+	if err := newConfig.UpdateFromFile(fileName); err != nil {
+		return err
+	}
+
+	// Reload all available options
+	if err := c.ReloadLogLevel(newConfig); err != nil {
+		return err
+	}
+
```
vrothberg: Can we add a `sysregisriesv2.InvalidateCache()` here? This will clear the cache and force crio to reload the `registries.conf` on the next query.

vrothberg: This might be worth documenting as well.

server/server.go
```diff
@@ -538,3 +548,38 @@ func (s *Server) StartExitMonitor() {
...
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, signals.Hup)
+
+	go func() {
+		for {
+			// Block until the signal is received
+			<-c
+			logrus.Infof("reloading configuration %q", fileName)
+			if err := reloadFunc(fileName); err != nil {
+				logrus.Warnf("unable to reload configuration: %v", err)
```
vrothberg: Can we make this an `Errorf()`? I think that's something to be screamed.

cmd/crio/config.go
```diff
@@ -17,6 +17,11 @@ var commentedConfigTemplate = template.Must(template.New("config").Parse(`
 #
 # Please refer to crio.conf(5) for details of all configuration options.
 
+# CRI-O supports partially configuration reload during runtime, which can be
```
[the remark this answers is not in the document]
saschagrunert: Fixed.

docs/crio.conf.5.md
```diff
@@ -8,6 +8,8 @@ crio.conf - configuration file of the CRI-O OCI Kubernetes Container Runtime dae
 # DESCRIPTION
 The CRI-O configuration file specifies all of the available configuration options and command-line flags for the [crio(8) OCI Kubernetes Container Runtime daemon][crio], but in a TOML format that can be more easily modified and versioned.
 
+CRI-O supports partially configuration reload during runtime, which can be done by sending SIGHUP to the running process. Currently supported options are explicitly marked with 'supports live reload'.
```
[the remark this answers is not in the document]
saschagrunert: Yes sure, fixed it.

server/config.go
saschagrunert: I would to carve out this to an additional PR if you don't mind. Within that PR I also would like to add the global system context to be able to modify it in a consistent way.

server/server.go
saschagrunert: Yes, I applied it.

server/config.go
vrothberg: Sure, sounds good to me :+1:

saschagrunert:
@vrothberg Thanks for the review! There is just one discussion point left open.

PR #2350, opened 2019-05-09
conmon: properly set conmon logs

haircommander:
conmon logging levels, syslog and container ID are now set correctly

This fix also fixes compilation for gcc 9 (which only breaks if syslog default is changed in the code, but it was annoying to deal with)


haircommander:
@giuseppe @mrunalp ptal

rhatdan:
LGTM

PR #2356, opened 2019-05-10
conmon: detect cgroup2 and skip OOM handling

giuseppe:
add detection for a cgroup v2 unified mode.  Skip setting up OOM for
now.  It is not proper support for will temporarily unlock Podman.

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -22,6 +22,8 @@
 #include <termios.h>
 #include <unistd.h>
 #include <inttypes.h>
+#include <sys/statfs.h>
+#include <linux/magic.h>
 
 #if __STDC_VERSION__ >= 199901L
 /* C99 or later */
@@ -34,6 +36,10 @@
 
 #include "cmsg.h"
 #include "config.h"
+
+#ifndef CGROUP2_SUPER_MAGIC
+#define CGROUP2_SUPER_MAGIC 0x63677270
+#endif
 
 static volatile pid_t container_pid = -1;
 static volatile pid_t create_pid = -1;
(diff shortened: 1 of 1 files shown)


conmon/conmon.c
```diff
@@ -878,9 +880,17 @@ static int setup_terminal_control_fifo()
 static void setup_oom_handling(int container_pid)
 {
 	/* Setup OOM notification for container process */
-	_cleanup_free_ char *memory_cgroup_path = process_cgroup_subsystem_path(container_pid, "memory");
+	_cleanup_free_ char *memory_cgroup_path = NULL;
 	_cleanup_close_ int cfd = -1;
+	struct statfs sfs;
 	int ofd = -1; /* Not closed */
+
+        if (statfs ("/sys/fs/cgroup", &sfs) == 0 && sfs.f_type == CGROUP2_SUPER_MAGIC) {
```
haircommander: will the statfs call complain if conmon is called as non-root?

haircommander:
LGTM

conmon/conmon.c
giuseppe: no, it works well.  I've tested it locally only as non root :-)

conmon/conmon.c
```diff
@@ -35,6 +37,10 @@
 #include "cmsg.h"
 #include "config.h"
 
+#ifndef CGROUP2_SUPER_MAGIC
+# define CGROUP2_SUPER_MAGIC 0x63677270
```
mheon: Space between `#` and `define`?

giuseppe: I usually use it to indent C preprocessor macros (as done by https://www.gnu.org/software/cppi/)

rhatdan:
LGTM

mrunalp:
@giuseppe clang format failed on the file - https://travis-ci.com/cri-o/cri-o/jobs/199234610

conmon/conmon.c
giuseppe: well clang doesn't like it, so I had to drop it :-)

PR #2359, opened 2019-05-13
Update unit test target to not run `mockgen`

saschagrunert:
The mockgen target takes too much time on local machines to run before
every test. This commit splits up the workflow and should provide a
better local developer experience when running unit tests. The CI
scripts have been adapted to check for modified mocks and fail in that
case.


rhatdan:
LGTM
@mrunalp @umohnani8 @haircommander @giuseppe @vrothberg PTAL

PR #2367, opened 2019-05-14
Cleanup SystemContext usage

saschagrunert:
The SystemContext is now created from the main binaries like
`cmd/crio/main.go` and then passed down to the actual methods. For this
we now store the `systemContext` at the `Server` in substition to the
`ImageContext` of the `ContainerServer`. All tests have been adapted as
well.

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"text/template"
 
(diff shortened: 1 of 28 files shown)


TomSweeneyRedHat:
LGTM
Realy nice cleanup @saschagrunert

PR #2370, opened 2019-05-15
Remove terminal watch after success

haircommander:
We used to continue to listen on the console socket for our terminal master, even though we've already cleaned everything else up.

Instead, we should just remove the callback, so there's no chance of calling it later and printing a very confusing looking loop of "trying to connect, failed to connect".

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -703,7 +703,9 @@ exit:
 	/* Clean up everything */
 	close(connfd);
 
-	return G_SOURCE_CONTINUE;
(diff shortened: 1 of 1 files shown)


haircommander:
(this PR is part fix, part test to see if this fix actually breaks something)

mrunalp:
LGTM

mrunalp:
@giuseppe ptal.

haircommander:
@mrunalp @giuseppe approve? :)

PR #2376, opened 2019-05-16
Add critest to integration test suite

saschagrunert:
The bats integration test suite is now able to run `critest` as well, but
only when specifying the `RUN_CRITEST` environment variable. Otherwise the
test will be skipped. critest will additionally run on CircleCI to not
slow down the overall CI pipeline. Beside that, the critest suite runs
in parallel which requires a working ginkgo executable. For that we use
the internally vendored ginkgo to ensure stable and reproducible test
results.

The Dockerfiles have been adapted to contain critest as well.


Dockerfile
```diff
@@ -70,16 +70,12 @@ RUN VERSION=v0.7.5 &&\
...
-RUN set -x \
-       && export GOPATH="$(mktemp -d)" \
-       && git clone https://github.com/kubernetes-sigs/cri-tools.git "$GOPATH/src/github.com/kubernetes-sigs/cri-tools" \
-       && cd "$GOPATH/src/github.com/kubernetes-sigs/cri-tools" \
-       && git checkout -q "$CRICTL_COMMIT" \
-       && go install github.com/kubernetes-sigs/cri-tools/cmd/crictl \
-       && cp "$GOPATH"/bin/crictl /usr/bin/ \
-       && rm -rf "$GOPATH"
+# Install crictl and critest
+RUN VERSION=v1.14.0 &&\
```
haircommander: Is there a way to get the cri-o version and parse it, instead of setting it here? Adding one here will likely be missed when bumping releases

haircommander: Otherwise, a pointer in version/version.go should exist to make sure it's not missed

saschagrunert: Hm, we could vendor critest and crictl into the repo and only rely on that. Updating would be done via dependabot. WDYT?

saschagrunert:
@chavafg I get a strange error on kata and I'm not completely sure if it is related to my change, WDYT?
```
ERROR: kata-log-parser: no log records to process
```

chavafg:
@saschagrunert 
Hi, I think it's failing because we are currently using `CRICTL_COMMIT` from the `Dockerfile` to get the correct version. I think we need to change that on our script: https://github.com/kata-containers/tests/blob/master/.ci/install_crio.sh#L59

saschagrunert:
> @saschagrunert
> Hi, I think it's failing because we are currently using `CRICTL_COMMIT` from the `Dockerfile` to get the correct version. I think we need to change that on our script: https://github.com/kata-containers/tests/blob/master/.ci/install_crio.sh#L59

Thanks for the hint. :pray: I reverted the change on the original Dockerfile, it is not necessarily needed at all.

test/critest.bats
```diff
@@ -0,0 +1,27 @@
...
+    if [[ -z $RUN_CRITEST ]]; then
+        skip "critest because RUN_CRITEST is not set"
+    fi
+
+    sysctl -w net.ipv4.conf.all.route_localnet=1
+    iptables -t nat -I POSTROUTING -s 127.0.0.1 ! -d 127.0.0.1 -j MASQUERADE
+
+    start_crio
+
+    run critest --parallel $JOBS \
```
chuanchang: Maybe, it had better to also support CRI_FOCUS and CRI_SKIP environment variables in critest, and then tester can only run specified cases or ignore error cases for debugging, the arguments look like this --ginkgo.focus="${CRI_FOCUS}" --ginkgo.skip="${CRI_SKIP}".

chuanchang: Maybe, it had better to also support CRI_FOCUS and CRI_SKIP environment variables in critest, and then tester can only run specified cases or ignore error cases for debugging, the arguments look like this --ginkgo.focus="${CRI_FOCUS}" --ginkgo.skip="${CRI_SKIP}".

chuanchang:
https://github.com/cri-o/cri-o/pull/2376/commits/93f8a79274473fa59c7c689f3c79c310f9391c39#r284807895

test/critest.bats
saschagrunert: Yes sure, I will add the flags

Dockerfile
saschagrunert: This should be resolved now since I do not touch the file any more. I will see If we can find another way to keep the Dockerfile dependencies up to date.

saschagrunert:

haircommander:
LGTM

PR #1721, opened 2018-08-06
Add default_systcls option to crio.conf

umohnani8:
Users can add default sysctls that they want all cri-o containers
to have in the crio.conf file now.
Adding net.ipv4.ping_group_range as a default so we can drop the
NET_RAW capability.

Remove sysctl parsing code from cri-o as we extract the sysctls from k8s directly.
Had a TODO from a while back to remove it, so removing.

Drop the NET_RAW capability from the list of cri-o defaults
as we are using the sysctl net.ipv4.ping_group_range instead
to allow users to do a ping.

Signed-off-by: umohnani8 <umohnani@redhat.com>

PR #2378, opened 2019-05-17
Remove NET_RAW capability and add net.ipv4.ping_group_range sysctl

rhatdan:
also grabbed some changes introduced by @Chenditang on github to
update the crio.conf man page

Increase the security of containers by removing the NET_RAW capabilty.

Add net.ipv4.ping_group_range so that ping will still work.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/contrib/test/integration/system-packages.yml
+++ b/contrib/test/integration/system-packages.yml
@@ -86,3 +86,6 @@
 - name: Update all packages RHEL 8 and Fedora >= 30
   shell: sudo yum update -y
   when: supports_python3 is defined and supports_python3
+
+- name: Allow processes inside of containers to send ping
+  shell: sysctl -w net.ipv4.ping_group_range='0 2147483647'
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -83,11 +83,11 @@ The `crio.api` table contains settings for the kubelet/gRPC interface.
 ## CRIO.RUNTIME TABLE
 The `crio.runtime` table contains settings pertaining to the OCI runtime used and options for how to set up and manage the OCI runtime.
 
-**default_ulimits**=[]
-  A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example:"nofile=1024:2048". If nothing is set here, settings will be inherited from the CRI-O daemon.
-
 **default_runtime**="runc"
   The _name_ of the OCI runtime to be used as the default.
+
+**default_ulimits**=[]
+  A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example:"nofile=1024:2048". If nothing is set here, settings will be inherited from the CRI-O daemon.
 
 **no_pivot**=false
   If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
@@ -122,22 +122,27 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
   The default list is:
 ```
   default_capabilities = [
-          "CHOWN",
-          "DAC_OVERRIDE",
-          "FSETID",
-          "FOWNER",
-          "NET_RAW",
-          "SETGID",
-          "SETUID",
-          "SETPCAP",
-          "NET_BIND_SERVICE",
-          "SYS_CHROOT",
-          "KILL",
+	  "CHOWN",
+	  "DAC_OVERRIDE",
+	  "FSETID",
+	  "FOWNER",
+	  "SETGID",
+	  "SETUID",
+	  "SETPCAP",
+	  "NET_BIND_SERVICE",
+	  "KILL",
   ]
 ```
 
 **default_sysctls**=[]
  List of default sysctls. If it is empty or commented out, only the sysctls defined in the container json file by the user/kube will be added.
+
+  One example would be allowing ping inside of containers.  On systems that support `/proc/sys/net/ipv4/ping_group_range`, the default list could be:
+```
+  default_sysctls = [
+       "net.ipv4.ping_group_range" = "0   2147483647",
+  ]
+```
 
 **additional_devices**=[]
   List of additional devices. Specified as "<device-on-host>:<device-on-container>:<permissions>", for example: "--additional-devices=/dev/sdc:/dev/xvdc:rwm". If it is empty or commented out, only the devices defined in the container json file by the user/kube will be added.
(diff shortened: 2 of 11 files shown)


rhatdan:
@umohnani8 @mrunalp @haircommander @vrothberg @giuseppe PTAL

rhatdan:
@saschagrunert PTAL

haircommander:
Lgtm assuming happy tests

umohnani8:
@rhatdan dropping NET_RAW and using net.ipv4.ping_group_range instead was causing certain failures. It doesn't work in every situation https://github.com/cri-o/cri-o/pull/1721#issuecomment-415176279.

rhatdan:
@umohnani8 Remind me again why this is a problem?

umohnani8:
@rhatdan the issue I ran into was that the sysctl net.ipv4.ping_group_range wasn
t available when in a user namespace on RHEL, https://github.com/cri-o/cri-o/pull/1721#issuecomment-415176279, so the tests were failing.

haircommander:
@rhatdan still working on this?

rhatdan:
If you would like to take if over, that would be great.  I would still like to get this in, but my time is limited.

AkihiroSuda:
Is this discussed in SIG-node? How many projects will break with this?

rhatdan:
I don't think the default settings for CRI-O require this kind of discussion.  If a distribution wants to set it's default kubernetes to run with tighter security, then a committee does not need to debate it.

tutorials/setup.md
```diff
@@ -341,6 +339,11 @@ default_sysctls = [
 ]
 ```
 
+One optional sysctl you might consider, if you want ping to work within
+contianers and your kernel supports it is setting
+
+	"net.ipv4.ping_group_range = 0   429296728",
```
AkihiroSuda: what's this magic number?

rhatdan: I believe this is the max number of gids on a system.

AkihiroSuda: 2147483647?

rhatdan: Ok, I will change, but I thought I found that number from the internet.  Also see a spelling mistake above.

haircommander:
I don't see any code that enables ping_group_range sysctl. Would that be done crio side, or does that also have to be done on the VM e2e tests run @rhatdan

rhatdan:
I think the test suite could turn it on.

haircommander:
Try adding an ansible target in `contrib/test/intregration/system.yml` that calls a shell to add the sysctl

haircommander:
hey @rhatdan can you rebase to fix conflict and get the rest of the ci fixes?

docs/crio.conf.5.md
```diff
@@ -158,6 +163,9 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 
     2) `/usr/share/containers/mounts.conf`: This is the default file read for mounts. If you want CRI-O to read from a different, specific mounts file, you can change the default_mounts_file. Note, if this is done, CRI-O will only add mounts it finds in this file.
 
+**bind_mount_prefix**="prefix"
+  A prefix to use for the source of the bind mounts.  This option would be useful if you were running CRI-O in a container and had the / on the host mounted as /host on your container.  Then if you ran CRI-O with the --bind-mount-prefix=/host option, CRI-O would add /host to any bind mounts it is handed over CRI.  If Kubernetes asked to have /var/lib/foobar bind mounted into the container, then CRI-O would bind mount /host/var/lib/foobar.  Since CRI-O itself is running in a container with / or the host mounted on /host, the container would end up with /var/lib/foobar from the host mounted in the container rather then /var/lib/foobar from the CRI-O container.
```
TomSweeneyRedHat: I'm not a fan of "you" in a man page.  Suggest:  "This option is useful when running CRI-O in a container and the '/' directory on the host is mounted as '/host' in the container.   Then when CRI-O runs with the --bind-mount-prefix=/host ..."

docs/crio.conf.5.md
```diff
@@ -203,6 +211,12 @@ The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  Th
 **runtime_path**=""
   Path to the OCI compatible runtime used for this runtime handler.
 
+**runtime_root**=""
+  Root directory used to store runtime data
+
+**runtime_type**="oci"
+  Type of the runtime used for this runtime handler. "oci", "vm"
```
TomSweeneyRedHat: missing period, ditto line 215

contrib/test/integration/system-packages.yml
```diff
@@ -147,3 +147,6 @@
 - name: Update all packages RHEL 8 and Fedora >= 30
   shell: sudo yum update -y
   when: supports_python3 is defined and supports_python3
+
+- name: Allow processes inside of containers to send ping
+  shell: sysctl -w net.ipv4.ping_group_range '0   2147483647'
```
haircommander: We have to similarly give the sysctl to the circle tests. I'm not really sure how to do that. Update the image maybe? @saschagrunert thoughts?

haircommander:
the location where the sysctl is added is only called in an ami update. I'm going to try to bump the amis after https://github.com/openshift/aos-cd-jobs/pull/2062 merges

haircommander:
that line is failing with:
`TASK [Allow processes inside of containers to send ping] ***********************
task path: /go/src/github.com/cri-o/cri-o/contrib/test/integration/system-packages.yml:151
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "sysctl -w net.ipv4.ping_group_range '0   2147483647'", "delta": "0:00:00.007495", "end": "2019-11-27 18:19:04.360600", "msg": "non-zero return code", "rc": 254, "start": "2019-11-27 18:19:04.353105", "stderr": "sysctl: \"net.ipv4.ping_group_range\" must be of the form name=value\nsysctl: \"0   2147483647\" must be of the form name=value", "stderr_lines": ["sysctl: \"net.ipv4.ping_group_range\" must be of the form name=value", "sysctl: \"0   2147483647\" must be of the form name=value"], "stdout": "", "stdout_lines": []}`

rhatdan:
@haircommander I would like to get back to this.  What do I need to do to update the AMI?

haircommander:
you do
/test ami
and see if the jobs pass. If they don't, AMIs can be manually tagged, but let's see how it goes

haircommander:
still got 
```
# PING 10.88.0.82 (10.88.0.82): 56 data bytes
# 
# ping: permission denied (are you root?)
# 
# non-zero exit code
```
for fedora, though the sysctl call ran...

for rhel
```
 TASK [Make sure we have all required packages] *********************************
task path: /go/src/github.com/cri-o/cri-o/contrib/test/integration/system-packages.yml:5
fatal: [localhost]: FAILED! => {"changed": false, "msg": "No package atomic-registries available."}
	to retry, use: --limit @/go/src/github.com/cri-o/cri-o/contrib/test/integration/main.retry
```
which means I think we should drop atomic-registries from here https://github.com/cri-o/cri-o/blob/92128fe37fe794d16546b6bafcfb53f316804018/contrib/test/integration/system-packages.yml#L9

haircommander:
https://github.com/cri-o/cri-o/blob/master/test/network_ping.bats
it doesn't seem like the sysctl is actually being given to any containers. can you apply the following patch?
```
diff --git a/test/network_ping.bats b/test/network_ping.bats
index 7457d822a..b7841cd20 100644
--- a/test/network_ping.bats
+++ b/test/network_ping.bats
@@ -29,6 +29,7 @@ function teardown() {
 }
 
 @test "Ping pod from another pod" {
+    export CONTAINER_DEFAULT_SYSCTLS='"net.ipv4.ping_group_range" = "0   2147483647"'
     start_crio
     run crictl runp "$TESTDATA"/sandbox_config.json
     echo "$output"
```

haircommander:
@rhatdan I have stolen your PR

rhatdan:
Thanks run with it.

PR #2379, opened 2019-05-17
Update scripts to find correct bash path

saschagrunert:
Some distributions have bash installed under a different path than
`/bin`, which can be retrieved via `env`.


TomSweeneyRedHat:
LGTM assuming happy tests which aren't so happy atm

PR #2388, opened 2019-05-20
Update dependencies

saschagrunert:
This is the weekly module update PR, this time with indirect dependencies included. This is necessary because some projects are now converted to use go modules and we would never get rid of these indirect dependencies when not updating them manually.


saschagrunert:

mrunalp:
I think that we should make these update dependency PRs single commits.

saschagrunert:
> I think that we should make these update dependency PRs single commits.

Alright, so I will squash them together and keep the details in the commit message if this fits better your needs.

PR #2393, opened 2019-05-21
Update e2e tests to be tested against k8s release-1.14

saschagrunert:
This is just yet another approach to see if we can stabilize the e2e tests by early adopting release branches instead of testing the Kubernetes master.

If we decide to go that way we would have to backport the fix to further release branches.

### DIFF
--- a/contrib/test/integration/main.yml
+++ b/contrib/test/integration/main.yml
@@ -24,8 +24,7 @@
     - name: clone build and install kubernetes
       include: "build/kubernetes.yml"
       vars:
-      vars:
-        k8s_git_version: "release-1.12"
+        k8s_git_version: "release-1.14"
         k8s_github_fork: "kubernetes"
         crio_socket: "/var/run/crio/crio.sock"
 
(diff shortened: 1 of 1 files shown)


saschagrunert:
@mrunalp the Kubernetes master and release-1.15 branch seems currently a bit unstable related to our e2e tests. Would you be fine to test with release-1.14 for now?

I think we could create a test document within the repository which specifies how we cross-test which versions of Kubernetes and CRI-O. I'm also not sure which strategy is overall the best, but testing the CRI-O master against a stable Kubernetes version seems reasonable to me (until we have no different appraoch).

My idea for the future would be to create the release-X.XX branch simultaneously to Kubernetes and adapt the test versions there. If the final Kubernetes release happens, we could merge the release branch back into the master to bump the new test versions. So we would have the same feature-freeze period like Kubernetes, but it *could* be more stable.

saschagrunert:
The CI issues seem fixed so this might be closed if not needed any more. :)

haircommander:
Yeah I think we're good to keep tracking master. Though it looks like we should clean up some of these yaml files if there are traces of 1.12 that don't actually run

mrunalp:
https://github.com/kubernetes/kubernetes/pull/78163 fixed the issues :)

I think we can close this.

PR #2395, opened 2019-05-21
conmon: support OOM monitor under cgroup v2

giuseppe:
Add support for OOM handling when running in cgroup v2 unified mode.

Wait for notifications using inotify on the
/sys/fs/cgroup/$PATH/memory.events file.

On an inotify event read the "oom" counter from the same file.

Create an "oom" file each time the counter is different than the
previous known value.

Closes: https://github.com/containers/conmon/issues/8

### DIFF
--- a/conmon/conmon.c
+++ b/conmon/conmon.c
@@ -24,6 +24,7 @@
 #include <inttypes.h>
 #include <sys/statfs.h>
 #include <linux/magic.h>
+#include <sys/inotify.h>
 
 #if __STDC_VERSION__ >= 199901L
 /* C99 or later */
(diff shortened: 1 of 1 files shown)


haircommander:
LGTM, though it may be nice to export cgroup/oom handling to a different module (which I'd like to do for a lot of the other parts of conmon to improve readability/maintainability). Definitely not a blocker though

rhatdan:
Are we ready to switch conmon development to be primarily in github.com/containers/conmon?  And then vendor the changes into CRI-O?

giuseppe:
> Are we ready to switch conmon development to be primarily in github.com/containers/conmon? And then vendor the changes into CRI-O?

I guess we can do another sync to the conmon repository before we make cri-o use conmon as a submodule.  If everyone agrees on that, I can prepare a PR that moves conmon to a submodule

haircommander:
@giuseppe conmon repo is in sync as of now!

haircommander:
I like the idea of using c/conmon, the only thing is we need a way to test with e2e here before we merge to master in conmon. That work flow isn't well defined. Maybe a staging branch we push to in c/conmon, then use the staging branch to run test PRs here, then merge in master of conmon if those pass?

giuseppe:
if we get this in, I'll prepare a PR for c/conmon as well

PR #2397, opened 2019-05-21
[1.14] Lay down version file

haircommander:
To be able to set auto-upgrades, we need the ability write the current CRI-O version to a file
Add this to version/version.go
Add package google/renameio for an atomic write of the version file
Call WriteVersionFile in main, where we have the gitCommit and we've just created the server

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -582,6 +582,11 @@ func main() {
 			logrus.Fatal(err)
 		}
 
+		// Immediately upon start up, write our new version file
+		if err := version.WriteVersionFile(server.CrioVersionPath, gitCommit); err != nil {
+			logrus.Fatal(err)
+		}
+
 		if c.GlobalBool("enable-metrics") {
 			metricsPort := c.GlobalInt("metrics-port")
 			me, err := service.CreateMetricsEndpoint()
--- a/server/config_unix.go
+++ b/server/config_unix.go
@@ -7,3 +7,6 @@ const CrioConfigPath = "/etc/crio/crio.conf"
 
 // CrioSocketPath is where the unix socket is located
 const CrioSocketPath = "/var/run/crio/crio.sock"
+
+// CrioVersionPath is where the CRI-O version file is located
+const CrioVersionPath = "/var/lib/crio/version"
--- a/version/version.go
+++ b/version/version.go
@@ -1,4 +1,58 @@
 package version
 
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/blang/semver"
+	"github.com/google/renameio"
+)
+
 // Version is the version of the build.
 const Version = "1.14.3-dev"
+
+// WriteVersionFile writes the version information to a given file
+// file is the location of the old version file
+// gitCommit is the current git commit version. It will be added to the file
+// to aid in debugging, but will not be used to compare versions
+func WriteVersionFile(file string, gitCommit string) error {
+	current, err := parseVersionConstant(gitCommit)
+	// Sanity check-this should never happen
+	if err != nil {
+		return err
+	}
+	json, err := current.MarshalJSON()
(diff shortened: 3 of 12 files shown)


haircommander:
it is no longer desirable to have storage wiping in CRI-O, so I have added this to replace #2340, where CRI-O is only responsible for laying down the version file

haircommander:
@mrunalp @umohnani8 @rhatdan @saschagrunert PTAL

cmd/crio/main.go
```diff
@@ -588,6 +588,11 @@ func main() {
 			logrus.Fatal(err)
 		}
 
+		// Immediately upon start up, write our new version file
+		if err := version.WriteVersionFile(server.CrioVersionPath, gitCommit); err != nil {
+			logrus.Errorf(err.Error())
```
saschagrunert:
```suggestion
			logrus.Error(err)
```

Do we want to stop there or continue the execution?

server/config_unix.go
```diff
@@ -7,3 +7,6 @@ const CrioConfigPath = "/etc/crio/crio.conf"
 
 // CrioSocketPath is where the unix socket is located
 const CrioSocketPath = "/var/run/crio/crio.sock"
+
+// CrioVersionPath is where the CRI-O version file is located
+const CrioVersionPath = "/etc/crio/version"
```
saschagrunert: I would make this configurable (also in relation to testing), but this definitely only a _nice to have_ in my POV.

version/version.go
```diff
@@ -1,4 +1,51 @@
...
+)
+
 // Version is the version of the build.
 const Version = "1.14.2-dev"
+
+// WriteVersionFile writes the version information to a given file
+// file is the location of the old version file
+// gitCommit is the current git commit version. It will be added to the file
+// to aid in debugging, but will not be used to compare versions
+func WriteVersionFile(file string, gitCommit string) error {
```
saschagrunert: Some unit tests would be great to see here. :innocent:

saschagrunert:
```suggestion
func WriteVersionFile(file, gitCommit string) error {
```

cmd/crio/main.go
haircommander: Hm, I opted to have it be fatal. I'm not married to that though @mrunalp WDYT

version/version.go
haircommander: Added :)

server/config_unix.go
giuseppe: does this information need to be on persistent storage?  If so we should better use `/var/lib`.  If it is not supposed to be persistent then we should use `/run`.

version/version.go
```diff
@@ -1,4 +1,51 @@
...
+	if err != nil {
+		return err
+	}
+	json, err := current.MarshalJSON()
+	// Sanity check-this should never happen
+	if err != nil {
+		return err
+	}
+
+	return renameio.WriteFile(file, json, 0644)
```
giuseppe: related to my previous comment.  Is it supposed to be on persistent storage?

If we write to non persistent storage, we can probably drop the additional dependency as the `fsync` doesn't have any effect on tmpfs.  We can replace it with a simpler write to a ioutil.TempFile() + `os.Rename()`

haircommander: @giuseppe it is supposed to persist, as it's used to check after a reboot whether the storage should be wiped.

haircommander: I went for /var/lib/crio/version

haircommander:
@mrunalp I had to rebase because of 14.2 and lost the lgtm. PTAL

haircommander:
/test e2e_rhel 
>:(

saschagrunert:
@haircommander Do we have a backport of the e2e fix for release-1.14?

haircommander:
@saschagrunert the errors in e2e aren't the same here as they were in master unfortunately :\

PR #2399, opened 2019-05-22
Ensure the test suite configures config directories.

jpeach:
If you run the test suite on a host that already has a default
OCI hooks directory, the test suite can fail because it is unable
to parse the host' hooks configuration. Since the tests should
be isolated from the host, ensure that we always specify a safe
default for directories that can be searched for configuration.

Verify that configuration directories are actually directories,
and add the ginkgo `--trace` option to assist in test failure
debugging.

Signed-off-by: James Peach <jpeach@apache.org>

### DIFF
--- a/utils/filesystem.go
+++ b/utils/filesystem.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"os"
 	"path/filepath"
+	"syscall"
 )
 
 // GetDiskUsageStats accepts a path to a directory or file
@@ -25,4 +26,24 @@ func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, err error) {
 	}
 
 	return dirSize, inodeCount, err
+}
+
+// IsDirectory tests whether the given path exists and is a directory. It
+// follows symlinks.
+func IsDirectory(path string) error {
+	info, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+
+	if !info.Mode().IsDir() {
+		// Return a PathError to be consistent with os.Stat().
+		return &os.PathError{
+			Op:   "stat",
+			Path: path,
+			Err:  syscall.ENOTDIR,
+		}
+	}
+
+	return nil
 }
--- a/Makefile
+++ b/Makefile
@@ -219,6 +219,7 @@ testunit: ${GINKGO}
 	${BUILD_BIN_PATH}/ginkgo \
 		${TESTFLAGS} \
 		-r \
+		--trace \
 		--cover \
 		--covermode atomic \
 		--outputdir ${COVERAGE_PATH} \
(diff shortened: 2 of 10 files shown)


jpeach:
```
              msg: "invalid hooks_dir: stat /var/empty: no such file or directory",
```

I'll update the test fixtures to create an empty test directory.

haircommander:
@sboeuf @chavafg PTAL

haircommander:
LGTM

utils/filesystem.go
```diff
@@ -26,3 +28,33 @@ func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, err error) {
...
+
+	return nil
+}
+
+// MustTempDir uses ioutil.TempDir to create a temporary directory
+// with the given prefix.  It panics on any error.
+func MustTempDir(prefix string) string {
+	path, err := ioutil.TempDir("", prefix)
+	if err != nil {
+		panic(err)
```
saschagrunert: I would recommend to either move this function into the test suite/framework or add non-panicing error handling.

jpeach: Moved it to `TestFramework` and added the cleanup there too.

utils/filesystem.go
```diff
@@ -26,3 +27,22 @@ func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, err error) {
...
 }
+
+// IsDirectory tests whether the given path exists and is a directory.
+func IsDirectory(path string) error {
+	info, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+
+	if !info.Mode().IsDir() {
```
TomSweeneyRedHat: If path is a symlink that points to a valid dir, does this fail and should it?

jpeach:
> On May 23, 2019, at 9:15 PM, Tom Sweeney <notifications@github.com> wrote:
> 
> @TomSweeneyRedHat commented on this pull request.
> 
> In utils/filesystem.go <https://github.com/cri-o/cri-o/pull/2399#discussion_r286892909>:
> 
> > @@ -26,3 +27,22 @@ func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, err error) {
>  
>  	return dirSize, inodeCount, err
>  }
> +
> +// IsDirectory tests whether the given path exists and is a directory.
> +func IsDirectory(path string) error {
> +	info, err := os.Stat(path)
> +	if err != nil {
> +		return err
> +	}
> +
> +	if !info.Mode().IsDir() {
> If path is a symlink that points to a valid dir, does this fail and should it?
> 
This follows symlinks. I can update the doc comment.

utils/filesystem.go
jpeach: Updated the docs to clarify that this follows symlinks. This is what the prior code did, so I kept it the same.

PR #2400, opened 2019-05-22
Add predefined build tags to .golangci.yml

saschagrunert:
Some IDEs are able to lint via a simple invocation of `golangci-lint` in
the background. The recommended buildtags needs to be set in the
configuration to ensure that these tests have strong defaults.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,11 @@
 ---
 run:
+  build-tags:
+    - apparmor
+    - containers_image_ostree_stub
+    - seccomp
+    - selinux
+    - test
   concurrency: 6
   deadline: 5m
 linters:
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -1,5 +1,11 @@
 ---
 run:
+  build-tags:
+    - apparmor
```
rhatdan: What happens if the distribution does not support the build-tag?

saschagrunert: I guess nothing, but let's see if travis complains (should have no apparmor support).

saschagrunert: On travis it works seamless:
https://travis-ci.com/cri-o/cri-o/jobs/202074032#L348

I also tested it on a modified Dockerfile without any apparmor support and this works too. :+1:

rhatdan:
LGTM

rhatdan:
@haircommander @mrunalp @umohnani8 PTAL

TomSweeneyRedHat:
LGTM and happy green tests

PR #2403, opened 2019-05-22
Add live reload integration tests

saschagrunert:
This commits adds the first integration test suite to the live reload
feature of CRI-O. A small refactoring of the start_crio function within
the helpers.bash is included as well. The HTTP "/info" endpoint contains
now a new field "last_reload", which specifies the last configuration
reload via SIGHUP independently if the reload has been successful or
not.

I had to remove the `-t/--tty` from the CircleCI console to avoid adding
newlines to the console output.

### DIFF
--- a/server/inspect.go
+++ b/server/inspect.go
@@ -98,6 +98,20 @@ func (s *Server) getContainerInfo(id string, getContainerFunc, getInfraContainer
 func (s *Server) GetInfoMux() *bone.Mux {
 	mux := bone.New()
 
+	mux.Get("/config", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		b, err := s.config.ToBytes()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/toml")
+		if _, err := w.Write(b); err != nil {
+			http.Error(w, fmt.Sprintf("unable to write TOML: %v", err),
+				http.StatusInternalServerError)
+		}
+	}))
+
 	mux.Get("/info", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		ci := s.getInfo()
 		js, err := json.Marshal(ci)
--- a/test/reload.bats
+++ b/test/reload.bats
@@ -1,0 +1,85 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
+    start_crio
+}
+
+function teardown() {
+    cleanup_test
+}
+
+function replace_config() {
+    sed -ie 's/\('$1' = "\).*\("\)/\1'$2'\2/' "$CRIO_CONFIG"
+}
+
+function reload_crio() {
+    kill -HUP $CRIO_PID
+}
+
+function wait_for_log() {
+    CNT=0
+    while true; do
+        if [[ $CNT -gt 50 ]]; then
+            echo wait for log timed out
+            exit 1
+        fi
+
+        if grep -q "$1" "$CRIO_LOG"; then
+            break
+        fi
+
+        echo "waiting for log entry to appear ($CNT): $1"
+        sleep 0.1
+        CNT=$((CNT + 1))
+    done
+}
+
+function expect_log_success() {
+    wait_for_log '"set config '$1' to \\"'$2'\\""'
+}
+
+function expect_log_failure() {
+    wait_for_log "unable to reload configuration: $1"
+}
+
+@test "should succeed to reload" {
+    # when
+    reload_crio
+
+    # then
+    ps --pid $CRIO_PID &>/dev/null
+}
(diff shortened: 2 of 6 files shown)


test/reload.bats
```diff
@@ -0,0 +1,71 @@
...
+function replace_config() {
+    sed -ie 's/\('$1' = "\).*\("\)/\1'$2'\2/' "$CRIO_CONFIG"
+}
+
+function reload_crio() {
+    # send SIGHUP
+    kill -HUP $CRIO_PID
+
+    # give it a bit time to deliver the signal
+    sleep 1
```
saschagrunert: The helpful reviewer might have a better approach here.

saschagrunert:
I'm experiencing integration test flakes because of gcr.io outages...

/test integration_rhel

test/reload.bats
mrunalp: We can store a reload attempt with success or failure status in the /info endpoint (server/inspect.go). We can then query the endpoint in a loop with a sleep to know whether the attempt succeeded or not and then proceed.

saschagrunert: Thanks for the hint :pray:, I will go for that.

server/inspect.go
```diff
@@ -44,6 +44,7 @@ func (s *Server) getInfo() types.CrioInfo {
 		StorageRoot:       s.config.Config.Root,
 		CgroupDriver:      s.config.Config.CgroupManager,
 		DefaultIDMappings: s.getIDMappingsInfo(),
+		LastReload:        s.lastReload.Unix(),
```
haircommander: Do we still care about cross compilation? there are _windows/_unsupported files, and having this line here will break builds. I am fine with leaving it if we don't care about cross compilation/it doesn't actually break anything

mrunalp:
While thinking about this, I think it will be even better to expose a /config endpoint that returns the current configuration. That way our tests can be more precise and admins can also verify the current configured state of cri-o. WDYT?

saschagrunert:
> While thinking about this, I think it will be even better to expose a /config endpoint that returns the current configuration. That way our tests can be more precise and admins can also verify the current configured state of cri-o. WDYT?

I generally like the idea to provide the configuration state via an API to the customer. So I played around with different approaches and was not very satisfied with any of them. Some issues I came across were:

- We not only have to provide customers access to the content of `/etc/crio/crio.conf` (or how configured), but also to configuration files in `/etc/containers` (like for registries.conf reload). 
- We do not track any configuration file paths at all and I'm not sure if we should, because some state tracking clearly contains to containers/{image/storage}.
- We cannot transform the TOML config into JSON with ease. That means If we simply marshal the JSON from the structs then the customer has to interpret the results like `RuntimePath` vs `runtime_path`. (A solution to this would be to add a struct field tag for JSON to every field, but does not solve the same issue for files in `/etc/containers`).
- The `/info` API would be obsolete.
- Do we still want to provide additional metadata like the last time of configuration reload?

We could add only fields to the API which support live reload. We could also stick only to the CRI-O internal config for now. WDYT?

server/inspect.go
giuseppe: /cc @vbatts

vbatts: We did make a big push for windows compiles, but I don't think that is a blocker for us any longer.

saschagrunert:
I thought a bit about this topic and would like to propose the following changes:

1. Add JSON support to all configuration fields, which have the same name as the existing TOML ones.
2. Propose a dedicated `/config` API to propose the applications configuration state. Since we have multiple configuration, we use the path of the configuration as key:
```json
{
  "/etc/crio/crio.conf": { ... },
  "/etc/containers/registries.conf": { ... }
}
```

WDYT?

mrunalp:
@saschagrunert We could return just the crio.conf; that way we don't worry if anything in the libraries changes. Also, we could just return toml instead of json for the /config endpoint. Does that work?

server/inspect.go
saschagrunert: It will be removed with the next commit. I still do not want to increase technical depth so I guess we should keep a _clean_ separation.

saschagrunert:
> @saschagrunert We could return just the crio.conf; that way we don't worry if anything in the libraries changes. Also, we could just return toml instead of json for the /config endpoint. Does that work?

I added the `/config` endpoint for now which returns the TOML config. The thing is that we cannot indicate reload failures any more (no reload attempt indicator, so the signal could simply not delivered). I let the tests still check for the log messages, which seems robust enough for me. PTAL

saschagrunert:

haircommander:
LGTM @mrunalp PTAL

umohnani8:
LGTM

saschagrunert:
@mrunalp do we want to get this in? I would stay with the new config API (toml based) and integration testing.

haircommander:
LGTM I say let's get this in before 1.15 @mrunalp PTAL

PR #2404, opened 2019-05-24
Create network directory if it doesn't exist

ganeshmaharaj:
The default network directory does not necessarily have to exist before
the cri-o service is started. IF the planned network directory does not
exist, create it when starting the service.

Signed-off-by: Ganesh Maharaj Mahalingam <ganesh.mahalingam@intel.com>

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -548,8 +548,15 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 // `nil`.
 func (c *NetworkConfig) Validate(onExecution bool) error {
 	if onExecution {
-		if err := utils.IsDirectory(c.NetworkDir); err != nil {
-			return errors.Wrapf(err, "invalid network_dir: %s", err)
+		err := utils.IsDirectory(c.NetworkDir)
+		if err != nil {
+			if os.IsNotExist(err) {
+				if err = os.MkdirAll(c.NetworkDir, 0755); err != nil {
+					return errors.Wrapf(err, "Cannot create network_dir: %s", c.NetworkDir)
+				}
+			} else {
+				return errors.Wrapf(err, "invalid network_dir: %s", c.NetworkDir)
+			}
 		}
 
 		for _, pluginDir := range c.PluginDirs {
--- a/lib/config_test.go
+++ b/lib/config_test.go
@@ -271,13 +271,32 @@ var _ = t.Describe("Config", func() {
 			Expect(err).To(BeNil())
 		})
 
-		It("should fail on invalid NetworkDir", func() {
+		It("should create the  NetworkDir", func() {
 			// Given
-			sut.NetworkConfig.NetworkDir = invalidPath
+			tmpDir := path.Join(os.TempDir(), invalidPath)
+			sut.NetworkConfig.NetworkDir = tmpDir
 			sut.NetworkConfig.PluginDirs = []string{validDirPath}
 
 			// When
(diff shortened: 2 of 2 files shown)


lib/config.go
```diff
@@ -545,8 +545,15 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
...
-		if err := utils.IsDirectory(c.NetworkDir); err != nil {
-			return errors.Wrapf(err, "invalid network_dir: %s", err)
+		err := utils.IsDirectory(c.NetworkDir)
+		if err != nil {
+			if os.IsNotExist(err) {
+				if err = os.MkdirAll(c.NetworkDir, 0755); err != nil {
+					return errors.Wrapf(err, "invalid network_dir: %s", err)
+				}
+			} else {
+				return errors.Wrapf(err, "invalid network_dir: %s", err)
```
saschagrunert: This path is not covered by an unit test, can you add one to `lib/config_test.go`?

lib/config.go
```diff
@@ -545,8 +545,15 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 // `nil`.
 func (c *NetworkConfig) Validate(onExecution bool) error {
 	if onExecution {
-		if err := utils.IsDirectory(c.NetworkDir); err != nil {
-			return errors.Wrapf(err, "invalid network_dir: %s", err)
+		err := utils.IsDirectory(c.NetworkDir)
+		if err != nil {
+			if os.IsNotExist(err) {
+				if err = os.MkdirAll(c.NetworkDir, 0755); err != nil {
+					return errors.Wrapf(err, "invalid network_dir: %s", err)
```
giuseppe: please change the error message to something like `"cannot create network_dir: %s", c.NetworkDir`

I see the issue was already existing, but since we are touching this part let's fix it.  There is no need to repeat `err` in the error message.  Having `c.NetworkDir` is more useful.

lib/config.go
giuseppe: same here

mythi:
@ganeshmaharaj one observation about `plugin_dir`:  I have it configured to `plugin_dir = ["/opt/cni/bin"]` but `/opt/cni/bin` does not exist. However, I get `config validation: invalid plugin_dir entry: stat /opt/cni/bin: no such file or directory` even it is supposed to be created.

lib/config.go
ganeshmaharaj: Done. :)

ganeshmaharaj:
Is there a way we can re-test this patch in CI. I am not sure if the failures are related to the changes in this patch.

haircommander:
Can you rebase off of latest master @ganeshmaharaj

rhatdan:
LGTM

lib/config_test.go
```diff
@@ -271,13 +271,31 @@ var _ = t.Describe("Config", func() {
...
 			err := sut.NetworkConfig.Validate(true)
 
+			// Then
+			Expect(err).To(BeNil())
+			os.RemoveAll(tmpDir)
+		})
+
+		It("should fail on invalid NetworkDir", func() {
+			// Given
+			tmpfile := path.Join(os.TempDir(), "wrong-file")
```
saschagrunert: Just a single non-blocking nit: We could use the `TestFramework` method `MustTempDir()` here, which should do the same job.

haircommander:
hey @ganeshmaharaj can you rebase this?

PR #2405, opened 2019-05-24
Update Makefile to be usable without git

saschagrunert:
Some build environments will use the CRI-O sources without containing
the complete git history. To avoid annoying fatal git error messages we
now fallback to "unknown" versions if the git repository is not
available.


TomSweeneyRedHat:
Neat approach @saschagrunert .  LGTM assuming happy test.

PR #2406, opened 2019-05-24
Fix possible runtime panic on store shutdown

saschagrunert:
This commit fixes a runtime panic which can happen on store shutdown,
for example if the `mountpoints.lock` file got deleted during runtime.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM assuming happy tests.

PR #2412, opened 2019-05-28
Version file

haircommander:
To be able to set auto-upgrades, we need the ability write the current CRI-O version to a file
Add this to version/version.go
Add package google/renameio for an atomic write of the version file
Call WriteVersionFile in main, where we have the gitCommit and we've just created the server

forward-port of: https://github.com/cri-o/cri-o/pull/2397
Signed-off-by: Peter Hunt pehunt@redhat.com


haircommander:
@mrunalp @umohnani8 @rhatdan PTAL

version/version.go
```diff
@@ -1,4 +1,58 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/blang/semver"
+	"github.com/google/renameio"
+)
+
 // Version is the version of the build.
-const Version = "1.14.0-dev"
+const Version = "1.14.2-dev"
```
umohnani8: We cut 1.14.2, so might want to change this.

haircommander: hm, do we even update the version file in master as we cut branches? I think this change isn't supposed to be here...

umohnani8: My bad actually, we cut in https://github.com/cri-o/cri-o/pull/2435 but it hasn't gone in yet. And yeah, master keeps track of the `-dev` versions, so it should be there.

haircommander: so we can resolve this?

Issue #2414, opened 2019-05-28
is there a way to detect cgroup driver for crio


chenzhiwei:
**Description**

For Docker, the `docker info` command can output the cgroup driver.

Is there a command can get the cgroup info from crio?

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `crio --version`:**

```
(paste your output here)
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

vrothberg:
You can inspect the configuration crio has been started with and look for the `cgroup_manager` option. The default path is `/etc/crio/crio.conf`. The default cgroup manager is cgroupfs. But there is no docker-info like command.

chenzhiwei:
I know this method, but it is not a perfect method.

it is a static detect method, it is not runtime detection.

think about this case, users modified this file, but did not restart crio,
so the detection will not accurate.

vrothberg:
That's a good point. CRI-O does not support inspecting that yet, but I have an idea using `crictl info` and adding some optional information to what CRI-O returns. I'll spin up a PR in a bit.

vrothberg:
Feel free to try out https://github.com/cri-o/cri-o/pull/2416. This is not at all carved into stone and I expect the PR to turn into something more generic to return other potentially useful configuration values as well. But it'll do the trick for your use case:
```bash
$ sudo crictl info{
  "status": {
    "conditions": [
      {
        "type": "RuntimeReady",
        "status": true,
        "reason": "",
        "message": ""
      },
      {
        "type": "NetworkReady",
        "status": true,
        "reason": "",
        "message": ""
      },
      {
        "type": "CgroupManager",
        "status": true,
        "reason": "",
        "message": "cgroupfs"
      }
    ]
  }
}
```

vrothberg:
Actually, there's a much simpler solution pointed out by @saschagrunert: https://github.com/cri-o/cri-o/pull/2416#issuecomment-496841562

mrunalp:
See #2403 We are adding /config endpoint that will return the settings.

chenzhiwei:
Thank you guys, the method mentioned here looks good to me: https://github.com/cri-o/cri-o/pull/2416#issuecomment-496841562

```
> sudo curl -s --unix-socket /var/run/crio/crio.sock http://localhost/info | jq
{
[...]

PR #2416, opened 2019-05-29
server status: add cgroup manager

vrothberg:
Fixes: #2414

### DIFF
--- a/server/runtime_status.go
+++ b/server/runtime_status.go
@@ -34,11 +34,18 @@ func (s *Server) Status(ctx context.Context, req *pb.StatusRequest) (resp *pb.St
 		networkCondition.Message = fmt.Sprintf("Network plugin returns error: %v", err)
 	}
 
+	cgroupManager := &pb.RuntimeCondition{
+		Type:    "CgroupManager",
+		Status:  true,
+		Message: s.config.CgroupManager,
+	}
+
 	resp = &pb.StatusResponse{
 		Status: &pb.RuntimeStatus{
 			Conditions: []*pb.RuntimeCondition{
 				runtimeCondition,
 				networkCondition,
+				cgroupManager,
 			},
 		},
 	}


vrothberg:
According to the API, only the runtime and network status are required and all other statuses are optional, so we can add any kind of data.

I think that other config options could be valuable as well and that we can generalize #2414 a bit more. But before jumping into that: WDYT @umohnani8 @saschagrunert @haircommander @mrunalp ?

saschagrunert:
> According to the API, only the runtime and network status are required and all other statuses are optional, so we can add any kind of data.

Sure, this looks good to me too.

> I think that other config options could be valuable as well and that we can generalize #2414 a bit more. 

Is there any reason why to not use the `/info` HTTP endpoint which already includes this information?

vrothberg:
> Is there any reason why to not use the /info HTTP endpoint which already includes this information?

Do you know if any cli-client is using that?

saschagrunert:
> > Is there any reason why to not use the /info HTTP endpoint which already includes this information?
> 
> Do you know if any cli-client is using that?

I don't think so, but we could trigger it with any HTTP client:

```
> sudo curl -s --unix-socket /var/run/crio/crio.sock http://localhost/info | jq
{
  "storage_driver": "btrfs",
  "storage_root": "/var/lib/containers/storage",
  "cgroup_driver": "cgroupfs",
  "default_id_mappings": {
    "uids": [
      {
        "container_id": 0,
        "host_id": 0,
        "size": 4294967295
      }
    ],
    "gids": [
      {
        "container_id": 0,
        "host_id": 0,
        "size": 4294967295
      }
    ]
  }
}
```

I guess we have to add documentation about that... :thinking:

vrothberg:
> I guess we have to add documentation about that...

Very good idea. I'll close the PR in favor of the http request :+1:

PR #2419, opened 2019-05-29
Fix runtime panic when having concurrent writes to runtime impl map

saschagrunert:
We now protect the map of runtime implementations with a RWMutex to
allow concurrent reads, but restrict single write access.


haircommander:
ooh good call, LGTM

haircommander:
@saschagrunert also rebase here :)

saschagrunert:

PR #2422, opened 2019-05-29
[1.14] Change plugin_dir to plugin_dirs

haircommander:
To maintain backwards compatibility, change plugin_dir to plugin_dirs in crio configs.
Mark plugin_dir depreciated, and send warning as such.

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -263,8 +263,8 @@ image_volumes = "{{ .ImageVolumes }}"
 network_dir = "{{ .NetworkDir }}"
 
 # Paths to directories where CNI plugin binaries are located.
-plugin_dir = [
-{{ range $opt := .PluginDir }}{{ printf "\t%q,\n" $opt }}{{ end }}]
+plugin_dirs = [
+{{ range $opt := .PluginDirs }}{{ printf "\t%q,\n" $opt }}{{ end }}]
 `))
 
 // TODO: Currently ImageDir isn't really used, so we haven't added it to this
--- a/lib/config.go
+++ b/lib/config.go
@@ -284,7 +284,10 @@ type NetworkConfig struct {
 	NetworkDir string `toml:"network_dir"`
 
 	// PluginDir is where CNI plugin binaries are stored.
-	PluginDir []string `toml:"plugin_dir"`
+	PluginDir string `toml:"plugin_dir,omitempty"`
+
+	// PluginDirs is where CNI plugin binaries are stored.
+	PluginDirs []string `toml:"plugin_dirs"`
 }
 
 // tomlConfig is another way of looking at a Config, which is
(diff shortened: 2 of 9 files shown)


cmd/crio/config.go
```diff
@@ -263,8 +263,8 @@ image_volumes = "{{ .ImageVolumes }}"
 network_dir = "{{ .NetworkDir }}"
 
 # Paths to directories where CNI plugin binaries are located.
-plugin_dir = [
-{{ range $opt := .PluginDir }}{{ printf "\t%q,\n" $opt }}{{ end }}]
+plugin_dirs = [
+{{ range $opt := .PluginDirs }}{{ printf "\t%q,\n" $opt }}{{ end }}]
 `))
```
mrunalp: We will have to retain the older plugin_dir as well, since that's what is used in the MCO template. See https://github.com/openshift/machine-config-operator/blob/6276485e84c53e6531adb540ce494e30170f5876/templates/worker/01-worker-container-runtime/_base/files/crio.yaml#L252

haircommander: I opted to append it to PluginDirs in lib/config validate(), so it does appear in the config because it had already been added to PluginDirs

haircommander: oh you mean print it. Ok

lib/config.go
```diff
@@ -505,10 +508,20 @@ func (c *NetworkConfig) Validate(onExecution bool) error {
...
 
-		for _, pluginDir := range c.PluginDir {
+		for _, pluginDir := range c.PluginDirs {
 			if _, err := os.Stat(pluginDir); err != nil {
+				return errors.Wrapf(err, "invalid plugin_dirs entry")
+			}
+		}
+		// While the plugin_dir option is being depreciated, we need this check
+		if c.PluginDir != "" {
+			logrus.Warnf("The config field plugin_dir is being depreciated. Please use plugin_dirs instead")
```
umohnani8: nit: depreciated to deprecated

umohnani8:
one nit, otherwise LGTM

lib/config.go
haircommander: fixed!

lib/config.go
```diff
@@ -500,10 +503,20 @@ func (c *NetworkConfig) Validate(onExecution bool) error {
...
-		for _, pluginDir := range c.PluginDir {
+		for _, pluginDir := range c.PluginDirs {
 			if _, err := os.Stat(pluginDir); err != nil {
+				return errors.Wrapf(err, "invalid plugin_dirs entry")
+			}
+		}
+		// While the plugin_dir option is being deprecated, we need this check
+		if c.PluginDir != "" {
+			logrus.Warnf("The config field plugin_dir is being deprecated. Please use plugin_dirs instead")
+			if _, err := os.Stat(c.PluginDir); err != nil {
```
jpeach: This should be `utils.IsDirectory()` for consistency and better checking.

jpeach: Oh, this is a backport? Never mind then :)

PR #2423, opened 2019-05-29
 Added EnvVar to cri-o cli flags 

samc24:
Added EnvVar: to each of the flags so that value can be set from the environment itself. EnvVar naming convention is same as the Name, except it is capitalized and separated by underscores instead of hyphens. Now able to define these variables in the environment instead of always defining the flags when starting cri-o.


haircommander:
cheers @samc24 congrats on the first PR :tada: ! 
A couple of nits: 
It looks like you'll need to run `make fmt` to fix the whitespace
It'd be best if you sent along only one commit. You can squash it by using `git rebase -i upstream/master` and choosing to squash the duplicate
Maybe a test or two in our bats test suit located at test/ would be good too :)

mrunalp:
Can you change the logic in the test helper to use this instead of converting env to crio flags?
See https://github.com/cri-o/cri-o/blob/master/test/helpers.bash#L262

haircommander:
hey @samc24 can you rebase off of latest master to fix e2e tests :)

rhatdan:
Should we use CONTAINER_, allowing us to use the same standard for the podman/buildah/skopeo?

samc24:
@mrunalp  @umohnani8 most tests pass locally, but most fail here through the CI tests. Tried debugging, but not sure why they are failing here. Some tests fail locally with the message `[ "$status" -eq 0 ]` or similar. I tried fixing this with Urvashi's help, but to no avail.

umohnani8:
@mrunalp yeah Travis is not liking the switch from the flags to the ENV. It keeps thinking that certain options are not set, and when we switch it back to the fags it works fine. Locally, the test work fine so I am guessing it is a Travis thing.
@samc24 the integration tests are valid failures, so double check the things you are changing.

rhatdan:
@samc24 Still working on this?  Needs a rebase.

haircommander:
@samc24 if we can't get the integration tests working with these, I'd still like the environment variable change. The integration tests can be migrated in a separate PR. WDYT

test/helpers.bash
```diff
@@ -124,6 +138,7 @@ CRIO_CONFIG="$TESTDIR/crio.conf"
 CRIO_CNI_CONFIG="$TESTDIR/cni/net.d/"
 CRIO_LOG="$TESTDIR/crio.log"
 
+# CONTAINER_CNI_CONFIG_DIR=${CONTAINER_CNI_CONFIG_DIR:-"$CRIO_CNI_CONFIG"}
```
umohnani8: Was this left over from earlier changes?

samc24: @umohnani8  Prior to commenting this out, it was failing 3 tests. There were two tests in network.bats that were failing, so even though cni_config_dir isn't a slice, I tried omitting this and it passed the tests, so only one fail now. Planned to delete it in the next commit when hopefully all tests pass.

haircommander:
LGTM

cmd/crio/main.go
```diff
@@ -444,56 +474,66 @@ func main() {
...
+			EnvVar: "CONTAINER_UID_MAPPINGS",
 		},
 		cli.StringFlag{
-			Name:  "gid-mappings",
-			Usage: fmt.Sprintf("specify the GID mappings to use for the user namespace (default: %q)", defConf.GIDMappings),
-			Value: "",
+			Name:   "gid-mappings",
+			Usage:  fmt.Sprintf("specify the GID mappings to use for the user namespace (default: %q)", defConf.GIDMappings),
+			Value:  "",
+			EnvVar: "CONTAINER_GID_MAPPINGS",
```
TomSweeneyRedHat: It would be nice to have the variables in the active and deprecated sections listed alphabetically by name.

PR #2426, opened 2019-05-30
test: set container runtime to remote for e2e and fixup crio.conf

haircommander:
Signed-off-by: Mrunal Patel mrunalp@gmail.com

### DIFF
--- a/contrib/test/integration/e2e-base.yml
+++ b/contrib/test/integration/e2e-base.yml
@@ -1,4 +1,15 @@
 ---
+# only fixup paths for e2e tests which expect to be able to pass 'test-handler' as the runtime handler
+- name: add test-handler runtime handler for Runtimes test
+  blockinfile:
+    path: /etc/crio/crio.conf
+    insertbefore: .*crio.runtime.runtimes.runc.*
+    backup: yes
+    block: |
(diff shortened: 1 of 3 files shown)


umohnani8:
LGTM

mrunalp:
This is missing crio restart after changing crio.conf so it won't pick the changes.

haircommander:
@mrunalp I'll try to do the change before we start crio instead and see how that works.

contrib/test/integration/e2e-base.yml
```diff
@@ -1,4 +1,18 @@
 ---
+# only fixup paths for e2e tests which expect to be able to pass 'test-handler' as the runtime handler
+- name: fixup default runtime for k8s suite
+  replace:
+    regexp: 'default_runtime = "runc"'
```
mrunalp: Instead of replacing runc, we should add another runtime. However, let the current e2e runs verify the fix first :)

haircommander: fix incoming once e2e_fedora also (hopefully) goes green

haircommander:
Nice @umohnani8 @mrunalp @rhatdan PTAL

mrunalp:
You will need to restart the crio service.

PR #2436, opened 2019-05-31
Move conmon to remote repo

haircommander:
We want conmon to solely live in containers/conmon to prevent duplication between repositories.


umohnani8:
Changes LGTM
@haircommander needs a rebase

haircommander:
/hold
lots of work to be done here actually

saschagrunert:
> lots of work to be done here actually

Hm, we could add a `conmon` make target which builds conmon and puts it into `build/bin` for later usage (like we do with ginkgo). :)

haircommander:
@saschagrunert sgtm I'll try it out!

rhatdan:
@haircommander Needs a rebase and to move forward.

haircommander:
/hold
I'm working on moving crio-config to containers/conmon

haircommander:
@saschagrunert how does one access the circleCI images? I need to change the pointer to conmon to a different place (other than cri-o/cri-o/bin/conmon), or install conmon there.

saschagrunert:
> @saschagrunert how does one access the circleCI images? I need to change the pointer to conmon to a different place (other than cri-o/cri-o/bin/conmon), or install conmon there.

The only way right now would be to build the image locally and push it somewhere. Then you could change the used image location in `.circleci/config.yml`. It would be cool if could push the images to a location where we all can access them.

haircommander:
@saschagrunert do you have a Dockerfile you based it off of?

saschagrunert:
> @saschagrunert do you have a Dockerfile you based it off of?
> 
> edit: I bet it's the dockerfile in the root dir of this repo haha

Yes, that’s the image we use for all tests :)

saschagrunert:
Sascha‘s hint of the day (just in case you did not know it already): 

In CircleCI on the jobs page, you can “rerun using SSH”. This restarts the job and provides an anonymous SSH connection to drop you directly into the build node and let you debug the test. 👌💚

saschagrunert:
@haircommander A patch like the attached one should fix the nix package related to the static builds failure. :+1: 
[nix.patch.txt](https://github.com/cri-o/cri-o/files/3460193/nix.patch.txt)

haircommander:
@saschagrunert you're the best! do I need to rebuild the Dockerfile-nix image and change it in the config.yaml, or will the patch alone do the trick?

saschagrunert:
> @saschagrunert you're the best! do I need to rebuild the Dockerfile-nix image and change it in the config.yaml, or will the patch alone do the trick?

It should work alone :) I can update the nix image (takes a hell of CPU time) after 1.16, then we can remove the fix in the `default.nix` derivation.

haircommander:
SGTM

.circleci/config.yml
```diff
@@ -158,6 +159,9 @@ jobs:
       - run:
           name: build
           command: make test-binaries -j $JOBS
+      - run:
+          name: install conmon
+          command: cp $CONMONLOCATION bin
```
saschagrunert: I guess this is needed because the integration tests expect conmon in `$GOPATH/src/github.com/cri-o/cri-o/bin/conmon`. Usually we should not need this because we support this "auto `$PATH` discovery feature". So, simply  removing the conmon path from the integration tests should work, right?

.circleci/config.yml
```diff
@@ -168,6 +172,7 @@ jobs:
             - test/bin2img/bin2img
             - test/checkseccomp/checkseccomp
             - test/copyimg/copyimg
+            - bin/conmon
```
saschagrunert: This can be removed as well if the upper comment applies.

.circleci/config.yml
haircommander: I adapted the test_runner to copy conmon from the path into the repo. there's one integration test that needs to modify conmon, so a copy is copied into the $TESTDIR. I'll look into instead modifying that test :+1:

saschagrunert: Yeah that would be wonderful :+1: Ah I see it's the test which `chmod` to the conmon executable.

.circleci/config.yml
```diff
@@ -141,7 +143,6 @@ jobs:
           command: |
             nix-build nix --argstr revision $CIRCLE_SHA1
             mkdir -p bin
-            cp result-*bin/bin/crio-* bin
```
saschagrunert: We still need this line to copy the statically build binaries to the correct target dir :)

haircommander: There shouldn't be any crio-* though right? that used to be crio-config, but now we just have crio-. Do other binaries have crio- prepended?

saschagrunert: Yes the static binaries :)

haircommander:
waiting on https://github.com/cri-o/cri-o/pull/2667 then this should be ready to go

haircommander:
/hold cancel

this is ready for review @saschagrunert @rhatdan @mrunalp @umohnani8 @vrothberg @giuseppe PTAL

haircommander:
Note to self: i need to update documentation and spec files

.golangci.yml
```diff
@@ -6,6 +6,7 @@ run:
     - seccomp
     - selinux
     - test
+    - exclude_graphdriver_btrfs
```
saschagrunert: May I ask you what the reason for this exclusion is? Can we re-add it?

.circleci/config.yml
```diff
@@ -4,7 +4,8 @@ version: 2.1
 stdenv: &stdenv
   environment:
     GOCACHE: &gocache /tmp/go-build
-    IMAGE: &image saschagrunert/criocircle:1.1.0
+    IMAGE: &image quay.io/haircommander/cri-o-go-1.12:latest
+    IMAGELEGACY: &imagelegacy quay.io/haircommander/cri-o-go-1.10:latest
```
saschagrunert: Can we versionize the images to `1.2.0`? Or do we want to pick another location where we all have write permissions?

Makefile
```diff
@@ -110,8 +110,6 @@ cfmt:
 	find . '(' -name '*.h' -o -name '*.c' ')' ! -path './vendor/*'  -exec clang-format -i {} \+
```
saschagrunert: I think the `fmt` and `cfmt` taret can be removed as well since we don't have any C code left on the repo.

.golangci.yml
haircommander: For some reason, refreshing the images caused lint errors in containers storage because btrfs wasn't installed on the images (a failed #include) . I should add a comment though

.circleci/config.yml
TomSweeneyRedHat: We ought to have these images in a more generic account on quay.io.  Something like quay.io/cri-o/ or the like.  If you need to, you can set up a email list to act as the account owner, that's what we did with quay.io/buildah and quay.io/podman

haircommander: we have quay.io/crio (I also reserved quay.io/cri-o, but we already use crio), I think I'm gonna push these images there, and rename them as @saschagrunert suggested above :)

TomSweeneyRedHat:
Got a bit of test unhappiness.  One comment that could be addressed later, but preferably in the near future.

.circleci/config.yml
haircommander: added them to quay.io/crio, which I think @mrunalp gave you write access :)

.golangci.yml
saschagrunert: I see that issue as well, I will fix this in some time since I'm currently working on a related topic :)

.circleci/config.yml
saschagrunert: Hm, I'm not sure If I have write access to this organization, my quay.io  account exists just since a couple of days: https://quay.io/user/saschagrunert :man_shrugging:

haircommander: I pinged @mrunalp you should be added soon :)

haircommander:
whelp, e2e is busted again...

saschagrunert:
Still looking good, do we have to adapt anything else to move forward with this?

haircommander:
Nope, this should be ready to go!

haircommander:
rebased to fix merge conflict, I'll need my label back

haircommander:
/cherrypick release-1.16

On Wednesday, September 18, 2019, OpenShift Merge Robot <
notifications@github.com> wrote:
> Merged #2436 into master.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.<
https://ci6.googleusercontent.com/proxy/ES0x_fU_rvcpdXKzIRTWlxNJNhs9cfXLmJTUh5BJhm-MDYLn7lpB6xlnyjgb1b0tvgigf898F4e-k0kIKzXWgtb1JUk-qid-HpgYqMVWl7A3KIPMH6ifcnCljS55cymT7Suwtz4vIlvaGSiSBEVzI4CJMwAXEWpzep8Vc2rPaZ1PaSGbeW9rviVKxukalVzH2tUx9sWJGP1RwOCOGw6ebWc8WVY-EDjH8rwAz7l3dmntFK258Pu9ZWIITeOayXA=s0-d-e1-ft#https://github.com/notifications/beacon/ACFJQJ7MHR5URYXKL7FU7X3QKHSRBA5CNFSM4HR5DBM2YY3PNVWWK3TUL52HS4DFWZEXG43VMVCXMZLOORHG65DJMZUWGYLUNFXW5KTDN5WW2ZLOORPWSZGOTWBZNZQ.gif>

PR #2439, opened 2019-06-03
Rebase containers/image to 2.0.0, buildah to 1.8.4, libpod to 1.4.1

mtrmac:
**- What I did**

Updated containers/image.  This adds mirror support to `sysregistriesv2`, using a new `registries.conf` format.

Then trivially convert all current users from the v1 package to v2.

**NOTE**: It's unclear to me why `lib.DefaultConfig` reads `registries.conf` for defaults; AFAICS the values end up only passed in `pkg/storage.GetImageSources`, which already defaults the configuration to the one `in sysregistriesv2`, so explicitly creating that default in the callers as well seems rather redundant.

**- How I did it*

See above.  Right now I expect the PR to fail to build, because buildah and podman need updating for breaking API changes in the sysregistriesv2 package as well.

**- How to verify it**

All existing tests for registries.conf should continue to work, at least. c/image and containers/skopeo have tests that exercise the mirror lookup code.

Supports v2 of the registries.conf format, notably allowing to specify registry mirrors.


saschagrunert:
Hey @mtrmac :wave:, the changes are looking good to me. Don't we want to wait for the official next release of containers/image before we put it in here?

mtrmac:
>Don't we want to wait for the official next release of containers/image before we put it in here?

This is blocked on buildah and podman updating (unless we wanted to patch the vendored copies… ew…), and buildah wants to wait for an official release in https://github.com/containers/buildah/pull/1634#issuecomment-498530594; so, I suppose we are waiting :)

So, this is now ultimately blocked on https://github.com/containers/image/issues/614, probably notably https://github.com/containers/image/pull/617 .

mrunalp:
We will need this in `release-1.14` as well.

umohnani8:
@mtrmac once we cut a new version in containers/image, we will need the changes vendored into release-1.14 as well.

mtrmac:
DO NOT MERGE as is, this includes unmerged containers/buildah#1634 and https://github.com/containers/libpod/pull/3252 . After that is merged, this should revendor from the upstream repos (or an upstream releases) again.

mtrmac:
Updated to only vendor from master branches. Note that the updates brought in a few necessary updates of dependencies, I don’t really understand the possible interactions at all.

saschagrunert:
Hm, the PR will conflict with #2476, which is already on its way into the master...

saschagrunert:
Changes still LGTM :)

mtrmac:
No worries, I can rebase this on top.

Issue #2454, opened 2019-06-07
crio reports names as null for old images that have no parent tag


cduchesne:
**Description**
When a new container image is pulled with the same tag as a previously existing image, the kubernetes status shows null as the `names` for the image. This should probably instead remove the labeled reference for the old image, but still maintain an image@sha256 reference.

**Steps to reproduce the issue:**
1) check status of node via kubernetes api
```
[root@master ~]# kubectl get node node01 -o yaml
apiVersion: v1
kind: Node
metadata:
  name: node01
spec:
  podCIDR: 10.100.0.0/24
status:
  images:
  - names:
    - docker.io/library/redis@sha256:478392fbb4f132b36b8ead3435087553ee7e6f0395b3d64987ae195268d13493
    - docker.io/library/redis:latest
    sizeBytes: 98259235
```

2) pull an updated image with same tag
```
[root@node01 ~]# crictl pull redis
Image is up to date for docker.io/library/redis@sha256:2720a72fb05febf7994e422b7cd327a031dcca2863a808e2825bf3cbb8d4c804
```

3) check updated status via kubernetes api
```
[root@master ~]# kubectl get node node01 -o yaml
apiVersion: v1
kind: Node
metadata:
  name: node01
spec:
  podCIDR: 10.100.0.0/24
status:
  images:
  - names:
    - docker.io/library/redis@sha256:2720a72fb05febf7994e422b7cd327a031dcca2863a808e2825bf3cbb8d4c804
    - docker.io/library/redis:latest
    sizeBytes: 98303778
  - names: null
    sizeBytes: 98259235
```

Once this status shows null for the name, it breaks compatibility with the spec. If you were to try and edit a node, and manually apply an annotation and save it, you would see an error that null is not an allowed value for names.

```
# nodes "node01" was not valid:
# * : Invalid value: "The edited file failed validation": ValidationError(Node.status.images[1]): missing required field "names" in io.k8s.api.core.v1.ContainerImage
```

This will impact anything that is doing validation on the node.status from the received objects. My example is just the simplest way to demonstrate.

**Output of `crio --version`:**
I have tested this in 1.11.12-dev and the latest of 1.14 release branch.
```
# crio --version          
crio version 1.14.3-dev
commit: "615c561b67ac140c1d155d9dc25767a0e81ce433"
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
virtualbox for example above, but should work in any environment

saschagrunert:
[...]

PR #2455, opened 2019-06-07
Fix missing image names on list

saschagrunert:
If images contain no tag or digest then they will now be stubbed with
`<none>:<none>` or `<none>@<none>` to fix invalid node configurations.
The functionality has been refactored into a dedicated function which
will be tested separately now.

Fixes: #2454

### DIFF
--- a/server/image_list.go
+++ b/server/image_list.go
@@ -3,6 +3,7 @@ package server
 import (
 	"time"
 
+	"github.com/cri-o/cri-o/pkg/storage"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
@@ -31,22 +32,43 @@ func (s *Server) ListImages(ctx context.Context, req *pb.ListImagesRequest) (res
 	}
 	resp = &pb.ListImagesResponse{}
 	for i := range results {
-		result := &results[i]
-		resImg := &pb.Image{
-			Id:          result.ID,
-			RepoTags:    result.RepoTags,
-			RepoDigests: result.RepoDigests,
-		}
-		uid, username := getUserFromImage(result.User)
-		if uid != nil {
-			resImg.Uid = &pb.Int64Value{Value: *uid}
-		}
-		resImg.Username = username
-		if result.Size != nil {
-			resImg.Size_ = *result.Size
-		}
-		resp.Images = append(resp.Images, resImg)
+		image := ConvertImage(&results[i])
+		resp.Images = append(resp.Images, image)
 	}
 	logrus.Debugf("ListImagesResponse: %+v", resp)
 	return resp, nil
+}
+
+func ConvertImage(from *storage.ImageResult) *pb.Image {
+	if from == nil {
+		return nil
+	}
+
+	repoTags := []string{"<none>:<none>"}
+	if len(from.RepoTags) > 0 {
+		repoTags = from.RepoTags
+	}
+
+	repoDigests := []string{"<none>@<none>"}
+	if len(from.RepoDigests) > 0 {
+		repoDigests = from.RepoDigests
+	}
+
+	to := &pb.Image{
+		Id:          from.ID,
+		RepoTags:    repoTags,
+		RepoDigests: repoDigests,
+	}
+
+	uid, username := getUserFromImage(from.User)
+	to.Username = username
+
+	if uid != nil {
+		to.Uid = &pb.Int64Value{Value: *uid}
+	}
+	if from.Size != nil {
+		to.Size_ = *from.Size
+	}
+
+	return to
 }
--- a/server/image_list_test.go
+++ b/server/image_list_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/cri-o/cri-o/pkg/storage"
+	"github.com/cri-o/cri-o/server"
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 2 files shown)


saschagrunert:
Hey @chavafg :wave:, it seems that we have an unusual issue with the kata test infrastructure:

```
12:21:08 make[1]: go: Command not found
```

Is there already a known issue?

haircommander:
LGTM

umohnani8:
LGTM

mrunalp:
@mtrmac @vrothberg ptal.

server/image_list.go
```diff
@@ -31,22 +32,43 @@ func (s *Server) ListImages(ctx context.Context, req *pb.ListImagesRequest) (res
...
+
+	repoTags := []string{"<none>:<none>"}
+	if len(from.RepoTags) > 0 {
+		repoTags = from.RepoTags
+	}
+
+	repoDigests := []string{"<none>@<none>"}
+	if len(from.RepoDigests) > 0 {
+		repoDigests = from.RepoDigests
+	}
```
mtrmac: This does not make any sense to me; why invent invalid values to enter into the array (i.e. invent a new string syntax), when an empty / missing array _completely correctly_ describes the actual state (i.e. we already have the ability to represent this with fully-typed data objects)?

If https://github.com/cri-o/cri-o/issues/2454 formats the missing/empty values as text that is then invalid to parse back, either fix the formatting to represent missing/empty in a different way, or fix the parser to accept that syntax to mean missing/empty.

_Either way_ the parser of the edited data in https://github.com/cri-o/cri-o/issues/2454 will need to be taught about a special case, and a “missing/empty value” special case seem much more straightforward than hard-coding “a value exists but has invalid syntax and we have to hard-code an extra check to remove the value — _and then_ run exactly the same special-case code what would have been added for the ‘missing/empty value’ case”

mtrmac: Looking at:
```
# * : Invalid value: "The edited file failed validation": ValidationError(Node.status.images[1]): missing required field "names" in io.k8s.api.core.v1.ContainerImage
```
Either the `names` field should actually be optional, and the validation is just wrong, or the `names` field is ultimately required by _something_, and that something is not going to appreciate an invalid `<none>` string in there (in fact if a name is required, there ideally should be a validator that refuses such a string).

saschagrunert: This implementation is the same behavior as for the current docker shim in Kubernetes. True it makes no sense and we should fix it within Kubernetes, but this does not work for < 1.15.

mtrmac: Ah. Actually, it’s not even the shim, it is returned by the underlying docker API. Ugh.

OK, this is then up to the definition in the CRI. I can’t see that the current spec documents the `<none>` special case, but I’m also completely out of my area of expertise.

>  True it makes up to some point no sense and we should fix it within Kubernetes, but this does not work for < 1.15.

_This_ PR is for the master branch.

Anyway, I don’t know nearly enough about all of this, disregard me.

saschagrunert: No, it’s not really documented and we could do other “solutions” as well where I thought the current one (sticking to the docker behavior to nix break other stuff) is the most “cool one”.

I thought about simply skipping those non-tag/digest images, but this strips of system information and sounded pretty breaking to me.

I will definitely follow up to fix this behavior in Kubernetes 1.16, so we could add comments to the current implementation and remove it after the EOL of k8s 1.15.

mrunalp: I think that for older branches, our best strategy would be to mimic dockershim behavior. We can follow up with upstream k8s fixes for 1.16 :+1: in sig-node.

mrunalp: I think this may have interactions with the kubelet image GC, so should probably check that as well.

cduchesne: Having looked at this and tested briefly, I don't don't think i'm quite onboard. The output should be more realistic, instead of none/none it should look like other images, just without a label.

normal image before losing active label:
```
  - names:
    - docker.io/library/redis@sha256:2720a72fb05febf7994e422b7cd327a031dcca2863a808e2825bf3cbb8d4c804
    - docker.io/library/redis:latest
    sizeBytes: 98305350
```

after losing label (this patch):
```
  - names:
    - <none>@<none>
    - <none>:<none>
    sizeBytes: 98267462
```
what it probably should be (accurate representation):
```
  - names:
    - docker.io/library/redis@sha256:2720a72fb05febf7994e422b7cd327a031dcca2863a808e2825bf3cbb8d4c804
    sizeBytes: 98305350
```

As you can see, the redis/latest is gone, but the sha256 reference should still be here, as it didn't lose that identity. Maybe the problem is occurring in how the image itself is updated when the vanity label is removed?

mtrmac: Part of the issue is that the `repoDigests` value has not actually been accurately tracked in c/storage for images pulled by tag; see what `makeRepoDigests` does.

cduchesne: I see - bummer. I haven't validated with docker, but if the idea is to emulate the same functionality for now, I'm fine with it. I'd still prefer fixing how the digests are stored and ask people to wipe their containers cache if necessary.

I just checked with containerd implementation and see the following:
```
  images:
  - names:
    - docker.io/library/redis@sha256:e549a30b3c31e6305b973e0d9113a3d38d60566708137af9ed7cbdce5650c5cc
    - docker.io/library/redis:latest
    sizeBytes: 35233875
  - names:
    - docker.io/library/redis@sha256:dd5b84ce536dffdcab79024f4df5485d010affa09e6c399b215e199a0dca38c4
    sizeBytes: 35222571
```
```
# /usr/local/bin/crictl images
IMAGE                              TAG                 IMAGE ID            SIZE
docker.io/library/redis            <none>              0f55cf3661e92       35.2MB
docker.io/library/redis            latest              d3e3588af5178       35.2MB
```

With crio (patched, same from comment above)
```
  - names:
    - docker.io/library/redis@sha256:2720a72fb05febf7994e422b7cd327a031dcca2863a808e2825bf3cbb8d4c804
    - docker.io/library/redis:latest
    sizeBytes: 98305350
  - names:
    - <none>@<none>
    - <none>:<none>
    sizeBytes: 98267462
```
```
[root@node01 ~]# crictl images
IMAGE                              TAG                 IMAGE ID            SIZE
docker.io/library/redis            latest              d3e3588af5178       98.3MB
<none>                             <none>              a55fbf438dfd8       98.3MB
```

saschagrunert: @cduchesne Beside the mentioned tracking enhancements (which are a bit out of scope of this fix here), do you see anything not working right now? For me, with crictl it seems to work:

Preparation:
```
> echo "FROM scratch" > Dockerfile

> sudo podman build -t test .
STEP 1: FROM scratch
STEP 2: COMMIT test
--> 703e240592773cd22bbbf811733b6b789eb375ce11f6f84c496afedcfebf163f

> sudo podman images
REPOSITORY       TAG      IMAGE ID       CREATED         SIZE
localhost/test   latest   703e24059277   3 seconds ago   2.07 kB

> sudo podman build -t test .
STEP 1: FROM scratch
STEP 2: COMMIT test
--> 199ae287855b09587e9c0d7ed8affdd217d9bd1cee83270d49341853b5ba3a77

> sudo podman images
REPOSITORY       TAG      IMAGE ID       CREATED          SIZE
localhost/test   latest   199ae287855b   3 seconds ago    2.07 kB
<none>           <none>   703e24059277   29 seconds ago   2.07 kB
```

Before:
```
> sudo crictl images
IMAGE               TAG                 IMAGE ID            SIZE
localhost/test      latest              199ae287855b0       2.07kB
<none>              <none>              703e240592773       2.07kB
```

After:
```
> sudo crictl images
IMAGE               TAG                 IMAGE ID            SIZE
<none>              <none>              703e240592773       2.07kB
localhost/test      latest              199ae287855b0       2.07kB
```

saschagrunert: > I think this may have interactions with the kubelet image GC, so should probably check that as well.

To me it looks like that the GC mostly relies on the image ID, which should not be influenced in any case:
https://github.com/kubernetes/kubernetes/blob/8de1569ddae62e8fab559fe6bd210a5d6100a277/pkg/kubelet/images/image_gc_manager.go#L214-L266

saschagrunert: The moby reference can be found here:
https://github.com/moby/moby/blob/618741ba87e97e72174eea6431a89b78bf829341/daemon/images/images.go#L180-L181

mrunalp: I think we can get this in and then follow up with tracking repoDigests next.

umohnani8:
LGTM

server/image_list.go
cduchesne: @saschagrunert all good here for the short-term fix

PR #2456, opened 2019-06-07
Fix gomega matcher syntax

saschagrunert:
To be consistent with testing we now use `ToNot` instead of `To(Not(...))`


rhatdan:
LGTM

PR #2461, opened 2019-06-10
Default to runc is default_runtime is not set

umohnani8:
To ensure we don't break upgrades, we set the default runtime to
runc if the default_runtime option in crio.conf is not set.

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -27,6 +27,9 @@ const (
 	defaultTransport    = "docker://"
 	apparmorProfileName = "crio-default"
 	cgroupManager       = oci.CgroupfsCgroupsManager
+	defaultRuntime      = "runc"
+	defaultRuntimePath  = "/usr/bin/runc"
+	defaultRuntimeType  = "oci"
 )
 
 // Config represents the entire set of configuration values that can be set for
@@ -369,11 +372,11 @@ func DefaultConfig() *Config {
 			FileLockingPath: lockPath,
 		},
 		RuntimeConfig: RuntimeConfig{
-			DefaultRuntime: "runc",
+			DefaultRuntime: defaultRuntime,
 			Runtimes: map[string]oci.RuntimeHandler{
-				"runc": {
-					RuntimePath: "/usr/bin/runc",
-					RuntimeType: "oci",
+				defaultRuntime: {
+					RuntimePath: defaultRuntimePath,
+					RuntimeType: defaultRuntimeType,
 				},
 			},
 			Conmon: conmonPath,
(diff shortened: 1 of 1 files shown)


umohnani8:
@mrunalp @haircommander @rhatdan ptal

lib/config.go
```diff
@@ -459,7 +459,19 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
 
 	// check we do have at least a runtime
 	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
-		return errors.New("no default runtime configured")
+		// Have this check to default to runc when default_runtime is not set
```
haircommander: nit: consider rewording

lib/config.go
```diff
@@ -459,7 +459,19 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
...
+		if c.DefaultRuntime == "" {
+			logrus.Warnf("Defaulting to `runc` as the runtime since default_runtime is not set")
+			// The default config sets runc and its path in the runtimes map, so check for that
+			// first. If it does not exist then we add runc + its path to the runtimes map.
+			if _, ok := c.Runtimes["runc"]; !ok {
```
haircommander: nit: maybe not have a magic string?

haircommander:
couple of nits, none blockers. All together LGTM

lib/config.go
umohnani8: Fixed

lib/config.go
umohnani8: Agreed, fixed

lib/config.go
```diff
@@ -27,6 +27,9 @@ const (
 	defaultTransport    = "docker://"
 	apparmorProfileName = "crio-default"
 	cgroupManager       = oci.CgroupfsCgroupsManager
+	runcRuntime         = "runc"
```
rhatdan: Can  you call these 
        defaultRuntime         = "runc"
	defaultRuntimePath     = "/usr/bin/runc"
	defaultRuntimeType     = "oci"

lib/config.go
```diff
@@ -459,7 +462,19 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
 
 	// check we do have at least a runtime
 	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
-		return errors.New("no default runtime configured")
+		// Set the default runtime to "runc" if default_runtime is not set
+		if c.DefaultRuntime == "" {
+			logrus.Warnf("Defaulting to `runc` as the runtime since default_runtime is not set")
```
rhatdan: logrus.Warnf("Defaulting to `%s` as the runtime since default_runtime is not set", defaultRuntime)

lib/config.go
umohnani8: Done

lib/config.go
umohnani8: Fixed!

rhatdan:
LGTM

lib/config.go
```diff
@@ -459,7 +462,19 @@ func (c *RuntimeConfig) Validate(onExecution bool) error {
 
 	// check we do have at least a runtime
 	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
-		return errors.New("no default runtime configured")
+		// Set the default runtime to "runc" if default_runtime is not set
+		if c.DefaultRuntime == "" {
+			logrus.Warnf("Defaulting to %q as the runtime since default_runtime is not set", defaultRuntime)
```
cgwalters: Why is this a warning?  This is a *completely normal* condition where we may have an older config file right?

Upgrading is a normal thing :wink:

cgwalters: I don't want to be a pain - but I really think it's worth emphasizing - we have to care *deeply* about upgrades.  It does mean "carrying the weight" of previous design decisions; files that get laid down on disk, particularly config files that admins might change are things that we need to avoid breaking *years* into the future.

I don't know the full backstory here and I may be missing something, but it looks to me like we're going out and emitting a message into the logs because a human (or code) didn't go by and tell cri-o "please use `runc`".  But the use case for changing the runtime seems sufficiently obscure (hm well I guess there's kata and stuff) that again it seems totally normal to me to just silently default to `runc`.

umohnani8: Would making it a `debug` be better? I am thinking of the cases where users don't realize that they have to set `default_runtime` to tell cri-o to use a runtime apart from runc.

umohnani8: Switched it to be a debug message instead.

PR #2469, opened 2019-06-12
Vendor Kubernetes 1.15.0-rc.1

saschagrunert:
The latest RC is just around the corner so we should now rely on that dependency.


rhatdan:
LGTM

PR #2475, opened 2019-06-13
[1.14] Update Makefile to be usable without git

saschagrunert:
### Backport of #2405

---

Some build environments will use the CRI-O sources without containing
the complete git history. To avoid annoying fatal git error messages we
now fallback to "unknown" versions if the git repository is not
available.

Cherry-picked: 33af10124c397511723ef6d59dd72edf11353277

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,7 @@
-include Makefile.inc
-
 GO ?= go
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

PR #2480, opened 2019-06-14
Fix Vagrantfile vendor inconsistency

saschagrunert:
We excluded two vendored files due to the gitignore of all Vagrantfiles
within the repository. This lead into small file inconsistencies on
disk, for example when executing `git clean -fdx`.


rhatdan:
LGTM

PR #2481, opened 2019-06-14
[1.14] When plugin_dir is set, only use that value

haircommander:
We should assume that both plugin_dir and plugin_dirs should not be set. Either a user expects the legacy configuration
or is using the new one. If the user expects the legacy, we should act as though they correctly specified with the new config, rather
than allowing a mixture of new and old


haircommander:
@umohnani8 @mrunalp PTAL

haircommander:
/hold
hm not this solution, not yet.

haircommander:
@saschagrunert yes we will, once I get the right solution :) PTAL at this version

haircommander:
/hold cancel
I feel good about this now

umohnani8:
LGTM

PR #2485, opened 2019-06-17
Re-enable image integration test

saschagrunert:
The mentioned registry issues do not exist any more and therefore we can
enable the test again.


saschagrunert:
The test seems still flakey, so closing this for now.

PR #2486, opened 2019-06-17
Add live reload feature to pause configuration

saschagrunert:
This commit adds the reload feature for the `pause_image`,
`pause_image_auth_file` and `pause_command` options. To be able to
reload the options on a server configuration level, we now have to
let the server pass in the options instead of adding them as member
of the `runtimeServer`. Documentation and tests have been adapted as
well.


haircommander:
nice, LGTM

PR #2488, opened 2019-06-17
Add flake attempts to critest integration testing

saschagrunert:
I see a [flaking critest spec](https://circleci.com/gh/openSUSE/cri-o/7306) when running critest in parallel. I think this happens because image layers get removed because of the parallelism which causes other tests to fail. For now I'd suggest to enable the ginkgo `flakeAttempts` to make the tests more reliable.

References https://github.com/kubernetes-sigs/cri-tools/issues/472

### DIFF
--- a/contrib/test/integration/critest.yml
+++ b/contrib/test/integration/critest.yml
@@ -22,7 +22,7 @@
     state: directory
 
 - name: run critest validation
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

PR #2489, opened 2019-06-17
Add codecov upload step to CircleCI config

saschagrunert:
This step was missing during the migration from travis to CircleCI.
We're now able to upload the coverage directly from CircleCI, too.


rhatdan:
LGTM

PR #2494, opened 2019-06-17
Registries.conf mirrors 1.14

mtrmac:
**- What I did**

This is a 1.14 variant of #2439.

Updated containers/image.  This adds mirror support to `sysregistriesv2`, using a new `registries.conf` format.

Then trivially converted all current users from the v1 package to v2.

**NOTE**: It's unclear to me why `lib.DefaultConfig` reads `registries.conf` for defaults; AFAICS the values end up only passed in `pkg/storage.GetImageSources`, which already defaults the configuration to the one `in sysregistriesv2`, so explicitly creating that default in the callers as well seems rather redundant.

**- How I did it**

See above. Note that this changes a _lot_ of dependencies. I have cherry-picked two pre-existing commits that updated c/storage, to also include the relevant CRI-O updates; other than that, I have no idea about possible version interactions of the various updated packages.

**- How to verify it**

All existing tests for registries.conf should continue to work, at least. c/image and containers/skopeo have tests that exercise the mirror lookup code.

Supports v2 of the registries.conf format, notably allowing to specify registry mirrors.


mrunalp:
My one concern is whether we will hit locking issues in c/storage that we regressed on multiple times during 1.13.x. 1.12.4 of containers/storage seems to be from that time frame and makes me nervous.

mtrmac:
> My one concern is whether we will hit locking issues in c/storage that we regressed on multiple times during 1.13.x.

Might well be the case, I have no idea.

> 1.12.4 of containers/storage seems to be from that time frame and makes me nervous.

Note that this PR ultimately updates to 1.12.10 (same as on the `master` branch); “1.12.4” is mentioned in the summary of a cherry-picked commit that was included for its updates of the CRI-O code that calls c/storage.

mrunalp:
@vrothberg ptal

vrothberg:
I am positive about the c/storage locks. When we hit the flakes, we downgraded instead of upgrading to the version that included fixes. Certainly, we need to make sure about that in the OpenShift CI.

lib/config.go
```diff
@@ -360,8 +359,19 @@ func (c *Config) ToFile(path string) error {
 
 // DefaultConfig returns the default configuration for crio.
 func DefaultConfig() (*Config, error) {
-	registries, _ := sysregistries.GetRegistries(&types.SystemContext{})
-	insecureRegistries, _ := sysregistries.GetInsecureRegistries(&types.SystemContext{})
+	registries, err := sysregistriesv2.UnqualifiedSearchRegistries(&types.SystemContext{})
```
vrothberg: @mtrmac, can we log the errors? Ultimately, I would prefer to return the error but we could defer that to another PR. Currently, there's no (easy) way a user/admin could figure out if the config is loaded; only via the http API or when pulling.

mtrmac: From the PR description:
> NOTE: It's unclear to me why `lib.DefaultConfig` reads `registries.conf` for defaults; AFAICS the values end up only passed in `pkg/storage.GetImage`<s>`Sources`</s>`Service`, which already defaults the configuration to the one in `sysregistriesv2`, so explicitly creating that default in the callers as well seems rather redundant.

So, whether or not this fails, the values read at this point are not actually indicative of what CRI-O will later use. I suppose `GetImageService` could log its final configuration, and that would be representative for the search path (which should ~never be used because it doesn’t work with credentials), but not quite for the insecure registry list (which is also implicitly applied deep inside c/image/docker).

mtrmac: If you don’t mind, I’ll add logging here in a separate PR, so that this PR is consistent between the 1.14 and master branches, and there are no surprises when updating from 1.14.

mtrmac: … hum, actually, there already is a (racy, partial) check in `RuntimeConfig.Validate`.

Looking further, it seems that the `DefaultConfig` value is supposed to be useful for `cri-o config --default`, to write a configuration file equivalent to the built-in defaults. The way I think about it, that really should not contain any copies of data from `registries.conf`; we should default to just using the `registries.conf` values unless the user _explicitly_ decides to override them.

If you agree, I’ll rip these calls out of `DefaultConfig` entirely, and we’ll rely on `GetImageService` calling `sysregistriesv2.*` if the config has not been explicitly set up by the user; the call in  `RuntimeConfig.Validate` can continue to serve as a sanity check that `registries.conf` is valid, at least at the time CRI-O is starting.

vrothberg: I am cool with merging this as is. As you suggested, we can figure out the details of that in another PR.

vrothberg:
LGTM, @mrunalp @umohnani8 PTAL

haircommander:
@rhatdan @umohnani8 @vrothberg could use an approval :)

vrothberg:
I can't since I am not in the OWNERS file. @saschagrunert, can you jump in?

PR #2490, opened 2019-06-17
Encapsulate lib.Config and reuse it directly in the OCI runtime

saschagrunert:
This PR does two major things in terms of refactoring and support of further live reload configuration options:

1. Move `lib.Config` to a dedicated package:
We need to have a dedicated package for the library configuration to
avoid cyclic imports and to be able to refactor functions with a
tremendous amount of parameters.

2. Use the new `*config.Config` within the OCI runtime directly:
This cleans-up the high amount of copying during OCI creation. Beside
this, we're now allowed to update the configuration structure from
outside and have direct influence on the OCI configuration without
required restart.

PR #2495, opened 2019-06-17
Fix error when run root is empty

haircommander:
There exist situations in which run root can be empty for a runtime. The main situation we could encounter is during an upgrade from 1.14 -> master (and soon 1.15), the crio.conf won't have a runtime.runroot specified (because it is a new feature). 

This PR has two possible solutions for this problem. One is we could only specify the run root to the runtime if it is set. This is a bit verbose, but doesn't require hardcoding a default in the runtime_oci.

The second is simply set a default if it is not set. It's much cleaner.

I am opening a WIP PR with both, and upon deciding which (or both) is better I'll update the PR and take the WIP off

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -28,6 +28,10 @@ const (
 	defaultTransport    = "docker://"
 	apparmorProfileName = "crio-default"
 	cgroupManager       = oci.CgroupfsCgroupsManager
+	defaultRuntime      = "runc"
+	defaultRuntimePath  = "/usr/bin/runc"
+	defaultRuntimeType  = "oci"
+	defaultRuntimeRoot  = "/run/runc"
 )
 
 // Config represents the entire set of configuration values that can be set for
@@ -388,12 +392,12 @@ func DefaultConfig(systemContext *types.SystemContext) (*Config, error) {
 			FileLockingPath: lockPath,
 		},
 		RuntimeConfig: RuntimeConfig{
-			DefaultRuntime: "runc",
+			DefaultRuntime: defaultRuntime,
 			Runtimes: map[string]oci.RuntimeHandler{
-				"runc": {
-					RuntimePath: "/usr/bin/runc",
-					RuntimeType: "oci",
-					RuntimeRoot: "/run/runc",
+				defaultRuntime: {
+					RuntimePath: defaultRuntimePath,
+					RuntimeType: defaultRuntimeType,
+					RuntimeRoot: defaultRuntimeRoot,
 				},
 			},
 			Conmon: conmonPath,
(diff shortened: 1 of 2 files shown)


haircommander:
Update: after a little bit of mental deliberation, I decided the second option was superior. The PR is reflective of that. You can still see both options at `github.com/haircommander/cri-o/tree/runroot-hotfix-save` but I feel this is better.

haircommander:
also included a forward port of https://github.com/cri-o/cri-o/pull/2461 while I was in the area :)

rhatdan:
LGTM

oci/runtime_oci.go
```diff
@@ -35,6 +35,9 @@ const (
 
 	// Command line flag used to specify the run root directory
 	rootFlag = "--root"
+
+	// default run root to be used for OCI runtime if it is not set by handler
+	defaultRunRoot = "/run/runc"
```
saschagrunert: If we give this PR precedence to #2490, then I'd have a fix in mind to avoid the duplication of that variable. :)

oci/runtime_oci.go
```diff
@@ -48,10 +51,14 @@ type runtimeOCI struct {
 
 // newRuntimeOCI creates a new runtimeOCI instance
 func newRuntimeOCI(r *Runtime, handler *RuntimeHandler) RuntimeImpl {
+	runRoot := defaultRunRoot
```
saschagrunert: Is the reason why we do this here and not during config validation because the issue can affect already running containers?

If so, then I'd be fine with that. If not I'd suggest setting the runRoot once on configuration validation or even before on CLI argument parsing.

lib/config.go
```diff
@@ -514,7 +518,19 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 
 	// check we do have at least a runtime
 	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
-		return errors.New("no default runtime configured")
+		// Set the default runtime to "runc" if default_runtime is not set
+		if c.DefaultRuntime == "" {
```
saschagrunert: I'm wondering about the reason behind this. Setting `default_runtime = ""` inside the config seems a clear misconfiguration to me, since the value is set but wrong (empty). On the other hand I'd be totally fine with that if we say something is not configured if the value is `""`.

We don't handle this in a consistent way in my point of view. For example we have no default `conmon` path set, and if it's `""`, we fail.

I'm not sure if we can fix these issues without adding workarounds like these, whereas I like the additional checks and we should go for strong defaults.

oci/runtime_oci.go
haircommander: I liked doing it here because it can be the default only for runtime_oci, and I didn't want to effect any other runtimes.

lib/config.go
haircommander: This is for upgrades actually. default_runtime was not a setting and we want to make sure we roll over configurations smoothly.

oci/runtime_oci.go
haircommander: I like the sounds of that!

oci/runtime_oci.go
saschagrunert: I see. I would say it is better to keep it there since with that solution already running workloads still work as expected.

haircommander: sorry--do you mean you'd prefer to move this check into validation? 'there' isn't clear to me

saschagrunert: No, it's fine like it is now. :)

haircommander: SGTM!

haircommander:
@umohnani8 @mrunalp PTAL

oci/runtime_oci.go
TomSweeneyRedHat: Can or should this be in a conf file so it could be changed by the admin if necessary?

umohnani8:
LGTM

PR #2496, opened 2019-06-17
[1.14] Disable file locking by default

haircommander:
File locking adds runtime overhead and thus should be opted into as a user. Disable it by default here


rhatdan:
Would be better to have an explanation of why this is needed in the PR.
Code LGTM

haircommander:
@rhatdan fixed!

PR #2497, opened 2019-06-17
Disable file locking by default and mark depreciated

haircommander:
File locking adds runtime overhead and thus should be opted into as a user. Disable it by default here

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -46,9 +46,11 @@ var commentedConfigTemplate = template.Must(template.New("config").Parse(`
 {{ range $opt := .StorageOptions }}{{ printf "#\t%q,\n" $opt }}{{ end }}#]
 
 # If set to false, in-memory locking will be used instead of file-based locking.
+# **Deprecated** this option will be removed in the future.
 file_locking = {{ .FileLocking }}
 
 # Path to the lock file.
+# **Deprecated** this option will be removed in the future.
 file_locking_path = "{{ .FileLockingPath }}"
 
 
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -361,9 +361,11 @@ func main() {
 			Name:  "storage-opt",
 			Usage: fmt.Sprintf("storage driver option (default: %q)", defConf.StorageOptions),
 		},
+		// XXX: DEPRECATED
 		cli.BoolFlag{
-			Name:  "file-locking",
-			Usage: fmt.Sprintf("enable or disable file-based locking (default: %t)", defConf.FileLocking),
+			Name:   "file-locking",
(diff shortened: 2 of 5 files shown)


rhatdan:
Please explain why this change is being made?

rhatdan:
I think we should document the ramifications of this change in the config file.  

Why do we have an option for file locking at all?

haircommander:
I don't really know why we have the option to be honest. @mrunalp any insight here?

mrunalp:
This came from a libpod option. Hindsight, we shouldn't have added it at all. We can default to false, deprecate it and then remove it in next release.

rhatdan:
SGTM

cmd/crio/main.go
```diff
@@ -362,9 +362,10 @@ func main() {
 			Name:  "storage-opt",
 			Usage: fmt.Sprintf("storage driver option (default: %q)", defConf.StorageOptions),
 		},
+		// XXX: DEPRECATED
 		cli.BoolFlag{
 			Name:  "file-locking",
-			Usage: fmt.Sprintf("enable or disable file-based locking (default: %t)", defConf.FileLocking),
+			Usage: fmt.Sprintf("enable or disable file-based locking (depreciated) (default: %t)", defConf.FileLocking),
```
saschagrunert: Do we want to use `deprecated` instead of `depreciated`?

rhatdan: Should we just Hide the option.  I see no reason to continue to document an option that we no longer want to support.  We can put a reference into the man page, but no reason to show the option crio --help

cmd/crio/config.go
```diff
@@ -46,9 +46,11 @@ var commentedConfigTemplate = template.Must(template.New("config").Parse(`
 {{ range $opt := .StorageOptions }}{{ printf "#\t%q,\n" $opt }}{{ end }}#]
 
 # If set to false, in-memory locking will be used instead of file-based locking.
+# **Depreciated** this option will be removed in the future.
```
TomSweeneyRedHat: Deprecated through out please.  Deprecated == obsolete, Depreciated == less monetary value

cmd/crio/main.go
TomSweeneyRedHat: Any concerns about people already using the option and adjustments they might need to make?

cmd/crio/config.go
haircommander: :O thanks!

cmd/crio/main.go
haircommander: I am personally not concerned about it. I don't think it's thought about very much

rhatdan: Existing users will continue to work, if hidden, and document the deprecation in the man page.

haircommander:
@TomSweeneyRedHat @saschagrunert @vrothberg @mrunalp @umohnani8 @rhatdan PTAL

PR #2498, opened 2019-06-18
[1.14] Add global_auth_file option to crio.image config

umohnani8:
Add a global_auth_file option under crio.image to
the crio config. This option points to a json file
that holds the credentials required to pull images
from a secure registry. The json file is similar to
that of .docker/config.json.

Will forward-port to master as well.

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -184,11 +184,14 @@ CRI-O reads its configured registries defaults from the system wide containers-r
 **default_transport**="docker://"
   Default transport for pulling images from a remote container storage.
 
+**global_auth_file**=""
+  The path to a file like /var/lib/kubelet/config.json holding credentials necessary for pulling images from secure registries.
+
 **pause_image**="k8s.gcr.io/pause:3.1"
   The image used to instantiate infra containers.
 
 **pause_image_auth_file**=""
- If not empty, the path to a docker/config.json-like file containing credentials necessary for pulling the image specified by pause_image above.
+ The path to a file like /var/lib/kubelet/config.json holding credentials specific to pulling the pause_image from above.
 
 **pause_command**="/pause"
   The command to run to have a container stay in the paused state.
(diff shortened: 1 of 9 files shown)


umohnani8:
@mrunalp @mtrmac @haircommander PTAL

haircommander:
LGTM

docs/crio.conf.5.md
```diff
@@ -184,6 +184,9 @@ CRI-O reads its configured registries defaults from the system wide containers-r
 **default_transport**="docker://"
   Default transport for pulling images from a remote container storage.
 
+**global_auth_file**=""
+  If not empty, the path to a docker/config.json-like file containing credentials necessary for pulling images from secure registries.
```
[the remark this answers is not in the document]
umohnani8: Fixed up the docs and comments

pkg/storage/image.go
```diff
@@ -407,7 +408,11 @@ func (svc *imageService) PullImage(systemContext *types.SystemContext, imageName
 		return nil, err
 	}
 	if options == nil {
-		options = &copy.Options{}
+		options = &copy.Options{
+			SourceCtx: &types.SystemContext{
+				AuthFilePath: svc.globalAuthFile,
+			},
```
mtrmac: Looking at current non-test callers, this will never trigger AFAICS; both are passing a non-nil `options`.

(And if `svc.globalAuthFile` is effectively never used, it should not exist.)

There are _quite a few_ other callers that refer to the image via `ImageReference.NewImage{,Source}`; shouldn’t all of them also have the `AuthFilePath` value set up, at some point?

I don’t quite know how to think about the conceptual separation between `ImageServer` and its callers; still, the current placement of the `AuthFilePath` option seems incomplete/inconsistent.

umohnani8: The main place we need this authFile to be set is when pulling down images right? I was looking at the other callers of `ImageReference.NewImage{}` and don't think those need AuthFilePath value set up. If you are worried about a particular part, can you point me to it in case I missed it.

mtrmac: https://github.com/cri-o/cri-o/pull/2510

But, wow, what a maze.  The `alltransports.ParseImageName` in `imageService.getRef` opens up many… possibilities.

umohnani8:
Closing in favour of #2510

PR #2500, opened 2019-06-19
WIP: Update RemoveImage to never fail

saschagrunert:
The CRI API protocol definition states that the RPC `RemoveImage` should
be idempotent, and must not return an error if the image has already
been removed or is not parsable. All test cases have been adapted as
well.

Relates to https://github.com/kubernetes-sigs/cri-tools/issues/472

This is the 2500th PR! :tada:

### DIFF
--- a/server/image_remove.go
+++ b/server/image_remove.go
@@ -10,7 +10,12 @@ import (
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
-// RemoveImage removes the image.
+// RemoveImage removes the image. If the image does not exist, then this
+// function should not error at all.
+//
+// For further documentation, see:
+// https://github.com/kubernetes/cri-api/blob/261df499b74595bc2ce546130f1edfcc6f05d2c1/pkg/apis/runtime/v1alpha2/api.proto#L123-L124
+// https://github.com/kubernetes/kubernetes/blob/71a7be41e0518a7df549eb1fa7b51e7c647d985b/pkg/kubelet/dockershim/docker_image.go#L122-L156
 func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (resp *pb.RemoveImageResponse, err error) {
 	const operation = "remove_image"
 	defer func() {
@@ -18,7 +23,7 @@ func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (r
 		recordError(operation, err)
 	}()
 
-	logrus.Debugf("RemoveImageRequest: %+v", req)
+	logrus.Debugf("%s: request: %+v", operation, req)
 	image := ""
 	img := req.GetImage()
 	if img != nil {
@@ -27,11 +32,8 @@ func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (r
 	if image == "" {
 		return nil, fmt.Errorf("no image specified")
 	}
-	var (
-		images  []string
-		deleted bool
-	)
-	images, err = s.StorageImageServer().ResolveNames(s.systemContext, image)
+
+	images, err := s.StorageImageServer().ResolveNames(s.systemContext, image)
 	if err != nil {
 		if err == storage.ErrCannotParseImageID {
 			images = append(images, image)
@@ -42,16 +44,12 @@ func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (r
 	for _, img := range images {
 		err = s.StorageImageServer().UntagImage(s.systemContext, img)
 		if err != nil {
-			logrus.Debugf("error deleting image %s: %v", img, err)
+			logrus.Infof("%s: error deleting image %s: %v", operation, img, err)
 			continue
 		}
-		deleted = true
 		break
 	}
-	if !deleted && err != nil {
-		return nil, err
-	}
 	resp = &pb.RemoveImageResponse{}
-	logrus.Debugf("RemoveImageResponse: %+v", resp)
+	logrus.Debugf("%s: response: %+v", operation, resp)
 	return resp, nil
 }
--- a/server/image_remove_test.go
+++ b/server/image_remove_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 // The actual test suite
-var _ = t.Describe("ImageRemove", func() {
+var _ = t.Describe("RemoveImage", func() {
 	// Prepare the sut
 	BeforeEach(func() {
 		beforeEach()
@@ -21,7 +21,7 @@ var _ = t.Describe("ImageRemove", func() {
 
 	const imageID = "imageID"
 
-	t.Describe("ImageRemove", func() {
+	t.Describe("RemoveImage", func() {
 		It("should succeed", func() {
 			// Given
 			gomock.InOrder(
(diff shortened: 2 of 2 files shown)


server/image_remove.go
```diff
@@ -17,41 +16,36 @@ func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (r
...
+	logrus.Debugf("%s: request: %+v", operation, req)
 	image := ""
 	img := req.GetImage()
 	if img != nil {
 		image = img.Image
 	}
 	if image == "" {
-		return nil, fmt.Errorf("no image specified")
+		logrus.Infof("%s: no image specified in request", operation)
+		return resp, nil
```
vrothberg: Just to double check: An empty string qualifies as "not parsable", right?

server/image_remove.go
```diff
@@ -17,41 +16,36 @@ func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (r
...
-	var (
-		images  []string
-		deleted bool
-	)
-	images, err = s.StorageImageServer().ResolveNames(s.systemContext, image)
+
+	images, err := s.StorageImageServer().ResolveNames(s.systemContext, image)
 	if err != nil {
 		if err == storage.ErrCannotParseImageID {
 			images = append(images, image)
```
vrothberg: Can we drop a comment for that special case linking to the CRI doc mentioning that?

server/image_remove.go
```diff
@@ -17,41 +16,36 @@ func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (r
...
 		} else {
-			return nil, err
+			logrus.Infof("%s: unable to resolve image %q: %v", operation, image, err)
+			return resp, nil
```
vrothberg: I think we need to return this error. Otherwise, we'd mask errors from, for instance, `sysregistriesv2`.

saschagrunert:
> Could we add some links/references to the CRI docs mentioning that? Especially the fact that parsing errors should be ignored sounds exotic enough for readers of the code to get stuck.

The only documentation I found was this one:
https://github.com/kubernetes/cri-api/blob/261df499b74595bc2ce546130f1edfcc6f05d2c1/pkg/apis/runtime/v1alpha2/api.proto#L123-L124

I think more does not exist, so I'm unsure If we still should return errors. WDYT?

server/image_remove.go
saschagrunert: Yes :+1:

vrothberg:
@saschagrunert, using that as a link (or only the file link in case the lines change) should suffice.

>  // This call is idempotent, and must not return an error if the image has
// already been removed.

I think we still need to return all errors other than the _already-removed_ case.

Do you have a reference for ignoring the parsing errors?

server/image_remove.go
saschagrunert: I added further docs above the function.

saschagrunert:
> I think we still need to return all errors other than the _already-removed_ case.

I agree and finally got it now.

> Do you have a reference for ignoring the parsing errors?

Yeah I added the reference implementation and the docs about the CRI API.

PTAL and thanks for taking your time to review this :pray:

server/image_remove.go
```diff
@@ -42,16 +45,11 @@ func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (r
...
 		if err != nil {
-			logrus.Debugf("error deleting image %s: %v", img, err)
+			logrus.Infof("%s: error deleting image %s: %v", operation, img, err)
 			continue
 		}
-		deleted = true
 		break
 	}
-	if !deleted && err != nil {
-		return nil, err
```
vrothberg: I'm still a bit scared about removing the errors entirely. In case an image has already been removed (or doesn't exist in general), we should get a `containers/image/storage.ErrNoSuchImage` (https://github.com/containers/image/blob/master/storage/storage_reference.go#L56) and only those should be ignored while others should be returned.

@nalind @mtrmac PTAL

saschagrunert: The issue originates from the flaking critest suite when running in parallel. The error I saw there was mainly "layer not known" because another test already removed the layer:
https://github.com/containers/storage/blob/4442d3aa2f8218e930c389535d26496edd0e07e6/errors.go#L15

vrothberg: This sounds like a bug in c/storage. @nalind, @rhatdan, @giuseppe can you have a look at that? I'll be heading to the train station in a bit.

server/image_remove.go
```diff
@@ -10,15 +10,21 @@ import (
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
-// RemoveImage removes the image.
+// RemoveImage removes the image. If the image does not exist, then this
+// function should not error at all.
+//
+// For further documentation, see:
+// https://github.com/kubernetes/cri-api/blob/261df499b74595bc2ce546130f1edfcc6f05d2c1/pkg/apis/runtime/v1alpha2/api.proto#L123-L124
+// https://github.com/kubernetes/kubernetes/blob/71a7be41e0518a7df549eb1fa7b51e7c647d985b/pkg/kubelet/dockershim/docker_image.go#L122-L156
```
mtrmac: Note that this implements something rather different in general: it removes all tags of the image and then presumably the image, whereas this implementation, via `UntagImage`, will just remove a tag(/digest?) if the input is a tag, leaving the image itself around.

(Assuming the comments at the link are correct and the input can only be an image ID, there probably isn’t a difference… but there are way too many abstraction layers that are way too liberal about the input they accept, so it’s hard to be sure.)

server/image_remove.go
mtrmac: I agree that a “layer not known” seems like a problem in c/storage if the caller is not manipulating layers individually.

> I'm still a bit scared about removing the errors entirely. In case an image has already been removed (or doesn't exist in general), we should get a containers/image/storage.ErrNoSuchImage (https://github.com/containers/image/blob/master/storage/storage_reference.go#L56) and only those should be ignored while others should be returned.

Yes, ideally. OTOH, with all the abstraction layers involved and all the helpful flexible helpers, it’s not _trivially_ obvious that the destination is always c/storage. AFAICS it _is_ the case, but only because `ResolveNames` enforces the input to be an image ID or a docker reference; without that, `UntagImage` would happily try to remove a `dir:latest`.

server/image_remove.go
```diff
@@ -10,15 +10,21 @@ import (
...
 func (s *Server) RemoveImage(ctx context.Context, req *pb.RemoveImageRequest) (resp *pb.RemoveImageResponse, err error) {
 	const operation = "remove_image"
 	defer func() {
 		recordOperation(operation, time.Now())
 		recordError(operation, err)
 	}()
+	resp = &pb.RemoveImageResponse{}
```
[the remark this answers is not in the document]
saschagrunert: I agree and changed it as suggested.

saschagrunert:
Rebased on top of the latest master branch to get in the latest CI fixes.

vrothberg:
https://github.com/cri-o/cri-o/pull/2500/files#r295275846 still holds. I don't think we can ignore these errors.

server/image_remove.go
vrothberg: Ewww. Well, we still cannot just ignore the errors here.

saschagrunert:
I will close this for now. Since the issue seems to be storage related I don't see any reason for changing the `RemoveImage` implementation.

PR #2502, opened 2019-06-20
RFC: Don't read registries.conf for the defaults of --registry and --insecure-registry

mtrmac:
**- What I did**

Motivated by https://github.com/cri-o/cri-o/pull/2494#pullrequestreview-251539419, changed `lib.DefaultConfig` to not read defaults from `registries.conf`.

**- How I did it**

Don't read `registries.conf` for the defaults of `--registry` and `--insecure-registry`:

In both cases, the `ImageService` falls back to `registries.conf` anyway if the values are not specified (explicitly for the search path, and implicitly in containers/image/docker for insecure registries).  The default values in `lib.DefaultConfig` really make a difference only:
- in the help text of `--registry`, which was a bit misleading by suggesting a hard-coded default instead of pointing to `registries.conf`
- in the output of `crio config --default`, which previously copied the search list from `registries.conf` into the generated output, probably something the user does not really want (but it did not matter anyway because the generated output comments the value out)

Also, the `--insecure-registry` list as implemented in CRI-O only works at the level of registries, while the configuration in `sysregistriesv2` now works at the repository level; this did not really hurt (the repository entries were ignored in CRI-O and applied in c/image/docker), but it caused invalid/ignored entries to be used throughout the CRI-O codebase.

It's overall much simpler to just use empty defaults, and let the fallback happen elsewhere.

(Note that `lib.RuntimeConfig.Validate` does call `sysregistriesv2.GetRegistries`, which ensures that `registries.conf` is reasonably valid at the time CRI-O is starting, so this does not affect `registries.conf` validation / impact on CRI-O startup.)

Then, simplified the call stack by dropping unnecessary `types.SystemContext` parameters.

**- How to verify it**

Existing tests, hopefully. The `crio --help` and `crio config --default` output should change a bit.

Data automatically used from `registries.conf` no longer shows up as implied-to-be-hard-coded defaults for CRI-O options.

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -371,7 +371,7 @@ func main() {
 		},
 		cli.StringSliceFlag{
 			Name:  "registry",
-			Usage: fmt.Sprintf("registry to be prepended when pulling unqualified images, can be specified multiple times (default: %q)", defConf.Registries),
+			Usage: fmt.Sprintf("registry to be prepended when pulling unqualified images, can be specified multiple times (default: configured in /etc/containers/registries.conf)"),
 		},
 		cli.StringFlag{
 			Name:  "default-transport",
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -274,7 +274,7 @@ func main() {
 	app.Version = strings.Join(v, "\n")
 
 	systemContext := &types.SystemContext{}
-	defConf, err := server.DefaultConfig(systemContext)
+	defConf, err := server.DefaultConfig()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "error loading server config: %v", err)
 		os.Exit(1)
--- a/server/config.go
+++ b/server/config.go
@@ -157,8 +157,8 @@ func (c *Config) ToBytes() ([]byte, error) {
 }
 
 // DefaultConfig returns the default configuration for crio.
-func DefaultConfig(systemContext *types.SystemContext) (*Config, error) {
-	conf, err := config.DefaultConfig(systemContext)
+func DefaultConfig() (*Config, error) {
+	conf, err := config.DefaultConfig()
 	if err != nil {
 		return nil, err
 	}
@@ -207,9 +207,9 @@ func (c *Config) Validate(systemContext *types.SystemContext, onExecution bool) 
 
 // Reload reloads the configuration with the config at the provided `fileName`
 // path. The method errors in case of any read or update failure.
-func (c *Config) Reload(systemContext *types.SystemContext, fileName string) error {
+func (c *Config) Reload(fileName string) error {
 	// Reload the config
-	newConfig, err := DefaultConfig(systemContext)
+	newConfig, err := DefaultConfig()
 	if err != nil {
 		return fmt.Errorf("unable to create default config")
 	}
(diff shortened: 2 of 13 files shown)


mtrmac:
Note to self: If this PR is accepted, it should be also done on 1.14 after #2494.

server/config.go
```diff
@@ -207,9 +207,9 @@ func (c *Config) Validate(systemContext *types.SystemContext, onExecution bool)
 
 // Reload reloads the configuration with the config at the provided `fileName`
 // path. The method errors in case of any read or update failure.
-func (c *Config) Reload(systemContext *types.SystemContext, fileName string) error {
+func (c *Config) Reload(fileName string) error {
 	// Reload the config
-	newConfig, err := DefaultConfig(systemContext)
+	newConfig, err := DefaultConfig()
```
vrothberg: Should we add an explicit `GetRegistries(systemContext)` here? I think it would be nice to keep that since a user gets immediate feedback when changing the configs and sending HUP to crio.

saschagrunert: Would that somehow incorporate with https://github.com/containers/image/pull/630?

saschagrunert:
I like that, makes everything less complicated.

/retest

cmd/crio/main.go
```diff
@@ -372,7 +372,7 @@ func main() {
 		},
 		cli.StringSliceFlag{
 			Name:  "registry",
-			Usage: fmt.Sprintf("registry to be prepended when pulling unqualified images, can be specified multiple times (default: %q)", defConf.Registries),
+			Usage: fmt.Sprintf("registry to be prepended when pulling unqualified images, can be specified multiple times (default: configured in /etc/containers/registries.conf)"),
```
haircommander: Do we want a similar comment about the default for `--insecure-registry`?

haircommander:
outstanding nits aside this change LGTM

umohnani8:
changes LGTM

server/config.go
vrothberg: Good point, yes. We could either use that or invalidate and reload (for now).

mtrmac: >I think it would be nice to keep that since a user gets immediate feedback when changing the configs and sending HUP to crio.

Without invalidation the `GetRegistries()` does not return updated data, so there is no feedback.

https://github.com/containers/image/pull/630 was created to benefit this location, but all of that is IMHO a new feature not really related to this initialization cleanup.

cmd/crio/main.go
mtrmac: Not exactly the same comment, because `--registry` directly overrides the list in `registries.conf`, but `--insecure-registry` adds new CRI-O-implemented items without disabling/overriding the insecure registries/repos configured in `registries.conf`.

Having a `--help` entry that says “this option does $FOO (but another config file also allows $FOO)” seems unexpected to me; it would be reasonable to put in a man page or other primary reference documentation, but it seems a bit much for the hopefully-concise `--help` output. (The `--insecure-registry` entry is already, at least, borderline, in that aspect.)

haircommander: ah makes sense. let's leave it as is then

vrothberg:
Someone else needs to approve :^)

rhatdan:
Ouch needs a rebase.

PR #2506, opened 2019-06-20
Add release bundle target

saschagrunert:
This commit adds a `bundle` Makefile target which can be used to create
a build artifact tarball which contains static binaries and necessary
configurations. The target will be used by CircleCI to provide
continuous results. We can also attach bundles for specific versions on
corresponding GitHub releases.

Example run:
https://8210-161810706-gh.circle-artifacts.com/0/bundle/crio-33a105c0e.tar.gz

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,10 @@ workflows:
       - build
       - build-static
       - build-test-binaries
+      - bundle:
+          requires:
+            - build
+            - build-static
       - clang-format
       - ginkgo
       - git-validation
@@ -75,6 +79,7 @@ workflows:
       - lint
       - results:
           requires:
+            - bundle
             - integration
             - integration-critest
             - integration-static-glibc
@@ -106,6 +111,8 @@ jobs:
           root: .
           paths:
             - bin
+            - crio.conf
+            - docs
 
   build-static:
     executor: nix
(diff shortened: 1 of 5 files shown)


saschagrunert:
I added the release label for 1.15 since I guess it will speed up the deployment of the static binaries.

/cc @mrunalp

saschagrunert:
Rebased on top of the latest master branch to get in the latest CI fixes.

.circleci/config.yml
```diff
@@ -144,6 +151,20 @@ jobs:
...
+      - run:
+          name: bundle
+          command: make bundle
+      - persist_to_workspace:
+          root: .
+          paths:
+            - bundle/*.tar.gz
+
   clang-format:
     executor: docker
```
rhatdan: No change required for this PR, but could we replace docker with podman ?

rhatdan:
LGTM

.circleci/config.yml
saschagrunert: Yeah I thought about using podman there as well. It would mean some slight changes to the CI, but in general it should work (even rootless). I will put it on my TBD list. :)

saschagrunert:
Rebased on top of the latest master branch to resolve the conflicts.

PR #2509, opened 2019-06-20
Don't disable SELinux in e2e tests anymore

haircommander:
By relabeling /var to be used by kubernetes

### DIFF
--- a/contrib/test/integration/e2e-features.yml
+++ b/contrib/test/integration/e2e-features.yml
@@ -19,20 +19,9 @@
             &> {{ artifacts }}/e2e.log
   # Fix vim syntax hilighting: "
 
-- block:
-
-    - name: Disable selinux during e2e tests
-      command: 'setenforce 0'
-      when: not e2e_selinux_enabled
-
-    - name: run e2e tests
-      shell: "{{ e2e_shell_cmd | regex_replace('\\s+', ' ') }}"
-      args:
-        chdir: "{{ ansible_env.GOPATH }}/src/k8s.io/kubernetes"
-      async: '{{ 60 * 60 * 4 }}'  # seconds
-      poll: 60
-
-  always:
-
-    - name: Re-enable SELinux after e2e tests
-      command: 'setenforce 1'
+- name: run e2e tests
+  shell: "{{ e2e_shell_cmd | regex_replace('\\s+', ' ') }}"
+  args:
+    chdir: "{{ ansible_env.GOPATH }}/src/k8s.io/kubernetes"
+  async: '{{ 60 * 60 * 4 }}'  # seconds
+  poll: 60
--- a/contrib/test/integration/e2e.yml
+++ b/contrib/test/integration/e2e.yml
@@ -25,18 +25,9 @@
 
 - block:
 
-    - name: Disable selinux during e2e tests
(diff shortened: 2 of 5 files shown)


contrib/test/integration/e2e-base.yml
```diff
@@ -61,3 +61,7 @@
 
 - name: Enable iptables NAT for the bridge
   command: sysctl -w net.bridge.bridge-nf-call-iptables=1
+
+- name: Relabel /var to be used by kubernetes
+  command: 'semanage fcontext -a -t container_file_t /var'
```
mrunalp: restorecon could be another option. cc: @rhatdan

haircommander: whoops I would have needed restorecon when using semanage. I am actually going to try chcon because I don't think these changes need to persist

contrib/test/integration/e2e-base.yml
```diff
@@ -61,3 +61,19 @@
 
 - name: Enable iptables NAT for the bridge
   command: sysctl -w net.bridge.bridge-nf-call-iptables=1
+
+- name: Change /var label
+  command: 'semanage fcontext -a -R -t container_file_t /var/'
```
rhatdan: If you want to just label everything under /var as container_file_t, for this boot, just do
chcon -t container_file_t -R /var

Don't do semanage or restorecon.

contrib/test/integration/e2e-base.yml
```diff
@@ -61,3 +61,19 @@
...
+  when: e2e_selinux_enabled
+
+- name: Reload /var label
+  command: 'restorecon -v /var/'
+  when: e2e_selinux_enabled
+
+- name: Change /etc label
+  command: 'semanage fcontext -a -R -t container_file_t /etc/'
```
rhatdan: You can do similar for /etc, but I think this is dangerous because other programs that are running on the system might want to read files in /etc like /etc/passwd and they will not be able to read them if /etc/passwd is labeled as container_file_t.

rhatdan: What AVC's are you seeing if you don't do the change of label for /etc?

haircommander: I am not seeing any AVCs anymore, but earlier on e2e_features_* runs I was seeing AVC denials for the name "etc-hosts" which I assumed was /etc/hosts. I then checked openshift nodes for the labelling there, and it seems as though all of /etc is labeled container_file_t, so I tried that.

contrib/test/integration/e2e-base.yml
haircommander: fixed

rhatdan:
@haircommander Still working on this one?

rhatdan:
@haircommander Why is this still around?
@umohnani8 @saschagrunert @haircommander @mrunalp Is this something we want?
Attempting to cleanup old CRI-O PRs

haircommander:
Yeah we eventually want this, but definitely want stable CI first

haircommander:
/hold cancel

we will need this to merge PRs now that we switched to f30

rhatdan:
@haircommander What is going on with this one?

mrunalp:
Closing in favor of #3305

PR #2510, opened 2019-06-20
[1.14] Add global_auth_file option to crio.image config

mtrmac:
> Add a global_auth_file option under crio.image to the crio config. This option points to a json file that holds the credentials required to pull images from a secure registry. The json file is similar to that of .docker/config.json.
>
> Will forward-port to master as well.

This starts with #2498, and:
- removes some dead code
- fixes the case when the CRI client provides credentials (for the primary registry)
- adds use of `globalAuthFile` for the pause image if no specific credentials are provided

This is a functionally minimal set of changes that reflect how the API is probably _intended_ to be used, but not quite exhaustive for how it _could_ be used (e.g. it seems that calling CRI `ImageStatus` with a `docker://example.com/foo:tar` image name would actually try to contact the remote registry!). I do plan to clean all of this up / make it explicit, but that will be a separate `master`-only PR.

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -225,11 +225,15 @@ ctr_stop_timeout = {{ .CtrStopTimeout }}
 # Default transport for pulling images from a remote container storage.
 default_transport = "{{ .DefaultTransport }}"
 
+# The path to a file containing credentials necessary for pulling images from
+# secure registries. The file is similar to that of /var/lib/kubelet/config.json
+global_auth_file = "{{ .GlobalAuthFile }}"
+
 # The image used to instantiate infra containers.
 pause_image = "{{ .PauseImage }}"
 
-# If not empty, the path to a docker/config.json-like file containing credentials
-# necessary for pulling the image specified by pause_image above.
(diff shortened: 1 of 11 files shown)


mrunalp:
@vrothberg ptal

mtrmac:
`master` version: #2510.

mtrmac:
> `master` version: #2510.

Oops, #2511.

vrothberg:

haircommander:
@mtrmac changes LGTM, but just to make sure-if this member is absent from the crio.conf, does CRI-O behave as expected? We need to make sure upgrades using an old crio.conf allow CRI-O to boot up.

TomSweeneyRedHat:
Nothing over the other comments.  Otherwise LGTM

umohnani8:
LGTM
@haircommander yeah it should be fine if this option is missing from crio.conf. We are not failing if value for this option is empty, in fact by default it is going to be empty n cri-o. MCO is the one that will be setting it.

mrunalp:
@saschagrunert We can have a follow-on PR for the flag.

saschagrunert:
Yes. :) no worries about that.

PR #2513, opened 2019-06-21
Add junit test files to .gitignore

saschagrunert:
Failing tests will cause the junit XML files to not being moved into the
test results directory. We now exclude them to avoid commiting them into
the repository.


rhatdan:
LGTM

PR #2514, opened 2019-06-21
pkg/storage test cleanups

mtrmac:
**- What I did**

Significantly rewritten the gomock-based tests in pkg/storage , to hopefully make it plausible to change the code base under test by making the test structure more readable.

**- How I did it**

See the individual commit messages for details.

(Resisted a _very serious_ temptation to just remove the tests; looking at the actual input parameters and control flow, they are, even after this fix, clearly not testing realistic scenarios, ignoring many values on input and returning nonsensical outputs. The computed code coverage is basically only good to confirm that it is possible not to crash on those lines.)

The most important idea is adding an `inOrder` helper, a more useful variant of `gomock.InOrder` that is composable; and based on that, creating a series of `mock$function` helpers that mirror the call stack of the implementation under test. For example, we can have a single `mockNewImage` call that sets up expectations for a series of calls to c/storage, and that can be reasoned about as a single unit.

Yes, this makes the test structure demonstrably very tightly coupled to the implementation (and implementation details of c/image/storage); but it was just as coupled before, only in a less visible way.

**- How to verify it**

Tests should continue to pass.  This does not touch non-test code, except for removing a dead method.

N/A.

### DIFF
--- a/pkg/storage/image_test.go
+++ b/pkg/storage/image_test.go
@@ -21,9 +21,10 @@ import (
 var _ = t.Describe("Image", func() {
 	// Test constants
 	const (
-		testRegistry  = "docker.io"
-		testImageName = "image"
-		testSHA256    = "2a03a6059f21e150ae84b0973863609494aad70f0a80eaeb64bddd8d92465812"
+		testRegistry            = "docker.io"
+		testImageName           = "image"
+		testNormalizedImageName = "docker.io/library/image:latest" // Keep in sync with testImageName!
+		testSHA256              = "2a03a6059f21e150ae84b0973863609494aad70f0a80eaeb64bddd8d92465812"
 	)
 
 	var (
@@ -42,7 +43,7 @@ var _ = t.Describe("Image", func() {
 
 		var err error
 		sut, err = storage.GetImageService(
-			context.Background(), nil, storeMock, "",
+			context.Background(), nil, storeMock, "docker://",
 			[]string{}, []string{testRegistry},
 		)
 		Expect(err).To(BeNil())
@@ -54,26 +55,25 @@ var _ = t.Describe("Image", func() {
 
 	mockGetRef := func() mockSequence {
 		return inOrder(
-			storeMock.EXPECT().Image(gomock.Any()).
-				Return(&cs.Image{ID: testImageName}, nil),
-			mockParseStoreReference(storeMock, ""), // FIXME: What image name is expected?
+			// parseStoreReference ("@"+testImageName) will fail, recognizing it as an invalid image ID
+			storeMock.EXPECT().Image(testImageName).
+				Return(&cs.Image{ID: testSHA256}, nil),
+
+			mockParseStoreReference(storeMock, testImageName),
 		)
 	}
 
-	mockListImage := func() mockSequence {
+	// Only about the first half of imageService.buidlImageCacheItem
+	mockBuildImageCacheItemStart := func(expectedImageNameOrID string) mockSequence {
 		return inOrder(
-			storeMock.EXPECT().Image(gomock.Any()).
-				Return(&cs.Image{ID: testImageName}, nil),
-			storeMock.EXPECT().ListImageBigData(gomock.Any()).
-				Return([]string{""}, nil),
-			storeMock.EXPECT().ImageBigDataSize(gomock.Any(), gomock.Any()).
-				Return(int64(0), nil),
-			storeMock.EXPECT().ImageBigData(gomock.Any(), gomock.Any()).
+			// NewImageSource:
+			mockResolveImage(storeMock, expectedImageNameOrID, testSHA256),
+			// imageSize:
+			mockStorageImageSourceGetSize(storeMock),
+			// imageConfigDigest calling storageImageSource.GetManifest:
+			storeMock.EXPECT().ImageBigData(testSHA256, gomock.Any()).
 				Return(testManifest, nil),
-			storeMock.EXPECT().Image(gomock.Any()).
-				Return(&cs.Image{ID: testImageName}, nil),
 		)
-
 	}
 
 	t.Describe("GetImageService", func() {
@@ -264,11 +264,9 @@ var _ = t.Describe("Image", func() {
 			// Given
 			inOrder(
 				mockGetRef(),
-				storeMock.EXPECT().Image(gomock.Any()).
-					Return(&cs.Image{ID: testImageName}, nil),
-				storeMock.EXPECT().Image(gomock.Any()).
-					Return(&cs.Image{ID: testImageName}, nil),
-				storeMock.EXPECT().DeleteImage(gomock.Any(), gomock.Any()).
+				mockGetStoreImage(storeMock, testNormalizedImageName, testSHA256),
+				mockResolveImage(storeMock, testNormalizedImageName, testSHA256),
+				storeMock.EXPECT().DeleteImage(testSHA256, true).
 					Return(nil, nil),
 			)
 
@@ -288,11 +286,11 @@ var _ = t.Describe("Image", func() {
 			Expect(err).NotTo(BeNil())
 		})
 
-		It("should fail to untag an image with invalid name", func() {
+		It("should fail to untag an image that can't be found", func() {
 			// Given
 			inOrder(
 				mockGetRef(),
-				storeMock.EXPECT().Image(gomock.Any()).Return(nil, t.TestError),
+				mockGetStoreImage(storeMock, testNormalizedImageName, ""),
 			)
 
 			// When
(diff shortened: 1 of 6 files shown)


mtrmac:
@saschagrunert @vrothberg RFC. While working on the `GlobalAuthFile` feature I have noticed quite a few opportunities for cleanup in uses of c/image, but I was completely unable to implement most of them because I just wasn’t able to correlate the code under test with the mocks series of expected calls. (It turns out that some of the existing helpers started or ended in the middle of function calls in the tested code, or in the middle of a deeper call stack called by the tested code, even.)

pkg/storage/mock_helpers_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+)
+
+type mockSequence struct {
+	first, last *gomock.Call // may be both nil (= the default value of mockSequence) to mean empty sequence
+}
+
+// like gomock.inOrder, but can be nested
+func inOrder(calls ...interface{}) mockSequence {
```
saschagrunert: This might be a valuable addition to gomock as well, do you think we should port it there?

pkg/storage/image_test.go
```diff
@@ -52,18 +52,12 @@ var _ = t.Describe("Image", func() {
...
-		)
-	}
-
-	mockGetRef := func() {
-		gomock.InOrder(
+	mockGetRef := func() mockSequence {
+		return inOrder(
 			storeMock.EXPECT().Image(gomock.Any()).
 				Return(&cs.Image{ID: testImageName}, nil),
+			mockParseStoreReference(storeMock, ""), // FIXME: What image name is expected?
```
saschagrunert: I would remove that `FIXME` since I guess the image name does not matter at all for these kind of tests, right?

pkg/storage/mock_helpers_test.go
```diff
@@ -38,3 +39,13 @@ func inOrder(calls ...interface{}) mockSequence {
 	}
 	return mockSequence{first, last}
 }
+
+// containers/image/storage.storageReference.StringWithinTransport
+func mockStorageReferenceStringWithinTransport(storeMock *containerstoragemock.MockStore) mockSequence {
+	return inOrder(
+		storeMock.EXPECT().GraphOptions().Return([]string{}),
```
saschagrunert: This is just an idea and not something we need to do right now:

Do you think it would be worth adding a mock'able public API interface to containers/{image,storage}? I think we could then avoid these deeply nested mocks a bit.

pkg/storage/mock_helpers_test.go
```diff
@@ -0,0 +1,130 @@
+package storage_test
```
saschagrunert: Do we want to add these helpers to the `test/framework/framework.go` to be re-usable accross all test suites?

pkg/storage/image_test.go
mtrmac: The name does matter now, see how `mockParseStoreReference` uses it.

Note that this `FIXME` does not remain at the end of this PR, the commit “Fix mockGetRef, and deal with all of the fallout” removes it.

(I do generally try to make my PRs close to a sequence of one-step-at-a-time commits with nothing left to fix up later, in this case it was a bit tricky because `mockNewImage` needs to be created and use the new stricter `mockParseStoreReference` for the “Fix mockGetRef, and deal with all of the fallout” commit to be possible.

OTOH, it _is_ possible not to have that FIXME even temporarily, by temporarily introducing two variants of `mockParseStoreReference` in the worst case, or maybe by taking more care to split “Fix mockGetRef, and deal with all of the fallout” into individual precisely-reasoned steps, that commit is rather a collection of whatever remained outstanding. Let me know if I should work on that.

pkg/storage/mock_helpers_test.go
mtrmac: Hum… I did spend some time looking in GoMock for various helpers, including ultimately something like this, so maybe it would be natural to propose.

OTOH (and this is with a huge disclaimer: this is the first project I’ve encountered with so much state to mock, and significant use of mocks in general, and I have not tried testing it in any other way at all, so I am really not justified to comment on the trade-offs or suitability in any way), it seems to me that once the use of mocking gets so widespread like here, that a several-level call stack of setting up mocks turns out to be necessary to make the mocks readable, it might be an indication that things have gone too far down a difficult road, and the tests might need reconsidering / restructuring in some other way (again, disclaimer: I have no idea what that would be) instead of making it any easier to go down that path.

(From what I’ve been able to gather, the GoMock implementation is not set up for enforcing a strict _order_ of calls, it seems to be more structured about ~stateless pairs of input→output that could be called in any order.  Which is one reason why test failures about the code call sequence not matching the mock sequence are so convoluted and difficult to read, they almost always refer to some other distant mocked call of the same method with different arguments and only secondarily mention the ordering constraints.)

pkg/storage/mock_helpers_test.go
mtrmac: Mockable in what way?  (See the disclaimer elsewhere: I don’t quite know what I am doing at this conceptual level.)

c/image does have a set of `types.Image{Transport,Reference,Source,Destination}` interfaces, and (although that’s a bit unusual for c/image), the `storage.Transport` even allows replacing the underlying `c/storage.Store` object), so it should be quite possible to write tests that ensure that c/image is invoked correctly (and that exercise the `c/image/copy.Image` function in full, perhaps), without actually having to execute the full implementation of the `c/image/storage.storageTransport` object: basically, somewhere in CRI-O, parametrize whatever creates `types.ImageReference` objects to return an `ImageReference` object that points to a test-only mock transport purpose-built for the test, instead of a real one.  For the most part, that would mean wrapping `alltransports.ParseImageName` (and probably `c/image/storage.Transport.ParseStoreReference`, and maybe a few others) in an interface that would allow CRI-O tests to return a mock instead of the real thing.

AFAICS this would be something that CRI-O needs to do; as long as CRI-O calls the `alltransports.ParseImageName` global directly, the return value of that can’t be mocked (well, without modifying `ParseImageName` to refer to global state, at least, which would be fairly ugly).

pkg/storage/mock_helpers_test.go
mtrmac: I don’t know… as mentioned elsewhere, (with huge disclaimers) the necessity of this seems to be a code smell to me. From a quick look, it seems that many other packages have the mocked call sequences smaller, so they don’t need this as urgently.

The composable `inOrder` helper does seem possibly useful in general, or at least it does not clearly belong into `pkg/storage`. If any other test package needs the `storeMock`-based helper abstractions, although it accesses `containers/storage` via the `pkg/storage` interface, maybe that test package should instead use a `pkg/storage` mock instead of a `containers/storage` mock with the `storeMock`-based helpers. (Then again, end-to-end testing can genuinely be useful.)

(Also, practically, right now most of CRI-O does not build on macOS and at least this week I can’t spend much time changing that nor changing my development setup, so it is cumbersome for me to modify tests in most of the other packages. And I originally wanted to clean up c/image callers instead of redesigning tests.)

mtrmac:
> Beside this, would you consider making the first commit message (first paragraph) a bit more formal? 😇

Done, I think. I’m sorry, I’ve only seen that word used and never thought to look it up in a dictionary. Please give me a direct pointer if I misunderstood.

saschagrunert:
I'm fine with these changes, LGTM :+1: 

These linting issues have to be fixed before continuing here:
https://circleci.com/gh/cri-o/cri-o/1800?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

pkg/storage/mock_helpers_test.go
saschagrunert: > Mockable in what way? (See the disclaimer elsewhere: I don’t quite know what I am doing at this conceptual level.)

In an ideal world, which we'll probably never have, we would unit-test only the code part which contains to a dedicated function. Everything else could be mocked, like IO side effects. The issue is now that we can only mock public interfaces in golang, which is not available in every case (like for reading and writing of files). This means that we have to go down the path which is out of scope of the function and this makes it hard to understand/write. In terms of unit testing, it would be great to consume an API which provides an public mockable interface at its outermost endpoints.

vrothberg:
Only one step away from merging: the linter [is not yet happy](https://circleci.com/gh/cri-o/cri-o/1800?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

Note that I only skimmed the tests as reading `gomock` is close to physical pain for me; it's too far from idiomatic go and the benefit of using it seems IMHO to be overruled by the decreased maintainability.

Besides my short rant: LGTM :+1:

PR #2519, opened 2019-06-21
[1.14] Add state of infracontainer to disk when stopped

umohnani8:
Add the state of the infracontainer when the pod is stopped.
This ensures that the pod is removed in a timely manner after
being stopped.


umohnani8:
@haircommander @mrunalp PTAL

haircommander:
LGTM

PR #2520, opened 2019-06-21
Add state of infracontainer to disk when stopped

umohnani8:
Add the state of the infracontainer when the pod is stopped.
This ensures that the pod is removed in a timely manner after
being stopped.

### DIFF
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -135,6 +135,9 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 	if err := s.StorageRuntimeServer().StopContainer(sb.ID()); err != nil && errors.Cause(err) != storage.ErrContainerUnknown {
 		logrus.Warnf("failed to stop sandbox container in pod sandbox %s: %v", sb.ID(), err)
 	}
+	if err := s.ContainerStateToDisk(podInfraContainer); err != nil {
+		logrus.Warnf("error writing pod infra container %q state to disk: %v", podInfraContainer.ID(), err)
+	}
 
 	logrus.Infof("Stopped pod sandbox: %s", podInfraContainer.Description())
 	sb.SetStopped()
--- a/server/sandbox_stop_test.go
+++ b/server/sandbox_stop_test.go
@@ -38,6 +38,8 @@ var _ = t.Describe("PodSandboxStatus", func() {
 					gomock.Any()).Return(nil),
 				runtimeServerMock.EXPECT().StopContainer(gomock.Any()).
 					Return(nil),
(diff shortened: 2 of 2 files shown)


umohnani8:
@haircommander @mrunalp PTAL

server/sandbox_stop_linux.go
```diff
@@ -135,6 +135,7 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 	if err := s.StorageRuntimeServer().StopContainer(sb.ID()); err != nil && errors.Cause(err) != storage.ErrContainerUnknown {
 		logrus.Warnf("failed to stop sandbox container in pod sandbox %s: %v", sb.ID(), err)
 	}
+	s.ContainerStateToDisk(podInfraContainer)
```
saschagrunert: The errcheck linter will complain that we do not check this error. We should print a warning if we don't want to fail here.

saschagrunert: Beside this, we have to add the following mock calls to `sandbox_stop_test.go` (line 40 and 67)

```
				ociRuntimeMock.EXPECT().UpdateContainerStatus(gomock.Any()).
					Return(nil),
```

umohnani8: Fixed, thanks!

server/sandbox_stop_test.go
```diff
@@ -36,6 +36,8 @@ var _ = t.Describe("PodSandboxStatus", func() {
 					gomock.Any(), gomock.Any()).Return(nil),
 				ociRuntimeMock.EXPECT().WaitContainerStateStopped(gomock.Any(),
 					gomock.Any()).Return(nil),
+				ociRuntimeMock.EXPECT().UpdateContainerStatus(gomock.Any()).
+					Return(nil),
```
saschagrunert: I guess this insertion has to be moved after the 
```
runtimeServerMock.EXPECT().StopContainer(gomock.Any()).
					Return(nil),
```
below. :point_down:

server/sandbox_stop_test.go
```diff
@@ -63,6 +65,8 @@ var _ = t.Describe("PodSandboxStatus", func() {
 					gomock.Any(), gomock.Any()).Return(nil),
 				ociRuntimeMock.EXPECT().WaitContainerStateStopped(gomock.Any(),
 					gomock.Any()).Return(nil),
+				ociRuntimeMock.EXPECT().UpdateContainerStatus(gomock.Any()).
```
saschagrunert: The same for this one.

server/sandbox_stop_test.go
umohnani8: yup, done

server/sandbox_stop_test.go
umohnani8: fixed

PR #2523, opened 2019-06-24
Fallback to default seccomp profile if not provided

saschagrunert:
We now include the default seccomp profile within the binary to provide
stronger configuration defaults. This profile will be used as fallback
if the user does not specify an own one.

Since we now use `github.com/seccomp/container-golang` as direct
dependency we can remove the internal implementation.

The seccomp related integration tests have been enabled in CircleCI,
which now work as well there.


vrothberg:
I like the idea of having a default one in memory but I feel rather strongly against having that in `cri-o/pkg`, mostly because it increases the fragmentation across the projects. As @saschagrunert mentions above, the profiles diverged again, so I'd prefer having the profile and the go package at one central place. One thing when doing that is to be very verbose about moving a the profile to another location/project to avoid breaking downstream packages.

saschagrunert:
Hm, what about moving it into `github.com/seccomp/libseccomp-golang`?

vrothberg:
> Hm, what about moving it into `github.com/seccomp/libseccomp-golang`?

@rhatdan WDYT?

rhatdan:
I thought we already did something with this in:

https://github.com/seccomp/containers-golang

vrothberg:
Perfect! @saschagrunert, are you cool with using the defaults from https://github.com/seccomp/containers-golang?

vrothberg:
Maybe, those need to be updated but I am not sure.

saschagrunert:
> Perfect! @saschagrunert, are you cool with using the defaults from https://github.com/seccomp/containers-golang?

I think we should go for a complete substitution of the internal package with the `containers-golang`. This will invoke more changes in the related repo so I will put this on hold for now.

saschagrunert:
This is now ready for review, the description and scope changed a bit.

@rhatdan @vrothberg @haircommander PTAL

server/server.go
```diff
@@ -353,15 +353,22 @@ func New(
...
-		var seccompConfig seccomp.Seccomp
-		if jsonErr := json.Unmarshal(seccompProfile, &seccompConfig); jsonErr != nil {
-			return nil, fmt.Errorf("decoding seccomp profile failed: %v", jsonErr)
+		if config.SeccompProfile != "" {
+			seccompProfile, fileErr := ioutil.ReadFile(config.SeccompProfile)
+			if fileErr != nil {
+				return nil, fmt.Errorf("opening seccomp profile (%s) failed: %v",
+					config.SeccompProfile, fileErr)
+			}
+			logrus.Infof("Using seccomp profile %q", config.SeccompProfile)
```
vrothberg: I'd move that after the `json.Unmarshal(...)` to only log it when we're sure loading has succeeded.

server/server.go
```diff
@@ -353,15 +353,22 @@ func New(
...
+			var seccompConfig seccomp.Seccomp
+			if jsonErr := json.Unmarshal(seccompProfile, &seccompConfig); jsonErr != nil {
+				return nil, fmt.Errorf("decoding seccomp profile failed: %v", jsonErr)
+			}
+			s.seccompProfile = &seccompConfig
+		} else {
+			logrus.Infof("No seccomp profile specified, using the internal default.")
```
vrothberg: Nit of a nit: remove the trailing dot for consistency

saschagrunert:
> @saschagrunert, did you verify that the seccomp profiles are identical?

Yes, but they are not completely equal:
```
> diff old.json new.json
```
```diff
308a309
>         "statx",
333c334
<         "umount2",
---
>         "umount",
337d337
<         "unshare",
589a590,599
>       },
>       "excludes": {}
>     },
>     {
>       "names": ["get_mempolicy", "mbind", "name_to_handle_at", "set_mempolicy"],
>       "action": "SCMP_ACT_ALLOW",
>       "args": [],
>       "comment": "",
>       "includes": {
>         "caps": ["CAP_SYS_NICE"]
```

server/container_create.go
```diff
@@ -648,7 +648,12 @@ func (s *Server) setupSeccomp(specgen *generate.Generator, profile string) error
...
 	if profile == seccompRuntimeDefault || profile == seccompDockerDefault {
-		return seccomp.LoadProfileFromStruct(&s.seccompProfile, specgen)
+		linuxSpecs, err := seccomp.LoadProfileFromConfig(s.seccompProfile, specgen.Config)
+		if err != nil {
+			return err
+		}
+		specgen.Config.Linux.Seccomp = linuxSpecs
+		return nil
 	}
 	if !strings.HasPrefix(profile, seccompLocalhostPrefix) {
```
vrothberg: The TrimPrefix below should also use seccompLocalhostPrefix.

server/container_create.go
```diff
@@ -648,7 +648,12 @@ func (s *Server) setupSeccomp(specgen *generate.Generator, profile string) error
 		return nil
 	}
 	if profile == seccompRuntimeDefault || profile == seccompDockerDefault {
```
vrothberg: Can we add some comments starting from here to this method? The code is not entirely obvious.

vrothberg:
The commit message should mention/scream that the `seccomp.json` is removed. Downstream packaging may depend on it and we should not miss it in the changelog. Maybe we can create a dedicated commit for this removal to make it super obvious?

vrothberg:
> > @saschagrunert, did you verify that the seccomp profiles are identical?
> 
> Yes, but they are not completely equal:
> 
> ```
> > diff old.json new.json
> ```
> 
> ```diff
> 308a309
> >         "statx",
> 333c334
> <         "umount2",
> ---
> >         "umount",
> 337d337
> <         "unshare",
> 589a590,599
> >       },
> >       "excludes": {}
> >     },
> >     {
> >       "names": ["get_mempolicy", "mbind", "name_to_handle_at", "set_mempolicy"],
> >       "action": "SCMP_ACT_ALLOW",
> >       "args": [],
> >       "comment": "",
> >       "includes": {
> >         "caps": ["CAP_SYS_NICE"]
> ```

Which files were in this diff? When I do a `$ diff -u $profileFromCRI-OMaster $profileFrom/seccomp/containers/golang` I get:
```diff
@@ -322,6 +322,7 @@
                                "stat64",
                                "statfs",
                                "statfs64",
+                               "statx",
                                "symlink",
                                "symlinkat",
                                "sync",
@@ -674,6 +675,23 @@
                                ]
                        },
                        "excludes": {}
+               },
+               {
+                       "names": [
+                               "get_mempolicy",
+                               "mbind",
+                               "name_to_handle_at",
+                               "set_mempolicy"
+                       ],
+                       "action": "SCMP_ACT_ALLOW",
+                       "args": [],
+                       "comment": "",
+                       "includes": {
+                               "caps": [
+                                       "CAP_SYS_NICE"
+                               ]
+                       },
+                       "excludes": {}
                },
                {
                        "names": [
```

saschagrunert:
> Which files were in this diff?

Yeah it was the CRI-O master `seccomp.json` and the generated `default.json` (from the struct) when running `make` in containers-golang.

server/server.go
saschagrunert: Adapted as suggested.

server/server.go
saschagrunert: Affirmative.

server/container_create.go
saschagrunert: Yes, changed as well.

server/container_create.go
saschagrunert: Did this to make the intention more clear.

saschagrunert:
> The commit message should mention/scream that the `seccomp.json` is removed. Downstream packaging may depend on it and we should not miss it in the changelog. Maybe we can create a dedicated commit for this removal to make it super obvious?

I made the removal of `seccomp.json` the head-liner of the commit message, so I hope an additional commit split-up is not necessary.

vrothberg:
> > Which files were in this diff?
> 
> Yeah it was the CRI-O master `seccomp.json` and the generated `default.json` (from the struct) when running `make` in containers-golang.

So the `seccomp.json` from seccomp/containers-golang is inherently different from the generated `default.json` :thinking: 
```diff
--- seccomp.json        2019-06-27 11:06:55.765572987 +0200
+++ default.json        2019-06-27 11:10:15.855699048 +0200
@@ -194,6 +194,7 @@
                                "mlockall",
                                "mmap",
                                "mmap2",
+                               "mount",
                                "mprotect",
                                "mq_getsetattr",
                                "mq_notify",
@@ -210,6 +211,7 @@
                                "munlock",
                                "munlockall",
                                "munmap",
+                               "name_to_handle_at",
                                "nanosleep",
                                "newfstatat",
                                "_newselect",
@@ -234,6 +236,7 @@
                                "readlink",
                                "readlinkat",
                                "readv",
+                               "reboot",
                                "recv",
                                "recvfrom",
                                "recvmmsg",
@@ -347,6 +350,7 @@
                                "truncate64",
                                "ugetrlimit",
                                "umask",
+                               "umount",
                                "uname",
                                "unlink",
                                "unlinkat",
@@ -359,12 +363,7 @@
                                "waitid",
                                "waitpid",
                                "write",
-                               "writev",
-                               "mount",
-                               "umount2",
-                               "reboot",
-                               "name_to_handle_at",
-                               "unshare"
+                               "writev"
                        ],
                        "action": "SCMP_ACT_ALLOW",
                        "args": [],
@@ -770,4 +769,4 @@
                        "excludes": {}
                }
        ]
-}
+}
```

I am a bit confused now since it's not clear to me what we or a user should use, see:
```
SECCOMP_PROFILE=${SECCOMP_PROFILE:-${CRIO_ROOT}/cri-o/vendor/github.com/seccomp/containers-golang/seccomp.json}
```

The default one is obviously more secure and would never grant a container the permission to reboot but I guess we need to fix that in the seccomp project. @rhatdan @mheon PTAL

saschagrunert:
> So the `seccomp.json` from seccomp/containers-golang is inherently different from the generated `default.json`

I sorted the arrays to have a cleaner diff. :wink:

vrothberg:
> > So the `seccomp.json` from seccomp/containers-golang is inherently different from the generated `default.json`
> 
> I sorted the arrays to have a cleaner diff. wink

:laughing: selective perception at it's best. Thanks for calming me down so quickly :) So it seems that umount and umount2 are left.

saschagrunert:
> So it seems that umount and umount2 are left.

`statx` vs `unshare` too :thinking:

docs/crio.8.md
```diff
@@ -177,7 +177,7 @@ If `hooks_dir` is unset, CRI-O will currently default to `/usr/share/containers/
 
 **--selinux**=**true**|**false**: Enable selinux support (default: false)
 
-**--seccomp-profile**="": Path to the seccomp json profile to be used as the runtime's default (default: "/etc/crio/seccomp.json")
+**--seccomp-profile**="": Path to the seccomp.json profile to be used as the runtime's default. If not specified, then the internal default seccomp profile will be used. (default: "")
```
TomSweeneyRedHat: is the "(default: "") still necessary here?  We don't have it in crio.conf.5.md below.  I'd vote to remove it.

TomSweeneyRedHat:
One possible nit comment.  As everything is green and this is a big PR, I'm fine ignoring it or circling back and tweaking it in a follow up PR if necessary.
LGTM otherwise.

docs/crio.8.md
saschagrunert: Okay, I removed it as suggested. :+1:

rhatdan:
We should update the seccomp files with the additional rules. Most of them have to do with systemd inside of a container.

saschagrunert:
> We should update the seccomp files with the additional rules. Most of them have to do with systemd inside of a container.

Opened https://github.com/seccomp/containers-golang/pull/21 to update the default profile

saschagrunert:
Done, updated to https://github.com/seccomp/containers-golang/releases/tag/v0.3.1

saschagrunert:

PR #2524, opened 2019-06-24
Search conmon executable in $PATH

saschagrunert:
If conmon in the config is empty or not available, we now search it
inside the $PATH variable before actually failing.

We could do the same for the runtimes, WDYT?

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -89,7 +89,7 @@ crio [GLOBAL OPTIONS] config [OPTIONS]
 
 **--config**="": path to configuration file
 
-**--conmon**="": path to the conmon executable (default: "/usr/local/libexec/crio/conmon")
+**--conmon**="": Path to the conmon binary, used for monitoring the OCI runtime. Will be searched for using $PATH if empty. (default: "")
 
 **--cpu-profile**="": set the CPU profile file path
 
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -80,8 +80,8 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 **no_pivot**=*false*
   If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
 
-**conmon**="/usr/local/libexec/crio/conmon"
-  Path to the conmon binary, used for monitoring the OCI runtime.
+**conmon**=""
+  Path to the conmon binary, used for monitoring the OCI runtime. Will be searched for using $PATH if empty.
 
 **conmon_env**=["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"]
   Environment variable list for the conmon process, used for passing necessary environment variables to conmon or the runtime.
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -402,7 +402,7 @@ func DefaultConfig() (*Config, error) {
 					RuntimeRoot: DefaultRuntimeRoot,
 				},
 			},
-			Conmon: conmonPath,
+			Conmon: "",
 			ConmonEnv: []string{
 				"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
 			},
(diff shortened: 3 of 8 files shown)


docs/crio.8.md
```diff
@@ -89,7 +89,7 @@ crio [GLOBAL OPTIONS] config [OPTIONS]
 
 **--config**="": path to configuration file
 
-**--conmon**="": path to the conmon executable (default: "/usr/local/libexec/crio/conmon")
+**--conmon**="": Path to the conmon binary, used for monitoring the OCI runtime. Will be searched inside $PATH if empty. (default: "")
```
TomSweeneyRedHat: suggest: "searched inside" to "searched for using"

docs/crio.conf.5.md
```diff
@@ -81,7 +81,7 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
   If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
 
 **conmon**="/usr/local/libexec/crio/conmon"
-  Path to the conmon binary, used for monitoring the OCI runtime.
+  Path to the conmon binary, used for monitoring the OCI runtime. Will be searched inside $PATH if empty.
```
TomSweeneyRedHat: if you take suggestion above, use here too please.

TomSweeneyRedHat:
Tests aren't hip.  Couple of wording suggestions for consideration.

docs/crio.8.md
saschagrunert: Yes, I changed it as suggested. :+1:

docs/crio.conf.5.md
saschagrunert: Yes, I changed it as suggested. :+1:

saschagrunert:
> Tests aren't hip. Couple of wording suggestions for consideration.

I was finally able to fix the tests. The issue was the new default value `""` for conmon, so I had to adapt the installation path for conmon during `make install` to be inside $PATH.

haircommander:
Why do you default to "" ? Did CI break when conmon was left as conmonPath, and the makefile was left the same. I feel like I'd like crio to look at `libexec` first because of conmon version skew between it and podman

saschagrunert:
> Why do you default to "" ? 

My main intention was to have a working default. If we default to something in `libexec` then this default value might not work.

> I feel like I'd like crio to look at `libexec` first because of conmon version skew between it and podman

Yeah somehow I agree. But podman does not install into `$(PREFIX)/bin` as well, right?

haircommander:
Podman doesn't actually install conmon. It looks in a couple of predefined places, of which one is `$PREFIX/bin` (though another is `$PREFIX/libexec/crio` so maybe my point is moot...

saschagrunert:
> Podman doesn't actually install conmon. It looks in a couple of predefined places, of which one is `$PREFIX/bin` (though another is `$PREFIX/libexec/crio` so maybe my point is moot...

I'm unsure, my main intention for this change was to provide a strong working default for CRI-O, like when running it on NixOS or just spinning up the static binary. If we still default to some thing like `.../libexec/...` which is outside of the users $PATH than the PR does not solve the issue at all. :man_shrugging:

rhatdan:
I would prefer that we run it from /usr/bin, and then fall back to /usr/libexec/crio/ when we have version skue.

saschagrunert:
Rebased on top of the latest master branch to resolve the conflicts.

haircommander:
Okay, I've thought about this more and I am okay with it. I was initially worried because some of my work didn't seem to be able to be backwards compatible, but it now is. Version skew is less of a concern, and thus LGTM.
@vrothberg @umohnani8 PTAL

lib/config/config.go
```diff
@@ -560,14 +561,32 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
...
+	if c.Conmon == "" {
+		conmon, err := exec.LookPath(executable)
+		if err != nil {
+			return errors.Wrapf(err, "%q not found in $PATH", executable)
+		}
+		c.Conmon = conmon
+		logrus.Debugf("using conmon from $PATH %q", conmon)
+	} else if _, err := os.Stat(c.Conmon); err != nil {
+		return errors.Wrapf(err, "invalid conmon path")
+	}
```
vrothberg: I'd prefer to always to a `logrus.Infof(...)` here since forgetting to set conmon explicitly could be a source of bugs/undesired behaviours etc.

vrothberg: To be more precise: I'd always log the path of the conmon we're using, independent where it came from.

saschagrunert: Added a log message in any succeeding case.

saschagrunert:
I had to rebase on top of the latest master branch and lost my `lgtm`. @umohnani8 may I ask you to drop it again here? :innocent:

saschagrunert:
It seems that we broke a single unit test on the master branch with the last merge, I will follow up with that.

saschagrunert:

PR #2527, opened 2019-06-24
[1.14] Fix up machine os content version and cri-o version in spec

umohnani8:
We fail if the cri-o version for 1.14 does not match a certain
pattern.
Use the 4.2 machine os content image for the ci.

### DIFF
--- a/contrib/test/ci/cri-o.spec
+++ b/contrib/test/ci/cri-o.spec
@@ -32,7 +32,7 @@
 %global service_name crio
 
 Name: %{repo}
-Version: 1.14
+Version: 1.14.0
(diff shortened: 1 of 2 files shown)


umohnani8:
@haircommander @mrunalp PTAL

haircommander:
LGTM

contrib/test/ci/cri-o.spec
```diff
@@ -32,7 +32,7 @@
 %global service_name crio
 
 Name: %{repo}
-Version: 1.14
+Version: 1.14.0
```
haircommander: not blocking but we should eventually put a note here why we need this .0

PR #2533, opened 2019-06-25
Only enable circleCI on master

haircommander:
Until we add support on other branches
(I'm not sure this will actually fix the problems in the other branches, because they have no config at all...)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,9 @@ workflows:
 
 jobs:
   build:
+    branches:
+      only:
+        - master
     executor: docker
     steps:
       - checkout


.circleci/config.yml
```diff
@@ -93,6 +93,9 @@ workflows:
 
 jobs:
   build:
+    branches:
+      only:
+        - master
```
saschagrunert: I don't think that this works. :)

The main issue is that CircleCI listens on webhooks incoming from GitHub and if there is no config available on the branch it simply fails (this is awkward, but okay).

haircommander:
@saschagrunert yeah I wasn't sure it would but figured I'd try. Your solutions are better :)

PR #2539, opened 2019-06-26
Use repository logo instead of rawgit

saschagrunert:
RawGit has reached the end of life, so we should consume the logo directly from the repository instead of there.


TomSweeneyRedHat:
LGTM and resolves well.

rhatdan:
Done wasting electricity on this one.  Force Merging.

PR #2540, opened 2019-06-26
WIP Conmon exec legacy smoketest

haircommander:
I've updated the conmon exec changes to be backwards compatible. This is a smoketest PR to test to see if that is the case.


haircommander:
nice

PR #2541, opened 2019-06-27
Discover runtimePath from $PATH environment

saschagrunert:
The default runtime path for `runc` is now `""` and we discover the
executable directly from the local $PATH environment. Is it nevertheless
possible to set the `runtime_path` directly, where we still test for
existence on runtime.

The validation of the runtime handlers within `oci.go` has been cleaned
up, to bring back the matter of concerns to `config.go`. We now use
pointers to easily update the runtime handlers during validation, where
a new type `Runtimes` got introduced.

All tests have been adapted as well.

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/BurntSushi/toml"
@@ -27,7 +28,6 @@ const (
 	defaultTransport    = "docker://"
 	apparmorProfileName = "crio-default"
 	defaultRuntime      = "runc"
-	defaultRuntimePath  = "/usr/bin/runc"
 	DefaultRuntimeType  = "oci"
 	DefaultRuntimeRoot  = "/run/runc"
 	cgroupManager       = "cgroupfs"
@@ -146,6 +146,9 @@ type RuntimeHandler struct {
 	RuntimeRoot string `toml:"runtime_root"`
 }
 
+// Multiple runtime Handlers in a map
+type Runtimes map[string]*RuntimeHandler
+
 // RuntimeConfig represents the "crio.runtime" TOML config table.
 type RuntimeConfig struct {
 	// ConmonEnv is the environment variable list for conmon process.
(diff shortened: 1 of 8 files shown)


TomSweeneyRedHat:
integration_fedora tests aren't happy.

lib/config/config.go
```diff
@@ -518,12 +521,16 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
...
+					RuntimePath: "",
+					RuntimeType: DefaultRuntimeType,
+					RuntimeRoot: DefaultRuntimeRoot,
+				}
 			}
 			// Set the DefaultRuntime to runc so we don't fail further along in the code
 			c.DefaultRuntime = defaultRuntime
 		} else {
-			return fmt.Errorf("default_runtime set to %q, but no runtime path is set for it", c.DefaultRuntime)
+			return fmt.Errorf("default_runtime set to %q, but no runtime entry found for it", c.DefaultRuntime)
```
TomSweeneyRedHat: nit "entry was found"

oci/oci.go
```diff
@@ -81,40 +81,32 @@ type RuntimeImpl interface {
...
 
 	runtimeHandler, ok := r.config.Runtimes[handler]
 	if !ok {
-		return config.RuntimeHandler{}, fmt.Errorf("failed to find runtime handler %s from runtime list %v",
+		return nil, fmt.Errorf("failed to find runtime handler %s from runtime list %v",
 			handler, r.config.Runtimes)
 	}
 	if runtimeHandler.RuntimePath == "" {
-		return config.RuntimeHandler{}, fmt.Errorf("empty runtime path for runtime handler %s", handler)
+		return nil, fmt.Errorf("empty runtime path for runtime handler %s", handler)
```
TomSweeneyRedHat: I don't see any changes to calls to ValidateRuntimeHandler() .  Will they all be happy getting a nil return?  i.e. they're not catching the error and trying to run with the default (now nil) handler.

lib/config/config.go
saschagrunert: Fixed :)

oci/oci.go
saschagrunert: I see only two uses of that function, where the error handling is fine in both cases:

https://github.com/cri-o/cri-o/blob/e2bd3aaebc7950ca954adbb7cfb708e33e242281/server/sandbox_run.go#L66-L68

https://github.com/cri-o/cri-o/blob/e2bd3aaebc7950ca954adbb7cfb708e33e242281/oci/oci.go#L194-L197

TomSweeneyRedHat: thx for checking @saschagrunert

TomSweeneyRedHat:
LGTM assuming the pesky fedora test will pass.

rhatdan:
@vrothberg @umohnani8 PTAL

umohnani8:
LGTM @haircommander PTAL

PR #2542, opened 2019-06-27
Remove duplicate Dockerfile's

saschagrunert:
With the move to CircleCI we do not need the second Dockerfile any more
and merge these into one. We now versionize these files to avoid clashes
during updates.


saschagrunert:
Just have to make sure that `make dbuild`, `make integration`, `sudo make dbuild` and `sudo make integration` still works...

saschagrunert:
Looks good now, PTAL

Makefile
```diff
@@ -175,10 +175,23 @@ crioimage: git-vars
...
+
+integration: ${GINKGO} crioimage
+	sudo $(CONTAINER_RUNTIME) run \
+		-e CI=true \
+		-e CRIO_BINARY \
+		-e RUN_CRITEST \
+		-e STORAGE_OPTIONS="--storage-driver=vfs" \
+		-e TESTFLAGS \
+		-e TEST_USERNS \
+		-it --privileged --rm \
```
vrothberg: Not sure if we want/need `-i`?

Makefile
```diff
@@ -175,10 +175,23 @@ crioimage: git-vars
 	$(CONTAINER_RUNTIME) build -t ${CRIO_IMAGE} .
 
 dbuild: crioimage
-	$(CONTAINER_RUNTIME) run --name=${CRIO_INSTANCE} -e BUILDTAGS --privileged -v ${PWD}:/go/src/${PROJECT} --rm ${CRIO_IMAGE} make binaries
-
-integration: crioimage
-	$(CONTAINER_RUNTIME) run -e STORAGE_OPTIONS="--storage-driver=vfs" -e TEST_USERNS -e TESTFLAGS -e CI -e CRIO_BINARY -t --privileged --rm -v ${CURDIR}:/go/src/${PROJECT} ${CRIO_IMAGE} make localintegration
+	$(CONTAINER_RUNTIME) run --rm --name=${CRIO_INSTANCE} --privileged \
+		-v $(shell pwd):/go/src/${PROJECT} -w /go/src/${PROJECT} -it \
```
vrothberg: Do we need `-it`?

test/apparmor.bats
```diff
@@ -14,6 +14,9 @@ function teardown() {
     if [[ "$enabled" -eq 0 ]]; then
         skip "skip this test since apparmor is not enabled."
     fi
+    if [[ "$CI" == "true" ]]; then
```
vrothberg: Can we add a comment why that's needed? It might be the heat but it's not obvious to me why it's needed now.

Makefile
saschagrunert: Hm, I will remove it since it was not there before.

Makefile
saschagrunert: Removed it.

test/apparmor.bats
saschagrunert: Hm, I never got this managed to work because of mainly this: https://stackoverflow.com/questions/25533666/cannot-reload-or-start-apparmor-in-docker/27029610#27029610

The tests are host dependent, so It's not nice to run them in parallel even if the host supports apparmor.

I will add a comment about that.

vrothberg: Docker uses a trick. There's a `dind` (Docker in Docker) somewhere in the tree to setup the mounts and privileges that AppArmor requires to be set.

saschagrunert:
I have to fix something in KATA here before we can continue: 
```
Failed at 70: critools_version=$(grep "ENV CRICTL_COMMIT" "${GOPATH}/src/${crio_repo}/Dockerfile" | cut -d " " -f3)
```

haircommander:
sorry @vrothberg were all of your comments addressed? I'll give you the final say here

vrothberg:
Let's merge this PR. We can enable the tests in a later one.

PR #2543, opened 2019-06-28
c/image API use cleanups

mtrmac:
**- What I did**

On top of #2514 , reviewed the uses of c/image API, and made various cleanups and smaller bug fixes.

Two notable major themes:
- `ListImages`/`ImageStatus` now do much less redundant work (decreasing the number of c/storage calls, and the implied locking operations necessary), and share more code
- `types.SystemContext` options that apply ~globally are set once in `server.Server.systemContext`, and automatically used throughout the call stack. Notably `GlobalAuthFile` is now handled only in one place.

**- How I did it**

(This depends on, and currently includes #2514, I will rebase this after that PR is merged.)

**- How to verify it**

Hopefully the existing CI.

`ListImages` and `ImageStatus` are now more efficient.

### DIFF
--- a/pkg/storage/image.go
+++ b/pkg/storage/image.go
@@ -270,15 +270,7 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
 	if err != nil {
 		return nil, err
 	}
-	imageFull, err := ref.NewImage(svc.ctx, systemContext)
-	if err != nil {
-		return nil, err
-	}
-	defer imageFull.Close()
-
-	size := imageSize(imageFull)
-	configDigest := imageFull.ConfigInfo().Digest
-	imageConfig, err := imageFull.OCIConfig(svc.ctx)
+	cacheItem, err := svc.buildImageCacheItem(systemContext, ref) // Single-use-only, not actually cached
 	if err != nil {
 		return nil, err
 	}
@@ -290,10 +282,10 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
 		Name:         name,
 		RepoTags:     tags,
 		RepoDigests:  repoDigests,
-		Size:         size,
+		Size:         cacheItem.size,
 		Digest:       imageDigest,
-		ConfigDigest: configDigest,
-		User:         imageConfig.Config.User,
+		ConfigDigest: cacheItem.configDigest,
+		User:         cacheItem.user,
 	}
 
 	return &result, nil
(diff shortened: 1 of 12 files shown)


rhatdan:
Need @saschagrunert and @vrothberg to review this.

saschagrunert:
> Need @saschagrunert and @vrothberg to review this.

I guess it's still work in progress, right? :)

mtrmac:
This was marked WIP to avoid merging #2514 without review; rebased now.

vrothberg:
I'll take a look now and hope be finished before the weekend starts :^)

server/image_pull_test.go
```diff
@@ -132,11 +132,7 @@ var _ = t.Describe("ImagePull", func() {
 
 		It("should fail credential decode erros", func() {
 			// Given
-			gomock.InOrder(
-				imageServerMock.EXPECT().ResolveNames(
-					gomock.Any(), gomock.Any()).
-					Return([]string{"image"}, nil),
-			)
+			gomock.InOrder()
```
saschagrunert: Please remove this line if not needed.

mtrmac: Fixed, thanks. (Kept the `// Given` comment based on `utils/utils_test.go` doing the same.)

pkg/storage/image.go
```diff
@@ -405,30 +349,23 @@ func (svc *imageService) PullImage(systemContext *types.SystemContext, imageName
 	if err != nil {
 		return nil, err
 	}
-	if options == nil {
-		options = &copy.Options{}
-	}
 
-	srcRef, err := svc.prepareReference(imageName, options)
+	options := *inputOptions // A shallow copy
```
haircommander: Are we no longer worried about null pointer dereference?

mtrmac: https://github.com/cri-o/cri-o/pull/2543/commits/39af0992a5d1cbaf0ab8bfee084387731aa9570a :
> Expect a non-nil copy.Options in ImageServer.PullImage
>
> Both callers pass a non-nil value already.

(I originally made the `copy.Options` a struct passed by value to make this enforced by the type system, but lint was complaining about a large parameter.  If only Go had explicit optionals instead of pointers with `nil`…)

mtrmac: That said, no further clean-ups depend on this, so I’d be happy to drop this change if the increased resilience is preferred.

haircommander: oops, missed the comment. If you're not worried about it, neither am I, I just wanted to make sure our expectations were mentioned at least

haircommander:
LGTM, awesome refactor @mtrmac !

PR #2545, opened 2019-06-28
cgroups: use libpod/pkg/cgroups

giuseppe:
replace containerd/cgroups with the implementation in libpod.  The
version in libpod is a minimal implementation tailored for
libpod/CRI-O.

The advantage is that libpod/pkg/cgroups is cgroups v2 aware, so CRI-O
would work on a cgroups v2 only system.


TomSweeneyRedHat:
LGTM assuming (fingers crossed) happy tests.

haircommander:
Why is this WIP @giuseppe ? LGTM

giuseppe:
> Why is this WIP @giuseppe ? LGTM

I pushed the PR before actually trying it out :-)  I'll rebase and drop the WIP, as it seems to work fine locally (at least I am able to run a pod on cgroup v2)

giuseppe:
@saschagrunert am I supposed to not commit changes to `vendor`?  I am getting a lot of changes under `vendor` but all I've done was to update `github.com/containers/libpod` and `github.com/godbus/dbus`

vrothberg:
> @saschagrunert am I supposed to not commit changes to `vendor`? I am getting a lot of changes under `vendor` but all I've done was to update `github.com/containers/libpod` and `github.com/godbus/dbus`

It's odd since those deps are still in `./vendor`

vrothberg:
@giuseppe, those are "hidden" behind libpod which has now migrated to go modules. Those deps are now taken from there and when doing `go mod vendor` those are mentioned in the `modules.txt`. Everything seems to be in order.

vrothberg:
FYI @edsantiago

giuseppe:
> @giuseppe, those are "hidden" behind libpod which has now migrated to go modules. Those deps are now taken from there and when doing `go mod vendor` those are mentioned in the `modules.txt`. Everything seems to be in order.

thanks!  I'll split the commit then, otherwise it is difficult to read it

saschagrunert:
> @saschagrunert am I supposed to not commit changes to `vendor`? I am getting a lot of changes under `vendor` but all I've done was to update `github.com/containers/libpod` and `github.com/godbus/dbus`

Sorry for being late. Yeah I found the issue with godbus/dbus too, we could bump it up to the latest master branch. We can do this later since this PR is already on its way into the master 👍

PR #2548, opened 2019-07-01
Remove unnecessary indirect dependency gopopulate

saschagrunert:
The dependency is not needed within our go.mod, so we remove it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ go 1.12
 module github.com/cri-o/cri-o
 
 require (
-	github.com/14rcole/gopopulate v0.0.0-20180821133914-b175b219e774 // indirect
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@vrothberg @giuseppe @umohnani8 @haircommander @mrunalp PTAL

vrothberg:
/lgtm

Note that golang v1.13 would remove that with `go mod tidy`. v1.12 still has some rough edges :(

saschagrunert:
> Note that golang v1.13 would remove that with `go mod tidy`. v1.12 still has some rough edges :(

No worries, it seems a pretty good sign that we have only a few indirect dependencies left.

PR #2549, opened 2019-07-01
Remove unnecessary sudo from Makefile

saschagrunert:
Dependent targets to also not run via sudo, which will break up
image-preparation when running via podman.


rhatdan:
/retest
LGTM

PR #2550, opened 2019-07-01
Fix seccomp checks for integration tests

saschagrunert:
The manually hard-enabled seccomp variable should not be needed if we
pass the right path for the "checkseccomp" executable into the CI.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #2551, opened 2019-07-01
Fix invalid CRIO_ROOT for integration tests

saschagrunert:
The `$CRIO_ROOT` variable seems to be not the root of the repository,
which make the integration tests fail outside of the usual
`../cri-o/cri-o` GOPATHs.


umohnani8:
LGTM
@haircommander @mrunalp @rhatdan PTAL

PR #2552, opened 2019-07-01
Remove invalid unit test to unblock master branch CI

saschagrunert:
The library creation does not validate the runtime config that tight any
more, so we should remove that invalid use case.


haircommander:
LGTM, I'll slap it with the tag once the test goes green

PR #2556, opened 2019-07-02
Add context based logging facade and interceptors

saschagrunert:
This commit adds stream and unary based logging interceptors which
automatically add an request ID (UUID) to the given context. The request
and responses are logged automatically from now on.

All functionality will be wrapped within the internal `log` package to
isolate implementation details, whereas logs within the `server` package
have been adapted to use the new `log` package.

---

We now add two interceptors to the GRPC server creation, whereas the request and responses are logged by default. A logging message would look like this for example:

```
DEBU[2019-07-02 10:27:29.269823804+02:00] request: &ListImagesRequest{Filter:&ImageFilter{Image:&ImageSpec{Image:,},},}  id=659a66be-d629-4fd0-9930-ce41e2888783
DEBU[2019-07-02 10:27:29.270081547+02:00] response: &ListImagesResponse{Images:[],}     id=659a66be-d629-4fd0-9930-ce41e2888783
```

We can see that we now have a UUID field attached to every message, whereas this UUID will be generated by the intercepting middleware. Everytime we now use the new `log` package within `cri-o/cri-o/pkg/log`, we are able to pass the context in and create a log entry which also uses this UUID.

### DIFF
--- a/server/container_attach.go
+++ b/server/container_attach.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/cri-o/cri-o/internal/oci"
-	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"k8s.io/client-go/tools/remotecommand"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
@@ -19,7 +18,6 @@ func (s *Server) Attach(ctx context.Context, req *pb.AttachRequest) (resp *pb.At
 		recordOperation(operation, time.Now())
 		recordError(operation, err)
 	}()
-	logrus.Debugf("AttachRequest %+v", req)
 
 	resp, err = s.getAttach(req)
 	if err != nil {
--- a/server/container_portforward.go
+++ b/server/container_portforward.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/cri-o/cri-o/internal/oci"
-	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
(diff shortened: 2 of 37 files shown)


server/container_attach.go
```diff
@@ -19,7 +18,6 @@ func (s *Server) Attach(ctx context.Context, req *pb.AttachRequest) (resp *pb.At
 		recordOperation(operation, time.Now())
 		recordError(operation, err)
 	}()
-	logrus.Debugf("AttachRequest %+v", req)
```
haircommander: was there a reason for dropping this?

server/container_portforward.go
```diff
@@ -18,7 +17,6 @@ func (s *Server) PortForward(ctx context.Context, req *pb.PortForwardRequest) (r
 		recordOperation(operation, time.Now())
 		recordError(operation, err)
 	}()
-	logrus.Debugf("PortForwardRequest %+v", req)
```
haircommander: same question here

server/container_attach.go
saschagrunert: All requests are now automatically logged via the interceptor:
https://github.com/cri-o/cri-o/blob/53f82f7858c7bfa47fb98e0fbf7599500413907a/pkg/log/interceptors.go#L57-L65

This way we have a homogeneous look and feel for the standard log messages without doing any further work.

pkg/log/log.go
```diff
@@ -0,0 +1,40 @@
+// Package log provides a global interface to logging functionality
+package log
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+)
+
+type id struct{}
```
jpeach: I think it would be more obvious to use strings for context value keys, e.g "crio/id". Unless it's for performance reasons, in which case this probably needs a comment.

jpeach: I think that using strings would let us attach a container ID as well?

saschagrunert: Golint complains when I use a string there:
```
should not use basic type string as key in context.WithValue
```

I think its about possible collisions we have to avoid. :)

jpeach: On interesting. Well that won't preclude us adding a container ID in the future so 👍

saschagrunert:
I guess this is now ready for review, too.

@haircommander @rhatdan @vrothberg PTAL :)

TomSweeneyRedHat:
LGTM and happy green tests

saschagrunert:

PR #2557, opened 2019-07-02
Move main configuration into Validate method

saschagrunert:
This allow testing the actual logic and cleans up some exposed variables
which can be internal now. The SELinux disabling already supports
multiple OS'es, where we can remove that from `cmd/crio` too.


rhatdan:
LGTM

PR #2558, opened 2019-07-02
Change default apparmor profile to actually contain the version

saschagrunert:
This ensures that the default configuration of CRI-O always installs
the default apparmor profile named `crio-default-vx.xx.x[-dev]`. The
logging verbosity has been increased during startup to give the user a
higher transparency about the current profile.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -22,13 +22,13 @@ import (
 	"github.com/containers/libpod/pkg/apparmor"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/cri-o/cri-o/lib"
+	libconfig "github.com/cri-o/cri-o/lib/config"
 	"github.com/cri-o/cri-o/lib/sandbox"
 	"github.com/cri-o/cri-o/oci"
 	"github.com/cri-o/cri-o/pkg/signals"
 	"github.com/cri-o/cri-o/pkg/storage"
 	"github.com/cri-o/cri-o/server/metrics"
 	"github.com/cri-o/cri-o/server/useragent"
-	"github.com/cri-o/cri-o/version"
 	"github.com/cri-o/ocicni/pkg/ocicni"
 	"github.com/fsnotify/fsnotify"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


vrothberg:
Can you elaborate why this change makes the defaults stronger? Please add that to the commit message as well. The idea of suffixing it was to assure smooth upgrades and to avoid collisions. Podman does that as well, and maybe even Docker (but I am not sure).

saschagrunert:
> The idea of suffixing it was to assure smooth upgrades and to avoid collisions. Podman does that as well, and maybe even Docker (but I am not sure).

I see your point, so I switched the implementation around that we now actually use the default profile (including the version) as default.

server/server.go
```diff
@@ -377,10 +375,13 @@ func New(
...
-	if s.appArmorEnabled && s.appArmorProfile == apparmorDefaultProfile {
-		if err := apparmor.InstallDefault(apparmorDefaultProfile); err != nil {
-			return nil, fmt.Errorf("ensuring the default apparmor profile %q is installed failed: %v", apparmorDefaultProfile, err)
+	if s.appArmorEnabled && config.ApparmorProfile == libconfig.DefaultApparmorProfile {
+		logrus.Infof("installing default apparmor profile: %v", libconfig.DefaultApparmorProfile)
+		if err := apparmor.InstallDefault(libconfig.DefaultApparmorProfile); err != nil {
+			return nil, fmt.Errorf("ensuring the default apparmor profile %q is installed failed: %v", libconfig.DefaultApparmorProfile, err)
 		}
+	} else {
+		logrus.Infof("assuming user-provided apparmor profile: %v", config.ApparmorProfile)
```
TomSweeneyRedHat: I'm personally not sure, so thought I'd solicit opinions.  Should this logrus here and the one at line 379 be Infof or Debugf?

TomSweeneyRedHat:
One question, otherwise LGTM

server/server.go
saschagrunert: I decided to go for the same verbosity of the seccomp profile log above and it only logs on server initialization. :) @vrothberg what do you think?

vrothberg: I don't feel strongly on that. `Info` might add to the logs but it's a security-related one and only printed at start-up, so it seems reasonable to me.

TomSweeneyRedHat: Since it follows the seccomp log then, let's leave it as Infof.  Thanks for entertaining the question.

PR #2440, opened 2019-06-03
Close temporary image in PullImage

m4burns:
Signed-off-by: Marc Burns <m4burns@uwaterloo.ca>

**- What I did**

The `ImageServer` interface in `pkg/storage` documents that images returned by `PrepareImage()` should have `Close()` called. `PullImage()` currently fails to close an image returned by `PrepareImage()`, which causes `/var/tmp` to fill with large temporary files when using at least the `docker-archive` transport. I fixed this issue.

**- How I did it**

I modified the `ImageServer` interface to allow `Close()` to be called on images returned by `PrepareImage()`. I added a deferred call to `Close()` in `PullImage()`.

**- How to verify it**

I updated the unit tests to check that `Close()` is being called as prescribed by the `ImageServer` interface.

**- Description for the changelog**

Fix temporary file leak when using `docker-archive` transport.

PR #2561, opened 2019-07-02
Close temporary image in PullImage

haircommander:
The ImageServer interface in pkg/storage documents that images returned
by PrepareImage() should have Close() called. Neglecting to close the
returned image will fill /var/tmp with large temporary files when using
at least the docker-archive transport.

This patch modifies said interface to allow Close() to be called and
adds a call to Close() in PullImage().

Signed-off-by: Marc Burns <m4burns@uwaterloo.ca>

Cherry-pick of: https://github.com/cri-o/cri-o/pull/2440

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/pkg/storage/image.go
+++ b/pkg/storage/image.go
@@ -89,7 +89,7 @@ type ImageServer interface {
 	ImageStatus(systemContext *types.SystemContext, filter string) (*ImageResult, error)
 	// PrepareImage returns an Image where the config digest can be grabbed
 	// for further analysis. Call Close() on the resulting image.
-	PrepareImage(imageName string, options *copy.Options) (types.Image, error)
+	PrepareImage(imageName string, options *copy.Options) (types.ImageCloser, error)
 	// PullImage imports an image from the specified location.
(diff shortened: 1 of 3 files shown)


mrunalp:
@mtrmac ptal

mtrmac:
Sure (although I question why a `docker-archive:` pull spec is a good idea, or should be supported at all).

mrunalp:
This was reported on crio channel on kubernetes slack. The user seems to be pulling from gitlab-registry.nordstrom.com/nordace/fastscore/engine:nordypy-0.3 and that doesn't look like `docker-archive`. Do we have a bug in parsing?

mtrmac:
I mean, closing the object is the right thing as far as the API goes anyway, so I have absolutely no objections to including this.

But, well, for `docker://` sources, `NewImage().Close()` does exactly nothing, and the users would _definitely_ tell if something mis-parsed `docker://` as `docker-archive:`.

mtrmac:
… so it _might_ be the case that this is not fixing whatever you think it’s fixing, or the story is not completely understood.

Either way, this PR won’t make anything worse AFAICT, and it can get it better if those rare transports really _are_ involved.

PR #2563, opened 2019-07-03
[1.14] Add username and homedir to generated password

mrunalp:
When generating a an entry in the /etc/passwd file, we
need to also add username and homedir, so that the user
can properly login.

Port of #2325

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -273,6 +273,17 @@ func setupContainerUser(specgen *generate.Generator, rootfs, mountLabel, ctrRunD
 		return fmt.Errorf("user group is specified without user or username")
 	}
 	imageUser := ""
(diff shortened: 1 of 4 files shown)


saschagrunert:
It looks like we have to make some lint and code adaptions to let the CI pass.

cgwalters:
Hm, so this logic to change `/etc/passwd` only exists in crio and not podman?

mrunalp:
@mheon Can you look into making sure that podman has this logic as well?

haircommander:
everything but codecov is green, manually merging

mrunalp:
Pushed the tag :+1:

rhatdan:
@cgwalters Podman should have this functionality as well.

PR #2570, opened 2019-07-04
Update codecov config to be less strict

saschagrunert:
I think this should solve our codecov issues with release branches. The configuration is not very well documented so I might miss some point here. Afterwards, we could enable codecov again.

/cc @umohnani8


saschagrunert:
Thanks, I would request another codecov integration after this has been merged. 👍

PR #2572, opened 2019-07-04
Move server config into lib/config

saschagrunert:
This removes a lots of code duplicates and simplifies testing. All tests
have been migrated to the lib/config package as well the overall logic.


lib/config/config.go
```diff
@@ -365,17 +409,28 @@ func (c *Config) UpdateFromFile(path string) error {
...
+
+	tc := tomlConfig{}
+	tc.fromConfig(c)
+
+	if err := e.Encode(tc); err != nil {
+		return nil, err
 	}
 
-	return ioutil.WriteFile(path, w.Bytes(), 0644)
+	return buffer.Bytes(), nil
```
jpeach: I think this code is fine, but it would be more conventional to encode to a io.Writer than to require bytes, e.g. `func (c *Config) WriteTo(out io.Writer) error`.

lib/config/config_test.go
```diff
@@ -25,6 +36,17 @@ var _ = t.Describe("Config", func() {
 			Expect(err).To(BeNil())
 		})
 
+		It("should succeed with runtime cheks", func() {
```
jpeach: s/cheks/checks/

lib/config/config_test.go
```diff
@@ -61,6 +95,77 @@ var _ = t.Describe("Config", func() {
...
+			// When
+			err := sut.Validate(nil, false)
+
+			// Then
+			Expect(err).NotTo(BeNil())
+		})
+
+		It("should fail on wrong default ulimits", func() {
+			// Given
+			sut.DefaultUlimits = []string{invalidPath}
```
jpeach: Maybe using a path here is a bit misleading, suggest `[]string{"invalid=-1:-1"}`

lib/config/reload.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	if err := c.ReloadPauseImage(newConfig); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// logConfig logs a config set operation as with info verbosity. Please use
+// always this function for setting configuration options to ensure consistent
```
jpeach: s/use always/always use/

lib/config/config.go
```diff
@@ -365,17 +409,28 @@ func (c *Config) UpdateFromFile(path string) error {
...
 
-	t := new(tomlConfig)
-	t.fromConfig(c)
+	return ioutil.WriteFile(path, b, 0644)
+}
 
-	if err := e.Encode(*t); err != nil {
-		return err
+// ToBytes encodes the config into a byte slice. It errors if the encoding
+// fails, which should never happen at all because of generael type safeness.
```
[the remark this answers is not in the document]
saschagrunert: :+1:

cmd/crio/config.go
```diff
@@ -313,10 +313,10 @@ var configCommand = cli.Command{
 		var err error
 		// At this point, app.Before has already parsed the user's chosen
 		// config file. So no need to handle that here.
-		config := c.App.Metadata["config"].(*server.Config) // nolint: errcheck
+		config := c.App.Metadata["config"].(*libconfig.Config) // nolint: errcheck
```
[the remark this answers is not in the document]
saschagrunert: Yes, fixed it as suggested.

lib/config/config_test.go
saschagrunert: Fixed :)

lib/config/reload.go
saschagrunert: Fixed :)

lib/config/config_test.go
saschagrunert: Applied as suggested.

lib/config/config.go
saschagrunert: Hm, I see your point but for now I would leave it as it is. I thought about removing this function entirely but it might have their use case.

saschagrunert:
Thanks for the review @jpeach, I applied your suggestions. :)

saschagrunert:
Rebased on top of the latest master branch to resolve conflicts.

saschagrunert:

rhatdan:
@mrunalp @haircommander @umohnani8 @vrothberg PTAL

haircommander:
LGTM

vrothberg:
`+661 −965` in one commit :scream: I don't have enough cycles today to review, but maybe tomorrow.

saschagrunert:
Is there anything wrong with tide right now? :thinking:

runcom:
this was wrong, moving config types (public and used by openshift as well) to internal makes them unvendorable

runcom:
the config types were public api as well and you cannot just move them and break between releases of kubernetes (for at least 2 releases iirc)

Issue #2575, opened 2019-07-07
Hard to cross-compile cri-o, due to config generation


afbjorklund:
**Description**

It is hard to build cri-o for another target (e.g. Buildroot), due to running tools on host.

This may fail for any number of reasons, including architecture and shared libraries.

**Steps to reproduce the issue:**
1. make minikube.iso
2.
3.

**Describe the results you received:**

`bin/crio-config: error while loading shared libraries: libdevmapper.so.1.02: cannot open shared object file: No such file or directory`

`./bin/crio: error while loading shared libraries: libdevmapper.so.1.02: cannot open shared object file: No such file or directory`

**Describe the results you expected:**

conmon/config.h

crio.conf

**Additional information you deem important (e.g. issue happens only occasionally):**

Happens every time, even if I provide my own configuration files.

Probably due to the usage of `git-vars` target, invalidating everything ?

**Output of `crio --version`:**

```
crio version 1.15.0
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

Buildroot 2018.05.3

saschagrunert:
Hey @afbjorklund :wave:, if I understand the issue correctly, then the `crio` binary builds correctly but it is not executable because the shared library for device mapper is not found? If so, then you may have to adapt the `LD_LIBRARY_PATH` in your build environment or exclude the device mapper support during build.

The `git-vars` target should be related to this issue in my point of view.

saschagrunert:
I opened-up a PR which should enhance the `git-vars` target to cover non-existing `git` executables as well: https://github.com/cri-o/cri-o/pull/2581

Can you verify that it works with that patch? If so, we can backport it to `release-1.15` as well.

afbjorklund:
The library does exist in the _target_ environment, it just does not exist in the _build_ environment.

It seems like I also need to do this make as root user, or it will generate crio.conf for rootless ?

afbjorklund:
My problem is not really `git-vars`, even though I don't really need that phony target (provide vars)

The problem is that _because_ of git-vars, it tries to generate the configuration using a target binary.

``` console
$ LANG=C make -dr conmon/config.h
```

Describe the results you received:
```
Considering target file 'conmon/config.h'.
  Considering target file 'git-vars'.
   File 'git-vars' does not exist.
   Finished prerequisites of target file 'git-vars'.
  Must remake target 'git-vars'.
  Successfully remade target file 'git-vars'.
[...]

PR #2581, opened 2019-07-08
Update git-vars target to check for executable as well

saschagrunert:
If the `git` executable is not available then the `git-vars` Makefile
target will now cover that as well.

Refers to #2575

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -1,6 +1,8 @@
+# vim: set syntax=make:
+
 .PHONY: git-vars
 git-vars:
-ifneq ($(wildcard .git),)
+ifeq ($(shell [[ `command -v git` && -d .git ]] && echo true),true)
 	$(eval COMMIT_NO :=$(shell git rev-parse HEAD 2> /dev/null || true))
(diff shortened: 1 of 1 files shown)


Makefile.inc
```diff
@@ -1,6 +1,8 @@
+# vim: set syntax=make:
+
 .PHONY: git-vars
 git-vars:
-ifneq ($(wildcard .git),)
+ifeq ($(shell [[ `command -v git` && -d .git ]] && echo true),true)
```
giuseppe: should the output for `[[ .. ]]` be redirected to `/dev/null`?

saschagrunert: I guess it will output nothing in any case. The `command -v git` returns a non-empty string only if the executable is found, which will make the `[[ ... ]]` test for `-d .git` afterwards. So the output within `[[ ... ]]` will never reach `std[out,err]`.

saschagrunert:
It seems that this PR does not fix the issue mentioned in #2575, but the issue could still happen: If I run a pure nix shell within the repository I won't be able to build CRI-O without the git executable. I would go forward with this change, since we have to ensure that both exists: the `.git` directory and the `git` executable.

rhatdan:
LGTM
@vrothberg @umohnani8 @haircommander @mrunalp @giuseppe PTAL

haircommander:
LGTM

PR #2583, opened 2019-07-08
Remove git-vars from conmon/config.h generation

saschagrunert:
The `git-vars` target force a rebuild of the `config.h` even if it
already exist. This can break build-setups, so we remove it from the
`config.h` generation.

Related discussion: https://github.com/kubernetes/minikube/pull/4703/files#r301143546

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ crio.conf: bin/crio
 release-note: ${RELEASE_TOOL}
 	${RELEASE_TOOL} -n $(release)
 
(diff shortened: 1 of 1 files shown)


saschagrunert:
I think we won't have this issue within the conmon repository at all, since the `make config` target it not a direct dependency during the build process. So in the end I guess we're fine with porting that one to release-1.15 and continuing our way to remove conmon here entirely.

/cc @haircommander

haircommander:
both SGTM and LGTM :)

rhatdan:
Needs a rebase.

saschagrunert:

PR #2585, opened 2019-07-09
Update codecov patch comparison to succeed

saschagrunert:
The patch comparison of codecov can still fail which is not what we
want. This can cause trouble if there are no coverage related code
changes involved.

Relevant codecov docs:
https://docs.codecov.io/docs/commit-status#section-patch-status


giuseppe:
LGTM

PR #2593, opened 2019-07-10
Fix git-vars target for shells which do not default to bash

saschagrunert:
Some environments default to `/bin/sh` as shell, which will break the
`git-vars` target because of its `[[ .. ]]` syntax. To avoid such, we
now execute a bash shell around it.

Another fix included in this commit is that the `.git` directory does
not necessary needs to be part of the current working directory, for
example when including the Makefile.inc in the `conmon` dir. For that
we now use the command `git rev-parse --git-dir` to avoid this issue.

The version.go file is now located in `internal/version`, which has been
fixed as well.

---

The output can be verified at: https://circleci.com/gh/openSUSE/cri-o/12346

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -2,7 +2,7 @@
 
 .PHONY: git-vars
 git-vars:
-ifeq ($(shell [[ `command -v git` && -d .git ]] && echo true),true)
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

PR #2598, opened 2019-07-11
Do not print 'unknown' versions

saschagrunert:
If the version is 'unknown' we don't print it anymore to avoid user
confusion.


rhatdan:
LGTM

PR #2599, opened 2019-07-11
Add missing config template values

saschagrunert:
Add missing `insecure_registries` and `bind_mount_prefix` to config template.

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -194,6 +194,9 @@ container_exits_dir = "{{ .ContainerExitsDir }}"
 # Path to directory for container attach sockets.
 container_attach_socket_dir = "{{ .ContainerAttachSocketDir }}"
 
+# The prefix to use for the source of the bind mounts.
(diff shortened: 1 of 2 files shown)


rhatdan:
The only thing with this, is we should tell the user that we would prefer that they make registries changes in registries.conf rather then in crio.conf.  That way the other tools, skopeo, Podman, Buildah will follow the rules. If you change just crio.conf the other tools have no idea.

saschagrunert:
> The only thing with this, is we should tell the user that we would prefer that they make registries changes in registries.conf rather then in crio.conf. That way the other tools, skopeo, Podman, Buildah will follow the rules. If you change just crio.conf the other tools have no idea.

Alright, I changed the message to take `/etc/containers/registries.conf` into account before using `insecure_registries`.

saschagrunert:
Found some more missing entries in `docs/crio.conf.5.md`, PTAL.

PR #2600, opened 2019-07-11
[1.14] run crio-wipe from /usr and add system-preset

haircommander:
In openshift, the machine-os-content container is spilled onto disk, and pivot is called. pivot only copies /usr and /etc, and ignores /var. Thus, for crio-wipe to actually exist on reboot, it must be installed into usr.

### DIFF
--- a/contrib/crio-wipe/50-crio-wipe.preset
+++ b/contrib/crio-wipe/50-crio-wipe.preset
@@ -1,0 +1,1 @@
+disable crio-wipe.service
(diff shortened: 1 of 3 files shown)


haircommander:
@umohnani8 @mrunalp @rhatdan @vrothberg PTAL

haircommander:
btw, if we want crio-wipe to run on reboot, we will need to enable it with a preset. I added the preset as disabled in the PR, and we can enable it on installation if needed. PTAL

rhatdan:
@haircommander Did you accidently push a second commit here?

haircommander:
@rhatdan no, I added that on purpose!
1: run from /usr
2: add system-preset for rpm to overload
3: bump version so we only have to go through CI race once

haircommander:
@rhatdan @umohnani8 @mrunalp needs approval too :)

PR #2602, opened 2019-07-12
[1.15] Add release bundle target and documentation

saschagrunert:
This commit adds a new `release-bundle` target for building a static
release archive locally. Documentation has been added as well. Since we
default to rootless podman, the additional `chown` is not needed any
more when compiling a static binary.


rhatdan:
LGTM
@umohnani8 @haircommander @mrunalp @vrothberg PTAL

PR #2307, opened 2019-04-29
[1.13] sandbox_run: Log a warning if we can't find a slice

mrunalp:
Don't fail hard if we can't find a slice.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

@umohnani8 ptal

PR #2605, opened 2019-07-12
[1.14] sandbox_run: Log a warning if we can't find a slice

mrunalp:
Don't fail hard if we can't find a slice.

Port of #2307 


haircommander:
Unfortunately, these aren't the flakes we're seeing in master these are consistent errors. Something changed with networking and i'm not really sure what. I don't think we'll succeed in spamming retest :/

saschagrunert:
We have to backport the `parse_pod_ip` fix of https://github.com/cri-o/cri-o/commit/73e72fcdf238156710ad0c4d6e58a6b64f35b2d7 to `release-1.1x`. 

The AMI update applies to all branches, which contained the CNI plugins update. I think it would be useful to have dedicated AMIs for the release branches to decouple the dependencies.

PR #2618, opened 2019-07-16
Remove deprecated file-locking

saschagrunert:
We deprecated file-locking with the last release so it could be removed
for the next one. Tests and documentation has been adapted as well.


giuseppe:
LGTM

PR #2621, opened 2019-07-17
Return if error loading cgroup CleanupConmonCgroup

mariusgrigoriu:
Return before using nil cgroup if there was an error loading conmon cgroup for a container

Fixes #2620 

**- What I did**
Return if error loading cgroup before calling cg.Delete()

**- How to verify it**
Looks like CleanupConmonCgroup doesn't have any tests. What would be a good way to test this?

**- Description for the changelog**
Don't panic if conmon cgroup could not be loaded during cleanup (Fixes #2620)

PR #2623, opened 2019-07-17
[1.15] Return if error loading cgroup CleanupConmonCgroup

vrothberg:
Return before using nil cgroup if there was an error loading conmon
cgroup for a container.

Signed-off-by: Marius Grigoriu <marius.grigoriu@me.com>
Backported-by: Valentin Rothberg <rothberg@redhat.com>

---

Backport of https://github.com/cri-o/cri-o/pull/2621


vrothberg:
Force merging because of codecov...

Issue #2625, opened 2019-07-17
kube-proxy pod unable to write to /sys with cri-o


matthias50:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------
**Description**

When I start kube-proxy using cri-o I get this error from kube-proxy:

```
E0717 20:59:57.644545       1 conntrack.go:127] sysfs is not writable: {Device:sysfs Path:/sys Type:sysfs Opts:[ro seclabel nosuid nodev noexec relatime] Freq:0 Pass:0} (mount options are [ro seclabel nosuid nodev noexec relatime])
```

The same pod manifest works with Docker (i.e. no such errors)

**Steps to reproduce the issue:**
1. start kube-proxy using a static pod manifest.

**Describe the results you received:**
kube-proxy can't write to /sys

**Describe the results you expected:**
kube-proxy is able to write to /sys since it is a privileged container.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `crio --version`:**

```
crio version 1.13.6-1.dev.rhaos4.1.gitee2e748.el7-dev
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

Running in AWS.

kube-proxy pod spec:

```
apiVersion: v1
kind: Pod
metadata:
 annotations:
   scheduler.alpha.kubernetes.io/critical-pod: ""
 creationTimestamp: null
 labels:
   k8s-app: kube-proxy
   tier: node
 name: kube-proxy
 namespace: kube-system
spec:
 priorityClassName: "system-node-critical"
 containers:
 - command:
   - /bin/sh
   - -c
   - exec kube-proxy
     --config=/srv/kubernetes/kube-proxy/kube-proxy-config.yaml
     2>&1
   image: <redacted>.dkr.ecr.us-west-2.amazonaws.com/mirror/kube-proxy:v1.13.7
   imagePullPolicy: IfNotPresent
   name: kube-proxy
   resources:
     requests:
       cpu: 100m
   securityContext:
     privileged: true
   volumeMounts:
   - mountPath: /srv/kubernetes/kube-proxy
     name: kube-proxy
     readOnly: true
   - mountPath: /etc/ssl/certs
     name: ssl-certs-hosts
     readOnly: true
   - mountPath: /run/xtables.lock
     name: xtables-lock
     readOnly: false
   - mountPath: /lib/modules
     name: lib-modules
     readOnly: false
 hostNetwork: true
 volumes:
 - hostPath:
[...]

PR #2627, opened 2019-07-18
privileged: set mounts to rw

vrothberg:
Set all OCI mounts as privileged when running in privileged mode.

Fixes: #2625

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -208,6 +208,7 @@ func makeAccessible(path string, uid, gid int) error {
 	return nil
 }
 
+// nolint:gocyclo
 func (s *Server) createSandboxContainer(ctx context.Context, containerID, containerName string, sb *sandbox.Sandbox, sandboxConfig *pb.PodSandboxConfig, containerConfig *pb.ContainerConfig) (*oci.Container, error) {
 	if sb == nil {
 		return nil, errors.New("createSandboxContainer needs a sandbox")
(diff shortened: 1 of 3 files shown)


server/container_create_linux.go
```diff
@@ -626,6 +627,18 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		if err := specgen.AddOrReplaceLinuxNamespace(string(rspec.NetworkNamespace), netNsPath); err != nil {
 			return nil, err
 		}
+
+		if privileged {
+			specgen.RemoveMount("/sys")
+			specgen.RemoveMount("/sys/cgroup")
```
giuseppe: it should be `/sys/fs/cgroup`

vrothberg:
@giuseppe @rhatdan PTAL

server/container_create_linux.go
```diff
@@ -686,6 +699,10 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		specgen.AddMount(mnt)
 	}
 
+	if privileged {
+		setOCIBindMountsPrivileged(&specgen)
```
giuseppe: does it also override the mode for bind mounts that were configured as `ro`?  Probably doesn't matter as the privileged container can always remount it

vrothberg: Those are added after that call, so we're good. Thanks for reviewing!

vrothberg:
Preparing backports now.

Issue #2255, opened 2019-04-15
Allow RuntimeImpl implementations to be provided as external .so file


sboeuf:
In order to limit the amount of dependencies onto external repositories, we should be able to provide through `crio.conf` an implementation of a specific type of `RuntimeImpl` as a shared library (`.so` file).
A concrete example is the `runtime_vm.go` implementation, that could live inside https://github.com/kata-containers/runtime repository. A user that would like to use this implementation would provide the `runtime_vm.so` file provided by Kata Containers, given that it would specify the path from the `crio.conf` file.

mrunalp:
:+1:

amshinde:
@mrunalp An update on this issue.
I moved the implementation of runtime_vm.go to  https://github.com/kata-containers/runtime repository and  vendored  the `oci` package under crio into kata as the vm implementation requires data structures defined in this package. With this I had to resolve all the dependency conflicts between the transitive dependencies introduced by the vendoring of crio oci package and several other packages that it depends on.

With this I changed the crio.conf file to point to the plugin path instead. The crio code was modified to open the plugin and perform a lookup for the functions within this plugin as specified in the plugin api.
I ran into a couple of issues while doing so. First I saw this panic:
`panic: http: multiple registrations for /debug/request`.

You can see a similar issue here: https://jira.hyperledger.org/browse/FAB-12031
The issue seems to be a problem with `golang.org/x/net` package, which when vendored twice can cause this panic. crio includes this in its vendor directory and the newly created plugin ends up requiring this vendor as well, so while opening this plugin, you see the above error.
To test it out, I simply deleted the package from `vendor` directory from crio repo and this seems to resolve the panic seen while opening the plugin.

We do need to find a workable solution for this though. Looking at the crio repo I see that it uses go modules, so was wondering why crio includes the `vendor` dir as well? 

The second issue that I am seeing now is with the Lookup call. I am using the plugin api to lookup a function in the plugin as : 
```
 newRuntimeFunc, err := plug.Lookup("NewRuntimeVM")
                if err != nil {    
                        return nil, err
                }
```

But I am seeing a panic with this:
```
panic: interface conversion: plugin.Symbol is func(string) oci.RuntimeImpl, not func(string) oci.RuntimeImpl (types from different scopes)
```

Here is a similar issue someone else has come accross: https://github.com/golang/go/issues/30838
According to the comment on the issue:
```
[...]

PR #2633, opened 2019-07-19
WIP: Make the OCI interface public

saschagrunert:
A first approach in splitting external and internal OCI types.


haircommander:
a quick skim LGTM, I'll do a more thorough review in a bit

my one concern is I wanted to start reducing duplication between libpod's oci implementation and cri-o's. Included in this was ensuring feature pairity of the oci runtime interface to the needs of libpod. I can escalate that work if this PR is desired, but I'd like to make sure libpod doesn't need an API extension before we make this public.

WDYT @vrothberg @saschagrunert

saschagrunert:
> a quick skim LGTM, I'll do a more thorough review in a bit
> 
> my one concern is I wanted to start reducing duplication between libpod's oci implementation and cri-o's. Included in this was ensuring feature pairity of the oci runtime interface to the needs of libpod. I can escalate that work if this PR is desired, but I'd like to make sure libpod doesn't need an API extension before we make this public.
> 
> WDYT @vrothberg @saschagrunert

Sounds good, I agree we should not do the work multiple times. Lets also wait for #2255 to see if such an API fits.

vrothberg:
> my one concern is I wanted to start reducing duplication between libpod's oci implementation and cri-o's. Included in this was ensuring feature pairity of the oci runtime interface to the needs of libpod. I can escalate that work if this PR is desired, but I'd like to make sure libpod doesn't need an API extension before we make this public.
> 
> WDYT @vrothberg @saschagrunert

In general, I have a bad feeling about exposing APIs. Containers are moving fast, and the golang ecosystem is not as strict as the C community when it comes to stability etc. Making other projects depend on CRI-O on that level makes me _very_ nervous. Using libpod would make sense as this would (or should) be the common ground of many projects. I cannot quantify it (yet) but I find libpod's API cleaner than CRI-O's. Less redundancy, less scattering, etc.

rhatdan:
I agree the interface should be in libpod, not CRI-O.

saschagrunert:
As I understood https://github.com/cri-o/cri-o/issues/2255 correctly, CRI-O should be able to depend on runtime implementations which follow a common functionality. If we now put the interface into libpod, then developers have to depend on libpod to build a runtime for CRI-O?

Do we want to add the runtime interface I exposed here to libpod and depend on that implementation?

rhatdan:
I want to eventually move most of the internals of CRI-O to libpod, and better unify the projects.

saschagrunert:
> I want to eventually move most of the internals of CRI-O to libpod, and better unify the projects.

Sounds reasonable, I will close this one for now.

PR #2640, opened 2019-07-19
[1.13] conmon: check it is a valid pid before killing it

haircommander:
always check it is a valid PID before passing it down to kill(2).

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1730919

Signed-off-by: Giuseppe Scrivano gscrivan@redhat.com


saschagrunert:
Have to merge manually, because I did not know that if a context does not exist, prow will never succeed at all... :man_facepalming: 

Following up on: https://github.com/openshift/release/pull/4457

PR #2615, opened 2019-07-15
[1.14] properly enable crio-wipe

mrunalp:
Carry of #2604 since the version bump commits over there are old and our release will miss fixes merged since then.

PR #2641, opened 2019-07-19
Crio wipe fixes

haircommander:
picks of https://github.com/cri-o/cri-o/pull/2600 and https://github.com/cri-o/cri-o/pull/2615 for master


saschagrunert:
LGTM, but do we need it for the static bundle as well?

haircommander:
yes if we use the service file here we would need it bundled with the static builds

PR #2643, opened 2019-07-21
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
**- What I did**
Make cri-o build reproducibly by default.

**- How I did it**
Default `BUILD_INFO` to `SOURCE_DATE_EPOCH`, if available.

**- How to verify it**
1. `export SOURCE_DATE_EPOCH=123456`
2. Do 2 clean builds within `taskset 1` (needed because of some go parallelism bug) and compare `/usr/bin/crio` - should have identical content between builds.

Allow to override build date with SOURCE_DATE_EPOCH

in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

While it is already possible to do reproducible builds without
this patch, this patch makes it the default,
so that distributions do not have to discover the BUILD_INFO variable.

Without this patch, `/usr/bin/crio` differs in ELF section .note.go.buildid

Signed-off-by: Bernhard M. Wiedemann <bwiedemann@suse.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ OCIUMOUNTINSTALLDIR=$(PREFIX)/share/oci-umount/oci-umount.d
 
 SELINUXOPT ?= $(shell selinuxenabled 2>/dev/null && echo -Z)
 
-BUILD_INFO := $(shell date +%s)
+SOURCE_DATE_EPOCH ?= $(shell date +%s)
 
 GO_MD2MAN := ${BUILD_BIN_PATH}/go-md2man
 GINKGO := ${BUILD_BIN_PATH}/ginkgo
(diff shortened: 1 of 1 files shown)


haircommander:
/ok-to-test
Change LGTM, though i'd almost prefer to replace the BUILD_INFO variable with SOURCE_DATE_EPOCH and set it with
`SOURCE_DATE_EPOCH ?= $(shell date +%s)`

saschagrunert:
Hey @bmwiedemann :wave:, thanks for the improvement. I agree with @haircommander and I think we can replace the `$(BUILD_INFO)` variable entirely.

bmwiedemann:
hmm, the integration test relies on BUILD_INFO somehow?

saschagrunert:
> hmm, does the integration test rely on BUILD_INFO somehow?

No, we have some flakes in the integration tests which we will tackle in the future :innocent:

PR #2649, opened 2019-07-23
Add crio-status client for convenient HTTP API access

saschagrunert:
The client API is now internal but we can provide a simple command line
interface for it.

### DIFF
--- a/cmd/crio-status/main.go
+++ b/cmd/crio-status/main.go
@@ -1,0 +1,160 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/cri-o/cri-o/internal/client"
+	"github.com/cri-o/cri-o/internal/version"
+	"github.com/urfave/cli"
+)
+
+const (
+	defaultSocket = "/var/run/crio/crio.sock"
+	idArg         = "id"
+	socketArg     = "socket"
+)
+
+func main() {
+	app := cli.NewApp()
+	app.Name = "crio-status"
+	app.Usage = "A tool for CRI-O status retrieval"
+	app.Version = version.Version
+	app.CommandNotFound = func(*cli.Context, string) { os.Exit(1) }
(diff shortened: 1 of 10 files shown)


vrothberg:
> Code LGTM but tests and man pages are missing.

And bash-completions would be great.

cmd/crio-status/main.go
```diff
@@ -0,0 +1,160 @@
...
+			),
+		},
+	}
+	app.Flags = flags
+	app.Commands = []cli.Command{{
+		Action:  config,
+		Aliases: []string{"c"},
+		Flags:   flags,
+		Name:    "config",
+		Usage:   "retrieve the configuration as TOML string",
```
TomSweeneyRedHat: nit suggestion  s/as TOML/as a TOML/

saschagrunert:
> Code LGTM but tests and man pages are missing.

Added it. :+1: 

> And bash-completions would be great.

Can we tackle this later on? The [urfave/cli](https://github.com/urfave/cli#bash-completion) generator offers only a two-step generation where I'm not sure if we want this... Maintaining seems also not nice to me, WDYT?

cmd/crio-status/main.go
saschagrunert: Fixed as suggested.

vrothberg:
> Can we tackle this later on? The [urfave/cli](https://github.com/urfave/cli#bash-completion) generator offers only a two-step generation where I'm not sure if we want this... Maintaining seems also not nice to me, WDYT?

Why not generate it now and add it? I think it's better to have a simple completion than none. If you prefer to add this later, please open an issue so we don't forget about it. My squirrel brain would certainly :^)

saschagrunert:
> Why not generate it now and add it? I think it's better to have a simple completion than none. If you prefer to add this later, please open an issue so we don't forget about it. My squirrel brain would certainly :^)

Added the bash completion and adapted the install scripts. As a little extra I added a fish completion, too! :heart_eyes:

PR #2652, opened 2019-07-23
Make image volumes writeable by the container user

### LINKED ISSUE #2562: Make image volumes writeable by the container user
In `server/container_create.go:addImageVolumes()`, the `config.ImageVolumesMkdir` volume is create with mode `0755` and UID `0`. Unless the container process also runs as UID `0`, it won't be able to write to the volume. Maybe the ownership or permissions should be adjusted to ensure that image volumes are always writeable from inside the container.

samc24:
Fixes #2562, to ensure that image volumes are always writeable from inside the container.
User will have correct permissions if `run_as_user` value is specified for the container.
Also added a test for this feature.

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containers/storage/pkg/idtools"
 	"github.com/cri-o/cri-o/internal/lib/config"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/pkg/log"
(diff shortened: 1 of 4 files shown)


server/container_create.go
```diff
@@ -165,7 +166,22 @@ func addImageVolumes(ctx context.Context, rootfs string, s *Server, containerInf
 		}
 		switch s.config.ImageVolumes {
 		case config.ImageVolumesMkdir:
-			if err1 := os.MkdirAll(fp, 0755); err1 != nil {
+			imageUser := ""
```
mrunalp: Let's not repeat the logic here for getting the container user. You can get it back from the spec after uid, gid are added to it and move the code around to make sure that we setup User before we add the image volumes.

umohnani8:
@saschagrunert any idea what we need to do to make codecov happy?

saschagrunert:
> @saschagrunert any idea what we need to do to make codecov happy?

Nothing I guess, prow will ignore it :) for now it’s only a visual indicator that the coverage dropped.

server/container_create.go
```diff
@@ -165,7 +166,10 @@ func addImageVolumes(ctx context.Context, rootfs string, s *Server, containerInf
 		}
 		switch s.config.ImageVolumes {
 		case config.ImageVolumesMkdir:
-			if err1 := os.MkdirAll(fp, 0755); err1 != nil {
+			uid := specgen.Config.Process.User.UID
+			gid := specgen.Config.Process.User.GID
+			IDs := idtools.IDPair{UID: int(uid), GID: int(gid)}
```
umohnani8: nit: you can do all this in one line `IDs := idtools.IDPair{UID: int(specgen.Config.Process.User.UID), GID: int(specgen.Config.Process.User.GID)}`

umohnani8:
One nit, otherwise changes LGTM

giuseppe:
/retest

LGTM

haircommander:
while we're at it 
/cherry-pick release-1.15

PR #2654, opened 2019-07-23
[1.13] Allow default route to not be present

haircommander:
First, attempt to find an IP address that is clearly the primary. If we fail to do so, find one with /proc/net/route

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -387,11 +387,37 @@ func New(ctx context.Context, config *Config) (*Server, error) {
 
 	hostIP := net.ParseIP(config.HostIP)
 	if hostIP == nil {
+		// First, attempt to find a primary IP from /proc/net/route deterministically
 		hostIP, err = knet.ChooseBindAddress(nil)
 		if err != nil {
-			return nil, err
+			// if that fails, check if we can find a primary IP address unambiguously
+			allAddrs, err2 := net.InterfaceAddrs()
+			if err2 != nil {
+				return nil, errors.Wrapf(err, "Failed to read InterfaceAddrs after failing to choose bind address")
+			}
+
+			// we are hoping for exactly 1 possible IP
+			numPossibleIPs := 0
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -387,11 +387,37 @@ func New(ctx context.Context, config *Config) (*Server, error) {
...
+		}
+
+		// We either found no hostIP, or too many. Attempt to find a
+		// primary IP from /proc/net/route deterministically
+		if numPossibleIPs != 1 {
+			hostIP, err = knet.ChooseBindAddress(nil)
+			if err != nil {
+				return nil, err
+			}
+		}
```
sjenning: lets fall through to one more case.  if `ChooseBindAddress` fails, but `numPossibleIPs != 0`, lets just choice the first IPv4 address.  @mrunalp what do you think?

haircommander: per internal conversations, we've decided this could lead to restarts of CRI-O having addresses changed from under it. marking as resolved

mrunalp:
Needs port to 1.14, 1.15 and master.

server/server.go
```diff
@@ -387,11 +387,37 @@ func New(ctx context.Context, config *Config) (*Server, error) {
...
 	hostIP := net.ParseIP(config.HostIP)
 	if hostIP == nil {
+		// First, attempt to find a primary IP from /proc/net/route deterministically
 		hostIP, err = knet.ChooseBindAddress(nil)
 		if err != nil {
-			return nil, err
+			// if that fails, check if we can find a primary IP address unambiguously
+			allAddrs, err2 := net.InterfaceAddrs()
+			if err2 != nil {
+				return nil, errors.Wrapf(err, "Failed to read InterfaceAddrs after failing to choose bind address")
```
vrothberg: Can we return `err2` as well (`...address: %v", err2`)

vrothberg: Nevermind :) Too late.

Issue #531, opened 2017-05-20
image digests TODOs and bugs


runcom:
@nalind @mtrmac here's the situation:

- when pulling an image by tag we need to **always** store the digested reference along with the tag (same as docker does)
- we don't have a `RepoDigests` for `ImageStatus` and we just fill `RepoTags` with `status.Names` which just contains tagged references for that image
- (guess what) we need `RepoDigests` slice
- digests must not be stored and shown as the ID of an image (see reproducer 1). The image ID in docker is the digest of its `config` blob which should be unique due to content addressability!
-  if you pull an image by digest corresponding to a tag different than `latest` image list reports that it has a `:latest` tag whether that's not true at all (see reproducer 2)
- image ID should be the config blob digest!

--

reproducer 1 (digest shown as image ID, the image ID should be the `config` digest blob!):
```
$ crioctl image pull nginx@sha256:4aacdcf186934dcb02f642579314075910f1855590fd3039d8fa4c9f96e48315
$ crioctl image list
ID: 4aacdcf186934dcb02f642579314075910f1855590fd3039d8fa4c9f96e48315
Tag: docker.io/library/nginx:latest
```

--

reproducer 2 (wrong tag applied to a digested image, there should not be any tag for this image unless someone has previously pulled the same image by tag, in which case we already have the digest as well):
```sh
# this is the digest for redis:alpine but redis:latest is shown below, incorrectly
# notice how the same issue as reproducer 1 is shown below (ID != digest)
$ crioctl image pull redis@sha256:03789f402b2ecfb98184bf128d180f398f81c63364948ff1454583b02442f73b
$ crioctl image list
ID: 03789f402b2ecfb98184bf128d180f398f81c63364948ff1454583b02442f73b
Tag: docker.io/library/redis:latest
```

mtrmac:
@runcom Is there anything specific I should be doing?

FWIW it may not be _obviously necessary_ to exactly replicate the Docker config digest / manifest digest dichotomy, if the CRI API does not already embed the assumption that this is the way it works

As long as truly different images are not mixed together, it may make sense to e.g. identify images by manifest digest and ignore config digests completely (I guess, as long as layer chains _are_ deduplicated, because that costs disk space).

runcom:
@mtrmac wasn't the whole idea of using the config blob digest to be content-addressable? I keep hearing that all over the places and using the manifest digest sounds, instead, weird to me.

mtrmac:
[...]

PR #2655, opened 2019-07-24
[1.14] container_status: Use cached value of image name

mrunalp:
Use the name we stored in the container object in container status
rather than recalculating it.

@nalind ptal. I am reversing what was introduced in f3b7065bd80a53e4c48e8c45c36dcb86717e1578.

### DIFF
--- a/server/container_status.go
+++ b/server/container_status.go
@@ -4,7 +4,6 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/containers/image/types"
 	"github.com/cri-o/cri-o/oci"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


mariusgrigoriu:
Does this apply only to 1.14?

mrunalp:
@mariusgrigoriu No, it will be back and forward ported. I am starting with 1.14.

nalind:
This might backtrack on some of #531, but so long as e2e tests pass, I guess we can afford to in the interest of performance.

nalind:
Okay, there's no loss here.  This bit's been redundant for a while.

mrunalp:
Pushed an update to make the change inline with rest of the response.

PR #2659, opened 2019-07-24
[1.15] Update libpod to the latest version

saschagrunert:
The used version of libpod requires `renameat2(2)`, which needs glibc 2.28.
Our build toolchains are just using a lower version of glibc which will
break up the CI if the build cache gets poisoned.

The updated version of OCICNI needs adaption in the network setup in
`server/utils.go` as well.


umohnani8:
LGTM
@vrothberg @haircommander @mrunalp PTAL

TomSweeneyRedHat:
Changes LGTM.
However, I don't think all of the changes here are related to the libpod version bump.  I'd rather see a PR for only libpod changes, then a separate PR for the other changes.

TomSweeneyRedHat:
I should have said, "However, in the future I don't....".  i'm fine with this going through if the tests are happy.

vrothberg:
@TomSweeneyRedHat, I think those changes come from using `go.mod` which is a really nice candy when using modules.

TomSweeneyRedHat:
@vrothberg even the .yml change?

vrothberg:
> @vrothberg even the .yml change?

No, I was only referring to the vendor changes. The .yml and API changes should have been broken into separate commits, your're absolutely right.

saschagrunert:
> > @vrothberg even the .yml change?
> 
> No, I was only referring to the vendor changes. The .yml and API changes should have been broken into separate commits, your're absolutely right.

From a development perspective I would agree, but since we're generating the changelogs from the commit messages I would say it might be better to have a single commit stating the user facing change and containing the rest within the message.

vrothberg:
>  From a development perspective I would agree, but since we're generating the changelogs from the commit messages I would say it might be better to have a single commit stating the user facing change and containing the rest within the message.

Splitting commits into logical chunks is common practice and especially important for maintenance. I wouldn't worry about the changelogs - those should in any case not include _all_ but _relevant_ changes.

PR #2660, opened 2019-07-24
Add AUDIT_WRITE to default capabilities

saschagrunert:
To keep the compatibility with Docker we should add the `AUDIT_WRITE`
capability as a default, too.

Reference: https://github.com/moby/moby/pull/7179

Needs a backport to release-1.15 if acceptable.

/cc @rhatdan

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -100,7 +100,7 @@ crio [GLOBAL OPTIONS] config [OPTIONS]
 
 **--cpu-profile**="": set the CPU profile file path
 
-**--default-capabilities**="": capabilities to add to the containers (default: "CHOWN, DAC_OVERRIDE, FSETID, FOWNER, NET_RAW, SETGID, SETUID, SETPCAP, NET_BIND_SERVICE, SYS_CHROOT, KILL)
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
LGTM, one question

saschagrunert:
> Any changes to documentation necessary?

Hm, I'm unsure how we should document that in a nice way, but it had to be added to the man pages, right. :+1:

rhatdan:
NO, We don't want compatibility with Docker.  We want to be tigher then Docker for running default containers.  AUDIT_WRITE should not be needed.  If kubernetes wants to launch a container that needs AUDIT_WRITE or any other additionl capability it should add it to the spec.

rhatdan:
Podman has this, but I also believe we should remove it from here.

vrothberg:
> Podman has this, but I also believe we should remove it from here.

I just wanted to point this out, but I trust your judgement.

rhatdan:
No problem. We have dropped a hole bunch of others, and I am working on dropping more.  I basically want to differentiate between running containers in production versus running containers and playing with them. Podman and Building them Buildah.

Issue #2478, opened 2019-06-13
lazily load hostIP


cgwalters:
Today crio on *startup* tries to detect the primary host IP based on the default route.  AIUI this is simply so that address can be reported for `hostNetwork` pods.

The problem with this is that it creates a dependency on networking during startup which is racy.

xref: https://bugzilla.redhat.com/show_bug.cgi?id=1713478
xref: https://github.com/cri-o/cri-o/pull/2102

Filing this specifically because this is impacting us for RHEL CoreOS doing basic host sanity testing in qemu; looks like we're getting an IPv6 address first before IPv4.

If we did this lazily then crio's startup wouldn't be dependent on networking.

haircommander:
Unfortunately it seems it also be a fallback for the IP of the stream server. So we would either need to make sure a user configures a valid StreamAddress, or we can start after networking? WDYT @umohnani8 @mrunalp

cgwalters:
> Unfortunately it seems it also be a fallback for the IP of the stream server.

Stream can listen on all interfaces right?

cgwalters:
See also https://github.com/openshift/installer/pull/1768

haircommander:
something worth note here: if dhcp takes a bit to figure out hostname, cri-o will mount the incorrect hostname in `/etc/hosts`. since these files are static upon container creation, the containers that come up before the network is fully initialized would have a wonky hostname.

Of course, this is something that should be figured out on the kube level, bc dhcp can change the hostname again later... I am not sure how we'd do that other than having the container's /etc/hosts be a hard link to the hosts, rather than a bind mount

PR #2662, opened 2019-07-24
Allow default route to not be present

haircommander:
First, attempt to find an IP address from /proc/net/route. If we fail to do so, attempt to find a clear primary

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -402,39 +402,10 @@ func New(
 	s.restore()
 	s.cleanupSandboxesOnShutdown(ctx)
 
-	hostIP := net.ParseIP(config.HostIP)
-	if hostIP == nil {
-		// First, attempt to find a primary IP from /proc/net/route deterministically
-		hostIP, err = knet.ChooseBindAddress(nil)
-		if err != nil {
-			// if that fails, check if we can find a primary IP address unambiguously
-			allAddrs, err2 := net.InterfaceAddrs()
-			if err2 != nil {
-				return nil, errors.Wrapf(err, "Failed to read InterfaceAddrs after failing to choose bind address")
-			}
-
-			// we are hoping for exactly 1 possible IP
-			numPossibleIPs := 0
-			// adapted from: https://stackoverflow.com/a/31551220
-			for _, addr := range allAddrs {
-				if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
-					if ipnet.IP.To4() != nil {
-						numPossibleIPs++
-						hostIP = ipnet.IP
-						if numPossibleIPs > 1 {
-							break
-						}
-					}
-				}
-			}
-
-			// If there is no clear primary IP, we should fail
-			if numPossibleIPs != 1 {
-				return nil, errors.Wrapf(err, "Failed to find one IP address after failing to choose bind address")
-			}
-		}
+	hostIP, err := s.getHostIP(config.HostIP)
+	if err != nil {
+		return nil, err
 	}
-
 	bindAddress := net.ParseIP(config.StreamAddress)
 	if bindAddress == nil {
 		bindAddress = hostIP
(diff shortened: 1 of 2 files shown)


umohnani8:
LGTM

sjenning:
This doesn't seem to work
```
# ip r
10.42.15.0/24 dev ens3 proto kernel scope link src 10.42.15.10 metric 100 
10.88.0.0/16 dev cni0 proto kernel scope link src 10.88.0.1 linkdown 
169.254.169.254 via 10.42.15.2 dev ens3 proto dhcp metric 100 

# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc fq_codel state UP group default qlen 1000
    link/ether fa:16:3e:af:f4:19 brd ff:ff:ff:ff:ff:ff
    inet 10.42.15.10/24 brd 10.42.15.255 scope global dynamic noprefixroute ens3
       valid_lft 86197sec preferred_lft 86197sec
    inet6 fe80::f816:3eff:feaf:f419/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
3: cni0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default qlen 1000
    link/ether ea:25:e4:25:ec:bf brd ff:ff:ff:ff:ff:ff
    inet 10.88.0.1/16 brd 10.88.255.255 scope global cni0
       valid_lft forever preferred_lft forever
    inet6 fe80::e825:e4ff:fe25:ecbf/64 scope link 
       valid_lft forever preferred_lft forever

# systemctl start crio
Job for crio.service failed because the control process exited with error code.
See "systemctl status crio.service" and "journalctl -xe" for details.

# journalctl -u crio.service --since="20 seconds ago"
systemd[1]: Starting Open Container Initiative Daemon...
crio[6470]: time="2019-07-29 21:58:08.860059205Z" level=fatal msg="Failed to find one IP address after failing to choose bind address: no default routes found in "/proc/net/route" or "/proc/net/ipv6_route""
systemd[1]: crio.service: Main process exited, code=exited, status=1/FAILURE
systemd[1]: crio.service: Failed with result 'exit-code'.
systemd[1]: Failed to start Open Container Initiative Daemon.
```

server/server.go
```diff
@@ -454,6 +455,55 @@ func New(
 	return s, nil
 }
 
+// getHostIP first attempts to read the IP address of the host from /proc/net/route
+// If that fails (there is no /proc/net/route), it then attempts to deterministically find a default IP
+// It does so by finding exactly one IP address that is not a loop back, and not created by CNI.
+func (s *Server) getHostIP() (net.IP, error) {
```
sjenning: we need to make this unit testable

server/server.go
```diff
@@ -454,6 +455,55 @@ func New(
...
+	// First, attempt to find a primary IP from /proc/net/route deterministically
+	hostIP, err := knet.ChooseBindAddress(nil)
+	if err != nil {
```
sjenning: we can lose a level of indentation by just 
```
if err == nil {
  return hostIP, nil
}
```

server/server.go
```diff
@@ -454,6 +455,55 @@ func New(
...
+		// if that fails, check if we can find a primary IP address unambiguously
+		ifaces, err2 := net.Interfaces()
```
sjenning: then we can just reuse `err2`

server/server.go
```diff
@@ -454,6 +455,55 @@ func New(
...
+		if err2 != nil {
+			return nil, errors.Wrapf(err, "Failed to read network Interfaces after failing to choose bind address")
+		}
+
+		// we are hoping for exactly 1 possible IP
+		numPossibleIPs := 0
+		for _, iface := range ifaces {
```
sjenning: could we break this loop into a different function that we can unit test?

sjenning: gah, seems like it make a ton of different calls though that we'd have to mock up :disappointed:

haircommander:
@sjenning thanks for the review, I'll get to the unit tests tomorrow :+1: :smile:

haircommander:
@sjenning after some deliberation, i have deemed the function un-unit-testable without having control over the networking stack of the vm setup. I was hoping it could be mocked, or members could be spoofed, but the implementation needs both function implementations and member variables, so no dice :\

Also on @mrunalp's slack messages about using the CNI config to parse the network names, unfortunately ocicni doesn't currently expose that over the CNIPlugin interface, as such I added a comment to mark future work will be needed if the format changes. the cni0 you found is probably podman's cni, as CRI-O wouldn't have set one up by that point. I also verified the cni configurations on a cluster are in the cni* format, which they are.

All in all, I think this is ready, PTAL

sjenning:
LGTM.  want @mrunalp to give final lgtm.

rhatdan:
@mrunalp Looks like this is ready to merge, once you give LGTM.

haircommander:
/hold
I don't think I'm happy with this solution...

haircommander:
/hold cancel

Ok, I came to a solution I am happy with. Here, we are emulating the way the kubelet chooses its nodeIP, with the one change of prioritizing the legacy CRI-O-way of doing it (ChooseHostIP is called earlier than kubelet, which should be fine because the default route is a good choice in general). 

filtering by cni* was both hacky, and not a good long term solution (https://github.com/containers/libpod/pull/3755 was a PR that originally could have broken this assumption). Instead, let's rely on the ordering provided by InterfaceAddrs(), which is more reliable. 

@sjenning, PTAL; can you confirm this works with the setup you used above?

cgwalters:
This is related to https://github.com/cri-o/cri-o/issues/2478 right?

sjenning:
@haircommander should we expect e2e to pass now?

haircommander:
@sjenning nope, we had 3 days of green CI and we're broken again :\ 
@cgwalters sort of, we still depend on routing, but it allows some aspects of routing to not be configured, so this may help your usecase

PR #2670, opened 2019-07-25
Fix parse of memory.limit_in_bytes on 32-bit machines

rkojedzinszky:
Signed-off-by: Richard Kojedzinszky <richard@kojedz.in>

**- What I did**
Forward port of #2668

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -726,7 +726,7 @@ func AddCgroupAnnotation(ctx context.Context, g generate.Generator, mountPath, c
 				// strip off the newline character and convert it to an int
 				strMemory := strings.TrimRight(string(fileData), "\n")
 				if strMemory != "" {
-					memoryLimit, err := strconv.Atoi(strMemory)
+					memoryLimit, err := strconv.ParseInt(strMemory, 10, 64)
 					if err != nil {
 						return "", errors.Wrapf(err, "error converting cgroup memory value from string to int %q", strMemory)
 					}


saschagrunert:
Thanks @rkojedzinszky, I guess we have to port this later on to 1.14 and 1.15 too.

server/sandbox_run_linux.go
```diff
@@ -726,7 +726,7 @@ func AddCgroupAnnotation(ctx context.Context, g generate.Generator, mountPath, c
 				// strip off the newline character and convert it to an int
 				strMemory := strings.TrimRight(string(fileData), "\n")
 				if strMemory != "" {
-					memoryLimit, err := strconv.Atoi(strMemory)
+					memoryLimit, err := strconv.ParseInt(strMemory, 10, 64)
```
TomSweeneyRedHat: It's pre-tea for me, but shouldn't 64 be 32 here?

https://golang.org/pkg/strconv/#ParseInt  The bitSize argument specifies the integer type that the result must fit into. Bit sizes 0, 8, 16, 32, and 64 correspond to int, int8, int16, int32, and int64.

rkojedzinszky: The point is that the integer parsed there does not fit in 32 bits.

TomSweeneyRedHat: Right, but with it set to 64 here, does it fit?  I thought it would need to be 32 but perhaps I'm mis-reading the doc.

rkojedzinszky: Actually, my platform has:
```
# cat /sys/fs/cgroup/memory/memory.limit_in_bytes 
8796093018112
```
That does not fit in a 32 bit integer, and with that I got runtime conversion errors. Now it can parse it, and does not fail.

rkojedzinszky: My amd64 platform has even a larger value and on 64 bit systems int type has 64 bits. So there it fits.
```
$ cat /sys/fs/cgroup/memory/memory.limit_in_bytes
9223372036854771712
```

giuseppe:
/retest

LGTM

PR #2671, opened 2019-07-25
WIP: Use libpod for ListImages RPC

saschagrunert:
#### Summary

This is one of the first steps in the CRI-O to libpod migration. We now
utilize the `Runtime` structure of `libpod/image` to list images.

The reference based filter logic has been mostly taken from podman,
whereas the `ConvertImage` function now uses the `Inspect` API to
retrieve the necessary information.

Since we now use the runtime of libpod to list the images, we can remove
unneeded parts from the internal implementation.

#### Still to do:

- [ ] vendor the latest libpod master instead of the forked branch
- [ ] find a solution for the removed unit tests

#### Discussion

Do you think we should use `libpod/runtime.Runtime` instead of the `libpod/image.Runtime`? I think later on we would have to migrate to the higher level API, but this implies additional configuration mapping from CRI-O to libpod.

Before continuing here, can we add a public interface for `libpod/image.Runtime` to provide a mock-able abstraction?

/cc @haircommander @mrunalp @rhatdan @vrothberg


vrothberg:
I highly appreciate the effort but won't ack (nor nack if others want to merge) changes before the [requirement's doc for the migration](https://docs.google.com/document/d/1GKub_NvdGCo7veR8XXf5w3PaEpIwzU6TnTMKNDcErEA/edit?usp=sharing) is finished. I am repeating myself but a forceful migration is likely to hurt us, so we need to work closely with the libpod folks.

We had a chat yesterday with @mheon who is also planning to work on that.

> Before continuing here, can we add a public interface for libpod/image.Runtime to provide a mock-able abstraction?

Sorry for sending negative vibes :( I feel rather strongly against mocking libpod _here_. It inverts the responsibility of testing. I see CRI-O as a pure consumer of libpod and trust the integration, unit and e2e tests (in addition to libpod's tests).

haircommander:
In general, I'm happy this is moving forward, but I agree with @vrothberg in that we should pull libpod folks in and do this meticulously. 

I think the mock interface should move into libpod, and the libpod testing should be extended. @mheon has mentioned this being high on his priority, as it would aid in separating libpod and podman.

what was the reasoning behind using the forked version?

saschagrunert:
> what was the reasoning behind using the forked version?

I found a runtime panic in libpod which is fixed in the fork (see https://github.com/containers/libpod/pull/3643).

saschagrunert:
Alright, if @mheon is already working on that one then I'm closing this one. :+1:

haircommander:
I like this as a proof of concept though, thanks for kicking this off @saschagrunert !

vrothberg:
@saschagrunert, thanks a lot! Let's try to sync again soon on that. @mrunalp was thinking about moving the CRI-O call a bit earlier or creating a shared one with the libpod folks, so that we can better collaborate on this effort.

rhatdan:
Yes I agree that we need to move forward on this, and happy to see it happening.  I think moving the CRI-O call to a libpod/cri-o call would be helpful and moving the time up to around 14:00 UTC.

PR #2684, opened 2019-07-30
Use podman for integration tests

saschagrunert:
We don't rely on docker any more and test the integration tests now via
podman. The caching has been adapted as well to speedup the CI process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ stdenv: &stdenv
     GOCACHE: &gocache /tmp/go-build
     IMAGE: &image saschagrunert/criocircle:1.1.0
     JOBS: &jobs 8
+    PODMANCACHE: &podmancache /tmp/podman
     WORKDIR: &workdir /go/src/github.com/cri-o/cri-o
 
 executors:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -39,6 +40,36 @@ executors:
...
+  run:
+    name: setup podman
+    command: |
+      # Install podman
+      sudo add-apt-repository ppa:projectatomic/ppa -y
+      sudo apt-get update
+      sudo apt-get install podman -y
+
+      # Setup registries
+      echo '[registries.search]' > tmp
```
haircommander: we could consolidate these with printf

haircommander:
and @rhatdan rejoices!
one nit, otherwise LGTM

mrunalp:
This upstream PR has broken the e2e for us https://github.com/kubernetes/kubernetes/pull/79667

saschagrunert:
/hold 
until podman is capable of IPv6

rhatdan:
@mheon PTAL

.circleci/config.yml
saschagrunert: Fixed :)

haircommander:
/hold
wait was I not supposed to cancel that hold @saschagrunert

saschagrunert:
Unfortunately I have to revisit this one when podman supports IPv6, since we now use it in our integration tests. I will reopen this if the time is right.

PR #2687, opened 2019-07-31
Make integration tests ready for IPv6

saschagrunert:
The current hostport manager of Kubernetes does not support IPv6
addresses, so we have to skip these to avoid sandbox setup failures.

The test configuration has been changed to add an IPv6 CIDR beside the
already existing IPv4 one. To let this work, we now have to enable IPv6
support via `--sysctl net.ipv6.conf.all.disable_ipv6=0` on docker based
runs. All test paths have been adapted for checking the new IPv6 CIDR,
whereas the pod IP parsing functions have been simplified as well.

Relates to: https://github.com/kubernetes/kubernetes/issues/80810


rhatdan:
Interesting difference here between Docker and Podman.
```
#  podman run fedora cat /proc/sys/net/ipv6/conf/all/disable_ipv6 
0
# docker run fedora cat /proc/sys/net/ipv6/conf/all/disable_ipv6 
1
```

saschagrunert:
> Interesting difference here between Docker and Podman.
> 
> ```
> #  podman run fedora cat /proc/sys/net/ipv6/conf/all/disable_ipv6 
> 0
> # docker run fedora cat /proc/sys/net/ipv6/conf/all/disable_ipv6 
> 1
> ```

I think docker/libnetwork disables it somehow, because I cannot find anything related to that within podman/libpod. :thinking:

Makefile
```diff
@@ -191,15 +191,18 @@ integration: ${GINKGO} crioimage
...
 		-e TEST_USERNS \
-		-t --privileged --rm \
+		-it --privileged --rm \
 		-v $(shell pwd):/go/src/${PROJECT} \
 		-v ${GINKGO}:/usr/bin/ginkgo \
 		-w /go/src/${PROJECT} \
-		${CRIO_IMAGE} make localintegration ${TESTFLAGS}
+		--sysctl net.ipv6.conf.all.disable_ipv6=0 \
+		${CRIO_IMAGE} \
+		make localintegration
```
TomSweeneyRedHat: should the `${TESTFLAGS}` still be here at the end?

saschagrunert: No we don't have to set it again, tested it locally :)

mrunalp:
Is this ready to be merged?

saschagrunert:
> Is this ready to be merged?

Yes, there is nothing blocking it any more. :)

PR #2688, opened 2019-07-31
Vendor in changes made to hooks monitoring in libpod

samc24:
Changes to vendor in (github.com/containers/libpod/pull/3580):
Improved hooks monitoring to work for specific edge cases with a simpler solution.
Re-reads hooks directories after any changes are detected by the watchers.
Added monitoring test for adding a different invalid hook to primary directory.
Some issues with prior code:
- ReadDir would stop when it encounters an invalid hook, rather than registering an error but continuing to read the valid hook.
- Wouldn’t account for Rename and Chmod events.
- After doing a mv of the hooks file instead of rm, it would still think the hooks file is in the directory, but it has been moved to another location.
- If a hook file was renamed, it would register the renamed file as a separate hook and not delete the original, so it would then execute the hook twice - once for the renamed file, and once for the original name which it did not delete.


umohnani8:
/test integration_fedora
changes LGTM

haircommander:
hey @samc24 you'll have to call `make vendor` and commit all the changes, then we should be good to merge here :)

samc24:
Closing in favor of  #2720, which already vendored the changes in

PR #2698, opened 2019-08-05
Update end-to-end tests to run with Kubernetes release branches

saschagrunert:
This should fix the failling e2e tests. I'd like to suggest to stick to the release branches since the [CI signal](https://github.com/orgs/kubernetes/projects/11) seems to change a lot lately.

We could add a end-to-end test to the master which is allowed to fail, but I'm not sure if this is easily achievable with prow. WDYT?

### DIFF
--- a/contrib/test/integration/main.yml
+++ b/contrib/test/integration/main.yml
@@ -118,7 +118,7 @@
       include: "build/kubernetes.yml"
       vars:
         force_clone: true
-        k8s_git_version: "master"
+        k8s_git_version: "release-1.15"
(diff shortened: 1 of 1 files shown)


rhatdan:
I hate to loose the testing on experimental branch of K8s, because the workload when they branch could become large.
@mrunalp WDYT?

haircommander:
I agree with @rhatdan, as well as think the issues we're seeing now would have to be dealt with later anyway. I intend on digging into the e2e failures this week.

saschagrunert:
Okay :+1:

Issue #2713, opened 2019-08-12
Installing cri-o on Fedora 30 based on documentation fails


adelton:
**Description**

The page https://cri-o.io/ says

> Fedora: Available on all supported Fedora versions 28, 29 and Rawhide.
> `dnf install cri-o`

The page https://github.com/cri-o/cri-o/blob/master/README.md says

> Fedora, CentOS, RHEL, and related distributions: `sudo yum install crio`

However, on Fedora 30 and rawhide, module needs to be installed first, to be able to get `cri-o`.

**Steps to reproduce the issue:**
1. on Fedora 30, run `dnf install -y cri-o`

**Describe the results you received:**

```
Fedora Modular 30 - x86_64                      3.0 MB/s | 2.7 MB     00:00    
Fedora Modular 30 - x86_64 - Updates            3.6 MB/s | 2.4 MB     00:00    
Fedora 30 - x86_64 - Updates                     13 MB/s |  19 MB     00:01    
Fedora 30 - x86_64                               25 MB/s |  70 MB     00:02    
No match for argument: cri-o
Error: Unable to find a match: cri-o
```

**Describe the results you expected:**

```
Dependencies resolved.
================================================================================
 Package                     Arch   Version                Repository      Size
================================================================================
Installing:
 cri-o                       x86_64 2:1.13.0-1.gite8a2525.module_f30+3068+5f412e7f
                                                           fedora-modular  11 M
Installing dependencies:
 container-selinux           noarch 2:2.111.0-1.fc30       updates         48 k
 containernetworking-plugins x86_64 0.8.1-1.fc30           updates         19 M
 containers-common           x86_64 1:0.1.37-2.fc30        updates         48 k
 runc                        x86_64 2:1.0.0-93.dev.gitb9b6cc6.fc30
                                                           updates        2.5 M
 selinux-policy              noarch 3.14.3-43.fc30         updates        126 k
 selinux-policy-targeted     noarch 3.14.3-43.fc30         updates         14 M
 libbsd                      x86_64 0.9.1-3.fc30           fedora         100 k
 libnet                      x86_64 1.1.6-17.fc30          fedora          61 k
 libnl3                      x86_64 3.4.0-8.fc30           fedora         286 k
 libselinux-utils            x86_64 2.9-1.fc30             fedora         147 k
 policycoreutils             x86_64 2.9-1.fc30             fedora         193 k
 protobuf-c                  x86_64 1.3.1-2.fc30           fedora          33 k
[...]

PR #2714, opened 2019-08-12
Update fedora installation instructions

haircommander:
fixes: https://github.com/cri-o/cri-o/issues/2713

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,9 @@ up a PR and add it to the list.
 ### Installing CRI-O
 To install CRI-O, you can use your distributions package manager:
 
-Fedora, CentOS, RHEL, and related distributions:
+Fedora 30 and later:
+```sudo dnf module install cri-o```
+Fedora 29, CentOS, RHEL, and related distributions:
 ```sudo yum install crio```
 openSUSE:
 ```sudo zypper install cri-o```


README.md
```diff
@@ -97,7 +97,9 @@ up a PR and add it to the list.
 ### Installing CRI-O
 To install CRI-O, you can use your distributions package manager:
 
-Fedora, CentOS, RHEL, and related distributions:
+Fedora 30 and later:
+```sudo dnf module install cri-o```
```
TomSweeneyRedHat: Did yum go away onF30?  @cgwalters  tries to keep us in line with yum....

mrunalp: Fedora should be using dnf.

haircommander: yum does in fact work though!

mrunalp: yeah it is a link to dnf. dnf should be preferred in newer versions of Fedora.
```
$ ls -l /usr/bin/yum
lrwxrwxrwx. 1 root root 5 Jul 23 00:07 /usr/bin/yum -> dnf-3
```

adelton:
Running merely
```
sudo dnf module install cri-o
```
as the commit suggests fails on Fedora 30:
```
Cannot enable more streams from module 'cri-o' at the same time
Unable to resolve argument cri-o
Error: Problems in request:
```

The module needs to be versioned ... which is sort of the problem because on each OS version, the module version will be different.

PR #2719, opened 2019-08-13
Add integration test to make sure annotations are passed through to the runtime

### LINKED ISSUE #2402: Add support for passing sandbox annotations to runtime
Recently I added a feature in `containerd` that enabled setting up a filter in `CRI/Containerd` to pass specific kubernetes pod/deployment annotations to OCI runtime annotations. 

OCI annotations can be used to set the sandbox related parameters. In the case of VM based runtimes, such as kata, this can be used to dynamically select the `vmdisk` or `initrd` that can be custom built for that container image.

 e.g. With ppc64le ultravisors' upcoming capability to run secure VM which need to be encrypted with the public key of target system, user can specifically intruct the runtime, like kata, to use a specific initrd on the system for the given container image. 

Kata already supports setting those values via OCI annotations, it's just that until now CRI wasn't setting it. The corresponding k8s pod.yaml looks something like this,

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: foobar-kat3
  "annotations": {
    "com.github.containers.virtcontainers.InitrdPath" : "/usr/share/kata-containers/custom-secure-vm.img",
    "com.github.containers.virtcontainers.InitrdHash" : "3971EBF11A18BCE250A282E6FF78739C9A0F7BCB9BC3552FE9CD8191E4A4DA573BB38983A47B2A86A63BFC4E41D3361D5D6AE1A1480BF3F27F730CC75FBC84DE" 
 }
spec:
  runtimeClassName: kataclass
  containers:
  - name: redis
    image: redis:latest 
```

more info on this work can be found in the containerd repo, https://github.com/containerd/cri/pull/1084

I believe this feature should also exist in `CRI-O`. 

N/A

N/A

N/A

Please check https://github.com/containerd/cri/pull/1084

N/A

N/A

samc24:
Added integration test to make sure annotations are passed through to the runtime. 
Fixes #2402.

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -1497,3 +1497,20 @@ function teardown() {
 	echo "$output"
 	[ "$status" -eq 0 ]
 }
+
+@test "annotations passed through" {
+	start_crio
+	run crictl runp "$TESTDATA"/sandbox_config.json
+	echo "$output"
+	[ "$status" -eq 0 ]
+	pod_id="$output"
+	run crictl inspectp $pod_id | run grep '"owner": "hmeng"'
+	run crictl inspectp $pod_id | run grep '"security.alpha.kubernetes.io/seccomp/pod": "unconfined"'
+	run crictl stopp "$pod_id"
+	echo "$output"
+	[ "$status" -eq 0 ]
+	run crictl rmp "$pod_id"
+	echo "$output"
+	[ "$status" -eq 0 ]
+	stop_crio
+}


samc24:
@umohnani8 @mrunalp PTAL :)

server/sandbox_run_test.go
```diff
@@ -300,4 +300,43 @@ var _ = t.Describe("RunPodSandbox", func() {
 			Expect(res).To(Equal(""))
 		})
 	})
+
+	t.Describe("PodAnnotationPassthroughContainerSpec", func() {
+		// Given
+		g, err := generate.New("linux")
+		Expect(err).To(BeNil())
```
saschagrunert: I think this has to be wrapped in an `BeforeEach` block of ginkgo to work, like this:
```suggestion
		var g generate.Generator

		BeforeEach(func() {
			var err error
			g, err = generate.New("linux")
			Expect(err).To(BeNil())
		})
```

saschagrunert: Hm, now the test seems to work technically, but they still fail :innocent:

saschagrunert: I think it is really not easy to test it via unit-testing because of the high complexity of the sandbox and container run methods. I recommend to add an integration test which uses custom pod annotations via the CLI and verifies that they exist on the corresponding sandbox and container. WDYT?

samc24:
@umohnani8 @mrunalp seems like the existing code already passes **all** annotations through [here](https://github.com/cri-o/cri-o/blob/b5f06157a872e338f971d64a3fd879ed169dd46e/server/sandbox_run_linux.go#L411). If I remove those lines, the passthrough works as expected, so is that still something we want - or do we just want to pass them all without filtering? 
(will work on fixing integration test too)

mrunalp:
@samc24 We should keep what we have.

umohnani8:
@saschagrunert @haircommander any idea why circle CI is failing on the grep line with an error deleting the network? The tests pass fine when run locally.

test/ctr.bats
```diff
@@ -1497,3 +1497,20 @@ function teardown() {
...
 	[ "$status" -eq 0 ]
 }
+
+@test "annotations passed through" {
+	start_crio
+	run crictl runp "$TESTDATA"/sandbox_config.json
+	echo "$output"
+	[ "$status" -eq 0 ]
+	pod_id="$output"
+	grep -E "\"owner\": \"hmeng\"" "$TESTDIR/crio-run/overlay-containers/$pod_id/userdata/config.json"
```
saschagrunert:
```suggestion
	grep -E "\"owner\": \"hmeng\"" "$TESTDIR/crio-run/vfs-containers/$pod_id/userdata/config.json"
```

test/ctr.bats
```diff
@@ -1497,3 +1497,20 @@ function teardown() {
...
+	grep -E "\"security.alpha.kubernetes.io/seccomp/pod\": \"unconfined\"" "$TESTDIR/crio-run/overlay-containers/$pod_id/userdata/config.json"
```
saschagrunert:
```suggestion
	grep -E "\"security.alpha.kubernetes.io/seccomp/pod\": \"unconfined\"" "$TESTDIR/crio-run/vfs-containers/$pod_id/userdata/config.json"
```

test/ctr.bats
saschagrunert: To avoid the issue with the changing storage drivers, we could do something like:
```
> crictl inspectp $pod_id | grep -q '"owner": "hmeng",'
```

samc24: Thanks! Should be good to go now

rhatdan:
This seems to have gotten lost in the weeds.
@samc24 Could you rebase and we can get this merged.

PR #2722, opened 2019-08-14
test: randomize cni config

haircommander:
to prevent parallel bats runs (with parallel crio's running) interfere on the network


rhatdan:
I think you also need to make sure they use different subnets...

rhatdan:
@haircommander Needs a rebase?  Or close?

haircommander:
yeah I think this can be closed

PR #2724, opened 2019-08-14
[1.14] Choose IP similarly to kubelet

haircommander:
Instead of failing when we can't find an IP unambiguously, choose an IP using the following heuristic:
1. configured value is chosen if set
2. legacy way of calling ChooseHostIP (this is the last step in the kubelet, but we want to keep behavior the same)
3. attempt to use the hostname to find it
4. loop through available interfaces, filter out ones the kubelet would, and choose the first available.

for the 4th step, these interfaces are ordered by when they're configured on operating system level, and not alphabetically.
This is good, because we don't want a CNI network preceeding one that will connect us to the net just because the former is 9.x... and latter 10.x...

Ultimately, we want to rely on ChooseHostIP as much as possible, but now have fallbacks in case a default route isn't set

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -381,39 +381,12 @@ func New(ctx context.Context, config *Config) (*Server, error) {
 	s.restore()
 	s.cleanupSandboxesOnShutdown(ctx)
 
-	hostIP := net.ParseIP(config.HostIP)
-	if hostIP == nil {
-		// First, attempt to find a primary IP from /proc/net/route deterministically
-		hostIP, err = knet.ChooseBindAddress(nil)
-		if err != nil {
-			// if that fails, check if we can find a primary IP address unambiguously
-			allAddrs, err2 := net.InterfaceAddrs()
-			if err2 != nil {
-				return nil, errors.Wrapf(err, "Failed to read InterfaceAddrs after failing to choose bind address")
-			}
-
-			// we are hoping for exactly 1 possible IP
-			numPossibleIPs := 0
-			// adapted from: https://stackoverflow.com/a/31551220
-			for _, addr := range allAddrs {
-				if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
-					if ipnet.IP.To4() != nil {
-						numPossibleIPs++
-						hostIP = ipnet.IP
-						if numPossibleIPs > 1 {
-							break
-						}
-					}
-				}
-			}
-
-			// If there is no clear primary IP, we should fail
-			if numPossibleIPs != 1 {
-				return nil, errors.Wrapf(err, "Failed to find one IP address after failing to choose bind address")
-			}
-		}
+	hostIP, err := s.getHostIP(config.HostIP)
+	if err != nil {
+		return nil, err
 	}
 
+	logrus.Infof("Choose host IP %s", hostIP)
 	bindAddress := net.ParseIP(config.StreamAddress)
 	if bindAddress == nil {
 		bindAddress = hostIP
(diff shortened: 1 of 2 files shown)


haircommander:
if this is approved, I'll forward port to 1.15

server/server.go
```diff
@@ -381,39 +381,10 @@ func New(ctx context.Context, config *Config) (*Server, error) {
...
-							break
-						}
-					}
-				}
-			}
-
-			// If there is no clear primary IP, we should fail
-			if numPossibleIPs != 1 {
-				return nil, errors.Wrapf(err, "Failed to find one IP address after failing to choose bind address")
-			}
```
TomSweeneyRedHat: I don't see this check for the numPossibleIPs != 1 anymore, is that no longer a concern or do I need another cup of tea?

haircommander: yeah the main concern there was crio choosing a different ip than the kubelet (so we wanted crio to choose on with no ambiguity). turns out that's not really necessary, and the kubelet chooses the first that works

TomSweeneyRedHat:
Thx for the 411 @haircommander .
LGTM if you can make the tests happy.

server/utils.go
```diff
@@ -279,3 +280,24 @@ func getUlimitsFromConfig(config Config) ([]ulimit, error) {
 func translateLabelsToDescription(labels map[string]string) string {
 	return fmt.Sprintf("%s/%s/%s", labels[types.KubernetesPodNamespaceLabel], labels[types.KubernetesPodNameLabel], labels[types.KubernetesContainerNameLabel])
 }
+
+// Validate given hostIP IP belongs to the current host
+// adapted from github.com/kubernetes/kubernetes/pkg/kubelet/kubelet_node_status.go
+func validateHostIP(hostIP net.IP) error {
+	if hostIP.To4() == nil {
```
saschagrunert: I think IPv6 will be allowed with 1.16.0 too, just to consider for the forward-port.

server/utils.go
```diff
@@ -279,3 +280,24 @@ func getUlimitsFromConfig(config Config) ([]ulimit, error) {
 func translateLabelsToDescription(labels map[string]string) string {
 	return fmt.Sprintf("%s/%s/%s", labels[types.KubernetesPodNamespaceLabel], labels[types.KubernetesPodNameLabel], labels[types.KubernetesContainerNameLabel])
 }
+
+// Validate given hostIP IP belongs to the current host
+// adapted from github.com/kubernetes/kubernetes/pkg/kubelet/kubelet_node_status.go
+func validateHostIP(hostIP net.IP) error {
```
saschagrunert: I recommend to return a bool here if we do not do anything with the error message. The we could prefix the function with `is`, like `isValidHostIP`.

server/server.go
```diff
@@ -465,6 +436,42 @@ func New(ctx context.Context, config *Config) (*Server, error) {
 	return s, nil
 }
 
+func (s *Server) getHostIP(configIP string) (net.IP, error) {
+	// emulate kubelet behavior of choosing hostIP
+	// ref: k8s/pkg/kubelet/nodestatus/setters.go
+
+	// use configured value if set
+	if hostIP := net.ParseIP(configIP); hostIP != nil {
```
saschagrunert: I would error here in case it is not parseable to give the user the feedback that CRI-O is misconfigured. Or validate this one in the config validation part of the `config` package. (At least in the forward-port)

server/utils.go
```diff
@@ -279,3 +280,24 @@ func getUlimitsFromConfig(config Config) ([]ulimit, error) {
...
+	if hostIP.To4() == nil {
+		return fmt.Errorf("hostIP must be a valid IP address")
+	}
+	if hostIP.IsLoopback() {
```
saschagrunert: Not? I’m wondering why 127.0.0.1 should not work...

server/utils.go
haircommander: Okay, I'll add it in the forward port :)

server/utils.go
haircommander: We want to choose an IP with as close a heuristic as the kubelet as possible. the kubelet filters loopback, so I did. I can't really think of a reason it wouldn't work though, but I think I'd like to keep it there.

haircommander:
I don't think we want this anymore
/close

PR #2728, opened 2019-08-15
Remove unnecessary store nil unit test

saschagrunert:
This test is not possible to succeed any more, whereas the test result is
wrong when running as `root`.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #2730, opened 2019-08-16
Privileged restrict devices

### LINKED ISSUE #2708: Add flag to overload default privileged behaviour to pass/not pass host devices.
See the discussion on : https://github.com/kata-containers/runtime/issues/1568#issuecomment-519475588

Passing all the host devices in priveleged mode is a security issue. There has been a discussion going on in CRI and docker to provide a flag to disable this behaviour. 
We need a similar flag in cri-o as well.

Eventually, the plan is to propose this as the Runtime level in k8s, with each runtime defining its own scope for Privileged.

amshinde:
oci: Introduce new runtime field to restrict devices in privileged mode
    
Add a new field "privileged_without_host_devices" to the RuntimeHandler
so that runtimes can restrict passing host devices when a container
is run as privileged.

Fixes #2708


amshinde:
Opening this for review, plan to add tests soon.

internal/lib/container_server.go
```diff
@@ -31,7 +31,7 @@ import (
 
 // ContainerServer implements the ImageServer
 type ContainerServer struct {
-	runtime              oci.RuntimeImpl
+	runtime              *oci.Runtime
```
sboeuf: I agree with this change, but I realized that @saschagrunert introduced the change from `*Runtime` into `RuntimeImpl` some time ago with this commit 109d3709c1085bbf3ecff12b7186ac31b3655996.
@saschagrunert is there any reason why you did that in the first place? Just want to make sure we're not potentially breaking something here.

saschagrunert: Yes, the major reason was to have something where the behavior is mock’able, which are in golang only public interfaces.

Do you think it would be possible to extend the interface by adding a `PrivilegedWithoutHostDevices()` method?

sboeuf: @saschagrunert 
>Yes, the major reason was to have something where the behavior is mock’able, which are in golang only public interfaces.
Do you think it would be possible to extend the interface by adding a PrivilegedWithoutHostDevices() method?

I see. But the interface was not created for this purpose and I'm afraid that we make code design decisions based on the capacity of mocking or not.
If we add `PrivilegedWithoutHostDevices()` to `RuntimeImpl` interface, this means `RuntimeVM` and `RuntimeOCI` implementations of this interface should also implement this new method. But that's not the goal. We want `Runtime` to remain a structure that provides the mapping into `RuntimeImpl` implementations. `Runtime` should not be seen as an implementation of `RuntimeImpl` IMO.

I think that if we really want to do proper mocking, we should add an extra abstraction layer dedicated for this, which `Runtime` would implement. But that sounds like a lot of indirection layers for the purpose of mocking.

amshinde: @saschagrunert Agree with @sboeuf there. Making this as an interface when it is not really required from an implementation point of view, kind of makes the code not very intuitive. 
This would also mean adding PrivilegedWithoutHostDevices() implementations for `RuntimeVM` and `RuntimeOCI`, when they are really not required or called.

internal/oci/oci.go
```diff
@@ -172,22 +172,43 @@ func (r *Runtime) WaitContainerStateStopped(ctx context.Context, c *Container) (
...
 			return nil, err
 		}
 
 		rh = runtimeHandler
 	}
 
+	return rh, nil
+}
+
+// PrivelegedRuntimeHandler returns a boolean value  configured for the
```
haircommander: nit: double space between value and configured

haircommander:
LGTM in general. One conceptual question though:
Why is it we store this value in oci? I would almost rather have it configured on the server level. I don't see a reason to involve oci, especially because the implementation doesn't differ between runtime_vm and runtime_oci. Unless you think there would be users who want to set different values for vm/oci? it seems like a setting that users would set for all or none to me

haircommander:
thanks for taking this on by the way @amshinde !

amshinde:
> Why is it we store this value in oci? I would almost rather have it configured on the server level. I don't see a reason to involve oci, especially because the implementation doesn't differ between runtime_vm and runtime_oci. Unless you think there would be users who want to set different values for vm/oci? it seems like a setting that users would set for all or none to me

@haircommander This field is specific to a Runtime. It could differ between different runtime_oci instances as well. Eventually, plan is to have this as part of the CRI api itself.
For example, with this change you could configure the runtimes in `/etc/crio/crio.conf` as:
```
[crio.runtime.runtimes.runc]
    runtime_path = "/usr/local/bin/crio-runc"
    runtime_type = "oci"
    runtime_root = "/run/runc"
    privileged_without_host_devices = false
[crio.runtime.runtimes.kata]
    runtime_path = "/usr/bin/kata-runtime"
    runtime_type = "oci"
    privileged_without_host_devices = true
[crio.runtime.runtimes.kata-qemu]
    runtime_path = "/opt/kata/bin/kata-qemu"
    runtime_type = "vm"
```

server/container_create_linux_test.go
```diff
@@ -73,3 +75,83 @@ func TestAddOCIBindsForSys(t *testing.T) {
 		t.Error("there is not a single /sys bind mount")
 	}
 }
+
+func TestAddDevicesPlatformPrivileged(t *testing.T) {
```
amshinde: @haircommander @saschagrunert I have added this unit test which checks if host devices are added for privileged containers , and have been trying to run the test with as below, but getting following errors:
```
go test -v -run TestAddDevicesPlatform 
# github.com/cri-o/cri-o/server [github.com/cri-o/cri-o/server.test]
./inspect_test.go:60:12: container.SetState undefined (type *oci.Container has no field or method SetState)
./inspect_test.go:172:12: container.SetState undefined (type *oci.Container has no field or method SetState)
FAIL    github.com/cri-o/cri-o/server [build failed]
```
What is the recommended way to run the tests? 
I tried running `sudo -E GOPATH=$GOPATH make testunit`, but seeing other unrelated errors. Does the mock interface issue needs to be addressed for that ?

internal/lib/container_server.go
saschagrunert: I think testability should be taken into consideration when designing an API, but that's not the key problem here. If we decide to leave the implementation as it currently is, then we have to remove this method which does not work any more:

https://github.com/cri-o/cri-o/blob/68070024cf96caaf252fa9b87265f8f20664bb06/internal/lib/container_server_test_inject.go#L22-L25

Afterwards, all relevant tests using this method (see `internal/lib` and `server` packages) have to be removed because they do not work any more too and I have to find another way to test these methods...maybe with a fake runtime but I'm not sure.

server/container_create_linux_test.go
saschagrunert: Yes, the tests do not work any more, see: https://github.com/cri-o/cri-o/pull/2730/files#r315060012

internal/lib/container_server.go
amshinde: @saschagrunert @sboeuf @haircommander @mrunalp 
Let me know what you think is the best way to go about this. I am going on vacation in a couple of days, and would like to get this resolved and possibly merged before that.

AkihiroSuda:
What is the next step?

internal/lib/container_server.go
saschagrunert: To unblock this and bring it forward I'm still fine with removing the tests for now. :)

saschagrunert:
> What is the next step?

I think fixing-up the CI pipeline and then we can continue our review. :)

internal/lib/container_server.go
amshinde: @saschagrunert Back now, I can remove the tests for now. Can you let me know what are the tests that I need to remove?

amshinde:
@saschagrunert I have fixed the unit tests and included your patch on top of my changes. PTAL.
The CI needs to run now.

internal/lib/config/config.go
```diff
@@ -140,9 +140,10 @@ type RootConfig struct {
 // RuntimeHandler represents each item of the "crio.runtime.runtimes" TOML
 // config table.
 type RuntimeHandler struct {
-	RuntimePath string `toml:"runtime_path"`
-	RuntimeType string `toml:"runtime_type"`
-	RuntimeRoot string `toml:"runtime_root"`
+	RuntimePath                  string `toml:"runtime_path"`
+	RuntimeType                  string `toml:"runtime_type"`
+	RuntimeRoot                  string `toml:"runtime_root"`
+	PrivilegedWithoutHostDevices bool   `toml:"privileged_without_host_devices"`
```
saschagrunert: Can we please add documentation about this flag? I'm also not sure if the naming is that concrete. What do you think about something like `mount_host_devices_on_privileged_containers`?

saschagrunert:
> @saschagrunert I have fixed the unit tests and included your patch on top of my changes. PTAL.
> The CI needs to run now.

Hey, I just have a single comment (see above)

amshinde:
@saschagrunert I have added a documentation comment for the field.
I also kept the name `privileged_without_host_devices` as is, as this is the name we have used in PRs submitted to docker and containerd for similar functionality. I just wanted to keep the naming consistent from a user point of view. Let me know if you think otherwise.

amshinde:
>     * [ ]  I think an integration test would be good, but we could add it later on
> 
>     * [ ]  The configuration template would need additional documentation, but we can add this later on too

Thanks @saschagrunert for all your help.
I can certainly address the above in a follow up PR.

amshinde:
@mrunalp Can you take a look, the CI needs to triggered I think.

amshinde:
@haircommander Looks like the fedora test has failed, can we rerun it?

amshinde:
I think this is ready to merge. 
This PR removes some existing tests(hence the reduction in the test coverage) as concrete struct `Runtime` was introduced to replace the `RuntimeImpl` interface added earlier for testing. Tests need to be added in the future taking this into consideration.

rhatdan:
@haircommander @mrunalp @umohnani8 PTAL

amshinde:
Any updates?

haircommander:
Sorry for the delay, i'll review tomorrow. It'll also fail against CI, we're working on fixing another flake now.

haircommander:
/lgtm
thanks for sticking with this @amshinde!

haircommander:
/retest
(also e2e*fedora will fail not at the fault of this PR)

amshinde:
@haircommander @saschagrunert Any updates on the status of this PR?

haircommander:
@amshinde sorry, we're still fighting CI fires. I would be fine merging this manually as is. WDYT @saschagrunert @mrunalp @umohnani8

saschagrunert:
Yeah I agree and I think we have to do as well:

/cherrypick release-1.16

haircommander:
hey @amshinde can you rebase to pick up CI fixes?

amshinde:
@haircommander Missed this. I have rebased the PR.

saschagrunert:
@amshinde thanks, can you please fix the single linting issue which is now reported? :)

saschagrunert:
This may need a backport to 1.16

amshinde:
Wondering what is the status of this PR.

saschagrunert:
@amshinde may I ask you to rebase to resolve the merge conflicts?

amshinde:
Done @saschagrunert

amshinde:
Checking in again @saschagrunert @haircommander @mrunalp @umohnani8

umohnani8:
LGTM
would like a final approval from either @haircommander @mrunalp

PR #2731, opened 2019-08-16
Remove deprecated implicit hooks directory handling

saschagrunert:
If `hooks_dir` is unset, CRI-O will not longer default to
`/usr/share/containers/oci/hooks.d` and `/etc/containers/oci/hooks.d`
because this behavior is depcrecated since some releases.

Documentation has been adapted as well.


TomSweeneyRedHat:
LGTM
I'm assuming this will be noted in the Release Announcement?

vrothberg:
We're seeing regressions by the changed behavior as some deployments were relying on these defaults. To be fair, CRI-O has been warning about the deprecation for almost a year but the fact we're regressing shows that warnings are not enough and that we didn't scream about this when releasing so package updates and/or config managers can properly set the `hooks_dir` in crio.conf.

This PR is just pulling the trigger of an earlier decision and doesn't deserve any blame but I still wanted to share the impact.

PR #2740, opened 2019-08-20
WIP: test reverting f22b67dd8f9f4c1ce7

haircommander:

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/contrib/test/integration/main.yml
+++ b/contrib/test/integration/main.yml
@@ -118,8 +118,8 @@
       include: "build/kubernetes.yml"
       vars:
         force_clone: true
-        k8s_git_version: "master"
-        k8s_github_fork: "kubernetes"
+        k8s_git_version: "fix-e2e-4"
+        k8s_github_fork: "haircommander"
         crio_socket: "/var/run/crio/crio.sock"
     - name: run k8s e2e tests
       include: e2e.yml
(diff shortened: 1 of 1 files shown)


haircommander:
/retest
had to drop 2c79ffe274c3bb5a2dad38cde866425018033e10 from kubernetes because it has some weird ssh breaking change. If this passes, we now have two culprits to fix :\

saschagrunert:
> /retest
> had to drop 2c79ffe274c3bb5a2dad38cde866425018033e10 from kubernetes because it has some weird ssh breaking change. If this passes, we now have two culprits to fix :\

Which one? I’d like to help you out :)

haircommander:
https://github.com/kubernetes/kubernetes/pull/81163 is the culprit of the ssh failures on the last run
https://github.com/kubernetes/kubernetes/pull/81305 seems to be the problem we've been seeing where the tests don't spawn (not positive, but the commit before passes and it and the one after it fails...) if this PR passes, then those are our two issues

mrunalp:
@saschagrunert @haircommander It looks like we need to setup KUBE_SSH_USER for these tests in the e2e.

mrunalp:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/ssh/ssh.go#L163 We can create a keypair and setup root to be able to ssh locally.

haircommander:
testing again for a sanity check

haircommander:
this particular issue is not present anymore

Issue #2725, opened 2019-08-15
crio-wipe fails if /var/lib/containers is a mountpoint


paradon:
**Description**

crio-wipe attempts to "rm -rf /var/lib/containers", however if that path is a mountpoint it fails with "Device or resource busy".

**Steps to reproduce the issue:**
1. Mount a filesystem at /var/lib/containers
2. Cause crio-wipe to attempt an update (e.g. remove /var/lib/crio/version)

**Describe the results you received:**
crio-wipe errors with "Device or resource busy", and blocks crio startup due to systemd unit dependency.

**Describe the results you expected:**
crio-wipe should handle this situation, probably by removing the contents instead of the entire directory.

**Output of `crio --version`:**

```
crio version 1.15.1-dev
commit: "68070024cf96caaf252fa9b87265f8f20664bb06"
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Archlinux VM.

haircommander:
@paradon thanks for opening the issue. I didn't even think of that! I have a fix here: https://github.com/cri-o/cri-o/pull/2726 which once approved, I will cherry-pick to 1.15 for you

PR #2746, opened 2019-08-22
wipe only crio

haircommander:
As it stands, crio-wipe is dangerous. If a user wants to preserve their podman containers, or buildah images on a node, they won't be able to. Downloading a new CRI-O, or installing for the first time, will delete them.

Fix this by being more deft about container and image deletion. This PR includes:

- a new binary: crio-wiper, which is downloaded into the contrib/crio-wipe/crio-wiper folder. It is responsible for reading from storage directly, filtering out crio containers and images, unmounting and deleting them.
- Because we want crio-wiper to be looking at the same storage location as crio, I have factored out the argument and conf file handling to a separate library: clicommon, which both crio and crio-wiper can read from.
- Added a make target for downloading crio-wiper
- added a test suite. To do so, the existing integration test suite had to be adapted a little (it didn't use set -u). The test suite is currently pretty sparse, but it demonstrates the usage of crio-wiper.

~This PR is technically a WIP, I need to make sure the spec file downloads it, and need to test it in the wild, as well as make sure I am sure about how images should work. I'm putting it out here for review to make sure my direction is good.~ 

also fixes: https://github.com/cri-o/cri-o/issues/2725


haircommander:
this PR is ready for review @mrunalp @saschagrunert @rhatdan @umohnani8 PTAL

rhatdan:
Why make this a separate executable at all?  Why not just do crio wipe?

haircommander:
that is a great idea lol

saschagrunert:
Please feel free to ping here again if you need review since I guess it is now again WIP, right?

haircommander:
hey @saschagrunert this is pretty much ready, I took @rhatdan 's advice. I'll just need to fix the integration tests, but feel free to review whenever, I'm happy with the design and code now.

cmd/crio/cli.go
```diff
@@ -0,0 +1,571 @@
+package main
```
saschagrunert: I like that! Do we want to move this file into an internal `internal/pkg/cli` package? This would have the benefit that we could write tests for it later on.

docs/crio.8.md
```diff
@@ -61,6 +61,7 @@ crio
 [--stream-port=[value]]
 [--uid-mappings=[value]]
 [--version|-v]
+[--version-file-location=[value]]
```
saschagrunert: I would omit the `-location` here since it is obvious that it is a path.

docs/crio.conf.5.md
```diff
@@ -39,6 +39,9 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 **storage_option**=[]
   List to pass options to the storage driver. Please refer to containers-storage.conf(5) to see all available storage options.
 
+**version_file_location**="/var/lib/crio/version"
+  Location for crio to lay the version file down.
```
saschagrunert: The description should match with the one of the CLI:
```suggestion
  Location for crio to lay down the version file.
```

internal/version/version.go
```diff
@@ -1,23 +1,70 @@
...
+// If there is a parsing error, then crio should wipe, and the error is returned.
+// if parsing is successful, it compares the major and minor versions
+// and returns whether the major and minor versions are the same.
+// If they differ, then crio should wipe.
+func ShouldCrioWipe(versionFileName string) (bool, error) {
+	return shouldCrioWipe(versionFileName, Version)
+}
+
+// shouldCrioWipe is an internal function for testing purposes
+func shouldCrioWipe(versionFileName, versionString string) (bool, error) {
```
saschagrunert: I would change the return type to `bool` and only `logrus.Errorf` the errors.

haircommander: I'll do that for the public version, but I'd like to retain the errors for testing on the internal version

haircommander: on second though, I'd like to leave it as is. At this point, logrus isn't configured, and it's not included in version. I'd prefer to allow the caller to choose what to do with an error.

saschagrunert: SGTM

docs/crio.8.md
```diff
@@ -217,6 +218,8 @@ If `hooks_dir` is unset, CRI-O will currently default to `/usr/share/containers/
 
 **--version, -v**: Print the version
 
+**--version-file**="": Location for crio to lay down the version file.
```
TomSweeneyRedHat: as much as I love a good ending period ('.'), remove the one here as it doesn't appear any other line has one.

saschagrunert:
Unfortunately this needs to be rebased now.

haircommander:
@mrunalp @rhatdan @umohnani8 PTAL

rhatdan:
Can you squash your commits? You need to rebase anyways.

haircommander:
@mrunalp @rhatdan @umohnani8 PTAL

haircommander:
/retest
:crossed_fingers:

PR #2751, opened 2019-08-29
Allow usage of short container ID in ContainerStats()

adk9:
This fix allows using short container IDs in the ContainerStats() request

**- What I did**
Unlike all of the other CRI RPC calls, ContainerStats() would require using the full container ID. This change enables the short container ID to be used for the ContainerStats() call.

**- How I did it**

**- How to verify it**
Call ContainerStats() with a ContainerStatsRequest that contains a short container id. 

Allow usage of short container ID in ContainerStats() request

### DIFF
--- a/server/container_stats.go
+++ b/server/container_stats.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/cri-o/cri-o/internal/oci"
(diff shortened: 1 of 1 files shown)


server/container_stats.go
```diff
@@ -37,10 +37,10 @@ func (s *Server) ContainerStats(ctx context.Context, req *pb.ContainerStatsReque
 		recordOperation(operation, time.Now())
 		recordError(operation, err)
 	}()
-
-	container := s.GetContainer(req.ContainerId)
-	if container == nil {
-		return nil, fmt.Errorf("invalid container")
+	logrus.Debugf("ContainerStatsRequest %+v", req)
```
saschagrunert: I think this is not needed since we automatically log every RPC in:
https://github.com/cri-o/cri-o/blob/30f477fa1d5014e8afd175d4fa84cae80d3521fe/internal/pkg/log/interceptors.go#L57

TomSweeneyRedHat:
LGTM assuming happy tests
Thanks for the PR @adk9!

haircommander:
thanks for the PR @adk9 LGTM. We are currently facing CI woes. considering this passed all the other tests, I hardly think we need to waste spin cycles on the e2e tests. @rhatdan want to hit the magic red button?

PR #2753, opened 2019-08-29
Add `log_dir`/`--log-dir` validation to be absolute

saschagrunert:
If the log_dir is an relative path, then the creation of the
container/sandbox will fail during runtime. To avoid this, we now
pre-check if the path is absolute. All tests have been adapted as well.


TomSweeneyRedHat:
The changes LGTM, needs a bit more doc work and tests aren't happy.

PR #2761, opened 2019-09-02
Fix server restore to not remove podman containers

saschagrunert:
removed.

by an common podman annotation.


rhatdan:
Should we call these out as Podman containers or just as nonCRI-O Containers?  Theoretically we could end up with Buildah containers or other tools that uses containers/storage.  Error messages should just be about Non CRI-O containers.

saschagrunert:
> Should we call these out as Podman containers or just as nonCRI-O Containers? Theoretically we could end up with Buildah containers or other tools that uses containers/storage. Error messages should just be about Non CRI-O containers.

I agree, and vendored (https://github.com/containers/libpod/pull/3927) some idea about generally adding an annotation about the "container manager". I changed the error messages regarding this as well.

If https://github.com/containers/libpod/pull/3927 gets accepted, then we can continue here. If not, I will find another solution :innocent:

TomSweeneyRedHat:
Changes LGTM, but some of the tests are less than happy.

saschagrunert:
This is now ready for review since the related PR has been merged. We could also wait for libpod 1.6.0.

haircommander:
Yeah i say let's wait for the release

saschagrunert:
Closing in favor of https://github.com/cri-o/cri-o/pull/2883

Issue #2756, opened 2019-08-30
use GOPROXY to speed up dependency management


vrothberg:
I've already opened PRs for the github.com/containers/* projects using `GOPROXY=https://proxy.golang.org` which shows vendoring speed ups of factor six to ten. CRI-O could benefit from it as well but the proxy returns 410 for `golangci/*` dependencies. Once those issues are resolved, we can update the Makefile.

This issue is meant to prevent redundant work (others looking into the same issue) and to track it.

PR #2759, opened 2019-08-30
download kubetest manually

haircommander:
and set --get false, in the hopes that getting kubetest ourselves will fix CI woes

Signed-off-by: Peter Hunt <pehunt@redhat.com>

PR #2766, opened 2019-09-04
use GOPROXY to speed up dependency management

vrothberg:
Set GOPROXY=https://proxy.golang.org in the Makefile to speed up
fetching dependencies.  Note that the go modules of
golangci/golangci-lint has some invalid version specifiers, so we had to
add some replacements in our go.mod to avoid getting 410's from the
GOPROXY.

Fixes: #2756


Makefile
```diff
@@ -1,5 +1,7 @@
 GO ?= go
 
+export GOPROXY=https://proxy.golang.org
```
saschagrunert: What happens if the proxy is unavailable? Can we set it to something like `https://proxy.golang.org,direct` to still allow direct connections as fallback?

vrothberg: That's the default in go v1.13, so I don't think we need to worry about the proxy being unavailable. The change is effectively an early opt-in for go < v1.13.

Regarding the fall back, I don't know if that's valid syntax for GOPROXY. I've checked the docs but didn't see a reference for that. Do you know if that's possible?

vrothberg: `invalid $GOPROXY setting: cannot have comma` ... it's really a mutually exclusive setting. If the proxy is unavailable, it will fail.

saschagrunert: Okay, so the comma separated list seems a go1.13 feature too, LGTM :)

vrothberg:
@saschagrunert, do you know if the e2e tests are generally broken? Otherwise, I'll have a look at the failures here.

saschagrunert:
> @saschagrunert, do you know if the e2e tests are generally broken? Otherwise, I'll have a look at the failures here.

I think https://github.com/cri-o/cri-o/pull/2759 is the closest solution right now, but we see new failures which does not seem to be reproducible locally (investigating this now too).

Makefile
saschagrunert: Do we also want to set `GOSUMDB=sum.golang.org`?

vrothberg: Sounds good to me. However, it'll only take effect on go >= 1.13.

go.mod
```diff
@@ -86,3 +86,19 @@ replace (
 	k8s.io/metrics => k8s.io/kubernetes/staging/src/k8s.io/metrics v0.0.0-20190807140250-28e88e3f19c3
 	k8s.io/sample-apiserver => k8s.io/kubernetes/staging/src/k8s.io/sample-apiserver v0.0.0-20190807140250-28e88e3f19c3
 )
+
+replace github.com/go-critic/go-critic v0.0.0-20181204210945-1df300866540 => github.com/go-critic/go-critic v0.3.5-0.20190526074819-1df300866540
```
saschagrunert: Sorry to bump you again, but can we check if this is still needed with [golangci-lint v1.18.0](https://github.com/golangci/golangci-lint/releases/tag/v1.18.0)?

vrothberg: That works like a charm now. Thanks for checking!

saschagrunert:
This does not look good...

/retest

PR #2767, opened 2019-09-04
[1.14] wipe only crio

haircommander:
As it stands, crio-wipe is dangerous. If a user wants to preserve their podman containers, or buildah images on a node, they won't be able to. Downloading a new CRI-O, or installing for the first time, will delete them.

Fix this by being more deft about container and image deletion. This PR includes:

    adding `crio wipe`. It is responsible for reading from storage directly, filtering out crio containers and images, unmounting and deleting them.
    added a test suite. To do so, the existing integration test suite had to be adapted a little (it didn't use set -u).
    get rid of old contrib/crio-wipe/*, as we now entirely encapsulate crio-wipe in the crio binary


haircommander:
happy tests PTAL @umohnani8 @mrunalp @rhatdan @saschagrunert

cmd/crio/config.go
```diff
@@ -4,6 +4,7 @@ import (
 	"os"
 	"text/template"
 
+	"github.com/cri-o/cri-o/pkg/criocli"
```
saschagrunert: Can we call this package only `cli` to reduce the noise a bit? :innocent:

cmd/crio/main.go
```diff
@@ -257,234 +94,24 @@ func main() {
 	app.Usage = "crio server"
 	app.Version = strings.Join(v, "\n")
 
-	defConf, err := server.DefaultConfig()
+	var err error
+	app.Flags, app.Metadata, err = criocli.GetFlagsAndMetadata()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "error loading server config: %v", err)
+		fmt.Fprintf(os.Stderr, err.Error())
```
saschagrunert:
```suggestion
		fmt.Fprintln(os.Stderr, err)
```

cmd/crio/config.go
haircommander: `cli` would conflict with `github.com/urfave/cli`, which is imported both in main and criocli. I'd have to import `criocli "github.com/cri-o/cri-o/pkg/cli`, which i find noiser

saschagrunert: I agree :+1:

cmd/crio/main.go
haircommander: Fixed!

haircommander:
@saschagrunert @mrunalp @umohnani8 PTAL

haircommander:
@rhatdan the release-tool binary moved repos. I'll need another label :)

PR #2775, opened 2019-09-06
[1.14] support manifest lists

nalind:
Vendor in WIP branches for manifest list support to let CI exercise the changes.


rhatdan:
LGTM

mrunalp:
Thanks! The e2e are good :+1:
I think we should some integration tests to pull manifest list images by latest, tag and digest to cover those scenarios in CI.

mrunalp:
@saschagrunert ptal

test/image.bats
```diff
@@ -5,6 +5,12 @@ load helpers
 IMAGE=quay.io/crio/pause
 SIGNED_IMAGE=registry.access.redhat.com/rhel7-atomic:latest
 UNSIGNED_IMAGE=quay.io/crio/hello-world:latest
+IMAGE_LIST_DIGEST=docker.io/library/alpine@sha256:7746df395af22f04212cd25a92c1d6dbc5a06a0ca9579a229ef43008d4d1302a
+IMAGE_LIST_DIGEST_AMD64=docker.io/library/alpine@sha256:bf1684a6e3676389ec861c602e97f27b03f14178e5bc3f70dce198f9f160cce9
```
haircommander: are we sure this digest won't change?

nalind: The list's digest is affected by the contents, so these two won't unless the tag is moved, in which case the list's digest would change, but the per-arch images might not (for example, if an image for a new arch was added to the list, which would change the list's digest, and the tag was moved to point to the new list).

mrunalp:
I think we will have to ensure that we return the same image for image status and image list APIs. For e.g. if someone tries to start a pod with a manifest list digest, we will get an image pull request and then an image status request and once we confirm that we have the image, then the containers in the pod are started. Is that already the case with the changes in the PR?

mrunalp:
@nalind could you rebase this PR? We merged a test fix.

haircommander:
/test integration fedora
hit the `Device or resource busy` flake

nalind:
Hmm, go 1.11.13 on my system, both on the host and in a container, is producing a `vendor/modules.txt` that's in a different order than the CI test is producing for comparison, which is... disappointing.

haircommander:
/retest
If you get e2e-aws to pass this should be merged

PR #2781, opened 2019-09-10
WIP: Add multi hostIP support for IPv6 host networks

saschagrunert:
This is a change on top of https://github.com/cri-o/cri-o/pull/2662 to support multiple host IPs, for example in dual stack IPv6 scenarios.


saschagrunert:
Ready for review.

@haircommander we might bring #2662 in before that I can rebase on top of it.

In any case I took my freedom to label it to 1.16 as crucial fix to make the IPv6 host network scenario work.

saschagrunert:
Rebased on top of the master to get the related PR in

saschagrunert:

mrunalp:
@squeed @dcbw @danwinship ptal

rhatdan:
@squeed @dcbw @danwinship Does this look ok to you?

server/server.go
```diff
@@ -469,43 +469,81 @@ func New(
...
+		}
+	}
+
+	if rootIP == "" {
+		rootIP = net.IPv4(127, 0, 0, 1).String()
+		logrus.Warnf("unable to find a host IP, falling back to %v", rootIP)
+	}
+
+	// Search for an IPv6
+	ips := []string{rootIP}
```
squeed: This should handle the case where rootIP is v6 but the interface has v4 addresses.

server/server.go
```diff
@@ -469,43 +469,81 @@ func New(
...
+		}
 	}
 
 	// attempt to find an IP from the hostname
-	if hostname, err := os.Hostname(); err == nil {
-		if hostIP := net.ParseIP(hostname); hostIP != nil && validateHostIP(hostIP) == nil {
-			return hostIP
+	if rootIP == "" {
+		if hostname, err := os.Hostname(); err == nil {
+			if hostIP := net.ParseIP(hostname); hostIP != nil && validateHostIP(hostIP) == nil {
```
squeed: validateHostIP is v4 only. What if the hostname is a v6 IP?

server/server.go
```diff
@@ -469,43 +469,81 @@ func New(
...
 	}
 
 	// if that fails, check if we can find a primary IP address unambiguously
-	if allAddrs, err := net.InterfaceAddrs(); err == nil {
-		// adapted from: https://stackoverflow.com/a/31551220
-		for _, addr := range allAddrs {
-			if ipnet, ok := addr.(*net.IPNet); ok {
-				if validateHostIP(ipnet.IP) == nil {
-					return ipnet.IP
+	if rootIP == "" {
```
squeed: The correct fallback here is to find the IP address with the default route.

server/server.go
```diff
@@ -469,43 +469,81 @@ func New(
 	return s, nil
 }
 
-func (s *Server) getHostIP(configIP string) net.IP {
+func (s *Server) getHostIPs(configIP string) []string {
```
squeed: this should still return `[]net.IP`. IPv6 addresses can be "spelled" in many ways, so it's important to always use the byte form.

server/server.go
```diff
@@ -469,43 +469,81 @@ func New(
...
-		return hostIP
+		rootIP = hostIP.String()
 	}
 
 	// Otherwise use kubernetes utility to choose hostIP
 	// there exists the chance for both hostIP and err to be nil, so check both
-	if hostIP, err := knet.ChooseHostInterface(); err != nil && hostIP != nil {
-		return hostIP
+	if rootIP == "" {
+		if hostIP, err := knet.ChooseHostInterface(); err != nil && hostIP != nil {
```
squeed: All ChooseHostInterface does is pick the interface with the default route. If v4 is not configured on the machine, it will return a v6 address.

server/server.go
```diff
@@ -469,43 +469,81 @@ func New(
...
+			if addrs, err := iface.Addrs(); err == nil {
+				searchThisInterface := false
+				for _, addr := range addrs {
+					if ipnet, ok := addr.(*net.IPNet); ok {
+						ip := ipnet.IP
+						if ip.String() == rootIP {
+							searchThisInterface = true
+						}
+						if searchThisInterface &&
+							ip.String() != rootIP &&
```
squeed: You can just use the function `IsGlobalUnicast()`

squeed:
The function needs a bit of re-thinking. I think the correct logic is:
1. Does the kubernetes Node object already have IPs? If so, just use them. (This may be a bug in the CRI that needs an upstream re-think).
2. Otherwise, do a `A` and `AAAA` lookup of the node's hostname and use what's returned
3. Otherwise (or if step 2 returns only one family), for each family, find the address with the default route.
4. Otherwise, if the hostname is a valid v4 or v6 ip address, use that. It is very unlikely that we will reach this step, as every node should have a default route.
5. Otherwise, return the first address for each family that `IsGlobalUnicast()`. This should be very rare.

@danehans do you have any thoughts?

server/server.go
saschagrunert: Alright, fixed :heavy_check_mark:

server/server.go
saschagrunert: I added neccesary IPv6 paths.

server/server.go
saschagrunert: Alright, fixed :heavy_check_mark:

server/server.go
saschagrunert: I removed it that it also allow IPv6 addresses.

server/server.go
saschagrunert: I added the necessary code paths.

saschagrunert:
Thanks for the review @squeed, I changed the code paths that it can also handle main IPv6 addresses, PTAL.

haircommander:
hey @saschagrunert can you rebase this PR please :)

haircommander:
@danehans @squeed any comments or does it LGTY

saschagrunert:

haircommander:
huzzah happy tests LGTM, @mrunalp @umohnani8 @dcbw @squeed PTAL

squeed:
looks good to me!

squeed:
even better:
/lgtm

haircommander:
> even better:
> /lgtm

even even better:
/lgtm

uablrek:
I can confirm that this works fine;
```
# crio -v
crio version 1.17.0-dev
commit: "55ccb1e9bd8508ced18198444a56fd51daa76611"
# kubectl get pod alpine-deployment-6bdb488b88-d5tkt -o json
...
        "hostIP": "192.168.1.2",
        "phase": "Running",
        "podIP": "192.168.1.2",
        "podIPs": [
            {
                "ip": "192.168.1.2"
            },
            {
                "ip": "1000::1:c0a8:102"
            }
        ],
```

#### About using the default-route

We are trying to isolate the K8s network for security and performace reasons. That means that external traffic will go to a different network, and hence also the default-route. IMO no plugin used in k8s shall make the assumption that the default-route is on the k8s network.

Also I define a multi-hop default route using ECMP;
```
# ip ro show default
default 
        nexthop via 192.168.1.201 dev eth1 weight 1 
        nexthop via 192.168.1.202 dev eth1 weight 1 
```
I have not seen any plugin that makes assumptions on the default route to handle this case. I fairness I have not even tested `cri-o` in thie respect.

#### Specify both addresses to crio

In the future I think there will be all kind of unforseen corner-cases, so I suggest that crio allows multiple addresses to the `--host-ip`, e.g;
```
crio -log-level debug -host-ip 192.168.1.2,1000::1:c0a8:102
```
Then the user can control the addresses fully.

PR #2807, opened 2019-09-16
Fallback to 127.0.0.1 if host IP is unavailable

saschagrunert:
The `getHostIP` method does not fail any more because we now use the
local loopback address if no network is available and nothing else is
configured.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #2808, opened 2019-09-16
Add metrics to configuration file

saschagrunert:
The metrics can now being configured via the crio.conf configuration
file, too. This means we move the metrics server creation logic into
the library and adapt all necessary documentation.

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	goflag "flag"
 	"fmt"
-	"net"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
@@ -242,6 +241,12 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) (string, error) {
 	}
 	if ctx.GlobalIsSet("stream-tls-key") {
 		config.StreamTLSKey = ctx.GlobalString("stream-tls-key")
+	}
+	if ctx.GlobalIsSet("enable-metrics") {
+		config.EnableMetrics = ctx.GlobalBool("enable-metrics")
+	}
+	if ctx.GlobalIsSet("metrics-port") {
+		config.MetricsPort = ctx.GlobalInt("metrics-port")
 	}
 
 	return path, nil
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
LGTM assuming happy tests

giuseppe:
LGTM

rhatdan:
/test integration_fedora
@mrunalp @haircommander @umohnani8 PTAL

cmd/crio/main.go
```diff
@@ -547,9 +552,9 @@ func main() {
 			Usage:  "enable metrics endpoint for the server on localhost:9090",
 			EnvVar: "CONTAINER_ENABLE_METRICS",
 		},
-		cli.IntFlag{
+		cli.StringFlag{
```
haircommander: the need for this change is unclear to me. Wouldn't we want the verification of the port actually being an int? And toml can handle int values (`grpc_max_recv_msg_size` for instance)

saschagrunert: Okay, I will put it back to an int. The change was more for sake of consistency since every port we parse is of type string.

haircommander: ah, if that's the case I guess leaving it isn't problematic

haircommander: haha sorry, maybe revert it. keeping consistent and adding a verification step later makes sense to me.

saschagrunert: But now it might be possible that we can keep the command integration tests I removed before :man_shrugging:

haircommander: I'll leave it up to you :)

saschagrunert: I would leave it for now, I'd like your suggestion :heart:

docs/crio.conf.5.md
```diff
@@ -239,6 +239,15 @@ The `crio.network` table containers settings pertaining to the management of CNI
 **plugin_dirs**=["/opt/cni/bin/",]
   List of paths to directories where CNI plugin binaries are located.
 
+## CRIO.METRICS TABLE
+The `crio.metrics` table containers settings pertaining to the Prometheus based metrics retrieval.
+
+**enable_metrics**=false
+  Globally enable or disable metrics support.
+
+**metrics_port**="9090"
```
haircommander: if we leave it an int this needs to change too

saschagrunert: Changed as suggested.

PR #2809, opened 2019-09-16
Use gRPC interceptors for metrics collection

saschagrunert:
This changes the default manual metrics retrieval to an automatic gRPC
interception based one. All necessary methods have been adapted.

Please note that the metrics name now change from snake_case to
CamelCase. I'm not sure if this a problem, but did not want to spend to many
cycles on string manipulations.


server/metrics/interceptors.go
```diff
@@ -0,0 +1,38 @@
...
+			Observe(SinceInMicroseconds(operationStart))
+
+		// record error metric if occurred
+		if err != nil {
+			CRIOOperationsErrors.WithLabelValues(operation).Inc()
+		}
+
+		return resp, err
+	}
+}
```
TomSweeneyRedHat: Maybe it's pre-tea time, but how/when does this get called from say image_status.go?  Is it part of the grpc processing that you get for free?

saschagrunert: Yes, the interceptors here are registered there as part of the gRPC request processing:
https://github.com/cri-o/cri-o/pull/2809/files#diff-702431da3a2e35392cf253955d15f774R741-R744

This means this function will be called for every request, like we do now for the logging of the requests and responses in debug mode.

TomSweeneyRedHat:
Thanks for the 411 @saschagrunert .  LGTM and happy green test buttons.

umohnani8:
LGTM
@haircommander @mrunalp PTAL

PR #2834, opened 2019-09-19
Burst for bandwidth can be only 4GB

CallMeFoxie:
**- What I did** 
Set bandwidth default burst limits to 4GB instead of UInt64 Max to accomodate the same logic as per [https://github.com/containernetworking/plugins/pull/389](here)

**- How I did it**
Changing from MaxUInt64 to MaxUint32 * 8

**- How to verify it**
Compile and use with the PR with [https://github.com/containernetworking/plugins/pull/389](this) CNI plugin

Limit bandwidth burst to 4GB

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -129,11 +129,11 @@ func (s *Server) newPodNetwork(sb *sandbox.Sandbox) (ocicni.PodNetwork, error) {
 		bwConfig = &ocicni.BandwidthConfig{}
 		if ingress > 0 {
 			bwConfig.IngressRate = uint64(ingress)
-			bwConfig.IngressBurst = math.MaxUint64
+			bwConfig.IngressBurst = math.MaxUint32 * 8 // 4GB burst limit
 		}
 		if egress > 0 {
 			bwConfig.EgressRate = uint64(egress)
-			bwConfig.EgressBurst = math.MaxUint64
+			bwConfig.EgressBurst = math.MaxUint32 * 8 // 4GB burst limit
 		}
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Is there any ordering required for which changes get merged and then code vendored for the three PR's that you've spun up?  i.e. do the changes first need to get into cni, then have that vendored here with these changes?

CallMeFoxie:
The best would be to have the CNI merged first, but it shouldn't overly matter as it doesn't apparently work properly now anyways due to the casting ("Cannot convert <uint64 max> to int"), at least with kube 1.16.0/crio 1.15.2/cni plugins 0.8.2.

saschagrunert:
> The best would be to have the CNI merged first, but it shouldn't overly matter as it doesn't apparently work properly now anyways due to the casting ("Cannot convert to int"), at least with kube 1.16.0/crio 1.15.2/cni plugins 0.8.2.

Do you need these changes in CRI-O 1.15, too?

CallMeFoxie:
> > The best would be to have the CNI merged first, but it shouldn't overly matter as it doesn't apparently work properly now anyways due to the casting ("Cannot convert to int"), at least with kube 1.16.0/crio 1.15.2/cni plugins 0.8.2.
> 
> Do you need these changes in CRI-O 1.15, too?

Quite possibly since the feature is in already.

CallMeFoxie:
Thanks! :)

saschagrunert:
Anytime :blush:

PR #2835, opened 2019-09-19
Add shell completion support

saschagrunert:
This adds generic bash, fish and zsh shell completion support via code
generation. The generated code is part of the repository and can be
installed into the correct locations via `make install`.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM assuming happy tests

haircommander:
/lgtm
as a follow up, we should add a circle ci target that runs the completions and checks if the tree is dirty (similar to the lint check or conmon config check) so we keep the completions up to date.

saschagrunert:
> /lgtm
> as a follow up, we should add a circle ci target that runs the completions and checks if the tree is dirty (similar to the lint check or conmon config check) so we keep the completions up to date.

Yep, I can add this as well.

Issue #2849, opened 2019-09-24
cri-o passes (possibly) reused namespaces on network teardown


squeed:
**Description**
CRI-O tears down networking after a reboot, risking pid reuse.

**Steps to reproduce the issue:**
Sadly, reproducing this is probabilistic. It should still be easy to fix, though

1. Reboot the node. Create some containers, so they have a low pid number
2. Reboot the node again
3. Kubelet starts tearing down sandboxes that were killed because of the reboot
4. cri-o issues a CNI delete with /proc/$pid/ns/net, even though `$pid` is meaningless since the reboot.

Even if you don't get a pid collision, I was able to see pretty clearly getting a CNI DEL for a stale pid. For example, from crio logs at level Info:

```
About to add CNI network lo (type=loopback)
Got pod network &{Name:alertmanager-main-1 Namespace:openshift-monitoring ID:... NetNS:/proc/8036/ns/net Networks:[] RuntimeConfig:map[]}

-- reboot --

About to del CNI network lo (type=loopback)
Error deleting network: failed to Statfs "/proc/8036/ns/net": no such file or directory
```

This clearly shows that it is *looking* for `/proc/8036...`, and it happens to not be a process. However, reboot enough times and you will eventually lose and it will point to a running pid (but not the one started by cri-o). We typically see this in about 1-in-10 reboots.

**Describe the results you received:**
We got a CNI Delete with the netns of `/proc/<pid>/ns/net`, which is correct, except that the node was rebooted in the mean time, and `/proc/<pid>/ns/net` pointed to the root netns.

**Describe the results you expected:**
The CNI delete should be with an empty netns parameter, which signifies to the plugins that the namespace is gone and only bookkeeping operations (e.g IPAM cleanup) are to be done. CRI-O should only pass the netns parameter if it points to a known-good crio-created process that is still running.

**Output of `crio --version`:**

```
crio version 1.14.10-0.19.dev.rhaos4.2.gita86dae7.el8
```

squeed:
Another solution would be to bind-mount the network namespace to `/var/run/netns/<containerid>`, and only pass that. Then, if there's a reboot, it will be invalidated automatically. This is what rkt does.

umohnani8:
@squeed I am currently working on fixing this up in CRI-O. I am returning an empty string as the netns path but I keep getting this error when tryng to stop the network and clean up
```
level=error msg="Error deleting network: failed to Statfs "": no such file or directory"
[...]

PR #2851, opened 2019-09-24
WIP: bind mount netns to /var/run/...

haircommander:
on a reboot, the kubelet tells CRI-O to delete all stale pods. CRI-O will then try to delete the netns at /proc/<ctrpid>/ns/net. However, because CRI-O is on a node that has rebooted, all those pids are stale. At best, CNI del fails, at worst, we delete the net ns of another process.

Attempt instead to bind mount to /var/run/netns/<ctrid>, which is sufficiently unique.

Added wip because I am mostly testing if the change is this easy. if it is, we may also want to try for the other namespaces.

could fix: #2849


haircommander:
surprise! it's not that easy

PR #2854, opened 2019-09-24
[1.14] sandbox_remove: Ensure that the pod infra container exit file is removed

mrunalp:
@rphillips @haircommander @umohnani8 ptal

### DIFF
--- a/server/sandbox_remove.go
+++ b/server/sandbox_remove.go
@@ -2,6 +2,8 @@ package server
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/containers/storage"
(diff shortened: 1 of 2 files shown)


server/sandbox_remove.go
```diff
@@ -99,6 +101,9 @@ func (s *Server) RemovePodSandbox(ctx context.Context, req *pb.RemovePodSandboxR
 	if err := s.CtrIDIndex().Delete(podInfraContainer.ID()); err != nil {
 		return nil, fmt.Errorf("failed to delete infra container %s in pod sandbox %s from index: %v", podInfraContainer.ID(), sb.ID(), err)
 	}
+	if err := os.Remove(filepath.Join(s.Config().RuntimeConfig.ContainerExitsDir, podInfraContainer.ID())); err != nil && !os.IsNotExist(err) {
+		return nil, errors.Wrapf(err, "failed to remove pod infra container exit file %s", podInfraContainer.ID())
```
rphillips: I almost think this should just just be a warning, but I don't have a all the context behind it.

haircommander: I would also be inclined to make tihs (and the corresponding one for non-infra containers) to be a warning

mrunalp: The most usual case here is the file already being deleted and we cover that in the precheck.

haircommander:
e2e failures are not the fault of this PR (same with 1.13). I'll look into it in an hour or so

PR #2870, opened 2019-10-04
Add missing files to static release bundle

saschagrunert:
We now add the CNI, systemd, policy and CRI-O wipe filed to the bundle
as well. Additional documentation how to use the bundle is now included,
too.

### DIFF
--- a/bundle/Makefile
+++ b/bundle/Makefile
@@ -1,5 +1,7 @@
 PREFIX ?= $(DESTDIR)/usr/local
 ETCDIR ?= $(DESTDIR)/etc
+CONTAINERS_DIR ?= $(ETCDIR)/containers
+CNIDIR ?= $(ETCDIR)/cni/net.d
 BINDIR ?= $(PREFIX)/bin
 MANDIR ?= $(PREFIX)/share/man
 OCIDIR ?= $(PREFIX)/share/oci-umount/oci-umount.d
@@ -7,6 +9,9 @@ SELINUX ?= $(shell selinuxenabled 2>/dev/null && echo -Z)
 BASHINSTALLDIR ?= $(PREFIX)/share/bash-completion/completions
 FISHINSTALLDIR ?= $(PREFIX)/share/fish/completions
 ZSHINSTALLDIR ?= $(PREFIX)/share/zsh/site-functions
+SYSTEMDDIR ?= $(PREFIX)/lib/systemd/system
+
+DEFAULT_BINARY ?= crio-x86_64-static-musl
 
 all: install
 
@@ -15,8 +20,11 @@ install:
 	install $(SELINUX) -d -m 755 $(BASHINSTALLDIR)
 	install $(SELINUX) -d -m 755 $(FISHINSTALLDIR)
 	install $(SELINUX) -d -m 755 $(ZSHINSTALLDIR)
+	install $(SELINUX) -d -m 755 $(CONTAINERS_DIR)
+	install $(SELINUX) -d -m 755 $(CNIDIR)
 	install $(SELINUX) -D -m 755 -t $(BINDIR) bin/crio-x86_64-static-glibc
 	install $(SELINUX) -D -m 755 -t $(BINDIR) bin/crio-x86_64-static-musl
+	ln -sf $(BINDIR)/$(DEFAULT_BINARY) $(BINDIR)/crio
 	install $(SELINUX) -D -m 755 -t $(BINDIR) bin/pause-x86_64-static-glibc
 	install $(SELINUX) -D -m 755 -t $(BINDIR) bin/pause-x86_64-static-musl
 	install $(SELINUX) -D -m 644 -t $(ETCDIR) etc/crictl.yaml
@@ -27,9 +35,15 @@ install:
 	install $(SELINUX) -D -m 644 -t $(BASHINSTALLDIR) completions/bash/crio
 	install $(SELINUX) -D -m 644 -t $(FISHINSTALLDIR) completions/fish/crio.fish
 	install $(SELINUX) -D -m 644 -t $(ZSHINSTALLDIR) completions/zsh/_crio
+	install $(SELINUX) -D -m 644 -t $(CNIDIR) contrib/10-crio-bridge.conf
+	install $(SELINUX) -D -m 644 -t $(CONTAINERS_DIR) contrib/policy.json
+	install $(SELINUX) -D -m 644 -t $(SYSTEMDDIR) contrib/crio-shutdown.service
+	install $(SELINUX) -D -m 644 -t $(SYSTEMDDIR) contrib/crio-wipe.service
+	install $(SELINUX) -D -m 644 -t $(SYSTEMDDIR) contrib/crio.service
 
 .PHONY: uninstall
 uninstall:
+	rm $(BINDIR)/crio
 	rm $(BINDIR)/crio-x86_64-static-glibc
 	rm $(BINDIR)/crio-x86_64-static-musl
 	rm $(BINDIR)/pause-x86_64-static-glibc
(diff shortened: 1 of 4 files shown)


saschagrunert:
@afbjorklund, I tested the default bundle on openSUSE and it works like intended, can you give this a try too? The built bundle can be found here: https://21233-161810706-gh.circle-artifacts.com/0/bundle/crio-c65786bee.tar.gz

bundle/README.md
```diff
@@ -0,0 +1,31 @@
...
+with this bundle:
+
+- [runc][0]
+- [conmon][1]
+- [CNI plugins][2]
+
+[0]: https://github.com/opencontainers/runc
+[1]: https://github.com/containers/conmon
+[2]: https://github.com/containernetworking/plugins
+
```
TomSweeneyRedHat: does/will crun come into play here at some point?

bundle/build
```diff
@@ -59,8 +69,16 @@ for FILE in "${FILES_ETC[@]}"; do
     fi
     cp "$FILE" $TMPDIR/etc
 done
+for FILE in "${FILES_CONTRIB[@]}"; do
+    if [[ ! -f "$FILE" ]]; then
+        echo "File '$FILE' does not exist"
+        exit 1
+    fi
+    cp "$FILE" $TMPDIR/contrib
+done
```
TomSweeneyRedHat: On the fence.  Would it be better to go through all the files, noting all that can't be copied, and then exit if one or more can't be?  That way the admin could know all of the issues, rather than fix one, rerun, fix next, rerun, fix next, ...

Your call.

bundle/README.md
saschagrunert: Hm yeah I just wanted to stick to the minimal requirements. :)

bundle/build
saschagrunert: I agree and will change it as suggested.

haircommander:
LGTM, I don't think this needs to make it past e2e*fedora, we can manually merge when enough people give the :+1: 

/cherrypick release-1.15
/cherrypick release-1.16

afbjorklund:
> @afbjorklund, I tested the default bundle on openSUSE and it works like intended, can you give this a try too? The built bundle can be found here: https://21233-161810706-gh.circle-artifacts.com/0/bundle/crio-c65786bee.tar.gz

@saschagrunert :
**Will give it a try, all I wanted was some easy-to-follow step-by-step instructions on how to swap Docker for CRI-O in the official kubernetes instructions on how to install a container runtime:**

**https://kubernetes.io/docs/setup/production-environment/container-runtimes/**

Currently it "easy" to install Docker, since they offer a script to do the setup and because they totally ignore CRI and therefore doesn't "need" to install `crictl` which is mandatory for other runtimes...

* https://get.docker.com
  (https://github.com/docker/docker-install)

They also have a simple tarball to download, which contains everything you need to run (containerd, runc). So that can be used as a workaround if you are not on any of the two supported distributions.

* https://docs.docker.com/install/linux/docker-ce/binaries/
  (https://download.docker.com/linux/static/stable/x86_64/)

---

Installing CRI-O is complicated enough to make most users give up before they even try it. And then trying to use CentOS instead of Ubuntu only makes it harder, especially if it's not in common repositories.

Also, since there is no way (yet?) to load images using only CRI we also need to install `podman`...
So that makes users have to learn about both cri-tools and libpod, in addition to crio and conmon.

Swapping runc for crun would only make the situation _worse_, so needs to be included as well.

My personal goal is to get this automated for minikube, so that it can use the "generic" driver.

----

PS. Most end users are still stuck at understanding the difference between runtime and runtime:

https://www.ianlewis.org/en/container-runtimes-part-1-introduction-container-r

saschagrunert:
@afbjorklund thanks for the detailed explanations :pray: 

> > @afbjorklund, I tested the default bundle on openSUSE and it works like intended, can you give this a try too? The built bundle can be found here: https://21233-161810706-gh.circle-artifacts.com/0/bundle/crio-c65786bee.tar.gz
> 
> @saschagrunert :
> **Will give it a try, all I wanted was some easy-to-follow step-by-step instructions on how to swap Docker for CRI-O in the official kubernetes instructions on how to install a container runtime:**
> 
> **https://kubernetes.io/docs/setup/production-environment/container-runtimes/**

Interesting, I think we can improve that documentation by extending it, too.

afbjorklund:
@saschagrunert : ultimately I want it to be a one-line change, to switch over from docker to cri-o for k8s.

But it seems that still has a _long_ way to go yet, and now there’s also issues with sudo on CentOS too.

saschagrunert:
> @saschagrunert : ultimately I want it to be a one-line change, to switch over from docker to cri-o for k8s.
> 
> But it seems that still has a _long_ way to go yet, and now there’s also issues with sudo on CentOS too.

I see and I think we should continue the discussion about how to improve the switch to CRI-O from other runtimes. (maybe on slack?)

Nevertheless I think these changes here are needed to make the bundle work better on Linux 64bit systemd based systems.

haircommander:
hey @saschagrunert can you rebase this?

saschagrunert:
> hey @saschagrunert can you rebase this?

Rebased on top of the latest master :)

haircommander:
seems circle could use a retrigger, I don't have the perms :\

bundle/Makefile
```diff
@@ -1,12 +1,18 @@
...
+CNIDIR ?= $(ETCDIR)/cni/net.d
 BINDIR ?= $(PREFIX)/bin
 MANDIR ?= $(PREFIX)/share/man
 OCIDIR ?= $(PREFIX)/share/oci-umount/oci-umount.d
 SELINUX ?= $(shell selinuxenabled 2>/dev/null && echo -Z)
 BASHINSTALLDIR ?= $(PREFIX)/share/bash-completion/completions
 FISHINSTALLDIR ?= $(PREFIX)/share/fish/completions
 ZSHINSTALLDIR ?= $(PREFIX)/share/zsh/site-functions
+SYSTEMDDIR ?= $(PREFIX)/lib/systemd/system
+CRIOWIPEDIR ?= $(PREFIX)/libexec/crio/crio-wipe
```
haircommander: this is no longer needed as of https://github.com/cri-o/cri-o/pull/2746

bundle/Makefile
```diff
@@ -15,8 +21,12 @@ install:
 	install $(SELINUX) -d -m 755 $(BASHINSTALLDIR)
 	install $(SELINUX) -d -m 755 $(FISHINSTALLDIR)
 	install $(SELINUX) -d -m 755 $(ZSHINSTALLDIR)
+	install $(SELINUX) -d -m 755 $(CONTAINERS_DIR)
+	install $(SELINUX) -d -m 755 $(CNIDIR)
+	install $(SELINUX) -d -m 755 $(CRIOWIPEDIR)
```
haircommander: we can also remove this

bundle/Makefile
```diff
@@ -27,9 +37,18 @@ install:
...
 	install $(SELINUX) -D -m 644 -t $(FISHINSTALLDIR) completions/fish/crio.fish
 	install $(SELINUX) -D -m 644 -t $(ZSHINSTALLDIR) completions/zsh/_crio
+	install $(SELINUX) -D -m 644 -t $(CNIDIR) contrib/10-crio-bridge.conf
+	install $(SELINUX) -D -m 644 -t $(CONTAINERS_DIR) contrib/policy.json
+	install $(SELINUX) -D -m 644 -t $(SYSTEMDDIR) contrib/crio-shutdown.service
+	install $(SELINUX) -D -m 644 -t $(SYSTEMDDIR) contrib/crio-wipe.service
+	install $(SELINUX) -D -m 644 -t $(SYSTEMDDIR) contrib/crio.service
+	install $(SELINUX) -D -m 644 -t $(CRIOWIPEDIR) contrib/crio-wipe.bash
+	install $(SELINUX) -D -m 644 -t $(CRIOWIPEDIR) contrib/helpers.bash
+	install $(SELINUX) -D -m 644 -t $(CRIOWIPEDIR) contrib/lib.bash
```
haircommander: and these (but we still need crio-wipe.service above)

bundle/Makefile
saschagrunert: Yep, dropped it

bundle/Makefile
saschagrunert: Yes, I removed everything except the service

saschagrunert:
TODO: manually pick to relevant release branches

PR #2871, opened 2019-10-04
[1.14] restore: Filter podman containers

haircommander:
before, the restore code would find every container in container storage. This would include podman containers. However, CRI-O never actually filtered podman containers out. This caused podman containers to have their configs and storage removed, so they couldn't be inspected.

Fix this by filtering for podman containers, similarly to how `crio wipe` does. While the current method isn't beautiful or elegant, it gets the job done, and opens the door for a lees implementation dependent solution

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1758500

### DIFF
--- a/pkg/storage/utils.go
+++ b/pkg/storage/utils.go
@@ -1,0 +1,8 @@
+package storage
+
+// IsCrioContainer returns whether a container coming from storage was created by CRI-O
+// CRI-O sandboxes and containers differ from libpod container and pods because they require
+// a PodName and PodID annotation
+func IsCrioContainer(md RuntimeContainerMetadata) bool {
+	return md.PodName != "" && md.PodID != ""
+}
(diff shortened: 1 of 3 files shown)


haircommander:
I cherry picked to 1.13 already, but will wait until the approach is approved before going to the rest.

rhatdan:
LGTM

pkg/storage/utils.go
```diff
@@ -0,0 +1,8 @@
+package storage
+
+// IsCrioContainer returns whether a container coming from storage was created by CRI-O
+// CRI-O sandboxes and containers differ from libpod container and pods because they require
```
TomSweeneyRedHat: who's "they" here?  CRI-O or Podman containers?  I'd specify it.

TomSweeneyRedHat: Any worries about Buildah containers?

PR #2872, opened 2019-10-04
[1.13] restore: Filter podman containers

haircommander:
before, the restore code would find every container in container storage. This would include podman containers. However, CRI-O never actually filtered podman containers out. This caused podman containers to have their configs and storage removed, so they couldn't be inspected.

Fix this by filtering for podman containers, by filtering containers that don't have a required CRI-O annotation

**- What I did**

**- How I did it**

**- How to verify it**


rhatdan:
LGTM
/approve

PR #2875, opened 2019-10-07
[1.16] replace crio-wipe with crio wipe

haircommander:
before, crio-wipe would nuke the entire storage directory, not worrying about there being podman containers
instead, let's be more fine grained about wiping before an upgrade.

add a crio subcommand wipe, which wipes only crio storage.
update tests and docs accordingly
introduce package criocli that makes main code cleaner, and allows for sharing with wipe.go


haircommander:
@saschagrunert you're looking for https://github.com/cri-o/cri-o/pull/2746 :)

saschagrunert:
> @saschagrunert you're looking for #2746 :)

Ah thanks, ...we have to keep the order of the cherry-picks to not get lost in merge conflicts :innocent:

saschagrunert (approved):

PR #2876, opened 2019-10-07
[1.16] test: add changes required for fedora 30 AMI

haircommander:
Including removing our own debugging callback
change from `package` to a manual shell call (because `python2-dnf` is sunsetted, but we still need python2 for oct)
disable nodev for userns test
Signed-off-by: Peter Hunt pehunt@redhat.com


haircommander:
I reverted the ami switch, so this passed against the old ami and isn't testing anything. 
/hold

haircommander:
Manually tagged a new fedora image, let's hope :crossed_fingers:

PR #2881, opened 2019-10-08
Fallback to pause image's entrypoint when config is empty

saschagrunert:
The general target is to allow `pause_command = ""`, which should take the
entrypoint into account.

If the `pause_command = ""`, then its entrypoint will now be considered as
process arg, whereas the specified command in the image config will be
added as well. This keeps the backwards compatibility with images using
only `Cmd` and no `Entrypoint`.

If the `pause_command` is commented out, then it will still defaults to
`/pause`.

The logic has been refactored into a dedicated function to reduce the
cyclomatic complexity and allow proper testing.


haircommander:
can you rebase to pick up CI fixes?

saschagrunert:

giuseppe:
LGTM

PR #2883, opened 2019-10-08
[1.16] Fix server restore to not remove podman containers

saschagrunert:
The previous behavior was that CRI-O forcely tried to remove running
podman containers on startup, which only succeeded for the container
storage but not the running process. The result was was a running
container process in podman, which got broken because its storage got
removed.

This is now fixed by excluding podman containers on restore, discovered
by the new podman annotation.

A dedicated unit test has been added as well.

PR #2890, opened 2019-10-11
[1.16] Fix server restore to not remove podman containers

haircommander:
removed.

by the new podman annotation.

taking over https://github.com/cri-o/cri-o/pull/2883


umohnani8:
LGTM
/approve

haircommander:
merge me when e2e_fedora passes

haircommander:
@umohnani8 @mrunalp merge me?

PR #2899, opened 2019-10-17
[1.16] fix ami rhel7

haircommander:
the ansible version on rhel 7 is old enough such that a sufficiently complex boolean expression is interpreted as a string. This causes any supports_python3 condition to not execute.

instead, use set_facts with a when, so we properly set supports_python3

cherry-pick of https://github.com/cri-o/cri-o/pull/2898


haircommander:
:open_mouth: PTAL @umohnani8 @mrunalp @rhatdan

PR #2900, opened 2019-10-18
[1.16] Set process label to privileged when host network is true

mrunalp:
This is a regression from 1.14.

**- What I did**

**- How I did it**

**- How to verify it**


umohnani8:
LGTM

rhatdan:
What is the reason for this.  SELinux should not be interfering between two containers on the same network, or on the host network.

PR #2916, opened 2019-10-23
Update golangci lint and apply fixes

saschagrunert:
The new recommended way is to use the golangci install script,
which we now use. All new linter issues have been fixed and
additionally added linters are commented out per default.


TomSweeneyRedHat:
Tests aren't hip.  Changes LGTM, but I'm shocked by the number of vendored files removed by this.

saschagrunert:
Had to rebase on top of the latest master to resolve the conflicts

PR #2918, opened 2019-10-23
Allow CRI-O to manage IPC and UTS namespaces

haircommander:
This includes:
 - change the options `ManageNetworkNSLifecycle` to `ManageNSLifecycle`
 - add a new set of files: `internal/lib/sandbox/namespaces*` that takes care of namespace related functionality
 - create a generic NamespaceIface interface for interacting with all three kinds of namespaces
 - refactor some of runPodSandbox to reduce cyclomatic complexity

still TODO: adapt the unit tests to use IPC and UTS as well


mrunalp:
We can do this for IPC and UTS but not for PID. PID requires the first pid in the namespace to be alive. 
So, once all this is in place, we should need a pod image (pause) only when pid namespace is shared. For the other cases, just the pinned namespaces are enough.

haircommander:
Fixed!

haircommander:
I'm doing some CI runs with `manage_ns_lifecycle = true`. We probably shouldn't merge it
/hold

internal/lib/sandbox/namespaces_linux.go
```diff
@@ -11,68 +11,183 @@ import (
...
-	return &NetNs{netNS: netNS, closed: false, restored: true}, nil
+		var origNS nspkg.NetNS
+		origNS, err = nspkg.GetNS(getCurrentThreadNSPath(nsType))
+		if err != nil {
+			return
+		}
+		defer origNS.Close()
+
+		// create a new netns on the current thread
+		err = unix.Unshare(flag)
```
mrunalp: We should unshare all 3 namespaces together for performance.

haircommander:
closing in favor of https://github.com/cri-o/cri-o/pull/3048

PR #2920, opened 2019-10-24
install /etc/crio/seccomp.json in rpm for ci

lsm5:

@haircommander @umohnani8 @mrunalp


contrib/test/ci/seccomp.json
```diff
@@ -0,0 +1,774 @@
+{
```
saschagrunert: Is this profile different from the default one? If yes, what are these? :innocent:

umohnani8:
@lsm5 is this still needed?

PR #2922, opened 2019-10-25
[1.16] bump to  critools 1.16

haircommander:

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -212,7 +212,7 @@ function teardown() {
 	run crictl start "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
-	run crictl rm "$ctr_id"
+	run crictl rm -f "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
 	run crictl stopp "$pod_id"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,8 +4,8 @@ version: 2.1
 stdenv: &stdenv
   environment:
     GOCACHE: &gocache /tmp/go-build
-    IMAGE: &image quay.io/crio/circle-1.12:1.2.0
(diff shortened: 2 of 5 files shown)


rhatdan:
LGTM

rhatdan:
@umohnani8 @saschagrunert @mrunalp PTAL

umohnani8:
@haircommander do you know if this fixes https://github.com/kubernetes-sigs/cri-tools/issues/548?

haircommander:
these are work arounds for at least one of the problems reported. I think those tests have to be updated

saschagrunert:
Then we can forward port this one and close my open PR, thanks! 🙏😊

haircommander:
/retest 
(apparently, if you open a PR using the same commit but different branch, it effects both PRs)

test/ctr.bats
```diff
@@ -212,7 +212,7 @@ function teardown() {
 	run crictl start "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
-	run crictl rm "$ctr_id"
```
TomSweeneyRedHat: Sorry if a dupe comment, having fun network issues.  Although very minimal, I'd really rather not see test changes in a version bump PR unless they would break due to the bump.  I think these should be in a separate PR.

haircommander: I got this change from @saschagrunert's PR. Thought it was needed, though never checked. Let's find out :smile:

haircommander: yes, it appears without the change the tests break

TomSweeneyRedHat: Thanks for checking @haircommander, I was fine with them in, more of a reminder for the future.  But saying that, I'm puzzled that it's required, I would not have guessed that.

haircommander:
Finally! Can i have that lgtm back?

PR #2925, opened 2019-10-28
Make multiple host IPs configurable

saschagrunert:
As a follow-up of #2781, we are now able to configure ~~multiple~~ up to two host IPs via the command line interface or configuration file. The specified IPs will now be verified during config validation.

@uablrek can you test if this works for you? It is now possible to run CRI-O like
```
> crio --host-ip 192.168.178.101 --host-ip 2a02:8106:20f:a000:3815:37ed:f926:e0f4
```

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -233,7 +233,7 @@ For the bind-mount conditions, only mounts explicitly requested by Kubernetes co
 
 **--grpc-max-send-msg-size**="": maximum grpc receive message size (default: 16 * 1024 * 1024)
 
-**--host-ip**="": host IP considered as the primary IP to use by CRI-O for things such as host network IP (default: "")
+**--host-ip**="": Host IPs are the addresses to be used for the host network and can be specified up to two times (default: [])
 
 **--manage-network-ns-lifecycle**: determines whether we pin and remove network namespace and manage its lifecycle (default: false)
 
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -51,7 +51,7 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 The `crio.api` table contains settings for the kubelet/gRPC interface.
 
 **host_ip**=""
-  Host IP considered as the primary IP to use by CRI-O for things such as host network IP.
+  Host IPs are the addresses to be used for the host network. It is not possible to assign more than two addresses right now.
 
 **listen**="/var/run/crio/crio.sock"
   Path to AF_LOCAL socket on which CRI-O will listen.
--- a/internal/lib/config/config.go
+++ b/internal/lib/config/config.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -353,7 +354,7 @@ type APIConfig struct {
 	StreamTLSCA string `toml:"stream_tls_ca"`
 
 	// HostIP is the IP address that the server uses where it needs to use the primary host IP.
-	HostIP string `toml:"host_ip"`
+	HostIP []string `toml:"host_ip"`
 }
 
 // MetricsConfig specifies all necessary configuration for Prometheus based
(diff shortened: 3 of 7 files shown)


docs/crio.8.md
```diff
@@ -233,7 +233,7 @@ For the bind-mount conditions, only mounts explicitly requested by Kubernetes co
 
 **--grpc-max-send-msg-size**="": maximum grpc receive message size (default: 16 * 1024 * 1024)
 
-**--host-ip**="": host IP considered as the primary IP to use by CRI-O for things such as host network IP (default: "")
+**--host-ip**="": Host IPs are the addresses to be used for sandboxes with specified host network, can be specified multiple times (default: [])
```
TomSweeneyRedHat: It might be me pre-tea, but I'm not getting what you're trying to say here.  This is the first and only reference to "sandboxes" on this page.  Maybe:  "Host IPs are the addresses  to be used for the host network and can be specified multiple times (default: [])"

docs/crio.conf.5.md
```diff
@@ -51,7 +51,7 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 The `crio.api` table contains settings for the kubelet/gRPC interface.
 
 **host_ip**=""
-  Host IP considered as the primary IP to use by CRI-O for things such as host network IP.
+  Host IPs are the addresses to be used for sandboxes with specified host network.
```
TomSweeneyRedHat: Ditto previous comment.

docs/crio.8.md
saschagrunert: Yes, I think it would be better to rephrase it, I applied your suggestion.

docs/crio.conf.5.md
saschagrunert: I applied your suggestion for the docs.

uablrek:
Hi,

With;
```
crio -log-level debug -host-ip 192.168.1.1 -host-ip 3300::1
```
I get;
```
        "hostIP": "192.168.1.1",
        "phase": "Running",
        "podIP": "192.168.1.1",
        "podIPs": [
            {
                "ip": "192.168.1.1"
            },
            {
                "ip": "3300::1"
            },
            {
                "ip": "1000::1:c0a8:101"
            }
        ],
```
Which is correct in principle but not what K8s expects I'm afraid. I think K8s only can handle 2 addresses, one ipv4 and one ipv6.

It seems like after setting the addresses defined on the command line you go on and use the other ways of finding addresses also.

I think if the used specifies more than one "--host-ip=..." he/she has taken full responsibility for assigning the addresses and no further processing should happen.

(BTW the "3300::1" address does not even exist on an interface on the node and you are doing the right thing IMO; use it anyway. The address may apperar later oc be some vip, you can not know)

saschagrunert:
Hey @uablrek, thank you for the feedback. I adapted the implementation that we now can only up to two addresses (will be validated before CRI-O start). In that case, we do not do any further processing.

uablrek:
I must go away for ~1h. I'll test the updated patch when I get back.

uablrek:
Perfect!
```
        "hostIP": "192.168.1.4",
        "phase": "Running",
        "podIP": "192.168.1.4",
        "podIPs": [
            {
                "ip": "192.168.1.4"
            },
            {
                "ip": "3300::4"
            }
        ],
```

docs/crio.8.md
```diff
@@ -233,7 +233,7 @@ For the bind-mount conditions, only mounts explicitly requested by Kubernetes co
 
 **--grpc-max-send-msg-size**="": maximum grpc receive message size (default: 16 * 1024 * 1024)
 
-**--host-ip**="": host IP considered as the primary IP to use by CRI-O for things such as host network IP (default: "")
+**--host-ip**="": Host IPs are the addresses to be used for the host network and can be specified up to two times (default: [])
```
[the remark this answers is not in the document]
saschagrunert: Hm, I thought we would stay backwards compatible with that change, since the only thing that changes is that we can define the flag multiple times. The previous behavior should stay (and that's why I did not rename it as well).

I can for sure add a new option, or I thought to wrap a small struct around the two IPs (4 and 6) and parse it from the CLI.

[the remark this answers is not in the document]
haircommander: hm I guess because it's a CLI flag it shouldn't be a problem. I didn't realize we didn't have a config var. nevermind I am fine with this

haircommander:
LGTM

docs/crio.conf.5.md
```diff
@@ -51,7 +51,7 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 The `crio.api` table contains settings for the kubelet/gRPC interface.
 
 **host_ip**=""
-  Host IP considered as the primary IP to use by CRI-O for things such as host network IP.
+  Host IPs are the addresses to be used for the host network. It is not possible to assign more than two addresses right now.
```
TomSweeneyRedHat: FWIW, much clearer now, thanks for the touch up.

TomSweeneyRedHat:
LGTM

haircommander:
hm, needs rebase anyway. hopefully fedora is just taking a nap and isn't broken :upside_down_face:

saschagrunert:

PR #2928, opened 2019-10-29
[1.16] reduce infra container's selinux privilege

haircommander:
before, if a pod was privileged, a pause container was set as privileged. This caused the rest of the containers in the pod to also need to inherit privileged labels from the pod (so they can join the pause container's namespaces).

In reality, there's little need for the pause container to have privileged (spc_t) label. It can manage an unprivileged namespace, and containers can join regardless of privilege. Further, the privilege of the pause container doesn't affect its ability to reap (if the pod shares pid).

Unconditionally set the pause container's process label, so the rest of the (unprivileged) containers in the pod inherit it.


umohnani8:
oh ha, I thought this was master. didn't pay close attention to the branches.

PR #2932, opened 2019-10-31
Support manifest lists

nalind:
This updates us to containers image v5, buildah v1.11.4, and a new enough libpod, allowing CRI-O to pull and use images using tags and digests that resolve to manifest lists.


rhatdan:
LGTM
/test integration_rhel

Issue #2665, opened 2019-07-24
integration tests flakes


haircommander:
**Description**
When running the bats integration test suite, we often fail a test where `/tmp/tmp.*/crio/overlay` is busy when attempting to remove the temp directory. As a temporary fix, we can remove the directory best effort as in here: https://github.com/cri-o/cri-o/pull/2582, but that is not a permanent solution. 

**Steps to reproduce the issue:**
1. run `sudo bats test/*.bats`on master or release-1.15
2. note tests that fail with `rm: cannot remove '/tmp/tmp.*/crio/overlay': Device or resource busy`

saschagrunert:
Just a pointer, the current workaround to be fixed can be found here: https://github.com/cri-o/cri-o/pull/2667#discussion_r312976058

haircommander:
also a pointer: https://github.com/cri-o/cri-o/pull/2667#discussion_r313051037 because I don't want to forget

haircommander:
`# time="2019-08-13T14:30:22Z" level=fatal msg="run pod sandbox failed: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_podsandbox1_redhat.test.crio_redhat-test-crio_1(e9f14f3892bcc8ae06001759eb29b6d415b225d661d4838b08b64fb9d874d5b9): failed to set bridge addr: could not add IP address to \"cni0\": file exists"`
should be an easy one to fix and happens every so often

saschagrunert:
> `# time="2019-08-13T14:30:22Z" level=fatal msg="run pod sandbox failed: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_podsandbox1_redhat.test.crio_redhat-test-crio_1(e9f14f3892bcc8ae06001759eb29b6d415b225d661d4838b08b64fb9d874d5b9): failed to set bridge addr: could not add IP address to \"cni0\": file exists"`
> should be an easy one to fix and happens every so often

Do you see a way to reproduce this? The log message comes directly from the bridge plugin, so a locking might be needed more there instead of within CRI-O... :thinking:

haircommander:
maybe, but I was thinking the problem is there are multiple crio's running at once, using the same interface (because of parallel bats). If we used cni$RAND instead of every one using cni0 that should alleviate it.

haircommander:
PR to test here https://github.com/cri-o/cri-o/pull/2722

haircommander:
[...]

PR #2942, opened 2019-11-04
Add disk usage for ListContainerStats

QiWang19:
Show disk usage in the container stats

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/server/container_stats.go
+++ b/server/container_stats.go
@@ -1,12 +1,31 @@
 package server
 
 import (
+	"path/filepath"
+
 	"github.com/cri-o/cri-o/internal/oci"
+	crioStorage "github.com/cri-o/cri-o/utils"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
-func buildContainerStats(stats *oci.ContainerStats, container *oci.Container) *pb.ContainerStats {
+func (s *Server) buildContainerStats(stats *oci.ContainerStats, container *oci.Container) *pb.ContainerStats {
+	// TODO: Fix this for other storage drivers. This will only work with overlay.
+	var writableLayer *pb.FilesystemUsage
+	if s.ContainerServer.Config().RootConfig.Storage == "overlay" {
+		diffDir := filepath.Join(filepath.Dir(container.MountPoint()), "diff")
+		bytesUsed, inodeUsed, err := crioStorage.GetDiskUsageStats(diffDir)
+		if err != nil {
+			logrus.Warnf("unable to get disk usage for container %s， %s", container.ID(), err)
+		}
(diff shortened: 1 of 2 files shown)


QiWang19:
follow  PR https://github.com/cri-o/cri-o/pull/1869#issuecomment-548998063
 @drakedevel PTAL

drakedevel:
/test integration_rhel
It looks like the tests flaked as described in #2665

@QiWang19 this looks right to me (it appears equivalent to b986e6a8d2af34451363492479d2671a68fd20a3), although I'm not a collaborator on this repository so I can't give a proper lgtm. I'll try to give it a shot in my test cluster this evening.

drakedevel:
A simple cherry-pick onto a783f2374ebcdc4c487c7dd6ca4329b72f2749cf (1.16.0-rc2) works for me (`crictl stats` returns sane numbers); I don't actually have an environment to test on master at the moment.

server/container_stats.go
```diff
@@ -1,12 +1,22 @@
...
+
 	"github.com/cri-o/cri-o/internal/oci"
+	crioStorage "github.com/cri-o/cri-o/utils"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
 func buildContainerStats(stats *oci.ContainerStats, container *oci.Container) *pb.ContainerStats {
+	// TODO: Fix this for other storage drivers. This will only work with overlay.
```
giuseppe: does it need to check overlay is used here and in case it is not leave `WritableLayer = nil`?

QiWang19: @giuseppe do you know where I can check if the overlay is used?

saschagrunert: Yes, via the `server.Config().Storage` variable, the function just has to be bound to the `*Server` instance.

haircommander:
This seems like an API that should be exposed at the driver layer in c/storage. I am not really happy about the TODO, as well as all of this implementation specific code on the client side. WDYT @nalind @rhatdan @giuseppe

rhatdan:
@haircommander Yes I agree that this should be an interface in Container/Storage.  Do we need this now in CRI-O and then can work on it in C/S at a later time?

haircommander:
ah, so I see this is already in 1.13 https://github.com/cri-o/cri-o/commit/b986e6a8d2af34451363492479d2671a68fd20a3#diff-537a8ea573af14a02a208d521e5654aa
but never made it into master.
/retest

when tests go green, I'll merge, and we should backport. but we should also open an issue in c/storage to track.

haircommander:
@QiWang19 can you backport this to 1.1{4..6} if the patch isn't present?

QiWang19:
@haircommander ok, I will  backport it

PR #2944, opened 2019-11-05
don't set PodIPs on host network pods

danwinship:
If the runtime returns an IP for a HostNetwork pod, then kubelet will accept that IP as the pod's PodIP. But if it returns nothing, kubelet will fill in the PodIP with whatever it considers the correct node IP to be. In particular, on a dual-stack host, letting kubelet decide will ensure that hostNetwork pods end up with the node's "preferred" IP according to kubelet.

(This change means that the `host_ip` config variable and all of the code for detecting it if it's not set is now *mostly* useless, although it still gets used to figure out `bindAddress`... Probably that can be simplified though.)

/cc @mrunalp @haircommander

### DIFF
--- a/server/sandbox_network.go
+++ b/server/sandbox_network.go
@@ -17,7 +17,7 @@ import (
 // or an error
 func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs []string, result cnitypes.Result, err error) {
 	if sb.HostNetwork() {
-		return s.hostIPs, nil, nil
+		return nil, nil, nil
 	}
 
 	podNetwork, err := s.newPodNetwork(sb)
@@ -89,7 +89,7 @@ func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs 
 // getSandboxIP retrieves the IP address for the sandbox
 func (s *Server) getSandboxIPs(sb *sandbox.Sandbox) (podIPs []string, err error) {
 	if sb.HostNetwork() {
-		return s.hostIPs, nil
+		return nil, nil
 	}
 
 	podNetwork, err := s.newPodNetwork(sb)
--- a/server/server.go
+++ b/server/server.go
@@ -67,7 +67,6 @@ type Server struct {
 	hostportManager hostport.HostPortManager
 
 	appArmorProfile string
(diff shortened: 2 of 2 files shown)


haircommander:
huh, most notably this passes *e2e and unit-tests (the other failures are mostly cosmetic, and integration_rhel is a known flake).
@saschagrunert PTAL

haircommander:
interesting that `integration` fails with `not ok 103 Connect to pod hostport from the host` but `integration_fedora` passed...

haircommander:
If we go down this route (which I'd like to), we could merge host_port and stream_port and deprecate host_port. That can be handled in a separate PR though

saschagrunert:
Yeah this makes sense to me. Is the kubelet flag `--node-ip` the only way how to (hardly) set the IP of the node? I'm just wondering if there are intended use cases to still have CRI-O set the IP, but I cannot find one right now.

We should document that somewhere, right?

danwinship:
> interesting that `integration` fails with `not ok 103 Connect to pod hostport from the host` but `integration_fedora` passed...

I'm thinking that must be a flake, because it's not even using a HostNetwork pod

haircommander:
```
/bin/bash: sudo: command not found
Exited with code 127
```
did we change the circle images @saschagrunert ? This has appeared twice for the completions job

danwinship:
> Is the kubelet flag `--node-ip` the only way how to (hardly) set the IP of the node?

Yes, I think so. In theory you shouldn't even need that; it should pick the right IP by default. But for a long time there were kubelet bugs that made the automatic choice of node IP unreliable on hosts with multiple IPs.

saschagrunert:
> ```
> /bin/bash: sudo: command not found
> Exited with code 127
> ```
> 
> did we change the circle images @saschagrunert ? This has appeared twice for the completions job

This should be fixed with https://github.com/cri-o/cri-o/pull/2945

haircommander:
A rebase **should** fix completions, and ultimately the failure isn't the fault of this PR, so I think we're good to get this in
@mrunalp @umohnani8 PTAL

saschagrunert:
/retest

The completions validation is still not fixed yet (PR incoming, should be no blocker).

PR #2950, opened 2019-11-06
Fix fish shell completion for new default conmon system.slice

saschagrunert:
This seems to have been slipped through during all the merges. Fixed now :)


haircommander:
LGTM

PR #2947, opened 2019-11-06
Vendor in latest buildah and libpod

umohnani8:
Pull in changes to secrets package that allows users
to set FIPS_DISABLE label in pod to disable mounting
the fips secret.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

PR #2957, opened 2019-11-08
Bump go.mod to use go 1.13

saschagrunert:
To avoid inconsistent behavior between go versions we now use go 1.13 per
default.

### DIFF
--- a/contrib/test/integration/golang.yml
+++ b/contrib/test/integration/golang.yml
@@ -51,8 +51,6 @@
 - name: install Go tools and dependencies
   shell: /usr/bin/go get -u "github.com/{{ item }}"
   with_items:
-    - tools/godep
     - onsi/ginkgo/ginkgo
     - onsi/gomega
-    - cloudflare/cfssl/cmd/...
     - jteeuwen/go-bindata/go-bindata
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
Change LGTM, but the tests don't.

haircommander:
Failures are from an ansible bump causing the provisioning code to fail

saschagrunert:
> Failures are from an ansible bump causing the provisioning code to fail

Is there anything I can do to solve this issue? Where did the ansible update happen?

haircommander:
I'm not really sure, I think our best bet is to fix the ansible ATM

umohnani8:
Could this help fix the failures I am seeing in https://github.com/cri-o/cri-o/pull/2947?

haircommander:
woah, gnarly. Yeah, it may

saschagrunert:
@haircommander I think the focus for the `e2e_features_*` tests does not work any more. Are you already aware of this issue?

haircommander:
uhhh I don't know why that's happening. Through, e2e_features* don't test anything e2e* doesn't so it's not a huge loss. definitely something to look into though.

saschagrunert:
Test are green again, please take a look.

haircommander:
thanks prow

PR #2967, opened 2019-11-14
Add pinns utility for pinning namespaces

mrunalp:
@haircommander @umohnani8 ptal.

This is an initial just enough code to work version. I will clean it up.

### DIFF
--- a/pinns/pinns.c
+++ b/pinns/pinns.c
@@ -1,0 +1,124 @@
+#define _GNU_SOURCE
+#include <fcntl.h>
+#include <getopt.h>
+#include <linux/limits.h>
+#include <sched.h>
+#include <signal.h>
+#include <stdbool.h>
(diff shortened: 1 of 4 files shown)


pinns/pinns.c
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  if (bind_ipc) {
+    if (bind_ns(pin_path, "ipc") < 0) {
+      exit(1);
+    }
+  }
+
+  if (bind_net) {
+    if (bind_ns(pin_path, "net") < 0) {
```
haircommander: I find it a bit odd we exit 1 for failure but EXIT_SUCCESS upon succeeding

mrunalp: will be cleaned up :)

haircommander:
/retest
LGTM

PR #2975, opened 2019-11-16
Remove double log of PodSandboxStatusResponse

saschagrunert:
We already log every request and response via the interceptors,
so there is no need to log it twice.


TomSweeneyRedHat:
LGTM and happy tests.  Would like a head nod from @mrunalp or @haircommander before merge

PR #2980, opened 2019-11-18
Update conmon to v2.0.3

haircommander:
- Bump go module
- Bump test images
- Bump conmon for ansible based tests

Cherry-picked-by: Peter Hunt <pehunt@redhat.com>


haircommander:
circle builds will fail, I haven't pushed the images yet

.circleci/config.yml
```diff
@@ -4,8 +4,8 @@ version: 2.1
 stdenv: &stdenv
   environment:
     GOCACHE: &gocache /tmp/go-build
-    IMAGE: &image quay.io/crio/circle-1.12:1.16.0
-    IMAGELEGACY: &imagelegacy quay.io/crio/circle-1.10:1.16.0
+    IMAGE: &image quay.io/crio/circle-1.12:1.16.1
+    IMAGELEGACY: &imagelegacy quay.io/crio/circle-1.10:1.16.1
```
saschagrunert: Can we use the same test images like for the master branch?

haircommander: fixed

haircommander: though one day I think we should have the images follow the release version (like master is 1.17.x where x is the image version) so we know which crictl is inside and such

rhatdan:
LGTM
although tests not looking good.

umohnani8:
LGTM

Makefile
```diff
@@ -178,6 +183,13 @@ bin/crio.cross.%: git-vars .gopathok .explicit_phony
 crioimage: git-vars
 	$(CONTAINER_RUNTIME) build -t ${CRIO_IMAGE} .
 
+local-image:
+	$(TESTIMAGE_SCRIPT)
+
+test-images:
+	$(TESTIMAGE_SCRIPT) -g 1.13 -a amd64
+	$(TESTIMAGE_SCRIPT) -g 1.10 -a amd64
```
kolyshkin, line 191: Wonder why are we using golang 1.13 and 1.10 here?

Asking since I'm trying to test 1.16 branch with newer runc, and it chokes on golang 1.10.

kolyshkin: @saschagrunert @haircommander PTAL ^^^

haircommander: I feel like this was a version that was in debian or some slow moving distro, but I forget exactly.

saschagrunert: We had a backwards compatibility to go 1.10 on 1.16.x. Since it is now highly EOL we can consider dropping it. Unfortunately we would need kind of a policy how to handle golang versions on release branches, which would further increase the maintenance efforts on our side.

PR #2992, opened 2019-11-25
WIP: set "readonly_rootfs" and "privileged" in the right places in tests

nalind:
The `readonly_rootfs` and `privileged` options are part of Linux-specific security context information ([definition](https://github.com/cri-o/cri-o/blob/release-1.16/vendor/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.pb.go#L2431)).  Setting them anywhere else doesn't appear to have any effect.

### DIFF
--- a/test/testdata/container_config.json
+++ b/test/testdata/container_config.json
@@ -37,7 +37,6 @@
 		"owner": "dragon",
 		"daemon": "crio"
 	},
-	"privileged": true,
 	"log_path": "",
 	"stdin": false,
 	"stdin_once": false,
@@ -58,6 +57,7 @@
 				"pid": 1
 			},
 			"readonly_rootfs": false,
+			"privileged": true,
 			"selinux_options": {
 				"user": "system_u",
 				"role": "system_r",
--- a/test/testdata/container_config_logging.json
+++ b/test/testdata/container_config_logging.json
@@ -39,8 +39,6 @@
 		"owner": "dragon",
 		"daemon": "crio"
 	},
-	"privileged": true,
-	"readonly_rootfs": true,
 	"log_path": "",
 	"stdin": false,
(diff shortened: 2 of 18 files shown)


rhatdan:
@nalind @haircommander Is this still being worked on? Still something we want?

test/testdata/container_config.json
```diff
@@ -58,6 +57,7 @@
 				"pid": 1
 			},
 			"readonly_rootfs": false,
+			"privileged": true,
```
haircommander: I think we should drop this one, and see which containers fail from not having priv

test/testdata/container_config_logging.json
```diff
@@ -57,6 +55,8 @@
 			"namespace_options": {
 				"pid": 1
 			},
+			"readonly_rootfs": true,
+			"privileged": true,
```
haircommander: same here, no priv needed methinks

haircommander:
I started a review, but thinking about it, each of these should have no privilege, with the exception of container_config_privileged.json. All the readonly_rootfs should be moved to make sure we're testing that as expected.

Thanks for the patch @nalind this is very much needed.

haircommander:
(btw, the ci errors are:
`# time="2019-12-04T19:14:40Z" level=fatal msg="Creating container failed: rpc error: code = Unknown desc = no privileged container allowed in sandbox"`
which is from the container asking for priv when the sandbox wasn't specified to have it

haircommander:
I am taking this over [here](https://github.com/cri-o/cri-o/pull/2992)

rhatdan:
@haircommander You pointed at your self?

haircommander:
lol [here](https://github.com/cri-o/cri-o/pull/3381) **

Issue #2996, opened 2019-11-26
cgroups mounted differently for privileged containers


nalind:
It looks like we've changed how we're running privileged containers and pods.  In 1.14, runc would be given these in its configuration for a privileged container:
```
        {
            "destination": "/sys",
            "type": "sysfs",
            "source": "sysfs",
            "options": [
                "nosuid",
                "noexec",
                "nodev",
                "rw"
            ]
        },
	{
            "destination": "/sys/fs/cgroup",
            "type": "cgroup",
            "source": "cgroup",
            "options": [
                "nosuid",
                "noexec",
                "nodev",
                "relatime",
                "rw"
            ]
        },
```
An unprivileged container would get the same mounts, but marked read-only.

In 1.16, a privileged container instead gets
```
        {
            "destination": "/sys",
            "type": "bind",
            "source": "/sys",
            "options": [
                "nosuid",
                "noexec",
                "nodev",
                "rw",
                "rbind"
            ]
        }
```
This appears to have been changed as a bug fix in https://github.com/cri-o/cri-o/pull/2627, but it breaks applications which are accustomed to reading control group settings that affect them from the root of cgroup hierarchies (e.g. `/sys/fs/cgroup/memory/memory.max_usage_in_bytes`).

In OpenShift we currently run builds with privileges, so checks executed by RUN instructions are getting tripped up by this (test disabled in https://github.com/openshift/origin/pull/24001, investigation in https://github.com/openshift/origin/pull/24157 and https://github.com/openshift/origin/pull/24174).

haircommander:
this should be fixed with #2997. please reopen if you disagree

PR #2997, opened 2019-11-26
[1.16] add rw mount of cgroup

haircommander:

fixes: https://github.com/cri-o/cri-o/issues/2996

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -620,9 +620,9 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 			specgen.RemoveMount("/sys/fs/cgroup")
 			sysMnt := rspec.Mount{
 				Destination: "/sys",
-				Type:        "bind",
-				Source:      "/sys",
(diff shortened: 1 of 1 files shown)


haircommander:
once approved I'll forward port

haircommander:
@nalind PTAL

nalind:
Hmm, this doesn't change the behavior I'm seeing.  it looks like we're depending on the order: moving the call that adds `sysMnt` so that it happens before `cgroupMnt` produces the desired result.

nalind:
The current revision calls `specgen.AddMount(cgroupMnt)` before `specgen.AddMount(sysMnt)`.  If we switch the order, things look right in my test VM.

haircommander:
updated, though the failing tests don't seem to be related (and thus are going to be annoying to fix)

nalind:
LGTM, thanks!

giuseppe:
LGTM

/retest

haircommander:
alright, tests should be fixed, PTAL @nalind

PR #3002, opened 2019-11-27
Show previous image tag if no current tag is present

saschagrunert:
Containers storage does now track the image name history, which can be
utilized by CRI-O to display an image name even if the image got
untagged due to pulling a newer image from a remote resource.


saschagrunert:
@mrunal @haircommander I think we have to mention that in a dedicated way in the release notes. It affects scheduling (in a positive way) as well as the information which images are available on the node.

server/image_list.go
```diff
@@ -36,6 +36,8 @@ func ConvertImage(from *storage.ImageResult) *pb.Image {
 	repoTags := []string{"<none>:<none>"}
 	if len(from.RepoTags) > 0 {
 		repoTags = from.RepoTags
+	} else if from.PreviousName != "" {
+		repoTags = []string{from.PreviousName + ":<none>"}
```
vrothberg: I didn't test it but could this lead to situations where we return a wrong (outdated) repoTags? (I didn't look at the call sites)

saschagrunert: Hm, I don't think so because `<none>` will be always treated as "tag does not exist"

PR #3004, opened 2019-11-27
Deflake network integration tests

saschagrunert:
The ping pod tests removed all sandboxes after the test which caused a
race condition in removal of the pod sandbox. Isolating them and
removing only the sandboxes they created should fix that issue.


rhatdan:
/test integration_fedora
LGTM

PR #3007, opened 2019-11-27
Remove unused containerd release-tool

saschagrunert:
We're not using this tool anymore so we can remove it, right? :innocent:


haircommander:
I do use it to draft the releases, but we can just rely on it being out of tree

saschagrunert:
> I do use it to draft the releases, but we can just rely on it being out of tree

Hm, do you use the version in this repo? Is there another script invocation somewhere? :eyes:

haircommander:
I use the latest after at least https://github.com/containerd/release-tool/commit/234eed92eab8c824a2b83c823ac2eef23db44b0b so we can use go modules in 1.14+

saschagrunert:
Okay, so closing this now. :+1:

PR #3010, opened 2019-12-02
Move SystemContext from Server into Config

saschagrunert:
This allows to simplify the configuration reload strategy as well as
updating parts of the system context during live reload. There are now
fewer locations where the context has to be passed around.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -18,8 +18,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/containers/image/v5/pkg/sysregistriesv2"
-	"github.com/containers/image/v5/types"
 	"github.com/containers/libpod/pkg/apparmor"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/cri-o/cri-o/internal/lib"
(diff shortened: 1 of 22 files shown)


saschagrunert:
Rebased on top of the latest master to get the CI fixes in.

server/server.go
```diff
@@ -470,12 +455,6 @@ func New(
 		logrus.Infof("unable to start config watcher: %v", err)
 	}
 
-	// Start a configuration watcher for the registries of the SystemContext
-	registriesPath := sysregistriesv2.ConfigPath(s.systemContext)
-	if _, err := s.StartConfigWatcher(registriesPath, s.ReloadRegistries); err != nil {
-		logrus.Infof("unable to start config watcher: %v", err)
-	}
-
```
TomSweeneyRedHat: So this isn't needed any more?

TomSweeneyRedHat:
One question.  I love the cleanup aspect of this, very nice @saschagrunert !

server/server.go
saschagrunert: Yes, this is the main benefit of the change. Since the system context is now part of config we can reload the registries directly there. :) (see reload.go:35). The naming `ConfigWatcher` is a bit misleading because it indicates that we're watching a file for changes, which we don't. We just watch for SIGHUP.

haircommander:
LGTM @mrunalp @umohnani8 PTAL

PR #3011, opened 2019-12-02
[1.16] add default hooks dir to crio.conf

haircommander:
In removing the implicit hooks dir, we also don't have any hooks dirs we ship by default. Given we'd like some sane defaults, we should add hooks.DefaultDir in the default config.

### DIFF
--- a/completions/fish/crio.fish
+++ b/completions/fish/crio.fish
@@ -10,7 +10,7 @@ function __fish_crio_no_subcommand --description 'Test if there has been any sub
 end
 
 complete -c crio -n '__fish_crio_no_subcommand' -f -l additional-devices -r -d 'devices to add to the containers (default: [])'
-complete -c crio -n '__fish_crio_no_subcommand' -f -l apparmor-profile -r -d 'default apparmor profile name (default: "crio-default-1.16.0")'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l apparmor-profile -r -d 'default apparmor profile name (default: "crio-default-1.16.1")'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l bind-mount-prefix -r -d 'specify a prefix to prepend to the source of a bind mount (default: "")'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l cgroup-manager -r -d 'cgroup manager (cgroupfs or systemd) (default: "cgroupfs")'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l cni-config-dir -r -d 'CNI configuration files directory (default: "/etc/cni/net.d/")'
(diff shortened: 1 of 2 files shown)


haircommander:
begins to address: https://bugzilla.redhat.com/show_bug.cgi?id=1777838

completions/fish/crio.fish
```diff
@@ -34,7 +34,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l gid-mappings -r -d 'specif
 complete -c crio -n '__fish_crio_no_subcommand' -f -l global-auth-file -r -d 'path to a file like /var/lib/kubelet/config.json holding credentials necessary for pulling images from secure registries (default: "")'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l grpc-max-recv-msg-size -r -d 'maximum grpc receive message size in bytes (default: %!q(int=16777216))'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l grpc-max-send-msg-size -r -d 'maximum grpc receive message size (default: %!q(int=16777216))'
```
saschagrunert: TODO: `%!q(int=16777216)` looks like a bug, I'll follow up on this.

pkg/config/config.go
```diff
@@ -498,6 +499,7 @@ func DefaultConfig() (*Config, error) {
 			DefaultSysctls:           []string{},
 			DefaultUlimits:           []string{},
 			AdditionalDevices:        []string{},
+			HooksDir:                 []string{hooks.DefaultDir},
```
saschagrunert: Do we need to add documentation somewhere else like in the template.go?

haircommander: we shouldn't. the template.go gets the hooks configured and generates based on that.

zvonkok:
@haircommander `/etc/containers/oci/hooks.d` should be the default hook directory, This one is only writable on RHCOS. The `/usr/share/containers/oci/hooks.d` is read-only.  BTW the `oci-systemd-hook.json` is per default in `/usr/share/containers/oci/hooks.d`. Is this not needed to be run? Is this hook optional?

haircommander:
@zvonkok https://github.com/openshift/machine-config-operator/pull/1299 adds /etc/... to RHCOS. I think it makes sense we use the `DefaultDir` as a general default for upstream CRI-O, then update it in different platforms as needed

zvonkok:
@haircommander thanks, this makes sense!

carbonin:
I saw mention in here that `oci-systemd-hooks` is not needed anymore on cri-o nodes. What does that mean exactly? How does running a container with init work without `privileged`?

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1807245

PR #3012, opened 2019-12-03
Fix CLI double default prints

saschagrunert:
We do not have to print out the defaults twice if a value is already
provided by the CLI. This fixes a miss-format for the GRPC send and
receive message sizes as well.

Refers to: https://github.com/cri-o/cri-o/pull/3011#discussion_r352701606


saschagrunert:
It looks like that the issue is not solved with containers/storage v1.15.1. I think I have to dive deeper into that issue...

haircommander:
woah happy tests
/lgtm

PR #3018, opened 2019-12-04
[1.16] Deflake network integration tests

haircommander:
The ping pod tests removed all sandboxes after the test which caused a
race condition in removal of the pod sandbox. Isolating them and
removing only the sandboxes they created should fix that issue.

### DIFF
--- a/test/network.bats
+++ b/test/network.bats
@@ -85,48 +85,6 @@ function teardown() {
 
 }
 
-@test "Ping pod from the host" {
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM but it looks like you're still fighting with the tests.

saschagrunert:
I'm going to fix the unit tests in https://github.com/cri-o/cri-o/pull/3019

PR #3023, opened 2019-12-06
Update libcontainer stats to use runc event stats

saschagrunert:
We can remove the libcontainer factory dependency by relying only on the runc event stats.


internal/oci/oci_linux.go
```diff
@@ -83,45 +85,47 @@ func newPipe() (parent, child *os.File, err error) {
 	return os.NewFile(uintptr(fds[1]), "parent"), os.NewFile(uintptr(fds[0]), "child"), nil
 }
 
-func loadFactory(root string) (libcontainer.Factory, error) {
-	abs, err := filepath.Abs(root)
+// eventStats gets the stats for the container with the given id from an OCI runtime
```
haircommander: this is a nit and a style difference, but generally I prefer caller functions to be above callee functions, so it can be read from top down.

haircommander:
one nit, otherwise LGTM

internal/oci/oci_linux.go
saschagrunert: Sure, I moved it one layer down :+1:

TomSweeneyRedHat:
LGTM assuming happy tests.

mrunalp:
Can we measure the performance impact?

saschagrunert:
> Can we measure the performance impact?

Hm, I could do some manual tests but don't think that they are representative. Do you think we should wait with this change until we have performance tests up and running?

internal/oci/oci_linux.go
```diff
@@ -83,45 +85,47 @@ func newPipe() (parent, child *os.File, err error) {
...
+
+	return stats, nil
 }
 
-// libcontainerStats gets the stats for the container with the given id from runc/libcontainer
-func (r *runtimeOCI) libcontainerStats(ctr *Container) (*libcontainer.Stats, error) {
-	factory, err := loadFactory(r.root)
+// eventStats gets the stats for the container with the given id from an OCI runtime
+func (r *runtimeOCI) eventStats(ctr *Container) (*types.Stats, error) {
+	res, err := utils.ExecCmd(r.path, rootFlag, r.root, "events", "--stats", ctr.ID())
```
giuseppe: I don't think we can assume the OCI runtime has the `events` command.  It will work only with runc.

saschagrunert: Hm, does creating a libcontainer factory work with other runtimes, like crun?

giuseppe: I thought so :( but looking at the code, it seems it is already broken.

We should read this information directly from the cgroups (possibly using github.com/containers/libpod/pkg/cgroups) and not depend from a single OCI runtime.

saschagrunert: Yeah, that would be an option. If I change this now we would definitely touch it again when we migrate to libpod.

Please take a look at: https://github.com/cri-o/cri-o/pull/3030

saschagrunert:
Closing in favor of https://github.com/cri-o/cri-o/pull/3030

PR #3025, opened 2019-12-06
Add network setup metrics

### LINKED ISSUE #2906: Add metric for CNI network setup time
We want to be able to accurately measure and improve pod startup time. We should track the time it takes for CNI to set up the network namespace for a pod so we can improve the performance and help debug issues in running clusters.

cc: @smarterclayton

saschagrunert:
This adds two new metrics `network_setup_pod` and
`network_setup_overall` which provides a better way for network startup
time debugging.

Closes #2906

---

As a follow-up I suggest adding a documentation part about the available metrics and what they actually do.

### DIFF
--- a/server/sandbox_network.go
+++ b/server/sandbox_network.go
@@ -5,17 +5,21 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@haircommander @umohnani8 @mrunalp @giuseppe PTAL

haircommander:
LGTM

TomSweeneyRedHat:
LGTM integration-userns test isn't happy, might be a flake.

PR #2991, opened 2019-11-24
Default manage_network_ns_lifecycle to true

saschagrunert:
It is recommended to set the configuration option `manage_network_ns_lifecycle = true`, which we will now default from the beginning of releases newer than v1.17.x. All test scenarios have been adapted as well.

PR #2988, opened 2019-11-22
Update urfave/cli to v2.0.0

saschagrunert:
The v2 introduces changes which need code adaptions on the CRI-O side. This PR fixes them as well.

PR #3028, opened 2019-12-10
Vendor kubernetes v1.17.0

saschagrunert:
That's an easy one, because we have no further code changes between the latest RC vendored and the final tag.


rhatdan:
Does this mean we should cut a release?

saschagrunert:
> Does this mean we should cut a release?

I think we should definitely do before Christmas. From my point of view these four would be good to be part of the release: #3029, #2991, #2988, #2912

saschagrunert:
I label them for the release

PR #3030, opened 2019-12-10
Update libcontainer stats to use cgroups stats

saschagrunert:
We can remove the runc/libcontainer dependency by relying only on the
cgroup stats.

Initiated by: https://github.com/cri-o/cri-o/pull/3023#discussion_r355991841
Refers to: #3023


internal/oci/stats.go
```diff
@@ -24,24 +24,27 @@ type ContainerStats struct {
...
 // Returns the total number of bytes transmitted and received for the given container stats
-func getContainerNetIO(stats *libcontainer.Stats) (received, transmitted uint64) {
-	for _, iface := range stats.Interfaces {
-		received += iface.RxBytes
-		transmitted += iface.TxBytes
-	}
-	return
+func getContainerNetIO() (received, transmitted uint64) {
+	/*
+		TODO(saschagrunert): find a way to retrieve the data here
```
saschagrunert: This triggers more effort on our side and will change as well when migrating to libpod.

mrunalp:
:+1: to this approach if we can make it feature complete :)

saschagrunert:
>  to this approach if we can make it feature complete :)

Alright I made it complete, please take a look :)

server/container_stats_test.go
```diff
@@ -19,21 +19,6 @@ var _ = t.Describe("ContainerStats", func() {
 	AfterEach(afterEach)
 
 	t.Describe("ContainerStats", func() {
-		It("should fail when container stats failed", func() {
```
saschagrunert: This test does not fail any more on root (CI user).

saschagrunert:
It would be awesome to have this in 1.17 as well.

internal/oci/stats.go
```diff
@@ -24,24 +28,37 @@ type ContainerStats struct {
...
 // Returns the total number of bytes transmitted and received for the given container stats
-func getContainerNetIO(stats *libcontainer.Stats) (received, transmitted uint64) {
-	for _, iface := range stats.Interfaces {
-		received += iface.RxBytes
-		transmitted += iface.TxBytes
-	}
-	return
+func getContainerNetIO(netNsPath string) (received, transmitted uint64) {
+	ns.WithNetNSPath(netNsPath, func(_ ns.NetNS) error { // nolint: errcheck
+		link, err := netlink.LinkByName(ocicni.DefaultInterfaceName)
```
haircommander: are we sure we used ocicni.DefaultInterfaceName?

haircommander:
in general LGTM. Are we using cri stats in our tests?

internal/oci/stats.go
saschagrunert: I thought so. Do we have other Interfaces inside containers than eth0?

[I mean the logic is now a bit turned around since we have now the containers perspective.]

saschagrunert:
> in general LGTM. Are we using cri stats in our tests?

Not right now I think. :thinking:

saschagrunert:
Rebased to solve the integration flakyness

haircommander:
@giuseppe PTAL

giuseppe:
LGTM

internal/oci/oci_linux.go
```diff
@@ -83,45 +83,31 @@ func newPipe() (parent, child *os.File, err error) {
...
+	stats.MemUsage = cgroupStats.Memory.Usage.Usage
+	stats.MemLimit = getMemLimit(cgroupStats.Memory.Usage.Limit)
 	stats.MemPerc = float64(stats.MemUsage) / float64(stats.MemLimit)
-	stats.PIDs = cgroupStats.PidsStats.Current
-	stats.BlockInput, stats.BlockOutput = calculateBlockIO(libcontainerStats)
-	stats.NetInput, stats.NetOutput = getContainerNetIO(libcontainerStats)
+	stats.PIDs = cgroupStats.Pids.Current
+	stats.BlockInput, stats.BlockOutput = calculateBlockIO(cgroupStats)
+
+	if netNsPath, err := ctr.NetNsPath(); err == nil {
```
haircommander: Whoops, i took out container level NetNsPath because it wasn't being used anywhere

haircommander:
The failures are legit btw

internal/oci/oci_linux.go
saschagrunert: Fixed it by manually setting the path in that place.

saschagrunert:
> The failures are legit

internal/oci/oci_linux.go
```diff
@@ -83,45 +83,32 @@ func newPipe() (parent, child *os.File, err error) {
...
+	stats.MemLimit = getMemLimit(cgroupStats.Memory.Usage.Limit)
 	stats.MemPerc = float64(stats.MemUsage) / float64(stats.MemLimit)
-	stats.PIDs = cgroupStats.PidsStats.Current
-	stats.BlockInput, stats.BlockOutput = calculateBlockIO(libcontainerStats)
-	stats.NetInput, stats.NetOutput = getContainerNetIO(libcontainerStats)
+	stats.PIDs = cgroupStats.Pids.Current
+	stats.BlockInput, stats.BlockOutput = calculateBlockIO(cgroupStats)
+
+	if ctr.state != nil {
+		netNsPath := fmt.Sprintf("/proc/%d/ns/net", ctr.state.Pid)
```
saschagrunert: I don't think that we have to quirk around with `filepath.Join` here since It will work only on linux in any case.

haircommander: This is the way it's handled elsewhere 👍

PR #3041, opened 2019-12-11
[1.16] Persist exit status 

mrunalp:
conmon will save the exit file of the container under
/var/lib so it gets persisted and available even on reboot.

Fixes to better handle exit code:

1. Change ExitCode to be a pointer to distinguish from
 exit code 0.
2. On reboot if we can't get runc state, then attempt
to read exit code from the persisted exit file.
This handles cases where crio was unable to persist
the state for the container but conmon had written
the exit file for the container.
3. When restoring state during CRI-O startup,
write back the state to disk so that any missed
exit file updates are accounted for.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -197,7 +197,7 @@ func (c *ContainerServer) Update() error {
 			ctr := c.GetContainer(id)
 			if ctr != nil {
 				// if the container exists, update its state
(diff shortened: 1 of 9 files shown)


mrunalp:
@giuseppe @haircommander This is ready for review.

TomSweeneyRedHat:
LGTM assuming happy tests

haircommander:
LGTM

PR #3022, opened 2019-12-05
add conmonmon

haircommander:
One of the many shortcomings of kernel space OOM killing is there isn't very much a process can do to prevent itself from being OOM killed. Even adjusting its score doesn't prevent the kernel from killing it.

Unforuntately for CRI-O, conmon being OOM killed is very bad. A container can be orphaned and left running, with no way to report it state to CRI-O. At worst, it could keep PIDs in CRI-O that really should have gone away.

Introducing conmonmon, a conmon monitor. conmonmon loops in a routine and intermittantly calls `kill 0` on conmon, seeing if its still alive. If it has for some reason been killed, then conmonmon kills its associated container, and sets the container's state as though it OOM'd. This way, we can report to the kubelet that the container OOM'd, even though there is no one listening to it.

Signed-off-by: Peter Hunt <pehunt@redhat.com>

PR #3042, opened 2019-12-11
[1.16] add conmonmon

haircommander:
One of the many shortcomings of kernel space OOM killing is there isn't very much a process can do to prevent itself from being OOM killed. Even adjusting itsscore doesn't prevent the kernel from killing it.

Unforuntately for CRI-O, conmon being OOM killed is very bad. A container can be orphaned and left running, with no way to report it state to CRI-O. At worse, it could keep PIDs in CRI-O that really should have gone away.

Introducing conmonmon, a conmon monitor. conmonmon loops in a routine and intermittantly calls `kill 0` on conmon, seeing if its still alive. If it has for some reason been killed, then conmonmon kills its associated container, and sets the container's state as though it OOM'd. This way, we can report to the kubelet that the container OOM'd, even though there is no one listening to it.

cherry-pick of:  https://github.com/cri-o/cri-o/pull/3022


haircommander:
let's see how far we can carry this
/cherry-pick release-1.15

PR #3044, opened 2019-12-12
[1.15] Add conmonmon

haircommander:
One of the many shortcomings of kernel space OOM killing is there isn't very much a process can do to prevent itself from being OOM killed. Even adjusting its score doesn't prevent the kernel from killing it.

Unforuntately for CRI-O, conmon being OOM killed is very bad. A container can be orphaned and left running, with no way to report it state to CRI-O. At worst, it could keep PIDs in CRI-O that really should have gone away.

Introducing conmonmon, a conmon monitor. conmonmon loops in a routine and intermittantly calls kill 0 on conmon, seeing if its still alive. If it has for some reason been killed, then conmonmon kills its associated container, and sets the container's state as though it OOM'd. This way, we can report to the kubelet that the container OOM'd, even though there is no one listening to


lib/conmonmon.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/containers/psgo"
+	"github.com/cri-o/cri-o/oci"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// the time between checking the registered conmons
+var sleepTime = 5 * time.Minute
```
TomSweeneyRedHat: Is this something that should be configurable via the conf file?

haircommander: that's one of the things I'd wire up if there's the ask

haircommander:
PTAL @umohnani8 @mrunalp @saschagrunert @rhatdan

rhatdan:
LGTM
/approve

haircommander:
```
13:13:43 + yq --trim=false w -i versions.yaml externals.critools.version 1.15.0
13:13:43 Error: unknown flag: --trim
```
weird, PTAL @chavafg

chavafg:
@haircommander fixed. we changed the version of `yq` we use for reading and writing yaml files, and this affected the kata job. I removed the `--trim` flag which is no longer supported. 

Can you please restart? 

Thanks  and sorry for the inconvenience. 
BTW, there might be other PRs affected.

haircommander:
no problem, thanks for the quick fix!

PR #3053, opened 2019-12-17
Add stats integration test

saschagrunert:
We also modify the test images to contain `jq` for testing purposes as
well as updating cri-tools.


haircommander:
I would rather not drop the `master-` prefix. I'd like the branch we're using in the circle images to be clear. If you'd rather, `1.17` works for me too.

haircommander:
`# /tmp/bats.31786.src: line 27: jq: command not found`
we may need an AMI update to install jq for the jenkins tests

saschagrunert:
> I would rather not drop the `master-` prefix. I'd like the branch we're using in the circle images to be clear. If you'd rather, `1.17` works for me too.

Alright, I re-push the images with the prefix.

> `# /tmp/bats.31786.src: line 27: jq: command not found`
> we may need an AMI update to install jq for the jenkins tests

Yep I see, I added an install `jq.yml`, let's see if that works.

saschagrunert:
Yeah it looks like that it’s working now.

/test integration_rhel

haircommander:
/test integration_rhel
/lgtm
(though one day I think we should move jq installation to ami build time)

saschagrunert:
@haircommander may I ask for a manual merge, I think we could skip the AMI tests in that case, WDYT?

PR #3054, opened 2019-12-17
test: use cri stats instead of cadvisor

haircommander:
by removing a hardcoded line to tell the kubelet to not use cri stats if it's using crio

### DIFF
--- a/contrib/test/integration/build/kubernetes.yml
+++ b/contrib/test/integration/build/kubernetes.yml
@@ -8,6 +8,15 @@
     version: "{{ k8s_git_version }}"
     force: "{{ force_clone | default(False) | bool}}"
 
+# replace hardcoded line not to use legacy stats provider
+- name: use CRI stats
+  lineinfile:
+    path: "{{ ansible_env.GOPATH }}/src/k8s.io/kubernetes/pkg/kubelet/cadvisor/util.go"
+    regexp: '^(\s*)runtimeEndpoint == CrioSocket \|\| runtimeEndpoint == \"unix\:\/\/\"\+CrioSocket$'
+    # keep the whitespace to ensure the file still looks pretty :)
+    line: '\1false'
+    backrefs: yes
+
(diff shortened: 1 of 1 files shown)


contrib/test/integration/build/kubernetes.yml
```diff
@@ -8,6 +8,15 @@
     version: "{{ k8s_git_version }}"
     force: "{{ force_clone | default(False) | bool}}"
 
+# replace hardcoded line to use legacy stats provider
```
haircommander: nit: `not use legacy stats provider`

haircommander:
I am not super familiar with the history, but did a bit of digging. 
The change originally came in https://github.com/kubernetes/kubernetes/pull/55642 which was supposed to be a temporary solution for: https://github.com/kubernetes/kubernetes/issues/51798 . the topic was largely dropped for a bit, and was briefly touch on here https://github.com/kubernetes/kubernetes/pull/76291#discussion_r273621481 .

I think it since fell off peoples radar. removing this line (saying "no, we should not use legacy cadvisor stats") and running it through tests should be a way of showing that we do satisfactorily implement the stats api (though I'm unsure if the test suite uses/tests CRI stats)

contrib/test/integration/build/kubernetes.yml
```diff
@@ -8,6 +8,15 @@
     version: "{{ k8s_git_version }}"
     force: "{{ force_clone | default(False) | bool}}"
 
+# replace hardcoded line not to use legacy stats provider
+- name: use CRI stats
+  lineinfile:
+    path: "{{ ansible_env.GOPATH }}/src/k8s.io/kubernetes/pkg/kubelet/cadvisor/util.go"
```
giuseppe: is there a PR to change it in Kubernetes as well?

haircommander: No I haven't opened one yet

PR #3058, opened 2019-12-21
managed namespaces: batch infra.State() calls

haircommander:
to save on holding the sandbox lock for each namespace, introduce NamespacePaths() to get all namespace paths the sandbox or infra container manage
also share a bit more code between infra and container namespace generation
As well as add unit tests and update some old ones

### DIFF
--- a/internal/lib/sandbox/namespaces.go
+++ b/internal/lib/sandbox/namespaces.go
@@ -45,21 +45,44 @@ type NamespaceIface interface {
 	Type() string
 }
 
-func (s *Sandbox) CreateManagedNamespaces(managedNamespaces []string, pinnsPath string) (map[string]string, error) {
+// ManagedNamespace is a structure that holds all the necessary information a caller would
+// need for a sandbox managed namespace
+// Where NamespaceIface does hold similar information, ManagedNamespace exists to allow this library
+// to not return data not necessarily in a NamespaceIface (for instance, when a namespace is not managed
+// by CRI-O, but instead is based off of the infra pid)
+type ManagedNamespace struct {
+	nsPath string
+	nsType string
+}
+
+// Type returns the namespace type
+func (m *ManagedNamespace) Type() string {
+	return m.nsType
+}
+
+// Type returns the namespace path
+func (m *ManagedNamespace) Path() string {
+	return m.nsPath
+}
+
+// CreateManagedNamespaces calls pinnsPath on all the managed namespaces for the sandbox.
+// It returns a slice of ManagedNamespaces it created.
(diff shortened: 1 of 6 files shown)


haircommander:
PTAL @rhatdan @saschagrunert @giuseppe

internal/lib/sandbox/namespaces.go
```diff
@@ -173,6 +197,17 @@ func (s *Sandbox) UtsNsJoin(nspath string) (err error) {
 	return err
 }
 
+// UserNs specific functions
+
+// UserNsPath returns the path to the user namespace of the sandbox.
+// If the sandbox uses the host namespace, nil is returned
```
saschagrunert: It’s not `nil` but an empty `string` :)

server/container_create_linux.go
```diff
@@ -577,25 +577,8 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
...
-	}
-
-	utsNsPath := sb.UtsNsPath()
-	if utsNsPath == "" {
-		return nil, errors.New("sandbox UTS namespace path returned empty")
-	}
-	if err := specgen.AddOrReplaceLinuxNamespace(string(rspec.UTSNamespace), utsNsPath); err != nil {
-		return nil, err
+	if err := configureGeneratorGivenNamespacePaths(sb.NamespacePaths(), specgen); err != nil {
+		return nil, errors.Wrapf(err, "failed to configure namespaces in container create")
```
saschagrunert: Just `Wrap()` should be enough here.

internal/lib/sandbox/namespaces.go
```diff
@@ -90,6 +91,29 @@ func (s *Sandbox) CreateNamespacesWithFunc(managedNamespaces []string, pinnsPath
 	return nsTypeToPath, nil
 }
 
+// NamespacePaths returns all the paths of the
+// namespaces of the sandbox. If a namespace is not
+// managed by the sandbox, the namespace of the infra
+// container will be returned.
+// It returns a map of nsType -> path, allowing for
+// callers to branch on the namespace type
+func (s *Sandbox) NamespacePaths() map[string]string {
```
saschagrunert: I think returning a structure for the namespace tuple would make it more error prone compared to a map, WDYT?

haircommander: hm yeah I agree. fixed

haircommander:
It also needs approval 🙂

haircommander:
> /approve
> 
> Merry Xmas

To you as well! 🎅🎄

PR #3060, opened 2019-12-29
Support pulling image specified by tag and digest.

### LINKED ISSUE #2351: Support pulling an image specified by both tag and digest
**Description**

Hi all! Currently trying to get skaffold to work. Skaffold works by pulling an image specified by both tag and digest. I wanted to know if you could support this in crio?

https://github.com/GoogleContainerTools/skaffold/issues/2089#issuecomment-491000366

Example: 
```
root@a2:~# crictl pull gcr.io/allthingssfx/hook-to-k8s:v0.28.0-116-ga1172365-dirty@sha256:9154d56ca7364f5f3a005c469ab0e53ee005641beb70fcbbf6d014af1f19858b
FATA[0000] pulling image failed: rpc error: code = Unknown desc = Invalid image name "gcr.io/allthingssfx/hook-to-k8s:v0.28.0-116-ga1172365-dirty@sha256:9154d56ca7364f5f3a005c469ab0e53ee005641beb70fcbbf6d014af1f19858b", unknown transport "gcr.io/allthingssfx/hook-to-k8s"
root@a2:~# crictl pull gcr.io/allthingssfx/hook-to-k8s:v0.28.0-116-ga1172365-dirty
Image is up to date for gcr.io/allthingssfx/hook-to-k8s@sha256:9154d56ca7364f5f3a005c469ab0e53ee005641beb70fcbbf6d014af1f19858b
```

Currently using crio version 1.13 built from source ~ 2 months ago.A
```
root@a1:~# crio --version
crio version 1.13.1-dev
commit: ""

```

Physical env

mgoltzsche:
Favour the digest over the tag if both specified.
This is required for docker compatibility.
fixes #2351.

_Docker supports canonical image references (`repo/image:tag@sha`) that contain both a tag and an image ID. The latter is actually used for resolution. Despite its ambiguity this feature is used by tools like skaffold to ensure that the correct image version is used within k8s manifests and enforce rollouts even when the tag was overwritten (e.g. `skaffold dev`) while having the tag there as a hint/documentation.
A downside of this feature - as for image ID based resolution in general - is that, when the image ID is deleted from the registry somehow, an image rebuild will not help as long as you don't regenerate the manifests and do a rollout as well since the new image will usually have a different ID/digest due to timestamps._

**- What I did**
When a remote image reference is specified ambiguously with both tag and digest its tag is removed to resolve based on the image ID only.

**- How I did it**
By stripping tags of ambiguous image names in `imageService.ResolveNames()`.

**- How to verify it**
I've added two unit test cases and improved the existing ones by initializing the testee (`imageService`) with a second system repository.

Also I ensured that this really fixes #2351 and verified that the CRI-O-internal image caching still works as well as the overall integration by running a corresponding [branch](https://github.com/mgoltzsche/kubernetes-setup/tree/crio-imageref-fix-verification) of my containerized kubernetes using this PR (`./make.sh clean && ./make.sh`).
After it started I've set up a docker registry and ran `skaffold dev` in an example project and verified that it was deployed successfully using the digest. Undeploying and redeploying using the same image also didn't cause further image lookups within the image registry so that apparently it also works fine with the internal image store and caching.

Support pulling image specified by tag and digest.

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -477,6 +477,30 @@ func splitDockerDomain(name string) (domain, remainder string) {
 	return
 }
 
+// imageNamesWithDigestOrTag strips the tag from ambiguous image references that have a digest as well (e.g. `image:tag@sha256:123...`).
+// Such image references are supported by docker but, due to their ambiguity,
+// explicitly not by containers/image.
+func imageNamesWithDigestOrTag(images []string) ([]string, error) {
+	normalized := make([]string, len(images))
+	for i, imageName := range images {
+		ref, err := reference.ParseNormalizedNamed(imageName)
+		if err != nil {
+			return nil, err
+		}
+		_, isTagged := ref.(reference.NamedTagged)
+		canonical, isDigested := ref.(reference.Canonical)
+		if isTagged && isDigested {
+			canonical, err = reference.WithDigest(reference.TrimNamed(ref), canonical.Digest())
+			if err != nil {
+				return nil, err
+			}
+			imageName = canonical.String()
+		}
+		normalized[i] = imageName
+	}
+	return normalized, nil
+}
+
 // ResolveNames resolves an image name into a storage image ID or a fully-qualified image name (domain/repo/image:tag).
 // Will only return an empty slice if err != nil.
 func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageName string) ([]string, error) {
@@ -507,7 +531,7 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 		if registry != nil && registry.Blocked {
 			return nil, fmt.Errorf("cannot use %q because it's blocked", imageName)
 		}
-		return []string{imageName}, nil
+		return imageNamesWithDigestOrTag([]string{imageName})
 	}
 	unqualifiedSearchRegistries, err := sysregistriesv2.UnqualifiedSearchRegistries(systemContext)
 	if err != nil {
@@ -540,7 +564,7 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 	if len(images) == 0 {
 		return nil, fmt.Errorf("all search registries for %q are blocked", remainder)
 	}
-	return images, nil
+	return imageNamesWithDigestOrTag(images)
 }
 
 // GetImageService returns an ImageServer that uses the passed-in store, and
--- a/internal/storage/image_test.go
+++ b/internal/storage/image_test.go
@@ -21,10 +21,13 @@ import (
 var _ = t.Describe("Image", func() {
 	// Test constants
 	const (
-		testRegistry            = "docker.io"
-		testImageName           = "image"
-		testNormalizedImageName = "docker.io/library/image:latest" // Keep in sync with testImageName!
-		testSHA256              = "2a03a6059f21e150ae84b0973863609494aad70f0a80eaeb64bddd8d92465812"
+		testDockerRegistry                  = "docker.io"
+		testExampleRegistry                 = "registry.example.org"
+		testImageName                       = "image"
+		testNormalizedImageName             = "docker.io/library/image:latest" // Keep in sync with testImageName!
+		testSHA256                          = "2a03a6059f21e150ae84b0973863609494aad70f0a80eaeb64bddd8d92465812"
+		testImageWithTagAndDigest           = "image:latest@sha256:" + testSHA256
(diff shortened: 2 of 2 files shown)


rhatdan:
/ok-to-test
@mtrmac @vrothberg PTAL

rhatdan:
@mgoltzsche Should we have a similar fix in Podman?

internal/pkg/storage/image.go
```diff
@@ -489,14 +489,21 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
...
 		return nil, err
 	}
 
+	// Use digest instead of tag when both provided (for docker compatibility)
+	_, isTagged := ref.(reference.NamedTagged)
+	canonical, isDigested := ref.(reference.Canonical)
+	if isTagged && isDigested {
+		imageName = canonical.Name() + "@" + canonical.Digest().String()
+	}
+
```
saschagrunert: I have the feeling that this logic should somehow go into containers/image, also considering that we want to move to the libpod image backend in the near future. (not blocking)

mtrmac: See e.g. the discussion in https://github.com/cri-o/cri-o/issues/2351 (and notably https://github.com/cri-o/cri-o/issues/2351#issuecomment-492726855 , which does suggest this could end up in Libpod).

One part of this is the ambiguity of intent/guarantees, another is that there are, sadly, quite a few manually-written incorrect parsers (“Is there a `@`? If so, the digest is everything after the first `:`”) that keep being added throughout the ecosystem. Rejecting :tag@digest in c/image somewhat limits the damage of such incorrect parsers. (OTOH, arguably, allowing :tag@digest could motivate maintainers of the code that use such parsers to fix the code.)

internal/pkg/storage/image.go
```diff
@@ -489,14 +489,21 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
...
 			return nil, ErrCannotParseImageID
 		}
 		return nil, err
 	}
 
+	// Use digest instead of tag when both provided (for docker compatibility)
+	_, isTagged := ref.(reference.NamedTagged)
+	canonical, isDigested := ref.(reference.Canonical)
+	if isTagged && isDigested {
+		imageName = canonical.Name() + "@" + canonical.Digest().String()
```
mtrmac: Something like
```go
imageName = reference.WithDigest(reference.TrimNamed(ref), canonical.Digest()).String()
```
please, to use the Go API instead of hard-coding string syntax (compare the other comment).

mgoltzsche: That's better indeed, done.

internal/pkg/storage/image.go
mgoltzsche: @saschagrunert previously I tried to solve it in [containers/image#784](https://github.com/containers/image/pull/784). However I realized what @mtrmac pointed out and that it would make the library unnecessarily more complicated or inconsistent while actually being a non-solution just for docker-compatibility.

mgoltzsche:
@rhatdan yes, consequently it would need to be added to podman as well for docker compatibility.

internal/pkg/storage/image.go
mtrmac: On second thought, oops… neither version is going to work as expected, because both (implied in `.Name()`/`.String()`) convert unqualified repo names (`busybox`) to the fully-qualified form (`docker.io/library/busybox`), breaking the `splitDockerDomain` check immediately below.

Sorry about not noticing that before.

So, it’s either doing this transformation on the _results_ of the search path expansion (a bit more work), or doing it on strings only without using the `reference` subpackage (eww).

(And having non-trivial tests for the unqualified name expansion would certainly have helped.)

mgoltzsche: Oh, right. I improved the tests correspondingly and fixed it.

rhatdan:
@mgoltzsche needs a rebase
@saschagrunert @mrunalp @umohnani8 @haircommander @kolyshkin PTAL

haircommander:
this LGTM (After rebase). Would like final :+1: from @mtrmac as well

haircommander:
@mgoltzsche  will merge upon rebase, thanks for the PR :tada:

mgoltzsche:
rebased

haircommander:
Reminder to self: some wacky errors going on here https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3060/test_pull_request_crio_integration_rhel/10077/artifacts/testout_userns.txt

/retest

srueg:
Thanks for the implementation!
Any plans on back porting this to let's say 1.16 (OpenShift 4.3)?

PR #3065, opened 2020-01-03
[1.16] oci: Handle timeouts correctly for probes

mrunalp:
We shouldn't return a custom error when an exec sync command
times out. Instead, we should return a non-zero exit code.

When the prober code in kubernetes sees a custom error
it goes into Unknown status and doesn't restart containers
on timed out probes as expected.

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"time"
 
+	conmonconfig "github.com/containers/conmon/runner/config"
 	"github.com/cri-o/cri-o/internal/pkg/findprocess"
 	"github.com/cri-o/cri-o/pkg/config"
 	"github.com/cri-o/cri-o/utils"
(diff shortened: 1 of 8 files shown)


mrunalp:
@haircommander @saschagrunert ptal

internal/oci/runtime_oci.go
```diff
@@ -454,6 +454,21 @@ func (r *runtimeOCI) ExecSyncContainer(c *Container, command []string, timeout i
 
 	logrus.Debugf("Received container exit code: %v, message: %s", ec.ExitCode, ec.Message)
 
+	// This has to be in sync with the message returned by conmon
+	timedOutMessage := "command timed out"
```
haircommander: This should be defined in `https://github.com/containers/conmon/blob/master/runner/config/config.go` and then we can import it

mrunalp: :+1: Opened https://github.com/containers/conmon/pull/110

mrunalp:
PR updated with conmon bump.

haircommander:
LGTM :+1:

PR #3066, opened 2020-01-03
oci: Handle timeouts correctly for probes

mrunalp:
We shouldn't return a custom error when an exec sync command
times out. Instead, we should return a non-zero exit code.

When the prober code in kubernetes sees a custom error
it goes into Unknown status and doesn't restart containers
on timed out probes as expected.

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -452,6 +452,21 @@ func (r *runtimeOCI) ExecSyncContainer(c *Container, command []string, timeout i
 
 	logrus.Debugf("Received container exit code: %v, message: %s", ec.ExitCode, ec.Message)
 
+	// This has to be in sync with the message returned by conmon
+	timedOutMessage := "command timed out"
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_oci.go
```diff
@@ -452,6 +452,21 @@ func (r *runtimeOCI) ExecSyncContainer(c *Container, command []string, timeout i
 
 	logrus.Debugf("Received container exit code: %v, message: %s", ec.ExitCode, ec.Message)
 
+	// This has to be in sync with the message returned by conmon
+	timedOutMessage := "command timed out"
```
rhatdan: Can we get this message from a vendored conmon?

mrunalp:
Yeah, I am going to open a PR to update conmon for master. We already did that for 1.16.

haircommander:
taking this over [here](https://github.com/cri-o/cri-o/pull/3231)

PR #3068, opened 2020-01-05
[1.16] Use golang 1.13 

mrunalp:
kubernetes requires go 1.13


mrunalp:
Merging since kata tests shouldn't be affected by this change.

PR #3073, opened 2020-01-07
Cleanup sandbox shared memory before removing it

saschagrunert:
If conmon gets killed either with running or without running CRI-O, then
we have no way to cleanup the shared memory for the pod. This will cause
an unlinkat failure, which is now fixed by removing the SHM before
actually touching the sandbox.

Reproduce via:

```bash
> sudo crictl runp test/testdata/sandbox_config.json
> sudo pkill conmon
> sudo crictl rmp -fa
7e4148565ae9430830515329a0cc3578c55fb041232bda4d77a9b3e11a0580d6
ERRO[0000] removing the pod sandbox "7e4148565ae9430830515329a0cc3578c55fb041232bda4d77a9b3e11a0580d6" failed: \
  rpc error: code = Unknown desc = failed to remove pod sandbox 7e4148565ae9430830515329a0cc3578c55fb041232bda4d77a9b3e11a0580d6: \
  unlinkat /var/run/containers/storage/overlay-containers/7e4148565ae9430830515329a0cc3578c55fb041232bda4d77a9b3e11a0580d6/userdata/shm: device or resource busy
FATA[0000] unable to remove sandbox(es)
> sudo crictl pods
POD ID              CREATED             STATE               NAME                NAMESPACE           ATTEMPT
7e4148565ae94       5 seconds ago       NotReady            podsandbox1         redhat.test.crio    1
```

Now:
```
> sudo crictl runp test/testdata/sandbox_config.json && sudo pkill conmon && sudo crictl rmp -fa
746c3b35b02e7c4eb80a4fd9440e7d718217ee61bb039acb638127176721d0c7
Removed sandbox 746c3b35b02e7c4eb80a4fd9440e7d718217ee61bb039acb638127176721d0c7
```

### DIFF
--- a/internal/lib/sandbox/sandbox.go
+++ b/internal/lib/sandbox/sandbox.go
@@ -7,10 +7,16 @@ import (
 	"time"
 
 	"github.com/cri-o/cri-o/internal/oci"
+	"github.com/docker/docker/pkg/mount"
+	"github.com/docker/docker/pkg/symlink"
+	"golang.org/x/sys/unix"
 	"k8s.io/apimachinery/pkg/fields"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 	"k8s.io/kubernetes/pkg/kubelet/dockershim/network/hostport"
 )
+
+// DevShmPath is the default system wide shared memory path
+const DevShmPath = "/dev/shm"
 
 // Sandbox contains data surrounding kubernetes sandboxes on the server
 type Sandbox struct {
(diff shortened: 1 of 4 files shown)


haircommander:
:open_mouth: good find, LGTM

internal/lib/sandbox/sandbox.go
```diff
@@ -315,3 +321,27 @@ func (s *Sandbox) Ready(takeLock bool) bool {
...
+	// /var/run/containers/storage/overlay-containers/CID/userdata/shm
+	// but /var/run on most system is symlinked to /run so we first resolve
+	// the symlink and then try and see if it's mounted
+	fp, err := symlink.FollowSymlinkInScope(s.ShmPath(), "/")
+	if err != nil {
+		return err
+	}
+	if mounted, err := mount.Mounted(fp); err == nil && mounted {
+		if err := unix.Unmount(fp, unix.MNT_DETACH); err != nil {
+			return err
```
rhatdan: Should we ignore EINVAL?
```
man umount
...
       EINVAL target is not a mount point.
```
This would happen if /dev/shm was already unmounted.

haircommander: we check if it's mounted before attempting to Unmount. Should protect against that case

PR #3076, opened 2020-01-07
Sandbox: Don't use an infra container in some cases

haircommander:
some changes this includes:
    allow for the infra container to not be created by the runtime, if sb.NeedsInfra() is false
    add concept of a 'spoofed' container. This allows us to make calls to ID(), Name(), and Description() without having the runtime create the container
    using spoof, change checks of podInfraContainer == nil to Spoofed(), as well as re-add Description() calls
    reorder some infra container creation code


haircommander:
Happy tests, PTAL @mrunalp @umohnani8 @saschagrunert @rhatdan @giuseppe

test/helpers.bash
```diff
@@ -36,6 +36,7 @@ RUNTIME_NAME=${RUNTIME_NAME:-runc}
 CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-runc}
 CONTAINER_UID_MAPPINGS=${CONTAINER_UID_MAPPINGS:-}
 CONTAINER_GID_MAPPINGS=${CONTAINER_GID_MAPPINGS:-}
+CONTAINER_MANAGE_NS_LIFECYCLE=${CONTAINER_MANAGE_NS_LIFECYCLE:-true}
```
saschagrunert: So I assume that the new code is now integration tested as well and we dont have to add more, right?

giuseppe:
LGTM

haircommander:
PTAL @mrunalp @umohnani8 @rhatdan

test/helpers.bash
haircommander: yeah it's being tested both in integration and e2e. I intend on adding more tests in the future, but figure we can get this in for now

mrunalp:
This doesn't work without the following patch. Without the patch, we always default to zero value for namespace mode which is pod and the infra container is always created for the pod.

```
diff --git a/server/sandbox_run_linux.go b/server/sandbox_run_linux.go
index a17ad221d..d9f989778 100644
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -514,6 +514,9 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 
        var container *oci.Container
        // Only create the container in the runtime if we need it
+
+       sb.SetNamespaceOptions(securityContext.GetNamespaceOptions())
+
        if sb.NeedsInfra(s.config.ManageNSLifecycle) {
                container, err = oci.NewContainer(id, containerName, podContainer.RunDir, logPath, labels, g.Config.Annotations, kubeAnnotations, "", "", "", nil, id, false, false, false, sb.Privileged(), sb.RuntimeHandler(), podContainer.Dir, created, podContainer.Config.Config.StopSignal)
                if err != nil {
@@ -554,7 +557,6 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
                g.AddAnnotation(fmt.Sprintf("%s.%d", annotations.IP, idx), ip)
        }
        sb.AddIPs(ips)
-       sb.SetNamespaceOptions(securityContext.GetNamespaceOptions())
 
        spp := securityContext.GetSeccompProfilePath()
        g.AddAnnotation(annotations.SeccompProfilePath, spp)

```

haircommander:
> This doesn't work without the following patch. Without the patch, we always default to zero value for namespace mode which is pod and the infra container is always created for the pod. (I just tested this quickly, so we may want to move that even earlier in the pod creation process).
> 
> ```
> diff --git a/server/sandbox_run_linux.go b/server/sandbox_run_linux.go
> index a17ad221d..d9f989778 100644
> --- a/server/sandbox_run_linux.go
> +++ b/server/sandbox_run_linux.go
> @@ -514,6 +514,9 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
>  
>         var container *oci.Container
>         // Only create the container in the runtime if we need it
> +
> +       sb.SetNamespaceOptions(securityContext.GetNamespaceOptions())
> +
>         if sb.NeedsInfra(s.config.ManageNSLifecycle) {
>                 container, err = oci.NewContainer(id, containerName, podContainer.RunDir, logPath, labels, g.Config.Annotations, kubeAnnotations, "", "", "", nil, id, false, false, false, sb.Privileged(), sb.RuntimeHandler(), podContainer.Dir, created, podContainer.Config.Config.StopSignal)
>                 if err != nil {
> @@ -554,7 +557,6 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
>                 g.AddAnnotation(fmt.Sprintf("%s.%d", annotations.IP, idx), ip)
>         }
>         sb.AddIPs(ips)
> -       sb.SetNamespaceOptions(securityContext.GetNamespaceOptions())
>  
>         spp := securityContext.GetSeccompProfilePath()
>         g.AddAnnotation(annotations.SeccompProfilePath, spp)
> ```

oof, good catch. Writing unit tests now to actually make sure things work as expected :sweat_smile:

haircommander:
PR updated with unit tests and patch

haircommander:
failures seem real

haircommander:
trying something out, it's not finished

/hold

haircommander:
this PR needs much more work before it is ready. Moving target to 1.18

rhatdan:
@haircommander Needs a rebase, and are you still working on it?

haircommander:
closing in favor of https://github.com/cri-o/cri-o/pull/3474

PR #3082, opened 2020-01-08
[1.15] Set process label to privileged when host network is true

haircommander:

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -363,11 +363,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 	}
 	hostIPC := containerConfig.GetLinux().GetSecurityContext().GetNamespaceOptions().GetIpc() == pb.NamespaceMode_NODE
 	hostPID := containerConfig.GetLinux().GetSecurityContext().GetNamespaceOptions().GetPid() == pb.NamespaceMode_NODE
+	hostNet := containerConfig.GetLinux().GetSecurityContext().GetNamespaceOptions().GetNetwork() == pb.NamespaceMode_NODE
 
 	// Don't use SELinux separation with Host Pid or IPC Namespace or privileged.
 	if hostPID || hostIPC {
 		processLabel, mountLabel = "", ""
 	}
+
+	if hostNet {
+		processLabel = ""
+	}
+
 	defer func() {
 		if err != nil {
 			err2 := s.StorageRuntimeServer().DeleteContainer(containerInfo.ID)


rhatdan:
I really would like to talk about why we are doing this.  SELinux should not care about which network the container is running on.  If it is running on the host network, the confinement should work fine, and we should NOT disable it.

haircommander:
> I really would like to talk about why we are doing this. SELinux should not care about which network the container is running on. If it is running on the host network, the confinement should work fine, and we should NOT disable it.
Maybe chat at contributors meeting? I know @matthias50 found trouble without this patch in running the e2e tests with SELinux on (hence my backport), as well as @mrunalp and @umohnani8 found it when debugging openshift issues (which lead to the original patch)

rhatdan:
I have all day meetings on Thursday, but will try to sneak away.  Could we turn it back on in master, and see what is failing, then I can look at it?  Are we seeing failings in the 1.15 branch now?

haircommander:
We don't have ci that tests with SELinux in upstream k8s (yet...) we could test in release-1.16 and see if e2e-aws fails though

matthias50:
> I have all day meetings on Thursday, but will try to sneak away. Could we turn it back on in master, and see what is failing, then I can look at it? Are we seeing failings in the 1.15 branch now?

@rhatdan, Here is a copy/paste of what I originally posted in the crio slack channel regarding this issue:

Hi, so one of the tests we run to make sure our Kubernetes automation hasn't done something foolish is the upstream Kubernetes conformance tests. It looks like crio 1.15 wth SELinux enabled breaks one of the test cases. We are seeing this:

```
STEP: Verifying /etc/hosts content of container is not kubelet-managed for pod with hostNetwork=true
Dec 5 13:36:25.070: INFO: ExecWithOptions {Command:[cat /etc/hosts] Namespace:e2e-kubelet-etc-hosts-8163 PodName:test-host-network-pod ContainerName:busybox-1 Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:false}
Dec 5 13:36:25.070: INFO: >>> kubeConfig: /root/.kube/config
Dec 5 13:36:25.789: INFO: Exec stderr: "cat: can't open '/etc/hosts': Permission denied"
Dec 5 13:36:25.789: INFO: Unexpected error occurred: command terminated with exit code 1
```
From looking at the test https://github.com/kubernetes/kubernetes/blob/v1.15.6/test/e2e/common/kubelet_etc_hosts.go, this makes sense as they are using a hostPath volume. Are you hitting this same issue on your end?

Then @haircommander replied with:
> we would if we ran e2e with selinux :confused: it's been on the backlog for a bit now, but bigger fires keep popping up
(note: we run openshift tests with selinux, but obviously that's not vanilla kube)
but that's interesting, having hostNetwork=true should allow the pod to touch /etc/hosts.
based on my understanding of what we're doing with selinux labelling
https://github.com/cri-o/cri-o/pull/3021 nevermind, wasn't backported

rhatdan:
This error indicates that the /etc/hosts file volume mounted into the container is not labeled correctly.

/etc/hosts on the host is labeled net_conf_t
```
$ ls -lZ /etc/hosts
-rw-r--r--. 1 root root system_u:object_r:net_conf_t:s0 158 Oct  9 04:18 /etc/hosts
```

Which is readable with the container label.
```
$ sesearch -A -s container_t -t net_conf_t  -c file -p read
allow container_domain net_conf_t:file { getattr ioctl lock open read };
```
Using podman this works perfectly:

```
podman run --net=host fedora cat /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
```

In the case of podman the /etc/hosts and /etc/resolv.conf from the hosts are actually copied into the container private memory, even allowing the container to write to the files.

I would guess this volume is coming from some other location then /etc/hosts and is a volume from the some directory under kubernetes.

I would guess that

matthias50:
> /etc/hosts on the host is labeled net_conf_t

I get the same result:

```
$ ls -lZ /etc/hosts
-rw-r--r--. root root system_u:object_r:net_conf_t:s0  /etc/hosts
```

> I would guess this volume is coming from some other location then /etc/hosts and is a volume from the some directory under kubernetes.

The test seems to have two volume mounts for the host level /etc/hosts which map to two different locations in the container. See https://github.com/kubernetes/kubernetes/blob/7015f71e75f670eb9e7ebd4b5749639d42e20079/test/e2e/common/kubelet_etc_hosts.go#L200. It looks like it is mounting /etc/hosts -> /etc/hosts and /etc/hosts -> /etc/hosts-original. Could this multi mounting somehow mess things up?

rhatdan:
The test should be fixed to set the SELinux label to container_file_t.

chcon -t contianer_file_t /root/.kube/config/etc/hosts 

Or whatever the source path was, then the tests should pass.

haircommander:
@mrunalp @umohnani8 @rhatdan @saschagrunert PTAL

as true as @rhatdan's comments are, this patch is in 1.16 and 1.14, which I find incredibly awkward. I think we should fix root cause in kubernetes, but in the meantime I think this patch should be merged

rhatdan:
Holding my nose
LGTM
/approve
As long as this is fixed in master branch in 1.17

Issue #3079, opened 2020-01-07
Unexpected behavior with empty seccomp profile


benclarkwood:
**Description**

When running cri-o with seccomp enabled, and a default profile specified, the runtime falls back to "unconfined" when no profile is specified.

**Steps to reproduce the issue:**
1. Enable seccomp
2. Specify a default profile
3. Create a Pod

**Describe the results you received:**
Empty profile default to `unconfined`

**Describe the results you expected:**
Empty profile defaulting to `runtime/default`

**Output of `crio --version`:**
1.14.9

saschagrunert:
Hey @benclarkwood, thank you for opening the issue. As already discussed I will assign the issue to you since you've state that you'd be happy to work on this issue.

I took my time to play around with seccomp and AppArmor profiles and would expect the same as described by @benclarkwood. Generally it was not possible for me to find any generalized description of:

- If a profile is specified to a pod, it applies to all containers as well
- If a profile is specified to a container, it overwrites the pods profile
- We always default to `runtime/default`

I really would like to enforce this from a security perspective and document it properly in a dedicated security section inside this repository. WDYT?

saschagrunert:
I proposed the change for AppArmor here: https://github.com/cri-o/cri-o/pull/3083

haircommander:
as we've seen [here](https://github.com/cri-o/cri-o/pull/3342) this is actually expected behavior, and is an invariant of the CRI. As such, it is not a problem in CRI-O. What do you think about closing this issue and opening one to track in upstream kube @benclarkwood

haircommander:
@saschagrunert would the upstream seccomp work fix this issue? i.e: are we targeting a non-empty default profile for 1.19 or 1.20?

saschagrunert:
> @saschagrunert would the upstream seccomp work fix this issue? i.e: are we targeting a non-empty default profile for 1.19 or 1.20?

Not right now. I don't think that the CRI will change without preserving the backwards compatibility: https://github.com/kubernetes/cri-api/blob/1eae59a7c4dee45a900f54ea2502edff7e57fd68/pkg/apis/runtime/v1alpha2/api.proto#L287

I suggest that we close this issue.

haircommander:
users can override the empty profile behavior by configuring `seccomp_use_default_when_empty` in 1.20 https://github.com/cri-o/cri-o/pull/4212

PR #3083, opened 2020-01-08
Change AppArmor profile handling to fallback to the default

saschagrunert:
If no AppArmor profile is specified, but AppArmor is supported, then we
automatically fallback to `runtime/default`, which is the profile
configured in `crio.conf`.

Beside this we now fail on startup if the profile provided via the
configuration does not exist. This ensures that we can avoid failures
during runtime.

Relates to #3079


rhatdan:
LGTM

TomSweeneyRedHat:

PR #3091, opened 2020-01-09
Add new NSType for available namespaces

saschagrunert:
We now add an additional abstraction layer on of of available namespace
types by defining a new type `NSType`, which is basically just a string.
This ensures that we (from a programming perspective) only can pass
around the available namespaces defined in `namespaces.go`.


haircommander:
Nice, LGTM

PR #3092, opened 2020-01-09
Use UUID generator for namespace path

saschagrunert:
We already use the dependency for logging interception so we can use it
for namespace path generation as well.


haircommander:
Clever!
/lgtm

PR #3096, opened 2020-01-10
Destroy the pod's network when it can't be restored

umohnani8:
If a pod cannot be restored after restart, destroy the
pod network and release the IPs for fututre use


umohnani8:
@mrunalp PTAL

PR #3098, opened 2020-01-10
server: create cgroupns when running on cgroup v2

giuseppe:
when running on a cgroup 2 host, create automatically the cgroupns.

This is being discussed as part of the cgroup v2 support KEP


TomSweeneyRedHat:
LGTM assuming happy tests

Issue #2984, opened 2019-11-19
openshift: `error reserving pod name` causing pods to take a long time to come up


dustymabe:
**Description**

Pods are taking a long time to come up because of `FailedCreatePodSandBox` with the text: `error reserving pod name` as is found [here](https://github.com/cri-o/cri-o/blob/26654d373138e9c93083777391d5f9c4932c83a0/internal/lib/container_server.go#L575-L583).

This is happening a lot but here are two snippets of output from `oc describe pod` that show it takes a long time for the container to be able to get past this: 

<details>

```
Events:
  Type     Reason                  Age                  From                           Message
  ----     ------                  ----                 ----                           -------
  Normal   Scheduled               <unknown>            default-scheduler              Successfully assigned openshift-authentication-operator/authentication-operator-54f5d6c8c9-qpmd5 to dmabe-r662f-master-0
  Warning  FailedCreatePodSandBox  17m                  kubelet, dmabe-r662f-master-0  Failed create pod sandbox: rpc error: code = DeadlineExceeded desc = context deadline exceeded
  Warning  FailedCreatePodSandBox  17m                  kubelet, dmabe-r662f-master-0  Failed create pod sandbox: rpc error: code = Unknown desc = error reserving pod name k8s_authentication-operator-54f5d6c8c9-qpmd5_openshift-authentication-operator_45d93718-e90f-4888-bad6-660cf9c37a87_0 for id f0325581e9547f0087b6e1f6bf1e8f15d5c926f7b1d6694582129358ff54dfb8
  Warning  FailedCreatePodSandBox  16m                  kubelet, dmabe-r662f-master-0  Failed create pod sandbox: rpc error: code = Unknown desc = error reserving pod name k8s_authentication-operator-54f5d6c8c9-qpmd5_openshift-authentication-operator_45d93718-e90f-4888-bad6-660cf9c37a87_0 for id 5910bcf2356ba371ea1e52b721c45d87ade6606b80806755b6c57b5d16f5eafc
  Warning  FailedCreatePodSandBox  16m                  kubelet, dmabe-r662f-master-0  Failed create pod sandbox: rpc error: code = Unknown desc = error reserving pod name k8s_authentication-operator-54f5d6c8c9-qpmd5_openshift-authentication-operator_45d93718-e90f-4888-bad6-660cf9c37a87_0 for id 3b4ba22f8384095d4632e5e3c58623a88d8df60d10aa4f00b236c4d39f4a1c62
[...]

PR #3102, opened 2020-01-11
[1.16] Check for context errors before returning from longer requests

mrunalp:
This is a possible fix for #2984 

The issue seems to be that the kubelet times out on the create request but crio doesn't error out when kubelet client times out. So, we end up with the pod being created and then kubelet keeps on trying to create the pod again and it can't since the name is reserved.

With the change, we error out on timeouts so the pod creation fails and the name is unreserved so later attempts by the kubelet to create a pod or container may succeed.

We may want to add more of these checks around long running operations rather than just at the end in follow-ups. 

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -623,6 +623,10 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 		log.Errorf(ctx, "%v", err)
 	}
 
+	if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+		return nil, ctx.Err()
+	}
+
 	log.Infof(ctx, "Created container: %s", container.Description())
 	resp := &pb.CreateContainerResponse{
 		ContainerId: containerID,
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -673,6 +673,10 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 		log.Errorf(ctx, "%v", err)
 	}
 
+	if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+		return nil, ctx.Err()
+	}
+
 	log.Infof(ctx, "ran pod sandbox with infra container: %s", container.Description())
 	resp = &pb.RunPodSandboxResponse{PodSandboxId: id}
 	return resp, nil


laurelnaiad:
Seems to fix the issue as I mentioned experiencing it. The `name is reserved` error was pretty frequent (though I won't bore you with the log lines)... now it's gone.

```log
# (before this PR)
  Normal   Created                 52m                  kubelet, ubu  Created container kube-controller-manager
  Normal   Started                 52m                  kubelet, ubu  Started container kube-controller-manager
  Warning  FailedCreatePodSandBox  44m                  kubelet, ubu  Failed to create pod sandbox: rpc error: code = Unknown desc = error reserving pod name k8s_kube-controller-manager-ubu_kube-system_ff3dda92fdafad8918c8b8c81e70d5fc_6 for id f37693177f3520cee653f9276c6560f481a3616fb493b76a4c7917d1046db31d: name is reserved
  Normal   SandboxChanged          44m (x2 over 44m)    kubelet, ubu  Pod sandbox changed, it will be killed and re-created.
  Normal   Pulled                  44m                  kubelet, ubu  Container image "k8s.gcr.io/kube-controller-manager:v1.17.0" already present on machine
  Normal   Created                 44m                  kubelet, ubu  Created container kube-controller-manager
  Normal   Started                 44m                  kubelet, ubu  Started container kube-controller-manager
# deployed this PR:
  Normal   SandboxChanged          12m                  kubelet, ubu  Pod sandbox changed, it will be killed and re-created.
  Normal   Pulled                  12m                  kubelet, ubu  Container image "k8s.gcr.io/kube-controller-manager:v1.17.0" already present on machine
  Normal   Created                 12m                  kubelet, ubu  Created container kube-controller-manager
  Normal   Started                 12m                  kubelet, ubu  Started container kube-controller-manager
  Normal   Pulled                  7m41s                kubelet, ubu  Container image "k8s.gcr.io/kube-controller-manager:v1.17.0" already present on machine
  Normal   Created                 7m40s                kubelet, ubu  Created container kube-controller-manager
  Normal   Started                 7m40s                kubelet, ubu  Started container kube-controller-manager
  Normal   SandboxChanged          2m19s                kubelet, ubu  Pod sandbox changed, it will be killed and re-created.
  Normal   Pulled                  2m12s                kubelet, ubu  Container image "k8s.gcr.io/kube-controller-manager:v1.17.0" already present on machine
  Normal   Created                 2m10s                kubelet, ubu  Created container kube-controller-manager
  Normal   Started                 2m10s                kubelet, ubu  Started container kube-controller-manager
  ```

**Comment by mrunalp** [comment]
Awesome, thanks for testing!

**Comment by laurelnaiad** [comment]
My pleasure! Thanks for making all the awesomeness!

File: server/container_create.go [hunk 1]
```diff
@@ -623,6 +623,10 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 		log.Errorf(ctx, "%v", err)
 	}
 
+	if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+		return nil, ctx.Err()
+	}
+
```
giuseppe: is there any possibility that the request is cancelled after this check and before the response is received?

umohnani8:
LGTM
/retest

haircommander:
/retest

LGTM, good find @mrunalp !

server/container_create.go
mrunalp: Yeah that is possible but that should be okay since we end up cleaning the pod in CRI-O which is the main goal of this PR.
The issue earlier was we weren't cleaning up at all so kubelet crio were out of sync w.r.t. the pod state. Also, I think we need these checks sprinkled through the function more as a follow up.

mtrmac:
> The issue seems to be that the kubelet times out on the create request but crio doesn't error out when kubelet client times out. So, we end up with the pod being created and then kubelet keeps on trying to create the pod again and it can't since the name is reserved.

Do you mean:
- Kubelet sends a request
- CRI-O successfully performs that request, and sends a success response
- Kubelet times out before reading the success response
- Kubelet tries again, and this time the operation fails because the name is already allocated

If so, I can’t see how this PR truly fixes the problem; it does narrow the failure timing significantly (assuming the CRI-O context timeout and the Kubelet timeouts are very similar), but it doesn’t eliminate it — if the timeout _value_ is the same for the Kubelet and CRI-O, it starts for the Kubelet at the latest when the request is sent, but for CRI-O at the earliest when it is received, so the Kubelet is _still_ going to timeout first.

It’s not even a timeout matter at all, the response could be lost or corrupted in the network during transfer.

AFAICS, in the above scenario, fundamentally Kubelet must be able to detect that “the operation has succeeded previously” (i.e. notice that the name is reused and that implies this step can now be skipped) or start a separate operation with no shared state with the one that was considered a failure (i.e. allocate an entirely new, different, pod name, and eventually somehow clean up the orphaned resources that were successfully allocated but the Kubelet doesn’t actually know about). (I haven’t looked into the details, that may well involve extending the CRI.)

mtrmac:
… or maybe the container names don’t need to exist at all? `CreateContainerResponse` only returns an ID, so is perhaps the need for a deterministic name something CRI-O decided to introduce and that can just be abandoned?

I clearly don’t know anything about how this all works nor about fixing this correctly, it’s just the combination of the motivation of the PR + the code change that don’t fit my very limited understanding.

mrunalp:
@mitr you are absolutely right.
> If so, I can’t see how this PR truly fixes the problem; it does narrow the failure timing significantly (assuming the CRI-O context timeout and the Kubelet timeouts are very similar), but it doesn’t eliminate it — if the timeout value is the same for the Kubelet and CRI-O, it starts for the Kubelet at the latest when the request is sent, but for CRI-O at the earliest when it is received, so the Kubelet is still going to timeout first.

Yes, the timeout value is the same, so the kubelet should be timing out first.



> AFAICS, in the above scenario, fundamentally Kubelet must be able to detect that “the operation has succeeded previously” (i.e. notice that the name is reused and that implies this step can now be skipped) or start a separate operation with no shared state with the one that was considered a failure (i.e. allocate an entirely new, different, pod name, and eventually somehow clean up the orphaned resources that were successfully allocated but the Kubelet doesn’t actually know about). (I haven’t looked into the details, that may well involve extending the CRI.)

There is a field called attempt number in the pod request. If the kubelet bumps that up on timeouts then we wouldn't have these clashes. We would still have to cleanup any older created resources from previous attempts if they escaped due to the _we returned success but kubelet timed out before it received success_ race. The current PR doesn't fix that race but is an attempt to reduce the amount of leaks compared to status quo.

I also thought about reusing previously created pod as an option. But I worry that it will be difficult to get right compared to having the attempt number be different from the kubelet or us appending a
random string at the end of the name and deleting an old pod when we see one requested with the same parameters.

server/container_create.go
mrunalp: @giuseppe actually after rereading this comment you are right we still have that race. See my response to @mtrmac on thoughts to fix this via bumping up attempt number or on crio side adding a random string to name and doing cleanup on older leaked resources.

mrunalp:
Thinking about the reusing older pod scenario, I think if we add a boolean that we are going to return success before returning success response, we can go off that boolean to decide whether to reuse the older created pod or cleanup and create a new one. Here are some scenarios:

Scenario 1:
1. kubelet requests name pod attempt 1
2. crio goes through all the steps include checking ctx for error. There are no errors.
3. crio sets pod succeeded to true for this pod name pod_1
4. kubelet times out 
4. crio returns response.
5. kubelet sends new request for name pod attempt 1
6. crio checks that pod_1 is already present and succeeded: true. crio also updates status of the pod to make sure it is running.
7. crio returns this pod id before kubelet times out this time.

Scenario 2:
1. kubelet requests name pod attempt 1
2. crio goes through all the steps include checking ctx for error. There are no errors.
3. crio sets pod succeeded to true for this pod name pod_1
4. kubelet times out.
5. kubelet sends new request for name pod attempt 1
6. crio see another request for same pod but it hasn't yet set boolean for first request so it cleans up the current pod 
7. crio returns response to first request but that's ignored and there is no cleanup since it is returning success.
8. crio services the new request by creating a new pod.

In Scenario 2, we tear down a pod which could have been reused, but the race window is very narrow.

In a similar scenario where we didn't set boolean to true for the first request by the time second shows up then we clean up the first pod.

If we keep on timing out in scenario 2 like situation we have to be careful with not introducing more races.

@giuseppe @mtrmac wdyt?

giuseppe:
> * kubelet requests name pod attempt 1
> 
> * crio goes through all the steps include checking ctx for error. There are no errors.
> 
> * crio sets pod succeeded to true for this pod name pod_1
> 
> * kubelet times out
> 
> * crio returns response.
> 
> * kubelet sends new request for name pod attempt 1
> 
> * crio checks that pod_1 is already present and succeeded: true. crio also updates status of the pod to make sure it is running.
> 
> * crio returns this pod id before kubelet times out this time.

would it be possible to just check whether there are existing containers in the requested pod.  If there are no containers created, and the pod is created, then assume the previous request timed-out and be successful this time.  If there are already containers, then something probably went wrong.

mtrmac:
To complicate this further:

The code being patched in this PR creates a _container_; AFAICS the pod is created via `RunPodSandbox` earlier. I can’t see any code path in through which `Server.CreateContainer` ends up failing at `ContainerServer.ReservePodName` and fail with `error reserving pod name` (that can happen through `Server.RunPodSandbox`). So, I can’t see how this change can be improving #3102; my best guess is that the failure observed was actually `error reserving ctr name`, but I’m not at all sure.

---

> Thinking about the reusing older pod scenario, I think if we add a boolean that we are going to return success before returning success response, we can go off that boolean to decide whether to reuse the older created pod or cleanup and create a new one.

Vaguely, yes. The way I think about this, each “create” operation should be idempotent, so that repeating it if it succeeded just results in a “yup, already done this”, and the conceptually easiest way to do that is to have the request come with a caller-generated unique value (“request UUID”) that is never reused. If the pod name is never reused to work this way (I haven’t checked), fine.

> Here are some scenarios:
> 
> Scenario 1:
> 
> 1. kubelet requests name pod attempt 1

(Attempt number exists in `PodSandboxMetadata`, but not in `CreateContainerRequest`, so it can’t impact this code path.)

> 2. crio goes through all the steps include checking ctx for error. There are no errors.
> 3. crio sets pod succeeded to true for this pod name pod_1
> 4. kubelet times out
> 5. crio returns response.
> 6. kubelet sends new request for name pod attempt 1
> 7. crio checks that pod_1 is already present and succeeded: true. crio also updates status of the pod to make sure it is running.

(What is this “update”? An edit of the recorded state? A verification that the actual system state matches the “pod_1 already present” record?)

mtrmac:
As a general comment, is this really the very first time error handling / Kubelet vs. CRI-O getting out of sync has been encountered? Isn’t there some kind of design document or previous discussions about these issues?

From a quick search / skim, https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1alpha2/api.proto doesn’t quite say (apart from mentioning the idempotency idea for _stopping/removing_ things), and https://github.com/kubernetes/kubernetes/blob/release-1.5/docs/proposals/container-runtime-interface-v1.md says
> The container runtime must not apply any transition (such as starting a new container) unless explicitly instructed by Kubelet. It is Kubelet's responsibility to enforce garbage collection, restart policy, and otherwise react to changes in lifecycle.

which kind of suggests that this might be a Kubelet responsibility, not a CRI-O bug.

laurelnaiad:
> ... So, I can’t see how this change can be improving #3102; my best guess is that the failure observed was actually `error reserving ctr name`, but I’m not at all sure.

Assuming you meant #2984 (since this PR is 3102), the OP and I both reported `error reserving pod name / name is reserved`. I am aware of ever having had an error reserving `ctr` name, though it is possible I overlooked it. The ones I noticed were all in re: pods.

mtrmac:
> > ... So, I can’t see how this change can be improving #3102; my best guess is that the failure observed was actually `error reserving ctr name`, but I’m not at all sure.
> 
> Assuming you meant #2984 (since this PR is 3102)

Yes. Thanks.

mrunalp:
We looked a bit more and kubelet should be able to detect rogue older pods running and remove them. That said, I think that this is still useful as it helps with a faster recovery from kubelet.

laurelnaiad:
> kubelet should be able to detect rogue older pods running and remove them

In my experience, running (with crun, not runc, fyi, in case that matters) without this patch can render a cluster unstartable after a reboot or after deleting certain pods. At first I thought they would always recover eventually, but it seems like I may have just been believing what the OP from the other issue had reported. My more recent experience comes from having forgotten to apply the patch after rebuilding the cluster.  It seems to afflict kube-scheduler and kube-controller-manager most frequently.

rhatdan:
@mrunalp Needs a rebase?

umohnani8:
Closing in favour of https://github.com/cri-o/cri-o/pull/3354

PR #3112, opened 2020-01-14
Fix man page header

saschagrunert:
The header was wrongly formatted which is now fixed.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #3115, opened 2020-01-15
Take total_inactive_file into consideration for memory usage

### LINKED ISSUE #3106: MemoryUsage.WorkingSetBytes should take total_inactive_file into consideration
containerd/CRI defines workingset as `usage - total_inactive_file`:

https://github.com/containerd/cri/blob/982e767745a9b0c8ea5f82d9795ab3c37f21c0f5/pkg/server/container_stats_list_unix.go#L84-L96

OTOH, CRI-O doesn't seem taking `total_inactive_file` into consideration:
https://github.com/cri-o/cri-o/blob/3314d8abfd80132bf89e408384ff6683b9799e5e/server/container_stats.go#L42

saschagrunert:
We should take the `total_inactive_file` into consideration
when providing the `WorkingSetBytes` to align with other
container runtimes.

Closes #3106

### DIFF
--- a/internal/oci/oci_linux.go
+++ b/internal/oci/oci_linux.go
@@ -3,9 +3,11 @@
 package oci
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -110,6 +112,18 @@ func (r *runtimeOCI) containerStats(ctr *Container, cgroup string) (*ContainerSt
 		stats.NetInput, stats.NetOutput = getContainerNetIO(netNsPath)
 	}
 
+	totalInactiveFile, err := getTotalInactiveFile()
+	if err != nil { // nolint: gocritic
+		logrus.Warnf("error in memory working set stats retrieval: %v", err)
+	} else if stats.MemUsage > totalInactiveFile {
+		stats.WorkingSetBytes = stats.MemUsage - totalInactiveFile
+	} else {
+		logrus.Debugf(
+			"unable to account working set stats: total_inactive_file (%d) > memory usage (%d)",
+			totalInactiveFile, stats.MemUsage,
+		)
+	}
+
 	return stats, nil
 }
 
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
LGTM

haircommander:
LGTM, PTAL @giuseppe

internal/oci/oci_linux.go
```diff
@@ -162,3 +176,35 @@ func metricsToCtrStats(c *Container, m *cgroups.Metrics) *ContainerStats {
 		PIDs:        pids,
 	}
 }
+
+// getTotalInactiveFile returns the value if `total_inactive_file` as integer
+// from `/sys/fs/cgroup/memory/memory.stat`. It returns an error if the file is
+// not parsable.
+func getTotalInactiveFile() (uint64, error) {
+	const memoryStat = "/sys/fs/cgroup/memory/memory.stat"
```
giuseppe: not sure how we can deal with it in cgroup v2, as this information is not present in the root cgroup and we should probably lookup every child.  For now could we just add a check and return 0 if it is under cgroup v2?

saschagrunert: Yes sure, I will add a check for cgroups v2.

internal/oci/oci_linux.go
```diff
@@ -162,3 +176,40 @@ func metricsToCtrStats(c *Container, m *cgroups.Metrics) *ContainerStats {
...
 }
+
+// getTotalInactiveFile returns the value if `total_inactive_file` as integer
+// from `/sys/fs/cgroup/memory/memory.stat`. It returns an error if the file is
+// not parsable.
+func getTotalInactiveFile() (uint64, error) {
+	// no cgroupv2 support right now
+	if isV2, err := cgroups.IsCgroup2UnifiedMode(); err == nil || isV2 {
+		return 0, nil
+	}
```
saschagrunert: @giuseppe is this the right way to use this function? I cannot omit the check for the `err` because the linter will complain that I not checked that error.

giuseppe: yeah I think it is fine.  Thanks for fixing it so quickly!

mrunalp:
We should backport this as well. Thanks!

PR #3119, opened 2020-01-15
remove NET_RAW and SYS_CHROOT capability by default

haircommander:
#### What type of PR is this?

/kind cleanup

This PR is built on top of https://github.com/cri-o/cri-o/pull/2378 and has some bonus additions:

- Increase the security of containers by removing the NET_RAW and SYS_CHROOT capabilty.
- integrate changes introduced by @Chenditang on github to update the crio.conf man page
- Add net.ipv4.ping_group_range to tests so that ping will still work
- fixes to urfave StringSlice handling, allowing StringSlices to properly be specified by EnvVars
- test: replace dnf with yum in system-packages installation, because the ami bump wasn't tested in https://github.com/cri-o/cri-o/commit/fcb6d26d685e2f13c244f7ce5c48d597ae6a82a9, and using the dnf module breaks rhel updates
- test: use a special fedora-ping container for ping_pod_in_pod, as the alpine image has a version of ping that doesn't work with net.ipv4.ip_forward and no NET_RAW

```release-note
CRI-O now runs containers without NET_RAW and SYS_CHROOT capabilities by default
```


haircommander:
Should we have `net.ipv4.ping_group_range` by default?

rhatdan:
Thanks @haircommander for trying to hammer this one home...

docs/crio.8.md
```diff
@@ -139,7 +139,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--decryption-keys-path**="": Path to load keys for image decryption. (default: "/etc/crio/keys/")
 
-**--default-capabilities**="": Capabilities to add to the containers (default: ["CHOWN" "DAC_OVERRIDE" "FSETID" "FOWNER" "NET_RAW" "SETGID" "SETUID" "SETPCAP" "NET_BIND_SERVICE" "SYS_CHROOT" "KILL"])
+**--default-capabilities**="": Capabilities to add to the containers (default: ["CHOWN" "DAC_OVERRIDE" "FSETID" "FOWNER" "SETGID" "SETUID" "SETPCAP" "NET_BIND_SERVICE" "KILL"]) (default: [])
```
giuseppe: should we drop the second `(default: [])`?

internal/pkg/criocli/criocli.go
```diff
@@ -535,8 +535,9 @@ func getCrioFlags(defConf *libconfig.Config, systemContext *types.SystemContext)
 			Usage:     fmt.Sprintf("Path to default mounts file (default: %q)", defConf.DefaultMountsFile),
 			EnvVars:   []string{"CONTAINER_DEFAULT_MOUNTS_FILE"},
 			TakesFile: true,
+			Hidden:    true,
 		},
-		&cli.StringFlag{
+		&cli.StringSliceFlag{
 			Name:    "default-capabilities",
 			Usage:   fmt.Sprintf("Capabilities to add to the containers (default: %q)", defConf.DefaultCapabilities),
```
saschagrunert: We should use the `Value` here to define the defaults.

docs/crio.8.md
saschagrunert: Yes, this can be fixed with https://github.com/cri-o/cri-o/pull/3119#discussion_r367373186

haircommander:
> why do we drop `CAP_SYS_CHROOT` as part of this change?

it doesn't seem to have a negative effect, so why not :man_shrugging: I should update the PR description though, this pr is doing a lot

haircommander:
So I'm still not positive what's going on here, but here's a quick diagnostic.

For some reason, busy box's ping on the AMI is getting EPERM'd, despite having the sysctl. @mrunalp and I were able to debug and find that much. Copying a fedora binary to a fedora container and running from there works. I can only guess busybox's has some need for NET_RAW other than what ping_group_range fixes... the weird part is the classic "it works on my machine", where my fedora 30 host is able to pass the test. I am guessing it's a missing shared library? I will  attempt other solutions here (including at worst adding NET_RAW for this test) tomorrow

rhatdan:
Have you tried installing the busybox ping on your system and seeing if it works?
Is the busybox ping setuid?  Maybe removing that flag will get it to do the right thing?
is this busybox ping doing a ping while in a container?

haircommander:
> Have you tried installing the busybox ping on your system and seeing if it works?
it's hard to tell what version of busybox is in the container, as well as which version of ping (Apparently there are multiple)
> Is the busybox ping setuid? Maybe removing that flag will get it to do the right thing?
I don't see how difference of setuid in the container would change behavior between hosts though
> is this busybox ping doing a ping while in a container?
yeah, this is in a busybox container from `quay.io/crio/busybox`, I've tried it with alpine and a supposedly better ping from iputils (instructions [here](https://github.com/gliderlabs/docker-alpine/issues/253#issuecomment-299518713)) to also no avail.

worst comes to worse, we can update the container to a fedora-minimal with a ping binary injected. kinda awkward, but at least gets the test to work

rhatdan:
@haircommander Any chance you can get back to the this one?

haircommander:
```
# time="2020-02-10 16:39:58.378053107Z" level=info msg="exec'd [ping -W 1 -c 2 10.88.0.82] in //" file="server/container_execsync.go:43" id=eecc3b7d-9e76-4097-9391-f16e6ac84ca0
# time="2020-02-10 16:39:58.378141416Z" level=debug msg="response: &ExecSyncResponse{Stdout:[FILTERED],Stderr:[],ExitCode:0,}" file="go-grpc-middleware/chain.go:25" id=eecc3b7d-9e76-4097-9391-f16e6ac84ca0
# PING 10.88.0.82 (10.88.0.82) 56(84) bytes of data.
# 64 bytes from 10.88.0.82: icmp_seq=1 ttl=64 time=0.196 ms
# 64 bytes from 10.88.0.82: icmp_seq=2 ttl=64 time=0.084 ms
# 
# --- 10.88.0.82 ping statistics ---
# 2 packets transmitted, 2 received, 0% packet loss, time 1000ms
# rtt min/avg/max/mdev = 0.084/0.140/0.196/0.056 ms
# time="2020-02-10 16:39:58.483453149Z" level=debug msg="request: &ExecSyncRequest{ContainerId:f4395a2542460cc19b9decdd68a1c874de4eb0e0ee99d6f071dc996a06396fbf,Cmd:[ip addr show dev eth0 scope global],Timeout:0,}" file="go-grpc-middleware/chain.go:25" id=67f4ba3c-7e96-4e3d-8425-b6af1d5ea9a6
# time="2020-02-10 16:39:58.627168863Z" level=debug msg="Received container exit code: 0, message: " file="oci/runtime_oci.go:455"
# time="2020-02-10 16:39:58.627529176Z" level=info msg="exec'd [ip addr show dev eth0 scope global] in //" file="server/container_execsync.go:43" id=67f4ba3c-7e96-4e3d-8425-b6af1d5ea9a6
# time="2020-02-10 16:39:58.627644727Z" level=debug msg="response: &ExecSyncResponse{Stdout:[FILTERED],Stderr:[],ExitCode:0,}" file="go-grpc-middleware/chain.go:25" id=67f4ba3c-7e96-4e3d-8425-b6af1d5ea9a6
# time="2020-02-10 16:39:58.662342798Z" level=debug msg="request: &ExecSyncRequest{ContainerId:7a74def0863b96221d46d9a3ec7655d809ca30cffabc21cc82bc0e7ec37a470d,Cmd:[ping6 -W 1 -c 2 1100:200::52],Timeout:0,}" file="go-grpc-middleware/chain.go:25" id=7f52c9de-b1ce-415a-8b40-59c8be9522ae
# time="2020-02-10 16:39:59.093238162Z" level=debug msg="Received container exit code: 2, message: " file="oci/runtime_oci.go:455"
# time="2020-02-10 16:39:59.093657153Z" level=info msg="exec'd [ping6 -W 1 -c 2 1100:200::52] in //" file="server/container_execsync.go:43" id=7f52c9de-b1ce-415a-8b40-59c8be9522ae
# time="2020-02-10 16:39:59.093731088Z" level=debug msg="response: &ExecSyncResponse{Stdout:[],Stderr:[FILTERED],ExitCode:2,}" file="go-grpc-middleware/chain.go:25" id=7f52c9de-b1ce-415a-8b40-59c8be9522ae
# 
# ping: socket: Operation not permitted
# 
```

ping6 gets eperm'd but ping does not ???

rhatdan:
Should we open this as a bug in the kernel?  Does ping6 blow up on Fedora as well?

haircommander:
> Should we open this as a bug in the kernel? Does ping6 blow up on Fedora as well?

no not on fedora, so yeah I think it's a kernel or iputils problem for this version of rhel
`Linux version 3.10.0-1062.9.1.el7.x86_64 (mockbuild@x86-vm-27.build.eng.bos.redhat.com) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-39) (GCC) )`

rhatdan:
So this blows up on RHEL7 only?  And only with non Red Hat based Container images?

haircommander:
> So this blows up on RHEL7 only? And only with non Red Hat based Container images?

actually I updated the image to be a fedora with ping and ip installed

rhatdan:
@haircommander Needs a rebase.  Would love to get this one in.

haircommander:
rebased, I think there were some remaining issues with it, which I may need help in pinning down

nalind:
> > why do we drop `CAP_SYS_CHROOT` as part of this change?
> 
> it doesn't seem to have a negative effect, so why not :man_shrugging: I should update the PR description though, this pr is doing a lot

We use `chroot()` in OpenShift builds.

haircommander:
> > > why do we drop `CAP_SYS_CHROOT` as part of this change?
> > 
> > 
> > it doesn't seem to have a negative effect, so why not man_shrugging I should update the PR description though, this pr is doing a lot
> 
> We use `chroot()` in OpenShift builds.

Ideally, we would then give only openshift builder pods sys_chroot, and leave them out of the default caps

TomSweeneyRedHat:
Another rebase needed @haircommander

haircommander:
the failure is not a flake, for some reason it gets EPERM'd on RHEL

rhatdan:
RHEL7 or RHEL8

haircommander:
7

rhatdan:
Probably the sysctl is not available on RHEL7 boxes, or not suppported,    Is there a way that we can only do this on RHEL8 boxes and leave the capabiltiy on RHEL7.

haircommander:
> Probably the sysctl is not available on RHEL7 boxes, or not suppported, Is there a way that we can only do this on RHEL8 boxes and leave the capabiltiy on RHEL7.

Yeah I think I'll just change the default caps for rhel 7 tests for now

nalind:
The failure in the integration test run doesn't include `system.yml` changes to the sysctl values in its log.  Is the AMI cached?  If so, does it need to be rebuilt?

The AMI test's integration run looks like it's failing with `ping6: permission denied (are you root?)`.  At first glance, it looks like a Fedora kernel will pay attention to `net.ipv4.ping_group_range` even for ipv6 sockets, but a RHEL 7.x kernel will not because that was added in kernel 3.11 in `net: ipv6: Add IPv6 support to the ping socket.`

rhatdan:
@haircommander Can we turn this test off for RHEL7 boxes?

docs/crio.conf.5.md
```diff
@@ -158,6 +163,9 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 
     2) `/usr/share/containers/mounts.conf`: This is the default file read for mounts. If you want CRI-O to read from a different, specific mounts file, you can change the default_mounts_file. Note, if this is done, CRI-O will only add mounts it finds in this file.
 
+**bind_mount_prefix**="prefix"
+  A prefix to use for the source of the bind mounts.  This option would be useful if you were running CRI-O in a container and had the / on the host mounted as /host on your container.  Then if you ran CRI-O with the --bind-mount-prefix=/host option, CRI-O would add /host to any bind mounts it is handed over CRI.  If Kubernetes asked to have /var/lib/foobar bind mounted into the container, then CRI-O would bind mount /host/var/lib/foobar.  Since CRI-O itself is running in a container with / or the host mounted on /host, the container would end up with /var/lib/foobar from the host mounted in the container rather then /var/lib/foobar from the CRI-O container.
+
```
TomSweeneyRedHat: I know it's elsewhere in here, but I'm not a fan of the word "you" in a man page.

haircommander:
/retest

please prow

haircommander:
/retest

integration_crun flaked, e2e_{fedora,rhel} have
```
Build was aborted
Aborted by [8mha:////4O8I/Lh1ZktNUL34o0oGIszT7V2fIWX9t81roO2UrGOzAAAApB+LCAAAAAAAAP9b85aBtbiIQTGjNKU4P08vOT+vOD8nVc83PyU1x6OyILUoJzMv2y+/JJUBAhiZGBgqihhk0NSjKDWzXb3RdlLBUSYGJk8GtpzUvPSSDB8G5tKinBIGIZ+sxLJE/ZzEvHT94JKizLx0a6BxUmjGOUNodHsLgAyhEgYJ/dLi1CL9/ILUvOKMzLQS3eRM3aL8pPwSAHsUFgTNAAAA[0mOpenShift CI Robot
```
which is odd..

haircommander:
:sunglasses: @rhatdan PTAL

docs/crio.conf.5.md
haircommander: I will do a follow up PR fixing all of these for you :slightly_smiling_face:

test/ctr.bats
```diff
@@ -1076,7 +1076,7 @@ function teardown() {
 	run crictl exec --sync $ctr_id grep Cap /proc/1/status
 	echo "$output"
 	[ "$status" -eq 0 ]
-	[[ "$output" =~ 00000000002425fb ]]
+	[[ "$output" =~ 00000000002005fb ]]
```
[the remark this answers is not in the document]
haircommander: @rhatdan where di this value come from?

AkihiroSuda:
Wouldn't this break buildah (chroot-mode) ?

haircommander:
> Wouldn't this break buildah (chroot-mode) ?

it would break using cri-o to run buildah inside of a pod, but a pod can request capabilities to be added, so developers who need buildah pods should specifically give their pod cap_sys_chroot (or admins can give all pods cap_sys_chroot)

pkg/config/config.go
```diff
@@ -550,6 +548,9 @@ func DefaultConfig() (*Config, error) {
 			CtrStopTimeout:           defaultCtrStopTimeout,
 			DefaultCapabilities:      DefaultCapabilities,
 			LogLevel:                 "info",
+			DefaultSysctls:           []string{},
+			DefaultUlimits:           []string{},
+			AdditionalDevices:        []string{},
```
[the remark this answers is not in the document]
haircommander: fixed

docs/crio.conf.5.md
haircommander: jk I fixed it here

haircommander:
this one is new:

`[k8s.io] Security Context when creating containers with AllowPrivilegeEscalation should allow privilege escalation when not explicitly set and uid != 0 [LinuxOnly] [NodeConformance] expand_more`
[ref](https://prow.svc.ci.openshift.org/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3119/test_pull_request_crio_e2e_crun_fedora_cgroupv2/1001/)

/retest

haircommander:
had to rebase because of conflicts

haircommander:
tests passed, tide seems to be hiding, but this is ready
@umohnani8 @rhatdan @saschagrunert @mrunalp PTAL

test/ctr.bats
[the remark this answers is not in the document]
rhatdan: Run a container and `grep CapEff /proc/self/status`
This is the bitflag of the effective capabilities available to the process.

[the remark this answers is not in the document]
saschagrunert: Ref https://github.com/cri-o/cri-o/pull/3477

vrothberg:
I am a bit afraid of removing the two capabilities.

> it would break using cri-o to run buildah inside of a pod with a default config when CAP_SYS_CHROOT is not requested, but a pod can request capabilities to be added, so developers who need buildah pods should specifically give their pod CAP_SYS_CHROOT (or admins can give all pods CAP_SYS_CHROOT)

I agreed with that if we would not have allowed it until now. I fear we're breaking users and should (assuming the removal is carved in stone) communicate such a breaking change for a couple of releases before actually removing it.

saschagrunert:
> I am a bit afraid of removing the two capabilities.
> 
> > it would break using cri-o to run buildah inside of a pod with a default config when CAP_SYS_CHROOT is not requested, but a pod can request capabilities to be added, so developers who need buildah pods should specifically give their pod CAP_SYS_CHROOT (or admins can give all pods CAP_SYS_CHROOT)
> 
> I agreed with that if we would not have allowed it until now. I fear we're breaking users and should (assuming the removal is carved in stone) communicate such a breaking change for a couple of releases before actually removing it.

We could add a deprecation warning per capability to alert users if the caps are not explicitly set in the config. Beside that we could change the release notes block in this PR to mention this deprecation as well.

vrothberg:
> We could add a deprecation warning per capability to alert users if the caps are not explicitly set in the config.

I like this idea :+1:  Adding to logs would help debug the issue.

> Beside that we could change the release notes block in this PR to mention this deprecation as well.

Absolutely, yes! We need to explain why potentially breaking users was worth the change. I think it's easy for NET_RAW but I would have a hard time for SYS_CHROOT. I believe that RPM builds would break.

rhatdan:
Yes launching a buildah container will now require the caller to specify the SYS_CHROOT Capability.  OpenShift Builder is currently running privileged, but eventually it would need the same, and perhaps MKNOD (Although I doubt many workloads require this).

But the whole concept of this change is to move to a higher level of security removing Capabilities that very few container images require, and then adding them to the containers that need it.

This is all about moving towards Papa Bear.  We can always move back to CAP_CHROOT by default, if we find tons of people hitting this, and OpenShift can choose to set its own defaults.

saschagrunert:
> > We could add a deprecation warning per capability to alert users if the caps are not explicitly set in the config.
> 
> I like this idea  Adding to logs would help debug the issue.

Awesome, I'll prepare a PR. :+1: 

> > Beside that we could change the release notes block in this PR to mention this deprecation as well.
> 
> Absolutely, yes! We need to explain why potentially breaking users was worth the change. I think it's easy for NET_RAW but I would have a hard time for SYS_CHROOT. I believe that RPM builds would break.

Maybe @haircommander can help us here by adapting the release notes block? I think we can already mention that it is only deprecated and explain the why's around it. This way I would not have a need for a detailed changelog in my PR. :innocent:

rhatdan:
@haircommander @umohnani8 Thanks for pushing this over the finish line.

We started working on this back on May 16, 2019, nice to see it finally merged.

vrothberg:
> @haircommander @umohnani8 Thanks for pushing this over the finish line.
> 
> We started working on this back on May 16, 2019, nice to see it finally merged.

Absolutely!

Just to re-iterate. I agree a 100 percent with blondie Locks (@rhatdan's talk is not on YouTube yet) being more secure. My main concern is that we need to make sure to communicate it loudly to users and help them resolve issues in case they regress. We need to make clear that this was an improvement for all.

saschagrunert:
On the other hand: If we stick with the new default then this would be the first real good use-case for a `crio config migrate` command. This way we could warn users that the caps have been removed very clearly...

haircommander:
I have updated the release note, PTAL

saschagrunert:
> I have updated the release note, PTAL

Awesome, you could also prefix your note with `action required: ` ([like here](https://github.com/kubernetes/kubernetes/pull/86173)) to have them in a dedicated section inside the release notes.

haircommander:
> > I have updated the release note, PTAL
> 
> Awesome, you could also prefix your note with `action required: ` ([like here](https://github.com/kubernetes/kubernetes/pull/86173)) to have them in a dedicated section inside the release notes.

done :+1:

vrothberg:
> I have updated the release note, PTAL

Great word smithing, @haircommander :+1:

saschagrunert:
Nice, thank you :heart:

haircommander:
> > I have updated the release note, PTAL
> 
> Great word smithing, @haircommander +1

:nerd_face:

rhatdan:
One of my big fears of Kubernetes/CRI-O is that we will not be able to increase the security, for fear of breaking existing workloads.    Which means that we end up having the same security in 10.0 that we had in version 1.0.  In a continuous deployment system, we will slowly update some features that have the change of breaking something,  Good thing is this change will only happen on a major release boundary.  The RPM package will also not modify an existing crio.conf that had not been changed by the admin or a daemonset/controller.   

But in order to make progress in improving things like container separation, at times, we will have to make changes that might break existing container workloads.

Issue #2579, opened 2019-07-08
Configuration migration support


saschagrunert:
I'm thinking about providing a better configuration migration to end-users via something like:

```
> crio config migrate /etc/crio/crio.conf --from v1.14.0 --to v1.15.0
```

This way we would have to keep track of changing configuration from a source code and documentation perspective. We could also implement some sort of automatism in correlation to `crio-wipe`.

WDYT?

rhatdan:
crio config migrate /etc/crio/crio.conf --from v1.14.0 --to v1.15.0

1  No need for parameter /etc/crio/crio.conf, this is default and can be overridden with --config.
2 I don't think we need --to v1.15.0, this should default to current version.  Do you think we would need to migrate to different older versions?
3 --from v1.14.0, should be optional, since this is what is assumed  (Current version -1)

rhatdan:
@haircommander @umohnani8 @vrothberg @mrunalp WDYT?

vrothberg:
> This way we would have to keep track of changing configuration from a source code and documentation perspective.

In theory, we should already be doing that :^)

> crio config migrate /etc/crio/crio.conf --from v1.14.0 --to v1.15.0

What is the problem we're trying to solve and who's the target audience? Would it fill potentially new options? Would it fill the default values for missing ones?

saschagrunert:
> What is the problem we're trying to solve and who's the target audience?

I think the target audience are mostly administrators and package maintainers. The idea is inspired by migration tools provided by large web frameworks to migrate the application from one version to another (like [Angular](https://update.angular.io/)). I mean the technical implementation is not trivial since we would have to versionize the structures somehow to be able to do a conversion. But in the end we could provide the user a CLI tool which provides information like:

```
> Migrating CRI-O config from 1.14 to 1.15
> Renaming 'plugin_dir' to 'plugin_dirs'
...
```

> Would it fill potentially new options?

Yeah, the new config should be fully functional :) We could either ask the user interactively or inform about a new default value has been added.

> Would it fill the default values for missing ones?

Yes it should I guess.

vrothberg:
>  I mean the technical implementation is not trivial since we would have to versionize the structures somehow to be able to do a conversion.

[...]

PR #3123, opened 2020-01-16
Add support for crio drop-in configuration files

saschagrunert:
We now support partial drop-in configuration files which will be assembled together on CRI-O startup.

A default (or user specified via the CLI) directory will be recursively iterated, whereas each file gets applied to the configuration in their processing order. This means that a configuration file named `/etc/crio/crio.conf.d/00-default` has a lower priority than a file `/etc/crio/crio.conf.d/01-my-config`. The configuration provided via `--config,-c` has always a higher priority than those inside the `--config-dir,-d`. The command line parameters have the highest configuration priority at all.

/cc @ereslibre @flavio @Martin-Weiss


Martin-Weiss:
> ult (or user specified via the CL

How do you plan to migrate from customized crio.conf to drop-in configuration files?
Do you plan to change the spec file for crio.conf to overwrite it instead of the creation of "rpmnew"?

saschagrunert:
> How do you plan to migrate from customized crio.conf to drop-in configuration files?
> Do you plan to change the spec file for crio.conf to overwrite it instead of the creation of "rpmnew"?

The plan would be that we do not use `crio.conf` any more and therefore provide a sane default in `/etc/crio/crio.conf.d/00-default`. Then users can customize their setup via `/etc/crio/crio.conf.d/01-change-something` et cetera.

rhatdan:
I wonder how we can merge these with the work we are doing on containers.conf for podman/buildah.

Martin-Weiss:
> 
> 
> > How do you plan to migrate from customized crio.conf to drop-in configuration files?
> > Do you plan to change the spec file for crio.conf to overwrite it instead of the creation of "rpmnew"?
> 
> The plan would be that we do not use `crio.conf` any more and therefore provide a sane default in `/etc/crio/crio.conf.d/00-default`, which is owned by the package. Then users can customize their setup via `/etc/crio/crio.conf.d/01-change-something` et cetera.

But in this case an upgrade will "remove" all customer made adjustments.

The current problem is that a customer had to configure global_auth - so with this change and ignoring the settings in crio.conf - the authentication to the registry will be gone after the upgrade and the next pull will fail.

So what ever change is done - IMO we also have to take care about what was there before and we have to migrate somehow.

vrothberg:
> This means that a configuration file named /etc/crio/crio.conf.d/00-default has a lower priority than a file /etc/crio/crio.conf.d/01-my-config.

Does it mean a reverse alphanumerical order?

test/helpers.bash
```diff
@@ -197,6 +197,8 @@ function setup_test() {
 	fi
 	CRIO_SOCKET="$TESTDIR/crio.sock"
 	CRIO_CONFIG="$TESTDIR/crio.conf"
+	CRIO_CONFIG_DIR="$TESTDIR/crio.conf.d"
+    mkdir "$CRIO_CONFIG_DIR"
```
haircommander: nit: indentation

haircommander:
> > > How do you plan to migrate from customized crio.conf to drop-in configuration files?
> > > Do you plan to change the spec file for crio.conf to overwrite it instead of the creation of "rpmnew"?
> > 
> > 
> > The plan would be that we do not use `crio.conf` any more and therefore provide a sane default in `/etc/crio/crio.conf.d/00-default`, which is owned by the package. Then users can customize their setup via `/etc/crio/crio.conf.d/01-change-something` et cetera.
> 
> But in this case an upgrade will "remove" all customer made adjustments.
> 
> The current problem is that a customer had to configure global_auth - so with this change and ignoring the settings in crio.conf - the authentication to the registry will be gone after the upgrade and the next pull will fail.
> 
> So what ever change is done - IMO we also have to take care about what was there before and we have to migrate somehow.

theoretically, we know what the defaults are and what the user changed values are. Upon rpm upgrade, we could have a command in the crio binary that translates between crio.conf to default-00 + changes-01

saschagrunert:
> > > How do you plan to migrate from customized crio.conf to drop-in configuration files?
> > > Do you plan to change the spec file for crio.conf to overwrite it instead of the creation of "rpmnew"?
> > 
> > 
> > The plan would be that we do not use `crio.conf` any more and therefore provide a sane default in `/etc/crio/crio.conf.d/00-default`, which is owned by the package. Then users can customize their setup via `/etc/crio/crio.conf.d/01-change-something` et cetera.
> 
> But in this case an upgrade will "remove" all customer made adjustments.
>
> The current problem is that a customer had to configure global_auth - so with this change and ignoring the settings in crio.conf - the authentication to the registry will be gone after the upgrade and the next pull will fail.
> 
> So what ever change is done - IMO we also have to take care about what was there before and we have to migrate somehow.

Yes. We already discussed upgrade migration features in https://github.com/cri-o/cri-o/issues/2579 but I'm open to re-visit that discussion.

> > This means that a configuration file named /etc/crio/crio.conf.d/00-default has a lower priority than a file /etc/crio/crio.conf.d/01-my-config.
> 
> Does it mean a reverse alphanumerical order?

Yes, we apply the configuration in filename order, so the priority is reversed from low to high. Does this make sense?

test/helpers.bash
saschagrunert: Fixed :+1:

mrunalp:
I think we should align with how systemd does some of this: https://www.freedesktop.org/software/systemd/man/systemd-system.conf.html.

One thing where we are differing in this implementation vs. systemd is in priority of crio.conf. I think t the files in conf.d should still allow overriding crio.conf. 

Thanks for putting this together. This has been on my list for so long :)

saschagrunert:
> One thing where we are differing in this implementation vs. systemd is in priority of crio.conf. I think the files in conf.d should still allow overriding crio.conf.

Alright :+1:, I changed the configuration priority to this order and adapted the documentation:
1. CLI arguments
1. `crio.conf.d/*`
1. `crio.conf`

docs/crio.8.md
```diff
@@ -125,6 +126,17 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--config, -c**="": Path to configuration file (default: /etc/crio/crio.conf)
 
+**--config-dir, -d**="": Path to the configuration drop-in directory.
+    This directory will be recursively iterated and each file gets applied
+    to the configuration in their processing order. This means that a
+    configuration file named '00-default' has a lower priority than a file
+    '01-my-overwrite'.
+    The global config file, either provided via '--config,-c' or per default in
+    /etc/crio/crio.conf, has always a lower priority than those inside the
```
TomSweeneyRedHat: "has always" to "always has"

docs/crio.8.md
```diff
@@ -125,6 +126,17 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--config, -c**="": Path to configuration file (default: /etc/crio/crio.conf)
 
+**--config-dir, -d**="": Path to the configuration drop-in directory.
+    This directory will be recursively iterated and each file gets applied
+    to the configuration in their processing order. This means that a
+    configuration file named '00-default' has a lower priority than a file
+    '01-my-overwrite'.
```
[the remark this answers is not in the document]
saschagrunert: Fixed as suggested :+1:

docs/crio.8.md
saschagrunert: Fixed as well :+1:

docs/crio.8.md
```diff
@@ -125,6 +126,17 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
...
+    The global config file, either provided via '--config,-c' or per default in
```
TomSweeneyRedHat: I'd drop 'either' in this line, it's a bit redundant with the 'or'

docs/crio.8.md
```diff
@@ -125,6 +126,17 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
...
+    /etc/crio/crio.conf, has always a lower priority than those inside the
+    '--config-dir,-d'.
```
TomSweeneyRedHat: suggest:  "than those inside the '--config-dir,-d'." to "than the files in the directory specified by '--config-dir,-d'."

docs/crio.8.md
```diff
@@ -125,6 +126,17 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
...
+    Beside that, the provided command line parameters still have the highest
+    configuration priority. (default: /etc/crio/crio.conf.d)
```
TomSweeneyRedHat: I'm not sure what this last bit is trying to explain.

completions/fish/crio.fish
```diff
@@ -16,6 +16,16 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l cgroup-manager -r -d 'cgro
...
+complete -c crio -n '__fish_crio_no_subcommand' -l config-dir -s d -r -d 'Path to the configuration drop-in directory.
+    This directory will be recursively iterated and each file gets applied
+    to the configuration in their processing order. This means that a
+    configuration file named \'00-default\' has a lower priority than a file
+    \'01-my-overwrite\'.
+    The global config file, either provided via \'--config,-c\' or per default in
+    /etc/crio/crio.conf, has always a lower priority than those inside the
+    \'--config-dir,-d\'.
+    Beside that, the provided command line parameters still have the highest
+    configuration priority.'
```
TomSweeneyRedHat: see below comments.

internal/pkg/criocli/criocli.go
```diff
@@ -289,6 +294,23 @@ func getCrioFlags(defConf *libconfig.Config, systemContext *types.SystemContext)
...
+			Usage: fmt.Sprintf("Path to the configuration drop-in directory."+`
+    This directory will be recursively iterated and each file gets applied
+    to the configuration in their processing order. This means that a
+    configuration file named '00-default' has a lower priority than a file
+    named '01-my-overwrite'.
+    The global config file, either provided via '--config,-c' or per default in
+    %s, always has a lower priority than those inside the
+    '--config-dir,-d'.
+    Beside that, the provided command line parameters still have the highest
+    configuration priority.`, libconfig.CrioConfigPath),
```
TomSweeneyRedHat: see prior comments.

docs/crio.8.md
saschagrunert: Changed it to:
```
	Beside that, provided command line parameters still have a higher priority
	than any configuration file.
```
I hope this makes more sense.

PR #3124, opened 2020-01-16
Fix generated docs formatting

saschagrunert:
This fixes the markdown as well as the man page formatting for the CLI
documentation.

### DIFF
--- a/completions/fish/crio.fish
+++ b/completions/fish/crio.fish
@@ -41,29 +41,34 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l hooks-dir -r -d 'Set the O
     the semantics of hook injection, see \'oci-hooks(5)\'. CRI-O
     currently support both the 1.0.0 and 0.1.0 hook schemas, although
     the 0.1.0 schema is deprecated.
-
     This option may be set multiple times; paths from later options
     have higher precedence (\'oci-hooks(5)\' discusses directory
     precedence).
-
-	For the annotation conditions, CRI-O uses the Kubernetes
+    For the annotation conditions, CRI-O uses the Kubernetes
     annotations, which are a subset of the annotations passed to the
     OCI runtime. For example, \'io.kubernetes.cri-o.Volumes\' is part of
     the OCI runtime configuration annotations, but it is not part of
     the Kubernetes annotations being matched for hooks.
-
     For the bind-mount conditions, only mounts explicitly requested by
     Kubernetes configuration are considered. Bind mounts that CRI-O
     inserts by default (e.g. \'/dev/shm\') are not considered.'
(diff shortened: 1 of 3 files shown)


rhatdan:
I am fine with this, but does the docs still stretch to fill larger screens?

saschagrunert:
> I am fine with this, but does the docs still stretch to fill larger screens?

Hm, in markdown yes, in the man page we now strip it down to the limited length:
```man
      --global-auth-file="": Path to a file like /var/lib/kubelet/config.json holding credentials necessary for pulling images from secure registries (default: "")

       --grpc-max-recv-msg-size="": Maximum grpc receive message size in bytes (default: 16777216)

       --grpc-max-send-msg-size="": Maximum grpc receive message size (default: 16777216)

       --help, -h: show help

       --hooks-dir="": Set the OCI hooks directory path (may be set multiple times) (default: ["/usr/share/containers/oci/hooks.d"])
           Each '*.json' file in the path configures a hook for CRI-O
           containers. For more details on the syntax of the JSON files and
           the semantics of hook injection, see 'oci-hooks(5)'. CRI-O
           currently support both the 1.0.0 and 0.1.0 hook schemas, although
           the 0.1.0 schema is deprecated.
           This option may be set multiple times; paths from later options
           have higher precedence ('oci-hooks(5)' discusses directory
           precedence).
           For the annotation conditions, CRI-O uses the Kubernetes
           annotations, which are a subset of the annotations passed to the
           OCI runtime. For example, 'io.kubernetes.cri-o.Volumes' is part of
           the OCI runtime configuration annotations, but it is not part of
           the Kubernetes annotations being matched for hooks.
           For the bind-mount conditions, only mounts explicitly requested by
           Kubernetes configuration are considered. Bind mounts that CRI-O
           inserts by default (e.g. '/dev/shm') are not considered. (default: [])

       --host-ip="": Host IPs are the addresses to be used for the host network and can be specified up to two times (default: [])

       --image-volumes="": Image volume handling ('mkdir', 'bind', or 'ignore')
           1. mkdir: A directory is created inside the container root filesystem for
              the volumes.
           2. bind: A directory is created inside container state directory and bind
              mounted into the container for the volumes.
           3. ignore: All volumes are just ignored and no action is taken. (default: mkdir)

       --insecure-registry="": Enable insecure registry communication, i.e., enable un-encrypted and/or untrusted communication.
           1. List of insecure registries can contain an element with CIDR notation to
              specify a whole subnet.
           2. Insecure registries accept HTTP or accept HTTPS with certificates from
              unknown CAs.
           3. Enabling '--insecure-registry' is useful when running a local registry.
              However, because its use creates security vulnerabilities, it should ONLY
              be enabled for testing purposes. For increased security, users should add
              their CA to their system's list of trusted CAs instead of using
              '--insecure-registry'. (default: [])
```

haircommander:
LGTM

PR #3126, opened 2020-01-17
Allow server to start without config

saschagrunert:
We already search required runtime dependencies inside paths so the
existence of global `crio.conf` is not necessary in all cases. We will
now not warn any more if the configuration file does not exist on the
system as well as set default the `log_level` to `info` to provide a
better default startup context.


mrunalp:
This is a logical step over #3123 :+1:

haircommander:
LGTM

PR #3127, opened 2020-01-17
[1.16] fail on network stop

haircommander:
Move stopNetwork call to earlier in sandbox stop function. This means we unconditionally stop the network.
Add sb.*NetworkStopped functions
Create files in infra's persistent dir to restore {,network}Stopped state
update tests accordingly

### DIFF
--- a/internal/lib/sandbox/sandbox.go
+++ b/internal/lib/sandbox/sandbox.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -13,6 +14,11 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 	"k8s.io/kubernetes/pkg/kubelet/dockershim/network/hostport"
+)
+
+var (
+	sbStoppedFilename        = "stopped"
+	sbNetworkStoppedFilename = "network-stopped"
 )
 
 func isSymbolicLink(path string) (bool, error) {
@@ -102,6 +108,7 @@ type Sandbox struct {
 	stopMutex          sync.RWMutex
 	created            bool
 	stopped            bool
+	networkStopped     bool
 	privileged         bool
 	hostNetwork        bool
 }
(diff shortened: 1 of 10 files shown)


server/sandbox_network.go
```diff
@@ -132,10 +133,10 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) {
 	podNetwork, err := s.newPodNetwork(sb)
 	if err != nil {
 		log.Warnf(ctx, err.Error())
-		return
+		return nil
```
mrunalp: We need to return this error as the next call will fail if this doesn't succeed.

internal/lib/sandbox/sandbox.go
```diff
@@ -431,6 +456,52 @@ func (s *Sandbox) Stopped() bool {
...
+		s.stopped = true
+	}
+	if s.fileExistsInInfraDir(sbNetworkStoppedFilename) {
+		s.networkStopped = true
+	}
+}
+
+func (s *Sandbox) fileExistsInInfraDir(filename string) bool {
+	infra := s.InfraContainer()
+	_, err := os.Stat(filepath.Join(infra.Dir(), filename))
```
mrunalp: we should log if it is some other error.

mrunalp:
circle ci is failing on iptables failures from stop pod.

internal/lib/sandbox/sandbox.go
haircommander: fixed

server/sandbox_stop_linux.go
```diff
@@ -45,6 +45,11 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 	stopMutex.Lock()
 	defer stopMutex.Unlock()
 
+	// Clean up sandbox networking and close its network namespace.
+	if err := s.networkStop(ctx, sb); err != nil {
```
saschagrunert: This is where OCI CNI fails with modifying the `ip6tables` chain. I cannot reproduce it on my local machine but in the CI every container stop does not seem to work any more. I think it might be related to docker networking and we may want to switch the integration tests to run completely without docker :thinking:

haircommander:
WIP: trying out https://github.com/cri-o/cri-o/pull/3133 on top of this to see if circle tests pass

saschagrunert:
The `.circleci/config.yml` seems not to contain the full patch, it's still trying to run `docker`.

internal/lib/sandbox/sandbox.go
```diff
@@ -431,6 +456,58 @@ func (s *Sandbox) Stopped() bool {
...
+		if err := s.createFileInInfraDir(sbNetworkStoppedFilename); err != nil {
+			return fmt.Errorf("failed to create state file in container directory. Restores may fail: %v", err)
+		}
+	}
+	return nil
+}
+
+func (s *Sandbox) createFileInInfraDir(filename string) error {
+	infra := s.InfraContainer()
+	_, err := os.Create(filepath.Join(infra.Dir(), filename))
```
rphillips: This will leak a file descriptor. The returned file needs to be closed.

saschagrunert:
I just backported the CI patch to release-1.16 (https://github.com/cri-o/cri-o/pull/3140) to ensure this adds no regressions in the release branch.

internal/lib/sandbox/sandbox.go
haircommander: nice catch, fixed

saschagrunert:
Seems to work now, right?

haircommander:
huzzah!

haircommander:
hm none of the integration tests use pid mode == pod

/hold

saschagrunert:
Do we want to forward port this?

haircommander:
Absolutely!
/cherry-pick release-1.17

haircommander:
Manual cherry pick it is

PR #3128, opened 2020-01-17
[1.16] Add image decryption capabilities

lumjjb:
Add decryption capabilities and also update completions to match 1.16.2.

Signed-off-by: Brandon Lum <lumjjb@gmail.com>


haircommander:
/ok-to-test

hm, I'm a bit hesitant to add a config variable this late in the release. Is there a reason you want/need this in 1.16 and can't wait for 1.17?

lumjjb:
I am pushing my luck here :). 

Well, we were trying to get an offering with decryption around the first half of this year in OpenShift. This is an ongoing discussion right now, but I'd thought of trying to see if it would be feasible to do it cleanly (from code merge perspective). 

We have a couple discussions continuing to see what is possible, or whether we would just wait on an OpenShift that uses 1.17 - depending on the timeframe of that release. Since 4.3 was announced 2 days ago, I assume that 4.4 is going to be a while hehe :)

haircommander:
/retest

fair enough! the code LGTM, I'll let someone else chime in on how feasible it is to add now. my preference is still to wait :slightly_smiling_face:

mrunalp:
My preference is that this goes into 1.17 if possible.

saschagrunert:
> My preference is that this goes into 1.17 if possible.

Yep I agree since 1.17.0 should be available in quite some time.

lumjjb:
Yea, that seems reasonable. Tagging @matthew-johnson1 for tracking.

rhatdan:
@lumjjb are you still working on this, or has this moved on to a more recent CRI-O and we can close?

lumjjb:
Yea, i'll close this! thanks.

PR #3133, opened 2020-01-21
Run integration tests natively in CircleCI

saschagrunert:
This way we don't need docker any more. :)


.circleci/config.yml
```diff
@@ -69,41 +71,26 @@ workflows:
           requires:
             - build
       - docs-validation
-      - ginkgo
       - git-validation
-      - integration:
-          requires:
-            - build
-            - build-test-binaries
+      - integration
```
TomSweeneyRedHat: Is this line valid?

haircommander:
Lgtm, thanks for taking this @saschagrunert

.circleci/config.yml
saschagrunert: Yes :) there are no parameters needed any more

PR #3134, opened 2020-01-21
Fail to start when stream server port already allocated

saschagrunert:
If the streaming server port is already used then we should fail on
CRI-O startup.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #3135, opened 2020-01-21
Add documentation about stream_port="0"

saschagrunert:
We should mention that the port is randomly selected if set to "0".


TomSweeneyRedHat:
LGTM assuming happy tests

PR #3137, opened 2020-01-21
Switch default cgroup manager to systemd

saschagrunert:
This provides a stronger default and implies the recommendation to use `systemd` as cgroup manager.


haircommander:
@saschagrunert if you updated the comment block here, would the release-notes get updated?

saschagrunert:
> @saschagrunert if you updated the comment block here, would the release-notes get updated?

I don't think so since this is already part of 1.18.0. Ah damn this has been slipped through.

haircommander:
growing pains :confused:

haircommander:
we can manually push the gh-pages branch though right?

saschagrunert:
Yes, that would be fine :)

PR #3138, opened 2020-01-21
[1.17] Revert net cleanup and ctx check

mrunalp:
We are seeing possible regressions from these fixes so reverting them.


runcom:
As a side note, we’re rolling back crio to 116 in the Ci to check back

mrunalp:
@runcom These are the only two changes which are somewhat related though I am not 100% convinced. We can merge this and see if it helps or explore other possibilities.

PR #3139, opened 2020-01-22
server: allow an apparmor-unconfined container

tych0:
Currently, it is not possible to start a container with the sentinel
profile "unconfined":

time="2020-01-21 19:29:37.209187136-07:00" level=info msg="assuming user-provided apparmor profile: unconfined"
time="2020-01-21 19:29:37.209263882-07:00" level=fatal msg="config provided AppArmor profile \"unconfined\" not loaded"

This profile doesn't need to be loaded, so checking for it is incorrect.

Signed-off-by: Tycho Andersen <tycho@tycho.ws>


mrunalp:
@saschagrunert ptal.

server/server.go
```diff
@@ -393,7 +393,7 @@ func New(
 				}
 				logrus.Tracef("default apparmor profile contents: %s", profileContent)
 			}
-		} else {
+		} else if config.ApparmorProfile != "unconfined" {
```
saschagrunert: I would prefer in having documentation around this in those places:
https://github.com/cri-o/cri-o/blob/544c64b0afbbf3e8944709313a604cfea5a6c508/pkg/config/template.go#L141-L143
https://github.com/cri-o/cri-o/blob/544c64b0afbbf3e8944709313a604cfea5a6c508/internal/pkg/criocli/criocli.go#L457-L459

Afterwards the CLI config can be generated via:
```
> make clean all
> sudo make docs-generation completions
```

haircommander:
LGTM, thanks @tych0

server/server.go
tych0: Can you elaborate on what you'd like? To me this seems like an implementation detail of crio.

saschagrunert: Yep my intention was to tell the end user how CRI-O behaves if we set `apparmor_profile = "unconfined"`. So it means that AppArmor will be disabled for workloads not specifying an own profile via the annotation, right?

tych0: I'm not sure what "an own profile via the annotation" means; if a task's apparmor label is set to "unconfined" in the kernel, apparmor will not deny anything.

saschagrunert: Yes, but this applies only if we do not specify any profile via Kubernetes Pod’s metadata annotation:
https://kubernetes.io/docs/tutorials/clusters/apparmor/#securing-a-pod

This means if the user specifies a custom AppArmor profile via a k8s pod, then the `apparmor_profile = "unconfined"` option will have no effect:
https://github.com/cri-o/cri-o/blob/544c64b0afbbf3e8944709313a604cfea5a6c508/server/container_create.go#L684-L689

This should be documented somehow if this is the intended behavior :)

tych0: I don't really know if it is the intended behavior or not, but I pushed some docs to that effect :)

saschagrunert: Hehe, thank you :pray: 

I think it could be the intended behavior, at least that is something _I_ would expect to happen. :grin:

PR #3142, opened 2020-01-22
Log path location when using binaries discovered in $PATH

saschagrunert:
Just a small enhancement to increase the debug capabilities.


TomSweeneyRedHat:
LGTM
nice tweak, that will be handy debugging down the road I'm sure.

PR #3144, opened 2020-01-22
Refactor sysctl handling and add unit tests

saschagrunert:
This fixes the CLI default_sysctl string slice handling and refactors
the validation/parsing of the sysctls in to the config package. We now
pre-check if the sysctls are parsable via the CLI to reduce the error
surface. Unit tests for all parsing and validation cases have been added
as well.


haircommander:
nice LGTM

TomSweeneyRedHat:
LGTM but would like another head nod before merging.

PR #3145, opened 2020-01-22
Fix unit tests for rootless runs

saschagrunert:
Running unit tests locally require a valid namespace path as well as pinns which can be faked around.


TomSweeneyRedHat:
LGTM assuming happy tests

PR #3146, opened 2020-01-22
Enable AppArmor tests in CircleCI

saschagrunert:
This also implies that an `--apparmor-profile=""` will now be treated as _the internal default_, which makes sense in my point of view.


saschagrunert:
Funny enough, the static build feature set was broken... :thinking:

PR #3153, opened 2020-01-23
drop host_ip{,s}

haircommander:
Since we now don't parse them to find a bind address, they're not needed anymore. we can drop the cli flag because we haven't released with it yet, and the configuration parameter will just be ignored.


haircommander:
@saschagrunert PTAL

haircommander:
PTAL @saschagrunert

pkg/config/config.go
```diff
@@ -610,6 +613,14 @@ func (c *APIConfig) Validate(onExecution bool) error {
 	if c.GRPCMaxRecvMsgSize <= 0 {
 		c.GRPCMaxRecvMsgSize = defaultGRPCMaxMsgSize
 	}
+	// merge deprecated option HostIP with new option HostIPs
+	// set HostIP to empty to have it no longer printed in configs
+	if c.HostIP != "" {
+		if len(c.HostIPs) < 2 {
```
saschagrunert: What should we do when `len(HostIPs) >= 2` and `HostIP` is set?

haircommander: fail, because we unconditionally append HostIP to HostIPs.

docs/crio.conf.5.md
```diff
@@ -50,7 +50,7 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 ## CRIO.API TABLE
 The `crio.api` table contains settings for the kubelet/gRPC interface.
 
-**host_ip**=""
+**host_ips**=[]
```
mrunalp: what was the reason for limiting to 2?

saschagrunert: I think it was the discussion that Kubernetes works only with 1 IPv4 and 1 additional/optional IPv6 right now.

haircommander:
hey all, so I actually am changing this to totally drop host_ips. since https://github.com/cri-o/cri-o/pull/3165 we don't use them anymore. If we cherry-pick both this and that into 1.17 before we release, we can avoid making the config change all together. WDYT @saschagrunert @mrunalp @umohnani8

saschagrunert:
> hey all, so I actually am changing this to totally drop host_ips. since #3165 we don't use them anymore. If we cherry-pick both this and that into 1.17 before we release, we can avoid making the config change all together. WDYT @saschagrunert @mrunalp @umohnani8

I'm in favor of this change. I think if it's still in the config of end-users it would not be considered, but the missing CLI flag might break end existing setups...

test/docs-validation/main.go
```diff
@@ -31,6 +31,7 @@ var (
 		"plugin_dir",                  // deprecated
 		"runtimes",                    // printed as separate table
 		"manage_network_ns_lifecycle", // deprecated
+		"host_ip",                     // deprecated
```
saschagrunert: I think we don't need this any more since it's gone from the config.

haircommander:
> > hey all, so I actually am changing this to totally drop host_ips. since #3165 we don't use them anymore. If we cherry-pick both this and that into 1.17 before we release, we can avoid making the config change all together. WDYT @saschagrunert @mrunalp @umohnani8
> 
> I'm in favor of this change. I think if it's still in the config of end-users it would not be considered, but the missing CLI flag might break end existing setups...


from what I can see, https://github.com/cri-o/cri-o/pull/2925/commits/f6dff3c6d5e9715b6b866c471d647831b34f056b is only in 1.17.0-rc1, and before we didn't have the flag, which means we technically haven't broken any users using a release branch. 

Do you still want me to keep the flag and mark it hidden so we don't error out?

test/docs-validation/main.go
haircommander: fixed!

saschagrunert:
> > > hey all, so I actually am changing this to totally drop host_ips. since #3165 we don't use them anymore. If we cherry-pick both this and that into 1.17 before we release, we can avoid making the config change all together. WDYT @saschagrunert @mrunalp @umohnani8
> > 
> > 
> > I'm in favor of this change. I think if it's still in the config of end-users it would not be considered, but the missing CLI flag might break end existing setups...
> 
> from what I can see, [f6dff3c](https://github.com/cri-o/cri-o/commit/f6dff3c6d5e9715b6b866c471d647831b34f056b) is only in 1.17.0-rc1, and before we didn't have the flag, which means we technically haven't broken any users using a release branch.
> 
> Do you still want me to keep the flag and mark it hidden so we don't error out?

No I guess that's fine :) Thanks for double checking it.

mrunalp:
Dropping sounds good. We will have to make this deprecated in 1.17 so there is a smooth path to this.

haircommander:
> Dropping sounds good. We will have to make this deprecated in 1.17 so there is a smooth path to this.

won't merging https://github.com/cri-o/cri-o/pull/3171 make a smooth path needless?

PR #3161, opened 2020-01-27
Fix integration tests by adjusting image digest

saschagrunert:
Beside that, I removed the digest test for architectures we do not test at all.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -820,7 +820,7 @@ var _ = t.Describe("Config", func() {
 	t.Describe("UpdateFromPath", func() {
 		It("should succeed with the correct priority", func() {
 			// Given
-			Expect(sut.LogLevel).To(Equal("error"))
+			Expect(sut.LogLevel).To(Equal("info"))
 
 			configDir := t.MustTempDir("config-dir")
 			Expect(ioutil.WriteFile(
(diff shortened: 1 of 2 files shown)


saschagrunert:
I added the unit test fix on top of that PR too to make the CI green again.

haircommander:
LGTM, @nalind PTAL

TomSweeneyRedHat:
Personally I'd rather keep the other arch in place, even if we're not using them at the moment.  I'm fine with letting this go in, but would like a tie breaker from someone else.

nalind:
Not entirely up to speed on what was failing there, but we definitely test on amd64.  Was the error that the digests of the arch-specific images no longer matched the ones in the manifest list when we switched which manifest list we were looking at?

saschagrunert:
> Not entirely up to speed on what was failing there, but we definitely test on amd64. Was the error that the digests of the arch-specific images no longer matched the ones in the manifest list when we switched which manifest list we were looking at?

Yes exactly. :)

nalind:
Well, we should at least keep the amd64 test, because we want to verify that we end up saving both the arch-specific image's manifest and the manifest list.  The updated digest values for the various arches can be dug out of `skopeo inspect --raw docker://docker.io/library/alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178`.

saschagrunert:
Alright, I re-added the amd64 test

PR #3165, opened 2020-01-28
stream server: Bind to all addresses

mrunalp:
If stream server address isn't specified, then we
should bind to all addresses.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -36,7 +36,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	seccomp "github.com/seccomp/containers-golang"
 	"github.com/sirupsen/logrus"
-	knet "k8s.io/apimachinery/pkg/util/net"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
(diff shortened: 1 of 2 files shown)


giuseppe:
/retest

LGTM

haircommander:
rebase to fix test failures

server/server.go
```diff
@@ -415,10 +414,13 @@ func New(
 	s.restore(ctx)
 	s.cleanupSandboxesOnShutdown(ctx)
 
+	var bindAddressStr string
 	bindAddress := net.ParseIP(config.StreamAddress)
 	if bindAddress == nil {
-		hostIPs := s.getHostIPs(config.HostIP)
-		bindAddress = hostIPs[0]
+		// If bind address isn't valid or unset, then listen on all IP addresses
+		bindAddressStr = ""
```
haircommander: do we need this case? can't we just check if bindAddress != nil, and if so set bindAddressStr? `var bindAddressStr string` sets it as ""

mrunalp: oops, yeah. we can do that in a follow cleanup PR.

haircommander: fixing it here https://github.com/cri-o/cri-o/pull/3153

PR #3170, opened 2020-01-29
[1.17] persist exit status

haircommander:
conmon will save the exit file of the container under
/var/lib so it gets persisted and available even on reboot.

Fixes to better handle exit code:

Change ExitCode to be a pointer to distinguish from
exit code 0.
On reboot if we can't get runc state, then attempt
to read exit code from the persisted exit file.
This handles cases where crio was unable to persist
the state for the container but conmon had written
the exit file for the container.
When restoring state during CRI-O startup,
write back the state to disk so that any missed
exit file updates are accounted for.

Signed-off-by: Mrunal Patel mrunalp@gmail.com
cherry-pick of #3041

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -199,7 +199,7 @@ func (c *ContainerServer) Update() error {
 			ctr := c.GetContainer(id)
 			if ctr != nil {
 				// if the container exists, update its state
-				if err := c.ContainerStateFromDisk(c.GetContainer(id)); err != nil {
+				if err := c.ContainerStateFromDisk(ctr); err != nil {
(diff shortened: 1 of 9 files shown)


haircommander:
Note: merging this will break kata, as it uses a conmon flag that doesn't exist in their tests. they are pinning an old version of conmon to fix a breakage of a newer conmon and execsync. I am tracking this in https://github.com/kata-containers/tests/issues/2271
for now, we can push this through here, and work on getting a fix into 1.17 and 1.16 before merging this into master

rhatdan:
You are storing this under /var/lib/containers/storage/DRIVER-containers?

PR #3172, opened 2020-01-30
[1.17] Add support for crio drop-in config files

haircommander:
We now support partial drop-in configuration files which will be
assembled together on CRI-O startup.

A default (or user specified via the CLI) directory will be recursively
iterated, whereas each file gets applied to the configuration in their
processing order. This means that a configuration file named
`/etc/crio/crio.conf.d/00-default` has a lower priority than a file
`/etc/crio/crio.conf.d/01-my-config`. The file provided via
`--config,-c` always has a higher priority than those inside the
`--config-dir,-d`, whereas the command line parameters always have the
highest configuration priority at all.

**- What I did**

**- How I did it**

**- How to verify it**


haircommander:
/retest

love me some network flakes from quay

PR #3176, opened 2020-01-31
Update dependencies

saschagrunert:
Give all dependencies a bump, whereas no direct source code interaction should be needed.


saschagrunert:
I'm pretty sure that we have a selinux based regression in somehow. I will come up with a new PR where I expect nothing to be broken.

PR #3177, opened 2020-01-31
build: make install providing systemd and config

dougsland:
Currently, make install is not installing systemd files and configs.
Based on that, the environment is incomplete when building and
installing the project manually.

**- What I did**: Improved Makefile

**- How I did it**: Changing the install target from Makefile

**- How to verify it**: make && make install (Now I see config and systemd files in the env)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -370,7 +370,7 @@ docs-generation:
 bundle:
 	bundle/build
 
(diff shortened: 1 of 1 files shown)


dougsland:
IMHO, the error doesn't seems related to the patch.

PR #3178, opened 2020-01-31
config: move default version file location to tmpfs

haircommander:
move default version file location to /var/run/crio/version instead of /var/lib/crio/version

as we should really wipe crio after every reboot, so let's put it on a tmpfs


rhatdan:
LGTM
/approve

PR #3185, opened 2020-02-03
Add the container IDs that cri-o assigns to various logs

umohnani8:
Having the extra container/pod ID in the logs helpw with
debugging. Fix the stop sandbox log to save "stopped" instead
of "removed".

Will cherry-pick into the release branches.


umohnani8:
@haircommander @mrunalp PTAL

haircommander:
LGTM

PR #3193, opened 2020-02-03
Setup container environment variables before user

### LINKED ISSUE #3186: cri-o must set home ton $HOME if specified in image
When cri-o is used  in OpenShift, it automatically creates the arbitrary user that OpenShift requires to run a specific image. as an example:
```
sh-4.2$ tail -1 /etc/passwd
1000580000:x:1000580000:0:1000580000 user:/:/sbin/nologin
sh-4.2$
```
However, the home directory is set to "/"  .
```
sh-4.2$ getent passwd $(id -u) | cut -f6 -d:
/
```

When the $HOME environment variable is set in the built container image, it is preferable that the user home matches the same value.

mrunalp:
The user setup code looks at the env to setup HOME.

Fixes #3186 

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -758,6 +758,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		specgen.AddAnnotation("org.opencontainers.image.stopSignal", containerImageConfig.Config.StopSignal)
 	}
 
+	// Add environment variables from image the CRI configuration
+	envs := mergeEnvs(containerImageConfig, containerConfig.GetEnvs())
+	for _, e := range envs {
+		parts := strings.SplitN(e, "=", 2)
+		specgen.AddProcessEnv(parts[0], parts[1])
+	}
+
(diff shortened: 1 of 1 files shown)


haircommander:
LGTM

server/container_create_linux.go
```diff
@@ -758,6 +758,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		specgen.AddAnnotation("org.opencontainers.image.stopSignal", containerImageConfig.Config.StopSignal)
 	}
 
+	// Add environment variables from image the CRI configuration
+	envs := mergeEnvs(containerImageConfig, containerConfig.GetEnvs())
+	for _, e := range envs {
+		parts := strings.SplitN(e, "=", 2)
+		specgen.AddProcessEnv(parts[0], parts[1])
```
TomSweeneyRedHat: as noted elsewhere, possible. albeit unlikely, panic here if there's not an equal sign.

PR #3196, opened 2020-02-04
Update golangci-lint to v1.23.3

saschagrunert:
We also print out the enabled/disabled linters as well as the version
for CI debugging purposes.


TomSweeneyRedHat:

PR #3200, opened 2020-02-04
contrib: 10-crio-bridge.conf change subnet

dougsland:
Currently, the subnet example for crio-bridge is 10.88.0.0/16.
However, libpod project also provides a cni with the same
subnet:

https://github.com/containers/libpod/blob/master/cni/87-podman-bridge.conflist

If both bridges are up and running will be a conflict. This patch
changes the crio-bridge subnet to 10.85.0.0/16.

**- What I did** Change crio-bridge subnet from 10-crio-bridge.conf

**- How I did it** Replacing subnet in 10-crio-bridge.conf

**- How to verify it** Making both bridges up (with same subnet) and run ./hack/local-up-cluster.sh. The kube-dns and kube-dashboard pods couldn't run.


dougsland:
new PR coming, /me forgot git commit -s.

PR #3202, opened 2020-02-05
Bump containers storage to v1.15.8

saschagrunert:
As the title says

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -185,6 +185,8 @@ github.com/containers/storage v1.15.4-0.20191218193401-e18327c57107 h1:p6uPxazL3
 github.com/containers/storage v1.15.4-0.20191218193401-e18327c57107/go.mod h1:v0lq/3f+cXH3Y/HiDaFYRR0zilwDve7I4W7U5xQxvF8=
 github.com/containers/storage v1.15.4 h1:eiUtV9MOTnPHibO18nDRI+aDhKudY7WmAiJdyVMsqSM=
 github.com/containers/storage v1.15.4/go.mod h1:v0lq/3f+cXH3Y/HiDaFYRR0zilwDve7I4W7U5xQxvF8=
+github.com/containers/storage v1.15.7 h1:ecPmv2y/qpxeSTHZ147jQLO6to8wDn8yUPtDCZlz0H4=
+github.com/containers/storage v1.15.7/go.mod h1:gLZIp+/hP8nFn9tLS0uJlnk4h1tSoDu3oS2eFiaIqkE=
 github.com/containers/storage v1.15.8 h1:ef7OfUMTpyq0PIVAhV7qfufEI92gAldk25nItrip+6Q=
 github.com/containers/storage v1.15.8/go.mod h1:zhvjIIl/fR6wt/lgqQAC+xanHQ+8gUQ0GBVeXYN81qI=
 github.com/coredns/corefile-migration v1.0.4/go.mod h1:OFwBp/Wc9dJt5cAZzHWMNhK1r5L0p0jDwIBc6j8NC8E=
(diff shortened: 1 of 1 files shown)


vendor/github.com/opencontainers/selinux/go-selinux/selinux_linux.go
```diff
@@ -370,7 +370,10 @@ func writeCon(fpath string, val string) error {
 	} else {
 		_, err = out.Write(nil)
 	}
-	return err
+	if err != nil {
+		return errors.Wrapf(err, "failed to set %s on procfs", fpath)
+	}
+	return nil
```
saschagrunert: This change in selinux-go seems to cause a regression on our side (introduced in https://github.com/opencontainers/selinux/pull/60)

TomSweeneyRedHat:
LGTM assuming happy tests

vendor/github.com/opencontainers/selinux/go-selinux/selinux_linux.go
```diff
@@ -289,7 +289,10 @@ func SetFileLabel(fpath string, label string) error {
 	if fpath == "" {
 		return ErrEmptyPath
 	}
-	return lsetxattr(fpath, xattrNameSelinux, []byte(label), 0)
+	if err := lsetxattr(fpath, xattrNameSelinux, []byte(label), 0); err != nil {
+		return errors.Wrapf(err, "failed to set file label on %s", fpath)
+	}
+	return nil
```
saschagrunert: This change in selinux-go seems to cause a regression on our side (introduced in https://github.com/opencontainers/selinux/pull/60)

saschagrunert: @rhatdan do you have any idea how to fix the SELinux related issue?
```
failed to set file label on /var/lib/containers/storage/overlay/ed4a2284370c886d47ef5a6b0a183a7007b67a354a1f1c29fe478eb8e4399a0a/merged: operation not supported
```

vrothberg: I am also hitting it -> https://github.com/cri-o/cri-o/pull/3205

rhatdan: Looks like we should ignore ENOSUP errors.

rhatdan: Was CRI-O examining the error, and checking for ENOSUP?  It needs to use errors.Cause now.

rhatdan: Where is this being called?

nalind: https://github.com/cri-o/cri-o/blob/master/server/label_linux.go#L11, I think.

saschagrunert: I was thinking the same but was not sure if it is an appropriate solution to skip those errors. What is the reason behind it?

I can spin up a follow-up PR tomorrow if nobody catches up first. :)

nalind: In a number of places, we're already comparing the error we get back to syscall.ENOTSUP.  The change would be to pass those errors to `errors.Cause()` to unwrap them before comparing them.

rhatdan: I believe this is the correct fix.

nalind: Opened #3212.  If you prefer to cherry-pick it into this one and close that one, I'm fine with that.

PR #3209, opened 2020-02-05
Fix reload behavior for unqualified search registries

saschagrunert:
If we reload via SIGHUP, then the systemContext may get populated with new
unqualified search registries. We have to check those in every pull
operation, because otherwise the update of the unqualified search
registries would have no effect.

```
unqualified-search-registries = ["docker.io"]
```


haircommander:
LGTM

TomSweeneyRedHat:
Just wish there was a way to not need a file lookup with every pull.

saschagrunert:
We don’t lookup the file directly since the registries are cached during runtime. The main issue was that the cached registries have not been considered yet. 🤭

PR #3158, opened 2020-01-24
Update urfave/cli to latest version

saschagrunert:
Hidden flags are now not visible in the documentation any more.

PR #3215, opened 2020-02-05
[1.17] drop host_ip

haircommander:
cherry pick of https://github.com/cri-o/cri-o/pull/3153 and https://github.com/cri-o/cri-o/pull/3214

drops host_ip as a config option. this should have no effect because we don't use it anymore as of https://github.com/cri-o/cri-o/pull/3171


haircommander:
Since `--host-ip` was an option in 1.16 (aka I was wrong here https://github.com/cri-o/cri-o/pull/3153#issuecomment-580273799), I also added it as a hidden flag, and marked it deprecated. In 1.18, we can drop support for it (which is conveniently already in master :smile: ) to do so, I had to cherry-pick https://github.com/cri-o/cri-o/pull/3158 to not have --host-ip littered throughout all of the docs (it remains in some :woman_shrugging: )

PR #3217, opened 2020-02-05
[1.16] Vendor latest opencontainers/runtime-tools

umohnani8:
Pulls in the fix that adds the missing clone
masked_eq seccomp rule which causes clone syscalls
to fail on s390x - opencontainers/runtime-tools#700

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/opencontainers/image-spec v1.0.2-0.20190823105129-775207bd45b6
 	github.com/opencontainers/runc v1.0.0-rc9
 	github.com/opencontainers/runtime-spec v1.0.1
-	github.com/opencontainers/runtime-tools v0.9.0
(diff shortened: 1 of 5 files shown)


umohnani8:
@vrothberg I am not sure why `make vendor` deleted a bunch of vendored files, but cri-o builds successfully. Wondering if I missed something...

vrothberg:
> ~@vrothberg I am not sure why `make vendor` deleted a bunch of vendored files, but cri-o builds successfully. Wondering if I missed something...~
> 
> Nvm, I think I fixed it

Awesome! I just saw the message in the morning now.

vrothberg:
Apologies. I merged without realizing that another LGTM was missing.

giuseppe:
> Apologies. I merged without realizing that another LGTM was missing.

I add it now, LGTM :-)

PR #3045, opened 2019-12-12
persist exit status

haircommander:
conmon will save the exit file of the container under
/var/lib so it gets persisted and available even on reboot.

Fixes to better handle exit code:

    Change ExitCode to be a pointer to distinguish from
    exit code 0.
    On reboot if we can't get runc state, then attempt
    to read exit code from the persisted exit file.
    This handles cases where crio was unable to persist
    the state for the container but conmon had written
    the exit file for the container.
    When restoring state during CRI-O startup,
    write back the state to disk so that any missed
    exit file updates are accounted for.

Signed-off-by: Mrunal Patel mrunalp@gmail.com
cherry-pick of https://github.com/cri-o/cri-o/pull/3041

PR #3218, opened 2020-02-05
[1.11] port forward: drain the stream on error

haircommander:

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1798193


haircommander:
(this is currently untested)

rhatdan:
LGTM
@haircommander Is this the one that Kata does not work for?

haircommander:
> LGTM
> @haircommander Is this the one that Kata does not work for?

no, you're thinking of https://github.com/cri-o/cri-o/pull/3045

kata isn't working here probably because 1.11 is wicked old. I'm also not super inclined to get it to work because 1.11 is out of support in the upstream, and this PR is mostly for openshift 3.11 support

rhatdan:
LGTM

PR #3223, opened 2020-02-06
[1.14] Bump up version to 1.14.12

umohnani8:
Bump up version to 1.14.12 for a new release.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -13,7 +13,7 @@ import (
 )
 
 // Version is the version of the build.
-const Version = "1.14.11"
+const Version = "1.14.12"
 
 // ShouldCrioWipe opens the version file, and parses it and the version string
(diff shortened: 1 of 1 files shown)


umohnani8:
@haircommander @vrothberg PTAL

version/version.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // Version is the version of the build.
-const Version = "1.14.11"
+const Version = "1.14.13-dev"
```
haircommander: I thought we stopped using -dev because of the automation scripts

umohnani8: Is that so? I saw we were still using dev in 1.15. Will remove it

haircommander: yeah I think we should drop dev everywhere (1.15 was probably an oversight because its our forgotten child)

PR #3227, opened 2020-02-07
[1.17] Fix server restore to not remove podman containers

haircommander:
removed.

by the new podman annotation.


rhatdan:
I dream of when we finally get libpod merged into CRI-O and we can do this in a non hacky fashion.
LGTM

vrutkovs:
@haircommander don't forget to cherry-pick to master so that we won't stumble on this again in 1.18

haircommander:
Thanks for the reminder @vrutkovs, taken care of here https://github.com/cri-o/cri-o/pull/3228

PR #3229, opened 2020-02-07
[1.14] add conmonmon

haircommander:
One of the many shortcomings of kernel space OOM killing is  there isn't very much a process can do to prevent itself from being OOM  killed. Even adjusting its score doesn't prevent the kernel from  killing it.

Unforuntately for CRI-O, conmon being OOM killed is very bad. A  container can be orphaned and left running, with no way to report it  state to CRI-O. At worst, it could keep PIDs in CRI-O that really should  have gone away.

Introducing conmonmon, a conmon monitor. conmonmon loops in a routine  and intermittantly calls kill 0 on conmon, seeing if its still alive.  If it has for some reason been killed, then conmonmon kills its  associated container, and sets the container's state as though it OOM'd.  This way, we can report to the kubelet that the container OOM'd, even  though there is no one listening to


rhatdan:
@haircommander Needs a rebase.

rhatdan:
@haircommander Needs a rebase.

haircommander:
Considering we're in the process of deprecating conmonmon, we don't need this anymore

PR #3235, opened 2020-02-10
contrib: use crun from the containers repository

giuseppe:
and use a fixed version


rhatdan:
Seems like a no brainer, merging.
/lgtm

PR #3237, opened 2020-02-10
Add pinns binary to static bundle

saschagrunert:
Fixes the missing binary in the bundle.


TomSweeneyRedHat:
Changes LGTM, but  the tests are already complaining

PR #3241, opened 2020-02-10
Make SIGHUP reload for drop-in config dir work

saschagrunert:
We now store the config dir and path locations directly at the `Config`
structure, which allows us to use that information on SIGHUP reload to
check both locations for possible new configuration files.


rhatdan:
LGTM

mrunalp:
If we copy a bunch of files at the same time into drop-in directory is the ordering handled correctly?
The files would most likely have different priorities.

mrunalp:
nvm, since we are relying on SIGHUP, the ordering should be fine.

PR #3246, opened 2020-02-11
Add live reload to DecryptionKeysPath

saschagrunert:
This adds the live configuration reload feature to the
`decryption_keys_path`. We still check per-pull if the path exists and
decide if the key should be loaded or not.


TomSweeneyRedHat:
LGTM assuming happy test

PR #3252, opened 2020-02-11
[1.14] exec: Close pipe fds to prevent hangs

haircommander:


oci/runtime_oci.go
```diff
@@ -305,6 +309,13 @@ func (r *runtimeOCI) ExecContainer(c *Container, cmd []string, stdin io.Reader,
 		}
 
 		cmdErr = execCmd.Run()
+
+		// The read side of the pipe should be closed after the container process has been started.
```
mrunalp: This looks incorrect. Did we miss the start/wait split here?

haircommander: ah, oops, we sure did

haircommander: fixed!

PR #3257, opened 2020-02-11
[1.13] Pass down the integer value of the stop signal

umohnani8:
Instead of converting the stop signal from int to its
string name, i.e sig 15 to "TERM", send down the int value
converted to a string instead i.e sig 15 to "15". This avoids
doing a look up for the signal in the sigMap in both cri-o
and runc and double back and forth conversion from int to string.
The sigMap in the runc code for example didn't have all the
signals that the sigMap in cri-o does. So when a container
with the ubi8-init image was run and stopped, a SIGRTMIN+3
is used, which wasn't being recognized by runc causing the
container to fail to stop.

### DIFF
--- a/lib/kill.go
+++ b/lib/kill.go
@@ -1,22 +1,24 @@
 package lib
 
 import (
+	"os"
+	"strconv"
+	"syscall"
+
 	"github.com/cri-o/cri-o/oci"
 	"github.com/cri-o/cri-o/utils"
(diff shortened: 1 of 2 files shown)


haircommander:
LGTM

umohnani8:
@rhatdan @mrunalp can you please add the approve label

PR #3261, opened 2020-02-12
Add high level Sandbox and Container interface

saschagrunert:
#### What type of PR is this?
/kind design

Target of this PR is to find a common interface for sandbox creation. The `Creator` interface contains methods which should be called in order to successfully create a sandbox.

Please note that the return and parameters of the functions are still empty, those would have to be defined after we can agree on the general approach.

None

None

```release-note
None
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -1,0 +1,5 @@
+package container
+
+type Container interface {
+	/* TODO: add methods */
+}
--- a/pkg/sandbox/add_container.go
+++ b/pkg/sandbox/add_container.go
@@ -1,0 +1,9 @@
+package sandbox
+
+import (
+	"github.com/cri-o/cri-o/pkg/container"
+)
+
+func (s *sandbox) AddContainer(container.Container) error {
+	return nil
+}
--- a/pkg/sandbox/create.go
+++ b/pkg/sandbox/create.go
@@ -1,0 +1,5 @@
+package sandbox
+
+func (s *sandbox) Create() error {
+	return nil
+}
--- a/pkg/sandbox/delete.go
+++ b/pkg/sandbox/delete.go
@@ -1,0 +1,5 @@
+package sandbox
+
+func (s *sandbox) Delete() error {
+	return nil
+}
--- a/pkg/sandbox/remove_container.go
+++ b/pkg/sandbox/remove_container.go
@@ -1,0 +1,9 @@
+package sandbox
+
+import (
+	"github.com/cri-o/cri-o/pkg/container"
+)
+
+func (s *sandbox) RemoveContainer(container.Container) error {
+	return nil
+}
--- a/pkg/sandbox/sandbox.go
+++ b/pkg/sandbox/sandbox.go
@@ -1,0 +1,32 @@
+package sandbox
+
+import (
+	"context"
+
+	"github.com/cri-o/cri-o/pkg/container"
+)
+
+// Sandbox is the interface for managing pod sandboxes
+type Sandbox interface {
+	Create() error
+
+	Start() error
+
+	Stop() error
+
+	Delete() error
+
+	AddContainer(container.Container) error
+
+	RemoveContainer(container.Container) error
+}
+
+// sandbox is the hidden default type behind the Sandbox interface
+type sandbox struct {
+	ctx context.Context
+}
+
+// New creates a new, empty Sandbox instance
+func New(ctx context.Context) Sandbox {
+	return &sandbox{ctx}
+}
--- a/pkg/sandbox/start.go
+++ b/pkg/sandbox/start.go
@@ -1,0 +1,5 @@
+package sandbox
+
+func (s *sandbox) Start() error {
+	return nil
+}
--- a/pkg/sandbox/stop.go
+++ b/pkg/sandbox/stop.go
@@ -1,0 +1,5 @@
+package sandbox
+
+func (s *sandbox) Stop() error {
+	return nil
+}


server/sandbox/create.go
```diff
@@ -0,0 +1,101 @@
...
+	// It has to be ensured that the sandbox name does not already exist.
+	GenerateName()
+
+	// PullPauseImage ensures that the pause image of the container exists
+	// before container creation.
+	PullPauseImage()
+
+	// CreateContainer builds the container from the provided names,
+	// identifiers and pause image.
+	CreateContainer()
```
haircommander: CreateInfraContainer() perhaps?

server/sandbox/create.go
```diff
@@ -0,0 +1,101 @@
...
+// pod sandbox.
+// TODO: add the necessary arguments to the methods
+type Creator interface {
+	// GenerateName prepares the name and unique identifier of the pod sandbox.
+	// It has to be ensured that the sandbox name does not already exist.
+	GenerateName()
+
+	// PullPauseImage ensures that the pause image of the container exists
+	// before container creation.
+	PullPauseImage()
```
haircommander: Let's keep the language to Infra
PullInfraImage

server/sandbox/create.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// CreateContainer builds the container from the provided names,
+	// identifiers and pause image.
+	CreateContainer()
+
+	// SetupOCISpec prepares the basic, host specific OCI configuration,
+	// like setting the process args for the pause image execution or DNS
+	// configuration.
+	SetupOCISpec()
```
haircommander: Hm, I am unsure this should be separate from CreateInfraContainer

server/sandbox/create.go
```diff
@@ -0,0 +1,101 @@
...
+	AddMetadata()
+
+	// AddLabels adds the provided labels to the sandbox.
+	AddLabels()
+
+	// AddAnnotation adds the provided annotations to the sandbox.
+	AddAnnotations()
+
+	// AddCapabilities adds the provided capabilities to the sandbox.
+	AddCapabilities()
```
haircommander: I feel compelled to keep the verbiage the same
`Setup...`
instead of having some be Add and some be Setup

server/sandbox/create.go
```diff
@@ -0,0 +1,101 @@
...
+	SetupPortMappings()
+
+	// SetupSysctls prepares the sysctls for the sandbox.
+	SetupSysctls()
+
+	// SetupCgroups ensures the correct parent cgroup for the sandbox.
+	SetupCgroups()
+
+	// SetupNetwork creates the network for the sandbox.
+	SetupNetwork()
```
haircommander: We should define well the relationship between setup network and setup namespace options, as network is one of the namespace options

server/sandbox_cri.go
```diff
@@ -0,0 +1,38 @@
+package server
```
haircommander: what is the point of this file? All sandbox related cri operations?

server/sandbox/create.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"context"
+)
+
+// Creator is the interface which reflects necessary steps in sandbox
+// creation. All methods have to be called in order to successfully creator a
+// pod sandbox.
+// TODO: add the necessary arguments to the methods
+type Creator interface {
```
mrunalp: This interface is something that could be used to implement infra container based sandboxes i.e. our default code path or the libpod integration. 

This doesn't work for pinned namespace based containers or vm based containers, though. So, for those cases we may not even need a creator like interface to begin with.

So, we may need to specialize the name more than just Creatorr.

server/sandbox/create.go
saschagrunert: Yes, I change it :+1:

server/sandbox_cri.go
saschagrunert: Hm, no I think we could put them into dedicated files like we currently have it.

server/sandbox/create.go
saschagrunert: So for VM based containers we cannot be more specific than the CRI (which already is an interface)? :thinking:

If so, them I'm wondering if another interface is the right approach or if we need something else to reduce the complexity… :grin:

haircommander: We may be able to get away with a single SandboxFactory function that takes the parameters and branches on the type to create, calling the constructor of whichever implementation of the Sandbox interface is needed, Instead of an entire Creator interface

server/sandbox/create.go
saschagrunert: This is obsolete with the latest version of this PR.

server/sandbox/create.go
saschagrunert: This is obsolete with the latest version of this PR.

server/sandbox/create.go
saschagrunert: This is obsolete with the latest version of this PR.

server/sandbox/create.go
saschagrunert: Okay, I tried to strip down the interface to a bare minimum and we could focus on two main high-level interfaces for now `Container` and `Sandbox`. WDYT? I think this way we could refactor into a plain new `pkg/{container,sandbox}` package.

saschagrunert:
If we decide to continue with that than this could be our starting base.

saschagrunert:

TomSweeneyRedHat:
LGTM, but this is one for @mrunalp and @haircommander to approve.
Happy green test buttons though!

rhatdan:
@haircommander @mrunalp @kolyshkin @umohnani8 PTAL

haircommander:
LGTM, we probably will tweak it more in the future, but this is a good start

PR #3269, opened 2020-02-12
Change CircleCI config to build all jobs for all tags

saschagrunert:
This seems the only way to force CircleCI to trigger jobs on tags.


saschagrunert:

PR #3270, opened 2020-02-12
conmonmon: errorf when OOM killing

haircommander:
when conmonmon decides a conmon is gone, and its container should have oom kill spoofed, that's a big deal.
As an admin, we'd typically want that feedback.
Error is a standard value for log level, so this message is more likely to show up. Error is more representative of the severity of the situation


rhatdan:
LGTM

PR #3274, opened 2020-02-12
[1.14] port forward: drain the stream on error

haircommander:


haircommander:
upon acceptance of this PR I'll forward port to 1.1{5..7} and master

rhatdan:
LGTM

PR #3279, opened 2020-02-13
[1.14] Restore sandbox selinux labels directly from config.json

mrunalp:


rhatdan:
These SELinux labels we previously generated, I believe so LGTM

PR #3281, opened 2020-02-13
Update pinns build and add small cleanups

saschagrunert:
Just a couple of enhancements to pinns:

- Strip the binary to reduce its size
- Make `bind_ns()` static to scope to single object
- Print a `:` separator on logging errno

### DIFF
--- a/pinns/Makefile
+++ b/pinns/Makefile
@@ -2,11 +2,14 @@ src = $(wildcard *.c)
 obj = $(src:.c=.o)
 
 override LIBS +=
-CFLAGS ?= -std=c99 -Os -Wall -Wextra
+CFLAGS ?= -std=c99 -Os -Wall -Werror -Wextra
 override CFLAGS += -static
+
+all: ../bin/pinns
 
 ../bin/pinns: $(obj) | ../bin
 	$(CC) -o $@ $^ $(CFLAGS) $(LIBS)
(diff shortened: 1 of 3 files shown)


rhatdan:
Boy would I love to get PINNS into Libpod/Podman hint, hint...

saschagrunert:
> Boy would I love to get PINNS into Libpod/Podman hint, hint...

Do we want to move it into a dedicated project first? I also thought about a fancy rewrite in Rust. :smiling_imp: WDYT?

rhatdan:
Yes I would prefer this be moved to libpod or to an independent package. Do not want libpod importing from CRI-O.  Don't care what it is written in.  Although I am getting too old to learn a new language :^)

Basically would love to drop pause container from podman pod...

saschagrunert:
> Yes I would prefer this be moved to libpod or to an independent package. Do not want libpod importing from CRI-O. Don't care what it is written in. Although I am getting too old to learn a new language :^)
> 
> Basically would love to drop pause container from podman pod...

How is that: https://github.com/saschagrunert/pinns.rs ? If you don't like it I'm also open to stick to the C version

cc @haircommander @mrunalp

saschagrunert:
Following up in https://github.com/containers/libpod/pull/5210

PR #3287, opened 2020-02-13
Remove version marker from AppArmor profile

saschagrunert:
Upgrade paths destroyed the CRI-O configuration multiple times because
of changing AppArmor profile names. We should not do that any more and
default to a simple profile name which does not change per release.


saschagrunert:
> > Upgrade paths destroyed the CRI-O configuration multiple times
> 
> What exactly broke? 

If customers edit the crio.conf and do an upgrade, then (depending on the package setting) that file might get not updated on disk. On a CRI-O restart, it now assumes that this is not the default profile any more and assumes a user-provided profile because it contains the previous version. If this profile does not exist (because of an restart for example), then CRI-O silently got broken because of the configuration.

saschagrunert:
> Are we force-loading the profile on each start up of CRI-O? We need to do that now to make sure to always load the latest profile.

Yes:
https://github.com/cri-o/cri-o/blob/ba4a8cdc9470a099563b78686a12ab92eb5f0fd1/server/server.go#L369-L372
https://github.com/containers/libpod/blob/e4e5efc12b7902a22561170ca59f3da90b7b80ce/pkg/apparmor/apparmor_linux.go#L80-L114

TomSweeneyRedHat:
LGTM once the tests are happy.

PR #3291, opened 2020-02-14
port forward: drain the stream on error

haircommander:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/creack/pty v1.1.9
 	github.com/cri-o/ocicni v0.2.1-0.20200409131010-b197cd13855b
 	github.com/cyphar/filepath-securejoin v0.2.2
+	github.com/docker/docker v1.4.2-0.20191219165747-a9416c67da9f // indirect
 	github.com/docker/go-units v0.4.0
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/go-zoo/bone v1.3.0
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -904,8 +904,9 @@ func (r *runtimeOCI) PortForwardContainer(c *Container, port int32, stream io.Re
 	defer func() {
 		if emptyStreamOnError {
 			go func() {
-				_, copyError := pools.Copy(ioutil.Discard, stream)
-				logrus.Errorf("error closing port forward stream after other error: %v", copyError)
+				if _, copyError := pools.Copy(ioutil.Discard, stream); copyError != nil {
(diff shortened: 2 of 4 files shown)


umohnani8:
LGTM

giuseppe:
looks like the error in CircleCI is not a flake

rhatdan:
@haircommander Could you check out this error?

server/container_portforward.go
```diff
@@ -20,7 +23,21 @@ func (s *Server) PortForward(ctx context.Context, req *pb.PortForwardRequest) (r
...
+		if emptyStreamOnError {
+			go func() {
+				_, copyError := pools.Copy(ioutil.Discard, stream)
+				logrus.Errorf("error closing port forward stream after other error: %v", copyError)
+			}()
+		}
+	}()
+
 	sandboxID, err := s.runtimeServer.PodIDIndex().Get(podSandboxID)
+
```
umohnani8: nit: extra line

haircommander:
PTAL @saschagrunert @rhatdan @umohnani8 @mrunalp

internal/oci/runtime_oci.go
```diff
@@ -871,6 +871,15 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
 
 // PortForwardContainer forwards the specified port provides statistics of a container.
 func (r *runtimeOCI) PortForwardContainer(c *Container, port int32, stream io.ReadWriter) error {
+	emptyStreamOnError := true
+	defer func() {
+		if emptyStreamOnError {
+			go func() {
+				_, copyError := pools.Copy(ioutil.Discard, stream)
+				logrus.Errorf("error closing port forward stream after other error: %v", copyError)
```
saschagrunert: Do we have to check if `copyEror != nil`?

haircommander: aw jeez we do

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/cpuguy83/go-md2man v1.0.10
 	github.com/cri-o/ocicni v0.1.1-0.20200218164758-d2881573038f
 	github.com/cyphar/filepath-securejoin v0.2.2
+	github.com/docker/docker v1.4.2-0.20191219165747-a9416c67da9f
```
saschagrunert: Do we have any alternative to this dependency?

haircommander: good catch, fixing

haircommander:
> This PR was merged in the time period when ci-operator was mistakenly reporting failed tests as passing.
> 
> If this repository has ci-operator jobs, please inspect their test results even if passing, and consider the need for fixing or even reverting.

let's see how https://github.com/cri-o/cri-o/pull/3716 does, tests shouldn't differ too much

PR #3297, opened 2020-02-18
Add cgroup namespace unsharing to pinns

saschagrunert:
This adds the ablity to unshare a cgroup namespace via pinns. We also
update the .clang-format file to enable proper auto-formatting.

### DIFF
--- a/pinns/pinns.c
+++ b/pinns/pinns.c
@@ -45,6 +45,7 @@ int main(int argc, char **argv) {
   bool bind_uts = false;
   bool bind_ipc = false;
   bool bind_user = false;
+  bool bind_cgroup = false;
 
   static const struct option long_options[] = {
       {"help", no_argument, NULL, 'h'},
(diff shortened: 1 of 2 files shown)


pinns/pinns.c
```diff
@@ -77,6 +81,19 @@ int main(int argc, char **argv) {
       bind_user = true;
       num_unshares++;
       break;
+    case 'p':
+      unshare_flags |= CLONE_NEWPID;
```
giuseppe: `unshare()` won't be enough to create a new PID namespace.

a PID namespace is created after a `fork()`

saschagrunert: Hah, thanks for the hint. I assume we just have to pin the non PID 1 namespace after the fork, right? (see latest implementation)

pinns/pinns.c
```diff
@@ -126,11 +143,31 @@ int main(int argc, char **argv) {
...
 
+  if (bind_pid) {
+    if (fork() < 0) {
+      pexit("unable to fork process");
+    }
+    // The first child created by a process after a call to unshare(2) using
+    // the CLONE_NEWPID flag) has the PID 1, and is the init process for the
+    // namespace. We just have to pin the process which is not PID 1 after the
+    // call to fork(2).
+    if (getpid() != 1 && bind_ns(pin_path, "pid") < 0) {
```
giuseppe: I think it should be the child to do the pinning.

You can use directly the return code from fork().

```
pid_t p = fork();
....

if (p == 0 && bind_ns(pin_path, "pid") < 0) {
```

mrunalp:
We don't need pid unsharing for pinns because in that case we need a pid 1 container running for the pod. So, that code path will work as it works today.

saschagrunert:
> We don't need pid unsharing for pinns because in that case we need a pid 1 container running for the pod. So, that code path will work as it works today.

Ah okay, alright then I change that PR to only add cgroup unsharing to pinns.

pinns/pinns.c
saschagrunert: Cool thank you for the hint. Unfortunately the pid namespace unsharing does not seem to be a valid use case any more.

saschagrunert:

PR #3306, opened 2020-02-20
Fix Fedora based tests

saschagrunert:
Trying to find out why the CI is broken for Fedora.

### DIFF
--- a/contrib/test/integration/main.yml
+++ b/contrib/test/integration/main.yml
@@ -36,7 +36,7 @@
         k8s_git_version: "cgroupv2-crio"
         k8s_github_fork: "giuseppe"
         crio_socket: "/var/run/crio/crio.sock"
-      when: "cgroupv2 | default(False) | bool"
+      when: "cgroupv2 | bool"
 
     - name: clone build and install kubetest
       include: "build/kubetest.yml"
@@ -45,11 +45,11 @@
 
     - name: clone build and install runc
       include: "build/runc.yml"
-      when: "build_runc | default(True) | bool"
+      when: "build_runc | bool"
 
     - name: clone build and install crun
       include: "build/crun.yml"
-      when: "build_crun | default(False) | bool"
+      when: "build_crun | bool"
 
     - name: clone build and install networking plugins
       include: "build/plugins.yml"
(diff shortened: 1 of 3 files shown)


test/helpers.bash
```diff
@@ -38,7 +38,7 @@ ULIMITS=${ULIMITS:-}
 DEVICES=${DEVICES:-}
 OVERRIDE_OPTIONS=${OVERRIDE_OPTIONS:-}
 RUNTIME_PATH=$(command -v $CONTAINER_RUNTIME || true)
-RUNTIME_BINARY=${RUNTIME_PATH:-/usr/local/sbin/runc}
+RUNTIME_BINARY=${RUNTIME_PATH:-$(which runc)}
```
TomSweeneyRedHat: interesting twist, I like!

saschagrunert:
I think the issue is:
```
which: no runc in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
```

contrib/test/integration/main.yml
```diff
@@ -89,6 +89,11 @@
     - name: clone build and install crun
       include: "build/crun.yml"
       when: "build_crun | default(False) | bool"
+
+    - name: clone build and install runc
+      include: "build/runc.yml"
+      when: "build_runc | default(True) | bool"
```
saschagrunert: @giuseppe Is it correct that `default(True)` only applies if `build_runc` is not set? That would mean in our case that it is always set becuse of vars.yml: `build_runc: False`.

saschagrunert:
Look how green :green_heart: PTAL @haircommander @giuseppe @mrunalp @rhatdan

contrib/test/integration/main.yml
giuseppe: I guess so :)  we can probably drop that

giuseppe:
/test e2e_fedora

LGTM

contrib/test/integration/main.yml
saschagrunert: Ah okay I'll do.

saschagrunert:
One more round, I'm optimistic that this works :)

PR #3309, opened 2020-02-21
Take localhost/ images into account during pull

saschagrunert:
#### What type of PR is this?
/kind feature

When pulling images, then we were not able to handle locally build
images before this patch.

Now, the last fallback to `localhost/` prefixed images will be the node,
because it could be theoretically possible that a registry runs on
`localhost`.

Refers to https://github.com/kubernetes-sigs/cri-tools/issues/474

Drawback of this approach is that the new behavior ignores the
`imagePullPolicy` of Kubernetes in conjunction with `localhost/`
prefixed images. This might be not problematic since prefixing local
images with `localhost/` is a general paradigm from buildah/podman.

```release-note
- Take local images into account when pulling images prefixed with `localhost/`
```

### DIFF
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -8,12 +8,15 @@ import (
 
 	"github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/types"
+	libpodImage "github.com/containers/libpod/libpod/image"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/storage"
 	"github.com/cri-o/cri-o/server/metrics"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
+
+var localRegistryPrefix = libpodImage.DefaultLocalRegistry + "/"
 
 // PullImage pulls a image with authentication config.
 func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp *pb.PullImageResponse, err error) {
@@ -61,6 +64,18 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp 
 		var tmpImg types.ImageCloser
 		tmpImg, err = s.StorageImageServer().PrepareImage(&sourceCtx, img)
 		if err != nil {
+			// We're not able to find the image remotely, check if it's
+			// available locally, but only for localhost/ prefixed ones.
+			// This allows pulling localhost/ prefixed images even if the
+			// `imagePullPolicy` is set to `Always`.
+			if strings.HasPrefix(img, localRegistryPrefix) {
+				if _, err := s.StorageImageServer().ImageStatus(
+					s.config.SystemContext, img,
+				); err == nil {
+					pulled = img
+					break
+				}
+			}
 			log.Debugf(ctx, "error preparing image %s: %v", img, err)
 			continue
 		}


TomSweeneyRedHat:
FWIW, we've had some pushback on adding `localhost/` to local image names in Buildah/Podman.

rhatdan:
LGTM

mrunalp:
This sounds good to me. @mtrmac @vrothberg ptal.

umohnani8:
LGTM

haircommander:
LGTM, waiting on final :+1: from @vrothberg or @mtrmac

rhatdan:
@saschagrunert Needs a rebase

saschagrunert:
> My initial impression of this (admittedly only after looking at this for a few minutes.): I can’t see why `PullImage` is the **EDIT only** place that needs updating for this semantics change.
> 
> E.g. the kubelet is calling ImageStatus with an image name before deciding to call PullImage. Another way to look at this, there are 4 different callers of `ResolveNames`. Don’t the other callers need updating too? (And that may not even be everything, basically everything that deals with image names could be affected in principle… which would be easier to audit if we didn’t use the same `string` type for pull specs and image IDs ~interchangeably in various locations.)

I gave it a local test and the added use case works as intended. The call to `ImageStatus` works as well:
```
time="2020-03-09 12:14:37.254398314+01:00" level=debug msg="Request: &ImageStatusRequest{Image:&ImageSpec{Image:localhost/test:latest,},Verbose:false,}" file="go-grpc-middleware/chain.go:25" id=5500fc92-692b-416f-95b6-594582a490af
time="2020-03-09 12:14:37.255260020+01:00" level=debug msg="Response: &ImageStatusResponse{Image:&Image{Id:c0f3954bbe8fab062f2632154fd41e186bdd8971013240946af3277b8f12eec5,RepoTags:[localhost/test:latest],RepoDigests:[localhost/test@sha256:1ed39c151e61317e2784e93a5b495afa92ff6df24d85bc0e7805b7b660cc1ac6],Size_:5858702,Uid:nil,Username:,},Info:map[string]string{},}" file="go-grpc-middleware/chain.go:25" id=5500fc92-692b-416f-95b6-594582a490af
```

mtrmac:
> I gave it a local test and the added use case works as intended. The call to `ImageStatus` works as well:

OK, then I guess I am just very confused.  Let me try to describe how I now understand the issue:
- https://github.com/kubernetes-sigs/cri-tools/issues/474 linked in the description is pretty much completely unrelated (apart from users trying to do something similar); those problems are caused by either using a partial image name (without `localhost/`) or by pulling images into a storage that is not used by CRI-O.
- `ImageStatus` does not need updating, because it can already automatically see Podman-pulled images
- Kubelet therefore can already see that images are available locally.
- Nevertheless, the Kubelet decides to ask for the image to be pulled anyway?! and that’s where the code introduced by this PR is invoked, and skips the pull.

So, is it fair to say that this PR essentially defeats `imagePullPolicy: Always`, and would not be necessary if the containers used a different `imagePullPolicy`? (`Always` is the default policy for the default `:latest` tag, so it still might be warranted to do this in CRI-O — at this point I’m just trying to clarify what this PR does.)

server/image_pull.go
```diff
@@ -61,6 +64,14 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
 		var tmpImg types.ImageCloser
 		tmpImg, err = s.StorageImageServer().PrepareImage(&sourceCtx, img)
 		if err != nil {
+			// We're not able to find the image remotely, check if it's
+			// available locally, but only for localhost/ prefixed ones
+			if strings.HasPrefix(img, localRegistryPrefix) {
+				if _, err := s.StorageImageServer().GetStore().Image(img); err == nil {
```
mtrmac: Could this use `s.StorageImageServer().ImageStatus()` instead? It’s a bit more expensive, but it has clearly the same semantics, unlike reaching through to the storage directly (I’m primarily worried about the extra lookups `storageReference.resolveImage` does, although the layering violation is not great either).

`ImageStatus()` is more expensive than `GetStore().Image()`, OTOH it is called after exiting the loop anyway, so if efficiency were a concern, the two can be deduplicated (both here and in the “already in store, skipping pull” path below), making `ImageStatus` ultimately cheaper.

saschagrunert:
> So, is it fair to say that this PR essentially defeats `imagePullPolicy: Always`, and would not be necessary if the containers used a different `imagePullPolicy`? (`Always` is the default policy for the default `:latest` tag, so it still might be warranted to do this in CRI-O — at this point I’m just trying to clarify what this PR does.)

Yes exactly. If we say that we want to point to the pull policy as workaround then we do not need this PR here. :)

mtrmac:
> > So, is it fair to say that this PR essentially defeats `imagePullPolicy: Always`, and would not be necessary if the containers used a different `imagePullPolicy`?
> 
> Yes exactly. If we say that we want to point to the pull policy as workaround then we do not need this PR here. :)

OK, thanks. I don’t have a strong opinion either way, and others like this, so, sure.

Mechanically I’d like to avoid the direct `GetImage` call, and thinking about maintainability, documenting this special case (and that it only matters for `imagePolicy:Always`) at least in the comment, if not elsewhere, would be nice.

mtrmac:
> I don’t have a strong opinion either way, and others like this, so, sure.

I mean “sure, let’s do what this PR proposes”, not “sure, let’s suggest changing the pull policy and reject this PR”, to be clear.

saschagrunert:
> > I don’t have a strong opinion either way, and others like this, so, sure.
> 
> I mean “sure, let’s do what this PR proposes”, not “sure, let’s suggest changing the pull policy and reject this PR”, to be clear.

Sounds good to me, I'll update my PR tomorrow regarding your review suggestions :)

server/image_pull.go
saschagrunert: Implemented as suggested and I also added a comment regarding the image pull policy.

saschagrunert:
Rebased on top of the latest master and implemented the review notes.

saschagrunert:
I'll manually cherry pick this in some time.

PR #3313, opened 2020-02-24
Add live configuration reload to AppArmor profile

saschagrunert:
The apparmor profile is now processed in a dedicated sub-package of the
`config` package and supports the live config reload feature. Necessary
test have been added as well.

### DIFF
--- a/internal/config/apparmor/apparmor.go
+++ b/internal/config/apparmor/apparmor.go
@@ -1,0 +1,134 @@
+package apparmor
+
+import (
+	"strings"
+
+	"github.com/containers/libpod/pkg/apparmor"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	k8sAppArmor "k8s.io/kubernetes/pkg/security/apparmor"
+)
+
+const (
+	// DefaultProfile is the default profile name
+	DefaultProfile = "crio-default"
+
+	unconfined = "unconfined"
+)
+
+// Config is the global AppArmor configuration type
+type Config struct {
+	enabled        bool
+	defaultProfile string
+}
+
+// New creates a new default AppArmor configuration instance
+func New() *Config {
+	return &Config{
+		enabled:        apparmor.IsEnabled(),
+		defaultProfile: DefaultProfile,
+	}
+}
+
+// LoadProfile can be used to load a AppArmor profile from the provided path.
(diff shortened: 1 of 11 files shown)


TomSweeneyRedHat:
LGTM

haircommander:
what would you think about having `pkg/config/apparmor` be in `internal/apparmor_config` or something like that? I imagine we'd want the fewest dependencies exported in `pkg/`. My hesitation is it could be a confusing file hierarchy. Non-blocking, mostly putting it out there for consideration

saschagrunert:
> what would you think about having `pkg/config/apparmor` be in `internal/apparmor_config` or something like that? I imagine we'd want the fewest dependencies exported in `pkg/`. My hesitation is it could be a confusing file hierarchy. Non-blocking, mostly putting it out there for consideration

Yes sounds good, I’ll move it over there (as well as the seccomp code). I think we could also add an `internal` package directly in `pkg/config`…

haircommander:
> > what would you think about having `pkg/config/apparmor` be in `internal/apparmor_config` or something like that? I imagine we'd want the fewest dependencies exported in `pkg/`. My hesitation is it could be a confusing file hierarchy. Non-blocking, mostly putting it out there for consideration
> 
> Yes sounds good, I’ll move it over there (as well as the seccomp code). I think we could also add an `internal` package directly in `pkg/config`…

considering that would probably look like `pkg/config/internal/apparmor` I'm hesitant. I think `internal/config/apparmor` sounds good, then seccomp can be `internal/config/seccomp`

saschagrunert:
> > > what would you think about having `pkg/config/apparmor` be in `internal/apparmor_config` or something like that? I imagine we'd want the fewest dependencies exported in `pkg/`. My hesitation is it could be a confusing file hierarchy. Non-blocking, mostly putting it out there for consideration
> > 
> > 
> > Yes sounds good, I’ll move it over there (as well as the seccomp code). I think we could also add an `internal` package directly in `pkg/config`…
> 
> considering that would probably look like `pkg/config/internal/apparmor` I'm hesitant. I think `internal/config/apparmor` sounds good, then seccomp can be `internal/config/seccomp`

Sounds good, moved the apparmor content into `internal/config/apparmor` and will spin-up another PR for seccomp.

saschagrunert:

PR #3314, opened 2020-02-24
Cleanup config default values

saschagrunert:
We do not set default struct values any more and focus only on the
actual differences in `DefaultConfig()` to reduce the noise. Unnecessary
const values have been merged into that function as well.


haircommander:
LGTM

PR #3316, opened 2020-02-24
Fix markdown for generated crio/crio-status docs

rhafer:
Many markdown tools (e.g. go-md2man, which cri-o uses to generate
man-pages) don't cope well with embedded HTML code. The HTML comment in
the generated markdown files caused the man-pages to be ill-formatted.
This commit replaces the comment with some valid markdown that is
invisible in the rendered HTML- and man-pages. It should also silence
the "WARNING: go-md2man does not handle node type HTMLBlock" message
during "make docs-generation".

Signed-off-by: Ralf Haferkamp <rhafer@suse.com>


haircommander:
nice thanks @rhafer , LGTM
/ok-to-test

PR #3319, opened 2020-02-25
Make bin/pinns a PHONY target

saschagrunert:
We have to make the `bin/pinns` target `PHONY` to have a rebuild based
on file system changes (via `pinns/Makefile`), otherwise we would omit
the "file changed check" inside the `pinns` directory.


rhatdan:
LGTM

TomSweeneyRedHat:

Issue #2979, opened 2019-11-18
make: -ldflags improvements 


mythi:
**Description**

1. `-X importpath.name=value` ldflags cannot be easily amended
2. `-X main.buildInfo=${SOURCE_DATE_EPOCH}` seems non-existent

I wanted to override `criocli.DefaultsPath` with make but it's not easy. A simple workaround would be to override `SHRINKFLAGS` with the default values plus my `-X ...DefaultsPath=` setting but not very nice.

While working with -X flags, I also noticed `main.buildInfo` is gone:

```
$ go tool nm bin/crio | grep -e buildInfo -e gitCommit
34adf70 d k8s.io/component-base/metrics.buildInfo
34363f0 d k8s.io/component-base/version.gitCommit
3436460 d main.gitCommit
2239460 r main.gitCommit.str
```

saschagrunert:
Hey @mythi, I'm working on a fix for this in #3320 and #3321

mythi:
@saschagrunert thanks!

PR #3320, opened 2020-02-25
Improve crio --version / version output

saschagrunert:
We now make the unnecessary public variables `version.GitCommit` and
`version.Version` private again.

The output of `crio -v` and the `VERSION` section of `crio -h` now show
the full version details, equally to `crio version`.

We can use the newly introduced `version.Get()` API to retrieve the
version information, for example in `useragent.go` and
`server/version.go`.

Refers to https://github.com/cri-o/cri-o/issues/2979


TomSweeneyRedHat:

PR #3321, opened 2020-02-25
Add DEFAULTS_PATH to Makefile

### LINKED ISSUE #2979: make: -ldflags improvements

saschagrunert:
#### What type of PR is this?
/kind feature

This allows user to easily inject the `criocli.DefaultsPath` variable
via `make DEFAULTS_PATH=<PATH>`. 

Fixes #2979

None

```release-note
- Provide the possibility to set the default config path via `make DEFAULTS_PATH=<PATH>`
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,8 +81,10 @@ GO_FILES := $(shell find . -type f -name '*.go' -not -name '*_test.go')
 VPATH := $(VPATH):$(GOPATH)
 SHRINKFLAGS := -s -w
 VERSION := $(shell $(GO_RUN) ./scripts/latest_version.go)
+DEFAULTS_PATH := ""
 
 BASE_LDFLAGS = ${SHRINKFLAGS} \
+	-X ${PROJECT}/internal/pkg/criocli.DefaultsPath=${DEFAULTS_PATH} \
 	-X ${PROJECT}/internal/version.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
 	-X ${PROJECT}/internal/version.gitCommit=${COMMIT_NO} \
 	-X ${PROJECT}/internal/version.gitTreeState=${GIT_TREE_STATE} \


rhatdan:
/test e2e_crun
LGTM

TomSweeneyRedHat:
LGTM

mythi:
LGTM

umohnani8:
/restest
LGTM

dougsland:
Just wondering, if we should document this option in cri-o github (devel) page.

saschagrunert:
> Just wondering, if we should document this option in cri-o github (devel) page.

Hm yeah I think this makes sense

mythi:
one note about the implementation still: this fulfills my original use-case (I wanted to set `DefaultsPath`) but it won't allow users to set *arbitrary* `-X ...values`.

saschagrunert:
> one note about the implementation still: this fulfills my original use-case (I wanted to set `DefaultsPath`) but it won't allow users to set _arbitrary_ `-X ...values`.

Which kind of other `-X` values would you like to inject?

mythi:
> Which kind of other -X values would you like to inject?

I don't have any other use case right now. Just wanted to make the point that if someone had one, Makefile patching would be needed AFAICS.

This change is LGTM as I wrote above.

saschagrunert:
> > Which kind of other -X values would you like to inject?
> 
> I don't have any other use case right now. Just wanted to make the point that if someone had one, Makefile patching would be needed AFAICS.
> 
> This change is LGTM as I wrote above.

Alright, thanks for the clarification :blush:

saschagrunert:

PR #3322, opened 2020-02-25
Rebuild bin/* targets on *.go file changes

saschagrunert:
This is not optimal, because it takes every *.go file into account. But
before this patch a binary, like `bin/crio` has not been re-build if any
dependent *.go file changed.


umohnani8:
LGTM

PR #3317, opened 2020-02-25
[WIP] CreateContainer: pass TerminationGracePeriod

kolyshkin:
Enable passing of sandbox's termination grace period
down to OCI runtime, as an annotation for systemd.

This is a glue between
* https://github.com/kubernetes/kubernetes/pull/88495 (note: pending merge)

and

* https://github.com/opencontainers/runc/pull/2224 (or https://github.com/containers/crun/pull/266)

that is a part of the fix for
* https://github.com/kubernetes/kubernetes/issues/77873

### TODO
* [ ] unit test

PR #3328, opened 2020-02-26
CreateContainer: pass TerminationGracePeriod

kolyshkin:
Enable passing of kubernetes termination grace period
down to OCI runtime, as an annotation for systemd.

This builds on top of

* https://github.com/opencontainers/runc/pull/2224   (or https://github.com/containers/crun/pull/266)

and is part of the fix for

* https://github.com/kubernetes/kubernetes/issues/77873

Replaces: #3317

### Testing notes

1. Install runc from latest git.
2. Compile cri-o with this patch, run it.
3. Run kubernetes so it uses the cri-o daemon compiled above.
4. Add the following into kubernetes pod yaml:
```yaml
apiVersion: v1
kind: Pod
....
spec:
  terminationGracePeriodSeconds: 111
....
```
5. Apply the above pod, wait for container(s) to start.
6. Find out any container id from the above pod (field `Container ID` from `kubectl describe pod ...` output, with `cri-o://` suffix removed).
7. Check its TimeoutStopUSec field by inspecting systemctl show output, for example
```console
$ CTID=8c0a8118d2e839d2fe9bc5cfcde83ab2ccfce06605b809ec81af4af5143936cb
$ systemctl show crio-$CTID.scope | grep TimeoutStop
TimeoutStopUSec=1min 51s
```

As you can see, 1 min 51s equals to what was specified in pod yaml (111s)

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -41,6 +41,9 @@ import (
 // A lower value would result in the container failing to start.
 const minMemoryLimit = 12582912
 
+// Copied from k8s.io/kubernetes/pkg/kubelet/kuberuntime/labels.go
+const podTerminationGracePeriodLabel = "io.kubernetes.pod.terminationGracePeriod"
+
 var (
 	_cgroupv2HasHugetlbOnce sync.Once
 	_cgroupv2HasHugetlb     bool
@@ -533,6 +536,15 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 			cgPath = filepath.Join(parent, scopePrefix+"-"+containerID)
 		}
 		specgen.SetLinuxCgroupsPath(cgPath)
+
+		if t, ok := kubeAnnotations[podTerminationGracePeriodLabel]; ok {
+			// currently only supported by systemd, see
+			// https://github.com/opencontainers/runc/pull/2224
+			if useSystemd {
+				specgen.AddAnnotation("org.systemd.property.TimeoutStopUSec",
+					"uint64 "+t+"000000") // sec to usec
+			}
+		}
 
 		if privileged {
 			specgen.SetupPrivileged(true)
--- a/contrib/test/integration/build/crun.yml
+++ b/contrib/test/integration/build/crun.yml
@@ -5,7 +5,7 @@
     repo: "https://github.com/containers/crun.git"
     dest: "{{ ansible_env.GOPATH }}/src/github.com/containers/crun"
     force: "{{ force_clone | default(False) | bool}}"
-    version: 0.12.2.1
+    version: "7c84e354169e3e4dd49d5ac21f13a143b87f5493" # 0.12.2.1 + https://github.com/containers/crun/pull/266
 
 - name: Install crun dependencies
   raw: >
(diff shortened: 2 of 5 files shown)


mrunalp:
:+1: We should add an integration test for this.

server/container_create_linux.go
```diff
@@ -534,6 +537,15 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		}
 		specgen.SetLinuxCgroupsPath(cgPath)
 
+		if t, ok := kubeAnnotations[podTerminationGracePeriodLabel]; ok {
+			// currently only supported by systemd, see
+			// https://github.com/opencontainers/runc/pull/2224
+			if useSystemd {
+				specgen.AddAnnotation("org.systemd.property.TimeoutStopUSec",
```
saschagrunert: Can we use something like `TimeoutStopSec` to avoid the `t+"000000"` below?

kolyshkin: @saschagrunert we could, but then @giuseppe would need to add support for `Sec` suffix in crun, and I understand why he'd rather not. So I'd rather have it as is for now, for the sake of crun compatibility.

saschagrunert: Okay sounds good. :+1:

kolyshkin:
TODO: unit test, thus WIP

server/container_create_linux.go
```diff
@@ -534,6 +537,15 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
...
+					"uint64 "+t+"000000") // sec to usec
```
mrunalp: I think we also want to protect ourselves from pods setting arbitrarily high timeouts delaying shut down so we could limit this to something like 5 minutes.

kolyshkin: Not sure it's a good idea to do that kind of validation in here. Such kind of validation belongs to either upper level, i.e. kubernetes, which validates input from user, or lower levels, i.e. systemd, which might find some values inapplicable, violating its settings, or otherwise bad.

In ths case, cri-o is just a messenger, and a messenger should not be overly smart.

mrunalp: I agree with you and I would prefer it to be at a higher level, too. It could be an admission hook or something like a Limit Range. However, either of those options may take time to deliver, so this suggestion is for short term where we can ship this (to help with issues we have at shutdown) without a bug opened for updates getting blocked by rogue pods.

kolyshkin: Makes sense (with something like `// TODO remove this once kubernetes PR #xxx is merged`).

In case of unreasonably high value (`val > MAX`), are you suggesting to
- return an error,
- skip the setting, or
- silently set it to `MAX`?

mrunalp: We can log a warning and set it to MAX if a value is set > MAX. Thanks!

kolyshkin:
Added a test case -- it should fail since I haven't bumped crun/runc yet.

kolyshkin:
OK, the test I wrote is passing locally but not in CI. I guess the reason is either there's no systemd, or old runc version is used.

kolyshkin:
We can't test it on CircleCI until the runc release with the feature is out, so I've added a skip

kolyshkin:
Skipped as expected on CircleCI:
> ok 122 kubernetes pod terminationGracePeriod passthru # skip runc v1.0.0-rc11 required

mrunalp:
@kolyshkin could you port this to release-1.17?

kolyshkin:
Backports:
* 1.16: https://github.com/cri-o/cri-o/pull/3370
* 1.17: https://github.com/cri-o/cri-o/pull/3371

PR #3342, opened 2020-03-02
seccomp: change empty profile behavior

benclarkwood:
If seccomp is enabled, but the profile is empty, fall back to the default seccomp profile rather than unconfined.

Fix for #3079

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -634,12 +634,12 @@ func isInCRIMounts(dst string, mounts []*pb.Mount) bool {
 }
 
 func (s *Server) setupSeccomp(ctx context.Context, specgen *generate.Generator, profile string) error {
-	if profile == "" {
-		// running w/o seccomp, aka unconfined
-		specgen.Config.Linux.Seccomp = nil
-		return nil
-	}
 	if s.Config().Seccomp().IsDisabled() {
+		if profile == "" {
+			// seccomp disabled and no profile specified
+			specgen.Config.Linux.Seccomp = nil
+			return nil
+		}
 		if profile != seccompUnconfined {
 			return fmt.Errorf("seccomp is not enabled in your kernel, cannot run with a profile")
 		}
@@ -651,8 +651,8 @@ func (s *Server) setupSeccomp(ctx context.Context, specgen *generate.Generator, 
 		return nil
 	}
 
-	// Load the default seccomp profile from the server if the profile is a default one
-	if profile == seccompRuntimeDefault || profile == seccompDockerDefault {
+	// Load the default seccomp profile from the server if the profile is a default one or empty
+	if profile == "" || profile == seccompRuntimeDefault || profile == seccompDockerDefault {
 		linuxSpecs, err := seccomp.LoadProfileFromConfig(s.Config().Seccomp().Profile(), specgen.Config)
 		if err != nil {
 			return err


haircommander:
/ok-to-test

Thanks for the PR! I'm unsure this works with the CRI definition but let's see 😄

saschagrunert:
> /ok-to-test
> 
> Thanks for the PR! I'm unsure this works with the CRI definition but let's see

Do you have any resource that the CRI should behave like this?

server/container_create.go
```diff
@@ -634,12 +634,12 @@ func isInCRIMounts(dst string, mounts []*pb.Mount) bool {
 }
 
 func (s *Server) setupSeccomp(ctx context.Context, specgen *generate.Generator, profile string) error {
-	if profile == "" {
-		// running w/o seccomp, aka unconfined
-		specgen.Config.Linux.Seccomp = nil
-		return nil
-	}
 	if s.Config().Seccomp().IsDisabled() {
+		if profile == "" {
```
saschagrunert: The test in `ctr_seccomp.bats` is now failing as well, I think we have to change it too.

server/container_create.go
```diff
@@ -651,8 +651,8 @@ func (s *Server) setupSeccomp(ctx context.Context, specgen *generate.Generator,
 		return nil
 	}
 
-	// Load the default seccomp profile from the server if the profile is a default one
-	if profile == seccompRuntimeDefault || profile == seccompDockerDefault {
+	// Load the default seccomp profile from the server if the profile is a default one or empty
+	if profile == "" || profile == seccompRuntimeDefault || profile == seccompDockerDefault {
```
saschagrunert: I would be good if we could add this behavior to criocli.go:
https://github.com/cri-o/cri-o/blob/5996f36f459ea2143b50dc98468e27ef69487e0f/internal/pkg/criocli/criocli.go#L463-L464

benclarkwood: > If not specified, then the internal default seccomp profile will be used

doesn't that imply the behavior?

saschagrunert: Somehow yes. 😬

server/container_create.go
mrunalp: http://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtimemptye/v1alpha2/api.proto#L282

rhatdan:
@benclarkwood I think this might need a rebase.

JAORMX:
Would be great to get this fixed.

JAORMX:
@benclarkwood will you have time to get back to this? Or do you want help in fixing up the tests?

haircommander:
based on the results of both critest runs:
```
/go/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/security_context_linux.go:608 Expected     <string>: Seccomp:	2      to contain substring     <string>: 0 /go/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/security_context_linux.go:1120
```
it seems this is in violation of the cri. I'm not sure we can do anything here without having the cri changed

JAORMX:
Would this enhancement help change the CRI default? https://github.com/kubernetes/enhancements/pull/1148

haircommander:
> Would this enhancement help change the CRI default? [kubernetes/enhancements#1148](https://github.com/kubernetes/enhancements/pull/1148)

it seems to be the first step. the proposal would only declare it GA and make it a first class pod spec value. Further work and proposals would be needed to have the CRI default to a profile when none is specified

haircommander:
closing this, as we have https://github.com/cri-o/cri-o/pull/4212 now

PR #3344, opened 2020-03-02
Use `Value` field in CLI for non-default values

saschagrunert:
#### What type of PR is this?

/kind cleanup

If the default value of a config contains a non-default CLI value, then
we can use the correct `Value: …` field. The documentation and fish
completions now changes slightly, but this should be the more correct
usage of the urfave/cli framework.

None
None

```release-note
None
```


completions/fish/crio.fish
```diff
@@ -91,33 +91,33 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l log-size-max -r -d 'Maximu
...
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-cert -r -d 'Path to the x509 certificate file used to serve the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes (default: "")'
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-key -r -d 'Path to the key file used to serve the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes (default: "")'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l uid-mappings -r -d 'Specify the UID mappings to use for the user namespace (default: "")'
-complete -c crio -n '__fish_crio_no_subcommand' -l version-file -r -d 'Location for CRI-O to lay down the version file (default: /var/run/crio/version)'
+complete -c crio -n '__fish_crio_no_subcommand' -l version-file -r -d 'Location for CRI-O to lay down the version file'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l help -s h -d 'show help'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l version -s v -d 'print the version'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l help -s h -d 'show help'
```
TomSweeneyRedHat: just curious as to why you're removing all the defaults from this file?

TomSweeneyRedHat:
One question, otherwise LGTM

completions/fish/crio.fish
saschagrunert: Those descriptions are only inherited from the `Usage` field of the flags, so if we remove the `(default: …)` from it then they're gone here as well.

haircommander:
LGTM

PR #3345, opened 2020-03-02
Add runc, conmon, crictl and CNI plugins to bundle

saschagrunert:
#### What type of PR is this?

/kind feature

This makes the bundle more universal and self-contained.

None
None

```release-note
- Added `conmon`, `runc` and `cni-plugins` to the static release bundle
```

### DIFF
--- a/bundle/README.md
+++ b/bundle/README.md
@@ -1,17 +1,6 @@
 # CRI-O static build bundle
 
-The following runtime dependencies are needed to let CRI-O work in conjunction
-with this bundle:
-
-- [runc][0]
-- [conmon][1]
-- [CNI plugins][2]
-
-[0]: https://github.com/opencontainers/runc
-[1]: https://github.com/containers/conmon
-[2]: https://github.com/containernetworking/plugins
-
-To install the bundle, run:
+To install the bundle, run the following from a development sandbox:
 
 ```
 $ sudo make install
--- a/bundle/build
+++ b/bundle/build
@@ -7,6 +7,11 @@ source "$(dirname "${BASH_SOURCE[0]}")"/vars
 
 cd "$(dirname "$0")"
 
+CONMON_VERSION=v2.0.11
+RUNC_VERSION=v1.0.0-rc10
+CNI_PLUGINS_VERSION=v0.8.5
+CRICTL_VERSION=v1.17.0
+
 FILES_BIN=(
     "../bin/crio-status-x86_64-static-glibc"
     "../bin/crio-status-x86_64-static-musl"
@@ -85,8 +90,36 @@ if [[ $ERRORED == 1 ]]; then
     exit 1
 fi
 
+# Local assets
 cp Makefile $TMPDIR
 cp README.md $TMPDIR
+
+curl_to() {
+    curl -sfL -o "$1" "$2"
+}
+TMP_BIN=$TMPDIR/bin
+
+# conmon
+curl_to $TMP_BIN/conmon \
+    https://github.com/containers/conmon/releases/download/$CONMON_VERSION/conmon
+chmod +x $TMP_BIN/conmon
+
+# runc
+curl_to $TMP_BIN/runc \
+    https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64
+chmod +x $TMP_BIN/runc
+
+# CNI plugins
+mkdir -p "$TMPDIR"/cni-plugins
+set -x
+curl_to - \
+    https://github.com/containernetworking/plugins/releases/download/$CNI_PLUGINS_VERSION/cni-plugins-linux-amd64-$CNI_PLUGINS_VERSION.tgz |
+    tar xfz - -C "$TMPDIR"/cni-plugins
+
+# crictl
+curl_to - \
+    https://github.com/kubernetes-sigs/cri-tools/releases/download/$CRICTL_VERSION/crictl-$CRICTL_VERSION-linux-amd64.tar.gz |
+    tar xfz - -C "$TMP_BIN"
 
 # Create the archive
 BUNDLE=crio-$(git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -251,9 +251,6 @@ jobs:
       - attach_workspace:
           at: .
       - run:
-          name: setup environment
-          command: scripts/circle-setup
-      - run:
           name: test
           command: sudo bundle/test
 
--- a/bundle/Makefile
+++ b/bundle/Makefile
@@ -10,19 +10,41 @@ BASHINSTALLDIR ?= $(PREFIX)/share/bash-completion/completions
 FISHINSTALLDIR ?= $(PREFIX)/share/fish/completions
 ZSHINSTALLDIR ?= $(PREFIX)/share/zsh/site-functions
 SYSTEMDDIR ?= $(PREFIX)/lib/systemd/system
+OPT_CNI_BIN_DIR ?= /opt/cni/bin
 
 DEFAULT_BINARY ?= crio-x86_64-static-musl
 
 all: install
 
 .PHONY: install
-install:
+install: \
+	install-cni \
+	install-conmon \
+	install-crio \
+	install-crictl \
+	install-pinns \
+	install-runc
+
+.PHONY: install-cni
+install-cni:
+	install $(SELINUX) -d -m 755 $(CNIDIR)
+	install $(SELINUX) -D -m 755 -t $(OPT_CNI_BIN_DIR) cni-plugins/*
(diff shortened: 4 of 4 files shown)


dougsland:
Maybe we should implement auto-tools for checking if the requirements are available before building? "make install" seems more related to the project itself and not for requirements?

saschagrunert:
> Maybe we should implement auto-tools for checking if the requirements are available before building? "make install" seems more related to the project itself and not for requirements?

Hm, we basically could check for `runc`/`conmon` available in `$PATH` and do not install it in that case. :thinking: @dougsland do you want to follow-up on that? I think we can add it later on (before 1.18.0).

dougsland:
> > Maybe we should implement auto-tools for checking if the requirements are available before building? "make install" seems more related to the project itself and not for requirements?
> 
> Hm, we basically could check for `runc`/`conmon` available in `$PATH` and do not install it in that
> case.  @dougsland do you want to follow-up on that? I think we can add it later on (before 1.18.0).

Sure. It's related to the work I am doing right now.

bundle/README.md
```diff
@@ -1,16 +1,5 @@
 # CRI-O static build bundle
```
TomSweeneyRedHat: I know this is pre-existing, and probably known by 99% of the people reading this, but should we add something like:

To install the bundle, run the following from a development sandbox:

saschagrunert: Fixed :+1:

bundle/README.md
```diff
@@ -1,17 +1,6 @@
...
-- [runc][0]
-- [conmon][1]
-- [CNI plugins][2]
-
-[0]: https://github.com/opencontainers/runc
-[1]: https://github.com/containers/conmon
-[2]: https://github.com/containernetworking/plugins
-
-To install the bundle, run:
+To install the bundle, the following from a development sandbox:
```
TomSweeneyRedHat: "the following" -> "run the following"

saschagrunert: Fixed as well, thanks!

rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 @cevich @mrunalp @kolyshkin @haircommander PTAL

bundle/build
```diff
@@ -85,9 +90,40 @@ if [[ $ERRORED == 1 ]]; then
...
 cp README.md $TMPDIR
 
+curl_to() {
+    curl -sfL -o "$1" "$2"
+}
+TMP_BIN=$TMPDIR/bin
+
+# conmon
+curl_to $TMP_BIN/conmon \
+    https://github.com/containers/conmon/releases/download/$CONMON_VERSION/conmon
```
dougsland: For security reasons, I would check the sha256sum from each file from curl.

bundle/build
```diff
@@ -85,9 +90,40 @@ if [[ $ERRORED == 1 ]]; then
...
+chmod +x $TMP_BIN/conmon
+
+# runc
+curl_to $TMP_BIN/runc \
```
dougsland: the same here.

bundle/build
```diff
@@ -85,9 +90,40 @@ if [[ $ERRORED == 1 ]]; then
...
+    https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64
+chmod +x $TMP_BIN/runc
+
+# CNI plugins
+CNI_PLUGINS=$TMPDIR/cni-plugins.tgz
+curl_to $CNI_PLUGINS \
```
dougsland: the same here.

bundle/build
```diff
@@ -85,9 +90,40 @@ if [[ $ERRORED == 1 ]]; then
...
+    https://github.com/containernetworking/plugins/releases/download/$CNI_PLUGINS_VERSION/cni-plugins-linux-amd64-$CNI_PLUGINS_VERSION.tgz
+mkdir $TMPDIR/cni-plugins
+tar xfz $CNI_PLUGINS -C $TMPDIR/cni-plugins
+rm $CNI_PLUGINS
+
+# crictl
+CRICTL=$TMPDIR/crictl.tar.gz
+curl_to $CRICTL \
```
dougsland: the same here.

bundle/build
```diff
@@ -85,9 +90,40 @@ if [[ $ERRORED == 1 ]]; then
...
+    https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64
+chmod +x $TMP_BIN/runc
+
+# CNI plugins
+CNI_PLUGINS=$TMPDIR/cni-plugins.tgz
+curl_to $CNI_PLUGINS \
+    https://github.com/containernetworking/plugins/releases/download/$CNI_PLUGINS_VERSION/cni-plugins-linux-amd64-$CNI_PLUGINS_VERSION.tgz
+mkdir $TMPDIR/cni-plugins
+tar xfz $CNI_PLUGINS -C $TMPDIR/cni-plugins
+rm $CNI_PLUGINS
```
kolyshkin: You can probably do this in place (i.e. without saving the tarball), something like
`curl $URL | tar -C $DESTDIR xf -`

bundle/build
```diff
@@ -85,9 +90,40 @@ if [[ $ERRORED == 1 ]]; then
...
+
+# crictl
+CRICTL=$TMPDIR/crictl.tar.gz
+curl_to $CRICTL \
+    https://github.com/kubernetes-sigs/cri-tools/releases/download/$CRICTL_VERSION/crictl-$CRICTL_VERSION-linux-amd64.tar.gz
+tar xfz $CRICTL -C $TMP_BIN
```
kolyshkin: ditto

bundle/build
kolyshkin: or `curl_to - $URL | tar -C $DESTDIR xf -`

bundle/build
kolyshkin: @dougsland Do you mean to hardcode the checksum next to version and check against it?

Something like
```shell
CONMON_VERSION=v2.0.11
CONMON_CHECKSUM=xxxxxx
....
curl_to() {
￼    curl -sfL -o "$1" "$2"
    if ! -z $3 -a ! sha512sum "$1" | grep -F "^$3"; then
        echo 2>&1 "Bad checksum for $2"
        exit 1
    fi
}
...

curl_to $TMP_BIN/conmon $CONMON_URL $CONMON_CHECKSUM
￼    https://github.com/containers/conmon/releases/download/$CONMON_VERSION/conmon
```

I mean, this will work, but it would be harder to bump versions

dougsland: @kolyshkin Not really, I thinking something like the pseudo-code bellow. However, I don't see in the assets file a conmon-$release.sha256sum available in here: https://github.com/containers/conmon/releases/

@haircommander might be a good idea to have a sha256sum file in case we don't have for validation. As example, runc provides it, see: https://github.com/opencontainers/runc/releases


  tmpdir_wget=$(mktemp -d)
  pushd "${tmpdir_wget}" &> /dev/null || exit

  wget conmon
  wget conmon.sha256sum

  sha256sum_from_tarfile=$(sha256sum conmon-$version.tar | awk '{print $1}')
  sha256sum_from_web=$(cat conmon-sha256-file)

    if [ "${sha256sum_from_tarfile}" != "${sha256sum_from_web}" ]; then
        fail....
        exit 1
    fi
    echo "passed..." 
   
  popd &> /dev/null || exit
  ${RM_BIN} -rf "${tmpdir_wget}"

bundle/build
saschagrunert: Fixed :+1:

bundle/build
saschagrunert: Fixed :+1:

bundle/build
saschagrunert: Hm, I'd like to follow-up on that in another PR. Is that okay for you?

bundle/build
saschagrunert: See https://github.com/cri-o/cri-o/pull/3345#discussion_r389588866

bundle/build
saschagrunert: See https://github.com/cri-o/cri-o/pull/3345#discussion_r389588866

bundle/build
saschagrunert: See https://github.com/cri-o/cri-o/pull/3345#discussion_r389588866

bundle/build
```diff
@@ -7,6 +7,11 @@ source "$(dirname "${BASH_SOURCE[0]}")"/vars
 
 cd "$(dirname "$0")"
 
+CONMON_VERSION=v2.0.11
+RUNC_VERSION=v1.0.0-rc10
+CNI_PLUGINS_VERSION=v0.8.5
+CRICTL_VERSION=v1.17.0
```
haircommander: I'm going to give this the lgtm, but can we source these from `scripts/versions` somehow?

saschagrunert: Ah yes I’ll propose this in a PR next week :)

PR #3349, opened 2020-03-03

nalind:
#### What type of PR is this?

/kind bug

Per https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-limits-are-run, when kubernetes runs a container with a memory usage limit set, it expects results comparable to using `docker run --memory=...`.

Per https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details, if the --memory flag is used, but the --memory-swap flag is _not_ used, the swap space limit value is expected to be equal to the memory limit value.

This swap space limit _includes_ the memory limit, though, so kubernetes effectively expects us to not allow the container to use swap.

None

None

```release-note
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -495,6 +495,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 				return nil, fmt.Errorf("set memory limit %v too low; should be at least %v", memoryLimit, minMemoryLimit)
 			}
 			specgen.SetLinuxResourcesMemoryLimit(memoryLimit)
+			specgen.SetLinuxResourcesMemorySwap(memoryLimit)
 
 			specgen.SetProcessOOMScoreAdj(int(resources.GetOomScoreAdj()))
 			specgen.SetLinuxResourcesCPUCpus(resources.GetCpusetCpus())
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -1251,6 +1251,16 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "$output" =~ "209715200" ]]
 
+	if test -r /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes ; then
+		expected=209715200
+	else
+		expected=0
+	fi
+	run crictl exec --sync "$ctr_id" sh -c "cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes 2> /dev/null || cat /sys/fs/cgroup/memory.swap.max 2> /dev/null"
+	echo "$output"
+	[ "$status" -eq 0 ]
+	[ "$output" -eq "$expected" ]
+
 	if test $(stat -f -c%T /sys/fs/cgroup) = cgroup2fs; then
 		run crictl exec --sync "$ctr_id" sh -c "cat /sys/fs/cgroup/cpu.max"
 		echo "$output"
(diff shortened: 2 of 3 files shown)


TomSweeneyRedHat:
LGTM assuming happy tests

haircommander:
failures look like registry flakes

/retest

e2e_{crun,fedora} failures are not the fault of this PR

saschagrunert:
@nalind May I ask you to use our pull request template? We will use it for release notes generation later and and this means that our PRs need proper labeling and formatting. :)

test/ctr.bats
```diff
@@ -1251,6 +1251,11 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "$output" =~ "209715200" ]]
 
+	run crictl exec --sync "$ctr_id" sh -c "cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes || cat /sys/fs/cgroup/memory.swap.max"
```
giuseppe: the swap limits work differently on cgroup v2.

in cgroup v1 the memory swap limit accounts also for the memory max, so `memory.memsw.limit_in_bytes` is really `memory limit + swap`, while on cgroup v2 it is just the net `swap` limit.

If you try on cgroup v1 to set `memory.memsw.limit_in_bytes` lower than `memory.mem.limit_in_bytes` the kernel will complain.

`crun` maintains the same semantic of the cgroup v1 OCI configuration also with cgroup v2, so the value that is set for `SetLinuxResourcesMemorySwap()` will be converted to `SWAP_LIMIT - MEMORY_LIMIT`:

on cgroup v1:
```
# podman run --rm -ti --memory 100M --memory-swap=100M fedora cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes 
104857600
```

on cgroup v2:
```
# podman run --rm -ti --memory 100M --memory-swap 100M alpine cat /sys/fs/cgroup/memory.swap.max
0
```

both of these containers cannot use any swap memory.

nalind: Is this the same behavior we get from runc?  If so, I'm fine with changing the logic here to set the swap value to double the memory value.

giuseppe: it will eventually follow the same behavior otherwise it is a breaking change for the cgroup v1 configuration

nalind: Right, so the code's doing the right thing (effectively disabling swap), but the test needs to be fixed.  Got it.

nalind:
Updated the description, updated tests per @giuseppe's feedback (thanks!), rebased.

nalind:
Changed test comparisons from `=~` to `-eq`, since a lot of things match a pattern of `0`.

haircommander:
crun and kata failure looks like a flakes

/retest

haircommander:
@saschagrunert we're hitting `memory.memsw.limit_in_bytes: permission denied` on circle here. It seems the circle VMs don't allow swap accounting? is there a way we can set kernel cmd args in circle?

server/container_create_linux.go
```diff
@@ -495,6 +495,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 				return nil, fmt.Errorf("set memory limit %v too low; should be at least %v", memoryLimit, minMemoryLimit)
 			}
 			specgen.SetLinuxResourcesMemoryLimit(memoryLimit)
+			specgen.SetLinuxResourcesMemorySwap(memoryLimit)
```
haircommander: a work around for kernels not having this would be checking if `/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes` exists before setting this. I would think this would be a responsibility of the runtime though...

saschagrunert:
> @saschagrunert we're hitting `memory.memsw.limit_in_bytes: permission denied` on circle here. It seems the circle VMs don't allow swap accounting? is there a way we can set kernel cmd args in circle?

Yes we can try to change the kernel parameters in `scripts/circle-setup`.

haircommander:
I am taking this over here https://github.com/cri-o/cri-o/pull/3497

PR #3354, opened 2020-03-04
[1.16] Check for context erroring before returning from longer requests

umohnani8:

#### What type of PR is this?

/kind bug

The issue seems to be that the kubelet times out on the create request but crio doesn't error out when kubelet client times out. So, we end up with the pod being created and then kubelet keeps on trying to create the pod again and it can't since the name is reserved.

With the change, we error out on timeouts so the pod creation fails and the name is unreserved so later attempts by the kubelet to create a pod or container may succeed.

We may want to add more of these checks around long running operations rather than just at the end in follow-ups.

Possible fix for https://github.com/cri-o/cri-o/issues/2984

Taking over https://github.com/cri-o/cri-o/pull/3102

No

```release-note

```

### DIFF
--- a/contrib/test/integration/golang.yml
+++ b/contrib/test/integration/golang.yml
@@ -54,5 +54,4 @@
     - tools/godep
     - onsi/ginkgo/ginkgo
     - onsi/gomega
-    - cloudflare/cfssl/cmd/...
     - jteeuwen/go-bindata/go-bindata
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -623,7 +623,11 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 		log.Errorf(ctx, "%v", err)
 	}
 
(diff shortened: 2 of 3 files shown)


umohnani8:
@mrunalp @haircommander PTAL

haircommander:
jenkins seems sad

/retest

LGTM

mtrmac:
I still don’t understand why this actually fixes a bug (https://github.com/cri-o/cri-o/pull/3102#issuecomment-573867731 suggests this PR should typically have no effect at all, and  https://github.com/cri-o/cri-o/pull/3102#issuecomment-574625689 that this is a Kubelet responsibility anyway), but then I don’t understand how the problem occurs…

umohnani8:
How can I re-run the circle ci? I don't have the permission to rerun it from the interface

haircommander:
I retriggered it. I don't remember how I got permission to do that. @saschagrunert gave it to me I think

umohnani8:
Thanks @haircommander! this is ready

PR #3362, opened 2020-03-05
Create PID file at startup

haircommander:
#### What type of PR is this?

/kind feature

Create a PID file at CRI-O's startup.

There are cases where it'd be useful to write a PID file when CRI-O starts up. Specifically, a PID file is a good way to for cadvisor to track usage in the crio cgroup in a way that's agnostic of systemd.

None

None

```release-note
CRI-O now writes its PID to a file at startup. The location can be configured with config value pid_path, or CLI value --pid-path```


haircommander:
PTAL @joelsmith

Would you be against using /var/run/crio/pid instead of /var/run/crio.pid? the former is more in line with how we format other files

mrunalp:
Can you check if using systemd `PIDFile=` makes sense to use with this?

haircommander:
> Can you check if using systemd `PIDFile=` makes sense to use with this?

I'd rather use a solution that doesn't rely on systemd if we don't have to

joelsmith:
Looks good. I'm waiting for a review on https://github.com/kubernetes/kubernetes/pull/88823 so it's good that this is done. Thanks!

haircommander:
closing as the kube PR is not needed

PR #3385, opened 2020-03-10
Fix unit tests for locally configured registries

saschagrunert:
#### What type of PR is this?

/kind cleanup

The system context is not stored any more in the image server, which
means that we cannot pass `nil` to sysregistries, because this will
access the host configured registries and might cause the unit tests to
fail.

None
`make testunit` should now work with any configuration in `/etc/containers/registries.conf`

```release-note
None
```
/cc @rhafer

### DIFF
--- a/internal/storage/image_test.go
+++ b/internal/storage/image_test.go
@@ -2,7 +2,6 @@ package storage_test
 
 import (
 	"context"
-	"io/ioutil"
 	"os"
 
 	"github.com/containers/image/v5/copy"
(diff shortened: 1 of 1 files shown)


rhafer:
fixes the issue we've been seeing when running `make testunit` locally. So LGTM.

TomSweeneyRedHat:
LGTM

haircommander:
gotta love flakes..
/retest

/lgtm

saschagrunert:
> gotta love provisioning flakes..

Do you know if we can solve that issue somehow? We just saw them recently

haircommander:
> > gotta love provisioning flakes..
> 
> Do you know if we can solve that issue somehow? We just saw them recently

our infra folks are going through some major rearchitectures right now. I'm not sure there's much we can do while that transition is happening.

PR #3363, opened 2020-03-05
[1.17] drop conmonmon

haircommander:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
drop conmonmon (as amazing as the name was)

conmonmon never really figured out how to handle a cgroup being torn down out of order (conmon before a container) and was generally pretty racy. With the integration of conmon 2.0.11, we now prevent conmon from ever being OOM killed. For now, drop conmonmon until a better solution is found
#### Which issue(s) this PR fixes:
should fix https://bugzilla.redhat.com/show_bug.cgi?id=1809593

#### Special notes for your reviewer:
In general, bumping conmon and droping this will lower the chance of false positive oom kills (where we think there was an oom but there actually wasn't)
#### Does this PR introduce a user-facing change?

```release-note
Fix bug resulting in false reports of OOM
```

PR #3389, opened 2020-03-10
Drop conmonmon 1.16

haircommander:
#### What type of PR is this?

/kind cleanup

drop conmonmon (as amazing as the name was)

cherry-pick of https://github.com/cri-o/cri-o/pull/3363

conmonmon never really figured out how to handle a cgroup being torn down out of order (conmon before a container) and was generally pretty racy. With the integration of conmon 2.0.11, we now prevent conmon from ever being OOM killed. For now, drop conmonmon until a better solution is found
should fix https://bugzilla.redhat.com/show_bug.cgi?id=1809593

In general, bumping conmon and droping this will lower the chance of false positive oom kills (where we think there was an oom but there actually wasn't)

```release-note
Fix bug resulting in false reports of OOM
```


haircommander:
oops

PR #3398, opened 2020-03-13
[1.18] stats: fix stats when systemd cgroups are used

haircommander:
#### What type of PR is this?

/kind bug
fixes cri stats for systemd cgroups

passing the slice directly to libpod.cgroups.Load() caused it to not be found. This is because it expected a fully qualified path, rather than a slice name.
write a function to translate slice name to path, and pass that down if we're using systemd cgroups

```release-note
Fixed a bug where systemd cgroups were incompatible with CRI stats endpoint
```

### DIFF
--- a/internal/oci/oci_linux.go
+++ b/internal/oci/oci_linux.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/containers/libpod/pkg/cgroups"
 	"github.com/cri-o/cri-o/utils"
+	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


haircommander:
```

not ok 35 ctr list filtering
# (in test file ./ctr.bats, line 647)
#   `[[ "$output" == "$ctr1_id" ]]' failed
# 0
```
I have never seen [that](https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3398/test_pull_request_crio_integration_rhel/10576/artifacts/testout_userns.txt) before...

/retest

haircommander:
we see the segfault here, but the fix is already upstream, and e2e_crun_cgroupv2 isn't blocking

@saschagrunert @umohnani8 @rhatdan @mrunalp PTAL

rhatdan:
/approve
LGTM

internal/oci/oci_linux.go
```diff
@@ -86,6 +87,16 @@ func newPipe() (parent, child *os.File, err error) {
 }
 
 func (r *runtimeOCI) containerStats(ctr *Container, cgroup string) (*ContainerStats, error) {
+	// this correction has to be made because the libpod cgroups package can't find a
+	// systemd cgroup that isn't converted to a fully qualified cgroup path
+	if r.config.CgroupManager == "systemd" {
```
saschagrunert:
```suggestion
	if r.config.CgroupManager == SystemdCgroupsManager {
```

internal/oci/oci_linux.go
```diff
@@ -86,6 +87,16 @@ func newPipe() (parent, child *os.File, err error) {
 }
 
 func (r *runtimeOCI) containerStats(ctr *Container, cgroup string) (*ContainerStats, error) {
```
saschagrunert:
```suggestion
func (r *runtimeOCI) containerStats(ctr *Container, cgroup string) (stats *ContainerStats, err error) {
```

internal/oci/oci_linux.go
```diff
@@ -86,6 +87,16 @@ func newPipe() (parent, child *os.File, err error) {
...
+	// this correction has to be made because the libpod cgroups package can't find a
+	// systemd cgroup that isn't converted to a fully qualified cgroup path
+	if r.config.CgroupManager == "systemd" {
+		var err error
```
saschagrunert:
```suggestion
```

PR #3400, opened 2020-03-13
systemd: use inactive-or-failed for CollectMode

haircommander:
#### What type of PR is this?

/kind cleanup

Configure every container to have its systemd CollectMode set to inactive-or-failed. This tells systemd to cleanup the cgroup if it failed

```release-note
none
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -533,13 +533,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		}
 		specgen.SetLinuxCgroupsPath(cgPath)
 
-		if t, ok := kubeAnnotations[podTerminationGracePeriodLabel]; ok {
-			// currently only supported by systemd, see
-			// https://github.com/opencontainers/runc/pull/2224
-			if useSystemd {
+		if useSystemd {
+			if t, ok := kubeAnnotations[podTerminationGracePeriodLabel]; ok {
+				// currently only supported by systemd, see
+				// https://github.com/opencontainers/runc/pull/2224
 				specgen.AddAnnotation("org.systemd.property.TimeoutStopUSec",
 					"uint64 "+t+"000000") // sec to usec
 			}
+			configureGeneratorWithCollectMode(specgen)
 		}
 
 		if privileged {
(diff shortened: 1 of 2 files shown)


haircommander:
PTAL @kolyshkin

haircommander:
huh, rhel 7.7 doesn't seem to have CollectMode

mrunalp:
@haircommander yeah, RHEL 7.x version of systemd doesn't have it.

server/container_create_linux.go
```diff
@@ -533,13 +533,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
...
-			// https://github.com/opencontainers/runc/pull/2224
-			if useSystemd {
+		if useSystemd {
+			if t, ok := kubeAnnotations[podTerminationGracePeriodLabel]; ok {
+				// currently only supported by systemd, see
+				// https://github.com/opencontainers/runc/pull/2224
 				specgen.AddAnnotation("org.systemd.property.TimeoutStopUSec",
 					"uint64 "+t+"000000") // sec to usec
 			}
+			configureGeneratorWithCollectMode(specgen)
```
mrunalp: This should be always applied whether termination grace is set or not.

server/sandbox_run_linux.go
```diff
@@ -712,6 +713,12 @@ func AddCgroupAnnotation(ctx context.Context, g generate.Generator, mountPath, c
 	return cgroupParent, nil
 }
 
+func configureGeneratorWithCollectMode(g generate.Generator) {
+	// we always want inactive-or-failed as the collect mode
+	// when using systemd
```
kolyshkin: here "when using systemd" does not make sense, because there is no such thing as "collect mode" without systemd

server/container_create_linux.go
haircommander: I switched the order of the blocks. this is now in the block `useSystemd {}` and is a sibling of the termination grace period check

rhatdan:
@haircommander Still working on this?

umohnani8:
@haircommander bumping this up, still working on this?

haircommander:
oops, sorry for ghosting, yeah I'm waiting on https://github.com/cri-o/cri-o/pull/3565. my plan is to use `internal/config/node` package to save whether we have the correct version of systemd to support this feature.

haircommander:
replaced by https://github.com/cri-o/cri-o/pull/3859

PR #3404, opened 2020-03-16
build: make uninstall improvements

dougsland:
#### What type of PR is this?

/kind bug

Make sure all installed files are removed.


dougsland:
The ci/circleci: git-validation error is not valid. I have added a empty (TAB) line for readability and give space between cleanup codes.

TomSweeneyRedHat:
LGTM once the tests are happy.

haircommander:
LGTM, though I don't really want to totally break git validation, so I'd prefer you fix as it is asking

PR #3409, opened 2020-03-16
avoid parallel pulls of the same image

vrothberg:
To suit the needs of OpenShift CI [1], we need to avoid parallel pulls.
While we have an open PR against containers/image [2] to detect parallel
copy operations across processes, we are still far from getting this
merged; it must be split into smaller chunks and we plan to tackle it in
the next few months but who knows how long this might take.

To support that in CRI-O, however, we enjoy the benefit of running a
daemon where we can easily avoid redundantly pulling the same image in
parallel as we're in the same process space and can do all
synchronisation in memory.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1785390
[2] https://github.com/containers/image/pull/611

### DIFF
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/storage"
 	"github.com/cri-o/cri-o/server/metrics"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
@@ -27,38 +28,89 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp 
 		image = img.Image
 	}
 
-	sourceCtx := *s.config.SystemContext // A shallow copy we can modify
+	pullArgs := pullArguments{image: image}
 	if req.GetAuth() != nil {
 		username := req.GetAuth().Username
 		password := req.GetAuth().Password
 		if req.GetAuth().Auth != "" {
 			username, password, err = decodeDockerAuth(req.GetAuth().Auth)
 			if err != nil {
-				log.Debugf(ctx, "error decoding authentication for image %s: %v", img, err)
+				log.Debugf(ctx, "error decoding authentication for image %s: %v", image, err)
 				return nil, err
 			}
 		}
 		// Specifying a username indicates the user intends to send authentication to the registry.
 		if username != "" {
-			sourceCtx.DockerAuthConfig = &types.DockerAuthConfig{
+			pullArgs.credentials = types.DockerAuthConfig{
 				Username: username,
 				Password: password,
 			}
 		}
 	}
 
+	// We use the server's pullOperationsInProgress to record which images are
+	// currently being pulled. This allows for avoiding pulling the same image
+	// in parallel. Hence, if a given image is currently being pulled, we queue
+	// into the pullOperation's waitgroup and wait for the pulling goroutine to
+	// unblock us and re-use its results.
+	pullOp, pullInProcess := func() (pullOp *pullOperation, inProgress bool) {
+		s.pullOperationsLock.Lock()
+		defer s.pullOperationsLock.Unlock()
+		pullOp, inProgress = s.pullOperationsInProgress[pullArgs]
+		if !inProgress {
+			pullOp = &pullOperation{}
+			s.pullOperationsInProgress[pullArgs] = pullOp
+			pullOp.wg.Add(1)
+		}
+		return pullOp, inProgress
+	}()
+
+	if !pullInProcess {
+		pullOp.err = errors.New("pullImage was aborted by a Go panic")
+		defer func() {
+			s.pullOperationsLock.Lock()
+			delete(s.pullOperationsInProgress, pullArgs)
+			pullOp.wg.Done()
+			s.pullOperationsLock.Unlock()
+		}()
+		pullOp.imageRef, pullOp.err = s.pullImage(ctx, &pullArgs)
+	} else {
+		// Wait for the pull operation to finish.
+		pullOp.wg.Wait()
+	}
+
+	if pullOp.err != nil {
+		return nil, pullOp.err
+	}
+
+	resp = &pb.PullImageResponse{
+		ImageRef: pullOp.imageRef,
+	}
+	return resp, nil
+}
+
+// pullImage performs the actual pull operation of PullImage. Used to separate
+// the pull implementation from the pullCache logic in PullImage and improve
+// readability and maintainability.
+func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string, error) {
+	var err error
+	sourceCtx := *s.config.SystemContext // A shallow copy we can modify
+	if pullArgs.credentials.Username != "" {
+		sourceCtx.DockerAuthConfig = &pullArgs.credentials
+	}
+
 	decryptConfig, err := getDecryptionKeys(s.config.DecryptionKeysPath)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
 	var (
 		images []string
 		pulled string
 	)
-	images, err = s.StorageImageServer().ResolveNames(s.config.SystemContext, image)
+	images, err = s.StorageImageServer().ResolveNames(s.config.SystemContext, pullArgs.image)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 	for _, img := range images {
 		var tmpImg types.ImageCloser
@@ -159,21 +211,21 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp 
 		pulled = img
 		break
 	}
+
 	if pulled == "" && err != nil {
-		return nil, err
+		return "", err
 	}
+
 	status, err := s.StorageImageServer().ImageStatus(s.config.SystemContext, pulled)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 	imageRef := status.ID
 	if len(status.RepoDigests) > 0 {
 		imageRef = status.RepoDigests[0]
 	}
-	resp = &pb.PullImageResponse{
-		ImageRef: imageRef,
-	}
-	return resp, nil
+
+	return imageRef, nil
 }
 
 func decodeDockerAuth(s string) (user, password string, err error) {
--- a/server/server.go
+++ b/server/server.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/cri-o/cri-o/internal/lib"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
@@ -61,6 +62,33 @@ type Server struct {
 	defaultIDMappings *idtools.IDMappings
 
 	updateLock sync.RWMutex
+
+	// pullOperationsInProgress is used to avoid pulling the same image in parallel. Goroutines
+	// will block on the pullResult.
+	pullOperationsInProgress map[pullArguments]*pullOperation
+	// pullOperationsLock is used to synchronize pull operations.
+	pullOperationsLock sync.Mutex
+}
+
+// pullArguments are used to identify a pullOperation via an input image name and
+// possibly specified credentials.
+type pullArguments struct {
+	image       string
+	credentials types.DockerAuthConfig
+}
+
+// pullOperation is used to synchronize parallel pull operations via the
+// server's pullCache.  Goroutines can block the pullOperation's waitgroup and
+// be released once the pull operation has finished.
+type pullOperation struct {
+	// wg allows for Goroutines trying to pull the same image to wait until the
+	// currently running pull operation has finished.
+	wg sync.WaitGroup
+	// imageRef is the reference of the actually pulled image which will differ
+	// from the input if it was a short name (e.g., alpine).
+	imageRef string
+	// err is the error indicating if the pull operation has succeeded or not.
+	err error
 }
 
 type certConfigCache struct {
@@ -324,12 +352,13 @@ func New(
 	}
 
 	s := &Server{
-		ContainerServer:   containerServer,
-		netPlugin:         netPlugin,
-		hostportManager:   hostportManager,
-		config:            *config,
-		monitorsChan:      make(chan struct{}),
-		defaultIDMappings: idMappings,
+		ContainerServer:          containerServer,
+		netPlugin:                netPlugin,
+		hostportManager:          hostportManager,
+		config:                   *config,
+		monitorsChan:             make(chan struct{}),
+		defaultIDMappings:        idMappings,
+		pullOperationsInProgress: make(map[pullArguments]*pullOperation),
 	}
 
 	if err := configureMaxThreads(); err != nil {


vrothberg:
Will fix the unit tests tomorrow

server/image_pull.go
```diff
@@ -18,24 +18,17 @@ import (
...
-// PullImage pulls a image with authentication config.
-func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp *pb.PullImageResponse, err error) {
-	// TODO: what else do we need here? (Signatures when the story isn't just pulling from docker://)
-	image := ""
-	img := req.GetImage()
-	if img != nil {
-		image = img.Image
-	}
-
+func (s *Server) pullImage(ctx context.Context, req *pb.PullImageRequest, image string) (string, error) {
```
haircommander: I generally prefer the called function to come after the calling function. That way it can be read top to bottom and go down the stack. Can you switch the location with PullImage?

haircommander: Also I think a comment explaining the reasoning behind having a public PullImage and private pullImage split up would be helpful

server/server.go
```diff
@@ -330,6 +339,7 @@ func New(
 		config:            *config,
 		monitorsChan:      make(chan struct{}),
 		defaultIDMappings: idMappings,
+		pullCache:         make(map[string]*pullResult),
```
haircommander: It's unclear to me where this pullCache is used

mrunalp:
@mtrmac ptal.

mrunalp:
@nalind ptal.

go.sum
```diff
@@ -798,6 +798,7 @@ github.com/stretchr/objx v0.2.0/go.mod h1:qt09Ya8vawLte6SNmTgCsAVtYtaKzEcn8ATUoH
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
+github.com/stretchr/testify v1.5.1 h1:nOGnQDM7FYENwehXlg/kFVnos3rEvtKTjRvOWSzb6H4=
```
saschagrunert: I think this change is unintended, right?

server/image_pull.go
vrothberg: Sounds good! I moved it down and made sure to add more verbose comments all around.

server/server.go
vrothberg: In `PullImage(...)`. I added more verbose comments (also to the internal types) to make the code more approachable.

go.sum
vrothberg: Not intended, thanks!

server/server.go
```diff
@@ -61,6 +61,25 @@ type Server struct {
 	defaultIDMappings *idtools.IDMappings
 
 	updateLock sync.RWMutex
+
+	// pullCache is used to avoid pulling the same image in parallel. Goroutines
+	// will block on the pullResult.
+	pullCache map[string]*pullOperation
```
mtrmac: Isn’t this more of a `pullsInProgress` than a “cache”?

server/image_pull.go
```diff
@@ -27,15 +27,57 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
+	pullOp := &pullOperation{}
+	s.pullCacheLock.Lock()
+	pr, pullInProcess := s.pullCache[image]
+	if pullInProcess {
+		pullOp = pr
+	} else {
+		s.pullCache[image] = pullOp
+		pullOp.wg.Add(1)
+	}
+	s.pullCacheLock.Unlock()
```
mtrmac: I’d feel safer with a `panic`-proof locking, i.e.

```go
lock.Lock()
func() {
    defer lock.Unlock()
    …
}()
```

OTOH if we are worried about that, there should also be a `defer` to mark a `pullCache` entry as failed on failed pull, and the `defer` scopes would overlap. Oops.

server/image_pull.go
```diff
@@ -27,15 +27,57 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
 		image = img.Image
 	}
 
+	// We use the server's pullCache to record which images are currently being
+	// pulled. This allows for avoiding pulling the same image in parallel.
+	// Hence, if a given image is currently being pulled, we queue into the
+	// pullOperation's waitgroup and wait for the pulling goroutine to unblock
+	// us and re-use its results.
+	pullOp := &pullOperation{}
```
mtrmac: Non-blocking: This could be `var pullOp *pullOperation`, initializing to `nil`; the heap allocation can be moved inside the `!pullInProcess` case.

server/server.go
```diff
@@ -61,6 +61,25 @@ type Server struct {
...
+	// pullCacheLock is used to synchronize the pullCache.
+	pullCacheLock sync.Mutex
+}
+
+// pullOperation is used to synchronize parallel pull operations via the
+// server's pullCache.  Goroutines can block the pullOperation's waitgroup and
+// be released once the pull operation has finished.
+type pullOperation struct {
+	// imageRef is the reference of the pulled image.
```
mtrmac: Non-blocking: “actually pulled image” (which makes a difference if the request used a short name)?

server/server.go
```diff
@@ -61,6 +61,25 @@ type Server struct {
...
+	imageRef string
+	// err is the error indicating if the pull operation has succeeded or not.
+	err error
+	// wg allows for Goroutines trying to pull the same image to wait until the
+	// currently running pull operation has finished.
+	wg sync.WaitGroup
```
mtrmac: (Absolutely non-blocking: Move the `wg` first, to have the fields in order of use, and to help anyone reading these doc strings to understand the concepts in order)

server/image_pull.go
```diff
@@ -27,15 +27,57 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
+		// Unblock waiting goroutines and delete the entry from the cache.
+		s.pullCacheLock.Lock()
+		delete(s.pullCache, image)
+		pullOp.wg.Done()
+		s.pullCacheLock.Unlock()
+	} else {
+		pullOp.wg.Wait()
+	}
+
+	if pullOp.imageRef == "" && pullOp.err != nil {
```
mtrmac: Why not just `pullOp.err != nil`? That implies the other condition anyway, and even if it were possible for an `imageRef != "" && err != nil` to happen, the code below would silently discard `err`, so that case is not really meaningful for the rest of the code.

vrothberg:
> My primary concern: Why is it safe to share the credentials across parallel requests?

I was thinking about that but concluded that images are shared in the storage in any case. However, this argument is weak and (if not for security) at least for the sake of determinism the key to the pullCache should incorporate:
* requested image name
* credentials
* decryption keys

Anything missing?

mtrmac:
Looking at the request object, only the image name and credentials are actually used (the encryption keys are purely theoretical currently, K8s does not have the concept yet).

server/server.go
vrothberg: Renamed to `pullOperationsInProgress` :+1:

server/image_pull.go
vrothberg: Nice idea!

server/image_pull.go
vrothberg: I addressed all comments but this one. We can spin off a new routine for each critical section but I feel it might be more harmful (wrt performance) in the general case. WDYT?

vrothberg: Hm ... now I can't live without. I'll address this one to :^)

vrothberg: Done.

vrothberg:
Thanks for the review, @mtrmac! Your comments are now addressed, PTanotherL.

vrothberg:
@mrunalp, do we need this in 1.17 as well?

haircommander:
> @mrunalp, do we need this in 1.17 as well?

why not :smile: 

/cherrypick release-1.17

/test kata

server/image_pull.go
```diff
@@ -27,38 +27,75 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
+	resp = &pb.PullImageResponse{
+		ImageRef: pullOp.imageRef,
+	}
+	return resp, nil
+}
+
+// pullImage performs the actual pull operation of PullImage. Used to separate
+// the pull implementation from the pullCache logic in PullImage and improve
+// readability and maintainability.
+func (s *Server) pullImage(ctx context.Context, req *pb.PullImageRequest, pullArgs *pullArguments) (string, error) {
```
haircommander: req is unused (lint is complaining)

vrothberg: Thanks!

vrothberg:
make: *** [Makefile:136: lint] Killed

... woops

server/image_pull.go
mtrmac: I am not one bit worried about the cost of a few stack frames, or at worst, heap allocations and atomic operations, when this is going to be blocked on downloads, decompression, and I/O in gigabytes, when trading it off for correctness.

But my initial guess was that this wasn’t going to be easy to do comprehensively anyway, so there was little point. (And, I forgot to mention, CRI-O is a top-level application, and has only one `recover` on a rare paths, so `panic`s are not _that_ much of a concern.)

For comprehensive `panic`-safety it would be necessary to also use `defer` using `defer` to make sure `pullOp.wg.Gone()` is _always_ called. And all of that makes the code very messy.

server/image_pull.go
```diff
@@ -27,38 +27,75 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
+	// pulled. This allows for avoiding pulling the same image in parallel.
+	// Hence, if a given image is currently being pulled, we queue into the
+	// pullOperation's waitgroup and wait for the pulling goroutine to unblock
+	// us and re-use its results.
+	pullOp, pullInProcess := s.pullOperation(&pullArgs)
+	if !pullInProcess {
+		go func() {
+			pullOp.imageRef, pullOp.err = s.pullImage(ctx, &pullArgs)
+			// Unblock waiting goroutines and delete the entry from the cache.
+			s.pullOperationsLock.Lock()
```
mtrmac: Non-blocking I guess: it’s not symmetric that starting a pull and finishing it are not both a named function or both an in-line lambda.

server/image_pull.go
```diff
@@ -27,38 +27,75 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
 
+	// We use the server's pullCache to record which images are currently being
+	// pulled. This allows for avoiding pulling the same image in parallel.
+	// Hence, if a given image is currently being pulled, we queue into the
+	// pullOperation's waitgroup and wait for the pulling goroutine to unblock
+	// us and re-use its results.
+	pullOp, pullInProcess := s.pullOperation(&pullArgs)
+	if !pullInProcess {
+		go func() {
+			pullOp.imageRef, pullOp.err = s.pullImage(ctx, &pullArgs)
```
mtrmac: If you want to deal with `panic`-safety in full:
```go
pullOp.err = errors.New("pullImage was aborted by a Go panic")
defer /* pullOp.wg.Done() etc., under the lock */
pullOp.ImageRef, pullOp.Err = s.pullImage(…)
```
(Or if you decide that it’s not worth it, I’m fine with that as well.)

server/image_pull.go
```diff
@@ -27,38 +27,75 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
+			// Unblock waiting goroutines and delete the entry from the cache.
+			s.pullOperationsLock.Lock()
+			defer s.pullOperationsLock.Unlock()
+			defer pullOp.wg.Done()
+			delete(s.pullOperationsInProgress, pullArgs)
```
mtrmac: Hum… if _either_ of the `.Done()` or the `delete` crashed, some waiter could block indefinitely. I can’t see how we can win here, we might just as well let the `panic` (at least in either of the two) bubble up and let Systemd restart us (if that’s what happens).

server/image_pull.go
mtrmac: > CRI-O is a top-level application

(I mean that libraries have to be stricter about their behavior on `panic` because they typically don’t understand and/or can’t influence the top-level application error recovery strategy.)

server/image_pull.go
vrothberg: Both lambdas now.

server/image_pull.go
```diff
@@ -27,38 +28,88 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
+		s.pullOperationsLock.Lock()
+		defer s.pullOperationsLock.Unlock()
+		pullOp, inProgress = s.pullOperationsInProgress[*pullArgs]
+		if !inProgress {
+			pullOp = &pullOperation{}
+			s.pullOperationsInProgress[*pullArgs] = pullOp
+			pullOp.wg.Add(1)
+		}
+		return pullOp, inProgress
+	}(&pullArgs)
```
mtrmac: (This parameter isn’t necessary, the lambda can use `pullArgs` from the surrounding scope directly.)

server/image_pull.go
```diff
@@ -27,38 +28,88 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp
...
+	return resp, nil
+}
+
+// pullImage performs the actual pull operation of PullImage. Used to separate
+// the pull implementation from the pullCache logic in PullImage and improve
+// readability and maintainability.
+func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string, error) {
+	var err error
+	sourceCtx := *s.config.SystemContext // A shallow copy we can modify
+	sourceCtx.DockerAuthConfig = &pullArgs.credentials
```
mtrmac: This unconditional assignment changes semantics — if `DockerAuthConfig != nil`, lookup in other credential stores is skipped.

So, that needs to happen only if `credentials.username != ""` (per the previous condition).

server/image_pull.go
mtrmac: Actually, this is already the best it can get: `wg.Done()` ensures the caller doesn’t block, and leaving the operation in the map will let any future callers to find it, not block, and progress; they’ll ”only” be unable to try the same pull again, which is pretty good considering the failure.

saschagrunert:
Can we have tests for that?

vrothberg:
> Can we have tests for that?

We decided for manual tests.

vrothberg:
What's up with the CI? It takes hours for the jobs to start.

haircommander:
so I'm not sure this is the fault of this PR, but we've hit a segfault, which is why e2e_rhel just failed
https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3409/test_pull_request_crio_e2e_rhel/11062/artifacts/journals/crio.service

haircommander:
shouldn't be the fault of this PR, and I've opened https://github.com/cri-o/cri-o/pull/3425 to fix, so no need to keep the failure around

/test e2e_rhel

haircommander:
/retest

> What's up with the CI? It takes hours for the jobs to start.

I'm not sure but CI has sure been funky lately...

haircommander:
/retest

(are the prow issues fixed :thinking: )

vrothberg:
Rebased, fingers crossed :crossed_fingers:

vrothberg:
Ready to merge.

PR #3415, opened 2020-03-17
Simplify container log path handling

saschagrunert:
#### What type of PR is this?

/kind cleanup

The `LogPath` inside the container config should be mostly relative
according to:

https://github.com/kubernetes/kubernetes/blob/84dc7046797aad80f258b6740a98e79199c8bb4d/pkg/kubelet/kuberuntime/helpers.go#L167-L170

This means we can remove the annoying warning for every container and
just fallback to a default log path if empty.

None
None

```release-note
- Removed warning about non-absolute container log paths when creating a container
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -453,19 +453,16 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		specgen.SetProcessApparmorProfile(profile)
 	}
 
-	logPath := containerConfig.GetLogPath()
 	sboxLogDir := sandboxConfig.GetLogDirectory()
 	if sboxLogDir == "" {
 		sboxLogDir = sb.LogDir()
 	}
+
+	logPath := containerConfig.GetLogPath()
 	if logPath == "" {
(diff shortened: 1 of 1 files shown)


haircommander:
LGTM
@mrunalp @umohnani8 @rhatdan PTAL

server/container_create_linux.go
```diff
@@ -453,19 +453,16 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
...
 		sboxLogDir = sb.LogDir()
 	}
+
+	logPath := containerConfig.GetLogPath()
 	if logPath == "" {
 		logPath = filepath.Join(sboxLogDir, containerID+".log")
-	}
-	if !filepath.IsAbs(logPath) {
-		// XXX: It's not really clear what this should be versus the sbox logDirectory.
-		log.Warnf(ctx, "requested logPath for ctr id %s is a relative path: %s", containerID, logPath)
```
rhatdan: Do we no longer need to worry about relative paths?

PR #3416, opened 2020-03-17
Remove utils.ExecCmdWithStdStreams in favor of utils.ExecCmd

saschagrunert:
#### What type of PR is this?

/kind cleanup

We only use standard streams so we can fallback to the `utils.ExecCmd`
function. Beside that, we printed results of lower level runtimes (like
runc) directly to stdout/err which is annoying when triaging the CRI-O logs.

None
None

```release-note
- Removed annoying logs coming directly from lower level runtimes like runc
```


haircommander:
LGTM, you may need to repoke lint job though

PR #3419, opened 2020-03-18
Mention starting cri-o for running with kubernetes

tedyu:
#### What type of PR is this?

/kind bug

Starting cri-o is required for running kubernetes with cri-o. Otherwise you would see:
```
W0318 08:26:30.130587   36467 clientconn.go:1208] grpc: addrConn.createTransport failed to connect to {/var/run/crio/crio.sock  <nil> 0 <nil>}. Err :connection error: desc = "transport: Error while dialing dial unix /var/run/crio/crio.sock: connect: no such file or directory". Reconnecting...
```

None
```release-note

```


haircommander:
on second thought, no need to run tests, this is docs only. merging manually

PR #3422, opened 2020-03-19
Isolate ContainerServer from Server package

saschagrunert:
#### What type of PR is this?

/kind cleanup

To be able to refactor further we need to split the ContainerServer from
the Server implementation. This allows us to use the ContainerServer in
the new interface without having the need to rely on the *Server.

None
/cc @rhafer 

```release-note
None
```

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -16,6 +16,7 @@ import (
 	_ "github.com/containers/libpod/pkg/hooks/0.1.0"
 	"github.com/containers/storage/pkg/reexec"
 	"github.com/cri-o/cri-o/internal/criocli"
+	"github.com/cri-o/cri-o/internal/lib"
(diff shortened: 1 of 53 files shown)


haircommander:
hmhmhm I'm not stoked about the global variable (but sync.Once symantics won't quite work with needing libconfig everytime we call ContainerServer() ). What do you need to call the container server from the sandbox package for?

saschagrunert:
I know, global mutable state is bad. :-/ The Server package calls to the Container Server during sandbox and container creation (and deletion etc.) multiple times. This was fine, since the server had the container server as direct member. The only way to bring an interface in between both is from my perspective the need to decouple them. This seems needed because the new interfaces should never import anything from the server package. 😶

saschagrunert:
Closing for now.

PR #3426, opened 2020-03-20
Switch back to machine executor for CI lint stage

saschagrunert:
#### What type of PR is this?

/kind bug

The container executor was hit by OOM kills because golangci-lint uses a
tremendous amount of memory. We now switch back to the machine executor
to have better memory constraints.
None

None

```release-note
None
```


TomSweeneyRedHat:

PR #3435, opened 2020-03-20
tree_status: show the git diff

haircommander:
#### What type of PR is this?

/kind cleanup

it is often the case that there's some git diff in hack/tree_status.sh in CI, but we don't know what it is. Fix this by printing it out when there is something different

```release-note
none
```


umohnani8:
LGTM
/approve

haircommander:
the reasoning behind these failures is unclear to me
/retest

Issue #3455, opened 2020-03-23
CRI-O: PLEG/ Node NotReady issue due to longer terminationGracePeriodSeconds in pods


tanweernoor:
**Description**
CRIO- PLEG/ Node NotReady issue due to longer terminationGracePeriodSeconds in pods

We met the scenario where we have two containers in a pod, when the pod lands to Termination state, one container goes to Exited state (container terminated) and another one is Running due to the nature of job ( Please see sample POD yaml ). we noticed that the the latter container stays in "Running" till terminatioGracePeriodSeconds reaches. if terminationGracePeriodSeconds is longer than 3 minute then the Node lands into NodReady state and comes back to Ready state when the offended Pod is Terminated by reaching terminatioGracePeriodSeconds. 

**Steps to reproduce the issue:**
Use this sample yaml to reproduce the problem
apiVersion: v1
kind: Pod
metadata:
  name: test-containers
spec:
  volumes:
  - name: shared-data
    emptyDir: {}
  containers:
  - name: nginx-container
    image: nginx:latest
    volumeMounts:
    - name: shared-data
      mountPath: /usr/share/nginx/html
  - name: busybox-container
    image: busybox:1.28
    volumeMounts:
    - name: shared-data
      mountPath: /pod-data
    command: ["/bin/sh", "-ec", "while :; do echo 'Hello from the test container >>/pod-data/index.html'; sleep 5 ; done"]
  terminationGracePeriodSeconds: 600

**Describe the results you received:**
Node lands to NotReady state after 180 seconds of pod delete operation

**Describe the results you expected:**
Termination should be graceful at terminationGracePeriodSeconds

**Additional information you deem important (e.g. issue happens only occasionally):**
1) this is always reproducible with above pod yaml
2) crictl stats hangs on running container, reports empty on Exited container ( that's expected ). 
3) Issue doesn't occur with containerd ( crictl stats doesn't hang during pod termination )
4) Not reproducible with one container and longer terminationGracePeriodSeconds

crictl stats output

crictl ps -a | grep -i 32c46840411bf
fd3bd11541116       busybox@sha256:74f634b1bc1bd74535d5209589734efbd44a25f4e2dc96d78784576a3eb5b335                                                      4 minutes ago        Running             busybox-container        0                   32c46840411bf
[...]

PR #3457, opened 2020-03-23
server/ContainerStatus: don't lock for c.State()

kolyshkin:
c.State() is using opLock which can sometimes be held for minutes
in e.g. runtime.ContainerStop. The opLock is there so that
operations changing the container state won't clash. There
is no reason to wait for the lock here.

The side effect is, ContainerStatus will return immediately,
not e.g. after container has finished starting or stopping,
and show the old, rather than new, state.

#### What type of PR is this?

/kind bug

This might fix https://github.com/cri-o/cri-o/issues/3455

```release-note
ContainerStatus no longer waits for a container operation (such as start or stop) to finish.
```

### DIFF
--- a/server/container_status.go
+++ b/server/container_status.go
@@ -48,7 +48,7 @@ func (s *Server) ContainerStatus(ctx context.Context, req *pb.ContainerStatusReq
 	}
 	resp.Status.Mounts = mounts
 
-	cState := c.State()
+	cState := c.StateNoLock()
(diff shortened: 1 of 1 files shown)


haircommander:
"Specified key doesn't exist"

/retest

haircommander:
LGTM

tedyu:
In Container#FromDisk():
```
	return dec.Decode(c.state)
```
We should take opLock prior to decoding.

@mrunalp @kolyshkin @rphillips 
What do you think ?

PR #3459, opened 2020-03-24
Add hint to release notes on gh-pages to README.md

saschagrunert:
#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it:
We now add an additional badge to the README as well as a custom note
about the release notes.
#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:

None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #3460, opened 2020-03-24
Make release notes generator capable of handling tags

saschagrunert:
#### What type of PR is this?

/kind feature

The release notes generator now sets the download variables correctly on tags.

None
None

```release-note
None
```

### DIFF
--- a/scripts/release-notes/release_notes.go
+++ b/scripts/release-notes/release_notes.go
@@ -82,6 +82,19 @@ func run() error {
 	}
 	defer func() { err = os.RemoveAll(templateFile.Name()) }()
 
+	// Check if we're on a tag and adapt variables if necessary
+	bundleVersion := head[:9]
+	shortHead := head[:7]
+	endRev := head
+	if output, err := command.New(
+		"git", "describe", "--exact-match",
+	).RunSilentSuccessOutput(); err == nil {
+		foundTag := output.OutputTrimNL()
+		bundleVersion = foundTag
+		shortHead = foundTag
+		endRev = foundTag
+	}
+
 	if _, err := templateFile.WriteString(fmt.Sprintf(`# CRI-O %s
 
 The release notes have been generated for the commit range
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Lint isn't happy @saschagrunert

saschagrunert:
I know, this is because of the build cache which I have fixed in another PR.

haircommander:
> Lint isn't happy @saschagrunert

I was able to fix this in https://github.com/cri-o/cri-o/pull/3291 by rerunning the workflow from the beginning

saschagrunert:
> > Lint isn't happy @saschagrunert
> 
> I was able to fix this in #3291 by rerunning the workflow from the beginning

Yeah I just bumped the lint cache to another version which causes the whole project to rebuild and cache again (see https://github.com/cri-o/cri-o/pull/3459/files). If I do it here again I will definitely end-up in a merge conflict.

scripts/release-notes/release_notes.go
```diff
@@ -95,6 +95,16 @@ func run() error {
 		return errors.Wrap(err, "open generated release notes")
 	}
 
+	// Check if we're on a tag and adapt variables if necessary
+	bundleVersion := head[:9]
+	shortHead := head[:7]
+	endRev := head
+	if !strings.Contains(tag, "dev") {
```
haircommander: we aren't using the `-dev` idiom anymore in release branches, will this set end tag for every new PR?

saschagrunert: We do still use it. But if we're exactly on a checked out tag, then the version will not have the `-dev` suffix. This is the indicator for us to link the actual release instead of the commit.

saschagrunert:
@haircommander I think this might be ready to get merged :balloon:

scripts/release-notes/release_notes.go
haircommander: for 1.13-1.17 we don't use the `-dev` suffix for version bumps (ref https://github.com/cri-o/cri-o/pull/3575/commits/a3564c52271d9641099f4477645f2fe6705e57a1). I think we need a different scheme to check if we're precisely on a version bump PR or not.

saschagrunert: I'm not sure If I get this correctly. We use the `-dev` for every commit which is not a tag right now:
https://github.com/cri-o/cri-o/blob/e8d1d22e7a43466627471e1cbb2189286b2a0153/scripts/latest-version/latest_version.go#L129

If you want then we can change this to something different, but in the end the `latest_version.go` has to return something that indicates that we're on a development version. (Maybe the git commit?)

umohnani8:
LGTM
@haircommander PTAL

saschagrunert:
Changed the implementation, PTAL @haircommander

saschagrunert:
@haircommander I rebased, do you think we can get this in?

PR #3461, opened 2020-03-24
Add release branch forward to CircleCI

saschagrunert:
#### What type of PR is this?

/kind feature

We now add the release branch forward script to CircleCI, which will on
on every master branch push.

None
None

```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,6 +126,14 @@ workflows:
       - lint:
           filters: *filterAllTags
 
+      - release-branch-forward:
+          requires:
+            - results
+          filters:
+            branches:
+              only:
+                - master
+
       - release-notes:
           filters: *filterAllTags
 
@@ -382,6 +390,26 @@ jobs:
           command: make lint
       - save_cache:
           key: v4-lint-{{ checksum "go.sum" }}
+          paths:
+            - *gocache
+
+  release-branch-forward:
+    executor: container
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-release-branch-forward-{{ checksum "go.sum" }}
+      - add_ssh_keys:
+          fingerprints:
+            - bd:a6:4b:50:1b:2a:a4:1d:79:67:64:05:66:c4:3a:10
+      - run:
+          name: Forward Latest Release Branch
+          command: make release-branch-forward
+          environment:
+            DRY_RUN: false
+      - save_cache:
+          key: v1-release-branch-forward-{{ checksum "go.sum" }}
           paths:
             - *gocache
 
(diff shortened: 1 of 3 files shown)


saschagrunert:
@haircommander PTAL. The idea is to run the script on every master push, which should be the right trigger in our case.

TomSweeneyRedHat:
Changes LGTM, but some test issues and I'm not a circle heavy.

umohnani8:
LGTM

haircommander:
wait
/hold

.circleci/config.yml
```diff
@@ -126,6 +126,14 @@ workflows:
       - lint:
           filters: *filterAllTags
 
+      - release-branch-forward:
+          requires:
+            - results
```
haircommander, line 131: is results the correct trigger here? I see results happens after all the circle tests pass, not after this PR is merged. if i understand correctly, that means this script would run on every passed CI run, not on every merged PR

.circleci/config.yml
```diff
@@ -126,6 +126,14 @@ workflows:
...
+          filters:
+            branches:
+              only:
+                - master
```
saschagrunert, line 135: This means it runs only on master commits.

.circleci/config.yml
saschagrunert: Yes it means that it runs after the results step. That it only applies to the master branch happens here: https://github.com/cri-o/cri-o/pull/3461#discussion_r404253416

haircommander: right, but results happens after every circle test passes (for instance, it's already run and passed on this PR  https://app.circleci.com/jobs/github/openSUSE/cri-o/42837)

this means we are merging from master at the wrong time. It would run every time circle runs and passes, but not every time a PR is merged. I suppose this means that for every PR merged, it would get backported to the last release when the next PR is submitted. I'm not sure that is what you intended here though.

or am I totally off?

saschagrunert: > right, but results happens after every circle test passes (for instance, it's already run and passed on this PR https://app.circleci.com/jobs/github/openSUSE/cri-o/42837)

The `results` step happens every time, correct. The `requires` only states that `release-branch-forward` runs after the `results` step (from a pipeline perspective).

> this means we are merging from master at the wrong time. It would run every time circle runs and passes, but not every time a PR is merged.

No, the filter `filters branches only master` below let this step run only on the master branch, not on PRs. For example it did not run on this PR, which is correct. But it will run if prow merges the changes into the master. :)

Did I miss anything?

saschagrunert: @haircommander what do you think?

haircommander:
Okay, let's give it a shot :+1: 
/lgtm

haircommander:
(before this merges, 1.18 is on f67fa53236a2e895e80a900f843b2c65705d0dfa), after this merges, it should update to current master

saschagrunert:
Let’s release the hold :)
/hold cancel

saschagrunert:
It works. :)

PR #3474, opened 2020-03-24
WIP Sandbox: drop infra and use pinns to configure sysctls

haircommander:
#### What type of PR is this?

/kind api-change

a new version of https://github.com/cri-o/cri-o/pull/3076 that adapts the sysctl code from crun and copies it into pinns, as it's the only part of libcrun that is needed. This, along with fixing a remaining port-forward bug, should be the last needed to pass tests while dropping infra container (still lots of more refactoring needed)

```release-note
If manage_ns_lifecycle is true, and pod isn't sharing PID namespace, pods will be deployed without an infra container
```


test/sysctl.bats
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bats
```
haircommander: todo: remove this test, it was just for not needing to run the entire ctr.bats

pinns/src/sysctl.c
```diff
@@ -0,0 +1,82 @@
...
+  }
+  if (!strlen (sysctl_value))
+  {
+	pwarnf ("sysctl must be in the form of 'key=value'; key is value");
+    return -1;
+  }
+  return 0;
+}
+
+static int write_sysctl_to_file (char * sysctl_key, char* sysctl_value)
```
haircommander: todo: give @giuseppe credit for this function

giuseppe: could we use directly libcrun or is it overkill?  It doesn't need to be linked and it could just be a git submodule

haircommander: It is kind of overkill. All the autoconf/automake configuration for a 15 line function ultimately doesn't seem worth it.

The interaction with libcrun was also awkward, because we need to call pinns before we write the `config.json` to disk. To get the sysctls from it, we needed to write a temporary one, then use it to configure the namespaces, then use those namespaces to configure the real one.

Ultimately, I like this flow better.

giuseppe: I thought we could use directly the function without going through the `config.json` part.

But I agree autoconf/automake for a function is overkill :-)

internal/lib/container_server.go
```diff
@@ -275,37 +277,43 @@ func (c *ContainerServer) LoadSandbox(id string) error {
...
 	}
 
-	scontainer, err := oci.NewContainer(m.Annotations[annotations.ContainerID], cname, sandboxPath, m.Annotations[annotations.LogPath], labels, m.Annotations, kubeAnnotations, "", "", "", nil, id, false, false, false, privileged, sb.RuntimeHandler(), sandboxDir, created, m.Annotations["org.opencontainers.image.stopSignal"])
-	if err != nil {
-		return err
-	}
-	scontainer.SetSpec(&m)
-	scontainer.SetMountPoint(m.Annotations[annotations.MountPoint])
+	var scontainer *oci.Container
+	if sb.NeedsInfra(c.config.ManageNSLifecycle) {
```
tedyu, line 281: When NeedsInfra() returns true, infra container would still be instantiated.
From the subject of the PR, I thought infra container would be gone.

haircommander: yeah it's still being used to hold pod level information (and is likely needed to store it for a restore). The first step is to drop the container at the runtime level

server/sandbox_remove.go
```diff
@@ -101,11 +101,9 @@ func (s *Server) RemovePodSandbox(ctx context.Context, req *pb.RemovePodSandboxR
 		}
 	}
 
-	if podInfraContainer != nil {
-		s.ReleaseContainerName(podInfraContainer.Name())
-		if err := s.CtrIDIndex().Delete(podInfraContainer.ID()); err != nil {
-			return nil, fmt.Errorf("failed to delete infra container %s in pod sandbox %s from index: %v", podInfraContainer.ID(), sb.ID(), err)
-		}
+	s.ReleaseContainerName(podInfraContainer.Name())
```
tedyu, line 104: We don't need the 'if podInfraContainer != nil ' check anymore ?

server/sandbox_run_linux.go
```diff
@@ -552,70 +570,56 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
...
+		}()
+
+		if err := s.ContainerStateToDisk(container); err != nil {
+			log.Warnf(ctx, "unable to write containers %s state to disk: %v", container.ID(), err)
 		}
-	}()
 
-	if err := s.ContainerStateToDisk(container); err != nil {
-		log.Warnf(ctx, "unable to write containers %s state to disk: %v", container.ID(), err)
+		if err := s.MonitorConmon(container); err != nil {
```
tedyu, line 620: I noticed the MonitorConmon() call originally on line 638 has been removed.
Do we still want the call here ?

internal/lib/sandbox/namespaces.go
```diff
@@ -67,21 +67,21 @@ func (m *ManagedNamespace) Path() string {
...
 	typesAndPaths := make([]*ManagedNamespace, 0, 4)
 	if len(managedNamespaces) == 0 {
 		return typesAndPaths, nil
 	}
 
-	namespaces, err := pinFunc(managedNamespaces, cfg)
+	namespaces, err := pinFunc(managedNamespaces, cfg, sysctls)
 	if err != nil {
-		return typesAndPaths, nil
+		return typesAndPaths, err
```
tedyu, line 84: Since err is returned, how useful is typesAndPaths ?

pinns/src/pinns.c
```diff
@@ -13,28 +14,9 @@
...
-
-  snprintf(ns_path, PATH_MAX - 1, "/proc/self/ns/%s", ns_name);
-  if (mount(ns_path, bind_path, NULL, MS_BIND, NULL) < 0) {
-    pwarnf("Failed to bind mount ns: %s", ns_path);
-    return -1;
-  }
-
-  return 0;
-}
+static int bind_ns(const char *pin_path, const char *ns_name);
```
tedyu, line 19: Current declaration is:
static int bind_ns(const char *pin_path, const char *filename, const char *ns_name);

filename is no longer needed ?

internal/lib/sandbox/namespaces.go
haircommander: it's not it could be nil

server/sandbox_run_linux.go
haircommander: no we've dropped conmonmon

pinns/src/pinns.c
haircommander: the way the function works has been changed

haircommander:
replacing with https://github.com/cri-o/cri-o/pull/3997

PR #3483, opened 2020-03-26
test: set cri-o systemd restart policy to no

haircommander:
#### What type of PR is this?

/kind failing-test

While in production we want cri-o to restart when it crashes, when we're testing we'd like to catch panics and crashes. Set restart policy in the systemd unit file to never restart, so we fail tests that make cri-o panic

also drop make install.{systemd,config} when building cri-o, as the targets are already hit when running make install

I consider this a failing test because we are failing to catch when CRI-O crashes

```release-note
none
```

### DIFF
--- a/contrib/test/integration/build/cri-o.yml
+++ b/contrib/test/integration/build/cri-o.yml
@@ -20,25 +20,15 @@
     dest: "/bin/pinns"
     state: link
 
-- name: install cri-o
-  make:
-    target: install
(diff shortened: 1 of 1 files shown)


haircommander:
```

TASK [install cri-o] ***********************************************************
task path: /go/src/github.com/cri-o/cri-o/contrib/test/integration/build/cri-o.yml:29
Killed by signal 15.
++ export status=FAILURE
++ status=FAILURE
```
???
/retest

mrunalp:
@haircommander This is great. Fail fast in CI :+1:

PR #3487, opened 2020-03-27
Add crio config --migrate feature

saschagrunert:
#### What type of PR is this?

/kind feature

We now offer a new option `crio config -m/--migrate` to migrate the current provided config to the latest version. The feature also logs which changes will be done to the config before printing the result:

```console
> sudo ./bin/crio config -m 1.17
INFO Migrating config from 1.17
INFO Checking for NET_RAW and SYS_CHROOT capabilities, which have been removed per default
INFO Removing "default_capabilities" entry "NET_RAW"
INFO Removing "default_capabilities" entry "SYS_CHROOT"
INFO Checking for default AppArmor profile, which does not contain the version number any more
INFO Checking for the log level, which has changed from error to info
INFO Checking for ctr_stop_timeout, which now has a minimum value of 30
INFO Changing "ctr_stop_timeout" to 30
# The CRI-O configuration file specifies all of the available configuration
# options and command-line flags for the crio(8) OCI Kubernetes Container Runtime
# daemon, but in a TOML format that can be more easily modified and versioned.
#
…
```

None

I will add tests if we agree that this is useful for users.

```release-note
- Add `crio config -m/--migrate` option which supports migrating a v1.17.0 configuration file to the latest version.
```

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -1,12 +1,18 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
+	"github.com/cri-o/cri-o/internal/config/migrate"
 	"github.com/cri-o/cri-o/internal/criocli"
 	"github.com/cri-o/cri-o/pkg/config"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 )
+
+var from string
 
 var configCommand = &cli.Command{
 	Name: "config",
@@ -18,8 +24,31 @@ it later with **--config**. Global options will modify the output.`,
 			Name:  "default",
 			Usage: "Output the default configuration (without taking into account any configuration options).",
 		},
+		&cli.StringFlag{
+			Name:        "migrate-defaults",
+			Aliases:     []string{"m"},
+			Destination: &from,
+			Usage: fmt.Sprintf(`Migrate the default config from a specified version.
+    To run a config migration, just select the input config via the global
+    '--config,-c' command line argument, for example:
+    `+"```"+`
+    crio -c /etc/crio/crio.conf.d/00-default.conf config -m 1.17
+    `+"```"+`
+    The migration will print converted configuration options to stderr and will
+    output the resulting configuration to stdout.
+    Please note that the migration will overwrite any fields that have changed
+    defaults between versions. To save a custom configuration change, it should
+    be in a drop-in configuration file instead.
+    Possible values: %q`, migrate.From1_17),
+			Value: migrate.FromPrevious,
+		},
 	},
 	Action: func(c *cli.Context) error {
+		logrus.SetFormatter(&logrus.TextFormatter{
+			DisableTimestamp: true,
+		})
+		logrus.SetLevel(logrus.InfoLevel)
+
 		conf, err := criocli.GetConfigFromContext(c)
 		if err != nil {
 			return err
@@ -29,6 +58,13 @@ it later with **--config**. Global options will modify the output.`,
 			conf, err = config.DefaultConfig()
 			if err != nil {
 				return err
+			}
+		}
+
+		if c.IsSet("migrate-defaults") {
+			logrus.Infof("Migrating config from %s", from)
+			if err := migrate.Config(conf, from); err != nil {
+				return errors.Wrap(err, "migrate config")
 			}
 		}
 
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -313,6 +313,19 @@ it later with **--config**. Global options will modify the output.
 
 **--default**: Output the default configuration (without taking into account any configuration options).
 
+**--migrate-defaults, -m**="": Migrate the default config from a specified version.
+    To run a config migration, just select the input config via the global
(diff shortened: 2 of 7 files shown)


haircommander:
hm often I do
`crio config > config.toml`, but this would add a bunch of debug logs to it. Could make logrus log to stderr for this command only? ([here's](https://github.com/sirupsen/logrus/issues/403) a thread, it's not simple but not impossible)

haircommander:
also this seems more as a `--defaults-migrate`. We would migrate the format by calling `crio config` anyway, but this migrates to new defaults

haircommander:
finally, in the help docs we should mention that this will rewrite the whole config, and that users should move their user changed config values to drop-in files if they don't want to lose potentially their values

saschagrunert:
> hm often I do
> `crio config > config.toml`, but this would add a bunch of debug logs to it. Could make logrus log to stderr for this command only? ([here's](https://github.com/sirupsen/logrus/issues/403) a thread, it's not simple but not impossible)

I think logrus already logs to stderr, so this scenario should be fine.

saschagrunert:
> also this seems more as a `--defaults-migrate`. We would migrate the format by calling `crio config` anyway, but this migrates to new defaults

We could also migrate configs like this, which already works:

```
crio --config old.conf config --migrate v1.17 > new.conf
```

It does not only migrate to defaults. For example lets take the capabilities: All custom specified capabilities will be still in afterwards, whereas we only remove the NET_RAW and SYS_CHROOT ones.

saschagrunert:
> finally, in the help docs we should mention that this will rewrite the whole config, and that users should move their user changed config values to drop-in files if they don't want to lose potentially their values

Sure, I will write a explicit documentation if we think we should continue here.

haircommander:
> > also this seems more as a `--defaults-migrate`. We would migrate the format by calling `crio config` anyway, but this migrates to new defaults
> 
> We could also migrate configs like this, which already works:
> 
> ```
> crio --config old.conf config --migrate v1.17 > new.conf
> ```
> 
> It does not only migrate to defaults. For example lets take the capabilities: All custom specified capabilities will be still in afterwards, whereas we only remove the NET_RAW and SYS_CHROOT ones.

right, but that's a default change. It is also migrating the config structure, but that's happening as part of the `crio config` without the `--migrate`

it's a pedantic difference, and I'm not that invested in the change, but I think we should be clear that this flag is going to change the defaults from the 1.17 defaults to 1.18 defaults.

saschagrunert:
Updated the PR with documentation and testing. PTAL

cmd/crio/config.go
```diff
@@ -18,8 +23,29 @@ it later with **--config**. Global options will modify the output.`,
...
+			Usage: `Migrate the default config from a certain version.
+    To run a config migration, just select the input config via the global
+    '--config,-c' command line argument, for example:
+
+    crio -c /etc/crio/crio.conf.d/00-default.conf config -m 1.17
+
+    The migration will print converted configuration options to stderr and will
+    output the resulting configuration to stdout. Please note that the migration
+    will only convert changed defaults between CRI-O versions by preserving
+    custom settings in the configuration.`,
```
haircommander: suggested edit:

`Please note that the migration will overwrite any fields that have changed defaults between versions. To save a custom configuration change, it should be in a drop-in configuration file instead.`

haircommander:
one nit, otherwise LGTM

cmd/crio/config.go
saschagrunert: Fixed :+1:

docs/crio.8.md
```diff
@@ -313,6 +313,18 @@ it later with **--config**. Global options will modify the output.
 
 **--default**: Output the default configuration (without taking into account any configuration options).
 
+**--migrate-defaults, -m**="": Migrate the default config from a certain version.
```
TomSweeneyRedHat: suggest:  certain version -> specified version

docs/crio.8.md
```diff
@@ -313,6 +313,18 @@ it later with **--config**. Global options will modify the output.
...
+    '--config,-c' command line argument, for example:
+    ```
+    crio -c /etc/crio/crio.conf.d/00-default.conf config -m 1.17
+    ```
+    The migration will print converted configuration options to stderr and will
+    output the resulting configuration to stdout.
+    Please note that the migration will overwrite any fields that have changed
+    defaults between versions. To save a custom configuration change, it should
+    be in a drop-in configuration file instead. (default: 1.17)
+
```
TomSweeneyRedHat: It might be nice to have an option that lists the valid values or point the user to a file where they could see the valid values.  Not such an issue now, but 5 or 6 versions from now in the future.  Something to consider, not necessary for this PR.

saschagrunert: I added the possible values so that we do not for get it in the future.

docs/crio.8.md
saschagrunert: Fixed as suggested :+1:

saschagrunert:
/retest

@haircommander I think this might be ready to be merged if we want to have it

PR #3488, opened 2020-03-28
build: clean generated bin/ dir

dougsland:
Previously, make clean command was not cleaning
bin/crio-status and bin/ dir. This patch make sure
the generated bin/ is cleaned.

#### What type of PR is this?

/kind bug

Fix make clean


saschagrunert:
Impressive how fast this succeeded. 😱

PR #3492, opened 2020-03-30
Use crictl for crio wipe

saschagrunert:
#### What type of PR is this?

/kind feature

We encountered multiple issues with crio wipe, for example that the
containers storage got not properly unmounted or processes are still
alive. To avoid such pitfalls, we can now utilize crictl to remove all
pods (including containers) and images via two simple commands. This now
requires to have a working crio instance running before executing `crio
wipe`.

None
@haircommander would this work for OpenShift too?

```release-note
- Changed `crio wipe` to use `crictl` for container and image removal. This means that we now
  require `crictl` to be available in `$PATH` as well as changing the execution order of the
  `crio-wipe.service` systemd unit.
```

### DIFF
--- a/cmd/crio/wipe.go
+++ b/cmd/crio/wipe.go
@@ -1,18 +1,18 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
-	"os"
 
-	cstorage "github.com/containers/storage"
 	"github.com/cri-o/cri-o/internal/criocli"
-	"github.com/cri-o/cri-o/internal/storage"
 	"github.com/cri-o/cri-o/internal/version"
+	"github.com/cri-o/cri-o/pkg/config"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
+	"k8s.io/release/pkg/command"
 )
+
+const crictl = "crictl"
 
 var wipeCommand = &cli.Command{
 	Name:   "wipe",
@@ -28,7 +28,7 @@ var wipeCommand = &cli.Command{
 }
 
 func crioWipe(c *cli.Context) error {
-	config, err := criocli.GetConfigFromContext(c)
+	cfg, err := criocli.GetConfigFromContext(c)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


haircommander:
I'd have to check the execution order to make sure crio-wipe.service and kubelet startup don't race in openshift. I'm hesitant but I can see this working

saschagrunert:
> I'd have to check the execution order to make sure crio-wipe.service and kubelet startup don't race in openshift. I'm hesitant but I can see this working

Yes this would be cool. I think we could avoid many issues by just going via the gRPC API instead of manipulating containers/storage directly :+1:

evrardjp:
Any news?

haircommander:
> Any news?

none yet, sorry. I'll prioritize testing this today

evrardjp:
Sorry, I didn't want to pressure you! I am just making sure this is not forgotten. : )

haircommander:
so I have been running into some problems with this. note, this is by hammering `systemctl restart crio kubelet`, so it's not entirely realistic.

once I got `identifier is not a container`
another couple of times we get `connection refused` errors. I am guessing because systemd starts crio-wipe before the socket is actually activated *sometimes*

I also wonder what the motivation behind this PR is. I think it'd be better we dropped crio requiring crio wipe, and allowed sysadmins to set it themselves. I have not heard of the errors you describe in openshift in a while (where crio is typically restarted after reboot).

@umohnani8 did find a problem with 1.18 and storage (where we'd sometimes panic), but that should be fixed once https://github.com/cri-o/cri-o/pull/3500 lands

saschagrunert:
> I also wonder what the motivation behind this PR is. I think it'd be better we dropped crio requiring crio wipe, and allowed sysadmins to set it themselves. I have not heard of the errors you describe in openshift in a while (where crio is typically restarted after reboot).
> 
> @umohnani8 did find a problem with 1.18 and storage (where we'd sometimes panic), but that should be fixed once #3500 lands

We found a runtime panic where sandboxes are not restored correctly when upgrading between 1.16 and the latest 1.17 release (report is incoming). 

I agree that removing the whole wiping story would be more useful but to do that, we would test those upgrade scenarios otherwise we have no chance to guarantee anything.

I'll close this for now, let's revisit if we have the error report.

PR #3493, opened 2020-03-30
Use HEAD for runc built from source

kolyshkin:
This is the master branch and it makes sense to use
the latest runc for CI.

Suggested-by: @mrunalp 

/kind dependency-change

```release-note
NONE
```

### DIFF
--- a/contrib/test/integration/build/runc.yml
+++ b/contrib/test/integration/build/runc.yml
@@ -4,7 +4,7 @@
   git:
     repo: "https://github.com/opencontainers/runc.git"
     dest: "{{ ansible_env.GOPATH }}/src/github.com/opencontainers/runc"
-    version: "13b1603fd0e37db1764433a140d494b2e9f05805"
+    version: "HEAD"
 
 - name: build runc
   make:


contrib/test/integration/build/runc.yml
```diff
@@ -4,7 +4,7 @@
   git:
     repo: "https://github.com/opencontainers/runc.git"
     dest: "{{ ansible_env.GOPATH }}/src/github.com/opencontainers/runc"
-    version: "13b1603fd0e37db1764433a140d494b2e9f05805"
+    version: "HEAD"
```
saschagrunert: Hm, if we make it this dynamic then we might silently break our CI if something in runc changes.

mrunalp: @saschagrunert That is kind of the intent here. We are merging a lot into runc recently and I fear we don't have enough coverage there so we want to catch any issues fast just like we are doing with Kubernetes master with our master branch. Otherwise we will end up sifting through 50+ runc commits vs. raising an early alarm and having to look at hopefully just -2-3 commits per day.

kolyshkin: That's the idea -- it's a master not a release branch, and we'd rather catch runc bugs earlier and fix those quick.

kolyshkin: But yeah, I agree this has pros and cons. The biggest "pro" argument is we will catch the runc problem (or the problem in runc+crio) before a release, not after. The biggest "con" is we will break CI without changing anything in crio, which is counterintuitive.

saschagrunert: Alright thank you for the clarification.

PR #3497, opened 2020-03-31

haircommander:
#### What type of PR is this?

/kind bug

Per https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-limits-are-run, when kubernetes runs a container with a memory usage limit set, it expects results comparable to using docker run --memory=....

Per https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details, if the --memory flag is used, but the --memory-swap flag is not used, the swap space limit value is expected to be equal to the memory limit value.

This swap space limit includes the memory limit, though, so kubernetes effectively expects us to not allow the container to use swap.

This PR also checks to make sure we have cgroup memory swap enabled. If we don't, we don't set this value, as it will error in the runtime

this is carrying https://github.com/cri-o/cri-o/pull/3349, but adds the check for memsw

```release-note
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -506,11 +506,12 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 			specgen.SetLinuxResourcesCPUShares(uint64(resources.GetCpuShares()))
 
 			memoryLimit := resources.GetMemoryLimitInBytes()
-			if memoryLimit != 0 && memoryLimit < minMemoryLimit {
-				return nil, fmt.Errorf("set memory limit %v too low; should be at least %v", memoryLimit, minMemoryLimit)
-			}
 			if memoryLimit != 0 {
+				if memoryLimit < minMemoryLimit {
+					return nil, fmt.Errorf("set memory limit %v too low; should be at least %v", memoryLimit, minMemoryLimit)
+				}
 				specgen.SetLinuxResourcesMemoryLimit(memoryLimit)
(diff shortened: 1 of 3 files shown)


haircommander:
this is actually part 1 of a 2 part fix. the second part involves refactoring cgroup use. I figure we can have this to easily cherry pick (it was asked for in 1.16), then follow on with a bigger refactor in a second PR

haircommander:
/test integration_crun

build failures again

haircommander:
I am unsure what happened with [this](https://prow.svc.ci.openshift.org/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3497/test_pull_request_crio_e2e_features_fedora/8970/) run, but it doesn't seem to be related

/retest

haircommander:
integration_crun keeps only showing started.json, and no other files. impossible to tell what is happening

/retest

:crossed_fingers:

haircommander:
```
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/gmake", "msg": "/bin/sh: m: command not found\ngmake[2]: [Makefile:2263: src/libcrun/signals.c] Error 127 (ignored)\n/usr/bin/ld: src/crun-kill.o:
```
so weird that only sometimes happens

/test integration_crun

haircommander:
I'm retrying in a new PR with the hopes that this one is just cursed and that's why crun fails each time with literally no data about why

PR #3506, opened 2020-04-01
Add the mounts that are required by systemd

umohnani8:
#### What type of PR is this?

 /kind bug

None

Signed-off-by: Urvashi Mohnani umohnani@redhat.com

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -665,6 +665,22 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		specgen.AddMount(cgroupMnt)
 	}
 
+	containerImageConfig := containerInfo.Config
+	if containerImageConfig == nil {
+		err = fmt.Errorf("empty image config for %s", image)
+		return nil, err
+	}
+
+	processArgs, err := buildOCIProcessArgs(ctx, containerConfig, containerImageConfig)
+	if err != nil {
+		return nil, err
+	}
+	specgen.SetProcessArgs(processArgs)
+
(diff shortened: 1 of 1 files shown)


server/container_create_linux.go
```diff
@@ -676,6 +676,30 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 		specgen.AddAnnotation(fmt.Sprintf("%s.%d", annotations.IP, idx), ip)
 	}
 
+	// systemd expects to have /run, /run/lock and /tmp on tmpfs
+	// It also expects to be able to write to /sys/fs/cgroup/systemd and /var/log/journal
+	if s.config.CgroupManager == oci.SystemdCgroupsManager {
```
mrunalp, line 681: This shouldn't be done unconditionally; key off the command to know that user is intending to runc systemd.

server/container_create_linux.go
```diff
@@ -878,6 +878,36 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
...
+		specgen.AddMount(tmpfsMnt)
+	}
+	systemdMnt := rspec.Mount{
+		Destination: "/sys/fs/cgroup/systemd",
+		Type:        "bind",
+		Source:      "/sys/fs/cgroup/systemd",
+		Options:     []string{"bind", "nodev", "noexec", "nosuid"},
+	}
+	specgen.AddMount(systemdMnt)
+	specgen.AddLinuxMaskedPaths("/sys/fs/cgroup/systemd/release_agent")
```
giuseppe, line 909: this should be done only on cgroup v1

umohnani8: Fixed

server/container_create_linux.go
umohnani8: fixed

PR #3509, opened 2020-04-01
pinns: pin to /var/run/*ns instead of /var/run/crio/ns/*

haircommander:
#### What type of PR is this?

/kind bug

This change is largely needed for third party networking plugins. These plugins often bind mount /var/run/netns and expect to be able to operate on the namespaces of the pods. Unfortunately, by pinning to /var/run/crio/ns*, we lose this capability. Rather than pinning everything to /var/run/netns/ns/uts or something odd, change the way pinns builds the pinned namespace. Now, we bind namespaces to /var/run/$NS_NAMEns/$ID

Also, add a test that verifies some different permutations of managing namespace lifecycle and restore, to make sure changes like this, as well as admins toggling namespace lifecycle management, don't mess up a restore

once https://github.com/cri-o/cri-o/pull/3487 merges, I think the NamespaceDir change is a good candidate.

```release-note
when pinning namespaces, CRI-O now pins to /var/run/$NS_NAMEns/$RAND_ID instead of /var/run/crio/ns/$RAND_ID/$NS_NAME for better compatibility with third party networking plugins
```

### DIFF
--- a/internal/lib/sandbox/namespaces.go
+++ b/internal/lib/sandbox/namespaces.go
@@ -81,7 +81,7 @@ func (s *Sandbox) CreateNamespacesWithFunc(managedNamespaces []NSType, cfg *conf
 
 	namespaces, err := pinFunc(managedNamespaces, cfg)
 	if err != nil {
-		return typesAndPaths, nil
+		return nil, err
 	}
 
 	for _, namespace := range namespaces {
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -232,7 +232,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--metrics-port**="": Port for the metrics endpoint (default: 9090)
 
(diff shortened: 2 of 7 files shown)


haircommander:
we may also want this cherry-picked to 1.17, considering it breaks assumptions third party networking plugins have, and thus could be considered a bug fix.

haircommander:
PTAL @saschagrunert @mrunalp @umohnani8

haircommander:
/test e2e_crun

it hit the github failures

internal/lib/sandbox/namespaces.go
```diff
@@ -81,7 +81,7 @@ func (s *Sandbox) CreateNamespacesWithFunc(managedNamespaces []NSType, cfg *conf
 
 	namespaces, err := pinFunc(managedNamespaces, cfg)
 	if err != nil {
-		return typesAndPaths, nil
+		return typesAndPaths, err
```
saschagrunert, line 84:
```suggestion
		return nil, err
```

haircommander: fixed!

haircommander:
/retest

error is documented https://github.com/cri-o/cri-o/issues/3527

PR #3515, opened 2020-04-02
Skip already uploaded artifacts with gsutil

saschagrunert:
#### What type of PR is this?

/kind failing-test

It might be possible that we re-run the CI and the artifact already
exists. In that case we will not fail the `results` step.

None
None

```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -428,7 +428,7 @@ jobs:
               gcloud auth activate-service-account --key-file=/tmp/key.json
 
               BUCKET=gs://k8s-conform-cri-o
-              gsutil cp build/bundle/*.tar.gz $BUCKET/artifacts
+              gsutil cp -n build/bundle/*.tar.gz $BUCKET/artifacts
             fi
       - store_test_results:
           path: build/junit


dougsland:
@saschagrunert looks good to me but I am wondering if we should clean the artifacts every-time we re-run the job instead of use cp -n?

saschagrunert:
> @saschagrunert looks good to me but I am wondering if we should clean the artifacts every-time we re-run the job instead of use cp -n? Might generate a false positive?

The retention strategy of the bucket does not allow removing or modifying files for 10 years. 😅 At least they’re safe.

saschagrunert:
> LGTM if you can get that last pesky test to behave.

This test is not marked as `required` and failing recently. I think we can ignore it in this change. 🤷🏻‍♂️

dougsland:
> > @saschagrunert looks good to me but I am wondering if we should clean the artifacts every-time we re-run the job instead of use cp -n? Might generate a false positive?
> 
> The retention strategy of the bucket does not allow removing or modifying files for 10 years.  At
>  least they’re safe.

Ah, I see. /me nods

PR #3518, opened 2020-04-02
[1.17] Add the mounts that are required by systemd

umohnani8:
#### What type of PR is this?

 /kind bug

```release-note

```

### DIFF
--- a/scripts/circle-setup
+++ b/scripts/circle-setup
@@ -110,7 +110,7 @@ install_cni_plugins() {
 
 install_runc() {
     export GOPATH=/home/circleci/go
-    go get github.com/opencontainers/runc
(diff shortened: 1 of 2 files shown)


dougsland:
I am new here, but I am wondering if setupSystemd() and mountExists() could be a good candidates for utils module (or even a systemd module for systemd related code) and not pinned for container_create_linux.go.

rhatdan:
Is there a way we could just share the code from libpod here. Basically make a libpod/pkg/systemd.

umohnani8:
@rhatdan yup I can do that. But can we get this in first, and I will work on sharing the code from libpod in a follow up PR.

haircommander:
I agree we could share with libpod, but no reason to hold this for that

PR #3519, opened 2020-04-03
[1.16] Add the mounts that are required by systemd

umohnani8:
#### What type of PR is this?

 /kind bug

```release-note

```


dougsland:
I am new here, but, I am wondering if setupSystemd() and mountExists() could be a good candidates for utils module (or even a systemd module for systemd related code) and not pinned for container_create_linux.go.

TomSweeneyRedHat:
Changes LGTM, but definitely want someone more systemd knowledgable to take a looksee.

Issue #3528, opened 2020-04-04
Can't debug cri-o static using delve


gattytto:
I recently started building cri-o from source inside a debian LXC Container with NIX as explained in the docs.

now I wanted to add debug symbols to the crio static binary to be able to debug it using delve, but running the command "dlv exec crio-static" triggers the following:

````bash
 dlv --api-version 2 exec /usr/bin/crio --log                                          
2020-04-04T03:40:29-03:00 info layer=debugger launching process with args: [/usr/bin/crio]
could not launch process: could not open debug info
````
then it starts shouting crio service startup logs to the terminal as a normal crio startup process.

I tried to modify default.nix file for building cri-o with debug symbols but it was easier to do it in the Makefile. After some googling I found using the "file" command along with " objdump --syms crio-static-binary" would help me figure if the binary was built really including the symbols.

after some attempts I got a ~65MB static binary with lots of debug symbols but dlv still goes with the "could not open debug info". 

here is the resulting build command from the nix build -f nix -L:
````verilog
nix build -f nix -L                                                                       SIGINT(2) ↵  ⚡  254  03:07:19
cri-o-static> unpacking sources
cri-o-static> unpacking source archive /nix/store/dm6fajmx9lix4iclvzw1pac8qa4v0ljn-cri-o
cri-o-static> source root is cri-o
cri-o-static> patching sources
cri-o-static> configuring
cri-o-static> building
cri-o-static> /build/go/src/github.com/cri-o/cri-o /build
cri-o-static> mkdir -p "/build/go:/nix/store/nha5hv6zzdixkdp6icz8319ylk0ii60n-go-1.14/share/go:/nix/store/nha5hv6zzdixkdp6icz8319ylk0ii60n-go-1.14/share/go/src/github.com/cri-o"
cri-o-static> ln -s "/build/go/src/github.com/cri-o/cri-o" "/build/go:/nix/store/nha5hv6zzdixkdp6icz8319ylk0ii60n-go-1.14/share/go:/nix/store/nha5hv6zzdixkdp6icz8319ylk0ii60n-go-1.14/share/go/src/github.com/cri-o/cri-o"
cri-o-static> touch "/build/go:/nix/store/nha5hv6zzdixkdp6icz8319ylk0ii60n-go-1.14/share/go:/nix/store/nha5hv6zzdixkdp6icz8319ylk0ii60n-go-1.14/share/go/.gopathok"
[...]

PR #3529, opened 2020-04-06
Enable debug symbols for binaries when make DEBUG=1

saschagrunert:
#### What type of PR is this?

/kind feature

Setting DEBUG=1 during build will now result in a binary containing the
debug symbols. The same applies to the nix build when setting "DEBUG=1"
in `nix/default.nix`.

Fixes https://github.com/cri-o/cri-o/issues/3528
None

```release-note
None
```

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -38,6 +38,8 @@ let
       buildInputs = old.buildInputs ++ (with pkgs; [ systemd ]);
       src = ./..;
       EXTRA_LDFLAGS = ''-linkmode external -extldflags "-static -lm"'';
+      dontStrip = true;
+      # DEBUG = 1; # Uncomment this line to enable debug symbols in the binary
     })).override {
       flavor = "-static";
       gpgme = (static pkgs.gpgme);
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM once the tests are happy

nix/default.nix
```diff
@@ -38,6 +38,7 @@ let
       buildInputs = old.buildInputs ++ (with pkgs; [ systemd ]);
       src = ./..;
       EXTRA_LDFLAGS = ''-linkmode external -extldflags "-static -lm"'';
+      # DEBUG = 1; # to enable debug symbols inside the binary
```
haircommander, line 41: what's the purpose of this comment? describing what needs to be done to set debugging symbols?

saschagrunert: Yes exactly. :)

haircommander: I would suggest the rewrite 
```
# DEBUG = 1; # Uncomment this line to enable debug symbols in the binary 
```
but it's a nit. I'm fine with a follow up if you want

saschagrunert: Fixed as suggested :+1:

saschagrunert:
/test kata-containers
connection timeout to azure

PR #3530, opened 2020-04-06
[1.17] pinns: pin to /var/run/*ns instead of /var/run/crio/ns/* 

haircommander:
#### What type of PR is this?

/kind api-change

This change is largely needed for third party networking plugins. These plugins often bind mount /var/run/netns and expect to be able to operate on the namespaces of the pods. Unfortunately, by pinning to /var/run/crio/ns*, we lose this capability. Rather than pinning everything to /var/run/netns/ns/uts or something odd, change the way pinns builds the pinned namespace. Now, we bind namespaces to /var/run/$NS_NAMEns/$ID

Also, add a test that verifies some different permutations of managing namespace lifecycle and restore, to make sure changes like this, as well as admins toggling namespace lifecycle management, don't mess up a restore

once https://github.com/cri-o/cri-o/pull/3487 merges, I think the NamespaceDir change is a good candidate.

```release-note
when pinning namespaces, CRI-O now pins to /var/run/$NS_NAMEns/$RAND_ID instead of /var/run/crio/ns/$RAND_ID/$NS_NAME for better compatibility with third party networking plugins
```


haircommander:
```

TASK [fetch Golang] ************************************************************
task path: /go/src/github.com/cri-o/cri-o/contrib/test/integration/golang.yml:8
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failure downloading https://storage.googleapis.com/golang/go1.13.4.linux-amd64.tar.gz, ('The read operation timed out',)"}
```
uh oh

/retest

:pray:

haircommander:
happy tests, PTAL @saschagrunert @mrunalp @umohnani8 @rhatdan 

I feel confident about this change in a release branch because I have tested changes from restore. Thus, if cri-o is upgraded and restarted, the old pods will use the old location and the new pods will use the new one.

PR #3537, opened 2020-04-07
manage ns: don't cleanup top level directories

haircommander:
#### What type of PR is this?

/kind bug
now that we bind mount to $dir/$ns/$rand instead of $dir/$rand/$ns, we should not cleanup the middle directory, as it will attempt to cleanup the namespace directory for all pods

```release-note

```


evanfoster:
Any chance this could be back-ported to 1.16? We've been hitting an issue that this PR fixes but we're not on 1.17 yet.

PR #3539, opened 2020-04-08
[1.16] server/ContainerStatus: don't lock for c.State()

mrunalp:
c.State() is using opLock which can sometimes be held for minutes
in e.g. runtime.ContainerStop. The opLock is there so that
operations changing the container state won't clash. There
is no reason to wait for the lock here.

The side effect is, ContainerStatus will return immediately,
not e.g. after container has finished starting or stopping,
and show the old, rather than new, state.


mrunalp:
@umohnani8 @kolyshkin PTAL.

PR #3553, opened 2020-04-09
server: skip setting memory limit to 0

giuseppe:
do not specify a memory limit=0 if there are no limits in place.

#### What type of PR is this?

/kind bug

do not set a memory limit=0 if the container doesn't have any limit in place.

OCI specifies -1 to be the value for unlimited memory.

```release-note
NONE
```


haircommander:
LGTM

PR #3564, opened 2020-04-13

haircommander:
#### What type of PR is this?

/kind bug

Per https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-limits-are-run, when kubernetes runs a container with a memory usage limit set, it expects results comparable to using docker run --memory=....

Per https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details, if the --memory flag is used, but the --memory-swap flag is not used, the swap space limit value is expected to be equal to the memory limit value.

This swap space limit includes the memory limit, though, so kubernetes effectively expects us to not allow the container to use swap.

This PR also checks to make sure we have cgroup memory swap enabled. If we don't, we don't set this value, as it will error in the runtime

this is carrying https://github.com/cri-o/cri-o/pull/3349, but adds the check for memsw

```release-note
```

### DIFF
--- a/contrib/test/integration/build/crun.yml
+++ b/contrib/test/integration/build/crun.yml
@@ -5,7 +5,7 @@
     repo: "https://github.com/containers/crun.git"
     dest: "{{ ansible_env.GOPATH }}/src/github.com/containers/crun"
     force: "{{ force_clone | default(False) | bool}}"
-    version: "0.13"
+    version: "4ad44b872b09a244dd7b2848952f4db0d0b968f4"
 
 - name: Install crun dependencies
   raw: >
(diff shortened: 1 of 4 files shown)


haircommander:
blocked on https://github.com/containers/crun/pull/327

TomSweeneyRedHat:
changes LGTM once the dependencies are solved.

haircommander:
I've updated to a newer version of crun. I'd like to get this into 1.18.0, so I have used a commit that is not a official release. @giuseppe advised to do this, as 0.14 will be coming out later. This is not a problem, because it's only for the tests, and is an easy update once 0.14 comes out :smiley:

haircommander:
@umohnani8 @mrunalp @rhatdan PTAL

mrunalp:
We need a follow on issue opened to remove this check and follow up on circle.

PR #3565, opened 2020-04-13
config: add cgroup manager

haircommander:
#### What type of PR is this?

/kind cleanup

previously, there was a lot of clunky code when branching on cgroup manager, as well as duplicated error handling on values that should be initialized from CRI-O startup (are we on a node that supports cgroupsv2? are we on a node that supports pid cgroup?)
    
since cgroup information is initialized and kept in place way before CRI-O starts, we should really be reading disk and branching on cgroup manager once, and then using those values from then on out. This cleans the code and will (barely) improve runtime performance
    
introducing internal/config/cgroupmanager: CRI-O's cgroup manager. It is responsible for initializing the cgroup information for the server, and being a point of contact for all things cgroup related. There are currently three flavors of cgroupmanager: cgroupfs, systemd v1 and systemd v2
This is built upon https://github.com/cri-o/cri-o/pull/3564 and should not be merged until that is

```release-note
pod creation now fails if conmon cannot be moved to the cgroup specified in `conmon_cgroup`. Our default value for `conmon_cgroup` is `system.slice`, which is invalid for cgroupfs. As such, if you use cgroupfs, you should change `conmon_cgroup` to `pod`
```


haircommander:
@mrunalp it does not seem possible to set kernel args for circle, and it does not seem to be possible to activate the memsw cgroup via sysctl. Thus, we need to drop the fatal error if memsw cgroup isn't activated. I think we are okay veering from docker here, as we've existed without memsw cgroup for the entirety of cri-o's life. WDYT? I'd like to get this in to start using the `internal/config/node` package in other new and exciting ways

haircommander:
PTAL @saschagrunert @umohnani8 @rhatdan

tedyu:
In crio.service for cgroupv2 test run, I saw:
```
May 08 14:58:40 ip-172-18-12-106.ec2.internal crio[58199]: time="2020-05-08 14:58:40.869552993Z" level=error msg="Container creation error: error stat'ing file `/var/lib/kubelet/pods/0776a09e-0ec1-4682-810f-d90f99208fdb/volumes/kubernetes.io~secret/default-token-j4bsm`: No such file or directory\n"
```
Such error occurred twice.
Still looking for other error(s).

haircommander:
> In crio.service for cgroupv2 test run, I saw:
> 
> ```
> May 08 14:58:40 ip-172-18-12-106.ec2.internal crio[58199]: time="2020-05-08 14:58:40.869552993Z" level=error msg="Container creation error: error stat'ing file `/var/lib/kubelet/pods/0776a09e-0ec1-4682-810f-d90f99208fdb/volumes/kubernetes.io~secret/default-token-j4bsm`: No such file or directory\n"
> ```
> 
> Such error occurred twice.
> Still looking for other error(s).

I think I found one source of cgroupv2 errors

internal/config/cgroupmanager/cgroupmanager.go
```diff
@@ -0,0 +1,69 @@
...
+	scopePrefix                      = "crio"
+	// minMemoryLimit is the minimum memory that must be set for a container.
+	// A lower value would result in the container failing to start.
+	minMemoryLimit = 12582912
+	// CgroupfsCgroupManager represents cgroupfs native cgroup manager
+	CgroupfsCgroupManager = "cgroupfs"
+	// SystemdCgroupManager represents systemd native cgroup manager
+	SystemdCgroupManager = "systemd"
+
+	DefaultCgroupManager = SystemdCgroupManager
```
saschagrunert, line 22: Can we make those consts private and only the struct types public? The consts should be accessible via the `Name()` method, right?

internal/config/cgroupmanager/cgroupfsmanager.go
```diff
@@ -0,0 +1,84 @@
...
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/containers/libpod/pkg/cgroups"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+)
+
+type CgroupfsManager struct{}
```
saschagrunert, line 16: Non blocking and leave it as you prefer: We could make it a string type alias if it has no members and just a name

internal/config/cgroupmanager/cgroupfsmanager.go
```diff
@@ -0,0 +1,84 @@
...
+		return "", "", fmt.Errorf("cri-o configured with cgroupfs cgroup manager, but received systemd slice as parent: %s", sbParent)
+	}
+	return sbParent, filepath.Join(sbParent, scopePrefix+"-"+sbID), nil
+}
+
+// MoveConmonToCgroup takes the container ID, cgroup parent, conmon's cgroup (from the config) and conmon's PID
+// It attempts to move conmon to the correct cgroup.
+// It returns the cgroupfs parent that conmon was put into
+// so that CRI-O can clean the cgroup path of the newly added conmon once the process terminates (systemd handles this for us)
+func (*CgroupfsManager) MoveConmonToCgroup(cid, cgroupParent, conmonCgroup string, pid int) string {
```
saschagrunert, line 56: I'm wondering why we do not return errors in this method and let the caller decide how to handle it.

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return cgParent, cgPath, nil
+}
+
+func verifyCgroupHasEnoughMemory(cgroupParent, memorySubsystemPath, memoryMaxFilename string) error {
+	// check memory limit is greater than the minimum memory limit of 4Mb
+	// expand the cgroup slice path
+	slicePath, err := systemd.ExpandSlice(cgroupParent)
+	if err != nil {
+		return errors.Wrapf(err, "error expanding systemd slice path for %q", cgroupParent)
```
saschagrunert, line 114: Personally I think repeating `error` is too noisy in our messages. Logrus already indicates that this is an error, so:
```suggestion
		return errors.Wrapf(err, "expanding systemd slice path for %q", cgroupParent)
```

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,182 @@
...
+	// strip off the newline character and convert it to an int
+	strMemory := strings.TrimRight(string(fileData), "\n")
+	if strMemory != "" && strMemory != "max" {
+		memoryLimit, err := strconv.ParseInt(strMemory, 10, 64)
+		if err != nil {
+			return errors.Wrapf(err, "error converting cgroup memory value from string to int %q", strMemory)
+		}
+		// Compare with the minimum allowed memory limit
+		if err := VerifyMemoryIsEnough(memoryLimit); err != nil {
+			return fmt.Errorf("pod %s", err.Error())
```
saschagrunert, line 135:
```suggestion
			return errors.Errorf("pod %v", err)
```

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,182 @@
...
+	if memoryLimit != 0 && memoryLimit < minMemoryLimit {
+		return fmt.Errorf("set memory limit %v too low; should be at least %v", memoryLimit, minMemoryLimit)
+	}
+	return nil
+}
+
+// convertCgroupFsNameToSystemd converts an expanded cgroupfs name to its systemd name.
+// For example, it will convert test.slice/test-a.slice/test-a-b.slice to become test-a-b.slice
+// NOTE: this is public right now to allow its usage in dockermanager and dockershim, ideally both those
+// code areas could use something from libcontainer if we get this style function upstream.
```
saschagrunert, line 151: Does not look like that this function is public.

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,182 @@
...
+func convertCgroupFsNameToSystemd(cgroupfsName string) string {
+	// TODO: see if libcontainer systemd implementation could use something similar, and if so, move
+	// this function up to that library.  At that time, it would most likely do validation specific to systemd
+	// above and beyond the simple assumption here that the base of the path encodes the hierarchy
+	// per systemd convention.
```
saschagrunert, line 156: Is this TODO still valid?

internal/config/node/cgroups.go
```diff
@@ -0,0 +1,94 @@
...
+	_cgroupHasPid     bool
+	_cgroupHasPidErr  error
+
+	_cgroupHasMemorySwapOnce sync.Once
+	_cgroupHasMemorySwap     bool
+	_cgroupHasMemorySwapErr  error
+
+	_cgroupIsV2Once sync.Once
+	_cgroupIsV2     bool
+	_cgroupIsV2Err  error
```
saschagrunert, line 31: Do we need those underscore variables? I don't think that this is a convention in go

internal/config/node/cgroups.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+	return _cgroupHasHugetlb
+}
+
+// CgroupHasPid returns whether the pid controller is present
+func CgroupHasPid() bool {
+	_cgroupHasPidOnce.Do(func() {
+		_, err := libctrcgroups.FindCgroupMountpoint("", "pids")
+		_cgroupHasPid = err == nil
+		_cgroupHasPidErr = err
```
saschagrunert, line 61: Ref to my comment above: It is hard to distinguish the blank identifier `_` with all those underscored vars.

internal/config/cgroupmanager/systemdmanager.go
haircommander: :shrug:

internal/config/node/cgroups.go
haircommander: I would guess we had them before to make explicitly clear they shouldn't be used anywhere else in the server package. I think we're good to drop though

haircommander:
comments addressed @saschagrunert

internal/config/cgroupmanager/cgroupfsmanager.go
haircommander: I think I like this way better

haircommander:
I also added a commit that makes a failure to move the conmon cgroup cause a failure to create the container

haircommander:
happy tests :smile:

haircommander:
@kolyshkin @giuseppe @mrunalp @umohnani8 @rhatdan PTAL

test/ctr.bats
```diff
@@ -1284,7 +1284,6 @@ function wait_until_exit() {
 	[ "$status" -eq 0 ]
 	[[ "$output" =~ "209715200" ]]
 
-
```
kolyshkin, line 1287: The patch content do not align with a description. Not sure why have this at all

internal/config/cgroupmanager/cgroupfsmanager_unsupported.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+type CgroupManager interface {
+	GetContainerCgroupPath(string, string) string
+	// cgroup parent, cgroup path, error
+	GetSandboxCgroupPath(string, string) string
+}
+
+func InitializeCgroupManager(cgroupManager string) (CgroupManager, error) {
+	return nil, fmt.Errorf("not implemented yet")
```
kolyshkin, line 16: nit: `s/fmt.Errorf/errors.New/` since you do not using any `%`-style formatting

internal/config/cgroupmanager/supported.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return nil
+}
+
+// CgroupHasHugetlb returns whether the hugetlb controller is present
+func CgroupHasHugetlb() bool {
+	_cgroupHasHugetlbOnce.Do(func() {
+		if CgroupIsV2() {
+			if _, err := ioutil.ReadDir("/sys/fs/cgroup/hugetlb"); err != nil {
+				_cgroupHasHugetlbErr = errors.Wrap(err, "readdir /sys/fs/cgroup/hugetlb")
```
kolyshkin: I bet `iotuils.ReadDir` already provides the directory name in the error it returns, so there's no need to wrap it.

internal/config/cgroupmanager/supported.go
```diff
@@ -0,0 +1,123 @@
...
+			return
+		}
+		_, err := os.Stat("/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes")
+		_cgroupHasMemorySwap = err == nil
+	})
+	return _cgroupHasMemorySwap
+}
+
+// CgroupIsV2 returns whether we are using cgroup v2 or v1
+func CgroupIsV2() bool {
```
kolyshkin: Hmm, `libpodcgroups.IsCgroup2UnifiedMode()` already provides the functionality of effectively checking if cgroup is v2 or not, only doing the check once and returning the cached value otherwise. I don't see how this wrapper adds any value to it.

internal/config/cgroupmanager/supported.go
kolyshkin: What ioutils.ReadDir does is opens a directory, read all the entries, calls `stat(2)` on every entry and returns all that to the caller. This is a relatively expensive operation, and is surely an overkill if all you need is to check whether the directory exists.

Calling `os.Stat()` and checking `IsDir()` is probably sufficient here.

kolyshkin: Looks like the logic of v1/v2 is wrong here. For v1, you need to check if `/sys/fs/cgroup/hugetlb` exists and is a directory. For v2, you need to read `/sys/fs/cgroup/cgroup.controllers` and see if it contains `hugetlb`.

internal/config/cgroupmanager/supported.go
```diff
@@ -0,0 +1,123 @@
...
+			_cgroupHasHugetlb = true
+			return
+		}
+
+		controllers, err := ioutil.ReadFile("/sys/fs/cgroup/cgroup.controllers")
+		if err != nil {
+			_cgroupHasHugetlbErr = errors.Wrap(err, "read /sys/fs/cgroup/cgroup.controllers")
+			return
+		}
+		_cgroupHasHugetlb = strings.Contains(string(controllers), "hugetlb")
```
kolyshkin: this can result in false positive if some future kernel will introduce say `hugetlb2`.

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/cri-o/cri-o/utils"
+	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const defaultSystemdParent = "system.slice"
+
+type Systemdv1Manager struct{}
```
kolyshkin, line 22: nit: `s/v1/V1/`

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// GetContainerCgroupPath takes arguments sandbox parent cgroup and container ID and returns
+// the cgroup path for that containerID. If parentCgroup is empty, it
+// uses the default parent system.slice
+func (*Systemdv2Manager) GetContainerCgroupPath(sbParent, containerID string) string {
+	return getContainerCgroupPath(sbParent, containerID)
+}
+
+// systemd cgroups (v1 and v2) have no difference in creating the container cgroup path
```
kolyshkin, line 73: I would create `SystemdManager` with the common stuff and then derive it so you won't have to repeat the same methods for v1 and v2.

internal/config/cgroupmanager/supported.go
haircommander: The added value I see is IsCgroup2UnifiedMode() returns a bool and error. Technically, we should only need to check this error once (on initialization) and handle the error there. After that, we should only do a bool check. I find IsCgroup2UnifiedMode to be awkward because we're compelled to handle the error, even though it'll never change

internal/config/cgroupmanager/cgroupmanager.go
```diff
@@ -0,0 +1,69 @@
...
+// New creates a new CgroupManager with defaults
+func New() CgroupManager {
+	// we can eat the error here because we control what the DefaultCgroupManager is
+	cm, _ := SetCgroupManager(DefaultCgroupManager)
+	return cm
+}
+
+// SetCgroupManager takes a string and branches on it to return
+// the type of cgroup manager configured
+func SetCgroupManager(cgroupManager string) (CgroupManager, error) {
```
kolyshkin, line 57: Not sure why there is a need to have this string to type conversion. Do you want to be able to set the type in a text config, or something like that?

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,183 @@
...
+func (*Systemdv2Manager) GetSandboxCgroupPath(sbParent, sbID string) (string, string, error) {
+	return getSandboxCgroupPathForSystemd(sbParent, sbID, cgroupMemorySubsystemMountPathV2, "memory.max")
+}
+
+func getSandboxCgroupPathForSystemd(sbParent, sbID, memorySubsystemPath, memoryPath string) (string, string, error) {
+	if sbParent == "" {
+		return "", "", nil
+	}
+
+	if len(sbParent) <= 6 || !strings.HasSuffix(path.Base(sbParent), ".slice") {
```
kolyshkin, line 94: 1. I think the optimization of checking len first is redundant, as this is _probably_ the first thing `strings.HasSuffix()` does. Are you trying to save the call to `path.Base()`?

2. Doesn't matter for Linux but strictly speaking it should be `filepath.Base()`.

internal/config/cgroupmanager/cgroupmanager.go
haircommander: yes, we need to translate user defined cgroup_manager to CgroupManager. This is called in `pkg/config/config.go`

internal/config/cgroupmanager/systemdmanager.go
haircommander: ah good idea

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	// read in the memory limit from the memory.limit_in_bytes file
+	fileData, err := ioutil.ReadFile(filepath.Join(memorySubsystemPath, slicePath, memoryMaxFilename))
+	if err != nil {
+		if os.IsNotExist(err) {
+			logrus.Warnf("Failed to find %s for slice: %q", memoryMaxFilename, cgroupParent)
+			return nil
+		} else {
+			return errors.Wrapf(err, "error reading %s file for slice %q", memoryMaxFilename, cgroupParent)
```
kolyshkin: The error from ioutil.ReadFile already contains the file name and the fact that reading failed. Not sure if you add value by wrapping it.

internal/config/cgroupmanager/supported.go
haircommander: @kolyshkin do you have a patch that you think is better? I'm not super inclined update the check to something more complicated (using regexp or something) unless the future of hugetlb2 feels realistic. Even then, considering how slowly the kernel moves, I would imagine we would have plenty of time to cross the bridge once we get there. This line is clear and comprehensive for now, which I prefer, unless you have something better in mind.

internal/config/cgroupmanager/systemdmanager.go
haircommander: :shrug: this was adopted from the old implementation, fixed as suggested

internal/config/cgroupmanager/systemdmanager.go
haircommander: I like having the wrap so there's a unique string to grep against. I dropped the MemoryMaxFilename though, that's certainly redundant

internal/config/cgroupmanager/systemdmanager.go
haircommander: fixed as suggested

internal/config/cgroupmanager/supported.go
haircommander: fixed as suggested

internal/config/cgroupmanager/cgroupfsmanager_unsupported.go
haircommander: fixed!

test/ctr.bats
haircommander: fixed!

haircommander:
comments addressed
@kolyshkin @giuseppe @mrunalp @umohnani8 @rhatdan PTAL

internal/config/cgroupmanager/supported.go
kolyshkin: Realistically, you're right. And yet most of the implementations I've dealt with would do something like
```go
for _, ctr := range strings.Split(string(controllers)) {
  if ctr == "hugetlb" {
    __cgroupHasHugetlb = true
    break
  }
}
```
A simple non-regex solution would be
```go
_cgroupHasHugetlb = strings.Contains(" "+string(controllers)+" ", " hugetlb ")
```

internal/config/cgroupmanager/cgroupfsmanager.go
```diff
@@ -0,0 +1,84 @@
...
+	// Move conmon to specified cgroup
```
kolyshkin, line 58: nit: the comment is redundant

internal/config/cgroupmanager/cgroupfsmanager.go
```diff
@@ -0,0 +1,84 @@
...
+	if conmonCgroup != "pod" && conmonCgroup != "" {
+		logrus.Errorf("conmon cgroup %s invalid for cgroupfs", conmonCgroup)
+		return ""
+	}
+
+	cgroupPath := filepath.Join(cgroupParent, "/crio-conmon-"+cid)
```
kolyshkin, line 63: having both `/` and `filepath.Join` is redundant -- you can drop either one. I would drop `filepath.Join` but either way is fine.

internal/config/cgroupmanager/cgroupmanager.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	cgroupMemorySubsystemMountPathV1 = "/sys/fs/cgroup/memory"
+	cgroupMemorySubsystemMountPathV2 = "/sys/fs/cgroup"
+	scopePrefix                      = "crio"
+	// minMemoryLimit is the minimum memory that must be set for a container.
+	// A lower value would result in the container failing to start.
+	minMemoryLimit = 12582912
```
kolyshkin, line 16: 1. Maybe add a comment that this is an empirical value valid for x86_64.

2. Maybe put it as `12 * 1024 * 1024` for readability.

internal/oci/runtime_oci.go
```diff
@@ -87,12 +87,11 @@ func (r *runtimeOCI) CreateContainer(c *Container, cgroupParent string) (err err
...
 	var args []string
-	if r.config.CgroupManager == SystemdCgroupsManager {
+	if r.config.CgroupManager().IsSystemd() {
 		args = append(args, "-s")
 	}
-	if r.config.CgroupManager == CgroupfsCgroupsManager {
-		args = append(args, "--syslog")
-	}
+
+	args = append(args, "--syslog")
```
kolyshkin: Not sure why it was conditional before and is now it's not.

internal/oci/oci_linux.go
```diff
@@ -201,7 +154,7 @@ func metricsToCtrStats(c *Container, m *cgroups.Metrics) *ContainerStats {
 // not parsable.
 func getTotalInactiveFile() (uint64, error) {
 	// no cgroupv2 support right now
-	if isV2, err := cgroups.IsCgroup2UnifiedMode(); err == nil || isV2 {
+	if isV2 := cgroupmanager.CgroupIsV2(); isV2 {
```
kolyshkin: nit: `if cgroupmanager.CgroupIsV2() {`

internal/config/cgroupmanager/systemdmanager.go
kolyshkin: not yet?

internal/config/cgroupmanager/supported.go
kolyshkin: In this case you can
1. call it during initialization to have a chance to return an error.
2. add a wrapper to call it and discard an error.

I think it would be better this way -- just for simplicity.

internal/config/cgroupmanager/supported.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// CgroupHasMemorySwap returns whether the memory swap controller is present
+func CgroupHasMemorySwap() bool {
+	_cgroupHasMemorySwapOnce.Do(func() {
+		if CgroupIsV2() {
+			_cgroupHasMemorySwap = true
+			return
+		}
+		_, err := os.Stat("/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes")
```
kolyshkin: this won't work for cgroup v2

internal/config/cgroupmanager/supported.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		_cgroupHasHugetlb = strings.Contains(string(controllers), "hugetlb")
+	})
+	return _cgroupHasHugetlb
+}
+
+// CgroupHasPid returns whether the pid controller is present
+func CgroupHasPid() bool {
+	_cgroupHasPidOnce.Do(func() {
+		_, err := libctrcgroups.FindCgroupMountpoint("", "pids")
```
kolyshkin: This is a super ineffective way to check, as it involves parsing /proc/self/mountinfo.

I think the whole section of checking for controllers need to be rewritten. For v2, we check the `cgroup.controllers` file. For v1, unless you have a weird setup, checking for a particular directory presence is fine.

IOW maybe generalize what we have for hugetlb, check for all controllers at once, cache the result, and return it on demand.

kolyshkin: I would do:

```go
_availableControllers map string[bool]
```

```
func CgroupHasController(c string) bool {
  has, ok := _availableControllers[c]
  if !ok {
    logrus.Warnf("CgroupHasController: invalid argument %s", c)
  }
  return has
}
```

(plus locking for the map)

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,183 @@
...
+		return "", "", err
+	}
+
+	cgPath := cgroupParent + ":" + scopePrefix + ":" + sbID
+
+	return cgroupParent, cgPath, nil
+}
+
+func verifyCgroupHasEnoughMemory(cgroupParent, memorySubsystemPath, memoryMaxFilename string) error {
+	// check memory limit is greater than the minimum memory limit of 4Mb
```
kolyshkin, line 104: Hmm, it was 12 MB above.

Also, this comment describes what the function does so it should go just before its declaration.

Either fix and move it, or drop it, either way is fine to me

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,183 @@
...
+		if err := VerifyMemoryIsEnough(memoryLimit); err != nil {
+			return fmt.Errorf("pod %s", err.Error())
+		}
+	}
+	return nil
+}
+
+func VerifyMemoryIsEnough(memoryLimit int64) error {
+	if memoryLimit != 0 && memoryLimit < minMemoryLimit {
+		return fmt.Errorf("set memory limit %v too low; should be at least %v", memoryLimit, minMemoryLimit)
```
kolyshkin, line 137: nit: `s/%v/%d/g`

internal/oci/runtime_oci.go
haircommander: I made the switch because I couldn't tell why it was conditional and it made no sense to have it so. it eventually got squashed into everything else.

internal/config/cgroupmanager/supported.go
haircommander: by @giuseppe's recommendation I opted to just assume memsw is enabled for v2 (see the case above)

internal/config/cgroupmanager/systemdmanager.go
haircommander: bye bye comment :wave:

internal/config/cgroupmanager/supported.go
haircommander: I see that as what this function is doing though, in a way that conforms to the way the other ones do it (providing clarity). Is there any action item here?

internal/config/cgroupmanager/systemdmanager.go
haircommander: I think you need a refresh, I've pushed a newer version

internal/config/cgroupmanager/supported.go
haircommander: okay I've adapted the first version in the new generic controller population code (as suggested below)

internal/config/cgroupmanager/supported.go
haircommander: I've gone for a slightly different solution that doesn't need a lock nor a map (which I think its a bit bloated if we only need two bools)

haircommander:
thanks for the comprehensive review @kolyshkin I think I addressed all of your comments. PTAL (and poke me if I missed something)

haircommander:
/kind api-change

(for the change making the conmon cgroup move fatal)

haircommander:
PTAL @mrunalp @umohnani8 @rhatdan @saschagrunert @kolyshkin

internal/config/node/cgroups.go
```diff
@@ -8,62 +8,33 @@ import (
...
-
-// CgroupHasPid returns whether the pid controller is present
-func CgroupHasPid() bool {
-	cgroupHasPidOnce.Do(func() {
-		_, err := libctrcgroups.FindCgroupMountpoint("", "pids")
-		cgroupHasPid = err == nil
-		cgroupHasPidErr = err
+// CgroupIsV2 returns whether we are using cgroup v2 or v1
+func CgroupIsV2() bool {
+	cgroupIsV2Once.Do(func() {
```
kolyshkin, line 32: You can just do

```go
func CgroupIsV2() bool {
  v2, _ := cgroups.IsCgroup2UnifiedMode()
  return v2
}
```

(and add error checking somewhere in init)

internal/config/cgroupmanager/cgroupmanager_unsupported.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+type CgroupManager interface {
+	GetContainerCgroupPath(string, string) string
+	// cgroup parent, cgroup path, error
+	GetSandboxCgroupPath(string, string) string
+}
+
+func InitializeCgroupManager(cgroupManager string) (CgroupManager, error) {
+	return nil, errors.New("not implemented yet")
```
kolyshkin, line 16: nit: I would return 
```go
errors.New("not implemented for "+runtime.GOOS)`
```

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,148 @@
+// +build linux
```
kolyshkin, line 1: Maybe file name should `systemd_manager.go` (or just `systemd.go`)?

The directory and the package can just  be `cgroup` or `cgroupmgr` or `cgmgr` as per recommendation from https://blog.golang.org/package-names

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,148 @@
...
+	return "", nil
+}
+
+type SystemdV1Manager struct {
+	SystemdManager
+}
+
+// GetSandboxCgroupPath takes the sandbox parent, and sandbox ID. It
+// returns the cgroup parent, cgroup path, and error.
+// it also checks there is enough memory in the given cgroup (4mb is needed for the runtime)
```
kolyshkin, line 67: 1. nit: `s/it /It /`

2. `4mb is needed...` comment is probably wrong and obsolete, I've seen 12 in some other place.

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,148 @@
...
+	parent := defaultSystemdParent
+	if sbParent != "" {
+		parent = sbParent
+	}
+	return parent + ":" + scopePrefix + ":" + containerID
+}
+
+// MoveConmonToCgroup takes the container ID, cgroup parent, conmon's cgroup (from the config) and conmon's PID
+// It attempts to move conmon to the correct cgroup.
+func (*SystemdManager) MoveConmonToCgroup(cid, cgroupParent, conmonCgroup string, pid int) (string, error) {
```
kolyshkin, line 49: I think it deserves to mention what the return value is, same as in other places.

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,148 @@
...
+	if strings.HasSuffix(conmonCgroup, ".slice") {
+		cgroupParent = conmonCgroup
+	}
+	conmonUnitName := fmt.Sprintf("crio-conmon-%s.scope", cid)
+	logrus.Debugf("Running conmon under slice %s and unitName %s", cgroupParent, conmonUnitName)
+	if err := utils.RunUnderSystemdScope(pid, cgroupParent, conmonUnitName); err != nil {
+		return "", errors.Wrapf(err, "Failed to add conmon to systemd sandbox cgroup")
+	}
+	return "", nil
```
kolyshkin, line 58: Makes sense to mention why you return "" (something like "return empty string as path because the cgroup cleanup is performed by systemd" or so).

internal/config/cgroupmanager/systemdmanager.go
```diff
@@ -0,0 +1,148 @@
...
+
+type SystemdV1Manager struct {
+	SystemdManager
+}
+
+// GetSandboxCgroupPath takes the sandbox parent, and sandbox ID. It
+// returns the cgroup parent, cgroup path, and error.
+// it also checks there is enough memory in the given cgroup (4mb is needed for the runtime)
+func (*SystemdV1Manager) GetSandboxCgroupPath(sbParent, sbID string) (cgParent, cgPath string, err error) {
+	return getSandboxCgroupPathForSystemd(sbParent, sbID, cgroupMemorySubsystemMountPathV1, "memory.limit_in_bytes")
```
kolyshkin, line 69: You can actually have 1 method instead of 3.

```go
struct SystemdManager {
  memoryPath, memoryMaxFile string
}
...
// initalize these ^^^ two in New()
...
func (m *systemdManager) SandboxCgroupPath(sbParent, sbID string) (cgParent, cgPath string, err error) {
...
if err = verifyCgroupHasEnoughMemory(cgParent, m.memoryPath, m.memoryMaxFile); err != nil {
￼		return "", "", err
￼	}
}
```

internal/config/node/cgroups.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		cgroupHasMemorySwap = true
+	})
+	return cgroupHasMemorySwap
+}
+
+// CgroupHasHugetlb returns whether the hugetlb controller is present
+func CgroupHasHugetlb() bool {
+	cgroupControllerOnce.Do(func() {
```
kolyshkin, line 62: You don't have to wrap a single function call into yet another `func()`.

internal/config/node/cgroups.go
```diff
@@ -0,0 +1,128 @@
...
+		checkRelevantControllers()
+	})
+	return cgroupHasHugetlb
+}
+
+// CgroupHasPid returns whether the pid controller is present
+func CgroupHasPid() bool {
+	cgroupControllerOnce.Do(func() {
```
kolyshkin, line 70: ditto

internal/config/node/cgroups.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func checkRelevantControllersV1(relevantControllers []controllerInfo) {
+	allControllers, cgroupControllerErr := ioutil.ReadDir("/sys/fs/cgroup")
+	if cgroupControllerErr != nil {
+		return
+	}
+	for _, toCheck := range relevantControllers {
+		for _, controllerStat := range allControllers {
+			if toCheck.name == controllerStat.Name() && controllerStat.IsDir() {
```
kolyshkin, line 107: nit: in general, the shorter the variable scope, the shorter its name should be. The scope of `controllerStat` is just two lines, the code will be more readable if you call it `c` or `st`.

nit2: `IsDir()` check is cheaper, thus it should go first.

internal/config/node/cgroups.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func checkRelevantControllersV2(relevantControllers []controllerInfo) {
+	allControllers, cgroupControllerErr := ioutil.ReadFile("/sys/fs/cgroup/cgroup.controllers")
+	if cgroupControllerErr != nil {
+		return
+	}
+	for _, toCheck := range relevantControllers {
+		for _, controllerName := range strings.Fields(string(allControllers)) {
```
kolyshkin, line 121: ditto wrt shorter var name

internal/config/node/node.go
```diff
@@ -0,0 +1,58 @@
...
+
+// ValidateConfig initializes and validates all of the singleton variables
+// that store the node's configuration.
+// Currently, we check hugetlb, cgroup v1 or v2, pid and memory swap support for cgroups.
+// We check the error at server configuration validation, and if we error, shutdown
+// cri-o early, instead of when we're already trying to run containers.
+func ValidateConfig() error {
+	toInit := []struct {
+		name        string
+		initializer func() bool
```
kolyshkin, line 18: nit: `s/initializer/init`

internal/config/node/node.go
```diff
@@ -0,0 +1,58 @@
...
+		err         *error
+		fatal       bool
+	}{
+		{
+			name:        "hugetlb cgroup",
+			initializer: CgroupHasHugetlb,
+			err:         &cgroupControllerErr,
+			fatal:       false,
```
kolyshkin, line 26: nit: you don't have to initialize struct fields which have default value, and for `bool` the default is `false`. It some cases omitting the default might hurt readability but I doubt it's the case here.

internal/config/node/node_unsupported.go
```diff
@@ -0,0 +1,11 @@
+// +build !linux
+
+package node
+
+import (
+	"github.com/pkg/errors"
+)
+
+func ValidateConfig() error {
+	return errors.Errorf("CRI-O is only supported on linux")
```
kolyshkin, line 10: 1. `s/CRI-O is only supported on linux/CRI-O only supports Linux/`
2. `s/errors.Errorf/errors.New/`
3. You can actually define an error globally (say in `internal/lib/`)
```go
var errLinuxOnly = errors.New("CRI-O only supports Linux")
```
and use it for all the places like this
```go
return lib.errLinuxOnly
```
return

internal/oci/oci_linux.go
```diff
@@ -201,7 +159,7 @@ func metricsToCtrStats(c *Container, m *cgroups.Metrics) *ContainerStats {
 // not parsable.
 func getTotalInactiveFile() (uint64, error) {
 	// no cgroupv2 support right now
```
kolyshkin, line 161: Maybe add `TODO:` here

internal/oci/oci_unsupported.go
```diff
@@ -7,7 +7,8 @@ import (
 	"syscall"
 )
 
-func (r *Runtime) createContainerPlatform(c *Container, cgroupParent string, pid int) {
+func (r *Runtime) createContainerPlatform(c *Container, cgroupParent string, pid int) error {
+	return nil
```
kolyshkin, line 11: `return lib.errLinuxOnly`?

kolyshkin: (see above for where this is coming from)

internal/oci/runtime_oci.go
```diff
@@ -107,7 +99,13 @@ func (r *runtimeOCI) CreateContainer(c *Container, cgroupParent string) (err err
 		"--exit-dir", r.config.ContainerExitsDir,
 		"--socket-dir-path", r.config.ContainerAttachSocketDir,
 		"--log-level", logrus.GetLevel().String(),
-		"--runtime-arg", fmt.Sprintf("%s=%s", rootFlag, r.root))
+		"--runtime-arg", fmt.Sprintf("%s=%s", rootFlag, r.root),
+		"--syslog"}
```
kolyshkin, line 103: syslog used to be added only for non-systemd case. What is the reason to change it? (sorry if I missed it)

pkg/config/config.go
```diff
@@ -774,12 +786,13 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 
 	// check for validation on execution
 	if onExecution {
-		if err := c.ValidateRuntimes(); err != nil {
+		var err error
+		if err = c.ValidateRuntimes(); err != nil {
```
kolyshkin, line 790: I don't think this is an improvement. Same for similar code below. Can you drop this?

server/container_create_linux.go
```diff
@@ -522,12 +458,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 			specgen.SetLinuxResourcesCPUShares(uint64(resources.GetCpuShares()))
 
 			memoryLimit := resources.GetMemoryLimitInBytes()
+
```
kolyshkin, line 461: nit: remove this newline

server/container_create_linux.go
```diff
@@ -1254,7 +1165,7 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
 		g.AddMount(tmpfsMnt)
 	}
 
-	if cgroups.IsCgroup2UnifiedMode() {
+	if node.CgroupIsV2() {
```
kolyshkin, line 1168: OK I just saw there is a version of the function the way you want it, without an error returned as a second parameter. Perhaps we can convert everyone to just use it?

Or, if you don't like the name, you can do
```go
package node
...
import (
...
"github.com/opencontainers/runc/libcontainer/cgroups"
...

var CgroupIsV2 = cgroups.IsCgroup2UnifiedMode()
```
and use `node.CgroupIsV2()` the way you do.

internal/config/cgroupmanager/cgroupmanager.go
```diff
@@ -58,9 +58,9 @@ func SetCgroupManager(cgroupManager string) (CgroupManager, error) {
 	switch cgroupManager {
 	case systemdCgroupManager:
 		if node.CgroupIsV2() {
-			return new(Systemdv2Manager), nil
+			return new(SystemdV2Manager), nil
```
kolyshkin, line 62: this belongs to the first commit in the series, isn't it?

internal/config/node/cgroups.go
TomSweeneyRedHat: I'll just counter the argument.  I know shorter is in many style guides, but I find it much clearer to read a variable name than just a char or two, even if it only exists across a few lines.  My $.02 and YMMV.

internal/config/node/node_unsupported.go
haircommander: I'd like to do this as a follow up PR. I think we're going to drop all distinction between platforms and make clear our plans are only linux support for the near future

internal/oci/runtime_oci.go
haircommander: ok I dropped this and will add in a follow up PR

internal/config/node/cgroups.go
haircommander: I used libcontainer as suggested below

internal/config/cgroupmanager/systemdmanager.go
haircommander: I went for cgmgr for the package and systemd/cgroupfs for the file names

internal/config/cgroupmanager/systemdmanager.go
haircommander: I just dropped that part

internal/config/cgroupmanager/systemdmanager.go
haircommander: added a comment :+1:

internal/config/cgroupmanager/systemdmanager.go
haircommander: also added one here

internal/config/node/cgroups.go
haircommander: I fixed the allControllers variable to be more informative, and then made the iterative have a shorter name: a compromise :)

internal/config/node/cgroups.go
haircommander: ditto as above

internal/config/node/node.go
haircommander: there are no fields that are fatal, so I've dropped that for now

pkg/config/config.go
haircommander: fixed

internal/config/cgroupmanager/cgroupmanager.go
haircommander: yeah the commits are all over the place. I am inclined to leave it but could fix if you want.

haircommander:
```

[sig-storage]  In-tree Volumes [Driver: hostPathSymlink] [Testpattern: Inline-volume  (default fs)] subPath should support readOnly file specified in the  volumeMount [LinuxOnly] expand_more | [sig-storage]  In-tree Volumes [Driver: hostPathSymlink] [Testpattern: Inline-volume  (default fs)] subPath should support readOnly file specified in the  volumeMount [LinuxOnly] expand_more
-- | --
[sig-storage]  In-tree Volumes [Driver: hostPathSymlink] [Testpattern: Inline-volume  (default fs)] subPath should support readOnly file specified in the  volumeMount [LinuxOnly] expand_more
[sig-storage] HostPath should support r/w [NodeConformance] expand_more | [sig-storage] HostPath should support r/w [NodeConformance] expand_more
[sig-storage] HostPath should support r/w [NodeConformance] expand_more

```
[hm](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3565/test_pull_request_crio_e2e_crun_fedora/4372/)
/retest

haircommander:
```
failed: [localhost] (item=onsi/gomega) => {"ansible_loop_var": "item", "changed": true, "cmd": "go get -u \"github.com/onsi/gomega\"", "delta": "0:00:34.974876", "end": "2020-05-19 16:13:10.402481", "item": "onsi/gomega", "msg": "non-zero return code", "rc": 1, "start": "2020-05-19 16:12:35.427605", "stderr": "# cd /go/src/gopkg.in/yaml.v2; git pull --ff-only\nremote: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/go-yaml/yaml.git/info/refs?service=git-upload-pack: net/http: request canceled (Client.Timeout exceeded while awaiting headers)\nfatal: unable to access 'https://gopkg.in/yaml.v2/': The requested URL returned error: 502\npackage gopkg.in/yaml.v2: exit status 1", "stderr_lines": ["# cd /go/src/gopkg.in/yaml.v2; git pull --ff-only", "remote: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/go-yaml/yaml.git/info/refs?service=git-upload-pack: net/http: request canceled (Client.Timeout exceeded while awaiting headers)", "fatal: unable to access 'https://gopkg.in/yaml.v2/': The requested URL returned error: 502", "package gopkg.in/yaml.v2: exit status 1"], "stdout": "", "stdout_lines": []}
changed: [localhost] => (item=jteeuwen/go-bindata/go-bindata) => {"ansible_loop_var": "item", "changed": true, "cmd": "go get -u \"github.com/jteeuwen/go-bindata/go-bindata\"", "delta": "0:00:00.690482", "end": "2020-05-19 16:13:11.258253", "item": "jteeuwen/go-bindata/go-bindata", "rc": 0, "start": "2020-05-19 16:13:10.567771", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```
/retest

haircommander:
```
failed: [localhost] (item=onsi/gomega) => {"ansible_loop_var": "item", "changed": true, "cmd": "go get -u \"github.com/onsi/gomega\"", "delta": "0:00:38.846451", "end": "2020-05-19 17:08:09.217251", "item": "onsi/gomega", "msg": "non-zero return code", "rc": 1, "start": "2020-05-19 17:07:30.370800", "stderr": "# cd /go/src/gopkg.in/yaml.v2; git pull --ff-only\nremote: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/go-yaml/yaml.git/info/refs?service=git-upload-pack: dial tcp: lookup github.com on 127.0.0.53:53: server misbehaving\nfatal: unable to access 'https://gopkg.in/yaml.v2/': The requested URL returned error: 502\npackage gopkg.in/yaml.v2: exit status 1", "stderr_lines": ["# cd /go/src/gopkg.in/yaml.v2; git pull --ff-only", "remote: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/go-yaml/yaml.git/info/refs?service=git-upload-pack: dial tcp: lookup github.com on 127.0.0.53:53: server misbehaving", "fatal: unable to access 'https://gopkg.in/yaml.v2/': The requested URL returned error: 502", "package gopkg.in/yaml.v2: exit status 1"], "stdout": "", "stdout_lines": []}
changed: [localhost] => (item=jteeuwen/go-bindata/go-bindata) => {"ansible_loop_var": "item", "changed": true, "cmd": "go get -u \"github.com/jteeuwen/go-bindata/go-bindata\"", "delta": "0:00:01.032629", "end": "2020-05-19 17:08:10.421864", "item": "jteeuwen/go-bindata/go-bindata", "rc": 0, "start": "2020-05-19 17:08:09.389235", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```
/retest

haircommander:
happy tests again
PTAL @mrunalp @kolyshkin @umohnani8 @saschagrunert @rhatdan

haircommander:
@mrunalp @kolyshkin @giuseppe re-ping

haircommander:
/close

replaced by https://github.com/cri-o/cri-o/pull/3810

Issue #3567, opened 2020-04-14
Unable to retrieve the image spec using crictl inspecti 


vivekanandpoojari:
**Description**
crictl inspecti does not return the OCI imagespec.

**Steps to reproduce the issue:**
1. Install cri-o on a ubuntu machine
2. Pull busybox image
3. crictl inspecti ##busyBoxImageID##

**Describe the results you received:**
The inspect data does not include the OCI Image spec in the info 

```
crictl inspecti 83aa35aa1c79e
{
  "status": {
    "id": "83aa35aa1c79e4b6957e018da6e322bfca92bf3b4696a211b42502543c242d6f",
    "repoTags": [
      "docker.io/library/busybox:latest"
    ],
    "repoDigests": [
      "docker.io/library/busybox@sha256:afe605d272837ce1732f390966166c2afff5391208ddd57de10942748694049d",
      "docker.io/library/busybox@sha256:b26cd013274a657b86e706210ddd5cc1f82f50155791199d29b9e86e935ce135"
    ],
    "size": "1441595",
    "uid": null,
    "username": ""
  }
}
```

**Describe the results you expected:**
When I execute the same command on containerd runtime.
I get the complete OCI image spec.

```
crictl inspecti 83aa35aa1c79e
{
  "status": {
    "id": "sha256:83aa35aa1c79e4b6957e018da6e322bfca92bf3b4696a211b42502543c242d6f",
    "repoTags": [
      "docker.io/library/busybox:latest"
    ],
    "repoDigests": [
      "docker.io/library/busybox@sha256:b26cd013274a657b86e706210ddd5cc1f82f50155791199d29b9e86e935ce135"
    ],
    "size": "764687",
    "uid": null,
    "username": ""
  },
  "info": {
    "chainID": "sha256:a6d503001157aedc826853f9b67f26d35966221b158bff03849868ae4a821116",
    "imageSpec": {
      "created": "2020-03-10T00:19:33.019716493Z",
      "architecture": "amd64",
      "os": "linux",
      "config": {
        "Env": [
          "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
        ],
        "Cmd": [
          "sh"
        ]
      },
      "rootfs": {
        "type": "layers",
        "diff_ids": [
          "sha256:a6d503001157aedc826853f9b67f26d35966221b158bff03849868ae4a821116"
        ]
      },
      "history": [
        {
[...]

PR #3570, opened 2020-04-14
Add OCI image spec to image stats info

saschagrunert:
#### What type of PR is this?

/kind feature

Added the full OCI image spec to the verbose `ImageStatusResponse`,
for example when running `crictl inspecti $IMAGE`.

Fixes https://github.com/cri-o/cri-o/issues/3567
I also added the output only if the request is specified in verbose mode.

```release-note
- Added the full OCI image spec to the verbose `ImageStatusResponse`, for example when running `crictl inspecti $IMAGE`.
```

### DIFF
--- a/server/image_status.go
+++ b/server/image_status.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
@@ -8,6 +9,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/cri-o/cri-o/internal/log"
 	pkgstorage "github.com/cri-o/cri-o/internal/storage"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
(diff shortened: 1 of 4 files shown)


harche:
Should we have corresponding unit tests in `image_test.go` and `image_status_test.go` for this change?

saschagrunert:
> Should we have corresponding unit tests in `image_status_test.go` and/or `bats` test for this change?

Sure, I added unit and integration tests. :+1:

TomSweeneyRedHat:
@saschagrunert lint isn't happy with this

haircommander:
lint fix in https://github.com/cri-o/cri-o/pull/3566

haircommander:
LGTM

@mtrmac @vrothberg PTAL

saschagrunert:

server/image_status.go
```diff
@@ -63,7 +65,13 @@ func (s *Server) ImageStatus(ctx context.Context, req *pb.ImageStatusRequest) (r
 				RepoDigests: status.RepoDigests,
 				Size_:       size,
 			},
-			Info: status.Labels,
```
mtrmac, line 66: `Info` was always documented to be set only if `req.Verbose`, but is it OK at this point in the release to remove it?

Also, the way `Info` now represents labels has changed. Is it OK to change that?

haircommander:
@mtrmac you're correct these changes haven't made it into a stable release yet, and thus I'm not bashful about changing the structure of the json

server/image_status.go
saschagrunert: Yes I think It is okay since we did not use `Info` at all before 1.18. :+1:

saschagrunert:
Alright, so I do not see any further blockers on this one, right?

PR #3583, opened 2020-04-15
[1.17] bump storage to 01ee1656d552

haircommander:
#### What type of PR is this?

/kind dependency-change

bumps c/storage to crio-release-1.17 which contains https://github.com/containers/storage/pull/581/commits/01ee1656d552287053be7f5b08eca72962a9ed9f, as we saw 
```

Apr 15 13:44:43 ip-10-0-129-128.us-east-2.compute.internal crio[70396]: panic: attempted to update last-writer in lockfile without the write lock
Apr 15 13:44:43 ip-10-0-129-128.us-east-2.compute.internal crio[70396]: goroutine 135 [running]:
Apr 15 13:44:43 ip-10-0-129-128.us-east-2.compute.internal crio[70396]: panic(0x560e8b171000, 0x560e8b4bc160)
``` 

```release-note
fix containers/storage panic
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/containers/image/v5 v5.3.1
 	github.com/containers/libpod v1.8.1
 	github.com/containers/ocicrypt v0.0.0-20190930154801-b87a4a69c741
-	github.com/containers/storage v1.16.6
+	github.com/containers/storage v1.16.7-0.20200415192751-01ee1656d552
 	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f // indirect
 	github.com/coreos/go-systemd/v22 v22.0.0
 	github.com/cpuguy83/go-md2man v1.0.10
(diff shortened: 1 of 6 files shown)


docs/crio-status.8.tmp
```diff
@@ -0,0 +1,114 @@
+<!-- This file has been autogenerated, please do not edit -->.TH "crio-status" "8" " A tool for CRI-O status retrieval"
```
mrunalp, line 1: Was this added my mistake?

docs/crio.conf.5.tmp
```diff
@@ -0,0 +1,388 @@
+.TH "crio.conf" "5" " Kubernetes Container Runtime Daemon for Open Container Initiative Containers" "Aleksa Sarai" "OCTOBER 2016"
```
mrunalp, line 1: This also looks like a tmp file.

mtnbikenc:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1779811

docs/crio-status.8.tmp
haircommander: hm weird

haircommander:
we should try to branch off of storage 1.16.6 instead of pulling in 1.18.2

haircommander:
test e2e_features_fedora

TomSweeneyRedHat:
Could you bump to c/storage v1.17 or higher?

haircommander:
> Could you bump to c/storage v1.17 or higher?

theoretically, but it would require many other version bumps, which at this stage I'm hesitant to do. Are there fixes in 1.17+ you think should be here @TomSweeneyRedHat

TomSweeneyRedHat:
My main concern is having a non-tagged release like this.  I'd much prefer seeing a release tag such as 1.16.3 in the vendoring for future debugging.  I think but I'm not sure you've the  "DeleteContainer(): wait for all goroutines" fix that you put together that's in 1.18.1 in here.  Also, the UID/GID fix for layers that went in for 1.17 would be good too.

haircommander:
I've moved cri-o 1.17 and 1.16 to use a storage branch crio-release-1.17, so we can keep getting vital patches into these branches without keeping up with, say, latest podman or c/common. this is a scheme we have with cri-o 1.13 as well. I don't think it's necessary to tag these branches as an official storage release, they're more of downstream branches to carry patches.

PR #3601, opened 2020-04-16
[WIP-Rebase] *: Add --pid-namespace=[container|pod|pod-container]

dougsland:
#### What type of PR is this?

/kind feature

So crio callers can pick their PID namespace approach.  This seems
like something that folks might want to configure per-pod, or possibly
even per-container, but for now follow --enable-shared-pid-namespace
and make it per-crio-daemon.

The "Deprecated:" paragraph approach in Go comments is recommended in
[1].

I tried to find a more compact form for the REDIS_IN_INFRA test, but:

* =~ is not in POSIX [2], so we need to use Bash's internal [[.
* '[ ! {expression} ]' is in POSIX [2], but Bash's [[ seems to look
  for ! before doing variable expansion.

Because of those, we can't use POSIX's ${parameter:+word} [3]:

  NOT="!"
  [[ ${REDIS_IN_INFRA:+${NOT}} "$output" =~ "redis" ]]

and the best alternative I can find is the if/else that I've gone with
in this commit.

The ADDITIONAL_CRIO_OPTIONS approach is easier to maintain than the
old approach with settings for every option, because we no longer need
to maintain defaults in two locations (lib/config.go and
test/helpers.bash).  For this commit, I've only dropped
ENABLE_SHARED_PID_NAMESPACE, but we may want to extend this approach
to more variables in the future.

PIDNamespaceType follows the ImageVolumes pattern at Mrunal's
recommendation [4].

[1]: https://blog.golang.org/godoc-documenting-go-code
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html
[3]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02
[4]: #1280 (comment)

Based on: https://github.com/cri-o/cri-o/pull/1280/

Signed-off-by: W. Trevor King <wking@tremily.us>

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -58,6 +58,7 @@ crio
 [--pause-command]=[value]
 [--pause-image-auth-file]=[value]
 [--pause-image]=[value]
+[--pid-namespace]=[value]
 [--pids-limit]=[value]
 [--pinns-path]=[value]
 [--profile-port]=[value]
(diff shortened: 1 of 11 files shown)


mrunalp:
This isn't really needed since k8s added support for sharing pid namespaces at the pod level.

dougsland:
> This isn't really needed since k8s added support for sharing pid namespaces at the pod level.

Sure @mrunalp. Thanks!

PR #3607, opened 2020-04-17
Stop container when there is error in createSandboxContainer

tedyu:
/kind bug

If createSandboxContainer returns non-nil error, we should stop the container.

```release-note
None
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -826,6 +826,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 	if err != nil {
 		return nil, fmt.Errorf("failed to mount container %s(%s): %v", containerName, containerID, err)
 	}
+	defer func() {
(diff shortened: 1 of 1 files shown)


haircommander:
/ok-to-test
/approve
LGTM

mrunalp:
@tedyu can you check and open backports if needed?

haircommander:
come to think of it, do we need this? I seem to remember @nalind saying the DeleteContainer call will umount (do the work of StopContainer) if necessary, and we already call DeleteContainer on a defer

tedyu:
Looking at the operation for runtimeService#StopContainer:
```
	_, err = r.storageImageServer.GetStore().Unmount(container.ID, false)
```
The above is not done by DeleteContainer().
So the call to StopContainer is needed.

umohnani8:
LGTM

PR #3606, opened 2020-04-17
bump to version 1.17.4

haircommander:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
as title says, as well as pull in https://github.com/cri-o/cri-o/pull/3607
#### Does this PR introduce a user-facing change?

```release-note

```

PR #3608, opened 2020-04-17
[release-1.17] Stop container when there is error in createSandboxContainer

tedyu:
/kind bug

If createSandboxContainer returns non-nil error, we should stop the container.

This is backport of #3607

```release-note
None
```


haircommander:
I've already pulled this in here https://github.com/cri-o/cri-o/pull/3606

tedyu:
Didn't see #3606 prior to submission

PR #3613, opened 2020-04-19
[1.18] Revert the auto-injecting version changes

umohnani8:

#### What type of PR is this?

 /kind design

Reverting this so we don't run into build issues with
openshift again. We will revisit in master to come up with
a less fragile way of doing this.

```release-note

```


internal/version/version.go
```diff
@@ -19,9 +19,11 @@ import (
 	"github.com/cri-o/cri-o/utils"
 )
 
+// Version is the version of the build.
+const Version = "1.18.0-rc1"
```
mrunalp, line 23: should this be `1.18.1`?

mrunalp: Actually let's make it `1.18.0-rc2` instead.

umohnani8: fixed

cgwalters:
Thanks for the revert; this is stalling RHCOS builds and we're trying to get some other things landed too.

I think gathering version information from git is useful as an *informational* thing, but it gets fragile if one is trying to actively make decisions based on it.  Even without the weird "lookaside cache" case where Koji does that breaks any relationship with the upstream git repo.  For example, it's reasonable for downstream project consumers to make their own git repositories and cherry pick patches there - and they shouldn't have to also manage syncing up tags and such.

It's just more robust to do `const version = "1.18"` in the source code and the cost of doing that is low.

But, it definitely *also* makes sense to record the git commit, I just wouldn't use it for things like crio-wipe.

cgwalters:
I also tried setting `VERSION=%{version}` in our RPM spec build, and I swear it worked once (I verified that `./crio version` had a version again) but then failed (still had no version tag) when I tried it a second time...didn't try too hard to debug it though.

mrunalp:
@lsm for new build once this merges.

PR #3621, opened 2020-04-21
[1.18] Fix CI

saschagrunert:
#### What type of PR is this?

/kind failing-test

Fixing the CircleCI pipeline for this branch.

None
None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ lint: .gopathok ${GOLANGCI_LINT}
 	${GOLANGCI_LINT} run
 
 shellfiles: ${SHFMT}
-	$(eval SHELLFILES=$(shell ${SHFMT} -f . | grep -v vendor/))
+	$(eval SHELLFILES=$(shell ${SHFMT} -f . | grep -v vendor/ | grep -v hack))
 
 shfmt: shellfiles
 	${SHFMT} -w -i 4 -d ${SHELLFILES}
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -92,6 +92,7 @@ else
 GCFLAGS = -gcflags '-N -l'
 endif
 
+VERSION ?= $(shell $(GO_RUN) ./scripts/latest-version)
```
haircommander, line 95: is this used anywhere?

saschagrunert: Yes in the release notes script for example.

haircommander:
I'm not a huge fan of partially using this script to scrape the version... the hardcoded version and the dynamic one could differ...
Unfortunately, we're stuck between a rock and a hard place here.

Either we:
- take the approach of this PR (use both hardcoded and dynamic)
- hardcode the VERSION in the makefile (fixing the release notes scripts)
    - one would think this would do what we want, but I think we would have similar problems pulling this into openshift that caused the revert in the first place, I have to check the spec file
- hardcode in Makefile AND in internal/version/version.go
    - which could also be prone to skew
- parse the version from the output of crio --version for the release notes
    - seems needlessly complicated...

saschagrunert:
> I'm not a huge fan of partially using this script to scrape the version... the hardcoded version and the dynamic one could differ...
> Unfortunately, we're stuck between a rock and a hard place here.
> 
> Either we:
> 
> * take the approach of this PR (use both hardcoded and dynamic)
> * hardcode the VERSION in the makefile (fixing the release notes scripts)
>   
>   * one would think this would do what we want, but I think we would have similar problems pulling this into openshift that caused the revert in the first place, I have to check the spec file
> * hardcode in Makefile AND in internal/version/version.go
>   
>   * which could also be prone to skew
> * parse the version from the output of crio --version for the release notes
>   
>   * seems needlessly complicated...

I'll check if I can fix this pipeline by relying on the version in the internal/version package.

haircommander:
/lgtm

thanks @saschagrunert

PR #3625, opened 2020-04-22
[WIP] Add support for KVM/Init label selection

rhatdan:

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

```release-note

```


rhatdan:
@umohnani8 Two problems I am having with this PR,

1 I can seem to vendor a non released version of libpod to get kvm selection, code.  I will attempt to back port this into podman-1.9.1 release.
2 I need to get the ContainerRuntimeType, before creating the container in order to select the SELinux type.  Not sure the best way to do this is.

server/container_create_linux.go
```diff
@@ -398,10 +399,37 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
...
+	if err != nil {
+		return nil, err
+	}
+	specgen.SetProcessArgs(processArgs)
+
 	mountLabel := containerInfo.MountLabel
 	var processLabel string
 	if !privileged {
 		processLabel = containerInfo.ProcessLabel
+		if strings.Contains(processArgs[0], "/sbin/init") || (filepath.Base(processArgs[0]) == oci.SystemdCgroupsManager) {
```
haircommander, line 418: can we assume that init will always be in `/sbin/`?

haircommander: nevermind, we already were assuming that

umohnani8:
@rhatdan you should be able to vendor a non-release version of libpod, this should do the trick:
```
go get github.com/containers/libpod@commit
make vendor
```
And for the containerRuntimeType, make we can use https://github.com/cri-o/cri-o/blob/master/pkg/config/config.go#L530 to determine it. I am guessing you need to know if we are using kata or not right?

rhatdan:
@umohnani8 Yes I know but this was failing for some reason.  I will attempt to look at this again, if I have time, otherwise you can take it over. BTW Later today I would like to talk to you about this PR.  I have some questions about how kubernetes will work with it.

haircommander:
absoleted by https://github.com/cri-o/cri-o/pull/3754/ and https://github.com/cri-o/cri-o/pull/3861

PR #3628, opened 2020-04-22
Add Installation Guide with Ansible

### LINKED ISSUE #3552: Ansible Role for CRI-O
I am now maintenance an Ansible Role for CRI-O installation (https://github.com/alvistack/ansible-role-cri_o), by downloading the static binary release archive and extract it for following OS:
- Ubuntu 16.04/18.04/19.10
- CentOS/RHEL 7/8
- openSUSE 15.1

Could we also introduce this as an alternative installation method, e.g. under https://github.com/cri-o/cri-o#installing-cri-o?

hswong3i:
#### What type of PR is this?

> /kind documentation

Provide additional installation options by using Ansible with https://github.com/alvistack/ansible-role-cri_o from source code.

Fixes #3552

```release-note

```

### DIFF
--- a/tutorials/setup.md
+++ b/tutorials/setup.md
@@ -14,6 +14,7 @@ This guide will walk you through the installation of [CRI-O](https://github.com/
     + [Debian - Raspbian - Ubuntu](#debian---raspbian---ubuntu)
   * [Get Source Code](#get-source-code)
   * [Build](#build)
+    + [Install with Ansible](#install-with-ansible)
     + [Build Tags](#build-tags)
   * [Static builds](#static-builds)
     + [Creating a release archive](#creating-a-release-archive)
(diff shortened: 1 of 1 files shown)


tutorials/setup.md
```diff
@@ -188,6 +189,20 @@ make BUILDTAGS=""
 sudo make install
 ```
 
+### Install with Ansible
+
+An [Ansible Role](https://github.com/alvistack/ansible-role-cri_o) is also available to automate the above steps:
```
saschagrunert, line 194: Do you think it would make sense to add the role to this repository here (as well)?

hswong3i: To be honest the role will not works in solo: it also depends on other roles, e.g. golang, git, runc, conmon, etc (see https://github.com/alvistack/ansible-role-cri_o/blob/master/ansible-role-requirements.yml).

Why the steps look such simple now, just because molecule will fetch the dependency before the converge start. So the truth is you will have more than single Ansible Role being used. IMHO, in this repo we may not hope to locally embed too many Ansible Role dependency, it shouldn’t be this repo target. 

Moreover for production deployment use cases, you may hope to use my Ansible Collection for Kubernetes (see https://github.com/alvistack/ansible-collection-kubernetes/blob/master/playbooks/cri_o-install.yml), for multi target DevOps management. BTW, promote the Ansible Collection here may again overkill.

saschagrunert: Alright thanks for the clarification. Sounds good to me

PR #3630, opened 2020-04-22
Add info logs where needed and check for context erroring

umohnani8:
#### What type of PR is this?

> /kind cleanup

Add info logs to all the defers in container and sandbox create.
Add logs for when we get a request from kubelet to stop and remove
a container and sandbox.
And Check for context erroring before returning from longer requests
The kubelet times out on the create request but crio doesn't error out
when kubelet client times out. So, we end up with the pod being created
and then kubelet keeps on trying to create the pod again and it can't
since the name is reserved.
With the change, we error out on timeouts so the pod creation fails and
the name is unreserved so later attempts by the kubelet to create a pod
or container may succeed.

```release-note

```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime/debug"
 	"strconv"
 	"strings"
 
@@ -546,6 +547,13 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 	if sb.Stopped() {
 		return nil, fmt.Errorf("CreateContainer failed as the sandbox was stopped: %v", sbID)
 	}
+
+	defer func() {
+		if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+			log.Infof(ctx, "createCtr: context was either canceled or the deadline was exceeded: %v", ctx.Err())
+			debug.PrintStack()
+		}
+	}()
 
 	ctr := container.New(ctx)
 	if err := ctr.SetConfig(req.GetConfig()); err != nil {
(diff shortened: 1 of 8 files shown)


server/container_create.go
```diff
@@ -607,6 +611,11 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 
 	newContainer.SetCreated()
 
+	if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+		log.Infof(ctx, "ctrCreate: context was either canceled or the deadline was exceeded: %v", ctx.Err())
```
haircommander, line 615: nit: above it's createCtr, here it's ctrCreate

I'm not a huge fan of abbreviating ctr here, but I understand we don't want to be logging unnecessarily.

Do we really need the function names? if so, then can you make them uniform?

haircommander:
I would rather have the error messages be sufficiently unique, rather than saying what function they happened in (and whether they happened in a defer)

server/container_create.go
mrunalp: having function names will make it easier to track these.

umohnani8: yeah, it is meant to be uniform - mistyped here. Quite a few of the functions in defer are the same for sandbox_run and container_create so I prefer to point out where the defer was occurring for easy debugging. I can take out the "defer" from the log, but would like to keep the function names.

umohnani8: Fixed

server/container_create.go
```diff
@@ -607,6 +611,11 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 
 	newContainer.SetCreated()
 
+	if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
```
rphillips, line 614: We could move the context check to the top of the function into a deferred and catch the context cancelled or deadline exceeded anywhere. it would be beneficial to debug.PrintStack() within that function as well to print the stack. It would give us function that the context cancelled or exceeded on.

umohnani8: Fixed

rphillips:
lgtm

/assign @haircommander

PR #3635, opened 2020-04-22
[1.17] crio wipe: version file persist

haircommander:
#### What type of PR is this?

/kind feature

Also, update the crio wipe tests to test different scenerios and move the tests to a higher directory so they're run on PRs

```release-note
add --version-file-persist for finer grained control on crio wipe. version-file-persist should be in a persistent directory, and version-file should be in a temporary directory, if using crio wipe.
```


haircommander:
one thing to note is upgrading to this version will mean, on first run of crio wipe, the images and containers will  be removed, despite it not being a minor version upgrade..

cmd/crio/wipe.go
```diff
@@ -51,15 +51,15 @@ func crioWipe(c *cli.Context) error {
...
 	// if we should not wipe, exit with no error
-	// since we won't wipe images without wiping containers, we can just check the one condition
 	if !shouldWipeContainers {
+		// we should not wipe images without wiping containers
+		// in a future release, we should wipe both container and images if only shouldWipeImages is true.
+		// However, now, we cannot expect users to have version-file-persist after having upgraded
+		// to this version. Skip the wipe, for now, and log about it.
+		if shouldWipeImages {
+			logrus.Infof("legacy version-file path found, but new version-file-persist path not. Skipping wipe")
+		}
```
TomSweeneyRedHat, line 62: Do we have a "TODO:" tag that we use to make sure we clean stuff up?  And/or is there an issue spun up for this that we could reference here?

haircommander: I intend on cherry-picking this version into 1.18, then writing the new version in master. I am leaving myself in charge of remembering :smile:

PR #3638, opened 2020-04-23
Fix linkmode for static binaries

saschagrunert:
#### What type of PR is this?

/kind bug

The actual output of ldd is in the error which caused static binaries to
return the wrong linkmode.

None
None

```release-note
- Fixed wrong linkmode output (on `crio version`) for static binaries
```


TomSweeneyRedHat:
LGTM

PR #3642, opened 2020-04-23
Makefile: include -nobuild install targets

lsm5:
Distro packaging often uses non-default build and linker flags.
The current Makefile cannot be used in the package build process as it
will end up rebuilding the binaries with the default flags.

This commit introduces install targets which do not depend on the build
targets. Distro packages should prefer these if they want to use any
non-default flags.

NOTE: This commit effectively leaves prior targets unchanged, so users
won't notice any difference.

#### What type of PR is this?

/kind dependency-change

(Couldn't find a better /kind)

This PR makes it easier for distros to re-use Makefile install targets
while continuing to build with custom flags/rules.

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -434,23 +434,29 @@ bundle-test:
 
 install: .gopathok install.bin install.man install.completions install.systemd install.config
 
-install.bin: binaries
+install.bin-nobuild:
 	install ${SELINUXOPT} -D -m 755 bin/crio $(BINDIR)/crio
 	install ${SELINUXOPT} -D -m 755 bin/crio-status $(BINDIR)/crio-status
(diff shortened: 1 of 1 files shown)


lsm5:
@dougsland @mrunalp @haircommander @saschagrunert PTAL

lsm5:
Inspired by https://github.com/containers/libpod/pull/5954

haircommander:
/approve
LGTM

haircommander:
test issues should be unrelated
they're being fixed [here](https://github.com/cri-o/cri-o/pull/3643)

TomSweeneyRedHat:
LGTM assuming happy tests

haircommander:
can you rebase off of latest master to get test fixes @lsm5

PR #3646, opened 2020-04-23
Obtain lock before storing decoded state

tedyu:
#### What type of PR is this?

/kind bug

This PR adds locking around the state decoding in Container#FromDisk

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -141,6 +141,7 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (err erro
 
 	select {
 	case err = <-createdCh:
+		f.Close()
 		if err != nil {
 			return errors.Errorf("CreateContainer failed: %v", err)
 		}


haircommander:
integration tests will fail, and it won't be this PRs fault. the fix is [here](https://github.com/cri-o/cri-o/pull/3643)

tedyu:
```
func (c *Container) Metadata() *pb.ContainerMetadata {
```
It seems the above method should obtain read lock.
@haircommander @mrunalp 
What do you think ?

tedyu:
```
func (c *Container) State() *ContainerState {
	c.opLock.RLock()
```
I think a copy of ContainerState should be made whose pointer is returned.

I will add the above change to this PR.

rphillips:
the decode should probably be done in a new instance, then locked so it is atomic. I'm still seeing the name error... still digging.

```diff
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -161,8 +161,15 @@ func (c *Container) FromDisk() error {
        }
        defer jsonSource.Close()

+       var state ContainerState
        dec := json.NewDecoder(jsonSource)
-       return dec.Decode(c.state)
+       if err := dec.Decode(&state); err != nil {
+               return err
+       }
+       c.opLock.Lock()
+       c.state = &state
+       c.opLock.Unlock()
+       return nil
 }
```

tedyu:
@rphillips confirmed that the change in server/sandbox_list.go for infra container is needed (for reducing the occurrence of 'ID does not exist' error).

server/sandbox_list.go
```diff
@@ -60,7 +60,7 @@ func (s *Server) ListPodSandbox(ctx context.Context, req *pb.ListPodSandboxReque
 		}
 
 		rStatus := pb.PodSandboxState_SANDBOX_NOTREADY
-		if sb.Ready(false) {
+		if sb.Ready(true) {
```
mrunalp, line 63: Concern with this is that PLEG may go unhealthy if some pod operation takes long.

tedyu:
[update-cont-stat.txt](https://github.com/cri-o/cri-o/files/4525701/update-cont-stat.txt)

server/sandbox_list.go
tedyu: Is it due to runtimeOCI#UpdateContainerStatus taking long ?
See update-cont-stat.txt below for proposal on reducing the duration opLock is held.

mrunalp: We skip over sandboxes that aren't created a few lines above. So, not-ready, not-created sandboxes shouldn't show up here.

tedyu:
Ran test suite locally with patch:
```
Ginkgo ran 17 suites in 44.24401059s
Test Suite Passed
```

server/sandbox_list.go
tedyu: See #3650

tedyu: Is it possible that the sandbox has exited by the time we retrieve the state ?

haircommander:
can you rebase with latest master @tedyu

tedyu:
@haircommander 
I have pushed again. Though the previous patch seems to have no conflict.

haircommander:
/retest

I was hoping for some CI fixes to be picked up, thanks!

server/sandbox_list.go
haircommander: I feel like this change is out of scope for the PR. Is there a reason to have it here?

tedyu: @haircommander 
I have taken this out of the PR.

haircommander:
/approve

umohnani8:
LGTM

internal/oci/container.go
```diff
@@ -170,6 +177,8 @@ func (c *Container) StatePath() string {
 
 // CreatedAt returns the container creation time
 func (c *Container) CreatedAt() time.Time {
+	c.opLock.RLock()
```
haircommander, line 180: If we make this change, then every call to CreatedAt() that follows a State() call should instead directly access the state.
from what I can see that's in both `server/container_list.go` and `server/container_status.go`

internal/oci/container.go
```diff
@@ -269,9 +278,15 @@ func (c *Container) Metadata() *pb.ContainerMetadata {
 
 // State returns the state of the running container
 func (c *Container) State() *ContainerState {
```
mrunalp, line 280: @tedyu can you break this change to a separate PR? I want us to land the changes in this PR in stages.

tedyu: I will keep Container#State() as is in this PR.

tedyu: @mrunalp 
Can you take another look ?

internal/oci/container.go
tedyu: Noted.
State() is reverted to current form.

I will keep this in mind if there is follow on PR which changes State() call.

tedyu:
w.r.t. crun e2e, I looked at the logs from https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3646/test_pull_request_crio_e2e_crun_fedora/3820/artifacts/journals/ but  didn't see much clue.

tedyu:
From kubelet.log under https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3646/test_pull_request_crio_e2e_crun_fedora/3821/artifacts/ :
```
I0429 21:22:18.162288   28703 factory.go:201] Factory "raw" can handle container "/user.slice/user-0.slice/session-c235.scope", but ignoring.
```
I searched for 'lice\/session-c235.scope' but there was no other log which didn't contain ignoring.

tedyu:
@haircommander @mrunalp 
Tests passed.
Please take another look.

internal/oci/container.go
tedyu: @mrunalp
Do you have other comment ?

rphillips:
lgtm

server/container_list.go
```diff
@@ -79,7 +79,7 @@ func (s *Server) ListContainers(ctx context.Context, req *pb.ListContainersReque
 			continue
 		}
 		podSandboxID := ctr.Sandbox()
-		cState := ctr.StateNoLock()
+		cState := ctr.State()
```
rphillips, line 82: Do we need these changes to State()? I don't think we do.

tedyu: Addressed

tedyu:
@haircommander @mrunalp 
I have reduced the scope of the PR.
Please take another look.

internal/oci/container.go
```diff
@@ -160,7 +160,14 @@ func (c *Container) FromDisk() error {
...
 	dec := json.NewDecoder(jsonSource)
-	return dec.Decode(c.state)
+	var state ContainerState
+	if err := dec.Decode(&state); err != nil {
+		return err
+	}
+	c.opLock.Lock()
+	c.state = &state
+	c.opLock.Unlock()
+	return nil
```
TomSweeneyRedHat, line 170: IDK dec.Decode() and couldn't find it quickly to verify, but I think the value in state can change in that right?  If so, should the c.opLock.Lock() be above that call?

tedyu: state is local variable.
It is expected that its value changes after the decoding.

TomSweeneyRedHat: DOH, that's right, the trick is it's a local variable and safe.  Sorry for the churn, thx @tedyu

TomSweeneyRedHat:

haircommander:
LGTM @mrunalp happy with it?

tedyu:
We can see how #3729 (a superset of this PR) goes.

haircommander:
it seems the PR description doesn't match the contents of the PR

PR #3647, opened 2020-04-23
[1.18] crio wipe: add version-file-persist

haircommander:
#### What type of PR is this?

/kind feature

note: we also don't wipe anything when we have temp but not persist. this is to address the update case where we update from an older version of 1.18 that didn't specify a persisted version file. The assumption is if a persisted version file doesn't exist, but the temporary dir does, then we should not update. In the future (when we can more safely assume there will be a persisted version file) this case may mean we should wipe everything.

Also, update the crio wipe tests to test different scenarios

```release-note
add --version-file-persist, a place to put the version file in persistent storage. Now, crio wipe wipes containers if --version-file is not present (presumably it is on temporary storage), and wipes images if both --version-file and --version-file-persist are out of date (presumably there has been an upgrade of cri-o's minor version
```


docs/crio.8.md
```diff
@@ -289,7 +290,9 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--version, -v**: print the version
 
-**--version-file**="": Location for CRI-O to lay down the version file (default: /var/run/crio/version)
+**--version-file**="": Location for CRI-O to lay down the temporary version file (default: /var/run/crio/version)
+
+**--version-file-persist**="": Location for CRI-O to lay down the persistent version file (default: /var/run/crio/version)
```
TomSweeneyRedHat, line 295: It would be good to add a little verbiage indicating why these files exist, how they are used, and if necessary, the contents.

docs/crio.8.md
```diff
@@ -289,7 +290,9 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--version, -v**: print the version
 
-**--version-file**="": Location for CRI-O to lay down the version file (default: /var/run/crio/version)
+**--version-file**="": Location for CRI-O to lay down the temporary version file. It is used to check if crio wipe should wipe containers, which should always happen on a node reboot (default: /var/run/crio/version)
+
+**--version-file-persist**="": Location for CRI-O to lay down the persistent version file. It is used to check if crio wipe should wipe images, which should only happen when CRI-O has been upgraded (default: /var/run/crio/version)
```
TomSweeneyRedHat, line 295: :+1:  TY!

TomSweeneyRedHat:
LGTM assuming happy tests

haircommander:
```
not ok 65 privileged ctr -- check for rw mounts
# (in test file ./ctr.bats, line 1574)
#   `[[ "$output" =~ "/sys/fs/cgroup tmpfs" ]]' failed
# 0
```

CONFUSING

PR #3648, opened 2020-04-23
crio wipe: add version-file-persist

haircommander:
#### What type of PR is this?

/kind feature

note: we also don't wipe anything when we have temp but not persist. this is to address the update case where we update from an older version of 1.18 that didn't specify a persisted version file. The assumption is if a persisted version file doesn't exist, but the temporary dir does, then we should not update. In the future (when we can more safely assume there will be a persisted version file) this case may mean we should wipe everything.

Also, update the crio wipe tests to test different scenarios

```release-note
add --version-file-persist, a place to put the version file in persistent storage. Now, crio wipe wipes containers if --version-file is not present (presumably it is on temporary storage), and wipes images if both --version-file and --version-file-persist are out of date (presumably there has been an upgrade of cri-o's minor version
```


docs/crio.conf.5.md
```diff
@@ -45,7 +45,10 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
   The default log directory where all logs will go unless directly specified by the kubelet. The log directory specified must be an absolute directory.
 
 **version_file**="/var/run/crio/version"
-  Location for crio to lay down the version file.
+  Location for crio to lay down the temporary version file.
+
+**version_file_persist**="/var/lib/crio/version"
+  Location for crio to lay down the persistent version file.
```
TomSweeneyRedHat, line 51: As noted in the 1.18 review, a little more background would be nice.

haircommander:
merge me plz @umohnani8 @mrunalp

haircommander:
can I have my label back plz @mrunalp @umohnani8

test/crio-wipe.bats
```diff
@@ -0,0 +1,136 @@
...
+
+function teardown() {
+	cleanup_test
+	run_podman_with_args stop -a
+	run_podman_with_args rm -fa
+}
+
+# run crio_wipe calls crio_wipe and tests it succeeded
+function run_crio_wipe() {
+	run $CRIO_BINARY --config "$CRIO_CONFIG" wipe
```
saschagrunert, line 31:
```suggestion
	run $CRIO_BINARY_PATH --config "$CRIO_CONFIG" wipe
```

This :point_up: should fix the static binary CI

haircommander: AH you're the best thanks

PR #3649, opened 2020-04-23
test: update digests and test

haircommander:
#### What type of PR is this?

/kind failing-test

round 2!

Before, the image pull and list by manifest list tag test was failing because of the wrong digest

because this has happened twice today, also update the test to inform us the value it should be

```release-note

```


haircommander:
/retest

seems like a flake

PR #3654, opened 2020-04-24
[1.17] Add info logs where needed

umohnani8:
#### What type of PR is this?

/kind cleanup

Add info logs to all the defers in container and sandbox create.
Add logs for when we get a request from kubelet to stop and remove
a container and sandbox.

```release-note

```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -568,6 +568,7 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 
 	defer func() {
 		if err != nil {
+			log.Infof(ctx, "createCtr: releasing container name %s", containerName)
 			s.ReleaseContainerName(containerName)
 		}
 	}()
@@ -578,6 +579,7 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 	}
 	defer func() {
 		if err != nil {
+			log.Infof(ctx, "createCtr: deleting container %s from storage", containerID)
 			err2 := s.StorageRuntimeServer().DeleteContainer(containerID)
 			if err2 != nil {
 				log.Warnf(ctx, "Failed to cleanup container directory: %v", err2)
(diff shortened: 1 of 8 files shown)


umohnani8:
@mrunalp @haircommander PTAL

umohnani8:
Do we want this in 1.16 as well?

haircommander:
we did switch log level to info in 4.3 so I think so

TomSweeneyRedHat:
LGTM assuming happy tests

server/container_create.go
```diff
@@ -546,6 +547,13 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 		return nil, fmt.Errorf("CreateContainer failed as the sandbox was stopped: %v", sbID)
 	}
 
+	defer func() {
```
mrunalp, line 550: Can we explicitly test this by adding a sleep to network start before merging?

server/sandbox_run_linux.go
```diff
@@ -53,6 +54,14 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 
 	// we need to fill in the container name, as it is not present in the request. Luckily, it is a constant.
 	log.Infof(ctx, "attempting to run pod sandbox with infra container: %s%s", translateLabelsToDescription(req.GetConfig().GetLabels()), leaky.PodInfraContainerName)
+
+	defer func() {
```
mrunalp, line 58: Actually this one can be tested by adding a sleep > 4 minutes in network start. We should test before merging this PR.

umohnani8: This was working fine, but I moved it back to the end of the function so that we return an error if the context is cancelled or deadline exceeded, making sure all the clean up happens when this occurs. Also got rid of the stack trace as it wasn't pointing to where the context deadline was exceeded - it would just point to when we check the context error, so I didn't think it was really helpful.

umohnani8:
Rebased

umohnani8:
Closing in favour of https://github.com/cri-o/cri-o/pull/3679

PR #3659, opened 2020-04-24
[1.17] Pass context from caller to ocicni methods

haircommander:
What type of PR is this?

/kind bug

What this PR does / why we need it:

Update ocicni vendored code to get new "WithContext" methods.
Pass a child context from the one supplied to cri-o calls to ocicni, with half of the timeout for that respective request

Which issue(s) this PR fixes:
Special notes for your reviewer:
Does this PR introduce a user-facing change?

None


haircommander:
manually picked https://github.com/cri-o/cri-o/pull/3661 into here

PR #3661, opened 2020-04-24
test: check for rw mounts

haircommander:
#### What type of PR is this?

/kind failing-test

the test is failing and this fixes it and I'm not sure why it was passing at all. today is confusing

```release-note

```

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -1566,7 +1566,9 @@ function wait_until_exit() {
 	run crictl start "$ctr_id"
 	[ "$status" -eq 0 ]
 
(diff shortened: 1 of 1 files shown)


haircommander:
ok so I think this is coming from a change in behavior in runc. as crun used to pass and now fails, whereas integration_{fedora,rhel} used to fail and now pass

haircommander:
I am trying a different approach to fix CI

haircommander:
if this passes we should merge to unblock CI and I'll open an issue to figure out what the behavior should be

PR #3663, opened 2020-04-25
Fix Linkmode path resolution

CertainLach:
In case if crio is located and run from `$PATH`, `Args[0]` is just `crio`, so `filepath.Abs` returns `${PWD}/crio`
os.Executables() should be used instead

#### What type of PR is this?

/kind bug

```release-note
None
```


saschagrunert:
Ah thank you for the fix!

PR #3668, opened 2020-04-27
Add upgrade path integration tests

saschagrunert:
#### What type of PR is this?

/kind cleanup

Adding upgrade path tests to ensure correct container restore between CRI-O versions.

None
None

```release-note
None
```

### DIFF
--- a/test/testdata/sandbox_config_apparmor.json
+++ b/test/testdata/sandbox_config_apparmor.json
@@ -1,7 +1,7 @@
 {
   "metadata": {
     "name": "podsandbox1",
-    "uid": "redhat-test-crio",
+    "uid": "redhat-test-crio-apparmor",
     "namespace": "redhat.test.crio",
     "attempt": 1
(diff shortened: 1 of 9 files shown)


haircommander:
neat idea! maybe we should build off of the latest in the release branches?

saschagrunert:
> neat idea! maybe we should build off of the latest in the release branches?

We can! :)

Do you suggest do build it from source instead of using the static binaries?

haircommander:
> > neat idea! maybe we should build off of the latest in the release branches?
> 
> We can! :)
> 
> Do you suggest do build it from source instead of using the static binaries?

I think from the latest tagged release on each branch would be sufficient

saschagrunert:
/retest

I'm pretty sure that there is an issue but I cannot access any logs :-/

saschagrunert:
I have to reopen this PR since I changed the forks.

PR #3677, opened 2020-04-27
[1.16] Disable the extra-source when building the rpm.

umohnani8:

#### What type of PR is this?

/kind failing-test

```release-note

```

### DIFF
--- a/hack/build-rpms.sh
+++ b/hack/build-rpms.sh
@@ -31,7 +31,7 @@ tar czf "${rpm_tmp_dir}/SOURCES/${OS_RPM_NAME}-test.tar.gz" \
 	.
 cp -r "${ci_data}/." "${rpm_tmp_dir}/SOURCES"
 
-echo "${OS_GIT_COMMIT}"
+sed -i 's/vault.centos.org\/centos\/$releasever/vault.centos.org\/centos\/7.7.1908/g' /etc/yum.repos.d/CentOS-Sources.repo
 yum-builddep -y "${OS_RPM_SPECFILE}"
 
 rpmbuild -ba "${OS_RPM_SPECFILE}" \


hack/build-rpms.sh
```diff
@@ -31,7 +31,7 @@ tar czf "${rpm_tmp_dir}/SOURCES/${OS_RPM_NAME}-test.tar.gz" \
 	.
 cp -r "${ci_data}/." "${rpm_tmp_dir}/SOURCES"
 
-echo "${OS_GIT_COMMIT}"
+yum-config-manager --disable extras-source
```
lsm5, line 34: @umohnani8 Could you delete this line and add this instead: `sed -i 's/vault.centos.org\/centos\/$releasever/vault.centos.org\/centos\/7.7.1908/g' /etc/yum.repos.d/CentOS-Sources.repo` . This will ensure the URL is correct is fetched.

umohnani8: Done, thanks!

hack/build-rpms.sh
```diff
@@ -31,7 +31,7 @@ tar czf "${rpm_tmp_dir}/SOURCES/${OS_RPM_NAME}-test.tar.gz" \
 	.
 cp -r "${ci_data}/." "${rpm_tmp_dir}/SOURCES"
 
-echo "${OS_GIT_COMMIT}"
+sed -i 's/vault.centos.org\/centos\/$releasever/vault.centos.org\/centos\/7.7.1908/g' /etc/yum.repos.d/CentOS-Sources.r
```
lsm5, line 34: @umohnani8 err, doesn't look like the file name extension got mentioned correctly.

lsm5: should be `.repo`

umohnani8: yup, typo - fixed!

PR #3678, opened 2020-04-27
Remove unnecessary error wrapping for runtimeVM#StartContainer

tedyu:
#### What type of PR is this?

/kind cleanup

The err is filled by the goroutine, at the time runtimeVM#StartContainer returns, err would be nil.
This PR removes the ineffective error wrapping.

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -243,15 +243,18 @@ func (r *runtimeVM) StartContainer(c *Container) error {
 
 	// Spawn a goroutine waiting for the container to terminate. Once it
 	// happens, the container status is retrieved to be updated.
-	var err error
 	go func() {
-		_, err = r.wait(r.ctx, c.ID(), "")
+		_, err := r.wait(r.ctx, c.ID(), "")
 		if err == nil {
-			err = r.UpdateContainerStatus(c)
+			if err1 := r.UpdateContainerStatus(c); err1 != nil {
+				logrus.Warningf("error updating container status %v", err1)
+			}
+		} else {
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_vm.go
```diff
@@ -243,15 +243,14 @@ func (r *runtimeVM) StartContainer(c *Container) error {
 
 	// Spawn a goroutine waiting for the container to terminate. Once it
 	// happens, the container status is retrieved to be updated.
-	var err error
 	go func() {
-		_, err = r.wait(r.ctx, c.ID(), "")
+		_, err := r.wait(r.ctx, c.ID(), "")
```
haircommander, line 247: I feel like we shouldn't just swallow this error...

tedyu: I added log for non-nil error

internal/oci/runtime_vm.go
```diff
@@ -243,15 +243,18 @@ func (r *runtimeVM) StartContainer(c *Container) error {
...
```
haircommander, line 247: another nit: personally I prefer multiple returns over if/elses
```
if _, err := r.wait(r.ctx, c.ID(), ""); err != nil {
    logrus.Warnf(...)
    return
}
if err := r.UpdateContainerStatus(c); err != nil {
    logrus.Warnf(...)
}
```
that's really a style nit though, feel free to ignore :smile:

tedyu:
@saschagrunert 
Can you give /lgtm ?

Thanks

tedyu:
@haircommander 
Can this be merged ?

Thanks

PR #3686, opened 2020-04-28
Return an error if context has been cancelled or deadline exceeded

umohnani8:
#### What type of PR is this?

 /kind cleanup

This will ensure that we do all the clean up when we timeout.

```release-note

```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"runtime/debug"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


server/container_create.go
```diff
@@ -548,13 +547,6 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 		return nil, fmt.Errorf("CreateContainer failed as the sandbox was stopped: %v", sbID)
 	}
 
-	defer func() {
-		if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
-			log.Infof(ctx, "createCtr: context was either canceled or the deadline was exceeded: %v", ctx.Err())
-			debug.PrintStack()
```
haircommander, line 554: do we not want the PrintStack() anymore?

umohnani8: nope, it doesn't give you any helpful notification. the point for it was it would point us to where the context deadline exceeded, but it just points to where we check for the context error.

umohnani8:
@haircommander @mrunalp can I please get a lgtm

PR #3689, opened 2020-04-29
Remove crio-wipe and crio-shutdown systemd units from bundle

saschagrunert:
#### What type of PR is this?

/kind cleanup

Both systemd units are not required and therefore should not be part of
the static bundle.

None
None

```release-note
- Removed `crio-wipe.service` and `crio-shutdown.service` systemd units from the static bundle since they are not required
```


TomSweeneyRedHat:

PR #3698, opened 2020-04-29
Close channels in runtimeOCI

tedyu:
#### What type of PR is this?

/kind bug

This PR closes receiveStdout and stdinDone in runtimeOCI#AttachContainer.
Also closes errorCh in runtimeOCI#ReopenContainerLog

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -863,6 +863,7 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
 	if outputStream != nil || errorStream != nil {
 		go func() {
 			receiveStdout <- redirectResponseToOutputStreams(outputStream, errorStream, conn)
+			close(receiveStdout)
 		}()
 	}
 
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_oci.go
```diff
@@ -878,6 +878,8 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
 		stdinDone <- err
 	}()
 
+	close(receiveStdout)
```
giuseppe, line 881: shouldn't this be done as part of the cleanup from the goroutines using these channels?

Otherwise it can cause a write to a closed channel.

tedyu: @giuseppe 
Your comment has been addressed.

tedyu:
@haircommander 
Can you approve this ?

internal/oci/runtime_oci.go
giuseppe: thanks, LGTM

PR #3707, opened 2020-04-30
Automatically label containers running systemd with the correct label

rhatdan:
/kind Feature

Newer versions of container-selinux, container-selinux-2.132.0 or newer,
supply a `container_init_t` label.  If CRI-O is running systemd or init inside
of the container, then the container will require different SELinux privs
to run the container.

By using the new SELinux label, we can run ordinary containers with a tighter
selinux policy then those running the init system, makeing the overlay system
more secure.

The eliminates the need to turn on the container_manage_cgroup SELinux boolean.

Ie. no need to execute

```
setsebool -P container_manage_cgroup 1
```

Any longer. On systems without the updated container-selnux package, the containers will still
attempt to run the standard container type `container_t`, and still require the boolean.

#### What type of PR is this?
/kind feature

```release-note

```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/buildah/pkg/secrets"
 	"github.com/containers/libpod/pkg/annotations"
 	"github.com/containers/libpod/pkg/rootless"
+	"github.com/containers/libpod/pkg/selinux"
 	createconfig "github.com/containers/libpod/pkg/spec"
 	"github.com/containers/storage/pkg/mount"
 	"github.com/cri-o/cri-o/internal/lib"
@@ -398,6 +399,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 	if err != nil {
 		return nil, err
 	}
+
 	mountLabel := containerInfo.MountLabel
 	var processLabel string
 	if !privileged {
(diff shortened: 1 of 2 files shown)


rhatdan:
@umohnani8 PTAL

umohnani8:
LGTM
I will test it out on a cluster as well

rhatdan:
Awesome, hopefully it works,  Make sure you have a newer container-selinux.

umohnani8:
@rhatdan tried on a cluster with container-selinuux v2.132.0 and with crictl inspect still seeing the label set as `"selinuxLabel": "system_u:system_r:container_t:s0:c86,c226"`
I added a print statement right after the label init too and see this in the logs 
```Apr 30 20:44:19 ip-10-0-143-64 crio[45156]: ---the process label is-----: system_u:system_r:container_t:s0:c86,c226```

**Comment by rhatdan** [comment]
On these systems do you have a /usr/share/containers/selinux/context file?

**Comment by umohnani8** [comment]
@rhatdan nope, I don't see that file on the nodes

**Comment by rhatdan** [comment]
Then they don't have the latest container-selinux package installed.

container-selinux-2.125.2 or later
```
commit ae0720d6ac32866b023babeff75e766870951f8f (tag: v2.125.2)
Author: Daniel J Walsh <dwalsh@redhat.com>
Date:   Mon Mar 23 15:29:10 2020 -0400

    Add container_contexts file and man page to be shipped with policy
    
```

rhatdan:
@jnovy Can you update the container-selinux package available for Openshift, to the latest package.
container-selinux-2.132.0

umohnani8:
@rhatdan I did do a `rpm-ostree override replace` for the container-selinux package to update it to v2.132.0 - but I don't see that file, I guess installing it doesn't add the new file.

mrunalp:
I think we should add an integration test for this to check the label for systemd.

umohnani8:
Was able to successfully test this on a cluster and it works! We can get this moving along :D

rhatdan:
SGTM

haircommander:
@rhatdan can you run `make vendor` with go 1.14 (or in a go1.14 container)? it seems the tree is dirty

rhatdan:
@umohnani8 or @haircommander Please take over this PR.  My environment is not working well.

haircommander:
yoink

PR #3710, opened 2020-04-30
Delete container Id from ctrIDIndex if podIDIndex.Add fails

tedyu:
#### What type of PR is this?

/kind bug

In ContainerServer#LoadSandbox, if podIDIndex.Add() call fails, we should delete container Id from ctrIDIndex.

```release-note
None
```

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -308,6 +308,13 @@ func (c *ContainerServer) LoadSandbox(id string) (retErr error) {
 	if err := c.ctrIDIndex.Add(scontainer.ID()); err != nil {
 		return err
 	}
+	defer func() {
+		if retErr != nil {
+			if err1 := c.ctrIDIndex.Delete(scontainer.ID()); err1 != nil {
+				logrus.Warnf("could not delete container ID %s: %v", scontainer.ID(), err1)
+			}
+		}
+	}()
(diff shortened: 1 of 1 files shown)


tedyu:
@saschagrunert 
Can you take a look ?

internal/lib/container_server.go
```diff
@@ -309,6 +309,9 @@ func (c *ContainerServer) LoadSandbox(id string) (retErr error) {
 		return err
 	}
 	if err := c.podIDIndex.Add(id); err != nil {
+		if err1 := c.ctrIDIndex.Delete(scontainer.ID()); err1 != nil {
```
mrunalp, line 312: Can we use a defer instead?

tedyu: Since this is toward the end of the method, I added the Delete() call directly.

If you feel there could be more code added after the if block in the future, I will change this to defer.

tedyu: @mrunalp 
Your comment has been addressed.

haircommander:
/approve

LGTM

tedyu:
@saschagrunert 
Can you take a look ?

saschagrunert:
@tedyu does this fix a user-facing bug? If so, please change your release note in the `release-note` block of the PR body.

tedyu:
This doesn't seem to be user-facing bug.

PR #3712, opened 2020-05-01
Delete container Id only when impl.DeleteContainer(c) passes

tedyu:
#### What type of PR is this?

/kind bug

In Runtime#DeleteContainer, we should only delete the container Id from runtimeImplMap if the call to impl.DeleteContainer(c) doesn't return non-nil error.

```release-note
None
```


haircommander:
/approve

LGTM

PR #3713, opened 2020-05-01
version: small tweaks

haircommander:
#### What type of PR is this?

> /kind bug

omitempty on String(), as we already have set omitempty on the JSON version, but the String() method left in empty strings
also return empty string when we fail to check the ldd status of the binary (as opposed to setting the field as the error)

```release-note

```


rhatdan:
LGTM
/test integration_rhel

Issue #3717, opened 2020-05-04
Debian Buster Default Kernel Always Failed with `open /sys/fs/cgroup/hugetlb: no such file or directory`


hswong3i:
**Description**

(Related to #3540?)

When running `kubeadm init` with CRI-O v1.18.0 + Debian Buster, always failed due to default kernel `4.19.0-6-amd64` + `CONFIG_CGROUP_HUGETLB` is not set by default:
```
root@aepu7zu4aih9-2:~# uname -a
Linux aepu7zu4aih9-2 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64 GNU/Linux
root@aepu7zu4aih9-2:~# cat /boot/config-4.19.0-6-amd64 | grep CONFIG_CGROUP_HUGETLB
# CONFIG_CGROUP_HUGETLB is not set
```

In case replace Kernel from Ubuntu Mainline Kernel PPA (e.g. https://kernel.ubuntu.com/~kernel-ppa/mainline/v5.6.7/) and reboot, above error message gone and all deployment works as expected, with `CONFIG_CGROUP_HUGETLB` enabled:
```
root@aepu7zu4aih9-2:~# uname -a
Linux aepu7zu4aih9-2 5.6.7-050607-generic #202004230933 SMP Thu Apr 23 09:35:28 UTC 2020 x86_64 GNU/Linux
root@aepu7zu4aih9-2:~# cat /boot/config-5.6.7-050607-generic | grep CONFIG_CGROUP_HUGETLB
CONFIG_CGROUP_HUGETLB=y
```

**Steps to reproduce the issue:**
1. Running a Debian Buster box with Vagrant, and ssh into it:
```
vagrant init generic/debian10
vagrant up
vagrant ssh
```

2. Run my Ansible Playbook for testing (which will failed with default kernel):
```
sudo su -
apt-get update
apt-get update && sudo apt-get install -y ca-certificates curl gcc git iproute2 pwgen python3 python3-dev sudo
mkdir -p ~/.ansible/roles
cd ~/.ansible/roles
git clone =b develop https://github.com/alvistack/ansible-role-kube_master.git kube_master
cd ~/.ansible/roles/kube_master
curl -skL https://bootstrap.pypa.io/get-pip.py | sudo -H python3
sudo -H pip3 install --upgrade --ignore-installed --requirement requirements.txt
molecule converge -s default
```

[...]

PR #3721, opened 2020-05-04
[1.18] make lack of hugetlb support non-fatal for cgroupv1

haircommander:
#### What type of PR is this?

/kind bug

lack of hugetlb support in the kernel should not be a fatal error in cgroupv1.

this PR patches the error in 1.18. https://github.com/cri-o/cri-o/pull/3565 will fix it in master

fixes https://github.com/cri-o/cri-o/issues/3717

```release-note
Fix a bug where CRI-O could not start a container if CONFIG_CGROUP_HUGETLB was not set in the kernel
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -47,7 +47,6 @@ const podTerminationGracePeriodLabel = "io.kubernetes.pod.terminationGracePeriod
 var (
 	_cgroupv1HasHugetlbOnce sync.Once
 	_cgroupv1HasHugetlb     bool
-	_cgroupv1HasHugetlbErr  error
 	_cgroupv2HasHugetlbOnce sync.Once
 	_cgroupv2HasHugetlb     bool
 	_cgroupv2HasHugetlbErr  error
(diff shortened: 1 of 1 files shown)


server/container_create_linux.go
```diff
@@ -58,17 +57,15 @@ var (
 
 // cgroupv1HasHugetlb returns whether the hugetlb controller is present on
 // cgroup v1.
-func cgroupv1HasHugetlb() (bool, error) {
+func cgroupv1HasHugetlb() bool {
 	_cgroupv1HasHugetlbOnce.Do(func() {
 		if _, err := ioutil.ReadDir("/sys/fs/cgroup/hugetlb"); err != nil {
-			_cgroupv1HasHugetlbErr = errors.Wrap(err, "readdir /sys/fs/cgroup/hugetlb")
 			_cgroupv1HasHugetlb = false
```
TomSweeneyRedHat, line 63: IDK, but might it be useful to have a debug log print out here if we set this to false?

TomSweeneyRedHat:
One questions, otherwise LGTM with happy tests.

server/container_create_linux.go
haircommander: good idea, added

haircommander: ach, nevermind, the function call flow doesn't make this simple or clear. I think we're okay with out.

haircommander:
PTAL @saschagrunert @mrunalp @umohnani8

PR #3723, opened 2020-05-05
Avoid unnecessary locking on runtimeImplMap

tedyu:
#### What type of PR is this?

/kind bug

In Runtime#DeleteContainer, when the impl was not found in the runtimeImplMap, there is no need to take write lock on the map at the end of the method.
Assuming the get from runtimeImplMap at beginning of DeleteContainer() returns nothing, the deferred func shouldn't remove map entry someone else puts in the map.

```release-note
None
```

### DIFF
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -302,18 +302,22 @@ func (r *Runtime) StopContainer(ctx context.Context, c *Container, timeout int64
 
 // DeleteContainer deletes a container.
 func (r *Runtime) DeleteContainer(c *Container) (err error) {
-	impl, err := r.RuntimeImpl(c)
-	if err != nil {
-		return err
-	}
-
-	defer func() {
-		if err == nil {
-			r.runtimeImplMapMutex.Lock()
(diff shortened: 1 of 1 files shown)


internal/oci/oci.go
```diff
@@ -302,19 +302,21 @@ func (r *Runtime) StopContainer(ctx context.Context, c *Container, timeout int64
...
 // DeleteContainer deletes a container.
 func (r *Runtime) DeleteContainer(c *Container) (err error) {
-	impl, err := r.RuntimeImpl(c)
-	if err != nil {
-		return err
+	r.runtimeImplMapMutex.RLock()
+	impl, ok := r.runtimeImplMap[c.ID()]
+	r.runtimeImplMapMutex.RUnlock()
+	if !ok {
+		impl, _ = r.newRuntimeImpl(c)
```
saschagrunert, line 309: Hm, what are we going to do with that error?

tedyu: Check added on the error return.

PR #3741, opened 2020-05-08
[1.18] Makefile: include -nobuild install targets

lsm5:
Distro packaging often uses non-default build and linker flags.
The current Makefile cannot be used in the package build process as it
will end up rebuilding the binaries with the default flags.

This commit introduces install targets which do not depend on the build
targets. Distro packages should prefer these if they want to use any
non-default flags.

NOTE: This commit effectively leaves prior targets unchanged, so users
won't notice any difference.

(cherry picked from commit c6233a2b4b26daa21195078f0883d1cbfc56ba49)

#### What type of PR is this?

/kind dependency-change

Allows packaging to reuse upstream Makefile

None

None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -431,23 +431,29 @@ bundle-test:
 
 install: .gopathok install.bin install.man install.completions install.systemd install.config
 
-install.bin: binaries
+install.bin-nobuild:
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM assuming happy tests

PR #3749, opened 2020-05-11
Remove socat runtime dependency

saschagrunert:
#### What type of PR is this?

/kind feature

This is the analogous implementation to
https://github.com/containerd/cri/pull/1470 to remove the socat runtime
dependency from CRI-O.

None

None

```release-note
- Removed `socat` runtime dependency which was needed for pod port forwarding
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -11,13 +11,14 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
-	"strings"
 	"syscall"
 	"time"
 
+	"github.com/containernetworking/plugins/pkg/ns"
 	conmonconfig "github.com/containers/conmon/runner/config"
 	"github.com/containers/storage/pkg/pools"
 	"github.com/cri-o/cri-o/internal/findprocess"
+	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/pkg/config"
 	"github.com/cri-o/cri-o/utils"
 	"github.com/fsnotify/fsnotify"
@@ -255,9 +256,9 @@ func prepareExec() (pidFile, parentPipe, childPipe *os.File, err error) {
 	return
 }
 
-func parseLog(log []byte) (stdout, stderr []byte) {
+func parseLog(l []byte) (stdout, stderr []byte) {
 	// Split the log on newlines, which is what separates entries.
-	lines := bytes.SplitAfter(log, []byte{'\n'})
+	lines := bytes.SplitAfter(l, []byte{'\n'})
 	for _, line := range lines {
 		// Ignore empty lines.
 		if len(line) == 0 {
@@ -898,73 +899,88 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
 	return nil
 }
 
-// PortForwardContainer forwards the specified port provides statistics of a container.
-func (r *runtimeOCI) PortForwardContainer(c *Container, port int32, stream io.ReadWriter) error {
-	emptyStreamOnError := true
-	defer func() {
-		if emptyStreamOnError {
-			go func() {
-				if _, copyError := pools.Copy(ioutil.Discard, stream); copyError != nil {
-					logrus.Errorf("error closing port forward stream after other error: %v", copyError)
-				}
-			}()
+// PortForwardContainer forwards the specified port into the provided container.
+func (r *runtimeOCI) PortForwardContainer(ctx context.Context, c *Container, netNsPath string, port int32, stream io.ReadWriteCloser) error {
+	log.Infof(ctx,
+		"Starting port forward for %s in network namespace %s", c.ID(), netNsPath,
+	)
+
+	// Adapted reference implementation:
+	// https://github.com/containerd/cri/blob/8c366d/pkg/server/sandbox_portforward_unix.go#L65-L120
+	if err := ns.WithNetNSPath(netNsPath, func(_ ns.NetNS) error {
+		defer stream.Close()
+
+		// TODO: hardcoded to tcp4 because localhost resolves to ::1 by default
+		// if the system has IPv6 enabled. However, not all applications are
+		// listening on the IPv6 localhost address. Theoretically happy
+		// eyeballs will try IPv6 first and fallback to IPv4 but resolving
+		// localhost doesn't seem to return and IPv4 address always, thus
+		// failing the connection.
+		conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))
+		if err != nil {
+			return errors.Wrapf(err, "dialing %d", port)
 		}
-	}()
-	containerPid := c.State().Pid
-	socatPath, lookupErr := exec.LookPath("socat")
-	if lookupErr != nil {
-		return fmt.Errorf("unable to do port forwarding: socat not found")
+		defer conn.Close()
+
+		errCh := make(chan error, 2)
+
+		debug := func(format string, args ...interface{}) {
+			log.Debugf(ctx, fmt.Sprintf(
+				"PortForward (id: %s, port: %d): %s", c.ID(), port, format,
+			), args...)
+		}
+
+		// Copy from the the namespace port connection to the client stream
+		go func() {
+			debug("copy data from container to client")
+			_, err := io.Copy(stream, conn)
+			errCh <- err
+		}()
+
+		// Copy from the client stream to the namespace port connection
+		go func() {
+			debug("copy data from client to container")
+			_, err := io.Copy(conn, stream)
+			errCh <- err
+		}()
+
+		// Wait until the first error is returned by one of the connections we
+		// use errFwd to store the result of the port forwarding operation if
+		// the context is cancelled close everything and return
+		var errFwd error
+		select {
+		case errFwd = <-errCh:
+			debug("stop forwarding in direction: %v", errFwd)
+		case <-ctx.Done():
+			debug("cancelled: %v", ctx.Err())
+			return ctx.Err()
+		}
+
+		// give a chance to terminate gracefully or timeout
+		const timeout = time.Second
+		select {
+		case e := <-errCh:
+			if errFwd == nil {
+				errFwd = e
+			}
+			debug("stopped forwarding in both directions")
+
+		case <-time.After(timeout):
+			debug("timed out waiting to close the connection")
+
+		case <-ctx.Done():
+			debug("cancelled: %v", ctx.Err())
+			errFwd = ctx.Err()
+		}
+
+		return errFwd
+	}); err != nil {
+		return errors.Wrapf(
+			err, "port forward into network namespace %q", netNsPath,
+		)
 	}
 
-	args := []string{"-t", fmt.Sprintf("%d", containerPid), "-n", socatPath, "-", fmt.Sprintf("TCP4:localhost:%d", port)}
-
-	nsenterPath, lookupErr := exec.LookPath("nsenter")
-	if lookupErr != nil {
-		return fmt.Errorf("unable to do port forwarding: nsenter not found")
-	}
-
-	commandString := fmt.Sprintf("%s %s", nsenterPath, strings.Join(args, " "))
-	logrus.Debugf("executing port forwarding command: %s", commandString)
-
-	command := exec.Command(nsenterPath, args...)
-	if v, found := os.LookupEnv("XDG_RUNTIME_DIR"); found {
-		command.Env = append(command.Env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", v))
-	}
-	command.Stdout = stream
-
-	stderr := new(bytes.Buffer)
-	command.Stderr = stderr
-
-	// If we use Stdin, command.Run() won't return until the goroutine that's copying
-	// from stream finishes. Unfortunately, if you have a client like telnet connected
-	// via port forwarding, as long as the user's telnet client is connected to the user's
-	// local listener that port forwarding sets up, the telnet session never exits. This
-	// means that even if socat has finished running, command.Run() won't ever return
-	// (because the client still has the connection and stream open).
-	//
-	// The work around is to use StdinPipe(), as Wait() (called by Run()) closes the pipe
-	// when the command (socat) exits.
-	inPipe, err := command.StdinPipe()
-	if err != nil {
-		return fmt.Errorf("unable to do port forwarding: error creating stdin pipe: %v", err)
-	}
-	var copyError error
-	go func() {
-		emptyStreamOnError = false
-		_, copyError = pools.Copy(inPipe, stream)
-		inPipe.Close()
-	}()
-
-	runErr := command.Run()
-
-	if copyError != nil {
-		return copyError
-	}
-
-	if runErr != nil {
-		return fmt.Errorf("%v: %s", runErr, stderr.String())
-	}
-
+	log.Infof(ctx, "Finished port forwarding for %q on port %d", c.ID(), port)
 	return nil
 }
 
--- a/server/container_portforward.go
+++ b/server/container_portforward.go
@@ -6,8 +6,9 @@ import (
 	"io/ioutil"
 
 	"github.com/containers/storage/pkg/pools"
+	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
-	"github.com/sirupsen/logrus"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
@@ -23,6 +24,8 @@ func (s *Server) PortForward(ctx context.Context, req *pb.PortForwardRequest) (r
 }
 
 func (s StreamService) PortForward(podSandboxID string, port int32, stream io.ReadWriteCloser) error {
+	ctx := log.AddRequestNameAndID(context.Background(), "PortForward")
+
 	// if we error in this function before Copying all of the content out of the stream,
 	// this stream will eventually get full, which causes leakages and can eventually brick CRI-O
 	// ref https://bugzilla.redhat.com/show_bug.cgi?id=1798193
@@ -31,7 +34,7 @@ func (s StreamService) PortForward(podSandboxID string, port int32, stream io.Re
 		if emptyStreamOnError && stream != nil {
 			go func() {
 				_, copyError := pools.Copy(ioutil.Discard, stream)
-				logrus.Errorf("error closing port forward stream after other error: %v", copyError)
+				log.Errorf(ctx, "error closing port forward stream after other error: %v", copyError)
 			}()
 		}
 	}()
@@ -41,11 +44,12 @@ func (s StreamService) PortForward(podSandboxID string, port int32, stream io.Re
 		return fmt.Errorf("PodSandbox with ID starting with %s not found: %v", podSandboxID, err)
 	}
 
-	c := s.runtimeServer.GetSandboxContainer(sandboxID)
-	if c == nil {
-		return fmt.Errorf("could not find container for sandbox %q", podSandboxID)
+	sb := s.runtimeServer.GetSandbox(sandboxID)
+	if sb == nil {
+		return fmt.Errorf("could not find sandbox %s", podSandboxID)
 	}
 
+	c := sb.InfraContainer()
 	if err := s.runtimeServer.Runtime().UpdateContainerStatus(c); err != nil {
 		return err
 	}
@@ -56,5 +60,12 @@ func (s StreamService) PortForward(podSandboxID string, port int32, stream io.Re
 	}
 
 	emptyStreamOnError = false
-	return s.runtimeServer.Runtime().PortForwardContainer(c, port, stream)
+
+	if sb.NetNsPath() == "" {
+		return errors.Errorf(
+			"network namespace path of sandbox %s is empty", sb.ID(),
+		)
+	}
+
+	return s.runtimeServer.Runtime().PortForwardContainer(ctx, c, sb.NetNsPath(), port, stream)
 }
--- a/internal/log/interceptors.go
+++ b/internal/log/interceptors.go
@@ -30,9 +30,7 @@ func StreamInterceptor() grpc.StreamServerInterceptor {
 		info *grpc.StreamServerInfo,
 		handler grpc.StreamHandler,
 	) error {
-		newCtx := addRequestName(
-			addRequestID(stream.Context()), info.FullMethod,
-		)
+		newCtx := AddRequestNameAndID(stream.Context(), info.FullMethod)
 		newStream := NewServerStream(stream)
 		newStream.NewContext = newCtx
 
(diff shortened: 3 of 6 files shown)


internal/oci/runtime_oci.go
```diff
@@ -898,73 +898,93 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
+// PortForwardContainer forwards the specified port into the provided container.
+func (r *runtimeOCI) PortForwardContainer(
+	c *Container, port int32, stream io.ReadWriteCloser,
+) error {
 	containerPid := c.State().Pid
-	socatPath, lookupErr := exec.LookPath("socat")
-	if lookupErr != nil {
-		return fmt.Errorf("unable to do port forwarding: socat not found")
-	}
+	netNsPath := fmt.Sprintf("/proc/%d/ns/net", containerPid)
```
haircommander, line 906: I'm not the hugest fan of using /proc here... we have the netNsPath saved in the sandbox, can you do something similar to https://github.com/cri-o/cri-o/pull/3474/commits/546e9817d85eeb27211706cb663e69e9d0348b10 instead?

internal/oci/runtime_oci.go
```diff
@@ -898,73 +898,93 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
+	logrus.Infof("Starting port forward in network namespace %s", netNsPath)
+
+	// TODO: use an intercepted context to anhance logging via the internal log
```
haircommander, line 909: nit: enhance

internal/oci/runtime_oci.go
haircommander: this could also be a follow up if you'd prefer

saschagrunert: Yep, fixed as suggested. :+1:

internal/oci/runtime_oci.go
saschagrunert: FIxed.

saschagrunert:
Two tests are failing:
https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3749/test_pull_request_crio_e2e_fedora/13020/

internal/oci/runtime_oci.go
```diff
@@ -898,73 +899,89 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
+		case errFwd = <-errCh:
+			debug("stop forwarding in direction: %v", errFwd)
+		case <-ctx.Done():
+			debug("cancelled: %v", ctx.Err())
+			return ctx.Err()
+		}
 
-	runErr := command.Run()
+		// give a chance to terminate gracefully or timeout
+		const timeout = 1 * time.Second
```
saschagrunert, line 961: Increasing the timeout to 1s fixes the failing e2e tests at least on my local machine.

aojea: great catch Sascha, I applied the same timeout to containerd and it fixes the e2e test.
Please, can you fix it in containerd too?

saschagrunert:
Referencing fix in containerd: https://github.com/containerd/cri/pull/1477

internal/oci/runtime_oci.go
saschagrunert: Yep, done :)

internal/oci/runtime_oci.go
```diff
@@ -898,73 +899,89 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
+	// Adapted reference implementation:
+	// https://github.com/containerd/cri/blob/8c366d/pkg/server/sandbox_portforward_unix.go#L65-L120
+	if err := ns.WithNetNSPath(nsPath, func(_ ns.NetNS) error {
+		defer stream.Close()
+
+		// TODO: hardcoded to tcp4 because localhost resolves to ::1 by default
+		// if the system has IPv6 enabled. Theoretically happy eyeballs will
+		// try IPv6 first and fallback to IPv4 but resolving localhost doesn't
+		// seem to return and IPv4 address, thus failing the connection.
+		conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))
```
haircommander, line 920: @danwinship @dcbw @squeed would this fail on ipv6 only nodes?

internal/oci/runtime_oci.go
```diff
@@ -898,73 +899,89 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
 
-	args := []string{"-t", fmt.Sprintf("%d", containerPid), "-n", socatPath, "-", fmt.Sprintf("TCP4:localhost:%d", port)}
+		errCh := make(chan error, 2)
 
-	nsenterPath, lookupErr := exec.LookPath("nsenter")
-	if lookupErr != nil {
-		return fmt.Errorf("unable to do port forwarding: nsenter not found")
-	}
+		debug := func(format string, args ...interface{}) {
+			log.Warnf(ctx, fmt.Sprintf(
```
haircommander, line 929: it strikes me a bit odd the function is called debug but it uses warn level

internal/oci/runtime_oci.go
aojea: I didn't dig more into this, it fails if localhost resolves to IPv6, at least in my tests with KIND, but we only have containerd in KIND. It works on IPv6 and IPv4 clusters this way.
I think that it depends on how the container is exposing the port inside the pod. 
It's common to still find some code in k/k hardcoded to 172.0.0.1 to avoid the `localhost` resolution ... :shrug:

server/container_portforward.go
```diff
@@ -56,5 +59,6 @@ func (s StreamService) PortForward(podSandboxID string, port int32, stream io.Re
 	}
 
 	emptyStreamOnError = false
-	return s.runtimeServer.Runtime().PortForwardContainer(c, port, stream)
+
+	return s.runtimeServer.Runtime().PortForwardContainer(ctx, c, sb.NetNsPath(), port, stream)
```
haircommander, line 63: NetNsPath() has the ability to return "", which I would deem an error here

internal/oci/runtime_oci.go
saschagrunert: Ah this should not happen, reverting it.

server/container_portforward.go
saschagrunert: The question is where we catch that. Is the `sb.NetNsPath() == ""` if we're running on host network? If yes, should we do the same like there: https://github.com/containerd/cri/blob/64aa9da76fc0ab333119f455f3b292244c1fae8c/pkg/server/sandbox_portforward_unix.go#L58-L60 ?

internal/oci/runtime_oci.go
danwinship: cri-o knows the pod's `eth0` IP doesn't it? Can't you just connect to that rather than to localhost?

aojea: that won't work for hostNetwork true, right?
I think that communication has to be done always through localhost, this runs inside the same namespace,  however, I couldn't understand how the port is published inside the namespace and why it is always bound to 127.0.0.1 (at least in containerd)

danwinship: You don't need port-forwarding in the hostNetwork case though...



> why it is always bound to 127.0.0.1 (at least in containerd)

Virtually all servers running in containers just bind to `0.0.0.0`, so it doesn't matter if you redirect to `127.0.0.1` or their primary IP. Golang-based servers will automatically bind to both `0.0.0.0` and `::` if you try to bind to either, but software written in other languages may not do that so, if we always send to `127.0.0.1` then it will fail for _some_ single-stack IPv6 containers.

server/container_portforward.go
haircommander: not quite. 
NetNsPath() returns "" iff we're `manage_ns_lifecycle` = false and the infra container PID is not initialized. Otherwise, we send `/proc/$pid/ns/net`, or `/var/run/netns/$id`
thus, empty NetNsPath is an error here (we somehow started port forwarding in a container without a pod infra container initialized)

saschagrunert: Thanks for the clarification, I added an error check

internal/oci/runtime_oci.go
saschagrunert: I changed the implementation to choose the first sandbox IP.

internal/oci/runtime_oci.go
```diff
@@ -898,73 +899,85 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
+	log.Infof(ctx,
+		"Starting port forward for %s in network namespace %s", c.ID(), netNsPath,
+	)
+
+	// Adapted reference implementation:
+	// https://github.com/containerd/cri/blob/8c366d/pkg/server/sandbox_portforward_unix.go#L65-L120
+	if err := ns.WithNetNSPath(netNsPath, func(_ ns.NetNS) error {
+		defer stream.Close()
+
+		conn, err := net.Dial("tcp4", fmt.Sprintf("%s:%d", ip, port))
```
aojea, line 916: for being IPv6 and IPv4 compatible you hast to change this to:

`net.Dial("tcp", net.JoinHostPort(ip, strconv.Itoa(port)))`

saschagrunert: Yes I changed it as suggested.

internal/lib/sandbox/sandbox.go
```diff
@@ -140,6 +140,18 @@ func (s *Sandbox) IPs() []string {
...
+// FirstIP returns either the first result of `IPs` or the host IP if the
+// sandbox is running on host network
+func (s *Sandbox) FirstIP() string {
+	// Use the first IP of the sandbox as target if non host-network
+	if len(s.IPs()) > 0 {
+		return s.IPs()[0]
+	}
+
+	// For host networks
+	return "localhost"
```
aojea, line 152: I suggest to not return localhost to avoid the resolution problem . it may resolve to ::1 or to 127.0.0.1 , but the application may not be bound to the same address
xref. https://gist.github.com/aojea/94f6f483173641647c731f582e52f0b0

@danwinship what do you think? is forcing to `127.0.0.1` safe adding a comment explaining or a TODO?

saschagrunert: Thanks, changed it to `127.0.0.1` for now.

aojea: rereading my comment it may suggest something different, I meant to keep the FirstIP code but instead of `return localhost` as the default just `return 127.0.0.1`

saschagrunert:
> er, wait, you dropped the `FirstIP` code and went back to hardcoding 127.0.0.1 all the time?

Yeah If I re-add it then the tests will fail in the CI: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3749/test_pull_request_crio_e2e_fedora/13065

Let me double check

danwinship:
> Yeah If I re-add it then the tests will fail in the CI: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3749/test_pull_request_crio_e2e_fedora/13065

`[sig-cli] Kubectl Port forwarding With a server listening on localhost should support forwarding over websockets`

Huh, it's explicitly checking that port-forwarding works to a server that is only listening on localhost. I wouldn't have expected that to be considered a real use case.

So in that case, the only way to pass that test, currently, is to be forwarding to `"localhost"`. Not `127.0.0.1`, because you have to be forwarding to whichever single IP the server is listening on, which is whichever one `"localhost"` resolves to.

I _think_ that test is broken and wrong, and it's certainly not "dual-stack-ready". But that's an issue for upstream. For now, for cri-o, I think you have to just go with `net.Dial("tcp", "localhost:$PORT")`. (`"tcp"` not `"tcp4"` like in the original patch)

saschagrunert:
Okay good thank you for the input @danwinship, I changed the PR as suggested and dropped my previous commit.

saschagrunert:
It might be possible that we have to adapt critest as well:
```
[k8s.io] Streaming runtime should support streaming interfaces runtime should support portforward [Conformance] 
/go/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/streaming.go:124
Timed out after 65.047s.
Expected
    <*url.Error | 0xc0001a12c0>: {
        Op: "Get",
        URL: "http://127.0.0.1:12001",
        Err: {
            Op: "read",
            Net: "tcp",
            Source: {IP: [127, 0, 0, 1], Port: 49574, Zone: ""},
            Addr: {IP: [127, 0, 0, 1], Port: 12001, Zone: ""},
            Err: {Syscall: "read", Err: 0x68},
        },
    }
to be nil
/go/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/networking.go:253
```

internal/oci/runtime_oci.go
```diff
@@ -898,73 +899,82 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
+	log.Infof(ctx,
+		"Starting port forward for %s in network namespace %s", c.ID(), netNsPath,
+	)
+
+	// Adapted reference implementation:
+	// https://github.com/containerd/cri/blob/8c366d/pkg/server/sandbox_portforward_unix.go#L65-L120
+	if err := ns.WithNetNSPath(netNsPath, func(_ ns.NetNS) error {
+		defer stream.Close()
+
+		conn, err := net.Dial("tcp", net.JoinHostPort("localhost", strconv.Itoa(int(port))))
```
danwinship, line 913: Oh, I think I see: I said you should do `"localhost"` to guarantee you get the same IP as the pod does, but this code isn't running fully inside the pod; it's in the pod's network namespace, but the _node_'s mount namespace, so it sees the node's `/etc/hosts`, `/etc/resolv.conf`, `/etc/nsswitch.conf`, so there's no reason to think that it's going to resolve "localhost" the same way as the pod would. Blah.

So ok, basically, trying to be dual-stack-ready is just a lose at this point. Hardcode to 127.0.0.1, and be ready to rewrite in some way when new upstream tests are added...

aojea: yeah, that's why  `conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))` previously
https://github.com/cri-o/cri-o/pull/3749#discussion_r423737909

saschagrunert: Changed the implementation to match the latest comments. Big thanks again! :pray:

saschagrunert:
Testing seems not green. PTAL @mrunalp @haircommander

internal/oci/runtime_oci.go
```diff
@@ -898,73 +899,82 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
+	log.Infof(ctx,
+		"Starting port forward for %s in network namespace %s", c.ID(), netNsPath,
+	)
+
+	// Adapted reference implementation:
+	// https://github.com/containerd/cri/blob/8c366d/pkg/server/sandbox_portforward_unix.go#L65-L120
+	if err := ns.WithNetNSPath(netNsPath, func(_ ns.NetNS) error {
+		defer stream.Close()
+
+		conn, err := net.Dial("tcp", net.JoinHostPort("127.0.0.1", strconv.Itoa(int(port))))
```
aojea, line 913: sorry for nitpicking, but if you have the chance of changing the PR again I prefer to carry on with the comment and implementation I used in containerd , if you don't have the whole context of the discussion this is easy to be changed by mistake
```suggestion
		// TODO: hardcoded to tcp4 because localhost resolves to ::1 by default if the system has IPv6 enabled.
		// Theoretically happy eyeballs will try IPv6 first and fallback to IPv4
		// but resolving localhost doesn't seem to return and IPv4 address always, thus failing the connection.
		conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))```

https://github.com/containerd/cri/pull/1470/files#diff-a60e53338184ff4aee2ff77caed7cabaR67-R70

Comment (saschagrunert) [reply internal/oci/runtime_oci.go:919]: Sure, added the comment as suggested.

Comment (aojea) [reply internal/oci/runtime_oci.go:919]: and the code	`conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))` 
:upside_down_face:

Comment (saschagrunert) [reply internal/oci/runtime_oci.go]: I was confused, changed that as well.

**Comment by haircommander** [comment]
LGTM, waiting on :+1: that this won't break ipv6 (or we're willing to break ipv6 for now) from @danwinship as I seem to have lost track of that conversation :laughing:

**Comment by aojea** [comment]
> LGTM, waiting on 👍 that this won't break ipv6 (or we're willing to break ipv6 for now) from @danwinship as I seem to have lost track of that conversation 😆

it won't break it because it was already using IPv4 only inside the namespace
> 	args := []string{"-t", fmt.Sprintf("%d", containerPid), "-n", socatPath, "-", fmt.Sprintf("TCP4:localhost:%d", port)}

it has to be revisited in the future for sure, but who knows when ;)

File: internal/oci/runtime_oci.go (hunk 4 above)
Comment (aojea) [reply internal/oci/runtime_oci.go:919]: for completenes, it seems that was hardcoded for security reasons in the past
https://github.com/containerd/cri/pull/1470/files#r424526233

**Comment by danwinship** [comment]
And we can't get ahead of the curve and deal with dual-stack while we're here because it's not obvious how to both correctly implement dual-stack _and_ pass the current blatantly-non-dual-stack kube e2e tests.

**Comment by haircommander** [comment]
in that case

/lgtm

**Comment by saschagrunert** [comment]
And in that case:
/retest

File: internal/oci/runtime_oci.go [hunk 11]
```diff
@@ -255,9 +256,9 @@ func prepareExec() (pidFile, parentPipe, childPipe *os.File, err error) {
 	return
 }
 
-func parseLog(log []byte) (stdout, stderr []byte) {
+func parseLog(l []byte) (stdout, stderr []byte) {
```
TomSweeneyRedHat, line 259: not a fan of this change.  "log" is so much easier to search for than 'l'.

saschagrunert: I can follow-up on that. `log` shadows the package import so we may rename it to `logBytes` or something else.

smarterclayton:
> Huh, it's explicitly checking that port-forwarding works to a server that is only listening on localhost. I wouldn't have expected that to be considered a real use case.

This is a fundamental Kube use case.  This is literally why portforward works the way it does :)

haircommander:
/hold 
those errors look legit
`[sig-network] Services should be able to preserve UDP traffic when server pod cycles for a NodePort service`

haircommander:
nevermind, I have seen that test fail [here](https://github.com/cri-o/cri-o/pull/3565) too, so it's not the fault of this PR

/hold cancel

saschagrunert:
> nevermind, I have seen that test fail [here](https://github.com/cri-o/cri-o/pull/3565) too, so it's not the fault of this PR
> 
> /hold cancel

The test is still failing everywhere but I was not able to reproduce it locally for now.

haircommander:
/test e2e

another day, another debugging session :smiley:

saschagrunert:
Is quay up again? If not we might give ourself a coffee ☕️ break.

haircommander:
didn't even make it that far
```

TASK [build runc] **************************************************************
task path: /go/src/github.com/cri-o/cri-o/contrib/test/integration/build/runc.yml:9
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/gmake 'BUILDTAGS=seccomp selinux'", "msg": "# github.com/opencontainers/runc/libcontainer/cgroups/fs\nlibcontainer/cgroups/fs/cpuset.go:63:14: undefined: fmt\ngmake: *** [Makefile:29: runc] Error 2", "rc": 2, "stderr": "# github.com/opencontainers/runc/libcontainer/cgroups/fs\nlibcontainer/cgroups/fs/cpuset.go:63:14: undefined: fmt\ngmake: *** [Makefile:29: runc] Error 2\n", "stderr_lines": ["# github.com/opencontainers/runc/libcontainer/cgroups/fs", "libcontainer/cgroups/fs/cpuset.go:63:14: undefined: fmt", "gmake: *** [Makefile:29: runc] Error 2"], "stdout": "go build \"-mod=vendor\" -buildmode=pie  -tags \"seccomp selinux\" -ldflags \"-X main.gitCommit=\"f369199ff61bafdb91b5308d8b30cb4c18270bd9\" -X main.version=1.0.0-rc10+dev \" -o runc .\n", "stdout_lines": ["go build \"-mod=vendor\" -buildmode=pie  -tags \"seccomp selinux\" -ldflags \"-X main.gitCommit=\"f369199ff61bafdb91b5308d8b30cb4c18270bd9\" -X main.version=1.0.0-rc10+dev \" -o runc ."]}

```

saschagrunert:
> didn't even make it that far
> 
> ```
> 
> TASK [build runc] **************************************************************
> task path: /go/src/github.com/cri-o/cri-o/contrib/test/integration/build/runc.yml:9
> fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/gmake 'BUILDTAGS=seccomp selinux'", "msg": "# github.com/opencontainers/runc/libcontainer/cgroups/fs\nlibcontainer/cgroups/fs/cpuset.go:63:14: undefined: fmt\ngmake: *** [Makefile:29: runc] Error 2", "rc": 2, "stderr": "# github.com/opencontainers/runc/libcontainer/cgroups/fs\nlibcontainer/cgroups/fs/cpuset.go:63:14: undefined: fmt\ngmake: *** [Makefile:29: runc] Error 2\n", "stderr_lines": ["# github.com/opencontainers/runc/libcontainer/cgroups/fs", "libcontainer/cgroups/fs/cpuset.go:63:14: undefined: fmt", "gmake: *** [Makefile:29: runc] Error 2"], "stdout": "go build \"-mod=vendor\" -buildmode=pie  -tags \"seccomp selinux\" -ldflags \"-X main.gitCommit=\"f369199ff61bafdb91b5308d8b30cb4c18270bd9\" -X main.version=1.0.0-rc10+dev \" -o runc .\n", "stdout_lines": ["go build \"-mod=vendor\" -buildmode=pie  -tags \"seccomp selinux\" -ldflags \"-X main.gitCommit=\"f369199ff61bafdb91b5308d8b30cb4c18270bd9\" -X main.version=1.0.0-rc10+dev \" -o runc ."]}
> ```

Okay, would it be okay for us to switch back to the static binary?

aojea:
but ... I don't see the relation between this PR and
```
[sig-network] Services should be able to preserve UDP traffic when server pod cycles for a NodePort service
```
is it only failing with this PR? 
someone opened a new issue in k/k about this test failing in openshift CI

haircommander:
> but ... I don't see the relation between this PR and
> 
> ```
> [sig-network] Services should be able to preserve UDP traffic when server pod cycles for a NodePort service
> ```
> 
> is it only failing with this PR?
> someone opened a new issue in k/k about this test failing in openshift CI

it's failing in cri-o everywhere

haircommander:
(note, we've opted to skip the test to unblock master)

haircommander:
https://github.com/opencontainers/runc/pull/2418 will fix runc compile errors

PR #3754, opened 2020-05-12
Automatically label containers running systemd with the correct label

haircommander:
/kind Feature

Newer versions of container-selinux, container-selinux-2.132.0 or newer,
supply a `container_init_t` label.  If CRI-O is running systemd or init inside
of the container, then the container will require different SELinux privs
to run the container.

By using the new SELinux label, we can run ordinary containers with a tighter
selinux policy then those running the init system, makeing the overlay system
more secure.

The eliminates the need to turn on the container_manage_cgroup SELinux boolean.

Ie. no need to execute

```
setsebool -P container_manage_cgroup 1
```

Any longer. On systems without the updated container-selnux package, the containers will still
attempt to run the standard container type `container_t`, and still require the boolean.

#### What type of PR is this?
/kind feature

```release-note
containers running `init` or `systemd` are now given a new selinux label `container_init_t`, giving it selinux privileges more appropriate for the workload
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -17,7 +17,7 @@ import (
 	"github.com/containers/buildah/pkg/secrets"
 	"github.com/containers/libpod/pkg/annotations"
 	"github.com/containers/libpod/pkg/rootless"
-	"github.com/containers/libpod/pkg/selinux"
+	selinux "github.com/containers/libpod/pkg/selinux"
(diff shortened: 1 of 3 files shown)


haircommander:
taking over https://github.com/cri-o/cri-o/pull/3707

umohnani8:
LGTM

Issue #3618, opened 2020-04-20
[1.18] unable to run pod - "cannot generate pod name without uid in metadata"


miabbott:
The RHEL CoreOS team has sanity tests for `cri-o` that started failing when we moved to `cri-o` 1.18

Our previous configuration (pod template/container template) no longer works with 1.18.  We are getting failures like:

```
            cluster.go:141: time="2020-04-20T18:01:00Z" level=fatal msg="run pod sandbox failed: rpc error: code = Unknown desc = setting pod sandbox name and id: cannot generate pod name without uid in metadata"
```

The error seems to match 2d5cedabe81ee94f4eef4c1656177faa542f9d1b

I suspect we'll be told to update our templates, but filing an issue to track the issue going from 1.17 to 1.18 for us.

haircommander:
@miabbott did you hit this on an upgrade? what's the template?

hswong3i:
I had upgrade my Ubuntu 18.04 + Kubernetes 1.17.5 + CRI-O 1.17.3, into Kubernetes 1.18.2 + CRI-O 1.18.0-RC1 successfully.

P.S. once upgraded all pod failed with “some” reason that I didn’t dig into it; as usual I reboot all nodes one by one, then kill all running pods once; after that everything resume as normal, even follow with another few rounds of reboot or pod restart.

miabbott:
@haircommander Yeah, hit this when RHCOS went from `cri-o` 1.17 to 1.18

Template:

```json
{
	"metadata": {
		"name": "rhcos-crio-pod-%s",
		"namespace": "redhat.test.crio"
	},
	"image": {
			"image": "localhost/%s:latest"
	},
	"args": [],
	"readonly_rootfs": false,
	"log_path": "",
	"stdin": false,
	"stdin_once": false,
	"tty": false,
	"linux": {
			"resources": {
					"memory_limit_in_bytes": 209715200,
					"cpu_period": 10000,
					"cpu_quota": 20000,
					"cpu_shares": 512,
					"oom_score_adj": 30,
					"cpuset_cpus": "0",
					"cpuset_mems": "0"
			},
			"cgroup_parent": "Burstable-pod-123.slice",
			"security_context": {
					"namespace_options": {
							"pid": 1
					},
					"capabilities": {
							"add_capabilities": [
								"sys_admin"
							]
					}
			}
	}
}
```

haircommander:
[...]

PR #3774, opened 2020-05-15
sandbox iface: don't fail if uid is not specified

haircommander:
#### What type of PR is this?

 /kind bug

before we started switching to have sandbox be in an interface, the only reason we fatally failed on run pod sandbox was if there was no Namespace. Uid did not need to be defined. I see no harm in continuing to do this.

fixes https://github.com/cri-o/cri-o/issues/3618

```release-note
Fixed bug where Pod creation would fail if Uid was not specified in Metadata of sandbox config passed in a run pod sandbox request
```

### DIFF
--- a/pkg/sandbox/sandbox.go
+++ b/pkg/sandbox/sandbox.go
@@ -92,10 +92,6 @@ func (s *sandbox) SetNameAndID() error {
 		return errors.New("cannot generate pod name without name in metadata")
 	}
 
-	if s.config.GetMetadata().GetUid() == "" {
-		return errors.New("cannot generate pod name without uid in metadata")
-	}
-
 	s.id = stringid.GenerateNonCryptoID()
 	s.name = strings.Join([]string{
 		"k8s",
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
Maybe it's just me with a light load of tea, but your PR description doesn't match the code change...

haircommander:
> Maybe it's just me with a light load of tea, but your PR description doesn't match the code change...

is that better?

haircommander:
PTAL @saschagrunert @umohnani8 @mrunalp @giuseppe @rhafer

umohnani8:
LGTM

TomSweeneyRedHat:
Title make more sense to my noggin now, thx.  IDK so thought I'd ask and sorry I didn't ask before merge.  Any value to changing the error to a Warning or a Debug statement instead of dropping it completely?

haircommander:
/shrug

I don't see a ton of value

PR #3784, opened 2020-05-19
[1.11] conmon: make un-OOM-killable

haircommander:
There are no situations that conmon should be OOM killed. In all situations conmon would be OOM killed, it is from a poorly acting container in conmon's cgroup. Even with -999, if conmon allocates memory at the wrong time, it will be OOM killed. If conmon is made un-OOM-killable, the container will be the one to be killed, as is correct.

/kind bug


TomSweeneyRedHat:
LGTM assuming happy tests

Issue #3792, opened 2020-05-21
CPU quota and period needed in runc master


haircommander:
tests are failing that run runc master because of https://github.com/opencontainers/runc/commit/06d7c1d2616052974ebb39a403a31e4ff95b9a20

```May 21 14:34:37 ip-172-18-15-199.ec2.internal crio[15379]: time="2020-05-21 14:34:37.102375216Z" level=error msg="Container creation error: time=\"2020-05-21T14:34:37Z\" level=warning msg=\"signal: killed\"\ntime=\"2020-05-21T14:34:37Z\" level=error msg=\"container_linux.go:353: starting container process caused: process_linux.go:304: applying cgroup configuration for process caused: CPU quota and period should both be set\"\n"```

it seems by the code we set them together from the resources
```
specgen.SetLinuxResourcesCPUPeriod(uint64(resources.GetCpuPeriod()))
specgen.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
```
so I am guessing one is not specified. we should probably figure out a default for each in the case one is specified, or not set either if one is not specified

haircommander:
cc @kolyshkin

haircommander:
this was fixed in runc

PR #3793, opened 2020-05-21
only set cpu period and quota when both are defined

haircommander:
#### What type of PR is this?

/kind bug

runc was previously doing this for us, but has since changed behavior. Now, we're responsible for not setting quota or period when either are undefined

fixes https://github.com/cri-o/cri-o/issues/3792

```release-note

```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -517,8 +517,11 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 	if linux != nil {
 		resources := linux.GetResources()
 		if resources != nil {
-			specgen.SetLinuxResourcesCPUPeriod(uint64(resources.GetCpuPeriod()))
-			specgen.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
+			// only set period or quota if both are configured, or else the runtime will fail
+			if resources.GetCpuPeriod() != 0 && resources.GetCpuQuota() != 0 {
+				specgen.SetLinuxResourcesCPUPeriod(uint64(resources.GetCpuPeriod()))
+				specgen.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
+			}
 			specgen.SetLinuxResourcesCPUShares(uint64(resources.GetCpuShares()))
 
 			memoryLimit := resources.GetMemoryLimitInBytes()
--- a/server/container_update_resources.go
+++ b/server/container_update_resources.go
@@ -30,15 +30,24 @@ func (s *Server) UpdateContainerResources(ctx context.Context, req *pb.UpdateCon
 
 // toOCIResources converts CRI resource constraints to OCI.
 func toOCIResources(r *pb.LinuxContainerResources) *rspec.LinuxResources {
-	var swap int64
+	var (
+		shares int64
+		swap   int64
+		quota  int64
+	)
 	memory := r.GetMemoryLimitInBytes()
 	if cgroupHasMemorySwap() {
 		swap = memory
 	}
+	// only set period or quota if both are configured, or else the runtime will fail
+	if r.GetCpuShares() != 0 && r.GetCpuQuota() != 0 {
+		shares = r.GetCpuShares()
+		quota = r.GetCpuQuota()
+	}
 	return &rspec.LinuxResources{
 		CPU: &rspec.LinuxCPU{
-			Shares: proto.Uint64(uint64(r.GetCpuShares())),
-			Quota:  proto.Int64(r.GetCpuQuota()),
+			Shares: proto.Uint64(uint64(shares)),
+			Quota:  proto.Int64(quota),
 			Period: proto.Uint64(uint64(r.GetCpuPeriod())),
 			Cpus:   r.GetCpusetCpus(),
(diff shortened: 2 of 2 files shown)


haircommander:
ah dang there's also the UpdateResources case...

mrunalp:
@kolyshkin ptal

server/container_create_linux.go
```diff
@@ -517,8 +517,11 @@ func (s *Server) createSandboxContainer(ctx context.Context, containerID, contai
 	if linux != nil {
 		resources := linux.GetResources()
 		if resources != nil {
-			specgen.SetLinuxResourcesCPUPeriod(uint64(resources.GetCpuPeriod()))
-			specgen.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
+			// only set period or quota if both are configured, or else the runtime will fail
+			if resources.GetCpuPeriod() != 0 && resources.GetCpuQuota() != 0 {
+				specgen.SetLinuxResourcesCPUPeriod(uint64(resources.GetCpuPeriod()))
+				specgen.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
+			}
```
kolyshkin, line 524: I believe instead of silently ignoring one setting if the other is not set, we should actually return an error here, as it's an invalid configuration (the two parameters depend on each other and it does not make sense to set one but not the other).

server/container_update_resources.go
```diff
@@ -30,15 +30,24 @@ func (s *Server) UpdateContainerResources(ctx context.Context, req *pb.UpdateCon
 
 // toOCIResources converts CRI resource constraints to OCI.
 func toOCIResources(r *pb.LinuxContainerResources) *rspec.LinuxResources {
-	var swap int64
+	var (
+		shares int64
+		swap   int64
+		quota  int64
+	)
```
kolyshkin, line 37: nit:
```go
var swap, shares, quota int64
```
although the way it is is also OK for me

server/container_update_resources.go
```diff
@@ -30,15 +30,24 @@ func (s *Server) UpdateContainerResources(ctx context.Context, req *pb.UpdateCon
...
 	memory := r.GetMemoryLimitInBytes()
 	if cgroupHasMemorySwap() {
 		swap = memory
 	}
+	// only set period or quota if both are configured, or else the runtime will fail
+	if r.GetCpuShares() != 0 && r.GetCpuQuota() != 0 {
+		shares = r.GetCpuShares()
+		quota = r.GetCpuQuota()
+	}
```
kolyshkin, line 46: same as above -- I think it makes more sense to return an error

kolyshkin:
@haircommander can you please show the actual failures? If those are coming from the tests, I think those tests need to be fixed. The quota and the period should always be used together, since they don't make sense otherwise. It's the same as giving a fraction number for which you only specify either the numerator or the denominator and omit the other.

haircommander:
example failed run: 
https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3305/test_pull_request_crio_e2e_rhel/13303/
example failed logs:
```
May 21 14:37:51 ip-172-18-4-68.ec2.internal crio[21233]: time="2020-05-21 14:37:51.497521880Z" level=info msg="Attempting to create container: kube-system/kube-dns-c74654849-wnrnj/kubedns" id=0cddd24d-28b0-4e78-9871-966a7fe560ee name=/runtime.v1alpha2.RuntimeService/CreateContainer
May 21 14:37:51 ip-172-18-4-68.ec2.internal crio[21233]: time="2020-05-21 14:37:51.584887009Z" level=error msg="Container creation error: time=\"2020-05-21T14:37:51Z\" level=warning msg=\"signal: killed\"\ntime=\"2020-05-21T14:37:51Z\" level=error msg=\"container_linux.go:353: starting container process caused: process_linux.go:304: applying cgroup configuration for process caused: CPU quota and period should both be set\"\n"
```

This error is coming from the runtime. It seems that kube is not required to send both at once. If we error out here, the same result will happen (we won't pass e2e tests because both resources are not being specified)

haircommander:
```
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/kubectl/kubectl.go:549 May 21 16:39:45.130: Expected     <string>:  to contain substring     <string>: stdin closed /go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/kubectl/kubectl.go:573
```
[ref](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3793/test_pull_request_crio_e2e_rhel/13311/)
uncool

/retest

haircommander:
/hold

a runc pr will replace this, but in the meantime I'm using it as a way to debug the fedora failures

kolyshkin:
Please hold on, I am going to fix this in runc instead.

tedyu:
@kolyshkin 
Can you link the runc fix please ?

haircommander:
fixed in https://github.com/opencontainers/runc/pull/2434

PR #3799, opened 2020-05-23
cgroup features detection speedup and simplification

kolyshkin:
One speedup, one fix, and a few simplifications to the code.

The speedup is not calling `cgroups.FindCgroupMountpoint` to test whether the pids controller is available. Believe me, this is huge.

The fix is not calling `cgroups.FindCgroupMountpoint` o test whether the pids controller is available (it will always say "yes, sir").

Simplification number one is reusing the same code to check for hugetlb controller.

Simplification number two is dropping sync.Once use to avoid an os.Stat() call.

Overall result: faster, simpler code.

Side note: this also prepares the codebase for upcoming changes in runc (https://github.com/opencontainers/runc/pull/2411)

---

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -5,7 +5,6 @@ package server
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


haircommander:
this feels like what https://github.com/cri-o/cri-o/pull/3565 is trying to do. I'm not a fan of dropping sync.Once, because it's needless re-calling of functions that will never change after cri-o starts up, which I don't believe will result in faster code execution (I do agree it complicates the code, but having the separate package handle that makes it clearer IMO). If you want, I can fold the changes to the runc functions into the linked PR

kolyshkin:
> this feels like what #3565 is trying to do

Yes, this is kind of a simpler alternative to it. What I like about this one is
- it's mostly removing (rather than adding) code, and the less code the better
- it actually improves performance (by not calling the dreadful `FindCgroupMountpoint`)

>  I'm not a fan of dropping sync.Once, because it's needless re-calling of functions that will never change after cri-o starts up

This, of course, depends on how many times we call the function in question, and how heavy (in terms of performance) is the function we call. I have carefully considered both metrics before dropping sync.Once.

For example, the second commit does stat(2) syscall. I do not think that sparing one stat(2) on a container start will affect the performance.

The first commit, well, it's a little more going on (reading/parsing a small proc file from the kernel) and I agree it could use a sync.Once.

Please see updated.

haircommander:
I have pulled some of these changes into a new PR here: https://github.com/cri-o/cri-o/pull/3810

kolyshkin:
Obsoleted by https://github.com/cri-o/cri-o/pull/3810 (which incorporates some of the ideas from here)

PR #3804, opened 2020-05-28
Cleanup and update nix derivation for static builds

hswong3i:
#### What type of PR is this?

> /kind cleanup
> /kind dependency-change

#### What this PR does / why we need it:

Similar PR goes for crun/conmon/libpod/cri-o/etc, too.

Also see:
  - ~~https://github.com/containers/crun/pull/372~~
  - ~~https://github.com/containers/conmon/pull/161~~
  - ~~https://github.com/containers/skopeo/pull/932~~
  - ~~https://github.com/containers/buildah/pull/2380~~
  - ~~https://github.com/containers/libpod/pull/6402~~
  - https://github.com/cri-o/cri-o/pull/3804

Static binaries:
  - [crun-0.13-linux-amd64](https://github.com/alvistack/crun/releases/download/0.13/crun-0.13-linux-amd64)
  - [conmon-v2.0.17-linux-amd64](https://github.com/alvistack/conmon/releases/download/v2.0.17/conmon-v2.0.17-linux-amd64)
  - [skopeo-v1.1.0-linux-amd64](https://github.com/alvistack/skopeo/releases/download/v1.1.0/skopeo-v1.1.0-linux-amd64)
  - [buildah-v1.15.0-linux-amd64](https://github.com/alvistack/buildah/releases/download/v1.15.0/buildah-v1.15.0-linux-amd64)
  - [podman-v1.9.3-linux-amd64](https://github.com/alvistack/libpod/releases/download/v1.9.3/podman-v1.9.3-linux-amd64)
  - [cri-o-v1.17.4-linux-amd64.tar.gz](https://github.com/alvistack/cri-o/releases/download/v1.17.4/cri-o-v1.17.4-linux-amd64.tar.gz)
  - [cri-o-v1.18.2-linux-amd64.tar.gz](https://github.com/alvistack/cri-o/releases/download/v1.18.2/cri-o-v1.18.2-linux-amd64.tar.gz)

Ansible Roles:
  - https://github.com/alvistack/ansible-role-crun
  - https://github.com/alvistack/ansible-role-conmon
  - https://github.com/alvistack/ansible-role-skopeo
  - https://github.com/alvistack/ansible-role-buildah 
  - https://github.com/alvistack/ansible-role-podman
  - https://github.com/alvistack/ansible-role-cri_o

#### Special notes for your reviewer:

Here I skip the btrfs and lvm2 support for static binary, because:
1. btrfs will not support in CentOS 8
[...]

PR #3805, opened 2020-05-28
[release-1.18] Cleanup nix derivation for static builds

hswong3i:
Also see #3804

nix-based static binary could be found from:
- https://github.com/alvistack/cri-o/releases/tag/v1.18.1
- https://github.com/alvistack/cri-o/releases/tag/v1.17.4


hswong3i:
Should be too difficult to keep CI update with such changes.

PR #3812, opened 2020-06-02
oci: Don't return early for non-tty attach if there is no stdin

mrunalp:
We want to return early from a non-tty attach only if stdin is
enabled otherwise we want to wait for the output as well.

#### What type of PR is this?

 /kind failing-test

This PR fixes a race exposed by recent changes in https://github.com/kubernetes/kubernetes/pull/91227

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -885,7 +885,12 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
 	case err := <-receiveStdout:
 		return err
 	case err := <-stdinDone:
-		if !c.StdinOnce() && !tty {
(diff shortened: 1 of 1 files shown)


mrunalp:
@haircommander @saschagrunert @giuseppe ptal. This fixes the tests.

giuseppe:
woho green tests!  I am fine to merge it but I still don't understand this requirement, why should we ignore stdout on `c.stdin && !c.StdinOnce() && !tty`, is this behaviour specified anywhere or is it some docker behavior?

haircommander:
it seems to be docker behavior. All of this is pretty poorly documented. Unfortunately, if a dockershim test passes and we fail, then we have to go to that :\

/lgtm

PR #3824, opened 2020-06-03
Use correct format for logPath removal log

tedyu:
#### What type of PR is this?

/kind bug

In ensureSaneLogPath(), if the call to os.RemoveAll fails, we log the error.
However, the format string is wrong. '%s' shouldn't be used with err.

This PR corrects the mistake.

```release-note
None
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -502,7 +502,7 @@ func ensureSaneLogPath(logPath string) error {
 	if os.IsNotExist(err) {
 		err = os.RemoveAll(logPath)
 		if err != nil {
-			return fmt.Errorf("ensureSaneLogPath remove bad logPath: %s", err)
+			return fmt.Errorf("failed to remove bad log path %s: %v", logPath, err)
 		}
 	}
 	return nil


server/container_create.go
```diff
@@ -502,7 +502,7 @@ func ensureSaneLogPath(logPath string) error {
 	if os.IsNotExist(err) {
 		err = os.RemoveAll(logPath)
 		if err != nil {
-			return fmt.Errorf("ensureSaneLogPath remove bad logPath: %s", err)
+			return fmt.Errorf("ensureSaneLogPath remove bad logPath %s : %v", logPath, err)
```
mrunalp, line 505: `failed to remove bad log path %s: %v", logPath, err`

tedyu: Your comment is addressed.

server/container_create.go
```diff
@@ -502,7 +502,7 @@ func ensureSaneLogPath(logPath string) error {
 	if os.IsNotExist(err) {
 		err = os.RemoveAll(logPath)
 		if err != nil {
-			return fmt.Errorf("ensureSaneLogPath remove bad logPath: %s", err)
+			return fmt.Errorf("failed to remove bad log path %s : %v", logPath, err)
```
mrunalp, line 505: nit: There is extra space before the colon which should be removed: `failed to remove bad log path %s: %v", logPath, err`

tedyu: Removed the space.

TomSweeneyRedHat:
LGTM

umohnani8:
LGTM

PR #3830, opened 2020-06-04
Close the done channel when there is watcher error

tedyu:
#### What type of PR is this?

/kind bug

When the following case is triggered, done channel is not closed.
```
case err := <-watcher.Errors:
   logrus.Debugf("watch error: %v", err)
   return
```
This PR closes the done channel before returning.

```release-note
None
```


haircommander:
/approve
LGTM

tedyu:
@saschagrunert 
Can you take a look please ?

PR #3836, opened 2020-06-04
Add sandbox IPs if there is no error in IP retrieval

tedyu:
#### What type of PR is this?

/kind bug

In Server#restore() 
```
ips, err := s.getSandboxIPs(sb)
if err != nil {
   logrus.Warnf("could not restore sandbox IP for %v: %v", sb.ID(), err)
}
sb.AddIPs(ips)
```
sb.AddIPs() should only be called when err return from getSandboxIPs() is nil.

```release-note
None
```


TomSweeneyRedHat:
LGTM, but it looks like the tests are kicking up their heels.

haircommander:
both intermittent failures, I blame the robots

/lgtm

PR #3839, opened 2020-06-05
managed_ns: deflake tests

haircommander:
#### What type of PR is this?

/kind flake

turns out, the PID 42 (as idiomatic a test variable it is) is not the best PID to use.
NamespacePaths() returns /proc entries if the PID is running, so sometimes there was a PID 42 on the system, and the tests returned valid paths, causing the test to fail

This PR updates the tests to use the values 0 (unset pid), 1 (running pid), and 4194305 (never running pid, as it's max_pid+1)

it also fixes a small inconsistency in reserving space for the slice of namespace paths

```release-note

```

### DIFF
--- a/internal/lib/sandbox/namespaces.go
+++ b/internal/lib/sandbox/namespaces.go
@@ -14,10 +14,11 @@ import (
 type NSType string
 
 const (
-	NETNS  NSType = "net"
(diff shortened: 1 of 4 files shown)


haircommander:
PTAL @umohnani8 @mrunalp

PR #3841, opened 2020-06-05
oci: check container status before stop atomically

haircommander:
#### What type of PR is this?

/kind flake

in stop, we need to check the pod's status before stopping it.
However, as it stands, we do it in two distinct steps: State() followed by ContainerStop()

There exists a race where:
the State() call happens
followed by an UpdateContainerStatus() (marking the container as stopped)
followed by the Container Stop()

we've hit this race quite a bit in circleCI running 'pod stop idempotent with ctrs already stopped'

fix this by moving the check branching on Status to after we take the opLock()

```release-note

```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/libpod/pkg/cgroups"
 	"github.com/containers/storage/pkg/idtools"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	"k8s.io/apimachinery/pkg/fields"
@@ -24,6 +25,7 @@ import (
 const defaultStopSignalInt = 15
 
 var defaultStopSignal = strconv.Itoa(defaultStopSignalInt)
+var ErrContainerStopped = errors.New("container is already stopped")
 
 // Container represents a runtime container.
 type Container struct {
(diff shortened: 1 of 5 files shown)


internal/oci/container.go
```diff
@@ -24,6 +25,7 @@ import (
 const defaultStopSignalInt = 15
 
 var defaultStopSignal = strconv.Itoa(defaultStopSignalInt)
+var ErrContainerStopped = errors.Errorf("container is already stopped")
```
mrunalp, line 28: errors.New

internal/oci/container.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/libpod/pkg/cgroups"
 	"github.com/containers/storage/pkg/idtools"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
```
mrunalp, line 17: We can use the golang `errors` package instead of using this.

mrunalp:
Couple of nits, otherwise looks fine.

mrunalp:
This should be backported to 1.18.

internal/oci/container.go
haircommander: I prefer pkg/errors, it's more comprehensive

internal/oci/container.go

internal/oci/container.go
mrunalp: I think we will have to check across the repo if we are using anything beyond https://blog.golang.org/go1.13-errors. But this is okay for now since we already are using it.

PR #3848, opened 2020-06-08
Container should only be added once after passing filter

tedyu:
#### What type of PR is this?

/kind bug

In ContainerServer#ListContainers, we should only include container once even if there are two filters which the container passes.

```release-note
None
```


nalind:
LGTM

haircommander:
/approve
LGTM

PR #3854, opened 2020-06-08
Send only single error to stdinDone

tedyu:
#### What type of PR is this?

/kind bug

In runtimeOCI#AttachContainer, if closeErr is not nil, there would be two writes into the stdinDone channel (and the closeErr may preempt utils.DetachError).

this PR makes sure that only one error is written to the channel.

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -875,14 +875,20 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
 
 	stdinDone := make(chan error)
 	go func() {
-		var err error
+		var err, closeErr error
 		if inputStream != nil {
 			_, err = utils.CopyDetachable(conn, inputStream, nil)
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_oci.go
```diff
@@ -875,14 +875,18 @@ func (r *runtimeOCI) AttachContainer(c *Container, inputStream io.Reader, output
...
-			if closeErr := conn.CloseWrite(); closeErr != nil {
-				stdinDone <- closeErr
-			}
+			closeErr = conn.CloseWrite()
+		}
+		if err != nil {
+			stdinDone <- err
+		} else if closeErr != nil {
+			stdinDone <- closeErr
+		} else {
```
haircommander, line 887: just thought of this: should probably have a comment here, or maybe instead of sending `stdinDone <- err` it would be more clear to have it be `stdinDone <- nil`

tedyu: @haircommander 
Please take another look.

PR #3857, opened 2020-06-09
Close the done channel in waitContainerStop

tedyu:
#### What type of PR is this?

/kind bug

In waitContainerStop, done channel should be closed by the goroutine.
```
case <-chControl:
   return
```
The close of channel was missing in the above case.

```release-note
None
```


tedyu:
@saschagrunert 
Can you take a look ?

PR #3859, opened 2020-06-09
use inactive-or-failed CollectMode if appropriate

haircommander:
#### What type of PR is this?

/kind cleanup

currently, systemd leaks some scopes if they're failed.
This can be changed with the CollectMode annotation (which runc will set).
However, only systemd version >= 236 has support for this, so we need a new entry in internal/config/node to check whether systemd is the correct version

```release-note

```

### DIFF
--- a/internal/config/node/node.go
+++ b/internal/config/node/node.go
@@ -41,6 +41,13 @@ func ValidateConfig() error {
 			activated: &cgroupHasMemorySwap,
 			fatal:     false,
 		},
+		{
+			name:      "systemd CollectMode",
+			init:      SystemdHasCollectMode,
+			err:       &systemdHasCollectModeErr,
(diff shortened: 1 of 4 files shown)


server/container_create_unsupported.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/oci"
 	"github.com/cri-o/cri-o/pkg/sandbox"
+	"github.com/docker/docker/container"
```
saschagrunert, line 11: Do we need this change?

haircommander: hm how'd that sneak in there

PR #3861, opened 2020-06-09
Use the container_kvm_t label when using kata as the runtime

umohnani8:

#### What type of PR is this?

/kind feature

Add a check to detect when cri-o is using kata as the runtime
and set the process label to container_kvm_t.

```release-note

```


umohnani8:
@rhatdan @mrunalp PTAL

server/container_create_linux.go
```diff
@@ -871,6 +871,19 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 		return nil, err
 	}
 
+	runtimeType, err := s.Runtime().ContainerRuntimeType(ociContainer)
+	if err != nil {
+		return nil, err
+	}
+	// If using kata runtime, the process label should be set to container_kvm_t
+	// Note: the requirement here is that the name used for the runtime class has "kata" in it
+	if runtimeType == libconfig.RuntimeTypeVM && strings.Contains(strings.ToLower(sb.RuntimeHandler()), "kata") {
```
haircommander, line 880: I know kata v1 is being deprecated, but right now, there is a RuntimeTypeOCI that is technically a kata type.

We could change this to an `||` instead of an `&&`, so we always set KVM label if we're runtime VM, and then use the kata keyword if we're OCI but using kata as an OCI type.

haircommander:
can you add a release note? I think the different label could be classified as user facing

PR #3864, opened 2020-06-10
make all operations read state atomically

haircommander:
#### What type of PR is this?

/kind bug

there are still two cases where the container state isn't checked atomically: container start and sandbox remove

fix these cases, so we don't risk updating state based on stale data

```release-note

```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -26,6 +26,7 @@ const defaultStopSignalInt = 15
 
 var defaultStopSignal = strconv.Itoa(defaultStopSignalInt)
 var ErrContainerStopped = errors.New("container is already stopped")
+var ErrContainerNotCreated = errors.New("container is not in created state")
 
 // Container represents a runtime container.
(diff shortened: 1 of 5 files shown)


haircommander:
I will get to this after https://github.com/cri-o/cri-o/pull/3868, I think we need an internal UpdateContainerStatus()

internal/oci/container.go
```diff
@@ -357,3 +358,10 @@ func (c *Container) ShouldBeStopped() error {
 	}
 	return nil
 }
+
+func (c *Container) ShouldBeStarted() error {
+	if c.state.Status == ContainerStateCreated {
```
umohnani8, line 363: Should we check if the container is in running state as well? If not, can we rename the function to `ShouldBeCreated`? `ShouldBeStarted` makes me think that running state is valid as well.

haircommander:
replacing this with https://github.com/cri-o/cri-o/pull/4406

PR #3866, opened 2020-06-10
[1.18] oci: repeat $runtime state calls if appropriate

haircommander:
#### What type of PR is this?

 /kind bug

we have seen cases where `$runtime state` calls fail supriously, but succeed later
this is not great, though, we shouldn't incorrectly label pods if this happens.

We now retry state calls up to three times if we determine the container is still running (by calling kill on its pid)

also carry https://github.com/cri-o/cri-o/pull/3853

```release-note

```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -344,6 +344,24 @@ func (c *Container) exitFilePath() string {
 	return filepath.Join(c.dir, "exit")
 }
 
+// IsRunning is a function that checks if a specific pid is running
+// it is used to check a container state when we don't want (or don't trust) a `$runtime state` call
+func (c *Container) IsRunning() bool {
+	pid := c.state.Pid
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		logrus.Infof("container %s not found: %v", c.id, err)
+		return false
+	}
+
+	if err := process.Signal(syscall.Signal(0)); err != nil {
+		logrus.Errorf("container %s not running %v", c.id, err)
+		return false
+	}
+
+	return true
+}
+
 // ShouldBeStopped checks whether the container state is in a place
 // where attempting to stop it makes sense
 // a container is not stoppable if it's paused or stopped
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -699,19 +699,30 @@ func (r *runtimeOCI) UpdateContainerStatus(c *Container) error {
 		return nil
 	}
 
-	cmd := exec.Command(r.path, rootFlag, r.root, "state", c.id) // nolint: gosec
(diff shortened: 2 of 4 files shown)


internal/oci/container.go
```diff
@@ -344,6 +344,24 @@ func (c *Container) exitFilePath() string {
 	return filepath.Join(c.dir, "exit")
 }
 
+// IsRunning is a function that checks if a specific pid is running
+// it is used to check a container state when we don't want (or don't trust) a `$runtime state` call
+func (c *Container) IsRunning() bool {
+	pid := c.state.Pid
+	process, err := os.FindProcess(pid)
```
mrunalp, line 351: Is this necessary? We could directly make the syscall since we already know the pid.

haircommander: I suppose not, though I personally would rather use the standard lib instead of directly using syscall. I am worried about having to handle things that the lib takes care of (this isn't a great example, but for go 1.14, longer syscalls need eintr handling https://golang.org/doc/go1.14#runtime). My inclination is to leave as is

internal/oci/runtime_oci.go
```diff
@@ -699,18 +699,30 @@ func (r *runtimeOCI) UpdateContainerStatus(c *Container) error {
...
+	// We have seen cases where this state call fails
+	// with little output, but the container is running and
+	// later calls succeed.
+	// There is little harm in trying again
+	for i := 0; i < 3; i++ {
+		out, err = r.attemptUpdateContainerStatus(c)
+		if err == nil {
+			break
+		}
+		logrus.Errorf("%v: attempt %d", err, i)
```
giuseppe, line 713: could we add more context to this error?

internal/oci/container.go
giuseppe: `os.FindProcess(pid)` doesn't add any extra cost on unix as it just populates a struct with the specified PID.

internal/oci/runtime_oci.go
haircommander: the context comes from the called function:
`failed to update container state for %s: %v", c.id, err`
I couldn't think of anything else to add out here that wouldn't be redundant. Is there anything you'd like to see added?

internal/oci/container.go
mrunalp: https://github.com/golang/go/blob/cb11c981df7b4dc40550ab71cc097c25d24d7a71/src/os/exec_unix.go#L97 :+1:

haircommander:
/hold
entirely possible we can get away without this

haircommander:
we have identified the underlying issue in runc. it's best not to hide flakes like this pr does.

PR #3867, opened 2020-06-11
reduce flakiness of UpdateContainerStatus

haircommander:
#### What type of PR is this?

 /kind bug

this PR does two things:
- forward ports https://github.com/cri-o/cri-o/pull/3866, to make UpdateContainerStatus more reliable
- checks a container pid instead of calling UpdateContainerStatus for execsyncs (all probes) which should improve performance

```release-note

```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -344,6 +344,23 @@ func (c *Container) exitFilePath() string {
 	return filepath.Join(c.dir, "exit")
 }
 
+// IsRunning is a function that checks if a specific pid is running
+// it is used to check a container state when we don't want (or don't trust) a `$runtime state` call
+func (c *Container) IsRunning() bool {
+	process, err := os.FindProcess(c.state.Pid)
+	if err != nil {
+		logrus.Infof("container %s not found: %v", c.id, err)
+		return false
+	}
+
+	if err := process.Signal(syscall.Signal(0)); err != nil {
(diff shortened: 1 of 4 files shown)


internal/oci/container.go
```diff
@@ -344,6 +344,24 @@ func (c *Container) exitFilePath() string {
 	return filepath.Join(c.dir, "exit")
 }
 
+// IsRunning is a function that checks if a specific pid is running
+// it is used to check a container state when we don't want (or don't trust) a `$runtime state` call
+func (c *Container) IsRunning() bool {
+	pid := c.state.Pid
+	process, err := os.FindProcess(pid)
```
saschagrunert, line 351:
```suggestion
	process, err := os.FindProcess(c.state.Pid)
```

haircommander: because the function does not hold the lock, I am worried about the pid changing from underneath it. saving the value causes this function to run atomically

haircommander: though, I see now we don't even use this variable anywhere else. taking the change as suggested...

haircommander:
> We should also store the process start time in container state and use that for matching the process to rule
> out any pid reuse scenarios.

I'd like to do that as a follow up, to use the implementation for all accesses of c.state.Pid

haircommander:
/hold

the approach is changing

haircommander:
closing in favor of https://github.com/cri-o/cri-o/pull/3868

PR #3868, opened 2020-06-11
oci: check pid before using it

haircommander:
#### What type of PR is this?

 /kind bug

As we found in the conmonmon saga, directly accessing pids on the host is a bit dangerous. Primarily, we risk believing a pid is the container process, when it is really a new process, because of pid wrap. While this is really unlikely to happen if you set it to the kernel pid_max (4 billion?), it still can happen, and can cause security problems (especially when using it for namespaces, when not managing namespace lifecycle)

Now, we drop every direct c.state.Pid access in favor of a helper function c.Pid() (and children). This allows us to do the proper verification (checking stime is the same for the pid we're trying to use as the container's pid originally was)

This PR also refactors how we use FindProcess(), as there was a bit of duplication, and direct accesses to c.state.Pid were littered in code associated with FindProcess()

This also carries the exec sync patch checking c.IsRunning() instead of updating container state, as a bonus.

TODO:
- [x] test restore case (I need to find the best way to persist the pid to disk)
- [x] add more tests
~- [ ] maybe add this pid handling stuff in a new package for better unit testing~

But I'm throwing this up to see how CI likes it

```release-note

```


internal/oci/container.go
```diff
@@ -58,6 +60,10 @@ type Container struct {
 	stdin              bool
 	stdinOnce          bool
 	created            bool
+	// the unix start time of the process, as found by psgo
+	// this is used to track whether the pid we have stored
+	// is the same as the same pid number on the host
+	psStartTime        string
```
mrunalp, line 66: How about `pidStartTime` ?

haircommander: yeah that's better. I was at a loss for a variable name :)

mrunalp: `processStartTime` is probably better than that.

internal/oci/container.go
```diff
@@ -344,6 +350,104 @@ func (c *Container) exitFilePath() string {
...
+// pid is the internal, non-locking version of pid. It checks
+// that we have a Pid defined in the state, that pid can be found
+// and it is the same process that was originally started by the runtime
+func (c *Container) pid() (int, error) {
+	if c.state == nil {
+		return -1, errors.New("state not initialized")
+	}
+	pid := c.state.Pid
+	if pid <= 0 {
+		return -1, errors.New("pid not initialized")
```
mrunalp, line 382: We can introduce an error variable here like `ErrPidNotRunning`

internal/oci/container.go
mrunalp: This will have to be persisted in state.

mrunalp: I am thinking that we should probably separate the pid returned from runc state from the container start pid. runc state could change to zero once stopped. We can sepearately store the startTime, Pid as top level fields in crio state that never change once populated.

server/container_create_linux.go
```diff
@@ -570,7 +570,11 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 
 		// share Pod PID namespace
 		// SEE NOTE ABOVE
-		pidNsPath := fmt.Sprintf("/proc/%d/ns/pid", infra.State().Pid)
+		pid, err := infra.Pid()
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to find pid for container %s", infra.ID())
+		}
+		pidNsPath := fmt.Sprintf("/proc/%d/ns/pid", pid)
```
haircommander, line 577: TODO: this should really be handled in internal/sandbox/namespaces*

haircommander:
integration tests are a runc problem fixed in https://github.com/opencontainers/runc/pull/2479, e2e test failures are legit

haircommander:
PTAL @umohnani8 @mrunalp @saschagrunert @kolyshkin

internal/oci/container.go
```diff
@@ -344,6 +386,109 @@ func (c *Container) exitFilePath() string {
...
+	if err != nil {
+		return found, err
+	}
+
+	if err := c.verifyPid(pid); err != nil {
+		return found, err
+	}
+
+	found = true
+	if err := process.Release(); err != nil {
```
giuseppe, line 457: I'd still try to release it anyway.  If it is not a direct child an error will be raised, and if it is a direct child then we are sure to free the process in case the verification step fails

internal/oci/container.go
haircommander: I'm not sure we need that. I'd be fine doing it but there's no place we really want to check against this as of now

internal/oci/container.go
haircommander: good idea, fixed as suggested

haircommander:
```

[sig-storage] In-tree Volumes [Driver: aws] [Testpattern: Dynamic PV (block volmode)] provisioning should provision storage with mount options [Suite:openshift/conformance/parallel] [Suite:k8s]

```
we have trouble with this one in upstream e2e too
[ref](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/cri-o_cri-o/3868/pull-ci-cri-o-cri-o-master-e2e-aws/1282743517545762816)
/test e2e-aws

TomSweeneyRedHat:
LGTM
but it looks like a rebase is now needed, and it was just happy green test buttons....

haircommander:
aaaaaaaaaaaaand we're back :sunglasses:

haircommander:
happy tests PTAL @giuseppe @mrunalp @kolyshkin @umohnani8

internal/oci/container.go
```diff
@@ -344,6 +344,23 @@ func (c *Container) exitFilePath() string {
 	return filepath.Join(c.dir, "exit")
 }
 
+// IsRunning is a function that checks if a specific pid is running
+// it is used to check a container state when we don't want (or don't trust) a `$runtime state` call
+func (c *Container) IsRunning() bool {
+	process, err := os.FindProcess(c.state.Pid)
```
kolyshkin: On UNIX, os.FindProcess always succeeds (see its doc), so you safely simplify by doing
```go
  // On Unix systems, FindProcess always succeeds
	process, _ := os.FindProcess(c.state.Pid)
```
and dropping the error check below

kolyshkin: ...unless this code is platform-agnostic, of course

internal/oci/container.go
```diff
@@ -344,6 +344,23 @@ func (c *Container) exitFilePath() string {
...
+	if err != nil {
+		logrus.Infof("container %s not found: %v", c.id, err)
+		return false
+	}
+
+	if err := process.Signal(syscall.Signal(0)); err != nil {
+		logrus.Errorf("container %s not running %v", c.id, err)
```
kolyshkin: nit:
```suggestion
		logrus.Errorf("container %s not running: %v", c.id, err)
```

internal/oci/container_test.go
```diff
@@ -238,4 +238,38 @@ var _ = t.Describe("Container", func() {
...
+		// When
+		err := sut.IsRunning()
+
+		// Then
+		Expect(err).To(Equal(false))
+	})
+	It("should succeed if pid is running", func() {
+		// Given
+		state := &oci.ContainerState{}
+		state.Pid = 1
```
kolyshkin: If this mean to `kill(1, 0)` (ie check if pid 1 (init) is running), this will fail with EPERM if the caller is not root -- so the test case should require root.

internal/oci/container_test.go
```diff
@@ -238,4 +238,38 @@ var _ = t.Describe("Container", func() {
...
+		err := sut.IsRunning()
+
+		// Then
+		Expect(err).To(Equal(true))
+	})
+	It("should be false if pid is not running", func() {
+		// Given
+		state := &oci.ContainerState{}
+		// the highest allowed pid + 1
+		state.Pid = 4194305
```
kolyshkin: Ideally, this should be the value from `/proc/sys/kernel/pid_max` + 1 as this is configurable.

From the other side, it seems that `4194304` (`0x400000`) is indeed the maximum value possible, so might be OK to leave as is.

```console
# echo 10000 > /proc/sys/kernel/pid_max
# cat /proc/sys/kernel/pid_max
10000
# echo 0x400000 > /proc/sys/kernel/pid_max
# cat /proc/sys/kernel/pid_max
4194304
# echo 0x400001 > /proc/sys/kernel/pid_max
bash: echo: write error: Invalid argument
```

internal/oci/container_test.go
haircommander: personally, I've had trouble running these tests as nonroot anyway

kolyshkin: You mean, the already existing test cases are failing when run as non-root, so there's no point caring about the ones you add?

Fine my be then.

internal/oci/runtime_oci.go
```diff
@@ -711,7 +711,6 @@ func (r *runtimeOCI) UpdateContainerStatus(c *Container) error {
 		}
 		out, err := cmd.Output()
 		if err != nil {
-			logrus.Errorf("Failed to update container state for %s: %v", c.id, err)
```
kolyshkin, line 714: Why is this line dropped?

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/containers/image/v5 v5.4.4
 	github.com/containers/libpod v1.9.2
 	github.com/containers/ocicrypt v1.0.2
+	github.com/containers/psgo v1.4.0
```
kolyshkin, line 19: Hmm, there should also be an entry in `go.sum` which I don't see in this commit, as well as some new files under `vendor/`.

internal/lib/sandbox/namespaces_test.go
```diff
@@ -495,6 +507,8 @@ func setupInfraContainerWithPid(pid int) {
 	cstate.State = specs.State{
 		Pid: pid,
 	}
+	// eat error here because callers may send invalid pids to test against
+	cstate.SetProcessInformation(pid) // nolint:errcheck
```
kolyshkin, line 511:
```suggestion
	_ := cstate.SetProcessInformation(pid) // nolint:errcheck
```
looks a tad more explicit

internal/oci/container.go
```diff
@@ -152,6 +160,12 @@ func (c *Container) StopSignal() syscall.Signal {
 }
 
 // FromDisk restores container's state from disk
+// Calls to FromDisk should always be proceeded by call to Runtime.UpdateContainerStatus
```
kolyshkin, line 163: did you mean preceded?

```suggestion
// Calls to FromDisk should always be preceded by call to Runtime.UpdateContainerStatus.
```

internal/oci/container.go
```diff
@@ -347,18 +390,108 @@ func (c *Container) exitFilePath() string {
...
+
+// Pid is the external, locking function that returns the container's
+// pid, if it's running, and actually the container process
+func (c *Container) Pid() (int, error) {
+	c.opLock.Lock()
+	defer c.opLock.Unlock()
+	return c.pid()
+}
+
+// pid is the internal, non-locking version of pid. It checks
```
kolyshkin, line 410:
```suggestion
// pid is the internal, non-locking version of Pid. It checks
```

internal/oci/container_test.go
haircommander: I don't personally think the unit tests need to be this precise

internal/oci/container.go
```diff
@@ -347,18 +390,108 @@ func (c *Container) exitFilePath() string {
...
+
+	if err := c.verifyPid(pid); err != nil {
+		return -1, err
+	}
+	return pid, nil
+}
+
+// findAndReleasePid attempts to find the container's pid
+// and if it's found, release it. In between these steps,
+// it also verified the pid running is the same as was originally
```
kolyshkin, line 439:
```suggestion
// it also verifies the pid running is the same as was originally
```

internal/oci/container.go
```diff
@@ -347,18 +390,108 @@ func (c *Container) exitFilePath() string {
...
+// started by the runtime
+func (c *Container) findAndReleasePid() (bool, error) {
+	pid := c.state.InitialPid
+	found := false
```
kolyshkin, line 443: IMHO using this variable in the code below decreases the readability, as you can use `return true` (or `false`) directly.

internal/oci/container.go
```diff
@@ -347,18 +390,108 @@ func (c *Container) exitFilePath() string {
...
+
+	if startTime != c.state.ProcessStartTime {
+		return errors.New("pid running but not the original container. pid wrap may have occurred")
+	}
+	return nil
+}
+
+// getProcessStartTime reads the kernel's /proc entry for stime for pid
+func getProcessStartTime(pid int) (string, error) {
+	psInfo, err := psgo.ProcessInfoByPids([]string{strconv.Itoa(pid)}, []string{"stime"})
```
kolyshkin, line 486: I would refrain from using `psgo` here. It reads five files in `/proc`, which is a bit too much for getting a single value.

If you only need a process start time, there is a trick you can use -- check any time (ctime, mtime, atime) of /proc/<PID> directory entry. This is a single call to `os.Stat()` and it will be way faster than parsing a few /proc files.

kolyshkin:
Left a few nits, and a suggestion to use ctime obtained from `os.Stat("/proc/"+strconv.Itoa(pid))` instead of using psgo.

There is a much bigger problem though. Using process start time won't work with CRIU (checkpoint/restore). In other words, if we checkpoint/restore the container, its start_time will be reset (to the restore time) and, with this PR, it won't be recognized.

If we're absolutely sure we won't support checkpoint/restore and CRIU, that is fine. Otherwise, we need to find other ways to make sure this is indeed our process. Unfortunately, I do not know of any easy way to do so.

kolyshkin:
OK, I think my comment about CRIU is not valid, if we're recreating container state upon restore.

So, I would
 - change psgo to stat on /proc/PID
 - make sure if works with checkpoint/restore

haircommander:
> Left a few nits, and a suggestion to use ctime obtained from `os.Stat("/proc/"+strconv.Itoa(pid))` instead of using psgo.
> 

great idea, done
> There is a much bigger problem though. Using process start time won't work with CRIU (checkpoint/restore). In other words, if we checkpoint/restore the container, its start_time will be reset (to the restore time) and, with this PR, it won't be recognized.
> 
> If we're absolutely sure we won't support checkpoint/restore and CRIU, that is fine. Otherwise, we need to find other ways to make sure this is indeed our process. Unfortunately, I do not know of any easy way to do so.

this is not a worry, until the CRI supports checkpointing/restoring, at which point we can address it. I find it unlikely

haircommander:
```
# time="2020-07-17T19:57:01Z" level=error msg="removing the pod sandbox \"e67945101ed9156978b11fb9a633a4414baaa2ea723b99b94e0b62be0cbf07b5\" failed: rpc error: code = Unknown desc = unable to remove managed namespaces: Removing namespaces encountered the following errors [unlinkat /var/run/utsns/1311f64c-70cc-404e-b4c7-e83275f80028: device or resource busy]"
```
[another one](https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3868/test_pull_request_crio_integration_rhel/13859/artifacts/testout.txt)
/retest

internal/oci/runtime_oci.go
haircommander: it's duplicated with error messages below

internal/oci/container_test.go
kolyshkin: My concern was if there are kernels out there that can have pid_max which is greater than the one you use in the test, the test would fail.

It looks like there are no such kernels (at least for now), so that is why I said it's OK to leave this as is.

internal/oci/container.go
```diff
@@ -344,6 +344,23 @@ func (c *Container) exitFilePath() string {
 	return filepath.Join(c.dir, "exit")
 }
 
+// IsRunning is a function that checks if a specific pid is running
```
kolyshkin, line 389: > if a specific pid is running

This is not a good choice of words. If by "pid" you mean a process, it can be in present but in a few other states than "running" (e.g. waiting for I/O to finish, traced, etc.).

I think it would be more correct to say "if a container's init pid exists" (or "present").

internal/oci/runtime_oci.go
kolyshkin: I see

kolyshkin:
I still see some remnants of psgo in the second commit

haircommander:
> I still see some remnants of psgo in the second commit

fixed!

internal/oci/container.go
haircommander: adapted as suggested!

internal/oci/container.go
kolyshkin: this one is still not addressed. Not a big issue but would save a few lines code (which is always good)

internal/oci/container.go
```diff
@@ -346,6 +346,23 @@ func (c *Container) exitFilePath() string {
 	return filepath.Join(c.dir, "exit")
 }
 
+// IsRunning is a function that checks if a container's init pid exists.
+// It is used to check a container state when we don't want a `$runtime state` call
+func (c *Container) IsRunning() bool {
```
kolyshkin, line 394: I am thinking, maybe change this to `IsAlive` or `IsInitAlive`, or something like this, as `IsRunning` is a bit misleading

internal/oci/container.go
```diff
@@ -25,8 +26,9 @@ import (
 const defaultStopSignalInt = 15
 
 var (
-	defaultStopSignal   = strconv.Itoa(defaultStopSignalInt)
-	ErrContainerStopped = errors.New("container is already stopped")
+	defaultStopSignal      = strconv.Itoa(defaultStopSignalInt)
+	ErrContainerStopped    = errors.New("container is already stopped")
+	errContainerPidStopped = errors.Errorf("insufficient ps information; pid likely stopped")
```
kolyshkin, line 31: 1. `s/errors.Errorf/errors.New/`
2. "pid is stopped" seems confusing to me (I think of it as if the process was stopped by SIGSTOP). Something like "no such pid" or "process unexpectedly gone" would be better.

internal/oci/container.go
```diff
@@ -72,12 +74,18 @@ type ContainerVolume struct {
...
-	Error     string    `json:"error,omitempty"`
+	Created    time.Time `json:"created"`
+	Started    time.Time `json:"started,omitempty"`
+	Finished   time.Time `json:"finished,omitempty"`
+	ExitCode   *int32    `json:"exitCode,omitempty"`
+	OOMKilled  bool      `json:"oomKilled,omitempty"`
+	Error      string    `json:"error,omitempty"`
+	InitialPid int       `json:"initialPid,omitempty"`
+	// the unix start time of the process, as found by parsing
+	// the /proc/ entry for the contaienr pid.
```
kolyshkin, line 85: nit: `s/contaienr/container/`

internal/oci/container.go
```diff
@@ -72,12 +74,18 @@ type ContainerVolume struct {
...
-	OOMKilled bool      `json:"oomKilled,omitempty"`
-	Error     string    `json:"error,omitempty"`
+	Created    time.Time `json:"created"`
+	Started    time.Time `json:"started,omitempty"`
+	Finished   time.Time `json:"finished,omitempty"`
+	ExitCode   *int32    `json:"exitCode,omitempty"`
+	OOMKilled  bool      `json:"oomKilled,omitempty"`
+	Error      string    `json:"error,omitempty"`
+	InitialPid int       `json:"initialPid,omitempty"`
+	// the unix start time of the process, as found by parsing
```
kolyshkin, line 84: We're no longer parsing the /proc/<PID> entry...

internal/oci/container.go
```diff
@@ -72,12 +74,18 @@ type ContainerVolume struct {
...
+	// the /proc/ entry for the contaienr pid.
+	// This is used to track whether the pid we have stored
+	// is the same as the same pid number on the host
```
kolyshkin, line 87: nit: `s/$/./`

internal/oci/container.go
kolyshkin: and the error name itself should be `errContainerInitGone` or something similar (`errContainerInitPidNotFound`?)

internal/oci/container.go
```diff
@@ -72,12 +74,18 @@ type ContainerVolume struct {
...
-	ExitCode  *int32    `json:"exitCode,omitempty"`
-	OOMKilled bool      `json:"oomKilled,omitempty"`
-	Error     string    `json:"error,omitempty"`
+	Created    time.Time `json:"created"`
+	Started    time.Time `json:"started,omitempty"`
+	Finished   time.Time `json:"finished,omitempty"`
+	ExitCode   *int32    `json:"exitCode,omitempty"`
+	OOMKilled  bool      `json:"oomKilled,omitempty"`
+	Error      string    `json:"error,omitempty"`
+	InitialPid int       `json:"initialPid,omitempty"`
```
kolyshkin, line 83: Suggestion: `s/InitialPid/InitPid/`

internal/oci/container.go
```diff
@@ -72,12 +74,18 @@ type ContainerVolume struct {
...
+	// the unix start time of the process, as found by parsing
+	// the /proc/ entry for the contaienr pid.
+	// This is used to track whether the pid we have stored
+	// is the same as the same pid number on the host
+	ProcessStartTime string `json:"processStartTime,omitempty"`
```
kolyshkin, line 88: similar here: `s/ProcessStartTime/InitStartTime/`.

internal/oci/container.go
```diff
@@ -349,18 +392,110 @@ func (c *Container) exitFilePath() string {
...
+
+// getProcessStartTime reads the kernel's /proc entry for stime for pid
+func getProcessStartTime(pid int) (string, error) {
+	st, err := os.Stat(fmt.Sprintf("/proc/%d", pid))
+	if err != nil {
+		return "", errContainerPidStopped
+	}
+
+	sysInfo, ok := st.Sys().(*syscall.Stat_t)
+	if !ok {
```
kolyshkin, line 494: this is a no-op as os.Stat() always return `syscall.Stat_t`.

You can avoid all this checking if you use `unix.Stat` instead.

internal/oci/container.go

internal/oci/container.go

internal/oci/container.go

internal/oci/container.go
kolyshkin: or `IsInitExists` as the documentation says

internal/oci/container.go
kolyshkin: does not look to be addressed

kolyshkin:
Ideally, you should avoid having the second commit fix the changes in the first one (unless it's logical).

For example, when the first commit introduces a line like this:
```go
logrus.Errorf("error getting pid %d status %v", pid, err)
```
and then the second commit fixes it to add the missing comma:
```go
logrus.Errorf("error getting pid %d status: %v", pid, err)
```
and both commits belong to the same PR, it's clear that it's the first commit who should have done it right from the start.

Otherwise the git history is unnecessary complex and it is harder to read (and the PR itself is harder to review).

haircommander:
> Ideally, you should avoid having the second commit fix the changes in the first one (unless it's logical).
> 
> For example, when the first commit introduces a line like this:
> 
> ```go
> logrus.Errorf("error getting pid %d status %v", pid, err)
> ```
> 
> and then the second commit fixes it to add the missing comma:
> 
> ```go
> logrus.Errorf("error getting pid %d status: %v", pid, err)
> ```
> 
> and both commits belong to the same PR, it's clear that it's the first commit who should have done it right from the start.
> 
> Otherwise the git history is unnecessary complex and it is harder to read (and the PR itself is harder to review).

alright, I think I cleaned it up

internal/oci/container.go
```diff
@@ -155,6 +161,12 @@ func (c *Container) StopSignal() syscall.Signal {
 }
 
 // FromDisk restores container's state from disk
+// Calls to FromDisk should always be preceded by call to Runtime.UpdateContainerStatus.
+// This is because FromDisk() initializes the InitStartTime for the saved container state
+// when CRI-O is being upgraded to a version that supports tracking pid,
+// but does no verification the container is actually still running. If we assume the container
+// is still running, we could incorrectly think a process with the same pid running on the host
+// is our container. a call to `$runtime state` will protect us against this.
```
kolyshkin, line 169: nit: `s/a call/A call/`.

internal/oci/container.go
```diff
@@ -163,7 +175,36 @@ func (c *Container) FromDisk() error {
...
+// SetInitPid initializes the InitPid and InitStartTime for the container state
+// given a pid.
+// These values should be set once, and not changed again.
+func (cstate *ContainerState) SetInitPid(pid int) (err error) {
+	if cstate.InitPid != 0 || cstate.InitStartTime != "" {
+		return errors.Errorf("pid and start time already initialized: %d %s", cstate.InitPid, cstate.InitStartTime)
+	}
+	cstate.InitPid = pid
+	cstate.InitStartTime, err = getInitStartTime(pid)
+	if err != nil {
```
kolyshkin, line 204: This is useless, since we use a named return value (`err`) and it is already initialized.

Note: in the spirit of that recent PR that renames `err` to `retErr` I would use `retErr` here as well (or not use a named return value at all -- up to you).

internal/oci/container.go
```diff
@@ -350,14 +391,105 @@ func (c *Container) exitFilePath() string {
...
+	_, err := c.pid()
+	if err != nil {
+		logrus.Errorf("checking if pid of %s is running failed: %v", c.id, err)
 		return false
 	}
 
 	return true
 }
 
+// Pid is the external, locking function that returns the container's
```
kolyshkin, line 403: Since the name starts with an uppercased letter, it's obvious that it is an external function, so you can omit saying it.
You can probably omit the "locking" part, too, since it's an implementation detail.

internal/oci/container.go
```diff
@@ -350,14 +391,105 @@ func (c *Container) exitFilePath() string {
...
+	if pid <= 0 {
+		return -1, errors.New("pid not initialized")
+	}
+
+	// container has stopped (as pid is initialized but the runc state has overwritten it)
+	if c.state.Pid == 0 {
+		return 0, findprocess.ErrNotFound
+	}
+
+	if _, err := findprocess.FindProcess(pid); err != nil {
```
kolyshkin, line 428: It seems that this step is redundant, since you're going to open `/proc/<pid>/` entry anyway.

internal/oci/container.go
```diff
@@ -350,14 +391,105 @@ func (c *Container) exitFilePath() string {
...
+	pid := c.state.InitPid
+
+	if err := c.verifyPid(pid); err != nil {
+		if !errors.Is(err, errContainerPidGone) {
+			return false, err
+		}
+		return false, nil
+	}
+
+	process, err := findprocess.FindProcess(pid)
```
kolyshkin, line 452: same here

internal/oci/container.go
```diff
@@ -79,6 +80,11 @@ type ContainerState struct {
 	ExitCode  *int32    `json:"exitCode,omitempty"`
 	OOMKilled bool      `json:"oomKilled,omitempty"`
 	Error     string    `json:"error,omitempty"`
+	InitPid   int       `json:"initPid,omitempty"`
+	// The unix start time of the container's init PID.
+	// This is used to track whether the PID we have stored
+	// is the same as the corresponding PID on the host.
+	InitStartTime string `json:"initStartTime,omitempty"`
```
kolyshkin, line 87: It never occurred to be before, but why are we using a `string` rather than `int64` here. Using an integer would be easier (no conversions), faster (to compare two ints rather than two strings), and it will look almost the same in json anyway.

internal/oci/container.go
```diff
@@ -350,14 +391,105 @@ func (c *Container) exitFilePath() string {
...
+}
+
+// getInitStartTime reads the kernel's /proc entry for stime for pid.
+func getInitStartTime(pid int) (string, error) {
+	var st unix.Stat_t
+	if err := unix.Stat(fmt.Sprintf("/proc/%d", pid), &st); err != nil {
+		return "", errors.Wrapf(errContainerPidGone, err.Error())
+	}
+
+	return strconv.FormatInt(st.Ctim.Nano(), 10), nil
```
kolyshkin, line 490: Nanosecond precision seems redundant in here -- I doubt it is possible to have a pid overlap happen in under 1 second.

internal/oci/container.go
```diff
@@ -350,14 +391,105 @@ func (c *Container) exitFilePath() string {
...
+	}
+
+	if startTime != c.state.InitStartTime {
+		return errors.New("pid running but not the original container. pid wrap may have occurred")
+	}
+	return nil
+}
+
+// getInitStartTime reads the kernel's /proc entry for stime for pid.
+func getInitStartTime(pid int) (string, error) {
```
kolyshkin, line 484: As suggested earlier, maybe have this return int64 (aka "unix time").

kolyshkin:
> alright, I think I cleaned it up

Thank you, it is much easier to review now. Left a few more comments (that I should have left earlier)

internal/oci/container.go
haircommander: I am not actually sure what house keeping process.Release() does, but it was there before. Without this call, we don't know what process we have to release. Do you think we should drop process.Release() here too?

internal/oci/container.go

internal/oci/container.go
haircommander: kernel.pid_max == 1 :thinking: 

haha, you're right, fixed

internal/oci/container.go

internal/oci/container.go

haircommander:
@kolyshkin updated, though I skipped https://github.com/cri-o/cri-o/pull/3868#discussion_r461178807 because I believe we still need it

/retest

internal/oci/container.go
```diff
@@ -347,17 +388,104 @@ func (c *Container) exitFilePath() string {
...
+	}
+
+	if startTime != c.state.InitStartTime {
+		return errors.New("PID running but not the original container. PID wrap may have occurred")
+	}
+	return nil
+}
+
+// getInitStartTime reads the kernel's /proc entry for stime for PID.
+func getInitStartTime(pid int) (int, error) {
```
kolyshkin, line 480: The word `Init` in the function name is misleading, since the function doesn't care if the `pid` is that of init or not.

internal/oci/oci_linux.go
```diff
@@ -85,8 +85,12 @@ func (r *runtimeOCI) containerStats(ctr *Container, cgroup string) (*ContainerSt
...
 	stats.BlockInput, stats.BlockOutput = calculateBlockIO(cgroupStats)
 
-	if ctr.state != nil {
-		netNsPath := fmt.Sprintf("/proc/%d/ns/net", ctr.state.Pid)
+	// Try our best to get the net namespace path.
+	// If pid() errors, the container has stopped, and the /proc entry
+	// won't exist anyway.
+	pid, _ := ctr.pid() // nolint:errcheck
+	if pid > 0 {
+		netNsPath := fmt.Sprintf("/proc/%d/ns/net", pid)
```
mrunalp, line 93: Why don't we return the pinned path here? We can get rid of the pid check then.

haircommander: a couple of reasons:

primarily: we don't have direct access to the Sandbox object, which owns the namespaces. We'd have to ask the server what the sandbox is, and then ask  the sandbox what the namespace is.
secondarily, is it possible to have a private net namespace? if so, the sandbox level namespace won't be the correct one

mrunalp: Atleast for the pinned ns case.

mrunalp: There is no private net namespaces in a pod. All the containers always share the pod network namespace.

haircommander:
@kolyshkin pointed out that most of findprocess is now obsolete with the existence of verifyPid(). I have removed it and associated calls

internal/oci/oci_linux.go
haircommander: ah, well we still have the access problem. I would replumb this as a follow up, if we'd like

internal/lib/sandbox/namespaces.go
```diff
@@ -303,9 +316,9 @@ func infraPid(infra *oci.Container) int {
 // calling a container.State() in batch operations
 func nsPathGivenInfraPid(ns NamespaceIface, nsType NSType, infraPid int) string {
 	// caller is responsible for checking if infraContainer
-	// is valid. If not, infraPid should be negative
+	// is valid. If not, infraPid should be less than or equal to 0
 	if ns == nil || ns.Get() == nil {
-		if infraPid >= 0 {
+		if infraPid > 0 {
```
TomSweeneyRedHat, line 321: this doesn't match the new comment on 319.  Did you mean to remove the `=` here?

haircommander: the comment is meant to say "if it's not valid". is the wording unclear?

kolyshkin: I also had to read this part twice, and it is correct, including the comment, but maybe not super clear. Not sure how to improve it though :(

server/container_execsync.go
```diff
@@ -17,12 +16,7 @@ func (s *Server) ExecSync(ctx context.Context, req *pb.ExecSyncRequest) (*pb.Exe
 		return nil, status.Errorf(codes.NotFound, "could not find container %q: %v", req.ContainerId, err)
 	}
 
-	if err := s.Runtime().UpdateContainerStatus(c); err != nil {
-		return nil, err
-	}
-
-	cState := c.State()
-	if !(cState.Status == oci.ContainerStateRunning || cState.Status == oci.ContainerStateCreated) {
+	if !c.IsAlive() {
```
kolyshkin, line 19: maybe

```go
if _, err := c.verifyPid(); err != nil {
```

and remove the whole `IsAlive` function?

Or, use `IsAlive` in all the places where you're using the `if` like the one in this comment. Otherwise it's a duplicated functionality.

haircommander: the nice thing about the different functions has to do with error handling. In the case of IsAlive, I believe it should be used in cases we presume the container to be alive, and it would be an error if it was not.

Conversely, verifyPid does not make an assertion of what the liveliness of the container should be

For instance, all uses in runtime_oci.go assume the container will eventually not be alive. in that case, we wouldn't want to print an error that the container wasn't alive

I think it is correct to error in `server/container_execsync.go`, as the client asked to exec in a container that is not running. Propagating the reason for this failure (pid wrap, container exited, we exec'ed too early somehow) is useful.

Do you have suggestions on how to balance these two use cases and error scenerios @kolyshkin

Issue #3872, opened 2020-06-13
Missing MKNOD capability makes rancher setup fail


kubealex:
**Description**

Some tools require creation of files/paths with MKNOD, like Rancher, and it fails hard when trying to install on k8s with cri-o.

**Steps to reproduce the issue:**
1. Install k8s
2. Install rancher
3. Enjoy failure on startup

**Describe the results you received:**

2020/05/20 08:38:26 [INFO] Rancher version v2.4.3 (684884c) is starting
2020/05/20 08:38:26 [INFO] Rancher arguments {ACMEDomains:[] AddLocal:auto Embedded:false HTTPListenPort:80 HTTPSListenPort:443 K8sMode:auto Debug:false Trace:false NoCACerts:false AuditLogPath:/var/log/auditlog/rancher-api-audit.log AuditLogMaxage:10 AuditLogMaxsize:100 AuditLogMaxbackup:10 AuditLevel:0 Features:}
2020/05/20 08:38:26 [INFO] Listening on /tmp/log.sock
2020/05/20 08:38:26 [INFO] Starting API controllers
2020/05/20 08:38:26 [INFO] Running in clustered mode with ID 10.85.0.52, monitoring endpoint cattle-system/rancher
2020/05/20 08:38:27 [INFO] Starting API controllers
2020/05/20 08:38:27 [INFO] Starting rbac.authorization.k8s.io/v1, Kind=Role controller
2020/05/20 08:38:27 [INFO] Starting rbac.authorization.k8s.io/v1, Kind=ClusterRole controller
2020/05/20 08:38:28 [FATAL] error running the jail command: exit status 2

From a debug pod:

tar: ./urandom: Cannot mknod: Operation not permitted
tar: ./zero: Cannot mknod: Operation not permitted
tar: ./tty: Cannot mknod: Operation not permitted
tar: ./full: Cannot mknod: Operation not permitted
tar: ./random: Cannot mknod: Operation not permitted
tar: ./null: Cannot mknod: Operation not permitted
tar: Exiting with failure status due to previous errors

**Describe the results you expected:**

Setup completes

**Additional information you deem important (e.g. issue happens only occasionally):**

Happens everytime, doesn't happen on docker or containerd.

**Output of `crio --version`:**

```
[kube@k8s-master-0 ~]$ crio --version
crio version 
Version:       1.18.1
GitCommit:     5cbf694c34f8d1af19eb873e39057663a4830635
GitTreeState:  clean
[...]

PR #3873, opened 2020-06-13
Add MKNOD to capabilities 

kubealex:
Signed-off-by: Alessandro Rossi <al.rossi87@gmail.com>

#### What type of PR is this?

 /kind feature

Removing MKNOD from capabilities doesn't allow installation of Rancher, for instance.

https://github.com/cri-o/cri-o/issues/3872

No

### DIFF
--- a/internal/config/capabilities/capabilities.go
+++ b/internal/config/capabilities/capabilities.go
@@ -23,6 +23,7 @@ func Default() Capabilities {
 		"SETPCAP",
 		"NET_BIND_SERVICE",
 		"KILL",
+                "MKNOD",
 	}
 }
 


internal/config/capabilities/capabilities.go
```diff
@@ -23,6 +23,7 @@ func Default() Capabilities {
 		"SETPCAP",
 		"NET_BIND_SERVICE",
 		"KILL",
+                "MKNOD",
```
TomSweeneyRedHat, line 26: looks like a tab/space issue here

mrunalp:
I think this can be solved by using a drop-in to add this capability where you need it. This isn't something that we should enable by default. We want to reduce the capabilities we need at runtime
over time.

rhatdan:
These were removed intentionally since 99.99% of all containers in the world do NOT need CAP_AUDIT_WRITE, CAP_MKNOD, CAP_NET_RAW, CAP_SYS_CHROOT.  We have removed them from the default CRI-O config.  If you need them then you can add them back in, or better yet specify them in the kubernetes.yaml config for the small subset of containers/pods that need them.

The general case should be as tight as possible, and not be limited to what someone decided at Docker upstream 7 years ago.

PR #3874, opened 2020-06-14
Remove the protocol filter from the portMappings constructor

janosi:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

There is a protocol filter in the current code that filters out all host ports that do not have TCP or UDP as protocol defined. Originally this protocol filter was introduced to prevent the loading of the SCTP kernel module on the nodes. But the iptables chain creation alone does not load the kernel module. The module would be loaded if an SCTP socket was created. The host port manager imported by CRI-O already filters out the SCTP ports from the set of ports on which it starts listening. As there is no socket operation (listening) on the SCTP ports the kernel module is not loaded. We can remove this filter here so the iptables chains will be created for SCTP host ports.

None

```release-note
Container HostPort with SCTP protocol is supported.
```

### DIFF
--- a/server/sandbox_run.go
+++ b/server/sandbox_run.go
@@ -74,9 +74,6 @@ func convertPortMappings(in []*pb.PortMapping) []*hostport.PortMapping {
 		if v.HostPort <= 0 {
 			continue
 		}
-		if v.Protocol != pb.Protocol_TCP && v.Protocol != pb.Protocol_UDP {
-			continue
-		}
 		out = append(out, &hostport.PortMapping{
 			HostPort:      v.HostPort,
 			ContainerPort: v.ContainerPort,


server/sandbox_run.go
```diff
@@ -74,9 +74,6 @@ func convertPortMappings(in []*pb.PortMapping) []*hostport.PortMapping {
 		if v.HostPort <= 0 {
 			continue
 		}
-		if v.Protocol != pb.Protocol_TCP && v.Protocol != pb.Protocol_UDP {
-			continue
-		}
```
TomSweeneyRedHat, line 79: I know there are only 3 values defined in the map now, but is there any value to keeping this but adding Protocol_SCTP to guard against a future addition that might clonk CRI-O?  I lean towards no, that is this change is fine as is, but thought I'd ask just in case.

TomSweeneyRedHat:
one question for consideration, otherwise LGTM

server/sandbox_run.go
haircommander: I think if there's an expectation new map values succeed in CRI-O, I'd rather get clonked (and fix it) if they fail than skip them

haircommander:
/lgtm

thanks @janosi

PR #3875, opened 2020-06-15
[1.17] oci: check state before stop atomically

haircommander:
#### What type of PR is this?

/kind flake

in stop, we need to check the pod's status before stopping it.
However, as it stands, we do it in two distinct steps: State() followed by ContainerStop()

There exists a race where:
the State() call happens
followed by an UpdateContainerStatus() (marking the container as stopped)
followed by the Container Stop()

we've hit this race quite a bit in circleCI running 'pod stop idempotent with ctrs already stopped'

fix this by moving the check branching on Status to after we take the opLock()

```release-note

```

### DIFF
--- a/internal/lib/stop.go
+++ b/internal/lib/stop.go
@@ -16,15 +16,17 @@ func (c *ContainerServer) ContainerStop(ctx context.Context, container string, t
 	}
 	ctrID := ctr.ID()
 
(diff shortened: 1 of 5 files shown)


haircommander:
/test launch-aws

:upside_down_face:

umohnani8:
LGTM
/retest

Issue #3845, opened 2020-06-07
Strange output from version


afbjorklund:
Previous output:

```console
                         _             _            
            _         _ ( )           ( )           
  ___ ___  (_)  ___  (_)| |/')  _   _ | |_      __  
/' _ ` _ `\| |/' _ `\| || , <  ( ) ( )| '_`\  /'__`\
| ( ) ( ) || || ( ) || || |\`\ | (_) || |_) )(  ___/
(_) (_) (_)(_)(_) (_)(_)(_) (_)`\___/'(_,__/'`\____)
$ crio --version
crio version 1.17.1
```
Updated output:

```console
$ crio --version
WARN[0000] Encountered error detecting link mode of binary: `ldd /usr/bin/crio` failed:   (exec: "ldd": executable file not found in $PATH) 
WARN[0000] Encountered error detecting link mode of binary: `ldd /usr/bin/crio` failed:   (exec: "ldd": executable file not found in $PATH) 
crio version 
Version:    1.18.1
GoVersion:  go1.13.8
Compiler:   gc
Platform:   linux/amd64

```

I _tried_ to supply the GIT_COMMIT also with the new make variables, but that info seems missing.
That is, it is not available when building the release from the tarball but only when building from git.

Laptop output:
```
crio version 
Version:       1.18.1
GitCommit:     5cbf694c34f8d1af19eb873e39057663a4830635
GitTreeState:  clean
BuildDate:     2020-06-07T15:03:33Z
GoVersion:     go1.13.8
Compiler:      gc
Platform:      linux/amd64
Linkmode:      dynamic
```

`make COMMIT_NO=5cbf694c34f8d1af19eb873e39057663a4830635 PREFIX=/usr binaries`

Is there something more needed ? Are there other variables to supply this version information ?

```console
 file out/buildroot/output/target/usr/bin/crio
out/buildroot/output/target/usr/bin/crio: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 5.4.0, Go BuildID=pnqFhA7h8z8BERVxjB6D/80jgCVpBQ1L4XCb7mONC/tDSx0k9EPYOozoKN1jb6/vx4KwRTOsbebCXcsecT9, stripped
 ldd out/buildroot/output/target/usr/bin/crio
	linux-vdso.so.1 (0x00007fff27d94000)
[...]

PR #3886, opened 2020-06-17
Restore version output from crio --version

afbjorklund:
The version was missing from the first output line,
ldd warnings were printed and the link information
was missing (due to reusing the $LDFLAGS variable).
This broke compatibility with earlier crio versions.

Cherry-picked to `master` from PR #3845


haircommander:
/ok-to-test

LGTM

PR #3887, opened 2020-06-17
server: re-add gocyclo skip

haircommander:
Signed-off-by: Peter Hunt <pehunt@redhat.com>

#### What type of PR is this?

/kind failing-test
-

#### Does this PR introduce a user-facing change?

```release-note

```

PR #3889, opened 2020-06-17
runtime_vm: Apply the correct label before the sandbox is created

fidencio:
#### What type of PR is this?

/kind bug

Although the current code does apply the correct label, it does it too
late, after the Sandbox has been created and the processes already being
ran.

None

Testing this PR is not something easy without a previously setup environment.
As I have one handy, please, let me know via a private message and I can give you access to it.

```release-note
None
```

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -9,11 +9,13 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containers/libpod/pkg/annotations"
+	selinux "github.com/containers/libpod/pkg/selinux"
 	"github.com/containers/storage"
 	"github.com/cri-o/cri-o/internal/config/node"
 	"github.com/cri-o/cri-o/internal/lib"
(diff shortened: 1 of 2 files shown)


fidencio:
@rhatdan, @haircommander, please, take a look.

haircommander:
can you fix the issues with
`make lint`?
also, is there any harm in doing both? wouldn't a new container process running in the rm also need the correct labels?

fidencio:
> can you fix the issues with
> `make lint`?

Sure!

> also, is there any harm in doing both? wouldn't a new container process running in the vm also need the correct labels?

There's no harm, @haircommander, but it's not useful either. The reason for that is kata not applying any SELinux labels on the VM workloads: https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L1102

What's your preference?

haircommander:
ah I see, can you add a comment making note of that?

fidencio:
> ah I see, can you add a comment making note of that?

Sure!

haircommander:
nvm on the lint, the only lint issue is fixed in https://github.com/cri-o/cri-o/pull/3887

haircommander:
integration failures are not the fault of this PR

server/sandbox_run_linux.go
```diff
@@ -482,6 +484,23 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 	if err != nil {
 		return nil, err
 	}
+
+	runtimeType, err := s.Runtime().ContainerRuntimeType(container)
+	if err != nil {
+		return nil, err
+	}
+	// If using kata runtime, the process label should be set to container_kvm_t
+	// Mind that kata does *not* apply any process label to the VM workloads.
```
haircommander, line 493: suggested rewording:
`Keep in mind that kata does *not* apply any process label to containers within the VM`

fidencio: Suggestion taken!

haircommander:
test failures are not the fault of this PR, but are legit. they're fixed by https://github.com/opencontainers/runc/pull/2479

fidencio:
I've rebased this one atop of the origin/main and forced pushed, in order to see if it improves the issues with the CI which are not related to the PR itself.

PR #3891, opened 2020-06-18
fix segfault with errors.Cause()

haircommander:
#### What type of PR is this?

/kind bug

a strange interaction between the go compiler and errors.Cause() seems to happen.
Sometimes, the condition `err != nil` is checked after `errors.Cause(err)...`
since errors.Cause(err) segfaults if err == nil, we need to put the second check in a new block

it sucks, but we've seen a segfault like in https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3887/test_pull_request_crio_e2e_fedora/13716/artifacts/journals/crio.service once already

```release-note

```


TomSweeneyRedHat:
LGTM
but it looks like you picked up a lint error.

umohnani8:
Well that is annoying, great find! LGTM
/retest

haircommander:
/hold

I am going to experiment with errors.Is() instead, as this is a drag :)

PR #3903, opened 2020-06-23
Add runtime_type as an option of "--runtimes"

### LINKED ISSUE #3877: Is there a way to setup runtime type in CRI-O tests?
We are trying to run the tests from CRI-O like https://github.com/cri-o/cri-o/blob/master/test/ctr.bats, however, for kata containers 2.0, we can performed the export of the RUNTIME to `containerd-shim-kata-v2` without a problem but how can we performed a export for `runtime_type = "vm"`? 
/cc @chavafg

fidencio:
#### What type of PR is this?

/kind feature

Let's add a new field to the "--runtimes" option, allowing consumers of
the cli to also define the type of the runtime.

This is particularly useful to ease testing of the "vm" runtime, used by
kata-containers.

Mind that this PR does *not* break the command-line backwards
compatibility as we do not mandate 4 fields to be defined, but in case a
4th field is define, it'll be treated as the runtime_type.

Fixes #3877

```release-note
Adds a new optional field, runtime_type, to the "--runtimes" option.
```

### DIFF
--- a/internal/criocli/criocli.go
+++ b/internal/criocli/criocli.go
@@ -128,12 +128,21 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
 		runtimes := StringSliceTrySplit(ctx, "runtimes")
 		for _, r := range runtimes {
 			fields := strings.Split(r, ":")
-			if len(fields) != 3 {
+
+			runtimeType := libconfig.DefaultRuntimeType
+
+			switch len(fields) {
+			case 4:
+				runtimeType = fields[3]
+				fallthrough
+			case 3:
+				config.Runtimes[fields[0]] = &libconfig.RuntimeHandler{
+					RuntimePath: fields[1],
+					RuntimeRoot: fields[2],
+					RuntimeType: runtimeType,
+				}
+			default:
 				return fmt.Errorf("wrong format for --runtimes: %q", r)
-			}
-			config.Runtimes[fields[0]] = &libconfig.RuntimeHandler{
-				RuntimePath: fields[1],
-				RuntimeRoot: fields[2],
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


internal/criocli/criocli.go
```diff
@@ -128,13 +128,23 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
...
 				RuntimeRoot: fields[2],
 			}
+
+			// Only in case the 4th field was passed to the command-line, assume
+			// it's the RuntimeType and set it accordingly in the RuntimeHandler
+			if len(fields) == 4 {
+				config.Runtimes[fields[0]] = &libconfig.RuntimeHandler{
+					RuntimeType: fields[3],
+				}
+			}
```
haircommander, line 147: can we use a switch statement for this?

fidencio: Please, let me know if this change is what you had in mind with your suggestion.

internal/criocli/criocli.go
```diff
@@ -128,13 +128,23 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
 		runtimes := StringSliceTrySplit(ctx, "runtimes")
 		for _, r := range runtimes {
 			fields := strings.Split(r, ":")
-			if len(fields) != 3 {
+
+			runtimeType := "oci"
+
+			nFields := len(fields)
+			switch nFields {
```
haircommander, line 135: nit: you can just do switch len(fields)

haircommander:
one more nit, otherwise LGTM
/approve

internal/criocli/criocli.go
fidencio: Changed!

haircommander:
/retest

LGTM

internal/criocli/criocli.go
```diff
@@ -128,13 +128,22 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
 		runtimes := StringSliceTrySplit(ctx, "runtimes")
 		for _, r := range runtimes {
 			fields := strings.Split(r, ":")
-			if len(fields) != 3 {
+
+			runtimeType := "oci"
```
saschagrunert, line 132: Please make this a global const. Or do we already have this somewhere defined?

fidencio: @saschagrunert, I've addressed this by using `libconfig.DefaultRuntimeType` instead of "oci".

Let me know if this is what you had in mind.

haircommander:
:angry: 

/lgtm

c'mon ci...

PR #3911, opened 2020-06-30
test: use correct flags instead of deprecated env var

haircommander:
#### What type of PR is this?

/kind flake

so that kubetest stops yelling at us

```release-note
NONE
```


haircommander:
note: this *may* not actually fix the problem...

haircommander:
I am starting to think it's go 1.14.4...

haircommander:
ooh e2e_cgroupv2 (with crun) is passing, I bet it's go 1.14.4 + runc

TomSweeneyRedHat:
Changes LGTM, but a rebase is needed now

PR #3924, opened 2020-07-02
runtime_vm: Ignore ttrpc.ErrClosed when shutting the container down

fidencio:
#### What type of PR is this?

/kind bug

When shutting the container down, we're dealing with the following piece
of code on Kata side:
https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/service.go#L785
```
func (s *service) Shutdown(ctx context.Context, r *taskAPI.ShutdownRequest) (_ *ptypes.Empty, err error) {
	defer func() {
		err = toGRPC(err)
	}()

	s.mu.Lock()
	if len(s.containers) != 0 {
		s.mu.Unlock()
		return empty, nil
	}
	s.mu.Unlock()

	s.cancel()

	os.Exit(0)

	// This will never be called, but this is only there to make sure the
	// program can compile.
	return empty, nil
}
```

The code shown above will simply stop the service, closing the ttrpc
channel, raising then the "ErrClosed" error, which is returned by the
Shutdown.

Differently from containerd code, which simply igores the error, CRI-O
propagates the error, leaving a bunch of processes behind that will
never ever be closed.

Here's what containerd does:
https://github.com/containerd/containerd/blob/master/runtime/v2/shim.go#L194
```
        _, err := s.task.Shutdown(ctx, &task.ShutdownRequest{
                ID: s.ID(),
        })
        if err != nil && !errors.Is(err, ttrpc.ErrClosed) {
                return errdefs.FromGRPC(err)
        }
```

Knowing that, let's mimic what's been done by containerd and ignore the
error in this specific case.

None

Again, not so trivial case to reproduce. If needed, I have an OCP cluster already setup that could be loaned for validating this issue.

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -540,7 +540,7 @@ func (r *runtimeVM) DeleteContainer(c *Container) error {
 		return err
 	}
 
-	if _, err := r.task.Shutdown(r.ctx, &task.ShutdownRequest{ID: c.ID()}); err != nil {
+	if _, err := r.task.Shutdown(r.ctx, &task.ShutdownRequest{ID: c.ID()}); err != nil && !errors.Is(err, ttrpc.ErrClosed) {
 		return err
 	}
 


haircommander:
/ok-to-test

e2e tests will fail and that's not the fault of this PR

haircommander:
/approve

LGTM, thanks @fidencio

giuseppe:
/retest

LGTM

haircommander:
hey @fidencio tests are fixed please rebase

internal/oci/runtime_vm.go
```diff
@@ -540,7 +540,7 @@ func (r *runtimeVM) DeleteContainer(c *Container) error {
 		return err
 	}
 
-	if _, err := r.task.Shutdown(r.ctx, &task.ShutdownRequest{ID: c.ID()}); err != nil {
+	if _, err := r.task.Shutdown(r.ctx, &task.ShutdownRequest{ID: c.ID()}); err != nil && !errors.Is(err, ttrpc.ErrClosed) {
```
TomSweeneyRedHat, line 543: might be nice to have a code comment here on why ttrpc.ErrClosed is special.

fidencio: I can add a comment, but that's well detailed in the commit message.
Let me know your preference, but I have to admit I'm afraid of touching this PR again as, for the first time, CI seems to have passed. :-)

PR #3929, opened 2020-07-06
Managed NS: update default configuration values

haircommander:
#### What type of PR is this?

/kind api-change

This PR does three things:
- set managed ns lifecycle to true by default
- finishes deprecation of ManageNetworkNSLifecycle flag and config value (set as deprecated in 4.4)
- fixes a small test inconsistency with how we restore a managed ns pod. Before, we removed the managed namespaces on sandbox stop. however, this caused crio to fail to restore the pod (as it was looking for the managed ns which didn't exist). instead, only remove the namespaces on sandbox remove

```release-note
CRI-O now manages namespace lifecycles by default (manage_ns_lifecycle = true). The config option manage_network_ns_lifecycle has also been fully deprecated
```


docs/crio.conf.5.md
```diff
@@ -215,10 +215,7 @@ the container runtime configuration.
 **ctr_stop_timeout**=30
   The minimal amount of time in seconds to wait before issuing a timeout regarding the proper termination of the container.
 
-**manage_network_ns_lifecycle**=false
-  **DEPRECATED** this option is being replaced by manage_ns_lifecycle, which is described below.
-
-**manage_ns_lifecycle**=false
+**manage_ns_lifecycle**=true
   Determines whether we pin and remove namespaces and manage their lifecycle
```
TomSweeneyRedHat, line 219: I know it's pre-existing, but missing ending period.

haircommander: I was the one that wrote the original anyway :sunglasses:

harche:
lgtm

PR #3933, opened 2020-07-07
Improves container memory usage stats

### LINKED ISSUE #3580: crictl stats return pod memory usage not per container
CRI-O: 1.17.2
K8s: 1.16.8
Cgroup driver: systemd

We found what `kubectl top pods --containers` return same cpu/memory usage for all containers in POD, after i check that by running crictl stats for 2 containers in same pod

```
root@K8S-N1:~# crictl ps | grep f432ef16bf3ca
538fbb2396de1       013718a8b3d932355c501429b8e58d0e27597070371ca5a0308ca1af24af0f01                                                         About an hour ago   Running             nginx                      0                   f432ef16bf3ca
168141c76d71c       9f1e244751343bbb999ac715eea9c7cd946f532b19e45097c4a86951f4757cff                                                         About an hour ago   Running             kube-proxy                 0                   f432ef16bf3ca

root@K8S-N1:~# crictl stats 538fbb2396de1
CONTAINER           CPU %               MEM                 DISK                INODES
538fbb2396de1       5.49                30.65MB             8.194kB             3

root@K8S-N1:~# crictl stats 168141c76d71c
CONTAINER           CPU %               MEM                 DISK                INODES
168141c76d71c       0.15                30.58MB             32.77kB             10

root@K8S-N1:~# crictl stats --pod f432ef16bf3ca
CONTAINER           CPU %               MEM                 DISK                INODES
```

So that's looks like CRI-O internal error with collecting/store metrics

Thanks

wgahnagl:
#### What type of PR is this?

/kind bug

Allows viewing of stats by individual container instead of the whole pod

Fixes #3580 

How the stats display for containers

```release-note
fixed a bug where crictl only showed pod level stats, not container level stats.

```

### DIFF
--- a/internal/config/cgmgr/cgroupfs.go
+++ b/internal/config/cgmgr/cgroupfs.go
@@ -41,6 +41,12 @@ func (*CgroupfsManager) ContainerCgroupPath(sbParent, containerID string) string
 	return filepath.Join(parent, crioPrefix+"-"+containerID)
 }
 
+// ContainerCgroupAbsolutePath just calls ContainerCgroupPath,
+// because they both return the absolute path
+func (m *CgroupfsManager) ContainerCgroupAbsolutePath(sbParent, containerID string) (string, error) {
+	return m.ContainerCgroupPath(sbParent, containerID), nil
+}
+
 // SandboxCgroupPath takes the sandbox parent, and sandbox ID. It
 // returns the cgroup parent, cgroup path, and error.
 func (*CgroupfsManager) SandboxCgroupPath(sbParent, sbID string) (cgParent, cgPath string, _ error) {
--- a/internal/config/cgmgr/systemd.go
+++ b/internal/config/cgmgr/systemd.go
@@ -46,6 +46,23 @@ func (*SystemdManager) ContainerCgroupPath(sbParent, containerID string) string 
 	return parent + ":" + crioPrefix + ":" + containerID
 }
 
+// ContainerCgroupAbsolutePath takes arguments sandbox parent cgroup and container ID and
+// returns the cgroup path on disk for that containerID. If parentCgroup is empty, it
+// uses the default parent system.slice
+func (*SystemdManager) ContainerCgroupAbsolutePath(sbParent, containerID string) (string, error) {
+	parent := defaultSystemdParent
+	if sbParent != "" {
+		parent = sbParent
+	}
+	logrus.Debugf("Expanding systemd cgroup slice %v", parent)
+	cgroup, err := systemd.ExpandSlice(parent)
+	if err != nil {
+		return "", errors.Wrapf(err, "error expanding systemd slice to get container %s stats", containerID)
+	}
+
+	return filepath.Join(cgroup, crioPrefix+"-"+containerID+".scope"), nil
+}
+
 // MoveConmonToCgroup takes the container ID, cgroup parent, conmon's cgroup (from the config) and conmon's PID
 // It attempts to move conmon to the correct cgroup.
 // cgroupPathToClean should always be returned empty. It is part of the interface to return the cgroup path
--- a/test/stats.bats
+++ b/test/stats.bats
@@ -24,18 +24,63 @@ function teardown() {
 
     # then
     JSON="$output"
-    echo $JSON | jq -e '.stats[0].attributes.id != ""'
+    run echo $JSON | jq -e '.stats[0].attributes.id != ""'
     [ "$status" -eq 0 ]
 
-    echo $JSON | jq -e '.stats[0].cpu.timestamp > 0'
+    run echo $JSON | jq -e '.stats[0].cpu.timestamp > 0'
     [ "$status" -eq 0 ]
 
-    echo $JSON | jq -e '.stats[0].cpu.usageCoreNanoSeconds.value > 0'
+    run echo $JSON | jq -e '.stats[0].cpu.usageCoreNanoSeconds.value > 0'
     [ "$status" -eq 0 ]
 
-    echo $JSON | jq -e '.stats[0].memory.timestamp > 0'
+    run echo $JSON | jq -e '.stats[0].memory.timestamp > 0'
     [ "$status" -eq 0 ]
 
-    echo $JSON | jq -e '.stats[0].memory.workingSetBytes.value > 0'
+    run echo $JSON | jq -e '.stats[0].memory.workingSetBytes.value > 0'
+    [ "$status" -eq 0 ]
+}
+
+@test "container stats" {
+    # given
+    container2config=$(cat "$TESTDATA"/container_redis.json | python -c 'import json,sys;obj=json.load(sys.stdin);obj["name"] = ["podsandbox1-redis2"];obj["metadata"]["name"] = "podsandbox1-redis2"; json.dump(obj, sys.stdout)')
+    echo "$container2config" > "$TESTDIR"/container_redis2.json
+    run crictl runp "$TESTDATA"/sandbox_config.json
+    echo "$output"
+    [ "$status" -eq 0 ]
+    pod_id="$output"
+    run crictl create "$pod_id" "$TESTDATA"/container_redis.json "$TESTDATA"/sandbox_config.json
+    echo "$output"
+    [ "$status" -eq 0 ]
+    ctr1_id="$output"
+    run crictl create "$pod_id" "$TESTDIR"/container_redis2.json "$TESTDATA"/sandbox_config.json
+    echo "$output"
+    [ "$status" -eq 0 ]
+    ctr2_id="$output"
+    run crictl start "$ctr1_id"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    run crictl start "$ctr2_id"
+    echo "$output"
+    [ "$status" -eq 0 ]
+
+    # when
+    run crictl stats -o json "$ctr1_id"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    ctr1_stats_JSON="$output"
+
+    run crictl stats -o json "$crt2_id"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    ctr2_stats_JSON="$output"
+
+    run echo $ctr1_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
+    ctr1_memory_bytes="$output"
+    run echo $ctr2_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
+    ctr2_memory_bytes="$output"
+
+    run echo $ctr1_memory_bytes != $ctr2_memory_bytes
     [ "$status" -eq 0 ]
 }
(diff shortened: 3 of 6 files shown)


internal/config/cgmgr/systemd.go
```diff
@@ -46,6 +46,24 @@ func (*SystemdManager) ContainerCgroupPath(sbParent, containerID string) string
...
+	if sbParent != "" {
+		parent = sbParent
+	}
+	logrus.Debugf("Expanding systemd cgroup slice %v", parent)
+	cgroup, err := systemd.ExpandSlice(parent)
+	if err != nil {
+		return "", errors.Wrapf(err, "error expanding systemd slice to get container %s stats", containerID)
+	}
+
+	cgroup = filepath.Join(cgroup, crioPrefix+"-"+containerID) + ".scope"
```
haircommander, line 63: nit, we're already adding the strings together inside of the call, I think we should do the same with scope
`crioPrefix+"-"+containerID+".scope"`

internal/config/cgmgr/systemd.go
```diff
@@ -46,6 +46,24 @@ func (*SystemdManager) ContainerCgroupPath(sbParent, containerID string) string
 	return parent + ":" + crioPrefix + ":" + containerID
 }
 
+// ContainerCgroupAbsolutePath takes arguments sandbox parent cgroup and container ID and
+// returns the cgroup path on disk for that containerID. If parentCgroup is empty, it
+// uses the default parent system.slice
+func (r *SystemdManager) ContainerCgroupAbsolutePath(sbParent, containerID string) (string, error) {
```
haircommander, line 52: hm, why `r` here?

wgahnagl: the linter wanted it
it used to be an m, but the linter was like "why is this an m, everything else is an r"

internal/config/cgmgr/systemd.go
wgahnagl: fixed!

internal/config/cgmgr/systemd.go
haircommander: I think we should change all cases to `m`, `r` has less meaning here IMO

haircommander:
you've got gitvalidation errors too https://app.circleci.com/pipelines/github/cri-o/cri-o/4425/workflows/2419657b-4a78-46ab-920e-56995a81fcd8/jobs/68082/steps
`has whitespace errors. See `git show --check 66eb76d356824880c2d86b8f2a50fd6dc9fc494f`

internal/config/cgmgr/systemd.go
wgahnagl: can do lol

internal/config/cgmgr/systemd.go
```diff
@@ -46,6 +46,24 @@ func (*SystemdManager) ContainerCgroupPath(sbParent, containerID string) string
...
+	if sbParent != "" {
+		parent = sbParent
+	}
+	logrus.Debugf("Expanding systemd cgroup slice %v", parent)
+	cgroup, err := systemd.ExpandSlice(parent)
+	if err != nil {
+		return "", errors.Wrapf(err, "error expanding systemd slice to get container %s stats", containerID)
+	}
+
+	cgroup = filepath.Join(cgroup, crioPrefix+"-"+containerID+".scope")
```
haircommander, line 63: nit: i think we should `return filepath.Join ..., nil` instead of reassigning the variable

internal/config/cgmgr/cgroupfs.go
```diff
@@ -41,6 +41,12 @@ func (*CgroupfsManager) ContainerCgroupPath(sbParent, containerID string) string
 	return filepath.Join(parent, crioPrefix+"-"+containerID)
 }
 
+// ContainerCgroupAbsolutePath just calls ContainerCgroupPath,
+// because they both return the absolute path
+func (r *CgroupfsManager) ContainerCgroupAbsolutePath(sbParent, containerID string) (string, error) {
```
haircommander, line 46: nit, an r snuck in here too

internal/config/cgmgr/systemd.go
wgahnagl: fixed!

internal/config/cgmgr/cgroupfs.go
wgahnagl: fixed!

haircommander:
thanks @wgahnagl LGTM
/approve

haircommander:
a rebase will make the tests happy again

haircommander:
```
# time="2020-07-14 15:45:41.514025886Z" level=error msg="Error checking loopback interface: failed to Statfs \"\": no such file or directory" file="ocicni/ocicni.go:739"
```
[ref](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3933/test_pull_request_crio_integration_rhel/13746/)

/retest

internal/config/cgmgr/systemd.go
```diff
@@ -46,6 +46,23 @@ func (*SystemdManager) ContainerCgroupPath(sbParent, containerID string) string
...
 
+// ContainerCgroupAbsolutePath takes arguments sandbox parent cgroup and container ID and
+// returns the cgroup path on disk for that containerID. If parentCgroup is empty, it
+// uses the default parent system.slice
+func (*SystemdManager) ContainerCgroupAbsolutePath(sbParent, containerID string) (string, error) {
+	parent := defaultSystemdParent
+	if sbParent != "" {
+		parent = sbParent
+	}
+	logrus.Debugf("Expanding systemd cgroup slice %v", parent)
```
kolyshkin, line 57: I think this debug can be dropped

test/stats.bats
```diff
@@ -39,3 +39,48 @@ function teardown() {
...
+    [ "$status" -eq 0 ]
+    ctr1_stats_JSON="$output"
+
+    run crictl stats -o json "$crt2_id"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    ctr2_stats_JSON="$output"
+
+    echo $ctr1_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
```
kolyshkin, line 78: Not sure what you check here -- `$status` is set by bats' `run` function.

test/stats.bats
```diff
@@ -39,3 +39,48 @@ function teardown() {
...
+    ctr1_memory_bytes="$output"
+    echo $ctr2_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
+    ctr2_memory_bytes="$output"
```
kolyshkin, line 82: Similar thing here -- `$output` is only set by bats' `run`.

test/stats.bats
```diff
@@ -39,3 +39,48 @@ function teardown() {
...
+    ctr1_stats_JSON="$output"
+
+    run crictl stats -o json "$crt2_id"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    ctr2_stats_JSON="$output"
+
+    echo $ctr1_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
+    ctr1_memory_bytes="$output"
```
kolyshkin, line 79: Similar thing here -- `$output` is only set by bats' `run`.

test/stats.bats
```diff
@@ -39,3 +39,48 @@ function teardown() {
...
+    echo $ctr2_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
+    ctr2_memory_bytes="$output"
+
+    echo $ctr1_memory_bytes != $ctr2_memory_bytes
+    [ "$status" -eq 0 ]
```
kolyshkin, line 85: ditto

test/stats.bats
haircommander: if the json didn't have the value field then the jq command would fail

kolyshkin: I understand the `jq` part, but the variable being checked, `$status`, was set last time the `run ...` was executed, and it was already checked earlier. In other words, the value of `$status` has nothing to do with `echo .. | jq` invocation in the previous line.

Also, bats sets `-e -o pipefail` options for bash (or so I hope), so if `jq` fails the test fails, there is no need to specifically check its exit code.

So, this code has (1) no effect and (2) is misleading as the reader thinks we're checking `echo | jq` exit code here, but it's not true.

haircommander: oh good catch, I didn't notice there was no run

kolyshkin: It's a little trickier with the `bats` -- you have to remember this is not just `bash`.

haircommander:
/hold

@wgahnagl can you address @kolyshkin comments above?

internal/config/cgmgr/systemd.go
wgahnagl: fixed!

test/stats.bats
wgahnagl: fixed!

test/stats.bats
wgahnagl: fixed!

test/stats.bats
wgahnagl: fixed!

test/stats.bats
wgahnagl: fixed!

test/stats.bats
```diff
@@ -39,3 +39,48 @@ function teardown() {
...
+    run crictl stats -o json "$crt2_id"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    ctr2_stats_JSON="$output"
+
+    echo $ctr1_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
+    ctr1_memory_bytes="$output"
+    echo $ctr2_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
+    [ "$status" -eq 0 ]
```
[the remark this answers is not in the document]
wgahnagl: fixed!

saschagrunert:
One more
/retest

haircommander:
c'mon prow

internal/config/cgmgr/systemd.go
kolyshkin: looks like it was not :( @wgahnagl

test/stats.bats
kolyshkin: So, afaik bats set `set -e` and `set -o pipefail`, meaning that if any command in a pipeline `echo ... | jq ... ` fails, the test case fails.

In other words, the following two code blocks are roughly equivalent:

```bash
echo $ctr1_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
```

```bash
run echo $ctr1_stats_JSON | jq -e '.stats[0].memory.workingSetBytes.value'
[ "$status" -eq 0 ]
```

and unless you want to do something about command exit code (say check if it's 32) and/or output, there is no need to use run. Sorry I should have been clear in my earlier comment -- what I hinted at is that `[ "$status" -eq 0 ]` should be removed (rather than `run ` prefix added).

Can surely be addressed as a followup

kolyshkin: Here's an example:

```console
[kir@kir-rhat runc]$ cat a.bats 
@test "use run" {
	run ls -l | false
	[ "$status" -eq 0 ]
}

@test "don't use run" {
	ls -l | false
}
[kir@kir-rhat runc]$ bats a.bats 
 ✗ use run
   (in test file a.bats, line 2)
     `run ls -l | false' failed
 ✗ don't use run
   (in test file a.bats, line 7)
     `ls -l | false' failed

2 tests, 2 failures
```

PR #3586, opened 2020-04-15
Use named error return for container_server

tedyu:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
This PR names error return variable to cover all possible error return code paths.

#### Which issue(s) this PR fixes:

None

```release-note
None
```

PR #3934, opened 2020-07-07
retErr all the things

haircommander:
#### What type of PR is this?

/kind bug

relates to https://github.com/cri-o/cri-o/pull/3596 and https://github.com/cri-o/cri-o/pull/3586

essentially, it is error prone and risky to name a return value `err`. Instead, let's name every occurrence of a named error `retErr`. Also, drop named return variables in favor of anonymous ones when they don't meet the following two conditions:
- the error is among the list of a list of variables whose purpose could use clarification that variable names can provide (parentPipe and childPipe is one such example)
- the error is checked in a defer

```release-note
fixed bug where pod names would sometimes leak on creation, causing the kubelet to fail to recreate
```

### DIFF
--- a/internal/findprocess/findprocess_unix.go
+++ b/internal/findprocess/findprocess_unix.go
@@ -7,8 +7,8 @@ import (
 	"syscall"
 )
 
-func findProcess(pid int) (process *os.Process, err error) {
-	process, err = os.FindProcess(pid)
+func findProcess(pid int) (*os.Process, error) {
+	process, err := os.FindProcess(pid)
 	if err != nil {
 		return process, err
 	}
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -241,7 +241,8 @@ func (r *runtimeOCI) StartContainer(c *Container) error {
 	return nil
 }
 
-func prepareExec() (pidFile, parentPipe, childPipe *os.File, err error) {
+func prepareExec() (pidFile, parentPipe, childPipe *os.File, _ error) {
+	var err error
 	parentPipe, childPipe, err = os.Pipe()
 	if err != nil {
 		return nil, nil, nil, err
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -129,7 +129,7 @@ func getMountInfo(mountInfos []*mount.Info, dir string) *mount.Info {
 	return nil
 }
 
-func getSourceMount(source string, mountInfos []*mount.Info) (path, optionalMountInfo string, errRet error) {
+func getSourceMount(source string, mountInfos []*mount.Info) (path, optionalMountInfo string, _ error) {
 	mountinfo := getMountInfo(mountInfos, source)
 	if mountinfo != nil {
 		return source, mountinfo.Optional, nil
--- a/server/container_stats.go
+++ b/server/container_stats.go
@@ -48,7 +48,7 @@ func (s *Server) buildContainerStats(ctx context.Context, stats *oci.ContainerSt
 
 // ContainerStats returns stats of the container. If the container does not
 // exist, the call returns an error.
-func (s *Server) ContainerStats(ctx context.Context, req *pb.ContainerStatsRequest) (resp *pb.ContainerStatsResponse, err error) {
+func (s *Server) ContainerStats(ctx context.Context, req *pb.ContainerStatsRequest) (*pb.ContainerStatsResponse, error) {
 	container, err := s.GetContainerFromShortID(req.ContainerId)
 	if err != nil {
 		return nil, err
--- a/server/image_fs_info.go
+++ b/server/image_fs_info.go
@@ -31,7 +31,7 @@ func getStorageFsInfo(store storage.Store) (*pb.FilesystemUsage, error) {
 }
 
 // ImageFsInfo returns information of the filesystem that is used to store images.
-func (s *Server) ImageFsInfo(ctx context.Context, req *pb.ImageFsInfoRequest) (resp *pb.ImageFsInfoResponse, err error) {
+func (s *Server) ImageFsInfo(ctx context.Context, req *pb.ImageFsInfoRequest) (*pb.ImageFsInfoResponse, error) {
 	store := s.StorageImageServer().GetStore()
 	fsUsage, err := getStorageFsInfo(store)
 
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -21,8 +21,9 @@ import (
 var localRegistryPrefix = libpodImage.DefaultLocalRegistry + "/"
 
 // PullImage pulls a image with authentication config.
-func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp *pb.PullImageResponse, err error) {
+func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (*pb.PullImageResponse, error) {
 	// TODO: what else do we need here? (Signatures when the story isn't just pulling from docker://)
+	var err error
 	image := ""
 	img := req.GetImage()
 	if img != nil {
@@ -86,10 +87,9 @@ func (s *Server) PullImage(ctx context.Context, req *pb.PullImageRequest) (resp 
 	}
 
 	log.Infof(ctx, "Pulled image: %v", pullOp.imageRef)
-	resp = &pb.PullImageResponse{
+	return &pb.PullImageResponse{
 		ImageRef: pullOp.imageRef,
-	}
-	return resp, nil
+	}, nil
 }
 
 // pullImage performs the actual pull operation of PullImage. Used to separate
@@ -272,7 +272,7 @@ func tryIncrementImagePullFailureMetric(ctx context.Context, img string, err err
 	}
 }
 
-func decodeDockerAuth(s string) (user, password string, err error) {
+func decodeDockerAuth(s string) (user, password string, _ error) {
 	decoded, err := base64.StdEncoding.DecodeString(s)
 	if err != nil {
 		return "", "", err
--- a/server/sandbox_network.go
+++ b/server/sandbox_network.go
@@ -101,7 +101,7 @@ func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs 
 }
 
 // getSandboxIP retrieves the IP address for the sandbox
-func (s *Server) getSandboxIPs(sb *sandbox.Sandbox) (podIPs []string, err error) {
+func (s *Server) getSandboxIPs(sb *sandbox.Sandbox) ([]string, error) {
 	if sb.HostNetwork() {
 		return nil, nil
 	}
@@ -120,6 +120,7 @@ func (s *Server) getSandboxIPs(sb *sandbox.Sandbox) (podIPs []string, err error)
 		return nil, fmt.Errorf("failed to get network JSON for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
 	}
 
+	podIPs := make([]string, 0, len(res.IPs))
 	for _, podIPConfig := range res.IPs {
 		podIPs = append(podIPs, strings.Split(podIPConfig.Address.String(), "/")[0])
 	}
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -680,7 +680,7 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 	return resp, nil
 }
 
-func setupShm(podSandboxRunDir, mountLabel string) (shmPath string, err error) {
+func setupShm(podSandboxRunDir, mountLabel string) (shmPath string, _ error) {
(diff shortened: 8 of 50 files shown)


haircommander:
now, running:
```
 $ grep -rl 'err error) {' --exclude-dir vendor/ | grep -v golangci 
```
only returns cases where an argument err is passed into a function:
```bash
server/image_pull.go
internal/version/version_test.go
internal/oci/container.go
cmd/crio/main.go
```

haircommander:
```
fatal: [localhost]: FAILED! => {"attempts": 100, "changed": true, "cmd": "/go/src/k8s.io/kubernetes/_output/bin/kubectl get nodes -o jsonpath='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}' 2>&1 | grep 'Ready=True'", "delta": "0:00:00.093969", "end": "2020-07-11 02:24:33.851969", "msg": "non-zero return code", "rc": 1, "start": "2020-07-11 02:24:33.758000", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```
Cluster seems to not come up on cgrouv2...
https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3934/test_pull_request_crio_e2e_crun_fedora_cgroupv2/2787/

internal/oci/runtime_oci.go
```diff
@@ -241,7 +241,8 @@ func (r *runtimeOCI) StartContainer(c *Container) error {
 	return nil
 }
 
-func prepareExec() (pidFile, parentPipe, childPipe *os.File, err error) {
+func prepareExec() (pidFile, parentPipe, childPipe *os.File, retErr error) {
```
TomSweeneyRedHat, line 244: do you need to declare 'retErr` here?

server/image_pull.go
```diff
@@ -272,7 +272,7 @@ func tryIncrementImagePullFailureMetric(ctx context.Context, img string, err err
 	}
 }
 
-func decodeDockerAuth(s string) (user, password string, err error) {
+func decodeDockerAuth(s string) (user, password string, errRet error) {
```
TomSweeneyRedHat, line 275: I don't think errRet needs to be defined here.

server/secrets.go
```diff
@@ -83,7 +83,7 @@ func readFile(root, name string) ([]SecretData, error) {
 }
 
 // getMountsMap separates the host:container paths
-func getMountsMap(path string) (host, container string, err error) {
+func getMountsMap(path string) (host, container string, retErr error) {
```
TomSweeneyRedHat, line 86: I think you can drop retErr here too.

server/image_pull.go
haircommander: in this case, it seems the return value is being used as documentation: which string is which. that's one of the cases I've decided to leave

server/secrets.go
haircommander: same as above

server/container_create.go
```diff
@@ -129,7 +232,7 @@ func getMountInfo(mountInfos []*mount.Info, dir string) *mount.Info {
 	return nil
 }
 
-func getSourceMount(source string, mountInfos []*mount.Info) (path, optionalMountInfo string, err error) {
+func getSourceMount(source string, mountInfos []*mount.Info) (path, optionalMountInfo string, errRet error) {
```
kolyshkin, line 235: in this case I would drop the names from the return list...

kolyshkin: and use
```
path := ...
mountinfo := ...
...
return path, mountinfo.Optional, nil
```

below

server/sandbox_network.go
```diff
@@ -30,13 +30,13 @@ func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs
...
 	if err != nil {
-		return
+		return nil, nil, err
 	}
 
 	// Ensure network resources are cleaned up if the plugin succeeded
 	// but an error happened between plugin success and the end of networkStart()
 	defer func() {
-		if err != nil {
+		if errRet != nil {
```
kolyshkin, line 39: Looks like this code won't compile, as  you forgot to change the function signature above, it still looks this way:

```go
func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs []string, result cnitypes.Result, err error) {
```

server/sandbox_run_linux.go
```diff
@@ -737,20 +737,19 @@ func (s *Server) configureGeneratorForSysctls(ctx context.Context, g generate.Ge
 // as well as whether CRI-O should be managing the namespace lifecycle.
 // it returns a slice of cleanup funcs, all of which are the respective NamespaceRemove() for the sandbox.
 // The caller should defer the cleanup funcs if there is an error, to make sure each namespace we are managing is properly cleaned up.
-func (s *Server) configureGeneratorForSandboxNamespaces(hostNetwork, hostIPC, hostPID bool, sb *libsandbox.Sandbox, g generate.Generator) (cleanupFuncs []func() error, err error) {
+func (s *Server) configureGeneratorForSandboxNamespaces(hostNetwork, hostIPC, hostPID bool, sb *libsandbox.Sandbox, g generate.Generator) (cleanupFuncs []func() error, retErr error) {
+	var err error
```
kolyshkin, line 741: I think it's better to drop this ...

server/sandbox_run_linux.go
```diff
@@ -737,20 +737,19 @@ func (s *Server) configureGeneratorForSysctls(ctx context.Context, g generate.Ge
...
 	managedNamespaces := make([]libsandbox.NSType, 0, 3)
 	if hostNetwork {
-		err = g.RemoveLinuxNamespace(string(spec.NetworkNamespace))
-		if err != nil {
+		if err = g.RemoveLinuxNamespace(string(spec.NetworkNamespace)); err != nil {
```
kolyshkin, line 744: ... and use `err := ....` here

server/sandbox_run_linux.go
```diff
@@ -737,20 +737,19 @@ func (s *Server) configureGeneratorForSysctls(ctx context.Context, g generate.Ge
...
 			return
 		}
 	} else if s.config.ManageNSLifecycle {
 		managedNamespaces = append(managedNamespaces, libsandbox.NETNS)
 	}
 
 	if hostIPC {
-		err = g.RemoveLinuxNamespace(string(spec.IPCNamespace))
-		if err != nil {
+		if err = g.RemoveLinuxNamespace(string(spec.IPCNamespace)); err != nil {
```
kolyshkin, line 752: and here

server/sandbox_run_linux.go
```diff
@@ -759,8 +758,7 @@ func (s *Server) configureGeneratorForSandboxNamespaces(hostNetwork, hostIPC, ho
 
 	// Since we need a process to hold open the PID namespace, CRI-O can't manage the NS lifecycle
 	if hostPID {
-		err = g.RemoveLinuxNamespace(string(spec.PIDNamespace))
-		if err != nil {
+		if err = g.RemoveLinuxNamespace(string(spec.PIDNamespace)); err != nil {
```
kolyshkin, line 761: and here

server/container_create.go
```diff
@@ -491,111 +594,6 @@ func addSecretsBindMounts(ctx context.Context, mountLabel, ctrRunDir string, def
 	return mounts, nil
 }
 
-// CreateContainer creates a new container in specified PodSandbox
```
kolyshkin, line 494: I am very opposed to move this around -- this pollutes the git history, making it harder to read it. The justification for the move ("looks better when it's defined before being used") is not strong enough.

utils/filesystem.go
```diff
@@ -8,8 +8,8 @@ import (
 
 // GetDiskUsageStats accepts a path to a directory or file
 // and returns the number of bytes and inodes used by the path
-func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, err error) {
-	err = filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, errRet error) {
```
umohnani8, line 11: Want to keep this consistent and use `retErr` instead?

kolyshkin:
LGTM in general

Two minor comments:
 - can you please format the changelog messages to not be wider than  80 columns? It's sort of harder to read very long lines
 - I don't like moving CreateContainer, would be great if you can revert it.

utils/utils.go
```diff
@@ -201,7 +205,7 @@ func openContainerFile(rootfs, path string) (io.ReadCloser, error) {
 
 // GetUserInfo returns UID, GID and additional groups for specified user
 // by looking them up in /etc/passwd and /etc/group
-func GetUserInfo(rootfs, userName string) (uid, gid uint32, additionalGids []uint32, err error) {
+func GetUserInfo(rootfs, userName string) (uid, gid uint32, additionalGids []uint32, errRet error) {
```
umohnani8, line 208: nit: same as above, use `retErr` instead.

umohnani8:
2 nits, but LGTM

server/container_stats.go
```diff
@@ -11,6 +11,27 @@ import (
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
+// ContainerStats returns stats of the container. If the container does not
```
kolyshkin, line 14: Same here -- I would not move the code around for the sole reason to be declared above its use.

server/image_fs_info.go
```diff
@@ -10,6 +10,20 @@ import (
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
+// ImageFsInfo returns information of the filesystem that is used to store images.
```
kolyshkin, line 13: ditto

server/container_create.go
haircommander: it isn't really "looks better", I personally find it incredibly difficult to read when functions do not read in order of their calling. the entire server/ package is really hard to read. Though, I understand your point on the git history pollution. I can drop this for now, but do intend on finding a way to move around every function that is defined before its called, hopefully we can find a way that satisfies both of our expectations

kolyshkin:
Some more general comments.

1. Named return values have a few side benefits that you might have overlooked.

1.1. They are a self-documenting code (for example, `func foo() (parent, child string)` instead of `func foo() (string, string)` might be crucial in understanding what are the two strings being returned are).
1.2. They help the compiler to avoid copying stuff around (see e.g. https://blog.minio.io/golang-internals-part-2-nice-benefits-of-named-return-values-1e95305c8687).

2. In case you don't use named value anywhere in the function, and yet you have to name it, because there are other named return values, use `_`. For example, `func foo() (parent, child string, _ err)` rather than `func foo() (parent, child string, retErr error)`.

Based on the above, can you please re-review your own code @haircommander ?

kolyshkin:
> I don't like moving CreateContainer, would be great if you can revert it.

Same for other functions that are move around.

haircommander:
```
[k8s.io] [sig-node] Pods Extended [k8s.io] Pod Container Status should never report success for a pending container
```
[ref](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/3934/test_pull_request_crio_e2e_crun_fedora_cgroupv2/2856/)

e2e_rhel was a flake (tests passed, but reported as failed?)
/retest

haircommander:
PTAL @umohnani8 @mrunalp @kolyshkin @giuseppe @saschagrunert

server/container_create.go
kolyshkin: I agree that ideally and usually a function should be placed right above its use, and it should be done in any new code whenever possible, but I am opposed to moving the code around for the sole reason of placing it before it's used. I find myself frequently going through git history to find out when a particular line of code was introduced or changed, and there are not many tools to make it easier than repeating `git log file` and `git blame <revision> -- file` ad infinitum.

Not sure what kind of editor/IDE are you using, but I find vim + [vim-lsp](https://github.com/prabirshrestha/vim-lsp) + gopls incredibly helpful when reading the code. Jumping back and forth from function use to its implementation, or listing the function callers, etc, is a keystroke away, and it's relatively fast with gopls even with bigger codebases. I'd like to add that my past experience with vim-go and go-guru is not so good.

Even without the advanced tools referenced above, something as simple as `grep`, `git grep`, or a search function built into an editor is enough to find an identifier.

internal/findprocess/findprocess_unix.go
```diff
@@ -7,8 +7,8 @@ import (
 	"syscall"
 )
 
-func findProcess(pid int) (process *os.Process, err error) {
-	process, err = os.FindProcess(pid)
+func findProcess(pid int) (*os.Process, error) {
+	process, err := os.FindProcess(pid)
```
kolyshkin, line 11: Not directly related to this PR, but err handling can be dropped for unix, i.e.
```go
// os.FindProcess never returns an error on unix
process, _ := os.FindProcess(pid)
```
should suffice

PR #3936, opened 2020-07-08
Provide functionality to disable and enable back the CPU load balancing

cynepco3hahue:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

This PR should give to a user option disable the CPU load balancing for the container via specifying relevant annotation.
For Telco workloads, it very important to avoid context switch also among CPUs from the same NUMA node, to make
it possible we need to disable CPU load balancing under the OS via setting scheduling flags for each domain under the 
`/proc/sys/kernel/sched_domain`.

Before the container start the code will check if the `high-performance` runtime used
and  annotation `cpu-load-balancing.crio.io` exists under the container and will
disable CPU load balancing for all CPUs used by this container.

Before the container stop the code will check if the `high-performance` runtime used
and  annotation `cpu-load-balancing.crio.io` exists under the container and will
enable CPU load balancing for back all CPUs used by this container.

```release-note
NONE
```

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>


cynepco3hahue:
@mrunalp Can you please review?

internal/oci/oci_linux.go
```diff
@@ -190,3 +199,68 @@ func getTotalInactiveFile() (uint64, error) {
 
 	return 0, errors.Errorf("%q not found in %v", totalInactiveFilePrefix, memoryStat)
 }
+
+func isCPULoadBalancingShouldBeDisabled(annotations fields.Set) bool {
```
haircommander, line 203: I find this name clunky, can we do `shouldCPULoadBalancingBeDisabled`

haircommander:
thanks @cynepco3hahue couple of nits, in general LGTM

internal/oci/oci_linux.go
```diff
@@ -19,6 +23,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const (
+	annotationCPUloadBalancing = "cpu-load-balancing.kubernetes.io"
```
mrunalp, line 27: We shouldn't use `kubernetes.io` here. Let's use `crio.io` instead.

internal/oci/runtime_oci.go
```diff
@@ -232,6 +232,14 @@ func (r *runtimeOCI) StartContainer(c *Container) error {
 	c.opLock.Lock()
 	defer c.opLock.Unlock()
 
+	// if the annotation cpu-load-balancing.kubernetes.io=true present under the container
```
MarSik, line 235: I wonder if cri-o should be doing some sanity checks too. As this is written if will allow any Pod to disable cpu balancing even if it is not the sole inhabitant of the cpu. That might disrupt other unrelated workloads.

In k8s world this should only be allowed when the Pod belongs to the guaranteed resource class. Meaning it is the only Pod using the requested cpus.

internal/oci/runtime_oci.go
```diff
@@ -622,6 +630,14 @@ func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout in
 		}
 	}
 
+	// if the annotation cpu-load-balancing.kubernetes.io=true present under the container
+	// enable back the CPU load balancing for all CPUs used by the container
+	if isCPULoadBalancingShouldBeDisabled(c.annotations) {
+		if err := setCPUSLoadBalancing(c, true); err != nil {
```
mrunalp, line 636: should this be `false` here?

internal/oci/runtime_oci.go
haircommander: CRI-O is generally pretty dumb, and listens to whatever the kubelet tells it to do. We'd need some admission controller to protect against this

MarSik: I'm just saying this is unsafe by default as it is based on an annotation that anyone can set.

cynepco3hahue:
@haircommander @mrunalp Hi guys, can you give me some advice regarding the testing, from what I can see it problematic to test it under unittest(I can add shcedDomainPath as an argument to `setCPUSLoadBalancing`, but in this case, I will only check the logic under the `setCPUSLoadBalancing` and I want to check the correctness of logic under Start and Stop container methods) and e2e tests run k8s e2e tests(please correct me if I wrong). From your perspective, what will be the best way to test the new functionality?

mrunalp:
We should move this logic up into the CRI calls and tie them to a custom runtime class with a special name. 
We can then have some access control for runtime class up the stack similar to what we need for kata as well.

cynepco3hahue:
@mrunalp Just to be sure that I understand correctly:

- I add condition check to the line https://github.com/cri-o/cri-o/blob/f9b12e50a8268fb24d2c749adade9019c6f8e20e/server/container_start.go#L38, something like that
```
if strings.Contains(strings.ToLower(runtimeHandler), "cpu-load-balancing") {
    // disable the CPU load balancing
}
```
- I will do a similar thing I will do for the stop container method
- I will add the new runtime under the config template

Please correct me if you meant something different.

cynepco3hahue:
> We should move this logic up into the CRI calls and tie them to a custom runtime class with a special name.
> We can then have some access control for runtime class up the stack similar to what we need for kata as well.

@mrunalp Can you have an example of such access control, I failed to find any relevant information regarding it.

mrunalp:
@cynepco3hahue Yes, moving it to the CRI calls like Start/Stop/Remove makes sense :+1: 
Also, we don't have to add this runtime handler by default. It will have to be configured when needed.

mrunalp:
> Can you have an example of such access control, I failed to find any relevant information regarding it.

It isn't done but something that we have to do.

cynepco3hahue:
> It isn't done but something that we have to do.

Give me know if you need help with this effort(do we have an issue or KEP for it?).

cynepco3hahue:
@mrunalp @haircommander Can you please take a look?

server/container_start.go
```diff
@@ -34,8 +36,20 @@ func (s *Server) StartContainer(ctx context.Context, req *pb.StartContainerReque
 		if err := s.ContainerStateToDisk(c); err != nil {
 			log.Warnf(ctx, "unable to write containers %s state to disk: %v", c.ID(), err)
 		}
+
+		if strings.Contains(runtimeHandler, runtimeHandlerHighPerformance) && shouldCPULoadBalancingBeDisabled(c.Annotations()) {
```
haircommander, line 40: nit: since we do this operation twice, can we save whether the runtimeHandler contains runtimeHandlerHighPerformance in a variable above?

cynepco3hahue: sure will do it

cynepco3hahue: done

cynepco3hahue:
@haircommander @mrunalp I provided some basic unittests for `setCPULoadBalancing`, to be honest, I wanted to add some tests under `container_start` and `container_stop` test files, but did not find a good way(without introducing a new field under the container struct that will contain the path to the sched_domain directory).

Please give me know if I missing something under the PR and a lot of thanks for the review.

haircommander:
ach, I'm afraid you have a merge conflict

cynepco3hahue:
@haircommander No worries, will update the PR :)

server/container_start.go
```diff
@@ -34,8 +39,20 @@ func (s *Server) StartContainer(ctx context.Context, req *pb.StartContainerReque
...
+		if useHighPerformanceRuntimeHandler && shouldCPULoadBalancingBeDisabled(c.Annotations()) {
+			if err := setCPUSLoadBalancing(c, true, schedDomainDir); err != nil {
+				log.Warnf(ctx, "failed enable back the container %q CPUs load balancing: %v", c.ID(), err)
+			}
+		}
 	}()
 
+	if useHighPerformanceRuntimeHandler && shouldCPULoadBalancingBeDisabled(c.Annotations()) {
+		if err := setCPUSLoadBalancing(c, true, schedDomainDir); err != nil {
+			return nil, fmt.Errorf("failed to disable the container %q CPUs load balalancing: %v", c.ID(), err)
```
haircommander, line 52: nit 1: the logic here feels backwards. the function is called `setCPUSLoadBalancing` but failure to set to true results in us returning an error that says we `failed to disable the ... load balancing`
nit 2: s/balalancing/balancing/g

cynepco3hahue: done

server/container_start.go
```diff
@@ -34,8 +39,20 @@ func (s *Server) StartContainer(ctx context.Context, req *pb.StartContainerReque
 		if err := s.ContainerStateToDisk(c); err != nil {
 			log.Warnf(ctx, "unable to write containers %s state to disk: %v", c.ID(), err)
 		}
+
+		if useHighPerformanceRuntimeHandler && shouldCPULoadBalancingBeDisabled(c.Annotations()) {
+			if err := setCPUSLoadBalancing(c, true, schedDomainDir); err != nil {
+				log.Warnf(ctx, "failed to set back the container %q CPUs load balancing to true: %v", c.ID(), err)
```
TomSweeneyRedHat, line 45:
```suggestion
				log.Warnf(ctx, "failed to set the container %q CPUs load balancing to true: %v", c.ID(), err)
```

server/container_start.go
```diff
@@ -34,8 +39,20 @@ func (s *Server) StartContainer(ctx context.Context, req *pb.StartContainerReque
...
+			}
+		}
 	}()
 
+	if useHighPerformanceRuntimeHandler && shouldCPULoadBalancingBeDisabled(c.Annotations()) {
+		if err := setCPUSLoadBalancing(c, true, schedDomainDir); err != nil {
```
haircommander, line 51: this is still true but the error says false

cynepco3hahue: thanks, done

haircommander:
thanks @cynepco3hahue this LGTM

/approve

PTAL @mrunalp @saschagrunert @harche @umohnani8 @kolyshkin

cynepco3hahue:
@mrunalp @haircommander I tested my change in the real environment, and I can see that annotation did not pass from the pod to the container spec, I was sure that the k8s passes all pod annotations to the container config :( So, for now, I want to remove the annotation check and to leave only the handler check.
I probably will need to create the PR under the k8s to propagate this annotation, but it no chance it will be ready for 1.19.
WDYT?

mrunalp:
@cynepco3hahue We have an association from containers to pods, so we should be able to check pod annotations for a container by looking at the sandbox object.

cynepco3hahue:
@mrunalp It is great, thanks for the info.

go.mod
```diff
@@ -55,6 +55,7 @@ require (
 	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
 	golang.org/x/sys v0.0.0-20200625212154-ddb9806d33ae
 	google.golang.org/grpc v1.30.0
+	google.golang.org/grpc/examples v0.0.0-20200716233830-6dc7938fe875 // indirect
```
saschagrunert, line 58: Can we remove this?

cynepco3hahue: I unsure why it was added in the first place, my pattern to update dependencies was:
`go clean -modcache && go mod tidy && go mod vendor`, so it was added automatically by `go mod tidy`.

saschagrunert: If you switch to go 1.14, manually remove the line and run `make vendor` then the change should persist. :)

cynepco3hahue: Done, thanks for the tip!

PR #3944, opened 2020-07-10
user namespaces: add support for userns-mode annotation

giuseppe:
#### What type of PR is this?

/kind feature

allow users to specify the user namespace to configure for the pod.
    
A namespace can be configured with an annotation like:

```
    annotations:
            io.kubernetes.cri-o.userns-mode: "private:uidmapping=0:1000:5000;gidmapping=0:1000:4000"
```

the user namespace will be created for the entire pod.  Each container will then join the same user namespace.
    
If any of runAsUser/runAsGroup/fsGroup is not present in the specified namespace, they will be automatically added.
    
The specified runAsUser/runAsGroup/fsGroup configuration will be honored from the host point of view, and they will be remapped to the new value inside of the user namespace.

```release-note
support user namespace creation via an annotation.
```


server/sandbox_run_linux.go
```diff
@@ -37,6 +39,208 @@ import (
...
+		for _, r := range strings.Split(parts[1], ";") {
+			kv := strings.SplitN(r, "=", 2)
+			if len(kv) != 2 {
+				return nil, errors.Errorf("invalid argument: %q", r)
+			}
+			values[kv[0]] = kv[1]
+		}
+	}
+
+	switch parts[0] {
```
rhatdan, line 87: Can we share this parsing between Podman and Cri-o?

giuseppe: I am afraid we will end up with a different set of options for CRI-O.

I am still not sure how `auto` will look like, all the assumptions we had with Podman are not valid anymore.  We cannot really inspect the image since it is not known at user namespace creation time.

rhatdan: Well we can just allocate the 65k and be done with it.  For now, if user does not specify it.

Kubernetes will keep better track of the containers and not leave them forever on disk.

giuseppe: that is a good idea.  I think it should be either specified or default to something reasonable like 65k

internal/storage/runtime.go
```diff
@@ -346,12 +348,12 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
 	}, nil
 }
 
-func (r *runtimeService) CreatePodSandbox(systemContext *types.SystemContext, podName, podID, imageName, imageAuthFile, imageID, containerName, metadataName, uid, namespace string, attempt uint32, idMappings *idtools.IDMappings, labelOptions []string, privileged bool) (ContainerInfo, error) {
-	return r.createContainerOrPodSandbox(systemContext, podName, podID, imageName, imageAuthFile, imageID, containerName, podID, metadataName, uid, namespace, attempt, idMappings, labelOptions, true, privileged)
+func (r *runtimeService) CreatePodSandbox(systemContext *types.SystemContext, podName, podID, imageName, imageAuthFile, imageID, containerName, metadataName, uid, namespace string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, privileged bool) (ContainerInfo, error) {
```
harche, line 351: This is not related to this PR, but the number argument `CreatePodSandbox` takes unusually high. Can we not have some of these embedded inside `*runtimeService` or maybe we can have `CreateSandoxInfo` struct to hold those values and pass it here as an argument. 

@haircommander WDYT?

haircommander: @wgahnagl is in the process of refactoring a lot of the sandbox/container creation code into new interfaces. I think it is a good idea to refactor the arguments here, but I think it should be `pkg/sandbox/sandbox`, not another intermediate struct

giuseppe:
@wgahnagl maybe it is not so bad as I thought :-)

giuseppe:
@rhatdan @mrunalp added the two checks we discussed earlier

internal/lib/sandbox/namespaces.go
```diff
@@ -114,11 +115,13 @@ func (s *Sandbox) CreateNamespacesWithFunc(managedNamespaces []NSType, cfg *conf
 				nsPath: namespace.Path(),
 			})
 		case USERNS:
-			s.userns = namespaceIface
-			typesAndPaths = append(typesAndPaths, &ManagedNamespace{
-				nsType: USERNS,
-				nsPath: namespace.Path(),
-			})
+			if idMappings != nil {
```
haircommander, line 118: we may want to fail here, otherwise I think we leak the userns (which shouldn't have been created in the first place)

giuseppe: the idMappings can be nil, when not using a userns.

Where do we leak the userns?  I'll fix that

haircommander: the way this code reads is:
after we pin the user namespace, verify the idMapping isn't nil. if it is nil, we do not pass the newly pinned user namespace to the caller. If this happens, the caller won't know to clean up the newly created user namespace.

We should fail before the call to `pinns` happens if idMappings is nil but the caller specified USERNS

giuseppe:
@haircommander @mrunalp @rhatdan fine to merge this as it is?  It will avoid to rebase again :) and I will follow up with the comments @haircommander had.  I think it will be easier to iterate

haircommander:
dropping this note publically (I talked with @giuseppe privately) I think the pinns/cri-o interaction is currently awkward. I've asked him to reimplement the IPC as CLI instead of have cri-o take care of setting up the mount after pinns does the work (requiring the stdout reading and process killing)

I am fine with that as a follow on though, along with my other nits

rhatdan:
LGTM
Great work @giuseppe

giuseppe:
just pushed a new version with the changes suggested by @haircommander

giuseppe:
rebased

docs/crio.conf.5.md
```diff
@@ -206,6 +206,10 @@ the container runtime configuration.
 **read_only**=false
   If set to true, all containers will run in read-only mode.
 
+**allow_userns_annotation**=false
+  Allow usage of the experimental feature for setting user namespaces through the io.kubernetes.cri-o.userns-mode annotation.  This is an experimental feature, do not
```
TomSweeneyRedHat, line 210:
```suggestion
  Allow usage of the experimental feature for setting user namespaces through the io.kubernetes.cri-o.userns-mode annotation.  This is an experimental feature, 
```

docs/crio.conf.5.md
```diff
@@ -206,6 +206,10 @@ the container runtime configuration.
...
+  enable in production.  It might be changed in future without any notice.
```
TomSweeneyRedHat, line 211:
```suggestion
  enable this annotation in a production environment.  It may be changed in the future without any notice.
```

giuseppe: thanks, fixed

giuseppe:
rebased again.

@haircommander @mrunalp @rhatdan are you fine with merging it?

giuseppe:
rebased again

docs/crio.conf.5.md
```diff
@@ -206,6 +206,10 @@ the container runtime configuration.
 **read_only**=false
   If set to true, all containers will run in read-only mode.
 
+**allow_userns_annotation**=false
+  Allow usage of the experimental feature for setting user namespaces through the io.kubernetes.cri-o.userns-mode annotation.  This is an experimental feature,
```
mrunalp, line 210: nit: maybe we should do it like the crio.8.md file as [EXPERIMENTAL] so it stands out.

internal/lib/container_server.go
```diff
@@ -31,6 +31,9 @@ import (
 // `io.container.manager`.
 const ContainerManagerCRIO = "cri-o"
 
+// UsernsMode is the user namespace mode to use
+const UsernsModeAnnotation = "io.kubernetes.cri-o.userns-mode"
```
mrunalp, line 35: This should go into `annotations` package. I am still not sure at what point we moved that to libpod since all the annotations are crio specific in that package.

pinns/pinns.c
```diff
@@ -100,44 +122,96 @@ int main(int argc, char **argv) {
...
   }
 
-  if (unshare(unshare_flags) < 0) {
-    pexit("Failed to unshare namespaces");
+  if (bind_user && (uid_mapping == NULL || gid_mapping == NULL))
+    nexit("Creating new user namespace but mappings not specified");
+  if (!bind_user && (uid_mapping != NULL || gid_mapping != NULL))
+    nexit("Mappings specified without creating a new user namespace");
+
+  if (! bind_user) {
```
mrunalp, line 130: nit: space after !

pinns/pinns.c
```diff
@@ -100,44 +122,96 @@ int main(int argc, char **argv) {
...
 
   if (bind_cgroup) {
-    if (bind_ns(pin_path, filename, "cgroup") < 0) {
+    if (bind_ns(pin_path, filename, "cgroup", pid) < 0) {
       return EXIT_FAILURE;
     }
   }
 
+  /* Avoid creating a zombie.  */
+  if (pid > 0 && kill (pid, SIGKILL) == 0)
```
mrunalp, line 208: nit: space after kill

pinns/pinns.c
```diff
@@ -38,6 +52,8 @@ int main(int argc, char **argv) {
       {"cgroup", optional_argument, NULL, 'c'},
       {"dir", required_argument, NULL, 'd'},
       {"filename", required_argument, NULL, 'f'},
+      {"uid-mapping", optional_argument, NULL, UID_MAPPING},
```
mrunalp, line 55: Hmm, reason why we use enum values here? Could it just be a long argument?

pinns/pinns.c
```diff
@@ -168,11 +248,61 @@ static int bind_ns(const char *pin_path, const char *filename, const char *ns_na
...
+    if (*it == '@')
+      *it = '\n';
+    else if (*it == '-')
+      *it = ' ';
+  }
+  content_size = it - content + 1;
+
+  fd = open(path, O_WRONLY | O_CLOEXEC);
+  if (fd < 0)
+    {
```
mrunalp, line 281: nit: style consistency :)

pinns/pinns.c
```diff
@@ -168,11 +248,61 @@ static int bind_ns(const char *pin_path, const char *filename, const char *ns_na
...
+      int saved_errno = errno;
+      free (content);
+      errno = saved_errno;
+      return -1;
+    }
+
+  if (write(fd, content, content_size) != content_size)
+    {
```
mrunalp, line 289: Same here.

server/sandbox_run_linux.go
```diff
@@ -37,6 +39,234 @@ import (
...
+		if s.defaultIDMappings != nil {
+			uids := s.defaultIDMappings.UIDs()
+			gids := s.defaultIDMappings.GIDs()
+			return &storage.IDMappingOptions{UIDMap: uids, GIDMap: gids}, nil
+		}
+		return nil, nil
+	}
+
+	// Ignore the annotation if not explicitly set in the config file.
+	if !s.config.AllowUsernsAnnotation {
```
mrunalp, line 74: Can we move this as the first check for the function or out into the calling code?

server/sandbox_run_linux.go
```diff
@@ -37,6 +39,234 @@ import (
...
+		}
+	}
+
+	switch parts[0] {
+	case "auto":
+		const t = "true"
+		ret := &storage.IDMappingOptions{
+			AutoUserNs: true,
+		}
+		keepID := values["keep-id"] == t
```
mrunalp, line 107: Could we add some comments here on the format for auto and what these values mean?

server/sandbox_run_linux.go
```diff
@@ -37,6 +39,234 @@ import (
...
+			return nil, errors.Errorf("cannot use both keep-id and map-to-root: %q", mode)
+		}
+		if v, ok := values["size"]; ok {
+			s, err := strconv.Atoi(v)
+			if err != nil {
+				return nil, err
+			}
+			ret.AutoUserNsOpts.Size = uint32(s)
+		} else {
+			ret.AutoUserNsOpts.Size = 65536
```
mrunalp, line 119: Can we define a constant for this value?

server/sandbox_run_linux.go
```diff
@@ -37,6 +39,234 @@ import (
...
+			ret.AutoUserNsOpts.AdditionalUIDMappings = append(ret.AutoUserNsOpts.AdditionalUIDMappings, uids...)
+		}
+		if v, ok := values["gidmapping"]; ok {
+			gids, err := idtools.ParseIDMap(strings.Split(v, ","), "GID")
+			if err != nil {
+				return nil, err
+			}
+			ret.AutoUserNsOpts.AdditionalGIDMappings = append(ret.AutoUserNsOpts.AdditionalGIDMappings, gids...)
+		}
+		if sc.GetRunAsUser() != nil {
```
mrunalp, line 135: How are we handling the case where the user name to run as is set in the Dockerfile?

mrunalp: `setupContainerUser` in container_create.go has the code to combine and come up with a final list of uid, gid and groups. I think we should refactor that code to return those values and reuse it here.

internal/lib/container_server.go
mrunalp: We can handle this as follow-up but we should leave a TODO here.

pinns/pinns.c
giuseppe: The reason is that other options seem to follow the `unshare` tool flags.  And these two new options are not available there, so to not risk I've added only the long argument variant

pinns/pinns.c
giuseppe: dropped

internal/lib/container_server.go
giuseppe: added

server/sandbox_run_linux.go
giuseppe: I don't think that can affect how we create the user namespace.  The user namespace is created for the POD so all the uids/gids must be already defined.  If a container is created later with a different uid/gid that is not part of the user namespace, then that cannot be handled.

I think this still requires some discussions and if we ccan come up with something better.

mrunalp: Yeah, just realized that after our conversations today :) :+1:

giuseppe:
rebased again

giuseppe:
found an issue in containers/storage while playing with this: https://github.com/containers/storage/pull/674

I think we can revendor later, once we have a new version for c/storage

giuseppe:
@rhatdan and another https://github.com/containers/storage/pull/675

I am finding these issues now as I am stressing punching holes in the user namespace and creating hundreds of pods

giuseppe:
recorded an asciinema: https://asciinema.org/a/351396

giuseppe:
last missing piece in containers/storage: https://github.com/containers/storage/pull/676

With that in place we can delete mapped images (this can happen at container deletion time if it is using a mapping)

server/sandbox_run_linux.go
```diff
@@ -37,6 +39,236 @@ import (
...
+			kv := strings.SplitN(r, "=", 2)
+			if len(kv) != 2 {
+				return nil, errors.Errorf("invalid argument: %q", r)
+			}
+			values[kv[0]] = kv[1]
+		}
+	}
+
+	_, uidMappingsPresent := values["uidmapping"]
+	_, gidMappingsPresent := values["uidmapping"]
```
mauriciovasquezbernal, line 92:
```suggestion
	_, gidMappingsPresent := values["gidmapping"]
```

server/sandbox_run_linux.go
```diff
@@ -37,6 +39,236 @@ import (
...
+	// allow these options only if running as root
+	if uidMappingsPresent || gidMappingsPresent {
+		if sc.GetRunAsUser().Value != 0 {
```
mauriciovasquezbernal, line 95: Missing `nil` check?

server/container_create_linux.go
```diff
@@ -201,6 +208,64 @@ func makeAccessible(path string, uid, gid int) error {
...
+	return nil
+}
+
+func toContainer(id uint32, idMap []idtools.IDMap) (uint32, error) {
+	hostID := int(id)
+	if idMap == nil {
+		return uint32(hostID), nil
+	}
+	for _, m := range idMap {
+		if (hostID >= m.HostID) && (hostID <= (m.HostID + m.Size - 1)) {
```
mauriciovasquezbernal, line 229:
```suggestion
		if hostID >= m.HostID && hostID < m.HostID + m.Size {
```

?

giuseppe: yes that is definitely better!

server/sandbox_run_linux.go
giuseppe: added

server/sandbox_run_linux.go
giuseppe: thanks for spotting my copy&pasta!

giuseppe:
@mauriciovasquezbernal thanks for the review.  I've amended your suggestions

rhatdan:
@giuseppe In your ascicinema video, you should have exec'd into the container to show that it was running as root.

giuseppe:
I'll need to re-vendor again containers/storage as there is a regression in the last release

giuseppe:
> I'll need to re-vendor again containers/storage as there is a regression in the last release

revendored to 1.23.1

giuseppe:
@mrunalp it is still stuck :/

mrunalp:
tide issue, we can merge :)

PR #3949, opened 2020-07-13
runtime_vm: Avoid premature container info cleanup on creation failure

fidencio:
#### What type of PR is this?

/kind bug

While the removing the container info of the failed container from the
containers' map seems harmless, it'll end up causing an error when
actually cleaning up / deleting the failed container.

The reason for this is that CreateContainer, on Kata side, will remove
the failed container, wich will result on DeleteContainer, on the CRI-O
side, being called and then failing to actually remove the container, as
its info has already been removed from the containers' map.

This issue has been reported as part of a PR opened by Evan Foster:
https://github.com/kata-containers/runtime/pull/2826#issuecomment-657754302

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -113,12 +113,6 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (err erro
 		cio: containerIO,
 	}
 
-	defer func() {
-		if err != nil {
-			delete(r.ctrs, c.ID())
-		}
-	}()
-
 	// We can now create the container, interacting with the server
 	request := &task.CreateTaskRequest{
 		ID:       c.ID(),


fidencio:
@amshinde, please, take a look

haircommander:
/ok-to-test

/approve

LGTM, thanks @fidencio

haircommander:

fidencio:
Just a note, please, do *not* merge this until https://github.com/kata-containers/runtime/pull/2826 is re-worked / merged.
I want to be 100% sure this won't cause any issue due to the changes in the kata PR.

saschagrunert:
> Just a note, please, do **not** merge this until [kata-containers/runtime#2826](https://github.com/kata-containers/runtime/pull/2826) is re-worked / merged.
> I want to be 100% sure this won't cause any issue due to the changes in the kata PR.

Thanks for the hint:
/hold

fidencio:
Actually, I'll withdraw this PR. While taking a look on a different issue, I sincerely think what I'm doing on relying on kata is wrong.
We do should do the cleanup, but the full cleanup instead of just deleting the container ID from the map. A cleaner solution would be a simple call to `r.DeleteContainer()`, but in a way it doesn't cause a deadlock. :-)

PR #3950, opened 2020-07-13
[1.18] change variable name err to retErr for deferred comparisons

haircommander:
if we're branching on whether we errored in execution at the end of a function, we should use the retErr name, as it should not get shadowed

partial port of https://github.com/cri-o/cri-o/pull/3934/commits

#### What type of PR is this?

/kind bug

```release-note
fixed bug where pod names would sometimes leak on creation, causing the kubelet to fail to recreate
```

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -40,7 +40,7 @@ import (
 const cgroupMemorySubsystemMountPathV1 = "/sys/fs/cgroup/memory"
 const cgroupMemorySubsystemMountPathV2 = "/sys/fs/cgroup"
 
-func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest) (resp *pb.RunPodSandboxResponse, err error) {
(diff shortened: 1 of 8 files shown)


haircommander:
```
[sig-storage] EmptyDir volumes volume on default medium should have the correct mode [LinuxOnly] [NodeConformance] [Conformance]
```
haven't seen this one before

/retest

server/sandbox_run_linux.go
```diff
@@ -670,13 +670,13 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 	return resp, nil
 }
 
-func setupShm(podSandboxRunDir, mountLabel string) (shmPath string, err error) {
-	shmPath = filepath.Join(podSandboxRunDir, "shm")
+func setupShm(podSandboxRunDir, mountLabel string) (string, error) {
```
mrunalp, line 673: Here we are losing documentation implicit in the return value names. If we do this, then we should add a comment for the function.

mrunalp: This case is where I would argue that keeping it as is probably fine as we don't have defers and it is a small function.

haircommander: this one actually snuck...

PR #3951, opened 2020-07-14
Move server tests to the server package

cynepco3hahue:
#### What type of PR is this?

> /kind api-change
> /kind bug
/kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

It a good practice to have unit tests under the same package
as a code that should be tested.

It allows to:
- get rid of unneeded imports
- to test private consts and variables under unittests
- to test private methods

```release-note
NONE
```
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/server/container_attach_test.go
+++ b/server/container_attach_test.go
@@ -1,4 +1,4 @@
-package server_test
+package server
 
 import (
 	"context"
--- a/server/container_create_test.go
+++ b/server/container_create_test.go
@@ -1,4 +1,4 @@
-package server_test
+package server
 
 import (
 	"context"
--- a/server/container_exec_test.go
+++ b/server/container_exec_test.go
@@ -1,4 +1,4 @@
-package server_test
+package server
 
 import (
 	"context"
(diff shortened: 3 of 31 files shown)


kolyshkin:
> It a good practice to have unit tests under the same package as a code that should be tested.

Depending on approach, we might want to test the code as a black box -- for such cases we use `packagename_test` for the testing package. I am not sure if this was the intention for `server_test` though.

cynepco3hahue:
> Depending on approach, we might want to test the code as a black box -- for such cases we use `packagename_test` for the testing package. I am not sure if this was the intention for `server_test` though.

I see, please give me know if you do not have the intention to change the server approach.

haircommander:
I think both have a use. Unit testing internally is useful for checking things we cannot test from the outside, but a lot of our tests do assume the blackbox approach. I would say we should leave the existing tests as is (maybe move them to a different package directories if it gets too cluttered?) and then add internal tests as needed

cynepco3hahue:
@haircommander Hi thanks for the answer, it makes sense. I will close a PR and will open a new one, once I will have enough additional internal tests.

PR #3952, opened 2020-07-14
internal/oci/runtime_vm: lock around map access

kolyshkin:
It looks like a concurrent map accces can occur here, since the only
lock taken is per-container. Guard such accesses with a mutex lock.

/kind bug

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -774,7 +774,7 @@ func (r *runtimeVM) remove(ctx context.Context, ctrID, execID string) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


haircommander:
/retest

LGTM, good catch
/approve

kolyshkin:
Wow, adding a lock to the structure revealed a typo :)

Fixed by adding a small commit before mine.

harche:
good catch. lgtm.

fidencio:
lgtm!

saschagrunert:
e2e_crun_cgroupv2 seems broken, may you have a look @giuseppe ?

PR #3956, opened 2020-07-15
runtimeVM: Improve CreateContainer cleanup in case of failures

fidencio:
/kind bug

In case of failures on CreateContainer, let's do a full cleanup by
closing the IO opened, removing the container, shutting it down, and
finally removing it from the containers' map, which is performed by the
recently introduced deleteContainer() function.

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -535,22 +535,30 @@ func (r *runtimeVM) DeleteContainer(c *Container) error {
 	c.opLock.Lock()
 	defer c.opLock.Unlock()
 
+	return r.deleteContainer(c, false)
+}
+
+// deleteContainer performs all the operations needed to delete a container.
+// force must only be used on clean-up cases.
+// It does **not** Lock the container, thus it's the caller responsibility to do so, when needed.
+func (r *runtimeVM) deleteContainer(c *Container, force bool) error {
 	r.Lock()
 	cInfo, ok := r.ctrs[c.ID()]
 	r.Unlock()
-	if !ok {
+	if !ok && !force {
 		return errors.New("Could not retrieve container information")
 	}
 
-	if err := cInfo.cio.Close(); err != nil {
+	if err := cInfo.cio.Close(); err != nil && !force {
 		return err
 	}
 
-	if err := r.remove(r.ctx, c.ID(), ""); err != nil {
+	if err := r.remove(r.ctx, c.ID(), ""); err != nil && !force {
 		return err
 	}
 
-	if _, err := r.task.Shutdown(r.ctx, &task.ShutdownRequest{ID: c.ID()}); err != nil && !errors.Is(err, ttrpc.ErrClosed) {
+	_, err := r.task.Shutdown(r.ctx, &task.ShutdownRequest{ID: c.ID()})
+	if err != nil && !errors.Is(err, ttrpc.ErrClosed) && !force {
 		return err
 	}
 
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -119,9 +119,10 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (err erro
 
 	defer func() {
 		if err != nil {
-			r.Lock()
-			delete(r.ctrs, c.ID())
-			r.Unlock()
+			logrus.WithError(err).Warnf("Cleaning up container %s", c.ID())
+			if cleanupErr := r.deleteContainer(c, true); cleanupErr != nil {
+				logrus.WithError(cleanupErr).Infof("deleteContainer failed for container %s", c.ID())
+			}
 		}
 	}()
 


fidencio:
This PR is rebased atop of https://github.com/cri-o/cri-o/pull/3924 and https://github.com/cri-o/cri-o/pull/3952

fidencio:
I've tested with the patches from https://github.com/kata-containers/runtime/pull/2826/ and it did work for me. But I'd like to have a confirmation from @evanfoster.

internal/oci/runtime_vm.go
```diff
@@ -109,13 +111,32 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (err erro
...
 
+	r.Lock()
 	r.ctrs[c.ID()] = containerInfo{
 		cio: containerIO,
 	}
+	r.Unlock()
 
 	defer func() {
 		if err != nil {
+			logrus.WithError(err).Warnf("Cleaning up container %v", c.ID())
```
haircommander, line 122: is all of the code here in delete? why not add a fatal bool that returns in an error case to a deleteContainer helper function. Or, you can take the step by step approach as others (the moment you provision a resource, defer its removal). or leave it as is :smiley:

haircommander: (the deleteContainer helper could also be non-locking)

internal/oci/runtime_vm.go
```diff
@@ -109,13 +111,32 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (err erro
...
+				logrus.WithError(cleanupErr).Infof("Error closing the container %v IO", c.ID())
+			}
+
+			if cleanupErr := r.remove(r.ctx, c.ID(), ""); cleanupErr != nil {
+				logrus.WithError(cleanupErr).Infof("Error removing the container %v from the background context", c.ID())
+			}
+
+			_, cleanupErr := r.task.Shutdown(r.ctx, &task.ShutdownRequest{ID: c.ID()})
+			if cleanupErr != nil && !errors.Is(cleanupErr, ttrpc.ErrClosed) {
+				logrus.WithError(cleanupErr).Infof("Error error shutting down %v", c.ID())
```
TomSweeneyRedHat, line 134: "Error error"?  I think just "Error" is sufficient.

TomSweeneyRedHat:
```suggestion
				logrus.WithError(cleanupErr).Infof("Error shutting down container %v", c.ID())
```

internal/oci/runtime_vm.go
fidencio: Yep, all of the code here is in DeleteContainer.

The very first version of patch I added a `cleanupContainer()` helper, which would take a `bool` to meaning force, with some lines of documentation saying that locking was up to the caller, but I must admit it could be more confusing than the approach taken in case someone decides to use it later on and, for some reason, doesn't carefully read the comments.

internal/oci/runtime_vm.go
fidencio: That was a typo, thanks! :-)

internal/oci/runtime_vm.go
haircommander: in general I'd prefer to have it in one place. I'd personally like to assume anyone adding a call to a function should have at least read its comment. It add maintenance overhead (and duplicated code) to have this in two places. what if there's a new step to cleaning up a container? more worth it to me to have it in one place

fidencio: Ack!

haircommander:
you've got a lint problem:
```
internal/oci/runtime_vm.go:123:21: Error return value of `r.deleteContainer` is not checked (errcheck)
 
			r.deleteContainer(c, true)
 
```

fidencio:
> you've got a lint problem:
> 
> ```
> internal/oci/runtime_vm.go:123:21: Error return value of `r.deleteContainer` is not checked (errcheck)
>  
> 			r.deleteContainer(c, true)
>  
> ```

Oh, let me fix that!

evanfoster:
I've tested this PR in conjunction with https://github.com/kata-containers/runtime/pull/2826 and I'm not seeing any processes hanging around. Seems like it's working well!

haircommander:
/approve
LGTM
PTAL @saschagrunert @umohnani8 @mrunalp

mrunalp:
Needs rebase.

internal/oci/runtime_vm.go
```diff
@@ -119,9 +119,10 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (err erro
 
 	defer func() {
 		if err != nil {
-			r.Lock()
-			delete(r.ctrs, c.ID())
-			r.Unlock()
+			logrus.WithError(err).Warnf("Cleaning up container %v", c.ID())
```
harche, line 122:
```suggestion
			logrus.WithError(err).Warnf("Cleaning up container %s", c.ID())
```

internal/oci/runtime_vm.go
```diff
@@ -119,9 +119,10 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (err erro
...
+			if cleanupErr := r.deleteContainer(c, true); cleanupErr != nil {
+				logrus.WithError(cleanupErr).Infof("deleteContainer failed")
```
harche, line 124:
```suggestion
				logrus.WithError(cleanupErr).Infof("deleteContainer failed for container %s", c.ID()")
```

internal/oci/runtime_vm.go
fidencio: Harshal, you're reviewing a previous version of the series.

Please, take a look in the latest commits.

fidencio: Ah, no, my mistake, sorry!

fidencio: Changed! Thanks for the review!

internal/oci/runtime_vm.go
fidencio: Changed! Thanks for the review!

TomSweeneyRedHat:
LGTM

evanfoster:
Quick question! Was a 1.17 backport PR ever created for this?

fidencio:
@evanfoster I'm backporting that soon, while testing it together with a few other patches that should go to 1.17.

Hopefully I'll open the PR  later Today / Tomorrow in the worst case scenario.

PR #3957, opened 2020-07-15
Close the done channel in runtimeOCI#ReopenContainerLog

tedyu:
####  What type of PR is this?

/kind bug

In runtimeOCI#ReopenContainerLog, if the goroutine receives watcher error, we would execute the following (around line 1059):
```
	case err := <-errorCh:
```
where the done channel is not closed.

This PR moves the closing of done channel to the main goroutine so that done channel is always closed.

```release-note
None
```


TomSweeneyRedHat:
LGTM

tedyu:
@TomSweeneyRedHat 
Can you give /lgtm ?

Thanks

PR #3960, opened 2020-07-16
container refactor

wgahnagl:
#### What type of PR is this?

> /kind api-change
> /kind bug

/kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Begins cleaning up the container create functions, and moves the spec generation into container where it can be vendored and reused, and also is easier to test. 

```release-note
none
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -2,12 +2,22 @@ package container
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"time"
 
+	"github.com/containers/libpod/v2/pkg/annotations"
 	"github.com/containers/storage/pkg/stringid"
+	"github.com/cri-o/cri-o/internal/lib"
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	oci "github.com/cri-o/cri-o/internal/oci"
+	"github.com/cri-o/cri-o/internal/storage"
 	"github.com/cri-o/cri-o/utils"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -62,6 +72,19 @@ type Container interface {
 	// SelinuxLabel returns the container's SelinuxLabel
 	// it takes the sandbox's label, which it falls back upon
 	SelinuxLabel(string) ([]string, error)
+
+	// spec functions
+
+	// returns the spec
+	Spec() *generate.Generator
+
+	// SpecAddMount adds a mount to the container's spec
+	// it takes the rspec mount object
+	// if there is already a mount at the path specified, it removes it.
+	SpecAddMount(rspec.Mount)
+
+	// SpecAddAnnotations adds annotations to the spec.
+	SpecAddAnnotations(sandbox *sandbox.Sandbox, containerVolume []oci.ContainerVolume, mountPoint, configStopSignal string, imageResult *storage.ImageResult, isSystemd, systemdHasCollectMode bool) error
 }
 
 // container is the hidden default type behind the Container interface
@@ -72,13 +95,117 @@ type container struct {
 	id         string
 	name       string
 	privileged bool
+	spec       generate.Generator
 }
 
 // New creates a new, empty Sandbox instance
-func New(ctx context.Context) Container {
-	return &container{
-		ctx: ctx,
+func New(ctx context.Context) (Container, error) {
+	spec, err := generate.New("linux")
+	if err != nil {
+		return nil, err
 	}
+	return &container{
+		ctx:  ctx,
+		spec: spec,
+	}, nil
+}
+
+// SpecAddMount adds a specified mount to the spec
+func (c *container) SpecAddMount(r rspec.Mount) {
+	c.spec.RemoveMount(r.Destination)
+	c.spec.AddMount(r)
+}
+
+// SpecAddAnnotation adds all annotations to the spec
+func (c *container) SpecAddAnnotations(sb *sandbox.Sandbox, containerVolumes []oci.ContainerVolume, mountPoint, configStopSignal string, imageResult *storage.ImageResult, isSystemd, systemdHasCollectMode bool) (err error) {
+	// Copied from k8s.io/kubernetes/pkg/kubelet/kuberuntime/labels.go
+	const podTerminationGracePeriodLabel = "io.kubernetes.pod.terminationGracePeriod"
+
+	kubeAnnotations := c.Config().GetAnnotations()
+	created := time.Now()
+	labels := c.Config().GetLabels()
+
+	image, err := c.Image()
+	if err != nil {
+		return err
+	}
+	logPath, err := c.LogPath(sb.LogDir())
+	if err != nil {
+		return err
+	}
+	c.spec.AddAnnotation(annotations.Image, image)
+	c.spec.AddAnnotation(annotations.ImageName, imageResult.Name)
+	c.spec.AddAnnotation(annotations.ImageRef, imageResult.ID)
+	c.spec.AddAnnotation(annotations.Name, c.Name())
+	c.spec.AddAnnotation(annotations.ContainerID, c.ID())
+	c.spec.AddAnnotation(annotations.SandboxID, sb.ID())
+	c.spec.AddAnnotation(annotations.SandboxName, sb.Name())
+	c.spec.AddAnnotation(annotations.ContainerType, annotations.ContainerTypeContainer)
+	c.spec.AddAnnotation(annotations.LogPath, logPath)
+	c.spec.AddAnnotation(annotations.TTY, strconv.FormatBool(c.Config().Tty))
+	c.spec.AddAnnotation(annotations.Stdin, strconv.FormatBool(c.Config().Stdin))
+	c.spec.AddAnnotation(annotations.StdinOnce, strconv.FormatBool(c.Config().StdinOnce))
+	c.spec.AddAnnotation(annotations.ResolvPath, sb.ResolvPath())
+	c.spec.AddAnnotation(annotations.ContainerManager, lib.ContainerManagerCRIO)
+	c.spec.AddAnnotation(annotations.MountPoint, mountPoint)
+	c.spec.AddAnnotation(annotations.SeccompProfilePath, c.Config().GetLinux().GetSecurityContext().GetSeccompProfilePath())
+	c.spec.AddAnnotation(annotations.Created, created.Format(time.RFC3339Nano))
+
+	metadataJSON, err := json.Marshal(c.Config().GetMetadata())
+	if err != nil {
+		return err
+	}
+	c.spec.AddAnnotation(annotations.Metadata, string(metadataJSON))
+
+	labelsJSON, err := json.Marshal(labels)
+	if err != nil {
+		return err
+	}
+	c.spec.AddAnnotation(annotations.Labels, string(labelsJSON))
+
+	volumesJSON, err := json.Marshal(containerVolumes)
+	if err != nil {
+		return err
+	}
+	c.spec.AddAnnotation(annotations.Volumes, string(volumesJSON))
+
+	kubeAnnotationsJSON, err := json.Marshal(kubeAnnotations)
+	if err != nil {
+		return err
+	}
+	c.spec.AddAnnotation(annotations.Annotations, string(kubeAnnotationsJSON))
+
+	for k, v := range kubeAnnotations {
+		c.spec.AddAnnotation(k, v)
+	}
+	for k, v := range labels {
+		c.spec.AddAnnotation(k, v)
+	}
+	for idx, ip := range sb.IPs() {
+		c.spec.AddAnnotation(fmt.Sprintf("%s.%d", annotations.IP, idx), ip)
+	}
+
+	if isSystemd {
+		if t, ok := kubeAnnotations[podTerminationGracePeriodLabel]; ok {
+			// currently only supported by systemd, see
+			// https://github.com/opencontainers/runc/pull/2224
+			c.spec.AddAnnotation("org.systemd.property.TimeoutStopUSec", "uint64 "+t+"000000") // sec to usec
+		}
+		if systemdHasCollectMode {
+			c.spec.AddAnnotation("org.systemd.property.CollectMode", "'inactive-or-failed'")
+		}
+	}
+
+	if configStopSignal != "" {
+		// this key is defined in image-spec conversion document at https://github.com/opencontainers/image-spec/pull/492/files#diff-8aafbe2c3690162540381b8cdb157112R57
+		c.spec.AddAnnotation("org.opencontainers.image.stopSignal", configStopSignal)
+	}
+
+	return nil
+}
+
+func (c *container) Spec() *generate.Generator {
+	return &c.spec
 }
 
 // SetConfig sets the configuration to the container and validates it
--- a/pkg/container/container_test.go
+++ b/pkg/container/container_test.go
@@ -1,19 +1,140 @@
 package container_test
 
 import (
+	"encoding/json"
+	"strconv"
+	"time"
+
+	"github.com/containers/libpod/v2/pkg/annotations"
+	"github.com/cri-o/cri-o/internal/lib"
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	oci "github.com/cri-o/cri-o/internal/oci"
+	"github.com/cri-o/cri-o/internal/storage"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	"k8s.io/kubernetes/pkg/kubelet/dockershim/network/hostport"
 )
 
 var _ = t.Describe("Container", func() {
 	var config *pb.ContainerConfig
 	var sboxConfig *pb.PodSandboxConfig
+	const defaultMounts = 6
 	BeforeEach(func() {
 		config = &pb.ContainerConfig{
 			Metadata: &pb.ContainerMetadata{Name: "name"},
 		}
 		sboxConfig = &pb.PodSandboxConfig{}
+	})
+	t.Describe("SpecAddMount", func() {
+		It("should add the mount to the spec", func() {
+			sut.SpecAddMount(rspec.Mount{
+				Destination: "test",
+				Type:        "test",
+				Source:      "test",
+				Options:     []string{"test"},
+			})
+			Expect(len(sut.Spec().Mounts())).To(Equal(defaultMounts + 1))
+		})
+		It("should add only one copy to the spec", func() {
+			sut.SpecAddMount(rspec.Mount{
+				Destination: "test",
+				Type:        "test",
+				Source:      "test",
+				Options:     []string{"test"},
+			})
+			sut.SpecAddMount(rspec.Mount{
+				Destination: "test",
+				Type:        "test",
+				Source:      "test",
+				Options:     []string{"test"},
+			})
+			Expect(len(sut.Spec().Mounts())).To(Equal(defaultMounts + 1))
+		})
+	})
+	t.Describe("Spec", func() {
+		It("should return the spec", func() {
+			Expect(sut.Spec()).ToNot(Equal(nil))
+		})
+	})
+	t.Describe("SpecAddAnnotations", func() {
+		It("should set the spec annotations", func() {
+			// Given
+			sandboxConfig := &pb.PodSandboxConfig{
+				Metadata: &pb.PodSandboxMetadata{Name: "name"},
+			}
+			containerConfig := &pb.ContainerConfig{
+				Metadata: &pb.ContainerMetadata{Name: "name"},
+				Linux: &pb.LinuxContainerConfig{
+					SecurityContext: &pb.LinuxContainerSecurityContext{
+						Privileged: true,
+					},
+				},
+				Image: &pb.ImageSpec{
+					Image: "img",
+				},
+			}
+			err := sut.SetConfig(containerConfig, sandboxConfig)
+			Expect(err).To(BeNil())
+			currentTime := time.Now()
+			volumes := []oci.ContainerVolume{}
+			imageResult := storage.ImageResult{}
+			mountPoint := "test"
+			configStopSignal := "test"
+
+			sb, err := sandbox.New("sandboxID", "", "", "", "test",
+				make(map[string]string), make(map[string]string), "", "",
+				&pb.PodSandboxMetadata{}, "", "", false, "", "", "",
+				[]*hostport.PortMapping{}, false, currentTime, "")
+			Expect(err).To(BeNil())
+
+			image, err := sut.Image()
+			Expect(err).To(BeNil())
+
+			logpath, err := sut.LogPath(sb.LogDir())
+			Expect(err).To(BeNil())
+
+			metadataJSON, err := json.Marshal(sut.Config().GetMetadata())
+			Expect(err).To(BeNil())
+
+			labelsJSON, err := json.Marshal(sut.Config().GetLabels())
+			Expect(err).To(BeNil())
+
+			volumesJSON, err := json.Marshal(volumes)
+			Expect(err).To(BeNil())
+
+			kubeAnnotationsJSON, err := json.Marshal(sut.Config().GetAnnotations())
+			Expect(err).To(BeNil())
+
+			Expect(currentTime).ToNot(BeNil())
+			Expect(sb).ToNot(BeNil())
+
+			err = sut.SpecAddAnnotations(sb, volumes, mountPoint, configStopSignal, &imageResult, false, false)
+			Expect(err).To(BeNil())
+
+			Expect(sut.Spec().Config.Annotations[annotations.Image]).To(Equal(image))
+			Expect(sut.Spec().Config.Annotations[annotations.ImageName]).To(Equal(imageResult.Name))
+			Expect(sut.Spec().Config.Annotations[annotations.ImageRef]).To(Equal(imageResult.ID))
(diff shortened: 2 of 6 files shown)


server/container_create_linux.go
```diff
@@ -569,35 +506,15 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 		if !isInCRIMounts("/sys", containerConfig.GetMounts()) {
 			specgen.RemoveMount("/sys")
 			specgen.RemoveMount("/sys/fs/cgroup")
```
haircommander, line 508: considering we call RemoveMount for effectively every call to add mount, I think it makes sense to add it to `AddMountToSpec` as well.

we'd still want this RemoveMount() call, because it wouldn't be removed when we remove `/sys`, but most others will be deduplicated

haircommander: this being the one that removes `/sys/fs/cgroup`, if that wasn't clear :slightly_smiling_face:

wgahnagl: it looks to me like there's quite a few places where we don't remove mounts :thinking: 
Could we call RemoveMount with an empty string to remove nothing, or maybe overload the function with another parameter ?

haircommander: I can't think of any case where calling RemoveMount before adding the mount would be problematic. The new mount should always replace the old, anyway

wgahnagl: Places like this though ?

server/container_create_linux.go
```diff
@@ -250,25 +234,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 		}
 		for target, mode := range mounts {
 			if !isInCRIMounts(target, containerConfig.GetMounts()) {
-				mnt := rspec.Mount{
```
wgahnagl, line 253: @haircommander I was thinking I shouldn't add removing the mount in places like this, because this doesn't seem to remove it? but maybe I'm wrong !

pkg/container/container_test.go
```diff
@@ -15,6 +22,83 @@ var _ = t.Describe("Container", func() {
 		}
 		sboxConfig = &pb.PodSandboxConfig{}
 	})
+	t.Describe("AddMountToSpec", func() {
+		It("should add the mount to the spec", func() {
+			sut.AddMountToSpec("test", "test", "test", []string{"test"})
+			Expect(len(sut.GetSpec().Mounts())).To(Equal(7))
```
haircommander, line 28: can we have a const value for the number of default mounts (6 is it?) then calculate what we should Equal() in relation to that? `defaultNumMount + mountsAdded`

pkg/container/container_test.go
```diff
@@ -15,6 +22,83 @@ var _ = t.Describe("Container", func() {
...
+		})
+		It("should add only one copy to the spec", func() {
+			sut.AddMountToSpec("test", "test", "test", []string{"test"})
+			sut.AddMountToSpec("test", "test", "test", []string{"test"})
```
haircommander, line 32: same here

pkg/container/container_test.go
```diff
@@ -15,6 +22,83 @@ var _ = t.Describe("Container", func() {
...
+			Expect(len(sut.GetSpec().Mounts())).To(Equal(8))
+		})
+	})
+	t.Describe("GetSpec", func() {
+		It("should return the spec", func() {
+			Expect(sut.GetSpec()).ToNot(Equal(nil))
+		})
+	})
+	t.Describe("AddSpecAnnotations", func() {
+		(It("should set the spec annotations", func() {
```
haircommander, line 42: what's the first parenthesis doing?

pkg/container/container_test.go
```diff
@@ -15,6 +22,83 @@ var _ = t.Describe("Container", func() {
...
+			mountPoint := "test"
+			configStopSignal := "test"
+
+			Expect(currentTime).ToNot(BeNil())
+			Expect(sb).ToNot(BeNil())
+
+			err = sut.AddSpecAnnotations(sb, volumes, mountPoint, configStopSignal, &imageResult, false, false)
+			Expect(err).To(BeNil())
+
+			Expect(sut.GetSpec().Config.Annotations[annotations.Image]).To(Equal("img"))
```
haircommander, line 79: for all of these, can we compare against variables instead of hard coding twice

server/container_create_linux.go
haircommander: I think we don't remove it because we don't expect it to be there (it's a target mount in the container's rootfs, `/var/lib/containers/storage/overlay/$randomID....` which would be remarkable if it was already there. I don't think there's any harm in double removing the mount, and I'm pretty sure all calls to AddMount() are meant to replace one that exists. I think calling them together would be best, unless something breaks when we delete if it's not there

pkg/container/container_test.go
wgahnagl: fixed!

pkg/container/container_test.go
wgahnagl: nothing important, evidently!
fixed!

pkg/container/container_test.go
wgahnagl: fixed!

pkg/container/container_test.go
wgahnagl: fixed!

server/container_create_linux.go
wgahnagl: fixed !

pkg/container/container.go
```diff
@@ -62,6 +71,21 @@ type Container interface {
 	// SelinuxLabel returns the container's SelinuxLabel
 	// it takes the sandbox's label, which it falls back upon
 	SelinuxLabel(string) ([]string, error)
+
+	// spec functions
+
+	// returns the spec
+	GetSpec() *generate.Generator
```
haircommander, line 78: idiomatic go says to drop `Get` from getters, so this should just be `Spec()`

pkg/container/container.go
```diff
@@ -62,6 +71,21 @@ type Container interface {
...
+
+	// AddMountToSpec adds a mount to the container's spec
+	// it takes the destination, the mount type, the source type, the mount to remove, and the options
+	AddMountToSpec(string, string, string, string, []string)
```
haircommander, line 82: maybe `AddSpecMount` or below have it be `AddAnnotationsToSpec`. Whichever one, but I'd like them to be of the same structure

haircommander:
couple more nits, otherwise LGTM :tada: 

@umohnani8 @mrunalp @saschagrunert @kolyshkin @harche PTAL

pkg/container/container.go
wgahnagl: maybe it should be SpecAddMount, and have AddAnnotationsToSpec be SpecAddAnnotations ? so when you're trying to use them they would be grouped by "spec" and be easy to use if you're using autocomplete ?

haircommander: that's an option! another one that I just thought of is just do `AddAnnotations` and `AddMounts` and do away with specifying the spec

haircommander: I am still not too opinionated as long as they're consistent :smiley:

wgahnagl: for sure lol

wgahnagl: renamed!

pkg/container/container.go
wgahnagl: fixed!

umohnani8:
/retest
LGTM

haircommander:
>:( 

/retest

pkg/container/container.go
```diff
@@ -62,6 +71,21 @@ type Container interface {
...
 	SelinuxLabel(string) ([]string, error)
+
+	// spec functions
+
+	// returns the spec
+	Spec() *generate.Generator
+
+	// SpecAddMount adds a mount to the container's spec
+	// it takes the destination, the mount type, the source type, the mount to remove, and the options
+	SpecAddMount(string, string, string, string, []string)
```
kolyshkin, line 82: 1. Having the "mount to remove" argument seems questionable, especially since we already have a method that does exactly that.

2. I'd suggest to use variable names here, as the form of self-documentation.

IOW
```go
SpecAddMount(destination, mountType, sourceType string, options []string)
```

It reads way better than "string, string, string, string" :) and you can now omit the parameters description from the docstring.

pkg/container/container.go
```diff
@@ -62,6 +71,21 @@ type Container interface {
...
+
+	// SpecAddAnnotations adds annotations to the spec.
+	// it takes the sandbox, the list of container volumes, the mount path
+	// the configStopSignal, an image result, a bool if it is systemd,
+	// and a bool systemdHasCollectMode
+	SpecAddAnnotations(*sandbox.Sandbox, []oci.ContainerVolume, string, string, *storage.ImageResult, bool, bool) error
```
kolyshkin, line 88: same here, these "string, string" and "bool, bool" are crying to be named.

pkg/container/container.go
```diff
@@ -72,13 +96,121 @@ type container struct {
...
 	return &container{
-		ctx: ctx,
+		ctx:  ctx,
+		spec: spec,
+	}, nil
+}
+
+// SpecAddMount adds a specified mount to the spec
+func (c *container) SpecAddMount(destination, mntType, sourceType, mountToRemove string, options []string) {
+	c.spec.RemoveMount(mountToRemove)
```
kolyshkin, line 116: as pointed out above, I find the "mountToRemove" thing questionable.

pkg/container/container.go
```diff
@@ -72,13 +96,121 @@ type container struct {
...
+	c.spec.AddAnnotation(annotations.ImageRef, imageResult.ID)
+	c.spec.AddAnnotation(annotations.Name, c.Name())
+	c.spec.AddAnnotation(annotations.ContainerID, c.ID())
+	c.spec.AddAnnotation(annotations.SandboxID, sb.ID())
+	c.spec.AddAnnotation(annotations.SandboxName, sb.Name())
+	c.spec.AddAnnotation(annotations.ContainerType, annotations.ContainerTypeContainer)
+	c.spec.AddAnnotation(annotations.LogPath, logPath)
+	c.spec.AddAnnotation(annotations.TTY, fmt.Sprintf("%v", c.Config().Tty))
+	c.spec.AddAnnotation(annotations.Stdin, fmt.Sprintf("%v", c.Config().Stdin))
+	c.spec.AddAnnotation(annotations.StdinOnce, fmt.Sprintf("%v", c.Config().StdinOnce))
```
kolyshkin, line 152: I know this is not something new but rather copied from the old code, and yet while at it can you switch to using `strconv.FormatBool` for these ^^^ three, as fmt.Sprintf is pretty heavy.

pkg/container/container_test.go
```diff
@@ -1,20 +1,126 @@
...
+			Expect(sut.Spec().Config.Annotations[annotations.ImageRef]).To(Equal(imageResult.ID))
+			Expect(sut.Spec().Config.Annotations[annotations.Name]).To(Equal(sut.Name()))
+			Expect(sut.Spec().Config.Annotations[annotations.ContainerID]).To(Equal(sut.ID()))
+			Expect(sut.Spec().Config.Annotations[annotations.SandboxID]).To(Equal(sb.ID()))
+			Expect(sut.Spec().Config.Annotations[annotations.SandboxName]).To(Equal(sb.Name()))
+			Expect(sut.Spec().Config.Annotations[annotations.ContainerType]).To(Equal(annotations.ContainerTypeContainer))
+			Expect(sut.Spec().Config.Annotations[annotations.LogPath]).To(Equal(logpath))
+			Expect(sut.Spec().Config.Annotations[annotations.TTY]).To(Equal(fmt.Sprintf("%v", sut.Config().Tty)))
+			Expect(sut.Spec().Config.Annotations[annotations.Stdin]).To(Equal(fmt.Sprintf("%v", sut.Config().Stdin)))
+			Expect(sut.Spec().Config.Annotations[annotations.StdinOnce]).To(Equal(fmt.Sprintf("%v", sut.Config().StdinOnce)))
```
kolyshkin, line 112: same here

server/container_create_linux.go
```diff
@@ -548,36 +505,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
-		specgen.AddMount(sysMnt)
-
-		cgroupMnt := rspec.Mount{
-			Destination: "/sys/fs/cgroup",
-			Type:        "cgroup",
-			Source:      "cgroup",
-			Options:     []string{"nosuid", "noexec", "nodev", "rw", "relatime"},
-		}
-		specgen.AddMount(cgroupMnt)
+		ctr.SpecAddMount("sys", "sysfs", "sysfs", "/sys", []string{"nosuid", "noexec", "nodev", "rw"})
```
kolyshkin, line 513: Whoops, seems that the mount destination `/sys` became `sys`. Or is it intentional?

server/container_create_linux.go
```diff
@@ -548,36 +505,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
+		ctr.SpecAddMount("sys/fs/cgroup", "cgroup", "cgroup", "/sys/fs/cgroup", []string{"nosuid", "noexec", "nodev", "rw", "relatime"})
```
kolyshkin, line 514: same here

server/container_create_linux.go
```diff
@@ -548,36 +505,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
 			specgen.RemoveMount("/sys")
-			specgen.RemoveMount("/sys/fs/cgroup")
-			sysMnt := rspec.Mount{
-				Destination: "/sys",
-				Type:        "sysfs",
-				Source:      "sysfs",
-				Options:     []string{"nosuid", "noexec", "nodev", "ro"},
-			}
-			specgen.AddMount(sysMnt)
+			ctr.SpecAddMount("/sys", "sysfs", "sysfs", "/sys/fs/cgroup", []string{"nosuid", "noexec", "nodev", "ro"})
```
kolyshkin, line 508: After reviewing the whole commit, it seems this change is hurting readability. While previously we had code which clearly shows what's what:
```go
specgen.RemoveMount("/sys/fs/cgroup")
sysMnt := rspec.Mount{
￼				Destination: "/sys",
￼				Type:        "sysfs",
￼				Source:      "sysfs",
...
```

now we have 
```go
ctr.SpecAddMount("/sys", "sysfs", "sysfs", "/sys/fs/cgroup", ...
```
which is somewhat harder to read.

I could be OK with the increased readability in case the change would add some benefit, but I don't see it.

That's just IMHO though, feel free to ignore.

pkg/container/container.go
```diff
@@ -72,13 +96,121 @@ type container struct {
...
+	}
 	return &container{
-		ctx: ctx,
+		ctx:  ctx,
+		spec: spec,
+	}, nil
+}
+
+// SpecAddMount adds a specified mount to the spec
+func (c *container) SpecAddMount(destination, mntType, sourceType, mountToRemove string, options []string) {
```
haircommander, line 115: instead of mountToRemove, we could just unconditionally remove destination

server/container_create_linux.go
haircommander: what if we passed in an rspec.Mount object?
Then SpecAddMount would RemoveMount(m.Destination), then AddMount(m)

The idea behind the change is to move the work of translating between cri->oci from server/container_create_linux to pkg/container. Eventually, more will be moved and the purpose will be more clear, and I believe readability will be improved

haircommander:
a rebase gave you this errror
```

fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/bin/make", "msg": "# github.com/cri-o/cri-o/server\nserver/container_create_linux.go:29:2: imported and not used: \"github.com/json-iterator/go\"
```
can you rebase and fix @wgahnagl

pkg/container/container.go
wgahnagl: fixed !

server/container_create_linux.go
wgahnagl: fixed!

server/container_create_linux.go
wgahnagl: fixed!

pkg/container/container_test.go
wgahnagl: fixed!

pkg/container/container.go
wgahnagl: fixed!

pkg/container/container.go
wgahnagl: fixed!

pkg/container/container.go
wgahnagl: fixed!

pkg/container/container.go
wgahnagl: fixed !

server/container_create_linux.go
wgahnagl: passing in an rspec.Mount object now!

pkg/container/container.go
```diff
@@ -62,6 +72,18 @@ type Container interface {
...
 	// it takes the sandbox's label, which it falls back upon
 	SelinuxLabel(string) ([]string, error)
+
+	// spec functions
+
+	// returns the spec
+	Spec() *generate.Generator
+
+	// SpecAddMount adds a mount to the container's spec
+	// it takes the destination, the mount type, the source type, the mount to remove, and the options
```
kolyshkin, line 82: > the mount to remove

looks like this comment is no longer valid

kolyshkin: Well, it might make sense to say instead that if there's another mount with the same destination, it is removed.

pkg/container/container.go
```diff
@@ -72,13 +94,115 @@ type container struct {
...
+	}
 	return &container{
-		ctx: ctx,
+		ctx:  ctx,
+		spec: spec,
+	}, nil
+}
+
+// SpecAddMount adds a specified mount to the spec
+func (c *container) SpecAddMount(r rspec.Mount) {
```
kolyshkin, line 113: Now we have two docstrings for SpecAddMount. I think one is enough

pkg/container/container.go
```diff
@@ -72,13 +94,115 @@ type container struct {
...
+	c.spec.RemoveMount(r.Destination)
+	c.spec.AddMount(r)
+}
+
+// SpecAddAnnotation adds all annotations to the spec
```
kolyshkin, line 118: ditto

server/container_create_linux.go
```diff
@@ -38,7 +35,6 @@ import (
 )
 
 // Copied from k8s.io/kubernetes/pkg/kubelet/kuberuntime/labels.go
```
kolyshkin, line 37: The const is moved out of here, and this ^^^ comment (that was about the const) feels lonely. Please help it to reunite with the const it describes.

server/container_create_linux.go
```diff
@@ -796,13 +694,11 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 	sort.Sort(orderedMounts(mounts))
 
 	for _, m := range mounts {
-		mnt = rspec.Mount{
+		ctr.SpecAddMount(rspec.Mount{
```
kolyshkin, line 697: Maybe pass `rspec.Mount` via a reference, otherwise it's too much data being copied over

pkg/container/container.go
kolyshkin: 2. Perhaps `s/SpecAddMount/AddMount/`?

server/container_create_linux.go
wgahnagl: fixed !

server/container_create_linux.go
wgahnagl: fixed !

pkg/container/container.go
wgahnagl: fixed !

pkg/container/container.go
wgahnagl: there's similar comments over other functions in the file, I think I we should either remove all of them or keep all of them.

haircommander:
oop, sorry @wgahnagl need to run `gofumpt`:
https://app.circleci.com/pipelines/github/cri-o/cri-o/5199/workflows/02f21ff7-4b38-482a-984a-8eaf9042e6a3/jobs/80302

wgahnagl:
@haircommander fumpted

server/container_create_linux.go
```diff
@@ -548,36 +508,29 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
-			specgen.RemoveMount("/sys/fs/cgroup")
-			sysMnt := rspec.Mount{
-				Destination: "/sys",
-				Type:        "sysfs",
-				Source:      "sysfs",
-				Options:     []string{"nosuid", "noexec", "nodev", "ro"},
-			}
-			specgen.AddMount(sysMnt)
+			ctr.SpecAddMount(
+				rspec.Mount{
```
haircommander, line 512: nit: extra \n

wgahnagl: fixed !

haircommander:
PTAL @saschagrunert @umohnani8 @mrunalp

haircommander:
unit tests is failing with
```
Failed to compile container:



/usr/local/go/pkg/tool/linux_amd64/link: signal: killed

```
no idea what that means

saschagrunert:
> unit tests is failing with
> 
> ```
> Failed to compile container:
> /usr/local/go/pkg/tool/linux_amd64/link: signal: killed
> ```
> 
> no idea what that means

I think they took to much memory in the CI, let's check…

saschagrunert:
Indeed there seems something hanging in the unit test execution:
```
Too long with no output (exceeded 10m0s): context deadline exceeded
```

server/container_create_linux.go
```diff
@@ -867,14 +769,18 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 
 	sort.Sort(orderedMounts(mounts))
 
+	rspecMount := rspec.Mount{
+		Destination: "",
```
kolyshkin, line 773: No need to initialize fields with default values (for string the default value is "").

server/container_create_linux.go
```diff
@@ -867,14 +769,18 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
+		Type:        "bind",
+		Source:      "",
+		Options:     nil,
```
kolyshkin, line 776: ditto

server/container_create_linux.go
```diff
@@ -867,14 +769,18 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
-			Type:        "bind",
-			Source:      m.Source,
-			Destination: m.Destination,
-			Options:     append(m.Options, "bind"),
-		}
-		specgen.AddMount(mnt)
+		options := append(m.Options, "bind")
+		rspecMount.Destination = m.Destination
+		rspecMount.Source = m.Source
+		rspecMount.Options = options
```
kolyshkin, line 782: nit: `rspecMount.Options = append(m.Options, "bind")`

kolyshkin: i.e. there's no need to use a temp var `options`

server/container_create_linux.go
wgahnagl: fixed!

server/container_create_linux.go
wgahnagl: fixed!

server/container_create_linux.go
```diff
@@ -866,14 +768,15 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
+	}
 	for _, m := range mounts {
-		mnt = rspec.Mount{
-			Type:        "bind",
-			Source:      m.Source,
-			Destination: m.Destination,
-			Options:     append(m.Options, "bind"),
-		}
-		specgen.AddMount(mnt)
+		rspecMount.Destination = m.Destination
```
umohnani8, line 776: The `rspec.Mount` values look the same as it was being done before in lines 870 to 876, why not leave it as it was?

wgahnagl: I'm a little confused as to what you mean ?

wgahnagl: I'm a little confused as to what you mean?

umohnani8: It was initially 
```
mnt = rspec.Mount{
			Type:        "bind",
			Source:      m.Source,
			Destination: m.Destination,
			Options:     append(m.Options, "bind"),
		}
specgen.AddMount(mnt)
```

so why not just modify it to be 
```
ctr.SpecAddMount(rspec.Mount{
			Type:        "bind",
			Source:      m.Source,
			Destination: m.Destination,
			Options:     append(m.Options, "bind"),
		})
```
Instead of doing what you have right now
```
rspecMount := rspec.Mount{
		Type:    "bind",
		Options: nil,
	}
...
rspecMount.Destination = m.Destination
rspecMount.Source = m.Source
rspecMount.Options = append(m.Options, "bind")
ctr.SpecAddMount(rspecMount)
```

wgahnagl: fixed!

haircommander:
/retest
LGTM 
PTAL @umohnani8 @kolyshkin @saschagrunert @mrunalp

server/container_create_linux.go
```diff
@@ -702,95 +645,60 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
 			return nil, err
 		}
 	}
-	specgen.AddAnnotation(annotations.SeccompProfilePath, spp)
 
 	mountPoint, err := s.StorageRuntimeServer().StartContainer(containerID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to mount container %s(%s): %v", containerName, containerID, err)
 	}
+	err = ctr.SpecAddAnnotations(sb, containerVolumes, mountPoint, containerImageConfig.Config.StopSignal, imgResult, s.config.CgroupManager().IsSystemd(), node.SystemdHasCollectMode())
```
haircommander, line 698: we actually need  to do this after the defer below, otherwise, we risk not cleaning up the container in the storage. If this merges before you get to it, can you do a follow up?

saschagrunert (approved):
/lgtm
/retest

server/container_create_linux.go
wgahnagl: fixed !

PR #3969, opened 2020-07-17
Fix static binary mode retrieval for musl toolchains

saschagrunert:
#### What type of PR is this?

/kind bug

The returned error is `Not a valid dynamic program` which is now catched
by the binary check as well.
None

None

```release-note
- Fixed `crio version` binary mode parsing on musl toolchains
```

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -176,7 +176,8 @@ func getLinkmode() string {
 		return ""
 	}
 	if _, err = utils.ExecCmd("ldd", abspath); err != nil {
-		if strings.Contains(err.Error(), "not a dynamic executable") {
+		if strings.Contains(err.Error(), "not a dynamic executable") ||
+			strings.Contains(strings.ToLower(err.Error()), "not a valid dynamic program") {
 			return "static"
 		}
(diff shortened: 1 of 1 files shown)


internal/version/version.go
```diff
@@ -176,7 +176,8 @@ func getLinkmode() string {
 		return ""
 	}
 	if _, err = utils.ExecCmd("ldd", abspath); err != nil {
-		if strings.Contains(err.Error(), "not a dynamic executable") {
+		if strings.Contains(err.Error(), "not a dynamic executable") ||
+			strings.Contains(strings.ToLower(err.Error()), "not a valid dynamic program") {
```
TomSweeneyRedHat, line 180: Always worries me to check error message strings rather than error types.  But `¯\_(ツ)_/¯`

TomSweeneyRedHat:
LGTM
and happy green test buttons.
one small concern raised and can be ignored.

internal/version/version.go
saschagrunert: The question is how we would do that when interacting with a CLI. :thinking:

PR #3971, opened 2020-07-17
Use /usr/bin/env bash in crio-shutdown.service

saschagrunert:
#### What type of PR is this?

/kind bug

Some distributions have bash in other paths so we should call out to
`/usr/bin/env` to verify that the unit file works there as well.

None

None

```release-note
- Fix path to bash via `/usr/bin/env` in crio-shutdown.service
```


haircommander:
LGTM

TomSweeneyRedHat:
LGTM
once tests are happy

PR #3973, opened 2020-07-17
node: fix panic if /sys/fs/cgroup failed to stat

haircommander:
#### What type of PR is this?

 /kind bug

our rhel 7 boxes that are used to package cri-o are failing with the following error:
```
panic: cannot statfs cgroup root
goroutine 1 [running]:
```
this is because the libcontainer cgroup packages opts to panic when the stat on /sys/fs/cgroup fails, rather than return an error

The point of the node package is to gracefully catch and handle these errors, so we switch to libpod version instead, and will now print
we failed to stat, rather than panic on it.

Note: these nodes are not suitable for running workloads, as runc will later panic, but packaging (in this case, building a config) should not panic because of it.

```release-note
fixes panic when /sys/fs/cgroup can't be stat'ed
```

### DIFF
--- a/internal/config/node/node.go
+++ b/internal/config/node/node.go
@@ -42,6 +42,13 @@ func ValidateConfig() error {
 			fatal:     false,
 		},
 		{
+			name:      "cgroup v2",
+			init:      CgroupIsV2,
+			err:       &cgroupIsV2Err,
+			activated: nil,
+			fatal:     false,
+		},
+		{
 			name:      "systemd CollectMode",
 			init:      SystemdHasCollectMode,
 			err:       &systemdHasCollectModeErr,
@@ -58,7 +65,9 @@ func ValidateConfig() error {
 			}
 			logrus.Error(err.Error())
 		}
-		logrus.Infof("node configuration value for %s is %v", i.name, *i.activated)
+		if i.activated != nil {
+			logrus.Infof("node configuration value for %s is %v", i.name, *i.activated)
+		}
 	}
 	return nil
 }
--- a/internal/config/node/cgroups.go
+++ b/internal/config/node/cgroups.go
@@ -6,7 +6,8 @@ import (
 	"os"
 	"sync"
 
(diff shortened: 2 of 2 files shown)


haircommander:
sorry @kolyshkin I know you were against this approach, but libcontainer's panic is making cri-o 1.19 not build on rhel 7 :\ I still really want to keep this initialization in the node package, but if you have other suggestions on how to handle this, please let me know

haircommander:
cc @lsm5

haircommander:
```
# time="2020-07-17T17:37:04Z" level=error msg="removing the pod sandbox \"cd5a923a3874c86cfb66854d37b083da2a7846e33daa3ecee63be0e2808c8fd2\" failed: rpc error: code = Unknown desc = unable to remove managed namespaces: Removing namespaces encountered the following errors [unlinkat /var/run/ipcns/85993b37-74f3-4422-8b00-6c3920f0aca7: device or resource busy unlinkat /var/run/netns/85993b37-74f3-4422-8b00-6c3920f0aca7: device or resource busy]"
```
I've never seen [this](https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3973/test_pull_request_crio_integration_rhel/13850/artifacts/testout.txt) before

internal/config/node/node.go
```diff
@@ -41,6 +41,13 @@ func ValidateConfig() error {
 			activated: &cgroupHasMemorySwap,
 			fatal:     false,
 		},
+		{
+			name:      "cgroup v2",
+			init:      CgroupIsV2,
+			err:       &cgroupIsV2Err,
+			activated: nil,
```
TomSweeneyRedHat, line 48: Is there any way to set an appropriate true/false value here?

haircommander: no, because it's internal to libpod. we'd have to keep our own boolean somehow, which I think would be clunky

umohnani8:
LGTM
@kir @giuseppe PTAL

PR #3987, opened 2020-07-21
runtime_vm: Avoid possible deadlock on UpdateContainerStatus()

fidencio:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

StartContainer() function will hold the opLock, while a goroutine will
be waiting for the container to terminate. Once it happens, the
goroutine would update the container status, calling then
UpdateContainerStatus() which will try to get the opLock and,
consequently, a deadlock will happen.

In order to avoid such situation, let's create the
updateContainerStatus() helper, which actually does status update but
does not try to get the opLock, leaving it up to the caller to do so.

This new helper will be used from both StartContainer() and
UpdateContainerStatus() functions.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #3998, opened 2020-07-22
runtimeVM: Cleanup a "Completed" container

fidencio:
#### What type of PR is this?

/kind bug

When the container finishes its execution a containerd-shim-kata-v2
process is left behind (only when using CRI-O).

The reason for that seems to be CRI-O not doing a cleanup of the process
whenever the container state has changed its state from running to
stopped.

The most reasonable way found to perform such cleanup seems to be taking
advantage of the goroutine used to update the container status and
performing the cleanup there, whenever it's needed.

Related: https://github.com/kata-containers/runtime/issues/2719

This is yet another part of the series trying to not leave container-shim-kata-v2 processes around, but this time focusing on cleaning up when the container finishes its execution without errors.

This series also includes the patch provided in https://github.com/cri-o/cri-o/pull/3987

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -258,6 +258,12 @@ func (r *runtimeVM) StartContainer(c *Container) error {
 			if err1 := r.updateContainerStatus(c); err1 != nil {
 				logrus.Warningf("error updating container status %v", err1)
 			}
+
+			if c.state.Status == ContainerStateStopped {
+				if err1 := r.deleteContainer(c, true); err1 != nil {
+					logrus.WithError(err1).Infof("deleteContainer failed for container %s", c.ID())
+				}
+			}
 		} else {
 			logrus.Warningf("wait for %s returned: %v", c.ID(), err)
 		}


fidencio:
@amshinde, we've talked about this on Slack a few days ago, would you mind taking a look on whether the approach is sane?

fidencio:
Oops, my bad, now the PR as it should be. :-)

internal/oci/runtime_vm.go
```diff
@@ -255,9 +255,15 @@ func (r *runtimeVM) StartContainer(c *Container) error {
 	go func() {
 		_, err := r.wait(r.ctx, c.ID(), "")
 		if err == nil {
-			if err1 := r.UpdateContainerStatus(c); err1 != nil {
+			if err1 := r.updateContainerStatus(c); err1 != nil {
 				logrus.Warningf("error updating container status %v", err1)
 			}
+
+			if c.state.Status == ContainerStateStopped {
```
mrunalp, line 262: Do we have to clean up the container here or we can wait for kubelet to notice the status from ContainerStatus and then call StopContainer and RemoveContainer?

fidencio: That's a really good question.

At this point, StopContainer() has been called already, we sent the signal down to kata.
virtiofs & qemu processes are stopped, but containerd-shim-kata-v2 will still be around and will be consuming a high amount of CPU, which doesn't happen at all when using containerd.

That's the way I, with a not so good understanding of the whole architecture, could come up with what may be a reasonable solution.

I'm 100% open to a different approach if this one is wrong. :-)

mrunalp: Thanks! I think it's fine to do the cleanup here to clean up heavy processes here and then leave the rest of the cleanup to the subsequent StopContainer call from the kubelet :)

haircommander:
/approve
LGTM

umohnani8:
/retest
LGTM

fidencio:
Rebased atop of main ...

fidencio:
Victory! :-)

PR #4004, opened 2020-07-23
crio: use github.com/pquerna/ffjson/ffjson instead of encoding/json

giuseppe:
#### What type of PR is this?

/kind cleanup

faster json encoding/decoding and better usage of memory since buffers can be added to a pool.

```release-note
NONE
```

### DIFF
--- a/internal/storage/runtime.go
+++ b/internal/storage/runtime.go
@@ -2,7 +2,6 @@ package storage
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"time"
 
@@ -14,6 +13,7 @@ import (
 	"github.com/containers/storage/pkg/idtools"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	json "github.com/pquerna/ffjson/ffjson"
 	"github.com/sirupsen/logrus"
 )
 
--- a/internal/storage/runtime.go
+++ b/internal/storage/runtime.go
@@ -256,6 +256,7 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
 	if err != nil {
 		return ContainerInfo{}, err
 	}
+	defer json.Pool(mdata)
 
 	// Build the container.
 	names := []string{metadata.ContainerName}
@@ -392,6 +393,7 @@ func (r *runtimeService) SetContainerMetadata(idOrName string, metadata *Runtime
 		logrus.Debugf("failed to encode metadata for %q: %v", idOrName, err)
 		return err
 	}
+	defer json.Pool(mdata)
 	return r.storageImageServer.GetStore().SetMetadata(idOrName, string(mdata))
 }
 
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -3,7 +3,6 @@
 package server
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -33,6 +32,7 @@ import (
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/pkg/errors"
+	json "github.com/pquerna/ffjson/ffjson"
 	"golang.org/x/net/context"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -345,11 +345,12 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 		return nil, err
 	}
 
-	volumesJSON, err := json.Marshal(containerVolumes)
-	if err != nil {
+	if volumesJSON, err := json.Marshal(containerVolumes); err != nil {
 		return nil, err
+	} else { // nolint: golint
+		specgen.AddAnnotation(annotations.Volumes, string(volumesJSON))
+		json.Pool(volumesJSON)
 	}
-	specgen.AddAnnotation(annotations.Volumes, string(volumesJSON))
 
 	configuredDevices, err := getDevicesFromConfig(ctx, &s.config)
 	if err != nil {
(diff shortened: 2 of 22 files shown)


TomSweeneyRedHat:
LGTM

haircommander:
LGTM, do we need to do any generation? I see in the docs one's suppose to call `ffjson` after changing the structs, should we add this to the makefile?

giuseppe:
> LGTM, do we need to do any generation? I see in the docs one's suppose to call `ffjson` after changing the structs, should we add this to the makefile?

the current implementation is just taking advantage of `json.Pool(...)` so we avoid unnecessary memory allocations.  I've seen the static parser is used by containers/storage but I've never looked int it.  Since we are already using ffjson, it probably makes sense to take full advantage of it.

mrunalp:
@kolyshkin ptal.

internal/storage/runtime.go
```diff
@@ -256,6 +256,7 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
 	if err != nil {
 		return ContainerInfo{}, err
 	}
+	defer json.Pool(mdata)
```
kolyshkin, line 259: in this case it might make more sense to call `json.Pool(mdata)` right after the data is used, i.e. after line 273

server/container_create_linux.go
```diff
@@ -349,6 +349,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 	if err != nil {
 		return nil, err
 	}
+	defer json.Pool(volumesJSON)
```
kolyshkin, line 352: same here, it makes more sense to explictly call `json.Pool(volumesJSON)` right after `specgen.AddAnnotation`,

server/container_create_linux.go
```diff
@@ -694,18 +695,21 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 	if err != nil {
 		return nil, err
 	}
+	defer json.Pool(metadataJSON)
```
kolyshkin, line 698: ditto

server/container_create_linux.go
```diff
@@ -694,18 +695,21 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
 	specgen.AddAnnotation(annotations.Metadata, string(metadataJSON))
 
 	labelsJSON, err := json.Marshal(labels)
 	if err != nil {
 		return nil, err
 	}
+	defer json.Pool(labelsJSON)
```
kolyshkin, line 705: ditto

server/container_create_linux.go
```diff
@@ -694,18 +695,21 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
 	specgen.AddAnnotation(annotations.Labels, string(labelsJSON))
 
 	kubeAnnotationsJSON, err := json.Marshal(kubeAnnotations)
 	if err != nil {
 		return nil, err
 	}
+	defer json.Pool(kubeAnnotationsJSON)
```
kolyshkin, line 712: ditto

server/container_create_linux.go
kolyshkin: In such case, though, if the variable will be used in the code below, we are doomed.

So maybe limit its scope, something like
```go
if volumesJSON, err := json.Marshal(containerVolumes); err != nil {
  return nil, err
} else {
  specgen.AddAnnotation(annotations.Volumes, string(volumesJSON))
  json.Pool(volumesJSON)
}
```

server/sandbox_run_linux.go
```diff
@@ -355,6 +359,7 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 	if err != nil {
 		return nil, err
 	}
+	defer json.Pool(portMappingsJSON)
```
kolyshkin, line 362: ditto

server/sandbox_run_linux.go
```diff
@@ -570,6 +575,7 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 			if err != nil {
 				return nil, err
 			}
+			defer json.Pool(cniResultJSON)
```
kolyshkin, line 578: ditto

kolyshkin:
LGTM overall, left some optional suggestions.

@giuseppe do you have any performance data?

giuseppe:
> @giuseppe do you have any performance data?

I've done a quick test running 50 containers with and without the patch.  Immediately after all the containers are running, I've read   the memory usage from the crio cgroup and it passes from 84M to 77M.

mheon:
Question: why FFJSON? We used it in Podman land for a while, but ended up dropping it because of incompatibilities with the `encoding/json` package. We eventually settled on `https://github.com/json-iterator/go` which seems to be faster and does not have the compatibility gotchas that we saw with ffjson.

giuseppe:
> Question: why FFJSON? We used it in Podman land for a while, but ended up dropping it because of incompatibilities with the `encoding/json` package. We eventually settled on `https://github.com/json-iterator/go` which seems to be faster and does not have the compatibility gotchas that we saw with ffjson.

I've only considered the memory usage.  What sort of incompatibilities?  Do you have a pointer?  If there are incompatibilities, then I think we should avoid it

giuseppe:
given the last comment by @mheon I am going to close this PR and give github.com/json-iterator/go a try first.

mheon:
I'll try and locate the Podman issues, but the gist was that it does not obey the `MarshalJSON()` method on structs and types, thus leading to certain things (one that sticks out in my memory is IP addresses) being represented in a way incompatible with the standard library.

giuseppe:
> I'll try and locate the Podman issues, but the gist was that it does not obey the `MarshalJSON()` method on structs and types, thus leading to certain things (one that sticks out in my memory is IP addresses) being represented in a way incompatible with the standard library.

no worries, I get similar results with json-iterator/go and it seems a better candidate anyway, I am going to open a new PR using json-iterator/go

PR #4006, opened 2020-07-23
Restore Sandbox createdAt timestamp on cri-o restart

rhafer:
#### What type of PR is this?

/kind bug

When restarting cri-o, try to restore the createdAt timestamp of Sandboxes from the annotations. Previously the timestamp was reset to time.Now() when restoring Sandboxes. As Sandbox are not guaranteed to be restored in creation order in server.restore() this could lead to ordering issues in calls like e.g. ListPodSandbox which would not return the Pods in the orginial creations order.

To reproduce this issue in crio, create a couple of Sandboxes (with `crictl`) using the same name but different "Attempt" numbers. Then restart crio and run `crio pods`. Two things should be visible:

* the `CREATED` time in the `crio pods` is reset with every crio restart. While the createdAt annoation seen with `crio inspectp` is staying the same
* The order in which the pods are listed might change during restarts and is not stable

The first issue might just be cosmetic. The second however is causing issues with kubelet, which seems to rely on the pods being returned in a certain order. See e.g. https://github.com/kubernetes/kubernetes/issues/93064 for some more details.
(I am also looking into a possible fix on the kubelet side, but still believe that we need this fix for cri-o  as well to preserve the timestamps after restart.

None

```release-note
Fixed crio restart behavior to make sure that Pod creation timestamps are restored and the order in the list of pods stays stable across restarts
```

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -194,7 +194,12 @@ func (c *ContainerServer) LoadSandbox(id string) (retErr error) {
 		return errors.Wrapf(err, "error unmarshalling %s annotation", annotations.NamespaceOptions)
 	}
 
-	sb, err := sandbox.New(id, m.Annotations[annotations.Namespace], name, m.Annotations[annotations.KubeName], filepath.Dir(m.Annotations[annotations.LogPath]), labels, kubeAnnotations, processLabel, mountLabel, &metadata, m.Annotations[annotations.ShmPath], m.Annotations[annotations.CgroupParent], privileged, m.Annotations[annotations.RuntimeHandler], m.Annotations[annotations.ResolvPath], m.Annotations[annotations.HostName], portMappings, hostNetwork)
+	created, err := time.Parse(time.RFC3339Nano, m.Annotations[annotations.Created])
+	if err != nil {
+		return errors.Wrap(err, "parsing created timestamp annotation")
+	}
+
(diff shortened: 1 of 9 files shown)


internal/lib/container_server.go
```diff
@@ -194,7 +194,12 @@ func (c *ContainerServer) LoadSandbox(id string) (retErr error) {
 		return errors.Wrapf(err, "error unmarshalling %s annotation", annotations.NamespaceOptions)
 	}
 
-	sb, err := sandbox.New(id, m.Annotations[annotations.Namespace], name, m.Annotations[annotations.KubeName], filepath.Dir(m.Annotations[annotations.LogPath]), labels, kubeAnnotations, processLabel, mountLabel, &metadata, m.Annotations[annotations.ShmPath], m.Annotations[annotations.CgroupParent], privileged, m.Annotations[annotations.RuntimeHandler], m.Annotations[annotations.ResolvPath], m.Annotations[annotations.HostName], portMappings, hostNetwork)
+	created, err := time.Parse(time.RFC3339Nano, m.Annotations[annotations.Created])
+	if err != nil {
+		return err
```
saschagrunert, line 199:
```suggestion
		return errors.Wrap(err, "parsing created timestamp annotation")
```

TomSweeneyRedHat:
LGTM
TYVM @rhafer !

haircommander:
nice find, LGTM
/approve

haircommander:
/hold

I agree with the nits (err wrapf and integration test)

internal/lib/container_server.go
rhafer: Error wrapping is done now. Working on a test now (which might be a little more involved ...)

rhafer: > Working on a test now (which might be a little more involved ...)

Actually, creating the test was easy. Getting the test suite to run locally was somewhat more complicated ;-)

@saschagrunert I've added the testcase to one of the existing tests in `restore.bats` which seemed to be a good fit. If you prefer I can also move it to a separate test in that file.

PR #4008, opened 2020-07-23
[1.19] Point to k8s release-1.19 branch for tests

umohnani8:

#### What type of PR is this?

/kind dependency-change

And update the tutorial doc to say release-1.19 instead
of release-1.18.

```release-note
None
```

### DIFF
--- a/contrib/test/integration/main.yml
+++ b/contrib/test/integration/main.yml
@@ -25,7 +25,7 @@
     - name: clone build and install kubernetes
       include: "build/kubernetes.yml"
       vars:
-        k8s_git_version: "release-1.18"
+        k8s_git_version: "release-1.19"
         k8s_github_fork: "kubernetes"
         crio_socket: "/var/run/crio/crio.sock"
       when: "(cgroupv2 is undefined) or (cgroupv2 == False) | bool"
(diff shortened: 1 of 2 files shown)


contrib/test/integration/main.yml
```diff
@@ -25,15 +25,15 @@
...
-        k8s_git_version: "release-1.18"
+        k8s_git_version: "release-1.19"
         k8s_github_fork: "kubernetes"
         crio_socket: "/var/run/crio/crio.sock"
       when: "(cgroupv2 is undefined) or (cgroupv2 == False) | bool"
 
     - name: clone build and install kubernetes for cgroup v2
       include: "build/kubernetes.yml"
       vars:
-        k8s_git_version: "master"
```
saschagrunert, line 36: What is the reason that we do not stick to the master here?

umohnani8: Yeah, I wasn't entirely sure about this one. In 1.18 we were using @giuseppe's branch. I was testing out if this works with the release-1.19 branch, and looks like it does. But if we want to keep cgroupv2 pointing to master, I will revert it here.

saschagrunert: Right now they're still the same with some slight delay, but if we apply Code Thaw (in some weeks) in Kubernetes, then we will not be able to test the latest master changes any more. I think we could simply stick to master, WDYT?

umohnani8: Sure, fixed!

TomSweeneyRedHat:
Changes LGTM once tests are happy

haircommander:
/hold

we've decided to wait until we stop forwarding from master to do this (should be in a week or two)

umohnani8:
/hold cancel

This can be merged now. @mrunalp @haircommander PTAL

PR #4009, opened 2020-07-23
Use CircleCI cache for static builds

saschagrunert:
#### What type of PR is this?

/kind feature

Using the CircleCI cache should speed-up building the static binary.

None
None

```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,12 @@ executors:
     <<: *stdenv
     working_directory: *workdir
 
+  container-nix:
+    docker:
+      - image: nixos/nix:latest
+    <<: *stdenv
+    working_directory: *workdir
+
   machine:
     machine:
       image: ubuntu-1604:201903-01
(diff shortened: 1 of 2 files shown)


hswong3i:
Could we just simply preserve/restore the `/nix` directory with CI cache? This could be more portable as example for crun/conmon/skopeo/build/podman with different CI service?

saschagrunert:
> Could we just simply preserve/restore the `/nix` directory with CI cache? This could be more portable as example for crun/conmon/skopeo/build/podman with different CI service?

Thank you for the hint, I more or less thought the same during the implementation, let my give it a try :)

saschagrunert:
Way faster, PTAL @haircommander @hswong3i

hswong3i:
@saschagrunert https://github.com/cri-o/cri-o/pull/4012 include this changes and cache now working, too.

Some more founding:
- Due to `nixos/nix` limitation which without `bash` bundled, not every CI with Docker engine direct support would happy with :-(
- Caching `/nix` directly may also make some CI not happy due to permission issue; therefore I cache it with `.cache` and export/import it in case already warm up, else copy from `nixos/nix` directly.
- Cache for Circle CI (for [CRI-O](https://github.com/cri-o/cri-o)) and GitHub Actions (for [crun](https://github.com/containers/crun)) are running well across multiple build with same cache key; BTW Cirrus CI (for [conmon](https://github.com/containers/conmon), [Skopeo](https://github.com/containers/skopeo) and [Buildah](https://github.com/containers/buildah)) only preserve it across multiple tasks inside a single build (please correct me) so caching didn't seems really functioning (although cache is generated and upload correctly, but always say "cache miss" during initial cache fetching phase) :-(

saschagrunert:
> @saschagrunert #4012 include this changes and cache now working, too.
> 
> Some more founding:
> 
> * Due to `nixos/nix` limitation which without `bash` bundled, not every CI with Docker engine direct support would happy with :-(

CircleCI does, so I prefer to use the native image over the vm-wrapped approach.

> * Caching `/nix` directly may also make some CI not happy due to permission issue; therefore I cache it with `.cache` and export/import it in case already warm up, else copy from `nixos/nix` directly.

This issue should be gone if we cache directly in nixos/nix (see the CI run here).

PR #4005, opened 2020-07-23
Remove stale nix targets and files

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
Those files and images are not needed any more so we can remove them.
#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4012, opened 2020-07-24
Speedup static build by utilizing CI cache on `/nix` folder

hswong3i:
#### What type of PR is this?

> /kind cleanup
> /kind feature

Speedup static build by utilizing CI cache on `/nix` folder

Also see:
- https://github.com/containers/crun/pull/428
- https://github.com/containers/conmon/pull/189
- https://github.com/containers/skopeo/pull/973
- https://github.com/containers/buildah/pull/2428
- https://github.com/containers/podman/pull/7076
- https://github.com/cri-o/cri-o/pull/4012

Fixes #4005
Fixes #4009

Some more founding:
- Due to `nixos/nix` limitation which without `bash` bundled, not every CI with Docker engine direct support would happy with :-(
- Caching `/nix` directly may also make some CI not happy due to permission issue; therefore I cache it with `.cache` and export/import it in case already warm up, else copy from `nixos/nix` directly.
- Cache for Circle CI (for [CRI-O](https://github.com/cri-o/cri-o)) and GitHub Actions (for [crun](https://github.com/containers/crun)) are running well across multiple build with same cache key; BTW Cirrus CI (for [conmon](https://github.com/containers/conmon), [Skopeo](https://github.com/containers/skopeo) and [Buildah](https://github.com/containers/buildah)) only preserve it across multiple tasks inside a single build (please correct me) so caching didn't seems really functioning (although cache is generated and upload correctly, but always say "cache miss" during initial cache fetching phase) :-(

None

```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,29 +156,42 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
+      - restore_cache:
+          keys:
+            - nix-v1-{{ checksum "nix/nixpkgs.json" }}
       - run:
           name: build
           command: |
             set -ex
             sudo mkdir -p /nix
-            sudo docker run --rm --privileged -ti -v /:/mnt nixos/nix cp -rfT /nix /mnt/nix
+            sudo mkdir -p .cache
+            sudo mount --bind .cache /nix
+            if [[ -z $(ls -A /nix) ]]; then sudo docker run --rm --privileged -ti -v /:/mnt nixos/nix cp -rfT /nix /mnt/nix; fi
             sudo docker run --rm --privileged -ti -v /nix:/nix -v ${PWD}:${PWD} -w ${PWD} nixos/nix nix --print-build-logs --option cores 8 --option max-jobs 8 build --file nix/
-            mkdir -p bin
-            cp -r result/bin bin/static
+            sudo chown -Rf $(whoami) .cache
+      - save_cache:
+          key: nix-v1-{{ checksum "nix/nixpkgs.json" }}
+          paths:
+            - .cache
       - run:
           name: Show CRI-O version for static binaries
           command: |
-            ./bin/static/crio version
-      - persist_to_workspace:
-          root: .
-          paths:
-            - bin/static
+            ./result/bin/crio version
       - store_artifacts:
           path: result/bin/crio
       - store_artifacts:
           path: result/bin/crio-status
       - store_artifacts:
           path: result/bin/pinns
+      - run:
+          name: persist_to_workspace
+          command: |
+            mkdir -p bin/static
+            cp -rfT result/bin bin/static
+      - persist_to_workspace:
+          root: .
+          paths:
+            - bin/static
 
   build-test-binaries:
     executor: container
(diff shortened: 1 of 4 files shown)


hswong3i:
@saschagrunert `sudo mount --bind .cache /nix` could even simplify the cache export/import issue.

Also I would suggest for using machine driver + nix in docker image for testing, rather than using docker driver directly: it could be more flexible for configuration, and more portable across different CI platform when compare with crun/conmon/skopeo/buildah/podman ;-)

haircommander:
```
[sig-network] IngressClass [Feature:Ingress] should set default value on new IngressClass [Serial]
```
[ref](https://prow.ci.openshift.org/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/4012/test_pull_request_crio_e2e_features_fedora/12519/)

I've seen this before, seems like a flaky one

/retest

hswong3i:
@haircommander those failed tests shouldn't related to this PR, am I correct?

haircommander:
you are correct, and e2e_crun_cgroupv2 is not required

/approve

PTAL @saschagrunert

.circleci/config.yml
```diff
@@ -156,29 +156,42 @@ jobs:
...
+            - nix-v1-{{ checksum "nix/nixpkgs.json" }}
       - run:
           name: build
           command: |
             set -ex
             sudo mkdir -p /nix
-            sudo docker run --rm --privileged -ti -v /:/mnt nixos/nix cp -rfT /nix /mnt/nix
+            sudo mkdir -p .cache
+            sudo mount --bind .cache /nix
+            if [[ -z $(ls -A /nix) ]]; then sudo docker run --rm --privileged -ti -v /:/mnt nixos/nix cp -rfT /nix /mnt/nix; fi
```
saschagrunert, line 169: I'm still wondering why we cannot get rid of this docker execution wrapped in a VM? Using the nixos/nix container image directly here seems way more straight forward from my point of view.

Makefile
```diff
@@ -175,10 +173,12 @@ bin/crio-status: $(GO_FILES) .gopathok
...
-	$(CONTAINER_RUNTIME) run --rm -it -v $(shell pwd):/cri-o:Z $(TESTIMAGE_NIX) sh -c \
-		"nix build -f cri-o/nix && \
-		mkdir -p cri-o/bin && \
-		cp -r result/bin cri-o/bin/static"
+	$(CONTAINER_RUNTIME) run --rm --privileged -ti -v /:/mnt \
+		nixos/nix cp -rfT /nix /mnt/nix
+	$(CONTAINER_RUNTIME) run --rm --privileged -ti -v /nix:/nix -v ${PWD}:${PWD} -w ${PWD} \
+		nixos/nix nix --print-build-logs --option cores 8 --option max-jobs 8 build --file nix/
+	mkdir -p bin
+	cp -r result/bin bin/static
```
saschagrunert, line 181: Same for this, why cannot we stick to a local nix installation here?

.circleci/config.yml
hswong3i: 1. Basically because not all CI service like container with non-bash based. `nixos/nix` only come with `/bin/sh` so when I tried with GitHub Action/Travis CI/Cirrus CI/Circle CI across containers/* projects, some of them (as I remember should be Cirrus CI which used by conmon/buildah/podman) are not happy with ;-(
2. Another use case is crun: it is now wrapping the build of static binary into its testing script, and the script itself already running inside VM. Migrate to Docker based require for a rewrite of CI logic :-(
3. The cache folder also need to be handle in care, e.g. Cirrus CI having bug with it (see https://github.com/cirruslabs/cirrus-ci-docs/issues/681#issuecomment-665410656). Using VM based could provide more flexibility for our implementation.

P.S. If we could simplify the choose of CI platform across containers/* and cri-o/cri-o, and that selected CI platform support Docker based test for nix build simply (+ cache), I would vote for Docker based for sure ;-)

PR #4020, opened 2020-07-24
Remove unnecessary err assignment

tedyu:
#### What type of PR is this?

/kind bug

In Server#runPodSandbox, when both parseDNSOptions and removeFile calls return error, we shouldn't assign err1 to err because that would erase the reason behind parseDNSOptions failure.

```release-note
None
```


haircommander:
/approve 

LGTM

PR #4022, opened 2020-07-27
added irq smp balance and cpu cfs quota control

jianzzha:
added handling for irq smp balance and cpu cfs quota control
added safety check step to make them only work on pods with exclusive cpus

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

One of the requirements for real time application is to minimize scheduling delay and interrupt. When a real time application is allocated with exclusive cpu set, we can prevent these cpus from handling interrupt work by disabling the irq smp balance on these cpus.  Additionally, since these cpus are to be solely used by this particular application, cpu cfs quota is not needed on these cpus. As a matter of fact, we have seen inappropriate throttle issues caused by cpu cfs quota. So it is a good practice to disable cpu cfs quota where it is not needed.
  

None

Yes 

```release-note
Added the capability to optionally disable IRQ SMP balancing and CPU CFS quota on a pod's exclusive CPU(s) via a "high-performance" runtime handler. The Kubernetes users can set up a RuntimeClass referring to this "high-performance" handler, and use this RuntimeClass in a pod spec. The prerequisite is: the pod has to be QoS "guaranteed" with whole CPU(s) requested. When this prerequisite is met, if the pod has an annotation irq-load-balancing.crio.io: "true", then IRQ SMP balancing will be disabled at the pod run time; if the pod has an annotation cpu-quota.crio.io: "true", then CPU CFS quota will be disabled at the pod run time.
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks.go
@@ -5,15 +5,19 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
 
-	"github.com/cri-o/cri-o/internal/log"
-
+	"github.com/cri-o/cri-o/internal/config/cgmgr"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
-
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 )
@@ -25,7 +29,12 @@ const (
 
 const (
 	annotationCPULoadBalancing = "cpu-load-balancing.crio.io"
+	annotationCPUQuota         = "cpu-quota.crio.io"
+	annotationIRQLoadBalancing = "irq-load-balancing.crio.io"
+	annotationTrue             = "true"
 	schedDomainDir             = "/proc/sys/kernel/sched_domain"
+	irqSmpAffinityProcFile     = "/proc/irq/default_smp_affinity"
+	cgroupMountPoint           = "/sys/fs/cgroup"
 )
 
 // HighPerformanceHooks used to run additional hooks that will configure a system for the latency sensitive workloads
@@ -33,10 +42,43 @@ type HighPerformanceHooks struct{}
 
 func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-start hook for the container %q", HighPerformance, c.ID())
+
+	if isCgroupParentBurstable(s) {
+		log.Infof(ctx, "Container %q is a burstable pod. Skip PreStart.", c.ID())
+		return nil
+	}
+	if isCgroupParentBestEffort(s) {
+		log.Infof(ctx, "Container %q is a besteffort pod. Skip PreStart.", c.ID())
+		return nil
+	}
+	if !isContainerRequestWholeCPU(c) {
+		log.Infof(ctx, "Container %q requests partial cpu(s). Skip PreStart", c.ID())
+		return nil
+	}
+
 	// disable the CPU load balancing for the container CPUs
 	if shouldCPULoadBalancingBeDisabled(s.Annotations()) {
+		log.Infof(ctx, "Disable cpu load balancing for container %q", c.ID())
 		if err := setCPUSLoadBalancing(c, false, schedDomainDir); err != nil {
+			return errors.Wrap(err, "set CPU load balancing")
+		}
+	}
+	// disable the IRQ smp load balancing for the container CPUs
+	if shouldIRQLoadBalancingBeDisabled(s.Annotations()) {
+		log.Infof(ctx, "Disable irq smp balancing for container %q", c.ID())
+		if err := setIRQLoadBalancing(c, false, irqSmpAffinityProcFile); err != nil {
+			return errors.Wrap(err, "set IRQ load balancing")
+		}
+	}
+	// disable the CFS quota for the container CPUs
+	if shouldCPUQuotaBeDisabled(s.Annotations()) {
+		log.Infof(ctx, "Disable cpu cfs quota for container %q", c.ID())
+		cpuMountPoint, err := cgroups.FindCgroupMountpoint(cgroupMountPoint, "cpu")
+		if err != nil {
 			return err
+		}
+		if err := setCPUQuota(cpuMountPoint, s.CgroupParent(), c, false); err != nil {
+			return errors.Wrap(err, "set CPU CFS quota")
 		}
 	}
 
@@ -45,18 +87,59 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 
 func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-stop hook for the container %q", HighPerformance, c.ID())
+
+	if isCgroupParentBurstable(s) {
+		log.Infof(ctx, "Container %q is a burstable pod. Skip PreStop.", c.ID())
+		return nil
+	}
+	if isCgroupParentBestEffort(s) {
+		log.Infof(ctx, "Container %q is a besteffort pod. Skip PreStop.", c.ID())
+		return nil
+	}
+	if !isContainerRequestWholeCPU(c) {
+		log.Infof(ctx, "Container %q requests partial cpu(s). Skip PreStop", c.ID())
+		return nil
+	}
+
 	// enable the CPU load balancing for the container CPUs
 	if shouldCPULoadBalancingBeDisabled(s.Annotations()) {
 		if err := setCPUSLoadBalancing(c, true, schedDomainDir); err != nil {
-			return err
+			return errors.Wrap(err, "set CPU load balancing")
 		}
 	}
+	// enable the IRQ smp balancing for the container CPUs
+	if shouldIRQLoadBalancingBeDisabled(s.Annotations()) {
+		if err := setIRQLoadBalancing(c, true, irqSmpAffinityProcFile); err != nil {
+			return errors.Wrap(err, "set IRQ load balancing")
+		}
+	}
+	// no need to reverse the cgroup CPU CFS quota setting as the pod cgroup will be deleted anyway
 
 	return nil
 }
 
 func shouldCPULoadBalancingBeDisabled(annotations fields.Set) bool {
-	return annotations[annotationCPULoadBalancing] == "true"
+	return annotations[annotationCPULoadBalancing] == annotationTrue
+}
+
+func shouldCPUQuotaBeDisabled(annotations fields.Set) bool {
+	return annotations[annotationCPUQuota] == annotationTrue
+}
+
+func shouldIRQLoadBalancingBeDisabled(annotations fields.Set) bool {
+	return annotations[annotationIRQLoadBalancing] == annotationTrue
+}
+
+func isCgroupParentBurstable(s *sandbox.Sandbox) bool {
+	return strings.Contains(s.CgroupParent(), "burstable")
+}
+
+func isCgroupParentBestEffort(s *sandbox.Sandbox) bool {
+	return strings.Contains(s.CgroupParent(), "besteffort")
+}
+
+func isContainerRequestWholeCPU(c *oci.Container) bool {
+	return *(c.Spec().Linux.Resources.CPU.Shares)%1024 == 0
 }
 
 func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string) error {
@@ -65,7 +148,7 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string) 
 		lspec.Resources == nil ||
 		lspec.Resources.CPU == nil ||
 		lspec.Resources.CPU.Cpus == "" {
-		return fmt.Errorf("failed to find the container %q CPUs", c.ID())
+		return errors.Errorf("find container %s CPUs", c.ID())
 	}
 
 	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
@@ -107,6 +190,99 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string) 
 			return ioutil.WriteFile(path, []byte(newContent), 0o644)
 		})
 		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func setIRQLoadBalancing(c *oci.Container, enable bool, irqSmpAffinityFile string) error {
+	lspec := c.Spec().Linux
+	if lspec == nil ||
+		lspec.Resources == nil ||
+		lspec.Resources.CPU == nil ||
+		lspec.Resources.CPU.Cpus == "" {
+		return errors.Errorf("find container %s CPUs", c.ID())
+	}
+
+	content, err := ioutil.ReadFile(irqSmpAffinityFile)
+	if err != nil {
+		return err
+	}
+	currentIRQSMPSetting := strings.TrimSpace(string(content))
+	newIRQSMPSetting, newIRQBalanceSetting, err := UpdateIRQSmpAffinityMask(lspec.Resources.CPU.Cpus, currentIRQSMPSetting, enable)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(irqSmpAffinityFile, []byte(newIRQSMPSetting), 0o644); err != nil {
+		return err
+	}
+	if _, err := exec.LookPath("irqbalance"); err != nil {
+		// irqbalance is not installed, skip the rest; pod should still start, so return nil instead
+		logrus.Warnf("irqbalance binary not found: %v", err)
+		return nil
+	}
+	// run irqbalance in daemon mode, so this won't cause delay
+	cmd := exec.Command("irqbalance", "--oneshot")
+	additionalEnv := "IRQBALANCE_BANNED_CPUS=" + newIRQBalanceSetting
+	cmd.Env = append(os.Environ(), additionalEnv)
+	return cmd.Run()
+}
+
+func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool) error {
+	var rpath string
+	var err error
+	var cfsQuotaPath string
+	var parentCfsQuotaPath string
+	var cgroupManager cgmgr.CgroupManager
+
+	if strings.HasSuffix(parentDir, ".slice") {
+		// systemd fs
+		if cgroupManager, err = cgmgr.SetCgroupManager("systemd"); err != nil {
+			return nil
+		}
+		parentPath, err := systemd.ExpandSlice(parentDir)
+		if err != nil {
+			return err
+		}
+		parentCfsQuotaPath = filepath.Join(cpuMountPoint, parentPath, "cpu.cfs_quota_us")
+		if rpath, err = cgroupManager.ContainerCgroupAbsolutePath(parentDir, c.ID()); err != nil {
+			return err
+		}
+		cfsQuotaPath = filepath.Join(cpuMountPoint, rpath, "cpu.cfs_quota_us")
+	} else {
+		// cgroupfs
+		if cgroupManager, err = cgmgr.SetCgroupManager("cgroupfs"); err != nil {
+			return nil
+		}
+		parentCfsQuotaPath = filepath.Join(cpuMountPoint, parentDir, "cpu.cfs_quota_us")
+		if rpath, err = cgroupManager.ContainerCgroupAbsolutePath(parentDir, c.ID()); err != nil {
+			return err
+		}
+		cfsQuotaPath = filepath.Join(cpuMountPoint, rpath, "cpu.cfs_quota_us")
+	}
+
+	if _, err := os.Stat(cfsQuotaPath); err != nil {
+		return err
+	}
+	if _, err := os.Stat(parentCfsQuotaPath); err != nil {
+		return err
+	}
+
+	if enable {
+		// there should have no use case to get here, as the pod cgroup will be deleted when the pod end
+		if err := ioutil.WriteFile(cfsQuotaPath, []byte("0"), 0o644); err != nil {
+			return err
+		}
+		if err := ioutil.WriteFile(parentCfsQuotaPath, []byte("0"), 0o644); err != nil {
+			return err
+		}
+	} else {
+		if err := ioutil.WriteFile(cfsQuotaPath, []byte("-1"), 0o644); err != nil {
+			return err
+		}
+		if err := ioutil.WriteFile(parentCfsQuotaPath, []byte("-1"), 0o644); err != nil {
 			return err
 		}
 	}
--- a/internal/runtimehandlerhooks/utils.go
+++ b/internal/runtimehandlerhooks/utils.go
@@ -1,0 +1,123 @@
+package runtimehandlerhooks
+
+import (
+	"encoding/hex"
+	"fmt"
+	"strings"
+	"unicode"
+
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+)
+
+func isASCII(s string) bool {
+	for i := 0; i < len(s); i++ {
+		if s[i] > unicode.MaxASCII {
+			return false
+		}
+	}
+	return true
+}
+
+func cpuMaskByte(c int) byte {
+	return byte(1 << c)
+}
+
+func mapHexCharToByte(h string) ([]byte, error) {
+	l := len(h)
+	var hexin string
+	if l%2 != 0 {
+		// expect even number of chars
+		hexin = "0" + h
+	} else {
+		hexin = h
+	}
+
+	breversed, err := hex.DecodeString(hexin)
+	if err != nil {
+		return nil, err
+	}
+
+	l = len(breversed)
+	var barray []byte
+	var rindex int
+	for i := 0; i < l; i++ {
+		rindex = l - i - 1
+		barray = append(barray, breversed[rindex])
+	}
+	return barray, nil
+}
+
+func mapByteToHexChar(b []byte) string {
+	var breversed []byte
+	var rindex int
+	l := len(b)
+	// align it to 8 byte
+	if l%8 != 0 {
+		lfill := 8 - l%8
+		l += lfill
+		for i := 0; i < lfill; i++ {
+			b = append(b, byte(0))
+		}
+	}
+
+	for i := 0; i < l; i++ {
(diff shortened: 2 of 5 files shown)


haircommander:
/ok-to-test

Thanks for the PR! @giuseppe @kolyshkin PTAL

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -124,3 +183,71 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+func setCPUQuota(ctx context.Context, parentDir string, c *oci.Container, enable bool) error {
+	original := c.Spec().Linux.CgroupsPath
+	var rpath string
+	if strings.Contains(original, ":") {
+		// systemd managed, first replace the colon in "...slice:" with "/"
+		step1 := strings.Replace(original, ":", "/", 1)
+		// then replace colon in "crio:" with "-"
+		step2 := strings.Replace(step1, ":", "-", 1)
+		// lastly append ".scope"
+		rpath = step2 + ".scope"
```
haircommander, line 219: we could use https://godoc.org/github.com/opencontainers/runc/libcontainer/cgroups/systemd#ExpandSlice here

haircommander:
thanks for the PR @jianzzha 

/ok-to-test

I would like us to have unit tests and integration tests for this as well, please.

PTAL @giuseppe @kolyshkin for the cgroups code

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -64,6 +105,24 @@ func shouldCPULoadBalancingBeDisabled(annotations fields.Set) bool {
 	return value == "true"
 }
 
+func shouldCPUQuotaBeDisabled(annotations fields.Set) bool {
+	value, ok := annotations[annotationCPUQuota]
+	if !ok {
+		return false
+	}
+
+	return value == "true"
```
kolyshkin, line 114: You can avoid checking `ok` since for non-existing map elements the value returned is `""`.

kolyshkin: I.e. 
```go
return annotations[annotationCPUQuota] == "true"
```
is both correct and sufficient.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -64,6 +105,24 @@ func shouldCPULoadBalancingBeDisabled(annotations fields.Set) bool {
...
+}
+
+func shouldIRQLoadBalancingBeDisabled(annotations fields.Set) bool {
+	value, ok := annotations[annotationIRQLoadBalancing]
+	if !ok {
+		return false
+	}
+
+	return value == "true"
```
kolyshkin, line 123: ditto

kolyshkin: I see that it's just a copy-paste, so not really your code.

What you can do is add a first commit which simplifies the existing code (by removing the `ok` check), then add your own commit on top of it.

kolyshkin: Addressed some of these in https://github.com/cri-o/cri-o/pull/4026, please feel free to cherry-pick (or wait for that one to be merged).

jianzzha: Thanks for the #4026 cleanup. I will add some tests while waiting that one merged

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: @haircommander not sure if ExpandSlice will work. Here is the pattern of Spec().Linux.CgroupsPath, 
parentfolder:crio:containerid 
the real path is:
parentfolder/crio-containerid.slic

it doesn't look like ExpandSlice can do that conversion.

haircommander: I believe if you have a path:
parentfolder-subparentfolder:crio:containerid
it's going to be
/parentfolder/parentfolder-subparentfolder/crio/containerid.scope

so I believe this will fail without ExpandSlice. We're looking for the location of this cgroup on disk right?

jianzzha: yes. a real example is to convert:
kubepods-pod344c6fac_64b4_431c_a449_10ed09a492ba.slice:crio:813c5bd769ca6b8198acf1e927236d10807630246702d94165e23a31515a6a1a

into an absolute path:
/sys/fs/cgroup/cpu/kubepods.slice/kubepods-pod344c6fac_64b4_431c_a449_10ed09a492ba.slice/crio-813c5bd769ca6b8198acf1e927236d10807630246702d94165e23a31515a6a1a.scope

I do not know why crio put the two colons around "crio". is it due to some convention? my code does do the job but if ExpandSlice can do the majic I would love to use it. but looking at its source: https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/common.go#L53
1. it expect a input string which .slice suffix
2. it expect "-" as separator
so as its doc says, its job is to convert  "test-a-b.slice" to  "/test.slice/test-a.slice/test-a-b.slice". This is not the input I'm observing. I'm testing it on openshift, with cgroup v1 layout.

jianzzha: based on the input example provided above, ExpandSlice can take care of parent.slice/parent-subparent.slice portion, but not the crio-<containerid>.scope with the two colons around "crio". Let me know your thought on that

haircommander: ah yes, that seems more correct

also, if you look at `internal/config/cgmgr` there are some related functions (especially in systemd.go). perhaps we could utilize some code there?

jianzzha: Given the input I'm observing on the parentDir and Spec().Linux.CgroupsPath, here is one way to use ExpandSlice:
ExpandSlice(parentDir) + "/" + "crio-" + fmt.Sprintf("%d", c.ID() + ".scope"

THis will at least take care if the parent folder; though the "crio...scope" portion still looks hardcoded.
what do you think ?

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -124,3 +183,71 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+		"/sys/fs/cgroup/cpu/kubepods.slice/",
+		"/sys/fs/cgroup/cpu/kubepods/",
+		"/sys/fs/cgroup/",
+		"/sys/fs/cgroup/kubepods.slice/",
+		"/sys/fs/cgroup/kubepods/",
+	}
+	var cfsQuotaFile string
+	var parentCfsQuotaFile string
+	for _, prefix := range prefixes {
+		cfsQuotaFile = prefix + rpath + "/cpu.cfs_quota_us"
```
giuseppe, line 236: please use `filepath.Join` here

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -124,3 +183,71 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+			} else {
+				ioutil.WriteFile(cfsQuotaFile, []byte("-1\n"), 0644)
+				ioutil.WriteFile(parentCfsQuotaFile, []byte("-1\n"), 0644)
+			}
+			return nil
+		}
+	}
+	log.Infof(ctx, "Failed to find cpu.cfs_quota_us file for the container %q", c.ID())
+	return nil
+}
```
giuseppe, line 253: I don't think writing directly to the cgroup is the right thing to do when using systemd.  systemd is the owner for the cgroup.  We should either use the dbus API to change these settings, or I'd suggest using either runc/ libcontainer/cgroups or containers/podman/pkg/cgroups

internal/runtimehandlerhooks/utils.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	for i := 0; i < l; i++ {
+		rindex = l - i - 1
+		breversed = append(breversed, b[rindex])
+	}
+	return hex.EncodeToString(breversed)
+}
+
+// UpdateIRQSmpAffinityMask take input cpus that need to change irq affinit mask and
```
giuseppe, line 66: typo on affinit

internal/runtimehandlerhooks/utils.go
```diff
@@ -0,0 +1,103 @@
...
+// enabled or disable in the mask
+func UpdateIRQSmpAffinityMask(cpus string, current string, set bool) (string, error) {
+	podcpuset, err := cpuset.Parse(cpus)
+	if err != nil {
+		return cpus, err
+	}
+
+	// only ascii string supported
+	if isASCII(current) == false {
+		return cpus, fmt.Errorf("non ascii charactor detected: %s", current)
```
giuseppe, line 77: typo on charactor

internal/runtimehandlerhooks/utils.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	// remove ","; now each element is "0-9,a-f"
+	s := strings.Replace(current, ",", "", -1)
+
+	// the index 0 correspondes to the cpu 0-7
```
giuseppe, line 83: typo on correspondes

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -124,3 +183,71 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+	}
+	newIRQSMPSetting += "\n"
+	ioutil.WriteFile(irqSmpAffinityFile, []byte(newIRQSMPSetting), 0644)
+	return nil
+}
+
+func setCPUQuota(ctx context.Context, parentDir string, c *oci.Container, enable bool) error {
+	original := c.Spec().Linux.CgroupsPath
+	var rpath string
+	if strings.Contains(original, ":") {
```
cynepco3hahue, line 213: can we consider to use some existing library to work with cgroups, I am pretty sure I saw some usage of such library under the repository `github.com/opencontainers/runc/libcontainer/cgroups`

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -25,32 +25,73 @@ const (
...
+		}
+	}
 
 	return nil
 }
 
 func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-stop hook for the container %q", HighPerformance, c.ID())
+
+	if strings.Contains(s.CgroupParent(), "burstable") || strings.Contains(s.CgroupParent(), "besteffort") {
```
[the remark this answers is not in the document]
jianzzha: ack

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: ack

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -25,32 +25,73 @@ const (
...
+		log.Infof(ctx, "Container %q is not a guaranteed pod. Skip PreStart.", c.ID())
+		// not guranteed pod, do nothing
+		return nil
+	}
+	if *(c.Spec().Linux.Resources.CPU.Shares)%1024 != 0 {
+		log.Infof(ctx, "Container %q requests cpu share %d: not whole cpu(s). Skip PreStart", c.ID(), *(c.Spec().Linux.Resources.CPU.Shares))
+		// requested cpu not whole number, do nothing
+		return nil
+	}
```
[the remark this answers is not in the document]
jianzzha: there are 3 cases we try to avoid
1. best effort container
2. burstable container
3. guaranteed but request/limit cpu is not whole number of cpu. for example, when request/limit cpu is 0.5, the cpuset will be all cpus in the default cpu pool, so we can't use that cpuset; but cpu.share will be 512. 

cpu.share%1024 == 0 is an easy way to check against case 3.

jianzzha:
@haircommander  @cynepco3hahue ack on adding the test files.

jianzzha:
pushed a new commit with extra unit tests and addressed some of the comments above. Please review again.

cynepco3hahue:
@jianzzha You should signoff all commits to pass the DCO check.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -25,32 +26,89 @@ const (
...
 
 func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-start hook for the container %q", HighPerformance, c.ID())
+
+	if isCgroupParentBurstable(s) {
+		log.Infof(ctx, "Container %q is a burstable pod. Skip PreStart.", c.ID())
+		return nil
+	}
+	if isCgroupParentBestEffort(s) {
+		log.Infof(ctx, "Container %q is a besteffort pod. Skip PreStart.", c.ID())
```
TomSweeneyRedHat, line 47:
```suggestion
		log.Infof(ctx, "Container %q is a best effort pod. Skip PreStart.", c.ID())
```

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -25,32 +26,89 @@ const (
...
 
 func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-stop hook for the container %q", HighPerformance, c.ID())
+
+	if isCgroupParentBurstable(s) {
+		log.Infof(ctx, "Container %q is a burstable pod. Skip PreStart.", c.ID())
+		return nil
+	}
+	if isCgroupParentBestEffort(s) {
+		log.Infof(ctx, "Container %q is a besteffort pod. Skip PreStart.", c.ID())
```
TomSweeneyRedHat, line 92:
```suggestion
		log.Infof(ctx, "Container %q is a best effort pod. Skip PreStart.", c.ID())
```

TomSweeneyRedHat: oops, never mind, just double checked and besteffort is correct.

internal/runtimehandlerhooks/high_performance_hooks.go
TomSweeneyRedHat: ignore

internal/runtimehandlerhooks/utils.go
jianzzha: ack

internal/runtimehandlerhooks/utils.go
jianzzha: ack

internal/runtimehandlerhooks/utils.go
jianzzha: ack

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -25,54 +26,127 @@ const (
...
+	}
 
 	return nil
 }
 
 func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-stop hook for the container %q", HighPerformance, c.ID())
+
+	if isCgroupParentBurstable(s) {
+		log.Infof(ctx, "Container %q is a burstable pod. Skip PreStart.", c.ID())
```
cbandy, line 88: Should these log messages in `PreStop()` say "PreStop"?

```suggestion
		log.Infof(ctx, "Container %q is a burstable pod. Skip PreStop.", c.ID())
```

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -124,3 +191,74 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+		// CgroupsPath in this pattern parent-subparent:crio:containerID
+		parts := strings.SplitN(original, ":", 3)
+		if len(parts) < 3 {
+			return fmt.Errorf("invalid cgroup entry: must contain at least two colons: %v", original)
+		}
+		// append ".scope"
+		rpath = parts[1] + "-" + parts[2] + ".scope"
+		parentPath, err := systemd.ExpandSlice(parts[0])
+		if err != nil {
+			return nil
```
cbandy, line 236: Should this return `err`?

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: good catch

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: ack

haircommander:
don't worry about ami_rhel @jianzzha, it's for rebuilding the VM images

jianzzha:
All done. Please take a final look. Thanks.

haircommander:
I don't super understand what this is doing, but the code looks fine to me
@kolyshkin @giuseppe @mrunalp @umohnani8 @saschagrunert PTAL
@jianzzha can you fix the release note? I do believe this is a user facing change.
also
/kind feature

jianzzha:
@haircommander It will take a bit of writing to explain why we want to disable IRQ and cpu cfs throttling on a pod's CPU(s) :) 

Basically this is for NFV use cases, for example for a dpdk application running inside a pod. Kubenetes provides static cpu policy to allocate exclusive cpu(s) for a pod that requires whole cpu(s). So we can make sure no other pods compete for the cpu resources allocated to a dpdk application. This, however, is not good enough. Sometime, kernel work, for example the interrupt handling on the cpu where the dpdk application is running can cause jitter which leads to packet drop. So we want to offer a way to optionally turn off the interrupt for a pod that doesn't want to be interrupted.

Turning off cpu cfs is for the same idea, but more for a safeguard - we have seen in the past that cpu cfs can make inaccurate accounting which leads to throttling and causes packet drop. See https://github.com/kubernetes/kubernetes/issues/67577 for some details. Admittedly that the inaccurate accounting can be a kernel issue, The thing is: when an entire cpu is exclusively allocated to a pod, there is no need for the cfs throttling on that cpu anyway - so we can turn it off as an option instead of waiting for a kernel fix. 

Hope this explains the purpose behind this PR.

haircommander:
thanks for the explanation @jianzzha !

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -25,38 +27,116 @@ const (
...
 )
 
 // HighPerformanceHooks used to run additional hooks that will configure a system for the latency sensitive workloads
 type HighPerformanceHooks struct{}
 
 func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-start hook for the container %q", HighPerformance, c.ID())
+
+	if isCgroupParentBurstable(s) {
+		log.Infof(ctx, "Container %q is a burstable pod. Skip PreStart.", c.ID())
```
saschagrunert, line 45: Do we wanna use `Debugf` messages here and down below instead?

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -25,38 +27,116 @@ const (
...
+	if !isContainerRequestWholeCPU(c) {
+		log.Infof(ctx, "Container %q requests partial cpu(s). Skip PreStart", c.ID())
+		return nil
+	}
+
 	// disable the CPU load balancing for the container CPUs
 	if shouldCPULoadBalancingBeDisabled(s.Annotations()) {
+		log.Infof(ctx, "Disable cpu load balancing for container %q", c.ID())
 		if err := setCPUSLoadBalancing(c, false, schedDomainDir); err != nil {
 			return err
```
saschagrunert, line 61: Let's wrap those errors to add more context:
```suggestion
			return errors.Wrap(err, "set CPU load balancing")
```

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -113,3 +193,89 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
 	return nil
 }
+
+func setIRQLoadBalancing(c *oci.Container, enable bool, irqSmpAffinityFile string) error {
+	lspec := c.Spec().Linux
+	if lspec == nil ||
+		lspec.Resources == nil ||
+		lspec.Resources.CPU == nil ||
+		lspec.Resources.CPU.Cpus == "" {
+		return fmt.Errorf("failed to find the container %q CPUs", c.ID())
```
saschagrunert, line 203:
```suggestion
		return errors.Errorf("find container %s CPUs", c.ID())
```

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -113,3 +193,89 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+	}
+
+	content, err := ioutil.ReadFile(irqSmpAffinityFile)
+	if err != nil {
+		return err
+	}
+	currentIRQSMPSetting := strings.Trim(string(content), "\n")
```
saschagrunert, line 210:
```suggestion
	currentIRQSMPSetting := strings.TrimSpace(string(content))
```

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -113,3 +193,89 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+}
+
+func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool) error {
+	var rpath string
+	var err error
+	var cfsQuotaPath string
+	var parentCfsQuotaPath string
+	var cgroupManager cgmgr.CgroupManager
+
+	if strings.Contains(parentDir, ".slice") {
```
saschagrunert, line 230:
```suggestion
	if strings.HasSuffic(parentDir, ".slice") {
```

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -113,3 +193,89 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+		}
+		parentPath, err := systemd.ExpandSlice(parentDir)
+		if err != nil {
+			return err
+		}
+		parentCfsQuotaPath = filepath.Join(cpuMountPoint, parentPath, "cpu.cfs_quota_us")
+		if rpath, err = cgroupManager.ContainerCgroupAbsolutePath(parentDir, c.ID()); err != nil {
+			return err
+		}
+		cfsQuotaPath = filepath.Join(cpuMountPoint, rpath, "cpu.cfs_quota_us")
```
saschagrunert, line 243: not blocking: I recommend to refactor this function (and the cgroupfs one) into own ones to be able to test them and reduce the complexity.

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: I like it!

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: ack

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: ack

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: ack

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: This one is on the boundary - the information is not verbose enough to make it to debug level so I feel like to keep it at the info level.

internal/runtimehandlerhooks/high_performance_hooks.go
jianzzha: I'm working on some cgroup common utils outside this PR.  Once that's done I will do the refactor using the common utils.

haircommander:
LGTM, waiting on final ack from @giuseppe or @kolyshkin

jianzzha:
put this PR on hold to investigate using irqbalance service in stead.

jianzzha:
@haircommander not sure why e2e-aws failed, I will do a rebase to master branch to see if it helps

haircommander:
/retest
/lgtm

thanks @jianzzha

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -113,3 +196,96 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
+	if err := ioutil.WriteFile(irqSmpAffinityFile, []byte(newIRQSMPSetting), 0o644); err != nil {
+		return err
+	}
+	if _, err := exec.LookPath("irqbalance"); err != nil {
+		// irqbalance is not installed, skip the rest; pod should still start, so return nil instead
+		logrus.Warnf("irqbalance binary not found: %v", err)
+		return nil
+	}
+	// run irqbalance in daemon mode, so this won't cause delay
+	cmd := exec.Command("irqbalance", "--oneshot")
```
pperiyasamy, line 227: If the irqbalance service already exists, then we may have to update /etc/sysconfig/irqbalance config file with IRQBALANCE_BANNED_CPUS and restart irqbalance service. we could run irqbalance --oneshot command only if the service is not present (This is what i did it here https://github.com/pperiyasamy/irq-smp-balance/blob/main/pkg/irq/util.go#L94). 
Shouldn't it be done this way ?

cynepco3hahue: yes, it can be nice, can you please open the PR and we will discuss it

jianzzha: yes, and we need to discuss whether/how to recover  /etc/sysconfig/irqbalance to its default if the computer node rebooted

JanScheurich: One additional challenge that I see with the approach to reconfigure the irqbalance service is its platform dependency. We have seen that Linux distributions like SLES, Ubuntu, RHEL all have slightly different approaches for configuring and managing the daemon, and even within one distribution the way can change between major releases.

Can we find a solution that will work on all relevant platforms? Perhaps add parameters to the cri-o config file to tell, which file to update and how to restart the service?

If not, can the solution be split into a generic part inside cri-o that manages a a file on the host with the wanted banned CPUs, and another platform-specific daemon that reconfigures the host's irqbalance service accordingly?

pperiyasamy: Yes, The irqbalance config would present at /etc/sysconfig/ directory in SLES, CentOS platforms whereas in Ubuntu it's present in /etc/default/ directory. Hope this file is just a source file which takes in the format of  `IRQBALANCE_BANNED_CPUS=<value>`. Of course we could pass the config file path using a new RuntimeConfig parameter and made it available to runtime_handler_hooks through Server.config. 
we could recover the irqbalance config at crio start time and banned mask derived from /proc/irq/default_smp_affinitity. 
Could we take this approach ?

PR #4026, opened 2020-07-27
Various nits

kolyshkin:
/kind cleanup

Optimize and/or improve readability of some code I came across. Please see individual commits for details.

```release-note
NONE
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks.go
@@ -56,12 +56,7 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s 
 }
 
 func shouldCPULoadBalancingBeDisabled(annotations fields.Set) bool {
-	value, ok := annotations[annotationCPULoadBalancing]
-	if !ok {
-		return false
-	}
-
-	return value == "true"
+	return annotations[annotationCPULoadBalancing] == "true"
 }
 
 func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string) error {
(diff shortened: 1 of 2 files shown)


kolyshkin:
CI failures do not seem to be related.

/retest

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -83,15 +79,9 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
 			if err != nil {
 				return err
 			}
-
-			if path == cpuSchedDomainDir {
-				return nil
-			}
-
-			if !strings.Contains(path, "flags") {
+			if info.Name() != "flags" {
```
TomSweeneyRedHat, line 82: Is this really equivalent?  Regardless, it might be good to still call out the cpuSchedDomainDir check.  However, if @haircommander @mrunalp or @saschagrunert are good with this change, then I'll trust their judgment.

haircommander: this change looks good, info.Name() is the final element of the path

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -79,15 +79,9 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
 			if err != nil {
 				return err
 			}
-
-			if path == cpuSchedDomainDir {
```
haircommander, line 87: how is this redundant?

haircommander:
one question, otherwise LGTM

internal/runtimehandlerhooks/high_performance_hooks.go
cynepco3hahue: the next `if info.Name() != "flags" ` will catch it anyway

kolyshkin:
updated the `filepath.Walk()` check to exclude non-regular files. It's not strictly required since we won't have a directory named 'flags' in there, but should speed things up a tiny bit since file mode flags check should be faster than file name string comparison.

PR #4033, opened 2020-07-30
Make filter container list be able to filter short pod IDs

saschagrunert:
#### What type of PR is this?

/kind feature

The `PodSandboxID` can now be truncated because we're just checking for its prefix.
Refers to https://github.com/kubernetes-sigs/cri-tools/pull/643

None

```release-note
- Added behavior to allow filtering by a partial Pod Sandbox ID
```

### DIFF
--- a/server/container_list.go
+++ b/server/container_list.go
@@ -1,6 +1,8 @@
 package server
 
 import (
+	"strings"
+
 	"github.com/cri-o/cri-o/internal/log"
 	oci "github.com/cri-o/cri-o/internal/oci"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
and happy green test buttons

server/container_list.go
```diff
@@ -35,28 +37,29 @@ func (s *Server) filterContainerList(ctx context.Context, filter *pb.ContainerFi
...
 			default:
-				return []*oci.Container{}
+				return nil
 			}
 		}
 	} else if filter.PodSandboxId != "" {
-		pod := s.ContainerServer.GetSandbox(filter.PodSandboxId)
-		if pod == nil {
-			return []*oci.Container{}
+		for _, sb := range s.ListSandboxes() {
```
mikebrow, line 55: suggest `sandboxID, err := s.PodIDIndex().Get(filter.PodSandboxId)` to use the index vs scan

haircommander:
/hold

thinking about this more, I agree with @mikebrow , that seems more idiomatic for how we handle short ids

server/container_list.go
saschagrunert: Yes makes sense, thank you for the hint :)

saschagrunert:
> thinking about this more, I agree with @mikebrow , that seems more idiomatic for how we handle short ids

Changed the implementation as suggested :+1:

PR #4040, opened 2020-08-02
Parameterize strip binary in pinns/Makefile for cross-compilation

bluestealth:
#### What type of PR is this?
/kind cleanup

Allows cross compile tool-chains to select the correct `strip` binary to use.
Without this change it will attempt to use the host tool-chain instead of the target tool-chain, which will fail if the underlying architecture is not the same.

```release-note
None
```

### DIFF
--- a/pinns/Makefile
+++ b/pinns/Makefile
@@ -1,6 +1,7 @@
 src = $(wildcard *.c)
 obj = $(src:.c=.o)
 
+STRIP ?= strip
 override LIBS +=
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests

haircommander:
thanks for the PR @bluestealth 

(lint failure is not the fault of this PR)

/retest
/lgtm

PR #4052, opened 2020-08-06
config: create hooks dir if not present

haircommander:
#### What type of PR is this?

/kind design

We've changed the handling of hooks_dir quite a bit, hopefully this is the last of such changes.

There are cases where directories are not present on the host, but admins would desire some entity put hooks into said directory. The primary use case I can think of is a temporary directory in /run that allows hooks to be wiped on reboot. Not really a use case worth packaging for everyone, but instead something cri-o should handle  

```release-note
hooks_dir entries are now created if they don't exist
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -780,15 +780,25 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 			return errors.Wrap(err, "invalid registries")
 		}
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

haircommander:
PTAL @umohnani8 @mrunalp

PR #4056, opened 2020-08-07
Refactor AppArmor logic into new container interface

saschagrunert:
#### What type of PR is this?

/kind cleanup

This also implies that we move the spec generator into the config
interface, too. Benefit is now that we're able to test the AppArmor
logic which was previously part of `server/container_create_linux.go`.

None

None

```release-note
None
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -7,7 +7,10 @@ import (
 	"strings"
 
 	"github.com/containers/storage/pkg/stringid"
+	"github.com/cri-o/cri-o/internal/config/apparmor"
+	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/utils"
+	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
LGTM
but would like a couple other head nods

haircommander:
I'd say we should unify the spec implementation between this and https://github.com/cri-o/cri-o/pull/3960
@wgahnagl @saschagrunert

haircommander:
I would personally vote we get https://github.com/cri-o/cri-o/pull/3960 in and @saschagrunert rebases this on top of it

saschagrunert:
> I would personally vote we get #3960 in and @saschagrunert rebases this on top of it

Sure, sounds good!

saschagrunert:
It's green! @haircommander may I get a /lgtm again?

server/container_create_linux.go
```diff
@@ -215,10 +215,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 	}
 
 	// creates a spec Generator with the default spec.
```
haircommander, line 217: nit (fine as follow up) we don't actually create the spec here, so the comment is out of date

pkg/container/container.go
```diff
@@ -276,3 +290,42 @@ func (c *container) SelinuxLabel(sboxLabel string) ([]string, error) {
...
+}
+
+// SpecGen returns the internal config spec generator.
+func (c *container) SpecGen() *generate.Generator {
+	return c.specGen
+}
+
+// ApplyAppArmor takes the AppArmor config and the profile name as string and
+// applies it to the internal config spec generator.
+func (c *container) ApplyAppArmor(appArmorConfig *apparmor.Config, profileName string) error {
```
haircommander, line 311: I've never seen it capatalized as AppArmor. Is that idiomatic? I think I prefer Apparmor

saschagrunert: Indeed it is `AppArmor` (see the never lying [Wikipedia article](https://en.wikipedia.org/wiki/AppArmor). I also changed that in libraries like containers/common recently (https://github.com/containers/common/pull/260/files)

server/container_create_linux.go
saschagrunert: Fixed the comment :+1:

pkg/container/container.go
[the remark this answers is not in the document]
saschagrunert: Makes sense, fixed that :+1:

PR #4057, opened 2020-08-07
crio: perform pulls in a new cgroup

giuseppe:
when cri-o is re-exec'ed create a new systemd scope to run it.

All pulled files are charged and that are still in memory, are charged
to the cgroup that first opened them.  Without the transient cgroup,
they would be charged to the CRI-O process even if they are kept in
memory by container processes.

Using a transient cgroup prevents that, as once the cgroup is
destroyed these files are charged to the cgroup that is using them.

#### What type of PR is this?

/kind design

prevents files to be charged to the CRI-O cgroup

```release-note
when running under systemd, image pulls happen in a new cgroup
```


cmd/crio/main.go
```diff
@@ -114,6 +119,7 @@ func main() {
 	// This must be done before reexec.Init()
 	ioutils.SetDefaultOptions(ioutils.AtomicFileWriterOptions{NoSync: true})
 
+	maybeMoveToCgroup()
```
haircommander, line 122: I'm confused. this reexec happens in the main crio process. Doesn't that mean we're just moving the whole crio process to a new cgroup?

haircommander: (I should probably learn how reexec.Init() works..)

giuseppe: no :) when we do a pull, we re-exec the crio process and the os.Args[0] is set to a special value.  containers/storage does it for running the extraction in a chroot environment (the logic is in `github.com/containers/storage/pkg/chrootarchive/archive_unix.go`).
The current implementation is a hack, to avoid changes in containers/storage and get an idea of what improvements we can expect.

mrunalp:
@vrothberg @mtrmac ptal.

internal/storage/image.go
```diff
@@ -407,9 +409,32 @@ type copyImageArgs struct {
...
+	if os.Getenv("INVOCATION_ID") == "" {
+		return nil
+	}
+
+	slice := "system.slice"
+	if rootless.IsRootless() {
+		slice = "user.slice"
+	}
+
+	unitName := fmt.Sprintf("crio-pull-image-%d.scope", os.Getpid())
```
kolyshkin, line 425: Why the new unique unit name for every copyImage? I mean, can we just use `crio-pull-image.scope` for all of them?

giuseppe: The goal is to run each pull in a separate cgroup, so we could apply limits if needed in future.

I've not looked if it is possible, but can we join an existing systemd scope?

kolyshkin: > I've not looked if it is possible, but can we join an existing systemd scope?

Not sure about systemd, but we can join the cgroup. That will be a kludge though :(

In general I think it make more sense for all pulls to be run within one single scope. That might require a different architecture though :-\

mrunalp: Eventually, we want to charge image pulls to a pod cgroup once k8s upstream has the plumbing to give us a pod cgroup for an image pull.

TomSweeneyRedHat:
LGTM
assuming the tests get happy

giuseppe:
@mrunalp added a flag to enable the new feature

docs/crio.conf.5.md
```diff
@@ -112,6 +112,9 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
   Additional environment variables to set for all the containers. These are overridden if set in the container image spec or in
 the container runtime configuration.
 
+**pull-in-a-cgroup**=false
+  If true, image pulls will happen in a new cgroup. It is currently supported only with the systemd cgroup manager.
```
haircommander, line 116: I feel like this should be in the `[crio.image]` table...
also, can we rename to `separate-pull-cgroup` or something like that?

giuseppe:
rebased

docs/crio.conf.5.md
```diff
@@ -279,6 +279,8 @@ CRI-O reads its configured registries defaults from the system wide containers-r
 **big_files_temporary_dir**=""
   Path to the temporary directory to use for storing big files, used to store image blobs and data streams related to containers image management.
 
+**separate-pull-cgroup**=false
+  If true, image pulls will happen in a new cgroup. It is currently supported only with the systemd cgroup manager.
```
TomSweeneyRedHat, line 283: Can you combine the messaging in this man and crio.8.md?  Perhaps in both places:
```suggestion
  If true, image pulls will happen in a new cgroup. It is currently supported only with the systemd cgroup manager.  (default: false)
```
Just based on this, there's no discussion about why one might want to use a new cgroup.  Is it covered elsewhere in this man page?  If not, we should add it either here or somewhere on the page.

giuseppe: we are not specifying any default value in `crio.8.md` so wouldn't it be better to be conformant with the rest of the file?

giuseppe:
tests are green now

giuseppe:
all green!

docs/crio.8.md
```diff
@@ -280,6 +281,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--selinux**: Enable selinux support (default: false)
 
+**--separate-pull-cgroup**: [EXPERIMENTAL] Pull in new cgroup (default: false)
```
mrunalp, line 284: I think instead of boolean we could make this a string. So, we have:
1. empty string: current behavior
2. POD: pull in pod cgroup
3. some.slice: Perform all pulls in that slice

server/image_pull.go
```diff
@@ -208,12 +214,21 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
...
+		if !s.config.SeparatePullCgroup {
+			cgroup = ""
+		}
+		_, err = s.StorageImageServer().PullImage(s.config.SystemContext, img, &storage.ImageCopyOptions{
 			SourceCtx:        &sourceCtx,
 			DestinationCtx:   s.config.SystemContext,
 			OciDecryptConfig: decryptConfig,
 			ProgressInterval: time.Second,
 			Progress:         progress,
+			UseNewCgroup:     s.config.SeparatePullCgroup,
```
mrunalp, line 230: Maybe group these two as cgroup options?

internal/storage/image.go
```diff
@@ -365,44 +388,247 @@ func (svc *imageService) prepareReference(inputSystemContext *types.SystemContex
...
+	if cgroup != "" {
+		if !strings.Contains(cgroup, ".slice") {
+			return fmt.Errorf("invalid systemd cgroup %q", cgroup)
+		}
+		slice = strings.ReplaceAll(strings.Trim(cgroup, "/"), "/", ":")
+	}
+
+	unitName := fmt.Sprintf("crio-pull-image-%d.scope", os.Getpid())
+
+	return utils.RunUnderSystemdScope(os.Getpid(), slice, unitName)
```
mrunalp, line 429: We probably need the `CollectMode` set to `inactive-or-failed` for the pull so they don't leak on failures.

mrunalp: Opened https://github.com/cri-o/cri-o/pull/4248 to support that.

docs/crio.conf.5.md
```diff
@@ -279,6 +279,8 @@ CRI-O reads its configured registries defaults from the system wide containers-r
 **big_files_temporary_dir**=""
   Path to the temporary directory to use for storing big files, used to store image blobs and data streams related to containers image management.
 
+**separate_pull_cgroup**=""
+  [EXPERIMENTAL] If its value is set, then images are pulled into the specified cgroup.  If its value is set to "pod", then the pod cgroup is used.  It is currently supported only with the systemd cgroup manager.
```
TomSweeneyRedHat, line 283: nit
```suggestion
  [EXPERIMENTAL] If its value is set, then images are pulled into the specified cgroup.  If its value is set to "pod", then the pod's cgroup is used.  It is currently supported only with the systemd cgroup manager.
```

docs/crio.conf.5.md
TomSweeneyRedHat: @giuseppe being conformant with crio.8.md and leaving the default out there is fine by me.  However, it would be best to list or not list defaults in all man pages and be consistent.  But that's a task for another day.

internal/storage/image.go
```diff
@@ -365,44 +397,258 @@ func (svc *imageService) prepareReference(inputSystemContext *types.SystemContex
...
+	ParentCgroup   string
+	SystemContext  *types.SystemContext
+	Options        *ImageCopyOptions
+	HasCollectMode bool
+
+	StoreOptions storage.StoreOptions
+}
+
+// moveSelfToCgroup moves the current process to a new transient cgroup.
+func moveSelfToCgroup(cgroup string, hasCollectMode bool) error {
```
haircommander, line 425: we seem to be hard coding assumptions about the node using systemd. Will this break users who aren't (if there are any?)

giuseppe: yes, currently only systemd is supported.  If CRI-O is running without systemd, then we wouldn't even get this far but fail earlier.
cgroupfs support is missing, we'll have to take care of cleaning up the cgroup by ourselves.

PR #1159, opened 2017-11-16
Replace crioctl with crictl

weiwei04:
**- What I did**

replace crioctl with crictl in e2e

**- How I did it**

vim

**- How to verify it**

make verify

make integration

**- Description for the changelog**

replace crioctl with crictl in e2e

PR #1156, opened 2017-11-15
replace crioctl with crictl in e2e

weiwei04:
Signed-off-by: Wei Wei <weiwei.inf@gmail.com>

**- What I did**

replace crioctl with crictl in e2e

**- How I did it**

vim

**- How to verify it**

make verify

make integration

**- Description for the changelog**

replace crioctl with crictl in e2e

PR #4060, opened 2020-08-10
test/*bats: fix shellcheck SC2076

kolyshkin:
/kind cleanup

```release-note
NONE
```

### 1. test/*bats: fix shellcheck SC2076

This fixes shellcheck warning SC2076, for example

```
In default_mounts.bats line 90:
    [[ "$output" =~ "test.txt" ]]
                    ^--------^ SC2076: Don't quote right-hand side of =~, it'll match literally rather than as a regex.
```

So, operator `~=` does a comparison against a regular expression, and yet
in all the places we are actually using it to search for a substring, not a regex.

Use the substring syntax (i.e. `[[ "$output" == *"substring"* ]]`) instead
to be more clear about what we're doing.

This commit is courtesy of
```bash
for f in *.bats; do sed -i -e 's/ =\~ "\(.*\)" ]]/ == *"\1"* ]]/g' $f; done
```

and a few manual fixes on top of it.

------

### 2. test/*.bats: fix checks that id is not present

> [[ "$output" != *"$ctr3_id"* ]]
> [[ "$output" != "$pod3_id" ]]

These two checks meant to check that the ID is not present in the
output. They don't work as they perform string comparison, while
we need to search for a substring.
    
Fix by adding * around the substring.

Fixes: b0b6611bdf (#1156)
Fixes: 25dfde9044 (#1159)

### DIFF
--- a/test/apparmor.bats
+++ b/test/apparmor.bats
@@ -57,7 +57,7 @@ load_a_specific_apparmor_profile_as_default_apparmor_and_run_a_container_with_it
     ctr_id="$output"
     run crictl exec --sync "$ctr_id" touch test.txt
     echo "$output"
-    [[ "$output" =~ "Permission denied" ]]
+    [[ "$output" == *"Permission denied"* ]]
 
     remove_apparmor_profile "$APPARMOR_TEST_PROFILE_PATH"
     cleanup_test
@@ -83,7 +83,7 @@ load_default_apparmor_profile_and_run_a_container_with_another_apparmor_profile(
     ctr_id="$output"
     run crictl exec --sync "$ctr_id" touch test.txt
     echo "$output"
-    [[ "$output" =~ "Permission denied" ]]
+    [[ "$output" == *"Permission denied"* ]]
 
     remove_apparmor_profile "$APPARMOR_TEST_PROFILE_PATH"
     cleanup_test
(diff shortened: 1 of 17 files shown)


mrunalp:
Test failures look related to the changes - https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/4060/test_pull_request_crio_integration_rhel/15866/artifacts/testout.txt

kolyshkin:
> Test failures look related to the changes - https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/4060/test_pull_request_crio_integration_rhel/15866/artifacts/testout.txt

Thanks for looking into it! My bad, I have somehow accidentally replaced `!=` with `==` (before writing a script to do that).

But the check was not working before. The way it was written is

```bash
[[ "$output" != "$ctr3_id" ]]
```

and it was meant to check that output does not contain a substring `$ctr3_id`, while it fact it was checking that output was not equal to `$ctrl3_id`.

Fix (of the original code):
```diff
-[[ "$output" != "$ctr3_id" ]]
+[[ "$output" != *"$ctr3_id"* ]]
```

I took a look and found a few more logical bugs like this.

Fixed the first commit, added the second one fixing such != checks.

mrunalp:
e2e-aws failures are unrelated.

PR #2943, opened 2019-11-04
run e2e tests using crun

giuseppe:
Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #4061, opened 2020-08-10
test/ctr.bats: fix test check

kolyshkin:
First, the code

        [ "$status" -ne 0 ] || [ "$output" =~ "Exit code: 1" ]

is not working since one can't use `=~` operator inside `[..]`
(it is only working inside `[[..]]`).

Here's an example:
```console
        $ [ "aaa bbb" =~ "aaa" ]; echo $?
        bash: [: =~: binary operator expected
        2
```

Since the test never failed, I guess $status was always non-zero,
so the second part was never evaluated, thus it is not needed.

This was found by shellcheck:

```
In ctr.bats line 1008:
        [ "$status" -ne 0 ] || [ "$output" =~ "Exit code: 1" ]
                                           ^-- SC2074: Can't use =~ in [ ]. Use [[..]] instead.
```

Second, the cleanup/stop stuff added by the commit is not needed,
since `teardown()` is executed after each test, and it contains
a call to `cleanup_test()` which calls the same 3 functions that
were added.

This reverts commit 2649ebb1a5d39208f8d2d68303e42a81048f4bdf
(from PR https://github.com/cri-o/cri-o/pull/2943/)

/kind bug

```release-note
NONE
```

@giuseppe PTAL

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -1005,10 +1005,7 @@ function wait_until_exit() {
 	[ "$status" -eq 0 ]
 	run crictl exec --sync "$ctr_id" doesnotexist
 	echo "$output"
-	[ "$status" -ne 0 ] || [ "$output" =~ "Exit code: 1" ]
-	cleanup_ctrs
-	cleanup_pods
-	stop_crio
+	[ "$status" -ne 0 ]
 }
 
 @test "ctr execsync exit code" {


kolyshkin:
Or maybe this second check is non needed at all?

mrunalp:
Looks like @giuseppe added that change for crun https://github.com/cri-o/cri-o/commit/2649ebb1a5d39208f8d2d68303e42a81048f4bdf

kolyshkin:
> Looks like @giuseppe added that change for crun 2649ebb

Yeah, I referred to that one in the commit message.

The problem is, this second check never worked:

```console
bash $ [ "aaa bbb" =~ "aaa" ]; echo $?
bash: [: =~: binary operator expected
2
```

and since the whole line (`[ "$status" -ne 0 ] || [ "$output" =~ "Exit code: 1" ]`) never failed it means the second check was never ran.

kolyshkin:
Re-did the commit to just revert 2649ebb1.

mrunalp:
This looks fine to me but will wait for @giuseppe to ack.

PR #4064, opened 2020-08-10
test/stats.bats: fixes

kolyshkin:
/kind bug

* test/stats.bats: fix typo in a variable name
* test/stats.bats: improve test case
* test/stats.bats: fix/improve container stats test case

See individual commits for more details.

~~Currently blocked by #4080 and #4068~~.

```release-notes
NONE
```

@wgahnagl @haircommander PTAL

PR #4068, opened 2020-08-12
oci_linux: fix working set calculation

kolyshkin:
/kind bug

Commit 512fdb2f9f49 (PR #3115) mistakenly used the value of total_inactive_file
from the top-level cgroup, thus the working set value was either
wrong (too low) or invalid (negative), for example:

> Unable to account working set stats: total_inactive_file (1572753409)
> memory usage (585728)" file="oci/oci_linux.go:93"

We need to use total_inactive_file and memory.usage_in_bytes from
the same cgroup, otherwise it does not make any sense.

While at it
 - promote the above message from debug to warning;
 - optimize getTotalInactiveFile() by using HasPrefix()
   rather than Contains().

None

```release-note
Fix working set calculation
```

### DIFF
--- a/internal/oci/oci_linux.go
+++ b/internal/oci/oci_linux.go
@@ -6,6 +6,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


kolyshkin:
@saschagrunert @TomSweeneyRedHat @giuseppe @haircommander @mrunalp PTAL

kolyshkin:
Found while fixing test/stats.bats in PR #4064 (see https://github.com/cri-o/cri-o/pull/4064#issuecomment-672330148), going to test it there.

haircommander:
/retest

LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #4071, opened 2020-08-12
selinux: override only specified values

giuseppe:
use the sandbox selinux label options from the pod when if they are
not overriden in the configuration.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1868122

#### What type of PR is this?

/kind bug
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1868122

```release-note
NONE
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -267,12 +267,27 @@ func (c *container) ReadOnly(serverIsReadOnly bool) bool {
 // it takes the sandbox's label, which it falls back upon
 func (c *container) SelinuxLabel(sboxLabel string) ([]string, error) {
 	selinuxConfig := c.config.GetLinux().GetSecurityContext().GetSelinuxOptions()
-	if selinuxConfig != nil {
-		return utils.GetLabelOptions(selinuxConfig), nil
-	}
+
+	labels := map[string]string{}
+
 	labelOptions, err := label.DupSecOpt(sboxLabel)
 	if err != nil {
 		return nil, err
 	}
-	return labelOptions, nil
+	for _, r := range labelOptions {
+		k := strings.Split(r, ":")[0]
+		labels[k] = r
+	}
+
+	if selinuxConfig != nil {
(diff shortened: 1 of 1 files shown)


pkg/container/container.go
```diff
@@ -267,12 +267,27 @@ func (c *container) ReadOnly(serverIsReadOnly bool) bool {
...
+
+	if selinuxConfig != nil {
+		for _, r := range utils.GetLabelOptions(selinuxConfig) {
+			k := strings.Split(r, ":")[0]
+			labels[k] = r
+		}
+	}
+	var ret []string
+	for _, v := range labels {
+		ret = append(labelOptions, v)
```
jean-edouard, line 290: Shouldn't this be `ret = append(ret, v)`?

giuseppe: yes obviously :)  Thanks for catching it!

umohnani8:
LGTM
/retest

pkg/container/container.go
```diff
@@ -267,12 +267,27 @@ func (c *container) ReadOnly(serverIsReadOnly bool) bool {
...
+		labels[k] = r
+	}
+
+	if selinuxConfig != nil {
+		for _, r := range utils.GetLabelOptions(selinuxConfig) {
+			k := strings.Split(r, ":")[0]
+			labels[k] = r
+		}
+	}
+	var ret []string
```
saschagrunert, line 288:
```suggestion
	ret := []string{}
```

TomSweeneyRedHat: I'm not arguing about the change but trying to learn.  Is this any better other than style?

TomSweeneyRedHat:
LGTM

pkg/container/container.go
saschagrunert: For me it does not matter but the linter complains about “considering a preallocation”. :)

giuseppe: fixed

haircommander:
/retest
LGTM

PR #368, opened 2017-02-22
Run without seccomp

apilloud:
This fixes issues running without seccomp compiled in and issues running with seccomp disabled. Before this change, the demo pod in sandbox_config fails to start with runc returning the following error:
`seccomp: config provided but seccomp not supported`

PR #211, opened 2016-11-23
Add support for Seccomp

runcom:
Relates to https://github.com/kubernetes-incubator/cri-o/issues/202

This PR adds support for seccomp in ocid. Right now we can only use either the `unconfined` profile or the `runtime/default` profile because per container profile on the node is an open issue upstream in k8s (see https://issues.k8s.io/36997).

This PR takes the docker's default seccomp profile for now. When k8s implements their own seccomp profile types we'll move to that of course.
Upstream opencontainers/runtime-spec is also going to modify the seccomp profile type to be more usable (there's an email thread going on right now and possible a PR soonish).

The linked issue at the top is going to remain open because we're still missing the `localhost/...` profile. I'll fix that once we figure that out and runtimes in CRI knows how to get a profile from the node they're running into.

PR #4078, opened 2020-08-13
Remove duplicate check for enabled seccomp

saschagrunert:
#### What type of PR is this?

/kind cleanup

We already test for the availability of seccomp in `setupSeccomp()` so
we can remove the additional check.

None

None

```release-note
None
```

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -444,10 +444,6 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 		return nil, err
 	}
 
-	if s.Config().Seccomp().IsDisabled() {
-		g.Config.Linux.Seccomp = nil
-	}
-
 	saveOptions := generate.ExportOptions{}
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
and happy green test buttons

kolyshkin:
The code being removed was introduced by commit 44e7e88ff365da02237e672762f88bf53891585c (PR #368).

The setupSeccomp() call was added earlier (commit 78ee03a8fc, PR #211), and is only called when the container is not privileged.

Question is: can we still run a privileged container using cri-o compiled with seccomp on a seccomp-less machine?

saschagrunert:
> Question is: can we still run a privileged container using cri-o compiled with seccomp on a seccomp-less machine?

Tested it locally with the (modified to `unconfined`) `test/testdata/sandbox_config_seccomp.json` and the code path ends up in https://github.com/cri-o/cri-o/blob/1c4d633661888bd6f3cd19cd130e9cfc3d4cf426/server/container_create.go#L616

PR #4086, opened 2020-08-17
Don't limit the size on /run for systemd based containers

rhatdan:
We had a customer incident where they ran out of space on /run.

If you don't specify size, it will be still limited to 50% or memory
available in the cgroup the container is running in.  If the cgroup is
unlimited then the /run will be limited to 50% of the total memory
on the system.

#### What type of PR is this?

/kind bug

We are limiting the size of /run, which is not unnecessary, and can cause random issues.

```release-note
None
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -1102,7 +1102,7 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
 			Destination: dest,
 			Type:        "tmpfs",
 			Source:      "tmpfs",
-			Options:     append(options, "tmpcopyup", "size=65536k"),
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1869322

haircommander:
LGTM

Can you add an entry in the release notes field at the bottom of the PR description (or none if you don't think it's appropriate)

TomSweeneyRedHat:
LGTM

rhatdan:
Added None to release notes.

kolyshkin:
CI is all green except for `codecov/patch`. Is there a way to force-merge?

PR #4087, opened 2020-08-17
Assorted test fixes (might help CI flakes)

kolyshkin:
/kind failing-test

A set of few fixes related to using bats' `run` function in the tests.
These are real fixes, not some refactoring or code style improvements.

##### 1. test/pull_test_containers: fix
    
Using `if ! run cmd` is useless as `run` always succeeds (and puts the
exit code of `cmd` to `$status`. Remove it.

##### 2. test/ping_pod_from_pod: fix
    
1. The bats `run` command always returns 0 exit code, so
       `ping_pod_from_pod` never tested anything, as it always succeeded.
    
2. Using `$output=$(run cmd)` is wrong, since `run` assigns the output
       itself, and do not print anything per se (so `$output` is always
       empty in this case).
    
3. Using `exit 1` from the function is excessive.
    
4. There is no need for the whole `if ... ` construct, as bats tests are
       run with `set -e` and if the ping fails the test will fail.
    
Fix the `ping_pod` in the same way.

##### 3. test/wait_until_reachable: fix
    
Commit 39aef1a09c1ad broke the command exit code check,
as `run` always returns 0.
    
Fix: do not use run.
    
This might help some CI flakes :crossed_fingers: 

##### 4. critest.bats wrong $status check

Variable `$output` is only set by the `run` function provided by bats.
Since we're not using `run` here, it does not make sense to check
$output, more to say, it is dangerous because it leads to test failure
(as most probably `$output` is unset and so check is failing).
    
This fixes the following issue:
    
     ...
     Ran 74 of 81 Specs in 57.964 seconds
     SUCCESS! -- 74 Passed | 0 Failed | 1 Flaked | 0 Pending | 7 Skipped
    
     Ginkgo ran 1 suite in 58.01304415s
     Test Suite Passed
     PASS
     not ok 1 run the critest suite
     # (in test file critest.bats, line 27)
     #   `[ "$status" -eq 0 ]' failed with status 2
     #

##### 5. test/critest.bats: move setup/cleanup out of test

This change results in

1. Not doing any unnecessary preparation steps in case we're going to
   skip the test.

2. Making sure cleanup is performed even if the test fails.

Note that stop_crio is in cleanup_test so there's no need to call it
again explicitly.

None

```release-note
None
```

### DIFF
--- a/test/critest.bats
+++ b/test/critest.bats
@@ -24,7 +24,5 @@ function teardown() {
                 --ginkgo.skip="${CRI_SKIP}" \
                 --ginkgo.flakeAttempts=3 >&3
 
-    [ "$status" -eq 0 ]
-
     stop_crio
 }
--- a/test/critest.bats
+++ b/test/critest.bats
@@ -3,7 +3,12 @@
 load helpers
 
 function setup() {
+    if [[ -z $RUN_CRITEST ]]; then
+        skip "critest because RUN_CRITEST is not set"
+    fi
+
     setup_test
+    start_crio
 }
 
 function teardown() {
@@ -11,18 +16,10 @@ function teardown() {
 }
 
 @test "run the critest suite" {
-    if [[ -z $RUN_CRITEST ]]; then
-        skip "critest because RUN_CRITEST is not set"
-    fi
-
-    start_crio
-
     critest --parallel 8 \
                 --runtime-endpoint "${CRIO_SOCKET}" \
                 --image-endpoint "${CRIO_SOCKET}" \
                 --ginkgo.focus="${CRI_FOCUS}" \
                 --ginkgo.skip="${CRI_SKIP}" \
                 --ginkgo.flakeAttempts=3 >&3
-
-    stop_crio
 }
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -577,27 +577,15 @@ function get_host_ip() {
 
 function ping_pod() {
     ipv4=$(parse_pod_ipv4 "$1")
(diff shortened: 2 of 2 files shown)


kolyshkin:
@saschagrunert PTAL

haircommander:
https://app.circleci.com/pipelines/github/cri-o/cri-o/5460/workflows/ea592fc1-e945-4942-8c14-c81b06037428/jobs/85208
```

# /tmp/bats-run-24234/bats.24264.src: line 27: [: : integer expression expected
```
cri test fails (as sad as I am to relinquish an otherwise green test run)

kolyshkin:
> cri test fails (as sad as I am to relinquish an otherwise green test run)

It's a super funny story actually. critest fails because it checks `$output` without using `run` before. It worked since some code in `helpers.bash` (erroneously) used `run` which set `$status`, which the `critest.bats` was checking.

The fix to that one is in https://github.com/cri-o/cri-o/pull/4091/commits/bbb9bbe6326385c091d84dcb88dbfc3f3e87b5cc

Let me move that commit from #4091 to here.

kolyshkin:
> Let me move that commit from #4091 to here.

done

haircommander:
> > cri test fails (as sad as I am to relinquish an otherwise green test run)
> 
> It's a super funny story actually. critest fails because it checks `$output` without using `run` before. It worked since some code in `helpers.bash` (erroneously) used `run` which set `$status`, which the `critest.bats` was checking.

:upside_down_face: 

/retest

PR #4090, opened 2020-08-18
Use Unmount w/o prior Mounted check

kolyshkin:
/kind bug

The mount.Mounted check is super expensive, as it has to parse the whole
nine yards of /proc/self/mountinfo, which is slow and sometimes racy
(for latter, see https://github.com/kolyshkin/procfs-test).

It is much simpler and faster to just try unmount, and ignore EINVAL
error. Note that EINVAL might be returned if the argument is not a mount
point, or if flags are wrong. Since the flags is hardcoded here, the
code will only get EINVAL in case the path is not mounted.

This should speed things up a little, and fix an occasional case of real
mount not being unmounted because of the mountinfo race.

Also, since we do not check the path in mountinfo, and umount(2) deals
just fine with symlinks, we no longer need to resolve the path, so
remove the path cleaning code (which in fact should have been
filepath.EvalSymlinks()).

None

```release-note
NONE
```

### DIFF
--- a/internal/lib/sandbox/namespaces_linux.go
+++ b/internal/lib/sandbox/namespaces_linux.go
@@ -10,9 +10,7 @@ import (
 	"sync"
 
 	nspkg "github.com/containernetworking/plugins/pkg/ns"
-	"github.com/containers/storage/pkg/mount"
 	"github.com/cri-o/cri-o/pkg/config"
-	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


kolyshkin:
Might have help CI issues like this one:

> time="2020-08-17T21:57:41Z" level=error msg="removing the pod sandbox \"7d85850fd5de35f44b5b4d73c144290a73ae6b5d6ac6c8c44471acc55943586d\" failed: rpc error: code = Unknown desc = unable to remove managed namespaces: Removing namespaces encountered the following errors [unlinkat /var/run/netns/e050a9be-5dd8-4a55-9229-d05cbe44ce7f: device or resource busy]"

which might be caused by a race in mount.Mounted(), which results in skipping the Unmount, which results in EBUSY we're seeing here.

internal/lib/sandbox/namespaces_linux.go
```diff
@@ -176,29 +174,14 @@ func (n *Namespace) Remove() error {
...
 	n.closed = true
 
-	// we got namespaces in the form of
-	// /var/run/$NSTYPEns/$NSTYPE-d08effa-06eb-a963-f51a-e2b0eceffc5d
-	// but /var/run on most system is symlinked to /run so we first resolve
-	// the symlink and then try and see if it's mounted
-	fp, err := securejoin.SecureJoin("/", n.Path())
-	if err != nil {
-		return errors.Wrapf(err, "unable to join '/' with %s path", n.Path())
-	}
```
cyphar, line 186: Yeah this is all quite :thinking:-worthy. And :+1: on removing the needless "is it mounted" check.

haircommander:
```
# time="2020-08-18T12:21:31Z" level=error msg="removing the pod sandbox \"676baa072e2d956cf423ee2c848ab63fae6ed53140db151426c3de2a0b1ba853\" failed: rpc error: code = Unknown desc = unable to remove managed namespaces: Removing namespaces encountered the following errors [unlinkat /var/run/ipcns/cd753410-5bf9-45bb-949e-d807bc839700: device or resource busy unlinkat /var/run/netns/cd753410-5bf9-45bb-949e-d807bc839700: device or resource busy]"
```
alas, it does not totally fix the problem, though it may mitigate it

TomSweeneyRedHat:
LGTM

kolyshkin:
The only failing CI is `codecov/patch`. I don't see any need to cover this with unit tests, the code is simple.

PR #4094, opened 2020-08-18
Validate that apparmor_parser binary is available on the host

saschagrunert:
#### What type of PR is this?

/kind feature

If AppArmor is enabled, then we now pre-check on CRI-O startup if the
`apparmor_parser` binary is either in `/sbin` or in `$PATH`. This way we
can let the user know that there is still a missing runtime dependency.

None

None

```release-note
- Added validation of the existence of the `apparmor_parser` binary (either in `/sbin` or in `$PATH`) on CRI-O start, if the feature is enabled by the build or the host.
```

### DIFF
--- a/internal/config/apparmor/apparmor.go
+++ b/internal/config/apparmor/apparmor.go
@@ -1,6 +1,9 @@
 package apparmor
 
 import (
+	"os"
+	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/containers/libpod/pkg/apparmor"
@@ -9,12 +12,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-const (
-	// DefaultProfile is the default profile name
-	DefaultProfile = "crio-default"
-
-	unconfined = "unconfined"
-)
+// DefaultProfile is the default profile name
+const DefaultProfile = "crio-default"
 
 // Config is the global AppArmor configuration type
 type Config struct {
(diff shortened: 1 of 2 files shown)


haircommander:
LGTM

TomSweeneyRedHat:
LGTM

kolyshkin:
IMHO this check should be added to github.com/containers/libpod/pkg/apparmor.IsEnabled() instead, since it is the one that uses and needs that binary (and cri-o does not).

It did that before by calling runc's libcontainer/apparmor.IsEnabled(), but now runc doesn't check binary presense (as it's not using it itself).

internal/config/apparmor/apparmor.go
```diff
@@ -38,9 +37,9 @@ func (c *Config) LoadProfile(profile string) error {
 		return nil
 	}
 
-	if profile == unconfined {
+	if profile == v1.AppArmorBetaProfileNameUnconfined {
 		logrus.Info("AppArmor profile is unconfined which basically disables it")
-		c.defaultProfile = unconfined
+		c.defaultProfile = v1.AppArmorBetaProfileNameUnconfined
```
kolyshkin, line 42: This should be a separate commit, please.

saschagrunert:
> IMHO this check should be added to github.com/containers/libpod/pkg/apparmor.IsEnabled() instead, since it is the one that uses and needs that binary (and cri-o does not).

The most recent version of Podman does use containers/common instead of its own implementation. We probably have to add it there and switch over to the new library. I really like to make the issue (binary is missing) visible to the user instead of just disabling AppArmor. It is also a matter _when_ to report the issue to the user. Having the validation on CRI-O startup seems more reasonable to me than doing it late on container creation time.

I'll add the check to containers/common and follow-up with this PR later on.

internal/config/apparmor/apparmor.go
saschagrunert: Will be a dedicated PR, thank you for the hint.

saschagrunert: See https://github.com/cri-o/cri-o/pull/4098

saschagrunert:
Following-up in https://github.com/containers/common/pull/269

PR #4099, opened 2020-08-19
tests: support default runtime different from runc

fgiudici:
#### What type of PR is this?

/kind bug

When running the tests specifing a different runtime, like, e.g.:
 $> CONTAINER_DEFAULT_RUNTIME=crun \
    RUNTIME_NAME=crun \
    CONTAINER_RUNTIME=crun \
    RUNTIME_ROOT=/run/containers \
    make localintegration
some tests (command.bats and config_migrate.bats) call the crio binary
without passing the RUNTIME_NAME, CONTAINER_RUNTIME and the RUNTIME_ROOT
environment variables, while CONTAINER_DEFAULT_RUNTIME is taken up
directly by the crio binary. This will cause the test to fail as the
default runtime is not found in the config.

(in test file test/command.bats, line 71)
 `[[ "$output" == *"log size max should be negative or >= 8192"* ]]' failed
time="2020-08-18 14:27:11.944731362+02:00" level=info msg="Node configuration value for hugetlb cgroup is true" time="2020-08-18 14:27:11.944804578+02:00" level=info msg="Node configuration value for pid cgroup is true" time="2020-08-18 14:27:11.944824466+02:00" level=info msg="Node configuration value for memoryswap cgroup is true" time="2020-08-18 14:27:11.950071414+02:00" level=info msg="Node configuration value for systemd CollectMode is true" time="2020-08-18 14:27:11.950143720+02:00" level=fatal msg="Validating runtime config: default_runtime set to \"crun\", but no runtime entry was found for it"

 None

This looked to me as the best approach to deal with tests that need to call the CRI-O binary directly if we want to support running integration tests with different runtimes than runc.

None

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -223,7 +223,7 @@ function setup_test() {
 # Run crio using the binary specified by $CRIO_BINARY_PATH.
 # This must ONLY be run on engines created with `start_crio`.
 function crio() {
-    "$CRIO_BINARY_PATH" --listen "$CRIO_SOCKET" "$@"
+    "$CRIO_BINARY_PATH" --listen "$CRIO_SOCKET" --runtimes "$RUNTIME_NAME:$RUNTIME_BINARY:$RUNTIME_ROOT:$RUNTIME_TYPE" "$@"
 }
 
 # Run crictl using the binary specified by $CRICTL_BINARY.
--- a/test/command.bats
+++ b/test/command.bats
@@ -3,20 +3,20 @@
 load helpers
 
 @test "crio commands" {
-	run ${CRIO_BINARY_PATH} -c /dev/null config > /dev/null
+	run crio -c /dev/null config > /dev/null
 	echo "$output"
 	[ "$status" -eq 0 ]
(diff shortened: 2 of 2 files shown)


umohnani8:
LGTM

test/helpers.bash
```diff
@@ -223,7 +223,7 @@ function setup_test() {
 # Run crio using the binary specified by $CRIO_BINARY_PATH.
 # This must ONLY be run on engines created with `start_crio`.
 function crio() {
```
haircommander, line 225: as a follow up (or you can address it here) we should change this function name (crio-with-args?) it is not clear we're not running the crio binary in $PATH when calling it above.

haircommander:
I actually think I have a different way of doing this which simplifies

TomSweeneyRedHat:
LGTM

fidencio:
@fgiudici, @haircommander, can we close this one in favour of https://github.com/cri-o/cri-o/pull/4100?

fgiudici:
> @fgiudici, @haircommander, can we close this one in favour of #4100?

Absolutely! Thanks, closing

PR #4100, opened 2020-08-19
test: reconfigure how runtimes are passed in

haircommander:
#### What type of PR is this?

 /kind cleanup

As well as update documentation describing how to specify an alternative runtime
And fix some inconsistencies in how the runtime is called in restore.bats

replaces https://github.com/cri-o/cri-o/pull/4099

```release-note
none
```

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -15,7 +15,7 @@ Integration tests are written in *bash* using the
 
 ### Containerized tests
 
-The easiest way to run integration tests is with Docker:
+The easiest way to run integration tests is with Podman:
 ```
 $ make integration
 ```
@@ -70,11 +70,16 @@ Tests on the host will run with `runc` as the default runtime.
 However you can select other OCI compatible runtimes by setting
 the `RUNTIME` environment variable.
 
-For example one could use the [Clear Containers](https://github.com/clearcontainers/runtime)
-runtime instead of `runc`:
-
+For example, to use [crun](https://github.com/containers/crun) instead of `runc`:
 ```
-make localintegration RUNTIME=cc-runtime
+make CONTAINER_DEFAULT_RUNTIME=crun localintegration
+```
+
+If you'd like to run the tests with a runtime of a different type, you need to also specify `$RUNTIME_TYPE`
+
+For example, to use [kata](https://github.com/kata-containers/runtime) with shim v2:
+```
+make CONTAINER_DEFAULT_RUNTIME=containerd-shim-kata-v2 RUNTIME_TYPE=vm localintegration
 ```
 
 ## Writing integration tests
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -263,11 +263,6 @@ function setup_crio() {
     CNI_DEFAULT_NETWORK=${CNI_DEFAULT_NETWORK:-crio}
     CNI_TYPE=${CNI_TYPE:-bridge}
 
-    # Workaround for https://github.com/containers/crun/pull/531.
-    # TODO: remove once crun > 0.15 is released and used here.
-    if $RUNTIME_BINARY --version | grep -q '^crun '; then
-        OVERRIDE_OPTIONS="$OVERRIDE_OPTIONS --selinux=false"
-    fi
     RUNTIME_ROOT="$TESTDIR/crio-runtime-root"
 
     # shellcheck disable=SC2086
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -35,16 +35,9 @@ CONTAINER_UID_MAPPINGS=${CONTAINER_UID_MAPPINGS:-}
 CONTAINER_GID_MAPPINGS=${CONTAINER_GID_MAPPINGS:-}
 OVERRIDE_OPTIONS=${OVERRIDE_OPTIONS:-}
 # Runtime
-CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-runc}
 CONTAINER_DEFAULT_RUNTIME=${CONTAINER_DEFAULT_RUNTIME:-runc}
-RUNTIME_NAME=${RUNTIME_NAME:-runc}
-RUNTIME_PATH=$(command -v "$CONTAINER_RUNTIME" || true)
-RUNTIME_BINARY=${RUNTIME_PATH:-$(command -v runc)}
+RUNTIME_BINARY_PATH=$(command -v "$CONTAINER_DEFAULT_RUNTIME")
 RUNTIME_TYPE=${RUNTIME_TYPE:-oci}
-if [[ $CONTAINER_RUNTIME == "kata-runtime" ]]; then
-    export RUNTIME_NAME="$CONTAINER_RUNTIME"
-    export CONTAINER_DEFAULT_RUNTIME="$RUNTIME_NAME"
-fi
 # Path of the apparmor_parser binary.
 APPARMOR_PARSER_BINARY=${APPARMOR_PARSER_BINARY:-/sbin/apparmor_parser}
 # Path of the apparmor profile for test.
@@ -200,7 +193,7 @@ function crictl() {
 
 # Run the runtime binary with the specified RUNTIME_ROOT
 function runtime() {
-    "$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" "$@"
+    "$RUNTIME_BINARY_PATH" --root "$RUNTIME_ROOT" "$@"
 }
 
 # Communicate with Docker on the host machine.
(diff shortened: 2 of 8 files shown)


haircommander:
@fgiudici  PTAL
the advantage here is we only need to specify 
`CONTAINER_DEFAULT_RUNTIME=crun` and the test takes care of the rest (or also RUNTIME_TYPE if it's vm based)

test/README.md
```diff
@@ -70,11 +70,16 @@ Tests on the host will run with `runc` as the default runtime.
...
+```
+make CONTAINER_DEFAULT_RUNTIME=crun localintegration
+```
+
+If you'd like to run the tests with a runtime of a different type, you need to also specify `$RUNTIME_TYPE`
 
+For example, to use [kata](https://github.com/kata-containers/runtime) with shim v2:
 ```
-make localintegration RUNTIME=cc-runtime
+make CONTAINER_DEFAULT_RUNTIME=kata RUNTIME_TYPE=vm localintegration
```
fgiudici, line 82: The kata shim v2 binary has a slightly cumbersome name ;-)
```suggestion
make CONTAINER_DEFAULT_RUNTIME=containerd-shim-kata-v2 RUNTIME_TYPE=vm localintegration
```
Not all tests will work with it (at least for now), but nonetheless we support specifying the RUNTIME_TYPE, so would be good to have this documented.

test/helpers.bash
```diff
@@ -304,9 +303,9 @@ function setup_crio() {
         --listen "$CRIO_SOCKET" \
         --registry "quay.io" \
         --registry "docker.io" \
-        --runtimes "$RUNTIME_NAME:$RUNTIME_BINARY:$RUNTIME_ROOT:$RUNTIME_TYPE" \
         -r "$TESTDIR/crio" \
         --runroot "$TESTDIR/crio-run" \
+        --runtimes "$CONTAINER_RUNTIMES" \
```
fgiudici, line 308: If we export CONTAINER_RUNTIMES can we also skip the --runtimes parameter here?

test/helpers.bash
```diff
@@ -33,15 +33,14 @@ CONTAINER_CONMON_CGROUP=${CONTAINER_CONMON_CGROUP:-pod}
 CONTAINER_SECCOMP_PROFILE=${CONTAINER_SECCOMP_PROFILE:-${CRIO_ROOT}/vendor/github.com/seccomp/containers-golang/seccomp.json}
 # Runtime
 CONTAINER_DEFAULT_RUNTIME=${CONTAINER_DEFAULT_RUNTIME:-runc}
-RUNTIME_NAME=${RUNTIME_NAME:-runc}
-CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-runc}
+RUNTIME_ROOT=${RUNTIME_ROOT:-/run/runc}
+RUNTIME_BINARY_PATH=$(command -v $CONTAINER_DEFAULT_RUNTIME || true)
+RUNTIME_TYPE=${RUNTIME_TYPE:-oci}
+# export here so direct calls to crio later inherit the variable
+export CONTAINER_RUNTIMES=${CONTAINER_RUNTIMES:-$CONTAINER_DEFAULT_RUNTIME:$RUNTIME_BINARY_PATH:$RUNTIME_ROOT:$RUNTIME_TYPE}
```
fgiudici, line 40: Brilliant!

fgiudici (commented):
LGTM

test/helpers.bash
haircommander: I found some weirdness without it, but I forget what that was. Let's try!

haircommander:
conflict resolved can I have my tag back

haircommander:
`Kubernetes e2e suite: [sig-network] IngressClass [Feature:Ingress] should set default value on new IngressClass [Serial]`
known flake
/retest

haircommander:
ptal @mrunalp @kolyshkin @umohnani8 @giuseppe @saschagrunert

test/restore.bats
```diff
@@ -100,7 +100,7 @@ function teardown() {
 	stop_crio
 
 	# simulate reboot with runc state going away
-	"$CONTAINER_RUNTIME" delete -f "$pod_id"
+	"$RUNTIME_BINARY_PATH" --root "$RUNTIME_ROOT" delete -f "$pod_id"
```
kolyshkin, line 103: so, since you're using `"$RUNTIME_BINARY_PATH" --root "$RUNTIME_ROOT"` from many places, makes sense to have something like

```bash
function runtime() {
   "$RUNTIME_BINARY_PATH" --root "$RUNTIME_ROOT" "$@"
}
```
and use it everywhere. For example, the above would become
```suggestion
	runtime delete -f "$pod_id"
```
which I think helps readability as well as maintenance (making future patches smaller).

test/helpers.bash
```diff
@@ -35,17 +35,15 @@ CONTAINER_UID_MAPPINGS=${CONTAINER_UID_MAPPINGS:-}
...
+RUNTIME_ROOT=${RUNTIME_ROOT:-/run/$CONTAINER_DEFAULT_RUNTIME}
+RUNTIME_BINARY_PATH=$(command -v "$CONTAINER_DEFAULT_RUNTIME" || true)
 RUNTIME_TYPE=${RUNTIME_TYPE:-oci}
-if [[ $CONTAINER_RUNTIME == "kata-runtime" ]]; then
-    export RUNTIME_NAME="$CONTAINER_RUNTIME"
-    export CONTAINER_DEFAULT_RUNTIME="$RUNTIME_NAME"
-fi
+# export here so direct calls to crio later inherit the variable
+export CONTAINER_RUNTIMES=${CONTAINER_RUNTIMES:-$CONTAINER_DEFAULT_RUNTIME:$RUNTIME_BINARY_PATH:$RUNTIME_ROOT:$RUNTIME_TYPE}
+CONTAINER_UID_MAPPINGS=${CONTAINER_UID_MAPPINGS:-}
```
kolyshkin, line 44: Apparently this one is duplicated?

test/helpers.bash
```diff
@@ -35,17 +35,15 @@ CONTAINER_UID_MAPPINGS=${CONTAINER_UID_MAPPINGS:-}
...
+CONTAINER_GID_MAPPINGS=${CONTAINER_GID_MAPPINGS:-}
+OVERRIDE_OPTIONS=${OVERRIDE_OPTIONS:-}
```
kolyshkin, line 46: ... as well as these two

test/helpers.bash
```diff
@@ -35,17 +35,15 @@ CONTAINER_UID_MAPPINGS=${CONTAINER_UID_MAPPINGS:-}
...
 OVERRIDE_OPTIONS=${OVERRIDE_OPTIONS:-}
 # Runtime
-CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-runc}
 CONTAINER_DEFAULT_RUNTIME=${CONTAINER_DEFAULT_RUNTIME:-runc}
-RUNTIME_NAME=${RUNTIME_NAME:-runc}
-RUNTIME_PATH=$(command -v "$CONTAINER_RUNTIME" || true)
-RUNTIME_BINARY=${RUNTIME_PATH:-$(command -v runc)}
-RUNTIME_ROOT=${RUNTIME_ROOT:-/run/runc}
+RUNTIME_ROOT=${RUNTIME_ROOT:-/run/$CONTAINER_DEFAULT_RUNTIME}
+RUNTIME_BINARY_PATH=$(command -v "$CONTAINER_DEFAULT_RUNTIME" || true)
```
kolyshkin, line 40: Maybe we should fail here (i.e. remove ` || true` part).

fidencio:
@fgiudici, @haircommander, is this still needed?  If so, @haircommander, would you need someone to take it over or do you have the cycles to take a look at @kolyshkin's comments and do a rebase?

haircommander:
I also cleaned up a work around fro crun that seems to be out of date (so I don't forget)

test/helpers.bash
```diff
@@ -198,6 +198,11 @@ function crictl() {
     "$CRICTL_BINARY" -r "unix://$CRIO_SOCKET" -i "unix://$CRIO_SOCKET" "$@"
 }
 
+# Run the runtime binary with the specified RUNTIME_ROOT
+function runtime() {
+    "$RUNTIME_BINARY_PATH" --root "$RUNTIME_ROOT" "$@"
```
kolyshkin, line 196: Hmm, did you mean `"$CONTAINER_RUNTIME"` here? There's no `"$RUNTIME_BINARY_PATH"` (yet?)

kolyshkin: Ah, I see, the order of commits seems to be wrong (or github is showing them in the wrong order), as this ^^^ belongs to the first commit, and `$RUNTIME_BINARY_PATH` is introduced by the third commit.

haircommander: oopsies, updated!

kolyshkin:
1. I think that patches 2 and 4 can be combined into one -- here it does not make sense to add `|| true` and then remove it.

2. Patch 2 breaks all the tests, as it removes the variable used by them. This might play bad if someone wants to use e.g. git bisect.

haircommander:
>     1. I think that patches 2 and 4 can be combined into one -- here it does not make sense to add `|| true` and then remove it.
> 
>     2. Patch 2 breaks all the tests, as it removes the variable used by them. This might play bad if someone wants to use e.g. git bisect.

okay, I have implemented point 1. For point 2, I have updated the second commit to not change the variable name in the `runtime()` function, and update it in the third commit. A git bisect should allow each commit to pass, now

kolyshkin (approved):
LGTM, thanks!

haircommander:
@mrunalp @saschagrunert @fidencio @kolyshkin @umohnani8 @QiWang19 plz merge

saschagrunert (approved):
/lgtm

PR #4101, opened 2020-08-19
managed ns: ignore `PID not initialized` on sandbox creation

haircommander:
#### What type of PR is this?

/kind bug

> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

everytime we create a sandbox with a pinned namespace, we spit out an error that says:
`Pid for infra container $id not found: PID not initialized`

in the pinned namespace case, we create the namespaces, create the networking stack (giving the created net namespace to the plugin), THEN create the pod.
Thus, we don't expect the infra container to be initialized.

Fix this by ignoring ErrNotInitialized when we're fetching the infraPid when finding the namespaces.

We can safely do this because the managed namespace case is covered (we don't use the pid, but use the managed namespace struct we've created)
And in the non-managed namespace case, we will (correctly) fail later because we return an invalid PID, which matches our invalid infra container.

```release-note
none
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -28,6 +28,7 @@ var (
 	defaultStopSignal   = strconv.Itoa(defaultStopSignalInt)
 	ErrContainerStopped = errors.New("container is already stopped")
 	ErrNotFound         = errors.New("container process not found")
+	ErrNotInitialized   = errors.New("container PID not initialized")
 )
 
 // Container represents a runtime container.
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
LGTM

saschagrunert (approved):
/lgtm
Noticed that as well, thanks for the fix!

internal/oci/container.go
```diff
@@ -413,10 +414,10 @@ func (c *Container) Pid() (int, error) {
 // and it is the same process that was originally started by the runtime.
 func (c *Container) pid() (int, error) {
 	if c.state == nil {
-		return 0, errors.New("state not initialized")
+		return 0, ErrNotInitialized
 	}
 	if c.state.InitPid <= 0 {
-		return 0, errors.New("PID not initialized")
+		return 0, ErrNotInitialized
```
TomSweeneyRedHat, line 420: Do we really want to have the same error message being spun up for both errors?

haircommander: yeah the above check is mostly a sanity check to prevent us from segfaulting. The result is the same for both: we're not in a position to check the pid yet

PR #4112, opened 2020-08-24
Fix git-validation by always using HEAD^

saschagrunert:
#### What type of PR is this?

/kind failing-test

The merge base is wrong for release branches. To decrease the complexity
we now just use the HEAD^ commit and assume that a PR consists of one
commit.

None
git-validation is currently broken for release-1.19. So we might also
/cherrypick release-1.19

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,11 +56,9 @@ SHELLCHECK := ${BUILD_BIN_PATH}/shellcheck
 ifeq ($(shell bash -c '[[ `command -v git` && `git rev-parse --git-dir 2>/dev/null` ]] && echo true'), true)
 	COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
 	GIT_TREE_STATE := $(if $(shell git status --porcelain --untracked-files=no),dirty,clean)
-	GIT_MERGE_BASE := $(shell git merge-base origin/master $(shell git rev-parse --abbrev-ref HEAD))
 else
 	COMMIT_NO := unknown
 	GIT_TREE_STATE := unknown
-	GIT_MERGE_BASE := HEAD^
 endif
 
 # pass crio CLI options to generate custom configuration options at build time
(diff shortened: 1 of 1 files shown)


haircommander:
I am unsure of this one, though not enough to put a hold on it. I would rather we find a way to actually only check the commits that differ between upstream and HEAD. There are definite cases to have more than one commit per PR, and we risk having those not signed if we don't validate...

saschagrunert:
> I am unsure of this one, though not enough to put a hold on it. I would rather we find a way to actually only check the commits that differ between upstream and HEAD. There are definite cases to have more than one commit per PR, and we risk having those not signed if we don't validate...

Hm I'm not sure about the git-validation in general. Do we really need it? We enforce the signing already via prow, and checking for whitespace changes is mostly done by our linters. The only thing we not check right now is the "short subject".

haircommander:
> > I am unsure of this one, though not enough to put a hold on it. I would rather we find a way to actually only check the commits that differ between upstream and HEAD. There are definite cases to have more than one commit per PR, and we risk having those not signed if we don't validate...
> 
> Hm I'm not sure about the git-validation in general. Do we really need it? We enforce the signing already via prow, and checking for whitespace changes is mostly done by our linters. The only thing we not check right now is the "short subject".

yeah the one whitespace fix we miss is in the actual git commit itself. though, I can't really see that being very useful. I am in favor of dropping it. wdyt @umohnani8 @mrunalp

saschagrunert:
> > > I am unsure of this one, though not enough to put a hold on it. I would rather we find a way to actually only check the commits that differ between upstream and HEAD. There are definite cases to have more than one commit per PR, and we risk having those not signed if we don't validate...
> > 
> > 
> > Hm I'm not sure about the git-validation in general. Do we really need it? We enforce the signing already via prow, and checking for whitespace changes is mostly done by our linters. The only thing we not check right now is the "short subject".
> 
> yeah the one whitespace fix we miss is in the actual git commit itself. though, I can't really see that being very useful. I am in favor of dropping it. wdyt @umohnani8 @mrunalp

I'll propose that as another PR so that we can continue the discussion there :+1:

PR #4120, opened 2020-08-27
Refactor hostport manager into own package

saschagrunert:
#### What type of PR is this?

/kind cleanup

Both hostport managers are now part of a new package which can be
tested.

Follow-up of https://github.com/cri-o/cri-o/pull/4116
/cc @aojea 

```release-note
None
```

### DIFF
--- a/internal/hostport/hostport.go
+++ b/internal/hostport/hostport.go
@@ -1,0 +1,75 @@
+package hostport
+
+import (
+	"net"
+
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"k8s.io/kubernetes/pkg/kubelet/dockershim/network/hostport"
+	iptablesproxy "k8s.io/kubernetes/pkg/proxy/iptables"
+	utiliptables "k8s.io/kubernetes/pkg/util/iptables"
+	utilexec "k8s.io/utils/exec"
+)
+
+// Manager is the main structure of this package.
+type Manager struct {
+	v4 hostport.HostPortManager
+}
+
+// New creates a new hostport.Manager instance.
+func New() *Manager {
+	iptInterface := utiliptables.New(utilexec.New(), utiliptables.ProtocolIPv4)
+	if _, err := iptInterface.EnsureChain(
+		utiliptables.TableNAT, iptablesproxy.KubeMarkMasqChain,
+	); err != nil {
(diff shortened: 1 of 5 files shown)


server/sandbox_network.go
```diff
@@ -150,20 +135,8 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
-		var err error
-		if utilnet.IsIPv6String(ips[0]) {
-			err = s.hostportManagerv6.Remove(sb.ID(), mapping)
-		} else {
-			err = s.hostportManager.Remove(sb.ID(), mapping)
-		}
-		if err != nil {
-			log.Warnf(ctx, "failed to remove hostport for pod sandbox %s(%s): %v",
-				sb.Name(), sb.ID(), err)
+		if err := s.hostportManager.Remove(sb, ips[0]); err != nil {
```
aojea, line 138: thinking in loud, if we pass the sandbox to the Remove() method we may implement the logic to check if there are still ips inside.

aojea:
/hold
I don 't know but I somehow break something :)

TomSweeneyRedHat:
LGTM

server/sandbox_network.go
haircommander: I agree with this, I think encapsulating the behavior makes sense :+1:

saschagrunert:
@haircommander @aojea rebased on top of the latest master and still think that we can wrap the manager in a new package to have testing.

aojea:
> @haircommander @aojea rebased on top of the latest master and still think that we can wrap the manager in a new package to have testing.

we should definitively do this, depending on the kubernetes code is risky https://github.com/kubernetes/kubernetes/pull/94382#issuecomment-686786468

internal/hostport/hostport.go
```diff
@@ -0,0 +1,75 @@
+package hostport
+
+import (
+	"net"
+
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"k8s.io/kubernetes/pkg/kubelet/dockershim/network/hostport"
```
aojea, line 9: maybe we should stop vendoring and having it here?
cc: @danwinship

saschagrunert: I'm in favor of that.

saschagrunert: Just wondering how… @mrunalp would it be fine if we copy the manager sources into CRI-O and refactor / add testing?

PR #4123, opened 2020-08-27
Revert "dual stack portmap support"

abhat:
Reverts cri-o/cri-o#4116


abhat:
The branch will be reverted manually. This PR is not needed.

PR #4125, opened 2020-08-27
[release-1.18] Backport "Use the container_kvm_t label when using kata as the runtime" (and follow-ups)

fidencio:
#### What type of PR is this?

/kind feature

In order to have take fully advantage of SELinux, would be good to have the commits enabling it on CRI-O side as part of release-1.18 as well (and not only 1.19).

While I know backporting features is not exactly the best practices, this only affects the VM runtime type and is a quite self-contained.

None

```release-note
None
```


haircommander:
LGTM

PR #4127, opened 2020-08-28
test/network.bats: fixes

kolyshkin:
/kind bug
```release-note
NONE
```

### 1. test/network.bats: fix "clean up network" cases
    
Commit 58657488e31 replaced the last component of
/var/lib/cni/networks path to be $RANDOM_CNI_NETWORK,
while in fact it should be $CNI_DEFAULT_NETWORK.
    
The bug was never found because bats do not set -o pipefail
and so the non-zero exit code from ls was ignored.
    
While at it,
 - move the common checking code into a separate function;
 - fix network cleanup in teardown as well;
 - helpers.bash:
   - add `fail` helper;
   - remove useless `echo 0` from functions;
   - eliminate write_plugin_test_args_network_conf;
   - simplify prepare_plugin_test_args_network_conf_malformed_result;
   - get rid of RANDOM_CNI_NETWORK to avoid confusion.

### 2. test/network.bats: simplify, fix shellcheck

1. Simplify the code.

For example,

        run crictl start ...
        echo "$output"
        [ "$status" -eq 0 ]

is equivalent to

        crictl start ...

as it much easier to read (yes, bats runs tests with set -e
so there is an error check in place).

One more example,
        run crictl runp ...
        echo "$output"
        [ "$status" -eq 0 ]
        pod_id="$output"

is roughly equivalent to

        pod_id=$(crictl runp ...)

(except it does not print stdout, only stderr).

Another example,

        cfg=$(python .... < $input)
        echo "$cfg" > $output

is equivalent to

        python ... < $input > $output

and avoids the use of a temporary variable (saves some memory).

Finally,

        run crictl start ...
        echo "$output"
        [ "$status" -ne 0 ]

is equivalent to

        crictl start ... && false

2. Optimize use of `CONMON_BINARY` variable, and remove unneeded cleanup.

3. Fix indentation, add some vertical white-space for better readability.

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -82,11 +82,9 @@ STREAM_PORT=${STREAM_PORT:-10010}
 CONTAINER_METRICS_PORT=${CONTAINER_METRICS_PORT:-9090}
 
 POD_IPV4_CIDR="10.88.0.0/16"
-POD_IPV4_CIDR_START="10.88"
 POD_IPV4_DEF_ROUTE="0.0.0.0/0"
 
 POD_IPV6_CIDR="1100:200::/24"
-POD_IPV6_CIDR_START="1100:200::"
 POD_IPV6_DEF_ROUTE="1100:200::1/24"
 
 # Make sure we have a copy of the redis:alpine image.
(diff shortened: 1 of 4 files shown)


kolyshkin:
@saschagrunert PTAL

kolyshkin:
This one is def unrelated as I was not touching network_ping.bats file.
```
not ok 114 Ping pod from another pod
# (from function `crictl' in file ./helpers.bash, line 230,
#  from function `ping_pod_from_pod' in file ./helpers.bash, line 586,
#  in test file ./network_ping.bats, line 53)
#   `ping_pod_from_pod $ctr2_id $ctr1_id' failed
```

Lots of k8s tests are also unrelated.

kolyshkin:
Oh wow

```
# --- 10.88.0.16 ping statistics ---
# 5 packets transmitted, 5 received, 0% packet loss, time 4080ms
# rtt min/avg/max/mdev = 0.024/0.042/0.101/0.029 ms
# time="2020-08-30 16:52:54.297012737-07:00" level=debug msg="Request: &ExecSyncRequest{ContainerId:2ae017d8e105e8b42cca9b5a7f48774f4e22e20c28634c29ca3e80db644e5bec,Cmd:[ip addr show dev eth0 scope global],Timeout:0,}" file="go-grpc-middleware/chain.go:25" id=3e035cdb-0514-43e3-960b-aa2d33472f2b name=/runtime.v1alpha2.RuntimeService/ExecSync
# time="2020-08-30 16:52:54.297068649-07:00" level=error msg="Checking if PID of 2ae017d8e105e8b42cca9b5a7f48774f4e22e20c28634c29ca3e80db644e5bec is running failed: PID running but not the original container. PID wrap may have occurred" file="oci/container.go:396"
# time="2020-08-30 16:52:54.297097497-07:00" level=debug msg="Response error: container is not created or running" file="go-grpc-middleware/chain.go:25" id=3e035cdb-0514-43e3-960b-aa2d33472f2b name=/runtime.v1alpha2.RuntimeService/ExecSync
# time="2020-08-30T16:52:54-07:00" level=fatal msg="execing command in container synchronously failed: rpc error: code = Unknown desc = container is not created or running"
```

There is something weird going on here

kolyshkin:
adding some debug patches to figure out what is going on

PR #4128, opened 2020-08-28
`crio-shutdown.service`: Support Graceful Reboot for AIO Node

hswong3i:
Currently our sample systemd service file
`contrib/systemd/crio-shutdown.service` handle shutdown dependency as
below:

    [Service]
    Type=oneshot
    ExecStart=/usr/bin/rm -f /var/lib/crio/crio.shutdown
    ExecStop=/usr/bin/env bash -c "/usr/bin/mkdir /var/lib/crio; /usr/bin/touch /var/lib/crio/crio.shutdown"
    RemainAfterExit=yes

For some rare condition, e.g. bare matel AIO deployment with Rook/CephFS
as loopback PVC, `crio.service` may not stop early enough before Ceph
components get stopped during graceful system reboot, therefore
containers may not able to be stopped correctly since PVC target already
gone, with current `/var/lib/crio/crio.shutdown` implementation.

This PR introduce:

  - `crio.service`: Ensure crio start after `remote-fs.target` so
    postpone as late as possible during startup, therefore always stop
    as early as possible during graceful system reboot
  - `crio-shutdown.service`: Always stop all running containers with
    `crictl` just before `crio.service` get stopped (i.e. avoid loopback
    deadlock)

The logic is concept proof by
<https://github.com/alvistack/ansible-role-etcd/tree/develop>; also
works as expected with Ceph + Kubernetes deployment by
<https://github.com/alvistack/ansible-collection-kubernetes/tree/develop>.
No more deadlock happened during graceful system reboot, both AIO
single/multiple node with loopback mount.

Also see:

  - <https://github.com/ceph/ceph/pull/36776>
  - <https://github.com/etcd-io/etcd/pull/12259>
  - <https://github.com/cri-o/cri-o/pull/4128>
  - <https://github.com/kubernetes/release/pull/1504>

Signed-off-by: Wong Hoi Sing Edison <hswong3i@gmail.com>

image

#### What type of PR is this?

> /kind cleanup
> /kind feature

```release-note
`crio-shutdown.service`: Support Graceful Reboot for AIO Node
```

### DIFF
--- a/contrib/systemd/crio-shutdown.service
+++ b/contrib/systemd/crio-shutdown.service
@@ -2,13 +2,11 @@
 Description=Shutdown CRI-O containers before shutting down the system
 Wants=crio.service
 After=crio.service
-Documentation=man:crio(8)
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/rm -f /var/lib/crio/crio.shutdown
-ExecStop=/usr/bin/env bash -c "/usr/bin/mkdir /var/lib/crio; /usr/bin/touch /var/lib/crio/crio.shutdown"
 RemainAfterExit=yes
+ExecStop=/usr/local/bin/crictl rm -af
 
 [Install]
 WantedBy=multi-user.target
--- a/contrib/systemd/crio-wipe.service
+++ b/contrib/systemd/crio-wipe.service
@@ -1,9 +1,10 @@
 [Unit]
 Description=CRI-O Auto Update Script
 Before=crio.service
-RequiresMountsFor=/var/lib/containers
+Wants=crio.service
 
 [Service]
+Type=oneshot
 EnvironmentFile=-/etc/sysconfig/crio
 ExecStart=/usr/local/bin/crio \
           $CRIO_CONFIG_OPTIONS \
@@ -12,8 +13,6 @@ ExecStart=/usr/local/bin/crio \
           $CRIO_NETWORK_OPTIONS \
           $CRIO_METRICS_OPTIONS \
           wipe
-
-Type=oneshot
 
 [Install]
 WantedBy=multi-user.target
--- a/contrib/test/ci/crio.service
+++ b/contrib/test/ci/crio.service
@@ -1,16 +1,16 @@
 [Unit]
-Description=Open Container Initiative Daemon
-Documentation=https://github.com/kubernetes-sigs/cri-o
-After=network-online.target
+Description=Container Runtime Interface for OCI (CRI-O)
+Documentation=https://github.com/cri-o/cri-o
+Wants=network-online.target local-fs.target remote-fs.target time-sync.target
+After=network-online.target local-fs.target remote-fs.target time-sync.target
 
 [Service]
 Type=notify
 EnvironmentFile=-/etc/sysconfig/crio
-EnvironmentFile=-/etc/sysconfig/crio-metrics
-EnvironmentFile=-/etc/sysconfig/crio-network
-EnvironmentFile=-/etc/sysconfig/crio-storage
 Environment=GOTRACEBACK=crash
 ExecStart=/usr/local/bin/crio \
+          $CRIO_CONFIG_OPTIONS \
+          $CRIO_RUNTIME_OPTIONS \
           $CRIO_STORAGE_OPTIONS \
           $CRIO_NETWORK_OPTIONS \
           $CRIO_METRICS_OPTIONS
--- a/contrib/systemd/crio.service
+++ b/contrib/systemd/crio.service
@@ -1,8 +1,8 @@
 [Unit]
 Description=Container Runtime Interface for OCI (CRI-O)
 Documentation=https://github.com/cri-o/cri-o
-Wants=network-online.target
-After=network-online.target
+Wants=network-online.target local-fs.target remote-fs.target time-sync.target
+After=network-online.target local-fs.target remote-fs.target time-sync.target
 
 [Service]
 Type=notify


contrib/systemd/crio-shutdown.service
```diff
@@ -2,13 +2,11 @@
 Description=Shutdown CRI-O containers before shutting down the system
 Wants=crio.service
 After=crio.service
-Documentation=man:crio(8)
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/rm -f /var/lib/crio/crio.shutdown
-ExecStop=/usr/bin/env bash -c "/usr/bin/mkdir /var/lib/crio; /usr/bin/touch /var/lib/crio/crio.shutdown"
```
haircommander, line 10: we still need this. there are some internal processes that need this file to be created

haircommander: both the ExecStart and ExecStop are needed

contrib/systemd/crio-wipe.service
```diff
@@ -1,9 +1,10 @@
 [Unit]
 Description=CRI-O Auto Update Script
 Before=crio.service
-RequiresMountsFor=/var/lib/containers
+Wants=crio.service
```
haircommander, line 4: isn't this contradictory with the Before above?

contrib/test/ci/crio.service
```diff
@@ -1,16 +1,16 @@
...
+Wants=network-online.target remote-fs.target
+After=network-online.target remote-fs.target
 
 [Service]
 Type=notify
 EnvironmentFile=-/etc/sysconfig/crio
-EnvironmentFile=-/etc/sysconfig/crio-metrics
-EnvironmentFile=-/etc/sysconfig/crio-network
-EnvironmentFile=-/etc/sysconfig/crio-storage
 Environment=GOTRACEBACK=crash
```
haircommander, line 10: why remove this?

contrib/systemd/crio-shutdown.service
hswong3i: I could only found https://github.com/cri-o/cri-o/pull/4128/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bL269-L273 which also mentioned we shouldn't do cleanup with this `/var/lib/crio/crio.shutdown` so I remove it, too.

Ideally, `crio-shutdown.service` will not be triggered during `crio.service` stop/restart so running containers could be keep; only system shutdown/reboot will trigger the `/usr/local/bin/crictl rm -af` before CRI-O daemon gone away.

contrib/systemd/crio-wipe.service
hswong3i: From https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Wants=

> Wants=
> 
> Configures requirement dependencies on other units. This option may be specified more than once or multiple space-separated units may be specified in one option in which case dependencies for all listed names will be created. Dependencies of this type may also be configured outside of the unit configuration file by adding a symlink to a .wants/ directory accompanying the unit file. For details, see above.
> 
> Units listed in this option will be started if the configuring unit is. However, if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole, and this unit will still be started. This is the recommended way to hook the start-up of one unit to the start-up of another unit.
> 
> Note that requirement dependencies do not influence the order in which services are started or stopped. This has to be configured independently with the After= or Before= options. If unit foo.service pulls in unit bar.service as configured with Wants= and no ordering is configured with After= or Before=, then both units will be started simultaneously and without any delay between them if foo.service is activated.

AFAIK, adding in Before/After + Wants/Requires combo could ensure dependencies being inject correctly.

contrib/test/ci/crio.service
hswong3i: Shall this file sync with https://github.com/cri-o/cri-o/pull/4128/files#diff-8f149e7773eb6fec135aa90b23aac30b?

hswong3i:
@haircommander since some legacy code cleanup from `server/server.go` and ` server/sandbox_stop.go` could we trigger CI for testing?

haircommander:
/ok-to-test

but honestly, I am not sure we want this. I do not consider the Shutdown code as legacy

hswong3i:
> /ok-to-test
> 
> but honestly, I am not sure we want this. I do not consider the Shutdown code as legacy

@haircommander since `crictl rm -af` could totally replace the functionality of Shutdown code (which `stopAllPodSandboxes()` didn't be used elsewhere), unless we have other use cases shouldn't it consider as legacy?

From the other point of view:
  - Relies on `crictl` looks not such strict forward for me...
  - Could we add a new wrapper command like `crio wipe`, e.g. `crio shutdown`, so we could access `stopAllPodSandboxes()` directly from command line (and so from systemd `ExecStop`, too)?

haircommander:
I'd like @mrunalp @umohnani8 and @saschagrunert to chime in as well. I'm still thinking....

contrib/systemd/crio-shutdown.service
```diff
@@ -2,13 +2,11 @@
...
 RemainAfterExit=yes
+ExecStop=/usr/local/bin/crictl rm -af
```
hswong3i, line 9: @haircommander @mrunalp @umohnani8 @saschagrunert could we have a wrapper CLI as direct call to `stopAllPodSandboxes()` here, similar as `crio wipe` shorthand?

```
ExecStart=/usr/local/bin/crio \
          $CRIO_CONFIG_OPTIONS \
          $CRIO_RUNTIME_OPTIONS \
          $CRIO_STORAGE_OPTIONS \
          $CRIO_NETWORK_OPTIONS \
          $CRIO_METRICS_OPTIONS \
          shutdown
```

haircommander: I would be more comfortable with that. can you spin up a PR

hswong3i: > I would be more comfortable with that. can you spin up a PR

I try to keep this PR clean and focus on systemd related issues and temporary handle shutdown with `crio rm -af`.

The implementation of `crio shutdown` goes to #4182; BTW, I have zero knowledge with golang... @haircommander please kindly advice how could we implement it ;-)

hswong3i:
@haircommander @saschagrunert could we get this PR merge first, then handle the improvement with #4128?

hswong3i:
@haircommander @saschagrunert PR is now refine as below:
```
After=network-online.target local-fs.target remote-fs.target time-sync.target
Wants=network-online.target local-fs.target remote-fs.target time-sync.target
```

Please kindly comment ;-)

hswong3i:
> @haircommander @saschagrunert could we get this PR merge first, then handle the improvement with #4128?

PR #4131, opened 2020-08-28
config: set internal RootConfig to default storage if not specified

haircommander:
#### What type of PR is this?

/kind bug

After we pull in the initial config, it is our source of truth for telling the world where our container storage is
in particular, if we, say, have an empty `storage` entry, we will return `StorageDriver: ""` on an Inspect request

the problem is, that is a lie. c/storage populates defaults for us, so we actually use overlay in that scenerio.

we don't encounter this normally, because if the toml field isn't specified (as it isn't by default), then we inherit the information from the default config
However, if a user (or an ignition config) overrides those fields to empty, we will listen to it as the source of truth, and never update to what is actually being used

Fix this by inheriting the values from c/storage on the validation step.

```release-note
Fix bug where empty config fields having to do with storage cause `/info` requests to return incorrect information (which causes cadvisor to fail to read imageFs information)
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -66,7 +66,7 @@ type Iface interface {
 }
 
 // GetStore returns the container storage for a given configuration
-func (c *Config) GetStore() (storage.Store, error) {
+func (c *RootConfig) GetStore() (storage.Store, error) {
 	return storage.GetStore(storage.StoreOptions{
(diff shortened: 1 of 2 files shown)


haircommander:

umohnani8:
LGTM

mrunalp (approved):

PR #4132, opened 2020-08-28
Disable the release branch forward script

umohnani8:
Looks like the master PRs are still being forwarded
to the release-1.19 branch even after we cut the rc.1
tag. Going to disable to script for now, will re-enable
when we create a branch for 1.20.

```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,13 +91,14 @@ workflows:
           name: integration-cgroupfs
           cgroup_manager: cgroupfs
       - lint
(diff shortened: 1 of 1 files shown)


haircommander:
for posterity, the reason we didn't disable it is because there are PRs that have an old version of the release-branch-forward script (and thus .circleci/config file). Merging this wouldn't solve the problem, and we'd need to require everyone to rebase to fully solve it. Luckily, we can just branch to 1.20, and have master forward PRs there, which we did

PR #4138, opened 2020-08-31
Preserve sandbox annotations for handling OCI hooks

saschagrunert:
#### What type of PR is this?

/kind bug

Some OCI hooks like the oci-seccomp-bpf-hook may re-use the sandbox
annotations which would get lost before applying this patch. This means
we now preserve all sandbox annotations for containers and overwrite
specific values during setup.

None
None

```release-note
- Fixed bug that all custom sandbox annotations will be passed to OCI hooks and therefore are also available on the containers
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -133,6 +133,13 @@ func (c *container) SpecAddAnnotations(sb *sandbox.Sandbox, containerVolumes []o
 	if err != nil {
 		return err
 	}
+
+	// Preserve the sandbox annotations. OCI hooks may re-use the sandbox
+	// annotation values to apply them to the container later on.
+	for k, v := range sb.Annotations() {
+		c.spec.AddAnnotation(k, v)
+	}
+
 	c.spec.AddAnnotation(annotations.Image, image)
 	c.spec.AddAnnotation(annotations.ImageName, imageResult.Name)
 	c.spec.AddAnnotation(annotations.ImageRef, imageResult.ID)


haircommander:
I am pretty sure this could break kata, as there is a specific annotation passed to kata that specifies whether it's an infra container or not. PTAL @fidencio @amshinde

saschagrunert:
> I am pretty sure this could break kata, as there is a specific annotation passed to kata that specifies whether it's an infra container or not. PTAL @fidencio @amshinde

I'd fear that something like this happens… :scream:

saschagrunert:
We could build something with reflection to check for unknown fields and sync them up, otherwise I’m not sure how to solve the issue appropriately.

haircommander:
theoretically, if we started the annotations based on the sandbox ones, and applies all the rest on top, it would override any container overrides, but leave the sandbox ones behind. maybe that could work?

fidencio:
> I am pretty sure this could break kata, as there is a specific annotation passed to kata that specifies whether it's an infra container or not. PTAL @fidencio @amshinde

@haircommander, @saschagrunert,

I've tested this one with the following reproducer ...
```
[root@k8s cri-o]# cat ~/reproducer.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: fio-test
  labels:
    role: fio-test
spec:
  containers:
    - name: fio
      image: dmonakhov/alpine-fio
      command: ["fio", "--name=random-writers-emptydir", "--filename=/test-volume/testfile", "--ioengine=libaio", "--iodepth=16", "--rw=randwrite", "--bs=512", "--direct=0", "--size=1G", "--numjobs=1", "--runtime=60s", "--group_reporting", "--fallocate=none", "--time_based", "--thread", "--invalidate=1", "--output-format=json"]
      volumeMounts:
        - name: test-volume
          mountPath: /test-volume
  initContainers:
    - name: fedora
      image: fedora
      command: ["date"]
    - name: centos
      image: centos
      command: ["date"]
  restartPolicy: Never
  runtimeClassName: kata
  volumes:
    - name: test-volume
      emptyDir: {}
```

... and I was able to spawn the pod without issues.

Let me know if there's any other scenario, maybe a more complex one, that I could try.

saschagrunert:
Thanks for the test @fidencio! :pray: @haircommander what do we wanna do now?

except…
/retest

TomSweeneyRedHat:
@saschagrunert codecov is complaining that it wants more coverage.

umohnani8:
Tests are passing and changes LGTM.
@saschagrunert @haircommander @fidencio are we good to get this in?

fidencio:
@umohnani8. from my side, yes! It didn't break kata-containers on a basic test I ran.

haircommander:
actually
/hold

I just realized we added annotations handling in pkg/container/ SpecAddAnnotations, can you maybe move this there? sorry for just remembering

saschagrunert:
> actually
> /hold
> 
> I just realized we added annotations handling in pkg/container/ SpecAddAnnotations, can you maybe move this there? sorry for just remembering

This comment has been slipped off… Rebased and addressed your comment. :+1:

haircommander:
none other than I forgot about it :)

PR #4142, opened 2020-08-31
Revert "runtime_vm: Cleanup process when the Container is Stopped"

fidencio:
#### What type of PR is this?

/kind bug

This reverts commit 814c1bb01e84d6e42cd01b5957cad6e441228fa5.

We cannot simply delete the container when its status changes to
stopped. The intention of this fix is okay-ish, but implemented in the
wrong layer.

What containerd does is calling Shutdown(), when the *pod* is finished,
and that's what we should do as well.

None

```release-note
None
```


haircommander:
LGTM, thanks!

saschagrunert (approved):
/lgtm

PR #4145, opened 2020-08-31
[release-1.17] Revert "runtime_vm: Cleanup process when the Container is Stopped"

fidencio:
#### What type of PR is this?

/kind bug

This reverts commit 67dcd05de1ea559008ff041729b390504a4b9d03.

We cannot simply delete the container when its status changes to
stopped. The intention of this fix is okay-ish, but implemented in the
wrong layer.

What containerd does is calling Shutdown(), when the *pod* is finished,
and that's what we should do as well.

None

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -276,12 +276,6 @@ func (r *runtimeVM) StartContainer(c *Container) error {
 			if err1 := r.updateContainerStatus(c); err1 != nil {
 				logrus.Warningf("error updating container status %v", err1)
 			}
-
-			if c.state.Status == ContainerStateStopped {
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM

TomSweeneyRedHat:
@fidencio you said we should call shutdown() when the pod is finished.  I don't see that change here, is it already present?

fidencio:
> @fidencio you said we should call shutdown() when the pod is finished. I don't see that change here, is it already present?

Tom, no, it's not. It'll be a follow up patch coming soon. Right now I just want to unblock the 1.17.5 release by reverting this commit.

PR #4146, opened 2020-08-31
oci: parse stat file instead of using ctime

haircommander:

#### What type of PR is this?

/kind bug

as we have seen the ctime of it's proc entry fluctuate, and the process start time (item 22 in /proc/pid/stat)
could be more reliable

this removes superfluous log lines like this: 
`0e06ea75835a28aee51b37751e51c6a6f4ac5a54fbe5b805d13c9184c075b74b: PID 3240 is running but has start time of 1598893538, whereas the saved start time is 1598893381. PID wrap may have occurred"` 

```release-note
none
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -1,7 +1,9 @@
 package oci
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -22,7 +24,14 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/types"
 )
 
-const defaultStopSignalInt = 15
+const (
+	defaultStopSignalInt = 15
+	// the following values can be verified here: https://man7.org/linux/man-pages/man5/proc.5.html
+	// the 22nd field is the process starttime
+	statStartTimeLocation = 22
+	// The 2nd field is the command, wrapped by ()
+	statCommField = 2
+)
 
 var (
 	defaultStopSignal   = strconv.Itoa(defaultStopSignalInt)
@@ -84,7 +93,7 @@ type ContainerState struct {
 	// The unix start time of the container's init PID.
 	// This is used to track whether the PID we have stored
 	// is the same as the corresponding PID on the host.
-	InitStartTime int `json:"initStartTime,omitempty"`
+	InitStartTime string `json:"initStartTime,omitempty"`
 }
 
 // NewContainer creates a container object.
@@ -182,11 +191,11 @@ func (c *Container) FromDisk() error {
 
 	// this is to handle the situation in which we're upgrading
 	// versions of cri-o, and we didn't used to have this information in the state
-	if tmpState.InitPid == 0 && tmpState.InitStartTime == 0 && tmpState.Pid != 0 {
+	if tmpState.InitPid == 0 && tmpState.InitStartTime == "" && tmpState.Pid != 0 {
 		if err := tmpState.SetInitPid(tmpState.Pid); err != nil {
 			return err
 		}
-		logrus.Infof("PID information for container %s updated to %d %d", c.id, tmpState.InitPid, tmpState.InitStartTime)
+		logrus.Infof("PID information for container %s updated to %d %s", c.id, tmpState.InitPid, tmpState.InitStartTime)
 	}
 	c.state = tmpState
 	return nil
@@ -196,8 +205,8 @@ func (c *Container) FromDisk() error {
 // given a PID.
 // These values should be set once, and not changed again.
 func (cstate *ContainerState) SetInitPid(pid int) error {
-	if cstate.InitPid != 0 || cstate.InitStartTime != 0 {
-		return errors.Errorf("pid and start time already initialized: %d %d", cstate.InitPid, cstate.InitStartTime)
+	if cstate.InitPid != 0 || cstate.InitStartTime != "" {
+		return errors.Errorf("pid and start time already initialized: %d %s", cstate.InitPid, cstate.InitStartTime)
 	}
 	cstate.InitPid = pid
 	startTime, err := getPidStartTime(pid)
@@ -438,7 +447,7 @@ func (c *Container) verifyPid() error {
 
 	if startTime != c.state.InitStartTime {
 		return errors.Errorf(
-			"PID %d is running but has start time of %d, whereas the saved start time is %d. PID wrap may have occurred",
+			"PID %d is running but has start time of %s, whereas the saved start time is %s. PID wrap may have occurred",
 			c.state.InitPid, startTime, c.state.InitStartTime,
 		)
 	}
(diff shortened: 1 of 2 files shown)


mrunalp:
@kolyshkin ptal

internal/oci/container.go
```diff
@@ -448,13 +456,18 @@ func (c *Container) verifyPid() error {
...
-	var st unix.Stat_t
-	if err := unix.Stat(fmt.Sprintf("/proc/%d", pid), &st); err != nil {
-		return 0, errors.Wrapf(ErrNotFound, err.Error())
+func getPidStartTime(pid int) (string, error) {
+	statContents, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/stat", pid))
+	if err != nil {
+		return "", errors.Wrapf(ErrNotFound, err.Error())
+	}
+
+	statFields := strings.Fields(string(statContents))
```
kolyshkin, line 465: 1. `strings.Fields` is a slow function. Using `strings.Split` is preferable.

2. Field 2 can contain spaces, so it's not that easy to parse; in fact I think it's not possible as there can be any characters in there:

```console
$ cp /bin/sleep 'ha ha ))) '
$ ./ha\ ha\ \)\)\)\  1h &
[1] 7397
$ cut -c-50 /proc/7397/stat
7397 (ha ha ))) ) S 3859 7397 3859 34816 7412 1077
```

In case it's a user-controlled process (and I believe it is), it can change its name freely.


3. Since you do not need all the fields, it is probably better to employ a faster algo -- get past the field 2 (which is cumbersome), skip a certain number of spaces, get the value.

saschagrunert: Alternatively, we could also retrieve the program name for the pid in a different command and strip it away. But this seems slower in that case. :shrug:

TomSweeneyRedHat:
LGTM once Kir's comment is addressed and the tests get happy.

internal/oci/container.go
haircommander: luckily, it's actually fairly simple, because the command is always wrapped in a set of parenthesis

haircommander:
ptal @mrunalp @saschagrunert @kolyshkin @umohnani8

internal/oci/container.go
```diff
@@ -448,13 +460,30 @@ func (c *Container) verifyPid() error {
...
+// inspiration for this function came from https://github.com/containers/psgo/blob/master/internal/proc/stat.go
+// some credit goes to the psgo authors
+func getPidStartTime(pid int) (string, error) {
+	data, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/stat", pid))
+	if err != nil {
+		return "", errors.Wrapf(ErrNotFound, err.Error())
+	}
+	// The command (2nd field) is wrapped in ().
+	// Filter it out
+	commEnd := strings.LastIndexByte(string(data), ')')
```
kolyshkin, line 472: this is the correct way to do that :+1:

internal/oci/container.go
```diff
@@ -448,13 +460,30 @@ func (c *Container) verifyPid() error {
...
+	// commEnd now points to the ending parenthesis.
+	// After it, there is a space, and then the next field starts
+	statFieldsAfterComm := strings.Split(string(data[commEnd+2:]), " ")
+	if len(statFieldsAfterComm) != statTotalFields-statCommField {
+		return "", errors.Wrapf(ErrNotFound, "invalid number of entries found in stat file %d %v", len(statFieldsAfterComm), statFieldsAfterComm)
 	}
 
-	return int(st.Ctim.Sec), nil
+	// and return the 20th
+	// because StartTime is the 22nd field, and we removed the first 2
```
kolyshkin, line 485: You have abstracted these numbers (22 and 2) to constants and then you're hardcoding them here. Yes, it's a comment, but it's still hardcoding :)

internal/oci/container.go
kolyshkin: but it's better to use `bytes.LastIndexByte` as you don't have to convert to the string.

internal/oci/container.go
```diff
@@ -448,13 +460,30 @@ func (c *Container) verifyPid() error {
...
+	// The command (2nd field) is wrapped in ().
+	// Filter it out
+	commEnd := strings.LastIndexByte(string(data), ')')
+	if commEnd == -1 {
+		return "", errors.Wrapf(ErrNotFound, "invalid format in stat file")
+	}
+
+	// commEnd now points to the ending parenthesis.
+	// After it, there is a space, and then the next field starts
+	statFieldsAfterComm := strings.Split(string(data[commEnd+2:]), " ")
```
kolyshkin, line 479: I suggest you don't split the whole string, but count 20 spaces from the position you found and then just take the number. Otherwise you're creating a slice of 50+ fields of which 49 you won't need.

haircommander: done!

internal/oci/container.go

internal/oci/container.go
haircommander: done, and I changed it all to use []byte

haircommander:
updated as suggested and added unit tests

haircommander:
PTAL @kolyshkin @saschagrunert @mrunalp @umohnani8

internal/oci/container.go
```diff
@@ -448,13 +457,41 @@ func (c *Container) verifyPid() error {
...
+	commEnd := bytes.LastIndexByte(data, ')')
+	if commEnd == -1 {
+		return "", errors.Wrapf(ErrNotFound, "invalid format in stat file")
+	}
+
+	// start on the space after the command
+	iter := commEnd + 1
+	// for the number of fields between command and stime, trim the beginning word
+	for field := 0; field < statStartTimeLocation-statCommField; field++ {
+		// trim from the beginning to the character after the last spac
```
mrunalp, line 484: nit: `spac` -> `space`

internal/oci/container.go
```diff
@@ -446,13 +455,41 @@ func (c *Container) verifyPid() error {
...
+	}
+
+	// start on the space after the command
+	iter := commEnd + 1
+	// for the number of fields between command and stime, trim the beginning word
+	for field := 0; field < statStartTimeLocation-statCommField; field++ {
+		// trim from the beginning to the character after the last space
+		data = data[iter+1:]
+		// find the next space
+		iter = bytes.IndexRune(data, ' ')
```
kolyshkin, line 485: No need to use `IndexRune` here as we do not expect any Unicode in the input.

kolyshkin:
Looks much better now, just a nit about using `bytes.IndexRune()` instead of `bytes.Index()`

kolyshkin:
> instead of `bytes.Index()`

or, yet better, `bytes.IndexByte()`

internal/oci/container.go
```diff
@@ -446,13 +455,41 @@ func (c *Container) verifyPid() error {
...
+func GetPidStartTimeFromFile(file string) (string, error) {
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		return "", errors.Wrapf(ErrNotFound, err.Error())
+	}
+	// The command (2nd field) can have spaces, but is wrapped in ()
+	// first, trim it
+	commEnd := bytes.LastIndexByte(data, ')')
+	if commEnd == -1 {
+		return "", errors.Wrapf(ErrNotFound, "invalid format in stat file")
```
saschagrunert, line 475:
```suggestion
		return "", errors.Wrapf(ErrNotFound, "unable to find ')' in stat file")
```

saschagrunert (approved):
Just a nit, otherwise LGTM

internal/oci/container.go
```diff
@@ -182,11 +191,11 @@ func (c *Container) FromDisk() error {
 
 	// this is to handle the situation in which we're upgrading
 	// versions of cri-o, and we didn't used to have this information in the state
-	if tmpState.InitPid == 0 && tmpState.InitStartTime == 0 && tmpState.Pid != 0 {
+	if tmpState.InitPid == 0 && tmpState.InitStartTime == "" && tmpState.Pid != 0 {
 		if err := tmpState.SetInitPid(tmpState.Pid); err != nil {
 			return err
 		}
-		logrus.Infof("PID information for container %s updated to %d %d", c.id, tmpState.InitPid, tmpState.InitStartTime)
+		logrus.Infof("PID information for container %s updated to %d %s", c.id, tmpState.InitPid, tmpState.InitStartTime)
```
TomSweeneyRedHat, line 198: maybe pre-tea time, but won't tmpState.InitStartTime always be "" here?  Is that useful in the message?

haircommander: SetInitPid updates the two values for tmpState :)

TomSweeneyRedHat:
One question
LGTM otherwise

saschagrunert (approved):
LGTM

kolyshkin (approved):
LGTM

Issue #4147, opened 2020-09-01
pinns failure diagnostics is not adequate


kolyshkin:
**Description**

Pinns failure diagnostics is not adequate. Here's an excerpt from logs:

```
# time="2020-08-31 16:48:27.990804565-07:00" level=warning msg="Failed to find memory.max for slice: \"pod_123-456.slice\"" file="cgmgr/systemd.go:117"
# time="2020-08-31 16:48:27.990879645-07:00" level=debug msg="Calling pinns with [-d /var/run -f 034cf2f1-b18e-4853-95b0-05b8421c67b4 -n -i -u]" file="sandbox/namespaces_linux.go:127"
# time="2020-08-31 16:48:27.993261284-07:00" level=info msg="RunSandbox: cleaning up namespaces after failing to run sandbox 45ff147e5b3db805d255594f0f597eef84e194c86458b230c03a56b795d0f98c" file="server/sandbox_run_linux.go:681" id=d6035872-bc9d-4d98-b3b4-b1e036c7546d name=/runtime.v1alpha2.RuntimeService/RunPodSandbox
# time="2020-08-31 16:48:27.993295622-07:00" level=info msg="RunSandbox: deleting pod ID 45ff147e5b3db805d255594f0f597eef84e194c86458b230c03a56b795d0f98c from idIndex" file="server/sandbox_run_linux.go:648" id=d6035872-bc9d-4d98-b3b4-b1e036c7546d name=/runtime.v1alpha2.RuntimeService/RunPodSandbox
# time="2020-08-31 16:48:27.993328819-07:00" level=info msg="RunSandbox: removing pod sandbox 45ff147e5b3db805d255594f0f597eef84e194c86458b230c03a56b795d0f98c" file="server/sandbox_run_linux.go:635" id=d6035872-bc9d-4d98-b3b4-b1e036c7546d name=/runtime.v1alpha2.RuntimeService/RunPodSandbox
# time="2020-08-31 16:48:27.993358646-07:00" level=info msg="RunSandbox: deleting container ID from idIndex for sandbox 45ff147e5b3db805d255594f0f597eef84e194c86458b230c03a56b795d0f98c" file="server/sandbox_run_linux.go:536" id=d6035872-bc9d-4d98-b3b4-b1e036c7546d name=/runtime.v1alpha2.RuntimeService/RunPodSandbox
[...]

PR #4148, opened 2020-09-01
Fix pinns error diagnostics, cleanup, and return

kolyshkin:
/kind bug

1. In case pinns fails, we want to return a clear error message saying
       it has failed, not that we failed to unmount some mounts. So, do not
       return the "failed to unmount" error as it's minor in nature, and log
       the failures instead.
    
2. In case pinns fails, we want to see its stderr, not just stdout.
       Before this patch, only stdout is collected (and now shown as it's empty).
    
3. We can't be sure that the mounts we are trying to unmount were
       actually mounted, so ignore EINVAL from unmount(2). EINVAL means
       that either argument is not the mount point (i.e. not mounted),
       or that the flags are wrong. We only provide MNT_DETACH which
       is not a wrong flag, ergo EINVAL means "not mounted" and it should
       be ignored.
    
4. Only log pinns output in case of error, and do it as a warning since
       it is pretty serious.

While at it, make some minor improvements around the code we're touching.

Fixes: https://github.com/cri-o/cri-o/issues/4147

```release-note
NONE
```

### DIFF
--- a/internal/lib/sandbox/namespaces_linux.go
+++ b/internal/lib/sandbox/namespaces_linux.go
@@ -143,7 +143,7 @@ func pinNamespaces(nsTypes []NSType, cfg *config.Config, idMappings *idtools.IDM
 		return nil, fmt.Errorf("failed to pin namespaces %v: %s %v", nsTypes, output, err)
 	}
 
-	returnedNamespaces := make([]NamespaceIface, 0)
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests.

PR #4150, opened 2020-09-02
Bump cri-tools to v1.19.0, CNI plugins and golang

saschagrunert:
#### What type of PR is this?

/kind dependency-change

This updates the test scripts and CI to their latest versions.

None
None

```release-note
None
```


saschagrunert:
Something is broken here.

saschagrunert:
:green_circle: tests now, PTAL @haircommander @mrunalp

test/helpers.bash
```diff
@@ -217,6 +217,9 @@ function setup_test() {
     cp "$INTEGRATION_ROOT"/cni_plugin_helper.bash "$CRIO_CNI_PLUGIN"
     sed -i "s;%TEST_DIR%;$TESTDIR;" "$CRIO_CNI_PLUGIN"/cni_plugin_helper.bash
 
+    # configure crictl globally
+    crictl config --set pull-image-on-create=true || true
```
kolyshkin, line 221: @saschagrunert have you entertained the possibility of leaving this at default (false)? We already pre-fetch and load a bunch of images during `setup_test`, and I think we can amend a few tests with `crictl pull` in `setup_test`. This will solve the problem of flakes caused by intermittent network issues (even if the image is there, it is re-checked on create and sometimes it fails).

saschagrunert: If we pre-fetch all images used in the tests, yeah then we could change this to point to the defaults.

kolyshkin: https://github.com/cri-o/cri-o/pull/4311

PR #4158, opened 2020-09-03
Print seccomp profile JSON only on trace log level

saschagrunert:
#### What type of PR is this?

/kind cleanup

To sync up the behavior with AppArmor, we now print the content of the
seccomp profile only on trace log level. Beside that, we print the JSON
instead of the struct to provide more information to the user.

None
None

```release-note
- Changed the output of the printed seccomp profile to JSON instead of the struct. The profile will be only printed on CRI-O startup and if the `--log-level`/`log_level` is set to `trace`.
```


haircommander:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #3997, opened 2020-07-21
WIP: drop the infra container (again!)

haircommander:
#### What type of PR is this?

/kind api-change

#### What this PR does / why we need it:
this is a https://github.com/cri-o/cri-o/pull/3474 redux, built for the modern era! it also contains https://github.com/cri-o/cri-o/pull/3868 and https://github.com/cri-o/cri-o/pull/3910

With this PR, users can specify `--drop-infra` and cri-o will drop the infra container when able.

#### Does this PR introduce a user-facing change?

```release-note
add `--drop-infra` option to ask CRI-O to drop the infra container when a pod level pid namespace isn't requested. This feature is considered experimental
```

PR #4165, opened 2020-09-04
Add option to drop infra container

haircommander:
#### What type of PR is this?

 /kind feature

     Sandbox: Don't use an infra container in some cases
    
    some changes this includes:
    - allow for the infra container to not be created by the runtime, if sb.NeedsInfra() is false
    - add concept of a 'spoofed' container. This allows us to make calls to ID(), Name(), and Description() without having the runtime create the container
    - using spoof, change checks of podInfraContainer == nil to Spoofed(), as well as re-add Description() calls
    - reorder some infra container creation code
    - add unit tests for NeedsInfra and Spoofed
    - add integration test for drop infra
    - sysctls are configured in the runtime after the process unshares in the shared namespaces. since we aren't using a runtime, we need to repeat this step.
            - After unsharing, but before binding, configure the sysctls for the namespace.
    - spoof container state
    - create a new annotations package for cri-o specific annotations that are experimental, and not yet locked in the API
    - refactor ContainerRuntimeType to get a runtime type without a container (and call it RuntimeType)

```release-note
add `--drop-infra-ctr` option to ask CRI-O to drop the infra container when a pod level pid namespace isn't requested. This feature is considered experimental

```


haircommander:
replaces https://github.com/cri-o/cri-o/pull/3997 (which has grown out of date, and has an extra patch that's not *needed*)

haircommander:
also carries https://github.com/cri-o/cri-o/pull/4146

internal/criocli/criocli.go
```diff
@@ -248,6 +248,9 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
 	if ctx.IsSet("manage-ns-lifecycle") {
 		config.ManageNSLifecycle = ctx.Bool("manage-ns-lifecycle")
 	}
+	if ctx.IsSet("drop-infra") {
```
mrunalp, line 251: Wondering if calling this `drop-infra-container` or `drop-infra-ctr` is better or leave it as is.

haircommander: yeah `drop-infra-ctr` is more clear... I may make that change on next rebase (the only other field in the config that has container or ctr is `ctr_stop_timeout` so we should keep with `ctr` imo)

internal/criocli/criocli.go
```diff
@@ -725,6 +728,11 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			Usage:   fmt.Sprintf("Determines whether we pin and remove IPC, network and UTS namespaces and manage their lifecycle (default: %v)", defConf.ManageNSLifecycle),
 			EnvVars: []string{"CONTAINER_MANAGE_NS_LIFECYCLE"},
 		},
+		&cli.BoolFlag{
+			Name:    "drop-infra-ctr",
+			Usage:   fmt.Sprintf("determines whether we drop the infra container when appropriate. Requires ManageNSLifecycle to be true (default: %v)", defConf.DropInfraCtr),
```
mrunalp, line 733: instead of when appropriate should we reword to `when pod pid namespace is not shared`?

haircommander: I thought it a bit verbose for a cli flag TBH but yeah it would better match the config values

internal/lib/container_server.go
```diff
@@ -468,7 +503,7 @@ func (c *ContainerServer) ContainerStateFromDisk(ctr *oci.Container) error {
 // ContainerStateToDisk writes the container's state information to a JSON file
 // on disk
 func (c *ContainerServer) ContainerStateToDisk(ctr *oci.Container) error {
-	if ctr == nil {
+	if ctr.Spoofed() {
```
mrunalp, line 506: Do we need a skip in `ContainerStateFromDisk` as well?

internal/lib/sandbox/namespaces_linux.go
```diff
@@ -77,6 +78,11 @@ func pinNamespaces(nsTypes []NSType, cfg *config.Config, idMappings *idtools.IDM
 		"-d", cfg.NamespacesDir,
 		"-f", pinnedNamespace,
 	}
+
+	if len(sysctls) != 0 {
+		pinnsArgs = append(pinnsArgs, "-s", getSysctlForPinns(sysctls))
```
mrunalp, line 83: Should we worry about cases where we exceed the command line size? If so, should we pass this as temp file like we do for runc exec?

internal/lib/sandbox/sandbox.go
```diff
@@ -417,3 +422,10 @@ func (s *Sandbox) UnmountShm() error {
 
 	return nil
 }
+
+// NeedsInfra is a function that returns whether the sandbox will need an infra container
```
mrunalp, line 426: nit: Add sentence with period.

internal/lib/sandbox/sandbox.go
```diff
@@ -417,3 +422,10 @@ func (s *Sandbox) UnmountShm() error {
...
+// if the server manages the namespace lifecycles, and the Pid option on the sandbox
```
mrunalp, line 427: nit: if -> If

internal/lib/sandbox/sandbox.go
```diff
@@ -417,3 +422,10 @@ func (s *Sandbox) UnmountShm() error {
...
+// is node level, the infra container is not needed
```
mrunalp, line 428: node -> pod

internal/lib/sandbox/namespaces_linux.go
haircommander:
```
getconf ARG_MAX
2097152
```
```
sudo sysctl -a |  awk  '{ print $1 }' | wc 
   1884    1884   65129
```
of course, that doesn't account for their actual args, but since its an entire order of magnitude lower, I am not super inclined to worry

internal/lib/sandbox/sandbox.go
mrunalp, line 428: node or container

internal/oci/container.go
```diff
@@ -234,14 +254,14 @@ func (c *Container) Name() string {
...
 func (c *Container) ID() string {
-	if c == nil {
-		return ""
-	}
 	return c.id
 }
 
 // CleanupConmonCgroup cleans up conmon's group when using cgroupfs.
 func (c *Container) CleanupConmonCgroup() {
+	if c.Spoofed() {
```
mrunalp, line 262: this can be `c.spoofed`

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+static int separate_sysctl_key_value (char* sysctl_key_value, char** sysctl_key, char** sysctl_value);
+static int write_sysctl_to_file (char * sysctl_key, char* sysctl_value);
+const char *sysctl_delim = ":";
+
+int configure_sysctls (char * const sysctls) {
+  char* sysctl = strtok(sysctls, sysctl_delim);
+  char* key = NULL;
+  char* value = NULL;
+  while (sysctl) {
+    if (separate_sysctl_key_value (sysctl, &key, &value) < 0)
```
mrunalp, line 18: We need to be consistent in the space between function call and arguments. I favor no spaces.

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+  }
+  if (!strlen (*sysctl_value))
+  {
+    nwarnf ("sysctl must be in the form of 'key=value'; value is empty");
+    return -1;
+  }
+  return 0;
+}
+
+static int write_sysctl_to_file (char * sysctl_key, char* sysctl_value)
```
mrunalp, line 74: nit: remove space.

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+    pwarnf ("sysctl key or value not initialized");
+    return -1;
+  }
+
+  // replace periods with / to create the sysctl path
+  for (char* it = sysctl_key; *it; it++)
+    if (*it == '.')
+      *it = '/';
+
+  _cleanup_close_ int dirfd = open ("/proc/sys", O_DIRECTORY | O_RDONLY);
```
mrunalp, line 87: nit: remove space

server/sandbox_network.go
```diff
@@ -110,6 +110,7 @@ func (s *Server) getSandboxIPs(sb *sandbox.Sandbox) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	log.Debugf(context.Background(), "%v", podNetwork)
```
mrunalp, line 113: drop or move to a different PR?

mrunalp:
@saschagrunert @kolyshkin @umohnani8 @harche ptal

internal/criocli/criocli.go
```diff
@@ -725,6 +728,11 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			Usage:   fmt.Sprintf("Determines whether we pin and remove IPC, network and UTS namespaces and manage their lifecycle (default: %v)", defConf.ManageNSLifecycle),
 			EnvVars: []string{"CONTAINER_MANAGE_NS_LIFECYCLE"},
 		},
+		&cli.BoolFlag{
+			Name:    "drop-infra",
+			Usage:   fmt.Sprintf("determines whether we drop the infra container when appropriate. Requires ManageNSLifecycle to be true (default: %v)", defConf.DropInfra),
```
kolyshkin: > Requires ManageNSLifecycle to be true (default: %v)", defConf.DropInfra)

As a user, I would think that `(default: ...)` prints the value of `ManageNSLifecycle`, not the DropInfra.

internal/lib/container_server.go
```diff
@@ -203,7 +200,7 @@ func (c *ContainerServer) LoadSandbox(id string) (retErr error) {
 		return errors.Wrap(err, "parsing created timestamp annotation")
 	}
 
-	sb, err := sandbox.New(id, m.Annotations[annotations.Namespace], name, m.Annotations[annotations.KubeName], filepath.Dir(m.Annotations[annotations.LogPath]), labels, kubeAnnotations, processLabel, mountLabel, &metadata, m.Annotations[annotations.ShmPath], m.Annotations[annotations.CgroupParent], privileged, m.Annotations[annotations.RuntimeHandler], m.Annotations[annotations.ResolvPath], m.Annotations[annotations.HostName], portMappings, hostNetwork, created, m.Annotations[UsernsModeAnnotation])
+	sb, err := sandbox.New(id, m.Annotations[annotations.Namespace], name, m.Annotations[annotations.KubeName], filepath.Dir(m.Annotations[annotations.LogPath]), labels, kubeAnnotations, processLabel, mountLabel, &metadata, m.Annotations[annotations.ShmPath], m.Annotations[annotations.CgroupParent], privileged, m.Annotations[annotations.RuntimeHandler], m.Annotations[annotations.ResolvPath], m.Annotations[annotations.HostName], portMappings, hostNetwork, created, m.Annotations[crioann.UsernsModeAnnotation])
```
kolyshkin, line 203: Not a showstopper, but I'd rather see this as a separate patch/commit

internal/lib/container_server.go
```diff
@@ -272,40 +271,76 @@ func (c *ContainerServer) LoadSandbox(id string) (retErr error) {
 		}
 	}()
 
-	scontainer, err := oci.NewContainer(m.Annotations[annotations.ContainerID], cname, sandboxPath, m.Annotations[annotations.LogPath], labels, m.Annotations, kubeAnnotations, m.Annotations[annotations.Image], "", "", nil, id, false, false, false, sb.RuntimeHandler(), sandboxDir, created, m.Annotations["org.opencontainers.image.stopSignal"])
-	if err != nil {
-		return err
+	var scontainer *oci.Container
+
+	// we should not take whether the server actively has DropInfra specified, but rather
```
kolyshkin: `s/actively/currently/`?

internal/lib/container_server.go
kolyshkin: Sorry I didn't see you're introducing `pkg/annotations` in the same commit.

test/network.bats
```diff
@@ -127,6 +127,9 @@ function teardown() {
 }
 
 @test "Clean up network if pod sandbox fails" {
+	# TODO FIXME find a way for sandbox setup to fail if manage ns is true
+	export CONTAINER_MANAGE_NS_LIFECYCLE=false
+	export CONTAINER_DROP_INFRA=false
```
kolyshkin, line 132: AFAIK there's no need for `export` here.

test/drop_infra.bats
```diff
@@ -0,0 +1,40 @@
...
+	CONTAINER_MANAGE_NS_LIFECYCLE=true start_crio
+}
+
+function teardown() {
+	cleanup_test
+}
+
+@test "test infra ctr dropped" {
+	temp_config=$(cat "$TESTDATA"/sandbox_config.json | python -c 'import json,sys;obj=json.load(sys.stdin);obj["linux"]["security_context"]["namespace_options"]["pid"] = 1; json.dump(obj, sys.stdout)')
+	echo "$temp_config" > "$TESTDIR"/sandbox_no_infra.json
```
kolyshkin, line 16:
```bash
python -c 'import json,sys;obj=json.load(sys.stdin);obj["linux"]["security_context"]["namespace_options"]["pid"] = 1; json.dump(obj, sys.stdout)' \
   < "$TESTDATA"/sandbox_config.json > "$TESTDIR"/sandbox_no_infra.json
```

test/drop_infra.bats
```diff
@@ -0,0 +1,40 @@
...
+	run crictl runp "$TESTDIR"/sandbox_no_infra.json
+	echo "$output"
+	[ "$status" -eq 0 ]
+	pod_id="$output"
```
kolyshkin, line 20:
```bash
pod_id=$(crictl runp "$TESTDIR"/sandbox_no_infra.json)
```
will have the same effect.

test/drop_infra.bats
```diff
@@ -0,0 +1,40 @@
...
+
+	run "$CONTAINER_RUNTIME" list
+	echo "$output"
+	[ "$status" -eq 0 ]
+	[[ ! "$output" =~ "$pod_id" ]]
```
kolyshkin, line 25:
```bash
[[ ! "$output" = *"$pod_id"* ]]
```

test/drop_infra.bats
```diff
@@ -0,0 +1,40 @@
...
+}
+
+@test "test infra ctr not dropped" {
+	temp_config=$(cat "$TESTDATA"/sandbox_config.json | python -c 'import json,sys;obj=json.load(sys.stdin);obj["linux"]["security_context"]["namespace_options"]["pid"] = 0; json.dump(obj, sys.stdout)')
+	echo "$temp_config" > "$TESTDIR"/sandbox_no_infra.json
+	run crictl runp "$TESTDIR"/sandbox_no_infra.json
+	echo "$output"
+	[ "$status" -eq 0 ]
+	pod_id="$output"
```
kolyshkin, line 34: same as above

test/drop_infra.bats
```diff
@@ -0,0 +1,40 @@
...
+
+	run "$CONTAINER_RUNTIME" list
+	echo "$output"
+	[ "$status" -eq 0 ]
+	[[ "$output" =~ "$pod_id" ]]
```
kolyshkin, line 39:
```bash
[[ "$output" == *"$pod_id"* ]]
```

(operator `=~` is used for regular expression matching, and `$pod_id` is definitely not a regular expression

server/sandbox_run_linux.go
```diff
@@ -826,76 +848,67 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
...
-	if err := s.createContainerPlatform(container, sb.CgroupParent(), sandboxIDMappings); err != nil {
-		return nil, err
-	}
 
-	if err := s.Runtime().StartContainer(container); err != nil {
-		return nil, err
-	}
+		if err := s.Runtime().StartContainer(container); err != nil {
+			return nil, err
+		}
```
kolyshkin: What caused this whitespace change? (I can't see it from github if it is correct or not)

pkg/config/config.go
```diff
@@ -775,6 +780,9 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 	if c.GIDMappings != "" && c.ManageNSLifecycle {
 		return fmt.Errorf("cannot use GIDMappings with ManageNSLifecycle")
 	}
+	if c.DropInfra && !c.ManageNSLifecycle {
+		return fmt.Errorf("cannot drop infra without ManageNSLifecycle")
```
kolyshkin, line 784: `s/fmt.Errorf/errors.New/`

pinns/src/pinns.c
```diff
@@ -83,12 +86,15 @@ int main(int argc, char **argv) {
       unshare_flags |= CLONE_NEWCGROUP;
       bind_cgroup = true;
       num_unshares++;
+      break;
```
kolyshkin, line 89: This seems like a real fix. If so, it deserves a separate commit.

kolyshkin (changes requested):
Might be good to split this into more reviewable pieces, such as
 - pinns: add sysctl support
 - pkg/annotations: add
etc.

internal/lib/sandbox/namespaces_linux.go
```diff
@@ -152,6 +158,14 @@ func pinNamespaces(nsTypes []NSType, cfg *config.Config, idMappings *idtools.IDM
 	return returnedNamespaces, nil
 }
 
+func getSysctlForPinns(sysctls map[string]string) string {
+	g := new(bytes.Buffer)
+	for key, value := range sysctls {
+		fmt.Fprintf(g, "'%s=%s':", key, value)
+	}
+	return strings.TrimSuffix(g.String(), ":")
```
kolyshkin, line 166: Do we assume that there are no values with `:` or `=` in them? Should we actually check it (not here, but maybe when parsing annotations).

internal/lib/container_server.go
```diff
@@ -272,40 +271,76 @@ func (c *ContainerServer) LoadSandbox(id string) (retErr error) {
...
+		if err := c.ContainerStateFromDisk(scontainer); err != nil {
+			return fmt.Errorf("error reading sandbox state from disk %q: %v", scontainer.ID(), err)
 		}
-		for _, cv := range containerVolumes {
-			scontainer.AddVolume(cv)
+
+		// We write back the state because it is possible that crio did not have a chance to
+		// read the exit file and persist exit code into the state on reboot.
+		if err := c.ContainerStateToDisk(scontainer); err != nil {
+			return fmt.Errorf("failed to write container state to disk %q: %v", scontainer.ID(), err)
```
kolyshkin, line 308: maybe put container ID after the word "container"? As in `failed to write container %q state to disk...`

internal/oci/oci_linux.go
```diff
@@ -50,6 +53,10 @@ func (r *runtimeOCI) containerStats(ctr *Container, cgroup string) (*ContainerSt
 	stats.Container = ctr.ID()
 	stats.SystemNano = time.Now().UnixNano()
 
+	if ctr.Spoofed() {
```
mrunalp, line 56: Not a change request but a thought that we shouldn't be queried about container stats from the kubelet.

internal/oci/runtime_oci.go
```diff
@@ -630,6 +650,12 @@ func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout in
 		return err
 	}
 
+	if c.Spoofed() {
+		c.state.Status = ContainerStateStopped
+		c.state.Finished = time.Now()
```
mrunalp, line 655: Do we guard at upper level so that we don't update Finished once it is set?

internal/lib/container_server.go
haircommander: we can for posterity, but the only time it's called is when restoring, in a block where we don't restore if the container used to be spoofed.

internal/lib/sandbox/namespaces_linux.go
haircommander: that's a great point. a quick `sudo sysctl -a | fold -w1 | sort -u` shows me some with `:`, though `=` is a special character in sysctl land, so that should be allowed
I've opted to deliminate them with `+` (not a special bash character, not included in the characters of sysctls on my machine). No spot seems right to find `+` though, we never traverse the whole string, mostly pass it from place to place

internal/oci/runtime_oci.go
haircommander: above this we check ShouldBeStopped(), which verifies the Status is not ContainerStateStopped

pinns/src/sysctl.c
haircommander: I agree, but @giuseppe introduced the convention. I'll add all to have spaces, and the three of us can bicker about it when he's back :)

server/sandbox_run_linux.go
haircommander: this does look weird, but I believe it's fixed in a later commit (which I will be pushing a version that's been squashed soon)

pinns/src/pinns.c
haircommander: good point, done

haircommander:
comments addressed. There are two TODOs before merging this:
- [ ] by default drop_infra_ctr is true, it should be false
- [ ] https://github.com/kata-containers/tests/pull/2834 needs resolving

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,10 @@
+package annotations
+
+const (
+	// UsernsMode is the user namespace mode to use
+	UsernsModeAnnotation = "io.kubernetes.cri-o.userns-mode"
+
+	// SpoofedContainer indicates a container was spoofed in the runtime
+	SpoofedContainer = "io.kubernetes.cri-o.Spoofed"
+
```
mrunalp, line 9: nit: extra space

internal/oci/runtime_oci.go
mrunalp: What happens if crio restarts after this point?

haircommander: that's a good point.

In this version, I've allowed spoofed containers to have their state written and read from disk in ContainerState{To,From}Disk. that will allow the appropriate state to be preserved across cri-o restarts

pkg/annotations/annotations.go
haircommander: fixed!

docs/crio.8.md
```diff
@@ -170,6 +171,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--default-ulimits**="": Ulimits to apply to containers by default (name=soft:hard) (default: []) (default: [])
 
+**--drop-infra-ctr**: determines whether we drop the infra container when appropriate. Requires ManageNSLifecycle to be true (default: true)
```
TomSweeneyRedHat, line 174: it would be nice to have a bit more detail to what "when appropriate" means.

haircommander: fixed!

pinns/src/pinns.c
haircommander: done

server/sandbox_run_linux.go
```diff
@@ -780,14 +752,66 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
...
+	}
+
+	runtimeType, err := s.Runtime().RuntimeType(runtimeHandler)
+	if err != nil {
+		return nil, err
+	}
+
+	// A container is kernel separated if we're using shimv2, or we're using a kata v1 binary
+	podIsKernelSeparated := runtimeType == libconfig.RuntimeTypeVM ||
```
haircommander, line 785: cc @amshinde does this condition look good? we can't strictly rely on runtime vm because we need the infra container for kata v1

docs/crio.conf.5.md
```diff
@@ -222,6 +222,10 @@ the container runtime configuration.
 **manage_ns_lifecycle**=true
   Determines whether we pin and remove namespaces and manage their lifecycle.
 
+**drop_infra_ctr**=false
+  Determines whether we drop the infra container when a pod does not have a private PID namespace, and it is not a kernel separated container.
+  Requies manage_ns_lifecycle to be true
```
kolyshkin, line 227: nit: `s/$/./` (missing `.` at EOL).

kolyshkin: nit: `s/manage_ns_lifecycle/**manage_ns_lifecycle**/`

kolyshkin: > a kernel separated container

I'm not sure I know what that means :thinking:

pkg/config/config.go
```diff
@@ -284,6 +284,10 @@ type RuntimeConfig struct {
 	// and manage their lifecycle
 	ManageNSLifecycle bool `toml:"manage_ns_lifecycle"`
 
+	// DropInfraCtr determines whether the infra container is dropped when appropriate
+	// Requires ManageNSLifecycle to be true
```
kolyshkin, line 288: missing `.` at EOL

pkg/config/config.go
```diff
@@ -284,6 +284,10 @@ type RuntimeConfig struct {
 	// and manage their lifecycle
 	ManageNSLifecycle bool `toml:"manage_ns_lifecycle"`
 
+	// DropInfraCtr determines whether the infra container is dropped when appropriate
```
kolyshkin, line 287: missing . at EOL

pkg/config/config.go
```diff
@@ -580,6 +584,7 @@ func DefaultConfig() (*Config, error) {
 			HooksDir:                 []string{hooks.DefaultDir},
 			NamespacesDir:            "/var/run",
 			ManageNSLifecycle:        true,
+			DropInfraCtr:             false,
```
kolyshkin, line 587: afaik you don't have to set a value if it's a default one for this type (for boolean the default is false).

kolyshkin: ...unless you're doing it for readability.

pkg/config/template.go
```diff
@@ -257,6 +257,12 @@ ctr_stop_timeout = {{ .CtrStopTimeout }}
 # and manage their lifecycle
 manage_ns_lifecycle = {{ .ManageNSLifecycle }}
 
+# drop_infra_ctr determines whether CRI-O drops the infra container
+# in cases where a pod doesn't have a pod level PID namespace, and when
+# the requested runtime is not running kernel separated containers.
+# It requires manage_ns_lifecycle to be true
```
kolyshkin, line 263: missing `.` at EOL

test/cgroups.bats
```diff
@@ -28,7 +28,7 @@ function teardown() {
 }
 
 @test "conmon custom cgroup" {
-	CONTAINER_CGROUP_MANAGER="systemd" CONTAINER_CONMON_CGROUP="customcrioconmon.slice" start_crio
+	CONTAINER_CGROUP_MANAGER="systemd" CONTAINER_DROP_INFRA_CTR=false CONTAINER_MANAGE_NS_LIFECYCLE=false CONTAINER_CONMON_CGROUP="customcrioconmon.slice" start_crio
```
kolyshkin, line 31: since `false` is the default for these two, why set them?

test/ctr_userns.bats
```diff
@@ -3,6 +3,8 @@
 load helpers
 
 function setup() {
+	export CONTAINER_MANAGE_NS_LIFECYCLE=false
+	export CONTAINER_DROP_INFRA_CTR=false
```
kolyshkin, line 7: ditto

test/network.bats
```diff
@@ -127,9 +127,12 @@ function teardown() {
 }
 
 @test "Clean up network if pod sandbox fails" {
+	# TODO FIXME find a way for sandbox setup to fail if manage ns is true
 	cp $(which conmon) "$TESTDIR"/conmon
-	CONMON_BINARY="$TESTDIR"/conmon
-	start_crio "" "prepare_plugin_test_args_network_conf"
+	CONTAINER_MANAGE_NS_LIFECYCLE=false \
+		CONTAINER_DROP_INFRA_CTR=false \
```
kolyshkin, line 133: Since false is the default for these two, why set them?

I suspect I'm missing something...

kolyshkin: same question to a few cases below.

test/helpers.bash
```diff
@@ -298,6 +298,11 @@ function setup_crio() {
         CNI_DEFAULT_NETWORK="crio-$RANDOM_CNI_NETWORK"
     fi
 
+    if [[ $CONTAINER_RUNTIME == "kata-runtime" ]]; then
+        export RUNTIME_NAME="$CONTAINER_RUNTIME"
+        export CONTAINER_DEFAULT_RUNTIME="$RUNTIME_NAME"
+    fi
```
kolyshkin, line 304: Maybe move this to global context (to where this variables are initially set).

pinns/Makefile
```diff
@@ -12,6 +12,9 @@ all: ../bin/pinns
 	$(CC) -o $@ $^ $(CFLAGS) $(LIBS)
 	$(STRIP) -s $@
 
+%.o: %.c $(HEADERS)
+	$(CC) $(CFLAGS) -g3 -o $@ -c $<
```
kolyshkin, line 16: why `-g3` here?

pinns/src/pinns.c
```diff
@@ -54,9 +56,10 @@ int main(int argc, char **argv) {
       {"filename", required_argument, NULL, 'f'},
       {"uid-mapping", optional_argument, NULL, UID_MAPPING},
       {"gid-mapping", optional_argument, NULL, GID_MAPPING},
+      {"sysctl", optional_argument, NULL, 's'},
```
kolyshkin, line 59: looks like `s/optional_argument/required_argument/`.

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+  char* key = NULL;
+  char* value = NULL;
+  while (sysctl) {
+    if (separate_sysctl_key_value (sysctl, &key, &value) < 0)
+      return -1;
+
+    if (write_sysctl_to_file (key, value) < 0)
+      return -1;
+	
+	sysctl = strtok (NULL, sysctl_delim);
```
kolyshkin, line 24: looks like wrong indentation (hard to see with github though)

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+
+  // if sysctl ends in a ', we should strip it
+  char* ending_char = strchr (equals_token, '\'');
+  if (ending_char)
+	  *ending_char = '\0';
+
+  // sysctl_value is now value
+  *sysctl_value = equals_token;
+
+  if (!strlen (*sysctl_key))
```
kolyshkin, line 61: this check can be done earlier, right after
```c
 char * equals_token = strchr (sysctl_key_value, '=');
￼  if (!equals_token) {
￼    nwarnf ("sysctl must be in the form of 'key=value'; '=' missing from %s", sysctl_key_value);
￼    return -1;
￼  }
```
you can add
```c
if equals_token == sysctl_key_value {
    nwarnf ("sysctl must be in the form of 'key=value'; key is empty");
￼    return -1;
}

File: pinns/src/sysctl.c [hunk 45]
```diff
@@ -0,0 +1,105 @@
...
+      return -1;
+	
+	sysctl = strtok (NULL, sysctl_delim);
+  }
+
+  return 0;
+}
+
+// sysctl_key_value should be in the form `'key=value'`
+static int separate_sysctl_key_value (char* sysctl_key_value, char** sysctl_key, char** sysctl_value)
```
kolyshkin, line 31: I think if you drop `sysctl_` prefix from all the variables here (not the function name though), it will result in a better looking code.

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+  *equals_token = '\0';
+
+  // sysctl_key is now key
+  *sysctl_key = sysctl_key_value;
+
+  // equals_token is now value'
+  ++equals_token;
+
+  // if sysctl ends in a ', we should strip it
+  char* ending_char = strchr (equals_token, '\'');
```
kolyshkin, line 54: 1. This should be stripped only if we found the opening `'` earlier.
2. What you say ("if sysctl ends in a ', we should strip it") is not what you do here (find `'` and strip it with the remainder.
3. I think shell will strip those single quotes for you.

pinns/src/sysctl.c
kolyshkin: same here: `key` and `value` would be simpler (easier to read).

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+    pwarnf ("sysctl key or value not initialized");
+    return -1;
+  }
+
+  // replace periods with / to create the sysctl path
+  for (char* it = sysctl_key; *it; it++)
+    if (*it == '.')
+      *it = '/';
+
+  _cleanup_close_ int dirfd = open ("/proc/sys", O_DIRECTORY | O_RDONLY);
```
kolyshkin, line 87: I think you need `O_PATH` instead of `O_RDONLY` here. Also `O_CLOEXEC`.

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+  if (UNLIKELY (dirfd < 0)) {
+    pwarnf ("failed to open /proc/sys");
```
kolyshkin, line 89:
```c
pwarnf ("failed to open /proc/sys: %m");
```

pinns/src/sysctl.c
```diff
@@ -0,0 +1,105 @@
...
+
+  _cleanup_close_ int fd = openat (dirfd, sysctl_key, O_WRONLY);
+  if (UNLIKELY (fd < 0)) {
+    pwarnf ("failed to open /proc/sys/%s", sysctl_key);
+    return -1;
+  }
+
+  int ret = TEMP_FAILURE_RETRY (write (fd, sysctl_value, strlen (sysctl_value)));
+  if (UNLIKELY (ret < 0)) {
+    pwarnf ("failed to write to /proc/sys/%s", sysctl_key);
```
kolyshkin, line 101: same here, add `%m`

pinns/src/pinns.c
```diff
@@ -174,6 +180,11 @@ int main(int argc, char **argv) {
       pexit("Cannot write gid mappings");
   }
 
+  if (sysctls && configure_sysctls(sysctls) < 0) {
+    pexit("Failed to configure sysctls after unshare");
+    return EXIT_FAILURE;
```
kolyshkin, line 185: no need to have `return` here, as `pexit` calls `exit`.

pinns/src/sysctl.c
kolyshkin: `pwarn ("failed to open /proc/sys");`

internal/oci/container.go
```diff
@@ -505,3 +525,11 @@ func (c *Container) ShouldBeStopped() error {
 	}
 	return nil
 }
+
+// Spoofed returns whether this container is spoofed
```
kolyshkin, line 529: missing `.` at EOL.

internal/oci/container.go
```diff
@@ -505,3 +525,11 @@ func (c *Container) ShouldBeStopped() error {
...
+// a container should be spoofed when it doesn't have to exist in the container runtime
```
kolyshkin, line 530: `s/a container/A container/`

internal/oci/container.go
```diff
@@ -505,3 +525,11 @@ func (c *Container) ShouldBeStopped() error {
...
+// but does need to exist in the storage. The main use of this is when an infra container
+// is not needed, but sandbox metadata should be stored with a spoofed infra container
```
kolyshkin, line 532: missing `.` at EOL

internal/lib/sandbox/sandbox.go
```diff
@@ -384,12 +384,17 @@ func (s *Sandbox) Created() bool {
 // Ready returns whether the sandbox should be marked as ready to the kubelet
 // if there is no infra container, it is always considered ready
 // takeLock should be set if we need to take the lock to get the infra container's state
+// if there is no infra container, it is always considered not ready
+// if the infra container is spoofed, the pod is considered ready when it's been created but not stopped
```
kolyshkin, line 388: Please fix punctuation and capitalization, it's a bit hard to follow otherwise.

test/cgroups.bats
haircommander: the reason is to future proof the test. These tests require we don't drop infra, regardless of what the default is. Setting them here decreases the difficulty of changing the default

also, manage_ns_lifecycle is true by default

docs/crio.conf.5.md
haircommander: fixed all!

test/cgroups.bats
haircommander: would you rather I don't set it @kolyshkin

pinns/Makefile
haircommander: honestly, I forget. changing to O3

pinns/src/pinns.c
haircommander: no it should be optional. below, I check whether sysctls was specified before processing them

pinns/src/sysctl.c
haircommander: very slick!

haircommander:
(most) comments address, thanks for the thorough review @kolyshkin 

pinns/src/sysctl.c
haircommander: I am going to leave the `'` handling (3), but everything else was addressed

docs/crio.8.md
```diff
@@ -170,6 +171,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--default-ulimits**="": Ulimits to apply to containers by default (name=soft:hard) (default: []) (default: [])
 
+**--drop-infra-ctr**: Determines whether pods are created without an infra container (when the pod is not using a pod level PID namespace). Requires ManageNSLifecycle to be true (default: false)
```
saschagrunert, line 174: I suggest we implicitly set `ManageNSLifecycle` to true if this toggle switches on.

saschagrunert (approved):
Just one suggestion around the general behavior, otherwise LGTM

docs/crio.8.md
haircommander: I think it is more clear (as an experimental feature) that we fail quickly when it's not set. By the time DropInfra is set to true by default, I like to think we won't have ManageNSLifecycle as an option anymore

PR #4199, opened 2020-09-15
 Provide support for checkpoint and restore

adrianreber:
#### What type of PR is this?

/kind feature

Provide support for checkpoint and restore

This implements minimal container checkpoint and restore support in cri-o.

The minimal implementation offers the possibility to checkpoint and restore a container. Currently checkpoints do not survive a reboot, which would be a nice additional feature. As 'bundlePath' is on a tmpfs (/run) the checkpoint is gone after reboot. It would be possible to store the checkpoint somewhere else to survive a reboot, but to restore a container config.json is also needed.

Similar to Podman the plan is to support exporting checkpoints which will include all necessary information to restore a container after a reboot or on another system but this is not part of this initial implementation. Using 'crictl' this looks like:
```
# crictl checkpoint CTR-ID
# crictl restore CTR-ID
```
In addition to the simple checkpoint and restore support this implementation also offers the possibility to restore a container into another sandbox.
```
# crictl checkpoint CTR-ID
# crictl restore -p POD-ID CTR-ID # POD-ID being the destination pod
```
This requires the latest runc (https://github.com/opencontainers/runc/pull/2583) version to be able to let runc tell CRIU to join the namespaces of the infrastructure container.

This is based on the Podman checkpoint/restore support.

As mentioned, this is only the initial, minimal change to get checkpoint/restore working. The corresponding API changes are not yet submitted to the corresponding upstream repository as I am not yet sure how that works. This is definitely, for now, a draft to get feedback about this implementation/approach.

```release-note
This adds minimal checkpoint/restore support to cri-o. The minimal support allows to checkpoint a container and restore it again. Optionally it can be restored in another pod.
```


haircommander:
/ok-to-test

thanks for taking this on @adrianreber ! I am curious what the motivation is to add this to cri-o. is there desire for checkpoint/restore to be added to k8s?

haircommander:
ah I just have looked at the issues, nevermind :laughing:

haircommander:
you likely will have to submit a [KEP](https://github.com/kubernetes/enhancements/tree/master/keps) for sig-node to discuss liability, use-cases, etc to get full support from the community and everything merged

adrianreber:
@haircommander Thanks for the KEP pointer. I submitted one here: https://github.com/kubernetes/enhancements/pull/1990

vendor/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto
```diff
@@ -1308,3 +1313,23 @@ message ReopenContainerLogRequest {
 
 message ReopenContainerLogResponse{
 }
+
+message CheckpointContainerRequest {
+    // ID of the container to be checkpointed
+    string container_id = 1;
```
schrej, line 1319: I chose to use a struct for the options of both the checkpoint and the restore request in [my extension](https://github.com/schrej/kubernetes/blob/feature/pod-migration/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto#L868-L902) so it's easier to extend in the future, for example if "actual" live-migration with pagefaultd, which would add a lot of parameters, should be supported in the future.
A struct makes it easier to add additional fields and also keeps the [interface used by kubernetes](https://github.com/schrej/kubernetes/blob/a9d7bcbe16427d26f81c0e9a7a193388ca9f6902/staging/src/k8s.io/cri-api/pkg/apis/services.go#L40-L43) clean. It's done the same way for other actions with lots of parameters.

adrianreber: @schrej Thanks for your input. I opened a KEP at https://github.com/kubernetes/enhancements/pull/1990 about checkpoint/restore in the API. That is probably the place at which to have the API discussion. If structures are better suited then I am happy to use structures.

yimin-zhao:
Hi @adrianreber, I met an error when I try to reproduce your work in my local environment, here is the symptom:
I followed the [tutorial](https://github.com/cri-o/cri-o/blob/master/tutorials/crictl.md) of crio and created a redis server in a pod sandbox, then I ran `checkpoint`, it returned successfully, but when I do `restore`, the cmd line is blocked. So I checked the process in the background, redis was bringing back alive, but I find a few lines of errors in restore.log:
```
0.123143) Running post-restore scripts
(00.123149)     RPC
(00.126661) mnt: Switching to new ns to clean ghosts
(00.142560) Unlock network
(00.142568) Running network-unlock scripts
(00.142572)     RPC
iptables-restore: line 5 failed
(00.146070) Error (criu/util.c:631): exited, status=1
ip6tables-restore: line 5 failed
(00.148616) Error (criu/util.c:631): exited, status=1
(00.148890) pie: 11: seccomp: mode 0 on tid 11
(00.148984) pie: 10: seccomp: mode 0 on tid 10
(00.149063) pie: 9: seccomp: mode 0 on tid 9
(00.149137) pie: 1: seccomp: mode 0 on tid 1
....
(00.150002) 22099 (native) is going to execute the syscall 15, required is 15
(00.150021) 22099 was stopped
(00.150054) 22099 was trapped
(00.150060) 22099 (native) is going to execute the syscall 11, required is 11
(00.150115) 22099 was stopped
(00.150126) Running pre-resume scripts
(00.150130)     RPC
(00.169641) Restore finished successfully. Tasks resumed.
(00.169652) Writing stats
(00.169737) Running post-resume scripts
(00.169741)     RPC
``` 
Does this error matters? For me, it looks like there is something wrong when returning the final status.
I rebased your code onto cri-o's master branch, and also build the crictl with your [crictl's pr](https://github.com/kubernetes-sigs/cri-tools/pull/662). 

Thanks

adrianreber:
To be able to present a working implementation of `kubectl drain 127.0.0.1 --checkpoint` for https://github.com/kubernetes/enhancements/pull/1990 I extended this PR from a proof of concept to a complete implementation of container and pod checkpoint/restore. With this PR I can checkpoint pods, reboot and restore those pods:
```
# crictl ps -a
CONTAINER           IMAGE                               CREATED             STATE               NAME                ATTEMPT             POD ID
1969cedbab7fc       quay.io/adrianreber/counter         2 seconds ago       Running             counter             0                   06e959aa6430e
b0793d84aeb18       quay.io/adrianreber/wildfly-hello   2 seconds ago       Running             wildfly             0                   06e959aa6430e
# curl `crictl inspect 1969cedbab7fc | jq -r '.info.runtimeSpec.annotations["io.kubernetes.cri-o.IP.0"]'`:8088
counter: 0
# curl `crictl inspect 1969cedbab7fc | jq -r '.info.runtimeSpec.annotations["io.kubernetes.cri-o.IP.0"]'`:8088
counter: 1
# curl `crictl inspect b0793d84aeb18 | jq -r '.info.runtimeSpec.annotations["io.kubernetes.cri-o.IP.0"]'`:8080/helloworld/
0
# curl `crictl inspect b0793d84aeb18 | jq -r '.info.runtimeSpec.annotations["io.kubernetes.cri-o.IP.0"]'`:8080/helloworld/
1
# crictl checkpoint --export=/tmp/cp.tar 06e959aa6430e
06e959aa6430e
# reboot
...
# crictl ps -a
CONTAINER           IMAGE               CREATED             STATE               NAME                ATTEMPT             POD ID
# crictl restore --import=/tmp/cp.tar
6f304b70f09d5d9cf85dbbdc448aa794369e4b955e6cd284de08cea04c8dc77f
# crictl ps -a
CONTAINER           IMAGE                                      CREATED             STATE               NAME                ATTEMPT             POD ID
6f304b70f09d5       quay.io/adrianreber/wildfly-hello:latest   2 seconds ago       Running             wildfly             0                   160cdcd98301f
cd5a64efdb752       quay.io/adrianreber/counter:latest         3 seconds ago       Running             counter             0                   160cdcd98301f
# curl `crictl inspect 6f304b70f09d5 | jq -r '.info.runtimeSpec.annotations["io.kubernetes.cri-o.IP.0"]'`:8080/helloworld/
2
# curl crictl inspect cd5a64efdb752 | jq -r '.info.runtimeSpec.annotations["io.kubernetes.cri-o.IP.0"]'`:8088
counter: 2
```

Upon restore a new Pod is created and into this new Pod the checkpointed containers are restored into and stateful applications (I am using a Python server (counter) and a Java server (wildfly)) are continuing to run from the previous state.

This is all based on my Podman checkpoint/restore implementation.

adrianreber:
@yimin-zhao Nice to see interest in this PR. I am always checkpointing and restoring a Wildfly and a Python based server.

I never tried redis before, but using the steps from the tutorial with this latest PR it works for me to checkpoint and restore the redis server. Can you retry with my latest changes?

adrianreber:
At this point I have the actual code, integration tests and unit tests.

CI so far seems to like the changes except for the 'vendor' test because the CRI API changes are not yet upstream. Once the CRI API changes are available upstream I will remove the WIP flag.

haircommander:
/ok-to-test

amazing, thanks @adrianreber

yimin-zhao:
@adrianreber Cool, the latest code is working.

adrianreber:
Rebased and added support to correctly restore Pods with DNS configuration set.

yimin-zhao:
@adrianreber Hi, I'd like to provide some information about a potential bug:
In short: run a redis-server, do `checkpoint` and `restore`, then redis-server is alive, but redis-cli cannot connect to it. 
```
CRIU version:
Version: 3.15
GitID: v3.14-174-gf68da4a86
```
```
runc version 1.0.0-rc92+dev
commit: 44f221e2fce70dd6a75e7d67e9bdde36706aef79-dirty
spec: 1.0.2-dev
go: go1.15.5
libseccomp: 2.3.1
```
I also observed that when I try to send an udp broadcast in the net namespace of the redis-server, I got `EPERM(Operation not permitted)`, for some reasons, I've already set both pod and container to privileged.

Thanks

adrianreber:
@yimin-zhao Do you restore the container into an existing Pod or is the Pod newly created.

Currently CRIU still sets up iptables rules in the network namespace of the Pod. This needs to be fixed in CRIU.

If you do a `ip netns list` and then `ip netns exec NETNS-ID iptables -F` you should be able to use the Pod just as before.

yimin-zhao:
Restore to an existing Pod, I guess, I didn't pass pod id when do `restore`.
Where to execute your command? On the host side?

yimin-zhao:
Ok, I got it, it works, thank you very much.

adrianreber:
Rebased and force pushed with the latest changes as shown in the demo at https://people.redhat.com/~areber/2021-01-29-kubectl-drain-checkpoint.mp4

adrianreber:
Updated the PR to use the experimental CRI API as proposed in the Kubernetes SIG Node meeting.

adrianreber:
Updated to use more Podman code as mentioned in https://github.com/containers/podman/pull/9521

The nice thing is that I can now export a checkpoint in Podman and import that exported checkpoint into an already running pod in CRI-O. It is now possible to migrate containers from Podman to CRI-O.

internal/oci/runtime_oci.go
```diff
@@ -1220,3 +1231,248 @@ func (r *Runtime) SpoofOOM(c *Container) {
...
+		c.state.ExitCode = utils.Int32Ptr(0)
+		c.state.Finished = time.Now()
+	}
+
+	return nil
+}
+
+// RestoreContainer restores a container.
+func (r *runtimeOCI) RestoreContainer(c *Container, sbSpec *rspec.Spec, infraPid int, cgroupParent string) error {
+	c.opLock.Lock()
```
yiminmv, line 1335: Here is a potential deadlock. In function `RestoreContainer()`, it will do `c.opLock.Lock()` then call `CreateContainer()` at line 1449 to restore container. If restore failed(like criu returns errors), it will call `DeleteContainer()` at line 200 to delete the container and it will try to do `c.opLock.Lock()` again , hence causes a deadlock.

internal/oci/runtime_oci.go
```diff
@@ -1220,3 +1231,248 @@ func (r *Runtime) SpoofOOM(c *Container) {
 func ConmonPath(r *Runtime) string {
 	return r.config.Conmon
 }
+
+// CheckpointContainer checkpoints a container.
+func (r *runtimeOCI) CheckpointContainer(c *Container, specgen *rspec.Spec, leaveRunning bool) error {
+	c.opLock.Lock()
```
yimin-zhao, line 1237: >Here is a potential deadlock. In function `RestoreContainer()`, it will do `c.opLock.Lock()` then call `CreateContainer()` at line 1449 to restore container. If restore failed(like criu returns errors), it will call `DeleteContainer()` at line 200 to delete the container and it will try to do `c.opLock.Lock()` again , hence causes a deadlock.

Sorry for the noisy, switch to a proper account and forward the comment again.

adrianreber:
After a couple of months I have finally updated this again. I had a lot of discussions with (and help from) @mrunalp

One idea we are currently following with this modified version of the PR is to keep the unmerged CRI API changes at a internal location (`server/cri/experimental`) while we keep on working on the corresponding KEP (https://github.com/kubernetes/enhancements/issues/2008). Using the CRI API changes in such a way enables us to implement all the checkpoint/restore necessary changes in CRI-O while the KEP is still being discussed. This way it should be possible, if desired, to merge the checkpoint/restore changes into CRI-O without being blocked by the ongoing KEP discussions.

Having the checkpoint/restore implementation currently independent of the CRI API discussion enables us to test checkpoint/restore earlier in CRI-O.

This PR also introduces a new small tool: `crio-cr`. `crio-cr` only has two subcommands: `checkpoint` and `restore`. `crio-cr` works in the same way as the corresponding `crictl` changes (https://github.com/kubernetes-sigs/cri-tools/pull/662) which are also blocked by the CRI API discussions. The two subcommands work in the same way as already known from Podman.

With the current form of the PR it is for example possible to checkpoint a container out of Kubernetes for analysis on another system. So if a container is behaving in some unusual way the container can be checkpointed and kept running. The checkpoint can then be transferred to another machine for further analysis without disrupting the original container.

This PR depends on a couple of other PRs in runc and CRIU for correctly handling SELinux labels on restore:

 * https://github.com/checkpoint-restore/criu/pull/1500
 * https://github.com/checkpoint-restore/criu/pull/1486
 * https://github.com/opencontainers/runc/pull/3005

And one more runc PR which has not been opened yet which is needed to provide support for changing selinux mount context.  The runc change mount context PR depends on https://github.com/checkpoint-restore/criu/pull/1500

Still working on getting CI happy. The needed CRIU version is not available yet and therefore there are still some errors.

adrianreber:
Without the newest version of CRIU (not yet released) the checkpoint/restore tests cannot run and therefore the code coverage is negative because the new code cannot be tested. Once a new enough version of CRIU is available the code coverage errors will be gone.

server/container_restore.go
```diff
@@ -0,0 +1,479 @@
...
+	if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+		log.Infof(ctx, "CreateCtr: context was either canceled or the deadline was exceeded: %v", ctx.Err())
+		return "", ctx.Err()
+	}
+	return ctr.ID(), nil
+}
+
+func (s *Server) importPodCheckpoint(ctx context.Context, req *types.RestoreContainerRequest, dir string) (podID string, opts []*lib.ContainerCheckpointRestoreOptions, retErr error) {
+	input := req.Options.CommonOptions.Archive
+	archiveFile, err := os.Open(input)
```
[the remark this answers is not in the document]
adrianreber: ignore

server/container_restore.go
```diff
@@ -0,0 +1,457 @@
...
+
+	response.ID = podID
+	return response, nil
+}
+
+// also taken from Podman
+func (s *Server) CRImportCheckpoint(ctx context.Context, input, sbID string) (ctrID string, retErr error) {
+	// First get the container definition from the
+	// tarball to a temporary directory
+	archiveFile, err := os.Open(input)
```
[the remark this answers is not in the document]
adrianreber: ignore

internal/oci/runtime_oci.go
adrianreber: Thanks. I fixed that.

adrianreber:
This needs runc 1.1.0: https://github.com/opencontainers/runc/milestone/12

adrianreber:
A couple of months ago @mrunalp suggested to try to merge the checkpoint/restore code for CRI-O using an internal/private CRI interface in combination with a CRI-O only command-line tool: `crioctl`.

This checkpoint/restore implementation does not use the suggested CRI changes from https://github.com/kubernetes/enhancements/pull/1990, but creates a CRI-O private CRI interface to be able to early test the checkpoint/restore interface.

This private CRI is identical to the one proposed in the KEP, but using a private CRI implementation in combination with `crioctl` make it possible to early test the checkpoint/restore code. We are confident that the corresponding KEP will be ready during 1.24 and using the private CRI allows us to test it early.

@mrunalp PTAL

adrianreber:
Has any one an idea if the crun failures are caused by my changes or if they happen regularly. For some of the failed crun tests I am not able to see the logs. It says that the tests are still running.

adrianreber:
Rebased but the crun tests are still failing. I see the crun tests also not working in other PRs. I assume this probably means that it is unrelated to this PR.

cmd/crioctl/main.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/urfave/cli/v2"
+	"golang.org/x/net/context"
+
+	pb "github.com/cri-o/cri-o/pkg/private"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+var checkpointCommand = cli.Command{
```
saschagrunert, line 23: can we merge this into crio-status or the other way around?

internal/criocli/criocli.go
```diff
@@ -1012,6 +1015,11 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			Usage:   "The number of seconds between collecting pod and container stats. If set to 0, the stats are collected on-demand instead.",
 			EnvVars: []string{"CONTAINER_STATS_COLLECTION_PERIOD"},
 		},
+		&cli.BoolFlag{
+			Name:    "disable-criu-support",
+			Usage:   "Disable CRIU integration even if the criu binary is found in $PATH.",
+			EnvVars: []string{"CONTAINER_DISABLE_CRIU_SUPPORT"},
+		},
```
saschagrunert, line 1022: How about inverting the logic by having `enable-criu-support` default to `true`?

internal/lib/checkpoint.go
```diff
@@ -0,0 +1,233 @@
...
+func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *ContainerCheckpointRestoreOptions) (string, error) {
+	ctr, err := c.LookupContainer(opts.Container)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to find container %s", opts.Container)
+	}
+
+	configFile := filepath.Join(ctr.BundlePath(), "config.json")
+	specgen, err := generate.NewFromFile(configFile)
+	if err != nil {
+		return "", errors.Wrapf(err, "Not able to read config for container %q", ctr.ID())
```
saschagrunert, line 39: nit
```suggestion
		return "", errors.Wrapf(err, "not able to read config for container %q", ctr.ID())
```

internal/lib/checkpoint.go
```diff
@@ -0,0 +1,233 @@
...
+				logrus.Debugf("Unable to remove file %s", file)
+			}
+		}
+	}
+
+	return ctr.ID(), nil
+}
+
+// Copied from libpod/diff.go
+var containerMounts = map[string]bool{
```
saschagrunert, line 87: Let's consider making those public in c/common and then import it here (can be done as follow-up for sure).

internal/lib/checkpoint.go
```diff
@@ -0,0 +1,233 @@
...
+	"/proc":              true,
+	"/run":               true,
+	"/run/.containerenv": true,
+	"/run/secrets":       true,
+	"/sys":               true,
+}
+
+// getDiff returns the file system differences
+// Copied from libpod/diff.go and simplified for the checkpoint use case
+func (c *ContainerServer) getDiff(id string) ([]archive.Change, error) {
```
saschagrunert, line 101: nit:
```suggestion
func (c *ContainerServer) getDiff(id string) (rchanges []archive.Change, err error) {
```
Then we can remove line 106

internal/lib/checkpoint.go
```diff
@@ -0,0 +1,233 @@
...
+	return nil
+}
+
+func (c *ContainerServer) exportCheckpoint(ctr *oci.Container, specgen *rspec.Spec, export string) error {
+	id := ctr.ID()
+	dest := ctr.Dir()
+	logrus.Debugf("Exporting checkpoint image of container %q to %q", id, dest)
+
+	includeFiles := []string{
+		"dump.log",
```
saschagrunert, line 159: I think this should be a `const` somewhere.

internal/lib/checkpoint.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// Put log file into checkpoint archive
+	_, err = os.Stat(specgen.Annotations[annotations.LogPath])
+	if err == nil {
+		src, err := os.Open(specgen.Annotations[annotations.LogPath])
+		if err != nil {
+			return errors.Wrapf(err, "error opening log file %q", specgen.Annotations[annotations.LogPath])
+		}
+		defer src.Close()
+		destLog, err := os.Create(filepath.Join(dest, annotations.LogPath))
```
saschagrunert, line 187: nit: create a new var for `filepath.Join(dest, annotations.LogPath)` and reuse it in the paths below

cmd/crioctl/main.go
adrianreber: I am open to whatever works best. The original idea behind crioctl came from @mrunalp to be able to test checkpoint/restore in CRI-O even before the CRI API changes are merged in Kubernetes. Now that the KEP has been merged it seems chances are high that the CRI API changes are merged and then we would not need the CRI-O only crioctl. We can just use crictl. I think the deadline for Kubernetes code changes for 1.24 is end of March so if checkpoint/restore is merged in Kubernetes we do not really need the whole crioctl setup. It was a workaround to be able to run tests.

saschagrunert (commented):
Pretty decent work @adrianreber! I just have a bunch of nits :point_down:

internal/criocli/criocli.go
adrianreber: I am not sure I get it.

The idea was to automatically enable CRIU support if the binary is found and have the possibility to disable it.

Maybe you mean something like this: `--enable-criu-support false` to disable it? But then it wouldn't be a boolean parameter. I think. Not sure. If you could provide some more details what you mean that would help me.

internal/lib/checkpoint.go
adrianreber: Changed.

internal/lib/checkpoint.go
adrianreber: Let's do it in a follow-up

internal/lib/checkpoint.go
adrianreber: changed

internal/lib/checkpoint.go
adrianreber: done

adrianreber:
@saschagrunert Thanks for your review. I updated the PR with your suggestion. I think I did not understand your comment regarding `--disable-criu-support`. Maybe you could help me understand it better.

internal/criocli/criocli.go
adrianreber: Ah, okay. Thanks. But (:wink:), the help text now sounds a bit strange. It is enabled by default and to disable it I need to specify `--enable-criu-support=false`, right? This seems odd. I am happy to do the change, it just feels strange.

saschagrunert: > It is enabled by default and to disable it I need to specify `--enable-criu-support=false`, right? 

Yeah.

> This seems odd. I am happy to do the change, it just feels strange.

Personally, it's not odd to me. But since you mentioned that it feels strange we probably have to consider the `disable-` flag. Let's pull in @haircommander, WDYT?

adrianreber: Sounds reasonable.  I will update the PR accordingly. Thanks for the clarification.

adrianreber: Changed it to --enable-criu-support with the default set to disabled.

cmd/crioctl/main.go
adrianreber: I integrated my changes into `crio-status` and renamed `crio-status` to `crioctl`. Please have a look. I hope CI still likes it.

haircommander:
great work as always @adrianreber ! I missed the discussion about adding crioctl and checkpoint/restore support before it's supported in kubernetes. I think I'd like to rehash that conversation because I have some doubts. I feel bad about asking because you've done so much work already, but I am concerned about implementing outside of kubernetes support cc @mrunalp

internal/lib/checkpoint.go
```diff
@@ -0,0 +1,234 @@
...
+		return errors.Wrapf(err, "error reading checkpoint directory %q", id)
+	}
+
+	outFile, err := os.Create(export)
+	if err != nil {
+		return errors.Wrapf(err, "error creating checkpoint export file %q", export)
+	}
+	defer outFile.Close()
+
+	// The resulting tar archive should not readable by everyone as it contains
```
rst0git, line 218: *should not be readable or writable by anyone except its owner...

adrianreber: changed

adrianreber:
Seems `make bundle` fails because I am using a git hash as version for `cri-tools` and it seems `make bundle` requires a released version. As long as there is no `cri-tools` release with `checkpoint` support this PR is stuck.

internal/lib/restore.go
```diff
@@ -0,0 +1,153 @@
...
+
+	metadata "github.com/checkpoint-restore/checkpointctl/lib"
+	"github.com/checkpoint-restore/go-criu/v5/stats"
+	"github.com/containers/podman/v4/pkg/checkpoint/crutils"
+	"github.com/cri-o/cri-o/internal/oci"
+	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/sirupsen/logrus"
+)
+
+func (c *ContainerServer) restoreFileSystemChanges(ctr *oci.Container, mountPoint string) error {
```
haircommander, line 17: nit: can this function be defined below the function where it's used

adrianreber: Sure. That is just what I am used to do in C :wink:

server/container_restore.go
```diff
@@ -0,0 +1,279 @@
...
+		return "", fmt.Errorf("setting container name and ID: %w", err)
+	}
+
+	if _, err = s.ReserveContainerName(ctr.ID(), ctr.Name()); err != nil {
+		return "", fmt.Errorf("kubelet may be retrying requests that are timing out in CRI-O due to system load: %w", err)
+	}
+
+	defer func() {
+		if retErr != nil {
+			log.Infof(ctx, "CreateCtr: releasing container name %s", ctr.Name())
```
haircommander, line 229: nit: shouldn't these errors be RestoreCtr

scripts/github-actions-packages
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+CRIU_REPO="https://download.opensuse.org/repositories/devel:/tools:/criu/xUbuntu_20.04"
```
haircommander, line 4: if I was to be pedantic this should be in the next commit. If you happen to do another rebase you could update, otherwise :shrug:

pkg/config/config.go
```diff
@@ -1067,6 +1071,17 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 			return fmt.Errorf("initialize nsmgr: %w", err)
 		}
 
+		if c.EnableCriuSupport {
+			if err := validateCriuInPath(); err != nil {
+				c.EnableCriuSupport = false
+				return errors.New("cannot enable checkpoint/restore support without the criu binary in $PATH")
+			} else {
```
haircommander, line 1078: nit: don't need this else because we return in the previous block

server/container_restore.go
adrianreber: Fixed

pkg/config/config.go
adrianreber: Right. Fixed.

scripts/github-actions-packages
adrianreber: Also correct and changed.

adrianreber:
Applied all suggestions and force pushed.

haircommander:
/approve

LGTM

@mrunalp @saschagrunert for the final lgtm?

saschagrunert (approved):

PR #4202, opened 2020-09-16
test: skip MetricsGrabber tests

haircommander:
#### What type of PR is this?

 /kind flake

Before, the tests were allowed to run even if the control plane pods weren't up.
This worked, because local-cluster-up doesn't actually start control plane "pods", but runs the binaries directly on the host

However, as of https://github.com/kubernetes/kubernetes/pull/94729, the tests started bailing out if the control plane pods didn't come up

technically, the tests should have been failing before too, but since we didn't register the node with name "master" [0], they were skipped.

The tests don't even make sense in our context, as it's looking for metrics of pods that are run as binaries on the host, so
it is fine to skip them.

[0] https://github.com/kubernetes/kubernetes/blob/303810c7a22d6dc4f940d7e183f7b0518c00569a/test/e2e/instrumentation/monitoring/metrics_grabber.go#L85

```release-note
none
```


umohnani8:
LGTM

mrunalp (approved):

saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

PR #4212, opened 2020-09-18
introduce seccomp override empty

haircommander:
#### What type of PR is this?

/kind feature
> /kind flake

right now, the CRI defaults to an unconfined profile when it's set to empty
However, it's possible admins want the behavior to default to the default profile instead.

add a knob to allow admins to override the empty behavior from unconfined to the default
Note, when set this technically makes CRI-O not CRI compliant, but for increased security, it's worth it.

```release-note
Add option `seccomp_override_empty` to override an unspecified seccomp profile from being unconfined to being the runtime default. Note: setting this option makes CRI-O not fully CRI compliant, but does increase security. 
```


haircommander:
dropped the WIP commits, added an integration test

saschagrunert (commented):
Do we have a chance to change the default in Kubernetes when graduating the CRI?

haircommander:
> Do we have a chance to change the default in Kubernetes when graduating the CRI?

Possibly, when is seccomp going to be GA? possibly then as well. The issue is, unless either of those make 1.20 and also update the default, users won't get it for a while. If upstream k8s changes the default, we can go to that and drop this option. Until then, I think it's something admins should be given the choice for

kolyshkin (commented):
Perpaps something like this instead?

```
--seccomp-empty=[unconfined|default]

Changes the meaning of an empty seccomp profile. By default (and according to CRI spec), empty profile means unconfined. This option allows to treat empty profile as default, which increases security.
```

kolyshkin:
OK, one more alternative is boolean's `--seccomp-treat-empty-as-default`, which makes more sense for bool but it's still somewhat cumbersome.

haircommander:
what about `--seccomp-use-default-when-empty` still verbose but definitely a clear boolean value

haircommander:
> Perpaps something like this instead?
> 
> ```
> --seccomp-empty=[unconfined|default]
> 
> Changes the meaning of an empty seccomp profile. By default
> (and according to CRI spec), empty profile means unconfined.
> This option allows to treat an empty profile as default profile,
> which might increase security.
> ```
> 
> Overall, though, it's too complicated (== semantically overloaded) either way (yours or mine).

I like this verbiage, but it's pretty verbose for help text, I'll add it to the config text though

saschagrunert (approved):
LGTM

mrunalp (approved):

kolyshkin (approved):
The test case needs a revamp but AFAIK @wgahnagl is working on this one, so she can cover this, too.

Issue #3996, opened 2020-07-21
rhel 7 and managed ns are not playing together well


haircommander:
**Description**
[we](https://github.com/cri-o/cri-o/pull/3979#issuecomment-661236332) [have](https://github.com/cri-o/cri-o/pull/3973#issuecomment-660257176) [had](https://github.com/cri-o/cri-o/pull/3987#issuecomment-661858243) [many](https://github.com/cri-o/cri-o/pull/3987#issuecomment-662043024) [of these failures](https://github.com/cri-o/cri-o/pull/3868#issuecomment-661941694)

suspiciously after https://github.com/cri-o/cri-o/pull/3929

It's hard to tell which specific change is causing it, but I am suspicious it's https://github.com/cri-o/cri-o/commit/d8615ec4693bece6b2d537463fde02263e7444cc

unfortunately, I tried reproducing locally, and on an amazon instance, and have gotten no where yet. Instead of adding comments to the issues, I'm going to compile instances here.

**Steps to reproduce the issue:**
1. run the integration tests on rhel

**Describe the results you received:**
sometimes, the managed namespaces are still busy

**Describe the results you expected:**
they should not be busy

**Additional information you deem important (e.g. issue happens only occasionally):**
so far, only on latest cri-o
**Additional environment details (AWS, VirtualBox, physical, etc.):**
so far, only aws

haircommander:
https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3960/test_pull_request_crio_integration_rhel/14278/artifacts/testout.txt

haircommander:
https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3482/test_pull_request_crio_integration_rhel/14388/artifacts/testout.txt

haircommander:
https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/4034/test_pull_request_crio_integration_rhel/15195/artifacts/testout.txt

haircommander:
got one on integration_crun, which is no bueno
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_crio_integration_crun_fedora/7317/s3/

haircommander:
got one on fedora
https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/4080/test_pull_request_crio_integration_rhel/16593/artifacts/testout.txt

[...]

PR #4210, opened 2020-09-18
[do not merge] debug ns file removal

kolyshkin:
Just trying to debug a failure that happened here:

* https://github.com/cri-o/cri-o/pull/4208#issuecomment-694593726
* https://github.com/cri-o/cri-o/pull/4208#issuecomment-694633670

The issue:
 * https://github.com/cri-o/cri-o/issues/3996

Similar bug: 
 * https://github.com/containerd/containerd/issues/3667

### Hypothesis of the hour

The mount gets propagated to other namespaces, and it takes time for the unmount event to be propagated. Might have to do with an older/buggy RHEL7 kernel. Might need to be mitigated by retrying with a wee sleep.

PR #4217, opened 2020-09-21
Fix bogus CI test failures

kolyshkin:
/kind failing-test

The following failure happens in CI on RHEL7 from time to time:

> removing the pod sandbox "0d5bf5eeb0048bb70ab8ee9bca0a497e216a6c6cfa42507a8d735f61c825784d": rpc error: code = Unknown desc = unable to remove managed namespaces: Removing namespaces encountered the following errors [unlinkat /var/run/netns/0376543d-4917-417f-b406-7e8ab07cb847: device or resource busy]

(with different test cases and different namespaces).

RHEL7 kernel has upstream commit [1] backported, but the feature is
controlled by a sysctl and is off by default. For the feature to work,
one needs to set fs.may_detach_mounts = 1.

On production RHEL7 systems, this is done by runc rpm (see [2]), but
we're not using those rpms for CI, so we have to set the sysctl
manually.

Add an integration test case to check the sysctl is set.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=8ed936b5671bfb33d89bc60bdcc7cf0470ba52fe
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1823374#c17

Fixes: https://github.com/cri-o/cri-o/issues/3996
Some more notes at https://github.com/cri-o/cri-o/pull/4210

```release-note
None
```

### DIFF
--- a/contrib/test/integration/main.yml
+++ b/contrib/test/integration/main.yml
@@ -93,6 +93,14 @@
       include: "build/runc.yml"
       when: "build_runc | bool"
 
+    - name: set fs.may_detach_mounts = 1
+      sysctl:
+        name: fs.may_detach_mounts
+        state: absent
+        value: 1
+        sysctl_set: yes
+        ignoreerrors: yes
+
   post_tasks:
     - name: Swap is disused and disabled as required for kubernetes
       include: "disable_swap.yml"
--- a/internal/config/node/node.go
+++ b/internal/config/node/node.go
@@ -55,6 +55,13 @@ func ValidateConfig() error {
 			activated: &systemdHasCollectMode,
 			fatal:     false,
 		},
+		{
+			name:      "fs.may_detach_mounts sysctl",
+			init:      checkFsMayDetachMounts,
+			err:       &checkFsMayDetachMountsErr,
+			activated: nil,
+			fatal:     true,
+		},
 	}
 	for _, i := range toInit {
 		i.init()
--- a/internal/config/node/sysctl.go
+++ b/internal/config/node/sysctl.go
@@ -1,0 +1,40 @@
+package node
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+var checkFsMayDetachMountsErr error
+
+// checkFsMayDetachMounts is called once from ValidateConfig(),
+// and its return value is ignored. It makes cri-o fail in case
+// checkFsMayDetachMountsErr is not nil.
+func checkFsMayDetachMounts() bool {
+	// this sysctl is specific to RHEL7 kernel
+	const file = "/proc/sys/fs/may_detach_mounts"
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			logrus.WithError(err).Debug("checkFsMayDetachMounts")
+		}
+		return true
+	}
+	str := strings.TrimSpace(string(data))
+	val, err := strconv.ParseInt(str, 10, 64)
+	if err != nil { // should never happen
+		logrus.WithError(err).Warnf("checkFsMayDetachMounts: file %s, value %q, can't convert to int", file, str)
+		return true
+	}
+	if val != 1 {
+		checkFsMayDetachMountsErr = fmt.Errorf("fs.may_detach_mounts sysctl: expected 1, got %d; this may result in \"device or resource busy\" errors while stopping or removing containers", val)
+		return false
+	}
+
+	return true
+}
--- a/test/sanity_checks.bats
+++ b/test/sanity_checks.bats
@@ -1,0 +1,20 @@
+#!/usr/bin/env bats
+
+# Sanity checks for test environment. If any of the below tests fail,
+# the test environment is not to be trusted (i.e. bogus failures are
+# quite possible), and the test environment need to be fixed.
+
+
+# RHEL7 kernels should have this sysctl enabled, otherwise
+# we can see EBUSY on mount point removal, container removal
+# etc. It is set in production via runc rpm %postin script,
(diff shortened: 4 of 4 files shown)


test/sanity_checks.bats
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bats
```
haircommander, line 1: I think more users are apt to find this out if it's in cri-o itself. we could have a log message come out of `internal/config/node` similarly to how we check other node level knobs. wdyt

contrib/test/integration/main.yml
```diff
@@ -93,6 +93,14 @@
       include: "build/runc.yml"
       when: "build_runc | bool"
 
+    - name: set fs.may_detach_mounts = 1
+      sysctl:
+        name: vm.overcommit_memory
```
mrunalp, line 98: `fs.may_detach_mounts`

haircommander:
/test ami

(I'm not sure the ami update jobs actually work anymore, but it's worth a shot)

kolyshkin:
Apparently it's working; from integration_rhel [output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/4217/test_pull_request_crio_integration_rhel/21098/artifacts/testout.txt):

> ok 163 if fs.may_detach_mounts is set

contrib/test/integration/main.yml
kolyshkin: Ahhhh. Updated.

kolyshkin:
A bunch of failed CI jobs doesn't have any logs...

/retest

test/sanity_checks.bats
kolyshkin: This one is strictly for CI, so it would fail in case we are not running in a sane env.

Adding a log message to cri-o itself is also on my radar. Added.

kolyshkin: Not sure if we should make it fatal or not. I'd rather make it fatal as otherwise we'll have to deal with some weird EBUSY bugs for no reason.

kolyshkin:
A minor fix to the error message...

mrunalp:
From e2e_fedora logs:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to reload sysctl: net.ipv4.conf.all.route_localnet = 1\nnet.ipv4.ping_group_range = 0 2147483647\nnet.bridge.bridge-nf-call-iptables = 1\nsysctl: cannot stat /proc/sys/fs/may_detach_mounts: No such file or directory\n"}
```

contrib/test/integration/main.yml
```diff
@@ -93,6 +93,14 @@
       include: "build/runc.yml"
       when: "build_runc | bool"
 
+    - name: set fs.may_detach_mounts = 1
+      sysctl:
+        name: fs.may_detach_mounts
+        state: present
+        value: 1
+        sysctl_set: yes
+        ignoreerrors: yes
```
kolyshkin, line 102: ignoreerrors boolean
Choices:
no ←
yes
Use this option to ignore errors about unknown keys.

Alas, this is not working for some reason

kolyshkin: Maybe it's because of `state: present` stance. Let me try removing it.

kolyshkin: retrying with `state: absent`.

internal/config/node/node.go
```diff
@@ -55,6 +55,13 @@ func ValidateConfig() error {
 			activated: &systemdHasCollectMode,
 			fatal:     false,
 		},
+		{
+			name:      "fs.may_detach_mounts sysctl",
+			init:      checkFsMayDetachMounts,
+			err:       &checkFsMayDetachMountsErr,
+			activated: nil,
+			fatal:     true,
```
kolyshkin, line 63: The only question is whether to make it fatal. I'm in favor to do so, and this is gross misconfiguration which will result in various failures later, so it's better to bail out.

kolyshkin: Basically if we fail here it means RPM packaging is broken or something like that.

internal/config/node/sysctl.go
```diff
@@ -0,0 +1,37 @@
...
+
+func checkFsMayDetachMounts() bool {
+	// this sysctl is specific to RHEL7 kernel
+	const file = "/proc/sys/fs/may_detach_mounts"
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			logrus.WithError(err).Debug("checkFsMayDetachMounts")
+		}
+		return false
```
mrunalp, line 23: Do we want the check to fail if the file isn't present?

kolyshkin: We do not. This parameter is only available for RHEL7 kernel, so, if absent, it means we're not running on RHEL7 (or this is very old RHEL7 kernel, or we are inside some kind of container/chroot).

kolyshkin: The return value is not used anywhere (AFAICS). We only fail if we set the `checkFsMayDetachMountsErr` error, and it is only set if the value is read and parsed but is not 1.

kolyshkin: Added a comment and changed a couple of falses to trues.

kolyshkin:
> ci/openshift-jenkins/integration_crun_cgroupv2 — Jenkins job failed.

and if you click on "Details" link it says

> Test started today at 2:36 PM is still running after 39m15s. (more info)

:(

kolyshkin:
Look ma, green CI :tada:

mrunalp (approved):

kolyshkin:
Backports:
* 1.17: #4231
* 1.18: #4232
* 1.19: #4228

PR #4219, opened 2020-09-24
vendor: revendor containers/image and containers/storage to v1.1

fgiudici:
#### What type of PR is this?

/kind dependency-change

```
crictl pull docker.io/fgiudici/test:1.0
crictl rmi docker.io/fgiudici/test:1.0
crictl pull docker.io/node:lts-slim
crictl pull docker.io/fgiudici/test:1.0

FATA[0003] pulling image failed: rpc error: code = Unknown desc = Error: blob sha256:6a73b40493d9a3fd12c93f0d03ec90c78e447dc9f182d119fb7477b75997ba72 is already present, but with size 22522274 instead of 23625492
```

Issue is in the vendored https://github.com/containers/image code.
Purpose of this PR is to evaluate how big would be to move cri-o 1.11 vendored code to an unaffected version.

None (none reported yet)

The issue is in the old vendored containerd/image code we have in cri-o 1.11, which is fixed with a rework on the local cache in a more recent branch.
This PR is to evaluate how big would be to move the containerd/image code to an official, unaffected version.
Exploring also if a punctual fix would be doable here:
https://github.com/containers/image/pull/1051

```release-note
None
```


fgiudici:
Superseded by #4261 . Closing.

PR #4238, opened 2020-10-02
server: be more careful about using not created pods and containers

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

there exist many chances for us to use a pod or container that hasn't finished creating. 

This PR fixes these issues, and unifies some code paths in finding containers

```release-note
none
```

### DIFF
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -24,6 +24,9 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 		if err == sandbox.ErrIDEmpty {
 			return nil, err
 		}
+		if err == errSandboxNotCreated {
+			return nil, fmt.Errorf("StopPodSandbox failed as the sandbox is not created: %s", sb.ID())
+		}
 
 		// If the sandbox isn't found we just return an empty response to adhere
 		// the CRI interface which expects to not error out in not found
--- a/server/server.go
+++ b/server/server.go
@@ -42,6 +42,8 @@ const (
 	rootlessEnvName     = "_CRIO_ROOTLESS"
 )
 
+var errSandboxNotCreated = errors.New("sandbox not created")
+
 // StreamService implements streaming.Runtime.
 type StreamService struct {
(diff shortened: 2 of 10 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests

haircommander:
```
Kubernetes e2e suite: [sig-network] IngressClass [Feature:Ingress] should set default value on new IngressClass [Serial]
```
[hm](https://prow.ci.openshift.org/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/4238/test_pull_request_crio_e2e_features_rhel/18794/)
/retest

server/sandbox_stop_linux.go
```diff
@@ -24,6 +24,9 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 		if err == sandbox.ErrIDEmpty {
 			return nil, err
 		}
+		if err == errSandboxNotCreated {
```
mrunalp, line 27: Do we have evidence that we get a stop call before we return an id? The more likely scenario is a direct kill of processes  at the cgroup level by the kubelet.

haircommander: that's a good question. yeah it's not likely the kubelet can get that container....

(I need this PR independent of this race between kill and start btw, but I can put it with the timeout fix if it doesn't make sense alone)

saschagrunert (approved):
LGTM

mrunalp (approved):

server/server.go
```diff
@@ -42,6 +42,8 @@ const (
 	rootlessEnvName     = "_CRIO_ROOTLESS"
 )
 
+var errSandboxNotCreated = errors.New("sandbox is not created")
```
kolyshkin, line 45: perhaps just
```suggestion
var errSandboxNotCreated = errors.New("sandbox not created")
```

haircommander:
comments addresed!

PR #4245, opened 2020-10-06
test/network.bats: simplify, fix shellcheck

kolyshkin:
_This used to be a part of #4127, which I am splitting into smaller PRs_

#### What type of PR is this?

/kind cleanup

1. Simplify the code.

For example,
```bash
	run crictl start ...
	echo "$output"
	[ "$status" -eq 0 ]
```
is equivalent to
```bash
	crictl start ...
```
as it much easier to read (yes, bats runs tests with set -e
so there is an error check in place).

One more example,
```bash
	run crictl runp ...
	echo "$output"
	[ "$status" -eq 0 ]
	pod_id="$output"
```
is roughly equivalent to
```bash
	pod_id=$(crictl runp ...)
```
(except it does not print stdout, only stderr).

Another example,
```bash
	cfg=$(python .... < $input)
	echo "$cfg" > $output
```
is equivalent to
```bash
	python ... < $input > $output
```
and avoids the use of a temporary variable (saves some memory).

Finally,
```bash
	run crictl start ...
	echo "$output"
	[ "$status" -ne 0 ]
```
is equivalent to
```bash
	crictl start ... && false
```
2. Optimize use of `CONMON_BINARY` variable, and remove unneeded cleanup.

3. Fix indentation, add some vertical whitespace for better readability.

None

None

```release-note
None
```

### DIFF
--- a/test/network.bats
+++ b/test/network.bats
@@ -13,26 +13,20 @@ function teardown() {
 
 @test "ensure correct hostname" {
 	start_crio
-	run crictl runp "$TESTDATA"/sandbox_config.json
-	echo "$output"
-	[ "$status" -eq 0 ]
-	pod_id="$output"
-	run crictl create "$pod_id" "$TESTDATA"/container_redis.json "$TESTDATA"/sandbox_config.json
-	echo "$output"
-	[ "$status" -eq 0  ]
-	ctr_id="$output"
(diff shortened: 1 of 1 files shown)


mrunalp (approved):

umohnani8:
LGTM
/test e2e_features_fedora

PR #4247, opened 2020-10-07
test/network.bats: fix "Clean up network" tests

kolyshkin:
#### What type of PR is this?
/kind cleanup

Commit 58657488e31 replaced the last component of
/var/lib/cni/networks path in network.bats to be $RANDOM_CNI_NETWORK,
while in fact it should be $CNI_DEFAULT_NETWORK
(see https://github.com/cri-o/cri-o/commit/58657488e31518ebebe3ea27ae3f540f8a9fb5b2#diff-807eaca3d52b5557160919f85ff738bc)
    
The bug was never found because bats do not `set -o pipefail`
and so the non-zero exit code from `ls` was ignored.
    
Let's fix the bug. While at it,
 - merge some 90% duplicated code;
 - introduce and use `fail` helper function.

None

```release-note
None
```

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -534,8 +534,6 @@ function prepare_network_conf() {
     }
 }
 EOF
-
-    echo 0
 }
 
 function write_plugin_test_args_network_conf() {
@@ -561,8 +559,6 @@ EOF
     if [[ -n "$1" ]]; then
         echo "DEBUG_ARGS=$1" >"$TESTDIR"/cni_plugin_helper_input.env
     fi
-
-    echo 0
 }
 
 function prepare_plugin_test_args_network_conf() {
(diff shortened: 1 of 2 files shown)


kolyshkin:
integration test failures seem to be caused by inability to pull images

/test pipeline

mrunalp (approved):

kolyshkin:
CI failure is image server glitch
> # time="2020-10-09T04:24:17Z" level=fatal msg="pulling image: rpc error: code = Unknown desc = Error reading blob sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4: Error fetching blob: invalid status code from registry 500 (Internal Server Error)"

/test integration_rhel

PR #4250, opened 2020-10-08
server: don't call ContainerStateToDisk() to a nil container

fidencio:
#### What type of PR is this?

/kind bug

I didn't face any specific issue about this, but there's the risk of
calling ContainerStateToDisk() when podContainerInfra is nil, and it
could be causing issues as down in the stack there's no guard to ensure
the container is not nil.

```release-note
None
```

### DIFF
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -67,7 +67,7 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 		for _, ctr := range containers[i:max] {
 			cStatus := ctr.State()
 			if cStatus.Status != oci.ContainerStateStopped {
-				if ctr.ID() == podInfraContainer.ID() {
+				if podInfraContainer != nil && ctr.ID() == podInfraContainer.ID() {
 					continue
 				}
 				c := ctr
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -112,6 +112,7 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 	if err := s.StorageRuntimeServer().StopContainer(sb.ID()); err != nil && !errors.Is(err, storage.ErrContainerUnknown) {
 		log.Warnf(ctx, "failed to stop sandbox container in pod sandbox %s: %v", sb.ID(), err)
 	}
+
 	if err := s.ContainerStateToDisk(podInfraContainer); err != nil {
 		log.Warnf(ctx, "error writing pod infra container %q state to disk: %v", podInfraContainer.ID(), err)
 	}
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
LGTM

server/sandbox_stop_linux.go
```diff
@@ -112,8 +112,11 @@ func (s *Server) stopPodSandbox(ctx context.Context, req *pb.StopPodSandboxReque
 	if err := s.StorageRuntimeServer().StopContainer(sb.ID()); err != nil && !errors.Is(err, storage.ErrContainerUnknown) {
 		log.Warnf(ctx, "failed to stop sandbox container in pod sandbox %s: %v", sb.ID(), err)
 	}
-	if err := s.ContainerStateToDisk(podInfraContainer); err != nil {
-		log.Warnf(ctx, "error writing pod infra container %q state to disk: %v", podInfraContainer.ID(), err)
+
+	if podInfraContainer != nil {
+		if err := s.ContainerStateToDisk(podInfraContainer); err != nil {
+			log.Warnf(ctx, "error writing pod infra container %q state to disk: %v", podInfraContainer.ID(), err)
+		}
```
TomSweeneyRedHat, line 119: Looks like a naked check at line 70 too:  `if ctr.ID() == podInfraContainer.ID() {`

fidencio: And also quite similar to: https://github.com/cri-o/cri-o/blob/49f667a77e393ab80e0725a9f3bf00fd89e7fcc9/server/sandbox_stop_linux.go#L99

kolyshkin: It looks like we're not expecting `podInfraContainer == nil` in here, so maybe just add a `!= nil` check after line 56 -- and error out if it's `nil`, removing all the subsequent checks?

haircommander: I am thinking we should just handle this in ContainerStateToDisk() instead of here

fidencio:
I've rebased the code atop of git main and moved the check inside `ContainerStateToDisk()`.

fidencio:
And I've added an extra patch addressing @TomSweeneyRedHat's comment.

umohnani8:
LGTM
/retest

haircommander:
/retest

@fidencio we still want this right

fidencio:
> /retest
> 
> @fidencio we still want this right

I do think so.
Thanks for the reminder, btw, it just got out my radar entirely.

haircommander:
same lol

haircommander:
I misread the comment as "I don't think so".. oops

fidencio:
I've just talked to @haircommander and I actually think we can drop this small series.

The main reasong being that tight now is unlikely we'll face infraPod as being nil, and in order to actually drop infraPod  (a case where this patch could add some guard) we'd need a huge refactory anyways, so we can safely address this later on.

I'm closing the PR. :-)

PR #4251, opened 2020-10-09
[WIP] more test fixes

### LINKED ISSUE #3662: fix check for rw mounts test
sometime today, the test `privileged ctr -- check for rw mounts` started failing in integration_{fedora,rhel} because `/sys/fs/cgroup` was not mounted `ro` by runc. Oddly, the test is named "check for rw mounts", but switching the grep line to `rw` makes it fail with crun.

instead of tracking down the source of the issue, I've made the test nearly no-op, by just catting /proc/mounts [here](https://github.com/cri-o/cri-o/pull/3661)

We should determine what is expected by that test, and figure out which runtime is wrong.

kolyshkin:
** This is a work in progress, more changes in the pipeline, just checking if it works **

#### What type of PR is this?

/kind bug

(this is also `cleanup` and `failed test`.

Fixes #3662

```release-note
None
```

### DIFF
--- a/test/apparmor.bats
+++ b/test/apparmor.bats
@@ -8,8 +8,8 @@ function teardown() {
 
 # AppArmor tests have to run in sequence since they modify the system state
 @test "apparmor tests (in sequence)" {
-    if [[ $(is_apparmor_enabled) -eq 0 ]]; then
-        skip "skip test since apparmor is not enabled."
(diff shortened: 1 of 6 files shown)


kolyshkin:
Closing this one (splitting into a few smaller PRs)

PR #4253, opened 2020-10-09
test/ctr.bats: fixes and improvements

kolyshkin:
#### What type of PR is this?

/kind bug

* fix a few bugs found in ctr.bats tests.
* make tests more compact (+394 −1,180 lines) and readable (mostly by removing excessive `run` use).
* fix the code to pass `shellcheck`.
* introduces `edit_json` helper to modify json using jq
  (the plan is to switch code using `sed` or `python` to this one).

The test fixed are:

##### 1. "annotations passed through".
    
Commit 5f49b2c1fa added this test case.
    
Obviously, `run crictl` do not output anything to stdout,
and `run grep` do not fail if grep exits with non-zero exit code.
So the test was not testing anything.

##### 2. "privileged ctr -- check for rw mounts".

It was broken from the very beginning, since it was supposed to run a
privileged container and check that /sys/fs/cgroup is mounted rw.
    
Commit d3a50d86f8803e21b ("privileged: set mounts to rw") added
test/testdata/container_config_privileged.json file, but it is not
used by the test (or any other tests). It also checks for ro mount
while it should check for rw (based on what I understood). Finally,
it used incorrect pod config while creating a container (not sure
if it makes any difference, just observing).
    
The test case was later amended to not check ro/rw bits since it was
failing on some configurations (see commit 78badc81ca4490).
    
This commit fixes the config used for container (and adds `edit_json`
function to which hopefully other tests will be eventually converted),
fixes the check for rw mount, and fixes the pod config used to create
container.
    
While at it, remove the never used container_config_privileged.json.

##### 3. "ctr lifecycle", "ctr execsync should not overwrite initial spec args", "privileged ctr -- check for rw mounts".

Commits a2ec1d40d64b and 247d465e8123e5 added a few checks using jq.
Alas, those checks do not work (and never worked); jq always succeeds.
    
This happened because
    
1. in `run cmd1 ... | cmd2` the part starting with the pipe
   character is not part of `run` statement;
    
2. `run` eats `cmd1 ...` output (into `$output` variable);
    
so `cmd2` is provided with empty input.
    
Now,
    
3. `jq` with empty input does not run any filters and thus succeeds
   (even with `-e`, see [1]).
    
The fix is to add a separate check that the output is not empty.
    
While at it, remove `run` where it's not needed from the other places
in those three tests we fix.
    
[1] https://github.com/stedolan/jq/issues/1628

Please see individual commits for more details.

Fixes: #3662

```release-note
None
```


test/ctr.bats
```diff
@@ -12,15 +12,17 @@ function teardown() {
...
 function wait_until_exit() {
     ctr_id=$1
     # Wait for container to exit
     attempt=0
     while [ $attempt -le 100 ]; do
         attempt=$((attempt + 1))
-        run crictl inspect -o table "$ctr_id"
-        echo "$output"
-        [ "$status" -eq 0 ]
+        output=$(crictl inspect -o table "$ctr_id")
```
TomSweeneyRedHat, line 25: In at least the Buildah and Podman repos, we've been moving these calls into a helper function.  Changing calls directly to the tool to `run_buildah` as it gives us some extra error handling.   Might be something to consider for cri-o. @edsantiago has driven that in those projects, I'd be interested in hearing his opinion on these changes.  Also, I'm not seeing a status check here, should there be?

kolyshkin: bats runs test with `set -e`, so the status check is implicit and one can run simple tests like
```bash
@test "my test" {
   command1 args ...
   command2 args ...
   [ -f output.file ]
}
```

rather than
```bash
@test "my test" {
   run command1 args ...
   echo "$output"
   [ "$status" -eq 0 ]

   run command2 args ...
   echo "$output"
   [ "$status" -eq 0 ]

   if [ ! -f output.file ]; then
      exit 1
   fi
```

These two examples are roughly equivalent, with some minor details, and I strongly prefer the former -- for clarity.

Besides, there are many inherent problems with `run`, some of them I tried to document in https://github.com/bats-core/bats-core/pull/343 and https://github.com/bats-core/bats-core/pull/366.

I guess I'll eventually come with a doc (for test directory) explaining how to use and not to use `run`.

test/ctr.bats
```diff
@@ -12,6 +12,10 @@ function teardown() {
 	cleanup_test
 }
 
+function is_v2() {
```
mrunalp, line 15: nit: Maybe rename this to `is_cgroupsv2`?

kolyshkin: sure

kolyshkin: I'd prefer `is_cgroupv2` though

kolyshkin: fixed; hope I didn't break anything this time :crossed_fingers:

mrunalp (approved):

kolyshkin:
Not sure what is going on. I have amended and pushed this branch, but github is not picking it up here.

kolyshkin:
single failure with e2e_features_fedora (which I've seen before):

```
1/5226 Tests Failed.
expand_less
Kubernetes e2e suite: [sig-network] IngressClass [Feature:Ingress] should not set default value if no default IngressClass [Serial] expand_less	0s
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/network/ingressclass.go:61
Oct 12 01:36:32.373: Unexpected error:
    <*errors.StatusError | 0xc003a5f9a0>: {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "ingressclasses.networking.k8s.io \"ingressclass1\" already exists",
            Reason: "AlreadyExists",
            Details: {
                Name: "ingressclass1",
                Group: "networking.k8s.io",
                Kind: "ingressclasses",
                UID: "",
                Causes: nil,
                RetryAfterSeconds: 0,
            },
            Code: 409,
        },
    }
    ingressclasses.networking.k8s.io "ingressclass1" already exists
occurred
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/network/ingressclass.go:63
```

There is an open issue (https://github.com/kubernetes/kubernetes/issues/92147) and looks like the related fix (https://github.com/kubernetes/kubernetes/pull/93427 is in our codebase but it's not enough).

test/ctr.bats
```diff
@@ -32,125 +34,66 @@ function wait_until_exit() {
...
-	run crictl create "$pod_id" "$TESTDIR"/container_config_error.json "$TESTDATA"/sandbox_config.json
-	echo "$output"
-	[ "$status" -eq 0 ]
-	ctr_id="$output"
-	run crictl start "$ctr_id"
-	echo "$output"
-	[ "$status" -eq 0 ]
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+
+	python -c 'import json,sys;obj=json.load(sys.stdin);obj["command"] = ["false"]; json.dump(obj, sys.stdout)' \
```
haircommander, line 55: why not edit_json here?

test/ctr.bats
```diff
@@ -32,125 +34,66 @@ function wait_until_exit() {
...
-	ulimits=$(cat "$TESTDATA"/container_config.json | python -c 'import json,sys;obj=json.load(sys.stdin); obj["command"] = ["/bin/sh", "-c", "sleep 600"]; json.dump(obj, sys.stdout)')
-	echo "$ulimits" > "$TESTDIR"/container_config_ulimits.json
-	run crictl create "$pod_id" "$TESTDIR"/container_config_ulimits.json "$TESTDATA"/sandbox_config.json
-	echo "$output"
-	[ "$status" -eq 0 ]
-	ctr_id="$output"
-	run crictl start "$ctr_id"
-	echo "$output"
-	[ "$status" -eq 0 ]
+	python -c 'import json,sys;obj=json.load(sys.stdin); obj["command"] = ["/bin/sh", "-c", "sleep 600"]; json.dump(obj, sys.stdout)' \
```
haircommander, line 70: same, I think all invocations of python should be replaced with edit_json

test/ctr.bats
```diff
@@ -726,154 +408,96 @@ function wait_until_exit() {
...
-	pod_id="$output"
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
 
-	ctrconfig=$(cat "$TESTDATA"/container_config.json | python -c 'import json,sys;obj=json.load(sys.stdin);obj["metadata"]["name"] = "ctr1";obj["labels"]["group"] = "test";obj["labels"]["name"] = "ctr1";obj["labels"]["version"] = "v1.0.0"; json.dump(obj, sys.stdout)')
-	echo "$ctrconfig" > "$TESTDATA"/labeled_container_redis.json
-	run crictl create "$pod_id" "$TESTDATA"/labeled_container_redis.json "$TESTDATA"/sandbox_config.json
-	echo "$output"
-	[ "$status" -eq 0 ]
-	ctr1_id="$output"
+	python -c 'import json,sys;obj=json.load(sys.stdin);obj["metadata"]["name"] = "ctr1";obj["labels"]["group"] = "test";obj["labels"]["name"] = "ctr1";obj["labels"]["version"] = "v1.0.0"; json.dump(obj, sys.stdout)' \
```
haircommander, line 413: same comment as above, edit_json?

test/ctr.bats
kolyshkin: It was not there yet at the time I wrote this.

I am going to have a separate PR attacking this python and sed (and trying to reduce the number of testdata files).

kolyshkin:
No logs for "failed" integration_crun, it seems it just went astray.

/retest

test/ctr.bats
kolyshkin: > In at least the Buildah and Podman repos, we've been moving these calls into a helper function. Changing calls directly to the tool to run_buildah as it gives us some extra error handling. Might be something to consider for cri-o.

@TomSweeneyRedHat can you please share a link with an example of such changes?

edsantiago: https://github.com/containers/buildah/blob/50be5719e09243b4d30a248a72eb6fd2b217b45a/tests/helpers.bash#L105-L107

This has repeatedly proven itself valuable over the years. The key aspects are that it does exit-status check automatically, logs the command run and its output (only displayed when there's an error), and there's the `expect_output` helper (further below) for string checking. In regular BATS all you get is "this did not match", but with no indication of what the actual values were; which makes it completely useless for diagnosing.

I've long had a PR planned for submission to BATS itself, altering `run` itself, but have never finished wrapping it up. Maybe if there's ever a quiet day...

TomSweeneyRedHat: What @edsantiago said.  Thanks Ed!

edsantiago: I've submitted my `run` modifications to the project: https://github.com/bats-core/bats-core/pull/367

kolyshkin: Thanks @edsantiago @TomSweeneyRedHat, let's continue this discussion in  bats-core/bats-core#367

PR #4240, opened 2020-10-02
server: create pod network as early as possible

haircommander:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
when we're managing ns lifecycle, we don't need to wait until the infra container has started before setting up the networking resources, as we have all that is needed (the network namespace, and sandbox annotations).

A key to this is this allows the network create to happen before the infra container creation, which prioritizes the work in case of long running network creations.

this also fixes a case where we may fail to clean up the networking resources if we fail in marshalling the returned structure
#### Does this PR introduce a user-facing change?

```release-note
none
```

PR #4244, opened 2020-10-06
Move back network setup to after adding infra container

mrunalp:
We use the infra container directory to save network stop status

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
We introduced a panic on network stop in #4240 and this PR reverts part of that work.

#### Special notes for your reviewer:
We can revisit this as part of a refactor to separate infra/pinned sandboxes.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4258, opened 2020-10-12
network stop: don't segfault if sandbox isn't created yet

haircommander:
#### What type of PR is this?

/kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

If we create the network before we have an infra container, but fail to fully create a sandbox,
    we attempt to clean up the network. Calling networkStop() causes CRI-O to place a file in the
    sandbox's infra container's directory, thus allowing us to restore the fact that the network had been stopped
    
The problem is, we don't have a infra container directory, so the call segfaults.
    
Instead, check if the sandbox has finished creating before attempting to create the file. if it hasn't, there will be
    no sandbox to restore, so we don't really need the temp file.
    
Another option would be to wire it so that the sandbox has access to the infraContainer.Dir() without actually having an infra container.
That requires another item in libsandbox.New(), which I find cumbersome. Further, I think sandbox creation code is itching for a refactor,
which can include that fix if we find it desireable. In the meantime, this work around is sufficient.

This PR un-reverts https://github.com/cri-o/cri-o/pull/4244 (i.e. reverting https://github.com/cri-o/cri-o/pull/4244), but also fixes https://github.com/cri-o/cri-o/pull/4240#issuecomment-703972496

```release-note
none
```

### DIFF
--- a/internal/lib/sandbox/sandbox.go
+++ b/internal/lib/sandbox/sandbox.go
@@ -347,6 +347,14 @@ func (s *Sandbox) SetNetworkStopped(createFile bool) error {
 }
 
 func (s *Sandbox) createFileInInfraDir(filename string) error {
+	// If the sandbox is not yet created,
+	// this function is being called when
+	// cleaning up a failed sandbox creation.
+	// We don't need to create the file, as there will be no
+	// sandbox to restore
+	if !s.created {
+		return nil
+	}
 	infra := s.InfraContainer()
 	f, err := os.Create(filepath.Join(infra.Dir(), filename))
 	if err == nil {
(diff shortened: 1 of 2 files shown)


internal/lib/sandbox/sandbox.go
```diff
@@ -348,6 +348,14 @@ func (s *Sandbox) SetNetworkStopped(createFile bool) error {
 
 func (s *Sandbox) createFileInInfraDir(filename string) error {
 	infra := s.InfraContainer()
+	// If the sandbox is not yet created,
+	// this function is being called when
+	// cleaning up a failed sandbox creation.
+	// We don't need to create the file, as there will be no
+	// sandbox to restore
+	if !s.created {
```
kolyshkin, line 356: nit: this should probably come before `infra := s.InfraContainer()` line.

kolyshkin:
Would be nice to amend the first commit message with explanation about why this was reverted. It's nice to be able to read git log without referring to PRs.

internal/lib/sandbox/sandbox.go
haircommander: fixed

saschagrunert (approved):

kolyshkin (approved):
LGTM

fidencio:
/lgtm
Thanks, @haircommander!

PR #4259, opened 2020-10-13
test/*.bats: use shfmt

kolyshkin:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This unifies formatting of `bats` files using `shfmt` which recently added `bats` support
(see https://github.com/mvdan/sh/issues/600).

Let's reformat all bats files using `shfmt`, and add a check job to CI to make sure files
are formatted correctly.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

Please review commit-by-commit, see commit messages for details.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4260, opened 2020-10-13
test/ctr.bats: use jq

kolyshkin:
#### What type of PR is this?

/kind cleanup

Currently, we have a plethora of json files in test/testdata, and many ways to generate modified ones.
Some tests use own files under `testdata, some employ `python` or `sed` to make slight modifications.

Let's settle on using `jq` to edit json files. This PR converts `test/ctr.json` to using `jq`.

There are some other minor changes there as well (please see individual commits for details).

None

The first commit should go away once #4259 is merged.

```release-note
None
```


test/ctr.bats
```diff
@@ -51,10 +50,9 @@ function wait_until_exit() {
 @test "ctr termination reason Error" {
 	start_crio
 	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
-
-	python -c 'import json,sys;obj=json.load(sys.stdin);obj["command"] = ["false"]; json.dump(obj, sys.stdout)' \
-		< "$TESTDATA"/container_config.json > "$TESTDIR"/container_config_error.json
-	ctr_id=$(crictl create "$pod_id" "$TESTDIR"/container_config_error.json "$TESTDATA"/sandbox_config.json)
+	jq '	  .command = ["false"]' \
```
haircommander, line 53: is the whitespace between ' and .command needed for the jq command to work?

kolyshkin: No, but in some cases there are multiple elements to be set, so I use this to unify formatting and be shfmt-compatible at the same time.

kolyshkin: Depending on github settings, it might look weird in github, but it looks fine in the terminal.

This is what I mean
Screenshot from 2020-10-13 11-48-05

I am using the same formatting everywhere so that more lines can be added later.

mrunalp (approved):

PR #4263, opened 2020-10-14
runtime_vm: Fix non terminating pods

fidencio:
#### What type of PR is this?

/kind bug

It's been reported by Robert Krawitz that some pods were not terminating when using the VM runtime type.  I've tried to reproduce the issue locally and failed.  Robert, then, gave me access to one of his macines and his scripts and turns out the situation doesn't happen all the time and, when it happens, it happens usually with an infra pod hanging and with pods containing init pods, and it seems to happen because the VM is shutdown, which causes a ttrpc.ErrClosed report, but we don't handle it properly.   This PR tries to fix the issue mentioned.

None

@haircommander, I remember in the past we had issues with removing pods on runtime VM side that caused kubelet to get confused.  Would you have any tip on how I could ensure this is not happening in this case?

Also, I'm not exactly comfortable with `runtime_vm: Don't let wait() return ttrpc.ErrClosed` and `runtime_vm: StopContainers() should not fail when the VM is shutdown`.  Maybe we could actually return ttrpc.ErrClosed (not passing this to errdefs.FromGRPC()) and check for it in the second patch.  That sounds cleaner to me, but I'd like to hear a feedback from the reviewers.

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -843,7 +843,7 @@ func (r *runtimeVM) remove(ctx context.Context, ctrID, execID string) error {
 	if _, err := r.task.Delete(ctx, &task.DeleteRequest{
 		ID:     ctrID,
 		ExecID: execID,
-	}); err != nil {
+	}); err != nil && !errors.Is(err, ttrpc.ErrClosed) {
 		return errdefs.FromGRPC(err)
 	}
 
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -629,7 +629,7 @@ func (r *runtimeVM) updateContainerStatus(c *Container) error {
 		ID: c.ID(),
 	})
 	if err != nil {
-		if errors.Is(err, ttrpc.ErrClosed) {
+		if !errors.Is(err, ttrpc.ErrClosed) {
 			return errdefs.FromGRPC(err)
 		}
 		return errdefs.ErrNotFound
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -814,7 +814,10 @@ func (r *runtimeVM) wait(ctx context.Context, ctrID, execID string) (int32, erro
 		ExecID: execID,
 	})
 	if err != nil {
-		return -1, errdefs.FromGRPC(err)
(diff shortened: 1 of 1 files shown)


haircommander:
I believe it was @evanfoster who found the problems with the kubelet interacting with cri-o and kata

/area vm

this LGTM, we could also introduce our own named error to be clearer about the case, but it makes no functional difference

/approve

fidencio:
> @fidencio are the pods going into a pending state? Could this possibly address some of the issues in https://bugzilla.redhat.com/show_bug.cgi?id=1884035 - Pods are illegally transitioning back to pending

@TomSweeneyRedHat, no, as far as I could see the pods are left forever in the "Terminating" state because of the ttrpc.ErrClosed not being properly treated. Mind, tho, that this only affects pods using "vm" (as in kata-containers) as their runtime class.

By the way, although I don't think it's related, quite nice linking skills, I must admit! I'll keep that BZ reference in mind in case we face something similar in the near future.

Thanks, @TomSweeneyRedHat!

internal/oci/runtime_vm.go
```diff
@@ -814,7 +817,10 @@ func (r *runtimeVM) wait(ctx context.Context, ctrID, execID string) (int32, erro
 		ExecID: execID,
 	})
 	if err != nil {
-		return -1, errdefs.FromGRPC(err)
+		if !errors.Is(err, ttrpc.ErrClosed) {
+			return -1, errdefs.FromGRPC(err)
+		}
+		return -1, errdefs.ErrNotFound
```
mrunalp, line 823: Should we define a custom error here instead?

fidencio: I'm really on the fence about this.

Defining a custom error seems the cleanest possible way to proceed, but we need to ensure it'll be handled inside our own code.  The reason for that is containerd also returning `errdefs.ErrNotFound` in case the error is `ttrpc.ErrClosed`, which makes me think that it's some kind of non-written convention.

Would you be okay with a follow-up PR where I'd switch the ErrNotFound by our own defined error, carefully taking into consideration the places where we actually can do that?

mrunalp: Sounds good :+1:

PR #603, opened 2017-06-15
test: add a test for /etc/resolv.conf in rw/ro mode

runcom:
This patch isn't adding a test for /etc/hosts as that requires host
network and we don't want to play with host's /etc/hosts when running
make localintegration on our laptops. That may change in the future
moving to some sort of in-container testing.

smoke test for https://github.com/kubernetes-incubator/cri-o/pull/599

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #4267, opened 2020-10-14
test/testdata: rm some files

kolyshkin:
#### What type of PR is this?

/kind cleanup

This removes some unused and infrequently-used files from `test/testdata`.
For the latter category, we chose to generate the needed files in place
rather than have a pre-created ones in testdata.

This contradicts suggestion in https://github.com/cri-o/cri-o/pull/603#pullrequestreview-44440778
but I believe this is the right thing to do, because

 1. The configs have many parameters, and in many cases it's not clear from the test
     which config setting are we testing. If we set it explicitly, it becomes obviously.

 2. The more configs in testdata, the harder they are to maintain. Files that were different
     in one place only become more and more different over time, it's harder to apply global
     fixes (such as removing `priviledged: true`) etc.

None

```release-note
None
```


mrunalp (approved):

saschagrunert (approved):
/lgtm

PR #1562, opened 2018-05-08
[WIP] support user namespaces

giuseppe:
Disclaimer:

This is a not working-yet WIP PR.

I am opening it mostly for running the CI tests and get some early feedback.

**- What I did**
Added support for user namespaces

**- How I did it**
Added a new configuration to specify what subuidname/subgidname use for setting up the user namespace

**- How to verify it**
Containers have a different uid/gid mapping

**- Description for the changelog**
user namespaces support

PR #599, opened 2017-06-14
Fix calico

runcom:
Fix #595 

Together with https://github.com/kubernetes-incubator/cri-o/pull/598 this should fix calico

ref: https://github.com/projectcalico/calico/issues/852

PR #4268, opened 2020-10-14
container_create: fix /etc/resolv.conf to be ro for a read-only container

kolyshkin:
#### What type of PR is this?

/kind bug

Commit 0dfec710f2a55c1183 (PR #599, merged Jun 14 2017) fixed the
`/etc/resolv.conf` and `/etc/hosts` bind mounts to be read-only
in case the container has `readonly_rootfs = true`.

A test case was added by commit 8b4e3036ea80304 (PR #603, merged
Jun 16 2017) to verify that `/etc/resolv.conf` is indeed read-only
for a read-only container.

Later, the test case was broken by commit 25dfde90443c2d (PR #1159,
merged Nov 20 2017) to expect success from writing to `/etc/resolv.conf`
for read-only container.

Later, `ro` for `/etc/resolv.conf` was broken by commit 4c096f2812
(PR #1562, merged May 18 2018), which went unnoticed until now
because of the broken test case.

This commit reinstates the `ro` flag, and fixes the test case.
This is important since this file is outside of a container,
and not having it ro basically enables a container to write
to a file system outside of the container.

None

This is currently based on top of #4267.

```release-note
Fix making /etc/resolv.conf bind-mount to be readonly for a readonly container.
```


mrunalp (approved):

umohnani8:
LGTM

saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #4281, opened 2020-10-19
add allowed_annotations to runtime handler

haircommander:
#### What type of PR is this?

/kind api-change

This allows admins to configure which runtime classes are allowed to interpret
which annotations.

Right now, the only supported annotation is io.kubernetes.cri-o.userns-mode, which provides further customizability for when usernamespaces are allowed

Also, fix some small things in the config template. more information in commits
also also, drop support for allow_userns_annotation config option, as it's now redundent

```release-note
add allowed_annotations option to runtime handler structure, which allows admins to gate which runtime classes interpret the annotation io.kubernetes.cri-o.userns-mode. In doing so, also drop the experimental allow_userns_annotation option.
```


haircommander:
I am not sure we need the top level AllowUsernsAnnotation option now that we have a runtime handler specific level. But adding the top level option may give the option more visibility? I am unsure. Luckily, this option has first been added in 1.20-dev so we still have time to change the configuration API before dealing with deprecation cycles.

WDYT @mrunalp @saschagrunert @giuseppe @umohnani8 @kolyshkin @fidencio

pkg/config/config.go
```diff
@@ -148,7 +148,10 @@ type RuntimeHandler struct {
 	RuntimeRoot string `toml:"runtime_root"`
 	// PrivilegedWithoutHostDevices can be used to restrict passing host devices
 	// to a container running as privileged.
-	PrivilegedWithoutHostDevices bool `toml:"privileged_without_host_devices"`
+	PrivilegedWithoutHostDevices bool `toml:"privileged_without_host_devices,omitempty"`
+	// RuntimeAllowUsernsAnnotation can be used to enable this runtime handler to interpret
+	// userns configuration in the pod's annotations
+	RuntimeAllowUsernsAnnotation bool `toml:"runtime_allow_userns_annotation,omitempty"`
```
saschagrunert, line 154: This option is very specific, are we planning to add other annotations later on? We may generalize it then into `allowAnnotations = ["userns"]`, etc… 

In any case I'd recommend to strip the `Runtime`/`runtime_` prefix from the option.

pkg/config/template.go
```diff
@@ -242,9 +299,10 @@ log_level = "{{ .LogLevel }}"
...
 log_filter = "{{ .LogFilter }}"
 
-# Allow usage of the annotation io.kubernetes.cri-o.userns-mode
-# for setting up a user namespace.  This feature is experimental, do not use
-# in production.  It might be changed in future without notice.
+# Allow usage of the annotation io.kubernetes.cri-o.userns-mode for setting up a user namespace.
+# A runtime class must be configured with "runtime_allow_userns_annotation" for this option to
+# have any effect.
+# This feature is experimental, do not use in production. It might be changed in future without notice.
 allow_userns_annotation = {{ .AllowUsernsAnnotation }}
```
saschagrunert, line 306: Yeah I agree having nearly the same option will be misleading to the users. Can we remove/deprecate it?

giuseppe: yes, if we support a finer control, let's just drop `allow_userns_annotation`.  Since it is experimental we can avoid the deprecation

pkg/config/config.go
haircommander: that is *way* better

fidencio:
> I am not sure we need the top level AllowUsernsAnnotation option now that we have a runtime handler specific level. But adding the top level option may give the option more visibility? I am unsure.

@haircommander, you have the technical input of all the folks with the technical experience to do so. IMHO, more like a consumer than a developer,  as long as we document the option in a way that a sysadmin can easily configure it, we're good to go.

One thing that worries me here is if this option ends up as hidden as `privileged_without_host_devices`, which I just found out about a few days ago (as it's not present in the crio.conf man page). Talking about this, your PR should also remove the `allow_user_ns_annotation` mention from the crio.conf man page, and add the info about `allowed_annotations` there.

Does this sound reasonable?

haircommander:
> > I am not sure we need the top level AllowUsernsAnnotation option now that we have a runtime handler specific level. But adding the top level option may give the option more visibility? I am unsure.
> 
> @haircommander, you have the technical input of all the folks with the technical experience to do so. IMHO, more like a consumer than a developer, as long as we document the option in a way that a sysadmin can easily configure it, we're good to go.
> 
> One thing that worries me here is if this option ends up as hidden as `privileged_without_host_devices`, which I just found out about a few days ago (as it's not present in the crio.conf man page). Talking about this, your PR should also remove the `allow_user_ns_annotation` mention from the crio.conf man page, and add the info about `allowed_annotations` there.
> 
> Does this sound reasonable?

Yeah there is a documentation fix commit in this PR for privileged_without_host_devices, as well as allowed_annotations, so I hope each of these options will now be more visible

docs/crio.8.md
```diff
@@ -116,8 +115,6 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--additional-devices**="": Devices to add to the containers  (default: [])
 
-**--allow-userns-annotation**: [EXPERIMENTAL] Allow usage of the annotation to setup user namespaces (default: false)
```
fidencio, line 119: `docs/crio.conf.5.md` also has to be updated, as it contains:
```
**allow_userns_annotation**=false
  [EXPERIMENTAL] Allow usage of the experimental feature for setting user namespaces through the io.kubernetes.cri-o.userns-mode annotation.  This is an experimental feature,
  do not enable in production.  It may be changed in the future without any notice.
```

haircommander: good catch, fixed!

fidencio:
As a consumer, LGTM!
Thanks, @haircommander!

internal/oci/oci.go
```diff
@@ -189,6 +190,22 @@ func (r *Runtime) PrivilegedWithoutHostDevices(handler string) (bool, error) {
...
+	}
+	for _, ann := range rh.AllowedAnnotations {
+		if ann == annotations.UsernsModeAnnotation {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
```
TomSweeneyRedHat, line 208: IDK, but do you expect there'll be many annotations at some point in the future?  If so, would it make sense to have this live in a new file?  Perhaps `internal/oci/annoctions.go`?

TomSweeneyRedHat:
One question for consideration.
LGTM once the tests get happy

internal/oci/oci.go
haircommander: there will be more in the future, so I think a new file will make sense, but for now I think I'll leave it here

saschagrunert (approved):
LGTM

pkg/config/template.go
```diff
@@ -301,12 +297,25 @@ pinns_path = "{{ .PinnsPath }}"
 #   omitted, an "oci" runtime is assumed.
 # - runtime_root (optional, string): root directory for storage of containers
 #   state.
+# - privileged_without_host_devices (optional, bool): an option for restricting
+#   host devices from being passed to privileged containers.
+# - allowed_annotations (optional, array of strings): an option for specifying
+#   a list of experimental annotations that this runtime handler is allowed to process.
+#   The only currently recognized value is "io.kubernetes.cri-o.userns-mode" for configuring
+#   a usernamespace for the pod.
```
mrunalp, line 305: nit: `user namespace`

PR #4284, opened 2020-10-20
[release-1.18] runtime_vm: Fix non terminating pods

fidencio:
#### What type of PR is this?

/kind bug

It's been reported by Robert Krawitz that some pods were not terminating when using the VM runtime type.  I've tried to reproduce the issue locally and failed.  Robert, then, gave me access to one of his macines and his scripts and turns out the situation doesn't happen all the time and, when it happens, it happens usually with an infra pod hanging and with pods containing init pods, and it seems to happen because the VM is shutdown, which causes a ttrpc.ErrClosed report, but we don't handle it properly.   This PR tries to fix the issue mentioned.

None

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -626,7 +626,7 @@ func (r *runtimeVM) updateContainerStatus(c *Container) error {
 		ID: c.ID(),
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM
/retest

mrunalp (approved):

PR #4297, opened 2020-10-22
Add interface to rust library

saschagrunert:
#### What type of PR is this?

/kind feature

This adds a for now empty interface to the Rust library. The idea is to
simply download the (static) library of not existing locally and link
against it. The new `ffi` package will be the main entrypoint around
wrapping structures later on.

None
None

```release-note
None
```


saschagrunert:
/retest 

This seems to work :)

Makefile
```diff
@@ -196,6 +196,20 @@ dependencies: ${GO_MOD_OUTDATED}
 	${GO_RUN} ./scripts/dependencies \
 		--output-path ${BUILD_PATH}/dependencies
 
+LIB_COMMIT := 8b01dbb
+
+.PHONY: update-lib
+update-lib:
+	FILE=lib-$(LIB_COMMIT).tar.gz && \
+	curl -sf https://storage.googleapis.com/k8s-conform-cri-o/lib/$$FILE \
```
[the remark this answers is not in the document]
saschagrunert: That's why I commited the tarball into the repository. We do not have to download anything on packaging. I was playing around with git lfs and stuff like that but just having the static library as tarball inside the repo seemed like the best solution to me.

saschagrunert:
@haircommander @mrunalp do we want to rename the `cri` project before integrating it here?

saschagrunert:
I'd say that we drop the 32bit binary, because this will not work with the Rust lib. We could also build a 32bit version of the rust library, but I do not see a strong reason to stick to 32bit binaries.

.circleci/config.yml
```diff
@@ -6,7 +6,6 @@ stdenv: &stdenv
     GOCACHE: &gocache /tmp/go-build
     IMAGE: &image quay.io/crio/crio-build-amd64-go1.15:master-1.5.0
     IMAGELEGACY: &imagelegacy quay.io/crio/crio-build-amd64-go1.14:master-1.5.0
-    IMAGE386: &image386 quay.io/crio/crio-build-386-go1.15:master-1.5.0
```
saschagrunert, line 9: I'd say that we drop the 32bit binary, because this will not work with the Rust lib. We could also build a 32bit version of the rust library, but I do not see a strong reason to stick to 32bit binaries.

Makefile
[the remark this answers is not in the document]
giuseppe: could we have something like "make dist" to prepare the dist tarball?  Having a 28M binary file in the git repo seems a bit expensive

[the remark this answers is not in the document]
saschagrunert: How would that work? Should we build the whole Rust project from this repository?

.circleci/config.yml
haircommander: I do know of some raspbian users and the 64 bit versions are still fairly recent.... I would vote we keep it

saschagrunert: That's ARM only, right? Phew…we probably have to build the static library for all of those architectures then… :thinking:

haircommander: yeah I believe only ARM even thinks about 32 bit these days :)

saschagrunert: Okay, what do we wanna do now? Build the library here in this repository? It would cause the biggest changes to our CI but seems the only reason to not put static libraries into this repo…

saschagrunert: I'll add the toolchain to our CI.

Makefile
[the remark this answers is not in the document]
saschagrunert: I think we have to add the toolchain to our CI, going for this approach now.

saschagrunert:
This seems to work now, PTAL @mrunalp @haircommander @giuseppe

haircommander:
LGTM, @lsm5 @jnovy are you going to be able to help me set up rust support in our packaging solutions/know anyone who is able to

saschagrunert:
> LGTM, @lsm5 @jnovy are you going to be able to help me set up rust support in our packaging solutions/know anyone who is able to

I think in a long term we have to create a dedicated package around the rust library and add it as dependency to CRI-O.

lsm5:
@haircommander no experience with rust or rust packaging so far, my first thought would be to have a subpackage, but don't know offhand how complicated it would make things with the prior golang stuff in there.

I guess I could look at some current rust packages and get back to you..

jnovy:
never done anything with Rust @haircommander

Issue #4298, opened 2020-10-22
Recent changes causing go modules to be downloaded in CI


saschagrunert:
With the merge of https://github.com/cri-o/cri-o/pull/4208 we're now downloading go modules in CircleCI, which should not happen. For example:
https://app.circleci.com/pipelines/github/cri-o/cri-o/6356/workflows/6223bde7-3cc2-4f16-a201-7ff5923d5ec3/jobs/102326/parallel-runs/0/steps/0-105

Either we remove the vendoring completely or we stick to the vendored sources. The issue seems to be introduced by the removal of the `--mod=vendor` flag.

mrunalp:
@kolyshkin ptal

kolyshkin:
This happens as CI is using golang 1.13 (not supported since August). I mistakenly thought everyone has switched to go 1.14+. With 1.13, `-mod=vendor` is still required.

kolyshkin:
Let me try to fix circleci scripts to have golang. The problem is we use Ubuntu Xenial which is pretty old, and the backports ppa we use (https://launchpad.net/~longsleep/+archive/ubuntu/golang-backports) only has go-1.13 for Xenial.

kolyshkin:
Let's see if this works... https://github.com/cri-o/cri-o/pull/4305

saschagrunert:
> Let me try to fix circleci scripts to have golang. The problem is we use Ubuntu Xenial which is pretty old, and the backports ppa we use (https://launchpad.net/~longsleep/+archive/ubuntu/golang-backports) only has go-1.13 for Xenial.

Ah yeah I see… I think CircleCI also provides some 20.04 images. Let's see if this fixes magically the issue. 
Xref #4306

PR #4305, opened 2020-10-23
circle-ci: use go 1.15.3

kolyshkin:
#### What type of PR is this?

/kind dependency-change

We were using golang from longsleep/golang-backports PPA.
Unfortunately, that repo is apparently no longer updated for Ubuntu
Xenial, with the latest golang being 1.13.

Since 1.13 is no longer supported, let's switch to binary distribution
from google.com, and update to 1.15.

Fixes: https://github.com/cri-o/cri-o/issues/4298

None

```release-note
None
```

### DIFF
--- a/scripts/versions
+++ b/scripts/versions
@@ -3,6 +3,7 @@ set -euo pipefail
 
 # Versions to be used
 declare -A VERSIONS=(
+    ["go"]=1.15.3
     ["cni-plugins"]=v0.8.7
     ["conmon"]=v2.0.20
     ["cri-tools"]=v1.19.0
--- a/scripts/circle-setup
+++ b/scripts/circle-setup
@@ -10,6 +10,7 @@ main() {
     prepare_system
 
     install_packages
+    install_golang
     install_bats
     install_conmon
     install_cri_tools
@@ -35,7 +36,7 @@ prepare_system() {
     sudo iptables -t nat -I POSTROUTING -s 127.0.0.1 ! -d 127.0.0.1 -j MASQUERADE
 }
 
-install_packages() {
+install_golang() {
     # remove old golang version
     sudo rm -rf /usr/local/go
     sudo rm -rf /usr/local/bin/go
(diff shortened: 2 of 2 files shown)


scripts/versions
```diff
@@ -3,6 +3,7 @@ set -euo pipefail
 
 # Versions to be used
 declare -A VERSIONS=(
+    ["go"]=1.15
```
saschagrunert, line 6: Let's use 1.15.3 here

Can we also use the latest CircleCI machine image which uses 20.04: ubuntu-2004:202010-01

kolyshkin: Right. It's too bad we have to specify a minor version.

Switched the image to 20.04 as well

saschagrunert (approved):
LGTM

fidencio:
LGTM, considering happy tests!
Thanks, @kolyshkin!

kolyshkin:
OK, using 20.04 is a separate issue I believe and it should be addressed separately. For now, when blindly switching to it, we have this:

```
Package e2fslibs-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  uuid-dev


No apt package "jq", but there is a snap with that name.
Try "snap install jq"

W: --force-yes is deprecated, use one of the options starting with --allow instead.
E: Unable to locate package btrfs-tools
E: Unable to locate package conntrack
E: Package 'e2fslibs-dev' has no installation candidate
E: Unable to locate package jq
E: Unable to locate package libaio-dev
E: Unable to locate package libapparmor-dev
E: Unable to locate package libcap-dev
E: Unable to locate package libdevmapper-dev
E: Unable to locate package libfuse-dev
E: Unable to locate package libgpgme11-dev
E: Unable to locate package libnet1-dev
E: Unable to locate package libnl-3-dev
E: Unable to locate package libprotobuf-c-dev
E: Unable to locate package libsystemd-dev
E: Unable to locate package libudev-dev

Exited with code exit status 100
```

Removing image bump commit for now.

mrunalp (approved):

kolyshkin:
> using 20.04 is a separate issue I believe and it should be addressed separately.

https://github.com/cri-o/cri-o/pull/4309

PR #4314, opened 2020-10-28
Refactor handling of devices

haircommander:
#### What type of PR is this?

/kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR does two things:
- move the handling of config level devices (additional_devices) to a dedicated package: internal/config/device"
- move the container configuration of devices from server/ to pkg/container

Both of these changes are in anticipation of adding devices from pod annotations in addition to these sources

```release-note
none
```

### DIFF
--- a/internal/config/device/device_test.go
+++ b/internal/config/device/device_test.go
@@ -1,0 +1,48 @@
+package device_test
+
+import (
+	"github.com/cri-o/cri-o/internal/config/device"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = t.Describe("DeviceConfig", func() {
+	var d *device.Config
+	BeforeEach(func() {
+		d = device.New()
+	})
+	t.Describe("LoadDevices", func() {
+		It("should fail with poorly formatted device", func() {
+			// Given
+			// When
+			err := d.LoadDevices([]string{"invalid:invalid"})
+			// Then
+			Expect(err).NotTo(BeNil())
+			Expect(d.Devices()).To(BeEmpty())
+		})
+		It("should fail if invalid device", func() {
+			// Given
+			// When
+			err := d.LoadDevices([]string{"/dev/invalid"})
+			// Then
+			Expect(err).NotTo(BeNil())
+			Expect(d.Devices()).To(BeEmpty())
+		})
+		It("should succeed with valid device", func() {
+			// Given
+			// When
+			err := d.LoadDevices([]string{"/dev/null:/dev/null:w"})
+			// Then
+			Expect(err).To(BeNil())
+			Expect(d.Devices()).NotTo(BeEmpty())
+		})
+		It("should succeed with empty", func() {
+			// Given
+			// When
+			err := d.LoadDevices([]string{""})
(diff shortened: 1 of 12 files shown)


pkg/container/device.go
```diff
@@ -0,0 +1,173 @@
...
+
+					return nil
+				})
+			}
+		}
+	}
+	return nil
+}
+
+// ResolveSymbolicLink resolves a possbile symlink path. If the path is a symlink, returns resolved
```
giuseppe, line 159: possbile -> possible

pkg/container/device.go
```diff
@@ -0,0 +1,173 @@
...
+// path; if not, returns the original path.
+func ResolveSymbolicLink(path, scope string) (string, error) {
+	info, err := os.Lstat(path)
+	if err != nil {
+		return "", err
+	}
+	if info.Mode()&os.ModeSymlink != os.ModeSymlink {
+		return path, nil
+	}
```
giuseppe, line 168: why do we need the extra logic here instead of using directly SecureJoin?  Don't we want to solve symlinks that are not the last path component?

saschagrunert (approved):

pkg/container/device.go
haircommander: I am not sure! This was basically copied over from server/. I will see what dropping it does

haircommander:
ptal @umohnani8 @kolyshkin

pkg/container/device.go
haircommander: we do need resolveSymbolicLinks for the remaining case in container_create_linux, as securejoin doesn't differentiate between a symlink and a path not existing. In that case, we need to create if it doesn't exist, so we still need the function. for the device case, I think we can drop it

fidencio:
LGTM, assuming happy tests.
Thanks, @haircommander!

internal/config/device/device_test.go
```diff
@@ -0,0 +1,48 @@
...
+			// When
+			err := d.LoadDevices([]string{"/dev/invalid"})
+			// Then
+			Expect(err).NotTo(BeNil())
+			Expect(d.Devices()).To(BeEmpty())
+		})
+		It("should succeed with valid device", func() {
+			// Given
+			// When
+			err := d.LoadDevices([]string{"/dev/null:/dev/null:w"})
```
TomSweeneyRedHat, line 34: Maybe I'll discover later down in the code, but is a spec like `"/dev/null::w"` valid?  Regardless you probably want to add a test here for it to make sure it's handled.

TomSweeneyRedHat:
Any man page changes needed or were they covered prior?

internal/config/device/device_test.go
haircommander: WOW thanks apparently we segfault :upside_down_face:

haircommander: fixed in https://github.com/containers/podman/pull/8200

haircommander:
> Any man page changes needed or were they covered prior?

there should be no user facing changes with this PR. there will be man page changes in a follow up to this, so stay tuned

PR #4316, opened 2020-10-29
Upload bundle in separate CI step

saschagrunert:
#### What type of PR is this?

/kind failing-test

We have issues with integration test timeouts, which is now increased
from 10 to 30 minutes. Beside that, we upload the binary bundle without
their influence in a separate step. This should increase the
availability of uploaded artifacts and make commands like this working
better:

```
> curl -f https://storage.googleapis.com/k8s-conform-cri-o/artifacts/\
    crio-$(git ls-remote https://github.com/cri-o/cri-o master | cut -c1-9).tar.gz -o crio.tar.gz
```

None
cc @harche 

```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,6 @@ workflows:
             - results
       - results:
           requires:
-            - bundle-test
             - integration
             - integration-cgroupfs
             - integration-critest
@@ -112,6 +111,9 @@ workflows:
       - shellcheck
       - shfmt
       - unit-tests
+      - upload-artifacts:
+          requires:
+            - bundle-test
       - vendor
 
 jobs:
@@ -449,9 +451,6 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
-      - run:
-          name: Upload artifacts
-          command: make upload-artifacts
       - store_test_results:
           path: build/junit
       - store_artifacts:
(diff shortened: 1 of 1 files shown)


harche:
thanks @saschagrunert

.circleci/config.yml
```diff
@@ -368,6 +370,7 @@ jobs:
       - run:
           name: integration test
           command: sudo -E test/test_runner.sh $TEST_ARGS 2>&1 | tee -a /tmp/testout
+          no_output_timeout: 30m
```
kolyshkin, line 373: Thanks for adding this (saw this "no output for 10 minutes" a lot recently), but is it related to the change in this PR ("Upload bundle in separate CI step")?

If not really, I'd prefer to have this one as a separate commit (same PR is totally fine)

saschagrunert: Yep, I added it as second commit. I'm not sure if this increased timeout is necessary because of actual changes to the integration tests or hardware changes in CircleCI. It _feels_ that their machines slowed down in the past weeks, but I might be wrong.

kolyshkin: Thanks!

Yes, I have the same feeling (slower machines, reduced concurrency).

kolyshkin (approved):
LGTM

PR #4329, opened 2020-11-01
ci: bump crun to 0.15/HEAD

kolyshkin:
1. Since this is a master branch, it makes sense to test crun from
   HEAD (rather than a released version) if we're compiling it from
   source. A similar change was done for runc in commit 9e10f54d3ba834d (PR #3493).

2. For binary crun version, bump to latest 1.15.

/kind dependency-change

```release-note
None
```

@giuseppe PTAL


giuseppe (approved):
LGTM

saschagrunert (approved):
/test e2e_rhel

PR #4310, opened 2020-10-27
test: do test user namespace

kolyshkin:
#### What type of PR is this?

/kind bug

(I consider it a bug in CI that prevented userns testing)

##### 1. test: bring back userns testing

Commit 92caff26d ("Added EnvVar to cri-o cli flags", May 29 2019)
changed some testing code to use CONTAINER_{U,G}ID_MAPPING
instead of {U,G}ID_MAPPING environment variables, but a place
in test/test_runner.sh was omitted. This, I guess, resulted
in not testing userns (at least partially -- there are tests that enable
it explicitly rather through a TEST_USERNS env var).

Next, commit b80c31b9cf ("Allow CRI-O to manage IPC and UTS namespaces",
Dec 16 2019) changed the option ManageNetworkNSLifecycle to
ManageNSLifecycle, and changed its default to be true. Since this
commit, it is now needed to disable the option to allow userns testing
(otherwise we get "cannot use UIDMappings with ManageNSLifecycle"
error).

This commit fixes both issues, allegedly bringing back user namespace
testing.

##### 2. misc

Disable various tests that are not possible to run with userns enabled.

Some of them need to be re-enabled later once we fix userns+manage_ns_lifecycle
working together. Those are marked with TODO.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4331, opened 2020-11-02
pkg: allow [UG]IDMappings with ManageNSLifecycle

giuseppe:
it seems to work fine, no need to not allow it.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

we can use ManageNSLifecycle with user namespaces

Users can use ManageNSLifecycle and user namespaces

```release-note
None
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -785,12 +785,6 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		logrus.Infof("The manage-ns-lifecycle option is being deprecated, and will be unconditionally true in the future")
 	}
 
-	if c.UIDMappings != "" && c.ManageNSLifecycle {
-		return errors.New("cannot use UIDMappings with ManageNSLifecycle")
-	}
-	if c.GIDMappings != "" && c.ManageNSLifecycle {
-		return errors.New("cannot use GIDMappings with ManageNSLifecycle")
-	}
 	if c.DropInfraCtr && !c.ManageNSLifecycle {
 		return errors.New("cannot drop infra without ManageNSLifecycle")
 	}
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -224,7 +224,7 @@ the container runtime configuration.
 
 **drop_infra_ctr**=false
   Determines whether we drop the infra container when a pod does not have a private PID namespace, and does not use a kernel separating runtime (like kata).
-  Requies **manage_ns_lifecycle** to be true.
+  Requires **manage_ns_lifecycle** to be true.
 
 **namespaces_dir**="/var/run"
   The directory where the state of the managed namespaces gets tracked. Only used when manage_ns_lifecycle is true
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -357,30 +357,6 @@ var _ = t.Describe("Config", func() {
 			Expect(err).NotTo(BeNil())
 		})
 
-		It("should fail wrong UID mappings", func() {
-			// Given
-			sut.UIDMappings = "value"
-			sut.ManageNSLifecycle = true
-
-			// When
-			err := sut.Validate(false)
-
-			// Then
-			Expect(err).NotTo(BeNil())
-		})
-
-		It("should fail wrong GID mappings", func() {
-			// Given
-			sut.GIDMappings = "value"
-			sut.ManageNSLifecycle = true
-
-			// When
-			err := sut.Validate(false)
-
-			// Then
-			Expect(err).NotTo(BeNil())
-		})
-
 		It("should fail wrong max log size", func() {
 			// Given
 			sut.LogSizeMax = 1
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -161,7 +161,7 @@ func addDevicesPlatform(ctx context.Context, sb *sandbox.Sandbox, containerConfi
 
 // createContainerPlatform performs platform dependent intermediate steps before calling the container's oci.Runtime().CreateContainer()
 func (s *Server) createContainerPlatform(container *oci.Container, cgroupParent string, idMappings *idtools.IDMappings) error {
-	if idMappings != nil {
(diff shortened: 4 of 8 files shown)


pkg/config/config.go
```diff
@@ -785,12 +785,6 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		logrus.Infof("The manage-ns-lifecycle option is being deprecated, and will be unconditionally true in the future")
 	}
 
-	if c.UIDMappings != "" && c.ManageNSLifecycle {
-		return errors.New("cannot use UIDMappings with ManageNSLifecycle")
```
saschagrunert, line 789: I think there are some unit tests which have to be removed/adapted for those code paths.

kolyshkin:
I'm afraid this is not enough -- I tried doing the same thing in https://github.com/cri-o/cri-o/pull/4310 but ended up with what I summarized at https://github.com/cri-o/cri-o/issues/4327

Currently the tests are passing because userns tests are broken (being fixed by #4310, see TODO items in there though).

giuseppe:
> I'm afraid this is not enough -- I tried doing the same thing in #4310 but ended up with what I summarized at #4327
> 
> Currently the tests are passing because userns tests are broken (being fixed by #4310, see TODO items in there though).

I'll put this PR on hold and rebase on top of it.

/hold

kolyshkin:
I see that you have rebased it. Now you need to
1. Remove `CONTAINER_MANAGE_NS_LIFECYCLE=false CONTAINER_DROP_INFRA_CTR=false` from `test/test_runner.sh`.
2. Re-enable those tests that I marked as TODO in https://github.com/cri-o/cri-o/pull/4310/commits/a3d0b391d5c0bf747707518d1ad3d210437ab970
3. Fix the panic in cri-o (this hunk should work: https://github.com/cri-o/cri-o/commit/e6e7cda588a0cdcd9aa748d930afa630f970a590#diff-1e0877112035840d3e2f073c693665f8daffa241a440e54a866ad7382b510ea2
4. Fix `open /proc/0/uid_map: no such file or directory` from one of the restore tests. Which is where I stopped yesterday...

kolyshkin:
The last problem I'm dealing with is
> CONTAINER_UID_MAPPINGS="0:100000:100000" CONTAINER_GID_MAPPINGS="0:200000:100000" bats -f 'crio restore > first managing then not managing' test/restore.bats
> ✗ crio restore first managing then not managing
>   (in test file test/restore.bats, line 241)
>     `ctr_id=$(crictl create "$pod_id" "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json)' failed
>   time="2020-11-01T13:47:33-08:00" level=fatal msg="creating container: rpc error: code = Unknown desc = cannot open > /proc/0/uid_map: open /proc/0/uid_map: no such file or directory"

Which is caused by the fact that `func (s *Server) getSandboxIDMappings(sb *libsandbox.Sandbox) (*idtools.IDMappings, error) {
` is getting the mappings from infra container's /proc/$PID/uid_map and gid_map. If there is no infra container, I'm not sure what to do in this case.

haircommander:
> The last problem I'm dealing with is
> 
> > CONTAINER_UID_MAPPINGS="0:100000:100000" CONTAINER_GID_MAPPINGS="0:200000:100000" bats -f 'crio restore > first managing then not managing' test/restore.bats
> > ✗ crio restore first managing then not managing
> > (in test file test/restore.bats, line 241)
> > `ctr_id=$(crictl create "$pod_id" "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json)' failed
> > time="2020-11-01T13:47:33-08:00" level=fatal msg="creating container: rpc error: code = Unknown desc = cannot open > /proc/0/uid_map: open /proc/0/uid_map: no such file or directory"
> 
> Which is caused by the fact that `func (s *Server) getSandboxIDMappings(sb *libsandbox.Sandbox) (*idtools.IDMappings, error) { ` is getting the mappings from infra container's /proc/$PID/uid_map and gid_map. If there is no infra container, I'm not sure what to do in this case.

this is partially the problem. This particular test case tests with us using `--manage-ns-lifecycle=true` then restarting crio with `--manage-ns-lifecycle=false`. 

In this case, we should *not* assume the infra container is present, but use the namespace that was pinned before.

can we try with
```
diff --git a/server/sandbox_run_linux.go b/server/sandbox_run_linux.go 
index d952b03d8..cafe60bd9 100644 
--- a/server/sandbox_run_linux.go 
+++ b/server/sandbox_run_linux.go 
@@ -247,7 +247,7 @@ func (s *Server) configureSandboxIDMappings(mode string, sc *pb.LinuxSandboxSecu 
  
 func (s *Server) getSandboxIDMappings(sb *libsandbox.Sandbox) (*idtools.IDMappings, error) { 
        ic := sb.InfraContainer() 
-       if ic != nil { 
+       if ic != nil && !ic.Spoofed() { 
                mappings := ic.IDMappings() 
                if mappings != nil { 
                        return mappings, nil
```

giuseppe:
> In this case, we should _not_ assume the infra container is present, but use the namespace that was pinned before.

In that case, how could we read the mappings back?  Should we fork a temporary process that joins the user namespace and read the mappings from `/proc/self/*id_map`?

haircommander:
> > In this case, we should _not_ assume the infra container is present, but use the namespace that was pinned before.
> 
> In that case, how could we read the mappings back? Should we fork a temporary process that joins the user namespace and read the mappings from `/proc/self/*id_map`?

how do we do it in the 100% manage-ns-lifecycle=true case? If reading {u,g}id_map is the only way, then we may need to fork to get the values

kolyshkin:
OK the fix was actually pretty easy: https://github.com/cri-o/cri-o/pull/4333/commits/21daa969b28b0cb1ffccb8154a680659de0efa32

(oh, and I opened a whole PR as an alternative to this one: #4333; PTAL)

giuseppe:
closing in favor of #4333

Issue #4343, opened 2020-11-03
Make script to install cri-o, for one-liner install instruction


afbjorklund:
It would be nice if there was an official script to install cri-o, similar to get.docker.com

https://github.com/docker/docker-install

That way, one could run that instead of having to copy/paste different instructions ?

e.g. https://get.cri-o.io (or something)

----

https://github.com/cri-o/cri-o/blob/master/install.md#install-packaged-versions-of-cri-o

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cri-o

saschagrunert:
Do we wanna use the static binary bundle for this? Or a swith-case for installing the right package on the distribution running the script?

afbjorklund:
I only did the apt and yum vendor packages (as a proof-of-concept), so didn't do the system packages (for suse) or the static binary

The background was that we were moving from the "boot2docker" concept to a more generic cloud image concept for minikube

https://github.com/kubernetes/minikube/issues/4733

https://github.com/kubernetes/minikube/pull/9545

By default, it will then bootstrap each node by install docker on it (for use as the container runtime for kubernetes, that is):

```go
func installDockerGeneric(p Provisioner, baseURL string) error {
	// install docker - until cloudinit we use ubuntu everywhere so we
	// just install it using the docker repos
	if output, err := p.SSHCommand(fmt.Sprintf("if ! type docker; then curl -sSL %s | sh -; fi", baseURL)); err != nil {
		return fmt.Errorf("error installing docker: %s", output)
	}

	return nil
}
```

So it would be nice if there was such a script available also for cri-o, if wanting to use that as the CRI instead ?

Note that the code above won't work (yet), since I didn't include any `sudo` in the commands where needed...

As long as that static binary works, why not ?

Currently there's also some packages missing...

afbjorklund:
@saschagrunert : there are actually no instructions on how to _use_ the static binaries, on [install.md](https://github.com/cri-o/cri-o/blob/master/install.md) ?

Seems to only show how to build them, says that they are actually meant for the nix package manager ?

afbjorklund:
As for distributions, I was mostly going off the officially supported list for kubeadm on:
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/

Which basically just means ubuntu 16.04+ (debian 9+) and centos 7+ (fedora 25+)
For centos/fedora we did the preparation by turning off selinux and disabling cgroups v2

[...]

PR #4344, opened 2020-11-03
Convert install.md text into install.sh script

afbjorklund:
Try to make an executable version of the install:

https://github.com/cri-o/cri-o/blob/master/install.md#install-packaged-versions-of-cri-o

Use `sudo` for getting the required root access

Quote things at random for shellcheck/shfmt...

For #4343


afbjorklund:
After the first pathetic attempts, it seems to work OK in centos and ubuntu containers (after first installing `curl` and `gnupg`)

However, not all versions of cri-o are available for operating systems - so maybe $VERSION should be a parameter ($1) ?

install.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh -ex
```
saschagrunert, line 1: Let's link this script in the readme/install instructions.

haircommander:
> After the first pathetic attempts, it seems to work OK in centos and ubuntu containers (after first installing `curl` and `gnupg`)
> 
> However, not all versions of cri-o are available for operating systems - so maybe $VERSION should be a parameter ($1) ?

yeah I definitely think $VERSION should be a parameter

haircommander:
thanks for picking this up @afbjorklund !

afbjorklund:
> yeah I definitely think $VERSION should be a parameter

Done.

haircommander:
/retest
/approve
LGTM, there's one outstanding nit about adding a pointer from the README and install docs

afbjorklund:
> LGTM, there's one outstanding nit about adding a pointer from the README and install docs

Were you able to find anywhere to host it ? Or would it have to use some "raw" github URL ?

haircommander:
> > LGTM, there's one outstanding nit about adding a pointer from the README and install docs
> 
> Were you able to find anywhere to host it ? Or would it have to use some "raw" github URL ?
> 
> Like https://get.docker.com

for starters, it can be self referential. like if we're updating one of the readme files, we can put a link to `(/install.sh)`

afbjorklund:
I think GitHub will add the usual frame to it, if you just link it in the markdown ?

Like so:[ install.sh](https://github.com/afbjorklund/cri-o/blob/install/install.sh) 

So you would need something like:

`curl -sSL https://raw.githubusercontent.com/afbjorklund/cri-o/install/install.sh | sh -`

haircommander:
I think we should just have a reference to it for now:
```
You can install from package managers with our [install script](/install.md), or by following the steps below:
```

after the title here https://github.com/cri-o/cri-o/blob/master/install.md#install-packaged-versions-of-cri-o

afbjorklund:
> I think we should just have a reference to it

What would be the use of having a _second_ install.md ?

The whole idea was to have something that was executable

haircommander:
haha I think I meant
```
You can install from package managers with our [install script](/install.sh), or by following the steps below:
```

haircommander:
thinking about this more, I Think we can add references to docs as a follow up

PTAL @umohnani8 @kolyshkin @saschagrunert @mrunalp

afbjorklund:
> haha I think I meant
> 
> ```
> You can install from package managers with our [install script](/install.sh), or by following the steps below:
> ```

It's still just a link to a web page, though.

afbjorklund:
Ultimately we would like something (an URL) that can be fed into the "Provisioner" of the libmachine replacement.

The role of the provisioner is to install the container runtime on the node, for later bootstrapping of Kubernetes...

Right now it is hardcoded to Docker, and only allows different _versions_ of it.

But we want to have equal support for docker and containerd - and also cri-o.

haircommander:
we can add it here https://cri-o.github.io/cri-o/ under the CRI-O tab, above the release-notes. I am not super sure how to do that though, and @saschagrunert is still out on vacation I believe

TomSweeneyRedHat:
The script itself LGTM.

I'm not sure if this is being suggested or not, but I would not completely replace install.md with this script.  For instance, there's a lot of good info on how to install/setup your sandbox in the install.md that I'd not want to lose.

PR #3881, opened 2020-06-16
[1.11] server/ContainerStatus: don't lock

haircommander:
r.ContainerStatus(c) is using opLock which can sometimes be held for minutes
in e.g. runtime.ContainerStop. The opLock is there so that
operations changing the container state won't clash. There
is no reason to wait for the lock here.

The side effect is, ContainerStatus will return immediately,
not e.g. after container has finished starting or stopping,
and show the old, rather than new, state.

adapted from https://github.com/cri-o/cri-o/pull/3457

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

Signed-off-by: Peter Hunt <pehunt@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

#### Does this PR introduce a user-facing change?

```release-note

```

PR #4352, opened 2020-11-06
[3.11] container status: check whether the container is stopped

haircommander:
before updating state

We currently do not hold the lock to get the initial state of the container when getting container status.
However, we should also only attempt to reset that state if we know the container is stopped

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

```release-note
none
```

### DIFF
--- a/server/container_status.go
+++ b/server/container_status.go
@@ -59,7 +59,7 @@ func (s *Server) ContainerStatus(ctx context.Context, req *pb.ContainerStatusReq
 
 	// If we defaulted to exit code -1 earlier then we attempt to
 	// get the exit code from the exit file again.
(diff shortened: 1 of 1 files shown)


haircommander:
oh wait, nvm, I don't actually think we need this

haircommander:
for posterity, someone ran into the issue that was supposed to be fixed by https://github.com/cri-o/cri-o/pull/3881.

This PR would be another nail in the coffin against that situation, but actually, their version just didn't have that PR

PR #4130, opened 2020-08-28
Revert "dual stack portmap support"

aojea:
This reverts commit f97ad7fd3fb3f0901e70a45e7caa57724715ce6c.

This PR uncovered a serial of issues in the portmap behavior that is present upstream too-

When a pod with a PortMapping is created, the portmapper opens a socket in ALL addresses to avoid that another applications can "steal" the port.

https://github.com/cri-o/cri-o/blob/master/vendor/k8s.io/kubernetes/pkg/kubelet/dockershim/network/hostport/hostport.go#L68-L84

However, as the test indicates, you should be able to user several host-ports in the same host on different addresses, this means that ALWAYS will fail to bind the first time.

It was working because, the second pod fails the first time

> Aug 28 15:03:59 crio-worker crio[18574]: time="2020-08-28 15:03:59.167098356Z" level=info msg="TEST adding hostport ips: 10.244.2.44 portmaps: &{ k8s_pod2_sched-pred-5589_0aa5a842-ed92-47b9-8523-3782181c83fd_0 [0xc000a825c0] false 10.244.2.44}" file="server/sandbox_network.go:89" id=3e9adfba-a430-4a1b-a513-eaee34cb2e4e name=/runtime.v1alpha2.RuntimeService/RunPodSandbox

if the pod fails to setup the network, it retries removing the failed pod

> Aug 28 15:03:59 crio-worker crio[18574]: time="2020-08-28 15:03:59.169900605Z" level=info msg="NetworkStart: stopping network for sandbox 7d2d754771fbaa3c7a337540cfb95f8cb31b4e22d1305c453ea8284df7404ee0" file="server/sandbox_network.go:41" id=3e9adfba-a430-4a1b-a513-eaee34cb2e4e name=/runtime.v1alpha2.RuntimeService/RunPodSandbox

and removing the portmap (of the second pod) and the process binding the port

> Aug 28 15:03:59 crio-worker crio[18574]: time="2020-08-28 15:03:59.169947955Z" level=info msg="TEST removing hostport ips: [] portmaps: [0xc000a825c0]" file="server/sandbox_network.go:152" id=3e9adfba-a430-4a1b-a513-eaee34cb2e4e name=/runtime.v1alpha2.RuntimeService/RunPodSandbox

The remove host-port does not have IP address, that's what I've changed in my PR, causing that we never release the socket binding to the hostport.

On the second try, it goes through, because it can bind the socket

[...]

PR #4361, opened 2020-11-10
Portmap support for dual-stack

aojea:
#### What type of PR is this?

> /kind bug
> /kind dependency-change

current portmap is hardcoded to use IPv4 only. We can add IPv6
support just detecting the IP family from the network configuration,
however, this won't work for dual stack.

We can add a portManager handler per IP family and use the corresponding
one based on the sandbox registered IPs.

However, due to a bug in the hostport code vendored from upstream, we had
to revert previous implementation https://github.com/cri-o/cri-o/pull/4130

Since upstream is not likely going to fix the code, because it is being depreacted,
we copy the code directly in crio-o.

Signed-off-by: Antonio Ojea aojea@redhat.com

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1872128

Kubernetes upstream hostport bug
https://github.com/kubernetes/kubernetes/pull/94382

```release-note
crio-o portMapping dual-stack support
```


aojea:
/hold
/assign @saschagrunert @dcbw 
I still need to test it better, just FYI will ping you once it is ready for review

aojea:
@mrunalp is the test 
```
validates that there is no conflict between pods with same hostPort but different hostIP and protocol
```
skipped somewhere, I can't see it running in ci/prow/e2e-aws  and it was the test that broke last time.

saschagrunert (approved):
LGTM

Can we add a README.md into the hostport package that this is copied and modified from k/k?

internal/hostport/README.md
```diff
@@ -0,0 +1,8 @@
+# Host port manager
```
TomSweeneyRedHat, line 1: Do we usually include the CRI-O logo in the top of the README.md files?  If not, we should for context.

internal/hostport/README.md
```diff
@@ -0,0 +1,8 @@
...
+
+Kubernetes started the deprecation of the dockershim code and there is a risk that the upstream code will not
+receive any new patches or fixes.
+
+This package come from https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/dockershim/network/hostport
```
TomSweeneyRedHat, line 6:
```suggestion
This package comes from https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/dockershim/network/hostport
```

danwinship:
> Since upstream is not likely going to fix the code, because it is being depreacted,

I don't think that is accurate. There is benefit in having a canonical correct implementation of that code, and some `k8s.io` module is obviously the right place to do that. There is uncertainty about exactly how to do that given the code's dependencies though.

aojea:
> > Since upstream is not likely going to fix the code, because it is being depreacted,
> 
> I don't think that is accurate. There is benefit in having a canonical correct implementation of that code, and some `k8s.io` module is obviously the right place to do that. There is uncertainty about exactly how to do that given the code's dependencies though.

The upstream fix is there since Sept, 1st https://github.com/kubernetes/kubernetes/pull/94382

Absolutely agree with you on having the code in k8s.io, and I've also tried that https://github.com/kubernetes/utils/pull/192
but that is not going to happen on time to fix this bug :shrug:

internal/hostport/README.md
aojea: added

mrunalp:
@aojea https://github.com/cri-o/cri-o/blob/master/contrib/test/integration/e2e.yml#L22 is the skip list for `e2e_rhel` and `e2e_fedora` . e2e-aws comes from `openshift/release`

aojea:
dockershim deprecation clock started ...
https://github.com/kubernetes/kubernetes/pull/94624

aojea:
/test e2e-aws

heh 142 tests failed, previous run was about 20, it seems environmental and not related to the PR

aojea:
> Failed to create injector pod: timed out waiting for the condition
/test e2e-aws


lots of test failed but I don't see a relation, this only touches hostport that are not used by most of the tests

internal/lib/container_server.go
```diff
@@ -24,7 +25,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
-	"k8s.io/kubernetes/pkg/kubelet/dockershim/network/hostport"
```
danwinship, line 27: It would be better to have one commit switching over to the new internal hostport manager (and removing the unused files) and then a second commit adding dual-stack support. Among other things it may make it easier to revert back to the upstream hostport manager when/if that gets updated

internal/hostport/README.md
```diff
@@ -0,0 +1,10 @@
+![CRI-O logo](https://github.com/cri-o/cri-o/blob/master/logo/crio-logo.svg?raw=true)
+# Host port manager
+
+CRIO-O was vendoring the host port manager implementation from Kubernetes.
+Kubernetes started the deprecation of the dockershim code and there is a risk that the upstream code will not
+receive any new patches or fixes.
```
danwinship, line 6: also might want a rewording

danwinship (commented):
>     fork hosport kubernetes code
>     
>     the code is being abandoned in the upstream repo, dockershim
>     moved to the portmap CNI implementation and it seems that only
>     the legacy kubnet is using it.
>     
>     It has an open bug and is not likely to be fixed, so it is better
>     for CRIO to use it and maintain it directly by now.

Commit message needs to be updated; the code is not being abandoned. It's just not likely to be fixed for 1.20.

Also "hosport", "kubnet"

aojea:
> > ```
> > 
> > ```
> 
> 
> Commit message needs to be updated; the code is not being abandoned. It's just not likely to be fixed for 1.20.

The  code is going to be removed next release
https://github.com/kubernetes/kubernetes/pull/94624

internal/lib/container_server.go
aojea: the code is going to be removed
https://github.com/kubernetes/kubernetes/pull/94624

aojea: I mean, is not going to be a revert to the same folder

internal/hostport/README.md
aojea: done

aojea:
The conformance test need to be modified for IPv6, because it uses loopback addreses that doesn't have the same behaviour than IPv4 loopback addresses
https://github.com/kubernetes/kubernetes/pull/96627

I tested it locally with IPv6 and the modified conformance test and it works

internal/hostport/README.md
```diff
@@ -0,0 +1,20 @@
+![CRI-O logo](https://github.com/cri-o/cri-o/blob/master/logo/crio-logo.svg?raw=true)
+# Host Port Manager
+
+The Host Port Managers map ports from the Host to the specified Container Port.
```
TomSweeneyRedHat, line 4:
```suggestion
The Host Port Manager maps ports from the Host to the specified Container Port.
```

internal/hostport/README.md
```diff
@@ -0,0 +1,20 @@
...
+It can also specify the Host IP that should be mapped to the container IP.
+For security reasons, it also open a socket in the host to avoid any process in the system
```
TomSweeneyRedHat, line 6:
```suggestion
For security reasons, it also opens a socket in the host to avoid any process in the system
```

internal/hostport/README.md
```diff
@@ -0,0 +1,20 @@
...
+can interfere with the traffic that is being port mapped.
```
TomSweeneyRedHat, line 7:
```suggestion
that could interfere with the traffic that is being port mapped.
```

internal/hostport/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+![hostport](HostPort.jpg "Container Host Port")
+
+This code is copied from the kubernetes/kubernetes repo
+https://github.com/kubernetes/kubernetes/tree/41533ecec8966636554faba44368a36d5d6c5347/pkg/kubelet/dockershim/network/hostport
```
TomSweeneyRedHat, line 12:
```suggestion
https://github.com/kubernetes/kubernetes/tree/41533ecec8966636554faba44368a36d5d6c5347/pkg/kubelet/dockershim/network/hostport.
```

internal/hostport/README.md
```diff
@@ -0,0 +1,20 @@
...
+since Kubernetes started the deprecation of that part of the code in
```
TomSweeneyRedHat, line 13:
```suggestion
Since Kubernetes started the deprecation of that part of the code in
```

internal/hostport/README.md
```diff
@@ -0,0 +1,20 @@
...
+https://github.com/kubernetes/kubernetes/pull/94624 and is missing some fixes and the features required to operate
```
TomSweeneyRedHat, line 14:
```suggestion
https://github.com/kubernetes/kubernetes/pull/94624, it is missing some fixes and the features required to operate
```

internal/hostport/README.md
```diff
@@ -0,0 +1,20 @@
...
+correctly in dual stack environments.
+
+It has a fix where the host port manager always open a socket in all addresses, instead of leveraging the HostIP field
```
TomSweeneyRedHat, line 17:
```suggestion
 The kubernetes/kubernetes repo has a fix where the host port manager always opens a socket in all addresses, instead of leveraging the HostIP field:
```

aojea:
These tests timeout in 2 different jobs
> Pods Extended [k8s.io] Pod Container Status should never report success for a pending container
but I can't find a relation, any hint how can I debug that?

aojea:
> but I can't find a relation, any hint how can I debug that

It was a perfmance or a race problem, I can't say, but caching the sandbox info it went through

This is ready for review, I tested it and it works local in an ipv6 cluster

server/sandbox_network.go
```diff
@@ -138,12 +146,19 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
-	if err := s.hostportManager.Remove(sb.ID(), &hostport.PodPortMapping{
+	mapping := &hostport.PodPortMapping{
 		Name:         sb.Name(),
 		PortMappings: sb.PortMappings(),
 		HostNetwork:  false,
-	}); err != nil {
-		log.Warnf(ctx, "failed to remove hostport for pod sandbox %s(%s): %v",
+	}
+	// portMapping removal does not need the IP address
+	if err := s.hostportManagerv6.Remove(sb.ID(), mapping); err != nil {
```
danwinship: It looks like `Remove()` will always rewrite the `KUBE-HOSTPORT` chain even if it doesn't actually remove anything from it. Might be worth adding a TODO about that...

server/sandbox_network.go
```diff
@@ -68,17 +67,13 @@ func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs
...
-		// the same host port twice
-		if idx == 0 && len(sb.PortMappings()) > 0 {
-			ip := net.ParseIP(podIP)
-			if ip == nil {
-				return nil, nil, fmt.Errorf("failed to get valid ip address for sandbox %s(%s)", sb.Name(), sb.ID())
-			}
+	for _, podIPConfig := range network.IPs {
+		ip := podIPConfig.Address.IP
+		if ip.To16() == nil {
+			return nil, nil, fmt.Errorf("failed to get valid ip address for sandbox %s(%s)", sb.Name(), sb.ID())
```
danwinship, line 81: `To16` can only fail if the IP is neither IPv4 nor IPv6, in which case it should have failed some check long before this

internal/hostport/meta_hostport_manager.go
```diff
@@ -0,0 +1,67 @@
...
+
+// NewMetaHostportManager creates a new HostPortManager
+func NewMetaHostportManager() HostPortManager {
+	exec := exec.New()
+	// Create IPv4 handler
+	iptInterface := utiliptables.New(exec, utiliptables.ProtocolIPv4)
+	if _, err := iptInterface.EnsureChain(utiliptables.TableNAT, iptablesproxy.KubeMarkMasqChain); err != nil {
+		klog.Warningf("unable to ensure iptables chain: %v", err)
+	}
+	hostportManager := NewHostportManager(iptInterface)
```
danwinship, line 28: might as well call it `hostportManagerv4` now

server/sandbox_network.go
```diff
@@ -79,14 +85,20 @@ func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs
 				IP:           ip,
 				HostNetwork:  false,
 			}
-
+			if utilnet.IsIPv6(ip) && foundIPv6 {
+				continue
+			} else if utilnet.IsIPv6(ip) && !foundIPv6 {
+				foundIPv6 = true
+			} else if foundIPv4 {
```
danwinship, line 99: if utilnet.IsIPv6(ip) {
            if foundIPv6 {
                    continue
            }
            foundIPv6 = true
    }

danwinship (changes requested):
Please squash your internal bugfixes and rewrites

server/sandbox_network.go
aojea: I think that is expected, a kind of reconciliation

aojea:
I had to do some changes to pass the linter,
I've tested locally in an IPv6 cluster with the e2e test fix https://github.com/kubernetes/kubernetes/pull/96627

server/sandbox_network.go
danwinship: no, I mean, it does `iptables-save`, reconciles the state to what it should be, and then does `iptables-restore`. But it does the `iptables-restore` even if it didn't make any changes from what it got from `iptables-save`.

Basically, the code was assuming that you wouldn't ever call it with a mapping that you hadn't previously added, which was more-or-less correct when there was only a single hostportmanager. But now in a single-stack IPv4 cluster, you'll still have a `hostportManagerv6`, and you'll never call `Add` on it, but you'll still call `Remove` on it and make it do unnecessary `iptables-restore`s

aojea: I think this should fix it 
8fee17d4a59894b6e0cacad17bdc9549ec7f3e67

aojea:
/hold cancel
CI is passing

internal/hostport/fake_iptables.go
```diff
@@ -146,6 +146,7 @@ func (f *fakeIPTables) ensureRule(position utiliptables.RulePosition, tableName
 		return true, nil
 	}
 
+	// nolint:gocritic // using a switch statement is not much different
```
kolyshkin, line 149: It will actually look cleaner with switch:
```go
switch(position) {
  case utiliptables.Prepend:
    chain.rules = append([]string{rule}, chain.rules...)
  case utiliptables.Append:
    chain.rules = append(chain.rules, rule)
  default:
    return false, fmt.Errorf("unknown position argument %q", position)
}

File: internal/hostport/hostport.go [hunk 17]
```diff
@@ -52,7 +52,17 @@ type PodPortMapping struct {
 	IP           net.IP
 }
 
+// ipFamily refers to a specific family if not empty, i.e. "4" or "6".
+type ipFamily string
+
+// Constants for valid IPFamily:
+const (
```
kolyshkin, line 59: So, it seems that this commit is too overloaded, i.e. it does too many things:
 - adds IPv6 support
 - fixes nits found by golangci-lint
 - adds tests for hostport pkg
 - switches cri-o to use the new hostport implementation

Perhaps it'll be better to split it into separate commits, in the following order:
1. fix golangci nits
2. add IPv6 support
3. add unit tests
4. switch crio to new hostport

or something like that

aojea: can we go with current status this time? :sweat_smile:

aojea:
/test e2e-agnostic

pod create sandboxes test failures

internal/hostport/hostport.go
aojea: :+1:

aojea: @kolyshkin I split it out, with a small different approach, please let me know if it works for you

haircommander:
@danwinship can we have a final +1 here?

danwinship:
/lgtm
but if you end up needing to do another round, there's still a typo in one of the commit messages: "fork **hosport** kubernetes code"

haircommander:
/approve

@mrunalp @umohnani8 @kolyshkin PTAL

aojea:
/retest
Multus create podsandbox test errors

aojea:
/retest
again failing the pod sandbox creation tests
```
 Failed to create pod sandbox: rpc error: code = Unknown 
```

aojea:
same errors
/retest

aojea:
@haircommander can you run the automatic cherry-pick with the bot for 1.20 and 1.19?
xref: https://bugzilla.redhat.com/show_bug.cgi?id=1872128

haircommander:
1.20 will have it automatically backported

/cherry-pick release-1.19

haircommander:
alas

aojea:
I will take it, thanks

PR #4362, opened 2020-11-11
honor the grace period for a container on sandbox deletion

rphillips:
#### What type of PR is this?
/kind bug

When a sandbox is stopped the default container timeout is 10 seconds. This PR adds support to honor the terminationGracePeriod to allow a container to shutdown correctly.

```release-note
NONE
```

### DIFF
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -4,6 +4,7 @@ package server
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


rphillips:
Talked to Mrunal, and the Kubelet's StopPodSandbox says this function in crio terminates the containers without the grace period on purpose. Closing...

PR #4367, opened 2020-11-12
[DO NOT MERGE] test e2e-aws failures

umohnani8:

```release-note
none
```

### DIFF
--- a/hack/build-rpms.sh
+++ b/hack/build-rpms.sh
@@ -28,6 +28,11 @@ tar czf "${rpm_tmp_dir}/SOURCES/${OS_RPM_NAME}-test.tar.gz" \
     .
 cp -r "${ci_data}/." "${rpm_tmp_dir}/SOURCES"
 
+# for file in /etc/yum.repos.d/*; do sed "s#vault.centos.org/centos/7#vault.centos.org/centos/7.8.2003#g" $file -i; done
+echo $releasever
+export releasever=7.8.2003
+echo $releasever
+for file in /etc/yum.repos.d/*; do cat $file; done
 yum-builddep -y "${OS_RPM_SPECFILE}"
 
 rpmbuild -ba "${OS_RPM_SPECFILE}" \


hack/build-rpms.sh
```diff
@@ -28,6 +28,11 @@ tar czf "${rpm_tmp_dir}/SOURCES/${OS_RPM_NAME}-test.tar.gz" \
     .
 cp -r "${ci_data}/." "${rpm_tmp_dir}/SOURCES"
 
+# for file in /etc/yum.repos.d/*; do sed "s#vault.centos.org/centos/7#vault.centos.org/centos/7.8.2003#g" $file -i; done
+echo $releasever
```
fgiudici, line 32: From the error in the CI job seems the script is run with set -u
```suggestion
echo ${releasever-""}
```

fidencio:
JTFR, I've opened https://pagure.io/centos-infra/issue/156 to track the status of the issue on CentOS Infra side.

fidencio:
Let me quite a comment from Fabian, in the issue mentioned above:
```
Hi,

Some  just reported the same issue on centos-devel list, and it seems that @hughesjr indeed forgot to update metadata. (and also some real src.rpm should also have landed there, probably still in his todo list for post-GA activities)

I've created empty repodata for now, so that shouldn't "block" you but I'll let this ticket assigned so that he can confirm and close it when properly implemented
```

I've re-triggered the CI on #4356 and it seems to have passed the problematic part.  So, we should be in a good shape about this issue now.

fidencio:
@umohnani8, we can drop this patch as the workaround provided by Fabian did fix the issues with CentOS. \o/.

And now we know the proper way to poke CentOS Infra (via a ticket) in case this happens again in the future.

umohnani8:
Yup! I was just trying to see if updating the version helps. Thanks for the update! Just saw that is has been fixed on the centos end, so closing.

PR #4370, opened 2020-11-13
not enforce seccomp profiles if disabled

aojea:
kubernetes forces sandbox to use runtime/default, however, users
may not want to use seccomp.

Allow to use pods with runtime/default profiles if users run
with seccomp disabled and they don' t want to enforce seccomp

#### What type of PR is this?

/kind bug

```release-note
CRIO allow to run pods with default runtime profile in the Pod.Spec, if seccomp is disabled
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -571,9 +571,17 @@ func (s *Server) setupSeccomp(ctx context.Context, specgen *generate.Generator, 
 			specgen.Config.Linux.Seccomp = nil
 			return nil
 		}
+		// default to SeccompProfileRuntimeDefault if user sets UseDefaultWhenEmpty
 		profile = k8sV1.SeccompProfileRuntimeDefault
 	}
+	// kubelet defaults sandboxes to run as `runtime/default`, we consider the default profile as unconfined if Seccomp disabled
+	// https://github.com/kubernetes/kubernetes/blob/12d9183da03d86c65f9f17e3e28be3c7c18ed22a/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go#L162-L163
 	if s.Config().Seccomp().IsDisabled() {
+		if profile == k8sV1.SeccompProfileRuntimeDefault {
+			// running w/o seccomp, aka unconfined
+			specgen.Config.Linux.Seccomp = nil
+			return nil
+		}
 		if profile != k8sV1.SeccompProfileNameUnconfined {
 			return fmt.Errorf("seccomp is not enabled in your kernel, cannot run with a profile")
 		}


aojea:
/assign @haircommander @saschagrunert @mrunalp 

Talked about this with Sascha in slack, the thing is that without this CRIO has a different behaviour than in 1.19,

This approach works,  but I don' t have clear what is the expected behaviour, so I open this for discussion.

aojea:
basically what I' m asking is to document this table in crio for users, I can do it as part of this PR, but I need your help to fill the gaps


| Scenario | API Server Result | Kubelet Result |
|--------------------------------------------------------------|------------------------------|----------------------------------------------------------------------------------------------|
| 1) Using custom or `runtime/default` profile when container runtime does not support seccomp. | Pod created | The outcome is container runtime dependent. In this scenario containers may 1) fail to start or 2) run normally without having its policies enforced. |
| 2) Using custom or `runtime/default` profile that restricts syscalls a container is trying to make. | Pod created | The outcome is workload and seccomp dependent. In this scenario containers may 1) fail to start, 2) misbehave or 3) log violations. |
| 3) Using custom profile that relies on unsupported or invalid seccomp actions (i.e. `SCPM_ACT_LOG` in versions earlier than 4.14 Linux kernel). | Pod created | Containers fail to start. Retry respecting RestartPolicy and back-off delay. |
| 4) Using custom profile that does not exist on the node's disk. | Pod created | Containers fail to start. Retry respecting RestartPolicy and back-off delay. |
| 5) Using a non supported runtime profile (i.e. `runtime/default-audit`). | Pod **not** created. | N/A |
| 6) Using localhost profile with invalid name (i.e. "/etc/profilename") | Pod **not** created. | N/A |

https://github.com/kubernetes/enhancements/blob/1274d448ea6fc09b976428bfc77237746ea1bb56/keps/sig-node/20190717-seccomp-ga.md#failure-and-fallback-strategy

server/container_create.go
```diff
@@ -571,9 +571,17 @@ func (s *Server) setupSeccomp(ctx context.Context, specgen *generate.Generator,
 			specgen.Config.Linux.Seccomp = nil
 			return nil
 		}
+		// default to SeccompProfileRuntimeDefault if user sets UseDefaultWhenEmpty
 		profile = k8sV1.SeccompProfileRuntimeDefault
 	}
+	// kubelet forces sandbox to run as `runtime/default` , if !UseDefaultWhenEmpty we consider it as unconfined if Seccomp disabled
+	// https://github.com/kubernetes/kubernetes/blob/12d9183da03d86c65f9f17e3e28be3c7c18ed22a/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go#L162-L163
 	if s.Config().Seccomp().IsDisabled() {
+		if profile == k8sV1.SeccompProfileRuntimeDefault && !s.Config().Seccomp().UseDefaultWhenEmpty() {
```
saschagrunert, line 580: I think we can apply this independently from the `UseDefaultWhenEmpty` option.
```suggestion
		if profile == k8sV1.SeccompProfileRuntimeDefault {
```

aojea: done

aojea:
/retest
some of the jobs doesn't have any output

aojea:
@saschagrunert I've tried to add some unit test to increase the coverage but I can't see an easy way to mock the config for the method `setupSeccomp()`,
is it a problem the 2 codecov job failing? 
any guidance if affirmative on how to add the unit test to increase the coverage?

haircommander:
> @saschagrunert I've tried to add some unit test to increase the coverage but I can't see an easy way to mock the config for the method `setupSeccomp()`,
> is it a problem the 2 codecov job failing?
> any guidance if affirmative on how to add the unit test to increase the coverage?

codecov is non blocking, but if you feel like figuring them out, it we'd love the extra unit tests! You could override the seccomp config for the tests with a public setter that's only added with the `test` build flag like in https://github.com/cri-o/cri-o/blob/master/internal/oci/container_test_inject.go

or you can just leave it as is :sweat_smile: 
/approve

LGTM

aojea:
what is missing here?

haircommander:
the final lgtm :smile: 

@mrunalp @umohnani8 @fidencio @saschagrunert @kolyshkin @giuseppe PTAL

saschagrunert (approved):
/lgtm

aojea:
/retest
> Kubernetes e2e suite: [k8s.io] [sig-node] Pods Extended [k8s.io] Pod Container Status should never report success for a pending container expand_more

PR #4372, opened 2020-11-16
WIP: Switch to CRI v1 protocol

saschagrunert:
#### What type of PR is this?

/kind api-change

We now use the graduated CRI v1 protocol instead of the v1alpha2.

None
None

```release-note
Changed Container Runtime Interface implementation to only support `v1` instead of `v1alpha2`.
```


v1/lib/rename.go
```diff
@@ -0,0 +1,116 @@
+package lib
```
haircommander, line 1: why'd we drop `internal`? doesn't that mean other libraries can vendor `v1/lib` which could mess with our versioning?

saschagrunert: Its impossible to import an `internal` package from a sub-package of the project from another part of the project.

saschagrunert:
Will open a new one

PR #4373, opened 2020-11-16
CI: remove unused job, move some jobs to github-actions

kolyshkin:
#### What type of PR is this?

/kind ci

This moves a few short and straightforward CI jobs
out of circleci pipeline and into github actions.

The motivation is circleci's parallelism limit, which appears
to be recently introduced. Because of that, jobs like integration
testing are often not started right away, since the jobs like
shfmt and shellcheck are run first.

The other benefit is now commits are annotated with warnings
from e.g. shellcheck. This is not done for all linters though.

Now, I can't demo the end result here in this repo, but
you can take a look at how it looks like here: https://github.com/kolyshkin/cri-o/pull/1
In particular, take a look at annotations in commits.

Below is detailed description, copy-pasted from commits.

1. circleci: rm build-test-binaries job
    
    It looks like it is obsoleted by 964245f943e4fe.
    
    Now when the circleci paralellism is limited, this might help up to
    speed up CI completion.

2. ci: move golangci-lint from circleci to GH actions
    
    It should be faster, and the side benefit of using github
    actions is linter warnings will be shown in the diff
    as annotations (similar to review comments).
    
    To make it work without installing additional packages,
    fix the build tags in .golangci.yml. In particular,
    disable seccomp and device mapper, as they need some C
    headers to be compiled.
    
    Note that this is merely to be able to compile packages
    used by crio, and since we are only linting crio itself,
    and not the packages it imports, changing the tags not
    used by crio doesn't matter much (if at all).
    
    Note we still have `lint` target in Makefile, to be used
    by developers (but not for CI).

3. ci: move shellcheck from circleci to github actions
    
    Note that
    
    1. No golang caching is required (as it was done in circleci)
       since we do not build any go code, and it does not look like
       it make sense to cache a single binary we donwload from github.
    
    2. Use a problem matcher, so the PRs will be annotated with
       shellcheck warnings.

4. ci: move shfmt from circleci to github actions
    
    shfmt output is unified diff, so it's hard to apply a problem matcher.
    Besides, I think the unified diff format is the best (well, ideally
    we'd use something like github "suggestion" thing, but I was not able
    to find a way of doing it).

5. ci: move docs-validation to github actions

None

```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,6 @@ workflows:
             - integration-static
             - integration-userns
             - unit-tests
-      - shellcheck
       - shfmt
       - unit-tests
       - upload-artifacts:
@@ -426,22 +425,6 @@ jobs:
       - store_artifacts:
           path: bundle
           destination: bundle
-
-  shellcheck:
-    executor: container
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - v1-shellcheck-{{ checksum "go.sum" }}
-      - run:
-          name: Check shell files
-          command: make shellcheck
-      - save_cache:
-          key: v1-shellcheck-{{ checksum "go.sum" }}
-          paths:
-            - *gocache
-            - build/bin/shellcheck
 
   shfmt:
     executor: container
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,7 +70,6 @@ workflows:
       - docs-generation:
           requires:
             - build
-      - docs-validation
       - integration
       - integration:
           name: integration-critest
@@ -252,21 +251,6 @@ jobs:
             hack/tree_status.sh
       - save_cache:
           key: v1-docs-generation-{{ checksum "go.sum" }}
-          paths:
-            - *gocache
-
-  docs-validation:
-    executor: container
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - v1-docs-validation-{{ checksum "go.sum" }}
-      - run:
-          name: doc validation
-          command: make docs-validation
-      - save_cache:
-          key: v1-docs-validation-{{ checksum "go.sum" }}
           paths:
             - *gocache
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,6 @@ workflows:
             - integration-static
             - integration-userns
             - unit-tests
-      - shfmt
       - unit-tests
       - upload-artifacts:
           requires:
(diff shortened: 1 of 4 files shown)


saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM
Assuming happy tests.

kolyshkin:
So, we have this check called "tide" which makes sure that all the required checks are passing. Alas I have no idea how to make it check the newly added github-actions jobs. Need some help here.

haircommander:
unfortunately, I believe tide is only tied to the openshift-jenkins spawned jobs. The documentation for tide is here https://github.com/kubernetes/test-infra/tree/master/prow/tide, maybe there's some knob to configure a job as required even if the openshift ci didn't create it?

saschagrunert:
> unfortunately, I believe tide is only tied to the openshift-jenkins spawned jobs. The documentation for tide is here https://github.com/kubernetes/test-infra/tree/master/prow/tide, maybe there's some knob to configure a job as required even if the openshift ci didn't create it?

The tide configuration can be found in openshift/release. For CRI-O we decided to skip unknown contexts (tests) per default there: 
https://github.com/openshift/release/blob/59b1a7634274955b7596436ed67b4828a974928c/core-services/prow/02_config/_config.yaml#L604-L607

The main reason for this is that codecov was blocking tide in the past. I think we might solve this by adding the github actions contexts there:
https://github.com/openshift/release/blob/59b1a7634274955b7596436ed67b4828a974928c/core-services/prow/02_config/_config.yaml#L49-L52

Then those contexts/tests would not be unknown any more and tide requires them before merge.

kolyshkin:
> The tide configuration can be found in openshift/release. For CRI-O we decided to skip unknown contexts (tests) per default there:
https://github.com/openshift/release/blob/59b1a7634274955b7596436ed67b4828a974928c/core-services/prow/02_config/_config.yaml#L604-L607

Nice, so that seems possible. I think we should require all the 4 jobs that are moved (which should be done after we merge this PR).

kolyshkin:
> I think we should require all the 4 jobs that are moved (which should be done after we merge this PR).

Opened https://github.com/openshift/release/pull/13708

kolyshkin:
> Opened [openshift/release#13708](https://github.com/openshift/release/pull/13708)

Alas, it's not working as expected (https://github.com/openshift/release/pull/13708#issuecomment-729934795).

Opened a PR to revert (https://github.com/openshift/release/pull/13736).

PR #4376, opened 2020-11-17
Add a performance hook banning network soft IRQs

marcel-apf:
#### What type of PR is this?
/kind feature

Add another performance hook to the high-performance runtime banning network soft IRQs.
    
Some guaranteed pods may have stricter performance constrains requiring network soft IRQs to not run on pod's CPUs.
Annotate those pods with "rps-load-balancing.crio.io: true".

Enable RPS on all the existing network devices (at the time the pod is created) masking out the pod's CPUs during container creation, restoring the previous configuration upon pod termination.

None

None

Signed-off-by: Marcel Apfelbaum <marcel@redhat.com>

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks.go
@@ -31,10 +31,13 @@ const (
 	annotationCPULoadBalancing = "cpu-load-balancing.crio.io"
 	annotationCPUQuota         = "cpu-quota.crio.io"
 	annotationIRQLoadBalancing = "irq-load-balancing.crio.io"
+	annotationRPSLoadBalancing = "rps-load-balancing.crio.io"
 	annotationTrue             = "true"
 	schedDomainDir             = "/proc/sys/kernel/sched_domain"
 	irqSmpAffinityProcFile     = "/proc/irq/default_smp_affinity"
+	onlineCPUsSysFile          = "/sys/devices/system/cpu/online"
 	cgroupMountPoint           = "/sys/fs/cgroup"
+	sysDevicesDir              = "/sys/devices/"
 )
 
 // HighPerformanceHooks used to run additional hooks that will configure a system for the latency sensitive workloads
@@ -81,6 +84,13 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 			return errors.Wrap(err, "set CPU CFS quota")
 		}
 	}
+	// enable RPS not allowing the container CPUs to be used by the existing net devices soft IRQs
+	if shouldRPSLoadBalancingBeEnabled(s.Annotations()) {
+		log.Infof(ctx, "Enable RPS excluding the container CPUs usage for container %q", c.ID())
+		if err := setRPSLoadBalancing(c, sysDevicesDir, onlineCPUsSysFile, true); err != nil {
+			return errors.Wrap(err, "set RPS load balancing")
+		}
+	}
 
 	return nil
 }
@@ -115,6 +125,13 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s 
 	}
 	// no need to reverse the cgroup CPU CFS quota setting as the pod cgroup will be deleted anyway
 
+	// Allow the container CPUs to be used by the net devices soft IRQs
+	if shouldRPSLoadBalancingBeEnabled(s.Annotations()) {
+		if err := setRPSLoadBalancing(c, sysDevicesDir, onlineCPUsSysFile, false); err != nil {
+			return errors.Wrap(err, "set RPS load balancing")
+		}
+	}
+
 	return nil
 }
 
@@ -128,6 +145,10 @@ func shouldCPUQuotaBeDisabled(annotations fields.Set) bool {
 
 func shouldIRQLoadBalancingBeDisabled(annotations fields.Set) bool {
 	return annotations[annotationIRQLoadBalancing] == annotationTrue
+}
+
+func shouldRPSLoadBalancingBeEnabled(annotations fields.Set) bool {
+	return annotations[annotationRPSLoadBalancing] == annotationTrue
 }
 
 func isCgroupParentBurstable(s *sandbox.Sandbox) bool {
@@ -288,4 +309,64 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
 	}
 
 	return nil
+}
+
+func setRPSLoadBalancing(c *oci.Container, sysDevicesDir, onlineCPUsSysFile string, enable bool) error {
+	lspec := c.Spec().Linux
+	if lspec == nil ||
+		lspec.Resources == nil ||
+		lspec.Resources.CPU == nil ||
+		lspec.Resources.CPU.Cpus == "" {
+		return errors.Errorf("cannot find container %s CPUs", c.ID())
+	}
+
+	content, err := ioutil.ReadFile(onlineCPUsSysFile)
+	if err != nil {
+		return err
+	}
+	cpus := strings.Trim(string(content), "\n")
+
+	machineCPUs, err := cpuset.Parse(cpus)
+	if err != nil {
+		return err
+	}
+
+	machineCPUsMask, _, err := computeCPUmask(cpus, "", true)
+	if err != nil {
+		return err
+	}
+
+	return filepath.Walk(sysDevicesDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.Mode().IsRegular() || info.Name() != "rps_cpus" {
+			return nil
+		}
+
+		content, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		rps := strings.TrimSpace(string(content))
+		if rps == "0" {
+			rps = machineCPUsMask
+		}
+
+		newRPS, _, err := computeCPUmask(lspec.Resources.CPU.Cpus, rps, !enable)
+		if err != nil {
+			return err
+		}
+
+		newRPSCPUs, err := cpuMaskToCPUSet(newRPS)
+		if err != nil {
+			return err
+		}
+
+		if newRPSCPUs.Equals(machineCPUs) {
+			newRPS = "0"
+		}
+		return ioutil.WriteFile(path, []byte(newRPS), 0o644)
+	})
 }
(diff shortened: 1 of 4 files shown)


internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -81,6 +83,16 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
...
 	}
+	// enable RPS not allowing the container CPUs to be used by the existing net devices soft IRQs
+	if shouldRPSLoadBalancingBeEnabled(s.Annotations()) {
+		log.Infof(ctx, "enable RPS excluding the container CPUs usage for container %q", c.ID())
+		ncpus, err := machineCPUsCount()
+		if err == nil {
+			if err := setRPSLoadBalancing(c, ncpus, sysDevicesDir, true); err != nil {
+				return errors.Wrap(err, "set RPS load balancing")
+			}
+		} // else - can't count machine cpus; the pod should still start
```
fromanirh, line 94: maybe add a log here?

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -114,6 +126,15 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s
 		}
 	}
 	// no need to reverse the cgroup CPU CFS quota setting as the pod cgroup will be deleted anyway
+	// Allow the container CPUs to be used by the net devices soft IRQs
```
MarSik, line 129: I would probably add an empty line here, this way the two unrelated comments can be confusing

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -211,7 +236,7 @@ func setIRQLoadBalancing(c *oci.Container, enable bool, irqSmpAffinityFile strin
 		return err
 	}
 	currentIRQSMPSetting := strings.TrimSpace(string(content))
-	newIRQSMPSetting, newIRQBalanceSetting, err := UpdateIRQSmpAffinityMask(lspec.Resources.CPU.Cpus, currentIRQSMPSetting, enable)
+	newIRQSMPSetting, newIRQBalanceSetting, err := filterCPUs(lspec.Resources.CPU.Cpus, currentIRQSMPSetting, enable)
```
MarSik, line 239: Is filterCPUs descriptive enough?

MarSik: What about using something like computeCPUmask or so?

internal/runtimehandlerhooks/high_performance_hooks.go
marcel-apf: > maybe add a log here? should not really happen that we can't learn which cpus are available

Will do, sure, I'll add a warn message.

internal/runtimehandlerhooks/high_performance_hooks.go
marcel-apf: ok, thanks

internal/runtimehandlerhooks/high_performance_hooks.go
marcel-apf: sounds good, will change

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -289,3 +314,54 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+		lspec.Resources.CPU.Cpus == "" {
+		return errors.Errorf("find container %s CPUs", c.ID())
+	}
+
+	machineCPUs, err := cpuset.Parse(fmt.Sprintf("0-%d", ncpus-1))
+	if err != nil {
+		return err
+	}
+
+	machineCPUsMask := cpuNumToCPUMask(ncpus)
```
MarSik, line 332: Why a new method? Can't you get the mask form machineCPUs? It represents the same data, right?

internal/runtimehandlerhooks/utils.go
```diff
@@ -121,3 +124,82 @@ func UpdateIRQSmpAffinityMask(cpus, current string, set bool) (cpuMask, bannedCP
 	}
 	return maskStringWithComma, invertedMaskStringWithComma, nil
 }
+
+// cpuMaskToCPUSet parses a CPUSet received in a Mask Format, see:
+// https://man7.org/linux/man-pages/man7/cpuset.7.html#FORMATS
+func cpuMaskToCPUSet(cpuMask string) (cpuset.CPUSet, error) {
```
MarSik, line 130: Same here, isn't cpuset providing this already?

internal/runtimehandlerhooks/utils.go
```diff
@@ -121,3 +124,82 @@ func UpdateIRQSmpAffinityMask(cpus, current string, set bool) (cpuMask, bannedCP
...
+	n := len(slice)
+	for i := 0; i < n/2; i++ {
+		slice[i], slice[n-i-1] = slice[n-i-1], slice[i]
+	}
+
+	return string(slice)
+}
+
+// machineCPUsCount returns the number of processing units available
+func machineCPUsCount() (int, error) {
```
MarSik, line 188: This is only used to create a long enough bitmask (allocating space and filling it with 1s). Executing an external tool for that seems to be a bit excessive. Is there a better / cheaper way? Reading /proc or /sys or maybe just using a big enough number (even though I would prefer something dynamic)?

internal/runtimehandlerhooks/high_performance_hooks.go
marcel-apf: We actually don't have access to this info, but indeed reading  /sys/devices/system/cpu/online will give us the info as a set already, and also we will not need to use the nproc tool. I'll make the changes, thanks.

internal/runtimehandlerhooks/utils.go
marcel-apf: We need the set, e.g 0-7, as a mask  "ff"

internal/runtimehandlerhooks/utils.go
marcel-apf: Indeed, as stated above reading /sys/devices/system/cpu/online will make this function obsolete

marcel-apf:
Addressed @fromanirh and @MarSik reviews:
 - cosmetic changes
 - read /sys/devices/system/cpu/online instead of executing nproc --all

internal/runtimehandlerhooks/utils_test.go
```diff
@@ -4,10 +4,11 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 )
 
 var _ = Describe("Utils", func() {
-	Describe("UpdateIRQSmpAffinityMask", func() {
+	Describe("filterCPUs", func() {
```
haircommander, line 11: this function name changed, should probably change the test description that goes along with it

marcel-apf: totally right, missed it.

marcel-apf: done, thanks

marcel-apf:
Addressed @haircommander's review:
- Fixed unit-test name

umohnani8:
@saschagrunert @haircommander @mrunalp PTAL

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -289,3 +310,63 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
 	return nil
 }
+
+func setRPSLoadBalancing(c *oci.Container, sysDevicesDir, onlineCPUsSysFile string, enable bool) error {
+	lspec := c.Spec().Linux
+	if lspec == nil ||
+		lspec.Resources == nil ||
+		lspec.Resources.CPU == nil ||
+		lspec.Resources.CPU.Cpus == "" {
+		return errors.Errorf("find container %s CPUs", c.ID())
```
umohnani8, line 320: Should the error message be `cannot find...`?

marcel-apf: I paraphrased https://github.com/cri-o/cri-o/blob/ae6220f779afb6c936d95816540021258ef5a5b2/internal/runtimehandlerhooks/high_performance_hooks.go#L151.
You are right and I will make the change.

marcel-apf:
Addressed @umohnani8's review:

- Improved an error message

umohnani8:
LGTM
@kolyshkin @fidencio PTAL

/retest

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -289,3 +310,63 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+	if err != nil {
+		return err
+	}
+
+	machineCPUsMask, _, err := computeCPUmask(cpus, "", true)
+	if err != nil {
+		return err
+	}
+
+	return filepath.Walk(sysDevicesDir, func(path string, info os.FileInfo, err error) error {
```
fromanirh, line 339: can we narrow down this path a bit further? for example: can we scan only the `/sys/class/net` tree? not really sure this is the best subtree, just an example.

marcel-apf: Sadly we can't. These files ar e only under /sys/device and under different sub-directories.
However only the net devices have the rps_cpus file, so we should be fine.
`
/sys/devices/pci0000:00/0000:00:03.0/virtio0/net/ens3/queues/rx-0/rps_cpus
/sys/devices/virtual/net/vxlan_sys_4789/queues/rx-0/rps_cpus
/sys/devices/virtual/net/veth2fdfb5e3/queues/rx-0/rps_cpus
/sys/devices/virtual/net/tun0/queues/rx-0/rps_cpus
/sys/devices/virtual/net/vetha8646254/queues/rx-0/rps_cpus
/sys/devices/virtual/net/vethf2fa53ed/queues/rx-0/rps_cpus
/sys/devices/virtual/net/br0/queues/rx-0/rps_cpus
/sys/devices/virtual/net/ovs-system/queues/rx-0/rps_cpus
/sys/devices/virtual/net/veth67ff0295/queues/rx-0/rps_cpus
/sys/devices/virtual/net/veth6afda312/queues/rx-0/rps_cpus
/sys/devices/virtual/net/veth84c456c3/queues/rx-0/rps_cpus
/sys/devices/virtual/net/lo/queues/rx-0/rps_cpus
/sys/devices/virtual/net/vethcd6422cb/queues/rx-0/rps_cpus
/sys/devices/virtual/net/veth321c2a31/queues/rx-0/rps_cpus
/sys/devices/virtual/net/veth55d926ae/queues/rx-0/rps_cpus
`

marcel-apf:
/hold
Setting the rps in the host might not be enough, we may need to set it also for the existing namespaces. I will double-check.

marcel-apf:
The approach is not enough to eliminate the latency spikes entirely.

PR #4386, opened 2020-11-20
Make NamespaceOption an internal type in sandbox

saschagrunert:
#### What type of PR is this?

/kind cleanup

The fact that we re-use the protobuf devotions in the lower level
implementation details of CRI-O increases the burden to support multiple
Container Runtime Interface versions. This patch is one of a series of
patches to remove the protobuf type usage from the internal lib. This
allows us to create a middle layer later on to support two CRI versions
in parallel.

Required for https://github.com/cri-o/cri-o/pull/4372
We have to do that with every `pb.Type` we're using in CRI-O. This seems the only way to me to cleanly re-use the implementations without having a need for copy-pasting code around.

```release-note
None
```


TomSweeneyRedHat:
LGTM
assuming happy tests

mrunalp (approved):

PR #4391, opened 2020-11-25
simplify adding annotations to spec

wgahnagl:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci

/kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

refactors sandbox_run_linux to simplify adding annotations to the spec

None 

```release-note
None
```

### DIFF
--- a/pkg/sandbox/sandbox.go
+++ b/pkg/sandbox/sandbox.go
@@ -3,12 +3,23 @@ package sandbox
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 
+	"github.com/containers/libpod/v2/pkg/annotations"
 	"github.com/containers/storage/pkg/stringid"
+	"github.com/cri-o/cri-o/internal/config/node"
+	"github.com/cri-o/cri-o/internal/hostport"
+	"github.com/cri-o/cri-o/internal/lib"
+	ann "github.com/cri-o/cri-o/pkg/annotations"
 	"github.com/cri-o/cri-o/pkg/container"
 	"github.com/cri-o/cri-o/server/cri/types"
+	json "github.com/json-iterator/go"
+	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/kubernetes/pkg/kubelet/leaky"
+	kubeletTypes "k8s.io/kubernetes/pkg/kubelet/types"
 )
 
 // Sandbox is the interface for managing pod sandboxes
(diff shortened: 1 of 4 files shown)


test/testdata/config.json
```diff
@@ -0,0 +1,78 @@
+{
```
haircommander, line 1: did you mean to add this?

pkg/sandbox/sandbox.go
```diff
@@ -78,6 +101,110 @@ func (s *sandbox) SetConfig(config *types.PodSandboxConfig) error {
...
+	s.spec.AddAnnotation(annotations.ShmPath, shmPath)
+	s.spec.AddAnnotation(annotations.PrivilegedRuntime, fmt.Sprintf("%v", privileged))
+	s.spec.AddAnnotation(annotations.RuntimeHandler, runtimeHandler)
+	s.spec.AddAnnotation(annotations.ResolvPath, resolvPath)
+	s.spec.AddAnnotation(annotations.HostName, hostname)
+	s.spec.AddAnnotation(annotations.NamespaceOptions, string(nsOptsJSON))
+	s.spec.AddAnnotation(annotations.KubeName, kubeName)
+	s.spec.AddAnnotation(annotations.HostNetwork, fmt.Sprintf("%v", hostNetwork))
+	s.spec.AddAnnotation(annotations.ContainerManager, lib.ContainerManagerCRIO)
+	s.spec.AddAnnotation(annotations.Created, created)
```
haircommander, line 167: is created used in runPodSandbox at all? if now, we can just call time.Now() here and drop the arg

haircommander:
I think this update is a great start, and the correct direction. I think it should be a goal to drop more of the args here rather than passing them into Annotations(), and also we should share more with `pkg/container/container`'s version of this function.

A couple of nits to start you off.. I'll do another comb through but it looks good so far

pkg/sandbox/sandbox.go
wgahnagl: I tried doing that but it was causing errors because there was a discrepancy between when the pod was "created" and when it set the annotation

kolyshkin:
> I think it should be a goal to drop more of the args here rather than passing them into Annotations()

I agree. Anything more than 3-5 string parameters already looks scary, and in here we have more than 10. This either calls for a structure, or a set of With-style methods modifying the object (as much as I don't like With-style stuff I think it might be a good idea to use it here).

haircommander:
> > I think it should be a goal to drop more of the args here rather than passing them into Annotations()
> 
> I agree. Anything more than 3-5 string parameters already looks scary, and in here we have more than 10. This either calls for a structure, or a set of With-style methods modifying the object (as much as I don't like With-style stuff I think it might be a good idea to use it here).

considering we have struct fields already, I think the majority of these fields can be saved in the struct. looking quickly, I think
shmPath, privileged, runtimeHandler, resolvPath, hostname, stopSignal, cgroupParent, mountPoint, hostnamePath, cniResult (not the json), created (in time form), isSystemd and spoofedContainer can all be saved in the structure themselves. They're fields that will prove useful in other attempts to move code from server to pkg/sandbox

kolyshkin:
I think we can close this one.

Issue #4221, opened 2020-09-24
CRI-O should be smarter about context timeouts


haircommander:
**Description**
On nodes under load, we run into issues where ContainerCreate and RunPodSandbox requests time out. The most common cases are i/o load, network creation taking a long time, SELinux relabelling of volumes taking a while

Right now, ContainerCreate and RunPodSandbox run to completion, but check at the end that we timed out. if so, we remove the resource we recently created, and try again. We do this because the kubelet works as follows:

Kubelet makes a request
CRI-O handles the request, reserving the resource name as asked by the kubelet
kubelet request times out, and it begins polling CRI-O, continuously trying to create the resource it first tries.
*note, it does not bump the attempt number, as this would cause CRI-O to simultaneously try to create multiples of the same container, further adding load to the already overloaded node*
Eventually, CRI-O learns the kubelet timed out, and cleans up after itself
*note: we clean up here so the next request will pass. if we didn't, the kubelet would say "give me the same container" and CRI-O would say "I already created that container!", even though the kubelet's original request failed*
Kubelet finally asks for the same resource again, and one day, assuming the node's load reduces, CRI-O succeeds in creating the container and returns it to the kubelet.

Notice, the kubelet is asking for the same resource the whole time (name, attempt, sandbox, etc), and yet, CRI-O has the possibility of creating it infinite number of times (if it keeps timing out on the request that actually is handling the creation of the resource), while returning errors while it's trying to create that resource ("name is reserved" errors)

we should do better

Instead of removing the resource upon finding out we timed out, CRI-O should put that newly created resource into a cache, report an error to the kubelet. When the kubelet eventually re-requests that resource (as it would have been doing upon the timeout), we should pull that resource from the cache, and return it, without creating it a second time.

There are a couple of gotchas we need to handle:
*What if the kubelet bumps the attempt number?*
If the kubelet asks to create attempt 0, and then the api-server tells it to create a new version, CRI-O should abandon attempt 0. 
*What if the kubelet gives up on that container?*
CRI-O should abandon the old container, as it's no longer needed.
*What if CRI-O is restarted before the container is correctly returned?*
CRI-O should be prepared to clean up that container if the kubelet doesn't ask for it.

So here's the proposed solution:
[...]

PR #4266, opened 2020-10-14
WIP: improve timeout handling

haircommander:
#### What type of PR is this?

/kind bug

> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

#### What this PR does / why we need it:
This PR changes the behavior of container and pod creation to save the progress of a creation request if we timeout, and quickly return that progress if it's re-requested.

Specifically, it adds
- ResourceCache package, which keeps track of pods and containers between a timed out request and a successful one
- updates server to save the progress of a pod or container request
- add unit tests and integration tests

#### Which issue(s) this PR fixes:

fixes https://github.com/cri-o/cri-o/issues/4221

#### Does this PR introduce a user-facing change?

```release-note
Fix a bug where a timeout in RunPodSandbox or CreateContainer requests caused CRI-O to delete the newly created resource. Now, it saves that resource, until the kubelet re-requests it, thus allowing kubelet and CRI-O to reconcile quicker when nodes are under load
```

PR #4241, opened 2020-10-02
server: refactor handling of cleanup funcs

haircommander:
#### What type of PR is this?

/kind cleanup

> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

#### What this PR does / why we need it:

it doesn't make very much sense to have so many deferred funcs queued, and check retErr each time
instead, we can check retErr once, and loop through a slice of cleanupFuncs

this also carries https://github.com/cri-o/cri-o/pull/4240 to avoid a merge conflict
#### Does this PR introduce a user-facing change?

```release-note
none
```

PR #4394, opened 2020-11-30
Improve timeout handling

haircommander:
#### What type of PR is this?

/kind bug

> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR changes the behavior of container and pod creation to save the progress of a creation request if we timeout, and quickly return that progress if it's re-requested.

Specifically, it adds
- ResourceCache package, which keeps track of pods and containers between a timed out request and a successful one
- updates server to save the progress of a pod or container request
- add unit tests and integration tests
- have each new request watch the resource and wait until it's created. This is important because currently this situation causes cri-o to spam the kubelet with "name is reserved" errors. Waiting until the resource is created (or for a timeout) will significantly reduce the number of these errors, making a standard error seem less catastrophic

This is carrying https://github.com/cri-o/cri-o/pull/4266, plus adding the watcher idiom.

fixes https://github.com/cri-o/cri-o/issues/4221

```release-note
Fix a bug where a timeout in RunPodSandbox or CreateContainer requests caused CRI-O to delete the newly created resource. Now, it saves that resource, until the kubelet re-requests it, thus allowing kubelet and CRI-O to reconcile quicker when nodes are under load.
```


haircommander:
A week ago I was able to verify this fix worked.
using https://github.com/RobertKrawitz/OpenShift4-tools/blob/master/clusterbuster and an openshift cluster with and without these patches lead to these two reports (using https://github.com/RobertKrawitz/OpenShift4-tools/blob/master/monitor-pod-status to generate the report):

`./clusterbuster -P server -b 5 -c 1  -p 10 -r 4 -N 75 -d 2 --parallel-deployments=2  --bytes=750000`

without this patch:
```
15:17:36 Pods: 0P 0R 0C 0O 0T 0E 0X 0I 0U  / 0   NS: 0A 0T  / 0   Sec: 0A 0S    Node: 3R 0N  / 3 
15:18:36 Pods: 0P 25R 0C 55O 0T 0E 0X 0I 0U  / 80   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:19:36 Pods: 62P 53R 0C 231O 0T 0E 0X 0I 0U  / 346   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:20:36 Pods: 98P 97R 0C 446O 0T 0E 0X 0I 0U  / 641   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:21:36 Pods: 80P 167R 0C 503O 0T 0E 0X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:22:36 Pods: 54P 258R 0C 438O 0T 0E 0X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:23:40 Pods: 54P 413R 0C 264O 0T 0E 19X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:24:36 Pods: 54P 420R 0C 232O 0T 0E 44X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 2R 1N  / 3 
15:25:36 Pods: 54P 432R 0C 181O 0T 0E 83X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 2R 1N  / 3 
15:26:36 Pods: 54P 432R 0C 181O 0T 0E 83X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 1R 2N  / 3 
15:27:39 Pods: 54P 432R 0C 181O 0T 0E 83X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 1R 2N  / 3 
15:28:39 Pods: 54P 432R 0C 181O 0T 0E 83X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 1R 2N  / 3 
```

with this patch:
```
15:56:03 Pods: 0P 0R 0C 0O 0T 0E 0X 0I 0U  / 0   NS: 0A 0T  / 0   Sec: 0A 0S    Node: 3R 0N  / 3 
15:57:03 Pods: 11P 95R 0C 91O 0T 0E 0X 0I 0U  / 197   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:58:05 Pods: 48P 221R 0C 196O 0T 0E 0X 0I 0U  / 465   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
15:59:05 Pods: 132P 286R 0C 288O 0T 0E 0X 0I 0U  / 706   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
16:00:03 Pods: 125P 358R 0C 267O 0T 0E 0X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
16:01:06 Pods: 97P 482R 0C 171O 0T 0E 0X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 3R 0N  / 3 
16:02:04 Pods: 96P 558R 0C 96O 0T 0E 0X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 2R 1N  / 3 
16:03:04 Pods: 96P 558R 0C 96O 0T 0E 0X 0I 0U  / 750   NS: 75A 0T  / 75   Sec: 0A 675S    Node: 1R 2N  / 3
```

Note: both clusters did eventually fail (the test case specifically had to overload it to trigger this situation) but with the patches we were able to get more pods running (`558R` vs `432R`). Also note in the "with patch" case, we have `0X` which means 0 container create errors vs `83X` or 83 containers that were spamming kubelet logs with "name is reserved" errors

saschagrunert (approved):
LGTM

server/sandbox_network.go
```diff
@@ -21,7 +21,7 @@ import (
 func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs []string, result cnitypes.Result, retErr error) {
 	overallStart := time.Now()
 	// give a network Start call 2 minutes, half of a RunPodSandbox request timeout limit
-	startCtx, startCancel := context.WithTimeout(ctx, 2*time.Minute)
+	startCtx, startCancel := context.WithTimeout(ctx, 4*time.Minute)
```
fidencio, line 24: Why? :-)

Mind to add an explanation in the commit message and also update the comment above (which still mentiones 2 minutes).

fidencio: Looking at:
```
resourcecache: add watcher idiom

Even if we use the resource cache as is, the user is still bombarded with messages saying the name is reserved.

This is bad UX, and we're capable of improving it.

Add watcher idiom to resource cache, allowing a handler routine of RunPodSandbox or CreateContainer to
wait for a resource to be available.

Something that is key here is if the resource becomes available while we're watching for it,
*we still need to error on this request*
This is because we could get the resource from the cache, remove it (thus meaning it won't be cleaned up),
and the kubelet's request could time out, and it could try again. This would cause us to leak a resource.

This way, if we get into this situation, there needs to be three requests:
first that times out
second that discovers the resource is ready, but still errors
third that actually retrives that resource and returns it.

This will result in many fewer "name is reserved" errors (one every 2 seconds to one every 4 minutes)
```

Is this the reason of updating it?

internal/resourcecache/resourcecache.go
```diff
@@ -58,14 +62,25 @@ func (rc *ResourceCache) AddResourceToCacheWithTimeout(name string, resource Ide
...
+	if !ok {
+		r = &Resource{}
+		rc.resources[name] = r
 	}
-	rc.resources[name] = &Resource{
-		resource:     resource,
-		cleanupFuncs: cleanupFuncs,
+	// make sure the resource hasn't already been added to the cache
+	if r.resource != nil || r.cleanupFuncs != nil {
+		return errors.Errorf("failed to add entry %s to ResourceCache; entry already exists", name)
```
fidencio, line 73: Shall we really error out if trying to add an entry that already exists to the Cache?
To me it's something we should simply log and sucessfully return.

fidencio (commented):
@haircommander,  a few basic comments but looks good in general.

internal/resourcecache/resourcecache.go
haircommander: it means that somehow two different goroutines are trying to create the same resource at the same time, and managed to race against each other perfectly. It's a really unlikely scenerio (especially because the locking is perfect :wink: ), but it is definitely an error situation. The most negative thing that would happen is the cleanup funcs for each will be different, so if one overwrote the other, there would be leaked resources

server/sandbox_network.go
haircommander: updated comment and commit, thanks!

haircommander:
blocked on https://github.com/cri-o/cri-o/pull/4241 btw

fidencio:
LGTM

haircommander:
this is ready for review

server/sandbox_network.go
```diff
@@ -20,8 +20,11 @@ import (
 // or an error
 func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs []string, result cnitypes.Result, retErr error) {
 	overallStart := time.Now()
-	// give a network Start call 2 minutes, half of a RunPodSandbox request timeout limit
-	startCtx, startCancel := context.WithTimeout(ctx, 2*time.Minute)
+	// Give a network Start call the entire RunPodSandbox request timeout limit.
+	// This is to prevent the CNI plugin from taking an unbounded amount of time,
+	// but to still allow a long-running sandbox creation to be cached and reused,
+	// rather than failing and recreating it.
+	startCtx, startCancel := context.WithTimeout(ctx, 4*time.Minute)
```
mrunalp, line 27: why not pass down the main context here if we are using the full timeout?

haircommander: yeah that's a good idea. I am now  actually considering creating a custom timeout independent of the outer context (like: 
`startCtx, startCancel := context.WithTimeout(context.Background(), 6*time.Minute)` ) which would really allow us to work independently of the kubelet's timeout

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var sleepTimeBeforeCleanup = 2 * time.Minute
```
mrunalp, line 11: Do want to make this configurable?

server/container_create.go
```diff
@@ -488,6 +488,11 @@ func (s *Server) CreateContainer(ctx context.Context, req *pb.CreateContainerReq
 	}()
 
 	if _, err = s.ReserveContainerName(ctr.ID(), ctr.Name()); err != nil {
+		if cachedID := s.ResourceFromCache(ctr.Name()); cachedID != "" {
```
mrunalp: We can guard this block against a specific typed error.

server/sandbox_run_linux.go
```diff
@@ -319,6 +319,12 @@ func (s *Server) runPodSandbox(ctx context.Context, req *pb.RunPodSandboxRequest
 	}()
 
 	if _, err := s.ReservePodName(sbox.ID(), sbox.Name()); err != nil {
+		if cachedID := s.ResourceFromCache(sbox.Name()); cachedID != "" {
```
mrunalp: Same: We can guard this block using a specific typed error.

server/utils.go
```diff
@@ -295,3 +296,7 @@ func getSourceMount(source string, mountinfos []*mount.Info) (path, optional str
 
 	return res.Mountpoint, res.Optional, nil
 }
+
+func isContextError(err error) bool {
+	return err == context.Canceled || err == context.DeadlineExceeded
```
mrunalp, line 303: Have we identified cases in which we want to retain a resource when the context is Canceled from the kubelet?

server/sandbox_network.go
```diff
@@ -20,8 +20,11 @@ import (
 // or an error
 func (s *Server) networkStart(ctx context.Context, sb *sandbox.Sandbox) (podIPs []string, result cnitypes.Result, retErr error) {
 	overallStart := time.Now()
-	// give a network Start call 2 minutes, half of a RunPodSandbox request timeout limit
-	startCtx, startCancel := context.WithTimeout(ctx, 2*time.Minute)
+	// Give a network Start call a full 5 minutes, independent of the context of the request.
+	// This is to prevent the CNI plugin from taking an unbounded amount of time,
+	// but to still allow a long-running sandbox creation to be cached and reused,
+	// rather than failing and recreating it.
+	startCtx, startCancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
mrunalp, line 27: @danwinship Thoughts on bumping up this timeout?

server/utils.go
```diff
@@ -300,3 +302,26 @@ func getSourceMount(source string, mountinfos []*mount.Info) (path, optional str
...
+	watcher := s.WatcherForResource(name)
+	if watcher == nil {
+		return "", errors.Errorf("error attempting to watch for %s %s: no longer found", resourceType, name)
+	}
+	log.Infof(ctx, "Could not find %s %s in cache; creation not yet finished", resourceType, name)
+	var err error
+	select {
+	case <-ctx.Done():
+		err = ctx.Err()
+	case <-time.After(time.Minute * 4):
```
mrunalp, line 320: Let's use constants for these timeouts.

server/utils.go
```diff
@@ -300,3 +302,26 @@ func getSourceMount(source string, mountinfos []*mount.Info) (path, optional str
...
+func (s *Server) getResourceOrWait(ctx context.Context, name, resourceType string) (string, error) {
+	if cachedID := s.ResourceFromCache(name); cachedID != "" {
+		log.Infof(ctx, "Found %s %s with ID %s in resource cache; using it", resourceType, name, cachedID)
+		return cachedID, nil
+	}
+	watcher := s.WatcherForResource(name)
+	if watcher == nil {
+		return "", errors.Errorf("error attempting to watch for %s %s: no longer found", resourceType, name)
+	}
+	log.Infof(ctx, "Could not find %s %s in cache; creation not yet finished", resourceType, name)
```
mrunalp, line 315: Modify the message to indicate that we are waiting for the creation to finish up to timeout value.

internal/resourcecache/resourcecache.go
haircommander: :shrug: doesn't seem like a tunable we *need* and our config is already pretty big. I'm not that inclined to add it

server/container_create.go
haircommander: what do you mean by that?

server/utils.go
haircommander: I figured we'd keep the behavior as we had before--checking both Canceled and DeadlineExceeded

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ResourceCache is a structure that saves information about a nearly created resource.
```
kolyshkin, line 13: `s/nearly/recently/`?

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+// Resources can be added and retrieved from the cache. A retrieval (ResourceFromCache) also removes the item from the cache.
+// If `sleepTimeBeforeCleanup` has passed, and the resource has not yet been retrieved, ResourceCache considers that resource stale,
+// and removes it.
```
kolyshkin, line 16: By this description it does not look like a cache (as cache hit does not removes the object). Maybe `ResourceStore`?

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return rc
+}
+
+// AddResourceToCache takes a unique resource name (retrieved from the client request, not generated by the server)
+// a newly created resource, and functions to cleanup that newly created resource.
+// It adds the Resource to the ResourceCache, as well as starts a go routine that is responsible for cleaning up the
+// resource if the server has not gotten another request for it (i.e. if ResourceFromCache has not been called).
+// AddResourceToCache expects `name`s passed to it to be unique. If a duplicate `name` is detected, it returns an error.
+func (rc *ResourceCache) AddResourceToCache(name string, resource IdentifiableCreatable, cleanupFuncs []func()) error {
```
kolyshkin, line 55: `s/AddResourceToCache/Add/`?

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+	return rc.AddResourceToCacheWithTimeout(name, resource, cleanupFuncs, sleepTimeBeforeCleanup)
+}
+
+// AddResourceToCacheWithTimeout is used for testing purposes. It allows the caller to set the timeout, allowing for faster tests.
+// Most callers should use AddResourceToCache instead.
+func (rc *ResourceCache) AddResourceToCacheWithTimeout(name string, resource IdentifiableCreatable, cleanupFuncs []func(), timeout time.Duration) error {
```
kolyshkin, line 61: `s/AddResourceToCacheWithTimeout/AddWithTimeout/`?

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+	defer rc.Unlock()
+
+	if _, ok := rc.resources[name]; ok {
+		return errors.Errorf("failed to add entry %s to ResourceCache; entry already exists", name)
+	}
+	rc.resources[name] = &Resource{
+		resource:     resource,
+		cleanupFuncs: cleanupFuncs,
+	}
+	go rc.cleanupResourceAfterSleep(name, timeout)
```
kolyshkin, line 78: Depending on number of resources you want to store, a goroutine per resource might not be a very scalable model.

The alternative is a single goroutine which wakes up say once a minute (or 10 minutes) and cleans up all stale entries. If you need better precision, you can check if the object is stale upon retrieve.

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+	for _, f := range r.cleanupFuncs {
+		f()
+	}
+}
+
+// ResourceFromCache attempts to look up a resource by its name.
+// If it's found, it's removed from the cache, and it is set as created.
+// ResourceFromCache returns an empty ID if the resource is not found,
+// and returns the value of the items ID() method if it is.
+func (rc *ResourceCache) ResourceFromCache(name string) string {
```
kolyshkin, line 133: `s/ResourceFromCache/Get/`?

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+func (rc *ResourceCache) cleanupResourceAfterSleep(name string, timeout time.Duration) {
+	time.Sleep(timeout)
+
+	rc.Lock()
+	r, ok := rc.resources[name]
+	if !ok {
+		rc.Unlock()
+		return
+	}
+	logrus.Infof("cleaning up %s after kubelet didn't re-request it", name)
```
kolyshkin, line 99: Maybe `cleaning up stale resource %q`?

internal/resourcecache/resourcecache.go
```diff
@@ -0,0 +1,109 @@
...
+	delete(rc.resources, name)
+	// no need to hold the lock when running the cleanup functions
+	rc.Unlock()
+
+	for _, f := range r.cleanupFuncs {
+		f()
+	}
```
kolyshkin, line 106: If the resource is removed and then instantly re-added, this code might call its cleanup functions while it's being re-added, i.e. at least theoretically there is a race here.

internal/resourcecache/resourcecache.go
kolyshkin: On a second thought, maybe it's better to have `Put` (instead of `AddResourceToCache`) and `Get` (instead of `ResourceFromCache`), and also `Got` (instead of `SetCreated`).

server/server.go
```diff
@@ -66,6 +67,8 @@ type Server struct {
 	pullOperationsInProgress map[pullArguments]*pullOperation
 	// pullOperationsLock is used to synchronize pull operations.
 	pullOperationsLock sync.Mutex
+
+	*resourcecache.ResourceCache
```
kolyshkin, line 71: With the naming changes I suggested above, surely it needs to have this one as a named member instead of embedding...

internal/resourcecache/resourcecache.go
kolyshkin: `s/var/const/`?

haircommander: good call

internal/resourcecache/resourcecache.go
haircommander: `SetCreated` is a function of Container and Sandbox that marks it as created, so I don't as much want to change that, though ACK on the other changes

internal/resourcecache/resourcecache.go
haircommander: *maybe* for users that aren't using pod/container name as a way to determine whether there's a duplicated request going on. the last cleanupFunc to be called is `Release{Pod,Container}Name` so, at worse, a goroutine adds itself as a watcher for a resource that's already been cleaned up, but that will time out in 4 minutes. Unless I misunderstand the flow you're imagining

server/utils.go
```diff
@@ -300,3 +302,26 @@ func getSourceMount(source string, mountinfos []*mount.Info) (path, optional str
...
+	var err error
+	select {
+	case <-ctx.Done():
+		err = ctx.Err()
+	case <-time.After(time.Minute * 4):
+		err = errors.Errorf("waited too long for request to timeout or %s %s to be created", resourceType, name)
+	case <-watcher:
```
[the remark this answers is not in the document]
haircommander: done!

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
+	helptext=$(crictl runp -h)
+	if [[ "$helptext" != *"cancel-timeout"* ]]; then
```
kolyshkin, line 7:
```bash
if ! crictl runp -h | grep -q cancel-timeout; then
...
```

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+}
+
+function teardown() {
+	cleanup_test
+}
+
+function create_conmon() {
+	timeout=$1
+	cat > "$TESTDIR"/tmp_conmon << EOF
+#!/bin/sh
```
kolyshkin, line 20: since you use `[[` below, which is a bashism, you need to have `/bin/bash` here.

internal/resourcecache/resourcecache.go
haircommander: this function has been dropped now that it's one goroutine

internal/resourcecache/resourcecache.go
haircommander: fair enough, I've moved to one goroutine. I can't imagine there would be that many, but why not future proof it.

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+	# cri-o should not report any pods
+	pods=$(crictl pods -q)
+	[[ -z "$pods" ]]
+
+	created_ctr_id=$("$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q)
+	[[ ! -z "$created_ctr_id" ]]
+
+	run crictl runp "$TESTDATA"/sandbox_config.json
+	echo "$output"
+	[ "$status" -eq 0 ]
```
kolyshkin, line 50:
```suggestion
	output=$(crictl runp "$TESTDATA"/sandbox_config.json)
```

In this case
 - stderr (in which we might be interested it if the test is failing) is not redirected (and will be shown by bats if test fails);
 - stdout (which we're not really interested in except for the below check) goes to `$output`;
 - exit code check is implicit.

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+
+	# cri-o should not report any pods
+	ctrs=$(crictl ps -aq)
+	[[ -z "$ctrs" ]]
+
+	# cri-o should have created a container
+	created_ctr_id=$("$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q | grep -v $pod_id)
+	[[ ! -z "$created_ctr_id" ]]
+
+	run crictl create $pod_id "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json
```
kolyshkin, line 75: ditto

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+	[ "$status" -ne 0 ]
+
+	# allow cri-o to catch up
+	sleep 4s
+
+	created_ctr_id=$("$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q)
+	[[ ! -z "$created_ctr_id" ]]
+
+	# we should create a new pod and not reuse the old one
+	run crictl runp <(cat "$TESTDATA"/sandbox_config.json | jq '.metadata.attempt = 2')
```
kolyshkin, line 96: `cat` and the pipe are redunant here, `jq '.metadata.attempt = 2' "$TESTDATA"/sandbox_config.json` would do

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+	[ "$status" -eq 0 ]
+
```
kolyshkin, line 98: extra newline

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+	[[ "$output" != *"$created_ctr_id"* ]]
+
+	sleep 150s
+
+	# the old, timed out container should have been removed
+	run "$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q
```
kolyshkin, line 104: `output=$("$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q)`

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+
+	# allow cri-o to catch up
+	sleep 4s
+
+	# cri-o should have created a container
+	created_ctr_id=$("$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q | grep -v $pod_id)
+	[[ ! -z "$created_ctr_id" ]]
+
+	# should create a new container and not reuse the old one
+	run crictl create $pod_id <(cat "$TESTDATA"/container_config.json | jq '.metadata.attempt = 2') "$TESTDATA"/sandbox_config.json
```
kolyshkin, line 126: `output=$(run crictl ...`

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+	# allow cri-o to catch up and clear its state of the pod
+	sleep 4s
+
+	# container should not have been cleaned up
+	created_ctr_id=$("$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q)
+	[[ ! -z "$created_ctr_id" ]]
+
+	run crictl create $created_ctr_id "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json
+	echo "$output"
+	[ "$status" -ne 0 ]
```
kolyshkin, line 209:
```suggestion
! crictl create $created_ctr_id "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json
```

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+	echo "$output"
+	[[ "$output" == *"context deadline exceeded"* ]]
+	[ "$status" -ne 0 ]
+
+	# allow cri-o to catch up
+	sleep 4s
+
+	# cri-o should have created a container
+	created_ctr_id=$("$CONTAINER_RUNTIME" --root "$RUNTIME_ROOT" list -q | grep -v $pod_id)
+	[[ ! -z "$created_ctr_id" ]]
```
kolyshkin, line 235:
```suggestion
[ -n "$created_ctr_id" ]
```

test/timeout.bats
```diff
@@ -0,0 +1,252 @@
...
+
+	run crictl start $created_ctr_id
+	echo "$output"
+	[ "$status" -ne 0 ]
```
kolyshkin, line 239:
```suggestion
	! crictl start $created_ctr_id
```

and similar in the below cases

server/utils.go
```diff
@@ -295,3 +298,41 @@ func getSourceMount(source string, mountinfos []*mount.Info) (path, optional str
 
 	return res.Mountpoint, res.Optional, nil
 }
+
+func isContextError(err error) bool {
+	return err == context.Canceled || err == context.DeadlineExceeded
+}
+
+func (s *Server) getResourceOrWait(ctx context.Context, name, resourceType string) (string, error) {
+	const timeToWaitBeforeGivingUp = time.Minute * 4
```
mrunalp, line 307: How about `resourceCreationWaitTime` ?

haircommander: fixed!

TomSweeneyRedHat:
@haircommander looks like the cgroupv2 tests are kicking up their heels.

haircommander:
there was a flake with the tests, I tweaked it a bit, I think this should be better

internal/resourcestore/resourcestore.go
```diff
@@ -0,0 +1,127 @@
...
+		return ""
+	}
+	delete(rc.resources, name)
+	r.resource.SetCreated()
+	return r.resource.ID()
+}
+
+// Put takes a unique resource name (retrieved from the client request, not generated by the server)
+// a newly created resource, and functions to cleanup that newly created resource.
+// It adds the Resource to the ResourceStore, as well as starts a go routine that is responsible for cleaning up the
```
kolyshkin, line 111: the part about go routine is obsoleted

PR #4396, opened 2020-12-01
Fix make vendor GOSUMDB

saschagrunert:
#### What type of PR is this?

/kind ci

This fixes the intentionally unset `GOSUMDB` variable when running `make
vendor`.

None
None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -294,7 +294,7 @@ ${SHELLCHECK}:
 	curl -sSfL $$URL | tar xfJ - -C ${BUILD_BIN_PATH} --strip 1 shellcheck-$$VERSION/shellcheck && \
 	sha256sum ${SHELLCHECK} | grep -q $$SHA256SUM
 
-vendor: export GOSUMDB := ""
+vendor: export GOSUMDB :=
 vendor:
 	$(GO) mod tidy
 	$(GO) mod vendor


Makefile
```diff
@@ -294,7 +294,7 @@ ${SHELLCHECK}:
 	curl -sSfL $$URL | tar xfJ - -C ${BUILD_BIN_PATH} --strip 1 shellcheck-$$VERSION/shellcheck && \
 	sha256sum ${SHELLCHECK} | grep -q $$SHA256SUM
 
-vendor: export GOSUMDB := ""
+vendor: export GOSUMDB=
```
kolyshkin, line 297: I guess `:=` needs to be retained, as we want immediate rather than deferred assignment here. IOW, just drop the quotes.

```suggestion
vendor: export GOSUMDB :=
```

kolyshkin: It would probably be nice to explain it in a commit message that this fixes the issue of setting GOSUMDB to a literal pair of quotes (""), and we want the value to be empty.

(if instead you want to unset `GOSUMDB`, use `undefine GOSUMDB`)

saschagrunert: Alright, changed as suggested :+1:

haircommander:
/retest

LGTM

PR #4398, opened 2020-12-01
typo on reference to ~/.docker/config.json vs. /var/lib/kubelet/config.json

piengeng:
1. global-auth-file likely reference docker's default config.json format
2. /var/lib/kubelet/ doesn't have config.json, but config.yaml, which is stark difference.

#### What type of PR is this?

/kind documentation

fix typo in documentation, so that the next person reading the documentation or using the application will understand how to work with this software.

None

please review.

```release-note
None
```


haircommander:
thanks for the PR @piengeng 
/ok-to-test

you'll need to sign-off your commit with `git commit --amend -s` or manually add a line `Signed-off-by: Henry Kwan <*email*>` to the bottom of your commit message

I am not super familiar with the different config formats or the history behind having this directory, but looking into a bit, it does seem you're right that `/var/lib/kubelet/config.json` is not a well established convention, and that we'd more likely reference something similar to `~/.docker/config.json`

PTAL @vrothberg @mtrmac @mrunalp to verify that ^

mtrmac (changes requested):
From a quick search finding https://github.com/openshift/machine-config-operator/search?q=global_auth_file&type=code, NAK.

It could be legitimate to find the actual users, and full history of how this feature came, and argue that the path is wrong — and then actually propose changes to the other references to be consistent, but from “likely reference” I can’t tell what, if any, research has been done.

haircommander:
> From a quick search finding https://github.com/openshift/machine-config-operator/search?q=global_auth_file&type=code, NAK.
> 
> It could be legitimate to find the actual users, and full history of how this feature came, and argue that the path is wrong — and then actually propose changes to the other references to be consistent, but from “likely reference” I can’t tell what, if any, research has been done.

To be fair, I do believe using MCO as evidence of other references is self-referential, as that path probably came from the same knowledge base (contributors of CRI-O). I can't find evidence of other developers using such a path for an auth file.

haircommander:
ah here it is https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry

mtrmac:
> ah here it is https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry

Well, kind of. That one documents what Kubelet reads to get credentials; this option determines what CRI-O reads to get credentials (when Kubelet doesn’t send any?). With OpenShift the two happen to read the same file, but it’s not, AFAIK, a design requirement for the two daemons to read the same file.

As for CRI-O’s global-auth-file option, it’s a perfectly legitimate objection that OpenShift/MCO is not necessarily an authoritative source for what the option _should_ point to — it _is_ up to the CRI-O project to make a decision/recommendation of how it should be used in larger k8s systems.

That one demonstrated user is, though, IMHO, sufficient to demonstrate that the path in documentation is not a “typo” and that more research or argument would be necessary for changing the current documentation which is, looking at the number of places this PR proposed to change, consistent about using the `/var/lib/kubelet/config.json` path.

mtrmac:
(I have no idea why Kubelet and CRI-O both have a fallback credentials file.)

piengeng:
a `git push --force ` breaks this PR. lol
i think globally, probably, people start containerization using docker, when I `docker login` by default it will create ~/.docker/config.json . json format is typical. it also says it's storing authentication information in text format. anyone uses docker knows this or will see this.

when I setup k8s cluster, inside k8s nodes:/var/lib/kubelet/ there's no config.json but config.yaml for me to refer, there's not much for me to search on how should the content of this 'global-auth-file' . as long the format is json and similar to `~/.docker/config.json` then setting global-auth-file is straight-forward.

mtrmac:
(I’m completely the wrong person to make prescriptions or recommendations for vendor-neutral K8s tool standards.)
- A large part of the k8s purpose is automation of cluster management, a manual `docker login` is inconsistent with that
- If the admin _does_ log into the node to run `docker login` for some purpose, I’d argue it makes more sense for that _not_ to influence what CRI-O does, to make the behavior of CRI-O consistent and reproducible.

haircommander:
@piengeng I totally agree that better documentation on CRI-O's end would help with this. Would you mind opening an issue describing the current gap in understanding you may have? Or, even better, would you have interest in submitting a PR to improve the documentation :)

PR #4422, opened 2020-12-09
test/timeout.bats: fix flakes

kolyshkin:
#### What type of PR is this?

/kind ci
/kind flake

1. Increase timeouts in timeout.bats test to fix flakes.
2. Fix some comments and other nitpicks.

None

```release-note
None
```


saschagrunert (approved):
LGTM

test/timeout.bats
```diff
@@ -26,16 +26,16 @@ EOF
 }
 
 @test "should not clean up pod after timeout" {
-	create_conmon 2s
+	create_conmon 10s
```
haircommander, line 29: changing this from 2 to 10 will not fix the cause of the flake. The flake happens because the  rest of the operation takes a variable amount of time. 

Basically, if you look at the failures, crio is stuck on the conmon call. that's because setting up the network and container rootfs took > 8 seconds. Then, conmon was still in the middle of sleeping when we got past the `sleep 10s` below, and then attempted to call `runc list`, which returned nothing (because conmon hadn't finished creating the container)

The best course of action is to bump the sleep below to 20s, but leave these as 2s. I had been trying to minimize the sleeps below to get the tests running faster, but I suppose there's no way to expect it'd take that little time

kolyshkin: So, the reason why I increased the 2s timeouts is not because of conmon, but because of `crictl create -T`. I have seen that 1 second timeout is too small for our CI, and thought that maybe 2s might also too small sometimes, and increased it just in case. But once we increase `crictl create` timeout, conmon sleep needs to be increased.

Reverting 2s -> 10s change, let's see.

haircommander: the `context timeout exceeded` error we see at the top of the output is actually the correct behavior btw. we want the create_conmon timeout to equal the -T given to crictl, so we are guaranteed to timeout.

kolyshkin: Yes, I understand that. Although I have seen `context timeout exceeded` error from `crictl create -T 1s` when it merely tries to connect to cri-o, and increased the timeout to avoid that one.

kolyshkin: Commit updated.

test/timeout.bats
```diff
@@ -23,19 +27,32 @@ fi
...
+# Allow cri-o to catch up. The sleep here should be less than
+# resourcestore.sleepTimeBeforeCleanup but enough for cri-o to
+# finish processing cancelled crictl create/runp.
+function wait_create() {
+	sleep 30s
+}
+
+# Allow cri-o to catch up and clean the state of a pod/container.
+# The sleep here should be > 2 * resourcestore.sleepTimeBeforeCleanup.
+function wait_clear() {
```
haircommander, line 43: maybe wait_clean? we don't use the verb clear anywhere really

kolyshkin: right

kolyshkin: updated

haircommander:
kata failures are unrelated, I'm merging this to unblock 1.20

PR #4430, opened 2020-12-10
[1.20] improve timeout handling and fix flakes

haircommander:
#### What type of PR is this?

/kind bug

> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

carries https://github.com/cri-o/cri-o/pull/4394 and https://github.com/cri-o/cri-o/pull/4422
replaces https://github.com/cri-o/cri-o/pull/4421

```release-note
Fix a bug where a timeout in RunPodSandbox or CreateContainer requests caused CRI-O to delete the newly created resource. Now, it saves that resource, until the kubelet re-requests it, thus allowing kubelet and CRI-O to reconcile quicker when nodes are under load.
```


internal/resourcestore/resourcestore.go
```diff
@@ -0,0 +1,157 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const sleepTimeBeforeCleanup = 1 * time.Minute
+
+// ResourceStore is a structure that saves information about a recently created resource.
+// Resources can be added and retrieved from the store. A retrieval (Get) also removes the Resource from the store.
```
TomSweeneyRedHat, line 14: a retrieval also removes?  That's unusual no?

haircommander: I suppose, though here we need to keep a single copy of the resource, so we're basically using move semantics. do you have a suggestion for a better method name?

mrunalp (approved):

PR #4435, opened 2020-12-15
allow handler allowed annotations to be set by a flag

wgahnagl:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake
> /kind other

allows the handler allowed annotations to be set with a CLI flag and and environment variable

```release-note
allows handler allowed annotations to be set using the CLI and an environment variable
```


haircommander:
/approve

LGTM

I believe a rebase will fix the tests for you

docs/crio.8.md
```diff
@@ -114,6 +115,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--additional-devices**="": Devices to add to the containers  (default: [])
 
+**--allowed-annotations**="": list of experimental annotations that this runtime handler is allowed to process
```
mrunalp, line 118: Is the comment accurate? I thought this PR was for the top level config and not specific to any runtime handler.

mrunalp: We should also describe or document how this intersects with a runtime handler specific annotations.

wgahnagl: revised comment! Let me know what you think!

haircommander:
@hasan4791  can you take this one over?

hasan4791:
> @hasan4791  can you take this one over?

On a bike trip for 2 weeks😁🥳 Will take this one if its still available 👍

haircommander:
> > @hasan4791  can you take this one over?
> 
> On a bike trip for 2 weeks :grin:  :partying_face:  Will take this one if its still available +1

enjoy!

haircommander:
I actually think the approach here should be different. instead of applying allowed annotations as a config flag, I think we should add a `default_workload` field similar to `default_runtime` and allow users to configure the workload with default allowed annotations

hasan4791:
@haircommander Could you provide some info regarding this PR. Would help me to understand the changes here

haircommander:
instead of this approach, I think the config should look something like:
```
[crio.runtime]
default_workload = "xyz"
[crio.runtime.workloads.xyz]
allowed_annotations = "abc"
```

haircommander:
>  Little confusion here, does the value for default_workload is set by user or do we need to provide something? Also whats the need of this new field though the users already has the ability to set the annotations based on workload config?

you make a good point here that I didn't consider. An admin does have the option to do something like 
```
[crio.runtime.workloads.default]
activation_annotation = "io.cri-o.default"
allowed_annotations = [ "io.cri-o.default" ]
```
and any user who specifies that annotation in their pod would get that feature. Since all of the allowed annotations require pod authors to opt-in, an admin can sort of already specify allowed annotations being allowed by everyone.


we may not actually need this after all...

hasan4791:
> > Little confusion here, does the value for default_workload is set by user or do we need to provide something? Also whats the need of this new field though the users already has the ability to set the annotations based on workload config?
> 
> you make a good point here that I didn't consider. An admin does have the option to do something like
> 
> ```
> [crio.runtime.workloads.default]
> activation_annotation = "io.cri-o.default"
> allowed_annotations = [ "io.cri-o.default" ]
> ```
> 
> and any user who specifies that annotation in their pod would get that feature. Since all of the allowed annotations require pod authors to opt-in, an admin can sort of already specify allowed annotations being allowed by everyone.
> 
> we may not actually need this after all...

Cool. And what about this PR? Is this valid still or we can close this!!

haircommander:
I think we can close this! default allowed annotations can be applied as above

PR #4439, opened 2020-12-16
Refactor managed namespace handling

haircommander:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation

/kind design

> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

This PR refactors the namespace pinning code, which was previously all over the place.

Now, we have a dedicated package for creating managed namespaces: nsmgr. When a pod needs a new set of namespaces, the server can request nsmgr to create them. 

This PR also updates a couple of things: docs, function names and object names, as well as cleans up unused code.
a subset of https://github.com/cri-o/cri-o/pull/4431: all of the non-behavior changing commits 

```release-note
none
```


internal/lib/sandbox/namespaces.go
```diff
@@ -58,54 +57,19 @@ type NamespaceOption struct {
...
-	Remove() error
-
-	// Path returns the bind mount path of the namespace
-	Path() string
-
-	// Type returns the namespace type (net, ipc or uts)
-	Type() NSType
-}
-
+// TODO FIXME do I keep this?
```
haircommander, line 60: oop

internal/config/nsmgr/namespaces.go
```diff
@@ -0,0 +1,59 @@
...
+	// Initialized returns true if already initialized
+	Initialized() bool
+
+	// Remove ensures this network namespace handle is closed and removed
+	Remove() error
+
+	// Path returns the bind mount path of the namespace
+	Path() string
+
+	// Type returns the namespace type (net, ipc or uts)
```
mrunalp: comment missing `user`

internal/config/nsmgr/namespaces.go
```diff
@@ -0,0 +1,59 @@
...
+	Type() NSType
+}
+
+type namespacePinner func([]NSType, *idtools.IDMappings, map[string]string) ([]NamespaceIface, error)
+
+func (mgr *managedNamespaceManager) NewPodNamespaces(managedNamespaces []NSType, idMappings *idtools.IDMappings, sysctls map[string]string) ([]NamespaceIface, error) {
+	return mgr.NewPodNamespacesWithFunc(managedNamespaces, idMappings, sysctls, mgr.pinNamespaces)
+}
+
+// CreateManagedNamespacesWithFunc is mainly added for testing purposes. There's no point in actually calling the pinns binary
```
mrunalp: Comment doesn't match function name.

internal/lib/sandbox/namespaces.go
```diff
@@ -114,72 +78,30 @@ func (m *ManagedNamespace) Path() string {
...
+		case nsmgr.NETNS:
+			s.netns = ns
+		case nsmgr.USERNS:
+			s.userns = ns
+		// this should never happen
 		default:
-			// This should never happen
-			err = errors.New("Invalid namespace type")
-			return typesAndPaths, err
+			continue
```
kolyshkin, line 102: Why are unknown entries skipped? If "this should never happen" and there's no way / need to return an error, the code can `panic()` here.

internal/config/nsmgr/namespaces.go
```diff
@@ -0,0 +1,59 @@
...
+	NETNS                NSType = "net"
+	IPCNS                NSType = "ipc"
+	UTSNS                NSType = "uts"
+	USERNS               NSType = "user"
+	PIDNS                NSType = "pid"
+	ManagedNamespacesNum        = 4
+)
+
+// NamespaceIface provides a generic namespace interface
+type NamespaceIface interface {
```
mrunalp: Could you rearrange the commits? I see you added code in this commit that is being dropped in the second commit.

haircommander: I believe rearranging would basically end up looking like the commits are squashed. would you just like me to do that? I left it this way to make the 1:1 relationship between this code living in nsmgr vs sandbox more clear

haircommander:
the only way to rearrange was to squash everything together. I managed to keep one commit out. If need be, I backed up the old commit structure so I can revert if there's other suggestions

internal/lib/sandbox/namespaces.go
haircommander: fixed!

internal/config/nsmgr/namespaces.go
haircommander: fixed!

internal/config/nsmgr/namespaces.go
haircommander: fixed!

internal/config/nsmgr/nsmgr.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// NamespaceManager manages the server's namespaces.
+// Specifically, it is an interface for how the server is creating namespaces.
+// and can be requested to create namespaces for a pod.
+type NamespaceManager interface {
```
kolyshkin, line 21: Personally I find the practice of using an interface that only has a single implementation excessive. I might be overlooking something, but what are the chances cri-o will have one more implementation of `NamespaceManager`? If chances are low, we should not be use an interface here, as this is an abstraction that comes with a price tag.

internal/config/nsmgr/nsmgr.go
```diff
@@ -0,0 +1,157 @@
...
+			if err != nil {
+				return nil, err
+			}
+			f, err := os.Create(pinPath)
+			if err != nil {
+				return nil, err
+			}
+			f.Close()
+			if err := os.Chown(pinPath, rootPair.UID, rootPair.GID); err != nil {
+				return nil, err
```
kolyshkin, line 99: perhaps this can be moved out to a function, to make `NewPodNamespaces` a bit short and easier to read.

internal/config/nsmgr/nsmgr.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+		}
+	}
+
+	if idMappings != nil {
+		pinnsArgs = append(pinnsArgs,
+			fmt.Sprintf("--uid-mapping=%s", getMappingsForPinns(idMappings.UIDs())),
+			fmt.Sprintf("--gid-mapping=%s", getMappingsForPinns(idMappings.GIDs())))
```
kolyshkin, line 107: I'd use simple string concatenation, i.e.
```go
"--uid-mapping=" + getMappingsForPinns(idMappings.UIDs())
```

kolyshkin: Of course, if you just move the code around it's better to make such changes in a separate commit.

internal/config/nsmgr/nsmgr.go
```diff
@@ -0,0 +1,157 @@
...
+	namespacesDir string
+	pinnsPath     string
+}
+
+// NewPodNamespaces creates new namespaces for a pod.
+// It's responsible for running pinns and creating the Namespace objects.
+// The caller is responsible for cleaning up the namespaces by calling Namespace.Remove().
+func (mgr *namespaceManager) NewPodNamespaces(managedNamespaces []NSType, idMappings *idtools.IDMappings, sysctls map[string]string) ([]Namespace, error) {
+	if len(managedNamespaces) == 0 {
+		return []Namespace{}, nil
```
kolyshkin, line 45: AFAIK this can be `return nil, nil` (no need to create an empty slice).

kolyshkin: It might not be good though (the caller would need an additional `!= nil` check), so ignore this.

internal/config/nsmgr/nsmgr.go
haircommander: that's a great point. there originally were two, but then we dropped `manage_ns_lifecycle`, and I don't see there being a new one any time soon

internal/config/nsmgr/nsmgr.go
haircommander: yeah I am worried about needing the nil check, so I'd rather just make a new one here

internal/config/nsmgr/nsmgr.go
haircommander: addressed

internal/config/nsmgr/nsmgr.go
haircommander: addressed

mrunalp:
LGTM

haircommander:
tests :heavy_check_mark: comments :heavy_check_mark: 
PTAL @kolyshkin @giuseppe @mrunalp @umohnani8 @saschagrunert

internal/config/nsmgr/nsmgr.go
```diff
@@ -0,0 +1,150 @@
...
+
+	"github.com/containers/storage/pkg/idtools"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// NamespaceManager manages the server's namespaces.
+// Specifically, it is an interface for how the server is creating namespaces.
```
saschagrunert, line 19: nittiest-nit:
```suggestion
// Specifically, it is an interface for how the server is creating namespaces
```

saschagrunert (approved):
LGTM, just found one nit which may not worth re-triggering the CI

PR #4449, opened 2020-12-23
Add initial Kata Containers support to contrib/test/integration playbook

fgiudici:
#### What type of PR is this?

/kind ci

This PR introduces support to Kata Containers  runtime in the contrib/test/integration playbook to allow to run integration tests via ansible. A new boolean environment variable has been added (build_kata). When set to true, Kata Container packages are installed and the environment configured to use Kata as the default runtime (shimv2 binary).
Only a small subset of the integration tests are run: many are not working with Kata Containers yet.
Targeted distributions are Fedora and CentOS.

None

This PR aims to bring initial support to Kata Containers. The set of integration tests run is small, but tested to pass on CentOS 7 and Fedora 32/33. More integration tests may be later added and Kata Containers packages may be updated as newer stable versions are released.

```release-note
None
```

### DIFF
--- a/contrib/test/integration/vars.yml
+++ b/contrib/test/integration/vars.yml
@@ -8,6 +8,7 @@ manage_ns_lifecycle: True
 
 build_runc: True
 build_crun: False
+build_kata: False
 cgroupv2: False
 
 # For results.yml Paths use rsync 'source' conventions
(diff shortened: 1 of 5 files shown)


contrib/test/integration/vars.yml
```diff
@@ -31,3 +32,33 @@ ssh_location: "{{ lookup('env','HOME') }}/.ssh/id_rsa"
...
+
+kata_int_test_env:
+    CONTAINER_RUNTIMES: "containerd-shim-kata-v2:/usr/bin/containerd-shim-kata-v2:/run/vc:vm"
+    CONTAINER_RUNTIME: "containerd-shim-kata-v2"
+    CONTAINER_DEFAULT_RUNTIME: "containerd-shim-kata-v2"
+    RUNTIME_NAME: "containerd-shim-kata-v2"
+    RUNTIME_ROOT: "/run/vc"
+    RUNTIME_TYPE: "vm"
+    TESTFLAGS: "ctr.bats"
+    JOBS: "1"
```
haircommander, line 44: why only one job here? can kata runtime not handle parallelization of more jobs?

haircommander:
this looks great @fgiudici I have one question but it largely LGTM 
/ok-to-test

contrib/test/integration/vars.yml
fgiudici: Well, you are right, it should... parallelization was disabled when the tests were run using devicemapper as the storage driver, but here it is not the case any more.
So, just tried to run the test enabling back parallelization... jobs started to fail :-( 
We need to disable parallelization for now, but to understand why some more investigation is needed.

haircommander: can you throw a TODO in there to not loose track of it, and upon this merging open an issue referencing the TODO?

fgiudici: Sure, nice advice! Done

mrunalp:
@fidencio ptal

fidencio:
Let me add a few notes for the future so those don't get lost on my mind.
* A lot of skipped tests: Would it be worth opening issues for each one of those?  Have you had the time to investigate why those are not passing?
* s390x? :-) ... For 2.0 we may need to focus on testing against x86_64 only, at least at the beginning, as we're still facing a bunch of issues with (the rust version of) kata-agent, when statically linked against glibc;
  * https://github.com/kata-containers/kata-containers/pull/1236

By the moment it gets merged, is it going to run on every PR? If so, yay, lovely!

fidencio:
LGTM, thanks @fgiudici! Also, thanks @mrunalp for pointing it out in my direction, I had missed it.

fgiudici:
> Let me add a few notes for the future so those don't get lost on my mind.
> 
>     * A lot of skipped tests: Would it be worth opening issues for each one of those?  Have you had the time to investigate why those are not passing?

I started investigated but then I stopped (for now) for two main reasons:

- we install the kata packaged with the distribution, so we end up testing different kata versions with the same playbook.
- as you noticed, we have a lot of failures.

I think it is not worth investigating right now: we will test newer versions of kata soon that may have some of the tests already fixed. Building kata from scratch would be another option, but we may end up testing with different built components than what is really used around. Both options may be viable btw.
So the plan is to start with something simple for now, something that just do basic testing that should pass.
Second step (in the near future) is to get a more recent version of kata (either waiting the packages in the distro to get updated or building in the playbook).
Third step is to start trying adding more and more tests, while investigating the failures. This last step may take really a lot of time, but we could do incrementally. 

> 
>     * s390x? :-) ... For 2.0 we may need to focus on testing against x86_64 only, at least at the beginning, as we're still facing a bunch of issues with (the rust version of) kata-agent, when statically linked against glibc;
>       
>       * [kata-containers/kata-containers#1236](https://github.com/kata-containers/kata-containers/pull/1236)
> 

Yeah, just put the check for enabling nested virtualization there... not really planning s390x, at least for now, but it will be there if it will be needed in the future :-)

> 
> By the moment it gets merged, is it going to run on every PR? If so, yay, lovely!
No, not yet. We have to schedule the job in prow. Some other changes needed, but small ones as soon as the playbook is merged and ready :-) In particular, adding CI test against Fedora should be very easy :-)

PR #4450, opened 2020-12-23
contrib/test/integration: parallel build process in CentOS

fgiudici:
#### What type of PR is this?

/kind ci

When running the contrib/test/integration main playbook on CentOS, the parallel installation fails as there is no packaged version of parallel in CentOS 7 nor in CentOS 8.
I guess we should build it as we do for RHEL.
Moreover, the build process ends adding a symbolic link to the just compiled parallel binary: if we don't force the link creation (-f) the operation will fail if the playbook runs twice.

None

Wondering if I'm missing an extra repo that we usually add here and that provides the parallel package for CentOS

```release-note
None
```

### DIFF
--- a/contrib/test/integration/build/parallel.yml
+++ b/contrib/test/integration/build/parallel.yml
@@ -6,18 +6,18 @@
   package:
     name: parallel
     state: present
-  when: (supports_python3 is not defined or not supports_python3) and ansible_distribution in ['Fedora', 'CentOS']
+  when: (supports_python3 is not defined or not supports_python3) and ansible_distribution in ['Fedora']
 
 - name: install parallel package for Fedora > 30
   shell: sudo yum install -y parallel
-  when: supports_python3 is defined and supports_python3 and ansible_distribution in ['Fedora', 'CentOS']
+  when: supports_python3 is defined and supports_python3 and ansible_distribution in ['Fedora']
 
 - name: download parallel sources
   unarchive:
     src: https://ftp.gnu.org/gnu/parallel/parallel-20190322.tar.bz2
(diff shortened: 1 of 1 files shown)


haircommander:
/ok-to-test

/approve

LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

giuseppe (approved):
LGTM

kolyshkin:
> Wondering if I'm missing an extra repo that we usually add here and that provides the parallel package for CentOS

EPEL? https://fedoraproject.org/wiki/EPEL

fgiudici:
> > Wondering if I'm missing an extra repo that we usually add here and that provides the parallel package for CentOS
> 
> EPEL? https://fedoraproject.org/wiki/EPEL

Oh right! Just checked and the parallel package is there.
Moreover, just realized that README in contrib/test/integration says: 
_"The playbooks assume" [...] "On RHEL, the repositories for EPEL, rhel-server, and extras repos are configured and functional."_

What bothered me was that the parallel package was built from sources in RHEL while installed via the package manager in CentOS (following Fedora conditional branch). My goal with this PR was to make CentOS and RHEL behavior the same and as seems to me we use more RHEL than CentOS in CRI-O CI I followed RHEL path without digging too much...
It is worth to notice that at least for the integration tasks it is just enough to let parallel be installed from sources to let CentOS complete the task without the need of the EPEL repo.
Wondering if would be easier at this point to just install parallel from the package manager in all the OS variants (so RHEL too)... but not a big issue IMHO as long as we keep consistent with RHEL and CentOS.

PR #4455, opened 2021-01-06
support short-name aliases

vrothberg:
#### What type of PR is this?

> /kind feature

* Deprecate (insecure) registries in crio.conf.
* Do not check for blocked registries and let containers/image handle it.
* Add support for short-name aliasing.

Please refer to the individual commits.

```release-note
CRI-O now supports short-name aliases which can be configured in the `containers-registries.conf(5)` configuration files.  Please refer to the following article for further details on short-name aliases: www.redhat.com/sysadmin/container-image-short-names

The `registries` option in the crio.conf has been deprecated and takes no effect any more.  CRI-O will warn when loading the config and note that `containers-registries.conf(5)` should be used instead for configuring unqualified-search registries.
```


vrothberg:
@rhatdan @mtrmac @mrunalp PTAL

docs/crio.conf.5.md
```diff
@@ -280,12 +280,6 @@ CRI-O reads its configured registries defaults from the system wide containers-r
 **image_volumes**="mkdir"
   Controls how image volumes are handled. The valid values are mkdir, bind and ignore; the latter will ignore volumes entirely.
 
-**insecure_registries**=[]
-  List of registries to skip TLS verification for pulling images.
```
mtrmac, line 284: Is this worth it? While I’m all for removing as much of the search mechanism as we can get away with, this just forces users to migrate an option from one file to another with not much to gain.

internal/storage/image.go
```diff
@@ -781,57 +770,28 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
...
 	}
 
-	domain, remainder := splitDockerDomain(imageName)
-	if domain != "" {
-		// this means the image is already fully qualified
-		registry, err := sysregistriesv2.FindRegistry(systemContext, imageName)
-		if err != nil {
-			return nil, err
-		}
-		if registry != nil && registry.Blocked {
```
mtrmac, line 791: Note that this looks up based on the user-specified name, but c/image/docker.newDockerClient` works with individual mirrors, so that’s a behavior change.

internal/storage/image.go
```diff
@@ -781,57 +770,28 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
...
+
+	if desc := resolved.Description(); len(desc) > 0 {
+		logrus.Info(desc)
 	}
-	if len(images) == 0 {
-		return nil, fmt.Errorf("all search registries for %q are blocked", remainder)
+
+	images := make([]string, len(resolved.PullCandidates))
+	for i := range resolved.PullCandidates {
+		images[i] = resolved.PullCandidates[i].Value.String()
```
mtrmac, line 784: > // Note that callers *must* call `(PullCandidate).Record` after a returned
> // item has been pulled successfully;

I suppose it’s not quite relevant here in practice; still, in principle it’s an API violation.

docs/crio.conf.5.md
vrothberg: I was under the impression that was the consensus yesterday.  Previously CRI-O did not even consult the registries.conf for insecure registries which drove me (even more) to remove it.

But I am completely open to revert that part.

internal/storage/image.go
```diff
@@ -781,57 +770,28 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
...
 	}
+
 	return imageNamesWithDigestOrTag(images)
```
mtrmac, line 787: This is the only remaining caller; so please update it to work with `resolved.PullCandidates[].Value` directly instead of the current round-trip via `….Value.String()`.

internal/storage/image.go
```diff
@@ -855,24 +815,17 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
...
+	// Look up insecure registries from the registries.conf and try to mark
+	// them as insecure CIDRs.
+	insecureRegistries := []string{}
+	registries, err := sysregistriesv2.GetRegistries(sc)
+	if err != nil {
+		return nil, err
+	}
+	for _, reg := range registries {
+		if reg.Insecure {
+			insecureRegistries = append(insecureRegistries, reg.Location)
```
mtrmac, line 827: - This should be `.Prefix` (the lookup key), not `.Location`
- `registries.conf` prefix values can be namespaces/repos, which this code is not ready to handle
- Using the `registries.conf` data via `IndexConfigs` changes semantics (matches by the user-specified primary name, not for each mirror individually)
- AFAICS this does not really work anyway, because CIDRs in `registries.conf`:
  - are invalid in the documented format; even if CRI-O interpreted them (and the whole implementation of CIDR matching here is completely wrong and unfixable with the Go HTTP stack, BTW), something else would happen with Podman and other consumers
  - are in practice not preserved (see `sysregistriesv2.parseLocation`), so they don’t work anyway.

internal/storage/image_test.go
```diff
@@ -157,16 +159,43 @@ var _ = t.Describe("Image", func() {
...
+				&types.SystemContext{
+					SystemRegistriesConfPath: "../../test/registries.conf",
+				},
+				testImageName,
+			)
 
 			// Then
 			Expect(err).To(BeNil())
 			Expect(names).To(Equal([]string{
-				testDockerRegistry + "/library/" + testImageName,
```
mtrmac, line 165: It’s rather useful to test the `/library/` special case; the three replacement registries are all supposed to behave the same, so that’s not as interesting.

mtrmac (commented):
I don’t see that dropping `insecure_registries` is viable, unless we are ready to drop all of the CIDR matching as well. (I’d love that to happen, but that’s a whole another fight, and one I have lost once or twice before.)

docs/crio.conf.5.md
mtrmac: `c/image/docker.{newDockerClient,newImageSourceAttempt}` apply the `registries.conf` configuration automatically, and that should affect CRI-O as well already without CRI-O having any explicit code to do so.

(Note the other comments, the semantics is a bit different.)

internal/storage/image.go
mtrmac: Of course making CRI-O follow the documented semantics instead of its own is nice — but it’s a behavior change otherwise unrelated to this PR.

docs/crio.conf.5.md
haircommander, line 284: unfortunately, I think we need a release to buffer this change. We shouldn't make such a quick change to crio.conf. We can mark as deprecated this release and drop next

vrothberg: The change for insecure_registries will be reverted.  Dropping registries is required for supporting short-name aliasing without bigger changes to c/image.  @rhatdan @mrunalp @mtrmac and I found that the best way forward, but please raise your voice if you have concerns.

haircommander: another option is we can mark as deprecated mid-1.20. since we only just release 1.20.0, and most people won't move there for a while anyway, I think that would be relatively low risk

vrothberg: So we're good merge into the main branch and would add deprecation warnings to the 1.20 (and 1.21?) branches?

internal/storage/image.go
vrothberg: Do you want me to drop it?

vrothberg: I really don't like that CRI-O is occasionally working around containers/image.  `containers-registries.conf(5)` should be the source of truth.  On the other hand, I understand that behavioural changes may not be desired.

docs/crio.conf.5.md
haircommander: main branch == 1.21 (it auto backports). so yeah if we manually add deprecation warnings in 1.20 I'm okay with dropping here

internal/storage/image.go
mtrmac: (To be fair, historically c/image has only added these features _after_ they have been implemented/copy+pasted in Libpod/… .)

I’m leaning towards dropping this, but I’ll defer to CRI-O maintainers.

vrothberg:
@rhatdan, I am seeing a lot of "could not get either XDG_CONFIG_HOME or HOME" in the logs.  Has that been fixed in c/storage?

vrothberg:
@haircommander @mtrmac @mrunalp @rhatdan PTanotherL

docs/crio.conf.5.md
mtrmac: Now that the `insecure_registries` part has been dropped, the release note in the PR description needs updating.

internal/storage/image.go
```diff
@@ -781,57 +745,45 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
...
+		if isTagged && isDigested {
+			canonical, err = reference.WithDigest(reference.TrimNamed(ref), canonical.Digest())
+			if err != nil {
+				return nil, err
+			}
+			name = canonical.String()
+		} else {
+			name = ref.String()
+		}
+		images[i] = name
```
mtrmac, line 776: Absolutely non-blocking:
```suggestion
			ref = canonical
		}
		images[i] = ref.String()
```
is a tiny bit simpler.

internal/storage/image.go
mtrmac: To be more precise, by “leaning towards dropping this” I mean “dropping this change from the PR, i.e. leaving the `.Blocked` check in”. But it’s a very weak preference and up to CRI-O maintainers to decide.

haircommander:
sounds like something that https://github.com/cri-o/cri-o/pull/4462 will fix

internal/storage/image.go
vrothberg: Thanks, @mtrmac!

@haircommander @mrunalp @rhatdan what do you think?  As Mirek mentioned, if a registry itself is blocked but has mirrors configured, CRI-O would not attempt to pull from these mirrors.

docs/crio.conf.5.md
vrothberg: Cheers!

internal/storage/image.go
vrothberg: Thanks!  I'll apply in case I need another iteration.

vrothberg:
The `integration-critest` in `pipeline` indicates that the environment does not ship a `registries.conf`. @haircommander, any advice?

vrothberg:
> The `integration-critest` in `pipeline` indicates that the environment does not ship a `registries.conf`. @haircommander, any advice?

I think I got it now

vrothberg:
| /test integration_fedora
-- | --

vrothberg:
@mtrmac @rhatdan @umohnani8 @haircommander PTAL

rhatdan:
LGTM

pkg/config/config.go
```diff
@@ -478,6 +478,13 @@ func (c *Config) UpdateFromFile(path string) error {
 		delete(c.Runtimes, defaultRuntime)
 	}
 
+	// Registries are deprecated in cri-o.conf and turned into a NOP.
+	// Users should use registries.conf instead, so let's log it.
+	if len(t.Crio.Image.Registries) > 0 {
+		t.Crio.Image.Registries = nil
+		logrus.Warnf("The 'registries'option has been deprecated but is referenced in %q.  Please use containers-registries.conf(5) instead.", path)
```
TomSweeneyRedHat, line 485:
```suggestion
		logrus.Warnf("The 'registries' option has been deprecated but is referenced in %q.  Please use containers-registries.conf(5) instead.", path)
```

docs/crio.conf.5.md
```diff
@@ -284,7 +284,7 @@ CRI-O reads its configured registries defaults from the system wide containers-r
   List of registries to skip TLS verification for pulling images.
 
 **registries**=["docker.io"]
-  List of registries to be used when pulling an unqualified image (e.g., "alpine:latest"). By default, registries is set to "docker.io" for compatibility reasons. Depending on your workload and usecase you may add more registries (e.g., "quay.io", "registry.fedoraproject.org", "registry.opensuse.org", etc.).
+  List of registries to be used when pulling an unqualified image. Note that this option has been deprecated and has no effect. Please refer to `containers-registries.conf(5)` for configuring unqualified-search registries.
```
TomSweeneyRedHat, line 287: The following is a soft suggestion/thought for the below.  Use if you think any of this is better.

IDK, but maybe this will draw the eye better?  Or some variant?   Perhaps "Deprecated as of v1.20."  Regardless, I don't think you need the "and has no effect." bit.

```suggestion
  List of registries to be used when pulling an unqualified image. DEPRECATED.  Please refer to `containers-registries.conf(5)` for configuring unqualified-search registries.
```

mtrmac: I tend to understand “deprecated” as “discouraged but continues to work”; if it is not just deprecated but basically removed, I think that should be documented explicitly. (Or, maybe, just remove the option documentation entirely.)

OTOH I don’t see much value in documenting what the option did _in the past_; that can distract readers from reading on to notice that it doesn’t do that any more.

mtrmac (commented):
Code LGTM; documentation nitpicking can continue or not depending on mood and consensus.

docs/crio.conf.5.md
haircommander: In this PR I think we should drop the entry entirely, and mark the deprecation notice in 1.20

internal/storage/image.go
haircommander: I would be fine dropping this field as well TBH, instead of trying to work around it

docs/crio.conf.5.md
TomSweeneyRedHat: @haircommander dropping would be fine with me.  I was just trying to nitpick something based on the prior comments.

pkg/config/config.go
haircommander, line 485: this part can go in the 1.20 version, and we can just drop this entirely

docs/crio.conf.5.md
vrothberg: I concur with @mtrmac's understanding of deprecated.  The important part is that "it takes no effect".  I see wisdom in helping users resolve a problem.  They may not (always) look at the logs.

pkg/config/config.go
vrothberg: I see wisdom in keeping this for a longer time.  Not all users go through all versions, docs can be outdated etc.  I think it doesn't hurt us to keep it around but can be a time safer for users.

haircommander:
All in all, the code itself LGTM, now we just work on messaging. Since we're in the process of deprecating/dropping insecure registries, it makes sense to me to drop any of the cri-o specific fields that don't make sense to y'all c/image maintainers. I don't think there's that much value add to having both (I would guess most users run either CRI-O or podman on a system, and often run podman rootless if they are using both). 

Thus, we can drop all traces of the unneeded fields in this PR, and the follow up will add the deprecation notices in the 1.20 branch

Thanks for taking this on @vrothberg

pkg/config/config.go
haircommander: :shrug: it's up to you, because any issues with the field will likely cause us to call you to help :laughing:

haircommander: to me, the messaging is clearer if we drop all cri-o specific fields at once

haircommander: sorry, for some reason I thought there was an additional field we were dropping in this PR...

vrothberg: I'll rephrase to state that field does not exist any more.  That may play better with planned deprecation notice for `1.20`.

docs/crio.conf.5.md
vrothberg: I'll drop it :)

pkg/config/config.go
haircommander: given that, I think this is a good strategy for telling users about the drop.
Given that, I suggest this wording: 
`Support for the 'registries' field has been dropped but is referenced in %q...`

docs/crio.conf.5.md
haircommander: sorry my brain is all over the place. we can leave this in for a PR, and I'd suggest the wording: 
```
List of registries to be used when pulling an unqualified image. Note that support for this option has been dropped and setting it has no effect. Please refer to `containers-registries.conf(5)` for configuring unqualified-search registries.
```

internal/storage/image.go
haircommander: I see now that we have dropped it. sorry this review is all over the place :upside_down_face:

pkg/config/config.go
haircommander: and also:
```
Support for the 'registries' option has been dropped but is referenced in %q.  Please use containers-registries.conf(5) instead
```

haircommander: ignore me, this is good, besides the wording nit

docs/crio.conf.5.md
vrothberg: SGTM :)

pkg/config/config.go
vrothberg: SGTM :)

vrothberg:
@haircommander updated :)  In the future, we can also drop the CLI flag.

haircommander:
/approve

LGTM

mtrmac:
> All in all, the code itself LGTM, now we just work on messaging. ~Since we're in the process of deprecating/dropping insecure registries, it makes sense to me to drop any of the cri-o specific fields that don't make sense to y'all c/image maintainers.

To be clear, we are dropping the CRI-O-only list of registries to search when receiving hostname-unqualified names (in favor of a generic registries.conf setting).

We are _not_ dropping the CRI-O-only list of registries to connect without TLS, because the CRI-O-only list allows IP address/network matching, and the c/image one does not. (The IP matching also doesn’t work securely, so I would not object to that being dropped, but that would almost certainly break users without giving them an alternative, and it’s entirely unrelated to this PR.)

haircommander:
PTAL @umohnani8 @mrunalp @kolyshkin @saschagrunert

haircommander:
wake up prow

haircommander:
@vrothberg can you submit a 1.20 PR with the docs changes?

vrothberg:
> @vrothberg can you submit a 1.20 PR with the docs changes?

Thanks for the reminder, will do.

PR #4459, opened 2021-01-06
Provide functionality to start infra containers on the specified set of CPUs

cynepco3hahue:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake
> /kind other

Infra containers can start on every online CPU under the system that can be a problem for the latency-sensitive workloads because every context switch can result in latency peaks.

```release-note
Provide a new configuration flag to specify CPUs that will be used to run infra containers
```


docs/crio.8.md
```diff
@@ -210,6 +211,10 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
        mounted into the container for the volumes.
 	3. ignore: All volumes are just ignored and no action is taken. (default: mkdir)
 
+**--infra-containers-cpus**="": CPU set to run infra containers, if not specified the CRI-O will use all online CPUs to run infra containers (default: "").
+    You can specify CPUs in the Linux CPU list format.
```
haircommander, line 215: suggestion: I think `--infra-cpus` would work too
nit: we should describe that these should be the same as the `guaranteed-cpus` passed to the kubelet, if the user wants their workloads to truly have a unique cpuset

internal/criocli/criocli.go
```diff
@@ -824,3 +839,15 @@ func StringSliceTrySplit(ctx *cli.Context, name string) []string {
 
 	return values
 }
+
+// cpuSetParse parses the the string
```
haircommander, line 843: can you update this comment a bit?

docs/crio.8.md
haircommander: actually, above we use `infra-ctr` so we should use that here `--infra-ctrs-cpus`

internal/oci/oci.go
```diff
@@ -445,6 +447,10 @@ func (r *Runtime) ReopenContainerLog(c *Container) error {
 	return impl.ReopenContainerLog(c)
 }
 
+func (r *Runtime) GetInfraCPUs() *cpuset.CPUSet {
+	return r.config.InfraContainersCPUs
+}
+
```
haircommander, line 453: I'm not sure this really makes sense in the runtime object. Either in the server config (`pkg/config/config.go`) or in a special package makes more sense to me

server/container_create_linux.go
```diff
@@ -343,9 +343,16 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
 			specgen.SetProcessOOMScoreAdj(int(resources.OomScoreAdj))
-			specgen.SetLinuxResourcesCPUCpus(resources.CPUsetCPUs)
 			specgen.SetLinuxResourcesCPUMems(resources.CPUsetMems)
 
+			specgen.SetLinuxResourcesCPUCpus(resources.CPUsetCPUs)
+			if resources.CPUsetCPUs == "" {
+				infraCPUs := s.Runtime().GetInfraCPUs()
+				if infraCPUs != nil {
+					specgen.SetLinuxResourcesCPUCpus(infraCPUs.String())
+				}
```
haircommander, line 353: we don't want to set this here I believe. what this is doing is settting a new, non-infra container to use infra cpus, if it has cpus set. What we should do is set infra containers to be in those cpus (in server/sandbox_run_linux.go). Or, we can move the infra container to the cpus when the container is created (similar to how a container is in container_update_resources)

does that make sense?

haircommander: thinking about this more, we should unconditionally set the infra container's cpuset if `infra-ctr-cpus` is set. we'll have to do that in `server/sandbox_run_linux.go`

docs/crio.8.md
cynepco3hahue: done

internal/criocli/criocli.go
cynepco3hahue: Thanks, I forgot to do it, done

internal/oci/oci.go
cynepco3hahue: I removed this function.

@haircommander Does it ok to have the `infra_ctr_cpus` parameter under the
```
[runtime]
...
infra_ctr_cpus=<something>
...
```
or should I move this parameter under the separate section(probably in the future we can have the same functionality for the memory pinning and more).

server/container_create_linux.go
cynepco3hahue: done

docs/crio.8.md
```diff
@@ -210,6 +211,10 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
        mounted into the container for the volumes.
 	3. ignore: All volumes are just ignored and no action is taken. (default: mkdir)
 
+**--infra-ctrs-cpus**="": CPU set to run infra containers, if not specified the CRI-O will use all online CPUs to run infra containers (default: "").
+    You can specify CPUs in the Linux CPU list format.
+
+
```
TomSweeneyRedHat, line 217: Remove extra carriage return please.
```suggestion
```

docs/crio.conf.5.md
```diff
@@ -225,6 +225,10 @@ the container runtime configuration.
 **drop_infra_ctr**=false
   Determines whether we drop the infra container when a pod does not have a private PID namespace, and does not use a kernel separating runtime (like kata).
   Requies **manage_ns_lifecycle** to be true.
+  
+**infra-ctrs-cpus**=""
+    Determines the CPU set to run infra containers, if not specified the CRI-O will use all online CPUs to run infra containers.
```
TomSweeneyRedHat, line 230:
```suggestion
    Determines the CPU set to run infra containers. If not specified, the CRI-O will use all online CPUs to run infra containers.
```

docs/crio.8.md
cynepco3hahue: Thanks, Tom, done.

docs/crio.conf.5.md
cynepco3hahue: done

cynepco3hahue:
/hold
I am still need to check it locally

docs/crio.8.md
mrunalp: How about `infra-cpuset`?

mrunalp: `cpus` is too broad and could also refer to amount of cpu shares reserved or cpu quota if we want to add those in the future. So, I think it may be better to be explicit.

cynepco3hahue: makes sense, done

pkg/config/template.go
```diff
@@ -253,6 +253,10 @@ ctr_stop_timeout = {{ .CtrStopTimeout }}
 # It requires manage_ns_lifecycle to be true.
 drop_infra_ctr = {{ .DropInfraCtr }}
 
+# infra_ctr_cpuset determines what CPUs will be used to run infra containers.
+# You can use linux CPU list format to specify desired CPUs.
```
haircommander, line 257: I also think it would be worth mentioning that this option needs to be specified with kubelet's `--reserved-cpus` for a container to be truly guaranteed a cpu

haircommander:
there's some infra instability btw, test failures not the fault of this pr

pkg/config/template.go
cynepco3hahue: Makes a lot of sense, done

completions/fish/crio.fish
```diff
@@ -109,10 +110,10 @@ complete -c crio -n '__fish_crio_no_subcommand' -l runroot -r -d 'The CRI-O stat
 complete -c crio -n '__fish_crio_no_subcommand' -f -l runtimes -r -d 'OCI runtimes, format is runtime_name:runtime_path:runtime_root:runtime_type'
 complete -c crio -n '__fish_crio_no_subcommand' -l seccomp-profile -r -d 'Path to the seccomp.json profile to be used as the runtime\'s default. If not specified, then the internal default seccomp profile will be used. (default: "")'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l seccomp-use-default-when-empty -r -d 'Use the default seccomp profile when an empty one is specified (default: false)'
-complete -c crio -n '__fish_crio_no_subcommand' -f -l selinux -d 'Enable selinux support (default: false)'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l selinux -d 'Enable selinux support (default: true)'
```
haircommander, line 113: due to a limitation of the `docs-generation` and `completion` targets, it currently takes into account your crio configuration (I'm looking into it now-it's been a minor annoyance for a long time). So you'll want to do
`sed -i 's/Enable selinux support (default: true)/Enable selinux support (default: false)/g' completions/fish/crio.fish docs/crio.8.md`

and 
`sed -i 's/default: "overlay"/default: ""/g' completions/fish/crio.fish docs/crio.8.md`

I recommend sed over manually editing because I use vim, and vim spits a EOL at the end of the files which we don't currently have. You can manually edit as well, but I find doing `:set binary` and `:set noeol` annoying

docs/crio.8.md
```diff
@@ -273,15 +276,15 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
...
 
 **--separate-pull-cgroup**="": [EXPERIMENTAL] Pull in new cgroup (default: "")
 
 **--signature-policy**="": Path to signature policy JSON file. (default: "", to use the system-wide default)
 
-**--storage-driver, -s**="": OCI storage driver (default: "")
+**--storage-driver, -s**="": OCI storage driver (default: "overlay")
 
-**--storage-opt**="": OCI storage driver option (default: [])
+**--storage-opt**="": OCI storage driver option (default: [overlay.mountopt=nodev,metacopy=on overlay.mountopt=nodev,metacopy=on])
```
haircommander, line 287: the same as above applies here too

completions/fish/crio.fish
cynepco3hahue: Good catch thanks, done.

docs/crio.8.md
cynepco3hahue: done

haircommander:
/approve

LGTM, thanks @cynepco3hahue 
PTAL @kolyshkin @mrunalp @umohnani8 @giuseppe

mrunalp:
Can you add an integration test?

server/sandbox_run_linux.go
```diff
@@ -745,6 +745,12 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 
 	g.SetLinuxResourcesCPUShares(PodInfraCPUshares)
 
+	// When infra-ctr-cpuset specified, set the infra container CPU set
+	if s.config.InfraCtrCPUSet != "" {
+		log.Infof(ctx, "Set the infra container cpuset to %q", s.config.InfraCtrCPUSet)
```
mrunalp, line 750: This could be debug instead.

cynepco3hahue: done

cynepco3hahue:
> Can you add an integration test?

Done

cynepco3hahue:
> Done

@mrunalp Do you know how many CPUs have the CI machines, if it has only one CPU my test will not check anything :(

haircommander:
> > Done
> 
> @mrunalp Do you know how many CPUs have the CI machines, if it has only one CPU my test will not check anything :(

I believe between 4 and 8. if there's just one you can skip the test

docs/crio.conf.5.md
```diff
@@ -225,6 +225,11 @@ the container runtime configuration.
 **drop_infra_ctr**=false
   Determines whether we drop the infra container when a pod does not have a private PID namespace, and does not use a kernel separating runtime (like kata).
   Requies **manage_ns_lifecycle** to be true.
+  
+**infra_ctr_cpuset**=""
+    Determines the CPU set to run infra containers. If not specified, the CRI-O will use all online CPUs to run infra containers.
```
TomSweeneyRedHat, line 230:
```suggestion
    Determines the CPU set to run infra containers. If not specified,  CRI-O will use all online CPUs to run infra containers.
```

internal/criocli/criocli.go
```diff
@@ -794,6 +798,11 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			EnvVars:   []string{"CONTAINER_VERSION_FILE_PERSIST"},
 			TakesFile: true,
 		},
+		&cli.StringFlag{
+			Name:    "infra-ctr-cpuset",
+			Usage:   "CPU set to run infra containers, if not specified the CRI-O will use all online CPUs to run infra containers (default: '').",
```
TomSweeneyRedHat, line 803:
```suggestion
			Usage:   "CPU set to run infra containers, if not specified CRI-O will use all online CPUs to run infra containers (default: '').",
```

docs/crio.conf.5.md
cynepco3hahue: done

internal/criocli/criocli.go
cynepco3hahue: done

PR #4462, opened 2021-01-07
[release-1.20] update containers/storage to v1.24.4

QiWang19:
#### What type of PR is this?

> /kind bug

This is mostly to have https://github.com/containers/storage/commit/83150e3a2fd6219d0a908102645e83770ed35856
which should fix the authentication failure during bootstrap node install.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1907770

```release-note
Fixed a bug where image authentication failed from not finding the auth file.
```


haircommander:
/approve

LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

docs/crio.8.md
```diff
@@ -276,15 +276,15 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--seccomp-use-default-when-empty**="": Use the default seccomp profile when an empty one is specified (default: false)
 
-**--selinux**: Enable selinux support (default: false)
+**--selinux**: Enable selinux support (default: true)
```
haircommander, line 279: unfortunately, our docs generation messes with these values. It seems you can just drop the changes to this file from the commit.

It may be easier to just run 
`sed -i 's|/var/run/|/run|g' docs/crio.8.md`

QiWang19: Only apply this /var/run to /run change for /var/run/containers/storage, or all the /var/run/crio/* in this readme should be change to /run/*?  It looks like the failure only complains about the **--runroot**

haircommander: sorry, you're right, we only need it for the run root

PR #4470, opened 2021-01-13
Properly implement metrics for Kata Containers when using CRI stats.

fidencio:
/kind bug

When using CRI stats, Kata Container pods have no metrics at all.  This happens because CRI-O doesn't know how to properly unmarshal the message coming from Kata Containers.

None

This PR, as it's at the moment it was opened, allows CRI-O to return something to kubelet, although the returned values for memory are clearly wrong. I'll keep updating the description according to the progress made.

Last but not least, we should compare results with what containerd provides, as I do believe we should be getting information about Block & Net Input & Ouput.

```release-note
None
```
Or should we add a note about this change?


fidencio:
Okay, with the last commit CRI stats are mostly functional when used with the "vm" runtime type.  I still need to write tests for those, though.

internal/oci/runtime_vm.go
```diff
@@ -733,6 +733,66 @@ func (r *runtimeVM) ContainerStats(ctx context.Context, c *Container, _ string)
...
+		cpuNano         uint64
+		memUsage        uint64
+		memLimit        uint64
+		memPerc         float64
+		workingSetBytes uint64
+		netInput        uint64
+		netOutput       uint64
+		blockInput      uint64
+		blockOutput     uint64
+		pids            uint64
```
TomSweeneyRedHat, line 748: Personal nit, and since you added workingSetBytes I'll ask, can you alphabetize these?  It's getting long enough where that would be useful during reviews at least.

TomSweeneyRedHat:
One nit comment and some unhappy tests.  
Otherwise LGTM

saschagrunert (approved):
LGTM

fidencio:
Hmm. Sad to see this one was not part of the v1.21.0 release.

PR #4473, opened 2021-01-14
Return runtime API version based on protocol

saschagrunert:
#### What type of PR is this?

/kind feature

To be able to distinguish the chosen CRI API version, we now return the
runtime API version based on the RPC code path.

None
The CRI protocol mentioned that the version has to be semver compatible:
https://github.com/kubernetes/cri-api/blob/3094f92ca2a23915931d9345285a6721be06cf62/pkg/apis/runtime/v1alpha2/api.proto#L143-L145

Before this patch we had returned `v1alpha1`, which is neither the current CRI API version nor semver compatible. Any thought on that?

```release-note
- Changed `VersionResponse.RuntimeApiVersion` to return either the `v1alpha2` or `v1` CRI API version
```

### DIFF
--- a/server/cri/v1/rpc_version.go
+++ b/server/cri/v1/rpc_version.go
@@ -9,7 +9,7 @@ import (
 func (s *service) Version(
 	ctx context.Context, req *pb.VersionRequest,
 ) (*pb.VersionResponse, error) {
-	resp, err := s.server.Version(ctx)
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests.

mrunalp (approved):

PR #4479, opened 2021-01-15
server: support setting raw unified cgroupv2 settings

giuseppe:
#### What type of PR is this?

/kind feature

allow to override the cgroup v2 unified configuration through a custom annotation.  Multiple values are separed with `;`.

Each value can be specified in the base64 format.  If it is not a valid base64 encoded string, then its original value is used.

The container is specified as part of the annotation name.

```
    apiVersion: v1
    kind: Pod
    metadata:
      name: unified-test
      annotations:
        io.kubernetes.cri-o.UnifiedCgroup.cgroup-test: "memory.high=8000000;memory.low=100000"
    spec:
      containers:
      - name: cgroup-test
        image: docker.io/alpine
        command: ["top"]
```

None

```release-note
it is possible to override cgroup v2 unified configuration through the io.kubernetes.cri-o.UnifiedCgroup.$CTR_NAME annotation
```

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -262,6 +262,7 @@ The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  Th
   "io.kubernetes.cri-o.userns-mode" for configuring a user namespace for the pod.
   "io.kubernetes.cri-o.Devices" for configuring devices for the pod.
   "io.kubernetes.cri-o.ShmSize" for configuring the size of /dev/shm.
+  "io.kubernetes.cri-o.UnifiedCgroup.$CTR_NAME" for configuring the cgroup v2 unified block for a container.
 
 ## CRIO.IMAGE TABLE
 The `crio.image` table contains settings pertaining to the management of OCI images.
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -4,6 +4,9 @@ const (
 	// UsernsMode is the user namespace mode to use
 	UsernsModeAnnotation = "io.kubernetes.cri-o.userns-mode"
 
+	// UnifiedCgroupAnnotation specifies the unified configuration for cgroup v2
+	UnifiedCgroupAnnotation = "io.kubernetes.cri-o.UnifiedCgroup"
+
 	// SpoofedContainer indicates a container was spoofed in the runtime
 	SpoofedContainer = "io.kubernetes.cri-o.Spoofed"
 
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -2,6 +2,7 @@ package container
 
 import (
 	"context"
+	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
 	"path/filepath"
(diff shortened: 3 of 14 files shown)


giuseppe:
@mrunalp PTAL

pkg/annotations/annotations.go
```diff
@@ -4,6 +4,9 @@ const (
 	// UsernsMode is the user namespace mode to use
 	UsernsModeAnnotation = "io.kubernetes.cri-o.userns-mode"
 
+	// UnifiedCgroupAnnotation is the user namespace mode to use
```
haircommander, line 7: this comment seems incorrect :)

docs/crio.conf.5.md
```diff
@@ -258,6 +258,7 @@ The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  Th
   "io.kubernetes.cri-o.userns-mode" for configuring a user namespace for the pod.
   "io.kubernetes.cri-o.Devices" for configuring devices for the pod.
   "io.kubernetes.cri-o.ShmSize" for configuring the size of /dev/shm.
+  "io.kubernetes.cri-o.UnifiedCgroup.$CTR_NAME" for configuring the cgroup v2 unified block for a container.
```
haircommander, line 261: why $CTR_NAME? why not have it apply to the whole pod?

server/container_create_linux.go
```diff
@@ -416,6 +470,9 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 			}
 		}
 	}
+	if err := s.addUnifiedResources(specgen, sb, containerConfig); err != nil {
+		return nil, err
```
haircommander, line 474: I'm not sure this fits in with `finalizeUserMapping`. This seems like something that should be in `pkg/container/container.go`

docs/crio.conf.5.md
mrunalp: cgroups are set at the container level in pod yaml

haircommander: .. fair enough :)

pkg/annotations/annotations.go
giuseppe: thanks for noticing my copy&pasta.  I'll fix it

server/container_create_linux.go
giuseppe: good hint!  I'll move it

pkg/container/container.go
```diff
@@ -429,3 +435,47 @@ func (c *container) SelinuxLabel(sboxLabel string) ([]string, error) {
 	}
 	return ret, nil
 }
+
+// AddUnifiedResources adds the cgroup-v2 resources specified in the io.kubernetes.cri-o.UnifiedCgroup annotation
+func (c *container) AddUnifiedResources(specgen *generate.Generator, sb *sandbox.Sandbox, containerConfig *types.ContainerConfig) error {
```
haircommander, line 440: nit: the spec generator and container config are already members, we don't need to pass them in :)

haircommander: I also think we could just pass in the annotations, which will make unit testing easier (don't have to initialize a whole sandbox). If you make this change, I'd also change the function to `AddUnifiedResourcesFromAnnotation` or something

haircommander:
a couple more nits, as well as: could you add some unit tests in `pkg/container` to cover some of the new code paths?

pkg/container/container.go
giuseppe: thanks!  Amended in the last version

TomSweeneyRedHat:
LGTM

haircommander:
LGTM

haircommander:
LGTM

@mrunalp @kolyshkin @umohnani8 @saschagrunert @fidencio @wgahnagl @QiWang19 PTAL

PR #4485, opened 2021-01-17
server: leave swap mem limit unset if not supported

giuseppe:
if the memory swap limit is not supported by cgroup, leave the value
unset instead of forcing it to 0.

#### What type of PR is this?
/kind bug

if swap memcg is not supported, leave the value unset instead of setting it to 0

None

```release-note
None
```

### DIFF
--- a/server/container_update_resources.go
+++ b/server/container_update_resources.go
@@ -35,10 +35,10 @@ func (s *Server) UpdateContainerResources(ctx context.Context, req *types.Update
 
 // toOCIResources converts CRI resource constraints to OCI.
 func toOCIResources(r *types.LinuxContainerResources) *rspec.LinuxResources {
-	var swap int64
+	var swap *int64
 	memory := r.MemoryLimitInBytes
 	if node.CgroupHasMemorySwap() {
-		swap = memory
+		swap = proto.Int64(memory)
 	}
 	return &rspec.LinuxResources{
 		CPU: &rspec.LinuxCPU{
@@ -50,7 +50,7 @@ func toOCIResources(r *types.LinuxContainerResources) *rspec.LinuxResources {
 		},
 		Memory: &rspec.LinuxMemory{
 			Limit: proto.Int64(memory),
-			Swap:  proto.Int64(swap),
+			Swap:  swap,
 		},
 		// TODO(runcom): OOMScoreAdj is missing
 	}


fidencio:
@giuseppe, thanks for the PR.
I'd really like to have a reference for the original issue in the commit message, so one can easily get to the reasoning behind this patch.

Anyways, LGTM!

lyind:
@giuseppe Hoping this would fix https://github.com/containers/crun/issues/570 I applied this PR to cri-o 1.20 (not master) and tested it. Issue is still present:
```
  Warning  Failed  98s (x12 over 3m48s)  kubelet  Error: container create failed: opening file `memory.swap.max` for writing: Permission denied
```

Out of curiousity I also looked into `CgroupHasMemorySwap()` and I do not quite understand:

Even on systems with the memory.swap knobs present:
```
$ ls -1 /sys/fs/cgroup/kubepods
...
memory.stat
memory.swap.current
memory.swap.events
memory.swap.high
memory.swap.max
pids.current
...
```

There is no such file `/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes` present. In fact, not event the directory `/sys/fs/cgroup/memory` exists on the host. Was that valid under cgroupv1?

The contents of `cgroup.controllers` also doesn't differ between a host with CONFIG_SWAP and without:

One reliable way to detect swap support on the host: **check if `/proc/swaps` is readable**.

Of course this only tests for CONFIG_SWAP and not for CONFIG_MEMCG_SWAP. Thus I just verified the Linux configuration options (`make menuconfig`). One can't specifically select/de-select `CONFIG_MEMCG_SWAP`, it is automatically enabled on `CONFIG_MEMCG && CONFIG_SWAP`:
```
--- Control Group support 
 [*]   Memory controller     
 [*]   IO controller              
 -*-   CPU controller  --->  
 [*]   PIDs controller          
 [ ]   RDMA controller        
 [*]   Freezer controller      
 [*]   Cpuset controller       
 [*]     Include legacy /proc/<pid>/cpuset file  
 [*]   Device controller                                    
 [*]   Simple CPU accounting controller         
 [*]   Perf controller                                        
 [*]   Support for eBPF programs attached to cgroups 
 [ ]   Debug controller  
```

lyind:
Event when using `stat('/proc/swaps')` in cri-o 1.20 `CgroupHasMemorySwap()` and deleting the pod, the re-created pod still fails on the non-swap host with:
```
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  72s               default-scheduler  Successfully assigned kube-system/coredns-74ff55c5b-trvpx to node-REDACTED
  Normal   Pulled     6s (x7 over 71s)  kubelet            Container image "k8s.gcr.io/coredns:1.7.0" already present on machine
  Warning  Failed     6s (x7 over 71s)  kubelet            Error: container create failed: opening file `memory.swap.max` for writing: Permission denied
```

giuseppe:
> ```
> 
> ```
> 
> 
> There is no such file `/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes` present. In fact, not even the directory `/sys/fs/cgroup/memory` exists on the Linux 5.10.7 cgroupv2 host. Was that valid under cgroupv1?

oh I see now the issue.  On cgroup v2 we are assuming the swap cgroup is always present:

```
// CgroupHasMemorySwap returns whether the memory swap controller is present
func CgroupHasMemorySwap() bool {
	cgroupHasMemorySwapOnce.Do(func() {
		if CgroupIsV2() {
			cgroupHasMemorySwap = true
			return
		}

		_, err := os.Stat("/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes")
		if err != nil {
			cgroupHasMemorySwapErr = errors.New("node not configured with memory swap")
			cgroupHasMemorySwap = false
			return
		}

		cgroupHasMemorySwap = true
	})
	return cgroupHasMemorySwap
}
```

giuseppe:
> @giuseppe, thanks for the PR.
> I'd really like to have a reference to the original issue in the commit message, so one can easily get to the reasoning behind this patch.
> 
> Anyways, LGTM!

Thanks for the review.  I've added more details to the commit message.  I've not any issue to point at, I was looking the reason for a different failure and spotted we were setting mem swap also when it is not supported.

TomSweeneyRedHat:
LGTM assuming happy tests.

giuseppe:
@saschagrunert @haircommander PTAL

saschagrunert (approved):
LGTM

PR #4487, opened 2021-01-18
Close the stdin/tty on server start to avoid shortname prompts

saschagrunert:
#### What type of PR is this?

/kind bug

This avoids prompting for shortname expansion if CRI-O runs inside of a terminal.

Refers to https://github.com/cri-o/cri-o/pull/4455
/assign @vrothberg 

```release-note
None
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -32,6 +32,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


vrothberg:
This should not happen in CRI-O unless CRI-O leaves stdin/tty open which it shouldn't as a daemon.  Forcing to disabled mode may not be desired when user want to enforce it in the registries.conf.

Did you run into CRI-O prompting?

saschagrunert:
> Did you run into CRI-O prompting?

Yes, I think we will not run into this issue when using CRI-O with systemd.

When running CRI-O in a standard TTY like my development setup, then it will ask for the name to resolve which seems annoying. Do you see a way to disable at least the prompt without breaking the behavior?

/hold

vrothberg:
> > Did you run into CRI-O prompting?
> 
> Yes. So, we will not run into this issue when using CRI-O with systemd.
> 
> When running CRI-O in a standard TTY like my development setup, then it will ask for the name to resolve which seems annoying. Do you see a way to disable at least the prompt without breaking the behavior?

Oh that's unfortunate - I should have caught that in my PR. Thanks for tackling that so quickly!

I think CRI-O could use the same trick as we applied to Podman: https://github.com/containers/podman/commit/b96c3489d043

saschagrunert:
Thank you for the hint to the commit @vrothberg, changed the PR as suggested.

vrothberg (commented):
LGTM, thanks @saschagrunert !

rhatdan:
lgtm

TomSweeneyRedHat:
LGTM

PR #4492, opened 2021-01-19
Do not log file path on ioutil.ReadFile

saschagrunert:
#### What type of PR is this?

/kind cleanup

The error message will already indicate which file is affected by the
failure. This way we do not have to need to log the filename multiple
times to lower the logging noise.

None
None

```release-note
None
```


TomSweeneyRedHat (commented):
LGTM
assuming happy tests

PR #4495, opened 2021-01-19
update pinns to support pinning host level namespaces

haircommander:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake
> /kind other

When we manage NS lifecycle, we currently do nothing for host level namespaces, instead using the /proc/ entry of the infra container for the location of that namespace.

The problem is, when we drop infra, there's no pid to use in the /proc/ entry, so we don't know where the namespace is.

This means, when we drop the infra container, we actually can't use port_forward on pods that use the host network... oops.

This PR, among many other things, updates pinns to support binding the location of the namespace, but not unsharing, thus binding the location of the host namespace.

replaces https://github.com/cri-o/cri-o/pull/4431

```release-note
Fix a bug where pods with `hostNetwork: true` couldn't have ports forwarded from them when drop_infra_ctr=true
```

### DIFF
--- a/internal/config/nsmgr/nsmgr.go
+++ b/internal/config/nsmgr/nsmgr.go
@@ -34,16 +34,19 @@ func New(namespacesDir, pinnsPath string) *NamespaceManager {
 // NewPodNamespaces creates new namespaces for a pod.
 // It's responsible for running pinns and creating the Namespace objects.
 // The caller is responsible for cleaning up the namespaces by calling Namespace.Remove().
-func (mgr *NamespaceManager) NewPodNamespaces(managedNamespaces []NSType, idMappings *idtools.IDMappings, sysctls map[string]string) ([]Namespace, error) {
-	if len(managedNamespaces) == 0 {
+func (mgr *NamespaceManager) NewPodNamespaces(cfg *PodNamespacesConfig) ([]Namespace, error) {
+	if cfg == nil {
+		return nil, errors.New("PodNamespacesConfig cannot be nil")
+	}
+	if len(cfg.Namespaces) == 0 {
 		return []Namespace{}, nil
 	}
 
 	typeToArg := map[NSType]string{
-		IPCNS:  "-i",
-		UTSNS:  "-u",
-		USERNS: "-U",
-		NETNS:  "-n",
+		IPCNS:  "--ipc",
+		UTSNS:  "--uts",
+		USERNS: "--user",
+		NETNS:  "--net",
 	}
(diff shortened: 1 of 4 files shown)


giuseppe (approved):

saschagrunert (approved):
LGTM

pinns/src/pinns.c
```diff
@@ -254,6 +263,11 @@ int main(int argc, char **argv) {
   return EXIT_SUCCESS;
 }
 
+// returns true if the option is empty or not equal to 'host'
+static bool is_not_host_ns(const char* const optarg) {
```
mrunalp, line 267: I think inverting this function to `is_host_ns` and adding the `!` at the caller is easier to read.

internal/config/nsmgr/nsmgr.go
```diff
@@ -69,6 +69,9 @@ func (mgr *NamespaceManager) NewPodNamespaces(cfg *PodNamespacesConfig) ([]Names
 		if !ok {
 			return nil, errors.Errorf("Invalid namespace type: %s", ns.Type)
 		}
+		if ns.Host {
+			arg += "=host"
```
mrunalp, line 73: may want to share a constant between pinns and here but could be done later.

pinns/src/pinns.c
haircommander: to be honest I forget why I did this...

internal/config/nsmgr/nsmgr.go
haircommander: yeah especially if pinns is taken out of this repo. It's not super trivial to do so, so I'll leave it out for now.

pinns/src/pinns.c
```diff
@@ -63,28 +66,34 @@ int main(int argc, char **argv) {
   while ((c = getopt_long(argc, argv, "pchuUind:f:s:", long_options, NULL)) != -1) {
     switch (c) {
     case 'u':
-      unshare_flags |= CLONE_NEWUTS;
+      if (is_host_ns (optarg))
```
mrunalp, line 69: You need to add a `!` for the `is_host_ns` invocations.

haircommander: what a foolish mistake :upside_down_face: thanks, fixed!

PR #4500, opened 2021-01-20
Make rootless namespace unit test execution work

saschagrunert:
#### What type of PR is this?

/kind failing-test

We should target to get the unit tests back to a state where we can
execute them in a rootless way. Right now the system user would not have
a chance to access pid 1's /proc/1/ns directory which make the test
fail. This patch uses the current pid, which should be accessible by the
current user executing the tests.

Can be verified by:
```
> make testunit TESTFLAGS=--focus=SandboxManagedNamespaces
```

None
None

```release-note
None
```

### DIFF
--- a/internal/lib/sandbox/namespaces_test.go
+++ b/internal/lib/sandbox/namespaces_test.go
@@ -1,6 +1,7 @@
 package sandbox_test
 
 import (
+	"os"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

TomSweeneyRedHat (commented):
LGTM

haircommander:
/lgtm

good idea!

PR #4501, opened 2021-01-20
Make config tests work rootless

saschagrunert:
#### What type of PR is this?

/kind failing-test

The storage retrieval does not work without providing an UID map in
rootless mode. To avoid adding such non-used options we now exclude
those tests if they run in rootless mode.

Beside that, the conmon path might be different than the hard coded
within the test. This means we now lookup the path during test
execution to make the tests more robust.

None
None

```release-note
None
```


giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #4504, opened 2021-01-20
Remove implicit GitHub action `name` fields

saschagrunert:
#### What type of PR is this?

/kind ci

Some name fields are duplicate because GitHub actions already names the
job based on its identifier. Beside that, it's easier to not name single
line bash jobs because then we can see the actual invocation better in
the UI.

None
None

```release-note
None
```


TomSweeneyRedHat:
LGTM

PR #4508, opened 2021-01-21
pinns: Allow pinning mount namespaces

lack:
Mount namespaces are a bit odd in that:
- Like user namespaces, you must fork and bindmount the mount namespace
  from the parent process (otherwise the bindmount is in the child
  mount namespace where no one sees it)
- Additionally, you cannot bindmount a mount namespace to a file if that
  file is in a shareable mount namespace, otherwise infinite loops are
  possible. So before attempting to bind the mount namespace we
  bind-mount the target directory to itself (if it's not already a
  mountpoint) and mark it as unbindable.

Usage:
    pinns -d ... -f ... [-m|--mnt]

Signed-off-by: Jim Ramsay <jramsay@redhat.com>

#### What type of PR is this?

/kind feature

This adds the ability for the pinns utility to pin mount namespaces.  This will be useful in the implementation of My [hide-container-mountpoints](https://github.com/lack/openshift-enhancements/blob/hide_container_mountpoints/enhancements/hide_container_mountpoints.md) Openshift enhancement proposal.

None

```release-note
None
```

### DIFF
--- a/pinns/src/pinns.c
+++ b/pinns/src/pinns.c
@@ -19,6 +19,8 @@
 #include "sysctl.h"
 
 static bool is_host_ns(const char* const optarg);
+static int setup_unbindable_bindpath(const char *pin_path, const char *ns_name);
+static int create_bind_root(char *bind_root, size_t size, const char *pin_path, const char *ns_name);
 static int bind_ns(const char *pin_path, const char *filename, const char *ns_name, pid_t pid);
 static int directory_exists_or_create(const char* path);
 
@@ -46,6 +48,7 @@ int main(int argc, char **argv) {
   bool bind_ipc = false;
   bool bind_user = false;
   bool bind_cgroup = false;
+  bool bind_mount = false;
   char *sysctls = NULL;
   char res;
 
@@ -56,6 +59,7 @@ int main(int argc, char **argv) {
       {"net", optional_argument, NULL, 'n'},
       {"user", optional_argument, NULL, 'U'},
       {"cgroup", optional_argument, NULL, 'c'},
+      {"mnt", optional_argument, NULL, 'm'},
       {"dir", required_argument, NULL, 'd'},
       {"filename", required_argument, NULL, 'f'},
       {"uid-mapping", optional_argument, NULL, UID_MAPPING},
@@ -63,7 +67,7 @@ int main(int argc, char **argv) {
       {"sysctl", optional_argument, NULL, 's'},
   };
 
-  while ((c = getopt_long(argc, argv, "pchuUind:f:s:", long_options, NULL)) != -1) {
+  while ((c = getopt_long(argc, argv, "mpchuUind:f:s:", long_options, NULL)) != -1) {
     switch (c) {
     case 'u':
       if (!is_host_ns (optarg))
@@ -99,6 +103,12 @@ int main(int argc, char **argv) {
       break;
 #endif
       pexit("unsharing cgroups is not supported by this pinns version");
+    case 'm':
+      if (!is_host_ns (optarg))
+        unshare_flags |= CLONE_NEWNS;
+      bind_mount = true;
+      num_unshares++;
+      break;
     case 'd':
       pin_path = optarg;
       break;
@@ -143,7 +153,7 @@ int main(int argc, char **argv) {
   if (!bind_user && (uid_mapping != NULL || gid_mapping != NULL))
     nexit("Mappings specified without creating a new user namespace");
 
-  if (!bind_user) {
+  if (!bind_user && !bind_mount) {
     /* Use pid=0 to indicate using the current process.  */
     pid = 0;
 
@@ -151,7 +161,7 @@ int main(int argc, char **argv) {
       pexit("Failed to unshare namespaces");
     }
   } else {
-    /* if we create a user namespace, we need a new process.  */
+    /* if we create a user or mount namespace, we need a new process. */
     if (socketpair(AF_UNIX, SOCK_SEQPACKET | SOCK_CLOEXEC, 0, p))
       pexit("socketpair");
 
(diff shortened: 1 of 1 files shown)


saschagrunert (commented):
/ok-to-test

saschagrunert (approved):
LGTM, I think we're at a point where we would need some tests for those code lines.

lack:
@saschagrunert I didn't see any tests about pinns specifically.  Can you point me to where I might add some?

saschagrunert:
> @saschagrunert I didn't see any tests about pinns specifically. Can you point me to where I might add some?

There are none right now, we may follow-up with that in another PR.

pinns/src/pinns.c
```diff
@@ -268,20 +294,62 @@ static bool is_host_ns(const char* const optarg) {
...
+  if (create_bind_root(bind_root, PATH_MAX - 1, pin_path, ns_name) < 0) {
+    return -1;
+  }
+
+  // TODO: Check if bind_root is already a mountpoint
+  // For now, just blindly try to bindmount itself to itself.
+  // If this succeeds, we know it's a mountpoint.  If it fails it probably
+  // failed because it's already a mountpoint, and if it's not, the call to set
+  // MS_UNBINDABLE will fail next.
+  if (mount(bind_root, bind_root, NULL, MS_BIND, NULL) < 0) {
```
haircommander, line 325: it seems we're not reacting to mount()<0. Why have the if statement?

pinns/src/pinns.c
```diff
@@ -268,20 +294,62 @@ static bool is_host_ns(const char* const optarg) {
...
+/* Mount namespaces can only be bound into unsharable mount namespaces (to
+ * avoid infinite loops), so force the pin_path to be a bind-mount to
+ * itself and then marked as unsharable. */
+static int setup_unbindable_bindpath(const char *pin_path, const char *ns_name) {
+  char bind_root[PATH_MAX];
+
+  // first, verify the /$PATH/$NSns directory exists
+  if (create_bind_root(bind_root, PATH_MAX - 1, pin_path, ns_name) < 0) {
+    return -1;
+  }
```
haircommander, line 318: can these functions go beneath their last caller?

haircommander: ah I see setup_unbindable_bindpath is, but create_bind_root could be moved to the bottom of the file

pinns/src/pinns.c
```diff
@@ -196,25 +208,29 @@ int main(int argc, char **argv) {
...
 
-    if (gid_mapping && write_mapping_file(pid, gid_mapping, true) < 0)
-      pexit("Cannot write gid mappings");
+      /* Write user mappings */
+      if (gid_mapping && write_mapping_file(pid, gid_mapping, true) < 0)
+        pexit("Cannot write gid mappings");
 
-    if (uid_mapping && write_mapping_file(pid, uid_mapping, false) < 0)
-      pexit("Cannot write gid mappings");
+      if (uid_mapping && write_mapping_file(pid, uid_mapping, false) < 0)
```
haircommander, line 225: nit: there seem to be some whitespace changes I wouldnt' expect (not sure if they're intentional)

haircommander:
@giuseppe PTAL

pinns/src/pinns.c
lack: Intentional: These were indented into the `if (bind_user) {` on [line 215 above](https://github.com/cri-o/cri-o/pull/4508/files/60986224f0d5dd4879aef283a5713e308e0533bd#diff-bfef5beff459416de7e125890a18464df27f7ada77aa3d6f89eb033fac790801R215).

pinns/src/pinns.c
lack: Will do.

pinns/src/pinns.c
lack: 2 reasons: A lot of coverage tools get cranky if you don't check the return value in some way, and I wanted to make it explicit that we're explicitly ignoring errors.

Happy to do it some other way.

lack: Trying to just ignore it to see if any tooling here complains...

haircommander:
note: ci failures aren't related https://github.com/cri-o/cri-o/issues/4509

pinns/src/pinns.c
```diff
@@ -268,20 +295,48 @@ static bool is_host_ns(const char* const optarg) {
...
+  // first, verify the /$PATH/$NSns directory exists
+  if (create_bind_root(bind_root, PATH_MAX - 1, pin_path, ns_name) < 0) {
+    return -1;
+  }
+
+  // TODO: Check if bind_root is already a mountpoint
+  // For now, just blindly try to bindmount itself to itself, ignoring
+  // failures.  If this succeeds, we know it's a mountpoint.  If it fails it
+  // probably failed because it's already a mountpoint, and if it's not, the
+  // call to set MS_UNBINDABLE will fail next.
```
giuseppe, line 313: could we just attempt an `umount` before the `mount` (or even better run in a loop until it succeeds)?  Just to make sure we don't pile up mounts if it runs multiple times.

giuseppe:
one note for `Hide Container Mountpoints`.  Moving CRI-O to a separate mount namespace will affect also Podman/Skopeo/Buildah, and they should be running from the same mount namespace, otherwise it could cause issues in the underlying storage.

pinns/src/pinns.c
lack: I think doing an `umount` and then mount again would be problematic if `bind_root` is already a mountpoint we don't manage, or if we're reusing the same bind_root to add an additional mount namespace pin to it.  The behavior right now is that if `bind_root` is already mounted (including if it's already a bind-mount to itself like we intend to do the 1st time around), this mount call fails and there's nothing that piles up.

I was looking around for a nice way to check if it's already mounted, but it looked like the only way is to parse and walk through /proc/self/mount[s|info], or to use something like libmount that does this already, and either way it seemed heavy-weight when a try-and-fail-if-it's already-mounted does the job okay.

lack:
FYI: I've done some manual testing of this change, including double-checking that 'nsenter --mount=...' can indeed use the mount namespace file pinned in this way.

pinns/src/pinns.c
giuseppe: > I was looking around for a nice way to check if it's already mounted, but it looked like the only way is to parse and walk through /proc/self/mount[s|info], or to use something like libmount that does this already, and either way it seemed heavy-weight when a try-and-fail-if-it's already-mounted does the job okay.

agreed.  It is not worth to add that complexity

haircommander:
LGTM

how do you feel about this @giuseppe

giuseppe (approved):
LGTM

haircommander:
/lgtm

thanks @lack

lack:
I rebase this to the tip of the current master branch, to try to unstick that CI job...

lack:
@haircommander : This will need a re-lgtm thanks to the rebase.

Issue #4489, opened 2021-01-18
Move to GitHub actions


saschagrunert:
CircleCI seemed to slow-down recently and now we're in a state where some jobs have been transferred to GitHub, whereas others are still running in CircleCI.

The integration tests have a fair amount of dependencies which have to be setup before we could run them in GitHub actions. ~~Beside that, it looks like that we cannot model a pipeline in the same way we do it in CircleCI right now.~~ Edit: We can.

Anyways, I'd prefer moving completely over to GitHub actions.

We also have access to packet.net (now Equinix Metal) resources where we could maintain our own GitHub actions runners (one per machine). This has the drawback that we actually have to maintain the machines on our own.

Any thoughts on that?

Ref https://github.com/cri-o/cri-o/issues/3008
cc @haircommander @kolyshkin

haircommander:
I would be interestd in moving to github actions. runc uses vagrant to set up VMs, so we could do something similar I'd think

Issue #4509, opened 2021-01-21
integration fedora failing


haircommander:
with this error
```
not ok 32 ctr log max with default value
# (in test file ./ctr.bats, line 226)
#   `[ "$len" -eq 250 ]' failed
```
because of this PR https://github.com/opencontainers/runc/pull/2727

because there is an unexpected log line:
```
# 2021-01-21T14:21:17.120377043+00:00 stderr F time="2021-01-21T14:21:17Z" level=warning msg="systemd v241 is too old to support CPUQuotaPeriodSec  (setting will still be applied to cgroupfs)"
```
ref: https://ci.openshift.redhat.com/jenkins/job/test_pull_request_crio_integration_fedora/21857/s3/
it really feels like this error should be captured by cri-o instead of in the container logs, but I don't know if there's anything to do about it.

cc @kolyshkin @mrunalp

haircommander:
it passes now

PR #4510, opened 2021-01-21
Move static build to GitHub actions

saschagrunert:
#### What type of PR is this?
/kind feature

This moves the whole static build pipeline from CircleCI to GitHub
actions, except the integration tests. The integration tests will follow
at some later point, if other integration tests are available and the
overall test strategy is clear.

Refers to #4489 
None

```release-note
- Added `crio-status.8` man page to static release bundle
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,18 @@ jobs:
           path: |
             bin/crio
             bin/crio-status
+      - uses: actions/upload-artifact@v2
+        with:
+          name: docs
+          path: |
+            docs/crio-status.8
+            docs/crio.8
+            docs/crio.conf.5
+            docs/crio.conf.d.5
+      - uses: actions/upload-artifact@v2
+        with:
+          name: config
+          path: crio.conf
 
   validate-docs:
     needs: build
@@ -43,6 +55,7 @@ jobs:
       - run: |
           chmod -R +x bin
           sudo rm /etc/containers/storage.conf
+      - run: |
           make docs-generation
           hack/tree_status.sh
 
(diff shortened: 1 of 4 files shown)


haircommander:
I think test failures are unrelatred https://github.com/cri-o/cri-o/issues/4509

.github/workflows/test.yml
```diff
@@ -43,6 +55,7 @@ jobs:
       - run: |
           chmod -R +x bin
           sudo rm /etc/containers/storage.conf
+      - run: |
```
kolyshkin, line 58: What is the reason for this change?

kolyshkin (approved):
LGTM

kolyshkin:
So I took a look at the artifacts produced (https://github.com/cri-o/cri-o/actions/runs/501561670), and it looks like `build` and `build-static` are identical, except `build-static` has `pinns` while `build` doesn't.

Also it probably make sense to set some retention policy, and/or remove some intermediate artifacts.

.github/workflows/test.yml
saschagrunert: My intention to not name the steps was to be able to see what is going to be executed from the jobs overview page. Unfortunately it only displays the first line from multi-line executions. By adding another run step, It now looks like this: 
screenshot

Seems more clear to me what is going on.

saschagrunert:
> So I took a look at the artifacts produced (https://github.com/cri-o/cri-o/actions/runs/501561670), and it looks like `build` and `build-static` are identical, except `build-static` has `pinns` while `build` doesn't.

`build` contains the dynamically linked binaries which are mainly going to be used for further validations and later on the integration tests.

`build-static` contains the statically linked binaries, which are intended to be directly consumed by users via the static binary bundle we provide for each commit/release.

> Also it probably make sense to set some retention policy, and/or remove some intermediate artifacts.

Right now we have an artifact and log retention set to 90 days via the project configuration settings:

screenshot
Do you have anything else in mind?

PR #4513, opened 2021-01-22
Move unit tests to GitHub actions

saschagrunert:
#### What type of PR is this?

/kind ci

We now use GitHub actions for the unit test execution instead of
CircleCi. We also separate the pipeline into the three steps of mock
verification, unit test execution and coverage upload.

Refers to #4489 
None

```release-note
None
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -143,3 +143,40 @@ jobs:
         env:
           GCS_BUCKET_SA: ${{ secrets.GCS_BUCKET_SA }}
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+
+  unit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.15'
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: go-unit-${{ hashFiles('**/go.sum') }}
+          restore-keys: go-unit-
+      - run: scripts/github-actions-packages
+      - run: |
+          make mockgen -j $(nproc)
+          hack/tree_status.sh
+      - run: make testunit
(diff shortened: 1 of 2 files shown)


fidencio:
This is really good news, @saschagrunert! Thanks and LGTM!

.github/workflows/test.yml
```diff
@@ -59,3 +59,57 @@ jobs:
...
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: go-mocks-${{ hashFiles('**/go.sum') }}
+          restore-keys: go-mocks-
+      - run: scripts/github-actions-packages
+      - run: |
+          make mockgen -j6
```
kolyshkin, line 79: nit: use `-j$(nproc)`

.github/workflows/test.yml
```diff
@@ -59,3 +59,57 @@ jobs:
...
+
+  coverage:
+    needs: unit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@v2
+        with:
+          name: unit
+          path: build
```
kolyshkin, line 114: What is the reason for having a separate `make coverage` job? It seems it could just be the last step of `unit` job, which would simplify things a lot.

kolyshkin: IOW
```suggestion
```

.github/workflows/test.yml
saschagrunert: Fixed :+1:

.github/workflows/test.yml
saschagrunert: Yeah I agree, merged both steps into a single one :+1:

saschagrunert:

saschagrunert:
Completions validation seems broken on the master. This PR is not the root cause for it.

saschagrunert:
/retest

@haircommander do you think we can move forward with this one?

PR #4514, opened 2021-01-22
Support pprof profile over unix socket

mrunalp:
#### What type of PR is this?
/kind feature

This PR adds a feature to enable pprof profile over the CRI-O unix socket.
This is useful for situations where we want to keep access to pprof locked down.

None

```release-note
Support enabling pprof profile over CRI-O's unix socket
```


haircommander:
/retest

is the only way to get this behaviour copying the entire library? is there some external library we could use instead? it's quite a bit of code to carry, and we have no official mechanism to carry patches for it (go mods or the like) which makes me hesitant.

mrunalp:
@haircommander I will add a README as a follow for the process to keep this updated. I looked around and didn't see any maintained library. Also, it makes sense for us to hook this into the mux that we use directly.

kolyshkin:
> This is useful for situations where we want to keep access to pprof locked down.

What is the scenario in which pprof is exposed? In our code it is listening on 127.0.0.1:6060 which is an unroutable address and thus is not exposed to the outside world (unless there's a DNAT netfilter rule is added). In such case the remedy would be to remove that DNAT rule (or add another rule, blocking port 6060).

mrunalp:
Pushed a simplified version w/o forking library code.

PR #4523, opened 2021-01-29
[WIP][CLI] "crio config" only prints the fields that are different than the default

### LINKED ISSUE #4506: `crio config` fixes
Currently, calling `crio config` spits out an entire config file. However, it would be prudent for us to move away from the monolithic config file, in favor of smart defaults, and drop-in files.

However, to do this, there needs to be a way for users to quickly create drop-in files, and have them not override the defaults where they don't need to.

The main advantage to this approach is it allows us to change defaults without as much churn in config files.

This proposal is to change `crio config` to render the configuration/flags it's passed, and only print the toml for the fields that are different than the default.

Sample could be:
`crio --log-level debug config`
would only output
```
[crio]
[crio.runtime]
log_level = "debug"
```

/kind feature

ityuhui:
This is an early draft. I put the code here to ask reviewers to check whether I'm in the right direction for this feature implementation.

#### What type of PR is this?

> /kind feature

As described in #4506 , this PR lets ``crio config`` only prints the fields that are differet than the default

Fixes #4506

Yes.

```release-note
[CLI] "crio config" only prints the fields that are differet than the default
```

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -74,6 +74,6 @@ it later with **--config**. Global options will modify the output.`,
 		}
 
 		// Output the commented config.
-		return conf.WriteTemplate(os.Stdout)
+		return conf.WriteTemplate(c, os.Stdout)
 	},
 }
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -19,9 +19,37 @@ func (c *Config) WriteTemplate(ctx *cli.Context, w io.Writer) error {
 func assembleTemplateString(ctx *cli.Context) string {
 	templateString := templateStringPrefix
 
+	// top level configuration
 	if ctx.IsSet("root") {
 		templateString += templateString_root
 	}
+
+	if ctx.IsSet("runroot") {
+		templateString += templateString_runroot
+	}
+
+	if ctx.IsSet("storage-driver") {
+		templateString += templateString_storage_driver
+	}
+
+	if ctx.IsSet("storage-opt") {
+		templateString += templateString_storage_option
+	}
+
+	if ctx.IsSet("log-dir") {
+		templateString += templateString_log_dir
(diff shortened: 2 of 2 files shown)


ityuhui:
Hi @haircommander

Could you please check whether my draft code is in the right direction ? Thanks

ityuhui:
Append a test

```shell
root@machine:~/cri-o# ./bin/crio --root /etc/crio-conf config
INFO[0000] Starting CRI-O, version: 1.20.0-dev, git: 42acc24ca393bf77dc58692fffbf9122be660486(dirty)
INFO Using default capabilities: CAP_CHOWN, CAP_DAC_OVERRIDE, CAP_FSETID, CAP_FOWNER, CAP_SETGID, CAP_SETUID, CAP_SETPCAP, CAP_NET_BIND_SERVICE, CAP_KILL
# The CRI-O configuration file specifies all of the available configuration
# options and command-line flags for the crio(8) OCI Kubernetes Container Runtime
# daemon, but in a TOML format that can be more easily modified and versioned.
#

# CRI-O supports partial configuration reload during runtime, which can be
# done by sending SIGHUP to the running process. Currently supported options
# are explicitly mentioned with: 'This option supports live configuration
# reload'.

# CRI-O reads its storage defaults from the containers-storage.conf(5) file
# located at /etc/containers/storage.conf. Modify this storage configuration if
# you want to change the system's defaults. If you want to modify storage just
# for CRI-O, you can change the storage configuration options here.
[crio]

# Path to the "root directory". CRI-O stores all of its data, including
# containers images, in this directory.
#root = "/etc/crio-conf"

```

haircommander:
Yeah this is a great start @ityuhui :smile:

ityuhui:
Close this PR because this feature is under development.

PR #4525, opened 2021-01-29
Rename separate_pull_cgroup to image_pull_cgroup

mrunalp:
The current name is more suitable as a boolean whereas we are really expecting a value so renaming
it to image pull cgroup.

#### What type of PR is this?
/kind cleanup

Rename the option so it is clearer 

None

```release-note
Rename separate_pull_cgroup option to image_pull_cgroup
```


haircommander:
since this option exists in 1.20, we'll need to mark it as deprecated and support it for a release before dropping it

giuseppe:
> since this option exists in 1.20, we'll need to mark it as deprecated and support it for a release before dropping it

Do we still need to go through the deprecation step considering the feature is marked as EXPERIMENTAL?

saschagrunert (approved):
LGTM

haircommander:
I like the idea of this PR but it's been long enough with the old name that I think we should leave it as is. If we really want to change the flag name, we should go through the configuration deprecation process

PR #4528, opened 2021-02-01
rootless: enable resource limit when cgroup v2 controllers are delegated

AkihiroSuda:
#### What type of PR is this?

/kind feature

rootless: enable resource limit when cgroup v2 controllers are delegated.

Required by [Usernetes](https://github.com/rootless-containers/usernetes)

The patch is from https://github.com/rootless-containers/usernetes/blob/v20210201.0/src/patches/crio/0001-rootless-enable-resource-limit-when-cgroup-v2-contro.patch .

```release-note
rootless: enable resource limit when cgroup v2 controllers are delegated.
```
- - -
# Test steps

## Prepare environment
Just as in Rootless Podman, cgroup v2 delegation needs to be configured: https://rootlesscontaine.rs/getting-started/common/cgroup2/

Also, the following sysctl is needed, depending on distro:
```bash
sudo sysctl kernel.dmesg_restrict=0
```

## Install Usernetes
[Usernetes v20210201.0](https://github.com/rootless-containers/usernetes) contains CRI-O binary (`./usernetes/bin/crio`) with this PR applied.

```bash
wget https://github.com/rootless-containers/usernetes/releases/download/v20210201.0/usernetes-x86_64.tbz
tar xjvf usernetes-x86_64.tbz
cd usernetes
./install.sh --cri=crio
```

## Run a test pod
```bash
export KUBECONFIG=$HOME/.config/usernetes/master/admin-localhost.kubeconfig

kubectl apply -f - << EOF
apiVersion: v1
kind: Pod
metadata:
  name: test-limits
spec:
  containers:
  - name: test-limits
    image: alpine
    command: ["top"]
    resources:
      limits:
        cpu: 420m
        memory: 42Mi
EOF
```

A new cgroup with the resource limits is created for the pod:

```console
$ cat /proc/$(pgrep top)/cgroup
0::/user.slice/user-1001.slice/user@1001.service/u7s-rootlesskit.service/615fc903ca7ea59a4b534816b2908c29cc30d669f1f7042467ec8ea15524f74c

$ cd /sys/fs/cgroup/user.slice/user-1001.slice/user@1001.service/u7s-rootlesskit.service/615fc903ca7ea59a4b534816b2908c29cc30d669f1f7042467ec8ea15524f74c

$ cat cpu.max 
42000 100000

$ cat memory.max 
44040192
```

### DIFF
--- a/server/rootless.go
+++ b/server/rootless.go
@@ -1,10 +1,14 @@
 package server
 
 import (
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
+	"github.com/opencontainers/runc/libcontainer/cgroups"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/sirupsen/logrus"
 )
 
 func hasNetworkNamespace(config *rspec.Spec) bool {
@@ -17,7 +21,43 @@ func hasNetworkNamespace(config *rspec.Spec) bool {
 }
 
 func makeOCIConfigurationRootless(g *generate.Generator) {
-	g.Config.Linux.Resources = nil
+	// Resource limitations requires cgroup v2 delegation (https://rootlesscontaine.rs/getting-started/common/cgroup2/).
+	if r := g.Config.Linux.Resources; r != nil {
+		// cannot control device eBPF with rootless
+		r.Devices = nil
+		if r.Memory != nil || r.CPU != nil || r.Pids != nil || r.BlockIO != nil || r.Rdma != nil || r.HugepageLimits != nil {
+			v2Controllers := getAvailableV2Controllers()
+			if _, ok := v2Controllers["memory"]; !ok && r.Memory != nil {
+				logrus.Warn("rootless: cgroup v2 memory controller is not delegated. Discarding memory limit.")
+				r.Memory = nil
+			}
+			if _, ok := v2Controllers["cpu"]; !ok && r.CPU != nil {
+				logrus.Warn("rootless: cgroup v2 cpu controller is not delegated. Discarding cpu limit.")
+				r.CPU = nil
+			}
+			if _, ok := v2Controllers["cpuset"]; !ok && r.CPU != nil {
+				logrus.Warn("rootless: cgroup v2 cpuset controller is not delegated. Discarding cpuset limit.")
+				r.CPU.Cpus = ""
+				r.CPU.Mems = ""
+			}
+			if _, ok := v2Controllers["pids"]; !ok && r.Pids != nil {
+				logrus.Warn("rootless: cgroup v2 pids controller is not delegated. Discarding pids limit.")
+				r.Pids = nil
+			}
+			if _, ok := v2Controllers["io"]; !ok && r.BlockIO != nil {
+				logrus.Warn("rootless: cgroup v2 io controller is not delegated. Discarding block I/O limit.")
+				r.BlockIO = nil
+			}
+			if _, ok := v2Controllers["rdma"]; !ok && r.Rdma != nil {
+				logrus.Warn("rootless: cgroup v2 rdma controller is not delegated. Discarding RDMA limit.")
+				r.Rdma = nil
+			}
+			if _, ok := v2Controllers["hugetlb"]; !ok && r.HugepageLimits != nil {
+				logrus.Warn("rootless: cgroup v2 hugetlb controller is not delegated. Discarding RDMA limit.")
+				r.HugepageLimits = nil
+			}
+		}
+	}
 	g.Config.Process.OOMScoreAdj = nil
 	g.Config.Process.ApparmorProfile = ""
 
@@ -44,4 +84,28 @@ func makeOCIConfigurationRootless(g *generate.Generator) {
 	}
 
 	g.SetLinuxCgroupsPath("")
+}
+
+// getAvailableV2Controllers returns the entries in /sys/fs/cgroup/<SELF>/cgroup.controllers
+func getAvailableV2Controllers() map[string]struct{} {
+	procSelfCgroup, err := cgroups.ParseCgroupFile("/proc/self/cgroup")
+	if err != nil {
+		logrus.Error(err)
+		return nil
+	}
+	v2Group := procSelfCgroup[""]
+	if v2Group == "" {
+		return nil
+	}
+	controllersPath := filepath.Join("/sys/fs/cgroup", v2Group, "cgroup.controllers")
+	controllersBytes, err := ioutil.ReadFile(controllersPath)
+	if err != nil {
+		logrus.Error(err)
+		return nil
+	}
+	result := make(map[string]struct{})
+	for _, controller := range strings.Split(strings.TrimSpace(string(controllersBytes)), " ") {
+		result[controller] = struct{}{}
+	}
+	return result
 }


AkihiroSuda:
cc @giuseppe

server/rootless.go
```diff
@@ -45,3 +83,27 @@ func makeOCIConfigurationRootless(g *generate.Generator) {
 
 	g.SetLinuxCgroupsPath("")
 }
+
+// getAvailableV2Controllers returns the entries in /sys/fs/cgroup/<SELF>/cgroup.controllers
```
giuseppe, line 87: can we memoize the return value with `sync.Once`?

AkihiroSuda: I don't think so, the file content can change dynamically

giuseppe: but something that can be fixed by restarting CRI-O instead of having the cost for each container?

Anyway, not a blocker, we can deal with it later if we need to

giuseppe (approved):
LGTM

server/rootless.go
haircommander: or, we could asynchronously check if the file has changed, and update if so. that may cause a delay in successful container creation if it does change, but the normal case would not add overhead per container create

server/rootless.go
```diff
@@ -17,7 +21,41 @@ func hasNetworkNamespace(config *rspec.Spec) bool {
...
+	// Resource limitations requires cgroup v2 delegation (https://rootlesscontaine.rs/getting-started/common/cgroup2/).
+	if g.Config.Linux.Resources != nil {
+		// cannot control device eBPF with rootless
+		g.Config.Linux.Resources.Devices = nil
+		v2Controllers := getAvailableV2Controllers()
+		if _, ok := v2Controllers["memory"]; !ok && g.Config.Linux.Resources.Memory != nil {
+			logrus.Warn("rootless: cgroup v2 memory controller is not delegated. Discarding memory limit.")
+			g.Config.Linux.Resources.Memory = nil
+		}
+		if _, ok := v2Controllers["cpu"]; !ok && g.Config.Linux.Resources.CPU != nil {
```
haircommander, line 33: nit: any reason to not unconditionally set ...CPU = nil if we don't have the controller? i.e. why are we checking if the field is not nil before setting it to be nil?

AkihiroSuda: Because when CPU is already nil, no need to print the warning. (But I haven't checked whether CPU can be nil)

AkihiroSuda:
CI seems failing in building runc. Doesn't seem to be related to this PR.

```
TASK [build runc] **************************************************************
task path: /go/src/github.com/cri-o/cri-o/contrib/test/integration/build/runc.yml:9
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/bin/make 'BUILDTAGS=seccomp selinux'", "msg": "# github.com/opencontainers/runc/libcontainer/cgroups/fs\nlibcontainer/cgroups/fs/cpuset.go:89:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:94:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:99:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:104:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:109:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:114:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:119:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:124:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:129:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:134:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:134:20: too many errors\nmake: *** [runc] Error 2", "rc": 2, "stderr": "# github.com/opencontainers/runc/libcontainer/cgroups/fs\nlibcontainer/cgroups/fs/cpuset.go:89:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:94:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:99:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:104:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:109:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:114:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:119:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:124:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:129:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:134:20: undefined: errors\nlibcontainer/cgroups/fs/cpuset.go:134:20: too many errors\nmake: *** [runc] Error 2\n", "stderr_lines": ["# github.com/opencontainers/runc/libcontainer/cgroups/fs", "libcontainer/cgroups/fs/cpuset.go:89:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:94:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:99:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:104:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:109:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:114:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:119:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:124:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:129:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:134:20: undefined: errors", "libcontainer/cgroups/fs/cpuset.go:134:20: too many errors", "make: *** [runc] Error 2"], "stdout": "go build -trimpath \"-mod=vendor\" \"-buildmode=pie\"  -tags \"seccomp selinux\" -ldflags \"-X main.gitCommit=\"4074b4797531102c2210c07b562e8a7da9b7f77e\" -X main.version=1.0.0-rc92+dev \" -o runc .\n", "stdout_lines": ["go build -trimpath \"-mod=vendor\" \"-buildmode=pie\"  -tags \"seccomp selinux\" -ldflags \"-X main.gitCommit=\"4074b4797531102c2210c07b562e8a7da9b7f77e\" -X main.version=1.0.0-rc92+dev \" -o runc ."]}
	to retry, use: --limit @/go/src/github.com/cri-o/cri-o/contrib/test/integration/main.retry
```
https://prow.ci.openshift.org/view/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/4528/test_pull_request_crio_e2e_rhel/22531/

AkihiroSuda:
Is CI failure related?

giuseppe:
> Is CI failure related?

I don't think so.

Let's just try again:

/retest

AkihiroSuda:
Any chance to get this merged? :pray:

haircommander:
LGTM, @kolyshkin PTAL and merge if LGTY

AkihiroSuda:
ping @kolyshkin

server/rootless.go
```diff
@@ -45,3 +83,27 @@ func makeOCIConfigurationRootless(g *generate.Generator) {
 
 	g.SetLinuxCgroupsPath("")
 }
+
+// getAvailableV2Controllers returns the entries in /sys/fs/cgroup/<SELF>/cgroup.controllers
+func getAvailableV2Controllers() map[string]struct{} {
+	procSelfCgroup, err := cgroups.ParseCgroupFile("/proc/self/cgroup")
+	if err != nil {
+		logrus.WithError(err).Error("failed to parse /proc/self/cgroup")
```
kolyshkin, line 91: nit: I think the error message already contains file name, so wrapping is redundant. Not a big deal though.

server/rootless.go
```diff
@@ -45,3 +83,27 @@ func makeOCIConfigurationRootless(g *generate.Generator) {
...
+		return nil
+	}
+	v2Group := procSelfCgroup[""]
+	if v2Group == "" {
+		return nil
+	}
+	controllersPath := filepath.Join("/sys/fs/cgroup", v2Group, "cgroup.controllers")
+	controllersBytes, err := ioutil.ReadFile(controllersPath)
+	if err != nil {
+		logrus.WithError(err).Errorf("failed to read %s", controllersPath)
```
kolyshkin, line 101: nit: `ioutil.ReadFile` error message already contains file name, so no need to wrap an error.
```suggestion
		logrus.Error(err)
```

server/rootless.go
```diff
@@ -45,3 +83,27 @@ func makeOCIConfigurationRootless(g *generate.Generator) {
...
+		return nil
+	}
+	result := make(map[string]struct{})
+	for _, controller := range strings.Fields(string(controllersBytes)) {
```
kolyshkin, line 105: I'd use `strings.Split` here as
 - it is faster;
 - the data returned from the kernel do not contain any extra spaces.

server/rootless.go
```diff
@@ -17,7 +21,41 @@ func hasNetworkNamespace(config *rspec.Spec) bool {
 }
 
 func makeOCIConfigurationRootless(g *generate.Generator) {
-	g.Config.Linux.Resources = nil
+	// Resource limitations requires cgroup v2 delegation (https://rootlesscontaine.rs/getting-started/common/cgroup2/).
+	if g.Config.Linux.Resources != nil {
```
kolyshkin, line 25: I would add something like
```go
r := g.Config.Linux.Resources
```
and use it below, for the sake of readability.

server/rootless.go
```diff
@@ -17,7 +21,41 @@ func hasNetworkNamespace(config *rspec.Spec) bool {
...
+		// cannot control device eBPF with rootless
+		g.Config.Linux.Resources.Devices = nil
```
kolyshkin, line 27: Do we need a warning here (like you do in the other cases below)?

server/rootless.go
```diff
@@ -17,7 +21,41 @@ func hasNetworkNamespace(config *rspec.Spec) bool {
...
+		v2Controllers := getAvailableV2Controllers()
```
kolyshkin, line 28: I'd do something like
```go
if r.Memory != nil || r.CPU != nil || r.Pids != nil || r.BlockIO != nil || r.Rdma != nil || r.HugepageLimits != nil {
   v2Controllers := getAvailableV2Controllers()
   .....
}
```

so we don't call getAvailableV2Controllers() unnecessarily.

server/rootless.go
```diff
@@ -17,7 +21,41 @@ func hasNetworkNamespace(config *rspec.Spec) bool {
...
+		if _, ok := v2Controllers["memory"]; !ok && g.Config.Linux.Resources.Memory != nil {
```
kolyshkin, line 29: Makes sense to reverse the order of checks:
```suggestion
		if r.Memory != nil && _, ok := v2Controllers["memory"]; !ok {
```
as the `!= nil` check is faster than a map lookup.

server/rootless.go
kolyshkin: This is not a file but two (`/proc/self/cgroup` and `/sys/fs/cgroup/<SELF>/cgroup.controllers`), and these are not usual files, but rather an API to the kernel. For some files, it is possible to register an event handler to be notified by the kernel if the file is being changed, but I am not sure if it's possible for these two.

Even if it is, I don't think it makes sense to do so (add a watcher, update the contents in async manner), as there is not too much contents to read/parse, and it would be a gross complication with little performance benefit (shave off a few ms on a rootless container start).

server/rootless.go
AkihiroSuda: updated

server/rootless.go
AkihiroSuda: updated

server/rootless.go
AkihiroSuda: updated

server/rootless.go
AkihiroSuda: updated

server/rootless.go
AkihiroSuda: updated

server/rootless.go
AkihiroSuda: No, because there is no action user can take.

server/rootless.go
AkihiroSuda: That syntax is uncompilable, and benefit seems negligible

AkihiroSuda:
Addressed comments

giuseppe (approved):
LGTM

AkihiroSuda:
@giuseppe @kolyshkin @rhatdan Can this be merged? :pray:

haircommander:
/lgtm

thanks @AkihiroSuda for sticking through with this

PR #4530, opened 2021-02-02
ResourceStore: fix segfault and update tests

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

There is a chance that a WatcherForResource() followed by a Get() would cause cri-o to segfault. Fix this
also extend the test suite to add a test for this and another situation, as well as move around the tests so they're all run

```release-note
fixed a bug that could cause CRI-O to segfault when a node is under heavy load
```

### DIFF
--- a/internal/resourcestore/resourcestore.go
+++ b/internal/resourcestore/resourcestore.go
@@ -34,6 +34,12 @@ type Resource struct {
 	name         string
 }
 
+// wasPut checks that a resource has been fully defined yet.
+// This is defined as a resource that only has watchers, but no associated resource.
+func (r *Resource) wasPut() bool {
+	return r != nil && r.resource != nil
+}
+
 // IdentifiableCreatable are the qualities needed by the caller of the resource.
 // Once a resource is retrieved, SetCreated() will be called, indicating to the server
 // that resource is ready to be listed and operated upon, and ID() will be used to identify the
@@ -101,6 +107,11 @@ func (rc *ResourceStore) Get(name string) string {
 	if !ok {
 		return ""
 	}
+	// It is possible there are existing watchers,
+	// but no resource created yet
+	if !r.wasPut() {
+		return ""
+	}
 	delete(rc.resources, name)
 	r.resource.SetCreated()
 	return r.resource.ID()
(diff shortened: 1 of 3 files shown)


internal/resourcestore/resourcestore.go
```diff
@@ -121,7 +132,7 @@ func (rc *ResourceStore) Put(name string, resource IdentifiableCreatable, cleanu
 		rc.resources[name] = r
 	}
 	// make sure the resource hasn't already been added to the store
-	if r.resource != nil || r.cleanupFuncs != nil {
+	if r.wasPut() {
```
kolyshkin, line 135: Looks like this code would benefit from
```suggestion
	if ok && r.wasPut() {
```

Or something similar (as there's no need to check wasPut if we just created a new empty resource.

kolyshkin:
Nit: looks like `WatcherForResource` doc needs fixing, too. Currently it says

```go
// WatcherForResource looks up a Resource by name, and gives it a watcher if it's found.
```

but in fact it gives a resource a watcher even if the lookup result is negative (by creating an uninitialized resource).

haircommander:
> Nit: looks like `WatcherForResource` doc needs fixing, too. Currently it says
> 
> ```go
> // WatcherForResource looks up a Resource by name, and gives it a watcher if it's found.
> ```
> 
> but in fact it gives a resource a watcher even if the lookup result is negative (by creating an uninitialized resource).

fixed!

kolyshkin:
I would also reorder commits to put the fix before the test -- otherwise if someone uses git-bisect, they find that the test commit breaks things.

Overall LGTM except a few nits above.

kolyshkin (approved):
LGTM

QiWang19 (commented):
LGTM

saschagrunert (approved):
LGTM

internal/resourcestore/resourcestore.go
```diff
@@ -53,12 +60,17 @@ func New() *ResourceStore {
...
+		closeChan: make(chan struct{}, 1),
 		timeout:   timeout,
 	}
 	go rc.cleanupStaleResources()
 	return rc
 }
 
+func (rc *ResourceStore) Close() {
+	rc.closeChan <- struct{}{}
+}
```
giuseppe, line 72: should it also do `close(rc.closeChan)`?

haircommander: good point, it should actually only close methinks, as the receiever will still get a value out of it and know it's time to return

fidencio:
LGTM, and it does address all the comments raised, based on that ...
/lgtm

PR #4540, opened 2021-02-04
[1.19] Set systemd property KillMode to mixed for containers

mrunalp:

#### What type of PR is this?

/kind bug

This PR sets the systemd KillMode for container scopes to mixed
so on stop, only the main process gets the SIGTERM and is 
responsible for forwarding it if needed to child processes.

Without this, we end up with double SIGTERMS leading to unexpected
failures in graceful termination on node shutdown or reboot.

https://bugzilla.redhat.com/show_bug.cgi?id=1882750

```release-note
Set systemd KillMode to mixed for container scopes modifying behavior on node shutdown
```


TomSweeneyRedHat:
LGTM
but tests are not at all happy

mrunalp:
Closing for now - https://bugzilla.redhat.com/show_bug.cgi?id=1925623

PR #4542, opened 2021-02-04
hostport manager clean up host ports

aojea:
The host port manager, when Remove the hostport from the system, exists fast if there are no
iptables rules to delete.
However, it was missing to call the method that remove the hostports bound, so there is a possibility that we leak sockets.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

Thanks to @johnbelamaric 
https://github.com/kubernetes/kubernetes/pull/98755#discussion_r570413563

> /kind bug
> /kind cleanup
```release-note
NONE
```

### DIFF
--- a/internal/hostport/hostport_manager.go
+++ b/internal/hostport/hostport_manager.go
@@ -238,8 +238,9 @@ func (hm *hostportManager) Remove(id string, podPortMapping *PodPortMapping) (er
 	}
 
 	// exit if there is nothing to remove
+	// don´t forget to clean up opened pod host ports
 	if len(existingChainsToRemove) == 0 {
-		return nil
+		return hm.closeHostports(hostportMappings)
 	}
 
 	natChains := bytes.NewBuffer(nil)


aojea:
/assign @danwinship 
/hold

until the upstream review finish https://github.com/kubernetes/kubernetes/pull/98755

It is a bit ironic, but since Conformance broke due to the bug in hostport, now the fix may get upstream.
However, the removal of dockershim is still scheduled for 1.21, so it may be this code will be backported only to 1.20 and disappears in 1.21 ... who knows :shrug:

saschagrunert (approved):
LGTM

aojea:
/hold cancel
kubernetes patch was approved

saschagrunert (approved):
/lgtm
/retest

internal/hostport/hostport_manager.go
```diff
@@ -238,8 +238,9 @@ func (hm *hostportManager) Remove(id string, podPortMapping *PodPortMapping) (er
 	}
 
 	// exit if there is nothing to remove
+	// don´t forget to clean up opened pod host ports
 	if len(existingChainsToRemove) == 0 {
-		return nil
+		return hm.closeHostports(hostportMappings)
 	}
```
TomSweeneyRedHat, line 245: There's a `return(err)` at line https://github.com/cri-o/cri-o/pull/4542/files#diff-1e8c8253b73c319cc3d39e1e6eebdbff23129076dfe7814b74754f640f68b1c5R261 and I think we can leave behind the hostportmappings there too.  Would it be better to create a defer function to close them there and to do the unlock too?

aojea: If there is an error programming the iptables rules, it will be possible that the iptables rules for the portmaps are still active so we should keep holding the ports and not close them.

PR #4544, opened 2021-02-04
Allow seccomp hook tracing for separate containers

saschagrunert:
#### What type of PR is this?

/kind feature

This change allows us to trace different containers into dedicated
files. For example the workload:

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  selector:
    matchLabels:
      app: test
  replicas: 1
  template:
    metadata:
      labels:
        app: test
      annotations:
        io.containers.trace-syscall/nginx: of:/tmp/nginx.json
        io.containers.trace-syscall/redis: of:/tmp/redis.json
    spec:
      containers:
        - name: nginx
          image: nginx:latest
        - name: redis
          image: redis:latest
```

Will now create two separate files containing only the syscalls from
each workload.

Fixes https://github.com/containers/oci-seccomp-bpf-hook/issues/74
None

```release-note
Added feature to allow the OCI seccomp BPF hook to trace containers into dedicated files by
using the annotation key `io.containers.trace-syscall/container-name`
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cri-o/cri-o/internal/config/device"
 	"github.com/cri-o/cri-o/internal/lib"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	"github.com/cri-o/cri-o/internal/log"
 	oci "github.com/cri-o/cri-o/internal/oci"
 	"github.com/cri-o/cri-o/internal/storage"
 	crioann "github.com/cri-o/cri-o/pkg/annotations"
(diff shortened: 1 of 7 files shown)


pkg/container/container.go
```diff
@@ -141,6 +142,28 @@ func (c *container) SpecAddAnnotations(sb *sandbox.Sandbox, containerVolumes []o
...
+		// attach a BFP module to the process. The recorded syscalls will be
+		// then stored in the output path file (annotation value prefixed with
+		// 'of:'). We now add a custom logic to be able to distinguish
+		// containers within pods in Kubernetes. If we suffix the container
+		// name within the annotation key like this:
+		// io.containers.trace-syscall/container
+		// Then we will rewrite the key to 'io.containers.trace-syscall' if the
+		// metadata name is equal to 'container'. This allows us to trace
+		// containers into distinguishable files.
+		const traceHookPrefix = "io.containers.trace-syscall"
```
haircommander, line 157: maybe this is a good candidate for AllowedAnnotations?

JAORMX (approved):

pkg/container/container.go
```diff
@@ -141,6 +142,28 @@ func (c *container) SpecAddAnnotations(sb *sandbox.Sandbox, containerVolumes []o
...
+		if strings.TrimPrefix(k, traceHookPrefix+"/") == c.config.Metadata.Name {
+			log.Debugf(c.ctx,
+				"Annotation key for container '%s' rewritten to '%s' (value is: '%s')",
```
vrothberg, line 160:
```suggestion
				"Annotation key for container %q rewritten to %q (value is: %q)",
```

vrothberg (commented):
:+1: cool idea!  Looking forward to seeing it in action.

pkg/container/container.go
saschagrunert: Yes, the annotation `io.containers.trace-syscall` now has to be part of the allowed annotations, otherwise they will get filtered-out at the sandbox level. PTAL :)

saschagrunert:
>  cool idea! Looking forward to seeing it in action.

I'm planning to demo it at containers plumbing days if my talk gets accepted :crossed_fingers: :grimacing:

pkg/container/container.go
saschagrunert: Thanks, fixed as suggested. :+1:

pkg/container/container.go
```diff
@@ -140,8 +142,35 @@ func (c *container) SpecAddAnnotations(sb *sandbox.Sandbox, containerVolumes []o
...
 	// Preserve the sandbox annotations. OCI hooks may re-use the sandbox
 	// annotation values to apply them to the container later on.
+	// The sandbox annotations are already filtered for the allowed
+	// annotations, there is no need to check it additionally here.
 	for k, v := range sb.Annotations() {
-		c.spec.AddAnnotation(k, v)
+		if strings.HasPrefix(k, pkgAnnotations.OCISeccompBPFHookAnnotation) {
+			// The OCI seccomp BPF hook
+			// (https://github.com/containers/oci-seccomp-bpf-hook)
+			// uses the annotation io.containers.trace-syscall as indicator
```
TomSweeneyRedHat, line 151: nit of a nit, only change if you've other changes
```suggestion
			// uses the annotation io.containers.trace-syscall as an indicator
```

TomSweeneyRedHat:
LGTM
only one tiny in code comment nit, don't worry about it unless you have to make other adjustments.

pkg/container/container.go
saschagrunert: Fixed as suggested, had to rebase.

saschagrunert:
@haircommander do you think we can get this one in?

PR #4546, opened 2021-02-04
[1.20] Set conmon scope KillSignal to SIGPIPE

mrunalp:
#### What type of PR is this?
 /kind bug

conmon forwards the SIGTERM to the main container process.
This is undesirable during node shutdown as it results in
the container receving 2 SIGTERMs. With this change, conmon
ignores the SIGPIPE and will exit after the main container
process dies. If it doesn't for some reason, then
it gets SIGKILL by systemd after 90 seconds.

We need a follow-on to adjust the 90s to a value higher
than the termination grace period as a follow on. This doesn't
matter today as we don't retain container state on reboot.

```release-note
Set conmon scope KillSignal to SIGPIPE
```

### DIFF
--- a/internal/config/cgmgr/systemd.go
+++ b/internal/config/cgmgr/systemd.go
@@ -8,10 +8,13 @@ import (
 	"path/filepath"
 	"strings"
 
+	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
 	"github.com/cri-o/cri-o/utils"
+	"github.com/godbus/dbus/v5"
 	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 const defaultSystemdParent = "system.slice"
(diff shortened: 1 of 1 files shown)


internal/config/cgmgr/systemd.go
```diff
@@ -70,8 +72,16 @@ func (*SystemdManager) MoveConmonToCgroup(cid, cgroupParent, conmonCgroup string
 		cgroupParent = conmonCgroup
 	}
 	conmonUnitName := fmt.Sprintf("crio-conmon-%s.scope", cid)
+
+	// Set the systemd KillSignal to SIGPIPE that conmon ignores.
+	// This helps during node shutdown so that conmon waits for the container
+	// to exit and doesn't forward the SIGTERM that it gets.
+	killSignalProp := systemdDbus.Property{
+		Name:  "KillSignal",
+		Value: dbus.MakeVariant(13),
```
kolyshkin, line 81: Maybe
 - use `int(unix.SIGPIPE)` as otherwise it's a non-descriptive magic number
 - add a comment explaining that SIGPIPE is ignored, and it is set here so conmon can ignore termination signal from systemd.

kolyshkin (changes requested):

internal/config/cgmgr/systemd.go
kolyshkin: (second issue fixed while I was writing my review comment, so please ignore)

mrunalp: Switched to using `int(unix.SIGPIPE)`. Thanks!

kolyshkin (approved):
LGTM

PR #4548, opened 2021-02-05
[1.20] oci: don't wait too long on a long stop

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

oci: don't wait too long on a long stop

if a long stop timeout comes in, and a shorter one comes in after
we never interrupt the old one. This can cause the new container stop to be DOSed

instead, interrupt the old stop, so the new one can proceed as expected

```release-note
none
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -72,6 +72,10 @@ type Container struct {
 	stdinOnce          bool
 	created            bool
 	spoofed            bool
+	stopping           bool
+	stopTimeoutChan    chan time.Duration
+	stoppedChan        chan struct{}
+	stopLock           sync.Mutex
 }
 
 // Metadata holds all necessary information for building the container name.
@@ -132,6 +136,8 @@ func NewContainer(id, name, bundlePath, logPath string, labels, crioAnnotations,
 		dir:             dir,
 		state:           state,
 		stopSignal:      stopSignal,
+		stopTimeoutChan: make(chan time.Duration, 1),
+		stoppedChan:     make(chan struct{}, 1),
 	}
 	return c, nil
 }
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -616,35 +616,89 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
 			}
 		}
 	}()
-	select {
-	case <-done:
-		return nil
-	case <-ctx.Done():
-		close(chControl)
-		return ctx.Err()
-	case <-time.After(timeout):
-		close(chControl)
-		if ignoreKill {
-			return fmt.Errorf("timeout reached after %.0f seconds waiting for container process to exit",
-				timeout.Seconds())
-		}
-		pid, err := c.pid()
-		if err != nil {
-			return err
-		}
-		if err := kill(pid); err != nil {
-			return fmt.Errorf("failed to kill process: %v", err)
+	// Operate in terms of targetTime, so that we can pause in the middle of the operation
+	// to catch a new timeout (and possibly ignore that new timeout if it's not correct to
+	// take a new one).
+	targetTime := time.Now().Add(timeout)
+	for {
+		select {
+		case <-done:
+			return nil
+		case <-ctx.Done():
+			close(chControl)
+			return ctx.Err()
+		case <-time.After(time.Until(targetTime)):
+			close(chControl)
+			if ignoreKill {
+				return fmt.Errorf("timeout reached after %.0f seconds waiting for container process to exit",
+					timeout.Seconds())
+			}
+			pid, err := c.pid()
+			if err != nil {
+				return err
+			}
+			if err := kill(pid); err != nil {
+				return fmt.Errorf("failed to kill process: %v", err)
+			}
+		case newTimeout := <-c.stopTimeoutChan:
+			// If we're already at the "killing" step,
+			// we should use the killingContainerTimeout,
+			// and not override.
+			if !ignoreKill {
+				continue
+			}
+			// If a new timeout comes in,
+			// interrupt the old one, and start a new one
+			newTargetTime := time.Now().Add(newTimeout)
+
+			targetTime = newTargetTime
+
+			continue
 		}
 	}
-
 	c.state.Finished = time.Now()
+	// Successfully stopped! This is to prevent other routines from
+	// racing with this one and waiting forever.
+	// close only the dedicated channel. If we close stopTimeoutChan,
+	// any other waiting goroutine will panic, not gracefully exit.
+	close(c.stoppedChan)
 	return nil
 }
 
 // StopContainer stops a container. Timeout is given in seconds.
-func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout int64) error {
+func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout int64) (retErr error) {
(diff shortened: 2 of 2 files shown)


mrunalp:
@kir @giuseppe ptal

internal/oci/runtime_oci.go
```diff
@@ -635,17 +635,46 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
 		if err := kill(pid); err != nil {
 			return fmt.Errorf("failed to kill process: %v", err)
 		}
+	case newTimeout := <-c.stopTimeoutChan:
+		// if a new timeout comes in,
+		// interrupt the old one, and start a new one
+		close(chControl)
+		return waitContainerStop(ctx, c, newTimeout, ignoreKill)
```
kolyshkin, line 642: I don't like recursion in here -- there might be a possibility to abuse it, leading to stack overflow. Maybe better to use `goto`.

internal/oci/container.go
```diff
@@ -72,6 +72,9 @@ type Container struct {
 	stdinOnce          bool
 	created            bool
 	spoofed            bool
+	stopping           bool
+	stopTimeoutChan    chan time.Duration
```
kolyshkin, line 76: I think we might get rid of `stopping` field by initializing `stopTimeoutChan` in `waitContainerStop`, and checking if it's nil in `StopContainer` (i.e. the idea is non-nil `stopTimeoutChan` means we're stopping). The only thing is to make sure to close stopTimeoutChan and set it to nil when exiting `waitContainerStop`.

Seems that `stopLock` is still required though.

kolyshkin: I'm not sure this will result in better code though -- just an idea, feel free to ignore.

internal/oci/runtime_oci.go
```diff
@@ -646,35 +646,34 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
...
-			c.stopping = false
-			c.stopLock.Unlock()
-		}
+		// regardless of whether we succeeded or not
+		// we need to decrement the stopping counter.
+		// This means we'll know for sure whether a stop is happening.
+		// If there is no stop happening (and all others failed),
+		// we'll need to start a new one.
+		c.stopLock.Lock()
+		defer c.stopLock.Lock()
```
kolyshkin, line 670: I'm against using defer here -- the flow is pretty simple to just have
```go
		c.stopLock.Lock()
￼		c.timesStopped--
￼		c.stopLock.Unlock()
```

internal/oci/container.go
```diff
@@ -72,7 +72,7 @@ type Container struct {
 	stdinOnce          bool
 	created            bool
 	spoofed            bool
-	stopping           bool
+	timesStopped       int
```
kolyshkin, line 75: I don't understand the reason to have a refcounter here, can you please explain?

kolyshkin:
Overall I like the idea of sending a new timeout via a channel -- looks quite elegant and neat.

Left some minor comments.

internal/oci/runtime_oci.go
```diff
@@ -635,14 +635,42 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
 		if err := kill(pid); err != nil {
 			return fmt.Errorf("failed to kill process: %v", err)
 		}
+	case newTimeout := <-c.stopTimeoutChan:
```
mrunalp, line 638: If we are in the kill phase then we can ignore the update.

internal/oci/runtime_oci.go
```diff
@@ -635,14 +635,42 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
...
 // StopContainer stops a container. Timeout is given in seconds.
 func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout int64) error {
+	// first, need to check if the container is already stopping
+	c.stopLock.Lock()
+	if c.timesStopped > 0 {
+		// If so, we shouldn't wait forever on the opLock.
+		// This can cause issues where the container stop gets DOSed by a very long
+		// timeout, followed by someone sending a very short one.
+		// Instead, interrupt the other stop with this new one.
+		c.stopTimeoutChan <- time.Duration(timeout) * time.Second
```
mrunalp, line 657: Isn't there a race if `waitContainerStop` is done with the select loop? This write would then block.

internal/oci/runtime_oci.go
haircommander: ugh I really don't love using goto, but a for loop feels excessive for one case. I resign to your suggestion :)

internal/oci/runtime_oci.go
haircommander: who would have thought, such a simple solution is not that simple after all...

internal/oci/container.go
haircommander: we can't close stopTimeoutChan upon leaving waitContainerStop, as we race against other routines putting a value in (risking inserting to a closed chan and panicing). I can't think of a way to avoid the bool :\

internal/oci/runtime_oci.go
haircommander: added!

internal/oci/runtime_oci.go
haircommander: okay I think I fixed this race by introducing a new channel.

haircommander:
okay, I have a version that is (sadly) less elegant and (hopefully) better handles races.

The last piece is we need to figure out when we are able to call close(c.stopTimeoutChan)
running https://play.golang.org/p/q-r_Fl8ZYPE indicates that no matter when we close the first channel, it's possible we panic if we read from the closed second

maybe DeleteContainer? though that will clutter up a very clean function..

haircommander:
PTAL @mrunalp @kolyshkin

mrunalp:
@haircommander While thinking more about this, I think we should model this using the observer pattern. The first Stop call starts a goroutine to kill the pod and registers to be notified once that is done. Any subsequent call that comes in updates the definition of the work via a channel and then adds itself to the notification list.

mrunalp:
@haircommander The latest impl is close to what I was thinking :)

internal/oci/runtime_oci.go
```diff
@@ -616,36 +616,88 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
...
+	// first, need to check if the container is already stopping
+	c.stopLock.Lock()
+	if c.stopping {
+		// If so, we shouldn't wait forever on the opLock.
+		// This can cause issues where the container stop gets DOSed by a very long
+		// timeout, followed by someone sending a very short one.
+		// Instead, interrupt the other stop with this new one.
+		select {
+		case c.stopTimeoutChan <- time.Duration(timeout) * time.Second:
+		case <-c.stoppedChan: // This case is to prevent from waiting forever once another routine has finished.
```
mrunalp, line 679: Can't we exit when we hit this case?

haircommander: good point, added

kolyshkin (approved):
LGTM

haircommander:
/hold

for a main branch version of this, as well as *hopefully* unit tests

haircommander:
1.20 is EOL

PR #4550, opened 2021-02-08
update pause image to 3.5（3.4.1 for multi-arch support and 3.5 for non-root）

pacoxu:
#### What type of PR is this?
/kind dependency-change

bumps pause image to 3.5
Per https://github.com/kubernetes/kubernetes/pull/100292

None

part of https://github.com/kubernetes/kubernetes/pull/98205 (changelogs there)

k8s.gcr.io/pause:3.4.1 is for both windows/linux. Details in kubernetes/kubernetes/#98205

- [x] https://github.com/containers/common/blob/master/pkg/config/default.go#L48 wait for vendor update?
- [x] after https://github.com/kubernetes/kubernetes/pull/98205 merged

```release-note
Default pause image is `k8s.gcr.io/pause:3.5`
```

### DIFF
--- a/internal/config/migrate/from_1_17.go
+++ b/internal/config/migrate/from_1_17.go
@@ -60,11 +60,10 @@ func migrateFrom1_17(cfg *config.Config) error {
 	}
 
 	// Upgrade pause image
-	// https://github.com/cri-o/cri-o/pull/3582
-	logrus.Infof("Checking for pause_image, which now should be k8s.gcr.io/pause:3.2 instead of k8s.gcr.io/pause:3.1")
-	newPauseImage := "k8s.gcr.io/pause:3.2"
-	if cfg.PauseImage == "k8s.gcr.io/pause:3.1" {
-		cfg.PauseImage = newPauseImage
+	// https://github.com/cri-o/cri-o/pull/4550
+	logrus.Infof("Checking for pause_image, which now should be %s instead of k8s.gcr.io/pause:3.1 or 3.2", config.DefaultPauseImage)
+	if cfg.PauseImage == "k8s.gcr.io/pause:3.1" || cfg.PauseImage == "k8s.gcr.io/pause:3.2" {
+		cfg.PauseImage = config.DefaultPauseImage
 		logrus.Infof(`Changing "pause_image" to %s`, cfg.PauseImage)
 	}
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -94,6 +94,8 @@ const (
 	ImageVolumesIgnore ImageVolumesType = "ignore"
 	// ImageVolumesBind option is for using bind mounted volumes
 	ImageVolumesBind ImageVolumesType = "bind"
+	// DefaultPauseImage is default pause image
+	DefaultPauseImage string = "k8s.gcr.io/pause:3.5"
 )
 
 const (
@@ -622,7 +624,7 @@ func DefaultConfig() (*Config, error) {
 		},
 		ImageConfig: ImageConfig{
 			DefaultTransport: "docker://",
-			PauseImage:       "k8s.gcr.io/pause:3.2",
+			PauseImage:       DefaultPauseImage,
 			PauseCommand:     "/pause",
 			ImageVolumes:     ImageVolumesMkdir,
 		},
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -256,7 +256,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--pause-command**="": Path to the pause executable in the pause image (default: /pause)
 
-**--pause-image**="": Image which contains the pause executable (default: k8s.gcr.io/pause:3.2)
+**--pause-image**="": Image which contains the pause executable (default: k8s.gcr.io/pause:3.5)
 
 **--pause-image-auth-file**="": Path to a config file containing credentials for --pause-image (default: "")
 
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -275,7 +275,7 @@ CRI-O reads its configured registries defaults from the system wide containers-r
 **global_auth_file**=""
   The path to a file like /var/lib/kubelet/config.json holding credentials necessary for pulling images from secure registries.
 
-**pause_image**="k8s.gcr.io/pause:3.2"
+**pause_image**="k8s.gcr.io/pause:3.5"
   The image used to instantiate infra containers. This option supports live configuration reload.
 
 **pause_image_auth_file**=""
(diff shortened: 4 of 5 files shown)


TBBle:
I'm assuming the vendor/ code change was just a global search-replace, but you probably need to revert that until https://github.com/kubernetes/kubernetes/pull/98205 lands so that `verify/vendor` passes.

vendor/github.com/containers/common/pkg/config/containers.conf
```diff
@@ -313,7 +313,7 @@
 # associated with the  pod.  This container does nothing other then sleep,
 # reserving the pods resources for the lifetime of the pod.
 #
-# infra_image = "k8s.gcr.io/pause:3.2"
+# infra_image = "k8s.gcr.io/pause:3.4.1"
```
TBBle, line 316: I'm assuming the vendor/ code change was just a global search-replace, but you probably need to revert changes in that tree until https://github.com/kubernetes/kubernetes/pull/98205 lands so that `verify/vendor` passes.

pacoxu: Thanks, @TBBle.

We should update dependencies in  https://github.com/containers/common/blob/master/pkg/config/default.go#L48 along with it? I updated the vendor package and will update it later.

pkg/config/config.go
```diff
@@ -622,7 +622,7 @@ func DefaultConfig() (*Config, error) {
 		},
 		ImageConfig: ImageConfig{
 			DefaultTransport: "docker://",
-			PauseImage:       "k8s.gcr.io/pause:3.2",
+			PauseImage:       "k8s.gcr.io/pause:3.4.1",
```
saschagrunert, line 625: We should make this a public const somewhere and re-use it in `from_1_17.go`.

vendor/github.com/containers/common/pkg/config/containers.conf
TBBle: Ah, yes. Sorry, I don't know why I thought that was vendored from k8s.io. -_- The PR for containers/common makes sense, but as noted there, that config isn't used by this project, so it wouldn't be a blocker for this project.

pkg/config/config.go
pacoxu: thanks for review. @saschagrunert  updated.

vendor/github.com/containers/common/pkg/config/containers.conf
pacoxu: Just do a quick fix. I missed many things.  

Thanks for helpful comments.

internal/config/migrate/from_1_17.go
```diff
@@ -60,11 +60,10 @@ func migrateFrom1_17(cfg *config.Config) error {
...
 	// Upgrade pause image
-	// https://github.com/cri-o/cri-o/pull/3582
-	logrus.Infof("Checking for pause_image, which now should be k8s.gcr.io/pause:3.2 instead of k8s.gcr.io/pause:3.1")
-	newPauseImage := "k8s.gcr.io/pause:3.2"
+	// https://github.com/cri-o/cri-o/pull/4550
+	logrus.Infof("Checking for pause_image, which now should be %s instead of k8s.gcr.io/pause:3.1", config.DefaultPauseImage)
 	if cfg.PauseImage == "k8s.gcr.io/pause:3.1" {
-		cfg.PauseImage = newPauseImage
+		cfg.PauseImage = config.DefaultPauseImage
 		logrus.Infof(`Changing "pause_image" to %s`, cfg.PauseImage)
```
TBBle, line 67: Does this imply there should be a a migrator from 1.18, 1.19, and 1.20 too? It seems like someone migrating from 1.17 (or earlier...) would end up with the new pause image, but migrating from 1.18, 1.19, and 1.20 to cri-o 1.21 would still have the older pause image in their config.

That said, poking around, #3929 also changed a default config, but didn't implement a migrator... So perhaps this isn't a thing anymore? cri-o 1.17 is out-of-maintenance now.

pacoxu: - 1.17 is using k8s.gcr.io/pause:3.1
- 1.18~1.20 is using k8s.gcr.io/pause:3.2 
If so, I need to add some logic for it. 

Sorry for some misunderstanding here as this is my first PR in cri-o.

pacoxu:
pause image in. github.com/containers/common  is not used in this project.(wait next release of containers/common)
We may update it later or in this PR. Either is OK.

saschagrunert (approved):
/ok-to-test
LGTM

TomSweeneyRedHat:
LGTM

haircommander:
I think this patch will fix the tests:
```
diff --git a/test/helpers.bash b/test/helpers.bash
index cf3aa199e..b7ad3b449 100644
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -84,7 +84,7 @@ POD_IPV6_CIDR_START="1100:2"
 POD_IPV6_DEF_ROUTE="1100:200::1/24"
 
 IMAGES=(
-    k8s.gcr.io/pause:3.2
+    k8s.gcr.io/pause:3.5
     quay.io/crio/busybox:latest
     quay.io/crio/fedora-ping:latest
     quay.io/crio/image-volume-test:latest
```

pacoxu:
@haircommander  Updated
Thanks.

internal/config/migrate/from_1_17.go
```diff
@@ -60,11 +60,10 @@ func migrateFrom1_17(cfg *config.Config) error {
...
 
 	// Upgrade pause image
-	// https://github.com/cri-o/cri-o/pull/3582
-	logrus.Infof("Checking for pause_image, which now should be k8s.gcr.io/pause:3.2 instead of k8s.gcr.io/pause:3.1")
-	newPauseImage := "k8s.gcr.io/pause:3.2"
+	// https://github.com/cri-o/cri-o/pull/4550
+	logrus.Infof("Checking for pause_image, which now should be %s instead of k8s.gcr.io/pause:3.1", config.DefaultPauseImage)
 	if cfg.PauseImage == "k8s.gcr.io/pause:3.1" {
-		cfg.PauseImage = newPauseImage
+		cfg.PauseImage = config.DefaultPauseImage
```
TomSweeneyRedHat, line 66: The man pages anyway said 3.2 prior.  Do we need to also check from 3.2 here?

pacoxu: Check both 3.1 and 3.2?
Kubernetes 1.17 by default uses 3.1.

pacoxu:
Updated @TomSweeneyRedHat

saschagrunert (approved):
/lgtm

pacoxu:
The cgroup v2 and  e2e-gcp failed for other reasons.
/skip
?

Issue #4551, opened 2021-02-08
Support for arm 64bit static binary bundles.


saschagrunert:
The main idea is to support a second static binary bundle which is able to run on 64bit arm machines.

#### Things to be done:
- [x] Create a nix derivation which produces the `crio*` binaries for arm (https://github.com/cri-o/cri-o/pull/4552)
- [x] Update the bundle build process to create two separate instead of just one
- [x] Check if all other binaries within the bundle are supported for arm, too

saschagrunert:
This is now done
/close

PR #4552, opened 2021-02-08
Add arm64 static build to GitHub actions and bundle

saschagrunert:
#### What type of PR is this?

/kind feature

This adds a new derivation to build statically linked binaries for
arm64.

Refers to https://github.com/cri-o/cri-o/issues/4551
None

```release-note
Added arm64 variant to static binary bundle. The bundle is now available via the URL
`https://storage.googleapis.com/k8s-conform-cri-o/artifacts/cri-o.$ARCH.$VERSION.tar.gz`,
where `$ARCH` can be `amd64` or `arm64` and `$VERSION` a tag or full length git SHA.
Beside that, we provide a new installation script
`curl https://raw.githubusercontent.com/cri-o/cri-o/master/scripts/get | bash`.
For this script, it is also possible to pass the architecture and version as arguments:
`get [amd64|arm64] [VERSION|SHA]`.
The static binary bundle now uses crun as default container runtime for CRI-O.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,18 +98,43 @@ up a PR and add it to the list.
 ## Getting started
 
 ### Installing CRI-O
+
 To install `CRI-O`, you can follow our [installation guide](install.md).
 Alternatively, if you'd rather build `CRI-O` from source, checkout our [setup
 guide](install.md#build-and-install-cri-o-from-source).
 We also provide a way in building [static binaries of `CRI-O`](install.md#static-builds) via nix.
 Those binaries are available for every successfully built commit on our [Google Cloud Storage Bucket][bucket].
-This means that the latest commit can be downloaded via:
+This means that the latest commit can be installed via our convinience script:
 
 [bucket]: https://console.cloud.google.com/storage/browser/k8s-conform-cri-o/artifacts
 
 ```shell
-> curl -f https://storage.googleapis.com/k8s-conform-cri-o/artifacts/crio-$(git ls-remote https://github.com/cri-o/cri-o master | cut -c1-9).tar.gz -o crio.tar.gz
+> curl https://raw.githubusercontent.com/cri-o/cri-o/master/scripts/get | bash
 ```
+
+Beside `amd64` we also support the `arm64` bit architecture. This can be
+selected via the script, too:
+
+```shell
+> curl https://raw.githubusercontent.com/cri-o/cri-o/master/scripts/get | bash -s -- -a arm64
+```
+
+It is also possible to select a specific git SHA or tag by:
+
+```shell
+> curl https://raw.githubusercontent.com/cri-o/cri-o/master/scripts/get | bash -s -- -t v1.21.0
+```
+
+The above script resolves to the download URL of the static binary bundle
+tarball matching the format:
+
+```
+https://storage.googleapis.com/k8s-conform-cri-o/artifacts/cri-o.$ARCH.$REV.tar.gz
+```
+
+where `$ARCH` can be `amd64` or `arm64` and `$REV` can be any git SHA or tag.
+Please be aware that using the latest `master` SHA might cause a race, because
+the CI has not finished publishing the artifacts yet or failed.
 
 ### Running kubernetes with CRI-O
 
--- a/contrib/bundle/vars
+++ b/contrib/bundle/vars
@@ -6,8 +6,7 @@ GIT_ROOT=$(git rev-parse --show-toplevel)
 ARCHIVE_PATH="$GIT_ROOT/build/bundle"
 mkdir -p "$ARCHIVE_PATH"
 
-BUNDLE=crio-$(git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD)
-ARCHIVE="$BUNDLE.tar.gz"
+ARCH=${1:-amd64}
+ARCHIVE="cri-o.$ARCH.$(git describe --tags --exact-match 2>/dev/null || git rev-parse HEAD).tar.gz"
 
-export BUNDLE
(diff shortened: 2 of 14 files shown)


saschagrunert:
Works :heavy_check_mark: 

There is a slack conversation ongoing where we could do some manual external testing: https://kubernetes.slack.com/archives/CAZH62UR1/p1612626775090600

/hold

saschagrunert:
I have to update the PR once crun 0.18 is available: https://github.com/containers/crun/pull/604

TomSweeneyRedHat:
LGTM
assuming happy tests

saschagrunert:
This is ready for shipping, PTAL @haircommander @mrunalp @harche 

/hold cancel

contrib/bundle/vars
```diff
@@ -6,8 +6,7 @@ GIT_ROOT=$(git rev-parse --show-toplevel)
 ARCHIVE_PATH="$GIT_ROOT/build/bundle"
 mkdir -p "$ARCHIVE_PATH"
 
-BUNDLE=crio-$(git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD)
-ARCHIVE="$BUNDLE.tar.gz"
+ARCH=${1:-amd64}
+ARCHIVE="cri-o.$ARCH.$(git describe --tags --exact-match 2>/dev/null || git rev-parse HEAD).tar.gz"
```
harche, line 10: Cool. 

/cc @odinuge

README.md
```diff
@@ -98,17 +98,25 @@ up a PR and add it to the list.
...
 We also provide a way in building [static binaries of `CRI-O`](install.md#static-builds) via nix.
 Those binaries are available for every successfully built commit on our [Google Cloud Storage Bucket][bucket].
-This means that the latest commit can be downloaded via:
+This means that the latest commit can be installed via our continence script:
 
 [bucket]: https://console.cloud.google.com/storage/browser/k8s-conform-cri-o/artifacts
 
 ```shell
-> curl -f https://storage.googleapis.com/k8s-conform-cri-o/artifacts/crio-$(git ls-remote https://github.com/cri-o/cri-o master | cut -c1-9).tar.gz -o crio.tar.gz
+> curl https://raw.githubusercontent.com/cri-o/cri-o/master/get | bash
```
harche, line 112: The url mentioned in https://github.com/cri-o/cri-o/pull/4587#discussion_r578314427 is still valid right? Because that's what I am using in https://github.com/kubernetes/test-infra/pull/20898

Would it be a good idea to have a complete URL, say as an example only for `amd64`, mentioned here? Just for someone who might not want to pipe a script from web into bash or someone like me who end up using complete URL in a systemd service file.

harche (commented):
Looks cool @saschagrunert. Just a couple of queries, but otherwise LGTM. :)

README.md
saschagrunert: Yes, that still applies. I changed the README.md to mention that URL path, too. But we have to be aware that this could cause a race (CI not finished or failed) in comparison to the script which will always succeed.

harche: Oh, ok, then in my case, I will change the systemd service file to first download the script and execute it locally.

saschagrunert: Yeah, this is way more safe and decouples our CI from the consumer of the script.

saschagrunert:
> Looks cool @saschagrunert. Just a couple of queries, but otherwise LGTM. :)

Thank you! Please also note:
```
The static binary bundle now uses crun as default container runtime for CRI-O.
```

get
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ARCH=${1:-amd64}
```
odinuge, line 4: It would be nice to just default to empty, and use `uname` to detec the correct arch.

```
$ uname -m #x86
x86_64 
$ uname -m #arm64 
aarch64
```

get
```diff
@@ -0,0 +1,47 @@
...
+VERSION=${2:-}
```
odinuge, line 5: Can we default to "latest", and fetch latest in that case (instead of the current `""`)? If so, it would be nice to have version first and arch in the end. You often might want a pinned version, but without explicitly setting the arch.

README.md
harche: Thanks. I was wondering, if it's possible to keep the old url around for sometime. It would take me sometime to get this tested and added in test-infra. Until then the periodic jobs, https://prow.k8s.io/?job=*crio*, will keep failing :( 

Just checking if it's possible, hate to see the prow dashboard in red :)

odinuge: I really like this approach @saschagrunert! The old one was kinda fragile, so thanks!

get
saschagrunert: I reworked the script to now accept parsed CLI arguments as well as a help output. We now fallback to `uname -m` (converted to amd64/arm64), too. :)

get
saschagrunert: Fixed that by the rework of the script, see https://github.com/cri-o/cri-o/pull/4552#discussion_r579089773

get
odinuge: Awesome!! :smile:

saschagrunert:
@haircommander do you think we can get this one in?

haircommander:
I know I just gave this lgtm, but I am hesitant to introduce a script to the top level of the tree

/lgtm cancel

can we move it to the scripts/ directory? or is there a reason to keep it top-level?

saschagrunert:
> can we move it to the scripts/ directory? or is there a reason to keep it top-level?

Yes sure, moved the script into the `scripts` directory.  :+1:

PR #4557, opened 2021-02-09
[CLI] "crio config" only prints the fields that are different than the default

### LINKED ISSUE #4506: `crio config` fixes

ityuhui:
As described in #4506 , this PR lets ``crio config`` only prints the fields that are different than the default.

Fixes #4506

Yes.

```release-note
"crio config" only prints the fields that are different than the default configuration.
```

### DIFF
--- a/cmd/crio/config.go
+++ b/cmd/crio/config.go
@@ -74,6 +74,6 @@ it later with **--config**. Global options will modify the output.`,
 		}
 
 		// Output the commented config.
-		return conf.WriteTemplate(os.Stdout)
+		return conf.WriteTemplate(c.Bool("default"), os.Stdout)
 	},
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1078,3 +1078,340 @@ func (c *Config) SetLocations(singleConfigPath, dropInConfigDir string) {
 func (c *NetworkConfig) CNIPlugin() ocicni.CNIPlugin {
 	return c.cniPlugin
 }
+
+type ChangeResult map[string]bool
+
+func StringSliceEqual(a, b []string) bool {
+	if (a == nil) && (b == nil) {
+		return true
+	}
+
+	if (a == nil) && (len(b) == 0) {
+		return true
+	}
+
+	if (b == nil) && (len(a) == 0) {
+		return true
+	}
+
+	if len(a) != len(b) {
+		return false
+	}
+
+	for i, v := range a {
+		if v != b[i] {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (c *Config) CompareToDefaultConfig() (ChangeResult, error) {
+	configChangeResult := make(map[string]bool)
+
+	dc, err := DefaultConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	// [crio] configuration
+	if c.Root != dc.Root {
+		configChangeResult["root"] = true
+	}
+
+	if c.RunRoot != dc.RunRoot {
+		configChangeResult["runroot"] = true
+	}
+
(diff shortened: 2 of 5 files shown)


saschagrunert:
Thank you @ityuhui 
/ok-to-test

haircommander:
this is an awesome @ityuhui !

I think I'd like to see more unit tests and integration tests. Unit tests can be added in `pkg/config` in go, and integration tests can be added in `test/` in bats.

haircommander:
thanks for taking this on

mrunalp:
We should have a boolean to control the behavior so we can still print everything if we want to.

haircommander:
we have the `crio config --default` flag to do so :)

ityuhui:
> I think I'd like to see more unit tests and integration tests. Unit tests can be added in `pkg/config` in go, and integration tests can be added in `test/` in bats.

Sure. I will do it.

ityuhui:
> we have the `crio config --default` flag to do so :)

Hi @haircommander 

The flag ``crio config --default`` will print all **default** configurations, other than all default + configured by crio.conf or crio.conf.d

Do we need add a new flag ``crio config --print-all`` to meet this requirement ?

haircommander:
> > we have the `crio config --default` flag to do so :)
> 
> Hi @haircommander
> 
> The flag `crio config --default` will print all **default** configurations, other than all default + configured by crio.conf or crio.conf.d
> 
> Do we need add a new flag `crio config --print-all` to meet this requirement ?

honestly, I don't think we want this. Moving forward, I think we should really push users to either use the cri-o defaults, or use a drop-in to override. the `--default` could just be for packagers that *really* want some defaults. But I don't think we should have a special flag that prints all the fields, including that which was overridden. That's what the `/etc/crio/crio.conf.d` is for :)

haircommander:
I also wonder if this can be refactored to be a bit more structured. Right now, there are like 4 places someone has to change to add a new config value, and I think this will add another. IT would be really cool if we could iterate through a slice of structs like
```go
type ConfigValue struct {
  templateString string,
  cliFlag              string,
  kind                  reflect.Kind, // may be useful for accessing the data
  compFunc        func(struct{}, struct{}) bool, // verify the two fields are the same
  defaultValue     struct{},
...
}
```

or something and iterate through that to apply these diffs/add template values.

Not really a blocking nit for this PR, but a general direction I think would be good. If you have the time to try something like that here, I'd be interested in seeing what you put together :smile:

ityuhui:
> I also wonder if this can be refactored to be a bit more structured. Right now, there are like 4 places someone has to change to add a new config value, and I think this will add another. IT would be really cool if we could iterate through a slice of structs like
> 
> ```go
> type ConfigValue struct {
>   templateString string,
>   cliFlag              string,
>   kind                  reflect.Kind, // may be useful for accessing the data
>   compFunc        func(struct{}, struct{}) bool, // verify the two fields are the same
>   defaultValue     struct{},
> ...
> }
> ```
> 
> or something and iterate through that to apply these diffs/add template values.
> 

I'd like to try.

PR #4563, opened 2021-02-10
config/node: refactor checking for CollectMode

haircommander:
shout out to Kir (@kolyshkin) for coming up with this method
Drastically simplify querying systemd for CollectMode.
This is more resistant to the feature being backported into the version shipped in distros.

#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

```release-note
Fix a bug where CollectMode wouldn't be set if the feature was backported to systemd (in RHEL/CentOS 7, for instance)
```

### DIFF
--- a/internal/config/node/systemd.go
+++ b/internal/config/node/systemd.go
@@ -4,8 +4,6 @@ package node
 
 import (
 	"os/exec"
-	"regexp"
-	"strconv"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


internal/config/node/systemd.go
```diff
@@ -19,26 +17,13 @@ var (
 
 func SystemdHasCollectMode() bool {
 	systemdHasCollectModeOnce.Do(func() {
-		stdout, err := exec.Command("systemctl", "--version").Output()
+		// This will show whether the currently running systemd supports CollectMode
+		_, err := exec.Command("systemctl", "show", "-p", "CollectMode", "systemd").Output()
 		if err != nil {
-			systemdHasCollectModeErr = err
+			systemdHasCollectModeErr = errors.Wrapf(err, "Failed to check systemd CollectMode")
```
saschagrunert, line 23: nit, errors already indicate that something has failed and should start lower case:
```suggestion
			systemdHasCollectModeErr = errors.Wrapf(err, "check systemd CollectMode")
```

haircommander: good point, fixed!

mrunalp (approved):

fidencio (commented):
LGTM, thanks @haircommander!

internal/config/node/systemd.go
```diff
@@ -19,26 +17,13 @@ var (
 
 func SystemdHasCollectMode() bool {
 	systemdHasCollectModeOnce.Do(func() {
-		stdout, err := exec.Command("systemctl", "--version").Output()
+		// This will show whether the currently running systemd supports CollectMode
+		_, err := exec.Command("systemctl", "show", "-p", "CollectMode", "systemd").Output()
 		if err != nil {
-			systemdHasCollectModeErr = err
+			systemdHasCollectModeErr = errors.Wrapf(err, "check systemd CollectMode")
```
kolyshkin, line 23: nit: I would exclude "exit code 1" as being an error.

haircommander: do you mean we should not set the error? or I shouldn't include the contents of `err` in the returned error ?

kolyshkin: Not set the error if we got exit code 1 (which 99% means that the option is not supported).

PR #4568, opened 2021-02-12
Move integration tests to GitHub actions

### LINKED ISSUE #4489: Move to GitHub actions

saschagrunert:
#### What type of PR is this?

/kind ci

This moves the last tests from CricleCI to GitHub actions, which
completes the migration.

Fixes #4489 
None

```release-note
None
```


saschagrunert:
I have to unlink the project from CircleCI afterwards, otherwise it will complain that there is no config.

PR #4573, opened 2021-02-12
sandbox: Initialize the infra container internally

haircommander:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
/kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

We should move to a world where most of the infra configuration happens in `pkg/sandbox/` instead of `server/`.

This PR begins that process, as well as cleans up some unused functions/files and makes the usage of context more idiomatic

```release-note
none
```

### DIFF
--- a/pkg/sandbox/infra.go
+++ b/pkg/sandbox/infra.go
@@ -1,0 +1,70 @@
+package sandbox
+
+import (
+	"fmt"
+
+	"github.com/cri-o/cri-o/internal/storage"
+	libconfig "github.com/cri-o/cri-o/pkg/config"
+	"github.com/cri-o/cri-o/pkg/container"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/runtime-tools/generate"
+)
+
+func (s *sandbox) InitInfraContainer(serverConfig *libconfig.Config, podContainer *storage.ContainerInfo) error {
+	var err error
+	s.infra, err = container.New()
+	if err != nil {
+		return err
+	}
+
+	// determine pause command before initializing generator
+	// so a failure will not result in a half configured generator
+	pauseCommand, err := PauseCommand(serverConfig, podContainer.Config)
+	if err != nil {
+		return err
+	}
+
+	g := s.infra.Spec()
+	g.HostSpecific = true
+	g.ClearProcessRlimits()
+
+	// setup defaults for the pod sandbox
+	g.SetRootReadonly(true)
+
+	// configure default ulimits
+	for _, u := range serverConfig.Ulimits() {
+		g.AddProcessRlimits(u.Name, u.Hard, u.Soft)
+	}
+	g.SetProcessArgs(pauseCommand)
+
+	return nil
+}
+
+// Spec can only be called after a successful call to InitInfraContainer
+func (s *sandbox) Spec() *generate.Generator {
+	return s.infra.Spec()
+}
+
+// PauseCommand returns the pause command for the provided image configuration.
+func PauseCommand(cfg *libconfig.Config, image *v1.Image) ([]string, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("provided configuration is nil")
+	}
+
+	// This has been explicitly set by the user, since the configuration
+	// default is `/pause`
+	if cfg.PauseCommand != "" {
+		return []string{cfg.PauseCommand}, nil
+	}
+	if image == nil || (len(image.Config.Entrypoint) == 0 && len(image.Config.Cmd) == 0) {
+		return nil, fmt.Errorf(
+			"unable to run pause image %q: %s",
+			cfg.PauseImage,
+			"neither Cmd nor Entrypoint specified",
+		)
+	}
+	cmd := []string{}
+	cmd = append(cmd, image.Config.Entrypoint...)
+	cmd = append(cmd, image.Config.Cmd...)
+	return cmd, nil
+}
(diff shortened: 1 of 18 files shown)


haircommander:
@saschagrunert @giuseppe @mrunalp @umohnani8 @kolyshkin @QiWang19 @wgahnagl PTAL

saschagrunert (approved):
LGTM

QiWang19 (commented):
LGTM

pkg/sandbox/infra.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return nil
+}
+
+func (s *sandbox) Spec() *generate.Generator {
+	return s.infra.Spec()
+}
+
+// PauseCommand returns the pause command for the provided image configuration.
+func PauseCommand(cfg *libconfig.Config, image *v1.Image) ([]string, error) {
```
kolyshkin, line 45: Why is this a public method?

pkg/sandbox/infra.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return err
+	}
+	g.SetProcessArgs(pauseCommand)
+
+	return nil
+}
+
+func (s *sandbox) Spec() *generate.Generator {
+	return s.infra.Spec()
```
kolyshkin, line 41: this might panic if called before `InitializeInfraContainer`.

pkg/sandbox/infra.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// setup defaults for the pod sandbox
+	g.SetRootReadonly(true)
+
+	// configure default ulimits
+	for _, u := range serverConfig.Ulimits() {
+		g.AddProcessRlimits(u.Name, u.Hard, u.Soft)
+	}
+
+	pauseCommand, err := PauseCommand(serverConfig, podContainer.Config)
```
kolyshkin, line 31: I'd move this call up so in case the function return an error we won't end up with half-initialized `g`

pkg/sandbox/infra.go
```diff
@@ -0,0 +1,67 @@
...
+
+// PauseCommand returns the pause command for the provided image configuration.
+func PauseCommand(cfg *libconfig.Config, image *v1.Image) ([]string, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("provided configuration is nil")
+	}
+
+	// This has been explicitly set by the user, since the configuration
+	// default is `/pause`
+	if cfg.PauseCommand == "" {
```
kolyshkin, line 52: I'd do
```go
if cfg.PauseCommand != "" {
    return []string{cfg.PauseCommand}, nil
}
if image == nil || 
...
```

kolyshkin:
LGTM overall, left a few nits.

pkg/sandbox/infra.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+
+	"github.com/cri-o/cri-o/internal/storage"
+	libconfig "github.com/cri-o/cri-o/pkg/config"
+	"github.com/cri-o/cri-o/pkg/container"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/runtime-tools/generate"
+)
+
+func (s *sandbox) InitializeInfraContainer(serverConfig *libconfig.Config, podContainer *storage.ContainerInfo) error {
```
kolyshkin, line 13: Perhaps `InitInfraContainer`?

pkg/sandbox/infra.go
haircommander: for unit tests, as we test from a different package

pkg/sandbox/infra.go
haircommander: I don't think we should gracefully recover from that situation, that feels like a good time for an assert. Would you like the panic to be more deliberate?

haircommander: documentation could use updating too

pkg/sandbox/infra.go
haircommander: good call

pkg/sandbox/infra.go
haircommander: fixed

pkg/sandbox/infra.go
haircommander: done!

pkg/sandbox/infra.go
haircommander: I updated the docs of the function to make clear the required ordering

pkg/sandbox/infra.go
```diff
@@ -0,0 +1,70 @@
...
+	// configure default ulimits
+	for _, u := range serverConfig.Ulimits() {
+		g.AddProcessRlimits(u.Name, u.Hard, u.Soft)
+	}
+	g.SetProcessArgs(pauseCommand)
+
+	return nil
+}
+
+// Must be called after a successful call to InitInfraContainer
```
kolyshkin, line 43: "Can only be called only ..."?

kolyshkin: Also, a docstring should start with the method name, as in
```go
// Spec <does this and that>...
```

It's merely a convention, but a good one. Not sure why our linters do not catch this -- afaik they should

kolyshkin: Possibly this is because it is an internal package.

kolyshkin (approved):
LGTM

haircommander:
refers to https://github.com/cri-o/cri-o/issues/4622

Issue #4589, opened 2021-02-19
runtimeVM: CRI-O doesn't properly convert the runtime path to the expected shimv2 name


fidencio:
**Description**

shim v2, used by the VM runtime type, expects the runtime name to be in the following format:
```
    ($dir.)?$prefix.$name.$version
    -------- ------ ----- ---------
        |     |     |      |
        v     |     |      |
"/usr/local/bin"    |      |
  (optional)  |     |      |
              v     |      |
       "containerd.shim."  |
                    |      |
                    v      |
               "kata-qemu" |
                           v
                          "v2"
```

The current code simply replaces, from the binary name, all the "-" by ".", which ends up working well as long as the $name doesn't have a "-" in it, which is, by luck, what's done by the Kata Operator.  However, whenever a name has "-" in it, which we can see happening on upstream kata-deploy - which uses $name as kata-qemu, for instance - it'll simply make containerd
library confused, not being able to invoke the binary.

**Steps to reproduce the issue:**
1. Configure a Kata Containers environment (sorry)
2. Change the name of the binary from `containerd-shim-kata-v2` to, `containers-shim-kata-something-v2`
3. Update the CRI-O config to match the change of the binary's name
3. Try to create a pod using the new config

**Describe the results you received:**
Pod is not created due to:
```
  Warning  FailedCreatePodSandBox  1s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = runtime "/usr/local/bin/containerd.shim.kata.qemu.v2" binary not installed "containerd-shim-qemu-v2": file does not exist
```

**Describe the results you expected:**
Pod would be started as expected.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `crio --version`:**

```
[fidencio@demo cri-o]$ crio --version
INFO[0000] Starting CRI-O, version: 1.21.0-dev, git: 2a1d1136213a5181cf9f3f6e5d00bb1a27dcf550(clean) 
crio version 1.21.0-dev
Version:       1.21.0-dev
GitCommit:     2a1d1136213a5181cf9f3f6e5d00bb1a27dcf550
GitTreeState:  clean
BuildDate:     2021-02-19T17:47:40Z
GoVersion:     go1.14.12
Compiler:      gc
Platform:      linux/amd64
Linkmode:      dynamic
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

PR #4590, opened 2021-02-19
runtimeVM: Fix shimv2 binary name construction

fidencio:
#### What type of PR is this?

/kind bug

shim v2 expected the runtime name to be in the following format:
```
    ($dir.)?$prefix.$name.$version
    -------- ------ ----- ---------
        |     |     |      |
        v     |     |      |
"/usr/local/bin"    |      |
  (optional)  |     |      |
              v     |      |
       "containerd.shim."  |
                    |      |
                    v      |
               "kata-qemu" |
                           v
                          "v2"
```

The current code simply replaces all the "-" by ".", which ends up
working well as long as the $name doesn't have a "-" in it, which is, by
luck, what's done by the Kata Operator.  However, whenever a name has
"-" in it, which we can see happening on upstream kata-deploy - which
uses $name as kata-qemu, for instance - it'll simply make containerd
library confused, not being able to invoke the binary.

With everything said above in mind, let's introduce a specialized
function to build the shim v2 path.

Fixes: #4589

```release-note
None
```

### DIFF
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -476,4 +478,42 @@ type ExecSyncError struct {
 
 func (e *ExecSyncError) Error() string {
 	return fmt.Sprintf("command error: %+v, stdout: %s, stderr: %s, exit code %d", e.Err, e.Stdout.Bytes(), e.Stderr.Bytes(), e.ExitCode)
+}
+
+// BuildContainerdBinaryName() is responsible for ensuring the binary passed will
+// be properly converted to the containerd binary naming pattern.
+//
+// This method should never ever be called from anywhere else but the runtimeVM,
+// and the only reason its exported here is in order to get some test coverage.
+func BuildContainerdBinaryName(path string) string {
+	// containerd expects the runtime name to be in the following pattern:
+	//        ($dir.)?$prefix.$name.$version
+	//        -------- ------ ----- ---------
+	//              |     |     |      |
+	//              v     |     |      |
+	//      "/usr/local/bin"    |      |
+	//        (optional)  |     |      |
+	//                    v     |      |
+	//             "containerd.shim."  |
+	//                          |      |
+	//                          v      |
+	//                     "kata-qemu" |
+	//                                 v
+	//                                "v2"
+	const expectedPrefix = "containerd-shim-"
+	const expectedVersion = "-v2"
+
+	const binaryPrefix = "containerd.shim"
+	const binaryVersion = "v2"
+
+	runtimeDir := filepath.Dir(path)
+	// This is only safe to do because the runtime_path, for the VM runtime_type, is validated in the config,
+	// allowing us to take the liberty to simply go ahead and check, without having to ensure we're receiving
+	// the binary in the expected form.
+	//
+	// For clarity, it could be ensured twice, but we count on the developer to never ever call this function
+	// in a different context from the one used in the runtime_vm.go file.
+	runtimeName := strings.SplitAfter(strings.Split(filepath.Base(path), expectedVersion)[0], expectedPrefix)[1]
+
+	return filepath.Join(runtimeDir, fmt.Sprintf("%s.%s.%s", binaryPrefix, runtimeName, binaryVersion))
 }
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -189,7 +189,7 @@ func (r *runtimeVM) startRuntimeDaemon(ctx context.Context, c *Container) error 
 	args = append(args, "start")
 
 	// Modify the runtime path so that it complies with v2 shim API
-	newRuntimePath := strings.ReplaceAll(r.path, "-", ".")
+	newRuntimePath := BuildContainerdBinaryName(r.path)
 
 	ctx = namespaces.WithNamespace(ctx, namespaces.Default)
 
--- a/internal/oci/oci_test.go
+++ b/internal/oci/oci_test.go
@@ -40,7 +40,8 @@ var _ = t.Describe("Oci", func() {
 				RuntimePath: "/bin/sh",
 				RuntimeType: "",
 				RuntimeRoot: "/run/runc",
-			}, invalidRuntime: {},
+			},
+			invalidRuntime: {},
 			usernsRuntime: {
 				RuntimePath:        "/bin/sh",
 				RuntimeType:        "",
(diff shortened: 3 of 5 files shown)


internal/oci/runtime_vm.go
```diff
@@ -52,6 +53,33 @@ type containerInfo struct {
...
+	//                                 v
+	//                                "v2"
+	const expectedPrefix = "containerd-shim-"
+	const expectedVersion = "-v2"
+
+	const binaryPrefix = "containerd.shim"
+	const binaryVersion = "v2"
+
+	runtimePrefix := filepath.Dir(path)
+	runtimeName := strings.TrimLeft(strings.TrimRight(filepath.Base(path), expectedVersion), expectedPrefix)
```
fidencio, line 78: Oh, I thought Trim{Left,Right} would take prefixes instead of cutsets.  Okay, TIL and this is something I need to update.

internal/oci/runtime_vm.go
```diff
@@ -52,6 +53,33 @@ type containerInfo struct {
...
+	const expectedPrefix = "containerd-shim-"
+	const expectedVersion = "-v2"
+
+	const binaryPrefix = "containerd.shim"
+	const binaryVersion = "v2"
+
+	runtimeDir := filepath.Dir(path)
+	runtimeName := strings.SplitAfter(strings.Split(filepath.Base(path), expectedVersion)[0], expectedPrefix)[1]
+
+	return filepath.Join(runtimeDir, fmt.Sprintf("%s.%s.%s", binaryPrefix, runtimeName, binaryVersion))
```
fgiudici, line 80: Here we enforce the containerd v2 shim prefix and suffix (also if not present in the passed path).

fgiudici (commented):
Here we assume that the VM runtime binary is in the form we expect (which is the containerd shimv2 format, i.e., containerd-shim-$RUNTIMENAME-v2). If that assumption is not met (wrong config file?) we can tweak the name or panic (see inline comments).
Maybe we already check the runtime VM name somewhere earlier... or maybe we don't need to deal with this here anyway (which would mean deal with unknown VM type runtimes or wrong configuration files): if so, just ignore the comments please.
Apart from this, nice patch, LGTM!

internal/oci/runtime_vm.go
```diff
@@ -52,6 +53,33 @@ type containerInfo struct {
 	cio *cio.ContainerIO
 }
 
+func buildShimV2Path(path string) string {
```
haircommander, line 56: I think this may be better suited for config validation step in pkg/config/config.go

that way, we fail early if a user configured vm runtime incorrectly

internal/oci/runtime_vm.go
haircommander: do we verify the file exists later?

internal/oci/runtime_vm.go
fidencio: @haircommander, let me see if I understand your suggestion correctly.  Your suggestion seems to be along the lines of what @fgiudici suggested here: https://github.com/cri-o/cri-o/pull/4590#discussion_r580187812, and I'm totally in for that.

With that in mind, we'd still need to build the "ShimV2Path", as what we pass to containerd is **not** exactly the same as the path of the binary.

For instance, for let's take the `containerd-shim-kata-v2` binary as example:
```
containerd-shim-kata-v2
                ----
                  |
                  v
         this should not change, and
    every other "-" should be replaces by "."
```
This will lead us to "containerd.shim.kata.v2" as the name passed to `containerd`.


Now, let's take `containerd-shim-kata-qemu-v2`, binary as example:
```
containerd-shim-kata-qemu-v2
                ---------
                    |
                    v
             this should not change, and
        every other "-" should be replaces by "."
```
This will lead us to "containerd.shim.kata-qemu.v2" as the name passed to `contained`.

So, AFAIU, we could validate that the binary name would be something like:
`(/\\w\/-+)+/containerd-shim-\\w+-v2` as part of the pkg/config/config.go (and, again, that does make sense), but we would still have to convert the `containerd-shim-$something-v2" to `containerd.shim.$something.v2`.

Based on the explanation above, do you still think the whole `buildShimV2Path` could be done at the config validation step in `pkg/config/config.go`?

internal/oci/runtime_vm.go
fidencio: We, as CRI-O, don't, but do we need?
Once this name is built it's passed to https://github.com/containerd/containerd/blob/08d765af954cd5355b7207fe628d06ed5a97c9d4/runtime/v2/shim/util.go#L41, which does the validation here https://github.com/containerd/containerd/blob/08d765af954cd5355b7207fe628d06ed5a97c9d4/runtime/v2/shim/util.go#L61-L94

fidencio: Oh, actually CRI-O does check the existence of the binary during the validation step.

haircommander:
I am fine with this here, and we can add validation in config later if we want

before we move forward with this PR though I'd like to see some unit tests :)

haircommander:
that button is too close to the comment button :upside_down_face:

fidencio:
@haircommander, @fgiudici,

Sorry, it took me longer than I wanted to be able to revisit this series.
I've addressed your comments, or I think I did, in this new series by:
* Enforcing a "containerd-shim-%s-v2" pattern (more or less), during the RuntimePath() validation, when using the "VM" runtime type;
* I've implemented some tests, the basic ones for the cases I'm trying to cover here;

Also as part of this series, I  ended up increasing the unit tests we already have so we cover other pieces related to the "VM" runtime type.

Ah, I must admit I'm not that comfortable with the `BuildContainerdBinaryName` being exposed as part of the `oci` package, but there's not much I can do, as that's the way to have it tested.  Meanwhile, let's count that the next developer using that piece of code will read the comments I added to that method.

fidencio:
And now the linter should be happier.

haircommander:
/approve

LGTM
/override ci/kata-jenkins

fgiudici:
LGTM

saschagrunert (approved):
/lgtm

internal/oci/oci.go
```diff
@@ -477,3 +479,41 @@ type ExecSyncError struct {
 func (e *ExecSyncError) Error() string {
 	return fmt.Sprintf("command error: %+v, stdout: %s, stderr: %s, exit code %d", e.Err, e.Stdout.Bytes(), e.Stderr.Bytes(), e.ExitCode)
 }
+
+// BuildContainrdBinaryName() is responsible for ensuring the binary passed will
```
TomSweeneyRedHat, line 483: nit of a nit
```suggestion
// BuildContainerdBinaryName() is responsible for ensuring the binary passed will
```

TomSweeneyRedHat:
LGTM
other than the nit which could be cleaned up later.

internal/oci/oci.go
fidencio: Fixed, thanks,

fidencio:
@saschagrunert, @haircommander, I may a new `/lgtm` label here. :-)

PR #4592, opened 2021-02-20
Fix oci container update config

odinuge:
#### What type of PR is this?

It is better to skip sending the values, instead of sending 0s.
crun will try to enforce a memory.Limit=0, making the contianer OOM.

We can also do the same as containerd to copy old spec and update the values, but I think
this is better.

> /kind bug

```release-note
Fix OOMs when using crun together with CPU Manager
```


harche:
Thanks @odinuge. 

/cc @haircommander

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM
and happy green buttons.

haircommander:
thanks a bunch @odinuge

/ok-to-test

PR #4600, opened 2021-02-24
oci: don't wait too long on a long stop

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

oci: don't wait too long on a long stop

if a long stop timeout comes in, and a shorter one comes in after
we never interrupt the old one. This can cause the new container stop to be DOSed

instead, interrupt the old stop, so the new one can proceed as expected

This PR is https://github.com/cri-o/cri-o/pull/4548 plus a couple of usability tweaks and unit tests

```release-note
none
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -72,6 +72,10 @@ type Container struct {
 	stdinOnce          bool
 	created            bool
 	spoofed            bool
+	stopping           bool
+	stopTimeoutChan    chan time.Duration
+	stoppedChan        chan struct{}
+	stopLock           sync.Mutex
 }
 
 // Metadata holds all necessary information for building the container name.
@@ -132,6 +136,8 @@ func NewContainer(id, name, bundlePath, logPath string, labels, crioAnnotations,
 		dir:             dir,
 		state:           state,
 		stopSignal:      stopSignal,
+		stopTimeoutChan: make(chan time.Duration, 1),
+		stoppedChan:     make(chan struct{}, 1),
 	}
 	return c, nil
 }
@@ -545,4 +551,35 @@ func (c *Container) ShouldBeStopped() error {
 // is not needed, but sandbox metadata should be stored with a spoofed infra container.
 func (c *Container) Spoofed() bool {
 	return c.spoofed
+}
+
+// SetAsStopping marks a container as being stopped.
+// If a stop is currently happening, it also sends the new timeout
+// along the stopTimeoutChan, allowing the in-progress stop
+// to stop faster, or ignore the new stop timeout.
+func (c *Container) SetAsStopping(timeout int64) {
+	// First, need to check if the container is already stopping
+	c.stopLock.Lock()
+	defer c.stopLock.Unlock()
+	if c.stopping {
+		// If so, we shouldn't wait forever on the opLock.
+		// This can cause issues where the container stop gets DOSed by a very long
+		// timeout, followed a shorter one coming in.
+		// Instead, interrupt the other stop with this new one.
+		select {
+		case c.stopTimeoutChan <- time.Duration(timeout) * time.Second:
+		case <-c.stoppedChan: // This case is to avoid waiting forever once another routine has finished.
+			return
+		}
+	}
+	// Regardless, set the container as actively stopping.
+	c.stopping = true
+}
+
+// SetAsNotStopping unsets the stopping field indicating to new callers that the container
+// is no longer actively stopping.
+func (c *Container) SetAsNotStopping() {
+	c.stopLock.Lock()
+	c.stopping = false
+	c.stopLock.Unlock()
 }
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -462,6 +462,9 @@ func (c *Container) pid() (int, error) {
 	if err := c.verifyPid(); err != nil {
 		return 0, err
 	}
+	if err := unix.Kill(c.state.InitPid, 0); err == unix.ESRCH {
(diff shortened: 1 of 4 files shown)


haircommander:
the third and fourth commit should be squashed to the second once we decide they're the right behaviour, I've kept them separate because they veer in behaviour from #4548

mrunalp:
@saschagrunert ptal

internal/oci/container.go
```diff
@@ -462,6 +468,9 @@ func (c *Container) pid() (int, error) {
 	if err := c.verifyPid(); err != nil {
 		return 0, err
 	}
+	if err := unix.Kill(c.state.InitPid, 0); errors.Is(err, unix.ESRCH) {
+		return 0, fmt.Errorf("check whether %d is running: %v", c.state.InitPid, err)
```
saschagrunert, line 472:
```suggestion
		return 0, errors.Wrapf(err, "check whether %d is running", c.state.InitPid)
```

saschagrunert (approved):
LGTM, just found one non-blocking nit.

internal/oci/container.go

internal/oci/runtime_oci.go
```diff
@@ -616,33 +616,68 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
...
+			// interrupt the old one, and start a new one
+			newTargetTime := time.Now().Add(newTimeout)
+
+			// but only if it's earlier
+			if !newTargetTime.Before(targetTime) {
+				continue
+			}
+
+			targetTime = newTargetTime
+			timeout = newTimeout
```
TomSweeneyRedHat, line 656: I get super twitchy when I see a select with no default, no matter how improbable it is to hit it.  At the very least I think you should have a debugf with a "we should never hit this" type of message.

haircommander: Unfortunately, I may need to leave you twitching, as it won't wait the timeout if there's a default
https://play.golang.org/p/hMqswDkjs8V

internal/oci/container.go
```diff
@@ -462,6 +462,9 @@ func (c *Container) pid() (int, error) {
 	if err := c.verifyPid(); err != nil {
 		return 0, err
 	}
+	if err := unix.Kill(c.state.InitPid, 0); errors.Is(err, unix.ESRCH) {
+		return 0, errors.Wrapf(err, "check whether %d is running", c.state.InitPid)
+	}
```
kolyshkin, line 473: Since `kill` is faster than `c.verifyPid()` I'd put it first.

OTOH if you want to check it as late as possible, leave it as is.

internal/oci/container.go
```diff
@@ -546,3 +552,34 @@ func (c *Container) ShouldBeStopped() error {
 func (c *Container) Spoofed() bool {
 	return c.spoofed
 }
+
+// SetAsStopping marks a container as being stopped
```
kolyshkin, line 556: nit: `.` is missing

internal/oci/container.go
```diff
@@ -462,6 +462,9 @@ func (c *Container) pid() (int, error) {
 	if err := c.verifyPid(); err != nil {
 		return 0, err
 	}
+	if err := unix.Kill(c.state.InitPid, 0); errors.Is(err, unix.ESRCH) {
```
kolyshkin, line 471: Also, since it's `unix`, the error is bare, and it would be a _tad_ faster to do a direct comparison, ie `err == unix.ESRCH`.

internal/oci/container.go
haircommander: waiting as long as possible allows us to *potentially* catch the case where the process has died but the proc entry hasn't been cleaned, which is most of the point of adding this kill. if we do it before, there's no way we'd catch it.

internal/oci/runtime_oci.go
```diff
@@ -616,33 +616,69 @@ func waitContainerStop(ctx context.Context, c *Container, timeout time.Duration,
...
 		}
 	}
-
 	c.state.Finished = time.Now()
+	// Successfully stopped! This is to prevent other routines from
+	// racing with this one and waiting forever.
+	// close only the dedicated channel. If we close stopTimeoutChan,
```
kolyshkin, line 662: nit: `s/close/Close/`.

internal/oci/container.go

internal/oci/container.go
haircommander: updated

saschagrunert (approved):
/lgtm

internal/oci/container.go
kolyshkin: I highly doubt that such situation (" the process has died but the proc entry hasn't been cleaned") is at all possible. Have you ever seen it?

Surely a process can disappear any time, say right after we read its /proc entry.

Adding yet another kill (with 0) to check if the process is still there before using this pid makes the race window a bit shorter but it does not close it. Instead we should either expect the pid to still be there (in most places), or disappear at any time (when stopping, like it is done in WaitContainerStop / Kill).

Anyway, syscalls are fast nowdays, I guess it won't hurt.

haircommander: I am not positive I've seen it, but before adding this kill, the window after running the kill  in the test needed to be higher, and adding the kill reduced the window. I believe it helped

PR #4610, opened 2021-03-02
Make `get` script independent from `make`

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now do not rely on `make` any more for sake of a higher portability
of the `get` script.

None
None

```release-note
None
```


scripts/get
```diff
@@ -94,8 +94,54 @@ download_and_install() {
...
+    CNIDIR=$ETCDIR/cni/net.d
+    BINDIR=$PREFIX/bin
+    MANDIR=$PREFIX/share/man
+    OCIDIR=$PREFIX/share/oci-umount/oci-umount.d
+    SELINUX=$(selinuxenabled 2>/dev/null && echo -Z)
+    BASHINSTALLDIR=$PREFIX/share/bash-completion/completions
+    FISHINSTALLDIR=$PREFIX/share/fish/completions
+    ZSHINSTALLDIR=$PREFIX/share/zsh/site-functions
+    SYSTEMDDIR=$PREFIX/lib/systemd/system
+    OPT_CNI_BIN_DIR=/opt/cni/bin
```
haircommander, line 111: is there any chance we can have a shared source file for these, in case the makefile values change?

saschagrunert: Hm, we could ship two additional scripts: `install` and `uninstall`. The `install` script could be re-used by the `get` script, but we would have to invoke an additional `curl` command... Not sure about how to solve it in a cleaner way...

scripts/get
```diff
@@ -94,8 +94,54 @@ download_and_install() {
...
+
+    install "$SELINUX" -d -m 755 $CNIDIR
```
harche, line 113: Thanks, I just verified `install` command exists on FCOS by default. It ships with very minimal set of commands, so I thought I will make sure anyway. 

```bash
[root@localhost core]# which install 
/usr/bin/install
```

scripts/get
kolyshkin: We can have something like `scripts/install-paths`:
```make
# NOTE this file is sourced from make and shell
PREFIX=/usr/local
ETCDIR=/etc
CONTAINERS_DIR=$ETCDIR/containers
....
```

and we can use this from Makefile:
```make
include ./scripts/install-paths

...

install: 
        install ${SELINUXOPT} -D -m 755 bin/crio $(DESTDIR)$(BINDIR)/crio
.....
```

as well as from the `scripts/get`:
```sh
. install-paths
```

That way we at least have a single source for all these paths.

NOTE that currently `DESTDIR` is not handled correctly by the Makefile. It should never be included into `PREFIX`, but rather specified directly as in my example above.

kolyshkin: > and uninstall

I would drop the `uninstall` target from the makefile. It is not very robust anyway, as no one keeps track of what's installed. For example, `make DESTDIR=$(pwd)/inst install` followed by `make PREFIX=/usr uninstall` might do some real harm.

saschagrunert: Removed the uninstall part from the makefile and changed the implementation to just provide an `install` script.

saschagrunert:
Added some GitHub actions test case for the code-generation part. This should now work, PTAL again @haircommander @kolyshkin

scripts/get
```diff
@@ -83,26 +83,77 @@ latest_version() {
...
+
+# In this section we include the contents of
+# contrib/bundle/install-paths and contrib/bundle/install
+
+# INCLUDE
+DESTDIR=${DESTDIR:-}
+PREFIX=${PREFIX:-/usr/local}
+ETCDIR=${ETCDIR:-/etc}
+CONTAINERS_DIR=${CONTAINERS_DIR:-$ETCDIR/containers}
+CNIDIR=${CNIDIR:-$ETCDIR/cni/net.d}
```
haircommander, line 114: are we ablet o source the contrib scripts here?

saschagrunert: We pipe the script through curl, which means we would have to `curl` again at this point. Do we wanna do that or stick to the current `make get-scripts` approach?

haircommander: ach, damn, of course. No that's fine. I'd like some way that we future proof these fields, so they're not inconsitent between the (now two) places they exist (top-level makefile and here). not blocking on this PR, but probably some auto generation and hack/tree_status.sh will be the method...

saschagrunert: We have it via `make get-script`, the new GitHub actions job will verify that the content is sane:
https://github.com/cri-o/cri-o/pull/4610/checks?check_run_id=2022361523

:)

haircommander: then the only thing that needs fixing is my ability to read :)

haircommander:
/approve

LGTM

harche (approved):
LGTM

TomSweeneyRedHat:
LGTM 
assuming happy tests

saschagrunert:
@harche I think the script is now ready to be used within the k8s CI :)

PR #4613, opened 2021-03-02
inspect: use first container when reporting pid of a pod with no infra

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

.. that's a mouthful!

When we drop the infra container, there's no PID associated with the pod. This is fine, unless cAdvisor is trying to find the network stats for the pod, as it uses the PID to parse `/proc/$pid/net/dev`, and skips for a PID that is 0.

Luckily, containers in the pod share the pod's network namespace. We can find the first running container in the pod, return that PID, and cAdvisor will parse the correct dev file

```release-note
When dropping the infra container, the container inspect endpoint now returns a PID when requesting the infra container. 
```

### DIFF
--- a/server/inspect.go
+++ b/server/inspect.go
@@ -56,11 +56,13 @@ var (
 
 func (s *Server) getContainerInfo(id string, getContainerFunc, getInfraContainerFunc func(id string) *oci.Container, getSandboxFunc func(id string) *sandbox.Sandbox) (types.ContainerInfo, error) {
 	ctr := getContainerFunc(id)
+	isInfra := false
 	if ctr == nil {
 		ctr = getInfraContainerFunc(id)
 		if ctr == nil {
 			return types.ContainerInfo{}, errCtrNotFound
 		}
+		isInfra = true
 	}
 	// TODO(mrunalp): should we call UpdateStatus()?
 	ctrState := ctr.State()
@@ -78,9 +80,26 @@ func (s *Server) getContainerInfo(id string, getContainerFunc, getInfraContainer
 			image = status.Name
 		}
 	}
+
+	pidToReturn := ctrState.Pid
+	if isInfra && pidToReturn == 0 {
+		// It is possible the infra container doesn't report a PID.
+		// That can either happen if we're using a vm based runtime,
+		// or if we've dropped the infra container.
+		// Since the Pid is used exclusively to find the network stats,
+		// and pods share their network (whether it's host or pod level)
+		// we can return the pid of a running container in the pod.
+		for _, c := range sb.Containers().List() {
+			ctrPid, err := c.Pid()
+			if ctrPid > 0 && err == nil {
+				pidToReturn = ctrPid
+				break
+			}
+		}
+	}
 	return types.ContainerInfo{
 		Name:            ctr.Name(),
-		Pid:             ctrState.Pid,
+		Pid:             pidToReturn,
 		Image:           image,
 		ImageRef:        ctr.ImageRef(),
 		CreatedTime:     ctrState.Created.UnixNano(),
--- a/internal/config/cgmgr/cgmgr.go
+++ b/internal/config/cgmgr/cgmgr.go
@@ -10,7 +10,9 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/containers/libpod/v2/pkg/cgroups"
 	"github.com/cri-o/cri-o/internal/config/node"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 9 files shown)


haircommander:
I've carried https://github.com/cri-o/cri-o/pull/4100/commits/846f17b6f7bce370e424499d7c2713f759be3a65 because I wanted it and it was taking too long to merge

server/inspect.go
```diff
@@ -78,9 +78,24 @@ func (s *Server) getContainerInfo(id string, getContainerFunc, getInfraContainer
...
 	}
+
+	pidToReturn := ctrState.Pid
+	if ctr.Spoofed() && pidToReturn == 0 {
+		// This is possible when we've dropped the infra container
+		// and this request is for a pod.
+		// Since the Pid is used exclusively to find the network stats,
+		// and pods share their network (whether it's host or pod level)
+		// we can return the pid of a running container in the pod.
+		for _, c := range sb.Containers().List() {
```
mrunalp, line 89: Can we audit and make sure that the pid isn't used for anything else on the cadvisor side? :)

saschagrunert (approved):
LGTM

server/inspect.go
```diff
@@ -78,9 +78,24 @@ func (s *Server) getContainerInfo(id string, getContainerFunc, getInfraContainer
...
+			ctrPid, err := c.Pid()
+			if ctrPid > 0 && err != nil {
```
fgiudici, line 91: Should this be `err == nil`?

server/inspect.go
haircommander: it is also used for printing the container:
`klog.V(4).Infof("Could not get PIDs for container %d: %v", h.pid, err)`

but the only functional uses seem to be network related

haircommander:
/hold
for @fgiudici and @fidencio :+1:

server/inspect.go
haircommander: ah crap

fgiudici:
LGTM, I think this PR should be merged.
There is one missing piece anyway in order for the pod metrics to be retrived: cAdvisor collects the container metrics from CRI-O querying for the containers one by one. Anyway, cAdvisor retrieves **the list** of the containers ID parsing the child cgroups under the POD control group created by Kubernetes; then it queries CRI-O for each container ID. The container child control groups (one for each container in the POD) are created by the OCI runtime.
CRI-O just fills the runtime spec structure with the cgroup to be crearted (https://github.com/cri-o/cri-o/blob/master/server/sandbox_run_linux.go#L631), then the runtime actually takes care of creating the container control group.
So, as we "spoof" the infra container, we never pass the runtime spec to the OCI runtime. So, no one creates the cgroup corresponding to the spoofed container we have in CRI-O. As a result, cAdvisor will not have it in the list of containers to be queried in CRI-O and will not collect the (network) stats.
In order to fix this, we can create the cgroup ourselves for the spoofed container. It will be a cgroup with no process attached that will just trigger cAdvisor to retrieve the stats for the spoofed container from CRI-O. Something like this:
https://github.com/fgiudici/cri-o/commit/a61b07b8bec633cebab54d6330f1490fa77f1afc
May it be a viable path?

haircommander:
> LGTM, I think this PR should be merged.
> There is one missing piece anyway in order for the pod metrics to be retrived: cAdvisor collects the container metrics from CRI-O querying for the containers one by one. Anyway, cAdvisor retrieves **the list** of the containers ID parsing the child cgroups under the POD control group created by Kubernetes; then it queries CRI-O for each container ID. The container child control groups (one for each container in the POD) are created by the OCI runtime.
> CRI-O just fills the runtime spec structure with the cgroup to be crearted (https://github.com/cri-o/cri-o/blob/master/server/sandbox_run_linux.go#L631), then the runtime actually takes care of creating the container control group.
> So, as we "spoof" the infra container, we never pass the runtime spec to the OCI runtime. So, no one creates the cgroup corresponding to the spoofed container we have in CRI-O. As a result, cAdvisor will not have it in the list of containers to be queried in CRI-O and will not collect the (network) stats.
> In order to fix this, we can create the cgroup ourselves for the spoofed container. It will be a cgroup with no process attached that will just trigger cAdvisor to retrieve the stats for the spoofed container from CRI-O. Something like this:
> [fgiudici@a61b07b](https://github.com/fgiudici/cri-o/commit/a61b07b8bec633cebab54d6330f1490fa77f1afc)
> May it be a viable path?

great idea! PTAL :)

fgiudici:
LGTM!

haircommander:
/hold cancel

this is ready

server/inspect.go
```diff
@@ -78,9 +80,25 @@ func (s *Server) getContainerInfo(id string, getContainerFunc, getInfraContainer
...
+		// It is possible the infra container doesn't report a PID.
+		// That can either happen if we're using a vm based runtime,
+		// or if we've dropped the infra container.
+		// Since the Pid is used exclusively to find the network stats,
+		// and pods share their network (whether it's host or pod level)
+		// we can return the pid of a running container in the pod.
+		for _, c := range sb.Containers().List() {
+			ctrPid, err := c.Pid()
+			if ctrPid > 0 && err == nil {
+				pidToReturn = ctrPid
```
saschagrunert, line 95: Can we `break` after this has been set?

TomSweeneyRedHat:
Other than Sascha's question, LGTM

server/inspect.go
haircommander: fixed!

saschagrunert (approved):
/lgtm

PR #4614, opened 2021-03-02
test: add pids_limit test

haircommander:
to test various values

#### What type of PR is this?

> /kind api-change
> /kind bug
/kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

```release-note
none
```

### DIFF
--- a/test/cgroups.bats
+++ b/test/cgroups.bats
@@ -21,8 +21,10 @@ function run_pids_limit_test() {
 	ctr_id=$(crictl run "$TESTDIR"/container_pids_limit.json "$TESTDATA"/sandbox_config.json)
 
 	output=$(crictl exec --sync "$ctr_id" sh -c 'cat /sys/fs/cgroup/pids/pids.max 2>/dev/null || cat /sys/fs/cgroup/pids.max')
+	echo "got $output, expecting $expected_limit"
 	[[ "$output" == "$expected_limit" ]]
 }
+
 
 @test "pids limit" {
 	if ! grep -qEw ^pids /proc/cgroups; then
@@ -51,7 +53,7 @@ function run_pids_limit_test() {
 		pids_root=/sys/fs/cgroup/pids
 		lookfor=pids
 		if is_cgroup_v2; then
-			￼ pids_root=/sys/fs/cgroup
+	￼		pids_root=/sys/fs/cgroup
 			lookfor=""
 		fi
 		default_limit=$(cat "$pids_root/$(awk -v lookfor="$lookfor" -F : '$2 == lookfor {print $3; exit}' /proc/self/cgroup)/pids.max")
--- a/test/cgroups.bats
+++ b/test/cgroups.bats
@@ -10,19 +10,53 @@ function teardown() {
 	cleanup_test
 }
 
-@test "pids limit" {
-	if ! grep -qEw ^pids /proc/cgroups; then
-		skip "pids cgroup controller is not available"
-	fi
-	CONTAINER_PIDS_LIMIT=1234 start_crio
+function run_pids_limit_test() {
+	local set_limit="$1"
+	local expected_limit="$2"
 
+	CONTAINER_PIDS_LIMIT="$set_limit" start_crio
 	jq '	  .command'='["/bin/sleep", "600"]' \
 		"$TESTDATA"/container_config.json > "$TESTDIR"/container_pids_limit.json
 
 	ctr_id=$(crictl run "$TESTDIR"/container_pids_limit.json "$TESTDATA"/sandbox_config.json)
 
 	output=$(crictl exec --sync "$ctr_id" sh -c 'cat /sys/fs/cgroup/pids/pids.max 2>/dev/null || cat /sys/fs/cgroup/pids.max')
-	[[ "$output" == "1234" ]]
+	[[ "$output" == "$expected_limit" ]]
+}
+
+@test "pids limit" {
+	if ! grep -qEw ^pids /proc/cgroups; then
+		skip "pids cgroup controller is not available"
+	fi
+	run_pids_limit_test "1234" "1234"
+}
+
+@test "negative pids limit" {
+	if ! grep -qEw ^pids /proc/cgroups; then
+		skip "pids cgroup controller is not available"
+	fi
+	run_pids_limit_test "-1" "max"
+}
+
+@test "zero pids limit" {
+	if ! grep -qEw ^pids /proc/cgroups; then
+		skip "pids cgroup controller is not available"
+	fi
+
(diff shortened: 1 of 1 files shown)


test/cgroups.bats
```diff
@@ -25,6 +25,37 @@ function teardown() {
...
+	CONTAINER_PIDS_LIMIT=0 start_crio
+
+	jq '	  .command'='["/bin/sleep", "600"]' \
+		"$TESTDATA"/container_config.json > "$TESTDIR"/container_pids_limit.json
+
+	ctr_id=$(crictl run "$TESTDIR"/container_pids_limit.json "$TESTDATA"/sandbox_config.json)
+
+	output=$(crictl exec --sync "$ctr_id" sh -c 'cat /sys/fs/cgroup/pids/pids.max 2>/dev/null || cat /sys/fs/cgroup/pids.max')
+	# TODO where does this number come from?
+	[[ "$output" == "37725" ]]
```
haircommander, line 56: lots of duplicate code here... I put it up because I was surprised by this value on my machine

haircommander: any theories why this value is what it is? @kolyshkin @giuseppe @mrunalp

kolyshkin: This comes from a parent cgroup I guess, and is set by systemd. For me, it's 18930. I haven't checked but believe systemd sets it depending on available RAM. Which probably means your laptop has 32 GB RAM, as mine has humble 16 :smile: 


```console
[kir@kir-rhat cri-o]$ cat /proc/self/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/vte-spawn-00c4c013-d8a2-4bd4-997e-c42605190449.scope
[kir@kir-rhat cri-o]$ cat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/vte-spawn-00c4c013-d8a2-4bd4-997e-c42605190449.scope/pids.max 
18930
```

test/cgroups.bats
```diff
@@ -10,19 +10,47 @@ function teardown() {
...
+@test "zero pids limit" {
+	if ! grep -qEw ^pids /proc/cgroups; then
+		skip "pids cgroup controller is not available"
+	fi
+	pids_root=/sys/fs/cgroup/pids
+	if is_cgroup_v2; then
+		pids_root=/sys/fs/cgroup
+	fi
+
+	# pids_limit of 0 sets the pids.max to that of pid 1
```
kolyshkin, line 51: This is not entirely true -- I think the limit is taken from the parent cgroup, not the one that init belongs to.

It's a coincidence that systemd sets the same limits for all cgroups.

test/cgroups.bats
```diff
@@ -10,19 +10,47 @@ function teardown() {
...
+	parent_limit=$(cat "$pids_root/$(grep pid '/proc/1/cgroup' | tr ':' ' ' | awk '{ printf $3 }')/pids.max")
```
kolyshkin, line 52: 1. This won't work for cgroup v2 as /proc/$PID/cgroup looks like:
```
kir@kir-rhat sys]$ cat /proc/1/cgroup 
0::/init.scope
[kir@kir-rhat sys]$ cat /proc/self/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/vte-spawn-d48a4976-89a3-4aac-96b6-b783b46a6113.scope
```

2. `awk -F : '$2 == "pid" {print $3}' /proc/1/cgroup` does the same as `grep pid '/proc/1/cgroup' | tr ':' ' ' | awk '{ printf $3 }'`

test/cgroups.bats
kolyshkin: I just checked on Ubuntu 20.04, init is not put to a pids cgroup:

```
kir@ubu2004:~$ grep pids /proc/1/cgroup 
8:pids:/
kir@ubu2004:~$ ls -l /sys/fs/cgroup/pids/
total 0
-rw-r--r--  1 root root 0 Mar  3 19:06 cgroup.clone_children
-rw-r--r--  1 root root 0 Mar  3 19:04 cgroup.procs
-r--r--r--  1 root root 0 Mar  3 19:06 cgroup.sane_behavior
-rw-r--r--  1 root root 0 Mar  3 19:06 notify_on_release
-rw-r--r--  1 root root 0 Mar  3 19:06 release_agent
drwxr-xr-x 38 root root 0 Mar  3 19:04 system.slice
-rw-r--r--  1 root root 0 Mar  3 19:06 tasks
drwxr-xr-x  3 root root 0 Mar  3 19:04 user.slice
```

i.e. your method would not work.

I suggest using /proc/self/cgroup instead.

test/cgroups.bats
kolyshkin: And to amend this for cgroupv2 case (not tested)

```sh
	pids_root=/sys/fs/cgroup/pids
	lookfor=pids
￼	if is_cgroup_v2; then
￼		pids_root=/sys/fs/cgroup
		lookfor=""
￼	fi
	path=$pids_root/$(awk -v lookfor="$lookfor" -F : '$2 == lookfor {print $3; exit}' /proc/self/cgroup)/pids.max
```

test/cgroups.bats
haircommander: /proc/self/cgroup fails when I run it locally though, as the container inherits it from systemd, but my user has "max" configured
```
 sudo bats test/cgroups.bats -t -f zero                                                                                                                                                                                                     1..1                                                                                                                                                                                                                                          not ok 1 zero pids limit                                                                                                                                                                                                                      ...                                                                
# 37725 max
```
(that came from me adding an echo before the test)

test/cgroups.bats
haircommander: updated!

test/cgroups.bats
kolyshkin:
```
not ok 4 zero pids limit
# (in test file ./cgroups.bats, line 55)
#   `parent_limit=$(cat "$pids_root$(awk -v lookfor="$lookfor" -F : '$2 == lookfor {print $3; exit}' /proc/1/cgroup)/pids.max")' failed
# cat: /sys/fs/cgroup/pids//pids.max: No such file or directory
```

As I said above, on Ubuntu 20.04, init is not put to a pids cgroup, and this is the reason why it's failing.

One way of figuring out the limit would be querying systemd. Here's an example from a couple of machines:

```console
kir@ubu2004:~$ systemctl show -p DefaultTasksMax
DefaultTasksMax=4592


[kir@kir-rhat cri-o]$ systemctl show -p DefaultTasksMax
DefaultTasksMax=18933
```

haircommander: oh yeah that seems like it's the way, updated

PR #4617, opened 2021-03-03
Add integration test for started/finished container time

saschagrunert:
#### What type of PR is this?

/kind cleanup

This adds an integration test to verify the container start/finished time.

None
None

```release-note
None
```

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -98,13 +98,32 @@ function wait_until_exit() {
 	[ -n "$output" ]
 	echo "$output" | jq -e ".info.privileged == false"
 
+	YEAR=$(date +"%Y")
+	CREATED=$(echo "$output" | jq -re '.status.createdAt')
+	echo "$output" | jq -e '.status.createdAt | startswith("'"$YEAR"'")'
+
+	echo "$output" | jq -e '.status.startedAt | startswith("'"$YEAR"'") | not'
+	echo "$output" | jq -e '.status.finishedAt | startswith("'"$YEAR"'") | not'
+
 	crictl start "$ctr_id"
-	crictl inspect "$ctr_id"
+	output=$(crictl inspect "$ctr_id")
+	[[ "$CREATED" == $(echo "$output" | jq -re '.status.createdAt') ]]
+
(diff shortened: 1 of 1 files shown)


haircommander:
code LGTM, though I'm curious why we're making this change

saschagrunert:
> code LGTM, though I'm curious why we're making this change

It more aligns to dockershim and would make a transition easier, if users parse back the value into golang and call `IsZero` on the time to verify if the field is set or not.

haircommander:
SGTM :)
/approve

giuseppe (approved):
/lgtm

server/container_status.go
```diff
@@ -65,23 +65,17 @@ func (s *Server) ContainerStatus(ctx context.Context, req *types.ContainerStatus
 		cState = c.State()
 	}
 
-	created := c.CreatedAt().UnixNano()
+	resp.Status.CreatedAt = cState.Created.UnixNano()
+	resp.Status.StartedAt = cState.Started.UnixNano()
+	resp.Status.FinishedAt = cState.Finished.UnixNano()
```
Iceber, line 70: It looks like the startedAt and finishedAt timestamp might be `-6795364578871345152`(`time.Time{}.UnixNano()`)
https://github.com/cri-o/cri-o/blob/f47aeb6bf10cc62b4b5af2283fa507ddc5242191/vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L925-L938
`time.Unix(0, -6795364578871345152).IsZero()` is false

Iceber: The current code should be correct

saschagrunert: Ah, yeah thank you for the wrap-up. This was a bit confusing.

saschagrunert: Let's keep the integration test.

saschagrunert:
The overall logic already was fine as @Iceber stated. Anyhow, let's keep the integration test case :)

TomSweeneyRedHat:
LGTM
assuming happy tests.

haircommander:
@saschagrunert still want the hold?

PR #4620, opened 2021-03-04
cri: Devices ownership from SecurityContext

mythi:
#### What type of PR is this?

> /kind feature

CRI container runtimes mount devices (set via kubernetes device plugins)
to containers by taking the host user/group IDs (uid/gid) to the
corresponding container device.
 
This triggers a problem when trying to run those containers with
non-zero (root uid/gid = 0) uid/gid set via runAsUser/runAsGroup:
the container process has no permission to use the device even when
its gid is permissive to non-root users because the container user
does not belong to that group.
  
It is possible to workaround the problem by manually adding the device
gid(s) to supplementalGroups. However, this is also problematic because
the device gid(s) may have different values depending on the workers'
distro/version in the cluster.
    
This patch suggests to take runAsUser/runAsGroup set via SecurityContext
as the device uid/gid, respectively. The feature can enabled by
setting device_ownership_from_security_context config boolean to true.
   
In cri-o, the functionality is limited to Devices set by the device
plugins. The uid/gid for devices set via annotations/cri-o config are not
modified.

Fixes: https://github.com/kubernetes/kubernetes/issues/92211

See the k/k issue for details. I have created a detailed Google doc about the
gap and proposed few alternatives how to address it. This PR implements our
"option A" which is what we prefer.

Our preference is that the gap is also addressed the same way in containerd too.
I have opened a PR with the same implementation there too.

Need help in testing...

```release-note
add `--device-ownership-from-security-context` which allows an admin to specify devices be configured to be owned by the container user and group, rather than unconditionally being root.
```


saschagrunert:
I have different feelings about this. On one hand I think this is a valuable addition when documented and clearly stated to the users how it behaves. On the other hand I think its confusing to mix-in such a change via a field which clearly states that we `runAsUser` and not mount anything as user…

The most sane way would be to add the field to the security context and pass down the information via the CRI. I think you already discussed that this is not a possible move forward, right? So then, I think this is the most feasible way to get the feature in.

mythi:
>  I think its confusing to mix-in such a change via a field which clearly states that we runAsUser and not mount anything as user…

The OCI spec is not clear on this. `Devices {}` can have optional

`uid` (uint32, OPTIONAL) - id of device owner in the container namespace.
`gid` (uint32, OPTIONAL) - id of device group in the container namespace.

and exactly the same notation is used for User:

`uid` (int, REQUIRED) specifies the user ID in the container namespace.
`gid` (int, REQUIRED) specifies the group ID in the container namespace.

It's not mentioned that they should always be different.

> The most sane way would be to add the field to the security context and pass down the information via the CRI.

We did not write a KEP about the `PodSecurityContext` changes. In our container-orchestrated-devices call last week it was concluded that since we don't have anything like `runAsUserDevices` in `PodSecurityContext` bridging the gap at the runtime level is the way to go for now.

pkg/container/device.go
```diff
@@ -6,13 +6,49 @@ import (
...
+	if userval := config.Linux.SecurityContext.RunAsUser; userval != nil {
+		uid := uint32(userval.Value)
+		if uid > 0 && !disableDeviceOwnershipFromSecurityContext(config.Annotations) {
+			return &uid
+		}
+	}
+	return &hostUID
+}
+
+func getDeviceGID(config *types.ContainerConfig, hostGID uint32) *uint32 {
```
haircommander, line 42: nit: can we have these functions defined below the function that calls them?

pkg/container/device.go
```diff
@@ -6,13 +6,49 @@ import (
...
 	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/opencontainers/runc/libcontainer/devices"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 )
 
+const (
+	// kubelet/device-plugins can set this annotation to "true" to opt-out for getting Devices
+	// UID/GID set as RunAsUser/RunAsGroup.
+	disableDeviceOwnershipFromSecurityContextAnnotation = "io.kubernetes.cri.DisableDeviceOwnershipFromSecurityContext"
```
haircommander, line 20: this should go in  `pkg/annotations`

saschagrunert: We should add this annotation to the allowed annotations docs, too.

mythi: OK, will do. Do you have unit tests I should add for this?

mythi:
dropped "rfc" and rebased to master.

pkg/container/device.go
```diff
@@ -163,3 +165,32 @@ func (c *container) specAddContainerConfigDevices() error {
...
 }
+
+func disableDeviceOwnershipFromSecurityContext(annotations map[string]string) bool {
+	if v, ok := annotations[crioann.DisableDeviceOwnershipFromSecurityContextAnnotation]; ok {
+		if v == "true" {
+			return true
+		}
+	}
+	return false
+}
```
haircommander, line 176: we'll also need this annotation to be in gated by allowed annotations:
add a function in `internal/oci/oci.go` similar to `AllowUnifiedCgroupAnnotation` and pass that in to getDeviceUID

pkg/container/device.go
```diff
@@ -163,3 +165,32 @@ func (c *container) specAddContainerConfigDevices() error {
...
+			return &uid
+		}
+	}
+	return &hostUID
+}
+
+func getDeviceGID(config *types.ContainerConfig, hostGID uint32) *uint32 {
+	if groupval := config.Linux.SecurityContext.RunAsGroup; groupval != nil {
+		gid := uint32(groupval.Value)
+		if gid > 0 && !disableDeviceOwnershipFromSecurityContext(config.Annotations) {
```
haircommander, line 191: this value will never change, so we should call disableDeviceOwnershipFromSecurityContext once and pass that value into these functions

pkg/container/device.go
mythi: done. does it look what you meant? I'll continue with more testing if this is what you had in mind.

pkg/container/device.go
mythi: done

pkg/container/device.go
haircommander: yes exactly, thanks!

pkg/container/device.go
```diff
@@ -108,8 +115,8 @@ func (c *container) specAddContainerConfigDevices() error {
 				Type:  string(dev.Type),
 				Major: dev.Major,
 				Minor: dev.Minor,
-				UID:   &dev.Uid,
-				GID:   &dev.Gid,
+				UID:   getDeviceUID(c.Config(), dev.Uid, disableDeviceOwnershipFromSecurityContext),
+				GID:   getDeviceGID(c.Config(), dev.Uid, disableDeviceOwnershipFromSecurityContext),
```
kad, line 119: maybe `dev.Gid`?

mythi: nice catch! fixed

mythi:
@haircommander I rebased this on top of the new 'allowed annotations' functionality.

saschagrunert (commented):
Let's add some docs to the `pkg/config/template.go` `templateStringCrioRuntimeRuntimesRuntimeHandler` string, please.

Otherwise LGTM

mythi:
> Let's add some docs to the pkg/config/template.go templateStringCrioRuntimeRuntimesRuntimeHandler string, please.

done, rebased and force pushed.

saschagrunert (approved):

pkg/annotations/annotations.go
```diff
@@ -16,6 +16,10 @@ const (
 	// DevicesAnnotation is a set of devices to give to the container
 	DevicesAnnotation = "io.kubernetes.cri-o.Devices"
 
+	// DisableDeviceOwnershipFromSecurityContextAnnotation is the K8S annotation used by kubelet/device-plugins
+	// to opt-out getting Devices UID/GID set as RunAsUser/RunAsGroup
+	DisableDeviceOwnershipFromSecurityContextAnnotation = "io.kubernetes.cri.DisableDeviceOwnershipFromSecurityContext"
+
```
haircommander, line 22: I think this should be opt-in to begin. Maybe just have:
`UseDeviceOwnershipFromSecurityContextAnnotation`.

Then the kubelet can pass that down to us, and we only interpret *if* we have that field.

that also means that a user has to specify this annotation in their list of `allowed_annotations` to have cri-o actually interpret it. This means admins will opt-into the special behavior

pkg/container/device.go
```diff
@@ -77,6 +79,8 @@ func (c *container) specAddHostDevicesIfPrivileged(privilegedWithoutHostDevices
 func (c *container) specAddContainerConfigDevices() error {
 	sp := c.Spec().Config
 
+	disableDeviceOwnershipFromSecurityContext := checkDisableDeviceOwnershipFromSecurityContextAnnotation(c.Config().Annotations)
```
haircommander, line 82: if we go with the approach above, then we only have to check if the useDeviceOwnershipFromSecurityContext is in the annotations still here. The filtering of the annotations has already happened, so we know if the annotation exists, we're allowed to use it

haircommander:
sorry @mythi I misunderstood what the PR was doing, I think this should be opt-in, not opt-out.

mythi:
> sorry @mythi I misunderstood what the PR was doing, I think this should be opt-in, not opt-out.

@haircommander I added my thoughts on opt-in/out in https://github.com/kubernetes/kubernetes/issues/92211#issuecomment-826560999

Today, if users set `runAsUser/runAsGroup` to non-zero uid/gid, they cannot use the devices. Therefore, they're forced to run containers as root. To "opt-in" for this feature, they can use non-zero values.

pkg/container/device.go
```diff
@@ -108,8 +112,8 @@ func (c *container) specAddContainerConfigDevices() error {
 				Type:  string(dev.Type),
 				Major: dev.Major,
 				Minor: dev.Minor,
-				UID:   &dev.Uid,
-				GID:   &dev.Gid,
+				UID:   getDeviceUID(c.Config(), dev.Uid, disableDeviceOwnershipFromSecurityContext),
+				GID:   getDeviceGID(c.Config(), dev.Gid, disableDeviceOwnershipFromSecurityContext),
```
haircommander, line 116: nit: we missed a case below (where the mount is a directory of devices). we should probably be consistent
also, in doing so, we should save the device{G,U}Id to not redo the work

haircommander:
sorry for the radio silence @mythi I am trying to make sure we aren't introducing risk by changing this behavior. from my perspective, it's possible node admins would want devices to be restricted to pods that run as root (especially because there exist tools that restrict who can run a pod as root, but not who can not add this annotation).

I am not really sure it's a risk, but it's a case I am worried about. The `allowed_annotation` feature in cri-o gives admins an option to toggle this behavior. I am not yet convinced the better user interface (having cri-o automatically give the correct UID/GID) is worth the risk of allowing users that are allowed to run containers as any user to have access to host devices. 

If a pod author sees their pod is being denied because the user they're using doesn't have access to the device, I don't see it as a particularly high hurdle to ask them to add an annotation to their pod (and hope their admin has enabled this annotation for them).

mythi:
@haircommander containerd folks have concerns too that this breaks things for some users but so far we haven't been able to think any scenario that would be impacted by default.

> If a pod author sees their pod is being denied because the user they're using doesn't have access to the device, I don't see it
> as a particularly high hurdle to ask them to add an annotation to their pod (and hope their admin has enabled this annotation
> for them).

It's not clear to me how CRI-O's annotations looks like I had missed how `allowed_annotations` are expected to be used. As written in the commit message, my vision was that only device-plugins/kubelet would be allowed to set this opt-out annotation. The annotation is from container's spec and not from pods annotation. With that, it sounds this `allowed_annotation` should be dropped completely.

Currently, devices can only be set via device plugins and if the functionality fails for some device, the device plugin can add that annotation to opt-out and there's no user-visible impact.

```
The annotation can be set by kubernetes device plugins or kubelet.
If several device resources are requested, it is enough if one device
plugin sets this annotation to opt-out for the configuration for
*all* the devices added.

In cri-o, the functionality is limited to Devices set by the device
plugins. The GID/UID for devices set via annotations are not modified.
```

haircommander:
We discussed this on the COD call today, and my concerns are resolved.
I was worried about 2 things: admins not being able to restrict access to devices for unprivileged users, and malicious actors gaining access to the kube api and mutating device permissions. The first concern is mitigated with resource quotas set on namespace/user level, and the latter was due to a misunderstanding and is not a risk.

One piece of investigation going forward should be for user namespaces (the security conetxt UID/GID may not be what they are on the host). One of my last nits for this PR is to either document that this behavior is not compatible with user namespaces or determine that it is.

The last nit is to drop this annotation from allowed_annotations. we'll use the native kube resource control mechanisms instead of having admins need to configure for runtime classes.

overall
/approve

thanks for pushing this through and walking me through my concerns @mythi

haircommander:
failures are not the fault of this PR

pkg/container/device.go
```diff
@@ -163,3 +164,41 @@ func (c *container) specAddContainerConfigDevices() error {
...
+// via annotations when pod's securitycontext runs as root/uid=0. When
+// enabled, the logic below does not change the behavior for existing
+// mappings unless container's securitycontext user/group overrides what
+// is set for the pod.
+//
+// Ref: https://github.com/kubernetes/kubernetes/issues/92211
+func getDeviceUID(config *types.ContainerConfig, hostUID uint32, enableDeviceOwnershipFromSecurityContext bool) *uint32 {
+	if userval := config.Linux.SecurityContext.RunAsUser; userval != nil {
+		uid := uint32(userval.Value)
+		if uid > 0 && enableDeviceOwnershipFromSecurityContext {
```
haircommander, line 189: can we refactor this to use a helper function for the common bits? we can pass a UInt64Value field into the helper:
```
func getID(userval UInt64Val, enableDeviceOwnershipFromSecurityContext bool) {
  ...
}
```

mythi: fixed!

haircommander:
hey @mythi here is a 
[patch](https://github.com/cri-o/cri-o/files/6939035/diff.txt)
 I'd appreciate you applying. it should fix the CI woes (I would have had you run it, but the completions/docs generation code is sometimes a pain to run)

mythi:
@haircommander thanks! applied and pushed

haircommander:
/approve

a good follow up could be a unit test, but this has languished for too long, so I don't think it's needed here.

@saschagrunert @wgahnagl @QiWang19 PTAL

mythi:
> a good follow up could be a unit test, but this has languished for too long, so I don't think it's needed here.

@haircommander +1, I can work on this. I have the unit tests in my containerd PR. I just need to take that to cri-o's Ginkgo style testing :-)

mythi:
@haircommander I added the test cases too

saschagrunert (approved):
LGTM

haircommander:
/retest
/lgtm

thanks @mythi for all of your work here

haircommander:
/override ci/prow/e2e-gcp
/override ci/prow/e2e-agnostic

(CI failures unrelated)

PR #4625, opened 2021-03-05
Cleanup portforward streams after their usage

saschagrunert:
#### What type of PR is this?

/kind bug

This implements a stream cleanup when using portforwardings. Before
applying this patch, the `streams  []httpstream.Stream` within
`spdy/connection.go` would fill-up for each streaming request. Now we
use the stream identifier to keep track of them.

None

/hold for obvious vendor changes

We have to decide if we wanna bring those changes into k/k, or if we move the vendored parts into this project (like we did with other parts of the streaming server).

```release-note
Fixed port-forward memory leak for long-running and heavily used connections.
```

### DIFF
--- a/vendor/k8s.io/apimachinery/pkg/util/httpstream/httpstream.go
+++ b/vendor/k8s.io/apimachinery/pkg/util/httpstream/httpstream.go
@@ -78,6 +78,8 @@ type Connection interface {
 	// SetIdleTimeout sets the amount of time the connection may remain idle before
 	// it is automatically closed.
 	SetIdleTimeout(timeout time.Duration)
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
Is there an issue or a PR upstream that you can point to in a comment as a breadcrumb for this change?
Otherwise LGTM
assuming happy tests

saschagrunert:
> Is there an issue or a PR upstream that you can point to in a comment as a breadcrumb for this change?

Unfortunately not, there is only a non-public Bugzilla issue open. :)

saschagrunert:
CI is green, but I think we should fix the issue in k/k.

Issue #4278, opened 2020-10-19
fatal error: runtime·unlock: lock count


nyxi:
**Description**
A few days after migrating to cri-o it crashed on one of our nodes. All nodes are logging a lot of these two warnings:

```
unable to retrieve network namespace link eth0: Link not found
unable to get stats for container b6d170426c4803dd8f8478c45821f9f27840e24c4911269a152b1af4033d3a5a: unable to load cgroup at /kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod4b5fcfe7_a146_4006_870a_2ba4c633694f.slice: cgroup deleted" id=b3a64d8d-3287-4222-942c-6d8019f1449c
```

**Steps to reproduce the issue:**
Can't really say, we were running our normal workload with Rook Ceph, multiple Prometheus instances and a few java applications.

**Describe the results you received:**
crio daemon crashed with:

```
fatal error: runtime·unlock: lock count
panic during panic
runtime stack:
goroutine 0 [runtime.throwidle(]:
0x560457766761fatal error: runtime·unlock: lock count
runtime.throwstack trace unavailable
(fatal error: runtime·unlock: lock count
fatal error: runtime·unlock: lock count
fatal error: runtime·unlock: lock count
```

**Describe the results you expected:**
For crio to not crash

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `crio --version`:**

```
crio version 1.17.5
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

- Virtual machines on Openstack
- OS is Fedora CoreOS 32.20200809.3.0
- Kernel 5.8.10-200.fc32.x86_64
- Kubernetes v1.17.6

haircommander:
can you post the logs leading up to the crash as well?

nyxi:
Sure;
[crio service log](https://github.com/cri-o/cri-o/files/5402878/logs.txt)

haircommander:
can you post the output of `crio config` as well as the contents of `/etc/cni/net.d`? it seems like you're using CRI stats, and not cadvisor stats, and that some configuration of your networking plugins are confusing our CRI stats path

haircommander:
what happens if you restart crio? could you restart crio and attach the full logs if it happens again? specifically, I want to see if it chooses the plugin correctly, and how long it takes for the node to hit this issue again.

this is further confusing because you should be using cadvisor stats, and I've never seen this error before...

nyxi:
[...]

PR #4626, opened 2021-03-05
stats: fix log spam

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

Because we don't lock in stats collection (to keep it speedy), we run into
races with other routines deleting stuff from underneath us.

As such, we should be wary to print errors in these expected situations.

Instead, ignore or reduce the log level for some common errors that are not really worth reporting.

```release-note
none
```

### DIFF
--- a/server/container_stats_list.go
+++ b/server/container_stats_list.go
@@ -1,9 +1,11 @@
 package server
 
 import (
+	"github.com/containers/libpod/v2/pkg/cgroups"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
 	"github.com/cri-o/cri-o/server/cri/types"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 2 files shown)


haircommander:
should fix https://github.com/cri-o/cri-o/issues/4278
refers to https://bugzilla.redhat.com/show_bug.cgi?id=1935461

haircommander:
a more rigorous solution would be to spawn a goroutine that checks whether the resource was actually cleaned up, and if not report the error. Doesn't seem worth the work, though.

haircommander:
checking a CI run: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-federated-results/pr-logs/pull/cri-o_cri-o/4623/test_pull_request_crio_e2e_fedora/22877/artifacts/journals/crio.service

we get 17 "instances of sandbox not found" and 133 "cgroup deleted".

We'll not get the link error because of the bug that some iteration of https://github.com/cri-o/cri-o/pull/4613 should fix

mrunalp (approved):

haircommander:
oopsies, introduced a segfault

server/container_stats_list.go
```diff
@@ -31,13 +33,20 @@ func (s *Server) ListContainerStats(ctx context.Context, req *types.ListContaine
 	for _, container := range ctrList {
 		sb := s.GetSandbox(container.Sandbox())
 		if sb == nil {
-			log.Warnf(ctx, "unable to get stats for container %s: sandbox %s not found", container.ID(), container.Sandbox())
+			// Because we don't lock, we will get situations where the container was listed, and then
+			// its sandbox was deleted before we got to checking its stats.
+			// We should not log in this expected situation.
```
TomSweeneyRedHat, line 38: IDK, any thought on keeping the old log message but converting to a debugf?

TomSweeneyRedHat:
One question for consideration.  
Changes LGTM
assuming happy tests

server/container_stats_list.go
haircommander: I thought about it, but I don't really think it's useful. the sandbox not existing anymore isn't an actionable error as the others *may* be (though likely aren't). If sandboxes are disappearing underneath us, many other things will break much louder

saschagrunert (approved):
/retest

Issue #4621, opened 2021-03-04
[RFE] Do not fail to start if a **non** default runtime is missing


fidencio:
**Description**

The current CRI-O behaviour is to not start if a **non** default runtime is missing.  While it's understandable, for **non** default runtimes we'd be better by:
* Loudly warning that the binary is not present;
* Not registering (or un-registering) the runtimeClass related to the missing binary;

**Steps to reproduce the issue:**
1. Add a second runtime to your crio configuration
2. Ensure CRI-O starts
3. Move the binary related to the configuration you added to a different place (without changing the config)
4. Restart CRI-O
5. See CRI-O failing to start

**Describe the results you received:**
```
mar 04 19:25:11 localhost crio[3425]: time="2021-03-04 19:25:11.294299243Z" level=fatal msg="Validating runtime config: runtime validation: invalid runtime_path for runtime 'kata': \"stat /usr/bin/containerd-shim-kata-v2: no such file or directory\"" file="crio/main.go:335"
```

**Describe the results you expected:**
A simple warning, and the inability to schedule something using the related runtime class;

**Additional information you deem important (e.g. issue happens only occasionally):**
This is an RFE.

**Output of `crio --version`:**

```
crio version 1.20.0-dev
Version:       1.20.0-dev
GitCommit:     a0f9ffc0d8154bd7e05003413c210cf5c756587f
GitTreeState:  dirty
BuildDate:     2021-03-02T08:22:17Z
GoVersion:     go1.14.12
Compiler:      gc
Platform:      linux/amd64
Linkmode:      dynamic
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

PR #4631, opened 2021-03-07
config: Don't fail if the non default runtime doesn't pass validation

fidencio:
#### What type of PR is this?

/kind bug

CRI-O is too strict by simply not starting up in case a **non** default runtime is misconfigured.

Fixes: #4621

I need to get to writing tests to it, but I wasn't successful on a  quick try. :-/

```release-note
Instead of failing to start, CRI-O now only prints a warning and ignores the runtime, in case a **non** default runtime is misconfigured.
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -899,12 +899,24 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 
 // ValidateRuntimes checks every runtime if its members are valid
 func (c *RuntimeConfig) ValidateRuntimes() error {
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
LGTM
/retest

TomSweeneyRedHat:
changes LGTM
assuming happy tests and an added test would be nice If possible.

fidencio:
Okay, I just added some basic unit test for this, and CI seems to be happy.

saschagrunert (approved):
/lgtm

PR #4639, opened 2021-03-09
use (sort of) happy-eyeballs for port-forwarding

aojea:
port-forward needs to open a connection inside the namespace to the localhost address.

localhost can resolve to both IPv4 and IPv6 addresses in dual-stack systems
but the application can be listening in one of the IP families only.

golang has enabled RFC 6555 Fast Fallback (aka HappyEyeballs)
by default in 1.12. It means that if a host resolves to both IPv6 and IPv4,
it will try to connect to any of those addresses and use the
working connection.

However, the implementation uses go routines to start both connections in parallel,
and this has limitations when running inside a namespace, so we try to the connections
serially disabling the Fast Fallback support.
xref golang/go#44922

Signed-off-by: Antonio Ojea <aojea@redhat.com>

#### What type of PR is this?

/kind feature

I've tested it with this snippet

```go
func main() {
	fmt.Println("GOOS:", runtime.GOOS)
	// Resolver
	addrs, err := net.LookupHost("localhost")
	fmt.Println("net.LookupHost addrs:", addrs, "err:", err)

	go listener("127.0.0.1", "8080")
	// go listener("[::1]", "8080")
	time.Sleep(1 * time.Second)

	// Dial
	conn, err := net.Dial("tcp", "localhost:8080")
	if err != nil {
		fmt.Println("Error connecting", err.Error())
		return
	}
	defer conn.Close()
	conn.Write([]byte("SYN"))
	fmt.Println("Connecting to: ", conn.RemoteAddr().String())
	buf := make([]byte, 1024)
	_, err = conn.Read(buf)
	if err != nil {
		fmt.Println("Error reading:", err.Error())
	}
	fmt.Println("Client Received", string(buf))
}

func listener(address, port string) {
	// open a server listening to establish the TCP connections
	listener, err := net.Listen("tcp", address+":"+port)
	if err != nil {
		panic(err)
	}
	defer listener.Close()
	listenerAddress := listener.Addr().String()
	fmt.Println("Listener: ", listenerAddress)

	for {
		conn, err := listener.Accept()
		if err != nil {
			panic(err)
		}

		buf := make([]byte, 1024)
		_, err = conn.Read(buf)
		if err != nil {
			fmt.Println("Error reading:", err.Error())
		}
		fmt.Println("Server Received", string(buf))
		conn.Write([]byte("ACK"))
		conn.Close()
	}
}
```

running it inside a pod
```
# ./resolve_localhost
GOOS: linux
net.LookupHost addrs: [127.0.0.1 ::1] err: <nil>
Listener:  127.0.0.1:8080
DEBUG: primaries [[::1]:8080] fallbacks [127.0.0.1:8080] 1
Connecting to:  127.0.0.1:8080
Server Received SYN
Client Received ACK
```

```release-note
port-forward works for IPv6 only pods
```
xref: https://github.com/kubernetes/kubernetes/issues/99850

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1020,15 +1020,21 @@ func (r *runtimeOCI) PortForwardContainer(ctx context.Context, c *Container, net
 	if err := ns.WithNetNSPath(netNsPath, func(_ ns.NetNS) error {
 		defer stream.Close()
 
-		// TODO: hardcoded to tcp4 because localhost resolves to ::1 by default
-		// if the system has IPv6 enabled. However, not all applications are
-		// listening on the IPv6 localhost address. Theoretically happy
-		// eyeballs will try IPv6 first and fallback to IPv4 but resolving
-		// localhost doesn't seem to return and IPv4 address always, thus
-		// failing the connection.
-		conn, err := net.Dial("tcp4", fmt.Sprintf("localhost:%d", port))
+		// localhost can resolve to both IPv4 and IPv6 addresses in dual-stack systems
+		// but the application can be listening in one of the IP families only.
+		// golang has enabled RFC 6555 Fast Fallback (aka HappyEyeballs) by default in 1.12
+		// It means that if a host resolves to both IPv6 and IPv4, it will try to connect to any
+		// of those addresses and use the working connection.
+		// xref https://github.com/golang/go/commit/efc185029bf770894defe63cec2c72a4c84b2ee9
+		// However, the implementation uses go routines to start both connections in parallel,
+		// and this has limitations when running inside a namespace, so we try to the connections
+		// serially disabling the Fast Fallback support.
+		// xref https://github.com/golang/go/issues/44922
+		var d net.Dialer
+		d.FallbackDelay = -1
+		conn, err := d.Dial("tcp", fmt.Sprintf("localhost:%d", port))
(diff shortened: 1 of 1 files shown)


saschagrunert:
The critest case seems to fail now:
```
Streaming runtime should support streaming interfaces [It] runtime should support portforward [Conformance] 
```

aojea:
> The critest case seems to fail now:
> 
> ```
> Streaming runtime should support streaming interfaces [It] runtime should support portforward [Conformance] 
> ```

/hold

aojea:
I can reproduce it .... it's interesting and can't say to which point is golang or forwarding logic problem
Will keep you posted

aojea:
@saschagrunert it was simpler than expected, critest is passing now

haircommander:
/ok-to-test
/approve

LGTM, thanks @aojea

aojea:
the cgroupv2 job fails with

```
+ true
+ git branch -D pr
error: branch 'pr' not found.
+ true
```

I don't know how to debug the kata job,

are those failures legit or known issues?

saschagrunert:
> are those failures legit or known issues?
Kata is broken right now, see #4648

saschagrunert (approved):
LGTM

saschagrunert:
@aojea please rebase to fix the unit test failures

PR #4642, opened 2021-03-10
Add GitHub API pages support to `get` script

saschagrunert:
#### What type of PR is this?

/kind bug

It may be possible that we do not find a result on the first page. This
means we do now add paging support to the search for a successful GitHub
actions run.

None
None

```release-note
None
```

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -66,7 +66,7 @@ verify_requirements() {
 }
 
 latest_version() {
-    GH_API_URL="https://api.github.com/repos/cri-o/cri-o/actions/runs"
+    GH_API_URL="https://api.github.com/repos/cri-o/cri-o/actions/runs?per_page=100"
 
     GH_HEADERS=(-H "Accept: application/vnd.github.v3+json")
     if [[ $GITHUB_TOKEN != "" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/get
```diff
@@ -75,10 +75,26 @@ latest_version() {
 
     if [[ $VERSION == "" ]]; then
         echo No version provided, finding latest successful GitHub actions run
-        VERSION=$(curl -sSfL "${GH_HEADERS[@]}" "$GH_API_URL" |
-            jq -r '.workflow_runs | map(select(.name == "test" and .head_branch == "master" and .conclusion == "success")) | first | .head_sha')
 
-        echo "Using latest successful GitHub action master: $VERSION"
+        for ((i = 1; i < 10; i++)); do
```
harche, line 79: I have never used GH API, but I was just wondering if they allow us to fetch pages by specifying size? Then how about fetching just one page with large enough page size. That way we avoid the pagination implementation at our side. 

Sorry if this sounds silly, I am just thinking out loud.

harche: Yeah this is cool, this way we don't have to hard code num pages to iterate on.

PR #4644, opened 2021-03-10
pkg/container: take process args

haircommander:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
/kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

with unit tests!

```release-note
none
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -21,6 +21,7 @@ import (
 	crioann "github.com/cri-o/cri-o/pkg/annotations"
 	"github.com/cri-o/cri-o/server/cri/types"
 	"github.com/cri-o/cri-o/utils"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/opencontainers/selinux/go-selinux/label"
@@ -96,6 +97,14 @@ type Container interface {
 
 	// AddUnifiedResourcesFromAnnotations adds the cgroup-v2 resources specified in the io.kubernetes.cri-o.UnifiedCgroup annotation
 	AddUnifiedResourcesFromAnnotations(annotationsMap map[string]string) error
+
+	// SpecSetProcessArgs sets the process args in the spec,
+	// given the image information and passed-in container config
+	SpecSetProcessArgs(imageOCIConfig *v1.Image) error
+
+	// WillRunSystemd checks whether the process args
+	// are configured to be run as a systemd instance.
+	WillRunSystemd() bool
 }
 
 // container is the hidden default type behind the Container interface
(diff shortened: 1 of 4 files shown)


pkg/container/container.go
```diff
@@ -505,3 +514,56 @@ func (c *container) AddUnifiedResourcesFromAnnotations(annotationsMap map[string
...
+
+// SpecSetProcessArgs sets the process args in the spec,
+// given the image information and passed-in container config
+func (c *container) SpecSetProcessArgs(imageOCIConfig *v1.Image) error {
+	// # Start the nginx container using the default command, but use custom
+	// arguments (arg1 .. argN) for that command.
+	// kubectl run nginx --image=nginx -- <arg1> <arg2> ... <argN>
+
+	// # Start the nginx container using a different command and custom arguments.
+	// kubectl run nginx --image=nginx --command -- <cmd> <arg1> ... <argN>
```
saschagrunert, line 526: Do we need those comments?

pkg/container/container.go
```diff
@@ -505,3 +514,56 @@ func (c *container) AddUnifiedResourcesFromAnnotations(annotationsMap map[string
...
+				kubeArgs = imageOCIConfig.Config.Cmd
+			}
+			if kubeCommands == nil {
+				kubeCommands = imageOCIConfig.Config.Entrypoint
+			}
+		}
+	}
+
+	if len(kubeCommands) == 0 && len(kubeArgs) == 0 {
+		return fmt.Errorf("no command specified")
```
saschagrunert, line 545:
```suggestion
		return errors.New("no command specified")
```

pkg/container/container.go
```diff
@@ -505,3 +514,56 @@ func (c *container) AddUnifiedResourcesFromAnnotations(annotationsMap map[string
...
+	var entrypoint string
+	var args []string
+	if len(kubeCommands) != 0 {
+		entrypoint = kubeCommands[0]
+		args = kubeCommands[1:]
+		args = append(args, kubeArgs...)
+	} else {
+		entrypoint = kubeArgs[0]
+		args = kubeArgs[1:]
+	}
```
saschagrunert, line 558: This is implicitely covered by the above statement:
```go
if len(kubeCommands) == 0 && len(kubeArgs) == 0 {
```
otherwise we could run into an index out of range if `len(kubeArgs) == 0 `.

How about combining both, for example:

```suggestion
	} else if len(kubeArgs) != 0 {
		entrypoint = kubeArgs[0]
		args = kubeArgs[1:]
	} else {
		return errors.New("no command specified")
	}
```

haircommander: good idea, updating

saschagrunert (approved):
LGTM

mrunalp (approved):

PR #4650, opened 2021-03-12
do not create & store context in runtime vm

### LINKED ISSUE #4634: runtime VM holds context
which is not idiomatic: https://golang.org/pkg/context/ (it's possible we do it in other places)

we should remove the context internally, and rework the runtime API to be passed a context when it's needed

pperiyasamy:
Signed-off-by: Periyasamy Palanisamy <periyasamy.palanisamy@est.tech>

#### What type of PR is this?

> /kind bug

This PR reuses existing context object and avoids creating a new context object in runtime VM module and storing in its runtimeVM structure.

The runtime VM creates its own context object and store in its structure which shouldn't be done as for crio process should contain only one context object which should be passed across API boundaries.  

Fixes #4634

`streaming#criAdapter{} `needed to store the existing `ctx` object, is that ok ?

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -38,10 +38,8 @@ import (
 type runtimeVM struct {
 	path    string
 	fifoDir string
-
-	ctx    context.Context
-	client *ttrpc.Client
-	task   task.TaskService
+	client  *ttrpc.Client
+	task    task.TaskService
 
 	sync.Mutex
 	ctrs map[string]containerInfo
@@ -71,13 +69,12 @@ func newRuntimeVM(path string, root string) RuntimeImpl {
 	return &runtimeVM{
 		path:    path,
 		fifoDir: filepath.Join(root, "crio", "fifo"),
-		ctx:     context.Background(),
 		ctrs:    make(map[string]containerInfo),
 	}
 }
 
 // CreateContainer creates a container.
-func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (retErr error) {
+func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupParent string) (retErr error) {
 	logrus.Debug("runtimeVM.CreateContainer() start")
 	defer logrus.Debug("runtimeVM.CreateContainer() end")
 
@@ -86,7 +83,7 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (retErr e
 	defer c.opLock.Unlock()
 
 	// First thing, we need to start the runtime daemon
-	if err := r.startRuntimeDaemon(c); err != nil {
+	if err := r.startRuntimeDaemon(ctx, c); err != nil {
 		return err
 	}
 
@@ -136,7 +133,7 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (retErr e
 	defer func() {
 		if retErr != nil {
 			logrus.WithError(err).Warnf("Cleaning up container %s", c.ID())
-			if cleanupErr := r.deleteContainer(c, true); cleanupErr != nil {
+			if cleanupErr := r.deleteContainer(ctx, c, true); cleanupErr != nil {
 				logrus.WithError(cleanupErr).Infof("deleteContainer failed for container %s", c.ID())
 			}
 		}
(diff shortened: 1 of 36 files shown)


internal/oci/runtime_vm.go
```diff
@@ -155,7 +152,7 @@ func (r *runtimeVM) CreateContainer(c *Container, cgroupParent string) (retErr e
 	createdCh := make(chan error)
 	go func() {
 		// Create the container
-		if resp, err := r.task.Create(r.ctx, request); err != nil {
+		if resp, err := r.task.Create(namespaces.WithNamespace(context.Background(), namespaces.Default), request); err != nil {
```
haircommander, line 155: why this change? also, why aren't we passing the ctx to CreateContainer as well?

haircommander:
awesome thanks for taking this on @pperiyasamy , you've got a lint issue and I've got one question, but generally it LGTM

internal/oci/runtime_vm.go
pperiyasamy: yes, missed to replace it with ctx, fixed now. 
ctx is updated with Namespace in startRuntimeDaemon, so its not needed anymore in other places. is my understanding right ?

server/streaming/streaming.go
```diff
@@ -62,8 +63,8 @@ type Server interface {
 
 // Runtime is the interface to execute the commands and provide the streams.
 type Runtime interface {
-	Exec(containerID string, cmd []string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
-	Attach(containerID string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
+	Exec(ctx context.Context, containerID string, cmd []string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
```
pperiyasamy, line 66: while fixing lint in [suite_test.go](https://github.com/cri-o/cri-o/blob/master/server/suite_test.go#L173), this change has a compatibility issue with kubelet interface https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cri/streaming/server.go#L63 as it doesn't pass ctx object, how would like to fix this ?

server/streaming/streaming.go
```diff
@@ -363,6 +364,7 @@ func (s *server) servePortForward(req *restful.Request, resp *restful.Response)
 // criAdapter wraps the Runtime functions to conform to the remotecommand interfaces.
 // The adapter binds the container ID to the container name argument, and the pod sandbox ID to the pod name.
 type criAdapter struct {
+	ctx context.Context
```
pperiyasamy, line 367: @haircommander is it ok to store ctx here ?

server/streaming/streaming.go
haircommander: damn yeah we'll have to drop them here and store them for this...

server/streaming/streaming.go
haircommander: yeah I think we're going to have to. do we create a criAdapter for each new request or for each runtime?

TomSweeneyRedHat:
LGTM
assuming happy tests.

haircommander:
/approve

LGTM, thanks @pperiyasamy !

haircommander:
@mrunalp @saschagrunert @umohnani8 @kolyshkin @giuseppe @QiWang19 PTAL

fidencio:
@pperiyasamy, thanks a lot for the PR. Codewise it looks good to me. 

One thing I'd like to ask you for is, could you squash those 3 commits into a single one?  The reason I'm asking for this is, when we end up bisecting something, we don't end up in a commit breaking the tests, leaving us without knowing whether the commit is the reason of the breakage we're debugging, or just something that had a fix pushed apart.

saschagrunert (approved):
/lgtm

fidencio:
@saschagrunert, sorry, I'm holding this one till we have all patches squashed.  Once we have that, I'll proceed with the `/lgtm`.

fidencio:
@saschagrunert taught me a nice trick: `/label tide/merge-method-squash`, let's see if it works. :-)

fidencio:
oh, too late it seems. :-)

PR #4654, opened 2021-03-12
fix default dependencies on container scopes to fix graceful shutdowns

rphillips:
#### What type of PR is this?
I'm debugging an issue with the Kube API server not shutting down fully on a reboot. I'm not seeing the scope being terminated since DefaultDependencies is disabled. This PR enables the DefaultDependency as documented in the systemd [documentation](https://github.com/systemd/systemd/blob/fee6441601c979165ebcbb35472036439f8dad5f/man/systemd.scope.xml#L70-L85).

/kind bug

```release-note
cri-o managed scopes now depend on the shutdown.target and network.target to allow graceful shutdowns.
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -237,6 +237,7 @@ func (c *container) SpecAddAnnotations(ctx context.Context, sb *sandbox.Sandbox,
 		if systemdHasCollectMode {
 			c.spec.AddAnnotation("org.systemd.property.CollectMode", "'inactive-or-failed'")
 		}
+		c.spec.AddAnnotation("org.systemd.property.DefaultDependencies", "true")
 	}
 
 	if configStopSignal != "" {


rphillips:
these passed, let's try them again

/test e2e-gcp
/test e2e-agnostic

mrunalp:
runc sets DefaultDependencies to false by default. See https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/v1.go#L164

History - https://github.com/opencontainers/runc/commit/99233fde8c4f58853a474a5831ef0bcf6bf866c5

We then added support for setting custom systemd properties that we use for CollectMode. These get added here https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/v1.go#L171

The code in this PR only changes it for the scopes that are started using this utility like conmon. What we want is 
change the property by default for all the containers that we start similar to how we set CollectMode. See https://github.com/cri-o/cri-o/blob/4efb9c28acaca32a72f42f433436e3f29578eea4/pkg/container/container.go#L238

mrunalp (approved):

PR #4662, opened 2021-03-16
Improve image name tag and digest retrieval

saschagrunert:
#### What type of PR is this?

/kind feature

The image result features a name and digest which can be re-used for the
image list result. This means we do not return `<none>:<none>` and
`<none>@<none>` any more in case we do not have any name available. This
aligns with the Kubernetes scheduler which takes the result of the image
list into account.

None
Refers to https://bugzilla.redhat.com/show_bug.cgi?id=1852637

```release-note
Changed the image list result to contain more useful information for the default kube scheduler
```

### DIFF
--- a/server/image_list.go
+++ b/server/image_list.go
@@ -33,16 +33,20 @@ func ConvertImage(from *storage.ImageResult) *types.Image {
 		return nil
 	}
 
-	repoTags := []string{"<none>:<none>"}
+	repoTags := []string{}
 	if len(from.RepoTags) > 0 {
 		repoTags = from.RepoTags
 	} else if from.PreviousName != "" {
 		repoTags = []string{from.PreviousName + ":<none>"}
 	}
 
-	repoDigests := []string{"<none>@<none>"}
-	if len(from.RepoDigests) > 0 {
+	repoDigests := []string{}
+	if len(from.RepoDigests) > 0 { // nolint: gocritic
 		repoDigests = from.RepoDigests
+	} else if from.Name != "" {
+		repoDigests = []string{from.Name + "@" + from.Digest.String()}
+	} else if from.PreviousName != "" {
+		repoDigests = []string{from.PreviousName + "@" + from.Digest.String()}
 	}
 
 	to := &types.Image{
--- a/server/image_list_test.go
+++ b/server/image_list_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/opencontainers/go-digest"
 )
 
 // The actual test suite
@@ -92,10 +93,8 @@ var _ = t.Describe("ImageList", func() {
 
 			// Then
 			Expect(result).NotTo(BeNil())
-			Expect(result.RepoTags).To(HaveLen(1))
-			Expect(result.RepoTags).To(ContainElement("<none>:<none>"))
-			Expect(result.RepoDigests).To(HaveLen(1))
-			Expect(result.RepoDigests).To(ContainElement("<none>@<none>"))
+			Expect(result.RepoTags).To(HaveLen(0))
+			Expect(result.RepoDigests).To(HaveLen(0))
 		})
 
 		It("should succeed with repo tags and digests", func() {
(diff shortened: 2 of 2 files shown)


haircommander:
/approve

@vrothberg @umohnani8 @mtrmac PTAL

server/image_list.go
```diff
@@ -33,16 +33,20 @@ func ConvertImage(from *storage.ImageResult) *types.Image {
...
 		repoTags = []string{from.PreviousName + ":<none>"}
 	}
 
-	repoDigests := []string{"<none>@<none>"}
-	if len(from.RepoDigests) > 0 {
+	repoDigests := []string{}
+	if len(from.RepoDigests) > 0 { // nolint: gocritic
 		repoDigests = from.RepoDigests
+	} else if from.Name != "" {
+		repoDigests = []string{from.Name + "@" + from.Digest.String()}
```
mtrmac, line 47: If the same image is pulled from different locations, and those locations have different digests, can’t this return an impossible name@digest combination (after pulling repo1@digest1 and repo2@digest2, returning a repo1@digest2 which does not physically exist)?

server/image_list.go
```diff
@@ -33,16 +33,20 @@ func ConvertImage(from *storage.ImageResult) *types.Image {
...
 	} else if from.PreviousName != "" {
 		repoTags = []string{from.PreviousName + ":<none>"}
 	}
 
-	repoDigests := []string{"<none>@<none>"}
-	if len(from.RepoDigests) > 0 {
+	repoDigests := []string{}
+	if len(from.RepoDigests) > 0 { // nolint: gocritic
 		repoDigests = from.RepoDigests
+	} else if from.Name != "" {
```
mtrmac, line 46: AFAICS `from.Name` is a name:tag or a name@digest, so this is invalid to concatenate with a @ digest.

server/image_list.go
mtrmac, line 47: `imageService.makeRepoDigests` has already made a similar (and much more involved) computation co create `from.RepoDigests`. How does this make a difference?

server/image_list.go
```diff
@@ -33,16 +33,20 @@ func ConvertImage(from *storage.ImageResult) *types.Image {
...
+		repoDigests = []string{from.Name + "@" + from.Digest.String()}
+	} else if from.PreviousName != "" {
+		repoDigests = []string{from.PreviousName + "@" + from.Digest.String()}
```
mtrmac, line 49: Same question about impossible name@digest pairs.

mtrmac (commented):
Is there a place that documents the overall situation?

I realize I have been arguing against these `<none>` entries back in #2455, but they have been added for a reason ( https://github.com/cri-o/cri-o/issues/2454 ), and so far this PR description just talks about the one place that does not want `<none>` entries and does not talk about the place that does want them (#2454 says empty values cause validation errors). Why is that requirement no longer applicable, or OK to break?

saschagrunert:
> Is there a place that documents the overall situation?

Yep, it's mainly written down there: https://bugzilla.redhat.com/show_bug.cgi?id=1852637

> I realize I have been arguing against these `<none>` entries back in #2455, but they have been added for a reason ( #2454 ), and so far this PR description just talks about the one place that does not want `<none>` entries and does not talk about the place that does want them (#2454 says empty values cause validation errors). Why is that requirement no longer applicable, or OK to break?

To be honest the same requirement still applies, my goal was to find a middle-way to provide more information for the scheduler, while still preserving seeing the information via tools like crictl. I think removing the `<none>:<none>`/`<none>@<none>` is a breaking change, because user may wanna delete those image IDs…

From a scheduler perspective we may already be fine, right? Every `<none>` will be ignored in any case, so there should be no information loss.

saschagrunert:
Unfortunately I don't think that this is the right approach and the current implementation reflects the on-disk state as best as we can. 
/close

mtrmac:
> To be honest the same requirement still applies, my goal was to find a middle-way to provide more information for the scheduler, while still preserving seeing the information via tools like crictl. I think removing the `<none>:<none>`/`<none>@<none>` is a breaking change, because user may wanna delete those image IDs…

Those _names_ are not very useful for deleting because there may be multiple images matching (OTOH it’s reasonable to argue that if the user decides to clean up, the goal will probably be to delete all of them anyway). Is _presence_ of the names necessary to allow deleting images?

PR #4673, opened 2021-03-18
Do not return `<none>` placeholders for images any more

saschagrunert:
#### What type of PR is this?

/kind feature

We now plainly use the `RepoTags` and `RepoDigests` from the returned
image result, with one exception: If there is a previous image name
available, then we use it together with the `Digest` to let tools like
`crictl` be able to display them together with the `<none>` placeholder.

Follow-up of https://github.com/cri-o/cri-o/pull/4662

@mtrmac please take a look

```release-note
CRI-O does not return `<none>` placeholders for missing tags and digests any more when doing a `ListImages` RPC
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,11 @@ default UID mappings (format <container>:<host>:<size>):
 
 Please refer to the [CRI-O Metrics guide](tutorials/metrics.md).
 
+#### Container Runtime Interface special cases
+
+Some aspects of the Container Runtime are worth some additional explanation.
+These details are summarized in a [dedicated guide](cri.md).
+
 ## Adopters
 
 An incomplete list of adopters of CRI-O in production environments can be found [here](ADOPTERS.md).
--- a/cri.md
+++ b/cri.md
@@ -1,0 +1,148 @@
+# Container Runtime Interface special cases
+
+The target of this document is to outline corner cases and common pitfalls in
+conjunction with the Container Runtime Interface (CRI). This document outlines
+CRI-O's interpretation of certain aspects of the interface, which may not be 
+completely formalized.
+
+The main documentation of the CRI can be found [in the corresponding protobuf
+definition][0], whereas this document follows it on the `service`/`rpc` level.
+
+## `ListImages`
+
+`ListImages` lists existing images. Its response consists of an array of
+`Image` types. Besides other information, an `Image` contains `repo_tags` and
+`repo_digests`, which are defined as:
+
+```proto
+// Other names by which this image is known.
+repeated string repo_tags = 2;
+
+// Digests by which this image is known.
+repeated string repo_digests = 3;
+```
+
+Both tags and digests will be used by:
+
+- The kubelet, which displays them in the node status as a flat list, for example:
+
+  ```
+  > kubectl get node 127.0.0.1 -o json | jq .status.images
+  ```
+
+  ```json
+  [
+    {
+      "names": [
+        "k8s.gcr.io/pause@sha256:4a1c4b21597c1b4415bdbecb28a3296c6b5e23ca4f9feeb599860a1dac6a0108",
+        "k8s.gcr.io/pause@sha256:927d98197ec1141a368550822d18fa1c60bdae27b78b0c004f705f548c07814f",
+        "k8s.gcr.io/pause:3.2"
+      ],
+      "sizeBytes": 688049
+    }
+  ]
+  ```
+
+  Right now, the amount of images shown is limited by the kubelet flag
+  `--node-status-max-images` (). The scheduler uses this list to
+  score nodes based on the information if a container image already exists.
+
+- crictl, which is able to output the image list in a human readable way:
+  ```
+  > sudo crictl images --digests
+  IMAGE                 TAG       DIGEST           IMAGE ID         SIZE
+  k8s.gcr.io/pause      3.2       4a1c4b21597c1    80d28bedfe5de    688kB
+  ```
+
+CRI-O implements the [`ConvertImage`][1] function to follow the following self-defined
+rules:
+
+- always return at least one `repo_digests` value
+- return zero or more `repo_tags` values
+
+There are multiple use-cases where this behavior is relevant. Those will be
+covered separately by using real world examples.
+
+### Pulling an image from a remote registry
+
+This is the standard behavior and already shown in the `pause` image example
+above. `crictl` is able to display all information, like the image name, tag and
+digest. There are multiple digests available for this image, which gets a
+correct representation within the `kubelet`'s node status.
+
+### Pulling an updated version of an image with the same tag
+
+Let's assume we pulled the image `quay.io/saschagrunert/hello-world` and
+afterwards its `latest` tag got updated. Now we pull the image again, which
+results in untagging the local image in favor of the new remote one.
+
+CRI-O would now have no available `RepoTags` nor `RepoDigests` within the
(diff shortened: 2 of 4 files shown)


server/image_list.go
```diff
@@ -33,16 +33,17 @@ func ConvertImage(from *storage.ImageResult) *types.Image {
...
-	repoTags := []string{"<none>:<none>"}
+	repoTags := []string{}
+	repoDigests := []string{}
+
 	if len(from.RepoTags) > 0 {
 		repoTags = from.RepoTags
-	} else if from.PreviousName != "" {
-		repoTags = []string{from.PreviousName + ":<none>"}
+	} else if from.PreviousName != "" && from.Digest != "" {
+		repoDigests = []string{from.PreviousName + "@" + string(from.Digest)}
```
saschagrunert, line 42: If we build a container image with the same tag two times in a row locally, then the first built image has:
- no repo tags set
- no repo digests set
- a previous name set
- a Digest set
Using the previous name and the `from.Digest` can then be used to assemble a single repoDigest together.

WDYT @mtrmac ?

mtrmac: Intuitively this looks good — but it’s just a guess, ultimately I don’t think we can avoid tracing down all the users, collecting the full set of constraints, and documenting what we are doing and how it fits together (and how can future users work with the API as intended). It’s the only way to escape the “here’s a test case that breaks and here’s a PR that fixes that one test case” treadmill.

- Does a built image _always_ have a name with this logic? (A plain `buildah commit` just returns an ID IIRC.) If not, isn’t there a caller that needs updating to handle that? (https://github.com/cri-o/cri-o/issues/2454 , or maybe the crictl code?)
-  (For me personally, this is a bit hard to follow.
   ```go
   repoTags = from.RepoTags // or just inline this into the `types.Image{}` constructor
   
   if len(from.RepoTags) == 0 && from.PreviousName… {
      repoDigests = []string{}
   }
   if len(from.RepoDigests) > 0 {
     repoDigests = append(repoDigests, from.RepoDigests...)
   }
   ```
   would feel more natural, one “paragraph” per goal.
- Actually, should we be using the PreviousName value if RepoDigests is set at all?

saschagrunert: Thank you, I simplified the `if` `else` paths and moved the `PreviousName` part to the actual digest handling. I added documentation of all necessary use cases and linked it from the corresponding sections.

README.md
```diff
@@ -204,6 +204,11 @@ default UID mappings (format <container>:<host>:<size>):
 
 Please refer to the [CRI-O Metrics guide](tutorials/metrics.md).
 
+#### Container Runtime Interface special cases
+
+Some aspects of the Container Runtime are worth some additional explanation.
+We summarized those information in a [dedicated guide](cri.md).
```
haircommander, line 210:
```suggestion
These details are summarized in a [dedicated guide](cri.md).
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+conjunction with the Container Runtime Interface (CRI). This document outlines
+CRI-O's interpretation of certain aspects of the interface, which may be not
+completely formalized.
+
+The main documentation of the CRI can be found [in the corresponding protobuf
+definition][0], whereas this document follows it on the `service`/`rpc` level.
+
+## `ListImages`
+
+`ListImages` should lists existing images. Its response consists of an array of
```
haircommander, line 13:
```suggestion
`ListImages` lists existing images. Its response consists of an array of
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+`Image` types. Beside other information, an `Iamge` contains `repo_tags` and
```
haircommander, line 14:
```suggestion
`Image` types. Besides other information, an `Image` contains `repo_tags` and
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+        "k8s.gcr.io/pause@sha256:4a1c4b21597c1b4415bdbecb28a3296c6b5e23ca4f9feeb599860a1dac6a0108",
+        "k8s.gcr.io/pause@sha256:927d98197ec1141a368550822d18fa1c60bdae27b78b0c004f705f548c07814f",
+        "k8s.gcr.io/pause:3.2"
+      ],
+      "sizeBytes": 688049
+    }
+  ]
+  ```
+
+  Right now the amount of images is limited by the kubelet flag
```
haircommander, line 46:
```suggestion
  Right now, the amount of images is limited by the kubelet flag
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+  score nodes based on the information if a container image already exists.
+
+- crictl, which is able to output the image list in a human readable way:
+  ```
+  > sudo crictl images --digests
+  IMAGE                 TAG       DIGEST           IMAGE ID         SIZE
+  k8s.gcr.io/pause      3.2       4a1c4b21597c1    80d28bedfe5de    688kB
+  ```
+
+CRI-O implement the [`ConvertImage`][1] function to follow those self-defined
```
haircommander, line 57:
```suggestion
CRI-O implements the [`ConvertImage`][1] function to follow the following self-defined
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+rules:
+
+- always return at least one `repo_digests` value
+- return zero or more `repo_tags` values
+
+There are multiple use cases where this behavior is relevant. Those will be
```
haircommander, line 63:
```suggestion
There are multiple use-cases where this behavior is relevant. Those will be
```

haircommander:
awesome doc! I learned a lot

saschagrunert:
> awesome doc! I learned a lot

Thanks! Fixed review as suggested. PTAL again

cri.md
```diff
@@ -0,0 +1,148 @@
+# Container Runtime Interface special cases
+
+Target of this document is to outline corner cases and common pitfalls in
```
TomSweeneyRedHat, line 3:
```suggestion
The target of this document is to outline corner cases and common pitfalls in
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+conjunction with the Container Runtime Interface (CRI). This document outlines
+CRI-O's interpretation of certain aspects of the interface, which may be not
```
TomSweeneyRedHat, line 5:
```suggestion
CRI-O's interpretation of certain aspects of the interface, which may not be 
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+
+```proto
+// Other names by which this image is known.
+repeated string repo_tags = 2;
+
+// Digests by which this image is known.
+repeated string repo_digests = 3;
+```
+
+Both, tags and digests, will be used by:
```
TomSweeneyRedHat, line 25:
```suggestion
Both tags and digests will be used by:
```

cri.md
```diff
@@ -0,0 +1,148 @@
...
+
+- The kubelet, which displays them in the node status as flat list, for example:
```
TomSweeneyRedHat, line 27:
```suggestion
- The kubelet, which displays them in the node status as a flat list, for example:
```

saschagrunert:
Thanks for the review @TomSweeneyRedHat. Applied the suggestions. :pray:

TomSweeneyRedHat:
LGTM
thanks @saschagrunert !

mtrmac (commented):
LGTM. Thanks for writing that down!

saschagrunert:
Thanks! @haircommander can we get this in?

PR #4677, opened 2021-03-19
refactor handling of allowed_annotations

haircommander:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
/kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

The code needed to support a new allowed_annotation was both difficult to maintain and inefficient.

instead of checking we allow an annotation (by searching through a vector each time),
filter out the disallowed annotations of the container and sandbox configs.

Now, if the container or pod still has the annotation, we know we can operate on it

```release-note
none
```


haircommander:
@giuseppe @cynepco3hahue @fromanirh PTAL

cynepco3hahue (commented):
Can you please add an example of usage under the PR description, just for documentation purposes?

pkg/config/config.go
```diff
@@ -1069,6 +1072,26 @@ func (r *RuntimeHandler) ValidateRuntimeType(name string) error {
...
+		disallowedAnnotations[ann] = struct{}{}
+	}
+	for _, allowed := range r.AllowedAnnotations {
+		if _, ok := disallowedAnnotations[allowed]; !ok {
+			return errors.Errorf("invalid allowed_annotation: %s", allowed)
+		}
+		delete(disallowedAnnotations, allowed)
+	}
+	for ann := range disallowedAnnotations {
+		r.DisallowedAnnotations = append(r.DisallowedAnnotations, ann)
```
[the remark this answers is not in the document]
haircommander: disallowed annotations is a map, so it can't be used to append to a slice :\

haircommander:
this PR is not updating against my branch, opening a new one

internal/oci/oci.go
```diff
@@ -191,65 +191,19 @@ func (r *Runtime) PrivilegedWithoutHostDevices(handler string) (bool, error) {
...
-		return false, err
+		return nil, err
 	}
-	for _, ann := range rh.AllowedAnnotations {
-		if ann == annotation {
-			return true, nil
+	for _, ann := range annotations {
+		for _, disallowed := range rh.DisallowedAnnotations {
+			if strings.HasPrefix(ann, disallowed) {
+				delete(annotations, disallowed)
```
fromanirh, line 202: is it ok to mutate the map given as argument?

server/container_create_linux.go
```diff
@@ -585,6 +571,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 			}
 		}
 	}()
+
+	// TODO: eventually, this should be in the container package, but it's going through a lot of churn
+	// and SpecAddAnnotations is already passed too many arguments
+	ctr.Config().Annotations, err = s.Runtime().FilterDisallowedAnnotations(sb.RuntimeHandler(), ctr.Config().Annotations)
```
fromanirh, line 577: Ok, this is probably not a big deal, but this invocation seems to suggest `FilterDisallowedAnnotations` is NOT mutating its argument - because we assign the return value. But it seems `FilterDisallowedAnnotations` is actually mutating the argument? Do we want to make this behaviour more explicit? perhaps either returning a copy or returning only error, so it's very clear the function mutates its argument.

fromanirh (commented):
The intent is clear and makes sense to me, unfortunately I'm not familiar enough with the codebase. Added a comment though.

PR #4680, opened 2021-03-22
refactor handling of allowed_annotations

haircommander:
The code needed to support a new allowed_annotation was both difficult to maintain and inefficient.

instead of checking we allow an annotation (by searching through a vector each time),
filter out the disallowed annotations of the container and sandbox configs.

Now, if the container or pod still has the annotation, we know we can operate on it.

An example of adding a new AllowedAnnotation:
define the annotation in `pkg/annotations`, and append it to the AllAllowedAnnotations slice
for any experimental feature, all one has to do is check that the annotation is in the sandbox's or container's
list of annotations (instead of also checking it is allowed)

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
/kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

```release-note
none
```

### DIFF
--- a/internal/runtimehandlerhooks/runtime_handler_hooks.go
+++ b/internal/runtimehandlerhooks/runtime_handler_hooks.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
-	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
 	crioann "github.com/cri-o/cri-o/pkg/annotations"
 	libconfig "github.com/cri-o/cri-o/pkg/config"
@@ -17,13 +16,8 @@ type RuntimeHandlerHooks interface {
 }
 
 // GetRuntimeHandlerHooks returns RuntimeHandlerHooks implementation by the runtime handler name
-func GetRuntimeHandlerHooks(ctx context.Context, config *libconfig.Config, handler string, annotations map[string]string) (RuntimeHandlerHooks, error) {
-	if strings.Contains(handler, HighPerformance) {
-		log.Warnf(ctx, "The usage of the handler %q without adding high-performance feature annotations under allowed_annotations will be deprecated under 1.21", HighPerformance)
-		return &HighPerformanceHooks{config.IrqBalanceConfigFile}, nil
-	}
+func GetRuntimeHandlerHooks(config *libconfig.Config, annotations map[string]string) (RuntimeHandlerHooks, error) {
 	if highPerformanceAnnotationsSpecified(annotations) {
-		log.Warnf(ctx, "The usage of the handler %q without adding high-performance feature annotations under allowed_annotations will be deprecated under 1.21", HighPerformance)
 		return &HighPerformanceHooks{config.IrqBalanceConfigFile}, nil
 	}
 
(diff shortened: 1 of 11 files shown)


haircommander:
replaces https://github.com/cri-o/cri-o/pull/4677

server/container_create_linux.go
```diff
@@ -585,6 +571,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 			}
 		}
 	}()
+
+	// TODO: eventually, this should be in the container package, but it's going through a lot of churn
+	// and SpecAddAnnotations is already passed too many arguments
+	ctr.Config().Annotations, err = s.Runtime().FilterDisallowedAnnotations(sb.RuntimeHandler(), ctr.Config().Annotations)
```
fromanirh, line 577: Ok, this is probably not a big deal, but this invocation seems to suggest FilterDisallowedAnnotations is NOT mutating its argument - because we assign the return value. But it seems FilterDisallowedAnnotations is actually mutating the argument? Do we want to make this behaviour more explicit? perhaps either returning a copy or returning only error, so it's very clear the function mutates its argument.

fromanirh (commented):
(copied from https://github.com/cri-o/cri-o/pull/4677#pullrequestreview-617834804)
The intent is clear and makes sense to me, unfortunately I'm not familiar enough with the codebase. Added a comment though.

server/container_create_linux.go
haircommander: alright, I made this more explicit :)

mrunalp:
@kolyshkin @saschagrunert ptal

saschagrunert (approved):
LGTM

mrunalp:
Needs rebase

internal/runtimehandlerhooks/runtime_handler_hooks.go
```diff
@@ -16,39 +17,26 @@ type RuntimeHandlerHooks interface {
...
-
-	if allowAnnotations {
+func GetRuntimeHandlerHooks(ctx context.Context, config *libconfig.Config, handler string, annotations map[string]string) (RuntimeHandlerHooks, error) {
+	if strings.Contains(handler, HighPerformance) {
+		log.Warnf(ctx, "The usage of the handler %q without adding high-performance feature annotations under allowed_annotations will be deprecated under 1.21", HighPerformance)
 		return &HighPerformanceHooks{config.IrqBalanceConfigFile}, nil
 	}
-
-	if !allowAnnotations && strings.Contains(handler, HighPerformance) {
+	if highPerformanceAnnotationsSpecified(annotations) {
```
cynepco3hahue, line 25: I think it should be `!highPerformanceAnnotationsSpecified`

cynepco3hahue (commented):
Some small comment, except it lgtm

internal/runtimehandlerhooks/runtime_handler_hooks.go
haircommander: I actually dropped that case, as we're preparing for 1.21 anyway

haircommander:
done in github.com/cri-o/cri-o/pull/4725

PR #4681, opened 2021-03-23
WIP: Add support for workload settings

mrunalp:

#### What type of PR is this?

/kind feature

This PR adds support for specifying settings for workload types.
The workloads types are identified by labels (todo).

```release-note
Add support for workload settings
```

### DIFF
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -345,6 +345,18 @@ allowed_annotations = [
 # Kata Containers with the Firecracker VMM
 #[crio.runtime.runtimes.kata-fc]
 
+#[crio.runtime.workloads.workload-type]
+# cpu_set = "0-1"
+# label = "workload-type.crio.io"
+# cpu_shares_annotation = "cpus.workload-type.crio.io"
+# Where:
+# cpu_set is the CPU set for the workload type
+# label is used to identify the workload type
+# cpu_shares_annotation is the prefix to use for setting cpu shares
+# for the containers in the pod. For the example above, there could be
+# multiple annotations for each container in the pod in the form
+# "cpus.workload-type.crio.io/{container_name}"
+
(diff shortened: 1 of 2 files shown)


pkg/config/template.go
```diff
@@ -345,6 +345,18 @@ allowed_annotations = [
 # Kata Containers with the Firecracker VMM
 #[crio.runtime.runtimes.kata-fc]
 
+#[crio.runtime.workloads.workload-type]
+# cpu_set = "0-1"
+# label = "workload-type.crio.io"
+# cpu_shares_annotation = "cpus.workload-type.crio.io"
+# Where:
+# cpu_set is the CPU set for the workload type
+# label is used to identify the workload type
```
saschagrunert, line 354: Can we remove the label and only use the `cpu_shares_annotation` for identifying that the workload option should apply?

haircommander: I think it could be useful to have a generic label `workload-type.crio.io` and a supported_labels slice `["cpus"]` 

then the full annotation would be constructed as:
```go
for _, label := range SupportedLabels {
 checkForAnnotation(container, label + genericLabel + "/" + ctr.Name()
}
```

pkg/config/template.go
```diff
@@ -345,6 +345,18 @@ allowed_annotations = [
 # Kata Containers with the Firecracker VMM
 #[crio.runtime.runtimes.kata-fc]
 
+#[crio.runtime.workloads.workload-type]
+# cpu_set = "0-1"
+# label = "workload-type.crio.io"
```
haircommander, line 350: nit: it'd be easier to check against if this was
`io.crio.workload-type` 
for instance the cpu one could be 
`io.crio.workload-type.cpus/container`, which would still allow us to check strings.Prefix(label) instead of having to check for a substring

PR #4683, opened 2021-03-23
Bump static binary dependency versions

saschagrunert:
#### What type of PR is this?

/kind dependency-change

#### What this PR does / why we need it:
This updates available dependencies for the static binary bundle.

#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4687, opened 2021-03-24
Add static bundle node e2e tests to GitHub actions

saschagrunert:
#### What type of PR is this?

/kind ci

This adds an additional e2e NodeConformance test suite for our static
binary bundle.

Fixes https://github.com/cri-o/cri-o/issues/4688
None

```release-note
None
```

### DIFF
--- a/contrib/bundle/test-e2e
+++ b/contrib/bundle/test-e2e
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Check for root
+if [[ $EUID -ne 0 ]]; then
+    echo "Please run this script as root"
+    exit 1
+fi
+
+# Assume we're running on this arch
+ARCH=amd64
+
+# Global vars to be used
+# shellcheck source=vars
+source "$(dirname "${BASH_SOURCE[0]}")"/vars
+
+pushd "$ARCHIVE_PATH"
+
+# Untar the bundle
+tar xfvz "$ARCHIVE"
+pushd cri-o
+
+# Install and prepare config
+./install
+
+# Use runc as default runtime
+rm /etc/crio/crio.conf.d/10-crun.conf
+cat <<EOT >>/etc/crio/crio.conf.d/10-config.conf
+[crio.runtime]
+log_level = "debug"
+
+[crio.runtime.runtimes.runc]
+runtime_path = "/usr/local/bin/runc"
+EOT
+
+runc --version
+conmon --version
+
+# Start CRI-O
+systemctl daemon-reload
+systemctl start crio
+systemctl is-active --quiet crio && echo CRI-O is running
+journalctl --no-pager -u crio
+
+# Start Kubernetes
+GOROOT="$GOROOT_1_16_X64"
+GOPATH="$HOME/go"
+PATH="$GOROOT/bin:$PATH"
+K8SPATH="$GOPATH/src/k8s.io"
+mkdir -p "$K8SPATH"
+pushd "$K8SPATH"
+git clone --depth=1 https://github.com/kubernetes/kubernetes
+pushd kubernetes
+
+LATEST=$(curl -sSfL https://dl.k8s.io/ci/latest.txt)
+echo "Using Kubernetes build $LATEST"
+
+URL="https://dl.k8s.io/ci/$LATEST"
+mkdir -p _output/bin
+
+echo Downloading server binaries
+curl -sSfL "$URL/kubernetes-server-linux-amd64.tar.gz" -o- | tar xfz -
+cp kubernetes/server/bin/{kubectl,kube-apiserver,kube-controller-manager,kube-proxy,kube-scheduler,kubelet} \
+    _output/bin
+
+echo Downloading test binaries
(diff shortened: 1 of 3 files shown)


contrib/bundle/test-e2e
```diff
@@ -0,0 +1,103 @@
...
+pushd "$ARCHIVE_PATH"
+
+# Untar the bundle
+tar xfvz "$ARCHIVE"
+pushd cri-o
+
+# Install and prepare config
+./install
+
+curl https://raw.githubusercontent.com/cri-o/cri-o/master/scripts/get | bash
```
haircommander, line 26: part of the intention of the issue was to catch issues before we merge. Unfortunately, this seems to only be able to catch them after. We do get a signal after the fact (tests start failing, we know which commit did it), but we don't prevent prow from going red upstream, which looks bad for the stability of the test.

I personally don't know how to inject the current binary somewhere, then have this pull it from there, but ideally that's what would happen here. does that seem reasonable?

saschagrunert: This line has to be removed, we pass the binary artifact during the build. Thanks for the hint!

contrib/bundle/test-e2e
```diff
@@ -0,0 +1,104 @@
...
+    _output/bin
+
+echo Downloading test binaries
+LATEST=$(curl -sSfL https://dl.k8s.io/ci/latest.txt)
+curl -sSfL "https://dl.k8s.io/ci/$LATEST/kubernetes-test-linux-amd64.tar.gz" \
+    -o- | tar xfz -
+cp kubernetes/test/bin/{ginkgo,e2e.test} _output/bin
+
+export CONTAINER_RUNTIME=remote
+export CGROUP_DRIVER=cgroupfs
```
harche, line 64: should be `systemd`?

contrib/bundle/test-e2e
```diff
@@ -0,0 +1,104 @@
...
+export KUBECONFIG=$KUBERUN/admin.kubeconfig
+export PATH="$GOPATH/src/k8s.io/kubernetes/_output/local/bin/linux/amd64:$PATH"
+export KUBE_MASTER_URL=$IP
+export KUBE_MASTER_IP=$IP
+export KUBE_MASTER=$IP
+
+export GINKGO_PARALLEL_NODES=4
+export GINKGO_PARALLEL=y
+
+_output/bin/e2e.test --provider=local --host="https://$IP:6443" \
```
harche, line 102: node e2e tests can also be run like this, https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/e2e-node-tests.md

harche: Although invoking with `e2e.test` might work, I have never tried that way.

contrib/bundle/test-e2e
harche: Complete set of node args we used in the prow job for node e2e, https://github.com/kubernetes/test-infra/blob/927be10bc7aee8a8eed3d97cd4286ccd09dea4cb/config/jobs/kubernetes/sig-node/sig-node-presubmit.yaml#L334

contrib/bundle/test-e2e
harche: Assuming crio is up and running, k/k is cloned. I presume the command to invoke the test locally with crio could look something like,

```
export KUBE_SSH_USER=core && make test-e2e-node RUNTIME=remote CONTAINER_RUNTIME_ENDPOINT="unix:///var/run/crio/crio.sock" FOCUS="\[NodeConformance\]|\[NodeFeature:.+\]" SKIP="\[Flaky\]|\[Slow\]|\[Serial\]" TEST_ARGS='--kubelet-flags="--cgroup-driver=systemd --cgroups-per-qos=true --cgroup-root=/ --runtime-cgroups=/system.slice/crio.service --non-masquerade-cidr=0.0.0.0/0" --extra-log="{\"name\": \"crio.log\", \"journalctl\": [\"-u\", \"crio\"]}"'
```

saschagrunert: My goal was to avoid the build time and download the artifacts from the CI bucket. This way we can re-use them directly by `hack/local-up-cluster.sh`.

contrib/bundle/test-e2e
saschagrunert: Tried to set the args in an additional commit.

contrib/bundle/test-e2e
harche: Makes sense.

haircommander:
nice LGTM,

can you do me a favor and revert https://github.com/cri-o/cri-o/pull/4683 to verify this is working as expected?

saschagrunert:
> nice LGTM,
> 
> can you do me a favor and revert #4683 to verify this is working as expected (as in, after reverting the test should fail)?

Yes, here is the run with runc rc92: https://github.com/cri-o/cri-o/runs/2200669742?check_suite_focus=true

Now, dropping the commit again should make the pipeline green.

TomSweeneyRedHat:
LGTM

haircommander:
@saschagrunert would you mind adding this to the conmon repo next? :smile:

saschagrunert:
> @saschagrunert would you mind adding this to the conmon repo next?

I think this would work, yes! :+1:

PR #4690, opened 2021-03-25
Add container out of memory metric

saschagrunert:
#### What type of PR is this?

/kind feature

This patch adds a new metric
`container_runtime_crio_containers_oom_total` and
`container_runtime_crio_containers_oom`, which collects out of memory
(oom) metrics global and by container name. This also includes
sandboxes, since we reuse the CRI-O internal name for every container.

None
None

```release-note
Added `container_runtime_crio_containers_oom_total` and  `container_runtime_crio_containers_oom` metrics,
which collects out of memory (oom) containers.
```


server/metrics/metrics.go
```diff
@@ -38,6 +38,9 @@ const (
 	// CRIOImageLayerReuseKey is the key for the CRI-O image layer reuse metrics.
 	CRIOImageLayerReuseKey = "crio_image_layer_reuse"
 
+	// CRIOContainersOOMKey is the key for the CRI-O container out of memory metrics.
+	CRIOContainersOOMKey = "crio_containers_oom"
```
lilic, line 42: My suggestion would be to use `_total` suffix here, as per best practices:

```suggestion
	CRIOContainersOOMKey = "crio_containers_oom_total"
```

saschagrunert: Fixed as suggested :+1:

saschagrunert:
@haircommander fine with merging this?

lilic:
I had another thought, for alerting to work on this correctly you might need a total metric without the container name, only the total count per node. Or you might need to remove this metric if the container no longer exists.

saschagrunert:
> I had another thought, for alerting to work on this correctly you might need a total metric without the container name, only the total count per node. Or you might need to remove this metric if the container no longer exists.

Alright, I added another metric to count the ooms without any label.

server/metrics/metrics.go
```diff
@@ -38,6 +38,12 @@ const (
...
```
lilic, line 42: Both should have _total suffix.

lilic (commented):
I meant my suggestion to be mutally exclusive, sorry for not being clear.

So how I see it is that we should either have a total number of container OOMs per node or per container name, but then we need to unregister/remove the metric when that container no longer exists.

Otherwise we end up with lots of active series about containers that are already gone.

server/metrics/metrics.go
saschagrunert: Removed that metric as per https://github.com/cri-o/cri-o/pull/4690#pullrequestreview-623167085

saschagrunert:
> So how I see it is that we should either have a total number of container OOMs per node or per container name, but then we need to unregister/remove the metric when that container no longer exists.
> 
> Otherwise we end up with lots of active series about containers that are already gone.

Makes sense, I removed the metric which collects by container name.

dgrisonnet:
I wonder if we really want to remove the container name from this metric.

I definitely agree with you @lilic that this metric would be useful to alert based on the number of container that were OOMKilled in a node, but I wonder if we could also use it to know if a specific container was OOM killed. To my knowledge, there are currently no metrics covering this in the ecosystem, and it could be used to create alert based on specific memory hungry containers being OOMKilled.

lilic:
Either work I think. For OOM containers killed alert, it depends how valuable container name can be, it can be very high cardinality, it could be that instance/node is valuable enough for an SRE to know there is huge OOM containers killed on that node. Depends really.

We can also start by adding this metric and add the container name specific one, if this is not enough?

dgrisonnet:
Sounds good to :+1:. I also think that having it per instance/node is more valuable, but I wanted to verify that we didn't miss any immediate use case for this metric.

saschagrunert:
@lilic @dgrisonnet I now assume that we're fine with the current approach :)

mrunalp (approved):

saschagrunert:
Re-adding the per container metric as per discussion thead in https://coreos.slack.com/archives/GSRDZ3NMN/p1617823886120700?thread_ts=1617821335.114600&cid=GSRDZ3NMN

server/metrics/metrics.go
```diff
@@ -144,6 +150,25 @@ var (
 		},
 		[]string{"name"},
 	)
+
+	// CRIOContainersOOMTotal collects container out of memory (oom) metrics for every container and sandboxes.
+	CRIOContainersOOMTotal = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      CRIOContainersOOMTotalKey,
+			Help:      "Amount of containers killed because they ran out of memory (OOM)",
```
TomSweeneyRedHat, line 159: Amount or "The number"?

Ditto line 168

tutorials/metrics.md
```diff
@@ -28,18 +28,20 @@ endpoint manually via [curl][1].
...
+| `crio_image_layer_reuse`                     | `name`                                                                                                                                  | Counter | Reused (not pulled) local image layer count by name.                                                 |
+| `crio_containers_oom_total`                  |                                                                                                                                         | Counter | Total amount of containers killed because they ran out of memory (OOM)                               |
+| `crio_containers_oom`                        | `name`                                                                                                                                  | Counter | Containers killed because they ran out of memory (OOM) by their name                                 |
```
TomSweeneyRedHat, line 44: Maybe it's late in the day brain, but I'm not gronking the difference between these last two....

saschagrunert: One is the total amount of OOM kills and the second is the oom kills by their container name :upside_down_face:

server/metrics/metrics.go
saschagrunert: Fixed as suggested.

saschagrunert:
Had to rebase to resolve the merge conflict.

PR #4694, opened 2021-03-26
Reload the main config file when reloading configs

nalind:
#### What type of PR is this?

/kind bug

When reloading the configuration, don't overwrite the name of the main configuration file with the name of the last drop-in that we parse, which would prevent us from reloading its contents the next time.

None

`pkg/config.Config.UpdateFromPath()` was calling `pkg/config.Config.UpdateFromFile()`, which always updated the `Config` object's `singleConfigPath` value, which it would later treat as the main configuration file in `pkg/config.Config.Reload()`.

```release-note
Fixed a bug which would cause configuration files to not be reloaded properly on SIGHUP when one or more "drop-in" configuration files were being used.
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -486,10 +486,25 @@ func (t *tomlConfig) fromConfig(c *Config) {
 	t.Crio.Metrics.MetricsConfig = c.MetricsConfig
 }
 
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
LGTM

haircommander:
/retest
/lgtm

thanks @nalind !

Issue #4693, opened 2021-03-26
Different container state between kubernetes and crictl


michalzxc:
I am opening this one in replacement of #4676 what wasn't accurate

**Description**
I found container what shows in kubectl describe that it is Running, while crioctl shows "Completed" with exit code 0:
```
  drone-1qljfp58m4cuf0qxndjo:
    Container ID:  cri-o://7f8d467da4d7288d9083887f5539fdac4ba5d6b3967638c8b88b31d2ab1e65f2
    Image:         ecr.ad.dice.fm/base:waitdbs-latest
    Image ID:      ecr.ad.dice.fm/base@sha256:5973219db525061334ccb4ef74f1e3220ad13f01ad0419599cd5be933bc73d2b
    Port:          <none>
    Host Port:     <none>
    Command:
      /bin/sh
      -c
    Args:
      echo "$DRONE_SCRIPT" | /bin/sh
    State:          Running
      Started:      Tue, 23 Mar 2021 17:15:21 +0000
    Last State:     Terminated
      Reason:       Error
      Exit Code:    2
      Started:      Tue, 23 Mar 2021 17:14:54 +0000
      Finished:     Tue, 23 Mar 2021 17:15:20 +0000
```

```
crictl inspect 7f8d467da4d72
{
  "status": {
    "id": "7f8d467da4d7288d9083887f5539fdac4ba5d6b3967638c8b88b31d2ab1e65f2",
    "metadata": {
      "attempt": 1,
      "name": "drone-1qljfp58m4cuf0qxndjo"
    },
    "state": "CONTAINER_EXITED",
    "createdAt": "2021-03-23T17:15:21.459089592Z",
    "startedAt": "2021-03-23T17:15:21.479616027Z",
    "finishedAt": "2021-03-23T17:15:24.646643722Z",
    "exitCode": 0,
    "image": {
      "image": "ecr.ad.dice.fm/base:waitdbs-latest"
    },
    "imageRef": "ecr.ad.dice.fm/base@sha256:5973219db525061334ccb4ef74f1e3220ad13f01ad0419599cd5be933bc73d2b",
    "reason": "Completed",
```

In kubectl describe I am not showing pod status, but status of particulat container inside a pod

It doesn't happen all the time, but in 24h I can find couple of them, on our CI/CD kubernetes nodes

Below is full crio debug log:
https://drive.google.com/file/d/1FRP4FBnt92wkMObmSQb2fHZVzjGZW_qA/view

[...]

PR #4695, opened 2021-03-29
Update container status on ListContainers

saschagrunert:
#### What type of PR is this?

/kind bug

There are cases where the container status is outdated when returned by
the `ListContainers` RPC. We now update the status for each container
before returning the list, which should mitigate this corner case.

Refers to https://github.com/cri-o/cri-o/issues/4693
None

```release-note
None
```

### DIFF
--- a/server/container_list.go
+++ b/server/container_list.go
@@ -85,6 +85,15 @@ func (s *Server) ListContainers(ctx context.Context, req *types.ListContainersRe
 		img := &types.ImageSpec{
 			Image: ctr.Image(),
 		}
+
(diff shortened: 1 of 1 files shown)


saschagrunert:
@haircommander please take a look

haircommander:
I am unsure about this. this will add quite a bit of overhead to ListContainers, which is called quite frequently. I think it may be better for us to find *why* we're in ContainerStateStopped but haven't set the ExitCode. That should be happening asynchronously from ListContainers. Even if we're hitting a race, it should be updated *soon*. in other words, I think we're hiding a missed case here.

PR #4696, opened 2021-03-29
test/testdata/sandbox_config.json: fix the dns_config

nalind:
#### What type of PR is this?

/kind other

8.8.8.8 would be a server, not a search domain, so change the section it appears in in the test sandbox config from "searches" to "servers".

None

```
None
```


haircommander:
a little late, but 
/retest
/lgtm

PR #4697, opened 2021-03-29
crio-wipe: only clear storage if CleanShutdownFile is supported

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

or else, we will unconditionally try to remove the whole directory after upgrading from 1.20,
rather than pointedly removing crio specific containers

```release-note
none
```

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -249,6 +249,16 @@ func main() {
 				logrus.Error(err)
 			}
 
+			// Write "$CleanShutdownFile".supported to show crio-wipe that
+			// we should be wiping if the CleanShutdownFile wasn't found.
+			// This protects us from wiping after an upgrade from a version that don't support
+			// CleanShutdownFile.
+			f, err := os.Create(config.CleanShutdownSupportedFileName())
+			if err != nil {
+				logrus.Errorf("Writing clean shutdown supported file: %v", err)
+			}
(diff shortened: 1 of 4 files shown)


cmd/crio/main.go
```diff
@@ -249,6 +249,16 @@ func main() {
...
 
+			// Write "$CleanShutdownFile".supported to show crio-wipe that
+			// we should be wiping if the CleanShutdownFile wasn't found.
+			// This protects us from wiping after an upgrade from a version that don't support
+			// CleanShutdownFile.
+			f, err := os.Create(config.CleanShutdownSupportedFileName())
+			if err != nil {
+				logrus.Errorf("Writing clean shutdown supported file: %v", err)
+			}
+			f.Close()
```
rphillips, line 260: if err is returned, the f will be nil

haircommander: I don't think that's a big deal. it'll return an error, but not segfault:
```
func (f *File) Close() error {
	if f == nil {
		return ErrInvalid
	}
	return f.file.close()
}

```

mrunalp (approved):

PR #4726, opened 2021-04-06
Bug 1942608: do not list the image with error locating manifest

QiWang19:
Fix Bug 1942608, during the call of ListImages, do not list the image if the there's no manifest exist.

#### What type of PR is this?

> /kind bug

https://bugzilla.redhat.com/show_bug.cgi?id=1942608

```release-note
None
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -93,6 +93,9 @@ type imageService struct {
 	ctx            context.Context
 }
 
+// ImageBeingPulled map[string]bool to keep track of the images haven't done pulling.
+var ImageBeingPulled sync.Map
+
 // CgroupPullConfiguration
 type CgroupPullConfiguration struct {
 	UseNewCgroup bool
@@ -316,6 +319,19 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext, filter s
 			}
 			results, err = svc.appendCachedResult(systemContext, ref, image, results, newImageCache)
 			if err != nil {
+				// skip reporting errors if the images haven't finished pulling
+				if os.IsNotExist(errors.Cause(err)) {
+					donePulling := true
+					for _, name := range image.Names {
+						if _, ok := ImageBeingPulled.Load(name); ok {
+							donePulling = false
+							break
+						}
+					}
+					if !donePulling {
+						continue
+					}
+				}
 				return nil, err
 			}
 		}
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -69,6 +69,7 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 		if !inProgress {
 			pullOp = &pullOperation{}
 			s.pullOperationsInProgress[pullArgs] = pullOp
+			storage.ImageBeingPulled.Store(pullArgs.image, true)
 			pullOp.wg.Add(1)
 		}
 		return pullOp, inProgress
@@ -79,6 +80,7 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 		defer func() {
 			s.pullOperationsLock.Lock()
 			delete(s.pullOperationsInProgress, pullArgs)
+			storage.ImageBeingPulled.Delete(pullArgs.image)
 			pullOp.wg.Done()
 			s.pullOperationsLock.Unlock()
 		}()


internal/storage/image.go
```diff
@@ -315,7 +315,7 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext, filter s
 				return nil, err
 			}
 			results, err = svc.appendCachedResult(systemContext, ref, image, results, newImageCache)
-			if err != nil {
+			if err != nil && !strings.Contains(err.Error(), fmt.Sprintf(`error locating item named "%s"`, "manifest")) {
```
mrunalp, line 318: This is too specific. Another option we discussed is keeping a list of images being pulled and then if we get any such error for an image in that list, we skip it.

haircommander:
/override ci/openshift-jenkins/critest_rhel
/override ci/openshift-jenkins/critest_fedora
/approve

LGTM

internal/storage/image.go
```diff
@@ -93,6 +93,8 @@ type imageService struct {
 	ctx            context.Context
 }
 
+var PulledImages map[string]int
```
mrunalp, line 96: Need sync on this.

mrunalp: use a sync.Map

QiWang19:
@mrunalp PTAL

QiWang19:
@vrothberg @saschagrunert PTAL

internal/storage/image.go
```diff
@@ -93,6 +93,8 @@ type imageService struct {
 	ctx            context.Context
 }
 
+var ImageBeingPulled sync.Map
```
haircommander, line 96: I think it could be useful to have a comment on the usage of this

vrothberg: I think this should be part of `type Server` since `pullOperationsInProgress` is already there. Talking about `pullOperationsInProgress`.  Why are we not reusing that map?

The `credentials` are also hashed so it may yield false negatives but I think the chances are very low.

server/image_pull.go
```diff
@@ -66,6 +70,10 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 		s.pullOperationsLock.Lock()
 		defer s.pullOperationsLock.Unlock()
 		pullOp, inProgress = s.pullOperationsInProgress[pullArgs]
+		if named, err := reference.ParseNormalizedNamed(pullArgs.image); err == nil {
```
vrothberg, line 73: The normalization to `docker.io` can lead to false positives. `fedora` and `library/fedora` would both resolve to `docker.io/library/fedora` although `fedora` has an alias to `registry.fedoraproject.org`.

I would really prefer we'd use `s.pullOperationsInProgress`.  Both have corner cases but using the existing struct, the corner cases would be consistent.

internal/storage/image.go
saschagrunert: Yes, let's reuse `pullOperationsInProgress` and maybe change it to a `sync.Map`

QiWang19:
@vrothberg @saschagrunert PTAL

server/server.go
```diff
@@ -67,23 +66,12 @@ type Server struct {
 
 	updateLock sync.RWMutex
 
-	// pullOperationsInProgress is used to avoid pulling the same image in parallel. Goroutines
-	// will block on the pullResult.
-	pullOperationsInProgress map[pullArguments]*pullOperation
```
vrothberg, line 72: Do we still need the `pullOperationsLock` below?

server/image_pull.go
```diff
@@ -65,10 +68,16 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 	pullOp, pullInProcess := func() (pullOp *pullOperation, inProgress bool) {
 		s.pullOperationsLock.Lock()
 		defer s.pullOperationsLock.Unlock()
-		pullOp, inProgress = s.pullOperationsInProgress[pullArgs]
+		op, inProgress := storage.PullOperationsInProgress.Load(pullArgs)
+		if inProgress {
+			pullOp, ok = op.(*pullOperation)
+			if !ok {
+				return &pullOperation{}, inProgress
```
vrothberg, line 75: I think this should be an error.  It _should_ never happen but only an error would tell us if it happens.

I really don't like `sync.Map` for having to deal with `Interface{}` (personal preference).

vrothberg (commented):
The untyped use of `sync.Map` is risky for an exported object (consistency is not enforced by the type system).

Looking at `internal/storage`, it seems that all we need to know is if a given image is currently downloaded.  The lookup at the moment is very expensive with O(N) walking the entire map.  I think we can separate the two concerns by keeping the `server` as is but keep an exported `map[string]int` in `internal/storage`. The `int` may be needed to account for multiple routines with different credentials pulling the same image.

The `server` would manage the map, and `internal/storage` would just access and query:
```golang
value, exists := map["alpine"]
if exists && value > 0 {
   // image is being pulled somewhere
}
```

The lookup would then be O(1).

QiWang19:
> The untyped use of `sync.Map` is risky for an exported object (consistency is not enforced by the type system).
> 
>  think we can separate the two concerns by keeping the `server` as is but keep an exported `map[string]int` in `internal/storage`. The `int` may be needed to account for multiple routines with different credentials pulling the same image.
> 

If the int value of the map may be incremented by different go routines, should we use it as sync.Map? see comments https://github.com/cri-o/cri-o/pull/4726#discussion_r610051350. Since it will use image nama string as the key, the search can also be done on O(1).

> The `server` would manage the map, and `internal/storage` would just access and query:
> 
> ```go
> value, exists := map["alpine"]
> if exists && value > 0 {
>    // image is being pulled somewhere
> }
> ```
> 
> The lookup would then be O(1).

Is it necessay to check the value > 0? if the image has done pulling in one routine, it can be deleted from the map so can we only check the key exist?

QiWang19:
@vrothberg @saschagrunert PTAL

server/image_pull.go
```diff
@@ -69,6 +69,7 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 		if !inProgress {
 			pullOp = &pullOperation{}
 			s.pullOperationsInProgress[pullArgs] = pullOp
+			storage.ImageBeingPulled.Store(pullArgs.image, 1)
```
vrothberg, line 72: I want to highlight one _potential_ error source. `pullArgs` include the credentials, so it is possible that the same image is being pulled by more than one routine. That's why I suggested an `int` which would allow for reference counting.

Looking at the PR now, I think that reference counting isn't necessary, so we could change `int` to `bool` and squeeze a bit of memory.

vrothberg (commented):
Other than the last nit, LGTM.

QiWang19:
@vrothberg @saschagrunert PTAL. Fixed the last nit.

vrothberg (commented):
LGTM

saschagrunert (approved):
/lgtm

TomSweeneyRedHat (commented):
LGTM

saschagrunert (approved):
/lgtm

PR #4730, opened 2021-04-07
Set short name mode to permissive

umohnani8:

#### What type of PR is this?

/kind feature

This will ensure that is the short name alias is not
found in the alias table it will always fall back to
looking through the search registries list as the user
cannot be prompted to select which registry to pull from.

None

```release-note
Fall back to search registries list if short name alias is not found in the alias table.

```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -745,6 +745,10 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 		return nil, err
 	}
 
+	// Set shortNameMode to permissive so we fall back to look through the search registries list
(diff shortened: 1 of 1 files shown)


umohnani8:
@vrothberg PTAL

mrunalp:
Is there a test we can add for this?

saschagrunert (approved):

vrothberg (commented):
LGTM

vrothberg:
> Is there a test we can add for this?

I don't know an easy way to test but I think we can rely on the very thorough tests we have containers/image.

TomSweeneyRedHat:
LGTM
once the one last hold out test is happy

PR #4744, opened 2021-04-09
server: cleanup sandbox cgroup after creation

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

otherwise they leak

```release-note
none
```

### DIFF
--- a/internal/config/cgmgr/cgmgr.go
+++ b/internal/config/cgmgr/cgmgr.go
@@ -66,6 +66,9 @@ type CgroupManager interface {
 	// CreateSandboxCgroup takes the sandbox parent, and sandbox ID.
 	// It creates a new cgroup for that sandbox, which is useful when spoofing an infra container.
 	CreateSandboxCgroup(sbParent, containerID string) error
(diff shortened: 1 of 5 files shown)


kolyshkin:
Do we need a test case?

sohankunkerkar:
@haircommander Do you want to pursue this? Or I can cover this as part of the CRI-O stats PR.

haircommander:
gosh I wish I had written down why I didn't push harder for this. I don't think these cgroups actually leak in practice because kubelet cleans them up, they only potentially leak when cgroups aren't being managed by the kubelet (like standalone with crictl) which is not really a supported case

sohankunkerkar:
> gosh I wish I had written down why I didn't push harder for this. I don't think these cgroups actually leak in practice because kubelet cleans them up, they only potentially leak when cgroups aren't being managed by the kubelet (like standalone with crictl) which is not really a supported case

OK, it sounds like we don't need this anymore. I will close this PR then.

PR #4745, opened 2021-04-09
WIP workloads: configure fields on the pod level

haircommander:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake
> /kind other

Before, configuring a workload would only do so for the container. This ignored stuff like conmon and the infra container--stuff under the pod's cgroup. we also have to configure that

```release-note
Workload customization is now applied on the pod slice level. It can be customized in a similar way as containers: with an annotation `$annotation_prefix.$resource/POD`
```

### DIFF
--- a/test/workloads.bats
+++ b/test/workloads.bats
@@ -12,6 +12,11 @@ function setup() {
 
 function teardown() {
 	cleanup_test
+	if [ -n "$tmp_cg" ]; then
+		for controller in cpu cpuset devices; do
+			cgdelete -g "$controller:$tmp_cg" -r
+		done
+	fi
 }
 
 function create_workload() {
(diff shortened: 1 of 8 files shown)


haircommander:
100% untested on cgroupv2

haircommander:
oh, and tests only work with cgroupfs rn

test/workloads.bats
```diff
@@ -25,7 +30,26 @@ resources = { "cpushares" =  "$cpushares", "cpuset" = "$cpuset" }
 EOF
 }
 
-function check_cpu_fields() {
+function setup_pod_tests() {
+	if [[ "$CONTAINER_CGROUP_MANAGER" == "cgroupfs" ]]; then
```
haircommander, line 34: @giuseppe @mrunalp @kolyshkin is there a nice way to create a slice with systemctl on the cmd line? docs didn't turn anything up

Issue #4727, opened 2021-04-07
CNI DEL not called on node reboot


caseydavenport:
**Description**

I'm seeing a behavior where, upon rebooting an OpenShift node using CRI-O, the CNI plugin does not appear to be called to tear down resources for the previously running pods on that node.

Basically, I am making note of the running pod sandbox IDs, rebooting the node with `sudo reboot now` (without draining the pods, effectively simulating a failure scenario) and letting the node come back up. When it does, I see the pods get launched again with new sandboxes. However, the old sandboxes never seem to get a CNI DEL. 

Similarly, I can see cache files in `/var/lib/cni/results` for both the new and old sandboxes. Does CRIO make any attempt to release these resources on node reboot?

**Steps to reproduce the issue:**

1. Run some pods on a node. Note the pod sandbox container IDs.
2. Reboot the node, see that the pods are re-launched with new pod sandboxes
3. See no evidence in the logs that the old sandbox IDs were released. 

**Describe the results you received:**

I expect CNI DEL to be called on the old sandbox IDs, providing a chance for CNI plugins to release any associated resources.

**Describe the results you expected:**

CNI DEL is not called.

**Additional information you deem important (e.g. issue happens only occasionally):**

I haven't looked at the code yet, but this seems very reproducible in my environment and from the logs it looks like no effort is even made to release the resources. It seems counter-intuitive to need this given a node-reboot will tear down namespaces, etc., but there are other resources which persist across node reboot and thus need to be released even if the original sandbox is no longer running.  

**Output of `crio --version`:**

```
crio version 1.18.2-18.rhaos4.5.git754d46b.el8
Version:    1.18.2-18.rhaos4.5.git754d46b.el8
GoVersion:  go1.13.4
Compiler:   gc
Platform:   linux/amd64
Linkmode:   dynamic
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

OpenShift, IPI in AWS using Calico CNI.

caseydavenport:
@haircommander this was another behavior I spotted while investigating https://github.com/cri-o/cri-o/issues/4719

I haven't looked into the root cause yet, though.

haircommander:
Ah see this is difficult though. When we reboot a node, the CNI pod is terminated. Here's a reboot flow:
`crio wipe` is called, which wipes all stale containers (to allow resources like held ports to be released, which allows the kube-apiserver to start after a non-graceful shutdown)
CNI pod is deleted with crio wipe
[...]

PR #4763, opened 2021-04-15
Cleanup pod network on sandbox removal

saschagrunert:
#### What type of PR is this?

/kind bug

We may encounter situations where sandboxes gets killed. In this case,
we now try to cleanup the network on removal of the sandbox to ensure
that no resources (like networks) left stale on the machine.

Refers to https://github.com/cri-o/cri-o/issues/4727
None

```release-note
Cleanup pod networks on sandbox removal, which now removes networks even if the pod sandbox got killed.
```

### DIFF
--- a/test/network.bats
+++ b/test/network.bats
@@ -1,5 +1,7 @@
 #!/usr/bin/env bats
 
+# vim:set ft=bash :
+
 load helpers
 
 function setup() {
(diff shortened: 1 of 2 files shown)


test/network.bats
```diff
@@ -140,3 +142,29 @@ function check_networking() {
...
+
+	CNI_RESULTS_DIR=/var/lib/cni/results
+	POD=$(crictl runp "$TESTDATA/sandbox_config.json")
+
+	# CNI result is there
+	# shellcheck disable=SC2010
+	[[ $(ls $CNI_RESULTS_DIR | grep "$POD") != "" ]]
+
+	# kill the sandbox
+	run "$CONTAINER_DEFAULT_RUNTIME" --root /var/run/runc kill "$POD"
```
haircommander, line 157: 1: we can call the `runtime` function here 
2: why use `run`? we're not checking output. if it's just to prevent a failure from kiling the test, we could use `|| true`

TomSweeneyRedHat:
LGTM
assuming happy tests

mrunalp (approved):

saschagrunert:
This is now included in https://github.com/cri-o/cri-o/pull/4767
/close

PR #4767, opened 2021-04-15
Introduce InternalWipe

haircommander:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind ci
> /kind cleanup
> /kind dependency-change
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind other

On node reboot, there are situations where CNI plugins aren't able to cleanup the resources they acquire. They rely on the caller of CNI add to send a corresponding CNI del.

The problem is, CRI-O was previously not able to send the CNI del, because `crio wipe` removed the evidence of the container's (and pod's) existence. As such, on every reboot, some CNI plugins have the ability to leak resources.

This PR is the first step in remediating it. It moves the responsibility of wiping images and containers from `crio wipe` to server startup. This behavior is gated by a config value `internal_wipe`.

It leaves the responsibility of checking whether the node shutdown and was able to sync to `crio wipe`, as well as leaves `crio wipe -f` which has proven useful to wipe the state of a node.

We also adapt the tests to the new behavior (previously, we were wiping the pause container, which we no longer do. I think this is the correct behavior)

fixes half of https://github.com/cri-o/cri-o/issues/4727

built on top of https://github.com/cri-o/cri-o/pull/4763

```release-note
Add the config field `internal_wipe` which moves the responsibility of wiping containers after a reboot and images after an upgrade from the external binary `crio wipe` to the main crio server. This has a handful of advantages, the main one being crio is now better able to cleanup CNI resources after a reboot.
```


docs/crio.8.md
```diff
@@ -232,6 +233,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
        their CA to their system's list of trusted CAs instead of using
        '--insecure-registry'. (default: [])
 
+**--internal-wipe**: Whether CRI-O should wipe containers on server startup. If set to false, 'crio wipe' will cleanup the containers instead.
```
TomSweeneyRedHat, line 236: what does "cleanup the containers" mean?  I don't see another instance of "cleanup" in this entire man page.  A bit more detail please.  It would also be good to be specific about what wipe does.

docs/crio.conf.5.md
```diff
@@ -54,6 +54,10 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
   It is used to check if crio wipe should wipe images, which should
   only happen when CRI-O has been upgraded
 
+**internal_wipe**=false
+  Whether CRI-O should wipe containers after a reboot on server startup.
+  If set to false, 'crio wipe' will cleanup the containers instead.
```
TomSweeneyRedHat, line 59: see prior

internal/resourcestore/resourcecleaner.go
```diff
@@ -1,33 +1,77 @@
...
 type ResourceCleaner struct {
+	ctx   context.Context
 	funcs []CleanupFunc
 }
 
 // NewResourceCleaner creates a new ResourceCleaner
-func NewResourceCleaner() *ResourceCleaner {
+func NewResourceCleaner(ctx context.Context) *ResourceCleaner {
 	return &ResourceCleaner{
+		ctx:   ctx,
```
haircommander, line 29: I don't think we're supposed to save ctx as a member variable

haircommander:
ref BZ https://bugzilla.redhat.com/show_bug.cgi?id=1948047 https://bugzilla.redhat.com/show_bug.cgi?id=1948137

internal/resourcestore/resourcecleaner.go
saschagrunert: The main reason is to be able to use `log.Errof` in `Cleanup()`. Do you think we should pass the context on `Cleanup(ctx)` as alternative?

saschagrunert:
@haircommander added another commit to run the cleanup in parallel

saschagrunert:
Something in the pod sandbox removal it broken since ec883ab13d310761651a120858763d1f64bf57d0 which made the integration tests fail.

saschagrunert:
Bundle test is not broken by this PR.

saschagrunert:
Some integration tests are still broken due to the sandbox cleanup behavior.

saschagrunert:
It looks like that the CNI del for the bridge plugin (v0.9.1) does fail if the network namespace (`/var/run/netns/…`) does not exist. In that case it will also not cleanup the resources in `/var/lib/cni/results`. I'm not sure if we can do anything from the CRI-O side to solve this kind of issue.

saschagrunert:
I think I found the issue in the parallel cleanup: We have to cleanup the namespace resources always after network teardown. This may solve the error from the CNI plugin.

saschagrunert:
@haircommander please take a look. I reverted some of my experimental commits and it seems that still some integration tests are failing due to the parallel cleanup.

server/sandbox_run_linux.go
```diff
@@ -685,11 +709,20 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
 			log.Errorf(ctx, "error stopping network on cleanup: %v", err2)
+			return err2
+		}
+		// retry with different context
+		// If we don't, this cleanup will always fail if the context times out
+		if err2 := s.networkStop(context.Background(), sb); err2 != nil {
+			log.Errorf(ctx, "error stopping network on cleanup: %v", err2)
+			return err2
 		}
+		return nil
```
saschagrunert, line 725: The issue with that is now if we return an error, it will cause a retry of the function. I don't think that it can work this way.

mrunalp:
@kolyshkin @QiWang19 ptal

saschagrunert (approved):
LGTM

QiWang19 (commented):
/lgtm

haircommander:
/hold

the unit test failure seems to be legit. I think I messed something up...

haircommander:
/hold cancel
fixed
/retest

internal/lib/container_server.go
```diff
@@ -199,30 +199,30 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (retErr er
...
+	}{
+		{rspecNS: rspec.NetworkNamespace, joinFunc: sb.NetNsJoin},
+		{rspecNS: rspec.IPCNamespace, joinFunc: sb.IpcNsJoin},
+		{rspecNS: rspec.UTSNamespace, joinFunc: sb.UtsNsJoin},
+		{rspecNS: rspec.UserNamespace, joinFunc: sb.UserNsJoin},
+	}
+	for _, namespaceToJoin := range namespacesToJoin {
+		path, err := configNsPath(&m, namespaceToJoin.rspecNS)
+		if err == nil {
+			if nsErr := namespaceToJoin.joinFunc(path); err != nil {
```
kolyshkin, line 223: Do you mean
```suggestion
			if nsErr := namespaceToJoin.joinFunc(path); nsErr != nil {
```

kolyshkin: Oh I see this error is in the current codebase. Makes sense to have a separate PR to fix it.

haircommander: ach damn

haircommander: https://github.com/cri-o/cri-o/pull/4796 to fix, I'll rebase this on that when evaluating review comments

cmd/crio/wipe.go
```diff
@@ -67,6 +67,14 @@ func crioWipe(c *cli.Context) error {
 		return handleCleanShutdown(config, store)
 	}
 
+	// If we make it here, and we're set to wipe internally
+	// crio wipe is done, unless force is set.
```
kolyshkin, line 71: Is there some punctuation missing between the first and the second line? It's a bit hard to parse this comment.

server/server.go
```diff
@@ -489,6 +492,66 @@ func New(
...
+func (s *Server) wipeIfAppropriate(ctx context.Context) {
+	if !s.config.InternalWipe {
+		return
+	}
+	// First, check if the node was rebooted.
+	// We know this happened because the VersionFile (which lives in a tmpfs)
+	// will not be there.
+	shouldWipeContainers, err := version.ShouldCrioWipe(s.config.VersionFile)
+	if err != nil {
+		log.Infof(ctx, "checking whether cri-o should wipe containers: %v", err)
```
kolyshkin, line 504: perhaps `log.Warnf(ctx, "error while checking ....`?

server/server.go
```diff
@@ -489,6 +492,66 @@ func New(
...
+	}
+
+	// there are two locations we check before wiping:
+	// one in a temporary directory. This is to check whether the node has rebooted.
+	// if so, we should remove containers
+	// another is needed in a persistent directory. This is to check whether we've upgraded
+	// if we've upgraded, we should wipe images
+	shouldWipeImages, err := version.ShouldCrioWipe(s.config.VersionFilePersist)
+	if err != nil {
+		log.Infof(ctx, "%v: triggering wipe of images", err.Error())
```
kolyshkin, line 514: `log.Warnf`?

server/server.go
```diff
@@ -489,6 +492,66 @@ func New(
...
+	}
+
+	shouldWipeContainers = shouldWipeContainers || shouldWipeImages
+
+	// First, save the images we should be wiping
+	// We won't remember if we wipe all the containers first
+	imagesToWipe := make([]string, 0)
```
kolyshkin, line 521: `var imagesToWipe []string`

(`make` makes sense if we know the size, otherwise it just makes things look more complicated when they should)

server/server.go
```diff
@@ -489,6 +492,66 @@ func New(
...
+
+	if shouldWipeContainers {
+		for _, sb := range s.ContainerServer.ListSandboxes() {
+			cleanupFunc := func() error {
+				if err := s.stopPodSandbox(ctx, sb); err != nil {
+					return err
+				}
+				return s.removePodSandbox(ctx, sb)
+			}
+			if err := cleanupFunc(); err != nil {
```
kolyshkin: What is the point of using the cleanupFunc here? It looks like it's equivalent to this (simpler) code:

```go
for _, sb := range s.ContainerServer.ListSandboxes() {
   err := s.stopPodSandbox(ctx, sb)
   if err == nil {
      err = s.removePodSandbox(ctx, sb)
   }
   if err != nil {
      log.Warnf(ctx, "Failed to cleanup pod %s: %v", sb.ID(), err)
   }
}
```

test/crio-wipe.bats
```diff
@@ -48,20 +48,26 @@ function test_crio_wiped_images() {
...
 	output=$(crictl images)
-	[[ "$output" == *"pause"* ]]
+	[[ "$output" == *"$IMAGE_USED"* ]]
 }
 
 function start_crio_with_stopped_pod() {
 	start_crio
 
+	# it must be everything before the tag, because crictl output won't match (the columns for image and tag are separated by space)
+	export IMAGE_USED
```
kolyshkin, line 64: I doubt that `export` is needed here. Is this just to eliminate a shellcheck warning?

internal/resourcestore/resourcecleaner.go
kolyshkin: Yes, I think it's better to pass `ctx` to `Cleanup` directly.

internal/resourcestore/resourcecleaner.go
```diff
@@ -1,33 +1,72 @@
...
 }
 
 // Cleanup cleans up the resource, running
 // the cleanup funcs in opposite chronological order
 func (r *ResourceCleaner) Cleanup() {
 	for i := len(r.funcs) - 1; i >= 0; i-- {
-		r.funcs[i]()
+		if err := retry(
+			r.funcs[i],
+		); err != nil {
```
kolyshkin: why not
```suggestion
		if err := retry(r.funcs[i]); err != nil {
```
?

internal/resourcestore/resourcecleaner.go
```diff
@@ -1,33 +1,72 @@
...
+func retry(fn func() error) error {
+	backoff := wait.Backoff{
+		Duration: 500 * time.Millisecond,
+		Factor:   1.5,
+		Steps:    5,
 	}
+
+	waitErr := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		if err := fn(); err != nil {
+			return false, nil
```
kolyshkin, line 74: is there any way not to discard the real error returned from `fn()`?

server/sandbox_network.go
```diff
@@ -180,3 +183,50 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
 
 	return sb.SetNetworkStopped(true)
 }
+
+func (s *Server) newPodNetwork(sb *sandbox.Sandbox) (ocicni.PodNetwork, error) {
+	var egress, ingress int64 = 0, 0
```
kolyshkin, line 188: No need to explicitly initialize to `0` -- this is the default value.

server/server.go
```diff
@@ -529,20 +529,28 @@ func (s *Server) wipeIfAppropriate(ctx context.Context) {
...
+			sb := sb
 			cleanupFunc := func() error {
 				if err := s.stopPodSandbox(ctx, sb); err != nil {
 					return err
 				}
 				return s.removePodSandbox(ctx, sb)
 			}
 			if err := cleanupFunc(); err != nil {
-				log.Warnf(ctx, "Failed to cleanup pod %s: %v", sb.ID(), err)
+				log.Warnf(ctx, "Failed to cleanup pod %s once, will try again: %v", sb.ID(), err)
```
kolyshkin, line 543:
```suggestion
				log.Warnf(ctx, "Failed to cleanup pod %s (will retry): %v", sb.ID(), err)
```
and maybe demote from warning to debug?

test/crio-wipe.bats
```diff
@@ -263,3 +263,41 @@ function start_crio_with_stopped_pod() {
 	test_crio_did_not_wipe_containers
 	test_crio_did_not_wipe_images
 }
+
+@test "internal_wipe eventually cleans network on forced restart of crio if network is slow to come up" {
+	export CNI_RESULTS_DIR=/var/lib/cni/results
```
kolyshkin, line 268: not sure if `export` is needed here.

test/crio-wipe.bats
```diff
@@ -263,3 +263,41 @@ function start_crio_with_stopped_pod() {
...
+
+	start_crio
+
+	# it must be everything before the tag, because crictl output won't match (the columns for image and tag are separated by space)
+	export IMAGE_USED
```
kolyshkin, line 273: ditto

test/crio-wipe.bats
```diff
@@ -263,3 +263,41 @@ function start_crio_with_stopped_pod() {
...
+	sleep 5s
+
+	# pretend like the CNI container has started
+	mv "$CRIO_CNI_PLUGIN"/"$CNI_TYPE"-hidden "$CRIO_CNI_PLUGIN"/"$CNI_TYPE"
+
+	# allow cri-o to catch up
+	sleep 5s
+
+	# make sure network resources were cleaned up
+	[[ -z $(ls "$CNI_RESULTS_DIR"/*"$pod_id"* || true) ]]
```
kolyshkin, line 302: easier to just do
```suggestion
	! ls "$CNI_RESULTS_DIR"/*"$pod_id"*
```

internal/resourcestore/resourcecleaner.go
```diff
@@ -5,56 +5,68 @@ import (
...
 }
 
+// A cleanupFunc is a function that cleans up one piece of
+// the associated resource.
+type cleanupFunc func() error
+
 // NewResourceCleaner creates a new ResourceCleaner
-func NewResourceCleaner(ctx context.Context) *ResourceCleaner {
-	return &ResourceCleaner{
-		ctx:   ctx,
```
kolyshkin: An earlier commit from this PR adds ctx to `ResourceCleaner`, and now you remove it. It's better to fix the first commit right away (less changes in history, easier to review).

internal/storage/runtime.go
```diff
@@ -71,6 +71,11 @@ type RuntimeServer interface {
 	// Pointer arguments can be nil.  Either the image name or ID can be
 	// omitted, but not both.  All other arguments are required.
 	CreatePodSandbox(systemContext *types.SystemContext, podName, podID, imageName, imageAuthFile, imageID, containerName, metadataName, uid, namespace string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, privileged bool) (ContainerInfo, error)
+	// RemovePodSandbox deletes a pod sandbox's infrastructure container.
+	// The CRI expects that a sandbox can't be removed unless its only
+	// container is its infrastructure container, but we don't enforce that
+	// here, since we're just keeping track of it for higher level APIs.
+	RemovePodSandbox(idOrName string) error
```
kolyshkin: This function was removed by an earlier commit, and now it's being re-added. I don't follow...

server/sandbox_run_linux.go
```diff
@@ -707,22 +708,23 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
 		// retry with different context
 		// If we don't, this cleanup will always fail if the context times out
 		if err2 := s.networkStop(context.Background(), sb); err2 != nil {
 			log.Errorf(ctx, "error stopping network on cleanup: %v", err2)
+
 			return err2
 		}
-		return nil
+
+		// Now that we've succeeded in stoppign network, cleanup namespaces
```
kolyshkin, line 725: typo: `stoppign`

server/sandbox_run_linux.go
```diff
@@ -707,22 +708,23 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
 		log.Infof(ctx, description)
-		if err2 := s.networkStop(ctx, sb); err2 != nil {
-			log.Errorf(ctx, "error stopping network on cleanup: %v", err2)
-			return err2
-		}
 		// retry with different context
 		// If we don't, this cleanup will always fail if the context times out
 		if err2 := s.networkStop(context.Background(), sb); err2 != nil {
 			log.Errorf(ctx, "error stopping network on cleanup: %v", err2)
+
```
kolyshkin, line 721: please remove

server/server.go
haircommander: because we expect this to fail in some cases (if the node has restarted but the networking plugin pods haven't been created, then any non-host network pods can't have their networking resources cleaned up)

test/crio-wipe.bats
haircommander: I think export is needed because this field is used outside this function

server/server.go
haircommander: oh I see, we don't yet actually do that. I think it was just to reduce churn between commits, but I can drop it for this commit and update in a later one

test/crio-wipe.bats
haircommander: hm not needed :)

internal/storage/runtime.go
haircommander: I have reworked the commits to only take the lint fixing parts of this commit. I believe this was readded by @saschagrunert to debug failures that other commits ended up fixing. we'll see how that works...

docs/crio.8.md
haircommander: alright I've updated the verbiage to hopefully be more clear

internal/resourcestore/resourcecleaner.go
mrunalp: Can we log the error here?

internal/resourcestore/resourcecleaner.go
```diff
@@ -1,33 +1,70 @@
...
 }
 
 // Cleanup cleans up the resource, running
 // the cleanup funcs in opposite chronological order
 func (r *ResourceCleaner) Cleanup() {
 	for i := len(r.funcs) - 1; i >= 0; i-- {
-		r.funcs[i]()
+		if err := retry(r.funcs[i]); err != nil {
+			log.Errorf(r.ctx, "Retried cleanup function too often, giving up")
+			// TODO: do we wanna return an error here?
```
mrunalp: We should bubble up for logging up the stack here.

haircommander: I have updated it to always return the error here

internal/resourcestore/resourcecleaner.go
haircommander: alright, done

haircommander:
okay comments addressed and I think this is ready again
@kolyshkin @saschagrunert @mrunalp @umohnani8 @QiWang19 PTAL

saschagrunert:
@haircommander this needs a rebase

PR #4791, opened 2021-04-22
Fix bundle e2e test while fixing coredns

saschagrunert:
#### What type of PR is this?

/kind failing-test

The bundle e2e tests were complaining about a coredns loop back to the
local machine. This was caused by the `/etc/resolv.conf`:

```
nameserver 127.0.0.53
options edns0 trust-ad
search kzav01uq0i3u1f1hch3ggm0mbb.bx.internal.cloudapp.net
```

This is now being fixed by using the local content from
`/run/systemd/resolve/resolv.conf`, which looks like:

```
nameserver 168.63.129.16
search gbcyptfplzzunl4htqv55y2vfc.jx.internal.cloudapp.net
```

None
None

```release-note
None
```

### DIFF
--- a/contrib/bundle/test-e2e
+++ b/contrib/bundle/test-e2e
@@ -7,6 +7,9 @@ if [[ $EUID -ne 0 ]]; then
     exit 1
 fi
 
+# Bypass local DNS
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM

Issue #4798, opened 2021-04-23
area/vm: pods are not reaching "completed" state


fidencio:
**Description**

Since #4650, kata pods are not reaching the "completed" state.  The main change introduced in that PR happens on the kind of context used inside the runtime VM code.

Prior to that patch, the context was created using `context.Background()`, which "is never canceled, has no values, and has no deadline", as shown [here](https://github.com/cri-o/cri-o/pull/4650/files#diff-7384fb0ac1915ef8b1fd72dc720492ebe0365f1b77ca29cdc836c9e584d2cdf4L74):
```
        return &runtimeVM{
		path:    path,
		fifoDir: filepath.Join(root, "crio", "fifo"),
		ctx:     context.Background(),
		ctrs:    make(map[string]containerInfo),
	}
```

With the change done by #4650, the we use a context coming from GRPC server as a parent, which has a deadline linked to it, when we create the context we'll use, as shown [here](https://github.com/cri-o/cri-o/blob/b99366680904420974463572f6f0b92166437b0e/internal/log/interceptors.go#L76-L78):
```
func addRequestName(ctx context.Context, req string) context.Context {
	return context.WithValue(ctx, Name{}, req)
}
```
Note 1: I was able to check that the context coming from GRPC has a deadline attached to it when printing it.

This change, simply this change, makes that the following pod never reaches its "Completed" status.
```
#
# Copyright (c) 2021 Red Hat, Inc.
#
# SPDX-License-Identifier: Apache-2.0
#
apiVersion: v1
kind: Pod
metadata:
  name: sharevol-kata
spec:
  runtimeClassName: kata
  restartPolicy: Never
  securityContext:
    runAsUser: 1001
    fsGroup: 123
  containers:
  - name: mounttest-container
    image: "k8s.gcr.io/e2e-test-images/agnhost:2.21"
    args:
      - mounttest
      - --fs_type=/test-volume
      - --new_file_0660=/test-volume/test-file
      - --file_perm=/test-volume/test-file
      - --file_owner=/test-volume/test-file
    volumeMounts:
      - name: emptydir-volume
        mountPath: /test-volume
  volumes:
  - name: emptydir-volume
    emptyDir: {}

```
Note 2: I was able to verify this by doing this change:
```
[fidencio@bump cri-o]$ git diff
diff --git a/internal/log/interceptors.go b/internal/log/interceptors.go
index cce141e22..ff1e6b4bc 100644
--- a/internal/log/interceptors.go
[...]

PR #4803, opened 2021-04-26
Use extra context for runtime VM

saschagrunert:
#### What type of PR is this?

/kind bug

We now partially revert https://github.com/cri-o/cri-o/pull/4650 to
still use the runtime internal context for its tasks.

Fixes: https://github.com/cri-o/cri-o/issues/4798
None

```release-note
Fixed bug where runtime VM created containers never reach their `completed` state. 
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -40,6 +40,7 @@ import (
 type runtimeVM struct {
 	path    string
 	fifoDir string
+	ctx     context.Context
 	client  *ttrpc.Client
 	task    task.TaskService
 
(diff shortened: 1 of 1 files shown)


fidencio:
@saschagrunert, thanks for the PR, it does solve the issue I reported.

tl;dr: LGTM, we should proceed with this PR, and I'd like to hear from @haircommander before adding my `lgtm`.

Now, the long reply ...

It bogus my mind, sincerely, and I've gone through this on Friday, coming up with pretty much the same solution proposed by in this PR, then I filed the issue.  I still don't fully get why we need the extra context, why the context coming from the GRPC doesn't do what it has to do.

@saschagrunert mentioned this may be a bug in the GRPC implementation of the runtime VM, and he may very much be right about that.  For now, I'm looking forward to have this partial revert merged, and backported to the `release-1.21` branch, as it'll unblock the 2.1.0-rc0 release of kata-containers.  Then we can track down what's broken on the runtime VM side. :-/

haircommander:
thanks for tracking this down y'all

PR #4804, opened 2021-04-26
main: still rely on logrus (rather than using the internal log)

fidencio:
#### What type of PR is this?

/kind bug

This is a partial revert of 975ffc6e482ab8581410b445c2268d60d64be62d,
where we started using the internal log when the context is avaiable.

It turns out that we missed a few pieces during the review, where we end
up using the internal log without having the interceptors set up.  This
was raised by @haircommander during the review, we checked a few parts
of the patch, but we still missed a few others.

We can only use the internal logging after
https://github.com/cri-o/cri-o/blob/d3dbaec060e33870e5cb5c3f7ec4207837804b00/cmd/crio/main.go#L222
happens.  From this moment, we start having different ways of logging in
the very same file and consistency becomes a problem then.

With the consistency in mind, I'd like to **only** use logrus for
debugging as part of this file.

Note: This was noticed when debugging
https://github.com/cri-o/cri-o/issues/4798

None

I think this one should also be backported to the `release-1.21` branch,

```release-note
None
```

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -63,9 +63,9 @@ func catchShutdown(ctx context.Context, cancel context.CancelFunc, gserver *grpc
 			case unix.SIGPIPE:
 				continue
 			case signals.Interrupt:
-				log.Debugf(ctx, "Caught SIGINT")
+				logrus.Debugf("Caught SIGINT")
 			case signals.Term:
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM

mrunalp (approved):

PR #4805, opened 2021-04-26
Add TLS support for serving metrics

saschagrunert:
#### What type of PR is this?

/kind feature

This adds TLS cert/key support for serving the metrics endpoint. To
allow backwards compatiblity, we now add an additional secure port as
serving instance.

Documentation and testing has been updated as well.

None
None

```release-note
Added secure metrics endpoint configuration options
```


TomSweeneyRedHat:
LGTM
assuming happy tests

server/server.go
```diff
@@ -552,9 +553,26 @@ func (s *Server) startMetricsServer() error {
 	}
 
 	if err := startMetricsEndpoint(
-		"tcp", fmt.Sprintf(":%v", s.config.MetricsPort), me,
+		"tcp", fmt.Sprintf(":%v", s.config.MetricsPort), "", "", me,
```
haircommander, line 556: should we be checking if this is 0, and not add the endpoint if so?

saschagrunert: Are you referring to the port? We may add this to the config validation:

- Provided cert but no key is invalid
- Provided key but no cert is invalid
- Provided cert and key but port is 0 is invalid

Maybe add a check if the file exists on disk.

haircommander: I more meant: we have two different ports configurable. if a user configures one with TLS, would they want the unsecured one? currently, we don't have a way to disable the unsecured one. is this something we want?

saschagrunert: Alright, removed the secure port and used the usual port for it. If cert and key are provided, then we will not serve the metrics as insecure any more.

saschagrunert:
@haircommander PTAL

haircommander:
/lgtm

/override ci/openshift-jenkins/e2e_crun
/override ci/openshift-jenkins/e2e_run_cgroupv2
/override ci/openshift-jenkins/integration_crun

(crun tests are currently busted)

PR #4844, opened 2021-05-04
add absent_mount_sources_to_reject option

haircommander:
#### What type of PR is this?

/kind feature
To match docker behavior, a directory is created when the src of a mount does not exist.

This causes issues when the src should not be a directory (/etc/hostname is a prime example).

Instead of relying on correct user behavior, we should urge them to do the right thing by erroring in this case (if a configured mount source is added but does not exist)

```release-note
Add support for `absent_mount_sources_to_reject`, which allows admins to configure paths that, when mounted into a container despite not existing on the host, causes a container creation request to fail. This is useful for paths like `/etc/hostname`, which causes trouble as a directory, but possibly shouldn't be created as a file either (in the case of a dynamic hostname).
```


server/container_create_linux.go
```diff
@@ -826,6 +826,10 @@ func addOCIBindMounts(ctx context.Context, mountLabel string, containerConfig *t
 		} else {
 			if !os.IsNotExist(err) {
 				return nil, nil, fmt.Errorf("failed to resolve symlink %q: %v", src, err)
+			} else if filepath.Clean(src) == "/etc/hostname" {
+				// special-case /etc/hostname, as we don't want it to be created as a directory
+				// This can cause issues with node reboot.
+				return nil, nil, errors.New("Cannot mount /etc/hostname: path does not exist and will cause issues as a directory. Consider mounting /proc/sys/kernel/hostname, or creating a static /etc/hostname")
```
saschagrunert, line 832: nit:
```suggestion
				return nil, nil, errors.New("cannot mount /etc/hostname: path does not exist and will cause issues as a directory. Consider mounting /proc/sys/kernel/hostname, or creating a static /etc/hostname")
```

saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM

mrunalp (approved):
/lgtm

PR #4853, opened 2021-05-05
server: add warning when bind mounting to the root

giuseppe:
if the bind mount points to the container root, raise a warning.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1950408

#### What type of PR is this?

/kind cleanup

warn the user when a bind mount targets the container root

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1950408

```release-note
raise a warning when creating a bind mount on the container root
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -814,9 +814,11 @@ func addOCIBindMounts(ctx context.Context, mountLabel string, containerConfig *t
 		if dest == "" {
 			return nil, nil, fmt.Errorf("mount.ContainerPath is empty")
 		}
-
 		if m.HostPath == "" {
 			return nil, nil, fmt.Errorf("mount.HostPath is empty")
+		}
+		if m.HostPath == "/" && dest == "/" {
+			log.Warnf(ctx, "configuration specifies mounting host root to the container root.  This is dangerous (especially with privileged containers) and should be avoided.")
 		}
 		src := filepath.Join(bindMountPrefix, m.HostPath)
 


TomSweeneyRedHat:
LGTM
thanks @giuseppe !

server/container_create_linux.go
```diff
@@ -814,7 +814,9 @@ func addOCIBindMounts(ctx context.Context, mountLabel string, containerConfig *t
 		if dest == "" {
 			return nil, nil, fmt.Errorf("mount.ContainerPath is empty")
 		}
-
+		if dest == "/" {
+			log.Warnf(ctx, "bind mounting %q to the container root.  It might have undesiderable effects", m.HostPath)
```
cgwalters, line 818:
```suggestion
			log.Warnf(ctx, "configuration specifies mounting host path %q to the container root.  This is dangerous (especially with privileged containers) and should be avoided.", m.HostPath)
```

cgwalters: Took a pass at a more strongly worded message.
(I'd at least s/undesiderable/undesirable/)

cgwalters:
I'm skeptical that admins are going to see this in the blizzard of node logs.

One pattern might be to force the container to crash once or something?  Or the equivalent of `sleep(60)?`

But, anyways better than nothing!

server/container_create_linux.go
giuseppe: thanks!  Amended now

giuseppe:
> I'm skeptical that admins are going to see this in the blizzard of node logs.
> 
> One pattern might be to force the container to crash once or something? Or the equivalent of `sleep(60)?`
> 
> But, anyways better than nothing!

wouldn't that go unnoticed anyway, at least until there are no visible effects?

Not sure if it is worth, but an alternative could be to make it configurable through a custom annotation and refuse to mount `/` unless the annotation is present

fgiudici:
LGTM!

server/container_create_linux.go
```diff
@@ -814,7 +814,9 @@ func addOCIBindMounts(ctx context.Context, mountLabel string, containerConfig *t
 		if dest == "" {
 			return nil, nil, fmt.Errorf("mount.ContainerPath is empty")
 		}
-
+		if dest == "/" {
```
rhatdan, line 817: Isn't this only dangerous is source="/" also?

giuseppe: it is less dangerous, but probably still worth reporting.  `$SOURCE/dev`, `$SOURCE/proc` and `$SOURCE/sys` can be propagated to the host

haircommander:
Part of me is thinking we should outright deny a request that tries to mount / to /, but ignore any that mounts / anywhere else

giuseppe:
> Part of me is thinking we should outright deny a request that tries to mount / to /, but ignore any that mounts / anywhere else

my doubt is, do we deny any valid use case?  This is possible only for privileged pods, users have plenty of ways to affect the host

haircommander:
it may be addressed upstream actually: https://github.com/kubernetes/kubernetes/issues/101749

giuseppe:
> it may be addressed upstream actually: [kubernetes/kubernetes#101749](https://github.com/kubernetes/kubernetes/issues/101749)

that seems like a better place to fix it.  IMO a warning is enough for now in CRI-O

server/container_create_linux.go
haircommander: I personally don't find it as risky, and openshift would report this by default with `oc debug node`, which diminishes the importance of the warning. I think we should only warn if source and dest are "/"

haircommander:
/lgtm

thanks @giuseppe

haircommander:
agnostic passed
/override ci/prow/e2e-gcp

PR #4858, opened 2021-05-06
Fix opt.semgrep.import-text-template error

soniasingla:
Signed-off-by: Sonia Singla <soniasingla.1812@gmail.com>

### DIFF
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -1,8 +1,8 @@
 package config
 
 import (
+	"html/template"
 	"io"
-	"text/template"
 )
 
 // WriteTemplate write the configuration template to the provided writer


vendor/github.com/containers/common/pkg/apparmor/apparmor_linux.go
```diff
@@ -11,7 +11,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
-	"text/template"
+	"html/template"
```
haircommander, line 14: we shouldn't change this here. It should be changed in the source and revendored

soniasingla: where is the source @haircommander ? thanks

soniasingla: @haircommander so i changed in template and run `make vendor`

haircommander: you can find it based on where the vendor directory is:
github.com/containers/common/tree/master/pkg/apparmor/apparmor_linux.go

soniasingla: thanks @haircommander so I should make the PR in that repo and close this one? thanks

haircommander: yes I would recommend that, thanks :)

soniasingla: ok thanks @haircommander

PR #4861, opened 2021-05-06
Attempt to fix Useless if statement, always the same behavior

soniasingla:
bugsbash


haircommander:
we shouldn't be changing things in the vendor directory, they'll just get clobbered when we run `make vendor`

PR #4862, opened 2021-05-06
(bugsbash) remove useless if statements

asquare14:
Signed-off-by: aSquare14 <atibhi.a@gmail.com>
This will always be true
```go
if true {
}
```

This will always be false and the code inside if will never be reached.
```go
if false {
// unreachable code
}
```
Hence, removing if-else for such cases.


haircommander:
@asquare14 sorry, we shouldn't make changes to the vendor directory (https://github.com/cri-o/cri-o/pull/4864 will update the checks to specify this)

PR #4863, opened 2021-05-06
bugsbash: Use net.JoinHostPort

soniasingla:


haircommander:
can you drop the changes to the vendor directory? https://github.com/cri-o/cri-o/pull/4864 will make them not necessary

vendor/github.com/go-git/go-git/v5/plumbing/transport/git/common.go
```diff
@@ -92,7 +93,7 @@ func (c *command) StdoutPipe() (io.Reader, error) {
 func endpointToCommand(cmd string, ep *transport.Endpoint) string {
 	host := ep.Host
 	if ep.Port != DefaultPort {
-		host = fmt.Sprintf("%s:%d", ep.Host, ep.Port)
+		host = net.JoinHostPort(ep.Host, strconv.Itoa(ep.Port))
```
soniasingla, line 96: @haircommander can you help if i used correct conv here and in other cases

haircommander: in general it looks good to me. though this change and every other in the vendor directory should be dropped :)

PR #4877, opened 2021-05-07
 Attempt to fix some errors under bugsmash assignemnt 

soniasingla:


haircommander:
Thanks for the submission, but I'm afraid we shouldn't be changing anything in the vendor directory. I tried to change the muse config to stop reflecting those entries, but it didn't quite work. I'll have to close this, though

PR #4879, opened 2021-05-07
runtimeVM: Use internal context to ensure goroutines are stopped

### LINKED ISSUE #4798: area/vm: pods are not reaching "completed" state

fidencio:
#### What type of PR is this?

/kind bug

https://github.com/cri-o/cri-o/pull/4803 partially reverted
https://github.com/cri-o/cri-o/pull/4650, but ended up missing two
places where the context is used to ensure the goroutines are stopped.

Fixes #4798 

This was caught due to some tests failing in a non deterministic manner
on kata-containers CI.

Related:
https://github.com/kata-containers/tests/pull/3503#issuecomment-834717567

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -331,7 +331,7 @@ func (r *runtimeVM) execContainerCommon(ctx context.Context, c *Container, cmd [
 	defer log.Debugf(ctx, "runtimeVM.execContainerCommon() end")
 
 	// Cancel the context before returning to ensure goroutines are stopped.
-	ctx, cancel := context.WithCancel(ctx)
+	ctx, cancel := context.WithCancel(r.ctx)
 	defer cancel()
 
 	// Generate a unique execID
@@ -506,7 +506,7 @@ func (r *runtimeVM) StopContainer(ctx context.Context, c *Container, timeout int
 	}
 
 	// Cancel the context before returning to ensure goroutines are stopped.
-	ctx, cancel := context.WithCancel(ctx)
+	ctx, cancel := context.WithCancel(r.ctx)
 	defer cancel()
 
 	stopCh := make(chan error)


TomSweeneyRedHat:
LGTM 
assuming happy tests

saschagrunert (approved):
/lgtm
/cherry-pick release-1.21

fidencio:
@haircommander, shall I override the integration_crun CI again?

PR #4885, opened 2021-05-10
storage: use volatile containers

giuseppe:
overlay supports a new mount option "volatile" that reduces I/O by
ignoring fsync and syncfs requests.

Enable it for CRI-O containers as they are "volatile" by nature and
are not supposed to survive a reboot.

Even in the case containers created by CRI-O are persistent, we
already have code in place to throw away the storage if the shutdown
was not clean, so they will be cleaned up in any case.

#### What type of PR is this?

/kind feature

Enable `volatile` for the overlay storage.

```release-note
Enable the  "volatile" option for the overlay drivers when it is supported by the underlying kernel.
```

### DIFF
--- a/internal/storage/runtime.go
+++ b/internal/storage/runtime.go
@@ -258,6 +258,7 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
 
 	coptions := storage.ContainerOptions{
 		LabelOpts: labelOptions,
+		Volatile:  true,
 	}
 	if idMappingsOptions != nil {
 		coptions.IDMappingOptions = *idMappingsOptions


saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests

haircommander:
/hold

actually I'm not sure about this. We rely on the container storage to cleanup CNI resources on reboot now (see https://github.com/cri-o/cri-o/pull/4767). Does the volatile option apply to the config.json as well (the information we'd need to restore a pod to then call CNI del)?

mrunalp:
@giuseppe This only applies to the rootfs?

haircommander:
looking into it, I misuderstood a bit what volatile did. Still, we need to be able to restore containers on reboot (or at least call CNI del on ones we fail to restore).

rhatdan:
This will only effect the overlay storage layer, not any of the other artifacts of the container. If the system crashes some of the content in the container image could be lost, but nothing else.

giuseppe:
yes, "volatile' affects only the rootfs of the container

internal/storage/runtime.go
```diff
@@ -258,6 +258,7 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
 
 	coptions := storage.ContainerOptions{
 		LabelOpts: labelOptions,
+		Volatile:  true,
```
mrunalp, line 261: Can we do this through a conf option? That way we can test and back out if needed easily.

haircommander: can't we pass it in storage_opts in storage.conf?

rhatdan: No because, we only want it for certain volume mounts.

giuseppe: what we could do is to have a global configuration `disable_volatile` that always disables it

rhatdan: Would this be ignored if the caller asked for it?  IE Podman run --rm?

giuseppe: yes, if the user disables completely support for "volatile" then it won't be used for any container.

PR: https://github.com/containers/storage/pull/903

I think it is just a last resort useful to CRI-O

giuseppe:
anything more holding this PR?

haircommander:
/hold cancel

My previous concerns are alleviated, LGTM

giuseppe:
can I've my lgtm back? :-)

PR #4900, opened 2021-05-14
server: rework internal wipe to play better with reboots

haircommander:
#### What type of PR is this?

/kind bug
Embarrassingly, the main use-case of internal wipe was one that I had not tested appropriately. It turns out, we never actually have any containers wipeIfAppropriate after we call restore.
This is because the managed ns no longer exists (it is in /run, a tmpfs). Even if its not in a tmpfs, it is not correctly mounted on reboot.

this is okay, though. It reduces complexity a bit. Now, instead of using the version_file to determine whether we rebooted, we just have to make sure we call CNI DEL as much as possible if we fail to restore a sandbox.

To do this, we must rework LoadSandbox to return the sandbox in question (as soon as possible, basically after unmarshalling all the data from the config.json), then call networkStop on that sandbox (inside a resource cleaner so we retry if the node is booting up).

As a result of this, we can simplify wipeIfAppropriate. It now must be passed a list of images, but will not need to do any thinking about containers. Another consequence is we can eventually deprecate version_file.

This PR also ignores some error cases that were not useful to log.

```release-note
Fix a bug in `internal_wipe` that would mean CNI resources would be leaked across reboots.
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -18,6 +18,7 @@ import (
 
 	imageTypes "github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/idtools"
+	storageTypes "github.com/containers/storage/types"
 	"github.com/cri-o/cri-o/internal/hostport"
 	"github.com/cri-o/cri-o/internal/lib"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
@@ -202,7 +203,7 @@ func (s *Server) restore(ctx context.Context) []string {
 		}
 		log.Warnf(ctx, "could not restore sandbox %s container %s: %v", metadata.PodID, sbID, err)
 		for _, n := range names[sbID] {
-			if err := s.Store().DeleteContainer(n); err != nil {
+			if err := s.Store().DeleteContainer(n); err != nil && err != storageTypes.ErrNotAContainer {
 				log.Warnf(ctx, "unable to delete container %s: %v", n, err)
 			}
 			// Release the infra container name and the pod name for future use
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -160,7 +160,12 @@ func (s *Server) getPortForward(req *types.PortForwardRequest) (*types.PortForwa
 	return s.stream.streamServer.GetPortForward(req)
 }
 
-func (s *Server) restore(ctx context.Context) {
+// restore attempts to restore the sandboxes and containers.
+// For every sandbox it fails to restore, it starts a cleanup routine attempting to call CNI DEL
+// For ever container it fails to restore, it returns that containers image, so that
```
mrunalp, line 166: nit: typo ever -> every

server/server.go
```diff
@@ -202,7 +203,7 @@ func (s *Server) restore(ctx context.Context) []string {
 		}
 		log.Warnf(ctx, "could not restore sandbox %s container %s: %v", metadata.PodID, sbID, err)
 		for _, n := range names[sbID] {
-			if err := s.Store().DeleteContainer(n); err != nil {
+			if err := s.Store().DeleteContainer(n); err != nil && err != storageTypes.ErrNotAContainer {
```
mrunalp, line 206: In what situations would this error be returned by storage?

test/crio-wipe.bats
```diff
@@ -57,6 +60,12 @@ function test_crio_did_not_wipe_images() {
 	[[ "$output" == *"$IMAGE_USED"* ]]
 }
 
+# simluate a reboot by unmounting and removing the namespaces
```
mrunalp, line 63: typo simulate

server/server.go
haircommander: The names stored for a container contain both the container ID and the container name. This comes up when we attempt to delete the container's name

test/crio-wipe.bats
haircommander: fixed

server/server.go
haircommander: good catch, fixed

haircommander:
/retest

PTAL @saschagrunert @mrunalp @umohnani8 @QiWang19 @wgahnagl

saschagrunert (approved):
LGTM

PR #4921, opened 2021-05-19
try again on EAGAIN from dbus

haircommander:
#### What type of PR is this?

/kind bug
when setting up conmon's cgroup

```release-note
Re-attempt setting up conmon's cgroup if it fails on EAGAIN from dbus
```

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -69,9 +69,14 @@ func RunUnderSystemdScope(pid int, slice, unitName string, properties ...systemd
 		properties = append(properties, systemdDbus.PropSlice(slice))
 	}
 	ch := make(chan string)
-	_, err = conn.StartTransientUnitContext(ctx, unitName, "replace", properties, ch)
-	if err != nil {
-		return err
+	for {
+		_, err = conn.StartTransientUnitContext(ctx, unitName, "replace", properties, ch)
+		if err == nil {
+			break
+		}
+		if !errors.Is(err, syscall.EAGAIN) {
+			return err
+		}
 	}
 	defer conn.Close()
 


utils/utils.go
```diff
@@ -67,9 +67,12 @@ func RunUnderSystemdScope(pid int, slice, unitName string, properties ...systemd
 		properties = append(properties, systemdDbus.PropSlice(slice))
 	}
 	ch := make(chan string)
-	_, err = conn.StartTransientUnit(unitName, "replace", properties, ch)
-	if err != nil {
-		return err
+	err = syscall.EAGAIN
+	for errors.Is(err, syscall.EAGAIN) {
```
mrunalp, line 71: How about this instead?
```
for {
    _, err = conn.StartTransientUnit(unitName, "replace", properties, ch)
    if err == nil {
        break
    } else {
        if !errors.Is(err, syscall.EAGAIN) {
            return err
        }
    }
}
```

haircommander: the curse of no do-whiles...

(done)

cgwalters:
I looked at this a bit out of curiosity.  It seems surprising that we'd see `EAGAIN` here, because I'd expect the Go runtime to retry in that scenario.  It looks like it's happening on the initial connection.

One thing I'd recommend is somewhere early in process startup, acquire a reference to the system bus to ensure the connection is set up first.  I think this is doing that implicity.

In other words call this: https://github.com/godbus/dbus/blob/e6cb346e4ee46b3c65ab915ceeb0f3739c563c9b/_examples/bluetooth_introspect.go#L11

Issue #4922, opened 2021-05-20
Drop ALL and add few caps doesn't work


mrunalp:
Creating a pod:

```
apiVersion: v1
kind: Pod
metadata:
  labels:
    run: shell
  name: shell
spec:
  containers:
  - image: busybox:1.24
    name: shell
    stdin: true
    tty: true
    securityContext:
      capabilities:
        drop:
          - ALL
        add:
          - NET_BIND_SERVICE
  dnsPolicy: ClusterFirst
  restartPolicy: Always
```

The running pod will include the default configured capabilities + NET_BIND_SERVICE when it should just be NET_BIND_SERVICE.

PR #4923, opened 2021-05-20
 Add support to drop ALL and add back few capabilities

mrunalp:
The current implemention has an issue where we add the default
capabilties when the user does drop ALL and add few.

#### What type of PR is this?

/kind bug

#4922 

```release-note
Fix drop ALL and add back few caps behavior to not include the default configured capabilities
```

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -446,11 +446,8 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 
 	// Add capabilities from crio.conf if default_capabilities is defined
 	capabilities := &types.Capability{}
-	if s.config.DefaultCapabilities != nil {
-		g.ClearProcessCapabilities()
-		capabilities.AddCapabilities = append(capabilities.AddCapabilities, s.config.DefaultCapabilities...)
-	}
-	if err := setupCapabilities(g, capabilities); err != nil {
+	g.ClearProcessCapabilities()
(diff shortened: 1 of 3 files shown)


saschagrunert (approved):
Code LGTM, we may add an integration test for the `ALL` cases

server/sandbox_run_linux.go
```diff
@@ -450,7 +450,7 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 		g.ClearProcessCapabilities()
 		capabilities.AddCapabilities = append(capabilities.AddCapabilities, s.config.DefaultCapabilities...)
```
haircommander, line 451: do we still need this line if we're adding them in setupCapabilities?

haircommander:
+1 on the tests. My preference would be to move this to pkg/container and add unit tests there, but that's not blocking :slightly_smiling_face:

server/sandbox_run_linux.go
mrunalp: Yeah, need to drop this. Fixing.

mrunalp:
For the tests, I was thinking that we add them in critest.

haircommander:
/lgtm

since we'll do tests as a follow up.

/cherry-pick release-1.21

server/sandbox_run_linux.go
mrunalp: done

PR #4929, opened 2021-05-20
server: reduce log verbosity and speed up startup on restore

haircommander:
#### What type of PR is this?

/kind bug
This PR has two QOL commits for internal wipe
The first commit simply reduces the verbosity of the logs (by not calling two warns on a failed LoadSandbox, as well as not re-calling DeleteContainer/ReleaseName for each previously deleted container

the second commit significantly speeds up crio's start up by not attempting to call CNI del before adding it as a cleanup func.

```release-note
Fix a bug where server startup was significantly slowed down by attempting to clean up CNI resources after a reboot.
```


mrunalp (approved):
/lgtm

PR #4943, opened 2021-05-25
oci: do not use conmon for exec sync

haircommander:
#### What type of PR is this?

/kind design
It is not really needed, and causes unnecessary overhead. Plus, we drop an unkillable child from the mix
allowing cri-o to keep track of its children better.

```release-note
ExecSync requests now don't use conmon, instead calling the runtime directly, which reduces overhead.
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -18,6 +18,7 @@ import (
 	"github.com/containers/storage/pkg/pools"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/pkg/config"
+	"github.com/cri-o/cri-o/server/cri/types"
 	"github.com/cri-o/cri-o/server/metrics"
 	"github.com/cri-o/cri-o/utils"
 	"github.com/fsnotify/fsnotify"
@@ -319,17 +320,14 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
 }
 
 // ExecSyncContainer execs a command in a container and returns it's stdout, stderr and return code.
-func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, command []string, timeout int64) (*ExecSyncResponse, error) {
+func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, command []string, timeout int64) (*types.ExecSyncResponse, error) {
 	if c.Spoofed() {
 		return nil, nil
 	}
 
 	processFile, err := prepareProcessExec(c, command, c.terminal)
 	if err != nil {
-		return nil, &ExecSyncError{
-			ExitCode: -1,
-			Err:      err,
-		}
+		return nil, err
 	}
 	defer os.RemoveAll(processFile)
 
@@ -348,12 +346,7 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 
 	err = cmd.Start()
 	if err != nil {
-		return nil, &ExecSyncError{
-			Stdout:   stdoutBuf,
-			Stderr:   stderrBuf,
-			ExitCode: -1,
-			Err:      err,
-		}
+		return nil, err
 	}
 
 	// wait till the command is done
@@ -374,7 +367,7 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 			// If the command timed out, we should return an ExecSyncResponse with a non-zero exit code because
 			// the prober code in the kubelet checks for it. If we return a custom error,
 			// then the probes transition into Unknown status and the container isn't restarted as expected.
-			return &ExecSyncResponse{
+			return &types.ExecSyncResponse{
 				Stderr:   []byte(conmonconfig.TimedOutMessage),
 				ExitCode: -1,
 			}, nil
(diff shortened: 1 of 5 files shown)


internal/oci/runtime_oci.go
```diff
@@ -484,85 +359,44 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
...
-	stdoutBytes, stderrBytes := parseLog(ctx, logBytes)
 	return &ExecSyncResponse{
-		Stdout:   stdoutBytes,
-		Stderr:   stderrBytes,
-		ExitCode: ec.ExitCode,
-	}, nil
+		Stdout:   stdout,
+		Stderr:   stderr,
+		ExitCode: exitCode,
+	}, waitErr
```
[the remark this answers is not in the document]
mrunalp: Please check if the kubelet will tolerate this change.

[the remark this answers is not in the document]
haircommander: there's something awry with the code. I am working on getting critests passing, and that will likely involve changing this block

internal/oci/runtime_oci.go
```diff
@@ -456,26 +339,18 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
...
 	defer os.RemoveAll(processFile)
 
-	args = append(args,
-		"--exec-process-spec", processFile,
-		"--runtime-arg", fmt.Sprintf("%s=%s", rootFlag, r.root))
-
-	cmd := exec.Command(r.config.Conmon, args...) // nolint: gosec
+	execCtx := ctx
+	if timeout > 0 {
+		execCtx, _ = context.WithTimeout(ctx, time.Second*time.Duration(timeout))
```
saschagrunert, line 344: We should probably cancel this new context once the exec is done.

haircommander: good point, fixed!

internal/oci/runtime_oci.go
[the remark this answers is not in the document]
haircommander: I just spent nearly a whole day looking for
```
        // gather exit code from waitErr
        exitCode := int32(0)
        if waitErr != nil {
-               if exitError, ok := err.(*exec.ExitError); ok {
+               if exitError, ok := waitErr.(*exec.ExitError); ok {
                        exitCode = int32(exitError.ExitCode())
                }
        }
```
:cry:

haircommander:
goodness this is a rabbit hole
we needed to handle the case where the command times out but the container process still runs. since the container process isn't a child of the runtime process, we can't kill the exec.Command struct, but rather find the process with pidFile and kill that (what conmon does).

I think this passes critest now, hard to tell because I keep getting rate limited. no idea what's up with the openshift jenkins tests either...

internal/oci/runtime_oci.go
```diff
@@ -393,176 +320,138 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
...
+	// TODO(haircommander): There does exist a race that we could time out before the runtime actually creates the file.
+	// Should we do inotify on this file to ensure it exists? Is that overkill?
+	ctrPid, ctrPgid, err := pidAndpgidFromFile(pidFile)
 	if err != nil {
-		return nil, &ExecSyncError{
-			Stdout:   stdoutBuf,
-			Stderr:   stderrBuf,
-			ExitCode: -1,
-			Err:      err,
+		log.Errorf(ctx, "failed to get pid (%d) or pgid (%d) from file %s: %v", ctrPid, ctrPgid, pidFile, err)
```
saschagrunert, line 426:
```suggestion
		log.Errorf(ctx, "Failed to get pid (%d) or pgid (%d) from file %s: %v", ctrPid, ctrPgid, pidFile, err)
		return
```

internal/oci/runtime_oci.go
```diff
@@ -393,176 +320,138 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
...
+		// First attempt to kill the container group
+		if err := syscall.Kill(-ctrPgid, syscall.SIGKILL); err != nil {
+			log.Errorf(ctx, "Failed to kill process group after timeout: %v", err)
+		}
+	} else if ctrPid > 0 {
+		// If that fails, kill the container PID itself
+		if err := syscall.Kill(ctrPid, syscall.SIGKILL); err != nil {
+			log.Errorf(ctx, "Failed to kill process after timeout: %v", err)
 		}
 	}
```
saschagrunert, line 440:
```suggestion
```

saschagrunert (approved):
Just two nits, otherwise LGTM

internal/oci/runtime_oci.go
haircommander: fixed!

internal/oci/runtime_oci.go
haircommander: fixed!

internal/oci/runtime_oci.go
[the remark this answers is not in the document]
saschagrunert: Oh, nice catch

saschagrunert (approved):
/lgtm
/hold
@mrunalp @haircommander feel free to lift the hold when ready.

saschagrunert:
Backported manually in https://github.com/cri-o/cri-o/pull/4953 and https://github.com/cri-o/cri-o/pull/4954

PR #4953, opened 2021-05-31
[1.19] oci: do not use conmon for exec sync

saschagrunert:
#### What type of PR is this?

/kind bug

It is not really needed, and causes unnecessary overhead. Plus, we drop an unkillable child from the mix
allowing cri-o to keep track of its children better.

Backport of https://github.com/cri-o/cri-o/pull/4943

eebef460c83f76db187f5c404267d09100ff347c is not part of this backport, because we do not have internal types at this point in time.

```release-note
ExecSync requests now don't use conmon, instead calling the runtime directly, which reduces overhead.
```

### DIFF
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -49,9 +49,9 @@ type Runtime struct {
 type RuntimeImpl interface {
 	CreateContainer(*Container, string) error
 	StartContainer(*Container) error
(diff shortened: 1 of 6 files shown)


saschagrunert:
@mrunalp @haircommander PTAL

PR #4974, opened 2021-06-07
cgmgr: reuse dbus connection

haircommander:
#### What type of PR is this?

/kind bug
reuse dbus connection and attempt to reconnect if we get a dbus error
this PR heavily adapts https://github.com/opencontainers/runc/blob/7362fa2d282feffb9b19911150e01e390a23899d/libcontainer/cgroups/systemd

eventually, the plan is to upstream the internal/dbusmgr package to godbus, but we're picking this here first

```release-note
Reuse connection when connecting to dbus, as well as reattempt the connection if it fails temporarily
```

### DIFF
--- a/internal/dbusmgr/dbusmgr.go
+++ b/internal/dbusmgr/dbusmgr.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"syscall"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
 	dbus "github.com/godbus/dbus/v5"
@@ -80,9 +81,16 @@ func (d *DbusConnManager) RetryOnDisconnect(op func(*systemdDbus.Conn) error) er
 			return err
 		}
 		err = op(conn)
+		if err == nil {
+			return nil
+		}
+		if errors.Is(err, syscall.EAGAIN) {
+			continue
+		}
 		if !errors.Is(err, dbus.ErrClosed) {
 			return err
 		}
+		// dbus connection closed, we should reconnect and try again
 		d.resetConnection(conn)
 	}
 }
--- a/internal/dbusmgr/dbusmgr.go
+++ b/internal/dbusmgr/dbusmgr.go
@@ -25,10 +25,13 @@ type DbusConnManager struct{}
 
 // NewDbusConnManager initializes systemd dbus connection manager.
 func NewDbusConnManager(rootless bool) *DbusConnManager {
+	dbusMu.Lock()
+	defer dbusMu.Unlock()
 	if dbusInited && rootless != dbusRootless {
 		panic("can't have both root and rootless dbus")
 	}
 	dbusRootless = rootless
+	dbusInited = true
 	return &DbusConnManager{}
 }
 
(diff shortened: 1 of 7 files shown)


utils/utils.go
```diff
@@ -70,18 +72,22 @@ func RunUnderSystemdScope(pid int, slice, unitName string, properties ...systemd
...
+		})
+
 		if err == nil {
 			break
 		}
 		if !errors.Is(err, syscall.EAGAIN) {
 			return err
 		}
 	}
-	defer conn.Close()
```
haircommander, line 81: it's not really clear to me how we call close on this anymore

haircommander: wait we're reusing this connection why would we close it :facepalm:

internal/dbusmgr/dbusmgr.go
```diff
@@ -0,0 +1,111 @@
...
+	dbusRootless bool
+)
+
+type DbusConnManager struct {
+}
+
+// NewDbusConnManager initializes systemd dbus connection manager.
+func NewDbusConnManager(rootless bool) *DbusConnManager {
+	if dbusInited && rootless != dbusRootless {
+		panic("can't have both root and rootless dbus")
```
kolyshkin, line 30: Nice, I was thinking about adding something like this but never got to it, mostly because it's not possible in a short-lived binary like `runc` (which is either rootless or not, can't be both).

internal/dbusmgr/dbusmgr.go
```diff
@@ -0,0 +1,111 @@
...
+
+var (
+	dbusC        *systemdDbus.Conn
+	dbusMu       sync.RWMutex
+	dbusInited   bool
+	dbusRootless bool
+)
+
+type DbusConnManager struct {
+}
```
kolyshkin, line 25: I guess it needs to be
```go
type DbusConnManager struct {}
```

(from the `gofumpt` POV)

internal/dbusmgr/dbusmgr.go
kolyshkin: But you do need to
(1) take a mutex before getting to these variables, otherwise races are possible (if two goroutines call `NewDbusConnManager`)
(2) do set `dbusInited = true` -- otherwise this check never works.

haircommander: hahaha then you can thank your past-self https://github.com/opencontainers/runc/blame/2f8e8e9d9775003bf76730a8a1c10f29b0000edf/libcontainer/cgroups/systemd/dbus.go#L25

utils/utils.go
```diff
@@ -70,18 +72,22 @@ func RunUnderSystemdScope(pid int, slice, unitName string, properties ...systemd
...
 
 	// Block until job is started
 	<-ch
+	close(ch)
```
kolyshkin, line 90: This belongs to a separate commit I think.

kolyshkin (changes requested):
LGTM aside from the hunk which do not belong to the commit.

Given my history with this feature, I think it might make sense to at least manually test it.

The test would be to start/stop lots of containers, doing `while :; do systemctl restart dbus.service; done` in parallel. Before this commit, this should inevitably result in errors from container start (i.e. `RunUnderSystemdScope`). This needs to be checked, too, that the test is valid.

Once test validity is established, test the patched version to make sure the error is no more.

internal/dbusmgr/dbusmgr.go
kolyshkin: 🤦🏻

kolyshkin: https://github.com/opencontainers/runc/pull/3006

kolyshkin (approved):
LGTM

mrunalp (approved):
/lgtm

PR #4983, opened 2021-06-08
Disable short name mode

umohnani8:

Disable the short name mode so we don't use the alias table
when pulling images. We will revisit this is in the next release.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1969599

```release-note
None
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -761,10 +761,9 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


mrunalp (approved):
/lgtm

mtrmac:
> Disable the short name mode so we don't use the alias table when pulling images

For the record, that’s not what that option does; it doesn’t affect alias resolution in any way, and assuming CRI-O is run in non-interactive mode without a TTY attached, this PR has absolutely no effect.

PR #4999, opened 2021-06-15
oci: fix issues with exec

haircommander:
#### What type of PR is this?

/kind bug
#### What this PR does / why we need it
previously we had lots of instances of
`Failed to get pid (-1) or pgid (-1) from file /tmp/pidfile097828553: strconv.Atoi: parsing \"\": invalid syntax"`
because we precreated the pid file. stop doing this, and also kill the runtime process if we timeout an exec run

```release-note
Fix a bug where an exec sync timeout would fail to cleanup the runtime exec process
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -331,11 +331,13 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 	}
 	defer os.RemoveAll(processFile)
 
-	pidFile, err := createPidFile()
+	pidDir, err := ioutil.TempDir("", "pidfile")
 	if err != nil {
 		return nil, err
 	}
-	defer os.RemoveAll(pidFile)
+	defer os.RemoveAll(pidDir)
+
+	pidFile := filepath.Join(pidDir, c.id)
 
 	cmd := r.constructExecCommand(ctx, c, processFile, pidFile)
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_oci.go
```diff
@@ -417,13 +419,15 @@ func createPidFile() (string, error) {
...
-	// TODO(haircommander): There does exist a race that we could time out before the runtime actually creates the file.
-	// Should we do inotify on this file to ensure it exists? Is that overkill?
 	ctrPid, ctrPgid, err := pidAndpgidFromFile(pidFile)
-	if err != nil {
-		log.Errorf(ctx, "Failed to get pid (%d) or pgid (%d) from file %s: %v", ctrPid, ctrPgid, pidFile, err)
+	if err != nil && ctrPid <= 0 {
+		// only kill the runtime process if we failed to find a ctrPid
+		// as this means the runtime exec hasn't successfully written the pid file
+		if killErr := cmd.Process.Kill(); killErr != nil {
+			log.Errorf(ctx, "Error killing runtime exec process(%v) after error finding runtime pid: (%v)", killErr, err)
```
mrunalp, line 429: nit: `after error finding container exec pid" (can be done later).

mrunalp (commented):
We also want to address not waiting indefinitely after the SIGKILL  as a follow-on as process could get stuck in D state.

PR #5007, opened 2021-06-16
Comments default values instead of omitting them

### LINKED ISSUE #4770: 1.21.0 installed config is empty
This was intentional, but it is not optimal.

For 1.21.0, we changed the `crio config` command to *only* print fields that are different than the default. This allows config migration to be easier, and config changes as well.

However, now running `make install.config` (which all packaged solutions do, as well as our recommendation for installation from source) installs an empty file--it's all the default!

This is sub-par--often users want to use the default config as a way to see the available options.

I take the blame for this one, I guided the solution to https://github.com/cri-o/cri-o/issues/4506 without checking the reprocussions and making sure all stakeholders were okay with the fix.

The good news is: we have another good first issue! 

Here are three proposed pieces:
- `crio config` should print *all the fields*, with the default ones commented out, instead of omitted. This will be used to populate the default configuration
- `crio config --all` should print *all the fields*, with both default and non-default commented out
- `crio config --verbose` should print only the non-default fields. This will be used for generating drop-in files

wgahnagl:
changes the default crio config to print default config values as comments instead of ommitting them

Fixes #4770 

```release-note
changes default config output to comment default values instead of omitting them

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,9 @@ check-log-lines:
 	./hack/log-capitalized.sh
 	./hack/tree_status.sh
 
+check-config-template: 
+	./hack/validate-config.sh
+
 shellfiles: ${SHFMT}
 	$(eval SHELLFILES=$(shell ${SHFMT} -f . | grep -v vendor/ | grep -v hack/lib | grep -v hack/build-rpms.sh))
 
--- a/hack/validate-config.sh
+++ b/hack/validate-config.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+if grep -q 'Comment' pkg/config/template.go | grep -vq '{{ $.Comment }}'; then
+    exit 1
+fi
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -3,6 +3,7 @@ package config
 import (
 	"io"
 	"reflect"
+	"strings"
 	"text/template"
 )
 
@@ -60,7 +61,9 @@ func crioTemplateString(group templateGroup, prefix string, displayAll bool, cri
 
 	for _, configItem := range crioTemplateConfig {
 		if group == configItem.group {
-			if displayAll || !configItem.isDefaultValue {
+			if !configItem.isDefaultValue || displayAll {
+				templateString += strings.ReplaceAll(configItem.templateString, "{{ $.Comment }}", "")
+			} else {
 				templateString += configItem.templateString
 			}
 		}
@@ -618,58 +621,58 @@ const templateStringCrio = `[crio]
 
 const templateStringCrioRoot = `# Path to the "root directory". CRI-O stores all of its data, including
 # containers images, in this directory.
-root = "{{ .Root }}"
+{{ $.Comment }}root = "{{ .Root }}"
 
 `
 
 const templateStringCrioRunroot = `# Path to the "run directory". CRI-O stores all of its state in this directory.
-runroot = "{{ .RunRoot }}"
+{{ $.Comment }}runroot = "{{ .RunRoot }}"
 
 `
 
 const templateStringCrioStorageDriver = `# Storage driver used to manage the storage of images and containers. Please
 # refer to containers-storage.conf(5) to see all available storage drivers.
-storage_driver = "{{ .Storage }}"
+{{ $.Comment }}storage_driver = "{{ .Storage }}"
 
 `
 
 const templateStringCrioStorageOption = `# List to pass options to the storage driver. Please refer to
 # containers-storage.conf(5) to see all available storage options.
-storage_option = [
-{{ range $opt := .StorageOptions }}{{ printf "\t%q,\n" $opt }}{{ end }}]
+{{ $.Comment }}storage_option = [
+{{ range $opt := .StorageOptions }}{{ $.Comment }}{{ printf "\t%q,\n" $opt }}{{ end }}{{ $.Comment }}]
 
 `
 
 const templateStringCrioLogDir = `# The default log directory where all logs will go unless directly specified by
 # the kubelet. The log directory specified must be an absolute directory.
-log_dir = "{{ .LogDir }}"
+{{ $.Comment }}log_dir = "{{ .LogDir }}"
 
 `
 
 const templateStringCrioVersionFile = `# Location for CRI-O to lay down the temporary version file.
 # It is used to check if crio wipe should wipe containers, which should
 # always happen on a node reboot
-version_file = "{{ .VersionFile }}"
+{{ $.Comment }}version_file = "{{ .VersionFile }}"
 
 `
 
 const templateStringCrioVersionFilePersist = `# Location for CRI-O to lay down the persistent version file.
 # It is used to check if crio wipe should wipe images, which should
 # only happen when CRI-O has been upgraded
-version_file_persist = "{{ .VersionFilePersist }}"
+{{ $.Comment }}version_file_persist = "{{ .VersionFilePersist }}"
 
 `
 
 const templateStringCrioCleanShutdownFile = `# Location for CRI-O to lay down the clean shutdown file.
 # It is used to check whether crio had time to sync before shutting down.
 # If not found, crio wipe will clear the storage directory.
-clean_shutdown_file = "{{ .CleanShutdownFile }}"
+{{ $.Comment }}clean_shutdown_file = "{{ .CleanShutdownFile }}"
 
 `
 
 const templateStringCrioInternalWipe = `# InternalWipe is whether CRI-O should wipe containers and images after a reboot when the server starts.
 # If set to false, one must use the external command 'crio wipe' to wipe the containers and images in these situations.
-internal_wipe = {{ .InternalWipe }}
+{{ $.Comment }}internal_wipe = {{ .InternalWipe }}
 
 `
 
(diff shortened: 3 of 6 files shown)


saschagrunert (approved):
/okt-ot-test

saschagrunert:
I like that one! The integration tests still do not seem to be happy, yet.

pkg/config/template.go
```diff
@@ -53,7 +54,9 @@ func crioTemplateString(group templateGroup, prefix string, displayAll bool, cri
 
 	for _, configItem := range crioTemplateConfig {
 		if group == configItem.group {
-			if displayAll || !configItem.isDefaultValue {
+			if !configItem.isDefaultValue {
+				templateString += strings.Replace(configItem.templateString, "{{ .Comment }} ", "", 1)
```
haircommander, line 58: I'm not seeing many cases where {{ .Comment }} is followed by a space, we may need to replace this with 
`strings.Replace(configItem.templateString, "{{ .Comment }}", "", 1)`

haircommander: also we need this to be `-1` so it replaces all instances of Comment

pkg/config/template.go
```diff
@@ -553,52 +556,53 @@ const templateStringCrio = `[crio]
...
 
 `
 
 const templateStringCrioStorageOption = `# List to pass options to the storage driver. Please refer to
 # containers-storage.conf(5) to see all available storage options.
-storage_option = [
-{{ range $opt := .StorageOptions }}{{ printf "\t%q,\n" $opt }}{{ end }}]
+
+{{ .Comment }}storage_option = [
+{{ range $opt := .StorageOptions }}{{ $.Comment }}{{ printf "\t%q,\n" $opt }}{{ end }}{{ $.Comment }}]
```
haircommander, line 578: will this $ get in the way of the Replace above?

pkg/config/template.go
```diff
@@ -609,135 +613,135 @@ internal_wipe = {{ .InternalWipe }}
...
 
 `
 
 const templateStringCrioRuntimeDefaultUlimits = `# A list of ulimits to be set in containers by default, specified as
 # "<ulimit name>=<soft limit>:<hard limit>", for example:
 # "nofile=1024:2048"
 # If nothing is set here, settings will be inherited from the CRI-O daemon
-default_ulimits = [
-{{ range $ulimit := .DefaultUlimits }}{{ printf "\t%q,\n" $ulimit }}{{ end }}]
+{{ .Comment }}default_ulimits = [
```
haircommander, line 686: same here, we may want to replace {{ $.Comment }} as well

pkg/config/template.go
```diff
@@ -962,19 +966,19 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
...
-runtime_root = "{{ $runtime_handler.RuntimeRoot }}"
-{{ if $runtime_handler.PrivilegedWithoutHostDevices }}
-privileged_without_host_devices = {{ $runtime_handler.PrivilegedWithoutHostDevices }}
-{{ end }}
-{{ if $runtime_handler.AllowedAnnotations }}
-allowed_annotations = [
-{{ range $opt := $runtime_handler.AllowedAnnotations }}{{ printf "\t%q,\n" $opt }}{{ end }}]
-{{ end }}
-{{ end }}
+{{ .Comment }}{{ range $runtime_name, $runtime_handler := .Runtimes  }}
```
haircommander, line 969: I don't think you need the comment here, this range function won't print anything on this line

pkg/config/template.go
```diff
@@ -962,19 +966,19 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
...
+{{ $.Comment }}[crio.runtime.runtimes.{{ $runtime_name }}]
+{{ $.Comment }}runtime_path = "{{ $runtime_handler.RuntimePath }}"
+{{ $.Comment }}runtime_type = "{{ $runtime_handler.RuntimeType }}"
+{{ $.Comment }}runtime_root = "{{ $runtime_handler.RuntimeRoot }}"
+{{ $.Comment }}{{ if $runtime_handler.PrivilegedWithoutHostDevices }}
```
haircommander, line 974: same here

haircommander:
@wgahnagl can you sign off your commit?

TomSweeneyRedHat:
LGTM
assuming happy tests

fidencio:
@haircommander, seems the kata-containers failure is legit.  I'll take some time to take a deeper look into this Tomorrow.  Please, do **not** override the kata-containers CI.

pkg/config/template.go
fidencio: As part of kata-containers CI we generate the crio.conf file by calling `crio -d "" --log-level "debug" --cgroup-manager "systemd" --storage-driver "overlay" --storage-opt "overlay.override_kernel_check=1" config > crio.conf`, and this will generate
```
storage_option = [
#       "overlay.override_kernel_check=1",
# ]
```

That's the reason the kata-containers CI is failing.

I'd strongly suggest adding a test case covering something similar.

fidencio (changes requested):
@wgahnagl, the kata-containers CI found a legit issue with the PR.
I've commented inline in the case that we hit, but we may face that in a few other parts of configuration file.

/cc @haircommander

fidencio:
The kata-containers CI is failing due to:
```
-- Logs begin at Wed 2020-05-20 18:55:39 UTC, end at Fri 2021-08-20 17:00:40 UTC. --
Aug 20 17:00:35 fedoradocker32-7603d0 systemd[1]: Starting Container Runtime Interface for OCI (CRI-O)...
Aug 20 17:00:35 fedoradocker32-7603d0 crio[94402]: time="2021-08-20T17:00:35Z" level=info msg="RDT is not enabled: failed to detect resctrl mount point: resctrl not found in /proc/mounts"
Aug 20 17:00:35 fedoradocker32-7603d0 crio[94402]: time="2021-08-20T17:00:35Z" level=fatal msg="unable to decode configuration /etc/crio/crio.conf: Near line 32 (last key parsed 'crio.storage_option'): expected value but found \"rdt\" instead"
Aug 20 17:00:35 fedoradocker32-7603d0 systemd[1]: crio.service: Main process exited, code=exited, status=1/FAILURE
Aug 20 17:00:35 fedoradocker32-7603d0 systemd[1]: crio.service: Failed with result 'exit-code'.
Aug 20 17:00:35 fedoradocker32-7603d0 systemd[1]: Failed to start Container Runtime Interface for OCI (CRI-O).
```

haircommander:
/retest
@fidencio @fgiudici are the kata failures the fault of this PR?

fidencio:
El mar., 24 ago. 2021 0:20, Peter Hunt ***@***.***> escribió:

> /retest
> @fidencio <https://github.com/fidencio> @fgiudici
> <https://github.com/fgiudici> are the kata failures the fault of this PR?
>

Yes, seems to be the same issue.  Have you tried to reproduce it locally
according to the instructions added in one of my comments?

Anyways, I will be back to working next week.  Please, do not override the
CI in this case

fidencio:
```

-- Logs begin at Wed 2020-05-20 18:55:39 UTC, end at Mon 2021-08-23
14:51:51 UTC. --
Aug 23 14:51:46 fedoradocker32-54c430 systemd[1]: Starting Container
Runtime Interface for OCI (CRI-O)...
Aug 23 14:51:46 fedoradocker32-54c430 crio[94909]:
time="2021-08-23T14:51:46Z" level=info msg="RDT is not enabled: failed
to detect resctrl mount point: resctrl not found in /proc/mounts"
Aug 23 14:51:46 fedoradocker32-54c430 crio[94909]:
time="2021-08-23T14:51:46Z" level=fatal msg="unable to decode
configuration /etc/crio/crio.conf: Near line 32 (last key parsed
'crio.storage_option'): expected value but found \"device\" instead"
Aug 23 14:51:46 fedoradocker32-54c430 systemd[1]: crio.service: Main
process exited, code=exited, status=1/FAILURE
Aug 23 14:51:46 fedoradocker32-54c430 systemd[1]: crio.service: Failed
with result 'exit-code'.
Aug 23 14:51:46 fedoradocker32-54c430 systemd[1]: Failed to start
Container Runtime Interface for OCI (CRI-O).

```

TomSweeneyRedHat:
LGTM
assuming tests get happy.

docs/crio.conf.5.md
```diff
@@ -125,7 +125,7 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
   Additional environment variables to set for all the containers. These are overridden if set in the container image spec or in
 the container runtime configuration.
 
-**selinux**=false
+**selinux**=true
```
haircommander, line 128: oops this seems to have slipped in

wgahnagl: make docs-validation fails with 
```go run -tags "containers_image_ostree_stub      containers_image_openpgp seccomp selinux " ./test/docs-va[5/2200]
INFO Verifying TOML tags of `config.go` to `TemplateString` and `docs/crio.conf.5.md`                            
ERRO Tag `selinux` with expected value `true` not found in `docs/crio.conf.5.md` ``` 
if it's not sent to true :thinking:

File: test/config.bats [hunk 7]
```diff
@@ -11,6 +11,12 @@ function teardown() {
 	cleanup_test
 }
 
+@test "default config should be empty" {
+	setup_crio
+	output=$("$CRIO_BINARY_PATH" config "|sed 's/#.*//' | tr -d '\n'")
+	[[ "$output" == *"[crio]"* ]]
```
haircommander, line 17: I believe this will fail now that we're not commenting the headers

haircommander: or rather, it should fail and I am actually not sure it does because we use a wildcard * at the end (thus allowing for anything after [crio]

I thiink this test case needs an update regardless

wgahnagl: should it check to find all of the headers?

haircommander: yeah, or it could check there are no lines that don't start with `[`

haircommander: that may be easier to maintain because we may add headers in the future

fidencio:
I see the kata-containers CI is failing, let me check locally why it's failing, but I'd like to double check that https://github.com/cri-o/cri-o/pull/5007#issuecomment-902968139 was addressed.

test/crio-wipe.bats
```diff
@@ -80,7 +80,8 @@ function start_crio_with_stopped_pod() {
 }
 
 @test "remove containers and images when remove both" {
-	CONTAINER_INTERNAL_WIPE=false start_crio_with_stopped_pod
+	crio -r "$TESTDIR/crio"
+    CONTAINER_INTERNAL_WIPE=false start_crio_with_stopped_pod
```
fidencio, line 84: Please, use a tab instead of 4 spaces here.

fidencio:
> I see the kata-containers CI is failing, let me check locally why it's failing, but I'd like to double check that [#5007 (comment)](https://github.com/cri-o/cri-o/pull/5007#issuecomment-902968139) was addressed.

I don't see this error happening anymore on my local tests, let's see what's going on with the kata-containers CI.

fidencio:
I think we will need to leave this one for next year, I couldn't find a way to reproduce it locally.

@littlejawa, would you mind to take a look at this one at some point?  I'd be up for a chat, in case you can easily reproduce the issues using @wgahnagl's commits.

fidencio:
From the kata-containers CI:
```
20:32:38 # # --- crio.log :: ---
20:32:38 # time="2021-12-22T19:31:48Z" level=fatal msg="unable to decode configuration /tmp/tmp.Svj65qRY4N/crio.conf.d/crio-custom.conf: toml: cannot load TOML value of type string into a Go boolean"
20:32:38 # # --- --- ---
```

It'd be worth adding a debug to show the configuration files, so we could pinpoint what's the string value we're wrongly parsing.  Note, we had a similar issue to this one in the past, with the `privileged_without_host_devices` option.

fidencio:
```
22:02:10 [crio.runtime.runtimes.containerd-shim-kata-v2]
22:02:10 runtime_path = "/usr/local/bin/containerd-shim-kata-v2"
22:02:10 runtime_type = "vm"
22:02:10 runtime_root = "/run/vc"
22:02:10 runtime_config_path = ""
22:02:10 
22:02:10 privileged_without_host_devices = "true"
```

Bingo! 
`privileged_without_host_devices = true|false` should be the correct form.  Let me just check what's setting that up.

pkg/config/template.go
```diff
@@ -1066,18 +1069,18 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
...
+{{ $.Comment }}runtime_config_path = "{{ $runtime_handler.RuntimeConfigPath }}"
 {{ if $runtime_handler.PrivilegedWithoutHostDevices }}
-privileged_without_host_devices = {{ $runtime_handler.PrivilegedWithoutHostDevices }}
-{{ end }}
-{{ if $runtime_handler.AllowedAnnotations }}
-allowed_annotations = [
-{{ range $opt := $runtime_handler.AllowedAnnotations }}{{ printf "\t%q,\n" $opt }}{{ end }}]
-{{ end }}
-{{ end }}
+{{ $.Comment }}privileged_without_host_devices = "{{ $runtime_handler.PrivilegedWithoutHostDevices }}"
```
fidencio, line 1077:
```suggestion
{{ $.Comment }}privileged_without_host_devices = {{ $runtime_handler.PrivilegedWithoutHostDevices }}
```

This should solve the issue faced with the `privileged_without_host_devices = "true"` and, instead, set `privileged_without_host_devices = true`.

It's worth double-checking all the boolean types to ensure the same mistake doesn't repeat itself.

wgahnagl: aha!

test/helpers.bash
```diff
@@ -422,6 +422,7 @@ function cleanup_test() {
         echo "# --- crio.log :: ---"
         cat "$CRIO_LOG"
         echo "# --- --- ---"
+        crio config >&3
```
fidencio, line 425: This bit can be dropped now that we figured out the kata-containers CI issues.

littlejawa:
> @littlejawa, would you mind to take a look at this one at some point? I'd be up for a chat, in case you can easily reproduce the issues using @wgahnagl's commits.

Sure ! But... too late apparently :-/

hack/validate-config.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+output=$(grep 'Comment' pkg/config/template.go | grep -v '{{ $.Comment }}')
+
+if [ -z "$output" ]; then
+    exit 0
+else
+    exit 1
```
haircommander, line 7: nit: no need to evaluate else here, because we exit above

hack/validate-config.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+output=$(grep 'Comment' pkg/config/template.go | grep -v '{{ $.Comment }}')
+
+if [ -z "$output" ]; then
```
haircommander, line 4: nit: we can compresss this without saving to `$output`

hack/validate-config.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+if [ -z "output=$(grep 'Comment' pkg/config/template.go | grep -v '{{ $.Comment }}')" ]; then
+    exit 0
+fi
```
haircommander, line 5: where'd our exit 1 go :cry:

haircommander:
:sunglasses: 
/approve

thanks for keeping up with this! LGTM

@cri-o/cri-o-maintainers PTAL

Makefile
```diff
@@ -144,6 +144,9 @@ check-log-lines:
 	./hack/log-capitalized.sh
 	./hack/tree_status.sh
 
+check-config-template: 
+	./hack/validate-config.sh`
```
saschagrunert, line 148:
```suggestion
	./hack/validate-config.sh
```

hack/validate-config.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
saschagrunert, line 1:
```suggestion
#!/usr/bin/env bash
```

fidencio (approved):
lgtm, thanks @wgahnagl!

haircommander:
cc @saschagrunert for the final LGTM?

saschagrunert (approved):
/lgtm

PR #5012, opened 2021-06-17
Generate self-signed metrics cert/key if provided ones are not available

saschagrunert:
#### What type of PR is this?

/kind feature

This allows starting CRI-O without having the certificates on disk. A Kubernetes controller can later on reconcile their own certificates into the same location for reload.

None
None

```release-note
Added generation of self-signed certificates for the secure metrics endpoint
if the provided cert and key paths are not available on disk.
```

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -3,8 +3,11 @@ package metrics
 import (
 	"crypto/tls"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


server/metrics/metrics.go
```diff
@@ -309,6 +313,37 @@ func newCertReloader(doneChan chan struct{}, certPath, keyPath string) (*certRel
 		keyPath:  keyPath,
 	}
 
+	// Generate self-signed certificate and key if the provided ones are not
+	// available.
+	_, errCertPath := os.Stat(certPath)
+	_, errKeyPath := os.Stat(keyPath)
+	if errCertPath != nil && os.IsNotExist(errCertPath) &&
+		errKeyPath != nil && os.IsNotExist(errKeyPath) {
+		logrus.Info("Metrics key and cert path does not exist, generating self-signed")
```
TomSweeneyRedHat, line 322: Do you need to worry about the case when you have the certpath but not the key, or vice versa?

saschagrunert: Right now it only serves as HTTPs endpoint when both are given as string. If one of both does not exist on disk then we error as usual. I think all cases are already covered.

saschagrunert:
@haircommander PTAL

PR #5026, opened 2021-06-23
network: pass pod UID to ocicni when performing network operations

dcbw:
Ongoing sandbox requests cannot be (or are not) canceled by kubelet, leading to a situation where short-lived pods (especially Kubernetes e2e tests for stateful sets) cause overlapping sandbox requests. If the CNI plugin needs to wait for network state to converge, it's pointless to wait for a sandbox who's pod has been deleted so the plugin should cancel the request and return to the runtime. However, it's impossible to do that race-free without the pod UID the sandbox was created for, since the there is a gap between when kubelet requests the sandbox creation and when the plugin gets the pod object from the apiserver when the pod could have been deleted and recreated, and the CNI plugin would retrieve information for the new pod, not the pod the sandbox was created for.

Passing the pod UID to the plugin allows the plugin to cancel the operation when the pod UID retrieved from the apiserver during plugin operation does not match the one the sandbox was created for.

@trozet @haircommander @mrunalp 

/kind feature

```release-note
CNI plugins are now passed a K8S_POD_UID environment variable containing the pod UID this sandbox was started for.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/coreos/go-systemd/v22 v22.3.2
 	github.com/cpuguy83/go-md2man v1.0.10
 	github.com/creack/pty v1.1.13
-	github.com/cri-o/ocicni v0.2.1-0.20210301205850-541cf7c703cf
+	github.com/cri-o/ocicni v0.2.1-0.20210623033107-4ea5fb8752cf
 	github.com/cyphar/filepath-securejoin v0.2.2
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/go-units v0.4.0
--- a/go.sum
+++ b/go.sum
@@ -373,8 +373,9 @@ github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ3
 github.com/creack/pty v1.1.11/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
 github.com/creack/pty v1.1.13 h1:rTPnd/xocYRjutMfqide2zle1u96upp1gm6eUHKi7us=
 github.com/creack/pty v1.1.13/go.mod h1:MOBLtS5ELjhRRrroQr9kyvTxUAFNvYEK993ew/Vr4O4=
-github.com/cri-o/ocicni v0.2.1-0.20210301205850-541cf7c703cf h1:k2wrxBiBseRfOD7h+9fABEuesABBQuUuW5fWwpARbeI=
 github.com/cri-o/ocicni v0.2.1-0.20210301205850-541cf7c703cf/go.mod h1:vingr1ztOAzP2WyTgGbpMov9dFhbjNxdLtDv0+PhAvY=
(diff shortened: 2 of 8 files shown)


saschagrunert (approved):
LGTM

dcbw:
kata-jenkins failure is:

```
07:21:15 #   Warning  FailedCreatePodSandBox  4s (x7 over 89s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = CreateContainer failed: unrecognised machinetype: pc: unknown
07:21:15 # pod "handlers" deleted
07:21:15 Failed at 69: bats "${K8S_TEST_ENTRY}"
```

which seems unrelated

dcbw:
e2e-gcp failure is:

```
fail [k8s.io/kubernetes@v1.21.1/test/e2e/storage/ubernetes_lite_volumes.go:163]: Unexpected error:     <*errors.errorString \| 0xc00263f220>: {         s: "PersistentVolumeClaims [pvc-1] not all in phase Bound within 5m0s",     }     PersistentVolumeClaims [pvc-1] not all in phase Bound within 5m0s
```

dcbw:
Updated with tests in network.bats

haircommander:
> Ongoing sandbox requests cannot be (or are not) canceled by kubelet, leading to a situation where short-lived pods (especially Kubernetes e2e tests for stateful sets) cause overlapping sandbox requests

weird this sounds like a kubelet bug. I would expect cri-o to fail to create a duplicate pod while the first request is ongoing, and synchronously wait on cni plugin, thus preventing duplicate calls.

this change is fine to me, but I fear we're putting a bandaid on a bigger wound
/lgtm

/override ci/prow/e2e-gcp

dcbw:
> weird this sounds like a kubelet bug. I would expect cri-o to fail to create a duplicate pod while the first request is ongoing, and synchronously wait on cni plugin, thus preventing duplicate calls.
> 
> this change is fine to me, but I fear we're putting a bandaid on a bigger wound

@haircommander the scenarios are something like this:

Scenario 1: pod recreated during sandbox wait
1) pod created, kubelet notices, asks CRI to create the sandbox
2) CRI creates sandbox, execs CNI plugin
3) CNI plugin gets pod from apiserver, starts setting up networking
4) something deletes and recreates the pod
5) CNI plugin waiting for networking to converge

Now in this scenario, the plugin *could* create a pod watch for delete events. But that's not race-proof since the pod could be deleted + recreated between steps 3 and 4 and the sandbox would be for an old pod (and be subsequently torn down by kubelet at some point in the future). Pod UID allows the plugin to notice the pod instance it gets in (3) or (5) is different and exit early.

Scenario 2: pod deleted during sandbox init
1) pod created, kubelet notices, asks CRI to create the sandbox
2) CRI creates sandbox, execs CNI plugin
3) something deletes and recreates the pod
4) CNI plugin gets pod from apiserver, starts setting up networking
5) CNI plugin still waiting for networking

in this scenario, the CNI plugin gets the *new* pod instance, which is still wrong for this sandbox setup. Pod UUID immediately tells the plugin that its pod is gone and it can exit early.

In all cases, kubelet will just tear the sandbox down anyway, and what we're trying to prevent is waiting longer than necessary before noticing that this sandbox is useless.

----

I suppose the real fix would be to allow a sandbox delete + CNI DEL while an existing add was going on, or a CANCEL operation that kubelet could execute to tell the CRI and plugins to stop the request. But that's a much longer arc to make happen (but it should happen, and at least on the CNI side we are working on that via gRPC).

haircommander:
thanks for the explaination @dcbw , makes sense!

PR #5028, opened 2021-06-24
[release-1.21] network: pass pod UID to ocicni when performing network operations

dcbw:
Ongoing sandbox requests cannot be (or are not) canceled by kubelet, leading to a situation where short-lived pods (especially Kubernetes e2e tests for stateful sets) cause overlapping sandbox requests. If the CNI plugin needs to wait for network state to converge, it's pointless to wait for a sandbox who's pod has been deleted so the plugin should cancel the request and return to the runtime. However, it's impossible to do that race-free without the pod UID the sandbox was created for, since the there is a gap between when kubelet requests the sandbox creation and when the plugin gets the pod object from the apiserver when the pod could have been deleted and recreated, and the CNI plugin would retrieve information for the new pod, not the pod the sandbox was created for.

Passing the pod UID to the plugin allows the plugin to cancel the operation when the pod UID retrieved from the apiserver during plugin operation does not match the one the sandbox was created for.

@trozet @haircommander @mrunalp

Cherry-pick of https://github.com/cri-o/cri-o/pull/5026

/kind feature

```release-note
CNI plugins are now passed a K8S_POD_UID environment variable containing the pod UID this sandbox was started for.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/coreos/go-systemd/v22 v22.3.1
 	github.com/cpuguy83/go-md2man v1.0.10
 	github.com/creack/pty v1.1.11
-	github.com/cri-o/ocicni v0.2.1-0.20210301205850-541cf7c703cf
+	github.com/cri-o/ocicni v0.2.1-0.20210623033107-4ea5fb8752cf
 	github.com/cyphar/filepath-securejoin v0.2.2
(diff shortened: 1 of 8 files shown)


dcbw:
@saschagrunert you approved the 1.22 and 1.20 backports, I assume this one for 1.21 is OK too? :) Thanks!

haircommander:
/approve
yeah I assume it's okay with him :)

PR #5034, opened 2021-06-25
[1.20] pull: do check for blocked registries

haircommander:
Do not check whether a registry is blocked.  c/image is taking care of
that implicitly when attempting to copy an image.

#### What type of PR is this?

/kind bug
or else mirrored registries cannot resolve there

```release-note
fix a bug where CRI-O prematurely stopped pulling from a blocked registry, even if there was a mirror for that registry it was allowed to pull from.
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -804,17 +804,7 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 			rem = "library/" + rem
 		}
 		image := r + "/" + rem
-		registry, err := sysregistriesv2.FindRegistry(systemContext, image)
-		if err != nil {
-			return nil, err
-		}
-		if registry != nil && registry.Blocked {
-			continue
-		}
 		images = append(images, image)
-	}
-	if len(images) == 0 {
-		return nil, fmt.Errorf("all search registries for %q are blocked", remainder)
 	}
 	return imageNamesWithDigestOrTag(images)
 }
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -783,14 +783,6 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 
 	domain, remainder := splitDockerDomain(imageName)
 	if domain != "" {
-		// this means the image is already fully qualified
-		registry, err := sysregistriesv2.FindRegistry(systemContext, imageName)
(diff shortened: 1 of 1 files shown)


mtrmac (changes requested):
`.Blocked` is also evaluated below in the unqualified-image case. I feel quite strongly that the two should behave the same.

---

IIRC this behavior difference was historically known when implementing the c/image approach, and the CRI-O code here was left around intentionally at that time, though I’m not 100% certain and I might be misremembering. Is there a specific rationale for either that old decision, or the new decision?

(OTOH I don’t feel strongly about preserving this behavior at all, I’m happy to defer decisions about CRI-O behavior stability to CRI-O maintainers. “The code will be simpler and consistent with other c/image users” might be a perfectly good justification.)

mtrmac (commented):
Thanks! Code LGTM, decision whether to change the behavior is up to you.

haircommander:
> `.Blocked` is also evaluated below in the unqualified-image case. I feel quite strongly that the two should behave the same.
> 
> IIRC this behavior difference was historically known when implementing the c/image approach, and the CRI-O code here was left around intentionally at that time, though I’m not 100% certain and I might be misremembering. Is there a specific rationale for either that old decision, or the new decision?
> 
> (OTOH I don’t feel strongly about preserving this behavior at all, I’m happy to defer decisions about CRI-O behavior stability to CRI-O maintainers. “The code will be simpler and consistent with other c/image users” might be a perfectly good justification.)

My updating this code was mostly because the current behavior wrt the interaction of blocked and mirrored registries is not intuitive to me. I am not married to defering the decision to c/image, but "simpler and consistent" are both also good reasons to do so :)

I removed the .Blocked handling below as well

mtrmac:
> My updating this code was mostly because the current behavior wrt the interaction of blocked and mirrored registries is not intuitive to me. 

The concept of blocking was created before c/image was aware of it, and IIRC before mirrors were a thing. So it’s quite possible that this before-mirror blocking was just inherited after mirrors, and left around because nobody wanted to affirmatively decide make a change to the behavior, i.e. this might have never been a conscious decision.

vrothberg (approved):
LGTM

haircommander:
@saschagrunert ptal

saschagrunert (approved):
/lgtm

PR #5041, opened 2021-06-28
runtimeVM: ExecSync should return a recognizable error on timeout

littlejawa:
#### What type of PR is this?

/kind bug

When running ExecSync with a timeout, do not report an error, but make an ExecSyncResponse showing that a timeout occurred.

This PR copies the logic found in runtime_oci.go:ExecSyncContainer(), and applies it to runtime_vm.go:ExecSyncContainer(), so that both have the same behaviour.

This issue was found by running the integration tests with Kata as the runtime.
The test "ctr execsync" was failing because the error reported on a timeout was not the one expected.
With this PR, the test passes.

```release-note
none
```

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -54,6 +54,11 @@ type containerInfo struct {
 	cio *cio.ContainerIO
 }
 
+const (
+	execError   = -1
+	execTimeout = -2
+)
+
 // newRuntimeVM creates a new runtimeVM instance
 func newRuntimeVM(path, root string) RuntimeImpl {
 	logrus.Debug("oci.newRuntimeVM() start")
@@ -320,7 +325,7 @@ func (r *runtimeVM) ExecSyncContainer(ctx context.Context, c *Container, command
 	}
 
 	// if the execution stopped because of the timeout, report it as such
-	if exitCode == -2 {
+	if exitCode == execTimeout {
 		return &types.ExecSyncResponse{
 			Stderr:   []byte(conmonconfig.TimedOutMessage),
 			ExitCode: -1,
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_vm.go
```diff
@@ -453,7 +462,8 @@ func (r *runtimeVM) execContainerCommon(ctx context.Context, c *Container, cmd [
 			return -1, killErr
 		}
 		<-execCh
-		return -1, errors.Errorf("ExecSyncContainer timeout (%v)", timeoutDuration)
+		// do not make an error for timeout: report it with a specific error code
+		return -2, nil
```
fidencio, line 466: Please, define an error instead of just using "-2",

fidencio (commented):
@littlejawa, the patch itself looks good, but I'd like to see a test for this (could be here, could be as part of the kata-containers' bats).

Tomorrow I'll try to exercise this code locally and finally approve (or ask for chances in) this PR.

Thanks for working on this, @littlejawa!

fidencio:
> @littlejawa, the patch itself looks good, but I'd like to see a test for this (could be here, could be as part of the kata-containers' bats).

Oh, I missed the fact this is fixing an issue found by thr ctr bats tests, sorry.
The comment about adding a new error still stands, though. :-)

internal/oci/runtime_vm.go
littlejawa: Done. I'm keeping it as a separate commit for now, just tell me if it's ok, and then I can squash if needed.

internal/oci/runtime_vm.go
```diff
@@ -318,6 +325,14 @@ func (r *runtimeVM) ExecSyncContainer(ctx context.Context, c *Container, command
 		return nil, errors.Wrap(err, "ExecSyncContainer failed")
 	}
 
+	// if the execution stopped because of the timeout, report it as such
+	if exitCode == execTimeout {
+		return &types.ExecSyncResponse{
+			Stderr:   []byte(conmonconfig.TimedOutMessage),
+			ExitCode: -1,
```
fidencio, line 332: Is this one not changed on purpose?

internal/oci/runtime_vm.go
fidencio: I think I prefer this in two separate commits, thanks @littlejawa!

internal/oci/runtime_vm.go
littlejawa: Yes. According to the [equivalent code un runtime_oci.go](https://github.com/cri-o/cri-o/blob/master/internal/oci/runtime_oci.go#L369), this is what the caller expects.

fidencio:
LGTM, thanks @littlejawa!

@haircommander, @saschagrunert, may I bother you for your input here as well?

haircommander:
I'll let @saschagrunert give the approval :)

saschagrunert (approved):
/lgtm

PR #5043, opened 2021-06-28
If capabilities are explicitly added then also add them to the ambien…

vinayakankugoyal:
…t set.

#### What type of PR is this?

/kind feature

his change makes it so that we add capabilities to the ambient set if they are explicitly added. So in kubernetes if we add a capability to the container securityContext then that capability will be added to the ambient set in addition to the inheritable, effective, permitted and bounding sets. This will allow users to run a container as non-root while giving it just the capabilities it needs, without having to apply file capabilities and allowing privilege escalation.

If ALL capabilities are added then we don't add ALL capabilities to the ambient set because that would make a non-root user essentially root. We also don't add the defaultUnixCaps i.e. the default capabilities that are always added to inheritable, effective, permitted and bounded sets to the ambient capability set as that will also make non-root user as power full as root.

```release-note
Explicitly added capabilities are added to the ambient set.
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -304,13 +304,15 @@ func setupCapabilities(specgen *generate.Generator, caps *types.Capability, defa
 	addAll := inStringSlice(caps.AddCapabilities, "ALL")
 	dropAll := inStringSlice(caps.DropCapabilities, "ALL")
 
+	var addDefaultCapbilities bool
+
 	// Only add the default capabilities to the AddCapabilities list
 	// if neither add or drop are set to "ALL". If add is set to "ALL" it
 	// is a super set of the default capabilties. If drop is set to "ALL"
 	// then we first want to clear the entire list (including defaults)
 	// so the user may selectively add *only* the capabilities they need.
 	if !(addAll || dropAll) {
-		caps.AddCapabilities = append(caps.AddCapabilities, defaultCaps...)
+		addDefaultCapbilities = true
 	}
 
 	// Add/drop all capabilities if "all" is specified, so that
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@vinayakankugoyal thanks for the PR but the CI requires PR's to be signed.  Please `git commit --amend -s`, fetch, rebase and push.

vinayakankugoyal:
@mrunalp - this is RE: kubernetes/enhancements#2757

haircommander:
/ok-to-test

full review will come later, but we can run CI for now, thanks for opening @vinayakankugoyal

server/container_create.go
```diff
@@ -351,6 +353,30 @@ func setupCapabilities(specgen *generate.Generator, caps *types.Capability, defa
 		}
 	}
 
+	if addDefaultCapbilities {
```
saschagrunert, line 356:
```suggestion
	if !addAll && !dropAll {
```

to make `addDefaultCapbilities` obsolete

mrunalp:
@vinayakankugoyal Thanks for the PR!
What about the case where a user drops ALL and adds back the default set?

vinayakankugoyal:
All those would be added to to the ambient set. But that in that case the user is choosing to do so, by explicitly adding them. We should be secure by default and not add anything to the ambient set if nothing is explicitly added is a safe starting point.

haircommander:
I think this PR doesn't make sense right now. I think we'd want an alternate way for container's to specify ambient caps, ideally with proper pod API support in kube. I am closing this. Thanks for your PR @vinayakankugoyal

PR #5046, opened 2021-06-29
Introduce CNIManager

haircommander:
#### What type of PR is this?

/kind bug
The way kubernetes deals with networking infra is that it doesn't. it's mostly up to the cri impl to figure out how pods are connected to each other. the problem with not being networking aware is that the kubelet won't wait for the network to be ready before starting pods. in resource constrained environments, this can cause large delays in node bootstrap, as the a pod that is the cni plugin is competing to start with all others.

instead, we can rate limit the pod creation by waiting until the cniplugin is ready before starting a pod. since it won't make progress until the cni plugin is up anyway, this should only improve how quickly a node comes up

```release-note
Wait for CNI plugins to be ready before starting non-host-network pods, to allow pods that may run CNI plugins to start faster
```

### DIFF
--- a/internal/config/cnimgr/cnimgr.go
+++ b/internal/config/cnimgr/cnimgr.go
@@ -34,19 +34,21 @@ func New(defaultNetwork, networkDir string, pluginDirs ...string) (*CNIManager, 
 
 func (c *CNIManager) pollUntilReady() {
 	// nolint:errcheck
-	_ = wait.PollInfinite(500*time.Millisecond, func() (bool, error) {
-		c.Lock()
-		defer c.Unlock()
-		if err := c.plugin.Status(); err != nil {
-			c.lastError = err
-			return false, nil
-		}
-		c.lastError = nil
-		for _, watcher := range c.watchers {
-			watcher <- struct{}{}
-		}
-		return true, nil
-	})
+	_ = wait.PollInfinite(500*time.Millisecond, c.pollFunc)
+}
+
+func (c *CNIManager) pollFunc() (bool, error) {
+	c.Lock()
+	defer c.Unlock()
+	if err := c.plugin.Status(); err != nil {
+		c.lastError = err
+		return false, nil
+	}
+	c.lastError = nil
+	for _, watcher := range c.watchers {
+		watcher <- struct{}{}
+	}
+	return true, nil
 }
 
 func (c *CNIManager) ReadyOrError() error {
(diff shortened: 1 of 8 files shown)


internal/config/cnimgr/cnimgr.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (c *CNIManager) pollUntilReady() {
+	backoff := wait.Backoff{
+		Duration: 500 * time.Millisecond,
+		Factor:   1.5,
+		Steps:    20,
+	}
+
+	waitErr := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
mrunalp, line 43: I would use periodic over an exponential here.

haircommander: fixed!

server/sandbox_run_linux.go
```diff
@@ -309,6 +309,16 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
 	}
 
+	securityContext := sbox.Config().Linux.SecurityContext
+	hostNetwork := securityContext.NamespaceOptions.Network == types.NamespaceModeNODE
+
+	if err := s.config.CNIPluginReadyOrError(); err != nil && !hostNetwork {
+		// if the cni plugin isn't ready yet, we should wait until it is
+		// before proceeding
+		watcher := s.config.CNIPluginAddWatcher()
+		<-watcher
```
mrunalp, line 319: After this point, we can check if we timed out before proceeding.

mrunalp (approved):
/lgtm

haircommander:
/hold

more kata failures...

server/suite_test.go
```diff
@@ -191,6 +191,10 @@ var setupSUT = func() {
 	// Inject the mock
 	sut.SetStorageImageServer(imageServerMock)
 	sut.SetStorageRuntimeServer(runtimeServerMock)
+
+	gomock.InOrder(
+		cniPluginMock.EXPECT().Status().Return(t.TestError),
+	)
```
saschagrunert, line 197: @haircommander this breaks some unit tests and will make them hang forever in 
https://github.com/cri-o/cri-o/blob/635b2d98dff9cdbbfd85d2448afa1627410659bd/server/sandbox_run_linux.go#L328

saschagrunert: Fix in https://github.com/cri-o/cri-o/pull/5099

PR #5047, opened 2021-06-30
[1.20] Update crun to v0.20.1

saschagrunert:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Update crun to the latest release to make it work with k8s.

#### Which issue(s) this PR fixes:

Refers to https://github.com/cri-o/cri-o/pull/5048

#### Special notes for your reviewer:

Per: https://kubernetes.slack.com/archives/CAZH62UR1/p1625037422136200

#### Does this PR introduce a user-facing change?

```release-note
Update crun to v0.20.1
```

PR #5048, opened 2021-06-30
[1.21] Update crun to v0.20.1

saschagrunert:
#### What type of PR is this?

/kind bug

Update crun to the latest release to make it work with k8s.

Refers to https://github.com/cri-o/cri-o/pull/5047

Per: https://kubernetes.slack.com/archives/CAZH62UR1/p1625037422136200

```release-note
Update crun to v0.20.1
```

### DIFF
--- a/scripts/versions
+++ b/scripts/versions
@@ -7,7 +7,7 @@ declare -A VERSIONS=(
     ["conmon"]=v2.0.26
     ["cri-tools"]=v1.21.0
     ["runc"]=v1.0.0-rc93
-    ["crun"]=0.20
+    ["crun"]=0.20.1
     ["bats"]=v1.3.0
 )
 export VERSIONS


fidencio:
LGTM, thanks @saschagrunert!

fidencio:
@saschagrunert, now revisiting this with way more coffee on my veins.  What happens when someone tries to run a not-so-new version of crun against 1.21 or even 1.20 with this PR merged?  Does it mean we'd require the latest crun binary? If so, what's the version present on the distros?  We should try to ensure we don't enforce a too new version of crun on systems that will not be able to have crun updated on their side.

saschagrunert:
> Does it mean we'd require the latest crun binary? If so, what's the version present on the distros? We should try to ensure we don't enforce a too new version of crun on systems that will not be able to have crun updated on their side.

I don't know if it's required in all cases that users have to use the latest version of crun. The only thing I was aware of is that in the discussion (https://kubernetes.slack.com/archives/CAZH62UR1/p1625037422136200) a problem has been fixed by updating crun to the latest release. So we probably should ship the latest release within our bundle.

fidencio:
> > Does it mean we'd require the latest crun binary? If so, what's the version present on the distros? We should try to ensure we don't enforce a too new version of crun on systems that will not be able to have crun updated on their side.
> 
> I don't know if it's required in all cases that users have to use the latest version of crun. The only thing I was aware of is that in the discussion (https://kubernetes.slack.com/archives/CAZH62UR1/p1625037422136200) a problem has been fixed by updating crun to the latest release. So we probably should ship the latest release within our bundle.

Being from the distro side, it kinda worries me a little bit that we may require the latest crun to run against 1.20, for instance.
1.20 is out already, people are already using it with their current versions of crun without issues ... then we bump it and possibly force then to move to the latest crun version because of a patch release.

I get your point, though, and I am not strongly enough against the change, as long as we're aware that we may break some setups, in case those changes break backwards compat with older versions of crun, it's more a "JFYI, it may happen".

fidencio:
so, all in all, I'm keeping the LGTM. :-)

saschagrunert:
@fidencio We still use runc as default runtime for CRI-O right? At least in kubic we have the latest version of crun available: https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/crun

fidencio:
> @fidencio We still use runc as default runtime for CRI-O right?

Yes, we do.

> At least in kubic we have the latest version of crun available: https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/crun

Perfect then!

PR #4986, opened 2021-06-08
[1.21] cgmgr: reuse dbus connection

haircommander:
#### What type of PR is this?

/kind bug
#### What this PR does / why we need it:
reuse dbus connection and attempt to reconnect if we get a dbus error
this PR heavily adapts https://github.com/opencontainers/runc/blob/7362fa2d282feffb9b19911150e01e390a23899d/libcontainer/cgroups/systemd

eventually, the plan is to upstream the internal/dbusmgr package to godbus, but we're picking this here first
#### Does this PR introduce a user-facing change?

```release-note
Reuse connection when connecting to dbus, as well as reattempt the connection if it fails temporarily
```

PR #5058, opened 2021-07-06
[1.20] cgmgr: reuse dbus connection

haircommander:
#### What type of PR is this?

/kind bug
reuse dbus connection and attempt to reconnect if we get a dbus error
this PR heavily adapts https://github.com/opencontainers/runc/blob/7362fa2d282feffb9b19911150e01e390a23899d/libcontainer/cgroups/systemd

eventually, the plan is to upstream the internal/dbusmgr package to godbus, but we're picking this here first

cherry-pick of https://github.com/cri-o/cri-o/pull/4986

```release-note
Reuse connection when connecting to dbus, as well as reattempt the connection if it fails temporarily
```


saschagrunert (approved):
LGTM

saschagrunert (approved):
/lgtm

utils/utils.go
```diff
@@ -67,9 +67,14 @@ func RunUnderSystemdScope(pid int, slice, unitName string, properties ...systemd
...
+	for {
+		_, err = conn.StartTransientUnit(unitName, "replace", properties, ch)
+		if err == nil {
+			break
+		}
+		if !errors.Is(err, syscall.EAGAIN) {
+			return err
+		}
 	}
 	defer conn.Close()
```
kolyshkin: Shouldn't this defer be moved earlier (to right after the error check from `systemdDbus.New()`)?

utils/utils.go
```diff
@@ -67,9 +67,14 @@ func RunUnderSystemdScope(pid int, slice, unitName string, properties ...systemd
...
 	ch := make(chan string)
-	_, err = conn.StartTransientUnit(unitName, "replace", properties, ch)
-	if err != nil {
-		return err
+	for {
+		_, err = conn.StartTransientUnit(unitName, "replace", properties, ch)
+		if err == nil {
+			break
+		}
+		if !errors.Is(err, syscall.EAGAIN) {
```
kolyshkin, line 80: Hmm, interesting... @haircommander  where the need for retry-on-EAGAIN comes from?

Asking because maybe we need to handle the same error in runc.

kolyshkin: OK, this comes from #4921

PR #5067, opened 2021-07-09
oci: wait until pidfile to be written before starting timer

haircommander:
#### What type of PR is this?

/kind bug
We want a way to judging how long the actual exec process is running in a container, to be totally fair in how we account for the exec. Unfortunately, there is not a great signal other than pid file creation (which still is a bit early). This improves the time the actual container process gets for exec by about 10ms

```release-note
fix a bug where a container exec process received a little less time than the timeout provided
```


utils/fsnotify.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/cri-o/cri-o/internal/log"
+	"github.com/fsnotify/fsnotify"
+)
+
+func Watch(ctx context.Context, path string, opsToWatch []fsnotify.Op) chan error {
```
saschagrunert, line 12: Do you think it makes sense to add some unit test cases for this function?

utils/fsnotify.go
```diff
@@ -0,0 +1,70 @@
...
+					if event.Op&op == op {
+						if event.Name == path {
+							done <- struct{}{}
+							return
+						}
+					}
+				}
+			case err := <-watcher.Errors:
+				if err != nil {
+					internalErrorCh <- fmt.Errorf("watch error %v", err)
```
saschagrunert, line 40: I would wrap those errors:
```suggestion
					internalErrorCh <- errors.Wrap(err, "watch failed")
```

utils/fsnotify.go
haircommander: it does!

utils/notify.go
```diff
@@ -0,0 +1,35 @@
...
+	externalErrorCh := make(chan error)
+
+	dir := filepath.Dir(path)
+	if err := notify.Watch(dir, c, opsToWatch...); err != nil {
+		externalErrorCh <- err
+		return externalErrorCh
+	}
+	go func() {
+		defer notify.Stop(c)
+		for {
```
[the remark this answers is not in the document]
haircommander: u right

haircommander:
I've opted to move to github.com/rjeczalik/notify  instead of fsnotify, partially because it seems more active, and also because I was having trouble with the unit tests and the accuracy of the cross-platform events. As a result of the new package, I have been able to shave off another ~30 ms, leaving the runc overhead at about 4ms on my machine, as opposed to 50 ms before this PR

haircommander:
I also had the idea to test what the behavior was like when conmon was running execs.

turns out, conmon was very generous to container processes--giving somewhere between 40ms and 50ms over the timeout:
`sudo crictl exec --sync --timeout 1 8d sleep 1.05`

in other words, this PR changes us from giving 100ms less from pre-conmonless-exec days to about 50ms less. BUT we are technically closer to the actual time the container process is execing

haircommander:
dropping a484419d9 for now to see if that broke stuff

go.mod
```diff
@@ -49,6 +49,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
 	github.com/psampaz/go-mod-outdated v0.8.0
+	github.com/rjeczalik/notify v0.9.2
```
saschagrunert, line 52: Do we wanna switch to this dependency completely instead of `github.com/fsnotify/fsnotify`?

internal/oci/runtime_oci.go
```diff
@@ -1099,3 +1074,24 @@ func prepareProcessExec(c *Container, cmd []string, tty bool) (processFile strin
 func (c *Container) conmonPidFilePath() string {
 	return filepath.Join(c.bundlePath, "conmon-pidfile")
 }
+
+func WatchForFile(ctx context.Context, path string, opsToWatch []notify.Event) chan error {
```
saschagrunert, line 1078: this function can be private

saschagrunert, line 1078:
```suggestion
func WatchForFile(ctx context.Context, path string, opsToWatch ...notify.Event) chan error {
```

saschagrunert (commented):
Generally LGTM, just a few nits

go.mod
haircommander: I found there were test failures with it and I want to jam this through first.

internal/oci/runtime_oci.go
haircommander: not if we want to unit test it, right?

internal/oci/runtime_oci.go
```diff
@@ -1099,3 +1055,26 @@ func prepareProcessExec(c *Container, cmd []string, tty bool) (processFile strin
...
+
+	dir := filepath.Dir(path)
+	if err := notify.Watch(dir, c, opsToWatch...); err != nil {
+		externalErrorCh <- err
+		return externalErrorCh
+	}
+	go func() {
+		defer notify.Stop(c)
+		for ei := range c {
+			fmt.Println(ei)
```
haircommander: todo gotta drop this

internal/oci/runtime_oci.go
haircommander: todo ctx is unused

saschagrunert: Uh, yeah right.

saschagrunert (approved):
/lgtm

haircommander:
```
time="2021-07-16T10:18:11Z" level=error msg="Error copying docker://quay.io/crio/oom:latest to dir:/go/src/github.com/cri-o/cri-o/.artifacts/oom-image: Error reading blob sha256:131fcea386e736dfaf075424d7e2a31ae8a7fcbb9a87ba93c8a70edd120b
Error pulling docker://quay.io/crio/oom:latest
```

quay flake
/retest

haircommander:
I dropped the bats test, because it was flaky, and I added some code to make sure we're not leaking channels or routines.

Issue #5077, opened 2021-07-15
regression: sandbox annotations are not propagated to container annotations


frasertweedale:
**Description**

Commit https://github.com/cri-o/cri-o/commit/8cf32223a1dc1208b53167763638886fd23bde67#diff-208130c20edcdc687fd2a4c092c9c22ad3f198f3fe2d27ea50c74a2656c4764d introduces some extra processing of seccomp-related sandbox annotations.  As a result of this change, most sandbox annotations are no longer propagated to the container annotations.

**Steps to reproduce the issue:**
1. create pod with k8s annotations
2. observe that the sandbox OCI config contains the k8s annotations (as expected)
3. observe that the container(s) OCI config(s) do not contain the k8s annotations.  They were not propagated from the sandbox config to the container config.

**Describe the results you expected:**

Sandbox annotations should also appear in the container's OCI spec.

**Additional information you deem important (e.g. issue happens only occasionally):**

Regression is in 1.21 (OpenShift 4.8).  So, the fix should be backported.

**Output of `crio --version`:**

```shell
sh-4.4# rpm -q cri-o
cri-o-1.21.1-12.rhaos4.8.git30ca719.el8.x86_64
sh-4.4# crio --version
crio version 1.21.1-12.rhaos4.8.git30ca719.el8
Version:    1.21.1-12.rhaos4.8.git30ca719.el8
GoVersion:  go1.16.4
Compiler:   gc
Platform:   linux/amd64
Linkmode:   dynamic
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**

OpenShift 4.8.0.

frasertweedale:
For good measure, I checked the behaviour of CRI-O 1.20 on an OpenShift 4.7.20 cluster.  And this is indeed a regression.

PR #5078, opened 2021-07-15
Fix propagation of annotations from sandbox to container

frasertweedale:
Commit 8cf32223a1dc1208b53167763638886fd23bde67 introduces some
extra processing of seccomp-related sandbox annotations.  But it
introduced a regression where most sandbox annotations are no longer
propagated to the container annotations.  Add an `else` clause to
ensure we propagate all the sandbox annotations to the container's
OCI configuration.

#### What type of PR is this?

/kind bug

Fixes: https://github.com/cri-o/cri-o/issues/5077

```release-note
None
```

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -157,7 +157,7 @@ func (c *container) SpecAddAnnotations(ctx context.Context, sb *sandbox.Sandbox,
 	// The sandbox annotations are already filtered for the allowed
 	// annotations, there is no need to check it additionally here.
 	for k, v := range sb.Annotations() {
-		if strings.HasPrefix(k, crioann.OCISeccompBPFHookAnnotation) {
+		if k == crioann.OCISeccompBPFHookAnnotation+"/"+c.config.Metadata.Name {
 			// The OCI seccomp BPF hook
 			// (https://github.com/containers/oci-seccomp-bpf-hook)
 			// uses the annotation io.containers.trace-syscall as indicator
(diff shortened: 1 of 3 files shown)


haircommander:
/ok-to-test

thanks @frasertweedale , since you also want it in 4.8
/cherry-pick release-1.21

pkg/container/container.go
```diff
@@ -182,6 +182,8 @@ func (c *container) SpecAddAnnotations(ctx context.Context, sb *sandbox.Sandbox,
 				// Annotation not suffixed with the container name
 				c.spec.AddAnnotation(k, v)
 			}
+		} else {
+			c.spec.AddAnnotation(k, v)
```
saschagrunert, line 186: Let's simplify the code by `continue`'ing after line 180 and removing the `} else {`s to just `c.spec.AddAnnotation(k, v)` in all other cases.

frasertweedale: Respectfully, I think such a change makes the code less readable, i.e. harder for a human to understand what's happening.  I think balanced if/else clauses are the way to go.

mrunalp:
@saschagrunert @haircommander We need tests here to make sure we don't regress again in the future.

pkg/container/container.go
frasertweedale: @saschagrunert alternatively I could simplify the whole thing to something like:
```go
for k, v := range sb.Annotations() {
  if k == crioann.OCISeccompBPFHookAnnotation + "/" + c.config.Metadata.Name {
    log.Debugf(ctx,
	       "Annotation key for container %q rewritten to %q (value is: %q)",
	       c.config.Metadata.Name, crioann.OCISeccompBPFHookAnnotation, v,
    )
    c.config.Annotations[crioann.OCISeccompBPFHookAnnotation] = v
    c.spec.AddAnnotation(crioann.OCISeccompBPFHookAnnotation, v)
  } else {
    c.spec.AddAnnotation(k, v)
  }
}
```

(Preserving the existing comments of course).  What do you think?

saschagrunert: Yeah sound good to me, thank you for the alternative appraoch.

frasertweedale:
> @saschagrunert @haircommander We need tests here to make sure we don't regress again in the future.

I am unfamiliar with the test suite.  If someone would identify a test that would make a good "template" for testing this behaviour, I'll be happy to have a go.  Otherwise it might be better use of everyone's time overall for someone more familiar with cri-o test suite to contribute a test.

saschagrunert:
> > @saschagrunert @haircommander We need tests here to make sure we don't regress again in the future.
> 
> Otherwise it might be better use of everyone's time overall for someone more familiar with cri-o test suite to contribute a test.

Sure thing, we can follow-up on that in a separate PR :+1:

TomSweeneyRedHat (commented):
LGTM
assuming happy tests

haircommander:
/approve
@saschagrunert @mrunalp I'll give y'all the final lgtm

haircommander:
ah @frasertweedale I thinki you need to run `gofmt -s -w pkg/container/container.go`

frasertweedale:
I fixed the lint nit and also added the regression test.

frasertweedale:
CI seems to be stalled somehow...

haircommander:
yeah we're unfortunately dealing with a cert rotation issue right now, thanks for the patience

saschagrunert (approved):
/lgtm

PR #5082, opened 2021-07-15
Monitor defunct processes in a node

raisaat:
#### What type of PR is this?

/kind feature

This patch adds `internal/process/defunct_processes.go` that retrieves the total number of defunct/zombie processes in a node and adds a new metric `crio_processes_defunct` that uses `internal/process/defunct_processes.go` to collect the total number of defunct processes.

None

None

```release-note
Added `internal/process/defunct_processes.go` and `crio_processes_defunct` metric to collect the total number of defunct/zombie processes in a node.
```

### DIFF
--- a/internal/process/defunct_processes.go
+++ b/internal/process/defunct_processes.go
@@ -1,0 +1,87 @@
+package process
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// ProcessFS is the process file system.
+const ProcessFS = "/proc"
+
+// Stat represents status information of a process from /proc/[pid]/stat.
+type Stat struct {
+	// Comm is the command name (usually the executable filename).
+	Comm string
+
+	// State is the state of the process.
+	State string
+}
+
+// DefunctProcesses returns the number of zombie processes in the node.
+func DefunctProcesses() (defunctCount uint, retErr error) {
+	directories, err := os.Open(ProcessFS)
+	if err != nil {
+		return 0, err
+	}
+	defer directories.Close()
+
+	names, err := directories.Readdirnames(-1)
+	if err != nil {
+		return 0, err
+	}
+
+	for _, name := range names {
+		// Processes have numeric names. If the name cannot
+		// be parsed to an int, it is not a process name.
+		if _, err := strconv.ParseInt(name, 10, 0); err != nil {
+			continue
+		}
+
+		stat, err := processStats(name)
+		if err != nil {
+			logrus.Debugf("Failed to get the status of process with PID %s: %v", name, err)
+			continue
+		}
+		if stat.State == "Z" {
+			logrus.Warnf("Found defunct process with PID %s (%s)", name, stat.Comm)
+			defunctCount++
+		}
+	}
+	return defunctCount, nil
+}
+
+// processStats returns status information of a process as defined in /proc/[pid]/stat
+func processStats(pid string) (*Stat, error) {
+	bytes, err := ioutil.ReadFile(filepath.Join(ProcessFS, pid, "stat"))
+	if err != nil {
+		return nil, err
+	}
+	data := string(bytes)
+
+	// /proc/[PID]/stat format is described in proc(5). The second field is process name,
+	// enclosed in parentheses, and it can contain parentheses inside. No other fields
+	// can have parentheses, so look for the last ')'.
+	i := strings.LastIndexByte(data, ')')
+	if i <= 2 || i >= len(data)-1 {
+		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
+	}
+
+	parts := strings.SplitN(data[:i], " (", 2)
+	if len(parts) != 2 {
+		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
+	}
+
+	return &Stat{
+		// The command name is field 2.
+		Comm: parts[1],
+
+		// The state is field 3, which is the first two fields and a space after.
+		State: string(data[i+2]),
+	}, nil
+}
--- a/server/metrics/collectors/collectors.go
+++ b/server/metrics/collectors/collectors.go
@@ -54,6 +54,9 @@ const (
 
 	// ContainersOOM is the key for the CRI-O container out of memory metrics per container name.
 	ContainersOOM Collector = crioPrefix + "containers_oom"
+
+	// ProcessesDefunct is the key for the total number of defunct processes in a node.
+	ProcessesDefunct Collector = crioPrefix + "processes_defunct"
 )
 
 // FromSlice converts a string slice to a Collectors type.
@@ -89,6 +92,7 @@ func All() Collectors {
 		ImageLayerReuse.Stripped(),
 		ContainersOOMTotal.Stripped(),
 		ContainersOOM.Stripped(),
+		ProcessesDefunct.Stripped(),
 	}
 }
 
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cri-o/cri-o/internal/process"
 	libconfig "github.com/cri-o/cri-o/pkg/config"
 	"github.com/cri-o/cri-o/server/metrics/collectors"
 	"github.com/fsnotify/fsnotify"
@@ -43,6 +44,7 @@ type Metrics struct {
 	metricImageLayerReuse         *prometheus.CounterVec
 	metricContainersOOMTotal      prometheus.Counter
 	metricContainersOOM           *prometheus.CounterVec
+	metricProcessesDefunct        prometheus.GaugeFunc
 }
 
 var instance *Metrics
@@ -165,6 +167,21 @@ func New(config *libconfig.MetricsConfig) *Metrics {
 				Help:      "Amount of containers killed because they ran out of memory (OOM) by their name",
 			},
 			[]string{"name"},
+		),
+		metricProcessesDefunct: prometheus.NewGaugeFunc(
+			prometheus.GaugeOpts{
+				Subsystem: collectors.Subsystem,
+				Name:      collectors.ProcessesDefunct.String(),
+				Help:      "Total number of defunct processes in the node",
+			},
+			func() float64 {
+				total, err := process.DefunctProcesses()
+				if err == nil {
+					return float64(total)
+				}
+				logrus.Warn(err)
+				return 0
+			},
 		),
 	}
 	return Instance()
@@ -335,6 +352,7 @@ func (m *Metrics) createEndpoint() (*http.ServeMux, error) {
 		collectors.ImageLayerReuse:         m.metricImageLayerReuse,
 		collectors.ContainersOOMTotal:      m.metricContainersOOMTotal,
 		collectors.ContainersOOM:           m.metricContainersOOM,
+		collectors.ProcessesDefunct:        m.metricProcessesDefunct,
 	} {
 		if m.config.MetricsCollectors.Contains(collector) {
 			logrus.Debugf("Enabling metric: %s", collector.Stripped())
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -261,7 +261,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--metrics-cert**="": Certificate for the secure metrics endpoint
 
-**--metrics-collectors**="": Enabled metrics collectors (default: [operations operations_latency_microseconds_total operations_latency_microseconds operations_errors image_pulls_by_digest image_pulls_by_name image_pulls_by_name_skipped image_pulls_failures image_pulls_successes image_pulls_layer_size image_layer_reuse containers_oom_total containers_oom])
+**--metrics-collectors**="": Enabled metrics collectors (default: [operations operations_latency_microseconds_total operations_latency_microseconds operations_errors image_pulls_by_digest image_pulls_by_name image_pulls_by_name_skipped image_pulls_failures image_pulls_successes image_pulls_layer_size image_layer_reuse containers_oom_total containers_oom processes_defunct])
 
 **--metrics-key**="": Certificate key for the secure metrics endpoint
 
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -341,7 +341,7 @@ The `crio.metrics` table containers settings pertaining to the Prometheus based 
 **enable_custom_shm_size**=false
 If set to true, enable users to set a custom shm size instead of using the default value of 64M. The shm size can be set through K8S annotation with the key "io.kubernetes.cri-o.ShmSize", and the value representing the size in human readable format. For example: "io.kubernetes.cri-o.ShmSize: 128Mi"
 
-**metrics_collectors**=["operations", "operations_latency_microseconds_total", "operations_latency_microseconds", "operations_errors", "image_pulls_by_digest", "image_pulls_by_name", "image_pulls_by_name_skipped", "image_pulls_failures", "image_pulls_successes", "image_pulls_layer_size", "image_layer_reuse", "containers_oom_total", "containers_oom"]
+**metrics_collectors**=["operations", "operations_latency_microseconds_total", "operations_latency_microseconds", "operations_errors", "image_pulls_by_digest", "image_pulls_by_name", "image_pulls_by_name_skipped", "image_pulls_failures", "image_pulls_successes", "image_pulls_layer_size", "image_layer_reuse", "containers_oom_total", "containers_oom", "processes_defunct"]
   Enabled metrics collectors
 
 **metrics_port**=9090
--- a/server/metrics/collectors/collectors_test.go
+++ b/server/metrics/collectors/collectors_test.go
@@ -49,11 +49,12 @@ var _ = t.Describe("Collectors", func() {
 				collectors.ImageLayerReuse,
 				collectors.ContainersOOMTotal,
 				collectors.ContainersOOM,
+				collectors.ProcessesDefunct,
 			} {
 				Expect(all.Contains(collector)).To(BeTrue())
 			}
 
-			Expect(all).To(HaveLen(13))
+			Expect(all).To(HaveLen(14))
 		})
 	})
 
--- a/tutorials/metrics.md
+++ b/tutorials/metrics.md
@@ -54,6 +54,7 @@ Beside the [default golang based metrics][2], CRI-O provides the following addit
 | `crio_image_layer_reuse`                     | `name`                                                                                                                                  | Counter   | Reused (not pulled) local image layer count by name.                                                 |
 | `crio_containers_oom_total`                  |                                                                                                                                         | Counter   | Total number of containers killed because they ran out of memory (OOM)                               |
 | `crio_containers_oom`                        | `name`                                                                                                                                  | Counter   | Containers killed because they ran out of memory (OOM) by their name                                 |
+| `crio_processes_defunct`                     |                                                                                                                                         | Gauge     | Total number of defunct processes in the node                                                        |
 
 - Available CRI-O RPC's from the [gRPC API][3]: `Attach`, `ContainerStats`, `ContainerStatus`,
   `CreateContainer`, `Exec`, `ExecSync`, `ImageFsInfo`, `ImageStatus`,


kolyshkin:
I'm afraid this is not a very efficient way to collect all zombies. If you look at what `process.Processes()` and `p.Status()` do, it's a lot of work, some of which is unnecessary, and it scales badly on a loaded node with lots of processes. It's OK for the proof of concept, but not good/scalable enough for production systems.

`process.Processes` starts nicely by reading all the entries from `/proc`, trying to convert them to intergers and adding those which can be converted. So far so good.

Next, it sends a `0` signal to every process in the system to make sure it exists. This is way too heavy and can be avoided.

`p.Status()`, called for every process, carefully parses the whole nine yards of `/proc/$PID/status` for the sole reason of checking a single `status` field, and leaving everything else for go garbage collector to take care of. This, again, is way too heavy.

I suggest writing own specialized parses, which reads every `/proc/*/stat` (which is faster than `status`) and parsing it. The only caveat is there can be `)` in process name, so you need to look for the last `) ` and then take the next character. For the example of such parser, see
https://github.com/kolyshkin/runc/blob/0eda452ee05f0e5d60128da337896fcfeb8823ca/libcontainer/system/proc.go#L75-L84
https://github.com/kolyshkin/runc/blob/0eda452ee05f0e5d60128da337896fcfeb8823ca/libcontainer/system/proc.go#L98-L100

saschagrunert:
I agree with @kolyshkin points so we can avoid any new dependency at all while staying performant.

Things which needs to be done here as well:
- Adding the new metric to [`server/metrics/metrics.go`](https://github.com/cri-o/cri-o/blob/master/server/metrics/metrics.go)
- Defining which labels we wanna use for the metric
- Updating the docs: https://github.com/cri-o/cri-o/blob/master/tutorials/metrics.md

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,25 @@
...
+	for _, p := range processes {
+		var status string
+		status, err = p.Status()
+
+		if err == nil && status == "Z" {
+			defunctProcesses = append(defunctProcesses, p)
+		}
+	}
+	return defunctProcesses, nil
+}
```
raisaat, line 25: Thank you @kolyshkin and @saschagrunert for your advice. I wiped out everything I had done in `internal/process/defunct_processes.go` earlier and used my own implementation where I am only returning the PIDs of all the zombie processes. Would greatly appreciate your input before I make another commit :)

```suggestion
package process

import (
	"fmt"
	"os"
	"strconv"
	"strings"
)

// DefunctProcesses returns the PIDs of all the zombie processes in the node
func DefunctProcesses() ([]int, error) {
	directories, err := os.Open("/proc")
	if err != nil {
		return nil, err
	}
	defer directories.Close()

	names, err := directories.Readdirnames(-1)
	if err != nil {
		return nil, err
	}

	var pids []int
	for _, name := range names {
		// processes have numeric names. If the name does not start with a number, skip.
		if name[0] < '0' || name[0] > '9' {
			continue
		}

		pid, err := strconv.ParseInt(name, 10, 0)
		if err != nil {
			continue
		}

		stat, err := status(int(pid))
		if err != nil {
			continue
		}
		if stat == "Z" {
			pids = append(pids, int(pid))
		}
	}
	return pids, nil
}

// status returns the status of a process
func status(pid int) (string, error) {
	bytes, err := os.ReadFile("/proc" + strconv.Itoa(pid) + "stat")
	if err != nil {
		return "", err
	}
	data := string(bytes)

	i := strings.LastIndexByte(data, ')')
	if i <= 2 || i >= len(data)-1 {
		return "", fmt.Errorf("invalid stat data (no comm): %q", data)
	}
	return string(data[i+2]), nil
}
```

saschagrunert: @raisaat please commit your suggestion, this way we can better comment line-by-line. :)

raisaat: @saschagrunert made a new commit :)

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,59 @@
...
+		if name[0] < '0' || name[0] > '9' {
+			continue
+		}
+
+		pid, err := strconv.ParseInt(name, 10, 0)
+		if err != nil {
+			continue
+		}
+
+		stat, err := status(int(pid))
```
saschagrunert, line 35: Pass the name string here to avoid the `strconv.Itoa(pid)` in `status()`.

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// DefunctProcesses returns the PIDs of all the zombie processes in the node
+func DefunctProcesses() ([]int, error) {
+	directories, err := os.Open("/proc")
```
saschagrunert, line 12: Let's make `/proc` a const of the package to re-use in several places.

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// DefunctProcesses returns the PIDs of all the zombie processes in the node
+func DefunctProcesses() ([]int, error) {
```
saschagrunert, line 11: I think we should keep the name of the process as well as the pid, so a wrapper struct could help here.
```suggestion
func DefunctProcesses() ([]Process, error) {
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,59 @@
...
+	directories, err := os.Open("/proc")
+	if err != nil {
+		return nil, err
```
saschagrunert, line 14: Let's wrap errors to add more context:
```suggestion
		return nil, errors.Wrap(err, "read proc file system")
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		pid, err := strconv.ParseInt(name, 10, 0)
+		if err != nil {
+			continue
+		}
+
+		stat, err := status(int(pid))
+		if err != nil {
+			continue
```
saschagrunert, line 37: I'd say we should `logrus.Warnf` in that case.

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,59 @@
...
+func status(pid int) (string, error) {
+	bytes, err := os.ReadFile("/proc/" + strconv.Itoa(pid) + "/stat")
+	if err != nil {
+		return "", err
+	}
+	data := string(bytes)
+
+	i := strings.LastIndexByte(data, ')')
+	if i <= 2 || i >= len(data)-1 {
+		return "", fmt.Errorf("invalid stat data (no comm): %q", data)
```
saschagrunert, line 56:
```suggestion
		return "", errors.Errorf("invalid stat data (no comm): %q", data)
```

saschagrunert (commented):
Thank you @raisaat, I gave it a first review. :pray:

internal/process/defunct_processes.go
saschagrunert: As discussed, we probably only want to implement a simple counter metric, like we do for OOM. This means we do not need the PID and name of the process at all for the metric and can only return the amount here.

It would be still helpful for debugging to have the pid as well as the process name in the logs.

internal/process/defunct_processes.go
kolyshkin: 1. `os.Open` provides enough context (as it returns `&os.PathError` with the file name, the op and the error) so I don't think it makes sense to wrap this.

2. Strictly speaking, we need to get rid of pkg/errors and use standard go 1.13+ error wrapping, i.e. `fmt.Errorf` with `%w`. But that's surely out of scope for this PR.

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	return string(data[i+2]), nil
```
kolyshkin, line 58: This calls for a comment explaining what you do here.

internal/process/defunct_processes.go
```diff
@@ -1,59 +1,90 @@
...
+const ProcessFS = "/proc"
+
+// Stat represents status information of a process from /proc/[pid]/stat.
+type Stat struct {
+	// Fname is the filename of the executable.
+	Fname string
+
+	// State is the state of the process.
+	State string
+}
```
raisaat, line 21: This struct allows `status()` to return the executable name of the process (along with its state) which is needed for logging as mentioned in https://github.com/cri-o/cri-o/pull/5082#discussion_r672302364.

internal/process/defunct_processes.go
```diff
@@ -1,59 +1,90 @@
...
+func status(pid string) (*Stat, error) {
+	bytes, err := os.ReadFile(ProcessFS + "/" + pid + "/stat")
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	data := string(bytes)
 
+	// Get the last index of field 2 in /proc/[pid]/stat.
+	// The last index of field 2 must contain a ')'.
```
kolyshkin, line 71: I would say something like

```go
// /proc/[PID]/stat format is described in proc(5). The second field is process name,
// enclosed in parentheses, and it can contain parentheses inside. No other fields
// can have parentheses, so we are looking for the last one.
```

internal/process/defunct_processes.go
```diff
@@ -1,59 +1,90 @@
...
+	}
+
+	parts := strings.SplitN(data[:i], " (", 2)
+	if len(parts) != 2 {
+		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
+	}
+
+	stat := &Stat{
+		// The executable filename is field 2.
+		Fname: parts[1],
```
kolyshkin, line 84: I'd call it `Comm` or `Name`, and describe as "command name (usually the executable filename)".

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,90 @@
...
+type Stat struct {
+	// Fname is the filename of the executable.
+	Fname string
+
+	// State is the state of the process.
+	State string
+}
+
+// DefunctProcesses returns the number of zombie processes in the node.
+func DefunctProcesses() (int, error) {
```
saschagrunert, line 24: nit: you can pre-declare `defunctCount` here and let's make it an uint:
```suggestion
func DefunctProcesses() (defunctCount uint, err error) {
```

internal/process/defunct_processes.go
```diff
@@ -1,59 +1,90 @@
...
+	directories, err := os.Open(ProcessFS)
 	if err != nil {
-		return nil, err
+		return -1, err
```
raisaat, line 27: @saschagrunert if I make `defunctCount` an `uint` as suggested in https://github.com/cri-o/cri-o/pull/5082#pullrequestreview-710249028, I cannot return -1 here. Should I return a 0? But that would create a confusion between whether there are 0 defunct processes or whether there is an error, right?

saschagrunert: If there is an `err != nil` returned, then we will not use the other result (`defunctCount`) in any case. We can only guarantee that the returned result is correct if `err == nil`, which means we can safely `return 0, err` here :)

internal/process/defunct_processes.go
```diff
@@ -80,11 +80,11 @@ func status(pid string) (*Stat, error) {
...
-		Fname: parts[1],
+		// The command name is field 2.
+		Comm: parts[1],
 
 		// The state is field 3, which is the first two fields and a space after.
 		State: string(data[i+2]),
 	}
 	return stat, nil
-}
+}
```
kolyshkin, line 90: nit: no newline at EOF

raisaat: Thank you for the catch! Added a newline :)

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,90 @@
...
+			logrus.Warnf("Found defunct process with PID %s (%s)", name, stat.Comm)
+			defunctCount++
+		}
+	}
+	return defunctCount, nil
+}
+
+// status returns status information of a process as defined in /proc/[pid]/stat
+func status(pid string) (*Stat, error) {
+	bytes, err := os.ReadFile(ProcessFS + "/" + pid + "/stat")
```
saschagrunert, line 63:
```suggestion
	bytes, err := os.ReadFile(filepath.Join(ProcessFS, pid, "stat"))
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	stat := &Stat{
+		// The command name is field 2.
+		Comm: parts[1],
+
+		// The state is field 3, which is the first two fields and a space after.
+		State: string(data[i+2]),
+	}
+	return stat, nil
```
saschagrunert, line 89: nit:
```suggestion
	return &Stat{
		// The command name is field 2.
		Comm: parts[1],

		// The state is field 3, which is the first two fields and a space after.
		State: string(data[i+2]),
	}, nil
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,90 @@
+package process
+
+import (
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"os"
+	"strconv"
+	"strings"
+)
```
saschagrunert, line 9:
```suggestion
import (
	"os"
	"path/filepath"
	"strconv"
	"strings"

	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
)
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,90 @@
...
+type Stat struct {
+	// Comm is the command name (usually the executable filename).
+	Comm string
+
+	// State is the state of the process.
+	State string
+}
+
+// DefunctProcesses returns the number of zombie processes in the node.
+func DefunctProcesses() (defunctCount uint, err error) {
```
haircommander, line 24: nit: can we name this err retErr or something? we have hit weird scoping issues in the past when we use the same name declared in the return values as we use throughout the function

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,90 @@
...
+	names, err := directories.Readdirnames(-1)
+	if err != nil {
+		return 0, err
+	}
+
+	for _, name := range names {
+		// Processes have numeric names. If the name does not start with a number, skip.
+		if name[0] < '0' || name[0] > '9' {
+			continue
+		}
```
haircommander, line 40: this reads as duplicate of ParseInt. is there any reason to not just call ParseInt here?

haircommander:
this is coming along well @raisaat :smiley:

internal/process/defunct_processes.go
raisaat: @haircommander Essentially yeah, `ParsInt()` would take care of this case, but I was trying to filter out the directory names that don't start with a number and hence, avoid calling `ParseInt()` on every directory name to reduce the number of function calls. Could get rid of it if the overhead of too many function calls is not negligible :)

internal/process/defunct_processes.go
raisaat: Sure, I'll do so.

internal/process/defunct_processes.go
raisaat: @saschagrunert what do you think?

server/metrics/metrics.go
```diff
@@ -208,6 +213,20 @@ func New(config *libconfig.MetricsConfig) *Metrics {
...
 			[]string{"name"},
 		),
+		metricDefunctProcessesTotal: prometheus.NewGaugeFunc(
+			prometheus.GaugeOpts{
+				Subsystem: subsystem,
+				Name:      CRIODefunctProcessesTotalKey,
+				Help:      "Total number of defunct processes in the node",
+			},
+			func() float64 {
+				total, err := process.DefunctProcesses()
```
raisaat, line 223: Not sure how to handle the error from `DefunctProcesses()` if it is not `nil`

server/metrics/metrics.go
```diff
@@ -208,6 +213,20 @@ func New(config *libconfig.MetricsConfig) *Metrics {
...
+				if err == nil {
+					return float64(total)
+				}
+				return 0
```
raisaat, line 227: Not sure how to handle the error from `DefunctProcesses()` if it is not `nil`.

kolyshkin:
1. It seems that the first 4 commits need to be squashed.
2. This needs a rebase.

server/metrics/metrics.go
saschagrunert: This looks generally good. We could log the error in case we cannot retrieve them.

server/metrics/metrics.go
```diff
@@ -61,6 +62,9 @@ const (
 	// CRIOContainersOOMKey is the key for the CRI-O container out of memory metrics per container name.
 	CRIOContainersOOMKey = "crio_containers_oom"
 
+	// CRIODefunctProcessesTotalKey is the key for the total number of defunct processes in a node.
+	CRIODefunctProcessesTotalKey = "crio_defunct_processes_total"
```
saschagrunert, line 66:
```suggestion
	CRIOProcessesDefunctTotalKey = "crio_processes_defunct_total"
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,91 @@
...
+
+		// If the name cannot be parsed to an int, it is not numeric.
+		_, err := strconv.ParseInt(name, 10, 0)
+		if err != nil {
+			continue
+		}
+
+		stat, err := status(name)
+		if err != nil {
+			logrus.Warnf("Failed to get the status of process with PID %s: %v", name, err)
```
saschagrunert, line 52: Let's move the verbosity down to debug
```suggestion
			logrus.Debugf("Failed to get the status of process with PID %s: %v", name, err)
```

saschagrunert (commented):
We're almost there. Please rebase and andraft that the CI can run.
/ok-to-test

server/metrics/metrics.go
simonpasquier: The `_total` suffix should only be used for counters, summaries and histograms (see https://prometheus.io/docs/instrumenting/writing_exporters/#naming).

```suggestion
	CRIODefunctProcessesTotalKey = "crio_defunct_processes"
```

server/metrics/metrics.go
simonpasquier: You could also implement a custom collector to cache the last know value + expose another metric `crio_processes_defunct_timestamp` reporting the "freshness" of the metric.

simonpasquier: but the way it's implemented right now might be good enough :)

internal/process/defunct_processes.go
saschagrunert: I assume it wouldn’t be faster with this check, so yes let’s just use ParseInt

server/metrics/metrics.go
saschagrunert: Good catch!

internal/process/defunct_processes.go
```diff
@@ -13,27 +13,26 @@ const ProcessFS = "/proc"
...
 	State string
 }
 
 // DefunctProcesses returns the number of zombie processes in the node.
-func DefunctProcesses() (int, error) {
+func DefunctProcesses() (defunctCount uint, err error) {
 	directories, err := os.Open(ProcessFS)
 	if err != nil {
-		return -1, err
+		return 0, err
```
raisaat, line 29:
```suggestion
logrus.Errorf("%v", err)
return 0, err
```

server/metrics/collectors/collectors.go
```diff
@@ -54,6 +54,9 @@ const (
 
 	// ContainersOOM is the key for the CRI-O container out of memory metrics per container name.
 	ContainersOOM Collector = crioPrefix + "containers_oom"
+
+	// ProcessesDefunctTotal is the key for the total number of defunct processes in a node.
+	ProcessesDefunctTotal Collector = crioPrefix + "processes_defunct"
```
saschagrunert, line 59:
```suggestion
	ProcessesDefunct Collector = crioPrefix + "processes_defunct"
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		logrus.Warn(err)
+		return 0, err
+	}
+
+	for _, name := range names {
+		// Processes have numeric names. If the name cannot
+		// be parsed to an int, it is not a process name.
+		_, err := strconv.ParseInt(name, 10, 0)
+		if err != nil {
```
saschagrunert, line 44: nit: line can be combined with previous one

saschagrunert (approved):
Just two nits, otherwise LGTM

PTAL @haircommander @kolyshkin

saschagrunert:
@raisaat please update the metrics docs, too

rphillips:
@saschagrunert there is a build failure being fixed as well

```
internal/process/defunct_processes.go:63:16: undefined: os.ReadFile
```

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,89 @@
...
+		if stat.State == "Z" {
+			logrus.Warnf("Found defunct process with PID %s (%s)", name, stat.Comm)
+			defunctCount++
+		}
+	}
+	return defunctCount, nil
+}
+
+// status returns status information of a process as defined in /proc/[pid]/stat
+func status(pid string) (*Stat, error) {
```
haircommander, line 62: nit: can we have a more informative name here? maybe statsForPid?

haircommander: or processStats

server/metrics/metrics.go
```diff
@@ -166,6 +168,20 @@ func New(config *libconfig.MetricsConfig) *Metrics {
...
+			prometheus.GaugeOpts{
+				Subsystem: collectors.Subsystem,
+				Name:      collectors.ProcessesDefunctTotal.String(),
+				Help:      "Total number of defunct processes in the node",
+			},
+			func() float64 {
+				total, err := process.DefunctProcesses()
+				if err == nil {
+					return float64(total)
+				}
```
haircommander, line 181: it doesn't look like we logged the error here

haircommander:
it may behoove us to have an integration test--one where we enable the metric, create a zombie process and verify the metric returns the correct value

internal/process/defunct_processes.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// State is the state of the process.
+	State string
+}
+
+// DefunctProcesses returns the number of zombie processes in the node.
+func DefunctProcesses() (defunctCount uint, retErr error) {
+	directories, err := os.Open(ProcessFS)
+	if err != nil {
+		logrus.Warn(err)
```
rphillips, line 29: nit: Adding descriptive errors with logrus.Warnf("opening proc filesystem failed: %v", err)... Probably should update the other logrus.Warns in the file as well.

server/metrics/metrics.go
raisaat: @haircommander Should a `logrus.Warnf("%v", err)` before `return 0` in line 182 suffice?

internal/process/defunct_processes.go
raisaat: @saschagrunert Like this: `_, err := strconv.ParseInt(name, 10, 0); if err != nil { continue }`

saschagrunert: Yes, with the `if` at the beginning of the line.

server/metrics/metrics.go
haircommander: yup that would be good!

internal/process/defunct_processes.go
raisaat: Thanks for the suggestion @rphillips! But I think `os.Open` returns enough context as @kolyshkin suggested in https://github.com/cri-o/cri-o/pull/5082#discussion_r672636392 so we may not need it :)

server/metrics/metrics.go
```diff
@@ -166,6 +168,21 @@ func New(config *libconfig.MetricsConfig) *Metrics {
...
+				Subsystem: collectors.Subsystem,
+				Name:      collectors.ProcessesDefunct.String(),
+				Help:      "Total number of defunct processes in the node",
+			},
+			func() float64 {
+				total, err := process.DefunctProcesses()
+				if err == nil {
+					return float64(total)
+				}
+				logrus.Warnf("%v", err)
```
raisaat, line 182: @saschagrunert since I am logging the error here, I'm not logging the error before returning from `DefunctProcesses()` in lines https://github.com/raisaat/cri-o/blob/monitor-defunct/internal/process/defunct_processes.go#L30 and https://github.com/raisaat/cri-o/blob/monitor-defunct/internal/process/defunct_processes.go#L36. Otherwise, I believe there would be redundant logs.

saschagrunert: Looks good to me, maybe just
```suggestion
				logrus.Warn(err)
```

saschagrunert (approved):
Just one nit, otherwise LGTM

haircommander:
I think you need to update a unit test: https://github.com/cri-o/cri-o/pull/5082/checks?check_run_id=3158926906

kolyshkin (approved):
LGTM

saschagrunert (approved):
/lgtm

saschagrunert:
@raisaat do you wanna follow-up with writing unit and integration tests? :)

PR #1386, opened 2018-02-28
execsync: Set terminal to true when we pass -t to conmon

mrunalp:
We may consider setting it to true all the time but this
should match our previous behavior before we started
using process json for exec.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

PR #310, opened 2017-01-11
Fix ExecSync support for runtimes other than runC

sameo:
Some OCI container runtimes (in particular the hypervisor based ones like e.g. Clear Containers) will typically create a shim process between the hypervisor and the runtime caller, in order to not rely on the hypervisor process for e.g. forwarding the output streams or getting a command exit code.

When executing a command inside a running container those runtimes will create that shim process and terminate.
Therefore calling and monitoring the runtime process directly from ExecSync() will fail. Instead we need to have a subreaper calling the runtime and monitoring the shim process, and conmon seems to be a natural place for doing so.

This PR does mostly 2 things:

1) Add a `-e` option to `conmon` to also monitor potential shim processes created by `<runtime> exec`
2) Modify `ExecSync` to call `conmon` instead of calling the runtime directly. `ExecSync` also now reads the exit status code back through the synchronization pipe.

Fixes #309

PR #5107, opened 2021-07-20
oci: properly handle tty on execsync

haircommander:
#### What type of PR is this?

/kind bug
or else containers may have trouble accessing the terminal

The background here is a bit weird. I tried figuring out why we actually specified this in the first place. It seems we first introduced this when adding initial support for execsync with conmon: https://github.com/cri-o/cri-o/pull/310

As the file before that was just https://github.com/cri-o/cri-o/blob/61e60bfe4740c0a5f2a574d259f98ba9834fa4e6/oci/oci.go#L181 which makes no mention of tty.

let's see what the tests think.

related to https://bugzilla.redhat.com/show_bug.cgi?id=1983205

edit: Turns out this is needed, for https://github.com/cri-o/cri-o/pull/1386/files and https://bugzilla.redhat.com/show_bug.cgi?id=1549029. changed the PR to account for this case

```release-note
fix an issue where a container started with a terminal fails on exec sync calls
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -265,54 +265,8 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
 		return err
 	}
 	defer os.RemoveAll(processFile)
-
 	execCmd := r.constructExecCommand(ctx, c, processFile, "")
-	var cmdErr, copyError error
-	if tty {
-		cmdErr = ttyCmd(execCmd, stdin, stdout, resize)
-	} else {
-		var r, w *os.File
-		if stdin != nil {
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
LGTM

haircommander:
ah we actually inhereited it in https://github.com/cri-o/cri-o/pull/1386. looking to make sure we don't break the same way we did

mrunalp:
Discussed with @haircommander offline that this will regress https://github.com/cri-o/cri-o/pull/1386/files for https://bugzilla.redhat.com/show_bug.cgi?id=1549029.

mrunalp (approved):
/lgtm

PR #5131, opened 2021-07-26
[1.19] introduce seccomp override empty

haircommander:
#### What type of PR is this?

/kind feature
> /kind flake

right now, the CRI defaults to an unconfined profile when it's set to empty
However, it's possible admins want the behavior to default to the default profile instead.

add a knob to allow admins to override the empty behavior from unconfined to the default
Note, when set this technically makes CRI-O not CRI compliant, but for increased security, it's worth it.

```release-note
Add option `seccomp_override_empty` to override an unspecified seccomp profile from being unconfined to being the runtime default. Note: setting this option makes CRI-O not fully CRI compliant, but does increase security. 
```


saschagrunert:
@haircommander what's the reasoning behind this backport? I think we should be conservative about cherry-picking features like this.

haircommander:
yeah I agree we should generally be careful, but this is a useful feature that allows users to cover a class of problems that are annoying to cover otherwise (making the seccomp profile applied by default)

saschagrunert (approved):
/lgtm
/retest

PR #5136, opened 2021-07-29
oci: fix a performance regression from execs

haircommander:
#### What type of PR is this?

/kind bug
turns out, starting a new inotify watcher for each exec probe was really inefficient. Replace it with one single watcher for all execs, which makes the RSS addition per exec much less (pretty much the overhead of exec.Cmd, which is unavoidable)

```release-note
fix a performance regression with exec probes
```


TomSweeneyRedHat:
Just curious why you're creating notify in CRI-O rather than fixing upstream?
Otherwise, LGTM once the last unhappy test gets happy.

haircommander:
> Just curious why you're creating notify in CRI-O rather than fixing upstream?
> Otherwise, LGTM once the last unhappy test gets happy.

I'm not super sure which upstream you're talking about. I would really like to get a timeout feature into runc, but there is a lot of nuance to the behavior here (if exec takes more than x seconds, sig kill it is not a very generic requirement). So we do the next best thing: our best in CRI-O.

TomSweeneyRedHat:
Maybe I misunderstood this part, but it looked like you dropped github.com/rjeczalik/notify and replaced it with your own spin on notify.

internal/oci/runtime_oci.go
```diff
@@ -1101,29 +1101,38 @@ func (c *Container) conmonPidFilePath() string {
...
+	}
+
 	ch := make(chan struct{})
 
 	dir := filepath.Dir(path)
-	if err := notify.Watch(dir, eiCh, opsToWatch...); err != nil {
-		return nil, err
-	}
 	go func() {
+		defer watcher.Close()
```
mrunalp, line 1107: This could be put after the err check on NewWatcher.

internal/oci/runtime_oci.go
```diff
@@ -1101,29 +1101,38 @@ func (c *Container) conmonPidFilePath() string {
...
 		defer close(ch)
-		defer close(eiCh)
-		defer notify.Stop(eiCh)
 		for {
 			select {
```
mrunalp, line 1110: Comparing with the container exit monitor, we are missing a case for wacher.Errors.

internal/oci/runtime_oci.go
```diff
@@ -285,13 +286,7 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
...
 	defer os.RemoveAll(processFile)
 
-	pidDir, err := ioutil.TempDir("", "pidfile")
-	if err != nil {
-		return nil, err
-	}
-	defer os.RemoveAll(pidDir)
-
-	pidFile := filepath.Join(pidDir, c.id)
+	pidFile := filepath.Join(r.execNotifier.Directory(), c.id+uuid.New().String())
```
saschagrunert, line 289: What about using `ioutil.TempFile` together with the `r.execNotifier.Directory`?
```suggestion
	pidFile, err := ioutil.TempFile(r.execNotifier.Directory(), "exec-")
```

saschagrunert (commented):
Just a nit, otherwise LGTM

internal/oci/runtime_oci.go
haircommander: ioutil.TempFile creates the file, which interacts strangely with the inotify code. I think it's better staying as is

internal/oci/runtime_oci.go
haircommander: I leave it here because there exists the chance we'll want to add a done channel in this goroutine (it does not seem to be needed now). Leaving it in the goroutine allows us just to extend the for loop, not move around the channels. If that isn't clear though I could document/change it. wydt

haircommander:
> Maybe I misunderstood this part, but it looked like you dropped github.com/rjeczalik/notify and replaced it with your own spin on notify.

ah not quite. I introduced a second library in the initial commit because it seemed to work better. I reverted that addition in this PR

saschagrunert (approved):
/lgtm
/retest

PR #5144, opened 2021-07-30
main: Don't log an error if the clean shutdown file doesn't exist

cgwalters:
This showed up when I was grepping a journal for `error:`, and it's
not an error.

```release-note
NONE
```

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -243,8 +243,7 @@ func main() {
 
 		if config.CleanShutdownFile != "" {
 			// clear out the shutdown file
-			if err := os.Remove(config.CleanShutdownFile); err != nil {
-				// not a fatal error, as it could have been cleaned up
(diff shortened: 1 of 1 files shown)


haircommander:
good point, thanks for the PR

/ok-to-test
/approve

TomSweeneyRedHat:
LGTM
assuming happy tests

cmd/crio/main.go
```diff
@@ -243,8 +243,8 @@ func main() {
 
 		if config.CleanShutdownFile != "" {
 			// clear out the shutdown file
-			if err := os.Remove(config.CleanShutdownFile); err != nil {
-				// not a fatal error, as it could have been cleaned up
+			if err := os.Remove(config.CleanShutdownFile); err != nil && !os.IsNotExist(err) {
+				// continue anyways
```
haircommander, line 247: apparently the linter thinks this is a commented out `continue` :upside_down_face:

cgwalters: Hah.  I removed it so we can continue

Issue #5148, opened 2021-07-31
CRI-O is not honoring Stopsignal passed as signal number


trishnendu:
In CRI-O code , the GetStopSignal() function is passing the input string directly to unix.SignalNum function without checking if it is an integer (in string format). As unix.SignalNum() supports only a limited number of POSIX signals, for special stop signals (like SIGRTMIN+3 needed to stop  systemd pods), we need to pass it by signal number. Containerd and docker are both supporting StopSignal to be passed as signal number or a signal name

haircommander:
fixed inhttps://github.com/cri-o/cri-o/pull/5366

PR #5149, opened 2021-07-31
Adding support to honor signal number passed as StopSignal

trishnendu:
Issue - [#5148](https://github.com/cri-o/cri-o/issues/5148)


internal/oci/container.go
```diff
@@ -183,6 +183,11 @@ func (c *Container) GetStopSignal() string {
 	if c.stopSignal == "" {
 		return defaultStopSignal
 	}
+
+	if _, err := strconv.Atoi(c.stopSignal); err == nil {
+		return c.stopSignal
+	}
```
saschagrunert, line 189: Let's use `unix.SignalName` here to verify the signal number actually exists.

internal/oci/container.go
```diff
@@ -199,6 +204,10 @@ func (c *Container) StopSignal() syscall.Signal {
 		return defaultStopSignalInt
 	}
 
+	if signal, err := strconv.Atoi(c.stopSignal); err == nil {
+		return syscall.Signal(signal)
+	}
```
saschagrunert, line 209: Same for this one, let's verify that the signal number actually exists.

saschagrunert:
Thank you for the contribution @trishnendu! :pray: Can you please sign-off your commits by `git commit --amend -s`?

haircommander:
/ok-to-test
/approve

besides other nits, this LGTM. thanks a bunch @trishnendu !

TomSweeneyRedHat:
LGTM
once @saschagrunert 's concerns are addressed and the tests are happy

internal/oci/container.go
trishnendu: The only issue I see with verifying the signal number using Unix.SignalName is that the go unix package has Signal Name defined only upto Signal 31. For few uses cases, we need to use real time signals (SIG 34-64), like to stop systemd pods gracefully SIGRTMIN+3 (signal 37) needs to be sent as stop signal.

saschagrunert: Oh, okay. Maybe we can extend the range then?

haircommander: @trishnendu wdyt adapting https://github.com/cri-o/cri-o/pull/5366 to fit this case?

haircommander: we could make sure the number is less that RTMAX (defined https://github.com/containers/common/blob/main/pkg/signal/signal_linux.go) and then case to syscall.Signal from there?

trishnendu: I wasn't aware of newly added ParseSignal function, #5366 lgtm. 
Right, a upper value check on returned value of ParseSignal is needed

PR #5157, opened 2021-08-03
display build tags in version command

rockmenjack:
/kind feature

#### What this PR does / why we need it:

When I install pre-compiled binary from distribution repo, there is no way to know whether some particular feature is compiled into the binary, e.g. is btrfs supported or excluded in purpose? The only way to know is to search the build script in the package src, which is very inconvenient.

This PR adds display of build tags information in `crio version`

```bash
$ ./bin/crio version
INFO[2021-08-03 11:54:02.066868576+08:00] Starting CRI-O, version: 1.22.0, git: 79a25c4c3123e1adb8511848564bcbcd0d853a3f(dirty) 
Version:       1.22.0
GitCommit:     79a25c4c3123e1adb8511848564bcbcd0d853a3f
GitTreeState:  dirty
BuildDate:     2021-08-03T03:53:21Z
GoVersion:     go1.16.6
Compiler:      gc
Platform:      linux/amd64
Linkmode:      dynamic
BuildTags:     containers_image_ostree_stub seccomp  #  <- new added
```

PR #5158, opened 2021-08-03
fix debian 10 build doc

rockmenjack:
/kind documentation

Based on my attempt to build crio with btrfs graph driver in Debian 10(where the package in opensuse kubic seems missing it).

With help of the info provided by https://github.com/cri-o/cri-o/pull/5157

### DIFF
--- a/install.md
+++ b/install.md
@@ -106,22 +106,24 @@ yum install cri-o
 Note: this tutorial assumes you have curl and gnupg installed
 
 To install on the following operating systems, set the environment variable $OS as the appropriate field in the following table:
+
 | Operating system | $OS               |
 | ---------------- | ----------------- |
 | Debian Unstable  | `Debian_Unstable` |
 | Debian Testing   | `Debian_Testing`  |
+| Debian 10        | `Debian_10`       |
 | Raspberry Pi OS  | `Raspbian_10`     |
 | Ubuntu 20.04     | `xUbuntu_20.04`   |
 | Ubuntu 19.10     | `xUbuntu_19.10`   |
 | Ubuntu 19.04     | `xUbuntu_19.04`   |
 | Ubuntu 18.04     | `xUbuntu_18.04`   |
 
-If installing cri-o-runc (recommended), you'll need to install libseccomp >= 2.4.1. This is not available in distros based on Debian buster or below, so buster backports will need to be enabled:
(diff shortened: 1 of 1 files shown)


saschagrunert:
@rockmenjack thank you for the contribution! :pray: Can you please sign-off you commit by `git commit --amend -s` and force-pushing again?

install.md
```diff
@@ -255,7 +257,9 @@ On Debian, Raspbian and Ubuntu distributions, [enable the Kubic project
 repositories](../README.md#installing-crio) and install the following packages:
 
 ```bash
-apt-get update -qq && apt-get install -y \
+. /etc/os-release
+apt-get update -qq && \
+apt-get install -y -t ${VERSION_CODENAME}-backports \
```
haircommander, line 262: I think we may want an explanation on what VERSION_CODENAME IS (maybe add it to the table above?)

haircommander (approved):
Thanks for the PR @rockmenjack! Change generally LGTM, just one nit

TomSweeneyRedHat:
LGTM 
once @haircommander 's concern is addressed.

install.md
rockmenjack: For Debian 10 the VERSION_CODENAME would be "buster".
Maybe I shall just split the instructions into Debian 10(buster) and others(because code name would be different and I don't have such systems at the moment. Any idea?

haircommander: I think the cleanest way is to change the table above so that there's a column "CODENAME" and all of the debian flavors have "buster" in it. Then, when we support debian 11, we can just extend the table with a different name

install.md
```diff
@@ -121,7 +123,7 @@ If installing cri-o-runc (recommended), you'll need to install libseccomp >= 2.4
 ```shell
 echo 'deb http://deb.debian.org/debian buster-backports main' > /etc/apt/sources.list.d/backports.list
 apt update
-apt install -y libseccomp2 || apt update -y libseccomp2
+apt install -y -t buster-backports libseccomp2 || apt update -y -t buster-backports libseccomp2
```
haircommander, line 126: we hardcode `buster` here  btw

install.md
rockmenjack: as shown in http://ftp.us.debian.org/debian/dists/, there will be no backport repo for Debian unstable, I think we have to give different instruction for different versions in the end

haircommander: Another option is we could say "if you're using debian 10 or related distros, you may need to add buster-backports to install the required packages" and then update for 11 one day

haircommander:
/approve
/ok-to-test
/release-note-none

@cri-o/cri-o-maintainers PTAL

PR #5162, opened 2021-08-03
[1.21] oci: fix a performance regression from execs

haircommander:
#### What type of PR is this?

/kind bug
turns out, starting a new inotify watcher for each exec probe was really inefficient. Replace it with one single watcher for all execs, which makes the RSS addition per exec much less (pretty much the overhead of exec.Cmd, which is unavoidable)
cherry-pick of https://github.com/cri-o/cri-o/pull/5136

```release-note
fix a performance regression with exec probes
```


mrunalp (approved):
/lgtm

PR #5173, opened 2021-08-05
drop crio.shutdown support

haircommander:
#### What type of PR is this?

/kind design

The interaction between crio.shutdown and crio wipe has always been kind of confusing. As we move towards internal wipe being the default (set in [this](github.com/cri-o/cri-o/pull/5218/) PR), it removes the need for crio.shutdown. Now, detecting whether a shutdown has happened happens when containers are restored, and no external files are consulted

```release-note
Drop support for the crio.shutdown.
```


TomSweeneyRedHat:
LGTM
but would like @mrunalp to take a looksee

server/server.go
```diff
@@ -295,26 +294,8 @@ func (s *Server) restore(ctx context.Context) []string {
 	return imagesOfDeletedContainers
 }
 
-// cleanupSandboxesOnShutdown Remove all running Sandboxes on system shutdown
-func (s *Server) cleanupSandboxesOnShutdown(ctx context.Context) {
-	_, err := os.Stat(shutdownFile)
-	if err == nil || !os.IsNotExist(err) {
-		log.Debugf(ctx, "Shutting down all sandboxes, on shutdown")
-		s.stopAllPodSandboxes(ctx)
```
saschagrunert, line 303: Isn't this function unused after that and can be removed as well?

haircommander:
good point! fixed

server/server.go
haircommander: good point! fixed

mrunalp (approved):
/lgtm

PR #5191, opened 2021-08-11
server: return a better error when container requests pod pid ns, but…

haircommander:
… infra not created

#### What type of PR is this?

/kind bug

```release-note
none
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -425,7 +425,10 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 	} else if securityContext.NamespaceOptions.Pid == types.NamespaceModePOD {
 		pidNsPath := sb.PidNsPath()
 		if pidNsPath == "" {
-			return nil, errors.New("PID namespace requested, but sandbox infra container invalid")
+			if sb.NamespaceOptions().Pid != types.NamespaceModePOD {
+				return nil, errors.New("Pod level PID namespace requested for the container, but pod sandbox was not similarly configured, and does not have an infra container")
+			}
+			return nil, errors.New("PID namespace requested, but sandbox infra container unexpectedly invalid")
 		}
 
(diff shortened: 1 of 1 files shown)


server/container_create_linux.go
```diff
@@ -425,7 +425,10 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 	} else if securityContext.NamespaceOptions.Pid == types.NamespaceModePOD {
 		pidNsPath := sb.PidNsPath()
 		if pidNsPath == "" {
-			return nil, errors.New("PID namespace requested, but sandbox infra container invalid")
+			if sb.NamespaceOptions().Pid != types.NamespaceModePOD {
+				return nil, errors.New("Pod level PID namespace requested for container, but pod sandbox was not similarly configured, and does not have an infra container.")
```
TomSweeneyRedHat, line 429:
```suggestion
				return nil, errors.New("Pod level PID namespace requested for the container, but pod sandbox was not similarly configured and does not have an infra container.")
```

TomSweeneyRedHat:
Couple nit suggestions, otherwise LGTM once tests are hip

server/container_create_linux.go
```diff
@@ -425,7 +425,10 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
...
+			}
+			return nil, errors.New("PID namespace requested, but sandbox infra container unexpectedly invalid.")
```
saschagrunert, line 431:
```suggestion
			return nil, errors.New("PID namespace requested, but sandbox infra container unexpectedly invalid")
```

server/container_create_linux.go
saschagrunert: Please also remove the dot from the error message at the end.

mrunalp (approved):
/lgtm

PR #5218, opened 2021-08-18
config: set internal_wipe to true by default, and deprecate the option

haircommander:

#### What type of PR is this?

/kind deprecation

```release-note
The internal_wipe option is now true by default. Further, it is being deprecated, and will be unconditionally true in the future.
```


mrunalp (approved):
/lgtm

Issue #5165, opened 2021-08-04
CRI-O fails to detect link mode of binary on non english language OS


fracarvic:
**Description**
I'm using CRI-O static in a ubuntu server with spanish locale (LANG=es_ES.UTF-8) and logs shows the following error every few seconds:

```
ago 04 14:18:36 server crio[862]: time="2021-08-04 14:18:36.573631718+02:00" level=warning msg="Encountered error detecting link mode of binary: `ldd /usr/local/bin/crio` failed: \tno es un ejecutable dinámico\n  (exit status 1)"
ago 04 14:18:46 server crio[862]: time="2021-08-04 14:18:46.633547844+02:00" level=warning msg="Encountered error detecting link mode of binary: `ldd /usr/local/bin/crio` failed: \tno es un ejecutable dinámico\n  (exit status 1)"
ago 04 14:18:56 server crio[862]: time="2021-08-04 14:18:56.687218704+02:00" level=warning msg="Encountered error detecting link mode of binary: `ldd /usr/local/bin/crio` failed: \tno es un ejecutable dinámico\n  (exit status 1)"
ago 04 14:19:06 server crio[862]: time="2021-08-04 14:19:06.747713189+02:00" level=warning msg="Encountered error detecting link mode of binary: `ldd /usr/local/bin/crio` failed: \tno es un ejecutable dinámico\n  (exit status 1)"
ago 04 14:19:16 server crio[862]: time="2021-08-04 14:19:16.804123603+02:00" level=warning msg="Encountered error detecting link mode of binary: `ldd /usr/local/bin/crio` failed: \tno es un ejecutable dinámico\n  (exit status 1)"
ago 04 14:19:26 server crio[862]: time="2021-08-04 14:19:26.866602961+02:00" level=warning msg="Encountered error detecting link mode of binary: `ldd /usr/local/bin/crio` failed: \tno es un ejecutable dinámico\n  (exit status 1)"
ago 04 14:19:36 server crio[862]: time="2021-08-04 14:19:36.918307177+02:00" level=warning msg="Encountered error detecting link mode of binary: `ldd /usr/local/bin/crio` failed: \tno es un ejecutable dinámico\n  (exit status 1)"
```

Seems that code is using text output from command ldd and expect an exact result, and it is not the case with non-English languages. The output of ldd /usr/local/bin/crio in my system is
```
$ ldd /usr/local/bin/crio
        no es un ejecutable dinámico
```

As a workaround , if I add 
```
[...]

PR #5221, opened 2021-08-19
Check `stdout` and `stderr` for linkmode

saschagrunert:
#### What type of PR is this?

/kind bug

We now check `stdout` as well as `stderr` for retrieving the linkmode of the
binary. If this does not work, then we do not warn and still report
`unknown`.

Refers to https://github.com/cri-o/cri-o/issues/5165
None

```release-note
Fixed linkmode detection by checking the `stderr` as well as `stdout` of the `ldd` command.
```

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -214,12 +214,18 @@ func getLinkmode() string {
 		return unknown
 	}
 
+	isStatic := func(input string) bool {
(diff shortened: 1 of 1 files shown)


saschagrunert:
I think switching to build-tags is better, I'll propose this in a new PR.

Issue #5219, opened 2021-08-19
possible bug in openpgp_tag.sh


rockmenjack:
Hi, I am trying to compile crio and enable image signing support, but in [hack/openpgp_tag.sh](https://github.com/cri-o/cri-o/blob/master/hack/openpgp_tag.sh), the check is a `! gpgme-config --libs`. Are we suppose to enable this feature when  `gpgme-config --libs` succeed?

saschagrunert:
I would say yes, the logic seems inverted. Do you wanna open a PR to fix that?

haircommander:
fixed by #5238

PR #5238, opened 2021-08-22
fix checking in openpgp_tag.sh

rockmenjack:
a handy fix, see https://github.com/cri-o/cri-o/issues/5219

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -39,6 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - run: scripts/github-actions-packages
     - run: make docs-validation
 
   vendor:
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
/ok-to-test

Thank you @rockmenjack

saschagrunert:
@rockmenjack the verify/docs CI job fails, I think we can solve it by adding something like
https://github.com/cri-o/cri-o/blob/master/.github/workflows/integration.yml#L27
to
https://github.com/cri-o/cri-o/blob/177ed5f1a66294a69764676a864777bbfa4d09c4/.github/workflows/verify.yml#L38-L42

TomSweeneyRedHat:
LGTM
assuming happy tests

saschagrunert (approved):
/lgtm

PR #5244, opened 2021-08-23
server: fix some hostNet handling in container creation

haircommander:
#### What type of PR is this?

/kind bug
as well as fix an issue where hostNet containers couldn't access cgroup heierarchy 

fixes https://github.com/kubernetes/kubernetes/issues/104034

```release-note
Fix a bug where host network containers couldn't read their cgroup hierarchy 
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -458,12 +458,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 		}
 
 		if !isInCRIMounts("/sys", containerConfig.Mounts) {
-			specgen.RemoveMount("/sys")
 			ctr.SpecAddMount(rspec.Mount{
 				Destination: "/sys",
 				Type:        "sysfs",
 				Source:      "sysfs",
 				Options:     []string{"nosuid", "noexec", "nodev", "ro"},
+			})
+			ctr.SpecAddMount(rspec.Mount{
+				Destination: "/sys/fs/cgroup",
+				Type:        "cgroup",
+				Source:      "cgroup",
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests

saschagrunert (commented):
Code LGTM, having an integration test for that would be great.

haircommander:
failures reproduced, looking

haircommander:
test fails when using a userns and I don't think it's the fault of this PR, so 'I've dropped it for now

haircommander:
old failures do trigger here https://github.com/cri-o/cri-o/pull/5344

we'll stall on a test addition until that's figured out, but we should move forward here

server/container_create_linux.go
```diff
@@ -458,13 +458,18 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 		}
 
 		if !isInCRIMounts("/sys", containerConfig.Mounts) {
-			specgen.RemoveMount("/sys")
```
mrunalp, line 461: What if we are privileged and host network?

haircommander: the mount generating code is all over the place tbh. Right after this block, we overwrite the mount if we're privileged

mrunalp: Can we skip the add if we are also privileged since they are added rw in that case?

haircommander:
@wgahnagl @QiWang19 @fgiudici PTAL

QiWang19:
LGTM

fgiudici (approved):
LGTM!

PR #5247, opened 2021-08-25
Bug 1989339: Remove log symlinks when executing wipe

harche:

#### What type of PR is this?

/kind bug

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1989339

```release-note
None
```

### DIFF
--- a/cmd/crio/wipe.go
+++ b/cmd/crio/wipe.go
@@ -1,7 +1,11 @@
 package main
 
 import (
+	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
+	"regexp"
 
 	cstorage "github.com/containers/storage"
 	"github.com/cri-o/cri-o/internal/criocli"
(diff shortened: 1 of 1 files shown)


cmd/crio/wipe.go
```diff
@@ -195,7 +201,36 @@ func (c ContainerStore) deleteContainer(id string) {
...
+
+func (c ContainerStore) removeLogSymlinks(id string) error {
+	files, err := ioutil.ReadDir(kubeletLogDir)
+	if err != nil {
+		return err
+	}
+
+	re := regexp.MustCompile(`\S.+-(\S{64})\.log`)
+
+	for _, f := range files {
```
saschagrunert, line 219: I'm wondering why we have to iterate at all. Can we just construct the logfile name from the ID and then try to remove it?

saschagrunert (commented):
Closing the loop: I think we may handle this in the kubelet, since there seems to be a cleanup routine for that already: https://github.com/kubernetes/kubernetes/blob/1a845ccd076bbf1b03420fe694c85a5cd3bd6bed/pkg/kubelet/kuberuntime/kuberuntime_gc.go#L351

cmd/crio/wipe.go
harche: Makes sense.

PR #5260, opened 2021-08-27
internal/process: add functionality to clean up zombie children

haircommander:
Now that exec sync requests are run by conmon, there are more processes in the mix and more possibility for zombies

this commit adds a zombie monitor to the defunct process metrics collection flow. it is a little clunky, but it would be weird
to have two different /proc parsers for very similar uses

#### What type of PR is this?

/kind feature

```release-note
add a zombie monitor to cleanup CRI-O's defunct children
```

### DIFF
--- a/internal/process/defunct_processes.go
+++ b/internal/process/defunct_processes.go
@@ -16,50 +16,60 @@ const ProcessFS = "/proc"
 
 // Stat represents status information of a process from /proc/[pid]/stat.
 type Stat struct {
+	// Pid is the PID of the process
+	Pid int
+
 	// Comm is the command name (usually the executable filename).
 	Comm string
 
 	// State is the state of the process.
 	State string
+
+	// PPid is the parent PID of the process
+	PPid int
 }
 
-// DefunctProcesses returns the number of zombie processes in the node.
-func DefunctProcesses() (defunctCount uint, retErr error) {
-	return DefunctProcessesForPath(ProcessFS)
+// ParseDefunctProcesses returns the number of defunct processes on the node,
+// as well as the number of defunct children of the current running process.
+func ParseDefunctProcesses() (defunctCount uint, defunctChildren []int, retErr error) {
+	return ParseDefunctProcessesForPathAndParent(ProcessFS, os.Getpid())
 }
 
-// DefunctProcessesForPath retrieves the number of zombie processes from
-// a specific process filesystem.
-func DefunctProcessesForPath(path string) (defunctCount uint, retErr error) {
+// ParseDefunctProcessesForPath retrieves the number of zombie processes from
+// a specific process filesystem, as well as the number of defunct children of a given parent.
+func ParseDefunctProcessesForPathAndParent(path string, parent int) (defunctCount uint, defunctChildren []int, retErr error) {
 	directories, err := os.Open(path)
 	if err != nil {
-		return 0, err
+		return 0, defunctChildren, err
 	}
 	defer directories.Close()
 
 	names, err := directories.Readdirnames(-1)
 	if err != nil {
-		return 0, err
+		return 0, defunctChildren, err
 	}
 
 	for _, name := range names {
 		// Processes have numeric names. If the name cannot
 		// be parsed to an int, it is not a process name.
-		if _, err := strconv.ParseInt(name, 10, 0); err != nil {
+		pid, err := strconv.ParseInt(name, 10, 0)
+		if err != nil {
 			continue
 		}
 
 		stat, err := processStats(path, name)
 		if err != nil {
-			logrus.Debugf("Failed to get the status of process with PID %s: %v", name, err)
 			continue
 		}
 		if stat.State == "Z" {
-			logrus.Warnf("Found defunct process with PID %s (%s)", name, stat.Comm)
 			defunctCount++
+			logrus.Warnf("Found defunct process with PID %s (%s)", name, stat.Comm)
+			if stat.PPid == parent {
+				defunctChildren = append(defunctChildren, int(pid))
+			}
 		}
 	}
-	return defunctCount, nil
+	return defunctCount, defunctChildren, nil
 }
 
 // processStats returns status information of a process as defined in /proc/[pid]/stat
@@ -73,14 +83,25 @@ func processStats(fsPath, pid string) (*Stat, error) {
 	// /proc/[PID]/stat format is described in proc(5). The second field is process name,
 	// enclosed in parentheses, and it can contain parentheses inside. No other fields
 	// can have parentheses, so look for the last ')'.
-	i := strings.LastIndexByte(data, ')')
-	if i <= 2 || i >= len(data)-1 {
+	commEnd := strings.LastIndexByte(data, ')')
+	if commEnd <= 2 || commEnd >= len(data)-1 {
 		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
 	}
 
-	parts := strings.SplitN(data[:i], " (", 2)
+	parts := strings.SplitN(data[:commEnd], " (", 2)
 	if len(parts) != 2 {
 		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
+	}
+
+	stateIdx := commEnd + 2
+
+	// the fourth field is PPid, and we can start looking after the space after State
+	ppidBegin := stateIdx + 2
+	ppidEnd := strings.IndexByte(data[ppidBegin:], ' ')
+
+	ppid, err := strconv.ParseInt(data[ppidBegin:ppidBegin+ppidEnd], 10, 0)
+	if err != nil {
+		return nil, errors.Errorf("invalid stat data (invalid ppid): %q", data[stateIdx+2:ppidEnd])
 	}
 
 	return &Stat{
@@ -88,6 +109,8 @@ func processStats(fsPath, pid string) (*Stat, error) {
 		Comm: parts[1],
 
 		// The state is field 3, which is the first two fields and a space after.
-		State: string(data[i+2]),
+		State: string(data[stateIdx]),
+
+		PPid: int(ppid),
 	}, nil
 }
--- a/internal/process/zombie_monitor.go
+++ b/internal/process/zombie_monitor.go
@@ -1,0 +1,53 @@
+package process
+
+import (
+	"syscall"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// ZombieMonitor is a structure for watching and cleaning up zombies on the node.
+// It is responsible for cleaning up zombies that are children of the currently running process.
+// It does so by occasionally polling for the zombie processes.
+// If any zombies are found, there is a delay between when they're identified and when they're cleaned.
+// This is to ensure ZombieMonitor doesn't interfere with the go runtime's own child management.
+type ZombieMonitor struct {
+	closeChan chan struct{}
+}
+
+// NewZombieMonitor creates and starts the zombie monitor.
+func NewZombieMonitor() *ZombieMonitor {
+	monitor := &ZombieMonitor{
+		closeChan: make(chan struct{}, 1),
+	}
+	go monitor.Start()
+	return monitor
+}
+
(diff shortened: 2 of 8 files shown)


internal/process/defunct_processes.go
```diff
@@ -16,50 +16,61 @@ const ProcessFS = "/proc"
...
+	return ParseDefunctProcessesForPathAndParent(ProcessFS, os.Getpid())
 }
 
-// DefunctProcessesForPath retrieves the number of zombie processes from
-// a specific process filesystem.
-func DefunctProcessesForPath(path string) (defunctCount uint, retErr error) {
+// ParseDefunctProcessesForPath retrieves the number of zombie processes from
+// a specific process filesystem, as well as the number of defunct children of the current running process.
+func ParseDefunctProcessesForPathAndParent(path string, parent int) (defunctCount uint, defunctChildren []int, retErr error) {
+	defunctChildren = make([]int, 0)
```
saschagrunert, line 41: This line seems not required
```suggestion
```

internal/process/zombie_monitor.go
```diff
@@ -0,0 +1,79 @@
...
+type ZombieMonitor struct {
+	zombieCount uint32
+	closeChan   chan struct{}
+}
+
+// InitializeZombieMonitor creates and starts the zombie monitor.
+// It fails if one has already been created.
+func InitializeZombieMonitor() error {
+	if monitor != nil {
+		return errors.New("zombie monitor already intialized")
```
saschagrunert, line 35: I know this will probably not happen, but we could return the existing `monitor` here.

server/metrics/metrics.go
```diff
@@ -175,12 +175,7 @@ func New(config *libconfig.MetricsConfig) *Metrics {
 				Help:      "Total number of defunct processes in the node",
 			},
 			func() float64 {
-				total, err := process.DefunctProcesses()
-				if err == nil {
-					return float64(total)
-				}
-				logrus.Warn(err)
-				return 0
+				return float64(process.TotalDefunctProcesses())
```
saschagrunert, line 178: This makes the "return the current state of the system" call time-wise dependent on the zombie monitor (runs every 5 seconds for now). I'm not sure if we wanna have this coupling, how about separating both?

internal/process/zombie_monitor.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Start begins the zombie monitor. It will populate the zombie count,
+// as well as begin the zombie cleaning process.
+func (zm *ZombieMonitor) Start() {
+	for {
+		count, zombieChildren, err := ParseDefunctProcesses()
+		if err != nil {
+			log.Warnf(context.Background(), "Failed to get defunct process information: %v", err)
+		}
+		atomic.StoreUint32(&zm.zombieCount, uint32(count))
```
saschagrunert, line 58: Should this run after the `case <-time.After(time.Second * 5):` below?

internal/process/zombie_monitor.go
```diff
@@ -0,0 +1,79 @@
...
+		select {
+		case <-zm.closeChan:
+			// Since the process will soon shutdown, and its children will be reparented, no need to delay the shutdown to cleanup.
+			return
+		case <-time.After(time.Second * 5):
```
saschagrunert, line 63: Is this too often? How about giving the whole cleanup more time, e.g. 15 minutes?

internal/process/zombie_monitor.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		for _, child := range zombieChildren {
+			if _, err := syscall.Wait4(child, nil, syscall.WNOHANG, nil); err != nil {
+				log.Errorf(context.Background(), "Failed to reap child process %d: %v", child, err)
```
saschagrunert, line 67: Let's use logrus if we do not have a context to add.

server/metrics/metrics.go
haircommander: yeah I was unsure but didnt' want to duplicate work. I will split it up, which will allow us to cleanup zombies less frequently

internal/process/zombie_monitor.go
haircommander: how about 5 minutes?

internal/process/zombie_monitor.go
haircommander: dropped now

internal/process/zombie_monitor.go
haircommander: done!

haircommander:
comments addressed!

internal/process/zombie_monitor.go
saschagrunert: Sounds good

saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM
but a few tests need some help

internal/process/defunct_processes.go
```diff
@@ -16,50 +16,60 @@ const ProcessFS = "/proc"
...
-// DefunctProcessesForPath retrieves the number of zombie processes from
-// a specific process filesystem.
-func DefunctProcessesForPath(path string) (defunctCount uint, retErr error) {
+// ParseDefunctProcessesForPath retrieves the number of zombie processes from
+// a specific process filesystem, as well as the number of defunct children of the current running process.
+func ParseDefunctProcessesForPathAndParent(path string, parent int) (defunctCount uint, defunctChildren []int, retErr error) {
 	directories, err := os.Open(path)
 	if err != nil {
-		return 0, err
+		return 0, defunctChildren, err
```
kolyshkin, line 43: Maybe something like
```go
retErr = err
return
```
would look better here.

internal/process/defunct_processes.go
```diff
@@ -73,21 +83,34 @@ func processStats(fsPath, pid string) (*Stat, error) {
...
+	parts := strings.SplitN(data[:endOfComm], " (", 2)
 	if len(parts) != 2 {
 		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
 	}
 
+	stateByte := endOfComm + 2
+
+	// the fourth field is PPid, and we can start looking after the space after State
+	ppidBegin := stateByte + 2
+	endOfPPid := strings.IndexRune(data[ppidBegin:], ' ')
```
kolyshkin, line 100: Maybe "ppidBegin" / "ppidEnd"?

internal/process/defunct_processes.go
```diff
@@ -73,21 +83,34 @@ func processStats(fsPath, pid string) (*Stat, error) {
 	// /proc/[PID]/stat format is described in proc(5). The second field is process name,
 	// enclosed in parentheses, and it can contain parentheses inside. No other fields
 	// can have parentheses, so look for the last ')'.
-	i := strings.LastIndexByte(data, ')')
-	if i <= 2 || i >= len(data)-1 {
+	endOfComm := strings.LastIndexByte(data, ')')
```
kolyshkin, line 86: `commEnd`?

internal/process/defunct_processes.go
kolyshkin, line 100: Please use `IndexByte` here.

internal/process/defunct_processes.go
```diff
@@ -73,21 +83,34 @@ func processStats(fsPath, pid string) (*Stat, error) {
...
 		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
 	}
 
-	parts := strings.SplitN(data[:i], " (", 2)
+	parts := strings.SplitN(data[:endOfComm], " (", 2)
 	if len(parts) != 2 {
 		return nil, errors.Errorf("invalid stat data (no comm): %q", data)
 	}
 
+	stateByte := endOfComm + 2
```
kolyshkin, line 96: `stateIdx`?

internal/process/zombie_monitor.go
```diff
@@ -0,0 +1,55 @@
...
+	"syscall"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// ZombieMonitor is a structure for watching and cleaning up zombies on the node.
+// It has two tasks, both related to detecting the existence of defunct processes:
+// 1: it is responsible for reporting the number of defunct PIDs currently on the node
+// 2: it is responsible for cleaning up zombies that are children of the currently running process
```
kolyshkin, line 13: `s/it is responsible for//`

internal/process/defunct_processes.go
```diff
@@ -16,50 +16,60 @@ const ProcessFS = "/proc"
...
+// ParseDefunctProcesses returns the number of defunct processes on the node,
+// as well as the number of defunct children of the current running process.
+func ParseDefunctProcesses() (defunctCount uint, defunctChildren []int, retErr error) {
+	return ParseDefunctProcessesForPathAndParent(ProcessFS, os.Getpid())
 }
 
-// DefunctProcessesForPath retrieves the number of zombie processes from
-// a specific process filesystem.
-func DefunctProcessesForPath(path string) (defunctCount uint, retErr error) {
+// ParseDefunctProcessesForPath retrieves the number of zombie processes from
```
kolyshkin, line 38: Is this function public only because we have to test it, and the test package is named `..._test`?

If yes, we're increasing the API surface (even if it's internal) to merely test it, and this does seem like a good practice to me .

kolyshkin:
@haircommander I think I missed the discussion on where those zombies are coming from, can you point me out to any details on that?

haircommander:
> @haircommander I think I missed the discussion on where those zombies are coming from, can you point me out to any details on that?

we've had a number of reports that there are `conmon` zombies that are leaking (https://bugzilla.redhat.com/show_bug.cgi?id=1952137 is one). We initially thought it was a bug in the exec implementation, but I am now thinking it was actually leaks that were closed in https://github.com/cri-o/cri-o/pull/5283. 

Part of me is less convinced we need this anymore, though it may not be a bad idea to sweep for them periodically

internal/process/defunct_processes.go
haircommander: hm I don't think I am a fan of that style

internal/process/defunct_processes.go
haircommander: correct, and it is (perhaps unfortunately) idiomatic in the cri-o repo. I think I would leave it as is for now

internal/process/defunct_processes.go
```diff
@@ -16,50 +16,60 @@ const ProcessFS = "/proc"
 
 // Stat represents status information of a process from /proc/[pid]/stat.
 type Stat struct {
+	// Pid is the PID of the process
```
kolyshkin, line 19: Nit: missing period at the end of sentence (feel free to ignore).

internal/process/defunct_processes.go
```diff
@@ -16,50 +16,60 @@ const ProcessFS = "/proc"
...
+	Pid int
+
 	// Comm is the command name (usually the executable filename).
 	Comm string
 
 	// State is the state of the process.
 	State string
+
+	// PPid is the parent PID of the process
```
kolyshkin, line 28: ditto

internal/process/defunct_processes.go
kolyshkin: So, essentially we do two things
1. Use `<package>_test` in tests in order to have "black box" testing (i.e. only test public APIs).
2. Make internal functions we want to test public.

Seems like 2 defeats the purpose of 1 and actually make things worse.

I totally agree this is not to be addressed in this PR, but overall this seems like an issue we'd want to address.

haircommander: one thing worth noting Is since this library is in `internal/` directory it can't actually be imported publically. that does confuse the interfaces in cri-o, but at least we're the only ones who have to resist the temptation to call a function that is external but should be treated as internal

that said, this is an idiom that is consistent across cri-o, so some random person could import functions in `server` that aren't really meant for public use either..

kolyshkin (approved):
Looks good overall, I'm just not sure we need a zombie reaper if there are no bugs in the code (and if they are, they need to be fixed).

haircommander:
yeah I am leaning with you as well, especially becaus we've recently found a leak which is likely to be the cause of some of the cases we've found. Closing for now

PR #5262, opened 2021-08-27
[1.21] oci: use conmon for exec again

haircommander:
After a long saga of bug fixing and issue squashing, it turns out this is not the correct approach.

Naively, one may exepct removing conmon from the picture would help reduce the load on the system.

However, the way the go runtime hogs memory after it's been aquired, and because conmon is comparatively lighter weight,
dropping conmon from exec syncs causes a memory regression in CRI-O.

We will have to find a new solution for the excessive defunct processes.

```release-note
Fix an RSS regression with exec sync requests
```


saschagrunert (approved):
/lgtm

PR #5263, opened 2021-08-30
server: remove ineffective `updateLock`.

klihub:
#### What type of PR is this?

/kind cleanup

    Remove effectively unused Server.updateLock. It is only
    ever RLocked()/RUnlocked(). A leftover that should have
    been removed in b56da85fc.

### Which issue(s) this PR fixes:

None

None

None

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -412,8 +412,6 @@ func hostNetwork(containerConfig *types.ContainerConfig) bool {
 func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainerRequest) (res *types.CreateContainerResponse, retErr error) {
 	log.Infof(ctx, "Creating container: %s", translateLabelsToDescription(req.Config.Labels))
 
-	s.updateLock.RLock()
-	defer s.updateLock.RUnlock()
 	sb, err := s.getPodSandboxFromRequest(req.PodSandboxID)
 	if err != nil {
 		if err == sandbox.ErrIDEmpty {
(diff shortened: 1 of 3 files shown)


server/container_create.go
```diff
@@ -412,8 +412,6 @@ func hostNetwork(containerConfig *types.ContainerConfig) bool {
 func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainerRequest) (res *types.CreateContainerResponse, retErr error) {
 	log.Infof(ctx, "Creating container: %s", translateLabelsToDescription(req.Config.Labels))
 
-	s.updateLock.RLock()
-	defer s.updateLock.RUnlock()
```
saschagrunert, line 416: Can we be sure that everything on top of this stack is properly locked?

saschagrunert (commented):
/ok-to-test

server/container_create.go
klihub: TBH, I don't know. I was proposing this commit simply because of this observation (unexported RWMutex only ever being read-locked):
```
klitkey1-mobl cri-o $ git grep updateLock HEAD^
HEAD^:server/container_create.go:       s.updateLock.RLock()
HEAD^:server/container_create.go:       defer s.updateLock.RUnlock()
HEAD^:server/sandbox_run_linux.go:      s.updateLock.RLock()
HEAD^:server/sandbox_run_linux.go:      defer s.updateLock.RUnlock()
HEAD^:server/server.go: updateLock sync.RWMutex
```
Unless it is already broken, removing that lock should not break anything.

saschagrunert: Yeah I agree. There is always the risk that we reveal some missing lower level locks which would cause a regression. I think we should be fine removing those locks. @haircommander WDYT?

haircommander: yeah I agree if we're only ever read locking it it doesn't seem like we need it.

saschagrunert (approved):
LGTM

haircommander:
/lgtm

thanks @klihub for the catch and the PR!

PR #5277, opened 2021-09-03
Make /sys/fs/cgroup writable when non-privileged and using cgroups v2

### LINKED ISSUE #5166: Make /sys/fs/cgroup writable when non-privileged and using cgroups v2
In cgroups v2, cgroups can be securely delegated and contained
[(per kernel doc on cgroups v2)](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#delegation-containment).

CRI-O allocates a private cgroup namespace when the node is using cgroupv2 and the pod is not privileged.  Making /sys/fs/cgroup writable would allow the processes in the container to safely manage their own cgroups hierarchy.

My use case:
On k8s, I want a process on the pod to be able to manage cgroups that are private to the pod.  Setting up the cgroup hierarchy for a pod is the responsibility of the CRI.

ajwock:
Signed-off-by: Drew Wock <drew.wock@sas.com>

#### What type of PR is this?

/kind other

This allows users running cri-o on cgroups v2 and with appropriate permissions with respect to the container's root cgroup to manage a private cgroups v2 hierarchy.

Fixes #5166

```release-note
Mount /sys/fs/cgroup as a writable cgroup fs when non-privileged and using cgroups v2
```


server/container_create_linux.go
```diff
@@ -948,13 +948,23 @@ func addOCIBindMounts(ctx context.Context, mountLabel string, containerConfig *t
...
+        Options:     []string{"nosuid", "noexec", "nodev", "relatime", "rw"},
+      }
+		  specgen.AddMount(m)
+    } else {
+      m := rspec.Mount{
+        Destination: "/sys/fs/cgroup",
+        Type:        "cgroup",
+        Source:      "cgroup",
+        Options:     []string{"nosuid", "noexec", "nodev", "relatime", "ro"},
+      }
```
saschagrunert, line 965: To reduce code duplication, let's do something like:
```suggestion
      m := rspec.Mount{
        Destination: "/sys/fs/cgroup",
        Type:        "cgroup",
        Source:      "cgroup",
        Options:     []string{"nosuid", "noexec", "nodev", "relatime"},
      }
      if node.CgroupIsV2() {
          m.Options = append(m.Options, "rw")
      } else {
          …
```

saschagrunert (commented):
@giuseppe PTAL

giuseppe:
I am not sure this should be done by default.

Delegation is safe on cgroup v2, but I think we need to control what containers can use it.  For example a container might create a lot of sub-cgroups and that could affect the host as well.

What happens if this is used without `nsdelegate`?  Can a container override its own limits?

ajwock:
From reading man 7 cgroups, it looks like using nsdelegate would prevent a vulnerability in a couple of particular cases- one, when the container's root is the real root.  In that case, root could write to controller interface files in the root directory of the namespace, thus allowing the container not to follow restrictions set on the container by k8s or the CRI.  The other case is when other parts of the cgroups hierarchy are exposed to the container, such as if someone hostpath mounted /sys/fs/cgroup.  As the real root, a container process could move the process outside of the container's cgroup hierarchy, also allowing it not to follow container restrictions.  So I think it would be good to check for nsdelegate.

Would creating a lot of descendant cgroups affect the host in a way that just creating a large number of files wouldn't also affect it?

Regardless of whether it would, it's possible to limit the number and depth of descendant cgroups with cgroup.max.depth and cgroup.max.descendants.

Perhaps this setting could be configured in crio.conf.  A configuration option for enabling this feature and another option for setting the max number of descendants.

haircommander:
also @kolyshkin PTAL

I am in favor of having a config field to allow an admin to deny access to this

giuseppe:
> I am in favor of having a config field to allow an admin to deny access to this

IMO, it should be off by default and treated in a similar way as we do for the userns annotation

server/container_create_linux.go
```diff
@@ -268,7 +268,12 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 		processLabel = ""
 	}
 
-	containerVolumes, ociMounts, err := addOCIBindMounts(ctx, mountLabel, containerConfig, specgen, s.config.RuntimeConfig.BindMountPrefix, s.config.AbsentMountSourcesToReject)
+	var cgroupRW bool;
+	if node.CgroupIsV2() && sb.Annotations()[crioann.CgroupRWAnnotation] == "true" {
+		cgroupRW = true;
+	}
```
haircommander, line 274: we could compress this to 
`cgroupRW := node.CgroupIsV2() && sb.Annotations()[crioann.CgroupRWAnnotation] == "true"`

pkg/annotations/annotations.go
```diff
@@ -8,6 +8,9 @@ const (
 	// UsernsMode is the user namespace mode to use
 	UsernsModeAnnotation = "io.kubernetes.cri-o.userns-mode"
 
+	// CgroupRW specifies mounting v2 cgroups as an rw filesystem.
+	CgroupRWAnnotation = "io.kubernetes.cri-o.cgroup-rw"
```
haircommander, line 12: I feel like the name implies this works for all cgroups. do we want it to be Cgroup2RW?

ajwock:
A question:  since `node.CgroupIsV2()` is not mockable with gmock, what would be the best way to approach testing this change (if that's an expectation at all)

haircommander:
> A question: since `node.CgroupIsV2()` is not mockable with gmock, what would be the best way to approach testing this change (if that's an expectation at all)
> 
> I mean it is possible to unit test addOCIBindMounts without that though.

yeah our CreateContainer() testing is pretty lacking rn. I would recommend just testing addOCIBindMounts for now

haircommander:
in the allowed_annotation field in the config template (pkg/config/template.go) you could add a blurb about it. If you feel so inclined, you could also put together a tutorial describing what you've done and how you did it :) or a blog would work too, which we could link in the AWESOME.md file

ajwock:
What's with the
```error: branch 'target' not found.```
and
```error: branch 'pr' not found.```
?

**Comment by haircommander** [comment]
that can be ignored, the actual error is
```
time="2021-09-29T16:41:43Z" level=error msg="Error copying docker://quay.io/crio/redis:alpine to dir:/go/src/github.com/cri-o/cri-o/.artifacts/redis-image: reading blob sha256:c72e0cff027d9cf1386d8c931e1a14e9e7d5bd7eb61ea0b4e24f021395ea5329: Error fetching blob: invalid status code from registry 500 (Internal Server Error)"
Error pulling docker://quay.io/crio/redis:alpine
make: *** [localintegration] Error 1
```

ajwock:
``` invalid status code from registry 500```
Does this relate to my PR or also just something to ignore?

haircommander:
definitely unrelated, I think quay is sad

TomSweeneyRedHat:
LGTM

ajwock:
Added the documentation blurb as described earlier

haircommander:
/retest

one last nit, otherwise LGTM

AkihiroSuda:
Is there a KEP to move this to  `Pod` spec (probably `securityOptions`)?

ajwock:
> Is there a KEP to move this to `Pod` spec (probably `securityOptions`)?

Sorry that I missed this.  As far as I know, there isn't one, but it would be a good idea to create one.

limck5856 (approved):

PR #5293, opened 2021-09-10
Do not log on startup if Intel RDT is not supported

saschagrunert:
#### What type of PR is this?

/kind cleanup

This reduces the noise on systems not having Intel RDT enabled. We already log that RDT is not enabled on profile load.

None
None

```release-note
Do not log on startup if Intel RDT is not supported.
```

### DIFF
--- a/internal/config/rdt/rdt.go
+++ b/internal/config/rdt/rdt.go
@@ -34,7 +34,6 @@ func New() *Config {
 	rdt.SetLogger(logrus.StandardLogger())
 
 	if err := rdt.Initialize(ResctrlPrefix); err != nil {
-		logrus.Infof("RDT is not enabled: %v", err)
 		c.supported = false
 	}
 	return c


haircommander:
would you remove the corresponding one in internal/config/blockio/blockio.go as well?

fgiudici:
> would you remove the corresponding one in internal/config/blockio/blockio.go as well?

Hey Peter, do you mean to remove the log from the Load() in blockio.go as well? If so,  wouldn't that be a bit different than here (removing the log from the New() only)?

haircommander:
> > would you remove the corresponding one in internal/config/blockio/blockio.go as well?
> 
> Hey Peter, do you mean to remove the log from the Load() in blockio.go as well? If so, wouldn't that be a bit different than here (removing the log from the New() only)?

you are right! nevermind haha

/lgtm
/retest

fgiudici:
lgtm

PR #5318, opened 2021-09-16
Add support for latest NRI with extended scope.

klihub:
## Background
This PR is one in a set of 3 related PRs which aim to extend NRI beyond its current v1 capabilities. The primary goal of the proposed extension is to allow NRI to act as a common infrastructure for plugging in extensions to runtimes, and to allow these extensions to implement, among other things, vendor- and domain-specific custom logic for improved container configuration.

This particular PR prototypes integrating cri-o with this extended scope NRI.

## Related PRs
 The other related PRs in the set are:

- [extend scope](https://github.com/containerd/nri/pull/16) of [current NRI](https://github.com/containerd/nri)
- [containerd integration](https://github.com/containerd/containerd/pull/6019) of the above

For a more detailed description of the full proposal, please refer to [this README](https://github.com/klihub/nri/blob/pr/proto/draft/v2alpha1/README.md).

## Current State
This implementation is not considered to be ready for merging (hence the RFC title tag). While there are still things to fix, plenty of rough edges to improve, unit tests are missing, etc., we do consider it to be in a good enough shape for a review round. The conceptual idea and the goals behind this proposal should be clear enough, likewise the approach chosen and the related consequences and compromises. If you are interested in any of the related subjects, please chime in. Any and all feedback is welcome.
/kind feature
```release-note
Add initial support for [Node Resource Interface (NRI)](https://github.com/containerd/nri/) v0.2.0. NRI allows vendors to customize container behavior and configuration using plugins. NRI plugins can register to various events in containers' lifecycle and make controlled changes to containers' configuration when these events occur. This feature is experimental and disabled by default. It can be enabled using the `--enable-nri` command line option or by setting `enable_nri = true` in the CRI-O configuration `[crio.nri]` table. The same table can be used to set other NRI-related configuration options. In addition to enabling NRI support in cri-o, an NRI configuration file also needs to be in place. The default location for this file is `/etc/nri/nri.conf` and it can be empty.
```


haircommander:
@cri-o/cri-o-maintainers PTAL

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/containerd/cgroups v1.0.2
 	github.com/containerd/containerd v1.5.7
 	github.com/containerd/cri-containerd v1.19.0
+	github.com/containerd/nri v0.1.0
```
saschagrunert, line 12: Are there any plans to move the NRI code out of containerd to make it more independent from the project?

internal/config/nri/nri.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// Config represents the CRI-O NRI configuration.
+type Config struct {
+	Enabled    bool   `toml:"enable_nri"`
+	ConfigPath string `toml:"nri_config_file"`
+	SocketPath string `toml:"nri_socket_path"`
+	PluginPath string `toml:"nri_plugin_dir"`
+
+	enabled bool
```
saschagrunert, line 15: What is the reason behind having this var and not just using `Enabled`?

internal/config/nri/nri.go
```diff
@@ -0,0 +1,57 @@
...
+	nri "github.com/containerd/nri/v2alpha1/pkg/runtime"
+	"github.com/pkg/errors"
+)
+
+// Config represents the CRI-O NRI configuration.
+type Config struct {
+	Enabled    bool   `toml:"enable_nri"`
+	ConfigPath string `toml:"nri_config_file"`
+	SocketPath string `toml:"nri_socket_path"`
+	PluginPath string `toml:"nri_plugin_dir"`
```
saschagrunert, line 13: Are those configuration variables actually being used?

saschagrunert (commented):
I just have a few questions about the code. Generally I'd be fine in having this kind of NRI support in CRI-O. The additional bindings to containerd related source code is a concern, while our target was to move away from them in the long run. But this should not be a blocker for the NRI from my point of view.

What are your thoughts on that?

go.mod
klihub: If our proposal to extend the scope gets eventually accepted then we'd like to move it somewhere to more neutral ground.

internal/config/nri/nri.go
klihub: They were not passed to NRI. Fixed.

internal/config/nri/nri.go
klihub: Hmm, nothing sensible. Got rid of it.

fidencio:
> I just have a few questions about the code. Generally I'd be fine in having this kind of NRI support in CRI-O. The additional bindings to containerd related source code is a concern, while our target was to move away from them in the long run. But this should not be a blocker for the NRI from my point of view.

My opinion is quite similar to @saschagrunert's opinion.  If there's something wrong with the approach, I don't see it.

One thing that crossed my mind is whether it would be reasonable to have this as **experimental** for one or two cycles, so there's still room for changes (if needed) and only then make it "official".

klihub:
> 
> My opinion is quite similar to @saschagrunert's opinion. If there's something wrong with the approach, I don't see it.
>

@saschagrunert @fidencio Thank you guys for carving out time to look at this. Really appreciated !
 
> One thing that crossed my mind is whether it would be reasonable to have this as **experimental** for one or two cycles, so there's still room for changes (if needed) and only then make it "official".

Yes, I think it should definitely be marked as experimental.

nri/config.go
```diff
@@ -0,0 +1,38 @@
+package nri
```
haircommander, line 1: nit: if we want this public, can it go in `pkg`. I'd actually lean towards internal--I wouldnt' expect anyone to import it, so we could then put it in `internal`

nri/domain.go
```diff
@@ -0,0 +1,109 @@
...
+	"sync"
+
+	"github.com/cri-o/cri-o/internal/log"
+	"github.com/sirupsen/logrus"
+
+	nri "github.com/containerd/nri/pkg/adaptation"
+)
+
+// Domain implements the functions the generic NRI interface needs to
+// deal with pods and containers from a particular containerd namespace.
```
haircommander, line 15: cri-o doesn't have a notion of  containerd namespaces, right? or does this PR introduce it and I haven't found it yet

nri/nri.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	l.state = make(map[string]State)
+
+	return l, nil
+}
+
+func (l *local) IsEnabled() bool {
+	return l != nil && !l.cfg.Disable
+}
```
haircommander, line 129: nit: I generally like having functions be defined underneath their last caller so the file can be read sequentially, but if you prefer it here it's not a hard requirement

nri/nri.go
```diff
@@ -0,0 +1,494 @@
...
+
+func (l *local) Start() error {
+	if !l.IsEnabled() {
+		return nil
+	}
+
+	err := l.nri.Start()
+	if err != nil {
```
haircommander, line 137: these can be compressed to one line

nri/nri.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	if cnt > 1 {
+		_, err = l.applyUpdates(ctx, response.Update[0:cnt-1])
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return response.Update[cnt-1].GetLinux().GetResources(), nil
```
haircommander, line 325: it looks like we could double update `cnt-1` am I misreading this?

nri/nri.go
```diff
@@ -0,0 +1,494 @@
...
+	if !l.IsEnabled() {
+		return nil
+	}
+
+	l.Lock()
+	defer l.Unlock()
+
+	err := l.stopContainer(ctx, pod, ctr)
+
+	return err
```
haircommander, line 356: this can be one line

nri/nri.go
```diff
@@ -0,0 +1,494 @@
...
+		return nil
+	}
+
+	request := &nri.StopContainerRequest{
+		Pod:       podSandboxToNRI(pod),
+		Container: containerToNRI(ctr),
+	}
+
+	response, err := l.nri.StopContainer(ctx, request)
+	l.setState(request.Container.Id, Stopped)
```
haircommander, line 370: are you sure the state be set to stopped if StopContainer returns an error? I haven't reviewed StopContainer enough to make sure, and I'll believe you if you double check for me :)

nri/nri.go
```diff
@@ -0,0 +1,494 @@
...
+	defer l.Unlock()
+
+	log.Infof(ctx, "Unsolicited container update from NRI")
+
+	failed, err := l.applyUpdates(ctx, req)
+	return failed, err
+}
+
+func (l *local) applyUpdates(ctx context.Context, updates []*nri.ContainerUpdate) ([]*nri.ContainerUpdate, error) {
+	// XXX TODO: should we pre-save state and attempt a rollback on failure ?
```
haircommander, line 453: is this TODO still standing?

server/container_remove.go
```diff
@@ -54,6 +55,16 @@ func (s *Server) removeContainerInPod(ctx context.Context, sb *sandbox.Sandbox,
 		if err := s.stopContainer(ctx, c, int64(10)); err != nil {
```
haircommander, line 55: I think stopContainer should have `s.nri.stopContainer()` so it doesn't have to be manually called below

test/nri/nri_test.go
```diff
@@ -0,0 +1,559 @@
...
+			plugins: []*plugin{nil},
+		}
+	)
+
+	t.Setup(stdT)
+	t.StartPlugins(WaitForPluginSync)
+	p := t.plugins[0]
+
+	pod := t.createPod()
+	require.NotNil(t, p.WaitEvent(RunPodEvent(pod), timeout), "pod creation event")
```
haircommander, line 97: any oppposition to using ginkgo in these? if we don't, ginkgo won't be able to filter tests out in the future (if we want to run just these or run just another test)

haircommander:
this is looking good, and close. a couple of structural nits and organization things, but it's very close.

nri/domain.go
klihub: Yes, the comment is off... needs updating. domain even as an unexported concept here is probably one of the most debatable implementation details. I gutted it out compared to the original code (which really had support for more than a single registered domain), but it's still there. If that is a show-stopper we can get rid of it. 

That said, I tried (maybe a bit too desperately) to keep the NRI integration bits here very similar to how it's done in containerd. Basically split up to two pieces: an 'egress/downstream' interface (these bits here) and the pod+container-lifecycle hook-in bits (server/nri-api.go et. al). There are two reasons. One is that it makes it a bit easier for us to maintain both as the mental burden of switching between these is smaller. The other is that I hope that eventually we could most of the NRI 'egress/downstream' bits into common code in the NRI repo itself and just import it here.

nri/nri.go
klihub: Hmm... I think we do it right. We apply all but the last update locally, and return the last one to the caller.

nri/config.go
klihub: I moved this to internal/nri.

nri/nri.go
klihub: Moved to be after last usage.

nri/nri.go
klihub: Updated.

nri/nri.go
klihub: Updated.

nri/nri.go
klihub: No. Removed.

test/nri/nri_test.go
klihub: This is actually the thing that I'm struggling/fighting with the most. So these are not really a unit tests. It is an end-to-end or integration test client for NRI and a bunch test cases.

The client part implements a rudimentary test infra. It allows test cases toestablish a gRPC CRI connection to cri-o, register one or more NRI plugins with cri-o, make sure images are pulled and present, create pods and containers, and provide some commonly used functions for test result verification.

Then there is a (currently pretty basic) set of NRI test cases implemented. They use the client infra to create/start/stop pods and containers, the NRI interface to register plugins, and both interfaces to verify that the right/expected things happen both in the plugins and to containers.

The golang/testing infra (github.com/stretchr) is merely used as a 'test driver' and for verification. The individual test cases implemented here are then tied in to end-to-end localintegration/bats tests using nri.bats, with each test case implemented here becoming an individual test case there for proper granular test failure/success reporting.

I think these could be rewritten so that they are ginkgo tests. I actually started it like that but was struggling with wrapping my head around what to put where wrt. ginkgo vs. the test infra we need, so in my frustration I switched to plain golang testing.

Now I'm struggling with a few remaining things related to this. One is a few tests failing in some of the integration runs, with something that looks like a timeout (so things just being too slow)... but I'm not sure if that's the real reason. Need to find a way to figure out.

Another one is what you pointed out above: ginkgo does not understand that these are not unit tests so it does not know to ignore them. That's a problem, because the tests cannot be run just like that. They need a properly configured cri-o (which the test cases in nri.bats set up when running them) so they just fail when those are not given during a plain unit test ginkgo run.

As a workaround, I  added and empty ginkgo test suite and a single no-op ginkgo test case and simple logic to TestMain to omit running the tests. This takes care of the ginkgo test failure, but leaves the tests without a coverage report... which still leads to a failure then. I don't know if this could be due to me skipping all tests by not doing a testing.M.Run(), but I suspect it might be. So now I'm thinking I'll try to change this into all tests just doing a testing.T.Skip() instead when we need to omit running.

Anyway, considering all this, can you think of a better way to do any or all of this which would eliminate the problems related to using ginkgo/unit tests for end-to-end testing ?

nri/nri.go
```diff
@@ -0,0 +1,494 @@
...
+func (l *local) setState(id string, state State) {
+	if state != Removed {
+		l.state[id] = state
+		return
+	}
+
+	delete(l.state, id)
+}
+
+func (l *local) getState(id string) State {
```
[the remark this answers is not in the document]
klihub: Moved after the last caller.

internal/nri/domain.go
```diff
@@ -0,0 +1,109 @@
...
+	"sync"
+
+	"github.com/cri-o/cri-o/internal/log"
+	"github.com/sirupsen/logrus"
+
+	nri "github.com/containerd/nri/pkg/adaptation"
+)
+
+// Domain implements the functions the generic NRI interface needs to
+// deal with pods and containers from a particular containerd namespace.
```
mrunalp, line 15: comment needs updating. We don't have containerd namespaces :)

internal/nri/domain.go
```diff
@@ -0,0 +1,109 @@
...
+	GetContainer(string) (Container, bool)
+
+	// UpdateContainer applies an NRI container update request in the namespace.
+	UpdateContainer(context.Context, *nri.ContainerUpdate) error
+
+	// EvictContainer evicts the requested container in the namespace.
+	EvictContainer(context.Context, *nri.ContainerEviction) error
+}
+
+// SetDomain registers an NRI domain for a containerd namespace.
```
mrunalp, line 39: same here

internal/nri/domain.go
```diff
@@ -0,0 +1,109 @@
...
+	GetPodSandbox(string) (PodSandbox, bool)
+
+	// GetContainer returns the container for the given ID.
+	GetContainer(string) (Container, bool)
+
+	// UpdateContainer applies an NRI container update request in the namespace.
+	UpdateContainer(context.Context, *nri.ContainerUpdate) error
+
+	// EvictContainer evicts the requested container in the namespace.
+	EvictContainer(context.Context, *nri.ContainerEviction) error
```
mrunalp, line 36: What's the use case for eviction? Eviction is handled by the kubelet today.

mrunalp:
Made a quick pass and left a couple of nits and questions that are non-blocking. Thanks :+1:
Will let @haircommander do the final review and merge.

internal/nri/domain.go
klihub: Ouch... thanks! Updated comments.

internal/nri/domain.go
klihub: Update comments.

nri/domain.go
klihub: Updated those comments.

server/container_remove.go
klihub: There is one code path where the control flow is NRI ingress -> Server.stopContainer() and that prevents me from simply folding unconditionally the NRI one into the corresponding CRI one like that.

test/nri/nri_test.go
klihub: Okay, after trying a few things, I ended up simply making ginkgo skipping the NRI test client.

internal/nri/domain.go
klihub: Hmm... eviction is probably a misnomer at the NRI level wrt. what this really does. Nothing is being evicted/removed here. EvictContainer stops the container. There are cases where an NRI plugin might need to stop a best-effort container and this interface is for that.

server/nri-api.go
```diff
@@ -0,0 +1,741 @@
...
+		criPod = a.cri.ContainerServer.GetSandbox(sandboxID)
+		if criPod == nil {
+			log.Errorf(ctx, "Failed to stop CRI container %q: can't find pod %q",
+				ctr.GetID(), sandboxID)
+			return nil
+		}
+	}
+
+	pod := nriPodSandbox(criPod)
+	err := a.nri.StopContainer(ctx, pod, ctr)
```
haircommander, line 298: nit, this can just be `return a.nri.Stop...`

server/nri-api.go
```diff
@@ -0,0 +1,741 @@
...
+	}
+
+	criPod := a.cri.getSandbox(ctx, criCtr.Sandbox())
+	pod := nriPodSandbox(criPod)
+	ctr := &criContainer{
+		api: a,
+		ctr: criCtr,
+	}
+
+	err := a.nri.PostUpdateContainer(ctx, pod, ctr)
```
haircommander, line 267: same here, can you go through and compress the lines where we do `err := ...; return err` and just do `return ...`

server/nri-api.go
```diff
@@ -0,0 +1,741 @@
...
+		spec: specgen.Config,
+	}
+
+	err := a.nri.StopContainer(ctx, pod, ctr)
+	if err != nil {
+		log.Errorf(ctx, "Failed to creation undo (stop) failed: %v", err)
+	}
+
+	err = a.nri.RemoveContainer(ctx, pod, ctr)
+	if err != nil {
```
haircommander, line 337: nit: compress these lines

server/nri-api.go
```diff
@@ -0,0 +1,741 @@
...
+	err = a.cri.stopContainer(ctx, ctr, 0)
+	if err != nil {
+		log.Errorf(ctx, "Failed to evict CRI container %q: %v", e.ContainerId, err)
+		return err
+	}
+
+	return nil
+}
+
+func todoMissingContext() context.Context {
```
haircommander, line 456: why don't we use context.TODO() instead of this?

server/nri-api.go
```diff
@@ -0,0 +1,741 @@
...
+		return nil
+	}
+	return api.FromOCILinuxNamespaces(p.spec.Linux.Namespaces)
+}
+
+func (p *criPodSandbox) GetPodLinuxOverhead() *api.LinuxResources {
+	if p.Sandbox == nil {
+		return nil
+	}
+	// TODO(klihub): AFAICT, CRI PodSandboxConfig is not stored, so we can only
```
haircommander, line 567: is this TODO expected to not be resolved? can we open an issue for it if this merges?

klihub: Yes, I don't expect those to be resolved now, that's why I left the TODOs in place. I can open an issue for this post-merge (and can also take a shot at implementing the missing bits).

klihub:
@kad PTAL

Makefile
```diff
@@ -92,6 +92,8 @@ GOPKGDIR := $(GOPATH)/src/$(PROJECT)
 GOPKGBASEDIR := $(shell dirname "$(GOPKGDIR)")
 GO_FILES := $(shell find . -type f -name '*.go' -not -name '*_test.go')
 
+GINKGO_SKIP_PACKAGES = test/nri
```
kad, line 95: might be deserving comment, why those are skipped? e.g. to mark that those are e2e style tests, not unit tests?

internal/factory/container/container.go
```diff
@@ -27,7 +27,7 @@ import (
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
-	"github.com/opencontainers/runtime-tools/validate"
+	validate "github.com/opencontainers/runtime-tools/validate/capabilities"
```
kad, line 30: this is expected or some artifact from rebases?

kad (commented):
LGTM, with few nits comments for tests.

Also, not sure about `internal/factory/container/container.go`, missing context of that change.

server/nri-api.go
klihub: Good point. Changed to that.

server/nri-api.go
klihub: Compressed to a single if-assign-test.

Makefile
klihub: Added a comment explaining the reason for skipping.

internal/factory/container/container.go
klihub: In runtime-tools we shuffled some functions around a bit a while ago (github.com/xeipuuv/gojson* isolation). Unfortunately in addition to properly bridging to the relocated functions, the original ones were also marked as deprecated. Our cri-o configuration asks golangci-lint to complain about using deprecated functions. So I had to change this to use the new location.

server/container_create_linux.go
```diff
@@ -816,6 +816,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		makeOCIConfigurationRootless(specgen)
 	}
 
+	err = s.nri.createContainer(ctx, specgen, sb, ociContainer)
+	if err != nil {
+		return nil, err
```
QiWang19, line 821: could you also compress it here?

server/container_start.go
```diff
@@ -97,11 +105,16 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
...
+		log.Warnf(ctx, "NRI post-start failed for container %q: %v", c.ID(), err)
+	}
+
 	log.WithFields(ctx, map[string]interface{}{
 		"description": c.Description(),
 		"containerID": c.ID(),
 		"sandboxID":   sandbox.ID(),
 		"PID":         state.Pid,
 	}).Infof("Started container")
+
```
QiWang19, line 118: remove the empty line

klihub: Updated.

go.mod
```diff
@@ -53,6 +54,7 @@ require (
 	github.com/opencontainers/runtime-spec v1.0.3-0.20220825212826-86290f6a00fb
 	github.com/opencontainers/runtime-tools v0.9.1-0.20221014010322-58c91d646d86
 	github.com/opencontainers/selinux v1.10.2
+	github.com/pkg/errors v0.9.1
```
kolyshkin, line 57: nit: this is something that's obsoleted since Go 1.13 (when native %w-style error wrapping was introduced). If possible, please do not add it back.

completions/bash/crio
```diff
@@ -76,6 +77,9 @@ h
 --minimum-mappable-uid
 --namespaces-dir
 --no-pivot
+--nri-config-file
+--nri-plugin-dir
+--nri-socket-path
```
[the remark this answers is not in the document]
klihub: Change command line option to `--nri-listen`.

go.mod
klihub: Argh, sorry and thanks for spotting that. I though I got rid of all of them but missed that one. Removed.

internal/nri/config.go
```diff
@@ -0,0 +1,38 @@
...
+	nri "github.com/containerd/nri/pkg/adaptation"
+)
+
+// Config data for NRI.
+type Config struct {
+	Disable    bool   `toml:"disable"`
+	ConfigPath string `toml:"config_file"`
+	SocketPath string `toml:"socket_path"`
+	PluginPath string `toml:"plugin_path"`
+}
```
kolyshkin, line 13: This mostly duplicates internal/config/nri/config. Perhaps you can reuse that one by embedding it here (or vice versa).

internal/nri/sandbox_linux.go
```diff
@@ -0,0 +1,38 @@
+//go:build linux
+// +build linux
+
+/*
+   Copyright The containerd Authors.
```
kolyshkin, line 5: Is this a copy-paste from containerd code?

internal/nri/container.go
```diff
@@ -0,0 +1,58 @@
+package nri
+
+import (
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+
+	nri "github.com/containerd/nri/pkg/adaptation"
+)
+
+// Container interface for interacting with NRI.
+type Container interface {
```
kolyshkin, line 10: I guess what I don't understand is why these interfaces are introduced here. I get it they are internal to cri-o.

Is there going to be more than 1 implementation? Or, in general, what is the reason to have these interfaces?

server/container_create_linux.go
```diff
@@ -816,6 +816,16 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		makeOCIConfigurationRootless(specgen)
 	}
 
+	if err := s.nri.createContainer(ctx, specgen, sb, ociContainer); err != nil {
+		return nil, err
```
kolyshkin, line 820: I guess I don't understand why for some nri calls (such as postCreateContainer) you choose to warn, but for some others (like this one) you choose to error out.

server/container_update_resources.go
```diff
@@ -27,13 +27,21 @@ func (s *Server) UpdateContainerResources(ctx context.Context, req *types.Update
 	}
 
 	if req.Linux != nil {
-		resources := toOCIResources(req.Linux)
+		updated, err := s.nri.updateContainer(ctx, c, req.Linux)
+		if err != nil {
+			return nil, err
```
kolyshkin, line 32: Same here. For most calls you choose to just warn, but in here you fail. Why is that?

server/container_update_resources.go
```diff
@@ -27,13 +27,21 @@ func (s *Server) UpdateContainerResources(ctx context.Context, req *types.Update
...
+		resources := toOCIResources(updated)
 		if err := s.Runtime().UpdateContainer(ctx, c, resources); err != nil {
 			return nil, err
 		}
 
 		// update memory store with updated resources
 		s.UpdateContainerLinuxResources(c, resources)
+
+		if err := s.nri.postUpdateContainer(ctx, c); err != nil {
+			log.Errorf(ctx, "NRI container post-update failed: %v", err)
```
kolyshkin, line 43: Again, there's an inconsistency here -- earlier it's mostly either `log.Warnf` or returning an error, in here you logging (but not returning) an error.

server/sandbox_run_linux.go
```diff
@@ -684,6 +684,7 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 	if err := s.addSandbox(ctx, sb); err != nil {
 		return nil, err
 	}
+
```
kolyshkin, line 687: nit: unrelated hunk, please remove

test/nri/nri_suite_test.go
```diff
@@ -0,0 +1,260 @@
...
+	"testing"
+
+	"github.com/containerd/nri/pkg/api"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+)
+
+var crio *runtime
+
+func Setup() {
```
kolyshkin, line 17: nit: does this have to be public?

test/nri/nri_suite_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+	err = r.PullImages()
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to pull test images")
+	}
+
+	crio = r
+}
+
+func Cleanup() {
```
kolyshkin, line 35: ditto

test/nri/nri_suite_test.go
```diff
@@ -0,0 +1,260 @@
...
+type testLogger struct {
+	Out io.Writer
+}
+
+func (t *testLogger) Write(p []byte) (n int, err error) {
+	fmt.Printf("%s", string(p))
+	return len(p), nil
+}
+
+func SetupLogging() {
```
kolyshkin, line 72: ditto

test/nri.bats
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env bats
+
+#
+# This file is used to hook each test from that binary into
+# BATS as a separate test case for more granular test result
+# reporting. It needs to be updated whenever test cases are
+# added to or removed from the test binary.
```
kolyshkin, line 7: Makes sense to be more explicit that we mean `NRITEST_BINARY`.

Ideally, there has to be some assertion to make sure that the list of tests is consistent between this file and nri.test.

.golangci.yml
```diff
@@ -228,3 +228,10 @@ linters-settings:
     min-complexity: 154
   nakedret:
     max-func-lines: 15
+
+issues:
+  exclude-rules:
+    - path: test/nri/nri_test.go
+      linters:
+        - dupl
+      text: "lines are duplicate of"
```
kolyshkin, line 237: Perhaps you can work around these (by moving some code into common functions or methods) instead of disabling it.

internal/nri/sandbox_linux.go
klihub: Well, I first added a (structurally very similar) implementation for containerd, then bootstrapped this one from the same code... and forgot to nuke the headers from files which didn't need any touching. I'd rather say I wrote both at the same time... Thanks for spotting it. Removed the header.

internal/nri/container.go
klihub: I tried to keep the containerd NRI integration and the crio one structurally as close to each other as possible. One reason was to lower the cost of context-switching between them and ease maintenance. But another reason was that I still hope that once the dust settles, I'd be able to eventually further generalize these bits enough to move them over to the NRI repo as a package, and then consume it from there in both runtimes. I just couldn't give that a serious try at this phase.

The bits here are mostly agnostic to the internal runtime-specific implementation/representation of pods and containers, but need to be able to introspect and manipulate these, to interact with and interface NRI plugins. The interfaces here are for that. The rest of the patch set uses this internal package to hook NRI into CRI request processing.  

On the containerd side it is even possible to have multiple sandbox and container implementations but support for that is exactly what I gutted out from the code here in cri-o.

server/container_create_linux.go
klihub: So the general theme is that
- errors in NRI functions which could have resulted in changes to containers cause the triggering CRI call to error out
- errors in calls to NRI functions which can't alter container state are logged (with a warning or error), but do not error out from the triggering CRI call

Potentially state-altering calls are create, update and stop. Because stop is semantically a post-stop, there we log errors but don't cause the CRI call to error out.

server/container_update_resources.go
klihub: It's a potentially state-altering call. We treat failure in those as a failure in the triggering CRI call, propagating the error all the way back.

server/container_update_resources.go
klihub: It's a post-update NRI call which can't alter state (inside the runtime). Therefore it does not cause the update to return with an error.

Now to be honest, I'm not 100% sure that the current approach (treat only the errors in potentially state-altering calls as an error of the original CRI call)  is the right one, or if it is always the right one. IMO, we haven't had/seen enough use cases where an NRI plugin would carry out extra-OCI actions (as opposed or in addition to altering container configuration via the NRI API) to be able to really say. We'll need to figure this out as we gain more experience ourselves and feedback from others.

test/nri.bats
klihub: True, I guess you mean 's/from that binary/from NRITEST_BINARY/'

Yes, about the consistency... initially I was generating the NRI bats test using a template and the list of NRI test cases from the client (-test.list Test). Actually the above comment mistake is the result of my obviously sloppy updating of the comment in what originally was the bats template. Anyway, the dynamic generation was disliked in earlier review comments, so I switched to this static one. What I haven't addressed yet for that review comment is adding a check that all listed tests are accounted for. But I'll do that now...

.golangci.yml
klihub: Those were 'false' positives, due to structural similarity of some of the test cases. But I'll take another look to see if I can extract some of the common bits and get rid of the duplication error, by somehow 'hyperparametrizing' the offending tests and folding them into one test function.

test/nri/nri_suite_test.go
klihub: No, it' doesn't. Unexported.

go.mod
```diff
@@ -419,6 +420,7 @@ replace (
 	// TODO: Remove me when Podman uses a pinned runc version:
 	// https://github.com/containers/podman/blob/74aa681e59352257eb5f25f089dffa66d6345a3a/go.mod#L78
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.1-0.20220617142545-8b9452f75cbc
+	github.com/opencontainers/runtime-tools => github.com/opencontainers/runtime-tools v0.0.0-20221026201742-946c877fa809
```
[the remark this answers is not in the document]
klihub: No, it's not you missing something. It's me missing that since we don't need to divert any more the NRI repo to a forked one, there is also no point in using a related replace for runtime-tools, as instead of making things clearer it just makes them more confusing. Removed replace and updated runtime-tools version in require.

internal/config/nri/nri.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Validate loads and validates the effective runtime NRI configuration.
+func (c *Config) Validate(onExecution bool) error {
+	var (
+		cfg *nri.Config
+		err error
+	)
+
+	if c.Enabled {
+		if cfg, err = nri.ReadConfig(c.ConfigPath); err != nil {
```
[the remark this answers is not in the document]
klihub: Updated.

internal/config/nri/nri.go
```diff
@@ -0,0 +1,50 @@
...
+	ConfigPath string `toml:"nri_config_file"`
+	SocketPath string `toml:"nri_listen"`
+	PluginPath string `toml:"nri_plugin_dir"`
+
+	config *nri.Config
+}
+
+// New returns the default CRI-O NRI configuration.
+func New() Config {
+	return Config{
```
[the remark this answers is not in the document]
klihub: Updated to return a pointer.

internal/nri/config.go
klihub: Simplified, eliminating internal/nri.Config altogether and switching to use directly internal/config/nri.Config instead.

test/nri/nri_suite_test.go
klihub: Updated.

test/nri/nri_suite_test.go
klihub: Updated.

test/nri.bats
klihub: Updated. Added Makefile rule and github action for checking test case consitency in test client vs. NRI bats test cases.

.golangci.yml
klihub: Okay, it's not going to be a beauty-contest winner, but I managed to fold both offending pairs into a parametrized single function and get rid of both false positive duplicate errors and the need for duplicate error suppression.

haircommander:
/lgtm
/approve

well done @klihub and thanks everyone for reviewing!

(@klihub I've added a section at the bottom to add a release note, can you fill it in please?)

PR #5324, opened 2021-09-17
small improvement

utam0k:
#### What type of PR is this?

/kind cleanup

reduce unnecessary assignments to err.

None

None

```release-note
none
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package server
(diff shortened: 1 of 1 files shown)


server/container_create_linux.go
```diff
@@ -751,7 +752,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
 	// compatible with Docker
 	mtab := filepath.Join(mountPoint, "/etc/mtab")
-	if err := idtools.MkdirAllAs(filepath.Dir(mtab), 0755, rootPair.UID, rootPair.GID); err != nil {
+	if idtools.MkdirAllAs(filepath.Dir(mtab), 0o755, rootPair.UID, rootPair.GID) != nil {
 		return nil, errors.Wrap(err, "error creating mtab directory")
```
haircommander, line 756: if we make this change then the `err` here will be some outdated (and likely nil) value

haircommander:
thanks for the PR @utam0k but I don't think we want this

utam0k:
> thanks for the PR @utam0k but I don't think we want this

Thanks for your review. I'm still in the process of learning Go, and I didn't learn enough.

PR #5386, opened 2021-10-08
Skip volume relabel for super privileged containers

saschagrunert:
#### What type of PR is this?

/kind feature

If a container or pod specifies the SELinux type `spc_t`, then we skip the volume relabel.

None

cc @mrunalp 

```release-note
Skip SELinux volume relabeling for super privileged containers (`securityContext.seLinuxOptions.type = "spc_t"`).
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -163,6 +163,6 @@ linters-settings:
       - unnamedResult
       - unnecessaryBlock
   gocyclo:
-    min-complexity: 122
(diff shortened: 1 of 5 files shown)


haircommander:
thanks! I am wondering if we need the allowed_annotation here. If the container can access the volume regardless of it being labeled, then the behavior won't change whether we allow it or not :thinking:

PR #5414, opened 2021-10-19
oci: set conmon cpuset to infra_ctr_cpuset

haircommander:
#### What type of PR is this?

/kind feature

```release-note
Now, if infra_ctr_cpuset is configured, conmon is put in the same cpuset
```

### DIFF
--- a/internal/oci/oci_linux.go
+++ b/internal/oci/oci_linux.go
@@ -5,8 +5,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/cri-o/cri-o/utils"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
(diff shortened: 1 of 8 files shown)


internal/oci/oci_linux.go
```diff
@@ -25,6 +27,13 @@ func (r *runtimeOCI) createContainerPlatform(c *Container, cgroupParent string,
 		},
 	}
 
+	// First, set the cpuset as the one for the infra container.
+	// This should be overridden if specified in a workload.
+	if r.config.InfraCtrCPUSet != "" && r.config.ConmonCgroup == utils.PodCgroupName {
```
mrunalp, line 32: Why we need the second check?

haircommander: updated with a comment--it's possible a user has configured the cpuset of the conmon cgroup

cynepco3hahue:
/lgtm
It is important to say that it is only part of the solution, we still have a small window when the conmon process runs on top of all CPUs, and it should be fixed once we will have clone3 functionality under the OS.

saschagrunert (approved):
LGTM

mrunalp (approved):
/lgtm

PR #5423, opened 2021-10-25
Bug 2012838: fix override storage options from storage.conf

QiWang19:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=2012838

Merge in the storage_option from drop-in crio.conf to the storage.conf, do not override the configs from storage.conf.
The override leads to configuration of containerruntimeconfig not working.

#### What type of PR is this?
/kind bug

https://bugzilla.redhat.com/show_bug.cgi?id=2012838
The storage_option from /etc/crio/crio.conf.d/00-default overrides the configurations from /etc/containers/storage.conf.

```release-note
Merge storage_option from drop-in files to sttorage_option from storage.conf
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -573,6 +573,10 @@ func (c *Config) UpdateFromFile(path string) error {
 // Returns errors encountered when reading or parsing the files, or nil
 // otherwise.
 func (c *Config) UpdateFromDropInFile(path string) error {
+	// keeps the storage options from storage.conf and merge it to crio config
+	var storageOpts []string
+	storageOpts = append(storageOpts, c.StorageOptions...)
+
 	data, err := ioutil.ReadFile(path)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


QiWang19:
@cri-o/cri-o-maintainers PTAL

haircommander:
one issue with this approach is when we run
`crio -c crio.conf config > crio2.conf && mv crio2.conf crio.conf` then I believe we would get an ever growing list. we may want to consider option deduplication

haircommander:
(by that I don't mean checking each key, but each key=value option pair)
another thing to think about is what is the correct behavior here:
`["option=value1", "option=value2", "option=value1"]` 

my intuition would be the resulting list should be
`["option=value2", "option=value1"]`

QiWang19:
Removed the duplicates and keeps the last appearance in the list.

QiWang19:
@cri-o/cri-o-maintainers PTAL

utils/utils.go
```diff
@@ -373,3 +373,21 @@ func Sync(path string) error {
 	}
 	return nil
 }
+
+// RemoveDupStorageOpts removes duplicated storage option from the list
```
haircommander, line 377: nit: I'm personally not a fan of putting functions in a utils package that isn't required in multiple packages. To me, utils should be used as a place to prevent cyclical dependencies. This function seems better suited for the config package IMO

haircommander:
:sunglasses: 
excellent, LGTM

ping @saschagrunert for the final tag

saschagrunert (approved):
/lgtm

PR #5427, opened 2021-10-26
stats: modernize stats collection and implement Pod level stats reporting

haircommander:
#### What type of PR is this?

/kind feature
CRI stats has been in sore need of an update, and is targeted to be updated in [1.23](https://github.com/kubernetes/enhancements/issues/2371). Part of this update is support in the main CRI implementations.

This PR is partially a refactor and partially a feature. In includes cleaning up some code that wasn't actually used and was thus wasteful, bumping CRI to get the new pod level stats calls, adding a StatsServer to give the ability to cache the stats on a periodic basis (rather than collect on-demand, as is default). 

There still should be some test updates, but integration tests will be hard before critools adds support for the new fields and cuts a release. In the meantime, lots of this is reviewable.

```release-note
Add a [crio.stats] table to the config for configuring all things stats-related.
Add ability (with config field StatsCollectionPeriod) to calculate and save stats for pods and containers, rather than return them on demand.
Add support for {,List}PodSandboxStats, which allows the kubelet to take advantage of the PodAndContainerStatsFromCRI feature gate--one that is set to be alpha in 1.23.
```


internal/config/cgmgr/stats.go
```diff
@@ -0,0 +1,174 @@
...
+		return err
+	}
+	systemNano := time.Now().UnixNano()
+	stats.CPU = createCPUStats(systemNano, cgroupStats)
+	stats.Memory, err = createMemoryStats(systemNano, cgroupStats, cgroupPath)
+	stats.Process = createProcessUsage(systemNano, cgroupStats)
+	if err != nil {
+		return err
+	}
+	return nil
```
saschagrunert, line 31:
```suggestion
	return err
```

internal/config/cgmgr/stats.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return err
+	}
+	systemNano := time.Now().UnixNano()
+	stats.CPU = createCPUStats(systemNano, cgroupStats)
+	stats.Memory, err = createMemoryStats(systemNano, cgroupStats, cgroupPath)
+	if err != nil {
+		return err
+	}
+	return nil
```
saschagrunert, line 46:
```suggestion
	return err
```

internal/config/cgmgr/stats.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return err
+	}
+	if usage > totalInactive {
+		memory.WorkingSetBytes.Value = usage - totalInactive
+	} else {
+		logrus.Warnf(
+			"unable to account working set stats: total_inactive_file (%d) > memory usage (%d)",
```
saschagrunert, line 161: nit
```suggestion
			"Unable to account working set stats: total_inactive_file (%d) > memory usage (%d)",
```

saschagrunert (approved):
Just a few nits, otherwise LGTM

fidencio:
@haircommander, can you hold this one till Tomorrow?  I'd like to do a review (together with some local tests) on this later Today / Tomorrow morning.

internal/oci/runtime_vm.go
```diff
@@ -812,31 +803,21 @@ func metricsToCtrStats(ctx context.Context, c *Container, m *cgroups.Metrics) *C
...
-			switch strings.ToLower(entry.Op) {
-			case "read":
-				blockInput += entry.Value
-			case "write":
-				blockOutput += entry.Value
-			}
-		}
+		rssBytes = m.Memory.RSS
+		pageFaults = m.Memory.PgFault
+		majorPageFaults = m.Memory.PgMajFault
```
fidencio, line 811: For kata-containers, this is the part that differs from the approach taken before and that we should ensure the CRI runtime has access to the correct information.

Those are collect, and then passed up to the CRI runtime, here:
https://github.com/kata-containers/kata-containers/blob/e89c06e68bcd465cc79dc68945751a706f230bfe/src/runtime/pkg/containerd-shim-v2/metrics.go#L98-L146

It means that, theoretically, from the kata-containers everything needed here is already provided by kata-containers, but I still need to get to do some more tests on this.

internal/oci/runtime_vm.go
```diff
@@ -812,31 +803,21 @@ func metricsToCtrStats(ctx context.Context, c *Container, m *cgroups.Metrics) *C
...
-		BlockInput:      blockInput,
-		BlockOutput:     blockOutput,
 		Container:       c.ID(),
-		CPU:             cpu,
 		CPUNano:         cpuNano,
-		MemLimit:        memLimit,
 		MemUsage:        memUsage,
-		MemPerc:         memPerc,
-		NetInput:        netInput,
-		NetOutput:       netOutput,
```
fidencio, line 836: @haircommander, I feel like our future selves would appreciate a comment, probably in the commit message, explaining why CPU*, Net*, and Block* stats coming from the containers are not relevant anymore.

server/cri/v1/rpc_list_pod_sandbox_stats.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+
+	"github.com/cri-o/cri-o/server/cri/types"
+	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
+)
+
+func (s *service) ListPodSandboxStats(
+	ctx context.Context, req *pb.ListPodSandboxStatsRequest,
+) (*pb.ListPodSandboxStatsResponse, error) {
+	r := &types.ListPodSandboxStatsRequest{}
```
kolyshkin, line 13: Seems this should belong to the next commit (the third one in the series, named "stub {List}PodSandboxStats" or so).

kolyshkin: or maybe not... I'm not quite following the logic

internal/oci/oci_linux.go
```diff
@@ -1,10 +1,10 @@
+//go:build linux
```
kolyshkin, line 1: It would be nice to have a separate PR that adds go:build tags for all files at once, to minimize the noise introduced by it.

Also, in this particular case, the `linux` build tag is not needed as the file ends with `_linux.go`

So, something similar to https://github.com/opencontainers/runc/pull/3185 would be nice to have.

internal/oci/oci_linux.go
```diff
@@ -95,70 +104,74 @@ func (r *runtimeOCI) containerStats(ctr *Container, cgroup string) (*ContainerSt
...
 			val, err := strconv.Atoi(
-				strings.TrimPrefix(scanner.Text(), varPrefix),
+				strings.TrimPrefix(scanner.Text(), field.prefix),
 			)
 			if err != nil {
-				return 0, errors.Wrap(err, "unable to parse total inactive file value")
+				return errors.Wrapf(err, "unable to parse %s", field.prefix)
 			}
-			return uint64(val), nil
+			valUint := uint64(val)
```
kolyshkin: Maybe use `strconv.ParseUint64` so we don't have to convert.

internal/config/cgmgr/cgmgr.go
```diff
@@ -54,10 +56,17 @@ type CgroupManager interface {
 	// returns the cgroup path on disk for that containerID. If parentCgroup is empty, it
 	// uses the default parent for that particular manager
 	ContainerCgroupAbsolutePath(string, string) (string, error)
+	// PopulateContainerCgroupStats takes arguments sandbox parent cgroup, container ID, and
+	// containers stats object. It fills the object with information from the cgroup found
+	// given that parent and ID
+	PopulateContainerCgroupStats(string, string, *types.ContainerStats) error
```
kolyshkin, line 62: You can make it more self-documenting here by adding parameter names, e.g.
```go
PopulateContainerCgroupStats(sbParent, cid string, stats *types.ContainerStats) error
```

internal/config/cgmgr/cgmgr_test.go
```diff
@@ -17,7 +17,7 @@ const (
 )
 
 // The actual test suite
-var _ = t.Describe("Config", func() {
+var _ = t.Describe("Cgmgr", func() {
```
kolyshkin, line 20: is this hunk related?

pkg/config/template.go
```diff
@@ -80,6 +83,7 @@ const (
 	crioNetworkConfig templateGroup = 5
 	crioMetricsConfig templateGroup = 6
 	crioTracingConfig templateGroup = 7
+	crioStatsConfig   templateGroup = 8
```
kolyshkin, line 86: Can simplify this to use iota. https://github.com/cri-o/cri-o/pull/5469

internal/lib/stats/stats_server.go
```diff
@@ -0,0 +1,339 @@
...
+	current.UsageNanoCores = &types.UInt64Value{
+		Value: usageNanoCores,
+	}
+}
+
+// populateWritableLayer attempts to populate the container stats writable layer.
+func (ss *StatsServer) populateWritableLayer(stats *types.ContainerStats, container *oci.Container) {
+	writableLayer, err := ss.writableLayerForContainer(container)
+	if err != nil {
+		logrus.Errorf("%v", err)
```
kolyshkin, line 173: nit: `logrus.Error(err)`.

internal/lib/stats/stats_server.go
```diff
@@ -0,0 +1,339 @@
...
+// writableLayerForContainer gathers information about the container's writable layer.
+// It does so by calling into the GraphDriver's endpoint to get the UsedBytes and InodesUsed.
+func (ss *StatsServer) writableLayerForContainer(container *oci.Container) (*types.FilesystemUsage, error) {
+	writableLayer := &types.FilesystemUsage{
+		Timestamp: time.Now().UnixNano(),
+		FsID:      &types.FilesystemIdentifier{Mountpoint: container.MountPoint()},
+	}
+	driver, err := ss.Store().GraphDriver()
+	if err != nil {
+		return writableLayer, errors.Wrapf(err, "Unable to get graph driver for disk usage for container %s", container.ID())
```
kolyshkin, line 187: Not for this PR, but in general: at some point we should ditch pkg/errors and switch to go native error wrapping (i.e. `fmt.Errorf(".... %w", err)`.

internal/lib/container_server.go
```diff
@@ -135,7 +137,9 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
 			processLevels:   make(map[string]int),
 		},
 		config: config,
-	}, nil
+	}
+	c.StatsServer = statsserver.New(c)
+	return c, nil
```
kolyshkin, line 142: Why not 
```go
return &ContainerServer{
...
	StatsServer: statsserver.New(c)
...
}
```

kolyshkin: Ah, because `c`. Nevermind :)

server/sandbox_list.go
```diff
@@ -8,72 +8,91 @@ import (
...
+			}
+		}
+		if filter.LabelSelector != nil {
+			sel := fields.SelectorFromSet(filter.LabelSelector)
+			if !sel.Matches(fields.Set(p.Labels)) {
+				return false
+			}
+		}
+	}
+	return true
```
kolyshkin, line 97: Why was this moved? Asking because it makes the review harder (and also git log / git blame).

kolyshkin:
LGTM overall. It's kind of hard to review because of so many code moved around, so I might have missed something.

kolyshkin (approved):
LGTM

internal/oci/oci_linux.go
haircommander: hm this was added by vim-go plugin.. I agree we should probably do it all in one fell swoop, though I may push for this to stay if the plugin continues to add it :sweat_smile:

internal/oci/runtime_vm.go
haircommander: the thing is they have never been relevant. I think they were added because they did come from the cgroup package we use but it does feel odd to leave a comment about code that was dead and wasteful

internal/config/cgmgr/cgmgr_test.go
haircommander: not technically related but also trivial and self evident so I didn't bother adding a different commit

pkg/config/template.go
haircommander: I've rebased on the pkg/config commit

server/sandbox_list.go
haircommander: it is a pet peeve of mine when functions are defined before they're called--preventing one reading the file top-down to understand the flow. I figured this one was a relatively small function and not too touch to check for differences

internal/lib/stats/stats_server.go
haircommander: agreed!

internal/config/cgmgr/cgmgr.go
haircommander: good idea, done

haircommander:
comments addressed, PTAL @cri-o/cri-o-maintainers

saschagrunert:
Unfortunately this needs a rebase.

haircommander:
rebased!

saschagrunert (approved):
/lgtm

PR #5433, opened 2021-11-01
oci: Fix a couple of deadlocks in container stop code

haircommander:
#### What type of PR is this?

/kind cleanup

```release-note
none
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -70,6 +70,7 @@ type Container struct {
 	stopping           bool
 	stopTimeoutChan    chan time.Duration
 	stoppedChan        chan struct{}
+	stopStoppingChan   chan struct{}
 	stopLock           sync.Mutex
 }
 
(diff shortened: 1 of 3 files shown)


haircommander:
for reference, we have a [goroutine stack](https://bugzilla.redhat.com/attachment.cgi?id=1838992) that has a ridiculous number of go routines stuck on [this line](https://github.com/cri-o/cri-o/blob/2e9a7119f763fb1f4e3eb0aeef15a7132a869602/internal/oci/oci.go#L144)

While this more seems like the go routine didn't receive correctly on the close of chControl, adding the buffer would also let the function make forward progress. i.e. I am suspecting something funky about the go runtime, but maybe we can work around it?

saschagrunert (approved):
LGTM

haircommander:
so I didn't actually know why this could work but I guessed it would. looking at https://github.com/cri-o/cri-o/blob/main/internal/oci/oci.go#L139-L146
```
			case <-chControl:
				return
			default:
				// Check if the container is stopped
				if err := impl.UpdateContainerStatus(ctx, c); err != nil {
					done <- err
					return
				}
				if c.State().Status == ContainerStateStopped {
					done <- nil
					return
				}
```
if chControl is closed while UpdateContainerStatus is running, we will wait forever on done, when we should check chControl. buffering the channel will allow us to make forward progress (without really complicating the code that much)

haircommander:
@cri-o/cri-o-maintainers PTAL

haircommander:
related to https://bugzilla.redhat.com/show_bug.cgi?id=2003206

Issue #4790, opened 2021-04-22
Support Target NamespaceMode 


haircommander:
We have managed to avoid processing Target NamespaceMode because there are no CRI tests nor k8s conformance tests that rely on it. However, we really should be supporting it (ephemeral containers are still alpha and seem to be the major consumer, but it will go beta one day)

This is a tracking issue for this work

haircommander:
it seems beta ephemeral containers [slipped 1.22](https://github.com/kubernetes/website/pull/27988) so I am going to push this to 1.23

haircommander:
also, to do this we'll need to revive https://github.com/cri-o/cri-o/pull/3910

haircommander:
I opened a PR to add support. @paulnivin do y'all have any interest in trying it out and making sure it works for you?

paulnivin:
@haircommander Sorry, I missed the updates here. We can give the code a spin. Thanks for adding

PR #5440, opened 2021-11-05
Add support for target namespaces

haircommander:
#### What type of PR is this?

/kind feature
Adds support for TARGET namespaces, which is needed by ephemeral containers (set to go beta in 1.23)

fixes https://github.com/cri-o/cri-o/issues/4790

```release-note
Add support for TARGET namespace mode, which enables support for ephemeral containers.
```


internal/config/nsmgr/nsmgr.go
```diff
@@ -180,8 +181,68 @@ func getSysctlForPinns(sysctls map[string]string) string {
...
+	// now create an empty file
+	f, err := os.Create(pinnedNamespace)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating namespace path")
+	}
+	f.Close()
+
+	defer func() {
+		if retErr != nil {
+			if err2 := os.Remove(pinnedNamespace); err != nil {
```
saschagrunert, line 201:
```suggestion
			if err := os.Remove(pinnedNamespace); err != nil {
```

rphillips:
I did a review pass... just Sascha's comment

test/test_runner.sh
```diff
@@ -26,7 +26,7 @@ function execute() {
 }
 
 # Tests to run. Defaults to all.
-TESTS=${*:-.}
+TESTS=${TESTS:-.}
```
littlejawa, line 29: This seems to be the cause of the error on ci/kata-jenkins : rather than running only the .bats file(s) given as parameter, the script is now expecting an environment variable to be set. Since our test script is not setting this variable, we default to running all bats files.

I can understand the change, but I'll need to adapt our test script. In the meantime, ci/kata-jenkins will fail systematically with this commit. I don't know if there are other jobs relying on it?

haircommander: honestly I didn't realize how this worked and thought it was just broken lol, I've reverted the change, thanks for the insight

saschagrunert (approved):
/retest-required

haircommander:
/hold
unit tests and kube e2e tests (in github actions) still fail, need to investigate those

haircommander:
@cri-o/cri-o-maintainers PTAL

pkg/container/namespaces.go
```diff
@@ -0,0 +1,95 @@
...
+		targetPID, err := targetCtr.Pid()
+		if err != nil {
+			return errors.Wrapf(err, "target PID namespace find PID")
+		}
+		ns, err := serverConfig.NamespaceManager().NamespaceFromProcEntry(targetPID, nsmgr.PIDNS)
+		if err != nil {
+			return errors.Wrapf(err, "target PID namespace namespace from proc")
+		}
+		if err := c.spec.AddOrReplaceLinuxNamespace(string(rspec.PIDNamespace), ns.Path()); err != nil {
+			return err
```
saschagrunert, line 59: Let's wrap this error for increasing verbosity.

pkg/container/namespaces.go
```diff
@@ -0,0 +1,95 @@
...
+			continue
+		}
+		nsForSpec := typeToSpec[ns.Type()]
+		if nsForSpec == "" {
+			return errors.Errorf("Invalid namespace type %s", ns.Type())
+		}
+		err := g.AddOrReplaceLinuxNamespace(string(nsForSpec), ns.Path())
+		if err != nil {
+			return err
+		}
```
saschagrunert, line 88:
```suggestion
		if err := g.AddOrReplaceLinuxNamespace(string(nsForSpec), ns.Path()); err != nil {
			return errors.Wrap(err, "add or replace linux namespace")
		}
```

saschagrunert (approved):
Just two nits, otherwise LGTM :+1:

server/sandbox_run_linux.go
```diff
@@ -353,12 +353,6 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 		return nil, errors.Wrapf(err, resourceErr.Error())
 	}
 
-	if sbox.Config().Linux == nil {
-		sbox.Config().Linux = &types.LinuxPodSandboxConfig{}
-	}
-	if sbox.Config().Linux.SecurityContext == nil {
-		sbox.Config().Linux.SecurityContext = newLinuxSandboxSecurityContext()
```
kolyshkin, line 360: The `newLinuxSandboxSecurityContext` is unused after this commit, and therefore should be removed (by this commit, not a later one).

(Also, I wish there was a way to run linter on every commit -- it would catch issues like this).

internal/config/nsmgr/nsmgr.go
```diff
@@ -187,6 +187,50 @@ func (mgr *NamespaceManager) dirForType(ns NSType) string {
...
+// This function is heavily based on containernetworking ns package found at:
+// https://github.com/containernetworking/plugins/blob/5c3c17164270150467498a32c71436c7cd5501be/pkg/ns/ns.go#L140
+// Credit goes to the CNI authors.
+func (mgr *NamespaceManager) NamespaceFromProcEntry(pid int, nsType NSType) (_ Namespace, retErr error) {
+	pinnedNamespace := uuid.New().String()
+
+	// now create an empty file
+	f, err := os.Create(pinnedNamespace)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating namespace path")
```
kolyshkin, line 195: At some point we should switch away from `golang.org/pkg/errors` into Go 1.13+ native error wrapping (i.e. `fmt.Errorf` with `%w` specifier).

internal/config/nsmgr/nsmgr.go
```diff
@@ -187,6 +187,50 @@ func (mgr *NamespaceManager) dirForType(ns NSType) string {
...
+	}
+	f.Close()
+
+	defer func() {
+		if retErr != nil {
+			if err := os.Remove(pinnedNamespace); err != nil {
+				logrus.Errorf("Failed to remove namespace path %s after failure to pin namespace: %v", pinnedNamespace, err)
```
kolyshkin, line 202: The error message from os.Remove already contains the file name, no need to have it twice.

internal/config/nsmgr/nsmgr.go
```diff
@@ -187,6 +187,50 @@ func (mgr *NamespaceManager) dirForType(ns NSType) string {
...
+// NamespaceFromProcEntry creates a new namespace object from a bind mount from a processes proc entry.
+// The caller is responsible for cleaning up the namespace by calling Namespace.Remove().
+// This function is heavily based on containernetworking ns package found at:
+// https://github.com/containernetworking/plugins/blob/5c3c17164270150467498a32c71436c7cd5501be/pkg/ns/ns.go#L140
+// Credit goes to the CNI authors.
+func (mgr *NamespaceManager) NamespaceFromProcEntry(pid int, nsType NSType) (_ Namespace, retErr error) {
+	pinnedNamespace := uuid.New().String()
+
+	// now create an empty file
+	f, err := os.Create(pinnedNamespace)
```
kolyshkin: Perhaps it's better to use `os.CreateTemp` instead -- it actually checks that the file being created has a unique name (rather than relying on UUID being unique), which seems more straightforward.

I'm also a bit concerned about _where_ the file is being created. I guess the code assumes we're in a particular directory -- maybe add a comment about it.

pkg/container/container.go
```diff
@@ -285,7 +288,7 @@ func (c *container) SetConfig(config *types.ContainerConfig, sboxConfig *types.P
 		return errors.New("sandbox config is already set")
 	}
 
-	c.config = config
+	c.config = cfg
```
kolyshkin, line 297: Nit: these hunks does not seem to be relevant.

internal/config/nsmgr/nsmgr.go
```diff
@@ -231,6 +225,12 @@ func (mgr *NamespaceManager) NamespaceFromProcEntry(pid int, nsType NSType) (_ N
 	return GetNamespace(pinnedNamespace, nsType)
 }
 
+// dirForType returns the sub-directory for that particular NSType
+// which is of the form `$namespaceDir/$nsType+"ns"`
+func (mgr *NamespaceManager) dirForType(ns NSType) string {
+	return filepath.Join(mgr.namespacesDir, string(ns)+"ns")
+}
+
```
kolyshkin: Ah! This should be part of the commit where you introduce `NamespaceFromProcEntry`.

test/namespaces.bats
```diff
@@ -32,3 +34,47 @@ function teardown() {
...
+
+	ctr_id=$(crictl run "$ctr2" "$pod2")
+
+	output1=$(crictl exec --sync "$target_ctr" ps | grep -v ps)
+	output2=$(crictl exec --sync "$ctr_id" ps | grep -v ps)
+	[[ "$output1" == "$output2" ]]
+
+	crictl rmp -fa
+	# make sure namespace is cleaned up
+	[[ -z $(ls "$namespaces_dir/pidns") ]]
```
kolyshkin, line 82: Perhaps it's more clear to check that the entry does not exist? Something like `[ ! -e "$namespaces_dir/pidns" ]`

.github/workflows/test.yml
```diff
@@ -246,7 +246,7 @@ jobs:
       - run: |
           make mockgen -j $(nproc)
           hack/tree_status.sh
-      - run: make testunit
+      - run: sudo PATH="$PATH" make testunit
```
kolyshkin, line 249: (commenting on the commit message, not the change itself)

Would be nice to explain why sudo is needed ("test case XXX added by an earlier commit performs mounts, which require sudo" or somesuch).

kolyshkin: Also, this probably belongs to the commit that added the test that needs sudo.

kolyshkin: > Also, this probably belongs to the commit that added the test that needs sudo.

That is, unless you have added a check that skips the test as non-root.

test/namespaces.bats
```diff
@@ -36,6 +36,9 @@ function teardown() {
 }
 
 @test "pid namespace mode target test" {
+	if [[ -z "$CONTAINER_UID_MAPPINGS" || $(id -u) == "0" ]]; then
+		skip "test fails in a user namespace"
+	fi
```
kolyshkin, line 41: The check seems to skip the test if there's NO userns enabled (as well as when the user is root), but the message suggests otherwise. I guess I'm missing something here 😕

kolyshkin:
It seems like a few last commits are fixups to earlier ones. Instead, they should be part of those earlier commits.

pkg/container/container.go
haircommander: we now import our config package, and we shadow the import unless we change the variable name

internal/config/nsmgr/nsmgr.go
haircommander: ooh great catch, it wasn't putting it in the right directory

haircommander: haha I guess I figured it out eventually

.github/workflows/test.yml
haircommander: I moved this commit before the other

mrunalp (approved):
/lgtm

PR #5469, opened 2021-11-19
Use iota

kolyshkin:
#### What type of PR is this?

/kind cleanup

Less hardcoded repeated numbers, more iotas!

None

```release-note
None
```

### DIFF
--- a/server/cri/types/types.go
+++ b/server/cri/types/types.go
@@ -6,22 +6,28 @@ const (
 	RuntimeReady = "RuntimeReady"
 	NetworkReady = "NetworkReady"
 
-	NamespaceModePOD       NamespaceMode = 0
-	NamespaceModeCONTAINER NamespaceMode = 1
-	NamespaceModeNODE      NamespaceMode = 2
-	NamespaceModeTARGET    NamespaceMode = 3
+	NamespaceModePOD NamespaceMode = iota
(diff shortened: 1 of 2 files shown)


kolyshkin:
Hmm, I am not sure why there are so many CI failures -- these changes are supposed to be non-disruptive, IOW they don't change anything but the way source code looks like.

server/cri/types/types.go
```diff
@@ -6,22 +6,28 @@ const (
...
-	MountPropagationPropagationBidirectional   MountPropagation = 2
-
-	ContainerStateContainerCreated ContainerState = 0
-	ContainerStateContainerRunning ContainerState = 1
-	ContainerStateContainerExited  ContainerState = 2
-	ContainerStateContainerUnknown ContainerState = 3
-
-	PodSandboxStateSandboxReady    PodSandboxState = 0
-	PodSandboxStateSandboxNotReady PodSandboxState = 1
+	NamespaceModePOD NamespaceMode = iota
```
haircommander, line 9: this PR is a good idea, but I think we should drop the types changes in favor of https://github.com/cri-o/cri-o/pull/5471

kolyshkin:
> this PR is a good idea, but I think we should drop the types changes in favor of #5471

Yes indeed! Thanks, closing.

PR #5472, opened 2021-11-22
Add Go 1.17+ go:build tags

kolyshkin:
#### What type of PR is this?

/kind cleanup

1. Remove redundant build tags
    
    The tags being removed are redundant because the file names already have
    them (e.g. specifying +build linux is redundant in foo_linux.go).

2. Add go 1.17+ go:build tags

    Go 1.17 introduced this new (and better) way to specify build tags.
    For more info, see https://golang.org/design/draft-gobuild.
    
    As a way to seamlessly switch from old to new build tags, gofmt (and
    gopls, and goimports) from go 1.17 adds the new tags along with the old
    ones.
    
    Later, when go < 1.17 is no longer supported, the old build tags
    can be removed.
    
    Now, as I started to use latest gopls (v0.7.1), it adds these tags
    while I edit. Rather than to randomly add new build tags in new PRs,
    I guess it is better to do it once for all files.
    
    Brought to you by goimports -w. It also did some other minor changes
    (whitespace fixes, removing unused imports), which looks OK to me.

None

```release-note
NONE
```

### DIFF
--- a/cmd/crio/daemon_linux.go
+++ b/cmd/crio/daemon_linux.go
@@ -1,5 +1,3 @@
-// +build linux
-
 package main
 
 import (
--- a/cmd/crio/daemon_unsupported.go
+++ b/cmd/crio/daemon_unsupported.go
@@ -1,3 +1,4 @@
+//go:build !linux
 // +build !linux
 
 package main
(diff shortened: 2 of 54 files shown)


saschagrunert (approved):
/lgtm

PR #5474, opened 2021-11-29
Fix runtime panic on stats server shutdown

saschagrunert:
#### What type of PR is this?

/kind bug

It may be possible that the stats servers `Shutdown()` get called twice
during CRI-O stop. To avoid calling `close()` on an already closed
channel, we now internally track if the stats server is still running or
not.

Introduced in https://github.com/cri-o/cri-o/pull/5427
None

```release-note
None
```


haircommander:
good catch, thanks :sweat_smile:

PR #5481, opened 2021-12-01
Allow swap to be configured if lower than memory on cgroupv2

haircommander:

/kind bug
```release-note
none
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -376,6 +376,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 						)
 					}
 					memoryLimit = resources.MemorySwapLimitInBytes
+					if node.CgroupIsV2() {
+						// Kubelet specifies these fields assuming cgroupv1. In cgroupv1,
+						// swap is memoryLimitInBytes + amount that can be used for swap.
+						// However, in cgroupv2, swap and memory are specified separately.
+						// Since kubelet assumes v1, it sends swap+memory as swap value,
+						// but the correct way to specify swap subtracts memory from that.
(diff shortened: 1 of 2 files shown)


server/container_create_linux.go
```diff
@@ -367,7 +367,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 				}
 				specgen.SetLinuxResourcesMemoryLimit(memoryLimit)
 				if resources.MemorySwapLimitInBytes != 0 {
-					if resources.MemorySwapLimitInBytes < resources.MemoryLimitInBytes {
+					if resources.MemorySwapLimitInBytes < resources.MemoryLimitInBytes && !node.CgroupIsV2() {
```
ehashman, line 370: This is not quite accurate from the spec, I think. I would expect the following logic:

- As written, this validation is correct.
- In the cgroups v1 case, set this as is currently specified on L378.
- In the cgroups v2 case, set swap to the difference between MemoryLimitInBytes and MemorySwapLimitInBytes. Since `MemoryLimitInBytes <= MemorySwapLimitInBytes`, `MemorySwapLimitInBytes - MemoryLimitInBytes >= 0`.

TomSweeneyRedHat:
@giuseppe WDYT?

giuseppe:
is it addressing any issue?

I think both runc and crun already have the same logic about swapMemory so we end up doing the same operation twice on cgroupv2

mrunalp:
As @giuseppe pointed out, the runtimes take care of this, so we don't need it.

See https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/utils.go#L433

PR #5487, opened 2021-12-04
metrics: introduce new metrics matching Prometheus best practices

### LINKED ISSUE #5426: Make metrics follow Prometheus best practices and reduce cardinality
The prometheus metric names don't follow the best practices:

- all counters should be suffixed by `_total`
- all metrics should use base units (so seconds, not microseconds)

Beside that, some label values are unbound, leading to high cardinality (at least `crio_image_pulls_by_*`).

Arguably some metric names could be improved (crio_image_pulls_by_digest should include bytes if it reflects bytes pulled).

Since changing the metrics is a breaking change for dashboards and alerts, ideally this happens only once and new metrics are carefully reviewed to follow the best practices.

/cc @bwplotka Maybe something for you? :)

swghosh:
#### What type of PR is this?

/kind other

* add metrics with new names that match naming best practices
  * use _total for all counters
  * use base unit seconds, bytes
* metrics that do not follow best practices have been marked deprecated,
  these can be removed in a future release, it is to ensure non-breaking change
  for couple of releases

Fixes #5426

New metrics:
`crio_operations_total, crio_operations_latency_seconds_total, crio_operations_latency_seconds,
crio_operations_errors_total, crio_image_pulls_by_digest_bytes_total,
crio_image_pulls_by_name_bytes_total, crio_image_pulls_by_name_skipped_bytes_total,
crio_image_pulls_successes_total, crio_image_pulls_failures_total,
crio_image_layer_reuse_total, crio_containers_oom_by_name_total`

```release-note
Introduce the following metrics:
crio_operations_total, crio_operations_latency_seconds_total, crio_operations_latency_seconds,
crio_operations_errors_total, crio_image_pulls_by_digest_bytes_total,
crio_image_pulls_by_name_bytes_total, crio_image_pulls_by_name_skipped_bytes_total,
crio_image_pulls_successes_total, crio_image_pulls_failures_total,
crio_image_layer_reuse_total, crio_containers_oom_by_name_total
while marking metric names that do not follow prometheus best practices as Deprecated in Prometheus metric help text.
```


tutorials/metrics.md
```diff
@@ -39,22 +39,34 @@ metrics_key = "/path/to/key.pem"
...
-| `crio_containers_oom`                            | `name`                                                                                                                                  | Counter   | Containers killed because they ran out of memory (OOM) by their name                                 |
-| `crio_processes_defunct`                         |                                                                                                                                         | Gauge     | Total number of defunct processes in the node                                                        |
+| Metric Key                                       | Possible Labels or Buckets                                                                                                              | Type      | Purpose                                                                                                           |
```
saschagrunert, line 42: @discordianfish may I ask you to give this table a look if it fits prometheus best practices?

tutorials/metrics.md
```diff
@@ -39,22 +39,34 @@ metrics_key = "/path/to/key.pem"
...
+| `crio_operations_total`                          | every CRI-O RPC\*                                                                                                                       | Counter   | Cumulative number of CRI-O operations by operation type.                                                          |
+| `crio_operations_latency_seconds_total`          | every CRI-O RPC\*,<br><br>`network_setup_pod` (CNI pod network setup time),<br><br>`network_setup_overall` (Overall network setup time) | Summary   | Latency in seconds of CRI-O operations. Split-up by operation type.                                               |
+| `crio_operations_latency_seconds`                | every CRI-O RPC\*                                                                                                                       | Gauge     | Latency in microseconds of individual CRI calls for CRI-O operations. Broken down by operation type.              |
```
discordianfish, line 60: It says microseconds in the description but should be seconds.
I'm also not sure what this is capturing. The latency of the last CRI call? That seems odd. The average? That would be besser calculated from crio_operations_latency_seconds_total.

tutorials/metrics.md
```diff
@@ -39,22 +39,34 @@ metrics_key = "/path/to/key.pem"
...
+| `crio_operations_errors_total`                   | every CRI-O RPC\*                                                                                                                       | Counter   | Cumulative number of CRI-O operation errors by operation type.                                                    |
+| `crio_image_pulls_by_digest_bytes_total`         | `name`, `digest`, `mediatype`, `size`                                                                                                   | Counter   | Bytes transferred by CRI-O image pulls by digest.                                                                 |
```
discordianfish, line 62: I'm a bit worried about the cardinality of these. Especially digest seems like something that could change very frequently and each dimension will stay exposed indefinitely.

tutorials/metrics.md
```diff
@@ -39,22 +39,34 @@ metrics_key = "/path/to/key.pem"
...
+| `crio_image_pulls_by_name_bytes_total`           | `name`, `size`                                                                                                                          | Counter   | Bytes transferred by CRI-O image pulls by name.                                                                   |
+| `crio_image_pulls_by_name_skipped_bytes_total`   | `name`                                                                                                                                  | Counter   | Bytes skipped by CRI-O image pulls by name.                                                                       |
```
discordianfish, line 64: Not sure what this means, might make sense to improve name and/or description. Is this due to image reuse? Are skipped bytes included in bytes_total? Is `crio_image_pulls_by_name_skipped_bytes_total/crio_image_pulls_by_name_bytes_total` somewhat like a "cache hit" ratio?

tutorials/metrics.md
discordianfish: Oh and size of course. Is that bounded? Otherwise this can't really be reasonably exposed: https://www.robustperception.io/cardinality-is-key

tutorials/metrics.md
```diff
@@ -39,22 +39,34 @@ metrics_key = "/path/to/key.pem"
...
+| `crio_operations_errors_total`                   | every CRI-O RPC\*                                                                                                                       | Counter   | Cumulative number of CRI-O operation errors by operation type.                                                    |
+| `crio_image_pulls_by_digest_bytes_total`         | `name`, `digest`, `mediatype`, `size`                                                                                                   | Counter   | Bytes transferred by CRI-O image pulls by digest.                                                                 |
+| `crio_image_pulls_by_name_bytes_total`           | `name`, `size`                                                                                                                          | Counter   | Bytes transferred by CRI-O image pulls by name.                                                                   |
```
discordianfish, line 63: Isn't this identical to `sum by (name, size) (crio_image_pulls_by_digest_bytes_total)`? In that case it's redundant.

discordianfish:
I'd also drop the `_by_name_` parts of the metric. It's uncommon to include the labels in the metric name like that and we don't do it consistently anyway.

tutorials/metrics.md
swghosh: Yes, this is kinda redundant. @saschagrunert can we go ahead and deprecate `crio_image_pulls_by_name` altogether instead of introducing `crio_image_pulls_by_name_bytes_total` as it's canonical.

swghosh:
> I'd also drop the `_by_name_` parts of the metric. It's uncommon to include the labels in the metric name like that and we don't do it consistently anyway.

@discordianfish That can be done for most parts (i.e. `image_pulls*`). There is some problem though:
- once PTAL at `containers_oom`, `containers_oom_total` and `containers_oom_by_name_total`, this PR in it's current state intends to introduce `containers_oom_by_name_total` in favour of deprecating `containers_oom` metric. Prometheus best practices suggest that all counters should be suffixed `_total`?
- one resolution would be too deprecate current `containers_oom_total` which we have right now and instead rename `containers_oom` to `containers_oom_total` ensuring that label `name` is still retained, for users willing to access total across names they can use `sum(containers_oom_total)`.

tutorials/metrics.md
haircommander: yeah that's a good way of thinking about it

swghosh: @haircommander @discordianfish So we can keep both of these metrics with `_bytes_total` suffix and remove the `_by_name` part as well, right?

swghosh: Also, am now worried about the cardinality of `name` label, should that be much concern?

tutorials/metrics.md
haircommander: the latency of the last cri call as of now. Then a user can determine the average for the time window they care about (did some operation increase the time a certain amount?)

tutorials/metrics.md
haircommander: it's likely digest will have quite high cardinality... For size we could sort into buckets instead of reporting the size directly

tutorials/metrics.md
haircommander: yeah we'll have to make a hard decision about the crio_image ones. If we do any reporting on the image level we're at risk of high cardinality. I don't *think* this number will get huge (order of 100s?), but I am not personally aware of what constitutes high

tutorials/metrics.md
haircommander: > Isn't this identical to sum by (name, size) (crio_image_pulls_by_digest_bytes_total)? In that case it's redundant.

good point, though if we end up reconsidering listing each by digest as we're discussing in https://github.com/cri-o/cri-o/pull/5487#discussion_r763360192 then we should keep.

haircommander: also won't all the _bytes fields have a large cardinality? maybe we should sort into buckets...

tutorials/metrics.md
swghosh: @haircommander @saschagrunert 
Sorting sizes into buckets seems like a reasonable way to reduce cardinality of size. What bucket sizes would be preferred? (any reference implementations?)
Also, we can decide if we really wanna drop digest altogether.

tutorials/metrics.md
swghosh: > also won't all the _bytes fields have a large cardinality? maybe we should sort into buckets...

Yes, similar to previous comment. 
Also, should be remove bytes label as well?

jan--f:
> * label `name` can lead to high cardinality as well?

I would think so. The `name` and `size` labels for the `*_image_*` metrics seem like they have potential for _very_ high cardinality, as possible values are essentially unbounded (iiuc). Furthermore those labels specifically might cause a sharp cardinality increase on cluster or operator updates (when a lot of images are changed to a new version).
If those labels aren't crucial it might be worth it to reconsider those. To still get stats for individual images a log message might be better?

server/metrics/metrics.go
```diff
@@ -184,6 +201,96 @@ func New(config *libconfig.MetricsConfig) *Metrics {
...
 			},
 		),
+
+		metricOperationsTotal: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Subsystem: collectors.Subsystem,
+				Name:      collectors.OperationsTotal.String(),
+				Help:      "Cumulative number of CRI-O operations by operation type.",
+			},
+			[]string{"operation_type"},
```
jan--f, line 211:
```suggestion
			[]string{"operation"},
```
Does this label name really require the `_type` suffix? Same for the other `operations` metrics.

swghosh: We can use just `operation` IMO.

tutorials/metrics.md
discordianfish: There is `crio_image_pulls_layer_size` already, which is essentially a counter of pulls by size.
Digest probably needs to get dropped

tutorials/metrics.md
discordianfish: Yeah not sure about the name either. On the most system this should be max in the 100s per node, but nothing might prevent users from pulling millions of images. Beside being a problem to ingest to many metrics, crio will also hold these in memory until it gets restarted.

discordianfish:
> one resolution would be too deprecate current containers_oom_total which we have right now and instead rename containers_oom to containers_oom_total ensuring that label name is still retained, for users willing to access total across names they can use sum(containers_oom_total).

That seems to be the best option. Just not sure to not break users in that case.

server/metrics/metrics.go
```diff
@@ -28,23 +28,40 @@ func SinceInMicroseconds(start time.Time) float64 {
...
-	metricImageLayerReuse         *prometheus.CounterVec
+	metricImageLayerReuse         *prometheus.CounterVec // to be deprecated
 	metricContainersOOMTotal      prometheus.Counter
-	metricContainersOOM           *prometheus.CounterVec
+	metricContainersOOM           *prometheus.CounterVec // to be deprecated
 	metricProcessesDefunct        prometheus.GaugeFunc
+
+	metricOperationsTotal                   *prometheus.CounterVec
+	metricOperationsLatencySeconds          *prometheus.GaugeVec
+	metricOperationsLatencySecondsTotal     *prometheus.SummaryVec
```
jan--f, line 56: I think these three metrics produce a lot of duplicate information. A Prometheus summary will include a total count and a sum. So at least `metricOperationsTotal` is not needed.
I wonder how useful `metricOperationsLatencySeconds` is as a gauge. A gauge would typically be used to get a moment in time value for something that can go up or down. If it is more interesting to look at changes over a time period a counter and `rate()` is often a good alternative. See also https://prometheus.io/docs/practices/instrumentation/#counter-vs-gauge-summary-vs-histogram.

So unless you really need the gauge I'd say that the summary `metricOperationsLatencySecondsTotal` has everything you need. The summary metric should then also be renamed to `metricOperationsLatencySeconds` as this type will add the buckets and a count and sum metric with the respective suffixes (see https://prometheus.io/docs/practices/histograms/)

tutorials/metrics.md
discordianfish: Oh and not sure if that was clear: Skipped bytes should be included in bytes_total if not and the doc should clarify that. Not sure if that's the case already but it makes calculating ratios more convenient.

swghosh: > Oh and not sure if that was clear: Skipped bytes should be included in bytes_total if not and the doc should clarify that. Not sure if that's the case already but it makes calculating ratios more convenient.

@discordianfish you mean exposing a metric like `skip_ratio` be more convenient that `skipped_bytes`, right?

discordianfish: @swghosh No, I'm just saying that skipped bytes should be included in total bytes so that `crio_image_pulls_by_name_skipped_bytes_total/crio_image_pulls_by_name_bytes_total` would calculate the ratio. Might be already the case in which I'd just point that out in the docs here.

server/metrics/metrics.go
swghosh: Resolved in 563ee1a.

tutorials/metrics.md
swghosh: Introduced metric names as 
`crio_image_pulls_bytes_total` and `crio_image_pulls_skipped_bytes_total` that follow naming best practices. Dropped high cardinality labels like `name`, `digest`. However, `size` is being used but with sorted buckets so as to keep cardinality in check.

tutorials/metrics.md
swghosh: Updated the doc in latest commit.

tutorials/metrics.md
swghosh: Label `digest` was dropped as also stated beneath.
> Introduced metric names as
> `crio_image_pulls_bytes_total` and `crio_image_pulls_skipped_bytes_total` that follow naming best practices. Dropped high cardinality labels like `name`, `digest`. However, `size` is being used but with sorted buckets so as to keep cardinality in check.

server/metrics/metrics.go
swghosh: As of now we're not dropping any existing metric in order to introduce a non-breaking change. We're just adding new metrics for idiomatic existing ones that follow Prometheus naming best practices and reduce cardinality. However, the insight shared is helpful and probably we can drop some metrics like the ones stated in a future release when refactor our metrics collector again to remove metrics instead.

swghosh:
Updated release note with metric names as per 
563ee1a.

server/metrics/collectors/collectors.go
```diff
@@ -16,37 +16,37 @@ const (
 
 	subsystemPrefix = Subsystem + "_"
 
-	// Operations is the key for CRI-O operation metrics.
+	// Operations is the key for CRI-O operation metrics. (to be deprecated)
```
saschagrunert, line 19: Let's add in which favor it has been deprecated.

server/metrics/collectors/collectors.go
```diff
@@ -57,6 +57,36 @@ const (
 
 	// ProcessesDefunct is the key for the total number of defunct processes in a node.
 	ProcessesDefunct Collector = crioPrefix + "processes_defunct"
+
+	// OperationsTotal is the key for CRI-O operation metrics.
+	OperationsTotal Collector = crioPrefix + "operations_total"
+
+	// OperationsLatency is the key for the operation latency metrics for each CRI call.
```
saschagrunert, line 64:
```suggestion
	// OperationsLatencySeconds is the key for the operation latency metrics for each CRI call.
```

server/metrics/collectors/collectors.go
```diff
@@ -57,6 +57,36 @@ const (
...
+	// OperationsLatencySecondsTotal is the key for the operation latency metrics.
+	OperationsLatencySecondsTotal Collector = crioPrefix + "operations_latency_seconds_total"
+
+	// OperationsErrorsTotal is the key for the operation error metrics.
+	OperationsErrorsTotal Collector = crioPrefix + "operations_errors_total"
+
+	// ImagePullsBytesTotal is the key for CRI-O image pull metrics by digest.
+	ImagePullsBytesTotal Collector = crioPrefix + "image_pulls_bytes_total"
+
+	// ImagePullsSkippedBytesTotal is the key for CRI-O skipped image pull metrics by name (skipped).
```
saschagrunert, line 76:
```suggestion
	// ImagePullsSkippedBytesTotal is the key for CRI-O skipped image pull metrics by name.
```

server/metrics/metrics.go
```diff
@@ -28,23 +28,67 @@ func SinceInMicroseconds(start time.Time) float64 {
...
+	metricImagePullsFailures      *prometheus.CounterVec // to be deprecated
+	metricImagePullsSuccesses     *prometheus.CounterVec // to be deprecated
 	metricImagePullsLayerSize     prometheus.Histogram
-	metricImageLayerReuse         *prometheus.CounterVec
+	metricImageLayerReuse         *prometheus.CounterVec // to be deprecated
 	metricContainersOOMTotal      prometheus.Counter
-	metricContainersOOM           *prometheus.CounterVec
+	metricContainersOOM           *prometheus.CounterVec // to be deprecated
 	metricProcessesDefunct        prometheus.GaugeFunc
+
```
saschagrunert, line 81: Let's remove this newline and reformat.
```suggestion
```

server/metrics/metrics.go
```diff
@@ -57,23 +101,23 @@ func New(config *libconfig.MetricsConfig) *Metrics {
 			prometheus.CounterOpts{
 				Subsystem: collectors.Subsystem,
 				Name:      collectors.Operations.String(),
-				Help:      "Cumulative number of CRI-O operations by operation type.",
+				Help:      "[DEPRECATED] Cumulative number of CRI-O operations by operation type.",
```
saschagrunert, line 104: Let's add in which favor it has been deprecated.

tutorials/metrics.md
```diff
@@ -39,22 +39,33 @@ metrics_key = "/path/to/key.pem"
...
+| Metric Key                                       | Possible Labels or Buckets                                                                                                              | Type      | Purpose                                                                                                           |
+| ------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------- | --------- | ----------------------------------------------------------------------------------------------------------------- |
+| `crio_operations`                                | every CRI-O RPC\*                                                                                                                       | Counter   | (DEPRECATED) Cumulative number of CRI-O operations by operation type.                                             |
```
saschagrunert, line 44: Let's move the deprecated ones to the end of the table and add which should be used instead.

swghosh: Changed in b1a1bfa.

server/metrics/metrics.go
swghosh: b1a1bfa.

server/metrics/collectors/collectors.go
swghosh: b1a1bfa

server/metrics/collectors/collectors.go
swghosh: Added in b1a1bfa.

saschagrunert (approved):
LGTM

/assign @haircommander

tutorials/metrics.md
swghosh: https://github.com/cri-o/cri-o/blob/854e2fda2f261efcce3744888724d97f348602be/server/image_pull.go#L148-L220
I think total pulled bytes also includes skipped bytes.

internal/oci/runtime_oci.go
```diff
@@ -896,7 +896,7 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 		metrics.Instance().MetricContainersOOMTotalInc()
 
 		// Collect metric by container name
-		metrics.Instance().MetricContainersOOMInc(c.Name())
+		metrics.Instance().MetricContainersOOMCountTotalInc(c.Name())
```
haircommander, line 899: we need to increment the old one too until we finish deprecating it

internal/oci/runtime_vm.go
```diff
@@ -707,7 +707,7 @@ func (r *runtimeVM) updateContainerStatus(ctx context.Context, c *Container) err
 			metrics.Instance().MetricContainersOOMTotalInc()
 
 			// Collect metric by container name
-			metrics.Instance().MetricContainersOOMInc(c.Name())
+			metrics.Instance().MetricContainersOOMCountTotalInc(c.Name())
```
haircommander, line 710: same here

server/image_pull.go
```diff
@@ -203,6 +205,13 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
...
+				// Metrics for image pulls bytes
+				metrics.Instance().MetricImagePullsBytesAdd(
+					float64(p.OffsetUpdate),
+					p.Artifact.MediaType,
+					p.Artifact.Size,
+				)
+
 				// Metrics for size histogram
 				if p.Event == imageTypes.ProgressEventDone {
 					metrics.Instance().MetricImagePullsLayerSizeObserve(p.Artifact.Size)
```
haircommander, line 217: I forget, are we deprecating this one?

server/metrics/collectors/collectors.go
```diff
@@ -17,46 +17,87 @@ const (
...
+	ImagePullsFailureTotal Collector = crioPrefix + "image_pulls_failure_total"
+
+	// ImagePullsSuccessTotal is the key for successful image downloads in CRI-O.
+	ImagePullsSuccessTotal Collector = crioPrefix + "image_pulls_success_total"
+
+	// ImageLayerReuseTotal is the key for the CRI-O image layer reuse metrics.
+	ImageLayerReuseTotal Collector = crioPrefix + "image_layer_reuse_total"
+
+	// ContainersOOMCountTotal is the key for the CRI-O container out of memory metrics per container name.
+	ContainersOOMCountTotal Collector = crioPrefix + "containers_oom_count_total"
```
haircommander, line 100: 1: Should we mention in docs or comments the possibility of oom metric by name having high cardinality, but the expectation that it shouldn't and that the metric is worth keeping?
2: shouldn't we mention `by name` or something. `containers` implies it's a total count for all containers. maybe I'm misunderstanding the purpose of this one though

server/image_pull.go
swghosh: Which one?
`ImagePullsLayerSizeObserve` we're not.

haircommander:
@discordianfish @jan--f any further thoughts?

server/metrics/collectors/collectors.go
swghosh: Yeah: we can mention in the docs about the high cardinality but why we intend to keep it with the reasonable assumption that number of containers OOMing would be low ideally.

swghosh: We're not keeping `_by_name` in the metric name as previously discussed in above thread(s) that `_by_name` part inclusion in metric is bad naming.

tutorials/metrics.md
swghosh: We can include about the skip ratio in doc. (that: `image_pulls_skipped_bytes_total / image_pulls_bytes_total` is actually a ratio)

internal/oci/runtime_oci.go
swghosh: So I added both here:
https://github.com/cri-o/cri-o/blob/5b7e1417fbb95bb8bda996bf96719700fbe4d06b/server/metrics/metrics.go#L420-L434

Assuming that we don't have to touch `internal/oci/runtime_oci.go` file when we want to remove it in course of deprecation.

internal/oci/runtime_vm.go
swghosh: > So I added both here:
> 
> https://github.com/cri-o/cri-o/blob/5b7e1417fbb95bb8bda996bf96719700fbe4d06b/server/metrics/metrics.go#L420-L434
> 
> Assuming that we don't have to touch `internal/oci/runtime_oci.go` file when we want to remove it in course of deprecation.


Similarly in `internal/oci/runtime_vm.go`.

server/image_pull.go
haircommander: doesn't this have the possibility of having high cardinality? (a different field for each distinct size)

internal/oci/runtime_oci.go
haircommander: good idea

server/image_pull.go
swghosh: It's a Histogram with already sorted bucket sizes: so no harm to cardinality.
https://github.com/cri-o/cri-o/blob/5b7e1417fbb95bb8bda996bf96719700fbe4d06b/server/metrics/metrics.go#L172-L191

haircommander: reading is hard :upside_down_face: thank you

haircommander:
docs validation is still failing, and one more nit for a doc update for oom, otherwise LGTM

server/metrics/collectors/collectors.go
swghosh: Added 1. in https://github.com/cri-o/cri-o/blob/8ec8b7396e018ad08f28b159a235aaff06d6287e/server/metrics/metrics.go#L308-L310

tutorials/metrics.md
swghosh: Included in 8ec8b7396.
https://github.com/cri-o/cri-o/blob/8ec8b7396e018ad08f28b159a235aaff06d6287e/tutorials/metrics.md#L49

server/metrics/metrics.go
```diff
@@ -305,6 +305,9 @@ func New(config *libconfig.MetricsConfig) *Metrics {
 				Name:      collectors.ContainersOOMCountTotal.String(),
 				Help:      "Amount of containers killed because they ran out of memory (OOM) by their name",
 			},
+			// The label `name` can have high cardinality sometimes but it is in the interest of users giving them the
```
haircommander, line 308: nit: i think this should go in metrics.md too

haircommander:
sorry, one more nit and also can you squash your commits (using `git rebase -i upstream/main` and selecting `s` for the commits you want to squash together)? Each commit should be a unit of work, and ideally they shouldn't undo each other (For instance https://github.com/cri-o/cri-o/pull/5487/commits/c5ea9ff418585c2617770a9cdfbeab329aff94df undoes a change in https://github.com/cri-o/cri-o/pull/5487/commits/bbd92d0d79ad6b8a8814d5558077d28502438205)

server/metrics/metrics.go
swghosh: Updated.

swghosh:
> sorry, one more nit and also can you squash your commits (using `git rebase -i upstream/main` and selecting `s` for the commits you want to squash together)? Each commit should be a unit of work, and ideally they shouldn't undo each other (For instance [c5ea9ff](https://github.com/cri-o/cri-o/commit/c5ea9ff418585c2617770a9cdfbeab329aff94df) undoes a change in [bbd92d0](https://github.com/cri-o/cri-o/commit/bbd92d0d79ad6b8a8814d5558077d28502438205))

@haircommander You can PTAL, in case we need to squash more from the commits that came from review comment related changes!

haircommander:
well done @swghosh thanks for your work!

saschagrunert:
Great job @swghosh!

Issue #5479, opened 2021-11-30
CRIO container is running with non-existing image reference


EvgenyAGRO:
**Description**

Some inconsistencies were observed in CRIO image storage and running containers. 

**Steps to reproduce the issue:**
1. Pull ubuntu:16.04, tag it as ubuntu:latest and push to some registry
2. Run pod with ubuntu:latest container with name ubuntu1
3. Delete the tag from the registry 
4. Pull ubuntu:18.04, tag it as ubuntu:latest and push to same registry
5. Start another pod with ubuntu:latest container with name ubuntu2

**Describe the results you received:**
Ubuntu2 container was behaving normally.
Ubuntu1 container on the other hand contained `imageRef` of digest that has vanished from the storage.

Running `crictl inspect <ubuntu1_container_id>` showed:
```
    "image": {
      "annotations": {},
      "image": "docker.io/library/ubuntu:latest"
    },
    "imageRef": "docker.io/library/ubuntu@sha256:3555f4996aea6be945ae1532fa377c88f4b3b9e6d93531f47af5d78a7d5e3761",
```

Querying this image reference with `crictl inspecti docker.io/library/ubuntu@sha256:3555f4996aea6be945ae1532fa377c88f4b3b9e6d93531f47af5d78a7d5e3761` just fails with no such image:
```
FATA[0000] no such image "docker.io/library/ubuntu@sha256:3555f4996aea6be945ae1532fa377c88f4b3b9e6d93531f47af5d78a7d5e3761" present 
```

I then tried to find that problematic image so i checked `crictl images | grep ubuntu`:
```
docker.io/library/ubuntu                                                             latest               ba6acccedd292      75.2MB
docker.io/library/ubuntu                                                             <none>            597ce1600cf4a       75.2MB
```

When checking image with none tag  `crictl inspecti 597ce1600cf4a` I found an image with no repoTags and repoDigests:
```
"status": {
    "id": "597ce1600cf4ac5f449b66e75e840657bb53864434d6bd82f00b172544c32ee2",
    "repoTags": [],
    "repoDigests": [],
    "size": "75159392",
    "uid": null,
    "username": "",
    "spec": null
  }
```

When running `cat /var/lib/containers/storage/overlay-images/images.json | jq | grep 597ce1600cf4a -A10`:
```
  {
    "id": "597ce1600cf4ac5f449b66e75e840657bb53864434d6bd82f00b172544c32ee2",
[...]

PR #5491, opened 2021-12-09
RFC: Do not use repo digests as image references

saschagrunert:
#### What type of PR is this?

/kind bug

Repo digests can change on pull or tag, while we then invalidate the
`imageRef` in the container status. To avoid such scenarios, we now
always reference the image ID.

Ref: https://github.com/kubernetes/cri-api/blob/c4e249a/pkg/apis/runtime/v1/api.proto#L1048-L1050

Fixes https://github.com/cri-o/cri-o/issues/5479

None

```release-note
Fixed bug where image references in container statuses are invalid digests because of an image pull or re-tag. We now always reference image ID's there.
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -220,9 +220,6 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrIface.Contai
 
 	imageName := imgResult.Name
 	imageRef := imgResult.ID
-	if len(imgResult.RepoDigests) > 0 {
-		imageRef = imgResult.RepoDigests[0]
-	}
 
 	labelOptions, err := ctr.SelinuxLabel(sb.ProcessLabel())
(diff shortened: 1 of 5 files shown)


saschagrunert:
@nalind PTAL

vrothberg (commented):
LGTM

It makes sense to use the ID.  A local index/manifest list may change and hence its digest while the ID remains.

mtrmac:
> A local index/manifest list may change and hence its digest while the ID remains.

The same digest directly implies the same ID. We also store all pulled manifests locally, and never remove them; any time an image can be looked up by this ID it could in principle also be looked up using a `repo@digest` reference described in #5479 , in principle.

Note to self: This requires understanding what the Kubelet (and k8s components that process Kubelet status or node status?) do with the data. Look into that before a review.

PR #5497, opened 2021-12-10
Use timeout context for conmon cgroup move

saschagrunert:
#### What type of PR is this?

/kind bug

We're now using the same context for moving conmon into the cgroup to
correctly timeout if needed.

Refers to https://github.com/cri-o/cri-o/pull/5496 and https://github.com/cri-o/cri-o/pull/5495

```release-note
Correctly timeout container creation when moving conmon into the cgroup (systemd cgroup driver)
```

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/containers/podman/v3/pkg/lookup"
 	"github.com/cri-o/cri-o/internal/dbusmgr"
@@ -54,9 +55,7 @@ func StatusToExitCode(status int) int {
 }
 
 // RunUnderSystemdScope adds the specified pid to a systemd scope
-func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName string, properties ...systemdDbus.Property) error {
-	ctx := context.Background()
-	var err error
+func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName string, properties ...systemdDbus.Property) (err error) {
 	// sanity check
 	if mgr == nil {
 		return errors.New("dbus manager is nil")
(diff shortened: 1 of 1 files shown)


internal/storage/image.go
```diff
@@ -472,31 +472,36 @@ func copyImageChild() {
 		os.Exit(1)
 	}
 
-	if err := moveSelfToCgroup(args.ParentCgroup, args.HasCollectMode); err != nil {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	if err := moveSelfToCgroup(ctx, args.ParentCgroup, args.HasCollectMode); err != nil {
 		fmt.Fprintf(os.Stderr, "%v", err)
 		os.Exit(1)
 	}
```
TomSweeneyRedHat, line 480: could we do `defer cancel()`

TomSweeneyRedHat:
One question for consideration, others LGTM once tests are happy.

mrunalp:
Yeah, this makes sense to me :+1:

mrunalp:
@haircommander can you review/merge?

haircommander:
I don't think we should use this context because it will negate the resource cache work. We can take a similar route as with network setup where we make a new context with a preposterously long context (6 min or something) to catch when requests go into the void

utils/utils.go
```diff
@@ -72,8 +71,10 @@ func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName
 	}
 	ch := make(chan string)
 	if err := mgr.RetryOnDisconnect(func(c *systemdDbus.Conn) error {
+		ctx, cancel := context.WithTimeout(context.Background(), 6*time.Minute)
```
saschagrunert, line 74: @haircommander I may miss some context here, but why do we decide to have a longer timeout than the default RPC one of 2minutes?

This would still cause a deadlock of 4minutes in the worst case.

saschagrunert:
> I don't think we should use this context because it will negate the resource cache work. We can take a similar route as with network setup where we make a new context with a preposterously long context (6 min or something) to catch when requests go into the void

Changed the PR to iterate faster on this. PTAL

internal/storage/image.go
saschagrunert: Not any more since I changed the PR to not touch this code section again. :)

utils/utils.go
haircommander: Basically, between the container rootfs creation, selinux relabeling, recursive chown, container creation and call to dbus to move conmon, it is quite possible we will spend more than 2 minutes on this call. IMO, the worst case is that we spend two minutes doing this work, stall at this call here, timeout and return to the kubelet with an error (cleaning up our provisioned resources on the way), after which the kubelet will retry, and this process repeats theoretically forever. We want to try our hardest to not undo any of the work we've already done, or else whatever is stalling the container creation past two minutes will only make things worse.

Cases like this are only to catch extremely rare problems where a request disappears. we don't want to wait forever, but we do want to wait as long as we can to try and save duplicated work

PR #5514, opened 2021-12-20
prepend commands with taskset if InfraCtrCPUSet is configured

haircommander:
#### What type of PR is this?

/kind feature
If InfraCtrCPUSet is configured, then there is an expectation that all commands land on certain CPUs. However, this isn't always the case. Thus, we need a way of spawning commands given a list of CPUs.

To do this, update cmdrunner to be a singleton object that stores an optional command and args to prepend to commands. Then use this singleton anytime exec.Command would be called.

```release-note
Add functionality to use taskset to spawn new commands cri-o runs. Now, if InfraCtrCPUSet is called, all newly spawned commands will be placed in the InfraCtrCPUSet (as it's expected to be set to the reserved CPU set that system commands should run on).
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -311,6 +311,7 @@ testunit-bin:
 	done
 
 mockgen: \
+	mock-cmdrunner \
 	mock-containerstorage \
 	mock-criostorage \
 	mock-lib-config \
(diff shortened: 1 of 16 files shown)


mrunalp:
For cgroups v2, longer term, we should utilize the new clone3 options that allow switching to a specific cgroup for the child process.

giuseppe:
> For cgroups v2, longer term, we should utilize the new clone3 options that allow switching to a specific cgroup for the child process.

unfortunately that is still not possible with systemd: https://github.com/systemd/systemd/issues/18843

giuseppe (approved):
LGTM

mrunalp (approved):
/lgtm

PR #5532, opened 2022-01-06
restore: ensure containers are wiped on reboot

haircommander:
If pods were successfully restored, but the node rebooted, then the pods should be removed

Note: this currently doesn't happen, as the pod currently fails to restore because
the their namespaces are in a tmpfs, but eventually a pod will be able to exist without its namespace
existing.

#### What type of PR is this?

/kind bug

```release-note
none
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -520,18 +520,56 @@ func (s *Server) wipeIfAppropriate(ctx context.Context, imagesToDelete []string)
 	if !s.config.InternalWipe {
 		return
 	}
+	var (
+		shouldWipeContainers, shouldWipeImages bool
+		err                                    error
+	)
+
 	// Check if our persistent version file is out of date.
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -520,18 +520,46 @@ func (s *Server) wipeIfAppropriate(ctx context.Context, imagesToDelete []string)
 	if !s.config.InternalWipe {
 		return
 	}
+	// Check if our version file is out of date.
+	// If so, we rebooted, and we should wipe containers.
+	shouldWipeContainers, err := version.ShouldCrioWipe(s.config.VersionFile)
```
mrunalp, line 525: Does this also cover a reboot without a version change?

haircommander: yeah if the file isn't present it wipes

saschagrunert:

PR #5535, opened 2022-01-10
Use timeout for conmon cgroup move

haircommander:
We're now using a timeout which is able to unblock gaining
the sandbox mutex in a container creation vs sandbox stop race.

Signed-off-by: Sascha Grunert <sgrunert@redhat.com>

#### What type of PR is this?

/kind bug

```release-note
none
```

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/containers/podman/v3/pkg/lookup"
 	"github.com/cri-o/cri-o/internal/dbusmgr"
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Change LGTM, but 6 minutes seems a bit long to wait.
Couple of tests still aren't hip.

mrunalp (approved):
/lgtm

utils/utils.go
```diff
@@ -73,14 +73,24 @@ func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName
...
+		close(ch)
+	case <-time.After(time.Minute * 6):
+		// This case is a work around to catch situations where the dbus library sends the
+		// request but it unexpectedly disappears. We set the timeout very high to make sure
+		// we wait as long as possible to catch situations where dbus is overwhelmed.
+		// We also don't use the native context cancelling behavior of the dbus library,
+		// because experience has shown that it does not help.
+		// TODO: Find cause of the request being dropped in the dbus library and fix it.
+		return errors.Errorf("timed out moving conmon with pid %d to cgroup", pid)
+	}
```
saschagrunert, line 93: Two ideas coming into my mind:
- What if we reduce the timeout to 1 minute and retry multiple times with an exponential backoff?
- What if we check for the error "Message recipient disconnected from message bus without replying" and restart dbus in that case?

haircommander: how would we be able to get the error in this case?

PR #5551, opened 2022-01-19
config: add allowed_devices field

haircommander:
#### What type of PR is this?

/kind feature
in cases where the experimental feature "io.kubernetes.cri-o.Devices" is enabled, any user is allowed to specify any device on the host, as long as they're in the proper workload. Unfortunately, this poses a security risk, as it essentially pokes a hole between the container and host boundry.

Instead, admins should have the ability to set a list of allowed devices their containers are allowed to use. that way, the access of those devices is more deliberate and secure

```release-note
add `allowed_devices` field to config, allowing admins to specify which devices are allowed to be specified in the "io.kubernetes.cri-o.Devices" allowed_annotation. The default for this config field is [/dev/fuse]
```


internal/config/device/device.go
```diff
@@ -74,6 +78,12 @@ func devicesFromStrings(devsFromConfig []string) ([]Device, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		if allowedDevices != nil {
+			if _, ok := allowedDevices[src]; !ok {
+				return nil, errors.Errorf("Device %s is specified but is not in allowed_devices", src)
```
saschagrunert, line 84: nit
```suggestion
				return nil, errors.Errorf("device %s is specified but is not in allowed_devices", src)
```

saschagrunert (approved):
LGTM

docs/crio.8.md
```diff
@@ -137,6 +138,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--additional-devices**="": Devices to add to the containers  (default: [])
 
+**--allowed-devices**="": Devices a user is allowed to specify with "io.kubernetes.cri-o.Devices" allowed annotation (default: [/dev/fuse])
```
TomSweeneyRedHat, line 141: Suggestion for consideration

```suggestion
**--allowed-devices**="": Devices a user is allowed to specify with the "io.kubernetes.cri-o.Devices" annotation (default: [/dev/fuse])
```

TomSweeneyRedHat: and if you use here, ditto below

TomSweeneyRedHat:
One small suggestion for consideration, otherwise LGTM.

mrunalp (approved):
/lgtm

PR #5552, opened 2022-01-20
userns: set User.UID to valid value in "auto" mode

frasertweedale:
#### What type of PR is this?

/kind feature

When creating a container that uses a user namespace, there are
three possibilities of how the initial value of
`process.user.[gu]id` relates to the user namespace of the sandbox
(Pod):

1. The value matches the host side of the mapping.  We replace it
   with the corresponding container UID/GID.

2. The value matches the container side of the mapping.  Commonly,
   this will `0` or another low UID, to allow executing a process as
   `root` or some other special UID inside the container, which is
   mapped to an unprivileged UID on the host.  We can use the value
   as-is.

   To allow the use of a low UID, the Pod must be admitted by a
   privileged account - even when using unprivileged host UIDs in the
   mapping.  This is because the admission system is not yet aware of
   user namespaces.  This is undesirable from a security
   perspective, and is the main issue this change addresses.

3. The value does not match either side of the mapping.  This can
   happen when the Pod is admitted by an unprivileged account.  We
   return the value unchanged.  Later, the container process will
   fail to start because it cannot be executed under an unmapped
   UID.

To enable the use of user namespaces without requiring special
privileges (e.g. `anyuid` SCC) for Pod admission, this commit
introduces new behaviour for scenario (3):

- In userns-mode "auto" we can be confident (up to `map-to-root` and
  `keep-id`) that CRI-O has chosen a *unique* mapping of
  unprivileged host UIDs for the Pod's user namespace (assuming
  sensible contents of `/etc/sub[gu]id`).  We cannot be so confident
  in "private" mode.  Therefore the new behaviour is restricted to
  "auto" mode.

- When `process.user.[gu]id` does not correspond to either the host
  or container side of the mapping, replace it with the smallest
  container ID in the mapping.  This heuristic assumes that (a) the
  user does want to run the containers, and (b) the lowest ID
  *inside the container* is an appropriate user to run it as.

- The `map-to-root` and `keep-id` options, when used, violate the
  assumption of unique, non-overlappping namespaces in "auto" mode,
  and present a risk of privilege escalation.  To mitigate this
  risk, also check the host IDs in the mapping and do not use any
  container ID that maps to host ID 0.

As a result of this change, it is possible for accounts without
privileges to create Pods with `RunAsUser: 0` (or other low number)
to use userns-mode "auto" to run containers in a user namespace,
with the container process running as UID 0 in the container and an
unprivileged user on the host.

None

```release-note
When using the "auto" mode for user namespaces, if the container process UID does not exist in the user
namespace of either the container or the host, CRI-O replaces it with a UID in the container's user namespace.
As a result, the container process can successfully execute.
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -92,7 +92,7 @@ func makeMountsAccessible(uid, gid int, mounts []rspec.Mount) error {
 	return nil
 }
 
-func toContainer(id uint32, idMap []idtools.IDMap) uint32 {
+func toContainer(id uint32, idMap []idtools.IDMap, usernsMode string) uint32 {
 	hostID := int(id)
 	if idMap == nil {
 		return uint32(hostID)
@@ -103,13 +103,50 @@ func toContainer(id uint32, idMap []idtools.IDMap) uint32 {
 			return uint32(contID)
 		}
 	}
+
 	// If the ID cannot be mapped, it means the RunAsUser or RunAsGroup was not specified
-	// so just use the original value.
+	// or that the specified host UID/GID is unmapped.
+	//
+	// From here, let's next check whether `id` corresponds to an ID on the container's
+	// side of the mapping.  If so, we can use it as-is.
+	for _, m := range idMap {
+		if hostID >= m.ContainerID && hostID < m.ContainerID+m.Size {
+			return id
+		}
+	}
+
+	// Finally, if we're in "auto" mode we know that CRI-O chose a unique host ID range
+	// to map into the container's user namespace (modulo 'keep-id' or 'map-to-root').
+	// Choose the lowest container ID in the mapping so that the container process can run,
+	// ignoring values that map to root on the host (to defend against privilege escalation).
+	if usernsMode == "auto" {
+		lowestContainerID := 0
+		lowestContainerIDSet := false
+		for _, m := range idMap {
+			containerID := m.ContainerID
+			if m.HostID == 0 { // host ID is root...
+				if m.Size < 2 {
+					continue // and no more IDs in this range; move on.
+				} else {
+					containerID += 1 // use next container ID
+				}
+			}
+			if !lowestContainerIDSet || containerID < lowestContainerID {
+				lowestContainerID = containerID
+				lowestContainerIDSet = true
+			}
+		}
+		if lowestContainerIDSet {
+			return uint32(lowestContainerID)
+		}
+	}
+
+	// None of the above heuristics worked; return original value
 	return id
 }
 
 // finalizeUserMapping changes the UID, GID and additional GIDs to reflect the new value in the user namespace.
-func (s *Server) finalizeUserMapping(specgen *generate.Generator, mappings *idtools.IDMappings) {
+func (s *Server) finalizeUserMapping(specgen *generate.Generator, mappings *idtools.IDMappings, usernsAnn string) {
 	if mappings == nil {
 		return
 	}
@@ -119,11 +156,14 @@ func (s *Server) finalizeUserMapping(specgen *generate.Generator, mappings *idto
 		return
 	}
 
-	specgen.Config.Process.User.UID = toContainer(specgen.Config.Process.User.UID, mappings.UIDs())
+	parts := strings.SplitN(usernsAnn, ":", 2)
+	usernsMode := parts[0]
+
+	specgen.Config.Process.User.UID = toContainer(specgen.Config.Process.User.UID, mappings.UIDs(), usernsMode)
 	gids := mappings.GIDs()
-	specgen.Config.Process.User.GID = toContainer(specgen.Config.Process.User.GID, gids)
+	specgen.Config.Process.User.GID = toContainer(specgen.Config.Process.User.GID, gids, usernsMode)
 	for i := range specgen.Config.Process.User.AdditionalGids {
-		gid := toContainer(specgen.Config.Process.User.AdditionalGids[i], gids)
+		gid := toContainer(specgen.Config.Process.User.AdditionalGids[i], gids, usernsMode)
 		specgen.Config.Process.User.AdditionalGids[i] = gid
 	}
 }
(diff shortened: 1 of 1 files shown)


giuseppe (commented):
this behavior looks a bit magic.  If the UID is wrong, isn't better to deal with the error and fix it instead of picking a random ID?

haircommander:
/ok-to-test

thanks @frasertweedale ! I'll take a deeper look later today

frasertweedale:
> this behavior looks a bit magic. If the UID is wrong, isn't better to deal with the error and fix it instead of picking a random ID?

@giuseppe This situation arises in OpenShift due to namespace UID range assignments and SCCs.  If you admit the Pod via an SCC that does not have `RunAsUser` (e.g. an ordinary account without any "special privileges" SCCs granted), then OpenShift will try to run it as a user from the namespace's assigned uid-range.  As a consequence, the Pod may fail to execute.

With this change, unprivileged accounts can use user namespaces and successfully execute workloads that expect to run as `root` or some other specific UID within the container's user namespace.  It is an important change to avoid having to grant excessive privileges to the accounts/service accounts that deploy applications.

I am happy to amend the commit message to mention this motivation/use case, if you like.

giuseppe:
the expectation is that `RunAsUser` is injected into the user namespace mappings: https://github.com/cri-o/cri-o/blob/main/server/sandbox_run_linux.go#L147-L164

Why does that fail?  Isn't `RunAsUser` specified?

haircommander:
I agree with @giuseppe that it feels a bit weird that an arbitrary value suddently becomes root of the user namespace. At least, I think there should be a separate option (`root-if-out-of-range` is clunky but descriptive). 

I suppose we could also have an option that just says `run-as-always-root` which always maps whatever RunAsUser is specified to to root in the user namespace. Not sure the security implications of that though

haircommander:
also, something I've just thought of: SCC not being aware of this means that it could choose a UID within the range of another container's mapping, I think. I don't know how SCC chooses the randomly high UID, but if we allocate a range inside the container and apply that mapping on the host, couldn't another container slip in to that range and possibly cause trouble? thinkign out loud here, I don't actually know...

giuseppe:
> also, something I've just thought of: SCC not being aware of this means that it could choose a UID within the range of another container's mapping, I think. I don't know how SCC chooses the randomly high UID, but if we allocate a range inside the container and apply that mapping on the host, couldn't another container slip in to that range and possibly cause trouble? thinkign out loud here, I don't actually know...

yes I think that could happen.  Is the range used by SCC defined anywhere?  I'd expect it to not overlap with the range specified to `containers` in `/etc/subuid` and `/etc/subgid`.

Said so, I don't think it is really an issue.  It is not different than running without any user namespace as of today, where all the IDs are shared by all pods.  While in this case it is only a conflict between the ID picked by SCC with an additional ID in a different user namespace that is probably not used.

frasertweedale:
@guiseppe here is an example use case that exhibits the problem that this PR proposes to solve:

- You are using OpenShift
- You want to run a workload in a user namespace
- The container image specifies UID 0 (root) - e.g. a workload that uses systemd for service management
- The account creating the Pod does not have anyuid (or similar) SCC assigned to it.  It uses the default `restricted` SCC which uses the `MustRunAsRange` policy for `runAsUser`
- Therefore the PodSpec can only specify `runAsUser` in the uid-range of the Namespace - or leave it blank, and the `MustRunAsRange` policy object will fill it in with the UID at the bottom of the range
- This results in a conflict where the `RunAsUser` is not mapped in the Pod's user namespace.

giuseppe:
why doesn't `map-to-root` work for this use case?  The original runAsUser is ignored inside the container and it is mapped to `0`

frasertweedale:
> also, something I've just thought of: SCC not being aware of this means that it could choose a UID within the range of another container's mapping, I think. I don't know how SCC chooses the randomly high UID, but if we allocate a range inside the container and apply that mapping on the host, couldn't another container slip in to that range and possibly cause trouble? thinking out loud here, I don't actually know...

Not directly related to this PR, but I did some digging.  The *default* range from which uid-range assignments are made is `1000000000-1999999999` in increments of `10000`. See https://github.com/openshift/cluster-policy-controller/blob/4de9b839bea33462ea1c4e6fb602bda3bfbaf5dd/pkg/cmd/cluster-policy-controller/openshiftcontrolplane_config.go#L35.

This seems to be configurable via the object `openshiftcontrollermanagers/cluster`.  But I haven't actually tried to configure something different.  See the following type definitions:

- `SecurityAllocator` at: https://github.com/openshift/api/blob/master/openshiftcontrolplane/v1/types.go#L373-L393 
- `OpenShiftControllerManagerConfig` at https://github.com/openshift/api/blob/master/openshiftcontrolplane/v1/types.go#L164-L197 (includes a `securityAllocator` field)

frasertweedale:
> why doesn't `map-to-root` work for this use case? The original `runAsUser` is ignored inside the container and it is mapped to `0`

@giuseppe thanks for your patience and suggestions.  I eventually worked out that it is a combination of behaviour of the OpenShift SCC behaviour and CRI-O's behaviour that was resulting in the failures.  In brief, the SCC checks default the `runAsUser` in the *Container'* `SecurityContext` but not the *Pod* `SecurityContext`.  It is the latter that CRI-O inspects when processing the `map-to-root` option.  The workaround is to explicitly set the *Pod* `securityContext.runAsUser` to the appropriate value.

The appropriate value must be determined by inspecting the Project or Namespace object that will be the Pod's home.  This is extra work for an operator to perform, but it is tolerable.  I will liase with the OpenShift apiserver-library-go maintainers to see if we can enhance the SCC processing to also set the *Pod* `securityContext.runAsUser` which would avoid this extra work/complexity.

More details in my blog post: https://frasertweedale.github.io/blog-redhat/posts/2022-02-02-openshift-user-ns-without-anyuid.html.

In light of all this, I am closing this PR.  Thank you for your review and suggestions.

PR #5561, opened 2022-01-21
[1.23] config: add allowed_devices field

haircommander:
#### What type of PR is this?

/kind feature
in cases where the experimental feature "io.kubernetes.cri-o.Devices" is enabled, any user is allowed to specify any device on the host, as long as they're in the proper workload. Unfortunately, this poses a security risk, as it essentially pokes a hole between the container and host boundry.

Instead, admins should have the ability to set a list of allowed devices their containers are allowed to use. that way, the access of those devices is more deliberate and secure

```release-note
add `allowed_devices` field to config, allowing admins to specify which devices are allowed to be specified in the "io.kubernetes.cri-o.Devices" allowed_annotation. The default for this config field is [/dev/fuse]
```


wgahnagl (commented):
/lgtm

PR #5587, opened 2022-02-02
Enable `--seccomp-use-default-when-empty` by default

saschagrunert:
#### What type of PR is this?

/kind feature

This is a premature step before the graduation of the `seccompDefault`
feature planned for Kubernetes v1.25. We now use the `runtime/default`
profile for every workload specifying none (empty) in the pod manifest.

None
None

```release-note
Enable `--seccomp-use-default-when-empty`/`seccomp_use_default_when_empty` by default.
This is a premature step before the graduation of the `seccompDefault` feature planned for
Kubernetes v1.25. We now use the `runtime/default` profile for every workload specifying 
none (empty) in the pod manifest. 
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -732,6 +732,7 @@ func DefaultConfig() (*Config, error) {
 		return nil, err
 	}
 	cgroupManager := cgmgr.New()
+	seccompConfig := seccomp.New()
 	return &Config{
 		Comment: "# ",
 		SystemContext: &types.SystemContext{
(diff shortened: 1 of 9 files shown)


saschagrunert:
Okay this was expected:
screenshot

haircommander:
building on this, I've pushed a commit that adds a minimal seccomp profile that only blocks unshare. Possibly, we can extend it to the other syscalls in the default profile that require cap_sys_admin. This is to test whether the openshift tests still fail with a minimal profile

pkg/config/config.go
```diff
@@ -753,33 +754,34 @@ func DefaultConfig() (*Config, error) {
...
+			DropInfraCtr:               true,
+			SeccompUseDefaultWhenEmpty: seccompConfig.UseDefaultWhenEmpty(),
+			seccompConfig:              seccompConfig,
+			apparmorConfig:             apparmor.New(),
+			blockioConfig:              blockio.New(),
+			rdtConfig:                  rdt.New(),
+			ulimitsConfig:              ulimits.New(),
+			cgroupManager:              cgroupManager,
+			deviceConfig:               device.New(),
+			namespaceManager:           nsmgr.New(defaultNamespacesDir, ""),
```
TomSweeneyRedHat, line 784: It would be nice to have these in alphanumeric order.  A change like this would be easier to spot, plus it makes my OCD happy.

TomSweeneyRedHat (commented):
LGTM
assuming happy tests

pkg/config/config.go
saschagrunert: Fixed :+1:

saschagrunert:
Fixed the critest suite by disabling the feature for them. I assume we have to fix this one `critest` with the graduation of the `SeccompDefault` feature in any case.

haircommander:
openshift tests pass! I'm going to progressively try to extend the seccomp default to see what is breaking openshift tests

saschagrunert:
@haircommander two pods (dns and metrics scraper) seem to fail in the e2e tests. I'm going to gather the audit logs from the nodes, let's see if that works.

haircommander:
/kind api-change

(we probably should make a lot of noise about this change, end-users are gonna hit eperms and are gonna be confused)

littlejawa:
@saschagrunert it seems the kata-jenkins job is failing on the test you modified:
-----
11:04:23 not ok 86 ctr seccomp profiles runtime/default by empty field
11:04:23 # (in test file ctr_seccomp_kata_integration_tests.bats, line 55)
11:04:23 #   `[ "$status" -ne 0 ]' failed
-----

We are skipping several of these tests for Kata, and I realize this one should be skipped too - your modification is not introducing a bug, it's revealing a known issue.

I will modify the kata test script to skip this additional test for now.
Please ignore this failure on this PR.

saschagrunert:
> I will modify the kata test script to skip this additional test for now. Please ignore this failure on this PR.

Thank you!

saschagrunert:
@haircommander PTAL

haircommander:
/lgtm

:tada: I know we aren't in the blog writing business, but we should probably write a blog describing the change and what folks can do if their containers suddenly hit EPERM

TomSweeneyRedHat:
Unless my eyes are crossing from too many reviews today, I believe this is showing happy green test buttons....

PR #5590, opened 2022-02-02
server: fix race with kubelet

haircommander:
previously, CRI-O assumed that the kubelet wouldn't bump the attempt number
if it returned an error after a series of retries.

in reality, the kubelet *is* doing that, which means situations where we have lots of load on a node
means the issue is made worse by this workaround, not better.

Mitigate this by waiting until the timeout is completed before returning the fact that the resource is ready.
This will cause the kubelet to continue to believe the timing out has happened, and retry with the same attempt number.

```release-note
fix a bug where situations of excessive load on nodes causes containers to never actually start
```

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -177,7 +177,9 @@ func isContextError(err error) bool {
 }
 
 func (s *Server) getResourceOrWait(ctx context.Context, name, resourceType string) (string, error) {
-	const resourceCreationWaitTime = time.Minute * 4
+	// In 99% of cases, we shouldn't hit this timeout. Instead, the context should be cancelled.
+	// This is really to catch an unlikely case where the kubelet doesn't cancel the context.
+	const resourceCreationWaitTime = time.Minute * 6
 
 	if cachedID := s.resourceStore.Get(name); cachedID != "" {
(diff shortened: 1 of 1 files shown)


server/utils.go
```diff
@@ -177,7 +177,9 @@ func isContextError(err error) bool {
 }
 
 func (s *Server) getResourceOrWait(ctx context.Context, name, resourceType string) (string, error) {
-	const resourceCreationWaitTime = time.Minute * 4
+	// In 99% of cases, we shouldn't hit this timeout. Instead, the context should be cancelled.
+	// This is really to catch an unlikely case where the kubelet doesn't cancel the context.
+	const resourceCreationWaitTime = time.Minute * 6
```
mrunalp, line 182: We may want to make this configurable in the future.

TomSweeneyRedHat:
LGTM
assuming happy tests.  I concur with Mrunal's thought on making that configurable.

saschagrunert (approved):
/lgtm

haircommander:
related to https://bugzilla.redhat.com/show_bug.cgi?id=2049832

PR #5596, opened 2022-02-03
Fix incorrect test setup order in cgroups bats test.

klihub:
Fix a number of problems related to CI tests. Huge thanks to @haircommander in helping me narrow down the test setup ordering problem:

  - fix incorrect test setup order in cgroups BATS test
  - force BATS /usr/bin symlink in place (in case CI worker node has distro-specific BATS installed)
  - fix test cases trying to dereference unset variables (recent BATS runs tests with `set -u`) 

#### What type of PR is this?

/kind failing-test

An incorrect order of calling `setup_test` and evaluating `$TESTDIR` causes cgroups test cases to share supposedly private test files and random test failures. This typically manifests itself as a conflict between the `ctr with swap should be configured` and `ctr with swap should fail when swap is lower` tests, with either one failing with a seemingly impossible condition. In reality this is caused by one of those test cases overwriting the container configuration input file generated by the other, before that file could be taken into use by the failing test case. 

None

```release-note
None
```

### DIFF
--- a/test/cgroups.bats
+++ b/test/cgroups.bats
@@ -3,8 +3,8 @@
 load helpers
 
 function setup() {
-	newconfig="$TESTDIR/config.json"
 	setup_test
+	newconfig="$TESTDIR/config.json"
 }
 
 function teardown() {
--- a/test/critest.bats
+++ b/test/critest.bats
@@ -3,7 +3,7 @@
 load helpers
 
 function setup() {
-	if [[ -z $RUN_CRITEST ]]; then
+	if [ ! -v RUN_CRITEST ]; then
 		skip "critest because RUN_CRITEST is not set"
 	fi
 
@@ -19,7 +19,7 @@ function teardown() {
 	critest --parallel 8 \
 		--runtime-endpoint "unix://${CRIO_SOCKET}" \
 		--image-endpoint "unix://${CRIO_SOCKET}" \
-		--ginkgo.focus="${CRI_FOCUS}" \
-		--ginkgo.skip="${CRI_SKIP}" \
+		--ginkgo.focus="${CRI_FOCUS:-}" \
+		--ginkgo.skip="${CRI_SKIP:-}" \
 		--ginkgo.flakeAttempts=3 >&3
 }
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -1,5 +1,18 @@
 #!/usr/bin/env bash
 
+# BATS is broken wrt. 'set -u' with bash < 4.4.
+# Until/unless https://github.com/bats-core/bats-core/pull/550 is
+# merged, make sure we don't run with 'set -u' on offending bash.
+bash_version=$((1000 * $(echo "${BASH_VERSION%.[^.]*}" | tr '.' '+')))
+if [ "$bash_version" -ge 4004 ]; then
+    set -u
+fi
+
+# Work around skip() in old BATS versions not handling substitutions properly
+# wrt. 'set -u'. (fixed in commit 9b2b659e5ba769752c0d00dadb3ad9b58f59cbaa).
+# This can be removed once all CI nodes have been setup with new enough BATS.
+BATS_TEARDOWN_STARTED=${BATS_TEARDOWN_STARTED:-}
+
 # Root directory of integration tests.
 INTEGRATION_ROOT=${INTEGRATION_ROOT:-$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")}
 
@@ -246,11 +259,7 @@ function setup_img() {
 }
 
 function setup_crio() {
-    apparmor=""
-    if [[ -n "$1" ]]; then
-        apparmor="$1"
-    fi
-
+    apparmor="${1:-}"
     for img in "${IMAGES[@]}"; do
         setup_img "$img"
     done
@@ -373,8 +382,8 @@ function cleanup_pods() {
 }
 
 function stop_crio_no_clean() {
-    local signal="$1"
-    if [ -n "${CRIO_PID+x}" ]; then
+    local signal="${1:-}"
+    if [ -v CRIO_PID ]; then
         kill "$signal" "$CRIO_PID" >/dev/null 2>&1 || true
         wait "$CRIO_PID"
         unset CRIO_PID
@@ -399,7 +408,7 @@ function restart_crio() {
 }
 
 function cleanup_lvm() {
-    if [ -n "${LVM_DEVICE+x}" ]; then
+    if [ -v LVM_DEVICE ]; then
         lvm lvremove -y storage/thinpool
         lvm vgremove -y storage
         lvm pvremove -y "$LVM_DEVICE"
@@ -416,7 +425,7 @@ function cleanup_testdir() {
 }
 
 function cleanup_test() {
-    [ -z "$TESTDIR" ] && return
+    [ ! -v TESTDIR ] && return
     # show crio log (only shown by bats in case of test failure)
     if [ -f "$CRIO_LOG" ]; then
         echo "# --- crio.log :: ---"
@@ -561,19 +570,17 @@ function create_runtime_with_allowed_annotation() {
 default_runtime = "$NAME"
 [crio.runtime.runtimes.$NAME]
 runtime_path = "$RUNTIME_BINARY_PATH"
-runtime_root = "$RUNTIME_ROOT"
+runtime_root = "${RUNTIME_ROOT:-}"
 runtime_type = "$RUNTIME_TYPE"
 allowed_annotations = ["$ANNOTATION"]
 EOF
 }
 
 function create_workload_with_allowed_annotation() {
-    local act="$2"
+    local act="${2:-}"
     # Fallback on the specified allowed annotation if
     # a specific activation annotation wasn't specified.
-    if [[ -z "$act" ]]; then
-        act="$1"
-    fi
+    act="${act:-$1}"
     cat <<EOF >"$CRIO_CONFIG_DIR/01-workload.conf"
 [crio.runtime.workloads.management]
 activation_annotation = "$act"
@@ -593,7 +600,7 @@ function set_swap_fields_given_cgroup_version() {
 
 function check_conmon_cpuset() {
     local ctr_id="$1"
-    local cpuset="$2"
+    local cpuset="${2:-}"
     systemd_supports_cpuset=$(systemctl show --property=AllowedCPUs systemd || true)
 
     if [[ "$CONTAINER_CGROUP_MANAGER" == "cgroupfs" ]]; then
--- a/test/namespaces.bats
+++ b/test/namespaces.bats
@@ -35,7 +35,7 @@ function teardown() {
 }
 
 @test "pid namespace mode target test" {
-	if [[ -n "$TEST_USERNS" ]]; then
+	if [[ -v TEST_USERNS ]]; then
 		skip "test fails in a user namespace"
 	fi
 	start_crio
--- a/test/config_migrate.bats
+++ b/test/config_migrate.bats
@@ -4,13 +4,13 @@
 load helpers
 
 @test "config migrate should succeed with default config" {
-	output=$(crio -c "" -d "" config -m 1.17 2>&1)
+	output=$("$CRIO_BINARY_PATH" -c "" -d "" config -m 1.17 2>&1)
 	[[ "$output" != *"Changing"* ]]
 }
 
 @test "config migrate should succeed with 1.17 config" {
 	# when
-	output=$(crio -c "$TESTDATA/config/config-v1.17.0.toml" -d "" config -m 1.17 2>&1)
+	output=$("$CRIO_BINARY_PATH" -c "$TESTDATA/config/config-v1.17.0.toml" -d "" config -m 1.17 2>&1)
 
 	# then
 	[[ "$output" == *'Changing \"apparmor_profile\" to \"crio-default\"'* ]]
@@ -25,7 +25,7 @@ load helpers
 
 @test "config migrate should fail on invalid version" {
 	# when
-	run crio -c "" -d "" config -m 1.16
+	run "$CRIO_BINARY_PATH" -c "" -d "" config -m 1.16
 
 	# then
 	[[ "$output" == *"unsupported migration version"* ]]
--- a/test/crio-wipe.bats
+++ b/test/crio-wipe.bats
@@ -22,9 +22,9 @@ function run_podman_with_args() {
 }
 
 function teardown() {
-	cleanup_test
 	run_podman_with_args stop -a
 	run_podman_with_args rm -fa
+	cleanup_test
 	cleanup_namespaces_dir
 }
 


test/cgroups.bats
```diff
@@ -65,8 +65,8 @@ function teardown() {
 	start_crio
 	# memsw should be greater than or equal to memory limit
 	# 210763776 = 1024*1024*200
-	jq '	  .linux.resources.memory_swap_limit_in_bytes = 209715200
-	    |     .linux.resources.memory_limit_in_bytes = 210763776' \
+	jq '	  .linux.resources.memory_swap_limit_in_bytes = 309715200
+	    |     .linux.resources.memory_limit_in_bytes = 310763776' \
```
haircommander, line 69: is this change intentional?

haircommander:
/approve

LGTM, thanks @klihub ! @cri-o/cri-o-maintainers PTAL

test/cgroups.bats
klihub: No. Just a forgotten verification hack after a painful 'No ! It surely can't be this'-sensation... Thanks for spotting. Reverted.

test/cgroups.bats
```diff
@@ -3,8 +3,8 @@
 load helpers
 
 function setup() {
-	newconfig="$TESTDIR/config.json"
 	setup_test
```
kolyshkin, line 6: Hmm. Since `setup_test` sets `$TESTDIR`, what do we have as its value before that?

If that's empty/unset, then maybe `set -u` would help

kolyshkin: Indeed it helps

```console
[kir@kir-rhat x]$ cat a.bats 
set -u

function setup_test() {
	TESTDIR=$(mktemp -d)
}

function setup() {
	dir="$TESTDIR"/xxx
	setup_test
}

function cleanup() {
	rm -f "$TESTDIR"
}

@test 1 {
	echo 1 $dir >&3
}
[kir@kir-rhat x]$ bats -t  a.bats 
1..1
not ok 1 1
# (from function `setup' in test file a.bats, line 8)
#   `dir="$TESTDIR"/xxx' failed
# /tmp/bats-run-5n7oz0/bats.1269087.src: line 8: TESTDIR: unbound variable
```

but it might uncover other issues (which, I guess, is good).

With `set -u` in place, we can remove the idiomatic check for `test_setup`.

klihub: So IIUC your suggestion is to add a `set -u` to `helpers.bash` to force it on every bats test script ?

I wonder how a sustainable idea that is ? With such a setup I *think* tests like `[ -n "$1" ]` and `[ -z "$2" ]` and evaluations like `foo "$@"` will start failing, despite looking like idiomatically correct shell expressions. They will need to be replaced with tautological expressions like `[ -n "${1:-}" ]` and `foo "${@:-}"`.

But I guess one way to find out if whether it is sustainable, is to try it out in practice. So I'll make that change and check if the fallout zone is of the size I can clean out.

fgiudici: Yeah, the shell expansion on $[1-9] will cause error if the parameter is not passed ($@ will be safe instead).
Having something able to catch more general errors (like the set -u) would be interesting indeed.

fgiudici (commented):
@klihub , nice catch, thank you!
Overall LGTM.
Seems to me the check-bats-setup.sh script will just check that the first occurrence of setup_test comes before the first usage of TESTDIR in the file. That may help, but will not check if this is true for all the tests: it will work only in the bats file where the setup_test is in the function_setup() (which honestly is always the case but for the apparmor.bats file). While this doesn't harm and may just help, having a more generic check may be a better option.

test/cgroups.bats
klihub: @fgiudici I'm testing now the `set -u` suggestion from @kolyshkin in practice, trying to go through the errors, understand where unbound arguments are due to intentional optionality, and fix those to work with `set -u` as well. It does not look too bad at first glance.

If I manage to get that working with a reasonably small change set, I will update this PR with that. I think it might be still a good idea to maybe put `set -u` behind an extra variable, so folks can do `make STRICT_MODE=true localintegration`, or the other way around if it's an opt-out... WDYT?

fgiudici: @klihub , nice!
The STRICT_MODE option is interesting... but I'm not sure it could be really useful.  In the end I think makes sense to have all the tests to comply to the STRICT_MODE or none of them: is there a reasonable use case to disable (or enable if disabled by default) the STRICT_MODE?

klihub: @fgiudic, @kolyshkin I've done a first round of set-u-ization of the helpers and tests. Although I tried to be careful enough to tell apart validly unset variables from strictly required ones, I might have made mistakes there. If you guys have the time/interest to look through the topmost patch (I'll update it if I find something more) to see if you spot something obviously wrong, any eyeballing is much appreciated.

hack/check-bats-setup.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+#
+# The idiomatic way to ensure that localintegarion/BATS tests don't
```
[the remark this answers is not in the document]
klihub: Abandoned the single-purpose script for the much saner `set -u` suggestion.

contrib/test/integration/test.yml
```diff
@@ -7,6 +7,18 @@
     regexp: ' go test \"\$'
     state: present
 
+- name: Print /usr/bin/bats version
+  shell: "/usr/bin/bats ] && /usr/bin/bats -v || echo no bats"
```
fgiudici:
```suggestion
  shell: "/usr/bin/bats && /usr/bin/bats -v || echo no bats"
```
Is the ']' a typo?

test/helpers.bash
```diff
@@ -555,23 +556,23 @@ function is_cgroup_v2() {
...
 EOF
 }
 
 function create_workload_with_allowed_annotation() {
-    local act="$2"
+    local act="${2:-}"
     # Fallback on the specified allowed annotation if
     # a specific activation annotation wasn't specified.
-    if [[ -z "$act" ]]; then
+    if [[ -z "${act:-}" ]]; then
```
fgiudici, line 575: Nit: we can drop the if clause and do:
act=${act:-$1}

test/helpers.bash
```diff
@@ -606,16 +607,16 @@ function check_conmon_cpuset() {
         fi
 
         found_cpuset=$(cat "$cpuset_path/pod_123-456/crio-conmon-$ctr_id/cpuset.cpus")
-        if [ -z "$cpuset" ]; then
+        if [ -z "${cpuset:-}" ]; then
```
fgiudici, line 610: Not needed (we ensure cpuset is defined above)

test/helpers.bash
```diff
@@ -606,16 +607,16 @@ function check_conmon_cpuset() {
...
             [[ $(cat "$cpuset_path/pod_123-456/cpuset.cpus") == *"$found_cpuset"* ]]
         else
             [[ "$cpuset" == *"$found_cpuset"* ]]
         fi
     else
         # don't test cpuset if it's not supported by systemd
-        if [[ -n "$systemd_supports_cpuset" ]]; then
+        if [[ -n "${systemd_supports_cpuset:-}" ]]; then
```
fgiudici, line 617: not needed, we defined systemd_supports_cpuset at the beginning of the function

test/helpers.bash
```diff
@@ -606,16 +607,16 @@ function check_conmon_cpuset() {
...
             info="$(systemctl show --property=AllowedCPUs crio-conmon-"$ctr_id".scope)"
-            if [ -z "$cpuset" ]; then
+            if [ -z "${cpuset:-}" ]; then
```
fgiudici, line 619: not needed

fgiudici (commented):
Thanks @klihub , the changes lgtm!
I found few places where we do double checks if the vars are set, but didn't spot any missing one, thank you!

test/helpers.bash
klihub: Thanks for spotting, fixed.

test/helpers.bash
klihub: Fixed.

test/helpers.bash
```diff
@@ -247,7 +248,7 @@ function setup_img() {
 
 function setup_crio() {
     apparmor=""
-    if [[ -n "$1" ]]; then
+    if [[ -n "${1:-}" ]]; then
```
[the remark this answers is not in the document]
klihub: Updated.

test/helpers.bash
klihub: Fixed.

test/critest.bats
```diff
@@ -3,7 +3,7 @@
 load helpers
 
 function setup() {
-	if [[ -z $RUN_CRITEST ]]; then
+	if [[ -z "${RUN_CRITEST:-}" ]]; then
```
kolyshkin, line 6: Perhaps `if [ -v RUN_CRITEST ]` is easier / more readable?

test/helpers.bash
```diff
@@ -123,6 +123,7 @@ for img in "${IMAGES[@]}"; do
 done
 
 function setup_test() {
+    set -u
```
kolyshkin, line 126: Maybe it's better to set it globally (i.e at the beginning of `helpers.bash`)

test/helpers.bash
```diff
@@ -373,8 +370,8 @@ function cleanup_pods() {
 }
 
 function stop_crio_no_clean() {
-    local signal="$1"
-    if [ -n "${CRIO_PID+x}" ]; then
+    local signal="${1:-}"
+    if [ -n "${CRIO_PID:-}" ]; then
```
kolyshkin, line 374: Same here, `-v CRIO_PID` seems better

test/helpers.bash
```diff
@@ -416,7 +413,7 @@ function cleanup_testdir() {
 }
 
 function cleanup_test() {
-    [ -z "$TESTDIR" ] && return
+    [ -z "${TESTDIR:-}" ] && return
```
kolyshkin, line 416: `[ ! -v TESTDIR ] && return`

or

`[ -v TESTDIR ] || return`

test/helpers.bash
```diff
@@ -555,25 +552,23 @@ function is_cgroup_v2() {
...
 default_runtime = "$NAME"
 [crio.runtime.runtimes.$NAME]
-runtime_path = "$RUNTIME_BINARY_PATH"
-runtime_root = "$RUNTIME_ROOT"
-runtime_type = "$RUNTIME_TYPE"
-allowed_annotations = ["$ANNOTATION"]
+runtime_path = "${RUNTIME_BINARY_PATH:-}"
+runtime_root = "${RUNTIME_ROOT:-}"
+runtime_type = "${RUNTIME_TYPE:-}"
+allowed_annotations = ["${ANNOTATION:-}"]
```
kolyshkin, line 563: Setting some parameters to empty values seem questionable (even if it was done the same way before).

So maybe do it conditionally instead?

Also, it seems weird that we are adding runtime without adding the path to its binary.

test/namespaces.bats
```diff
@@ -35,7 +35,7 @@ function teardown() {
 }
 
 @test "pid namespace mode target test" {
-	if [[ -n "$TEST_USERNS" ]]; then
+	if [[ -n "${TEST_USERNS:-}" ]]; then
```
kolyshkin, line 38: `-v TEST_USERNS`

kolyshkin:
@klihub thanks for working on this!

I think the order of commit should be different -- first you fix all cases where `set -u` result in unwanted error, and then you enable it.

Also, commits where you force bats symlinking and ignoring exit code from crictl need more explanation in commit messages.

contrib/test/integration/build/cri-o.yml
```diff
@@ -82,10 +82,9 @@
 
 - name: add quay.io and docker.io as default registries
   copy:
-    dest: /etc/crio/crio.conf.d/01-registries.conf
+    dest: /etc/containers/registries.conf
```
TomSweeneyRedHat, line 85: I've a vague memory that CRI-O wasn't supposed to use this file?  Bad memory, or have things changed?

klihub: I only stumbled upon it when combing through failed CI logs, saw this warning message, then git-grepped for it.

https://github.com/cri-o/cri-o/blob/93cdb4600d8a1332fcad08ea89ad6cf3b8f3ec2a/pkg/config/config.go#L631

test/helpers.bash
klihub: Actually, we don't even need to set this ourselves, if can only we make sure that we really run BATS test using the intended version (v1.3.0 ATM). Ever since commit 04ff588a BATS has set this itself.

I think it would be better to leave this unset here and instead add a BATS test to verify the version of BATS itself (probably forcing it to be the very first test.
WDYT?

test/critest.bats
klihub: Ah, thanks ! I didn't know such a test existed in `bash`. Looking at it, I guess 4.0.2 is safe/old enough to assume as the least version. Updated tests accordingly.

test/helpers.bash
klihub: True, exact same semantict. Updated.

test/helpers.bash
klihub: Yes, now that I actually tried to read the setup code paths, it looks to me as if all of those are always set.

test/helpers.bash
klihub: Actually we should just make sure that the nodes running CI are properly set up (`contrib/test/integration/main.yml` playbook successfully run with `setup` ansible tag) and this whole problem should go away. So I opted for to reverting the added BATS version verification test.

klihub:
> @klihub thanks for working on this!
> 
> I think the order of commit should be different -- first you fix all cases where `set -u` result in unwanted error, and then you enable it.
> 
> Also, commits where you force bats symlinking and ignoring exit code from crictl need more explanation in commit messages.

@kolyshkin I updated the link-forcing commit message explaining why it's needed. I also took out the `set -u` altogether. It should not be needed as long as we really run the tests with a proper version of BATS. A simple verification prior to running the tests can be added to the `Makefile` or `test/test_runner.sh` if deemed necessary.

kolyshkin:
> also took out the `set -u` altogether. It should not be needed as long as we really run the tests with a proper version of BATS.

I'm afraid this is not the case, i.e. bats does not do `set -u`:

```console
[kir@kir-rhat cli]$ bats  a.bats
 ✓ unset vars

1 test, 0 failures

[kir@kir-rhat cli]$ cat a.bats 
@test "unset vars" {
	unset ONE TWO THREE
	echo == $ONE $TWO $THREE ==
}
[kir@kir-rhat cli]$ bats ./a.bats 
 ✓ unset vars

1 test, 0 failures

[kir@kir-rhat cli]$ bats --version
Bats 1.5.0
```

At least this is not the case for bats 1.5.0. Are you seeing a different picture, @klihub?

klihub:
> > also took out the `set -u` altogether. It should not be needed as long as we really run the tests with a proper version of BATS.
> 
> I'm afraid this is not the case, i.e. bats does not do `set -u`:
> 
> ```
> [kir@kir-rhat cli]$ bats  a.bats
>  ✓ unset vars
> 
> 1 test, 0 failures
> 
> [kir@kir-rhat cli]$ cat a.bats 
> @test "unset vars" {
> 	unset ONE TWO THREE
> 	echo == $ONE $TWO $THREE ==
> }
> [kir@kir-rhat cli]$ bats ./a.bats 
>  ✓ unset vars
> 
> 1 test, 0 failures
> 
> [kir@kir-rhat cli]$ bats --version
> Bats 1.5.0
> ```
> 
> At least this is not the case for bats 1.5.0. Are you seeing a different picture, @klihub?

Hmm, no, unfortunately I do not. You are right, BATS only sets -u for itself, not for the tests being executed. So we do need to set it ourselves.

haircommander:
/retest-required

relevant tests look happy to me

/approve

I tag @kolyshkin for the final LGTM

haircommander:
thanks for working through this @klihub

test/helpers.bash
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+set -u
+
+# Work around skip() in old BATS versions not handling substitutions properly
+# wrt. 'set -u'. (fixed in commit 9b2b659e5ba769752c0d00dadb3ad9b58f59cbaa).
+# This can be removed once all CI nodes have been setup with new enough BATS.
```
kolyshkin, line 7: would be nice to specify the version here (is it v1.3.0?)

test/helpers.bash
```diff
@@ -1,5 +1,12 @@
...
+BATS_TEARDOWN_STARTED=${BATS_TEARDOWN_STARTED:-}
```
kolyshkin, line 8:
```suggestion
: "${BATS_TEARDOWN_STARTED:=}"
```

kolyshkin:
@klihub have you found where in CI we still have old bats?

klihub:
> @klihub have you found where in CI we still have old bats?

At least these are running older versions:
  - ci/openshift-jenkins/integration_crun_cgroupv2: ++ time bats --jobs 16 --tap . (using Bats 1.2.0-dev)
  - ci/openshift-jenkins/integration_crun : ++ time bats --jobs 16 --tap . (using Bats 1.2.1)
  - ci/openshift-jenkins/integration_fedora: ++ time bats --jobs 16 --tap . (using Bats 1.2.1)

I suspect ci/openshift-jenkins/integration_rhel is the same, but looks like this time it failed earlier so the corresponding logs were not collected.

test/helpers.bash
klihub: Yes, AFAICT the earliest tagged version with that fix/commit is `v1.3.0`. `git` describes it as `v1.2.1-17-g9b2b659` and there are no other tagged versions between `v1.2.1` and `v1.3.0`.

@kolyshkin, I can update that comment with the version info. That said, do we want to merge such a commit ? Or would it be better to instead change how tests are run so that we check BATS version, and clone/install the correct version if we find a too old (or just a different) one ?

test/helpers.bash
klihub: I don't mind updating it like that per se (although I have to say that writing it like would not come to me naturally). However, wouldn't that change be inconsistent with the rest of the `test/*.ba{sh,ts}` code which AFAICT currently uses the `:-`-notation everywhere to set up default/fallback values ?

kolyshkin:
@klihub can you please rebase this (ideally once https://github.com/cri-o/cri-o/pull/5663 is merged)?

klihub:
> @klihub can you please rebase this (ideally once #5663 is merged)?

@kolyshkin Rebased on latest `main`, #5663 included.

klihub:
This has been split up to multiple PRs and merged piecemeal. I think everything present here has been covered by those. Closing this accordingly...

PR #5610, opened 2022-02-08
server: filter sysctls passed through CRI if host namespace

haircommander:
#### What type of PR is this?

/kind bug

fixes https://github.com/cri-o/cri-o/security/advisories/GHSA-w2j5-3rcx-vx7x

```release-note
Fix a bug where a pod given a host IPC or network namespace could configure sysctls on the host
```

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -1018,7 +1018,7 @@ func (s *Server) configureGeneratorForSysctls(ctx context.Context, g *generate.G
 
 	for _, sysctl := range defaultSysctls {
 		if err := sysctl.Validate(hostNetwork, hostIPC); err != nil {
-			log.Warnf(ctx, "Skipping invalid sysctl %s: %v", sysctl, err)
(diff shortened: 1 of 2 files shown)


haircommander:
running CI here to make sure nothing unexpected fails

giuseppe (approved):
LGTM

although I am not sure about the security implications.  For using a host namespace, the container is already privileged.  AFAIK, runc has already the same check and it was added to the last release of crun as well.

fgiudici (approved):
/lgtm

haircommander:
> although I am not sure about the security implications. For using a host namespace, the container is already privileged. AFAIK, runc has already the same check and it was added to the last release of crun as well.

yeah I waiver between this being moderate and low severity. most kubernetes clusters don't do anything to prevent users from using host namespaces (until updated PSP, which is fairly new), which makes it more nefarious for upstream consumers (OCP has SCC, which mitigates this). However, the underlying principle is that a container affects the host in a way that wasn't really intended, which does feel like a container breakout

PR #5624, opened 2022-02-11
Multi storage support and selection using the runtime class

### LINKED ISSUE #5578: [RFE] Select the storage driver based on the workload/runtime type
**Description**

Today, cri-o supports a single storage driver at the time and the suggested/default is overlayfs for standard containers. For VM-based runtimes, the current solution is to share the container filesystem mount point using [9p](https://www.kernel.org/doc/Documentation/filesystems/9p.txt) or [virtiofs](https://virtio-fs.gitlab.io/). However, this doesn't offer the best performance, and passing a disk would be the ideal solution.

Cri-o is already able to identify and choose different runtime types based on the selected runtime class. We could use this information and select the best storage driver for each runtime type. The required changes in cri-o are the ability to enable multiple storage drivers at the same time and select one based on the runtime class.

The benefits of this enhancement are that we avoid the conflicts between standard and VM-based containers. We could get the best performance for both kind of workloads. Giving an example, kata-container could use device-mapper instead of overlays for passing the container filesystem as a device.

In the long term, this extension could also facilitate the introduction of new storage driver types for confidential computing. These new technologies come with a lot of challenges and requirements that filesystem-based storage drivers probably cannot  satisfy.

I'd like to collect any feedback and ideas and check if the community would be open to such a change :)

alicefr:
#### What type of PR is this?
/kind feature

Extend cri-o to configure multiple storage drivers at the same time and select the storage driver based on the runtime.

Today, cri-o supports a single storage driver at the time and the suggested/default is overlayfs for standard containers. For VM-based runtimes, the current solution is to share the container filesystem mount point using [9p](https://www.kernel.org/doc/Documentation/filesystems/9p.txt) or [virtiofs](https://virtio-fs.gitlab.io/). However, this doesn't offer the best performance, and passing a disk would be the ideal solution.

The benefits of this enhancement are that we avoid the conflicts between standard and VM-based containers. We could get the best performance for both kind of workloads. Giving an example, kata-container could use device-mapper instead of overlays for passing the container filesystem as a device.

In the long term, this extension could also facilitate the introduction of new storage driver types for confidential computing. These new technologies come with a lot of challenges and requirements that filesystem-based storage drivers probably cannot satisfy.

Fixes #5578 

Missing pieces:
  * if we download the same image for different storage drivers, it is listed multiple times? Do we need to filter it
  * Display the storage information for image inspection

Right now, I've been testing these changes by:
Starting cri-o with overlay and devicemapper:
```bash
$ bin/crio -l debug \
	--internal-wipe \
	--root /var/lib/containers-test/storage \
	--runroot /var/lib/containers-test/storage \
	--cgroup-manager systemd \
	--runtimes runc0:/usr/bin/runc:/run/runc0:oci:false \
	--runtime-storage runc0:devicemapper:dm.directlvm_device=/dev/loop0:dm.fs=ext4:dm.directlvm_device_force=true

```

```release-note
Support for multi storage drivers and selection using the runtimeclass
```


haircommander:
/ok-to-test

thanks @alicefr !

alicefr:
Update PR with 2 new interfaces for the multi store support. The unit test are passing locally. However, I needed to put a couple of AnyTimes(). So probably we want to get rid of those but at least I wanted to check the entire testsuite. Now, I'd like to check what is passing/failing in the upstream CI

alicefr:
I'm using inline tables for the `storage_option` because they are correctly parsed with toml, but I'd like to rework that and use a custom type. The array of storage_driver and the mapping for the storage_option is also redundant and it can be unified. Maybe some table like:
```toml
[crio.storage.overlay]  
 options = ["overlay.mountopt=nodev,metacopy=on","overlay.mountopt=nodev,metacopy=on"] 
[crio.storage.devicemapper]  
 options = ["dm.directlvm_device=/dev/loop0", "dm.fs=ext4", "dm.directlvm_device_force=true"]
```

contrib/test/integration/build/cri-o.yml
```diff
@@ -78,7 +78,8 @@
     dest: /etc/crio/crio.conf.d/01-overlay.conf
     content: |
       [crio]
-      storage_driver = "overlay"
+      default_storage = "overlay"
+      storage_driver = [ "overlay" ]
```
saschagrunert, line 82: Do we wanna make the change backwards compatible, like keeping `storage_driver` and add something like `additional_storage_drivers`?

internal/storage/multi_store.go
```diff
@@ -0,0 +1,333 @@
...
+	return iserver
+}
+
+func (s *multiStoreServer) GetStore() MultiStore {
+	return s.multiStore
+}
+
+func (s *multiStoreServer) ListAllImages(ctx *ctypes.SystemContext, filter string) ([]ImageResult, error) {
+	var imageResults []ImageResult
+	var lastError error
```
saschagrunert, line 62: nit:
```suggestion
func (s *multiStoreServer) ListAllImages(ctx *ctypes.SystemContext, filter string) (imageResults []ImageResult, lastError error) {
```

internal/storage/multi_store.go
```diff
@@ -0,0 +1,333 @@
...
+	return graphOptions
+}
+
+func (s *multiStore) Containers() (containers []cstorage.Container, lastError error) {
+	for _, v := range s.store {
+		c, err := v.Containers()
+		// TODO afrosi handles multiple errors
+		if err != nil {
+			lastError = err
+		}
```
saschagrunert, line 261: We can just wrap the errors with errors.Wrap or `fmt.Errorf("my error: %w", err)`

internal/storage/multi_store.go
```diff
@@ -0,0 +1,333 @@
...
+	return s.multiStore.GetDefaultStorage()
+}
+
+// GetImageServer returns the ImageServer for the selected storage driver. If the driver is empty or not found, the default storage driver is returned
+func (s *multiStoreServer) GetImageServer(driver string) ImageServer {
+	iserver, ok := s.store[driver]
+	if !ok {
+		v := s.store[s.multiStore.GetDefaultStorageDriver()]
+		return v
+	}
```
saschagrunert, line 52: I'm not sure if we should error here, at least if the not found case which should be impossible.

saschagrunert (commented):
Great work @alicefr! I think the way how we lookup the store is fine and integrates into the concept of the container storage. I would just have a few review nits later on if @haircommander agrees that the overall architecture is fine.

I'm now wondering how it impacts performance. Maybe we can do the e2e tests where the main storage driver is the second or third in a list of storage drivers and then compare the overall time of the test suites to get a rough idea?

alicefr:
> I'm now wondering how it impacts performance. Maybe we can do the e2e tests where the main storage driver is the second or third in a list of storage drivers and then compare the overall time of the test suites to get a rough idea?

Good idea! I can try this. Otherwise, we could implement a custom map type where the default storage is always the first to be checked

contrib/test/integration/build/cri-o.yml
alicefr: yes, this is important! I can leave `storage_driver` and `storage_opt`  and use this as the default and add what you proposed as a table:
```toml
[crio]
storage_driver = "overlay"
storage_opt = ["overlay.mountopt=nodev,metacopy=on","overlay.mountopt=nodev,metacopy=on"]

[crio.additional_storage.devicemapper]
options = ["dm.directlvm_device=/dev/loop0", "dm.fs=ext4", "dm.directlvm_device_force=true"]
[crio.additional_storage.btrfs]


```
Same for the crio cli flags.

internal/criocli/criocli.go
```diff
@@ -1033,3 +1042,69 @@ func StringSliceTrySplit(ctx *cli.Context, name string) []string {
 
 	return values
 }
+
+func splitMapOptions(v string) (key string, values []string, err error) {
```
haircommander, line 1046: nit, can this function be defined below the function that calls it so it can be read top to bottom

internal/lib/container_server.go
```diff
@@ -37,8 +38,7 @@ const ContainerManagerCRIO = "cri-o"
 // ContainerServer implements the ImageServer
 type ContainerServer struct {
 	runtime              *oci.Runtime
-	store                cstorage.Store
-	storageImageServer   storage.ImageServer
+	storageImageServer   storage.MultiStoreServer
```
haircommander, line 41: hm what was the motivation behind calling it storageImageServer? I would imagine it would be something like "stores" or "multiStore"

test/testdata/container_storage_sleep.json
```diff
@@ -0,0 +1,49 @@
...
+	},
+	"image": {
+		"image": "quay.io/crio/redis:alpine",
+		"annotations": {
+			"io.kubernetes.cri-o.Storage": "devicemapper"
+		}
+	},
+        "annotations": {
+                "io.kubernetes.cri-o.Storage": "devicemapper"
+        },
```
haircommander, line 13: nit; indenting here is off

haircommander: second nit: in server/sandbox_run_linux.go, we processed the annotation for the Storage. however, here we're adding the annotation on the container level. I would think we'd need it on the sandbox level

haircommander: third nit (sorry for not compressing into one comment and instead spamming :sweat_smile: ): there was a push to pair down the number of json files here and instead use jq directly in the bats suite to change the information in the default files. What motivation do you have for wanting this example that wouldn't be solved by jq in the bats suite

haircommander:
couple of nits, sorry I forgot to review yesterday!

I find myself wondering if processesing the annotaiton on the pod level is the best route. this seems like something an admin would want to specify in the runtime class or workload structure, not have be dynamic. what motivation is there for allowing a user to specify an annotation on their pod to be in a new runtime class?

generally though the structure looks good to me!

alicefr:
> couple of nits, sorry I forgot to review yesterday!
> 
> I find myself wondering if processesing the annotaiton on the pod level is the best route. this seems like something an admin would want to specify in the runtime class or workload structure, not have be dynamic. what motivation is there for allowing a user to specify an annotation on their pod to be in a new runtime class?
> 
> generally though the structure looks good to me!

Many thanks for the review, I'll rework next week :) .

But the annotation needs to specify per pod and the admin doesn't know this. This is similar to what we do when we specify the runtime class on the pod definition. Of course, we don't have this in the k8s API so we need to use the annotation. Does this answer the question?

haircommander:
> But the annotation needs to specify per pod and the admin doesn't know this. This is similar to what we do when we specify the runtime class on the pod definition. Of course, we don't have this in the k8s API so we need to use the annotation. Does this answer the question?

yeah! I think it would be best if this was accessed per-runtime class or per-workload. workloads can be triggered by an annotation, so it still subverts the kube api, but they have to be configured by the admin.

runtime class has the advantage of being part of what we're trying to do here--separate storage for pods in vm and oci runtime types. If each runtime class also could have a storage driver then all a user would have to do is specify the runtime class and cri-o would handle the rest. I think this would be better than having the pod author need to specify an annotation each time

alicefr:
> > But the annotation needs to specify per pod and the admin doesn't know this. This is similar to what we do when we specify the runtime class on the pod definition. Of course, we don't have this in the k8s API so we need to use the annotation. Does this answer the question?
> 
> yeah! I think it would be best if this was accessed per-runtime class or per-workload. workloads can be triggered by an annotation, so it still subverts the kube api, but they have to be configured by the admin.
> 
> runtime class has the advantage of being part of what we're trying to do here--separate storage for pods in vm and oci runtime types. If each runtime class also could have a storage driver then all a user would have to do is specify the runtime class and cri-o would handle the rest. I think this would be better than having the pod author need to specify an annotation each time

I need to find a solution for the image pulling. Right now, I don't where to find the runtime information during this phase :).

haircommander:
> I need to find a solution for the image pulling. Right now, I don't where to find the runtime information during this phase :).

it's a bit roundabout, but the image pull call passes in the pod sandbox config, which we could then use to find the pod sandbox, which we can use to find the runtime class. I thought image pull passed runtime handler but I was wrong

alicefr:
@haircommander @saschagrunert maybe you have again some good hints for me :). 
Currently, I need to loop over all the sandboxes to compare the sanbox name from the [PodSandboxConfig](https://github.com/cri-o/cri-o/blob/main/vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go#L6371) in order to find the whole [PodSandbox](https://github.com/cri-o/cri-o/blob/main/vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go#L2084) configuration where the runtime information is located. The main issue is that the PullRequest only contains the PodSandboxConfig. This doesn't include the PodSandbox ID that is required to quickly retrieve the sandbox information.

saschagrunert:
> or is it the function [`PodIDForName`](https://github.com/cri-o/cri-o/blob/main/internal/lib/container_server.go#L530) the right one to retrieve this information

Yeah this should return the ID.

alicefr:
hi @saschagrunert @haircommander , I want to implement some functional tests and I need to configure a second storage driver. So far, I tested on my laptop using overlay as default and devicemapper with a loopback device. I'm not quite sure if this is possible on cri-o ci.

alicefr:
I'm not sure how much this is complicated to put in place, but maybe we could add add an additional lane to the CI for testing the multi-storage with runtime configuration. So far, I have been able to run critest using this script:
```bash
#!/bin/bash

set -x

function setup_storage() {
    ls $1
    if [ $? -ne 0 ]; then 
    	fallocate -l 5G $1
    fi
    
    DEVICE=$(losetup -l | grep "$1" | awk '{print $1}')
    if [ -z "$DEVICE" ]; then 
    	sudo losetup -f --show $1
    fi
    if [ -z $DEVICE]; then 
	echo "failed in setting the device"
	exit 1
    fi
}

function clean_up {
    # Remove eventual lefover shim
    sudo killall -9 conmon
    # Unmount any devicemapper mount points in order to safely remove the snapshots
    sudo umount $(sudo  df -Th | grep devicemapper | awk '{print $7}')
    # Remove the snapshots for the containers
    sudo dmsetup remove $(sudo dmsetup ls | grep container | awk '{print $1}')
    sudo vgremove -f storage
    sudo pvremove $DEVICE
}

IMAGE=/tmp/test.img
export CONTAINER_STORAGE_DRIVER=overlay
export CONTAINER_RUNTIMES=runc0:/usr/bin/runc:/run/runc0:oci:false
export RUNTIME_TYPE=runc0

setup_storage $IMAGE

export CONTAINER_RUNTIME_STORAGE=runc0:devicemapper:dm.directlvm_device=$DEVICE:dm.fs=ext4:dm.directlvm_device_force=true

sudo -E PATH=$PATH make CRI_FOCUS="runtime should support removing running container" JOBS=1 RUN_CRITEST=1 TESTFLAGS=critest.bats localintegration

clean_up
```
and hacking the test/critest.bats in order to set the `--runtime-handler` flag
```diff
diff --git a/test/critest.bats b/test/critest.bats
index 50ddbed84..37e34785b 100644
--- a/test/critest.bats
+++ b/test/critest.bats
@@ -22,5 +22,6 @@ function teardown() {
                --image-endpoint "unix://${CRIO_SOCKET}" \
                --ginkgo.focus="${CRI_FOCUS}" \
                --ginkgo.skip="${CRI_SKIP}" \
-               --ginkgo.flakeAttempts=3 >&3
+               --ginkgo.flakeAttempts=3 >&3 \
+               --runtime-handler runc0
 }

```

saschagrunert:
Had a discussion with @alicefr on slack about the previous two comments. :)

alicefr:
@saschagrunert I've implemented a couple of functional tests. However, I need to use the `--with-pull` option (like [here](https://github.com/alicefr/cri-o/blob/multi-storage/test/multi_storage.bats#L71)) because the images are not prepulled for the additional storage. Is it something we want? Or is it fine to always pull the images during the test?

saschagrunert:
> @saschagrunert I've implemented a couple of functional tests. However, I need to use the `--with-pull` option (like [here](https://github.com/alicefr/cri-o/blob/multi-storage/test/multi_storage.bats#L71)) because the images are not prepulled for the additional storage. Is it something we want? Or is it fine to always pull the images during the test?

I think that's fine, avoiding network connections during tests is more an ideal rather than a must. :+1:

alicefr:
hi @saschagrunert and @haircommander! When you have a bit of time, could you please have a second look at this PR. I tried to address the issues from the previous review, cleaned, and simplified the code.
Modifications:
  * the storage is now selected based on the runtime only. In order to set the additional storage, the users need to configure it under the runtime (here I*m using `runc0` as additional runtime): 
```yaml
[crio.runtime.runtimes.runc0]
  storage_driver = "devicemapper"
  storage_option = [
          "dm.directlvm_device=/dev/loop0",
          "dm.fs=ext4",
          "dm.directlvm_device_force=true"
  ]
]
```
or from command line: `crio   --runtimes runc0:/usr/bin/runc:/run/runc0:oci:false  --runtime-storage runc0:devicemapper:dm.directlvm_device=/dev/loop0:dm.fs=ext4:dm.directlvm_device_force=true`
 * Added 2 interfaces for internal/storage `MultiStore` and `MultiStoreServer`. The implementation of the 2 interfaces uses an internal iterator that checks at first the default storage driver. 
 * Created unit tests for the new 2 multi-storage interfaces
 * Adjusted the already exiting unit tests with the new flow and mock functions (remove all the AnyTimes functions I added in the first version)
 * Add a new lane in the integration that run an integration test for the multi-store support with device-mapper configured

saschagrunert:
I'll revisit this PR later this week. :+1:

.github/workflows/integration.yml
```diff
@@ -4,8 +4,7 @@ on:
     tags:
       - v*
     branches:
-      - main
-      - release-*
+      - multi-storage
```
saschagrunert, line 7: I think this has to be reverted, right?

internal/lib/stats/stats_server.go
```diff
@@ -185,7 +185,11 @@ func (ss *StatsServer) writableLayerForContainer(container *oci.Container) (*typ
 		Timestamp: time.Now().UnixNano(),
 		FsId:      &types.FilesystemIdentifier{Mountpoint: container.MountPoint()},
 	}
-	driver, err := ss.Store().GraphDriver()
+	store, err := ss.Store().GetStoreForContainer(container.ID())
+	if err != nil {
+		return writableLayer, errors.Wrapf(err, "Unable to get store for container %s", container.ID())
```
saschagrunert, line 190: ultra nit, we start errors with lowercase usually:
```suggestion
		return writableLayer, errors.Wrapf(err, "unable to get store for container %s", container.ID())
```

saschagrunert (commented):
The overall approach looks good to me, great work @alicefr! I see a bunch of TODO left in the code, are you going to fix them as well?

@haircommander PTAL

.github/workflows/integration.yml
alicefr: yes, there is a commit that I need to remove. It is to have the CI running on this branch

alicefr:
> The overall approach looks good to me, great work @alicefr! I see a bunch of TODO left in the code, are you going to fix them as well?
> 
> @haircommander PTAL

Many thanks for the review!
Yes, there a couple of lines left for debugging the CI that I need to remove. There is also an additional commit for running the multi-storage integration test and I need to reintegrate 2 test2 I left.
However, there are 3 CI jobs failing that I don't quite understand. The `validate-docs`, `ci/prow/e2e-agnostic` and `
ci/prow/e2e-gcp`. The docs probably I'm doing something wrong because I'm generating the docs on fedora where selinux is enabled and I'm overwriting manually back to the default value. For the other 2 prow jobs, I've some difficulties in understanding what is going wrong. Any suggestions where I need to look?

docs/crio.8.md
```diff
@@ -449,4 +452,4 @@ Shows a list of commands or help for one command
 # SEE ALSO
 
 crio.conf(5), crio.conf.d(5), oci-hooks(5), policy.json(5), registries.conf(5),
-storage.conf(5)
\ No newline at end of file
+storage.conf(5)
```
saschagrunert: Can you please revert that change? This may fix the docs validation job.

alicefr: There are new options to the criocli for the runtime-storage, so I think they are necessary :/

saschagrunert: I mean only the newline at the end of the file.

alicefr: Not quite sure where it came from :/

alicefr:
/test e2e_fedora
--

alicefr:
There is an issue with the test ` Horizontal pod autoscaling (scale resource: CPU) ReplicationController light Should scale from 1 pod to 2 pods` for both failing lanes. I don't quite understand what is going wrong. Any help on this would be very much appreciated

docs/crio.8.md
```diff
@@ -328,7 +329,9 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--runroot**="": The CRI-O state directory (default: /run/containers/storage)
 
-**--runtimes**="": OCI runtimes, format is runtime_name:runtime_path:runtime_root:runtime_type:privileged_without_host_devices:runtime_config_path (default: [])
+**--runtime-storage**="": Storage for the runtime, format is runtime_name:storage_driver:opt1:opt2...:optN (default: [])
+
+**--runtimes**="": OCI runtimes, format is runtime_name:runtime_path:runtime_root:runtime_type:privileged_without_host_devices:runtime_config_path:storage_driver=opt1=value,opt2=value (default: [])
```
TomSweeneyRedHat, line 334: A little more verbosity for the descriptions would be good, along with an example or two would be helpful.  For instance, runtime_type, what are the valid values?

internal/criocli/criocli.go
```diff
@@ -130,15 +130,42 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
...
 		runtimes := StringSliceTrySplit(ctx, "runtimes")
+		storageRuntimeMap := make(map[string]storageOpts)
+		if ctx.IsSet("runtime-storage") {
+			for _, s := range StringSliceTrySplit(ctx, "runtime-storage") {
+				fields := strings.Split(s, ":")
+				if len(fields) < 2 {
+					return fmt.Errorf("invalid format runtime-storage: %s", s)
+				}
+				var opts []string
+				if len(fields) > 3 {
```
TomSweeneyRedHat, line 147: should this be?
```suggestion
				if len(fields) >= 3 {
```
Your testing prior for `< 2`

docs/crio.8.md
alicefr: @TomSweeneyRedHat thanks for the review! Do you mean adding some examples under `--runtime-storage` or `--runtimes` flag? At least the `runtime_type` you are mentioning isn't a new flag. But I just realized that new option  `storage_driver=opt1=value,opt2=value` in the runtime is useless, so the `--runtimes` will remain as it is.

internal/criocli/criocli.go
alicefr: You are perfectly right, thanks!

server/image_fs_info.go
```diff
@@ -32,13 +33,22 @@ func getStorageFsInfo(store storage.Store) (*types.FilesystemUsage, error) {
...
+	var lastError error
+	store := s.GetAllStores()
+	for _, s := range store {
+		fsUsage, err := getStorageFsInfo(s)
+		if err != nil {
+			if lastError == nil {
+				lastError = err
+			} else {
+				lastError = errors.Wrap(lastError, err.Error())
+			}
```
alicefr, line 46: I need to add a continue if the error is not nil

alicefr:
Rebase and replace the `errors.Wrap` with `fmt.Errorf`

alicefr:
> Looks like there are still some issues left open: https://github.com/cri-o/cri-o/runs/7088306948?check_suite_focus=true

Yes, I missed one `errors.Wrap` in the unit tests. Hopefully, now the unit and lint are happier :). The real issue is still the jobs `e2e-agnostic` and `e2e-gcp` with the test [`sig-autoscaling] [Feature:HPA] Horizontal pod autoscaling (scale resource: CPU) ReplicationController light Should scale from 1 pod to 2 pods`](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/5624/pull-ci-cri-o-cri-o-main-e2e-gcp/1541693611970662400) that is failing. I'm stuck in understanding what is the cause of this

saschagrunert:
> > Looks like there are still some issues left open: https://github.com/cri-o/cri-o/runs/7088306948?check_suite_focus=true
> 
> Yes, I missed one `errors.Wrap` in the unit tests. Hopefully, now the unit and lint are happier :). The real issue is still the jobs `e2e-agnostic` and `e2e-gcp` with the test [`sig-autoscaling] [Feature:HPA] Horizontal pod autoscaling (scale resource: CPU) ReplicationController light Should scale from 1 pod to 2 pods`](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/5624/pull-ci-cri-o-cri-o-main-e2e-gcp/1541693611970662400) that is failing. I'm stuck in understanding what is the cause of this :/

Isn't this a flake? I thought we recently fixed some test regarding HPA, but this looks like a new one.

alicefr:
> > > Looks like there are still some issues left open: https://github.com/cri-o/cri-o/runs/7088306948?check_suite_focus=true
> > 
> > 
> > Yes, I missed one `errors.Wrap` in the unit tests. Hopefully, now the unit and lint are happier :). The real issue is still the jobs `e2e-agnostic` and `e2e-gcp` with the test [`sig-autoscaling] [Feature:HPA] Horizontal pod autoscaling (scale resource: CPU) ReplicationController light Should scale from 1 pod to 2 pods`](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/5624/pull-ci-cri-o-cri-o-main-e2e-gcp/1541693611970662400) that is failing. I'm stuck in understanding what is the cause of this :/
> 
> Isn't this a flake? I thought we recently fixed some test regarding HPA, but this looks like a new one.

I don't think so because it was failing also a month ago before the rebase and it has always been the same test and it never passes. However, when we looked together last time, it simply saw that the second pod was not created at all without any meaningful messages. I need to look at the tests again more carefully. I did also ask for help the sig-autoscaling slack channel without any answers

alicefr:
@saschagrunert I have the suspicion that the test `[sig-autoscaling] [Feature:HPA] Horizontal pod autoscaling (scale resource: CPU) ReplicationController light Should scale from 1 pod to 2 pods` is failing because it fails to fetch the [CPU metrics](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/cri-o_cri-o/5624/pull-ci-cri-o-cri-o-main-e2e-agnostic/1541773731834105856/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-monitoring_prometheus-adapter-b885475d9-5svlj_prometheus-adapter.log). It doesn't get the CPU metrics so the replicaset isn't triggered to scale and create the second pod. Did you experience a similar issue in the past?

saschagrunert:
> @saschagrunert I have the suspicion that the test `[sig-autoscaling] [Feature:HPA] Horizontal pod autoscaling (scale resource: CPU) ReplicationController light Should scale from 1 pod to 2 pods` is failing because it fails to fetch the [CPU metrics](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/cri-o_cri-o/5624/pull-ci-cri-o-cri-o-main-e2e-agnostic/1541773731834105856/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-monitoring_prometheus-adapter-b885475d9-5svlj_prometheus-adapter.log). Therefore, the replicaset isn't triggered to scale and create the second pod. Did you experience a similar issue in the past?

Are those CPU metrics provided via the CRI?

saschagrunert:
> Good point. So, I had a look at the pod and the provider is failing [here](https://github.com/openshift/k8s-prometheus-adapter/blob/master/pkg/resourceprovider/provider.go#L191) because the container doesn't have the metrics set. Looking in the code, this basically this information comes from kubernetes API because it [lists all the pods in the namespace](https://github.com/openshift/k8s-prometheus-adapter/blob/master/vendor/sigs.k8s.io/metrics-server/pkg/api/pod.go#L84). I need to check from where k8s takes those values but I think you are right

There are two stats providers available in Kubernetes, the CRI stats provider and the cadvisor one: https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/stats

Based on https://github.com/cri-o/cri-o/pull/5926, we should now use the CRI stats provider per default. It may be possible that we broke something in this PR regarding that or we revealed a more generous bug.

alicefr:
It would make sense that it is a bug in this PR as I changed the `internal/lib/stats/stats_server.go`. I really need to reproduce the issue locally :/

alicefr:
@saschagrunert We might still rely on cadvisor as I can see in kubelet log the error and events `failed to get imageFs info: non-existent label "crio-images` and this correspond to this [error](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/cadvisor_stats_provider.go#L88). We already noticed this error in the past, but we haven't thought it could be the root cause. In fact, it can be. We don't get the sandbox  stats and the replica set doesn't create the second pod

alicefr:
ok, reading a bit cadvisor code, I'm worried that the mutli storage won't work. It seems they get the containers/pod information directly from the filesystem :/ not form CRI APIs. This still doesn't explain the failure for the test as I'm not configuring any additional storage. But we need to rethink if this is a feasible features :/

alicefr:
I think the bug comes from some wrong info sent by crio. Cadvisor has a crio client where [it should get the Root](https://github.com/google/cadvisor/blob/master/fs/fs.go#L299) this is empty and it cannot set the crio-images level and monitors the pod stats

saschagrunert:
Awesome, great work @alicefr!

saschagrunert:
Can you please fixup the docs validation as well?

alicefr:
@saschagrunert yes, I can there is some mismatch how the documentation is generated on my fedora laptop. However, I'd like to discuss about the problem with cadvisor I mentioned in  https://github.com/cri-o/cri-o/pull/5624#issuecomment-1171048390. I hope I can attend the crio meeting today

saschagrunert (approved):
LGTM

saschagrunert:
@haircommander that's a huge one so I prefer having multiple eyes here.

internal/storage/multi_store.go
```diff
@@ -0,0 +1,445 @@
...
+		return nil
+	}
+	if e, ok := i.multiStoreServer.store[i.keys[i.counter]]; ok {
+		i.counter++
+		return e
+	}
+	return nil
+}
+
+type MultiStoreServer interface {
```
haircommander, line 257: I think it would be useful to comment the distinction between the different interfaces. and generally what each of them do

internal/storage/runtime.go
```diff
@@ -343,19 +348,23 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
...
+func (r *runtimeService) CreateContainer(systemContext *types.SystemContext, podName, podID, imageName, imageID, containerName, containerID, metadataName string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, privileged bool, storageDriver string) (ContainerInfo, error) {
+	return r.createContainerOrPodSandbox(systemContext, podName, podID, imageName, "", imageID, containerName, containerID, metadataName, "", "", attempt, idMappingsOptions, labelOptions, false, privileged, storageDriver)
 }
 
 func (r *runtimeService) deleteLayerIfMapped(imageID, layerID string) {
 	if layerID == "" {
 		return
 	}
-	store := r.storageImageServer.GetStore()
+	store, err := r.multiStoreServer.GetStoreForImage(imageID)
```
haircommander, line 363: this is an archetectural point that I'm not sure is in scope here, but there seems to be 4 layers of abstraction to access the container information: runtimeService->multiStorageServer->multiStorage->store

Is there any way we can simplify this? i.e: maybe merging runtimeService with multiStorageServer? I am generally in favor of fewer abstractions unless there's a good reason to have them (and it's possible I'm missing said good reason that exists here)

server/image_pull.go
```diff
@@ -19,6 +20,36 @@ import (
 
 var localRegistryPrefix = "localhost/"
 
+func (s *Server) findStorageDriver(ctx context.Context, req *types.PullImageRequest) string {
```
haircommander, line 23: same as above, can this go below its last caller?

server/image_remove.go
```diff
@@ -22,9 +22,23 @@ func (s *Server) RemoveImage(ctx context.Context, req *types.RemoveImageRequest)
 	return s.removeImage(ctx, imageRef)
 }
 
+func (s *Server) untagImageForAllImageServers(img string) error {
+	iServer, err := s.StorageImageServerPerImage(img)
+	if err != nil {
+		return err
+	}
+	for _, is := range iServer {
```
haircommander, line 30: in many cases, there is a specific multiStore endpoint for this kind of batch iterator work, why not one here?

scripts/setup_storage
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
haircommander, line 1: I think I could use a comment here. what storage are we setting up?

haircommander:
a herculean effort @alicefr, well done! I have left some nits and architectural thoughts. I think I'd like @giuseppe , @kolyshkin and/or @nalind to take a look too. This is getting close, and I appreciate your work and patience!

internal/storage/multi_store.go
```diff
@@ -0,0 +1,445 @@
...
+// ListAllImages lists all the images known by the MultiStoreServer.
+func (s *multiStoreServer) ListAllImages(ctx *ctypes.SystemContext, filter string) (imageResults []ImageResult, lastError error) {
+	s.iterator.initialize()
+	for is := s.iterator.next(); is != nil; is = s.iterator.next() {
+		images, err := is.ListImages(ctx, filter)
+		if err != nil {
+			lastError = wrapMultipleErrors(lastError, err)
+			continue
+		}
+		imageResults = append(imageResults, images...)
```
giuseppe, line 316: do we need to care for duplicates?

internal/storage/runtime.go
```diff
@@ -343,19 +348,23 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
...
+	if err != nil {
+		logrus.Debugf("Failed to storage for  image %q: %v", imageID, err)
```
giuseppe, line 365: nit: two white spaces

internal/storage/runtime.go
giuseppe: is there a possibility the image is in multiple stores? In that case, shouldn't we delete it from them all instead of leaking some images or do we wait for the Kubelet to retrieve again the list and delete again?

internal/storage/multi_store.go
alicefr: @giuseppe it depends. Ilsted this issue under the special notes. There isn't a way how I could print the storage where the image is located. If I print only once then we don't get the information that we actually have 2 (or more) copies of the image. This is still an aspect I don't know how to solve in a nice way

internal/storage/runtime.go
alicefr: Yes, you are right. I think the function `GetStoreForImage` needs to return an array for Storage :)

alicefr:
> Is the iterator pattern (initialize(), next(), ...) being made safe somehow, or will two goroutines calling into the ListAllImages() method on the same multiStoreServer at the same time be a problem?

@nalind it is an interesting question. The iterator shouldn't particularly change the previous behavior for accessing the maps. This defines the order in how we access the values and every call to the functions instantiates a new iterator. So, I don't think it is a problem but please correct me if you have any concerns.

alicefr:
I added 2 commits that:
1. a4537e16fa23508097563f39fdd4d3e4ba5a40c4 : fix @giuseppe 's [comment](https://github.com/cri-o/cri-o/pull/5624#discussion_r914800770) about the image deletion
2. https://github.com/cri-o/cri-o/pull/5624/commits/576bd4108202df01549ebb7a905c62383b0ee040 : fix @nalind 's [comment](https://github.com/cri-o/cri-o/pull/5624#issuecomment-1176644801) about the iterator

The remaining points to solve for me are @haircommander 's [comment](https://github.com/cri-o/cri-o/pull/5624#discussion_r914214573) about the interfaces and @giuseppe 's [comment](https://github.com/cri-o/cri-o/pull/5624#discussion_r914795129) about the image listing. You can see my replies directly in the threads

docs/crio.8.md
```diff
@@ -328,6 +329,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--runroot**="": The CRI-O state directory (default: /run/containers/storage)
 
+**--runtime-storage**="": Storage for the runtime, format is runtime_name:storage_driver:opt1:opt2...:optN. The runtime_name needs to be listed with the  runtimes, otherwise this option doesn't have any effect.
```
TomSweeneyRedHat, line 332:
```suggestion
**--runtime-storage**="": Storage for the runtime, format is runtime_name:storage_driver:opt1:opt2...:optN. The runtime_name needs to be listed with the runtimes. Otherwise, this option doesn't have any effect.
```

alicefr:
@saschagrunert here are my findings about the new multi storage failure. I have the impression that the failure is due to the new update for containers/storage. I update the package in the [latest crio version](https://github.com/alicefr/cri-o/tree/update-storage). 
Start crio with devicemapper:
```bash
sudo -E PATH=$PATH bin/crio -l debug \
        --internal-wipe \
        --root /var/lib/containers-test/storage \
        --runroot /var/lib/containers-test/storage \
        -s devicemapper \
        --storage-opt dm.directlvm_device=/dev/loop1 \
        --storage-opt dm.fs=ext4 \
        --storage-opt dm.directlvm_device_force=true \
        --cgroup-manager systemd \
        --internal-wipe \
        --config $(pwd)/crio-custom.conf
```
Try to create a pod:
```bash
sudo crictl runp test/testdata/sandbox_config.json
```
Then, there is an error loop in crio:
```
 runSandbox: removing pod sandbox from storage: af0de4bbeb36bee781d75a62203c43618512a7926d3d92169db1ea66827cfe9a  file="resourcestore/resourcecleaner.go:69" id=463abb77-415d-4e2b-9a6b-4d526df645ce name=/runtime.v1.RuntimeService/RunPodSandbox
DEBU[2022-07-25 09:09:18.237857963+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:09:18.238558320+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:09:18.238667215+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:09:18.238732591+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:09:18.238928503+02:00] Failed to retrieve layer "4936c12c8c0a47b888a2522a34b2ddd81bb4c4eda3cb9cc333ff82a2d0bce0b5": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:395"
DEBU[2022-07-25 09:09:18.239037914+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:09:18.239593918+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:09:18.239673433+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:09:18.239729617+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:09:18.239867056+02:00] Failed to delete container "af0de4bbeb36bee781d75a62203c43618512a7926d3d92169db1ea66827cfe9a": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:399"
ERRO[2022-07-25 09:09:18.239957244+02:00] Failed to cleanup (probably retrying): unmarshaling previous device setup config: %!w(<nil>)  file="resourcestore/resourcecleaner.go:71" id=463abb77-415d-4e2b-9a6b-4d526df645ce name=/runtime.v1.RuntimeService/RunPodSandbox
INFO[2022-07-25 09:10:55.554566432+02:00] runSandbox: removing pod sandbox from storage: af0de4bbeb36bee781d75a62203c43618512a7926d3d92169db1ea66827cfe9a  file="resourcestore/resourcecleaner.go:69" id=463abb77-415d-4e2b-9a6b-4d526df645ce name=/runtime.v1.RuntimeService/RunPodSandbox
DEBU[2022-07-25 09:10:55.554816897+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:10:55.555737271+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:10:55.555873387+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:10:55.555980893+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:10:55.556288333+02:00] Failed to retrieve layer "4936c12c8c0a47b888a2522a34b2ddd81bb4c4eda3cb9cc333ff82a2d0bce0b5": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:395"
DEBU[2022-07-25 09:10:55.556519468+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:10:55.558143183+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:10:55.558352366+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:10:55.558485492+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:10:55.558762249+02:00] Failed to delete container "af0de4bbeb36bee781d75a62203c43618512a7926d3d92169db1ea66827cfe9a": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:399"
ERRO[2022-07-25 09:10:55.558932161+02:00] Failed to cleanup (probably retrying): unmarshaling previous device setup config: %!w(<nil>)  file="resourcestore/resourcecleaner.go:71" id=463abb77-415d-4e2b-9a6b-4d526df645ce name=/runtime.v1.RuntimeService/RunPodSandbox
 INFO[2022-07-25 09:13:21.524515067+02:00] runSandbox: removing pod sandbox from storage: af0de4bbeb36bee781d75a62203c43618512a7926d3d92169db1ea66827cfe9a  file="resourcestore/resourcecleaner.go:69" id=463abb77-415d-4e2b-9a6b-4d526df645ce name=/runtime.v1.RuntimeService/RunPodSandbox
DEBU[2022-07-25 09:13:21.524857586+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:13:21.525794149+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:13:21.525962804+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:13:21.526089025+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:13:21.526431094+02:00] Failed to retrieve layer "4936c12c8c0a47b888a2522a34b2ddd81bb4c4eda3cb9cc333ff82a2d0bce0b5": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:395"
DEBU[2022-07-25 09:13:21.526633930+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:13:21.527370219+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:13:21.527560615+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:13:21.527704888+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:13:21.527988904+02:00] Failed to delete container "af0de4bbeb36bee781d75a62203c43618512a7926d3d92169db1ea66827cfe9a": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:399"
ERRO[2022-07-25 09:13:21.528166060+02:00] Failed to cleanup (probably retrying): unmarshaling previous device setup config: %!w(<nil>)  file="resourcestore/resourcecleaner.go:71" id=463abb77-415d-4e2b-9a6b-4d526df645ce name=/runtime.v1.RuntimeService/RunPodSandbox

```

alicefr:
@saschagrunert here are my finding about the new failure in the multi store lane. AFAIU, the new failure is due to the update in the containers/storage. I updated the package in the [latest crio version](https://github.com/alicefr/cri-o/tree/update-storage) and then reproduced in this way:
Start crio with devicemapper:
```bash
sudo -E PATH=$PATH bin/crio -l debug \
        --internal-wipe \
        --root /var/lib/containers-test/storage \
        --runroot /var/lib/containers-test/storage \
        -s devicemapper \
        --storage-opt dm.directlvm_device=/dev/loop1 \
        --storage-opt dm.fs=ext4 \
        --storage-opt dm.directlvm_device_force=true \
        --cgroup-manager systemd \
        --internal-wipe
```
Try to create a pod:
```bash
sudo crictl runp test/testdata/sandbox_config.json
```
Then, you get:
```
DEBU[2022-07-25 09:18:55.913282153+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:18:55.913448104+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:18:55.913518646+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:18:55.913685050+02:00] Failed to delete container "ff88e3cbc3fab33a4a7f6ca768669805dcdddacb03a1057d250adfa2e6514ef7": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:399"
ERRO[2022-07-25 09:18:55.913783213+02:00] Failed to cleanup (probably retrying): unmarshaling previous device setup config: %!w(<nil>)  file="resourcestore/resourcecleaner.go:71" id=2a53ebcf-897b-4985-878e-a6f6d78a3808 name=/runtime.v1.RuntimeService/RunPodSandbox
INFO[2022-07-25 09:19:39.163711915+02:00] runSandbox: removing pod sandbox from storage: ff88e3cbc3fab33a4a7f6ca768669805dcdddacb03a1057d250adfa2e6514ef7  file="resourcestore/resourcecleaner.go:69" id=2a53ebcf-897b-4985-878e-a6f6d78a3808 name=/runtime.v1.RuntimeService/RunPodSandbox
DEBU[2022-07-25 09:19:39.163965639+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:19:39.164607756+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:19:39.164710315+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:19:39.164777340+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:19:39.164964270+02:00] Failed to retrieve layer "30c883c7c00809b626d9a06b4b2610af6a11af1b897b416443435b5dc6a37942": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:395"
DEBU[2022-07-25 09:19:39.165078930+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:19:39.165699088+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:19:39.165788470+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:19:39.165853151+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:19:39.165999751+02:00] Failed to delete container "ff88e3cbc3fab33a4a7f6ca768669805dcdddacb03a1057d250adfa2e6514ef7": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:399"
ERRO[2022-07-25 09:19:39.166097878+02:00] Failed to cleanup (probably retrying): unmarshaling previous device setup config: %!w(<nil>)  file="resourcestore/resourcecleaner.go:71" id=2a53ebcf-897b-4985-878e-a6f6d78a3808 name=/runtime.v1.RuntimeService/RunPodSandbox
INFO[2022-07-25 09:20:44.041624727+02:00] runSandbox: removing pod sandbox from storage: ff88e3cbc3fab33a4a7f6ca768669805dcdddacb03a1057d250adfa2e6514ef7  file="resourcestore/resourcecleaner.go:69" id=2a53ebcf-897b-4985-878e-a6f6d78a3808 name=/runtime.v1.RuntimeService/RunPodSandbox
DEBU[2022-07-25 09:20:44.041935384+02:00] [graphdriver] trying provided driver "devicemapper"  file="drivers/driver.go:324"
DEBU[2022-07-25 09:20:44.042825947+02:00] devicemapper: kernel dm driver version is 4.46.0  file="devmapper/deviceset.go:1505"
DEBU[2022-07-25 09:20:44.042956452+02:00] devmapper: Deferred removal support enabled.  file="devmapper/deviceset.go:1671"
DEBU[2022-07-25 09:20:44.043025150+02:00] devmapper: Deferred deletion support enabled.  file="devmapper/deviceset.go:1679"
DEBU[2022-07-25 09:20:44.043295404+02:00] Failed to retrieve layer "30c883c7c00809b626d9a06b4b2610af6a11af1b897b416443435b5dc6a37942": unmarshaling previous device setup config: %!w(<nil>)  file="storage/runtime.go:395"
```

alicefr:
@saschagrunert I think we need to to the [same fix](https://github.com/containers/storage/pull/1294/files) you did for the function [readLVMConfig](https://github.com/containers/storage/blob/main/drivers/devmapper/device_setup.go#L170)

saschagrunert:
> @saschagrunert I think we need to do the [same fix](https://github.com/containers/storage/pull/1294/files) you did for the function [readLVMConfig](https://github.com/containers/storage/blob/main/drivers/devmapper/device_setup.go#L170)
> 
> EDIT: Posted fix in containers/storage: [containers/storage#1298](https://github.com/containers/storage/pull/1298)

Awesome, thank you for fixing this one!

alicefr:
@nalind and @giuseppe could you please have a second look at this PR when you have some time?
@giuseppe I still haven't addressed your comment about the image list, currently, if an image is present in multiple drivers, we list the same image for all the storage drivers.

go.sum
```diff
@@ -1617,7 +1617,10 @@ github.com/klauspost/compress v1.14.1/go.mod h1:/3/Vjq9QcHkK5uEr5lBEmyoZ1iFhe47e
 github.com/klauspost/compress v1.15.1/go.mod h1:/3/Vjq9QcHkK5uEr5lBEmyoZ1iFhe47etQ6QUkpK6sk=
 github.com/klauspost/compress v1.15.4/go.mod h1:PhcZ0MbTNciWF3rruxRgKxI5NkcHHrHUDtV4Yw2GlzU=
 github.com/klauspost/compress v1.15.7/go.mod h1:PhcZ0MbTNciWF3rruxRgKxI5NkcHHrHUDtV4Yw2GlzU=
+<<<<<<< HEAD
 github.com/klauspost/compress v1.15.8/go.mod h1:PhcZ0MbTNciWF3rruxRgKxI5NkcHHrHUDtV4Yw2GlzU=
+=======
+>>>>>>> b172b4587 (update conainers/storage)
```
haircommander, line 1623: looks like we need to re-run `make vendor`

alicefr: oh sorry, messed up with the rebase

alicefr: Fixed

alicefr:
/test ci-integration
--

alicefr:
I need to fix the checkpoint/restore support

alicefr:
Unfortunately, I'm not working on this anymore. However, I think in the future we might need to be able to differentiate the storage driver based on the runtime or kind of workload and this PR can be used as reference.

Closing it.

PR #5626, opened 2022-02-11
image: use imageCache value for ImageStatus()

haircommander:
#### What type of PR is this?

/kind bug
Otherwise, we repopulate the image cache each time.
This reduces the number of attempts to do a full unlock of the LayerStore

```release-note
none
```


haircommander:
some context here: on a very overloaded node, I see cri-o holding open a ton of goroutines (in the order of thousands) being stuck on the LayerStore lock. Looking into it, there are two cases we seem to get stuck on:
- we attempt to unlock the layer store each image status, which seems excessive since we have an image cache, and since we populate that image cache each time we call ListImages. 
- a lot of churn of container creations cause us to create and remove container rootfs's

this PR attempts to fix the first case, to give the second case more room to do the cleanups

mtrmac (commented):
Code LGTM, but it’s a bit tricky.

- the values in `imageCacheItem` are immutable properties of a single image (= single storage _ID_ like `@{image.ID}` (assuming the same `SystemContext`, e.g. architecture choice)
  - Note that `ImageResult` is _not_ immutable, `image.Names` can change by pulling
  - Note that other ways to refer to images, like _name_ like `ubi8:latest` (which seem to be acceptable inputs to `ImageStatus`) are _not_ immutable, the names can point to a different image after a fresh pull)
- The key of `imageCache` is the image ID. So this is safe.

… Actually, once we compute an `imageCache` entry for an image ID, AFAICS we don’t need to rebuild it ever again. So it might possible to speed up `ListImages` quite a bit by preserving pre-existing cache entries (and only updating for added/removed images).

---

To help preserve these properties, could you add a comment saying something like ”The contents of imageItem are fixed properties of an image (storage's image.ID), and never need recomputing” to the documentation of the imageCacheItem` type?

haircommander:
> To help preserve these properties, could you add a comment saying something like ”The contents of imageItem are fixed properties of an image (storage's image.ID), and never need recomputing” to the documentation of the imageCacheItem` type?

added!

> So it might possible to speed up ListImages quite a bit by preserving pre-existing cache entries (and only updating for added/removed images).

It looks like we currently do this. we check for a cache hit in appendCachedResult after we've listed images from storage

mtrmac (commented):
LGTM.

mrunalp (approved):
/lgtm

PR #5639, opened 2022-02-14
contrib/bundle: remove deprecated kubelet option.

klihub:
#### What type of PR is this?

/kind failing-test

Don't pass the deprecated `--non-masquerade-cidr` command line option to `kubelet`. It was recently removed altogether and now breaks tests in CI.

None

```release-note
None
```

### DIFF
--- a/contrib/bundle/test-e2e
+++ b/contrib/bundle/test-e2e
@@ -86,7 +86,7 @@ cp kubernetes/test/bin/{ginkgo,e2e.test} _output/bin
 export CONTAINER_RUNTIME=remote
 export CGROUP_DRIVER=systemd
 export CGROUP_ROOT=/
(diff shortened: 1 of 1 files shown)


fidencio:
LGTM, thanks @klihub!

A nit (which I do **not** think is worth re-spinning the PR because of that), would be nice to have a reference for the deprecation, even if as a comment in this PR.

klihub:
> LGTM, thanks @klihub!
> 
> A nit (which I do **not** think is worth re-spinning the PR because of that), would be nice to have a reference for the deprecation, even if as a comment in this PR.

Kubernetes PR https://github.com/kubernetes/kubernetes/pull/107096, commit https://github.com/kubernetes/kubernetes/commit/21809043b5af2469a7fd76e173ade1a2caf80ba0

PR #4385, opened 2020-11-19
test/e2e: disable a flaky test

kolyshkin:
/kind ci
/kind flake
/release-note-none

#### What this PR does / why we need it:

This is failing a lot recently. There is an open issue [1],
but let's disable this one until the issue is fixed.

[1] https://github.com/kubernetes/kubernetes/issues/96565

PR #5643, opened 2022-02-15
test/crio-wipe: fixups

kolyshkin:
Commit 6c69b44953 remove bats "run" from `run_podman_with_args`, meaning
variables $output and $status are no longer set after using it.

Commit f48f90d365 added a new test case which checks $output. Most
probably the $output is not even set, so the check isn't working.

Fix by using podman container exists which is the most straightforward
check.

Use the same check in `start_crio_with_stopped_pod`.

```release-note
NONE
```

### DIFF
--- a/test/crio-wipe.bats
+++ b/test/crio-wipe.bats
@@ -128,7 +128,7 @@ function start_crio_with_stopped_pod() {
 
 	run_crio_wipe
 
-	run_podman_with_args ps -a | grep test
+	run_podman_with_args container exists test
 }
 
 @test "do clear everything when shutdown file not found" {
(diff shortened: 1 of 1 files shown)


haircommander:
/approve
LGTM, nice investigative work

TomSweeneyRedHat:
LGTM

kolyshkin:
The failures are *definitely* unrelated since this PR does not changes anything in cri-o.

One failure seems to be a flaky test, which at some point was disabled (see https://github.com/cri-o/cri-o/pull/4385) but its signature has now changed, as so it no longer matches. Having said that, the bug which caused a flake is supposed to be fixed, but I am not quite sure in which version.

saschagrunert (approved):
LGTM

saschagrunert (approved):
/lgtm

PR #5655, opened 2022-02-16
Deny empty `localhost/` AppArmor profiles

saschagrunert:
#### What type of PR is this?

/kind bug

Specifying a `localhost/` profile for AppArmor would run containers as
unconfined and not the selected default profile. This means we now check
that case explicitly and error if so.

None
None

```release-note
Forbid AppArmor profiles with the name `localhost/`.
```


klihub:
LGTM

PR #5662, opened 2022-02-18
CI: use kubernetes and cri-tools from git tip (aka HEAD, aka master)

kolyshkin:
#### What type of PR is this?

/kind ci

Use latest kubernetes and cri-tools in CI.

None

```release-note
NONE
```

### DIFF
--- a/contrib/test/integration/build/kubernetes.yml
+++ b/contrib/test/integration/build/kubernetes.yml
@@ -53,7 +53,7 @@
 
       # Whole script added by Ansible from build/kubernetes.yml
       export PATH=/usr/local/go/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/root/bin:{{ ansible_env.GOPATH }}/bin:{{ ansible_env.GOPATH }}/src/k8s.io/kubernetes/third_party/etcd:{{ ansible_env.GOPATH }}/src/k8s.io/kubernetes/_output/local/bin/linux/amd64/
-      export FEATURE_GATES="AllAlpha=false,RunAsGroup=true"
+      export FEATURE_GATES="AllAlpha=false"
       export CONTAINER_RUNTIME=remote
       export CGROUP_DRIVER=systemd
       export CONTAINER_RUNTIME_ENDPOINT='{{ crio_socket }}'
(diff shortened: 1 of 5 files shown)


saschagrunert (approved):
LGTM

haircommander:
for main branch I think we should use `master` for both kubernetes and cri-tools

also It looks like we need to drop RunAsGroup=true from https://github.com/cri-o/cri-o/blob/main/contrib/test/integration/build/kubernetes.yml#L56

TomSweeneyRedHat:
LGTM
once tests are happy and @haircommander 's concerns are addressed.

kolyshkin:
> for main branch I think we should use `master` for both kubernetes and cri-tools

This will bring it more instability into CI (and thus all the PRs), but it still makes sense.

Done.

> 
> also It looks like we need to drop RunAsGroup=true from https://github.com/cri-o/cri-o/blob/main/contrib/test/integration/build/kubernetes.yml#L56

Right! Done.

haircommander:
we should add `Conformance Tests should have at least two untainted nodes` to the skipped tests as well.

kolyshkin:
> Conformance Tests should have at least two untainted nodes

Can you provide a quick one-liner as to why this test is being disabled?

haircommander:
> > Conformance Tests should have at least two untainted nodes
> 
> Can you provide a quick one-liner as to why this test is being disabled?

"While fully conformant kubernetes clusters require more than 2 nodes, verifying that CRI-O is a compliant CRI implementation does not warrant multiple nodes. TODO(haircommander): when refactoring CI, we should consider spawning multiple nodes to drop this line"

haircommander:
/hold
@kolyshkin the dependencies check is failing because `contrib/test/integration/main.yml` still has 1.22 for cri-tools version. 

also, wdyt about setting cri-tools version to master as well in this branch?

kolyshkin:
> the dependencies check is failing because contrib/test/integration/main.yml still has 1.22 for cri-tools version.
> also, wdyt about setting cri-tools version to master as well in this branch?

Well, we have a problem.

We can not set version to `master` in `scripts/versions` because the binary cri-tools is used, i.e. it must be a release tag:
https://github.com/cri-o/cri-o/blob/71f577203d6cb726c306388c291a7ee31dc6949b/contrib/bundle/build#L135

OTOH we can use `master` from `scripts/github-actions-setup` because in there it is compiled from source.

The third place we get cri-tools is in `contrib/test/integration/build/cri-tools.yml` -- in there it is also compiled from source.

So, to use `master`, we need to change `contrib/bundle/build` from getting a binary release to compiling cri-tools from source. This can be addressed separately, and I'd rather not do it in this PR (which is mainly motivated by reducing CI failures).

Now, I have to use cri-tools 1.23.0 _everywhere_ since otherwise we'll have a dependency check complaining (as it does now).

Updated accordingly.

haircommander:
> Now, I have to use cri-tools 1.23.0 everywhere since otherwise we'll have a dependency check complaining (as it does now).

ah well, works for me

/lgtm

PR #5676, opened 2022-02-23
server: stop deleting pod from idIndex if already gone

haircommander:

#### What type of PR is this?

/kind bug

```release-note
none
```

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -692,6 +692,10 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 		log.Infof(ctx, description)
 		err := s.PodIDIndex().Delete(sbox.ID())
 		if err != nil {
+			// already deleted
+			if strings.Contains(err.Error(), noSuchID) {
+				return nil
+			}
 			log.Warnf(ctx, "Could not delete pod id %s from idIndex", sbox.ID())
 		}
 		return err


saschagrunert (approved):
LGTM

server/sandbox_run_linux.go
```diff
@@ -692,6 +692,10 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 		log.Infof(ctx, description)
 		err := s.PodIDIndex().Delete(sbox.ID())
```
saschagrunert, line 693: Idea but not sure if that makes sense: How about making the `s.PodIDIndex().Delete` generally not fail if `noSuchID` is errored?

klihub: Another alternative would be to change  `containers/storage/pkg/truncindex` to return the already defined `ErrNotExist` instead of the dynamically constructed ID-specific error.

That way 1) the caller had a pretty cheap way of special-casing/ignoring that error (which looks like the only potential source of failure ATM), 2) existing external `truncindex` package behavior would not change in case some other users rely on `Delete()` failing for non-existing keys, 3) the caller could construct an ID-specific error, for instance using `errors.Wrap()`, if that's necessary.

haircommander: yeah either would work really, we talked about fixing hardcoding the error message before and never got around to it.

PR #5681, opened 2022-02-24
CI: always install required BATS version.

klihub:
#### What type of PR is this?

/kind ci

Currently CI nodes are expected to be set up correctly by regularly scheduled
ansible runs with the `setup` tag. However, empirical evidence (logging BATS
version from `integration` runs on CI nodes) proves that many of the CI nodes
are still running older versions (1.2.1 or 1.2.1-dev).

So instead of relying nodes to get independently set up with the correct BATS
version this PR installs that one right before running the tests. Furthermore, it
makes sure that the freshly installed version is correctly symlinked in place, to
replace any previously present/distro-provided versions.

None

```release-note
None
```

### DIFF
--- a/contrib/test/integration/build/bats.yml
+++ b/contrib/test/integration/build/bats.yml
@@ -16,3 +16,4 @@
     src: /usr/local/bin/bats
     dest: /usr/bin/bats
     state: link
+    force: true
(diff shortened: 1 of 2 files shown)


haircommander:
/approve

LGTM

saschagrunert (approved):
/lgtm

PR #5683, opened 2022-02-24
resource store: save stage of resource creation and report to callers

haircommander:
#### What type of PR is this?

/kind feature
Often, users are met with the error "Kubelet may be retrying requests that are timing out in CRI-O..." but aren't provided with any details as to why.

This PR adds functionality that saves the stage of container and pod creations, as well as logs them. Now, users will be able to see where their pod or container is

```release-note
CRI-O now logs the stage of container or pod creation under system load. This allows users to find why their creation requests are stalling.
```

### DIFF
--- a/internal/resourcestore/resourcestore.go
+++ b/internal/resourcestore/resourcestore.go
@@ -34,6 +34,7 @@ type Resource struct {
 	watchers []chan struct{}
 	stale    bool
 	name     string
+	stage    string
 }
 
 // wasPut checks that a resource has been fully defined yet.
@@ -194,4 +195,21 @@ func (rc *ResourceStore) WatcherForResource(name string) chan struct{} {
 	}
 	r.watchers = append(r.watchers, watcher)
 	return watcher
+}
+
+func (rc *ResourceStore) SetStageForResource(name, stage string) {
+	rc.mutex.Lock()
+	defer rc.mutex.Unlock()
+	r, ok := rc.resources[name]
+	if !ok {
+		logrus.Tracef("Initializing stage for resource %s to %s", name, stage)
+		rc.resources[name] = &Resource{
+			watchers: []chan struct{}{},
+			name:     name,
+			stage:    stage,
+		}
+		return
+	}
+	logrus.Tracef("Setting stage for resource %s from %s to %s", name, r.stage, stage)
+	r.stage = stage
 }
--- a/internal/resourcestore/resourcestore.go
+++ b/internal/resourcestore/resourcestore.go
@@ -8,7 +8,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const sleepTimeBeforeCleanup = 1 * time.Minute
+const (
+	sleepTimeBeforeCleanup = 1 * time.Minute
+	StageUnknown           = "unknown"
+)
 
 // ResourceStore is a structure that saves information about a recently created resource.
 // Resources can be added and retrieved from the store. A retrieval (Get) also removes the Resource from the store.
@@ -181,20 +184,20 @@ func (rc *ResourceStore) Put(name string, resource IdentifiableCreatable, cleane
 // This is useful for situations where clients retry requests quickly after they "fail" because
 // they've taken too long. Adding a watcher allows the server to slow down the client, but still
 // return the resource in a timely manner once it's actually created.
-func (rc *ResourceStore) WatcherForResource(name string) chan struct{} {
+func (rc *ResourceStore) WatcherForResource(name string) (watcher chan struct{}, stage string) {
 	rc.mutex.Lock()
 	defer rc.mutex.Unlock()
-	watcher := make(chan struct{}, 1)
+	watcher = make(chan struct{}, 1)
 	r, ok := rc.resources[name]
 	if !ok {
 		rc.resources[name] = &Resource{
 			watchers: []chan struct{}{watcher},
 			name:     name,
 		}
-		return watcher
+		return watcher, StageUnknown
 	}
 	r.watchers = append(r.watchers, watcher)
-	return watcher
+	return watcher, r.stage
 }
 
 func (rc *ResourceStore) SetStageForResource(name, stage string) {
--- a/server/utils.go
+++ b/server/utils.go
@@ -164,11 +164,11 @@ func (s *Server) getResourceOrWait(ctx context.Context, name, resourceType strin
 		log.Infof(ctx, "Found %s %s with ID %s in resource cache; using it", resourceType, name, cachedID)
 		return cachedID, nil
 	}
-	watcher := s.resourceStore.WatcherForResource(name)
+	watcher, stage := s.resourceStore.WatcherForResource(name)
 	if watcher == nil {
 		return "", fmt.Errorf("error attempting to watch for %s %s: no longer found", resourceType, name)
 	}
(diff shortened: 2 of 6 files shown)


haircommander:
/hold

I don't think a single map is the best choice here. we may need something shareded. need to do a perf analysis under load.
also need tests here


still, I wanted to spread to a wider audience so folks can chime in on stages that shoudl be added or removed.

rphillips:
can we move the logic into the Resource?

internal/resourcestore/resourcestore.go
```diff
@@ -18,13 +18,36 @@ const sleepTimeBeforeCleanup = 1 * time.Minute
...
-	closeChan chan struct{}
-	closed    bool
-	sync.Mutex
+	resources    map[string]*Resource
+	stages       map[string]Stage
+	timeout      time.Duration
+	closeChan    chan struct{}
+	closed       bool
+	resourceLock sync.Mutex
+	stageLock    sync.Mutex
```
haircommander, line 27: my current thouht is separating the locking mechanisms will reduce contention. still, if there are hundreds of containers or pods being created then this will get very busy. may need something sharded by key so each item can be updated atomically without bothering the others...

haircommander:
> can we move the logic into the Resource?

I thought about it, see https://github.com/cri-o/cri-o/pull/5683#discussion_r814297858 on why I decided doing it this way. I'm open if we can find a performant way

internal/resourcestore/resourcestore.go
rphillips: migrating to a RWLock might have some benefit.

haircommander: unfortunately, this value is write often, and rarely read. I have thought about using a RW mutex on the whole map and only adding names when we have a write lock, and then having an [atomic value](https://pkg.go.dev/sync/atomic#Value) for the Stage...

saschagrunert:

haircommander:
@cri-o/cri-o-maintainers PTAL, I have revitalized this and reimplemented the `stages` map to handle concurrency better

haircommander:
per reviews from both @rphillips and @kolyshkin I have just added the stage to Resource which simplifies things a bit

internal/resourcestore/resourcestore.go
```diff
@@ -131,15 +134,18 @@ func (rc *ResourceStore) Get(name string) string {
 
 	r, ok := rc.resources[name]
 	if !ok {
+		rc.mutex.Unlock()
 		return ""
 	}
 	// It is possible there are existing watchers,
 	// but no resource created yet
 	if !r.wasPut() {
+		rc.mutex.Unlock()
```
kolyshkin, line 143: This seem like a bugfix, so maybe put it as a separate commit.

internal/resourcestore/resourcestore.go
```diff
@@ -180,10 +186,10 @@ func (rc *ResourceStore) Put(name string, resource IdentifiableCreatable, cleane
 // This is useful for situations where clients retry requests quickly after they "fail" because
 // they've taken too long. Adding a watcher allows the server to slow down the client, but still
 // return the resource in a timely manner once it's actually created.
-func (rc *ResourceStore) WatcherForResource(name string) chan struct{} {
+func (rc *ResourceStore) WatcherForResource(name string) (watcher chan struct{}, stage string) {
```
kolyshkin, line 189: This does not belong to the first commit, as you still return a single value (i.e. the code can't be compiled).

server/utils.go
```diff
@@ -196,7 +196,7 @@ func (s *Server) getResourceOrWait(ctx context.Context, name, resourceType strin
 		err = fmt.Errorf("the requested %s %s is now ready and will be provided to the kubelet on next retry", resourceType, name)
 	}
 
-	return "", fmt.Errorf("kubelet may be retrying requests that are timing out in CRI-O due to system load: %w", err)
+	return "", fmt.Errorf("kubelet may be retrying requests that are timing out in CRI-O due to system load. Currently at stage %v: %w", stage, err)
```
kolyshkin, line 199: Should this belong to the second commit?

internal/resourcestore/resourcestore.go
```diff
@@ -25,6 +25,8 @@ type ResourceStore struct {
 	mutex     sync.Mutex
 }
 
+const StageUnknown = "unknown"
```
kolyshkin, line 28: This doesn't have to be global IMO.

internal/resourcestore/resourcestore.go
```diff
@@ -195,3 +201,28 @@ func (rc *ResourceStore) WatcherForResource(name string) chan struct{} {
...
+			watchers: []chan struct{}{},
+			name:     name,
+			stage:    stage,
+		}
+		return
+	}
+	r.stage = stage
+}
+
+func (rc *ResourceStore) StageForResource(name string) string {
```
kolyshkin, line 220: Seems that this method  is not used anywhere.

kolyshkin (changes requested):

internal/resourcestore/resourcestore.go
haircommander: I have left it for testing

TomSweeneyRedHat:
LGTM assuming happy tests

haircommander:
@saschagrunert PTAL

haircommander:
wakeup prow

/retest

haircommander:
prow was frozen, so I rebased, @saschagrunert can I have my tag back please :smiley:

haircommander:
hm 
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/5683/pull-ci-cri-o-cri-o-main-ci-rhel-critest/1570759588955820032
cc @sohankunkerkar @wgahnagl 

/retest

internal/resourcestore/resourcestore.go
```diff
@@ -181,20 +184,20 @@ func (rc *ResourceStore) Put(name string, resource IdentifiableCreatable, cleane
...
+func (rc *ResourceStore) WatcherForResource(name string) (watcher chan struct{}, stage string) {
 	rc.mutex.Lock()
 	defer rc.mutex.Unlock()
-	watcher := make(chan struct{}, 1)
+	watcher = make(chan struct{}, 1)
 	r, ok := rc.resources[name]
 	if !ok {
 		rc.resources[name] = &Resource{
 			watchers: []chan struct{}{watcher},
 			name:     name,
```
kolyshkin, line 195: Perhaps you need to set the stage to unknown here.

kolyshkin: ...and in every other place where you create a new `Resource`.

Alternatively, don't initialize it, but every time you return the stage value, check it for being empty and return `StageUnknown`. Is this is what you're doing, then nothing has to be changed in the code.

kolyshkin (approved):
LGTM

internal/resourcestore/resourcestore.go
```diff
@@ -34,6 +37,7 @@ type Resource struct {
 	watchers []chan struct{}
 	stale    bool
 	name     string
+	stage    string
```
mrunalp, line 40: Can we make this a first class type or enum?

haircommander: I had named strings a bit ago, but I think this is clearer and makes the source of the name closer to where it's relevant

PR #5691, opened 2022-02-28
test: allow state of failing tests to be kept intact.

klihub:
Allow tests to be kept intact upon failure for better debugging and diagnosing.

#### What type of PR is this?

/kind other

Allow tests to be kept intact upon failure for better debugging and diagnosing. If set, the newly introduced `TEST_KEEP_ON_FAILURE` variable causes failing tests to skip their teardown phase, leaving all test artifacts and any related cri-o instances intact.

This should allow one to look for flakes locally with something like this:
`while true; do TEST_KEEP_ON_FAILURE=1 JOBS=8 ./test/test_runner.sh || break; done`

None

```release-note
None
```

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -430,11 +430,21 @@ function cleanup_test() {
         cat "$CRIO_LOG"
         echo "# --- --- ---"
     fi
-    cleanup_ctrs
-    cleanup_pods
-    stop_crio
-    cleanup_lvm
-    cleanup_testdir
+
+    # Leave the test artifacts intact for failing tests if requested.
+    #
+    # BATS_TEST_COMPLETED is set by BATS to 1 if the test passed, otherwise
+    # it is left unset. The variable is also set if the test was skipped.
+    # See https://bats-core.readthedocs.io/en/stable/faq.html#how-can-i-check-if-a-test-failed-succeeded-during-teardown for more details.
+    if [ -z "$TEST_KEEP_ON_FAILURE" ] || [ "${BATS_TEST_COMPLETED:-}" = "1" ]; then
+        cleanup_ctrs
+        cleanup_pods
+        stop_crio
+        cleanup_lvm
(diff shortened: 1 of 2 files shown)


test/helpers.bash
```diff
@@ -430,11 +430,16 @@ function cleanup_test() {
         cat "$CRIO_LOG"
         echo "# --- --- ---"
     fi
-    cleanup_ctrs
-    cleanup_pods
-    stop_crio
-    cleanup_lvm
-    cleanup_testdir
+
+    if [ -z "$TEST_KEEP_ON_FAILURE" ] || [ "${BATS_TEST_COMPLETED:-}" = "1" ]; then
```
haircommander, line 434: should we do != 0 in case it exits with a different code?

klihub: Do you mean the failure condition test `"$BATS_TEST_COMPLETED" = "1"` bit ?

If you do, then that specific bit I dug that out of the [BATS documentation](https://bats-core.readthedocs.io/en/stable/faq.html#how-can-i-check-if-a-test-failed-succeeded-during-teardown). It says the BATS_TEST_COMPLETED variable will be set to 1 if the test passed (and I believe also in the case the test is skipped, at least that's what my test runs indicate), otherwise it is left unset.

haircommander: Ah I see, could we have a link here to describe that?

TomSweeneyRedHat:
LGTM once @haircommander 's comment is addressed

test/helpers.bash
klihub: Here is the BATS docs bit describing how test completion is reflected by `BATS_TEST_COMPLETED`: https://bats-core.readthedocs.io/en/stable/faq.html#how-can-i-check-if-a-test-failed-succeeded-during-teardown

If you consider it necessary, I can add a reference to that in the commit message.

PR #5713, opened 2022-03-09
Handle manually paused containers

saschagrunert:
#### What type of PR is this?

/kind feature

We now handle manually paused containers (via `runc pause`) better,
which means:

- Pods with paused containers can be stopped by Kubernetes, because we
  now automatically resume them before attempting to stop the container.
- We're now able to handle the paused state in CRI-O internally by going
  back to `CONTAINER_RUNNING` if a user invoked `runc resume`. We also
  set the creation and started date accordingly to not throw Kubernetes
  events complaining that the dates are not set.

None
Unfortunately we cannot easily detect if a container is in paused state without updating the container status internally. Because of that we have to trigger an RPC which automatically updates the status, like `execsync`.

```release-note
Improved handling of manually paused containers.
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -712,6 +712,18 @@ func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout in
 		}
 	}()
 
+	// Paused containers cannot be stopped, so we have to resume them and
+	// update the container status accordingly.
+	if c.state.Status == ContainerStatePaused {
+		logrus.Warnf("Container %s is paused, resuming before trying to stop", c.ID())
+		if err := r.UnpauseContainer(ctx, c); err != nil {
+			return errors.Wrap(err, "resume container")
+		}
+		if err := r.UpdateContainerStatus(ctx, c); err != nil {
+			return errors.Wrap(err, "update container status")
+		}
+	}
+
 	c.opLock.Lock()
 	defer c.opLock.Unlock()
 
--- a/server/container_status.go
+++ b/server/container_status.go
@@ -54,7 +54,12 @@ func (s *Server) ContainerStatus(ctx context.Context, req *types.ContainerStatus
 
 	// If we defaulted to exit code not set earlier then we attempt to
 	// get the exit code from the exit file again.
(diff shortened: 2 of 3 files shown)


server/container_status.go
```diff
@@ -54,7 +54,12 @@ func (s *Server) ContainerStatus(ctx context.Context, req *types.ContainerStatus
 
 	// If we defaulted to exit code not set earlier then we attempt to
 	// get the exit code from the exit file again.
-	if cState.Status == oci.ContainerStateStopped && cState.ExitCode == nil {
+	if (cState.Status == oci.ContainerStateStopped && cState.ExitCode == nil) ||
+		// Manually paused containers should be covered as well.
+		// Kubernetes has no way if displaying that information (state will be
```
TomSweeneyRedHat, line 59:
```suggestion
		// Kubernetes has no way of displaying that information (state will be
```

internal/oci/runtime_oci.go
```diff
@@ -712,6 +712,18 @@ func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout in
 		}
 	}()
 
+	// Paused containers cannot be stopped, so we have to resume them and
+	// update the container status accordingly.
+	if c.state.Status == ContainerStatePaused {
+		logrus.Warnf("Container %s is paused, resuming before trying to stop", c.ID())
+		if err := r.UnpauseContainer(ctx, c); err != nil {
+			return errors.Wrap(err, "resume container")
```
TomSweeneyRedHat, line 720: would it be better to have "failed to resume container"?  If so, similar comment for other error messages.

server/container_status.go
saschagrunert: Fixed as suggested :+1:

internal/oci/runtime_oci.go
saschagrunert: Well' I'd prefer reducing the amount of word repetitions in error messages. Usually the return value (an `error` type) already indicates that something went wrong. Assuming that those errors are chained, the resulting message would look like:
`error in x: did y: on my way to z` compared to: `error in x: failed to do y: failed to z`

PR #5726, opened 2022-03-16
README: Update EOL & Version Skew links

travier:
#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it:

Minor doc update

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

N/A

```release-note
NONE
```

PR #5728, opened 2022-03-16
pinns: Pass sysctls as repeated '-s' arguments

travier:
#### What type of PR is this?

/kind cleanup

Use repeated '-s' arguments to pass sysctls keys and values to pinns to replace the serialization in cri-o and parsing in pinns.

To simplify the logic, this allocates an 'argc/2' sized array to store the pointers to the sysctls `k=v` pairs. This array will in most cases be much bigger than what is really needed but this should be safe as argc is limited in size on Linux.

This is a followup fix for https://github.com/cri-o/cri-o/security/advisories/GHSA-6x2m-w449-qwx7.

None

```release-note
None
```

### DIFF
--- a/internal/config/nsmgr/nsmgr.go
+++ b/internal/config/nsmgr/nsmgr.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 	"syscall"
 
 	nspkg "github.com/containernetworking/plugins/pkg/ns"
@@ -87,12 +86,8 @@ func (mgr *NamespaceManager) NewPodNamespaces(cfg *PodNamespacesConfig) ([]Names
 		"-f", pinnedNamespace,
 	}
 
-	if len(cfg.Sysctls) != 0 {
-		pinnsSysctls, err := getSysctlForPinns(cfg.Sysctls)
-		if err != nil {
-			return nil, errors.Wrapf(err, "invalid sysctl")
-		}
-		pinnsArgs = append(pinnsArgs, "-s", pinnsSysctls)
+	for key, value := range cfg.Sysctls {
+		pinnsArgs = append(pinnsArgs, "-s", fmt.Sprintf("%s=%s", key, value))
 	}
 
 	var rootPair idtools.IDPair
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -19,15 +19,15 @@ CRI-O follows the Kubernetes release cycles with respect to its minor versions
 (`1.x.0`). Patch releases (`1.x.y`) for CRI-O are not in sync with those from
 Kubernetes, because those are scheduled for each month, whereas CRI-O provides
 them only if necessary. If a Kubernetes release goes [End of
-Life](https://github.com/kubernetes/website/blob/main/content/en/releases/patch-releases.md),
```
saschagrunert, line 22: I guess this change is already part of https://github.com/cri-o/cri-o/pull/5726

internal/config/nsmgr/nsmgr.go
```diff
@@ -88,11 +87,9 @@ func (mgr *NamespaceManager) NewPodNamespaces(cfg *PodNamespacesConfig) ([]Names
 	}
 
 	if len(cfg.Sysctls) != 0 {
-		pinnsSysctls, err := getSysctlForPinns(cfg.Sysctls)
-		if err != nil {
-			return nil, errors.Wrapf(err, "invalid sysctl")
+		for key, value := range cfg.Sysctls {
+			pinnsArgs = append(pinnsArgs, "-s", "%s=%s", key, value)
```
haircommander, line 91: "%s=%s" needs to be wrapped in single quotes, as sysctls can have spaces in them

haircommander:
thanks for opening this @travier !

internal/config/nsmgr/nsmgr.go
travier: Isn't that preserved as we pass each "k=v" as distinct arguments?

haircommander: like `"'%s=%s'"`

internal/config/nsmgr/nsmgr.go
```diff
@@ -88,11 +87,9 @@ func (mgr *NamespaceManager) NewPodNamespaces(cfg *PodNamespacesConfig) ([]Names
 	}
 
 	if len(cfg.Sysctls) != 0 {
```
saschagrunert, line 89: We have no need to check the length if we only `range` over it.

internal/config/nsmgr/nsmgr.go
travier: `/proc/sys/%s` > I had forgotten to format the `%s=%s` string before passing it as an argument, thus the `%s` in the error here.

haircommander:
hm I guess I was wrong and we don't need the quotes

/approve

saschagrunert (approved):
LGTM

PR #5745, opened 2022-03-21
pinns: Check calloc return value

travier:
#### What type of PR is this?

/kind cleanup

This is unlikely to ever happen but let's do it just to be safe.

Fixes: 1af1f8af2 pinns: Pass sysctls as repeated '-s' arguments

```release-note
None
```

### DIFF
--- a/pinns/src/pinns.c
+++ b/pinns/src/pinns.c
@@ -69,6 +69,8 @@ int main(int argc, char **argv) {
   };
 
   sysctls = calloc(argc/2, sizeof(char *));
+  if (UNLIKELY(sysctls == NULL))
+      pexit("Failed to calloc");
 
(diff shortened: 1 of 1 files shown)


pinns/src/pinns.c
```diff
@@ -69,6 +69,8 @@ int main(int argc, char **argv) {
   };
 
   sysctls = calloc(argc/2, sizeof(char *));
+  if (sysctls == NULL)
```
haircommander, line 72: we can use the UNLIKELY macro to let the compiler optimize

travier: Done

haircommander:
/retest
/lgtm

thank you @travier !

Issue #1921, opened 2018-11-16
Default pids_limit too low


steven-sheehy:
**Description**
For heavily mult-threaded workloads like databases, the default of 1024 for pids-limit is too low. This parameter controls not just the number of processes but also the amount of threads (since a thread is technically a process with shared memory). When installing CRI-O via the PPA, it comes pre-packaged with a crio.conf that defaults to 1024. Also CRI-O defaults internally to 1024 according to `crio --help | grep pids-limit`. 

I'm not sure what a good default is, but I set it to 64000 in my environment as that is what MongoDB recommended for nproc.

**Steps to reproduce the issue:**
1. Start CRI-O with defaults
2. Run the following test pod which kicks of 2048 threads:
```yaml
kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: pid-limit
spec:
  containers:
  - image: docker.io/ssheehy/pid-limit
    name: pid-limit
EOF
```

**Describe the results you received:**
Pod crashes repeatedly since it can't create threads past 1024

**Describe the results you expected:**
Pod continues to run and can create > 1024 threads.

**Additional information you deem important (e.g. issue happens only occasionally):**
Discussed in slack with @runcom and @mrunalp 

**Output of `crio --version`:**

```
crio version 1.11.9-dev
commit: ""
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Ubuntu 18.04
`ulimit -u` is 47931
Kubernetes v1.11.4

cyphar:
That's a bit of a worry. It means that containers under containerd can trivially fork-bomb the system.

runcom:
Right, that's the reason why (fork bomb) we are defaulting as low but give the possibility to users to tune it the way they need for their workloads.

chrisnegus:
I'd like to document this here: https://access.redhat.com/documentation/en-us/openshift_container_platform/3.11/html-single/cri-o_runtime/#configure-crio-use-crio-engine.  Questions: 1) Can pids_limit be set when the node is deployed or should you edit pids_limit in the crio.conf file on the node? 2) What guidance can we provide on when you need a higher limit (databases?)?  3) What can happen if the pids_limit is set too high or set to unlimited?

cyphar:
For the record, I think 1024 is **far** too small for a default. systemd did this for user slices and we saw (at SUSE) lots of regressions as a result. 4096 (the systemd default for services) is much more reasonable.

[...]

PR #5755, opened 2022-03-24
Raise DefaultPidsLimit from 1024 to 4096

robbmanes:
To match some workloads such as databases with default configuration, increase the default pid_limit to 4096.

Signed-off-by: Robb Manes <rmanes@redhat.com>

#### What type of PR is this?

/kind feature

The default `pids_limit` was defined at 1024, which is no longer suitable for many out-of-the-box workloads (notably we've seen database workloads deployed that easily exceed this).  Raising this to 4096 should no longer be of as much concern as it was previously as the defaults elsewhere, such as within `systemd`, are also 4096, as well as other container engines. 

Fixes: https://github.com/cri-o/cri-o/issues/1921

Feedback appreciated.  We have quite a number of issues regarding the defaults needing to be overridden, and quite a few requests to have the defaults just be larger as some common application deployments are exceeding 1024 (databases, mostly).

None

```release-note

```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -114,7 +114,7 @@ const (
 const (
 	// DefaultPidsLimit is the default value for maximum number of processes
 	// allowed inside a container
-	DefaultPidsLimit = 1024
+	DefaultPidsLimit = 4096
 
 	// DefaultLogSizeMax is the default value for the maximum log size
 	// allowed for a container. Negative values mean that no limit is imposed.


saschagrunert (approved):
Hey @robbmanes, thank you for this contribution. Let's add a release note for that! Code LGTM if we find consensus with the change.
/ok-to-test

haircommander:
I'll raise you one: I think we should set it to `-1` (or `0`) and mark it as deprecated. The Kubelet has PodPidsLimit parameter that does the same thing, but on a pod level. The existence of both is confusing, as folks in the past have had to set both to properly set it as desired. I was thinking of deprecating for 1.24 anyway. While we're at it, I was also thinking of deprecating `log-size-max` (it's already set to no limit)

haircommander:
would you like to take care of that @robbmanes or would you like the cri-o team to?

robbmanes:
> would you like to take care of that @robbmanes or would you like the cri-o team to?

I got it, and will also add a release note.  Re-push incoming.  TYVM!

haircommander:
/hold

as this PR isn't in the correct state. I'll take it over tomorrow and move to the correct defaults

haircommander:
replaced by https://github.com/cri-o/cri-o/pull/5831

saschagrunert:

haircommander:
we've instead moved to `-1`

briantward:
@haircommander KubeletConfig does set `PodPidsLimit` (in the upstream spec that's passed through to the machineconfiguration spec).  I found that shortly after posting my question here. 

https://docs.openshift.com/container-platform/4.10/post_installation_configuration/machine-configuration-tasks.html#create-a-kubeletconfig-crd-to-edit-kubelet-parameters_post-install-machine-configuration-tasks

That leaves one last question, which is, what will be the default value of `PodPidsLimit` ? It looks like 4096 based on this:

https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-kubelet/_base/files/kubelet.yaml#L21

I think all that's left is cleaning up the docs for the release where this changes.  I'll open a docs BZ for that, unless you know of that being covered already?

haircommander:
yeah that looks like the default. Eventually, we will need a conversion webhook, but for now, we should move recommendations from using ctrcfg to using kubeletconfig. I don't know of a bz yet, thank you for looking into this @briantward

PR #5758, opened 2022-03-24
Add bats test for infra_ctr_cpuset taskset

lack:
Signed-off-by: Jim Ramsay <jramsay@redhat.com>

#### What type of PR is this?

/kind ci

This adds a bats test to check the code change in #5514

None

```release-note
None
```

### DIFF
--- a/test/infra_ctr_cpuset.bats
+++ b/test/infra_ctr_cpuset.bats
@@ -1,4 +1,5 @@
 #!/usr/bin/env bats
+# vim:set ft=bash :
 
 load helpers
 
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
Awesome, thanks! LGTM

haircommander:
/approve
/lgtm

thanks @lack 

I wish there was a way to make sure every child of crio was always in the right cpuset. the real nefarious behavior comes if the process is scheduled wrong from the beginning, but that's hard (impossible?) to check

lack:
> I wish there was a way to make sure every child of crio was always in the right cpuset. the real nefarious behavior comes if the process is scheduled wrong from the beginning, but that's hard (impossible?) to check

Do you mean "the process" as in crio itself?

We could have a test that resets crio's taskset after starting it, then checks that a container and the related conmon are both taskset correctly?

haircommander:
ah I mean the process as in conmon, runc, taskset basically all the processes crio and its children start. we've seen issues where it looks like the processes aren't being put in the right place and it would be sweet to rule out that possiblity

lack:
> ah I mean the process as in conmon, runc, taskset basically all the processes crio and its children start. we've seen issues where it looks like the processes aren't being put in the right place and it would be sweet to rule out that possiblity

Cool, I'll see what I can come up with...

PR #5775, opened 2022-04-04
server: Canonize selinux label for comparison with filesystem label

manuelluis:
#### What type of PR is this?

/kind bug

If you try to use the annotation: io.kubernetes.cri-o.TrySkipVolumeSELinuxLabel, It doesn't work because, the selinux label with the mcs from the annotation in the namespace, has the wrong order in the mcs.
```
$ oc get project test -o yaml | grep mcs
openshift.io/sa.scc.mcs: s0:c75,c35
```
The selinux label for the comparison with the label in the filesystem will be:  **system_u:object_r:container_file_t:s0:c75,c35**
but in the filesystem the label is: **system_u:object_r:container_file_t:s0:c35,c75**

None

```release-note
Fixed `io.kubernetes.cri-o.TrySkipVolumeSELinuxLabel` annotation usage with the OpenShift MCS.
```

### DIFF
--- a/server/label_linux.go
+++ b/server/label_linux.go
@@ -3,6 +3,7 @@ package server
 import (
 	"fmt"
 
+	selinux "github.com/opencontainers/selinux/go-selinux"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


server/label_linux.go
```diff
@@ -11,12 +12,17 @@ import (
...
-		if err == nil && currentLabel == secLabel {
-			logrus.Debugf(
-				"Skipping relabel for %s, as TrySkipVolumeSELinuxLabel is true and the label of the top level of the volume is already correct",
-				path)
-			return nil
+		canonicalSecLabel, err := selinux.CanonicalizeContext(secLabel)
+		if err != nil {
+			logrus.Errorf("Canonicalize label failed %s: %v", secLabel, err)
+		} else {
+			currentLabel, err := label.FileLabel(path)
```
haircommander, line 19: should we Canonicalize this one as well, or do we think it'll be correct because we're reading it from disk (after it's been procesed by the kernel)

haircommander:
/approve

saschagrunert (approved):
/lgtm
/retest-required

saschagrunert:
Integration tests have been fixed in another PR.

PR #5778, opened 2022-04-05
Bug 2064741: Disable systemd-mode cgroup detection if /sys/fs/cgroup is bind mounted from the host

harche:
Disable systemd-mode cgroup detection if /sys/fs/cgroup is bind mounted from the host 

#### What type of PR is this?

/kind bug

Disable systemd-mode cgroup detection if /sys/fs/cgroup is bind mounted from the host

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=2064741

```release-note
Disable systemd-mode cgroup detection if /sys/fs/cgroup is bind mounted from the host 
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -37,6 +37,11 @@ import (
 	"github.com/intel/goresctrl/pkg/blockio"
 )
 
+const (
+	cgroupSysFsPath        = "/sys/fs/cgroup"
+	cgroupSysFsSystemdPath = "/sys/fs/cgroup/systemd"
+)
+
 // createContainerPlatform performs platform dependent intermediate steps before calling the container's oci.Runtime().CreateContainer()
 func (s *Server) createContainerPlatform(ctx context.Context, container *oci.Container, cgroupParent string, idMappings *idtools.IDMappings) error {
 	if idMappings != nil && !container.Spoofed() {
@@ -508,7 +513,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 				Options:     []string{"nosuid", "noexec", "nodev", "ro"},
 			})
 			ctr.SpecAddMount(rspec.Mount{
-				Destination: "/sys/fs/cgroup",
+				Destination: cgroupSysFsPath,
 				Type:        "cgroup",
 				Source:      "cgroup",
 				Options:     []string{"nosuid", "noexec", "nodev", "relatime", "ro"},
@@ -524,7 +529,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 			Options:     []string{"nosuid", "noexec", "nodev", "rw", "rslave"},
 		})
 		ctr.SpecAddMount(rspec.Mount{
-			Destination: "/sys/fs/cgroup",
+			Destination: cgroupSysFsPath,
 			Type:        "cgroup",
 			Source:      "cgroup",
 			Options:     []string{"nosuid", "noexec", "nodev", "rw", "relatime", "rslave"},
@@ -1009,7 +1014,7 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
 
 	if _, mountSys := mountSet["/sys"]; !mountSys {
 		m := rspec.Mount{
-			Destination: "/sys/fs/cgroup",
+			Destination: cgroupSysFsPath,
 			Type:        "cgroup",
 			Source:      "cgroup",
 			Options:     []string{"nosuid", "noexec", "nodev", "relatime"},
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

server/container_create_linux.go
```diff
@@ -1076,18 +1076,32 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
...
+			}
+			g.AddMount(systemdMnt)
 		}
-		g.AddMount(systemdMnt)
 		g.AddLinuxMaskedPaths("/sys/fs/cgroup/systemd/release_agent")
 	}
 	g.AddProcessEnv("container", "crio")
 }
 
+func doesntHaveSysFsCgroupFromHost(mounts []rspec.Mount) bool {
```
haircommander, line 1096: naming bikeshed: can this  be `doesNot...` or even better, we could use `NoCgroupMount()` The only mounts that exist are from host, so I think that piece is redundant at least...

server/container_create_linux.go
```diff
@@ -1076,18 +1076,32 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
...
-		systemdMnt := rspec.Mount{
-			Destination: "/sys/fs/cgroup/systemd",
-			Type:        "bind",
-			Source:      "/sys/fs/cgroup/systemd",
-			Options:     []string{"bind", "nodev", "noexec", "nosuid"},
+		// If the /sys/fs/cgroup is bind mounted from the host,
+		// then systemd-mode cgroup should be disabled
+		// https://bugzilla.redhat.com/show_bug.cgi?id=2064741
+		if doesntHaveSysFsCgroupFromHost(g.Mounts()) {
+			systemdMnt := rspec.Mount{
```
haircommander, line 1083: what happens when we just don't add this mount? in addOCIBindMounts, we add `/sys/fs/cgroup` if we haven't already mounted `/sys`. This means, theoretically, this code block would only run if we're mounting sys, which would inherit `/sys/fs/cgroup/systemd` anyway, right? (cc @giuseppe )

harche: @haircommander You are right! Completely removing that code block also fixes the issue. 

https://github.com/cri-o/cri-o/pull/5785/files#diff-1e0877112035840d3e2f073c693665f8daffa241a440e54a866ad7382b510ea2R1079-R1093

We are handling this already in `addOCIBindMounts` function for both cgroups v1 and v2,

https://github.com/cri-o/cri-o/blob/main/server/container_create_linux.go#L1010-L1024

So IMHO, we can completely get rid of,

https://github.com/cri-o/cri-o/blob/main/server/container_create_linux.go#L1069-L1085

@giuseppe @haircommander WDYT?

harche: cc @saschagrunert

haircommander: less code is better to me!

giuseppe: I am not sure that would work.

on cgroup v2, systemd requires `/sys/fs/cgroup` to be writeable, while on cgroupv1 it requires `/sys/fs/cgroup/systemd` to be writeable.

The mount could exist when we bind mount `/sys/fs/cgroup` but it is not writeable

harche: I haven't tested in cgroups v2, but I guess for now we can leave the cgroups v2 part as is and only update the code related to cgroups v1

harche: The existing code already takes care of mounting with appropriate read/write options, 

https://github.com/cri-o/cri-o/blob/main/server/container_create_linux.go#L1018-L1023

giuseppe: the existing code, AFAICS, depends on an annotation to be present, and it doesn't address `/sys/fs/cgroup/systemd` being writeable on cgroup v1

server/container_create_linux.go
```diff
@@ -1076,13 +1076,6 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
 		}
 		g.AddMount(systemdMnt)
 	} else {
-		systemdMnt := rspec.Mount{
-			Destination: "/sys/fs/cgroup/systemd",
-			Type:        "bind",
-			Source:      "/sys/fs/cgroup/systemd",
-			Options:     []string{"bind", "nodev", "noexec", "nosuid"},
-		}
-		g.AddMount(systemdMnt)
```
giuseppe, line 1085: I think we need this code chunk when running on cgroup v1 with a fresh `/sys/fs/cgroup` mount

server/container_create_linux.go
```diff
@@ -1076,18 +1076,32 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
...
 		}
-		g.AddMount(systemdMnt)
 		g.AddLinuxMaskedPaths("/sys/fs/cgroup/systemd/release_agent")
 	}
 	g.AddProcessEnv("container", "crio")
 }
 
+func NoCgroupMount(mounts []rspec.Mount) bool {
+	for _, m := range mounts {
+		if m.Destination == "/sys/fs/cgroup" {
```
giuseppe, line 1098: don't we also need to check that there is `bind` or `rbind` in the mount options?

I think in addition to `/sys/fs/cgroup` we should also detect `/sys` and `/sys/fs`.

haircommander: I agree

harche: Thanks, updated.

giuseppe (approved):
LGTM

haircommander:
@harche can you rebase to get the latest CI fixes?

PR #5785, opened 2022-04-06
WIP: Bug 2064741  purge mounting fs cgroup

harche:

```release-note

```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -1076,16 +1076,30 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
 		}
 		g.AddMount(systemdMnt)
 	} else {
-		systemdMnt := rspec.Mount{
-			Destination: "/sys/fs/cgroup/systemd",
-			Type:        "bind",
-			Source:      "/sys/fs/cgroup/systemd",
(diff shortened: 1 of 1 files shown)


harche:
/hold

Created for testing with cluster-bot

server/container_create_linux.go
```diff
@@ -1076,18 +1076,32 @@ func setupSystemd(mounts []rspec.Mount, g generate.Generator) {
...
 		}
-		g.AddMount(systemdMnt)
 		g.AddLinuxMaskedPaths("/sys/fs/cgroup/systemd/release_agent")
 	}
 	g.AddProcessEnv("container", "crio")
 }
 
+func doesntHaveSysFsCgroupFromHost(mounts []rspec.Mount) bool {
+	for _, m := range mounts {
+		if m.Destination == "/sys/fs/cgroup" {
```
haircommander, line 1098: does this PR work if the user manually mounts `/sys`? i.e. we currently skip if the direct parent is mounted, but don't check any indirect. should we?

haircommander:
/ok-to-test
/approve

I think this approach is the one we want, so we don't risk systemd not being able to access something  because the mount is ro

BTW, you've got a lint issue

harche:
/close 

We have decided to go with https://github.com/cri-o/cri-o/pull/5778

PR #5788, opened 2022-04-06
seccomp: drop unshare syscall from default profile

haircommander:
#### What type of PR is this?

/kind api-change
when it comes to confining processes, having a contained process have access to unshare has historically lead to a host of vulnerabilities. CRI-O attempts to be secure by default, so in addition to adding a seccomp profile by default, we should also drop unshare by default, unless a container has CAP_SYS_ADMIN

```release-note
Block containers without CAP_SYS_ADMIN on `unshare` syscall in the default seccomp profile, to better contain unprivileged container processes.
```

### DIFF
--- a/internal/config/seccomp/seccomp.go
+++ b/internal/config/seccomp/seccomp.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/containers/common/pkg/seccomp"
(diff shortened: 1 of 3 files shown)


giuseppe:
This could be a problem for buildah.

@rhatdan PTAL

wgahnagl (approved):
lgtm

haircommander:
/hold

real quick

saschagrunert (approved):

PR #5740, opened 2022-03-17
Refactor Stop and Remove methods in ContainerServer and Server

haircommander:
#### What type of PR is this?

/kind cleanup
#### What this PR does / why we need it:
I personally find it confusing that the ContainerServer has a stop and remove method. Besides server restore (which is a separate matter) it is not the entities that provision the storage/cgroup/container, yet it has methods to stop and remove them. I think it makes more sense to have these methods in one place, and Server seems like the better fit.

This also includes some deduplication of methods
#### Does this PR introduce a user-facing change?

```release-note
none
```

PR #5823, opened 2022-04-25
config: move conmon specific fields in to runtime handler

haircommander:
#### What type of PR is this?

/kind deprecation
This makes the transition to conmonrs easier, as we can update conmonrs fields inpendently of the conmon ones.

```release-note
Deprecate `conmon`, `conmon-cgroup`, and `conmon-env` config options in favor of runtime handler specific fields `monitor-path`, `monitor-cgroup`, and `monitor-env`
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -56,6 +56,7 @@ const (
 	defaultNamespacesDir       = "/var/run"
 	RuntimeTypeVMBinaryPattern = "containerd-shim-([a-zA-Z0-9\\-\\+])+-v2"
 	tasksetBinary              = "taskset"
+	defaultMonitorCgroup       = "system.slice"
 )
 
 // Config represents the entire set of configuration values that can be set for
@@ -1090,7 +1091,7 @@ func defaultRuntimeHandler() *RuntimeHandler {
 		MonitorEnv: []string{
 			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
 		},
-		MonitorCgroup: "system.slice",
+		MonitorCgroup: defaultMonitorCgroup,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


haircommander:
good point, I've updated the kind so it'll be correctly catagorized

haircommander:
interestingly, I think some of the tests will be helped by https://github.com/cri-o/cri-o/pull/5740

/retest

pkg/config/template.go
```diff
@@ -1103,6 +1106,12 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
 {{ $.Comment }}allowed_annotations = [
 {{ range $opt := $runtime_handler.AllowedAnnotations }}{{ $.Comment }}{{ printf "\t%q,\n" $opt }}{{ end }}{{ $.Comment }}]
 {{ $.Comment }}{{ end }}
+{{ $.Comment }}monitor_path = "{{ $runtime_handler.MonitorPath }}"
+{{ $.Comment }}{{ if $runtime_handler.MonitorEnv }}
+{{ $.Comment }}monitor_env = [
+{{ range $opt := $runtime_handler.MonitorEnv }}{{ $.Comment }}{{ printf "\t%q,\n" $opt }}{{ end }}{{ $.Comment }}]
+{{ $.Comment }}{{ end }}
+{{ $.Comment }}monitor_cgroup = "{{ $runtime_handler.MonitorCgroup }}"
```
donpenney, line 1114: The new option names are missing the runtime_ prefix to align with the declarations in config.go

pkg/config/config.go
```diff
@@ -1111,6 +1109,11 @@ func (c *RuntimeConfig) ValidateRuntimes() error {
 			logrus.Warnf("'%s is being ignored due to: %q", name, err)
 			failedValidation = append(failedValidation, name)
 		}
+		if c.DefaultRuntime == name {
+			if err := c.TranslateMonitorFields(handler); err != nil {
```
donpenney, line 1113: It looks like there's no default set for MonitorPath, as defaultRuntimeHandler doesn't set one and TranslateMonitorFields is only called for c.DefaultRuntime. Without runtime_monitor_path explicitly set in the config file, I think this means MonitorPath will be an empty string for any non-default runtime handlers.

pkg/config/template.go
haircommander: we already have privileged_without_host_devices not prefixed. I personally think it's alright, as these fields are configuring the monitor, not the runtime

pkg/config/config.go
haircommander: good point! I've updated to run if the runtime is of type `oci`

pkg/config/template.go
donpenney: Aren't the new options defined with a runtime_ prefix in config.go, though? ie. runtime_monitor_path? So the example config this template generates would be missing that prefix and not be parsed?

haircommander: oh jeez, you're right. I'll fix it there

haircommander: fixed

saschagrunert (approved):
/lgtm

PR #5831, opened 2022-04-28
config: increase pids limit to unlimited and deprecate it and logSizeMax

haircommander:

#### What type of PR is this?

/kind deprecation

```release-note
Mark --pids-limit and --log-size-max options as deprecated, as well as set the default pids-limit to unlimited. Users should migrate to using the Kubelet's flags --pod-pids-limit and --container-log-max-size respectively.
```


mrunalp (approved):
/lgtm

PR #5836, opened 2022-05-02
main(): do not treat reexec.Init() == true as an error

nalind:
#### What type of PR is this?

/kind bug

`reexec.Init()` returning true is an indication that a handler was invoked and that we should return immediately because we're not the parent process.  It isn't an error condition, and we shouldn't be treating it as an error.

h/t to @adrianreber who asked about this while working on something else.

```release-note
None
```

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -111,8 +111,7 @@ func main() {
 	log.InitKlogShim()
 
 	if reexec.Init() {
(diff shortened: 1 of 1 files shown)


adrianreber:
@nalind Thanks. That fixes my attempts to create checkpoint images from CRI-O during checkpointing.

PR #5837, opened 2022-05-02
Exec cgroup

donpenney:
#### What type of PR is this?

/kind bug

Currently, exec sync containers are launched under the crio cgroup. For applications with a large number of exec probes, this can impact cpu usage for reserved platform cores, in addition to affecting accounting. To address this, this PR introduces an opt-in exec_cgroup option (defaults to false) that moves the exec probe process to the pod's cgroup. This ensures cpu usage and resource accounting are restricted to the pod's allowances.

None

```release-note
Add `monitor_exec_cgroup` to the configuration's runtime handler struct. This allows an admin to specify which cgroup the monitor for exec sync requests runs in (defaults to that of CRI-O).
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -16,6 +16,7 @@ import (
 	"github.com/containernetworking/plugins/pkg/ns"
 	conmonconfig "github.com/containers/conmon/runner/config"
 	"github.com/containers/storage/pkg/pools"
+	"github.com/cri-o/cri-o/internal/config/cgmgr"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/pkg/config"
 	"github.com/cri-o/cri-o/server/metrics"
@@ -440,7 +441,17 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 			Err:      err,
 		}
 	}
+
+	childStartPipe, parentStartPipe, err := newPipe()
+	if err != nil {
+		return nil, &ExecSyncError{
+			ExitCode: -1,
+			Err:      err,
+		}
+	}
+
 	defer parentPipe.Close()
+	defer parentStartPipe.Close()
 	defer func() {
 		if e := os.Remove(pidFile); e != nil {
 			log.Warnf(ctx, "Could not remove temporary PID file %s", pidFile)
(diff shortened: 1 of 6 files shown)


internal/oci/runtime_oci.go
```diff
@@ -521,6 +577,42 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 
 	// We don't need childPipe on the parent side
 	childPipe.Close()
+	childStartPipe.Close()
+
+	if r.execCgroup && r.config.InfraCtrCPUSet != "" {
+		// Update the exec's cgroup
+		containerPid, err := c.pid()
```
haircommander, line 584: I don't think this is the pid we want. we write to a pidFile above, I think the pid we want is there

internal/oci/runtime_oci.go
```diff
@@ -427,6 +435,32 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
 	return cmdErr
 }
 
+// moveExecToCgroup moves Exec command pid to the pod group
+func moveExecToCgroup(containerPid, commandPid int) error {
```
haircommander, line 439: 1: is there any way we can merge this funcitonality with the CreateContainer variant?

2: I think we want this to be in `internal/config/cgmgr` so we can toggle based on cgroup type and version

internal/oci/runtime_oci.go
donpenney: Wouldn't pidFile contain the pid for the exec probe, which we're getting from cmd.Process.Pid for the call to moveExecToCgroup? We're getting the container pid here to lookup the pod's cgroup.

haircommander: ah I see, makes sense to me

internal/oci/runtime_oci.go
donpenney: I'll look into this. Thanks for the suggestion!

donpenney: Moved the function to cgmgr.go. I don't see a way to merge this with CreateContainer, though

donpenney:
Rebased to pick up newly merged PR #5823

internal/oci/runtime_oci.go
```diff
@@ -521,6 +544,42 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 
 	// We don't need childPipe on the parent side
 	childPipe.Close()
+	childStartPipe.Close()
+
+	if r.handler.MonitorExecCgroup && r.config.InfraCtrCPUSet != "" {
```
haircommander, line 549: we need to defer the container process kill if it errors here (with a closure like in CreateContainer), otherwise we'll leak a process because conmon will be waiting forever on the pipe

donpenney: thanks for the comment... I've updated the code to do the same kill as CreateContainer

haircommander:
I would love to dedup this code in the future if possible, but I would not consider that blocking it in the first place

/approve

haircommander:
@saschagrunert PTAL

donpenney:
> 

Done. If monitor_exec_cgroup is set to a string other than "container", you'll see an error log like the following:

I0509 16:46:55.212517 3089960 prober.go:121] "Probe failed" probeType="Readiness" pod="xxxxxx" podUID=c6ba8458475c9dd26941a04586b8e318 containerName="xxxxxx" probeResult=failure output="Unsupported monitor_exec_cgroup value: not-container"

PR #5840, opened 2022-05-04
Move WillRunSystemd call after iterating the mounts 

harche:

#### What type of PR is this?

/kind bug

The changes introduced in https://github.com/cri-o/cri-o/pull/5778 weren't perfect. When iterating over the mount options of the sysfs mount provided in the pod definition will not have `rbin` or `bind` set. This causes that fix to not work. I might have goofed up in testing and might have tested with https://github.com/cri-o/cri-o/compare/240e311ea8e9daeb2ba34a2e81000c4560215421..9a7037b9f87f94120e3f8df33e8cd3729a60bb38 instead of the final version by mistake. Apologies for that. 

But I just tested this patch with 4.10 and 4.11 nightlies and it seem to fix this issue now. 

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=2064741

```release-note
None
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -546,14 +546,6 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		return nil, err
 	}
 
-	if ctr.WillRunSystemd() {
-		processLabel, err = selinux.InitLabel(processLabel)
-		if err != nil {
-			return nil, err
-		}
-		setupSystemd(specgen.Mounts(), *specgen)
-	}
-
 	// When running on cgroupv2, automatically add a cgroup namespace for not privileged containers.
 	if !ctr.Privileged() && node.CgroupIsV2() {
 		if err := specgen.AddOrReplaceLinuxNamespace(string(rspec.CgroupNamespace), ""); err != nil {
@@ -739,6 +731,14 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 			Source:      m.Source,
 		}
 		ctr.SpecAddMount(rspecMount)
+	}
+
+	if ctr.WillRunSystemd() {
+		processLabel, err = selinux.InitLabel(processLabel)
+		if err != nil {
+			return nil, err
+		}
+		setupSystemd(specgen.Mounts(), *specgen)
 	}
 
 	if s.ContainerServer.Hooks != nil {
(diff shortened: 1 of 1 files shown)


harche:
cc @haircommander

haircommander:
@giuseppe LGTY? if so please approve

giuseppe:
Isn't that the expected thing?  If `sysfs` is not a bind mount, then we need the mount for systemd.

In the new version, wouldn't `NoCgroupMount` always find `/sys/fs/cgroup` because it is part of the specs?

harche:
> Isn't that the expected thing? If `sysfs` is not a bind mount, then we need the mount for systemd.
> 
> In the new version, wouldn't `NoCgroupMount` always find `/sys/fs/cgroup` because it is part of the specs?

@giuseppe Whether pod spec adds volume `/sys/fs/cgroup` or not, the mounts of the container look like this,

```
May 05 07:02:54 harpatil9997-rts8r-worker-a-n69nr crio[24743]: time="2022-05-05 07:02:54.023162884Z" level=info msg="TEST1111: SSSSSSS [{Destination:/proc Type:proc Source:proc Options:[nosuid noexec nodev]} {Destination:/dev Type:tmpfs Source:tmpfs Options:[nosuid strictatime mode=755 size=65536k]} {Destination:/dev/pts Type:devpts Source:devpts Options:[nosuid noexec newinstance ptmxmode=0666 mode=0620 gid=5]} {Destination:/dev/shm Type:tmpfs Source:shm Options:[nosuid noexec nodev mode=1777 size=65536k]} {Destination:/dev/mqueue Type:mqueue Source:mqueue Options:[nosuid noexec nodev]} {Destination:/sys Type:sysfs Source:sysfs Options:[nosuid noexec nodev rw rslave]} {Destination:/sys/fs/cgroup Type:cgroup Source:cgroup Options:[nosuid noexec nodev rw relatime rslave]}]" id=2e08524e-9914-4ee4-9212-52f540848d09 name=/runtime.v1.RuntimeService/CreateContainer
```

As you can see,  [this condition](https://github.com/cri-o/cri-o/pull/5840/files#diff-1e0877112035840d3e2f073c693665f8daffa241a440e54a866ad7382b510ea2L1112) in the old code will never be true because, crio always mounts `sysfs` as type `sysfs` and `/sys/fs/cgroup` as type `cgroup`.  That's why I have proposed to remove it. 

I tested the code in this PR with following 2 pod yamls,

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: tmp-pod2
spec:
  containers:
  - name: tmp-pod2
    image: registry.access.redhat.com/ubi7/ubi-init:latest
    tty: true
    securityContext:
      privileged: true
    volumeMounts:
    - name: cgroup
      mountPath: /sys/fs/cgroup
  volumes:
  - name: cgroup
    hostPath:
      path: /sys/fs/cgroup
```
and 
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: tmp-pod2
spec:
  containers:
  - name: tmp-pod2
    image: registry.access.redhat.com/ubi7/ubi-init:latest
    tty: true
    securityContext:
      privileged: true
```

Do you have any pod yaml that can break because of the code in this PR? I can test it on my local cluster. 

cc @haircommander

giuseppe (approved):
LGTM

haircommander:
/lgtm

Good teamwork here, thanks to you both!

PR #5843, opened 2022-05-04
crio:fix a bug about log container

jwcesign:
crio:fix a bug about log container

Signed-off-by: jiang wei <jwcesign@gmail.com>

#### What type of PR is this?
/kind bug

When container is in unknown state, will have this error when log it(using critl and kubectl).
image

fix bug to log container, in upper layer interface, we need `status.CreatedAt`

none

```release-note
none
```

### DIFF
--- a/server/container_status.go
+++ b/server/container_status.go
@@ -62,19 +62,16 @@ func (s *Server) ContainerStatus(ctx context.Context, req *types.ContainerStatus
 		cState = c.State()
 	}
 
-	created := c.CreatedAt().UnixNano()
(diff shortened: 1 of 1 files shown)


jwcesign:
assign @haircommander

jwcesign:
I check the integration test case, the looks like unpause container work not on time sometimes, it case failed case when finish. 
So I plan to change the logic to delete container when it's in paused state with unpause it first.

haircommander:
yeah that make sense to me! we're going to cut 1.24.0 today, so it'll have to be backported

/lgtm

PR #5876, opened 2022-05-15
feature:Try to force delete ctr when in paused state

jwcesign:
Signed-off-by: jiang wei <jwcesign@gmail.com>

#### What type of PR is this?

/kind feature

When in pause state, there is case to delete it directly(for serverless framework-knative), but it will delete failed. So this is to delete it force

none

related info: https://github.com/cri-o/cri-o/pull/5843#issuecomment-1118084548

```release-note
none
```


jwcesign:
cc @haircommander

internal/lib/remove.go
```diff
@@ -20,17 +20,23 @@ func (c *ContainerServer) Remove(ctx context.Context, container string, force bo
...
-	case oci.ContainerStateCreated, oci.ContainerStateRunning:
 		if force {
-			if err = c.StopContainer(ctx, ctr, 10); err != nil {
-				return "", errors.Wrapf(err, "unable to stop container %s", ctrID)
+			if err = c.runtime.UnpauseContainer(ctx, ctr); err != nil {
+				return "", errors.Wrapf(err, "unable to unpause container %s", ctrID)
 			}
 		} else {
+			return "", errors.Errorf("cannot remove paused container %s", ctrID)
+		}
```
haircommander, line 29: can we rework this to be
```
if !force {
     return ....
}
if err = c.runtime.UnpauseContainer...
    return...
}
```

internal/lib/stop.go
```diff
@@ -11,6 +10,14 @@ import (
 
 // ContainerStop stops a running container with a grace period (i.e., timeout).
 func (c *ContainerServer) StopContainer(ctx context.Context, ctr *oci.Container, timeout int64) error {
+	if ctr.State().Status == oci.ContainerStatePaused {
+		if err := c.Runtime().UnpauseContainer(ctx, ctr); err != nil {
```
jwcesign, line 14: shoud we give a pause.go/unpause.go? @haircommander

server/container_list.go
```diff
@@ -84,7 +84,7 @@ func (s *Server) ListContainers(ctx context.Context, req *types.ListContainersRe
 		switch cState.Status {
 		case oci.ContainerStateCreated:
 			rState = types.ContainerState_CONTAINER_CREATED
-		case oci.ContainerStateRunning:
+		case oci.ContainerStateRunning, oci.ContainerStatePaused:
```
jwcesign, line 87: herer I check the code of containerd, u can check https://github.com/containerd/containerd/blob/77d53d2d230c3bcd3f02e6f493019a72905c875b/pkg/cri/store/container/status.go#L103 @haircommander

haircommander: intersting, I could be compelled by this. @saschagrunert WDYT

jwcesign: if its unknow state, kubelet will delete the paused ctr and restart one new.

jwcesign:
image
Maybe can fix this?

server/container_list.go
saschagrunert: I think hiding the information from the kubelet is the only way to make it work. It would require a kubernetes change to display the paused statuts to the end-user.

jwcesign: I think this is the reason why containrd do like this

jwcesign:
Hi, I am confused about why failed, https://ci.openshift.redhat.com/jenkins/job/test_pull_request_crio_integration_rhel/27206/s3/
Can u give a hint? @haircommander

internal/lib/remove.go
```diff
@@ -20,17 +20,22 @@ func (c *ContainerServer) Remove(ctx context.Context, container string, force bo
...
-		} else {
+		if !force {
 			return "", errors.Errorf("cannot remove running container %s", ctrID)
 		}
 	}
 
+	if err = c.StopContainer(ctx, ctr, 10); err != nil {
+		return "", errors.Wrapf(err, "unable to stop container %s", ctrID)
+	}
+
```
haircommander, line 38: this StopContainer looks like it should be inside of the block above

internal/lib/remove.go
```diff
@@ -20,17 +20,22 @@ func (c *ContainerServer) Remove(ctx context.Context, container string, force bo
 	cStatus := ctr.State()
 	switch cStatus.Status {
 	case oci.ContainerStatePaused:
-		return "", errors.Errorf("cannot remove paused container %s", ctrID)
+		if !force {
+			return "", errors.Errorf("cannot remove paused container %s", ctrID)
+		}
+		if err = c.runtime.UnpauseContainer(ctx, ctr); err != nil {
+			return "", errors.Wrapf(err, "unable to unpause container %s", ctrID)
+		}
```
haircommander, line 28: If we're calling unpause in StopContainer, we can merge these two cases
```
	case oci.ContainerStateCreated, oci.ContainerStatePaused, oci.ContainerStateRunning:
	    if !force {
                return ...
            }
            c.StopContainer(...)
        }
```

haircommander:
it looks like the tests are not progressing in `teset/ctr.bats` have you run that suite locally? it's possible my nits will help

internal/lib/remove.go
jwcesign: fixde

internal/lib/remove.go
jwcesign: fixed

internal/lib/stop.go
```diff
@@ -11,6 +10,15 @@ import (
 
 // ContainerStop stops a running container with a grace period (i.e., timeout).
 func (c *ContainerServer) StopContainer(ctx context.Context, ctr *oci.Container, timeout int64) error {
+	cStatus := ctr.StateNoLock()
```
jwcesign, line 13: if I use` ctr.State()`, the it case will failed, I think there is bug in `c.runtime.UpdateContainerStatus` which take the lock for long time. I will find it out latter

jwcesign:
cc @haircommander , everything is ok now

internal/lib/stop.go
```diff
@@ -2,7 +2,6 @@ package lib
 
 import (
 	"context"
-
```
haircommander, line 5: we typically separate the stdlib imports from the ones imported by go modules

jwcesign: fixed

server/container_list.go
TomSweeneyRedHat: This just seems wrong, but I'll let you all noodle it out.

PR #5895, opened 2022-05-18
oci: kill children of container if it is in the host pid namespace

haircommander:
When PID 1 of a PID namespace is killed, all of the remaining processes are also killed.
However, for containers with a host PID namespace, this does not happen.

While we will eventually kill these processes when the kubelet tears down the sandbox's cgroup,
it looks like a leak (and could cause weird situations with duplicated processes running)

Fix it by killing the processes in this case.

#### What type of PR is this?

/kind bug

```release-note
Fix a bug where child processes of containers in the host's PID namespace appear to leak after the child exits
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -613,3 +613,16 @@ func (c *Container) RemoveManagedPIDNamespace() error {
 func (c *Container) IsInfra() bool {
 	return c.ID() == c.Sandbox()
 }
+
+// nodeLevelPIDNamespace searches through the container spec to see if there is
+// a PID namespace specified. If not, it returns `true` (because the runtime spec
+// defines a node level namespace as being absent from the Namespaces list)
+func (c *Container) nodeLevelPIDNamespace() bool {
+	for i := range c.spec.Linux.Namespaces {
+		// If it's specified in the namespace list, then it is something other than Node level
+		if c.spec.Linux.Namespaces[i].Type == specs.PIDNamespace {
+			return false
+		}
+	}
+	return true
+}
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1001,7 +1001,12 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 		// Collect metric by container name
 		metrics.Instance().MetricContainersOOMCountTotalInc(c.Name())
 	}
-
+	// If this container had a node level PID namespace, then any children processes will be leaked to init.
+	// Eventually, the processes will get cleaned up when the pod cgroup is cleaned by the kubelet,
+	// but this situation is atypical and should be avoided.
+	if c.nodeLevelPIDNamespace() {
+		return r.signalContainer(c, syscall.SIGKILL, true)
+	}
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


haircommander:
related to https://bugzilla.redhat.com/show_bug.cgi?id=2078634

internal/oci/container.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/cri-o/cri-o/internal/config/nsmgr"
 	ann "github.com/cri-o/cri-o/pkg/annotations"
 	json "github.com/json-iterator/go"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
```
rphillips, line 21: this import is defined on the next line as 'specs'.

internal/oci/runtime_oci.go
```diff
@@ -1001,7 +1001,12 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 		// Collect metric by container name
 		metrics.Instance().MetricContainersOOMCountTotalInc(c.Name())
 	}
-
+	// If this container had a node level PID namespace, then any children processes will be leaked to init.
+	// Eventually, the processes will get cleaned up when the pod cgroup is cleaned by the kubelet,
+	// but this situation is atypical and should be avoided.
+	if c.nodeLevelPIDNamespace() {
```
mrunalp, line 1007: Is there a race possible where this could end up getting called twice? Is the call re-entrant?
Another approach may be to add a call to signal in the exit monitor.

test/ctr.bats
```diff
@@ -957,3 +969,30 @@ function check_oci_annotation() {
...
+
+	jq '	  .image.image = "quay.io/crio/redis:alpine"
+		| .linux.security_context.namespace_options.pid = 2
+		| .command = ["/bin/sh", "-c", "sleep 1m& exec sleep 2m"]' \
+		"$TESTDATA"/container_config.json > "$newconfig"
+
+	ctr_id=$(crictl run "$newconfig" "$newsandbox")
+	processes=$(list_all_children "$(pidof conmon)")
+
+	pid=$(runtime list -f json | jq .[].pid)
```
littlejawa, line 988: "runtime" is a wrapper to call the runtime binary directly from the command line. It assumes a runc-style runtime.
This is not compatible with the containerd-shim interface. This is why the kata CI is failing.
You may want to skip this test when RUNTIME_TYPE=vm, at least until we find a shim-compatible way of getting the same information

littlejawa: See #5896 for reference

internal/oci/runtime_oci.go
haircommander: since this call happens after we set the ExitCode for the container, there's no way to get back here (as we stop above if ExitCode has been defined)

test/ctr.bats
haircommander: good catch, fixed!

internal/oci/runtime_oci.go
```diff
@@ -1051,8 +1044,21 @@ func (r *runtimeOCI) SignalContainer(ctx context.Context, c *Container, sig sysc
 		return errors.Errorf("unable to find signal %s", sig.String())
 	}
 
+	return r.signalContainer(ctx, c, sig, false)
+}
+
+func (r *runtimeOCI) signalContainer(ctx context.Context, c *Container, sig syscall.Signal, all bool) error {
```
TomSweeneyRedHat, line 1050: do you anticipate using ctx later?

haircommander: probably not, I've dropped

haircommander:
addressed comments and fixed tests, PTAL @cri-o/cri-o-maintainers

TomSweeneyRedHat:
LGTM

saschagrunert (approved):
/lgtm

PR #5898, opened 2022-05-19
[1.19] oci: kill children of container if it is in the host pid namespace

haircommander:
When PID 1 of a PID namespace is killed, all of the remaining processes are also killed.
However, for containers with a host PID namespace, this does not happen.

While we will eventually kill these processes when the kubelet tears down the sandbox's cgroup,
it looks like a leak (and could cause weird situations with duplicated processes running)

Fix it by killing the processes in this case.

#### What type of PR is this?

/kind bug
/hold

```release-note
Fix a bug where child processes of containers in the host's PID namespace appear to leak after the child exits
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -533,3 +533,16 @@ func (c *Container) ShouldBeStopped() error {
 func (c *Container) Spoofed() bool {
 	return c.spoofed
 }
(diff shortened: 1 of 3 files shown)


saschagrunert:

PR #5914, opened 2022-05-27
[1.21] utils/RunUnderSystemdScope: fix wrt channel deadlock

kolyshkin:
As seen in [1], sometimes coreos/go-systemd/dbus package deadlocks: the
jobCompete is stuck trying to send job result string to the channel
while holding the jobListener lock, while startJob (called by
StartTransientUnit) waits for the same lock.

Alas, it is not clear why the channel is not being read, nor was I able
to reproduce it locally.

Make the job result channel buffered, so jobJistener won't block on
channel send and thus StartTransientUnit won't be stuck either.

While at it,

 - move the error wrapping out of mgr.RetryOnDisconnect function,
   and use fmt.Errorf with %w instead of obsoleted errors.Wrap;

 - improve error messages, printing the systemd unit name (so we can
   check it in systemd log);

 - do check the job result string -- in case it is not "done",
   return an error back to the caller, which should help avoid other
   issues down the line.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2082344

```release-note
Avoid a deadlock while moving conmon to a different system scope (RHBZ#2082344)
```

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -69,18 +69,24 @@ func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName
 	if slice != "" {
 		properties = append(properties, systemdDbus.PropSlice(slice))
 	}
-	ch := make(chan string)
+	// Make a buffered channel so that the sender (go-systemd's jobComplete)
+	// won't be blocked on channel send while holding the jobListener lock
+	// (RHBZ#2082344).
+	ch := make(chan string, 1)
 	if err := mgr.RetryOnDisconnect(func(c *systemdDbus.Conn) error {
 		_, err = c.StartTransientUnit(unitName, "replace", properties, ch)
-		return errors.Wrap(err, "start transient unit")
-	}); err != nil {
 		return err
+	}); err != nil {
+		return fmt.Errorf("start transient unit %q: %w", unitName, err)
 	}
 
-	// Block until job is started
+	// Wait for the job status.
 	select {
-	case <-ch:
+	case s := <-ch:
 		close(ch)
+		if s != "done" {
+			return fmt.Errorf("error moving conmon with pid %d to systemd unit %s: got %s", pid, unitName, s)
+		}
 	case <-time.After(time.Minute * 6):
 		// This case is a work around to catch situations where the dbus library sends the
 		// request but it unexpectedly disappears. We set the timeout very high to make sure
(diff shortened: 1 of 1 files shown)


utils/utils.go
```diff
@@ -69,26 +69,32 @@ func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName
...
-	// Block until job is started
+	// Wait for the job status.
 	select {
-	case <-ch:
+	case s := <-ch:
 		close(ch)
+		if s != "done" {
+			return fmt.Errorf("error moving conmon with pid %d to systemd unit %s: got %s", pid, unitName, s)
+		}
 	case <-time.After(time.Minute * 6):
```
haircommander, line 90: I find myself wondering if we still need this case. Ideally, the buffered channel would fix the deadlock long term. From the sounds of it, having the request timeout just caused a less obvious deadlock

kolyshkin:
lint job fails since it uses an old version of golangci-lint but a new version of golang (IOW release-1.21 branch needs some love).

utils/utils.go
kolyshkin: I believe we do. The buffered channel fixes the deadlock, but if we haven't received a reply from systemd we should still say so and return an error.

haircommander:
/retest
/approve

kolyshkin:
> lint job fails since it uses an old version of golangci-lint but a new version of golang (IOW release-1.21 branch needs some love).

Being fixed in https://github.com/cri-o/cri-o/pull/5918

rphillips:
@kolyshkin I think we should get this into 'main' first.

Can we give the customer a test build to try out?

kolyshkin:
> @kolyshkin I think we should get this into 'main' first.

Does it matter if we do the backport or forward port?

> Can we give the customer a test build to try out?

Yes, that would be a good thing to do, despite the lack of clear repro.

kolyshkin:
Forward-port to main branch: https://github.com/cri-o/cri-o/pull/5922

kolyshkin:
> let's start with main and pull back sequentially, so we order when they merge by inverse relative stability (1.21 should be _very_ stable, 1.24 should be stable)

Meaning I can still do a forward-port, but the order of merging is important, right?

Here's one for main: https://github.com/cri-o/cri-o/pull/5922

haircommander:
related https://bugzilla.redhat.com/show_bug.cgi?id=2095387

PR #5918, opened 2022-05-31
[1.21] ci/gha: fixup golangci-lint job

kolyshkin:
Apparently, ubuntu-latest now comes with a recent version of golang,
which is too new for golangci-lint we use in this branch. As a result,
we have errors like

	 ...could not import internal/goarch (-: could not load export data: cannot import \"internal/goarch\" (unknown iexport format version 2), export data is newer version - update tool)))))]]"

Explicitly installing golang 1.16 and upgrading the action itself should fix the issue.

/kind ci

```release-note
NONE
```

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -12,7 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


kolyshkin:
Weird, it is still using go 1.18.2.

kolyshkin:
OK, updating the action itself helped

PR #5917, opened 2022-05-31
Bump golang dependencies

saschagrunert:
#### What type of PR is this?

/kind dependency-change

#### What this PR does / why we need it:
Update some non-breaking golang dependencies to their latest versions.
#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
Updated golang dependencies
```

PR #5506, opened 2021-12-14
deps: bump runc/libcontainer 1.0.3 -> 1.1.2

kolyshkin:
#### What type of PR is this?

/kind dependency-change

#### What this PR does / why we need it:

Bump runc from 1.0.3 to 1.1.2. For release notes, see https://github.com/opencontainers/runc/releases

Modify internal/config/cgmgr/cgroupfs.go to use a new (better, simpler)
way of getting a cgroup manager.

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

```release-note
Update runc/libcontainer dependency to v1.1.2
```

PR #5922, opened 2022-05-31
utils/RunUnderSystemdScope: fix wrt channel deadlock

kolyshkin:
As seen in [1], sometimes coreos/go-systemd/dbus package deadlocks: the
jobCompete is stuck trying to send job result string to the channel
while holding the jobListener lock, while startJob (called by
StartTransientUnit) waits for the same lock.

Alas, it is not clear why the channel is not being read, nor was I able
to reproduce it locally.

Make the job result channel buffered, so jobJistener won't block on
channel send and thus StartTransientUnit won't be stuck either.

While at it,

 - move the error wrapping out of mgr.RetryOnDisconnect function,
   and use fmt.Errorf with %w instead of obsoleted errors.Wrap;

 - improve error messages, printing the systemd unit name (so we can
   check it in systemd log);

 - do check the job result string -- in case it is not "done",
   return an error back to the caller, which should help avoid other
   issues down the line.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2082344

/kind bug

```release-note
Fix a rare deadlock while communicating to systemd (RHBZ 2082344)
```
----

This is a forward-port of https://github.com/cri-o/cri-o/pull/5914 to main branch.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -70,16 +70,24 @@ func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName
 	if slice != "" {
 		properties = append(properties, systemdDbus.PropSlice(slice))
 	}
-	ch := make(chan string)
+	// Make a buffered channel so that the sender (go-systemd's jobComplete)
+	// won't be blocked on channel send while holding the jobListener lock
+	// (RHBZ#2082344).
+	ch := make(chan string, 1)
 	if err := mgr.RetryOnDisconnect(func(c *systemdDbus.Conn) error {
 		_, err = c.StartTransientUnitContext(ctx, unitName, "replace", properties, ch)
-		return errors.Wrap(err, "start transient unit")
-	}); err != nil {
 		return err
+	}); err != nil {
+		return fmt.Errorf("start transient unit %q: %w", unitName, err)
 	}
 
-	// Block until job is started
+	// Wait for the job status.
 	select {
+	case s := <-ch:
+		close(ch)
+		if s != "done" {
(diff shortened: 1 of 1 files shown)


kolyshkin:
Test failures _seem_ to be unrelated.

haircommander:
e2e_fedora and e2e_crun* are unrelated (being investigated in https://github.com/cri-o/cri-o/pull/5924). however, agnostic and gcp may not be

kolyshkin:
rebased

kolyshkin:
Failures in e2e-gcp and e2e-agnostic are the same as in #5506  which makes me think they are unrelated to this PR (or #5506).

Failures in e2e-cgroupv2 and e2e-fedora:

> Kubernetes e2e suite: [sig-network] Services should be rejected for evicted pods (no endpoints exist) expand_less
> Kubernetes e2e suite: [sig-node] Pods Extended Pod Container lifecycle evicted pods should be terminal expand_more

are the same as in e.g. and #5917, so also seem unrelated.

saschagrunert (approved):

utils/utils.go
```diff
@@ -70,16 +70,24 @@ func RunUnderSystemdScope(mgr *dbusmgr.DbusConnManager, pid int, slice, unitName
...
-	// Block until job is started
+	// Wait for the job status.
 	select {
+	case s := <-ch:
+		close(ch)
+		if s != "done" {
+			return fmt.Errorf("error moving conmon with pid %d to systemd unit %s: got %s", pid, unitName, s)
+		}
 	case <-ch:
 		close(ch)
```
kolyshkin, line 92: Oops, this should not be there.

kolyshkin: Opened https://github.com/cri-o/cri-o/pull/6124 to address it.

PR #5927, opened 2022-06-02
Add annotations to allow power management configuration

bartwensley:
Adding the following high performance feature annotations:

cpu-c-states.crio.io: \<enable|disable>
This annotation will enable or disable c-states for each cpu by writing to the
power/pm_qos_resume_latency_us for the cpu under sysfs.

cpu-freq-governor.crio.io: \<any supported cpu frequency governor>
This annotation will set the cpufreq governor for each cpu by writing to the
cpufreq/scaling_governor for the cpu under sysfs.

When a container is stopped, the original pm_qos_resume_latency_us and
scaling_governor values are restored.

As with the other high performance feature annotations, the new annotations
are only considered for pods with a guaranteed QoS Class.

Signed-off-by: Bart Wensley <bwensley@redhat.com>

#### What type of PR is this?

/kind feature

In order to support high performance (i.e. low latency) containers on a server, the user will typically configure the BIOS and/or kernel options to disable c-states and use the performance cpufreq governor for all CPUs in the server. This ensures the best performance and lowest latency for all containers run on that server, but it also has the effect of consuming a large amount of power.

In many cases, users do not need all their containers to run in this high performance (and high power consumption) mode. To allow for power savings in cases where there is a mix of high performance and regular containers, we need to allow the server to be configured for lower power usage by default (e.g. by enabling some c-states or by using a cpufreq governor like schedutil) and then only configure specific containers for high performance. This can be done by introducing annotations that allow c-states and the cpufreq governor to be configured for individual pods/containers. This is similar to existing high performance annotations like cpu-load-balancing.crio.io (for example) that improve the performance of a container by disabling functions that could impact performance of the cpus the container is using.

None

None

```release-note
This PR introduces additional annotations that can be optionally used to configure power management features for containers. It does not impact existing users who do not use the new annotations.
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks.go
@@ -2,9 +2,11 @@ package runtimehandlerhooks
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -36,10 +38,13 @@ const (
 const (
 	annotationTrue       = "true"
 	annotationDisable    = "disable"
+	annotationEnable     = "enable"
 	schedDomainDir       = "/proc/sys/kernel/sched_domain"
 	cgroupMountPoint     = "/sys/fs/cgroup"
 	irqBalanceBannedCpus = "IRQBALANCE_BANNED_CPUS"
 	irqBalancedName      = "irqbalance"
+	sysCPUDir            = "/sys/devices/system/cpu"
+	sysCPUSaveDir        = "/var/run/crio/cpu"
 )
 
 // HighPerformanceHooks used to run additional hooks that will configure a system for the latency sensitive workloads
@@ -90,6 +95,34 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 		}
 	}
 
+	// Configure c-states for the container CPUs.
+	if configure, value := shouldCStatesBeConfigured(s.Annotations()); configure {
+		log.Infof(ctx, "Configure c-states for container %q to %q", c.ID(), value)
+		switch value {
+		case annotationEnable:
+			// Enable all c-states.
+			if err := setCPUPMQOSResumeLatency(c, "0"); err != nil {
+				return fmt.Errorf("set CPU PM QOS resume latency: %w", err)
+			}
+		case annotationDisable:
+			// Lock the c-state to C0.
+			if err := setCPUPMQOSResumeLatency(c, "n/a"); err != nil {
+				return fmt.Errorf("set CPU PM QOS resume latency: %w", err)
+			}
+		default:
+			return fmt.Errorf("invalid annotation value %s", value)
+		}
+	}
+
+	// Configure cpu freq governor for the container CPUs.
+	if configure, value := shouldFreqGovernorBeConfigured(s.Annotations()); configure {
+		log.Infof(ctx, "Configure cpu freq governor for container %q to %q", c.ID(), value)
+		// Set the cpu freq governor to specified value.
+		if err := setCPUFreqGovernor(c, value); err != nil {
+			return fmt.Errorf("set CPU scaling governor: %w", err)
+		}
+	}
+
 	return nil
 }
 
@@ -125,6 +158,24 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s 
 
 	// no need to reverse the cgroup CPU CFS quota setting as the pod cgroup will be deleted anyway
 
+	// Restore the c-state configuration for the container CPUs (only do this when the annotation is
+	// present - without the annotation we do not modify the c-state).
+	if configure, _ := shouldCStatesBeConfigured(s.Annotations()); configure {
+		// Restore the original resume latency value.
+		if err := setCPUPMQOSResumeLatency(c, ""); err != nil {
+			return fmt.Errorf("set CPU PM QOS resume latency: %w", err)
+		}
+	}
+
+	// Restore the cpu freq governor for the container CPUs (only do this when the annotation is
+	// present - without the annotation we do not modify the governor).
+	if configure, _ := shouldFreqGovernorBeConfigured(s.Annotations()); configure {
+		// Restore the original scaling governor.
+		if err := setCPUFreqGovernor(c, ""); err != nil {
+			return fmt.Errorf("set CPU scaling governor: %w", err)
+		}
+	}
+
 	return nil
 }
 
@@ -153,6 +204,16 @@ func shouldIRQLoadBalancingBeDisabled(annotations fields.Set) bool {
 
 	return annotations[crioannotations.IRQLoadBalancingAnnotation] == annotationTrue ||
 		annotations[crioannotations.IRQLoadBalancingAnnotation] == annotationDisable
+}
+
+func shouldCStatesBeConfigured(annotations fields.Set) (present bool, value string) {
+	value, present = annotations[crioannotations.CPUCStatesAnnotation]
+	return
+}
+
+func shouldFreqGovernorBeConfigured(annotations fields.Set) (present bool, value string) {
+	value, present = annotations[crioannotations.CPUFreqGovernorAnnotation]
+	return
 }
 
 func annotationValueDeprecationWarning(annotation string) string {
@@ -347,6 +408,188 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
 			return err
 		}
 		if err := os.WriteFile(parentCfsQuotaPath, []byte("-1"), 0o644); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// setCPUPMQOSResumeLatency sets the pm_qos_resume_latency_us for a cpu and stores the original
+// value so it can be restored later. If the latency is an empty string, the original latency
+// value is restored.
+func setCPUPMQOSResumeLatency(c *oci.Container, latency string) error {
+	return doSetCPUPMQOSResumeLatency(c, latency, sysCPUDir, sysCPUSaveDir)
+}
+
+// doSetCPUPMQOSResumeLatency facilitates unit testing by allowing the directories to be specified as parameters.
+func doSetCPUPMQOSResumeLatency(c *oci.Container, latency, cpuDir, cpuSaveDir string) error {
+	lspec := c.Spec().Linux
+	if lspec == nil ||
+		lspec.Resources == nil ||
+		lspec.Resources.CPU == nil ||
+		lspec.Resources.CPU.Cpus == "" {
+		return fmt.Errorf("find container %s CPUs", c.ID())
+	}
+
+	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
+	if err != nil {
+		return err
+	}
+
+	for _, cpu := range cpus.ToSlice() {
+		latencyFile := fmt.Sprintf("%s/cpu%d/power/pm_qos_resume_latency_us", cpuDir, cpu)
+		cpuPowerSaveDir := fmt.Sprintf("%s/cpu%d/power", cpuSaveDir, cpu)
+		latencyFileOrig := path.Join(cpuPowerSaveDir, "pm_qos_resume_latency_us")
+
+		if latency != "" {
+			// Retrieve the current latency.
+			latencyOrig, err := os.ReadFile(latencyFile)
+			if err != nil {
+				return err
+			}
+
+			// Save the current latency so we can restore it later.
+			err = os.MkdirAll(cpuPowerSaveDir, 0o750)
+			if err != nil {
+				return err
+			}
+			err = os.WriteFile(latencyFileOrig, latencyOrig, 0o644)
+			if err != nil {
+				return err
+			}
+
+			// Update the pm_qos_resume_latency_us.
+			err = os.WriteFile(latencyFile, []byte(latency), 0o644)
+			if err != nil {
+				return err
+			}
+
+			continue
+		}
+
+		// Retrieve the original latency.
+		latencyOrig, err := os.ReadFile(latencyFileOrig)
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				// The latency may have already been restored by a previous invocation of the hook.
+				return nil
+			}
+			return err
+		}
+
+		// Restore the original latency.
+		err = os.WriteFile(latencyFile, latencyOrig, 0o644)
+		if err != nil {
+			return err
+		}
+
+		// Remove the saved latency.
+		err = os.Remove(latencyFileOrig)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// isCPUGovernorSupported checks whether the cpu governor is supported for the specified cpu.
+func isCPUGovernorSupported(governor, cpuDir string, cpu int) error {
+	// Get available cpu scaling governors.
+	availGovernorFile := fmt.Sprintf("%s/cpu%d/cpufreq/scaling_available_governors", cpuDir, cpu)
+	availGovernors, err := os.ReadFile(availGovernorFile)
+	if err != nil {
+		return err
+	}
+
+	// Is the scaling governor supported?
+	for _, availableGovernor := range strings.Fields(string(availGovernors)) {
+		if availableGovernor == governor {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("governor %s not available for cpu %d", governor, cpu)
+}
+
+// setCPUFreqGovernor sets the scaling_governor for a cpu and stores the original
+// value so it can be restored later. If the governor is an empty string, the original
+// scaling_governor value is restored.
+func setCPUFreqGovernor(c *oci.Container, governor string) error {
+	return doSetCPUFreqGovernor(c, governor, sysCPUDir, sysCPUSaveDir)
+}
+
+// doSetCPUFreqGovernor facilitates unit testing by allowing the directories to be specified as parameters.
+func doSetCPUFreqGovernor(c *oci.Container, governor, cpuDir, cpuSaveDir string) error {
+	lspec := c.Spec().Linux
+	if lspec == nil ||
+		lspec.Resources == nil ||
+		lspec.Resources.CPU == nil ||
+		lspec.Resources.CPU.Cpus == "" {
+		return fmt.Errorf("find container %s CPUs", c.ID())
+	}
+
+	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
+	if err != nil {
+		return err
+	}
+
+	for _, cpu := range cpus.ToSlice() {
+		governorFile := fmt.Sprintf("%s/cpu%d/cpufreq/scaling_governor", cpuDir, cpu)
+		cpuFreqSaveDir := fmt.Sprintf("%s/cpu%d/cpufreq", cpuSaveDir, cpu)
+		governorFileOrig := path.Join(cpuFreqSaveDir, "scaling_governor")
+
+		if governor != "" {
+			// Retrieve the current scaling governor.
+			governorOrig, err := os.ReadFile(governorFile)
+			if err != nil {
+				return err
+			}
+
+			// Is the scaling governor supported?
+			if err := isCPUGovernorSupported(governor, cpuDir, cpu); err != nil {
+				return err
+			}
+
+			// Save the current governor so we can restore it later.
+			err = os.MkdirAll(cpuFreqSaveDir, 0o750)
+			if err != nil {
+				return err
+			}
+			err = os.WriteFile(governorFileOrig, governorOrig, 0o644)
+			if err != nil {
+				return err
+			}
+
+			// Update the governor.
+			err = os.WriteFile(governorFile, []byte(governor), 0o644)
+			if err != nil {
+				return err
+			}
+
+			continue
+		}
+
+		// Retrieve the original scaling governor.
+		governorOrig, err := os.ReadFile(governorFileOrig)
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				// The governor may have already been restored by a previous invocation of the hook.
+				return nil
+			}
+			return err
+		}
+
+		// Restore the original governor.
+		err = os.WriteFile(governorFile, governorOrig, 0o644)
+		if err != nil {
+			return err
+		}
+
+		// Remove the saved governor.
+		err = os.Remove(governorFileOrig)
+		if err != nil {
 			return err
 		}
 	}
--- a/internal/runtimehandlerhooks/high_performance_hooks_test.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_test.go
@@ -228,6 +228,284 @@ var _ = Describe("high_performance_hooks", func() {
 		})
 	})
 
+	Describe("setCPUPMQOSResumeLatency", func() {
+		var pmQosResumeLatencyUs, pmQosResumeLatencyUsOriginal string
+		cpuDir := filepath.Join(fixturesDir, "cpu")
+		cpuSaveDir := filepath.Join(fixturesDir, "cpuSave")
+
+		//nolint:dupl
+		verifySetCPUPMQOSResumeLatency := func(latency string, expected string, expected_save string, expect_error bool) {
+			err := doSetCPUPMQOSResumeLatency(container, latency, cpuDir, cpuSaveDir)
+			if !expect_error {
+				Expect(err).ShouldNot(HaveOccurred())
+			} else {
+				Expect(err).Should(HaveOccurred())
+			}
+
+			if expected != "" {
+				for _, cpu := range []string{"cpu0", "cpu1"} {
+					content, err := os.ReadFile(filepath.Join(cpuDir, cpu, "power", "pm_qos_resume_latency_us"))
+					Expect(err).To(BeNil())
+
+					Expect(strings.Trim(string(content), "\n")).To(Equal(expected))
+				}
+			}
+
+			if expected_save != "" {
+				for _, cpu := range []string{"cpu0", "cpu1"} {
+					content, err := os.ReadFile(filepath.Join(cpuSaveDir, cpu, "power", "pm_qos_resume_latency_us"))
+					Expect(err).To(BeNil())
+					Expect(strings.Trim(string(content), "\n")).To(Equal(expected_save))
+				}
+			}
+		}
+
+		JustBeforeEach(func() {
+			// set container CPUs
+			container.SetSpec(
+				&specs.Spec{
+					Linux: &specs.Linux{
+						Resources: &specs.LinuxResources{
+							CPU: &specs.LinuxCPU{
+								Cpus: "0,1",
+							},
+						},
+					},
+				},
+			)
+
+			// create tests power files
+			for _, cpu := range []string{"cpu0", "cpu1"} {
+				powerDir := filepath.Join(cpuDir, cpu, "power")
+				err = os.MkdirAll(powerDir, os.ModePerm)
+				Expect(err).To(BeNil())
+
+				if pmQosResumeLatencyUs != "" {
+					err = os.WriteFile(filepath.Join(powerDir, "pm_qos_resume_latency_us"), []byte(pmQosResumeLatencyUs), 0o644)
+					Expect(err).To(BeNil())
+				}
+				if pmQosResumeLatencyUsOriginal != "" {
+					powerSaveDir := filepath.Join(cpuSaveDir, cpu, "power")
+					err = os.MkdirAll(powerSaveDir, os.ModePerm)
+					err = os.WriteFile(filepath.Join(powerSaveDir, "pm_qos_resume_latency_us"), []byte(pmQosResumeLatencyUsOriginal), 0o644)
+					Expect(err).To(BeNil())
+				}
+			}
+		})
+
+		AfterEach(func() {
+			for _, cpu := range []string{"cpu0", "cpu1"} {
+				if err := os.RemoveAll(filepath.Join(fixturesDir, cpu)); err != nil {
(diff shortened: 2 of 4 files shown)


internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,176 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			cpuScalingGovernorFileOrig := fmt.Sprintf("%s/cpu%d/cpufreq/scaling_governor", cpuSaveDir, cpu)
+			content, err := ioutil.ReadFile(cpuScalingGovernorFileOrig)
+			if err != nil {
+				if errors.Is(err, os.ErrNotExist) {
+					// the governor may have already been restored by a previous invocation of the hook
+					return nil
+				}
+				return err
+			}
+			governorOrig := strings.TrimSpace(string(content))
```
kolyshkin, line 570: It seems that this is not needed (in case you merely want to write this back to the kernel).

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,176 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+
+			// get available cpu scaling governors
+			cpuScalingAvailableGovernorFile := fmt.Sprintf("%s/cpu%d/cpufreq/scaling_available_governors", cpuDir, cpu)
+			content, err = ioutil.ReadFile(cpuScalingAvailableGovernorFile)
+			if err != nil {
+				return err
+			}
+			availableGovernors := strings.Split(strings.TrimSpace(string(content)), " ")
```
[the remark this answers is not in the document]
kolyshkin: Now, all CPUs have the same available governors, so you don't have to read it for every CPUI -- can only read it once.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,176 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+
+	for _, cpu := range cpus.ToSlice() {
+		if governor != "" {
+			// retrieve the current scaling governor
+			cpuScalingGovernorFile := fmt.Sprintf("%s/cpu%d/cpufreq/scaling_governor", cpuDir, cpu)
+			content, err := ioutil.ReadFile(cpuScalingGovernorFile)
+			if err != nil {
+				return err
+			}
+			governorOrig := strings.TrimSpace(string(content))
```
kolyshkin, line 520: This `[]byte` -> `string` conversion is entirely unnecessary, since you are only using this to write it to a file, in where you convert it back to `[]byte`. Trimming is also not necessary.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,176 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			cpuScalingGovernorFile := fmt.Sprintf("%s/cpu%d/cpufreq/scaling_governor", cpuDir, cpu)
+			err = ioutil.WriteFile(cpuScalingGovernorFile, []byte(governorOrig), 0o644)
+			if err != nil {
+				return err
+			}
+
+			// remove the saved governor
+			err = os.Remove(cpuScalingGovernorFileOrig)
+			if err != nil {
+				return err
```
kolyshkin, line 582: I think we need to ignore `ENOENT` (aka `os.ErrNotExist`) here.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,176 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			cpuPMQOSResumeLatencyFile := fmt.Sprintf("%s/cpu%d/power/pm_qos_resume_latency_us", cpuDir, cpu)
+			content, err := ioutil.ReadFile(cpuPMQOSResumeLatencyFile)
+			if err != nil {
+				return err
+			}
+			latencyOrig := strings.TrimSpace(string(content))
+
+			// save the current latency so we can restore it later
+			cpuPowerSaveDir := fmt.Sprintf("%s/cpu%d/power", cpuSaveDir, cpu)
+			err = os.MkdirAll(cpuPowerSaveDir, 0o750)
```
kolyshkin, line 449: why not `0o700`?

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,176 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+
+	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
+	if err != nil {
+		return err
+	}
+
+	for _, cpu := range cpus.ToSlice() {
+		if latency != "" {
+			// retrieve the current latency
+			cpuPMQOSResumeLatencyFile := fmt.Sprintf("%s/cpu%d/power/pm_qos_resume_latency_us", cpuDir, cpu)
```
kolyshkin, line 440: Since the function already have  `CPUPMQOSResumeLatency` in its name, you can use `file` as a variable name here, there won't be any ambiguity here.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,176 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			}
+			latencyOrig := strings.TrimSpace(string(content))
+
+			// save the current latency so we can restore it later
+			cpuPowerSaveDir := fmt.Sprintf("%s/cpu%d/power", cpuSaveDir, cpu)
+			err = os.MkdirAll(cpuPowerSaveDir, 0o750)
+			if err != nil {
+				return err
+			}
+			cpuLatencyFileOrig := fmt.Sprintf("%s/pm_qos_resume_latency_us", cpuPowerSaveDir)
```
kolyshkin, line 453: Same here, this can just be named `fileOrig` or e.g. `savedFile`

kolyshkin (changes requested):
Left some comments. In general, please
 - try to use shorter variable names if their context is small and well known; it is easier to read this way.
 - avoid unnecessary strings.Trim and string to []byte and back conversion, this will result in less work for garbage collector.

bartwensley:
> Left some comments. In general, please
> 
> * try to use shorter variable names if their context is small and well known; it is easier to read this way.
> * avoid unnecessary strings.Trim and string to []byte and back conversion, this will result in less work for garbage collector.

@kolyshkin - Thanks for the quick review!!! I will reply to all your comments once I have made the changes and retested.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: I based this on other files/dirs under /var/run/crio - they seem to all have permissions for the group.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Shortened the name.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: OK - I removed the conversions and the trimming here and in the other similar places.

internal/runtimehandlerhooks/high_performance_hooks.go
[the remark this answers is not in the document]
bartwensley: I addressed the first two items (using strings.Fields() and removing the intermediate varaible). However, although all CPUs currently have the same available governors, I'd rather not make that assumption in the code, since sysfs allows these to be different for each CPU.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: OK - removed the intermediate variable.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: I don't think we need to do that - we previously ensure the file exists when we read it above (see line 562). I think it is OK to treat the unexpected disappearance of the file as an error.

bartwensley:
@kolyshkin - I have made all the changes you requested (or replied to your comments). Please take another look when you have a chance. Thanks.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,171 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+				return err
+			}
+
+			// save the current latency so we can restore it later
+			cpuPowerSaveDir := fmt.Sprintf("%s/cpu%d/power", cpuSaveDir, cpu)
+			err = os.MkdirAll(cpuPowerSaveDir, 0o750)
+			if err != nil {
+				return err
+			}
+			latencyFileOrig := fmt.Sprintf("%s/pm_qos_resume_latency_us", cpuPowerSaveDir)
```
kolyshkin, line 452: nit: since this is not C, it's better to use simple strings concatenation:
```go
cpuPowerSaveDir + "/pm_qos_resume_latency_us"
```
or path.Join:
```go
path.Join(cpuPowerSaveDir, "pm_qos_resume_latency_us")
````

File: internal/runtimehandlerhooks/high_performance_hooks.go [hunk 9]
```diff
@@ -356,6 +417,171 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			if err != nil {
+				if errors.Is(err, os.ErrNotExist) {
+					// the latency may have already been restored by a previous invocation of the hook
+					return nil
+				}
+				return err
+			}
+
+			// restore the original latency
+			latencyFile := fmt.Sprintf("%s/cpu%d/power/pm_qos_resume_latency_us", cpuDir, cpu)
```
kolyshkin, line 476: Somewhere above the string `"enable"` used once, and it is defined as `const`.

Here `pm_qos_resume_latency_us` is used 4 times, and it is not defined as const.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,171 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			governorFound := false
+			for _, availableGovernor := range strings.Fields(string(availGovernors)) {
+				if availableGovernor == governor {
+					governorFound = true
+					break
+				}
+			}
+			if !governorFound {
+				return errors.Errorf("governor %s not available for cpu %d", governor, cpu)
+			}
```
kolyshkin, line 536: I think it is safe to assume that the list of available governors is the same for all CPUs, so it should only be checked once.

Yet better, you do not want to check it at all -- since writing of an invalid string to `scaling_governor` should result in a particular error (`unix.EINVAL`).

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,171 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+		}
+	}
+
+	return nil
+}
+
+// setCPUFreqGovernor sets the scaling_governor for a cpu and stores the original
+// value so it can be restored later. If the governor is an empty string, the original
+// scaling_governor value is restored.
+func setCPUFreqGovernor(c *oci.Container, governor, cpuDir, cpuSaveDir string) error {
```
kolyshkin, line 496: Since `cpuDir` and `cpuSaveDir` never change, it does not make sense to have them as parameters.

For the test case, you can maybe do `chroot`, or make `sysCPUDir` and `sysCPUSaveDir` variables that are changed by the test case.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +417,171 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
+	if err != nil {
+		return err
+	}
+
+	for _, cpu := range cpus.ToSlice() {
+		if latency != "" {
+			// retrieve the current latency
+			latencyFile := fmt.Sprintf("%s/cpu%d/power/pm_qos_resume_latency_us", cpuDir, cpu)
+			latencyOrig, err := ioutil.ReadFile(latencyFile)
```
kolyshkin, line 441: Here and below `os.ReadFile` and `os.WriteFile` should be used (since Go 1.16 or so).

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Done

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: In the case of pm_qos_resume_latency, since the usages are all in a single function are are mostly as part of a longer file path, I think that converting it to a constant is going to make the code less readable. I would prefer to leave this as is.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: So looking at the kernel code, it looks like there is only a single set of governors (stored in the cpufreq_governor_list global):
https://github.com/torvalds/linux/blob/268db333b561c77dee3feb6831806412293b4a7e/drivers/cpufreq/cpufreq.c#L2441

However, I think the way I've written this is more "correct" since the scaling_available_governors is exposed on a per cpu basis. My existing code is more future proof if the kernel implementation was ever to change. Is your concern related to the extra sysfs file access required for each cpu when a container is created (and has the new annotation)? If this is something where realtime is critical then I can make the change, but I just wanted to confirm we really need to do this.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: I based this code on the existing code in this file where the paths are passed in as parameter in order to facilitate unit testing. When I wrote the code, I also thought this was strange so I did some research to understand if there was a nice way for unit tests to mock out calls to os.ReadFile, os.WriteFile, etc... In short, there does not seem to be a reasonable way to do this. The best solution I found (other than passing in the paths which is what most people do) is to pass in a filesystem instead of the paths, which allows the unit test to pass in a "fake" filesystem. This would be a variation on https://talks.golang.org/2012/10things.slide#8

I don't see any testcases in crio (or elsewhere) that use chroot for this. I also don't want to make sysCPUDir and sysCPUSaveDir into global variables to facilitate unit testing - that also has potential issues (e.g. with parallel testcase execution).

So... given that I'm following an existing pattern from code in this file and there don't seem to be any existing examples of this type of mocking being done in the crio-o codebase (correct me if I'm wrong), I'd prefer to leave this as is.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -127,6 +161,24 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s
 
 	// no need to reverse the cgroup CPU CFS quota setting as the pod cgroup will be deleted anyway
 
+	// restore the c-state configuration for the container CPUs
+	configureCState, _ := shouldCStatesBeConfigured(s.Annotations())
+	if configureCState {
```
haircommander, line 166: can we note why we don't need to restore the state when it was disabled?

internal/runtimehandlerhooks/high_performance_hooks.go
haircommander: we can make a note above the functions that say this reasoning for keeping them parameters. If this is the way we're doing it now, we can leave it and possibly address in the future

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: I added a comment. We only restore the c-state when the annotation is present - without the annotation we don't modify the c-state when the container is started or stopped.

bartwensley:
@kolyshkin and @haircommander - when you have time it would be great if you could take another look and either mark your comments as resolved or add a note to the comments where you aren't satisfied with my reply. Thanks!!!

haircommander:
/approve

LGTM, thanks @bartwensley 

@kolyshkin any other comments?

bartwensley:
@kolyshkin @haircommander - I squashed my commits (as requested by Kir). There are no changes - just the squash. Please take another look.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -92,6 +96,36 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 		}
 	}
 
+	// configure c-states for the container CPUs
+	configureCState, value := shouldCStatesBeConfigured(s.Annotations())
+	if configureCState {
```
kolyshkin, line 101: nit: perhaps these two can be combined (added benefit is limiting the variables scope:

```go
if ok, value := shouldCStatesBeConfigured(s.Annotations()); ok {
...
}
```

Note that I call the variable `ok` as it is only used in this one line, and it's clear it is related to `shouldCStatesBeConfigured`. In general, the shorter the scope, the shorter the variable name should be.

internal/runtimehandlerhooks/high_performance_hooks.go
kolyshkin: One other way to solve this is to have something like
```
func setCPUFreqGovernor(c *oci.Container, governor, cpuDir, cpuSaveDir string) error {
   return doSetCPUFreqGovernor(c, governor, sysCPUDir, sysCPUSaveDir)
}

func doSetCPUFreqGovernor(c *oci.Container, governor, cpuDir, cpuSaveDir string) error {
....
}
```

then use the first function in the code, and the second function in the tests.

Overall it's not a big deal, so it's fine the way it is, but please consider refactoring the code later to simplify it.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +420,175 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+		return errors.Errorf("find container %s CPUs", c.ID())
+	}
+
+	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
+	if err != nil {
+		return err
+	}
+
+	for _, cpu := range cpus.ToSlice() {
+		if governor != "" {
```
kolyshkin, line 518: It looks like `governorFile` and `governorFileOrig` are used in both branches (`if` and `else`), so it makes sense to move them out of the block. IOW:

```go
	for _, cpu := range cpus.ToSlice() {
		governorFile := fmt.Sprintf("%s/cpu%d/cpufreq/scaling_governor", cpuDir, cpu)
		cpuFreqSaveDir := fmt.Sprintf("%s/cpu%d/cpufreq", cpuSaveDir, cpu)
		governorFileOrig := path.Join(cpuFreqSaveDir, "scaling_governor")
		if governor != "" {
			....
```

This way we are sure we're using the same files on set and restore.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +420,175 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			governorFound := false
+			for _, availableGovernor := range strings.Fields(string(availGovernors)) {
+				if availableGovernor == governor {
+					governorFound = true
+					break
+				}
+			}
+			if !governorFound {
+				return errors.Errorf("governor %s not available for cpu %d", governor, cpu)
+			}
```
kolyshkin, line 543: I would move the logic of checking if the governor is supported to a separate function, for clarity, so the code above is reduced to 
```go
if err := isCPUGovernorSupported(governor, cpuDir, cpu); err != nil {
   return err
}
```

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +420,175 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+		return errors.Errorf("find container %s CPUs", c.ID())
+	}
+
+	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
+	if err != nil {
+		return err
+	}
+
+	for _, cpu := range cpus.ToSlice() {
+		if latency != "" {
```
kolyshkin, line 443: Same as https://github.com/cri-o/cri-o/pull/5927/commits/f2859a0268d9d1d02ae6ec89c98dca72a48ab603#r906250537

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -356,6 +420,175 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			latencyFileOrig := path.Join(cpuPowerSaveDir, "pm_qos_resume_latency_us")
+			err = os.WriteFile(latencyFileOrig, latencyOrig, 0o644)
+			if err != nil {
+				return err
+			}
+
+			// update the pm_qos_resume_latency_us
+			err = os.WriteFile(latencyFile, []byte(latency), 0o644)
+			if err != nil {
+				return err
```
kolyshkin, line 466: Question: if we were unable to set the new latency, should we not be saving the original one (as there's no need to restore it)?

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: I like that - it keeps the calling functions simple (no need to pass the paths) but allows for clean and simple unit tests. I made the change.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: OK - I made the change - but I named the variable "configure" instead of "ok" as I personally read "ok" as a pass/fail thing where this return as more meaning.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Done - good suggestion.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Done

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Done

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: It should be an extremely rare case where we couldn't set the new latency. The only improvement here would be to delete the saved latency if we couldn't set the new latency - I still want to save the original latency first, as I wouldn't want to change the latency if I knew I wasn't going to be able to restore it to the original value. Since this failure case is extremely unlikely and there is no adverse impact of attempting to restore the original (unchanged) latency, I'd prefer to keep the code simple and not add in the extra error handling.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -355,6 +415,186 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
 	return nil
 }
 
+// setCPUPMQOSResumeLatency sets the pm_qos_resume_latency_us for a cpu and stores the original
+// value so it can be restored later. If the latency is an empty string, the original latency
+// value is restored.
+// Note: Passing the directories as parameters to facilitate unit tests (following the
+// pattern used by other functions in this file).
```
kolyshkin, line 422: This comment is now obsoleted.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -355,6 +415,186 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+func setCPUPMQOSResumeLatency(c *oci.Container, latency string) error {
+	return doSetCPUPMQOSResumeLatency(c, latency, sysCPUDir, sysCPUSaveDir)
+}
+
+// doSetCPUPMQOSResumeLatency facilitates unit testing by allowing the directories to be specified as parameters
```
kolyshkin, line 427: nit: missing period.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -355,6 +415,186 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// isCPUGovernorSupported checks whether the cpu governor is supported for the specified cpu
```
kolyshkin, line 497: nit: missing period.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -156,6 +206,16 @@ func shouldIRQLoadBalancingBeDisabled(annotations fields.Set) bool {
 		annotations[crioannotations.IRQLoadBalancingAnnotation] == annotationDisable
 }
 
+func shouldCStatesBeConfigured(annotations fields.Set) (present bool, value string) {
+	value, present = annotations[crioannotations.CPUCStatesAnnotation]
```
kolyshkin, line 210: I don't really like how the two values are swapped; perhaps it's better to return it in the same order, so the user would do

```go
if value, configure := shouldCStatesBeConfigured(s.Annotations()); configure {
   ....
```

With that change implemented, maybe we don't need this function at all, and just have
```go
if value, ok := s.Annotations()[crioannotations.CPUCStatesAnnotation]; ok {
   ...
```

but if you think that having this as a separate function improves readability or enables better code reuse, please feel free to leave it.

kolyshkin (approved):
Left a few more comments, mostly nits except for the obsoleted comment.

Also, can you check that all your code comments start with an Uppercase letter and end with a period? This, while being a nitpick, helps to maintain the code in the long run.

kolyshkin (changes requested):
see above

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Removed

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Added period.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: I did this on purpose to be consistent with the other functions that were already here (e.g. shouldCPUQuotaBeDisabled). The existing functions are returning a similar bool as their first (and only parameter) so I felt returning the bool as the first parameter was more consistent. That makes sense (at least in my mind) so I'd prefer to leave these as they are.

bartwensley:
Hmm... looks like my changes no longer build due 0a3bfc07817ef14f3feec2471d18c49b2727602c - I will rebase.

kolyshkin (approved):
LGTM, thank you!

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -354,6 +415,184 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			if err != nil {
+				return err
+			}
+
+			// Update the pm_qos_resume_latency_us.
+			err = os.WriteFile(latencyFile, []byte(latency), 0o644)
+			if err != nil {
+				return err
+			}
+		} else {
```
haircommander, line 467: nit: can we continue here instead of nesting the rest with `else {`

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -354,6 +415,184 @@ func setCPUQuota(cpuMountPoint, parentDir string, c *oci.Container, enable bool)
...
+			if err != nil {
+				return err
+			}
+
+			// Update the governor.
+			err = os.WriteFile(governorFile, []byte(governor), 0o644)
+			if err != nil {
+				return err
+			}
+		} else {
```
haircommander, line 568: same here

haircommander:
one more set of nits, thanks for your patience @bartwensley !

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Judging by your comment below - I assume this isn't really a "nit" because you are asking me to change this before it gets merged. I actually think it's more readable the way it is but that could be because I'm new to golang. I guess the golang preference is to avoid nesting where possible and return (or continue) early? I can update this, but won't be until next week as tomorrow is a holiday here in Canada.

haircommander: yeah in general I prefer flows with the least indentation. Thank you, no rush on my end :+1:

TomSweeneyRedHat:
No changes needed for docs?

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Done

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Done

bartwensley:
@TomSweeneyRedHat - I checked and I don't see the existing "*.crio.io" annotations mentioned in the cri-o documentation so I don't think a doc update is required.

haircommander:
@saschagrunert PTAL

haircommander:
/approve

LGTM, thanks @bartwensley !

saschagrunert (approved):
/lgtm

PR #5933, opened 2022-06-06
oci: add support for capping memory and disk usage from exec sync output

haircommander:
#### What type of PR is this?

/kind bug

```release-note
Fix a bug where ExecSync requests (exec probes) could use an arbitrary amount of memory and disk. Output from ExecSync requests is now limited to 16MB (the amount that exec output was limited to in the dockershim). Disk limiting requires conmon 2.1.2 to work. See https://github.com/cri-o/cri-o/security/advisories/GHSA-fcm2-6c3h-pg6j and CVE-2022-1708 for more information.
```


haircommander:
merged with the ghsa

PR #5956, opened 2022-06-09
Update containerd vendor

littlejawa:
#### What type of PR is this?

/kind dependency-change

Update containerd golang dependency

None

This PR completes the work done on #5917 

None


littlejawa:
@saschagrunert can you have a look?

littlejawa:
> @littlejawa, the code itself seems reasonable, but I really would like to have as much info as possible in the commit message describing why we need this change.
> [...]
> WDYT?

Sure, I will add that :-)

saschagrunert (approved):
/lgtm
/release-note-none

fidencio:
@littlejawa, could you add the following comment to the commit that's actually doing that?
```
Had to modify the way we call the Command() function from containerd:
- use of the CommandConfig structure as parameter
- use the actual binary name
  (no more converting to containerd binary naming pattern)

This later change was needed as in Containerd 1.6, the validation of the given
binary path is not done in Command() anymore - the path is used as is, allowing
custom runtime paths.
(see https://github.com/containerd/containerd/commit/6870f3b1)
```

littlejawa:
> @littlejawa, could you add the following comment to the commit that's actually doing that?

I did (I think?). The commit with this comment is updating containerd AND modifying runtim_vm.go to make it compatible with the new API. That what is described here.

The other commit is cleanup, that remove the function that is no longer called (because of the change in the first commit).

Am I missing something? Or should I rephrase it somehow?

fidencio:
> > could you add the following comment to the commit that's actually doing that?
> 
> @fidencio - I did (I think?). The commit with this comment is updating containerd AND modifying runtim_vm.go to make it compatible with the new API. That what is described here.
> 
> The other commit is cleanup, that remove the function that is no longer called (because of the change in the first commit).
> 
> Am I missing something? Or should I rephrase it somehow?

Oh, no, we're fine.  I missed one bit here on my side. :-)

fidencio:
LGTM!

PR #5970, opened 2022-06-13
systemd: use on-failure as restart policy

haircommander:
#### What type of PR is this?

/kind bug

```release-note
none
```


TomSweeneyRedHat:
LGTM
once the last unhappy test is fixed up.

haircommander:
I think we should not go this route actually

PR #5971, opened 2022-06-13
Add test for "Canonize selinux label"

holzman:
Signed-off-by: Burt Holzman <burt@fnal.gov>

#### What type of PR is this?

/kind other

Adds test case for PR #5775 

None

```release-note
None
```


holzman:
Hang on, this test isn't testing what I want to test  - will resubmit (with the right shellcheck disable line too) in a sec.

PR #5974, opened 2022-06-15
Enter mount namespace if set in $KUBENSMNT environment

lack:
#### What type of PR is this?

/kind feature

This is in support of OpenShift Enhancement https://github.com/openshift/enhancements/pull/1127

CRI-O will now check whether `$KUBENSMNT` is set in the environment.  If this
environment variable is defined, assume the file it points at is a bindmount to
a mount namespace that CRI-O should join:
- If the environment variable is not set, take no action
- If the environment variable is set and points at a valid mount
  namespace, join that mount namespace.
- If the environment variable is set but points at a missing file, or
  the file is not a valid mount namespace, print a warning but continue
  to run in the original mount namespace.

The approach makes use of a cgo init block to ensure the namespace is
joined before any go threads start up.

None

Ready for final review. Includes CI and docs changes.

```release-note
If `$KUBENSMNT` is defined in the environment, assume the file it points at is
a bindmount to a mount namespace that CRI-O should join:
- If the environment variable is not set, take no action.
- If the environment variable is set and points at a valid mount
  namespace, CRI-O will join that mount namespace.
- If the environment variable is set but points at a missing file, or the file
  is not a valid mount namespace, CRI-O will print a warning but continue to
  run in the original mount namespace.
```

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containers/kubensmnt"
 	_ "github.com/containers/podman/v4/pkg/hooks/0.1.0"
 	"github.com/containers/storage/pkg/reexec"
 	"github.com/cri-o/cri-o/internal/criocli"
@@ -106,6 +107,12 @@ scope of the CRI.
 5. Monitoring and logging required to satisfy the CRI.
 6. Resource isolation as required by the CRI.`
 
+const kubensmntHelp = `Path to a bind-mounted mount namespace that CRI-O
+should join before launching any containers. If the path does not exist,
+or does not point to a mount namespace bindmount, CRI-O will run in its
+parent's mount namespace and log a warning that the requested namespace
+was not joined.`
+
 func main() {
 	log.InitKlogShim()
 
@@ -135,6 +142,10 @@ func main() {
 
 	sort.Sort(cli.FlagsByName(app.Flags))
 	sort.Sort(cli.FlagsByName(configCommand.Flags))
+
+	app.Metadata["Env"] = map[string]string{
+		kubensmnt.EnvName: kubensmntHelp,
+	}
 
 	app.Commands = criocli.DefaultCommands
 	app.Commands = append(app.Commands, []*cli.Command{
@@ -224,6 +235,16 @@ func main() {
 		if c.Args().Len() > 0 {
 			cancel()
 			return fmt.Errorf("command %q not supported", c.Args().Get(0))
+		}
+
+		// Check if we joined a mount namespace
+		nsname, err := kubensmnt.Status()
+		if nsname != "" {
+			if err != nil {
+				logrus.Warn(err)
+			} else {
+				logrus.Infof("Joined mount namespace %q", nsname)
+			}
 		}
 
 		config, ok := c.App.Metadata["config"].(*libconfig.Config)
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -448,6 +448,18 @@ Shows a list of commands or help for one command
   Storage configuration file specifies all of the available container storage
   options for tools using shared container storage.
 
+# ENVIRONMENT
+
+All command-line options may also be specified as environment variables.
+The options detailed in this section, however, can only be set via
+environment variables.
+
+**KUBENSMNT**: Path to a bind-mounted mount namespace that CRI-O
+should join before launching any containers. If the path does not exist,
+or does not point to a mount namespace bindmount, CRI-O will run in its
+parent's mount namespace and log a warning that the requested namespace
+was not joined.
+
 # SEE ALSO
 
 crio.conf(5), crio.conf.d(5), oci-hooks(5), policy.json(5), registries.conf(5),
--- a/test/namespaces.bats
+++ b/test/namespaces.bats
@@ -80,3 +80,64 @@ function teardown() {
 	# make sure namespace is cleaned up
 	[[ -z $(ls "$CONTAINER_NAMESPACES_DIR/pidns") ]]
 }
+
+@test "KUBENSMNT mount namespace" {
+	original_ns=$(readlink /proc/self/ns/mnt)
+
+	PIN_ROOT=$TESTDIR/kubens
+	mkdir -p "$PIN_ROOT"
+
+	# Set up a pinned mount namespace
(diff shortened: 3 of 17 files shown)


cmd/crio/kubensenter/kubensenter.go
```diff
@@ -0,0 +1,33 @@
...
+	errorMessage[0] = '\0';
+	nsenterConfig = getKubeNsMnt();
+	nsenterResult = joinMountNamespace(nsenterConfig, errorMessage, ERR_LIMIT);
+}
+*/
+import "C"
+
+import "errors"
+
+func Status() (string, error) {
```
haircommander, line 27: the name Status doesn't imply that it is also joining the namespace. Maybe "CheckAndEnter" or something?

docs/crio.8.md
```diff
@@ -448,6 +448,18 @@ Shows a list of commands or help for one command
...
   options for tools using shared container storage.
 
+# ENVIRONMENT
+
+All command-line options may also be specified as environment variables.
+The options detailed in this section however can only be set via
+environment variables.
+
+**KUBENSMNT**: Path to a bind-mounted mount namespace that CRI-O
+should joinlbefore launching any containers. If the path does not exist,
```
TomSweeneyRedHat, line 458:
```suggestion
should join before launching any containers. If the path does not exist,
```

lack:
@TomSweeneyRedHat : Thanks for the review!  All typos now retyped.

haircommander:
can you think of any consequences of crio starting outside of the kubensmnt, and then being restarted within the kubensmnt? or vice versa? in `test/reload.bats` we have a couple of tests checking that pods restore correctly when changing different config options. we probably should add some tests for this case as well

lack:
> can you think of any consequences of crio starting outside of the kubensmnt, and then being restarted within the kubensmnt? or vice versa? in `test/reload.bats` we have a couple of tests checking that pods restore correctly when changing different config options. we probably should add some tests for this case as well

@haircommander : An interesting question!
- If crio starts outside, starts some pods, shuts down, and then starts up inside, what should happen?  If we assume the new namespace is a slave to the original namespace, crio should see all the containers and all their mount points, and everything should work fine.
- If crio starts inside, starts some pods, shuts down, and then starts up outside, what should happen?  I think from the perspective of crio managing pods, everything *might* be okay, though crio would not have access to the mountpoints created when the pods were first created (does it need them?  maybe not?).  The containers themselves might become confused depending on the use-case, especially if you're doing some kind of CSI and later-started pods outside the namespace need to access something shareable mounted by a pod inside the namespace...

I'll add some tests to this effect and see what happens!

haircommander:
/approve

thanks @lack! @saschagrunert @kolyshkin PTAL

haircommander:
> [APPROVALNOTIFIER] This PR is **APPROVED**
> 
> This pull-request has been approved by: _[haircommander](https://github.com/cri-o/cri-o/pull/5974#issuecomment-1175128647)_, _[lack](https://github.com/cri-o/cri-o/pull/5974#)_
> 
> The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=cri-o%2Fcri-o).
> 
> The pull request process is described [here](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process)
> Needs approval from an approver in each of these files:

I would expect the coverage to drop here, as it only counts the unit test coverage, not integration test coverage. It can be ignored

saschagrunert (approved):
LGTM

lack:
The CGO+init seems to be the best (and only) way to have Go code enter a mount namespace :)

cmd/crio/main.go
```diff
@@ -226,6 +237,16 @@ func main() {
 			return fmt.Errorf("command %q not supported", c.Args().Get(0))
 		}
 
+		// Check if we joined a mount namespace
+		nsname, err := kubensmnt.Status()
+		if nsname != "" {
+			if err != nil {
+				logrus.Warnf("%v", err)
```
kolyshkin, line 244: nit: `logrus.Warn(err)`

test/namespaces.bats
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bats
+# vim:set ft=bash :
```
kolyshkin, line 2: nit: not related to this PR.

kolyshkin (approved):
SGTM aside from a couple of nits. Will review the bats tests later.

test/restore.bats
```diff
@@ -324,3 +324,92 @@ function teardown() {
...
+@test "crio restore upon entering KUBENSMNT" {
+	start_crio
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	pod_list_info=$(crictl pods --quiet --id "$pod_id")
+
+	output=$(crictl inspectp -o json "$pod_id")
+	[[ "$output" != "" ]]
+	pod_status_info=$(echo "$output" | jq ".status.state")
+	pod_ip=$(echo "$output" | jq ".status.ip")
+	pod_created_at=$(echo "$output" | jq ".status.createdAt")
```
kolyshkin, line 337: would be nice to use here-file for this, but since other places doesn't use it, never mind.

test/restore.bats
```diff
@@ -324,3 +324,92 @@ function teardown() {
...
+}
+
+@test "crio restore upon exiting KUBENSMNT" {
+	setup_kubensmnt
+	start_crio
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	pod_list_info=$(crictl pods --quiet --id "$pod_id")
+
+	output=$(crictl inspectp -o json "$pod_id")
+	[[ "$output" != "" ]]
```
kolyshkin, line 379: nit: `[[ -n "$output" ]]`

vendor/github.com/containers/kubensmnt/kubensmnt_linux.go
```diff
@@ -0,0 +1,50 @@
+//go:build linux && cgo
+// +build linux,cgo
+
```
kolyshkin, line 3: Not related to this PR but these build tags are redundant.

lack: My editor adds both of them automagically for backwards-compatibility reasons.

test/restore.bats
lack: I'm going to go make changes anyway, so why not :)

PR #5978, opened 2022-06-17
Update conmon

sohankunkerkar:
#### What type of PR is this?

/kind ci

Update conmon

None

```
E0615 16:42:37.918430  296222 remote_runtime.go:711] "ExecSync cmd from runtime service failed" err=<
	rpc error: code = Unknown desc = command error: EOF, stdout: conmon: option parsing failed: Unknown option --log-global-size-max
	, stderr: , exit code -1
```
We need to update the version of conmon here to address the current crio-test CI failure. We're already installing that dependency as part of [this](https://github.com/cri-o/cri-o/blob/main/contrib/test/integration/build/cri-o.yml#L13) step. So there's no need to install that again using ansible.

No


saschagrunert (approved):
/lgtm

PR #6004, opened 2022-06-27
server: userns support

giuseppe:
Honor the CRI user namespace configuration when it is present in the
sandbox configuration.

The CRI-O custom support for user namespaces that is done through
annotations is ignored in this case, and the configuration coming from
the Kubelet has precedence over the annotation.

#### What type of PR is this?

/kind feature

add support for user namespaces

KEP 127 defines the new feature

```release-note
Now CRI-O honors the user namespace configuration that the Kubelet specified
```


server/sandbox_run_linux.go
```diff
@@ -62,7 +62,33 @@ func addToMappingsIfMissing(ids []idtools.IDMap, id int64) []idtools.IDMap {
 	return append(ids, newMapping)
 }
 
+func convertToStorageIDMap(mappings []*types.IDMapping) []idtools.IDMap {
```
haircommander, line 65: nit: can we define this function underneath where it's called?

server/container_create_linux.go
```diff
@@ -126,6 +126,10 @@ func (s *Server) finalizeUserMapping(specgen *generate.Generator, mappings *idto
 		return
 	}
 
+	if sb.Annotations()[crioann.UsernsModeAnnotation] == "" {
+		return
+	}
+
```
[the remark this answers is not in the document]
giuseppe: could we track (and implement) the deprecation separately?  At the moment we have more features available with the annotations, and I am not sure they will be supported in Kubernetes

server/container_create_linux.go
```diff
@@ -126,6 +126,10 @@ func (s *Server) finalizeUserMapping(specgen *generate.Generator, mappings *idto
 		return
 	}
 
+	if sb.Annotations()[crioann.UsernsModeAnnotation] == "" {
```
saschagrunert, line 129: Should we check for a unset field here?

```suggestion
	if _, ok := sb.Annotations()[crioann.UsernsModeAnnotation]; !ok {
```

go.mod
```diff
@@ -418,7 +418,7 @@ replace (
 	k8s.io/component-base => k8s.io/kubernetes/staging/src/k8s.io/component-base v0.0.0-20220524121711-3ddd0f45aa91
 	k8s.io/component-helpers => k8s.io/kubernetes/staging/src/k8s.io/component-helpers v0.0.0-20220524121711-3ddd0f45aa91
 	k8s.io/controller-manager => k8s.io/kubernetes/staging/src/k8s.io/controller-manager v0.0.0-20220524121711-3ddd0f45aa91
-	k8s.io/cri-api => k8s.io/kubernetes/staging/src/k8s.io/cri-api v0.0.0-20210928072042-90051602459c
+	k8s.io/cri-api => k8s.io/kubernetes/staging/src/k8s.io/cri-api v0.0.0-20220626121858-f4abde9e57cc
```
saschagrunert, line 421: We create 1.25.0-alpha.2 in k/k today, let's update every replaced component to this tag/commit to avoid vendoring incompatible packages.

giuseppe: thanks.  I've pushed a new version now

saschagrunert (approved):
/lgtm

PR #4072, opened 2020-08-12
Add /sys/dev as a masked path

umohnani8:
Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
If the container is not privileged, mask the /sys/dev
path so that we don't leak extra devices in a container.

#### Which issue(s) this PR fixes:
Fixes #1772993 (https://bugzilla.redhat.com/show_bug.cgi?id=1772993)

#### Does this PR introduce a user-facing change?

```release-note
None

```

PR #6025, opened 2022-07-01
server: Fix Unmasked ProcMountType

### LINKED ISSUE #6009: Unmasked ProcMountType is ignored
### What happened?

After enabling the ProcMountType feature gate setting `procMount: Unmasked` in a pod spec results in still seeing paths like `/proc/kcore` mounted.

`mount | grep /proc` within the container only show one mount.

Add `--feature-gates=ProcMountType=true` to Kubernetes components then within a pod's container configure:

```yaml
securityContext:
  procMount: Unmasked
```

I can fix this by deleting the blocks of code in https://github.com/cri-o/cri-o/blob/main/server/container_create_linux.go#L443-L456 that call `AddLinuxMaskedPaths` and `AddLinuxReadonlyPaths`.

I believe this is safe because Kubernetes populates the masked paths itself: https://github.com/kubernetes/kubernetes/blob/master/pkg/securitycontext/util.go#L217 -- but filing as an issue rather than a PR as I may have missed something, so would be good to get some eyes on this (happy to make a PR if you prefer).

There is one exception -- cri-o has added `/sys/dev`, which is a bit strange as that tree symlinks to `/sys/devices` which isn't masked, so I'm not sure what the intention of that is.

This is particularly useful for me, as combined with user namespace support in cri-o it's possible (and safe) to use an unmasked /proc, which allows running nested containers more easily (any container can be made to work, rather than having to get it to mount /proc in a particular way).

<details>

```console
$ crio --version
crio version 1.24.0
Version:        1.24.0
GitCommit:      8b001fb3cf7637e491fd31a1af731d4656413b44
GitCommitDate:  2022-06-25T00:45:48Z
GitTreeState:   dirty
BuildDate:      2022-06-27T03:31:09Z
GoVersion:      go1.18.3
Compiler:       gc
Platform:       linux/amd64
Linkmode:       dynamic
BuildTags:
  containers_image_ostree_stub
  exclude_graphdriver_btrfs
  btrfs_noversion
  libdm_no_deferred_remove
  containers_image_openpgp
  seccomp
LDFlags:          -s -w -X github.com/cri-o/cri-o/internal/pkg/criocli.DefaultsPath="" -X github.com/cri-o/cri-o/internal/version.buildDate=2022-06-27T03:31:09Z
SeccompEnabled:   true
AppArmorEnabled:  false
Dependencies:
```

```console
$ kubectl version
kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.1", GitCommit:"3ddd0f45aa91e2f30c70734b175631bec5b5825a", GitTreeState:"clean", BuildDate:"2022-05-24T12:26:19Z", GoVersion:"go1.18.2", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.1", GitCommit:"3ddd0f45aa91e2f30c70734b175631bec5b5825a", GitTreeState:"clean", BuildDate:"2022-05-24T12:18:48Z", GoVersion:"go1.18.2", Compiler:"gc", Platform:"linux/amd64"}
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

<details>

</details>

dgl:
#### What type of PR is this?

/kind bug

Set masked and readonly paths based on the securityContext only.
Kubernetes as of v1.13 will always populate these based on
ProcMountType.

This can be used to allow running Docker within a container, with fewer contortions (particularly when combined with user namespaces this can be safer, as the container isn't root on the host, but root within a user namespace). Obviously there's other ways to achieve running containers within containers without this, but given this is an option the Kubernetes API supports (even if at alpha stage) it seems worthwhile supporting it.

Fixes #6009.

The list in CRI-O has /sys/dev in it from #4072, note however the code as it was before this change overwrites it with the list supplied from Kubernetes, so I am fairly sure this is not used. (I don't see /sys/dev masked when running a container with CRI-O on Kubernetes 1.24).

The list in Kubernetes for comparison is: https://github.com/kubernetes/kubernetes/blob/master/pkg/securitycontext/util.go#L214-L235

```release-note
If the ProcMountType feature gate is enabled, CRI-O now honors the ProcMount securityContext set.
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -440,39 +440,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		specgen.SetProcessNoNewPrivileges(securityContext.NoNewPrivs)
 
 		if !ctr.Privileged() {
-			for _, mp := range []string{
-				"/proc/acpi",
-				"/proc/kcore",
-				"/proc/keys",
-				"/proc/latency_stats",
-				"/proc/timer_list",
-				"/proc/timer_stats",
-				"/proc/sched_debug",
-				"/proc/scsi",
-				"/sys/firmware",
-				"/sys/dev",
-			} {
-				specgen.AddLinuxMaskedPaths(mp)
-			}
 			if securityContext.MaskedPaths != nil {
-				specgen.Config.Linux.MaskedPaths = nil
 				for _, path := range securityContext.MaskedPaths {
 					specgen.AddLinuxMaskedPaths(path)
 				}
 			}
 
-			for _, rp := range []string{
-				"/proc/asound",
-				"/proc/bus",
-				"/proc/fs",
-				"/proc/irq",
-				"/proc/sys",
-				"/proc/sysrq-trigger",
-			} {
-				specgen.AddLinuxReadonlyPaths(rp)
-			}
 			if securityContext.ReadonlyPaths != nil {
-				specgen.Config.Linux.ReadonlyPaths = nil
 				for _, path := range securityContext.ReadonlyPaths {
 					specgen.AddLinuxReadonlyPaths(path)
 				}


server/container_create_linux.go
```diff
@@ -440,39 +440,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		specgen.SetProcessNoNewPrivileges(securityContext.NoNewPrivs)
 
 		if !ctr.Privileged() {
-			for _, mp := range []string{
```
haircommander, line 443: I am not seeing any toggling of this behavior based on the security context

dgl: There isn't really a way to do that, either Kubernetes provides the masked paths, or it doesn't.

In containerd they do have an option to disable this support, which they suggest enabling on Kubernetes <=1.11: https://github.com/containerd/containerd/blob/2b661b890f689716b2928408e386e656d8dd7118/pkg/cri/config/config.go#L283-L285

(The actual implementation in containerd is similar to this: https://github.com/containerd/containerd/blob/main/pkg/cri/server/container_create_linux.go#L202-L212)

I could add a similar option, but I don't believe cri-o aims to support Kubernetes versions that old?

haircommander: ah I see so kubelet unconditionally passes these down, unless the feature gate is on and the security context knob is set.

haircommander:
/ok-to-test
/approve

LGTM, thanks @dgl

saschagrunert (approved):
LGTM

PR #6023, opened 2022-06-30
server: update container status if state has bogus information

haircommander:
#### What type of PR is this?

/kind bug 
#### What this PR does / why we need it:
if CreatedAt is zero, the Kubelet will complain and also be confused by the container.
Attempt to update in this case.

Signed-off-by: Peter Hunt <pehunt@redhat.com>
#### Does this PR introduce a user-facing change?

```release-note
fix a bug where the status of a container ends up being bogus, causing Kubelet to print `verify ContainerStatus failed" err="status.CreatedAt is not set` repeatedly.
```

PR #5672, opened 2022-02-22
server: reduce object creation in List{Containers,PodSandboxes}

haircommander:
by only creating a new sandbox or container object when they change

Signed-off-by: Peter Hunt <pehunt@redhat.com>

#### What type of PR is this?

/kind bug
#### Does this PR introduce a user-facing change?

```release-note
none
```

PR #6026, opened 2022-07-01
oci/server: store container status in the container object

haircommander:
#### What type of PR is this?

/kind cleanup
this is part follow up of https://github.com/cri-o/cri-o/pull/5672, part alternative to https://github.com/cri-o/cri-o/pull/6023.

This came from investigation of #6023. Basically, it seems createdAt is subject to change because we return what is directly in the state. This got me thinking that we shouldn't be updating createdAt, finishedAt or startedAt, as they should be static.

That, paired with the effort in https://github.com/cri-o/cri-o/pull/5672 to reduce the number of new objects that are created, lead to this!

The idea is we know exactly when states change for containers. In UpdateContainerSTatus calls after we catch an exit, we know the time of exit, we know when we created the container and when we started it. Consulting `runc update` and updating our state can leave us in weird situations (like that in https://github.com/kubernetes/kubernetes/issues/105332 and https://bugzilla.redhat.com/show_bug.cgi?id=2094865).

Overall, this should make the flow simpler, and the number of objects created lower, at the cost of a decently heavy refactor.

```release-note
fix a bug where the status of a container ends up being bogus, causing Kubelet to print `verify ContainerStatus failed" err="status.CreatedAt is not set` repeatedly.
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -95,7 +95,7 @@ type Container interface {
 	SpecAddMount(rspec.Mount)
 
 	// SpecAddAnnotations adds annotations to the spec.
-	SpecAddAnnotations(ctx context.Context, sandbox *sandbox.Sandbox, containerVolume []oci.ContainerVolume, mountPoint, configStopSignal string, imageResult *storage.ImageResult, isSystemd, systemdHasCollectMode bool) error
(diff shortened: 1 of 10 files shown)


sohankunkerkar:
@haircommander what's the status of this PR? It's good to have these changes.

haircommander:
> @haircommander what's the status of this PR? It's good to have these changes.

yeah they seem like good changes though the PR needs some work. I don't think it's very high priority right now as the underlying issue was resolved

PR #6030, opened 2022-07-06
Retry dependency report push if failed

saschagrunert:
#### What type of PR is this?

/kind ci

We now retry pushing the dependency report in the same way we do it with the release notes since they both push to the same branch.

None
None

```release-note
None
```


saschagrunert:
@haircommander PTAL

PR #6056, opened 2022-07-14
Allow runtimes to be added and default runtime to be modified on config reload

### LINKED ISSUE #6036: Support reload of cri-o config to add runtimes
Is it possible to trigger a restart of crio to apply new runtime configurations without requiring a node reboot? e.g. through `SIGHUP` or by watching for changes at `/etc/crio/crio.d`.

## Motivation

When configuring additional runtimes (e.g. the NVIDIA Container Runtime or Kata), we want to make modifications to the `crio.conf` file (or a file in `/etc/crio/crio.d`). For example adding:
```
$ cat crio.conf

[crio]

  [crio.runtime]
    default_runtime = "nvidia"

    [crio.runtime.runtimes]

      [crio.runtime.runtimes.nvidia]
        runtime_path = "/usr/bin/nvidia-container-runtime"
        runtime_type = "oci"
```
(we don't always set the default runtime)

When using the GPU Operator to deploy the components required to enable access to GPUs in containers, this configuration update is performed by an operand managed by the operator. In the case of `docker` and `containerd` the equivalent configuration modifications are made and the docker or containerd daemons are restarted (e.g. using SIGHUP) to ensure that the newly configured runtime is available to run containers -- either using runtime classes, or as the default runtime.

These configuration options are also cleaned up when the operand responsible for applying the configuration is shut down, triggering another reload of the configuration.

elezar:
#### What type of PR is this?

/kind feature

This PR adds support for reloading runtime configs on SIGHUP. This allows for runtimes to be added dynamically and the default runtimes to be adjusted.

This is useful for cases where runtimes are added / configured after crio is started. One example of this is the NVIDIA Container Runtime which is managed by the NVIDIA GPU Operator. Without this change, adding a runtime to the config requires a node restart which is disruptive.

Fixes #6036

```release-note
Allow for a reload to add additional runtimes and change the default runtime
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -965,21 +965,8 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		return err
 	}
 
-	// check we do have at least a runtime
-	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
-		// Set the default runtime to "runc" if default_runtime is not set
-		if c.DefaultRuntime == "" {
-			logrus.Debugf("Defaulting to %q as the runtime since default_runtime is not set", defaultRuntime)
-			// The default config sets runc and its path in the runtimes map, so check for that
-			// first. If it does not exist then we add runc + its path to the runtimes map.
-			if _, ok := c.Runtimes[defaultRuntime]; !ok {
-				c.Runtimes[defaultRuntime] = defaultRuntimeHandler()
-			}
-			// Set the DefaultRuntime to runc so we don't fail further along in the code
-			c.DefaultRuntime = defaultRuntime
-		} else {
-			return fmt.Errorf("default_runtime set to %q, but no runtime entry table [crio.runtime.runtimes.%s] was found", c.DefaultRuntime, c.DefaultRuntime)
-		}
+	if err := c.ValidateDefaultRuntime(); err != nil {
+		return err
 	}
 
 	if c.LogSizeMax >= 0 && c.LogSizeMax < OCIBufSize {
@@ -1107,6 +1094,27 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		}
 	}
 
+	return nil
+}
+
+// ValidateDefaultRuntime ensures that the default runtime is set and valid.
+func (c *RuntimeConfig) ValidateDefaultRuntime() error {
+	// check we do have at least a runtime
+	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
+		// Set the default runtime to "runc" if default_runtime is not set
+		if c.DefaultRuntime == "" {
+			logrus.Debugf("Defaulting to %q as the runtime since default_runtime is not set", defaultRuntime)
+			// The default config sets runc and its path in the runtimes map, so check for that
+			// first. If it does not exist then we add runc + its path to the runtimes map.
+			if _, ok := c.Runtimes[defaultRuntime]; !ok {
+				c.Runtimes[defaultRuntime] = defaultRuntimeHandler()
+			}
+			// Set the DefaultRuntime to runc so we don't fail further along in the code
+			c.DefaultRuntime = defaultRuntime
+		} else {
+			return fmt.Errorf("default_runtime set to %q, but no runtime entry table [crio.runtime.runtimes.%s] was found", c.DefaultRuntime, c.DefaultRuntime)
+		}
+	}
 	return nil
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1099,22 +1099,26 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 
 // ValidateDefaultRuntime ensures that the default runtime is set and valid.
 func (c *RuntimeConfig) ValidateDefaultRuntime() error {
-	// check we do have at least a runtime
-	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
-		// Set the default runtime to "runc" if default_runtime is not set
-		if c.DefaultRuntime == "" {
-			logrus.Debugf("Defaulting to %q as the runtime since default_runtime is not set", defaultRuntime)
-			// The default config sets runc and its path in the runtimes map, so check for that
-			// first. If it does not exist then we add runc + its path to the runtimes map.
-			if _, ok := c.Runtimes[defaultRuntime]; !ok {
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
The code LGTM for what you're intending to do.  However, I'll have to lean on @giuseppe , @haircommander and @saschagrunert on the safeness of swapping out the runtime like this.

pkg/config/reload.go
```diff
@@ -228,3 +231,32 @@ func (c *Config) ReloadRdtConfig(newConfig *Config) error {
...
+	for runtime := range newConfig.Runtimes {
+		if _, ok := c.Runtimes[runtime]; !ok {
+			continue
+		}
+		newRuntimes = append(newRuntimes, runtime)
+	}
+
+	for _, runtime := range newRuntimes {
+		c.Runtimes[runtime] = newConfig.Runtimes[runtime]
+	}
```
haircommander, line 247: couldn't we just compress these lines and do 
`		c.Runtimes[runtime] = newConfig.Runtimes[runtime]` if it wasn't found?

pkg/config/reload.go
```diff
@@ -228,3 +231,32 @@ func (c *Config) ReloadRdtConfig(newConfig *Config) error {
 	}
 	return nil
 }
+
+// ReloadRuntimes reloads the runtimes configuration if changed
+func (c Config) ReloadRuntimes(newConfig *Config) error {
+	var newRuntimes []string
+	for runtime := range newConfig.Runtimes {
+		if _, ok := c.Runtimes[runtime]; !ok {
+			continue
```
haircommander, line 240: I think we should log a warning that we skipped because it was already present

pkg/config/reload_test.go
```diff
@@ -325,4 +326,40 @@ var _ = t.Describe("Config", func() {
...
+			newConfig.DefaultRuntime = "invalid"
+
+			// When
+			err := sut.ReloadRuntimes(newConfig)
+
+			// Then
+			Expect(err).NotTo(BeNil())
+		})
+
+		It("should add a new runtime", func() {
```
haircommander, line 352: could we add a case where we verify a runtime is unchanged because it's already present?

elezar: Sure. I will add more test cases. I wanted to get this out for early feedback.

pkg/config/config.go
```diff
@@ -1089,6 +1076,27 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 	return nil
 }
 
+// ValidateDefault ensures that the default runtime is set and valid.
+func (c *RuntimeConfig) ValidateDefault() error {
+	// check we do have at least a runtime
+	if _, ok := c.Runtimes[c.DefaultRuntime]; !ok {
+		// Set the default runtime to "runc" if default_runtime is not set
+		if c.DefaultRuntime == "" {
```
haircommander, line 1084: nit, can we invert this condition and compress the else?

`if c.DefaultRuntime != "" { return } ...`

haircommander: I suppose this was already there, but doesn't hurt to update now 🙂

saschagrunert (commented):
The unit test failures may be legit. Can we also add a test to https://github.com/cri-o/cri-o/blob/main/test/reload_config.bats, maybe by injecting a fake runtime?

elezar:
> The unit test failures may be legit. Can we also add a test to https://github.com/cri-o/cri-o/blob/main/test/reload_config.bats, maybe by injecting a fake runtime?

OK. I will have a look.

elezar:
@saschagrunert I have updated the unit tests and also added some tests as suggested. I don't have my development environment quite set up, so I've been resting on the CI a bit more heavily than I probably should. Please take another look (especially at the tests) and let me know if I'm missing something.

I'll try to get my dev env set up a bit more cleanly if more changes are required.

haircommander:
would you still consider this to be a draft @elezar ?

elezar:
> would you still consider this to be a draft @elezar ?

@haircommander no, I don't think so. I just wanted to get the tests looking in a better shape. Not having a local dev environment set up is making that a bit slow.

elezar:
I have rebased the changes here and updated the status.

What are the best docs for getting a development environment set up? Ideally this would be containerized.

pkg/config/reload.go
```diff
@@ -237,3 +240,36 @@ func (c *Config) ReloadRdtConfig(newConfig *Config) error {
...
+		}
+		c.Runtimes[runtime] = newConfig.Runtimes[runtime]
+		logrus.Infof("Registered new runtime %q", runtime)
+		updated = true
+	}
+
+	if c.DefaultRuntime != newConfig.DefaultRuntime {
+		c.DefaultRuntime = newConfig.DefaultRuntime
+		if err := c.ValidateDefault(); err != nil {
```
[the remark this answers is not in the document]
elezar: This was a typo after a rebase. Should be fixed now.

haircommander:
@elezar is this ready?

/retest

haircommander:
/approve

LGTM, thanks @elezar !

@cri-o/cri-o-maintainers PTAL

zvonkok:
@mrunalp PTAL, if this gets merged, in which OCP version can we expect this to land?

haircommander:
> @mrunalp PTAL, if this gets merged, in which OCP version can we expect this to land?

main is currently feeding into 1.26, so 4.13 would be the target

@saschagrunert PTAL

pkg/config/reload.go
```diff
@@ -237,3 +240,36 @@ func (c *Config) ReloadRdtConfig(newConfig *Config) error {
...
+		logConfig("default_runtime", c.DefaultRuntime)
+		updated = true
+	}
+
+	if !updated {
+		return nil
+	}
+
+	if err := c.ValidateRuntimes(); err != nil {
+		return fmt.Errorf("unabled to reload runtimes: %v", err)
```
saschagrunert, line 271:
```suggestion
		return fmt.Errorf("unabled to reload runtimes: %w", err)
```

saschagrunert (commented):
Just one tiny error wrapping nit, otherwise LGTM

pkg/config/reload.go
elezar: Updated and rebased.

haircommander:
thanks for your patience as we went through a CI migration and multiple outages

/lgtm

haircommander:
oh bother
/hold

elezar:
@haircommander thanks for giving me the heads-up regarding the CI issues. I have rebased the changes off `main` again. Are CI issues still expected?

PR #6070, opened 2022-07-18
Inherit capabilities for container execs

### LINKED ISSUE #6069: Setcap doesn't work with cri-o v1.24.1
### What happened?

A test with a non-root container fails on master;
```
vm-002 ~ # kubectl exec -it alpine-test-kc2c5 -- sh
~ $ id
uid=1000(1000) gid=3000
~ $ getcap /sbin/ip
/sbin/ip cap_net_admin=eip
~ $ /sbin/ip addr add 10.10.10.10/32 dev lo
RTNETLINK answers: Operation not permitted
```
The `/sbin/ip` has cap_net_admin=eip and this works on earlier version. We have a requirement on non-root containers but we must use the `ip` command in multi-network setups.

I thought this was K8s, but it was my upgrade of `cri-o`, v1.22.0 -> v1.24.1.

Please see https://github.com/kubernetes/kubernetes/issues/111196.

Everything is there and it's late so I don't want to repeat it.

Works as for cri-o v1.22.0

Please see https://github.com/kubernetes/kubernetes/issues/111196.

_No response_

```
  23:17:59 Server Version: v1.24.3
  23:17:59 Linux 5.18.1 #8 SMP PREEMPT_DYNAMIC Sat Jun 18 12:53:37 CEST 2022
  23:17:59 CNI-plugin; bridge
  23:17:59 Proxy-mode: "ipvs"
  23:17:59 crio version 1.24.1
```

### OS version

Own BusyBox Linux

Kvm VMs

saschagrunert:
#### What type of PR is this?

/kind bug

We dropped the inheritable capabilities support for the container creation in 3e9d7725714057d5fc2ce7eb9f905dc50c96647c to fix a security vulnerability. This had the side effect that all exec's within containers break the existing behavior if they rely on the initially set capabilities, because the exec process does not inherit the capabilities of the main container any more. We now explicitly set the inheritable capabilities for all exec's to restore the changed behavior.

Fixes #6069

It feels that this slightly decreases security.

```release-note
Fixed a bug where the capabilities of containers are not applied to exec (sync) requests.
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1340,6 +1340,10 @@ func prepareProcessExec(c *Container, cmd []string, tty bool) (processFile strin
 	if tty {
 		pspec.Terminal = true
 	}
+	// Allow the same capabilities to be used as those from the actual container process.
+	if pspec.Capabilities != nil {
+		pspec.Capabilities.Inheritable = pspec.Capabilities.Bounding
+	}
 	processJSON, err := json.Marshal(pspec)
 	if err != nil {
 		return "", err
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -669,6 +669,18 @@ function check_oci_annotation() {
 	crictl create "$newconfig" "$TESTDATA"/sandbox_config.json
 }
 
+@test "ctr exec caps" {
+	start_crio
+
+	jq '.linux.security_context.capabilities = { "add_capabilities": ["net_admin"] }' \
+		"$TESTDATA"/container_redis.json > "$newconfig"
+
+	ctr_id=$(crictl run "$newconfig" "$TESTDATA"/sandbox_config.json)
+
+	cap_inh=$(crictl exec --sync "$ctr_id" cat /proc/self/status | grep CapInh: | cut -f2)
+	[[ "$cap_inh" != "0000000000000000"* ]]
+}
+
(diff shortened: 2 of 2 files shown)


haircommander:
I wanna give this one some thought. It seems to be against the spirit of the fix for the security issue, but that one was more about idiomatic configuration and not so much about an actual vulnerability. It's possible this is the correct choice for us, but I need to mull more

internal/oci/runtime_oci.go
```diff
@@ -1340,6 +1340,8 @@ func prepareProcessExec(c *Container, cmd []string, tty bool) (processFile strin
 	if tty {
 		pspec.Terminal = true
 	}
+	// Allow the same capabilities to be used as those from the actual container process.
+	pspec.Capabilities.Inheritable = pspec.Capabilities.Bounding
```
TomSweeneyRedHat, line 1344: Is there a need to check if there's a value in `pspec.Capabilities.Inheritable` first, or will it always be empty?

TomSweeneyRedHat:
One question, otherwise LGTM

internal/oci/runtime_oci.go
saschagrunert: I think we can check for `pspec.Capabilities != nil` before setting the field to be on the safe side.

saschagrunert:
Closing in favor of https://github.com/cri-o/cri-o/pull/6236

PR #6094, opened 2022-07-26
Support for Evented PLEG

sairameshv:
#### What type of PR is this?
/kind feature

This PR implements the following enhancement related to the [Evented PLEG](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/3386-kubelet-evented-pleg/README.md)
Fixes # https://github.com/kubernetes/enhancements/issues/3386

This PR contains a new config option to enable "evented-pleg" while starting/running the crio binary. The same has been updated wherever required.

```release-note
Added a new boolean configuration flag "--evented-pleg"(defaulted to "false") to enable the evented pleg mechanism in cri-o. The environment variable "EVENTED_PLEG" when set to "true" also enables the evented pleg in the cri-o.
```

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -46,6 +46,7 @@ crio
 [--enable-metrics]
 [--enable-profile-unix-socket]
 [--enable-tracing]
+[--evented-pleg]
 [--gid-mappings]=[value]
 [--global-auth-file]=[value]
 [--grpc-max-recv-msg-size]=[value]
@@ -214,6 +215,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 **--enable-profile-unix-socket**: Enable pprof profiler on crio unix domain socket
 
 **--enable-tracing**: Enable OpenTelemetry trace data exporting
+
+**--evented-pleg**: Enable CRI-O to generate the container pod-level events in order to optimize the performance of the Pod Lifecycle Event Generator (PLEG) module in Kubelet.
 
 **--gid-mappings**="": Specify the GID mappings to use for the user namespace (default: "")
 
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -451,7 +451,12 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 		if err := s.Runtime().UpdateContainerStatus(ctx, newContainer); err != nil {
 			return nil, fmt.Errorf("failed to update the container status %s: %w", newContainer.ID(), err)
 		}
-		s.ContainerEventsChan <- types.ContainerEventResponse{ContainerId: newContainer.ID(), ContainerEventType: types.ContainerEventType_CONTAINER_CREATED_EVENT, CreatedAt: time.Now().UnixNano(), PodSandboxMetadata: s.GetSandbox(newContainer.CRIContainer().PodSandboxId).Metadata()}
+		select {
+		case s.ContainerEventsChan <- types.ContainerEventResponse{ContainerId: newContainer.ID(), ContainerEventType: types.ContainerEventType_CONTAINER_CREATED_EVENT, CreatedAt: time.Now().UnixNano(), PodSandboxMetadata: s.GetSandbox(newContainer.CRIContainer().PodSandboxId).Metadata()}:
+			log.Debugf(ctx, "Container created event generated for %s", newContainer.ID())
+		default:
+			log.Errorf(ctx, "CreateCtr: failed to send container created event %s", newContainer.ID())
+		}
 	}
 
 	log.Infof(ctx, "Created container %s: %s", newContainer.ID(), newContainer.Description())
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -446,6 +446,10 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 
 	newContainer.SetCreated()
 
+	// s.Runtime().UpdateContainerStatus(ctx, newContainer) // TODO - handle returned err
+
+	// s.ContainerEventsChan <- types.ContainerEventResponse{ContainerId: newContainer.ID(), ContainerEventType: types.ContainerEventType_CONTAINER_CREATED_EVENT, PodSandboxMetadata: s.GetSandbox(newContainer.CRIContainer().PodSandboxId).Metadata()}
+
 	log.Infof(ctx, "Created container %s: %s", newContainer.ID(), newContainer.Description())
 	return &types.CreateContainerResponse{
 		ContainerId: ctr.ID(),
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/mount"
@@ -446,9 +447,12 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 
 	newContainer.SetCreated()
 
-	// s.Runtime().UpdateContainerStatus(ctx, newContainer) // TODO - handle returned err
-
-	// s.ContainerEventsChan <- types.ContainerEventResponse{ContainerId: newContainer.ID(), ContainerEventType: types.ContainerEventType_CONTAINER_CREATED_EVENT, PodSandboxMetadata: s.GetSandbox(newContainer.CRIContainer().PodSandboxId).Metadata()}
+	if s.config.EventedPLEG {
+		if err := s.Runtime().UpdateContainerStatus(ctx, newContainer); err != nil {
+			return nil, fmt.Errorf("failed to update the container status %s: %w", newContainer.ID(), err)
+		}
+		s.ContainerEventsChan <- types.ContainerEventResponse{ContainerId: newContainer.ID(), ContainerEventType: types.ContainerEventType_CONTAINER_CREATED_EVENT, CreatedAt: time.Now().UnixNano(), PodSandboxMetadata: s.GetSandbox(newContainer.CRIContainer().PodSandboxId).Metadata()}
+	}
 
 	log.Infof(ctx, "Created container %s: %s", newContainer.ID(), newContainer.Description())
 	return &types.CreateContainerResponse{
--- a/server/container_remove.go
+++ b/server/container_remove.go
@@ -31,6 +31,10 @@ func (s *Server) RemoveContainer(ctx context.Context, req *types.RemoveContainer
 		return err
 	}
 
+	s.Runtime().UpdateContainerStatus(ctx, c)
+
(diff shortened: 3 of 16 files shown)


docs/crio.8.md
```diff
@@ -212,6 +213,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--enable-tracing**: Enable OpenTelemetry trace data exporting
 
+**--evented-pleg**: Whether CRI-O should generate the container events in order to optimize the performance of Kubelet to rely on these events.
```
TomSweeneyRedHat, line 216: I'm not terribly familiar with this option, and this description is not helping whatsoever.  Plus, is it a boolean?  If you make changes here, please also change in crio.fish.

Maybe?
```suggestion
**--evented-pleg**: Enable CRI-O to generate the container pod-level events in order to optimize the performance of the Pod Lifecycle Event Generator (PLEG) module in Kubelet.
```

sairameshv: Thanks Tom, Noted.

server/container_create.go
```diff
@@ -425,6 +425,13 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 
 	newContainer.SetCreated()
 
+	// if s.config.EventedPLEG {
+	// if err := s.Runtime().UpdateContainerStatus(ctx, newContainer); err != nil {
+	// return fmt.Errorf("failed to update the container status %s: %v", newContainer.ID(), err)
+	// }
+	// s.ContainerEventsChan <- types.ContainerEventResponse{ContainerId: newContainer.ID(), ContainerEventType: types.ContainerEventType_CONTAINER_CREATED_EVENT, PodSandboxMetadata: s.GetSandbox(newContainer.CRIContainer().PodSandboxId).Metadata()}
+	// }
```
saschagrunert, line 433: Let's remove this if not needed.

server/container_remove.go
```diff
@@ -22,6 +23,12 @@ func (s *Server) RemoveContainer(ctx context.Context, req *types.RemoveContainer
 		return err
 	}
 
+	if s.config.EventedPLEG {
+		if err := s.Runtime().UpdateContainerStatus(ctx, c); err != nil {
+			return fmt.Errorf("failed to update container status %s: %v", req.ContainerId, err)
```
saschagrunert, line 28:
```suggestion
			return fmt.Errorf("failed to update container status %s: %w", req.ContainerId, err)
```

also for the other occurrences.

server/container_start.go
```diff
@@ -57,6 +57,13 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
 	if err := s.Runtime().StartContainer(ctx, c); err != nil {
 		return fmt.Errorf("failed to start container %s: %v", c.ID(), err)
 	}
+	if s.config.EventedPLEG {
+		if err := s.Runtime().UpdateContainerStatus(ctx, c); err != nil {
+			return fmt.Errorf("failed to update container status %s: %v", c.ID(), err)
+		}
+		// time.Sleep(3 * time.Second)               // TODO - remove this sleep
```
saschagrunert, line 64:
```suggestion
```

server/container_start.go
```diff
@@ -57,6 +57,13 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
 	if err := s.Runtime().StartContainer(ctx, c); err != nil {
 		return fmt.Errorf("failed to start container %s: %v", c.ID(), err)
 	}
+	if s.config.EventedPLEG {
+		if err := s.Runtime().UpdateContainerStatus(ctx, c); err != nil {
+			return fmt.Errorf("failed to update container status %s: %v", c.ID(), err)
+		}
```
saschagrunert, line 63: I'm wondering why we have to update the container status? Before throwing an event?

vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto
```diff
@@ -18,7 +18,7 @@ limitations under the License.
 syntax = "proto3";
 
 package runtime.v1;
-option go_package = "v1";
+option go_package = "k8s.io/cri-api/pkg/apis/runtime/v1";
```
saschagrunert, line 21: I guess we have to vendor the latest Kubernetes via go.mod here, right?

docs/crio.8.md
```diff
@@ -212,6 +213,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--enable-tracing**: Enable OpenTelemetry trace data exporting
 
+**--evented-pleg**: Enable CRI-O to generate the container pod-level events in order to optimize the performance of the Pod Lifecycle Event Generator (PLEG) module in Kubelet.
```
saschagrunert, line 216: I'm not sure If I'm in favor of manually enabling/disabling the feature. What is the reason for not automatically decide that?

sairameshv: According to the following [discussion](https://kubernetes.slack.com/archives/C03HA69L6R1/p1657179643142429?thread_ts=1657126350.580129&cid=C03HA69L6R1), to be on sync with the feature gate in kubelet, a config switch in containerd, this has been added in CRI-O.

cc: @harche @rphillips

server/container_create.go
sairameshv: Sure, noted.

server/container_remove.go
sairameshv: Addressed

server/container_start.go
sairameshv: This [link](https://kubernetes.slack.com/archives/C03HA69L6R1/p1658751854843449) actually summarizes the issue and the fix looks to be updating the pod status prior to the generation of an event.

cc: @harche

server/container_start.go
sairameshv: Addressed

server/sandbox_remove.go
```diff
@@ -67,6 +67,13 @@ func (s *Server) removePodSandbox(ctx context.Context, sb *sandbox.Sandbox) erro
 		return fmt.Errorf("failed to delete pod sandbox %s from index: %w", sb.ID(), err)
 	}
 
+	if s.config.EventedPLEG {
+		if err := s.Runtime().UpdateContainerStatus(ctx, sb.InfraContainer()); err != nil {
```
haircommander, line 71: why this UpdateContainerStatus?

server/cri/v1/rpc_container_events.go
```diff
@@ -0,0 +1,29 @@
...
+
+	if s.server.Config().EventedPLEG {
+		for containerEvent := range s.server.ContainerEventsChan {
+			if err := ces.Send(&containerEvent); err != nil {
+				return err
+			}
+		}
+	}
+
+	// Dummy code to generate a sample events stream
```
haircommander, line 17: we should remove this in the final version.

server/sandbox_run_linux.go
```diff
@@ -962,6 +962,12 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 		return nil, ctx.Err()
 	}
 	sb.SetCreated()
+	if s.config.EventedPLEG {
+		if err := s.Runtime().UpdateContainerStatus(ctx, sb.InfraContainer()); err != nil {
```
haircommander, line 966: again here. updatecontainerstatus is a fairly expensive call, so we should call it as few times as possible

go.mod
```diff
@@ -431,7 +431,7 @@ replace (
 	k8s.io/component-base => k8s.io/kubernetes/staging/src/k8s.io/component-base v0.0.0-20220624161656-6219eed24f3c
 	k8s.io/component-helpers => k8s.io/kubernetes/staging/src/k8s.io/component-helpers v0.0.0-20220624161656-6219eed24f3c
 	k8s.io/controller-manager => k8s.io/kubernetes/staging/src/k8s.io/controller-manager v0.0.0-20220624161656-6219eed24f3c
-	k8s.io/cri-api => k8s.io/kubernetes/staging/src/k8s.io/cri-api v0.0.0-20220624161656-6219eed24f3c
+	k8s.io/cri-api => k8s.io/kubernetes/staging/src/k8s.io/cri-api v0.0.0-20220802183113-668b2440c559
```
haircommander, line 434: I think they'll cut 1.25.0.beta.0 soon so we should wait till that IMO

saschagrunert: We also should vendor all k8s.io dependendencies to the same commit/tag.

server/cri/v1/rpc_container_events.go
sairameshv: Sure.

server/sandbox_run_linux.go
sairameshv: This [link](https://kubernetes.slack.com/archives/C03HA69L6R1/p1658751854843449) actually summarizes the issue and the possible fix looks to be updating the pod status prior to the generation of an event.

cc: @harche

go.mod
sairameshv: Sure

server/sandbox_remove.go
harche: @haircommander without this we found that kubelet pod cache was not getting correctly updated. In Generic PLEG in about 1s the cache is refreshed, but in evented pleg there is no update to the cache unless another event is generated for that pod. During testing, we saw that kubelet was stuck with old pod status without that call to `UpdateContainerStatus`. 

Even though they are expensive, they are executed only in case the evented pleg is enabled.

haircommander: yeah but in the future evented pleg will be the default. Maybe we should figure out how to update the status more efficiently so we incur less overhead. I guess we don't have to block, but we should open an issue and add TODOs to remove them or make them more efficient

haircommander: also, would you mind reminding me, it's my naive assumption the PLEG causes the pod cache to update. Why is sending the event not sufficient?

server/sandbox_stop_linux.go
```diff
@@ -84,6 +86,12 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error
 
 	log.Infof(ctx, "Stopped pod sandbox: %s", sb.ID())
 	sb.SetStopped(true)
+	if s.config.EventedPLEG {
+		if err := s.Runtime().UpdateContainerStatus(ctx, sb.InfraContainer()); err != nil {
```
mrunalp, line 90: This is called during the stop logic, so not sure why we are adding a call here.

server/sandbox_remove.go
harche: The way it works is, 
1. CRIO sends an event
2. Kubelet requests pod status 
3. Kube cache is updated with the latest pod status. 

During testing we had noticed that incorrect pod status was getting updated in the kubelet without `UpdateContainerStatus`. Having said that, I need to some testing to see if we can remove this call from `remove` operation.

server/sandbox_stop_linux.go
harche: I agree, will report here after testing if we really need to have this `UpdateContainerStatus` during stop/remove

sairameshv:
/close
A new PR will be opened to support the Evented PLEG changes

PR #6097, opened 2022-07-28
WIP: Add cached based name reservation error handling

saschagrunert:
#### What type of PR is this?

/kind bug

We now allow up to 10 consecutive "name is reserved" errors within 10 minutes before CRI-O removes the container by itself. This allows container being stuck in name reservation to be released so that the kubelet can create them accordingly.

None

None

```release-note
None
```


saschagrunert:
Not required any more.

PR #6115, opened 2022-08-02
node_e2e_installer: be use runc/crun from PATH

haircommander:
the get script only installs runc if it can't be found in path, which is
good behavior. Use the one found in path when installing the drop-in
files for node e2e tests

#### What type of PR is this?

/kind bug

```release-note
none
```

### DIFF
--- a/scripts/node_e2e_installer
+++ b/scripts/node_e2e_installer
@@ -30,20 +30,17 @@ install_crio() {
     # Setup log level
     echo "CONTAINER_LOG_LEVEL=debug" >>/etc/sysconfig/crio
 
-    # Setup secondary runtime
+    cat <<EOF >/etc/crio/crio.conf.d/10-crun.conf
+[crio.runtime]
+[crio.runtime.runtimes]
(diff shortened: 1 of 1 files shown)


haircommander:
related to https://github.com/kubernetes/test-infra/pull/26998

scripts/node_e2e_installer
```diff
@@ -30,20 +30,22 @@ install_crio() {
     # Setup log level
     echo "CONTAINER_LOG_LEVEL=debug" >>/etc/sysconfig/crio
 
+    runc_path=$(command -v runc || true)
```
saschagrunert, line 33: Do we wanna fail if it's not in $PATH? Or: do we wanna fallback to `/usr/local/bin/runc` in that case?

saschagrunert: @haircommander yeah CRI-O should find it in $PATH as well :+1:

saschagrunert (approved):
/lgtm

Issue #5720, opened 2022-03-11
Add seccomp option to stop containers based on blocked syscall


saschagrunert:
I've created a PoC which utilized the seccomp notify feature to stop containers based on blocked seccomp syscalls:
https://github.com/cri-o/cri-o/compare/main...saschagrunert:seccomp-notifier

The idea is to rewrite the seccomp profile (runtime/default in case of the PoC) to change the actions `ActErrno`/`ActKill`/`ActKillProcess`/`ActKillThread` to `ActNotify` as well as setting the `ListenerPath` to a temporarily location. After that, a unix socket as well as libseccomp can be used to pass around the fd's and notify if any syscall may get blocked or lead into termination of the process. If that happens, we immediately stop the container and set the exit reason to something like `seccomp:syscall`.

For example, if we run `swapoff -a` within a container running `RuntimeDefault` as well as having a `RestartPolicy: "Never"` would then look like this:
![img](https://user-images.githubusercontent.com/695473/157830739-3b6f97a6-3847-4634-8c30-61775483dcca.png)

We could add this as optional feature to CRI-O to increase the seccomp observability. Throwing metrics would be a good addition as well. A main drawback is that we only catch the first erroneous syscalls, because from that point in time we declare the workload as "undefined state" and stop the container.

The whole idea was inspired by the seccomp agent: https://github.com/opencontainers/runc/tree/main/contrib/cmd/seccompagent

saschagrunert:
A friendly reminder that this issue had no activity for 30 days.

haircommander:
@saschagrunert did you still want to work on this?

saschagrunert:
PR open in  #6120

PR #6120, opened 2022-08-03
Add seccomp notifier feature

saschagrunert:
#### What type of PR is this?

/kind feature

Added seccomp notifier feature to support stopping containers on seccomp syscalls.

Fixes https://github.com/cri-o/cri-o/issues/5720

None

```release-note
Added seccomp notifier feature, which can be enabled by setting the annotation `io.kubernetes.cri-o.seccompNotifierAction` either to `stop` (for terminating the workload) or anything else to just create metrics or logs. This also includes the new metric `crio_containers_seccomp_notifier_count_total`. For more information on its usage, please refer to the [crio.conf.5](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md) man page.
```


saschagrunert:
Ready for review. PTAL @cri-o/cri-o-maintainers

docs/crio.conf.5.md
```diff
@@ -138,6 +138,31 @@ the container runtime configuration.
...
+  To be able to use this feature, just enable it via the flag below and
+  configure a runtime which has the annotation
+  "io.kubernetes.cri-o.seccompNotifier" in the `allowed_annotations` array.
+
+  It also requires at least runc 1.1.0 or crun 0.19 which support the notifier
+  feature.
+
+  If everything is setup, CRI-O will modify chosen seccomp profiles for
+  containers if the annotation "io.kubernetes.cri-o.seccompNotifier" is set on
+  the Pod sandbox. CRI-O will then get notified if a container is using a
```
haircommander, line 157: what is the difference with cri-o terminating because of a notification vs a pod failing on EPERM from a blocked syscall? I wonder if we should notify in a different way (metric and cri-o log, maybe?)

server/server.go
```diff
@@ -652,6 +662,91 @@ func (s *Server) MonitorsCloseChan() chan struct{} {
 	return s.monitorsChan
 }
 
+func (s *Server) startSeccmpNotifierWatcher() error {
```
haircommander, line 665: I know it's a long function, but I find it weird that we drop the o in seccomp here

haircommander: was that intentional

haircommander:
this is really awesome! I have some structural notes. I will do another pass after we address those.

@kolyshkin @giuseppe @mrunalp I think we want some more eyes here as well

docs/crio.conf.5.md
```diff
@@ -138,6 +138,31 @@ the container runtime configuration.
   Changes the meaning of an empty seccomp profile.  By default (and according to CRI spec), an empty profile means unconfined.
   This option tells CRI-O to treat an empty profile as the default profile, which might increase security.
 
+**seccomp_use_notifier**=false
```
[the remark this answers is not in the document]
saschagrunert: Good, removed that option and also the corresponding integration test. Updated the docs as well.

server/server.go
saschagrunert: No, fixed it :+1:

docs/crio.conf.5.md
saschagrunert: We cannot find the root cause of the pod termination easily because audit does not logs act kill for example.

A metric is interesting, but we will only catch the first wrong syscall because then the termination happens.

I was thinking if it would make sense to collect the wrong syscalls (until when?) and then throw the metrics. Hm… :thinking:

haircommander: from my perspective, the notify feature should be used to do information collection. I'm thinking it'd be used for syscalls that folks aren't worried about, but they want to tighten the syscalls allowed to reduce attack surface. Maybe as the second step, after using the bpf hook for a bit to filter the commonly used ones. The remaining (non dangerous) could be set to notify, and cri-o would notify if they're used. If a notification happens, the user could update the profile to block or allow, depending on how they feel about the syscall.

saschagrunert: Alright, I see the following use cases for now:

- Removing unnecessary syscalls from my already existing seccomp profile
- Validating if a profile has negative impact on my application

Both could be achieved via metrics and logging, but would require manual monitoring of them. The container stop is immediately visible to users, which I see as a benefit. How about:

- Adding metrics for the notifications
- Making the container stop behavior optional

giuseppe:
Interesting idea!

How will it work when CRI-O is restarted?

saschagrunert:
> How will it work when CRI-O is restarted?

Thank you! On CRI-O restart it tries to recover the state and reconnects to the socket: https://github.com/cri-o/cri-o/pull/6120/files#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4R673-R706

This means that we probably loose notifier events from previous connections (like the start PID of the container workload).

docs/crio.conf.5.md
saschagrunert: Made the container stop behavior configurable. I think we should not add metrics since all of them would have a high cardinality.

saschagrunert:
@haircommander I added the metric and rebased this PR. PTAL again.

docs/crio.conf.5.md
```diff
@@ -337,6 +337,32 @@ A workload is chosen for a pod based on whether the workload's **activation_anno
...
+  "io.kubernetes.cri-o.seccompNotifier" for enabling the seccomp notifier feature.
+
+#### Using the seccomp notifier feature:
+
+This feature can help you to debug seccomp related issues, for example if
+blocked syscalls (permission denied errors) have negative impact on the
+workload.
+
+To be able to use this feature, configure a runtime which has the annotation
+"io.kubernetes.cri-o.seccompNotifier" in the `allowed_annotations` array.
```
haircommander, line 349: sorry should have mentioned it earlier: what do you think about having this annotation be `seccompNotifierAction` and have it take an action. we can start with `stop` or `ignore/""` but maybe in the future we'd want another action

saschagrunert: Alright, changed as suggested :+1:

saschagrunert:
Green tests, PTAL again @cri-o/cri-o-maintainers

internal/config/seccomp/notifier.go
```diff
@@ -0,0 +1,370 @@
...
+			go handler(ctx, containerID, msgChan, libseccomp.ScmpFd(newFd))
+		}
+	}()
+
+	action, ok := annotationMap[annotations.SeccompNotifierActionAnnotation]
+	if !ok {
+		return nil, fmt.Errorf("%s annotation not set on container", annotations.SeccompNotifierActionAnnotation)
+	}
+
+	const actionStop = "stop"
```
haircommander, line 213: WDYT about putting this in `pkg/annotations` or some similar public place? I know we document the value but it also seems worth allowing folks to vendor it. I'm 50/50 so I am fine as a followup or fine to not take that nit

saschagrunert: Good thought, changed it as suggested :+1:

haircommander:
/lgtm

nice work here!

saschagrunert:
Uhm, it’s since hours in merge pool. Merging now.

PR #6122, opened 2022-08-03
oci: take opLock for UpdateContainer

haircommander:
UpdateContainer requests can come in at any time in a container's
lifecycle, even when it's shutting down.

In some cases, it can cause a container to get stuck: runc pauses the
cgroup on v1 and systemd to prevent systemd from mucking around while
the cgroup is being updated. In some cases, CRI-O calls
`UpdateContainerStatus` during that time, and gets `paused` state. CRI-O
does not expect a container to be paused, so it doesn't know how to
handle it.

Fix this by taking opLock when updating

#### What type of PR is this?

/kind bug
related to https://bugzilla.redhat.com/show_bug.cgi?id=2094865

```release-note
fix a bug where a container is stuck in INIT because CRI-O believes it to be paused, and never updates the state
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -698,6 +698,9 @@ func TruncateAndReadFile(ctx context.Context, path string, size int64) ([]byte, 
 
 // UpdateContainer updates container resources
 func (r *runtimeOCI) UpdateContainer(ctx context.Context, c *Container, res *rspec.LinuxResources) error {
+	c.opLock.Lock()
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests

haircommander:
@saschagrunert PTAL

haircommander:
@saschagrunert PTAL

saschagrunert (approved):
/lgtm

PR #6111, opened 2022-08-02
WIP: Revert "server: properly handle duplicate requests"

saschagrunert:
This reverts commit 9c5eb67c3f11b450a1de56704ce31bd592356869.

PR #6123, opened 2022-08-03
server: return already created ID for duplicated requests

haircommander:
turns out, kubelet does behave like this in some scenerios (static
pods). Instead of failing forever, return the already created pod.

#### What type of PR is this?

/kind bug
alternative to https://github.com/cri-o/cri-o/pull/6111

```release-note
Fix a bug where static pods cannot be created because they've already been created.
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -374,9 +374,9 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 			return nil, fmt.Errorf("failed to get ID of container with reserved name (%s), after failing to reserve name with %v: %w", ctr.Name(), getErr, getErr)
 		}
 		// if we're able to find the container, and it's created, this is actually a duplicate request
-		// from a client that does not behave like the kubelet (like crictl)
+		// Just return that container
 		if reservedCtr := s.GetContainer(reservedID); reservedCtr != nil && reservedCtr.Created() {
-			return nil, err
+			return &types.CreateContainerResponse{ContainerId: reservedID}, nil
 		}
 		cachedID, resourceErr := s.getResourceOrWait(ctx, ctr.Name(), "container")
 		if resourceErr == nil {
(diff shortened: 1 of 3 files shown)


saschagrunert (approved):
/lgtm
/hold
Feel free to unhold if this is ready to be merged.

TomSweeneyRedHat:
LGTM assuming happy tests

haircommander:
for you, anything

saschagrunert (approved):
/lgtm

haircommander:
/hold

I can't figure out why just kata is failing

@littlejawa do you have any insight?

TomSweeneyRedHat (commented):
LGTM
once the tests are hapy

littlejawa:
> /hold
> 
> I can't figure out why just kata is failing
> 
> @littlejawa do you have any insight?

Hey @haircommander,
I am not sure, but I think you're finding that this test (shm_size.bats : "check /dev/shm fails with incorrect values") actually never worked the way we thought with kata.
It is using the helpers.bat function `create_runtime_with_allowed_annotation`, and it seems we are skipping all the tests that use this function in kata because they fail - I suspect a bad interaction between this function, which changes the default runtime, and our test framework that makes sure kata is being used.
I am not sure that in the kata CI job, `create_runtime_with_allowed_annotation` actually changes the runtime being used by cri-o, and so the annotations are just not set.

The reason why we were not skipping this specific test is because it seemed to work, but looking at it, it is checking for failures, and doesn't verify the reason of the failure. So I am guessing that it was failing up to now because annotations were just not taken into account, and this made the test "pass", for the wrong reasons.

I am not sure why it changed its behavior now, but apparently your change makes the second call to `crictl runp` find an existing pod and returning it (this is what the cri-o log show), ignoring the potential issue with annotation, and so reporting success?
I don't know why the pod is still around at that time though.

Anyway - TL;DR: I think you should ignore this error, and I will modify our test script to skip this test, until we fix the function in question.

haircommander:
/hold

actually, ci is broken at the moment, let's wait till it's ready

saschagrunert:
@haircommander can you please rebase that we can merge it soon'ish?

saschagrunert (approved):

PR #6128, opened 2022-08-04
[release-1.24] Add packit configuration

LorbusChris:
#### What type of PR is this?
/kind ci

This change adds a config file for the Packit-as-a-Service GH app (see: https://packit.dev).
With this config, packit will automatically build an RPM for each new release tag.
The config has to exist only on the branches that are supposed to be built, i.e. currently `release-1.24`.
In the future, a copy of this config file will have to be added to `release-1.25`+ branches, too.

None

/cc @haircommander 

None

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,22 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+specfile_path: cri-o.spec
+
+# name in upstream package repository or registry (e.g. in PyPI)
+upstream_package_name: cri-o
+upstream_tag_template: v{version}
+# downstream (Fedora) RPM package name
+downstream_package_name: cri-o
+actions:
+  post-upstream-clone: "wget https://src.fedoraproject.org/rpms/cri-o/raw/rawhide/f/cri-o.spec -O cri-o.spec"
+jobs:
+- job: copr_build
+  trigger: release
+  owner: "@OKD"
+  project: okd
+  targets:
+    - centos-stream-9-aarch64
+    - centos-stream-9-x86_64
+    - fedora-all-aarch64
+    - fedora-all-x86_64


haircommander:
/ok-to-test

so is okd looking for 4.11 support now?

LorbusChris:
> so is okd looking for 4.11 support now?

@haircommander no, the OKD on CentOS Stream CoreOS variant the packit builds are for is aimed at 4.12, so I assume by then there'll be a cri-o 1.25 release, to which OKD will then switch.

haircommander:
yeah we should aim this PR to main, we pull back changes to main to release-1.25 automatically

LorbusChris:
The problem with that is that the config file has to be on the branch that is to be built. main does not have all the release tags on it, so if I want to automatically get a build of future v1.24.2 (or even the existing v1.24.1), the config file has to be present on release-1.24.

haircommander:
what if it was targeted at release-1.25?

LorbusChris:
It really depends on what packit is used for.
With this config, packit will only look at the latest tag on the branch, and build that, which makes sense for both `release-1.24` and `release-1.25`, but not for main. It does this by running `git describe --tags --abbrev=0`, which on both main and release-1.25 yields `v1.24.0` currently.

For main, I think it would make more sense to configure packit to build PRs instead of release tags.

haircommander:
I actually think we should base on PR. For openshift, we autobuild cri-o from whatever relevant version we're building against (we're already autobuilding 1.25 into 4.12 now). When we cut 1.25.0, we stop auto forwarding from main, but keep autobuilding. We could do a similar approach with packit so okd is properly acting as an upstream. Otherwise, the cri-o version in the latest openshift version could be ahead of the one for okd. Since we're not currently applying PRs to 1.25 directly (but doing from main and backporting) I think we should apply this to main, and once we cut the tag for 1.25, we remove the packit configuration from main. when we're reading to build from 1.26 for okd 4.13, we can reintroduce in main and remove from 1.25

wdyt?

LorbusChris:
ahh OK, if 4.12 already gets main/unreleased 1.25 builds that might make sense indeed. I do need stable versions built for CentOS Stream 9.

There are actually three different types of triggers, PRs (pre-commit), commits (post), and releases.
So my proposal would now be to add both commit AND release build triggers on main/release-1.25. In addition a PR trigger could be added for additional CI testing.

haircommander:
that works with me :smile:

PR #6164, opened 2022-08-23
Adding annotations for image and sandbox name.

littlejawa:
#### What type of PR is this?

/kind bug

These annotations are present for the container, but not the sandbox.
They are required for peer-pods VMs, used to identify the right image to use.
Without them, the remote VM fails to pull the pause image.

```release-note
Added ImageName and SandboxName annotations to the sandbox
```


saschagrunert (approved):
/ok-to-test

Thank you @littlejawa! I think it would be good to enhance an integration test to test against those fields, but I'm not blocking on that request.

PR #6181, opened 2022-08-29
support checkpointing to oci image

adrianreber:
/kind feature

This adds the Podman support to export checkpoints to OCI images which can be pushed to registry to CRI-O.

With this change it is possible to do something like this:
```shell
# crictl checkpoint --export=localhost/checkpoint-image:tag1 CTR-ID
# podman push localhost/checkpoint-image:tag1 quay.io/adrianreber/checkpoint-image:tag1
```
The image `quay.io/adrianreber/checkpoint-image:tag1` can then be used on the same or another system to restore the container using `crictl create` and `crictl start`.

The image `quay.io/adrianreber/checkpoint-image:tag1` can also be used in Kubernetes to start a container from a checkpoint image without Kubernetes knowing that it is restore.

This PR is not targeting the 1.25.0 CRI-O release.

A side effect from this PR is that `crictl inspecti` will contain `spec.annotations`.

None

```release-note
This introduces the ability to store checkpoint archives as OCI images and push the checkpoint images to a remote registry. Important to remember is that the checkpoint image contains all memory pages of the checkpoint and therefore might contain sensitive information (password, encryption keys, ...).
```

### DIFF
--- a/internal/lib/restore.go
+++ b/internal/lib/restore.go
@@ -9,6 +9,7 @@ import (
 	metadata "github.com/checkpoint-restore/checkpointctl/lib"
 	"github.com/checkpoint-restore/go-criu/v5/stats"
 	"github.com/containers/podman/v4/pkg/checkpoint/crutils"
+	"github.com/containers/storage/pkg/archive"
 	"github.com/cri-o/cri-o/internal/oci"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/sirupsen/logrus"
@@ -63,8 +64,43 @@ func (c *ContainerServer) ContainerRestore(ctx context.Context, opts *ContainerC
 	}
 
 	if ctr.RestoreArchive() != "" {
-		if err := crutils.CRImportCheckpointWithoutConfig(ctr.Dir(), ctr.RestoreArchive()); err != nil {
-			return "", err
+		if ctr.RestoreIsOCIImage() {
+			logrus.Debugf("Restoring from %v", ctr.RestoreArchive())
+			imageMountPoint, err := c.StorageImageServer().GetStore().MountImage(ctr.RestoreArchive(), nil, "")
+			if err != nil {
+				return "", err
+			}
+			logrus.Debugf("Checkpoint image mounted at %v", imageMountPoint)
+			defer func() {
+				_, err := c.StorageImageServer().GetStore().UnmountImage(ctr.RestoreArchive(), true)
+				if err != nil {
+					logrus.Errorf("Failed to unmount checkpoint image: %q", err)
+				}
+			}()
+
+			// Import all checkpoint files except ConfigDumpFile and SpecDumpFile. We
+			// generate new container config files to enable to specifying a new
+			// container name.
+			checkpoint := []string{
+				"artifacts",
+				metadata.CheckpointDirectory,
+				metadata.DevShmCheckpointTar,
+				metadata.RootFsDiffTar,
+				metadata.DeletedFilesFile,
+				metadata.PodOptionsFile,
+				metadata.PodDumpFile,
+			}
+			for _, name := range checkpoint {
+				src := filepath.Join(imageMountPoint, name)
+				dst := filepath.Join(ctr.Dir(), name)
+				if err := archive.NewDefaultArchiver().CopyWithTar(src, dst); err != nil {
+					logrus.Debugf("Can't import '%s' from checkpoint image", name)
+				}
+			}
+		} else {
+			if err := crutils.CRImportCheckpointWithoutConfig(ctr.Dir(), ctr.RestoreArchive()); err != nil {
+				return "", err
+			}
 		}
 		if err := c.restoreFileSystemChanges(ctr, mountPoint); err != nil {
 			return "", err
--- a/server/container_restore.go
+++ b/server/container_restore.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	metadata "github.com/checkpoint-restore/checkpointctl/lib"
+	istorage "github.com/containers/image/v5/storage"
 	"github.com/containers/podman/v4/pkg/annotations"
 	"github.com/containers/podman/v4/pkg/errorhandling"
 	"github.com/containers/storage/pkg/archive"
@@ -13,12 +14,46 @@ import (
 	"github.com/cri-o/cri-o/internal/lib"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/log"
+	crioann "github.com/cri-o/cri-o/pkg/annotations"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 	kubetypes "k8s.io/kubernetes/pkg/kubelet/types"
 )
+
+func (s *Server) checkIfCheckpointOCIImage(ctx context.Context, input string) (bool, error) {
+	if _, err := os.Stat(input); err == nil {
+		return false, nil
+	}
+	imageStatusRespone, err := s.ImageStatus(
+		ctx,
+		&types.ImageStatusRequest{
+			Image: &types.ImageSpec{
+				Image: input,
+			},
+		},
+	)
+	if err != nil {
+		return false, err
+	}
+
+	if imageStatusRespone == nil ||
+		imageStatusRespone.Image == nil ||
+		imageStatusRespone.Image.Spec == nil ||
+		imageStatusRespone.Image.Spec.Annotations == nil {
+		return false, nil
+	}
+
+	ann, ok := imageStatusRespone.Image.Spec.Annotations[crioann.CheckpointAnnotationName]
+	if !ok {
+		return false, nil
+	}
+
+	logrus.Debugf("Found checkpoint of container %v in %v", ann, input)
+
+	return true, nil
+}
 
 // taken from Podman
 func (s *Server) CRImportCheckpoint(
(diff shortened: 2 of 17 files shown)


adrianreber:
Ready for review. The three CI failures seem to happen everywhere right now and unrelated to this PR.

internal/lib/checkpoint.go
```diff
@@ -54,8 +81,67 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
 		return "", fmt.Errorf("failed to checkpoint container %s: %w", ctr.ID(), err)
 	}
 	if opts.TargetFile != "" {
-		if err := c.exportCheckpoint(ctr, specgen.Config, opts.TargetFile); err != nil {
-			return "", fmt.Errorf("failed to write file system changes of container %s: %w", ctr.ID(), err)
+		logrus.Debugf("Trying to save the checkpoint as %s\n", opts.TargetFile)
```
saschagrunert, line 84:
```suggestion
		logrus.Debugf("Trying to save the checkpoint as %s", opts.TargetFile)
```

internal/lib/checkpoint.go
```diff
@@ -54,8 +81,67 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
...
+		if strings.HasPrefix(opts.TargetFile, "/") || !strings.Contains(opts.TargetFile, "/") {
```
saschagrunert, line 85: Is this check for verifying that the path is absolute?

internal/lib/checkpoint.go
```diff
@@ -54,8 +81,67 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
...
+					FromImage: "scratch",
+				},
+			)
+			if err != nil {
+				return "", err
+			}
+			// Clean-up buildah working container
+			defer func() {
+				if err := importBuilder.Delete(); err != nil {
+					logrus.Errorf("Could not clean-up buildah working container: %q", err)
```
saschagrunert, line 109:
```suggestion
					logrus.Errorf("Could not clean-up buildah working container: %v", err)
```

internal/lib/restore.go
```diff
@@ -63,8 +64,44 @@ func (c *ContainerServer) ContainerRestore(ctx context.Context, opts *ContainerC
 	}
 
 	if ctr.RestoreArchive() != "" {
-		if err := crutils.CRImportCheckpointWithoutConfig(ctr.Dir(), ctr.RestoreArchive()); err != nil {
-			return "", err
+		if ctr.RestoreIsOCIImage() {
+			logrus.Debugf("Restoring from %v\n", ctr.RestoreArchive())
```
saschagrunert, line 68:
```suggestion
			logrus.Debugf("Restoring from %v", ctr.RestoreArchive())
```

internal/lib/restore.go
```diff
@@ -63,8 +64,44 @@ func (c *ContainerServer) ContainerRestore(ctx context.Context, opts *ContainerC
...
+			imageMountPoint, err := c.StorageImageServer().GetStore().MountImage(ctr.RestoreArchive(), nil, "")
+			if err != nil {
+				return "", err
+			}
+			logrus.Debugf("Checkpoint image mounted at %v\n", imageMountPoint)
```
saschagrunert, line 73:
```suggestion
			logrus.Debugf("Checkpoint image mounted at %v", imageMountPoint)
```

internal/lib/restore.go
```diff
@@ -63,8 +64,44 @@ func (c *ContainerServer) ContainerRestore(ctx context.Context, opts *ContainerC
...
+				}
+			}()
+
+			// Import all checkpoint files except ConfigDumpFile and SpecDumpFile. We
+			// generate new container config files to enable to specifying a new
+			// container name.
+			checkpoint := []string{
+				"artifacts",
+				metadata.CheckpointDirectory,
+				// metadata.CheckpointVolumesDirectory,
```
saschagrunert, line 87: Is this intentional? Do we wanna left a TODO if so?

pkg/annotations/annotations.go
```diff
@@ -44,6 +44,35 @@ const (
...
+	// to create the container.
+	CheckpointAnnotationRootfsImageID = "io.kubernetes.cri-o.annotations.checkpoint.rootfsImageID"
+
+	// CheckpointAnnotationRootfsImageName is used by Container Checkpoint when
+	// creating a checkpoint image to specify the original image name used to
+	// create the container.
+	CheckpointAnnotationRootfsImageName = "io.kubernetes.cri-o.annotations.checkpoint.rootfsImageName"
+
+	// CheckpointAnnotationCRIOVersion is used by Container Checkpoint when
+	// creating a checkpoint image to specify the version of Podman used on the
```
saschagrunert, line 68:
```suggestion
	// creating a checkpoint image to specify the version of CRI-O used on the
```
Is that correct?

server/container_restore.go
```diff
@@ -6,69 +6,137 @@ import (
...
+		imageStatusRespone.Image.Spec.Annotations == nil {
+		return false, nil
+	}
+
+	ann, ok := imageStatusRespone.Image.Spec.Annotations[crioann.CheckpointAnnotationName]
+	if !ok {
+		return false, nil
+	}
+
+	logrus.Debugf("Found checkpoint of container %v in %v\n", ann, input)
```
saschagrunert, line 53:
```suggestion
	logrus.Debugf("Found checkpoint of container %v in %v", ann, input)
```

internal/lib/restore.go
adrianreber: That was a leftover. I removed it.

internal/lib/checkpoint.go
adrianreber: I added a long comment what this does. It is basically a workaround for the limited choices the CRI API gives us.

internal/lib/checkpoint.go
```diff
@@ -54,8 +81,86 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
...
+		// As the crio.sock is only accessible by root and CRIU support currently
+		// needs to be enable explicitly in CRI-O only root is able to trigger
+		// a checkpoint request.
+		if strings.HasPrefix(opts.TargetFile, "/") || !strings.Contains(opts.TargetFile, "/") {
+			if err := c.exportCheckpoint(ctr, specgen.Config, opts.TargetFile); err != nil {
+				return "", fmt.Errorf("failed to write file system changes of container %s: %w", ctr.ID(), err)
+			}
+		} else {
+			// Create a checkpoint image in the local image store
+			imageRef, err := istorage.Transport.ParseStoreReference(c.store, opts.TargetFile)
```
haircommander, line 110: nit: can this block be a separate function?

pkg/annotations/annotations.go
```diff
@@ -44,6 +44,35 @@ const (
 
 	// CPUFreqGovernorAnnotation sets the cpufreq governor for CPUs used by the container
 	CPUFreqGovernorAnnotation = "cpu-freq-governor.crio.io"
+
```
haircommander, line 47: I feel like we should create a new file for these

internal/lib/checkpoint.go
adrianreber: done

pkg/annotations/annotations.go
adrianreber: done

adrianreber:
Not sure if the test failures are related to this PR or not. Doesn't look like it and I see the conmon-rs test also failing in other PRs.

internal/lib/checkpoint.go
```diff
@@ -26,6 +32,27 @@ type ContainerCheckpointRestoreOptions struct {
 	libpod.ContainerCheckpointOptions
 }
 
+func addCheckpointImageMetadata(importBuilder *buildah.Builder, ctr *oci.Container) error {
```
haircommander, line 35: nit: can this function be defined below where it's last called?

adrianreber: sure, done

adrianreber:
I addressed all review comments. Ready for further reviews. The CI errors seem unrelated to my PR.

haircommander:
a hesitation I have here is this goes beyond the requirements of the CRI. while this is a neat direction to take this feature, part of me feels like this could be achieved if we checkpoint the file, then use buildah directly on the cli to create the image/artifact to push. since we already need podman to push, it's not like we can do it all self contained within crio. having a clear separation of concerns (crio checkpointing because it owns the containers, other tools doing the sharing distrobution) sounds more in line with what we've been doing. wdyt @mrunalp @saschagrunert

haircommander:
> You are right, but I would hope/prefer to have it still as part of CRI-O.

are there plans to do the same in containerd? Or codify the behavior in the checkpoint KEP?

adrianreber:
> > You are right, but I would hope/prefer to have it still as part of CRI-O.
> 
> are there plans to do the same in containerd? Or codify the behavior in the checkpoint KEP?

The whole idea of checkpoint artefacts in an OCI image is based on sig-node discussions with input from the containerd developers. containerd has support for such images for a much longer time. They do not have it wired through from the CRI (my PR is still pending) but the whole idea is based on containerd. Unfortunately we were not able to use the format containerd uses for Podman/CRI-O as containerd includes a couple of internal protobuf structures which do not make sense to have outside of containerd. My goal is at some point to have discussions around standardization of the checkpoint image format based on JSON, basically what we use in Podman and in this PR. For now the checkpoint OCI images between containerd and Podman/CRI-O are not compatible but at this point it is just about getting the metadata right. It feels to earlier to have a standard as we do not know yet exactly what metadata is necessary.

adrianreber:
@haircommander Should I remove the code to create the checkpoint OCI image or can it be left as part of CRI-O?

haircommander:
I personally would prefer to wait until we standardize in the KEP. WDYT @saschagrunert @mrunalp

adrianreber:
> I personally would prefer to wait until we standardize in the KEP. WDYT @saschagrunert @mrunalp

Hmm. Okay. There is no immediate plan from my side to add this to a KEP as the image specification would be independent of Kubernetes from my point of view. If a standardization happens it would happen at https://github.com/opencontainers/image-spec , but that is a step we did not plan in the next couple of months.

Currently is is an image format used by Podman and specific to Podman and the goal was to bring this to CRI-O.

Not having it in CRI-O is not something we really planed for. Not really sure what to do now. Happy for any additional input.

adrianreber:
@haircommander What about just loading checkpoint images as created by Podman and not creating them by CRI-O? Just the changes in the restore code path?

saschagrunert:
I see both of your points @haircommander @adrianreber. The question is if we value the separation of concerns over usability. Since we just wrap a bunch of buildah APIs in an isolated fashion I'm leaning towards having the feature complete in CRI-O.

haircommander:
> @haircommander What about just loading checkpoint images as created by Podman and not creating them by CRI-O? Just the changes in the restore code path?

I think this is a good compromise and works for me. How does that sound @saschagrunert ?

adrianreber:
Okay, let me rework this PR and ping you again. Thanks.

adrianreber:
Okay, let's see if CI like the changes.

adrianreber:
Is this good to go?

haircommander:
/approve

LGTM, thanks @adrianreber 
@saschagrunert @mrunalp PTAL

saschagrunert (approved):
/lgtm
/retest

PR #6192, opened 2022-08-31
config: avoid segfault when workloads.resources is nil

haircommander:

#### What type of PR is this?

/kind bug
fixes https://github.com/cri-o/cri-o/issues/6188

```release-note
fix a segfault in `crio config` when `runtime.workloads.resources` is nil 
```

### DIFF
--- a/test/config.bats
+++ b/test/config.bats
@@ -96,3 +96,16 @@ EOF
 	[[ "$RES" == *"monitor_cgroup = \"pod\""* ]]
 	[[ "$RES" == *"monitor_path = \"/bin/true\""* ]]
 }
+
+@test "handle nil workloads" {
+	# when
+	unset CONTAINER_DEFAULT_RUNTIME
+	cat << EOF >> "$TESTDIR"/workload.conf
+[crio.runtime.workloads.userns]
+activation_annotation = "io.kubernetes.cri-o.userns-mode"
+allowed_annotations = ["io.kubernetes.cri-o.userns-mode"]
+EOF
+
+	# then
+	"$CRIO_BINARY_PATH" -c "$TESTDIR"/workload.conf -d "" config
+}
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):

test/config.bats
```diff
@@ -65,3 +65,15 @@ function teardown() {
...
+@test "handle nil workloads" {
+	# when
+	cat << EOF >> "$TESTDIR"/workload.conf
+[crio.runtime.workloads.userns]
+activation_annotation = "io.kubernetes.cri-o.userns-mode"
+allowed_annotations = ["io.kubernetes.cri-o.userns-mode"]
+EOF
+
+	# then
+	"$CRIO_BINARY_PATH" -c "$TESTDIR"/workload.conf -d "" config
```
littlejawa, line 78: kata-jenkins job fails here, because it cannot find the definition for the default runtime we've set.

`# level=fatal msg="validating runtime config: default_runtime set to \"containerd-shim-kata-v2\", but no runtime entry table [crio.runtime.runtimes.containerd-shim-kata-v2] was found"`

Maybe need to customize the conf file you use with an entry for the default runtime?

haircommander:
@saschagrunert PTAL

test/config.bats
littlejawa: Another (maybe simpler) suggestion: if the test doesn't imply the runtime, maybe you can unset the CONTAINER_RUNTIMES env variable for this test, like it's done in another test above?

haircommander: good idea! trying that

saschagrunert (approved):
/lgtm

test/config.bats
littlejawa: Ok, it didn't help :-(
I just tried locally. The problem is the default runtime, not the list of runtimes.
You need to unset CONTAINER_DEFAULT_RUNTIME for it to work.
I think it's ok to do that in this context, as the runtime is not involved - we're testing the config engine.

test/config.bats
```diff
@@ -96,3 +96,16 @@ EOF
 	[[ "$RES" == *"monitor_cgroup = \"pod\""* ]]
 	[[ "$RES" == *"monitor_path = \"/bin/true\""* ]]
 }
+
+@test "handle nil workloads" {
+	# when
+	unset CONTAINER_RUNTIMES
```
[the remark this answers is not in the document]
haircommander: thanks! adapted as suggested

haircommander:
/retest
@saschagrunert PTAL

saschagrunert (approved):
/lgtm

PR #6193, opened 2022-08-31
Adds rhel9 variant-specific changes to CI

sohankunkerkar:
Signed-off-by: Sohan Kunkerkar <sohank2602@gmail.com>

#### What type of PR is this?

/kind ci 

#### What this PR does / why we need it:
adds rhel9 variant-specific changes to CI
#### Does this PR introduce a user-facing change?

```release-note
None

```

PR #6196, opened 2022-08-31
cgmgr: use NewSystemd from createSandboxCgroup

kolyshkin:
It has been reported that sometimes pod creation fails with an
error like this one:

> cgroup: error creating cgroup path /pod_123.slice/pod_123-456.slice/crio-0dd29e75c4072e6d8227a338500c5a9a0cae2b41215c136150640c21e3e07fdf.scope: write /sys/fs/cgroup/pod_123.slice/pod_123-456.slice/cgroup.subtree_control: no such file or directory"

The error comes from containers/common/pkg/cgroups.New, which
eventually calls createCgroupv2Path, which does Mkdir immediately
followed by WriteFile, which fails with ENOENT.

It seems that this is caused by systemd which seems an unknown cgroup
hierarchy and removes it (in this case, in between Mkdir and WriteFile).

The solution is to not create paths when using systemd driver, since
systemd is going to create those for us.

Use cgroups.NewSystemd when appropriate.

#### What type of PR is this?

/kind bug
/kind flake

See above.

None

None

```release-note
NONE
```

### DIFF
--- a/internal/config/cgmgr/cgmgr.go
+++ b/internal/config/cgmgr/cgmgr.go
@@ -147,7 +147,11 @@ func createSandboxCgroup(sbParent, containerID string, mgr CgroupManager) error 
 	if err != nil {
 		return err
 	}
-	_, err = cgroups.New(path, &cgcfgs.Resources{})
+	if mgr.IsSystemd() {
+		_, err = cgroups.NewSystemd(path, &cgcfgs.Resources{})
+	} else {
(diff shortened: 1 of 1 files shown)


kolyshkin:
CI failures (2 of the same failures in 3 different tests) appear to be unrelated (they are the same in any recent PR, e.g. https://github.com/cri-o/cri-o/pull/6193)

saschagrunert (approved):
LGTM

PR #6236, opened 2022-09-19
add "add_inheritable_capabilities" option

haircommander:
#### What type of PR is this?

/kind feature
add an option to re-add the inheritable capabilities. there are situations where users previously expected non-root users to get these capabilities, which adding them to the inheritable list enabled. 

related to https://github.com/kubernetes/kubernetes/issues/111196 and https://github.com/cri-o/cri-o/security/advisories/GHSA-4hj2-r2pm-3hc6

```release-note
Add an option "add_inheritable_capabilities" which adds added capabilities to the inheritable list as well. As of CRI-O 1.24.0, CRI-O drops the inheritable capabilities to fix CVE-2022-27652 . However, this can cause regressions in workloads that attempt to pass capabilities to non-root users through inheritable capabilities. 
```

### DIFF
--- a/completions/bash/crio
+++ b/completions/bash/crio
@@ -13,6 +13,7 @@ wipe
 help
 h
 --absent-mount-sources-to-reject
+--add-inheritable-capabilities
 --additional-devices
 --address
 --allowed-devices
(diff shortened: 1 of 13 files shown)


saschagrunert (approved):
LGTM

PR #6220, opened 2022-09-12
Bump conmon-rs code to latest `main`

saschagrunert:
#### What type of PR is this?

/kind dependency-change

#### What this PR does / why we need it:
Update the latest conmon-rs golang client code.

#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
Requires https://github.com/containers/conmon-rs/pull/702 to make the CI pass again.
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6237, opened 2022-09-19
use cri-tools version from dependencies.yaml

sohankunkerkar:
Signed-off-by: Sohan Kunkerkar <sohank2602@gmail.com>

#### What type of PR is this?

/kind cleanup

```release-note
None

```


haircommander:
/approve
/lgtm

thanks!

contrib/test/ci/setup.yml
```diff
@@ -5,7 +5,7 @@
 - name: install Golang tools
   include: "golang.yml"
   vars:
-    version: "1.19"
+    golang_version: "1.19"
```
saschagrunert, line 8: I don't think we need this since we use `{{ golang_version }}` directly in golang.yml.

sohankunkerkar: Yup, good catch!

contrib/test/ci/main.yml
```diff
@@ -1,195 +0,0 @@
----
```
saschagrunert, line 1: What's the reason that we removed that file?

sohankunkerkar: This file has not been used for running any of our new CI jobs.

sohankunkerkar:
There's a [PR](https://github.com/cri-o/cri-o/pull/6220) to fix the `e2e-conomn-rs` job

PR #6248, opened 2022-09-22
Allow runtime configs to be updated completely

elezar:
#### What type of PR is this?

/kind feature

This feature extends the work done for #6056 to allow for complete updates of the runtime configuration. The previous work limited the changes to the runtimes config to adding runtimes and updating the default runtime. These changes allow the complete `crio.runtime.runtimes` table to be updated on reload.

This was originally discussed in #6036.

None

```release-note
Allow `crio.runtime.runtimes` configuration to be updated on reload.
```

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -291,7 +291,7 @@ the container runtime configuration.
   Enable CRIU integration, requires that the criu binary is available in $PATH. (default: false)
 
 ### CRIO.RUNTIME.RUNTIMES TABLE
-The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  The runtime to use is picked based on the runtime handler provided by the CRI.  If no runtime handler is provided, the runtime will be picked based on the level of trust of the workload. This option supports live configuration reload. Note that on reload runtimes can only be added to this table but not modified or removed.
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -1479,3 +1538,24 @@ func (c *NetworkConfig) CNIManagerShutdown() {
 func (c *Config) SetSingleConfigPath(singleConfigPath string) {
 	c.singleConfigPath = singleConfigPath
 }
+
+// Equals checks whether to runtimes tables are equal.
+func (r Runtimes) Equals(o Runtimes) bool {
```
haircommander, line 1543: we could use RuntimesEqual in pkg/config/template.go instead

elezar: Thanks. I have switched to this. One thing that I was concerned about was to ignore ordering in the `RuntimeHandler` members that are slices but this may be over engineering things at this stage.

Note that I call `RuntimesEqual` in `reload.go` directly instead of wrapping it here.

haircommander: worst case, we get a false positive because the members are in a different order, and we reload the config. we'll spend some unnecessary cpu time, but it doesn't seem like a horrible case. I think the code succinct-ness is worth it

haircommander:
/lgtm
/approve
thank you @elezar !

PR #6255, opened 2022-09-26
config: translate monitor fields when printing config

haircommander:
A situation arises when printing the config where
`translateMonitorFields` didn't previously run. This causes situations where the `conmon_cgroup` field looks like it's some other value, but the corresponding `monitor_cgroup` field does not reflect it

note: this is only a cosmetic change. The underlying config does eventually render correctly, just didn't print right.

#### What type of PR is this?

/kind bug

```release-note
fix a bug where `conmon_cgroup` and `monitor_path` became out of sync
```

### DIFF
--- a/test/config.bats
+++ b/test/config.bats
@@ -84,3 +84,15 @@ EOF
 	[[ "$RES" == *"crio.runtime.runtimes.runc"* ]]
 	[[ "$RES" == *"crio.runtime.runtimes.crun"* ]]
 }
+
+@test "monitor fields should be translated" {
+	if [[ "$RUNTIME_TYPE" == "vm" ]]; then
+		skip "not applicable to vm runtime type"
+	fi
+	# when
+	RES=$("$CRIO_BINARY_PATH" --conmon-cgroup="pod" --conmon="/bin/true" -c "" -d "" config 2>&1)
+
+	# then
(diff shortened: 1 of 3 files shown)


test/config.bats
```diff
@@ -65,3 +65,12 @@ function teardown() {
 	[[ "$RES" == *"crio.runtime.runtimes.runc"* ]]
 	[[ "$RES" == *"crio.runtime.runtimes.crun"* ]]
 }
+
+@test "monitor fields should be translated" {
+	# when
+	RES=$("$CRIO_BINARY_PATH" --conmon-cgroup="pod" --conmon="/bin/true" -c "" -d "" config 2>&1)
```
littlejawa, line 71: This command is failing in the kata CI because the default runtime is set to a value that is not found in the default conf (using CONTAINER_DEFAULT_RUNTIME env var).
Also the error message is not shown in the CI logs because stderr is redirected.

If this test is only about conmon options, maybe we can test for "RUNTIME_TYPE=vm" and skip it in that case? Conmon is not used with runtime_vm.

haircommander: makes sense, thank you!

mrunalp (approved):
/lgtm

haircommander:
@saschagrunert @mrunalp I had to fix some tests, PTAL

saschagrunert (approved):
/lgtm

PR #3772, opened 2020-05-15
Do not remove existing runtime handler

harche:
Fixes - #3536

Signed-off-by: Harshal Patil <harpatil@redhat.com>

#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

This PR addresses the concerns raised in #3536 about having runtimes other than default `runc` be responsible for making sure the existing `runc` configuration is not broken. For example, please refer to #3536 for additional info on how it affects the users of Kata runtime. 

#### Which issue(s) this PR fixes:

Fixes #3536 

#### Does this PR introduce a user-facing change?
Yes

```release-note
Adding additional runtime handler doesn't require the user to copy existing default runtime handler configuration. The existing default runtime handler configuration will be preserved while adding the new runtime handler. 
```

PR #6257, opened 2022-09-27
config: do not remove runc if different default runtime

haircommander:
This commit reverts d2ded1d73c9ce282028ec2b4d6016546267322a9 and adds a test that maintains what it was trying to do.

When upgrading from a cluster that uses runc to one that uses crun, there may be workloads that specifically request the old default. Those workloads will fail on restart. To fix this, don't remove runc from the map. We've since added code that means we don't fail if one of the non-default runtimes fails validation (if the binary doesn't exist) so we can handle that situation, which
d2ded1d73c9ce282028ec2b4d6016546267322a9 meant to fix.

#### What type of PR is this?

/kind bug

```release-note
Fix a bug where updating `default_runtime` would cause the `runc` entry in the runtimes table to be deleted
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -645,15 +645,9 @@ func (c *Config) UpdateFromDropInFile(path string) error {
 	t := new(tomlConfig)
 	t.fromConfig(c)
 
-	metadata, err := toml.Decode(string(data), t)
+	_, err = toml.Decode(string(data), t)
 	if err != nil {
 		return fmt.Errorf("unable to decode configuration %v: %w", path, err)
-	}
-
-	runtimesKey := []string{"crio", "runtime", "default_runtime"}
-	if metadata.IsDefined(runtimesKey...) &&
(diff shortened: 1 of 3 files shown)


haircommander:
@harche we return again, PTAL, lemme know how you feel about this

haircommander:
also @cynepco3hahue because you chimed in [here](https://github.com/cri-o/cri-o/pull/3772#issuecomment-650990717)

haircommander:
@littlejawa should I wait to merge this till that one does, or should I override this because that relies on this?

littlejawa:
> @littlejawa should I wait to merge this till that one does, or should I override this because that relies on this?

@haircommander - the problem if we merge this PR is that all subsequent PRs targetting the main branch will fail because of these tests :-(
I will ping people and try to get it merged ASAP.
Sorry for the delay

haircommander:
thank you @littlejawa 

/cherry-pick release-1.25

PR #6258, opened 2022-09-28
contrib/test/ci: set cgroup_config for fedora

sohankunkerkar:
This is needed to run the kubernetes cluster without any hiccups on Fedora 36

Signed-off-by: Sohan Kunkerkar <sohank2602@gmail.com>

#### What type of PR is this?

/kind ci

```release-note
None
```

### DIFF
--- a/contrib/test/ci/build/kubernetes.yml
+++ b/contrib/test/ci/build/kubernetes.yml
@@ -82,6 +82,15 @@
       ./hack/local-up-cluster.sh
     mode: "u=rwx,g=rwx,o=x"
 
+- name: Set CGROUP_CONFIG for fedora
+  become: yes
+  lineinfile:
+    dest: /usr/local/bin/createcluster.sh
+    line: export CGROUP_CONFIG=systemd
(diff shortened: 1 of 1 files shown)


contrib/test/ci/build/kubernetes.yml
```diff
@@ -82,6 +82,15 @@
       ./hack/local-up-cluster.sh
     mode: "u=rwx,g=rwx,o=x"
 
+- name: Set CGROUP_CONFIG for fedora
+  become: yes
+  lineinfile:
+    dest: /usr/local/bin/createcluster.sh
```
mrunalp, line 88: Why can't we put this inside the createcluster.sh script?

sohankunkerkar: I don't think we're hitting this issue for other variants, so tried the safer approach.

haircommander:
@sohankunkerkar do we still need this?

sohankunkerkar:
I don't think we need this anymore

PR #6273, opened 2022-10-05
fix: give loopback a name

STRRL:
Signed-off-by: STRRL <im@strrl.dev>

#### What type of PR is this?

/kind bug

- from CNI spec, https://github.com/containernetworking/cni/blob/main/SPEC.md, the field `name` should be required. And there is also a logic for validating it:

https://github.com/containernetworking/cni/blob/8dba3824fbfdf9fe69e72cd2f2330c06303e2a84/pkg/skel/skel.go#L184-L186

None

```release-note
None 
```

### DIFF
--- a/contrib/cni/99-loopback.conf
+++ b/contrib/cni/99-loopback.conf
@@ -1,4 +1,5 @@
 {
     "cniVersion": "0.3.1",
+    "name": "loopback",
     "type": "loopback"
 }


dilyanpalauzov:
According to https://github.com/containernetworking/plugins/issues/759 CRI-O does not use the loopback-cni-plugin.

haircommander:
/retest

> According to https://github.com/containernetworking/plugins/issues/759 CRI-O does not use the loopback-cni-plugin.

We don't use the plugin, but we do use the config to tell us to use the internal loopback instead of some other CNI plugin

dilyanpalauzov:
> We don't use the plugin, but we do use the config to tell us to use the internal loopback instead of some other CNI plugin

I do not understand this.  CRI-O does not use the undocumented, deprecated loopback plugin, but it requires configuration file for that plugin.

haircommander:
> > We don't use the plugin, but we do use the config to tell us to use the internal loopback instead of some other CNI plugin
> 
> I do not understand this. CRI-O does not use the undocumented, deprecated loopback plugin, but it requires configuration file for that plugin.

I need to double check, but I believe so. If we don't use the existence of the configuration file to say to use loopback, I don't know how we would differentiate between "CNI plugin pod hasn't come up yet empty configuration directory" and "Intentionally empty configuration directory because the internal loopback plugin is used"

haircommander:
/override ci/prow/e2e-agnostic
/lgtm
/approve

thanks @STRRL

STRRL:
hi @haircommander, it seems the CI got stuck with `ci/kata-jenkins` with a long-time `Pending` status.

And `tide` also got stuck with `Retesting 16 jobs`.

meierra:
Seems this fix was overwritten by another change
[Fix-6080: Update the CNI version to 1.0.0](https://github.com/cri-o/cri-o/commit/f317b267ddef21aee5ffc92d890a77112b006815)
f317b267ddef21aee5ffc92d890a77112b006815

So the source for 1.26.1 is back with the name missing from loopback. Can we reopen this request?

meierra:
Actually seems that :

contrib/cni/99-loopback.conf 
           was replaced by
contrib/cni/99-loopback.conflist

Which reverted the changes, wasn't picked up as .conf was deleted and there was no diff to flag

haircommander:
I've asked @saschagrunert to do so in https://github.com/cri-o/cri-o/pull/6541

PR #6280, opened 2022-10-06
test: config tests should ignore default runtime environment variable

littlejawa:
#### What type of PR is this?

/kind ci

Config tests are meant to verify the configuration code, by exercising it in specific configurations.
When run in an environment where the default runtime is modified, these tests fail because the configuration they use do not have this runtime defined.
As the runtime itself is not involved in these tests, it seems safe to unset the CONTAINER_DEFAULT_RUNTIME variable in their context.

None

```release-note
None
```

### DIFF
--- a/contrib/test/ci/integration.yml
+++ b/contrib/test/ci/integration.yml
@@ -54,7 +54,6 @@
         state: present
       loop: "{{ ['cgroups.bats'] | product(kata_skip_cgroups_tests) \
         + ['command.bats'] | product(kata_skip_command_tests) \
-        + ['config.bats'] | product(kata_skip_config_tests) \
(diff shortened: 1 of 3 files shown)


littlejawa:
@haircommander - could you take a look? This is generalizing what we did for your PR #6192

littlejawa:
> any reason not to do this in `setup()` instead?

I didn't want to go that far, because some of the tests do not need it : it makes sense only for tests that plays with the default config.

haircommander:
you've got a shfmt error @littlejawa

littlejawa:
@haircommander - this one is pretty old and low priority, but I think it still makes sense.
What do you think?

saschagrunert (approved):

saschagrunert (approved):

PR #6294, opened 2022-10-11
Add logs to OpenTelemetry traces

saschagrunert:
#### What type of PR is this?

/kind feature

We now use the log interceptor to enrich the otel traces with additional data. For example we now have detailed logging information from CRI-O and propagate the error to the traces.

None

For example, the output now looks like this on container creation:

screenshot

And the error propagation looks like this:

screenshot

Side note: We need the additional span to capture most of the logs, otherwise we will miss some information (see the error propagation).

cc @sallyom 

```release-note
Added logs and GRPC error codes to OpenTelemetry traces.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,6 +65,7 @@ require (
 	go.opentelemetry.io/otel v1.10.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.10.0
 	go.opentelemetry.io/otel/sdk v1.10.0
+	go.opentelemetry.io/otel/trace v1.10.0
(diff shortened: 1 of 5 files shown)


sallyom (approved):
nice! AFIAK there are no definite guidelines for trace content - and attributes to correlate logs will be useful.

saschagrunert:
> nice! AFIAK there are no definite guidelines for trace content - and attributes to correlate logs will be useful.

Thank you! PTAL @haircommander

saschagrunert:
/override ci/prow/ci-cgroupv2-e2e

Seems to be stuck in merge pool :thinking:

PR #6322, opened 2022-10-18
Add `--tracing-sample-always`/`tracing_sample_always` option

saschagrunert:
#### What type of PR is this?

/kind feature

This allows to enable tracing as well as sampling for each RPC independently of the sample rate as well as any kubelet configuration.

None
None

```release-note
Added `--tracing-sample-always`/`tracing_sample_always` option to enable tracing independently of the sample rate as well as any kubelet configuration.
```


saschagrunert:
/test e2e-agnostic

@haircommander PTAL

internal/opentelemetry/tracing_config.go
```diff
@@ -44,8 +44,12 @@ func InitTracing(ctx context.Context, collectorAddress string, samplingRate int)
 
 	// Only emit spans when the kubelet sends a request with a sampled trace
 	sampler := sdktrace.NeverSample()
-	// Or, emit spans for a fraction of transactions
-	if samplingRate > 0 {
+
+	if sampleAlways {
+		// Or, always if set
+		sampler = sdktrace.AlwaysSample()
```
haircommander, line 50: is this different than specifying 1000000 for samplingRate?

saschagrunert: No, since it translates to that if the faction is `>=1:` https://github.com/cri-o/cri-o/blob/557f22901941f79598f2cb9ffdfe78552f6dda30/vendor/go.opentelemetry.io/otel/sdk/trace/sampling.go#L107-L110

From a usability perspective I think that's a bit odd. But do you think we could just document that users have to set `--tracing-sampling-rate-per-million=1000000` for always sampling?

haircommander: yeah I'm hesitant to add a new config field if the existing one can satisfy

saschagrunert: Alright, sounds good to me :+1:

PR #6339, opened 2022-10-26
Disable checkpoint image check as early as possible

adrianreber:
#### What type of PR is this?

/kind other

To see if the specified container image is a checkpoint image, each image is checked, if it is a checkpoint image during container create. If checkpoint support is not enabled in CRI-O return as early as possible from this check without even looking at the image.

None

```release-note
None
```


saschagrunert (approved):
/ok-to-test
/lgtm

PR #6343, opened 2022-10-28
Extend distributed tracing

vrutkovs:
#### What type of PR is this?
/kind feature

This PR extends distributed tracing framework we already have in place:
* Ensures logrus messages are attached as events to trace via hook (instead of using custom function)
* Starts span when most functions are called and stops it when it ends, adds a handy function to `internal/log/log.go`
* Updates most functions to pass context to child functions to preserve parent trace reference (does this qualify as api change?)

image
image

No

```release-note
More information available in tracing spans
```

### DIFF
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -75,7 +75,7 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
 		for _, del := range cleanup {
 			file := filepath.Join(ctr.Dir(), del)
 			if err := os.Remove(file); err != nil {
-				logrus.Debugf("Unable to remove file %s", file)
+				logrus.WithContext(ctx).Debugf("Unable to remove file %s", file)
 			}
 		}
 	}
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -13,10 +13,10 @@ import (
 	"github.com/containers/podman/v4/pkg/annotations"
 	"github.com/containers/podman/v4/pkg/checkpoint/crutils"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
-	"github.com/sirupsen/logrus"
 )
 
 type ContainerCheckpointRestoreOptions struct {
@@ -62,7 +62,7 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
 		return "", fmt.Errorf("failed to unmount container %s: %w", ctr.ID(), err)
 	}
 	if err := c.ContainerStateToDisk(ctx, ctr); err != nil {
-		logrus.WithContext(ctx).Warnf("Unable to write containers %s state to disk: %v", ctr.ID(), err)
+		log.Warnf(ctx, "Unable to write containers %s state to disk: %v", ctr.ID(), err)
 	}
 
 	if !opts.Keep {
@@ -75,7 +75,7 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
 		for _, del := range cleanup {
 			file := filepath.Join(ctr.Dir(), del)
 			if err := os.Remove(file); err != nil {
-				logrus.WithContext(ctx).Debugf("Unable to remove file %s", file)
+				log.Debugf(ctx, "Unable to remove file %s", file)
 			}
 		}
 	}
@@ -220,7 +220,7 @@ func (c *ContainerServer) prepareCheckpointExport(ctr *oci.Container) error {
 func (c *ContainerServer) exportCheckpoint(ctx context.Context, ctr *oci.Container, specgen *rspec.Spec, export string) error {
 	id := ctr.ID()
 	dest := ctr.Dir()
-	logrus.WithContext(ctx).Debugf("Exporting checkpoint image of container %q to %q", id, dest)
+	log.Debugf(ctx, "Exporting checkpoint image of container %q to %q", id, dest)
 
 	includeFiles := []string{
 		stats.StatsDump,
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -28,7 +28,7 @@ type ContainerCheckpointRestoreOptions struct {
 
 // ContainerCheckpoint checkpoints a running container.
 func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *ContainerCheckpointRestoreOptions) (string, error) {
-	ctr, err := c.LookupContainer(opts.Container)
+	ctr, err := c.LookupContainer(ctx, opts.Container)
 	if err != nil {
 		return "", fmt.Errorf("failed to find container %s: %w", opts.Container, err)
 	}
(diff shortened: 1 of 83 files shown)


saschagrunert:
Thank you so much for this @vrutkovs!

/ok-to-test

internal/lib/checkpoint.go
```diff
@@ -217,10 +217,10 @@ func (c *ContainerServer) prepareCheckpointExport(ctr *oci.Container) error {
 	return nil
 }
 
-func (c *ContainerServer) exportCheckpoint(ctr *oci.Container, specgen *rspec.Spec, export string) error {
+func (c *ContainerServer) exportCheckpoint(ctx context.Context, ctr *oci.Container, specgen *rspec.Spec, export string) error {
 	id := ctr.ID()
 	dest := ctr.Dir()
-	logrus.Debugf("Exporting checkpoint image of container %q to %q", id, dest)
+	logrus.WithContext(ctx).Debugf("Exporting checkpoint image of container %q to %q", id, dest)
```
saschagrunert, line 223: If we have a context, do we then use our own log wrapper here?
```suggestion
	log.Debugf(ctx, "Exporting checkpoint image of container %q to %q", id, dest)
```

internal/log/log.go
```diff
@@ -67,8 +47,17 @@ func entry(ctx context.Context) *logrus.Entry {
...
+	return logrus.NewEntry(logger).WithContext(ctx)
+}
+
+func StartSpan(ctx context.Context) (context.Context, trace.Span) {
+	spanName := ""
+	// Use function signature as a span name if available
+	if pc, _, _, ok := runtime.Caller(1); ok {
+		spanName = runtime.FuncForPC(pc).Name()
+	}
+	return trace.SpanFromContext(ctx).TracerProvider().Tracer("").Start(ctx, spanName)
```
saschagrunert, line 62:
```suggestion
	return trace.SpanFromContext(ctx).TracerProvider().Tracer("crio").Start(ctx, spanName)
```

saschagrunert (commented):
I'm wondering if we still need this part:

https://github.com/cri-o/cri-o/blob/c4a9c6d123c9c038b1935a9f3c30e1af98596757/internal/log/interceptors.go#L54-L66

If not, then we could probably also remove: 

https://github.com/cri-o/cri-o/blob/c4a9c6d123c9c038b1935a9f3c30e1af98596757/internal/opentelemetry/tracing_config.go#L18-L24

internal/lib/checkpoint.go
vrutkovs: I'd rather do that in the follow-up PRs, along with many smaller changes

internal/log/log.go
vrutkovs: my understanding was that using `""` would get us default (== `crio`) so that we wouldn't have to hardcode it twice

internal/log/log.go
```diff
@@ -67,8 +47,17 @@ func entry(ctx context.Context) *logrus.Entry {
...
-		return logger.WithField("id", id).WithField("name", name)
+		return logger.WithField("id", id).WithField("name", name).WithContext(ctx)
 	}
 
-	return logrus.NewEntry(logger)
+	return logrus.NewEntry(logger).WithContext(ctx)
+}
+
+func StartSpan(ctx context.Context) (context.Context, trace.Span) {
+	spanName := ""
```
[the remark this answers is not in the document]
vrutkovs: Right, I could not reproduce the situation where `runtime` is unable to fetch a function name too. Lets use `unknown` then maybe?

internal/log/log.go
saschagrunert: Ah alright.

internal/lib/checkpoint.go
saschagrunert: Sounds good!

internal/log/log.go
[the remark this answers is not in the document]
saschagrunert: Yep, sound good! We could also log an error message of `!ok` below, but I'm not sure if that's too much of the verbosity.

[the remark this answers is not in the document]
vrutkovs: Done, using `Debug` for now. Tracing is advanced user / developer feature, so I guess it should be sufficient

vrutkovs:
>I'm wondering if we still need this part:

I think this is covered by `otlptracegrpc` but I'd rather remove it in the follow-up PR

internal/log/log.go
```diff
@@ -67,8 +47,19 @@ func entry(ctx context.Context) *logrus.Entry {
...
+	return logrus.NewEntry(logger).WithContext(ctx)
+}
+
+func StartSpan(ctx context.Context) (context.Context, trace.Span) {
+	spanName := "unknown"
+	// Use function signature as a span name if available
+	if pc, _, _, ok := runtime.Caller(1); ok {
+		spanName = runtime.FuncForPC(pc).Name()
+	} else {
+		Debugf(ctx, "unable to retrieve a caller when starting span")
```
saschagrunert, line 62:
```suggestion
		Debugf(ctx, "Unable to retrieve a caller when starting span")
```

saschagrunert (approved):
just a nit, otherwise LGTM

PTAL @haircommander

internal/lib/checkpoint.go
```diff
@@ -54,15 +54,15 @@ func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *Contain
...
 			return "", fmt.Errorf("failed to write file system changes of container %s: %w", ctr.ID(), err)
 		}
 	}
-	if err := c.storageRuntimeServer.StopContainer(ctr.ID()); err != nil {
+	if err := c.storageRuntimeServer.StopContainer(ctx, ctr.ID()); err != nil {
 		return "", fmt.Errorf("failed to unmount container %s: %w", ctr.ID(), err)
 	}
 	if err := c.ContainerStateToDisk(ctx, ctr); err != nil {
-		logrus.Warnf("Unable to write containers %s state to disk: %v", ctr.ID(), err)
+		logrus.WithContext(ctx).Warnf("Unable to write containers %s state to disk: %v", ctr.ID(), err)
```
haircommander, line 65: since we have ctx we can replace this with the internal log package instead of logrus

internal/lib/container_server.go
```diff
@@ -595,22 +603,26 @@ func (c *ContainerServer) HasContainer(id string) bool {
...
 	sb := c.state.sandboxes.Get(sbID)
 	if sb == nil {
 		return
 	}
-	sb.RemoveContainer(ctr)
+	sb.RemoveContainer(ctx, ctr)
 	c.RemoveStatsForContainer(ctr)
 	if err := ctr.RemoveManagedPIDNamespace(); err != nil {
-		logrus.Errorf("Failed to remove container %s PID namespace: %v", ctr.ID(), err)
+		logrus.WithContext(ctx).Errorf("Failed to remove container %s PID namespace: %v", ctr.ID(), err)
```
haircommander, line 617: same here (and all cases): instead of using WithContext, I think we should just use our internal log package

haircommander:
note to self: we should run some scale tests here to make sure we don't regress, and if we do, we should make sure it's configurable (so we're only collecting and emitting the spans if someone opts into them)

vrutkovs:
>we should run some scale tests here

This is a useful idea, but metrics are more suitable. They answer the question "is everything all right", while tracing helps answering "this particular request is slow and I want to know why"

internal/lib/checkpoint.go
haircommander: I am open either way, but I agree we should do a `port all logrus` commit

haircommander: if we're going to change and then immediately change again, I guess doing it here makes sense

vrutkovs: good point, added a new commit to port these

saschagrunert:
@vrutkovs are you still working on follow-ups of this or have priorities changed?

vrutkovs:
Lets file issues for each remaining piece and I could help with some (as a volunteer)

saschagrunert:
> Lets file issues for each remaining piece and I could help with some (as a volunteer)

Closing the loop: https://github.com/cri-o/cri-o/issues/6419

PR #6347, opened 2022-11-03
Remove CRI v1alpha2 support

saschagrunert:
#### What type of PR is this?

/kind api-change

Kubernetes v1.26.0 will not support CRI v1alpha2 any more after the merge of: https://github.com/kubernetes/kubernetes/pull/110618

This means we can also remove the code parts within CRI-O.

None
None

```release-note
Removed support for CRI `v1alpha2`, means that CRI-O now only supports `v1`.
```


haircommander:
ding dong the witch is removed :tada: 

/lgtm

follow up would be removing the compat layer, do we want to take that on in 1.26 too?

saschagrunert:
> follow up would be removing the compat layer, do we want to take that on in 1.26 too?

I'm wondering if we want to keep it just in case we get a v2 :zipper_mouth_face:

saschagrunert:
/override ci/prow/ci-critest
/override ci/prow/ci-integration

They don't run :thinking:

Issue #6350, opened 2022-11-07
runtime_vm: golang panic: "protobuf tag not enough fields in Status.state" when handling errors from the kata shim


littlejawa:
### What happened?

While testing kata with cri-o 1.25, we found that cri-o was crashing sometimes.
Further investigation shown that this is happening when the kata shim returns an error to cri-o on any calls it makes.

This is the same bug as #3991, but this time it occurs only on errors, not on every call.

Looking into the code, it seems the error was introduced by this commit: https://github.com/cri-o/cri-o/commit/40df9c9ae05e2ea376f65a8b8431a429b1ff112c

It is removing the "replace" directive we had in go.mod to downgrade genproto to a version that doesn't trigger this issue in gogo/protobuf.
The issue was not found immediately because it's not systematic anymore - I'm not sure why.

Downgrading genproto again seems not possible, as opentelemetry requires a higher version of genproto.

### What did you expect to happen?

cri-o should not crash

### How can we reproduce it (as minimally and precisely as possible)?

Modify the kata configuration to make the hypervisor path point to a non-existing binary.
When starting a container with the kata runtime, the kata shim will error out because of the wrong path.

### CRI-O and Kubernetes version

<details>

The issue was found in cri-o v1.25. It can also be seen on main.
It is not visible in 1.24

</details>

### OS version

<details>

</details>

### Additional environment details (AWS, VirtualBox, physical, etc.)

<details>

</details>

littlejawa:
@fidencio - FYI

littlejawa:
Looking at various versions of genproto - the earliest version I find that allows cri-o to compile is v0.0.0-20210713002101-d411969a0d9a. But this is too high already for gogo/protobuf, as the panic happens with it.

littlejawa:
Containerd has moved away from gogo/protobuf in their "main" branch, but the code is not released yet.

https://github.com/containerd/ttrpc/issues/62
https://github.com/containerd/containerd/issues/6564

When it is released (with the upcoming 1.7.0?), and when kata takes it into account, then we can expect this bug to go away for good.

Now until it's available, we may need a temporary solution - at least to avoid the crash in 1.25.

PR #6371, opened 2022-11-10
Fix issue in gogo/protobuf code, causing a crash in cri-o.

littlejawa:
#### What type of PR is this?

/kind bug

Looking at what others did to fix the gogo/protobuf issue, I've found this suggestion:
https://github.com/jaegertracing/jaeger/pull/2857#issuecomment-791704993

The idea is to ignore the fields that were added to the problematic data structure, letting gogo/protobuf handle the ones it knows about.
I replicated this change in cri-o's vendored gogo/protobuf code, and it seems to do the trick - kata containers are still running, and when the kata shim returns an error, it is properly reported in the container's status, without crashing.

I don't think this is a proper fix, but it may be a good workaround for 1.25 and later releases, until we can rely on the containerd version where gogo/protobuf is removed from the shim v2 interface.

Fixes: #6350

```release-note
None
```

### DIFF
--- a/vendor/github.com/gogo/protobuf/proto/table_unmarshal.go
+++ b/vendor/github.com/gogo/protobuf/proto/table_unmarshal.go
@@ -324,7 +324,9 @@ func (u *unmarshalInfo) computeUnmarshalInfo() {
 			}
 			panic("bad type for XXX_extensions field: " + f.Type.Name())
 		}
-		if f.Name == "XXX_NoUnkeyedLiteral" || f.Name == "XXX_sizecache" {
+		// ignore all fields that are not known to gogo/protobuf, to avoid panic
+		// this is a temporary hack, waiting for containerd removal of gogo/protobuf dependency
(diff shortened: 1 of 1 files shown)


littlejawa:
@fidencio - I would like your comments on this one as you've fought with this problem in the past

littlejawa:
The vendor verification fails, of course :-(

I'm not too familiar with go, but I guess the clean way of doing it would be to fork gogo/protobuf, put my change there, and vendor that instead of the original gogo/protobuf (with a "replace" directive) ?
Is there an easier way to make it?

Anyway, before looking into it, I'd like some discussion/agreement on the direction I'm taking.

haircommander:
> 

yeah that's the right direction, unless they'd upstream this patch. do you want to try upstreaming first?

haircommander:
> > 
> 
> yeah that's the right direction, unless they'd upstream this patch. do you want to try upstreaming first?

LOL I should have looked at the patch first. no chance they'd accept. I agree, replace directive is the way. I am generally not stoked about this change but as a temporary way to unbork kata I am fine with it

littlejawa:
I think #6381 is a cleaner solution - closing

PR #6375, opened 2022-11-11
Add Spoofed Container Runtime

HonakerM:
#### What type of PR is this?

/kind feature

This PR adds support for a new spoofed CRIO runtime which can mock the creation of containers and sandboxes. These containers report status as usual and can be interacted with using crictl; however, they do not start runc containers saving resources and image pull time. The original idea behind this functionality was to speed up development and testing of [operators](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/), which require orchestrating a large number of pods that can have lengthy startup and image pull times. 

This is my first PR for CRIO so any feedback would be greatly appreciated. I opened this PR after [some discussion on slack](https://kubernetes.slack.com/archives/CAZH62UR1/p1668093677867629?thread_ts=1668031829.089849&cid=CAZH62UR1) to get more feedback on the functionality.

no user facing change but it does add more configuration options. I'm not sure if that counts as a user-facing change.
 
```release-note
None
```

### DIFF
--- a/internal/oci/runtime_spoofed.go
+++ b/internal/oci/runtime_spoofed.go
@@ -58,7 +58,7 @@ func (r *runtimeSpoofed) CreateContainer(ctx context.Context, c *Container, cgro
 	}
 	err := cmd.Start()
 	if err != nil {
-		return fmt.Errorf("error creating spoofed process: %v", err)
+		return fmt.Errorf("Error creating spoofed process: %v", err)
 	}
 	pid := cmd.Process.Pid
 	logrus.Infof("Setting spoofed Pid to %d", pid)
@@ -147,7 +147,7 @@ func (r *runtimeSpoofed) StopContainer(ctx context.Context, c *Container, timeou
 	}
 
 	if err = proc.Kill(); err != nil {
-		log.Warnf(ctx, "failed to kill process: %s", err)
+		log.Warnf(ctx, "Failed to kill process: %s", err)
 	}
 
 	c.state.Status = ContainerStateStopped
--- a/internal/oci/runtime_spoofed.go
+++ b/internal/oci/runtime_spoofed.go
@@ -1,0 +1,245 @@
+package oci
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"syscall"
+	"time"
+
+	"github.com/cri-o/cri-o/internal/log"
+	"github.com/cri-o/cri-o/pkg/config"
+	"github.com/cri-o/cri-o/utils/cmdrunner"
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/net/context"
+	"k8s.io/client-go/tools/remotecommand"
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+)
+// runtimeSpoofed is the Runtime interface implementation relying on conmon to
+// interact with the container runtime.
+type runtimeSpoofed struct {
+	*Runtime
+
+	root    string
+	handler *config.RuntimeHandler
+}
+
+// newRuntimeSpoofed creates a new runtimeSpoofed instance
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -233,6 +234,13 @@ type RuntimeConfig struct {
 	// DropInfraCtr determines whether the infra container is dropped when appropriate.
 	DropInfraCtr bool `toml:"drop_infra_ctr"`
 
+	//TODO Beter place for these?
```
HonakerM, line 237: I'm not very familiar with CRIO's config format so please suggest better names and/or location for these values.

haircommander:
Thanks for the PR @HonakerM 

This feels like a natural fit for a new runtime class. a new runtime handler could be created in `internal/oci/`, and instead of a top-level config field (Spoofed and SpoofedPassThrough), the admin could just specify `default_runtime = "spoofed"`

HonakerM:
@haircommander ohhhhhh! That makes alot more sense! Just to confirm you're talking about adding a new file like `internal/oci/runtime_spoofed.go` which just focuses on spoofing containers. 

And on a different note do you have any suggestion regarding the `SpoofedPassThrough` configuration? From my understanding if the admin selects `default_runtime = "spoofed"` all containers would be spoofed regardless.

haircommander:
> @haircommander ohhhhhh! That makes alot more sense! Just to confirm you're talking about adding a new file like `internal/oci/runtime_spoofed.go` which just focuses on spoofing containers.

Yeah exactly 

> And on a different note do you have any suggestion regarding the `SpoofedPassThrough` configuration? From my understanding if the admin selects `default_runtime = "spoofed"` all containers would be spoofed regardless. Are there configuration options for specific runtime classes?

a pod can opt out of it by setting a different runtime through kubernetes runtime classes

HonakerM:
@haircommander I've updated my PR to implement a separate `spoofed` runtime. Please let me know what you think. I'm not sure where to add test cases for it.... I'm not to familiar with go's test suite just yet so any tips would be greatly appreciated.

internal/oci/runtime_spoofed.go
```diff
@@ -0,0 +1,245 @@
...
+		handler: handler,
+	}
+}
+
+
+// CreateContainer creates a container.
+func (r *runtimeSpoofed) CreateContainer(ctx context.Context, c *Container, cgroupParent string, restore bool) (retErr error) {
+
+	// Start sleep command to create process space for container
+	cmd := cmdrunner.Command("/bin/sleep","infinity") // nolint: gosec
```
haircommander, line 49: I would like to figure out a way to avoid having this process. IT doesn't looks like it's being stopped anywhere, and adds additional overhead. What was the motivation for having it?

HonakerM: The original motivation was to create a process space that the CNI could still hook into as I started running into issues if the PID of the container was 0. However, I'm not sure if it's needed as we could just use the PID of crio.

If this duplicate process is actually needed I can add in process termination/cleanup into` StopContainer`.

HonakerM:
@haircommander thank you for all of your help, I'm sorry for my multiple questions. I'm a little confused on how to add more runtime tests. I see that there is already a set of container tests in `test/ctr.bats` and to run them using the `spoofed` runtime I can use `make RUNTIME_TYPE=spoofed localintegration`. Other than updating the tests to skip when using spoofed [similarly to skipping the pod runtime](https://github.com/cri-o/cri-o/blob/7591cffa2a8b3bfc77190a45db54d90d4e45f61d/test/ctr.bats#L186) is there anything else I need to do?

HonakerM:
Sorry about that! I'm still working on adding tests but am busy with end of year madness. I'll hopefully have this PR in a reviewable state in January

HonakerM:
/remove-lifecycle stale

@haircommander sorry for the delay I've finally got back around to this PR. I've added tests and added process cleanup. Any suggestions regarding [the issue regarding process namespace](https://github.com/cri-o/cri-o/pull/6375#discussion_r1021866210) would be greatly appreciated.

PR #6381, opened 2022-11-16
Use containerd v1.7.0-beta.0 

### LINKED ISSUE #6350: runtime_vm: golang panic: "protobuf tag not enough fields in Status.state" when handling errors from the kata shim

littlejawa:
#### What type of PR is this?

/kind bug

containerd has released a new version v1.7.0-beta.0, where the gogo/protobuf dependency was removed.
I have tried it with cri-o, and it seems to be working as expected.

This is still a beta release though, so we have to be cautious in using it. But at least, it shows that the next release of containerd should fix our problem.

If accepted, this replaces the PR #6371 

Fixes #6350 

```release-note
None
```


saschagrunert (approved):
LGTM

haircommander:
/lgtm

I don't think CI issues are related

fidencio (approved):
lgtm, thanks @littlejawa!

littlejawa:
@haircommander - it seems I can't set the milestone here, but this is the PR that I think should be part of 1.26

haircommander:
@littlejawa can you rebase to get the ci fixes

PR #6384, opened 2022-11-18
WIP: Use conmonrs for namespace handling

saschagrunert:
#### What type of PR is this?

/kind feature

Testing https://github.com/containers/conmon-rs/pull/908 in CRI-O.

Refers to https://github.com/containers/conmon-rs/issues/871

None

```release-note
None
```


internal/oci/oci.go
```diff
@@ -436,3 +437,20 @@ func (r *Runtime) RestoreContainer(ctx context.Context, c *Container, sbSpec *rs
 
 	return impl.RestoreContainer(ctx, c, sbSpec, infraPid, cgroupParent)
 }
+
+func (r *Runtime) CreateNamespaces(ctx context.Context, c *Container) (string, error) {
```
haircommander, line 441: It's possible we may want to define a sandbox level interface that runtimePod also satisfies, and we can have a separate interface for oci and VM (where pinns is called). that may be cleaner from a code organization perspective

saschagrunert:
Will spin-up a new PR for this.

PR #6403, opened 2022-11-28
ResourceStore: cleanup created entries

haircommander:
ResourceStore leaks memory if entries added because the stage is saved are never removed. Fix this by deleting the entry after they're used

Signed-off-by: Peter Hunt~ <pehunt@redhat.com>

#### What type of PR is this?

/kind bug

```release-note
Fix a bug where `internal/resourcestore.(*ResourceStore).SetStageForResource` leaks memory 
```


saschagrunert (approved):
LGTM

PR #6404, opened 2022-11-29
Support for Evented PLEG

sairameshv:
#### What type of PR is this?
/kind feature

This PR implements the following enhancement related to the [Evented PLEG](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/3386-kubelet-evented-pleg/README.md)
Fixes # https://github.com/kubernetes/enhancements/issues/3386

This PR contains a new config option to enable "evented-pleg" while starting/running the crio binary. The same has been updated wherever required.

```release-note
Added a new boolean configuration flag "--evented-pleg"(defaulted to "false") to enable the evented pleg mechanism in cri-o. The environment variable "EVENTED_PLEG" when set to "true" also enables the evented pleg in the cri-o.
```


server/cri/v1/rpc_container_events.go
```diff
@@ -5,6 +5,13 @@ import (
...
 func (s *service) GetContainerEvents(req *pb.GetEventsRequest, ces pb.RuntimeService_GetContainerEventsServer) error {
-	// Spoofing the GetContainerEvents method to return nil response until actual support code is added.
+
+	if s.server.Config().EventedPLEG {
+		for containerEvent := range s.server.ContainerEventsChan {
+			if err := ces.Send(&containerEvent); err != nil {
+				return err
+			}
+		}
+	}
```
klihub, line 15: Is this really how this should work ? Maybe I misread the code, but to me it looks like that you are
  - using `s.server.ContainerEventsChan` to buffer at most a 1000 events at a time
  - you send all pending/'buffered' events from the channel to the stream when you get a `GetContainerEvents()`
  - you never send any subsequent events to the stream (even if the client sticks around)

I might be totally wrong here, but my impression of the intention of this new CRI API was that when a client makes a `GetContainerEvents()` call, it should keep receiving container events over the stream until the client itself goes away (or until the server decides to stop sending events and close the stream). And that there might be several concurrent clients listening for container events...

server/container_stop.go
```diff
@@ -38,6 +38,12 @@ func (s *Server) StopContainer(ctx context.Context, req *types.StopContainerRequ
 		return err
 	}
 
+	if s.config.EventedPLEG {
+		err := s.generateCRIEvent(ctx, c, types.ContainerEventType_CONTAINER_DELETED_EVENT)
+		if err != nil {
```
haircommander, line 43: nit: these can be compressed:
`if err := ...; err != nil`

server/container_stop.go
```diff
@@ -38,6 +38,12 @@ func (s *Server) StopContainer(ctx context.Context, req *types.StopContainerRequ
...
+			log.Errorf(ctx, "Unable to generate event %s for container %s due to err %s", types.ContainerEventType_CONTAINER_DELETED_EVENT, c.ID(), err)
```
haircommander, line 44: TODO item: should we emit a metric for events that were failed to generate?

server/cri/v1/rpc_container_events.go
```diff
@@ -5,6 +5,13 @@ import (
 )
 
 func (s *service) GetContainerEvents(req *pb.GetEventsRequest, ces pb.RuntimeService_GetContainerEventsServer) error {
```
haircommander, line 7: a nit about code structure: we have a convention of having functions in this folder this to call directly into a function in the server (see the other rpcs for examples)

server/server.go
```diff
@@ -851,7 +861,114 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+			return []*types.ContainerStatus{}, err
+		}
+		containerStatuses = append(containerStatuses, resp.GetStatus())
+
+	}
+
+	return containerStatuses, nil
+}
+
+func (s *Server) generateCRIEvent(ctx context.Context, container *oci.Container, eventType types.ContainerEventType) error {
```
haircommander, line 940: Since this is a function of the server, can we move the 
`if s.config.EventedPLEG {` piece in here (and return nil if it is) so the callers don't have to check and we can reduce duplication?

server/server.go
```diff
@@ -338,6 +339,10 @@ func (s *Server) Shutdown(ctx context.Context) error {
 		}
 	}
 
+	if s.config.EventedPLEG {
+		close(s.ContainerEventsChan)
```
haircommander, line 343: any reason we can't unconditionally close this?

docs/crio.8.md
```diff
@@ -218,6 +219,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--enable-tracing**: Enable OpenTelemetry trace data exporting.
 
+**--evented-pleg**: Enable CRI-O to generate the container pod-level events in order to optimize the performance of the Pod Lifecycle Event Generator (PLEG) module in Kubelet.
```
rphillips, line 222: What do you think of `--emitted-events`. @haircommander thoughts on the name?

Seems weird to tie this directly with the pleg.

haircommander: I get what you mean, but I'm not as worried about it, as kubelet is the only client we care about. I think it would be most clear as `--emit-pod-and-container-events` but that's too verbose. I think pleg describes what they're for

haircommander:
something worth noting is, as it is now, all clients share the same events source, so if we add a `crictl events` call, kubelet will miss those events... I don't think it's vital to address now but is something we may want to fix in the future

server/cri/v1/rpc_container_events.go
klihub: Again IIUC, the exhibited behavior of this implementation will be that when a client does a `GetContainerEvents()`, it will get all the queued events from the `ContainerEventsChan` and then receive nothing ever after. Also, if there has been more than a 1000 events emitted/generated by that time, the client will get the oldest 1000 with anything newer being already lost.

So shouldn't this instead work somehow so that:
- in `GetContainerEvents()` you simply store the new client stream along with any other existing ones
- you set up an 'event pump' goroutine which sits in loop reading events from `ContainerEventsChan` and multiplexes them to all client streams
- in the event pump you close/remove any client stream on which sending an event fails (client gone, or too slow to read)

haircommander: +1 to a multiplexed approach

sairameshv: @klihub, @haircommander  Thanks for the suggestions, let me check this out and get back.

docs/crio.8.md
saschagrunert: I think we should follow the `--enable-…` pattern for features to be used. Let me throw in a third opinion, how about: `--enable-pod-events`?

internal/criocli/criocli.go
```diff
@@ -1048,6 +1050,11 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			EnvVars: []string{"CONTAINER_ENABLE_CRIU_SUPPORT"},
 			Value:   false,
 		},
+		&cli.BoolFlag{
+			Name:    "evented-pleg",
+			Usage:   fmt.Sprintf("If true, the crio starts sending the container events to the Kubelet (default: %v)", defConf.EventedPLEG),
```
saschagrunert, line 1055:
```suggestion
			Usage:   "If true, CRI-O starts sending the container events to the kubelet",
```

server/container_create.go
```diff
@@ -470,6 +470,13 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 
 	newContainer.SetCreated()
 
+	if s.config.EventedPLEG {
+		err := s.generateCRIEvent(ctx, newContainer, types.ContainerEventType_CONTAINER_CREATED_EVENT)
+		if err != nil {
```
saschagrunert, line 475:
```suggestion
		if err := s.generateCRIEvent(ctx, newContainer, types.ContainerEventType_CONTAINER_CREATED_EVENT); err != nil {
```

server/container_remove.go
```diff
@@ -42,7 +42,12 @@ func (s *Server) RemoveContainer(ctx context.Context, req *types.RemoveContainer
 			}
 		}
 	}
-
+	if s.config.EventedPLEG {
+		err := s.generateCRIEvent(ctx, c, types.ContainerEventType_CONTAINER_DELETED_EVENT)
+		if err != nil {
+			log.Errorf(ctx, "Unable to generate event %s for container %s due to err %s", types.ContainerEventType_CONTAINER_DELETED_EVENT, c.ID(), err)
```
saschagrunert, line 48: If we just log the errors, why do we even bother returning it from `generateCRIEvent` ? We could combine the error logging into the method to reduce the overhead. We can also check `if s.config.EventedPLEG` in `generateCRIEvent` to make it a no op.

server/container_stop.go
saschagrunert: Yes, that's a great idea! Not per container but just having a counter collecting failures.

server/server.go
```diff
@@ -851,7 +861,114 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+func (s *Server) getSandboxStatuses(ctx context.Context, sandboxId string) (*types.PodSandboxStatus, error) {
+	sandboxStatusRequest := &types.PodSandboxStatusRequest{PodSandboxId: sandboxId}
+	sandboxStatus, err := s.PodSandboxStatus(ctx, sandboxStatusRequest)
+
+	if IsNotFound(err) {
+		return nil, err
+	}
+
+	if err != nil {
+		return nil, err
```
saschagrunert, line 884: Let's wrap errors to be trace-able:
```suggestion
		return nil, fmt.Errorf("get pod sandbox statuses: %w", err)
```

server/server.go
```diff
@@ -851,7 +861,114 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+	}
+
+	return sandboxStatus.GetStatus(), nil
+}
+
+func (s *Server) getContainerStatuses(ctx context.Context, sandboxUid string) ([]*types.ContainerStatus, error) {
+	listContainerRequest := &types.ListContainersRequest{Filter: &types.ContainerFilter{LabelSelector: map[string]string{"io.kubernetes.pod.uid": sandboxUid}}}
```
saschagrunert, line 891: Where does the label selector come from?

server/server.go
```diff
@@ -851,7 +861,114 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+	containers, err := s.ListContainers(ctx, listContainerRequest)
+
+	if err != nil {
+		return []*types.ContainerStatus{}, err
+	}
+
+	containerStatuses := []*types.ContainerStatus{}
+	for _, cc := range containers.GetContainers() {
+		containerStatusRequest := &types.ContainerStatusRequest{ContainerId: cc.Id, Verbose: false}
```
saschagrunert, line 900:
```suggestion
		containerStatusRequest := &types.ContainerStatusRequest{ContainerId: cc.Id}
```

server/server.go
```diff
@@ -851,7 +861,114 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+	}
+
+	sandboxStatuses, err := s.getSandboxStatuses(ctx, s.GetSandbox(container.Sandbox()).ID())
+
+	if IsNotFound(err) {
+		return nil
+	}
+
+	if err != nil {
+		return fmt.Errorf("generateCRIEvent: failed to get sandbox statuses of the pod %s due to err %v", sandboxStatuses.Metadata.Uid, err)
```
saschagrunert, line 956:
```suggestion
		return fmt.Errorf("generateCRIEvent: failed to get sandbox statuses of the pod %s: %w", sandboxStatuses.Metadata.Uid, err)
```

server/server.go
```diff
@@ -851,7 +861,114 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+	}
+
+	containerStatuses, err := s.getContainerStatuses(ctx, sandboxStatuses.Metadata.Uid)
+
+	if err != nil {
+		return fmt.Errorf("generateCRIEvent: failed to get container statuses of the pod %s due to err %v", sandboxStatuses.Metadata.Uid, err)
```
saschagrunert, line 962:
```suggestion
		return fmt.Errorf("generateCRIEvent: failed to get container statuses of the pod %s: %w", sandboxStatuses.Metadata.Uid, err)
```

vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go
```diff
@@ -2068,9 +2068,13 @@ type PodSandboxStatusResponse struct {
 	// value should be in json format. The information could include anything useful for
```
saschagrunert, line 2068: Woud vendoring the latest Kubernetes v1.26 RC help here?

saschagrunert:
> something worth noting is, as it is now, all clients share the same events source, so if we add a `crictl events` call, kubelet will miss those events... I don't think it's vital to address now but is something we may want to fix in the future

Yeah we would have to store the clients and broadcast to them.

server/server.go
sairameshv: Close of a `nil` channel might `panic`

server/cri/v1/rpc_container_events.go
sairameshv: Addressed

server/server.go
sairameshv: Close of a `nil` channel results in a `panic`
We are [creating](https://github.com/cri-o/cri-o/pull/6404/files#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4R443) this channel only in the case of enabled Evented PLEG scenario.

server/server.go
sairameshv: Agreed, Addressed

server/container_stop.go
sairameshv: Addressed

server/container_start.go
```diff
@@ -105,3 +112,15 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
 	}).Infof("Started container")
 	return nil
 }
+
+func IsNotFound(err error) bool {
```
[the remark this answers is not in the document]
sairameshv: Done

internal/criocli/criocli.go
sairameshv: Addressed

server/container_remove.go
sairameshv: Done

server/server.go
```diff
@@ -948,27 +950,43 @@ func (s *Server) generateCRIEvent(ctx context.Context, container *oci.Container,
...
 	}
 
 	select {
 	case s.ContainerEventsChan <- types.ContainerEventResponse{ContainerId: container.ID(), ContainerEventType: eventType, CreatedAt: time.Now().UnixNano(),
 		PodSandboxStatus: sandboxStatuses, ContainersStatuses: containerStatuses}:
 		log.Debugf(ctx, "Container event %s generated for %s", eventType, container.ID())
 	default:
-		return fmt.Errorf("generateCRIEvent: failed to generate event %s for container %s", eventType, container.ID())
+		err := fmt.Errorf("generateCRIEvent: failed to generate event %s for container %s", eventType, container.ID())
+		log.Errorf(ctx, "%w", err)
```
saschagrunert, line 975: Only log if `err != nil`, otherwise we will throw an `<nil>` log message.

sairameshv: This `err` can never be a `nil` right?

saschagrunert: Ugh, my fault :+1:

server/server.go
```diff
@@ -851,7 +863,130 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+			return []*types.ContainerStatus{}, err
+		}
+		containerStatuses = append(containerStatuses, resp.GetStatus())
+
+	}
+
+	return containerStatuses, nil
+}
+
+func (s *Server) generateCRIEvent(ctx context.Context, container *oci.Container, eventType types.ContainerEventType) error {
```
saschagrunert, line 937: No need to return the error here if we do not check it anywhere. :)

server/server.go
haircommander: can you put a comment about that to make that clearer?

server/server.go
sairameshv: I agree, but in future, if we have to rely on the returned error (Ex: metric calculation?) in the respective functions, I thought this signature would help. WDYT?

saschagrunert: Hm, the linter will definitely complain that the error is not checked.

sairameshv: Addressed

docs/crio.8.md
rphillips: I like `--enable-pod-events`.

haircommander: +1

test/ctr.bats
```diff
@@ -147,6 +147,43 @@ function check_oci_annotation() {
...
+	ENABLE_POD_EVENTS=true start_crio
+
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	output=$(crictl pods --quiet)
+	[[ "$output" == "$pod_id" ]]
+
+	ctr_id=$(crictl create "$pod_id" "$TESTDATA"/container_redis.json "$TESTDATA"/sandbox_config.json)
+	output=$(crictl ps --quiet --state created)
+	[[ "$output" == "$ctr_id" ]]
+	wait_for_log "Container created event generated"
```
littlejawa, line 160: The test is failing because there is no such line in the logs. But I see:
`msg="Container event CONTAINER_CREATED_EVENT generated for 575c19f6a48f25b8d8789ce6b265c0071b3364698260e9e83eb8094984b1fddf" file="server/server.go:971"`

You probably need to update the strings here?

server/container_events.go
```diff
@@ -0,0 +1,17 @@
+package server
+
+import (
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+)
+
+// GetContainerEvents sends the stream of container events to the client - kubelet
```
haircommander, line 7: while we primarily care about the kubelet it's not necessarily the only client. there may be a `crictl events` call one day

server/server.go
```diff
@@ -851,7 +866,123 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
...
+
+	return containerStatuses, nil
+}
+
+func (s *Server) generateCRIEvent(ctx context.Context, container *oci.Container, eventType types.ContainerEventType) {
+	// returning no error if the Evented PLEG feature is not enabled
+	if !s.config.EnablePodEvents {
+		return
+	}
+	if err := s.Runtime().UpdateContainerStatus(ctx, container); err != nil {
```
haircommander, line 941: I am still pretty anxious about adding this call. this means for every single pod event there's a corresponding call to `runc state` and all the overhead of managing a new process. We really shouldn't need this call, and if we do, we're not reacting to a container exit correctly, or are sending the event too early

server/server.go
klihub: We already vendor k8s pkg/kubelet/types and use it in a few places, including server/*.go. Why not to just `import kubetypes "k8s.io/kubernetes/pkg/kubelet/types"` in server.go and use the const name `kubetypes.KubernetesPodUIDLabel` instead of the duplicated constant value ?

haircommander:
@sairameshv there's a bunch of lint, shfmt, vendor and docs issues to address before this can merge (cc @harche )

PR #6412, opened 2022-12-01
Wipe faster and pass around timeouts

saschagrunert:
#### What type of PR is this?

/kind feature

We now define the default timeout in the server and pass it around directly from the RPC level. In CRI-O wipe we use a timeout of `0` to just wipe immediately rather than waiting the default 10 seconds.

None
None

```release-note
Updated CRI-O wipe to stop containers immediately.
```


server/server.go
```diff
@@ -586,7 +587,13 @@ func (s *Server) wipeIfAppropriate(ctx context.Context, imagesToDelete []string)
 			}
 		}
 		for _, sb := range s.ContainerServer.ListSandboxes() {
-			if err := s.removePodSandbox(ctx, sb); err != nil {
+			const timeoutSec = 0
+
+			if err := s.stopPodSandbox(ctx, sb, timeoutSec); err != nil {
```
haircommander, line 592: I would not expect any containers to be running in this case. We should have already figured out it's a reboot scenerio

saschagrunert: Should we avoid calling `s.restore` in case of a wipe: https://github.com/cri-o/cri-o/blob/5a46e7eb35333c4d4bb71e7d470b71236cea67ac/server/server.go#L451

haircommander: hmm it would break image wiping if we didn't know which containers previously were running. plus, we need to be able to call into CNI to remove the networking resources, which restore sets up

saschagrunert: Hm, this is tricky so I do not want to change the logic too much to break existing setups. The linked bug (https://issues.redhat.com/browse/OCPBUGS-3183) indicates that there are containers running which _maybe got restored_? The timeout of 10 seconds will make CRI-O wipe slow, especially since we do not stop them in parallel.

mrunalp: Are we missing the state that tells us whether it is a reboot vs. a crio restart? If we detect a reboot we could take a big hammer and all of the state w/o having to go through individual pods.

saschagrunert:
Gave this one another rebase.

saschagrunert:
Rebased, PTAL @cri-o/cri-o-maintainers

PR #6416, opened 2022-12-02
Fix conmonrs cgroup when infra container is dropped

mrunalp:
#### What type of PR is this?
/kind bug

We didn't handle the case of moving conmonrs to the right
cgroup when infra container is dropped. This PR fixes that.

```release-note
Fix conmonrs cgroup when infra container is dropped
```


mrunalp:
I know what's wrong with the failing test. I will push a fix tomorrow.

PR #6438, opened 2022-12-08
[1.22] Wipe faster and pass around timeouts

saschagrunert:
#### What type of PR is this?

/kind feature

We now define the default timeout in the server and pass it around directly from the RPC level. In CRI-O wipe we use a timeout of `0` to just wipe immediately rather than waiting the default 10 seconds.

Manual cherry-pick of e8ab5cad9532a03665057ba4d0fe4e5e5eaa7aee

Refers to: https://github.com/cri-o/cri-o/pull/6412

Refers to https://issues.redhat.com/browse/OCPBUGS-3183

None

```release-note
Updated CRI-O wipe to stop containers immediately.
```

### DIFF
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package server
@@ -15,7 +16,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error {
+func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox, timeout int64) error {
 	stopMutex := sb.StopMutex()
 	stopMutex.Lock()
 	defer stopMutex.Unlock()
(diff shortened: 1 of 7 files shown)


server/sandbox_stop_linux.go
```diff
@@ -55,7 +56,7 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error
 				}
 				c := ctr
 				waitGroup.Go(func() error {
-					if err := s.StopContainerAndWait(ctx, c, int64(10)); err != nil {
+					if err := s.StopContainerAndWait(ctx, c, timeout); err != nil {
```
EmmanuelKasper, line 59: Instead of a 10 seconds hardcoded timeout to stop containers,  we will have now a configured timeout, with a value of 10. How will this speed up actually the crio wipe ? Should  end userss change the value in the crio config too ?

haircommander: The timeout shouldn't need to be configured for crio wipe

If the node rebooted (which is the likliest reason the pod can't be restored) then there's no process to stop, so this is basically just updating runtime metadata

If just crio restarted, and there was some issue with the restore, then we likely want to restart the pod asap to allow for a new pod to be created that is not in this state. Note: I would consider a failure to restore a rare event

PR #6450, opened 2022-12-14
server: fail if HOME variable has a newline 

haircommander:
Signed-off-by: Peter Hunt~ <pehunt@redhat.com>

#### What type of PR is this?

/kind deprecation

```release-note
Fix CVE-2022-4318 by failing to create container if it's passed a HOME environment variable with a newline
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -201,6 +201,9 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
 	for _, env := range specgen.Config.Process.Env {
 		if strings.HasPrefix(env, "HOME=") {
 			homedir = strings.TrimPrefix(env, "HOME=")
+			if idx := strings.Index(homedir, `\n`); idx > -1 {
+				return fmt.Errorf("invalid HOME environment; newline not allowed")
+			}
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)


haircommander:
@cri-o/cri-o-maintainers PTAL

haircommander:
@umohnani8 @mrunalp @kolyshkin PTAL

haircommander:
@QiWang19 @rphillips @sohankunkerkar @wgahnagl @harche @klihub   PTAL

server/container_create.go
```diff
@@ -201,6 +201,9 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
 	for _, env := range specgen.Config.Process.Env {
 		if strings.HasPrefix(env, "HOME=") {
 			homedir = strings.TrimPrefix(env, "HOME=")
+			if idx := strings.Index(homedir, `\n`); idx > -1 {
+				return fmt.Errorf("invalid HOME environment; newline not allowed")
```
harche, line 205: Instead of failing, can we continue by dropping `\n`  from the variable?

haircommander: if the value is set to a path with a newline because the path genuinely has a newline, then the HOME won't be set and the container will fail more nefariously. I think it'd be better to fail quickly and notify the user why, rather than let the homedir be set wrong

harche: Oh ok, makes sense. Thanks.

mrunalp (approved):
/lgtm

sohankunkerkar:
LGTM

sohankunkerkar (approved):

PR #6456, opened 2022-12-18
server/metrics: Update seccomp notifier metrics to reduce cardinality

### LINKED ISSUE #6422: seccomp notifier metric cardinality
### Discussed in https://github.com/cri-o/cri-o/discussions/6418

<div type='discussions-op-text'>

<sup>Originally posted by **pdf** December  2, 2022</sup>
Looking at the recently added seccomp metrics, it looks like the `syscalls` label has essentially unbounded cardinality, since it appears that syscalls and counts are concatenated as a string value, e.g.:

```
container_runtime_crio_containers_seccomp_notifier_count_total{name="…",syscalls="chroot (1x)"} 1
container_runtime_crio_containers_seccomp_notifier_count_total{name="…",syscalls="chroot (2x), swapoff (2x)"} 1
container_runtime_crio_containers_seccomp_notifier_count_total{name="…",syscalls="chroot (1x), swapoff (2x)"} 1
container_runtime_crio_containers_seccomp_notifier_count_total{name="...",syscalls="swapoff (5x)"} 1
```

So, every possible combination of syscall and count may published to this label, making the possible value permutations infinite.

I'm not sure what the ideal structure for these metrics should be, but I don't think the current metrics should stand as they are.</div>

pdf:
Signed-off-by: Peter Fern <github@0xc0dedbad.com>

#### What type of PR is this?

/kind bug

As discussed in #6418 this change reduces cardinality of the seccomp notifier metrics by adjusting the metric output format.

Fixes #6422 

I wasn't able to sort out running the test suite locally in time for release, marking as WIP here to make use of the CI test runners.

This change does modify the metric output format, however the previous metrics have not been included in any prior release.

```release-note
None
```


server/server.go
```diff
@@ -782,7 +782,7 @@ func (s *Server) startSeccompNotifierWatcher(ctx context.Context) error {
 				})
 			}
 
-			metrics.Instance().MetricContainersSeccompNotifierCountTotalInc(ctr.Name(), usedSyscalls)
+			metrics.Instance().MetricContainersSeccompNotifierCountTotalInc(ctr.Name(), syscall)
```
haircommander, line 785: above we pull `notifier.UsedSyscalls`which got use usedSyscalls. I don't see us finding all of the syscalls that were reported. shouldn't we call this function for each syscall?

pdf: This is called in a loop, every time a seccomp notification arrives on the channel:

https://github.com/cri-o/cri-o/blob/093d680dd32276457a62cb3615b45b01610e5030/server/server.go#L746-L750

`usedSyscalls` accumulates syscalls from each iteration in the same loop:

https://github.com/cri-o/cri-o/blob/093d680dd32276457a62cb3615b45b01610e5030/server/server.go#L764

So we do call the function for each syscall - we just incr the metric count for the specific syscall name received each iteration, ie:

| iteration | msg.Syscall() | metric values |
|----------|----------------|----------------|
| 0 | swapoff | `{name="...", syscall="swapoff"} = 1` |
| 1 | swapoff | `{name="...", syscall="swapoff"} = 2` |
| 2 | chroot | `{name="...", syscall="swapoff"} = 2`, `{name="...", syscall="chroot"} = 1` |

pdf:
Looks like the original tests may not have been functional - all the exit code tests for commands that should make forbidden syscalls are failing, eg:

https://github.com/cri-o/cri-o/blob/093d680dd32276457a62cb3615b45b01610e5030/test/seccomp_notifier.bats#L36

Is the test runner perhaps not restricting these calls?

haircommander:
> Looks like the original tests may not have been functional - all the exit code tests for commands that should make forbidden syscalls are failing, eg:
> 
> https://github.com/cri-o/cri-o/blob/093d680dd32276457a62cb3615b45b01610e5030/test/seccomp_notifier.bats#L36
> 
> Is the test runner perhaps not restricting these calls?

That's a known failure we've been investigating (and subsequently ignoring in order to get the release out) in rhel e2e. They're working in fedora. I personally have no idea what's going on, and those that could possibly look into it are already on holiday break.

haircommander:
Thanks a bunch @pdf  for taking this on!

PR #6463, opened 2022-12-20
oci: Enable checkpointing of file locks

rst0git:
#### What type of PR is this?

/kind bug

Since https://github.com/checkpoint-restore/criu/pull/1357, CRIU would fail if it finds file locks being used by the application that is being checkpointed and the --file-locks option has not been specified. This pull request enables checkpointing of file locks by default.

cc: @adrianreber 
```release-note
none
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1502,6 +1502,7 @@ func (r *runtimeOCI) CheckpointContainer(ctx context.Context, c *Container, spec
 	args = append(
 		args,
 		"checkpoint",
+		"--file-locks",
 		"--image-path",
 		imagePath,
 		"--work-path",


haircommander:
/ok-to-test
/approve

thanks @rst0git ! LGTM, @adrianreber PTAL

adrianreber:
@rst0git Is it problematic to always pass that option to CRIU?

rst0git:
> Is it problematic to always pass that option to CRIU?

The support for checkpoint/restore of file locks in CRIU was introduced long time ago with https://github.com/checkpoint-restore/criu/commit/95801808aad3b2786a73e1f4faaab38ceb9b39f0 and https://github.com/checkpoint-restore/criu/commit/8f96ff85ff325e04d5e44083963be8f5f5b10565 and in runc with https://github.com/opencontainers/runc/pull/55.

As far as I understand, checkpoint/restore of file locks is considered to be safe when used with containers.

adrianreber:
> > Is it problematic to always pass that option to CRIU?
> 
> The support for checkpoint/restore of file locks in CRIU was introduced long time ago with [checkpoint-restore/criu@9580180](https://github.com/checkpoint-restore/criu/commit/95801808aad3b2786a73e1f4faaab38ceb9b39f0) and [checkpoint-restore/criu@8f96ff8](https://github.com/checkpoint-restore/criu/commit/8f96ff85ff325e04d5e44083963be8f5f5b10565), and in runc with [opencontainers/runc#55](https://github.com/opencontainers/runc/pull/55). As far as I understand, checkpoint/restore of file locks is considered to be safe when used with containers.
> 
> It should not be problematic for checkpointing because this feature can always be disabled during restore.

I am mainly concerned if it breaks existing cases. Why doesn't CRIU always enable it? Should CRI-O enabled it always if CRIU doesn't?

You should probably remove the `Fixes` to the Kubernetes issue. That issue is about more than just file locks.

rst0git:
> I am mainly concerned if it breaks existing cases.

It shouldn't break existing use-cases. With `--file-locks` specified, it would only enable the checkpointing of file locks instead of exiting with an error. This used to be what CRIU would do by default before https://github.com/checkpoint-restore/criu/pull/1357.

> Why doesn't CRIU always enable it?

CRIU can be used to checkpoint/restore processes that are not running in a container. In such case, CRIU can't guarantee that all processes using a specific file lock are being checkpointed. Thus, to prevent unintentional deadlock, this feature should be enabled by the user.

> Should CRI-O enabled it always if CRIU doesn't?

In contrast to CRIU, CRI-O is used to checkpoint containers. We have seen several use cases that require this feature to be enabled [^1][^2]. According to https://criu.org/File_locks, checkpoint/restore of file locks can only be absolutely safe for containers because a container naturally includes all processes that use a file lock.

[^1]: https://github.com/checkpoint-restore/criu/issues/2018
[^2]: https://github.com/containers/podman/pull/12333

rst0git:
@adrianreber The only more complicated use-case I can think of is when a volume is shared across multiple containers and file locking is used on shared files for read and write. In this case, we need all containers that use the shared volume to be checkpointed simultaneously.

sftim:
Can I check that restore from checkpoint is off by default? (If it isn't, I'm going to have concerns about avenues for misuse).

rst0git:
> Can I check that restore from checkpoint is off by default? (If it isn't, I'm going to have concerns about avenues for misuse).

Yes, when `--file-locks` is not specified, CRIU would skip restoring file locks included in the checkpoint:
https://github.com/checkpoint-restore/criu/blob/a315774e/criu/file-lock.c#L699

sftim:
I want to confirm that CRO-O itself does _not_ have any restore from checkpoint behavior enabled by default. I want people to have to opt in. If it is on by default, at CRI-O level, that's where I'll have a concern.

adrianreber:
> I want to confirm that CRO-O itself does _not_ have any restore from checkpoint behavior enabled by default. I want people to have to opt in. If it is on by default, at CRI-O level, that's where I'll have a concern.

Checkpoint/Restore support is disabled unless you specify --enable-criu-support=true.

Just like mentioned in https://kubernetes.io/blog/2022/12/05/forensic-container-checkpointing-alpha/

adrianreber:
> > Should CRI-O enabled it always if CRIU doesn't?
> 
> In contrast to CRIU, CRI-O is used to checkpoint containers. We have seen several use cases that require this feature to be enabled [1](#user-content-fn-1-53eb66927e98a21d81be3e211c305324)[2](#user-content-fn-2-53eb66927e98a21d81be3e211c305324). According to https://criu.org/File_locks, checkpoint/restore of file locks can only be absolutely safe for containers because a container naturally includes all processes that use a file lock.

The container argument sounds reasonable. So it indeed makes sense to always enable this.

There is still the problem that we need a way to pass additional CRIU parameters from Kubernetes through CRI-O, runc to CRIU somehow, but this parameter sounds like something to have always enabled.

rst0git:
@sftim PTAL

adrianreber:
LGTM

TomSweeneyRedHat:
LGTM

rst0git:
@saschagrunert @haircommander PTAL

PR #6464, opened 2022-12-20
Add additional metadata to inspect and checkpoint

adrianreber:
#### What type of PR is this?

/kind feature

This PR adds additional metadata to the inspect command if checkpointing support is active. The following three fields are added to inspect:

```json
  "checkpointedAt": "timestamp",
  "restored": "true or false",
```

With "checkpointedAt" it is possible to see when a checkpoint was created. With "restoredAt" is is possible to see when this container was restored from a checkpoint. "restored" is true if the container has been restored from a checkpoint for all other container this is set to false.

These information are also part of the checkpoint archive. If checkpoint archive contains a container that has been restored "restoredAt" will be empty but "checkpointedAt" will still contain the time of the checkpointing.

To add these additional fields it was necessary to pull in a newer version of checkpoint-restore/checkpointctl which also required to remove some unused checkpoint code concerning pods. The pod checkpointing code was a left over from the initial merge and the initial proof of concept about checkpointing and restoring pods.

None

```release-note
Extended checkpoint archive metadata with time of checkpointing.
```

### DIFF
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 
 	metadata "github.com/checkpoint-restore/checkpointctl/lib"
 	"github.com/checkpoint-restore/go-criu/v6/stats"
@@ -19,18 +20,15 @@ import (
 	"github.com/opencontainers/runtime-tools/generate"
 )
 
-type ContainerCheckpointRestoreOptions struct {
-	Container string
-	Pod       string
-
-	libpod.ContainerCheckpointOptions
-}
-
 // ContainerCheckpoint checkpoints a running container.
-func (c *ContainerServer) ContainerCheckpoint(ctx context.Context, opts *ContainerCheckpointRestoreOptions) (string, error) {
-	ctr, err := c.LookupContainer(ctx, opts.Container)
+func (c *ContainerServer) ContainerCheckpoint(
+	ctx context.Context,
+	config *metadata.ContainerConfig,
+	opts *libpod.ContainerCheckpointOptions,
+) (string, error) {
+	ctr, err := c.LookupContainer(ctx, config.ID)
 	if err != nil {
-		return "", fmt.Errorf("failed to find container %s: %w", opts.Container, err)
+		return "", fmt.Errorf("failed to find container %s: %w", config.ID, err)
 	}
 
 	configFile := filepath.Join(ctr.BundlePath(), "config.json")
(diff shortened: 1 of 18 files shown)


internal/lib/checkpoint.go
```diff
@@ -164,6 +162,9 @@ func (c *ContainerServer) prepareCheckpointExport(ctr *oci.Container) error {
 			}
 			return c.config.DefaultRuntime
 		}(),
+		CheckpointedAt: time.Now(),
+		RestoredAt:     ctr.RestoredAt(),
+		Restored:       ctr.Restore(),
```
haircommander, line 167: I'm confused, why would there be Restored/RestoredAt if it's checkpointed here? is this for previous checkpoint/restore cycles?

server/container_checkpoint.go
```diff
@@ -23,139 +18,27 @@ func (s *Server) CheckpointContainer(ctx context.Context, req *types.CheckpointC
 		return fmt.Errorf("checkpoint/restore support not available")
 	}
 
-	var opts []*lib.ContainerCheckpointRestoreOptions
-	var podCheckpointDirectory string
-	var checkpointedPodOptions metadata.CheckpointedPodOptions
-
 	_, err := s.GetContainerFromShortID(ctx, req.ContainerId)
 	if err != nil {
-		// Maybe the user specified a Pod
```
haircommander, line 32: would you mind separating this piece into a different commit?

internal/lib/checkpoint.go
adrianreber: CheckpointedAt will always be the time of last checkpoint. Either when the current container was checkpointed or if the current container has not been checkpointed but restored from a checkpoint archive it will be time of the last checkpoint on a possible other system.

RestoredAt will be the time of the last restore. This field is kind of unnecessary because if `Restored` is true it will be the same time as `startedAt`. It gives information about that this is restored container and when it was restored.

One could argue that having both CheckpointedAt and Restored is unnecessary, but a container could have been checkpointed but not restored and then Restored would be false.

Restore and RestoredAt in combination with `startedAt` could be simplified by removing RestoredAt as it does not carry any additional information. It is just easier to see when it was restored.

adrianreber:
This should fix the test coverage now.

adrianreber:
The integration tests are failing because they are using the same file concurrently. I need to create unique temporary file names.

adrianreber:
The `integration / test-userns` failure seems unrelated to my changes.

haircommander:
`not ok 174 metrics container oom` is a known flake :)

internal/lib/checkpoint.go
haircommander: yeah I guess my feeling is it wouldn't hurt to simplify the metadata, as it would reduce code complexity and the tiny amount of overhead it takes to transmit. if you can glean the information needed from less, it woulid be my preference to go that route. WDYT?

adrianreber: removed `restoredAt`

adrianreber:
@haircommander PTAL

saschagrunert (approved):

PR #6485, opened 2023-01-04
mocks: update with new c/storage mocks

haircommander:
Signed-off-by: Peter Hunt~ <pehunt@redhat.com>

#### Does this PR introduce a user-facing change?

/kind ci
```release-note
none
```

PR #6488, opened 2023-01-05
Allow cross building from non-linux

dfr:
This allows building crio.cross.linux.amd64 on non-linux platforms. The only thing which I needed to change was to add non-cgo stubs for seccomp. The main use-case for this is to avoid breaking the Linux build while refactoring to allow for a FreeBSD port of cri-o.

Signed-off-by: Doug Rabson <dfr@rabson.org>

#### What type of PR is this?

/kind cleanup

This is useful to support porting cri-o to non-linux platforms.

None

```release-note
None
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,8 @@ jobs:
       - run: scripts/github-actions-packages
       - run: make
       - run: bin/crio version
+      - run: make bin/crio.cross.linux.amd64
+      - run: bin/crio.cross.linux.amd64 version
       - uses: actions/upload-artifact@v3
         with:
           name: build
--- a/internal/config/seccomp/notifier.go
+++ b/internal/config/seccomp/notifier.go
@@ -1,3 +1,6 @@
+//go:build linux && cgo
+// +build linux,cgo
+
 package seccomp
 
 import (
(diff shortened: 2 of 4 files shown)


haircommander:
oh boy, it's cri-o's turn for the freeBSD port!

/ok-to-test
/approve

thanks @dfr, LGTM

dfr:
I don't understand the lint failure - its complaining about github.com/containers/storage which should not be affected by this PR.

haircommander:
it's unrelated and fixed by https://github.com/cri-o/cri-o/pull/6485

dfr:
Rebased to pick up #6485

dfr:
I'm not sure exactly how my change could cause these test failures - are these flaky?

haircommander:
welcome to cri-o, where our CI is rarely green :sweat_smile: (yeah, the ones failing are known flakes/failures)
@saschagrunert PTAL

saschagrunert (approved):
Code LGTM, it would be great to have a CI test for those non-linux builds just to verify that we do not break anything in the future.

dfr:
> Code LGTM, it would be great to have a CI test for those non-linux builds just to verify that we do not break anything in the future.

I could add a cross-build test in github.com/workflows/test.yml which will at least test that the nocgo build works. In an ideal world, this cross build should happen in a freebsd or darwin VM but that doesn't seem justitifed at this point.

dfr:
Rebased, added a cross-build test to test.yml

haircommander:
as a note: many tests are failing right now, I'll point out any that are legit and override the rest :)

PR #6489, opened 2023-01-05
OCPNODE-1336: Enable upstream CI to test node e2e with evented pleg feature

sairameshv:
Signed-off-by: Sai Ramesh Vanka <svanka@redhat.com>

#### What type of PR is this?
/kind ci

This code change helps in enabling the evented pleg feature in cri-o while running the CI(node e2e tests) jobs in the upstream k8s/test-infra

```release-note
None
```

### DIFF
--- a/scripts/node_e2e_installer
+++ b/scripts/node_e2e_installer
@@ -3,6 +3,7 @@ set -euo pipefail
 
 # Commit to run upstream node e2e tests
 NODE_E2E_COMMIT=085e3a03cf17d8a97e14551ec83137954328f627
+ENABLE_POD_EVENTS=${ENABLE_POD_EVENTS:-"false"}
 
 enable_selinux() {
     # Make sure SELinux is enabled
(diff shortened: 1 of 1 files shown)


scripts/node_e2e_installer
```diff
@@ -30,6 +31,13 @@ install_crio() {
     # Setup log level
     echo "CONTAINER_LOG_LEVEL=debug" >>/etc/sysconfig/crio
 
+    for condition in "true" "True" "TRUE"; do
+        if [[ "${ENABLE_POD_EVENTS}" = "$condition" ]]; then
+            echo "ENABLE_POD_EVENTS=true" >>/etc/sysconfig/crio
```
haircommander, line 36: this likely is a follow up because I see some other examples, but I think we should use drop-in files instead of env variables in sysconfig files, as I would say it's a preferred method of configuring cri-o

sairameshv: Okay

PR #6496, opened 2023-01-07
revert d06cae7: contrib/cni - use cniVersion 1.0.0 for ipv4 bridge config

amarlearning:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Revert d06cae7 commit. cc @haircommander 

We have updated the [other CNI files](https://github.com/cri-o/cri-o/tree/main/contrib/cni) to version `1.0.0`, ipv4 bridge config version was reverted back to 0.3.1 to fix the build issue. The build issue has been identified and [fixed](https://github.com/cri-o/cri-o/pull/6476). Hence reverting this change.

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6541, opened 2023-01-17
Update CNI plugins to v1.2.0

saschagrunert:
#### What type of PR is this?

/kind dependency-change

Update CNI plugins to their latest release.

None
None

```release-note
Updated CNI plugins to v1.2.0.
```

### DIFF
--- a/contrib/bundle/build
+++ b/contrib/bundle/build
@@ -178,7 +178,7 @@ $BOM validate -e "$SPDX_PATH" -d "$CRIODIR"
 pushd "$TMPDIR"
 export DESTDIR=test/
 ./install
-EXP_CNT=64
+EXP_CNT=65
 if ! command -v runc; then
     EXP_CNT=$((EXP_CNT + 1))
 fi
--- a/contrib/cni/11-crio-ipv4-bridge.conflist
+++ b/contrib/cni/11-crio-ipv4-bridge.conflist
@@ -1,5 +1,5 @@
 {
-  "cniVersion": "0.3.1",
+  "cniVersion": "1.0.0",
   "name": "crio",
   "plugins": [
     {
--- a/contrib/cni/99-loopback.conflist
+++ b/contrib/cni/99-loopback.conflist
@@ -1,4 +1,9 @@
 {
     "cniVersion": "1.0.0",
-    "type": "loopback"
+    "name": "loopback",
+    "plugins": [
+        {
+            "type": "loopback"
+        }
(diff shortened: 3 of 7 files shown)


haircommander:
would you also add a commit to check changing the bridge plugin config from 0.3.1 doesn't break things?

haircommander:
also, while you're there, can you bring back https://github.com/cri-o/cri-o/pull/6273/ (the patch was accidentally reverted)

meierra:
Just some quick feedback and testing... I've been getting parsing errors while using the default loopback.conflist.

First was: 
- error parsing configuration list: no 'name' key

Thus I reached out to have https://github.com/cri-o/cri-o/pull/6273 brought back, but while testing now, just adding a name key does not fully resolve,
Second error is:
- error parsing configuration list: no 'plugins' key

If I change the filename to loopback.conf then it does not complain, not sure if the right solution is to rename the file back to .conf or add all the required keys for .conflist?

meierra:
little bit more troubleshooting, (sorry I'm very new at this) seems that this configuration works with conflist:
{
    "cniVersion": "1.0.0",
    "name": "loopback",
    "plugins": [
        {
            "type": "loopback"
        }
    ]
}

Seems that according to spec the plugin's used need to be in a list:
- plugins (list): A list of CNI plugins and their configuration, which is a list of plugin configuration objects.

Hope this helps!

saschagrunert:
Thank you for the pointers @meierra and @haircommander, indeed it looks like that the new config works as expected. Updated as suggested.

meierra:
@saschagrunert thanks, just check, I believe there may still be a typo in the correction, looks like the key "type" is repeated twice
image

the first instance should be "name"

saschagrunert:
> @saschagrunert thanks, just check, I believe there may still be a typo in the correction, looks like the key "type" is repeated twice image
> 
> the first instance should be "name"

Good catch! Fixed as suggested :+1:

haircommander:
one more, can you pick https://github.com/cri-o/cri-o/pull/6496/files to make sure 1.2.0 works as we want?

saschagrunert:
> one more, can you pick https://github.com/cri-o/cri-o/pull/6496/files to make sure 1.2.0 works as we want?

Absolutely, let's see how that works out.

haircommander:
woo! 

/override ci/prow/ci-rhel-integration
/override ci/kata-jenkins

/lgtm

Issue #5653, opened 2022-02-15
Add support for in-place vertical pod autoscaling


haircommander:
which will land in 1.24

more information on expectations on CRI: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources

hasan4791:
@haircommander
Hello, Would like to pick up this and went though the doc,

Aren't we already handling the "UpdateContainerResources". After all the validation done by the kubelet, this api needs to be called if I'm not wrong. So what are we actually looking into this task.

haircommander:
looks like VPA support bumped to 1.25

hasan4791:
Gotcha, we aren't returning any response as of now this day except error.

kangclzjc:
@hasan4791 , are you still working on this 🙂?

keremgocen:
I'm also happy to have a go if no one is on it

saschagrunert:
We require https://github.com/kubernetes/kubernetes/pull/102884 to be merged to get the latest CRI changes in. In the meanwhile we could also work with the fork / PR.

haircommander:
slipped 1.26

PR #6570, opened 2023-01-24
Update container resources in ContainerStatusResponse

hasan4791:
#### What type of PR is this?
/kind other

Return resource status in ContainerStatusResponse

Fixes https://github.com/cri-o/cri-o/issues/5653

Yes

```release-note
Added support to return ContainerResources in ContainerStatusResponse.
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -79,6 +79,7 @@ type Container struct {
 	restore            bool
 	restoreArchive     string
 	restoreIsOCIImage  bool
+	resources          *types.ContainerResources
 }
 
 func (c *Container) CRIAttributes() *types.ContainerAttributes {
@@ -202,6 +203,7 @@ func (c *Container) CRIContainer() *types.Container {
 // SetSpec loads the OCI spec in the container struct
 func (c *Container) SetSpec(s *specs.Spec) {
 	c.spec = s
+	c.SetResources(s)
 }
 
 // Spec returns a copy of the spec for the container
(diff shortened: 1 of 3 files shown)


hasan4791:
> 

Thank you for quick look, Let me check on that too.
Btw do we need linux host to run the prerequisite commands like(make lint or compile the binary & check it)

haircommander:
lint I don't think so, compile I'm pretty sure to get real feedback (it should cross compile but lots of code is linux only)

hasan4791:
@haircommander would you give me a small hint on the container object & state, Im in early stages on getting familiar with crio language 👍🏽

haircommander:
it's defined in `internal/oci/container.go` and can be initialized when the container object is created. Then, a getter can be created and called

hasan4791:
@haircommander container object already has the spec which has the resource status. Still do we need to maintain it on container object?

haircommander:
> @haircommander container object already has the spec which has the resource status. Still do we need to maintain it on container object?

yeah the idea is to reduce the number of objects created and destroyed, and reduce the CPU of convering the objects. Trading the bit of memory to keep this information and reduce CPU has been the direction we've been going with these more frequenly called operations

hasan4791:
@haircommander Could you please take a look at the changes.

server/container_status.go
```diff
@@ -55,6 +55,12 @@ func (s *Server) ContainerStatus(ctx context.Context, req *types.ContainerStatus
 	}
 	resp.Status.Mounts = mounts
 
+	containerSpec := c.Spec()
+	if containerSpec.Linux != nil {
+		resourceStatus := c.GetResources()
+		resp.Status.Resources = resourceStatus
```
haircommander, line 61: this could be done in one step

internal/oci/container.go
```diff
@@ -695,3 +696,55 @@ func (c *Container) RestoreIsOCIImage() bool {
 func (c *Container) SetRestoreIsOCIImage(restoreIsOCIImage bool) {
 	c.restoreIsOCIImage = restoreIsOCIImage
 }
+
+// SetResources loads the OCI Spec.Linux.Resources in the container struct
```
haircommander, line 700: just had this thought: what if we combined this with `SetSpec`? then we wouldn't have to remember to call each time SetSpec was called

hasan4791: makes sense

server/container_status.go
hasan4791: Done

internal/oci/container.go
hasan4791: Done

haircommander:
would you mind squashing your commits?

haircommander:
/approve

hasan4791:
@haircommander Aren;t we supposed to wait for this PR to be merged?
https://github.com/cri-o/cri-o/issues/5653#issuecomment-1162805557

internal/oci/container.go
```diff
@@ -79,6 +79,7 @@ type Container struct {
 	restore            bool
```
jrvaldes, line 79: @hasan4791 Please add a footer in the commit message to reference the issue this PR is fixing. See [contribution guidelines](https://github.com/cri-o/cri-o/blob/main/CONTRIBUTING.md#submitting-pull-requests). 

For example:
```
Closes #5653
```

In addition, consider adding to the commit message what this commit is fixing and why.

internal/oci/container.go
```diff
@@ -707,3 +709,55 @@ func (c *Container) RestoreIsOCIImage() bool {
...
+			if s.Process.OOMScoreAdj != nil {
+				resourceStatus.Linux.OomScoreAdj = int64(*s.Process.OOMScoreAdj)
+			}
+		}
+		resourceStatus.Linux.Unified = make(map[string]string)
+		for key, value := range linuxResources.Unified {
+			resourceStatus.Linux.Unified[key] = value
+		}
+		resourceStatus.Linux.HugepageLimits = []*types.HugepageLimit{}
+		for _, entry := range linuxResources.HugepageLimits {
```
jrvaldes, line 750: Consider adding a unit test for the newly introduced function, it will also make code coverage happy.

jrvaldes (changes requested):
@hasan4791  Thanks for putting this together. PTAL at my comments.

TomSweeneyRedHat:
Other than the remaining comments, LGTM

internal/oci/container.go
hasan4791: Done :)

internal/oci/container.go
hasan4791: Sure, will keep it in mind

hasan4791:
@haircommander Please take a look once !!

haircommander:
still LGTM, @cri-o/cri-o-maintainers PTAL

jrvaldes (approved):
/lgtm

jrvaldes:
@hasan4791  Does this needs to be backported?

cc @haircommander

haircommander:
/retest

> @hasan4791 Does this needs to be backported?

I don't believe so, as the values weren't used until in place VPA support was added which happened this release

TomSweeneyRedHat:
LGTM

PR #6579, opened 2023-01-27
[FEAT]Add new parameter hostnetwork-disable-selinux

### LINKED ISSUE #5501: Host network and SELinux type
When running a Pod with `hostNetwork` the container process will always take the privileged `spc_t` type.

This issue was reported in [openshift/origin/issues/25488](https://github.com/openshift/origin/issues/25488). From the comments the relevant file [container_create_linux.go](https://github.com/cri-o/cri-o/blob/main/server/container_create_linux.go) and [commit](https://github.com/cri-o/cri-o/commit/349b85a7f2b97d23fd65490289bf5d66e888e0f9).

It's not clear what is the reason for not allowing it. If possible, can we remove the restriction?

hasan4791:
Signed-off-by: T K Chandra Hasan <t.k.chandra.hasan@ibm.com>

#### What type of PR is this?
/kind feature

Fixes #5501

Yes

```release-note
Add HostNetworkDisableSELinux option to configuration, 
allowing users to enable/disable selinux within a pod when it is running in the host network namespace.
```


hasan4791:
@haircommander Do we need to add this condition to runPodSandbox function also?

hasan4791:
> > @haircommander Do we need to add this condition to runPodSandbox function also?
> 
> I don't think so /ok-to-test
> 
> Thanks for going for this @hasan4791 I wish there was a more terse flag/config field that would describe the situation, but I'm having trouble coming up with something. Maybe `HostNetworkDisableSELinux`?

Sure, also how about the parameter description? If it needs any update, please provide your comments, will push all in one go

completions/fish/crio.fish
```diff
@@ -131,6 +131,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l rdt-config-file -r -d 'Pat
 complete -c crio -n '__fish_crio_no_subcommand' -f -l read-only -d 'Setup all unprivileged containers to run as read-only. Automatically mounts the containers\' tmpfs on `/run`, `/tmp` and `/var/tmp`.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l registry -r -d 'Registry to be prepended when pulling unqualified images. Can be specified multiple times.'
 complete -c crio -n '__fish_crio_no_subcommand' -l root -s r -r -d 'The CRI-O root directory.'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l run-as-spc-when-hostnetwork -d 'Determines whether pods are run as SPC, when hostNetwork is set.'
```
haircommander, line 134: nit: maybe `Whether SELinux should be disabled within a pod when it is running in the host network namespace`

haircommander:
I think anytime we add a config field it warrants a release note, can you add one, please :)

hasan4791:
> I think anytime we add a config field it warrants a release note, can you add one, please :)

Than you @haircommander Updated PR and will keep in mind :)

haircommander:
/test kata-containers

@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):
LGTM

hasan4791:
Validation is failing for "irqbalance-config-restore-file" which isn't related to this PR.
@haircommander shall i include the fix here?

haircommander:
how about you open a new PR for it and rebase this on that?

docs/crio.8.md
```diff
@@ -261,6 +262,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
     Kubernetes configuration are considered. Bind mounts that CRI-O
     inserts by default (e.g. '/dev/shm') are not considered. (default: "/usr/share/containers/oci/hooks.d")
 
+**--hostnetwork-disable-selinux**: Determines whether SELinux should be disabled within a pod when it is running in the host network namespace.
```
TomSweeneyRedHat, line 265: It would probably be good to at least add a warning that turning SELinux off is not a recommended configuration.

rhatdan: Remind me why running in the host network would ever require you to disable SELinux separation?  It shouldn't require that, as far as I can tell.

hasan4791: This was introduced in this commit

https://github.com/cri-o/cri-o/commit/349b85a7f2b97d23fd65490289bf5d66e888e0f9

In this PR, we are introducing a new flag to enable/disable the selinux for pod by not breaking what's already there.

atiratree:
@hasan4791 I welcome this PR, but we shouldn't say that it closes the original issue https://github.com/cri-o/cri-o/issues/5501, because it also links to this issue https://github.com/openshift/origin/issues/25488 which reports similar problems with `hostPID=true` and `hostIPC=true` which we are also running into in https://github.com/openshift/oc/issues/641.

Can we extend the scope for these settings as well in this PR or in a followup?

saschagrunert:
> Can we extend the scope for these settings as well in this PR or in a followup?

Let's keep the scope of this PR and move the other topic to https://github.com/cri-o/cri-o/pull/6780

PR #6603, opened 2023-02-02
Fix issues with PLEG events for kata

### LINKED ISSUE #6481: kata tests are broken
### Which jobs are failing?

kata-jenkins

### Which tests are failing?

looks like cri-o has been failing to setup: http://jenkins.katacontainers.io/job/kata-containers-2-crio-PR/4525/consoleFull

### Since when has it been failing?

Pre-holiday, I overrode it to release 1.26.0 a number of times

### Testgrid link

_No response_

### Reason for failure (if possible)

_No response_

_No response_

littlejawa:
#### What type of PR is this?

/kind bug

The integration test "ctr pod lifecycle with evented pleg enabled" is failing for kata containers.
This is because of how the container's exit is handled for kata.

For regular containers, the exit is detected by the `server.exitsMonitor()`, and it calls `handleExit()` for it, where the event `CONTAINER_STOPPED_EVENT` is generated.

For kata containers, the exit is detected by a wait function started when the container is started, and it doesn't send the event.

The event can be sent by calling the function `server.generateCRIEvent()`. But having a handle on the server in runtimeVM seems cumbersome.
Also I think runtimeVM containers could benefit from leveraging the handleExit() code in general - rather than handling the exit separately.

This is why I modified the code on the wait function for runtimeVM so that it just creates a file in the exitsPath, where exitsMonitor() is watching. This makes the watcher trigger the handleExit() function for kata containers too, and generates the event.

The second patch is fixing an issue that is triggered by the removal of the container: the code is updating the container status, but the container is already gone at that time, and this was triggering an error on the runtimeVM side, which prevented generating the `CONTAINER_DELETED_EVENT`. Ignoring the error on this specific situation seems to solve the issue.

Fixes #6481 

This PR stems from the discussion that happened on #6531. While the original fix on this PR was wrong, the discussion helps understand the requirements for evented PLEG.

```release-note
none
```

### DIFF
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -176,7 +176,7 @@ func (r *Runtime) newRuntimeImpl(c *Container) (RuntimeImpl, error) {
 	}
 
 	if rh.RuntimeType == config.RuntimeTypeVM {
-		return newRuntimeVM(rh.RuntimePath, rh.RuntimeRoot, rh.RuntimeConfigPath), nil
+		return newRuntimeVM(rh.RuntimePath, rh.RuntimeRoot, rh.RuntimeConfigPath, r.config.RuntimeConfig.ContainerExitsDir), nil
 	}
 
 	if rh.RuntimeType == config.RuntimeTypePod {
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
Changes look OK to me, but lint isn't buying it.

Issue #6625, opened 2023-02-09
Runtime service generates unexpected CONTAINER_DELETED_EVENT on container stop


akornatskyy:
### What happened?

Runtime service generates unexpected CONTAINER_DELETED_EVENT on container stop

### What did you expect to happen?

container stop should not generate CONTAINER_DELETED_EVENT.

### How can we reproduce it (as minimally and precisely as possible)?

1. start server
2. client getContainerEvents
3. run pod -> CONTAINER_STARTED_EVENT
4. create container -> CONTAINER_CREATED_EVENT
5. start container -> CONTAINER_STARTED_EVENT
6. stop container -> CONTAINER_STOPPED_EVENT, CONTAINER_DELETED_EVENT
7. rm container -> CONTAINER_DELETED_EVENT

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
crio version 1.26.1
Version:        1.26.1
GitCommit:      69945ab622a70ab01f4b0df10b107bc48e38dc9e
GitCommitDate:  2023-01-10T21:20:34Z
GitTreeState:   dirty
BuildDate:      1980-01-01T00:00:00Z
GoVersion:      go1.19.4
Compiler:       gc
Platform:       linux/amd64
Linkmode:       static
BuildTags:      
  static
  netgo
  osusergo
  exclude_graphdriver_btrfs
  exclude_graphdriver_devicemapper
  seccomp
  apparmor
  selinux
LDFlags:          unknown
SeccompEnabled:   true
AppArmorEnabled:  false
Dependencies:     

```
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
Debian
$ uname -a
6.1
```

</details>

### Additional environment details (AWS, VirtualBox, physical, etc.)

-

hasan4791:
Looks like 2 events are being generated, DELETED then STOPPED
https://github.com/cri-o/cri-o/blob/main/server/container_stop.go#L45
https://github.com/cri-o/cri-o/blob/main/server/sandbox_stop_linux.go#L97

akornatskyy:
fantastic... just closed since no one cares.

haircommander:
I'm pretty sure this was closed because https://github.com/cri-o/cri-o/pull/6633 fixed it. did you get a chance to check it out?

akornatskyy:
with the next release, I guess it will come soon.

Issue #6631, opened 2023-02-10
Fix Event Types generated in the Evented PLEG


sairameshv:
> The only places in cri-o where `CONTAINER_STOPPED_EVENT` is generated are :
> 
> * [server.go:899](https://github.com/cri-o/cri-o/blob/a03a46ec6c92d91c805e9b18d72da77c8405d5c9/server/server.go#L899), during `handleExit()`.  There's no check on the status of the container, the only other exit paths are for internal errors. To me, it means this event is generated whenever the container exits, whether or not it goes to completion?

`handleExit()` handles containers that CRIO thinks have exited (in other words run to completion). It handles both regular containers as well as sandboxes that have run to completion. 

> * [sandbox_stop_linux.go:97](https://github.com/cri-o/cri-o/blob/a03a46ec6c92d91c805e9b18d72da77c8405d5c9/server/sandbox_stop_linux.go#L97) for the sandbox only, when the `stopPodSandbox()` function is called. Again, I'm not sure it takes the completion into account, it just stops everything.
> 

@sairameshv This looks like a bug. We shouldn't really generate `ContainerEventType_CONTAINER_STOPPED_EVENT` in that file. That's not where we detect the sandbox has run to completion (exited). `handleExit()` function gets invoked when crio gets notified (using fsnotify) that the container has exited and we are already emitting `ContainerEventType_CONTAINER_STOPPED_EVENT` there. 

> Shouldn't we check the status of the container in `handleExit()` before sending the event, if the event depends on the completion? Or am I missing something about what the "completion" status is?
> 

Container running to completion or exited mean same thing. In broad sense, it's just means that the container has terminated itself without kubelet asking for that action. The runtime [has already determined](https://github.com/cri-o/cri-o/blob/a03a46ec6c92d91c805e9b18d72da77c8405d5c9/server/server.go#L884) that container has terminated (with the help of fsnotify). We just want to convey that information to the kubelet when `Evented PLEG` is turned on. 

> The fact that we generate `CONTAINER_STOPPED_EVENT` on `stopPodSandbox()` and not `stopContainer()` is part of why I made this change in the first place - I expected them to be symmetric somehow. If we're not supposed to send `CONTAINER_STOPPED_EVENT` on `stopContainer()`, should we remove it from `stopPodSandbox()` too? If not, why?
> 

`stopPodSandbox` also should not emit `CONTAINER_STOPPED_EVENT`. 

[...]

PR #6633, opened 2023-02-10
Fix events generated by Evented PLEG

sairameshv:
Fixes https://github.com/cri-o/cri-o/issues/6625

Signed-off-by: Sai Ramesh Vanka <svanka@redhat.com>

#### What type of PR is this?

/kind bug
https://github.com/cri-o/cri-o/issues/6625, https://github.com/cri-o/cri-o/issues/6631

```release-note
None
```

### DIFF
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -94,7 +94,12 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error 
 
 	log.Infof(ctx, "Stopped pod sandbox: %s", sb.ID())
 	sb.SetStopped(ctx, true)
-	s.generateCRIEvent(ctx, sb.InfraContainer(), types.ContainerEventType_CONTAINER_STOPPED_EVENT)
+
+	if podInfraContainer.Spoofed() {
+		// event generation would be needed in case of a spoofed infra container where there is no
+		// exit process that hits the handleExit() code.
+		s.generateCRIEvent(ctx, sb.InfraContainer(), types.ContainerEventType_CONTAINER_STOPPED_EVENT)
+	}
 
 	return nil
 }
--- a/server/container_stop.go
+++ b/server/container_stop.go
@@ -42,7 +42,6 @@ func (s *Server) StopContainer(ctx context.Context, req *types.StopContainerRequ
 		log.Warnf(ctx, "NRI stop failed for container %q: %v", c.ID(), err)
 	}
 
-	s.generateCRIEvent(ctx, c, types.ContainerEventType_CONTAINER_DELETED_EVENT)
 	log.Infof(ctx, "Stopped container %s: %s", c.ID(), c.Description())
 	return &types.StopContainerResponse{}, nil
 }


sairameshv:
Verified this change by running the node e2e tests and no test has failed as follows:
```
Ran 183 of 387 Specs in 618.975 seconds
SUCCESS! -- 183 Passed | 0 Failed | 0 Pending | 204 Skipped


Ginkgo ran 1 suite in 10m19.962207252s
Test Suite Passed
You're using deprecated Ginkgo functionality:
=============================================
  --untilItFails is deprecated, use --until-it-fails instead
  Learn more at: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#changed-command-line-flags

To silence deprecations that can be silenced set the following environment variable:
  ACK_GINKGO_DEPRECATIONS=2.7.0


Success Finished Test Suite on Host ramesh-fedora-coreos-37-20230122-3-0-gcp-x86-64
<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
<                              FINISH TEST                               <
<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<

```

Also verified the behavior by running a simple pod config and observed that all the events are generated as expected.
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: busybox 
    image: docker.io/busybox:latest 
    command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep 5']
  restartPolicy: Never
```

server/sandbox_stop_linux.go
```diff
@@ -94,7 +93,6 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error
 
 	log.Infof(ctx, "Stopped pod sandbox: %s", sb.ID())
 	sb.SetStopped(ctx, true)
-	s.generateCRIEvent(ctx, sb.InfraContainer(), types.ContainerEventType_CONTAINER_STOPPED_EVENT)
```
mrunalp, line 97: Does the kubelet expect a stop for pod infra container or does it need a separate pod stop event?

harche: CRI events are generated per container, which also includes infra container. If for some reason infra container runs to completion the `ContainerEventType_CONTAINER_STOPPED_EVENT` will be generated by crio in [handleExit()](https://github.com/cri-o/cri-o/blob/812bc5d664d4ec21c01ff172bce1a2d83f88d5c8/server/server.go#L899) function. This event will be [received and forwarded](https://github.com/kubernetes/kubernetes/blob/8d25cc319391450d9e8b27db0ff8e288da926340/pkg/kubelet/pleg/evented.go#L274) to the syncLoop in Kubelet in the form of `PodLifeCycle` event. 


This PR is fixing the issue where we were incorrectly generating `ContainerEventType_CONTAINER_STOPPED_EVENT` while intentionally stopping the infra container rather than generating it when that infra container runs to completion (for which the relevant code in `handleExit()` already existed as mentioned above)

In retrospect, I feel we should have named `ContainerEventType_CONTAINER_STOPPED_EVENT` as `ContainerEventType_CONTAINER_EXITED_EVENT` because it is generating some confusion with action that is stopping the container (e.g. https://github.com/cri-o/cri-o/pull/6531#issuecomment-1409365765)

harche: @haircommander do you have a reproducer for this scenario?

littlejawa (approved):
lgtm
Thanks!

server/sandbox_stop_linux.go
haircommander: if you create a pod that doesn't share a pod level pid namespace (`shareProcessNamespace: false`) then the infra container will be spoofed

sairameshv: Thanks @haircommander , I added a comment describing this scenario in the code.

sairameshv:
> /retest @sairameshv what's the current status of this PR?

@sohankunkerkar , Thanks for the reminder. I updated the PR today.
I would request @haircommander , @mrunalp, @harche  to have a look whenever available.

Thanks !!

server/sandbox_stop_linux.go
haircommander: @sairameshv can you only call this
`if podInfraContainer.Spoofed()`? 

otherwise, there will be two stopped events when the infra container is not dropped (one from handleExit, and one from here)

sairameshv: Yeah @haircommander , Updated.

haircommander:
/approve

LGTM, thanks @sairameshv , PTAL @cri-o/cri-o-maintainers

PR #6652, opened 2023-02-16
server: fix the permission issue for `/etc`

sohankunkerkar:
Fixes: https://issues.redhat.com/browse/OCPBUGS-6026 
This fixes the folder ownership issue for `/etc` inside the pod.

#### What type of PR is this?

/kind bug

```release-note
None
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -802,13 +802,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	if containerIDMappings == nil {
 		rootPair = idtools.IDPair{UID: 0, GID: 0}
 	}
+
+	etc := filepath.Join(mountPoint, "/etc")
+	// create the `/etc` folder only when it doesn't exist
+	if _, err := os.Stat(etc); err != nil && os.IsNotExist(err) {
+		if err := idtools.MkdirAllAndChown(etc, 0o755, rootPair); err != nil {
+			return nil, fmt.Errorf("error creating mtab directory: %w", err)
+		}
+	}
 	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
 	// compatible with Docker
-	mtab := filepath.Join(mountPoint, "/etc/mtab")
-	if err := idtools.MkdirAllAs(filepath.Dir(mtab), 0o755, rootPair.UID, rootPair.GID); err != nil {
-		return nil, fmt.Errorf("error creating mtab directory: %w", err)
-	}
-	if err := os.Symlink("/proc/mounts", mtab); err != nil && !os.IsExist(err) {
+	if err := os.Symlink("/proc/mounts", filepath.Join(etc, "mtab")); err != nil && !os.IsExist(err) {
 		return nil, err
 	}
 
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -100,6 +100,7 @@ IMAGES=(
     quay.io/crio/oom:latest
     quay.io/crio/redis:alpine
     quay.io/crio/stderr-test:latest
+    quay.io/crio/etc-permission:latest
 )
 
 function img2dir() {
--- a/test/pod.bats
+++ b/test/pod.bats
@@ -318,3 +318,27 @@ function teardown() {
 	crictl stopp "$id"
 	[[ $(find "$CONTAINER_NAMESPACES_DIR" -type f) -eq 0 ]]
 }
+
+@test "pod with the correct etc folder ownership" {
+	start_crio
+	etc_perm_config="$TESTDIR"/container_sleep.json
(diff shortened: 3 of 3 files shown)


test/testdata/container_config_folder_perm.json
```diff
@@ -0,0 +1,57 @@
+{
+	"metadata": {
+		"name": "folder-perm",
+		"attempt": 1
+	},
+	"image": {
+		"image": "quay.io/sohankunkerkar/etc-permission:latest"
```
sohankunkerkar, line 7: we can move this image under `quay.io/crio`

server/container_create_linux.go
```diff
@@ -802,14 +802,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
-	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
-	// compatible with Docker
-	mtab := filepath.Join(mountPoint, "/etc/mtab")
-	if err := idtools.MkdirAllAs(filepath.Dir(mtab), 0o755, rootPair.UID, rootPair.GID); err != nil {
-		return nil, fmt.Errorf("error creating mtab directory: %w", err)
-	}
-	if err := os.Symlink("/proc/mounts", mtab); err != nil && !os.IsExist(err) {
-		return nil, err
+	// create the `/etc` folder only when it doesn't exist
+	if _, err := os.Stat("/etc"); err != nil && os.IsNotExist(err) {
```
haircommander, line 806: You want /etc inside of the container. You need to use `filepath.Join(mountPoint, "/etc")` here

test/testdata/container_config_folder_perm.json
haircommander: instead of having a new json file, can you use jq to generate this?

server/container_create_linux.go
```diff
@@ -802,14 +802,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
-	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
-	// compatible with Docker
-	mtab := filepath.Join(mountPoint, "/etc/mtab")
-	if err := idtools.MkdirAllAs(filepath.Dir(mtab), 0o755, rootPair.UID, rootPair.GID); err != nil {
-		return nil, fmt.Errorf("error creating mtab directory: %w", err)
-	}
-	if err := os.Symlink("/proc/mounts", mtab); err != nil && !os.IsExist(err) {
-		return nil, err
+	// create the `/etc` folder only when it doesn't exist
+	if _, err := os.Stat(filepath.Join(mountPoint, "/etc")); err != nil && os.IsNotExist(err) {
```
haircommander, line 806: maybe we should create a variable for etc , and use it below instead of `filepath.Dir(mtab)`.

server/container_create_linux.go
```diff
@@ -802,14 +802,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+		// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in
+		// the container compatible with Docker
+		mtab := filepath.Join(mountPoint, "/etc/mtab")
```
haircommander, line 809: if we do above, we'd only use mtab once, and not need the variable

server/container_create_linux.go
```diff
@@ -802,14 +802,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+		if err := idtools.MkdirAllAndChown(filepath.Dir(mtab), 0o755, rootPair); err != nil {
+			return nil, fmt.Errorf("error creating mtab directory: %w", err)
+		}
+		if err := os.Symlink("/proc/mounts", mtab); err != nil && !os.IsExist(err) {
+			return nil, err
+		}
```
haircommander, line 815: this should be out of the if, we want to symlink this unconditionally, but conditionally create /etc

haircommander:
/approve

well done with the investigation here!

server/container_create_linux.go
```diff
@@ -802,13 +802,17 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	if containerIDMappings == nil {
 		rootPair = idtools.IDPair{UID: 0, GID: 0}
 	}
+
+	etc := filepath.Join(mountPoint, "/etc")
+	// create the `/etc` folder only when it doesn't exist
+	if _, err := os.Stat(etc); err != nil && os.IsNotExist(err) {
+		if err := idtools.MkdirAllAndChown(filepath.Join(etc, "mtab"), 0o755, rootPair); err != nil {
```
haircommander, line 809: wait 
/hold
 this should be `MkdirAllAndChown(etc...`

sohankunkerkar: Fixed

PR #6668, opened 2023-02-21
Configure systemd install path based on OS

### LINKED ISSUE #6666: Tarball install script should install crio.service file into /etc/systemd/system on CoreOS
### What happened?

When installing cri-o via tarball on CoreOS which installs it by default into `/usr/local` and subdirectories, the systemd unit file `crio.service` is installed into directory `/usr/local/lib/systemd/system`.
Due to CoreOS issue: https://github.com/coreos/rpm-ostree/issues/1936 this file is not picked up by systemd on boot since `/usr/local` on CoreOS is just a symlink to `/var/usrlocal` and `/var` is mounted after systemd already starts.
The end result is that crio doesn't start after OS reboot.
Perhaps the unit file could be installed into directory `/etc/systemd/system` instead? At least on CoreOS?

I expected that enabled crio.service systemd unit is started at boot.

Install crio-o via tarball install script on CoreOS. Reboot OS. See that crio is not started:

```
[root@uniapp ~]# systemctl status crio
○ crio.service - Container Runtime Interface for OCI (CRI-O)
     Loaded: loaded (/usr/local/lib/systemd/system/crio.service; enabled; preset: disabled)
     Active: inactive (dead)
       Docs: https://github.com/cri-o/cri-o
```

_No response_

crio 1.26.1

### OS version

Fedora CoreOS 37.20230122.3.0

VirtualBox in my situation, but could be anywhere CoreOS is possible to install.

hasan4791:
Closes: #6666

#### What type of PR is this?
/kind bug

Handle core-os install path based on the /etc/os-release file

Fixes #6666 

None

```release-note
Update systemd install path for coreos based distribution
(fedora coreos & rhcos)
```

### DIFF
--- a/contrib/bundle/install-paths
+++ b/contrib/bundle/install-paths
@@ -9,8 +9,16 @@ OCIDIR=${OCIDIR:-$PREFIX/share/oci-umount/oci-umount.d}
 BASHINSTALLDIR=${BASHINSTALLDIR:-$PREFIX/share/bash-completion/completions}
 FISHINSTALLDIR=${FISHINSTALLDIR:-$PREFIX/share/fish/completions}
 ZSHINSTALLDIR=${ZSHINSTALLDIR:-$PREFIX/share/zsh/site-functions}
-SYSTEMDDIR=${SYSTEMDDIR:-$PREFIX/lib/systemd/system}
 OPT_CNI_BIN_DIR=${OPT_CNI_BIN_DIR:-/opt/cni/bin}
+
+# Update systemddir based on OS
+source /etc/os-release
+if { [[ "${ID}" == "fedora" ]] && [[ "${VARIANT_ID}" == "coreos" ]]; } ||
+    [[ "${ID}" == "rhcos" ]]; then
+    SYSTEMDDIR=${SYSTEMDDIR:-/etc/systemd/system}
+else
+    SYSTEMDDIR=${SYSTEMDDIR:-$PREFIX/lib/systemd/system}
+fi
 
 SELINUX=
 if selinuxenabled 2>/dev/null; then
--- a/scripts/get
+++ b/scripts/get
@@ -212,8 +212,16 @@ OCIDIR=${OCIDIR:-$PREFIX/share/oci-umount/oci-umount.d}
 BASHINSTALLDIR=${BASHINSTALLDIR:-$PREFIX/share/bash-completion/completions}
 FISHINSTALLDIR=${FISHINSTALLDIR:-$PREFIX/share/fish/completions}
 ZSHINSTALLDIR=${ZSHINSTALLDIR:-$PREFIX/share/zsh/site-functions}
-SYSTEMDDIR=${SYSTEMDDIR:-$PREFIX/lib/systemd/system}
 OPT_CNI_BIN_DIR=${OPT_CNI_BIN_DIR:-/opt/cni/bin}
+
+# Update systemddir based on OS
+source /etc/os-release
+if { [[ "${ID}" == "fedora" ]] && [[ "${VARIANT_ID}" == "coreos" ]]; } ||
+    [[ "${ID}" == "rhcos" ]]; then
+    SYSTEMDDIR=${SYSTEMDDIR:-/etc/systemd/system}
+else
+    SYSTEMDDIR=${SYSTEMDDIR:-$PREFIX/lib/systemd/system}
+fi
 
 SELINUX=
 if selinuxenabled 2>/dev/null; then


TomSweeneyRedHat:
LGTM 
assuming happy tests.

haircommander:
any risk in always putting them in `/etc`

hasan4791:
> any risk in always putting them in `/etc`

Only concern is when updating to newer versions.

haircommander:
I think generally I would prefer having the file be put in the same place across OSs but I agree running the get script to install a new version would leave an old file behind on non-coreos systems. We could potentially remove the old file location, but that is prone to issues too... I am fine with this

/approve

LGTM, thanks @hasan4791, PTAL @cri-o/cri-o-maintainers

contrib/bundle/install-paths
```diff
@@ -9,9 +9,15 @@ OCIDIR=${OCIDIR:-$PREFIX/share/oci-umount/oci-umount.d}
 BASHINSTALLDIR=${BASHINSTALLDIR:-$PREFIX/share/bash-completion/completions}
 FISHINSTALLDIR=${FISHINSTALLDIR:-$PREFIX/share/fish/completions}
 ZSHINSTALLDIR=${ZSHINSTALLDIR:-$PREFIX/share/zsh/site-functions}
-SYSTEMDDIR=${SYSTEMDDIR:-$PREFIX/lib/systemd/system}
 OPT_CNI_BIN_DIR=${OPT_CNI_BIN_DIR:-/opt/cni/bin}
 
+# Update systemdir based on OS
+if grep -qi "coreos" </etc/os-release; then
```
saschagrunert, line 15: I think we should source the file via `. /etc/os-release` and then check for the corresponding `ID` variable.

hasan4791: this grep would be very generic as it looks for coreos string. Also we dont need to worry about specific string needs to be set for ID.

saschagrunert: Honestly I think it's too generic, we don't have a list of possible distributions which may or may not match here. I think we should focus the script on being more robust with respect to the OS ID.

hasan4791: @saschagrunert the ID field indicated the distro name "fedora" which is same for both coreos & non-coreos OS. also the variant is something we need to look but its avaiable only for fedora core-os. I cross verified with the rhcos, and its quite challenging to which variable we need to stick with.

fedora-coreos
```
[core@uniapp ~]$ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="37.20230122.3.0 (CoreOS)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora CoreOS 37.20230122.3.0"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
HOME_URL="https://getfedora.org/coreos/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora-coreos/"
SUPPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
BUG_REPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
SUPPORT_END=2023-11-14
VARIANT="CoreOS"
VARIANT_ID=coreos
OSTREE_VERSION='37.20230122.3.0'
```

rhcos
```
sh-4.4# cat /etc/os-release

ID="rhcos"
ID_LIKE="rhel fedora"
VERSION="412.86.202301061548-0"
VERSION_ID="4.12"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux CoreOS 412.86.202301061548-0 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::coreos"
DOCUMENTATION_URL="https://docs.openshift.com/container-platform/4.12/"
REDHAT_BUGZILLA_PRODUCT_VERSION="4.12"
REDHAT_SUPPORT_PRODUCT_VERSION="4.12"
OPENSHIFT_VERSION="4.12"
RHEL_VERSION="8.6"
OSTREE_VERSION="412.86.202301061548-0"
```
fedora non-coreos
```
[root@fedora ~]# cat /etc/os-release 
NAME="Fedora Linux"
VERSION="37 (Thirty Seven)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora Linux 37 (Thirty Seven)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f37/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
SUPPORT_END=2023-11-14
```

hasan4791: @saschagrunert ^^what do you think?

saschagrunert: Oh, I see. How about using `ID="rhcos"` and alternatively `VARIANT="CoreOS"` if set?

hasan4791: @saschagrunert ptal 👍

hasan4791:
@saschagrunert PTAL

contrib/bundle/install-paths
hasan4791: @saschagrunert any comments about the updated change.

saschagrunert (approved):

saschagrunert:
/skip

This should only affect GitHub actions.

PR #6718, opened 2023-03-09
Extend cpu-c-states.crio.io annotation to add max latency

bartwensley:
#### What type of PR is this?
/kind feature

The cpu-c-states.crio.io annotation currently allows c-states to be disabled for a pod (with guaranteed QoS). The annotation is defined as follows:
cpu-c-states.crio.io: <enable|disable>

This annotation enables or disables c-states for each cpu by writing to the power/pm_qos_resume_latency_us for the cpu under sysfs.

However, users of this new annotation now want to be able to have more control over the c-states for each pod - instead of just disabling c-states completely, they would like to be able to specify a maximum latency for the pod.

The cpu-c-states.crio.io annotation is being extended to support these values:
- enable: enable all c-states (cpu-c-states.crio.io: "enable")
- disable: disable all c-states (cpu-c-states.crio.io: "disable")
- max_latency: enable c-states with a maximum latency in microseconds (for example,  cpu-c-states.crio.io: "max_latency:10")

Note:
- The changes are 100% backwards compatible with the current annotation.

None

None

```release-note
This PR extends the existing cpu-c-states.crio.io annotation in a backwards compatible manner, by adding a new max_latency option. The cpu-c-states.crio.io annotation now supports these values:
- enable: enable all c-states (cpu-c-states.crio.io: "enable")
- disable: disable all c-states (cpu-c-states.crio.io: "disable")
- max_latency: enable c-states with a maximum latency in microseconds (for example,  cpu-c-states.crio.io: "max_latency:10")
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks.go
@@ -70,20 +70,16 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 
 	// Configure c-states for the container CPUs.
 	if configure, value := shouldCStatesBeConfigured(s.Annotations()); configure {
-		log.Infof(ctx, "Configure c-states for container %q to %q", c.ID(), value)
-		switch value {
-		case annotationEnable:
-			// Enable all c-states.
-			if err := setCPUPMQOSResumeLatency(c, "0"); err != nil {
+		maxLatency, err := convertAnnotationToLatency(value)
+		if err != nil {
+			return err
+		}
+
+		if maxLatency != "" {
+			log.Infof(ctx, "Configure c-states for container %q to %q (pm_qos_resume_latency_us: %q)", c.ID(), value, maxLatency)
+			if err := setCPUPMQOSResumeLatency(c, maxLatency); err != nil {
 				return fmt.Errorf("set CPU PM QOS resume latency: %w", err)
 			}
-		case annotationDisable:
-			// Lock the c-state to C0.
-			if err := setCPUPMQOSResumeLatency(c, "n/a"); err != nil {
-				return fmt.Errorf("set CPU PM QOS resume latency: %w", err)
-			}
-		default:
-			return fmt.Errorf("invalid annotation value %s", value)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -659,3 +655,130 @@ func RestoreIrqBalanceConfig(ctx context.Context, irqBalanceConfigFile, irqBanne
...
+//
+// The cpu-c-states.crio.io annotation can be used to control c-states in several ways:
+//
+//	enable: enable all c-states (cpu-c-states.crio.io: "enable")
+//	disable: disable all c-states (cpu-c-states.crio.io: "disable")
+//	max_latency: enable c-states with a maximum latency in microseconds
+//	             (for example,  cpu-c-states.crio.io: "max_latency:10")
+//	cstate.id: enable c-states with a maximum kernel CPU idle level ID
+//	           (for example,  cpu-c-states.crio.io: "cstate.id:1")
+//	cstate.name: enable c-states using a case sensitive name of the maximum C-state allowed
```
TomSweeneyRedHat, line 670: just a nit, you're using "C-state" and "c-state" interchangeably in this section. I'd stick with one, and lean towards "c-state"

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -659,3 +655,130 @@ func RestoreIrqBalanceConfig(ctx context.Context, irqBalanceConfigFile, irqBanne
...
+// It is also possible to specify multiple fallback values separated by '|' as the C-state names and/or
+// IDs may not be available on some systems.
+//
+// Note: this is loosely based on how latencies are specified for the tuned cpu plugin.
+//
+// Examples:
+//
+// cpu-c-states.crio.io: "disable" (disable all c-states)
+// cpu-c-states.crio.io: "enable" (enable all c-states)
+// cpu-c-states.crio.io: "cstate.name:C6|cstate.id:3|max_latency:10" (Use the latency of C6 c-state. If C6 does not exist, use the latency of CPU idle level ID 3 (state3) in sysfs. If this is not found, use a latency of 10us.)
```
TomSweeneyRedHat, line 684: I'm not 100% sure if GitHub/Go will be happy with ths, but here's a suggestion for consideration.

```suggestion
// cpu-c-states.crio.io: "cstate.name:C6|cstate.id:3|max_latency:10" (Use the latency of C6 c-state. If C6 does not exist, use the latency of CPU idle level ID 3 (state3) in sysfs. If this is not found, use a latency of 10μs.)
```

bartwensley: Just for simplicity, I'd prefer not using special characters in the comments. Also "us" is already used for microseconds throughout the code in different identifiers and I'd like to be consistent.

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: Thanks Tom - I will address this nit if I upload an additional set of changes.

bartwensley:
Looks like the tests that are failing are failing for all PRs at the moment. Do I just wait until someone fixed them?

haircommander:
yeah I have opened https://github.com/cri-o/cri-o/issues/6738 to track those issues

internal/runtimehandlerhooks/high_performance_hooks.go
bartwensley: I made this change when I did the rebase just now.

MarSik:
This gives users a very micro-management level of control and will be a pain to document, support and debug. There is also no hw abstraction so if ever we start supporting ARM or AMD for example, the C-state levels might need to be different per node. Also think about heterogenous clusters where not all nodes are equal.

I am OK with including something like power_management_latency:<us> version that will select the proper C-state automatically without exposing the actual C-state (or other mechanism) to the user. I just do not like the exposure of node specific C-states. Also from workload perspective, the customers know and can describe the latency requirements, but they do not always know the future hardware or the node the pod will be scheduled to.

(Btw as a side note, I think an useful extension would be to allow for per-container configuration as not all containers in a pod are typically latency sensitive)

bartwensley:
> This gives users a very micro-management level of control and will be a pain to document, support and debug. There is also no hw abstraction so if ever we start supporting ARM or AMD for example, the C-state levels might need to be different per node. Also think about heterogenous clusters where not all nodes are equal.
> 
> I am OK with including something like power_management_latency:[us] that will select the proper C-state automatically without exposing the actual C-state (or other mechanism) to the user. I just do not like the exposure of node specific C-states. Also from workload perspective, the customers know and can describe the latency requirements, but they do not always know the future hardware or the node the pod will be scheduled to.

Thanks Martin. I have removed the c-state.name and c-state.id annotations as you suggested. The only addition now is the "max_latency:[us]" annotation which will select the appropriate c-state automatically.

MarSik (approved):

bartwensley:
> /approve
> 
> thanks @bartwensley ! can you update the release note to have a bit more information about the change? I will lgtm after that

@haircommander - done. Thanks for reviewing.

haircommander:
/lgtm

thanks!

haircommander:
@bartwensley can you rebase to fix tests?

bartwensley:
> @bartwensley can you rebase to fix tests?

@haircommander - I have rebased and the tests are running

PR #6724, opened 2023-03-10
[WIP] Setting up a per-container imageserver mapping

littlejawa:
#### What type of PR is this?

/kind feature

In order to support the peer-pods use case, cri-o needs to be able to relay PullImage requests (at least), and probably other Image-related requests, to the remove VM used for running the container.

In the current implementation using kata containers, the way to do so is to call a new GRPC entry in the kata shim.
From cri-o's standpoint, that involves calling the runtimeVM instance for the container, when processing PullImage.

This PR provides a way to do that: it introduces a list of ImageServer instances, and provides a way to register a new ImageServer, associated to a containerID.
The current implementation of ImageServer is kept separately, as the "default ImageServer". It will be used for most of the workflow, to limit the overhead outside of peer-pods.

The registration of new ImageServer is done in the call to newRuntimeImpl(), where the runtime implementation is created, and where we can decide (depending on the type of runtime) whether or not a different ImageServer should be used, and link this ImageServer to the runtime.

I also included a very basic implementation of the ImageServer interface as part of the runtimeVM definition. This allows to use the runtimeVM object itself as ImageServer, and will simplify calling the kata shim process when performing image-related operations.
It is currently just a wrapper over the default ImageServer, and is used for testing only.
The registration in newRuntimeImpl() is commented out on purpose, and will be uncommented in a later PR, when the actual implementation of the PullImage forward will be done.

None

NOTE: I have looked at #5624 which does something similar, but the association of ImageServer to runtimeVM is a requirement that it didn't have, and causes additional complexity. Discussing it with @alicefr, we agreed that I should try my own design.

```release-note
None
```


TomSweeneyRedHat:
Any adjustments to man pages needed?

littlejawa:
force-push: fixing lint and vendor issues

littlejawa:
> Any adjustments to man pages needed?

No, I don't think so.
The changes are internal to cri-o. There is no configuration required, nor change in behavior.
This is all preparatory work to be able to plug a different ImageServer in a future PR, but for the moment, I don't think there is anything to document.

littlejawa:
There is a discussion going on regarding how cri-o support for confidential container can happen.
See https://github.com/confidential-containers/community/issues/87

Keeping this on hold as we're exploring other leads.

PR #6726, opened 2023-03-13
Use ImageRef instead of ImageName for restore

adrianreber:
#### What type of PR is this?

/kind bug
/kind feature

Up until now CRI-O was using ImageName to remember the base image of the restored container. This works, but has two problems:

 * The output from the restored container was different, especially for the Kubernetes use case. The output of 'crictl ps' contains for the original container under 'IMAGE' something like 'registry/path/container@sha256:123444444...'. The restored container was, however, only displaying something like 'registry/path/container'.

 * More problematic, however, was that CRI-O might pull the wrong image from the registry.  If the container in the registry was updated (new latest tag for example) all of a sudden the wrong base image would be downloaded and the restore would potentially fail.

With this change the output of Kubernetes containers in 'crictl ps' 'IMAGE' does not change after restoring the container and the restored container will run on exactly the same base image.

None

```release-note
None
```

### DIFF
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -153,6 +153,8 @@ func (c *ContainerServer) prepareCheckpointExport(ctr *oci.Container) error {
 	config := &metadata.ContainerConfig{
 		ID:              ctr.ID(),
 		Name:            ctr.Name(),
+		RootfsImage:     ctr.Image(),
+		RootfsImageRef:  ctr.ImageRef(),
 		RootfsImageName: ctr.ImageName(),
 		CreatedTime:     ctr.CreatedAt(),
 		OCIRuntime: func() string {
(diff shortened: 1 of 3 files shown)


adrianreber:
Hmm, I had a look over the CI failures and it does not seem to be related to my changes. Not totally sure, however.

umohnani8:
LGTM
/retest

haircommander:
/approve

LGTM

TomSweeneyRedHat:
LGTM
once the tests are hip

adrianreber:
> LGTM once the tests are hip

Unfortunately it is not clear to me if the failures are related to my change. I looked at a couple of the test failures and it seems unrelated, but hard to tell for me.

haircommander:
I'm pretty sure they're unrelated https://github.com/cri-o/cri-o/issues/6738

haircommander:
/lgtm

thanks!

PR #6737, opened 2023-03-15
tests: use crictl binary directly when checking its capabilities

littlejawa:
#### What type of PR is this?

/kind failing-test

The timeout.bats tests are all skipped with the message `must have a crictl with the -T option to test CRI-O's timeout handling`.

Looking into it, I think it started with https://github.com/cri-o/cri-o/commit/5cad5f287482c045ac7c6c1aa70a1352636b843a, where the `crictl()` wrapper was modified to use a config file in the `TESTDIR` folder.
Because of that, calling `crictl()` before `setup_test()` means calling it with a non-existing config file, causing an error.

This PR modifies the test to use the crictl binary directly, skipping the wrapper function.

None

```release-note
None
```

### DIFF
--- a/test/timeout.bats
+++ b/test/timeout.bats
@@ -3,7 +3,9 @@
 load helpers
 
 function setup() {
-	if ! crictl runp -h | grep -q "cancel-timeout"; then
+	# do not use the crictl() wrapper function here: we need to test the crictl
+	# features with no additional arg.
+	if ! "$CRICTL_BINARY" runp -h | grep -q "cancel-timeout"; then
 		skip "must have a crictl with the -T option to test CRI-O's timeout handling"
 	fi
 	setup_test
--- a/test/timeout.bats
+++ b/test/timeout.bats
@@ -33,6 +33,22 @@ EOF
 	export CONTAINER_CONMON="$TESTDIR/tmp_conmon"
 }
 
+function create_pinns() {
+	local timeout=$1
+
+	cat > "$TESTDIR"/tmp_pinns << EOF
+#!/bin/bash
+if [[ "\$1" != "--version" ]]; then
+    echo "Delaying pinns by $timeout"
+	sleep $timeout
+fi
+$PINNS_BINARY_PATH \$@
+EOF
+	chmod +x "$TESTDIR/tmp_pinns"
+
+	export CONTAINER_PINNS_PATH="$TESTDIR/tmp_pinns"
+}
+
 # Allow cri-o to catch up. The sleep here should be less than
 # resourcestore.sleepTimeBeforeCleanup but enough for cri-o to
 # finish processing cancelled crictl create/runp.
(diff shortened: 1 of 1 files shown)


littlejawa:
It seems there is a different behavior now than when the test was written.

The tests that expect `crictl runp -T [timeout] [...]` to fail are broken. This command just returns the created pod ID - in other word, it just works.

On the other hand, the tests that create a pod and then run `crictl create -T [timeout] [podid] [...]` are passing. The command fails as expected, with the `"context deadline exceeded"` message.

I'm not sure what has changed, if it is on crio, crictl, or in the test itself. I've tried modifying the timeout (increasing or decreasing), with no effect. If I understand the test logic, the pod creation is just too quick (?).
Not sure where to look at for the moment.

haircommander:
Thanks for taking this on @littlejawa 

A reason the pod timeout could not be triggered but container timeouts are is it's possible we're not actually creating the infra container as intended. I find myself wondering if we're successfully setting the drop_infra_ctr option. We could run a 
`$CRIO_STATUS_BINARY --socket "$CRIO_SOCKET"` before the `runp`s to see if we correctly set the `drop_infra_ctr` option to false.

littlejawa:
After verification: the flag `drop_infra_ctr` is set to `false` during the test, as expected.

I'm trying to understand the test itself : what would make `runp` timeout? How long is it to create the pod without a container in it?
I feel that when it succeeds, it should be very quick, so the timeout can't trigger.
Should we add something to the test to make the pod creation wait somehow?

haircommander:
> I'm trying to understand the test itself : what would make runp timeout? How long is it to create the pod without a container in it?

it could be network creation. pod creation creates namespaces (which calls pinns, and should be fast), calls into cni to create the network, and creates some files.

littlejawa:
> it could be network creation. pod creation creates namespaces (which calls pinns, and should be fast), calls into cni to create the network, and creates some files.

Ok. So I think I know how to fix that.

We can do with "pinns" what is done with "conmon" : replace it for the time of the test by a script that delays the call (with a sleep). This seems to fix the issue (at least when testing locally - waiting on CI result).

I'm changing "pinns" this way only for the pod timeout tests, not for all tests like we do for conmon.
The rationale is that if we make the pod creation timeout, then the container timeout tests will succeed because the pod timeout, which tells nothing about what we do on actual container timeout behaviour.
Not sure if that's relevant, but it seemed cleaner.

Tell me if you think we should just call "create_pinns" as part of the setup.

littlejawa:
Ok - so it failed on "metrics container oom" for two integration jobs, but timeout tests have passed.
I don't think this failure is related to my change, right?

haircommander:
> Tell me if you think we should just call "create_pinns" as part of the setup.

Very clever @littlejawa great idea!

> Ok - so it failed on "metrics container oom" for two integration jobs, but timeout tests have passed. I don't think this failure is related to my change, right?

correct that's a known flake

/approve
/hold cancel
@cri-o/cri-o-maintainers PTAL

LGTM

PR #6744, opened 2023-03-17
OCPBUGS-11396,OCPNODE-1536: high performance hooks: update to a cgroup approach for load balance disabling

haircommander:
#### What type of PR is this?

/kind feature
on newer kernels, the sysfs that was previously used to disable cpu load balacing has been moved to read only. As such, a cgroupfs approach needs to be used.

The general approach is CRI-O will set cpuset.sched_load_balance to disable for all containers that specify the load balance annotation. It is the responsibility of other entities on the node to make sure the cgroup hierarchy is configured to ensure the kernel will actually disable load balancing 

```release-note
Update high performance hooks to disable load balancing through cgroups, rather than directly through the kernel's sysfs
```


haircommander:
/hold

this requires some work to verify it actually works as expected e2e

haircommander:
> I hope this can help.

Thanks! CPU load balancing is not expected to work on RHEL 9, this patch will add support for it (though there are some other pieces that are needed)

haircommander:
I've also tossed in a fix for an issue that causes the `cpu-quota` annotation to fail to be respected. On newer systemds, it expects a scope property be set to set cpu quota. Since runc will do that anyway, and we know at container creation that we want to disable cpu quota, I've moved the handling of the annotation there. PTAL @ffromani @MarSik

server/container_create_linux.go
```diff
@@ -408,6 +409,9 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		if resources != nil {
 			specgen.SetLinuxResourcesCPUPeriod(uint64(resources.CpuPeriod))
 			specgen.SetLinuxResourcesCPUQuota(resources.CpuQuota)
+			if runtimehandlerhooks.ShouldCPUQuotaBeDisabled(sb.Annotations()) {
```
MarSik, line 412: Notice the original place actually checked that it is disabling quota for a guaranteed QoS class container with dedicated cpus. This code does not. So it makes a pretty big security hole by disabling resource accounting for burstable and best effort containers as well.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -259,50 +217,25 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
 		return fmt.Errorf("find container %s CPUs", c.ID())
 	}
 
-	cpus, err := cpuset.Parse(lspec.Resources.CPU.Cpus)
-	if err != nil {
-		return err
+	if node.CgroupIsV2() {
```
MarSik, line 220: Will this cause a container launch failure when the annotation is present on a system with cgroups v2? That would be a serious regression if that is the end results.

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -259,50 +217,25 @@ func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string)
...
-		if err != nil {
-			return err
-		}
+	cpusetPath, ok := controllers["cpuset"]
+	if !ok {
+		return fmt.Errorf("failed to get cpuset of container %s", c.ID())
 	}
 
-	return nil
+	return cgroups.WriteFile("/sys/fs/cgroup/cpuset"+cpusetPath, "cpuset.sched_load_balance", "0")
```
MarSik, line 238: This needs much better docs right here explaining that all other cpusets must lose access to the cpus or this does not work. Where is that part going to be handled? Just by setting up the slices properly?

MarSik (changes requested):

internal/runtimehandlerhooks/high_performance_hooks.go
MarSik: Or maybe it is actually a good thing and we will fix it later when the v2 support lands.

browsell: There is no support for this functionality in v2.

haircommander: unfortunately this needs to be the case. The cpuset controller for cgroup2 needs some modifications to work with kubernetes because of the unified hierarchy and new knobs that control this

server/container_create_linux.go
haircommander: good catch. I updated the condition to check for those

internal/runtimehandlerhooks/high_performance_hooks.go
haircommander: added documentation to the top of the function!

internal/runtimehandlerhooks/high_performance_hooks.go
MarSik: Yep, but maybe the error message should be a bit more explicit about what the issue is.

haircommander: updated, is that better @MarSik ?

MarSik: Much better, thanks.

ffromani (commented):
/lgtm

PR #6750, opened 2023-03-21
server: wire support for userns volumes

giuseppe:
#### What type of PR is this?

/kind feature

it passes down to the OCI runtime the mappings configuration as specified from the Kubelet.  It is necessary for supporting user namespaces (KEP-127)

```release-note
CRI-O honors the per-volume mappings specified by the Kubelet
```


server/container_create_linux.go
```diff
@@ -866,6 +868,21 @@ func setupWorkingDirectory(rootfs, mountLabel, containerCwd string) error {
 	return nil
 }
 
+func getOCIMappings(m []*types.IDMapping) []rspec.LinuxIDMapping {
```
haircommander, line 871: nit: can the function be defined below when it's called

haircommander:
test failures aren't related

server/container_create_linux.go
giuseppe: sure, done now

haircommander:
you may need to rebase @giuseppe

giuseppe:
> you may need to rebase @giuseppe

rebased

PR #6765, opened 2023-03-28
OCPBUGS-10970: Fix the interrupt mask width when encoding

MarSik:
#### What type of PR is this?

/kind bug

Kernel got more strict in parsing the interrupt cpu mask strings, we are no longer allowed to pass in more 4B (32 bit) words than necessary to cover all cpus on a given system.

See https://bugzilla.redhat.com/show_bug.cgi?id=2181546 for more detail about what kernel accepts and what gets rejected.

The logic of mapByteToHexChar (from internal/runtimehandlerhooks/utils.go) was rounding to 8 bytes (64 bits) which triggered the linked bug.

Fixes https://issues.redhat.com/browse/OCPBUGS-10970

```release-note
None
```

### DIFF
--- a/internal/runtimehandlerhooks/utils_test.go
+++ b/internal/runtimehandlerhooks/utils_test.go
@@ -48,6 +48,14 @@ var _ = Describe("Utils", func() {
 				input:    Input{cpus: "4-13", mask: "ffff,ffffc00f", set: true},
 				expected: Expected{mask: "0000ffff,ffffffff", invMask: "00000000,00000000"},
 			}),
+			Entry("clear a single bit that was one when odd mask is present", TestData{
+				input:    Input{cpus: "9", mask: "fff", set: false},
+				expected: Expected{mask: "00000dff", invMask: "0000f200"},
+			}),
+			Entry("clear two bits from a short mask", TestData{
+				input:    Input{cpus: "2-3", mask: "ffffff", set: false},
+				expected: Expected{mask: "00fffff3", invMask: "0000000c"},
+			}),
 		)
 
 		Context("UpdateIRQBalanceConfigFile", func() {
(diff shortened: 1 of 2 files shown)


internal/runtimehandlerhooks/utils_test.go
```diff
@@ -48,6 +48,14 @@ var _ = Describe("Utils", func() {
...
 				expected: Expected{mask: "0000ffff,ffffffff", invMask: "00000000,00000000"},
 			}),
+			Entry("clear a single bit that was one when odd mask is present", TestData{
+				input:    Input{cpus: "9", mask: "fff", set: false},
+				expected: Expected{mask: "00000dff", invMask: "0000f200"},
+			}),
+			Entry("clear two bits from a short mask", TestData{
+				input:    Input{cpus: "2-3", mask: "ffffff", set: false},
+				expected: Expected{mask: "00fffff3", invMask: "0000000c"},
+			}),
```
ffromani, line 58: do we have a test that checks the mask is not bigger than it is supposed to be (per the [very nice!]) comment you added in utils.go

MarSik: This very test (short mask) should cover that case I hope. I am not running an exhaustive check though..

MarSik:
I do not think this is my fault...

code: RESOURCE_ERROR
  location: /deployments/ci-op-843jlf83-37133-vpc/resources/ci-op-843jlf83-37133-network
  message: "{\"ResourceType\":\"compute.v1.network\",\"ResourceErrorCode\":\"QUOTA_EXCEEDED\"\
    ,\"ResourceErrorMessage\":\"Quota 'NETWORKS' exceeded.  Limit: 300.0 globally.\"\
    }"

TomSweeneyRedHat:
Once the test are hip, LGTM

haircommander:
@ffromani LGTY?

internal/runtimehandlerhooks/utils_test.go
ffromani: should be a good enough start

ffromani (commented):
/lgtm

ffromani:
> @ffromani LGTY?

yep! added

PR #6772, opened 2023-03-29
stop using KUBE-MARK-MASQ for hostport chains

danwinship:
#### What type of PR is this?
/kind bug

TL;DR: cri-o's hostport code expects that some other component will create a working `KUBE-MARK-MASQ` chain, but this is no longer guaranteed starting in kube 1.27.

[KEP-3178](https://github.com/kubernetes/enhancements/issues/3178) removes the creation of the `KUBE-MARK-MASQ` iptables chain from kubelet, such that it is now only created by kube-proxy, and is considered private to kube-proxy. (["Kubernetes’s IPTables Chains Are Not API"](https://kubernetes.io/blog/2022/09/07/iptables-chains-not-api/)). (Related: f61332ed96e8b59f68c3da557b47eab6f2c3182d :unamused:)

I had checked a while back to make sure cri-o was going to be OK with this change, and somehow convinced myself that it was... I think I got confused by the fact that [_one_ of its masquerading rules doesn't use `KUBE-MARK-MASQ`](https://github.com/cri-o/cri-o/blob/v1.26.2/internal/hostport/hostport.go#L162) and didn't notice that [another one does](https://github.com/cri-o/cri-o/blob/v1.26.2/internal/hostport/hostport_manager.go#L141).

Even worse, [cri-o actually creates the `KUBE-MARK-MASQ` chain itself if it doesn't exist](https://github.com/cri-o/cri-o/blob/v1.26.2/internal/hostport/meta_hostport_manager.go#L24), meaning that in an environment where KEP-3178 is enabled (eg, kube 1.27 or later), and kube-proxy is not running (eg, clusters using ovn-kubernetes), then cri-o will create a `KUBE-MARK-MASQ` chain that doesn't work but then later it will create rules that expect that it _does_ work, and may also end up hiding other components that were still depending on `KUBE-MARK-MASQ` too. :grimacing: 

None

~(Draft because it still needs testing, though the overall shape should be correct.)~

I am going to rewrite this code entirely in the near future to use nftables rather than iptables (since iptables is deprecated in RHEL 9 and will go away in RHEL 10) but I'm not sure when that will be, and this fix is needed for OCP 4.14 (and it may end up blocking the 1.27 rebase; at a minimum the e2e hostport tests will fail). But anyway, point being, while there is additional refactoring that bould be done here (like renaming the other `KUBE-` chains to `CRIO-`) it may not be worthwhile since that would require additional backward-compatibility and testing too, and the code may only last one release before being replaced again?

```release-note
None
```


haircommander:
/cc @aojea 

if you'd be so kind, as you moved the hostport code here

aojea:
> /cc @aojea
> 
> if you'd be so kind, as you moved the hostport code here

bookmarked

internal/hostport/meta_hostport_manager_test.go
```diff
@@ -203,36 +203,50 @@ func TestMetaHostportManager(t *testing.T) {
 
 	lines := strings.Split(raw.String(), "\n")
 	expectedLines := map[string]bool{
-		`*nat`:                              true,
```
danwinship, line 206: (best viewed with [gear icon] → Hide whitespace)

internal/hostport/hostport_manager.go
```diff
@@ -121,42 +119,54 @@ func (hm *hostportManager) Add(id string, podPortMapping *PodPortMapping, natInt
...
 				"-m", "comment", "--comment", fmt.Sprintf(`"%s hostport %d"`, podFullName, pm.HostPort),
 				"-m", protocol, "-p", protocol, "-d", pm.HostIP,
 				"-j", "DNAT", fmt.Sprintf("--to-destination=%s", hostPortBinding))
 		}
+
+		// SNAT hairpin traffic. There is no "ctorigaddrtype" so we can't
+		// _exactly_ match only the traffic that was definitely DNATted by our
+		// rule as opposed to someone else's. But if the traffic has been DNATted
+		// and has src=dst=podIP then _someone_ needs to masquerade it, and the
+		// worst case here is just that "-j MASQUERADE" gets called twice.
```
aojea, line 163: didn't have an issue in the past because masquerading twice? I remember you were the one that solved that issue, but I do not know if is the same case

aojea: however, this seems very unlikely to happen

aojea:
lgtm
some comment for a typo, but this code is tricky, we have e2e tests that covers this, so we have some assurance there too.

The number additional iptables rules should not be a problem, hostports is not a widely used feature so I can't see any scale problem

internal/hostport/hostport_manager.go
danwinship: ah, yes (https://github.com/kubernetes/kubernetes/pull/92035). But that only happened with `-j MASQUERADE --random-fully`, which we don't use here. (And was also caused by a kernel bug which has been fixed for a while now.)

aojea:
> The latter _could_ be tested but currently isn't.

https://github.com/kubernetes/kubernetes/blob/4f6714ef42dcdb1b35bb2d8581c1ef83d59f02d1/test/e2e/network/hostport.go#L119-L122

;)

danwinship:
> https://github.com/kubernetes/kubernetes/blob/4f6714ef42dcdb1b35bb2d8581c1ef83d59f02d1/test/e2e/network/hostport.go#L119-L122

OK, so we do test localhost, good. But I was confused anyway; this PR doesn't affect localhost, because [that already didn't use KUBE-MARK-MASQ](https://github.com/cri-o/cri-o/blob/v1.26.2/internal/hostport/hostport.go#L162). This PR only affects the hairpin case, which AFAICT kube e2e does not test?

aojea:
> OK, so we do test localhost, good. But I was confused anyway; this PR doesn't affect localhost, because [that already didn't use KUBE-MARK-MASQ](https://github.com/cri-o/cri-o/blob/v1.26.2/internal/hostport/hostport.go#L162). This PR only affects the hairpin case, which AFAICT kube e2e does not test?

I was more worried about breaking hostport in general, this only changes one thing but with iptables 🤷 , yeah, AFAIK we don't test hairpin for hostport, that is a bit weird case, right?

Issue #6751, opened 2023-03-21
e2e and integration tests are failing


sohankunkerkar:
### Which jobs are failing?

e2e-{aws,gcp}-ovn
{fedora,cgroupv2}-integration
{cgroupv2,rhel,crun,conmonrs}-e2e

### Which tests are failing?

one of the [upstream sig-network tests](https://github.com/kubernetes/kubernetes/pull/116232) is failing for all e2e jobs and the following bats tests are failing for {fedora,cgroupv2}-integration:
```
not ok 169 setup_file failed
# (from function `setup_file' in test file ./irqbalance.bats, line 16)
#   `if ! command -v irqbalance > /dev/null; then' failed with status 0
ok 174 metrics with default port # skip Metrics port 9090 not listening
ok 175 metrics with random port
ok 176 metrics with operations quantile
ok 177 secure metrics with random port
ok 178 secure metrics with random port and missing cert/key
not ok 179 metrics container oom
# (in test file ./metrics.bats, line 117)
#   `METRIC=$(curl -sf "http://localhost/:$PORT/metrics" | grep '^container_runtime_crio_containers_oom_total')' failed
```

### Since when has it been failing?

Since two weeks

### Testgrid link

e2e failure: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/6750/pull-ci-cri-o-cri-o-main-ci-cgroupv2-e2e/1638178250037399552

integration tests failure: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/6750/pull-ci-cri-o-cri-o-main-ci-cgroupv2-integration/1638178252549787648

### Reason for failure (if possible)

All e2e jobs failed due to https://github.com/kubernetes/kubernetes/pull/116232 and the integration tests failure is still unknown, it needs more digging.

sohankunkerkar:
I don't think the failure for `e2e-{aws-gcp}-ovn` is related to cri-o
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/6752/pull-ci-cri-o-cri-o-main-e2e-gcp-ovn/1638313267560452096

sohankunkerkar:
[This](https://github.com/cri-o/cri-o/pull/6777) PR fixed the integration tests failure

saschagrunert:
Closing this, feel free to reopen or creating a new issue if required.

PR #6777, opened 2023-03-30
test: run irqbalance tests serially to avoid race condition

sohankunkerkar:
Addresses the integration tests failure for https://github.com/cri-o/cri-o/issues/6751

#### What type of PR is this?

/kind ci
/kind flake

```release-note
None
```

### DIFF
--- a/test/test_runner.sh
+++ b/test/test_runner.sh
@@ -29,8 +29,11 @@ function execute() {
 # Tests to run. Default is "." (i.e. the current directory).
 TESTS=("${@:-.}")
 
-# The number of parallel jobs to execute
+# The number of parallel jobs to execute tests
 export JOBS=${JOBS:-$(($(nproc --all) * 4))}
 
 # Run the tests.
 execute bats --jobs "$JOBS" --tap "${TESTS[@]}"
+# Set this var to run irqbalance tests
+export TEST_SERIAL="Yes"
+execute bats --tap ./irqbalance.bats
--- a/test/irqbalance.bats
+++ b/test/irqbalance.bats
@@ -84,6 +84,9 @@ function teardown() {
 #	and the mask must have value stated in "IRQBALANCE_BANNED_CPUS" field
 #   from irqbalance service configuration file.
 @test "irqbalance cpu ban list save" {
+	if [[ -z "$TEST_SERIAL" ]]; then
+		skip "test must be run serially"
+	fi
 	# given
 	if ! grep -Eq '^[1,3,7,f]{1,}$' /proc/irq/default_smp_affinity; then
 		skip "requires default IRQ smp affinity (not banned CPUs)"
@@ -120,6 +123,9 @@ function teardown() {
 #	and save the mask value in "IRQBALANCE_BANNED_CPUS" field
 #   of irqbalance service configuration file.
 @test "irqbalance cpu ban list restore - default" {
+	if [[ -z "$TEST_SERIAL" ]]; then
+		skip "test must be run serially"
+	fi
 	# given
 	if ! grep -Eq '^[1,3,7,f]{1,}$' /proc/irq/default_smp_affinity; then
 		skip "requires default IRQ smp affinity (not banned CPUs)"
@@ -153,6 +159,9 @@ function teardown() {
 #	so it reads banned cpus mask from "IRQBALANCE_BANNED_CPUS" field
 #   and save it in a file pointer by "BANNEDCPUS_CONF"  env var
 @test "irqbalance cpu ban list restore - disable and file missing" {
+	if [[ -z "$TEST_SERIAL" ]]; then
+		skip "test must be run serially"
+	fi
 	# given
 	if ! grep -Eq '^[1,3,7,f]{1,}$' /proc/irq/default_smp_affinity; then
 		skip "requires default IRQ smp affinity (not banned CPUs)"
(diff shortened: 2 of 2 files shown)


test/serial/irqbalance.bats
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bats
 # vim: set syntax=sh:
 
-load helpers
+load ../helpers
```
haircommander, line 4: another option if this doesn't work: you could call `skip` in the `setup()` if an env variable isn't present and then set the env variable on the second bats call

haircommander: kind of like `TEST_USERNS` handling

test/test_runner.sh
```diff
@@ -27,10 +27,13 @@ function execute() {
...
+# Serial tests
+SERIAL_TESTS=("${TESTS_ROOT[@]}/serial")
 
-# The number of parallel jobs to execute
+# The number of parallel jobs to execute non-serial tests
 export JOBS=${JOBS:-$(($(nproc --all) * 4))}
 
 # Run the tests.
-execute bats --jobs "$JOBS" --tap "${TESTS[@]}"
+execute bats --jobs "$JOBS" --tap "${TESTS_ROOT[@]}"
```
mrunalp, line 38: execute replaces the current binary with bats so we can't use bash syntax to combine with other invocation or split this out to the next line. Instead, can we just invoke bats without the exec?

sohankunkerkar: Fixed

test/serial/irqbalance.bats
jlojosnegros: > another option if this doesn't work: you could call `skip` in the `setup()` if an env variable isn't present and then set the env variable on the second bats call

it works I tested it yesterday

test/test_runner.sh
```diff
@@ -27,10 +27,13 @@ function execute() {
...
+# Serial tests
+SERIAL_TESTS=("${TESTS_ROOT[@]}/serial")
 
-# The number of parallel jobs to execute
+# The number of parallel jobs to execute non-serial tests
 export JOBS=${JOBS:-$(($(nproc --all) * 4))}
 
 # Run the tests.
-execute bats --jobs "$JOBS" --tap "${TESTS[@]}"
+bats --jobs "$JOBS" --tap "${TESTS_ROOT[@]}"
```
jlojosnegros, line 38: I think the return value of test_runner.sh should take into account the result of both test sets ... 

somethink like
```suggestion
# Run the tests.
bats --jobs "$JOBS" --tap "${TESTS_ROOT[@]}"
parallel_test_result=$?
bats --tap "${SERIAL_TESTS[@]}"
serial_test_result=$?

[ parallel_test_result && serial_test_result ]
```

test/test_runner.sh
```diff
@@ -27,10 +27,13 @@ function execute() {
...
-TESTS=("${@:-.}")
+TESTS_ROOT=("${@:-.}")
+# Serial tests
+SERIAL_TESTS=("${TESTS_ROOT[@]}/serial")
 
-# The number of parallel jobs to execute
+# The number of parallel jobs to execute non-serial tests
 export JOBS=${JOBS:-$(($(nproc --all) * 4))}
 
 # Run the tests.
```
jlojosnegros, line 37: I think the return value of test_runner.sh should take into account the result of both test sets ... 

somethink like
```suggestion
# Run the tests.
bats --jobs "$JOBS" --tap "${TESTS_ROOT[@]}"
parallel_test_result=$?
bats --tap "${SERIAL_TESTS[@]}"
serial_test_result=$?

[ parallel_test_result && serial_test_result ]
```

test/test_runner.sh
```diff
@@ -27,10 +27,13 @@ function execute() {
...
-execute bats --jobs "$JOBS" --tap "${TESTS[@]}"
+bats --jobs "$JOBS" --tap "${TESTS_ROOT[@]}"
+bats --tap "${SERIAL_TESTS[@]}"
```
jlojosnegros, line 39: I think the return value of test_runner.sh should take into account the result of both test sets ... 

somethink like
```suggestion
# Run the tests.
bats --jobs "$JOBS" --tap "${TESTS_ROOT[@]}"
parallel_test_result=$?
bats --tap "${SERIAL_TESTS[@]}"
serial_test_result=$?

[ parallel_test_result && serial_test_result ]
```

jlojosnegros:
It seems `test_runner.sh` is run with the variable `TESTFLAGS` as parameter in some places. 
That variable sometimes contain `.bat` filenames to be executed, so this approach would not work as it is.
Right now `test_runner.sh` just handle its parameters as the test collection to execute which can be a single file, a list of files or a folder.

test/test_runner.sh
```diff
@@ -21,16 +21,12 @@ fi
...
 TESTS=("${@:-.}")
 
-# The number of parallel jobs to execute
+# The number of parallel jobs to execute tests
 export JOBS=${JOBS:-$(($(nproc --all) * 4))}
 
 # Run the tests.
-execute bats --jobs "$JOBS" --tap "${TESTS[@]}"
+bats --jobs "$JOBS" --tap "${TESTS[@]}"
+TEST_SERIAL="Yes" bats --tap ./irqbalance.bats
```
haircommander, line 32: does calling `execute` twice not work?

sohankunkerkar: I haven't tried but just followed [this](https://github.com/cri-o/cri-o/pull/6777#discussion_r1153819894) comment

test/test_runner.sh
haircommander: @mrunalp I believe `execute` was just a function we defined to run time and put the stderr into stdout, I don't think we need to drop it (though having two separate executes for the two  lines seems like a good idea)

haircommander:
/approve

LGTM @cri-o/cri-o-maintainers PTAL

Issue #6774, opened 2023-03-30
Release notes for patch releases are wrong


saschagrunert:
The release notes generation is wrong for patches: https://github.com/cri-o/cri-o/actions/runs/4561549708/jobs/8047639568

Its using:
```
level=info msg="Using start tag v1.25.0"
level=info msg="Using end tag v1.26.0"
```

But in this case it should use `v1.26.2` as start and `v1.26.3` as end tag.

PR #6778, opened 2023-03-30
Fix the release version for release notes

sohankunkerkar:
Fixes https://github.com/cri-o/cri-o/issues/6774

#### What type of PR is this?

/kind bug

```release-note
None
```


internal/version/version.go
```diff
@@ -248,3 +250,15 @@ func (i *Info) JSONString() (string, error) {
...
+// GetLatestReleaseVersion fetches the latest github release
+// version from cri-o.
+func GetLatestReleaseVersion(token string) (string, error) {
+	ctx := context.Background()
+	client := github.NewTokenClient(ctx, token)
+	tags, _, err := client.Repositories.ListTags(ctx, "cri-o", "cri-o", nil)
+	if err != nil {
+		return "", err
+	}
+	return tags[0].GetName(), nil
```
haircommander, line 263: I think we should do something to make sure the minor version matches the branch we're currently in. Sometimes multiple releases are done in a day and this would mean we'd need to be more strategic about when those happen

sohankunkerkar: ok, with the latest push, I'm evaluating the release `Version` and checking the number of releases after that. Then picking the last release tag from the sorted list.

sohankunkerkar: But this change needs to get backported to all the major releases.

internal/version/version.go
```diff
@@ -248,3 +250,27 @@ func (i *Info) JSONString() (string, error) {
 	}
 	return string(b), nil
 }
+
+// GetLatestReleaseVersion fetches the latest github release
+// version from cri-o.
+func GetLatestReleaseVersion(token string) (string, error) {
```
haircommander, line 256: this seems like something we want in scripts/release-notes instead. otherwise, we're suddenly linking the github dependencies in cri-o for something we're only using as a cri-o utility

internal/version/version.go
```diff
@@ -248,3 +250,27 @@ func (i *Info) JSONString() (string, error) {
...
+	ctx := context.Background()
+	client := github.NewTokenClient(ctx, token)
+	tags, _, err := client.Repositories.ListTags(ctx, "cri-o", "cri-o", nil)
+	if err != nil {
+		return "", err
+	}
+	rng, err := semver.ParseRange(fmt.Sprintf(">= %s", Version))
```
haircommander, line 263: can there be some comments about this and the next couple of steps? it's not immediately obvious what's happening here

haircommander:
/approve

@saschagrunert for the LGTM?

haircommander:
you need to update [the vendor ](https://github.com/cri-o/cri-o/actions/runs/4622350615/jobs/8174925566?pr=6778)

saschagrunert (approved):
Thank you for the work on this! 

/lgtm

PR #6780, opened 2023-03-31
Allow SELinux unconfined containers

saschagrunert:
#### What type of PR is this?

/kind feature

We now let process labels starting with `unconfined_u:unconfined_r:unconfined_t` be applied by CRI-O.

Refers to https://github.com/openshift/oc/pull/842#issuecomment-1491051478

For discussion @mrunalp @haircommander 

```release-note
Allow SELinux process labels being applied starting with `unconfined_u:unconfined_r:unconfined_t`.
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -279,7 +279,8 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 
 	mountLabel := containerInfo.MountLabel
 	var processLabel string
-	if !ctr.Privileged() {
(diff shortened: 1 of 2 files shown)


travier:
This might need a new rule in the policy to allow that domain transition.

travier:
https://github.com/containers/container-selinux/blob/main/container.te#L652 > This would suggest that this is already allowed but not sure.

rhatdan:
You do not need to transition the user or the role.

system_u:system_t:unconfined_t:MCS is fine.

This should be allowed.

saschagrunert:
This is not required any more due to https://github.com/containers/container-selinux/releases/tag/v2.208.0

PR #6784, opened 2023-04-03
Enforce umask on container and sandbox creation

saschagrunert:
#### What type of PR is this?

/kind bug

Sometimes CRI-O runs as umask 0000 but starts with the default one 0022. We now double check that we have the right umask enforced on sandbox and container creation to mitigate that issue.

None

None

```release-note
Enforce CRI-O default umask 0022 even on sandbox and container creation.
```


saschagrunert:
@cri-o/cri-o-maintainers PTAL, this addresses https://issues.redhat.com/browse/OCPBUGS-10704

haircommander:
it's not clear to me why we need this every pod and container creation. It feels like this is hiding a bug somewhere else

PR #6795, opened 2023-04-05
internal/storage.runtimeService.createContainerOrPodSandbox(): read ID maps correctly

nalind:
#### What type of PR is this?

/kind cleanup

When creating a storage library `Container`, read the ID mappings that are assigned to that container from the `Container` structure that `CreateContainer()` returns, rather than expecting the library to modify the options structure that we passed to `CreateContainer()`.  It will stop doing that in its 1.46 release.

```release-note
none
```


TomSweeneyRedHat (commented):
LGTM
once the tests get hip

Issue #5896, opened 2022-05-19
tests: the "runtime()" wrapper can't be used to test with RUNTIME_TYPE=vm


littlejawa:
### What happened?

The helpers.bat script defines a wrapper function "runtime()", used to call the runtime binary directly.
It assumes a runc-style interface for the parameters.

With RUNTIME_TYPE=vm, cri-o uses the containerd shim v2 interface (gRPC) to talk to the runtime binary. The command line interface is then not available directly form the binary, and the "runtime" wrapper fails.

### What did you expect to happen?

Calling the runtime binary directly is necessary in some tests to find discrepancies between the status cri-o has internally, and the actual status of the container as seen from the runtime itself. It is also used to access the container directly (for instance: simulate a crash or a reboot by killing a container, or deleting its associated files from the filesystem).

All of these need to be possible with a "vm"-type runtime.

### How can we reproduce it (as minimally and precisely as possible)?

The following integration tests are failing when run with kata containers (using RUNTIME_TYPE=vm):

- drop_infra.bats:
  - @test "test infra ctr dropped"
  - @test "test infra ctr not dropped"

- crio-wipe.bats:
  - @test "internal_wipe eventually cleans network on forced restart of crio if network is slow to come up"

- workloads.bats:
  - @test "test workload gets configured to defaults"
  - @test "test workload can override defaults"
  - @test "test workload should not be set if annotation not specified"

- network.bats:
  - @test "Clean up network if pod sandbox gets killed"

Other test files are using the runtime() wrapper in multiple places:
- timeout.bats
- cgroups.bats
- restore.bats
- ctr.bats
- inspect.bats

### Anything else we need to know?

I am currently looking for a way to retrieve the same information from a shim runtime, in a way that would not be kata-specific.
This is still work in progress.

I have identified 4 different use cases for the runtime() wrapper:

- runtime kill
Used in network.bats and crio-wipe.bats.

"runc kill" will send the SIGTERM signal to the container, forcing it to stop.
It is used in the test script to simulate the crash of a container and verify that cri-o correctly cleans up.

- runtime list
Used in cgroups.bats, drop_infra.bats and timeout.bats.

This command is used to compare the list of containers running from the runtime, compared to internal status of cri-o.
It is needed when we expect a desynchronization of the two states.
For instance: starting a pod with a timeout on creation time, cri-o will abort and ignore the pod before it is actually started, so the container may still be running in the background.
Checking that it is not the case can only be done by asking the runtime itself.

[...]

PR #6804, opened 2023-04-11
WIP - Fix "runtime list" calls from helpers.bash

littlejawa:
#### What type of PR is this?

/kind failing-test

This PR starts an implementation of the missing features in the runtime() wrapper function in helpers.bats, when dealing with kata containers.

It modifies the wrapper to call a separate script, rather than the actual runtime, when using a container with RUNTIME_TYPE=vm. This new script will implement the various use cases of the runtime() wrapper for kata containers.

In this PR, we only implement the "runtime list" use case. Additional use cases will be added with more PRs.

The "runtime list" call is implemented, and another one "runtime_pid" is added to address a very specific need.
Several of the tests have been adapted accordingly.

These test will be enabled in the kata CI job when this PR is merged.

See #5896 for more information

None

Not marking #5896 as fixed here, as this is just part of the full implementation needed to fix it.

```release-note
None
```

### DIFF
--- a/test/kata_cmdline_tool.bash
+++ b/test/kata_cmdline_tool.bash
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+# In helpers.bats, the "runtime()" wrapper is used to call the runtime
+# binary directly. It is used for 4 commands across the test files:
+#
+# list:
+#   Provides the list of containers running under the runtime.
+#   In this implementation, the output comes from a call to "ps" and contains
+#   the PID and command line used to run the kata process.
+#   with the "-q" parameter, the list is limited to the container ID of running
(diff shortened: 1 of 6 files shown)


test/kata_cmdline_tool.bash
```diff
@@ -0,0 +1,44 @@
...
+
+"kill") ;;
+
+"delete") ;;
+
+"state") ;;
+
+*)
+    echo "Not supported command: $1"
+    ;;
```
haircommander, line 42: I think these should probably exit non zero until they're implemented

littlejawa: Fixed. Thanks!

PR #6812, opened 2023-04-13
OCPBUGS-11777,OCPNODE-1536: [1.26] high performance hooks: update to a cgroup approach for load balance disabling

haircommander:
#### What type of PR is this?

/kind feature
on newer kernels, the sysfs that was previously used to disable cpu load balacing has been moved to read only. As such, a cgroupfs approach needs to be used.

The general approach is CRI-O will set cpuset.sched_load_balance to disable for all containers that specify the load balance annotation. It is the responsibility of other entities on the node to make sure the cgroup hierarchy is configured to ensure the kernel will actually disable load balancing 

```release-note
Update high performance hooks to disable load balancing through cgroups, rather than directly through the kernel's sysfs
```


haircommander:
@saschagrunert @sohankunkerkar @jrvaldes PTAL

jrvaldes (approved):
/lgtm

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -222,38 +174,13 @@ func annotationValueDeprecationWarning(annotation string) string {
...
-			}
-			return false, err
-		}
-		return true, nil
-	})
-}
-
-func setCPUSLoadBalancing(c *oci.Container, enable bool, schedDomainDir string) error {
+// disableCPULoadBalancing relies on the cpuset cgroup to disable load balancing for containers.
+// The requisite condition to allow this is `cpuset.sched_load_balance` field must be set to 0 for all cgroups
```
TomSweeneyRedHat, line 178: nit
```suggestion
// The requisite condition to allow this is the `cpuset.sched_load_balance` field must be set to 0 for all cgroups
```

PR #6817, opened 2023-04-16
Add basic crictl info config with sandboxImage

### LINKED ISSUE #6816: Information about pause image requested in runtime status
Since kubernetes 1.27.0, `kubeadm` is now looking for information about the pause image configuration of the runtime.

* https://github.com/kubernetes/kubernetes/issues/106893

```
getting status of runtime: failed to template data: template: tmplExecuteRawJSON:1:9: executing \"tmplExecuteRawJSON\" at <.config.sandboxImage>: map has no entry for key \"config\"
```

It was hardcoded, to look in the containerd config. So now all other container runtimes needs to implement that field.

---

The user can do `crictl info -q`, if just wanting the old status.

With the new info, `kubeadm` will warn if cri-o has not been configured:

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#override-pause-image-cri-o

Without the info, there will be the "info" output (above) but no "warn".

afbjorklund:
#### What type of PR is this?

/kind feature

Adds information about pause image to `crictl info`, as used by kubeadm in 1.27

Fixes #6816

```release-note
a new field "config" has been added to the verbose output of crictl info
```

`crictl info`
```json
{
  "status": {
    "conditions": [
      {
        "type": "RuntimeReady",
        "status": true,
        "reason": "",
        "message": ""
      },
      {
        "type": "NetworkReady",
        "status": false,
        "reason": "NetworkPluginNotReady",
        "message": "Network plugin returns error: No CNI configuration file in /etc/cni/net.d/. Has your network provider started?"
      }
    ]
  },
  "config": {
    "sandboxImage": "registry.k8s.io/pause:3.9"
  }
}

```

### DIFF
--- a/server/runtime_status.go
+++ b/server/runtime_status.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"golang.org/x/net/context"
@@ -27,12 +28,31 @@ func (s *Server) Status(ctx context.Context, req *types.StatusRequest) (*types.S
 		networkCondition.Message = fmt.Sprintf("Network plugin returns error: %v", err)
 	}
 
-	return &types.StatusResponse{
+	resp := &types.StatusResponse{
 		Status: &types.RuntimeStatus{
 			Conditions: []*types.RuntimeCondition{
 				runtimeCondition,
 				networkCondition,
 			},
 		},
-	}, nil
+	}
+	if req.Verbose {
+		info, err := s.createRuntimeInfo()
+		if err != nil {
+			return nil, fmt.Errorf("creating runtime info: %w", err)
+		}
+		resp.Info = info
+	}
+	return resp, nil
+}
+
+func (s *Server) createRuntimeInfo() (map[string]string, error) {
+	config := map[string]interface{}{
+		"sandboxImage": s.config.ImageConfig.PauseImage,
+	}
+	bytes, err := json.Marshal(config)
+	if err != nil {
+		return nil, fmt.Errorf("marshal data: %w", err)
+	}
+	return map[string]string{"config": string(bytes)}, nil
 }
--- a/server/runtime_status_test.go
+++ b/server/runtime_status_test.go
@@ -46,5 +46,16 @@ var _ = t.Describe("Status", func() {
 				Expect(condition.Status).To(BeTrue())
 			}
 		})
+
+		It("should return info as part of a verbose response", func() {
+			// When
+			response, err := sut.Status(context.Background(),
+				&types.StatusRequest{Verbose: true})
+
+			// Then
+			Expect(err).To(BeNil())
+			Expect(response).NotTo(BeNil())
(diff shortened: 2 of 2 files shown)


server/runtime_status.go
```diff
@@ -27,12 +28,31 @@ func (s *Server) Status(ctx context.Context, req *types.StatusRequest) (*types.S
...
+		info, err := createRuntimeInfo(s)
+		if err != nil {
+			return nil, fmt.Errorf("creating runtime info: %w", err)
+		}
+		resp.Info = info
+	}
+	return resp, nil
+}
+
+func createRuntimeInfo(s *Server) (map[string]string, error) {
```
saschagrunert, line 49: I'd say we either pass the `s.config` or we make it a method of `*Server`.

server/runtime_status.go
```diff
@@ -27,12 +28,31 @@ func (s *Server) Status(ctx context.Context, req *types.StatusRequest) (*types.S
...
+	config := map[string]interface{}{
+		"sandboxImage": s.config.ImageConfig.PauseImage,
+	}
+	bytes, err := json.Marshal(config)
+	if err != nil {
+		return nil, fmt.Errorf("marshal data: %v: %w", config, err)
```
saschagrunert, line 55: I don't think we need the whole config in the error message, but I'm not blocking on that.

saschagrunert (commented):
Thank you, just two nits otherwise LGTM.

server/runtime_status.go
afbjorklund: I was trying to do the same as with the other verbose "info" fields, for containers and images

server/runtime_status.go
afbjorklund: Maybe just pass the pause image, since it is an internal function anyway ?

afbjorklund: It is a bit of a mix:
```go
func (s *Server) createContainerInfo(container *oci.Container) (map[string]string, error)
func createImageInfo(result *pkgstorage.ImageResult) (map[string]string, error)
```

afbjorklund:
```go
func (s *Server) createRuntimeInfo() (map[string]string, error)
```

afbjorklund:
I don't think there is any requirement or definition on "config", other than what kubeadm is looking for...

https://github.com/kubernetes/kubernetes/pull/115610/commits/d12b4d4c6772bc7735c3765d79ecc46c69595088

saschagrunert (approved):
LGTM, ptal @cri-o/cri-o-maintainers 

/hold
because of auto approval. Maintainers feel free to unhold if it looks good to y'all.

afbjorklund:
Maybe a unit test would improve coverage, but I don't really understand the format of it. Seems both tests are the same ?

saschagrunert:
Yeah that's known

/override ci/prow/e2e-gcp-ovn

haircommander:
/unhold
/lgtm

thanks!

PR #6842, opened 2023-04-26
Support `RegistryUnavailable` type

saschagrunert:
#### What type of PR is this?

/kind feature

If we agree on the fix in https://github.com/kubernetes/kubernetes/pull/117612, then we can support the `RegistryUnavailable` error message as well. This has the benefit that we display the information directly to the end user as better error indicator.

/hold

None
None

```release-note
Support `RegistryUnavailable` error on container image pull.
```


saschagrunert:
/unhold

@cri-o/cri-o-maintainers PTAL

server/image_pull.go
```diff
@@ -93,6 +94,10 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 	}
 
 	if pullOp.err != nil {
+		if errors.Is(pullOp.err, syscall.ECONNREFUSED) {
+			// from https://github.com/kubernetes/kubernetes/blob/806b301/pkg/kubelet/images/types.go#L40-L41
+			return nil, errors.New("RegistryUnavailable")
```
haircommander, line 99: I am more in favor of updating vendor and using the variable directly. is that an option?

saschagrunert: We can do that, but it requires an OTEL bump since we're too far ahead of time with our dependencies. 

Ref: https://github.com/kubernetes/kubernetes/pull/117743

saschagrunert: Looks like we do not see an update in the near future because it introduces a huge amount of transient dependencies. Going for an override for now

saschagrunert:
@haircommander PTAL again

haircommander:
thank you!

/lgtm

Issue #6657, opened 2023-02-18
Networking metrics missing with cri-o 1.26


rkojedzinszky:
### What happened?

Upgraded cri-o from 1.25 to 1.26, and observed that container_network_receive_bytes_total per-pod metric (and perhaps others) has disappeared.

### What did you expect to happen?

Expected this metric to be present after the upgrade.

### How can we reproduce it (as minimally and precisely as possible)?

Have a 1.26 kubernetes cluster with cri-o-1.26, and check that kubelet/cadvisor does not provide metrics container_network_receive_bytes_total for each container.

### CRI-O and Kubernetes version

<details>

```console
# crio --version
crio version 1.26.1
Version:        1.26.1
GitCommit:      36d50b214b422753e9b33cc1b04e8b9de9e20419
GitCommitDate:  2023-02-09T06:17:48Z
GitTreeState:   clean
BuildDate:      2023-02-09T06:14:46Z
GoVersion:      go1.19.5
Compiler:       gc
Platform:       linux/arm64
Linkmode:       dynamic
BuildTags:      
  containers_image_ostree_stub
  apparmor
  exclude_graphdriver_btrfs
  btrfs_noversion
  exclude_graphdriver_devicemapper
  libdm_no_deferred_remove
  containers_image_openpgp
  seccomp
  selinux
LDFlags:          -s -w -X github.com/cri-o/cri-o/internal/pkg/criocli.DefaultsPath="" -X github.com/cri-o/cri-o/internal/version.buildDate=2023-02-09T06:14:46Z 
SeccompEnabled:   true
AppArmorEnabled:  false
Dependencies:     
  
```

```console
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.0", GitCommit:"b46a3f887ca979b1a5d14fd39cb1af43e7e5d12d", GitTreeState:"clean", BuildDate:"2022-12-08T19:58:30Z", GoVersion:"go1.19.4", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
[...]

Issue #6349, opened 2022-11-07
1.24.3 leaks pidns with cgroup v2


rkojedzinszky:
### What happened?

Observed that on a node where many cronjobs get scheduled (i.e.many pods being created than stopped), syslog entries appear:
```
Nov 07 08:22:18 k8s-node16 kubelet[699]: W1107 08:22:18.605965     699 container.go:589] Failed to update stats for container "/pids/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-podeb779fdd_e230_4fb0_b027_f9f91f117b46.slice/crio-f1800404321299f38a4ba5dfedf95df94b825507597527b48bf3a6d9f2f3eb13.scope": unable to determine device info for dir: /var/lib/containers/storage/overlay/304bfd8df6d35aaf9a5632b9bcdc69bdd54aa9277baf2c18c4930072d017514c/diff: stat failed on /var/lib/containers/storage/overlay/304bfd8df6d35aaf9a5632b9bcdc69bdd54aa9277baf2c18c4930072d017514c/diff with error: no such file or directory, continuing to push stats
```

That is because entries under `/sys/fs/cgroup/pids/kubepods.slice/kubepods-burstable.slice/` are not cleaned up. I suspect the count of `crictl pods` and entries under the mentioned dir should be in sync, but I may be wrong.

```
# crictl pods
POD ID              CREATED             STATE               NAME                     NAMESPACE           ATTEMPT             RUNTIME
6b309d1ef5223       13 minutes ago      Ready               kube-router-dr2vh        kube-system         0                   (default)
41b1338d583b2       13 minutes ago      Ready               truenas-csi-node-5lkh5   truenas-csi         0                   (default)
b12fd8fb24ccb       13 minutes ago      Ready               promtail-vfflj           monitoring          0                   (default)
950e109f187a0       13 minutes ago      Ready               node-exporter-km99c      kube-system         0                   (default)
# ls -la /sys/fs/cgroup/pids/kubepods.slice/kubepods-burstable.slice/
total 0
drwxr-xr-x 11 root root 0 Nov  7 08:13 .
drwxr-xr-x  3 root root 0 Nov  7 08:13 ..
-r--r--r--  1 root root 0 Nov  7 08:13 cgroup.controllers
-r--r--r--  1 root root 0 Nov  7 08:13 cgroup.events
-rw-r--r--  1 root root 0 Nov  7 08:13 cgroup.freeze
[...]

PR #6856, opened 2023-04-28
cgmgr: create cgroups for systemd cgroup driver for dropped infra pods

haircommander:
The history here is a bit convoluted. Originally, runc created the cgroup for the infra container.
    cAdvisor was built to assume the cgroup for the infra container would be created, and it uses
    this to find the network metrics for the pod. When we dropped the infra container, cri-o needed to make this
    cgroup so cAdvisor could still find the network metrics.
    
    However, systemd didn't like the way we did it, and would remove the cgroup mid pod creation, which was fixed in
    https://github.com/cri-o/cri-o/pull/6196. This actually caused the cgroup to not be created at all, which then caused
    the networking metrics to not be gathered at all.
    
    Thus, we do need to create a cgroup underneath the systemd cgroup. Attempt to use a slice for this, as systemd won't require a
    process be underneath it.

#### What type of PR is this?

/kind bug
reverts https://github.com/cri-o/cri-o/pull/6196 
fixes https://github.com/cri-o/cri-o/issues/6657 

```release-note
Fix a bug where network metrics collection is broken with systemd cgroup driver and dropped infra containers.
```

### DIFF
--- a/internal/config/cgmgr/cgmgr.go
+++ b/internal/config/cgmgr/cgmgr.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/cri-o/cri-o/internal/config/node"
 	libctr "github.com/opencontainers/runc/libcontainer/cgroups"
+	libctrCgMgr "github.com/opencontainers/runc/libcontainer/cgroups/manager"
+	cgcfgs "github.com/opencontainers/runc/libcontainer/configs"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
@@ -72,6 +74,9 @@ type CgroupManager interface {
 	// CreateSandboxCgroup takes the sandbox parent, and sandbox ID.
 	// It creates a new cgroup for that sandbox, which is useful when spoofing an infra container.
 	CreateSandboxCgroup(sbParent, containerID string) error
+	// RemoveSandboxCgroup takes the sandbox parent, and sandbox ID.
+	// It removes the cgroup for that sandbox, which is useful when spoofing an infra container.
+	RemoveSandboxCgroup(sbParent, containerID string) error
 }
 
 // New creates a new CgroupManager with defaults
@@ -158,4 +163,42 @@ func MoveProcessToContainerCgroup(containerPid, commandPid int) error {
 		}
 	}
 	return nil
+}
+
+// createSandboxCgroup takes the path of the sandbox parent and the desired containerCgroup
+// It creates a cgroup through cgroupfs (as opposed to systemd) at the location cgroupRoot/sbParent/containerCgroup.
+func createSandboxCgroup(sbParent, containerCgroup string) error {
+	cg := &cgcfgs.Cgroup{
+		Name:   containerCgroup,
+		Parent: sbParent,
+		Resources: &cgcfgs.Resources{
+			SkipDevices: true,
+		},
+	}
+	mgr, err := libctrCgMgr.New(cg)
+	if err != nil {
+		return err
+	}
+
+	return mgr.Apply(-1)
+}
+
+func removeSandboxCgroup(sbParent, containerCgroup string) error {
+	cg := &cgcfgs.Cgroup{
+		Name:   containerCgroup,
+		Parent: sbParent,
+		Resources: &cgcfgs.Resources{
+			SkipDevices: true,
+		},
+	}
+	mgr, err := libctrCgMgr.New(cg)
+	if err != nil {
+		return err
+	}
+
+	return mgr.Destroy()
+}
+
+func containerCgroupPath(id string) string {
+	return crioPrefix + "-" + id
 }
--- a/internal/config/cgmgr/cgmgr.go
+++ b/internal/config/cgmgr/cgmgr.go
@@ -180,7 +180,23 @@ func createSandboxCgroup(sbParent, containerCgroup string) error {
 		return err
 	}
 
-	return mgr.Apply(-1)
+	if err := mgr.Apply(-1); err != nil {
+		return err
+	}
+
+	// The reasoning for this code is slightly obscure. In situation where CPU load balancing is desired,
+	// all cgroups must either have cpuset.sched_load_balance=0 or they should not have an intersecting cpuset
+	// with the set that load balancing should be disabled on.
+	// When this cgroup is created, it is easiest to set sched_load_balance to 0, especially because there will
+	// not be any processes in this cgroup (or else we wouldn't need to call this).
+	if !node.CgroupIsV2() {
+		path := mgr.Path("cpuset")
+		if path == "" {
+			return fmt.Errorf("failed to find cpuset for newly created cgroup")
+		}
+		return libctr.WriteFile(path, "cpuset.sched_load_balance", "0")
+	}
+	return nil
 }
 
 func removeSandboxCgroup(sbParent, containerCgroup string) error {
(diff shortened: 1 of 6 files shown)


haircommander:
/hold

I think there are some cgroupfs issues here and I haven't tested thoroughly with cgroupv2 yet either

internal/config/cgmgr/cgmgr.go
```diff
@@ -72,6 +74,9 @@ type CgroupManager interface {
 	// CreateSandboxCgroup takes the sandbox parent, and sandbox ID.
 	// It creates a new cgroup for that sandbox, which is useful when spoofing an infra container.
 	CreateSandboxCgroup(sbParent, containerID string) error
+	// CreateSandboxCgroup takes the sandbox parent, and sandbox ID.
```
kolyshkin, line 77: typo: `s/CreateSandboxCgroup/RemoveSandboxCgroup/`.

Perhaps we should use a linter to catch this.  Revive would do (alas enabling it for the whole repo will cause tons of warnings)

internal/config/cgmgr/cgmgr.go
```diff
@@ -159,3 +164,39 @@ func MoveProcessToContainerCgroup(containerPid, commandPid int) error {
...
+		Resources: &cgcfgs.Resources{
+			SkipDevices: true,
+		},
+	}
+	mgr, err := libctrCgMgr.New(cg)
+	if err != nil {
+		return err
+	}
+
+	return mgr.Apply(-1)
```
kolyshkin, line 181: Well this does not make much sense to me.

1. Since you haven't specified `UseSystemd: true`, you are using fs manager. It just creates a bunch of directories.
2. It's a scope, so its lifetime is bound to the existence of at least one process in the scope. As we don't have any processes, systemd may just remove this.
3. I am not sure that having a `name=systemd` controller in there helps with (2), but overall this smells bad -- we create something that looks very much like what systemd would create, but systemd knows nothing about it.

Let's step back. Here is my limited understanding of the problem.

We have a pod (which is a scope in systemd terms), and a bunch of containers in it. We used to have a dummy infra container running a single pause process in order for cadvisor to identify the pod and get its stats. Now we don't have an infra container, and other containers may come and go, so there's no container that will always exist in the pod, and as a result cadvisor loses the pod (stops reporting its stats).

If what's written above is correct, than this is a problem with the approach of referring to a pod via a particular container which is in it. Containers can come and go,  and a pod should be identified by a pod cgroup path or something like that, not via any container.

kolyshkin (changes requested):

internal/config/cgmgr/cgmgr.go
haircommander: > so there's no container that will always exist in the pod, and as a result cadvisor loses the pod (stops reporting its stats)

Sort of. cadvisor assumes there's an infra container and an associated cgroup. the only way it registers network metrics is if there's a cgroup associated with the pod container. We can trick it to gather network metrics for the pod by pretending like there's the infra container by creating this dummy cgroup. I agree it smells bad--I've been wanting to change this interation for years with CRI stats KEP, but it's been delayed a number of times.

addressing the points more directly:
> Since you haven't specified UseSystemd: true, you are using fs manager. It just creates a bunch of directories.

yes this is the point. libcontainer will create the empty cgroup, and cadvisor (Which as inotify on the cgroup hierarchy) will catch its creation, see the pod ID, and then ask cri-o about that pod

> It's a scope, so its lifetime is bound to the existence of at least one process in the scope. As we don't have any processes, systemd may just remove this.

I was wondering whether this will happen. I believe this is the cause of the issue requiring https://github.com/cri-o/cri-o/pull/6196. I am not sure if this PR will help that or not, I am hoping it wil..

> I am not sure that having a name=systemd controller in there helps with (2), but overall this smells bad -- we create something that looks very much like what systemd would create, but systemd knows nothing about it.

agreed, though I'm not sure we can achieve anything better in the meantime. Folks need their network metrics, and we want systemd to not remove the cgroup. if we can achieve both of those internally in cri-o, it will be worth it for now, while we work on fixing it more legitimately

haircommander:
after a conversation with @kolyshkin offline, we opted to have crio create a slice for the infra container, as systemd won't require a process be underneath it.

test/pod.bats
```diff
@@ -278,22 +278,17 @@ function teardown() {
 }
 
 @test "kubernetes pod terminationGracePeriod passthru" {
-	[ -v CIRCLECI ] && skip "runc v1.0.0-rc11 required" # TODO remove this
+	if [[ "$CONTAINER_CGROUP_MANAGER" != "systemd" ]]; then
+		skip "need systemd cgroup manager"
+	fi
 	# Make sure there is no XDG_RUNTIME_DIR set, otherwise the test might end up using the user instance.
 	# There is an assumption in the test to use the system instance of systemd (systemctl show).
```
kolyshkin, line 285: nit: this comment is now misplaced

internal/config/cgmgr/systemd.go
```diff
@@ -197,8 +198,41 @@ func convertCgroupFsNameToSystemd(cgroupfsName string) string {
...
+		logrus.Infof("Not creating sandbox cgroup: sbParent is empty")
+		return nil
+	}
+	return &cgcfgs.Cgroup{
+		Name:   containerCgroupPath(containerID) + ".slice",
+		Parent: convertCgroupFsNameToSystemd(sbParent),
+		Resources: &cgcfgs.Resources{
+			SkipDevices: true,
+		},
+		Systemd: true,
```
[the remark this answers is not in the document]
haircommander: we basically do this now by having two versions of this function. that's the pattern we've adopted for branching between the two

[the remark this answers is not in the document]
kolyshkin: Ah, sorry, this *IS* systemd cgroup manager, so it's always `true`.

test/cgroups.bats
```diff
@@ -72,6 +72,28 @@ EOF
...
+	CONTAINER_DROP_INFRA_CTR=true start_crio
+
+	jq --arg cg "$parent" '	  .linux.cgroup_parent = $cg' \
+		"$TESTDATA"/sandbox_config.json > "$TESTDIR"/sandbox_config_slice.json
+
+	pod_id=$(crictl runp "$TESTDIR"/sandbox_config_slice.json)
+	ls "$cgroup_base"/"$parent"/crio-"$pod_id"*
+
+	crictl rmp -fa
+	! ls "$cgroup_base"/"$parent"/crio-"$pod_id"*
```
kolyshkin, line 94: Offtopic, but that reminds I recently learned that this form won't work, unless this is the last statement in a function. I'll send a fix.

kolyshkin:
So, I would like to have it tested before the merge, to make sure
 - the issue (#6657) is solved
 - the older bug (#6349) is not reintroduced

kolyshkin (approved):
LGTM

haircommander:
/hold

to double check 

>     * the issue ([Networking metrics missing with cri-o 1.26 #6657](https://github.com/cri-o/cri-o/issues/6657)) is solved
> 
>     * the older bug ([1.24.3 leaks pidns with cgroup v2 #6349](https://github.com/cri-o/cri-o/issues/6349)) is not reintroduced

haircommander:
/hold cancel

* the issue (https://github.com/cri-o/cri-o/issues/6657) is solved 

I now see the network metrics

* the older bug ( https://github.com/cri-o/cri-o/issues/6349) is not reintroduced

I don't see a pid cgroup being created. I think using libcontainer has fixed this

I also have monitored the cgroup for a bit and i don't see systemd cleaning it up. I think we're all set

kolyshkin (approved):
/lgtm

Issue #6835, opened 2023-04-20
Support Pinned Images


mrunalp:
Add functionality to specify pinned images in crio.conf to request the kubelet not to gc those images.

[pinned_images]
quay.io/pause:4.4
acme.io/k8s/*

It should support globbing/regex to specify a group of images.

mrunalp:
@sohankunkerkar is working this.

PR #6862, opened 2023-05-01
*: add support for pinned_images in crio configuration

sohankunkerkar:
Fixes: https://github.com/cri-o/cri-o/issues/6835

#### What type of PR is this?

/kind feature

Yes

```release-note
Add support for pinned_images in crio configuration
With this change, users can indicate to the kubelet not to perform garbage collection on certain images.
```


server/image_list.go
```diff
@@ -23,9 +26,15 @@ func (s *Server) ListImages(ctx context.Context, req *types.ListImagesRequest) (
...
+	filteredImagesList, err := FilterPinnedImages(results, s.config.PinnedImages)
+	if err != nil {
+		return nil, err
+	}
+
 	resp := &types.ListImagesResponse{}
-	for i := range results {
-		image := ConvertImage(&results[i])
+	for i := range filteredImagesList {
+		image := ConvertImage(&filteredImagesList[i])
```
haircommander, line 37: CRI has a [pinned](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L1404) field already. I think  we should continue to show the kubelet the pinned images but make sure we add the pinned field to them

server/image_list.go
```diff
@@ -70,3 +79,62 @@ func ConvertImage(from *storage.ImageResult) *types.Image {
...
 	return to
 }
+
+// FilterImageResults filters out the pinned images from the given list of image results
+func FilterPinnedImages(images []storage.ImageResult, pinnedImages []string) ([]storage.ImageResult, error) {
+	if len(pinnedImages) == 0 {
+		return images, nil
+	}
+	filteredImages := make([]storage.ImageResult, 0, len(images))
+	regexps, err := compileRegexpsForPinnedImages(pinnedImages)
```
haircommander, line 89: can these regexes be compiled on crio startup and stored so we don't need to recompile each time we list?

haircommander: they can be stored inside of the config similarly to how we do the seccomp profile or runtime handlers

internal/storage/image.go
```diff
@@ -873,3 +880,43 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
...
 	return is, nil
 }
+
+// FilterPinnedImage checks if the give image needs to be pinned
+// to exclude from kubelet's image GC.
+func FilterPinnedImage(image string, config *config.Config) bool {
+	if len(config.PinnedImages) == 0 {
+		return false
+	}
+	regexps := compileRegexpsForPinnedImages(config.PinnedImages)
```
haircommander, line 890: these should be saved in the imageService structure instead of compiled each time we pull

internal/storage/image.go
```diff
@@ -830,7 +838,7 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 // which will prepend the passed-in DefaultTransport value to an image name if
 // a name that's passed to its PullImage() method can't be resolved to an image
 // in the store and can't be resolved to a source on its own.
-func GetImageService(ctx context.Context, sc *types.SystemContext, store storage.Store, defaultTransport string, insecureRegistries []string) (ImageServer, error) {
+func GetImageService(ctx context.Context, store storage.Store, imageConfig *config.Config) (ImageServer, error) {
```
haircommander, line 841: nit: this is more the serverConfig

internal/storage/image.go
```diff
@@ -873,3 +883,43 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
...
+		if pinnedImage.MatchString(image) {
+			return true
+		}
+	}
+	return false
+}
+
+// CompileRegexpsForPinnedImages compiles regular expressions for the given
+// list of pinned images.
+func CompileRegexpsForPinnedImages(patterns *config.Config) []*regexp.Regexp {
```
haircommander, line 904: again, this is the server config. Since this only needs PinnedImages, maybe we should jsut pass that into this function instead of the whole config? will be easier to unit test that way

haircommander:
couple more nits, and it would be really cool to unit test CompileRegexpsForPinnedImages to have some examples of what we expect to work, otherwise LGTM

haircommander:
/approve

LGTM, nicely done

@cri-o/cri-o-maintainers PTAL

pkg/config/config.go
```diff
@@ -469,6 +469,11 @@ type ImageConfig struct {
 	// PauseCommand is the path of the binary we run in an infra
 	// container that's been instantiated using PauseImage.
 	PauseCommand string `toml:"pause_command"`
+	// PinnedImages is a list of container images that should be pinned
+	// and not subject to garbage collection by kubelet.
+	// Pinned images will remain in the container runtime's storage until
+	// they are manually removed. Default value: empty list (no images pinned)
+	PinnedImages []string `toml:"pinned_images"`
```
saschagrunert, line 476: We should mention in the docs that those are regular expressions and how it behaves. Having examples would be great as well.

internal/storage/image.go
```diff
@@ -873,3 +883,43 @@ func GetImageService(ctx context.Context, sc *types.SystemContext, store storage
...
+	return false
+}
+
+// CompileRegexpsForPinnedImages compiles regular expressions for the given
+// list of pinned images.
+func CompileRegexpsForPinnedImages(patterns []string) []*regexp.Regexp {
+	regexps := make([]*regexp.Regexp, 0, len(patterns))
+	for _, pattern := range patterns {
+		var re *regexp.Regexp
+		switch {
```
saschagrunert, line 908: Hm, I’ll not block on this but I’d prefer having plain globs or regular expressions without any further checks.

pkg/config/template.go
```diff
@@ -1324,6 +1329,12 @@ const templateStringCrioImagePauseCommand = `# The command to run to have a cont
 
 `
 
+const templateStringCrioImagePinnedImages = `# List of container images to be skipped from the kubelet's garbage collection.
```
saschagrunert, line 1332: non blocking: can we make it work with config reload via `systemctl reload crio`?

pkg/config/config.go
sohankunkerkar: The names of the images provided under [pinned_image] are not exactly regular expressions. While users may enter patterns that resemble regular expressions, our code is designed to handle them in a suitable manner. This is mainly a backend logic and not a user-facing change.

sohankunkerkar: How about this one?
```
The `pinned_images` field allows specifying image names using either exact, glob, or keyword patterns. Exact matches must match the entire name, glob matches can have a wildcard * at the end, and keyword matches can have wildcards on both ends.
```

saschagrunert: Sounds good!

sohankunkerkar:
fixed the validate-docs warning: https://github.com/cri-o/cri-o/actions/runs/4895446554/jobs/8741105904?pr=6862

internal/storage/image.go
```diff
@@ -249,13 +255,14 @@ func (svc *imageService) buildImageCacheItem(systemContext *types.SystemContext,
 			return imageCacheItem{}, err
 		}
 	}
-
+	name, _, _ := sortNamesByType(image.Names)
```
mtrmac, line 258: How is this supposed to work with images that have multiple names?

sohankunkerkar: Are you asking how `sortNamesByType` will behave with multiple names?

mtrmac: I did read that code. I don’t see that the current implementation provides a predictable behavior for users (the order of `image.Names`, in the current implementation, seems to depend on the order in which image names were pulled, and that’s _not at all promised by the API_).

My first guess is that all names in `image.Names` should be processed by the filter, but I didn’t look into the use cases of pinned images.

haircommander: > My first guess is that all names in image.Names should be processed by the filter, but I didn’t look into the use cases of pinned images.

the idea is to disqualify an image for kubelet garbage collection. Since the order of the names isn't guaranteed, it does sound like each should be considered to see if the image should be disqualified

PR #6880, opened 2023-05-05
go.{mod,sum}, vendor: update NRI.

klihub:
#### What type of PR is this?

/kind bug

Update `github.com/containerd/nri` dependency to latest. Current code assumes that any resource update has a non-nil runtime-spec.LinuxMemory field. Tries to blindly dereference it in `internal/lib/container_server.go` without a nil-check. This is not a problem for internally generated `LinuxResources`, because the implicit assumption there is true. NRI used to violate this assumption and the latest commit fixes that. Without that fix an NRI plugin updating a container without touching memory resources/cgroupfs parametrization triggers a nil-pointer dereference/SIGSEGV panic and a crash in CRI-O. 

None

```release-note
None
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/containerd/containerd v1.7.1
 	github.com/containerd/cri-containerd v1.19.0
 	github.com/containerd/fifo v1.1.0
-	github.com/containerd/nri v0.3.0
+	github.com/containerd/nri v0.3.1-0.20230504231226-94185418e253
 	github.com/containerd/ttrpc v1.2.2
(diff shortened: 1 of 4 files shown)


klihub:
> @klihub could you rebase on top of the main branch?

Rebased on latest main/HEAD.

TomSweeneyRedHat (commented):
LGTM
once the tests get hip

klihub:
> LGTM once the tests get hip

The test failures look like flakey ones to me. At least the integration test failure is one of the usual suspects (metrics container OOM).

PR #6892, opened 2023-05-08
Use native crierrors package for registry unavailable

saschagrunert:
#### What type of PR is this?

/kind cleanup

After vendoring the latest Kubernetes `master`, we're now able to use the CRI errors package without having a need to pin OTEL any more.

None
None

```release-note
None
```


server/image_pull.go
```diff
@@ -96,7 +96,7 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 
 	if pullOp.err != nil {
 		if errors.Is(pullOp.err, syscall.ECONNREFUSED) {
-			return nil, k8simages.ErrRegistryUnavailable
+			return nil, crierrors.ErrRegistryUnavailable
```
sohankunkerkar, line 99: not a blocker, but the PR would be easy to read if you push this change in a separate commit.

haircommander: +1 I agree

saschagrunert: Done, split-up the commits.

PR #6903, opened 2023-05-10
internal/storage: add sandbox/pause image to the list of pinned_images

sohankunkerkar:
Update CRI-O to include configured pause image in pinned images list. This eliminates the need for users to manually add the pause image to the list when a different image is configured in Kubernetes.

#### What type of PR is this?

/kind cleanup

```release-note
Add sandbox/pause image to the list of pinned_images if configured by the user.
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -855,6 +855,10 @@ func GetImageService(ctx context.Context, store storage.Store, serverConfig *con
 		IndexConfigs:          make(map[string]*indexInfo),
 		InsecureRegistryCIDRs: make([]*net.IPNet, 0),
 	}
+	// add the sandbox/pause image configured by the user (if any) to the list of pinned_images.
+	if serverConfig.PauseImage != "" {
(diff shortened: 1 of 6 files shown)


internal/storage/image.go
```diff
@@ -861,7 +861,7 @@ func GetImageService(ctx context.Context, store storage.Store, serverConfig *con
 		imageCache:           make(map[string]imageCacheItem),
 		ctx:                  ctx,
 		config:               serverConfig,
-		regexForPinnedImages: CompileRegexpsForPinnedImages(serverConfig.PinnedImages),
+		regexForPinnedImages: CompileRegexpsForPinnedImages(append(serverConfig.PinnedImages, config.DefaultPauseImage)), // add sandbox/pause image to the list of pinned_images
```
haircommander, line 864: can you elaborate on rational here? like say how we always add the sandbox image even if not specified to protect the user and sandbox image or something more eloquent

sohankunkerkar: Addressed this concern in the latest push. Let me know your thoughts.

saschagrunert (approved):

PR #6907, opened 2023-05-10
fix kubectl version in bug report template

jwhb:
#### What type of PR is this?

/kind documentation

The command `kubectl --version` in bug report template is invalid.

`kubectl version` is correct, but is deprecated: output will be equal to `kubectl version --short` in the future.

JSON or YAML output will continue to contain full version information.

This PR updates the bug report template to output JSON, which is stable long-term.

None

```release-note
None
```


mrunalp (approved):
/lgtm

PR #6917, opened 2023-05-11
OCPBUGS-13542: high perf hooks: disable CPU quota with libcontainer as a pre start hook

haircommander:
#### What type of PR is this?

/kind bug
    if cpu quota is disabled as part of container creation, then the pod
    cgroup does not have it disabled, which causes the cgroup to continue to
    have cpu quota. Instead, partially revert efaea10606ce6c7012 and use
    libcontainer to fix the issue it was originally trying to (an issue
    where newer systemd reverts the disabling of cpu quota because it was
    only applied on the cgroupfs)

```release-note
Fix a bug where the `cpu-quota.crio.io` annotation was not propagated to the pod cgroup, meaning cpu quota was not disabled for the container
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks.go
@@ -7,9 +7,11 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 
+	"github.com/cri-o/cri-o/internal/config/cgmgr"
 	"github.com/cri-o/cri-o/internal/config/node"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/log"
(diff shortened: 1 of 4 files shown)


haircommander:
@cri-o/cri-o-maintainers PTAL

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -252,6 +274,52 @@ func setIRQLoadBalancing(ctx context.Context, c *oci.Container, enable bool, irq
...
+		cgroupManager cgmgr.CgroupManager
+		err           error
+	)
+
+	if strings.HasSuffix(parentDir, ".slice") {
+		if cgroupManager, err = cgmgr.SetCgroupManager("systemd"); err != nil {
+			return nil
+		}
+	} else if cgroupManager, err = cgmgr.SetCgroupManager("cgroupfs"); err != nil {
+		return nil
```
rphillips, line 288: wdyt of adding a comment on why we are ignoring the error or logging the error?

applies to line 285 as well

haircommander: good point, I actually updated it to panic because it shouldn't be possible to fail unless there's a programming error

rphillips: lgtm

umohnani8:
LGTM

Issue #6805, opened 2023-04-12
1.25.3 cri-o - missing container_network* metrics in kubelet cAdvisor


Noksa:
### What happened?

Containter_network metrics are missing once we upgraded 1.25.2 to 1.25.3:
```
❯ ~ k get --raw /api/v1/nodes/NODE_NAME/proxy/metrics/cadvisor | grep "container_network"
# HELP container_network_receive_bytes_total Cumulative count of bytes received
# TYPE container_network_receive_bytes_total counter
container_network_receive_bytes_total{container="",id="/",image="",interface="cali4d130b05435",name="",namespace="",pod=""} 3.195685419e+09 1681300536414
container_network_receive_bytes_total{container="",id="/",image="",interface="caliecaa2707004",name="",namespace="",pod=""} 3.82519493e+08 1681300536414
container_network_receive_bytes_total{container="",id="/",image="",interface="ens3",name="",namespace="",pod=""} 4.698722634e+09 1681300536414
container_network_receive_bytes_total{container="",id="/",image="",interface="tunl0",name="",namespace="",pod=""} 2.596887196e+09 1681300536414
# HELP container_network_receive_errors_total Cumulative count of errors encountered while receiving
```

So - no metrics per container/pod.

If we use `1.25.2` cri-o, we see the metrics:
```
❯ ~ k get --raw "/api/v1/nodes/NODE/proxy/metrics/cadvisor" | grep "container_network_receive_bytes_total"
# HELP container_network_receive_bytes_total Cumulative count of bytes received
# TYPE container_network_receive_bytes_total counter
container_network_receive_bytes_total{container="",id="/",image="",interface="cali5488198f6fd",name="",namespace="",pod=""} 9.5952393e+07 1681300719581
container_network_receive_bytes_total{container="",id="/",image="",interface="cali5eda668cb9c",name="",namespace="",pod=""} 5.823608e+06 1681300719581
container_network_receive_bytes_total{container="",id="/",image="",interface="enp0s6",name="",namespace="",pod=""} 7.86041475e+08 1681300719581
container_network_receive_bytes_total{container="",id="/",image="",interface="tunl0",name="",namespace="",pod=""} 3.6428018e+07 1681300719581
[...]

PR #6930, opened 2023-05-17
[1.25] cgmgr: create cgroups for systemd cgroup driver for dropped infra pods

haircommander:
The history here is a bit convoluted. Originally, runc created the cgroup for the infra container.
    cAdvisor was built to assume the cgroup for the infra container would be created, and it uses
    this to find the network metrics for the pod. When we dropped the infra container, cri-o needed to make this
    cgroup so cAdvisor could still find the network metrics.
    
    However, systemd didn't like the way we did it, and would remove the cgroup mid pod creation, which was fixed in
    https://github.com/cri-o/cri-o/pull/6196. This actually caused the cgroup to not be created at all, which then caused
    the networking metrics to not be gathered at all.
    
    Thus, we do need to create a cgroup underneath the systemd cgroup. Attempt to use a slice for this, as systemd won't require a
    process be underneath it.

#### What type of PR is this?

/kind bug
reverts https://github.com/cri-o/cri-o/pull/6196 
fixes https://github.com/cri-o/cri-o/issues/6657 
partial cherry-pick of https://github.com/cri-o/cri-o/pull/6856, ignoring the last commit because it's not relevant for 1.25

```release-note
Fix a bug where network metrics collection is broken with systemd cgroup driver and dropped infra containers.
```

### DIFF
--- a/internal/config/cgmgr/cgmgr.go
+++ b/internal/config/cgmgr/cgmgr.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/cri-o/cri-o/internal/config/node"
 	libctr "github.com/opencontainers/runc/libcontainer/cgroups"
+	libctrCgMgr "github.com/opencontainers/runc/libcontainer/cgroups/manager"
+	cgcfgs "github.com/opencontainers/runc/libcontainer/configs"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


haircommander:
fixes https://github.com/cri-o/cri-o/issues/6805

umohnani8:
LGTM

haircommander:
/retest
@cri-o/cri-o-maintainers this is ready for merge :smiley:

saschagrunert (approved):

PR #6931, opened 2023-05-17
contrib/test: re-enable block volmod tests

sohankunkerkar:
#### What type of PR is this?

/kind ci

```release-note
None
```


haircommander:
eventually, we should try something like this for each of the disabled tests to see if we can run any more :)

mrunalp (approved):
/lgtm

PR #6944, opened 2023-05-22
internal/factory/container: take dedicated CRI ContainerConfig.CDIDevices field into use.

klihub:
#### What type of PR is this?

/kind other

Take the recently introduced CRI ContainerConfig.CDIDevices field into use as the primary source for CDI device resolution and injection..

None

Note that this commit does not remove CDI device injection using annotations. That will get deprecated and eventually removed in tandem with CDI itself.

```release-note
- add support for CDI device resolution and injection using the dedicated CRI ContainerConfig.CDIDevices field.
```

### DIFF
--- a/internal/factory/container/device.go
+++ b/internal/factory/container/device.go
@@ -176,12 +176,53 @@ func (c *container) specAddContainerConfigDevices(enableDeviceOwnershipFromSecur
 }
 
 func (c *container) specInjectCDIDevices() error {
-	// TODO: Once CRI is extended with native CDI support this will need to be updated...
-	_, names, err := cdi.ParseAnnotations(c.Config().GetAnnotations())
+	var (
+		cdiDevices = c.Config().CDIDevices
+		fromCRI    = map[string]struct{}{}
+		requested  = make([]string, 0, len(cdiDevices))
+		annotated  []string
+		err        error
+	)
+
+	// Take CDI devices from the dedicated CDIDevices CRI field.
+	for _, dev := range cdiDevices {
+		requested = append(requested, dev.Name)
+		fromCRI[dev.Name] = struct{}{}
+	}
+
+	// Extract CDI devices from annotations which is still supported as a means
+	// of injecting CDI devices to give people time to update their DRA drivers.
+	// TODO(klihub): Change the log message to a warning once annotations are
+	// deprecated, and to an error once support is removed altogether.
+	_, annotated, err = cdi.ParseAnnotations(c.Config().GetAnnotations())
 	if err != nil {
 		return fmt.Errorf("failed to parse CDI device annotations: %w", err)
 	}
-	if names == nil {
+
+	// Allow injecting the same device using both a dedicated field and an
+	// annotation. This helps the transition from annotations to dedicated
+	// CDI fields. DRA drivers can be updated to first inject devices using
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat (commented):
Any man page changes needed?

klihub:
> Any man page changes needed?

There shouldn't be any. This change is about how a Kubernetes DRA driver requests CDI device injection into containers. It should be a non-user-visible change.

Originally we lacked a dedicated CRI protocol field for this, so a CDI-specific annotation was used. Now, with a dedicated CRI protocol field in place, it became the primary/preferred way of communication. The original original annotation is still supported for backward compatibility. It will soon get marked as deprecated, at which point we'll start giving a warning about its usage. Eventually it will be removed altogether and we'll turn the warning into an error.

haircommander:
/approve

LGTM, test failures don't seem related

internal/factory/container/device.go
```diff
@@ -176,12 +176,53 @@ func (c *container) specAddContainerConfigDevices(enableDeviceOwnershipFromSecur
...
+	// drivers can be updated to stop using annotations. This also mirrors
+	// the behavior implemented in containerd.
+	if len(annotated) > 0 {
+		for _, name := range annotated {
+			if _, ok := fromCRI[name]; ok {
+				// TODO(klihub): change to a warning once annotations are deprecated
+				log.Infof(context.TODO(),
+					"Skipping duplicate annotated CDI device %s", name)
+				continue
+			}
```
bart0sh, line 216: Above is a valid warning. Can you add a test case covering this?

sohankunkerkar: @klihub would you like to handle this as a follow-up?

sohankunkerkar: Let's handle this in the follow up PR

sohankunkerkar (approved):
/lgtm

internal/factory/container/device.go
klihub: Yeah, if it is OK for folks to just get this in and then handle that and anything else that might come up with with a follow-up PR, then I'd be for that, too.

klihub: But let me rebase on latest main and check at the same time if I could add such a test with minimal effort.

internal/factory/container/device.go
```diff
@@ -176,12 +176,53 @@ func (c *container) specAddContainerConfigDevices(enableDeviceOwnershipFromSecur
...
+	for _, dev := range cdiDevices {
+		requested = append(requested, dev.Name)
+		fromCRI[dev.Name] = struct{}{}
+	}
+
+	// Extract CDI devices from annotations which is still supported as a means
+	// of injecting CDI devices to give people time to update their DRA drivers.
+	// TODO(klihub): Change the log message to a warning once annotations are
+	// deprecated, and to an error once support is removed altogether.
+	_, annotated, err = cdi.ParseAnnotations(c.Config().GetAnnotations())
```
haircommander, line 197: should we mark them as deprecated for 1.28 and then remove them in 1.29?

klihub: Yes, I think we should mark it deprecated. About removing support for it altogether, I think we should go in sync with what @bart0sh is doing on that front in containerd.

PR #1612, opened 2018-06-13
container_create: fixes

runcom:
**- What I did**

Various fixes, see commits

**- How I did it**

**- How to verify it**

**- Description for the changelog**

PR #6964, opened 2023-05-25
Revert "[1.23] Vendor c/storage PR #1612"

saschagrunert:
@sohankunkerkar this should not have been merged, I should have put a hold on it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -78,8 +78,6 @@ require (
 	sigs.k8s.io/zeitgeist v0.3.0
 )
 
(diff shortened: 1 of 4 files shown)


shitou12345678:
I have a doubt that if using defer compressor.Close(),  the data stored in r.tspath(layer.ID) in the code `if err := ioutils.AtomicWriteFile(r.tspath(layer.ID), tsdata.Bytes(), 0600)` may be incomplete. Because in the function compressor.Close(), it completes data writing.

PR #6971, opened 2023-05-25
ci/gha: fix double caching

kolyshkin:
Since commit b34038f8f4b1bc / PR #6735, gha ci uses actions/setup-go v4. This version adds implicit caching, very similar to what we did, and so we don't have to use actions/cache explicitly anymore.

The only catch is, it needs go.mod, so make sure that actions/checkout is run before actions/setup-go.

Remove own caching, except for build-i386.

/kind ci

```release-note
NONE
```

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -16,17 +16,10 @@ jobs:
   critest:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


sohankunkerkar (approved):
Nice catch!
/lgtm

PR #6977, opened 2023-05-26
Pause container during checkpointing

### LINKED ISSUE #6972: Error on restoring customized mysql application via Kubernetes C/R
### What happened?

We run a customized mysql application internally. Checkpoint and Restore works for open-source mysql. For our internal application, I ran into the issue for FD limit when doing checkpoint, but that issue is fixed following https://github.com/cri-o/cri-o/issues/6936#issuecomment-1553533775 recommendation. After checkpoint, I created an image using buildah (following this [tutorial](https://martinheinz.dev/blog/85)). Now I am trying to restore and getting an error but error message is not descriptive enough. Attached the log file for CRIU.

I expect the checkpoint and restore to work. It works with open-source mysql application but is having issue with internal customized mysql.

I am following this [tutorial](https://martinheinz.dev/blog/85) for C/R using K8s. 

Response from @adrianreber from [ticket](https://github.com/checkpoint-restore/criu/issues/2185) I created at CRIU repo.

```
The following seems to be the reason for your failure:

(00.214201)   1581: Error (criu/files-reg.c:2094): File var/log/ndb/lst_log/mysqld.n209-035-156.invalid-user.log.WARNING.001 has bad size 378073 (expect 374776)

CRIU stops restoring if any open file has a different file size. Restoring a process with open files that have changed are an indication that something is not right and might result in failures in the future. In you case it could mean that CRIU positions the file pointer at the wrong location during restore and existing data in the log file will be overwritten.

The CRI-O code is based on the Podman code and default behaviour in Podman is that the container does not continue running. In CRI-O and Kubernetes the only possibility is to checkpoint a container and keep it running. From the KEP introducing this into Kubernetes it is the only possibility.

The problem you are now hitting is that your application is checkpointed and then continues running and then we collect in CRI-O all the files that have changed. By the time we collect your files the application has changed the log file and the size we put in the checkpoint archive is now larger. So this is a CRI-O not handling checkpointing process correctly.

CRI-O needs to pause the container before checkpointing and let it continue to run just after all files have been collected.

Should be an easy change. Again, this is a CRI-O problem and not a CRIU problem.
```

<details>

```console
$ crio --version
root@n209-035-156:~# crio --version
crio version 1.27.0
Version:        1.27.0
GitCommit:      ac97112b9adc9f670506452a02d22d6e43be7d89
GitCommitDate:  2023-05-17T12:33:02Z
GitTreeState:   dirty
BuildDate:      2023-05-18T23:03:23Z
GoVersion:      go1.19
Compiler:       gc
Platform:       linux/amd64
Linkmode:       dynamic
BuildTags:      
  containers_image_ostree_stub
  containers_image_openpgp
  seccomp
  selinux
LDFlags:          unknown
SeccompEnabled:   true
AppArmorEnabled:  false
```

```console
root@n209-035-156:~# kubectl version --output=json
{
  "clientVersion": {
    "major": "1",
    "minor": "25+",
    "gitVersion": "v1.25.6-dirty",
    "gitCommit": "ff2c119726cc1f8926fb0585c74b25921e866a28",
    "gitTreeState": "dirty",
    "buildDate": "2023-05-24T23:29:19Z",
    "goVersion": "go1.19",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "kustomizeVersion": "v4.5.7"
}
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
root@n209-035-156:~# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

$ uname -a
root@n209-035-156:~# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
root@n209-035-156:~# uname -a
Linux n209-035-156 5.4.56.bsk.9-amd64 #5.4.56.bsk.9 SMP Debian 5.4.56.bsk.9 Wed Aug 25 04:08:18 UTC 20 x86_64 GNU/Linux

```

</details>

<details>
Physical Server
</details>
[criu.log](https://github.com/cri-o/cri-o/files/11568894/criu.log)

adrianreber:
#### What type of PR is this?

/kind bug

The initial implementation of checkpointing in CRI-O was based on Podman and initially the default workflow would have been that the container is stopped after checkpointing. CRIU will just kill all processes in the container.

As the initial Kubernetes checkpointing feature is based on the Forensic Container Checkpointing KEP, the container keeps on running after checkpointing.

If the container keeps on running after checkpointing it can happen that the files in the container are changed once we put the into the checkpoint archive. This means that the files are different during restore than they were during checkpointing.

CRIU aborts restoring if the file size of any open file has changed as restoring will put the FD pointer at the same location as during checkpointing. If the file, however, is different this can lead to data loss or crashes.

To solve this, this commit pauses the container before checkpointing and unpauses it after the checkpoint archive has been written to disk.

As checkpointing only works with the OCI runtimes currently, we do not have to handle the pause during restore.

The OCI runtime (runc/crun) uses the cgroup freezer to pause the process. CRIU also uses the cgroup freezer to pause all processes in the container. CRIU does not change the state of the cgroup freezer if the cgroup is already frozen. So the cgroup was already always frozen during checkpointing. With this change the frozen time is now controlled by CRI-O and not CRIU, but we still do not have to handle it during restore.

Fixes #6972 

```release-note
None
```

### DIFF
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -42,6 +42,35 @@ func (c *ContainerServer) ContainerCheckpoint(
 		return "", fmt.Errorf("container %s is not running", ctr.ID())
 	}
 
+	// At this point the container needs to be paused. As we first checkpoint
+	// the processes in the container and the container will continue to run
+	// after checkpointing, there is a chance that the changed files we include
+	// in the checkpoint archive might change by the now again running processes
+	// in the container.
+	// Assuming this uses runc/crun (the OCI runtime is currently the only one
+	// supporting checkpointing), PauseContainer() will use the cgroup freezer
+	// to freeze the processes. CRIU will also use the cgroup freezer to freeze
+	// the processes if possible. If the cgroup is already frozen by runc/crun
+	// CRIU will not change the freezer status.
+	if err = c.runtime.PauseContainer(ctx, ctr); err != nil {
+		return "", fmt.Errorf("failed to pause container %q before checkpointing: %w", ctr.ID(), err)
+	}
+	defer func() {
+		if err := c.runtime.UpdateContainerStatus(ctx, ctr); err != nil {
+			log.Errorf(ctx, "Failed to update container status: %q: %v", ctr.ID(), err)
+		}
+		if ctr.State().Status == oci.ContainerStatePaused {
+			err := c.runtime.UnpauseContainer(ctx, ctr)
+			if err != nil {
+				log.Errorf(ctx, "Failed to unpause container: %q: %v", ctr.ID(), err)
+			}
+		}
+		// container state needs to be written _after_ unpausing
+		if err = c.ContainerStateToDisk(ctx, ctr); err != nil {
+			log.Warnf(ctx, "Unable to write containers %s state to disk: %v", ctr.ID(), err)
+		}
+	}()
+
 	if opts.TargetFile != "" {
 		if err := c.prepareCheckpointExport(ctr); err != nil {
 			return "", fmt.Errorf("failed to write config dumps for container %s: %w", ctr.ID(), err)
@@ -56,11 +85,10 @@ func (c *ContainerServer) ContainerCheckpoint(
 			return "", fmt.Errorf("failed to write file system changes of container %s: %w", ctr.ID(), err)
 		}
 	}
-	if err := c.storageRuntimeServer.StopContainer(ctx, ctr.ID()); err != nil {
-		return "", fmt.Errorf("failed to unmount container %s: %w", ctr.ID(), err)
-	}
-	if err := c.ContainerStateToDisk(ctx, ctr); err != nil {
-		log.Warnf(ctx, "Unable to write containers %s state to disk: %v", ctr.ID(), err)
+	if !opts.KeepRunning {
+		if err := c.storageRuntimeServer.StopContainer(ctx, ctr.ID()); err != nil {
+			return "", fmt.Errorf("failed to unmount container %s: %w", ctr.ID(), err)
+		}
 	}
 
 	if !opts.Keep {
--- a/internal/lib/checkpoint_test.go
+++ b/internal/lib/checkpoint_test.go
@@ -110,7 +110,7 @@ var _ = t.Describe("ContainerCheckpoint", func() {
 
 			// Then
 			Expect(err).ToNot(BeNil())
-			Expect(err.Error()).To(ContainSubstring(`failed to checkpoint container containerID`))
+			Expect(err.Error()).To(ContainSubstring(`failed to pause container "containerID" before checkpointing`))
 		})
 	})
 	t.Describe("ContainerCheckpoint", func() {
--- a/server/container_checkpoint_test.go
+++ b/server/container_checkpoint_test.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/containers/podman/v4/pkg/criu"
 	"github.com/cri-o/cri-o/internal/oci"
-	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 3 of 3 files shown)


internal/lib/checkpoint.go
```diff
@@ -37,11 +37,44 @@ func (c *ContainerServer) ContainerCheckpoint(
 		return "", fmt.Errorf("not able to read config for container %q: %w", ctr.ID(), err)
 	}
 
+	if err := c.runtime.UpdateContainerStatus(ctx, ctr); err != nil {
+		return "", err
+	}
+
```
haircommander, line 43: is this protection against the container stopping before being paused?

adrianreber: It was not necessary. I removed it.

internal/lib/checkpoint.go
```diff
@@ -42,6 +42,35 @@ func (c *ContainerServer) ContainerCheckpoint(
...
+	// to freeze the processes. CRIU will also use the cgroup freezer to freeze
+	// the processes if possible. If the cgroup is already frozen by runc/crun
+	// CRIU will not change the freezer status.
+	if err = c.runtime.PauseContainer(ctx, ctr); err != nil {
+		return "", fmt.Errorf("failed to pause container %q before checkpointing: %w", ctr.ID(), err)
+	}
+	defer func() {
+		if err := c.runtime.UpdateContainerStatus(ctx, ctr); err != nil {
+			log.Errorf(ctx, "Failed to update container status: %q: %v", ctr.ID(), err)
+		}
```
haircommander, line 61: same here, is this one necessary?

internal/lib/checkpoint.go
```diff
@@ -42,6 +42,35 @@ func (c *ContainerServer) ContainerCheckpoint(
...
+		if ctr.State().Status == oci.ContainerStatePaused {
+			err := c.runtime.UnpauseContainer(ctx, ctr)
+			if err != nil {
+				log.Errorf(ctx, "Failed to unpause container: %q: %v", ctr.ID(), err)
+			}
+		}
+		// container state needs to be written _after_ unpausing
+		if err = c.ContainerStateToDisk(ctx, ctr); err != nil {
+			log.Warnf(ctx, "Unable to write containers %s state to disk: %v", ctr.ID(), err)
+		}
```
haircommander, line 71: if we pause and always unpause, I don't know if we need to persist the state here (esp because we're not doing so with the pause above)

haircommander:
idk if you've seen, but you've got a unit test failure as well

internal/lib/checkpoint.go
adrianreber: This is needed for the case the container does not keep running after checkpointing. This is currently not used in CRI-O because the KEP does not contain the possibility to stop the container after checkpointing.

This code is basically a left over from the initial implementation based on Podman where the container usually is stopped after checkpointing for the migration use case.

There is also additional metadata like `Checkpointed:` true and `CheckpointedAt:` which is probably written by this line. I think this could be left as it is.

In this PR I just moved it to the `defer` function to do the cleanup in the right order.

internal/lib/checkpoint.go
adrianreber: Yes, this one is needed. For some reason the CRI-O internal state of the checkpointed container is "running". Only after `UpdateContainerStatus()` it has the correct value of "paused". Not sure why the state is wrong. Especially as during actual checkpointing (`runc checkpoint`) the state is only changed to "stopped" if `KeepRunning` is false.

It is always false in CRI-O (in contrast to Podman) as that is the only mode supported by the KEP and the CRI.

adrianreber:
> idk if you've seen, but you've got a unit test failure as well

Thanks. Will fix.

adrianreber:
The decreased code test coverage of 0.01% seems to be all about error code paths (`err != nil`).

TomSweeneyRedHat:
LGTM
once the tests are hip

adrianreber:
> LGTM once the tests are hip

both failing tests are unrelated to this PR as far as I understand it

adrianreber:
Can we move this PR forward? The last round of reviews was done if I understood it correctly.

The test failures seem unrelated.

saschagrunert (approved):

Issue #6979, opened 2023-05-26
bad logic in "ctr with node level pid namespace should not leak children" 


kolyshkin:
### What happened?

_filing this so I won't forget_

In this test case, we get ALL conmon pids, and thus it is not working as it should in case of parallel tests, or if someone is running other containers, etc.

Here's the problematic line:

https://github.com/cri-o/cri-o/blob/395d32218aa2173c6f9c0cfa940741c90c29d4c0/test/ctr.bats#L1005

(instead of getting OUR conmon pid, we get all of them).

The second problem is, conmon itself is not included into the check (should it? I think so).

Now, there are a few approaches to fixing this:

1. Get the conmon pid from the pidfile, e.g.

```bash
       conmon_pid="$(find $TESTDIR -path '*/userdata/conmon-pidfile' -type f -exec cat '{}' \;)"
       kill -0 "$conmon_pid" # Sanity check -- make sure conmon_pid is valid.
       processes=$(list_all_children "$conmon_pid" $conmon_pid)
```

Here we also add the conmon's own pid.

2. In case the conmon pid is not needed (it was not used in the original test except for getting childrent), we can get container init pid and go from there.

3. We can get the list of pids from the cgroup.

kolyshkin:
@haircommander do you think we should add conmon's own pid to the list of pids being checked, or not? The approach to use depends on the answer.

haircommander:
(we spoke offline, but answering for posterity) conmon's own pid should not be included, the test is looking for all of the pids inside the container

PR #6950, opened 2023-05-23
Fix/improve bats tests for modern bats

kolyshkin:
1. Using `! cmd` does not result in what we want except when in the last line of the test case (see https://bats-core.readthedocs.io/en/stable/gotchas.html#my-negated-statement-e-g-true-does-not-fail-the-test-even-when-it-should). Fix such checks using `run ! cmd`.

2. With bats 1.5.0+, we can make use of `run` command to
 - check exit codes;
 - print what was executed;
 - print command output.
 
 With that, a bunch of `$status` checks and `echo $output` lines can be removed.

PR #6973, opened 2023-05-25
test: simplify and fix metrics oom test

kolyshkin:
This test fails sometimes, because apparently no OOM is happening.

This is an attempt to fix it.

/kind ci

```release-note
NONE
```

PR #6995, opened 2023-05-31
test/ctr: fix "should not leak children" test case

kolyshkin:
_Based on (and currently includes) #6950; draft until that one is merged._

    
The test was getting ALL conmon PIDS, which doesn't really work in case
we run tests in parallel (which we do), or in case conmon-rs is used.
    
In fact, what we need is PIDs of all processes in the container. The
best place to get these is cgroups.
    
Do that, and simplify process status check using grep.
    
Fixes: 9fea6e219e
Fixes: #6979 

/kind ci

```release-note
NONE
```

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -11,18 +11,6 @@ function teardown() {
 	cleanup_test
 }
 
-# list_all_children lists children of a process recursively
-function list_all_children {
-	children=$(pgrep -P "$1")
-	for i in ${children}; do
-		if [ -z "$i" ]; then
-			exit
-		fi
-		echo -n "$i "
-		list_all_children "$i"
-	done
-}
-
 function check_oci_annotation() {
 	# check for OCI annotation in container's config.json
 	local ctr_id="$1"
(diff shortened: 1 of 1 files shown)


test/ctr.bats
```diff
@@ -1006,11 +997,18 @@ function check_oci_annotation() {
...
 
 	# make sure crio syncs state
-	for process in ${processes}; do
-		# Ignore Z state (zombies) as the process has just been killed and reparented. Systemd will get to it.
-		# `pgrep` doesn't have a good mechanism for ignoring Z state, but including all others, so:
-		# shellcheck disable=SC2143
-		[ -z "$(ps -p "$process" o pid=,stat= | grep -v ' Z')" ]
+	for pid in ${processes}; do
+		if kill -0 "$pid"; then
+			# The proces is still there. If it's a zombie,
```
haircommander, line 1002: process*

kolyshkin: fixed

kolyshkin:
Test failures ("not ok 182 metrics container oom" and "not ok 291 emit metric when container is re-requested") are not related and are being addressed elsewhere (#6973 and #7019).

PR #7000, opened 2023-05-31
OCPBUGS-13980,OCPBUGS-14018: Fix bugs with high performance hooks

haircommander:
#### What type of PR is this?

/kind bug
This PR fixes two separate bugs with high performance hooks, and does some code reorganization while it's at it.
It's kept in one PR to simplify backporting.
- Fix a bug where the wrong cgroup path was chosen by libcontainer when disabling cpu quota
- Reoganize hooks and NRI plugins so they're called in all cases containers are stopped
- Fix a bug where stale containers (ones that have stopped but not been cleaned up) break cpu load balancing by disabling load balancing after they're stopped.

```release-note
Fix a bug with cpu quota annotation that manifests like:
`pod with cpu-quota.crio.io: disable fails with error: set CPU CFS quota: invalid slice name: /kubepods.slice`
Fix a bug where stopped containers break cpu load balancing being disabled
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks.go
@@ -148,6 +148,15 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s 
 	return nil
 }
 
+// If CPU load balancing is enabled, then *all* containers must run this PostStop hook.
+func (*HighPerformanceHooks) PostStop(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
+	// We could check if `!cpuLoadBalancingAllowed()` here, but it requires access to the config, which would be
+	// odd to plumb. Instead, always assume if they're using a HighPerformanceHook, they have CPULoadBalanceDisabled
+	// annotation allowed.
+	h := &DefaultCPULoadBalanceHooks{}
+	return h.PostStop(ctx, c, s)
+}
+
 func shouldCPULoadBalancingBeDisabled(annotations fields.Set) bool {
 	if annotations[crioannotations.CPULoadBalancingAnnotation] == annotationTrue {
 		log.Warnf(context.TODO(), annotationValueDeprecationWarning(crioannotations.CPULoadBalancingAnnotation))
@@ -275,6 +284,20 @@ func setIRQLoadBalancing(ctx context.Context, c *oci.Container, enable bool, irq
 }
 
 func setCPUQuota(parentDir string, c *oci.Container) error {
+	containerCgroup, containerCgroupParent, systemd, err := containerCgroupAndParent(parentDir, c)
+	if err != nil {
+		return err
+	}
+	podCgroup := filepath.Base(containerCgroupParent)
+	podCgroupParent := filepath.Dir(containerCgroupParent)
+
+	if err := disableCPUQuotaForCgroup(podCgroup, podCgroupParent, systemd); err != nil {
+		return err
+	}
+	return disableCPUQuotaForCgroup(containerCgroup, containerCgroupParent, systemd)
+}
+
+func containerCgroupAndParent(parentDir string, c *oci.Container) (ctrCgroup, parentCgroup string, systemd bool, _ error) {
 	var (
 		cgroupManager cgmgr.CgroupManager
 		err           error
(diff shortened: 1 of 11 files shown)


server/container_stop.go
```diff
@@ -22,34 +23,31 @@ func (s *Server) StopContainer(ctx context.Context, req *types.StopContainerRequ
...
-		log.Warnf(ctx, "NRI stop failed for container %q: %v", c.ID(), err)
-	}
-
 	log.Infof(ctx, "Stopped container %s: %s", c.ID(), c.Description())
 	return &types.StopContainerResponse{}, nil
 }
 
 // stopContainer stops a running container with a grace period (i.e., timeout).
-func (s *Server) stopContainer(ctx context.Context, ctr *oci.Container, timeout int64) error {
+func (s *Server) stopContainer(ctx context.Context, ctr *oci.Container, sb *sandbox.Sandbox, timeout int64) error {
```
kolyshkin, line 36: Why don't you add
```go
sb := s.getSandbox(ctx, ctr.Sandbox())
```
instead of adding a new parameter?

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -275,6 +284,20 @@ func setIRQLoadBalancing(ctx context.Context, c *oci.Container, enable bool, irq
...
+	podCgroup := filepath.Base(containerCgroupParent)
+	podCgroupParent := filepath.Dir(containerCgroupParent)
+
+	if err := disableCPUQuotaForCgroup(podCgroup, podCgroupParent, systemd); err != nil {
+		return err
+	}
+	return disableCPUQuotaForCgroup(containerCgroup, containerCgroupParent, systemd)
+}
+
+func containerCgroupAndParent(parentDir string, c *oci.Container) (ctrCgroup, parentCgroup string, systemd bool, _ error) {
```
MarSik, line 300: There is logic here, how will we make sure it stays in sync with the container creation code?

internal/runtimehandlerhooks/runtime_handler_hooks.go
```diff
@@ -44,3 +54,23 @@ func highPerformanceAnnotationsSpecified(annotations map[string]string) bool {
 	}
 	return false
 }
+
+func cpuLoadBalancingAllowed(config *libconfig.Config) bool {
+	cpuLoadBalancingEnabledOnce.Do(func() {
+		for _, runtime := range config.Runtimes {
+			for _, ann := range runtime.AllowedAnnotations {
+				if ann == crioann.CPULoadBalancingAnnotation {
+					cpuLoadBalancingEnabled = true
```
MarSik, line 63: The name is a bit confusing, should it be something like

```suggestion
					cpuLoadBalancingControlEnabled = true
```

internal/runtimehandlerhooks/high_performance_hooks.go
haircommander: at the very least, the new test case should filter any changes in the cgroup hierarchy that causes this to not work. I'm not sure we can do much more than expect the tests to save us. FWIW I don't expect the hierarchy to change so much it's pretty stable for v1

internal/runtimehandlerhooks/runtime_handler_hooks.go
haircommander: I went for `cpuLoadBalancingAllowedAnywhere` does that clarify?

server/container_stop.go
haircommander: done thanks!

haircommander:
this is ready for review @cri-o/cri-o-maintainers

saschagrunert (approved):

internal/runtimehandlerhooks/runtime_handler_hooks.go
```diff
@@ -44,3 +54,23 @@ func highPerformanceAnnotationsSpecified(annotations map[string]string) bool {
 	}
 	return false
 }
+
+func cpuLoadBalancingAllowed(config *libconfig.Config) bool {
+	cpuLoadBalancingAllowedAnywhereOnce.Do(func() {
+		for _, runtime := range config.Runtimes {
+			for _, ann := range runtime.AllowedAnnotations {
+				if ann == crioann.CPULoadBalancingAnnotation {
+					cpuLoadBalancingAllowedAnywhere = true
```
MarSik, line 63: The reason I was questioning the name is that the annotation is only used to disable cpu balancing.

Issue #7002, opened 2023-05-31
Deprecate and remove device mapper storage driver


kolyshkin:
It seems that no one is using devicemapper storage driver nowadays, since overlayfs is so much better.

Can it be deprecated and removed? I see that moby/moby did that recently.

See also: https://github.com/containers/storage/issues/1619

haircommander:
+1 from me, but we should have a slower deprecation process. 1.28, we can warn the user if they're using devmapper and mark as deprecated, and 1.29 we can drop support

saschagrunert:
+1 from my side, putting it into the project board

PR #7003, opened 2023-05-31
Remove device mapper storage driver support

kolyshkin:
#### What type of PR is this?

/kind deprecation

Device mapper storage driver is hard to set up, slow, and obsoleted by overlayfs.
My best guess is nobody is using it nowadays.

Let's remove it.

Fixes: #7002

```release-note
Remove device mapper storage driver.
```


kolyshkin:
OK, this is blocked by buildah which can't be built without libdevmapper.

Moving to draft until we have buildah release with https://github.com/containers/buildah/pull/4832 included

fidencio:
@kolyshkin, devmapper can be used with Kata Containers and in the Confidential Containers front it'll be heavily used.

We're using devmapper as TEE capable kernels do not support virtio-fs, virtio-9p is not supported by several distros (and also by several hypervisors), leaving us with the only option to use devmapper as the way to share the container rootfs between host and guest via a block device.

I know, Red Hat OpenShift is not using it, AFAIU, but other users of CRi-O could.

This is not a show stopper for this PR, but I'd love to have it taken into consideration.

/cc @littlejawa @bpradipt

sohankunkerkar:
>Moving to draft until we have buildah release with https://github.com/containers/buildah/pull/4832 included

This fix has been included in https://github.com/containers/buildah/releases/tag/v1.31.0

kolyshkin:
Resurrected as #8019

PR #7005, opened 2023-06-01
Update `README.md` version table

saschagrunert:
#### What type of PR is this?

/kind documentation

We always forget to add the new release, which is now not required any more considering a new format. We also outline that the patch release of CRI-O (`.y`) does not have to match with them from Kubernetes (`.z`).

Since we only have two data rows within the table, we also can avoid the key mapping and add their description directly to it.

None
None

```release-note
None
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,18 +29,10 @@ applies to features which are independent from Kubernetes.
 For more information visit the [Kubernetes Version Skew
 Policy](https://kubernetes.io/releases/version-skew-policy/).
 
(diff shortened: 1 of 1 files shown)


haircommander:
LGTM, @cri-o/cri-o-maintainers WDYT?

umohnani8:
LGTM

sohankunkerkar (commented):
Yeah, it makes sense to me. Less maintenance!
/lgtm

PR #7006, opened 2023-06-01
Drop support for path based seccomp profiles

saschagrunert:
#### What type of PR is this?

/kind cleanup

Field based profiles should be standard since many Kubernetes releases, means we can drop the path based logic and cleanup some code.

None

None

```release-note
Dropped support for annotation based seccomp profiles, Kubernetes uses a native field since many releases.
```

### DIFF
--- a/contrib/test/integration/vars.yml
+++ b/contrib/test/integration/vars.yml
@@ -12,7 +12,7 @@ build_kata: False
 cgroupv2: False
 
 critest_mirror_repo: quay.io/crio
-cri_tools_git_version: "v1.27.0"
+cri_tools_git_version: master
 golang_version: "1.20"
 
 # For results.yml Paths use rsync 'source' conventions
--- a/contrib/test/ci/vars.yml
+++ b/contrib/test/ci/vars.yml
@@ -14,7 +14,7 @@ use_conmonrs: '{{ USE_CONMONRS | default(False) | bool }}'
 evented_pleg_fg: '{{ EVENTED_PLEG | default(False) | bool }}'
 
 critest_mirror_repo: quay.io/crio
-cri_tools_git_version: "v1.27.0"
+cri_tools_git_version: master
 golang_version: "1.20"
 
 # For results.yml Paths use rsync 'source' conventions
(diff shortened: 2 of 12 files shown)


saschagrunert:
For critest this requires https://github.com/kubernetes-sigs/cri-tools/pull/1181

saschagrunert:
@sohankunkerkar looks like it's still using critest v1.27.0, do you have any idea how we can switch to `master`?

sohankunkerkar:
> @sohankunkerkar looks like it's still using critest v1.27.0, do you have any idea how we can switch to `master`?

We should merge this change as-is because the CI cache images still rely on v1.27.0. Alternatively, we can submit a different pull request by modifying the cri_tools_git_version.

contrib/test/integration/vars.yml
```diff
@@ -12,7 +12,7 @@ build_kata: False
 cgroupv2: False
 
 critest_mirror_repo: quay.io/crio
-cri_tools_git_version: "v1.27.0"
+cri_tools_git_version: master
```
TomSweeneyRedHat, line 15: any way this can be converted to "main"?

saschagrunert:
> would you mind also tying in dropping traces of `seccomp_default_when_empty` or does that feel out of scope?

I was thinking about it and re-added the behavior for the path based profiles in:
https://github.com/cri-o/cri-o/pull/7006/files#diff-824bf5b0cfe0c703e0eaac741045e010bc8b20297184bef4ab47d0e7d44c3b48R190

I can follow-up on that PR and remove the feature since it's deprecated for a while now.

contrib/test/integration/vars.yml
saschagrunert: Yes, when we agree on https://github.com/kubernetes-sigs/cri-tools/issues/1100

saschagrunert:
@cri-o/ocicni-maintainers PTAL

haircommander:
critest will fail until a new image is built

/override ci/prow/ci-fedora-critest
/override ci/prow/ci-rhel-critest
/lgtm

PR #7013, opened 2023-06-02
OCPBUGS-13747,OCPBUGS-14021: [1.26] Fix bugs with high performance hooks

haircommander:
This is a manual cherry-pick of #7000 and #7012
/kind bug

```release-note
Fix a bug with cpu quota annotation that manifests like:
`pod with cpu-quota.crio.io: disable fails with error: set CPU CFS quota: invalid slice name: /kubepods.slice`
Fix a bug where stopped containers break cpu load balancing being disabled
```


server/container_stop.go
```diff
@@ -82,5 +80,15 @@ func (s *Server) stopContainer(ctx context.Context, ctr *oci.Container, timeout
 		log.Warnf(ctx, "Unable to write containers %s state to disk: %v", ctr.ID(), err)
 	}
 
+	if hooks != nil {
+		if err := hooks.PostStop(ctx, ctr, sb); err != nil {
+			return fmt.Errorf("failed to run post-stop hook for container %q: %w", ctr.ID(), err)
```
MarSik, line 85: This must be just a log. The container is dead and there is nothing that can be done. Otherwise it will confuse the platform. See https://github.com/cri-o/cri-o/pull/7032

MarSik: In fact.. I wonder if it should have been below the stopContainer...

haircommander: I cherry-picked https://github.com/cri-o/cri-o/pull/7013/commits/93c4c561298e5e6fbc1a8dcaf9cfc3be7aab5e89

sdodson:
Spent a few minutes digging into the bug deps there's a weird situation where one of the bugs this depends on had previously gone to QA but got rejected because additional fixes were necessary and that bug is currently POST waiting on this PR to merge. Rather than try to disentangle all of that I'm just going to override the labels. We will likely need to manually move https://issues.redhat.com//browse/OCPBUGS-13747 to MODIFIED state once this merges.

/label jira/valid-bug
/unlabel jira/invalid-bug

MarSik:
@haircommander @sdodson Folks, we need to get this merged.

sdodson:
I forgot cri-o doesn't require valid bug/jira labels so not blocked on that.

haircommander:
okay, I picked a version of https://github.com/cri-o/cri-o/pull/7120 for this branch. this should now have all of the required fixes and the 1.27 variant has been verified

PR #7016, opened 2023-06-05
Add support for namespaced signature policies

saschagrunert:
#### What type of PR is this?

/kind feature

Adding support for the namespace separated signature policies on image pull.

None
None

```release-note
Added support for namespaced signature policies via `--signature-policy-namespace-root` / `signature_policy_namespace_root`.
```


saschagrunert:
@cri-o/cri-o-maintainers @QiWang19 PTAL

completions/bash/crio
```diff
@@ -110,6 +110,7 @@ h
 --selinux
 --separate-pull-cgroup
 --signature-policy
+--signature-policy-namespace-root
```
haircommander, line 113: I am not a huge fan of the wording here but I don't have anything else to suggest. If I or no one else can think of anything better then I am fine with this, but I find the name clunky as-is

umohnani8: What about `--signature-policy-dir` and the description can mention that it is where namespace separated signature policies are stored in the format `namespace.json`. So the final policy path used will be `signaturePolicyDir/Namespace.json`

haircommander: yeah I think that's a clean approach

umohnani8:
LGTM

Issue #6981, opened 2023-05-29
/var/lib/containers/storage/overlay unused containers root FS not getting cleaned


shashipratap:
### What happened?

/var/lib/containers/storage/overlay on k8s nodes contains the root FS of all containers running on that node , there are a lot of root FS which do not belong to any running container which is causing space to be used on nodes and eventually pods getting evicted due to low ephermal storage

### What did you expect to happen?

unused root FS sha directory which are not associated with any container should be removed from the /var/lib/containers/storage/overlay directory

### How can we reproduce it (as minimally and precisely as possible)?

create a cluster in OCI console and use oracle linux 7.9 nodes in it 

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
crio version 1.23.3-144.el7
Version:          1.23.3-144.el7
GitCommit:        unknown
GitTreeState:     unknown
BuildDate:        2023-03-15T20:49:19Z
GoVersion:        go1.19.6 A1727 X:boringcrypto
Compiler:         gc
Platform:         linux/amd64
Linkmode:         dynamic
BuildTags:        exclude_graphdriver_devicemapper, containers_image_ostree_stub, seccomp, selinux
SeccompEnabled:   true
AppArmorEnabled:  false
```

```console
$ kubectl version --output=json
{
  "clientVersion": {
    "major": "1",
    "minor": "23",
    "gitVersion": "v1.23.5",
    "gitCommit": "c285e781331a3785a7f436042c65c5641ce8a9e9",
    "gitTreeState": "clean",
    "buildDate": "2022-03-16T15:51:05Z",
    "goVersion": "go1.17.8",
    "compiler": "gc",
    "platform": "darwin/arm64"
  },
  "serverVersion": {
    "major": "1",
    "minor": "23",
    "gitVersion": "v1.23.4",
    "gitCommit": "0b25cacbeba0b73f720022544dd02ffe9e67017e",
    "gitTreeState": "clean",
    "buildDate": "2022-03-16T11:33:38Z",
    "goVersion": "go1.17.7 BoringCrypto",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
NAME="Oracle Linux Server"
VERSION="7.9"
ID="ol"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.9"
[...]

PR #7020, opened 2023-06-06
main: Added a call to GarbageCollect

whygyc:

Fixes: #6981

```release-note
None
```

### DIFF
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -342,6 +342,13 @@ func main() {
 			}
 		}
 
+		// We always use 'Volatile: true' when creating containers, which means that in
+		// the event of an unclean shutdown, we might lose track of containers and layers.
+		// We need to call the garbage collection function to clean up the redundant files.
+		if err := crioServer.Store().GarbageCollect(); err != nil {
+			logrus.Errorf("Attempts to clean up unreferenced old container leftovers failed: %v", err)
+		}
+
 		v1.RegisterRuntimeServiceServer(grpcServer, crioServer)
 		v1.RegisterImageServiceServer(grpcServer, crioServer)
 


cmd/crio/main.go
```diff
@@ -342,6 +342,13 @@ func main() {
 			}
 		}
 
+		// We always use "Volatile: true" to create containers, which means that we might
```
haircommander, line 345: I don't believe `Volatile: true` is a default value. can this be reworded to 
```
In situations where "Volatile: true" is specified in the container storage, there are containers and layers
that may be lost track of in an unclean shutdown. 
We need to call the garbage collection function to clean up the redundant files.
```

kolyshkin:
@whygyc a nit -- your added code has trailing whitespaces, which trigger warnings from two different linters; please fix this.

whygyc:
> @whygyc a nit -- your added code has trailing whitespaces, which trigger warnings from two different linters; please fix this.

Thanks for your reminding, I have modified it

cmd/crio/main.go
whygyc: I have modified it according to your requirement. But what I can see from the code is that `Volatile: true` is a fixed value and always true.
https://github.com/cri-o/cri-o/blob/e46a5207f2a05ad224b98dfc49f77fb996bb30ce/internal/storage/runtime.go#L260-L267

haircommander: ah I had the wrong option in mind (I was thinking they were for a transient store, not volatile containers)
you are correct, we are always setting volatile, I think you can revert to what it was before. sorry!

whygyc: > ah I had the wrong option in mind (I was thinking they were for a transient store, not volatile containers) you are correct, we are always setting volatile, I think you can revert to what it was before. sorry!

That's not a problem. I'm not very confident in my English. 
I trust your judgment and I will revert it if you think it was better before. It's easy and doesn't take much time.

sohankunkerkar (approved):
/lgtm

saschagrunert (approved):

PR #7024, opened 2023-06-07
Use staging `k8s.io/kubelet/cri/streaming` package

saschagrunert:
#### What type of PR is this?

/kind cleanup

After the merge of https://github.com/kubernetes/kubernetes/pull/118253, we can now use the upstream alternative rather than our own implementation to avoid code duplication.

None

None

```release-note
None
```


sohankunkerkar:
Is it possible to keep the generated code in a separate commit? It might help reviewers.

saschagrunert:
> Is it possible to keep the generated code in a separate commit? It might help reviewers.

Hm, everything in `vendor` is usually hidden from the GitHub diff output. I'm not a big fan of having commits that don't build in the history, but I get your point. There are some code changes required for this (switching the resize channel type), though.

sohankunkerkar:
> > Is it possible to keep the generated code in a separate commit? It might help reviewers.
> 
> Hm, everything in `vendor` is usually hidden from the GitHub diff output. I'm not a big fan of having commits that don't build in the history, but I get your point. There are some code changes required for this (switching the resize channel type), though.

ok

PR #7029, opened 2023-06-08
Deprecate device mapper

kolyshkin:
#### What type of PR is this?

/kind deprecation

```release-note
Devicemapper storage driver is now deprecated and will be removed in cri-o v1.28.
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -964,6 +964,10 @@ func (c *RootConfig) Validate(onExecution bool) error {
 		if err := os.MkdirAll(c.LogDir, 0o700); err != nil {
 			return fmt.Errorf("invalid log_dir: %w", err)
 		}
+		if c.Storage == "devicemapper" {
+			// TODO: return error.
+			logrus.Warn("devicemapper storage driver is deprecated and will be removed in cri-o v1.28")
+		}
 		store, err := c.GetStore()
 		if err != nil {
 			return fmt.Errorf("failed to get store to set defaults: %w", err)
--- a/test/command.bats
+++ b/test/command.bats
@@ -54,3 +54,8 @@ load helpers
 	run ! "$CRIO_BINARY_PATH" --log-size-max 18446744073709551616
 	[[ "$output" == *"value out of range"* ]]
 }
+
+@test "devicemapper is deprecated" {
+	run "$CRIO_BINARY_PATH" -s devicemapper
+	[[ "$output" == *"devicemapper storage driver is deprecated"* ]]
+}


test/command.bats
```diff
@@ -54,3 +54,8 @@ load helpers
 	run ! "$CRIO_BINARY_PATH" --log-size-max 18446744073709551616
 	[[ "$output" == *"value out of range"* ]]
 }
+
+@test "devicemapper is deprecated" {
+	"$CRIO_BINARY_PATH" -s devicemapper
```
littlejawa, line 59: You probably need to use "run", like the test above this one :
```suggestion
	run ! "$CRIO_BINARY_PATH" -s devicemapper
```

This will run the command and expect an error, and it will set the "ouput" variable that you're testing below.

With the current line, the test fails because this call causes an error:
```
06:14:38 # time="2023-06-08 04:14:38.073069018Z" level=warning msg="devicemapper storage driver is deprecated and will be removed in cri-o v1.28"
06:14:38 # time="2023-06-08 04:14:38.073268619Z" level=debug msg="[graphdriver] trying provided driver \"devicemapper\"" file="drivers/driver.go:335"
06:14:38 # time="2023-06-08 04:14:38.073347519Z" level=error msg="Failed to GetDriver graph devicemapper /var/lib/containers/storage" file="drivers/driver.go:308"
06:14:38 # time="2023-06-08 04:14:38.073400820Z" level=fatal msg="validating root config: failed to get store to set defaults: failed to GetDriver graph devicemapper /var/lib/containers/storage: driver not supported" file="crio/main.go:448"
```

pkg/config/config.go
```diff
@@ -964,6 +964,11 @@ func (c *RootConfig) Validate(onExecution bool) error {
 		if err := os.MkdirAll(c.LogDir, 0o700); err != nil {
 			return fmt.Errorf("invalid log_dir: %w", err)
 		}
+		if c.Storage == "devicemapper" {
```
haircommander, line 967: I am guessing this is not defined anywhere in c/storage, right?

test/command.bats
haircommander: I don't think we're compiling with devmapper support, hence this failure. we actually do expect to be able to specify it now but ci isn't set up to do so

pkg/config/config.go
```diff
@@ -964,6 +964,11 @@ func (c *RootConfig) Validate(onExecution bool) error {
...
+			// TODO: return error.
+			err := errors.New("devicemapper storage driver is deprecated and will be removed in cri-o v1.28")
+			logrus.Warn(err)
```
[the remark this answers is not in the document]
kolyshkin: This is done this way so we can change this to `return err` later. But yes, we sure can.

pkg/config/config.go
kolyshkin: No it's not (unlike for other drivers). More to say, sometimes documentation says the name is `devmapper` which is obviously wrong, but I'm not going to fix this since we're removing devicemapper anyway.

PR #7032, opened 2023-06-08
OCPBUGS-14750: Pod termination must succeed when a hook fails

MarSik:
#### What type of PR is this?

/kind bug

The pod is already dead when the PostStop hook is called and so there is nothing to block anyway. Just log the error.

Fixes: https://issues.redhat.com/browse/OCPBUGS-14750

```release-note
Failure in a PostStop hook does not prevent proper termination of a container anymore.
```

### DIFF
--- a/server/container_stop.go
+++ b/server/container_stop.go
@@ -81,7 +81,8 @@ func (s *Server) stopContainer(ctx context.Context, ctr *oci.Container, timeout 
 
 	if hooks != nil {
 		if err := hooks.PostStop(ctx, ctr, sb); err != nil {
-			return fmt.Errorf("failed to run post-stop hook for container %q: %w", ctr.ID(), err)
+			log.Errorf(ctx, "Failed to run post-stop hook for container %s: %v", ctr.ID(), err)
+			// The hook failure MUST NOT prevent the Pod termination
 		}
 	}
 


saschagrunert (changes requested):
Thank you for the fix @MarSik, may I ask you to sign-off your commit?

server/container_stop.go
```diff
@@ -81,7 +81,8 @@ func (s *Server) stopContainer(ctx context.Context, ctr *oci.Container, timeout
 
 	if hooks != nil {
 		if err := hooks.PostStop(ctx, ctr, sb); err != nil {
-			return fmt.Errorf("failed to run post-stop hook for container %q: %w", ctr.ID(), err)
+			log.Errorf("Failed to run post-stop hook for container %s: %v", ctr.ID(), err)
```
saschagrunert, line 84:
```suggestion
			log.Errorf(ctx, "Failed to run post-stop hook for container %s: %v", ctr.ID(), err)
```

MarSik: Ups, I was wondering why the IDE complained :) Thanks.

saschagrunert (approved):
/ok-to-test

server/container_stop.go
```diff
@@ -81,7 +81,8 @@ func (s *Server) stopContainer(ctx context.Context, ctr *oci.Container, timeout
 
 	if hooks != nil {
 		if err := hooks.PostStop(ctx, ctr, sb); err != nil {
```
MarSik, line 83: @haircommander Hey, the error fix is needed for sure, but I was actually wondering if the hook should move below the stopContainer call.

haircommander: that's nri.stopContainer, the actual container stopping has been done already

MarSik: Ack.

MarSik:
@haircommander You do not need a backport to 1.26, it does not have the bug yet. Your PR there has it fixed now.

PR #7046, opened 2023-06-13
Verify image signature policy on container creation

saschagrunert:
#### What type of PR is this?

/kind feature

As per https://github.com/openshift/enhancements/pull/1402#discussion_r1223681568, we need to verify the container image signature policy also on container creation.

None

None

```release-note
None
```


internal/storage/image.go
```diff
@@ -398,6 +401,43 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+
+	unparsedImage := cimage.UnparsedInstance(imageSource, &storeImage.Digest)
+
+	isAllowed, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage)
+	if err != nil {
+		return false, fmt.Errorf("policy is allowed: %w", err)
+	}
+
+	return isAllowed, nil
+}
```
saschagrunert, line 439: @mtrmac PTAL on this part.

internal/storage/image.go
```diff
@@ -398,6 +401,43 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	policy, err := signature.DefaultPolicy(systemContext)
+	if err != nil {
+		return false, fmt.Errorf("get default policy: %w", err)
+	}
+
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return false, fmt.Errorf("create policy context: %w", err)
+	}
+
```
mtrmac, line 419: (I imagine at least the file reading+parsing in `DefaultPolicy`, and _possibly_ `NewPolicyContext`, could be pre-created or preserved across innovations. (`PolicyContext` only allows a single thread at at a time, though.))

internal/storage/image.go
```diff
@@ -398,6 +401,43 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	defer imageSource.Close()
+
+	storeImage, err := istorage.Transport.GetStoreImage(svc.store, ref)
+	if err != nil {
+		return false, fmt.Errorf("get store image: %w", err)
+	}
+
+	unparsedImage := cimage.UnparsedInstance(imageSource, &storeImage.Digest)
+
+	isAllowed, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage)
```
mtrmac, line 433: This is using a policy for `containers-storage:`, not for the relevant registry. And that policy typically allows everything (or, if not configured so, falls back to the generic global default.)

---

In general, verifying the signature on an already-locally-pulled-image, is… interesting. We do store copies of the signatures locally, so it should be _possible_, but
- right now that would require a ~hack, creating a custom `ImageReference` implementation that points to a registry but fetches data locally. Ugly but viable
- we do not _always_ manage to preserve the originally-signed manifest: https://github.com/containers/image/issues/1768 , and encryption is a supported CRI-O feature. So those images would just break.

OTOH it _could_ be quite a bit simpler, and thus, at least in a sense, less risky, than contacting the registry and then trying to match the two images (and dealing with tags that move concurrently with our pull+start and the like).

server/image_pull.go
```diff
@@ -130,7 +147,12 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
-	sourceCtx := *s.config.SystemContext   // A shallow copy we can modify
+	sourceCtx, err := s.contextForNamespace(pullArgs.namespace)
+	if err != nil {
+		return "", fmt.Errorf("get context for namespace: %w", err)
+	}
+	log.Debugf(ctx, "Using pull policy path for image %s: %s", pullArgs.image, sourceCtx.SignaturePolicyPath)
```
mtrmac, line 154: This duplicates already existing code below.

server/container_create.go
```diff
@@ -299,6 +300,26 @@ func generateUserString(username, imageUser string, uid *types.Int64Value) strin
 func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainerRequest) (res *types.CreateContainerResponse, retErr error) {
 	log.Infof(ctx, "Creating container: %s", translateLabelsToDescription(req.GetConfig().GetLabels()))
 
+	if req.SandboxConfig != nil && req.SandboxConfig.Metadata != nil && req.Config != nil && req.Config.Image != nil {
```
mtrmac, line 303: AFAICS this does not work at all.

`req.Config.Image` is set, ultimately, from either CRI-O’s `ImageStatusResponse.Image.Id` (does not carry a repo information of any kind), or from CRI-O’s `PullImageResponse.ImageRef` (either an image ID again, or a `RepoDigests` value that might even point to a wrong repo).

Either way this does does not provide the information about the expected image identity, necessary to both select the right policy scope, and to correctly protect against image substitution.

At a first glance it seems to me that this would require a change in the CRI, to pass the original user-requested image name.

mtrmac (changes requested):
I’m afraid I think this does not work, and I can’t see any easy way to make it work.

internal/storage/image.go
saschagrunert: I added a cache around the policy context.

internal/storage/image.go
```diff
@@ -398,6 +404,56 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	log.Debugf(ctx, "Checking if image %s is allowed to run", imageName)
+
+	ref, err := svc.getRef(imageName)
+	if err != nil {
+		return false, fmt.Errorf("get ref for name or ID: %w", err)
+	}
+
+	// Check the cache first
+	var policyContext *signature.PolicyContext
+	item := svc.policyContexts.Get(imageName)
```
mtrmac, line 417: This needs to be keyed on `systemContext` — or probably directly on the namespace name CRI-O uses to decide on which policy to look for.

Otherwise two different namespaces with two different policies, but pulling the same `imageName`, would interfere.

internal/storage/image.go
```diff
@@ -859,13 +915,21 @@ func GetImageService(ctx context.Context, store storage.Store, serverConfig *con
...
+		defaultCacheTimeout time.Duration = 24 * time.Hour
+	)
 	is := &imageService{
 		lookup:               ils,
 		store:                store,
 		imageCache:           make(map[string]imageCacheItem),
 		ctx:                  ctx,
 		config:               serverConfig,
 		regexForPinnedImages: CompileRegexpsForPinnedImages(serverConfig.PinnedImages),
+		policyContexts: ttlcache.New(
```
mtrmac, line 929: It might be useful for this to interact with CRI-O’s existing “trigger a reload” features.

server/container_create.go
```diff
@@ -299,6 +300,28 @@ func generateUserString(username, imageUser string, uid *types.Int64Value) strin
...
+	if req.Config.UserSpecifiedImage != "" {
+		systemCtx, err := s.contextForNamespace(req.SandboxConfig.Metadata.Namespace)
+		if err != nil {
+			return nil, fmt.Errorf("get context for namespace: %w", err)
+		}
+
+		img := req.Config.UserSpecifiedImage
+		log.Debugf(ctx, "Using pull policy path for image %s: %s", img, systemCtx.SignaturePolicyPath)
+
+		isAllowed, err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, img)
```
mtrmac, line 312: This **must** pair the image ID used to work with the image with the `UserSpecifiedImage` value. Checking only one of these is not sufficient.

internal/storage/image.go
saschagrunert: Ah makes sense, I'm using `systemContext.SignaturePolicyPath` now, which should be implicitly refer to the namespace.

internal/storage/image.go
saschagrunert: Yep, done :+1:

server/container_create.go
mtrmac: If https://github.com/cri-o/cri-o/pull/7046#discussion_r1228433790 is correct, `IsRunningImageAllowed(…Image)` is not possible; and making two different signature checks on two possibly different objects seems not useful to me, or at least I can’t see the relationship.

This should:
- Resolve the `config.Image.Image` into a specific immutable c/storage image (either `storage.Image` object, or an `storage.Image.ID` value); or, maybe, to a specific manifest digest on a registry.
- Construct a `types.ImageSource` for exactly that immutable image.
- Validate the that immutable image, against the `UserSpecifiedImage`: use `UserSpecifiedImage` to look up the right policy scope, and verify the identity in the signature against `UserSpecifiedImage`
- Never do another `config.Image.Image` lookup (noting that `getRef` accepts names that can move); do all other operations involved in the container creation against the already-resolved immutable image identity.

internal/storage/image.go
```diff
@@ -398,6 +404,75 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
 	return &result, nil
 }
 
+func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, configImage, userSpecifiedImage string) (bool, error) {
+	// Resolve the config.Image.Image into a specific immutable c/storage image
+	configImageRef, err := istorage.Transport.ParseStoreReference(svc.store, configImage)
```
mtrmac, line 409: (I’m not immediately sure this works… if  https://github.com/cri-o/cri-o/pull/7046#discussion_r1228433790 is correct, the input might not always be an image ID.

OTOH if it is either an image ID or a RepoDigests value, and _nothing else_, the digest should fix a specific image, assuming no multi-arch shenanigans.

Ideally it would be possible to turn a `storage.Transport` `ImageReference` directly into a specific `*store.Image`; `storageTransport.GetStoreImage` seems intended to do that but it might not be quite good enough right now. So we might need another function there.

---

Ideally, at some point the input would be resolved into a `*storage.Image` or an image ID, and then create a reference using `ParseStoreReference("@"+id)` == `NewStoreReference(…, nil, id)`.

server/container_create.go
mtrmac: > > * Resolve the `config.Image.Image` into a specific immutable c/storage image (either `storage.Image` object, or an `storage.Image.ID` value); or, maybe, to a specific manifest digest on a registry.
> > * Construct a `types.ImageSource` for exactly that immutable image.
> 
> I've done this in https://github.com/cri-o/cri-o/pull/7046/files#diff-73f26dcabd7e433970fc0612ff0068d2d8f7cd61dda790bcd603195197aaad97R408-R418 via:

It’s not 100% clear to me that this will always resolve to the same image — nothing seems to be rejecting `configImage` values that use tags, at least within CRI-O. The lax parsing is not helping us here.

If the input were only an image ID, that would be clearly fine. If the input is a RepoDigests value, … maybe?

A tag would be a problem.

I’d rather prefer for this to be guaranteed to be immutable by the CRI-O codebase, not relying on Kubelet to never change what it submits (especially considering how underdocumented assumptions of the CRI API are).

---

>  > * Validate the that immutable image, against the `UserSpecifiedImage`: use `UserSpecifiedImage` to look up the right policy scope, and verify the identity in the signature against `UserSpecifiedImage`
> 
> I think I require more help on the API usage here. How can I move from the `UserSpecifiedImage` to a policy scope? Where is the link between the previously resolved store image and the resulting policy scope?

That’s one of the ways resolving an already-pulled image gets ”interesting” :)

Ordinarily, a signature verification during pull uses a `docker.Transport` implementation, so the signature is read from a registry and uses the policy section for `docker`.

This code would want to read an image from the c/storage transport, but still use a policy section for `docker`.

In practice, that means creating a new private `ImageReference` implementation, one that returns a `docker`-like values in `Transport()`, `PolicyConfigurationIdentity()` and `PolicyConfigurationNamespaces()`, but a `containers-storage` implementation of `NewImageSource`; and giving such an `ImageReference` implementation to `policyContext.IsRunningImageAllowed`. Probably use the existing `c/image/docker.{Parse,New}Reference`, and override the `NewImage*` methods.

> > * Never do another `config.Image.Image` lookup (noting that `getRef` accepts names that can move); do all other operations involved in the container creation against the already-resolved immutable image identity.
> 
> I think we only have lookups later for checkpoint-restore, which does an `ImageStatus` at some later point.

I would think the `config.Image.Image` must _somehow_ in actually creating the container. And that code path should also make sure to only refer to the already-resolved immutable identity. E.g. https://github.com/cri-o/cri-o/blob/4c8758b99741b5a0cf046271963ad759aeffe0be/server/container_create_linux.go#L229  is doing _some kind_ of lookup.

(I don’t immediately know whether that needs code changes deep in the container creation path, or imposing a strict format restriction on the input `config.Image.Image` field to be clearly immutable. We might want both.)

saschagrunert: Thank you for the insights @mtrmac! The resolving and especially the `ImageReference` implementation worries me with respect to complexity. I'm now thinking if it would be better to go one step back and do a plain image pull on container creation. The biggest question which comes into my mind would be: What would we do with the auth credentials? Would we have to pass them via the CRI? This would be a much bigger change compared to just passing the user specified image.

mtrmac: The _implementation_ of `ImageReference` is really not that bad:
```go
type myReference {
   types.ImageReference
  storage types.ImageReference
}

func new(expectedIdentity string, local *store.Image) types.ImageReference {
    identity, … := docker.ParseReference("//"+expectedIdentity)
    storage, … := storage.NewStoreReference(…, nil, local.ID)
   return &myReference{ImageReference: identity, storage: storage}
}

func (r *myReference) NewImageSource() … {
   return r.storage.NewImageSource()
}
```

To me it seems tedious, not difficult.

---

I think the far more relevant tradeoff is that signatures on encrypted images don’t validate against already-pulled images. vs. that pulling on `CreateContainer` potentially changes semantics (if the reference is to a tag that is moved) _and_ significantly changes the performance profile of that operation.

internal/storage/image.go
```diff
@@ -398,6 +407,75 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	}
+
+	log.Debugf(ctx, "Getting image status of user specified image: %s", configImage)
+	userSpecifiedImageStatus, err := svc.ImageStatus(systemContext, userSpecifiedImage)
+	if err != nil {
+		return fmt.Errorf("image status from user specified image: %w", err)
+	}
+
+	log.Debugf(ctx, "Comparing digest %s with %s, both must be equal.", configImageStatus.Digest, userSpecifiedImageStatus.Digest)
+	if configImageStatus.Digest != userSpecifiedImageStatus.Digest {
```
mtrmac, line 424: - I don’t think this works reliably if `UserSpecifiedImage` is a moving tag. What really matters is whether the image we want to use was signed with the requested tag, more than whether the registry _currently_ points at the image we want to use, especially in the current model where Kubelet triggers a pull by name, gets an image ID, and then requests to run that ID, not the original name.
  - … and really I think this is unnecessary. Resolve the `configImage` into a specific `*storage.Image` or `*storage.Image.ID`, and then do both a signature check and container start using that image object. Resolving `userSpecifiedImage` into an image ID is not an essential part of the logic.
  - Alternatively, there is a good argument that for moving tags, what is on the registry matters, and that we should not blindly continue to use an older image if the on-registry tag has moved; but in that case this failure would need to trigger a re-pull to update the image!
- Comparing image `ID` is a bit more precise than `Digest`; several digests can be deduplicated to the same ID, and it’s not explicitly promised that `Image.Digest` will not change on such deduplication.

internal/storage/image.go
```diff
@@ -398,6 +407,75 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+
+	log.Debugf(ctx, "Creating image source")
+	imageSrc, err := imageRef.NewImageSource(ctx, systemContext)
+	if err != nil {
+		return fmt.Errorf("create image source: %w", err)
+	}
+
+	log.Debugf(ctx, "Check if image is allowed to run")
+	unparsedImage := cimage.UnparsedInstance(imageSrc, &userSpecifiedImageStatus.Digest)
+	if _, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage); err != nil {
```
mtrmac, line 467: ⚠️ 

This checks whether the on-registry image is correctly signed. But _nothing_ connects the on-registry image artifact with the already-locally-pulled `config.Image`, i.e. this does not verify anything about the image that is about to be run.

mtrmac (changes requested):

internal/storage/image.go
saschagrunert: I removed that check and did the `IsRunningImageAllowed` with `configImage`.

We would need the `userSpecifiedImage` only for moving tags, right?

saschagrunert:
The e2e tests won't work without https://github.com/kubernetes/kubernetes/pull/118652

internal/storage/image.go
```diff
@@ -398,6 +408,94 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+		if err != nil {
+			return fmt.Errorf("create policy context: %w", err)
+		}
+
+		log.Debugf(ctx, "Caching policy context")
+		svc.policyContexts.Set(systemContext.SignaturePolicyPath, policyContext, ttlcache.DefaultTTL)
+	}
+
+	log.Debugf(ctx, "Parsing docker reference from %s", configImageStatus.Name)
+	imageRef, err := docker.ParseReference("//" + configImageStatus.Name)
```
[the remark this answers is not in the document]
saschagrunert: Hm, we probably should use the docker transport of the user specified image: https://github.com/cri-o/cri-o/pull/7046/files#diff-73f26dcabd7e433970fc0612ff0068d2d8f7cd61dda790bcd603195197aaad97R441-R453

internal/storage/image.go
```diff
@@ -398,6 +408,94 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	if err != nil {
+		return err
+	}
+
+	// Add a dummy auth for unauthenticated verifications
+	if len(userSpecifiedImageAuth) == 0 {
+		userSpecifiedImageAuth = []*runtimeapi.AuthConfig{nil}
+	}
+
+	for _, auth := range userSpecifiedImageAuth {
```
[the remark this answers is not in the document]
saschagrunert: You mean like this?

``golang
	// Validate the policy for already pulled images
	policy, err := signature.DefaultPolicy(systemContext)
	if err != nil {
		return err
	}

	policyContext, err := signature.NewPolicyContext(policy)
	if err != nil {
		return err
	}

	ref, err := svc.getRef(configImage)
	if err != nil {
		return err
	}

	img, err := istorage.Transport.GetStoreImage(svc.store, ref)
	if err != nil {
		return err
	}

	src, err := ref.NewImageSource(ctx, systemContext)
	if err != nil {
		return err
	}

	unparsedImage := cimage.UnparsedInstance(src, &img.Digest)
	if _, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage); err != nil {
		return err
	}
```

This still needs the auth credentials for the manifest lookup:
```
DEBU[2023-06-27 10:59:02.576747946+02:00] Accessing "quay.io/saschagrunert/signed:latest" failed: reading manifest latest in quay.io/saschagrunert/signed: unauthorized: access to the requested resource is not authorized  file="docker/docker_image_src.go:95"
```

**Comment by saschagrunert** [comment]
/hold 

for the `TODO`

**Comment by saschagrunert** [comment]
Needs https://github.com/cri-o/cri-o/pull/7121 and https://github.com/kubernetes-sigs/cri-tools/pull/1211

File: server/server.go [hunk 14]
```diff
@@ -103,8 +103,9 @@ type Server struct {
 type pullArguments struct {
 	image         string
 	sandboxCgroup string
-	credentials   imageTypes.DockerAuthConfig
+	credentials   imageTypes.DockerAuthConfig // parsed auth configuration
 	namespace     string
+	auth          *types.AuthConfig // original auth configuration
```
mtrmac, line 108: See how `s.pullOperationsInProgress` is indexed, all of `pullArguments` is compared _by value_.

Including a pointer into the value, one that is quite likely to be different on each operation, AFAICS completely breaks this, because no new pull request is likely to match any existing `pullOperationsInProgress` member.

internal/storage/image.go
```diff
@@ -398,6 +412,145 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
 
+// nolint:staticcheck
+func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, configImage, userSpecifiedImage string, auth *runtimeapi.AuthConfig) error {
+	if userSpecifiedImage == "" {
+		if systemContext.SignaturePolicyPath == "" {
+			log.Infof(ctx, "User specified image is empty and no policy applied, skipping check if image is allowed to run")
+			return nil
+		}
+
+		return errors.New("user specified image is empty")
```
mtrmac, line 423: Eventually this should probably be more helpful to users — ”Kubelet did not set the CRI Config.Image.UserSpecifiedImage field, please update it to use signature policies” or something.

internal/storage/image.go
```diff
@@ -398,6 +412,145 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	}
+
+	log.Debugf(ctx, "Parsing docker reference from %s", userSpecifiedImage)
+	imageRef, err := docker.ParseReference("//" + userSpecifiedImage)
+	if err != nil {
+		return err
+	}
+
+	// Check the auth cache first
+	authCacheItem := svc.authCache.Get(userSpecifiedImage)
```
mtrmac, line 460: ⚠️ Doesn’t this leak credentials across namespaces = tenants? Many tenants can use the same `example.com/vendor` repositories, each with their own credentials (and that might conceivably affect billing or usage metrics per customer).

---

(A reminder: https://github.com/cri-o/cri-o/pull/7046#discussion_r1240051558 is still an issue, and maybe needs solving first, to see if this cache is even needed.)

server/server.go
saschagrunert: Hm, yeah I removed support for auth for now, I think we need a different way to make it work.

internal/storage/image.go
saschagrunert: I removed the auth cache for now, private registries can be supported later on.

> This checks whether the on-registry image is correctly signed. But nothing connects the on-registry image artifact with the already-locally-pulled config.Image, i.e. this does not verify anything about the image that is about to be run.

I tried something similar like we do in image pull, but I'm absolutely unsure if that's the right way: https://github.com/cri-o/cri-o/pull/7046/files#diff-73f26dcabd7e433970fc0612ff0068d2d8f7cd61dda790bcd603195197aaad97R465-R473

saschagrunert:
Green tests @cri-o/cri-o-maintainers PTAL

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
 	return &result, nil
 }
 
+// nolint:staticcheck
+func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, configImage, userSpecifiedImage string) error {
+	if systemContext.SignaturePolicyPath == "" {
+		log.Infof(ctx, "No policy applied, skipping check if image is allowed to run")
+		return nil
+	} else if userSpecifiedImage == "" {
```
haircommander, line 418: nit: it's my preference to not have the else if we return in the previous branch, but if this is better for you :shrug:

haircommander:
one very small nit, otherwise LGTM

I'd love a final +1 from @mtrmac before we merge

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+		return fmt.Errorf("create image source: %w", err)
+	}
+
+	// Connect the on-registry image with the already locally-pulled one
+	imageCloser, err := svc.PrepareImage(systemContext, userSpecifiedImage)
+	if err != nil {
+		return fmt.Errorf("prepare user specified image: %w", err)
+	}
+	defer imageCloser.Close()
+	if imageCloser.ConfigInfo().Digest.Encoded() != configImageStatus.ID {
```
mtrmac, line 471: ⚠️ 

Uh… comparing `ConfigInfo.Digest` values _could_ be sufficient to establish image identity in theory, but it isn’t right now; the way image pull works, nothing ever cares about `config.RootFS.DiffIDs`, so “the same config” does not imply “the same layers”.

That didn’t previously matter in the `PullImage` “a correctly-signed image already exists locally” fast path, because the signature policy in the original pull and on reuse would have been the same.[1]

That can’t be assumed now, and the registry could be colluding with one of the tenants. So an attacker tenant could trigger a pull of a malicious image with a config digest matching a victim image (with a permissive policy, or with a policy that requires a signature that exists for the malicious image), and then instruct the registry to serve the correctly-signed victim image. A victim tenant would then check the signature for the victim image, see that the config digest matches, and accept the already-pulled malicious image.

We would need to match the whole manifest bytes (or equivalently the manifest digest) from the signature-verified `UnparsedImage`, and match that against _one of_ the manifests stored with a `*storage.Image`. That sounds… possible but tricky and difficult.

[1](And it doesn’t matter in `podman pull` or CRI-O’s `ImagePull`, other than the “image exists” shortcut added in this PR, because on those paths we first deduplicate layers based on trusted data, and require layers to be equal while deduplicating by config digest.)

server/image_pull.go
```diff
@@ -193,6 +205,10 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
 				log.Debugf(ctx, "Image %s already in store, skipping pull", img)
```
mtrmac, line 205: ⚠️ Now that policies can vary across tenants after #7016, a config digest comparison is not sufficient to establish trust; see elsewhere.

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+		return errors.New("local and remote image identifiers do not match")
+	}
+
+	log.Debugf(ctx, "Check if image %s is allowed to run", userSpecifiedImage)
+	unparsedImage := cimage.UnparsedInstance(imageSrc, &configImageStatus.Digest)
```
mtrmac, line 476: Yes, for multi-arch images this must choose the right instance … but `configImageStatus.Digest` could be a value from _any_ of the deduplicated images; it might not be a digest that exists on this remote registry/repo (or it might exist but it might not be signed with `userSpecifiedImage`.

The right one can be determined by
- if the input reference is a repo@digest one, doing a storage lookup on that repo@digest
- If the input reference is an image ID, doing a storage lookup on userSpecifiedImage@imageID (which might be repo@digest@imageID)

… and, in both cases, the @digest value might refer to a manifest list; in that case it needs to be turned into the right per-arch instance if `manifest.MIMETypeIsMultiImage`, via `manifest.ListFromBlob` and the like. (I vaguely remember we’ve run into this lookup needing an enhancement when dealing with multiple architectures, but I can’t now find that CRI-O issue.)

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	if _, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage); err != nil {
+		var policyErr signature.PolicyRequirementError
+		if errors.As(err, &policyErr) {
+			return fmt.Errorf("%v: %w", crierrors.ErrSignatureValidationFailed, err)
```
mtrmac, line 480: (Note that this doesn’t create an error that _matches_ `crierrors.ErrSignatureValidationFailed` on a comparison. I guess that doesn’t matter?)

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+		return fmt.Errorf("prepare user specified image: %w", err)
+	}
+	defer imageCloser.Close()
+	if imageCloser.ConfigInfo().Digest.Encoded() != configImageStatus.ID {
+		return errors.New("local and remote image identifiers do not match")
+	}
+
+	log.Debugf(ctx, "Check if image %s is allowed to run", userSpecifiedImage)
+	unparsedImage := cimage.UnparsedInstance(imageSrc, &configImageStatus.Digest)
+	if _, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage); err != nil {
```
mtrmac, line 477: Non-blocking: 

For belt-and-suspenders paranoia,
```go
allowed, err := …
if err != nil { … }
if !allowed { return fmt.Errorf("internal error: running image not allowed but no error returned") }
```
? Admittedly that extra return value really has no benefit, it just asks for this added ceremony.

mtrmac (commented):
I’m afraid this doesn’t quite work … and it hints at various difficulties in making it work in presence of deduplication.

Major items of concern:
- After #7016, the risk profile of referring to images by name changes — every unprivileged tenant can trigger a re-pull / re-tag of a name (assuming a cooperating, or already-compromised) registry, so the code must be _much_ more careful about resolving names to immutable IDs and using IDs to identify images. That certainly applies to how ImagePull turns a name:tag into an image ID returned to Kubelet, and I didn’t check the rest of the codebase for possible other issues.
- Checking for the same config digest is not sufficient to establish image identity.
- Even if we have a local image ID and `userSpecifiedImage`, turning that into the right manifest digest requires extra code.

The “same config digest” problem is really worrisome… There are some ways to solve it (e.g. c/image could add a check of config’s `RootFS.DiffIDs`), but still, that would be a fair bit of complexity in the critical path of the signature/identity check, and that complexity would not be invoked on any other code path.

So I now think that verifying the locally-pulled image against the `docker` policy, per https://github.com/cri-o/cri-o/pull/7046#discussion_r1238841665 , is really the better option. That does require some tedious code to find the right manifest to check signatures for (and breaks for encrypted images, which we can hopefully get fixed in the future), but at least the signature verification code is then, I think a very straightforward thing that directly verifies signatures, and there is no need to worry about deduplicating “what was signed” vs. “what we care about” — and the whole set of concerns around registry credentials for mirrors goes away permanently.

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
 	return &result, nil
 }
 
+// nolint:staticcheck
+func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, configImage, userSpecifiedImage string) error {
+	if systemContext.SignaturePolicyPath == "" {
+		log.Infof(ctx, "No policy applied, skipping check if image is allowed to run")
```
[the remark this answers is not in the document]
mtrmac: If this does have that effect… I wonder where we should have a test that would notice that. It’s perfectly reasonable that CRI-O’s test set `--signature-policy` to be isolated form the rest of the host system, but that means that this situation isn’t tested.

internal/storage/image.go
mtrmac: IOW, _within local c/storage_, the same config digest implies the same image ID implies exactly the same image.

On registries on the internet, the same config digest (without comparing the rest of the manifest, and with a pull implementation that doesn’t care about config’s `RootFS.DiffIDs`) does not guarantee exactly the same image _if the two were pulled_.

saschagrunert:
> So I now think that verifying the locally-pulled image against the `docker` policy, per [#7046 (comment)](https://github.com/cri-o/cri-o/pull/7046#discussion_r1238841665) , is really the better option. That does require some tedious code to find the right manifest to check signatures for (and breaks for encrypted images, which we can hopefully get fixed in the future)

I'd like to address this before going to the other review points, but I think I really need your help and experience with that. I added a commit to move to the custom reference implementation in https://github.com/cri-o/cri-o/pull/7046/commits/a917606100342941b3c55b70004fad9bd6926746, where this part seems clearly wrong, because we have to find the right manifest (integration tests are failing now):

```go
unparsedImage := cimage.UnparsedInstance(imageSrc, &storageImage.Digest)
```

Do you have any insights how we could achieve that with the existing c/image/storage API?

mtrmac:
> we have to find the right manifest

I was thinking something like:
```go
func newReference(…) … {
   dockerRef := reference.ParseNormalizedNamed(expectedIdentity)
   identity := docker.NewReference(dockerRef)

   // risk: this may fail if the image was re-pulled and re-tagged. In that case, try with (store, nil, local.ID) and hope for the best? Or try to find a RepoDigests value that matches dockerRef.Name(), but that can also be ambiguous and incorrect.
   // ideally PullImage should return (the single correct!) RepoDigests value, and that should be available here.
   // we might need to extend c/image to choose a manifest by digest without dockerRef.Name() matching.
   storageRef := storage.Transport.NewStoreReference(store, dockerRef, local.ID)
}

// then
var unparsedInstance
unparsedToplevel := image.UnparsedInstance(rawSource, nil)
topManifest, topMIMEType, … := unparsedToplevel.Manifest()
if manifest.MIMETypeIsMultiImage(topMimeType) {
   // risk: this can theoretically break if the image was pulled using a different architecture configuration than the one we are using now, and find an instance that is not available locally.
   instanceDigest := manifest.ListFromBlob(topManifest, toMIMEType).ChooseInstance()
   unparsedInstance = image.unparsedInstance(imageSrc, &instanceDigest)
} else {
  unparsedInstance = unparsedToplevel
}
```

Warning: This is untested; let me know if I should experiment and try to get a full prototype.

saschagrunert:
> let me know if I should experiment and try to get a full prototype.

That would be wonderful! :pray: 

I added your latest suggestions on top of the last commit of this PR. A test like this still fails:

https://github.com/cri-o/cri-o/blob/a917606100342941b3c55b70004fad9bd6926746/test/policy.bats#L113-L127

With:

```
# time="2023-07-24 09:38:58.934493937+02:00" level=info msg="Creating container: //" file="server/container_create.go:314" id=a6784ec9-1cdc-41e3-a273-57596d03fabc name=/runtime.v1.RuntimeService/CreateContainer
# time="2023-07-24 09:38:58.934515856+02:00" level=debug msg="Creating policy context" file="storage/image.go:429" id=a6784ec9-1cdc-41e3-a273-57596d03fabc name=/runtime.v1.RuntimeService/CreateContainer
# time="2023-07-24 09:38:58.934702811+02:00" level=debug msg="Caching policy context" file="storage/image.go:440" id=a6784ec9-1cdc-41e3-a273-57596d03fabc name=/runtime.v1.RuntimeService/CreateContainer
# time="2023-07-24 09:38:58.934960631+02:00" level=debug msg="parsed reference into \"[overlay@/tmp/tmp.FVLVxmxWFa/crio+/tmp/tmp.FVLVxmxWFa/crio-run]quay.io/crio/signed:latest\"" file="storage/storage_transport.go:185"
# time="2023-07-24 09:38:58.935008773+02:00" level=debug msg="Creating image source" file="storage/image.go:459" id=a6784ec9-1cdc-41e3-a273-57596d03fabc name=/runtime.v1.RuntimeService/CreateContainer
# time="2023-07-24 09:38:58.935063840+02:00" level=debug msg="Check if image quay.io/crio/signed is allowed to run" file="storage/image.go:492" id=a6784ec9-1cdc-41e3-a273-57596d03fabc name=/runtime.v1.RuntimeService/CreateContainer
# time="2023-07-24 09:38:58.935079734+02:00" level=debug msg="IsRunningImageAllowed for image containers-storage:[overlay@/tmp/tmp.FVLVxmxWFa/crio]quay.io/crio/signed:latest@acfccbcc273e8a4d9bca765205bf69d3210d4da1a73bcc544e338c276fe8c744" file="signature/policy_eval.go:274"
# time="2023-07-24 09:38:58.935098902+02:00" level=debug msg=" Using default policy section" file="signature/policy_eval.go:162"
# time="2023-07-24 09:38:58.935111726+02:00" level=debug msg="Requirement 0: denied, done" file="signature/policy_eval.go:285"
# time="2023-07-24 09:38:58.935164128+02:00" level=debug msg="Response error: SignatureValidationFailed: Running image containers-storage:[overlay@/tmp/tmp.FVLVxmxWFa/crio+/tmp/tmp.FVLVxmxWFa/crio-run]quay.io/crio/signed:latest@acfccbcc273e8a4d9bca765205bf69d3210d4da1a73bcc544e338c276fe8c744 is rejected by policy." file="otel-collector/interceptors.go:71" id=a6784ec9-1cdc-41e3-a273-57596d03fabc name=/runtime.v1.RuntimeService/CreateContainer
```

While it should be able to start the container. :thinking:

mtrmac:
I’m afraid I have been pointing out an impossible path, sorry about that.

The trouble is:
- I was wrong about how to override the identity when checking a signature of a local image; `IsRunningImage` consumes an `UnparsedImage`, so that’s what needs adjusting; easy enough…
- But doing that needs access to currently-c/image-private `internal/unparsedimage` and `internal/private` to support _sigstore_ signatures. An override/wrapper can, apparently, only override/wrap publicly-defined interface method, and the API-frozen `types.ImageSource` can not express sigstore signatures, only simple-signing ones. To this point, all of the c/image code dealing with sigstore signatures has been private.

---

As a prototype, with https://github.com/containers/image/pull/2056 this seems to work:

```console
% cat registries-d/default.yaml 
default-docker:
 use-sigstore-attachments: true
% go run ./main.go ./crio-policy.json $(pwd)/t quay.io/crio/signed
```
with
```go
package main

import (
	"context"
	"errors"
	"fmt"
	"io"
	"os"
	"path/filepath"

	"github.com/containers/image/v5/copy"
	"github.com/containers/image/v5/docker"
	"github.com/containers/image/v5/docker/reference"
	"github.com/containers/image/v5/image"
	"github.com/containers/image/v5/manifest"
	"github.com/containers/image/v5/signature"
	istorage "github.com/containers/image/v5/storage"
	"github.com/containers/image/v5/transports"
	"github.com/containers/image/v5/types"
	"github.com/containers/storage"
	"github.com/containers/storage/pkg/reexec"
	"github.com/opencontainers/go-digest"
	"github.com/sirupsen/logrus"
)

func tempStore(storeRoot string) (storage.Store, error) {
	// uidMap := []idtools.IDMap{{
	// 	ContainerID: 0,
	// 	HostID:      os.Getuid(),
	// 	Size:        1,
	// }}
	// gidMap := []idtools.IDMap{{
	// 	ContainerID: 0,
	// 	HostID:      os.Getgid(),
	// 	Size:        1,
	// }}
	return storage.GetStore(storage.StoreOptions{
		RunRoot:            filepath.Join(storeRoot, "run"),
		GraphRoot:          filepath.Join(storeRoot, "root"),
		GraphDriverName:    "vfs",
		GraphDriverOptions: []string{"vfs.ignore_chown_errors=true"},
		// UIDMap:             uidMap,
		// GIDMap:             gidMap,
	})
}

func pull(sys *types.SystemContext, policyContext *signature.PolicyContext, store storage.Store, userSpecifiedImage reference.Named) (digest.Digest, error) {
	srcRef, err := docker.NewReference(userSpecifiedImage)
	if err != nil {
		return "", fmt.Errorf("creating docker:// reference for %q: %w", userSpecifiedImage.String(), err)
	}

	destRef, err := istorage.Transport.NewStoreReference(store, userSpecifiedImage, "")
	if err != nil {
		return "", fmt.Errorf("creating containers-storage: reference for %q: %w", userSpecifiedImage.String(), err)
	}

	manifestBytes, err := copy.Image(context.TODO(), policyContext, destRef, srcRef, &copy.Options{
		SourceCtx:      sys,
		DestinationCtx: sys,
		ReportWriter:   io.MultiWriter(os.Stdout), // The MultiWriter is a hack to hide os.Stdout being a TTY, so that we don't get progress bars and they don't interfere with the debug log
	})
	if err != nil {
		return "", fmt.Errorf("pulling %q to %q: %w", transports.ImageName(srcRef), transports.ImageName(destRef), err)
	}
	return digest.FromBytes(manifestBytes), nil
}

func checkSig(sys *types.SystemContext, policyContext *signature.PolicyContext, store storage.Store, userSpecifiedImage reference.Named, localManifestDigest digest.Digest) error {
	userIdentityRef, err := docker.NewReference(userSpecifiedImage)
	if err != nil {
		return fmt.Errorf("creating docker:// reference for %q: %w", userSpecifiedImage.String(), err)
	}

	storeNamedReference, err := reference.WithDigest(reference.TrimNamed(userSpecifiedImage), localManifestDigest)
	if err != nil {
		return fmt.Errorf("creating a digested storage reference for %q + %q: %w", userSpecifiedImage, localManifestDigest, err)
	}
	storageRefWithoutID, err := istorage.Transport.NewStoreReference(store, storeNamedReference, "")
	if err != nil {
		return fmt.Errorf("creating containers-storage: reference for %q: %w", storeNamedReference.String(), err)
	}
	logrus.Debugf("Created storageRefWithoutID = %q", transports.ImageName(storageRefWithoutID))

	// storageRefWithoutID should be good enough, but just to show the workflow with a known local image ID:
	localImage, err := istorage.Transport.GetStoreImage(store, storageRefWithoutID)
	if err != nil {
		return fmt.Errorf("looking up a local image for %q: %w", transports.ImageName(storageRefWithoutID), err)
	}
	storageRef, err := istorage.Transport.NewStoreReference(store, storeNamedReference, localImage.ID)
	if err != nil {
		return fmt.Errorf("creating containers-storage: reference for %q + %q: %w", storeNamedReference.String(), localImage.ID, err)
	}
	logrus.Debugf("Created storageRef = %q", transports.ImageName(storageRef))

	storageSource, err := storageRef.NewImageSource(context.TODO(), sys)
	if err != nil {
		return fmt.Errorf("creating image source for local store image: %w", err)
	}
	defer storageSource.Close()

	unparsedToplevel := image.UnparsedInstance(storageSource, nil)
	topManifest, topMIMEType, err := unparsedToplevel.Manifest(context.TODO())
	if err != nil {
		return fmt.Errorf("get top level manifest: %w", err)
	}

	var unparsedInstance *image.UnparsedImage
	if manifest.MIMETypeIsMultiImage(topMIMEType) {
		manifestList, err := manifest.ListFromBlob(topManifest, topMIMEType)
		if err != nil {
			return fmt.Errorf("parsing list manifest: %w", err)
		}

		instanceDigest, err := manifestList.ChooseInstance(sys)
		if err != nil {
			return fmt.Errorf("choosing instance: %w", err)
		}

		unparsedInstance = image.UnparsedInstance(storageSource, &instanceDigest)
	} else {
		unparsedInstance = unparsedToplevel
	}

	mixedUnparsedInstance := image.UnparsedInstanceWithReference(unparsedInstance, userIdentityRef)

	allowed, err := policyContext.IsRunningImageAllowed(context.TODO(), mixedUnparsedInstance)
	if err != nil {
		return fmt.Errorf("verifying signatures: %w", err)
	}
	if !allowed {
		return fmt.Errorf("internal inconsistency: !allowed and no error")
	}

	return nil
}

func main2() error {
	if len(os.Args) != 4 {
		return errors.New("Expected ./policy ./$program ./temp-store-dir docker/reference:tag@or-digest")
	}
	policyPath := os.Args[1]
	storeRoot := os.Args[2]
	userSpecifiedImageStr := os.Args[3]

	logrus.SetLevel(logrus.DebugLevel)

	registriesDDir, err := filepath.Abs("./registries-d") // NOTE: A LOCAL HACK, to set use-sigstore-attachments: true
	if err != nil {
		return fmt.Errorf("registries-d: %w", err)
	}
	sys := &types.SystemContext{
		OSChoice:          "linux", // A hack to get usable images on a mac
		RegistriesDirPath: registriesDDir,
	}

	userSpecifiedImage, err := reference.ParseNormalizedNamed(userSpecifiedImageStr)
	if err != nil {
		return fmt.Errorf("parsing input: %w", err)
	}
	userSpecifiedImage = reference.TagNameOnly(userSpecifiedImage)

	store, err := tempStore(storeRoot)
	if err != nil {
		return fmt.Errorf("initializing store at %q: %w", storeRoot, err)
	}

	policy, err := signature.DefaultPolicy(&types.SystemContext{SignaturePolicyPath: policyPath})
	if err != nil {
		return fmt.Errorf("loading signature policy: %w", err)
	}
	policyContext, err := signature.NewPolicyContext(policy)
	if err != nil {
		return fmt.Errorf("creating policy context: %w", err)
	}
	defer policyContext.Destroy()

	logrus.Infof("Pulling...")
	localManifestDigest, err := pull(sys, policyContext, store, userSpecifiedImage)
	if err != nil {
		return fmt.Errorf("pulling %q: %w", userSpecifiedImage.String(), err)
	}
	logrus.Infof("... pulling done, local digest %s", localManifestDigest)

	if err := checkSig(sys, policyContext, store, userSpecifiedImage, localManifestDigest); err != nil {
		return fmt.Errorf("checking signature of %q: %w", userSpecifiedImage.String(), err)
	}

	return nil
}

func main() {
	if reexec.Init() {
		return
	}
	if err := main2(); err != nil {
		logrus.Errorf("%v", err)
		os.Exit(1)
	}
}
```

This also demonstrates how to turn the result of `copy.Image` into an image reference which is not redirected by tag changes (but, I think, may still be _broken_ by tag changes). (And if that is _always_ how the storage reference is obtained, the signature checking code does not need to deal with the `MIMETypeIsMultiImage` case at all, the digest reference points at the right-per-architecture image. But that might not be the case for the “image already exists” shortcut path, I’m not immediately sure.)

---

This is only smoke-tested with that one image; I didn’t test multi-arch handling at all, and I’m not immediately sure what else might need adjusting to make this viable on the “pull found a local image” shortcut. I’m filing that PR as a draft to hopefully unblock your progress.

internal/storage/image.go
saschagrunert: Yep, that's totally intentional.

server/image_pull.go
```diff
@@ -130,22 +147,17 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
-	sourceCtx := *s.config.SystemContext   // A shallow copy we can modify
+	sourceCtx, err := s.contextForNamespace(pullArgs.namespace)
+	if err != nil {
+		return "", fmt.Errorf("get context for namespace: %w", err)
+	}
+	log.Debugf(ctx, "Using pull policy path for image %s: %s", pullArgs.image, sourceCtx.SignaturePolicyPath)
```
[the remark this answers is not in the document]
saschagrunert: Changed it to become a `%q`.

saschagrunert:
That's awesome, thank you so much @mtrmac! 

critest needs https://github.com/kubernetes-sigs/cri-tools/pull/1224, let's see if the integration tests pass.

internal/storage/image.go
saschagrunert: This does not exist any more.

test/critest.bats
```diff
@@ -8,6 +8,7 @@ function setup() {
 	fi
 
 	export CONTAINER_SECCOMP_USE_DEFAULT_WHEN_EMPTY=false
+	export SIGNATURE_POLICY=""
```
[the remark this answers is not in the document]
saschagrunert: This is gone.

server/image_pull.go
```diff
@@ -193,6 +205,10 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
 				log.Debugf(ctx, "Image %s already in store, skipping pull", img)
 				pulled = img
 
+				if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &sourceCtx, img, img); err != nil {
```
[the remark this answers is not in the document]
saschagrunert: I assume this point is resolved with the latest implementation.

internal/storage/image.go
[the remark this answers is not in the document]
saschagrunert: This code path is not available any more.

internal/storage/image.go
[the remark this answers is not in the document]
saschagrunert: This code path is not available any more.

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+		return nil
+	} else if userSpecifiedImage == "" {
+		return errors.New("client did not set the CRI Config.Image.UserSpecifiedImage field, please update it to use signature policies")
+	}
+
+	log.Debugf(ctx, "Getting image status of config image: %s", configImage)
```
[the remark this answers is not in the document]
saschagrunert: This code path is not available any more.

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	configImageStatus, err := svc.ImageStatus(systemContext, configImage)
+	if err != nil {
+		return fmt.Errorf("image status from config image: %w", err)
+	}
+
+	// Check the cache first
+	var policyContext *signature.PolicyContext
```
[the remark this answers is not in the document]
saschagrunert: Removed the cache because we require parallel access.

internal/storage/image.go
```diff
@@ -398,6 +410,87 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	if err != nil {
+		var e errcode.Error
+		if as := errors.As(err, &e); as && e.ErrorCode().Descriptor().HTTPStatusCode == http.StatusUnauthorized {
+			return fmt.Errorf("verifying policies against private registries it not supported yet: %w", err)
+		}
+		return fmt.Errorf("create image source: %w", err)
+	}
+
+	// Connect the on-registry image with the already locally-pulled one
+	imageCloser, err := svc.PrepareImage(systemContext, userSpecifiedImage)
```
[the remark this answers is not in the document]
saschagrunert: This code path does not exist any more.

internal/storage/image.go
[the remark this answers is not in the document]
saschagrunert: This code path does not exist any more.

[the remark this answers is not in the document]
saschagrunert: This code path does not exist any more.

saschagrunert: This code path does not exist any more.

server/image_pull.go
saschagrunert: This code path does not exist any more.

internal/storage/image.go
saschagrunert: This code path does not exist any more.

saschagrunert:
Tests are green @mtrmac PTAL again.

internal/storage/image.go
```diff
@@ -398,6 +408,145 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	if err != nil {
+		return "", fmt.Errorf("creating docker:// reference for %q: %w", userSpecifiedImage.String(), err)
+	}
+
+	destRef, err := istorage.Transport.NewStoreReference(svc.store, userSpecifiedImage, "")
+	if err != nil {
+		return "", fmt.Errorf("creating containers-storage: reference for %q: %w", userSpecifiedImage.String(), err)
+	}
+
+	manifestBytes, err := copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
```
mtrmac, line 456: Without credentials this doesn’t work for authenticated registries at all.

The prototype’s `manifestBytes` was intended to be collected in `PullImage` and to be used to form the correct `RepoDigests` value.

internal/storage/image.go
```diff
@@ -398,6 +408,145 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+		SourceCtx:      sys,
+		DestinationCtx: sys,
+		ReportWriter:   io.MultiWriter(os.Stdout), // The MultiWriter is a hack to hide os.Stdout being a TTY, so that we don't get progress bars and they don't interfere with the debug log
```
mtrmac, line 459: I don’t think CRI-O needs this particular hack, it isn’t typically run with /dev/stdout being a TTY. That was only a concern for the interactively-run prototype.

internal/storage/image.go
```diff
@@ -398,6 +408,145 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+	if err != nil {
+		return fmt.Errorf("creating a digested storage reference for %q + %q: %w", userSpecifiedImage, localManifestDigest, err)
+	}
+	storageRefWithoutID, err := istorage.Transport.NewStoreReference(svc.store, storeNamedReference, "")
+	if err != nil {
+		return fmt.Errorf("creating containers-storage: reference for %q: %w", storeNamedReference.String(), err)
+	}
+	log.Debugf(ctx, "Created storageRefWithoutID = %q", transports.ImageName(storageRefWithoutID))
+
+	// storageRefWithoutID should be good enough, but just to show the workflow with a known local image ID:
```
mtrmac, line 495: Yeah, CRI-O should use the data it has available.

If it doesn’t have the local image ID, the extra lookup to get an ID and then look up by ID adds not value. OTOH the ImagePull “local image already present” shortcut might need to be ID-based to make sure the signature check, config digest check, and final return value all reliably refer to the same image.

mtrmac (commented):
Just a quick note for faster feedback while I read the code more carefully…

internal/storage/image.go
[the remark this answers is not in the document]
mtrmac: … and is there a design for the upgrade path, then?

(The new code does `reference.ParseNormalizedNamed`, which rejected `""` just like the old code path did.)

internal/storage/image.go
```diff
@@ -398,6 +408,145 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+
+	policy, err := signature.DefaultPolicy(systemContext)
+	if err != nil {
+		return fmt.Errorf("get default policy: %w", err)
+	}
+
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return fmt.Errorf("create policy context: %w", err)
+	}
```
mtrmac, line 427: `policyContext.Destroy` needs to be called eventually.

internal/storage/image.go
```diff
@@ -398,6 +408,145 @@ func (svc *imageService) ImageStatus(systemContext *types.SystemContext, nameOrI
...
+
+	localManifestDigest, err := svc.pullLocalManifestDigest(ctx, systemContext, policyContext, userSpecifiedImageParsed)
+	if err != nil {
+		return fmt.Errorf("pull local manifest digest: %w", err)
+	}
+	if localManifestDigest == "" {
+		return nil
```
mtrmac, line 434: It’s not OK to silently continue on failures.

This effectively gives a malicious registry a way to bypass signature enforcement by maliciously returning an “unauthorized” response.

server/container_create.go
```diff
@@ -359,6 +359,19 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
...
+	if err != nil {
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, req.Config.Image.UserSpecifiedImage); err != nil {
+		return nil, err
+	}
+	log.Debugf(ctx,
+		"Is allowed to run config image %s (user requested image: %s, policy path: %s)",
+		req.Config.Image.Image, req.Config.Image.UserSpecifiedImage, systemCtx.SignaturePolicyPath,
```
mtrmac, line 372: `%q` for policy path?

But really it would make a bit more sense to me to have this kind of logging inside `IsRunningImageAllowed`; that function is more authoritative as to which inputs / environment values are relevant for the policy check, and it would mean the logging is consistent (right now ImagePull is not logging this).

mtrmac (commented):
Older outstanding review comments, for visibility:
- https://github.com/cri-o/cri-o/pull/7046#discussion_r1270015485 on reliably identifying images
- https://github.com/cri-o/cri-o/pull/7046#discussion_r1270041672 on upgrade path
- https://github.com/cri-o/cri-o/pull/7046#discussion_r1270067885 on the “pulled image already exists locally” shortcut

server/container_create.go
```diff
@@ -359,6 +359,19 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 		}, nil
 	}
 
+	systemCtx, err := s.contextForNamespace(req.SandboxConfig.Metadata.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, req.Config.Image.UserSpecifiedImage); err != nil {
```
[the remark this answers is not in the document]
saschagrunert: Does that mean we don't need the `req.Config.Image.UserSpecifiedImage` at all? 

If we switch that to `req.Config.Image.Image`, then we need additional logic for those two tests:

https://github.com/cri-o/cri-o/blob/efd1654b9994ddbc1fd9f563a817d98b128dfd89/test/policy.bats#L144-L175


Otherwise we will get the error:

```
# time="2023-07-28 09:43:28.733989619+02:00" level=debug msg="Response error: parsing user specified image: invalid repository name (2f1bf9c599183e0ef813e9b76a8cdd17c9b2cb224c94b605c178dcef8d3a54e3), cannot specify 64-byte hexadecimal strings" file="otel-collector/interceptors.go:71" id=31cde433-4fda-4373-a473-6f438172e1d7 name=/runtime.v1.RuntimeService/CreateContainer
```

If we have an image ID in `req.Config.Image.Image`, should we then fallback to `req.Config.Image.UserSpecifiedImage`?

internal/storage/image.go
saschagrunert: A Kubernetes e2e test is using a private registry and we get no credentials on container creation. Not sure how we should handle that unsupported use case.

server/container_create.go
saschagrunert: Moved that log into `IsRunningImageAllowed`.

internal/storage/image.go
saschagrunert: Added

internal/storage/image.go
saschagrunert: Removed that one.

server/container_create.go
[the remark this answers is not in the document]
mtrmac: > Does that mean we don't need the `req.Config.Image.UserSpecifiedImage` at all?

No?! Policy verification has three main inputs:
- what is the users’ intent / instruction
- what are the rules for deciding whether something matches the users’ intent
- what is being verified against that intent

For us, the three are
- `UserSpecifiedImage`
- `policy.json`
- A single `*storage.Image`.

---

> If we have an image ID in `req.Config.Image.Image`, should we then fallback to `req.Config.Image.UserSpecifiedImage`?

What would that _mean_ in terms of the three inputs discussed above?

---

The `PullImage`+`CreateContainer` pair needs to be considered together. IIRC the situation is one of:
- a fresh pull
  - Pull the image, with policy enforcement included
  - Get a manifest digest from `copy.Image`, construct _the right_ `RepoDigest` value, return that in `PullImageResponse`
  - Receive exactly that `RepoDigest` value in `Config.Image.Image`
  - Use it to look up a `*storage.Image` (`NewStoreReference(store, repoDigest, "")`).
  - Verify signatures against policy and `UserSpecifiedImage`
- a pre-existing image reuse short path:
  - Find a candidate _commit_ to an `*Image.ID`
  - Verify a registry match
  - Verify that image’s signature on the pull short path (`NewStoreReference(store, nil, imageID)` or `NewStoreReference(store, RepoDigest, imageID)` if we have a repoDigest input — or, well, ideally, get a `*storage.Image` directly from the local lookup without going through an `*Image` -> ID -> `*Image` roundtrip), against the image we were requested to pull
  - Return _the right_ RepoDigest if known, or an image ID (if not-pulling by tag, IIRC) in `PullImageResponse`
  - If we received an ID in `Config.Image.Image`, use it to look up an image, possibly combined with UserSpecifiedImage (same `NewStoreReference` calls)
  - Verify signatures against policy and `UserSpecifiedImage`

internal/storage/image.go
mtrmac: See elsewhere for how I expect the local manifest digest to be involved.

server/container_create.go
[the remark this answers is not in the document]
mtrmac: > Verify a registry match

Hum, that could potentially be used to obtain the right manifest digest if users request a pull by tag. But, in principle, that does matter because…

---

… there is a third situation, per https://github.com/cri-o/cri-o/pull/7046#discussion_r1228433790 : Where there never is an `ImagePull`, and we get a `CreateContainer` with an image ID, based on Kubelet calling `ImageStatus`. So `CreateContainer` must be ready to accept repo-less digest IDs either way.

internal/storage/image.go
saschagrunert: Removed that code path.

server/image_pull.go
```diff
@@ -130,22 +147,17 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
...
-			sourceCtx.SignaturePolicyPath = policyPath
-		} else if !os.IsNotExist(err) {
-			return "", fmt.Errorf("read policy path %s: %w", policyPath, err)
-		}
-	}
-	log.Debugf(ctx, "Using pull policy path for image %s: %s", pullArgs.image, sourceCtx.SignaturePolicyPath)
-
 	decryptConfig, err := getDecryptionKeys(s.config.DecryptionKeysPath)
 	if err != nil {
 		return "", err
```
mtrmac, line 163: There was code here to enforce signatures in the `Image %s already in store` path; why is it safe to remove it?

server/image_pull.go
```diff
@@ -298,13 +310,9 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
 	// Update metric for successful image pulls
 	metrics.Instance().MetricImagePullsSuccessesInc(pulled)
 
-	status, err := s.StorageImageServer().ImageStatus(s.config.SystemContext, pulled)
+	imageRef, err := s.StorageImageServer().PullLocalManifestDigest(ctx, &sourceCtx, pulled)
```
mtrmac, line 313: `PullImage` already has that data internally. This discards that data, and effectively does a whole new pull, with no progress reporting etc.

Among other things, that makes the “image is already in store” optimization completely pointless, because that one tries to avoid issuing even a metadata-only pull.

server/image_pull.go
saschagrunert: I reverted that change, we still require that path when we consider https://github.com/cri-o/cri-o/pull/7046#discussion_r1286303003

server/image_pull.go
saschagrunert: Alright, I moved the logic into `PullImage` of the container server. This way we can directly retrieve the `manifestBytes` there.

PR #7058, opened 2023-06-16
OCPBUGS-13980: runtime handler hooks: run default hooks on graceful container exit

haircommander:
#### What type of PR is this?

/kind bug
#### What this PR does / why we need it:
previously, some containers were missing the default hooks, as they didn't run on container exit.

#### Does this PR introduce a user-facing change?

```release-note
Fix a bug where the PostStop runtime handler hooks weren't being called when a container exited gracefully
```

PR #7059, opened 2023-06-16
server: move post stop container behavior to server.go:handleExit

haircommander:
#### What type of PR is this?

/kind bug
Currently, we're duplicating a lot of the container stop operations. Whenever a container stops, conmon will write the exit file
and CRI-O should catch the exit, no matter whether it was triggered by CRI-O or happened naturally.

Not adding these will prevent duplicated work like UpdateContainerStatus, which calls `runc update` and is not negligible.

```release-note
Reduce duplicated work in container stops
```

### DIFF
--- a/server/container_stop.go
+++ b/server/container_stop.go
@@ -1,8 +1,10 @@
 package server
 
 import (
+	"errors"
 	"fmt"
 
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
 	"github.com/cri-o/cri-o/internal/runtimehandlerhooks"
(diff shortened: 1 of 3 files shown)


haircommander:
carrying https://github.com/cri-o/cri-o/pull/7058 for simplicity

haircommander:
@klihub the reasoning behind the NRI errors aren't clear to me, can you help me find the issue?

haircommander:
I'd like to use this PR to note some oddness about the high performance PreStop hooks: they run at different times depending on how the container stops. If it stops gracefully (not triggered by the kubelet), then the handleExit function will not call the preStop hooks. They will be called in container remove (as it calls containerStop, which calls the hooks), so the hook is functionally a postStop, preRemove hook. for manually stopped containers, the hooks will be called as expected before the container stops.

I don't know the consequences of this, but it does paint an odd state diagram for containers. I think it would be better if the hook behavior was consistent between the two different stopping conditions. cc @MarSik

TomSweeneyRedHat:
LGTM

klihub:
> @klihub the reasoning behind the NRI errors aren't clear to me, can you help me find the issue?

@haircommander The test fails, expecting an NRI StopContainer event which it does not receive after it has stopped the test container. Currently I don't know why that event is not received.

klihub:
> > @klihub the reasoning behind the NRI errors aren't clear to me, can you help me find the issue?

@haircommander Okay, I (think I) know what happens. Before the changes in this PR, the NRI StopContainer event was delivered before the CRI StopContainer call returned but now this is not the case any more (AFAICT). 

The NRI tester acts both as a CRI client and an NRI plugin. The NRI plugin code records all events that the plugin gets. The test cases exercise the CRI interface then wait for and check that the expected events in response show up. Waiting for the container stop event happened with a 0 timeout which used to be okay, but is not with this PR in place.

haircommander:
thanks @klihub applied the patch

MarSik:
@haircommander Anything I can do to help expedite this and the 4.13 backport?

haircommander:
hm still failing
/hold

PR #7069, opened 2023-06-20
Add kata integration tests

littlejawa:
#### What type of PR is this?

/kind ci

We have a job (`kata-jenkins`) running on each PR, that tests the PR with the kata runtime.
This job is using scripts that are in the kata-containers/tests repository. This is cumbersome, as we have to rely on an external repository to fix testing issues that are sometimes specific to cri-o.

In the past, we used to have this test located in the cri-o repository, along with all other integration tests, but it had to be moved away because kata requires virtualization, and the provider we had at the time did not allow nested virtualization.
Now with the newer provider, we can enable these tests again in the cri-o environment.

This PR makes the test run using the existing scripts for cri-o as much as possible, reducing the complexity of the test, and making it easier to maintain in the future.
It also modifies the pervious kata-specific test process in the following way:
- do not install kata from packages. Instead, use the binary archive published by the upstream kata project. This allow to get a newer version than what we may have in packages, and ensure we can decide when we want to update to a new version for testing
- add more test files than just the ctr.bats file. This is in line with what the `kata-jenkins` job is testing today. The current PR does not enable as many tests as the `kata-jenkins` job because I realized the `kata-jenkins` job was skipping all tests in some bats files, which makes little sense. The list I'm proposing in this PR is closer to what can actually be tested with kata at the time being.

If this is approved and merged, the next step will be to enable this test on cri-o PRs (see below).
And ultimately, we will remove the `kata-jenkins` job.

None

I have [a corresponding PR on openshift/release](https://github.com/openshift/release/pull/38121) that I used to test this code in real conditions.

The results of the last run of the test, exercising the code in this PR, is available [here](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/38121/rehearse-38121-pull-ci-littlejawa-cri-o-add_kata_integration_tests-ci-kata-integration/1670821033118011392).

```release-note
None
```


littlejawa:
@wgahnagl - this is the result of our discussion (a LONG time ago).
Can you get a look?

contrib/test/ci/build/cri-o.yml
```diff
@@ -95,17 +95,19 @@
...
-    dest: /etc/crio/crio.conf.d/50-kata.conf
-    content: |
-      [crio.runtime]
-      default_runtime = "kata"
-      [crio.runtime.runtimes.kata]
-      runtime_path = "/usr/bin/containerd-shim-kata-v2"
-      runtime_type = "vm"
-      runtime_root = "/run/vc"
-  when: "build_kata | default(False) | bool"
+#- name: use kata
```
haircommander, line 98: should this be deleted instead?

haircommander:
LGTM when the kata job passes, nice cleanup!

littlejawa:
> LGTM when the kata job passes, nice cleanup!

Thanks!
Yes, the "old" kata CI jobs is trying to prevent the new one from happening :)

contrib/test/ci/build/kata.yml
```diff
@@ -57,96 +57,10 @@
...
-      dest: /usr/share/kata-containers/defaults/configuration.toml
-      line: 'default_vcpus = -1'
-      regexp: '^default_vcpus'
-      state: present
-
-  when: ansible_distribution == "CentOS" and ansible_distribution_major_version|int >= 8
-
+    - name: Get and extract kata tarball from upstream release
+      unarchive:
+        src: "https://github.com/kata-containers/kata-containers/releases/download/3.0.2/kata-static-3.0.2-x86_64.tar.xz"
```
sohankunkerkar, line 64: can we parameterize the version here?

littlejawa: Right. I've added it.
Thanks!

haircommander:
/approve

LGTM, thanks!

PR #7074, opened 2023-06-21
governance: tweak voting behavior

haircommander:
#### What type of PR is this?

/kind documentation
as suggested in https://github.com/cncf/tag-contributor-strategy/issues/436

```release-note
none
```

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -268,9 +268,12 @@ In the case of abstention, an approver will not be included in the voting approv
 group when deciding the majority. In the case of a tie, the issue will not be considered
 to have passed.
 
-Most votes require a simple majority of all voting approvers to succeed. Approvers
-can be removed by a 2/3 majority vote of all voting approvers, and changes to this
-Governance require a 2/3 vote of all approvers.
+Most votes require a simple majority of all voting approvers to succeed.
+Approvers can be removed by a 2/3 majority vote of all voting approvers,
+and changes to this Governance require a 2/3 vote of all approvers. The
+approvers will make every reasonable effort to make sure that any removal
+or amendment to the Governance be affirmed with votes by employees of
+more than one organization.
 
 ## Miscellaneous Administrivia
 


haircommander:
for future reference: changes like this should have a bit more soak time to allow all the maintainers to reach consensus. It's part my fault for marking the change as ready. I don't think we need revert it, but @cri-o/cri-o-maintainers if there are any issues with this change, please chime in and we can revert. In the future, we should wait the full week :)

sohankunkerkar:
> for future reference: changes like this should have a bit more soak time to allow all the maintainers to reach consensus. It's part my fault for marking the change as ready. I don't think we need revert it, but @cri-o/cri-o-maintainers if there are any issues with this change, please chime in and we can revert. In the future, we should wait the full week :)

Thanks for bringing this point to the notice.

GOVERNANCE.md
```diff
@@ -268,9 +268,12 @@ In the case of abstention, an approver will not be included in the voting approv
...
 
-Most votes require a simple majority of all voting approvers to succeed. Approvers
-can be removed by a 2/3 majority vote of all voting approvers, and changes to this
-Governance require a 2/3 vote of all approvers.
+Most votes require a simple majority of all voting approvers to succeed.
+Approvers can be removed by a 2/3 majority vote of all voting approvers,
+and changes to this Governance require a 2/3 vote of all approvers. The
+approvers will make every reasonable effort to make sure that any removal
+or amendment to the Governance be affirmed with votes by employees of
+more than one organization.
```
fidencio, line 276: I agree with @haircommander that any kind of change like this one should be open for at least a week, giving enough time and ensuring that "The approvers will make every reasonable effort to amke sure that any removal or amendment to the Governance be affirmed with the votes by employees of mnore than one organization".

Regardless of this one being merged or not, I'm 100% in favour of the wording, as it's mostly wordsmith here and the project's always been open (and actually pushing forward to) having a good mix of folks working on different companies.

Thanks, @haircommander, lgtm!

PR #7076, opened 2023-06-21
 Set mount type HostToContainer for mounts that include container storage root

harche:
#### What type of PR is this?

/kind bug

Fixes # https://issues.redhat.com/browse/OCPBUGS-14351

```release-note
 Set mount type HostToContainer for mounts that include container storage root
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -324,7 +324,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	cgroup2RW := node.CgroupIsV2() && sb.Annotations()[crioann.Cgroup2RWAnnotation] == "true"
 
 	s.resourceStore.SetStageForResource(ctx, ctr.Name(), "container volume configuration")
-	containerVolumes, ociMounts, err := addOCIBindMounts(ctx, ctr, mountLabel, s.config.RuntimeConfig.BindMountPrefix, s.config.AbsentMountSourcesToReject, maybeRelabel, skipRelabel, cgroup2RW)
+	containerVolumes, ociMounts, err := addOCIBindMounts(ctx, ctr, mountLabel, s.config.RuntimeConfig.BindMountPrefix, s.config.AbsentMountSourcesToReject, maybeRelabel, skipRelabel, cgroup2RW, s.Config().Root)
 	if err != nil {
 		return nil, err
 	}
@@ -914,7 +914,7 @@ func clearReadOnly(m *rspec.Mount) {
 	m.Options = append(m.Options, "rw")
 }
 
-func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel, bindMountPrefix string, absentMountSourcesToReject []string, maybeRelabel, skipRelabel, cgroup2RW bool) ([]oci.ContainerVolume, []rspec.Mount, error) {
+func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel, bindMountPrefix string, absentMountSourcesToReject []string, maybeRelabel, skipRelabel, cgroup2RW bool, storageRoot string) ([]oci.ContainerVolume, []rspec.Mount, error) {
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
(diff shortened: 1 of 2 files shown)


server/container_create_linux.go
```diff
@@ -969,6 +969,9 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
 		if m.HostPath == "/" && dest == "/" {
 			log.Warnf(ctx, "Configuration specifies mounting host root to the container root.  This is dangerous (especially with privileged containers) and should be avoided.")
 		}
+		if m.HostPath == "/var/lib/containers" {
```
haircommander, line 972: we also should check /var/ and /var/lib (and also /lib)

haircommander: and `/`

server/container_create_linux.go
```diff
@@ -969,6 +969,9 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
...
+			m.Propagation = types.MountPropagation_PROPAGATION_HOST_TO_CONTAINER
```
haircommander, line 973: can we log an info to let the user know what's happening? 

I also wonder if this should be configurable to start... wdyt @cri-o/cri-o-maintainers

saschagrunert: Can we infer `/var/lib/containers` from an existing configuration?

server/container_create_linux.go
```diff
@@ -969,6 +969,16 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
 		if m.HostPath == "/" && dest == "/" {
 			log.Warnf(ctx, "Configuration specifies mounting host root to the container root.  This is dangerous (especially with privileged containers) and should be avoided.")
 		}
+
+		hostPaths := []string{"/var", "/var/lib", "/var/lib/containers", "/lib", "/"}
+
+		for _, path := range hostPaths {
```
haircommander, line 975: another nit: can we name this variable something more informative? `pathsContainingStorageRoot` or something?

server/container_create_linux.go
haircommander: I am inclined to agree with @saschagrunert here. I would not say the priority here is high enough to not have a complete solution.
using https://pkg.go.dev/path/filepath#Rel may be a possibility to calculate whether a path contains the storage root

haircommander: alternatively, maybe
`strings.Contains(storage.Root, m.HostPath)` may give us what we want
https://go.dev/play/p/OU5jiFvElL3

haircommander:
a couple more nits: a test would be sweet as well, and the release note could mention all of the other paths that are configured this way

server/utils.go
```diff
@@ -223,3 +223,28 @@ func (s *Server) FilterDisallowedAnnotations(toFind, toFilter map[string]string,
 
 	return s.config.Workloads.FilterDisallowedAnnotations(allowed, toFilter)
 }
+
+// isSubDirectoryOf checks if the base path contains the target path.
```
haircommander, line 227: I think the behavior is more `containsSubDirectory`

server/utils.go
```diff
@@ -223,3 +223,28 @@ func (s *Server) FilterDisallowedAnnotations(toFind, toFilter map[string]string,
```
haircommander, line 223: nit I'm personally a fan of keeping a function closest to the function that uses it, and reserving utils files for when the function is needed across multiple files. WDYT about moving this and the associated test?

server/utils.go
haircommander: nevermind I misread, the name is correct

server/container_create_linux.go
```diff
@@ -914,7 +914,7 @@ func clearReadOnly(m *rspec.Mount) {
 	m.Options = append(m.Options, "rw")
 }
 
-func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel, bindMountPrefix string, absentMountSourcesToReject []string, maybeRelabel, skipRelabel, cgroup2RW bool) ([]oci.ContainerVolume, []rspec.Mount, error) {
+func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel, bindMountPrefix string, absentMountSourcesToReject []string, maybeRelabel, skipRelabel, cgroup2RW bool, storageRoot string) ([]oci.ContainerVolume, []rspec.Mount, error) {
```
haircommander, line 917: one day I want this to all be refactored, it's really hard to read and figure out what happens to a mount

harche: +1

server/utils.go
harche: Moved from utils.go to container_create_linux.go

server/container_create_linux.go
```diff
@@ -969,6 +969,12 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
 		if m.HostPath == "/" && dest == "/" {
 			log.Warnf(ctx, "Configuration specifies mounting host root to the container root.  This is dangerous (especially with privileged containers) and should be avoided.")
 		}
+
+		if isSubDirectoryOf(storageRoot, m.HostPath) {
+			log.Infof(ctx, "Mount propogration for the host path %s will be set to HostToContainer", m.HostPath)
```
haircommander, line 974: nit: can you include the rational "as it includes the container storage root"

server/container_create_linux.go
```diff
@@ -969,6 +969,12 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
 		if m.HostPath == "/" && dest == "/" {
```
haircommander, line 969: I think we can drop this now

haircommander:
/approve

saschagrunert (approved):

PR #7077, opened 2023-06-22
WIP: Pull image on create to verify signature

saschagrunert:
Alternative to https://github.com/cri-o/cri-o/pull/7046


server/container_create.go
```diff
@@ -297,8 +298,64 @@ func generateUserString(username, imageUser string, uid *types.Int64Value) strin
...
+			}
+			log.Warnf(ctx, "Unable to pull image %s, trying next auth method", image.Image)
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("pull image: %w", err)
+		}
+
+		log.Infof(ctx, "Pulled image ref from user specified image %s: %s", image.Image, pullResp.ImageRef)
+		req.Config.Image.Image = pullResp.ImageRef
```
mtrmac, line 351: What does completely overriding this field do to Kubelet’s model of the runtime?

mtrmac (commented):
This is simpler for CRI-O to do, OTOH it pretty significantly changes the semantics and especially performance and reliability profile of `PullImage` vs. `CreateContainer`; i.e. it moves the complexity of the analysis from CRI-O to Kubelet.

Kubelet does pull retries with backoff; does it do that for CreateContainer at all?

I don’t particularly _love_ the idea of enforcing signatures on already-pulled images; but this is not clearly better.

---

If we are touching the CRI semantics to this extent all, I think it would be conceptually most correct (and maybe less disruptive) for Kubelet to do namespaced pulls, i.e. to issue a PullImage even if the image already exists, providing (credentials, namespace, UserSpecifiedImage). That’s, kind of, equivalent to this PR in effect, but it is less disruptive to code assumptions (if Kubelet is tuned for slow image pulls _that need retrying_ and fast container creations that rarely fail).

But both this PR and the hypothetical namespaced pull would cause registry metadata accesses to happen much more frequently than it is currently the case; so, verifying signatures on already-pulled images locally, as in #7046, might really be the best thing to do. Assuming we somehow fix (or defer?) the encrypted image case.

saschagrunert:
> This is simpler for CRI-O to do, OTOH it pretty significantly changes the semantics and especially performance and reliability profile of `PullImage` vs. `CreateContainer`; i.e. it moves the complexity of the analysis from CRI-O to Kubelet.
> 
> Kubelet does pull retries with backoff; does it do that for CreateContainer at all?

Yes, there is the same backoff for container creation.

https://github.com/kubernetes/kubernetes/blob/9325a57125e8502941d1b0c7379c4bb80a678d5c/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1181-L1218

> I don’t particularly _love_ the idea of enforcing signatures on already-pulled images; but this is not clearly better.
> 
> If we are touching the CRI semantics to this extent all, I think it would be conceptually most correct (and maybe less disruptive) for Kubelet to do namespaced pulls, i.e. to issue a PullImage even if the image already exists, providing (credentials, namespace, UserSpecifiedImage). That’s, kind of, equivalent to this PR in effect, but it is less disruptive to code assumptions (if Kubelet is tuned for slow image pulls _that need retrying_ and fast container creations that rarely fail).

The image pull policies `IfNotPresent` and `Never` would probably still require checking the signatures on container creation, right?

server/container_create.go
saschagrunert: Nothing really, I guess we could also omit this line.

mtrmac:
> The image pull policies `IfNotPresent` and `Never` would probably still require checking the signatures on container creation, right?

`Never` is interesting… does that mean that `root` on each individual node needs to pre-pull images? Or does that “just” mean that one Pod might trigger a pull and another might use `Pull: Never` afterwards? Either seems fairly impractical to me.

If the latter is expected to work, and the two pods can be in different namespaces, yes, we would need to do a signature check not on pull, probably container creation, because that’s the first time the (namespace, `UserSpecifiedImage`) pair shows up.

PR #7094, opened 2023-06-28
Use cache for integration tests

saschagrunert:
#### What type of PR is this?

/kind ci

We use a bunch of golang to install dependencies in the `github-actions-setup` script, means we should use a cache to speedup the CI.

None
None

```release-note
None
```

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -20,6 +20,15 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: ${{ env.GO_VERSION }}
+      - run: sudo chown -R $(id -u):$(id -g) /root
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+            /root/go/pkg/mod
+            /root/.cache/go-build
+          key: go-integration-${{ hashFiles('scripts/versions') }}-${{ hashFiles('scripts/github-actions-setup') }}
       - run: scripts/github-actions-packages
       - run: scripts/github-actions-setup
       - run: make all test-binaries
@@ -27,6 +36,7 @@ jobs:
         env:
           JOBS: '1'
           RUN_CRITEST: '1'
+      - run: sudo chown -R $(id -u):$(id -g) /root
 
   critest-conmonrs:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -20,13 +20,23 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: ${{ env.GO_VERSION }}
+      - run: sudo chown -R $(id -u):$(id -g) /root
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/go/pkg/mod
```
sohankunkerkar, line 27: I think a somewhat similar change has been removed recently in [this](https://github.com/cri-o/cri-o/commit/aa86e94f04c806b26abfafb323a94ab470de66d1) commit.

haircommander:
cc @kolyshkin

.github/workflows/integration.yml
kolyshkin: actions/setup-go@v4 caches a few directories -- `$(go env GOMODCACHE)` and `$(go env GOCACHE)`, to be precise, those are usually `~/go/pkg/mod` and `~/.cache/go-build`.

I think we should build all sources and tests as non-root (and I don't see any cases where we run `go build` or `go test` as root, maybe I'm mistaken?), so the above caching should be sufficient.

.github/workflows/integration.yml
```diff
@@ -20,13 +20,23 @@ jobs:
...
+            ~/.cache/go-build
+            /root/go/pkg/mod
+            /root/.cache/go-build
+          key: go-integration-${{ hashFiles('scripts/versions') }}-${{ hashFiles('scripts/github-actions-setup') }}
```
kolyshkin, line 31: and, if you want hashes of these files to be used as cache key input, there's a `cache-dependency-path` parameter described in actions/setup-go README.md (if unset, it looks for `go.sum` file in `$GITHUB_WORKSPACE`)

Something like this should work:

```yml
    cache-dependency-path: |
        ./go.sum
        ./scripts-versions
        ./scripts-github-actions-setup
```

OTOH, I think that changing a version of a tool we download should not invalidate go cache, so this looks like not being necessary.

saschagrunert:
I think we can solve this by changing the script.

PR #7099, opened 2023-06-30
container_test: fix "AddCapabilities ALL" test case

### LINKED ISSUE #7071: Trying to run unit tests locally fails
### Which jobs are failing?

When I run `make testunit`, it throws an error.

### Which tests are failing?

```
[1687310869] Container - 78/78 specs ••••••••••••••••••••••••••••••••••
• [FAILED] [0.021 seconds]
cri-o: Container cri-o: SpecSetupCapabilities [It] AddCapabilities ALL DropCapabilities one should drop that one
/root/cri-o/internal/factory/container/container_test.go:577

  [FAILED] Expected
      <int>: 37
  to equal
      <int>: 40
  In [It] at: /root/cri-o/internal/factory/container/container_test.go:544 @ 06/21/23 01:34:11.711

  Full Stack Trace
    github.com/cri-o/cri-o/internal/factory/container_test.glob..func5.13.1(0x4000481980, 0x1be01d0?)
        /root/cri-o/internal/factory/container/container_test.go:544 +0xc4
    github.com/cri-o/cri-o/internal/factory/container_test.glob..func5.13.5()
        /root/cri-o/internal/factory/container/container_test.go:585 +0x198
•••••••••••••••••••••••••••••••••••••••••••

Summarizing 1 Failure:
  [FAIL] cri-o: Container cri-o: SpecSetupCapabilities [It] AddCapabilities ALL DropCapabilities one should drop that one
  /root/cri-o/internal/factory/container/container_test.go:544

Ran 78 of 78 Specs in 0.102 seconds
FAIL! -- 77 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestContainer (0.16s)
FAIL
```

### Since when has it been failing?

In my environment, the tests keep failing with assertion errors.

### Testgrid link

_No response_

### Reason for failure (if possible)

I believe line 585 should be modified as follows: 
verifyCapValues(sut.Spec().Config.Process.Capabilities, **_int(validate.LastCap())+1_**-1)

https://github.com/cri-o/cri-o/blob/da46d0f2b50495d6107b7a8e1d90e619cc9306c1/internal/factory/container/container_test.go#L577-L586

The final value of `AddCapabilities: []string{"ALL"}` is different in different environments. Our actual value range is `[0,validate.LastCap()]` instead of `[0,capability.List()]`. As follows:
https://github.com/cri-o/cri-o/blob/da46d0f2b50495d6107b7a8e1d90e619cc9306c1/internal/factory/container/container.go#L747-L756

The difference lies in the fact that `validate.LastCap()` is determined by the operating environment, while `len(capability.List())` is a fixed value and can lead to failures in different environments.

In my environment the value of `validate.LastCap()` is 37.
```
# cat /proc/sys/kernel/cap_last_cap
37
```

whygyc:
"AddCapabilities ALL" means different values in different environments, and our base value should be last rather than a fixed list value.

#### What type of PR is this?
/kind failing-test

Fixes #7071 

```release-note
None
```

### DIFF
--- a/internal/factory/container/container_test.go
+++ b/internal/factory/container/container_test.go
@@ -19,7 +19,7 @@ import (
 	. "github.com/onsi/gomega"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
-	"github.com/syndtr/gocapability/capability"
+	validate "github.com/opencontainers/runtime-tools/validate/capabilities"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 	kubeletTypes "k8s.io/kubernetes/pkg/kubelet/types"
 )
@@ -582,7 +582,9 @@ var _ = t.Describe("Container", func() {
 			serverCaps := []string{}
 
 			Expect(sut.SpecSetupCapabilities(caps, serverCaps, false)).To(BeNil())
-			verifyCapValues(sut.Spec().Config.Process.Capabilities, len(capability.List())-1)
+			// `int(validate.LastCap())+1` represents the total number of `ALL` capabilities
+			// in the current environment, while `-1` indicates the removal of `CHOWN` from `ALL`.
+			verifyCapValues(sut.Spec().Config.Process.Capabilities, int(validate.LastCap())+1-1)
 		})
 		It("AddCapabilities one DropCapabilities ALL should add that one", func() {
 			caps := &types.Capability{


haircommander:
/ok-to-test
/approve

LGTM, thank you @whygyc 
PTAL @cri-o/cri-o-maintainers

internal/factory/container/container_test.go
```diff
@@ -582,7 +582,7 @@ var _ = t.Describe("Container", func() {
 			serverCaps := []string{}
 
 			Expect(sut.SpecSetupCapabilities(caps, serverCaps, false)).To(BeNil())
-			verifyCapValues(sut.Spec().Config.Process.Capabilities, len(capability.List())-1)
+			verifyCapValues(sut.Spec().Config.Process.Capabilities, int(validate.LastCap())+1-1)
```
littlejawa, line 585: I don't get the "+1-1" statement. Isn't it a no-op ?

whygyc: > I don't get the "+1-1" statement. Isn't it a no-op ?

From the final result, yes, it can be simply written as `int(validate.LastCap())`. 
This is mainly to avoid confusion because `int(validate.LastCap())+1` represents the number of `ALL` capabilities in the current environment, and `-1` indicates removing `CHOWN` from `ALL`.
Without the `-1`, the idea of excluding `CHOWN` from `ALL` in the test case cannot be reflected.

littlejawa: Got it. I think a comment would be clearer then, but that's a nit as the Add/Remove capabilities is just a couple lines above.
Thank you!

haircommander: yeah  I agree would you mind adding a comment justifying it @whygyc ?

whygyc: Sure, I agree. I will add a comment as you suggested.

sohankunkerkar (approved):
/lgtm

PR #7119, opened 2023-07-05
cgmgr: set CPUAffinity when AllowedCPUs are set

haircommander:
#### What type of PR is this?

/kind bug
if we don't do this, newer kernels won't listen to the cpuset (as it will take the subset of each)

```release-note
Fix a bug where conmon had AllowedCPUs set, but not CPUAffinity, causing the CPUs used to be incorrect
```


haircommander:
there was a different workaround chosen instead

PR #7129, opened 2023-07-12
oci: simplify stopping code

haircommander:
#### What type of PR is this?

/kind bug

Before, there was the possibility load could cause cri-o to segfault from double closing of channels.

this PR aims to simplify container stop code while retaining the required behavior.

Now, the first stop begins a registration process where the container stop begins and new timeouts
come in to interrupt. There are two communication channels, and only one location where they can be closed.
This also adds a watcher mechanism so callers can wait on the container stop

This PR also includes some cleanups:
- replace IsAlive() with Living()
- move around where ShouldBeStopped() is called

```release-note
Fix a very rare panic from a double closed channel in container stop
```


internal/oci/runtime_oci.go
```diff
@@ -784,151 +784,113 @@ func (r *runtimeOCI) UpdateContainer(ctx context.Context, c *Container, res *rsp
...
+	if err := c.ShouldBeStopped(); err != nil {
+		if err == ErrContainerStopped {
+			err = nil
+		}
+		return err
+	}
+
+	// The initial container process either doesn't exist, or isn't ours.
+	if err := c.IsAlive(); err != nil {
+		c.state.Finished = time.Now()
```
kolyshkin, line 807: Do we need to set `c.state.Status` here?

kolyshkin:
Left some nits. Note I did not review tests changes. Overall LGTM.

I think the last commit needs to be squashed into the previous one.

internal/oci/runtime_oci.go
haircommander: the state machine is kind of weird. TBH I don't think we should set either ever, and rely on UpdateContainerStatus to set them, but this is here for now.

kolyshkin (approved):
lgtm

internal/oci/runtime_oci.go
```diff
@@ -784,148 +784,113 @@ func (r *runtimeOCI) UpdateContainer(ctx context.Context, c *Container, res *rsp
...
+	defer span.End()
+
+	if c.Spoofed() {
+		c.state.Status = ContainerStateStopped
+		c.state.Finished = time.Now()
+		return nil
+	}
+
+	if err := c.ShouldBeStopped(); err != nil {
+		if err == ErrContainerStopped {
```
saschagrunert, line 799: Any reason to not use `errors.Is` here?

internal/oci/runtime_vm.go
```diff
@@ -521,14 +521,17 @@ func (r *runtimeVM) StopContainer(ctx context.Context, c *Container, timeout int
 	log.Debugf(ctx, "RuntimeVM.StopContainer() start")
 	defer log.Debugf(ctx, "RuntimeVM.StopContainer() end")
 
-	// Lock the container
-	c.opLock.Lock()
-	defer c.opLock.Unlock()
-
 	if err := c.ShouldBeStopped(); err != nil {
+		if err == ErrContainerStopped {
```
saschagrunert, line 525: Same here, let's use `errors.Is` if possible.

server/container_remove.go
```diff
@@ -53,7 +53,7 @@ func (s *Server) removeContainerInPod(ctx context.Context, sb *sandbox.Sandbox,
 	defer span.End()
 	if !sb.Stopped() {
 		if err := s.stopContainer(ctx, c, int64(10)); err != nil {
-			return fmt.Errorf("failed to stop container for removal")
+			return fmt.Errorf("failed to stop container for removal %v", err)
```
saschagrunert, line 56:
```suggestion
			return fmt.Errorf("failed to stop container for removal %w", err)
```

internal/oci/runtime_oci.go
```diff
@@ -784,148 +784,113 @@ func (r *runtimeOCI) UpdateContainer(ctx context.Context, c *Container, res *rsp
...
+	}
+
+	// The initial container process either doesn't exist, or isn't ours.
+	if err := c.Living(); err != nil {
+		c.state.Finished = time.Now()
+		return nil
+	}
+
+	if c.SetAsStopping() {
+		go r.StopLoopForContainer(context.Background(), c)
```
saschagrunert, line 812: No need to pass a context here if we always use `context.Background()`.

saschagrunert (changes requested):
Do we want to write an integration test for this, which triggers stop multiple times and ensures that everything is sane afterwards?

saschagrunert (approved):

haircommander:
we will be backporting it. I am not sure 1.25 will get it at this point but at least 1.26 I would guess

PR #6925, opened 2023-05-16
server: do not take lock to populate verbose information for containe…

haircommander:
#### What type of PR is this?

/kind bug
#### What this PR does / why we need it:
…r status

We can do this best-effort, and don't need to take the lock to do so. Also, by using the InitPid, we're not prone to reading a variable that changes underneath us, as the init pid is set once and not changed.

#### Does this PR introduce a user-facing change?

```release-note
Fix a bug where sending information to cadvisor is stalled on taking the container's state lock
```

PR #7131, opened 2023-07-12
[1.26] server: do not take lock to populate pid in container status and inspect

sohankunkerkar:
This is a manual cherry-pick of #6925

/assign haircommander

```release-note
Fix a bug where sending information to cadvisor is stalled on taking the container's state lock
```

### DIFF
--- a/server/container_status.go
+++ b/server/container_status.go
@@ -128,7 +128,7 @@ func (s *Server) createContainerInfo(container *oci.Container) (map[string]strin
 		Privileged  bool      `json:"privileged"`
 	}{
 		container.Sandbox(),
-		container.State().Pid,
+		container.StateNoLock().Pid,
 		container.Spec(),
 		metadata.Privileged,
 	}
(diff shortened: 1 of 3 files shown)


internal/oci/container.go
```diff
@@ -296,6 +298,16 @@ func (c *Container) CreatedAt() time.Time {
...
+// CheckpointedAt returns the container checkpoint time
+func (c *Container) CheckpointedAt() time.Time {
+	return c.state.CheckpointedAt
+}
+
+// SetCheckpointedAt sets the time of checkpointing
+func (c *Container) SetCheckpointedAt(checkpointedAt time.Time) {
+	c.state.CheckpointedAt = checkpointedAt
+}
+
```
haircommander, line 310: I don't think we should take the CheckpointedAt pieces

sohankunkerkar: Fixed

PR #7142, opened 2023-07-18
Fix filemode for device

littlejawa:
#### What type of PR is this?

/kind bug

When running a container with the kata runtime and setting up a block device, the device appears in the container with permissions set to 0.
The kata agent, creating this device in the VM, is applying the file mode provided in the spec file by cri-o - this field is not filled by cri-o, and defaults to 0.

This PR makes sure that the device's permission is provided in the spec, so that the agent can apply them appropriately.

None

```release-note
Fixed an issue preventing the use of block devices with kata containers
```


saschagrunert (approved):
Thank you!

Issue #7143, opened 2023-07-19
imageRef in container status is not image ID and lead to kubelet image gc work abnormally


ytinirt:
### What happened?

The imageRef of containers, returned by runtime cri-o, may not be image ID. This will mislead kubelet's image-gc logical.

In [CRI spec](https://github.com/kubernetes/kubernetes/blob/90c362b3430bcbbf8f245fadbcd521dab39f1d7c/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go#L5544), it consider imageRef as image ID:
```
// Reference to the image in use. For most runtimes, this should be an
// image ID.
ImageRef string `protobuf:"bytes,5,opt,name=image_ref,json=imageRef,proto3" json:"image_ref,omitempty"`
```

When do image gc logical, kubelet will get image and container status from cri-o runtime service. And then, compare currently used images (by container) and existing images by image ID.

If imageRef return by cri-o is not image ID, kubelet image gc can not work properly.

Kubelet's key logs may look like this:
```
I0719 16:36:14.498634 xxx image_gc_manager.go:233] "Container uses image" pod="multus/multus-admission-controller-7ff9c677b7-2p6lx" containerName="multus-admission-controller" containerImage="38da782fd6a202da62c652532b5d289278819d84bf2f7280ad40d214903db6e3" imageID="image.foo.bar/multus-admission-controller@sha256:36f44236d5c1aa57bd5e930179f020c089e2bde0d20a5d349f5af8fefe2d6235"
I0719 16:36:14.500180 xxx image_gc_manager.go:244] "Adding image ID to currentImages" imageID="38da782fd6a202da62c652532b5d289278819d84bf2f7280ad40d214903db6e3"
I0719 16:36:14.500186 xxx image_gc_manager.go:261] "Image ID has size" imageID="38da782fd6a202da62c652532b5d289278819d84bf2f7280ad40d214903db6e3" size=316580906
I0719 16:36:14.500192 xxx image_gc_manager.go:264] "Image ID is pinned" imageID="38da782fd6a202da62c652532b5d289278819d84bf2f7280ad40d214903db6e3" pinned=false
I0719 16:36:14.534022 xxx image_gc_manager.go:371] "Evaluating image ID for possible garbage collection" imageID="38da782fd6a202da62c652532b5d289278819d84bf2f7280ad40d214903db6e3"
[...]

PR #1116, opened 2017-11-03
Update containers/image and containers/storage

nalind:
Update to currently-proposed changes for containers/image.  This is mainly a placeholder so that we can be extra-sure that those changes won't negatively impact cri-o.

PR #7149, opened 2023-07-20
Fix ImageRef field for containers to default to an image ID

saschagrunert:
#### What type of PR is this?

/kind bug

When restoring containers, then we always restore an image ID based on the annotation. When creating containers, then we use the first digest with a higher priority than the image ID. We now streamline this behavior by always using an image ID.

Fixes https://github.com/cri-o/cri-o/issues/7143
Refers to https://github.com/kubernetes-incubator/cri-o/issues/531
cc @QiWang19 @mtrmac @nalind 

The change got introduced in https://github.com/cri-o/cri-o/pull/1116/commits/f3b7065bd80a53e4c48e8c45c36dcb86717e1578#diff-1ec6bb9cf833fd4ded46ce1feb6440f1d7532294e0445623f2f2e0bb2726ea63L976-R975

```release-note
Fixed bug where `ImageRef` refers to a digest rather than an image ID, which negatively impacted kubelets garbage collection.
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -237,9 +237,6 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 
 	imageName := imgResult.Name
 	imageRef := imgResult.ID
-	if len(imgResult.RepoDigests) > 0 {
-		imageRef = imgResult.RepoDigests[0]
-	}
 
 	labelOptions, err := ctr.SelinuxLabel(sb.ProcessLabel())
 	if err != nil {
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -211,7 +211,7 @@ function check_images() {
     eval "$(jq -r '.images[] |
         select(.repoTags[0] == "quay.io/crio/fedora-crio-ci:latest") |
         "REDIS_IMAGEID=" + .id + "\n" +
-	"REDIS_IMAGEREF=" + .repoDigests[0]' <<<"$json")"
+	"REDIS_IMAGEREF=" + .id' <<<"$json")"
 }
 
 function start_crio_no_setup() {
--- a/test/image.bats
+++ b/test/image.bats
@@ -79,7 +79,6 @@ function teardown() {
 
 	output=$(crictl inspect -o yaml "$ctr_id")
 	[[ "$output" == *"image: $IMAGE_LIST_DIGEST"* ]]
-	[[ "$output" == *"imageRef: $IMAGE_LIST_DIGEST"* ]]
 }
 
 @test "image pull and list" {


lance5890:
+1

haircommander:
this is related to https://issues.redhat.com/browse/OCPBUGS-16240 as well, right?

mtrmac:
#7143 does look like a compelling reason for why the field should always be an image ID.

I don’t have a good understanding of the history and motivations that lead to using the digested value in `ImageRef` when available.

saschagrunert:
> this is related to https://issues.redhat.com/browse/OCPBUGS-16240 as well, right?

Yes I think this may fix the issue, too.

haircommander:
chatted with @nalind in community meeting, let's move forward with this

/lgtm

lcarva:
With this change, how can someone tell the digest of the image used in a Pod's container from its status?

saschagrunert:
> With this change, how can someone tell the digest of the image used in a Pod's container from its status?

Not if it got pulled by tag or referenced by ID. If the image got pulled by digest then it's still available in the `.status.image.image`.

lcarva:
> > With this change, how can someone tell the digest of the image used in a Pod's container from its status?
> 
> Not if it got pulled by tag or referenced by ID. If the image got pulled by digest then it's still available in the `.status.image.image`.

Using a tag when creating a pod is useful in some cases. Not being able to tell which exact image was used in a Pod makes auditing extremely difficult, if not impossible. From a security perspective, this is concerning.

saschagrunert:
> Not being able to tell which exact image was used in a Pod makes auditing extremely difficult, if not impossible. From a security perspective, this is concerning.

A tag can refer to multiple repo digests, means we cannot know which one is the _right_ one. 

We could add annotations to the pods referencing the repo digests, though.

mtrmac:
“The right” repo digest can only be determined if the Kubelet issues a `PullImage`. 

If the image is present locally, and Kubelet’s `ImageStatus` with the tag finds it, the “raw” data is only an image ID with a _list_ of possible digests.

in the above case:

This only _really_ matters in the fairly unlikely case that the same tag has, over time, pointed at two different digests which nevertheless map to the same uncompressed image (same config digest). And in that case I don’t know why anyone inspecting the pod should care which one is reported.

There’s also the less unlikely case where different _repos_ point at the same uncompressed image; in that case we should, _in principle_, be able to provide a relevant digest from the right repo — but that doesn’t currently work either; we don’t store enough data for that. That could be fixed without too much effort (but it requires changes throughout the stack, at least in c/image and Podman as well).

mtrmac:
AFAICS https://github.com/openshift/kubernetes/blob/deb2c6012264c4b300833165e51134de25441985/pkg/kubelet/images/image_manager.go#L138 , via https://github.com/openshift/kubernetes/blob/deb2c6012264c4b300833165e51134de25441985/pkg/kubelet/kuberuntime/kuberuntime_image.go#L86C36-L86C36 , uses https://github.com/cri-o/cri-o/blob/44cb5088a1b4a6107b6803109e08a02609bec9b6/server/image_status.go#L68 ; and that is used to create a container.

So the digest is never provided as a part of container data when creating a container on that path.

lcarva:
@mtrmac, digging through that, how do I trigger the behavior with an older version of cri-o? I tried different variations of `ImagePullPolicy` and whether the image was present on the cluster or not. In every case, I get a digest in `ImageID`.

mtrmac:
@lcarva OK then I’m afraid I have no idea what you mean by `ImageID`. The only field with that name in CRI-O is https://github.com/cri-o/cri-o/blob/44cb5088a1b4a6107b6803109e08a02609bec9b6/internal/storage/runtime.go#L124 , and that is never filled with a digest.

lcarva:
@mtrmac, I mean `ContainerStatus.ImageID`:

```
$ oc get pod img-test-pod-c4m55 -o json | jq '.status.containerStatuses'
[
  {
    "containerID": "cri-o://f0296139333367a75f962b0710457f98031454b2f1e0cf0d82c14a5388c2490b",
    "image": "registry.redhat.io/ubi8:latest",
    "imageID": "registry.redhat.io/ubi8@sha256:b7e686e30346e9ace664fa09c0275262f8b9a443ed56d22165a0e201f6488c13",
    "lastState": {},
    ...
  }
]
```

mtrmac:
@lcarva If this is about the _Kubernetes_ `ContainerStatus.ImageID`, the CRI-O equivalent `ContainerStatus.ImageRef` (affected by this PR) contains _some_ repo@digest value for that image, but that might not correspond to the users’ request.

If the user asks for `redhat.io/ubi@digest1`, the repo digest value before this PR could easily have been `mirror.local/ubi-recompressed@digest2`.

lcarva:
> If the user asks for `redhat.io/ubi@digest1`, the repo digest value before this PR could easily have been `mirror.local/ubi-recompressed@digest2`.

Yes, exactly! I had filed [OCPBUGS-8428](https://issues.redhat.com/browse/OCPBUGS-8428) which is how I got redirected to this PR.

To be clear, the previous behavior was certainly broken (as you correctly call out), but at least the digest was there. A consumer could get the right image ref by combining the repo from `Image` and the digest from `ImageID`, for example. The new behavior is significantly inferior in this area as the digest is not recorded anywhere.

mtrmac:
> A consumer could get the right image ref by combining the repo from Image and the digest from ImageID, for example. 

To cross-link things, no. https://github.com/tektoncd/pipeline/issues/6970#issuecomment-1673727730

Issue #6492, opened 2023-01-05
Porting cri-o to FreeBSD


dfr:
This is a tracking issue for the FreeBSD porting process to facilitate communications and visibility.

haircommander:
activity has stalled here, so I'm removing it from the 1.27 milestone

dfr:
Sorry about that - too much multitasking over the last few months pushed this back a bit. I should be able to get back to it soon.

haircommander:
no problem! 1.27 was a ambitious milestone to aim for anyway :)

PR #7153, opened 2023-07-20
Allow internal/config packages to build on non-linux

dfr:
This simply moves linux specific files to have a linux-only suffix and adds non-functional stubs for other platforms. There should be no functional changes.

#### What type of PR is this?

/kind cleanup

This is part of a larger project to allow cri-o to work on FreeBSD (#6492)

None

```release-note
None
```


internal/config/nsmgr/types_unsupported.go
```diff
@@ -0,0 +1,28 @@
+//go:build !linux
+// +build !linux
```
haircommander, line 2: is there a way we can avoid this file? I see a lot of the same content as types_linux, so maybe we can take the types.go content and put any platform specific pieces in nsmgr_*

internal/config/node/systemd_unsupported.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+var (
+	systemdHasCollectModeOnce sync.Once
+	systemdHasCollectMode     bool
+	systemdHasCollectModeErr  error
+
+	systemdHasAllowedCPUsOnce sync.Once
+	systemdHasAllowedCPUs     bool
+	systemdHasAllowedCPUsErr  error
```
haircommander, line 17: I don't see a need for these if they're internal

internal/config/nsmgr/types_unsupported.go
dfr: In this PR, I was aiming to blindly rename things to make it easier to review. I can add a separate commit to this PR which factor out the common parts though.

haircommander: I am fine with a slightly harder review--I think I'd prefer it squashed with the existing commits (but maintaining the different commits per directory)

dfr: Looking at the freebsd version of types, the sharable parts are mainly the bits that I moved to types_unsupported.go so I'll re-work this commit to leave those bits in types.go.

internal/config/node/systemd_unsupported.go
dfr: Removed

dfr:
I also added a commit which fixes the non-cgo linux cross build from freebsd which I use for sanity checking.

internal/config/nsmgr/nsmgr_unsupported.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"runtime"
+)
+
+// NamespaceManager manages the server's namespaces.
+// Specifically, it is an interface for how the server is creating namespaces,
+// and can be requested to create namespaces for a pod.
+type NamespaceManager struct {
+	namespacesDir string
+	pinnsPath     string
```
haircommander, line 16: nit: these can be dropped

internal/config/nsmgr/nsmgr_unsupported.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// New creates a new NamespaceManager.
+func New(namespacesDir, pinnsPath string) *NamespaceManager {
+	return &NamespaceManager{
+		namespacesDir: namespacesDir,
+		pinnsPath:     pinnsPath,
+	}
+}
```
haircommander, line 25: and here, `_, _ string`

dfr:
The ci/prow/e2e-gcp-ovn test is consistently failing but I'm not sure how to debug it. Any ideas?

haircommander:
no need :)

/override ci/prow/e2e-gcp-ovn

dfr:
Rebased

saschagrunert (changes requested):
Do we want to add a CI test for this? It's likely that we break it at some point if not.

dfr:
> Do we want to add a CI test for this? It's likely that we break it at some point if not.

I have a few follow-up PRs which add enough stubs to support building a non-functioning crio on FreeBSD and also as a cross build target on Linux. When we get that far, I think it would be useful to add a CI smoke test but not just yet.

dfr:
Rebased.

Is there anything else needed for this one? I have a followup stack of about 10 commits which takes things to the point where I can build a non-functional crio binary on freebsd. This will make it possible to add a freebsd cross build to CI.

haircommander:
/approve
/lgtm

thanks!

PR #7036, opened 2023-06-09
test/metrics: simplify OOM test

kolyshkin:
#### What type of PR is this?

/kind ci

#### What this PR does / why we need it:

Currently, this test case keeps waiting for the container that was
already killed in case OOM detection has failed. This is a huge (15
minutes) waste of time.

To fix, change the logic to wait for container exit, and then check the
OOM. This also makes the test code more readable.

Also, add some debug info in case the test has failed.

Also, add a 1 second sleep to the container command, which apparently
improves the chances for the test to succeed.

#### Which issue(s) this PR fixes:

None

Related to: https://github.com/cri-o/cri-o/issues/7035

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

PR #7167, opened 2023-07-25
Deflake oom killer integration test

saschagrunert:
#### What type of PR is this?

/kind flake

The test regularly fails.

None
None

```release-note
None
```

### DIFF
--- a/test/metrics.bats
+++ b/test/metrics.bats
@@ -97,7 +97,7 @@ function teardown() {
 	CONTAINER_ENABLE_METRICS=true CONTAINER_METRICS_PORT=$PORT start_crio
 
 	jq '.linux.resources.memory_limit_in_bytes = 15728640
-        | .command = ["sh", "-c", "dd if=/dev/zero of=/dev/null bs=20M"]' \
+        | .command = ["sh", "-c", "sleep 5 && dd if=/dev/zero of=/dev/null bs=20M && sleep 30"]' \
 		"$TESTDATA/container_config.json" > "$TESTDIR/config.json"
 	CTR_ID=$(crictl run "$TESTDIR/config.json" "$TESTDATA/sandbox_config.json")
 


saschagrunert:
@cri-o/cri-o-maintainers PTAL, this refers to https://github.com/kubernetes/kubernetes/pull/116082 and https://github.com/containerd/containerd/issues/8180 and should make the CI a bit more robust.

haircommander:
nice find! LGTM, @kolyshkin WDYT

sohankunkerkar:
looks like the test still flakes: https://github.com/cri-o/cri-o/actions/runs/5657093961/job/15325427752?pr=7167

haircommander:
there's some discussion here https://github.com/cri-o/cri-o/pull/7036 but it seems we need to update CRI-O to drop setting `inactive-or-failed` property when runc is version 1.8 or newer

test/metrics.bats
```diff
@@ -97,7 +97,7 @@ function teardown() {
 	CONTAINER_ENABLE_METRICS=true CONTAINER_METRICS_PORT=$PORT start_crio
 
 	jq '.linux.resources.memory_limit_in_bytes = 15728640
-        | .command = ["sh", "-c", "dd if=/dev/zero of=/dev/null bs=20M"]' \
+        | .command = ["sh", "-c", "sleep 5 && dd if=/dev/zero of=/dev/null bs=20M && sleep 30"]' \
```
kolyshkin, line 100: The second sleep makes no sense since `dd` is running until killed.

The first sleep is added by #7036 (although I'm using `sleep 1` in there. Can we please merge #7036?

The problem with the flake is, sometimes neither conmon nor systemd detect an OOM event for cgroup v2. I am not entirely sure why.

saschagrunert: Ah I did not see #7036, sorry about that.

PR #7179, opened 2023-07-27
LinkLogs: give everyone access to log files

haircommander:
#### What type of PR is this?

/kind bug
or else any non root uid can't access the logs

```release-note
give 600 to the linked logs directory so non-root IDs can access it
```


haircommander:
turns out we need to change a long standing conmon behavior to get this, which I don't think we should do

PR #7180, opened 2023-07-28
*: add platform_runtime_paths to RuntimeHandler

sohankunkerkar:
This helps to track the runtime paths for different platforms.

#### What type of PR is this?

/kind feature

```release-note
add platform_runtime_paths option to the runtime handler structure, which allows admins to define specific runtime paths based on different platforms.
```

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -331,6 +331,9 @@ The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  Th
   "io.kubernetes.cri-o.UnifiedCgroup.$CTR_NAME" for configuring the cgroup v2 unified block for a container.
   "io.containers.trace-syscall" for tracing syscalls via the OCI seccomp BPF hook.
 
+**platform_runtime_paths**={}
+  A mapping of platforms to the corresponding runtime executable paths for the runtime handler.
+
 ### CRIO.RUNTIME.WORKLOADS TABLE
 The "crio.runtime.workloads" table defines a list of workloads - a way to customize the behavior of a pod and container.
 A workload is chosen for a pod based on whether the workload's **activation_annotation** is an annotation on the pod.
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -439,6 +439,11 @@ func (c *ContainerServer) LoadContainer(ctx context.Context, id string) (retErr 
 		imgRef = ""
 	}
 
+	platformRuntimePath, ok := m.Annotations[crioann.PlatformRuntimePath]
+	if !ok {
+		platformRuntimePath = ""
+	}
+
 	kubeAnnotations := make(map[string]string)
 	if err := json.Unmarshal([]byte(m.Annotations[annotations.Annotations]), &kubeAnnotations); err != nil {
 		return err
@@ -468,6 +473,8 @@ func (c *ContainerServer) LoadContainer(ctx context.Context, id string) (retErr 
 		return fmt.Errorf("failed to write container state to disk %q: %w", ctr.ID(), err)
 	}
 	ctr.SetCreated()
+
+	ctr.SetRuntimePathForPlatform(platformRuntimePath)
 
 	c.AddContainer(ctx, ctr)
 
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -79,6 +79,7 @@ type Container struct {
 	restoreArchive     string
 	restoreIsOCIImage  bool
 	resources          *types.ContainerResources
+	runtimePath        string // runtime path for a given platform
 }
 
 func (c *Container) CRIAttributes() *types.ContainerAttributes {
@@ -754,4 +755,17 @@ func (c *Container) SetResources(s *specs.Spec) {
 // GetResources returns a copy of the Linux resources from Container
 func (c *Container) GetResources() *types.ContainerResources {
 	return c.resources
+}
+
+// SetRuntimePathForPlatform sets the runtime path for a given platform.
+func (c *Container) SetRuntimePathForPlatform(runtimePath string) {
+	c.runtimePath = runtimePath
+}
+
+// RuntimePathForPlatform returns the runtime path for a given platform.
+func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
+	if c.runtimePath == "" {
+		return r.handler.RuntimePath
+	}
+	return c.runtimePath
 }
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -112,7 +112,7 @@ func (r *runtimeOCI) CreateContainer(ctx context.Context, c *Container, cgroupPa
 		"-P", c.conmonPidFilePath(),
 		"-p", filepath.Join(c.bundlePath, "pidfile"),
 		"--persist-dir", c.dir,
-		"-r", r.handler.RuntimePath,
+		"-r", c.RuntimePathForPlatform(r),
 		"--runtime-arg", fmt.Sprintf("%s=%s", rootFlag, r.root),
 		"--socket-dir-path", r.config.ContainerAttachSocketDir,
 		"--syslog",
@@ -409,7 +409,7 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
 
 	args := r.defaultRuntimeArgs()
 	args = append(args, "exec", "--process", processFile, c.ID())
-	execCmd := cmdrunner.Command(r.handler.RuntimePath, args...) // nolint: gosec
+	execCmd := cmdrunner.Command(c.RuntimePathForPlatform(r), args...) // nolint: gosec
 	if v, found := os.LookupEnv("XDG_RUNTIME_DIR"); found {
 		execCmd.Env = append(execCmd.Env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", v))
 	}
@@ -525,7 +525,7 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 	args := []string{
 		"-c", c.ID(),
 		"-n", c.name,
-		"-r", r.handler.RuntimePath,
+		"-r", c.RuntimePathForPlatform(r),
 		"-p", pidFile,
 		"-e",
 		"-l", logPath,
@@ -764,7 +764,7 @@ func (r *runtimeOCI) UpdateContainer(ctx context.Context, c *Container, res *rsp
 		return nil
 	}
 
-	cmd := cmdrunner.Command(r.handler.RuntimePath, rootFlag, r.root, "update", "--resources", "-", c.ID()) // nolint: gosec
+	cmd := cmdrunner.Command(c.RuntimePathForPlatform(r), rootFlag, r.root, "update", "--resources", "-", c.ID()) // nolint: gosec
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
 	cmd.Stdout = &stdout
@@ -1439,8 +1439,8 @@ func (r *runtimeOCI) defaultRuntimeArgs() []string {
 func (r *runtimeOCI) CheckpointContainer(ctx context.Context, c *Container, specgen *rspec.Spec, leaveRunning bool) error {
 	c.opLock.Lock()
 	defer c.opLock.Unlock()
-
-	if err := r.checkpointRestoreSupported(); err != nil {
+	runtimePath := c.RuntimePathForPlatform(r)
+	if err := r.checkpointRestoreSupported(runtimePath); err != nil {
 		return err
 	}
 
@@ -1481,7 +1481,7 @@ func (r *runtimeOCI) CheckpointContainer(ctx context.Context, c *Container, spec
 
 	_, err := r.runtimeCmd(args...)
 	if err != nil {
-		return fmt.Errorf("running %q %q failed: %w", r.handler.RuntimePath, args, err)
+		return fmt.Errorf("running %q %q failed: %w", runtimePath, args, err)
 	}
 
 	c.SetCheckpointedAt(time.Now())
@@ -1496,7 +1496,7 @@ func (r *runtimeOCI) CheckpointContainer(ctx context.Context, c *Container, spec
 
 // RestoreContainer restores a container.
 func (r *runtimeOCI) RestoreContainer(ctx context.Context, c *Container, cgroupParent, mountLabel string) error {
-	if err := r.checkpointRestoreSupported(); err != nil {
+	if err := r.checkpointRestoreSupported(c.RuntimePathForPlatform(r)); err != nil {
 		return err
 	}
 
@@ -1560,11 +1560,11 @@ func (r *runtimeOCI) RestoreContainer(ctx context.Context, c *Container, cgroupP
 	return nil
 }
 
-func (r *runtimeOCI) checkpointRestoreSupported() error {
+func (r *runtimeOCI) checkpointRestoreSupported(runtimePath string) error {
 	if err := criu.CheckForCriu(criu.PodCriuVersion); err != nil {
 		return fmt.Errorf("check for CRIU %w", err)
 	}
-	if !crutils.CRRuntimeSupportsCheckpointRestore(r.handler.RuntimePath) {
+	if !crutils.CRRuntimeSupportsCheckpointRestore(runtimePath) {
 		return fmt.Errorf("configured runtime does not support checkpoint/restore")
 	}
 	return nil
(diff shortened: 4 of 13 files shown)


internal/oci/runtime_oci.go
```diff
@@ -112,13 +113,22 @@ func (r *runtimeOCI) CreateContainer(ctx context.Context, c *Container, cgroupPa
...
 		"-p", filepath.Join(c.bundlePath, "pidfile"),
 		"--persist-dir", c.dir,
-		"-r", r.handler.RuntimePath,
 		"--runtime-arg", fmt.Sprintf("%s=%s", rootFlag, r.root),
 		"--socket-dir-path", r.config.ContainerAttachSocketDir,
 		"--syslog",
 		"-u", c.ID(),
 	}
 
+	imageResult, err := getImageDetailsForContainer(r.Runtime, ctx, c)
```
kolyshkin, line 122: This might be somewhat expensive, especially given the fact it is needed for wasm only, but the code is executed every time.

internal/oci/runtime_oci.go
```diff
@@ -112,13 +113,22 @@ func (r *runtimeOCI) CreateContainer(ctx context.Context, c *Container, cgroupPa
...
+	if err != nil {
+		return fmt.Errorf("failed to get image details for container: %w", err)
+	}
+	if archRuntimePath, ok := r.handler.ArchRuntimePaths[imageResult.OCIConfig.Platform.Architecture]; ok {
```
kolyshkin, line 126: what if the map is `nil`?

sohankunkerkar: good catch, fixed!

internal/oci/runtime_oci.go
sohankunkerkar: Yeah, I agree with you. At the same time, I feel getting platform details of an image required creating an `ImageServer` that needs a bunch of other stuff to compute. I'm happy to hear the better suggestions here.

haircommander: instead, I think you should execute this in server/ where we already have the image store initialized. There, you can maybe call a oci.Container function SetArch(), and here we can just call c.Arch() and see if it's defined in the runtime table

test/image.bats
```diff
@@ -293,3 +293,27 @@ function teardown() {
...
+[crio.runtime.runtimes.crun-wasm]
+runtime_path = "/usr/bin/crun"
+
+[crio.runtime.runtimes.crun-wasm.arch_runtime_paths]
+wasm = "/usr/bin/crun-wasm"
+EOF
+	start_crio
+
+	# Assuming the $IMAGE_LIST_DIGEST environment variable contains the new image value
+	IMAGE_LIST_DIGEST="quay.io/sohankunkerkar/test-wasm:latest"
```
sohankunkerkar, line 308: keeping this for testing purposes. I will move it to `quay.io/cri-o` once the test executes successfully.

internal/oci/container.go
```diff
@@ -755,3 +756,13 @@ func (c *Container) SetResources(s *specs.Spec) {
 func (c *Container) GetResources() *types.ContainerResources {
 	return c.resources
 }
+
+// SetArch sets the architecture of the image for a container.
+func (c *Container) SetArch(arch string) {
```
haircommander, line 761: TODO: we'll need to make sure this is called in server restore code

server/container_create_linux.go
```diff
@@ -867,6 +867,8 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	ociContainer.SetSpec(specgen.Config)
 	ociContainer.SetMountPoint(mountPoint)
 	ociContainer.SetSeccompProfilePath(seccompRef)
+	// Set the architecture of the image for container
+	ociContainer.SetArch(containerInfo.Config.Platform.Architecture)
```
haircommander, line 871: TODO: do we need to account for os as well? https://github.com/containers/podman/pull/16434/files#diff-c15f51066222fb402fefe74faa0f789de498024988e4a0414f5713f3b40bdf33R628

sohankunkerkar: I don't think so. We are already extracting the Arch from the [Platform](https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/config.go#L101) struct.

test/image.bats
```diff
@@ -293,3 +294,29 @@ function teardown() {
 
 	cleanup_images
 }
+
+@test "run container in pod with crun-wasm enabled" {
```
haircommander, line 298: we need to skip if RUNTIME_TYPE is vm as well

haircommander:
`ok 163 run container in pod with crun-wasm enabled`
:tada:

internal/oci/runtime_oci.go
```diff
@@ -525,14 +532,21 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
...
 		"-n", c.name,
-		"-r", r.handler.RuntimePath,
 		"-p", pidFile,
 		"-e",
 		"-l", logPath,
 		"--socket-dir-path", r.config.ContainerAttachSocketDir,
 		"--log-level", logrus.GetLevel().String(),
 	}
 
+	if archRuntimePath, ok := r.handler.ArchRuntimePaths[c.GetArch()]; ok {
```
haircommander, line 542: the equivalent piece in podman uses both os and platform https://github.com/containers/common/blob/cac40138f7e3c2b29ca32e64348535516bf6aa51/pkg/config/config.go#L695

@flouthoc what are some cases we need to include os as well as arch here? like does the runtime potentially differ based on OS between different wasm arches?

flouthoc: @haircommander There are some defaults here https://github.com/containers/common/pull/1225, so i think cases can be pretty much `wasm/wasi*`

haircommander: I am wary of supporting wildcards if we don't need to. I suppose we could just have this be os to path. Do you see a world where we'd need a different runtime for a different arch @flouthoc ?

flouthoc: My bad no I did not mean wildercard it was just to mention all three formats 
```
"wasi/wasm
"wasi/wasm32
"wasi/wasm64
```

haircommander: I am leaning towards including os and arch similar to podman to have the two implementations line up

haircommander: works for me

mrunalp: Golang recently added something for wasm for OS and Arch. Let's look at that.

flouthoc: Yes, Go added `OS=wasip1` for `wasi_snapshot_preview1` so maybe also add `wasip1/wasm`

docs/crio.conf.5.md
```diff
@@ -331,6 +331,9 @@ The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  Th
   "io.kubernetes.cri-o.UnifiedCgroup.$CTR_NAME" for configuring the cgroup v2 unified block for a container.
   "io.containers.trace-syscall" for tracing syscalls via the OCI seccomp BPF hook.
 
+**platform_runtime_paths**={}
```
haircommander, line 334: the structure of this is no longer correct, this should be a table

internal/oci/container.go
```diff
@@ -755,3 +756,18 @@ func (c *Container) SetResources(s *specs.Spec) {
 func (c *Container) GetResources() *types.ContainerResources {
 	return c.resources
 }
+
+// SetPlatformForImage sets the platform of the image for container.
+func (c *Container) SetPlatformForImage(os, arch string) { //nolint:gocritic
```
haircommander, line 761: what was gocritic complaining about?

internal/oci/container.go
```diff
@@ -755,3 +756,18 @@ func (c *Container) SetResources(s *specs.Spec) {
...
+	c.platform = fmt.Sprintf("%s/%s", os, arch)
+}
+
+// GetRuntimePath returns the runtime path for container.
+func (c *Container) GetRuntimePath(r *runtimeOCI) string {
```
haircommander, line 766: nit: usually go recommends against prepending getters with "Get" instead just going for the variable it's getting: `RuntimePath()`

That said, it may be more self documenting if we mention `RuntimePathGivenPlatform` or something

server/container_create_linux.go
```diff
@@ -867,6 +867,8 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	ociContainer.SetSpec(specgen.Config)
 	ociContainer.SetMountPoint(mountPoint)
 	ociContainer.SetSeccompProfilePath(seccompRef)
+	// Set the platform of the image for container
+	ociContainer.SetPlatformForImage(containerInfo.Config.Platform.OS, containerInfo.Config.Platform.Architecture)
```
haircommander, line 871: I think you should move https://github.com/cri-o/cri-o/pull/7180/files#diff-bcc340e07299e056e27ad412ee0817b6f9d8f7350834c1a4214cf1e253f66264R767-R772 here. Basically, the runtime handler and platform won't change for the lifecycle of the container. You can call something like a `SetRuntimePathGivenPlatform` function, which you pass the platform, runtime handler config and it sets the RuntimePath. Then, instead of doing a map lookup each `RuntimePathGivenPlatform()`, you are just returning a saved string

internal/oci/container.go
sohankunkerkar: It's about `os`, since we are also importing `os` package, it got confused with that keyword here

pkg/config/template.go
```diff
@@ -1252,6 +1255,9 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
 {{ if $runtime_handler.AllowedAnnotations }}{{ $.Comment }}allowed_annotations = [
 {{ range $opt := $runtime_handler.AllowedAnnotations }}{{ $.Comment }}{{ printf "\t%q,\n" $opt }}{{ end }}{{ $.Comment }}]{{ end }}
 {{ $.Comment }}privileged_without_host_devices = {{ $runtime_handler.PrivilegedWithoutHostDevices }}
+{{ if $runtime_handler.PlatformRuntimePaths }}{{ $.Comment }}platform_runtime_paths = {
+{{ range $key, $value := $runtime_handler.PlatformRuntimePaths }}{{ $.Comment }}{{ printf "\t%q = %q,\n" $key $value }}{{ end }}{{ $.Comment }}}
+{{ end }}
```
haircommander, line 1260: also the format here is not right, see https://github.com/cri-o/cri-o/pull/7180/files#diff-0b5b5d1619dddec727d1e347a5e9777d71b98ec5f35034690eb8b5693cf5ab4bR309-R310

internal/oci/container.go
haircommander: can you use `OS` instead so it doesn't shadow?

test/image.bats
```diff
@@ -293,3 +294,30 @@ function teardown() {
...
+	cat << EOF > "$CRIO_CONFIG_DIR/99-crun-wasm.conf"
+[crio.runtime]
+default_runtime = "crun-wasm"
+
+[crio.runtime.runtimes.crun-wasm]
+runtime_path = "/usr/bin/crun"
+
+[[crio.runtime.runtimes.crun-wasm.platform_runtime_paths]]
+platform = "wasi/wasm32"
+runtime_path = "/usr/bin/crun-wasm"
```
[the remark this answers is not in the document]
mrunalp: Can we do something like this: 
```
[crio.runtime.runtimes.crun]

platform_runtimes_paths = { "wasi/wasm32":  "/usr/bin/crun-wasm", "abc/def", "/usr/bin/acme" }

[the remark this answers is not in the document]
Comment (sohankunkerkar) [reply test/image.bats:311]: @mrunalp Are you suggesting that instead of creating a separate table, we should include the `platform_runtime_paths` field in the same RuntimeHandlers table?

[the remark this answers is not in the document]
Comment (mrunalp) [reply test/image.bats:311]: Yes.

File: test/image.bats [hunk 14]
```diff
@@ -293,3 +294,29 @@ function teardown() {
...
+	fi
+	cat << EOF > "$CRIO_CONFIG_DIR/99-crun-wasm.conf"
+[crio.runtime]
+default_runtime = "crun-wasm"
+
+[crio.runtime.runtimes.crun-wasm]
+runtime_path = "/usr/bin/crun"
+
+[crio.runtime.runtimes.crun-wasm.platform_runtime_paths]
+"wasi/wasm32" = "/usr/bin/crun-wasm"
```
haircommander, line 310: doesn't match what you generate in the template above

sohankunkerkar: Fixed

test/image.bats
```diff
@@ -293,3 +294,28 @@ function teardown() {
...
+		skip "crun-wasm not installed or runtime type is VM"
+	fi
+	cat << EOF > "$CRIO_CONFIG_DIR/99-crun-wasm.conf"
+[crio.runtime]
+default_runtime = "crun-wasm"
+
+[crio.runtime.runtimes.crun-wasm]
+runtime_path = "/usr/bin/crun"
+
+platform_runtime_paths = {"wasi/wasm32" = "/usr/bin/crun-wasm", "abc/def" = "/usr/bin/acme"}
```
mrunalp, line 309: Can drop the fake "abc/def" here.

sohankunkerkar: I had an offline discussion with @mrunalp. This change is only meant for testing purposes.

internal/lib/container_server.go
```diff
@@ -439,6 +439,11 @@ func (c *ContainerServer) LoadContainer(ctx context.Context, id string) (retErr
 		imgRef = ""
 	}
 
+	platformRuntimePath, ok := m.Annotations[crioann.PlatformRuntimePath]
+	if !ok {
+		platformRuntimePath = ""
+	}
+
```
haircommander, line 446: nit: this can be acheieved with 
```
platformRuntimePath, _ := m.Annotations[crioann.PlatformRuntimePath]
```

server/container_create_linux.go
```diff
@@ -652,8 +652,13 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		// TODO: patch runtime-tools to support setting ClosID via a helper func similar to SetLinuxIntelRdtL3CacheSchema()
 		specgen.Config.Linux.IntelRdt = &rspec.LinuxIntelRdt{ClosID: rdt.ResctrlPrefix + rdtClass}
 	}
-
-	err = ctr.SpecAddAnnotations(ctx, sb, containerVolumes, mountPoint, containerImageConfig.Config.StopSignal, imgResult, s.config.CgroupManager().IsSystemd(), node.SystemdHasCollectMode(), seccompRef)
+	// compute the runtime path for a given container
+	platform := fmt.Sprintf("%s/%s", containerInfo.Config.Platform.OS, containerInfo.Config.Platform.Architecture)
```
haircommander, line 656: nit: you can just use string concatination for this, it's a bit more efficient:
```
platform := containerInfo.Config.Platform.OS +"/"+ containerInfo.Config.Platform.Architecture
```

internal/lib/container_server.go
sohankunkerkar: ok, I thought of maintaining the same code styling as this one: https://github.com/cri-o/cri-o/blob/main/internal/lib/container_server.go#L437

haircommander: IMO that one is wrong too, though it predates me so I hadn't noticed

haircommander: chatted offline, my suggestion gives linter errors (that are also wrong IMO :upside_down_face: )

haircommander:
/approve

LGTM, well done here @sohankunkerkar 

@cri-o/cri-o-maintainers PTAL

server/container_create_linux.go
```diff
@@ -867,6 +872,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	ociContainer.SetSpec(specgen.Config)
 	ociContainer.SetMountPoint(mountPoint)
 	ociContainer.SetSeccompProfilePath(seccompRef)
+	ociContainer.SetRuntimePathForPlatform(runtimePath)
```
mrunalp, line 875: Should this only be set when not empty?

sohankunkerkar: That's fine bc later on we will check and assign the value: https://github.com/cri-o/cri-o/pull/7180/files#diff-bcc340e07299e056e27ad412ee0817b6f9d8f7350834c1a4214cf1e253f66264R766

mrunalp (approved):
/lgtm

PR #7184, opened 2023-07-31
OCPBUGS-11291: skip when ListImage with filter image hasn't done pulling

QiWang19:
#### What type of PR is this?

/kind bug

Apply a previous fix https://github.com/cri-o/cri-o/pull/4726 to ListImage with `filter`.
If the file not found error was caused by image not done pulling when ListImages() with filter, skip the image and the error.

Previous bug https://bugzilla.redhat.com/show_bug.cgi?id=1942608  appeared in https://issues.redhat.com/browse/OCPBUGS-11291. 
The log hyperkube ListImages with filters. `"ListImages with filter from image service failed" err="rpc error: code = Unknown desc = error locating item named \"manifest\" for image with ID \"9e11cc7364c9791ebf68949c852bf121cb4ae5829c925a1da6546be6f63b3536\": file does not exist" filter="nil"`
https://issues.redhat.com/browse/OCPBUGS-11291

```release-note
None
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -332,6 +332,11 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext, filter s
 		if image, err := istorage.Transport.GetStoreImage(svc.store, ref); err == nil {
 			results, err = svc.appendCachedResult(systemContext, ref, image, []ImageResult{}, nil)
 			if err != nil {
+				if os.IsNotExist(err) {
+					if !imageDonePulling(image) {
+						return nil, nil
+					}
+				}
 				return nil, err
 			}
 		}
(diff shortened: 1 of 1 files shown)


internal/storage/image.go
```diff
@@ -374,6 +372,17 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext, filter s
 	return results, nil
 }
 
+func imageDonePulling(image *storage.Image) bool {
+	donePulling := true
+	for _, name := range image.Names {
+		if _, ok := ImageBeingPulled.Load(name); ok {
+			donePulling = false
```
haircommander, line 379: we can just `return false` here, and `return true` below, no need for donePulling if we separate the functions

QiWang19: thanks. fixed

QiWang19:
/hold
I am not sure this fix makes sense.

QiWang19:
I don't think this PR makes sense for the bug.

PR #7190, opened 2023-08-04
feat: check for clean shutdown on reboot

### LINKED ISSUE #7177: make use of c/storage Check() and Repair() functions
currently, crio wipe checks if the node improperly shut down, and if so just removes the whole image store. this is a very heavy hammer. in https://github.com/containers/storage/pull/1569 @nalind added the abiliity to do consistency checks and repair the OS. We should update crio wipe to utilize them instead of removing the directory.

MdSahil-oss:
#### What type of PR is this?
/kind feature

This PR adds a feature to verify at reboot time that shutdown was clean or not, If it was not clean then applies repair logic implemented [here](https://github.com/containers/storage/pull/1569/files#diff-1f42697300efefc989929182bd025d0bee2c5b2512708659550ab314f30fba0dR18-R42)
Fixes #7177 

```release-note
Attempt to call c/storage Check() and Repair() when an unclean shutdown is detected
```


internal/lib/container_server.go
```diff
@@ -769,3 +784,19 @@ func (c *ContainerServer) UpdateContainerLinuxResources(ctr *oci.Container, reso
 
 	c.state.containers.Add(ctr.ID(), ctr)
 }
+
+func shutdownWasUnclean(config *libconfig.Config) bool {
```
haircommander, line 788: instead of just copy-pasting, can you export this function somewhere and import it in both places? Ideally export here because I think internal/criocli already needs to import this package

MdSahil-oss: @haircommander done 👍🏽

haircommander:
the structure here looks good, I'm writing a test case to validate it, so let's 
/hold
for that. Thanks @MdSahil-oss !

MdSahil-oss:
Sure @haircommander I can cherry pick https://github.com/haircommander/cri-o/commit/bb5c01ec4011052015519b24096628a8528a9411.

haircommander:
```

go build  -trimpath  -ldflags '-s -w -X github.com/cri-o/cri-o/internal/version.buildDate='2023-08-08T19:35:28Z' ' -tags "selinux seccomp exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_ostree_stub containers_image_openpgp" -o bin/crio github.com/cri-o/cri-o/cmd/crio
# github.com/cri-o/cri-o/internal/lib
internal/lib/container_server.go:116:23: undefined: errors.Join 
```
compile error

internal/lib/container_server.go
```diff
@@ -102,6 +104,24 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
 		return nil, fmt.Errorf("cannot create container server: interface is nil")
 	}
 
+	if ShutdownWasUnclean(config) {
+		checkOptions := cstorage.CheckEverything()
+		report, err := store.Check(checkOptions)
+		if err != nil {
+			return nil, err
```
haircommander, line 111: @nalind I would say, ideally, cri-o does not exit in this case. What cases would cause this to fail? do you think it makes sense to just remove the c/storage like we did before in this error case?

nalind: Generally, these errors are going to be about permissions or locking.  If returning here means we go from automatically wiping storage to requiring manual intervention, that seems excessive.

haircommander: agreed. @MdSahil-oss can you refactor so `handleCleanShutdown` (which really should be called `handleUncleanShutdown`) is also exported and called if the check or repair fail?

MdSahil-oss: @haircommander done 👍🏽

haircommander:
/approve

LGTM, thanks!
@cri-o/cri-o-maintainers PTAL

haircommander:
@nalind one of the tests is failing because the images are wiped when the test doesn't expect it:
https://github.com/cri-o/cri-o/actions/runs/5806255088/job/15749116292?pr=7190
the situation: cri-o is foracbly stopped but there should not be a corruption. Does anything in the logs indicate why? also, some strange behavior I see is that the image is removed but the container is still able to be restored. is that expected?

nalind:
@haircommander not sure.  I modified a local checkout of this PR to log the report that `Check()` returns to hopefully see what it didn't like about the image or one of its layers, but the tests aren't failing when I run them, either with or without `TEST_USERNS=true` set in the environment.

haircommander:
yeah I was also not able to reproduce...

> also, some strange behavior I see is that the image is removed but the container is still able to be restored. is that expected?

WDYT of this @nalind ? do we expect the image to be removed if there's a running container?

nalind:
No, I wouldn't expect containers to exist in any form after we've removed the images that they're based on.
I've seen this _kind of_ happen elsewhere when CRI-O was using `transient_store` (which stores information about containers in a different location) and someone ran a `podman prune` was built using a storage library that predates that option, and which therefore had no idea it was creating such an inconsistency by removing images that it was unaware were in use.

haircommander:
@nalind helped me debug this issue, it comes from the kernel's handling of chowning files in the lower layer (which is done on userns tests because we need to chown the files for the UID mapping). 

I am thinking we can skip if USERNS, but I am also thinking this will break any user namespace usage in cri-o, so we probably have to put it behind a flag until it's fixed in the kernel.

haircommander:
@MdSahil-oss I can work with you monday to add a flag to toggle this behavior.

glennswest:
The is a significant improvement. One further enhancement to this would be to do this async. Currently this validation/fixup happens before crio operation. The issue that concerns me, is in large systems, this may have significant runtime, resulting in crio being unavailable for hours. I recommend that in future that this be async to crio start. So once we move to this, we can let this happen async to crio start. Yes, some containers will fail due to bad images, but most of the time everything starts fast and clean. If a image is bad, it will get cleaned up, and the restart will happen. This will work far better in large installs.

haircommander:
> resulting in crio being unavailable for hours

have you observed this or is this a hunch?

internal/criocli/criocli.go
```diff
@@ -1137,6 +1137,11 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			EnvVars: []string{"DISABLE_HOSTPORT_MAPPING"},
 			Value:   defConf.DisableHostPortMapping,
 		},
+		&cli.BoolFlag{
+			Name:  "internal-repair",
+			Usage: "If true, CRI-O will check if contents of an image was modified or removed and repair the image.",
+			Value: defConf.InternalRepair,
```
haircommander, line 1143: can you add an EnvVars `CONTAINER_INTERNAL_REPAIR`?

MdSahil-oss: Done 👍🏽

test/crio-wipe.bats
```diff
@@ -320,3 +320,26 @@ function start_crio_with_stopped_pod() {
...
+	touch "$CONTAINER_CLEAN_SHUTDOWN_FILE.supported"
+
+	# Remove a random layer
+	layer=$(find "$TESTDIR/crio/overlay" -maxdepth 1 -regextype sed -regex '.*/[a-f0-9\-]\{64\}.*' | sort -R | head -n 1)
+	rm -fr "$layer"
+
+	# Since the clean shutdown supported file is created,
+	# but the clean shutdown file is absent, we will do the
+	# c/storage check/repair.
+	start_crio_no_setup
```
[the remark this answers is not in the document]
MdSahil-oss: Done 👍🏽

haircommander:
@MdSahil-oss the test is failing because there are a handful more places you need to wire the configuration to. I have added two commits I think should be picked here and squashed into your other commits: https://github.com/haircommander/cri-o/tree/wipe-check-shut-down

internal/criocli/criocli.go
```diff
@@ -1137,6 +1137,12 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			EnvVars: []string{"DISABLE_HOSTPORT_MAPPING"},
 			Value:   defConf.DisableHostPortMapping,
 		},
+		&cli.BoolFlag{
+			Name:    "internal-repair",
+			Usage:   "If true, CRI-O will check if contents of an image was modified or removed and repair the image.",
```
haircommander, line 1142: nit: I think this should be
```
If true, CRI-O will check if the container and image storage was corrupted after a sudden restart, and attempt to repair the storage if it was.
```

MdSahil-oss: @haircommander Done 👍🏽

haircommander:
>    Aug 21 15:45:53.149: INFO: Unexpected error occurred: rpc error: code = Unknown desc = container create failed: time="2023-08-21T15:45:53Z" level=error msg="runc create failed: unable to start container process: container init was OOM-killed (memory limit too low?)" 

[unexpected](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/7190/pull-ci-cri-o-cri-o-main-ci-rhel-critest/1693646036213436416)
/retest

haircommander:
/approve

LGTM, thanks @MdSahil-oss 

PTAL @cri-o/cri-o-maintainers

saschagrunert (approved):

Issue #6699, opened 2023-03-02
Containers can get stuck in stopping state until cri-o is restarted


PaulFurtado:
### What happened?

A pod is deleted and kubelet tells the container to stop
```
Mar 02 19:05:32 ip-172-18-52-92 crio[3133611]: time="2023-03-02 19:05:32.885463979Z" level=info msg="Stopping container: 252c0f00d0e3f9ff99016b2dbc7fe4c2877b29d53f6d8a66dbc0fb81710ae8a9 (timeout: 30s)" id=4ee0306c-e4b3-43c1-9f1f-747364e6edb1 name=/runtime.v1.RuntimeService/StopContainer
```
The stop times out
```
Mar 02 19:06:03 ip-172-18-52-92 crio[3133611]: time="2023-03-02 19:06:03.124419954Z" level=warning msg="Stopping container 252c0f00d0e3f9ff99016b2dbc7fe4c2877b29d53f6d8a66dbc0fb81710ae8a9 with stop signal timed out: timeout reached after 30 seconds waiting for container process to exit" id=4ee0306c-e4b3-43c1-9f1f-747364e6edb1 name=/runtime.v1.RuntimeService/StopContainer
```
But it appears that the cgroups, etc get torn down anyway:
```
Mar 02 19:08:02 ip-172-18-52-92 crio[3133611]: time="2023-03-02 19:08:02.885652165Z" level=warning msg="Failed to retrieve cgroup stats: open /sys/fs/cgroup/memory/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod05ae14ac_623d_4685_a19b_2c6ecf03d24f.slice/crio-252c0f00d0e3f9ff99016b2dbc7fe4c2877b29d53f6d8a66dbc0fb81710ae8a9.scope/memory.usage_in_bytes: no such file or directory"
Mar 02 19:08:02 ip-172-18-52-92 crio[3133611]: time="2023-03-02 19:08:02.885701842Z" level=warning msg="Failed to retrieve cgroup stats: open /sys/fs/cgroup/pids/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod05ae14ac_623d_4685_a19b_2c6ecf03d24f.slice/crio-252c0f00d0e3f9ff99016b2dbc7fe4c2877b29d53f6d8a66dbc0fb81710ae8a9.scope/pids.current: no such file or directory"
```

[...]

PR #7196, opened 2023-08-08
feat(cmdrunner): add support for context

terinjokes:
Support passing context through to *exec.Command by adding CommandContext to the CommandRunner interface. This allows context cancellation to terminate the running command.

#### What type of PR is this?

/kind feature

Support creating an `*exec.Command` with a context, and link the lifetime of commands started by `ExecContainer` to that of the passed in context.

#6699 

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -409,12 +409,13 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
 
 	args := r.defaultRuntimeArgs()
 	args = append(args, "exec", "--process", processFile, c.ID())
-	execCmd := cmdrunner.Command(c.RuntimePathForPlatform(r), args...) // nolint: gosec
+	execCmd := cmdrunner.CommandContext(ctx, c.RuntimePathForPlatform(r), args...) // nolint: gosec
 	if v, found := os.LookupEnv("XDG_RUNTIME_DIR"); found {
 		execCmd.Env = append(execCmd.Env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", v))
 	}
 	var cmdErr, copyError error
 	if tty {
+		execCmd.WaitDelay = 30 * time.Second
 		cmdErr = ttyCmd(execCmd, stdin, stdout, resizeChan)
 	} else {
 		var r, w *os.File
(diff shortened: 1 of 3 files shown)


terinjokes:
I update the PR to make  `cmdrunner.Command` call through to `cmdrunner.CommandContext` with `context.Background()`.

internal/oci/runtime_oci.go
```diff
@@ -409,7 +409,7 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
 
 	args := r.defaultRuntimeArgs()
 	args = append(args, "exec", "--process", processFile, c.ID())
-	execCmd := cmdrunner.Command(r.handler.RuntimePath, args...) // nolint: gosec
+	execCmd := cmdrunner.CommandContext(ctx, r.handler.RuntimePath, args...) // nolint: gosec
 	if v, found := os.LookupEnv("XDG_RUNTIME_DIR"); found {
 		execCmd.Env = append(execCmd.Env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", v))
```
terinjokes, line 414: Because the pty may not drain io, we may want `exec.WaitDelay`. Should since this needs a duration, should I defer to a separate PR?

haircommander: I think including it here makes sense to me. we could either hardcode it, or set it to the container's stop timeout (though that could be arbitrarily long which could defeat the purpose of this PR)

terinjokes: For now I can hardcode it.

terinjokes:
`WaitDelay` was added in [Go 1.20](https://go.dev/doc/go1.20#os/exec), which is what cri-o declares compatibility with, but I guess one of the test runners is using an older version. I can't easily find what releases cri-o officially supports.

terinjokes:
Still seeing errors suggesting the wrong version of Go is being used.

pims:
@haircommander is there something we can do to make sure all tests are running with Go >= 1.20 ?

sohankunkerkar (commented):
/lgtm

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

haircommander:
/approve

thanks for your patience here!

PR #7208, opened 2023-08-11
cgmgr: reorder setting of sched_load_balance for pod cgroup

haircommander:
#### What type of PR is this?

/kind bug
If we call mgr.Apply() first, libcontainer's cpusetCopyIfNeeded() will copy the parent cpuset and set load balancing to 1 by default. This causes the kernel to set the cpus to not load balanced for a brief moment which causes churn.

instead, create the cgroup and set load balance, then have Apply() copy the values into it.

```release-note
fix a bug in cpuset load balancing where cpusets flip between load balanced and not due to an ordering issue.
```

### DIFF
--- a/internal/config/cgmgr/cgmgr_linux.go
+++ b/internal/config/cgmgr/cgmgr_linux.go
@@ -180,23 +180,28 @@ func createSandboxCgroup(sbParent, containerCgroup string) error {
 		return err
 	}
 
-	if err := mgr.Apply(-1); err != nil {
-		return err
-	}
-
 	// The reasoning for this code is slightly obscure. In situation where CPU load balancing is desired,
 	// all cgroups must either have cpuset.sched_load_balance=0 or they should not have an intersecting cpuset
 	// with the set that load balancing should be disabled on.
 	// When this cgroup is created, it is easiest to set sched_load_balance to 0, especially because there will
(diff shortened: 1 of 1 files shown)


internal/config/cgmgr/cgmgr_linux.go
```diff
@@ -180,23 +180,28 @@ func createSandboxCgroup(sbParent, containerCgroup string) error {
...
+	// Note: this should be done before Apply(-1) below, as Apply contains cpusetCopyIfNeeded(), which will
+	// populate the cpuset with the parent's cpuset. However, it will be initialized to sched_load_balance=1
+	// which will cause the kernel to move all cpusets out of their isolated sched_domain, causing unnecessary churn.
 	if !node.CgroupIsV2() {
 		path := mgr.Path("cpuset")
 		if path == "" {
 			return fmt.Errorf("failed to find cpuset for newly created cgroup")
 		}
-		return libctr.WriteFile(path, "cpuset.sched_load_balance", "0")
+		if err := os.Mkdir(path, 0o755); err != nil {
```
rphillips, line 196: We should probably catch the Exists case.

```
err != nil && !os.IsExist(err)
```

kolyshkin (approved):
lgtm

rphillips:
lgtm

PR #7214, opened 2023-08-15
Allow crio to build on non-linux platforms

dfr:
This is a large change I'm afraid. I have split it into multiple commits, one for each affected directory. If it woud help, I can split this into multiple pull requests.

The main theme is to move obviously linux-only code to files with a _linux.go suffix and add and/or fix stubs for other platforms. This is enough to build the crio binary on FreeBSD but the result is not functional. Follow up PRs will add FreeBSD support.

#### What type of PR is this?

/kind cleanup

This helps to isolate the Linux-specific parts of the crio codebase and open up the possiblity of ports to other platforms such as FreeBSD (#6492).

None

```release-note
None
```


haircommander:
@saschagrunert WDYT about this for 1.28? if we wait, backporting will be harder but it is a large change so close to the deadline

dfr:
I don't think it should cause big problems with backports if this waits until after 1.28. I've been iterating on this stack of diffs for months now and haven't had big problems with merge conflicts - the code I'm changing here is quite stable.

As far as a working crio for FreeBSD goes, I have about 2000 lines of code to follow up on this baseline, mostly adding working freebsd versions of the stubs being added here. Outside of cri-o, I also have about 2000 lines of code to get kubelet working and another 1000 lines for kube-proxy which I need to clean up and get reviewed which is going to take time.

saschagrunert:
Alright so I assume we're fine merging this after 1.28. :+1:

saschagrunert (approved):

saschagrunert:
@dfr do we want to add a test for the non-linux build? Otherwise I fear we break your work in a matter of time.

dfr:
> @dfr do we want to add a test for the non-linux build? Otherwise I fear we break your work in a matter of time.

Good idea. I'll work on adding a FreeBSD cross-build target to CI.

dfr:
The single failing test seems unrelated to this PR.:
```
# (from function `wait_until_exit' in file ./helpers.bash, line 592,
#  in test file ./seccomp_notifier.bats, line 40)
#   `EXPECTED_EXIT_STATUS=137 wait_until_exit "$CTR"' failed
...
```

I'll try re-running it - hopefully its a flake.

dfr:
Rebased

internal/runtimehandlerhooks/utils_linux.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/cri-o/cri-o/utils/cmdrunner"
+	"github.com/sirupsen/logrus"
+	"k8s.io/utils/cpuset"
+)
+
+// UpdateIRQSmpAffinityMask take input cpus that need to change irq affinity mask and
+// the current mask string, return an update mask string and inverted mask, with those cpus
+// enabled or disable in the mask.
+func UpdateIRQSmpAffinityMask(cpus, current string, set bool) (cpuMask, bannedCPUMask string, err error) {
```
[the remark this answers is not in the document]
haircommander: as in: I think we can make all of the runtime handler hooks be linux only for now

pkg/config/config_freebsd.go
```diff
@@ -0,0 +1,38 @@
...
+
+	defaultRuntime       = "ocijail"
+	DefaultRuntimeType   = "oci"
+	DefaultRuntimeRoot   = "/var/run/ocijail"
+	defaultMonitorCgroup = ""
+
+	// ImageVolumesBind option is for using bind mounted volumes
+	ImageVolumesBind ImageVolumesType = "nullfs"
+	// DefaultPauseImage is default pause image
+	DefaultPauseImage string = "quay.io/dougrabson/pause:latest"
```
haircommander, line 37: eventually we could push this to the crio repo if you don't find another place for it

server/container_create_unsupported.go
```diff
@@ -17,6 +16,6 @@ func findCgroupMountpoint(name string) error {
 	return fmt.Errorf("no cgroups on this platform")
 }
 
-func (s *Server) createSandboxContainer(ctx context.Context, ctr container.Container, sb *sandbox.Sandbox) (*oci.Container, error) {
+func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Container, sb *sandbox.Sandbox) (cntr *oci.Container, retErr error) {
```
haircommander, line 19: nit: why the named return values?

haircommander:
sorry a couple more drive-bys. we should try to get this in soon IMO @cri-o/cri-o-maintainers

internal/runtimehandlerhooks/utils_linux.go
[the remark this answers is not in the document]
dfr: It should be possible to reduce the non-linux side down to stub implementations of GetRuntimeHandlerHooks, RestoreIrqBalanceConfig and IrqSmpAffinityProcFile. I'll try that approach.

pkg/config/config_freebsd.go
dfr: That would be nice. This image just contains a FreeBSD build of catatonit:
```
FROM registry.lab.rabson.org/freebsd13.2-minimal AS builder
RUN env ASSUME_ALWAYS_YES=yes pkg update && pkg install -y catatonit

FROM scratch
COPY --from=builder /usr/local/bin/catatonit /pause
CMD ["/pause", "-P"]
```
(ignore refeferences to my home lab infrastructure)

saschagrunert (approved):
I agree, let's get this in

dfr:
Thanks everyone! I'll try to make subsequent PRs for FreeBSD a bit less monolithic.

PR #7273, opened 2023-09-06
Update install.md

timmy61109:
#### What type of PR is this?

/kind other

```release-note
None
```

### DIFF
--- a/install.md
+++ b/install.md
@@ -181,15 +181,20 @@ apt update
 apt install -y -t buster-backports libseccomp2 || apt update -y -t buster-backports libseccomp2
 ```
 
+Need Packed
+
+```shell
+sudo apt install apt-transport-https ca-certificates curl gnupg2 software-properties-common -y
+```
+
 And then run the following as root:
 
 ```shell
-echo "deb [signed-by=/usr/share/keyrings/libcontainers-archive-keyring.gpg] https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/ /" > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
-echo "deb [signed-by=/usr/share/keyrings/libcontainers-crio-archive-keyring.gpg] https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/$VERSION/$OS/ /" > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable:cri-o:$VERSION.list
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -181,18 +181,23 @@ apt update
 apt install -y -t buster-backports libseccomp2 || apt update -y -t buster-backports libseccomp2
```
haircommander, line 181: can this section be updated with sudo if the section below is?

TomSweeneyRedHat:
LGTM
once @haircommander 's concerns are addressed.

sohankunkerkar:
@timmy61109 do you have the bandwidth to address @haircommander's comments and make this PR ready to merge?

timmy61109:
> @timmy61109 do you have the bandwidth to address @haircommander's comments and make this PR ready to merge?

I'm sorry, I don't really know what you mean by "bandwidth"?Can you explain this to me?

In addition, to clarify the above modification, I just refer to the official website's installation method, but the result is unsuccessful. I checked the official website and GitHub WiKi wrote differently, and then I read other people's articles on the internet, and I found that the method is actually the same as that written by GitHub WiKi, and after I organized it, I succeeded.

What I don't understand is why the `sudo` command is not provided, after all, not everyone can install as root, and the Ubuntu server version is also for general users.

Translated with www.DeepL.com/Translator (free version)

haircommander:
>   I'm sorry, I don't really know what you mean by "bandwidth"?Can you explain this to me?

bandwidth in this case is "capacity" as in: do you have enough time and energy to address what I've said

> What I don't understand is why the sudo command is not provided, after all, not everyone can install as root, and the Ubuntu server version is also for general users.

We usually assume CRI-O will be run and installed as root and by a user who has root permissions, which is why we didn't really add sudo before. I am fine adding the sudo, but we should do it consistently through the whole document

timmy61109:
> > I'm sorry, I don't really know what you mean by "bandwidth"?Can you explain this to me?
> 
> bandwidth in this case is "capacity" as in: do you have enough time and energy to address what I've said
> 
> > What I don't understand is why the sudo command is not provided, after all, not everyone can install as root, and the Ubuntu server version is also for general users.
> 
> We usually assume CRI-O will be run and installed as root and by a user who has root permissions, which is why we didn't really add sudo before. I am fine adding the sudo, but we should do it consistently through the whole document

我了解了，我後續會找時間來修改，請問目前有多少篇呢？

I understand, I will find time to revise it later, may I ask how many articles there are at present?

haircommander:
> I understand, I will find time to revise it later, may I ask how many articles there are at present?

you can just focus on `install.md` for now. thanks!

timmy61109:
> > I understand, I will find time to revise it later, may I ask how many articles there are at present?
> 
> you can just focus on `install.md` for now. thanks!

OK

PR #7285, opened 2023-09-08
Remove update container status

### LINKED ISSUE #7221: Calling stateCmd twice in the function UpdateContainerStatus is unnecessary.
Hi All,
After the first execution of stateCmd, the container will only continue to execute if it is in the stopped state. The exit code of the container is then read. As the container cannot transition from the stopped state to other states (such as creating, created, running), the second execution of stateCmd should not be necessary. 
I hope it makes sense to you :)

```func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) error {
	defer span.End()
	c.opLock.Lock()
	defer c.opLock.Unlock()

	if c.Spoofed() {
		return nil
	}

	if c.state.ExitCode != nil && !c.state.Finished.IsZero() {
		log.Debugf(ctx, "Skipping status update for: %+v", c.state)
		return nil
	}

	stateCmd := func() (*ContainerState, bool, error) {
		out, err := r.runtimeCmd("state", c.ID())
		if err != nil {
			// there are many code paths that could lead to have a bad state in the
			// underlying runtime.
			// On any error like a container went away or we rebooted and containers
			// went away we do not error out stopping kubernetes to recover.
			// We always populate the fields below so kube can restart/reschedule
			// containers failing.
			if exitErr, isExitError := err.(*exec.ExitError); isExitError {
				log.Errorf(ctx, "Failed to update container state for %s: stdout: %s, stderr: %s", c.ID(), out, string(exitErr.Stderr))
			} else {
				log.Errorf(ctx, "Failed to update container state for %s: %v", c.ID(), err)
			}
			c.state.Status = ContainerStateStopped
			if err := updateContainerStatusFromExitFile(c); err != nil {
				c.state.Finished = time.Now()
				c.state.ExitCode = utils.Int32Ptr(255)
			}
			return nil, true, nil
		}
		state := *c.state
		if err := json.NewDecoder(strings.NewReader(out)).Decode(&state); err != nil {
			return &state, false, fmt.Errorf("failed to decode container status for %s: %s", c.ID(), err)
		}
		return &state, false, nil
	}
	state, canReturn, err := stateCmd()
	if err != nil {
		return err
	}
	if canReturn {
		return nil
	}

	if state.Status != ContainerStateStopped {
		*c.state = *state
		return nil
	}
	// release the lock before waiting
	c.opLock.Unlock()
	exitFilePath := c.exitFilePath()
	err = kwait.ExponentialBackoff(
		kwait.Backoff{
			Duration: 500 * time.Millisecond,
			Factor:   1.2,
			Steps:    6,
		},
		func() (bool, error) {
			_, err := os.Stat(exitFilePath)
			if err != nil {
				// wait longer
				return false, nil
			}
			return true, nil
		})
	c.opLock.Lock()
	// run command again
	state, _, err2 := stateCmd()
	if err2 != nil {
		return err2
	}
	if state == nil {
		return fmt.Errorf("state command returned nil")
	}
	*c.state = *state
	if err != nil {
		log.Warnf(ctx, "Failed to find container exit file for %v: %v", c.ID(), err)
	} else {
		if err := updateContainerStatusFromExitFile(c); err != nil {
			return err
		}
		log.Debugf(ctx, "Found exit code for %s: %d", c.ID(), *c.state.ExitCode)
	}

	oomFilePath := filepath.Join(c.bundlePath, "oom")
	if _, err = os.Stat(oomFilePath); err == nil {
		c.state.OOMKilled = true

		// Collect total metric
		metrics.Instance().MetricContainersOOMTotalInc()

		metrics.Instance().MetricContainersOOMCountTotalInc(c.Name())
	}
	// If this container had a node level PID namespace, then any children processes will be leaked to init.
	// Eventually, the processes will get cleaned up when the pod cgroup is cleaned by the kubelet,
	// but this situation is atypical and should be avoided.
	if c.nodeLevelPIDNamespace() {
		return r.signalContainer(c, syscall.SIGKILL, true)
	}
	return nil
}```

https://github.com/cri-o/cri-o/blob/d7a647bf0d0d34731baa96e5194262879d2a17ed/internal/oci/runtime_oci.go#L1008C30-L1008C30

### PR DESCRIPTION
**Pull request by Eresear**
#### What type of PR is this?
/kind cleanup

Remove UpdateContainerStatus
1. When the handleExit function is executed, entering ContainerStateToDisk will also call updatecontainerstatus. There is no need to call it repeatedly.
2. s.StorageRuntimeServer().StopContainer(ctx, ctr.ID()) does not need to obtain the latest status of the container, so there is no need to call updatecontainerstatus in advance. It will be called by ContainerStateToDisk later on.

No
No

```release-note
none
```


haircommander:
/approve

great find! LGTM, @cri-o/cri-o-maintainers PTAL

haircommander:
@Eresear could you also drop the UpdateContainerStatus() call in ContainerStateFromDisk? it's similarly duplicated.

saschagrunert (approved):
/retest

saschagrunert:
> @Eresear could you also drop the UpdateContainerStatus() call in ContainerStateFromDisk? it's similarly duplicated.

/hold

haircommander:
/hold cancel
/retest

we can get this in and follow up

Issue #7262, opened 2023-08-29
Check idmap mounts support before creating a container


rata:
### What happened?

Hi!

I wanted to raise awareness on the fact that is not safe to use idmap mounts in high-level container runtimes without checking the OCI runtime features support. 

The thing is, old runtimes (withot idmap mounts support)  can't fail on unknown fields because the spec [requires to ignore them][spec]. therefore, old OCI runtimes ignore the idmap mounts configs (the new `uidMapping` and `gidMapping` on mounts), which is an issue to use it safely (more on this later). This is sad, but that is how runtimes behave now.

For crun probably this is not an issue here, as you release them in lockstep, right? So CRI-O XX goes with crun YY, and that is always the case. Am I right? In that case it all works, as you can make sure crun YY supports idmap mounts and all is done.

The issue, IIUC, comes with runc support. If CRI-O is used with an old runc that ignore idmap mounts and just start the container, volumes will be mixed up. The UID/GID will be the hostUID/GID of the pod, that is basically garbage, and completely unusable as the hostUID/GID is not guaranteed to be stable over time.

In containerd I'm currently doing the following, to handle this case and give a proper error to users:
 * If userns are in use ==> run "runc features" and check for idmap mounts support.
 * If idmap mounts supports is there, go ahead and create the container. Otherwise, throw a clear error.
 * The container creation might work or not (depends if the host supports idmap mounts for all the fs), but it will never be silently ignored. In the worst case, runc throws an error.

You can check the code I wrote to do this in containerd [here](
https://github.com/containerd/containerd/pull/8287/files#diff-17085223b7307c8c0ed7f2cdcfb904efa5a184a71b186a83961d7feef9e3556fR145-R150).

Exposing info on idmap mounts support is something I did in the runtime-spec due to this issue (not being able to use idmap mounts safely from a high-level container runtime). You can see the PR here: https://github.com/opencontainers/runtime-spec/pull/1219

I've also added support for that in runc and crun:
 * https://github.com/containers/crun/pull/1279
 * https://github.com/opencontainers/runc/pull/3993

Although these changes are not yet included on a release.

Let me know if something is not clear :)

[spec]: https://github.com/opencontainers/runtime-spec/blob/e8c413418efcb816b963be9762ddc202e181ba88/config.md#extensibility

### What did you expect to happen?

---

### How can we reproduce it (as minimally and precisely as possible)?

---

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
# paste output here
[...]

PR #7289, opened 2023-09-08
config/server: add functions to check IDMap support in runtime

haircommander:
if we don't, we risk container creation failures or security holes from the runtime not correctly setting up idmaps

#### What type of PR is this?

/kind feature

```release-note
Check the runtime supports IDMap support before specifying 
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -321,7 +321,8 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	cgroup2RW := node.CgroupIsV2() && sb.Annotations()[crioann.Cgroup2RWAnnotation] == "true"
 
 	s.resourceStore.SetStageForResource(ctx, ctr.Name(), "container volume configuration")
-	containerVolumes, ociMounts, err := addOCIBindMounts(ctx, ctr, mountLabel, s.config.RuntimeConfig.BindMountPrefix, s.config.AbsentMountSourcesToReject, maybeRelabel, skipRelabel, cgroup2RW, s.Config().Root)
+	idMapSupport := s.Runtime().RuntimeSupportsIDMap(sb.RuntimeHandler())
+	containerVolumes, ociMounts, err := addOCIBindMounts(ctx, ctr, mountLabel, s.config.RuntimeConfig.BindMountPrefix, s.config.AbsentMountSourcesToReject, maybeRelabel, skipRelabel, cgroup2RW, idMapSupport, s.Config().Root)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


haircommander:
cc @rata

sohankunkerkar:
Fixes https://github.com/cri-o/cri-o/issues/7262

server/container_create_linux.go
```diff
@@ -1064,12 +1065,17 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
 			SelinuxRelabel: m.SelinuxRelabel,
 		})
 
+		uidMappings := getOCIMappings(m.UidMappings)
+		gidMappings := getOCIMappings(m.GidMappings)
+		if (uidMappings != nil || gidMappings != nil) && !idMapSupport {
+			return nil, nil, fmt.Errorf("IDMappings specified but runtime does not support them")
```
rata, line 1071: This is shown like this in the kubectl describe output:
```
 Warning  Failed                  56s (x12 over 3m5s)  kubelet, 127.0.0.1  Error: IDMappings specified but runtime does not support them

```

I wonder if we can make it clearer for the user what they need to do? Maybe it is fine as it is, but maybe we can give more actionable info? I can't think of anything now, but:

```suggestion
			return nil, nil, fmt.Errorf("idmap mounts specified but OCI runtime does not support them, OCI runtime too old?")
```

server/container_create_linux_test.go
```diff
@@ -159,6 +160,47 @@ func TestAddOCIBindsCGroupRW(t *testing.T) {
 	}
 }
 
+func TestAddOCIBindsErrorWithoutIDMap(t *testing.T) {
+	ctr, err := container.New()
+	if err != nil {
+		t.Error(err)
```
rata, line 166: Do we want t.Fatal() here, so the execution doesn't continue?

```suggestion
		t.Fatal(err)
```

server/container_create_linux_test.go
```diff
@@ -159,6 +160,47 @@ func TestAddOCIBindsCGroupRW(t *testing.T) {
...
+		},
+		Metadata: &types.ContainerMetadata{
+			Name: "testctr",
+		},
+	}, &types.PodSandboxConfig{
+		Metadata: &types.PodSandboxMetadata{
+			Name: "testpod",
+		},
+	}); err != nil {
+		t.Error(err)
```
rata, line 191:
```suggestion
		t.Fatal(err)
```

rata (approved):
LGTM, thanks!

Left some minot comments, even if you want to act on them, I don't think I need to review again :)

haircommander:
@cri-o/cri-o-maintainers PTAL 

This is related to https://github.com/kubernetes/website/pull/42778/ which is due tomorrow. Ideally we'd have a version of this PR backported to there and released as 1.28.1

sohankunkerkar:
You may want to look into resolving some of the failing GitHub action tests.

sohankunkerkar (approved):
/lgtm
/retest

PR #7302, opened 2023-09-13
RFC: Make some subpackages build (not work) on non-Linux systems

mtrmac:
#### What type of PR is this?

/kind cleanup

Change various subpackages so that `./server` and `./internal/storage` can be built on non-Linux systems, just because I’m working in that area and using an IDE against code that doesn’t compile is a hassle.

None

In many cases, that means just adding always-failing non-Linux stubs, which might be counterproductive for actually making CRI-O _work_ on those platforms.

So, let me know what you think; and closing this PR as something not desirable for the project would be completely fair, I wouldn’t mind. Or maybe some parts are useful and should be split into a smaller acceptable PR.

See individual commits, and commit messages, for details.

```release-note
None
```


haircommander:
without looking deeply, I'm gonna guess a lot of this is duplicated with https://github.com/cri-o/cri-o/pull/7214 I don't mind which direction we go but we probably should reduce duplication of work somehow :)

mtrmac:
Thanks for the pointer. #7214 intends to port _all_ of CRI-O, which is clearly more valuable.

I’ll keep this around for a bit, just because minimizing the deviation from current `main` helps me prepare the changes I actually want to make, and I need to maintain a separate branch either way — as a way to hedge against #7214 not being merged.

But the review efforts etc. should probably focus on #7214 .

dfr:
I do indeed have a working port of cri-o on FreeBSD (as well as Kubernetes). The #7214 PR is the first step and I have a stack of about 15-20 commits which I indent to turn into a smaller number of PRs after #7214 lands.

dfr:
Reading through the diffs, it looks like we are both going in the same direction and I believe that #7214 covers at least some of what you are doing here although probably not all.

mtrmac:
@dfr This PR contains just changes to make things build, and explicitly don’t care about them working (not even passing unit tests). That’s strictly inferior to your effort and I’m perfectly fine with throwing this PR away after #7214 merges.

(I keep rebasing it primarily because for my other work, I need an up-to-date version of this, separate from that actual work. It’s by no means intended to be a competitor with #7214.)

dfr:
That makes perfect sense, thanks for clarifying. In #7214, I'm also not trying to get anything to work - that comes in followup PRs.

dfr:
Thanks for this. Perhaps it would be a good idea to land this one soon, at least so that I can rebase my stack of FreeBSD diffs on top of it.

mtrmac:
@dfr I really don’t care how this is merged; if you’d rather integrate these changes (or at least the ones that make macOS build) into your outstanding work and merge them at a timeline that fits you better, that’s perfectly fine with me. Let me know.

mtrmac:
A CI cross build would be useful — but this PR is not yet sufficient for that. It only focuses on `internal/storage` and `server`, and the required prerequisites. And even in the affected packages, unit tests frequently fail.

TomSweeneyRedHat:
LGTM

saschagrunert (approved):

saschagrunert (approved):

saschagrunert:
/retest 

I assume we have to land something like https://github.com/cri-o/cri-o/pull/7463 then.

saschagrunert (approved):

PR #7306, opened 2023-09-14
Bump version to 1.29.0

saschagrunert:
#### What type of PR is this?

/kind cleanup

The version is old and wrong, so we bump it to the latest development version.

None
None

```release-note
None
```

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -20,7 +20,7 @@ import (
 )
 
 // Version is the version of the build.
-const Version = "1.27.0"
+const Version = "1.29.0"
 
 // Variables injected during build-time
 var (


internal/version/version.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 // Version is the version of the build.
-const Version = "1.27.0"
+const Version = "1.29.0-dev"
```
haircommander, line 23: we haven't had `-dev` in the version in a while, I am open to having it but what do other @cri-o/cri-o-maintainers think

saschagrunert: The dev was added as indicator for the deb/rpm package that we do not publish the final version. Not sure if it's still used anywhere else like the release notes.

saschagrunert: Hm, I guess we can also omit the dev :)

sohankunkerkar:
Sounds good!
/retest

PR #7309, opened 2023-09-14
Simplify, and optimize, ImageServer.ResolveNames

mtrmac:
#### What type of PR is this?

/kind cleanup

As a prerequisite of work on sigstore signature signing (earlier #7046), I need to tighten how CRI-O refers to images, to use types and values with strict semantics. That significantly revolves around the values consumed and returned by `ResolveNames`; I will want to completely replace that with code more explicit about the semantics/heuristics.

Let’s start with some baby steps:
- Eliminate the `ErrCannotParseImageID` special error, which almost all callers special-case exactly the same way. Instead, have `ResolveNames` do what the callers need.
- This allows optimizing out one c/storage lookup (= 2 FS lock hits), when starting a container from an already-pulled image or when deleting an image by Kubelet’s GC
- Also fix some other smaller bugs and remove unnecessary code.

None

```release-note
None
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -793,7 +793,7 @@ func (svc *imageService) ResolveNames(systemContext *types.SystemContext, imageN
 		}
 	}
 
-	// Disable short name alias mode. Will enable it once we settle on a shortname alias table.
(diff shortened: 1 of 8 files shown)


mtrmac:
Cc: @saschagrunert

saschagrunert (approved):
Thank you! LGTM

PR #7323, opened 2023-09-21
Disambiguate image identification in container creation

mtrmac:
#### What type of PR is this?

/kind cleanup

As one of prerequisites of per-namespace signature enforcement (the effort of #7046 ), it is _strictly necessary_ that exactly the image ID for which we have verified a signature is used to eventually create the container. This is a part of that, one of the necessary steps, starting fairly deep in the call stack. (With possibly more to come, as I actually come to understand the relevant parts of the full container lifecycle).

I also want to more strictly limit the semantics of values passed to `PullImage`, to make the problem space more manageable. This does not change anything about the value itself, but the simplification makes analysis a tiny bit more tractable.

Currently, `runtimeService.createContainerOrPodSandbox` handles both non-infra containers and the pause=infra container, with quite flexible code paths with somewhat unclear assumptions. Looking at the (single) callers, the non-infra containers and the pause=infra container paths, are, in fact, almost entirely distinct. So, pull the image lookup code into the callers of `createContainerOrPodSandbox`:
- When dealing with the pause image, we only have a name (with unclear format assumptions); use it to potentially pull, and to look up an image ID by name
- When dealing with non-infra containers, we already must have a resolved image ID; make sure to use only that value, and only potentially look up an image name.

This is not an end-to-end cleanup (notably the caller of runtimeService.CreateContainer` is _also_ providing a fallback name, so one of those can probably be eliminated); that can come over time. I’m trying to file smaller, more manageable, PRs.

---

The net effect of this is:
- Relevant to me: It is clearer, and better-documented how images are identified on these code paths. (I don't think this ends up providing a user-relevant image ID promise, yet; I need to extend the analysis into the callers etc.
- Relevant to most others: Some slight optimizations, because we do more lookups by IDs instead parsing and resolving names.
- Some only tangentially-related error handling bug fixes and optimizations.

None

```release-note
None
```

### DIFF
--- a/internal/storage/runtime.go
+++ b/internal/storage/runtime.go
@@ -141,12 +141,13 @@ func (metadata *RuntimeContainerMetadata) SetMountLabel(mountLabel string) {
 	metadata.MountLabel = mountLabel
 }
 
+// imageID must be provided and should refer to an image which existed at before calling this function (but that can change at any time).
+// The caller is also responsible for setting imageName (or "" if unknown)
 func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.SystemContext, podName, podID, imageName, imageAuthFile, imageID, containerName, containerID, metadataName, uid, namespace string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, isPauseImage, privileged bool) (ci ContainerInfo, retErr error) {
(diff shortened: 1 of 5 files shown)


saschagrunert (approved):
Thanks!
/lgtm

PR #7325, opened 2023-09-21
RFC: Use RuntimeContainerMetadata for passing container data

mtrmac:
#### What type of PR is this?

/kind cleanup

The parameters to `createContainerOrPodSandbox` pass, individually, all but 2 fields of `RuntimeContainerMetadata`. Using a struct, where the parameters are named, seems much more readable, right now - OTOH that might depend on how the function API vs. metadata JSON format evolves.

Should not change behavior.

None

Just because I was working in this area.

I’m not at all sure the metadata struct and the parameters are going to remain so similar; it’s quite possible this is a bad idea, I’m perfectly fine with this PR being rejected.

```release-note
None
```

### DIFF
--- a/internal/storage/runtime.go
+++ b/internal/storage/runtime.go
@@ -141,24 +141,30 @@ func (metadata *RuntimeContainerMetadata) SetMountLabel(mountLabel string) {
 	metadata.MountLabel = mountLabel
 }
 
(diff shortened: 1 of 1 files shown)


haircommander:
great idea!

/approve

LGTM, thanks

sohankunkerkar (approved):
/lgtm

saschagrunert (approved):
/ok-to-test

PR #5197, opened 2021-08-12
WIP: Add an endpoint for heap dumps

mrunalp:
This is to aid inspecting memory usage

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
This PR adds support to get crio heap dumps for inspecting memory usage.

#### Which issue(s) this PR fixes:
None

#### Does this PR introduce a user-facing change?

```release-note
Add support for gathering heap dumps
```

PR #7328, opened 2023-09-22
enable heap dumps

### LINKED ISSUE #7307: Add an endpoint for heap dumps
### What happened?

This issue aims to add an endpoint for heap dumps in CRI-O, which can be a valuable debugging and diagnostics feature. You can get help from the existing WIP [PR](https://github.com/cri-o/cri-o/pull/5197).

N/A

N/A

https://github.com/cri-o/cri-o/pull/5197

<details>

```console
$ crio --version
# paste output here
```

```console
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

<details>

</details>

hernandanielg:
#### What type of PR is this?

/kind feature

This will help on the memory debugging process

Fixes #7307

This PR will add a `--enable-heap-dump` flag to enable head dumps on an specific file
```release-note

```

### DIFF
--- a/internal/criocli/criocli.go
+++ b/internal/criocli/criocli.go
@@ -408,6 +408,9 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
 	if ctx.IsSet("disable-hostport-mapping") {
 		config.DisableHostPortMapping = ctx.Bool("disable-hostport-mapping")
 	}
+	if ctx.IsSet("enable-heap-dump") {
+		config.EnableHeapDump = ctx.Bool("enable-heap-dump")
+	}
 	return nil
 }
 
@@ -1163,6 +1166,12 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			Usage:   "If true, CRI-O would disable the hostport mapping.",
 			EnvVars: []string{"DISABLE_HOSTPORT_MAPPING"},
 			Value:   defConf.DisableHostPortMapping,
+		},
+		&cli.BoolFlag{
+			Name:    "enable-heap-dump",
+			Usage:   "If true, CRI-O generates a file at the specified location containing heap dump information.",
+			EnvVars: []string{"ENABLE_HEAP_DUMP"},
+			Value:   defConf.EnableHeapDump,
 		},
 	}
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -470,6 +470,10 @@ type RuntimeConfig struct {
 	// Option to disable hostport mapping in CRI-O
 	// Default value is 'false'
 	DisableHostPortMapping bool `toml:"disable_hostport_mapping"`
+
+	// EnableHeapDump option will make CRI-O to generate a file at the specified location
+	// containing heap dump information.
+	EnableHeapDump bool `toml:"enable_heap_dump"`
 }
 
 // ImageConfig represents the "crio.image" TOML config table.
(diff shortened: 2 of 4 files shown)


sohankunkerkar:
@hernandanielg Thanks for the PR.
You are on the right track. There are still a few places where you need to make changes. Did you see [this](https://github.com/cri-o/cri-o/pull/5197) PR to understand the actual logic? The idea is to dump the heap-related information if `--enable-heap-dump` is mentioned. Have you gone through [this](https://github.com/cri-o/cri-o/pull/6862) example PR to understand how to add a command line flag to the crio config?

sohankunkerkar:
@hernandanielg are you still interested in pursuing this PR?

hernandanielg:
yes! sorry I've been busy lately but will continue on this today 👋🏻

hernandanielg:
hey @sohankunkerkar 

As you may see I have added configuration option and reload function

I know it's just a matter to execute `debug.WriteHeapDump(f.Fd())` as done [here](https://github.com/cri-o/cri-o/pull/5197/files#diff-1f041e86ed8093e2372db3368ff66ed9f10df7ce06b4a5089cb817eae69cc48bR206)

I am trying to understand the code and find where and when it's going to be executed on runtime, maybe somewhere around [here](https://github.com/cri-o/cri-o/blob/6ab8bca8858220fef966477e44e7834c04d5bf53/server/runtime_status.go#L15)?

If you can guide me on this it would be highly appreciated!

Looking forward

sohankunkerkar:
@hernandanielg thanks for the changes. I think you need to add the flag details to `docs/` as well. As far as I know, you can keep the logic in `server/inspect.go` and call it only when that flag is enabled.

pkg/config/reload.go
```diff
@@ -176,6 +177,15 @@ func (c *Config) ReloadDecryptionKeyConfig(newConfig *Config) {
 	}
 }
 
+// ReloadEnableHeapDump updates the DecryptionKeysPath with the provided
+// `newConfig`.
+func (c *Config) ReloadEnableHeapDump(newConfig *Config) {
```
kwilczynski, line 182: @hernandanielg, wasn't this meant to be a command-line argument only?

haircommander:
@hernandanielg are you still working on this? if not, @shipra101  could take it over

hernandanielg:
Yes, I'm sorry for the delay, please gimme until tomorrow to update this PR

pkg/config/reload.go
sohankunkerkar: The rationale behind this logic is to allow the user to change the location of the heap dump by reloading the config. So, we have to change the config option to a string type, rather than a boolean.

haircommander: generally, if a malicious user has access to the crio socket, they can do harm in a lot of different ways. admins should ensure only trusted users can access nodes. For instance, a user could DOS CRI-O by having systemd shut it off. Since this and the goroutine profiles are both debugging options, I think it's safe to guess a user who has access can be trusted to not DOS the node's memory

hernandanielg: Sorry I was confused, I thought this was only a command line option to generate it as it could be done by sending `SIGUSR1` to CRI-O engine. what do you think? it should be an endpoint?

kwilczynski: @hernandanielg, we've had some internal conversations about how to move forward with this feature.

There is, of course, still an opportunity to complete this work here, so please be sure to bear with us a little bit. :smile: 

That said, I would like to bring everyone's attention to the following discussion:

- https://github.com/cri-o/cri-o/discussions/7615

hernandanielg: hey any news about this?

kwilczynski: @hernandanielg, sorry for the delay!

A decision on how to proceed has stalled, so it seems. Apologies!

PR #7329, opened 2023-09-25
skip setting the process args when restoring a container checkpoint

tu1h:
#### What type of PR is this?

/kind bug

When restoring a container checkpoint, and there is no command/arg setting both in image spec and kubernetes manifest, it will cause failure. But actually, restoring a checkpoint won't used the command/arg.

```release-note
none
```


adrianreber:
> /retest
> 
> makes sense to me, @adrianreber WDYT

Makes totally sense. But maybe it would make more sense to do the check directly in `internal/factory/container/container.go` in the `SpecSetProcessArgs()` function. The function has direct access to the information if it is a restored container or not.

haircommander:
/approve

LGTM, thanks!
/retest

sohankunkerkar (approved):
/lgtm

PR #7336, opened 2023-09-26
RFC: Turn ListImages(filter) into ImageStatus

mtrmac:
#### What type of PR is this?

/kind cleanup

Turn `ListImages(filter)` into `ImageStatus`. That "filter" was never really much of a filter, we always only got one image.
    
The current logic is, effectively, the only user which accepts that specific form of heuristically-interpreted syntax (because all other non-`PauseImage` users of internal/storage use `ResolveNames`), so removing that syntax would consolidate and clarify the semantics of internal/storage names.
    
Kubelet never calls `ListImages` with a filter AFAIK, so this should make no difference in practice - and it means we get to remove a chunk of code and tests.

None

Do note that this changes how `ListImagesRequest.Filter.Image.Image` is interpreted (to be the same as the same field in of the same struct in `ImageStatusRequest.Image.Image`), and that might be an incompatible change.

It’s quite possible that there are other non-Kubelet users of this API I’m unaware of.

I don’t _really_ need this change to happen for signing — it clarifies how CRI-O interprets names on input, and I think that’s a bit valuable in itself, as well as the code simplification it would eventually allow (ultimately removing `getRef` entirely). But this PR is mostly a side-effect of me clarifying my thinking about which values matter and how they are parsed

```release-note
If a filter is specified in the `ListImages` CRI method, it is now interpreted the same way image names in the `ImageStatus` methods are.
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -303,7 +303,7 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext) ([]Image
 	newImageCache := make(imageCache, len(images))
 	for i := range images {
 		image := &images[i]
-		ref, err := istorage.Transport.ParseStoreReference(svc.store, "@"+image.ID)
+		ref, err := istorage.Transport.NewStoreReference(svc.store, nil, image.ID)
 		if err != nil {
 			return nil, err
 		}
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -294,20 +294,14 @@ func (svc *imageService) buildImageResult(image *storage.Image, cacheItem imageC
 	}
 }
 
-func (svc *imageService) appendCachedResult(systemContext *types.SystemContext, ref types.ImageReference, image *storage.Image, results []ImageResult, newImageCache imageCache) ([]ImageResult, error) {
+func (svc *imageService) getImageCacheItem(systemContext *types.SystemContext, ref types.ImageReference, image *storage.Image) (imageCacheItem, error) {
 	svc.imageCacheLock.Lock()
 	cacheItem, ok := svc.imageCache[image.ID]
 	svc.imageCacheLock.Unlock()
-	if !ok {
-		var err error
-		cacheItem, err = svc.buildImageCacheItem(systemContext, ref)
-		if err != nil {
-			return results, err
-		}
+	if ok {
+		return cacheItem, nil
 	}
-	newImageCache[image.ID] = cacheItem
-
-	return append(results, svc.buildImageResult(image, cacheItem)), nil
+	return svc.buildImageCacheItem(systemContext, ref)
 }
 
 func (svc *imageService) ListImages(systemContext *types.SystemContext) ([]ImageResult, error) {
@@ -315,7 +309,7 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext) ([]Image
 	if err != nil {
 		return nil, err
 	}
-	var results []ImageResult
+	results := make([]ImageResult, 0, len(images))
 	newImageCache := make(imageCache, len(images))
 	for i := range images {
 		image := &images[i]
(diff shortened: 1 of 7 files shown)


mtrmac:
Cc: @saschagrunert

server/image_status.go
```diff
@@ -35,11 +34,12 @@ func (s *Server) ImageStatus(ctx context.Context, req *types.ImageStatusRequest)
 		return nil, err
 	}
 	var (
+		status   *pkgstorage.ImageResult
 		notfound bool
 		lastErr  error
 	)
 	for _, image := range images {
-		status, err := s.StorageImageServer().ImageStatus(s.config.SystemContext, image)
+		status_, err := s.StorageImageServer().ImageStatus(s.config.SystemContext, image)
```
haircommander: I would expect us to not need the additional variable. is it the case that `ImageStatus` could return both an error and a non-nil value?

mtrmac: The `status_` variable is removed in a later commit; it is used here because `status, err :=` would, IIRC, not update the `status` variable from outside the loop.

Alternatively, this could be

```go
var err error
status, err = …
```

In a later commit, both `status` and `err` become local to the loop body, and this inelegance disappears.

mtrmac:
(Unit tests fail, I’ll return to this next week.)

server/image_list.go
```diff
@@ -11,15 +11,31 @@ import (
...
+			if status == nil {
+				return &types.ListImagesResponse{
+					Images: []*types.Image{},
+				}, nil
+			}
+			return &types.ListImagesResponse{
+				Images: []*types.Image{
+					ConvertImage(status),
+				},
+			}, nil
```
haircommander, line 34: we could update ConvertImage() to return an empty list instead of a single element, and then compress all of this to 
```
			return &types.ListImagesResponse{
				Images: ConvertImage(status),
			}, nil
```

mtrmac: `ConvertImage` is also used below, in the primary `ListImages` cause, where there are almost always multiple images to collect in `Images`.

It might be cleaner to move all of the `filterImage.Image != "" ` code out of the main body, as a rare case into a separate function out of sight. OTOH “out of sight” might be a bad idea for a corner case — entirely up to CRI-O maintainers.

---

Or would you prefer
```go
if status != nil {
  resp.Images = append(resp.Images, ConvertImage(status))
}
return resp, nil
```
here? I don’t have a preference either way.

haircommander: > as a rare case into a separate function out of sight.

I am open to this. Part of me feels like the code smells a bit that filters never really worked, and we probably should have them working in case kubelet uses them. 

> ConvertImage is also used below, in the primary ListImages cause, where there are almost always multiple images to collect in Images.

Another option is ConvertImages, which could take the filtered status here and then it could be called below. I am not feeling very convinced in any direction her--all of these are suggestions for consideration

server/image_status.go
haircommander: WFM, I was going commit by commit

server/image_list.go
haircommander: I've thought of it, I think I would prefer:

```
if status != nil {
  resp.Images = append(resp.Images, ConvertImage(status))
}
return resp, nil
```

mtrmac: Done.

internal/storage/image.go
```diff
@@ -296,80 +296,49 @@ func (svc *imageService) buildImageResult(image *storage.Image, cacheItem imageC
...
-					for _, name := range image.Names {
-						if _, ok := ImageBeingPulled.Load(name); ok {
-							donePulling = false
-							break
-						}
-					}
-					if !donePulling {
-						continue
-					}
+				if os.IsNotExist(err) && imageIsBeingPulled(image) { // skip reporting errors if the images haven't finished pulling
```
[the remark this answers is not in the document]
mtrmac: Moved.

test/image.bats
```diff
@@ -110,9 +110,6 @@ function teardown() {
 	imageid=$(crictl images --quiet "$IMAGE:go")
 	[ "$imageid" != "" ]
 
-	output=$(crictl images --quiet @"$imageid")
```
mtrmac, line 113: Note all these changes: the `@`_id_ syntax is no longer recognized for a “filter”, with the minimal version of this PR.

Alternatively, the new code could explicitly recognize and support the `@`_id_ syntax.

(Or, going the other extreme, all of these test commands could use `crictl inspecti` instead.)

saschagrunert (approved):

Issue #7351, opened 2023-10-07
cannot remove pod created without a UID in YAML file


maxloge:
### What happened?

Hi,

I am using `crictl` to create a pod with the following yaml file:

`bug.yaml`
```YAML
---
metadata:
  name: bug
  namespace: hello
linux:
  security_context:
    namespace_options:
      network: 2

```

and get the following error when trying to remove the pod:

`getting sandbox status of pod "crictl rmp 4a790291df0b7": metadata.Name, metadata.Namespace or metadata.Uid is not in metadata "&PodSandboxMetadata{Name:bug,Uid:,Namespace:hello,Attempt:0,}"`

### What did you expect to happen?

I expect the pod to be removed even if I don't specify a UID in the metadata section of the yaml file.

### How can we reproduce it (as minimally and precisely as possible)?

```bash
root@vps:~/k8s# crictl runp bug.yaml
4a790291df0b7855ae8e734243d3e9fb794fcf3e9a6e466403b0a8a6611e24eb

root@vps:~/k8s# crictl pods
POD ID              CREATED             STATE               NAME                NAMESPACE           ATTEMPT             RUNTIME
4a790291df0b7       2 seconds ago       Ready               bug                 hello               0                   (default)

root@vps:~/k8s# crictl rmp 4a790291df0b7
getting sandbox status of pod "crictl rmp 4a790291df0b7": metadata.Name, metadata.Namespace or metadata.Uid is not in metadata "&PodSandboxMetadata{Name:bug,Uid:,Namespace:hello,Attempt:0,}"

root@vps:~/k8s# crictl pods
POD ID              CREATED             STATE               NAME                NAMESPACE           ATTEMPT             RUNTIME
4a790291df0b7       4 minutes ago       Ready               bug                 hello               0                   (default)
```

### Anything else we need to know?

crictl version:

<details>
root@vps:~/k8s# crictl --version
crictl version v1.28.0
</details>

I am not sure if it is an issue with crictl or cri-o so I have also reported the issue at cri-tools: https://github.com/kubernetes-sigs/cri-tools/issues/1273

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
crio version 1.28.1
Version:        1.28.1
GitCommit:      unknown
GitCommitDate:  unknown
GitTreeState:   clean
BuildDate:      2023-09-15T13:43:48Z
GoVersion:      go1.19
Compiler:       gc
Platform:       linux/amd64
Linkmode:       dynamic
[...]

PR #7353, opened 2023-10-09
Validate pod uid

saschagrunert:
#### What type of PR is this?

/kind bug

Pre-validate sandbox uid values before creation in the same way it's already done for the container name.

Fixes https://github.com/cri-o/cri-o/issues/7351

**_This is a revert of https://github.com/cri-o/cri-o/pull/3774_**

```release-note
Validate pod namespace and uid before creation.
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -305,7 +305,7 @@ func (c *container) SetConfig(cfg *types.ContainerConfig, sboxConfig *types.PodS
 	}
 
 	if cfg.Metadata.Name == "" {
-		return errors.New("name is nil")
+		return errors.New("name is empty")
 	}
 
 	if sboxConfig == nil {
--- a/internal/factory/sandbox/sandbox.go
+++ b/internal/factory/sandbox/sandbox.go
@@ -72,7 +72,7 @@ func (s *sandbox) SetConfig(config *types.PodSandboxConfig) error {
 	}
 
 	if config.Metadata.Name == "" {
-		return errors.New("PodSandboxConfig.Metadata.Name should not be empty")
+		return errors.New("metadata.Name should not be empty")
 	}
 
 	if config.Linux == nil {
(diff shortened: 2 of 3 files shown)


saschagrunert:
Ah that's a revert of https://github.com/cri-o/cri-o/pull/3774

saschagrunert:
/unhold

per https://github.com/kubernetes/kubernetes/issues/119692#issuecomment-1752694156

saschagrunert:
@cri-o/cri-o-maintainers PTAL

haircommander:
I'm unsure if we need this with https://github.com/kubernetes-sigs/cri-tools/pull/1274 . From what I can tell, the error on removal in https://github.com/cri-o/cri-o/issues/7351comes from CRI, not from CRI-O. There are users that don't specify metadata: like https://github.com/cri-o/cri-o/pull/3774 (or at least they did). doesn't the crictl change protect them?

haircommander:
I'm mostly anxious about breaking the coreos tests again and I don't remember where they are cc @miabbott

saschagrunert:
> I'm unsure if we need this with [kubernetes-sigs/cri-tools#1274](https://github.com/kubernetes-sigs/cri-tools/pull/1274) . From what I can tell, the error on removal in https://github.com/cri-o/cri-o/issues/7351comes from CRI, not from CRI-O. There are users that don't specify metadata: like #3618 (or at least they did). doesn't the crictl change protect them?

Yes, we can stick with the cri-tools solution for now. Other clients would have a chance to create pods without UID which then would cause another lock.

Feel free to close if we can take that risk.

travier:
I thin we removed / disabled our cri-o tests in RHCOS as they weren't very valuable for us.

travier:
Hum, never mind, we still have them: https://github.com/coreos/coreos-assembler/blob/main/mantle/kola/tests/crio/crio.go

Let me know if you need us to validate this change with our tests / update our tests before merging this one.

saschagrunert:
> Hum, never mind, we still have them: https://github.com/coreos/coreos-assembler/blob/main/mantle/kola/tests/crio/crio.go
> 
> Let me know if you need us to validate this change with our tests / update our tests before merging this one.

The tests would require an update with `crictl` v1.29.0 in any case, because we merged https://github.com/kubernetes-sigs/cri-tools/pull/1274

PR #7355, opened 2023-10-09
checkpoint: clean-up checkpoint dir after export

rst0git:
#### What type of PR is this?

/kind bug

The CRI-O implementation for container checkpointing is based on Podman. One of the differences between CRI-O and Podman is that with CRI-O, containers continue to run after checkpointing, and the checkpoint is exported as a tar archive file. However, after the checkpoint has been created, all checkpoint files remain in the userdata directory of the container. As a result, subsequent checkpoints include files from previous checkpoints. To fix this problem, this patch ensures that we clean up the checkpoint directory after the checkpoint files have been exported.

The original pull request introducing container checkpoint (https://github.com/cri-o/cri-o/pull/4199) contains more information about exporting of checkpoints.

```release-note
Clean up container checkpoint directory after export
```


rst0git:
@mrunalp PTAL

saschagrunert (approved):

PR #7358, opened 2023-10-10
RFC: Explicitly define a much stricter format for pause_image

mtrmac:
#### What type of PR is this?

/kind cleanup

The handling of `pause_image` in `CreatePodSandbox` currently heuristically accepts many different values, and most would lead to very surprising behavior:

- Don't accept storage IDs (they work locally but can't be pulled)
- Don't accept prefixes of storage IDs either (if they ever match locally, that would be VERY surprising)
- Don't accept short names (`dir:foo` could locally search for `docker.io/library/dir:foo`, but then try to "pull" from local filesystem at `./foo`)
- Don't accept `transport:`_..._ values

Instead, just accept ordinary registry references, qualified with a host name (and, _undocumented_, use the traditional `docker.io/[library]` fallback semantics).

Also parse the `pause_image` option already at daemon start, and every reload.

I don’t expect this to affect production workloads — I think most of the accepted syntaxes are just not practical, but I am steeling myself to be surprised.

In theory, values starting with any of `docker://`, `docker-archive:`, `docker-daemon:`, `atomic:`, `containers-storage:` _might_ currently work; some in case where the referenced image is already available locally, some could also be pulled.

Note what is currently _not_ supported: short names (`CreatePodSandbox` does not do alias / search registry resolution), and `DefaultTransport` is not involved. Either way the values are currently very different from what e.g. `PullImage` accepts.

None

- I don’t _strictly_ need this for the signing work — it’s a byproduct of clarifying my thoughts; and it would allow (along with #7336 ) completely removing various code paths in `internal/storage`, ultimately removing _all_ heuristics from there. But I also think restricting the accepted values seems a bit valuable either way, just for reliability.
- It might make sense to completely reject short names (= names without a host name) instead of defaulting to `docker.io`. I am a bit worried that this could really break someone.
- It seems somewhat attractive to make `ImageConfig.PauseImage` itself a well-typed value, instead of a string and parsing the value later, but that seems like a bit more invasive change to me.

```release-note
The semantics of the `pause_image` config option has been clarified, and invalid values are rejected at daemon start and reload.
```

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -415,7 +415,9 @@ CRI-O reads its configured registries defaults from the system wide containers-r
   The path to a file like /var/lib/kubelet/config.json holding credentials necessary for pulling images from secure registries.
 
 **pause_image**="registry.k8s.io/pause:3.9"
-  The image used to instantiate infra containers. This option supports live configuration reload.
+  The on-registry image used to instantiate infra containers.
+  The value should start with a registry host name.
+  This option supports live configuration reload.
 
 **pause_image_auth_file**=""
  The path to a file like /var/lib/kubelet/config.json holding credentials specific to pulling the pause_image from above. This option supports live configuration reload.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -16,6 +16,7 @@ import (
 	"github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
 	"github.com/containers/common/pkg/hooks"
 	conmonconfig "github.com/containers/conmon/runner/config"
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/podman/v4/pkg/rootless"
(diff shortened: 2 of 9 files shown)


mtrmac:
Cc: @saschagrunert

docs/crio.conf.5.md
```diff
@@ -415,7 +415,9 @@ CRI-O reads its configured registries defaults from the system wide containers-r
   The path to a file like /var/lib/kubelet/config.json holding credentials necessary for pulling images from secure registries.
 
 **pause_image**="registry.k8s.io/pause:3.9"
-  The image used to instantiate infra containers. This option supports live configuration reload.
+  The on-regiostry image image used to instantiate infra containers.
```
saschagrunert, line 418:
```suggestion
  The on-registry image used to instantiate infra containers.
```

pkg/config/config.go
```diff
@@ -487,8 +488,9 @@ type ImageConfig struct {
 	// containing credentials necessary for pulling images from secure
 	// registries.
 	GlobalAuthFile string `toml:"global_auth_file"`
-	// PauseImage is the name of an image which we use to instantiate infra
-	// containers.
+	// PauseImage is the name of an image on a registry which we use to instantiate infra
+	// containers. It should start with a registry host name.
+	// Most users should prefer ParsePauseImage().
```
saschagrunert, line 493: I think we can just mention that `PauseImage` gets validated on config validation.

saschagrunert (commented):
Just two nits, otherwise LGTM. Thanks!

docs/crio.conf.5.md
mtrmac: Oops. Fixed, thanks.

saschagrunert (approved):

PR #7373, opened 2023-10-12
OCPBUGS-20492: highperfhooks: apply cgroup setting to crun cgroup

haircommander:

```release-note
fix a bug where the cgroup crun configures is different than that CRI-O sets load balancing/cpu quota on
```

### DIFF
--- a/test/cgroups.bats
+++ b/test/cgroups.bats
@@ -219,4 +219,7 @@ EOF
 	cgroup_file="cpu.cfs_quota_us"
 	[[ $(cat "$CTR_CGROUP"/"$cgroup_file") == "-1" ]]
 	[[ $(cat "$POD_CGROUP"/"$cgroup_file") == "-1" ]]
+	if [[ "$CONTAINER_DEFAULT_RUNTIME" == "crun" ]]; then
+		[[ $(cat "$CTR_CGROUP"/container/"$cgroup_file") == "-1" ]]
+	fi
 }
--- a/internal/config/cgmgr/cgmgr_linux.go
+++ b/internal/config/cgmgr/cgmgr_linux.go
@@ -193,7 +193,7 @@ func createSandboxCgroup(sbParent, containerCgroup string) error {
 		if path == "" {
 			return fmt.Errorf("failed to find cpuset for newly created cgroup")
 		}
-		if err := os.Mkdir(path, 0o755); err != nil && !os.IsNotExist(err) {
+		if err := os.MkdirAll(path, 0o755); err != nil && !os.IsNotExist(err) {
 			return fmt.Errorf("failed to create cpuset for newly created cgroup: %w", err)
 		}
 		if err := libctr.WriteFile(path, "cpuset.sched_load_balance", "0"); err != nil {
--- a/internal/runtimehandlerhooks/default_cpu_load_balance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/default_cpu_load_balance_hooks_linux.go
@@ -2,12 +2,9 @@ package runtimehandlerhooks
 
 import (
 	"context"
-	"fmt"
 
-	"github.com/cri-o/cri-o/internal/config/node"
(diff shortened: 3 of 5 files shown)


test/cgroups.bats
```diff
@@ -219,4 +219,7 @@ EOF
 	cgroup_file="cpu.cfs_quota_us"
 	[[ $(cat "$CTR_CGROUP"/"$cgroup_file") == "-1" ]]
 	[[ $(cat "$POD_CGROUP"/"$cgroup_file") == "-1" ]]
+	if [[ "$CONTAINER_DEFAULT_RUNTIME" == "crun" ]]; then
+		[[ $(cat "$CTR_CGROUP"/container"$cgroup_file") == "-1" ]]
```
kolyshkin, line 223: Looks like a slash is missing after `container`

haircommander: oops, and that revealed I was testing it wrong!

internal/runtimehandlerhooks/default_cpu_load_balance_hooks.go
```diff
@@ -34,17 +33,16 @@ func (*DefaultCPULoadBalanceHooks) PostStop(ctx context.Context, c *oci.Containe
...
 		return err
 	}
-	mgr, err := libctrManager(containerCgroup, containerCgroupParent, systemd)
-	if err != nil {
-		return err
-	}
-	path := mgr.Path("cpuset")
-	if path == "" {
-		return fmt.Errorf("failed to find cpuset for newly created cgroup")
+	for i := len(containerManagers) - 1; i >= 0; i-- {
```
rphillips, line 40: nonblocking: This block could be refactored to a function, since the same code is line high_performance_hooks.go:225

haircommander: done!

internal/runtimehandlerhooks/high_performance_hooks.go
```diff
@@ -204,34 +204,24 @@ func annotationValueDeprecationWarning(annotation string) string {
...
-		return fmt.Errorf("find container %s CPUs", c.ID())
-	}
-
+func disableCPULoadBalancing(parentDir string, c *oci.Container) error {
 	if node.CgroupIsV2() {
 		return fmt.Errorf("disabling CPU load balancing on cgroupv2 not yet supported")
 	}
 
-	pid, err := c.Pid()
+	_, containerManagers, err := libctrManagersForPodAndContainerCgroup(c, parentDir)
```
kolyshkin, line 212: Instantiating two cgroup managers for the sake of getting the path to cpuset cgroup seems like an overkill.

I would do it differently -- get the path to cgroup* , and in case its last component is "container", write to the parent cgroup as well.

* as for getting the path to cgroup, you can do it in the same way like it's done before this patch, OR maybe you can actually cache the cgroup path in Container struct somehow (similar to how you do it for Sandbox and CgroupParent).

kolyshkin (changes requested):
I don't like using two instances of cgroup manager to merely get cgroup path.

internal/runtimehandlerhooks/high_performance_hooks.go
MarSik: I agree the managers are heavy and complicate things. However hardcoding a name based detection is ugly and not future proof.

haircommander: > you can do it in the same way like it's done before this patch

this won't work for the sched_load_balance case because we need to disable after the container has stopped too (in case it has load balancing while  it was running but we need to reallocate the cpus it used to use later), which means the proc entry is gone

haircommander:
/hold

this may not be working correctly

haircommander:
/hold cancel

this also needs https://github.com/containers/crun/pull/1332 but the crio pieces are ready

haircommander:
@cri-o/cri-o-maintainers PTAL

Issue #7035, opened 2023-06-09
OOM detection is racy when using systemd cgroup driver


kolyshkin:
### What happened?

Filing this to summarise current findings, and track the work needed to fix this.

Cri-o relies on conmon (or conmon-rs) to detect that an OOM kill has happened. Alas, it does not work reliably in case systemd cgroup driver is used, because what happens is:

(1) container is being oom-killed;
(2) systemd receives cgroup event and removes container cgroup;
(3) conmon receives cgroup event and reads cgroup's `memory.events` file.

There's a race between (2) and (3). If (2) happens first, (3) fails:

> Failed to open cgroups file: /sys/fs/cgroup/pod_123.slice/pod_123-456.slice/crio-b0410a6682627b30dc5758d59d89d60c67450e995902965a13c7398c74094419.scope/memory.events

Also, in some cases it looks like conmon exits before the OOM event is handled.

kolyshkin:
Technically, this bug belongs to conmon (and conmon-rs), but maybe there's no way to fix it in there, because of the above race.

The solution, when using systemd cgroup driver, would be to ask systemd for unit status from cri-o, since systemd also detects that the unit was OOM-killed. The problems are:
 1. Unless `OOMPolicy` is set to `kill` (or `memory.oom.group` to `1`), the oom-kill event won't result in failed unit.
 2. Since crio sets `CollectMode: inactive-or-failed` property, the unit will be removed on failure.

Alas, runc delete does not call resetFailedUnit but this is easy to fix! opencontainers/runc#3888
(I'm not sure if something like this has to be added to crun as well -- if the unit created by crun also enters the failed state then yes; cc @giuseppe).

Once that's in, backported, and a runc release is made, we can modify cri-o to
- skip setting CollectMode;
- check with systemd if the container has been oom-killed (by checking unit status, not by reading the logs like the quoted code does);
- use new runc/crun to clean the unit status (I assume "runc delete" is called down the line). Or, cri-o can call `reset-failed` itself after getting its status, but I think it's better be done in runc.

kolyshkin:
The alternative to the above is to use journalctl --unit `crio-$CTID.scope` to see if there were any logs related to OOM killing. This is ugly.

saschagrunert:
@kolyshkin removing the assignee, feel free to reassign if you plan to work on this.

PR #7374, opened 2023-10-12
drop inactive-or-failed property

haircommander:
it was originally added because failed scopes weren't being cleaned up. This is the incorrect solution. Instead, the oci runtime should reset a failed unit after the container is done.

With inactive-or-failed property, systemd races with conmon reading the oom file of a container.

Note: this means that CRI-O should run with crun 1.9 and runc 1.1.8 or later. This should be enforced in the package.

#### What type of PR is this?

/kind deprecation
fixes https://github.com/cri-o/cri-o/issues/7035

```release-note
No longer set inactive-or-failed collect mode property, which fixes a race with container that OOM kill. As a result, users should upgrade to runc 1.1.8 or crun 1.9
```

### DIFF
--- a/internal/config/node/node_linux.go
+++ b/internal/config/node/node_linux.go
@@ -52,13 +52,6 @@ func ValidateConfig() error {
 			fatal:     false,
 		},
 		{
-			name:      "systemd CollectMode",
-			init:      SystemdHasCollectMode,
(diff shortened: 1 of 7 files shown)


mrunalp:
Can we link to run/crun changes for the cleanup?

sohankunkerkar (approved):
/lgtm

PR #7375, opened 2023-10-12
Use strongly encapsulated types with strict semantics for RemoveImage

mtrmac:
#### What type of PR is this?

/kind cleanup

Split `ImageServer.ResolveNames` into two well-typed parsers
    
We will push the heuristics to the callers to make the heuristics explicit, very loud to readers, and most importantly to give callers well-typed data instead of ambiguous strings.

This introduces `storage.StorageImageID` and `storage.RegistryImageReference` types, with the intent that users dealing with `internal/storage` should only use these strongly-typed values, not strings, for image identification/referencing.

None

See individual commit messages for details. Overall, this should not actually change behavior of `RemoveImage`.

This is the beginning of the payoff of the recent API/semantics PRs. I have two similar directly related items queued to follow:
- Migrate `ImageServer.ImageStatus` to accept well-typed input (`StorageImageID` or `RegistryImageReference`)
- Migrate `PrepareImage`/`PullImage` to accept `RegistryImageReference`

and that will eliminate (or push to callers) all the heuristic “does it parse this way, that way, another way?” code paths in `internal/storage`, allowing the pull code to precisely reason about the names and to manipulate them with confidence.

This PR is filed separately (leaving the rules-violating `ResolveNames` in place) to get the review of the API types and general approach quickly, to surface concerns, if any, earlier.

Cc: @saschagrunert 

```release-note
None
```

### DIFF
--- a/server/image_remove.go
+++ b/server/image_remove.go
@@ -31,10 +31,20 @@ func (s *Server) removeImage(ctx context.Context, imageRef string) error {
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
-	images, err := s.StorageImageServer().ResolveNames(s.config.SystemContext, imageRef)
-	if err != nil {
-		return err
+	var images []string
+	if id := s.StorageImageServer().HeuristicallyTryResolvingStringAsIDPrefix(imageRef); id != nil {
+		images = []string{id.IDStringForOutOfProcessConsumptionOnly()} // This violates rules of references.StorageImageID, but it will be removed soon.
+	} else {
+		potentialMatches, err := s.StorageImageServer().CandidatesForPotentiallyShortImageName(s.config.SystemContext, imageRef)
+		if err != nil {
+			return err
+		}
+		for _, ref := range potentialMatches {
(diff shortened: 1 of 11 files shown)


server/image_remove.go
```diff
@@ -31,14 +31,27 @@ func (s *Server) removeImage(ctx context.Context, imageRef string) error {
...
 
-	images, err := s.StorageImageServer().ResolveNames(s.config.SystemContext, imageRef)
+	// FIXME: The CRI API definition says
+	//      This call is idempotent, and must not return an error if the image has
+	//      already been removed.
+	// and this code doesn’t seem to conform to that.
+
+	// Actually Kubelet is only ever calling this with full image IDs.
+	// So we don't really need to accept ID prefixes nor short names;
+	// or is there another user?!
```
saschagrunert, line 41: cri-tools does an `ImageStatus` before removing the specified image, which leads into a ResolveNames call. So I guess we're fine.

saschagrunert (approved):
Thank you!

PR #7376, opened 2023-10-13
Switch to registries.conf.d for static binary bundle

saschagrunert:
#### What type of PR is this?

/kind feature

We now use a v2 config as well as the `registries.conf.d` location for the configuration file.

Refers to https://github.com/cri-o/packaging/pull/17
None

```release-note
Switch to `registries.conf.d` v2 format for static binary bundle.
```


haircommander (approved):
/lgtm

Should we move this stuff into the packaging repo so it's closer to the code that calls it?

saschagrunert:
> /lgtm
> 
> Should we move this stuff into the packaging repo so it's closer to the code that calls it?

That could be a long term effort, yes. Where would we split the responsibilities between the repos? For exaxmple, we build the binaries in cri-o/cri-o and just push them to the bucket? cri-o/packaging then uses the binaries and assembles a bundle, pushes it to the bucket and builds packages from it?

haircommander:
> ?

yeah this makes sense to me

contrib/bundle/crio-registries.conf
```diff
@@ -0,0 +1 @@
+unqualified-search-registries = ["docker.io", "quay.io"]
```
mtrmac, line 1: A drive-by comment on a PR which, I think, doesn’t change anything in this area:

Shipping two registries, both of which allow registration by ~untrusted users, in the default configuration, is a security risk: https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name

saschagrunert: Thanks! We added this for convenience, but can reconsider once I'm done with https://github.com/cri-o/cri-o/issues/7380

PR #7392, opened 2023-10-16
Use strongly encapsulated types with strict semantics for PullImage

mtrmac:
#### What type of PR is this?

/kind cleanup

Use `RegistryImageReference` throughout `PullImage` and the pull code in `internal/storage`. Various now-unreachable fallback paths and corner cases were removed, so the pull code now has clear semantics of image names, enforced by the Go type system.

Note that this, strictly speaking, changes the accepted image syntax for `PullImage`, to reject (prefixes of) local storage IDs _if the user has set up `default_transport` to a non-default value_. AFAICT that could only ever have worked in transports where the input would then be interpreted as a local path against CRI-O’s current directory, something that is clearly unwanted. See the "Behavior change…" commit.

It might make sense to deprecate `default_transport` entirely (it can only reasonably be set to `docker://`, and to `atomic` which probably works but where OpenShift has long removed/broken that protocol on the server-side), but that’s out of scope for me right now.

None

This is the 3rd part out of 3 in #7375.

The heuristic-accepting `ImageStatus` and `ResolveNames` are gone. This should hopefully be the last cleanup for pull path (not yet the container creation path) prior to actually adding signature enforcement.

```release-note
The `PullImage` operation no longer accepts image ID values/prefixes, relevant in extremely rare configurations with `default_transport` set to a transport which accesses local files.
```

I’m not sure this is even worth worrying most users about, but then… if any users did do that, they would probably want to know.


mtrmac:
Cc: @saschagrunert

saschagrunert (approved):

PR #7393, opened 2023-10-17
Use typed and explicitly named names for image fields in containers

mtrmac:
#### What type of PR is this?

/kind cleanup

As a ~side-effect of an audit of how image-related data is produced, stored on disk, and used, rename the fields typically named (`image`, `imageName` and `imageRef`) to (`userRequestedHeuristicImage`, `someNameOfTheContainersImage` and `imageID`), and use strong types for the latter two, throughout the codebase.

As a side effect of this analysis, this allows us to remove one on-disk image lookup on the container creation path.

None

See individual commit messages for details; notably this introduces new failure cases on invalid image names in storage, that might not be the best choice.

I’m stopping (after all those changes…) at the point where I’m confident that once we have an `imageID`, it is the only value that matters for running containers (assuming the checkpoint creation was created with a version that sets `RootfsImageRef`); and where it is ~explicit whether various APIs return the `userRequestedHeuristicImage` or the `someNameOfTheContainersImage` value.

AFAICT Kubelet never uses the image name values (only `ImageRef` = image ID), they “only”, show up in Pod statuses, so that’s where relevance to image signing goes.

Arguably the PR should do _something_ about the `userRequestedHeuristicImage` / `someNameOfTheContainersImage` values: maybe only ever use one of them, quite possibly make some effort to make `someNameOfTheContainersImage` more related to users’ input. If you think that’s required, please say so (and preferably choose a design).

Cc: @saschagrunert 

```release-note
None
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -73,8 +73,8 @@ type Container interface {
 	// DisableFips returns whether the container should disable fips mode
 	DisableFips() bool
 
-	// Image returns the image specified in the container spec, or an error
-	Image() (string, error)
+	// UserRequestedImage returns the image specified in the container spec, or an error
+	UserRequestedImage() (string, error)
 
 	// ReadOnly returns whether the rootfs should be readonly
 	// it takes a bool as to whether crio was configured to
@@ -172,7 +172,7 @@ func (c *container) SpecAddAnnotations(ctx context.Context, sb *sandbox.Sandbox,
 	created := time.Now()
 	labels := c.Config().Labels
 
-	image, err := c.Image()
+	userRequestedImage, err := c.UserRequestedImage()
 	if err != nil {
 		return err
 	}
@@ -211,7 +211,7 @@ func (c *container) SpecAddAnnotations(ctx context.Context, sb *sandbox.Sandbox,
 		}
 	}
 
-	c.spec.AddAnnotation(annotations.Image, image)
+	c.spec.AddAnnotation(annotations.Image, userRequestedImage)
 	imageName := ""
 	if imageResult.SomeNameOfThisImage != nil {
 		imageName = imageResult.SomeNameOfThisImage.StringForOutOfProcessConsumptionOnly()
@@ -464,8 +464,8 @@ func (c *container) DisableFips() bool {
 	return false
 }
 
-// Image returns the image specified in the container spec, or an error
-func (c *container) Image() (string, error) {
+// UserRequestedImage returns the image specified in the container spec, or an error
+func (c *container) UserRequestedImage() (string, error) {
 	imageSpec := c.config.Image
 	if imageSpec == nil {
 		return "", errors.New("CreateContainerRequest.ContainerConfig.Image is nil")
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -213,7 +213,7 @@ func (c *container) SpecAddAnnotations(ctx context.Context, sb *sandbox.Sandbox,
 
 	c.spec.AddAnnotation(annotations.Image, image)
 	c.spec.AddAnnotation(annotations.ImageName, imageResult.Name)
-	c.spec.AddAnnotation(annotations.ImageRef, imageResult.ID)
+	c.spec.AddAnnotation(annotations.ImageRef, imageResult.ID.IDStringForOutOfProcessConsumptionOnly())
 	c.spec.AddAnnotation(annotations.Name, c.Name())
 	c.spec.AddAnnotation(annotations.ContainerID, c.ID())
 	c.spec.AddAnnotation(annotations.SandboxID, sb.ID())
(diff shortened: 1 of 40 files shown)


server/container_start.go
```diff
@@ -30,13 +30,19 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
 		// into the restore code.
 		log.Debugf(ctx, "Restoring container %q", req.ContainerId)
 
+		// Using someNameOfTheContainersImage as a file looks quite suspicious. Luckily(?), AFAICS, as of 2023-10, the TargetFile
+		// field is not used by ContainerRestore at all, so it doesn’t matter.
```
mtrmac, line 34: I don’t understand what is really going on here…

server/mock_helpers_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+type mockSequence struct {
+	first, last *gomock.Call // may be both nil (= the default value of mockSequence) to mean empty sequence
+}
+
+// like gomock.inOrder, but can be nested
+func inOrder(calls ...interface{}) mockSequence {
```
mtrmac, line 15: This is a copy of code from `internal/storage`; is there a good place to share it?

server/server.go
```diff
@@ -212,7 +212,12 @@ func (s *Server) restore(ctx context.Context) []string {
 			pods[containers[i].ID] = &metadata
 		} else {
 			podContainers[containers[i].ID] = &metadata
-			containersAndTheirImages[containers[i].ID] = containers[i].ImageID
+			imageID, err := storage.ParseStorageImageIDFromOutOfProcessData(containers[i].ImageID)
+			if err != nil {
+				log.Warnf(ctx, "Error parsing image ID %q of container %q: %v, ignoring", containers[i].ImageID, containers[i].ID, err)
+				continue
+			}
+			containersAndTheirImages[containers[i].ID] = imageID
```
mtrmac, line 220: Looking at coverage data… apparently the existing `"should succeed with container restore"` test exercises this using 3 containers with `ID` and `ImageID` both empty. I’m… getting tired of fixing mocks that just don’t represent real-world operation.

saschagrunert (approved):

server/mock_helpers_test.go
saschagrunert: How about creating a new package in `internal` for this?

internal/oci/container.go
```diff
@@ -55,33 +56,34 @@ type Container struct {
...
-	pidns              nsmgr.Namespace
-	restore            bool
-	restoreArchive     string
-	restoreIsOCIImage  bool
-	resources          *types.ContainerResources
-	runtimePath        string // runtime path for a given platform
+	dir        string
+	stopSignal string
+	// Some name of the image imageID, if any; it may have NO RELATIONSHIP to the users’ requested image name.
+	someNameOfTheContainersImage *references.RegistryImageReference
```
haircommander, line 62: personally, I think continuing to refer to it as `imageName` and commenting the caveats is sufficient.

internal/factory/container/container.go
```diff
@@ -73,8 +73,8 @@ type Container interface {
 	// DisableFips returns whether the container should disable fips mode
 	DisableFips() bool
 
-	// Image returns the image specified in the container spec, or an error
-	Image() (string, error)
+	// UserRequestedHeuristicImage returns the image specified in the container spec, or an error
```
haircommander, line 76: similarly here. While i can see a case for being clearer, I think the variable name is very verbose and could be shortened

haircommander: maybe `UserRequestedImage` ?

mtrmac: My thinking was that scary concepts should have scary names, to discourage future readers from adding any more uses… and even perhaps motivate (… someone else …) to replace these values entirely.

But really 90% of the value of the exercise for me is in having found/tracked those values, so I’m happy with any names the maintainers prefer.

server/container_start.go
mtrmac: … anyway, I have outright removed that field value. It clearly doesn’t do anything, so let’s not even pretend to.

server/mock_helpers_test.go
mtrmac: Moved to `internal/mockutils`.

internal/oci/container.go
mtrmac: Reverted (but kept the explicit type).

internal/factory/container/container.go
mtrmac: Changed to `UserRequestedImage`.

server/server.go
mtrmac: I have changed that to at least use 3 proper, and distinct, values. Not that it exercises the wipe logic all that much…

saschagrunert (approved):
/hold

for @haircommander discussions

LGTM from my side

PR #7396, opened 2023-10-17
static binaries: install crio-registries.conf from correct location

haircommander:
#### What type of PR is this?

/kind ci

```release-note
none
```


haircommander:
hm this is not the solution

haircommander:
while those paths look incorrect (as in the cri-o repo they're located in contrib/bundle) the bundle script moves them from there to contrib/ so changing the paths makes the issue worse.

PR #7405, opened 2023-10-19
docs: install `libseccomp-devel` instead of `libseccomp`

Tal-or:
#### What type of PR is this?

/kind documentation

For compiling crio  the `libseccomp` development version is needed otherwise `pkg-config` failed to find the library
and failed with
```
go build github.com/seccomp/libseccomp-golang:
# pkg-config --cflags  -- libseccomp
Package libseccomp was not found in the pkg-config search path.
Perhaps you should add the directory containing `libseccomp.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libseccomp', required by 'virtual:world', not found
pkg-config: exit status 1
```

None

```release-note
none
```

### DIFF
--- a/install.md
+++ b/install.md
@@ -322,7 +322,7 @@ yum install -y \
   libassuan \
   libassuan-devel \
   libgpg-error \
-  libseccomp \
+  libseccomp-devel \
   libselinux \
   pkgconf-pkg-config \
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

PR #7435, opened 2023-10-25
Make `ImageStatusByName` race-free

mtrmac:
#### What type of PR is this?

/kind bug

Currently, `ImageStatusByName` looks up the incoming name three times (once in `GetStoreImage`, once in `NewImage`, once in `NewImageSource`. If the name uses a tag and not a digest, that can result reading data from three different images, making the resulting `ImageResult` inconsistent.

A fix for this requires a new `storage.ResolveReference` function in c/image, being added in https://github.com/containers/image/pull/2056 .

This PR is currently a proof-of-concept before merging that c/image feature, depending on an unmerged c/image PR.

None

It seems possible to me that we don’t actually need this for signature verification (as long as we get _one_ `StorageID`, we have something to work with, and the other fields of `ImageResult` might not matter), but it is just simpler for me fix `ImageStatusByName` than to worry about the details of that situation.

And I think CRI-O benefits form this fix either way.

Cc: @saschagrunert 

```release-note
None
```


mtrmac:
Note to self:

>   [FAILED] failed to get image status: rpc error: code = Unknown desc = reference "[overlay@/tmp/tmp.BdnzWSMRdG/crio+/tmp/tmp.BdnzWSMRdG/crio-run]registry.k8s.io/e2e-test-images/busybox:1.29-2" does not resolve to an image ID: identifier is not an image

Those are real enough, `ResolveReference` returns a different error than `GetStoreImage` used to, and `ImageStatus` looks for that error.

mtrmac:
Prerequisites are merged, this is now ready for review.

PR #7436, opened 2023-10-26
RFC: Add image ref test

saschagrunert:
#### What type of PR is this?

/kind bug

_The tests added in this PR fails right now to outline the problem._

When running a container by referencing a repo digest as image, then the container status should not resolve into a tag when locally available. Is this the correct intention?

Found in https://github.com/kubernetes/kubernetes/pull/121531

PTAL @mtrmac 

```release-note
TBD
```

### DIFF
--- a/test/image_ref.bats
+++ b/test/image_ref.bats
@@ -1,0 +1,35 @@
+#!/usr/bin/env bats
+# vim:set ft=bash :
+
+load helpers
+
+function setup() {
+	setup_test
+}
+
+function teardown() {
+	cleanup_test
+}
+
+IMAGE_REF=quay.io/crio/alpine:3.9
+IMAGE_DIGEST=quay.io/crio/alpine@sha256:414e0518bb9228d35e4cd5165567fb91d26c6a214e9c95899e1e056fcd349011
+
+function run_using_digest() {
+	jq '.image.image = "'"$IMAGE_DIGEST"'"' "$TESTDATA/container_config.json" > "$TESTDIR/ctr.json"
+	CTR_ID=$(crictl run "$TESTDIR"/ctr.json "$TESTDATA"/sandbox_config.json)
+
+	IMAGE=$(crictl inspect "$CTR_ID" | jq -r .status.image.image)
+	[[ "$IMAGE" == "$IMAGE_DIGEST" ]]
+}
+
+@test "should reference image digest if specified" {
+	start_crio
+	crictl pull "$IMAGE_DIGEST"
+	run_using_digest
+}
+
+@test "should reference image digest if specified and tag is referencing digest" {
+	start_crio
+	crictl pull "$IMAGE_REF"
+	run_using_digest
+}


test/image_ref.bats
```diff
@@ -0,0 +1,35 @@
...
+@test "should reference image digest if specified" {
+    start_crio
+    crictl pull "$IMAGE_DIGEST"
+    run_using_digest
+}
+
+@test "should reference image digest if specified and tag is referencing digest" {
+    start_crio
+    crictl pull "$IMAGE_REF"
+    run_using_digest
```
saschagrunert, line 34: Fails in line `22`, because we have a tag referencing the repo digest locally.

test/image_ref.bats
```diff
@@ -0,0 +1,35 @@
...
+    CTR_ID=$(crictl run "$TESTDIR"/ctr.json "$TESTDATA"/sandbox_config.json)
+
+    IMAGE=$(crictl inspect "$CTR_ID" | jq -r .status.image.image)
+    [[ "$IMAGE" == "$IMAGE_DIGEST" ]]
+}
+
+@test "should reference image digest if specified" {
+    start_crio
+    crictl pull "$IMAGE_DIGEST"
+    run_using_digest
```
saschagrunert, line 28: Succeeds

test/image_ref.bats
mtrmac: Yes, because it uses `oci.Container.ImageName` = `ImageResult.SomeNameOfThisImage`. Which name of the many available is chosen is not related to the image specified in the container spec at all; the code prefers to return _some_ tagged name over _some_ digested name (I… have a vague impression the goal was to prefer digests, but the code has been unchanged since 2017 IIRC). And, also, I think (I didn’t check in detail), because the “does this image exist locally already?” path resolves the _repo_`@`_digest_ reference without actually _adding_ the digested reference to `storage.Image.Names`; so `SomeNameOfThisImage` couldn’t possibly contain the digested reference.

---

FWIW, as a principle of the thing, “pull using one spec, create container using a different one” is _not_ a flow I am considering in the signature enforcement work; in the sense that if signatures are enforced, they will be enforced correctly, but I’m at this point not trying to ensure that that combination of pull+container is guaranteed to _work_. It… probably will?

---

The general situation of `oci.Container` carrying two different names, neither of which is really _correct_ (per the original explicit naming in https://github.com/cri-o/cri-o/pull/7393), can probably be substantially improved but it’s not something I’m immediately interested in tackling (it is irrelevant to signature enforcement, and it would involve understanding the Kubelet and higher-level ecosystems to see which value they want to see and why).

One thing that can be done immediately is to set `oci.Container.ImageName` to the “resolved” version of the input, in https://github.com/cri-o/cri-o/blob/7d55719d06da58074d639a5e666e22c5d2924d67/server/container_create_linux.go#L228 . (I’m not saying what to do on the `id := HeuristicallyTryResolvingStringAsIDPrefix…; id != nil` path.)

Eventually, Real Soon Now (but I’m not working for the rest of this week), I will get to `PullImage(tag)` returning the _correct_ resolved digested reference; and possibly `CreateContainer` also choosing a specific digested reference, and that might change more of the situation.

mtrmac: Actually, this is simpler with the new Kubelet versions which provide `UserRequestedImage`. In that case the code can always use that value as is (after checking that the supplied image can be actually found under that name, i.e. it was not untagged or the tag was not moved to a fresher image by a pull?), even if the input is an image ID.

Relying on `UserRequestedImage` would nicely avoid the whole `SomeNameOfThisImage` business on these code paths — at the cost of one extra c/storage lookup (multiple disk hits and a two short-term locks, IIRC).

(If the goal were to  resolve an input tagged reference into a digested reference, that’s still not possible: the information to do that correctly isn’t currently available in c/storage. But this PR doesn’t ask to do that.)

PR #7408, opened 2023-10-19
server: allow Bidirectional mounts that contain storage root

haircommander:
previously we would never allow a container to get anything other than HostToContainer for something that contains the storage root. This breaks situations where the container wishes to update mount points within the container and see them propegated to the host.

Instead, we should only override this if the mount is Private.

#### What type of PR is this?

/kind bug
#### Which issue(s) this PR fixes:
fixes https://github.com/cri-o/cri-o/issues/7308

#### Does this PR introduce a user-facing change?

```release-note
fix a bug where CRI-O would override a Bidirectional mount in favor of a HostToContainer if the mount contained the host's container storage
```

PR #7457, opened 2023-10-31
[release-1.27] NO-ISSUE: server: allow Bidirectional mounts that contain storage root

kwilczynski:
This is a manual cherry-pick of https://github.com/cri-o/cri-o/pull/7076 and https://github.com/cri-o/cri-o/pull/7408

/assign kwilczynski

```release-note
- Set mount type HostToContainer for mounts that include container storage root
- Fix a bug where CRI-O would override a Bidirectional mount in favor of a HostToContainer if the mount contained the host's container storage
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -979,7 +979,7 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
 			log.Warnf(ctx, "Configuration specifies mounting host root to the container root.  This is dangerous (especially with privileged containers) and should be avoided.")
 		}
 
-		if isSubDirectoryOf(storageRoot, m.HostPath) {
+		if isSubDirectoryOf(storageRoot, m.HostPath) && m.Propagation == types.MountPropagation_PROPAGATION_PRIVATE {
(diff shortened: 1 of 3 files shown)


saschagrunert (approved):

haircommander:
hm why did we backport this? was someone asking for this behavior? I worry about having introduced more issues (even past the one that is now fixed) and I'm not sure we should introduce such instability into this branch

kwilczynski:
Hi @haircommander,

> hm why did we backport this? was someone asking for this behavior? I worry about having introduced more issues (even past the one that is now fixed) and I'm not sure we should introduce such instability into this branch

What do you mean when you say "instability"? Do you foresee issues?

I am asking as we backported this to release 1.28, and we are about to cut a 1.28.2 release.

Similarly, with 1.27.2 - it was with anticipation of this new release and the fact that we made a backport to a different release branch I decided to backport this fix. Both are our active releases, and this fixes issues with some CSI that people use.

If we weren't entirely sure about this, why backport to 1.28 and not just leave it on the main branch?

Nevertheless, we can revert before you cut the 1.27.2 release. :smile:

haircommander:
FWIW: the initial commit here is kind of a feature. It fixes an issue when users mount /host and aren't able to see updates to /var/lib/containers. that said, I think it's fine. let's leave it for now

PR #7460, opened 2023-10-31
Fix inconsistency in restore image name resolution

mtrmac:
#### What type of PR is this?

/kind bug

The code was calling `checkIfCheckpointOCIImage`, which resolves short names into all candidates, and then using `ParseStoreReference`, which interprets short names as `docker.io/[library]`.
    
Instead, have `checkIfCheckpointOCIImage` return the resolved storage ID, use that, and avoid the redundant lookup entirely.

None

Initially, fixing this bug seemed easier than building the test mocks to work with the previous version of the code in #7302. Other than that, this has no relationship to signature verification.

Alternatively, the input → image ID resolution could use some other rules; either not resolving short names and hard-coding `docker.io`, or maybe even conceivably requiring an image ID as input. I don’t know what’s the right thing to do here; using the same rules as for regular `CreateContainer` paths seems reasonable to me as an outsider.

`checkIfCheckpointOCIImage` is still being called twice in quick succession from `CreateContainer` (and if the input were tagged, it could potentially resolve to different images?); that could be cleaned up further, but I’ll leave that to others.

Cc: @saschagrunert 

```release-note
None
```

(This could possibly advertise that short-name restore images work, but I have no idea how restore is used and whether that could ever happen. Even if it could, it shouldn’t…)

### DIFF
--- a/internal/lib/restore.go
+++ b/internal/lib/restore.go
@@ -58,16 +58,18 @@ func (c *ContainerServer) ContainerRestore(
 		return "", err
 	}
 
-	if ctr.RestoreArchive() != "" {
-		if ctr.RestoreIsOCIImage() {
-			log.Debugf(ctx, "Restoring from %v", ctr.RestoreArchive())
-			imageMountPoint, err := c.StorageImageServer().GetStore().MountImage(ctr.RestoreArchive(), nil, "")
+	if ctr.RestoreArchivePath() != "" || ctr.RestoreStorageImageID() != nil {
+		if ctr.RestoreStorageImageID() != nil {
+			log.Debugf(ctx, "Restoring from %v", ctr.RestoreStorageImageID())
+			// This is not out-of-process, but it is at least out of the CRI-O codebase; containers/storage uses raw strings.
+			imageMountPoint, err := c.StorageImageServer().GetStore().MountImage(ctr.RestoreStorageImageID().IDStringForOutOfProcessConsumptionOnly(), nil, "")
 			if err != nil {
 				return "", err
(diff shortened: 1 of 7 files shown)


saschagrunert (approved):
Thanks!

mtrmac:
(Updated with just a `s/SetRrestoreStorageImageID/SetRestoreStorageImageID/` typo fix.)

saschagrunert (approved):

mtrmac:
@saschagrunert for the past few days, I see Prow tests all failing with
```
WARNING:  Python 3.6.x is no longer officially supported by the Google Cloud CLI
and may not function correctly. Please use Python version 3.8 and up.
…
  File "/usr/local/gcloud/google-cloud-sdk/lib/googlecloudsdk/command_lib/util/apis/yaml_command_schema_util.py", line 248
    elif builtin_type := BUILTIN_TYPES.get(arg_type, None):
                       ^
SyntaxError: invalid syntax
```

That doesn’t look like something I could have caused

saschagrunert:
@mtrmac I know, we're trying to find out what's wrong here.

saschagrunert:
Closing the loop: https://github.com/openshift/release/pull/45228

mtrmac:
Almost a month later, looking at recently merged PRs, CI still seems to be broken?

haircommander:
timely question-- @sohankunkerkar has just updated it and it should be fixed now
/retest

saschagrunert:
The registry seems throwing `50x` errors :thinking: 
/retest

PR #7461, opened 2023-10-31
Prepare image pulling for per-namespace signature enforcement

mtrmac:
#### What type of PR is this?

/kind bug

This is mostly refactoring:
- Split the 180-line `Server.pullImage` function into 3 major components, making it more manageable and simplifying the control flow
- Modify the `copyImageChild` subprocess interface to be able to return values. This will be necessary to return the correct resolved image reference after pulling an image.
- Then share _everything_ in the implementation of `PullImage` between the in-process and `copyImageChild` variants, so that we can consistently enforce signature policies in the future without duplicating it in two implementations (which have _already_ diverged by now).

+ some smaller cleanups and bug fixes along the way.

None

I didn’t even smoke-test the “experimental” `copyImageChild` code path.

Cc: @saschagrunert . Almost there now.

```release-note
None
```

### DIFF
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -246,7 +246,7 @@ func (s *Server) pullImageCandidate(ctx context.Context, sourceCtx *imageTypes.S
 	defer close(progress) // nolint:gocritic
 	go metricsFromProgressGoroutine(ctx, progress, remoteCandidateName, tmpImg)
 
-	_, err = s.StorageImageServer().PullImage(s.config.SystemContext, remoteCandidateName, &storage.ImageCopyOptions{
+	_, err = s.StorageImageServer().PullImage(remoteCandidateName, &storage.ImageCopyOptions{
 		SourceCtx:        sourceCtx,
 		DestinationCtx:   s.config.SystemContext,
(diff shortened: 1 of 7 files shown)


server/image_pull.go
```diff
@@ -14,11 +14,13 @@ import (
 
 	"github.com/containers/image/v5/signature"
 	imageTypes "github.com/containers/image/v5/types"
+	encconfig "github.com/containers/ocicrypt/config"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/storage"
 	"github.com/cri-o/cri-o/server/metrics"
 	"github.com/cri-o/cri-o/utils"
 	"github.com/docker/distribution/registry/api/errcode"
+	. "github.com/onsi/ginkgo/v2"
```
saschagrunert, line 23: Unfortunately the linter is not happy with this, do we even need ginkgo as part of this non-test file?

mtrmac: I’m sorry, that’s just me trying to diagnose test failures. I’ll mark this as draft until the tests pass.

saschagrunert (approved):
/lgtm

PR #7478, opened 2023-11-14
WIP: [1.26] linked logs: drop intermediate logs directory

haircommander:

```release-note

```

### DIFF
--- a/internal/linklogs/link_logs.go
+++ b/internal/linklogs/link_logs.go
@@ -31,17 +31,13 @@ func MountPodLogs(ctx context.Context, kubePodUID, emptyDirVolName, namespace, k
 	if _, err := os.Stat(emptyDirLoggingVolumePath); err != nil {
 		return fmt.Errorf("failed to find %v: %w", emptyDirLoggingVolumePath, err)
 	}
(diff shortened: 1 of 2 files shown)


oybed:
@haircommander thx for this PR!

Tested this change and overall it looks good since the `logs` sub-directory is no longer there.

However, one additional difference I am spotting based on customer work is that the symlink at the "top level" is missing the `.log` extension. See screenshot below from testing this PR. The customer implementation in this case would have a log file with the `.log` extension at the top level like the one shown below - i.e.: not a symlink and a file name ending in `.log`. I don't necessarily think the symlink is going to be an issue, but I know they are using match strings like "*.log" - hence having the `.log` extension is important. 
```
> ls -l /logging-volume/
 :
-rw-r-----. 1 root root 11 Dec 12 03:52 38e0cad1609615bcefd9313d311f3b3f0d4044c1ab7d38bb9d352f3fe0fd188d.log
 : 
```

Can the missing `.log` extension be addressed by this PR as well? Once done, I'd be happy to give it another go in the customer environment to verify that the symlink isn't going to be an issue. 

image

haircommander:
updated @oybed

oybed:
Thank you @haircommander. I have validated that the latest is working in the customer environment end-to-end.

Issue #7479, opened 2023-11-15
Crio installation/update failure on x64 Ubuntu 20.04 


seantywork:
### What happened?

I'm frequently installing/updating crio as a Kubernetes OCI runtime engine on x64 Ubuntu 20.04

The issue has been recurring from roughly 8:00 AM November 14, 2023 with no automatic or random success observed.

I think the issue below 

- https://github.com/containers/podman/issues/20690

is somewhat reflecting the same thing I'm struggling with now.

Below is the installation log

```bash
+ sudo apt-get update
Hit:1 http://kr.archive.ubuntu.com/ubuntu focal InRelease
Hit:2 https://baltocdn.com/helm/stable/debian all InRelease
Get:3 http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/1.27/xUbuntu_20.04  InRelease [1632 B]
Hit:4 http://kr.archive.ubuntu.com/ubuntu focal-updates InRelease
Hit:5 http://kr.archive.ubuntu.com/ubuntu focal-backports InRelease
Get:6 https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  InRelease [1642 B]
Hit:7 http://kr.archive.ubuntu.com/ubuntu focal-security InRelease
Fetched 3274 B in 2s (2167 B/s)
Reading package lists... Done
+ sudo apt-get install cri-o cri-o-runc -y
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  conmon containers-common
Suggested packages:
  containernetworking-plugins
The following NEW packages will be installed:
  conmon containers-common cri-o cri-o-runc
0 upgraded, 4 newly installed, 0 to remove and 83 not upgraded.
Need to get 3631 kB/32.4 MB of archives.
After this operation, 133 MB of additional disk space will be used.
Err:1 https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  cri-o-runc 1.0.1~2
  Redirection from https to 'http://mirror.aardsoft.fi/opensuse/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/amd64/cri-o-runc_1.0.1~2_amd64.deb' is forbidden [IP: 195.135.223.226 443]
[...]

PR #7480, opened 2023-11-15
apt installationn redirection block fix

seantywork:
#### What type of PR is this?

/kind documentation

During the installation/update process, there is an issue where apt is blocking https -> http mirror redirection.

This PR simply counts that fact into the installation document, so for others to have less hassle over it. 

Fixes https://github.com/cri-o/cri-o/issues/7479

None

```release-note

```


install.md
```diff
@@ -200,6 +200,30 @@ mkdir -p /usr/share/keyrings
...
+apt-get install cri-o cri-o-runc
+```
+
+If apt doesn't allow redirection from 'https' -> 'http' mirror then modify the protocol as follows
+
+```shell
+
+# if something similar to the below log happens...
+
+Redirection from https to 'http://mirror.blablah.com/opensuse/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:versionblah/osblah' is forbidden
```
saschagrunert, line 213: I'd say we change `versionblah` and `osblah` to something without `blah` :)

saschagrunert (commented):
/ok-to-test

install.md
seantywork: great idea, I'll fix them

saschagrunert:
/hold

I assume that's an issue with the OBS repositories and should be fixed there rather than downgrading to HTTP.

seantywork:
yep, that's what I've just done. I think you're definitely right. I downgraded the protocol because the checks were failing due to the same issue :) Let's wait out then?

fix

kwilczynski:
@seantywork @saschagrunert, the issues should have been resolved now.

Do we still need this documentation update? Personally, I would prefer people to only ever use HTTPS and report issues if this does not work.

PR #7256, opened 2023-08-28
internal/lib/stats: leverage libcontainer API for gathering pod/container stats and pod sandbox metrics

sohankunkerkar:
Related to https://issues.redhat.com/browse/OCPNODE-1008

#### What type of PR is this?

/kind api-change
/kind other

#### Does this PR introduce a user-facing change?

```release-note
Leverage libcontainer API for gathering pod/container stats and pod sandbox metrics
```

PR #7481, opened 2023-11-15
Replace `containers/common` with` libcontainer/cgroups` for stats collection

Mo-Fatah:
on top of #7256 
#### What type of PR is this?

/kind dependency-change

```release-note

```


internal/lib/stats/stats_server.go
```diff
@@ -130,6 +147,174 @@ func (ss *StatsServer) updateSandbox(sb *sandbox.Sandbox) *types.PodSandboxStats
...
+			updatedIncludedMetrics = append(updatedIncludedMetrics, val)
+		}
+	}
+	// Create a cgroup manager for the container if it doesn't exist
+	if ss.ctrCgMgrs[c.ID()] == nil {
+		if err := ss.updateCgMgrs(sb, c); err != nil {
+			logrus.Errorf("Unable to Create a cgroup manager for the container %s: %v", c.ID(), err)
+			return nil
+		}
+	}
```
haircommander, line 173: since we're on the same container here, we can initialize the cgmgr here and then use it throughout, instead of duplicating the map lookup

PR #7493, opened 2023-11-20
Warn users when trying to use `noNewPrivileges` on privileged or `CAP_SYS_ADMIN`

saschagrunert:
#### What type of PR is this?

/kind cleanup

Combining both settings does not make much sense because the option `noNewPrivileges` will do nothing for containers using privileged or `CAPCAP_SYS_ADMIN`. We now warn users if that's the case.

None
None

```release-note
Warn users on container creation when trying to use `noNewPrivileges` on privileged or `CAP_SYS_ADMIN`.
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -477,6 +477,20 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 				return nil, err
 			}
 		}
+
+		if securityContext.NoNewPrivs {
+			const sysAdminCap = "CAP_SYS_ADMIN"
+			for _, cap := range specgen.Config.Process.Capabilities.Bounding {
+				if cap == sysAdminCap {
+					log.Warnf(ctx, "Setting `noNewPrivileges` flag has no effect because container has %s capability", sysAdminCap)
+				}
+			}
+
+			if ctr.Privileged() {
+				log.Warnf(ctx, "Setting `noNewPrivileges` flag has no effect because container is privileged")
+			}
+		}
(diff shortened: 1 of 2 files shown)


server/container_create_linux.go
```diff
@@ -477,6 +477,20 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 				return nil, err
 			}
 		}
+
+		if securityContext.NoNewPrivs {
+			const sysAdminCap = "CAP_SYS_ADMIN"
+			for _, cap := range specgen.Config.Process.Capabilities.Bounding {
+				if cap == sysAdminCap {
+					log.Warnf(ctx, "Setting `noNewPrivileges` flag has no effect because container has %s capability", sysAdminCap)
```
kwilczynski, line 485: Just a nit, so probably nothing to do here.

Would the flag name, and I assume we mean the one passed down from Kubernetes, be `noNewPrivs`?  I think the flag name is `no_new_privs` internally, as defined in the API, per the Protocol Buffer schema.

saschagrunert: Yeah I'm leaning towards using the runtime spec flag: https://github.com/opencontainers/runtime-spec/blob/99235413e05f23548c28dae02d72532e092d87a2/specs-go/config.go#L86

kwilczynski: @saschagrunert, ahh! Makes sense. :+1:

kwilczynski:
@saschagrunert, this should be good to go, I believe.

PR #7485, opened 2023-11-16
runtime handler hooks: add support for cgroupv2 cpu load balancing

haircommander:
#### What type of PR is this?

/kind feature
#### Does this PR introduce a user-facing change?

```release-note
add support for cpu load balancing annotation for cgroupv2 
```

PR #7502, opened 2023-11-21
CNF-8808: perfruntime: add shared cpus to container's cgroup

Tal-or:
#### What type of PR is this?

/kind feature

Given an annotation CRI-O appends additional CPUs (named shared CPUs)
into the container's cgroups.

The operation is done via a performance runtime hook.

This will allow the container to run light-weight tasks
on the shared cpus, while all the other task that
requires full isolation will keep running on the guaranteed CPUs.

It also injects env variables in to the container
in order to distinguish between the isolated and shared CPUs.

None
Related to:

* https://github.com/openshift/kubernetes/pull/1795
* https://github.com/openshift/kubernetes/pull/1799
* https://github.com/openshift/cluster-node-tuning-operator/pull/853

Depends on:
* https://github.com/cri-o/cri-o/pull/7485 

EP: https://github.com/openshift/enhancements/commit/76ea48a88bca15ef6157aacf90b8544a763e6075

```release-noteAdd support for shared CPUs in high performance hooks, which allows users to specify exclusive and shared CPUs for high performance containers
```


Tal-or:
Depends on: https://github.com/cri-o/cri-o/pull/7485

internal/config/node/cgroups_linux.go
```diff
@@ -26,6 +27,38 @@ var (
 	cgroupIsV2Err error
 )
 
+// CgroupHierarchy is a cgroup version agnostic struct that allows
```
haircommander, line 30: I don't think this is best here. This package has a pretty particular use (for initializing variables that CRI-O needs to react to on the node). If it's not being used anywhere else, I'd prefer it in runtime_handler_hooks

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -76,6 +89,15 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 		}
 	}
 
+	if requestedSharedCPUs(s.Annotations(), c.CRIContainer().GetMetadata().GetName()) {
+		if h.sharedCPUs == "" {
+			return fmt.Errorf("shared CPUs were requested for container %q but none are defined", c.Name())
+		}
+		if err = setSharedCPUs(ctx, c, podManager, containerManagers, h.sharedCPUs, shouldCPULoadBalancingBeDisabled(s.Annotations())); err != nil {
```
haircommander, line 96: we should do `if err :=` here to not shadow the variable,  I think it's more idiomatic

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -940,3 +968,114 @@ func convertAnnotationToLatency(annotation string) (maxLatency string, err error
...
+	if err != nil {
+		return fmt.Errorf("failed to parse container %q cpus: %w", c.Name(), err)
+	}
+	sharedCPUSet, err := cpuset.Parse(sharedCPUs)
+	if err != nil {
+		return fmt.Errorf("failed to parse shared cpus: %w", err)
+	}
+	if sharedCPUSet.IsEmpty() {
+		return fmt.Errorf("shared CPU set is empty")
+	}
```
haircommander, line 987: isn't this already being checked above with `h.sharedCPUs == ""` check?

haircommander:
heya @Tal-or! I have structural comment that was tedious enough I implemented it. Basically, as it is now, on v2 we set the isolated CPUs then immediately unset them and inject the new cgroup. Instead, I think it would be better to do it in one feld swoop. I had to break up the code a bit to do this:
 https://github.com/haircommander/cri-o/tree/lbv2_mixed_cpus
I did not adapt the unit tests, nor did I test it. WDYT? I think this results in cleaner code and less duplicated work, at the expense of making some of the logic a bit less immediately apparent. I also incorporated some of my other comments and cleaned up some things. Feel free to adapt my work if you think it'd be better a different way.  Iwould say the only blocking issue I have is the duplicated work in v2, so the other pieces you can integrate as you please

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -48,12 +46,27 @@ const (
...
+
+const (
+	cpusetPartition      = "cpuset.cpus.partition"
+	cpusetExclusive      = "cpuset.cpus.exclusive"
+	cgroupSubTreeControl = "cgroup.subtree_control"
+	cpusetCpus           = "cpuset.cpus"
+	cpusetCpusExclusive  = "cpuset.cpus.exclusive"
+)
+
+const (
```
mrunalp, line 60: Yes, can you merge with the const block above?

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -938,3 +987,132 @@ func convertAnnotationToLatency(annotation string) (maxLatency string, err error
...
+		return nil, fmt.Errorf("failed to parse container %q cpus: %w", c.Name(), err)
+	}
+	if sharedCPUs == "" {
+		return nil, fmt.Errorf("shared CPUs were requested for container %q but none are defined", c.Name())
+	}
+	sharedCPUSet, err := cpuset.Parse(sharedCPUs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse shared cpus: %w", err)
+	}
+	if err = containerManagers[len(containerManagers)-1].Set(&configs.Resources{
```
rphillips, line 1007: Need to guard here for the len(containerManagers) == 0 case.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -938,3 +987,132 @@ func convertAnnotationToLatency(annotation string) (maxLatency string, err error
...
+		return err
+	}
+
+	// container level operations
+	ctrCPUSet := isolatedCPUSet.Union(sharedCPUSet)
+	ctrQuota, err := calculateMaximalQuota(&ctrCPUSet, *(cpuSpec.Period))
+	if err != nil {
+		return fmt.Errorf("failed to calculate container %s quota: %w", c.ID(), err)
+	}
+	return containerManagers[len(containerManagers)-1].Set(&configs.Resources{
```
rphillips, line 1088: ditto for the -1 guard.

haircommander:
needs https://github.com/cri-o/cri-o/pull/7566

haircommander:
/override ci/prow/e2e-gcp-ovn

aws passed

PR #7514, opened 2023-11-27
Use `libctr/cgroups` for stats collection

Mo-Fatah:
#### What type of PR is this?

/kind dependency-change

Use `libcontainer/cgroups` to collect sandbox/container stats instead of `containers/common`. This will be useful once we move on from cAdvisor. 

```release-note
None
```

### DIFF
--- a/internal/config/cgmgr/cgmgr_linux.go
+++ b/internal/config/cgmgr/cgmgr_linux.go
@@ -62,8 +62,8 @@ type CgroupManager interface {
 	// ContainerCgroupManager takes the cgroup parent, and container ID.
 	// It returns the raw libcontainer cgroup manager for that container.
 	ContainerCgroupManager(sbParent, containerID string) (libctr.Manager, error)
-	// RemoveCtrCgManager removes the cgroup manager for the container
-	RemoveCtrCgManager(containerID string)
+	// RemoveContainerCgManager removes the cgroup manager for the container
+	RemoveContainerCgManager(containerID string)
 	// SandboxCgroupPath takes the sandbox parent, and sandbox ID. It
 	// returns the cgroup parent, cgroup path, and error. For systemd cgroups,
 	// it also checks there is enough memory in the given cgroup
@@ -74,8 +74,8 @@ type CgroupManager interface {
 	// SandboxCgroupManager takes the cgroup parent, and sandbox ID.
 	// It returns the raw libcontainer cgroup manager for that sandbox.
 	SandboxCgroupManager(sbParent, sbID string) (libctr.Manager, error)
-	// RemoveSbCgroupManager removes the cgroup manager for the sandbox
-	RemoveSbCgManager(sbID string)
+	// RemoveSandboxCgroupManager removes the cgroup manager for the sandbox
+	RemoveSandboxCgManager(sbID string)
 	// MoveConmonToCgroup takes the container ID, cgroup parent, conmon's cgroup (from the config), conmon's PID, and some customized resources
 	// It attempts to move conmon to the correct cgroup, and set the resources for that cgroup.
 	// It returns the cgroupfs parent that conmon was put into
--- a/internal/config/cgmgr/cgmgr_linux.go
+++ b/internal/config/cgmgr/cgmgr_linux.go
@@ -59,6 +59,11 @@ type CgroupManager interface {
 	// PopulateContainerCgroupStats fills the stats object with information from the cgroup found
 	// given a cgroup parent and container ID.
 	PopulateContainerCgroupStats(sbParent, containerID string, stats *types.ContainerStats) error
+	// ContainerCgroupManager takes the cgroup parent, and container ID.
+	// It returns the raw libcontainer cgroup manager for that container.
+	ContainerCgroupManager(sbParent, containerID string) (libctr.Manager, error)
+	// RemoveCtrCgManager removes the cgroup manager for the container
+	RemoveCtrCgManager(containerID string)
 	// SandboxCgroupPath takes the sandbox parent, and sandbox ID. It
 	// returns the cgroup parent, cgroup path, and error. For systemd cgroups,
 	// it also checks there is enough memory in the given cgroup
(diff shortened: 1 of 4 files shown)


internal/config/cgmgr/cgroupfs_linux.go
```diff
@@ -90,25 +90,25 @@ func (m *CgroupfsManager) ContainerCgroupStats(sbParent, containerID string) (cg
...
 	return cgstats, nil
 }
 
 // GetCtrCgroupManager takes the cgroup parent, and container ID.
 // It returns the raw libcontainer cgroup manager for that container.
-func (m *CgroupfsManager) GetCtrCgroupManager(sbParent, containerID string) (cgMgr libctrCg.Manager, err error) {
+func (m *CgroupfsManager) GetCtrCgroupManager(sbParent, containerID string) (libctrCg.Manager, error) {
 	if !node.CgroupIsV2() {
-		if cgMgr = m.v1CtrCgMgr[containerID]; cgMgr != nil {
+		if cgMgr := m.v1CtrCgMgr[containerID]; cgMgr != nil {
```
haircommander, line 103: so this will create a new entry in the map, you need to do `cgMgr, ok := ..; ok && cgMgr != nil {` to not add an entry

haircommander, line 103: I think we need some locking here

internal/config/cgmgr/cgroupfs_linux.go
```diff
@@ -90,25 +90,25 @@ func (m *CgroupfsManager) ContainerCgroupStats(sbParent, containerID string) (cg
...
+		OtherMemStats: generateOtherMemoryStats(&libCtrStats.MemoryStats),
 		SystemNano:    time.Now().UnixNano(),
 	}
 	return cgstats, nil
 }
 
 // GetCtrCgroupManager takes the cgroup parent, and container ID.
 // It returns the raw libcontainer cgroup manager for that container.
-func (m *CgroupfsManager) GetCtrCgroupManager(sbParent, containerID string) (cgMgr libctrCg.Manager, err error) {
+func (m *CgroupfsManager) GetCtrCgroupManager(sbParent, containerID string) (libctrCg.Manager, error) {
```
haircommander, line 101: Naming suggestion: golang sugggests not prepending `Get` to getters. Also ,the other functions in this package spell out Container, so maybe `ContainerCgroupManager` ?

internal/config/cgmgr/cgroupfs_linux.go
haircommander: oh shoot sorry I commented this on the wrong PR, lemme move this convo over there

internal/config/cgmgr/cgmgr_linux.go
```diff
@@ -59,13 +59,29 @@ type CgroupManager interface {
 	// PopulateContainerCgroupStats fills the stats object with information from the cgroup found
 	// given a cgroup parent and container ID.
 	PopulateContainerCgroupStats(sbParent, containerID string, stats *types.ContainerStats) error
+	// ContainerCgroupStats returns a stats object with information from the cgroup found
+	// given a cgroup parent and container ID.
+	ContainerCgroupStats(sbParent, containerID string) (*CgroupStats, error)
```
haircommander, line 64: is this called anywhere? I think I'd prefer to not have this endpoint for now. The caller of ContainerCgroupManager can implement it, no?

internal/config/cgmgr/cgroupfs_linux.go
```diff
@@ -66,6 +75,65 @@ func (m *CgroupfsManager) ContainerCgroupAbsolutePath(sbParent, containerID stri
...
+	if !node.CgroupIsV2() {
+		// cache only cgroup v1 managers
+		m.v1CtrCgMgr[containerID] = cgMgr
+	}
+	return cgMgr, nil
+}
+
+// RemoveCtrCgManager removes and destroys the cgroup manager for the container
+func (m *CgroupfsManager) RemoveCtrCgManager(sbParent, containerID string) {
+	cgMgr, err := m.GetCtrCgroupManager(sbParent, containerID)
```
haircommander, line 124: couldn't we just directly access the map? doesn't GetCtr... create the cgroup manager if it doesn't exist?

haircommander:
handful of comments, also before this merges the commits should be squashed

internal/config/cgmgr/cgroupfs_linux.go
Mo-Fatah: To destroy the cgroup, we should get it from the map (if we use cgroup v1), or create the cgroup object on the fly for cgroup v2. `GetCtrCgroupManager` internally does this.

haircommander: FYI I don't think we should destroy the cgruop in these functions, that's done by the runtime when the container is cleaned up. We should just remove the entry from the map, if present

internal/config/cgmgr/cgmgr_linux.go
Mo-Fatah: it isn't called anywhere yet, not in this PR. Later I am planning of removing `PopulateContainerCgroupStats` and similar functions, and use `ContainerCgroupStats` and `SandboxCgroupStats`. Both  `ContainerCgroupStats` and `SandboxCgroupStats` should be called directly from the stats server's loop.

Mo-Fatah:
@haircommander can you have another look now ?

internal/config/cgmgr/cgmgr_linux.go
```diff
@@ -59,13 +59,23 @@ type CgroupManager interface {
 	// PopulateContainerCgroupStats fills the stats object with information from the cgroup found
 	// given a cgroup parent and container ID.
 	PopulateContainerCgroupStats(sbParent, containerID string, stats *types.ContainerStats) error
+	// ContainerCgroupManager takes the cgroup parent, and container ID.
+	// It returns the raw libcontainer cgroup manager for that container.
+	ContainerCgroupManager(sbParent, containerID string) (libctr.Manager, error)
+	// RemoveCtrCgManager removes the cgroup manager for the container
+	RemoveCtrCgManager(containerID string)
```
haircommander, line 66: I think we should keep the names consistent. RemoveContainer* and RemoveSandbox* below

saschagrunert (approved):

saschagrunert:
/hold 

for @haircommander final look

haircommander:
/hold cancel

well done, thank you!

PR #7551, opened 2023-12-05
pinns: improve write_mapping_file errors

giuseppe:
include the mappings that we tried to write in the error message. Also fix the error type since the uid mapping failure was reported as gid mapping.

#### What type of PR is this?

/kind cleanup

```release-note
None
```

### DIFF
--- a/pinns/src/pinns.c
+++ b/pinns/src/pinns.c
@@ -227,10 +227,10 @@ int main(int argc, char **argv) {
 
       /* Write user mappings */
       if (gid_mapping && write_mapping_file(pid, gid_mapping, true) < 0)
-        pexit("Cannot write gid mappings");
+        pexitf("Cannot write gid mappings: %s", gid_mapping);
 
       if (uid_mapping && write_mapping_file(pid, uid_mapping, false) < 0)
-        pexit("Cannot write gid mappings");
+        pexitf("Cannot write uid mappings: %s", uid_mapping);
 
       /* Notify that the mappings were written.  */
       if (TEMP_FAILURE_RETRY(write(p[0], "0", 1)) < 0)


saschagrunert (approved):
Thanks!

pinns/src/pinns.c
```diff
@@ -227,10 +227,10 @@ int main(int argc, char **argv) {
 
       /* Write user mappings */
       if (gid_mapping && write_mapping_file(pid, gid_mapping, true) < 0)
-        pexit("Cannot write gid mappings");
+        pexit("Cannot write gid mappings: %s", gid_mapping);
 
       if (uid_mapping && write_mapping_file(pid, uid_mapping, false) < 0)
-        pexit("Cannot write gid mappings");
+        pexitf("Cannot write uid mappings: %s", uid_mapping);
```
saschagrunert, line 233: Pinns does not build unfortunately:

```
src/pinns.c: In function ‘main’:
src/pinns.c:230:59: error: macro "pexit" passed 2 arguments, but takes just 1
  230 |         pexit("Cannot write gid mappings: %s", gid_mapping);
      |                                                           ^
In file included from src/pinns.c:18:
src/utils.h:31: note: macro "pexit" defined here
   31 | #define pexit(s)                                                               \
      | 
src/pinns.c:230:9: error: ‘pexit’ undeclared (first use in this function); did you mean ‘exit’?
  230 |         pexit("Cannot write gid mappings: %s", gid_mapping);
      |         ^~~~~
      |         exit
```

saschagrunert (changes requested):

pinns/src/pinns.c
giuseppe: sorry, I had pushed the wrong version.

Could you please take a look now?

saschagrunert (approved):
Thank you!

PR #7582, opened 2023-12-12
bump buildah to latest version

kannon92:
#### What type of PR is this?

/kind dependency-change
Buildah is on v1.33.1 as of github.

```release-note
None
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -63,6 +63,12 @@ dependencies:
     - path: contrib/test/ci/vars.yml
       match: cri_tools_git_version
 
+  - name: buildah
+    version: v1.33.2
+    refPaths:
+    - path: scripts/versions
+      match: buildah
+
   - name: runc
     version: v1.1.9
     refPaths:
--- a/scripts/versions
+++ b/scripts/versions
@@ -10,6 +10,6 @@ declare -A VERSIONS=(
     ["runc"]=v1.1.9
     ["crun"]=1.9.2
     ["bats"]=v1.10.0
-    ["buildah"]=v1.28.0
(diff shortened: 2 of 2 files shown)


scripts/versions
```diff
@@ -10,6 +10,6 @@ declare -A VERSIONS=(
     ["runc"]=v1.1.9
     ["crun"]=1.9.2
     ["bats"]=v1.10.0
-    ["buildah"]=v1.28.0
+    ["buildah"]=v1.33.1
```
saschagrunert, line 13: Can we add this to `dependencies.yaml`? One reason that it's so much outdated is that the yaml file is the entry point for regularly checking new versions (manually).

saschagrunert (changes requested):
/ok-to-test

scripts/versions
kannon92: Done. PTAL

saschagrunert (approved):

dependencies.yaml
```diff
@@ -63,6 +63,12 @@ dependencies:
     - path: contrib/test/ci/vars.yml
       match: cri_tools_git_version
 
+  - name: buildah
+    version: v1.33.1
```
saschagrunert, line 67: Ah: https://github.com/containers/buildah/releases/tag/v1.33.2
```suggestion
    version: v1.33.2
```

scripts/versions
saschagrunert: This one has to be changed now as well to v1.33.2

kannon92:
@haircommander or @saschagrunert can I get an override of the e2e-gcp-ovn job since it looks like the aws one passed?

Issue #7550, opened 2023-12-05
provision Pod/Container with local Timezone during runtime


alireza-taleghani-stratsys:
Following OCI common spec we could instruct Containers to respect local TZ accordingly...
while podman and some other OCI complient clients (assume utilizing crun under the hood) follow common containers configuration as following Container runtime is able to report date/time with correct timezone indeed

```bash
# /etc/containers/containers.conf
...
[containers]
#tz="UTC"
tz = "Europe/Stockholm"
...
```
- ref: https://github.com/containers/common/blob/d0a7910051b13fe6cc69ca8efeee1d7e0005b4cd/pkg/config/containers.conf#L274

usage without/with configuration file/entry inplace:

```bash
$ podman run --rm -it docker.io/library/debian:jessie date
Tue Dec  5 09:21:36 CET 2023

# set tz=CET on [ /etc/containers/containers.conf ] 

$ podman run --rm -it docker.io/library/debian:jessie date
Tue Dec  5 08:21:46 UTC 2023
```

having that inplace will help us to have all pods/containers report the write timestamp on logs/msg without any need of extra env/volume mount as instructed here and there...

```yaml
# TZ via ENV
spec:
  containers:
  - name: my-container
    image: gcr.io/google-samples/node-hello:1.0
    env:
    - name: TZ
      value: "America/Toronto"
```

```yaml
# TZ via Volume
spec:
  containers:
  - name: my-container
    image: gcr.io/google-samples/node-hello:1.0
    volumeMounts:
    - name: tz-config
      mountPath: /etc/localtime
  volumes:
    - name: tz-config
      hostPath:
        path: /usr/share/zoneinfo/America/Toronto
        type: File
```

saschagrunert:
Hey @alireza-taleghani-stratsys, thank you for the enhancement request!

To sum it up, the request is that CRI-O interprets the data from `/etc/containers/containers.conf` to set the timezone for the workload?

saschagrunert:
@alireza-taleghani-stratsys thank you!

So the timezone support is part of c/common since https://github.com/containers/common/pull/200, and then got introduced into podman with https://github.com/containers/podman/pull/6836, https://github.com/containers/podman/pull/9537, https://github.com/containers/podman/pull/11128 or https://github.com/containers/podman/pull/18756.

I think we could optimize the shared code to avoid too much copy-pasta. What do other @cri-o/cri-o-maintainers think about it in general?

haircommander:
[...]

PR #7589, opened 2023-12-13
*: add support for specifying timezone for pod/container 

sohankunkerkar:
Addresses https://github.com/cri-o/cri-o/issues/7550

#### What type of PR is this?

/kind feature

```release-note
Add support for specifying timezone for pods and containers
```


docs/crio.conf.5.md
```diff
@@ -319,6 +319,9 @@ Enable CRI-O to generate the container pod-level events in order to optimize the
 **disable_hostport_mapping**=false
  Enable/Disable the container hostport mapping in CRI-O. Default value is set to 'false'.
 
+**time_zone**=""
+ The timezone to use for the container. If not specified, the CRI-O will use the host's timezone.
```
haircommander, line 323: nit: `for all containers` right? also, replace `the CRI-O` with `CRI-O`

pkg/config/reload.go
```diff
@@ -270,3 +271,22 @@ func (c *Config) ReloadRuntimes(newConfig *Config) error {
 
 	return nil
 }
+
+// ReloadTimeZone updates the TimeZone with the provided `newConfig`. It errors
+// if the timezone is not valid.
+func (c *Config) ReloadTimeZone(newConfig *Config) error {
```
haircommander, line 277: I don't know if reload support is really necessary

server/container_create_linux.go
```diff
@@ -906,6 +914,95 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	return ociContainer, nil
 }
 
+func copyTimezoneFile(mountLabel, zonePath string, c *oci.Container) (string, error) {
```
haircommander, line 917: can this be defined below where it's called?

internal/oci/container.go
```diff
@@ -717,3 +718,8 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
 	}
 	return c.runtimePath
 }
+
+// SetTimeZoneForPlatform sets the timezone for a given platform.
+func (c *Container) SetTimeZone(timezone string) {
+	c.timezone = timezone
```
haircommander, line 724: are we using this anywhere?

server/container_create_linux.go
```diff
@@ -906,6 +913,95 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+		ctr.SpecAddMount(rspec.Mount{
+			Destination: "/etc/localtime",
+			Type:        "bind",
+			Source:      localtimePath,
+			Options:     append(options, "bind"),
+		})
+	} else {
+		// File exists, let's create a symlink according to localtime(5)
+		log.Debugf(ctx, "Create localtime symlink for %s", timezonePath)
+		err = unix.Symlinkat(".."+timezonePath, etcFd, "localtime")
```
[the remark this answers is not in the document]
sohankunkerkar: It's the parent directory, forming a path to the desired timezone configuration file or dir.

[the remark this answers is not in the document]
sohankunkerkar: Done!

test/image.bats
```diff
@@ -338,3 +338,23 @@ EOF
 	output=$(crictl images -o json | jq '.images[] | select(.repoTags[] == "quay.io/crio/hello-wasm:latest") | .pinned')
 	[ "$output" == "true" ]
 }
+
+@test "run container in pod with timezone set" {
+	cat << EOF > "$CRIO_CONFIG_DIR/99-timezone.conf"
+[crio.runtime]
+time_zone = "Asia/Singapore"
+EOF
+	start_crio
```
haircommander, line 347: you could use `CONTAINER_TIME_ZONE="Asia/Signapore" start_crio` instead

haircommander:
/approve

pkg/config/config.go
```diff
@@ -1051,6 +1058,13 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		return err
 	}
 
+	if c.TimeZone != "" {
```
kannon92, line 1061: Can you cover this with a unit test? I see a corresponding `config_test.go` in this directory so you should be able to cover.

test/image.bats
```diff
@@ -338,3 +338,18 @@ EOF
 	output=$(crictl images -o json | jq '.images[] | select(.repoTags[] == "quay.io/crio/hello-wasm:latest") | .pinned')
 	[ "$output" == "true" ]
 }
+
+@test "run container in pod with timezone set" {
```
kannon92, line 342: Not sure if we want to cover it here but it'd be nice to see what happens if time zone is invalid?

Not existent timezone and an empty string for CONTAINER_TIME_ZONE.

If for some reason we can't read the timezone database or it doesn't exist, do we panic or just ignore it?

server/container_create_linux.go
```diff
@@ -906,6 +913,95 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	}
+
+	hostPath, err := securejoin.SecureJoin(mountPoint, timezonePath)
+	if err != nil {
+		return fmt.Errorf("resolve zoneinfo path in the container: %w", err)
+	}
+
+	if _, err := os.Stat(hostPath); err != nil {
+		// File does not exist, which means tzdata is not installed in the container.
+		// Create /etc/locatime as a copy from the host.
```
kannon92, line 949:
```suggestion
		// Create /etc/localtime as a copy from the host.
```

pkg/config/config.go
sohankunkerkar: Sure, fixed.

test/image.bats
sohankunkerkar: >Not sure if we want to cover it here but it'd be nice to see what happens if time zone is invalid?

I don't think we need an integration test for this. Runtime config validation will cover this use case, also addressed in `config_test.go`.

> Not existent timezone and an empty string for CONTAINER_TIME_ZONE.

For an empty string, it will not do anything, which means it will take the host machine time zone. This is as good as saying this feature doesn't exist.

>If for some reason we can't read the timezone database or it doesn't exist, do we panic or just ignore it?

We are not panicking but throwing appropriate errors wherever required, covered in `setContainerTimeZone`.

server/container_create_linux.go
```diff
@@ -906,6 +913,95 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+
+	// Make sure to remove any existing localtime file in the container to not create invalid links
+	err = unix.Unlinkat(etcFd, "localtime", 0)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("removing /etc/localtime: %w", err)
+	}
+
+	hostPath, err := securejoin.SecureJoin(mountPoint, timezonePath)
+	if err != nil {
+		return fmt.Errorf("resolve zoneinfo path in the container: %w", err)
```
kannon92, line 944: What is meant by zoneinfo?

server/container_create_linux.go
```diff
@@ -906,6 +913,95 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+		// File does not exist, which means tzdata is not installed in the container.
+		// Create /etc/localtime as a copy from the host.
+		log.Debugf(ctx, "Timezone %s does not exist in the container, create our own copy from the host", timezonePath)
+		localtimePath, err := copyTimezoneFile(mountLabel, timezonePath, ociContainer)
+		if err != nil {
+			return fmt.Errorf("setting timezone for container %s: %w", ociContainer.ID(), err)
+		}
+
+		ctr.SpecAddMount(rspec.Mount{
+			Destination: "/etc/localtime",
```
kannon92, line 957: You pass in `etc` as a variable here but then you use a hardcoded value of `/etc/localtime`.

test/image.bats
kannon92: Sounds good to me. As long as they are covered somewhere.

server/container_create_linux.go
sohankunkerkar:
```sh
$ ls -ltr /usr/share/zoneinfo/
total 356
-rw-r--r--.  1 root root   4446 Jan  7  2023 iso3166.tab
-rw-r--r--.  1 root root  10666 Jan 20  2023 leap-seconds.list
-rw-r--r--.  1 root root   3392 Jan 20  2023 leapseconds
-rw-r--r--.  1 root root  18855 Jan 23  2023 zone.tab
-rw-r--r--.  1 root root  17551 Jan 23  2023 zone1970.tab
-rw-r--r--.  8 root root    114 Mar 28  2023 Zulu
-rw-r--r--.  2 root root   1535 Mar 28  2023 W-SU
-rw-r--r--.  1 root root   1905 Mar 28  2023 WET
-rw-r--r--.  8 root root    114 Mar 28  2023 UTC
-rw-r--r--.  8 root root    114 Mar 28  2023 Universal
-rw-r--r--.  8 root root    114 Mar 28  2023 UCT
-rw-r--r--.  1 root root 109248 Mar 28  2023 tzdata.zi
-rw-r--r--.  3 root root   1947 Mar 28  2023 Turkey
-rw-r--r--.  3 root root    415 Mar 28  2023 Singapore
-rw-r--r--.  2 root root    617 Mar 28  2023 ROK
-rw-r--r--.  2 root root    761 Mar 28  2023 ROC
-rw-r--r--.  1 root root   2310 Mar 28  2023 PST8PDT
-rw-r--r--.  5 root root    561 Mar 28  2023 PRC
-rw-r--r--.  3 root root   3552 Mar 28  2023 posixrules
-rw-r--r--.  2 root root   3497 Mar 28  2023 Portugal
-rw-r--r--.  2 root root   2654 Mar 28  2023 Poland
-rw-r--r--.  2 root root   2068 Mar 28  2023 NZ-CHAT
-rw-r--r--.  4 root root   2437 Mar 28  2023 NZ
-rw-r--r--.  4 root root   2460 Mar 28  2023 Navajo
-rw-r--r--.  1 root root   2310 Mar 28  2023 MST7MDT
-rw-r--r--.  1 root root    114 Mar 28  2023 MST
-rw-r--r--.  1 root root   2094 Mar 28  2023 MET
-rw-r--r--.  2 root root    625 Mar 28  2023 Libya
-rw-r--r--.  2 root root    316 Mar 28  2023 Kwajalein
-rw-r--r--.  2 root root    309 Mar 28  2023 Japan
-rw-r--r--.  2 root root    482 Mar 28  2023 Jamaica
-rw-r--r--.  3 root root   2388 Mar 28  2023 Israel
-rw-r--r--.  2 root root   1262 Mar 28  2023 Iran
-rw-r--r--. 14 root root    148 Mar 28  2023 Iceland
-rw-r--r--.  1 root root    115 Mar 28  2023 HST
-rw-r--r--.  2 root root   1233 Mar 28  2023 Hongkong
-rw-r--r--. 10 root root    114 Mar 28  2023 Greenwich
-rw-r--r--. 10 root root    114 Mar 28  2023 GMT0
-rw-r--r--. 10 root root    114 Mar 28  2023 GMT-0
-rw-r--r--. 10 root root    114 Mar 28  2023 GMT+0
-rw-r--r--. 10 root root    114 Mar 28  2023 GMT
-rw-r--r--.  7 root root   3664 Mar 28  2023 GB-Eire
-rw-r--r--.  7 root root   3664 Mar 28  2023 GB
-rw-r--r--.  1 root root    116 Mar 28  2023 Factory
-rw-r--r--.  1 root root   2310 Mar 28  2023 EST5EDT
-rw-r--r--.  1 root root    114 Mar 28  2023 EST
-rw-r--r--.  2 root root   3492 Mar 28  2023 Eire
-rw-r--r--.  2 root root   2399 Mar 28  2023 Egypt
-rw-r--r--.  1 root root   1908 Mar 28  2023 EET
-rw-r--r--.  2 root root   2416 Mar 28  2023 Cuba
-rw-r--r--.  1 root root   2310 Mar 28  2023 CST6CDT
-rw-r--r--.  1 root root   2094 Mar 28  2023 CET
drwxr-xr-x.  1 root root    798 May 11  2023 Africa
drwxr-xr-x.  1 root root     24 May 11  2023 Arctic
drwxr-xr-x.  1 root root    170 May 11  2023 Antarctica
drwxr-xr-x.  1 root root   2474 May 11  2023 America
drwxr-xr-x.  1 root root     46 May 11  2023 Chile
drwxr-xr-x.  1 root root    132 May 11  2023 Canada
drwxr-xr-x.  1 root root     42 May 11  2023 Brazil
drwxr-xr-x.  1 root root    308 May 11  2023 Australia
drwxr-xr-x.  1 root root    188 May 11  2023 Atlantic
drwxr-xr-x.  1 root root   1456 May 11  2023 Asia
drwxr-xr-x.  1 root root    366 May 11  2023 Etc
drwxr-xr-x.  1 root root     46 May 11  2023 Mexico
drwxr-xr-x.  1 root root    164 May 11  2023 Indian
drwxr-xr-x.  1 root root    922 May 11  2023 Europe
drwxr-xr-x.  1 root root    190 May 11  2023 US
drwxr-xr-x.  1 root root    602 May 11  2023 Pacific
drwxr-xr-x.  1 root root    656 May 11  2023 posix
drwxr-xr-x.  1 root root    656 May 11  2023 right

```

server/container_create_linux.go
sohankunkerkar: so the `etc` [value](https://github.com/cri-o/cri-o/pull/7589/files#diff-1e0877112035840d3e2f073c693665f8daffa241a440e54a866ad7382b510ea2R855) passed to this function is different than the one used here.

kannon92 (approved):
/lgtm

pkg/config/config.go
```diff
@@ -478,6 +479,11 @@ type RuntimeConfig struct {
 	// Option to disable hostport mapping in CRI-O
 	// Default value is 'false'
 	DisableHostPortMapping bool `toml:"disable_hostport_mapping"`
+
+	// Option to specify the timezone inside the container.
+	// If set to "Local," it means the container will use
+	// the same timezone as the host machine.
```
saschagrunert, line 485: The overall behavior should be also documented in the CLI and markdown docs. I think we can also state the unsetting the field refers to UTC. I'd say we use `local` instead of `Local` and/or lowercase the input.

pkg/config/config.go
```diff
@@ -882,6 +888,7 @@ func DefaultConfig() (*Config, error) {
 			ulimitsConfig:               ulimits.New(),
 			HostNetworkDisableSELinux:   true,
 			DisableHostPortMapping:      false,
+			TimeZone:                    "",
```
saschagrunert, line 891: No need to set that field.

pkg/config/template.go
```diff
@@ -1355,6 +1360,13 @@ const templateStringCrioRuntimeDisableHostPortMapping = `# disable_hostport_mapp
 
 `
 
+const templateStringCrioRuntimeTimeZone = `# time_zone specifies the timezone configuration for CRI-O.
+# If an empty string is provided, CRI-O will use the timezone of the host machine.
```
saschagrunert, line 1364: That's not true, is it? We should also add the complete docs about the behavior.

internal/criocli/criocli.go
```diff
@@ -1173,6 +1176,12 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			EnvVars: []string{"DISABLE_HOSTPORT_MAPPING"},
 			Value:   defConf.DisableHostPortMapping,
 		},
+		&cli.StringFlag{
+			Name:    "time-zone",
```
saschagrunert, line 1180: Let's add a `tz` alias to the flag.

server/container_create_linux.go
```diff
@@ -906,6 +913,95 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	}
+	defer unix.Close(etcFd)
+
+	// Make sure to remove any existing localtime file in the container to not create invalid links
+	err = unix.Unlinkat(etcFd, "localtime", 0)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("removing /etc/localtime: %w", err)
+	}
+
+	hostPath, err := securejoin.SecureJoin(mountPoint, timezonePath)
```
saschagrunert, line 942: nit: `hostPath` is misleading, how about something like `timezonePathCtr`?

server/container_create_linux.go
```diff
@@ -906,6 +913,95 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	return ociContainer, nil
 }
 
+func setContainerTimeZone(ctx context.Context, tz, mountPoint, etc, mountLabel string, options []string, ociContainer *oci.Container, ctr ctrfactory.Container) error {
+	var err error
+	if tz == "" {
+		return nil
+	}
+
+	timezonePath := filepath.Join("/usr/share/zoneinfo", tz)
```
saschagrunert, line 922: We should either make this configurable or parse the `TZDIR` environment variable: https://sourceware.org/git/?p=glibc.git;a=blob;f=time/tzfile.c;h=8a923d0cccc927a106dc3e3c641be310893bab4e;hb=HEAD#l149

pkg/config/template.go
sohankunkerkar: It uses the timezone of the host machine but displays it in UTC

saschagrunert: @sohankunkerkar let's document that as well

sohankunkerkar: Done

pkg/config/config.go
sohankunkerkar: Fixed

saschagrunert:
CI is not happy with the change: https://github.com/cri-o/cri-o/actions/runs/7821892538/job/21339890665?pr=7589

internal/criocli/criocli.go
```diff
@@ -1182,6 +1185,13 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
...
+			Aliases: []string{"tz"},
+			Usage:   "Specifies the timezone configuration for CRI-O. If not specified, CRI-O will use the host's timezone but displays in UTC.",
```
kwilczynski, line 1191: This help is different from the one we quote in the configuration file. The last few phrases of the sentence could be made a little better. :smile:

pkg/config/config.go
```diff
@@ -1056,6 +1062,13 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		return err
 	}
 
+	if c.TimeZone != "" && !strings.EqualFold(c.TimeZone, "local") {
+		_, err := time.LoadLocation(c.TimeZone)
+		if err != nil {
+			return fmt.Errorf("invalid timezone %s", c.TimeZone)
```
kwilczynski, line 1068:
```suggestion
			return fmt.Errorf("invalid time zone: %s", c.TimeZone)
```

kwilczynski: However, here I am on the fence about whether to reference the configuration file item or spell it as "time zone" (separate words). Yes, it might be a nit pick. :smile:

pkg/config/template.go
```diff
@@ -1368,6 +1373,13 @@ const templateStringCrioRuntimeDisableHostPortMapping = `# disable_hostport_mapp
 
 `
 
+const templateStringCrioRuntimeTimeZone = `# time_zone specifies the timezone configuration for CRI-O.
+# If an empty string is provided, CRI-O will use the timezone of the host machine but displays it in UTC.
+# By default, CRI-O manages the timezone configuration.
```
kwilczynski, line 1378: This is self-explanatory.

```suggestion
```

pkg/config/config.go
kwilczynski: Why? For consistency, the command-line option uses a dash to separate words, the English language seems to favour "time zone" as separate words, and then we have `time_zone` in the configuration file.

server/container_create_linux.go
```diff
@@ -921,6 +904,57 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	}
+	if localTimePath != "" {
+		if err := securityLabel(localTimePath, mountLabel, false, false); err != nil {
+			return err
+		}
+		ctr.SpecAddMount(rspec.Mount{
+			Destination: "/etc/localtime",
+			Type:        "bind",
+			Source:      localTimePath,
+			Options:     append(options, "bind"),
```
kwilczynski, line 952: Should this be made read-only?

```suggestion
			Options:     []string{"ro", "bind"},
```

Or perhaps even the whole shebang:

```suggestion
			Options:     []string{"ro", "bind", "nodev", "nosuid", "noexec"},
```

kwilczynski: Not sure what is our take on such system-wide files? Perhaps also for a privileged container, this should be made read/write?

sohankunkerkar: Yeah, that's a good point. I don't think we need `r/w`, let me change that to `Options:     append(options, "ro", "bind", "nodev", "nosuid", "noexec"),`

.golangci.yml
```diff
@@ -233,7 +233,7 @@ linters-settings:
       # - filepathJoin
       # - whyNoLint
   gocyclo:
-    min-complexity: 160
+    min-complexity: 165
```
sohankunkerkar, line 236: Discussed this with @haircommander offline. I'm going to handle the refactoring of `createSandboxContainer` in a separate PR.

completions/fish/crio.fish
```diff
@@ -161,6 +161,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l stream-port -r -d 'Bind po
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-ca -r -d 'Path to the x509 CA(s) file used to verify and authenticate client communication with the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes.'
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-cert -r -d 'Path to the x509 certificate file used to serve the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes.'
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-key -r -d 'Path to the key file used to serve the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes.'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l time-zone -s tz -r -d 'Specifies the timezone configuration for CRI-O. If an empty string is provided, CRI-O will use the timezone of the host machine but displays it in UTC.'
```
haircommander, line 164: nit: time zone*

pkg/config/config.go
```diff
@@ -479,6 +480,11 @@ type RuntimeConfig struct {
 	// Option to disable hostport mapping in CRI-O
 	// Default value is 'false'
 	DisableHostPortMapping bool `toml:"disable_hostport_mapping"`
+
+	// Option to specify the timezone inside the container.
```
haircommander, line 484: same here

completions/fish/crio.fish
haircommander: @sohankunkerkar mentioned offline that this is the american spelling, which we have typically favored over the british.

haircommander:
/approve

LGTM

server/container_create_linux.go
haircommander: I may be wrong here, but I'm pretty sure we're copying the host tz file to the container, or symlinking from one container path to another: https://github.com/cri-o/cri-o/pull/7589/files#diff-d1b079a5798eb933a9cee5821c9b039d99cd82fe192caf20966f5b734f0dd0a0R58-R73
@sohankunkerkar is that correct? or are we mounting the host path into the container?

sohankunkerkar: whoops, I missed that part. @haircommander you are right

kwilczynski: Nice catch! @haircommander :+1: 

I didn't look at the new library, admittedly.

sohankunkerkar:
If the time zone is set to an empty string `" "`, CRI-O retains its default behavior, aligning with the podman codebase.
xref: https://github.com/containers/common/blob/main/pkg/timezone/timezone.go#L24

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

completions/bash/crio
```diff
@@ -127,6 +127,7 @@ h
 --stream-tls-ca
 --stream-tls-cert
 --stream-tls-key
+--time-zone
```
kwilczynski, line 130: Not moving to `--timezone` for consistency?

completions/zsh/_crio
```diff
@@ -134,6 +134,7 @@ it later with **--config**. Global options will modify the output.'
         '--stream-tls-ca'
         '--stream-tls-cert'
         '--stream-tls-key'
+        '--time-zone'
```
kwilczynski, line 137: Not moving to `--timezone` for consistency?

docs/crio.8.md
```diff
@@ -123,6 +123,7 @@ crio
 [--stream-tls-ca]=[value]
 [--stream-tls-cert]=[value]
 [--stream-tls-key]=[value]
+[--time-zone|--tz]=[value]
```
kwilczynski, line 126: Not moving to --timezone for consistency?

kwilczynski, line 126: Not moving to `--timezone` for consistency?

docs/crio.conf.5.md
```diff
@@ -325,6 +325,9 @@ Enable CRI-O to generate the container pod-level events in order to optimize the
 **disable_hostport_mapping**=false
  Enable/Disable the container hostport mapping in CRI-O. Default value is set to 'false'.
 
+**time_zone**=""
```
kwilczynski, line 328: Not moving to `timezone=` for consistency?

internal/criocli/criocli.go
```diff
@@ -414,6 +414,9 @@ func mergeConfig(config *libconfig.Config, ctx *cli.Context) error {
 	if ctx.IsSet("disable-hostport-mapping") {
 		config.DisableHostPortMapping = ctx.Bool("disable-hostport-mapping")
 	}
+	if ctx.IsSet("time-zone") {
```
kwilczynski, line 417: As, with some the prior comments. :smile:

server/container_create_linux.go
kwilczynski: Just to follow up, as I don't know where we are at? Are we keeping the bind mount here? If so, then are we making it also read-only?

pkg/config/config.go
```diff
@@ -1056,6 +1062,13 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		return err
 	}
 
+	if c.Timezone != "" && !strings.EqualFold(c.Timezone, "local") {
+		_, err := time.LoadLocation(c.Timezone)
+		if err != nil {
+			return fmt.Errorf("invalid timezone %s", c.Timezone)
```
kwilczynski, line 1068: Just a nit (also, for consistency with other errors):

```suggestion
			return fmt.Errorf("invalid timezone: %s", c.Timezone)
```

pkg/config/config.go
```diff
@@ -1056,6 +1062,13 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		return err
 	}
 
+	if c.Timezone != "" && !strings.EqualFold(c.Timezone, "local") {
```
kwilczynski, line 1065: I think this could be simplified to just the following:
```suggestion
	if c.Timezone != "local" {
```

Why? There is no UTF-8 or Unicode in the TZ identifier to speak of.

pkg/config/config.go
```diff
@@ -479,6 +480,11 @@ type RuntimeConfig struct {
 	// Option to disable hostport mapping in CRI-O
 	// Default value is 'false'
 	DisableHostPortMapping bool `toml:"disable_hostport_mapping"`
+
+	// Option to specify the timezone inside the container.
+	// If it's empty, the container will use the same timezone
+	// as the host machine but in UTC.
```
kwilczynski, line 486: The host might have some local time zone set. So it will use whatever this time zone would be. The `UTC` (or `Etc/UTC`) here is a bit confusing.

internal/oci/oci.go
```diff
@@ -182,6 +182,11 @@ func (r *Runtime) RuntimeType(runtimeHandler string) (string, error) {
 	return rh.RuntimeType, nil
 }
 
+// Timezone returns the time zone for a given platform.
```
kwilczynski, line 185: What is _platform_ here?

pkg/config/template.go
```diff
@@ -1368,6 +1373,12 @@ const templateStringCrioRuntimeDisableHostPortMapping = `# disable_hostport_mapp
 
 `
 
+const templateStringCrioRuntimeTimezone = `# time_zone specifies the timezone configuration for CRI-O.
```
kwilczynski, line 1376: This would need to be aligned with the command-line help output. We also would need to mention the ability for the user to set `local` as a viable timezone to denote CRI-O setting the timezone within the container to that of the host.

completions/fish/crio.fish
```diff
@@ -161,6 +161,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l stream-port -r -d 'Bind po
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-ca -r -d 'Path to the x509 CA(s) file used to verify and authenticate client communication with the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes.'
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-cert -r -d 'Path to the x509 certificate file used to serve the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes.'
 complete -c crio -n '__fish_crio_no_subcommand' -l stream-tls-key -r -d 'Path to the key file used to serve the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes.'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l time-zone -s tz -r -d 'Specifies the timezone configuration for CRI-O. If an empty string is provided, CRI-O retains its default behavior.'
```
kwilczynski, line 164: The empty string should be the default here. Perhaps something like this could do?

> The timezone that CRI-O will set for the running container. When no timezone is set or an empty value is provided, then no timezone will be set, leaving the container with its default. Using "local" as a timezone will cause CRI-O to set the container timezone to that of the host.

kwilczynski: Why? CRI-O _sets_ the timezone for a container. We don't technically set timezone or locales for CRI-O itself.

docs/crio.8.md
```diff
@@ -420,6 +421,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--stream-tls-key**="": Path to the key file used to serve the encrypted stream. This file can change and CRI-O will automatically pick up the changes within 5 minutes.
 
+**--time-zone, --tz**="": Specifies the timezone configuration for CRI-O. If an empty string is provided, CRI-O retains its default behavior.
```
kwilczynski, line 424: Not moving to `--timezone` for consistency?

pkg/config/config.go
sohankunkerkar: I believe we shouldn't solely check for 'local' because if the time zone is empty (' '), it enters that condition and executes `LoadLocation`, which returns UTC. Our hypothesis relies on the assumption that CRI-O will not take any action if the time zone is an empty string.

pkg/config/config.go
sohankunkerkar: I removed that line completely to avoid confusion.

completions/fish/crio.fish
sohankunkerkar: I would incorporate this information into the configuration description rather than adding it here and making the description bulkier.

pkg/config/config.go
kwilczynski: Ahh. Makes sense. Good point. My focus was to remove the comparison function useful for Unicode. Which will not be the case here, I believe.

kwilczynski:
@sohankunkerkar, thank you for working on this! Looks amazing! 🎉

PR #7592, opened 2023-12-13
server: deprecate CRI-O userns support

haircommander:
as support will be beta in kubernetes soon hopefully

```release-note
Mark `io.kubernetes.cri-o.userns-mode`  annotation, `uid_mappings` and `gid_mappings` configuration options as deprecated. They will be replaced with Kubernetes native user namespace support in the future. 
```


haircommander:
@cri-o/cri-o-maintainers PTAL

server/sandbox_run_linux.go
```diff
@@ -426,6 +426,9 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 	kubeAnnotations := sbox.Config().Annotations
 
 	usernsMode := kubeAnnotations[ann.UsernsModeAnnotation]
+	if usernsMode != "" {
+		log.Errorf(ctx, "Annotation 'io.kubernetes.cri-o.userns-mode' is deprecated, and will be replaced with native Kubernetes support for user namespaces in the future")
```
saschagrunert, line 430: nit: Do we want to warn here instead of error?

pkg/config/template.go
```diff
@@ -1105,27 +1105,31 @@ const templateStringCrioRuntimeLogFilter = `# Filter the log messages by the pro
 const templateStringCrioRuntimeUIDMappings = `# The UID mappings for the user namespace of each container. A range is
 # specified in the form containerUID:HostUID:Size. Multiple ranges must be
 # separated by comma.
+# This option is deprecated, and will be replaced with native Kubernetes user namespace support in the future.
```
saschagrunert, line 1108: Do we want to include a link to the k8s.io docs here?

server/sandbox_run_linux.go
haircommander: done!

pkg/config/template.go
haircommander: hm how about adding the KEP number instead?:

saschagrunert (approved):

PR #7603, opened 2023-12-18
Ensure that the exec-process-* temporary file is correctly deleted wh…

lidajia:
#### What type of PR is this?
issue: #7604 

/kind bug

Ensure that the exec-process-* temporary file is correctly deleted when an error occurs.

When there is an error in writing processFile (such as when the disk where the/tmp directory is located is full), the return statement will modify the ‘processFile’ var to "". In the defer function, the ‘processFile’ uses a global variable, so the processFile file will not be deleted correctly, resulting in a large number of empty processFile files

issue: #7604 

```release-note
None
```


saschagrunert (approved):
Great catch!

PR #7608, opened 2023-12-18
OCPBUGS-25647: runtimehandler: mixedcpus: fix quota calculation

Tal-or:
#### What type of PR is this?

/kind bug

Turns out the `podManager.GetCgroups()` doesn\'t return any information with regards to its existing resources.

We should put some extra effort in order to find and calculate the actual pod quota.
We need to sum the quota value from each container + the shared cpus and that should be our new pod quota.

None

```release-note
None
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -51,6 +51,8 @@ const (
 
 const (
 	cgroupSubTreeControl = "cgroup.subtree_control"
+	cgroupV1QuotaFile    = "cpu.cfs_quota_us"
+	cgroupV2QuotaFile    = "cpu.max"
 	cpusetCpus           = "cpuset.cpus"
 	cpusetCpusExclusive  = "cpuset.cpus.exclusive"
 	IsolatedCPUsEnvVar   = "OPENSHIFT_ISOLATED_CPUS"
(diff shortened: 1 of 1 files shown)


Tal-or:
/hold 
More local testing are needed

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1119,7 +1115,24 @@ func calculateMaximalQuota(cpus *cpuset.CPUSet, period uint64) (quota int64, err
...
-func calculatePodQuota(sharedCpus *cpuset.CPUSet, existingQuota int64, period uint64) (int64, error) {
+func calculatePodQuota(sharedCpus *cpuset.CPUSet, containerManagers []cgroups.Manager, period uint64) (int64, error) {
+	var existingQuota int64
+	var quotaPath string
+	if !node.CgroupIsV2() {
+		quotaPath = "cpu"
+	}
+	for _, mng := range containerManagers {
+		controllerPath := mng.Path(quotaPath)
+		q, err := cgroups.ReadFile(controllerPath, "cpu.cfs_quota_us")
```
haircommander, line 1126: this isn't present for v2

Tal-or: you're right, I was focusing on the v1 fix.

Tal-or: Fixed for both

Tal-or:
Tested on both v1 and v2

haircommander:
/ok-to-test
/approve

thanks, @cri-o/cri-o-maintainers PTAL

haircommander:
@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):

PR #7613, opened 2023-12-19
NO_PUBKEY bug fix with buster-backports install

alephpt:
#### What type of PR is this?
/kind documentation

More accurate documentation steps, and easier, faster and safer install process than alternative recommendations out there.

None (that I know of)

This single note would have saved me close to an hour of debugging, and research, and trial and hour - not to mention, that I could have stopped half-way and left a security risk, had I not taken the extra steps to figure this out.

Yes. 

```release-note
The modified steps proposed in the documentation were needed to install the dependencies required for CRI-O on AWS Ubuntu 22.

Adding backports, as recommended, returns no NO_PUBKEY error. 

Common "fixes" online shows adding the key to '/etc/apt/trusted.gpg.d' with apt-key which is deprecated due to security reasons.
```

### DIFF
--- a/install.md
+++ b/install.md
@@ -190,7 +190,7 @@ apt update
 apt install -y -t buster-backports libseccomp2 || apt update -y -t buster-backports libseccomp2
 ```
 
-If the above provides a GPG error stating `The following signatures couldn't be verified because the public key is not available: NO_PUBKEY <Key>` the following steps may be necessary:
+If the above results in a GPG error stating `The following signatures couldn't be verified because the public key is not available: NO_PUBKEY <Key>` the following steps may be necessary:
 
 ```shell
 apt-key export <Key> | sudo gpg --dearmour >> /usr/share/keyrings/backports.gpg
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -190,6 +190,17 @@ apt update
 apt install -y -t buster-backports libseccomp2 || apt update -y -t buster-backports libseccomp2
 ```
 
+If the above provides a GPG error stating `The following signatures couldn't be verified because the public key is not available: NO_PUBKEY <Key>` the following steps may be necessary:
```
kwilczynski, line 193:
```suggestion
If the above results in a GPG error stating `The following signatures couldn't be verified because the public key is not available: NO_PUBKEY <Key>` the following steps may be necessary:
```

alephpt:
> /ok-to-test

Are we OK to retest? I'm not sure why this minor revision was held up?

kwilczynski:
@alephpt, sorry for the delay!

I poked the Pull Request to get the tests restarted. I see markdown lint is complaining. Something to have a look into.

PR #7614, opened 2023-12-19
Add an annotation for asking for a RW sysfs mount

dgl:
#### What type of PR is this?

/kind feature

This allows some additional use-cases for user namespaces, in particular when running systemd + docker within a container, it allows using `docker run --privileged`.

Even with `procMountType: Unmasked` because of how masked mounts work, a container cannot mount sysfs read-write, because /sys is already mounted read-only. Mounting sysfs read-write *within* a user namespace is something that a standard user on Linux can do:

```
$ unshare -Urnm bash
root@dev4:~# mkdir /tmp/sys
root@dev4:~# mount -t sysfs none /tmp/sys
root@dev4:~# mount | grep /tmp/sys
none on /tmp/sys type sysfs (rw,relatime)
```

So therefore this isn't as scary as it may seem. It is obviously not safe if the container is not within a user namespace.

Note this does go against the [systemd container interface spec](https://systemd.io/CONTAINER_INTERFACE/) which specifies sysfs should be mounted read-only. I have tested this with systemd and docker and haven't yet noticed any unexpected behaviour as a result, potentially this should be investigated more, but this seems fine for an experimental feature for now. (We could potentially consider mounting a read-write sysfs elsewhere to keep the systemd interface correct.)

None

```release-note
Add an experimental `io.kubernetes.cri-o.sysfs-mount-rw` annotation to request `/sys` is mounted read-write.
```

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -344,6 +344,7 @@ The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  Th
   "io.kubernetes.cri-o.userns-mode" for configuring a user namespace for the pod.
   "io.kubernetes.cri-o.Devices" for configuring devices for the pod.
   "io.kubernetes.cri-o.ShmSize" for configuring the size of /dev/shm.
+  "io.kubernetes.cri-o.sysfs-mount-rw" to mount sysfs writable when set to "true" (only safe combined with userns-mode).
   "io.kubernetes.cri-o.UnifiedCgroup.$CTR_NAME" for configuring the cgroup v2 unified block for a container.
   "io.containers.trace-syscall" for tracing syscalls via the OCI seccomp BPF hook.
 
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -72,6 +72,9 @@ const (
 
 	// PlatformRuntimePath indicates the runtime path that CRI-O should use for a specific platform.
 	PlatformRuntimePath = "io.kubernetes.cri-o.PlatformRuntimePath"
+
+	// SysfsRWAnnotation specifies mounting sysfs as an rw filesystem.
+	SysfsRWAnnotation = "io.kubernetes.cri-o.sysfs-mount-rw"
 )
 
 var AllAllowedAnnotations = []string{
@@ -94,4 +97,5 @@ var AllAllowedAnnotations = []string{
 	PodLinuxResources,
 	LinkLogsAnnotation,
 	CPUSharedAnnotation,
+	SysfsRWAnnotation,
 }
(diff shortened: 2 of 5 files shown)


saschagrunert (approved):
Code LGTM

/ok-to-test

/hold
do we want to include this in v1.29?

docs/crio.conf.5.md
```diff
@@ -344,6 +344,7 @@ The "crio.runtime.runtimes" table defines a list of OCI compatible runtimes.  Th
   "io.kubernetes.cri-o.userns-mode" for configuring a user namespace for the pod.
   "io.kubernetes.cri-o.Devices" for configuring devices for the pod.
   "io.kubernetes.cri-o.ShmSize" for configuring the size of /dev/shm.
+  "io.kubernetes.cri-o.sysfs-mount-rw" to mount sysfs writable when set to "true" (only safe combined with userns-mode).
```
saschagrunert, line 347: Can we restrict that to only work with userns mode?

kannon92: ping @dgl

t-cas:
Would you be open to reviving this? We hit exactly the problem described in the original description, from the other direction: trying to run `kind` (which uses `docker run --privileged` for its node containers) inside a non-privileged, user-namespaced pod on OpenShift/CRI-O 1.33.

The kernel refuses the nested rw sysfs mount via `mount_too_revealing()`, because `/sys` in a non-privileged container is read-only at the superblock level. CRI-O mounts `/sys` rw only for privileged containers (`addSysfsMounts()`), and privileged **plus** userns doesn't work either, 
the privileged path skips the `cgroup2-mount-hierarchy-rw` ownership fix, leaving the pod cgroup owned by an unmapped uid, so a rootful dockerd can't delegate cgroup v2 controllers. The two requirements are mutually exclusive today, which forces us onto plain privileged dind with no user namespace, strictly worse than what this PR would enable.

Measured on OCP 4.20.25 / CRI-O 1.33.12:

| Config | `/sys` | pod cgroup | Outcome |
|---|---|---|---|
| non-privileged + userns | `ro` (sb `ro`) | uid=0 `755` writable | dockerd OK, `kind` fails |
| privileged + userns | `rw` (sb `rw`) | uid=65534 `555` not writable | dockerd fails |
| privileged, no userns | `rw` | writable | works, but no userns confinement |

Happy to do the rebase, `main` has since moved annotations to `pkg/annotations/v2` with paired V2/V1 names, and `addOCIBindMounts` now takes `sb`, so it's a port rather than a rebase. One question before writing it: would you want the annotation **gated on the container actually having a user namespace**, rather than honoured unconditionally? The safety argument rests entirely on the userns, and a cluster admin allowlisting the annotation on a runtime handler won't necessarily know that.

dgl:
@t-cas I actually have an updated version of this, we've been using it with Kind too. It's on my list to revisit and upstream, I'll aim to get to it next week.

t-cas:
> @t-cas I actually have an updated version of this, we've been using it with Kind too. It's on my list to revisit and upstream, I'll aim to get to it next week.

Great to hear! We've hit this exact issue (kind inside a user-namespaced pod on OCP 4.20 / CRI-O 1.33), so we're very interested in this. Feel free to reach out if you'd like an extra set of eyes on testing or review when you're ready to push upstream.

PR #7269, opened 2023-09-01
CRIO changes to support split filesystem 

kannon92:
#### What type of PR is this?
/kind api-change

#### What this PR does / why we need it:
Implement Image_Fs_Info as part of the separate gc KEP.  We should not merge unless this KEP is approved.  TODO: Add a link.  

Opening this up now to gather feedback on feasibility of this implementation.

Note, since I had a few changes in other areas, I vendored these changes here.  I already opened up a PR into c/storage.  Kubernetes is based on KEP approval.
#### Special notes for your reviewer:
We are having some problems with the integration tests picking up an older version of crictl. I moved the integration tests to https://github.com/cri-o/cri-o/pull/7269 just to move this PR through.
#### Does this PR introduce a user-facing change?

```release-note
Added more file system information in `ImageFsInfo` as part of the garbage collection KEP.
```

PR #7618, opened 2023-12-20
add imagefs integration tests

kannon92:
#### What type of PR is this?
/kind ci

Add integration tests for split disk KEP-4191.

```release-note
NONE
```


kannon92:
@sohankunkerkar @haircommander I moved the integration tests to this PR so we can iterate on that and get the `imagefsinfo` changes into main.

The test is mostly failing due to an older version of crictl present on the VM.

kannon92:
/hold
Needs https://github.com/cri-o/cri-o/pull/7269 to merge first.

kannon92:
@saschagrunert @haircommander I got stuck on this PR actually. For some reason, we were discovering that crictl was an older version and it wasn't actually building the correct one.

Do either of you have an idea of a path forward?

@sohankunkerkar and I looked at it but couldn't really find a reason why crictl was old.

saschagrunert:
> @saschagrunert @haircommander I got stuck on this PR actually. For some reason, we were discovering that crictl was an older version and it wasn't actually building the correct one.
> 
> Do either of you have an idea of a path forward?
> 
> @sohankunkerkar and I looked at it but couldn't really find a reason why crictl was old.

In the CI image? May it be possible that it gets pre-shipped in the image and we have to override it?

kannon92:
@kwilczynski I could use some help on this one actually.

kannon92:
@cri-o/cri-o-maintainers PTAL.

I think with @sohankunkerkar updates to VM builds this may be ready to merge.

test/imagefsinfo.bats
```diff
@@ -0,0 +1,39 @@
...
+	start_crio
+}
+
+function teardown() {
+	cleanup_test
+}
+
+@test "image fs info with default settings should return matching container_filesystem and image_filesystem" {
+	version=$(crictl --version)
+	echo "$version"
```
kwilczynski, line 17: To make sure that you get the correct binary, you can use `$CRICTL_BINARY` variable, or there is a function, a helper, called `crictl` too. Not sure if this would work for you?

The following is probably the same:

```suggestion
	crictl --version
```

That said, the version output won't be shown unless the test fails, I believe. You might need to save it to a file somewhere and then add `cat /tmp/debug.log` in the test runner script. Or something to that effect.

I wish there was a simple `log` or `debug` function available in bats.

test/imagefsinfo.bats
```diff
@@ -0,0 +1,39 @@
...
+
+	container_output=$(jq -e '.status.containerFilesystems[0]' <<< "$output")
+	image_output=$(jq -e '.status.imageFilesystems[0]' <<< "$output")
+	# if these are the same we can directly compare.
+	[ "$container_output" = "$image_output" ]
+}
+
+@test "image fs info with imagestore set should return different filesystems" {
+	version=$(crictl --version)
+	echo "$version"
```
kwilczynski, line 29: Same as earlier comment.

test/imagefsinfo.bats
saschagrunert: Do we need that `$version` at all?

test/imagefsinfo.bats
```diff
@@ -0,0 +1,39 @@
...
+
+	stop_crio
+	CONTAINER_IMAGESTORE="$TESTDIR/imagestore" start_crio
```
saschagrunert, line 32: I'd say we remove the `start_crio` from the setup and have no need to stop it again.

test/imagefsinfo.bats
kannon92: I'll just drop it.

test/imagefsinfo.bats
kannon92: I need to start it with a different parameter.

saschagrunert: I mean `stop_crio` would not be required if `setup()` would not container `start_crio`. The other test above can also run `start_crio` while this one calls `CONTAINER_IMAGESTORE="$TESTDIR/imagestore" start_crio`.

kannon92: Ah that is very clear now. Pushed a change for that.

saschagrunert (approved):

PR #7648, opened 2024-01-02
OCPBUGS-25926: mixedcpus: append the shared CPUs upon `UpdateContainerResourcesRequest`

Tal-or:
#### What type of PR is this?
/kind bug

After Kubelet restart, the CPU manager (CM) sends UpdateContainerResourcesRequest to all containers, to update their resources.

Since the CM is not aware of the shared CPUs concept, it sends only the guaranteed/isolated cpus. Then we end up with a container without the shared CPUs under its cgroup.

We should append the shared CPUs back upon UpdateContainerResourcesRequest if the container had them initially.

None

```release-note
None
```

### DIFF
--- a/server/container_update_resources.go
+++ b/server/container_update_resources.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/cri-o/cri-o/internal/config/node"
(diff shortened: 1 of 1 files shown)


haircommander:
/approve
LGTM
/cherry-pick release-1.29

Issue #7249, opened 2023-08-25
Deleted container never removed from crio_containers_oom_count_total


simonpasquier:
### What happened?

Using OCP (upcoming 4.14), I created a 10-replicas deployment with memory limits that trigger an immediate crash of the container. As expected the `crio_containers_oom_count_total` metric highlighted the problem.

After a while I've scaled down the deployment to 0 replicas.

### What did you expect to happen?

After scaling down the deployment, I expected the `crio_containers_oom_count_total` metric to disappear since all containers were now deleted.

### How can we reproduce it (as minimally and precisely as possible)?

Deployment example

```
kind: Deployment
apiVersion: apps/v1
metadata:
  name: hello-openshift
  namespace: test
spec:
  replicas: 10
  selector:
    matchLabels:
      app: hello-openshift
  template:
    metadata:
      labels:
        app: hello-openshift
    spec:
      containers:
        - name: nginx
          image: nginx
          resources:
            limits:
              memory: 20Mi
```

Deploy the manifest, query the `crio_containers_oom_count_total` metric in Prometheus and check that it returns a few series.

Then scale down the deployment to 0 replicas. Query again the  `crio_containers_oom_count_total` metric.

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
# paste output here
```

```console
$ kubectl version --output=json
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Additional environment details (AWS, VirtualBox, physical, etc.)

<details>

</details>

haircommander:
interesting idea. the current behavior is just a global counter counting up as time goes on. We theoretically could track which container did the OOM, and when the pod it belongs to is removed we could decrease the count, but that has some tracking overhead to figure out how many OOMs a pod had. also the name `oom_count_total` implies to me we're counting the total number of OOM kills on that particular node, which seems accurate to me

simonpasquier:
IIUC `crio_containers_oom_count_total` counts the number of ooms per `node` and `container`. 

https://gist.github.com/simonpasquier/2acfdd405cb42a7d29c917b95e08f552

`container_runtime_crio_containers_oom_total` is almost the same but only partioned by `node`.

https://gist.github.com/simonpasquier/4100eae5dade254f6bce66bd7cfbba36

[...]

PR #7656, opened 2024-01-04
metrics: remove deleted containers from OOM count metrics

haircommander:
to prevent an ever growing vector and cardinality

#### What type of PR is this?

/kind bug

```release-note
cleanup deleted container from `container_runtime_crio_containers_oom` metric, preventing leaks and high cardinality
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -905,6 +905,11 @@ func (r *runtimeOCI) DeleteContainer(ctx context.Context, c *Container) error {
 		return nil
 	}
 
+	if c.state.OOMKilled {
+		// Collect metric by container name
+		metrics.Instance().MetricContainersOOMCountTotalDelete(c.Name())
+	}
+
 	_, err := r.runtimeCmd("delete", "--force", c.ID())
 	return err
 }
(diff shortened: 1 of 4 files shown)


haircommander:
fixes https://github.com/cri-o/cri-o/issues/7249

internal/oci/runtime_oci.go
```diff
@@ -905,6 +905,11 @@ func (r *runtimeOCI) DeleteContainer(ctx context.Context, c *Container) error {
 		return nil
 	}
 
+	if c.state.OOMKilled {
```
simonpasquier, line 908: You could remove the if branch since DeleteLabelValues() should be a no-op when the label value doesn't exist.

haircommander: will it skim through all the values to check though? I figured we could avoid the lookup because we already know whether it's present or not

simonpasquier: I don't know in depth the internals of prometheus/client_golang but IIUC a MetricVec indexes series by their label hash so the lookup should be fast (https://github.com/prometheus/client_golang/blob/046e3203b61d1b5000b162e210a91975f098d742/prometheus/vec.go#L319).

haircommander: thanks for the suggestion, I think I prefer checking the bool as it'll be the fastest and doesn't increase the complexity too much

haircommander:
@cri-o/cri-o-maintainers PTAL

sohankunkerkar (approved):
/lgtm

Issue #6747, opened 2023-03-20
No metrics reported from CRI-O: Unable to account working set stats: total_inactive_file > memory usage


0xC0ncord:
### What happened?

On a 3-node cluster deployed with kubeadm and using CRI-O, I am seeing a situation where after some amount of workload is allocated to a node, the node stops reporting any metrics (CPU/memory usage is 0% or `<unknown>` in `kubectl top nodes`). In the system journal for CRI-O, I can see these warnings being emitted every second or so on the affected nodes:
```
Mar 18 16:09:13 ram.fuwafuwatime.moe crio[72618]: time="2023-03-18 16:09:13.067279194-04:00" level=warning msg="Unable to account working set stats: total_inactive_file (3827683328) > memory usage (3233787904)"
Mar 18 16:09:18 ram.fuwafuwatime.moe crio[72618]: time="2023-03-18 16:09:18.919610520-04:00" level=warning msg="Unable to account working set stats: total_inactive_file (3827683328) > memory usage (3233787904)"
Mar 18 16:09:24 ram.fuwafuwatime.moe crio[72618]: time="2023-03-18 16:09:24.932763766-04:00" level=warning msg="Unable to account working set stats: total_inactive_file (3827683328) > memory usage (3233746944)"
Mar 18 16:09:31 ram.fuwafuwatime.moe crio[72618]: time="2023-03-18 16:09:31.687808594-04:00" level=warning msg="Unable to account working set stats: total_inactive_file (3827683328) > memory usage (3233759232)"
```

If I execute `kubectl top nodes` on the cluster, I get this output:
```
NAME                         CPU(cores)   CPU%        MEMORY(bytes)   MEMORY%     
frederica.fuwafuwatime.moe   1867m        23%         17545Mi         55%         
ram.fuwafuwatime.moe         <unknown>    <unknown>   <unknown>       <unknown>   
rem.fuwafuwatime.moe         <unknown>    <unknown>   <unknown>       <unknown>   
```

And in `k9s`, the pods running on these nodes report no CPU/memory usage (0%):
![image](https://user-images.githubusercontent.com/8011001/226431622-f306fb9f-5628-499f-adeb-783945429d7c.png)

### What did you expect to happen?

Metrics should be visible from these nodes, regardless of workload.

### How can we reproduce it (as minimally and precisely as possible)?

This is what I have done:

[...]

PR #7658, opened 2024-01-05
Collect container and Sandbox stats through libctr cgroup managers

Mo-Fatah:
#### What type of PR is this?

/kind feature
stats server currently manually collects stats through the `Populate**` functions in cgmgr package which doesn't report correct stats. This PR bases the stats collection on libcontainer cgroup managers.

```release-note
Use libcontainer cgroup manager to collect container and Sandbox stats
```


haircommander:
/ok-to-test
/approve

LGTM

sohankunkerkar:
@Mo-Fatah do you mind pasting the stats o/p with this change? I just wanted to see the actual stats.

Mo-Fatah:
Sure! @sohankunkerkar 
I built crio from the current branch, ran a pod, ran a sleep container, finally used the command `crictl statsp -o json`

stats from the current changes:
```
...
           "cpu": {
              "timestamp": "1704727367994867370",
              "usageCoreNanoSeconds": {
                "value": "39525000"
              },
              "usageNanoCores": {
                "value": "0"
              }
            },
            "memory": {
              "timestamp": "1704727367994867370",
              "workingSetBytes": {
                "value": "839680"
              },
              "availableBytes": {
                "value": "265420800"
              },
              "usageBytes": {
                "value": "3014656"
              },
              "rssBytes": {
                "value": "98304"
              },
              "pageFaults": {
                "value": "2561"
              },
              "majorPageFaults": {
                "value": "17"
              }
            },
            "writableLayer": {
              "timestamp": "1704727367994869583",
              "fsId": {
                "mountpoint": "/var/lib/containers/storage/overlay/833e7fe8b7b3de29c496abea65b7260059af7cf54aa97d8f42bf519cd70b28f1/merged"
              },
              "usedBytes": {
                "value": "0"
              },
              "inodesUsed": {
                "value": "4"
              }
            },
            "swap": {
              "timestamp": "1704727367994867370",
              "swapAvailableBytes": {
                "value": "265420800"
              },
              "swapUsageBytes": {
                "value": "3014656"
              }
            }
...
```

doing the same steps with the main branch:
```
...
            "cpu": {
              "timestamp": "1704727905610530972",
              "usageCoreNanoSeconds": {
                "value": "35000000"
              },
              "usageNanoCores": null
            },
            "memory": {
              "timestamp": "1704727905610530972",
              "workingSetBytes": {
                "value": "102400"
              },
              "availableBytes": {
                "value": "18446744073441218560"
              },
              "usageBytes": {
                "value": "102400"
              },
              "rssBytes": {
                "value": "0"
              },
              "pageFaults": {
                "value": "2541"
              },
              "majorPageFaults": {
                "value": "0"
              }
            },
            "writableLayer": {
              "timestamp": "1704727905610609128",
              "fsId": {
                "mountpoint": "/var/lib/containers/storage/overlay/12351a22aa4c2f4339754a2e8d670f17b3dc00aeebf903a8c9b7c77a95dd4466/merged"
              },
              "usedBytes": {
                "value": "0"
              },
              "inodesUsed": {
                "value": "4"
              }
            },
            "swap": null
...
```
Notable changes in this PR:
- adding the `swap` values. (it was neglected before)
- some memory values are present (e.g. RSS)
- The numbers should be a bit different since the stats reported from libcontainer uses different values from the `containers/common` package (e.g. in the memory/usageBytes value: `containers/common` uses `memory.stats:anon` value from the cgroup while libcontainer uses `memory.current` which should be the correct value.

kwilczynski:
@Mo-Fatah, this value (an excerpt from your example):

```json
              "availableBytes": {
                "value": "18446744073441218560"
              },
```

That is about 18 Exabytes. if I am not mistaken :smile: How does this sort of value work with metrics?

Mo-Fatah:
@kwilczynski  I am not sure how this huge number is generated :sweat_smile: . However, the `availableBytes` isn't related to libcontainer because we calculate it ourselves in the code. k8s calculate it by `memory limit - memory usage ( or workingset)` 
https://github.com/kubernetes/kubernetes/blob/94f15bbbcbe952762b7f5e6e3f77d86ecec7d7c2/pkg/kubelet/stats/helper.go#L69

which is currently miscalculated on main branch in cri-o
https://github.com/cri-o/cri-o/blob/main/internal/config/cgmgr/stats_linux.go#L73

we calculate it the opposite order, and usually the memory limit > usage. So, this is my guess why we get this huge number.

haircommander:
> @kwilczynski I am not sure how this huge number is generated 😅 . However, the `availableBytes` isn't related to libcontainer because we calculate it ourselves in the code. k8s calculate it by `memory limit - memory usage ( or workingset)` https://github.com/kubernetes/kubernetes/blob/94f15bbbcbe952762b7f5e6e3f77d86ecec7d7c2/pkg/kubelet/stats/helper.go#L69
> 
> which is currently miscalculated on the main branch in cri-o https://github.com/cri-o/cri-o/blob/main/internal/config/cgmgr/stats_linux.go#L73
> 
> cri-o calculates it the opposite order, and usually the memory limit > usage. So, this is my guess why we get this huge number.

whoops, this probably was me a couple of years ago. I wonder if this causes https://github.com/cri-o/cri-o/issues/6747 as well

Mo-Fatah:
> > @kwilczynski I am not sure how this huge number is generated 😅 . However, the `availableBytes` isn't related to libcontainer because we calculate it ourselves in the code. k8s calculate it by `memory limit - memory usage ( or workingset)` https://github.com/kubernetes/kubernetes/blob/94f15bbbcbe952762b7f5e6e3f77d86ecec7d7c2/pkg/kubelet/stats/helper.go#L69
> > which is currently miscalculated on the main branch in cri-o https://github.com/cri-o/cri-o/blob/main/internal/config/cgmgr/stats_linux.go#L73
> > cri-o calculates it the opposite order, and usually the memory limit > usage. So, this is my guess why we get this huge number.
> 
> whoops, this probably was me a couple of years ago. I wonder if this causes #6747 as well

I guess so, will have a look at it tomorrow.

Mo-Fatah:
I couldn't reproduce the failures on my machine. They also seem to be persistently failing, don't think it's a flakiness issue

littlejawa:
> @littlejawa can you PTAL?

With kata, the error seems to come from :

```
# time="2024-01-11 14:34:09.959493738Z" level=error msg="Error getting container stats e53309b9b475e2fcd9b0b1f5691b1c9ee3a3a913e25ae46014d91c48d5b3f1fa: error while statting cgroup v2: [openat2 /sys/fs/cgroup/pod_123.slice/pod_123-456.slice/crio-e53309b9b475e2fcd9b0b1f5691b1c9ee3a3a913e25ae46014d91c48d5b3f1fa.scope/cgroup.procs: no such file or directory]" file="stats/stats_server_linux.go:119"
```

As I understand it, libcontainer tries to access the cgroup information from the host, but with kata the container's cgroup information is within a VM, not accessible to the host.

The way we do it today is through the call to 
`cStats, err := ss.Runtime().ContainerStats(context.TODO(), c, sb.CgroupParent())`
that you removed.

It allows a different behaviour between runtimes: for regular (oci) runtimes, it is conlcuded by a call to [PopulateContainerCgroupStats](https://github.com/mo-fatah/cri-o/blob/af3be187fa1101728d281ac3c30501b81d4c4280/internal/oci/oci_linux.go#L92).
For kata containers, it is done by an exchange with the kata agent within the VM, which returns the expected values. This is done in [ContainerStats](https://github.com/mo-fatah/cri-o/blob/af3be187fa1101728d281ac3c30501b81d4c4280/internal/oci/runtime_vm.go#L944-L945)

I'm wondering if we could keep this indirection: keep calling ss.Runtime().ContainerStats(), and modify the oci part to call cgroupStats() and return the stats in cgroup format? The rest of you code (converting from cgroup to CRI) wouldn't change.

I would need to change the kata-specific part to return cgroup data too.
I will need to do some conversion, as the VM cgroup version may differ from the host's. Today I'm converting it to CRI anyway, but if we expect cgroup stats, I'll need to make sure I provide the right version of it.
But returning cgroup stats seems right here, as the conversion to CRI would then be centralized for all runtimes.

Do you think that would work, or do you have other suggestions?

kwilczynski:
Aside from the bug that we had for a while, and which @Mo-Fatah, fixed (thank you!). We need to make sure that these new metrics hold water.

@Mo-Fatah, were you able to compare these metrics with what `/proc/<PID>/statm` and `/proc/<PID>/status` shows?

For instance, from your example, both of these values are the same:

```json
"availableBytes": {
  "value": "265420800"
},
```

```json
"swapAvailableBytes": {
  "value": "265420800"
},
```

A problem here or merely a happy coincidence? :smile: If there is no swap, does it show as a value of `0` or a `null` in JSON?

Basically, everything looks very good, and it's solid work, but we need to make some sanity checks to see if things are what they say they are and that values also make sense for certain metrics.

Generally, these metrics are such that process, a CRI-O process, I suppose, metrics are stored alongside what seems to be system-wide metrics.

```json
"memory": {
  "timestamp": "1704727367994867370",
  "workingSetBytes": {
    "value": "839680"
  },
  "availableBytes": {
    "value": "265420800"
  },
  "usageBytes": {
    "value": "3014656"
  },
  "rssBytes": {
    "value": "98304"
  },
  "pageFaults": {
    "value": "2561"
  },
  "majorPageFaults": {
    "value": "17"
  }
},
```

For example, within the above, we don't have virtual address space allocated for the CRI-O process, which is also useful to track, even though we have `workingSetBytes` that most likely include anonymous and virtual address space too.

So, basically, verify some things, make sure that everything looks reasonable, check the Prometheus labels, etc.

Mo-Fatah:
Thanks for the details! @littlejawa . Your approach seems reasonable to me. 
- for oci I will use `cStats, err := ss.Runtime().ContainerStats(context.TODO(), c, sb.CgroupParent())` and will change `ContainerStats` internally to return the new [`CgroupStats`](https://github.com/cri-o/cri-o/pull/7658/files#diff-b9141e04c913bc0a793240448278d523cd5b1bd22cf036b9a0f3bcf1e32113c3R22) type.
- for kata (vm runtime), will change the `ContainerStats` function to return `CgroupStats` instead of CRI stats by converting the metrics into `CgroupStats`. I see that we already have [conversion](https://github.com/cri-o/cri-o/blob/main/internal/oci/runtime_vm.go#L972-L976) functions, so I will edit those. 

the conversion to CRI finally will be remain in the stats server. (correct me if i lost my way)

Mo-Fatah:
@kwilczynski 
> "availableBytes": {
  "value": "265420800"
},
"swapAvailableBytes": {
  "value": "265420800"
},
A problem here or merely a happy coincidence? 😄 If there is no swap, does it show as a value of 0 or a null in JSON?

Not a coincidence. The reason why  `availableBytes` = `swapAvailableBytes` and `usageBytes` = `swapUsageBytes` is due to libcontainer's choice of being compatible with cgroupv1 (the numbers I provided are by cgroupv2) https://github.com/opencontainers/runc/blob/main/libcontainer/cgroups/fs2/memory.go#L122-L128 They add the the memory availableBytes , usageBytes to swapAvailableBytes, swapUsageBytes respectively. 
In the stats i provided, the swap values are 0, therefore their values are equal to the memory ones.

> were you able to compare these metrics with what /proc/<PID>/statm and /proc/<PID>/status shows?
So, basically, verify some things, make sure that everything looks reasonable, check the Prometheus labels, etc.

will have a look again at the numbers and verify them

haircommander:
> For example, within the above, we don't have virtual address space allocated for the CRI-O process, which is also useful to track, even though we have workingSetBytes that most likely include anonymous and virtual address space too.

yeah unfortunately for now we can't fuss with the metrics too much. We basically need 1:1 metrics coverage between what currently exists and what cri-o will collect. In the future we can investigate adding additional ones that better reflect the state of the processes

also note: these are for pods and containers, so the address space of cri-o itself is not relevant here

kwilczynski:
> > "availableBytes": {
> > "value": "265420800"
> > },
> > "swapAvailableBytes": {
> > "value": "265420800"
> > },
> > A problem here or merely a happy coincidence? 😄 If there is no swap, does it show as a value of 0 or a null in JSON?
> 
> Not a coincidence. The reason why `availableBytes` = `swapAvailableBytes` and `usageBytes` = `swapUsageBytes` is due to libcontainer's choice of being compatible with cgroupv1 (the numbers I provided are by cgroupv2) https://github.com/opencontainers/runc/blob/main/libcontainer/cgroups/fs2/memory.go#L122-L128 They add the the memory availableBytes , usageBytes to swapAvailableBytes, swapUsageBytes respectively. In the stats i provided, the swap values are 0, therefore their values are equal to the memory ones.

Got it. So, the question is: would this be a breaking change? Or should we fix it to report a value of `0` as before, or should we simply exclude some fields from being reported when we are dealing with cgroups v1, still?

I personally would rather have `0` or no value, rather than a value that does not make sense. As in, if there is no support for something, then no value should be reported for such a field.

Unless our goal was also to make it report something in such a case.

Thoughts?

[...]
> > For example, within the above, we don't have virtual address space allocated for the CRI-O process, which is also useful to track, even though we have workingSetBytes that most likely include anonymous and virtual address space too.
> 
> Can you elaborate that part? I mean, those stats are part of the cri-api

Feel free to ignore this part. @haircommander added more context in a different comment.

haircommander:
> Got it. So, the question is: would this be a breaking change? Or should we fix it to report a value of 0 as before, or should we simply exclude some fields from being reported when we are dealing with cgroups v1, still?

yeah I'd consider it breaking. we'd have to introduce a new metrics field and then deprecate the old. Unfortunately, many alerts are built off of the existing metric names so we need to consider them a stable API and undergo a long deprecation process

sohankunkerkar:
Hey @Mo-Fatah once you mark this PR ready for review, could you also squash the redundant commits and place them in the meaningful ones?

Mo-Fatah:
@sohankunkerkar  I did some squash using github desktop ( wanted to do some in-between commits squash, thought it would be easier to use the GUI app) . It seems all my commits are now `Unverified`. should this be fixed or not a problem ?

internal/config/cgmgr/stats_test.go
```diff
@@ -1,91 +0,0 @@
-package cgmgr_test
```
Mo-Fatah, line 1: This file just tests UpdateWithMemoryStatsFromFile function which is deleted and no longer used

internal/lib/stats/stats_server_linux.go
```diff
@@ -40,11 +47,12 @@ func (ss *StatsServer) updateSandbox(sb *sandbox.Sandbox) *types.PodSandboxStats
 		if c.StateNoLock().Status == oci.ContainerStateStopped {
 			continue
 		}
-		cStats, err := ss.Runtime().ContainerStats(context.TODO(), c, sb.CgroupParent())
+		cgstats, err := ss.Runtime().ContainerStats(context.TODO(), c, sb.CgroupParent())
```
littlejawa, line 50: There is a problem here: you're expecting `ContainerStats()` to return a `CgroupStats` structure, but you're modifying `ContainerStats` only in the next commit.
Your first commit then doesn't build - you get an error when you try to use `cgstats` as a parameter to `containerCRIStats()`

```
internal/lib/stats/stats_server_linux.go:55:31: cannot use cgstats (variable of type *"k8s.io/cri-api/pkg/apis/runtime/v1".ContainerStats) as *cgmgr.CgroupStats value in argument to containerCRIStats
```

If you don't want to merge the two commits (which would make it pretty big), maybe you could introduce that function code changes only in the second commit?
Not sure what option is best, but I think we need to have independently buildable commits

internal/config/cgmgr/stats_linux.go
```diff
@@ -20,22 +20,49 @@ import (
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
-// CgroupStats is a structure used to augment libctr.Stats object, because
-// it does not have all of the information required for all of the stats we're
-// interested in.
+// This is a universal stats object to be used across different runtime implementations.
```
haircommander, line 15: I find myself worried about the translation here. there is already high cpu usage on stats collection from golang object GC (each tiny object is created and destroyed quite quickly). I think I am okay with this for now but it's possible we may want to find a different way. For instance, having the statsserver run linux only for now

haircommander: If you'd rather do that as a follow up (or not deal with that and have someone else do it) @Mo-Fatah let me know

internal/lib/stats/stats_server_linux.go
Mo-Fatah: I would go with merging the two commits if no problem with that. it will be a big commit but I guess all changes are relevant. ( will make the commit message a bit meaningful)

internal/config/cgmgr/stats_linux.go
Mo-Fatah: I personally wanted to go the way Sohan did before [here](https://github.com/cri-o/cri-o/pull/7256/files#diff-d9db4f1729b93e9ec3b256488380d0ee3b8b06756cfc30d406b9de6976c30779R22-R25) to just use the , but due to the newly-added FreeBSD build workflow, apparently any inclusion of libcontainer in runtime_vm.go (through importing libcontainer or adding a type that uses a libcontainer type internally) will lead to FreeBSD build failure (at least this what I noticed from multiple CI failures).

 I would prefer to solve this in a separate PR if it's Ok.

sohankunkerkar: >I would prefer to solve this in a separate PR if it's Ok.
That sounds good!

sohankunkerkar:
/lgtm
/hold 
@cri-o/cri-o-maintainers  Please feel free unhold if everything looks good.

haircommander:
/hold cancel
/approve

well done @Mo-Fatah !

PR #6022, opened 2022-06-30
do not wipe images by default

haircommander:
#### What type of PR is this?

/kind bug
#### What this PR does / why we need it:
many users don't want the crio wipe functionality that removes images on reboot. Update internal wipe to not do so by default
#### Does this PR introduce a user-facing change?

```release-note
Update `version-file-persist` to be empty by default, and do not wipe images when it is so.
```

PR #7666, opened 2024-01-09
wrong version-file-persist value

lance5890:
#### What type of PR is this?
/kind bug

wrong parameter

```release-note
None
```

### DIFF
--- a/internal/criocli/criocli.go
+++ b/internal/criocli/criocli.go
@@ -1092,7 +1092,7 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 		&cli.StringFlag{
 			Name:      "version-file-persist",
 			Usage:     "Location for CRI-O to lay down the persistent version file. It is used to check if crio wipe should wipe images, which should only happen when CRI-O has been upgraded.",
-			Value:     defConf.VersionFile,
+			Value:     defConf.VersionFilePersist,
 			EnvVars:   []string{"CONTAINER_VERSION_FILE_PERSIST"},
 			TakesFile: true,
 		},
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -427,7 +427,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--version-file**="": Location for CRI-O to lay down the temporary version file. It is used to check if crio wipe should wipe containers, which should always happen on a node reboot. (default: "/var/run/crio/version")
 
-**--version-file-persist**="": Location for CRI-O to lay down the persistent version file. It is used to check if crio wipe should wipe images, which should only happen when CRI-O has been upgraded. (default: "/var/run/crio/version")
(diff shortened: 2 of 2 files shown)


kwilczynski:
@lance5890, thank you for the contribution!

Can you expand upon the "_Definitely the wrong parameter_" a little more? What is the problem you are trying to fix with the changes here?

lance5890:
> @lance5890, thank you for the contribution!
> 
> Can you expand upon the "_Definitely the wrong parameter_" a little more? What is the problem you are trying to fix with the changes here?

1 https://github.com/cri-o/cri-o/blob/521fdfb80a055d2d9d53f0649c9eee49e89cb7da/internal/criocli/criocli.go#L1092-L1097

the version-file-persist should use the defConf.VersionFilePersist, but not the defConf.VersionFile,

2 
image

the --version-file-persist default value is "/var/lib/crio/version"

sohankunkerkar:
@lance5890 Thanks for your PR. We removed [this](https://github.com/cri-o/cri-o/pull/6022) functionality to enable users to retain images on reboot.

internal/criocli/criocli.go
```diff
@@ -1092,7 +1092,7 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 		&cli.StringFlag{
 			Name:      "version-file-persist",
 			Usage:     "Location for CRI-O to lay down the persistent version file. It is used to check if crio wipe should wipe images, which should only happen when CRI-O has been upgraded.",
-			Value:     defConf.VersionFile,
+			Value:     defConf.VersionFilePersist,
```
haircommander, line 1095: yeah @sohankunkerkar is correct, though you correctly found an issue here.

haircommander: the rest should be "" though

lance5890: maybe we should set the --version-file-persist to "", but not the "/var/run/crio/version" which is shown to user

internal/criocli/criocli.go
```diff
@@ -1092,7 +1092,7 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 		&cli.StringFlag{
 			Name:      "version-file-persist",
 			Usage:     "Location for CRI-O to lay down the persistent version file. It is used to check if crio wipe should wipe images, which should only happen when CRI-O has been upgraded.",
-			Value:     defConf.VersionFile,
+			Value:     "",
```
haircommander, line 1095: I think this can stay defConf.VersionFilePersist, as it will future proof it. I think the documentation changes will remain

lance5890: done

haircommander:
hm docs validation is failing https://github.com/cri-o/cri-o/actions/runs/7482992377/job/20367580669?pr=7666

lance5890:
> hm docs validation is failing https://github.com/cri-o/cri-o/actions/runs/7482992377/job/20367580669?pr=7666

docs validation is ok now, I think the other failures has nothing to do with this PR

haircommander:
/ok-to-test
/approve 

LGTM, thanks @lance5890

PR #7715, opened 2024-01-29
OCPBUGS-28586: higeperfhooks: mixedcpus: set only exclusive cpus in child cgroup

Tal-or:
#### What type of PR is this?
/kind bug

When shared cpus are configured for the container, the child state object that represents that cpuset under the container's child cgroup is already exists under the managers slice.

This means, we should override last element instead of append a new one, because the child state is already there.

None

```release-note
None
```


haircommander:
/approve

Issue #4003, opened 2020-07-23
LFX 2024: Fully automated releases


saschagrunert:
<details>
<summary>Original issue description</summary>

Right now we have to do the following steps to cut a release:
1. Bump CRI-O's version in `internal/version/version.go`
2. Create a PR for that change and merge it
3. Create a release via the GitHub UI
4. Manually trigger CircleCI on the release branch
5. Update the release notes via the GitHub UI

To fully automate this process we could:
1. Add a new (golang based) script which:
    - Bumps the version automatically
    - Creates the PR automatically (via go-github and a set GITHUB_TOKEN env var)
2. Let GitHub actions *automatically* do the rest:
    - Tag the specified version
    - (Maybe bump the version again to *-dev?)
    - Create the GitHub release and push the changelog and artifacts

WDYT? I'm not sure if tagging the PR commit will cause issues in the future.

Refs: https://github.com/cri-o/cri-o/pull/4017

</details>

### Update

This topic is now part of the LFX mentoring program, where @olad5 will work on it

We had an onboarding call on March 12 (2024) which can be summarized as follows:

- Focus on the current state of the CI pipeline to automate the patch releases for now (the minor versions will be considered later)
- Use a GitHub action cron to propose a version bump [like this](https://github.com/cri-o/cri-o/pull/7810) once a month
- The PR needs to get manually approved by maintainers and merged by prow
- The merge will trigger the automatic release creation including the tag
- The tag will trigger the rest of the pipeline, [which is already in place](https://github.com/cri-o/cri-o/blob/ccc3e5b21399cab1e08ecc7482e772c997c745e6/.github/workflows/test.yml#L222-L239)

We probably can recycle code parts from (if not, then we may remove them):
- https://github.com/cri-o/cri-o/pull/7173
- https://github.com/cri-o/cri-o/tree/main/scripts/release

haircommander:
Another step further would be to build for respective distributions once the release is cut

I am certainly in favor of automating :smile:

haircommander:
we still want this!

anshgoyalevil:
@haircommander Can I start working on some part of this issue by creating child issues to it? I really want to work upon it under LFX Fall'23

haircommander:
good initiative! yeah please feel free to open child issues and begin breaking down the issue

anshgoyalevil:
@saschagrunert Can we put this project to GSoC 2024?

saschagrunert:
[...]

PR #7718, opened 2024-01-31
ci: add version-bump workflow

ipshitagoyal:
#### What type of PR is this?

/kind ci

This PR adds a Manual Trigger Workflow that would allow the maintainers to bump the version inside `internal/version/version.go` on the go by just running the workflow.

With this PR, the maintainers would not need to bump version in `internal/version/version.go` locally, and also doesn't require PRs to be created for the change.

Now simply run the workflow from the `Actions` tab on GitHub, and voila its all done for the version_bump side.

image

**What's next?**
This workflow will be used to trigger the automated release pipeline as described in issue https://github.com/cri-o/cri-o/issues/4003

part of: https://github.com/cri-o/cri-o/issues/4003

```release-note

```

### DIFF
--- a/.github/workflows/version_bump.yaml
+++ b/.github/workflows/version_bump.yaml
@@ -1,0 +1,32 @@
+name: Manual Version Bump
+
+on:
+  workflow_dispatch:
+    inputs:
+      bump_type:
+        description: 'Type of version bump (major, minor, patch)'
+        required: true
+
+jobs:
+  version_bump:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: 1.21
+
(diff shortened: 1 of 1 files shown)


ipshitagoyal:
//cc @haircommander @saschagrunert

kwilczynski:
@ipshitagoyal, thank you for working on this.

Even though the workflow idea is very nice, I believe the ultimate goal is to have everything automated. As such, there will be no maintainer or user interaction—no need to type or provide a new release version, etc.

So, a solution... whatever that will be... would prepare a new release and then also update the version number as part of that process.

saschagrunert:
@ipshitagoyal thank you for the PR! I would love to see this solution as part of the LFX program rather than tackling it here and now.

ipshitagoyal:
Thanks for your thoughts @kwilczynski and @saschagrunert 🙂 

I would close the PR, to keep it as a reference for anyone who would work on the project during LFX term.

PR #6153, opened 2022-08-18
oci: specify --systemd-cgroup for all runtime commands

haircommander:
#### What type of PR is this?

/kind bug
#### What this PR does / why we need it:
fixes https://github.com/cri-o/cri-o/issues/5490
#### Does this PR introduce a user-facing change?

```release-note
Fix an issue with leaked systemd scopes by always sending runc the `--systemd-cgroup` flag when using systemd cgroups
```

PR #7726, opened 2024-02-02
When crio restarts, restore the infraContainers

### LINKED ISSUE #5490: Error log: Failed to create existing container
**Description**  
I checked the similar issues #3259 and #4465 , but fail to fix with the workaround.  
My situation is that the error messages were triggered after restarting crio & kubelet.  
Though the error messages can disappear after reboot, is there any solution to this problem without rebooting?   

1. `systemctl stop kubelet`
2. `systemctl restart crio`
3. `systemctl start kubelet`
4. `journalctl -xeu kubelet -f`

I executed tests with different cgroup drivers, and I received errors and warnings below in kubelet logs every minute.  

#kubelet & crio cgroup driver use systemd:
```
kubelet[16645]: E1119 12:07:05.393398   16645 manager.go:1123] Failed to create existing container: /kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod1e900d5d_3559_4bc2_9b52_761eb5ed3c3f.slice/crio-217e7ff7799d62d02282afd1a0f2a8bbba85e62c8614ad4bf0aa9c029f2a0661.scope: Error finding container 217e7ff7799d62d02282afd1a0f2a8bbba85e62c8614ad4bf0aa9c029f2a0661: Status 404 returned error &{%!s(*http.body=&{0xc001297aa0 <nil> <nil> false false {0 0} false false false <nil>}) {%!s(int32=0) %!s(uint32=0)} %!s(bool=false) <nil> %!s(func(error) error=0x55fd88bf3020) %!s(func() error=0x55fd88bf2fa0)}
```

#kubelet & crio cgroup driver use cgroupfs:
```
kubelet[7608]: E1203 12:30:49.996168    7608 manager.go:1123] Failed to create existing container: /kubepods/burstable/pod5f65e3d8d489e8fea295a7cd01aff842/crio-a3e66d2af18c551b3a281b5369de361a19ed1ea1ed508563fbdf94366cb321eb: Error finding container a3e66d2af18c551b3a281b5369de361a19ed1ea1ed508563fbdf94366cb321eb: Status 404 returned error &{%!s(*http.body=&{0xc000b6cfd8 <nil> <nil> false false {0 0} false false false <nil>}) {%!s(int32=0) %!s(uint32=0)} %!s(bool=false) <nil> %!s(func(error) error=0x55fe06c23f40) %!s(func() error=0x55fe06c23ec0)}
```

In kubelet logs, no errors or warnings of "Failed to create existing container".

I received these logs every 60 seconds after restarting crio and kubelet.

```
crio version 1.22.1
Version:          1.22.1
GitCommit:        63ca93845d5fe05cdca826367afcb601ece8d7ad
GitTreeState:     clean
BuildDate:        2021-11-11T20:24:17Z
GoVersion:        go1.16.8
Compiler:         gc
Platform:         linux/amd64
Linkmode:         dynamic
BuildTags:        exclude_graphdriver_devicemapper, seccomp
SeccompEnabled:   true
AppArmorEnabled:  false
```

OS: Centos 7  
Kubernetes v1.22.1  

crio configuration:
```
...
[crio.runtime]
conmon_cgroup = "system.slice"
cgroup_manager = "systemd"
...
```

Kubelet configuration:
```
apiVersion: kubelet.config.k8s.io/v1beta1
authentication:
  anonymous:
    enabled: false
  webhook:
    cacheTTL: 0s
    enabled: true
  x509:
    clientCAFile: /etc/kubernetes/pki/ca.crt
authorization:
  mode: Webhook
  webhook:
    cacheAuthorizedTTL: 0s
    cacheUnauthorizedTTL: 0s
cgroupDriver: systemd
clusterDNS:
- 10.96.0.10
clusterDomain: cluster.local
cpuManagerPolicy: static
cpuManagerReconcilePeriod: 5s
evictionPressureTransitionPeriod: 0s
fileCheckFrequency: 0s
healthzBindAddress: 127.0.0.1
healthzPort: 10248
httpCheckFrequency: 0s
imageMinimumGCAge: 0s
kind: KubeletConfiguration
kubeReserved:
  cpu: 500m
kubeletCgroups: /systemd/system.slice
logging: {}
memorySwap: {}
nodeStatusReportFrequency: 0s
nodeStatusUpdateFrequency: 0s
rotateCertificates: true
runtimeRequestTimeout: 0s
shutdownGracePeriod: 0s
shutdownGracePeriodCriticalPods: 0s
staticPodPath: /etc/kubernetes/manifests
streamingConnectionIdleTimeout: 0s
syncFrequency: 0s
volumeStatsAggPeriod: 0s
```

dsxing:
#### What type of PR is this?

/kind bug

Although there is a PR #6153 to fix issue #5490, but the problem still persists (see: https://github.com/cri-o/cri-o/issues/5490#issuecomment-1403716666). The root cause of the issue is that after crio restarts, the data for infraContainers is lost as  it is stored in memory(https://github.com/cri-o/cri-o/blob/main/internal/lib/container_server.go#L585). During the restore process, the sandbox and containers are reconstructed and restored through **LoadSandbox** and **LoadContainer** in restore funtion(https://github.com/cri-o/cri-o/blob/main/server/server.go#L222-L279), but the infraContainers are not restored during **LoadSandbox**.

 As a result, when kubelet cadvisor queries and retrieves infraContainer information, it returns an **errCtrNotFound** error(https://github.com/cri-o/cri-o/blob/main/server/inspect.go#L68), causing kubelet cadvisor to continuously log error messages.

If we rebuild and restore the infraContainers data when restarting crio, which is necessary, it won't cause cadvisor to fail to retrieve data and, consequently, won't lead to print error log

fixes #5490

None

```release-note
Restore infra containers state on CRI-O restart. Without this, the infra containers will be accounted as missing, leading to a spurious error message.
```

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -319,6 +319,9 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
 		return sb, err
 	}
 
+	// We should restore the infraContainer to the container state store
+	c.AddInfraContainer(ctx, scontainer)
+
 	sb.RestoreStopped()
 	// We add an NS only if we can load a permanent one.
 	// Otherwise, the sandbox will live in the host namespace.
--- a/internal/lib/container_server_test.go
+++ b/internal/lib/container_server_test.go
@@ -157,6 +157,19 @@ var _ = t.Describe("ContainerServer", func() {
 			Expect(err).To(BeNil())
 		})
 
+		It("should succeed load infraContainer", func() {
+			// Given
+			createDummyState()
+			mockDirs(testManifest)
+
+			// When
+			_, err := sut.LoadSandbox(context.Background(), "id")
+
+			// Then
+			Expect(err).To(BeNil())
+			Expect(sut.GetInfraContainer(context.Background(), sandboxID)).NotTo(BeNil())
+		})
+
 		It("should succeed with invalid network namespace", func() {
 			// Given
 			createDummyState()


internal/lib/container_server.go
```diff
@@ -319,6 +319,9 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
 		return sb, err
 	}
 
+	// We should restore the infraContainer to the container state store
+	c.AddInfraContainer(ctx, scontainer)
```
sohankunkerkar, line 323: Is it feasible to add test coverage for this change?

sohankunkerkar:
We might need a release note if it's a bug.

internal/lib/container_server.go
kwilczynski: @sohankunkerkar, I was wondering about this.

This seems like an obvious thing to do. So I wonder why we didn't in the past? Perhaps indeed a bug.

kwilczynski: Obvious thing to do, as in, load infra containers on restart.

kwilczynski:
/release-note-edit ```release-note Restore infra containers state on CRI-O restart. Without this, the infra containers will be accounted as missing, leading to a spurious error message.```

haircommander:
wow! great investigation and find!
/approve

haircommander:
one day, I'd love to consolidate the restore and creation flow so there's no way to miss pieces like this...

haircommander:
we should backport it to all supported branches

internal/lib/container_server.go
dsxing: > Is it feasible to add test coverage for this change?

can we add a unit test to check if the infraContainer is loaded?  like this:
https://github.com/cri-o/cri-o/pull/7726/files#diff-5df3a97ff31edf6084a160b044038d23d5ef6d2ab76f5ad0e5b57d059c37bdbcR160

dsxing:
> unit test seems to fail https://github.com/cri-o/cri-o/actions/runs/7778795849/job/21227832077?pr=7726

I made a small mistake in the unit test code which I added. It should query infraContainer with **sandboxId**. It's now fixed.

haircommander:
/retest
/approve

LGTM, thanks for finding this @dsxing

PR #7735, opened 2024-02-05
oci: handle early exited container faster in stop loop

haircommander:
if we hit a TOCTOU race with c.Living() and StopLoopForContainer, then we will be forced to wait until the kubelet re-requests. instead, update watchers immediately so this doesn't happen

#### What type of PR is this?

/kind bug

```release-note
fix a bug where a container that exits soon after a StopContainer request is issued will be paused in termination for longer than the grace period
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -594,6 +594,17 @@ func (c *Container) WaitOnStopTimeout(ctx context.Context, timeout int64) {
 	}
 }
 
+func (c *Container) SetAsDoneStopping() {
+	c.stopLock.Lock()
+	for _, watcher := range c.stopWatchers {
+		close(watcher)
+	}
+	c.stopWatchers = make([]chan struct{}, 0)
+	c.stopping = false
+	close(c.stopTimeoutChan)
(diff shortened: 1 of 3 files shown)


sohankunkerkar (approved):
/retest
/lgtm

sohankunkerkar:
/cherry-pick release-1.27
//cherry-pick release-1.26
//cherry-pick release-1.25

kwilczynski:
@sohankunkerkar, you might need to kick off `release-1.26` and `release-1.25` again. I think it didn't pick on the edit to correct the slashes situation. :smile:

kwilczynski:
@sohankunkerkar, did someone ask for 1.24?

I think, it might be safe to stop at 1.25 at this point. We should be out of support for 1.24 since 1.29 came to be.

Also, among some of the users of 1.24 and 1.25, the OpenShift 4.10 and 4.11 is also officially out of support.

sohankunkerkar:
Yeah, I realized that. I was under the impression that version 1.24 of cri-o is included in OCP 4.12. Thanks for the pointers.

PR #7736, opened 2024-02-05
OCPBUGS-28981: fix busyloop when killing container

rphillips:
This fixes the busy loop on a failed kill (D-state process) and reschedules the timer to sometime in the future. Since targetTime has expired, the loop re-enters the kill case statement without waiting. The pod is terminal at this point, so a 10ms pause seems ok.

```release-note
Fix busy loop on killing a D-state container
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -31,10 +31,10 @@ import (
 const (
 	defaultStopSignalInt = 15
 	// the following values can be verified here: https://man7.org/linux/man-pages/man5/proc.5.html
+	// the 3rd field is the process state
+	statStateLocation = 3
 	// the 22nd field is the process starttime
 	statStartTimeLocation = 22
-	// The 2nd field is the command, wrapped by ()
-	statCommField = 2
 )
 
 var (
@@ -42,6 +42,11 @@ var (
 	ErrNotFound         = errors.New("container process not found")
 	ErrNotInitialized   = errors.New("container PID not initialized")
 )
+
+type StatData struct {
+	StartTime string
+	State     string
+}
 
 // Container represents a runtime container.
 type Container struct {
@@ -292,11 +297,11 @@ func (cstate *ContainerState) SetInitPid(pid int) error {
 		return fmt.Errorf("pid and start time already initialized: %d %s", cstate.InitPid, cstate.InitStartTime)
 	}
 	cstate.InitPid = pid
-	startTime, err := getPidStartTime(pid)
+	data, err := getPidData(pid)
 	if err != nil {
 		return err
 	}
-	cstate.InitStartTime = startTime
+	cstate.InitStartTime = data.StartTime
 	return nil
 }
 
@@ -527,17 +532,24 @@ func (c *Container) pid() (int, error) {
 // This is the simplest way to verify we are operating on the container
 // process, and haven't run into PID wrap.
 func (c *Container) verifyPid() error {
-	startTime, err := getPidStartTime(c.state.InitPid)
+	data, err := getPidData(c.state.InitPid)
 	if err != nil {
 		return err
 	}
 
-	if startTime != c.state.InitStartTime {
+	if data.StartTime != c.state.InitStartTime {
 		return fmt.Errorf(
 			"PID %d is running but has start time of %s, whereas the saved start time is %s. PID wrap may have occurred",
-			c.state.InitPid, startTime, c.state.InitStartTime,
+			c.state.InitPid, data.StartTime, c.state.InitStartTime,
 		)
 	}
+
+	// Skip states which are typically considered as invalid.
+	switch data.State {
+	case "D", "X", "Z":
+		return fmt.Errorf("PID %d has an invalid state: %s", c.state.InitPid, data.State)
+	}
+
 	return nil
 }
 
--- a/internal/oci/container_freebsd.go
+++ b/internal/oci/container_freebsd.go
@@ -29,3 +29,12 @@ func getPidStartTime(pid int) (string, error) {
 	kp := (*C.struct_kinfo_proc)(unsafe.Pointer(&data[0]))
 	return fmt.Sprintf("%d,%d", kp.ki_start.tv_sec, kp.ki_start.tv_usec), nil
 }
+
+// getPidData reads the kernel's /proc entry for various data.
+func getPidData(pid int) (*StatData, error) {
+	startData := getPidStartTime(pid)
+	return &StatData{
+		StartTime: startTime,
+		State:     "not implemented",
+	}, nil
+}
--- a/internal/oci/container_freebsd_nocgo.go
+++ b/internal/oci/container_freebsd_nocgo.go
@@ -18,3 +18,15 @@ func getPidStartTime(pid int) (string, error) {
 	fields := strings.Fields(string(data))
 	return fields[7], nil
 }
+
+// getPidData reads the kernel's /proc entry for various data.
+func getPidData(pid int) (*StatData, error) {
+	startTime, err := getPidStartTime(pid)
+	if err != nil {
+		return nil, err
+	}
+	return &StatData{
+		StartTime: startTime,
+		State:     "not implemented",
+	}, nil
+}
--- a/internal/oci/container_linux.go
+++ b/internal/oci/container_linux.go
@@ -2,6 +2,7 @@ package oci
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 
@@ -41,40 +42,47 @@ func (c *Container) SeccompProfilePath() string {
 	return c.seccompProfilePath
 }
 
-// getPidStartTime reads the kernel's /proc entry for stime for PID.
+// getPidData reads the kernel's /proc entry for various data.
 // inspiration for this function came from https://github.com/containers/psgo/blob/master/internal/proc/stat.go
 // some credit goes to the psgo authors
-func getPidStartTime(pid int) (string, error) {
-	return GetPidStartTimeFromFile(fmt.Sprintf("/proc/%d/stat", pid))
+func getPidData(pid int) (*StatData, error) {
+	return getPidStatData(fmt.Sprintf("/proc/%d/stat", pid))
 }
 
-// GetPidStartTime reads a file as if it were a /proc/$pid/stat file, looking for stime for PID.
+// GetPidStartTimeFromFile reads a file as if it were a /proc/$pid/stat file, looking for stime for PID.
 // It is abstracted out to allow for unit testing
 func GetPidStartTimeFromFile(file string) (string, error) {
+	data, err := getPidStatData(file)
+	if err != nil {
+		return "", err
+	}
+	return data.StartTime, nil
+}
+
+func getPidStatData(file string) (*StatData, error) {
 	data, err := os.ReadFile(file)
 	if err != nil {
-		return "", fmt.Errorf("%v: %w", err, ErrNotFound)
+		return nil, fmt.Errorf("%v: %w", err, ErrNotFound)
 	}
 	// The command (2nd field) can have spaces, but is wrapped in ()
 	// first, trim it
 	commEnd := bytes.LastIndexByte(data, ')')
 	if commEnd == -1 {
-		return "", fmt.Errorf("unable to find ')' in stat file: %w", ErrNotFound)
+		return nil, fmt.Errorf("unable to find ')' in stat file: %w", ErrNotFound)
 	}
 
-	// start on the space after the command
+	// skip space after the command
 	iter := commEnd + 1
-	// for the number of fields between command and stime, trim the beginning word
-	for field := 0; field < statStartTimeLocation-statCommField; field++ {
-		// trim from the beginning to the character after the last space
-		data = data[iter+1:]
-		// find the next space
-		iter = bytes.IndexByte(data, ' ')
-		if iter == -1 {
-			return "", fmt.Errorf("invalid number of entries found in stat file %s: %d: %w", file, field-1, ErrNotFound)
-		}
+	// skip the character after the space
+	iter++
+
+	fields := bytes.Fields(data[iter:])
+	if len(fields) <= statStartTimeLocation {
+		return nil, errors.New("unable to parse stat file")
 	}
 
-	// and return the startTime (not including the following space)
-	return string(data[:iter]), nil
+	return &StatData{
+		StartTime: string(fields[statStartTimeLocation-3]),
+		State:     string(fields[statStateLocation-3]),
+	}, nil
 }
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -871,15 +871,14 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
 			log.Warnf(ctx, "Stopping container %v with stop signal timed out. Killing", c.ID())
 			if _, err := r.runtimeCmd("kill", c.ID(), "KILL"); err != nil {
 				log.Errorf(ctx, "Killing container %v failed: %v", c.ID(), err)
+				targetTime = time.Now().Add(10 * time.Millisecond)
 			}
 			if err := c.Living(); err != nil {
 				finished = true
-				break
 			}
 
 		case <-done:
 			finished = true
-			break
 		}
 	}
 


haircommander:
I think this would be better added to the code already i n https://github.com/cri-o/cri-o/blob/main/internal/oci/container.go#L530-L540

rphillips:
@haircommander If I move it to container_linux, then the freebsd portion is not implemented.

haircommander:
sorry I may not have been clear. Queries to /proc are not trivially cheap and I think we should consolidate reading from /proc/pid/status if possible. We're already reading the start time whenever we check a Living(), we could extend the [getPidStartTime](https://github.com/cri-o/cri-o/blob/main/internal/oci/container_linux.go#L47) function to be getPidStartAndState, and read it in the same function. we could then treat processes in D state as not running

internal/oci/container_freebsd.go
```diff
@@ -29,3 +30,7 @@ func getPidStartTime(pid int) (string, error) {
 	kp := (*C.struct_kinfo_proc)(unsafe.Pointer(&data[0]))
 	return fmt.Sprintf("%d,%d", kp.ki_start.tv_sec, kp.ki_start.tv_usec), nil
 }
+
+func GetPidState(pid int) (string, error) {
```
haircommander, line 34: these seem unused to me

internal/oci/runtime_oci.go
```diff
@@ -854,7 +854,8 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
 	// to catch a new timeout (and possibly ignore that new timeout if it's not correct to
 	// take a new one).
 	targetTime := time.Unix(1<<50-1, 0)
-	for finished := false; !finished; {
+killLoop:
+	for {
```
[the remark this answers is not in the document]
rphillips: The break statement is failing on the linter, and is only breaking out of the select statement.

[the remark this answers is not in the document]
rphillips: It's not really a goto... It's a label for the for loop.

internal/oci/container_freebsd.go
rphillips: fixed

internal/oci/runtime_oci.go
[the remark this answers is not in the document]
rphillips: fixed.

internal/oci/container_freebsd_nocgo.go
```diff
@@ -18,3 +18,15 @@ func getPidStartTime(pid int) (string, error) {
...
+
+// getPidData reads the kernel's /proc entry for various data.
+func getPidData(pid int) (*StatData, error) {
+	startTime, err := getPidStartTime(pid)
+	if err != nil {
+		return nil, err
+	}
+	return &StatData{
+		StartTime: startTime,
+		State:     "not implemented",
```
haircommander, line 30: could be a follow up, but it seems freebsd uses status as well: https://man.freebsd.org/cgi/man.cgi?query=proc&apropos=0&sektion=5&manpath=CentOS+7.1&arch=default&format=html the location is just slightly different. Could maybe just have platform set `statStateLocation`

internal/oci/container.go
```diff
@@ -527,17 +532,25 @@ func (c *Container) pid() (int, error) {
...
-	if startTime != c.state.InitStartTime {
+	if data.StartTime != c.state.InitStartTime {
 		return fmt.Errorf(
 			"PID %d is running but has start time of %s, whereas the saved start time is %s. PID wrap may have occurred",
-			c.state.InitPid, startTime, c.state.InitStartTime,
+			c.state.InitPid, data.StartTime, c.state.InitStartTime,
 		)
 	}
+
+	if data.State == "D" {
```
kwilczynski, line 547: We could skip states that are considered invalid per:

```go
// Skip states which are typically considered as invalid.
switch data.State {
case "D", "X", "Z":
	return fmt.Errorf("PID %d has an invalid state: %s", c.state.InitPid, data.State)
}
```

internal/oci/container_linux.go
```diff
@@ -41,40 +42,47 @@ func (c *Container) SeccompProfilePath() string {
...
-		iter = bytes.IndexByte(data, ' ')
-		if iter == -1 {
-			return "", fmt.Errorf("invalid number of entries found in stat file %s: %d: %w", file, field-1, ErrNotFound)
-		}
+	// skip the character after the space
+	iter++
+
+	fields := bytes.Split(data[iter:], []byte{' '})
+	if len(fields) <= statStartTimeLocation {
+		return nil, errors.New("invalid index finding proc field")
```
kwilczynski, line 81: I am unsure what the current error message means.

```suggestion
		return nil, errors.New("unable to parse stat file")
```

rphillips: fixed. thanks!

internal/oci/container.go
rphillips: fixed. thanks

internal/oci/container_freebsd_nocgo.go
rphillips: This man page is for centos... The freebsd [man page](https://man.freebsd.org/cgi/man.cgi?query=procfs&apropos=0&sektion=0&manpath=FreeBSD+13.2-RELEASE&arch=default&format=html) is lacking a pid state.

haircommander: lmao so silly it would even list that as an option

internal/oci/container_linux.go
```diff
@@ -41,40 +42,47 @@ func (c *Container) SeccompProfilePath() string {
...
 func GetPidStartTimeFromFile(file string) (string, error) {
+	data, err := getPidStatData(file)
+	if err != nil {
+		return "", err
+	}
+	return data.StartTime, nil
+}
+
+func getPidStatData(file string) (*StatData, error) {
 	data, err := os.ReadFile(file)
```
kwilczynski, line 63: Some ideas for small optimisations that we can do here if we want to.

1) We can use something that won't call `fstat()` on the file, which does not work for procfs/sysfs anyway
2) We can only request for a page of data (so typically a 4 KiB on x86_64) to be read at once

The (2) is because procfs/sysfs (contrary to debugfs, etc.) only serves a page worth of data.

This can reduce the number of allocations a little bit per:

```console
$ go test -v -bench=. -benchmem 
goos: linux
goarch: amd64
pkg: test-procfs
cpu: AMD Ryzen 9 6900HX with Radeon Graphics        
BenchmarkA
BenchmarkA-16    	  256002	      4549 ns/op	    2160 B/op	       8 allocs/op
BenchmarkB
BenchmarkB-16    	  284290	      4327 ns/op	    1976 B/op	       8 allocs/op
PASS
ok  	test-procfs	3.074s
```

The `os.ReadFile()` call to `fstat()` will only return `0` (or 4 KiB; as it would for most procfs/sysfs objects) per:

```console
fstat(3, {st_mode=S_IFREG|0444, st_size=0, ...}) = 0
``` 

There is also a potential to leverage `fmt.Fscan()` over a bytes buffer in lieu of `strings.Split()` to potentially further reduce memory allocation.

internal/oci/runtime_oci.go
```diff
@@ -871,15 +871,14 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
 			log.Warnf(ctx, "Stopping container %v with stop signal timed out. Killing", c.ID())
 			if _, err := r.runtimeCmd("kill", c.ID(), "KILL"); err != nil {
 				log.Errorf(ctx, "Killing container %v failed: %v", c.ID(), err)
+				targetTime = time.Now().Add(10 * time.Millisecond)
```
andreaskaris, line 874: Doesn't this still mean that in case of a container that can't be killed with -9 due to uninetteruptible sleep, we'd be logging every 10 ms?

internal/oci/container.go
```diff
@@ -527,17 +532,24 @@ func (c *Container) pid() (int, error) {
 // This is the simplest way to verify we are operating on the container
 // process, and haven't run into PID wrap.
 func (c *Container) verifyPid() error {
```
kwilczynski, line 534: There is a thing now with this function since it understands the process state. We might be baling out on things too early/eagerly, for example:

https://github.com/cri-o/cri-o/blob/5c111c94c50a633edf10e89a3c5969c5fe55edfe/internal/oci/runtime_oci.go#L805-L809

We wouldn't attempt to deliver SIGTERM or SIGKILL should the process be in the undesired state. However, this state might be brief, or the process might still be killable.

What I am saying, we need to make sure the old flow needs to be preserved/attempted at least once, I suppose.

haircommander: oh shoot you're right. yeah we need to have two different functions. One to check for Living pre-SIGKILL, one for LivingOrSleeping for after. If we filter "D" state out for all processes we check for living, the process can flip back to R and we would totally skip the kill
/hold

internal/oci/runtime_oci.go
kwilczynski: @andreaskaris, assuming that we've attempted to send the graceful termination signal (SIGTERM) already and then _at least once_ send the SIGKILL signal, then the aim with the new changes is to add this 10 ms delay/stagger and also bail out for any process which is in the state such as "D", "X" or "Z". Otherwise, yes, it would also be a lot of log entries (10 ms is not a lot relatively).

As for the "D" state, to us, it does not matter whether it's killable or uninterruptible—at this point, we are already on the path to termination. There should be at least one pending signal (since we should have attempted to send SIGKILL at least once) awaiting such a process to awake (when it will be delivered). There is also no easy way to see whether the "D" status has a TASK_KILLABLE or TASK_UNINTERRUPTIBLE flag set (userspace seems to have limited visibility via /proc in that specific case).

internal/oci/container.go
kwilczynski: We can have two functions or pass a simple boolean to `Living()` and then to `verifyPid()` to toggle the process status check so it can be skipped on demand, for example:

```go
func (c *Container) verifyPid(checkState bool) error {
	data, err := getPidData(c.state.InitPid)
	if err != nil {
		return err
	}

	if data.StartTime != c.state.InitStartTime {
		return fmt.Errorf(
			"PID %d is running but has start time of %s, whereas the saved start time is %s. PID wrap may have occurred",
			c.state.InitPid, data.StartTime, c.state.InitStartTime,
		)
	}

        // Don't check the process (PID) state.
	if !checkState {
                return nil
        }

        // Skip states which are typically considered as invalid.
        switch data.State {
	case "D", "X", "Z":
		return fmt.Errorf("PID %d has an invalid state: %s", c.state.InitPid, data.State)
	}

	return nil
}
```

kwilczynski: Alternatively, `verifyPid()` would be split into `verifyPIDWrap()` and `verifyPIDState()`, and used accordingly.

kwilczynski: I suppose, we _always_ want to check for a PID wrap condition. Which might render the idea of having two separate helpers moot.

haircommander:
I'm wondering if we should just add an exponential backoff and call it a day. We can't easily communicate to the kubelet the process is in "D" state without exiting StopLoopForContainer, which would require it is reentrant, or potentially cause other transient failures (runc kill decides to take a while?) from killing the container. The safest and simplest is to just try less frequently and wait for the kernel to catch up. What do folks think?

kwilczynski:
> [...]
> The safest and simplest is to just try less frequently and wait for the kernel to catch up. What do folks think?

Personally, I believe a simple solution might be the best served here. Otherwise, there are a lot of edge cases we might have to potentially address to make the stopping code robust.

---

To set the scene a little:

- Signals are reentrant, non-blocking and asynchronous
- Sending a signal to a process that is in the uninterruptible sleep without the TASK_KILLABLE set will be set as pending
- There is no way for us to correctly ascertain from the userspace the type of the "D" process state
  - Has the TASK_UNINTERRUPTIBLE state been set without the TASK_KILLABLE flag or not
- Process can be "blocked" directly or become irresponsible should one of its children is currently in a blocking state

---

While testing the proposed changes I was able to identify a few potential issues:

- Testing a single process within a container that was in the blocked state revealed a potential race or ordering issue

```console
panic: send on closed channel

goroutine 1810 [running]:
github.com/cri-o/cri-o/internal/oci.(*Container).WaitOnStopTimeout(0xc00089c000, {0x2e3beb0, 0xc0005d0150}, 0xa)
	github.com/cri-o/cri-o/internal/oci/container.go:604 +0xc5
github.com/cri-o/cri-o/internal/oci.(*runtimeOCI).StopContainer(0xc000153620, {0x2e3beb0?, 0xc0005d0120?}, 0xc00089c000, 0xa)
	github.com/cri-o/cri-o/internal/oci/runtime_oci.go:816 +0x352
github.com/cri-o/cri-o/internal/oci.(*Runtime).StopContainer(0xc0003cfb90, {0x2e3beb0?, 0xc0005d0090?}, 0xc00089c000, 0xa)
	github.com/cri-o/cri-o/internal/oci/oci.go:309 +0x102
github.com/cri-o/cri-o/server.(*Server).stopContainer(0xc000452808, {0x2e3beb0?, 0xc001069bf0?}, 0xc00089c000, 0xa)
	github.com/cri-o/cri-o/server/container_stop.go:60 +0x40b
github.com/cri-o/cri-o/server.(*Server).stopPodSandbox.func1()
	github.com/cri-o/cri-o/server/sandbox_stop_linux.go:63 +0x38
golang.org/x/sync/errgroup.(*Group).Go.func1()
	golang.org/x/sync@v0.6.0/errgroup/errgroup.go:78 +0x56
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1785
	golang.org/x/sync@v0.6.0/errgroup/errgroup.go:75 +0x96
```

Related logs:

<details>

```console
time="2024-02-20 02:36:26.182569470Z" level=warning msg="Stopping container 314423af59714f38f65ea8a23072248bccf3bacd43147fdc1db64527d515b791 with stop signal timed out. Killing" file="oci/runtime_oci.go:875"
time="2024-02-20 02:36:26.189833904Z" level=debug msg="Unmounted container 314423af59714f38f65ea8a23072248bccf3bacd43147fdc1db64527d515b791" file="storage/runtime.go:495" id=e026368c-a07e-478e-ac8e-bb9218b5fec3 name=/runtime.v1.RuntimeService/StopContainer
time="2024-02-20 02:36:26.195471499Z" level=info msg="Stopped container 314423af59714f38f65ea8a23072248bccf3bacd43147fdc1db64527d515b791: default/test/test" file="server/container_stop.go:29" id=e026368c-a07e-478e-ac8e-bb9218b5fec3 name=/runtime.v1.RuntimeService/StopContainer
time="2024-02-20 02:36:26.196017909Z" level=debug msg="Request: &StopPodSandboxRequest{PodSandboxId:07327609189e09d497a5647c3befaa3e6f3d64119591ba82d6821919039ec726,}" file="otel-collector/interceptors.go:62" id=08b90184-ad3d-4172-9f2e-9058d0d1b9ca name=/runtime.v1.RuntimeService/StopPodSandbox
time="2024-02-20 02:36:26.196048272Z" level=info msg="Stopping pod sandbox: 07327609189e09d497a5647c3befaa3e6f3d64119591ba82d6821919039ec726" file="server/sandbox_stop.go:18" id=08b90184-ad3d-4172-9f2e-9058d0d1b9ca name=/runtime.v1.RuntimeService/StopPodSandbox
```

</details>

The panic occurs following the `SetAsDoneStopping()` function that closes the `stopTimeoutChan` channel while there might still be a pending new timeout to send through from the `StopContainer()` function.

- The 10 milliseconds delay is not set (renewed) as a new timeout

The following block is never run as sending a signal to a blocked process, albeit a process that exists, would not solicit any errors from the kernel (syscall).

```go
	targetTime = time.Now().Add(10 * time.Millisecond)
}
```

Thus, when `runc` or `crun` are asked to kill the containers, their internal code does not fail, and as such, we see an error is set, which means that we won't be updating the `targetTime` variable with the new value.

- The `Living()` function had to be taught to skip process state checks

A need to pass a boolean through the `Living()` to the `pid()` to then `verifyPid()` denoting whether checking for the process state should be performed or not. This was to allow for an early, or initial, check of the process liveness to be performed that needed to ignore a potentially momentary "D" state of a given process (a process might have been sleeping momentarily and such).

- We can only check the process state of the main, or the "init" process, running within the container

If a single process runs within the container or so happens this process is the first process, then we can easily access and check its state—should it be "D" and we were going to terminate the process anyway, then the undesired state will be correctly detected and the loop will be terminated early, as intended.

However, if the process inside spawned a child process (one or more) and one of these children that aren't detached is currently blocked and might remain in such a state for a longer period of time, then the parent will also be unresponsive, and such, sending termination signals to it wouldn't have any desired effect—the signals will be simply pending.

This, in turn, means that we won't be able to detect the "D" state of the process, container, etc., while on the path to termination, so the loop will continue to spin, exhibiting the same behaviour as the original code currently does.

This is because we only pass the `state.InitPid` from the `Container` type each time we look at the process running within the container. Since the first or "init" process would birth the container into existence, it does make sense to check for this specific process only.

- Can container storage be unmounted while the process is still running?

When a single process running within the container is due to be terminated and it happens to be in the blocking "D" state, then we would terminate the busy loop early. Then, the following code would run:

https://github.com/cri-o/cri-o/blob/87ff09f9e045797ae43c1a7bde121b6cecb76954/internal/storage/runtime.go#L468-L497

The question here would be: is this safe to run for a process (container) that might still be running for a while?

haircommander:
> The question here would be: is this safe to run for a process (container) that might still be running for a while?

hard to tell without knowing exactly the stage the oci runtime fails to kill the container, but my guess is we'd get EBUSY on attempting to unmount the storage. I think retrying with an exponential backoff until the process is gone from the process table is the only safe and sustainable solution.

PR #7759, opened 2024-02-08
RFC: [1.28] server: treat image Digest as ID

haircommander:
It turns out, many users were relying on ImageRef as a reference to the resolved image. Even though this breaks the Kubelet garbage collection (see cri-o/cri-o#7149).

Try to get the best of both worlds by using the Digest of the image as the ID. It shouldn't make a difference to the kubelet as long as we do it consistently, and end-users can use the `$image + ':$assumedAlgorithm' + $imageID` to construct a name of the resolved image.

```release-note
Use the Digest of the image instead of the generated storage ID, which allows an end user to construct the resolved image by using a combination of the image name and image ID
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -236,7 +236,11 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 	}
 
 	imageName := imgResult.Name
-	imageRef := imgResult.ID
+	// Instead of using the image ID, which is generated on the pull, use the digest of the image.
+	// This allows us to construct a pull reference of the image by stripping the tag from imageName
+	// and appending the digest at the end.
+	// One limitation of this is a user needs to figure out the algorithm...
+	imageRef := imgResult.Digest.Encoded()
 
 	labelOptions, err := ctr.SelinuxLabel(sb.ProcessLabel())
 	if err != nil {
--- a/server/image_list.go
+++ b/server/image_list.go
@@ -53,7 +53,9 @@ func ConvertImage(from *storage.ImageResult) *types.Image {
 	}
 
 	to := &types.Image{
-		Id:          from.ID,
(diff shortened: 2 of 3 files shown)


haircommander:
note: this won't apply cleanly to main, and it breaks some of the abstractions done there recently.
cc @sascha @mtrmac

mtrmac (commented):
Even casting aside the ~aesthetics of incorrectly mixing different kinds of values, and the fact that I don’t think this works, I can’t see that this improves on the fundamental situation at all.

The core of the problem is:
> One could easily end up in a situation structurally like this:
> - Pod1 is set up with repo1:tag1, resolving to digest1
> - Pod2 is set up with repo2:tag2, resolving to digest2
> The two images happen to be “the same image”, and are deduplicated on pull
> RepoDigests[0] is set to repo1@digest2 (because the information to return the right value isn’t tracked), an image which doesn’t exist and couldn’t possibly be scanned.

and this doesn’t change the characteristics of the available information _at all_; it’s just that instead of `RepoDigests[0]`, an unpredictable repoX@digestX combination, the digest is now `status.Digest` == `storage.Image.Digest`, also one of the unpredictable digestX values.

---

If the goal is “only” to manipulate `ContainerStatus` in some way that the information is exposed, while not fixing the repo1@digest2 situation, \*shrug\*, I would point at https://github.com/mtrmac/cri-o/commit/1eae23ee64548a90642e605ba920946970a71e2d : apparently CRI gives us wide latitude to return whatever we want in that field, so returning an always-digested, even if nonsensical, repo1@digest2 value in that field would be an option.

That would, certainly, have a fairly high risk of breaking something (including those “security” products), and is probably not well-suited to a stable branch.

As an outsider, it seems to me that the only reasonable way to add a field to the status objects is to _add a field_.

haircommander:
yeah I thought it was a long shot.. closed

PR #7762, opened 2024-02-09
Revert "Fix ImageRef field for containers to default to an image ID"

saschagrunert:
#### What type of PR is this?

/kind bug

This reverts commit 41b13e28dd75828ab9cc24806ad4d2d3b4a605b9 and therefore https://github.com/cri-o/cri-o/pull/7149

None
None

```release-note
Reverted image ID field change in the container status. While the change did fix an issue with Kubelet image GC, it caused an unintended change in the container status of the Kubernetes pod API, breaking an accidentally stable API
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -125,11 +125,17 @@ type ContainerState struct {
 // at any future time.
 // imageName, if set, is _some_ name of the image imageID; it may have NO RELATIONSHIP to the users’ requested image name.
 // imageID is nil for infra containers.
-func NewContainer(id, name, bundlePath, logPath string, labels, crioAnnotations, annotations map[string]string, userRequestedImage string, imageName *references.RegistryImageReference, imageID *storage.StorageImageID, md *types.ContainerMetadata, sandbox string, terminal, stdin, stdinOnce bool, runtimeHandler, dir string, created time.Time, stopSignal string) (*Container, error) {
+// someRepoDigest, if set, is some repo@some digest of the image imageID; it
+// may have NO RELATIONSHIP to the users’ requested image name (and, which
+// should be fixed eventually, may be a repo@digest combination which has never
+// existed on a registry).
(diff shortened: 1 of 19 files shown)


saschagrunert:
/hold

for final discussion.

internal/oci/container.go
```diff
@@ -125,11 +125,13 @@ type ContainerState struct {
 // at any future time.
 // imageName, if set, is _some_ name of the image imageID; it may have NO RELATIONSHIP to the users’ requested image name.
 // imageID is nil for infra containers.
```
mtrmac, line 127: Please document _precisely_ what is the value / purpose of this parameter.
> someRepoDigest, if set, is _some_ repo@_some_ digest of the image imageID; it may have NO RELATIONSHIP to the users’ requested image name (and, which should be fixed eventually, may be a repo@digest combination which has never existed on a registry.

or

> containerStatusImageRef, if set, is a value intended to be returned in ContainerStatus [and other CRI APIs??]; it’s the caller’s responsibility to confirm with CRI semantics

or something.

We know these things now, so let’s some poor soul 5 years later the trouble of finding out.

mtrmac (commented):
#7149 was originally added because it is required by Kubelet’s image GC. What is the plan for fixing/preserving that fixed functionality?

haircommander:
/hold

it's my preference not to revert. I think there are potentially some other avenues we can explore first

saschagrunert:
> #7149 was originally added because it is required by Kubelet’s image GC. What is the plan for fixing/preserving that fixed functionality?

We're still negotiating on this.

internal/oci/container.go
saschagrunert: Alright, added the comment and changed the variable name

server/container_create_linux.go
```diff
@@ -245,6 +245,10 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 
 	imageName := imgResult.SomeNameOfThisImage
 	imageID := imgResult.ID
+	imageRef := ""
```
mtrmac, line 248: Consider renaming this to `someRepoDigest` as well

saschagrunert: Yep, done :+1:

PR #7764, opened 2024-02-09
CRI: An empty DNSConfig != unspecified

roman-kiselenko:
if DnsConfig empty, it should copy the host's resolv.conf

#### What type of PR is this?

/kind other

This PR coordinates the behavior with the containerd. https://github.com/containerd/containerd/pull/9730

None

Reference https://github.com/kubernetes/kubernetes/issues/120748

I guess some users will find this behavior a breaking change.

```release-note
In case of an empty struct of DSNConfig /etc/resolv.conf be copied.
```

### DIFF
--- a/internal/factory/sandbox/infra.go
+++ b/internal/factory/sandbox/infra.go
@@ -88,14 +88,16 @@ func PauseCommand(cfg *libconfig.Config, image *v1.Image) ([]string, error) {
 
 func (s *sandbox) createResolvConf(podContainer *storage.ContainerInfo) (retErr error) {
 	// set DNS options
+	s.resolvPath = fmt.Sprintf("%s/resolv.conf", podContainer.RunDir)
+
 	if s.config.DnsConfig == nil {
-		return nil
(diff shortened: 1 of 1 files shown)


roman-kiselenko:
> the alternative would be treating a nil object like an empty one and creating an empty `s.config.DnsServer` if it isn't defined.

So, just doing something like this:

```go
if s.config.DnsConfig == nil {
   s.config.DnsConfig = &types.DnsConfig{}
}
```

saschagrunert (approved):

roman-kiselenko:
Not sure why tests are failed, will take a look closer 👀 
/retest-required

sohankunkerkar:
>Something like this 🤔 :

>if s.config.DnsConfig == nil {
>  s.config.DnsConfig = &types.DnsConfig{}
>}

@roman-kiselenko are you planning to add this code?

roman-kiselenko:
> >Something like this 🤔 :
> 
> >if s.config.DnsConfig == nil {
> >  s.config.DnsConfig = &types.DnsConfig{}
> >}
> 
> @roman-kiselenko are you planning to add this code?

yep 👍

PR #7767, opened 2024-02-12
Add image digests to `crio status` API

saschagrunert:
#### What type of PR is this?

/kind feature

We now feature a `-j/--json` output for `crio status`. Beside this, the output has been extended to contain the image digests for the used container image. This will be achieved by utilizing a new reserved annotation `io.kubernetes.cri-o.ImageDigests`.

None
Follow-up on https://github.com/cri-o/cri-o/pull/7762

```release-note
Added image digests to `crio status` API as well as `-j/--json` flag.
```

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -8,8 +8,9 @@ import (
 type ContainerInfo struct {
 	Name            string            `json:"name"`
 	Pid             int               `json:"pid"`
-	Image           string            `json:"image"`     // If set, _some_ name of the image imageID; it may have NO RELATIONSHIP to the users’ requested image name.
-	ImageRef        string            `json:"image_ref"` // In the format of StorageImageID.StringForOutOfProcessConsumptionOnly(), or "".
+	Image           string            `json:"image"`         // If set, _some_ name of the image imageID; it may have NO RELATIONSHIP to the users’ requested image name.
+	ImageRef        string            `json:"image_ref"`     // In the format of StorageImageID.StringForOutOfProcessConsumptionOnly(), or "".
+	ImageDigests    []string          `json:"image_digests"` // Possible repo digests of the used container image.
 	CreatedTime     int64             `json:"created_time"`
 	Labels          map[string]string `json:"labels"`
 	Annotations     map[string]string `json:"annotations"`
--- a/vendor/github.com/containers/podman/v4/pkg/annotations/annotations.go
+++ b/vendor/github.com/containers/podman/v4/pkg/annotations/annotations.go
@@ -40,6 +40,9 @@ const (
 	// ImageRef is the container image ref annotation.
 	ImageRef = "io.kubernetes.cri-o.ImageRef"
 
+	// ImageDigests is the container image digests annotation.
+	ImageDigests = "io.kubernetes.cri-o.ImageDigests"
+
 	// KubeName is the kubernetes name annotation.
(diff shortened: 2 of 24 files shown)


vendor/github.com/containers/podman/v4/pkg/annotations/annotations.go
```diff
@@ -40,6 +40,9 @@ const (
 	// ImageRef is the container image ref annotation.
 	ImageRef = "io.kubernetes.cri-o.ImageRef"
 
+	// ImageDigests is the container image digests annotation.
+	ImageDigests = "io.kubernetes.cri-o.ImageDigests"
```
saschagrunert, line 44: @cri-o/cri-o-maintainers are we happy with that annotation? I have to add it to Podman and vendor it back in here afterwards.

haircommander: why does it have to be added there?

saschagrunert: We have the other ones there as well. Would you prefer a different location?

haircommander: not necessarily, mostly curious :)

haircommander: name LGTM

pkg/types/types.go
```diff
@@ -8,8 +8,9 @@ import (
 type ContainerInfo struct {
 	Name            string            `json:"name"`
 	Pid             int               `json:"pid"`
-	Image           string            `json:"image"`     // If set, _some_ name of the image imageID; it may have NO RELATIONSHIP to the users’ requested image name.
-	ImageRef        string            `json:"image_ref"` // In the format of StorageImageID.StringForOutOfProcessConsumptionOnly(), or "".
+	Image           string            `json:"image"`         // If set, _some_ name of the image imageID; it may have NO RELATIONSHIP to the users’ requested image name.
+	ImageRef        string            `json:"image_ref"`     // In the format of StorageImageID.StringForOutOfProcessConsumptionOnly(), or "".
+	ImageDigests    []string          `json:"image_digests"` // Possible repo digests of the used container image.
```
mtrmac, line 13: Is this exactly what the users need?

- If the container is created after pulling image, and we conceptually know the _right_ repo@digest combination (at the time of pull, and we return a repo@digest value from pull), shouldn’t this ideally contain only that one? The documentation of this field seems to commit us to returning all of RepoDigests (“of the used container image”, not “possibly used when creating this container”).
- If the container is created after ImageStatus (and we don’t know the _right_ repo@digest right now, what are the users going to do? Combine the container’s image repo with this new set of RepoDigests to find out the right digest? That could work (but it is imprecise in case the precise digest matters, e.g. WRT presence of signatures) _except_ that this struct does not give them the repo of the image. As documented above, `Image` is `ImageResult.SomeNameOfThisImage`, and has no relationship with the user-requested repo.
- What is the long-term plan? Is this API eventually going to return the one right value? Or is this going to be completely deprecated in favor of container statuses in Pod.Status? If the latter, that should probably be already documented here right now.

(Starting at the exposed API; implementation details would follow from these decisions.)

saschagrunert:
The plan is right now to first restore the original behavior (in https://github.com/cri-o/cri-o/pull/7762) and then think about enhancements like this.

PR #7771, opened 2024-02-12
WIP: Reload CRI-O configurations in particular order

### LINKED ISSUE #7586: Reloading CRI-O configuration using SIGHUP overrides settings passed through command-line arguments
### What happened?

When the SIGHUP signal was sent to a currently running CRI-O process, it reloaded its configuration.

However, the CRI-O process was started with several command-line arguments, which were used to override default configuration options such as the log level or the pause image reference.

Following the request to CRI-O to reload its runtime configuration, the options specific via the command-line arguments were overridden with either a default or an item from the /etc/crio/crio.conf (or a drop-in from /etc/crio/crio.conf.d) file.

From this point onwards, the log level was changed, and the reference for the pause container image was updated to a hardcoded default.

The configuration to be reloaded with the options set through the applied command-line arguments to be preserved.

The precedence of configuration parsing should remain unchanged, and everything should also be merged accordingly, whether it was a fresh start of a runtime configuration reload request via the associated Linux signal delivered to the running process.

Additionally, the message that notifies the user about configuration reload should always be printed - this is not the case currently, as the log line itself is subjected to the presently set log level.

Preferably, it would also be nice if CRI-O had printed its runtime configuration on startup or even following a configuration reload, mainly to capture this type of detail in the log file for later reference (which can be helpful during troubleshooting). However, if the amount of logs this would produce is substantial, then perhaps the entire configuration will only be printed at the debug log level, and otherwise, only (so at the "info" level) difference to the default options would be printed - not sure if this would be desirable, though.

Only a few steps are needed to reproduce this behaviour reliably:

1. Prepare a configuration file drop-in to override the default log level (which is "info") to the level of warning (or "warn")

```shell
# cat > /etc/crio/crio.conf.d/99-log-level.conf
[crio.runtime]
log_level = "warn"
^D
```

2. Start a new CRI-O process manually overriding the log level and setting it to the "debug" level

```bash
# crio --log-level debug
````

```console
INFO[2023-12-12 22:13:11.794601654Z] Starting CRI-O, version: 1.29.0, git: 7380d8ff553b9b2ead8bf3173e8c1db0bc9004db(clean) 
```

_(... log lines that follow omitted for brevity ...)_

At this point, you can observe CRI-O process log level being set to "debug", and a rather substantial number of log lines being produced. Aside from the logs themselves, you can use the `crio` binary itself (assuming a modern version of CRI-O; otherwise, the `crio-status` can be used) to verify the current runtime configuration log level and what it's being set to, per:

```shell
# crio status config print | grep log_level
INFO[2023-12-12 22:08:57.099952659Z] Starting CRI-O, version: 1.28.1, git: unknown(clean) 
    log_level = "debug"
```

3. Send the **SIGHUP** signal to the running CRI-O process in a different terminal session

```bash
$ kill -HUP $(pgrep crio) # Or using pkill -HUP crio, whichever is more convenient.
```
Please return your attention to the terminal session running the CRI-O process in the foreground. At this point, you can observe the following lines appearing amongst other messages:

_(... a great many other log lines omitted for brevity ...)_

```console
DEBU[2023-12-12 22:13:13.762844491Z] received signal                               file="crio/main.go:57" signal=hangup
INFO[2023-12-12 22:13:13.763082980Z] Reloading configuration                       file="config/reload.go:18"
DEBU[2023-12-12 22:13:13.763491295Z] Using /sbin/apparmor_parser binary            file="supported/supported.go:61"
DEBU[2023-12-12 22:13:13.763702316Z] Using /sbin/apparmor_parser binary            file="supported/supported.go:61"
INFO[2023-12-12 22:13:13.764018888Z] Updating config from file /etc/crio/crio.conf  file="config/reload.go:27"
INFO[2023-12-12 22:13:13.764863975Z] Updating config from path /etc/crio/crio.conf.d  file="config/reload.go:36"
INFO[2023-12-12 22:13:13.765185045Z] Set config log_level to "warn"                file="config/reload.go:83"
```

Note the confirmation of signal reception and configuration reload. There is also a log line corresponding to a change in the log level, which is crucial to note.

Similarly to before, you can use the `crio` binary itself to obtain the current runtime configuration, per:

```bash
# crio status config print | grep log_level
INFO[2023-12-12 22:15:27.997584925Z] Starting CRI-O, version: 1.28.1, git: unknown(clean) 
    log_level = "warn"
```

This observed behaviour following a runtime configuration reload is not necessarily desirable.

Personally, I believe we should aim to preserve the order of precedence when reloading the runtime configuration. One could argue that "_but, what when the user makes a change to the configuration file, and would like or override option set prior via a command-line arguments?_" - well, it's a fair point. Nonetheless, what was set via the configuration arguments should remain unchanged and be the default behaviour. Otherwise, we risk violating the principle of least surprise for the user, which we should avoid if it can be helped.

The suggested order of preference when merging configuration would be (_higher_ to _lower_):

- Any command-line arguments that were passed
- Environment variables set
- Configuration files (including any files from the so-called "conf.d" directory where the drop-ins are to be stored)
- Default configuration

Now, the "conf.d" directory also has its own rules of how files from there are to be read and in what order - a lot of users respect files to be sorted and read in some order, most likely lexicographic (sort with the "C" locale to remove ambiguity), such that file called "01-foo.conf" would be loaded before files such as "zzz_bar.conf" or "99-baz.conf", I suppose.

Much of this is already working as intended, of course.

---

The following code is relevant:

- Main and runtime configuration types

https://github.com/cri-o/cri-o/blob/7380d8ff553b9b2ead8bf3173e8c1db0bc9004db/pkg/config/config.go#L65-L80

https://github.com/cri-o/cri-o/blob/7380d8ff553b9b2ead8bf3173e8c1db0bc9004db/pkg/config/config.go#L236-L477

- Configuration reload handler

https://github.com/cri-o/cri-o/blob/7380d8ff553b9b2ead8bf3173e8c1db0bc9004db/server/server.go#L583-L600

- The Config type concrete Reload() function

https://github.com/cri-o/cri-o/blob/7380d8ff553b9b2ead8bf3173e8c1db0bc9004db/pkg/config/reload.go#L17-L77

- The function that handles log level reload and set-up

https://github.com/cri-o/cri-o/blob/7380d8ff553b9b2ead8bf3173e8c1db0bc9004db/pkg/config/reload.go#L88-L102

<details>

```console
$ crio --version
crio version 1.29.0
Version:        1.29.0
GitCommit:      7380d8ff553b9b2ead8bf3173e8c1db0bc9004db
GitCommitDate:  2023-12-12T19:22:16Z
GitTreeState:   clean
BuildDate:      2023-12-12T21:54:46Z
GoVersion:      go1.21.5
Compiler:       gc
Platform:       linux/amd64
Linkmode:       static
BuildTags:      
  containers_image_ostree_stub
  libdm_no_deferred_remove
  containers_image_openpgp
  seccomp
  selinux
  apparmor
  static
  netgo
  exclude_graphdriver_btrfs
LDFlags:          unknown
SeccompEnabled:   true
AppArmorEnabled:  true
```

</details>

### OS version

<details>

```console
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

```console
$ uname -a
Linux worker-node01 5.15.0-83-generic #92-Ubuntu SMP Mon Aug 14 09:30:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

</details>

<details>
None.
</details>

SD-13:

Fixes #7586 

```release-note

```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -370,7 +370,7 @@ type RuntimeConfig struct {
 	MinimumMappableUID int64 `toml:"minimum_mappable_uid"`
 
 	// GIDMappings specifies the GID mappings to have in the user namespace.
-	// A range is specified in the form containerUID:HostUID:Size.  Multiple
+	// A range is specified in the form containerUID:HostUID:Size. Multiple
 	// ranges are separated by comma.
 	GIDMappings string `toml:"gid_mappings"`
 
@@ -507,7 +507,7 @@ type ImageConfig struct {
 	PinnedImages []string `toml:"pinned_images"`
 	// SignaturePolicyPath is the name of the file which decides what sort
 	// of policy we use when deciding whether or not to trust an image that
-	// we've pulled.  Outside of testing situations, it is strongly advised
+	// we've pulled. Outside of testing situations, it is strongly advised
 	// that this be left unspecified so that the default system-wide policy
 	// will be used.
 	SignaturePolicyPath string `toml:"signature_policy"`
@@ -691,7 +691,7 @@ func (c *Config) UpdateFromFile(path string) error {
 }
 
 // UpdateFromDropInFile populates the Config from the TOML-encoded file at the
-// given path.  The file may be the main configuration file, or it can be one
+// given path. The file may be the main configuration file, or it can be one
 // of the drop-in files which are used to supplement it.
 // Returns errors encountered when reading or parsing the files, or nil
 // otherwise.
--- a/pkg/config/reload.go
+++ b/pkg/config/reload.go
@@ -41,6 +41,11 @@ func (c *Config) Reload() error {
 		logrus.Infof("Skipping not-existing config path %q", c.dropInConfigDir)
 	}
 
+	// finding better way to know if any flags were set from cmd line args, even for ENV vars...
+	if len(os.Args) > 2 {
+		newConfig = c
+	}
+
 	// Reload all available options
 	if err := c.ReloadLogLevel(newConfig); err != nil {
 		return err


SD-13:
@kwilczynski I know it needs to be improved a lot and the code is little incomplete, but can you please take a look and let me know whether this is a good way of solving the bug? I feel like there is lots of boilerplate code. Can you please help in pointing out how it needs to be improved? Thanks!

SD-13:
> However, we shouldn't be parsing anything manually. We have a library that does this for us that we use.

Here did you mean the `cli` (github.com/urfave/cli/v2) library?

SD-13:
Hey @kwilczynski I was having some health issues which caused the delay :)

I took a closer look and now not able to figure out how to update the config with the flagset. Let me try to break in down

> However, we shouldn't be parsing anything manually. We have a library that does this for us that we use.

Here did you mean `cli`?

I tried to follow along and see [here](https://github.com/cri-o/cri-o/blob/6ff7c533bab71a0b2fd23ee3debaddc13729172e/vendor/github.com/urfave/cli/v2/command.go#L161) a flagset object is being created where command line argument values is being set as `Value: ...`.
Here are we updating the config with the flagset values? Seems like [this line](https://github.com/cri-o/cri-o/blob/6ff7c533bab71a0b2fd23ee3debaddc13729172e/vendor/github.com/urfave/cli/v2/command.go#L227) tells something else which don't understand clearly. Could you please explain how the flags are being set here to run the app?
btw, I started navigating from [main.go](https://github.com/cri-o/cri-o/blob/6ff7c533bab71a0b2fd23ee3debaddc13729172e/cmd/crio/main.go#L457).

> Have a look at how configuration is parsed and then how the runtime configuration object is assembled. Then, look at how this path might differ when the reload is requested. Why is it different?

What did you mean by this path?

I am trying to do this and figuring out the first 2 cases -

if cmd args were set:
    parse the flags and get the updated config
else if env vars were set:
    parse the env vars and get the updated config
else if a file path was provided:
    Use that as config
else
    Use default config

SD-13:
@kwilczynski @saschagrunert Can you please help by pointing out where the default config is being overridden with the command line arguments. I am new to Go and getting lost while navigating.

saschagrunert:
> @kwilczynski @saschagrunert Can you please help by pointing out where the default config is being overridden with the command line arguments.

Yes, this will happen here: 
https://github.com/cri-o/cri-o/blob/d705c02cec33b70e6a8b7a6eee44d23e744b9618/internal/criocli/criocli.go#L40

PR #7778, opened 2024-02-13
[WIP] *: refactor createSandboxContainer to reduce cyclomatic complexity

sohankunkerkar:

```release-note
None
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -129,6 +129,12 @@ type Container interface {
 	// WillRunSystemd checks whether the process args
 	// are configured to be run as a systemd instance.
 	WillRunSystemd() bool
+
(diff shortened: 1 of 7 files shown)


internal/factory/container/container.go
```diff
@@ -759,3 +762,13 @@ func getOCICapabilitiesList() []string {
 	}
 	return caps
 }
+
+// IsInCRIMounts checks if a mount path is in the list of mounts
+func (c *container) IsInCRIMounts(dst string) bool {
```
haircommander, line 767: TBH I dislike this function. I think it's a side effect of poor organization. I think we should refactor all of the mount code so we don't have to constantly loop through the mounts like a bunch of times. Focusing on that would reduce the complexity a lot. WDYT?

sohankunkerkar: Yeah, that's a good idea.

PR #7788, opened 2024-02-16
server: add support for reloading mirror registries automatically

### LINKED ISSUE #7748: Mirror configuration automatic reload
Currently cri-o supports reloading mirror configuration by either restarting the process or sending a SIGHUP signal to the process. In comparison Containerd reload the mirror configuration every time an image is pulled. Would there be interest in exploring the possibility of automatically reloading mirror configuration?

The reason I am asking is because I am looking at adding support for cri-o in my project https://github.com/XenitAB/spegel/. One of the blockers right now is that the end user experience would be a lot more complicated for cri-o users if they have to figure out some method of writing mirror configuration on their own.

sohankunkerkar:
Fixes #7748 

```release-note
Add support for automatically reloading mirror registries when there's an update to the registries.conf.d directory 
```


kwilczynski:
/retitle [WIP] server: add support for reloading mirror registries automatically

server/server.go
```diff
@@ -943,3 +946,57 @@ func isNotFound(err error) bool {
...
+// "/etc/containers/registries.conf.d". The routine reloads the registries
+// configuration when a file is modified or created in the directory.
+func (s *Server) startReloadWatcherForMirrorRegistries(ctx context.Context, registriesConfDDir string) {
+	if registriesConfDDir == "" {
+		log.Infof(ctx, "No registries.conf.d directory specified, using default")
+		registriesConfDDir = "/etc/containers/registries.conf.d"
+	}
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Fatalf(ctx, "Failed to create new watch: %v", err)
```
kwilczynski, line 961:
```suggestion
		log.Fatalf(ctx, "Failed to create new watcher: %s", err)
```

server/server.go
```diff
@@ -943,3 +946,57 @@ func isNotFound(err error) bool {
...
 }
+
+// startReloadWatcherForMirrorRegistries starts a new goroutine that
+// monitors the "registries.conf.d" directory. If not set, it defaults to
+// "/etc/containers/registries.conf.d". The routine reloads the registries
+// configuration when a file is modified or created in the directory.
+func (s *Server) startReloadWatcherForMirrorRegistries(ctx context.Context, registriesConfDDir string) {
+	if registriesConfDDir == "" {
+		log.Infof(ctx, "No registries.conf.d directory specified, using default")
+		registriesConfDDir = "/etc/containers/registries.conf.d"
```
kwilczynski, line 957: A candidate to make into a constant.

server/server.go
```diff
@@ -943,3 +946,57 @@ func isNotFound(err error) bool {
...
+				return
+			}
+			// Skip events for files starting with a dot
+			if strings.HasPrefix(filepath.Base(event.Name), ".") {
+				continue
+			}
+			if event.Op&(fsnotify.Write|fsnotify.Create) != 0 {
+				log.Infof(ctx, "File %s modified or created, reloading registries configuration", event.Name)
+				if err := s.config.ReloadRegistries(); err != nil {
+					log.Errorf(ctx, "Failed to reload registry configuration: %v", err)
```
kwilczynski, line 992: You do like to use `%v` for errors, where we should rely on `%s` to ensure that the type `String()` method is called.

```suggestion
					log.Errorf(ctx, "Failed to reload registry configuration: %s", err)
```

server/server.go
```diff
@@ -943,3 +946,57 @@ func isNotFound(err error) bool {
...
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+			// Skip events for files starting with a dot
+			if strings.HasPrefix(filepath.Base(event.Name), ".") {
+				continue
+			}
+			if event.Op&(fsnotify.Write|fsnotify.Create) != 0 {
```
kwilczynski, line 989: Deletions are tricky. Do we handle them correctly?

server/server.go
```diff
@@ -943,3 +946,57 @@ func isNotFound(err error) bool {
...
+				log.Infof(ctx, "File %s modified or created, reloading registries configuration", event.Name)
+				if err := s.config.ReloadRegistries(); err != nil {
+					log.Errorf(ctx, "Failed to reload registry configuration: %v", err)
+				}
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return
+			}
+			log.Errorf(ctx, "Watcher error: %v", err)
```
kwilczynski, line 999: Same here.

```suggestion
			log.Errorf(ctx, "Watcher error: %s", err)
```

server/server.go
kwilczynski: What about the default `/etc/containers/registries.conf` file?

Tere is also the `XDG_CONFIG_HOME` (albeit, I am not sure if this is strictly followed by the underlying package) where the user can put their configuration into `${HOME}/.config/containers/registries.conf`.

kwilczynski:
Generally, robust file events watching is non-trivial. Especially when you aim to watch a directory. Aside from cross-platform implementation issues (we have *BSD support in the works; albeit this won't be an issue for a while), which the Go package here aims to abstract, there are cross-platform behaviour issues. And even so, on the same platform (let it be Linux), there have been issues between different kernel versions and different sysctl settings (for fsnotify).

That said, there are also issues with the notifications themselves. File system events are not atomic and can also be sent multiple times for the same path being watched, depending on the platform and the software that interacts with the file somewhere. More often, the software behaviour is something one has to compensate for, but in some cases, the OS will also not behave as expected.

For example, consider the following examples (collected using the very same Go package as the one CRI-O uses):

```console
# date > test

04:11:05.6593   1 CREATE        "/etc/containers/registries.conf.d/test"
04:11:05.6599   2 WRITE         "/etc/containers/registries.conf.d/test"

# mv test /tmp

04:11:22.9017   3 RENAME        "/etc/containers/registries.conf.d/test"

# mv /tmp/test .

04:11:30.2516   4 CREATE        "/etc/containers/registries.conf.d/test"

# rm test

04:11:38.1324   5 REMOVE        "/etc/containers/registries.conf.d/test"

# vim test

04:13:33.7303   6 CREATE        "/etc/containers/registries.conf.d/.test.swp"
04:13:33.7304   7 CREATE        "/etc/containers/registries.conf.d/.test.swpx"
04:13:33.7306   8 REMOVE        "/etc/containers/registries.conf.d/.test.swpx"
04:13:33.7306   9 REMOVE        "/etc/containers/registries.conf.d/.test.swp"
04:13:33.7307  10 CREATE        "/etc/containers/registries.conf.d/.test.swp"
04:13:33.7307  11 WRITE         "/etc/containers/registries.conf.d/.test.swp"
04:13:37.7906  12 WRITE         "/etc/containers/registries.conf.d/.test.swp"
04:13:46.8045  13 WRITE         "/etc/containers/registries.conf.d/.test.swp"
04:13:49.8956  14 CREATE        "/etc/containers/registries.conf.d/test"
04:13:49.8958  15 WRITE         "/etc/containers/registries.conf.d/test"
04:13:49.8959  16 WRITE         "/etc/containers/registries.conf.d/test"
04:13:49.8971  17 WRITE         "/etc/containers/registries.conf.d/.test.swp"
04:13:53.9019  18 WRITE         "/etc/containers/registries.conf.d/.test.swp"
04:13:55.0267  19 CREATE        "/etc/containers/registries.conf.d/4913"
04:13:55.0268  20 CHMOD         "/etc/containers/registries.conf.d/4913"
04:13:55.0269  21 REMOVE        "/etc/containers/registries.conf.d/4913"
04:13:55.0271  22 RENAME        "/etc/containers/registries.conf.d/test"
04:13:55.0271  23 CREATE        "/etc/containers/registries.conf.d/test~"
04:13:55.0271  24 CREATE        "/etc/containers/registries.conf.d/test"
04:13:55.0272  25 WRITE         "/etc/containers/registries.conf.d/test"
04:13:55.0273  26 WRITE         "/etc/containers/registries.conf.d/test"
04:13:55.0280  27 CHMOD         "/etc/containers/registries.conf.d/test"
04:13:55.0281  28 REMOVE        "/etc/containers/registries.conf.d/test~"
04:13:55.0287  29 REMOVE        "/etc/containers/registries.conf.d/.test.swp"

# nano test

04:15:38.8292  38 CREATE        "/etc/containers/registries.conf.d/.test.swp"
04:15:38.8292  39 WRITE         "/etc/containers/registries.conf.d/.test.swp"
04:15:40.0551  40 REMOVE        "/etc/containers/registries.conf.d/.test.swp"
04:15:40.0551  41 CREATE        "/etc/containers/registries.conf.d/.test.swp"
04:15:40.0552  42 WRITE         "/etc/containers/registries.conf.d/.test.swp"
04:15:44.8382  43 CREATE        "/etc/containers/registries.conf.d/test"
04:15:44.8383  44 WRITE         "/etc/containers/registries.conf.d/test"
04:15:44.8393  45 REMOVE        "/etc/containers/registries.conf.d/.test.swp"

# touch test

04:16:35.3824  49 CHMOD         "/etc/containers/registries.conf.d/test"

# >test
# ln -f test test2
# date > test2
# ls -l

-rw-r--r-- 2 root root 32 Feb 21 04:18 test
-rw-r--r-- 2 root root 32 Feb 21 04:18 test2

04:18:06.0837  58 CREATE        "/etc/containers/registries.conf.d/test"
04:18:10.9388  59 CREATE        "/etc/containers/registries.conf.d/test2"
04:18:13.8936  60 WRITE         "/etc/containers/registries.conf.d/test2"
04:18:13.8940  61 WRITE         "/etc/containers/registries.conf.d/test2"

# ln -sf test2 test3
# date > test3
# ls -l

-rw-r--r-- 2 root root 32 Feb 21 04:18 test
-rw-r--r-- 2 root root 32 Feb 21 04:18 test2
lrwxrwxrwx 1 root root  5 Feb 21 04:19 test3 -> test2

04:19:06.4723  62 CREATE        "/etc/containers/registries.conf.d/test3"
04:19:16.3944  63 WRITE         "/etc/containers/registries.conf.d/test2"
04:19:16.3947  64 WRITE         "/etc/containers/registries.conf.d/test2"
```

Some of the above can be translated into popular use cases:

- Editing a file with a text editor
- Creating a file via output redirection (when the file does not exist before)
- Wanting to trigger a file/configuration reload using `touch` to update file metadata
- Moving a file to a different directory (not a delete, and yet the file will be removed from the current path)

You can also see that some operations on file on the user's end generate events that are unexpected.

However, the biggest problem is almost always having to deal with multiple events arriving on the same path. Are these duplicate events? Some might be, but some might be distinct events (for example, a large write split into smaller chunks).

Why is this important? The code, as it stands now, would keep reloading the registry configuration many times over in response to operation on even a single file. This might be undesirable—additional file reads will be done, additional objects will be created that need to be garbage-collected eventually, etc. There is also no lock or delay of any sort to prevent containers from being created and started while registries are being reloaded, and there might be a race here, albeit this is potentially the worst possible outcome and highly unlikely.

To deal with this sort of problem, one would have to implement event deduplication or events aggregator. or rate limiter so that multiple events for the same path are seen once or are staggered or throttled.

server/server.go
kwilczynski: There might be issues watching a single file, though.

server/server.go
sohankunkerkar: I'm trying to be consistent with the codebase. I've reused this line from `StartMontiorExit`, so I need to rectify that as well. Again, I want to be consistent with the codebase regarding `%v` over `%s`. I can switch to `%#v/%+v` or `%w` for the complex struct.

server/server.go
sohankunkerkar: I'm unsure about scenarios where users directly modify /etc/containers/registries.conf. I thought we should monitor user-configured drop-in directories or `/etc/containers/registries.conf.d`. Currently, Cri-O reloads the configuration with a `SIGHUP` signal. Given the mirror registries reloading issue, it seems unlikely users would directly alter `registries.conf`, but this is just my personal opinion. To comprehensively address use cases, we should consider edge cases as well.

server/server.go
sohankunkerkar: hmm... that's a good point. Currently, the code doesn't handle that scenario.

sohankunkerkar:
>Why is this important? The code, as it stands now, would keep reloading the registry configuration many times over in response to operation on even a single file. This might be undesirable—additional file reads will be done, additional objects will be created that need to be garbage-collected eventually, etc. There is also no lock or delay of any sort to prevent containers from being created and started while registries are being reloaded, and there might be a race here, albeit this is potentially the worst possible outcome and highly unlikely.

>To deal with this sort of problem, one would have to implement event deduplication or events aggregator. or rate limiter so that multiple events for the same path are seen once or are staggered or throttled.

Thanks for the feedback. I completely agree with your observations. The current implementation lacks event deduplication, and the absence of locks or delays could potentially lead to undesirable outcomes, though the likelihood is minimal. While we are here, we need to identify and document all potential use cases that the current implementation should handle (apart from the ones you listed), just to be on the same page and facilitate a quicker code-compile-debug cycle.

server/server.go
haircommander: I think there are functions in c/storage somewhere to get this file correctly. I think we probably should use them in case a user uses crio as a non root user (though this is not really supported). I also think we can ignore registires.conf for this feature, and focus on conf.d. That seems like a reasonable limitation and simplifies the code

server/server.go
```diff
@@ -943,3 +946,57 @@ func isNotFound(err error) bool {
...
+	log.Infof(ctx, "Registered reload watcher for mirror registries configuration")
+	done := make(chan struct{})
+	go s.reloadMirrorRegistriesConfiguration(ctx, watcher, done, registriesConfDDir)
+	<-done
+}
+
+func (s *Server) reloadMirrorRegistriesConfiguration(ctx context.Context, watcher *fsnotify.Watcher, done chan struct{}, path string) {
+	defer close(done)
+	if err := watcher.Add(path); err != nil {
+		log.Errorf(ctx, "Watcher.Add(%q) failed: %s", path, err)
```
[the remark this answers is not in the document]
haircommander: is there a way we can generalize StartExitMonitor to not repeat the code?

server/server.go
```diff
@@ -531,6 +531,9 @@ func New(
 	log.Debugf(ctx, "Sandboxes: %v", s.ContainerServer.ListSandboxes())
 
 	s.startReloadWatcher(ctx)
+	go func() {
+		s.startReloadWatcherForMirrorRegistries(ctx, s.config.SystemContext.SystemRegistriesConfDirPath)
```
haircommander, line 535: could this behavior be configurable?

server/server.go
haircommander: I can imagine users will directly modify

server/server.go
kwilczynski: We spoke about this with @sohankunkerkar.

We would either want this enabled by default with an option to disable this feature or disabled with an option to enable it. The latter is probably safer. Why? We have some customers using CRI-O who use or might use a lot of registry entries, and having this reload everything would be very costly.

server/server.go
kwilczynski: We probably should rely more on `%w` for correctness—with the cautious exception of an `os.PathError` where %v might be a better choice. Then, we stick with `%v` over `%s`. The difference is subtle and boils down to how the `nil`-values are represented, plus `go vet` picks on things when `%s` is used, which can be helpful.

I removed other review comments and corrected this one per the preference.

server/server.go
sohankunkerkar: Yup, I like the latter. Let's disable this feature by default, and users can enable it anytime if they wish. I will need to add that option in the CRI-O config

server/server.go
sohankunkerkar: Yup, makes sense.

server/server.go
sohankunkerkar: I had an offline discussion with @mtrmac regarding the deprecation note mentioned above. It seems that the function will reliably trigger a reload without a problem. The deprecation note is specifically related to the return value, now incorporating more configuration items than the utilized V2RegistriesConf type. Also, we don't use the configuration returned by that function anywhere in the codebase.

server/server.go
```diff
@@ -943,3 +948,92 @@ func isNotFound(err error) bool {
...
+	if err := watcher.Add(path); err != nil {
+		log.Errorf(ctx, "Failed to add watcher for path %q: %s", path, err)
+		return
+	}
+	var (
+		mu                   sync.Mutex
+		lastReload           time.Time
+		pendingReload        = false
+		reloadCount          int
+		reloadLimitPerMinute = 10
```
sohankunkerkar, line 987: We could rate-limit the number of times the registry can be reloaded. I'm not sure about the right number here, but I'm open to suggestions. Additionally, I'm open to discussing the possible scenarios in case we decide to proceed with this approach.

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

completions/bash/crio
```diff
@@ -19,6 +19,7 @@ h
 --address
 --allowed-devices
 --apparmor-profile
+--automatic-reload-mirror-registry
```
haircommander, line 22: myabe `automatically-reload-registries`?

pkg/config/config.go
```diff
@@ -859,37 +864,38 @@ func DefaultConfig() (*Config, error) {
...
+			apparmorConfig:                apparmor.New(),
+			blockioConfig:                 blockio.New(),
+			cgroupManager:                 cgroupManager,
+			deviceConfig:                  device.New(),
+			namespaceManager:              nsmgr.New(defaultNamespacesDir, ""),
+			rdtConfig:                     rdt.New(),
+			ulimitsConfig:                 ulimits.New(),
+			HostNetworkDisableSELinux:     true,
+			DisableHostPortMapping:        false,
+			AutomaticReloadMirrorRegistry: false,
```
haircommander, line 898: I think this would be better in the image config

server/server.go
```diff
@@ -49,6 +49,7 @@ const (
 	certRefreshInterval            = time.Minute * 5
 	rootlessEnvName                = "_CRIO_ROOTLESS"
 	irqBalanceConfigRestoreDisable = "disable"
+	eventStaggerDuration           = 50 * time.Millisecond
```
haircommander, line 52: have you tested this with Spegel yet? is this stagger enough?

server/server.go
haircommander: the original point still stands: is this the best way to find the registries directory?

server/server.go
```diff
@@ -943,3 +948,81 @@ func isNotFound(err error) bool {
...
+				pendingReload = true
+				go func() {
+					defer func() {
+						mu.Lock()
+						pendingReload = false
+						mu.Unlock()
+					}()
+					// Wait for some time before accepting more requests
+					time.Sleep(eventStaggerDuration)
+					if event.Op&(fsnotify.Write|fsnotify.Create|fsnotify.Remove|fsnotify.Chmod) != 0 {
```
haircommander, line 1008: hm I'm unsure about Chmod

server/server.go
```diff
@@ -943,3 +948,81 @@ func isNotFound(err error) bool {
...
+							log.Errorf(ctx, "Failed to reload registry configuration: %v", err)
+						}
+					}
+				}()
+			}
+			mu.Unlock()
+		case <-ticker.C:
+			mu.Lock()
+			pendingReload = false
+			mu.Unlock()
```
haircommander, line 1020: it's not clear to me what this is doing

server/server.go
sohankunkerkar: I need to check this data once.

server/server.go
sohankunkerkar: I tried to address this comment: https://github.com/cri-o/cri-o/pull/7788#discussion_r1498630932

server/server.go
```diff
@@ -943,3 +948,81 @@ func isNotFound(err error) bool {
...
+			}
+			mu.Lock()
+			if !pendingReload {
+				pendingReload = true
+				go func() {
+					defer func() {
+						mu.Lock()
+						pendingReload = false
+						mu.Unlock()
+					}()
```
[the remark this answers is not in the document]
sohankunkerkar: The use of defer in this context ensures that the mutex is unlocked even if there's a panic within the goroutine. Although the function itself doesn't explicitly return, it can exit prematurely due to a panic.

server/server.go
sohankunkerkar: This ensures that even if there are no file events triggering a reload, the `pendingReload` flag will eventually be reset, allowing the system to accept new events again. So, when <-ticker.C is received, it resets the `pendingReload` flag to false after the specified time interval.

server/server.go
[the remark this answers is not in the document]
haircommander: true but wouldn't cri-o crash if there was a panic in the goroutine? it's not like we have external cleanup we need to do

server/server.go
sohankunkerkar: I'm not using this stuff anymore. Leftover from the previous push.

sohankunkerkar: I guess for environments with relatively low event frequency, a stagger duration of around 100-200 milliseconds may suffice.

server/server.go
```diff
@@ -943,3 +948,79 @@ func isNotFound(err error) bool {
...
+// configuration when a file is modified, created or deleted in the directory.
+func (s *Server) startReloadWatcherForMirrorRegistries(ctx context.Context, registriesConfDDir string) {
+	if registriesConfDDir == "" {
+		log.Infof(ctx, "No registries.conf.d directory specified, defaulting to /etc/containers/registries.conf.d")
+		registriesConfDDir = "/etc/containers/registries.conf.d"
+	}
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Fatalf(ctx, "Failed to create new watcher: %v", err)
+		return
```
saschagrunert, line 964: No need to return, we will `exit(1)` on `Fatalf`

server/server.go
```diff
@@ -943,3 +948,79 @@ func isNotFound(err error) bool {
...
+	}
+	defer watcher.Close()
+
+	log.Infof(ctx, "Registered reload watcher for mirror registries configuration")
+	done := make(chan struct{})
+	go s.reloadMirrorRegistriesConfiguration(ctx, watcher, done, registriesConfDDir)
```
saschagrunert, line 970: Do we need this additional goroutine? We just create the routine and then wait for `done`. Everything could be inline from my point of view since we're already in a new go routine.

server/server.go
```diff
@@ -943,3 +948,79 @@ func isNotFound(err error) bool {
...
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+			// Skip events for files starting with a dot
+			if strings.HasPrefix(filepath.Base(event.Name), ".") {
+				continue
+			}
```
saschagrunert, line 996: Should we be more strict and only watch for `.conf` files? Editors or programs may create backup files which we would pickup.

server/server.go
```diff
@@ -943,3 +948,79 @@ func isNotFound(err error) bool {
...
+func (s *Server) reloadMirrorRegistriesConfiguration(ctx context.Context, watcher *fsnotify.Watcher, done chan struct{}, path string) {
+	defer close(done)
+	if err := watcher.Add(path); err != nil {
+		log.Errorf(ctx, "Failed to add watcher for path %q: %s", path, err)
+		return
+	}
+	var (
+		mu            sync.Mutex
+		pendingReload bool
+	)
```
saschagrunert, line 983: Can we replace the mutex + bool with a semaphore?

sohankunkerkar: Yeah, that's a good suggestion.

server/server.go
sohankunkerkar: Yup, let me fixed it.

server/server.go
```diff
@@ -943,3 +948,68 @@ func isNotFound(err error) bool {
...
+
+			// Check for .conf files only
+			if !strings.HasSuffix(filepath.Base(event.Name), ".conf") {
+				continue
+			}
+
+			<-reloadLock // Acquire the semaphore
+			go func() {
+				// Wait for some time before accepting more requests
+				time.Sleep(eventStaggerDuration)
```
haircommander, line 996: The net effect of this is still a bunch of routines queuing up to call ReloadRegistries. From what I can tell they'll all still enter this block (you only send one event on the channel). A WaitGroup may achieve what you're looking for here

haircommander: hm actually a waitgroup doesn't do it either...

sohankunkerkar: In that case, we need to rely on the rate limiter or the sliding window approach that I proposed earlier: https://github.com/cri-o/cri-o/pull/7788#discussion_r1513588252

server/server.go
```diff
@@ -531,7 +532,11 @@ func New(
 	log.Debugf(ctx, "Sandboxes: %v", s.ContainerServer.ListSandboxes())
 
 	s.startReloadWatcher(ctx)
-
+	if s.config.AutomaticallyReloadRegistries {
+		go func() {
```
mtrmac, line 536: Does the `func(){}()` do anything here?

server/server.go
```diff
@@ -943,3 +948,68 @@ func isNotFound(err error) bool {
...
+	}
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Fatalf(ctx, "Failed to create new watcher: %v", err)
+	}
+	defer watcher.Close()
+
+	log.Infof(ctx, "Registered reload watcher for mirror registries configuration")
+
+	s.reloadMirrorRegistriesConfiguration(ctx, watcher, registriesConfDDir)
```
mtrmac, line 969: (It’s unclear to me what is the benefit of splitting this into two functions, and if the first one is “start reload watcher”, shouldn’t this one be “reload watcher”? “reload” is not quite an accurate name for the function operating on a watcher. See also elsewhere.)

server/server.go
```diff
@@ -943,3 +948,68 @@ func isNotFound(err error) bool {
...
+}
+
+func (s *Server) reloadMirrorRegistriesConfiguration(ctx context.Context, watcher *fsnotify.Watcher, path string) {
+	if err := watcher.Add(path); err != nil {
```
mtrmac, line 973: Everything up to this point could happen in the main goroutine, giving the code more control about error handling.

(and, see elsewhere, _that_ might be the most natural point to split the two functions.)

server/server.go
```diff
@@ -943,3 +948,68 @@ func isNotFound(err error) bool {
...
+			if !strings.HasSuffix(filepath.Base(event.Name), ".conf") {
+				continue
+			}
+
+			<-reloadLock // Acquire the semaphore
+			go func() {
+				// Wait for some time before accepting more requests
+				time.Sleep(eventStaggerDuration)
+
+				if event.Op&(fsnotify.Write|fsnotify.Create|fsnotify.Remove|fsnotify.Chmod) != 0 {
```
mtrmac, line 998: Why doesn’t this happen before the `time.Sleep`? (And before taking the lock?)

server/server.go
```diff
@@ -943,3 +948,68 @@ func isNotFound(err error) bool {
...
+}
+
+func (s *Server) reloadMirrorRegistriesConfiguration(ctx context.Context, watcher *fsnotify.Watcher, path string) {
+	if err := watcher.Add(path); err != nil {
+		log.Errorf(ctx, "Failed to add watcher for path %q: %s", path, err)
+		return
+	}
+
+	reloadLock := make(chan struct{}, 1)
+	reloadLock <- struct{}{}
```
mtrmac, line 979: This might not be _quite_ the approach? If there is a bust of activity, generating 5 events, AFAICS this going to result in 5 reloads, over 5*`eventStaggerDuration`.

I’d expect the loop to look for changes to “settle”, e.g. by arming/re-arming a timer on every change, and triggering a reload if the timer fires without any more changes arriving in the meantime.

server/server.go
mtrmac: The c/image answer is “don’t”.

There _is_ `sysregistriesv2.ConfigDirPath`, but that’s deprecated because we added more places where configuration can be found since, and c/image doesn’t want to constrain itself for the future not to ever add any _more_ places; so a stable API like `ConfigDirPath` promising “this is where data will be read from” is making unsustainable commitments about the future.

So we now have `ConfigurationSourceDescription` but that’s _explicitly_ designed to not be machine-consumable. (Some of the sources might not files at all, in principle…)

---

I realize that’s not all that helpful.

The supported path is to invert the flow, for the caller to _define_ the location (via `SystemContext.SystemRegistriesConfDirPath`) and for c/image to respect that choice. But that’s, of course, not the same thing, and basically not answering the “so which path should be used” question at all.

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

pkg/config/config.go
```diff
@@ -893,11 +897,12 @@ func DefaultConfig() (*Config, error) {
...
-			PauseImage:         DefaultPauseImage,
-			PauseCommand:       "/pause",
-			ImageVolumes:       ImageVolumesMkdir,
-			SignaturePolicyDir: "/etc/crio/policies",
+			DefaultTransport:              "docker://",
+			PauseImage:                    DefaultPauseImage,
+			PauseCommand:                  "/pause",
+			ImageVolumes:                  ImageVolumesMkdir,
+			SignaturePolicyDir:            "/etc/crio/policies",
+			AutomaticallyReloadRegistries: false,
```
kolyshkin, line 905: Don't have to explicitly initialize it to `false` as this is a default value for bool (i.e. it's initialized to false if you omit explicit initialization).

server/server.go
```diff
@@ -943,3 +946,75 @@ func isNotFound(err error) bool {
 
 	return false
 }
+
+// startWatcherForMirrorRegistries starts a new goroutine that
```
kolyshkin, line 950: This description is incorrect as this method does not start a new goroutine.

test/reload_system.bats
```diff
@@ -6,11 +6,14 @@ load helpers
 function setup() {
 	setup_test
 	TEST_IMAGE=quay.io/saschagrunert/hello-world
+	TEST_IMAGE2=quay.io/crio/hello-wasm
 	export CONTAINER_REGISTRIES_CONF="$TESTDIR/containers/registries.conf"
-	mkdir "$TESTDIR/containers"
+	export CONTAINER_REGISTRIES_CONF_DIR="$TESTDIR/containers/registries.conf.d"
+	mkdir -p "$TESTDIR/containers/registries.conf.d"
```
kolyshkin, line 12: nit: you can do `mkdir -p "CONTAINER_REGISTRIES_CONF_DIR"` here.

test/reload_system.bats
```diff
@@ -6,11 +6,14 @@ load helpers
...
 	printf "[[registry]]\nlocation = 'quay.io/saschagrunert'\nblocked = true" \
 		>> "$CONTAINER_REGISTRIES_CONF"
-	start_crio
```
kolyshkin, line 13: Can you explain why this line is removed? Was it redundant?

kolyshkin: nevermind

test/reload_system.bats
```diff
@@ -21,12 +24,17 @@ function expect_log_success() {
 	wait_for_log "applied new registry configuration"
 }
 
+function expect_log_warning() {
+	wait_for_log "rate limit exceeded, skipping reload"
+}
+
```
kolyshkin, line 30: I don't see this being used anywhere in this commit.

test/reload_system.bats
```diff
@@ -60,3 +70,33 @@ function expect_log_failure() {
...
 }
+
+@test "system registries should succeed with new registry without reload" {
+	# given
+	cat << EOF > "$CRIO_CONFIG_DIR/00-automaticReload.conf"
+[crio.image]
+automatically_reload_registries = true
+EOF
+	start_crio
+	replace_config "log_level" "debug"
```
kolyshkin, line 81: Why is this done after, not before starting crio?

test/reload_system.bats
sohankunkerkar: Removed it

test/reload_system.bats
sohankunkerkar: I added that in a drop in config.

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

docs/crio.8.md
```diff
@@ -165,6 +166,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--apparmor-profile**="": Name of the apparmor profile to be used as the runtime's default. This only takes effect if the user does not specify a profile via the Kubernetes Pod's metadata annotation. (default: "crio-default")
 
+**--automatically-reload-registries**: if true, CRI-O will automatically reload the mirror registry when there is an update to the 'registries.conf.d' directory.
```
kwilczynski, line 169: This and other instances of this help sentence.

```suggestion
**--automatically-reload-registries**: If true, CRI-O will automatically reload the mirror registry when there is an update to the 'registries.conf.d' directory.
```

docs/crio.conf.5.md
```diff
@@ -476,6 +476,9 @@ CRI-O reads its configured registries defaults from the system wide containers-r
 **separate_pull_cgroup**=""
   [EXPERIMENTAL] If its value is set, then images are pulled into the specified cgroup.  If its value is set to "pod", then the pod's cgroup is used.  It is currently supported only with the systemd cgroup manager.
 
+**automatically_reload_registries**=false
+ Enable/Disable the automatic reloading of mirror registry in CRI-O when there is an update to the 'registries.conf.d' directory. Default value is set to 'false'.
```
kwilczynski, line 480: This sentence is different than the one for the command-line switch, should both be identical?

internal/criocli/criocli.go
```diff
@@ -922,6 +925,12 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			EnvVars: []string{"CONTAINER_BIG_FILES_TEMPORARY_DIR"},
 			Value:   defConf.BigFilesTemporaryDir,
 		},
+		&cli.BoolFlag{
+			Name:    "automatically-reload-registries",
+			Usage:   "if true, CRI-O will automatically reload the mirror registry when there is an update to the 'registries.conf.d' directory.",
```
kwilczynski, line 930: Ditto.

pkg/config/template.go
```diff
@@ -1472,6 +1477,13 @@ const templateStringCrioImageBigFilesTemporaryDir = `# Temporary directory to us
 
 `
 
+const templateStringCrioImageAutomaticallyReloadRegistries = `# automatically_reload_registries determines whether to
+# enable/disable the reloading of mirror registry in CRI-O when there is an update to the 'registries.conf.d' directory.
+# Default value is set to 'false'
```
kwilczynski, line 1482: Ditto.

server/server.go
```diff
@@ -946,3 +948,80 @@ func isNotFound(err error) bool {
 
 	return false
 }
+
+// startWatcherForMirrorRegistries sets up a file watcher to monitor changes
+// in the "registries.conf.d" directory (default: "/etc/containers/registries.conf.d").
+// It then delegates the monitoring task to watchAndReloadMirrorRegistriesConfiguration,
```
kwilczynski, line 954: Perhaps? Unless there is something missing there and would continue in the next line.

```suggestion
// It then delegates the monitoring task to watchAndReloadMirrorRegistriesConfiguration.
```

server/server.go
```diff
@@ -946,3 +948,80 @@ func isNotFound(err error) bool {
...
+func (s *Server) startWatcherForMirrorRegistries(ctx context.Context, registriesConfDDir string) {
+	if registriesConfDDir == "" {
+		log.Infof(ctx, "No registries.conf.d directory specified, defaulting to /etc/containers/registries.conf.d")
+		registriesConfDDir = "/etc/containers/registries.conf.d"
```
kwilczynski, line 958: This could be factored out into a constant.

server/server.go
```diff
@@ -946,3 +948,80 @@ func isNotFound(err error) bool {
...
+				close(reloadChannel)
+				return
+			}
+			if !strings.HasSuffix(filepath.Base(event.Name), ".conf") {
+				continue
+			}
+			if event.Op&(fsnotify.Write|fsnotify.Create|fsnotify.Remove|fsnotify.Chmod) != 0 {
+				if timer != nil {
+					if !timer.Stop() {
+						<-timer.C // Drain the channel if Stop fails and timer already expired
```
kwilczynski, line 1009:
```suggestion
						<-timer.C // Drain the channel if Stop() fails and the timer already expired.
```

server/server.go
```diff
@@ -946,3 +948,80 @@ func isNotFound(err error) bool {
...
+					}
+				}
+				timer = time.NewTimer(200 * time.Millisecond)
```
kwilczynski, line 1012: Could be factored out into a named variable. Also, would `Reset()` from the `Timer` type work here? In lieu of creating a new timer every time?

test/reload_system.bats
```diff
@@ -6,11 +6,23 @@ load helpers
...
+	mkdir -p "$CONTAINER_REGISTRIES_CONF_DIR"
 	printf "[[registry]]\nlocation = 'quay.io/saschagrunert'\nblocked = true" \
 		>> "$CONTAINER_REGISTRIES_CONF"
-	start_crio
+	printf "[[registry]]\nlocation = 'quay.io/crio'\nblocked = true" \
+		>> "$CONTAINER_REGISTRIES_CONF_DIR/01-registry.conf"
+}
+
+function drop_in_for_automatically_reload_registries() {
+	cat << EOF > "$CRIO_CONFIG_DIR/00-automatically-reload-registries.conf"
```
kwilczynski, line 20:
```suggestion
	cat <<EOF > "$CRIO_CONFIG_DIR/00-automatically-reload-registries.conf"
```

test/reload_system.bats
```diff
@@ -25,8 +37,24 @@ function expect_log_failure() {
 	wait_for_log "system registries reload failed"
 }
 
+function count_log_entries() {
+	local search_pattern="$1"
+	local count
+
+	# Use grep to count occurrences of the pattern.
+	count=$(grep -ico "$search_pattern" "$CRIO_LOG")
```
kwilczynski, line 45: If you drop `-i`, then the provided pattern match would be more exact. Thoughts?

test/reload_system.bats
```diff
@@ -25,8 +37,24 @@ function expect_log_failure() {
...
+
+	# Check if grep found any occurrences, if not, return 0.
+	if [[ -z "$count" ]]; then
+		echo 0
+	else
+		echo "$count"
+	fi
```
kwilczynski, line 52: At least on Linux using GNU version of grep, with the `-c` and `-o` options, it would return 0 and n, per:

```console
$ t() { grep -co "$1" "$2" ; } 
$ t "something" /etc/resolv.conf
0
$ t "nameserver" /etc/resolv.conf
1
```

As such, you could simply this function a little bit.

completions/bash/crio
```diff
@@ -19,6 +19,7 @@ h
 --address
 --allowed-devices
 --apparmor-profile
+--automatically-reload-registries
```
kwilczynski, line 22: I know that @haircommander asked you to rename this before, but we could keep this shorter and also aligned with the other registries-related options we already have, per:

_(an excerpt from the man page)_

```
        '--registries-conf'
        '--registries-conf-dir'
```

So adding `---registries-watch` or `--registries-watcher` would be short and to the point. The `--automatically-reload-registries` is really a bit of a mouthful. Thoughts?

server/server.go
```diff
@@ -946,3 +948,80 @@ func isNotFound(err error) bool {
...
+	reloadChannel := make(chan string, 1)
+	reloadChannel <- ""
+
+	go func() {
+		// The for loop ensures that the channel is properly drained, even if
+		// no new events are received, thus preventing potential deadlocks.
+		// For each event name received, the goroutine checks if it's not
+		// an empty string and then reloads the registries.
+		for evenName := range reloadChannel {
+			if evenName != "" {
```
kwilczynski, line 987: Aside from the initialisation (a few lines above), when did you see an empty event? I wonder if there is a race with the new (a few lines below) goroutine? Not sure.

server/server.go
```diff
@@ -946,3 +948,80 @@ func isNotFound(err error) bool {
...
+				log.Infof(ctx, "File %s modified, created, or deleted, reloading registries configuration", evenName)
```
kwilczynski, line 988: The reason why the registries file changed here is not so important—it could be captured using a `log.Debug()`, but to the end users, the net outcome is a reload, and as such, we could make the message simpler:

```suggestion
				log.Infof(ctx, "File %q changed, reloading registries configuration", evenName)
```

Or perhaps:

```suggestion
				log.Infof(ctx, "File %q changes detected. Reloading registries configuration", evenName)
```

sohankunkerkar: Good point!

server/server.go
sohankunkerkar: I removed that logic altogether.

completions/bash/crio
sohankunkerkar: What do you think about `--auto-reload-registries`? I believe clarity in variable names, reflecting their functionality, is important.

test/reload_system.bats
sohankunkerkar: ok, fixed.

test/reload_system.bats
sohankunkerkar: It looks pretty consistent with the other tests, unless my editor is a bit off

server/server.go
sohankunkerkar: Fixed

test/reload_system.bats
kwilczynski: Oh, if so, then ignore me. It was a habit while reviewing the code.

completions/bash/crio
kwilczynski: @sohankunkerkar, that looks nice! Thank you!

server/server.go
kwilczynski: Thank you, @mtrmac, for the details! Appreciated.

PR #7790, opened 2024-02-19
config, docs, completions: enable NRI by default.

klihub:
#### What type of PR is this?

/kind other

Flip NRI on in the default configuration.

None

As discussed in the last community meeting, this PR flips NRI on in the default configuration.

For maintaining command line backward compatibility this PR leaves the original `--enable-nri` command line option intact, only changing its default to `true` from `false` (as opposed to flipping it to `--disable-nri=[false]`). Hence, NRI can now be disabled on the command line using the `--enable-nri=false` option.

```release-note
Flip NRI on by default.
To revert to a configuration with NRI disabled, either either pass the `--enable-nri=false` option to crio
on the command line, or set the `enable_nri = false` in the `[crio.nri]` section of the configuration.
```

### DIFF
--- a/completions/fish/crio.fish
+++ b/completions/fish/crio.fish
@@ -54,7 +54,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l disable-hostport-mapping -
 complete -c crio -n '__fish_crio_no_subcommand' -f -l drop-infra-ctr -d 'Determines whether pods are created without an infra container, when the pod is not using a pod level PID namespace.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-criu-support -d 'Enable CRIU integration, requires that the criu binary is available in $PATH.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-metrics -d 'Enable metrics endpoint for the server.'
-complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-nri -d 'Enable NRI (Node Resource Interface) support. (default: false)'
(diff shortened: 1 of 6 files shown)


klihub:
/hold

@saschagrunert @haircommander 
I think I know what's wrong. Some of the tests (that used to run with NRI turned off) now clash on the stock NRI socket path when run in parallel... I'll take a look at it.

sohankunkerkar (approved):
/retest
/lgtm

klihub:
@haircommander I recall having seen occasional failures of the timezone tests in images.bat, and IIRC either with [this local timezone test](https://github.com/cri-o/cri-o/blob/bf30758326e51d3fed59d941c97500b6cbc9c653/test/image.bats#L366) failing or the previous one with the explicit/specific timezone test.

I cannot prove it at the moment without doubt, but looking at the actual test case implementation, I suspect the reason for flakes is that it implicitly assumes that running foo=$(date ...) twice, once on the host then in the container and with second resolution will always produce the same timestamp. IOW, it is assumed that running the first date command will never happen close enough to the 'end' of a second with the second date run taking long enough for the time to flip over to the next second... in which case the test is guaranteed to fail.

I have a change (I'd like to call it a fix but in lack of definite proof I won't), which takes a single timestamp in seconds and reuses that in both date formatting to verify that the timezones are correctly taken into account... which is what the test really tries to test.

@haircommander Do you think it is worth PR/fix for me to file ?

haircommander:
> @haircommander Do you think it is worth a PR/fix for me to file 

absolutely!

PR #7814, opened 2024-02-26
In-memory OCI artifact pull

saschagrunert:
#### What type of PR is this?

/kind cleanup

CRI-O now pulls OCI artifacts directly in-memory after the discussions in https://github.com/containers/image/pull/2306 and https://github.com/kubernetes/website/pull/45121.

CRI-O also enforces the config media type
`application/vnd.cncf.seccomp-profile.config.v1+json` for seccomp profiles.

None
None

```release-note
None
```

### DIFF
--- a/internal/config/ociartifact/ociartifact.go
+++ b/internal/config/ociartifact/ociartifact.go
@@ -2,19 +2,23 @@ package ociartifact
 
 import (
 	"context"
+	"encoding/hex"
+	"errors"
 	"fmt"
+	"io"
 
-	"github.com/containers/common/libimage"
-	"github.com/containers/common/pkg/config"
+	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/blobinfocache"
 	"github.com/containers/image/v5/types"
-	"github.com/containers/storage"
 
 	"github.com/cri-o/cri-o/internal/log"
 )
 
 // Impl is the main implementation interface of this package.
 type Impl interface {
-	Pull(context.Context, *types.SystemContext, string) (*Artifact, error)
+	Pull(context.Context, string, *PullOptions) (*Artifact, error)
 }
 
 // New returns a new OCI artifact implementation.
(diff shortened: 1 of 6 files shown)


internal/config/ociartifact/ociartifact.go
```diff
@@ -24,54 +26,81 @@ func New() Impl {
...
+		return nil, fmt.Errorf("artifact needs at least one layer")
 	}
+	layer := layers[0]
 
-	cleanup := func() {
-		if err := image.Unmount(true); err != nil {
-			log.Warnf(ctx, "Unable to unmount OCI artifact path %s: %v", mountPath, err)
-		}
+	bic := blobinfocache.DefaultCache(sys)
+	rc, size, err := src.GetBlob(ctx, layer.BlobInfo, bic)
```
saschagrunert, line 87: > ⚠️💀 Critical: verify each blob’s digest against the data in the manifest.

Is this implicitly done by using `GetBlob` with the data from the manifest (`parsedManifest.LayerInfos()[0].BlobInfo`)?

saschagrunert:
/hold 
for https://github.com/cri-o/cri-o/pull/7814#discussion_r1502285322

kwilczynski:
@saschagrunert, are we going to have more of these _"CRI-O ate my memory!!!"_ issues when we start pulling things to store in a heap every time a container is started?

I wonder if we should use things and store securely using something like the [memguard](https://github.com/awnumar/memguard) package, and be smart about pulling the same thing over and over, perhaps check ETag and such things, and pull if we cannot determine whether the object changed at the remote location?

Not sure. Just thinking out loud here. Thoughts?

saschagrunert:
> @saschagrunert, are we going to have more of these _"CRI-O ate my memory!!!"_ issues when we start pulling things to store in a heap every time a container is started?

So the artifact pull is limited to 1MiB right now (should me much smaller in terms of a seccomp profile) and the memory should be freed after the RPC (when the spec is written to disk).

> I wonder if we should use things and store securely using something like the [memguard](https://github.com/awnumar/memguard) package, and be smart about pulling the same thing over and over, perhaps check ETag and such things, and pull if we cannot determine whether the object changed at the remote location?
>
> Not sure. Just thinking out loud here. Thoughts?

Yes we can follow-up on that specific area. Managing the artifacts can be done on-disk, which is probably a requirement for WASM.

mtrmac (commented):
Thanks, I like this much better than involving c/storage to get the data.

(OTOH this new version always pulls from the registry.)

internal/config/ociartifact/ociartifact.go
```diff
@@ -24,54 +29,90 @@ func New() Impl {
...
+
+	bic := blobinfocache.DefaultCache(sys)
+	rc, _, err := src.GetBlob(ctx, layer.BlobInfo, bic)
+	if err != nil {
+		return nil, fmt.Errorf("get layer blob: %w", err)
+	}
+	defer rc.Close()
+
+	layerBytes, err := io.ReadAll(io.LimitReader(rc, maxArtifactSize))
+	if len(layerBytes) == maxArtifactSize {
```
kwilczynski, line 97: When my image happens to be exactly 1 MiB, it will be a failure?

The `GetBlob()` function returns size if `Content-Size` header is present in the response (we assume remote server is well-behaved), so we could leverage this first to make a check and fail early (no need to actually read anything then), and if we got `-1`, then we could rely on this check here.

It would be nice to make this check a > 1 MiB, if at all possible.

Thoughts?

mtrmac: (Compare the implementation in `c/image/internal/iolimits`.)

internal/config/ociartifact/ociartifact.go
```diff
@@ -24,54 +29,90 @@ func New() Impl {
...
 	if err != nil {
-		return nil, fmt.Errorf("mount OCI artifact: %w", err)
+		return nil, fmt.Errorf("read from blob: %w", err)
 	}
 
-	cleanup := func() {
-		if err := image.Unmount(true); err != nil {
-			log.Warnf(ctx, "Unable to unmount OCI artifact path %s: %v", mountPath, err)
-		}
+	hash := sha256.New()
```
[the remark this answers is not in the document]
saschagrunert: Improved as suggested.

internal/config/ociartifact/ociartifact.go
```diff
@@ -24,54 +29,90 @@ func New() Impl {
...
+func (*defaultImpl) Pull(
+	ctx context.Context,
+	sys *types.SystemContext,
+	img string,
+	enforceConfigMediaType string,
+) (*Artifact, error) {
+	log.Infof(ctx, "Pulling OCI artifact from ref: %s", img)
 
-	storeOpts, err := storage.DefaultStoreOptions(false, 0)
+	name, err := libimage.NormalizeName(img)
```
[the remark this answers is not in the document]
saschagrunert: Improved as suggested.

internal/config/ociartifact/ociartifact.go
saschagrunert: Done, fixed.

saschagrunert:
I'll add unit tests in a follow-up PR to the OCI artifact pull implementation.

/unhold

@cri-o/cri-o-maintainers PTAL, too.

internal/config/ociartifact/ociartifact.go
```diff
@@ -24,54 +28,135 @@ func New() Impl {
...
+	if len(layers) < 1 {
+		return nil, errors.New("artifact needs at least one layer")
+	}
+
+	// Just supporting one layer here. This can be later enhanced by extending
+	// the PullOptions.
+	layer := layers[0]
+
+	bic := blobinfocache.DefaultCache(opts.SystemContext)
+	rc, _, err := src.GetBlob(ctx, layer.BlobInfo, bic)
```
kwilczynski, line 99: If you want, you could still leverage the returned blob size here to short-circuit size check and return early.

internal/config/ociartifact/ociartifact.go
```diff
@@ -24,54 +28,135 @@ func New() Impl {
...
+		return nil, fmt.Errorf("exceeded maximum allowed size of %d bytes", limit)
+	}
+
+	return res, nil
+}
+
+func verifyDigest(layer *manifest.LayerInfo, layerBytes []byte) error {
+	expectedDigest := layer.BlobInfo.Digest
+	if err := expectedDigest.Validate(); err != nil {
+		return fmt.Errorf("invalid digest %s: %w", expectedDigest, err)
```
mtrmac, line 142:
```suggestion
		return fmt.Errorf("invalid digest %q: %w", expectedDigest, err)
```
it may be empty or contain control characters

internal/config/ociartifact/ociartifact.go
```diff
@@ -24,54 +28,135 @@ func New() Impl {
...
-	log.Infof(ctx, "Pulling OCI artifact from ref: %s", ref)
+// defaultMaxArtifactSize is the default maximum artifact size.
+const defaultMaxArtifactSize = 1024 * 1024 // 1 MiB
+
+// Pull downloads the artifact content by using the provided image name and the specified options.
+func (*defaultImpl) Pull(ctx context.Context, img string, opts *PullOptions) (*Artifact, error) {
+	log.Infof(ctx, "Pulling OCI artifact from ref: %s", img)
 
-	storeOpts, err := storage.DefaultStoreOptions(false, 0)
+	name, err := reference.ParseNormalizedNamed(img)
```
mtrmac, line 58: (This requires the caller to explicitly include a tag or digest, it does not default to `:latest`. That may well be fine.)

saschagrunert: Fixed by using `reference.TagNameOnly(name)`.

internal/config/ociartifact/ociartifact.go
saschagrunert: Changed as suggested.

internal/config/ociartifact/ociartifact.go
saschagrunert: Yeah I was considering that. But the size only reads the field in the manifest, which can be anything or `-1` (unknown). I'd like to stick to the limit reader to limit the actual size.

kwilczynski: I see the size being taken from the HTTP response, per:

```go
func getBlobSize(resp *http.Response) int64 {
	size, err := strconv.ParseInt(resp.Header.Get("Content-Length"), 10, 64)
	if err != nil {
		size = -1
	}
	return size
}
```

The call chain is `GetBlob()` -> `getExternalBlob()` -> `getBlobSize()`.

I was thinking something along the lines of:

```go
rc, size, err := src.GetBlob(ctx, layer.BlobInfo, bic)
if err != nil {
	return nil, fmt.Errorf("get layer blob: %w", err)
}
defer rc.Close()

maxArtifactSize := defaultMaxArtifactSize
if opts.MaxSize > 0 {
	maxArtifactSize = opts.MaxSize
}
if size != -1 && size > maxArtifactSize + 1 {
	return nil, fmt.Errorf("exceeded maximum allowed size of %d bytes", maxArtifactSize)
}
```

Basically, the reader passed is reading the content of the HTTP response. So, we might as well leverage the size of the response body (content) if set to avoid having to read it in the first place.

Am I thinking wrong about this?

saschagrunert: Ah, that's interesting. Yeah I'll add the code path to early exit :+1:

mtrmac: Well, if the input is OCI, the expected size is already in the manifest. No need to even trigger a HTTP request.

(It’s still a good idea to use a `LimitReader` because the registry may be malicious. Actually the expected limit could be `min(BlobInfo.Size, maxArtifactSize)`, _typically_ that’s much less than 1 MB.)

PR #7818, opened 2024-02-26
server: misc swap fixes

haircommander:
- unify swap setting between update and create into one function
- in create: don't rely on memory != 0 to set swap
- in update: take into account memory swap values sent from kubelet

#### What type of PR is this?

/kind bug

```release-note
fix a bug where on create a container wouldn't be given swap if memory is not specified, and on update swap wasn't taken into account
```


iholder101:
@haircommander is this still relevant?

PR #7827, opened 2024-02-28
Do not overwrite stored latency/governor on container restart

bartwensley:
#### What type of PR is this?

/kind bug

When a container in a pod with the cpu-c-states.crio.io or cpu-freq-governor.crio.io annotations is restarted, the high performance pre-start hook is called again. The current code will overwrite the stored pm_qos_resume_latency and/or scaling_governor with those that were written to /sys/devices/system/cpu/cpuX the first time the pre-start hook was called. Then when the pod is deleted, the pre-stop hook will restore the incorrect pm_qos_resume_latency and/or scaling_governor values.

The fix is to only store the pm_qos_resume_latency and/or scaling_governor the first time the pre-start hook is run. If the stored values already exist the pre-start hook will not overwrite them.

None

None

```release-note
None
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -734,20 +734,24 @@ func doSetCPUPMQOSResumeLatency(c *oci.Container, latency, cpuDir, cpuSaveDir st
 		latencyFileOrig := path.Join(cpuPowerSaveDir, "pm_qos_resume_latency_us")
 
 		if latency != "" {
-			// Retrieve the current latency.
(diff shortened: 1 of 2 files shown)


bartwensley:
Adding Peter since he reviewed the original commit for this code.
/cc @haircommander

bartwensley:
Added a new unit test to address the -0.01% change in code coverage (hopefully).

kwilczynski:
@cri-o/cri-o-maintainers, this should be good to go. Needs some flappy testes to be overridden.

saschagrunert (approved):

kwilczynski:
@cri-o/cri-o-maintainers, this probably requires an override.

PR #7836, opened 2024-03-04
Change process metrics collector log levels

kwilczynski:
#### What type of PR is this?

/kind cleanup
/kind other

The number of defunct (zombie) processes, ideally zero, can go up and down on a given node. Such orphaned processes might be a sign of a problem with some software or a script. This is something we track as part of CRI-O's process metrics collection.

Collecting data about processes in such a state is helpful, but there is no need for the metrics collector code, which also runs quite often, to produce a log line every time a defunct process is found. Why? There is nothing CRI-O can do about these processes, and eventually system init process will reparent them and collect their pending statuses, hopefully.

However, given that it might take a while for a defunct process to be taken care of, the current code would produce a substantial number of log lines that added to the total volume of CRI-O logs and are also not so helpful.

Thus, move the log level to debug for when a defunct process has been found. While at it, move the log line about missing a missing PID (process) from debug to warning log level.

Related:

- https://github.com/cri-o/cri-o/pull/5082

None

```release-note
None
```

### DIFF
--- a/internal/process/defunct_processes.go
+++ b/internal/process/defunct_processes.go
@@ -50,11 +50,11 @@ func DefunctProcessesForPath(path string) (defunctCount uint, retErr error) {
 
 		stat, err := processStats(path, name)
 		if err != nil {
-			logrus.Debugf("Failed to get the status of process with PID %s: %v", name, err)
+			logrus.Warnf("Failed to get the status of process with PID %s: %v", name, err)
 			continue
 		}
 		if stat.State == "Z" {
-			logrus.Warnf("Found defunct process with PID %s (%s)", name, stat.Comm)
+			logrus.Debugf("Found defunct process with PID %s (%s)", name, stat.Comm)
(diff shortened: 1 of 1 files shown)


internal/process/defunct_processes.go
```diff
@@ -50,11 +50,11 @@ func DefunctProcessesForPath(path string) (defunctCount uint, retErr error) {
 
 		stat, err := processStats(path, name)
 		if err != nil {
-			logrus.Debugf("Failed to get the status of process with PID %s: %v", name, err)
+			logrus.Warnf("Failed to get the status of process with PID %s: %v", name, err)
```
kwilczynski, line 53: **To to reviewer:**

Even though I moved this log line to the warning level, I am still unsure about how useful this line would also be here. In the end, this is just the metrics collector code.

saschagrunert (approved):

PR #7837, opened 2024-03-04
Add OCI artifact layer cache

saschagrunert:
#### What type of PR is this?

/kind feature

We now utilize an on-disk key/value cache for OCI artifacts. The digests will be used to identify if a layer has to be re-pulled or not. The artifact API now returns a path to the file on disk.

None
None

```release-note
Added OCI artifact layer cache to avoid repeatedly pulling the same container image.
```

### DIFF
--- a/internal/config/ociartifact/impl.go
+++ b/internal/config/ociartifact/impl.go
@@ -3,6 +3,7 @@ package ociartifact
 import (
 	"context"
 	"io"
+	"os"
 
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
@@ -21,6 +22,11 @@ type Impl interface {
 	LayerInfos(manifest.Manifest) []manifest.LayerInfo
 	GetBlob(context.Context, types.ImageSource, types.BlobInfo, types.BlobInfoCache) (io.ReadCloser, int64, error)
 	ReadAll(io.Reader) ([]byte, error)
+	MkdirAll(string, os.FileMode) error
+	ReadDir(string) ([]os.DirEntry, error)
+	ReadFile(string) ([]byte, error)
+	RemoveAll(string) error
+	WriteFile(string, []byte, os.FileMode) error
 }
 
 // defaultImpl is the default implementation for the OCI artifact handling.
@@ -61,4 +67,24 @@ func (*defaultImpl) GetBlob(ctx context.Context, src types.ImageSource, bi types
 
 func (*defaultImpl) ReadAll(r io.Reader) ([]byte, error) {
 	return io.ReadAll(r)
+}
+
+func (*defaultImpl) MkdirAll(path string, perm os.FileMode) error {
+	return os.MkdirAll(path, perm)
+}
+
+func (*defaultImpl) ReadDir(name string) ([]os.DirEntry, error) {
+	return os.ReadDir(name)
+}
+
+func (*defaultImpl) ReadFile(name string) ([]byte, error) {
+	return os.ReadFile(name)
+}
+
+func (*defaultImpl) RemoveAll(path string) error {
+	return os.RemoveAll(path)
+}
+
+func (*defaultImpl) WriteFile(name string, data []byte, perm os.FileMode) error {
+	return os.WriteFile(name, data, perm)
 }
--- a/internal/config/ociartifact/ociartifact.go
+++ b/internal/config/ociartifact/ociartifact.go
@@ -6,6 +6,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
+	"time"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/manifest"
@@ -31,12 +34,30 @@ func New() *OCIArtifact {
 type Artifact struct {
 	// Data is the actual artifact content.
 	Data []byte
+
+	// Path is the file representation of the artifact on disk. Can be empty if
+	// the cache is not used.
+	Path string
 }
 
 // PullOptions can be used to customize the pull behavior.
 type PullOptions struct {
 	// SystemContext is the context used for pull.
 	SystemContext *types.SystemContext
+
+	// CachePath is the path to the used artifact cache. The cache can be
+	// disabled if the value is an empty string.
+	CachePath string
+
+	// CacheEntryMaxAge is the maximum age cache entries can have before
+	// getting removed.
+	//
+	// Defaults to 3 days.
+	// Can be set to a negative value for disabling the removal.
+	//
+	// Note: Items only get removed on calling Pull(), which means they can be
+	// technically older on disk.
+	CacheEntryMaxAge time.Duration
 
 	// EnforceConfigMediaType can be set to enforce a specific manifest config media type.
 	EnforceConfigMediaType string
@@ -46,8 +67,13 @@ type PullOptions struct {
 	MaxSize int
 }
 
-// defaultMaxArtifactSize is the default maximum artifact size.
-const defaultMaxArtifactSize = 1024 * 1024 // 1 MiB
+const (
+	// defaultMaxArtifactSize is the default maximum artifact size.
+	defaultMaxArtifactSize = 1024 * 1024 // 1 MiB
+
+	// defaultCacheEntryMaxAge is the default max age for a cache entry item.
+	defaultCacheEntryMaxAge = 3 * 24 * time.Hour // 3 days
+)
 
 // Pull downloads the artifact content by using the provided image name and the specified options.
 func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (*Artifact, error) {
@@ -98,7 +124,12 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
 
 	// Just supporting one layer here. This can be later enhanced by extending
 	// the PullOptions.
-	layer := layers[0]
+	layer := &layers[0]
+
+	useCache := o.prepareCache(ctx, opts)
+	if artifact := o.tryLookupCache(ctx, useCache, opts, layer); artifact != nil {
+		return artifact, nil
+	}
 
 	bic := blobinfocache.DefaultCache(opts.SystemContext)
 	rc, size, err := o.impl.GetBlob(ctx, src, layer.BlobInfo, bic)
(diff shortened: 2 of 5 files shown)


saschagrunert:
/unhold 

ready for review.

internal/config/ociartifact/ociartifact.go
```diff
@@ -98,7 +111,33 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
 
 	// Just supporting one layer here. This can be later enhanced by extending
 	// the PullOptions.
-	layer := layers[0]
+	layer := &layers[0]
+
+	if opts.DatabasePath == "" {
+		opts.DatabasePath = defaultDatabasePath
+	}
+	db := diskv.New(diskv.Options{
```
harche, line 119: Would it make sense to move this to crio start up? or do we need to have this here for every pull?

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,12 +160,17 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
 		return nil, fmt.Errorf("read with limit: %w", err)
 	}
 
-	if err := verifyDigest(&layer, layerBytes); err != nil {
+	if err := verifyDigest(layer, layerBytes); err != nil {
 		return nil, fmt.Errorf("verify digest of layer: %w", err)
 	}
 
+	if err := o.impl.DatabaseWrite(db, key, layerBytes); err != nil {
```
kwilczynski, line 167: No way to leverage diskv's streaming API here?

Note: diskv's streaming API seems to be a bit buggy at the moment.

internal/config/ociartifact/impl.go
```diff
@@ -19,8 +20,11 @@ type Impl interface {
 	ManifestFromBlob([]byte, string) (manifest.Manifest, error)
 	ManifestConfigInfo(manifest.Manifest) types.BlobInfo
 	LayerInfos(manifest.Manifest) []manifest.LayerInfo
+	DatabaseRead(*diskv.Diskv, string) ([]byte, error)
+	DatabaseErase(*diskv.Diskv, string) error
 	GetBlob(context.Context, types.ImageSource, types.BlobInfo, types.BlobInfoCache) (io.ReadCloser, int64, error)
 	ReadAll(io.Reader) ([]byte, error)
+	DatabaseWrite(*diskv.Diskv, string, []byte) error
```
kwilczynski, line 27: Perhaps?

```suggestion
	DatabaseRead(*diskv.Diskv, string) ([]byte, error)
	DatabaseWrite(*diskv.Diskv, string, []byte) error
	DatabaseErase(*diskv.Diskv, string) error
	GetBlob(context.Context, types.ImageSource, types.BlobInfo, types.BlobInfoCache) (io.ReadCloser, int64, error)
	ReadAll(io.Reader) ([]byte, error)

```

kwilczynski: I am not a fan calling this "database" everywhere. It's a cache, really, and we also use it as such.

It might nice to move this to an interface, then embed it here? This would allow us to have more flexibility over what backend is the cache using–reading and writing through memory for testing? perhaps just a passthrough? 

Perhaps even ability to disable this via a configuration option?

Thoughts?

go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	github.com/opencontainers/runtime-spec v1.1.1-0.20230922153023-c0e90434df2a
 	github.com/opencontainers/runtime-tools v0.9.1-0.20230914150019-408c51e934dc
 	github.com/opencontainers/selinux v1.11.0
+	github.com/peterbourgon/diskv/v3 v3.0.1
```
kwilczynski, line 56: I would prefer if we used something that is maintained. Peter does not seem to be actively maintaining this project any more, at least per some of the bugs there, that would be best closed, per:

- https://github.com/peterbourgon/diskv/issues/63
- https://github.com/peterbourgon/diskv/issues/66

This is a bit of a no-no in my book.

I don't want to chase future bugs in some new dependency if there is no need to.

We could use something like Badger, which has in-memory mode (good for testing), TTL on cached objects, and garbage collection, and even supports encryption (if we ever need such a thing). However, it's a large project primarily focusing on performance, and it's not memory-usage friendly, sadly.

There is also the bbolt, a fork by us, by CoreOS team, to keep BoltDB going. This is something that etcd is using at the moment.

Thoughts?

saschagrunert: I decided against bolt because it's not really made for larger files and does not represent the files on disk in a 1:1 mapping. The maintainability is tricky here, right. Maybe we should implement it on our own? Or just use the parts of diskv we actually require?

internal/config/ociartifact/ociartifact.go
saschagrunert: We need the bytes in any case to validate the digests. That's why I thought we should not write anything to disk which is invalid.

internal/config/ociartifact/ociartifact.go
saschagrunert: Probably, I'll check what can be done.

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+	tryRemoveKeyPath := false
+
+	if value, err := o.impl.ReadFile(key); err == nil {
+		if err := verifyDigest(layer, value); err == nil {
+			log.Infof(ctx, "Using cached artifact layer for digest %q", layer.BlobInfo.Digest)
+			return &Artifact{
+				Data: value,
+				Path: keyPath,
+			}
+		} else {
```
kwilczynski, line 172: You could remove `else` here since there is a `return` above.

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
 }
 
+func (o *OCIArtifact) tryLookupCache(ctx context.Context, opts *PullOptions, layer *manifest.LayerInfo) *Artifact {
+	if opts.CachePath == "" {
+		return nil
+	}
+
+	if err := o.impl.MkdirAll(opts.CachePath, 0o700); err != nil {
+		log.Errorf(ctx, "Unable to create cache path: %v", err)
+		// Unset cache path to avoid write
```
kwilczynski, line 157:
```suggestion
		// Unset cache path to avoid write.
```

internal/config/ociartifact/ociartifact.go
saschagrunert: I think we can have individual caches per API user. So keeping logically in the same way without using `diskv` any more.

internal/config/ociartifact/ociartifact.go
saschagrunert: Done

internal/config/ociartifact/impl.go
saschagrunert: Changed by not using diskv.

internal/config/ociartifact/ociartifact.go
saschagrunert: Yep, fixed :+1:

internal/config/ociartifact/ociartifact.go
saschagrunert: Fixed

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+			log.Infof(ctx, "Using cached artifact layer for digest %q", layer.BlobInfo.Digest)
+			return &Artifact{
+				Data: value,
+				Path: keyPath,
+			}
+		}
+
+		log.Warnf(ctx, "Unable to verify cached digest (removing it): %v", err)
+		tryRemoveKeyPath = true
+	} else if !os.IsNotExist(err) {
```
kwilczynski, line 176: If the file does not exist, then attempting to remove it would create a potentially spurious warning log message. Or, am I wrong about this? We could just ignore non-existent file (key).

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+	if value, err := o.impl.ReadFile(key); err == nil {
+		if err := verifyDigest(layer, value); err == nil {
+			log.Infof(ctx, "Using cached artifact layer for digest %q", layer.BlobInfo.Digest)
+			return &Artifact{
+				Data: value,
+				Path: keyPath,
+			}
+		}
+
+		log.Warnf(ctx, "Unable to verify cached digest (removing it): %v", err)
```
kwilczynski, line 174:
```suggestion
		log.Warnf(ctx, "Unable to verify cached artifact layer: %v", err)
```

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+		tryRemoveKeyPath = true
+	} else if !os.IsNotExist(err) {
+		log.Warnf(ctx, "Unable to read cache key path (removing it): %v", err)
```
kwilczynski, line 177: Why? The "key path" is an implementation detail here.

```suggestion
		log.Warnf(ctx, "Unable to read cached artifact layer: %v", err)
```

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+		tryRemoveKeyPath = true
+	}
+
+	if tryRemoveKeyPath {
+		if err := o.impl.RemoveAll(keyPath); err != nil {
```
kwilczynski, line 182:
```suggestion
		log.Infof(ctx, "Removing cached artifact layer for digest %q", layer.BlobInfo.Digest)
		if err := o.impl.RemoveAll(keyPath); err != nil {
```

If the above messages are updated, some paper trail will be in order.

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+
+func (o *OCIArtifact) tryWriteCache(ctx context.Context, opts *PullOptions, layer *manifest.LayerInfo, layerBytes []byte) (keyPath string) {
+	if opts.CachePath == "" {
+		return ""
+	}
+
+	_, keyPath = keyAndPath(opts, layer)
+
+	if err := o.impl.WriteFile(keyPath, layerBytes, 0o600); err != nil {
+		log.Errorf(ctx, "Unable to write artifact to cache: %v", err)
```
kwilczynski, line 198: Nit. Only for consistency.

```suggestion
		log.Errorf(ctx, "Unable to write artifact layer to cache: %v", err)
```

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +135,79 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+		return ""
+	}
+
+	return keyPath
+}
+
+func keyAndPath(opts *PullOptions, layer *manifest.LayerInfo) (key, keyPath string) {
+	key = string(layer.BlobInfo.Digest)
+	keyPath = filepath.Join(opts.CachePath, key)
+	return key, keyPath
```
kwilczynski, line 208: Return with named return values? Best of both worlds 😄 

I know that the linter will complain otherwise, potentially.

internal/config/ociartifact/ociartifact.go
saschagrunert: Fixed.

internal/config/ociartifact/ociartifact.go
saschagrunert: We remove the file only if it exists :)

saschagrunert:
Added a cache max age. @cri-o/cri-o-maintainers @kwilczynski  PTAL again

internal/config/ociartifact/ociartifact.go
```diff
@@ -46,8 +67,13 @@ type PullOptions struct {
...
 }
 
-// defaultMaxArtifactSize is the default maximum artifact size.
-const defaultMaxArtifactSize = 1024 * 1024 // 1 MiB
+const (
+	// defaultMaxArtifactSize is the default maximum artifact size.
+	defaultMaxArtifactSize = 1024 * 1024 // 1 MiB
+
+	// defaultCacheEntryMaxAge is the default max age for a cache entry item.
+	defaultCacheEntryMaxAge = 3 * 24 * time.Hour
```
kwilczynski, line 75:
```suggestion
	defaultCacheEntryMaxAge = 3 * 24 * time.Hour // 3 days
```

internal/config/ociartifact/ociartifact.go
```diff
@@ -121,15 +152,115 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+		for _, entry := range entries {
+			info, err := entry.Info()
+			if err != nil {
+				log.Errorf(ctx, "Unable to get cache item info for %q: %v", entry.Name(), err)
+				continue
+			}
+
+			if diff := now.Sub(info.ModTime()); diff > maxAge {
+				p := filepath.Join(opts.CachePath, info.Name())
+				log.Infof(ctx, "Deleting old cache item %q (age: %v)", p, diff.Round(time.Second))
```
kwilczynski, line 199: Nit. If we want to be consistent with:

```golang
log.Infof(ctx, "Removing cached artifact layer for digest %q", layer.BlobInfo.Digest)
```

However, perhaps using "Deleting" in both places would be better sounding?

```suggestion
				log.Infof(ctx, "Removing old cache item %q (age: %v)", p, diff.Round(time.Second))
```

kwilczynski: Also, should we call things "cache item" or "artifact layer" throughout? What would be more appropriate?

Thoughts?

saschagrunert: Fixed, went for `artifact layer` all over the place.

internal/config/ociartifact/ociartifact.go
saschagrunert: Fixed as suggested.

kwilczynski:
/release-note-edit ```release-note Add OCI artifact layer cache to avoid repeatedly pulling the same container image.```

kwilczynski:
/release-note-edit ```release-note Add OCI artifact layer cache to avoid repeatedly pulling the same container image.```

kwilczynski:
/remove-label release-note-none
/release-note-edit ```release-note Add OCI artifact layer cache to avoid repeatedly pulling the same container image.```

kwilczynski:
> /release-note-edit ```release-note Add OCI artifact layer cache to avoid repeatedly pulling the same container image.```

@saschagrunert, not sure why bot does not seem to do what I ask. 🙂

If you could help me out here. Thank you!

saschagrunert:
Done :+1:

PR #7850, opened 2024-03-06
Add image removal lock to image server

saschagrunert:
#### What type of PR is this?

/kind bug

We now avoid removing images when a sandbox or container creation is in progress. This should close the gap in the outlined corner case in:

https://github.com/kubernetes/kubernetes/issues/123631

Refers to https://github.com/kubernetes/kubernetes/pull/123711 as well.

None

None

```release-note
Fixed bug where images can be removed during sandbox or container creation.
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -95,6 +95,7 @@ type imageService struct {
 	ctx                  context.Context
 	config               *config.Config
 	regexForPinnedImages []*regexp.Regexp
+	imageRemovalLock     sync.Map
 }
 
 // ImageBeingPulled map[string]bool to keep track of the images haven't done pulling.
@@ -155,6 +156,15 @@ type ImageServer interface {
 
 	// UpdatePinnedImagesList updates pinned and pause images list in imageService.
 	UpdatePinnedImagesList(imageList []string)
+
+	// ImageRemovalLock returns the removal lock for the specified image storage reference.
+	ImageRemovalLock(id StorageImageID) *sync.RWMutex
+}
+
+// ImageRemovalLock returns the removal lock for the specified image storage reference.
+func (svc *imageService) ImageRemovalLock(id StorageImageID) *sync.RWMutex {
+	value, _ := svc.imageRemovalLock.LoadOrStore(id, &sync.RWMutex{})
+	return value.(*sync.RWMutex) // nolint:errcheck,forcetypeassert // type matches line above
 }
 
 func parseImageNames(image *storage.Image) (someName *RegistryImageReference, tags []reference.NamedTagged, digests []reference.Canonical, err error) {
@@ -729,6 +739,14 @@ func (svc *imageService) DeleteImage(systemContext *types.SystemContext, id Stor
 	if err != nil {
 		return err
 	}
+
+	// Avoid removing images which are currently used for container creation
+	imageRemovalLock := svc.ImageRemovalLock(id)
+	imageRemovalLock.Lock()
+	defer imageRemovalLock.Unlock()
+
+	// Cleanup memory when the image got removed
+	defer svc.imageRemovalLock.Delete(id)
 
 	return ref.DeleteImage(svc.ctx, systemContext)
 }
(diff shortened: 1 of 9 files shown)


internal/storage/image.go
```diff
@@ -94,6 +94,7 @@ type imageService struct {
 	ctx                  context.Context
 	config               *config.Config
 	regexForPinnedImages []*regexp.Regexp
+	imageRemovalLock     sync.Map
```
kwilczynski, line 97: Question about locks here: we already have `imageCacheLock`. Which is also just a normal `sync.Mutex`, nothing as fancy as `sync.Map`. Would it be possible to use it too? Rename to some generic `imageLock` or just embedd directly in the type, and then use? Or do we need `sync.Map`?

kwilczynski: Looking at things more closely.  I suppose, having a per-image lock has the advantage of removing unnecessary contention since it allows for more granural locking.

internal/storage/image.go
```diff
@@ -151,6 +152,14 @@ type ImageServer interface {
 	// CandidatesForPotentiallyShortImageName resolves an image name into a set of fully-qualified image names (domain/repo/image:tag|@digest).
 	// It will only return an empty slice if err != nil.
 	CandidatesForPotentiallyShortImageName(systemContext *types.SystemContext, imageName string) ([]RegistryImageReference, error)
+	// ImageRemovalLock returns the removal lock for the specified image storage reference.
+	ImageRemovalLock(id StorageImageID) *sync.RWMutex
+}
+
+// ImageRemovalLock returns the removal lock for the specified image storage reference.
+func (svc *imageService) ImageRemovalLock(id StorageImageID) *sync.RWMutex {
+	value, _ := svc.imageRemovalLock.LoadOrStore(id, &sync.RWMutex{})
```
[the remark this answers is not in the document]
kwilczynski: @haircommander, it's not an error.

It's a boolean to indicate whether the value was loaded or stored. In case you need to know which operations succeeded.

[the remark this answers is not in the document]
haircommander: RTFM @haircommander :)

server/container_create_linux.go
```diff
@@ -245,6 +245,12 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 
 	imageName := imgResult.SomeNameOfThisImage
 	imageID := imgResult.ID
+
+	// Avoid image removal during container creation
+	imageRemovalLock := s.StorageImageServer().ImageRemovalLock(imageID)
+	imageRemovalLock.RLock()
+	defer imageRemovalLock.RUnlock()
```
haircommander, line 252: can we move this below the container storage call? after we mount the image, the GC will fail, so we can continue with image deltion attempts and report an error to the kubelet (instead of stalling and potentially timing out the gc request)

haircommander: this being the unlock

saschagrunert: Yes, like this now?

server/container_create_linux.go
```diff
@@ -290,6 +291,11 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		}
 	}()
 
+	// Avoid image removal during container creation
+	imageRemovalLock := s.StorageImageServer().ImageRemovalLock(imageID)
+	imageRemovalLock.RLock()
+	defer imageRemovalLock.RUnlock()
+
```
haircommander, line 298: i don't know if we need this here. The StorageRuntimeServer().CreateContainer() will mount the image into the container's new rootfs, so a RemoveImage request will fail (AFAIU). Why do we need it after that?

saschagrunert: I think I misunderstood https://github.com/cri-o/cri-o/pull/7850#discussion_r1514831228, moved it back to the previous location.

I did a test with a `time.Sleep(…)` and was able to produce an image removal during container creation. With the code in this PR, that issue was fixed by waiting on the deletion and then returning the right error (image in use by container).

PR #7854, opened 2024-03-07
OCPBUGS-28981: Add exponential backoff to the container stop loop

kwilczynski:
#### What type of PR is this?

/kind bug

Currently, when CRI-O attempts to stop a container where the process within, especially an init process (the so-called "PID 1"), is in an uninterruptible blocking state (for example, it's sleeping and waiting for a disk I/O completion, etc.), CRI-O will enter a broken state where it tries to delivery termination signals to such a process as fast as possible.

Nonetheless, a blocked process might not promptly respond to the signals delivered, causing CRI-O to enter a "busy loop" while it repeatedly tries to signal delivery. This seemingly unbound loop can render CRI-O unresponsive and result in high CPU usage while it happens.

Thus, add exponential backoff support to the container stop loop to fix the possible busy loop issue irregardless of the current state of the process to be terminated. The exponential backoff will stagger termination signals delivery for as long as the process is still running, allowing it to eventually terminate on its own volition (or crash, whichever comes first).

Related to:

- https://github.com/cri-o/cri-o/pull/7736

None

None

```release-note
- Add exponential backoff to the container stop loop to fix a busy loop issue when a process running with the container is in an uninterruptible blocking state and would become unresponsive to signals delivery during container termination.
- Add a new error log line to notify the user that a process (the container init) has been blocked in uninterruptible sleep for some time.```

### DIFF
--- a/contrib/test/ci/vars.yml
+++ b/contrib/test/ci/vars.yml
@@ -129,6 +129,7 @@ kata_skip_ctr_tests:
   - 'test "annotations passed through"'
   - 'test "ctr with default_env set in configuration"'
   - 'test "ctr with absent mount that should be rejected"'
+  - 'test "ctr stop loop kill retry attempts"'
 kata_skip_devices_tests:
   - 'test "additional devices permissions"'
   - 'test "annotation devices support"'
--- a/internal/oci/container_linux.go
+++ b/internal/oci/container_linux.go
@@ -3,11 +3,27 @@ package oci
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/containers/common/pkg/cgroups"
 	"github.com/cri-o/cri-o/internal/log"
 	"golang.org/x/net/context"
+)
+
+const (
+	procStatFile = "/proc/%d/stat"
+
+	// Fields from the /proc/<PID>/stat file. see:
+	//   https://man7.org/linux/man-pages/man5/proc.5.html
+	//
+	// Field no. 3, the process state, such as "R", "S", "D", etc.
+	// Field no. 22, the process start time, using clock ticks since the system boot.
+	//
+	// The index values are shifted three fields to the left
+	// with the process name field skipped over during parsing.
+	stateFieldIndex     = 0
+	startTimeFieldIndex = 19
 )
 
 // CleanupConmonCgroup cleans up conmon's group when using cgroupfs.
@@ -31,50 +47,60 @@ func (c *Container) CleanupConmonCgroup(ctx context.Context) {
 	}
 }
 
-// SetSeccompProfilePath sets the seccomp profile path
+// SetSeccompProfilePath sets the seccomp profile path.
 func (c *Container) SetSeccompProfilePath(pp string) {
 	c.seccompProfilePath = pp
 }
 
-// SeccompProfilePath returns the seccomp profile path
+// SeccompProfilePath returns the seccomp profile path.
 func (c *Container) SeccompProfilePath() string {
 	return c.seccompProfilePath
 }
 
-// getPidStartTime reads the kernel's /proc entry for stime for PID.
-// inspiration for this function came from https://github.com/containers/psgo/blob/master/internal/proc/stat.go
-// some credit goes to the psgo authors
-func getPidStartTime(pid int) (string, error) {
-	return GetPidStartTimeFromFile(fmt.Sprintf("/proc/%d/stat", pid))
+// GetPidStartTimeFromFile reads a file as if it were a /proc/<PID>/stat file,
+// looking for a process start time for a given PID. It is abstracted out to
+// allow for unit testing.
+func GetPidStartTimeFromFile(file string) (string, error) {
+	_, startTime, err := getPidStatDataFromFile(file)
+	return startTime, err
 }
 
-// GetPidStartTime reads a file as if it were a /proc/$pid/stat file, looking for stime for PID.
-// It is abstracted out to allow for unit testing
-func GetPidStartTimeFromFile(file string) (string, error) {
-	data, err := os.ReadFile(file)
+// getPidStartTime returns the process start time for a given PID.
+func getPidStartTime(pid int) (string, error) {
+	_, startTime, err := getPidStatDataFromFile(fmt.Sprintf(procStatFile, pid))
+	return startTime, err
+}
+
+// getPidStatData returns the process state and start time for a given PID.
+func getPidStatData(pid int) (string, string, error) { //nolint:gocritic // Ignore unnamedResult.
+	return getPidStatDataFromFile(fmt.Sprintf(procStatFile, pid))
+}
+
+// getPidStatData parses the kernel's /proc/<PID>/stat file,
+// looking for the process state and start time for a given PID.
+func getPidStatDataFromFile(file string) (string, string, error) { //nolint:gocritic // Ignore unnamedResult.
+	f, err := os.Open(file)
 	if err != nil {
-		return "", fmt.Errorf("%v: %w", err, ErrNotFound)
+		return "", "", err
 	}
-	// The command (2nd field) can have spaces, but is wrapped in ()
-	// first, trim it
-	commEnd := bytes.LastIndexByte(data, ')')
-	if commEnd == -1 {
-		return "", fmt.Errorf("unable to find ')' in stat file: %w", ErrNotFound)
+	defer f.Close()
+
+	data, err := io.ReadAll(io.LimitReader(f, 4096))
+	if err != nil {
+		return "", "", err
 	}
 
-	// start on the space after the command
-	iter := commEnd + 1
-	// for the number of fields between command and stime, trim the beginning word
-	for field := 0; field < statStartTimeLocation-statCommField; field++ {
-		// trim from the beginning to the character after the last space
-		data = data[iter+1:]
-		// find the next space
-		iter = bytes.IndexByte(data, ' ')
-		if iter == -1 {
-			return "", fmt.Errorf("invalid number of entries found in stat file %s: %d: %w", file, field-1, ErrNotFound)
-		}
+	bracket := bytes.LastIndexByte(data, ')')
+	if bracket < 0 {
+		return "", "", fmt.Errorf("unable to find ')' in stat file: %w", ErrNotFound)
 	}
 
-	// and return the startTime (not including the following space)
-	return string(data[:iter]), nil
+	// Skip the process name and the white space after the right bracket.
+	statFields := bytes.Fields(data[bracket+2:])
+
+	if len(statFields) < startTimeFieldIndex+1 {
+		return "", "", fmt.Errorf("unable to parse malformed stat file: %w", ErrNotFound)
+	}
+
+	return string(statFields[stateFieldIndex]), string(statFields[startTimeFieldIndex]), nil
 }
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -34,6 +35,7 @@ import (
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/remotecommand"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+	kclock "k8s.io/utils/clock"
 	utilexec "k8s.io/utils/exec"
 )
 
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: internal/oci/runtime_oci.go [hunk 1]
```diff
@@ -851,11 +857,28 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
...
+	// the last value, it will wrap around, allowing for a shorter interval
+	// to be used again and, as such, for burst periods.
+	getTickerInterval := func() time.Duration {
+		tickerIndex %= len(stopTickerIntervals)
+		t := time.Duration(stopTickerIntervals[tickerIndex]*1000) * time.Millisecond
+		tickerIndex++
+		return t
+	}
+
+	ticker := time.NewTicker(getTickerInterval())
```
haircommander, line 873: I think you can use the ExponentialBackoff in k8s.io/apimachinery/pkg/util/wait/ which I believe will simplify this  a bit

kwilczynski: @haircommander, can do. 👍 

I looked at this first, as we use it throughout the code base, but then decided that having a ability to burst and then also to roll-over to eventually try shorter intervals again and burst, was an interesting idea. Simple to add too.

That said, I will convert it. So we will have progression with an increasing internal between retries up to the limit of 5 minutes, from where it will always use 5 minutes as an interval going forward.

kwilczynski: Done.

internal/oci/runtime_oci.go
```diff
@@ -45,6 +47,15 @@ const (
 	rootFlag = "--root"
 )
 
+var (
+	// Configuration for the stop loop exponential backoff manager.
```
kwilczynski, line 51: **To the reviewer:**

Hopefully, these values are somewhat acceptable. We can always raise the maximum and reset backoff limits as needed to make the loop stagger slightly more.

internal/oci/runtime_oci.go
```diff
@@ -45,6 +47,15 @@ const (
 	rootFlag = "--root"
 )
 
+var (
```
saschagrunert, line 50:
```suggestion
const (
```

And merge with the block above.

saschagrunert (commented):
Is it somehow possible to write an integration test for this?

kwilczynski:
> Is it somehow possible to write an integration test for this?

@saschagrunert, I am open to suggestions.

I use a custom kernel module to switch process states while manually testing and verifying.

internal/oci/runtime_oci.go
kwilczynski: Done.

saschagrunert:
> I use a custom kernel module to switch process states while manually testing and verifying.

That seems to be too much from my point of view. I'm wondering if we could create a workload triggering the backoff at least.

andreaskaris:
> > I use a custom kernel module to switch process states while manually testing and verifying.
> 
> That seems to be too much from my point of view. I'm wondering if we could create a workload triggering the backoff at least.

No kernel module needed, it can be done with a script: https://github.com/andreaskaris/uninterruptible-sleeper/blob/master/uninterruptible.sh | https://github.com/andreaskaris/uninterruptible-sleeper/blob/master/fedora.yaml | https://chrisdown.name/2024/02/05/reliably-creating-d-state-processes-on-demand.html
But you still need to be root / privileged

kwilczynski:
>  [...]
> Or write a dummy runtime that doesn't react to the KILL signal and test with that?

@andreaskaris, this is what I settled with too. Probably the cleanest approach.

internal/oci/runtime_oci.go
```diff
@@ -846,42 +866,55 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
...
+	// The API is due to be deprecated. However, the replacement is completely broken, see:
+	//   https://github.com/kubernetes/kubernetes/issues/118638
+	backoffManager := kwait.NewExponentialBackoffManager(
+		stopInitialBackoff,
+		stopMaximumBackoff,
+		stopResetBackoff,
+		stopBackoffFactor,
+		stopBackoffJitter,
+		&kclock.RealClock{},
+	)
```
haircommander, line 889: nit: personal preference to not define the variable if we don't need it. we can inline it in the BackoffUntil function below

internal/oci/runtime_oci.go
```diff
@@ -821,6 +837,10 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
+	// Allow for SIGINT to correctly interrupt the stop loop, especially
+	// when CRI-O is run directly in the foreground in the terminal.
+	ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
```
haircommander, line 842: I assume you've hit this while debugging. I think it's in part incomplete because I would guess there are other pieces in cri-o we're not correctly catching SIGINT. I am fine with this for now, but maybe we should plumb all the CRI calls with this so we have better INT handling throughout?

internal/oci/runtime_oci.go
```diff
@@ -846,42 +866,55 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
...
+			time.Sleep(stopProcessWatchSleep)
 		}
 	}()
 
 	// Operate in terms of targetTime, so that we can pause in the middle of the operation
 	// to catch a new timeout (and possibly ignore that new timeout if it's not correct to
 	// take a new one).
-	targetTime := time.Unix(1<<50-1, 0)
-	for finished := false; !finished; {
+	targetTime := time.Now().AddDate(+1, 0, 0) // A year from this one.
```
haircommander, line 877: yeah I suppose the end of unix time is a bit much...

internal/oci/runtime_oci.go
kwilczynski: @haircommander, yes.

We need to plumb context correctly where it should have been passed. But this would be more work and out of scope for this change.

A work cut out for a future Pull Request.

internal/oci/runtime_oci.go
kwilczynski: I thought of doing it, but when done the readibility took a bit of a hit for the worse. The large block is in the middle which makes it somewhat unwildely.

Happy to change it for you. 😀

kwilczynski: Done.

kwilczynski:
> Is it somehow possible to write an integration test for this?

@saschagrunert, done.

saschagrunert (approved):

kwilczynski:
/hold

Another issue has been found while working on this fix. I will update the Pull Request accordingly.

kwilczynski:
/unhold

The other issue will be taken care of separately. I don't want to block a release of this specific fix.

haircommander:
kata test failures are legit https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/7854/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1767191175355174912/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt

kwilczynski:
> kata test failures are legit https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/7854/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1767191175355174912/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt

@haircommander, let me see! Thank you!

contrib/test/ci/vars.yml
```diff
@@ -129,6 +129,7 @@ kata_skip_ctr_tests:
   - 'test "annotations passed through"'
   - 'test "ctr with default_env set in configuration"'
   - 'test "ctr with absent mount that should be rejected"'
+  - 'test "ctr stop loop kill retry attempts"'
```
kwilczynski, line 132: **To the reviewer:**

The Runtime VM (Kata) implements its own `StopContainer()` with custom logic, log lines, etc. Hence, the OCI test does not apply to it.

kwilczynski:
/hold

Holding for a moment...

internal/oci/runtime_oci.go
```diff
@@ -809,21 +825,36 @@ func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout in
...
+			),
+		)
 	}
 
 	c.WaitOnStopTimeout(ctx, timeout)
 	return nil
 }
 
-func (r *runtimeOCI) StopLoopForContainer(c *Container) {
+func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager) {
```
kwilczynski, line 846: **To the reviewer:**

I decided to allow for passing a BackoffManager implementation so that we can tweak the one used in the unit tests. Otherwise, it would increase the total run time of the test (nearly double), which is not ideal for a unit test.

kwilczynski:
@cri-o/cri-o-maintainers, I think this could now go in as is.

saschagrunert (approved):

mrunalp:
I don't see the code to detect and log that the process in in D state. I think that's important to add so we can easily classify such issues by looking at the logs.

kwilczynski:
> I don't see the code to detect and log that the process in in D state. I think that's important to add so we can easily classify such issues by looking at the logs.

@mrunalp, the previous implementation wasn't fit for use and was not picked up.

I would like to get this quite urgent bug fix on the way, which addresses the issue of a busy loop in the stop code. We can revisit adding improved reporting on the state of the processes within the container as a follow-up. Would this be acceptable?

kwilczynski:
/hold

I found another issue while doing some manual testing... Putting a hold for now.

kwilczynski:
> > I don't see the code to detect and log that the process in in D state. I think that's important to add so we can easily classify such issues by looking at the logs.
> 
> @mrunalp, the previous implementation wasn't fit for use and was not picked up.
> 
> I would like to get this quite urgent bug fix on the way, which addresses the issue of a busy loop in the stop code. We can revisit adding improved reporting on the state of the processes within the container as a follow-up. Would this be acceptable?

Done.

I've added an error log that will be shown when a process is blocked for some time. The log will be repeated periodically as a constant reminder.

internal/oci/runtime_oci.go
```diff
@@ -832,50 +869,68 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container) {
...
 
 	// Operate in terms of targetTime, so that we can pause in the middle of the operation
 	// to catch a new timeout (and possibly ignore that new timeout if it's not correct to
 	// take a new one).
-	targetTime := time.Unix(1<<50-1, 0)
-	for finished := false; !finished; {
+	targetTime := time.Now().AddDate(+1, 0, 0) // A year from this one.
+
+	blockedTimer := time.AfterFunc(stopProcessBlockedInternal, func() {
+		if state, err := c.Living(); err == nil && state == "D" {
```
andreaskaris, line 892: I'm not sure that this happens *only* on uninterruptible sleep (it's likely only on uninterruptible sleep, but in another customer case under investigation, I could not find evidence that the process in question was in uninterruptible sleep). Anyway, this state is for the runtime's init process, right?
Nit: "Detected process (%d) blocked in uninterruptible sleep for more than %d seconds for container: %s" would IMO make it clearer, especially given that you test for "D"

kwilczynski: @andreaskaris, most of the time, we only track the init PID within the container within CRI-O, so if the process is blocked or waiting either for one of its children or a thread (albeit, this case is slightly different), then you could also see things blocked despite the primary process within the container sleeping with "S" state, etc.

We can add a message that spells out a list of children and threads of the init process with their state, which could help a little more. This is something that I ran through @mrunalp already, but it's also something he wants to include as a later, perhaps follow-up, change.

kwilczynski: [...]
> With the test for "D" only: "Detected process (%d) blocked **in uninterruptible sleep** for more than %d seconds for container: %s" would IMO make it clearer, especially given that you test for "D"

Done.

kwilczynski:
@cri-o/cri-o-maintainers @andreaskaris, please have a look. We need to get this out as soon as possible.

Thank you!

internal/oci/runtime_oci.go
andreaskaris: thanks!

internal/oci/container_linux.go
```diff
@@ -31,50 +34,66 @@ func (c *Container) CleanupConmonCgroup(ctx context.Context) {
...
+
+	// Skip the process name and the white space after the right bracket.
+	fields := bytes.Fields(data[bracket+2:])
+
+	// Fields from the /proc/<PID>/stat file. See:
+	//   https://man7.org/linux/man-pages/man5/proc.5.html
+	//
+	// Field no. 3, the process state, such as "R", "S", "D", etc.
+	// Field no. 22, the process start time (clock ticks) since the system boot.
+	const startTimeFieldLocation = 22
```
andreaskaris, line 90: Not sure if this is a nit or a minuscule bug ..

Not that it would matter because the stat file is 52 fields long (50 when factoring out pid and name), but startTimeFieldLocation 22 is too high, given that you're looking into fields[19].

By explicitly naming the fields, the code would be more readable, and the offset / length error could be avoided:
```
const startTimeFieldLocation = 19
const stateFieldLocation = 0
(...)
if len(fields) < startTimeFieldLocation + 1 {
(...)
return string(fields[procFieldLocation]), string(fields[startTimeFieldLocation]), nil
```

andreaskaris:
Feel free to disregard my nits. Most of the code is too complex for me, so I defer to the maintainers for review. I do like the error message now and the general approach with the backoff looks o.k. to me

internal/oci/container_linux.go
kwilczynski: Surprisingly, the code originally looked like that, but then I decided to eliminate the variables to make it more concise.

kwilczynski: Done. Thank you!

kwilczynski:
> Feel free to disregard my nits. Most of the code is too complex for me, so I defer to the maintainers for review. I do like the error message now and the general approach with the backoff looks o.k. to me

@andreaskaris, no worries! I appreciate every review! :muscle:

kwilczynski:
@cri-o/cri-o-maintainers, I believe this should be good to go.

kwilczynski:
@haircommander, added the `ProcessState()` function as requested, in lieu of modifying `Living()`.

kwilczynski:
OK. Requires manual backport.

PR #7859, opened 2024-03-11
[WIP-Refactor part 1] Move container mounts to internal/factory/container

### LINKED ISSUE #7275: Refactor of taking container creation pieces from server to internal/factory
@haircommander can provide more input about this refactoring item. Feel free to edit the description of this roadmap topic.

hasan4791:
#### What type of PR is this?
/kind api-change
/kind ci
/kind cleanup
/kind other

This PR is part-1 of container creation code refactoring where the mount related codes are moved to internal/factory/container/mounts file. 

Fixes #7275 

```release-note
None
```


hasan4791:
UTs are still in progress.
As per the current changes, the integrations & other build related test should pass without having issues.
Also this PR has lot of changes and would be great if it can be reviewed as it requires considerable effor.

hasan4791:
@dfr FYI as you're working on freebsd changes!!!

internal/factory/container/mounts.go
```diff
@@ -0,0 +1,285 @@
...
+	specAddMounts(c)
+
+	// Add mountlabel to the specgen
+	c.setMountLabel(containerInfo.MountLabel)
+
+	return containerVolumes, secretMounts, nil
+}
+
+// SpecAddPostOCIMounts add mounts after creating ocicontainer
+func (c *container) SpecAddPostOCIMounts(ctx context.Context, serverConfig *sconfig.Config, containerInfo storage.ContainerInfo, ociContainer *oci.Container, mountPoint string, timeZone string, rootPair idtools.IDPair) error {
```
haircommander, line 56: So this doesn't add mounts to the spec, it directly changes the rootfs. Maybe consider changing the name to reflect this?

Containerfile
```diff
@@ -0,0 +1,58 @@
+FROM fedora:latest
```
haircommander, line 1: what is  this included for?

Makefile
```diff
@@ -324,6 +324,13 @@ testunit-bin:
 			--gcflags '-N' -c -o ${TESTBIN_PATH}/$$(basename $$PACKAGE) ;\
 	done
 
+testunit-package:
+	mkdir -p ${TESTBIN_PATH}
+	go test ${PACKAGE} \
+		--tags "test $(BUILDTAGS)" \
+		--gcflags '-N' -c -o ${TESTBIN_PATH}/$$(basename ${PACKAGE})
+
```
haircommander, line 332: this can be acheieved by doing `TESTFLAGS='--focus="package' make testunit`

internal/factory/container/impl.go
```diff
@@ -0,0 +1,6 @@
+package container
+
+// Impl is the main implementation interface of this package.
+type Impl interface {
+	SecurityLabel(path, secLabel string, shared, maybeRelabel bool) error
+}
```
haircommander, line 6: I don't think this is needed if you just add a `label_unsupported.go` file that defines SecurityLabel as an empty function

internal/factory/container/mounts.go
```diff
@@ -0,0 +1,285 @@
...
+		criMounts: make(map[string]*rspec.Mount),
+		mounts:    make(map[string]*rspec.Mount),
+	}
+}
+
+func clearMountInfo(c *container) {
+	c.mountInfo.criMounts = nil
+	c.mountInfo.mounts = nil
+	c.mountInfo = nil
+}
```
haircommander, line 285: why not leave it for the duration of the container creation process? the container factory won't live that long

internal/factory/container/mounts.go
```diff
@@ -0,0 +1,285 @@
...
+	criSysSet bool
+	mounts    map[string]*rspec.Mount
+}
+
+type orderedMounts []*rspec.Mount
+
+// SpecAddPreOCIMounts add mounts to the spec before creating ocicontainer
+func (c *container) SpecAddPreOCIMounts(ctx context.Context, resourceStore *resourcestore.ResourceStore, serverConfig *sconfig.Config, sb *sandbox.Sandbox, containerInfo storage.ContainerInfo, mountPoint string, idMapSupport bool) ([]oci.ContainerVolume, []rspec.Mount, error) {
+	// Create temp mountInfo
+	c.mountInfo = newMountInfo()
```
haircommander, line 35: do we need to save mountInfo into the container if we're not referencing it after this function?

internal/factory/container/mounts.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// Clear temp mountInfo
+	defer clearMountInfo(c)
```
haircommander, line 38: if we clear mountinfo between these two calls (pre and post) then we lose the deduplication right?

Containerfile
hasan4791: Please ignore this file, currently im running integration tests in container locally and using this file.

PR #7865, opened 2024-03-11
Testing imagestore patch for KEP-4191

kannon92:
We are using this PR to verify if the image store patch fixes the upstream test failures we are seeing.

https://github.com/containers/storage/pull/1859 is open to push this into c/storage.


kwilczynski:
The `containers/storage` project has dedicated branches for that sort of things, I believe.

Also, why are we doing this? There is no description provided.

kannon92:
> The `containers/storage` project has dedicated branches for that sort of things, I believe.
> 
> Also, why are we doing this? There is no description provided.

Sorry. This was a POC to see if this fixes some e2e tests in kube. I’ll be moving this to a branch of c/storage today. They didn’t have one yet so it was easiest to put it here for testing.

PR #7889, opened 2024-03-14
Enable seccomp tests for arm64 arch

hasan4791:
#### What type of PR is this?
/kind ci

Enable seccomp tests for arm64 arch

```release-note
None
```

### DIFF
--- a/test/seccomp_notifier.bats
+++ b/test/seccomp_notifier.bats
@@ -8,10 +8,6 @@ function setup() {
 		skip "seccomp is not enabled"
 	fi
 
-	if [[ "$ARCH" != "$ARCH_X86_64" ]]; then
-		skip "not supported on arch $ARCH"
-	fi
-
 	setup_test
 }
 


saschagrunert (approved):
Thank you!

hasan4791:
@saschagrunert Weird, it does work here!!!
Btw the images we use for containers is available for both x86 & aarch64 right?

saschagrunert:
> @saschagrunert Weird, it does work here!!! Btw the images we use for containers in integration tests is available for both x86 & aarch64 right?

Yes, they should be multi-arch.

hasan4791:
> > @saschagrunert Weird, it does work here!!! Btw the images we use for containers in integration tests is available for both x86 & aarch64 right?
> 
> Yes, they should be multi-arch.

Yes, where can I find more about the runner we're using for arm46?

saschagrunert:
> > > @saschagrunert Weird, it does work here!!! Btw the images we use for containers in integration tests is available for both x86 & aarch64 right?
> > 
> > 
> > Yes, they should be multi-arch.
> 
> Yes, where can I find more details about the runner we're using for arm64?

Which information do you need? We're running on the default actuated runners.

saschagrunert:
This is done, thank you!

PR #7904, opened 2024-03-18
Save CRIU restore.log

adrianreber:
#### What type of PR is this?

/kind other

If there is an error restoring a container then there is usually an error message pointing to the log file created by CRIU. Unfortunately this log file is created in the bundle directory which is remove after container restore. The user gets told to look at a file which is automatically deleted.

In the case of an error the log file is now copied to a temporary directory:

  `os.CreateTemp("", fmt.Sprintf("restore-%s-*.log", ctr.ID()))`

The resulting error message is also adapted to point to this copy of the log file.

This change comes with a test that tries to trigger a restore error by creating a container with an established TCP connection and telling CRIU to handle established TCP connection while not telling CRIU to handle established TCP connection during restore. The restore will fail and the test can verify that the reported log file copy exists.

The main reason for this change is that one of most asked questions about checkpoint/restore in Kubernetes/CRI-O is that the log file is not readable.

None

```release-note
The CRIU log file (restore.log) is no longer deleted if the restore of the container fails. It will be copied to a temporary location.
```

### DIFF
--- a/internal/lib/restore.go
+++ b/internal/lib/restore.go
@@ -3,8 +3,10 @@ package lib
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	metadata "github.com/checkpoint-restore/checkpointctl/lib"
 	"github.com/checkpoint-restore/go-criu/v7/stats"
@@ -253,7 +255,32 @@ func (c *ContainerServer) ContainerRestore(
 		sb.CgroupParent(),
 		sb.MountLabel(),
 	); err != nil {
-		return "", fmt.Errorf("failed to restore container %s: %w", ctr.ID(), err)
+		// Save the CRIU log file as the original log file will be deleted pretty soon by cleanup code
+		// This is the only indication for a user why the checkpoint might have failed.
+		f, errLog := os.CreateTemp("", fmt.Sprintf("restore-%s-*.log", ctr.ID()))
+		if errLog != nil {
+			return "", fmt.Errorf("failed to restore container %s: %w", ctr.ID(), errLog)
+		}
+		defer f.Close()
+
+		src, errLog := os.Open(filepath.Join(ctr.BundlePath(), metadata.RestoreLogFile))
+		if errLog != nil {
+			return "", fmt.Errorf("failed to restore container %s: %w", ctr.ID(), errLog)
+		}
+		defer src.Close()
+
+		if _, errLog := io.Copy(f, src); errLog != nil {
+			return "", fmt.Errorf("failed to restore container %s: %w", ctr.ID(), errLog)
+		}
+
+		// Adapt the original error message to point to the new logfile
+		errorMessage := strings.ReplaceAll(
+			fmt.Sprintf("%s", err),
+			filepath.Join(ctr.BundlePath(), metadata.RestoreLogFile),
+			f.Name(),
+		)
+
+		return "", fmt.Errorf("failed to restore container %s: %s", ctr.ID(), errorMessage)
 	}
 	if err := c.ContainerStateToDisk(ctx, ctr); err != nil {
 		log.Warnf(ctx, "Unable to write containers %s state to disk: %v", ctr.ID(), err)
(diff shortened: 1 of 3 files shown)


adrianreber:
Hmm, I know why the unit tests are failing, but the `make vendor` change is unexpected.

adrianreber:
> @adrianreber, is there a way to save this file where CRI-O saves its own log file at the moment?
> 
> I actually don't know.

I am open to save the log file wherever it makes most sense.

haircommander:
do you think it would be useful to log this in crio instead? I feel like it may be tough for an administrator to find this file

adrianreber:
> do you think it would be useful to log this in crio instead? I feel like it may be tough for an administrator to find this file

Do you mean dumping the complete content of the file in the CRI-O log?

Currently the location of the log file is part of the error message the kubelet gives the user.

Being part of CRI-O would also work. It is, however, a quite large log file. So, it would be a lot of data.

I am open to suggestions. Just trying to preserve the information in the case of an error.

internal/lib/restore.go
```diff
@@ -253,7 +255,32 @@ func (c *ContainerServer) ContainerRestore(
 		sb.CgroupParent(),
 		sb.MountLabel(),
 	); err != nil {
-		return "", fmt.Errorf("failed to restore container %s: %w", ctr.ID(), err)
+		// Save the CRIU log file as the original log file will be deleted pretty soon by cleanup code
+		// This is the only indication for a user why the checkpoint might have failed.
+		f, err1 := os.CreateTemp("", fmt.Sprintf("restore-%s-*.log", ctr.ID()))
```
kwilczynski, line 260:
```suggestion
		f, errLog := os.CreateTemp("", fmt.Sprintf("restore-%s-*.log", ctr.ID()))
```

kwilczynski:
> > @adrianreber, is there a way to save this file where CRI-O saves its own log file at the moment?
> > I actually don't know.
> 
> I am open to save the log file wherever it makes most sense.

My idea was to save the file to the exact location of the default CRI-O log file. However, CRI-O would log to the standard error as default rather than to any given log file, making it harder to log into the exact location. Unless someone provides a file path via the `--log` command-line option.

I suppose the temporary directory would suffice.

internal/lib/restore.go
adrianreber: done

rst0git:
@adrianreber Would it make sense to use something similar to the [logCriuErrors](https://github.com/opencontainers/runc/pull/3816/commits/c77aaa3f95ec7cffbb24c53b0ee8a0b6d19d8839) functionality in runc?

adrianreber:
> @adrianreber Would it make sense to use something similar to the [logCriuErrors](https://github.com/opencontainers/runc/pull/3816/commits/c77aaa3f95ec7cffbb24c53b0ee8a0b6d19d8839) functionality in runc?

This change will be part of runc 1.2.0 and will probably lead to a more useful error message runc passes to CRI-O and then to Kubernetes. For a complete problem analysis we still need the complete log file which will be deleted by CRI-O the way it is currently setup. The runc change will be helpful but not always enough.

internal/lib/restore.go
```diff
@@ -3,8 +3,10 @@ package lib
 import (
```
rst0git, line 3: @adrianreber There are a few typos in the commit message. Would it be possible to fix them?

test/checkpoint.bats
```diff
@@ -143,3 +143,52 @@ function teardown() {
 	[[ "$container_name" == "restored-sleep-container" ]]
 	[[ "$pod_name" == "restoresandbox2" ]]
 }
+
+@test "checkpoint and try to restore one container with an open TCP connection and expected failure" {
+	# This test needs to fail during restore to see if the log file is saved at a location
+	# the user can access it.
+	# This test is starting a container and creates an established TCP connections.
```
rst0git, line 150: A simpler way to simulate an error could be by adding unrecognised option to the CRIU configuration file just before restore.

rst0git (approved):

adrianreber:
Can someone re-open this?

PR #7905, opened 2024-03-18
add selinux package to crio

kannon92:
#### What type of PR is this?
/kind cleanup

SELinux package from podman is mostly for displaying some KVM labels.

Seems easy enough just to move this to crio to drop that dependency.

```release-note
NONE
```


server/container_create_linux.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/containers/common/pkg/timezone"
 	"github.com/containers/common/pkg/util"
 	"github.com/containers/podman/v4/pkg/rootless"
-	selinux "github.com/containers/podman/v4/pkg/selinux"
```
haircommander, line 18: I don't think we need a whole new package, we can include htis in server

saschagrunert (approved):

PR #7937, opened 2024-03-26
oci: keep track of exec PIDs and stop them on container stop

haircommander:
#### What type of PR is this?

/kind bug

```release-note
Keep track of exec calls for a container, and make sure to kill them when a container is being stopped
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -74,6 +74,7 @@ type Container struct {
 	restoreStorageImageID *storage.StorageImageID
 	resources             *types.ContainerResources
 	runtimePath           string // runtime path for a given platform
+	execPIDs              map[int]bool
 }
 
 func (c *Container) CRIAttributes() *types.ContainerAttributes {
@@ -159,6 +160,7 @@ func NewContainer(id, name, bundlePath, logPath string, labels, crioAnnotations,
 		stopSignal:      stopSignal,
 		stopTimeoutChan: make(chan int64, 10),
 		stopWatchers:    []chan struct{}{},
+		execPIDs:        map[int]bool{},
 	}
 	return c, nil
 }
@@ -760,4 +762,55 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
 		return r.handler.RuntimePath
 	}
 	return c.runtimePath
+}
+
+// AddExecPID registers a PID associated with an exec session.
+// It is tracked so exec sessions can be cancelled when the container is being stopped.
+// If the PID is conmon, shouldKill should be false, as we should not call SIGKILL on conmon.
+// If it is an exec session, shouldKill should be true, as we can't guarantee the exec process
+// will have a SIGINT handler.
+func (c *Container) AddExecPID(pid int, shouldKill bool) error {
+	c.stopLock.Lock()
+	defer c.stopLock.Unlock()
+	if c.stopping {
+		return errors.New("cannot register an exec PID: container is stopping")
+	}
+	c.execPIDs[pid] = shouldKill
+	return nil
+}
+
+// DeleteExecPID is for deregistering a pid after it has exited.
+func (c *Container) DeleteExecPID(pid int) {
+	c.stopLock.Lock()
+	defer c.stopLock.Unlock()
+	delete(c.execPIDs, pid)
+}
+
+// KillExecPIDs loops through the saved execPIDs and sends a signal to them.
+// If shouldKill is true, the signal is SIGKILL. Otherwise, SIGINT.
+func (c *Container) KillExecPIDs() {
+	c.stopLock.Lock()
+	toKill := c.execPIDs
+	c.stopLock.Unlock()
+
+	for len(toKill) != 0 {
+		unkilled := map[int]bool{}
+		for pid, shouldKill := range toKill {
+			if pid == 0 {
+				// The caller may accidentally register `0` (for instance if the PID of the cmd has already exited)
+				// and killing 0 is the way to ask the kernel to kill the whole process group of the calling process.
+				// We definitely don't want to kill the CRI-O process group, so add this check just in case.
+				continue
+			}
+			sig := syscall.SIGINT
+			if shouldKill {
+				sig = syscall.SIGKILL
+			}
+			if err := syscall.Kill(pid, sig); err != nil && !errors.Is(err, syscall.ESRCH) {
+				unkilled[pid] = shouldKill
+			}
+		}
+		toKill = unkilled
+		time.Sleep(stopProcessWatchSleep)
+	}
 }
--- a/internal/oci/oci_unix.go
+++ b/internal/oci/oci_unix.go
@@ -29,15 +29,21 @@ func setSize(fd uintptr, size remotecommand.TerminalSize) error {
 	return unix.IoctlSetWinsize(int(fd), unix.TIOCSWINSZ, winsize)
 }
 
-func ttyCmd(execCmd *exec.Cmd, stdin io.Reader, stdout io.WriteCloser, resizeChan <-chan remotecommand.TerminalSize) error {
+func ttyCmd(execCmd *exec.Cmd, stdin io.Reader, stdout io.WriteCloser, resizeChan <-chan remotecommand.TerminalSize, c *Container) error {
 	p, err := pty.Start(execCmd)
 	if err != nil {
 		return err
 	}
 	defer p.Close()
-
 	// make sure to close the stdout stream
 	defer stdout.Close()
+
+	pid := execCmd.Process.Pid
+	if err := c.AddExecPID(pid, true); err != nil {
+		return err
+	}
+
+	defer c.DeleteExecPID(pid)
 
 	utils.HandleResizing(resizeChan, func(size remotecommand.TerminalSize) {
 		if err := setSize(p.Fd(), size); err != nil {
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -537,6 +537,21 @@ function check_oci_annotation() {
 	[[ $(crictl exec --sync "$ctr_id" /bin/sh -c "for i in $(seq 1 50000000); do echo -n 'a'; done" | wc -c) -le 16777216 ]]
 }
 
+@test "ctr exec{,sync} should be cancelled when container is stopped" {
+	start_crio
+	ctr_id=$(crictl run "$TESTDATA"/container_sleep.json "$TESTDATA"/sandbox_config.json)
+
+	crictl exec --sync "$ctr_id" /bin/bash -c 'while true; do echo XXXXXXXXXXXXXXXXXXXXXXXX; done' &
+	pid1=$!
+	crictl exec "$ctr_id" /bin/bash -c 'while true; do echo XXXXXXXXXXXXXXXXXXXXXXXX; done' || true &
+	pid2=$!
+
+	sleep 1s
+
+	crictl stop "$ctr_id"
+	wait "$pid1" "$pid2"
+}
+
 @test "ctr device add" {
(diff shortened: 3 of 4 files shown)


internal/oci/container.go
```diff
@@ -761,3 +763,50 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
...
+
+// KillExecPIDs loops through the saved execPIDs and sends a signal to them.
+// If shouldKill is true, the signal is SIGKILL. Otherwise, SIGINT.
+func (c *Container) KillExecPIDs() {
+	c.stopLock.Lock()
+	toKill := c.execPIDs
+	unkilled := c.execPIDs
+	c.stopLock.Unlock()
+
+	for len(unkilled) == 0 {
```
kolyshkin, line 797: 1. Did you mean `!= 0` here?
2. It's cleaner to use `len(toKill)` here and move  `unkilled` instantiation to inside the loop.

haircommander: two great points!

kolyshkin (commented):
Hmm, I was thinking cri-o maintains a list of running execs per container somewhere, at least those with a terminals, as it has to read from those terminals (doing that in a separate goroutine I guess). Can we reuse that list?

(and yes, we don't have to kill all execs, just those with a terminal, although that's not important)

haircommander:
> Hmm, I was thinking cri-o maintains a list of running execs per container somewhere, at least those with a terminals, as it has to read from those terminals (doing that in a separate goroutine I guess). Can we reuse that list?
> 
> (and yes, we don't have to kill all execs, just those with a terminal, although that's not important)

yeah right it's in a separate goroutine, and even the context diverges between the different CRI calls, so wiring them  together would be tricky.

internal/oci/container.go
```diff
@@ -761,3 +763,49 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
...
+
+	for len(toKill) != 0 {
+		unkilled := map[int]bool{}
+		for pid, shouldKill := range toKill {
+			sig := syscall.SIGINT
+			if shouldKill {
+				sig = syscall.SIGKILL
+			}
+			if err := syscall.Kill(pid, sig); err != nil && !errors.Is(err, syscall.ESRCH) {
+				unkilled[pid] = shouldKill
```
kolyshkin, line 804: So, you can either kill a process or not, and it doesn't make sense to retry. From kill(2) man page:
```
ERRORS
       EINVAL An invalid signal was specified.

       EPERM  The calling process does not have permission to send the signal to any of the target processes.

       ESRCH  The target process or process group does not exist.  Note that an existing process might be a zombie, a process that has terminated execution, but has not yet been wait(2)ed for.
```

Also, kill is async, meaning it can take some time for the kernel to actually release the process (and, in case of SIGINT, it also depends on what the app is doing).

There are two ways to make kill synchronous, if you need that here. Alas neither way is perfect.
1. Keep killing pid(s) till you get ESRCH (with some sensible timeout). The downside is there's a chance of PID wrap issue, meaning you'll be killing a wrong process. Practically, if the time the killing code is running is small enough (seconds), this issue is highly unlikely to happen.
2. Use pidfd and wait on events. Unfortunately this requires a new kernel and some amount of non-trivial code.

The current code is **kind of** implementing the first approach, but
- it does not look like it (as you keep sending the same signal);
- the number of iterations (and thus the total time spend killing) need to be limited.

So I suggest to rewrite this as following:
 - kill all execs (no loop needed)
 - check that they are killed (sending signal 0 in a loop with limited number of tries)
 - (optionally) use KILL for those PIDs that you sent INT to before, as a last resort. This can be done somewhere in the middle of the loop.

kolyshkin: * return (or log) an error if some execs still can't be killed.

kwilczynski: Something that I ran through @haircommander recently.

We could also leverage setting the `Setpgid` via `syscall.SysProcAttr{}` to first deliver a signal to the given process group. We do this today when creating the main container process but not using an Exec-type approach.

There is also `unix.PR_SET_PDEATHSIG` we could set to use `unix.SIGKILL` using `unix.Prctl()` better clean up children, etc.

If we want to tackle the PID wrap issue, then to avoid having to track everything internally, we could inject an environment variable such as `_OCI_COOKIE` or simply `CRIO` with some value (e.g., a random SHA256 or UUID generated at the CRI-O startup) and then look for this when we are about to kill a given PID.

haircommander: > The downside is there's a chance of PID wrap issue, meaning you'll be killing a wrong process. Practically, if the time the killing code is running is small enough (seconds), this issue is highly unlikely to happen.

yeah I don't think we need to worry about PID wrap in this case

internal/oci/container.go
```diff
@@ -761,3 +763,54 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
 	}
 	return c.runtimePath
 }
+
+// AddExecPID registers a PID associated with an exec session.
+// It is tracked so exec sessions can be cancelled when the container is being stopped.
+// If the PID is conmon, shouldKill should be false, as we should not call SIGKILL on conmon
```
kolyshkin, line 769: Nit: missing period

test/ctr.bats
```diff
@@ -537,6 +537,19 @@ function check_oci_annotation() {
...
+	start_crio
+	ctr_id=$(crictl run "$TESTDATA"/container_sleep.json "$TESTDATA"/sandbox_config.json)
+
+	crictl exec --sync "$ctr_id" /bin/bash -c 'while true; do echo XXXXXXXXXXXXXXXXXXXXXXXX; done' &
+	crictl exec "$ctr_id" /bin/bash -c 'while true; do echo XXXXXXXXXXXXXXXXXXXXXXXX; done' || true &
+
+	sleep 1s
+
+	crictl stop "$ctr_id"
+	fg || true
```
kolyshkin, line 550: I don't think job control is enabled inside bats. What you need here is wait:

```bash
	crictl exec --sync "$ctr_id" /bin/bash -c 'while true; do echo XXXXXXXXXXXXXXXXXXXXXXXX; done' &
	pid1=$!
	crictl exec "$ctr_id" /bin/bash -c 'while true; do echo XXXXXXXXXXXXXXXXXXXXXXXX; done' &
	pid2=$!

....

	wait $pid1 $pid2
```

haircommander: updated, thanks for the suggestion!

haircommander:
@cri-o/cri-o-maintainers this is ready, PTAL

internal/oci/container.go
```diff
@@ -761,3 +763,54 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
...
+				sig = syscall.SIGKILL
+				// In the case we're killing a container, we should kill the process group
+				// pid *= -1
+			}
+			if err := syscall.Kill(pid, sig); err != nil && !errors.Is(err, syscall.ESRCH) {
+				unkilled[pid] = shouldKill
+			}
+		}
+		toKill = unkilled
+		time.Sleep(stopProcessWatchSleep)
```
rphillips, line 814: c.execPIDS is not updated to be empty after this function runs... We should update it, right?

haircommander: so the stopping code is not reentrant, and we won't add any additional execPIDs after we begin the stopping process. functionally, after this runs the variable is no longer useful and can be ignored. I could update it if you want, but it shouldn't change behavior

rphillips: ok. Doesn't need to be updated.

internal/oci/container.go
```diff
@@ -761,3 +763,54 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
...
+func (c *Container) KillExecPIDs() {
+	c.stopLock.Lock()
+	toKill := c.execPIDs
+	c.stopLock.Unlock()
+
+	for len(toKill) != 0 {
+		unkilled := map[int]bool{}
+		for pid, shouldKill := range toKill {
+			if pid == 0 {
+				// Just in case
```
kwilczynski, line 800: Just in case of what? You want to prevent the following from having place:

>       If pid equals 0, then sig is sent to every process in the process
>       group of the calling process.

internal/oci/container.go
```diff
@@ -761,3 +763,54 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
...
+				continue
+			}
+			sig := syscall.SIGINT
+			if shouldKill {
```
kwilczynski, line 804: We could save the multiple passes using a single loop over a slice. First, send SIGINT, then sleep, and then follow up with the SIGKILL. This is what it boils down in the end, isn't it?

internal/oci/container.go
kwilczynski: The quote is from [kill(2)](https://www.man7.org/linux/man-pages/man2/kill.2.html).

kwilczynski:
/hold

To allow for review comments to be scrutinised.

internal/oci/container.go
```diff
@@ -761,3 +763,54 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
...
+				sig = syscall.SIGKILL
+				// In the case we're killing a container, we should kill the process group
+				// pid *= -1
```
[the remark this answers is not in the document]
haircommander: whoops good catch

internal/oci/container.go
haircommander: no we should never SIGKILL conmon, the `shouldKill` bool is basically `isConmon`. I hear you on the loop. We could just go through once if you'd prefer, but I am worried the sigint may be lost... WDYT?

internal/oci/container.go
kwilczynski: No worries. What I was after was a better comment that mentions this potential issue. So that whoever reads the code is aware of why we do this check.

internal/oci/container.go
kwilczynski: No issue with the SIGINT vs SIGKILl for conmon. My comment was more to avoid having to loop more than once. I also wanted to avoid using a map in Go, as these things are just utterly broken at the moment, per:

- https://github.com/golang/go/issues/20135

This bothers me greatly every time I see a map in Go nowadays.

haircommander:
/unhold

updated, thanks @kwilczynski PTAL

internal/oci/oci_unix.go
```diff
@@ -29,16 +29,22 @@ func setSize(fd uintptr, size remotecommand.TerminalSize) error {
...
 	defer p.Close()
-
 	// make sure to close the stdout stream
 	defer stdout.Close()
 
+	pid := execCmd.Process.Pid
+	if err := c.AddExecPID(pid, true); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
kwilczynski: This is just for consistency with the other such code block.

internal/oci/container.go
haircommander: > My comment was more to avoid having to loop more than once.

if the processes exit, then we won't loop more :) A concern could be busy looping though...

> I also wanted to avoid using a map in Go, as these things are just utterly broken at the moment, per:

I don't think we need to worry as much about this, as the map will only survive for the lifecycle of the container. Though, if you'd prefer, we could track the pids in separate slices or something, though that would make it more expensive to delete them...

kwilczynski: @haircommander, sounds good!

kwilczynski:
@haircommander, anything against cherry-pick to release 1.28 to 1.25? I will be happy to do it manually, if needed.

haircommander:
nothing at all, sounds good!

kwilczynski:
OK. Needs a manual cherry-pick. No worries.

PR #7941, opened 2024-03-27
Add cri-o-runc source to install.md

tvorogme:
#### What type of PR is this?

I've installed Debian 11 and copy&run command from `install.md`, however it wasn't working due `E: Package 'cri-o-runc' has no installation candidate`, one of repo is forgotten

To install cri-o WITH crio-o-runc, otherwise it'll not start

General installation of cri-o on debian

Try to use those commands without my fix :)

None

### DIFF
--- a/install.md
+++ b/install.md
@@ -572,6 +572,7 @@ and export it as a variable, like so:
 ```shell
 rm /etc/apt/sources.list.d/devel:kubic:libcontainers:stable:cri-o:$VERSION.list
 
+echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/ /"| sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -572,6 +572,7 @@ and export it as a variable, like so:
 ```shell
 rm /etc/apt/sources.list.d/devel:kubic:libcontainers:stable:cri-o:$VERSION.list
 
+echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/ /"| sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
```
saschagrunert, line 575: Hey @tvorogme, can you provide a reproducer for this kind of issue? Usually the kubic repositories should not affected by changes in https://github.com/cri-o/packaging, so I'm wondering what we're missing here.

tvorogme: You can start debian 11 container and run without change with OS Debian_11, on any version. It will not install cri-o-runc from apt (no such package). I add this echo from other tutorials and do `apt install cri-o-runc` successfully

saschagrunert: @haircommander do we need `crio-o-runc` for the kubic repos as well? I'm confused.

haircommander: only for the legacy obs packages. I think we can have this but we are moving away from this packaging scheme so we'll soon no longer use cri-o-runc

PR #7945, opened 2024-03-29
oci: AttachContainer: always read attach socket

xdavidwu:
conmon uses blocking writes on attach socket, if we do not read at all, conmon will be stuck.

#### What type of PR is this?

/kind bug

On an attach request with both stdout and stderr disabled, when there are output, since we never read from conmon attach socket in this case, conmon will be stuck at writing, and as a result, stdio of the container will be frozen.

It can be reproduced with a container that runs `cat` and an attach request with large enough data (above 1MB seems to be enough) for stdin, and disabled stdout, stderr.

None

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1185,12 +1185,10 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
 	defer conn.Close()
 
 	receiveStdout := make(chan error)
-	if outputStream != nil || errorStream != nil {
-		go func() {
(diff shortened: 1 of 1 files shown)


haircommander:
this makes sense to me. Thanks for opening a PR @xdavidwu !
/ok-to-test
/approve

@cri-o/cri-o-maintainers PTAL

kwilczynski:
@haircommander, should we backport this to older releases? Thoughts?

haircommander:
yeah let's start with 1.30 and maybe 1.29. Let's not backport quickly in case we can catch any odd regressions
/cherry-pick release-1.30

PR #7951, opened 2024-04-02
use wasmedge and drop crun-wasm

kannon92:
#### What type of PR is this?
/kind cleanup

use wasm-edge and drop crum-wasm from fedora package

```release-note
NONE
```


sohankunkerkar (commented):
/lgtm

PR #7953, opened 2024-04-02
add x to set for traceability

kannon92:
#### What type of PR is this?
/kind cleanup

add -x to bash allows one to trace the commands.

```release-note
NONE
```


kannon92:
@haircommander I am confused on why this would be causing a failure in the CI. But I rerun these tests a few times and it is failing..

PR #7828, opened 2024-02-29
Disable Recursive Read-only (RRO) mounts feature

saschagrunert:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

The feature can be enabled when we implemented it in CRI-O.

#### Which issue(s) this PR fixes:

Follow-up on https://github.com/cri-o/cri-o/pull/7765#discussion_r1506892815
Refers to https://github.com/kubernetes-sigs/cri-tools/pull/1344#issuecomment-1970245805

#### Special notes for your reviewer:

cc @giuseppe @AkihiroSuda 

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #7962, opened 2024-04-04
OCPNODE-2213: Add support for Recursive Read-only (RRO) mounts

kwilczynski:
#### What type of PR is this?

/kind feature

Currently, the read-only option set for the host mounts does not correctly propagate recursively from the top level of the given mount point. As such, any sub-mounts that happen to be part of the tree would retain their current options, and should it be a read/write type mount, then it could be modified without any restriction. This is often both unexpected and undesirable.

Thus, add support for so-called Recursive Read-only (RRO) mounts. This allows a mount point to propagate the read-only option through the tree, marking the entire mount point read-only.

Support for the RRO mounts has already been available in the runtimes, such as **crun** and **runc**, for some time now.

> [!NOTE] 
> Support of the Recursive Read-only (RRO) mounts requires:
>
> - For Linux, kernel version 5.12
> - Kubernetes v1.30🔹
> - cri-tools v1.30
> - crun v1.4
> - runc v1.1
>
> 🔹⚠️ Feature gate called **RecursiveReadOnlyMounts** needs to be enabled given the _Alpha_ status of the feature.

Related to:

- https://github.com/kubernetes/enhancements/issues/3857
- https://github.com/kubernetes/enhancements/pull/3858
- https://github.com/kubernetes/kubernetes/pull/123272
- https://github.com/kubernetes-sigs/cri-tools/pull/1344
- https://github.com/opencontainers/runc/pull/3272
- https://github.com/containers/crun/pull/778

A temporary workaround:

- https://github.com/cri-o/cri-o/pull/7828

None

```release-note
Add support for the Recursive Read-only (RRO) mounts following runtimes and kernel support to enable read-only mounts to be recursively read-only.
```

### DIFF
--- a/contrib/test/ci/vars.yml
+++ b/contrib/test/ci/vars.yml
@@ -130,6 +130,10 @@ kata_skip_ctr_tests:
   - 'test "ctr with default_env set in configuration"'
   - 'test "ctr with absent mount that should be rejected"'
   - 'test "ctr stop loop kill retry attempts"'
+  - 'test "ctr that mounts container storage as read-only option but not recursively"'
+  - 'test "ctr that mounts container storage as recursively read-only"'
+  - 'test "ctr that fails to mount container storage as recursively read-only without readonly option"'
+  - 'test "ctr that fails to mount container storage as recursively read-only without private propagation"'
 kata_skip_devices_tests:
   - 'test "additional devices permissions"'
   - 'test "annotation devices support"'
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -190,6 +190,12 @@ func (c *RootConfig) GetStore() (storage.Store, error) {
 	})
 }
 
+// runtimeHandlerFeatures represents the supported features of the runtime.
+type runtimeHandlerFeatures struct {
+	RecursiveReadOnlyMounts bool `json:"-"` // Internal use only.
+	features.Features
+}
+
 // RuntimeHandler represents each item of the "crio.runtime.runtimes" TOML
 // config table.
 type RuntimeHandler struct {
@@ -243,7 +249,7 @@ type RuntimeHandler struct {
 
 	// Output of the "features" subcommand.
 	// This is populated dynamically and not read from config.
-	features features.Features
+	features runtimeHandlerFeatures
 }
 
 // Multiple runtime Handlers in a map
@@ -1258,7 +1264,7 @@ func (c *RuntimeConfig) ValidateRuntimes() error {
 				return err
 			}
 
-			logrus.Warnf("'%s is being ignored due to: %q", name, err)
+			logrus.Warnf("Runtime handler %q is being ignored due to: %v", name, err)
 			failedValidation = append(failedValidation, name)
 		}
 	}
@@ -1275,12 +1281,12 @@ func (c *RuntimeConfig) initializeRuntimeFeatures() {
 	for name, handler := range c.Runtimes {
 		versionOutput, err := cmdrunner.CombinedOutput(handler.RuntimePath, "--version")
 		if err != nil {
-			logrus.Errorf("Unable to determine version of runtime handler %s: %v", name, err)
+			logrus.Errorf("Unable to determine version of runtime handler %q: %v", name, err)
 			continue
 		}
 
 		versionString := strings.ReplaceAll(strings.TrimSpace(string(versionOutput)), "\n", ", ")
-		logrus.Infof("Using runtime %s", versionString)
+		logrus.Infof("Using runtime handler %s", versionString)
 
 		memoryBytes, err := handler.SetContainerMinMemory()
 		if err != nil {
@@ -1298,10 +1304,33 @@ func (c *RuntimeConfig) initializeRuntimeFeatures() {
 			logrus.Errorf("Getting %s OCI runtime features failed: %s: %v", handler.RuntimePath, output, err)
 			continue
 		}
-		// Ignore errors to Unmarshal too, we can't populate it.
-		if err := json.Unmarshal(output, &handler.features); err != nil {
-			logrus.Errorf("Unmarshalling OCI features failed: %s", err)
+
+		// Ignore error if we can't load runtime features.
+		if err := handler.LoadRuntimeFeatures(output); err != nil {
+			logrus.Errorf("Unable to load OCI features for runtime handler %q: %v", name, err)
+			continue
 		}
+
+		if handler.RuntimeSupportsIDMap() {
+			logrus.Debugf("Runtime handler %q supports User and Group ID-mappings", name)
+		}
+
+		// Recursive Read-only (RRO) mounts require runtime handler support,
+		// such as runc v1.1 or crun v1.4. For Linux, the minimum kernel
+		// version 5.12 or a kernel with the necessary changes backported
+		// is required.
+		rro := handler.RuntimeSupportsMountFlag("rro")
+		if rro {
+			logrus.Debugf("Runtime handler %q supports Recursive Read-only (RRO) mounts", name)
+
+			// A given runtime might support Recursive Read-only (RRO) mounts,
+			// but the current kernel might not.
+			if err := checkKernelRROMountSupport(); err != nil {
+				logrus.Warnf("Runtime handler %q supports Recursive Read-only (RRO) mounts, but kernel does not: %v", name, err)
+				rro = false
+			}
+		}
+		handler.features.RecursiveReadOnlyMounts = rro
 	}
 }
 
@@ -1621,6 +1650,29 @@ func (r *RuntimeHandler) SetContainerMinMemory() (int64, error) {
 	return memoryBytes, nil
 }
 
+// LoadRuntimeFeatures loads features for a given runtime handler using the "features"
+// sub-command output, where said output contains a JSON document called "Features
+// Structure" that describes the runtime handler's supported features.
+func (r *RuntimeHandler) LoadRuntimeFeatures(input []byte) error {
+	if err := json.Unmarshal(input, &r.features); err != nil {
+		return fmt.Errorf("unable to unmarshal features structure: %w", err)
+	}
+
+	// All other properties of the Features Structure are optional and might be
+	// either absent, empty, or set to the null value, with the exception of
+	// OCIVersionMin and OCIVersionMax, which are required. Thus, the lack of
+	// them should indicate that the Features Structure document is potentially
+	// not valid.
+	//
+	// See the following for more details about the Features Structure:
+	//   https://github.com/opencontainers/runtime-spec/blob/main/features.md
+	if r.features.OCIVersionMin == "" || r.features.OCIVersionMax == "" {
+		return errors.New("runtime features structure is not valid")
+	}
+
+	return nil
+}
+
 // RuntimeSupportsIDMap returns whether this runtime supports the "runtime features"
 // command, and that the output of that command advertises IDMap mounts as an option
 func (r *RuntimeHandler) RuntimeSupportsIDMap() bool {
@@ -1631,6 +1683,11 @@ func (r *RuntimeHandler) RuntimeSupportsIDMap() bool {
 		return false
 	}
 	return true
+}
+
+// RuntimeSupportsRROMounts returns whether this runtime supports the Recursive Read-only mount as an option.
+func (r *RuntimeHandler) RuntimeSupportsRROMounts() bool {
+	return r.features.RecursiveReadOnlyMounts
 }
 
 // RuntimeSupportsMountFlag returns whether this runtime supports the specified mount option.
--- a/pkg/config/config_linux.go
+++ b/pkg/config/config_linux.go
@@ -1,6 +1,16 @@
 package config
 
-import selinux "github.com/opencontainers/selinux/go-selinux"
+import (
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/containers/storage/pkg/parsers/kernel"
+	selinux "github.com/opencontainers/selinux/go-selinux"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
 
 // Defaults if none are specified
 const (
@@ -14,6 +24,11 @@ const (
 	DefaultPauseImage string = "registry.k8s.io/pause:3.9"
 )
 
+var (
+	kernelRROSupportOnce  sync.Once
+	kernelRROSupportError error
+)
+
 func selinuxEnabled() bool {
 	return selinux.GetEnabled()
 }
@@ -23,4 +38,106 @@ func (c *RuntimeConfig) ValidatePinnsPath(executable string) error {
 	c.PinnsPath, err = validateExecutablePath(executable, c.PinnsPath)
 
 	return err
+}
+
+// checkKernelRROMountSupport checks the kernel support for the Recursive Read-only (RRO) mounts.
+func checkKernelRROMountSupport() error {
+	kernelRROSupportOnce.Do(func() {
+		kernelRROSupportError = (func() error {
+			// Check the current kernel version for RRO mounts support...
+			err := validateKernelRROVersion()
+			if err != nil {
+				versionErr := err
+				// ... and if the kernel version does not match the minimum required version,
+				// then verify whether the kernel supports RRO mounts regardless, as often
+				// Linux distributions provide heavily patched kernel releases, and the
+				// current kernel might include backported support.
+				err = validateKernelRROMount()
+				if err != nil {
+					err = fmt.Errorf("%w: %w", versionErr, err)
+				}
+			}
+			return err
+		})()
+	})
+
+	return kernelRROSupportError
+}
+
+// validateKernelRROVersion checks whether the current kernel version matches the release 5.12 or newer,
+// which is the minimum required kernel version that supports Recursive Read-only (RRO) mounts.
+func validateKernelRROVersion() error {
+	kv, err := kernel.GetKernelVersion()
+	if err != nil {
+		return fmt.Errorf("unable to retrieve kernel version: %w", err)
+	}
+
+	result := kernel.CompareKernelVersion(*kv,
+		kernel.VersionInfo{
+			Kernel: 5,
+			Major:  12,
+			Minor:  0,
+		},
+	)
+	if result < 0 {
+		return fmt.Errorf("kernel version %q does not support recursive read-only mounts", kv)
+	}
+
+	return nil
+}
+
+// validateKernelRROMount checks whether the current kernel can support Recursive Read-only mounts.
+// It uses a test mount of tmpfs against which an attempt will be made to set the required attributes.
+// If there is no failure in doing so, then the kernel has the required support.
+func validateKernelRROMount() error {
+	path, err := os.MkdirTemp("", "crio-rro-*")
+	if err != nil {
+		return fmt.Errorf("unable to create directory: %w", err)
+	}
+	defer func() {
+		if err := os.RemoveAll(path); err != nil {
+			logrus.Errorf("Unable to remove directory: %v", err)
+		}
+	}()
+
+	for {
+		err = unix.Mount("", path, "tmpfs", 0, "")
+		if !errors.Is(err, unix.EINTR) {
+			break
+		}
+	}
+	if err != nil {
+		return fmt.Errorf("unable to mount directory %q using tmpfs: %w", path, err)
+	}
+	defer func() {
+		var unmountErr error
+		for {
+			unmountErr = unix.Unmount(path, 0)
+			if !errors.Is(unmountErr, unix.EINTR) {
+				break
+			}
+		}
+		if unmountErr != nil {
+			logrus.Errorf("Unable to unmount directory %q: %v", path, unmountErr)
+		}
+	}()
+
+	for {
+		err = unix.MountSetattr(-1, path, unix.AT_RECURSIVE,
+			&unix.MountAttr{
+				Attr_set: unix.MOUNT_ATTR_RDONLY,
+			},
+		)
+		if !errors.Is(err, unix.EINTR) {
+			break
+		}
+	}
+	if err != nil {
+		if !errors.Is(err, unix.ENOSYS) {
+			return fmt.Errorf("unable to set mount attribute for directory %q: %w", path, err)
+		}
+		return fmt.Errorf("unable to set recursive read-only mount attribute: %w", err)
+	}
+
+	return nil
 }
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -333,7 +333,8 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 
 	s.resourceStore.SetStageForResource(ctx, ctr.Name(), "container volume configuration")
 	idMapSupport := s.Runtime().RuntimeSupportsIDMap(sb.RuntimeHandler())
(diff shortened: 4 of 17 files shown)


server/server_linux.go
```diff
@@ -139,3 +141,47 @@ func configureMaxThreads() error {
...
+		}
+
+		index := bytes.IndexByte(uname.Release[:], 0)
+		version := string(uname.Release[:index])
+
+		if _, err := fmt.Sscanf(version, "%d.%d", &major, &minor); err != nil {
+			return fmt.Errorf("unable to parse kernel version %q: %w", version, err)
+		}
+
+		hasMajor := major > kernelVersion/100
```
haircommander, line 168: I am fine with this for now though eventually it would be better if we found a way to check the feature without referencing the version, as it could have been backported.

haircommander: ah nevermind I see the TODO below

server/server_linux.go
```diff
@@ -139,3 +141,47 @@ func configureMaxThreads() error {
 	logrus.Debugf("Golang's threads limit set to %d", maxThreads)
 	return nil
 }
+
+// checkKernelRROMountSupport checks the kernel support for the Recursive Read-only (RRO) mounts.
+func checkKernelRROMountSupport() error {
+	// The minimal required kernel version is 5.12 or newer.
+	const kernelVersion = 512
+
+	checkKernelRROSupport := func() error {
```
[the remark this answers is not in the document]
haircommander: BTW I think I'd prefer this inline below, rather than saving the functor and calling it below. I don't think we save on indentation or code readability as-is

server/runtime_status.go
```diff
@@ -48,9 +48,14 @@ func (s *Server) Status(ctx context.Context, req *types.StatusRequest) (*types.S
 			}
 		}
 
-		// TODO: enable when CRI-O implemented Recursive Read-only (RRO) mounts
-		// rro := runtime.RuntimeSupportsMountFlag("rro")
 		rro := false
+		if runtime.RuntimeSupportsMountFlag("rro") {
+			// A given runtime might support Recursive Read-only (RRO) mounts,
+			// but the underlying kernel might not.
+			if err := checkKernelRROMountSupport(); err == nil {
```
haircommander, line 55: we call this a lot, potentially unnecessarily. I think we should check this in `initializeRuntimeFeatures()` in `pkg/config/config.go` and if the kernel doesn't support the flag, then remove it from the features structure. Then, from that point on, we can just ask the runtime whether it's supported or not

server/server_linux.go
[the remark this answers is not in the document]
kwilczynski: Sure! Will do.

server/container_create_linux.go
```diff
@@ -1119,6 +1136,35 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
...
+		} else {
+			if m.RecursiveReadOnly {
+				return nil, nil, fmt.Errorf(
+					"recursive read-only mount conflicts with read-write mount for hostPath %q",
+					m.HostPath,
+				)
+			}
+			mountOption = "rw"
+		}
+		options = append(options, mountOption)
```
haircommander, line 1166: we can do this with a flat if/else:
```
if rro && ro {
// rro
} else if rro
// err
} else if ro {
// ro
} else {
// rw
}
```
which I think would be easier to read

server/server_linux.go
haircommander: yeah makes sense to me

server/container_create_linux.go
kwilczynski: Sure. I can do it like that.

For the record, I personally find nested if-else-if blocks harder to read.

haircommander: yeah It would be best as a switch but we have two variables to check :(

server/runtime_status.go
kwilczynski: I see what you mean. I think we can plumb this like that. Let me see. :+1:

server/container_create_linux.go
kwilczynski: Done.

server/server_linux.go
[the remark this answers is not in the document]
kwilczynski: Done.

server/runtime_status.go
kwilczynski: We unmarshal the JSON atop the following:

_(from [opencontainers/runtime-spec/specs-go/features/features.go](https://github.com/opencontainers/runtime-spec/blob/main/specs-go/features/features.go#L5-L33)):_

```go
type Features struct {
	// OCIVersionMin is the minimum OCI Runtime Spec version recognized by the runtime, e.g., "1.0.0".
	OCIVersionMin string `json:"ociVersionMin,omitempty"`

	// OCIVersionMax is the maximum OCI Runtime Spec version recognized by the runtime, e.g., "1.0.2-dev".
	OCIVersionMax string `json:"ociVersionMax,omitempty"`

	// Hooks is the list of the recognized hook names, e.g., "createRuntime".
	// Nil value means "unknown", not "no support for any hook".
	Hooks []string `json:"hooks,omitempty"`

	// MountOptions is the list of the recognized mount options, e.g., "ro".
	// Nil value means "unknown", not "no support for any mount option".
	// This list does not contain filesystem-specific options passed to mount(2) syscall as (const void *).
	MountOptions []string `json:"mountOptions,omitempty"`

	// Linux is specific to Linux.
	Linux *Linux `json:"linux,omitempty"`

	// Annotations contains implementation-specific annotation strings,
	// such as the implementation version, and third-party extensions.
	Annotations map[string]string `json:"annotations,omitempty"`

	// PotentiallyUnsafeConfigAnnotations the list of the potential unsafe annotations
	// that may appear in `config.json`.
	//
	// A value that ends with "." is interpreted as a prefix of annotations.
	PotentiallyUnsafeConfigAnnotations []string `json:"potentiallyUnsafeConfigAnnotations,omitempty"`
}
```

This is not the same as the CRI API type, which I think you had in mind here?

server/server_linux.go
kwilczynski: Done.

server/container_create_linux.go
kwilczynski: I am going to rewrite this using a switch statement, since linter picks up on it, per:

```golang
server/container_create_linux.go:1119:3: ifElseChain: rewrite if-else to switch statement (gocritic)
		if m.Readonly && m.RecursiveReadOnly {
```

server/container_create_linux.go
```diff
@@ -992,7 +993,7 @@ func clearReadOnly(m *rspec.Mount) {
 	m.Options = append(m.Options, "rw")
 }
 
-func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel, bindMountPrefix string, absentMountSourcesToReject []string, maybeRelabel, skipRelabel, cgroup2RW, idMapSupport bool, storageRoot string) ([]oci.ContainerVolume, []rspec.Mount, error) {
+func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel, bindMountPrefix string, absentMountSourcesToReject []string, maybeRelabel, skipRelabel, cgroup2RW, idMapSupport, rroSupport bool, storageRoot string) ([]oci.ContainerVolume, []rspec.Mount, error) {
```
kwilczynski, line 996: **To the reviewer:**

We need to refactor this at some point. The number of arguments is slowly becoming unwieldy.

server/container_create_linux.go
```diff
@@ -1119,6 +1116,36 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
...
+		// set to private.
+		switch {
+		case m.Readonly && m.RecursiveReadOnly:
+			if !rroSupport {
+				return nil, nil, fmt.Errorf(
+					"recursive read-only mount support is not available for hostPath %q",
+					m.HostPath,
+				)
+			}
+			if m.Propagation != types.MountPropagation_PROPAGATION_PRIVATE {
```
kwilczynski, line 1130: **To the reviewer:**

When deployed atop Kubernetes, we have the advantage of having the API request validation to tell us that required attributes are not set or have wrong values. However, when creating containers using the API, for example, using crictl, we don't have this added value, and as such, we perform this validation ourselves.

server/container_restore.go
```diff
@@ -331,6 +333,14 @@ func (s *Server) CRImportCheckpoint(
...
 		}
 
+		// Support for the Recursive Read-only (RRO) mount requires
+		// the mount to be read-only and the mount propagation to be
+		// set to private.
+		if mount.RecursiveReadOnly {
+			mount.Readonly = true
+			mount.Propagation = types.MountPropagation_PROPAGATION_PRIVATE
+		}
+
```
kwilczynski, line 343: **To the reviewer:**

Hopefully, this is the right thing to do when restoring a container.

pkg/config/config.go
```diff
@@ -1277,6 +1289,27 @@ func (c *RuntimeConfig) initializeRuntimeFeatures() {
...
+			logrus.Debugf("Runtime handler %q supports User and Group ID-mappings", name)
+		}
+
+		// Recursive Read-only (RRO) mounts require runtime handler support,
+		// such as runc v1.1 or crun v1.4. For Linux, the minimal kernel
+		// version 5.12 or a kernel with the necessary changes backported
+		// are required.
+		rro := handler.RuntimeSupportsMountFlag("rro")
+		if rro {
+			logrus.Debugf("Runtime handler %q supports Recursive Read-only (RRO) mounts", name)
```
kwilczynski, line 1303: **To the reviewer:**

Perhaps this should be an `logrus.Infof()`?

kwilczynski: If so, we should add as nice a neat runtime handler features print on startup.

pkg/config/config.go
```diff
@@ -1277,6 +1289,27 @@ func (c *RuntimeConfig) initializeRuntimeFeatures() {
...
+
+			// A given runtime might support Recursive Read-only (RRO) mounts,
+			// but the underlying kernel might not.
+			if err := checkKernelRROMountSupport(); err != nil {
+				logrus.Warnf("Runtime handler %q supports Recursive Read-only (RRO) mounts, but kernel does not: %v", name, err)
```
kwilczynski, line 1308: **To the reviewer:**

An unsupported platform emits:

```console
WARN[2024-04-09 10:25:56.665685980Z] Runtime handler "runc" supports Recursive Read-only (RRO) mounts, but kernel does not: kernel version "5.10.214" does not support recursive read-only mounts: unable to set recursive read-only mount attribute: function not implemented 
WARN[2024-04-09 10:25:56.671210757Z] Runtime handler "crun" supports Recursive Read-only (RRO) mounts, but kernel does not: kernel version "5.10.214" does not support recursive read-only mounts: unable to set recursive read-only mount attribute: function not implemented 
```

pkg/config/config.go
```diff
@@ -1266,18 +1278,41 @@ func (c *RuntimeConfig) ValidateRuntimes() error {
...
 			continue
 		}
-		// Ignore errors to Unmarshal too, we can't populate it.
-		if err := json.Unmarshal(output, &handler.features); err != nil {
-			logrus.Errorf("Unmarshalling OCI features failed: %s", err)
+
+		// Ignore error if we can't load runtime features.
+		if err := handler.LoadRuntimeFeatures(output); err != nil {
+			logrus.Errorf("Unable to load OCI features for runtime handler %q: %v", name, err)
+			continue
```
kwilczynski, line 1293: **To the reviewer:**

This was originally missing. I think it was a small bug.

server/container_create_linux.go
haircommander: very true. @hasan4791 is working on a refactor which could handle this

server/container_create_linux.go
haircommander: I kinda feel we should do this validation in crictl to match kube doing it but I am open to double validating

server/container_create_linux.go
```diff
@@ -1119,6 +1116,35 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
...
+				return nil, nil, fmt.Errorf(
+					"recursive read-only mount requires private propagation for hostPath %q, got: %s",
+					m.HostPath, m.Propagation,
+				)
+			}
+			options = append(options, "rro")
+		case m.Readonly:
+			options = append(options, "ro")
+		default:
+			if m.RecursiveReadOnly {
```
haircommander, line 1139: why not another case here?

server/container_restore.go
```diff
@@ -331,6 +333,13 @@ func (s *Server) CRImportCheckpoint(
 			}
 		}
 
+		// Recursive Read-only (RRO) support requires the mount to be
+		// read-only and the mount propagation set to private.
+		if mount.RecursiveReadOnly {
+			mount.Readonly = true
+			mount.Propagation = types.MountPropagation_PROPAGATION_PRIVATE
+		}
+
```
kwilczynski, line 342: **To the reviewer:**

Hopefully, this is the right thing to do when restoring a container.

server/container_create_linux.go
kwilczynski: It depends on how you look at it. If you take CRI-O as an API server, then providing server-side validation, so to speak, would be more than fine, like the 412 HTTP status code in RESTful endpoints.

server/container_create_linux.go
kwilczynski: Sure.

server/container_create_linux.go
kwilczynski: Thank you, @hasan4791, in advance! :tada:

server/container_create_linux.go
kwilczynski: Done.

haircommander:
/approve

LGTM, thanks! @cri-o/cri-o-maintainers PTAL

pkg/config/config_linux.go
```diff
@@ -24,3 +33,115 @@ func (c *RuntimeConfig) ValidatePinnsPath(executable string) error {
...
+			err := validateKernelRROVersion()
+			if err != nil {
+				versionErr := err
+				// ... and if the kernel version does not match the minimum required version,
+				// then verify whether the kernel supports RRO mounts regardless, as often
+				// Linux distributions provide heavily patched kernel releases, and the
+				// current kernel might include backported support.
+				err = validateKernelRROMount()
+				if err != nil {
+					return fmt.Errorf("%v: %w", versionErr, err)
```
saschagrunert, line 51: nit: we can use `errors.Join` here.

kwilczynski:
/hold

For a brief moment.

pkg/config/config_linux.go
kwilczynski: The problem with `errors.Join()` is this:

```
// The error formats as the concatenation of the strings obtained
// by calling the Error method of each element of errs, with a newline
// between each string.
```

This sadly breaks logrus per:

```console
WARN[2024-04-11 08:03:38.207720457Z] Runtime handler "runc" supports Recursive Read-only (RRO) mounts, but kernel does not: kernel version "5.10.214" does not support recursive read-only mounts
unable to set recursive read-only mount attribute: function not implemented
WARN[2024-04-11 08:03:38.215434604Z] Runtime handler "crun" supports Recursive Read-only (RRO) mounts, but kernel does not: kernel version "5.10.214" does not support recursive read-only mounts
unable to set recursive read-only mount attribute: function not implemented
```

This is an unfortunate API that Go has that does not allow you to customise the `Error()` method of the type that handles joining errors.

I can replace `%v` with `%w`, as from 1.20 onwards, `fmt.Errorf()` supports multiple error wrapping. This will have the same effect, sans the better error unwrapping if needed.

test/ctr.bats
```diff
@@ -996,6 +996,136 @@ function check_oci_annotation() {
...
+
+	start_crio
+
+	ctr_id=$(crictl run "$TESTDIR"/config "$TESTDATA"/sandbox_config.json)
+
+	run ! --separate-stderr crictl exec --sync "$ctr_id" touch /host/test
+	[[ "$stderr" == *"Read-only file system"* ]]
+
+	crictl exec --sync "$ctr_id" touch /host/test-sub-volume/test
+}
```
kwilczynski, line 1032: @littlejawa, do you happen to know off the cuff, perhaps, whether any read-only mounts under Kata are also recursively read-only for the entire tree? So, would a sub-volume mount also be read-only?

I am asking, as this simple above test does not pass under Kata, whereas it has no issues on a vanilla tux container.

kwilczynski: Probably not. I am getting the same failure on Fedora.

contrib/test/ci/vars.yml
```diff
@@ -130,6 +130,10 @@ kata_skip_ctr_tests:
   - 'test "ctr with default_env set in configuration"'
   - 'test "ctr with absent mount that should be rejected"'
   - 'test "ctr stop loop kill retry attempts"'
+  - 'test "ctr that mounts container storage as read-only option but not recursively"'
+  - 'test "ctr that mounts container storage as recursively read-only"'
+  - 'test "ctr that fails to mount container storage as recursively read-only without readonly option"'
+  - 'test "ctr that fails to mount container storage as recursively read-only without private propagation"'
```
kwilczynski, line 136: **To the reviewer:**

I am not sure if Kata supports this correctly. I couldn't get it to work. Perhaps it's the [virtio-fs](https://virtio-fs.gitlab.io/) that's behind the scenes that requires some setting to be correctly enabled.

haircommander:
/approve

if there's another round of reviews, refactoring the commits to break up the test changes `supports_selinux` pieces and the addition of the code would be nice, but nothing to block on IMO

@cri-o/cri-o-maintainers PTAL

kwilczynski:
> if there's another round of reviews, refactoring the commits to break up the test changes `supports_selinux` pieces and the addition of the code would be nice, but nothing to block on IMO

@haircommander, done.

pkg/config/config.go
kolyshkin: If we're ignoring the error here, should it be `logrus.Warnf` not `Errorf`?

pkg/config/config.go
```diff
@@ -64,6 +65,11 @@ const (
 	MonitorExecCgroupContainer = "container"
 )
 
+var (
+	kernelRROSupportOnce  sync.Once
+	kernelRROSupportError error
+)
+
```
kolyshkin, line 72: AFAICS this needs to be moved to config_linux.go, as it's only used from there.

pkg/config/config_linux.go
```diff
@@ -24,3 +33,115 @@ func (c *RuntimeConfig) ValidatePinnsPath(executable string) error {
...
+
+	supportedMajor := major > kernelVersion/100
+	supportedMinor := major == kernelVersion/100 && minor >= kernelVersion%100
+
+	if !supportedMajor && !supportedMinor {
+		return fmt.Errorf("kernel version %q does not support recursive read-only mounts", version)
+	}
+
+	return nil
+}
```
kolyshkin, line 91: Maybe it's easier to use github.com/containers/storage/pkg/parsers/kernel rather than reimplement uname parsing here?

pkg/config/config_linux.go
```diff
@@ -24,3 +33,115 @@ func (c *RuntimeConfig) ValidatePinnsPath(executable string) error {
...
+// It uses a test mount of tmpfs against which an attempt will be made to set the required attributes.
+// If there is no failure in doing so, then the kernel has the required support.
+func validateKernelRROMount() error {
+	path, err := os.MkdirTemp("", "crio-rro-*")
+	if err != nil {
+		return fmt.Errorf("unable to create directory: %w", err)
+	}
+	defer func() {
+		if err := os.RemoveAll(path); err != nil {
+			logrus.Errorf("Unable to remove directory %q: %v", path, err)
```
kolyshkin, line 103: nit: os.RemoveAll error is like "rmdir %s: %w" (meaning you don't have to include the directory name and the fact that directory removal failed, as it's already in the error).

pkg/config/config.go
kwilczynski: The idea was to share this amongst other implementations. However, I don't know if we will ever get FreeBSD one to work with. I will move it, indeed.

pkg/config/config_linux.go
kwilczynski: I had looked at this package before; it was quite elaborate, given what we needed here. However, I will switch to it to benefit from a much more widely used and better-tested implementation. 👍

pkg/config/config_linux.go
kwilczynski: Ah, yes. The `PathError` type captures it. Will change.

pkg/config/config.go
kwilczynski: Done.

pkg/config/config_linux.go
kwilczynski: Done.

pkg/config/config_linux.go
kwilczynski: Done.

pkg/config/config.go
kwilczynski: I can change these to warnings. However, an error carries much more weight than a warning when we want to bring something to the user's attention, especially during initialisation. Perhaps that is why this was originally done that way.

contrib/test/ci/vars.yml
littlejawa: I can't answer from the top of my head. It's probably better to keep them in the skip list for now.
Thanks!

kwilczynski:
@cri-o/cri-o-maintainers, please have a look again. Thank you!

pkg/config/config.go
kwilczynski: The output format remains the same even after following the refactor as suggested in the review comment:

- https://github.com/cri-o/cri-o/pull/7962#pullrequestreview-2002056079

sohankunkerkar (approved):
/lgtm

test/ctr.bats
```diff
@@ -996,6 +996,125 @@ function check_oci_annotation() {
 	crictl exec --sync "$ctr_id" findmnt -no TARGET,PROPAGATION "$CTR_DIR" | grep -v private
 }
 
+@test "ctr that mounts container storage as read-only option but not recursively" {
+	# See https://www.shellcheck.net/wiki/SC2154 for more details.
+	declare stderr
+
+	# Parent of "--root", keep in sync with test/helpers.bash file.
+	TEST_VOLUME="$TESTDIR"/test-volume
```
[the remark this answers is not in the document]
kwilczynski: @sohankunkerkar, we could.

However, this specific set of `tmpfs` mount points is only ever going to be used in these two tests. I am unsure if abstracting it out adds the benefits of reusability here. We need it to be done specifically this way, and it's not applicable outside these two tests, in other words.

Thoughts?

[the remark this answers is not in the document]
sohankunkerkar: Yeah, I agree with you. I was thinking if we could do it in the same test, but feel free to ignore this suggestion if it doesn't make sense.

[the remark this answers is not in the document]
kwilczynski: No worries. This is now done.

kwilczynski:
OK. Requires manual cherry-pick.

saschagrunert:
> OK. Requires manual cherry-pick.

No need to cherry-pick this one as discussed.

kwilczynski:
> > OK. Requires manual cherry-pick.
> 
> No need to cherry-pick this one as discussed.

To add more background here:

I initially thought we would need to backport this to 1.29, to support older Kubernetes and, as such, OpenShift. Thinking that we were missing some crucial functionality; however, this feature, as I remembered later, is only made available as Alpha in Kubernetes 1.30. Thus, there is no point in having any backports.

PR #7965, opened 2024-04-05
add release versions for manual workflow run 

### LINKED ISSUE #4003: LFX 2024: Fully automated releases

olad5:
#### What type of PR is this?

/kind feature

* Add the env arg for the version in patch-release script to manually run the workflow 
Fixes #4003

* add the env for the version to make release from
* remove unused release.go code
None

```release-note

```

### DIFF
--- a/scripts/patch-release/patch-release.go
+++ b/scripts/patch-release/patch-release.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/blang/semver/v4"
@@ -23,6 +24,8 @@ const (
 	versionFile       = "internal/version/version.go"
 	branchPrefix      = "release-"
 )
+
+var releaseMinorVersions = []string{"1.29", "1.28", "1.27"}
 
 func main() {
 	logrus.SetFormatter(&logrus.TextFormatter{DisableTimestamp: true})
@@ -55,29 +58,59 @@ func run() error {
 		return fmt.Errorf("open local repo: %w", err)
 	}
 
-	currentBranch, err := repo.CurrentBranch()
-	if err != nil {
-		return fmt.Errorf("get current branch: %w", err)
-	}
-	logrus.Infof("Using current branch: %s", currentBranch)
+	for _, minorVersion := range releaseMinorVersions {
+		baseBranchName := branchPrefix + minorVersion // returns "release-x.y"
 
-	newVersion, err := incVersion(version.Version)
-	if err != nil {
-		return fmt.Errorf("increment version: %w", err)
-	}
+		sv, err := getCurrentVersionFromReleaseBranch(repo, baseBranchName) // returns "x.y.z"
+		if err != nil {
+			return fmt.Errorf("current version from release branch %s to semver: %w", minorVersion, err)
+		}
 
-	logrus.Infof("Using new version: %s", util.SemverToTagString(newVersion))
+		// Bump up the patch version
+		sv.Patch++
 
-	if err := updateVersionAndCreatePR(
-		repo, newVersion, currentBranch, org, remote,
-	); err != nil {
-		return fmt.Errorf("update version in local repository: %w", err)
+		if err := updateVersionAndCreatePR(
+			repo, sv, baseBranchName, org, remote,
+		); err != nil {
+			return fmt.Errorf("update version in local repository: %w", err)
+		}
 	}
 
 	return nil
 }
 
-func incVersion(tag string) (res semver.Version, err error) {
+func getCurrentVersionFromReleaseBranch(repo *git.Repo, baseBranchName string) (res semver.Version, err error) {
+	logrus.Infof("Switching to branch: %s", baseBranchName)
+	if err := repo.Checkout(baseBranchName); err != nil {
+		return res, fmt.Errorf("checkout branch %s: %w", baseBranchName, err)
+	}
+
+	versionFromVersionFile, err := getCurrentVersionFromVersionFile(versionFile) // returns "x.xx.x"
+	if err != nil {
+		return res, fmt.Errorf("reading latest version: %w", err)
+	}
+
+	return convertStringToSemver(versionFromVersionFile)
+}
+
+func getCurrentVersionFromVersionFile(versionFile string) (string, error) {
+	versionPattern := `const\s+Version\s+=\s+"(.+)"`
+
+	content, err := os.ReadFile(versionFile)
+	if err != nil {
+		return "", err
+	}
+
+	re := regexp.MustCompile(versionPattern)
+	matches := re.FindStringSubmatch(string(content))
+	if len(matches) < 2 {
+		return "", fmt.Errorf("unable to find version in %q", versionFile)
+	}
+
+	return matches[1], nil
+}
+
+func convertStringToSemver(tag string) (res semver.Version, err error) {
 	sv, err := util.TagStringToSemver(strings.TrimSpace(tag))
 	if err != nil {
 		return res, fmt.Errorf("convert tag string %s to semver: %w", tag, err)
@@ -85,15 +118,11 @@ func incVersion(tag string) (res semver.Version, err error) {
 
 	// clear any suffix like `-dev`
 	sv.Pre = nil
-
-	// New patch version
-	sv.Patch++
-
 	return sv, nil
 }
 
 func updateVersionAndCreatePR(
-	repo *git.Repo, newVersion semver.Version, branch, org, remote string,
+	repo *git.Repo, newVersion semver.Version, baseBranchName, org, remote string,
 ) error {
 	logrus.Info("Updating repository")
 
@@ -152,7 +181,7 @@ func updateVersionAndCreatePR(
 		newVersion, "/release-note-none",
 	)
 
-	pr, err := gh.CreatePullRequest("cri-o", "cri-o", branch,
+	pr, err := gh.CreatePullRequest("cri-o", "cri-o", baseBranchName,
 		headBranchName,
 		title,
 		body,
--- a/scripts/patch-release/patch-release_test.go
+++ b/scripts/patch-release/patch-release_test.go
@@ -11,6 +11,33 @@ import (
 
 var _ = t.Describe("Automated Releases", func() {
 	t.Describe("Patch Release", func() {
+		It("should read the version file", func() {
+			tempFileName := "tempVersionFile"
+			tmpfile, err := os.CreateTemp("", tempFileName)
+			if err != nil {
+				Expect(err).ToNot(HaveOccurred())
+			}
+			defer os.Remove(tempFileName)
+			currentVersion := "1.30.0"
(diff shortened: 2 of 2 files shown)


scripts/release/release.go
```diff
@@ -1,153 +0,0 @@
-package main
```
haircommander, line 1: If this is being removed anyway, I'm curious why we didn't write the patch-release code over this to preserve git history

saschagrunert (changes requested):
Can we also modify the GitHub action to be able to utilize the arg?

saschagrunert:
/ok-to-test

The linter is not happy: https://github.com/cri-o/cri-o/actions/runs/8617701845/job/23618280557?pr=7965

scripts/patch-release/patch-release.go
```diff
@@ -24,6 +24,8 @@ const (
 	branchPrefix      = "release-"
 )
 
+var releaseVersions = []string{"1.29.4", "1.28.4", "1.27.4"}
```
saschagrunert, line 27: This should be just the release branches:

```suggestion
var releaseBranches = []string{"release-1.29", "release-1.28", "release-1.27"}
```

Or you just use something like `releaseMinorVersions` which is then `1.29`, `1.28` and you automatically preprend `release-`.

saschagrunert (changes requested):

scripts/patch-release/patch-release.go
```diff
@@ -55,45 +57,47 @@ func run() error {
 		return fmt.Errorf("open local repo: %w", err)
 	}
 
-	currentBranch, err := repo.CurrentBranch()
-	if err != nil {
-		return fmt.Errorf("get current branch: %w", err)
+	currentReleaseBranch := branchPrefix + version.Version
+	for _, version := range releaseVersions {
```
saschagrunert, line 61: You need to iterate over the release branches and retrieve the right `version` for them.

scripts/patch-release/patch-release.go
olad5: I think i would go with the releaseMinorVersions and prepend the "release-"

scripts/patch-release/patch-release.go
olad5: I don't understand  what  you mean by "right version".
I thought we would have 	releaseMinorVersions = []string{"1.29", "1.28", "1.27"}
then loop through that slice, prepend the release prefix and create the PR.

olad5: I still don't get it. Currently now, we have 1.30.0  in the internal/version/version.go
but we have 1.29 as the first element of the slice, how does that work?

or do you mean We should take the elements of the slices, using 1.29 as an example, 
then bump up the patch version so it becomes 1.29.1, then make the release PR
"release-1.29.1"

saschagrunert: > or do you mean We should take the elements of the slices, using 1.29 as an example,
then bump up the patch version so it becomes 1.29.1, then make the release PR
"release-1.29.1"

Exactly. You have to retrieve `1.29.1` by incrementing the latest release on the `release-1.29` branch. That's the tricky part here.

scripts/release/release.go
haircommander: /hold

still curious about this

scripts/patch-release/patch-release.go
```diff
@@ -55,45 +57,37 @@ func run() error {
...
-		return fmt.Errorf("increment version: %w", err)
-	}
-
-	logrus.Infof("Using new version: %s", util.SemverToTagString(newVersion))
-
-	if err := updateVersionAndCreatePR(
-		repo, newVersion, currentBranch, org, remote,
-	); err != nil {
-		return fmt.Errorf("update version in local repository: %w", err)
+	currentReleaseBranch := branchPrefix + version.Version
```
saschagrunert, line 60: `currentReleaseBranch` is the base for the PR's we create. This value is now the same for all `releaseMinorVersions`, which is wrong. The base for a PR to update `v1.29.x` should be `release-1.29` and so on.

scripts/patch-release/patch-release.go
```diff
@@ -55,45 +57,37 @@ func run() error {
...
+	for _, version := range releaseMinorVersions {
+		sv, err := convertStringToSemver(version)
```
saschagrunert, line 62: The value of `sv` is now just `1.29`, `1.28` or `1.27`, which is not correct. We need the current value of `version.Version` for each release branch and increment that.

saschagrunert (changes requested):

scripts/release/release.go
olad5: Hi @haircommander , apologies for the late reply.
We decided to focus on a patch releases for now, just to see how the workflow will work and we agreed that the file is not in use and so we should remove it.

scripts/patch-release/patch-release.go
olad5: Added the changes as requested.

scripts/patch-release/patch-release.go
olad5: Added the changes as requested.

scripts/patch-release/patch-release.go
```diff
@@ -55,45 +57,37 @@ func run() error {
...
-	); err != nil {
-		return fmt.Errorf("update version in local repository: %w", err)
+	for _, version := range releaseMinorVersions {
+		baseBranchName := branchPrefix + version
+		sv, err := convertStringToSemver(version + ".0") // returns "1.27.0"
+		if err != nil {
+			return fmt.Errorf("convert tag string %s to semver: %w", version, err)
+		}
+		// Bump up the patch version
+		sv.Patch++
```
saschagrunert, line 67: What if the latest version for `release-1.27` is `v1.27.3`? Then we need to propose `v1.27.4` and not `v1.27.0`.

olad5: okay. Currently, I do not know where i can read that value from. I mean the latest version (v1.27.3).

olad5: Or can I read it from the tags?

saschagrunert: You checkout the release branch, read the file `internal/version/version.go` using `os.ReadFile` and parse the version from that:

https://github.com/cri-o/cri-o/blob/c8553396972c019ae02963913cafed608884bd8b/internal/version/version.go#L24

scripts/release/release.go
haircommander: @saschagrunert @kwilczynski are you opposed to making the changes in place instead of creating a new file to preserve git history?

saschagrunert: Hm, no we can replace the file if you want.

kwilczynski: We should keep the old file for now. We can remove it once things are ready. This gives a clear commit then.

olad5: reverted the file as requested. @haircommander

scripts/patch-release/patch-release.go
```diff
@@ -55,45 +58,71 @@ func run() error {
 		return fmt.Errorf("open local repo: %w", err)
 	}
 
-	currentBranch, err := repo.CurrentBranch()
-	if err != nil {
-		return fmt.Errorf("get current branch: %w", err)
+	for _, minorVersion := range releaseMinorVersions {
+		baseBranchName := branchPrefix + minorVersion // returns "release-x.xx"
+
+		sv, err := getCurrentVersionFromReleaseBranch(repo, baseBranchName) // returns "x.xx.x"
```
saschagrunert, line 64: technically it should be:

```suggestion
		sv, err := getCurrentVersionFromReleaseBranch(repo, baseBranchName) // returns "x.y.z"
```

scripts/patch-release/patch-release.go
```diff
@@ -55,45 +58,71 @@ func run() error {
 		return fmt.Errorf("open local repo: %w", err)
 	}
 
-	currentBranch, err := repo.CurrentBranch()
-	if err != nil {
-		return fmt.Errorf("get current branch: %w", err)
+	for _, minorVersion := range releaseMinorVersions {
+		baseBranchName := branchPrefix + minorVersion // returns "release-x.xx"
```
saschagrunert, line 62: technically it should be
```suggestion
		baseBranchName := branchPrefix + minorVersion // returns "release-x.y"
```

Makefile
```diff
@@ -196,9 +196,6 @@ build-static:
 crio.conf: bin/crio
 	./bin/crio -d "" --config="" $(CONF_OVERRIDES) config > crio.conf
 
-release:
-	${GO_RUN} ./scripts/release
-
```
saschagrunert, line 201: This needs to be reverted as well.

Makefile
```diff
@@ -551,7 +548,6 @@ metrics-exporter: bin/metrics-exporter
 	upload-artifacts \
 	bin/metrics-exporter \
 	metrics-exporter \
-	release \
```
saschagrunert, line 554: This needs to be reverted as well.

saschagrunert (commented):
Just a few nits, otherwise LGTM

scripts/patch-release/patch-release.go
olad5: so I do that for this as well?

baseBranchName := branchPrefix + minorVersion // returns "release-x.xx"

saschagrunert (approved):

PR #7975, opened 2024-04-08
Do not abort startup if CRIU binary not found

adrianreber:
#### What type of PR is this?

/kind bug

Now that checkpoint/restore is turned on by default, CRI-O no longer starts up if the CRIU binary is missing. This changes the error if the CRIU binary is not found to an information message. Checkpoint/Restore support is turned of but CRI-O still starts.

None

I would have preferred a way to detect if the option has been explicitly set by the user, but that seems hard to detect (if possible at all). If the user explicitly requests `--enable-criu-support=true` then it would make sense to fail if the CRIU binary is not found. If the user is just using the default setting then it  should not be fatal if the CRIU binary is missing.

In Kubernetes I have seen something like this in combination with `*int` and if the value is `!=nil` something was selected by the user and if it is `nil` the default value is used. I have not seen something like this in CRI-O.

```release-note
None
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1160,11 +1160,12 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 		if c.EnableCriuSupport {
 			if err := validateCriuInPath(); err != nil {
 				c.EnableCriuSupport = false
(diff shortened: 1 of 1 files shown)


adrianreber:
CC: @saschagrunert

saschagrunert (approved):

PR #7958, opened 2024-04-03
test: use kata 3.6.0 for testing

littlejawa:
#### What type of PR is this?

/kind ci

#### What this PR does / why we need it:

Update the version of kata to test with - using the latest 3.6.0 version. 
Also make kata test specific changes (affecting only the fedora-kata job).

Details:
- move to kata 3.6.0
- enable hooks tests, as the fix for it was released with 3.5.0
- make sure the storage.overlay.skip_mount_home flag is set to true (problem introduced with kata 3.3.0)
- enable debug logs for kata
- cleanup lingering kata processes on teardown
- add an integration test specific to kata, testing we're actually using the kata runtime. Useful to validate config/script changes in the CI

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #7976, opened 2024-04-08
Reload config update pinned_images

### LINKED ISSUE #7954: crio v1.29.2 config reload does not update pinned_images
### What happened?

```
$ crio --version
crio version 1.29.2-9.rhaos4.16.git5052e1a.el9
Version:        1.29.2-9.rhaos4.16.git5052e1a.el9
```
If I update pinned_images  config

```
[crio]
  [crio.image]
    pinned_images = ["quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:86d26e7ebcccd6f07a75db5b1e56283b25c2ee1c6a755d6ffc5a4d59beb9c504"]
```
reload service
```
$ systemctl reload crio.service
```
config appearss to reload correct after catching HUP signal.
```
$ journalctl -u crio | grep "pinned images"
Apr 02 17:44:09 ip-10-0-14-61 crio[1096516]: time="2024-04-02 17:44:09.569932578Z" level=info msg="Updated new pinned images: [quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:86d26e7ebcccd6f07a75db5b1e56283b25c2ee1c6a755d6ffc5a4d59beb9c504]
```

client reports the image as unpinned
```
crictl image --pinned quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:86d26e7ebcccd6f07a75db5b1e56283b25c2ee1c6a755d6ffc5a4d59beb9c504
IMAGE                                            TAG                 IMAGE ID            SIZE                PINNED
quay.io/openshift-release-dev/ocp-v4.0-art-dev   <none>              7948809a7f475       672MB               false
```
if I restart the service it shows pinned

reload of config would update pinned_images

this was tested on nightly 4.16 payload

_No response_

```
$ crio --version
crio version 1.29.2-9.rhaos4.16.git5052e1a.el9
Version:        1.29.2-9.rhaos4.16.git5052e1a.el9
```

kubelet --version
Kubernetes v1.29.2+258f1d5

### OS version

```
Linux ip-10-0-14-61 5.14.0-432.el9.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar 20 20:26:59 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```
```
# cat /etc/os-release
NAME="CentOS Stream CoreOS"
ID="rhcos"
ID_LIKE="rhel fedora"
VERSION="416.94.202403300731-0"
VERSION_ID="4.16"
VARIANT="CoreOS"
VARIANT_ID=coreos
PLATFORM_ID="platform:el9"
PRETTY_NAME="CentOS Stream CoreOS 416.94.202403300731-0"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:9::coreos"
HOME_URL="https://centos.org/"
DOCUMENTATION_URL="https://docs.okd.io/latest/welcome/index.html"
REDHAT_BUGZILLA_PRODUCT_VERSION="4.16"
REDHAT_SUPPORT_PRODUCT_VERSION="4.16"
OPENSHIFT_VERSION="4.16"
RHEL_VERSION=9.4
OSTREE_VERSION="416.94.202403300731-0"
```

aws OCP 4.16 nightly

roman-kiselenko:
#### What type of PR is this?

/kind bug
/kind feature

Updates pinned images list on config reload

Fixes #7954 

None

```release-note

Updates pinned images list on config reload

```

/cc @haircommander @sohankunkerkar @hexfusion @kwilczynski

### DIFF
--- a/contrib/test/ci/vars.yml
+++ b/contrib/test/ci/vars.yml
@@ -98,6 +98,10 @@ kata_skip_cgroups_tests:
 kata_skip_command_tests:
   - 'test "crio commands"'
   - 'test "log max boundary testing"'
+kata_skip_reload_config_tests:
+  - "test \"reload config should update 'pinned_images'\""
+  - "test \"reload config should update 'pinned_images' and only 'pause_image' is pinned\""
+  - "test \"reload config should update 'pause_image' and it becomes 'pinned_images'\""
 kata_skip_config_tests:
   - 'test "choose different default runtime should succeed"'
   - 'test "runc not existing when default_runtime changed should succeed"'
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -152,6 +152,9 @@ type ImageServer interface {
 	// CandidatesForPotentiallyShortImageName resolves an image name into a set of fully-qualified image names (domain/repo/image:tag|@digest).
 	// It will only return an empty slice if err != nil.
 	CandidatesForPotentiallyShortImageName(systemContext *types.SystemContext, imageName string) ([]RegistryImageReference, error)
+
+	// UpdatePinnedImagesList updates pinned and pause images list in imageService.
+	UpdatePinnedImagesList(imageList []string)
 }
 
 func parseImageNames(image *storage.Image) (someName *RegistryImageReference, tags []reference.NamedTagged, digests []reference.Canonical, err error) {
@@ -896,6 +899,11 @@ type nativeStorageTransport struct{}
 
 func (st nativeStorageTransport) ResolveReference(ref types.ImageReference) (types.ImageReference, *storage.Image, error) {
 	return istorage.ResolveReference(ref)
+}
+
+// UpdatePinnedImagesList updates pinned images list in imageService.
+func (svc *imageService) UpdatePinnedImagesList(pinnedImages []string) {
+	svc.regexForPinnedImages = CompileRegexpsForPinnedImages(pinnedImages)
 }
 
 // FilterPinnedImage checks if the given image needs to be pinned
--- a/pkg/config/reload.go
+++ b/pkg/config/reload.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/cri-o/cri-o/internal/log"
@@ -20,7 +21,7 @@ func (c *Config) Reload() error {
 	// Reload the config
 	newConfig, err := DefaultConfig()
 	if err != nil {
-		return errors.New("unable to create default config")
+		return fmt.Errorf("unable to create default config: %w", err)
 	}
 
 	if _, err := os.Stat(c.singleConfigPath); !os.IsNotExist(err) {
@@ -145,17 +146,23 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
 }
 
 // ReloadPinnedImages updates the PinnedImages with the provided `newConfig`.
+// The method print log in case of any updates.
 func (c *Config) ReloadPinnedImages(newConfig *Config) {
 	updatedPinnedImages := make([]string, len(newConfig.PinnedImages))
+	updatedPinnedImageList := false
+
 	for i, image := range newConfig.PinnedImages {
 		if i < len(c.PinnedImages) && image == c.PinnedImages[i] {
-			updatedPinnedImages[i] = c.PinnedImages[i]
-		} else {
-			updatedPinnedImages[i] = image
+			continue
 		}
+		updatedPinnedImages[i] = image
+		updatedPinnedImageList = true
 	}
-	logrus.Infof("Updated new pinned images: %+v", updatedPinnedImages)
-	c.PinnedImages = updatedPinnedImages
+
+	if updatedPinnedImageList {
+		logConfig("pinned_images", strings.Join(updatedPinnedImages, ", "))
+		c.PinnedImages = updatedPinnedImages
+	}
 }
 
 // ReloadRegistries reloads the registry configuration from the Configs
--- a/server/server.go
+++ b/server/server.go
@@ -577,6 +577,9 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 				logrus.Errorf("Unable to reload configuration: %v", err)
 				continue
 			}
+			// ImageServer compiles the list with regex for both
+			// pinned and sandbox/pause images, we need to update them
+			s.StorageImageServer().UpdatePinnedImagesList(append(s.config.PinnedImages, s.config.PauseImage))
 		}
 	}()
 
--- a/test/reload_config.bats
+++ b/test/reload_config.bats
@@ -230,3 +230,39 @@ EOF
 	#then
 	wait_for_log '"updating runtime configuration"'
 }
+
+@test "reload config should update 'pinned_images'" {
+	OPTION="pinned_images"
+	EXAMPLE_IMAGE=quay.io/crio/fedora-crio-ci:latest
+	# add pinned_images config
+	printf '[crio.image]\npinned_images = [""]\n' > "$CRIO_CONFIG_DIR"/00-default
+	printf '[crio.image]\npinned_images = ["%s"]\n' $EXAMPLE_IMAGE > "$CRIO_CONFIG_DIR"/01-overwrite
+	# image is not pinned
+	output=$(crictl images -o json | jq ".images[] | select(.repoTags[] == \"$EXAMPLE_IMAGE\") |.pinned")
+	[ "$output" == "false" ]
+	reload_crio
+	# image becomes pinned
+	expect_log_success $OPTION $EXAMPLE_IMAGE
+	output=$(crictl images -o json | jq ".images[] | select(.repoTags[] == \"$EXAMPLE_IMAGE\") |.pinned")
+	[ "$output" == "true" ]
+}
+
+@test "reload config should update 'pinned_images' and only 'pause_image' is pinned" {
+	OPTION="pause_image"
+	printf '[crio.image]\npinned_images = [""]\n' > "$CRIO_CONFIG_DIR"/00-default
+	reload_crio
+	output=$(crictl images -o json | jq '.images[] | select(.pinned == true) | .repoTags[]')
+	# only pause image is pinned
+	[[ "$output" == *"pause"* ]]
+}
+
+@test "reload config should update 'pause_image' and it becomes 'pinned_images'" {
+	OPTION="pause_image"
+	EXAMPLE_IMAGE=quay.io/crio/fedora-crio-ci:latest
+	printf '[crio.image]\npinned_images = [""]\npause_image = "%s"\n' $EXAMPLE_IMAGE > "$CRIO_CONFIG_DIR"/04-overwrite
+	reload_crio
+	expect_log_success $OPTION $EXAMPLE_IMAGE
+	output=$(crictl images -o json | jq '.images[] | select(.pinned == true) | .repoTags[]')
+	# pause image is pinned
+	[[ "$output" == *"fedora-crio-ci"* ]]
+}
--- a/contrib/test/ci/integration.yml
+++ b/contrib/test/ci/integration.yml
@@ -54,6 +54,7 @@
         state: present
       loop: "{{ ['cgroups.bats'] | product(kata_skip_cgroups_tests) \
         + ['command.bats'] | product(kata_skip_command_tests) \
+        + ['reload_config.bats'] | product(kata_skip_reload_config_tests) \
         + ['config.bats'] | product(kata_skip_config_tests) \
         + ['config_migrate.bats'] | product(kata_skip_config_migrate_tests) \
         + ['crio-wipe.bats'] | product(kata_skip_crio_wipe_tests) \
--- a/test/mocks/criostorage/criostorage.go
+++ b/test/mocks/criostorage/criostorage.go
@@ -185,6 +185,18 @@ func (mr *MockImageServerMockRecorder) UntagImage(arg0, arg1 interface{}) *gomoc
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "UntagImage", reflect.TypeOf((*MockImageServer)(nil).UntagImage), arg0, arg1)
 }
 
+// UpdatePinnedImagesList mocks base method.
+func (m *MockImageServer) UpdatePinnedImagesList(arg0 []string) {
+	m.ctrl.T.Helper()
+	m.ctrl.Call(m, "UpdatePinnedImagesList", arg0)
+}
+
+// UpdatePinnedImagesList indicates an expected call of UpdatePinnedImagesList.
+func (mr *MockImageServerMockRecorder) UpdatePinnedImagesList(arg0 interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "UpdatePinnedImagesList", reflect.TypeOf((*MockImageServer)(nil).UpdatePinnedImagesList), arg0)
+}
+
 // MockRuntimeServer is a mock of RuntimeServer interface.
 type MockRuntimeServer struct {
 	ctrl     *gomock.Controller


server/server.go
```diff
@@ -583,6 +583,19 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 	log.Infof(ctx, "Registered SIGHUP reload watcher")
 }
 
+// reloadConfigByHUP reload the config and UpdateImageCache
+func (s *Server) reloadConfigByHUP() error {
+	if err := s.config.Reload(); err != nil {
+		return err
+	}
+	if len(s.config.PinnedImages) > 0 {
```
sohankunkerkar, line 591: I think instead of checking the length, how about we change the logic of `ReloadPinnedImages` to return a boolean type? This will ensure that `UpdateImageCache` is called only when `ReloadPinnedImages` returns true indicating an update to the list of pinned_images.

sohankunkerkar (commented):
@roman-kiselenko Thanks for working on this issue. There's already an integration test for this to cover the use case of updating the pinned_images when cri-o restarts, could you add another one when cri-o reloads?

sohankunkerkar:
How about we update the image cache in `reloadConfigByHUP` only when there's an update to the list of pinned_images, instead of passing the `updateImageCache func([]string)` to `Reload`?
```go
func (c *Config) ReloadPinnedImages(newConfig *Config) bool {
	updatedPinnedImages := make([]string, len(newConfig.PinnedImages))
	updatedPinnedImageList := false

	for i, image := range newConfig.PinnedImages {
		if i < len(c.PinnedImages) && image == c.PinnedImages[i] {
			continue
		}
		updatedPinnedImages[i] = image
		updatedPinnedImageList = true
	}

	if updatedPinnedImageList {
		logrus.Infof("Updated new pinned images: %+v", updatedPinnedImages)
		c.PinnedImages = updatedPinnedImages
	}
	
	return updatedPinnedImageList
}
```

And then in reloadConfigByHUP, use this:
```go
// reloadConfigByHUP reload the config and UpdateImageCache
func (s *Server) reloadConfigByHUP() error {
	if err := s.config.Reload(); err != nil {
		return err
	}
	if if s.config.ReloadPinnedImages() {
		if err := s.StorageImageServer().UpdateImageCache(s.config.PinnedImages); err != nil {
			return err
		}
	}
	return nil
}
```

roman-kiselenko:
> How about we update the image cache in `reloadConfigByHUP` only when there's an update to the list of pinned_images, instead of passing the `updateImageCache func([]string)` to `Reload`?
> 
> ```go
> func (c *Config) ReloadPinnedImages(newConfig *Config) bool {
> 	updatedPinnedImages := make([]string, len(newConfig.PinnedImages))
> 	updatedPinnedImageList := false
> 
> 	for i, image := range newConfig.PinnedImages {
> 		if i < len(c.PinnedImages) && image == c.PinnedImages[i] {
> 			continue
> 		}
> 		updatedPinnedImages[i] = image
> 		updatedPinnedImageList = true
> 	}
> 
> 	if updatedPinnedImageList {
> 		logrus.Infof("Updated new pinned images: %+v", updatedPinnedImages)
> 		c.PinnedImages = updatedPinnedImages
> 	}
> 	
> 	return updatedPinnedImageList
> }
> ```
> 
> And then in reloadConfigByHUP, use this:
> 
> ```go
> // reloadConfigByHUP reload the config and UpdateImageCache
> func (s *Server) reloadConfigByHUP() error {
> 	if err := s.config.Reload(); err != nil {
> 		return err
> 	}
> 	if if s.config.ReloadPinnedImages() {
> 		if err := s.StorageImageServer().UpdateImageCache(s.config.PinnedImages); err != nil {
> 			return err
> 		}
> 	}
> 	return nil
> }
> ```

cool, I'm going to use your solutions, thanks 🙏

sohankunkerkar:
> The only thing here is how to pass `newConfig` variable. That the reason for passing `updateImageCache func([]string)` to `Reload()` 🤔
> 
> ```go
> s.config.ReloadPinnedImages() 
> ```


hmm... I think you need to wrap `ReloadPinnedImages` in another function, maybe?

```go
func (c *Config) IsPinnedImagesUpdated() bool {
	return c.ReloadPinnedImages(c)
}
```

pkg/config/reload.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 // Reload reloads the configuration for the single crio.conf and the drop-in
-// configuration directory.
-func (c *Config) Reload() error {
+// configuration directory. Method takes a function to update PinnedImages as parameter.
+func (c *Config) Reload(updateImageCache func([]string) error) error {
```
haircommander, line 17: is it possible to use the `_inject_test.go` pattern where we can override for the tests specifically by calling a hidden setter?

roman-kiselenko: I've changed a bit of implementation and used another way of passing `Config.` PTAL

roman-kiselenko:
> > The only thing here is how to pass `newConfig` variable. That the reason for passing `updateImageCache func([]string)` to `Reload()` 🤔
> > ```go
> > s.config.ReloadPinnedImages() 
> > ```
> 
> hmm... I think you need to wrap `ReloadPinnedImages` in another function, maybe?
> 
> ```go
> func (c *Config) IsPinnedImagesUpdated() bool {
> 	return c.ReloadPinnedImages(c)
> }
> ```

I've changed a bit of implementation and used another way of passing Config. PTAL

roman-kiselenko:
> > The only thing here is how to pass `newConfig` variable. That the reason for passing `updateImageCache func([]string)` to `Reload()` 🤔
> > ```go
> > s.config.ReloadPinnedImages() 
> > ```
> 
> hmm... I think you need to wrap `ReloadPinnedImages` in another function, maybe?
> 
> ```go
> func (c *Config) IsPinnedImagesUpdated() bool {
> 	return c.ReloadPinnedImages(c)
> }
> ```

The `Reload()` function contains a `newConfig` variable with fresh PinnedImages. I will rewrite the `Reload()` logic a little so the `PinnedImages` attribute will update appropriately.

sohankunkerkar (approved):
Changes LGTM
/lgtm

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

internal/storage/image.go
```diff
@@ -897,6 +901,26 @@ func (st nativeStorageTransport) ResolveReference(ref types.ImageReference) (typ
 	return istorage.ResolveReference(ref)
 }
 
+// UpdateImageCache invalidate imageCache if the pinned image from config exists;
```
haircommander, line 904: nit: invalidates*

test/reload_config.bats
```diff
@@ -230,3 +230,13 @@ EOF
...
 }
+
+@test "reload config should update 'pinned_images'" {
+	# given
+	printf '[crio.image]\npinned_images = [""]\n' > "$CRIO_CONFIG_DIR"/00-default
+	printf '[crio.image]\npinned_images = ["quay.io/crio/hello-wasm:latest"]\n' > "$CRIO_CONFIG_DIR"/01-overwrite
+	# when
+	reload_crio
+	#then
+	wait_for_log 'Updated new pinned images'
```
haircommander, line 241: I think this test could do a `crictl images -o json` and find the pinned images, then do so after we do the reload and make sure the images are updated

server/server.go
```diff
@@ -573,10 +573,17 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 		for {
 			// Block until the signal is received
 			<-ch
-			if err := s.config.Reload(); err != nil {
+			newConfig, err := s.config.Reload()
+			if err != nil {
 				logrus.Errorf("Unable to reload configuration: %v", err)
 				continue
 			}
+			if s.config.IsPinnedImagesUpdated(newConfig) {
```
haircommander, line 581: It's not clear to me how this condition will be true if we're replacing the config in place in Reload(). won't newConfig == s.config at this point?

roman-kiselenko: The `Reload()` does not replace the config; it only updates the `newConfig` variable, and all the changes vanish with returns.

https://github.com/cri-o/cri-o/blob/0ae4367ef3b337894b13cd199b007653a5c2b6be/pkg/config/reload.go#L26-L38

test/reload_config.bats
roman-kiselenko: This comment led me to understand that the current implementation is wrong, and it doesn't update the pinned attribute :) 

Thanks 🙏

haircommander: I trust the config hadn't pinned anything else, but just to future proof, can you also do the opposite check before we reload_crio?

server/server.go
haircommander: wait but we do replace the c.config in reload: https://github.com/cri-o/cri-o/blob/0ae4367ef3b337894b13cd199b007653a5c2b6be/pkg/config/reload.go#L148-L156 we absorb the newConfig, and then take select changes into the c.config (old config) so how would the newConfig and c.config ever differ  WRT pinned images?

roman-kiselenko: We're merging the old config pinned image list with the new pinned image list, and if there are changes, we update the pinned image list in `imageService`. Because we're relying on the returned state from `ReloadPinnedImages` we can perform `imageService` update only outside the Reload method.

hexfusion:
Just checking in on this how are we feeling about the current direction @haircommander @sohankunkerkar

internal/storage/image.go
```diff
@@ -897,6 +900,11 @@ func (st nativeStorageTransport) ResolveReference(ref types.ImageReference) (typ
 	return istorage.ResolveReference(ref)
 }
 
+// UpdatePinnedImagesList updates pinned images list in imageService.
+func (svc *imageService) UpdatePinnedImagesList(configPinnedImage []string) {
+	svc.regexForPinnedImages = append(svc.regexForPinnedImages, CompileRegexpsForPinnedImages(configPinnedImage)...)
```
haircommander, line 905: I'm not sure we should append here. Wouldnt' we just want to overwrite the slice? someone could theoretically unpin an image and we'd want it to qualify for GC then

pkg/config/reload.go
```diff
@@ -13,20 +13,20 @@ import (
 )
 
 // Reload reloads the configuration for the single crio.conf and the drop-in
-// configuration directory.
-func (c *Config) Reload() error {
+// configuration directory. Returns updated Config struct.
+func (c *Config) Reload() (*Config, error) {
```
hexfusion, line 17: this feels a little awkward since Reload is a method of Config.

pkg/config/reload_test.go
```diff
@@ -35,9 +35,10 @@ var _ = t.Describe("Config", func() {
 			Expect(sut.UpdateFromFile(filePath)).To(Succeed())
 
 			// When
-			err := sut.Reload()
+			cfg, err := sut.Reload()
```
hexfusion, line 38: In this use case sut is aleady reference to `Config` right?

roman-kiselenko:
@haircommander @sohankunkerkar @hexfusion 

The trick with `Reload()` and return config confuses readers. I've changed the implementation to make it more readable. 

The `Reload()` method now updates the config struct, and fields `PinnedImages` and `PauseImage` are passed to `UpdatePinnedImagesList`.
https://github.com/cri-o/cri-o/blob/fbb899bb89a73917da0649a5c702964637454b32/internal/storage/image.go#L857-L859 

Both fields are updated in ImageService; hence, we must update both when HUP happens.

Also, I've added a test to catch some corner cases.

@cri-o/cri-o-maintainers PTAL

haircommander:
Yes I like this better! Good rewrite, and thanks for your patience
/approve
Tagging another @cri-o/cri-o-maintainers for the final review

server/server.go
```diff
@@ -577,6 +577,12 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 				logrus.Errorf("Unable to reload configuration: %v", err)
 				continue
 			}
+			// ImageServer compiles the list with regex for both
+			// pinned and sandbox/pause images, we need to update them
+			s.StorageImageServer().UpdatePinnedImagesList(
+				s.config.PauseImage,
```
haircommander, line 583: Totally non blocking nit: I think i'd prefer combining the config fields here instead of passing two args

internal/storage/image.go
```diff
@@ -897,6 +900,14 @@ func (st nativeStorageTransport) ResolveReference(ref types.ImageReference) (typ
 	return istorage.ResolveReference(ref)
 }
 
+// UpdatePinnedImagesList updates pinned images list in imageService.
+func (svc *imageService) UpdatePinnedImagesList(pauseImage string, pinnedImages []string) {
+	if pauseImage != "" {
+		pinnedImages = append(pinnedImages, pauseImage)
```
haircommander, line 906: If pause image is changed to ", what happens here? I forget, do we fallback to anything in sandbox_run_linux.go, and if so should we include it here?

haircommander (approved):
Actually a couple of notes but non blocking

internal/storage/image.go
roman-kiselenko: That's a good catch; I'll update it. :pray:

roman-kiselenko: I'm a bit confused; could you elaborate? 

AFAIU, the pause image in runPodSandbox comes from the config. Hence, it will be updated on HUP.  

https://github.com/cri-o/cri-o/blob/fbb899bb89a73917da0649a5c702964637454b32/server/sandbox_run_linux.go#L461-L464

roman-kiselenko: > If pause image is changed to "

This will not happen because the empty value in the config will not be parsed

> Unable to reload configuration: invalid reference format

Hence, `DefaultPauseImage` will always be in the list of `PinnedImages` and as `PauseImage`.

pkg/config/reload.go
```diff
@@ -73,6 +75,7 @@ func (c *Config) Reload() error {
 	}
 	cdi.GetRegistry(cdi.WithSpecDirs(newConfig.CDISpecDirs...))
 
+	cdi.GetRegistry(cdi.WithSpecDirs(newConfig.CDISpecDirs...))
```
sohankunkerkar, line 78: why do we have two entries here?

roman-kiselenko: ah, bad rebase thx

pkg/config/reload.go
```diff
@@ -143,17 +146,23 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
 }
 
 // ReloadPinnedImages updates the PinnedImages with the provided `newConfig`.
```
hexfusion, line 148: optional: Might make sense to add a few words on what the bool this public method now returns indicates?

server/server.go
```diff
@@ -577,6 +577,9 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 				logrus.Errorf("Unable to reload configuration: %v", err)
 				continue
 			}
+			// ImageServer compiles the list with regex for both
+			// pinned and sandbox/pause images, we need to update them
+			s.StorageImageServer().UpdatePinnedImagesList(append(s.config.PinnedImages, s.config.PauseImage))
```
sohankunkerkar, line 582: With this change, you can drop this code: https://github.com/cri-o/cri-o/blob/main/internal/storage/image.go#L857

pkg/config/reload.go
```diff
@@ -143,17 +144,23 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
...
+// The method print log in case of any updates.
```
hexfusion, line 147: :+1:

internal/storage/image.go
sohankunkerkar: Could you add an integration test for this?

sohankunkerkar: Just to make sure everything works as expected.

pkg/config/reload.go
roman-kiselenko: Thx, I've rewritten this method and removed the `bool` return. Not log print happens in the same way as in other methods with usage of `logConfig()`

internal/storage/image.go
roman-kiselenko: I've implemented that checking here https://github.com/cri-o/cri-o/pull/7976/files#diff-4ea762d405a8752ac8b728bfbe94b8ee5f962013a1e65d4e5d20208556418447R255-R257

server/server.go
sohankunkerkar: ah, right!

roman-kiselenko:
The code has been changed a lot since last approve, I've re-request review 🙏

hexfusion (commented):
I am not a maintainer of this code but this approach looks much better :+1:

haircommander:
/approve
Still, @sohankunkerkar for the final lgtm?

hexfusion:
The integrations might be flakes but I think you need to fix the shfmt linter
```
make shfmt
```
probably looking for whitespace

sohankunkerkar:
> The integrations might be flakes but I think you need to fix the shfmt linter
> 
> ```
> make shfmt
> ```
> 
> probably looking for whitespace

@roman-kiselenko Thanks for your patience. One last fix before `/lgtm`

hexfusion:
fedora-kata test is failing due to rate limiting
> rateLimitExceeded

roman-kiselenko:
> fedora-kata test is failing due to rate limiting
> 
> > rateLimitExceeded

[integration / userns / amd64](https://github.com/cri-o/cri-o/actions/runs/8737513570/job/23974871352?pr=7976#step:13:416) - failed and I cant find a reason. 🤔 

[integration / conmon / arm64](https://github.com/cri-o/cri-o/actions/runs/8737513570/job/23974870432?pr=7976#step:13:352) - NRI tests are flaky

sohankunkerkar (approved):
/lgtm

sohankunkerkar:
>last failure might not of been a flake

Yup, it's a legit one.

hexfusion:
cc @roman-kiselenko PTAL

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/7976/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1780966933546405888/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt
```
not ok 216 reload config should update 'pinned_images' and 'pause_image' is pinned automatically
# (in test file reload_config.bats, line 260)
#   `[[ "$output" == *"wasm"* ]]' failed
```

roman-kiselenko:
> cc @roman-kiselenko PTAL
> 
> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/7976/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1780966933546405888/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt
> ```
> not ok 216 reload config should update 'pinned_images' and 'pause_image' is pinned automatically
> # (in test file reload_config.bats, line 260)
> #   `[[ "$output" == *"wasm"* ]]' failed
> ```

I can't reproduce it locally :sad:

going to poke around logs for a while

roman-kiselenko:
> I think this failure is only for Kata; the other two integration tests are looking fine

Do you think we can disable them for Kata?

I can't find the reason why it failed.

sohankunkerkar:
@littlejawa  Do you have any idea here?

sohankunkerkar:
```sh
 level=error msg="Getting /opt/kata/bin/containerd-shim-kata-v2 OCI runtime features failed: io.containerd.kata.v2: shim namespace cannot be empty: exit status 1" file="config/config.go:1306"
```

roman-kiselenko:
> ```shell
>  level=error msg="Getting /opt/kata/bin/containerd-shim-kata-v2 OCI runtime features failed: io.containerd.kata.v2: shim namespace cannot be empty: exit status 1" file="config/config.go:1306"
> ```

That looks like a problem; I've found this https://github.com/kata-containers/kata-containers/issues/541 

TBH I don't know how CRI-O works with Kata.  😞

roman-kiselenko:
Also [image pull & list disable on Kata](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8019/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1780041107673976832/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt#:~:text=ok%2030%20image%20pull%20and%20list%20by%20manifest,and%20manifest%20list%20digest%27%20not%20working%20in%20kata)

contrib/test/ci/vars.yml
```diff
@@ -44,7 +44,7 @@ ssh_user: "{{ lookup('env','USER') }}"
 ssh_location: "{{ lookup('env','HOME') }}/.ssh/id_rsa"
 
 # variables for testing with the kata-containers runtime
-kata_version: "3.1.3"
+kata_version: "3.3.0"
```
sohankunkerkar, line 47: There's already a [PR](https://github.com/cri-o/cri-o/pull/7958) to test the new version of Kata.

roman-kiselenko: oh, thanks, I didn't spot this.

sohankunkerkar:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/7976/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1782419051419537408/artifacts/fedora-integration/cri-o-gather/artifacts/testout.txt

I think [this](https://github.com/cri-o/cri-o/pull/8057) should solve the issue.

hexfusion:
Have we tested inside a kata instance to verify what is going on? Does the test provide the details of how to do this?

hexfusion:
cc @roman-kiselenko

roman-kiselenko:
> We can skip this test here: https://github.com/cri-o/cri-o/blob/main/contrib/test/ci/vars.yml#L101 and open an issue to track the test failure.

I've skipped this test for `kata.` I'm going to create an issue for this after a successful merge.

sohankunkerkar (approved):
/lgtm

hexfusion:
cc @sohankunkerkar can we get a cherry pick to 1.29?

hexfusion:
thanks for seeing this through @roman-kiselenko

littlejawa:
Hey there,
Sorry for being late on this one, I was offline for a while. I'll have a look at the issue and try to understand what's wrong on the kata side.

roman-kiselenko:
> Hey there,
> Sorry for being late on this one; I was offline for a while. I'll have a look at the issue and try to understand what's wrong on the kata side.

hey, thank you for your time.

We have an umbrella issue here https://github.com/cri-o/cri-o/issues/8074

Issue #7866, opened 2024-03-11
Aim to drop containers/podman as a dependency


kannon92:
As we were working on #https://github.com/cri-o/cri-o/pull/7865, I noticed that we have some dependencies on podman.

@haircommander mentioned that we should start phasing that out or moving components that we need to c/common.

From a quick search, we use rootless, annotations, criu, and errorhandling.

- [x] remove annotations: https://github.com/cri-o/cri-o/pull/7867
- [x] remove checkpoint/criu utilties https://github.com/containers/common/pull/1904
- [x] rootless https://github.com/cri-o/cri-o/pull/7990
- [x] selinux
- [x] /specgen/generate: https://github.com/cri-o/cri-o/pull/7891
- [x] lookup: https://github.com/cri-o/cri-o/pull/7893
- [x] errorhandling: https://github.com/cri-o/cri-o/pull/7868

Maybe we can start moving what we need to c/common so we can drop this as a library.
kehannon@kehannon-thinkpadp1gen4i:~/Work/cri-o$ grep -r 'podman' * --exclude-dir=vendor
```
internal/config/cgmgr/cgroupfs_linux.go:        "github.com/containers/podman/v4/pkg/rootless"
internal/config/cgmgr/systemd_linux.go: "github.com/containers/podman/v4/pkg/rootless"
internal/config/device/device_linux.go: createconfig "github.com/containers/podman/v4/pkg/specgen/generate"
internal/factory/container/container_test.go:   "github.com/containers/podman/v4/pkg/annotations"
internal/factory/container/container.go:        "github.com/containers/podman/v4/pkg/annotations"
internal/lib/checkpoint.go:     "github.com/containers/podman/v4/pkg/annotations"
internal/lib/checkpoint.go:     "github.com/containers/podman/v4/pkg/checkpoint/crutils"
internal/lib/checkpoint.go:     // To correctly track deleted files, let's go through the output of 'podman diff'
internal/lib/restore.go:        "github.com/containers/podman/v4/pkg/annotations"
internal/lib/restore.go:        "github.com/containers/podman/v4/pkg/checkpoint/crutils"
internal/lib/checkpoint_test.go:        "github.com/containers/podman/v4/pkg/criu"
internal/lib/container_server.go:       "github.com/containers/podman/v4/pkg/annotations"
internal/lib/container_server_freebsd.go:       "github.com/containers/podman/v4/pkg/annotations"
[...]

PR #7990, opened 2024-04-09
Use unshare instead of rootless

kannon92:
#### What type of PR is this?
/kind cleanup

Drop podman rootless in favor of unshare in c/storage.

Partially #7866 

```release-note
NONE
```


server/sandbox_run_linux.go
```diff
@@ -320,20 +320,27 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
 		return nil, errors.New("infra container not found")
 	}
 
-	uids, err := rootless.ReadMappingsProc(fmt.Sprintf("/proc/%d/uid_map", ic.State().Pid))
+	uids, gids, err := unshare.GetHostIDMappings(strconv.Itoa(ic.State().Pid))
```
kannon92, line 323: This is the ugly code.

Struct has the exact same shape but it comes from different packages so technically it is a different type. Did a type cast effectively.

haircommander:
@giuseppe would you PTAL?

it looks fine to me, not overly ugly (though the conversion code isn't ideal, I think it's alright)

server/sandbox_run_linux.go
```diff
@@ -320,20 +320,28 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
-		return nil, err
-	}
-
-	mappings := idtools.NewIDMappingsFromMaps(uids, gids)
+	mappings := convertToMappings(uids, gids)
 	ic.SetIDMappings(mappings)
 	return mappings, nil
 }
 
+func convertToMappings(uids, guids []spec.LinuxIDMapping) *idtools.IDMappings {
```
kwilczynski, line 332: Can you make this function look more like the one above? The `convertToStorageIDMap()`. Especially as you know the number of UIDs and GIDs, etc.  Also, you could move, so both functions are closer together.

kannon92: Which one are you suggesting?

I have to do this conversion because the rootless and unshare code have slightly different types.

kannon92: This function is one line past the end. Not sure how it can get closer. I could define it in the function if you'd prefer

kwilczynski: This is what I had in mind (an example):

```go
func convertToStorageIDMappings(uidMappings, gidMappings []spec.LinuxIDMapping) *idtools.IDMappings {
	uids := make([]idtools.IDMap, len(uidMappings))
	gids := make([]idtools.IDMap, len(gidMappings))

	for i, v := range uidMappings {
		uids[i] = idtools.IDMap{
			ContainerID: int(v.ContainerID),
			HostID:      int(v.HostID),
			Size:        int(v.Size),
		}
	}
	for i, v := range gidMappings {
		gids[i] = idtools.IDMap{
			ContainerID: int(v.ContainerID),
			HostID:      int(v.HostID),
			Size:        int(v.Size),
		}
	}

	return idtools.NewIDMappingsFromMaps(uids, gids)
}
```

To make it look more like the other.  So, avoid `append()`s there, since we know the length in advance, etc.

giuseppe (approved):
LGTM

server/sandbox_run_linux.go
kannon92: Holding to address your comment. Will do that today.

kannon92: Updated.

PR #7471, opened 2023-11-09
kata: support confidential container's api for pulling image in the guest

littlejawa:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

For confidential container, we want to have the ability to pull the image on the guest side, and not rely on the host's repository.

To achieve this, an extension to the kata shim's mount options for CreateContainer was added. Provided the right information through these options, the kata agent in the remote VM will proceed with the PullImage before handling the actual CreateContainer.

This commit adds support for this method.

As part of this, I am adding an option to the runtime configuration, making it possible to select whether or not the additional data should be sent.
We need it to be configurable, because there are use cases for confidential containers where pull-in-guest is not desired (like: sharing a volume from the host to the VM, containing the container images).

Additionally: with this feature enabled, cri-o will still comply with kubernetes' "PullImageRequest", even though the resulting container image is not used.
This can be an issue, not only in terms of resource usage: if the workload is encrypted and cri-o can't access it, cri-o will fail the PullImageRequest, and kubernetes will never send CreateContainer.
I'm looking for a way to avoid this situation, maybe making cri-o skip PullImage for a specific runtime. This will be a separate PR - suggestions are welcome.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

Here is how I've been testing it:

Prerequisite: build/install kata from the CCv0 branch (https://github.com/kata-containers/kata-containers/tree/CCv0).
Build/install cloud-api-adaptor and related podvm image, using libvirt as the cloud provider (and have a libvirt server available that you can talk to).
(https://github.com/confidential-containers/cloud-api-adaptor/)

Add the following configuration to crio, to define new runtime classes:
- kata: regular kata container
- kata-remote: peer-pod container - in this scenario, cloud-api-adapator makes a PullImage request to the kata-agent as part of processing the CreateContainer request. This method is considered a hack, and is bound to be deprecated.
- kata-cc: confidential container - in this scenario, the changes in this PR are exercised: cri-o provides the needed information to the agent as part of the CreateContainer request, cloud-api-adaptor forwards it without doing anything more, and the agent performs the PullImage on its own.

/etc/crio/crio.conf.d/50-kata-runtimes
```
[crio.runtime.runtimes.kata]
  runtime_path = "/usr/local/bin/containerd-shim-kata-v2"
  runtime_root = "/run/vc"
  runtime_type = "vm"
  privileged_without_host_devices = true
[...]

PR #8008, opened 2024-04-12
Confidential Containers - Skip pullimage for runtimes that are handling it

### LINKED ISSUE #8261: RFC: how to support the use of encrypted containers in Confidential Containers
In order to support Confidential Containers in CRI-O, we have added the ability for cri-o to let the kata runtime pull the container image in the guest VM, ignoring the image that CRI-O have pulled on the host.
See: https://github.com/cri-o/cri-o/pull/7471

This does not prevent CRI-O from pulling the image, but the image is ignored, which is fine from the Confidential Containers point of view.

Now pulling the image requires reading its layer. In the case of encrypted images, since the decryption key may not be available to CRI-O, the pull will fail, preventing any further progress in the container creation.

In order to avoid that, we tried to prevent the pull from happening.
Answering to Kubernetes' "Pull request" with an acknowledgement is easy enough, and allows to proceed with CreateContainer. But then we faced issues in CRI-O itself.

During the CreateContainer phase, CRI-O needs to prepare the storage for the container, and generate the spec file.
Those actions are performed by the [createSandboxContainer()](https://github.com/cri-o/cri-o/blob/9b9452f662381706cd909a36336514025fc5fb5c/server/container_create_linux.go#L127) function.

It requires using the container image layers as the base for the storage, and need access to the image's metadata (manifest and config).
Since the image was not downloaded, the function fails because it doesn't find the required information.

## What we need

In order to properly start a container with Confidential Containers, we need CRI-O to create a folder that can be used as the working directory for the runtime process. It doesn't need to contain the rootfs of the container, but the other files usually found there should be available (e.g: config.json file with appropriate settings).

We also need to ensure that CRI-O can still report the right container status to Kubernetes (e.g: image digest).

## What I tried so far

I've tried to modify the createSandboxContainer() function to use empty values or sane default values wherever possible when we are in the "Confidential Containers" case. This leads to a lot of changes everywhere in the function, and ended up with failures outside of it, including in kata, because of missing information.

In order to get actual meaningful settings, I tried pretending that the image to use was the pause image.
The change is simpler, but then we end up reporting to kubernetes that the container is running the pause image, and CRI-O tells the kata agent to run the pause image's command line.

This work can be reviewed here: https://github.com/cri-o/cri-o/pull/8008
I've stopped digging in this direction when I realized I would need the image's metadata anyway.

## Image manifest and config

The whole problem seems to be that the manifest and config are not available to cri-o when it prepares the container storage.

In an encrypted image scenario, we want to avoid reading the layers (which are encrypted), but accessing the manifest and config should be possible.
If that's the case, I would need a way to download/access those information without involving the download/decryption of the layers.

At the moment, cri-o is pulling the image using a single call to a ["copy.Image()"
function](https://pkg.go.dev/github.com/containers/image/v5@v5.29.2/copy) from the containers/image module.
I've tried to find a way to pull only the metadata with this module, but failed to find the right way to do it.

This is my understanding so far, based on the above investigations, but I really need confirmation and guidance from the cri-o community to confirm that, and to understand how to proceed from there.

littlejawa:
#### What type of PR is this?

/kind feature

With https://github.com/cri-o/cri-o/pull/7471 we introduced a way to support Confidential Container's feature of pulling the image in the guest VM.
When this happens, cri-o is still pulling the image on the host, because kubernetes keeps sending the "PullImageRequest", and cri-o has no reason to not process it.

This has two drawbacks:
1) it is a waste of time and resources, as the image that crio will pull and prepare will not be used anyway
2) it can block the actual execution of the container in the Confidential Container use case: if the image is encrypted, and the key is not shared with crio, then crio will fail to prepare the image because it can't read it.
   This failure will block the container creation, as kubernetes will not proceed with CreateContainer if PullImage failed.

This PR is meant to make crio skip the pull image phase when the runtime is configured with the "runtime_pull_image" flag introduced in #7471 

Fixes: #8261 

I have modified the code in crio to skip the pull image processing when the runtime is configured with the "runtime_pull_image" flag.
Crio then just reports a success status to kubernetes, which is happy with that, even when subsequent ImageList or ImageStatus report the image is missing.

Now without an image in the local store, cri-o will fail to create the container. This required further changes in create_sandbox_container and runtimeService, to get information from the remote repository, without pulling the image.

I kept those changes in separate commits for easier review, but I suspect it would make sense to squash them, and get a single, functional commit.

```release-note
Allow the runtime to be configured to skip the pull image phase, and make sure that cri-o can still run the container with it. This allows the use of the Confidential Containers runtime, that pulls the container image separately.
```

### DIFF
--- a/internal/config/ociartifact/ociartifact.go
+++ b/internal/config/ociartifact/ociartifact.go
@@ -11,9 +11,11 @@ import (
 	"time"
 
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/blobinfocache"
 	"github.com/containers/image/v5/types"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 
 	"github.com/cri-o/cri-o/internal/log"
 )
@@ -84,37 +86,14 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
 		opts = &PullOptions{}
 	}
 
-	name, err := o.impl.ParseNormalizedNamed(img)
+	src, err := o.getSourceFromImageName(ctx, img, opts)
 	if err != nil {
-		return nil, fmt.Errorf("parse image name: %w", err)
-	}
-	name = reference.TagNameOnly(name) // make sure to add ":latest" if needed
-
-	ref, err := o.impl.NewReference(name)
-	if err != nil {
-		return nil, fmt.Errorf("create docker reference: %w", err)
+		return nil, err
 	}
 
-	src, err := o.impl.NewImageSource(ctx, ref, opts.SystemContext)
+	parsedManifest, err := o.getParsedManifest(ctx, src, opts)
 	if err != nil {
-		return nil, fmt.Errorf("build image source: %w", err)
-	}
-
-	manifestBytes, mimeType, err := o.impl.GetManifest(ctx, src, nil)
-	if err != nil {
-		return nil, fmt.Errorf("get manifest: %w", err)
-	}
-
-	parsedManifest, err := o.impl.ManifestFromBlob(manifestBytes, mimeType)
-	if err != nil {
-		return nil, fmt.Errorf("parse manifest: %w", err)
-	}
-	if opts.EnforceConfigMediaType != "" && o.impl.ManifestConfigInfo(parsedManifest).MediaType != opts.EnforceConfigMediaType {
-		return nil, fmt.Errorf(
-			"wrong config media type %q, requires %q",
-			o.impl.ManifestConfigInfo(parsedManifest).MediaType,
-			opts.EnforceConfigMediaType,
-		)
+		return nil, err
 	}
 
 	layers := o.impl.LayerInfos(parsedManifest)
@@ -162,6 +141,130 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
 		Data: layerBytes,
 		Path: keyPath,
 	}, nil
+}
+
+func (o *OCIArtifact) GetManifest(ctx context.Context, img string, opts *PullOptions) (manifest.Manifest, error) {
+	log.Infof(ctx, "Pulling manifest from ref: %s", img)
+
+	// Use default pull options
+	if opts == nil {
+		opts = &PullOptions{}
+	}
+
+	src, err := o.getSourceFromImageName(ctx, img, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	parsedManifest, err := o.getParsedManifest(ctx, src, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return parsedManifest, nil
+}
+
+func (o *OCIArtifact) GetConfig(ctx context.Context, img string, opts *PullOptions) (*v1.Image, error) {
+	log.Infof(ctx, "Getting config from ref: %s", img)
+
+	// Use default pull options
+	if opts == nil {
+		opts = &PullOptions{}
+	}
+
+	src, err := o.getSourceFromImageName(ctx, img, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	unparsedToplevel := image.UnparsedInstance(src, nil)
+	topManifest, topMIMEType, err := unparsedToplevel.Manifest(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("get manifest: %w", err)
+	}
+
+	unparsedInstance := unparsedToplevel
+	if manifest.MIMETypeIsMultiImage(topMIMEType) {
+		// This is a manifest list. We need to choose a single instance to work with.
+		manifestList, err := manifest.ListFromBlob(topManifest, topMIMEType)
+		if err != nil {
+			return nil, fmt.Errorf("parsing primary manifest as list: %w", err)
+		}
+		instanceDigest, err := manifestList.ChooseInstance(opts.SystemContext)
+		if err != nil {
+			return nil, fmt.Errorf("choosing an image from manifest list: %w", err)
+		}
+
+		unparsedInstance = image.UnparsedInstance(src, &instanceDigest)
+	}
+
+	sourcedImage, err := image.FromUnparsedImage(ctx, opts.SystemContext, unparsedInstance)
+	if err != nil {
+		return nil, fmt.Errorf("getting sourced image from unparsed image: %w", err)
+	}
+	return sourcedImage.OCIConfig(ctx)
+}
+
+func (o *OCIArtifact) getSourceFromImageName(ctx context.Context, img string, opts *PullOptions) (types.ImageSource, error) {
+	name, err := o.impl.ParseNormalizedNamed(img)
+	if err != nil {
+		return nil, fmt.Errorf("parse image name: %w", err)
+	}
+	name = reference.TagNameOnly(name) // make sure to add ":latest" if needed
+
+	ref, err := o.impl.NewReference(name)
+	if err != nil {
+		return nil, fmt.Errorf("create docker reference: %w", err)
+	}
+
+	src, err := o.impl.NewImageSource(ctx, ref, opts.SystemContext)
+	if err != nil {
+		return nil, fmt.Errorf("build image source: %w", err)
+	}
+	return src, nil
+}
+
+func (o *OCIArtifact) getParsedManifest(ctx context.Context, src types.ImageSource, opts *PullOptions) (manifest.Manifest, error) {
+	manifestBytes, mimeType, err := o.impl.GetManifest(ctx, src, nil)
+	if err != nil {
+		return nil, fmt.Errorf("get manifest: %w", err)
+	}
+
+	if manifest.MIMETypeIsMultiImage(mimeType) {
+		// This is a manifest list. We need to choose a single instance to work with.
+		manifestList, err := manifest.ListFromBlob(manifestBytes, mimeType)
+		if err != nil {
+			return nil, fmt.Errorf("parsing primary manifest as list: %w", err)
+		}
+		instanceDigest, err := manifestList.ChooseInstance(opts.SystemContext)
+		if err != nil {
+			return nil, fmt.Errorf("choosing an image from manifest list: %w", err)
+		}
+
+		unparsedInstance := image.UnparsedInstance(src, &instanceDigest)
+
+		sourcedImage, err := image.FromUnparsedImage(ctx, opts.SystemContext, unparsedInstance)
+		if err != nil {
+			return nil, fmt.Errorf("getting sourced image from unparsed image: %w", err)
+		}
+		manifestBytes, mimeType, err = sourcedImage.Manifest(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("getting manifest bytes from sourced image: %w", err)
+		}
+	}
+
+	parsedManifest, err := o.impl.ManifestFromBlob(manifestBytes, mimeType)
+	if err != nil {
+		return nil, fmt.Errorf("parse manifest: %w", err)
+	}
+	if opts.EnforceConfigMediaType != "" && o.impl.ManifestConfigInfo(parsedManifest).MediaType != opts.EnforceConfigMediaType {
+		return nil, fmt.Errorf(
+			"wrong config media type %q, requires %q",
+			o.impl.ManifestConfigInfo(parsedManifest).MediaType,
+			opts.EnforceConfigMediaType,
+		)
+	}
+	return parsedManifest, nil
 }
 
 func (o *OCIArtifact) prepareCache(ctx context.Context, opts *PullOptions) (useCache bool) {
--- a/internal/storage/runtime.go
+++ b/internal/storage/runtime.go
@@ -12,6 +12,7 @@ import (
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
 
+	"github.com/cri-o/cri-o/internal/config/ociartifact"
 	"github.com/cri-o/cri-o/internal/log"
 )
 
@@ -76,7 +77,7 @@ type RuntimeServer interface {
 	// CreateContainer creates a container with the specified ID.
 	// Pointer arguments can be nil.
 	// All other arguments are required.
-	CreateContainer(systemContext *types.SystemContext, podName, podID, userRequestedImage string, imageID StorageImageID, containerName, containerID, metadataName string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, privileged bool) (ContainerInfo, error)
+	CreateContainer(systemContext *types.SystemContext, podName, podID, userRequestedImage string, imageID StorageImageID, containerName, containerID, metadataName string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, privileged, isRuntimePullImage bool) (ContainerInfo, error)
 	// DeleteContainer deletes a container, unmounting it first if need be.
 	DeleteContainer(ctx context.Context, idOrName string) error
 
@@ -157,7 +158,7 @@ type runtimeContainerMetadataTemplate struct {
 	privileged   bool   // Applicable to both PodSandboxes and Containers
 }
 
-func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.SystemContext, containerID string, template *runtimeContainerMetadataTemplate, idMappingsOptions *storage.IDMappingOptions, labelOptions []string) (ci ContainerInfo, retErr error) {
+func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.SystemContext, containerID string, template *runtimeContainerMetadataTemplate, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, isRuntimePullImage bool) (ci ContainerInfo, retErr error) {
 	if template.podName == "" || template.podID == "" {
 		return ContainerInfo{}, ErrInvalidPodName
 	}
@@ -187,21 +188,31 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
 
 	// Pull out a copy of the image's configuration.
 	// Ideally we would call imageID.imageRef(r.storageImageServer), but storageImageServer does not have access to private data.
-	ref, err := istorage.Transport.NewStoreReference(r.storageImageServer.GetStore(), nil, template.imageID.privateID)
-	if err != nil {
-		return ContainerInfo{}, err
-	}
-	image, err := ref.NewImage(r.ctx, systemContext)
-	if err != nil {
-		return ContainerInfo{}, err
-	}
-	defer image.Close()
+	var imageConfig *v1.Image
+	if isRuntimePullImage {
+		var err error
+		artifact := ociartifact.New()
 
-	imageConfig, err := image.OCIConfig(r.ctx)
-	if err != nil {
-		return ContainerInfo{}, err
-	}
+		imageConfig, err = artifact.GetConfig(r.ctx, metadata.ImageName, &ociartifact.PullOptions{})
+		if err != nil {
+			return ContainerInfo{}, err
+		}
+	} else {
+		ref, err := istorage.Transport.NewStoreReference(r.storageImageServer.GetStore(), nil, template.imageID.privateID)
+		if err != nil {
+			return ContainerInfo{}, err
+		}
+		image, err := ref.NewImage(r.ctx, systemContext)
+		if err != nil {
+			return ContainerInfo{}, err
+		}
+		defer image.Close()
 
+		imageConfig, err = image.OCIConfig(r.ctx)
+		if err != nil {
+			return ContainerInfo{}, err
+		}
+	}
 	metadata.Pod = (containerID == metadata.PodID) // Or should this be hard-coded in callers? The caller should know whether it is creating the infra container.
 	metadata.CreatedAt = time.Now().Unix()
 	mdata, err := json.Marshal(&metadata)
@@ -222,7 +233,12 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
 	if idMappingsOptions != nil {
 		coptions.IDMappingOptions = *idMappingsOptions
 	}
-	container, err := r.storageImageServer.GetStore().CreateContainer(containerID, names, template.imageID.privateID, "", string(mdata), &coptions)
+	imageName := template.imageID.privateID
+	if isRuntimePullImage {
+		// store.CreateContainer accepts empty image names
+		imageName = ""
+	}
+	container, err := r.storageImageServer.GetStore().CreateContainer(containerID, names, imageName, "", string(mdata), &coptions)
 	if err != nil {
 		if metadata.Pod {
 			logrus.Debugf("Failed to create pod sandbox %s(%s): %v", metadata.PodName, metadata.PodID, err)
@@ -350,10 +366,10 @@ func (r *runtimeService) CreatePodSandbox(systemContext *types.SystemContext, po
 		namespace:          namespace,
 		attempt:            attempt,
 		privileged:         privileged,
-	}, idMappingsOptions, labelOptions)
+	}, idMappingsOptions, labelOptions, false)
 }
 
-func (r *runtimeService) CreateContainer(systemContext *types.SystemContext, podName, podID, userRequestedImage string, imageID StorageImageID, containerName, containerID, metadataName string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, privileged bool) (ContainerInfo, error) {
+func (r *runtimeService) CreateContainer(systemContext *types.SystemContext, podName, podID, userRequestedImage string, imageID StorageImageID, containerName, containerID, metadataName string, attempt uint32, idMappingsOptions *storage.IDMappingOptions, labelOptions []string, privileged, isRuntimePullImage bool) (ContainerInfo, error) {
 	return r.createContainerOrPodSandbox(systemContext, containerID, &runtimeContainerMetadataTemplate{
 		podName:            podName,
 		podID:              podID,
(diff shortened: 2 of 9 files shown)


littlejawa:
I managed to get to something that works.
With the current code, I can run a container with CoCo, without pulling an image locally.
And running `crictl inspect` on it shows consistent information with what is expected for the container.

@saschagrunert  - as you suggested, I am using the OCI artifact implementation to retrieve the manifest and config.
I created dedicated functions there, and using them where needed.

I did not do a separate implementation of RuntimeServer, because I thought there is not enough changes in it to separate them. But I can revisit that as needed.

server/image_pull.go
```diff
@@ -20,10 +21,43 @@ import (
 
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/storage"
+	"github.com/cri-o/cri-o/pkg/config"
 	"github.com/cri-o/cri-o/server/metrics"
 	"github.com/cri-o/cri-o/utils"
 )
 
+// GetRuntimeHandlerForPod returns the runtime handler for the given pod config.
+func (s *Server) GetRuntimeHandlerForPod(ctx context.Context, cfg *types.PodSandboxConfig) (string, *config.RuntimeHandler) {
```
haircommander, line 30: nit: can this be defined below where it's called.
does this need to be exported?

server/container_create_linux.go
```diff
@@ -161,10 +164,30 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	var someNameOfTheImage *references.RegistryImageReference
+	var imageID storage.StorageImageID
+	var someRepoDigest string
+	var imageAnnotations map[string]string
+
 	// Get imageName and imageID that are later requested in container status
-	someNameOfTheImage, imageID, someRepoDigest, imageAnnotations, err := s.getInfoFromImage(userRequestedImage)
-	if err != nil {
-		return nil, err
+	if isRuntimePullImage {
```
haircommander, line 181: it's not clear to me why this bool toggles these functions.

haircommander, line 181: why are we using the artifact functions here?

haircommander: ah @saschagrunert suggested this. it feels weird to me to use the artifact functions here, why did we go this route?

haircommander:
@mtrmac would you be able to PTAL

server/container_create_linux.go
littlejawa: If the image is not there, we need a way to access the remote repo's metadata to get all needed info, without pulling the full image.
In a way or another, I need to access the data without pulling the layers. The containers/image API did not allow me to do that (or I missed it?) so Sascha suggested this alternative, which allows me to get just the manifest and config that I need.

haircommander: yeah I think the approach is good  but maybe naming the functions differently so it's clearer that's what we're doing here?

internal/storage/runtime.go
```diff
@@ -187,21 +188,31 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
...
-	if err != nil {
-		return ContainerInfo{}, err
-	}
-	image, err := ref.NewImage(r.ctx, systemContext)
-	if err != nil {
-		return ContainerInfo{}, err
-	}
-	defer image.Close()
+	var imageConfig *v1.Image
+	if isRuntimePullImage {
```
mtrmac, line 192: I don’t know why this parameter needs to exist, and I’m generally suspicious of call trees where a function deep down behaves _significantly_ differently based on an option (and I do consider reading local files vs. reaching out to a network significant); can’t the code be refactored to
- read `imageConfig` earlier
- possibly turn the reading of `imageConfig` into a stand-alone helper,
- then move the read of `imageConfig` to the callers?

Callers could then pass the `imageConfig` directly instead of the `isRuntimePullImage` flag. That’s not any more complex, at least.

internal/storage/runtime.go
```diff
@@ -187,21 +188,31 @@ func (r *runtimeService) createContainerOrPodSandbox(systemContext *types.System
...
+		var err error
+		artifact := ociartifact.New()
 
-	imageConfig, err := image.OCIConfig(r.ctx)
-	if err != nil {
-		return ContainerInfo{}, err
-	}
+		imageConfig, err = artifact.GetConfig(r.ctx, metadata.ImageName, &ociartifact.PullOptions{})
```
mtrmac, line 196: Where do credentials for registry access come from? Can this work without them, only relying on the node-global cluster pull secret ??!

server/container_create_linux.go
```diff
@@ -161,40 +164,30 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
-	if id := s.StorageImageServer().HeuristicallyTryResolvingStringAsIDPrefix(userRequestedImage); id != nil {
-		imgResult, err = s.StorageImageServer().ImageStatusByID(s.config.SystemContext, *id)
+	if isRuntimePullImage {
+		someNameOfTheImage, imageID, someRepoDigest, imageAnnotations, err = s.getInfoFromArtifact(ctx, userRequestedImage)
 		if err != nil {
 			return nil, err
 		}
 	} else {
-		potentialMatches, err := s.StorageImageServer().CandidatesForPotentiallyShortImageName(s.config.SystemContext, userRequestedImage)
+		someNameOfTheImage, imageID, someRepoDigest, imageAnnotations, err = s.getInfoFromImage(userRequestedImage)
```
mtrmac, line 187: “get info” says ~nothing useful. `resolveUserRequestedToLocalImageFrom…`?

server/container_create_linux.go
```diff
@@ -161,40 +164,30 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	var imageID storage.StorageImageID
+	var someRepoDigest string
+	var imageAnnotations map[string]string
+
 	// Get imageName and imageID that are later requested in container status
-	var imgResult *storage.ImageResult
-	if id := s.StorageImageServer().HeuristicallyTryResolvingStringAsIDPrefix(userRequestedImage); id != nil {
-		imgResult, err = s.StorageImageServer().ImageStatusByID(s.config.SystemContext, *id)
+	if isRuntimePullImage {
+		someNameOfTheImage, imageID, someRepoDigest, imageAnnotations, err = s.getInfoFromArtifact(ctx, userRequestedImage)
```
mtrmac, line 182: There’s nothing obvious connecting `RuntimePullImage` (whatever that means) and “from artifact”.

I think there probably is some opportunity to clarify the thinking, and naming — what are user wishes, what are the properties that CRI-O cares about, and what are internal implementation details?

(Compare elsewhere about “artifacts don’t have configs”)

server/container_create_linux.go
```diff
@@ -858,6 +852,77 @@ func setContainerConfigSecurityContext(containerConfig *types.ContainerConfig) {
...
+	someRepoDigest = ""
+	if len(imgResult.RepoDigests) > 0 {
+		someRepoDigest = imgResult.RepoDigests[0]
+	}
+	return imgResult.SomeNameOfThisImage, imgResult.ID, someRepoDigest, imgResult.Annotations, nil
+}
+
+func (s *Server) getInfoFromArtifact(ctx context.Context, userRequestedImage string) (imageName *references.RegistryImageReference, imageID storage.StorageImageID, someRepoDigest string, imageAnnotations map[string]string, err error) {
+	artifact := ociartifact.New()
+	pullOpts := &ociartifact.PullOptions{}
```
mtrmac, line 896: Where do credentials come from?

server/container_create_linux.go
```diff
@@ -858,6 +852,77 @@ func setContainerConfigSecurityContext(containerConfig *types.ContainerConfig) {
...
+
+	var imgManifest manifest.Manifest
+	imgManifest, err = artifact.GetManifest(ctx, userRequestedImage, pullOpts)
+	if err != nil {
+		return nil, storage.StorageImageID{}, "", nil, err
+	}
+
+	var name references.RegistryImageReference
+	name, err = references.ParseRegistryImageReferenceFromOutOfProcessData(userRequestedImage)
```
mtrmac, line 905: Why do “container images” support image IDs, image ID prefixes, and short name lookups, and “artifacts” (however that differs from “container images”) don’t?

Are the users OK with that?

server/container_create_linux.go
```diff
@@ -858,6 +852,77 @@ func setContainerConfigSecurityContext(containerConfig *types.ContainerConfig) {
...
+	if err != nil {
+		return nil, storage.StorageImageID{}, "", nil, err
+	}
+
+	var diffIDs []digest.Digest
+	var strID string
+	strID, err = imgManifest.ImageID(diffIDs)
```
mtrmac, line 912: Doing this with empty `diffIDs` is not correct for schema1 images. Maybe that’s not relevant here, but it is worth a comment justifying that with a reason.

server/container_create_linux.go
```diff
@@ -858,6 +852,77 @@ func setContainerConfigSecurityContext(containerConfig *types.ContainerConfig) {
...
+	if err != nil {
+		return nil, storage.StorageImageID{}, "", nil, err
+	}
+
+	var ID storage.StorageImageID
+	ID, err = storage.ParseStorageImageIDFromOutOfProcessData(strID)
```
mtrmac, line 918: - The assumption that local image ID is the same as the one from `Manifest.ImageID` is, to an extent, an implementation decision of c/storage which is not _obviously_ sustainable long-term. Consider sha256 vs. sha512 digests, for example. (And, also, the recent zstd:chunked image ID drama.) Is this introducing a _new_ place where that assumption is going to be hard-coded? If so, are there reasonable alternatives where we would not do that? What are the _actual properties_ the VM backend maintains?
- `StorageImageID` is inherently a reference to local c/storage. If this is not going to use any image in c/storage… why is it using the same data type? Shouldn’t the difference in image locations, or whatever this is, be explicitly modeled throughout the codebase, as separate fields with separate types, or as a properly abstracted interface with clearly documented properties (and non-properties)?

server/container_create_linux.go
```diff
@@ -858,6 +852,77 @@ func setContainerConfigSecurityContext(containerConfig *types.ContainerConfig) {
...
+	if err != nil {
+		return nil, storage.StorageImageID{}, "", nil, err
+	}
+
+	return &name, ID, imgManifest.ConfigInfo().Digest.String(), imgManifest.ConfigInfo().Annotations, nil
```
mtrmac, line 923: `RepoDigest` is not the config digest.

server/image_pull.go
```diff
@@ -20,10 +21,43 @@ import (
...
+	if cfg == nil || cfg.Metadata == nil {
+		return "", nil
+	}
+	name := strings.Join([]string{
```
mtrmac, line 34: Is there any way to avoid a FOURTH copy of the same hard-coded convention?

server/image_pull.go
```diff
@@ -49,6 +83,17 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 		}
 	}
 
+	rhName, rh := s.GetRuntimeHandlerForPod(ctx, req.SandboxConfig)
+	if rh != nil {
+		if rh.RuntimePullImage {
+			// don't pull the image in crio, as it will be done by the runtime
```
mtrmac, line 89: What happens if the image does not exist at all?

What happens if the network is flaky? This effectively blinds Kubelet’s visibility into pulls, and breaks its retry logic.

The credentials provided by Kubelet are discarded.

server/image_pull.go
```diff
@@ -49,6 +83,17 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
...
+			log.Debugf(ctx, "Skip image pull for runtime %s - image %s", rhName, image)
+			return &types.PullImageResponse{
+				ImageRef: req.Image.Image,
```
mtrmac, line 92: Nowadays, the returned `ImageRef` is always a digested reference. If this code path starts putting less structured user-controlled data into the field, I don’t immediately know what that means. E.g. does `ImageStatus` accept these values?

mtrmac, line 92: At least per lines just above, `req.Image` can be `nil`. And the code has already dealt with that, and the line above uses the result; i.e. the debug log and the actually-returned value are not consistent.

internal/config/ociartifact/ociartifact.go
```diff
@@ -164,6 +143,130 @@ func (o *OCIArtifact) Pull(ctx context.Context, img string, opts *PullOptions) (
...
+
+	parsedManifest, err := o.getParsedManifest(ctx, src, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return parsedManifest, nil
+}
+
+func (o *OCIArtifact) GetConfig(ctx context.Context, img string, opts *PullOptions) (*v1.Image, error) {
```
mtrmac, line 167: _In general_, artifacts don’t have image configs. If they have image configs, they are images. (I can see a semantic argument that “image is an artifact”, but… what are we doing here?)

mtrmac (commented):
I’m afraid I don’t understand the full problem space.

Still, this seems closer to a naive prototype, with _a lot_ of divergence from the current CRI-O behavior.
- Doing the pull outside of `PullImage` means Pod’s image pull secrets will not be available. I think that’s simply not viable, especially for the security-paranoid customers this feature might attract. So if the VM API requires the pull to happen only later in CreateContainer, it seems to me that the VM API needs to be redesigned _now_, before this could gain (any more) users.
- Users, sadly (and the more “enterprise” they are, the worse), rely on short names, and tags. These things must be resolved _once_ into a stable image identifier (what `PullImage` returns as an `ImageRef`). This PR does not implement the (regrettable) unqualified search at all, and it is adding code paths with tag-change races all over the place.
- I have no idea what it means to have `StorageImageID` values that no longer refer to local storage, _with nothing added to support that_. I suspect that things break all over the codebase (for starters, the `IsRunningImageAllowed` call will simply fail — and that code path is used by default on all RHEL and OpenShift systems). 
  - E.g. what happens with `ImageStatus` (when Kubelet asks, checking whether it needs to pull at all)? With `ListImages` and Kubelet-driven GC?
  - Overall I find this PR very unlikely to be correct. If this PR were a complete reimplementation of `ImageServer` for the VM backend, I don’t know, my eyes would probably glaze over but I would think that it’s _possible_ that the rest of CRI-O can work with no changes. Just injecting `StorageImageID` values that don’t refer to anything… I can’t believe that works.

server/image_pull.go
mtrmac: Also, Kubelet can gather credentials for several sources and will _loop_ over them in `PullImage` until it finds credentials accepted by the registry.

I don’t particularly _like_ that behavior but that’s one more reason for `PullImage` to be supported properly, whatever that means (a RPC to the VM?).

mtrmac:
> Still, this seems closer to a naive prototype

I realize that, at least with #7471, many parts of that are already true:
- There is a race between CRI-O’s and the VM’s pulls of tagged images, and short-name resolution
- The VM is pulling images in CreateContainer, and without access to the Pod’s pull secret and with no Kubelet visibility
- CRI-O has ~no visibility into VM’s use of the images

But, well, it’s been over a year since that PR. If it is not going to be handled now, is it _ever_ going to be handled?

server/container_create_linux.go
saschagrunert: Yeah we can split the existing bits if required. :+1:

littlejawa:
Hey @mtrmac ,
Thank you for your review.

Let me try to answer some of your questions:

> There is a race between CRI-O’s and the VM’s pulls of tagged images, and short-name resolution

Name resolution needs to happen on the VM side, where the pull occurs.
I have to double-check if/how we do it, but we definitely need to take care of short-names.
Thanks for pointing that out.

> The VM is pulling images in CreateContainer, and without access to the Pod’s pull secret and with no Kubelet visibility

That is actually done on purpose.

The problem we're trying to address with Confidential Containers is when customers want to run a sensitive workload on a cluster they don't trust.
Typically: using public clouds, to benefit from specific hardware without compromising security.
In this approach we can't take any secret or image coming from the host, because by definition they could be compromised.
And we have to prevent the host from accessing the content of the VM in any way.

The current approach in CoCo is to provision the image pull secret inside the VM before image pull. This secret along with any other configs like the container policy, signature verification keys, decryption keys are provisioned after verification of the  VM environment. Post that the image pull is executed.
In other words: nothing is taken from the host, and nothing gets pulled within the VM before we have verified that we're running in a trusted environment.

> CRI-O has ~no visibility into VM’s use of the images

My feeling is that, from CRI-O's perspective, there is just a container being created.
If the creation fails (because a pull error occurred within the VM for instance), the runtime will report it to CRI-O so that the user can see it and address it.
For everything else, we actually want to keep CRI-O un-aware of what happens within the VM.

> I have no idea what it means to have StorageImageID values that no longer refer to local storage, with nothing added to support that. 

Me neither :)
I don't know enough of the internals of cri-o to understand all the implications of not pulling the image.
That's why I asked for help since the beginning (see https://github.com/cri-o/cri-o/issues/8261).



Some questions from me about that (also to @haircommander):
I wanted to make CRI-O ignore the image, but the StorageID (and others things) are required all through the code base, so it seems like I actually need to make something up. Now you're telling me that's wrong, and I don't disagree.
Is there a way to have CRI-O ignore all these?
Rather than building a fake ID, can we return and use something that clearly state we're not dealing with a local image?
One of the things I tried in early versions was to trick CRI-O into using the pause image's storage as the local reference - not as a solution, more like a PoC. Should we have something similar, like an "empty image", that we could reference?

@mtrmac and @saschagrunert mentioned the ImageServer. I can make one used exclusively for the Confidential use case, but wouldn't it still need to output some references to (fake) storage IDs? I didn't go that route because it didn't seem to fix my problem, but I'm willing to go that way if it helps.

Again, any suggestion is welcome.

mtrmac:
> > There is a race between CRI-O’s and the VM’s pulls of tagged images, and short-name resolution
> 
> Name resolution needs to happen on the VM side, where the pull occurs.

See a parallel conversation; I think CRI-O and the VM _must_ structurally be setup to so that they agree on the identity of the image which is used.

> > The VM is pulling images in CreateContainer, and without access to the Pod’s pull secret and with no Kubelet visibility
> 
> That is actually done on purpose.
> 
> The problem we're trying to address with Confidential Containers is when customers want to run a sensitive workload on a cluster they don't trust. Typically: using public clouds, to benefit from specific hardware without compromising security. In this approach we can't take any secret or image coming from the host, because by definition they could be compromised. And we have to prevent the host from accessing the content of the VM in any way.

I have only a very vague idea of how confidential containers work, so this might be completely wrong: I thought confidential containers are confidential by encrypting the _contents_ of the image; in this threat model, the encrypted contents of the image are ~public enough (e.g. it must be possible to copy the container image to the cloud where it is to be run). Is it really the case that the _registry credentials_ must be invisible to CRI-O = that the registry is a trusted component in the confidential container design?

> The current approach in CoCo is to provision the image pull secret inside the VM before image pull. … signature verification keys …are provisioned after verification of the VM environment.

And all of that happens invisible to CRI-O? That… complicates maintenance a lot. How is feature parity going to be maintained over time if there is a parallel implementation invisible to the CRI-O codebase (and drive-by contributors ~like me not being aware of it at all)?

> > CRI-O has ~no visibility into VM’s use of the images
> 
> My feeling is that, from CRI-O's perspective, there is just a container being created.

Kubelet is still going to report on node’s disk usage, and it is still going to run its image garbage collection loop, isn’t it? And possibly try to evict / make scheduling decisions based on node’s disk usage. AFAICS right now that is all based on the filesystem hosting the node’s c/storage store, with no idea that the VMs might be elsewhere, possibly unaware that the node’s filesystem is critically overloaded.

> Some questions from me about that (also to @haircommander): I wanted to make CRI-O ignore the image, but the StorageID (and others things) are required all through the code base

As a first guess, I think all of that code ~assumes that images exist locally in c/storage. If that is not going to be the case, that code needs to either be, function by function, modified to handle in-VM images, or analyzed to make an argument that CRI-O does not actually need to do anything because $reason. Hypothetically it might be the case that everything is in the latter category and we can just fake values, but I think that’s very unlikely.

I’d expect at least `ImageServer`, or some modified version of it (one that does not expose the raw `.GetStore()`, and instead provides more high-level operations) would probably need to have a separate implementation for in-VM images. (And that would show up in the after-CRI-O restart container reload code, in snapshot/restore, in the disk usage management.)

IOW, if the design is that we have to very different storage backends, I’d expect that to be directly reflected / modeled in a CRI-O codebase. It can’t be one or two `ifs` and the rest of the codebase being completely unaware.

And if the same node could contain both c/storage images and in-VM images simultaneously… that would also need to be explicitly modeled.

Some good news is that `StorageImageID` is not an external API; Kubelet uses that values, but IIRC it just uses it as an opaque string. So it _should_ be possible to virtualize it / to model the two storage mechanisms, even if the VM storage had a rather different image reference mechanism. But, also, … the bad news that some structural assumptions (images can be pulled and then run; images can be listed; images don’t need to be pulled if they already exist on the node) _are_ hard-coded in the shape of the CRI.

mtrmac:
> Kubelet uses the provided values

Also, some of the APIs / code paths are ambiguous WRT whether they accept registry references or storage IDs. In some cases Kubelet does not actually _care_, to the extent that it just obtains the value from one CRI API and provides it to another ({`PullImage`,`ImageStatus`} -> `CreateContainer`) … but other users like security software really want to see repo@digest values.

haircommander:
> I’d expect at least ImageServer, or some modified version of it (one that does not expose the raw .GetStore(), and instead provides more high-level operations) would probably need to have a separate implementation for in-VM images. (And that would show up in the after-CRI-O restart container reload code, in snapshot/restore, in the disk usage management.)

yeah I agree I think any place we use storageID we'll probably want to evaluate how to delegate the understanding of that image to the VM.

is it possible to have the vm tell CRI-O what the storage ID is? then, cri-o can populate that everywhere. IT also could use the Pinned CRI field to make sure the kubelet doesn't consider it for garbage collection, while still reporting it so securityy scanners and the like can identify them. or is that too much information passed up to cri-o @littlejawa

littlejawa:
> is it possible to have the vm tell CRI-O what the storage ID is? 

@haircommander - We don't have a way in our current API to report that information.
But in any case, I'm afraid we would have a chicken-and-egg problem: the Storage ID is required in CRI-O during the CreateContainer phase before calling the kata runtime, but the image is pulled in the VM when it receives the CreateContainer request from the runtime. At the time we need it, the StorageID doesn't exist, even in the VM.

Now reading [mtrmak comment](https://github.com/cri-o/cri-o/pull/8008/files#r1937991354)
```
StorageImageID is inherently a reference to local c/storage.
If this is not going to use any image in c/storage… why is it using the same data type?
Shouldn’t the difference in image locations, or whatever this is, be explicitly modeled
throughout the codebase, as separate fields with separate types, or as a properly
abstracted interface with clearly documented properties (and non-properties)?
```

I've tried to copy the same data type, assuming this was a hard requirement. But maybe that's just wrong?
It brings a lot of assumption about its content that are just not true with our model.

If we make an abstraction at the ImageServer level, can we then manage another kind of ID that would be VM-specific?
I don't think we would need to get it from the VM itself, but I'll need a better understanding of what Storage ID actually refers to.

If the goal is to have a 1:1 relationship, making sure an ID refers to a uniquely identifiable image, maybe we can find a way to build this ID based on what CRI-O know about the pod, the runtime process that manages the VM, etc...
Something that clearly points to a single VM and a single image in it - knowing that the VM deals with a single image per container anyway, because every time the image is changed, the pod is restarted, and the VM with it.
If the ID doesn't provide data about the way the image is stored within the VM, but allows to point to that image in a unique way, is that enough?

mtrmac:
I think this needs to start with “what are the immutable parts”. Then we figure out what the behavior should be. And _last_ what the internal structure of the code to implement that behavior should be. (In principle, this is Open Source, and _everything_ can be changed.). But, let’s say that the CRI, and most of Kubelet’s behavior are immutable.

IIRC, Kubelet, primarily:
- Issues pulls: ImageStatus(userInput) = “not found”; PullImage(userInput) = imageRef; createContainer(imageRef)
- Avoids pulls: ImageStatus(userInput) = imageID; createContainer(imageID)
- Has a garbage-collection controller (ListImages, some way to determine what is used, DeleteImage) — I only vaguely remember this part.
- Reports some things in Node / Pod status objects.

That needs to be carefully re-analyzed, but let’s start with that as a hypothesis.

Now, Kubelet has fairly strict expectations on the format of userInput; for imageID, IIRC it doesn’t care ~at all — _but_ the values show up in K8s resources and are user-visible, so it’s possible that various users’ custom scripts expect things.

So, what _should_ the above map to with Confidential Containers?

---

> I've tried to copy the same data type, assuming this was a hard requirement. But maybe that's just wrong?

`StorageImageID` exists (as a specific strict type) because everything around is strings, and it was confusing (and _frequently confused_) what the acceptable syntax for various string fields is in various pieces of the codebase.

It is CRI-O’s ~choice to use the c/storage.Image.ID value as imageID in the CRI API; it plausibly could do something else (but, see above about user-visible values and custom scripts). If Confidential Containers don’t have c/storage, they obviously don’t have c/storage.Image.ID. Still, it _might_ turn out to be necessary to produce exactly the same values (typically = config digest without `sha256:`)… I don’t know, that’s a K8s question.

My (fairly strong) preference is that the CRI-O codebase should retain the strong type separation between various string values. So there should be some “image ID” type, distinct from “user input” or “repo@digest reference” and others. That might be a single type with a single syntax, or maybe a holder wrapping a interface, or something else, I don’t currently have an opinion — that follows from choosing and designing behavior.

littlejawa:
For the records - I'm still looking into that, but have been pre-empted by other tasks.
I'm trying to figure out the right answer for mtrmac's questions, but there is no obvious mapping, as we don't have an API for ImageStatus/ListImage/PullImage.

Now we have an implementation of Confidential Containers on the containerd side that works with encrypted containers.
So somehow, we've managed to make containerd happy with not pulling the image, and still return valid information to Kubernetes.
I want to dig into that implementation and figure out what IDs/structure is returned by Confidential Containers in the context of containerd, it may help find the right answer for CRI-O.
I'm not familiar with the code though, so I need to find time to do that investigation.

littlejawa:
Superseded by #9907

PR #8022, opened 2024-04-16
Lower verbosity of `Allowed annotations are specified for workload` message

saschagrunert:
#### What type of PR is this?

/kind cleanup

It's a bit annoying to have that a generic warning because the configuration is set intentionally. Having it still as an `Info` log should preserve it's major visibility.

None
cc @haircommander 

```release-note
Set the verbosity of the `Allowed annotations are specified for workload` message to `info` (instead of `warn`).
```


haircommander:
/lgtm

yeah it should have been this way from the start...

PR #8054, opened 2024-04-22
Filter namespace creation args to clone in default seccomp policy

dgl:
#### What type of PR is this?

/kind bug

The default seccomp policy used by cri-o did not previously block `clone()` with `CLONE_NEWNS` and other namespace creation flags. This PR blocks those flags. In addition it has to disable `clone3` as it is not currently possible to filter a struct argument with seccomp, this matches the approach taken by containerd.

I can't actually get the bats test to pass in my environment, but then I don't see some of the existing seccomp tests pass, so not sure what's happening there.

```release-note
The default seccomp policy now blocks clone system calls that create a Linux namespace, this provides defense-in-depth against Linux kernel vulnerabilities and matches the default seccomp policy provided by containerd.
```

### DIFF
--- a/internal/config/seccomp/seccomp.go
+++ b/internal/config/seccomp/seccomp.go
@@ -19,6 +19,7 @@ import (
 	json "github.com/json-iterator/go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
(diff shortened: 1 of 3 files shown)


test/clone-ns.c
```diff
@@ -0,0 +1,17 @@
+#define _GNU_SOURCE
```
haircommander, line 1: TESTDATA is `test/testdata` , I think that's a better fit for this anyway and moving it there will fix the test as is I think

PR #8056, opened 2024-04-22
Replace patch-release.go with release.go

olad5:
#### What type of PR is this?

/kind cleanup

- This PR removes the  scripts/release/release.go in favor of scripts/patch-release/patch-release.go to run automated patch releases

No

```release-note
None

```


saschagrunert (approved):
/ok-to-test

saschagrunert:
> I maintain that removing this file means we lose the git history on it. I think it would be better if we used `release.go` instead of `patch-release.go` going forward so we preserve the git history for `release.go`. If other maintainers disagree then this is the last I'll mention the point :)

Right, then let's do it that way @olad5

saschagrunert (approved):

haircommander:
/lgtm

thanks!

kwilczynski:
/hold

The actual commit message has to be updated to reflect change in direction. Also, please, no semantic commits, it hasn't been a standard here.

saschagrunert (approved):

PR #7832, opened 2024-03-04
Add configurable container minimum memory limit

roman-kiselenko:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Provide a way to configure container minimum memory limit per OCI runtime.

#### Which issue(s) this PR fixes:

Fixes #7694

#### Special notes for your reviewer:

None

#### Does this PR introduce a user-facing change?

```release-note
Add configurable container minimum memory limit that can be set per OCI runtime.
```

PR #8057, opened 2024-04-22
Stop using BytesSize when merging the config

bitoku:
#### What type of PR is this?

/kind bug

When you provide `--runtime` option (or `CONTAINER_RUNTIMES` variable) with no `container_min_memory`, it doesn't respect the default container_min_memory (12MiB) and it will be 0B. 

```
sudo crio --log-level=debug --runtimes=runc:/usr/local/sbin/runc:

...
INFO[2024-04-22 12:17:08.039694857Z] Using runtime handler runc version 1.1.12, commit: v1.1.12-0-g51d5e946, spec: 1.0.2-dev, go: go1.20.13, libseccomp: 2.5.4  file="config/config.go:1291"
DEBU[2024-04-22 12:17:08.039769758Z] Runtime handler "runc" container minimum memory set to 0 bytes  file="config/config.go:1300"
...
```

It should be 12 MiB.

None

```release-note
None
```

### DIFF
--- a/contrib/test/ci/vars.yml
+++ b/contrib/test/ci/vars.yml
@@ -150,6 +150,11 @@ kata_skip_image_tests:
   - 'test "image pull and list by manifest list tag"'
   - 'test "image pull and list by manifest list and individual digest"'
   - 'test "image pull and list by individual and manifest list digest"'
+  - 'test "run container with memory_limit_in_bytes -1"'
+  - 'test "run container with memory_limit_in_bytes 12.5MiB"'
+  - 'test "run container with container_min_memory 17.5MiB"'
+  - 'test "run container with container_min_memory 5.5MiB"'
+  - 'test "run container with empty container_min_memory"'
 kata_skip_namespaces_tests:
   - 'test "pid namespace mode target test"'
 kata_skip_network_tests:
--- a/test/image.bats
+++ b/test/image.bats
@@ -380,12 +380,12 @@ runtime_path = "$RUNTIME_BINARY_PATH"
 EOF
 	start_crio
 
-	case $(go env GOARCH) in
-	amd64)
+	case $ARCH in
+	x86_64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		;;
-	arm64)
+	aarch64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		;;
@@ -410,12 +410,12 @@ container_min_memory = "7.5MiB"
 EOF
 	start_crio
 
-	case $(go env GOARCH) in
-	amd64)
+	case $ARCH in
+	x86_64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		;;
-	arm64)
+	aarch64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		;;
@@ -427,7 +427,7 @@ EOF
 		| .image.image = $image' \
 		"$TESTDATA"/container_config.json > "$TESTDIR"/memory.json
 
-	run crictl run "$TESTDIR"/memory.json "$TESTDATA"/sandbox_config.json
+	crictl run "$TESTDIR"/memory.json "$TESTDATA"/sandbox_config.json
 }
 
 @test "run container with container_min_memory 17.5MiB" {
@@ -440,12 +440,12 @@ container_min_memory = "17.5MiB"
 EOF
 	start_crio
 
-	case $(go env GOARCH) in
-	amd64)
+	case $ARCH in
+	x86_64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		;;
-	arm64)
+	aarch64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		;;
@@ -457,7 +457,7 @@ EOF
 		| .image.image = $image' \
 		"$TESTDATA"/container_config.json > "$TESTDIR"/memory.json
 
-	run crictl run "$TESTDIR"/memory.json "$TESTDATA"/sandbox_config.json
+	run ! crictl run "$TESTDIR"/memory.json "$TESTDATA"/sandbox_config.json
 }
 
 @test "run container with container_min_memory 5.5MiB" {
@@ -470,12 +470,12 @@ container_min_memory = "5.5MiB"
 EOF
 	start_crio
 
-	case $(go env GOARCH) in
-	amd64)
+	case $ARCH in
+	x86_64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}
 		;;
-	arm64)
+	aarch64)
 		crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		IMAGE=${IMAGE_LIST_DIGEST_FOR_TAG_ARM64}
 		;;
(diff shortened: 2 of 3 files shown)


saschagrunert (commented):
@roman-kiselenko PTAL

I don't think we need a release note for this @bitoku because we have not released the feature yet. Thank you for the PR!

test/image.bats
```diff
@@ -516,3 +516,32 @@ EOF
...
+
+@test "run container having memory limit below default 12 MiB with empty container_min_memory" {
+	cat << EOF > "$CRIO_CONFIG_DIR/99-mem.conf"
+[crio.runtime]
+default_runtime = "mem"
+[crio.runtime.runtimes.mem]
+runtime_path = "$RUNTIME_BINARY_PATH"
+EOF
+	start_crio
+
```
roman-kiselenko, line 528: Can we check if the log file contains the line about the `container_min_memory` default value?
Something like:
```sh
wait_for_log "Runtime handler \"runc\" container minimum memory set to 12582912 bytes"
```

This can supplement the test with the necessary verification. 🤔

roman-kiselenko (approved):
Nothing is blocking it, just small test improvements.

test/image.bats
```diff
@@ -516,3 +516,32 @@ EOF
...
+		;;
+	esac
+
+	jq --arg image "$IMAGE" '.metadata.name = "memory"
+		| .command = ["/bin/sh", "-c", "sleep 600"]
+    | .linux.resources.memory_limit_in_bytes = 12582911
+		| .image.image = $image' \
+		"$TESTDATA"/container_config.json > "$TESTDIR"/memory.json
+
+	run ! crictl run "$TESTDIR"/memory.json "$TESTDATA"/sandbox_config.json
```
haircommander, line 546: `run` here won't report whether the command failed or not, it'll save it in variables. I recommend dropping `run` if we expect this command to fail

bitoku: according to the doc https://bats-core.readthedocs.io/en/stable/writing-tests.html#when-not-to-use-run,

> Note: In contrast to the above, testing that a command failed is best done via
`run ! command args ...`

However I think just checking the log line is enough as @/roman-kiselenko said.

I'm going to remove this test and add log check to the existing test.

test/image.bats
```diff
@@ -457,7 +457,7 @@ EOF
 		| .image.image = $image' \
 		"$TESTDATA"/container_config.json > "$TESTDIR"/memory.json
 
-	run crictl run "$TESTDIR"/memory.json "$TESTDATA"/sandbox_config.json
+	run ! crictl run "$TESTDIR"/memory.json "$TESTDATA"/sandbox_config.json
```
roman-kiselenko, line 460: I think there is a typo or something; this test is supposed to run [a container and doesn't fail.](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8057/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1782433161666367488/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt#:~:text=alpine%40sha256%3A65b3a80ebe7471beecbc090c5b2cdd0aafeaefa0715f8f12e40dc918a3a70e32-,%23%20f0b894c92ead47bdeffc3e64066d8631b22bfa7edc679156b8843283acf82d6f,-%23%20%23%20%2D%2D%2D%20crio.log%20%3A%3A%20%2D%2D%2D%0A%23%20time) 

`container_min_memory` is `"17.5MiB"` and it is enough to launch a container.

bitoku: container_min_memory is a minimum memory so the container memory should be above the minimum, right?
https://github.com/cri-o/cri-o/blob/9bac613f5fd24c3d690dfb2233f600d157e94baa/internal/config/cgmgr/cgmgr_linux.go#L149-L155

It sets 12MiB as a container memory so it should be failed, shouldn't it?

https://github.com/cri-o/cri-o/blob/c65ecf08ccd32b90e4fcf97b8f5c46418904c506/test/image.bats#L456

roman-kiselenko: This function does the check here:
https://github.com/cri-o/cri-o/blob/d90763d245d0756977cf4db0f9b8c0bbf2796f8f/server/container_create_linux.go#L434-L436 

- `memoryLimit` - `resources.memory_limit_in_bytes` memory requested for container
- `minMemory`   - `container_min_memory` minimum memory needs to run a container

```go
func VerifyMemoryIsEnough(memoryLimit /* 12MiB */, minMemory /* 17MiB */) error { 
  if 12MiB < 17MiB {
    return error
  }
  // ...
}
```

Hence, if `memoryLimit` is above `minMemory`, the container can't start.

roman-kiselenko: @cri-o/cri-o-maintainers ^^^ WDYT?

bitoku: sorry I can't still get it.
12 MiB < 17 MiB is true, so `VerifyMemoryIsEnough` should return an error and the container can't start...?

the same as https://github.com/cri-o/cri-o/pull/7832/files#diff-6391e62911e38278fada73fcfa52f4480e77cb2aea66483934da7e66871dddb5

bitoku:
It seems crio doesn't respect the default runtime `mem` set in conf because of this.
The test fails because container_min_memory in `containerd-shim-kata-v2` is default (12 MiB).

https://github.com/cri-o/cri-o/blob/08c7f32cdc0fa121738154fd8bf05195dd97e1a0/contrib/test/ci/vars.yml#L81

bitoku:
Can we skip the tests for kata in the same way as the ones written in the vars.yml?

test/image.bats
roman-kiselenko: Oh, I see, you're right `run ! ...` is the right assertion here.

roman-kiselenko (approved):
/approve

haircommander:
hm
```
# (from function `crictl' in file ./helpers.bash, line 87,
#  in test file ./image.bats, line 503)
#   `crictl pull ${IMAGE_LIST_DIGEST_FOR_TAG_AMD64}' failed
# time="2024-04-22 16:42:21.070657374Z" level=info msg="Starting CRI-O, version: 1.30.0, git: 837db7e253aac215e7f3d6b49665567dbe80d71d(clean)"
# time="2024-04-22 16:42:21.120060555Z" level=info msg="Starting CRI-O, version: 1.30.0, git: 837db7e253aac215e7f3d6b49665567dbe80d71d(clean)"
# time="2024-04-22T16:42:21Z" level=fatal msg="validate service connection: validate CRI v1 runtime API for endpoint \"unix:///tmp/tmp.zsmL7eIvKa/crio.sock\": rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /tmp/tmp.zsmL7eIvKa/crio.sock: connect: no such file or directory\""
# E0422 16:43:22.568155   67612 remote_image.go:180] "PullImage from image service failed" err="rpc error: code = Unknown desc = can't talk to a V1 container registry" image="quay.io/crio/alpine@sha256:65b3a80ebe7471beecbc090c5b2cdd0aafeaefa0715f8f12e40dc918a3a70e32"
# time="2024-04-22T16:43:22Z" level=fatal msg="pulling image: can't talk to a V1 container registry"
```
/hold

kwilczynski:
Some of the integration tests fail on Arm64 with:

```console
/home/runner/actions-runner/_work/cri-o/cri-o/test/./image.bats: line 501: go: command not found
```

Go not installed? Not in `PATH`?

bitoku:
There hasn't been Go cli in the arm64 runner all along, but because crictl was wrapped with `run` which ignored the failure, the tests didn't fail.
I changed `go env GOARCH` to use `ARCH`.

contrib/test/ci/vars.yml
```diff
@@ -150,6 +150,11 @@ kata_skip_image_tests:
   - 'test "image pull and list by manifest list tag"'
   - 'test "image pull and list by manifest list and individual digest"'
   - 'test "image pull and list by individual and manifest list digest"'
+  - 'test "run container with memory_limit_in_bytes -1"'
+  - 'test "run container with memory_limit_in_bytes 12.5MiB"'
+  - 'test "run container with container_min_memory 17.5MiB"'
```
roman-kiselenko, line 155: Can we skip only the failing test? WDYT?

kwilczynski: Which one is it? Why only a single one would fail, but not all of them for Kata? :thinking:

roman-kiselenko: AFAIK only this test fails `run container with container_min_memory 17.5MiB"`

bitoku: For kata, the runtime settings written in the test cases are not used, but the kata runtime is used instead. (the settings are specified in vars.yml)

In the settings for kata, it doesn't set `container_min_memory`  explicitly so it will be 12MiB (default).

In the 17.5 MiB test, normally it is expected to be failed because the default runtime   `container_min_memory` is supposed to have 17.5 MiB but the container's memory limit is 12MiB (12MiB < 17.5 MiB).
However for kata, kata's `container_min_memory` is 12MiB and it succeeds in creating the container, which caused the test failure.

In my opinion, the other tests just happen to have the same result in both non-kata and kata, so it should be removed from the kata tests.
Also it's confusing because we might think the runtime set in the test cases are used although they are not used.

PR #8011, opened 2024-04-12
server: allow containers within a cluster to opt out of FIPS mode

sohankunkerkar:
With [this](https://github.com/opencontainers/runc/pull/4246) change, we can disable FIPS mode in containers for a cluster that is FIPS-enabled.

#### What type of PR is this?

/kind feature

#### Special notes for your reviewer:

This isn't fully tested; I'm relying on our CI, which can build `runc` with Go 1.21

#### Does this PR introduce a user-facing change?

```release-note
Allow containers within a Kubernetes cluster to opt out of FIPS mode when required
```

PR #8077, opened 2024-04-26
Install `cri-tools` from `master`

saschagrunert:
#### What type of PR is this?

/kind ci

Use their corresponding main branches for testing instead of the pinned version.

None
None

```release-note
None
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -46,8 +46,6 @@ dependencies:
   - name: cri-tools
     version: v1.30.0
     refPaths:
-    - path: scripts/versions
-      match: cri-tools
     - path: contrib/test/ci/vars.yml
       match: cri_tools_git_version
 
--- a/scripts/github-actions-setup
+++ b/scripts/github-actions-setup
@@ -58,15 +58,29 @@ prepare_system() {
     sudo systemctl restart systemd-journald
 }
 
-install_bats() {
-    git clone https://github.com/bats-core/bats-core
-    pushd bats-core
-    git checkout "${VERSIONS["bats"]}"
-    sudo ./install.sh /usr/local
+install_from_github_org_repo() {
+    URL=https://github.com/$1/$2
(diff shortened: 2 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -9,7 +9,7 @@ on:
       - update-nixpkgs-*
   pull_request:
 env:
-  GO_VERSION: '1.22'
+  GO_VERSION: '1.21'
```
sohankunkerkar, line 12: I think you can still keep the Go version as 1.22 and build runc with 1.21.9 (https://github.com/cri-o/cri-o/pull/8011/files#diff-2864e5b0ec49a60b63556bdbf2ffa9c257ebae8656bb9d9ebbbf718d3eb40e8cR111) unless we decided to change the version to 1.21

haircommander: I thought runc main branch is failing to build with any go 1.22 right now

sohankunkerkar: Yeah, that's why we need to use the same hack as the one used in CI: https://github.com/cri-o/cri-o/pull/8011/files#diff-2864e5b0ec49a60b63556bdbf2ffa9c257ebae8656bb9d9ebbbf718d3eb40e8cR109-R129

saschagrunert: Reverted as suggested.

dependencies.yaml
```diff
@@ -57,12 +55,6 @@ dependencies:
     - path: scripts/versions
       match: buildah
 
-  - name: runc
```
sohankunkerkar, line 60: I would retain this logic (similar to CI). There may be instances where we need to stick to a specific version due to regressions in runc

saschagrunert: Alright, reverted that change.

sohankunkerkar (approved):
/lgtm

PR #8095, opened 2024-04-30
Add debug log to expose details of an exec PID lifecycle

kwilczynski:
#### What type of PR is this?

/kind cleanup

The following Pull Request introduced the concept of internally tracking the PID of the process created using the exec (whether synchronous or not) calls. This is very useful and immediately stops CRI-O from losing track of the processes created as part of the exec invocation.

However, aside from a debug log line added for each exec request, nothing else can be used to track a particular exec invocation and termination.

Thus, add a debug log line to expose details of an exec PID being newly started and to be killed, which can aid with troubleshooting.

Related:

- https://github.com/cri-o/cri-o/pull/7937

None

None

```release-note
None
```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -807,6 +807,8 @@ func (c *Container) KillExecPIDs() {
 			if shouldKill {
 				sig = syscall.SIGKILL
 			}
+
+			logrus.Debugf("Stopping exec PID %d for container %s with signal %s ...", pid, c.ID(), unix.SignalName(sig))
 			if err := syscall.Kill(pid, sig); err != nil && !errors.Is(err, syscall.ESRCH) {
 				unkilled[pid] = shouldKill
 			}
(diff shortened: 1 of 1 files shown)


kwilczynski:
Now that I think of it, perhaps it would be useful to add debug log for when an exec call spawns a process? What do you think, @haircommander, would it be useful?

kwilczynski:
The new debug log line looks as follows:

```console
DEBU[2024-04-30 19:04:00.489499405Z] Stopping exec PID 8124 for container a5cc9d55c32f72992f1e484bd32faac4eb5263d09a0c4d715f0e6abb6283b1ec with signal SIGINT ...  file="oci/container.go:811"
```

haircommander:
/approve

LGTM

sohankunkerkar:
IIUC, is this primarily a cleanup effort rather than a new feature? Because every feature typically requires a release note, right?

kwilczynski:
> IIUC, is this primarily a cleanup effort rather than a new feature? Because every feature typically requires a release note, right?

Could be a clean-up. I suppose, the debug log was missing originally? I will update the kind accordingly.

sohankunkerkar (approved):
/lgtm

saschagrunert (approved):

kwilczynski:
OK. Requires a manual cherry-pick.

PR #8102, opened 2024-05-01
Replace go env GOARCH with ARCH environment variable

bitoku:
#### What type of PR is this?

/kind cleanup
/kind ci

The `go` command is not installed in the `arm64` runner. This hasn't been a problem because the test doesn't fail because of a failure in command substitutions. However, this gives us the wrong idea that we can use the `go` command generally, which is, in fact, not always available.

I found this when working on https://github.com/cri-o/cri-o/pull/8057#issuecomment-2073973560.

None

None

```release-note
None
```

### DIFF
--- a/test/image.bats
+++ b/test/image.bats
@@ -170,8 +170,8 @@ function teardown() {
 	[ "$output" != "" ]
 	[[ "$output" == *"RepoDigests: ${IMAGE_LIST_DIGEST}"* ]]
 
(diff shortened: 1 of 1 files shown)


haircommander:
@bitoku thanks! you may need to rebase to pickup test fixes

bitoku:
I think I was able to reproduce the test failure in my env. will take a look.

PR #8110, opened 2024-05-02
Topic/docs completions config updates for runtime pull image

fidencio:

The `runtime_pull_image` configuration option was introduced as part of fff6493e456, but missed documentation, completion, and adding it as part of the configuration template.

None

```release-note
Add missing information about the `runtime_pull_image` to the documentation, completion, and default crio configuration.
```


littlejawa:
Hey @fidencio,
Thanks for filling a hole here :-)

Now I have a concern: this is not just documenting "runtime_pull_image" and adding it to the template config, it also adds it to the "--runtimes" parameter, which I didn't do in the original PR.

The parsing for it is not done, so if you do it, you need to change the following to actually parse the new parameter within the string : https://github.com/cri-o/cri-o/blob/d5666af4172c38c6ed56b40eddfb9af11edb7c62/internal/criocli/criocli.go#L124-L169


And then also https://github.com/cri-o/cri-o/blob/d5666af4172c38c6ed56b40eddfb9af11edb7c62/internal/criocli/criocli.go#L654 to keep things consistant.

And then the crio CI scripts using it should be adapted in the same PR, otherwise some of them might (silently) fail.
See:
- https://github.com/cri-o/cri-o/blob/d5666af4172c38c6ed56b40eddfb9af11edb7c62/contrib/test/ci/vars.yml#L79
- https://github.com/cri-o/cri-o/blob/d5666af4172c38c6ed56b40eddfb9af11edb7c62/test/helpers.bash#L162


Now I have another concern: changing the "--runtimes" string the way you do, the order of parameters is modified. It means that all users of this parameter will see their configuration broken next time they update crio.
Or worse: it will silently work with wrong parameters.
If we add something in this string, I'm afraid it needs to be at the end of it, for backward compatibility...

completions/fish/crio.fish
```diff
@@ -143,7 +143,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l read-only -d 'Setup all un
 complete -c crio -n '__fish_crio_no_subcommand' -f -l registry -r -d 'Registry to be prepended when pulling unqualified images. Can be specified multiple times.'
 complete -c crio -n '__fish_crio_no_subcommand' -l root -s r -r -d 'The CRI-O root directory.'
 complete -c crio -n '__fish_crio_no_subcommand' -l runroot -r -d 'The CRI-O state directory.'
-complete -c crio -n '__fish_crio_no_subcommand' -f -l runtimes -r -d 'OCI runtimes, format is \'runtime_name:runtime_path:runtime_root:runtime_type:privileged_without_host_devices:runtime_config_path:container_min_memory\'.'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l runtimes -r -d 'OCI runtimes, format is \'runtime_name:runtime_path:runtime_root:runtime_type:runtime_pull_image:privileged_without_host_devices:runtime_config_path:container_min_memory\'.'
```
haircommander, line 146: it doesn't look like this was wired into the cli flag so I don't think this is accurate

contrib/test/ci/vars.yml
```diff
@@ -88,6 +88,7 @@ kata_int_test_env:
     PRIVILEGED_WITHOUT_HOST_DEVICES: true
     RUNTIME_CONFIG_PATH: "/opt/kata/share/defaults/kata-containers/configuration.toml"
     PATH: "/opt/kata/bin:{{ ansible_env.PATH }}"
+    RUNTIME_PULL_IMAGE: true
```
littlejawa, line 91: Actually, I think we should keep it "false" here (or not set it at all as it is the default)
Our testing is done on kata alone - not peer-pods nor CoCo. We don't want to enable that feature systematically.

littlejawa:
There is still one occurrence of "usage" string not updated :
https://github.com/cri-o/cri-o/blob/d5666af4172c38c6ed56b40eddfb9af11edb7c62/internal/criocli/criocli.go#L654

Otherwise lgtm

Thanks!

PR #8124, opened 2024-05-04
Wait for configuration reload completion to stabilize e2e tests

bitoku:
Fixes https://github.com/cri-o/cri-o/issues/8074

#### What type of PR is this?

/kind ci
/kind flake

This PR makes some "reload config" tests to wait for the completion of reload.
Currently, those tests in `reload_config.bats` don't wait for `PinnedImagesList` updated.

- reload config should update 'pinned_images'
- reload config should update 'pinned_images' and only 'pause_image' is pinned
- reload config should update 'pause_image' and it becomes 'pinned_images'

It causes a race condition and test failures.
Specifically, The test sends SIGHUP and runs `crictl images`.
It expects the results to be updated with the reloaded new config.
However when `crictl images` runs before `PinnedImagesList` is updated, the `regexForPinnedImages` are not updated, and the image which should be pinned won't be regarded as pinned.

https://github.com/cri-o/cri-o/blob/7a172352193ed137dd284ef7d68545d6d8c4d2cc/internal/storage/image.go#L904-L921

https://github.com/cri-o/cri-o/blob/7a172352193ed137dd284ef7d68545d6d8c4d2cc/internal/storage/image.go#L307-L312

example: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8102/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1786581177407115264/artifacts/fedora-integration/cri-o-gather/artifacts/testout.txt

None

```release-note
None
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -583,6 +583,7 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 			// ImageServer compiles the list with regex for both
 			// pinned and sandbox/pause images, we need to update them
 			s.StorageImageServer().UpdatePinnedImagesList(append(s.config.PinnedImages, s.config.PauseImage))
+			logrus.Info("Configuration reload completed")
 		}
 	}()
 
--- a/contrib/test/ci/integration.yml
+++ b/contrib/test/ci/integration.yml
@@ -54,7 +54,6 @@
         state: present
       loop: "{{ ['cgroups.bats'] | product(kata_skip_cgroups_tests) \
         + ['command.bats'] | product(kata_skip_command_tests) \
-        + ['reload_config.bats'] | product(kata_skip_reload_config_tests) \
         + ['config.bats'] | product(kata_skip_config_tests) \
         + ['config_migrate.bats'] | product(kata_skip_config_migrate_tests) \
         + ['crio-wipe.bats'] | product(kata_skip_crio_wipe_tests) \
--- a/contrib/test/ci/vars.yml
+++ b/contrib/test/ci/vars.yml
@@ -99,10 +99,6 @@ kata_skip_cgroups_tests:
 kata_skip_command_tests:
   - 'test "crio commands"'
   - 'test "log max boundary testing"'
-kata_skip_reload_config_tests:
-  - "test \"reload config should update 'pinned_images'\""
-  - "test \"reload config should update 'pinned_images' and only 'pause_image' is pinned\""
-  - "test \"reload config should update 'pause_image' and it becomes 'pinned_images'\""
 kata_skip_config_tests:
   - 'test "choose different default runtime should succeed"'
   - 'test "runc not existing when default_runtime changed should succeed"'
--- a/test/reload_config.bats
+++ b/test/reload_config.bats
@@ -243,6 +243,7 @@ EOF
 	reload_crio
 	# image becomes pinned
 	expect_log_success $OPTION $EXAMPLE_IMAGE
+	wait_for_log "Configuration reload completed"
 	output=$(crictl images -o json | jq ".images[] | select(.repoTags[] == \"$EXAMPLE_IMAGE\") |.pinned")
 	[ "$output" == "true" ]
 }
@@ -251,6 +252,7 @@ EOF
 	OPTION="pause_image"
 	printf '[crio.image]\npinned_images = [""]\n' > "$CRIO_CONFIG_DIR"/00-default
 	reload_crio
+	wait_for_log "Configuration reload completed"
 	output=$(crictl images -o json | jq '.images[] | select(.pinned == true) | .repoTags[]')
 	# only pause image is pinned
 	[[ "$output" == *"pause"* ]]
@@ -262,6 +264,7 @@ EOF
 	printf '[crio.image]\npinned_images = [""]\npause_image = "%s"\n' $EXAMPLE_IMAGE > "$CRIO_CONFIG_DIR"/04-overwrite
 	reload_crio
 	expect_log_success $OPTION $EXAMPLE_IMAGE
+	wait_for_log "Configuration reload completed"
 	output=$(crictl images -o json | jq '.images[] | select(.pinned == true) | .repoTags[]')
 	# pause image is pinned
 	[[ "$output" == *"fedora-crio-ci"* ]]


bitoku:
cc: @roman-kiselenko

server/server.go
```diff
@@ -583,6 +583,7 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 			// ImageServer compiles the list with regex for both
 			// pinned and sandbox/pause images, we need to update them
 			s.StorageImageServer().UpdatePinnedImagesList(append(s.config.PinnedImages, s.config.PauseImage))
+			logrus.Debugf("Reload completed.")
```
kwilczynski, line 586: Might make sense to add what has completed a reload.

```suggestion
			logrus.Debugf("Configuration reload completed")
```

I also wonder whether we should pass Context to the goroutine and use logger that takes context. Not a requirement to solve this here, though.

kwilczynski: Would probably also make sense to make this an `logrus.Info()`, I suppose.

kwilczynski: Also, probably best to put this message at the end of the following function:

https://github.com/cri-o/cri-o/blob/7a172352193ed137dd284ef7d68545d6d8c4d2cc/pkg/config/reload.go#L18-L80

Thoughts?

bitoku: > Also, probably best to put this message at the end of the following function:

I thought that first, but I don't think that reduces flakiness effectively.
The problem is that `crictl images` runs before `UpdatePinnedImagesList` otherwise the tests fail.
It's important to check that `UpdatePinnedImagesList` is done before `crictl images`, so the log message should be after `UpdatePinnedImagesList` (it shouldn't necessarily be the log message though)

roman-kiselenko: `Debugf` - is for formating.
```golang
logrus.Debug("Configuration reload completed")
```

roman-kiselenko: I wonder if this related 🤔 

https://github.com/cri-o/cri-o/issues/8074

bitoku: could be. let me try to enable them again.

kwilczynski:
/hold 

To see if there is a different way to solve this.

server/server.go
sohankunkerkar: > > Also, probably best to put this message at the end of the following function:
> 
> I thought that first, but I don't think that reduces flakiness effectively. The problem is that `crictl images` runs before `UpdatePinnedImagesList` otherwise the tests fail. It's important to check that `UpdatePinnedImagesList` is done before `crictl images`, so the log message should be after `UpdatePinnedImagesList` (although it shouldn't necessarily be a log message if there is any better way)

Yes, either adding a sleep command or waiting for the appropriate message would work in this scenario. I would lean towards the latter, which is also the current approach.

kwilczynski: As much as I have nothing against adding a new log line, which can also be informative and a sensible thing to do, something about this specific case makes me wary of something we are potentially papering over.

For instance, how long does it take to reload CRI-O? I would imagine it does not take that long. Perhaps it should be even immediate (to avoid saying "instantaneous" here). But maybe we do have some bottleneck somewhere? 

Even without the new log line, adding something like `sleep 5` would be a bit odd. Do we really take that long to reload the current configuration? Especially during the test, CRI-O does nothing upon being started, so it's not busy servicing anything else.

I might also be unnecessarily worried about this. :smile:

bitoku: According to the [failed test](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8102/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1786581177407115264/artifacts/fedora-integration/cri-o-gather/artifacts/testout.txt) the other day
```
# time="2024-05-04 02:38:43.970683870Z" level=info msg="Reloading configuration" file="config/reload.go:19"

# time="2024-05-04 02:38:43.974799347Z" level=info msg="Set config pinned_images to \"quay.io/crio/fedora-crio-ci:latest\"" file="config/reload.go:86"

# time="2024-05-04 02:38:44.032783194Z" level=debug msg="Request: &ImageFsInfoRequest{}" file="otel-collector/interceptors.go:62" id=486737d3-9fb6-4f29-a149-abe6be4167dd name=/runtime.v1.ImageService/ImageFsInfo

# time="2024-05-04 02:38:44.061444616Z" level=debug msg="Runtime handler \"runc\" supports Recursive Read-only (RRO) mounts" file="config/config.go:1340"
```
It took ~90 ms until the last log output.

The tests wait for `Set config pinned_images` before running `crictl`. From that point, it took ~80 ms.
https://github.com/cri-o/cri-o/blob/c3bfcd3b4cbe8b364a9c69c149e966d61cb732b6/test/reload_config.bats#L245

I tested in my env too:
```
# time="2024-05-07 14:06:27.923797979Z" level=info msg="Reloading configuration" file="config/reload.go:19"

# time="2024-05-07 14:06:27.932525811Z" level=info msg="Set config pinned_images to \"quay.io/crio/fedora-crio-ci:latest\"" file="config/reload.go:86"

# time="2024-05-07 14:06:28.005166710Z" level=debug msg="Runtime handler \"runc\" supports Recursive Read-only (RRO) mounts" file="config/config.go:1340"
# time="2024-05-07 14:06:28.005531316Z" level=info msg="Configuration reload completed" file="server/server.go:586"
```
It took ~80 ms.

It runs runtime command in the validation step, which seems taking some time (~40-50 ms).
https://github.com/cri-o/cri-o/blob/67b59ab39eeafea94891b476bb3f1f4479bbd614/pkg/config/config.go#L1298

I still think that it needs to wait even if it's short enough to finish the reload, because it's not guaranteed that the test will finish by `crictl` run.

bitoku: I would incline to put a log message rather than `sleep` because it also doesn't guarantee the reload finish (in the most case it should work though).

roman-kiselenko:
Since the `kata` tests have passed, I'm going to update https://github.com/cri-o/cri-o/issues/8074 after the merge.

server/server.go
kwilczynski: @bitoku, thank you for looking into this! Appreciated.

PR #8141, opened 2024-05-07
Refactor sandbox creation function arguments

saschagrunert:
#### What type of PR is this?

/kind cleanup

- Add a dedicated list of setters for the various pod sandbox fields and use them directly on sandbox creation
- Add a new method to convert a runtime spec to a sandbox and use it on restore.
- Enable the revive linter and set it to a minimum amount of arguments.

None
None

```release-note
None
```


internal/lib/sandbox/sandbox.go
```diff
@@ -320,6 +393,111 @@ func (s *Sandbox) RemoveContainer(ctx context.Context, c *oci.Container) {
 	s.containers.Delete(c.Name())
 }
 
+// SetName can be used to set the name of the sandbox.
+func (s *Sandbox) SetID(id string) {
```
haircommander, line 397: do you think it makes sense to use `internal/factory/sandbox` package for these? My thought is it can be the translation package between a spec or a cri request, and define all of these setters. then, the sandbox object we pass around doesn't have these setters we only need on initialization. thoughts?

saschagrunert: We can utilize `SetNameAndID` and set them from there, I'll change it as suggested.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

saschagrunert:
@cri-o/cri-o-maintainers PTAL

haircommander:
I'm thinking about this more, and I find myself wondering if the `internal/factory` packages were actually a mistake. Part of me wants to move these functions into that package, but doing so would either cause us to do what you're doing here where we ultimately have setters inside of the sandbox package, or it would cause us to keep the problem as-is (a large number of args in sandbox.New()), or it could cause us to overuse an intermediate struct (copying all of the values from server/ into a SandboxBuilder object, and then passing that object ultimately to the sandbox package.

Maybe instead we should really move `internal/factory` packages into `oci` and `sandbox` packages , and create a new file that is `factory.go` in each, where we define these setters/translation functions) 

Ultimately, we want a way to translate CRI spec (and server configuration) to sandbox/container instance, as well as a way to restore each of those from a file

WDYT @cri-o/cri-o-maintainers? Maybe we discuss this at next week's community meeting?

saschagrunert:
> Maybe instead we should really move `internal/factory` packages into `oci` and `sandbox` packages , and create a new file that is `factory.go` in each, where we define these setters/translation functions)

Is the goal to have an immutable sandbox after its creation? Then a dedicated builder would help, which is probably a complete replacement for the `factory`. 

This PR also clashes with https://github.com/cri-o/cri-o/pull/7859

/hold

haircommander:
> Is the goal to have an immutable sandbox after its creation? Then a dedicated builder would help, which is probably a complete replacement for the factory.

I mean it could be partially mutable, but I don't think we need this level of flexibility. there are definitely fields that will never change and we shouldn't expose the ability to change externally.

PR #8148, opened 2024-05-07
Set an integration test timeout

saschagrunert:
#### What type of PR is this?

/kind ci

The integration test suite should not run longer.

None
None

```release-note
None
```

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -70,12 +70,14 @@ jobs:
             critest: 1
             userns: 0
             jobs: 1
+            timeout: 10
           - name: critest / conmon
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -70,70 +70,81 @@ jobs:
             critest: 1
             userns: 0
             jobs: 1
+            timeout: 10
```
kwilczynski, line 73: Why `10` and not `120` or `60`?

saschagrunert: critest usually runs about 5minutes, so I gave them 10.

PR #8149, opened 2024-05-07
pinns: write sysctls in correct process when userns

haircommander:
#### What type of PR is this?

/kind bug

```release-note
fix a bug where pinns wasn't setting the sysctls at the correct time when it was also pinning a user namespace 
```


internal/config/nsmgr/nsmgr_linux.go
```diff
@@ -132,6 +132,10 @@ func (mgr *NamespaceManager) NewPodNamespaces(cfg *PodNamespacesConfig) ([]Names
 		return nil, fmt.Errorf("failed to pin namespaces %v: %s %w", cfg.Namespaces, output, err)
 	}
 
+	if len(output) != 0 {
+		logrus.Debugf("Got output from pinns: %s", output)
+	}
```
kwilczynski, line 137: No need to wrap debug log with a check, you probably want to see any value there, if debug level is set.

```suggestion
	logrus.Debugf("Got output from pinns: %v", output)
```

kwilczynski:
@haircommander, there might be a genuine failure in the logs, per:

```console
# E0508 00:02:06.091415   27274 remote_runtime.go:193] "RunPodSandbox from runtime service failed" err=<
# 	rpc error: code = Unknown desc = CreateContainer failed: usernamespace
#
# 	Caused by:
# 	    0: idmapping uid
# 	    1: No idmap has size > 0: unknown
```

giuseppe (approved):
LGTM (for the .c part)

contrib/test/ci/vars.yml
```diff
@@ -169,6 +169,7 @@ kata_skip_network_tests:
   - 'test "Clean up network if pod sandbox gets killed"'
 kata_skip_pod_tests:
   - 'test "pass pod sysctls to runtime"'
+  - 'test "pass pod sysctls to runtime when in userns"'
```
sohankunkerkar, line 172: Should we create an issue for this to track the failure?

haircommander: we could though hostUsers isn't super meaningful in a kata context because the IDs are already separated from the host through virtualization

Issue #8083, opened 2024-04-29
crio is crashed during container restore


tonyaw:
### What happened?

1. I'm following this procedure to test container checkpoint/restore with Kubernetes 1.30:
    https://kubernetes.io/blog/2022/12/05/forensic-container-checkpointing-alpha/#restore-checkpointed-container-standalone
2. After I created checkpoint of container started by Kubernetes successfully, I deleted Kubernetes cluster, and try to restore a container via crio.
3. I used following json to create pod sandbox:
    crictl runp pod-config.json
```json
{
    "metadata": {
        "name": "clone-sandbox",
        "namespace": "default",
        "attempt": 1,
        "uid": "hdishd83djaidwnduwk28bcsb"
    },
    "linux": {
    }
}
````
5. Then I used following json to restore container from checkpoint:
    crictl create ef80f5a8c32b7 container-config.json pod-config.json
```json
{
  "metadata": {
      "name": "test"
  },
  "image":{
      "image": "/home/tonyaw/k8s/kubelet/checkpoints/checkpoint-tonyaw-deploy-78fdd8bbf5-5p2zf_default-test-2024-04-29T20:43:15+08:00.tar"
  }
}
```
6. crictl is failed, and crio is crashed:
```
[root@test-host kubelet]# crictl create 96480f6885a14 container-config.json pod-config.json
 E0429 20:50:58.512622   44825 remote_runtime.go:319] "CreateContainer in sandbox from runtime service failed" err="rpc error: code = Dead
 lineExceeded desc = context deadline exceeded" podSandboxID="96480f6885a14"
```

crio journal log and backtrace:
[crio-crash-log.txt](https://github.com/cri-o/cri-o/files/15149831/crio-crash-log.txt)

### What did you expect to happen?

container is restored successfully.

### How can we reproduce it (as minimally and precisely as possible)?

Follow procedure https://kubernetes.io/blog/2022/12/05/forensic-container-checkpointing-alpha/#restore-checkpointed-container-standalone to restore a container with Kubernetes 1.30.
The issue shall happen when you restore the container.

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
WARN[0000] Failed to decode the keys ["storage.options.override_kernel_check"] from "/etc/containers/storage.conf"
crio version 1.30.0
Version:        1.30.0
GitCommit:      a3de11a8aa3070c8b040054461ce82314c95baf2
GitCommitDate:  2024-04-26T16:51:03Z
[...]

PR #8150, opened 2024-05-08
Check for nil values when importing container definition

kwilczynski:
#### What type of PR is this?

/kind bug

The Protocol Buffers specification does not mandate that the Linux Resources for a container definition be present or that any of its attributes be required—most values are left to be the default, often unset, values. This means that the container definition stored as JSON document also does not require nor mandate that the "linux" attribute (with any of its sub-attributes) has to be present, etc. As such, once the JSON document is parsed, any missing sections will be set to `nil` within the `ContainerConfig` type, and attempting to access such object attributes to dereference them will result in a panic.

Thus, add `nil` check to ensure that the `Linux` field of the `ContainerConfig` type has been set before accessing any fields from within it, which would prevent a potential panic.

While at it, add a `nil` and empty value check to verify whether the "image" section and attribute have been set, making it the one and only mandatory field within the JSON document that is the container definition. This also short-circuits error handling further down the execution stack, so we can fail early instead.

Fixes https://github.com/cri-o/cri-o/issues/8083.

None

```release-note
Check for nil values when importing container definition for a given container checkpoint to be restored.
```

### DIFF
--- a/server/container_restore.go
+++ b/server/container_restore.go
@@ -74,7 +74,7 @@ func (s *Server) CRImportCheckpoint(
 
 	var restoreArchivePath string
 	if restoreStorageImageID != nil {
-		log.Debugf(ctx, "Restoring from oci image %s\n", inputImage)
+		log.Debugf(ctx, "Restoring from oci image %s", inputImage)
 
 		// This is not out-of-process, but it is at least out of the CRI-O codebase; containers/storage uses raw strings.
 		mountPoint, err = s.ContainerServer.StorageImageServer().GetStore().MountImage(restoreStorageImageID.IDStringForOutOfProcessConsumptionOnly(), nil, "")
--- a/server/container_restore.go
+++ b/server/container_restore.go
@@ -57,19 +57,24 @@ func (s *Server) CRImportCheckpoint(
 ) (ctrID string, retErr error) {
 	var mountPoint string
 
-	input := createConfig.Image.Image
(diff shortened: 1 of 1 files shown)


server/container_restore.go
```diff
@@ -57,19 +57,24 @@ func (s *Server) CRImportCheckpoint(
...
-	restoreStorageImageID, err := s.checkIfCheckpointOCIImage(ctx, input)
+	restoreStorageImageID, err := s.checkIfCheckpointOCIImage(ctx, inputImage)
 	if err != nil {
 		return "", err
 	}
 
 	var restoreArchivePath string
 	if restoreStorageImageID != nil {
-		log.Debugf(ctx, "Restoring from oci image %s\n", input)
+		log.Debugf(ctx, "Restoring from oci image %s\n", inputImage)
```
saschagrunert, line 77:
```suggestion
		log.Debugf(ctx, "Restoring from oci image %s", inputImage)
```

saschagrunert (commented):
Just a nit, otherwise LGTM. :+1:

server/container_restore.go
kwilczynski: Good clean up! :+1:

kwilczynski: Done.

saschagrunert (approved):

kwilczynski:
OK. Requires manual cherry-pick.

PR #8156, opened 2024-05-08
[release-1.29] pinns: write sysctls in correct process when userns

haircommander:
#### What type of PR is this?

/kind bug

```release-note
fix a bug where pinns wasn't setting the sysctls at the correct time when it was also pinning a user namespace 
```


sohankunkerkar (approved):
/lgtm

PR #8161, opened 2024-05-10
Delete deprecated registries config

bitoku:
#### What type of PR is this?

/kind cleanup
/kind deprecation

Since #4455, `registries` option has been no effect.

https://github.com/cri-o/cri-o/blob/9712c53d2291f0436a75cacecb70ad9a3a823eed/pkg/config/config.go#L765-L770

However, there are some places that don't mention explicitly that the option has been deprecated.
This PR changes the messages which don't mention the deprecation, and removes the remained configuration.

None

Since the deprecation has already been released and there was a release note, I'm not sure if we should need a new release-note for this.
I made it `None` for now.

```release-note
None
```

### DIFF
--- a/contrib/test/ci/setup.yml
+++ b/contrib/test/ci/setup.yml
@@ -71,12 +71,5 @@
       [crio]
       storage_driver = "overlay"
 
-- name: add quay.io and docker.io as default registries
-  copy:
-    dest: /etc/crio/crio.conf.d/01-registries.conf
-    content: |
-      [crio.image]
-      registries = [ "quay.io", "docker.io" ]
-
 - name: build parallel
   include_tasks: "build/parallel.yml"
--- a/internal/criocli/criocli.go
+++ b/internal/criocli/criocli.go
@@ -638,7 +638,7 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 		&cli.StringSliceFlag{
 			Name:    "registry",
 			Value:   cli.NewStringSlice(defConf.Registries...),
-			Usage:   "Registry to be prepended when pulling unqualified images. Can be specified multiple times.",
+			Usage:   "Registry to be prepended when pulling unqualified images. Can be specified multiple times. This option is deprecated, and it has no effect.",
 			EnvVars: []string{"CONTAINER_REGISTRY"},
 		},
 		&cli.StringFlag{
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -543,6 +543,8 @@ type ImageConfig struct {
 	// ImageVolumes controls how volumes specified in image config are handled
 	ImageVolumes ImageVolumesType `toml:"image_volumes"`
 	// Registries holds a list of registries used to pull unqualified images
+	// Deprecated: Support for this option has been dropped, and it has no effect.
+	// Please refer to containers-registries.conf(5) for configuring unqualified-search registries.
 	Registries []string `toml:"registries"`
 	// Temporary directory for big files
 	BigFilesTemporaryDir string `toml:"big_files_temporary_dir"`
--- a/completions/fish/crio.fish
+++ b/completions/fish/crio.fish
@@ -143,7 +143,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l profile-mem -r -d 'Write a
 complete -c crio -n '__fish_crio_no_subcommand' -f -l profile-port -r -d 'Port for the pprof profiler.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l rdt-config-file -r -d 'Path to the RDT configuration file for configuring the resctrl pseudo-filesystem.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l read-only -d 'Setup all unprivileged containers to run as read-only. Automatically mounts the containers\' tmpfs on \'/run\', \'/tmp\' and \'/var/tmp\'.'
-complete -c crio -n '__fish_crio_no_subcommand' -f -l registry -r -d 'Registry to be prepended when pulling unqualified images. Can be specified multiple times.'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l registry -r -d 'Registry to be prepended when pulling unqualified images. Can be specified multiple times. This option is deprecated, and it has no effect.'
 complete -c crio -n '__fish_crio_no_subcommand' -l root -s r -r -d 'The CRI-O root directory.'
 complete -c crio -n '__fish_crio_no_subcommand' -l runroot -r -d 'The CRI-O state directory.'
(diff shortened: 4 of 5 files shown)


internal/criocli/criocli.go
```diff
@@ -638,7 +638,7 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 		&cli.StringSliceFlag{
 			Name:    "registry",
 			Value:   cli.NewStringSlice(defConf.Registries...),
-			Usage:   "Registry to be prepended when pulling unqualified images. Can be specified multiple times.",
+			Usage:   "Registry to be prepended when pulling unqualified images. Can be specified multiple times. This option is deprecated, and it has no effect.",
```
kwilczynski, line 641: Using the same format as the other deprecated command-line switch.

```suggestion
			Usage:   "Registry to be prepended when pulling unqualified images. Can be specified multiple times. DEPRECATED: This option will be removed in the future, and it has no effect.",
```

kwilczynski: ... or maybe not. Other older deprecations didn't follow this sort of loud way.

pkg/config/config.go
```diff
@@ -543,6 +543,8 @@ type ImageConfig struct {
 	// ImageVolumes controls how volumes specified in image config are handled
 	ImageVolumes ImageVolumesType `toml:"image_volumes"`
 	// Registries holds a list of registries used to pull unqualified images
+	// Deprecated: Support for this option has been dropped, and it has no effect.
+	// Please refer to containers-registries.conf(5) for configuring unqualified-search registries.
 	Registries []string `toml:"registries"`
```
kwilczynski, line 548: I wonder if we should just be done with it and remove it already? It's been a while since 2021...

@cri-o/cri-o-maintainers, thoughts on removing this completely?

contrib/test/ci/setup.yml
```diff
@@ -71,12 +71,5 @@
       [crio]
       storage_driver = "overlay"
 
-- name: add quay.io and docker.io as default registries
-  copy:
-    dest: /etc/crio/crio.conf.d/01-registries.conf
-    content: |
-      [crio.image]
-      registries = [ "quay.io", "docker.io" ]
-
```
haircommander, line 80: we do need this config in /etc/containers/registries.conf.d

haircommander:
I would be okay with dropping support already. you're correct in that it's been deprecated for a while, and there is a better replacement with registries.conf

cc @mtrmac

contrib/test/ci/setup.yml
bitoku: Do we?
It hasn't been used since #4455. 
According to the log, it wasn't used.
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8102/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1786581177407115264/artifacts/fedora-integration/cri-o-gather/artifacts/testout.txt
```
# time="2024-05-04T02:38:42Z" level=warning msg="Support for the 'registries' option has been dropped but it is referenced in \"/etc/crio/crio.conf.d/01-registries.conf\".  Please use containers-registries.conf(5) for configuring unqualified-search registries instead."
```

I don't know how we configure the registries.conf in prow CIs, but at least for github actions, there is already `registries.conf`.

https://github.com/cri-o/cri-o/blob/9712c53d2291f0436a75cacecb70ad9a3a823eed/scripts/github-actions-setup#L161

https://github.com/cri-o/cri-o/blob/9712c53d2291f0436a75cacecb70ad9a3a823eed/test/registries.conf#L1-L4

mtrmac:
In general, the better replacement is “don’t do this at all”. `unqualified-search-registries` is invisible to Kubelet and interacts _extremely badly_ with Kubelet’s code to choose a credential to use in the `ImagePull` operation (it will always send a `docker.io` credential for short image names, if one is available to Kubelet, and that credential will be sent to the target registry).

So I’m fully in favor of removing anything which makes this misfeature easy to configure, or indeed, reachable by CRI-O users at all.

**Do not** add any new users of this option, especially not by default. (Yes, https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-container-runtime/_base/files/container-registries.yaml does that currently.)

---

Users who want a layer of indirection so that their Pod definitions don’t change in different deployment sites should define a “canonical” location for each image, which includes a host name (`vendor.example/product/component:version`, and use `registries.d`’s `prefix`/`location` redirection mechanism (in OpenShift, `ImageContentSourcePolicy` / `ImageTagMirrorSet` / `ImageDigestMirrorSet`). That works consistently everywhere, and it _scales_ across many different products/components — unlike short names, where different product might result in conflicting short names and no way to resolve that.

contrib/test/ci/setup.yml
haircommander: yeah since we have completely different install paths for github actions and prow, I think we may still need this.

can you try with it to see if the ci failures are due to this?

haircommander:
great points, which further supports removing this field (so users at least have one less place to be able to incorrectly specify)

It would be cool to drop support for short names generally but given the precedent of following docker's age-old lead, that may be tricky.

contrib/test/ci/setup.yml
haircommander: jk I think failures are not flakes from this https://github.com/cri-o/cri-o/issues/8167

kwilczynski: @bitoku @haircommander, what would be the way forward here?

kwilczynski:
> So is our way to go to drop this feature completely now, not adding more deprecation messages? Or is this PR ok as the first step to drop this feature in the future?

@bitoku, we should drop it completely.

It's been a while since this wasn't functional per: https://github.com/cri-o/cri-o/pull/8161#discussion_r1596530906. I also think that @haircommander has nothing against doing so and that it would be fine to retire this code... provided that we don't break things (like the need to have the configuration file; albeit, you pointed out that CI has passed with the file being removed).

@cri-o/cri-o-maintainers, thoughts about removing this old code?

pkg/config/config.go
saschagrunert: Have we ever deprecated this? If not, then we have to according to https://github.com/cri-o/cri-o/blob/main/deprecating_process.md

kwilczynski: Would this stand in for deprecation?

- https://github.com/cri-o/cri-o/blob/main/pkg/config/config.go#L765-L770

This has been defunct for about two years now.

saschagrunert: Somehow, yes.

kwilczynski: We can:

1. Remove it completely, given that it has been defunct for a while, with a relevant log line being produced each time this option has been enabled/used.

2. Improve messaging about this option being detected in command-line use and documentation, and perhaps even make the log line more explicit.

@cri-o/cri-o-maintainers, thoughts?

saschagrunert:
Yeah I guess we can remove it. :+1:

bitoku:
Thank you all!
Then I will create a new PR and will close this one.

PR #8162, opened 2024-05-10
Kill exec PIDs after main container exited

saschagrunert:
#### What type of PR is this?

/kind bug
/kind regression

Before applying this patch we killed the exec PIDs right away on container stop which leads into the failing e2e test:

```
[sig-node] [NodeFeature:SidecarContainers] Containers Lifecycle should terminate sidecars simultaneously if prestop doesn't exit
```

This regression is now fixed by killing the exec PIDs after the main container as well as in the same thread.

Needs a cherry-pick since the enhancement got already backported into supported release branches.

Fixes https://github.com/kubernetes/kubernetes/issues/124743

Follow-up on https://github.com/cri-o/cri-o/pull/7937

```release-note
Fixed pod lifecycle regression where the exec PID's got killed before the actual container.
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -869,8 +869,6 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 
 	startTime := time.Now()
 
-	go c.KillExecPIDs()
-
 	// Allow for SIGINT to correctly interrupt the stop loop, especially
 	// when CRI-O is run directly in the foreground in the terminal.
 	ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_oci.go
```diff
@@ -955,6 +953,10 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 		}
 	}, bm, true, ctx.Done())
 
+	// Kill the exec PIDS after the main container to avoid pod lifecycle regressions:
```
kwilczynski, line 956:
```suggestion
	// Kill the exec PIDs after the main container to avoid pod lifecycle regressions:
```

internal/oci/runtime_oci.go
```diff
@@ -869,8 +869,6 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 
 	startTime := time.Now()
 
-	go c.KillExecPIDs()
```
saschagrunert, line 872: @haircommander any reason this is a separate go routine?

saschagrunert (commented):
/hold

for @haircommander's review

internal/oci/runtime_oci.go
kwilczynski: Not sure why. However, it probably shouldn't be.

haircommander: I didn't want to block the container stop, but if we do it after it should be fine

haircommander:
thanks for the investigation!

/lgtm

PR #8166, opened 2024-05-13
test: use config file rather than environment variables for the kata CI job

littlejawa:
#### What type of PR is this?

/kind ci

There are a number of integration tests that rely on creating a new runtime class with specific parameters, and make it the default before running some tests.
Theses tests fail with kata, because the kata CI job setup uses environment variables (CONTAINER_DEFAULT_RUNTIME typically), which take precedence in crio's config.

This PR modify the kata job's configuration so that it relies on config files too.

To achieve this, this PR does the following:
- create a new helper function that creates a runtime config file with all needed parameters
- use that function in all tests that need such a feature
- use the same function for configuring the kata runtime when needed

Fixes: #8074

This PR also solves several integration test failures with no tracking issue.

```release-note
None
```


littlejawa:
/hold

This PR contains a commit that we don't want to merge.

littlejawa:
@cri-o/cri-o-maintainers can you have a look?

This one has been in the work for a while.
This is solving several issues for kata, by running the tests with the correct runtime settings.
The trick was to make sure the test's custom runtime configuration was taken into account while still running the container with kata.
I did some tests (both local and here) showing that this is now the case.

PR #8174, opened 2024-05-14
server: support ping_group_range if in a userns

haircommander:
#### What type of PR is this?

/kind bug
ping_group_range sysctl requires the upper bound be within the range of IDs the user has access to. ping_group_range is often set to the max allowable range "0 2147483647", but this will break for every usernamespace pod.

Instead, hack around it by updating the max GID to be the largest one we find in the IDMappings

```release-note
fix a bug where a pod with a userns would fail to be created when `ping_group_range` sysctl was specified for it (and the max of that range was outside of the pods user namespace)
```

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -746,7 +746,7 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 	}
 
 	// Add default sysctls given in crio.conf
-	sysctls := s.configureGeneratorForSysctls(ctx, g, hostNetwork, hostIPC, req.Config.Linux.Sysctls)
+	sysctls := s.configureGeneratorForSysctls(ctx, g, hostNetwork, hostIPC, sandboxIDMappings, req.Config.Linux.Sysctls)
 
 	// set up namespaces
 	s.resourceStore.SetStageForResource(ctx, sbox.Name(), "sandbox namespace creation")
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
LGTM

server/sandbox_run_linux.go
```diff
@@ -1091,7 +1091,43 @@ func (s *Server) configureGeneratorForSysctls(ctx context.Context, g *generate.G
 		g.AddLinuxSysctl(key, value)
 		sysctlsToReturn[key] = value
 	}
-	return sysctlsToReturn
+	return configurePingGroupRangeGivenIDMappings(ctx, g, sandboxIDMappings, sysctlsToReturn)
+}
+
+func configurePingGroupRangeGivenIDMappings(ctx context.Context, g *generate.Generator, sandboxIDMappings *idtools.IDMappings, sysctls map[string]string) map[string]string {
+	// We have to manually fuss with this specific sysctl.
+	// It's commonly set to the max range by default "0 2147483647".
```
kwilczynski, line 1099: **For posterity:**

This value, for the upper limit, is the same as `2^31-1`.

server/sandbox_run_linux.go
```diff
@@ -1091,7 +1091,43 @@ func (s *Server) configureGeneratorForSysctls(ctx context.Context, g *generate.G
...
+		return sysctls
+	}
+
+	maxID := 0
+	for _, mapping := range sandboxIDMappings.GIDs() {
+		topOfRange := mapping.ContainerID + mapping.Size - 1
+		if maxID < topOfRange {
+			maxID = topOfRange
+		}
+	}
```
kwilczynski, line 1124: The recent Go version has the `max()` built-in:

```go
	var maxID int
	for _, mapping := range sandboxIDMappings.GIDs() {
		maxID = max(maxID, mapping.ContainerID + mapping.Size - 1)
	}
```

server/sandbox_run_linux.go
kwilczynski: Also, a fun fact:

- https://en.wikipedia.org/wiki/2,147,483,647

server/sandbox_run_linux.go
haircommander: added a follow up https://github.com/cri-o/cri-o/pull/8184

PR #8187, opened 2024-05-16
Create tag on reconciler

saschagrunert:
#### What type of PR is this?

/kind ci

The tag we try to push is not being created beforehand, which is now fixed when applying this patch.

None
None

```release-note
None
```

### DIFF
--- a/scripts/tag-reconciler/tag-reconciler.go
+++ b/scripts/tag-reconciler/tag-reconciler.go
@@ -65,6 +65,11 @@ func run() error {
 }
 
 func pushTagToRemote(repo *git.Repo, tag, remote string) error {
+	logrus.Infof("Adding tag to repository: %s", tag)
+	if err := repo.Tag(tag, tag); err != nil {
+		return fmt.Errorf("unable to tag repository: %w", err)
+	}
+
 	logrus.Infof("Pushing tag to origin: %s", tag)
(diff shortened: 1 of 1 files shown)


saschagrunert:
cc @olad5

scripts/tag-reconciler/tag-reconciler.go
```diff
@@ -65,6 +65,11 @@ func run() error {
 }
 
 func pushTagToRemote(repo *git.Repo, tag, remote string) error {
+	logrus.Info("Tagging repository")
```
kwilczynski, line 68:
```suggestion
	logrus.Infof("Adding tag to repository: %s", tag)
```

scripts/tag-reconciler/tag-reconciler.go
```diff
@@ -65,6 +65,11 @@ func run() error {
...
+	if err := repo.Tag(tag, tag); err != nil {
+		return fmt.Errorf("tag repository: %w", err)
```
kwilczynski, line 70:
```suggestion
		return fmt.Errorf("unable to tag repository: %w", err)
```

scripts/tag-reconciler/tag-reconciler.go
saschagrunert: Applied as suggested.

scripts/tag-reconciler/tag-reconciler.go
saschagrunert: Applied as suggested.

PR #8203, opened 2024-05-20
Fix memory leakage when port-forwarding

bitoku:
#### What type of PR is this?

/kind bug

This PR forces cri-o to drain the stream when it finishes port-forward in order to avoid failure to close the connection which causes memory leakage.

The issue details are here.
https://issues.redhat.com/browse/OCPBUGS-30978
> while sending multiple failing port-forwarding requests, crio memory usage increases until the node dies

It had been reported before (https://bugzilla.redhat.com/show_bug.cgi?id=1798193) and fixed in https://github.com/cri-o/cri-o/commit/7016c3e13fe19272d6a680ca2172df02da75a35f.
However at some point, it regressed.

Fixes https://issues.redhat.com/browse/OCPBUGS-30978

```release-note
Fix memory leakage when sending a failing port-forward request
```

### DIFF
--- a/server/container_portforward.go
+++ b/server/container_portforward.go
@@ -25,17 +25,18 @@ func (s StreamService) PortForward(ctx context.Context, podSandboxID string, por
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
-	// if we error in this function before Copying all of the content out of the stream,
-	// this stream will eventually get full, which causes leakages and can eventually brick CRI-O
+	// Drain the stream to prevent failure to close the connection and memory leakage.
 	// ref https://bugzilla.redhat.com/show_bug.cgi?id=1798193
-	emptyStreamOnError := true
+	// ref https://issues.redhat.com/browse/OCPBUGS-30978
 	defer func() {
-		if emptyStreamOnError && stream != nil {
-			go func() {
-				_, copyError := pools.Copy(io.Discard, stream)
-				log.Errorf(ctx, "Error closing port forward stream after other error: %v", copyError)
-			}()
+		if stream == nil {
+			return
 		}
+		go func() {
+			if _, err := pools.Copy(io.Discard, stream); err != nil {
+				log.Errorf(ctx, "Unable to drain the stream data: %v", err)
+			}
+		}()
 	}()
 
 	sandboxID, err := s.runtimeServer.PodIDIndex().Get(podSandboxID)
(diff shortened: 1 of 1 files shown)


server/container_portforward.go
```diff
@@ -25,17 +25,19 @@ func (s StreamService) PortForward(ctx context.Context, podSandboxID string, por
...
-				log.Errorf(ctx, "Error closing port forward stream after other error: %v", copyError)
-			}()
+		if stream == nil {
+			return
 		}
+		go func() {
+			_, err := pools.Copy(io.Discard, stream)
+			if err != nil {
+				log.Errorf(ctx, "could not drain the stream with the error: %v", err)
+			}
```
kwilczynski, line 39:
```suggestion
			if _, err := pools.Copy(io.Discard, stream); err != nil {
				log.Errorf(ctx, "Unable to drain the stream data: %v", err)
			}
```

server/container_portforward.go
```diff
@@ -25,17 +25,19 @@ func (s StreamService) PortForward(ctx context.Context, podSandboxID string, por
...
+	// ref https://issues.redhat.com/browse/OCPBUGS-30978
 	defer func() {
-		if emptyStreamOnError && stream != nil {
-			go func() {
-				_, copyError := pools.Copy(io.Discard, stream)
-				log.Errorf(ctx, "Error closing port forward stream after other error: %v", copyError)
-			}()
+		if stream == nil {
+			return
 		}
```
kwilczynski, line 34: If we keep the if statement, but remote the `emptyStreamOnError` from it, then we have the smallest changeset.

haircommander: I like the updated structure

server/container_portforward.go
```diff
@@ -25,17 +25,19 @@ func (s StreamService) PortForward(ctx context.Context, podSandboxID string, por
...
+		go func() {
+			_, err := pools.Copy(io.Discard, stream)
+			if err != nil {
```
haircommander, line 37: nit: this can be compressed to one line

server/container_portforward.go
bitoku: thanks!

server/container_portforward.go
bitoku: If there is no objections about it, I'll keep the update one. thanks!

haircommander:
/approve

thank you!

PR #8212, opened 2024-05-23
Image verification for namespaced policies

harche:
#### What type of PR is this?

/kind feature

Adds support for verification using namespaced policies during container creation. 

None

```release-note
Add support for validating signatures on container creation. Now, if there is a namespaced policy in the `signature_policy_dir`, CRI-O will validate the signature defined in `signature_policy_dir`/`NS`.json for pods in namespace `NS`
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -183,8 +183,8 @@ type ImageServer interface {
 	// - ctx: The context for controlling the function's execution
 	// - systemContext: server's system context for the given namespace
 	// - imageName: A RegistryImageReference that identifies the container image
-	// - localManifestDigest: The digest of the local image
-	IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, imageName RegistryImageReference, localManifestDigest digest.Digest) error
+	// - imageID: A StorageImageID of the image
+	IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, imageName RegistryImageReference, imageID StorageImageID) error
 }
 
 func parseImageNames(image *storage.Image) (someName *RegistryImageReference, tags []reference.NamedTagged, digests []reference.Canonical, err error) {
@@ -474,7 +474,7 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
 	return &result, nil
 }
 
-func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, imageName RegistryImageReference, localManifestDigest digest.Digest) error {
+func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, imageName RegistryImageReference, imageID StorageImageID) error {
 	policy, err := signature.DefaultPolicy(systemContext)
 	if err != nil {
 		return fmt.Errorf("get default policy: %w", err)
@@ -491,7 +491,7 @@ func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContex
 		}
 	}()
 
-	if err := svc.checkSignature(ctx, systemContext, policyContext, imageName, localManifestDigest); err != nil {
+	if err := svc.checkSignature(ctx, systemContext, policyContext, imageName, imageID); err != nil {
 		return fmt.Errorf("checking signature of %q: %w", imageName, err)
 	}
 
@@ -500,23 +500,19 @@ func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContex
 	return nil
 }
 
-func (svc *imageService) checkSignature(ctx context.Context, sys *types.SystemContext, policyContext *signature.PolicyContext, imageName RegistryImageReference, localManifestDigest digest.Digest) error {
+func (svc *imageService) checkSignature(ctx context.Context, sys *types.SystemContext, policyContext *signature.PolicyContext, imageName RegistryImageReference, imageID StorageImageID) error {
 	userIdentityRef, err := docker.NewReference(imageName.Raw())
 	if err != nil {
 		return fmt.Errorf("creating docker:// reference for %q: %w", imageName.Raw().String(), err)
 	}
 
-	storeNamedReference, err := reference.WithDigest(reference.TrimNamed(imageName.Raw()), localManifestDigest)
+	storageRef, err := imageID.imageRef(svc)
 	if err != nil {
-		return fmt.Errorf("creating a digested storage reference for %q + %q: %w", imageName.Raw().String(), localManifestDigest, err)
+		return fmt.Errorf("creating containers-storage: reference for %v: %w", storageRef, err)
 	}
-	storageRefWithoutID, err := istorage.Transport.NewStoreReference(svc.store, storeNamedReference, "")
-	if err != nil {
-		return fmt.Errorf("creating containers-storage: reference for %v: %w", storageRefWithoutID, err)
-	}
-	log.Debugf(ctx, "Created storageRefWithoutID = %q", transports.ImageName(storageRefWithoutID))
+	log.Debugf(ctx, "Created storageRef = %q", transports.ImageName(storageRef))
 
-	storageSource, err := storageRefWithoutID.NewImageSource(ctx, sys)
+	storageSource, err := storageRef.NewImageSource(ctx, sys)
 	if err != nil {
 		return fmt.Errorf("creating image source for local store image: %w", err)
 	}
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -615,9 +615,10 @@ type pullImageArgs struct {
 }
 
 type pullImageOutputItem struct {
-	Progress          *types.ProgressProperties `json:",omitempty"`
-	Result            string                    `json:",omitempty"` // If not "", in the format of transport.ImageName()
-	CanonicalImageRef reference.Canonical       `json:",omitempty"`
+	Progress *types.ProgressProperties `json:",omitempty"`
+	Result   string                    `json:",omitempty"` // If not "", in the format of transport.ImageName()
+	Name     string                    `json:",omitempty"`
+	Digest   digest.Digest             `json:",omitempty"`
 }
 
 func pullImageChild() {
@@ -662,7 +663,8 @@ func pullImageChild() {
 		fmt.Fprintf(os.Stderr, "%v", err)
 		os.Exit(1)
 	}
-	output <- pullImageOutputItem{Result: transports.ImageName(destRef), CanonicalImageRef: canonicalRef}
+
+	output <- pullImageOutputItem{Result: transports.ImageName(destRef), Name: canonicalRef.Name(), Digest: canonicalRef.Digest()}
 
 	close(output)
 	<-outputWritten
@@ -756,9 +758,15 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
 				break
 			}
 
-			if item.CanonicalImageRef != nil {
-				canonicalRef = item.CanonicalImageRef
+			named, err := reference.ParseNamed(item.Name)
+			if err != nil {
+				break
 			}
+			canonicalRef, err = reference.WithDigest(named, item.Digest)
+			if err != nil {
+				break
+			}
+
 			if item.Progress != nil && progress != nil {
 				progress <- *item.Progress
 			}
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -15,7 +15,10 @@ import (
 	"time"
 
 	"github.com/containers/image/v5/copy"
+	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
+	cimage "github.com/containers/image/v5/image"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/shortnames"
 	"github.com/containers/image/v5/signature"
 	istorage "github.com/containers/image/v5/storage"
@@ -30,7 +33,9 @@ import (
 	digest "github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
+	crierrors "k8s.io/cri-api/pkg/errors"
 
+	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/storage/references"
 	"github.com/cri-o/cri-o/pkg/config"
 )
@@ -38,6 +43,11 @@ import (
 const (
 	minimumTruncatedIDLength = 3
 )
+
+//nolint:gochecknoinits // Init function is intentional here.
+func init() {
+	reexec.Register("crio-pull-image", pullImageChild)
+}
 
 // ImageResult wraps a subset of information about an image: its ID, its names,
 // and the size, if known, or nil if it isn't.
@@ -133,7 +143,17 @@ type ImageServer interface {
 	// for further analysis. Call Close() on the resulting image.
 	PrepareImage(systemContext *types.SystemContext, imageName RegistryImageReference) (types.ImageCloser, error)
 	// PullImage imports an image from the specified location.
-	PullImage(ctx context.Context, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, error)
+	//
+	// Arguments:
+	// - ctx: The context for controlling the function's execution
+	// - imageName: A RegistryImageReference representing the image to be pulled
+	// - options: Pointer to ImageCopyOptions, which contains various options for the image copy process
+	//
+	// Returns:
+	// - types.ImageReference: The reference to the pulled image
+	// - reference.Canonical: A canonical reference to the image, typically including the digest
+	// - error: An error object if pulling the image fails, otherwise nil
+	PullImage(ctx context.Context, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, reference.Canonical, error)
 
 	// DeleteImage deletes a storage image (impacting all its tags)
 	DeleteImage(systemContext *types.SystemContext, id StorageImageID) error
@@ -156,6 +176,15 @@ type ImageServer interface {
 
 	// UpdatePinnedImagesList updates pinned and pause images list in imageService.
 	UpdatePinnedImagesList(imageList []string)
+
+	// IsRunningImageAllowed verifies if running of the container image is allowed.
+	//
+	// Arguments:
+	// - ctx: The context for controlling the function's execution
+	// - systemContext: server's system context for the given namespace
+	// - imageName: A RegistryImageReference that identifies the container image
+	// - localManifestDigest: The digest of the local image
+	IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, imageName RegistryImageReference, localManifestDigest digest.Digest) error
 }
 
 func parseImageNames(image *storage.Image) (someName *RegistryImageReference, tags []reference.NamedTagged, digests []reference.Canonical, err error) {
@@ -445,6 +474,92 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
 	return &result, nil
 }
 
+func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, imageName RegistryImageReference, localManifestDigest digest.Digest) error {
+	policy, err := signature.DefaultPolicy(systemContext)
+	if err != nil {
+		return fmt.Errorf("get default policy: %w", err)
+	}
+
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return fmt.Errorf("create policy context: %w", err)
+	}
+
+	defer func() {
+		if err := policyContext.Destroy(); err != nil {
+			log.Errorf(ctx, "Error destroying policy: %+v", err)
+		}
+	}()
+
+	if err := svc.checkSignature(ctx, systemContext, policyContext, imageName, localManifestDigest); err != nil {
+		return fmt.Errorf("checking signature of %q: %w", imageName, err)
+	}
+
+	log.Debugf(ctx, "Is allowed to run config image %s (policy path: %q)", imageName, systemContext.SignaturePolicyPath)
+
+	return nil
+}
+
+func (svc *imageService) checkSignature(ctx context.Context, sys *types.SystemContext, policyContext *signature.PolicyContext, imageName RegistryImageReference, localManifestDigest digest.Digest) error {
+	userIdentityRef, err := docker.NewReference(imageName.Raw())
+	if err != nil {
+		return fmt.Errorf("creating docker:// reference for %q: %w", imageName.Raw().String(), err)
+	}
+
+	storeNamedReference, err := reference.WithDigest(reference.TrimNamed(imageName.Raw()), localManifestDigest)
+	if err != nil {
+		return fmt.Errorf("creating a digested storage reference for %q + %q: %w", imageName.Raw().String(), localManifestDigest, err)
+	}
+	storageRefWithoutID, err := istorage.Transport.NewStoreReference(svc.store, storeNamedReference, "")
+	if err != nil {
+		return fmt.Errorf("creating containers-storage: reference for %v: %w", storageRefWithoutID, err)
+	}
+	log.Debugf(ctx, "Created storageRefWithoutID = %q", transports.ImageName(storageRefWithoutID))
+
+	storageSource, err := storageRefWithoutID.NewImageSource(ctx, sys)
+	if err != nil {
+		return fmt.Errorf("creating image source for local store image: %w", err)
+	}
+	defer storageSource.Close()
+
+	unparsedToplevel := cimage.UnparsedInstance(storageSource, nil)
+	topManifest, topMIMEType, err := unparsedToplevel.Manifest(ctx)
+	if err != nil {
+		return fmt.Errorf("get top level manifest: %w", err)
+	}
+
+	unparsedInstance := unparsedToplevel
+	if manifest.MIMETypeIsMultiImage(topMIMEType) {
+		manifestList, err := manifest.ListFromBlob(topManifest, topMIMEType)
+		if err != nil {
+			return fmt.Errorf("parsing list manifest: %w", err)
+		}
+
+		instanceDigest, err := manifestList.ChooseInstance(sys)
+		if err != nil {
+			return fmt.Errorf("choosing instance: %w", err)
+		}
+
+		unparsedInstance = cimage.UnparsedInstance(storageSource, &instanceDigest)
+	}
+
+	mixedUnparsedInstance := cimage.UnparsedInstanceWithReference(unparsedInstance, userIdentityRef)
+
+	allowed, err := policyContext.IsRunningImageAllowed(ctx, mixedUnparsedInstance)
+	if err != nil {
+		var policyErr signature.PolicyRequirementError
+		if errors.As(err, &policyErr) {
+			return fmt.Errorf("%w: %w", crierrors.ErrSignatureValidationFailed, err)
+		}
+		return fmt.Errorf("verifying signatures: %w", err)
+	}
+	if !allowed {
+		panic("Internal inconsistency: IsRunningImageAllowed returned !allowed and no error when checking image signature")
+	}
+
+	return nil
+}
+
 func imageSize(img types.Image) *uint64 {
 	if sum, err := img.Size(); err == nil {
 		usum := uint64(sum)
@@ -490,10 +605,6 @@ func (svc *imageService) PrepareImage(inputSystemContext *types.SystemContext, i
 	return srcRef.NewImage(svc.ctx, systemContext)
 }
 
-func init() { //nolint
-	reexec.Register("crio-pull-image", pullImageChild)
-}
-
 type pullImageArgs struct {
 	Lookup       *imageLookupService
 	ImageName    string // In the format of RegistryImageReference.StringForOutOfProcessConsumptionOnly()
@@ -504,8 +615,9 @@ type pullImageArgs struct {
 }
 
 type pullImageOutputItem struct {
-	Progress *types.ProgressProperties `json:",omitempty"`
-	Result   string                    `json:",omitempty"` // If not "", in the format of transport.ImageName()
+	Progress          *types.ProgressProperties `json:",omitempty"`
+	Result            string                    `json:",omitempty"` // If not "", in the format of transport.ImageName()
+	CanonicalImageRef reference.Canonical       `json:",omitempty"`
 }
 
 func pullImageChild() {
@@ -545,12 +657,12 @@ func pullImageChild() {
 	}()
 	args.Options.Progress = progress
 
-	destRef, err := pullImageImplementation(context.Background(), args.Lookup, store, imageName, args.Options)
+	destRef, canonicalRef, err := pullImageImplementation(context.Background(), args.Lookup, store, imageName, args.Options)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%v", err)
 		os.Exit(1)
 	}
-	output <- pullImageOutputItem{Result: transports.ImageName(destRef)}
+	output <- pullImageOutputItem{Result: transports.ImageName(destRef), CanonicalImageRef: canonicalRef}
 
 	close(output)
 	<-outputWritten
@@ -577,26 +689,26 @@ func formatPullImageOutputItemGoroutine(dest io.Writer, items <-chan pullImageOu
 	}
 }
 
-func (svc *imageService) pullImageParent(ctx context.Context, imageName RegistryImageReference, parentCgroup string, options *ImageCopyOptions) (types.ImageReference, error) {
+func (svc *imageService) pullImageParent(ctx context.Context, imageName RegistryImageReference, parentCgroup string, options *ImageCopyOptions) (types.ImageReference, reference.Canonical, error) {
 	progress := options.Progress
 	// the first argument imageName is not used by the re-execed command but it is useful for debugging as it
 	// shows in the ps output.
 	cmd := reexec.CommandContext(ctx, "crio-pull-image", imageName.StringForOutOfProcessConsumptionOnly())
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		return nil, fmt.Errorf("error getting stdout pipe for image copy process: %w", err)
+		return nil, nil, fmt.Errorf("error getting stdout pipe for image copy process: %w", err)
 	}
 	defer stdout.Close()
 
 	stderr, err := cmd.StderrPipe()
 	if err != nil {
-		return nil, fmt.Errorf("error getting stderr pipe for image copy process: %w", err)
+		return nil, nil, fmt.Errorf("error getting stderr pipe for image copy process: %w", err)
 	}
 	defer stderr.Close()
 
 	stdin, err := cmd.StdinPipe()
 	if err != nil {
-		return nil, fmt.Errorf("error getting stdin pipe for image copy process: %w", err)
+		return nil, nil, fmt.Errorf("error getting stdin pipe for image copy process: %w", err)
 	}
 
 	stdinArguments := pullImageArgs{
@@ -616,18 +728,19 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
 
 	stdinArguments.Options.Progress = nil
 	if err := cmd.Start(); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	if err := json.NewEncoder(stdin).Encode(&stdinArguments); err != nil {
 		stdin.Close()
 		if waitErr := cmd.Wait(); waitErr != nil {
-			return nil, fmt.Errorf("%w: %w", waitErr, err)
+			return nil, nil, fmt.Errorf("%w: %w", waitErr, err)
 		}
-		return nil, fmt.Errorf("json encode to pipe failed: %w", err)
+		return nil, nil, fmt.Errorf("json encode to pipe failed: %w", err)
 	}
 	stdin.Close()
 
 	resultChan := make(chan string)
+	var canonicalRef reference.Canonical
 	go func() {
 		defer func() {
 			close(resultChan) // Future reads, if any, will get "".
@@ -643,9 +756,13 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
 				break
 			}
 
+			if item.CanonicalImageRef != nil {
+				canonicalRef = item.CanonicalImageRef
+			}
 			if item.Progress != nil && progress != nil {
 				progress <- *item.Progress
 			}
+
 			if item.Result != "" {
 				resultChan <- item.Result
 			}
@@ -657,23 +774,28 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
 	errOutput, errReadAll := io.ReadAll(stderr)
 	if err := cmd.Wait(); err != nil {
 		if errReadAll == nil && len(errOutput) > 0 {
-			return nil, fmt.Errorf("pull image: %s", string(errOutput))
+			return nil, nil, fmt.Errorf("pull image: %s", string(errOutput))
 		}
-		return nil, err
+		return nil, nil, err
 	}
 
 	if result == "" {
-		return nil, errors.New("pull child finished successfully but didn’t send a result")
+		return nil, nil, errors.New("pull child finished successfully but didn’t send a result")
 	}
 
 	destRef, err := alltransports.ParseImageName(result)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	return destRef, nil
+
+	if canonicalRef == nil {
+		return nil, nil, errors.New("canonical image reference cannot be nil")
+	}
+
+	return destRef, canonicalRef, nil
 }
 
-func (svc *imageService) PullImage(ctx context.Context, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, error) {
+func (svc *imageService) PullImage(ctx context.Context, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, reference.Canonical, error) {
 	if options.CgroupPull.UseNewCgroup {
 		return svc.pullImageParent(ctx, imageName, options.CgroupPull.ParentCgroup, options)
 	} else {
@@ -685,26 +807,26 @@ func (svc *imageService) PullImage(ctx context.Context, imageName RegistryImageR
 // NOTE: That means this code can run in a separate process, and it should not access any CRI-O global state.
 //
 // It returns a c/storage ImageReference for the destination.
-func pullImageImplementation(ctx context.Context, lookup *imageLookupService, store storage.Store, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, error) {
+func pullImageImplementation(ctx context.Context, lookup *imageLookupService, store storage.Store, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, reference.Canonical, error) {
 	srcSystemContext, srcRef, err := lookup.prepareReference(options.SourceCtx, imageName)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	destRef, err := istorage.Transport.NewStoreReference(store, imageName.Raw(), "")
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	policy, err := signature.DefaultPolicy(options.SourceCtx)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	policyContext, err := signature.NewPolicyContext(policy)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	_, err = copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
+	manifestBytes, err := copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
 		SourceCtx:        srcSystemContext,
 		DestinationCtx:   options.DestinationCtx,
 		OciDecryptConfig: options.OciDecryptConfig,
@@ -712,9 +834,15 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 		Progress:         options.Progress,
 	})
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	return destRef, err
+
+	canonicalRef, err := reference.WithDigest(reference.TrimNamed(imageName.Raw()), digest.FromBytes(manifestBytes))
+	if err != nil {
+		return nil, nil, fmt.Errorf("create canonical reference: %w", err)
+	}
+
+	return destRef, canonicalRef, err
 }
 
 func (svc *imageService) UntagImage(systemContext *types.SystemContext, name RegistryImageReference) error {
--- a/internal/storage/runtime.go
+++ b/internal/storage/runtime.go
@@ -316,7 +316,7 @@ func (r *runtimeService) CreatePodSandbox(systemContext *types.SystemContext, po
 		if imageAuthFile != "" {
 			sourceCtx.AuthFilePath = imageAuthFile
 		}
-		ref, err = r.storageImageServer.PullImage(context.Background(), pauseImage, &ImageCopyOptions{
+		ref, _, err = r.storageImageServer.PullImage(context.Background(), pauseImage, &ImageCopyOptions{
 			SourceCtx:      &sourceCtx,
 			DestinationCtx: systemContext,
 		})
--- a/internal/storage/runtime_test.go
+++ b/internal/storage/runtime_test.go
@@ -766,7 +766,7 @@ var _ = t.Describe("Runtime", func() {
 				imageServerMock.EXPECT().GetStore().Return(storeMock),
 				mockResolveReference(storeMock, storageTransportMock,
 					"docker.io/library/pauseimagename:latest", "", ""),
-				imageServerMock.EXPECT().PullImage(gomock.Any(), pauseImageRef, expectedCopyOptions).Return(pulledRef, nil),
+				imageServerMock.EXPECT().PullImage(gomock.Any(), pauseImageRef, expectedCopyOptions).Return(pulledRef, nil, nil),
 				mockResolveReference(storeMock, storageTransportMock,
 					"docker.io/library/pauseimagename:latest", "", imageID.IDStringForOutOfProcessConsumptionOnly()),
 				imageServerMock.EXPECT().GetStore().Return(storeMock),
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -238,24 +238,19 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 
 	userSpecifiedImage := ctr.Config().GetImage().UserSpecifiedImage
 
-	if userSpecifiedImage == "" {
-		// Attempt to check if the image is a checkpoint OCI image
-		if _, err := s.checkIfCheckpointOCIImage(ctx, userRequestedImage); err != nil {
-			return nil, fmt.Errorf("failed to check if this is a checkpoint image: %w", err)
-		} else { // Fallback to ensure that userSpecifiedImage is valid
-			userSpecifiedImage = imageName.StringForOutOfProcessConsumptionOnly()
+	// Skip this check on a restore, as it's from a file not from a registry so there's no notion of a registry,
+	// and thus no notion of a named reference.
+	if userSpecifiedImage != "" || !ctr.Restore() {
+		var userImageRef references.RegistryImageReference
+
+		userImageRef, err = references.ParseRegistryImageReferenceFromOutOfProcessData(userSpecifiedImage)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get userImageRef from user specified image %q: %w", userSpecifiedImage, err)
 		}
-	}
 
-	var userImageRef references.RegistryImageReference
-
-	userImageRef, err = references.ParseRegistryImageReferenceFromOutOfProcessData(userSpecifiedImage)
-	if err != nil {
-		return nil, fmt.Errorf("unable to get userImageRef from user specified image %q: %w", userSpecifiedImage, err)
-	}
-
-	if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, userImageRef, imgResult.Digest); err != nil {
-		return nil, err
+		if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, userImageRef, imgResult.Digest); err != nil {
+			return nil, err
+		}
 	}
 
 	labelOptions, err := ctr.SelinuxLabel(sb.ProcessLabel())
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -190,6 +190,10 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		}
 	}
 
+	// userRequestedImage is the way to locate the image.
+	// When called by Kubelet, it is either the ImageRef as returned by PullImage
+	// (for us, always a RegistryImageReference using a repo@digest), or an ImageID as returned by ImageStatus (a full StorageImageID).
+	// We accept other inputs, like short names and digest prefixes, because previous CRI-O versions did, just to be conservative against breakage.
 	userRequestedImage, err := ctr.UserRequestedImage()
 	if err != nil {
 		return nil, err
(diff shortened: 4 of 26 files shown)


internal/storage/image.go
```diff
@@ -420,6 +438,200 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+	}
+
+	var imageParsed reference.Named
+	// Check if the configImage is a local image ID if the image is already pulled
+	if localManifestDigest.String() == "" {
+		img, err := svc.GetStore().Image(configImage)
+		if err != nil && !errors.Is(err, storage.ErrImageUnknown) {
+			return fmt.Errorf("error retrieving image %s: %w", configImage, err)
+		}
+		if img != nil {
```
rphillips, line 534: perhaps double check the length of img here.

harche: @rphillips `img` is of Image type, https://github.com/cri-o/cri-o/blob/a934077689a3ff1dc09a4f4130de60dfabd74e88/vendor/github.com/containers/storage/images.go#L32

Hence the check for if it is `nil`.

internal/storage/image.go
```diff
@@ -420,6 +438,200 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+	policy, err := signature.DefaultPolicy(systemContext)
+	if err != nil {
+		return "", fmt.Errorf("get default policy: %w", err)
+	}
+
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return "", fmt.Errorf("create policy context: %w", err)
+	}
+	defer policyContext.Destroy() //nolint:errcheck
```
saschagrunert, line 457: Let's log the error in case if occurs.

saschagrunert (commented):
The unit test mocks needs to be updated (see GitHub actions).

Is this still work in progress (WIP)?

internal/storage/image.go
rphillips: Right, but the length of the array could be zero, and we are indexing into it.

harche: @rphillips something like,
```
if img != nil && img.Names != nil && len(img.Names) > 0
```
?

internal/storage/image.go
harche: Thanks, updated.

internal/storage/image.go
rphillips: I'm going to retract my comment here... I think the better approach is to check if err != nil if the alternative case, on line 542, and remove the `if img != nil` entirely.

harche: you can have `err == nil` and `img == nil` at the same time. It just means that, there was no error looking up in the image store but the image doesn't exist in that store.

server/image_pull_test.go
```diff
@@ -66,137 +46,14 @@ var _ = t.Describe("ImagePull", func() {
 			Expect(response).NotTo(BeNil())
 		})
 
-		It("should succeed when already pulled", func() {
```
mtrmac, line 69: Outright removing tests makes me worried. Why is that OK? It might well _be OK_, maybe the test is directly redundant, but the PR doesn’t say.

(Based on past experience, I am very skeptical of the mock tests, they tend to trigger very unrealistic code paths without anyone noticing, but still…)

server/container_create.go
```diff
@@ -403,6 +403,15 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 		}, nil
 	}
 
+	systemCtx, err := s.contextForNamespace(req.SandboxConfig.Metadata.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, req.Config.Image.Image, req.Config.Image.UserSpecifiedImage); err != nil {
```
mtrmac, line 411: ⚠️ This is racy and insufficient.

The policy check must happen _exactly_ on the resolved image (`imageID := imgResult.ID` in `container_create_linux.go`. Making a name lookup here is racy WRT concurrent pulls changing what `Image` resolves to.

internal/storage/image.go
```diff
@@ -154,6 +166,12 @@ type ImageServer interface {
 
 	// UpdatePinnedImagesList updates pinned and pause images list in imageService.
 	UpdatePinnedImagesList(imageList []string)
+
+	// PullLocalManifestDigest pull the manifest digest and assembles the image ref from it.
+	PullLocalManifestDigest(ctx context.Context, systemContext *types.SystemContext, image string) (string, error)
```
mtrmac, line 171: What is calling this? I can’t see any user.

internal/storage/image.go
```diff
@@ -154,6 +166,12 @@ type ImageServer interface {
...
+
+	// IsRunningImageAllowed verifies if running of the container image is allowed.
+	IsRunningImageAllowed(context.Context, *types.SystemContext, string, string) error
```
mtrmac, line 174: Document the parameters, in detail.

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
 	return &result, nil
 }
 
+func (svc *imageService) PullLocalManifestDigest(ctx context.Context, systemContext *types.SystemContext, userSpecifiedImage string) (string, error) {
```
mtrmac, line 441: (Ignoring this because I can’t see a caller.)

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+	imageParsed, err := reference.ParseNormalizedNamed(userSpecifiedImage)
```
mtrmac, line 442: (As a matter of general policy, push string parsing as close to the top of the call stack as possible; use well-specified types like `RegistryImageReference` instead of strings. If you _must_ use strings, document the format as precisely as possible.)

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+		if errors.As(err, &policyErr) {
+			return "", fmt.Errorf("%w: %w", crierrors.ErrSignatureValidationFailed, err)
+		}
+
+		return "", fmt.Errorf("pulling %q to %q: %w", transports.ImageName(srcRef), transports.ImageName(destRef), err)
+	}
+	return digest.FromBytes(manifestBytes), nil
+}
+
+func (svc *imageService) IsRunningImageAllowed(ctx context.Context, systemContext *types.SystemContext, configImage, userSpecifiedImage string) error {
```
mtrmac, line 511: Preferably use strict types like `RegistryImageReference` or `StorageImageID` instead of strings. Document string parameter semantics in detail if that’s not possible.

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+
+	defer func() {
+		if err := policyContext.Destroy(); err != nil {
+			log.Errorf(ctx, "Unable to destroy policy: %v", err)
+		}
+	}()
+
+	var localManifestDigest digest.Digest
+	// On fresh pulls retrieve the full image and digest as configImage
+	if parsedConfigImage, err := reference.Parse(configImage); err == nil {
```
mtrmac, line 530: (Push parsing heuristics up the call stack. This is too deep to be ignoring parsing errors. I suspect this will have to change anyway to resolve the races in `container_create.go`.

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+		if canonical, isDigested := parsedConfigImage.(reference.Canonical); isDigested {
+			localManifestDigest = canonical.Digest()
+		}
+	}
+
+	var imageParsed reference.Named
+	// Check if the configImage is a local image ID if the image is already pulled
```
mtrmac, line 537: Heuristics must happen at a higher level.

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+	if localManifestDigest.String() == "" {
+		img, err := svc.GetStore().Image(configImage)
```
mtrmac, line 539: This is a ~heuristic lookup. Don’t do heuristic lookups here.

And if anything, do that through `storageTransport.ResolveReference`, to correctly handle references when several platforms of a multi-platform image are present. Raw `storage.Store.Image` is almost certainly a mistake.

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+		if err != nil && !errors.Is(err, storage.ErrImageUnknown) {
+			return fmt.Errorf("error retrieving image %s: %w", configImage, err)
+		}
+		if img != nil {
+			imageParsed, err = reference.ParseNormalizedNamed(img.Names[0])
```
mtrmac, line 544: ⚠️ 

`img.Names[0]` is a basically arbitrary name, can be assumed to be chosen by an attacker (by triggering a pull of an image which “duplicates” the victim but users the attacker’s name. It must not participate in any policy decision.

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+		if err != nil {
+			return fmt.Errorf("parsing config image %+v : %w", configImage, err)
+		}
+		imageParsed = reference.TagNameOnly(imageParsed)
+	}
+
+	// Tagged references are mainly used by tools like crictl or critest
+	if localManifestDigest.String() == "" {
+		if parsedRef, err := reference.ParseNormalizedNamed(configImage); err == nil {
+			ref, err := svc.pullLocalManifestDigest(ctx, systemContext, policyContext, reference.TagNameOnly(parsedRef))
```
mtrmac, line 563: ⚠️ 

Why is an “is running allowed” predicate triggering a pull? That’s _extremely_ surprising.

(And without credentials it can’t be assumed to succeed, anyway.)

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+		imageParsed, err = reference.ParseNormalizedNamed(configImage)
+		if err != nil {
+			return fmt.Errorf("parsing config image %+v : %w", configImage, err)
+		}
+		imageParsed = reference.TagNameOnly(imageParsed)
+	}
+
+	// Tagged references are mainly used by tools like crictl or critest
+	if localManifestDigest.String() == "" {
+		if parsedRef, err := reference.ParseNormalizedNamed(configImage); err == nil {
```
mtrmac, line 562: (I am generally confused about the intended relationship of `configImage`/`imageParsed` here. Which one is authoritative for what? But I didn’t spend the time to read the code in detail.)

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+	if err != nil {
+		return "", fmt.Errorf("creating docker:// reference for %q: %w", imageParsed.String(), err)
+	}
+
+	destRef, err := istorage.Transport.NewStoreReference(svc.store, imageParsed, "")
+	if err != nil {
+		return "", fmt.Errorf("creating containers-storage: reference for %q: %w", imageParsed.String(), err)
+	}
+
+	manifestBytes, err := copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
```
mtrmac, line 489: On the `IsRunningImageAllowed` caller at least, no credentials available = can’t be relied upon.

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+			}
+			localManifestDigest = ref
+		}
+	}
+
+	if localManifestDigest.String() == "" {
+		return fmt.Errorf("not interpretable config image: %q", configImage)
+	}
+
+	if err := svc.checkSignature(ctx, systemContext, policyContext, imageParsed, localManifestDigest); err != nil {
```
mtrmac, line 575: ⚠️

AFAICS this can trigger verification against an _attacker-chose_ `imageParse` identity.

It’s noticeable that `userSpecifiedImage` is not actually used here at all!

internal/storage/image.go
```diff
@@ -420,6 +438,209 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+		return fmt.Errorf("checking signature of %q: %w", imageParsed.String(), err)
+	}
+
+	log.Debugf(ctx, "Is allowed to run config image %s (policy path: %q)", userSpecifiedImage, systemContext.SignaturePolicyPath)
+
+	return nil
+}
+
+func (svc *imageService) checkSignature(ctx context.Context, sys *types.SystemContext, policyContext *signature.PolicyContext, imageParsed reference.Named, localManifestDigest digest.Digest) error {
+	userIdentityRef, err := docker.NewReference(imageParsed)
```
mtrmac, line 585: Push parsing towards the top of the call stack.

internal/storage/image.go
```diff
@@ -633,23 +849,23 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
...
+		return nil, "", errors.New("pull child finished successfully but didn’t send a result")
 	}
 
 	destRef, err := alltransports.ParseImageName(result)
 	if err != nil {
-		return nil, err
+		return nil, "", err
 	}
-	return destRef, nil
+	return destRef, "", nil
```
mtrmac, line 865: Shouldn’t this return a real value?

internal/storage/image.go
```diff
@@ -131,7 +143,7 @@ type ImageServer interface {
 	// for further analysis. Call Close() on the resulting image.
 	PrepareImage(systemContext *types.SystemContext, imageName RegistryImageReference) (types.ImageCloser, error)
 	// PullImage imports an image from the specified location.
-	PullImage(ctx context.Context, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, error)
+	PullImage(ctx context.Context, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, string, error)
```
mtrmac, line 146: Document semantics of security-relevant return values. Avoid unbounded `string` types if possible.

internal/storage/image.go
```diff
@@ -661,36 +877,48 @@ func (svc *imageService) PullImage(ctx context.Context, imageName RegistryImageR
...
 		OciDecryptConfig: options.OciDecryptConfig,
 		ProgressInterval: options.ProgressInterval,
 		Progress:         options.Progress,
 	})
 	if err != nil {
-		return nil, err
+		return nil, "", err
+	}
+
+	imageParsed, err := reference.ParseNormalizedNamed(imageName.Raw().String())
```
mtrmac, line 910: ??! Why is this round-tripping through text?

internal/storage/image.go
```diff
@@ -661,36 +877,48 @@ func (svc *imageService) PullImage(ctx context.Context, imageName RegistryImageR
...
+	if err != nil {
+		return nil, "", fmt.Errorf("parse image name: %w", err)
 	}
-	return destRef, err
+	canonicalRef, err := reference.WithDigest(reference.TrimNamed(imageParsed), digest.FromBytes(manifestBytes))
+	if err != nil {
+		return nil, "", fmt.Errorf("create canonical reference: %w", err)
+	}
+
+	repoDigest := canonicalRef.String()
```
mtrmac, line 919: Use well-typed values. At the very least a `RegistryImageReference`, if not a new type specifically promising to include a digest.

mtrmac (changes requested):
I’m sorry, I definitely want to review this carefully, but I just can’t spare the time in the next 1-2 weeks.

This is a ~single pass over one hour, probably very redundant/repetitive.

Right now, this seems incorrect; start with the ⚠️ comments.

server/container_create.go
harche: That's a great point. Thanks. I will update the code.

internal/storage/image.go
harche: I agree, that one can be dropped.

mtrmac (commented):
Just a very quick peek at changes in the diff

internal/storage/image.go
```diff
@@ -445,6 +463,168 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+		return fmt.Errorf("checking signature of %q: %w", imageName, err)
+	}
+
+	log.Debugf(ctx, "Is allowed to run config image %s (policy path: %q)", imageName, systemContext.SignaturePolicyPath)
+
+	return nil
+}
+
+func (svc *imageService) checkSignature(ctx context.Context, sys *types.SystemContext, policyContext *signature.PolicyContext, imageName RegistryImageReference, localManifestDigest digest.Digest) error {
+	imageParsed, err := reference.ParseNormalizedNamed(imageName.Raw().String())
```
mtrmac, line 563: Why is this round-tripping through a string?!
```suggestion
	imageParsed, err := reference.ParseNormalizedNamed(imageName.Raw().String())
```

harche: @mtrmac I changed it from `imageName.Raw().String())`  to `imageName.StringForOutOfProcessConsumptionOnly()`   

Is this a better approach?

internal/storage/image.go
harche: I am looking for a way to convert `references.RegistryImageReference` to `reference.Named`. Any clue?

internal/storage/image.go
mtrmac: No, it should not be using a string _at all_. Just use `imageName.Raw()` directly, what is the benefit of the `ParseNormalizedNamed` round-trip?

mtrmac: And, uh, I was rather serious about the `ForOutOfProcessConsumptionOnly` name.

internal/storage/image.go
mtrmac: See in the other recent thread.

internal/storage/image.go
harche: >No, it should not be using a string at all. Just use imageName.Raw() directly, what is the benefit of the ParseNormalizedNamed round-trip?

Aha, got it! Thanks.

harche: > And, uh, I was rather serious about the `ForOutOfProcessConsumptionOnly` name.

:D

server/image_pull_test.go
saschagrunert: It’s fine because the tested code paths do not exist any more due to the simplification.

server/image_pull.go
```diff
@@ -133,7 +132,12 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
-	sourceCtx := *s.config.SystemContext   // A shallow copy we can modify
+	sourceCtx, err := s.contextForNamespace(pullArgs.namespace)
```
QiWang19, line 135: `contextForNamespace()` wraps the namespace check below. Should we drop the lines around? https://github.com/cri-o/cri-o/blob/73b6f0dbc6da5c349aacb8393f0b33143325a694/server/image_pull.go#L142

QiWang19: For clarity, this is the diff: https://github.com/harche/cri-o/compare/sigstore_namespace...QiWang19:cri-o:pr?expand=1

harche: /cc @saschagrunert WDYT?

saschagrunert: Yes let's simplify that.

server/container_create_linux.go
```diff
@@ -228,6 +228,15 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		someRepoDigest = imgResult.RepoDigests[0]
 	}
 
+	systemCtx, err := s.contextForNamespace(sb.Metadata().Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, *imageName, imgResult.Digest); err != nil {
```
mtrmac, line 236: ⚠️ 

Not a review, a very minimal look: This is incorrect. The signature must be verified against users’ intent (I think that’s `ctr.UserRequestedImage` but I didn’t now verify that). This does it against `SomeNameOfThisImage`, which is _explicitly documented_ to not be usable.

I … thought I might have gone overboard with the verbose `SomeNameOf` naming or with the shouting in the doc strings, it clearly isn’t sufficient. Can you suggest some other way to make these warnings stick?

mtrmac (changes requested):

server/container_create_linux.go
mtrmac: And `SomeNameOfThisImage` is, also, explicitly documented to not always be available, as suggested by the `*string` type.

harche:
```golang
	userRequestedImage, err := ctr.UserRequestedImage()
	if err != nil {
		return nil, err
	}
	// Get imageName and imageID that are later requested in container status
	var imgResult *storage.ImageResult
	if id := s.StorageImageServer().HeuristicallyTryResolvingStringAsIDPrefix(userRequestedImage); id != nil {
		imgResult, err = s.StorageImageServer().ImageStatusByID(s.config.SystemContext, *id)
		if err != nil {
			return nil, err
		}
	} else {
		potentialMatches, err := s.StorageImageServer().CandidatesForPotentiallyShortImageName(s.config.SystemContext, userRequestedImage)
		if err != nil {
			return nil, err
		}
		var imgResultErr error
		for _, name := range potentialMatches {
			imgResult, imgResultErr = s.StorageImageServer().ImageStatusByName(s.config.SystemContext, name)
			if imgResultErr == nil {
				break
			}
		}
		if imgResultErr != nil {
			return nil, imgResultErr
		}
	}
	// At this point we know userRequestedImage is not empty; "" is accepted by neither HeuristicallyTryResolvingStringAsIDPrefix
	// nor CandidatesForPotentiallyShortImageName. Just to be sure:
	if userRequestedImage == "" {
		return nil, errors.New("internal error: successfully found an image, but userRequestedImage is empty")
	}

```

`imageName` is obtained from `imgResult` which in turn comes from `ctr.UserRequestedImage()`

harche: But let us see how to get `imageName` from `imageResult` without using `SomeNameOfThisImage`. Any pointer is highly appreciated.

harche: `references.ParseRegistryImageReferenceFromOutOfProcessData(userRequestedImage)` might work, let me test it.

mtrmac: The name to enforce policy against is an _input_ to the security decision whether to trust an image. It logically does not _depend_ on image lookup. Looking for it in `imageResult` is structurally incorrect already.

---

Even without that conceptual thought, just looking at the shape of the code:

- In the `ImageStatusByID` case, the input structurally isn’t a name at all, so there is no way for it to end up in `SomeNameOfThisImage` in a way that could work.
- If you read how `ImageStatusByName` (or `…ByID`, it’s the same) constructs `imgResult`, the input `name` parameter does not participate in the construction of `SomeNameOfThisImage` at all.

---

Actually, `UserRequestedImage` is not the right field, it seems (it’s the output of Kubelet’s ImagePull and/or ImageStatus, so it can contain a non-name ID, as in the `ImageStatusByID` case); the policy needs to be driven by `UserSpecifiedImage`,  added by https://github.com/kubernetes/kubernetes/pull/118652 extra for this purpose. ( https://github.com/cri-o/cri-o/pull/7046/files does use that field .)

harche: Thanks, I changed to code to use `userSpecifiedImage := ctr.Config().GetImage().UserSpecifiedImage`  instead of  `UserRequestedImage`

harche: Also it uses `userSpecifiedImageReference, err := references.ParseRegistryImageReferenceFromOutOfProcessData(userSpecifiedImage)` to contruct `RegistryImageReference` instead of relying on `SomeNameOfThisImage`

harche: passing all policy bats tests,

```
ok 1 accept unsigned image with default policy
ok 2 deny unsigned image with restrictive policy
ok 3 deny unsigned image with restrictive policy if already pulled
ok 4 accept signed image with default policy
ok 5 accept signed image with restrictive policy
ok 6 accept unsigned image with not existing namespace policy
ok 7 accept unsigned image with higher priority namespace policy
ok 8 deny unsigned image with higher priority namespace policy
ok 9 accept signed image with higher priority namespace policy
ok 10 allow signed image with restrictive policy on container creation1 (fresh pull)
ok 11 deny unsigned image with restrictive policy on container creation2 (fresh pull)
ok 12 allow signed image with restrictive policy on container creation3 if already pulled (by ID)
ok 13 deny unsigned image with restrictive policy on container creation4 if already pulled (by ID)
ok 14 allow signed image with restrictive policy on container creation5 if already pulled (by tag)
ok 15 deny unsigned image with restrictive policy on container creation6 if already pulled (by tag)
```

harche: /hold till the CI passes.

test/policy.bats
```diff
@@ -97,3 +109,98 @@ SANDBOX_CONFIG="$TESTDATA/sandbox_config.json"
...
+
+	SIGNATURE_POLICY="$RESTRICTIVE_POLICY" start_crio
+	POD_ID=$(crictl runp "$TESTDATA/sandbox_config.json")
+	CTR_CONFIG="$TESTDIR/config.json"
+	jq '.image.image = "'"$UNSIGNED_IMAGE"'" | .image.user_specified_image = "'"$UNSIGNED_IMAGE"'"' "$TESTDATA/container_config.json" > "$CTR_CONFIG"
+
+	run ! crictl create "$POD_ID" "$CTR_CONFIG" "$TESTDATA/sandbox_config.json"
+
+	[[ "$output" == *"SignatureValidationFailed"* ]]
+}
```
mtrmac, line 206: Other interesting tests:
- Preparation:
  - Tag the same image as nameSigned:tagSigned and nameUnsigned:tagUnsigned. Sign the …Signed name. 
  - Using a permissive policy, pull both nameSigned:tagSigned, and nameUnsigned:tagUnsigned. Sanity-check that the local image ID is the same.
- Then, using a signature-enforcing policy:
  - (Start a container using UserSpecifiedImage: nameSigned:tagSigned, both by tag and by ID: should work)
  - Start a container using UserSpecifiedImage: nameUnsigned:tagUnsigned, by ID (mirroring a Kubelet that does ImageStatus(…Unsigned) and does not trigger a pull): should fail
  - Remove the nameUnsigned:tagUnsigned tag; start a container using UserSpecifiedImage: nameUnsigned:tagUnsigned, by ID (mirroring a Kubelet that does imageStatus, racing with a concurrent pull of a new version of …Unsigned moving the tag): should fail
  - Extra credit, not essential? Remove the  Remove the nameSigned:tagSigned tag; start a container using UserSpecifiedImage: nameSigned:tagSigned, by ID: should succeed.

I think the last one would currently fail because the code looks up in local storage by repo. OTOH this interacts with the “use incorrect digest” concern — we might have settled on this lookup approach in previous discussions.

Some of the “should fail” would have incorrectly succeeded in yesterday’s version of this PR.

mtrmac (commented):
Just a WIP bunch of comments, filing them now to help make progress.

I don’t see a recorded recorded resolution for https://github.com/cri-o/cri-o/pull/8212#discussion_r1683142317 ; is there one?

server/container_create_linux.go
```diff
@@ -228,6 +229,20 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		someRepoDigest = imgResult.RepoDigests[0]
 	}
```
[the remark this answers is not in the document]
harche: I modified to the code to check if the image is checkpoint image.

server/image_pull.go
```diff
@@ -177,69 +181,39 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
...
+			if repoDigest != nil {
+				return repoDigest.String(), nil
 			}
-			imageRef := status.ID.IDStringForOutOfProcessConsumptionOnly()
-			if len(status.RepoDigests) > 0 {
-				imageRef = status.RepoDigests[0]
-			}
-
-			return imageRef, nil
+			return "", nil
```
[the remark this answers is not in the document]
harche: changed `pullImageCandidate` to always return valid digest.

server/image_pull.go
```diff
@@ -177,69 +180,36 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
...
-			return imageRef, nil
+			return repoDigest, nil
 		}
 		lastErr = err
 	}
 	return "", lastErr
 }
 
-func (s *Server) pullImageCandidate(ctx context.Context, sourceCtx *imageTypes.SystemContext, remoteCandidateName storage.RegistryImageReference, decryptConfig *encconfig.DecryptConfig, cgroup string) error {
-	tmpImg, err := s.StorageImageServer().PrepareImage(sourceCtx, remoteCandidateName)
```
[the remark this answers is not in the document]
harche: /cc @saschagrunert

internal/storage/image.go
```diff
@@ -657,23 +768,38 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	canonicalRef, err := reference.WithDigest(reference.TrimNamed(imageName.Raw()), storageImg.Digest)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return destRef, canonicalRef, nil
```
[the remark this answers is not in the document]
harche: Ok, let me try this out.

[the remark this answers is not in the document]
harche: Thanks for that input. I have updated the code to not recompute canonical reference in `pullImageParent`, but rather reused the one calculated in `pullImageImplementation` which is passed on via `pullImageChild`

harche:
When I am trying to run the tests using `ginkgo run --tags test --focus 'ContainerRestore from archive into new pod' -v ./server` 3 of them fail with following error, 

```
cri-o: ContainerRestore cri-o: ContainerRestore from archive into new pod should succeed ({"rootfsImageName": "image"})
/home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:449
  [FAILED] in [It] - /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90 @ 08/30/24 13:25:03.007
• [FAILED] [0.006 seconds]
cri-o: ContainerRestore cri-o: ContainerRestore from archive into new pod [It] should succeed ({"rootfsImageName": "image"})
/home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:449

  [FAILED] Unexpected call to *criostoragemock.MockImageServer.CandidatesForPotentiallyShortImageName([0xc0005950e0 image]) at /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90 because: 
  expected call at /home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:59 doesn't match the argument at index 1.
  Got: image (string)
  Want: is equal to does-not-exist.tar (string)all expected calls for method "CandidatesForPotentiallyShortImageName" have been exhausted
  In [It] at: /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90 @ 08/30/24 13:25:03.007
cri-o: ContainerRestore cri-o: ContainerRestore from archive into new pod should succeed ({"rootfsImageRef": "8a788232037eaf17794408ff3df6b922a1aedf9ef8de36afdae3ed0b0381907b"})
/home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:449
  [FAILED] in [It] - /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90 @ 08/30/24 13:25:03.014
• [FAILED] [0.007 seconds]
cri-o: ContainerRestore cri-o: ContainerRestore from archive into new pod [It] should succeed ({"rootfsImageRef": "8a788232037eaf17794408ff3df6b922a1aedf9ef8de36afdae3ed0b0381907b"})
/home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:449

  [FAILED] Unexpected call to *criostoragemock.MockImageServer.CandidatesForPotentiallyShortImageName([0xc00012b0e0 8a788232037eaf17794408ff3df6b922a1aedf9ef8de36afdae3ed0b0381907b]) at /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90 because: 
  expected call at /home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:59 doesn't match the argument at index 1.
  Got: 8a788232037eaf17794408ff3df6b922a1aedf9ef8de36afdae3ed0b0381907b (string)
  Want: is equal to does-not-exist.tar (string)
  expected call at /home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:543 doesn't match the argument at index 1.
  Got: 8a788232037eaf17794408ff3df6b922a1aedf9ef8de36afdae3ed0b0381907b (string)
  Want: is equal to image (string)
  In [It] at: /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90 @ 08/30/24 13:25:03.014
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
[AfterSuite] 
/home/harshal/go/src/github.com/cri-o/cri-o/server/suite_test.go:83
  [FAILED] in [AfterSuite] - /home/harshal/go/src/github.com/cri-o/cri-o/vendor/github.com/golang/mock/gomock/controller.go:269 @ 08/30/24 13:25:03.015
[AfterSuite] [FAILED] [0.000 seconds]
[AfterSuite] 
/home/harshal/go/src/github.com/cri-o/cri-o/server/suite_test.go:83

  [FAILED] missing call(s) to *criostoragemock.MockRuntimeServer.StartContainer(is anything) /home/harshal/go/src/github.com/cri-o/cri-o/server/container_restore_test.go:577
  In [AfterSuite] at: /home/harshal/go/src/github.com/cri-o/cri-o/vendor/github.com/golang/mock/gomock/controller.go:269 @ 08/30/24 13:25:03.015
[DeferCleanup (Suite)] 
/home/harshal/go/src/github.com/cri-o/cri-o/vendor/github.com/golang/mock/gomock/controller.go:137
[DeferCleanup (Suite)] PASSED [0.000 seconds]

Summarizing 3 Failures:
  [FAIL] cri-o: ContainerRestore cri-o: ContainerRestore from archive into new pod [It] should succeed ({"rootfsImageName": "image"})
  /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90
  [FAIL] cri-o: ContainerRestore cri-o: ContainerRestore from archive into new pod [It] should succeed ({"rootfsImageRef": "8a788232037eaf17794408ff3df6b922a1aedf9ef8de36afdae3ed0b0381907b"})
  /home/harshal/go/src/github.com/cri-o/cri-o/server/image_status.go:90
  [FAIL] [AfterSuite] 
  /home/harshal/go/src/github.com/cri-o/cri-o/vendor/github.com/golang/mock/gomock/controller.go:269

Ran 12 of 168 Specs in 0.079 seconds
FAIL! -- 10 Passed | 2 Failed | 0 Pending | 156 Skipped
--- FAIL: TestServer (0.08s)
FAIL
```

test/policy.bats
harche: With the latest changes all those existing tests pass, 
```
1..15
ok 1 accept unsigned image with default policy
ok 2 deny unsigned image with restrictive policy
ok 3 deny unsigned image with restrictive policy if already pulled
ok 4 accept signed image with default policy
ok 5 accept signed image with restrictive policy
ok 6 accept unsigned image with not existing namespace policy
ok 7 accept unsigned image with higher priority namespace policy
ok 8 deny unsigned image with higher priority namespace policy
ok 9 accept signed image with higher priority namespace policy
ok 10 allow signed image with restrictive policy on container creation1 (fresh pull)
ok 11 deny unsigned image with restrictive policy on container creation2 (fresh pull)
ok 12 allow signed image with restrictive policy on container creation3 if already pulled (by ID)
ok 13 deny unsigned image with restrictive policy on container creation4 if already pulled (by ID)
ok 14 allow signed image with restrictive policy on container creation5 if already pulled (by tag)
ok 15 deny unsigned image with restrictive policy on container creation6 if already pulled (by tag)
``` 

but I am not sure what is going on with the mocked tests, https://github.com/cri-o/cri-o/pull/8212#issuecomment-2322025295

saschagrunert:
@harche I fixed the unit tests in https://github.com/cri-o/cri-o/pull/8212/commits/a7fa338433f556ec69df8921def8d639a401c697, but that comes with a drawback: We now double the calls for `CandidatesForPotentiallyShortImageName` and `ImageStatusByName` (see https://github.com/cri-o/cri-o/commit/a7fa338433f556ec69df8921def8d639a401c697#diff-e97f814e4217ecc24f2f480bd745fbb247972dc646f8eace746920c3af3b3601R537-R587).

We may wanna live with that or optimize further, I also had to change the path `userSpecifiedImage = imageName.StringForOutOfProcessConsumptionOnly()`, because otherwise we would use an image ID but we need a name.

saschagrunert:
@cri-o/cri-o-maintainers PTAL. CI seems to become green.

server/container_create_linux.go
```diff
@@ -229,6 +231,33 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	userSpecifiedImage := ctr.Config().GetImage().UserSpecifiedImage
+
+	if userSpecifiedImage == "" {
+		// Attempt to check if the image is a checkpoint OCI image
+		if _, err := s.checkIfCheckpointOCIImage(ctx, userRequestedImage); err != nil {
+			return nil, fmt.Errorf("failed to check if this is a checkpoint image: %w", err)
+		} else { // Fallback to ensure that userSpecifiedImage is valid
```
haircommander, line 245: non-blocking nit: we don't need the else because the above block returns

haircommander:
/approve
/lgtm

well done team! potentially some follow ups here but let's get the base feature in and iterate

haircommander:
/lgtm

(rebased)

server/container_create_linux.go
```diff
@@ -229,6 +231,33 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	}
+
+	var userImageRef references.RegistryImageReference
+
+	userImageRef, err = references.ParseRegistryImageReferenceFromOutOfProcessData(userSpecifiedImage)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get userImageRef from user specified image %q: %w", userSpecifiedImage, err)
+	}
+
+	if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, userImageRef, imgResult.Digest); err != nil {
```
mtrmac, line 257: Using `imgResult.Digest` here is not a _security_ issue _yet_ but seems incorrect. `.Digest` is _some_ digest of this deduplicated image, not necessarily the one the user wanted or one that is signed.

It is possible to have a single image deduplicated image for `repo1:tag1 == digest1` and `repo2:tag2 == digest2`. And we can have `UserSpecifiedImage` (user intent) for repo1:tag1, but `imgResult` returning `digest2`. The way `IsRunningImageAllowed` is implemented, this will check the `digest2` manifest for signatures, and even if the image is correctly signed for `repo1:tag1`, the signature verification will fail. (To make that example even more vivid, consider user input `repo1@digest1`, with `imgResult.Digest == digest2`. That should intuitively fail any policy check.)

---

⚠️ 

This _will_ become a security issue in a bit: Currently the same image digest implies the same image ID, but that’s no longer the case with zstd:chunked (where it matters _how_ we pulled the image).

Basically, the “which image are we verifying” parameter to `IsRunningImageAllowed` should be `imageID`, and the `imageID` should participate in `istorage.Transport.NewStoreReference`.

Then, in the `IsRunningImage` implementation, we can discuss the tradeoffs of calling `NewStoreReference` _only_ with the image ID vs. also with `userSpecifiedImageRef`, or possibly trying both.

mtrmac (commented):
Just looking at recent changes…

server/container_create_linux.go
```diff
@@ -229,6 +231,33 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	}
+
+	userSpecifiedImage := ctr.Config().GetImage().UserSpecifiedImage
+
+	if userSpecifiedImage == "" {
+		// Attempt to check if the image is a checkpoint OCI image
+		if _, err := s.checkIfCheckpointOCIImage(ctx, userRequestedImage); err != nil {
+			return nil, fmt.Errorf("failed to check if this is a checkpoint image: %w", err)
+		} else { // Fallback to ensure that userSpecifiedImage is valid
+			userSpecifiedImage = imageName.StringForOutOfProcessConsumptionOnly()
```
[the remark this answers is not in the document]
mtrmac: Note to self: Re-propose the PR that explicitly used the ugly `UserRequestedImage`/`SomeNameOfThisImage` names throughout the codebase, now that we have ~proof that it’s hard to keep these things in mind.

server/container_create_linux.go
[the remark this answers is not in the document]
mtrmac: Note to self: This is being handled elsewhere.

server/image_pull.go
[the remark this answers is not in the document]
mtrmac: [_If_ this were to be removed, this removes the only users of `.PrepareImage`, `ImageResult.ConfigDigest`, and maybe more. So there would be various opportunities to simplify.]

server/container_create_linux.go
```diff
@@ -229,6 +231,33 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	systemCtx, err := s.contextForNamespace(sb.Metadata().Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	userSpecifiedImage := ctr.Config().GetImage().UserSpecifiedImage
+
+	if userSpecifiedImage == "" {
+		// Attempt to check if the image is a checkpoint OCI image
+		if _, err := s.checkIfCheckpointOCIImage(ctx, userRequestedImage); err != nil {
```
mtrmac, line 243: Oh, and, of course: This is looking up `userRequestedImage` _again_, so it is racy against tag changes. It should be only working with `imgResult.ID`.

---

After `imgResult.ID` is first read, a
```go
// == Image lookup done.
// == NEVER USE userRequestedImage for anything but diagnostic logging past this point; it might
// resolve to a different image.
```
comment would not urt.

server/container_create_linux.go
haircommander: updated to use the imageID, which simplifies things anyway. Because we already have access to imageRef, I chose not to use userSpecifiedImageRef there.

saschagrunert (approved):
Code LGTM, let’s make sure the integration test CI is green.

internal/storage/image.go
```diff
@@ -445,6 +474,88 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+	if err != nil {
+		var policyErr signature.PolicyRequirementError
+		if errors.As(err, &policyErr) {
+			return fmt.Errorf("%w: %w", crierrors.ErrSignatureValidationFailed, err)
+		}
+		return fmt.Errorf("verifying signatures: %w", err)
+	}
+	if !allowed {
+		return errors.New("internal inconsistency: !allowed and no error")
+	}
```
kwilczynski, line 554: Should this be something more severe? Like a `panic()` or `log.Fatal()`? Would this be a "this can never happen" conditions?

server/container_restore.go
```diff
@@ -226,6 +226,15 @@ func (s *Server) CRImportCheckpoint(
 		return "", fmt.Errorf("specified sandbox not found: %s: %w", sbID, err)
 	}
 
+	systemCtx, err := s.contextForNamespace(sb.Metadata().Namespace)
+	if err != nil {
+		return "", fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	if systemCtx.SignaturePolicyPath != "" {
+		return "", fmt.Errorf("namespaced signature policy defined for pods in namespace %s; signature validation is currently incompatible with restoring", sb.Metadata().Namespace)
```
kwilczynski, line 235: Nit: make the language of the error simpler and more explicit, perhaps.

```suggestion
		return "", fmt.Errorf("namespaced signature policy defined for pods in namespace %s; signature validation is supported for container restore", sb.Metadata().Namespace)
```

haircommander: done (though I corrected "not supported")

internal/storage/image.go
haircommander: updated to log.Fatalf

server/container_restore.go
kwilczynski: @haircommander, ah! Sorry! Thank you!

internal/storage/image.go
kwilczynski: Just to make sure: would this semantic be acceptable? We will now terminate CRI-O here. So want to make sure we do the right thing...

kwilczynski: I mean, the new semantics/behaviour with `log.Fatalf()`. We already got a linter complaint. :smile:

kwilczynski: OK. I saw latest update. Makes sense.

haircommander: I updated with a direct panic which the linter doesn't complain about. my understanding is defers still run in a panic scenerio

haircommander: though I'm not sure why log.Fatalf doesn't work the same way...

kwilczynski: @haircommander, nice!

_For someone who stumbles here:_

The `panic()` has proper housekeeping within the runtime (panic is part of the "defer", "panic" and "recover" trio that can be used for control flow; interestingly, only "defer" here is a keyword, rest is more like `println()` [not the `fmt` package one]), whereas `log.Fatalf()` just prints (at the right log level) and then calls `os.Exit()`, which then does nothing other than calls the underlying OS-specific exit implementation, I suppose.

haircommander:
 ( fixed in https://github.com/cri-o/cri-o/pull/8578)
/retest

haircommander:
/lgtm

(had to update the CR test)

internal/storage/image.go
mtrmac: FWIW the only thing I care about here is that the pull does not succeed. 

WRT the question whether CRI-O wants to panic (and presumably be restarted?) or just return an error: because this supposedly can’t happen, I don’t really know how to choose which one to prefer.

Purely hypothetically, if there were a way to trigger this code path: if it were 100% deterministic and it lead to failure, the only true effect would be an attacker triggering a CRI-O restart. Possibly that’s a performance degradation?

If there were some randomized way for the evaluation to return unexpected data… on unexpected policy rejections, it doesn’t matter. On hypothetical unexpected policy successes… as long as the node continues to make policy evaluations (i.e. whether this returns an error or panics and triggers a CRi-O restart), there’s a risk.

So I suppose the one _slightly_ better alternative would be for CRI-O to take the node down and to have it _stay down_, but that’s clearly unpalatable. And we wouldn’t be writing this new mechanism for a “should never happen” error case.

mtrmac: … as a point of reference, the pull itself just returns an error:
https://github.com/containers/image/blob/31d4ad14fe4da8ef3969ff67297831ab291c76f1/copy/single.go#L77-L79

In retrospect, this two-returns API has not really been useful. c/image promises API stability so we can’t just _change_ it, but adding a new function might happen in a future.

kwilczynski: @mtrmac, to add a bit more colour here.

[...]
> Purely hypothetically, if there were a way to trigger this code path: if it were 100% deterministic and it lead to failure, the only true effect would be an attacker triggering a CRI-O restart. Possibly that’s a performance degradation?
[...]

The way to deploy and set things up varies, but for a typical OpenShift node, CRI-O crashing repeatedly would eventually lead to systemd marking the unit as broken, and CRI-O would remain stopped. This leads to the kubelet losing its connectivity to the CRI, eventually marking the node as NotReady. Hopefully, monitoring will pick this up, and the operator will know something is wrong. So, this can be a bit of a _saving grace_, so to speak.

An excerpt from the systemd unit for CRI-O (OpenShift 4.12.29):

```
StartLimitIntervalUSec=10s
StartLimitBurst=5
```

From [man 5 systemd.unit]( https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html) (section: _StartLimitIntervalSec=interval, StartLimitBurst=burst_):

> Configure unit start rate limiting. Units which are started more than burst times within an interval time span are not permitted to start any more. Use StartLimitIntervalSec= to configure the checking interval and StartLimitBurst= to configure how many starts per interval are allowed.

Someone would have to perform a "slow loris" type attack here to keep CRI-O crashing without triggering the systemd limits, I suppose.

mtrmac: Thanks, I admit I didn’t check. In that case, terminating the process is more likely to highlight the problem, and seems a bit better to me.

mtrmac (commented):
Late review … now comprehensive.

I’ll file a PR for some parts, and an issue for the remaining outstanding things.

server/image_pull.go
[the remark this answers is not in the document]
mtrmac: Some cleanups proposed in #8590 .

server/container_create_linux.go
```diff
@@ -228,6 +229,20 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	if err != nil {
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	userSpecifiedImage := ctr.Config().GetImage().UserSpecifiedImage
+	userSpecifiedImageReference, err := references.ParseRegistryImageReferenceFromOutOfProcessData(userSpecifiedImage)
+	if err != nil {
+		return nil, err
+	}
+	if err := s.StorageImageServer().IsRunningImageAllowed(ctx, &systemCtx, userSpecifiedImageReference, imgResult.Digest); err != nil {
```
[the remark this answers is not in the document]
mtrmac: The final version looks up the image by storage ID. #8590 comments what else we might do.

server/image_pull.go
```diff
@@ -177,69 +181,38 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string
...
-			status, err := s.StorageImageServer().ImageStatusByName(s.config.SystemContext, remoteCandidateName)
-			if err != nil {
-				return "", err
-			}
-			imageRef := status.ID.IDStringForOutOfProcessConsumptionOnly()
-			if len(status.RepoDigests) > 0 {
-				imageRef = status.RepoDigests[0]
+			if repoDigest != nil {
+				return repoDigest.String(), nil
 			}
```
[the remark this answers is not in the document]
mtrmac: Should be fixed in #8590 .

server/container_create_linux.go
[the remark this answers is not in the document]
mtrmac: #8590 re-proposes the verbose names (going even further than before).

server/container_create_linux.go
mtrmac: #8590 adds that comment.

server/image_pull_test.go
```diff
@@ -18,9 +18,7 @@ import (
 var _ = t.Describe("ImagePull", func() {
 	imageCandidate, err := references.ParseRegistryImageReferenceFromOutOfProcessData("docker.io/library/image:latest")
 	Expect(err).ToNot(HaveOccurred())
-	imageID, err := storage.ParseStorageImageIDFromOutOfProcessData("2a03a6059f21e150ae84b0973863609494aad70f0a80eaeb64bddd8d92465812")
-	Expect(err).ToNot(HaveOccurred())
-	otherImageID, err := storage.ParseStorageImageIDFromOutOfProcessData("3a03a6059f21e150ae84b0973863609494aad70f0a80eaeb64bddd8d92465812")
+	canonicalImageCandidate, err := reference.WithDigest(imageCandidate.Raw(), digest.Digest("sha256:340d9b015b194dc6e2a13938944e0d016e57b9679963fdeb9ce021daac430221"))
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590 .

internal/storage/runtime.go
```diff
@@ -316,7 +316,7 @@ func (r *runtimeService) CreatePodSandbox(systemContext *types.SystemContext, po
 		if imageAuthFile != "" {
 			sourceCtx.AuthFilePath = imageAuthFile
 		}
-		ref, err = r.storageImageServer.PullImage(context.Background(), pauseImage, &ImageCopyOptions{
+		ref, _, err = r.storageImageServer.PullImage(context.Background(), pauseImage, &ImageCopyOptions{
```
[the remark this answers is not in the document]
mtrmac: Done in #8590.

internal/storage/runtime_test.go
```diff
@@ -766,7 +766,7 @@ var _ = t.Describe("Runtime", func() {
 				imageServerMock.EXPECT().GetStore().Return(storeMock),
 				mockResolveReference(storeMock, storageTransportMock,
 					"docker.io/library/pauseimagename:latest", "", ""),
-				imageServerMock.EXPECT().PullImage(gomock.Any(), pauseImageRef, expectedCopyOptions).Return(pulledRef, nil),
+				imageServerMock.EXPECT().PullImage(gomock.Any(), pauseImageRef, expectedCopyOptions).Return(pulledRef, nil, nil),
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

server/container_restore.go
```diff
@@ -226,6 +226,15 @@ func (s *Server) CRImportCheckpoint(
 		return "", fmt.Errorf("specified sandbox not found: %s: %w", sbID, err)
 	}
 
+	systemCtx, err := s.contextForNamespace(sb.Metadata().Namespace)
+	if err != nil {
+		return "", fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	if systemCtx.SignaturePolicyPath != "" {
```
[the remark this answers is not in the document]
mtrmac: #8590 tries doing this.

internal/storage/image.go
```diff
@@ -506,6 +613,8 @@ type pullImageArgs struct {
 type pullImageOutputItem struct {
 	Progress *types.ProgressProperties `json:",omitempty"`
 	Result   string                    `json:",omitempty"` // If not "", in the format of transport.ImageName()
+	Name     string                    `json:",omitempty"`
+	Digest   digest.Digest             `json:",omitempty"`
```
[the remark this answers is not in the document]
mtrmac: #8590 goes back to a single `Result`.

internal/storage/image.go
```diff
@@ -643,9 +754,19 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
...
 			}
 
+			named, err := reference.ParseNamed(item.Name)
+			if err != nil {
+				break
+			}
+			canonicalRef, err = reference.WithDigest(named, item.Digest)
+			if err != nil {
+				break
+			}
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

internal/storage/image.go
```diff
@@ -643,9 +754,19 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
 				break
 			}
 
+			named, err := reference.ParseNamed(item.Name)
+			if err != nil {
+				break
+			}
+			canonicalRef, err = reference.WithDigest(named, item.Digest)
+			if err != nil {
+				break
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

internal/storage/image.go
```diff
@@ -643,9 +754,19 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
 				break
 			}
 
+			named, err := reference.ParseNamed(item.Name)
+			if err != nil {
+				break
+			}
+			canonicalRef, err = reference.WithDigest(named, item.Digest)
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

internal/storage/image.go
```diff
@@ -133,7 +143,17 @@ type ImageServer interface {
...
 	// PullImage imports an image from the specified location.
-	PullImage(ctx context.Context, imageName RegistryImageReference, options *ImageCopyOptions) (types.ImageReference, error)
+	//
+	// Arguments:
+	// - ctx: The context for controlling the function's execution
+	// - imageName: A RegistryImageReference representing the image to be pulled
+	// - options: Pointer to ImageCopyOptions, which contains various options for the image copy process
+	//
+	// Returns:
+	// - types.ImageReference: The reference to the pulled image
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

internal/storage/image.go
```diff
@@ -133,7 +143,17 @@ type ImageServer interface {
...
+	// - reference.Canonical: A canonical reference to the image, typically including the digest
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

internal/storage/image.go
```diff
@@ -156,6 +176,15 @@ type ImageServer interface {
 
 	// UpdatePinnedImagesList updates pinned and pause images list in imageService.
 	UpdatePinnedImagesList(imageList []string)
+
+	// IsRunningImageAllowed verifies if running of the container image is allowed.
+	//
+	// Arguments:
+	// - ctx: The context for controlling the function's execution
+	// - systemContext: server's system context for the given namespace
+	// - imageName: A RegistryImageReference that identifies the container image
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

internal/storage/image.go
```diff
@@ -156,6 +176,15 @@ type ImageServer interface {
 
 	// UpdatePinnedImagesList updates pinned and pause images list in imageService.
 	UpdatePinnedImagesList(imageList []string)
+
+	// IsRunningImageAllowed verifies if running of the container image is allowed.
+	//
+	// Arguments:
+	// - ctx: The context for controlling the function's execution
+	// - systemContext: server's system context for the given namespace
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

internal/storage/image.go
```diff
@@ -445,6 +474,88 @@ func (svc *imageService) imageStatus(systemContext *types.SystemContext, unstabl
...
+	if err := svc.checkSignature(ctx, systemContext, policyContext, imageName, imageID); err != nil {
+		return fmt.Errorf("checking signature of %q: %w", imageName, err)
+	}
+
+	log.Debugf(ctx, "Is allowed to run config image %s (policy path: %q)", imageName, systemContext.SignaturePolicyPath)
+
+	return nil
+}
+
+func (svc *imageService) checkSignature(ctx context.Context, sys *types.SystemContext, policyContext *signature.PolicyContext, imageName RegistryImageReference, imageID StorageImageID) error {
```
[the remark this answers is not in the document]
mtrmac: Fixed in #8590.

server/container_create_linux.go
```diff
@@ -229,6 +235,37 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		someRepoDigest = imgResult.RepoDigests[0]
 	}
 
+	systemCtx, err := s.contextForNamespace(sb.Metadata().Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("get context for namespace: %w", err)
+	}
+
+	if systemCtx.SignaturePolicyPath != "" {
```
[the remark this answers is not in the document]
mtrmac: Commented in #8590 .

mtrmac:
I have filed #8590 with various cleanups; I’d be happy to hand off finishing that.
And #8591 for possible new tests; I’m definitely leaving that part to CRI-O maintainers.

PR #8213, opened 2024-05-24
Reload config should remove pinned images when an empty list is provided

### LINKED ISSUE #8202: crio v1.29.4 config reload does not update removed pinned_images
### What happened?

```
crio --version
crio version 1.29.4-6.rhaos4.16.git0e93ae2.el9
Version:        1.29.4-6.rhaos4.16.git0e93ae2.el9
```

Verify  status of unpinned image
```
crictl image --pinned quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5f93519952ee4389c0b5a380a49eb74cbb8b3b040da843f31fee5b3ba8fa8515
IMAGE                                            TAG                 IMAGE ID            SIZE                PINNED
quay.io/openshift-release-dev/ocp-v4.0-art-dev   <none>              94a4bca6c68c4       364MB               false
```

Update crio config adding image to set and reload crio
```
[crio]
  [crio.image]
    pinned_images = ["quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5f93519952ee4389c0b5a380a49eb74cbb8b3b040da843f31fee5b3ba8fa8515"]
```

Image now shown as pinned
```
crictl image --pinned quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5f93519952ee4389c0b5a380a49eb74cbb8b3b040da843f31fee5b3ba8fa8515
IMAGE                                            TAG                 IMAGE ID            SIZE                PINNED
quay.io/openshift-release-dev/ocp-v4.0-art-dev   <none>              94a4bca6c68c4       364MB               true
```
Remove image from set

```
[crio]
  [crio.image]
```
or
```
[crio]
  [crio.image]
    pinned_images = []
```
reload service and it still shows pinned

```
crictl image --pinned quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5f93519952ee4389c0b5a380a49eb74cbb8b3b040da843f31fee5b3ba8fa8515
IMAGE                                            TAG                 IMAGE ID            SIZE                PINNED
quay.io/openshift-release-dev/ocp-v4.0-art-dev   <none>              94a4bca6c68c4       364MB               true
```
```
crio status config | grep pinned_images
INFO[2024-05-20 21:09:55.934592808Z] Starting CRI-O, version: 1.29.4-6.rhaos4.16.git0e93ae2.el9, git: unknown(clean) 
    pinned_images = ["quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5f93519952ee4389c0b5a380a49eb74cbb8b3b040da843f31fee5b3ba8fa8515"]
```

cache to be updated when images are removed from set

above example is reproducible with the version described

original bug ref: https://github.com/cri-o/cri-o/issues/7954

<details>

```console
$ crio --version
crio version 1.29.4-6.rhaos4.16.git0e93ae2.el9
Version:        1.29.4-6.rhaos4.16.git0e93ae2.el9
GitCommit:      unknown
GitCommitDate:  unknown
GitTreeState:   clean
GoVersion:      go1.21.9 (Red Hat 1.21.9-1.el9_4)
Compiler:       gc
Platform:       linux/amd64
Linkmode:       dynamic
BuildTags:      
  rpm_crashtraceback
  libtrust_openssl
  selinux
  seccomp
  exclude_graphdriver_btrfs
  containers_image_ostree_stub
LDFlags:           -compressdwarf=false -B 0xd3920d959b04383b0a796489d198b4c103702925 -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 '
SeccompEnabled:   true
AppArmorEnabled:  false
```
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
ID="rhcos"
ID_LIKE="rhel fedora"
VERSION="416.94.202405081407-0"
VERSION_ID="4.16"
VARIANT="CoreOS"
VARIANT_ID=coreos
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux CoreOS 416.94.202405081407-0"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos::coreos"
DOCUMENTATION_URL="https://docs.okd.io/latest/welcome/index.html"
REDHAT_BUGZILLA_PRODUCT_VERSION="4.16"
REDHAT_SUPPORT_PRODUCT_VERSION="4.16"
OPENSHIFT_VERSION="4.16"
RHEL_VERSION=9.4
OSTREE_VERSION="416.94.202405081407-0"

$ uname -a
Linux ip-10-0-56-46 5.14.0-427.16.1.el9_4.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Apr 26 18:16:09 EDT 2024 x86_64 x86_64 x86_64 GNU/Linux

```

</details>

<details>
AWS 4.16 nightly AWS
</details>

roman-kiselenko:
#### What type of PR is this?

/kind bug

Fixes #8202

```release-note
Reload config should remove pinned images when an empty list is provided
```

### DIFF
--- a/pkg/config/reload.go
+++ b/pkg/config/reload.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/cri-o/cri-o/internal/log"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/sirupsen/logrus"
 	"tags.cncf.io/container-device-interface/pkg/cdi"
 )
@@ -145,24 +147,34 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
 	return nil
 }
 
-// ReloadPinnedImages updates the PinnedImages with the provided `newConfig`.
-// The method print log in case of any updates.
+// ReloadPinnedImages replace the PinnedImages
+// with the provided `newConfig.PinnedImages`.
+// The method skips empty items and prints a log message.
 func (c *Config) ReloadPinnedImages(newConfig *Config) {
-	updatedPinnedImages := make([]string, len(newConfig.PinnedImages))
-	updatedPinnedImageList := false
+	if len(newConfig.PinnedImages) == 0 {
+		c.PinnedImages = []string{}
+		logConfig("pinned_images", "[]")
+		return
+	}
 
-	for i, image := range newConfig.PinnedImages {
-		if i < len(c.PinnedImages) && image == c.PinnedImages[i] {
-			continue
+	if cmp.Equal(c.PinnedImages, newConfig.PinnedImages,
+		cmpopts.SortSlices(func(a, b string) bool {
+			return a < b
+		}),
+	) {
+		return
+	}
+
+	pinnedImages := []string{}
+	for _, img := range newConfig.PinnedImages {
+		if img != "" {
+			pinnedImages = append(pinnedImages, img)
 		}
-		updatedPinnedImages[i] = image
-		updatedPinnedImageList = true
 	}
 
-	if updatedPinnedImageList {
-		logConfig("pinned_images", strings.Join(updatedPinnedImages, ", "))
-		c.PinnedImages = updatedPinnedImages
-	}
+	logConfig("pinned_images", strings.Join(pinnedImages, ","))
+
+	c.PinnedImages = pinnedImages
 }
 
 // ReloadRegistries reloads the registry configuration from the Configs
--- a/contrib/test/ci/integration.yml
+++ b/contrib/test/ci/integration.yml
@@ -56,6 +56,7 @@
         + ['command.bats'] | product(kata_skip_command_tests) \
         + ['config.bats'] | product(kata_skip_config_tests) \
         + ['config_migrate.bats'] | product(kata_skip_config_migrate_tests) \
+        + ['reload_config.bats'] | product(kata_skip_reload_config) \
         + ['crio-wipe.bats'] | product(kata_skip_crio_wipe_tests) \
         + ['ctr.bats'] | product(kata_skip_ctr_tests) \
         + ['devices.bats'] | product(kata_skip_devices_tests) \
--- a/contrib/test/ci/vars.yml
+++ b/contrib/test/ci/vars.yml
@@ -106,6 +106,8 @@ kata_skip_config_tests:
 kata_skip_config_migrate_tests:
   - 'test "config migrate should succeed with default config"'
   - 'test "config migrate should succeed with 1.17 config"'
+kata_skip_reload_config:
+  - 'test "reload config should remove pinned images when an empty list is provided"'
 kata_skip_crio_wipe_tests:
   - 'test "clear neither when remove persist"'
   - "test \"don't clear containers on a forced restart of crio\""
--- a/go.mod
+++ b/go.mod
@@ -39,6 +39,7 @@ require (
 	github.com/godbus/dbus/v5 v5.1.1-0.20230522191255-76236955d466
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/mock v1.6.0
+	github.com/google/go-cmp v0.6.0
 	github.com/google/renameio v1.0.1
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
@@ -138,7 +139,6 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/gnostic-models v0.6.9-0.20230804172637-c7be7c783f49 // indirect
-	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/google/go-containerregistry v0.19.1 // indirect
 	github.com/google/go-github/v60 v60.0.0 // indirect
 	github.com/google/go-intervals v0.0.2 // indirect
--- a/test/reload_config.bats
+++ b/test/reload_config.bats
@@ -269,3 +269,29 @@ EOF
 	# pause image is pinned
 	[[ "$output" == *"fedora-crio-ci"* ]]
 }
+
+@test "reload config should remove pinned images when an empty list is provided" {
+	EXAMPLE_IMAGE=quay.io/crio/fedora-crio-ci:latest
+
+	# Add a pinned image to the configuration
+	printf '[crio.image]\npinned_images = ["%s", ""]\n' $EXAMPLE_IMAGE > "$CRIO_CONFIG_DIR"/01-overwrite
+	reload_crio
+	wait_for_log 'Set config pinned_images to \\"quay.io/crio/fedora-crio-ci:latest\\"'
+	wait_for_log "Configuration reload completed"
+	wait_for_log 'pinned_images = \[\\"quay.io/crio/fedora-crio-ci:latest\\"\]'
+
+	# Verify that the image is pinned
+	output=$(crictl images -o json | jq ".images[] | select(.repoTags[] == \"$EXAMPLE_IMAGE\") |.pinned")
+	[ "$output" == "true" ]
+
+	# Remove the pinned image from the configuration
+	printf '[crio.image]\npinned_images = []\n' > "$CRIO_CONFIG_DIR"/01-overwrite
+	reload_crio
+	wait_for_log 'Set config pinned_images to \\"\[\]\\"'
+	wait_for_log "Configuration reload completed"
+	wait_for_log 'pinned_images = \[\]'
+
+	# Verify that the image is no longer pinned
+	output=$(crictl images -o json | jq ".images[] | select(.repoTags[] == \"$EXAMPLE_IMAGE\") |.pinned")
+	[ "$output" == "false" ]
+}
--- a/vendor/github.com/google/go-cmp/cmp/cmpopts/equate.go
+++ b/vendor/github.com/google/go-cmp/cmp/cmpopts/equate.go
@@ -1,0 +1,185 @@
+// Copyright 2017, The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package cmpopts provides common options for the cmp package.
+package cmpopts
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"reflect"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func equateAlways(_, _ interface{}) bool { return true }
+
+// EquateEmpty returns a [cmp.Comparer] option that determines all maps and slices
+// with a length of zero to be equal, regardless of whether they are nil.
+//
+// EquateEmpty can be used in conjunction with [SortSlices] and [SortMaps].
+func EquateEmpty() cmp.Option {
+	return cmp.FilterValues(isEmpty, cmp.Comparer(equateAlways))
+}
+
+func isEmpty(x, y interface{}) bool {
+	vx, vy := reflect.ValueOf(x), reflect.ValueOf(y)
+	return (x != nil && y != nil && vx.Type() == vy.Type()) &&
+		(vx.Kind() == reflect.Slice || vx.Kind() == reflect.Map) &&
+		(vx.Len() == 0 && vy.Len() == 0)
+}
+
+// EquateApprox returns a [cmp.Comparer] option that determines float32 or float64
+// values to be equal if they are within a relative fraction or absolute margin.
+// This option is not used when either x or y is NaN or infinite.
+//
+// The fraction determines that the difference of two values must be within the
+// smaller fraction of the two values, while the margin determines that the two
+// values must be within some absolute margin.
+// To express only a fraction or only a margin, use 0 for the other parameter.
+// The fraction and margin must be non-negative.
+//
+// The mathematical expression used is equivalent to:
+//
+//	|x-y| ≤ max(fraction*min(|x|, |y|), margin)
+//
+// EquateApprox can be used in conjunction with [EquateNaNs].
+func EquateApprox(fraction, margin float64) cmp.Option {
+	if margin < 0 || fraction < 0 || math.IsNaN(margin) || math.IsNaN(fraction) {
+		panic("margin or fraction must be a non-negative number")
+	}
+	a := approximator{fraction, margin}
+	return cmp.Options{
+		cmp.FilterValues(areRealF64s, cmp.Comparer(a.compareF64)),
+		cmp.FilterValues(areRealF32s, cmp.Comparer(a.compareF32)),
+	}
+}
+
+type approximator struct{ frac, marg float64 }
+
+func areRealF64s(x, y float64) bool {
+	return !math.IsNaN(x) && !math.IsNaN(y) && !math.IsInf(x, 0) && !math.IsInf(y, 0)
+}
+func areRealF32s(x, y float32) bool {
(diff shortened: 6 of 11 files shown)


pkg/config/reload.go
```diff
@@ -148,6 +150,16 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
 // ReloadPinnedImages updates the PinnedImages with the provided `newConfig`.
 // The method print log in case of any updates.
 func (c *Config) ReloadPinnedImages(newConfig *Config) {
+	if len(newConfig.PinnedImages) == 0 {
+		c.PinnedImages = newConfig.PinnedImages
+		return
+	}
```
kwilczynski, line 156: Setting the value to an empty slice might be more explicit.

```suggestion
	if len(newConfig.PinnedImages) == 0 {
		c.PinnedImages = []string{}
		return
	}
```

pkg/config/reload.go
```diff
@@ -148,6 +150,16 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
...
+	sort.Strings(newConfig.PinnedImages)
+	sort.Strings(c.PinnedImages)
+
+	if cmp.Equal(newConfig.PinnedImages, c.PinnedImages) {
+		return
+	}
```
kwilczynski, line 162: This is a nice optimisation here. Correct? This can also be streamlined as follows:

```suggestion
	if cmp.Equal(c.PinnedImages, newConfig.PinnedImages,
		cmpopts.SortSlices(func(a, b string) bool {
			return a < b
		}),
	) {
		return
	}
```

Perhaps it's a bit awkward-looking when wrapped with an if-statement.

Assume that there is "github.com/google/go-cmp/cmp/cmpopts" an import added.

kwilczynski: That said, this `cmp.Equal()` could probably be used to retire the rest of this function body, correct? I am not sure what the rest of it aims to do, really. It also appears broken, for example:

- Initiate state of pinned_images set using configuration file

```toml
  [crio.image]
    default_transport = "docker://"
    global_auth_file = ""
    pause_image = "registry.k8s.io/pause:3.9"
    pause_image_auth_file = ""
    pause_command = "/pause"
    pinned_images = ["abc", "def"]
    signature_policy = ""
    signature_policy_dir = "/etc/crio/policies"
    image_volumes = "mkdir"
    big_files_temporary_dir = ""
    auto_reload_registries = false
```

- Updated the configuration file and send SIGHUP to reload configuration

```toml
  [crio.image]
    default_transport = "docker://"
    global_auth_file = ""
    pause_image = "registry.k8s.io/pause:3.9"
    pause_image_auth_file = ""
    pause_command = "/pause"
    pinned_images = ["", "", "xyz"]
    signature_policy = ""
    signature_policy_dir = "/etc/crio/policies"
    insecure_registries = ["127.0.0.0/8"]
    image_volumes = "mkdir"
    big_files_temporary_dir = ""
    auto_reload_registries = false
```

What? That can't be right...

There is also a matter of silently injecting pause_image, which makes sense, but should probably be handled a little bit better, for example:

- Initial state, CRI-O was just started

```toml
[crio.image]
default_transport = "docker://"
global_auth_file = ""
pause_image = "registry.k8s.io/pause:3.9"
pause_image_auth_file = ""
pause_command = "/pause"
signature_policy = ""
signature_policy_dir = "/etc/crio/policies"
image_volumes = "mkdir"
big_files_temporary_dir = ""
auto_reload_registries = false
``` 

Note: the pinned_images list is entirely missing.

- After sending SIGHUP to reload current configuration, but there were no new changes provided...

```toml
[crio.image]
default_transport = "docker://"
global_auth_file = ""
pause_image = "registry.k8s.io/pause:3.9"
pause_image_auth_file = ""
pause_command = "/pause"
pinned_images = ["registry.k8s.io/pause:3.9"]
signature_policy = ""
signature_policy_dir = "/etc/crio/policies"
insecure_registries = ["127.0.0.0/8"]
image_volumes = "mkdir"
big_files_temporary_dir = ""
auto_reload_registries = false
```

Hmm, something's afoot here...

We should fix this, I think.

roman-kiselenko: That's a good catch 👍 

By the way, I'm thinking about [`regexForPinnedImages`](https://github.com/cri-o/cri-o/blob/main/internal/storage/image.go#L96) and the `pinned_images` list. They might be merged into one list and passed to `imageService`. 

We could build a `pinned_images` list in the `config` package and move the logic below to the right place. Hence `internal/storage/image` would work with a compiled list of `pinned_images` + `pause_image`. 

https://github.com/cri-o/cri-o/blob/747c152817970360fafa2b7264818681436c0939/internal/storage/image.go#L858-L860

@cri-o/cri-o-maintainers WDYT?

kwilczynski: Something that @LenkaSeg might want to keep a note of, given the other work on configuration reload.

pkg/config/reload.go
```diff
@@ -148,21 +151,29 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
...
 // The method print log in case of any updates.
 func (c *Config) ReloadPinnedImages(newConfig *Config) {
-	updatedPinnedImages := make([]string, len(newConfig.PinnedImages))
-	updatedPinnedImageList := false
+	if len(newConfig.PinnedImages) == 0 {
+		c.PinnedImages = []string{}
+		return
+	}
+	sort.Strings(newConfig.PinnedImages)
+	sort.Strings(c.PinnedImages)
```
kwilczynski, line 159: These are not needed as `cmp.Equal()` uses the `cmpopts.SortSlices()` callback.

pkg/config/reload.go
```diff
@@ -148,21 +151,29 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
...
-	for i, image := range newConfig.PinnedImages {
-		if i < len(c.PinnedImages) && image == c.PinnedImages[i] {
+	if cmp.Equal(c.PinnedImages, newConfig.PinnedImages,
+		cmpopts.SortSlices(func(a, b string) bool {
+			return a < b
+		}),
+	) {
+		return
+	}
+	newList := []string{}
```
kwilczynski, line 168:
```suggestion

	pinnedImages := []string{}
```

pkg/config/reload.go
```diff
@@ -148,21 +151,29 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
...
-		updatedPinnedImages[i] = image
-		updatedPinnedImageList = true
-	}
-
-	if updatedPinnedImageList {
-		logConfig("pinned_images", strings.Join(updatedPinnedImages, ", "))
-		c.PinnedImages = updatedPinnedImages
+		newList = append(newList, img)
 	}
+	logConfig("pinned_images", strings.Join(newList, ", "))
```
kwilczynski, line 175: We probably don't care about the space?

```suggestion
	logConfig("pinned_images", strings.Join(newList, ","))
```

roman-kiselenko:
I guess there is a bug somewhere (in the test), `fedora-kata` tests don't [wait](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8213/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1797689029882286080/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt#:~:text=%23%20time%3D%222024%2D06%2D03%2019%3A11%3A57.305737380Z%22%20level%3Dinfo%20msg%3D%22Reloading%20configuration%22%20file%3D%22config/reload.go%3A21%22) 

`Configuration reload completed` log entry and returns early.

haircommander:
cc @littlejawa

roman-kiselenko:
@haircommander 

I think the `kata` itself does nothing here. The problem is that the test waits log message `Configuration reload completed` two times and, in both cases catches only the first entry. 

I'm going to change a test to reflect both reloads

littlejawa:
Took a quick look at it but from what I understand, the changes and the test itself are independent of the runtime. Kata is not involved here.
If there is a difference of behavior with kata, it must be because of the environment variables and config file settings that are added for kata in the CI job itself. They can conflict with other assumptions made by the test scripts.
I've been hurt by problems like that before, so I can help troubleshoot. But the bottom line is that there's no reason for "kata" to make a difference here.

roman-kiselenko:
It looks like I catch https://github.com/kubernetes-sigs/cri-tools/pull/1429 

I'm going to wait for a revert.

hexfusion:
@roman-kiselenko thanks for the progress its unclear what needs to be reverted are you saying this[1] pr?

[1] https://github.com/kubernetes-sigs/cri-tools/pull/1429

roman-kiselenko:
> @roman-kiselenko thanks for the progress its unclear what needs to be reverted are you saying this[1] pr?
> 
> [1] [kubernetes-sigs/cri-tools#1429](https://github.com/kubernetes-sigs/cri-tools/pull/1429)

I was waiting for this [PR](https://github.com/kubernetes-sigs/cri-tools/pull/1440) to be merged. 

/unhold

roman-kiselenko:
@cri-o/cri-o-maintainers PTAL

haircommander:
/approve

thank you!
/cherry-pick release-1.30

roman-kiselenko:
> hmm... it looks the test is a bit flaky on `fedora-kata` https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8213/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1802693012262752256/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt

AFAIK it's related to Kata, should we disable it?

littlejawa:
> > hmm... it looks the test is a bit flaky on `fedora-kata` https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8213/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1802693012262752256/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt
> 
> AFAIK it's related to Kata, should we disable it?

I'm still not sure about that.
This is all about crio config, and kata is not involved there.

From the logs pointed by the link above, the test is running fine with kata until the end, including reloading the config with no pinned images. The test fails when doing a "crictl images" and seeing the image is still marked as pinned.

Looking up the logic of the test, and the workflow of the code, I think there is a possible race condition:
- In the [reload go routine](https://github.com/cri-o/cri-o/blob/e34116dd2ec84de7b000832003ca86684b53577c/server/server.go#L579), we call "s.config.Reload()", then "s.storageImageServer().UpdatePinnedImagesList()"
  The first will reload the config, and generates the log messages that the test is expecting.
  The second will update the regexp that we use to mark images as pinned or not when we do ListImages or ImageStatus.

- The test is expecting the log "pinned_images = []", which is generated by "Reload()". When it finds it, the next thing it does is to call "crictl images".

I think that if somehow the call to UpdatePinnedImagesList() is not done by the time the crictl call is answered, you can get the wrong result.

As part of the reload (still according to the logs above), the runtime config is reloaded after the pinned images list. Since we're adding a runtime with kata, the timing for the reload function must be *slightly* different than without kata. I actually see that after the ListImages log (wich is the answer to your "crictl images" call), we have some logs showing that the runtime config is still being processed by the "Reload()" function.
I think this explains the error that I see in the logs, and why it is flaky.

Again, this is not kata-specific. If it is a timing condition, any non-vanilla config file could generate the same error.

Could you add a wait_for_log call in your test, and wait for the reload function's "Configuration reload completed" message before doing your "crictl" calls?
If I'm not wrong, this should synchronize everything.

roman-kiselenko:
@littlejawa 

> Could you add a wait_for_log call in your test, and wait for the reload function's "Configuration reload completed" message before doing your "crictl" calls?

I've added `wait_for_log "Configuration reload completed"` but my assumption is a log entry 
 `"Configuration reload completed"` occurs several times and the test finds the first entry.

littlejawa:
> @littlejawa
> 
> > Could you add a wait_for_log call in your test, and wait for the reload function's "Configuration reload completed" message before doing your "crictl" calls?
> 
> I've added `wait_for_log "Configuration reload completed"` but my assumption is a log entry `"Configuration reload completed"` occurs several times and the test finds the first entry.

Ah... you may be right :-( I'm not sure how "wait_for_log" works.
Maybe not the right fix then. But I'm pretty sure this is the root cause.

I hate saying this, but sometimes "sleep" calls helps :-p

sohankunkerkar (approved):
CI looks green now
/lgtm

pkg/config/reload.go
```diff
@@ -145,24 +147,34 @@ func (c *Config) ReloadPauseImage(newConfig *Config) error {
...
-// The method print log in case of any updates.
+// ReloadPinnedImages replace the PinnedImages
+// with the provided `newConfig.PinnedImages`.
+// The method skips empty items and prints a log message.
 func (c *Config) ReloadPinnedImages(newConfig *Config) {
-	updatedPinnedImages := make([]string, len(newConfig.PinnedImages))
-	updatedPinnedImageList := false
+	if len(newConfig.PinnedImages) == 0 {
+		c.PinnedImages = []string{}
+		logConfig("pinned_images", "[]")
```
kwilczynski, line 156: Ah nice catch. Makes sense :+1:

kwilczynski:
@cri-o/cri-o-maintainers, we could use some overrides here. Thank you!

littlejawa:
> @cri-o/cri-o-maintainers, we could use some overrides here. Thank you!

Following our discussion above, the test may still be flaky. The use of "wait_for_log" on the "Configuration reload completed" message may not be reliable...
Please review carefully. Any advice is welcome to make sure the test doesn't try to get image status before the config reload is complete.

roman-kiselenko:
> > @cri-o/cri-o-maintainers, we could use some overrides here. Thank you!
> 
> Following our discussion above, the test may still be flaky. The use of "wait_for_log" on the "Configuration reload completed" message may not be reliable... Please review carefully. Any advice is welcome to make sure the test doesn't try to get image status before the config reload is complete.

I've added a `sleep 0.5` point, let's try with it. 🤞

kwilczynski:
> > @cri-o/cri-o-maintainers, we could use some overrides here. Thank you!
> 
> Following our discussion above, the test may still be flaky.
[...]
> Please review carefully.

@littlejawa, no worries. I was under the impression that we were able to resolve issues here already.

kwilczynski:
/hold

To resolve issues with flaky tests.

sohankunkerkar:
I think we need to unblock this work by disabling the test on `fedora-kata` and filing a tracking issue for it.

roman-kiselenko:
> I think we need to unblock this work by disabling the test on `fedora-kata` and filing a tracking issue for it.

SGTM @cri-o/cri-o-maintainers There is an umbrella [issue](https://github.com/cri-o/cri-o/issues/8074).

sohankunkerkar:
> > I think we need to unblock this work by disabling the test on `fedora-kata` and filing a tracking issue for it.
> 
> SGTM @cri-o/cri-o-maintainers There is an umbrella [issue](https://github.com/cri-o/cri-o/issues/8074).

@roman-kiselenko I'm not sure how it's related to the above issue. Could you create a separate one? 
Also, you need to skip this test on `fedora-kata`.

roman-kiselenko:
@sohankunkerkar, I've skipped the flaky test on kata. 

I will create a separate issue after the successful merge.

kwilczynski:
@roman-kiselenko, thank you!

Let's get this merged and we can fix/investigate the other issues separately, indeed.

PR #8225, opened 2024-05-28
server: use SecureJoin when setting container /etc directory

haircommander:

```release-note
Fix CVE-2024-5154 where a malicious container image could make a symlink of `/proc/mounts` on the host, out of the container's rootfs
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -846,25 +846,38 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		specgen.Config.Process.User.Umask = &umask
 	}
 
-	etc := filepath.Join(mountPoint, "/etc")
-	// create the `/etc` folder only when it doesn't exist
-	if _, err := os.Stat(etc); err != nil && os.IsNotExist(err) {
+	// The /etc directory can be subjected to various attempts on the path (directory)
+	// traversal attacks. As such, we need to ensure that its path will be relative to
+	// the base (or root, if you wish) of the container to mitigate a container escape.
+	etcPath, err = securejoin.SecureJoin(mountPoint, "/etc")
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve container /etc directory path: %w", err)
+	}
+
+	// Create the /etc directory only when it doesn't exist.
+	if _, err := os.Stat(etcPath); err != nil && os.IsNotExist(err) {
 		rootPair := idtools.IDPair{UID: 0, GID: 0}
 		if containerIDMappings != nil {
 			rootPair = containerIDMappings.RootPair()
 		}
-		if err := idtools.MkdirAllAndChown(etc, 0o755, rootPair); err != nil {
-			return nil, fmt.Errorf("error creating mtab directory: %w", err)
+		if err := idtools.MkdirAllAndChown(etcPath, 0o755, rootPair); err != nil {
+			return nil, fmt.Errorf("failed to create container /etc directory: %w", err)
 		}
 	}
-	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
-	// compatible with Docker
-	if err := os.Symlink("/proc/mounts", filepath.Join(etc, "mtab")); err != nil && !os.IsExist(err) {
+
+	// Add a symbolic link from /proc/mounts to /etc/mtab to keep compatibility with legacy
+	// Linux distributions and Docker.
+	//
+	// We cannot use SecureJoin here, as the /etc/mtab can already be symlinked from somewhere
+	// else in some cases, and doing so would resolve an existing mtab path to the symbolic
+	// link target location, for example, the /etc/proc/self/mounts, which breaks container
+	// creation.
+	if err := os.Symlink("/proc/mounts", filepath.Join(etcPath, "mtab")); err != nil && !os.IsExist(err) {
 		return nil, err
 	}
 
 	// Configure timezone for the container if it is set.
-	if err := configureTimezone(s.Runtime().Timezone(), ociContainer.BundlePath(), mountPoint, mountLabel, etc, ociContainer.ID(), options, ctr); err != nil {
+	if err := configureTimezone(s.Runtime().Timezone(), ociContainer.BundlePath(), mountPoint, mountLabel, etcPath, ociContainer.ID(), options, ctr); err != nil {
 		return nil, fmt.Errorf("failed to configure timezone for container %s: %w", ociContainer.ID(), err)
 	}
 
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -846,6 +846,19 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
 		specgen.Config.Process.User.Umask = &umask
 	}
 
+	etcPath := filepath.Join(mountPoint, "/etc")
+
+	// Warn users if the container /etc directory path points to a location
+	// that is not a regular directory. This could indicate that something
+	// might be afoot.
+	etc, err := os.Lstat(etcPath)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+	if err == nil && !etc.IsDir() {
+		log.Warnf(ctx, "Detected /etc path for container %s is not a directory", ctr.ID())
+	}
+
 	// The /etc directory can be subjected to various attempts on the path (directory)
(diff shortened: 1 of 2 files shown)


haircommander:
@eriksjolund  @kwilczynski can you check the newest version? `/etc/mtab` was already being linked somewhere somehow, and so ci was failing. this should cover that

eriksjolund:
LGTM

kwilczynski:
> @eriksjolund @kwilczynski can you check the newest version? `/etc/mtab` was already being linked somewhere somehow, and so ci was failing. this should cover that

Let me look into this. The code isn't that fundamentally different, so the new paths must differ quite a bit.

server/container_create_linux.go
```diff
@@ -846,25 +846,52 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
-		return nil, err
+
+	// we cannot use SecureJoin here, as /etc/mtab is already being symlinked from somewhere in some cases
+	// and doing so resolves mtabPath to mountPoint/etc/proc/self/mounts, which breaks container creation.
+	// Let's just check if mtab is created already, and if not, symlink it.
+	mtabPath := filepath.Join(etcPath, "mtab")
+	if _, err := os.Stat(mtabPath); err != nil && os.IsNotExist(err) {
+		// Add symbolic link from /proc/mounts to /etc/mtab to keep compatibility with
+		// legacy Linux distributions and Docker.
+		if err := os.Symlink("/proc/mounts", mtabPath); err != nil && !os.IsExist(err) {
```
kwilczynski, line 888: We can probably now drop the following:

```suggestion
		if err := os.Symlink("/proc/mounts", mtabPath); err != nil {
```

Since we only add the symbolic link now when it does not exist.

server/container_create_linux.go
```diff
@@ -846,25 +846,52 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
 		}
 	}
-	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
-	// compatible with Docker
-	if err := os.Symlink("/proc/mounts", filepath.Join(etc, "mtab")); err != nil && !os.IsExist(err) {
-		return nil, err
+
+	// we cannot use SecureJoin here, as /etc/mtab is already being symlinked from somewhere in some cases
+	// and doing so resolves mtabPath to mountPoint/etc/proc/self/mounts, which breaks container creation.
+	// Let's just check if mtab is created already, and if not, symlink it.
```
kwilczynski, line 883: This is an interesting side effect of using `securejoin.SecureJoin()` since it will try to resolve to underlying symbolic link.

For example, our problematic container image `quay.io/crio/fedora-crio-ci:latest` already has the following:

```console
lrwxrwxrwx 1 root root         19 Aug 10  2023 mtab -> ../proc/self/mounts
```

Which cases the `SecureJoin()` function to resolve it to (an example from a failed run):

```go
(string) (len=111) "/var/lib/containers/storage/overlay/ef720b916452ac083f3c8438627d9f5ca0f8180c5142a5befcb40cb2f80045ca/merged/etc"
(string) (len=116) "/var/lib/containers/storage/overlay/ef720b916452ac083f3c8438627d9f5ca0f8180c5142a5befcb40cb2f80045ca/merged/etc/mtab"
(string) (len=128) "/var/lib/containers/storage/overlay/ef720b916452ac083f3c8438627d9f5ca0f8180c5142a5befcb40cb2f80045ca/merged/etc/proc/self/mounts"
```

However, when run under Docker (moby), the /etc/mtab symbolic link will be recreated per:

```console
lrwxrwxrwx 1 root root      12 May 29 08:11 mtab -> /proc/mounts
```

This is because of the following code:

- https://github.com/moby/moby/blob/master/daemon/initlayer/setup_unix.go#L19-L72

I wonder whether it would be safer to also remove existing and add new /etc/mtab symbolic link? The procfs nowadays creates /proc/mounts as symbolic link to /proc/self/mounts anyway (see [man 5 proc_pid_mounts](https://manpages.debian.org/unstable/manpages/proc_pid_mounts.5.en.html)), and as such, it would be safe to always create it from scratch.

server/container_create_linux.go
eriksjolund: Oh, I first misread that line. I missed the `!`. I think this would make it even more explicit:

```
if err := os.Symlink("/proc/mounts", mtabPath); err != nil && os.IsExist(err) 
```

server/container_create_linux.go
haircommander: hm maybe we should remove the symlink in the image as it may actually be bogus and then readd the change?

kwilczynski: @haircommander that is what I am suggesting too.

Per the everpresent documentation, the `/etc/mtab`, if present, should be a symbolic link to `/proc/mounts`. There is no need for it to be different, I suppose. This is also to retain some backwards compatibility with older software and Linux distributions, plus to retain what is expected to be a canonical (not to say, historical even) thing for this specific file/symbolic link.

haircommander: okay so I did some investigation, and all the fedora family creates this symlink. It's not present everywhere or magically injected, as the ubuntu image doesn't have it. thus, I think the code as-is is correct

server/container_create_linux.go
haircommander: good catch

server/container_create_linux.go
kwilczynski: Yeah, the code is fine. :+1: 

What I wanted to bring up, was that if we remove it, no matter what it is and where it's currently pointing, and add it ourselves, then we can always guarantee its correctness.

Albeit, leaving it alone is also an option.

kwilczynski:
/approve
/lgtm

Thank you for help @haircommander and @eriksjolund :tada:

server/container_create_linux.go
```diff
@@ -846,25 +846,48 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
 		}
 	}
-	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
-	// compatible with Docker
-	if err := os.Symlink("/proc/mounts", filepath.Join(etc, "mtab")); err != nil && !os.IsExist(err) {
+
+	// Add symbolic link from /proc/mounts to /etc/mtab to keep compatibility with
+	// legacy Linux distributions and Docker.
+	// we cannot use SecureJoin here, as /etc/mtab is already being symlinked from somewhere in some cases
+	// and doing so resolves mtabPath to mountPoint/etc/proc/self/mounts, which breaks container creation.
```
kwilczynski, line 884:
```suggestion
	// Add symbolic link from /proc/mounts to /etc/mtab to keep compatibility with legacy Linux
	// distributions and Docker. We cannot use SecureJoin here, as the /etc/mtab can already be
	// symlinked from somewhere else in some cases and doing so resolves mtabPath to the
	// symbolic link target location, such  as, for example, /etc/proc/self/mounts, which breaks
	// container creation.
```

server/container_create_linux.go
```diff
@@ -846,25 +846,49 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+			return nil, fmt.Errorf("failed to create container /etc directory: %w", err)
 		}
 	}
-	// add symlink /etc/mtab to /proc/mounts allow looking for mountfiles there in the container
-	// compatible with Docker
-	if err := os.Symlink("/proc/mounts", filepath.Join(etc, "mtab")); err != nil && !os.IsExist(err) {
+
+	// Add a symbolic link from /proc/mounts to /etc/mtab to keep compatibility with legacy
+	// Linux distributions and Docker. We cannot use SecureJoin here, as the /etc/mtab can
+	// already be symlinked from somewhere else in some cases, and doing so resolves mtabPath
```
eriksjolund, line 883: The word _mtabPath_ is not used elsewhere. (The meaning can probably be guessed though)

kwilczynski: Good point. We don't have this variable any longer. Let me fix that.

kwilczynski: Done.

kwilczynski:
We would need to squash my code comment changes to rid of them, and then rebase when the following will be merged:

- https://github.com/cri-o/cri-o/pull/8226

To fix the unit tests failures.

sohankunkerkar (approved):
/lgtm

sohankunkerkar:
Let's get this PR in as it addresses a CVE. We can expect the unit test failures to disappear once the rebase happens

/unhold

PR #8240, opened 2024-06-03
Fix container stats label filter

saschagrunert:
#### What type of PR is this?

/kind bug

We need to filter the labels for each container to be able to support the corresponding cri-tools feature.

Fixes: https://github.com/kubernetes-sigs/cri-tools/issues/950

None

```release-note
Fixed container stats label filtering.
```

### DIFF
--- a/server/container_stats_list.go
+++ b/server/container_stats_list.go
@@ -16,14 +16,21 @@ func (s *Server) ListContainerStats(ctx context.Context, req *types.ListContaine
 	if err != nil {
 		return nil, err
 	}
-	filter := req.Filter
-	if filter != nil {
+	if req.Filter != nil {
 		cFilter := &types.ContainerFilter{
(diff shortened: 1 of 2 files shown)


saschagrunert:
/retest 

@cri-o/cri-o-maintainers PTAL

kwilczynski:
@saschagrunert, looks good!

An idea: we could also rid of the `filter` variable on the line 19 while at it... I think it was intended to be used with in the scope of the if-statement, but that never seems to have happened. Have a look... it might be a nice clean-up too.

Screenshot from 2024-06-04 01-57-56

saschagrunert:
> we could also rid of the `filter` variable on the line 19 while at it... I think it was intended to be used with in the scope of the if-statement, but that never seems to have happened. Have a look... it might be a nice clean-up too.

Sure, fixed as suggested.

PR #6547, opened 2023-01-18
server: only start one resource store routine on network cleanup in r…

haircommander:
…estore

to reduce the amount of CPU usage of many routines spinning and waiting for the CNI plugin to become ready

Signed-off-by: Peter Hunt~ <pehunt@redhat.com>

#### What type of PR is this?

/kind bug
#### Does this PR introduce a user-facing change?

```release-note
none
```

PR #8245, opened 2024-06-03
Call network plugin GC on startup

jcaamano:
#### What type of PR is this?

/kind feature

As a last measure to clean up potential stale network resources, call
the network plugin GC on startup referencing pods which were restored
successfully which should clean up for pods that could not be restored.

```release-note
Call network plugin GC on startup to attempt cleaning up stale network 
resources of pods that could not be restored
```


jcaamano:
@squeed PTAL

server/server.go
```diff
@@ -280,7 +285,13 @@ func (s *Server) restore(ctx context.Context) []storage.StorageImageID {
 		}
 	}
 
-	// Cleanup the deletedPods in the networking plugin
+	// First cleanup any potential stale network resources not associated to any
+	// pod known to us
+	if err := s.networkGC(ctx, knownPods); err != nil {
```
haircommander, line 290: so unfortunately this will run before the networking plugin has a chance to startup, which is why the other pieces are wrapped in the resource cleaner below. 

can you adapt https://github.com/cri-o/cri-o/pull/6547 into here, and call the GC first, and then loop through the pods all in one routine?

jcaamano: Thanks, wasn't aware of that.

The extra complication is that we can't be attending new pod requests as the list of valid pods needs to be stable while we call GC. I will see how we can go about this.

haircommander: an option is you can create all of the pods that are hostNetwork, then run GC, then create the rest?

jcaamano: @haircommander 

I went with a (maybe naive) option of calling GC in the background while holding a lock that prevents running non host network pods in the meantime.

I didn't adopt your code as there was no good reason for the specific order, both GC and DEL should be able to run parallel.

jcaamano: I could adopt your could though if that still helps.

haircommander:
another option which I find myself leaning towards is just adding the GC call here: https://github.com/cri-o/cri-o/blob/main/internal/config/cnimgr/cnimgr.go#L50 before we notify the watching routines their plugin is ready, we call the GC. functionally similar, but we don't need to introduce another lock in the mix. WDYT?

jcaamano:
> another option which I find myself leaning towards is just adding the GC call here: https://github.com/cri-o/cri-o/blob/main/internal/config/cnimgr/cnimgr.go#L50 before we notify the watching routines their plugin is ready, we call the GC. functionally similar, but we don't need to introduce another lock in the mix. WDYT?

Adding the call there seems convenient but having the necessary context available to come up with the list of valid pods is less so. We can fail the readiness until someone sets a list of valid pods to cnimgr to call GC with (or informs cnimgr that the GC call has been made). Let me know if you had that in mind.

haircommander:
hm good point. What if you added a call to cnimgr that took a list of stale pods that `server/server.go` produces, and calls GC on them? Such a function could first do the same check as ReadyOrError(), if the plugin is ready call GC immediately, if not save the pods, and then GC could be called right before the watchers are notified. 

another thing to note: cri-o does support restarting and picking up pods from the previous run. in that case, I don't think we want to GC (as the pods are valid). we only want to call GC if we follow up by deleting the pods

jcaamano:
> hm good point. What if you added a call to cnimgr that took a list of stale pods that `server/server.go` produces, and calls GC on them? Such a function could first do the same check as ReadyOrError(), if the plugin is ready call GC immediately, if not save the pods, and then GC could be called right before the watchers are notified.
> 
> another thing to note: cri-o does support restarting and picking up pods from the previous run. in that case, I don't think we want to GC (as the pods are valid). we only want to call GC if we follow up by deleting the pods

ok, let me give that a go.

A clarification: 

GC is called with a list of valid pods and CNI cleans everything else. As long as we have a stable list of valid pods, we could potentially call GC. On startup, I consider the stale pods for which CNI DEL is going to be called as valid in the context of the GC call. In general, basically any pod for which the network can still be managed in one way or another should be considered valid in terms of GC call and the GC call is no substitute for CNI DEL.

jcaamano:
@haircommander ready for another round

internal/config/cnimgr/cnimgr.go
```diff
@@ -90,3 +99,31 @@ func (c *CNIManager) Shutdown() {
...
+	defer c.mutex.Unlock()
+	if c.lastError != nil {
+		c.validPodList = validPodList
+		return nil
+	}
+	return c.doGC(ctx, validPodList)
+}
+
+func (c *CNIManager) doGC(ctx context.Context, validPodList PodNetworkLister) error {
+	if validPodList != nil {
```
haircommander, line 118: nit: this could be inverted and then the rest of the function would be less indented

haircommander:
I like this better! structurally it LGTM, just a couple of nits

internal/config/cnimgr/cnimgr.go
```diff
@@ -90,3 +99,31 @@ func (c *CNIManager) Shutdown() {
...
+// will be deferred until it is ready logging any errors then and returning nil
+// error here.
+func (c *CNIManager) GC(ctx context.Context, validPodList PodNetworkLister) error {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	if c.lastError != nil {
+		c.validPodList = validPodList
+		return nil
+	}
+	return c.doGC(ctx, validPodList)
```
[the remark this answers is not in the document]
jcaamano: Yeah, we can. I was just being overly jealous I imagine not setting it if it was not going to be used (because the plugin was actually ready). Will change it around.

haircommander:
can you add a note above each doGC call that says something like:
```
This call is and should be duplicated. We need to account for plugins that are ready at cri-o startup, and ones that need to wait until a pod is created. We could plumb in additional code to wait for a valid list of pods, but that would further complicate the code. This is safe to do because CNI GC is idempotent.
```
or something like that (assuming the last sentence is true which I believe to be the case)? I had to muscle through that logically and I'd like to save myself from doing so again in the future

jcaamano:
> can you add a note above each doGC call that says something like:
> 
> ```
> This call is and should be duplicated. We need to account for plugins that are ready at cri-o startup, and ones that need to wait until a pod is created. We could plumb in additional code to wait for a valid list of pods, but that would further complicate the code. This is safe to do because CNI GC is idempotent.
> ```
> 
> or something like that (assuming the last sentence is true which I believe to be the case)? I had to muscle through that logically and I'd like to save myself from doing so again in the future

I added comments, but something different to what you suggested. Let me know if it helps. On your version, I was a bit confused with `ones that need to wait until a pod is created` and `safe to do because CNI GC is idempotent`, not that it wouldn't be true but I was not clear to me on the cause/effect relationship with the duplicated call in the code.

jcaamano:
Rebased and changed bump to libcni to v1.2.2 which is suposed to fix a problem with one of the integration tests here (from https://github.com/containernetworking/cni/pull/1095), this might replace the bump effort here https://github.com/containernetworking/cni/pull/1095

jcaamano:
@haircommander can you clue me in on what might be failing on the CI jobs/builds? I looked but couldn't really find a thread to pull from.

haircommander:
[here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/8245/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1815375533416910848/artifacts/fedora-integration/cri-o-gather/artifacts/testout.txt) it looks like [this](https://github.com/cri-o/cri-o/blob/2851b0c9a76bab99943fd0a4b1a10035f50bbcc8/test/inspect.bats#L21) test

jcaamano:
@haircommander sorry it takes time for me to look at this, I have a ton of other work but I try to come back to this every now and then.

So on that test, even though the pod seems to be attached correctly
```
# time="2024-07-22 13:39:06.602569244Z" level=debug msg="CNI setup result: &{0.3.1 [{Name:cni0 Mac:3a:cd:7d:ac:24:45 Sandbox:} {Name:veth0bc103e5 Mac:be:fb:96:1b:3d:d2 Sandbox:} {Name:eth0 Mac:6e:e5:56:66:a2:4d Sandbox:/var/run/netns/92fde6de-5903-41e5-a285-17148f635ff8}] [{Version:4 Interface:0xc0005f3e78 Address:{IP:10.88.0.105 Mask:ffff0000} Gateway:10.88.0.1} {Version:6 Interface:0xc0005f3eb8 Address:{IP:1100:200::69 Mask:ffffff00000000000000000000000000} Gateway:1100:200::1}] [{Dst:{IP:0.0.0.0 Mask:00000000} GW:<nil> MTU:0 AdvMSS:0 Priority:0 Table:<nil> Scope:<nil>} {Dst:{IP:1100:200::1 Mask:ffffff00000000000000000000000000} GW:<nil> MTU:0 AdvMSS:0 Priority:0 Table:<nil> Scope:<nil>}] {[]  [] []}}" file="server/sandbox_network.go:76" id=23485206-bf2e-4f56-a980-7d95bc07b25b name=/runtime.v1.RuntimeService/RunPodSandbox
...
# time="2024-07-22 13:39:06.952739782Z" level=debug msg="Response: &CreateContainerResponse{ContainerId:7188d1342df9e33b5310596c2dca91a11a7025653b85c49dc8525d76814557c9,}" file="otel-collector/interceptors.go:75" id=9400ca3b-21db-4137-ab77-1474a20012c4 name=/runtime.v1.RuntimeService/CreateContainer
```
the test fails to grab the IPs
```
# time="2024-07-22T13:39:07Z" level=fatal msg="execing command in container synchronously: command 'ip -4 addr show dev eth0 scope global' exited with 1: Device \"eth0\" does not exist.\n"
# time="2024-07-22T13:39:07Z" level=fatal msg="execing command in container synchronously: command 'ip -6 addr show dev eth0 scope global' exited with 1: Device \"eth0\" does not exist.\n"
```
...
```
# time="2024-07-22 13:39:07.260794470Z" level=debug msg="Received container exit code: 1, message: " file="oci/runtime_oci.go:705" id=9f595776-5d95-4432-a803-c7d30f180590 name=/runtime.v1.RuntimeService/ExecSync
```

It failed in some platforms but not all. I don't see the relationship unless the bumps actually have something to do with it. I rebased and will check the new results, and ask around in the CNI upstream group.

jcaamano:
@haircommander do the `fedora-integration` tests run in parallel in any way? 

I see some failing test deleting something that does not seem spawned in that test, so it is possible that other tests are deleting stuff from the tests that fail.

```
# time="2024-08-12 12:48:26.892858300Z" level=error msg="GARBAGE COLLECT called on poll" file="cnimgr/cnimgr.go:63"
# time="2024-08-12 12:48:26.892952889Z" level=error msg="GARBAGE COLLECT called with valid pods []*ocicni.PodNetwork{}" file="cnimgr/cnimgr.go:137"
# time="2024-08-12 12:48:26.893004438Z" level=error msg="GARBAGE COLLECT for network crio" file="ocicni/ocicni.go:795"
# time="2024-08-12 12:48:26.893319772Z" level=error msg="GARBAGE COLLECT deleted a552cd5d85debcb4261d04397fea2d7f0b07a07c6a78223b1c690b9666e0284e" file="libcni/api.go:809"
# time="2024-08-12 12:48:26.995851235Z" level=error msg="GARBAGE COLLECT call done" file="cnimgr/cnimgr.go:142"
```

jcaamano:
> @haircommander do the `fedora-integration` tests run in parallel in any way?
> 
> I see some failing test deleting something that does not seem spawned in that test, so it is possible that other tests are deleting stuff from the tests that fail.
> 
> ```
> # time="2024-08-12 12:48:26.892858300Z" level=error msg="GARBAGE COLLECT called on poll" file="cnimgr/cnimgr.go:63"
> # time="2024-08-12 12:48:26.892952889Z" level=error msg="GARBAGE COLLECT called with valid pods []*ocicni.PodNetwork{}" file="cnimgr/cnimgr.go:137"
> # time="2024-08-12 12:48:26.893004438Z" level=error msg="GARBAGE COLLECT for network crio" file="ocicni/ocicni.go:795"
> # time="2024-08-12 12:48:26.893319772Z" level=error msg="GARBAGE COLLECT deleted a552cd5d85debcb4261d04397fea2d7f0b07a07c6a78223b1c690b9666e0284e" file="libcni/api.go:809"
> # time="2024-08-12 12:48:26.995851235Z" level=error msg="GARBAGE COLLECT call done" file="cnimgr/cnimgr.go:142"
> ```

I guess it might?
```
++ nproc --all
+ export JOBS=8
+ JOBS=8
+ bats --version
+ execute bats --jobs 8 --tap .
+ echo ++ bats --jobs 8 --tap .
++ bats --jobs 8 --tap .
+ bats --jobs 8 --tap .
```
testing without it

haircommander:
I believe each test should be sandboxed from the others but that isolation isn't perfect. but yeah they're run in parallel

haircommander:
I'm fairly certain integration-v2 tests are failing everywhere

jcaamano:
@haircommander so the two problems were
* uninitialized error in `cnimgr` causing GC to be called before it had to, code fix
* parallel tests garbage collecting interfaces of other tests

For that last issue, it looks like I will have to add a new config parameter for the [cache dir ](https://github.com/cri-o/ocicni/blob/745c5e82582632fa2dca5e3e2750dd98cde1ecc4/pkg/ocicni/ocicni.go#L211) so that it test can run with its independent cache, otherwise we would just have to disable GC for the tests.

haircommander:
I would also be okay disabling gc for tests for now. I'm open either way

haircommander:
hey @jcaamano are you able to rebase this? or would you like a cri-o maintainer to take it over?

jcaamano:
> hey @jcaamano are you able to rebase this? or would you like a cri-o maintainer to take it over?

I will try to wrap it up next week.

haircommander:
/approve

LGTM, thanks @jcaamano . ideally we'd have an integration test in `test/` but adding ones for cni plugins are a pain so I am fine to defer that. 

@cri-o/cri-o-maintainers PTAL, this could be a good candidate for 1.32

saschagrunert (approved):
/retest

PR #8251, opened 2024-06-03
server/*: Fix a bug where the GID is not added to /etc/group when run_as_group is set

### LINKED ISSUE #8216: Containers runs as a member of group should have its gid in /etc/group inside container
### What happened?

1. Create POD with below securitycontext  
securityContext:
    runAsUser: 64892
    runAsGroup: 6263
    fsGroup: 123

2. Inside container no group id is created at /etc/group
bash-5.2$ id
uid=64892(64892) gid=6263 groups=6263,123
bash-5.2$ id 64892
uid=64892(64892) gid=0(root) groups=0(root)
bash-5.2$ cat /etc/passwd | grep 64892
64892:x:64892:0:64892 user:/:/sbin/nologin
bash-5.2$ cat /etc/group | grep 6263
bash-5.2$

3.it leads to sysrepo starts failure inside container as it checks if group id exists at /etc/group 
https://github.com/sysrepo/sysrepo

if group id is appointed by runAsGroup, this group id should exist at /etc/group inside container.
just like uid case in below ticket
https://github.com/cri-o/cri-o/pull/2022

set runAsGroup explicitly and check if group id exists in /etc/group inside container.

_No response_

<details>

```console
$ crio --version
crio version 1.27.4-7.2.rhaos4.14.git082c52f.el9
Version:        1.27.4-7.2.rhaos4.14.git082c52f.el9
GitCommit:      unknown
GitCommitDate:  unknown
GitTreeState:   clean
GoVersion:      go1.20.12
Compiler:       gc
Platform:       linux/amd64
Linkmode:       dynamic
BuildTags:
  rpm_crashtraceback
  libtrust_openssl
  selinux
  seccomp
  exclude_graphdriver_btrfs
  containers_image_ostree_stub
LDFlags:           -compressdwarf=false -B 0x65f4e71070d8476fd679ce6e0b5a627a1d7db545 -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 '
SeccompEnabled:   true
AppArmorEnabled:  false

{
  "clientVersion": {
    "major": "1",
    "minor": "27",
    "gitVersion": "v1.27.4",
    "gitCommit": "f7b14a92857ad1f28be2e2eaec3ae8229d4c2761",
    "gitTreeState": "clean",
    "buildDate": "2024-04-15T20:46:22Z",
    "goVersion": "go1.20.12 X:strictfipsruntime",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "kustomizeVersion": "v5.0.1",
  "serverVersion": {
    "major": "1",
    "minor": "27",
    "gitVersion": "v1.27.12+7bee54d",
    "gitCommit": "656b6ddbf993fdc14b6e8cfeeb7e7eebc6c9a072",
    "gitTreeState": "clean",
    "buildDate": "2024-04-16T11:59:57Z",
    "goVersion": "go1.20.12 X:strictfipsruntime",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}



### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
ID="rhcos"
ID_LIKE="rhel fedora"
VERSION="414.92.202404162000-0"
VERSION_ID="4.14"
VARIANT="CoreOS"
VARIANT_ID=coreos
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux CoreOS 414.92.202404162000-0 (Plow)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::coreos"
DOCUMENTATION_URL="https://docs.openshift.com/container-platform/4.14/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT_VERSION="4.14"
REDHAT_SUPPORT_PRODUCT_VERSION="4.14"
OPENSHIFT_VERSION="4.14"
RHEL_VERSION="9.2"
OSTREE_VERSION="414.92.202404162000-0"
$ uname -a
Linux master0.hztt-vrf2-10-69-54-66.ocp.hz.nsn-rdnet.net 5.14.0-284.62.1.el9_2.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Apr 5 09:44:49 EDT 2024 x86_64 x86_64 x86_64 GNU/Linux


</details>


<details>

</details>
run at physical server
it's openshift from Redhat
oc version
Client Version: 4.14.0-202404151639.p0.gf7b14a9.assembly.stream-f7b14a9
Kustomize Version: v5.0.1
Server Version: 4.14.22
Kubernetes Version: v1.27.12+7bee54d

### PR DESCRIPTION
**Pull request by PannagaRao**
When securityContext of runAsGroup is added the
gid has to be added in /etc/group. Test case for
the same is also being added

#### What type of PR is this?

/kind bug

This PR has changes to add gid [value of RunAsGroup in SecurityContext] under /etc/group

Fixes #8216

```release-note
Fix a bug where the GID is not added to /etc/group when run_as_group is set
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -229,11 +229,19 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
 	}
 
 	genPasswd := true
+	genGroup := true
 	for _, mount := range specgen.Config.Mounts {
-		if mount.Destination == "/etc" ||
-			mount.Destination == "/etc/" ||
-			mount.Destination == "/etc/passwd" {
+		switch mount.Destination {
+		case "/etc", "/etc/":
 			genPasswd = false
+			genGroup = false
+		case "/etc/passwd":
+			genPasswd = false
+		case "/etc/group":
+			genGroup = false
+		}
+
+		if !genPasswd && !genGroup {
 			break
 		}
 	}
@@ -255,6 +263,28 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
 				Options:     []string{"rw", "bind", "nodev", "nosuid", "noexec"},
 			}
 			specgen.AddMount(mnt)
+		}
+	}
+	if genGroup {
+		if sc.RunAsGroup != nil {
+			gid = uint32(sc.RunAsGroup.Value)
+		}
+
+		// verify gid exists in containers /etc/group, else generate a group with the group entry
+		groupPath, err := utils.GenerateGroup(gid, rootfs, ctrRunDir)
+		if err != nil {
+			return err
+		}
+		if groupPath != "" {
+			if err := securityLabel(groupPath, mountLabel, false, false); err != nil {
+				return err
+			}
+			specgen.AddMount(rspec.Mount{
+				Type:        "bind",
+				Source:      groupPath,
+				Destination: "/etc/group",
+				Options:     []string{"rw", "bind", "nodev", "nosuid", "noexec"},
+			})
 		}
 	}
 
--- a/test/pod.bats
+++ b/test/pod.bats
@@ -390,3 +390,32 @@ function teardown() {
 	output=$(crictl exec --sync "$ctr_id" ls -ld /etc)
 	[[ "$output" == *"test test"* ]]
 }
+
+@test "verify RunAsGroup in container" {
+	start_crio
+
+	jq '
+    .linux.security_context.run_as_user = { value: 1000 }
+    | .linux.security_context.run_as_group = { value: 1001 }
+  ' "$TESTDATA"/sandbox_config.json > "$TESTDIR/modified_sandbox_config.json"
+
+	jq '
+    .linux.security_context.run_as_user = { value: 1000 }
+    | .linux.security_context.run_as_group = { value: 1002 }
+  ' "$TESTDATA"/container_sleep.json > "$TESTDIR/modified_container_sleep_config"
+
+	# Create a new pod using the modified sandbox configuration
+	pod_id=$(crictl runp "$TESTDIR/modified_sandbox_config.json")
+
+	# Create a new container within the pod using the modified container configuration
+	ctr_id=$(crictl create "$pod_id" "$TESTDIR/modified_container_sleep_config" "$TESTDIR/modified_sandbox_config.json")
+	crictl start "$ctr_id"
+
+	# Verify that the gid is present in the /etc/group file
+	exec_output=$(crictl exec "$ctr_id" cat /etc/group)
+	echo "$exec_output" | grep "x:1002" || fail "RunAsGroup ID 1002 not found in /etc/group"
+
+	# Clean up the pod and container
+	crictl stop "$ctr_id"
+	crictl stopp "$pod_id"
+}
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -179,56 +179,146 @@ func GetUserInfo(rootfs, userName string) (uid, gid uint32, additionalGids []uin
 // GeneratePasswd generates a container specific passwd file,
 // iff uid is not defined in the containers /etc/passwd
 func GeneratePasswd(username string, uid, gid uint32, homedir, rootfs, rundir string) (string, error) {
-	// if UID exists inside of container rootfs /etc/passwd then
-	// don't generate passwd
 	if _, err := GetUser(rootfs, strconv.Itoa(int(uid))); err == nil {
 		return "", nil
 	}
-	passwdFile := filepath.Join(rundir, "passwd")
-	originPasswdFile, err := securejoin.SecureJoin(rootfs, "/etc/passwd")
-	if err != nil {
-		return "", fmt.Errorf("unable to follow symlinks to passwd file: %w", err)
+
+	passwdFilePath, stat, err := secureFilePath(rootfs, "/etc/passwd")
+	if err != nil || stat.Size == 0 {
+		return "", err
 	}
-	var st unix.Stat_t
-	err = unix.Stat(originPasswdFile, &st)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return "", nil
-		}
-		return "", fmt.Errorf("unable to stat passwd file %s: %w", originPasswdFile, err)
-	}
-	// Check if passwd file is world writable.
-	if st.Mode&0o022 != 0 {
+
+	if checkFilePermissions(&stat, uid, stat.Uid) {
 		return "", nil
 	}
 
-	if uid == st.Uid && st.Mode&0o200 != 0 {
-		return "", nil
+	origContent, err := readFileContent(passwdFilePath)
+	if err != nil || origContent == nil {
+		return "", err
 	}
 
-	orig, err := os.ReadFile(originPasswdFile)
-	if err != nil {
-		// If no /etc/passwd in container ignore and return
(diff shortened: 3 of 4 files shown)


server/container_create.go
```diff
@@ -229,13 +229,17 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
 	}
 
 	genPasswd := true
+	genGroup := true
 	for _, mount := range specgen.Config.Mounts {
 		if mount.Destination == "/etc" ||
 			mount.Destination == "/etc/" ||
```
haircommander, line 235: mount.Destination == "/etc" would be a reason to set genGroup to false as well.
I suggest a switch statement:
```
switch mount.Destination {
case "/etc/":
genPasswd, genGroup = false
case "/etc/passwd":
genPasswd = false
case "/etc/group"
genCgroup = false
}
```
you may need a separate function though as break statements won't work in a switch (to break out of the loop). also: don't break out of the loop on only the /etc/passwd case, as a user may have mounted /etc/passwd and /etc/group but this code as-is would only catch the former.

server/container_create.go
```diff
@@ -229,13 +229,17 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
 	}
```
haircommander, line 229: I would expect https://github.com/cri-o/cri-o/pull/8251/files#diff-1ec6bb9cf833fd4ded46ce1feb6440f1d7532294e0445623f2f2e0bb2726ea63R226 to take the RunAsGroup field into account and then we could just rely on gid from this point on

server/container_create.go
```diff
@@ -258,6 +262,25 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
 		}
 	}
 
+	if genGroup && sc.RunAsGroup != nil {
```
haircommander, line 265: I think we should use gid here, assuming that we can take RunAsGroup into account for gid https://github.com/cri-o/cri-o/pull/8251/files#r1625011421

test/pod.bats
```diff
@@ -390,3 +390,33 @@ function teardown() {
 	output=$(crictl exec --sync "$ctr_id" ls -ld /etc)
 	[[ "$output" == *"test test"* ]]
 }
+
+@test "verify RunAsGroup in container" {
+  start_crio
+
+  jq '
+    .linux.security_context.run_as_user = { value: 1000 }
+    | .linux.security_context.run_as_group = { value: 1001 }
```
haircommander, line 399: doing this in the sandbox won't really change anything (the pause container may get it, but it won't do anything with it). I thinhk you can just use sandbox_config.json here

utils/utils.go
```diff
@@ -231,6 +231,88 @@ func GeneratePasswd(username string, uid, gid uint32, homedir, rootfs, rundir st
 	return passwdFile, nil
 }
 
+// GenerateGroup generates a container specific group file,
+// iff gid is not defined in the containers /etc/group
+func GenerateGroup(gid uint32, rootfs, rundir string) (string, error) {
```
haircommander, line 236: can you refactor the shared bits between these 4 functions (Generate* and Get*)?

server/container_create.go
PannagaRao: Will implement the suggestion

server/container_create.go
PannagaRao: Will make this change by setting gid to pick sc.RunAsGroup value (when not null)

server/container_create.go
kwilczynski: For the sake of completeness, the switch statement would be:

```go
	switch mount.Destination {
	case "/etc", "/etc/":
		genPasswd, genGroup = false, false
	case "/etc/passwd":
		genPasswd = false
	case "/etc/group":
		genGroup = false
	}
```

test/pod.bats
PannagaRao: Using different RunAsGroup values at pod and container level to highlight that container specific one gets picked.

utils/utils.go
```diff
@@ -179,56 +179,146 @@ func GetUserInfo(rootfs, userName string) (uid, gid uint32, additionalGids []uin
...
+	}
+
+	groupContent := fmt.Sprintf("%s%d:x:%d:\n", string(origContent), gid, gid)
+	groupFile := filepath.Join(rundir, "group")
+
+	return createAndSecureFile(groupFile, groupContent, os.FileMode(stat.Mode), int(stat.Uid), int(stat.Gid))
+}
+
+// checkFilePermissions checks file permissions to decide whether to skip file modification.
+func checkFilePermissions(stat unix.Stat_t, id, statId uint32) bool {
```
haircommander, line 241: nit: can these functions be ordered in the order they're called from the calling functions?

utils/utils.go
```diff
@@ -179,56 +179,146 @@ func GetUserInfo(rootfs, userName string) (uid, gid uint32, additionalGids []uin
...
+		return "", fmt.Errorf("failed to chown file: %w", err)
 	}
+	return filepath, nil
+}
 
-	orig, err := os.ReadFile(originPasswdFile)
+// GetGroup takes a containermount path and group name or ID and returns
+// a matching Group structure from /etc/group.  If it cannot locate a group
+// with the provided information, an ErrNoGroupEntries is returned.
+func GetGroup(containerMount, groupIDorName string) (*user.Group, error) {
```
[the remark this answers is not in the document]
PannagaRao: Done

server/container_create.go
```diff
@@ -229,11 +229,19 @@ func setupContainerUser(ctx context.Context, specgen *generate.Generator, rootfs
...
 			genPasswd = false
+			genGroup = false
+		case "/etc/passwd":
+			genPasswd = false
+		case "/etc/group":
+			genGroup = false
+		}
+
+		if !genPasswd && !genGroup {
 			break
```
haircommander, line 245: are we sure the "break" here breaks us out of the loop and not the switch statement? we may need a goto (even though I don't like how they look)

haircommander:
I think pulling latest and rebasing this PR should fix the test failures

It LGTM besides one remaining question (it wouldn't hurt to have it as-is)
/approve

haircommander:
@cri-o/cri-o-maintainers PTAL

sohankunkerkar (approved):
/lgtm

kwilczynski:
/hold

Please update both the commit subject and the title of this Pull Request such that it explains what is being added.

We already know that there are "changes" made here, since this is a Pull Request.

utils/utils.go
```diff
@@ -179,56 +179,146 @@ func GetUserInfo(rootfs, userName string) (uid, gid uint32, additionalGids []uin
...
+	origContent, err := readFileContent(passwdFilePath)
+	if err != nil || origContent == nil {
+		return "", err
+	}
+
+	if username == "" {
+		username = "default"
+	}
+	if homedir == "" {
+		homedir = "/tmp"
```
kwilczynski, line 204: Maybe it would be better to set this to `/nonexistent`?

@cri-o/cri-o-maintainers, thoughts?

kwilczynski:
OK. Will cherry-pick manually.

PR #8290, opened 2024-06-17
Restore container logs from checkpoint

rst0git:
#### What type of PR is this?

/kind bug

A container checkpoint includes a copy of the logs file. However, this file is currently not used during restore. As a result, the output of `kubectl logs` for a restored container is empty. This pull request extends the current functionality to enable restoring of container logs. This is particularly useful with applications such as Jupyter Notebooks that output an authentication token and access instructions in the container logs.

None

None

```release-note
Enabled restoring container logs from a checkpoint.
```

### DIFF
--- a/internal/lib/restore.go
+++ b/internal/lib/restore.go
@@ -3,6 +3,7 @@ package lib
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):

saschagrunert:
Thanks!

Issue #8030, opened 2024-04-17
Stopping a container blocks all further stop attempts for the same container


juanfresia:
### What happened?

We've noticed that with version 1.29.1 we no longer can issue multiple `crictl stop` commands for the same container. If there is a `StopContainer` operation active, any other `StopContainer` operation will timeout at the client side waiting for the server to take a lock that is hold by the initial stop container operation.

## Context

In a scenario where you have one container that is unwilling to die gracefully over a long period of time (e.g. a big database that ignores SIGTERM for a long period of time while closing connections), the following regression happens:

* Sending a `crictl stop -t 300 <container>`, will send a SIGTERM and after 300s send a SIGKILL. This command hangs as it should while we wait for the container to die gracefully or get forcefully terminated after 300s.
* The container ignores SIGTERM, and won't die on its own. So we need to wait for the full timeout.
* We decided that we don't want to wait for 300s, and we now want to issue `crictl stop -t 0 <container>` to immediately kill it.

Currently, the second command will not take effect, and `crictl` will get a context timeout from `crio` server.

The following logs show the symptoms :

```
$ crictl stop dd5d5449948de
E0417 09:57:18.003776  750032 remote_runtime.go:366] "StopContainer from runtime service failed" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded" containerID="dd5d5449948de"
FATA[0002] stopping the container "dd5d5449948de": rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

This behavior was not present on version 1.27.1 and multiple stop container operations can be issued.

### What did you expect to happen?

I'd expect to have the same behavior we did in 1.27; where the second `crictl stop` command goes through and manages to kill the container. The container runtime should honor all stopping requests, even if there is one currently waiting in a longer timeout, we should be able to stop a container forcefully even if we mistakenly or purposely tried with a longer timeout before.
I think this might be a regression introduced around the time of [this refactor](https://github.com/cri-o/cri-o/commit/be5bac87bb56bbc1b01807668006dd246cab173c).

### How can we reproduce it (as minimally and precisely as possible)?

To reproduce the issue, in a host running k8s with cri-o as container runtime:

1) Create the following pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: sleepy-dummy
spec:
  terminationGracePeriodSeconds: 3600
  containers:
    - name: sleepy-dummy
      image: docker.io/library/busybox
      command: ["sleep", "600"]
[...]

PR #8300, opened 2024-06-18
OCPNODE-2408: internal/oci: remove redundant ShouldBeStopped check for stopping containers

sohankunkerkar:
#### What type of PR is this?

/kind bug

This commit removes the `ShouldBeStopped` function to address the issue where stopping container blocks all further stop attempts for the same container. That function acquired the `opLock` mutex to check the container's state, which could lead to contention and potential deadlocks if other goroutines were waiting to acquire the same lock.

Additionally, the container's state is checked later in the `StopContainer` function by calling the `Living` method, which checks if the container's process is still running. Therefore, the `ShouldBeStopped` function is redundant and can be removed.

Fixes https://github.com/cri-o/cri-o/issues/8030.

None

```release-note
Fixed a bug where stopping a container would block all further stop attempts for the same container.
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -868,6 +868,11 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 	ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
 
 	c.opLock.Lock()
+	if c.state.Status == ContainerStatePaused {
+		if _, err := r.runtimeCmd("resume", c.ID()); err != nil {
+			log.Errorf(ctx, "Failed to unpause container %s: %v", c.Name(), err)
+		}
+	}
 
 	// Begin the actual kill.
 	if _, err := r.runtimeCmd("kill", c.ID(), c.GetStopSignal()); err != nil {
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -831,13 +831,6 @@ func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout in
 		return nil
 	}
 
-	if err := c.ShouldBeStopped(); err != nil {
-		if errors.Is(err, ErrContainerStopped) {
-			err = nil
-		}
-		return err
-	}
-
 	// The initial container process either doesn't exist, or isn't ours.
 	if err := c.Living(); err != nil {
 		c.state.Finished = time.Now()
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -868,6 +868,14 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 	ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
 
 	c.opLock.Lock()
+	defer func() {
+		// Kill the exec PIDs after the main container to avoid pod lifecycle regressions:
+		// Ref: https://github.com/kubernetes/kubernetes/issues/124743
+		c.KillExecPIDs()
+		c.state.Finished = time.Now()
+		c.opLock.Unlock()
+		c.SetAsDoneStopping()
+	}()
 	if c.state.Status == ContainerStatePaused {
 		if _, err := r.runtimeCmd("resume", c.ID()); err != nil {
 			log.Errorf(ctx, "Failed to unpause container %s: %v", c.Name(), err)
@@ -880,8 +888,6 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 			// The initial container process either doesn't exist, or isn't ours.
 			// Set state accordingly.
 			c.state.Finished = time.Now()
-			c.opLock.Unlock()
-			c.SetAsDoneStopping()
 			return
 		}
 	}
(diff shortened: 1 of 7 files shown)


internal/oci/runtime_vm.go
```diff
@@ -669,6 +669,27 @@ func (r *runtimeVM) StopContainer(ctx context.Context, c *Container, timeout int
...
+// stopping. It determines if stopping the container makes sense
+// based on its current state. A container cannot be stopped if
+// it is already stopped or paused. If the container is paused,
+// the function attempts to unpause it and update its status.
+func (r *runtimeVM) shouldBeStopped(ctx context.Context, c *Container) error {
+	switch c.State().Status {
+	case ContainerStateStopped: // no-op
+		return ErrContainerStopped
+	case ContainerStatePaused:
+		logrus.Warnf("Cannot stop paused container")
```
haircommander, line 682: nit: we could use the `log` package here since we have a ctx

internal/oci/runtime_oci.go
```diff
@@ -824,19 +824,21 @@ func (r *runtimeOCI) StopContainer(ctx context.Context, c *Container, timeout in
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
+	if c.State().Status == ContainerStatePaused {
```
haircommander, line 827: this suffers the same way as the existing code, we need to move this check after we take the `opLock` in `StopLoopForContainer`

sohankunkerkar:
@juanfresia This PR should fix [this](https://github.com/cri-o/cri-o/issues/8030) issue. Sorry, it took some time.

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

haircommander:
/approve

LGTM, thanks for investigating here! @cri-o/cri-o-maintainers (specifically @kwilczynski as you've poked this code lately) PTAL

kwilczynski:
/hold

So I can have a pass at code review.

internal/oci/runtime_vm.go
```diff
@@ -670,6 +670,27 @@ func (r *runtimeVM) StopContainer(ctx context.Context, c *Container, timeout int
...
 }
 
+// shouldBeStopped checks whether the container's state permits
+// stopping. It determines if stopping the container makes sense
+// based on its current state. A container cannot be stopped if
+// it is already stopped or paused. If the container is paused,
+// the function attempts to unpause it and update its status.
+func (r *runtimeVM) shouldBeStopped(ctx context.Context, c *Container) error {
+	switch c.State().Status {
+	case ContainerStateStopped: // no-op
```
kwilczynski, line 680: Returning an error is not a no-op. Skipping it or allowing a fallthrough would be.

If this needs to be treated as such, then it might warrant a better comment.

kwilczynski:
How did we go from removing a redundant `ShouldBeStopped()` function there to refactoring the main stop loop?

server/container_stop.go
```diff
@@ -49,15 +49,6 @@ func (s *Server) stopContainer(ctx context.Context, ctr *oci.Container, timeout
...
 	}
 
-	if ctr.StateNoLock().Status == oci.ContainerStatePaused {
-		if err := s.Runtime().UnpauseContainer(ctx, ctr); err != nil {
-			return fmt.Errorf("failed to stop container %s: %w", ctr.Name(), err)
-		}
-		if err := s.Runtime().UpdateContainerStatus(ctx, ctr); err != nil {
-			return fmt.Errorf("failed to update container status %s: %w", ctr.Name(), err)
-		}
-	}
```
kwilczynski, line 59: Why move these? Was there a problem with these being here?

internal/oci/runtime_oci.go
```diff
@@ -937,30 +939,33 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
...
-
-			// Reschedule the timer so that the periodic reminder can continue.
-			blockedTimer.Reset(stopProcessBlockedInterval)
+			goto killContainer
 
 		case <-done:
 			stop()
+			return
+		case <-ctx.Done():
+			return
```
kwilczynski, line 948: When the function returns here, will the KillExecPIDs() et al still take place? This includes updating the final timestamp, etc.

sohankunkerkar: Yeah, that's a good point. I think I will put the last two lines in the defer statement, that way we can ensure that these operations are always executed.

sohankunkerkar (commented):
>How did we go from removing a redundant ShouldBeStopped() function there to refactoring the main stop loop?

Removing the redundant `ShouldBeStopped()` function helped in addressing the issue of blocking multiple stop calls. However, this change caused other BATS tests to time out, specifically `ctr stop timeouts should decrease` and `ctr stop loop kill retry attempts`. These tests were timing out due to the retry loop. Therefore, the code was refactored to separate the processing of the timeout logic from the backoff loop.

kwilczynski:
> > How did we go from removing a redundant ShouldBeStopped() function there to refactoring the main stop loop?
> 
> Removing the redundant `ShouldBeStopped()` function helped in addressing the issue of blocking multiple stop calls. However, this change caused other BATS tests to time out, specifically `ctr stop timeouts should decrease` and `ctr stop loop kill retry attempts`. These tests were timing out due to the retry loop. Therefore, the code was refactored to separate the processing of the timeout logic from the backoff loop.

@sohankunkerkar, thank adds a lot of context I was missing. Thank you!

So, if I understand this accurately, the retry loop was in the way, so to speak, of the tests completing in a timely manner, so with the current change, it was factored out to be a standalone code path once the kill timeout kicks in, correct?

sohankunkerkar:
>So, if I understand this accurately, the retry loop was in the way, so to speak, of the tests completing in a timely manner, so with the current change, it was factored out to be a standalone code path once the kill timeout kicks in, correct?

Yes, that's correct. Separating the timeout logic into a standalone code path ensures that the kill timeout is processed independently.

sohankunkerkar:
The test failure is not related to this PR.

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

PR #8317, opened 2024-06-24
Add OCI/Image Volume Source support

saschagrunert:
#### What type of PR is this?

/kind feature

Adding support for the new OCI Volume Source feature.

Refers to https://github.com/kubernetes/enhancements/issues/4639

Requires https://github.com/kubernetes/kubernetes/pull/125659

```release-note
Added support for the Kubernetes OCI / image Volume Source ([KEP-4639](https://github.com/kubernetes/enhancements/issues/4639)).
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -26,6 +26,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/reexec"
 	json "github.com/json-iterator/go"
+	"github.com/moby/sys/mountinfo"
 	digest "github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
@@ -56,6 +57,7 @@ type ImageResult struct {
 	OCIConfig           *specs.Image
 	Annotations         map[string]string
 	Pinned              bool // pinned image to prevent it from garbage collection
+	MountPoint          string
 }
 
 type indexInfo struct {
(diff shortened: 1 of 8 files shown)


server/container_create_linux.go
```diff
@@ -1228,6 +1234,26 @@ func addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container, mountLabel,
...
+	log.Debugf(ctx, "Image ref to mount: %s", imageRef)
+	status, err := s.storageImageStatus(ctx, types.ImageSpec{Image: imageRef})
+	if err != nil {
+		return "", fmt.Errorf("get storage image status: %w", err)
+	}
+
+	id := status.ID.IDStringForOutOfProcessConsumptionOnly()
+	log.Debugf(ctx, "Image ID to mount: %v", id)
+
+	mountPoint, err := s.Store().MountImage(id, []string{"noexec"}, mountLabel)
```
kwilczynski, line 1248: Should we make it also "ro", "rbind", "nosuid" and "nodev"?

kwilczynski: Oh interesting Re: "robind". I wonder why...

internal/storage/image.go
```diff
@@ -309,6 +311,30 @@ func (svc *imageService) buildImageResult(image *storage.Image, cacheItem imageC
...
+	}
+	// Check if the mount actually exists
+	if mountPoint != "" {
+		infos, err := mountinfo.GetMounts(mountinfo.SingleEntryFilter(mountPoint))
+		if err != nil {
+			logrus.Warnf("Unable to get mount info for path %s: %v", mountPoint, err)
+			mountPoint = ""
+		}
+		if len(infos) == 0 {
+			logrus.Warnf("Mount path %s for image (%s) is not mounted any more, unmounting it in c/storage", mountPoint, image.ID)
```
kwilczynski, line 330: What about something like this, an example:

> Unable to find mount path %s for image %s, assuming image is no longer mounted

or

> Mount path %s is no longer present, assuming image %s is no longer mounted

The idea is not to expose the implementation details too much to the end user. I assume that "unmounting it in c/storage" means cleaning up some internal state? Otherwise, if c/storage tracks OS-level mounts, then it should also notice the mount point is gone, I assume.

saschagrunert: Sure, adapted as suggested :+1:

internal/storage/image.go
```diff
@@ -465,7 +492,7 @@ func (svc *imageService) PrepareImage(inputSystemContext *types.SystemContext, i
 	return srcRef.NewImage(svc.ctx, systemContext)
 }
 
-//nolint
+// nolint
```
kwilczynski, line 495: Oh, we got a complaint here...

saschagrunert: That's my vim plugin, let me do it in a way which should be fine.

PR #8346, opened 2024-07-05
Add `kubernetes` group in dependabot

saschagrunert:
#### What type of PR is this?

/kind ci

As a result we should get one PR for Kubernetes dependency bumps instead of multiple ones.

None
None

```release-note
None
```


sohankunkerkar (approved):
/lgtm

PR #8352, opened 2024-07-07
Merge cli flags into config after hot reload

LenkaSeg:
#### What type of PR is this?

/kind bug

Merges cli flags into config on hot reload.

Should fix the Issue #7586 

Still work in progress. While solving this issue I ran into some circular imports, so it needed refactoring. Not sure if this is preferred way of the code refactor.

Yes

```release-note
Command line arguments preserved after hot reload.
```


LenkaSeg:
With the changes this PR proposes, these two tests are failing:

https://github.com/cri-o/cri-o/blob/main/test/reload_config.bats#L36-L56

The reason is that with the change we retain the command line arguments after the event of hot reload.

LenkaSeg:
To confirm that the changes in this PR are merging the CLI flags into config:

```
bin/crio --pinns-path=$(pwd)/bin/pinns
```

From a different terminal:

```
bin/crio status config print | grep log_level
```

Output:
```
log_level = "debug"
```

```
kill -HUP $(pgrep crio)
```

```
bin/crio status config print | grep log_level
```

Output (is still debug):
```
log_level = "debug"
```

kwilczynski:
> To confirm that the changes in this PR are merging the CLI flags into config:
> 
> ```
> bin/crio --pinns-path=$(pwd)/bin/pinns
> ```
> 
> From a different terminal:
> 
> ```
> bin/crio status config print | grep log_level
> ```
> 
> Output:
> 
> ```
> log_level = "debug"
> ```
> 
> Let's do the hot reload:
> 
> ```
> kill -HUP $(pgrep crio)
> ```
> 
> Check the log level again:
> 
> ```
> bin/crio status config print | grep log_level
> ```
> 
> Output (is still debug):
> 
> ```
> log_level = "debug"
> ```

@LenkaSeg very nice!

How did you set the log level here? Was CLI using a different level, and `crio.conf` had a different one? Looking at the above, I am not sure which way it was. :smile:

kwilczynski:
/hold

So that I can make a review.

LenkaSeg:
> How did you set the log level here? Was CLI using a different level, and `crio.conf` had a different one? Looking at the above, I am not sure which way it was. 😄

Ah, sorry, the log level was passed as a cli argument. The first command was supposed to be:

```
sudo bin/crio --pinns-path=$(pwd)/bin/pinns --log-level=debug
```

I copied the wrong line :)

LenkaSeg:
The conf had log_level set to "warn":

```
lenkaseg@fedora:$ cat /etc/crio/crio.conf.d/99-log-level.conf
[crio.runtime]
log_level = "warn"
```

kwilczynski:
@LenkaSeg, looks good so far! Thank you or persevering with this.

We might have to update some tests. Also, if you could rebase your branch to get some of the latest updates, then it would have with some other test failures we have here.

kwilczynski:
> Considering the "debug" option here:
> 
> https://github.com/cri-o/cri-o/blob/d622521f4ee8a7b21cf009b87236706b5616a101/test/helpers.bash#L222
> 
> Would it be ok to disable those failing tests? Or create a different bats file, add the test covering this change and customize `start_crio`?

@LenkaSeg, we seem to have some behaviour change now, which probably makes sense since the previous (old) behaviour was a bit buggy, so to speak. So now we need to figure out why we have these failing tests and how to update them so that these would pass.

If we delete a test, we can lose some signal unless we are sure that the test is no longer valid or feasible.

What do you think?

That said, I didn't have a look, yet, what the issues are. :bow:

test/reload_config.bats
```diff
@@ -33,31 +33,22 @@ function expect_log_failure() {
 	ps --pid "$CRIO_PID" &> /dev/null
 }
 
-@test "reload config should succeed with 'log_level'" {
+@test "log_level set via command line remains after reload" {
```
sohankunkerkar, line 36: Is there a plan to add these tests in a separate BATS file to ensure their coverage?

sohankunkerkar: I meant the ones you removed.

LenkaSeg: Sorry, forgot to include that file. Added now!

test/reload_config_cli_options.bats
LenkaSeg, line 1: Two tests started failing after the fix introduced in this PR: 
- reload config should succeed with 'log_level'
- reload config should fail with 'log_level' if invalid
The original `start_crio_no_setup` function in helpers.bash contains log level flag:

https://github.com/cri-o/cri-o/blob/d5b40bbdcbcbaf8d5efe52e01772ecf48a1fc62f/test/helpers.bash#L219-228

With the fix that preserves cli flags now these tests did not result with log level "warn", which is taken from the config, but `debug`.
To keep these tests intact, they were moved into a new file `reload_config_cli_options.bats`, and crio is started from the function `start_crio_without_debug` which does not contain the log_level flag.

LenkaSeg:
@kwilczynski Would you mind having a look?

kwilczynski:
> @kwilczynski Would you mind having a look?

@LenkaSeg sorry for the delay! I will have a look in a day or so.

sohankunkerkar:
I think we still need this PR.

PR #8325, opened 2024-06-28
[release-1.29] OCPBUGS-33350: Reload config should remove pinned images when an empty list is provided

sohankunkerkar:
This is a manual cherry-pick of https://github.com/cri-o/cri-o/pull/8213

/assign sohankunkerkar

```release-note
Reload config should remove pinned images when an empty list is provided
```

PR #8362, opened 2024-07-09
server: propagate the pinned status of an image when queried via crictl 

sohankunkerkar:
#### What type of PR is this?

/kind cleanup

```bash
$ sudo crictl inspecti registry.k8s.io/pause:3.9 | jq .status.pinned
true
```

None

None

```release-note
None
```

### DIFF
--- a/server/image_status.go
+++ b/server/image_status.go
@@ -52,6 +52,7 @@ func (s *Server) ImageStatus(ctx context.Context, req *types.ImageStatusRequest)
 			Spec: &types.ImageSpec{
 				Annotations: status.Annotations,
 			},
+			Pinned: status.Pinned,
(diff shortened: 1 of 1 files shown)


kwilczynski:
@sohankunkerkar, would this be more of a bug since the information was missing?

sohankunkerkar:
> @sohankunkerkar, would this be more of a bug since the information was missing?

IMO, this is more of a UX improvement than a bug as the image status was missing that information. Currently, we can get that information from `crictl images --pinned`.

kwilczynski:
The **release-1.29** already included the code change referenced here through the following Pull Request:

- https://github.com/cri-o/cri-o/pull/8325

Issue #6743, opened 2023-03-17
fsync() contention when container logs get rotated can block containers writing to stdout/stderr


rmt:
### What happened?

When kubelet rotates logs and asks the remote runtime to reopen logs, a blocking fsync() occurs in conmon.  The pipe buffers can fill in the meantime while waiting for fsync() to complete, blocking writes to stdout/stderr.

conmon has an option called --no-sync-log which can alleviate this problem (and should arguably be the default)
However, podman does not support setting this.
And neither does cri-o, which uses podman.

The required changes should probably be driven by CRI-O.

This will usually not be noticed on a local disk or with low-volume logs, and very intermittently if the max-log-file-size is sufficiently large.  However, we have logs on an AWS EBS (GP3) volume and there is latency involved, and the max log file size was quite low given the amount of logs produced, which made the issue very obvious and got us to dig into it.

I suspect that this is that this is behind a significant number of (usually minor) latency spikes that people see.

### What did you expect to happen?

Writing to stdout from a container should not block under normal circumstances

### How can we reproduce it (as minimally and precisely as possible)?

Do lots writes and fsync()s on the container log volume while producing lots logs in a tight loop from one or more k8s containers, and set the max-log-file-size lower.   During log rotation / when the pipe is blocked, fewer log lines will be produced.  In a real world scenario, this might be HTTP requests that are delayed.

### Anything else we need to know?

Calling fsync on the critical path when writing log files is probably a bad thing

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
crio version 1.23.5
Version:          1.23.5
GitCommit:        d12c2842e9e4d051a2496e1740f846ffa22b3f1d
GitTreeState:     clean
BuildDate:        2023-01-10T21:30:25Z
GoVersion:        go1.17.5
Compiler:         gc
Platform:         linux/amd64
Linkmode:         dynamic
BuildTags:        exclude_graphdriver_devicemapper, seccomp
SeccompEnabled:   true
AppArmorEnabled:  false
```

```console
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.14", GitCommit:"3321ffc07d2f046afdf613796f9032f4460de093", GitTreeState:"clean", BuildDate:"2022-11-09T13:40:19Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #8354, opened 2024-07-08
Sanitize CRI-O version set in the User-Agent header

kwilczynski:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

The version CRI-O uses to set the **User-Agent** header when making requests against a remote container registry can be overridden during the build process, and it might contain more information than just a simple semver compatible version string in the "X.Y.Z" format. For example, the OpenShift-specific build of CRI-O includes release and build information as part of the version string, such as the following:

```console
1.27.8-2.rhaos4.14.gitbfac241.el9
```

This extra information, while useful as part of binary package metadata or when requesting a version using command-line switches such as `--version`, would not necessarily be useful when included in the User-Agent header. Usually, there isn't any need for the User-Agent header to include a very detailed version, release and build information.

Thus, ensure that even if the version string has been overridden and some extra information appended to it, the version included as part of the User-Agent header will only ever contain a simple version string, with any surplus information removed.

While at it, remove the not needed `VersionInfo` type slice and `append()` usage.

Related to:

- https://github.com/openshift/os/issues/1542
- https://github.com/kubernetes/registry.k8s.io/issues/286
- https://github.com/containers/image/pull/201
- https://github.com/containerd/containerd/issues/6474

This also potentially addresses the following valid concern raised by @AkihiroSuda in a different project:

Screenshot from 2024-07-10 12-33-04

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #8363, opened 2024-07-09
Allow disabling fsync on log rotate and pod shutdown

rtreffer:
#### What type of PR is this?

/kind feature

On log rotate or container exit the logs are closed and fsync is used to flush the data to disk.
This results in good crash recovery behavior but it may cause IO locks that can cause pods to wait on logging to stdout/stderr. #6743 describes the issue.

`conmon` supports `--no-sync-log` to avoid this issue. This PR adds the same (pass through) `no_sync_log` option to cri-o.

Fixes: #6743 

This is my first PR to cri-o.  I have been able to test this locally (manually), `conmon` starts up with `--no-sync-log`.

```release-note
Add `no_sync_log` option to disable fsync on container log rotation and container exit. This can improve performance at the cost of potential data loss on machine crashes.
```


pkg/config/config.go
```diff
@@ -433,6 +433,10 @@ type RuntimeConfig struct {
 	// Negative values indicate that the log file won't be truncated.
 	LogSizeMax int64 `toml:"log_size_max"`
 
+	// NoSyncLog if enabled will disable fsync on log rotation and container exit.
+	// This can improve performance but may result in date loss on hard system crashes.
+	NoSyncLog bool `toml:"no_sync_log"`
```
haircommander, line 438: @cri-o/cri-o-maintainers I wonder if this should be a member of the runtime config structure, as it doesn't work for runtime pod or vm right now

rtreffer: I just checked other log related settings in this struct and they seem to have mixed scopes as well
- `LogSizeMax` seems to apply to oci and pod
- `LogToJournald` seems to apply to oci only

Would it be acceptable to document this correctly?
E.g. appending `Applies to oci containers only` to all doc strings for the setting?

rtreffer:
The remaining test failures (rawhide rpm build) are in line with other PRs / builds and as far as I can tell unrelated.

pkg/config/config.go
kwilczynski: We have, indeed, overloaded **RuntimeConfig** a little bit. I still am of a belief that **LogLevel** should not be there, for example. I suppose it was also done with some of the options for convenience since it's easier to access some of these depending on which objects are being passed around...

@haircommander, I think it would be fine... for now. We then can figure out how to clean this up.

Thoughts?

kwilczynski:
> The remaining test failures (rawhide rpm build) are in line with other PRs / builds and as far as I can tell unrelated.

@rtreffer, that is fine to ignore for now. We know why this is failing. See:

- https://github.com/cri-o/cri-o/pull/8354#issuecomment-2216960687
- https://github.com/capnproto/go-capnp/pull/576

So nothing to worry about for now :smile:

pkg/config/config.go
```diff
@@ -433,6 +433,10 @@ type RuntimeConfig struct {
 	// Negative values indicate that the log file won't be truncated.
 	LogSizeMax int64 `toml:"log_size_max"`
 
+	// NoSyncLog if enabled will disable fsync on log rotation and container exit.
+	// This can improve performance but may result in date loss on hard system crashes.
```
ccmtaylor, line 437: nit: typo
```suggestion
	// This can improve performance but may result in data loss on hard system crashes.
```

kwilczynski: Well spotted!

rtreffer: Fixed. Thank you!

haircommander:
I find myself wondering if we don't need this afterall, whenever https://github.com/containers/conmon/pull/512 gets released. We still do the fsync for pod shutdown, but that happens way less frequently than log rotate and I find myself wondering if the underlying issue has been fixed

rtreffer:
Good question.
Now that I think about it: how does this behave for crashing sidecars? Will we get into a fsync loop without this patch?

This sounds like more of a problem than the log rotate.

haircommander:
> Good question. Now that I think about it: how does this behave for crashing sidecars? Will we get into a fsync loop without this patch?
> 
> This sounds like more of a problem than the log rotate.

since there are more steps in the container creation process, I Wouldn't expect fsync to be the piece that really stalls or over consumes CPU in that situation. But I don't have any data to back that up.

rtreffer:
@haircommander I tried to reproduce the problem, I attached it to the issue, https://github.com/cri-o/cri-o/issues/6743#issuecomment-2241329430

I am pretty confident by now that this _is_ an issue, and potentially larger than I assumed. But it's getting really late today.....

pkg/config/config.go
haircommander: sorry by runtime config I mean `crio.runtime.runtimes` runtime (talk about overloaded!) . this only works for runtime type OCI, and I'd love to see it as a member there instead

rtreffer: Ah, ok, so
* Move it to RuntimeHandler
* Add validation that it only applies to OCI

The thing I haven't figured out is what to do about the cli option in that case.
But I'll update this PR once I get that sorted.

haircommander: I personaly don't think the cli option is needed. Do you really want it @rtreffer ?

rtreffer: Well, it was _convinient_, but looks like it might be annoying to expose it as a flag, so I am in favor of dropping it for simplicity.

rtreffer:
I've reworked the PR as mentioned, `no_sync_log` it is now a member of the runtimes table, validation rules and test cases added.

Documentation of the setting contains a warning about data loss. Setting it to true will cause a warning before a conmon is spawned as that's the place where we can be sure it is in use.

rtreffer:
I came across a tiny nitpick that might be worth changing, the `runtime_type` should be validated before the other validators, PR https://github.com/cri-o/cri-o/pull/8427

internal/oci/runtime_oci.go
```diff
@@ -145,6 +145,11 @@ func (r *runtimeOCI) CreateContainer(ctx context.Context, c *Container, cgroupPa
 	if r.config.LogSizeMax >= 0 {
 		args = append(args, "--log-size-max", strconv.FormatInt(r.config.LogSizeMax, 10))
 	}
+	if r.handler.NoSyncLog {
+		log.Warnf(ctx, "NoSyncLog is enabled for %s (%s), this can lead to lost log data",
```
haircommander, line 149: this will log for each container creation. can we instead log it when we validate it in `pkg/config/config.go`?

rtreffer: Sure, this has the advantage of being very explicit of each use of it (even without using `ps` to verify).
I'll move it to the validation.

haircommander:
/approve

thanks @rtreffer

saschagrunert (approved):

PR #8368, opened 2024-07-11
Remove go-setup from jobs not using golang

saschagrunert:
#### What type of PR is this?

/kind ci

We do not need go for binary related checks, so we remove that step.

None
None

```release-note
None
```


kwilczynski:
Makes sense.

/approve
/lgtm

PR #8369, opened 2024-07-11
Move `internal/client` to `pkg/client`

saschagrunert:
#### What type of PR is this?

/kind api-change

Make the client package public to be re-used from other projects.

None
None

```release-note
Make `pkg/client` public to be re-used from third party applications.
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
@@ -9,13 +10,25 @@ import (
 	"syscall"
 	"time"
 
-	json "github.com/json-iterator/go"
-
 	"github.com/cri-o/cri-o/pkg/types"
-	"github.com/cri-o/cri-o/server"
 )
 
 const (
+	// ConfigEndpoint is the endpoint for the config retrieval.
+	ConfigEndpoint = "/config"
+
+	// ContainersEndpoint is the endpoint for the container information retrieval.
+	ContainersEndpoint = "/containers"
+
+	// InfoEndpoint is the endpoint for the daemon information retrieval.
+	InfoEndpoint = "/info"
+
+	// PauseEndpoint is the endpoint for pausing containers.
+	PauseEndpoint = "/pause"
+
+	// UnpauseEndpoint is the endpoint for unpausing containers.
+	UnpauseEndpoint = "/unpause"
+
 	maxUnixSocketPathSize = len(syscall.RawSockaddrUnix{}.Path)
 )
 
@@ -71,11 +84,10 @@ func (c *crioClientImpl) getRequest(path string) (*http.Request, error) {
 	return req, nil
 }
 
-// DaemonInfo return cri-o daemon info from the cri-o
-// info endpoint.
+// DaemonInfo return CRI-O daemon info from the CRI-O info endpoint.
 func (c *crioClientImpl) DaemonInfo() (types.CrioInfo, error) {
 	info := types.CrioInfo{}
-	req, err := c.getRequest(server.InspectInfoEndpoint)
+	req, err := c.getRequest(InfoEndpoint)
 	if err != nil {
 		return info, err
 	}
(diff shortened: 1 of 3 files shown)


kwilczynski:
@saschagrunert this is as per the ask on Kubernetes Slack, correct? To be able to fetch CRI-O's runtime configuration via an API request, etc.

elezar:
> @saschagrunert this is as per the ask on Kubernetes Slack, correct? To be able to fetch CRI-O's runtime configuration via an API request, etc.

Yes, that is correct. The discussion is here: https://kubernetes.slack.com/archives/CAZH62UR1/p1720627584166869

pkg/client/client.go
elezar, line 1: @saschagrunert one question about this move. It seems as if the `github.com/cri-o/cri-o/server` import is only to get the various `*Endpoint` constants. Are there concerns with the dependency graph that this will introduce when an third-party application imports the `client` package?

saschagrunert: Yes we probably should clean that up.

saschagrunert: Now the required import looks like this:

```gomod
module github.com/saschagrunert/crio-client

go 1.22.5

require github.com/cri-o/cri-o v0.0.0-00010101000000-000000000000

require (
	github.com/containers/storage v1.54.0 // indirect
	github.com/docker/go-units v0.5.0 // indirect
	github.com/json-iterator/go v1.1.12 // indirect
	github.com/moby/sys/mountinfo v0.7.1 // indirect
	github.com/moby/sys/user v0.1.0 // indirect
	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
	github.com/modern-go/reflect2 v1.0.2 // indirect
	github.com/sirupsen/logrus v1.9.3 // indirect
	golang.org/x/sys v0.21.0 // indirect
)

replace github.com/cri-o/cri-o => github.com/saschagrunert/cri-o v1.22.1-0.20240711090808-67540e26389d
```

saschagrunert: Alright, now an application like this:

```go
package main

import (
	"fmt"

	"github.com/cri-o/cri-o/pkg/client"
)

func main() {
	c, err := client.New("/var/run/crio/crio.sock")
	if err != nil {
		panic(err)
	}

	cfg, err := c.ConfigInfo()
	if err != nil {
		panic(err)
	}
	fmt.Print(cfg)
}
```

would import

```gomod
github.com/sirupsen/logrus
```

haircommander:
I am happy to merge this, but as a note I don't think we should commit to any API compatibility . These endpoints have been relatively stable but I don't want to need to worry about breaking external users of it. does that sound fair @elezar ?

elezar:
> I am happy to merge this, but as a note I don't think we should commit to any API compatibility . These endpoints have been relatively stable but I don't want to need to worry about breaking external users of it. does that sound fair @elezar ?

Sure, that sounds reasonable. As I mentioned in the Slack channel, our main intent is to get the runtimes that are configured so as to construct a correct drop-in file for our shim. I'm willing to accept some API risks to get this functionality.

kwilczynski:
> > I am happy to merge this, but as a note I don't think we should commit to any API compatibility . These endpoints have been relatively stable but I don't want to need to worry about breaking external users of it. does that sound fair @elezar ?
> 
> Sure, that sounds reasonable. As I mentioned in the Slack channel, our main intent is to get the runtimes that are configured so as to construct a correct drop-in file for our shim. I'm willing to accept some API risks to get this functionality.

Given @haircommander's point, it makes me wonder if we need to expose anything. Anyone can make a request against the endpoint, whether using a client of their choosing or curl, etc. I am not sure if we need to provide a client.

This would be something that will become available for use (given it no longer be a private package); as such, we would have to be careful not to break it for the users potentially. Or make it very clear that this is not to be relieved on since there aren't any guarantees around backward compatibility, etc. So then, why bother to expose a client (aside from obvious things being the convenience)?

Perhaps we should maintain a small SDK/client package to wrap CRI-O APIs around? (probably an overkill)

Thoughts?

elezar:
@kwilczynski I'm ok with using a rest client to do this instead -- especially since I'm primarily interested in the `/config` endpoint which outputs a basic string anyway. I have not had time to really investigate this on my end though.

saschagrunert:
> @kwilczynski I'm ok with using a rest client to do this instead -- especially since I'm primarily interested in the `/config` endpoint which outputs a basic string anyway. I have not had time to really investigate this on my end though.

Alright, then let's keep it private for now.

PR #8372, opened 2024-07-12
Allow mounting a cgroup v2 namespace for privileged containers

hensur:
#### What type of PR is this?

/kind feature

Adds the `io.kubernetes.cri-o.cgroup2-ns-privileged` annotation, which allows to mount a cgroup2 namespace different from the host namespace into privileged containers.
This is needed if one would want to run containers inside of privileged cri-o containers with cgroup nesting.

None

```release-note
Added the `io.kubernetes.cri-o.cgroup2-ns-privileged` annotation which allows mounting a cgroup v2 namespace for privileged containers.
```

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -11,6 +11,9 @@ const (
 	// CgroupRW specifies mounting v2 cgroups as an rw filesystem.
 	Cgroup2RWAnnotation = "io.kubernetes.cri-o.cgroup2-mount-hierarchy-rw"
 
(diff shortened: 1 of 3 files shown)


haircommander:
/ok-to-test

code LGTM, thank you @hensur . any chance you fancy adding a bats test to our integration suite in `test` ?  it could be added to cgroups.bats I think

hensur:
i've added two test cases now. I hope those work, I didn't want to go through the hassle of setting up a local test environment (yet).
I also rebased on current main.

PR #8397, opened 2024-07-18
Cleanup/refactor create sandbox container

littlejawa:
#### What type of PR is this?

/kind cleanup

This PR is made of cherry-pics from #8008: The refactoring that is done there can make sense outside of the PR.
The goal here is to reduce the complexity of createSandboxContainer() by extracting some functions. Some are moved to the factory/container code, others are just made sub-functions in the same file.

Hopefully, this can make createSandboxContainer() easier to maintain.

None

I'm not sure that the way I split the functions makes sense in the way the "container" object is used. My focus was on code readability, and extracting related parts of the code as sub-functions. We can discuss it and adapt.
See this as a first draft.

```release-note
None
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -24,7 +24,9 @@ import (
 	kubeletTypes "k8s.io/kubelet/pkg/types"
 
 	"github.com/cri-o/cri-o/internal/config/capabilities"
+	"github.com/cri-o/cri-o/internal/config/cgmgr"
 	"github.com/cri-o/cri-o/internal/config/device"
+	"github.com/cri-o/cri-o/internal/config/node"
 	"github.com/cri-o/cri-o/internal/config/nsmgr"
 	"github.com/cri-o/cri-o/internal/lib"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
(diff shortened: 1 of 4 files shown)


server/container_create_linux.go
```diff
@@ -124,6 +123,23 @@ func (s *Server) finalizeUserMapping(sb *sandbox.Sandbox, specgen *generate.Gene
 	}
 }
 
+// this function takes a container config and makes sure its SecurityContext
+// is not nil. If it is, it makes sure to set default values for every field.
+func setContainerConfigSecurityContext(containerConfig *types.ContainerConfig) {
```
haircommander, line 128: can this be defined below where it's called?

internal/factory/container/container.go
```diff
@@ -759,3 +767,98 @@ func getOCICapabilitiesList() []string {
 	}
 	return caps
 }
+
+func (c *container) SpecSetPrivileges(ctx context.Context, securityContext *types.LinuxContainerSecurityContext, cfg *config.Config) error {
```
haircommander, line 771: part of the hope of this package is to also add unit tests to test the spec as we modify it. I know this has been sitting for a long time, are you able to add some unit tests here @littlejawa ? if not, we can maybe find another contributor to add them

littlejawa: Sure, makes sense.
I'll need a little more time, but if you're ok with waiting a bit, I will look into it

haircommander:
a couple of unit, lint and build failures @littlejawa

littlejawa:
> a couple of unit, lint and build failures

Looks like I fixed them...
I merged the fix for the freebsd build with the commit that introduced the problem (b0b584ac886d02c0ca9cf46d66a8beabac1872a9).

littlejawa:
@haircommander - do you think we can move forward with this?

haircommander:
/approve

LGTM, thanks @cri-o/cri-o-maintainers PTAL

FYI @xw19

xw19:
noted

kwilczynski:
/hold

So I can make a review, too.

PR #8403, opened 2024-07-20
Fix bug where deleting a container using instance name resulted in th…

lidajia:
…e container not being deleted

In the restore processing logic, if the sandbox fails to load and attempts to delete a container under the pod, the container cannot be deleted due to incorrect use of the container name

#### What type of PR is this?

/kind bug

Fixes # 8401

```release-note

```


haircommander:
hm that delete container function should be able to take an id or a name. What is the exact error you get? we should also be ignoring `ErrNotAContainer` which I'd imagine would be the error you'd hit in the situation described

PR #8408, opened 2024-07-22
Make ImageVolume garbage collection work

saschagrunert:
#### What type of PR is this?

/kind feature

We need to return the image ID for the localhost mount, otherwise the kubelet garbage collection is not able to handle them.

Refers to https://github.com/kubernetes/enhancements/issues/4639

Follow-up on https://github.com/cri-o/cri-o/pull/8317
None

```release-note
None
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -1075,10 +1075,12 @@ func (s *Server) addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container,
 			return nil, nil, errors.New("mount.ContainerPath is empty")
 		}
 		if m.Image != nil && m.Image.Image != "" {
-			m.HostPath, err = s.mountImage(ctx, m.Image.Image, mountLabel)
+			mountPoint, imageID, err := s.mountImage(ctx, m.Image.Image, mountLabel)
 			if err != nil {
 				return nil, nil, fmt.Errorf("mount image: %w", err)
 			}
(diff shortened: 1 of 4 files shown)


server/container_status.go
```diff
@@ -135,6 +137,36 @@ func (s *Server) ContainerStatus(ctx context.Context, req *types.ContainerStatus
 	return resp, nil
 }
 
+// hostPathToPossibleImageSpec tries to find an image by the provided host path if the image is mounted locally.
+// It returns nil if nothing was found and logs corresponding errors to not fail unexpectedly.
+func (s *Server) hostPathToPossibleImageSpec(ctx context.Context, hostPath string) *types.ImageSpec {
```
haircommander, line 142: do you think we could save this in the volume structure? recalculating on each status may get expensive

saschagrunert: Yes, done :+1:

server/container_create_linux.go
```diff
@@ -1235,11 +1238,11 @@ func (s *Server) addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container,
 }
 
 // mountImage mounts the provided imageRef using the mountLabel and returns the hostPath on success.
```
kwilczynski, line 1240: Nit: I wonder if it would make sense to update the comment to reflect the new return value?

saschagrunert: Fixed as suggested :+1:

PR #8415, opened 2024-07-24
Remove error log message if mountpoint cannot be found

saschagrunert:
#### What type of PR is this?

/kind cleanup

There are several use cases around this, so we do not need to log an error here.

None
None

```release-note
None
```


kwilczynski:
@saschagrunert, not useful as a debug log message?

PR #8417, opened 2024-07-25
Use custom storage check options for CRI-O internal wipe

kwilczynski:
#### What type of PR is this?

/kind feature

When enabled, the internal repair feature uses the `storage.CheckEverything()`  helper function to configure what the `storage.Check()` should check for when scanning the storage directory. The following options are enabled:

```go
func CheckEverything() *CheckOptions {
	return &CheckOptions{
		LayerDigests:   true,
		LayerMountable: true,
		LayerContents:  true,
		LayerData:      true,
		ImageData:      true,
		ContainerData:  true,
	}
}
```

Two of these options, `LayerDigests` and `LayerContents`, are known to be I/O and CPU intensive, and both of these options are used to perform storage integrity checks and require the contents of the images to be read, checksums generated, and the current state compared against the desired state. As such, depending on the number of images, the layers each image has, and the number of files within each layer, it can take a considerable amount of time before the storage scan with both of these options enabled will complete.

Even though CRI-O only performs storage checks and repairs following an unclean shutdown, with the two checks mentioned earlier options enabled, it can take CRI-O some time to finish storage repairs. During this time, it won't be able to serve API requests. This can lead to the kubelet not being able to get responses to its requests in time, and eventually assuming that CRI-O is not responsive and proceed to make this specific node `NotReady`, which is undesirable. 

Thus, to mitigate the problem with storage scans potentially taking a long time to complete, we are going to switch to a custom configuration for the `storage.Check()` that turns off these expensive checks and only turns on checks that allow CRI-O to perform a "quick" scan of the storage directory.

This quick scan should cover most of the common types of storage failures and take significantly less time to complete. However, this comes at the expense of container image integrity verification, where files that are missing from the image or files that have their content altered will no longer be detected.

The goal of the internal repair is to ensure that CRI-O can recover from issues surrounding a corrupted storage directory on start-up, with an added option only to remove corrupted images and containers when attempting a repair, compared to removing the entire storage directory content. Hence, trading scan run time for correctness is acceptable.

While at it, change the internal repair option to be enabled by default and add a new `crio check` sub-command, which allows to check the storage directory for errors and repair it, if needed.

Related:

- https://github.com/containers/storage/pull/1569
- https://github.com/cri-o/cri-o/pull/7190
- https://github.com/cri-o/cri-o/issues/7177

None

None

```release-note
Update the type of checks the internal repair feature performs on CRI-O's start-up following an unclean shutdown, enable the internal repair option by default, and add a new `crio check` sub-command.
```

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -62,7 +62,7 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
   Whether CRI-O should wipe containers after a reboot and images after an upgrade when the server starts.
   If set to false, one must run `crio wipe` to wipe the containers and images in these situations.
 
-**internal_repair**=false
+**internal_repair**=true
   InternalRepair is whether CRI-O should check if the container and image storage was corrupted after a sudden restart.
   If it was, CRI-O also attempts to repair the storage.
 
--- a/internal/criocli/check.go
+++ b/internal/criocli/check.go
@@ -1,0 +1,195 @@
+package criocli
+
+import (
+	"fmt"
+
+	"github.com/containers/storage"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+
+	"github.com/cri-o/cri-o/internal/lib"
+	"github.com/cri-o/cri-o/utils"
+)
+
+type checkErrors map[string][]error
+
+var CheckCommand = &cli.Command{
+	Name:   "check",
+	Usage:  usageText,
+	Action: crioCheck,
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    "age",
+			Aliases: []string{"a"},
+			Value:   "24h",
+			Usage:   "Maximum allowed age for unreferenced layers",
+		},
+		&cli.BoolFlag{
+			Name:    "force",
+			Aliases: []string{"f"},
+			Usage:   "Remove damaged containers",
+		},
+		&cli.BoolFlag{
+			Name:    "repair",
+			Aliases: []string{"r"},
+			Usage:   "Remove damaged images and layers",
+		},
+		&cli.BoolFlag{
+			Name:    "quick",
+			Aliases: []string{"q"},
+			Usage:   "Perform only quick checks",
+		},
+		&cli.BoolFlag{
+			Name:    "wipe",
+			Aliases: []string{"w"},
+			Usage:   "Wipe storage directory on repair failure",
+		},
+	},
+}
+
+func crioCheck(c *cli.Context) error {
+	config, err := GetConfigFromContext(c)
+	if err != nil {
+		return fmt.Errorf("unable to load configuration: %w", err)
+	}
+
+	store, err := config.GetStore()
+	if err != nil {
+		return fmt.Errorf("unable to open storage: %w", err)
+	}
+	defer func() {
+		if _, err := store.Shutdown(true); err != nil {
+			logrus.Errorf("Unable to shutdown storage: %v", err)
+		}
+	}()
+
+	graphRoot := store.GraphRoot()
+	logrus.Infof("Checking storage directory %s for errors", graphRoot)
+
+	checkOptions := storage.CheckEverything()
+	if c.Bool("quick") {
+		// This is not the same as the "quick" check that CRI-O performs during its start-up
+		// following an unclean shutdown, as this one would set the `LayerDigests` option,
+		// which is I/O and CPU intensive, whereas the other one does not.
+		checkOptions = storage.CheckMost()
+	}
+
+	// The maximum unreferenced layer age.
+	layerAge := c.String("age")
+	if layerAge != "" {
+		age, err := utils.ParseDuration(layerAge)
+		if err != nil {
+			return fmt.Errorf("unable to parse age duration: %w", err)
+		}
+		checkOptions.LayerUnreferencedMaximumAge = &age
+	}
+
+	report, err := store.Check(checkOptions)
+	if err != nil {
+		return fmt.Errorf("unable to check storage: %w", err)
+	}
+
+	// Walk each report and show details...
+	for prefix, checkReport := range map[string]checkErrors{
+		"layer":           report.Layers,
+		"read-only layer": report.ROLayers,
+		"image":           report.Images,
+		"read-only image": report.ROImages,
+		"container":       report.Containers,
+	} {
+		for identifier, errs := range checkReport {
+			for _, err := range errs {
+				logrus.Debugf("%s: %s: %v", prefix, identifier, err)
+			}
+		}
+	}
+
+	seenStorageErrors := lib.CheckReportHasErrors(report)
+	logrus.Debugf("Storage directory %s has errors: %t", graphRoot, seenStorageErrors)
+
+	if !c.Bool("repair") {
+		if seenStorageErrors {
+			logrus.Warnf("Errors found while checking storage directory %s for errors", graphRoot)
+			return fmt.Errorf(
+				"%d layer errors, %d read-only layer errors, %d image errors, %d read-only image errors, %d container errors",
+				len(report.Layers),
+				len(report.ROLayers),
+				len(report.Images),
+				len(report.ROImages),
+				len(report.Containers),
+			)
+		}
+		return nil
+	}
+
+	force := c.Bool("force")
+	if force {
+		logrus.Warn("The `force` option has been set, repair will attempt to remove damaged containers")
+	}
+	logrus.Infof("Attempting to repair storage directory %s", graphRoot)
+
+	errs := store.Repair(report, &storage.RepairOptions{
+		RemoveContainers: force,
+	})
+	if len(errs) != 0 {
+		for _, err := range errs {
+			logrus.Error(err)
+		}
+
+		if c.Bool("wipe") {
+			// Depending on whether the `force` option is set or not, this will remove the
+			// storage directory completely while ignoring any running containers. Otherwise,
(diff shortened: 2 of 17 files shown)


kwilczynski:
@cri-o/cri-o-maintainers, please have a look. Thank you!

docs/crio.conf.5.md
```diff
@@ -57,12 +57,12 @@ CRI-O reads its storage defaults from the containers-storage.conf(5) file locate
 **imagestore**=""
   Store newly pulled images in the specified path, rather than the path provided by --root.
 
-**internal_wipe**=true
+**internal_wipe**=false
```
haircommander, line 60: hm i disagree with this, why is this here? internal_wipe isn't deprecated because it will be false in the future, it's deprecated because it will be set to unconditionally true in the future

haircommander: that's on me for not saying what deprecated meant here

kwilczynski: @haircommander, no worries. I took the following comment/note for it's face value:

> This option is currently DEPRECATED, and will be removed in the future.

kwilczynski: Done.

internal/lib/container_server.go
```diff
@@ -105,20 +106,25 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
...
-		checkOptions := cstorage.CheckEverything()
-		report, err := store.Check(checkOptions)
-		if err != nil {
-			err = HandleUncleanShutdown(config, store)
-			if err != nil {
-				return nil, err
+		graphRoot := store.GraphRoot()
+		log.Infof(ctx, "Checking storage directory %s for errors because of unclean shutdown", graphRoot)
+
+		var wipeStorage bool
```
saschagrunert, line 112: nit:
```suggestion
		wipeStorage := false
```

saschagrunert (commented):
Is it possible to write a reproducible integration test for this scenario?

internal/lib/container_server.go
kwilczynski: Done.

internal/lib/container_server.go
```diff
@@ -807,20 +812,81 @@ func ShutdownWasUnclean(config *libconfig.Config) bool {
...
-	}
-	// unmount storage or else we will fail with EBUSY
-	if _, err := store.Shutdown(false); err != nil {
-		return fmt.Errorf("failed to shutdown storage before wiping: %w", err)
+		logrus.Infof("Failed to wipe storage: %v", err)
+	}
+
+	// Unmount storage or else we will fail with -EBUSY.
+	if _, err := store.Shutdown(true); err != nil {
+		// CRI-O and Podman are rarely used together on the same node.
```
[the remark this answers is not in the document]
kwilczynski: Oops. Not true? What does Podman do on these nodes? Let me update the comment, though.

internal/criocli/check.go
```diff
@@ -0,0 +1,195 @@
...
+	for prefix, checkReport := range map[string]checkErrors{
+		"layer":           report.Layers,
+		"read-only layer": report.ROLayers,
+		"image":           report.Images,
+		"read-only image": report.ROImages,
+		"container":       report.Containers,
+	} {
+		for identifier, errs := range checkReport {
+			for _, err := range errs {
+				logrus.Debugf("%s: %s: %v\n", prefix, identifier, err)
```
saschagrunert, line 102:
```suggestion
				logrus.Debugf("%s: %s: %v", prefix, identifier, err)
```

internal/criocli/check.go
```diff
@@ -0,0 +1,195 @@
...
+		logrus.Warn("The `force` option has been set, repair will attempt to remove damaged containers")
+	}
+	logrus.Infof("Attempting to repair storage directory %s", graphRoot)
+
+	errs := store.Repair(report, &storage.RepairOptions{
+		RemoveContainers: force,
+	})
+	if len(errs) != 0 {
+		for _, err := range errs {
+			logrus.Errorf("%v\n", err)
```
saschagrunert, line 136:
```suggestion
			logrus.Error(err)
```

internal/criocli/check.go
```diff
@@ -0,0 +1,195 @@
...
+			len(report.Containers),
+		)
+		if force {
+			err = fmt.Errorf(
+				"%d read-only layer errors, %d read-only image errors",
+				len(report.ROLayers),
+				len(report.ROImages),
+			)
+		}
+		return err
```
saschagrunert, line 167:
```suggestion
		if force {
			return fmt.Errorf(
				"%d read-only layer errors, %d read-only image errors",
				len(report.ROLayers),
				len(report.ROImages),
			)
		}
		return fmt.Errorf(
			"%d read-only layer errors, %d read-only image errors, %d container errors",
			len(report.ROLayers),
			len(report.ROImages),
			len(report.Containers),
		)
```

saschagrunert (commented):
Just a couple of nits, otherwise LGTM

internal/criocli/check.go
kwilczynski: Done.

internal/criocli/check.go
kwilczynski: Done.

internal/criocli/check.go
kwilczynski: Done.

internal/criocli/check.go
```diff
@@ -0,0 +1,195 @@
...
+
+	graphRoot := store.GraphRoot()
+	logrus.Infof("Checking storage directory %s for errors", graphRoot)
+
+	checkOptions := storage.CheckEverything()
+	if c.Bool("quick") {
+		// This is not the same as the "quick" check that CRI-O performs during its start-up
+		// following an unclean shutdown, as this one would set the `LayerDigests` option,
+		// which is I/O and CPU intensive, whereas the other one does not.
+		checkOptions = storage.CheckMost()
```
haircommander, line 74: personally I think the quick here should match the one the crio binary does but I'm not feeling stubborn about that

kwilczynski: This is done on purpose. The `CheckMost()` picks more issues than our particular version. Since it also performs a `LayerDigests` scan, which is what we want. The complete scan can run twice as long (worst case), so often is not desirable, unless the user is willing to try it.

This is a good choice for a CLI utility, even though CRI-O on startup does something different to save on runtime.

I hope this makes sense.

kwilczynski: To add, once we extend the check feature capability to add file size and existence checks, then we can align the two, and introduce "quick", "default" and "complete", etc.

haircommander:
```
not ok 32 recover from badly corrupted storage directory
# (in test file crio-wipe.bats, line 363)
#   `umount -R -l -f "$TESTDIR"/crio/overlay' failed
# time="2024-08-09 12:22:23.605946349Z" level=info msg="Updating config from single file: /etc/crio/crio.conf"
# time="2024-08-09 12:22:23.605989403Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf"
# time="2024-08-09 12:22:23.606935825Z" level=info msg="Updating config from path: /etc/crio/crio.conf.d"
# time="2024-08-09 12:22:23.606997365Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/01-ns-lifecycle.conf"
# time="2024-08-09 12:22:23.607093229Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/01-overlay.conf"
# time="2024-08-09 12:22:23.607125756Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/99-log-level.conf"
# time="2024-08-09 12:22:23.643382454Z" level=info msg="Updating config from path: "
# time="2024-08-09T12:22:23Z" level=fatal msg="validate service connection: validate CRI v1 runtime API for endpoint \"unix:///tmp/tmp.pLNidEVRyE/crio.sock\": rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /tmp/tmp.pLNidEVRyE/crio.sock: connect: no such file or directory\""
# 1c2460bcafe6c2d536875edefccd6c0da43fa18caa2422d568ebbe5232b593b5
# umount: /tmp/tmp.pLNidEVRyE/crio/overlay: not mounted
```
kata failure. @kwilczynski can you override the test for now and open an issue so we can track the fix?

kwilczynski:
[...]
> ```
> # umount: /tmp/tmp.pLNidEVRyE/crio/overlay: not mounted
> ```
> 
> kata failure. @kwilczynski can you override the test for now and open an issue so we can track the fix?

@haircommander, should be resolved with a simple fix.

I am not entirely sure how Kata users c/storage.

If there are any more issues, I will exclude this test from being run under Kata, and have a look some more.

kwilczynski:
[...]
> If there are any more issues, I will exclude this test from being run under Kata, and have a look some more.

@haircommander, there is something else going on that makes these tests (for `crio check` and `crio wipe`) behave differently when run on Kata Containers per:

```console
# umount: /tmp/tmp.r6XWDPgPOM/crio/overlay: not mounted
# rm: cannot remove '/tmp/tmp.r6XWDPgPOM/crio/overlay/49596f1dcc172a24792e553ad4ae1c39f7a7ddf2519d13f666be930c2581f060/merged': Device or resource busy
# rm: cannot remove '/tmp/tmp.r6XWDPgPOM/crio/overlay/7c35e529e22d16f1997139e990e44bc3263476c0eb397912061a29679b2515cd/merged': Device or resource busy
```

I excluded both tests from being run on Kata Containers for now.

PR #8434, opened 2024-07-30
internal/oci: improve container termination process in killContainer

sohankunkerkar:
#### What type of PR is this?

/kind cleanup

This change aims to reduce unnecessary kill attempts and error logs when terminating containers that have already stopped or don't exist.

```sh
ERRO[2024-07-30 14:13:08.095902227-04:00] Killing container d159549c352df42b630974b3057f09146da18c37e266fe3e3d82f325eb557ebd failed: `/usr/bin/crun --root /run/runc --systemd-cgroup kill d159549c352df42b630974b3057f09146da18c37e266fe3e3d82f325eb557ebd KILL` failed: open pidfd: No such process
 : exit status 1 
```

```release-note
None
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -894,20 +894,24 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 
 	done := make(chan struct{})
 	go func() {
+		statusCheckTicker := time.NewTicker(stopProcessWatchSleep)
+		defer statusCheckTicker.Stop()
 		for {
-			if err := c.Living(); err != nil {
-				// The initial container process either doesn't exist, or isn't ours.
-				if !errors.Is(err, ErrNotFound) {
-					log.Warnf(ctx, "Failed to find process for container %s: %v", c.ID(), err)
-				}
-				close(done)
-				return
-			}
 			select {
 			case <-ctx.Done():
 				return
-			case <-time.After(stopProcessWatchSleep):
-				// Continue watching
+			case <-statusCheckTicker.C:
+				// Periodically check if the container is still running.
+				// This avoids busy-waiting and reduces resource usage while
+				// ensuring timely detection of container termination.
+				if err := c.Living(); err != nil {
+					// The initial container process either doesn't exist, or isn't ours.
+					if !errors.Is(err, ErrNotFound) {
+						log.Warnf(ctx, "Failed to find process for container %s: %v", c.ID(), err)
+					}
+					close(done)
+					return
+				}
 			}
 		}
 	}()
@@ -956,11 +960,15 @@ killContainer:
 	// We cannot use ExponentialBackoff() here as its stop conditions are not flexible enough.
 	kwait.BackoffUntil(func() {
 		if _, err := r.runtimeCmd("kill", c.ID(), "KILL"); err != nil {
-			log.Errorf(ctx, "Killing container %v failed: %v", c.ID(), err)
+			if !errors.Is(err, ErrNotFound) {
+				log.Errorf(ctx, "Killing container %v failed: %v", c.ID(), err)
+			}
+			log.Debugf(ctx, "Error while killing container %s: %v", c.ID(), err)
 		}
 
 		if err := c.Living(); err != nil {
 			stop()
+			return
 		}
 		// Reschedule the timer so that the periodic reminder can continue.
 		blockedTimer.Reset(stopProcessBlockedInterval)
@@ -1402,6 +1410,18 @@ func (r *runtimeOCI) runtimeCmd(args ...string) (string, error) {
 
 	err := cmd.Run()
 	if err != nil {
+		stdErrStr := stderr.String()
+		switch {
+		// crun, for most of the commands.
+		case strings.Contains(stdErrStr, "no such process"):
+			fallthrough //nolint:gocritic
+		// runc, for most of the commands.
+		case strings.Contains(stdErrStr, "container not running"):
+			fallthrough //nolint:gocritic
+		// runc, on a rare occasion.
+		case strings.Contains(stdErrStr, "invalid process"):
+			err = ErrNotFound
+		}
 		return "", fmt.Errorf("`%v %v` failed: %v %v: %w", r.handler.RuntimePath, strings.Join(runtimeArgs, " "), stderr.String(), stdout.String(), err)
 	}
 


internal/oci/runtime_oci.go
```diff
@@ -956,11 +956,16 @@ killContainer:
 	// We cannot use ExponentialBackoff() here as its stop conditions are not flexible enough.
 	kwait.BackoffUntil(func() {
 		if _, err := r.runtimeCmd("kill", c.ID(), "KILL"); err != nil {
+			if errors.Is(err, unix.ESRCH) {
+				// Process is already gone, stop the loop.
+				stop()
+				return
+			}
```
kwilczynski, line 963: The current value of errno is not passed between different processes. This is not the error CRI-O sees.

internal/oci/runtime_oci.go
```diff
@@ -956,11 +956,16 @@ killContainer:
...
 			log.Errorf(ctx, "Killing container %v failed: %v", c.ID(), err)
 		}
-
 		if err := c.Living(); err != nil {
```
kwilczynski, line 966: Are there issues with `Living()` not working currently? The idea is for it to ascertain whether a process is still "alive" and break the kill loop if not.

internal/oci/runtime_oci.go
sohankunkerkar: My previous approach was to check for the string ` if strings.Contains(err.Error(), "No such process")` but it doesn't appear like a full-proof one.

internal/oci/runtime_oci.go
sohankunkerkar: The issue appears to be related to the timing and sequence of operations. Specifically, `r.runtimeCmd("kill", c.ID(), "KILL")` successfully kills the container, but CRI-O waits for a response from crun. During this time, crun may have already deleted the container, leading to the "No such process" error. The error occurs because the container process is already gone when CRI-O checks its status.

internal/oci/runtime_oci.go
sohankunkerkar: > The current value of errno is not passed between different processes. This is not the error CRI-O sees.

whoops, I think that's my bad.

kwilczynski: This is not great, but for now this probably would have to do. Albeit, abstracting this away into a helper function would be an option, too.

A better option would be to talk to @kolyshkin and ask for more detailed exit codes to runc and crun so CRI-O could attempt to differentiate what happened. Currently, the exit code is just **1** most of the time on error.

Or, the runtime could also provide JSON document to stdout, as it does for other things such as `start` or `status` when terminating with a non-zero exit code, where said the document would carry a failure message and some other useful details we could parse on our side. However, I am not sure if this would be OCI-compliant. But probably on an error we could not care as much.

kwilczynski: This is what happens at the moment:

```console
([]string) (len=3 cap=3) {
 (string) (len=4) "kill",
 (string) (len=64) "d2b22ef4e5909c375e834aadda5fe7e16b74044e638ef164fd4625e95023264c",
 (string) (len=4) "KILL"
}
([]string) (len=6 cap=8) {
 (string) (len=6) "--root",
 (string) (len=9) "/run/runc",
 (string) (len=16) "--systemd-cgroup",
 (string) (len=4) "kill",
 (string) (len=64) "d2b22ef4e5909c375e834aadda5fe7e16b74044e638ef164fd4625e95023264c",
 (string) (len=4) "KILL"
}
(bytes.Buffer) open pidfd: No such process     <--- stderr (contains newline)

(bytes.Buffer)                                                          <--- stdout (empty)
(*exec.ExitError)(0xc00048be60)(exit status 1) <--- crun's exit code
```

This is what we then see as:

```console
ERRO[2024-07-31 14:19:55.694146382Z] Killing container d2b22ef4e5909c375e834aadda5fe7e16b74044e638ef164fd4625e95023264c failed: `/usr/bin/crio-crun --root /run/runc --systemd-cgroup kill d2b22ef4e5909c375e834aadda5fe7e16b74044e638ef164fd4625e95023264c KILL` failed: open pidfd: No such process
 : exit status 1
```

kolyshkin: @kwilczynski this needs to be done in runtime-spec first, so that multiple runtimes (runc, crun, youki) can behave in the same way. I think the best way to start the discusion about it is to open a PR against https://github.com/opencontainers/runtime-spec.

kolyshkin: cc @giuseppe ^^^

kwilczynski: @kolyshkin, sounds good! I will get to it then.

kwilczynski: @sohankunkerkar, I will take it to the runtime-spec community for feedback.

Meanwhile, we can proceed with a non-zero exit code plus a stderr string match. I suggested a helper function to do since we need to cover runc and crun, and they might have slightly different messages per:

- runc

```console
ERRO[2024-08-01 15:18:24.576675123Z] Killing container 1b3e518083c85f13bbe9b145da9f4f0dc7c254e48ce9fd7decbbcd6ae46acc1e failed: `/usr/bin/crio-runc --root /run/runc --systemd-cgroup kill 1b3e518083c85f13bbe9b145da9f4f0dc7c254e48ce9fd7decbbcd6ae46acc1e KILL` failed: time="2024-08-01T15:18:24Z" level=error msg="container not running"
 : exit status 1 
```

- crun

```console
 ERRO[2024-08-01 15:29:41.701878085Z] Killing container 52a27641c62eb42ec810d689c637dc86a329cdae2d46880c83a47122306d17f1 failed: `/usr/bin/crio-crun --root /run/runc --systemd-cgroup kill 52a27641c62eb42ec810d689c637dc86a329cdae2d46880c83a47122306d17f1 KILL` failed: open pidfd: No such process
 : exit status 1 
```

Also, runc can also emit an "invalid process" error, at some circumstances, I think.

internal/oci/runtime_oci.go
```diff
@@ -906,7 +908,7 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
 			select {
 			case <-ctx.Done():
 				return
-			case <-time.After(stopProcessWatchSleep):
```
kwilczynski, line 909: Why? The end result is exactly the same. Have you seen some issues?

kwilczynski: Also, I am glad you are looking in this direction. I wanted to ask about this function. We have it and the other, inner-loop, invocation of `Living()`, and it seems it's either a race or these aren't stopping the loop to stop the spurious kills. Hmm.

sohankunkerkar: I think the new logic eliminates the need for error string matching by centralizing the c.Living() check within a ticker-based goroutine, which reduces the likelihood of race conditions.

sohankunkerkar: > Why? The end result is exactly the same. Have you seen some issues?

I didn't find any issues. I just thought it was more idiomatic.

kwilczynski: > I think the new logic eliminates the need for error string matching by centralizing the c.Living() check within a ticker-based goroutine, which reduces the likelihood of race conditions.

@sohankunkerkar, that is much better, indeed!

Where we would still have issues would be the kill loop if the timeout is set to 0. However, this can only really be set when talking to the CRI directly, so with crictl, for example.

Curious that I couldn't set a timeout for termination grace period lower than 2 seconds. Haven't looked into why.

sohankunkerkar: >Where we would still have issues would be the kill loop if the timeout is set to 0. However, this can only really be set when talking to the CRI directly, so with crictl, for example.

I tried running the following example locally (a modified version of `ctr multiple stop calls`), and it worked without any problem.
```sh
@test "ctr stop with zero timeout" {
    start_crio

    # Create a container with a long-running command to simulate a scenario where
    # a container takes a while to stop gracefully.
    jq '.command = ["/bin/sh", "-c", "sleep 600"]' \
        "$TESTDATA"/container_config.json > "$newconfig"
    ctr_id=$(crictl run "$newconfig" "$TESTDATA"/sandbox_config.json)

    # Issue the crictl stop command with a timeout of 0.
    crictl stop --timeout 0 "$ctr_id" &> /dev/null

    # Verify that the container has either stopped or exited.
    final_state=$(crictl inspect "$ctr_id" | grep -Po '(?<="state": ")[^"]*')
    if [ "$final_state" != "CONTAINER_STOPPED" ] && [ "$final_state" != "CONTAINER_EXITED" ]; then
        echo "Test failed: Container did not stop or exit as expected."
        exit 1
    fi
}
```

sohankunkerkar: >Curious that I couldn't set a timeout for termination grace period lower than 2 seconds. Haven't looked into why.

I used this same example to set the `timeout` value as `1` and it worked.

kwilczynski: > > Curious that I couldn't set a timeout for termination grace period lower than 2 seconds. Haven't looked into why.
> 
> I used this same example to set the `timeout` value as `1` and it worked.

Sorry, I should have mentioned `terminationGracePeriodSeconds` from the Pod object specification rather than just the "grace period".

kwilczynski: > > Where we would still have issues would be the kill loop if the timeout is set to 0. However, this can only really be set when talking to the CRI directly, so with crictl, for example.
> 
> I tried running the following example locally (a modified version of `ctr multiple stop calls`), and it worked without any problem.
[...]

I am glad it works! However, we were trying to suppress spurious logs here in the Pull Request initially. When you set timeout to 0, that does not happen, per:

```console
root@worker-node01:~# crictl run container.json sandbox.json 
810eb62698d7bab94f923b144fcd915551ad2de30e60248cefb76e6c9c66506b
```

```console
root@worker-node01:~# crictl ps
CONTAINER           IMAGE                                CREATED             STATE               NAME                ATTEMPT             POD ID              POD
810eb62698d7b       quay.io/crio/fedora-crio-ci:latest   10 seconds ago      Running             busybox             0                   554ce2b4a1fd1       unknown
root@worker-node01:~# 
```

```console
root@worker-node01:~# crictl stop --timeout=0 810eb62698d7b
810eb62698d7b
root@worker-node01:~# echo $?
0
```

``console
root@worker-node01:~# crictl inspect 810eb62698d7b | grep -Po '(?<="state": ")[^"]*'
CONTAINER_EXITED
```

```console
INFO[2024-08-05 15:13:36.229891070Z] Stopping container: 810eb62698d7b (timeout: 0s)  id=4a7b454a-739c-4881-bdce-f4fb9c8619e4 name=/runtime.v1.RuntimeService/StopContainer
WARN[2024-08-05 15:13:36.235362788Z] Stopping container 810eb62698d7bab94f923b144fcd915551ad2de30e60248cefb76e6c9c66506b with stop signal timed out. Killing... 
WARN[2024-08-05 15:13:36.235403944Z] XXX 1 810eb62698d7bab94f923b144fcd915551ad2de30e60248cefb76e6c9c66506b 
ERRO[2024-08-05 15:13:36.239394660Z] Killing container 810eb62698d7bab94f923b144fcd915551ad2de30e60248cefb76e6c9c66506b failed: `/usr/bin/crio-runc --root /run/runc --systemd-cgroup kill 810eb62698d7bab94f923b144fcd915551ad2de30e60248cefb76e6c9c66506b KILL` failed: time="2024-08-05T15:13:36Z" level=error msg="container not running"
 : exit status 1 
INFO[2024-08-05 15:13:36.249060431Z] Stopped container 810eb62698d7bab94f923b144fcd915551ad2de30e60248cefb76e6c9c66506b: //  id=4a7b454a-739c-4881-bdce-f4fb9c8619e4 name=/runtime.v1.RuntimeService/StopContainer
```

This is what I meant. The `CONTAINER_EXITED` status is most likely set for a container that terminated.

kwilczynski: > > > Curious that I couldn't set a timeout for termination grace period lower than 2 seconds. Haven't looked into why.
> > 
> > 
> > I used this same example to set the `timeout` value as `1` and it worked.
> 
> Sorry, I should have mentioned `terminationGracePeriodSeconds` from the Pod object specification rather than just the "grace period".

An example:

```console
$ k get pod test-3 -o yaml | grep terminationGracePeriodSeconds:
  terminationGracePeriodSeconds: 1
```

```console
INFO[2024-08-05 15:37:15.109684729Z] Stopping container: f9ef70b73d1e7f0085ca69497ce290b7306a7b7061b9173335ec91d9a9566758 (timeout: 2s)  id=b90d2f54-8419-4241-ac22-e22fc760bc64 name=/runtime.v1.RuntimeService/StopContainer
INFO[2024-08-05 15:37:15.223443674Z] Stopped container f9ef70b73d1e7f0085ca69497ce290b7306a7b7061b9173335ec91d9a9566758: default/test-3/test-3  id=b90d2f54-8419-4241-ac22-e22fc760bc64 name=/runtime.v1.RuntimeService/StopContainer
INFO[2024-08-05 15:37:15.223824002Z] Stopping pod sandbox: a4c0fa4d3043a2e7f94359cfc5676658181226328f4627b42d845c1f24c26426  id=6b14eb06-2442-4f0a-b1b4-f340fe1437a6 name=/runtime.v1.RuntimeService/StopPodSandbox
INFO[2024-08-05 15:37:15.223977168Z] Got pod network &{Name:test-3 Namespace:default ID:a4c0fa4d3043a2e7f94359cfc5676658181226328f4627b42d845c1f24c26426 UID:550731f3-136d-412b-b684-a1ee7b85c0c4 NetNS:/var/run/netns/034fa416-0c1f-442c-81d5-f2237ac83c14 Networks:[{Name:crio Ifname:eth0}] RuntimeConfig:map[crio:{IP: MAC: PortMappings:[] Bandwidth:<nil> IpRanges:[] CgroupPath:kubepods-besteffort-pod550731f3_136d_412b_b684_a1ee7b85c0c4.slice PodAnnotations:0xc00014a560}] Aliases:map[]} 
INFO[2024-08-05 15:37:15.224052221Z] Deleting pod default_test-3 from CNI network "crio" (type=bridge) 
INFO[2024-08-05 15:37:15.311514819Z] Stopped pod sandbox: a4c0fa4d3043a2e7f94359cfc5676658181226328f4627b42d845c1f24c26426  id=6b14eb06-2442-4f0a-b1b4-f340fe1437a6 name=/runtime.v1.RuntimeService/StopPodSandbox
INFO[2024-08-05 15:37:16.207227134Z] Removing container: f9ef70b73d1e7f0085ca69497ce290b7306a7b7061b9173335ec91d9a9566758  id=3b6147f7-dc59-48ac-9cde-289f3da8d9f9 name=/runtime.v1.RuntimeService/RemoveContainer
INFO[2024-08-05 15:37:16.214599391Z] Removed container f9ef70b73d1e7f0085ca69497ce290b7306a7b7061b9173335ec91d9a9566758: default/test-3/test-3  id=3b6147f7-dc59-48ac-9cde-289f3da8d9f9 name=/runtime.v1.RuntimeService/RemoveContainer
```

I haven't looked why that is, to be completely honest. Albeit, very curious, indeed.

sohankunkerkar: @kwilczynski Thanks for the thorough examples. Let me re-visit this logic.

sohankunkerkar: I think for `terminationGracePeriodSeconds`, some of the pieces are covered in https://github.com/kubernetes/kubernetes/issues/83916

sohankunkerkar: >I am glad it works! However, we were trying to suppress spurious logs here in the Pull Request initially. When you set timeout to 0, that does not happen, per:

I'm leaning towards the combination of the existing logic + the error matching that we discussed earlier, thoughts?

internal/oci/container.go
```diff
@@ -35,6 +35,9 @@ var (
 	ErrContainerStopped = errors.New("container is already stopped")
 	ErrNotFound         = errors.New("container process not found")
 	ErrNotInitialized   = errors.New("container PID not initialized")
+	ErrorNotRunning     = errors.New("container is not running")
+	ErrorNoSuchProcess  = errors.New("no such process")
+	ErrorInvalidProcess = errors.New("invalid process")
```
haircommander, line 40: nit: the others are just `Err*` not `Error*`

haircommander:
/approve

internal/oci/runtime_oci.go
kwilczynski: @sohankunkerkar, that would be fine, indeed.

internal/oci/runtime_oci.go
```diff
@@ -953,14 +957,23 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
...
+	// temporary workaround based on error string matching.
+	isProcessGoneError := func(err error) bool {
+		return errors.Is(err, ErrNoSuchProcess) ||
+			errors.Is(err, ErrNotRunning) ||
+			errors.Is(err, ErrInvalidProcess)
+	}
 	// We cannot use ExponentialBackoff() here as its stop conditions are not flexible enough.
 	kwait.BackoffUntil(func() {
-		if _, err := r.runtimeCmd("kill", c.ID(), "KILL"); err != nil {
+		if _, err := r.runtimeCmd("kill", c.ID(), "KILL"); err != nil && !isProcessGoneError(err) {
```
kwilczynski, line 970: @sohankunkerkar, do you plan to update `runtimeCmd()` to do a string match and return the proper error type? 

Otherwise, the `errors.Is()` wouldn't do us any good, as it does not do any string match; there is nothing that calls `<T>.Error()` and does a comparison of any sort with the string value there. An example:

```go
package main

import (
	"errors"
	"fmt"
)

var e = errors.New("test error")
var x = errors.New("test error")

func t1() error {
	return errors.New("test error")
}

func t2() error {
	return e
}

func t3() error {
	return errors.New("this is a test error of sorts")
}

func t4() error {
	return x
}

func main() {
	fmt.Println(errors.Is(t1(), e))
	fmt.Println(errors.Is(t2(), e))
	fmt.Println(errors.Is(t3(), e))
	fmt.Println(errors.Is(t4(), e))
}
```

```console
false
true
false
false
```

Internally, the error string is backed by a pointer to the internal `errors.errorString` type, which holds the actual text. This ensures that no two errors (types) with the same string are to be considered identical.

internal/oci/runtime_oci.go
kwilczynski: > I think for `terminationGracePeriodSeconds`, some of the pieces are covered in [kubernetes/kubernetes#83916](https://github.com/kubernetes/kubernetes/issues/83916)

@sohankunkerkar, thank you for this!

That looks interesting. Not quite the same thing, since for the simple example, I wasn't updating the value. Also, this is something that we should support in CRI-O correctly, too. Hopefully we are. :smile:

internal/oci/runtime_oci.go
sohankunkerkar: Yeah, that's a fair point. Let me update the code to reflect this.

sohankunkerkar: @kwilczynski PTAL

internal/oci/runtime_oci.go
```diff
@@ -1402,7 +1415,19 @@ func (r *runtimeOCI) runtimeCmd(args ...string) (string, error) {
...
+		case strings.Contains(stderrStr, "no such process"):
+			runtimeErr = ErrNoSuchProcess
+		case strings.Contains(stderrStr, "container not running"):
+			runtimeErr = ErrNotRunning
+		case strings.Contains(stderrStr, "invalid process"):
+			runtimeErr = ErrInvalidProcess
+		default:
+			runtimeErr = err
+		}
+		return "", fmt.Errorf("`%v %v` failed: %v %v: %w", r.handler.RuntimePath, strings.Join(runtimeArgs, " "), stderr.String(), stdout.String(), runtimeErr)
```
kwilczynski, line 1430: We can probably keep it simple, for example:

```go
err := cmd.Run()
if err != nil {
	// TODO: Remove this function when we have standardized exit codes or
	// error reporting from container runtimes. Current approach is a
	// temporary workaround based on error string matching.
	errString := stderr.String()
	switch {
	// crun, for most of the commands.
	case strings.Contains(errString, "no such process"):
		fallthrough
	// runc, for most of the commands.
	case strings.Contains(errString, "container not running"):
		fallthrough
	// runc, on a rare occassion.
	case strings.Contains(errString, "invalid process"):
		err = ErrNotFound
	}
	return "", fmt.Errorf("`%v %v` failed: %v %v: %w", r.handler.RuntimePath, strings.Join(runtimeArgs, " "), stderr.String(), stdout.String(), err)
}
```

The kill loop would get (same as the background goroutine):

```go
	log.Debugf(ctx, "Error while killing container %s: %v", c.ID(), err)
	if !errors.Is(err, ErrNotFound) {
	}
}
```

The `log.Debugf()` is there to ensure visibility in case something is failing and we do not eat the error up. Can be optional, of course.

internal/oci/runtime_oci.go
```diff
@@ -953,14 +957,23 @@ func (r *runtimeOCI) StopLoopForContainer(c *Container, bm kwait.BackoffManager)
...
 	}
 killContainer:
+	// TODO: Remove this function when we have standardized exit codes or
+	// error reporting from container runtimes. Current approach is a
+	// temporary workaround based on error string matching.
+	isProcessGoneError := func(err error) bool {
+		return errors.Is(err, ErrNoSuchProcess) ||
+			errors.Is(err, ErrNotRunning) ||
+			errors.Is(err, ErrInvalidProcess)
+	}
```
kwilczynski, line 967: We might not need this. See some other comments.

internal/oci/container.go
```diff
@@ -35,6 +35,9 @@ var (
 	ErrContainerStopped = errors.New("container is already stopped")
 	ErrNotFound         = errors.New("container process not found")
 	ErrNotInitialized   = errors.New("container PID not initialized")
+	ErrNotRunning       = errors.New("container not running")
+	ErrNoSuchProcess    = errors.New("no such process")
+	ErrInvalidProcess   = errors.New("invalid process")
```
kwilczynski, line 40: Since we already have `ErrNotFound` and use it, we can keep it and reuse it for our needs, too.

See some other comments.

internal/oci/runtime_oci.go
kwilczynski: @sohankunkerkar, I left a few comments/ideas.

internal/oci/runtime_oci.go
sohankunkerkar: Yeah, this is much better.

internal/oci/runtime_oci.go
```diff
@@ -1402,6 +1410,18 @@ func (r *runtimeOCI) runtimeCmd(args ...string) (string, error) {
 
 	err := cmd.Run()
 	if err != nil {
+		stdErrStr := stderr.String()
+		switch {
+		// crun, for most of the commands.
+		case strings.Contains(stdErrStr, "no such process"):
+			fallthrough //nolint:gocritic
```
kwilczynski, line 1417: Very interesting to know that go-critic has some opinions about this, which are not necessarily correct.

Related:

- https://github.com/go-critic/go-critic/issues/645

A lot of parser code would use this, and the Go standard library is filled with such language constructs.

I am glad we disabled this nonsense. :+1:

kwilczynski:
@sohankunkerkar very nice! Thank you for taking care of this! :bow:

PR #8437, opened 2024-07-31
Block clone with namespace flag in seccomp default profile

PannagaRao:
#### What type of PR is this?

/kind bug

The default seccomp policy used by CRI-O did not previously block clone(2) system call (syscall) with CLONE_NEWNS and other namespace creation flags. Changes here block those flags. In addition, also disable the clone3(2) syscall as it is not currently possible to filter a struct argument (struct clone_args) with seccomp; this matches the approach containerd took.

None

None

```release-note
The default seccomp policy now blocks clone and clone3 system calls that create a Linux namespace. This matches the default seccomp policy containerd uses.
```

### DIFF
--- a/test/testdata/clone-ns.c
+++ b/test/testdata/clone-ns.c
@@ -1,0 +1,17 @@
+#define _GNU_SOURCE
+#include <unistd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sched.h>
+
+int entry() {
+        system("id");
+        return 0;
+}
+
+int main() {
+        void *stack = malloc(1024*1024);
+        if (clone(entry, (stack+1024*1024), CLONE_NEWUSER|CLONE_NEWNET, 0) == -1) {
+          perror("clone");
+        }
+}
--- a/internal/config/seccomp/seccomp.go
+++ b/internal/config/seccomp/seccomp.go
@@ -42,7 +42,7 @@ func DefaultProfile() *seccomp.Seccomp {
 		}{
 			{"clone", 1, 23},
 			{"clone3", 1, 24},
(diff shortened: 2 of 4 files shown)


kwilczynski:
This change would break the ability to run anything within the container using Exec or ExecSync, or pretty much anything that will ever attempt to use fork()/vfork()/exec*()/posix_spawn(), anything backed by clone system call, that is.

PannagaRao:
@dgl Could you please let me know if there were any checks to see if the implementation blocks only such clone() calls which have namespace flags as arguments.

test/testdata/clone-ns.c
```diff
@@ -0,0 +1,17 @@
...
+#include <sched.h>
+
+int entry() {
+        system("id");
+        return 0;
+}
+
+int main() {
+        void *stack = malloc(1024*1024);
+        if (clone(entry, (stack+1024*1024), CLONE_NEWUSER|CLONE_NEWNET, 0) == -1) {
```
haircommander, line 14: can we toggle the clone flags on a cli flag so we can compress the two files and trigger each case from the same executable

dgl:
[@kwilczynsk](https://github.com/cri-o/cri-o/pull/8437#issuecomment-2259764006):
> This change would break the ability to run anything within the container using Exec or ExecSync, or pretty much anything that will ever attempt to use fork()/vfork()/exec*()/posix_spawn(), anything backed by clone system call, that is.

The idea was it allows all clone args that aren't creating a namespace, then blocks the disallowed args (unless the container has CAP_SYS_ADMIN).

kwilczynski:
Superseded by:

- https://github.com/cri-o/cri-o/pull/8514.

PR #8438, opened 2024-07-31
Enable and fix `gomoddirectives`, `gomodguard`, `noctx` and `wastedassign` linters

saschagrunert:
#### What type of PR is this?

/kind cleanup

Enabling the linters and fixing the reports.

None

None

```release-note
None
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -552,13 +552,11 @@ func (c *container) AddUnifiedResourcesFromAnnotations(annotationsMap map[string
 		}
 		d, err := b64.StdEncoding.DecodeString(parts[1])
 		// if the value is not specified in base64, then use its raw value.
-		v := ""
 		if err == nil {
-			v = string(d)
+			c.spec.Config.Linux.Resources.Unified[parts[0]] = string(d)
 		} else {
-			v = parts[1]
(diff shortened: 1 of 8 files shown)


internal/factory/container/container.go
```diff
@@ -552,13 +552,11 @@ func (c *container) AddUnifiedResourcesFromAnnotations(annotationsMap map[string
...
 		// if the value is not specified in base64, then use its raw value.
-		v := ""
 		if err == nil {
-			v = string(d)
+			c.spec.Config.Linux.Resources.Unified[parts[0]] = string(d)
 		} else {
-			v = parts[1]
+			c.spec.Config.Linux.Resources.Unified[parts[0]] = parts[1]
 		}
-		c.spec.Config.Linux.Resources.Unified[parts[0]] = v
```
kwilczynski, line 561: This linter suggestion is an interesting one. The original code does not look all that bad, definitely more readable.

saschagrunert: That was more or less my decision to get rid of `v`. Do you want to have it back as `var v string`?

kwilczynski: No worries. Personally, I find the old version slightly easier to read, let it be "easier on the eyes", so to speak. But, we can keep the new one, if that is your preference. :smile:

PR #8448, opened 2024-08-05
Run critest in parallel

saschagrunert:
#### What type of PR is this?

/kind ci

Make the critest suite succeed faster by running them in parallel.

None
None

```release-note
None
```

### DIFF
--- a/test/critest.bats
+++ b/test/critest.bats
@@ -19,6 +19,11 @@ function teardown() {
 	critest \
 		--runtime-endpoint "unix://${CRIO_SOCKET}" \
 		--image-endpoint "unix://${CRIO_SOCKET}" \
+		--parallel="$(nproc)" \
+		--ginkgo.randomize-all \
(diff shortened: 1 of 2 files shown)


test/critest.bats
```diff
@@ -19,6 +19,11 @@ function teardown() {
 	critest \
 		--runtime-endpoint "unix://${CRIO_SOCKET}" \
 		--image-endpoint "unix://${CRIO_SOCKET}" \
+		--parallel="$(nproc)" \
```
kwilczynski, line 22: Do we want to make it **-1** to keep some CPU for the rest of the things running and servicing I/O?

kwilczynski:
/hold

For the review comment.

test/critest.bats
saschagrunert: Hm, I don't think the tests use that much CPU so I'd like to keep it simple :)

PR #8470, opened 2024-08-09
start hostport manager cleanup

danwinship:
#### What type of PR is this?
/kind cleanup

Some cleanup/refactoring to prepare for the eventual iptables→nftables port:

**First**, this drops the portOpener stuff. In the old days, the kubelet hostport manager and kube-proxy nodeport manager both had code to open listening sockets on all ports that were being used as hostports/nodeports, so as to ensure that the same port could not be bound as both a hostport and a nodeport, and so that you couldn't use a system port (like "22") as a hostport/nodeport. But this didn't work very well for various reasons, and that code was removed from kube-proxy a while back; the consensus was that it was better to limit NodePorts to specific ranges (which we do), and limit HostPorts to only be available to privileged users (which PSA does) because otherwise you can do damage with them even with the portOpeners in place. So anyway, this just syncs the cri-o hostport code up with that decision.

(FTR one of the problems with the old code is that it didn't resync after a restart, and I'm pretty sure this is still true of the cri-o code, so if you don't want to remove the portOpener stuff, you should fix that. :slightly_smiling_face: .)

**Second**, this squashes together the original single-stack HostPortManager and the "new" dual-stack MetaHostPortManager into a single object, since cri-o never actually uses the single-stack manager on its own. (And the future nftables version will be inherently dual-stack since you don't need to use separate helper binaries for IPv4 and IPv6.)

None

```release-note
NONE
```


kwilczynski:
@danwinship, you need to add Signed-of-by tag to your commits.

saschagrunert (approved):

PR #8480, opened 2024-08-13
Fix incorrect tab handling while converting Info to string and add unit tests

xw19:
/kind bug

Fixes incorrect tab character handling in the internal/version/String
function when dealing with array values. Previously, a tab was added
after field names containing arrays, which was unnecessary as array
values should be on the next line.

Additional tab character on String method. Which was causing additional space being created after key name for slice values.

```release-note
None
```

### DIFF
--- a/internal/version/version_test.go
+++ b/internal/version/version_test.go
@@ -12,7 +12,62 @@ var _ = t.Describe("Version", func() {
 	tempFileName := "tempVersionFile"
 	tempVersion := "1.1.1"
 	tempVersion2 := "1.13.1"
-
+	testInfo := Info{
+		Version:         "1.0.0",
+		GitCommit:       "abcdef123456",
+		GitCommitDate:   "2024-08-13T12:34:56Z",
+		GitTreeState:    "clean",
+		BuildDate:       "2024-08-13T12:34:56Z",
+		GoVersion:       "go1.20.5",
+		Compiler:        "gc",
+		Platform:        "linux/amd64",
+		Linkmode:        "external",
+		BuildTags:       []string{"tag1", "tag2"},
+		LDFlags:         "-X main.Version=1.0.0",
+		SeccompEnabled:  true,
+		AppArmorEnabled: false,
+		Dependencies:    []string{"dep1", "dep2"},
+	}
+	testInfoStr := `Version:        1.0.0
+GitCommit:      abcdef123456
+GitCommitDate:  2024-08-13T12:34:56Z
+GitTreeState:   clean
+BuildDate:      2024-08-13T12:34:56Z
+GoVersion:      go1.20.5
+Compiler:       gc
+Platform:       linux/amd64
+Linkmode:       external
+BuildTags:
+  tag1
+  tag2
+LDFlags:          -X main.Version=1.0.0
+SeccompEnabled:   true
+AppArmorEnabled:  false
+Dependencies:
+  dep1
+  dep2`
+	testJSONInfoStr := `{
+  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


internal/version/version_test.go
```diff
@@ -170,4 +226,15 @@ var _ = t.Describe("Version", func() {
 			Expect(err).To(HaveOccurred())
 		})
 	})
+	t.Describe("test generating string from info", func() {
+		It("should succeed returning a formatted string", func() {
+			infoString := testInfo.String()
+			Expect(infoString).To(Equal(testInfoStr))
+		})
+		It("should succeed returning a JSON formatted string", func() {
```
kwilczynski, line 234:
```suggestion
		It("should succeed returning a JSON document", func() {
```

But also, why a different description?

internal/version/version_test.go
```diff
@@ -12,6 +12,62 @@ var _ = t.Describe("Version", func() {
...
+		"    \"tag2\"" + "\n" +
+		"  ]," + "\n" +
+		"  \"ldFlags\": \"-X main.Version=1.0.0\"," + "\n" +
+		"  \"seccompEnabled\": true," + "\n" +
+		"  \"appArmorEnabled\": false," + "\n" +
+		"  \"dependencies\": [" + "\n" +
+		"    \"dep1\"," + "\n" +
+		"    \"dep2\"" + "\n" +
+		"  ]" + "\n" +
+		"}"
```
kwilczynski, line 70: This is not how Go developers would often approach larger blocks of text.

internal/version/version_test.go
xw19: Fixed

internal/version/version_test.go
xw19: I thought it just looked better now using text block.

xw19:
Tested command `crio -h`

```
...
VERSION:
   1.31.0
   GitCommit:      35eda06ca31195d5a2c42a1cc93c42feee05f86d
   GitCommitDate:  2024-08-14T11:04:55Z
   GitTreeState:   dirty
   BuildDate:      2024-08-14T11:52:42Z
   GoVersion:      go1.22.5
   Compiler:       gc
   Platform:       linux/amd64
   Linkmode:       dynamic
   BuildTags:
     containers_image_ostree_stub
     containers_image_openpgp
     seccomp
     selinux
   LDFlags:          unknown
   SeccompEnabled:   true
   AppArmorEnabled:  false
...

```

saschagrunert (approved):

kwilczynski:
@xw19, we need to update the commit subject and commit log to reflect the fact that you also fixed a problem you found.

xw19:
@kwilczynski I have added detailed commit message and modified the PR Subject

kwilczynski:
> @kwilczynski I have added detailed commit message and modified the PR Subject

@xw19, what we need, would be an updated concise commit subject, without the "internal/version/String", this is not needed, and then update the Pull Request with the new details.

Look how other Pull Request that were accepted looked like.

xw19:
Thank you for your guidance,

 @kwilczynski. I’ve added the requested information. Please let me know if this is correct.

This version is more concise.

kwilczynski:
> Thank you for your guidance,
> 
> @kwilczynski. I’ve added the requested information. Please let me know if this is correct.
> 
> This version is more concise.

@xw19, looks good! You can drop the DCO mention from the commit message (the actual commit, not the PR). We know what it is. :smile:

xw19:
@kwilczynski Done

kwilczynski:
@xw19, perfect. Let's ship!

Issue #7448, opened 2023-10-30
Use 127.0.0.1 as the default address to listen on for other listeners


kwilczynski:
Some time ago, the streaming endpoint was moved to only listen on localhost (127.0.0.1) and a random port to improve the default security stance in CRI-O; see:

- https://github.com/cri-o/cri-o/pull/1714

This change has also been incorporated in OpenShift via the Machine Config Operator (MCO):

- https://github.com/openshift/machine-config-operator/pull/3853

However, there are still other listeners that can be enabled in CRI-O that would listen on any address (0.0.0.0), often without a way to move them to either localhost or any other address for that matter, per:

- The Go profiling (pprof) endpoint:

**Note:** This does not affect accessing the profiling endpoint via the Unix socket CRI-O uses.

https://github.com/cri-o/cri-o/blob/5633698da2814bc4e97f56f701d747853069a91c/internal/criocli/criocli.go#L804-L809

https://github.com/cri-o/cri-o/blob/5633698da2814bc4e97f56f701d747853069a91c/cmd/crio/main.go#L233-L242

- The Prometheus (metrics) endpoint:

**Note:** This does not affect accessing the metrics endpoint via the Unix socket that can be additionally configured.

https://github.com/cri-o/cri-o/blob/5633698da2814bc4e97f56f701d747853069a91c/pkg/config/config.go#L889-L892

https://github.com/cri-o/cri-o/blob/5633698da2814bc4e97f56f701d747853069a91c/server/metrics/metrics.go#L364-L370

- The Tracing endpoint:

https://github.com/cri-o/cri-o/blob/5633698da2814bc4e97f56f701d747853069a91c/pkg/config/config.go#L893-L897

https://github.com/cri-o/cri-o/blob/5633698da2814bc4e97f56f701d747853069a91c/internal/oci/runtime_pod.go#L61-L74

Tracing shares implementation details with the [conmon-rs](https://github.com/containers/conmon-rs) project:

- [conmon-rs/pkg/client/client.go#L101-L111](https://github.com/containers/conmon-rs/blob/5d9cd8d30fa57ae548514124dc3bcba093366a61/pkg/client/client.go#L101-L111)

```go
type Tracing struct {
	// Enabled tells the server to run with OpenTelemetry tracing.
	Enabled bool

	// Endpoint is the GRPC tracing endpoint for OLTP.
	// Defaults to "http://localhost:4317"
	Endpoint string

	// Tracer allows the client to create additional spans if set.
[...]

PR #8495, opened 2024-08-19
Change default tracing endpoint to 127.0.0.1

bitoku:
#### What type of PR is this?

/kind other

It is prudent to move this endpoint to listen on 127.0.0.1 as a sensible default, which would improve the default security stance a little more, especially as there aren't any reasons for these services to listen on anything else than localhost.

We don't have to change conmon-rs because it already uses localhost.

https://github.com/containers/conmon-rs/blob/main/conmon-rs/server/src/config.rs#L119-L125

Fixes a part of #7448 

Users who are using this endpoint via not localhost(127.0.0.1) will need to set the tracingEndpoint option explicitly.

```release-note
Move the tracing endpoint listener to use 127.0.0.1 as the new default.
```

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -426,7 +426,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--timezone, --tz**="": To set the timezone for a container in CRI-O. If an empty string is provided, CRI-O retains its default behavior. Use 'Local' to match the timezone of the host machine.
 
-**--tracing-endpoint**="": Address on which the gRPC tracing collector will listen. (default: "0.0.0.0:4317")
+**--tracing-endpoint**="": Address on which the gRPC tracing collector will listen. (default: "127.0.0.1:4317")
 
 **--tracing-sampling-rate-per-million**="": Number of samples to collect per million OpenTelemetry spans. Set to 1000000 to always sample. (default: 0)
 
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -533,7 +533,7 @@ The certificate key for the secure metrics server.
 **enable_tracing**=false
 Globally enable or disable OpenTelemetry trace data exporting.
 
-**tracing_endpoint**="0.0.0.0:4317"
+**tracing_endpoint**="127.0.0.1:4317"
 Address on which the gRPC trace collector will listen.
 
 **tracing_sampling_rate_per_million**=""
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -946,7 +946,7 @@ func DefaultConfig() (*Config, error) {
 			MetricsCollectors: collectors.All(),
 		},
 		TracingConfig: TracingConfig{
-			TracingEndpoint:               "0.0.0.0:4317",
+			TracingEndpoint:               "127.0.0.1:4317",
 			TracingSamplingRatePerMillion: 0,
 			EnableTracing:                 false,
 		},
--- a/tutorials/tracing.md
+++ b/tutorials/tracing.md
@@ -14,7 +14,7 @@ enable_tracing = true
 ```
 
 Traces in CRI-O get exported via the [OpenTelemetry Protocol][otlp] by using an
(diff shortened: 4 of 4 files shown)


mrunalp:
Will this work with ipv6?

kwilczynski:
> Will this work with ipv6?

On this side? Yes. On conmon and/or conmon-rs side? I am not sure. The OTEL package for Rust could potentially support dual-stack, too. But I don't know.

However, there is no need for this endpoint to support anything other than connecting via 127.0.0.1, especially given that this is a tracing endpoint for which primary clients are conmon and/or conmon-rs. This is not something that users would typically need to send requests to manually.

Prior art about IPv6 on local listeners for reference:

- https://github.com/openshift/machine-config-operator/pull/3853#issuecomment-1761346423
- https://github.com/openshift/machine-config-operator/pull/3853#issuecomment-1761622765

kwilczynski:
@bitoku, we probably should change conmon-rs to explicitly say `127.0.0.1` for the same reasons we changed some other clients or listeners where we moved from `localhost` to an explicit loopback IP address.

bitoku:
@kwilczynski
Thank you for the explanation about IPv4 and 6.

I'm thinking about opening new PRs for each endpoints.
If we want to have it in the same RP, I'll add them.

bitoku:
I opened a PR for conmon-rs https://github.com/containers/conmon-rs/pull/2391

kwilczynski:
> @kwilczynski Thank you for the explanation about IPv4 and 6.

I will add another interesting bit concerning the resolution of the `localhost` name. Some software has it hardcoded, such as when it seems "localhost" as an argument, it automatically assumes `127.0.0.1`, and no name resolution lookup will occur. For the software that does not have this optimisation, so to speak, a lookup will be performed.

Normally, on systems that run GNU C library (glibc), which would be most Linux distributions, whether IPv6 is enabled or not, because of `/etc/gai.conf` (see [man 5 gai.conf](https://www.man7.org/linux/man-pages/man5/gai.conf.5.html)) by default favouring IPv6 lookups (so `AAAA` over `A` for any other external lookup), two addresses would be returned: `127.0.0.1` and `::1` (which is `0:0:0:0:0:0:0:1` with all the abbreviated zeros removed).

Thus, depending on the order (which should be normally set by the /etc/gai.conf), you might get the IPv6 loopback as the first address or the second (some software might choose to sort responses such that `A` is favourite over `AAAA` or generally IPv4 over IPv6). If you get IPv6 one, then a `connect()` a lot of the time would result in an RST sent back behind the scenes (especially if the listener is not dual-stack enabled and only listens on IPv4), which resets the connection attempt and depending on the implementation it might be an error the software would treat as terminal and exit, or connection attempt will be retried on the second address, which is the IPv4 one that was returned. This can add some delay, a few milliseconds usually, and as much as it's not an issue for a connection that is established once and reused, then for a client that makes a connection every time it needs to contact the server, then these milliseconds of delay can compound to a larger value eventually adding up to some latency for a service, etc.

An example of this would be curl users on Windows have most definitely stumbled into this issue and learned the hard way about how this works, usually: [Slow connection to localhost on Windows (happy eyeballs algo)](https://github.com/curl/curl/issues/14144) or [slow TCP connect on Windows](https://daniel.haxx.se/blog/2024/08/14/slow-tcp-connect-on-windows/). I mention Windows here as this is a lot more impactful there than it is on Linux, to many Windows users' dismay.

For reference:

A `/etc/gai.conf` that lowers the IPv6 priority for lookups can be used on any desktop or server computer that does not use IPv6 to speed up DNS lookups a bit and remove unnecessary retries, etc.

```
reload no

label ::1/128       0
label ::/0          1
label 2002::/16     2
label ::/96         3
label ::ffff:0:0/96 4
label fec0::/10     5
label fc00::/7      6
label 2001:0::/32   7

precedence  ::1/128       50
precedence  ::/0          40
precedence  2002::/16     30
precedence ::/96          20
precedence ::ffff:0:0/96  100

scopev4 ::ffff:169.254.0.0/112  2
scopev4 ::ffff:127.0.0.0/104    2
scopev4 ::ffff:0.0.0.0/96       14
```

saschagrunert (approved):

PR #8506, opened 2024-08-20
Change profile endpoint to 127.0.0.1

bitoku:
#### What type of PR is this?

/kind other

This PR changes profile endpoint to 127.0.0.1 according to #7448  

Fixes a part of #7448

```release-note
Move the tracing profile listener to use 127.0.0.1 as the new default.
```


cmd/crio/main.go
```diff
@@ -240,7 +240,7 @@ func main() {
 
 		if c.Bool("profile") {
 			profilePort := c.Int("profile-port")
-			profileEndpoint := fmt.Sprintf("localhost:%v", profilePort)
+			profileEndpoint := fmt.Sprintf("127.0.0.1:%v", profilePort)
```
saschagrunert, line 243:
```suggestion
			profileEndpoint := fmt.Sprintf("127.0.0.1:%d", profilePort)
```

saschagrunert:
Please also update the completions and docs (see GitHub actions CI).

kwilczynski:
@bitoku @saschagrunert, we allow for the address on which it will listen (bind) to be specified for some other listeners.

An example of this would be the metrics endpoint.

Should we also allow for this to be customised? I cannot see why someone would like to have the profiling endpoint listen on anything other than localhost, but perhaps for the sake of consistency, it should also be done for this endpoint.

Thoughts?

saschagrunert:
> Should we also allow for this to be customised? I cannot see why someone would like to have the profiling endpoint listen on anything other than localhost, but perhaps for the sake of consistency, it should also be done for this endpoint.

Probably, yes. Maybe for testing or something :thinking:

bitoku:
IMHO, we don't have to unless we receive reasonable requests to allow to configure the endpoint.
We've been left this `localhost` since the beginning, and do we get the request?

Making new options will also increase the cost of maintenance (even though it's very little, but every little bit counts).

kwilczynski:
> > Should we also allow for this to be customised? I cannot see why someone would like to have the profiling endpoint listen on anything other than localhost, but perhaps for the sake of consistency, it should also be done for this endpoint.
> 
> Probably, yes. Maybe for testing or something 🤔

@bitoku @saschagrunert same question for the following:

- https://github.com/cri-o/cri-o/pull/8495

bitoku:
Tracing endpoint is already customisable.
I don't think it has to be as well, but because we already have the feature, it's hard to get rid of it.
That's why I feel reluctant to add customisable features which are not that wanted by users.

I'm not strongly opposing because consistency is also important.

kwilczynski:
[...]
> I'm not strongly opposing because consistency is also important.

@bitoku @saschagrunert, one of the goals of the GitHub issue was indeed to make things consistent, if possible.

bitoku:
@kwilczynski 
ok! Do you want me to include that in this PR?
I think it's good to keep them separate to make things small and simple.

kwilczynski:
> @kwilczynski ok! Do you want me to include that in this PR? I think it's good to keep them separate to make things small and simple.


@bitoku @saschagrunert, only if the two of you think it's the right thing to do. We can have a consensus over this.

We do not _have_ to do it. If we feel it's not needed or we are trying to do too much. :smile:

bitoku:
@kwilczynski 
I think we can discuss it separately from this PR. Or do you want to add that change with this change?

kwilczynski:
> @kwilczynski I think we can discuss it separately from this PR. Or do you want to add that change with this change?

@bitoku, sounds good!

saschagrunert (approved):

PR #8514, opened 2024-08-23
Block clone with namespace flag in seccomp default profile

bitoku:
#### What type of PR is this?

/kind bug

Taking over https://github.com/cri-o/cri-o/pull/8437

> The default seccomp policy used by CRI-O did not previously block clone(2) system call (syscall) with CLONE_NEWNS and other namespace creation flags. Changes here block those flags. In addition, also disable the clone3(2) syscall as it is not currently possible to filter a struct argument (struct clone_args) with seccomp; this matches the approach containerd took.

None

None

```release-note
The default seccomp policy now blocks clone and clone3 system calls that can create a Linux namespace. This matches the default seccomp policy containerd uses.
```

### DIFF
--- a/internal/config/seccomp/seccomp.go
+++ b/internal/config/seccomp/seccomp.go
@@ -17,6 +17,7 @@ import (
 	json "github.com/json-iterator/go"
 	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 
 	"github.com/cri-o/cri-o/internal/config/seccomp/seccompociartifact"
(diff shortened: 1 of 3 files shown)


internal/config/seccomp/seccomp.go
```diff
@@ -29,50 +30,106 @@ var (
...
+			})
 		defaultProfile = prof
 	})
 
 	return defaultProfile
 }
 
+// validateSyscallIndex checks if the syscall's index matches the default profile's index.
+// We know the default profile at compile time though a vendor change may update it.
+// Panic on error and have CI catch errors on vendor bumps to avoid combing through.
```
kwilczynski, line 115:
```suggestion
// validateSyscallIndex checks if the syscall's index matches the default profile's index.
// We know the default profile at compile time, though a vendor change may update it.
// Panic on error and have CI catch errors on vendor bumps to avoid combing through.
```

internal/config/seccomp/seccomp.go
```diff
@@ -29,50 +30,106 @@ var (
...
+		if n == name {
+			_, file, _, _ := runtime.Caller(1)
+			logrus.Errorf("Change the Index for %q in %s to %d", name, file, i)
+			break
+		}
+	}
+	logrus.Fatalf(
+		"Default seccomp profile updated and syscall moved. Found unexpected syscall: %q",
+		prof.Syscalls[parentStructIndex].Names[index],
+	)
```
kwilczynski, line 130: I wonder if we should collapse both errors in a single error, perhaps a `Fatal()` log, that is the best of both, so to speak.

internal/config/seccomp/seccomp.go
```diff
@@ -29,50 +31,110 @@ var (
...
+	if prof.Syscalls[parentStructIndex].Names[index] == name {
+		return
+	}
+	_, file, _, _ := runtime.Caller(1)
+	i := slices.Index(prof.Syscalls[parentStructIndex].Names, name)
+	if i == -1 {
+		logrus.Fatalf("Default seccomp profile updated and syscall moved. Change the parentStructIndex for %q in %s",
+			name, file)
+	}
+	logrus.Fatalf(
```
haircommander, line 132: this is being logged unconditionally it seems

bitoku: It's not unconditional. If the indexes are correct, it returns here (using early return style not to indent a lot).

https://github.com/cri-o/cri-o/blob/56162933764ae95f5697f778c643391f3acacb2c/internal/config/seccomp/seccomp.go#L123-L125

Otherwise it should always fail IIUC, or am I missing something?

haircommander:
/approve

LGTM, thanks @bitoku @PannagaRao and @dgl

internal/config/seccomp/seccomp.go
```diff
@@ -29,50 +31,106 @@ var (
...
+// We know the default profile at compile time, though a vendor change may update it.
+// Panic on error and have CI catch errors on vendor bumps to avoid combing through.
+func validateSyscallIndex(prof *seccomp.Seccomp, name string, parentStructIndex, index int) {
+	if prof.Syscalls[parentStructIndex].Names[index] == name {
+		return
+	}
+	i := slices.Index(prof.Syscalls[parentStructIndex].Names, name)
+	if i == -1 {
+		logrus.Fatalf("Default seccomp profile updated and syscall moved. Change the parentStructIndex for %q", name)
+	}
```
kwilczynski, line 129:
```suggestion
	if i := slices.Index(prof.Syscalls[parentStructIndex].Names, name); i == -1 {
		logrus.Fatalf("Default seccomp profile updated and syscall moved. Change the parentStructIndex for %q", name)
	}
```

internal/config/seccomp/seccomp.go
```diff
@@ -29,50 +31,106 @@ var (
...
+	logrus.Fatalf(
+		"Default seccomp profile updated and syscall moved. Change the Index for %q to %d", name, i)
```
kwilczynski, line 131:
```suggestion
	logrus.Fatalf("Default seccomp profile updated and syscall moved. Change the Index for %q to %d", name, i)
```

internal/config/seccomp/seccomp.go
kwilczynski: So, when the user sees this error message and CRI-O exists, what ought they do? What does it mean to "change the parentStructIndex"?

haircommander: yeah maybe this should be "contact the CRI-O maintainers if you see this" sort of message.

bitoku: This is not a message for users. This error is supposed to be caught during CI, at least `ctr_seccomp.bats` should catch it if the upstream default is changed.

https://github.com/cri-o/cri-o/blob/6b41d86972a240ed62663fb984a1b829a2837357/internal/config/seccomp/seccomp.go#L119-L121

bitoku: You mean the users are the contributors?
The cri-o users can't see this error because this function always passes if CI passes.
This function just checks the discrepancies of the indexes [1] and default seccomp [2], and doesn't take any user configurations.

However the contributors to CRI-O will see this error message when they upgrade seccomp package (they can't ignore it though because CI fails), and the message can be improved as you say.

[1] https://github.com/cri-o/cri-o/blob/6b41d86972a240ed62663fb984a1b829a2837357/internal/config/seccomp/seccomp.go#L39-L47

[2] https://github.com/cri-o/cri-o/blob/a46e1d98ac29bd6088184266a0a7c2f9592cd90f/vendor/github.com/containers/common/pkg/seccomp/default_linux.go#L50-L58

bitoku: How about something like this?
```go
	if i == -1 {
		logrus.Errorf("Change the ParentStructIndex for %q", name)
	} else {
		logrus.Errorf("Change the Index for %q to %d", name, i)
	}
	logrus.Fatalf(
		`Default seccomp has been changed, and cri-o couldn't remove some risky syscalls.
You are seeing this error because you updated "github.com/containers/common/pkg/seccomp".
Please look up "vendor/github.com/containers/common/pkg/seccomp/default_linux.go,"
and change DefaultProfile() in internal/config/seccomp/seccomp.go according to the message above.`,
		)
```

dgl: > The cri-o users can't see this error because this function always passes if CI passes.

If someone compiles it themselves and potentially does something like changing how the module vendoring works they could see it.

> `Default seccomp has been changed, and cri-o couldn't remove some risky syscalls. You are seeing this error because you updated "github.com/containers/common/pkg/seccomp". Please look up "vendor/github.com/containers/common/pkg/seccomp/default_linux.go," and change DefaultProfile() in internal/config/seccomp/seccomp.go according to the message above.`,

I think that error assumes too much about the situation. Something like this is accurate but gives some leeway for various things going wrong and doesn't accuse the user themselves of changing something:

```
`The default internal seccomp policy has been changed and cri-o can't adjust some risky syscalls. You are likely seeing this error because "github.com/containers/common/pkg/seccomp" was updated. Please contact the developers or change DefaultProfile() in internal/config/seccomp/seccomp.go to match according to the message above.`,
```

haircommander:
hey @bitoku there's one more issue https://github.com/cri-o/cri-o/actions/runs/10606319107/job/29396842445?pr=8514

jlebon:
Though relatedly, are there plans to allow user namespacing by default in the future?

haircommander:
> Though relatedly, are there plans to allow user namespacing by default in the future?

yeah that'll happen in upstream kube one day (probably 1.33ish for API skew reasons)

PR #8521, opened 2024-08-26
Add additional bind mount to image volumes

saschagrunert:
#### What type of PR is this?

/kind feature

This should allow using different SELinux contexts if multiple pods use a single image.

Inherited from https://github.com/cri-o/cri-o/commit/61d4a591b2f57bae57e6718d84e3f41404a2f251#diff-1e0877112035840d3e2f073c693665f8daffa241a440e54a866ad7382b510ea2R1240

Follow-up on https://github.com/cri-o/cri-o/pull/8317

Refers to:
- https://github.com/containers/storage/issues/1994
- https://github.com/kubernetes/kubernetes/issues/126956
- https://github.com/kubernetes/website/pull/47588

/hold

This needs to be well tested.

```release-note
None
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -3,6 +3,7 @@ package server
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -1069,18 +1070,24 @@ func (s *Server) addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container,
 	if err != nil {
 		return nil, nil, err
 	}
+
+	imageVolumesPath, err := s.ensureImageVolumesPath(ctx, mounts)
+	if err != nil {
+		return nil, nil, fmt.Errorf("ensure image volumes path: %w", err)
+	}
+
 	for _, m := range mounts {
 		dest := m.ContainerPath
 		if dest == "" {
 			return nil, nil, errors.New("mount.ContainerPath is empty")
 		}
 		if m.Image != nil && m.Image.Image != "" {
-			mountPoint, imageID, err := s.mountImage(ctx, m.Image.Image, mountLabel)
+			volume, err := s.mountImage(ctx, specgen, imageVolumesPath, m)
 			if err != nil {
 				return nil, nil, fmt.Errorf("mount image: %w", err)
 			}
-			m.Image.Image = imageID // Set the ID for container status later on
-			m.HostPath = mountPoint // Adjust the host path to use the mount point
+			volumes = append(volumes, *volume)
+			continue
 		}
 		if m.HostPath == "" {
 			return nil, nil, errors.New("mount.HostPath is empty")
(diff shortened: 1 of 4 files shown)


server/container_create_linux.go
```diff
@@ -1249,15 +1255,50 @@ func (s *Server) mountImage(ctx context.Context, imageRef, mountLabel string) (h
...
-
 	log.Infof(ctx, "Image mounted to: %s", mountPoint)
+
+	const overlay = "overlay"
+	specgen.AddMount(rspec.Mount{
+		Type:        overlay,
+		Source:      overlay,
+		Destination: containerPath,
+		Options: []string{
+			"lowerdir=" + mountPoint + ":" + imageVolumesPath,
```
carlory, line 1270: We may need to get the kind of the storage driver that is used by the cri-o. I'm not sure whether it can work properly on other backends, i.e. zfs, btrfs, etc.

server/container_create_linux.go
```diff
@@ -1075,7 +1076,7 @@ func (s *Server) addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container,
 			return nil, nil, errors.New("mount.ContainerPath is empty")
 		}
 		if m.Image != nil && m.Image.Image != "" {
-			mountPoint, imageID, err := s.mountImage(ctx, m.Image.Image, mountLabel)
+			mountPoint, imageID, err := s.mountImage(ctx, specgen, m.Image.Image, m.ContainerPath)
 			if err != nil {
 				return nil, nil, fmt.Errorf("mount image: %w", err)
 			}
```
saschagrunert, line 1082: We would need to `continue` the loop here because we don't have a local mount point any more. Not sure if that's the overall right approach because it would make the implementation clumsy from my point of view. :thinking:

server/container_create_linux.go
```diff
@@ -1249,15 +1264,63 @@ func (s *Server) mountImage(ctx context.Context, imageRef, mountLabel string) (h
...
+
+	const overlay = "overlay"
+	specgen.AddMount(rspec.Mount{
+		Type:        overlay,
+		Source:      overlay,
+		Destination: containerPath,
+		Options: []string{
+			"lowerdir=" + mountPoint + ":" + imageVolumesPath,
+		},
+	})
```
saschagrunert, line 1281: @haircommander is that enough to propagate the right label to the container when we call afterwards:

https://github.com/cri-o/cri-o/blob/c940a11495c28f0f1cfeeeb461c805796944c2af/server/container_create_linux.go#L812-L813

saschagrunert: At least it works in my local VM-based test.

server/container_create_linux.go
saschagrunert: I don't think so, the `lowerdir=` specification will mainly work for overlay. But overlay should be available most of the times, right?

saschagrunert:
/retest
/unhold

@cri-o/cri-o-maintainers PTAL

saschagrunert:
@cri-o/cri-o-maintainers PTAL, CI is green

server/container_create_linux.go
```diff
@@ -1237,25 +1244,92 @@ func (s *Server) addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container,
...
+	if err := os.MkdirAll(imageVolumesPath, 0o700); err != nil {
+		return "", fmt.Errorf("create image volumes path: %w", err)
+	}
+
+	f, err := os.Open(imageVolumesPath)
+	if err != nil {
+		return "", fmt.Errorf("open image volumes path %s: %w", imageVolumesPath, err)
+	}
+
+	_, readErr := f.Readdirnames(1)
```
kwilczynski, line 1324: Nit: you can already use the most performant API here.

```suggestion
	_, readErr := f.ReadDir(1)
```

saschagrunert: Fixed as suggested, thank you!

Issue #8547, opened 2024-09-03
CVE-2024-24790 - Bump Go Version


FilipeNas:
### What happened?

Last release still has the CVE-2024-24790 with critical vulnerability.

CVE-2024-24790 - Base Score: [9.8 CRITICAL]

### What did you expect to happen?

No critical or high vulnerabilities in the latest release

### How can we reproduce it (as minimally and precisely as possible)?

Just scan the package

### Anything else we need to know?

Vulnerable Versions
< 1.21.11, 1.22.0 ≤ Version < 1.22.4

Fix version
1.21.11, 1.22.4

### CRI-O and Kubernetes version

Latest CRI-O Version

### OS version

n/a

### Additional environment details (AWS, VirtualBox, physical, etc.)

n/a

kwilczynski:
@FilipeNas, CRI-O does not directly use any of the affected methods, so its impact on us is not critical, and there is no immediate need for concern. Even if some security scanners would pick an older version of Go from the `go.mod` file being used.

kwilczynski:
Scanners most likely would pick on the following:

```
╭──────────────────────────────┬──────┬───────────┬─────────┬─────────┬────────╮
│ OSV URL                      │ CVSS │ ECOSYSTEM │ PACKAGE │ VERSION │ SOURCE │
├──────────────────────────────┼──────┼───────────┼─────────┼─────────┼────────┤
│ https://osv.dev/GO-2024-2598 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
│ https://osv.dev/GO-2024-2599 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
│ https://osv.dev/GO-2024-2600 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
│ https://osv.dev/GO-2024-2609 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
│ https://osv.dev/GO-2024-2610 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
│ https://osv.dev/GO-2024-2687 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
│ https://osv.dev/GO-2024-2824 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
│ https://osv.dev/GO-2024-2887 │      │ Go        │ stdlib  │ 1.22.0  │ go.mod │
[...]

PR #8548, opened 2024-09-03
Fix CVE-2024-24790

FilipeNas:
Bump go version to fix CVE.

Please note I dont know how to work with GO.
I hope this bumps the version.

Referring to this problem:
https://github.com/cri-o/cri-o/issues/8547

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
-go 1.22.0
+go 1.22.6
 
-toolchain go1.22.2
+toolchain go1.22.6
 
 module github.com/cri-o/cri-o
 
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-go 1.22.6
+go 1.22.0
 
 toolchain go1.22.6
 


go.mod
```diff
@@ -1,6 +1,6 @@
 go 1.22.0
 
-toolchain go1.22.2
+toolchain go1.22.6
```
kwilczynski, line 3: This is now how one would upgrade Go version. The above `go` directive would be what you want to change, see:

- [go.mod file reference](https://go.dev/doc/modules/gomod-ref)

kwilczynski:
@saschagrunert, do you think we could move to Go 1.23 already?

kwilczynski:
@FilipeNas, for the future reference. Please do not disregard our Pull Request template; we have it for a reason.

saschagrunert:
> @FilipeNas, for the future reference. Please do not disregard our Pull Request template; we have it for a reason.

this depends a bit if we have go 1.23 in downstream, but I would prefer moving directly to it, yes.

kwilczynski:
> > @saschagrunert, do you think we could move to Go 1.23 already?
> 
> this depends a bit if we have go 1.23 in downstream, but I would prefer moving directly to it, yes.

So, for the time being, the answer is a "no", then. Given that the CVE isn't affecting us, there would be nothing to do here.

FilipeNas:
> > > @saschagrunert, do you think we could move to Go 1.23 already?
> > 
> > 
> > this depends a bit if we have go 1.23 in downstream, but I would prefer moving directly to it, yes.
> 
> @saschagrunert so, for the time being, the answer is a "no", then.
> 
> Given that the CVE isn't affecting us, there would be nothing to do here.

Can't we upgrade to the latest patch version?
I would like for the security scanners to not pick this :))

PR #8431, opened 2024-07-30
Make container and image removal/stop idempotent

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
Fully conform to the CRI as well as the new tests from https://github.com/kubernetes-sigs/cri-tools/pull/1536
#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
Made `StopContainer`, `RemoveContainer` and `RemoveImage` idempotent per CRI API definition:
https://github.com/kubernetes/cri-api/blob/c20fa40/pkg/apis/runtime/v1/api.proto
```

PR #8549, opened 2024-09-03
default to crun (v2)

haircommander:
#### What type of PR is this?

/kind api-change
crun has been a compliant OCI runtime for many years now.
It advertises faster startup time, smaller memory footprint, and additional features
    
Given these advantages and its relative stability (and that it's the default in fedora, and podman),
update crun to be the default runtime in CRI-O as well.

```release-note
Update `crun` to be the default OCI runtime
```


.github/workflows/integration.yml
```diff
@@ -258,15 +258,16 @@ jobs:
...
+          sudo ln -sf $(command -v runc) $(command -v crun)
+          crun --version
       - name: Run the test suite
         run: sudo -E test/test_runner.sh
         env:
           JOBS: ${{ matrix.run.jobs }}
           RUN_CRITEST: ${{ matrix.run.critest }}
           RUNTIME_TYPE: ${{ matrix.run.runtimeType }}
           TEST_USERNS: ${{ matrix.run.userns }}
+          CONTAINER_DEFAULT_RUNTIME: ${{ matrix.run.defaultRuntime }}
```
saschagrunert, line 273: Do we need this?

Edit: Ah looking at the integration test changes below we probably do. :thinking:

.github/workflows/test.yml
```diff
@@ -231,6 +231,9 @@ jobs:
         run: |
           make mockgen -j $(nproc)
           hack/tree_status.sh
+      - name: spoof crun on arm64
+        if: ${{ matrix.run.arch == "arm64" }}
+        run: cp $(which runc) /usr/bin/crun
```
saschagrunert, line 236: Hm, which tests require the binary to be existent? Could we mock them as executable `os.Tempfile` to be system independent.

internal/oci/runtime_oci.go
```diff
@@ -995,6 +996,9 @@ func (r *runtimeOCI) DeleteContainer(ctx context.Context, c *Container) error {
 	}
 
 	_, err := r.runtimeCmd("delete", "--force", c.ID())
+	if errors.Is(err, ErrNotFound) {
+		return nil
```
saschagrunert, line 1000: Should we lock something here? Which use case are we skipping?

pkg/config/config.go
```diff
@@ -1581,6 +1575,10 @@ func (r *RuntimeHandler) Validate(name string) error {
 	if err := r.ValidateRuntimeAllowedAnnotations(); err != nil {
 		return err
 	}
+	if err := r.ValidateContainerMinMemory(name); err != nil {
+		logrus.Errorf("Unable to set minimum container memory for runtime handler %q: %v", name, err)
```
saschagrunert, line 1579: Should we return the error here?

.github/workflows/integration.yml
haircommander: I think it's more idiomatic. The copying of binaries and spoofing is a little weird IMO (though I still attempted to do so just in case)

.github/workflows/test.yml
haircommander: I agree, I think that could be a follow up. I could open an issue if this merges (only because it's been such a pain to get tests passing here)

internal/oci/runtime_oci.go
haircommander: I don't remember the exact situation but we hit this in a couple of integration tests. I don't think it was ultimately fatal but I think it's good to have

pkg/config/config.go
haircommander: we fallback to the default min memory so the config ends up valid, which was the original author's intent

.github/workflows/test.yml
saschagrunert: Yes, maybe more general: All unit tests should be independent from external commands.

haircommander: agreed!

saschagrunert (approved):

kwilczynski:
/hold

So I can make a review. :smile:

.github/workflows/test.yml
kwilczynski: This could have been a symbolic link, but a `cp` if fine, too. :smile:

contrib/test/ci/vars.yml
```diff
@@ -169,6 +164,11 @@ kata_skip_pod_tests:
   - 'test "systemd cgroup_parent correctly set"'
   - 'test "kubernetes pod terminationGracePeriod passthru"'
   - 'test "disable crypto.fips_enabled when FIPS_DISABLE is set"'
+  - 'test "run container with memory_limit_in_bytes -1"'
+  - 'test "run container with memory_limit_in_bytes 12.5MiB"'
+  - 'test "run container with container_min_memory 17.5MiB"'
+  - 'test "run container with container_min_memory 5.5MiB"'
+  - 'test "run container with empty container_min_memory"'
```
kwilczynski, line 171: An idea (perhaps to follow up later or not):

Do we want to include a test that uses the minimal viable memory size for crun? Just to ensure that it can actually start with that low memory limit? Now, and going forward, that is...

Thoughts?

internal/oci/runtime_oci.go
```diff
@@ -965,8 +965,9 @@ killContainer:
 		if _, err := r.runtimeCmd("kill", c.ID(), "KILL"); err != nil {
 			if !errors.Is(err, ErrNotFound) {
 				log.Errorf(ctx, "Killing container %v failed: %v", c.ID(), err)
+			} else {
+				log.Debugf(ctx, "Error while killing container %s: %v", c.ID(), err)
```
kwilczynski, line 969: @haircommander, wasn't this part of some changes that @sohankunkerkar added some time ago?

My recollection might be wrong here, but this does look familiar, a bit.

pkg/config/config.go
kwilczynski: :+1:

internal/oci/runtime_oci.go
haircommander: this is similar to https://github.com/cri-o/cri-o/pull/8431 though it's an additional case

pkg/config/config.go
```diff
@@ -1675,22 +1673,22 @@ func (r *RuntimeHandler) ValidateNoSyncLog() error {
...
+func (r *RuntimeHandler) ValidateContainerMinMemory(name string) error {
 	if r.ContainerMinMemory == "" {
 		r.ContainerMinMemory = units.BytesSize(defaultContainerMinMemory)
 	}
 
-	memoryBytes, err := units.RAMInBytes(r.ContainerMinMemory)
+	memorySize, err := units.RAMInBytes(r.ContainerMinMemory)
 	if err != nil {
-		err = fmt.Errorf("unable to set runtime memory to %q: %w", r.ContainerMinMemory, err)
+		err = fmt.Errorf("unable to set runtime memory to %q: %w. Setting to %q instead", r.ContainerMinMemory, err, defaultContainerMinMemory)
```
kwilczynski, line 1685: Nit. Just to keep errors in lower case in spirit.

```suggestion
		err = fmt.Errorf("unable to set runtime memory to %q: %w; setting to %q instead", r.ContainerMinMemory, err, defaultContainerMinMemory)
```

internal/oci/runtime_oci.go
kwilczynski: Ah, I see. You prefer to have if-else there, no worries. We actually left it to be a catch-all on purpose in the past.

PR #8553, opened 2024-09-04
Adding builder for sandbox

### LINKED ISSUE #8289: Refactor sandbox creation function arguments
This issue is created to perform the following
- Add a dedicated list of setters for the various pod sandbox fields and use them directly on sandbox creation
- Add a new method to convert a runtime spec to a sandbox and use it on restore.
- Enable the revive linter and set it to a minimum amount of arguments.

The modifications discussed in [](https://github.com/cri-o/cri-o/pull/8141) suggest a different approach. It is proposed to relocate internal/factory packages into oci and sandbox packages. Additionally, a new file named factory.go should be created in each package to define setters and translation functions. The objective is to establish a method for converting CRI specifications and server configurations into sandbox/container instances, and also to enable the restoration of these instances from a file.

A factory object, upon creation, should generate a sandbox object. Through the use of setters, its configurations are determined, and subsequently, the sandbox object is instantiated with the defined attributes. Once created, the sandbox object becomes immutable.

xw19:
#### What type of PR is this?

/kind bug

Fixes #8289 

None

```release-note

```


internal/lib/sandbox/sandbox.go
```diff
@@ -502,3 +513,84 @@ func (s *Sandbox) Ready(takeLock bool) bool {
 
 	return cState.Status == oci.ContainerStateRunning
 }
+
+func (s *Sandbox) SetResolvPath(resolvPath string) {
```
haircommander, line 517: wouldn't these setters be able to be justl ocal to `builder.go` because it's all the same package?

internal/lib/sandbox/builder/sandbox.go
```diff
@@ -0,0 +1,443 @@
+package builder
```
haircommander, line 1: ah this should be in `internal/lib/sandbox/builder.go` IMO (keep it internal in the package, that's the point of having the builder be next to the sandbox, so we don't need to expose setters on fields we won't set past the first instance

xw19: I actually tried the problem is it basically creates a circular dependency. The current factory/sandbox requires factory/container. factory/container in turns requires lib/sandbox. If we move factory/sandbox this means we are creating an circular dependency.

PR #8573, opened 2024-09-09
Fix invalid syntax in test workflow and unit tests

saschagrunert:
#### What type of PR is this?

/kind ci

Fixing the CI `test` workflow syntax.

None
None

```release-note
None
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -231,14 +231,22 @@ jobs:
         run: |
           make mockgen -j $(nproc)
           hack/tree_status.sh
-      - name: spoof crun on arm64
-        if: ${{ matrix.run.arch == "arm64" }}
-        run: cp $(which runc) /usr/bin/crun
+      - name: Download crun on arm64 # Required for checkpoint restore (criu) tests.
+        if: ${{ matrix.run.arch == 'arm64' }}
+        run: |
+          sudo curl -sSfL --retry 5 --retry-delay 3 \
+            -o $CRUN_PATH \
+            https://github.com/containers/crun/releases/download/$CRUN_VERSION/crun-$CRUN_VERSION-linux-arm64
+          sudo chmod +x $CRUN_PATH
+          crun --version
+        env:
+          CRUN_VERSION: 1.16.1
+          CRUN_PATH: /usr/bin/crun
       - name: Run unit tests as root
         if: ${{ matrix.run.type == 'root' }}
         run: |
           sudo PATH="$PATH" GOCACHE="$(go env GOCACHE)" GOMODCACHE="$(go env GOMODCACHE)" make testunit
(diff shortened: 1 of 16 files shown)


.github/workflows/test.yml
```diff
@@ -231,14 +231,22 @@ jobs:
         run: |
           make mockgen -j $(nproc)
           hack/tree_status.sh
-      - name: spoof crun on arm64
-        if: ${{ matrix.run.arch == "arm64" }}
-        run: cp $(which runc) /usr/bin/crun
+      - name: Download crun on arm64 # required for checkpoint restore tests
```
kwilczynski, line 234:
```suggestion
      - name: Download crun on arm64 # Required for checkpoint restore (criu) tests.
```

.github/workflows/test.yml
```diff
@@ -231,14 +231,22 @@ jobs:
...
+          crun --version
+        env:
+          CRUN_VERSION: 1.16.1
+          CRUN_PATH: /usr/bin/crun
       - name: Run unit tests as root
         if: ${{ matrix.run.type == 'root' }}
         run: |
           sudo PATH="$PATH" GOCACHE="$(go env GOCACHE)" GOMODCACHE="$(go env GOMODCACHE)" make testunit
-          sudo chown -R $(id -u):$(id -g) "$(go env GOCACHE)" "$(go env GOMODCACHE)"
+          sudo chown -R $(id -u):$(id -g) "$(go env GOCACHE)" "$(go env GOMODCACHE)" || true
```
kwilczynski, line 249: Can this fail? Just curious because of the `|| true`.

pkg/config/config.go
```diff
@@ -895,9 +895,9 @@ func DefaultConfig() (*Config, error) {
 		RuntimeConfig: RuntimeConfig{
 			AllowedDevices:     []string{"/dev/fuse", "/dev/net/tun"},
 			DecryptionKeysPath: "/etc/crio/keys/",
-			DefaultRuntime:     defaultRuntime,
+			DefaultRuntime:     DefaultRuntime,
 			Runtimes: Runtimes{
-				defaultRuntime: defaultRuntimeHandler(),
+				DefaultRuntime: defaultRuntimeHandler(),
 			},
```
kwilczynski, line 901: Can't be private any more? How did we miss this last time... :smile:

saschagrunert: I think having it public is not too bad, isn't it?

.github/workflows/test.yml
saschagrunert: Yes, when the modcache is not existing.

.github/workflows/test.yml
saschagrunert: Fixed as suggested :+1:

pkg/config/config.go
kwilczynski: Can be. :+1:

haircommander:

this should have no effect on that

PR #8583, opened 2024-09-10
WIP: Make unit tests independent from third party binaries

saschagrunert:
#### What type of PR is this?

/kind ci

Let the unit tests work without having a need for runtime binaries.

None
None

```release-note
None
```


test/framework/framework.go
```diff
@@ -91,6 +92,23 @@ func (t *TestFramework) MustTempFile(pattern string) string {
...
+func (t *TestFramework) EnsureRuntimeDeps() string {
+	dir := t.MustTempDir("crio-testunig-default-runtime-")
+	for dep, content := range map[string]string{
+		"crun":    "",
+		"conmon":  "#!/bin/sh\necho 'conmon version 2.1.12'",
+		"nsenter": "",
+	} {
+		runtimeDep := filepath.Join(dir, dep)
+		gomega.Expect(os.WriteFile(runtimeDep, []byte(content), 0o755)).
+			NotTo(gomega.HaveOccurred())
```
kwilczynski, line 106:
```suggestion
		gomega.Expect(os.WriteFile(runtimeDep, []byte(content), 0o755)).NotTo(gomega.HaveOccurred())
```

I am not sure where this break-on-dot comes from, but this is not Go style.

pkg/config/reload_test.go
```diff
@@ -345,7 +346,7 @@ var _ = t.Describe("Config", func() {
 	t.Describe("ReloadRuntimes", func() {
 		var existingRuntimePath string
 		BeforeEach(func() {
-			existingRuntimePath = t.MustTempFile("runc")
+			existingRuntimePath = filepath.Join(t.EnsureRuntimeDeps(), config.DefaultRuntime)
```
kwilczynski, line 349: Now the `DefaultRuntime` shines.

test/framework/framework.go
saschagrunert: It's the valid syntax to break long function chains. :shrug:

kwilczynski: Fair enough. You can tell I am not a fan of this approach. :smile: 

_For posterity:_

This works because of the implicit semicolon insertion rules. The same reason why this...

```go
Object
  .A()
  .B()
  .C()
```

would not work, but...

```go
Object.
  A().
  B().
  C()
```

works. The same way the following works:

```go
	for
	i := 0
	i < 10
	i++ {
		// Some code here.
	}
```

But we tend not do that.

PR #8590, opened 2024-09-11
Cleanups on top of #8212

mtrmac:
#### What type of PR is this?

/kind bug

Assorted cleanups identified in reviews of #8212 .

None

It might be better to split this, perhaps to discuss controversial parts separately, or because tests might be failing.

```release-note
None
```


internal/oci/runtime_vm.go
```diff
@@ -110,12 +110,12 @@ func newRuntimeVM(handler *config.RuntimeHandler, exitsPath string) RuntimeImpl
...
 
-	imageSource := c.Spec().Annotations[annotations.ImageName]
-	log.Infof(ctx, "Adding mount info to pull image %s", imageSource)
+	someNameOfTheImageSource := c.Spec().Annotations[annotations.SomeNameOfTheImage]
+	log.Infof(ctx, "Adding mount info to pull image %s", someNameOfTheImageSource)
 
 	volume := &katavolume.KataVirtualVolume{
 		VolumeType: katavolume.KataVirtualVolumeImageGuestPullType,
-		Source:     imageSource,
+		Source:     someNameOfTheImageSource,
```
mtrmac, line 118: This situation worries me, but it’s also way out of my area of expertise.

saschagrunert (approved):
Thank you!

saschagrunert:
/lgtm cancel

@harche has some review feedback as well.

server/container_create_linux.go
```diff
@@ -228,18 +228,23 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr ctrfactory.Cont
...
+	// == NEVER USE userRequestedImage (or even someNameOfTheImage) for anything but diagnostic logging past this point; it might
+	// resolve to a different image.
 
 	systemCtx, err := s.contextForNamespace(sb.Metadata().Namespace)
 	if err != nil {
 		return nil, fmt.Errorf("get context for namespace: %w", err)
 	}
 
+	// WARNING: This hard-codes an assumption that SignaturePolicyPath set specifically for the namespace is never less restrictive
+	// than the default system-wide policy, i.e. that if an image is successfully pulled, it always conforms to the system-wide policy.
```
haircommander, line 247: not for this PR: is there some way to confirm this invariant programatically? or change this check to also account for the system-wide one?

harche:
I was wondering if this PR can be broken into multiple PRs, if that would help in understanding what bug an individual commit is trying to fix.

server/container_restore.go
```diff
@@ -75,6 +75,21 @@ func (s *Server) CRImportCheckpoint(
 
 	var restoreArchivePath string
 	if restoreStorageImageID != nil {
+		sb, err := s.getPodSandboxFromRequest(ctx, sbID) // Note that we might call getPodSandboxFromRequest with a different sbID later. Is that necessary?
```
haircommander, line 78: we haven't run
```
	if sbID == "" {
		// restore into previous sandbox
		sbID = dumpSpec.Annotations[annotations.SandboxID]
		ctrID = config.ID
	} else {
		ctrID = ""
	}

```
yet here (line 147) so i don't think we know for sure what's the sandbox ID yet. that's why I opted to have it below

pkg/config/config_unsupported.go
```diff
@@ -26,3 +26,7 @@ func selinuxEnabled() bool {
 func checkKernelRROMountSupport() error {
 	return errdefs.ErrNotImplemented
 }
+
+func (c *RuntimeConfig) ValidatePinnsPath(executable string) error {
```
haircommander, line 30: should we add a gh action that compiles on mac so you don't need to keep playing catchup each time you commit to cri-o?

internal/storage/image.go
```diff
@@ -638,7 +638,12 @@ func pullImageChild() {
 		os.Exit(1)
 	}
 
-	output <- pullImageOutputItem{Result: transports.ImageName(destRef), Name: canonicalRef.Name(), Digest: canonicalRef.Digest()}
+	rawCanonical, ok := canonicalRef.Raw().(reference.Canonical)
+	if !ok {
+		fmt.Fprintf(os.Stderr, "Returned reference %v is not canonical", canonicalRef.Raw().String())
```
haircommander: can you use the `cri-o/internal/log` package here instead of logging right to Stderr?

internal/storage/image.go
```diff
@@ -638,7 +638,12 @@ func pullImageChild() {
...
+		os.Exit(1)
```
haircommander: is this a fatal programming error? like can cri-o not recover at all? if so I think we should `panic()` here. If not, we should find a way to error the image pull

internal/storage/image.go
haircommander: I see this (and my below comment) are done above, could you cleanup both cases?

internal/factory/container/container.go
```diff
@@ -215,11 +215,11 @@ func (c *container) SpecAddAnnotations(ctx context.Context, sb *sandbox.Sandbox,
 	}
 
 	c.spec.AddAnnotation(annotations.UserRequestedImage, userRequestedImage)
-	imageName := ""
+	someNameOfThisImage := ""
```
haircommander, line 218: I'm not personally a huge fan of how verbose this variable is. What is the risk to cri-o programmers if we leave the name as-is? do the comments present not document that risk?

internal/storage/image.go
```diff
@@ -589,9 +587,8 @@ type pullImageArgs struct {
 }
 
 type pullImageOutputItem struct {
-	Progress     *types.ProgressProperties `json:",omitempty"`
-	Result       string                    `json:",omitempty"` // If not "", in the format of transport.ImageName()
-	CanonicalRef string                    `json:",omitempty"` // If not "", in the format of RegistryImageReference.StringForOutOfProcessConsumptionOnly(), and always contains a digest.
+	Progress *types.ProgressProperties `json:",omitempty"`
+	Result   string                    `json:",omitempty"` // If not "", in the format of RegistryImageReference.StringForOutOfProcessConsumptionOnly(), and always contains a digest.
```
haircommander, line 591: I find myself wondering if we should keep calling this CanonicalRef so we don't have confusions between 1.30 and 1.31 in making this change (same variable name, same type, different semantic meaning)

mtrmac:
> I was wondering if this PR can be broken into multiple PRs, if that would help in understanding what bug an individual commit is trying to fix.

I intended to document that in commit messages; it’s probably easier to review by commit.

I’m fine with splitting those 14 any way you like; I assume 14 individual PRs with 14 individual `/retest` trains is not desirable.

server/container_restore.go
mtrmac: Yes, but _that_ sandbox ID comes from inside the metadata image.

For simplicity, suppose the node had a policy “deny anything from `docker.io`”; then any attempt to restore a snapshot using a metadata image from `docker.io` should fail.

I don’t _really_ know what I’m doing — I’m assuming Kubelet is always passing the same kind of metadata (at least, there’s only effectively a single `CreateContainer` call site in Kubelet), so using the same sandbox ID the non-snapshot-restore “create container” code path is using should be correct here.

But then again, I don’t know restoring a sandbox ID / namespace from contents of the metadata image follows namespace restrictions, so I might well be missing something.

pkg/config/config_unsupported.go
mtrmac: I’d _personally_ like that, but if you don’t have it already, you probably don’t need it to be effective in maintaining the software.

(There’s _much_ more in CRI-O that doesn’t compile on Mac; this was only targeted at making `internal/storage` unit tests run.)

internal/storage/image.go
mtrmac: Both of these are in `pullImageChild`, a separate process. The error message is explicitly expected to go to Stderr, and `pullImageParent` reads that (as `errOutput`) and turns it into a Go error.

From skimming `internal/log`, it seems to add Logrus’ typical log metadata; I think that should not be included in the Go error returned from `pullImageParent`.

internal/storage/image.go
mtrmac: See above, this is in a subprocess and this is the currently-designed error reporting mechanism.

internal/factory/container/container.go
mtrmac: Yes, I know you’re not a fan. I’ll take it out if you decide so.

The risk has materialized in https://github.com/cri-o/cri-o/pull/8212#discussion_r1734986296 / https://github.com/cri-o/cri-o/pull/8212#discussion_r1747337067 . (And I don’t know what https://github.com/cri-o/cri-o/pull/8590#discussion_r1752967914 deos.)

I don’t know that these ugly verbose names would _help_ but also I don’t have any better ideas.

---

Ideally, perhaps, CRI-O would change the logic to record some other values that are more known to be relevant (e.g. `PullImage` now returns a pretty well-specified repo@digest value). But that’s not this PR, and it would almost certainly change the API outputs; so I’ll leave the CRI-O maintainers to navigate that.

The ugly verbose names might serve as a gentle encouragement :)

internal/storage/image.go
mtrmac: I did wonder about this — it is used via `reexec.Command`, i.e. talking to `/proc/self/exe`; i.e. there should never be a version drift between the producer’s and consumer’s implementation.

(And, also, `types.ImageReference` and `RegistryImageReference` string formats are mutually incompatible, so the consumer would fail parsing a mismatching value either way.)

---

I don’t care about the name of the field at all; it’s purely internal between `pullImageChild`/`pullImageParent`. Let me know what you want it to be.

server/container_restore.go
mtrmac: > suppose the node had a policy…

suppose the _namespace_ invoking the operation had a policy, rather.

mtrmac:
What do I need to do to get this accepted? Split into some specific parts? Change anything?

pkg/annotations/internal.go
```diff
@@ -31,8 +31,10 @@ const (
 	// SeccompProfilePath is the node seccomp profile path.
 	SeccompProfilePath = "io.kubernetes.cri-o.SeccompProfilePath"
 
-	// Image is the container image ID annotation.
-	Image = "io.kubernetes.cri-o.Image"
+	// UserRequestedImage is ann annotation containing the image specified in the container spec
```
haircommander: nit: `ann` -> `an`

mtrmac: Thanks, fixed.

server/container_restore.go
haircommander: @adrianreber wdyt?

haircommander:
what would you think about keeping everything but https://github.com/cri-o/cri-o/pull/8590/commits/dca8939a3706d34b3c3b0bbb17be3cbdcd2981ed and https://github.com/cri-o/cri-o/pull/8590/commits/fba4f75668694bb2dd5545c4bbd3ddf8e797556f and putting those in a separate PR for discussion? i know those are some of the more important commits, but I don't wanna keep you blocked on all of these for those to be muscled through

server/container_restore.go
adrianreber: The code to restore into an existing sandbox ID is a left-over from the initial pull request which supported to restore a pod. This is no longer possible with the current code. Most of the code was removed. This should haven been also removed.

internal/oci/runtime_vm.go
mtrmac: Further discussion in https://github.com/cri-o/cri-o/pull/8706#discussion_r1813425897 .

server/container_restore.go
mtrmac: Further discussion to follow in https://github.com/cri-o/cri-o/pull/8706#discussion_r1813427411 .

internal/factory/container/container.go
mtrmac: Further discussion in https://github.com/cri-o/cri-o/pull/8706#discussion_r1813429259 , I have dropped this part from this PR.

mtrmac:
> what would you think about keeping everything but [dca8939](https://github.com/cri-o/cri-o/commit/dca8939a3706d34b3c3b0bbb17be3cbdcd2981ed) and [fba4f75](https://github.com/cri-o/cri-o/commit/fba4f75668694bb2dd5545c4bbd3ddf8e797556f) and putting those in a separate PR for discussion?

Thanks, done. #8706 currently includes the proposals from this PR, a merge commit, and the two commits.

haircommander:
/lgtm

thanks!

saschagrunert (approved):

saschagrunert:
Are there any commits within this PR which should be backported to release-1.31 because they fix a regression?

mtrmac:
I don’t think there are any regression fixes here. The net effect should be a minor efficiency improvement by removing no-longer-used operations (including removing a remote registry access, potentially improving reliability), and perhaps removing some small proven-to-be-unreachable code paths.

PR #8601, opened 2024-09-12
Refactoring factory/container to remove references of lib/sandbox

xw19:
#### What type of PR is this?

/kind cleanup

Instead of passing whole sandbox object we will pass parameters that are needed This will remove interdependencies for future refactor. Also ManagedNamespace extracted to its own package inside lib so that it can be referenced for both sandbox and container.

Part of #8289 

No

```release-note
Restructuring of packages.
```

### DIFF
--- a/internal/factory/container/sandbox.go
+++ b/internal/factory/container/sandbox.go
@@ -1,0 +1,38 @@
+package container
+
+import (
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+
+	"github.com/cri-o/cri-o/internal/lib/namespace"
+)
+
+// SandboxIFace represents an interface for interacting with sandbox related information.
+// This is to ensure we don't import the lib sandbox and just use its methods.
+type SandboxIFace interface {
+	// LogDir returns the location of the logging directory for the sandbox.
+	LogDir() string
+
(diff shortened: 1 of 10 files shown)


xw19:
This PR allows to move factory/sandbox to move over to lib/snadbox

internal/factory/container/sandbox.go
```diff
@@ -0,0 +1,19 @@
+package container
+
+import (
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+
+	"github.com/cri-o/cri-o/internal/lib/namespace"
+)
+
+type SandboxIFace interface {
```
haircommander, line 9: can you describe what this is an interface for  and the motivation (especially because there are a handful of Sandbox objects littered throughought the project)

xw19: I have add information please check.

haircommander: I am still missing the motivation. can you describe in the comment why we added this interface instead of directly using Sandbox?

haircommander:
/approve

LGTM, thanks @xw19 , @cri-o/cri-o-maintainers PTAL

Issue #8603, opened 2024-09-13
[V1.31.0] Error: checking signature : Docker references with both a tag and digest are currently not supported


Nemric:
### What happened?

Hi,
I did the upgrade from K8S & Cri-O from v1.30.4 to v1.31.0
I don't know if this issue is really related to Cri-O but I can't install ingress-nginx because of this behavior taken from `kubectl get events` when applying https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.11.2/deploy/static/provider/baremetal/deploy.yaml

```
pod/ingress-nginx-admission-create-s6x86         Error: checking signature of "registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.4.3@sha256:a320a50cc91bd15fd2d6fa6de58bd98c1bd64b9a6f926ce23a600d87043455a3": creating docker:// reference for "registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.4.3@sha256:a320a50cc91bd15fd2d6fa6de58bd98c1bd64b9a6f926ce23a600d87043455a3": Docker references with both a tag and digest are currently not supported
```

Crio logs looks like :
`Sep 13 09:10:35 Marie crio[2048]: time="2024-09-13 09:10:35.088965171Z" level=info msg="Image status: &ImageStatusResponse{Image:&Image{Id:ce263a8653f9cdabdabaf36ae064b3e52b5240e6fac90663ad3b8f3a9bcef242,RepoTags:[],RepoDigests:[registry.k8s.io/ingress-nginx/kube-webhook-certgen@sha256:1b792367d0e1350ee869b15f851d9e4de17db10f33fadaef628db3e6457aa012 registry.k8s.io/ingress-nginx/kube-webhook-certgen@sha256:a320a50cc91bd15fd2d6fa6de58bd98c1bd64b9a6f926ce23a600d87043455a3],Size_:55821706,Uid:&Int64Value{Value:65532,},Username:,Spec:&ImageSpec{Image:,Annotations:map[string]string{},UserSpecifiedImage:,RuntimeHandler:,},Pinned:false,},Info:map[string]string{},}" id=e2ce05ae-4b61-4ab6-b7f1-8b940feb277c name=/runtime.v1.ImageService/ImageStatus`

and kubelet logs looks like :
[...]

PR #8605, opened 2024-09-13
image: serialize RegistryImageReferences when checking signatures

haircommander:
#### What type of PR is this?

/kind bug
So we can check a signature of an image that has both a tag and a digest
fixes https://github.com/cri-o/cri-o/issues/8603

```release-note
Fix a bug where signature checking failed if an image specified both a tag and a digest
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -966,21 +966,9 @@ func (svc *imageService) CandidatesForPotentiallyShortImageName(systemContext *t
 
 	images := make([]RegistryImageReference, len(resolved.PullCandidates))
 	for i := range resolved.PullCandidates {
-		// Strip the tag from ambiguous image references that have a
-		// digest as well (e.g.  `image:tag@sha256:123...`).  Such
-		// image references are supported by docker but, due to their
-		// ambiguity, explicitly not by containers/image.
-		ref := resolved.PullCandidates[i].Value
-		_, isTagged := ref.(reference.NamedTagged)
-		canonical, isDigested := ref.(reference.Canonical)
-		if isTagged && isDigested {
-			canonical, err = reference.WithDigest(reference.TrimNamed(ref), canonical.Digest())
-			if err != nil {
-				return nil, err
-			}
-			ref = canonical
-		}
-		images[i] = references.RegistryImageReferenceFromRaw(ref)
+		// This function will strip the tag if both tag and digest are specified, as it's supported
+		// by Docker (and thus CRI-O by example) but not c/image.
+		images[i] = references.RegistryImageReferenceFromRaw(resolved.PullCandidates[i].Value)
 	}
 
 	return images, nil
--- a/internal/storage/references/registry_reference.go
+++ b/internal/storage/references/registry_reference.go
@@ -12,7 +12,7 @@ import (
 //
 // More specifically:
 // - The name always specifies a registry; it is not an alias nor a short name input to a search
-// - The name contains a tag or digest; it does not specify just a repo.
+// - The name contains a tag xor digest; it does not specify just a repo.
 //
 // This is intended to be a value type; if a value exists, it contains a valid reference.
 type RegistryImageReference struct {
@@ -24,8 +24,22 @@ type RegistryImageReference struct {
 // RegistryImageReferenceFromRaw is an internal constructor of a RegistryImageReference.
 //
 // This should only be called from internal/storage.
-// It’s the caller’s responsibility to provide a valid value (!IsNameOnly, and registry-qualified).
+// It will modify the reference if both digest and tag are specified, stripping the tag and leaving the digest.
+// It will also verifies the image is not only a name. If it is only a name, the function errors.
 func RegistryImageReferenceFromRaw(rawNamed reference.Named) RegistryImageReference {
+	_, isTagged := rawNamed.(reference.NamedTagged)
+	canonical, isDigested := rawNamed.(reference.Canonical)
+	// Strip the tag from ambiguous image references that have a
+	// digest as well (e.g.  `image:tag@sha256:123...`).  Such
+	// image references are supported by docker but, due to their
+	// ambiguity, explicitly not by containers/image.
+	if isTagged && isDigested {
+		canonical, err := reference.WithDigest(reference.TrimNamed(rawNamed), canonical.Digest())
+		if err != nil {
+			panic("internal error, reference.WithDigest was not passed a digest, which should not be possible")
+		}
+		rawNamed = canonical
+	}
 	// Ideally this would be better encapsulated, e.g. in internal/storage/internal, but
 	// that would require using a type defined with the internal package with a public alias,
 	// and as of 2023-10 mockgen creates code that refers to the internal target of the alias,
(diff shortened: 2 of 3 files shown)


internal/storage/references/registry_reference.go
```diff
@@ -36,6 +36,24 @@ func RegistryImageReferenceFromRaw(rawNamed reference.Named) RegistryImageRefere
 	return RegistryImageReference{privateNamed: rawNamed}
 }
 
+// RegistryImageReferenceFromRawSerialized is an internal constructor of a RegistryImageReference, that also
+// strips the tag from ambiguous image references that have a digest as well (e.g.  `image:tag@sha256:123...`).
+// Such image references are supported by docker but, due to their ambiguity, explicitly not by containers/image.
+//
+// This should only be called from internal/storage.
+func RegistryImageReferenceFromRawSerialized(rawNamed reference.Named) (RegistryImageReference, error) {
```
haircommander, line 44: not crazy about this name, @mtrmac do you have any suggestions?

mtrmac: See the other comment for a bit higher-level view…

… which still leaves open the naming of this canonicalization. My starting point is `canonicalize{TagDigest,DockerLike}` somewhere in the name, but that might get awfully long.

test/policy.bats
```diff
@@ -204,3 +204,38 @@ SANDBOX_CONFIG="$TESTDATA/sandbox_config.json"
 
 	[[ "$output" == *"SignatureValidationFailed"* ]]
 }
+
+@test "allow signed image with restrictive policy on container creation7 if already pulled (by tag and ID)" {
```
mtrmac, line 208: This is fundamentally confusing things. “Image ID” (`StorageImageID`) is, historically (nowadays not always) the sha256 digest of the config, and usually expressed without the leading `sha256:`.

“image digest” is the sha256 of the manifest. The two are ~never the same for a single image.

In `busybox:latest@sha256:…`, the second value is the manifest digest. So assuming the value from `crictl images -q` is an image ID, using it in the manifest digest syntax should never find a match; and the test name is misleading.

(Also for the other added test.)

mtrmac (commented):
ACK to the overall functionality this achieves.

---

WRT the implementation … my question is “what rules are implied by the `RegistryImageReference` data type”? I think it should be one of:
- :tag@digest references are unambiguously allowed by the data type.
  - In that case the mapping of :tag@digest to @digest-only must be in consumers of `RegistryImageReference`, i.e. in `PullImage`/`IsRunningImageAllowed` (and what value should be used in `MetricImagePullsSuccessesInc`, with or without the tag?)
  - or maybe `RegistryImageReference` should have a `RawCanonicalized` method that returns a value suitable for c/image, and most callers of `Raw` should migrate
- :tag@digest references are unambiguously invalid in the data type. In that case
  - _the constructor of `RegistryImageReferenceFromRaw` should reject that value to enforce the rule_
  - It would be consistent for `ParseRegistryImageReferenceFromOutOfProcessData` to reject the data as well
  - We’d probably need some kind of `…FromRawCanonicalizing` and `Parse…CanonicalizingFromOutOfProcesss…` helpers, so that top-level callers don’t manipulate values directly.

The current PR is ambiguous about the permissibility of such values — the text parser canonicalizes them, but `…FromRaw` does not.

I weakly prefer the variant where :tag@digest values are invalid in `RegistryImageReference`, forcing its users to explicitly choose, by the constructor they call, whether they want to accept such values or not. If nothing else, the `parseImageParent`/`parseImageChild` RPC should never generate such values, and it can use the stricter parser/constructor.

haircommander:
I agree the library should take care of canonicalizing. In fact, I just replaced *FromRaw with this behavior. Also i believe I updated the test to test digest and not storage ID. PTAL again @mtrmac

internal/storage/references/registry_reference.go
```diff
@@ -24,16 +24,27 @@ type RegistryImageReference struct {
...
+// It does two validations on rawNamed:
+// - It validates the image either has a tag or a digest, but not both (preferring digest)
+// - Verifies the image is not only a name.
+func RegistryImageReferenceFromRaw(rawNamed reference.Named) (RegistryImageReference, error) {
+	_, isTagged := rawNamed.(reference.NamedTagged)
+	canonical, isDigested := rawNamed.(reference.Canonical)
+	if isTagged && isDigested {
+		canonical, err := reference.WithDigest(reference.TrimNamed(rawNamed), canonical.Digest())
+		if err != nil {
+			return RegistryImageReference{}, err
```
mtrmac, line 36: (The only way this can fail is if `canonical.Digest()` is invalid, and that shouldn’t have happened by induction, so I’d be fine with a `panic` here. But returning an error explicitly is perfectly valid as well.)

internal/storage/references/registry_reference.go
```diff
@@ -48,7 +59,7 @@ func ParseRegistryImageReferenceFromOutOfProcessData(input string) (RegistryImag
 		return RegistryImageReference{}, err
 	}
 	ref = reference.TagNameOnly(ref)
-	return RegistryImageReferenceFromRaw(ref), nil
+	return RegistryImageReferenceFromRaw(ref)
```
mtrmac, line 62: The documentation of `ParseRegistryImageReferenceFromOutOfProcessData` should document that it removes the tag from tag@digest inputs.

internal/storage/image.go
```diff
@@ -966,21 +972,10 @@ func (svc *imageService) CandidatesForPotentiallyShortImageName(systemContext *t
 
 	images := make([]RegistryImageReference, len(resolved.PullCandidates))
 	for i := range resolved.PullCandidates {
-		// Strip the tag from ambiguous image references that have a
-		// digest as well (e.g.  `image:tag@sha256:123...`).  Such
-		// image references are supported by docker but, due to their
-		// ambiguity, explicitly not by containers/image.
```
mtrmac, line 972: (Absolutely non-blocking: Consider preserving some form of this comment, to have a documentation for the reason we modify the inputs.)

mtrmac (commented):
I’d prefer two variants of constructors, one only in the “XOR” form, one canonicalizing — but this works, up to you.

Assuming this design, code LGTM, just some documentation suggestions.

internal/storage/references/registry_reference.go
haircommander: that's better because then I don't need to change the function signature

internal/storage/references/registry_reference.go
```diff
@@ -40,6 +54,8 @@ func RegistryImageReferenceFromRaw(rawNamed reference.Named) RegistryImageRefere
 //
 // It is only intended for communication with OUT-OF-PROCESS APIs,
 // like registry references provided by CRI by Kubelet.
+// It will modify the reference if both digest and tag are specified, stripping the tag and leaving the digest.
+// It will also verifies the image is not only a name. If it is only a name, the function errors.
```
mtrmac, line 58: The function does not fail in such a situation: The `TagNameOnly` call turns name-only references into _name_`:latest`

internal/storage/references/registry_reference.go
mtrmac: Also using `panic` does not leave around (supposedly) 100% unreachable error handling code paths, considering that CRI-O is measuring code coverage.

mtrmac (commented):
Do you also want to fix the linter-highlighted typo?

LGTM either way.

haircommander:
whoops, fixed!

mtrmac (commented):
/lgtm

PR #8610, opened 2024-09-17
Moving container sanbox to lib sandbox

xw19:
#### What type of PR is this?

/kind cleanup

This is cleanup to move the factory sandbox to libsandbox. This will create a sandbox object which will be immutable.

Part of #8289

None

```release-note
- Move factory/sandbox to lib/sandbox

```


internal/lib/sandbox/builder.go
```diff
@@ -0,0 +1,379 @@
...
+	Spec() *generate.Generator
+
+	// ResolvPath returns the sandbox's resolvPath.
+	ResolvPath() string
+
+	// SetDNSConfig sets the dns configs.
+	SetDNSConfig(*types.DNSConfig)
+
+	// SetCriSandbox sets and creates CriSandbox.
+	SetCriSandbox(string, time.Time, map[string]string, map[string]string, *types.PodSandboxMetadata)
```
haircommander, line 53: nit: can this be `SetCRISandbox`

internal/factory/container/container_test.go
```diff
@@ -98,10 +99,30 @@ var _ = t.Describe("Container", func() {
...
-			sb, err := sandbox.New("sandboxID", "", "", "", "test",
-				make(map[string]string), make(map[string]string), "", "",
-				&types.PodSandboxMetadata{}, "", "", false, "", "", "",
-				[]*hostport.PortMapping{}, false, currentTime, "", nil, nil)
+			sbuilder := sandbox.NewBuilder()
+
+			sbuilder.SetID("sandboxID")
+			sbuilder.SetName("")
+			sbuilder.SetNamespace("")
+			sbuilder.SetKubeName("")
```
haircommander, line 107: I don't think we need to call these if we're just passing ""

internal/factory/container/container_test.go
```diff
@@ -98,10 +99,30 @@ var _ = t.Describe("Container", func() {
...
+			sbuilder.SetLogDir("test")
+			sbuilder.SetCriSandbox(sbuilder.ID(), currentTime, make(map[string]string), make(map[string]string), &types.PodSandboxMetadata{})
+			sbuilder.SetShmPath("")
+			sbuilder.SetCgroupParent("")
+			sbuilder.SetPrivileged(false)
```
haircommander, line 112: similar here, only call if we're setting non default

internal/lib/sandbox/builder.go
haircommander: also I feel like we could do this step at sandbox generation step, and fail if the respective fields aren't defined. We pass all of these things in through setters beforehand right? what if a caller messes up and sets them differently? I think we should inherit as much as possible to reduce code

internal/factory/container/container_test.go
haircommander: these notes apply to all the _test.go functions, but not for the pieces that run in production

server/sandbox_run_linux.go
```diff
@@ -375,30 +376,30 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
+	if _, err := s.ReservePodName(sandboxID, sandboxName); err != nil {
+		reservedID, getErr := s.PodIDForName(sandboxName)
 		if getErr != nil {
-			return nil, fmt.Errorf("failed to get ID of pod with reserved name (%s), after failing to reserve name with %w: %w", sbox.Name(), getErr, getErr)
+			return nil, fmt.Errorf("failed to get ID of pod with reserved name (%s), after failing to reserve name with %w: %w", sandboxName, getErr, getErr)
 		}
 		// if we're able to find the sandbox, and it's created, this is actually a duplicate request
 		// Just return that sandbox
-		if reservedSbox := s.GetSandbox(reservedID); reservedSbox != nil && reservedSbox.Created() {
+		if reservedsbuilder := s.GetSandbox(reservedID); reservedsbuilder != nil && reservedsbuilder.Created() {
```
haircommander, line 386: this shouldn't be a builder this should be a sandbox, why the name change?

haircommander: it would reduce code churn too to start which would make this easier to review as well

server/sandbox_run_linux.go
```diff
@@ -698,43 +699,84 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
+	sbuilder.SetPortMappings(portMappings)
+	sbuilder.SetHostNetwork(hostNetwork)
+	sbuilder.SetUsernsMode(usernsMode)
+	sbuilder.SetPodLinuxOverhead(overhead)
+	sbuilder.SetPodLinuxResources(resources)
+	sbuilder.SetDNSConfig(sbuilder.Config().DnsConfig)
+	sbuilder.SetHostnamePath(hostnamePath)
+	sbuilder.SetNamespaceOptions(securityContext.NamespaceOptions)
+	sbuilder.SetSeccompProfilePath(seccompRef)
+	sbuilder.SetCreatedAt(created)
```
haircommander, line 760: I think these steps should happen incrementally throughout, though we can leave them like this for now if it's helpful to make sure we're calling everything (before we have validation on the sbuilder object before it spits out a sandbox)

server/sandbox_run_linux.go
```diff
@@ -1026,21 +1045,21 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
 	// Since it's not a context error, we can delete the resource from the store, it will be tracked in the server from now on.
-	s.resourceStore.Delete(sbox.Name())
+	s.resourceStore.Delete(sandboxName)
 
 	sb.SetCreated()
 	s.generateCRIEvent(ctx, sb.InfraContainer(), types.ContainerEventType_CONTAINER_STARTED_EVENT)
 
 	log.Infof(ctx, "Ran pod sandbox %s with infra container: %s", container.ID(), container.Description())
-	resp = &types.RunPodSandboxResponse{PodSandboxId: sbox.ID()}
+	resp = &types.RunPodSandboxResponse{PodSandboxId: sandboxID}
```
haircommander, line 1062: I think for this PR I'd prefer to keep this sbox.ID() -> sandboxID (and Name similarly) out of it so we can focus on the structural changes if you wouldn't mind

internal/factory/container/container_test.go
xw19: Resolved

server/sandbox_run_linux.go
xw19: We delete the sandbox hence if we use sbox.ID() it tends to cause problems. Same for sbox.Name()

internal/lib/sandbox/builder.go
```diff
@@ -0,0 +1,379 @@
...
+	libconfig "github.com/cri-o/cri-o/pkg/config"
+)
+
+// Builder is the interface for managing pod sandboxes.
+type Builder interface {
+	// SetConfig sets the sandbox configuration and validates it
+	SetConfig(*types.PodSandboxConfig) error
+
+	// SetNameAndID sets the sandbox name and ID
+	SetNameAndID() error
```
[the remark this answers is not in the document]
xw19: Replace with GenerateNameAndID

internal/lib/sandbox/builder.go
xw19: So now we will use the same createdAt to set the time here and hence removed it from params but now it does require to set created at before hand else it will throw error

server/sandbox_run_linux.go
```diff
@@ -375,53 +383,57 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
...
 	}
-	resourceCleaner.Add(ctx, "runSandbox: releasing pod sandbox name: "+sbox.Name(), func() error {
-		s.ReleasePodName(sbox.Name())
+	resourceCleaner.Add(ctx, "runSandbox: releasing pod sandbox name: "+sboxName, func() error {
+		s.ReleasePodName(sboxName)
 		return nil
 	})
 
-	s.resourceStore.SetStageForResource(ctx, sbox.Name(), "sandbox creating")
+	// TODO: Pass interface instead of individual field.
```
haircommander, line 407: what's your plan for TODO's? address as follow ups? if so, should we open an issue to track them?

haircommander:
/retest
/approve

one question on TODOs, but this is a good step. In the future i'd also love to have a validation function so we know when we spit out a sandbox at the end of the building process we've set all of the fields we expect to have

PTAL @cri-o/cri-o-maintainers

server/sandbox_run_linux.go
sohankunkerkar: @xw19 Any inputs on this?
Could you also squash the commits into a single one and prepare you PR for merging?

xw19: @haircommander @sohankunkerkar I have created two issues and we will work on them
https://github.com/cri-o/cri-o/issues/8823
https://github.com/cri-o/cri-o/issues/8822

xw19: @sohankunkerkar Squashed into a single commit.

internal/lib/sandbox/builder.go
```diff
@@ -0,0 +1,393 @@
...
+
+	// SetCreatedAt sets the created at time.
+	SetCreatedAt(createdAt time.Time)
+}
+
+// builder is the hidden default type behind the Sandbox interface.
+type builder struct {
+	config *types.PodSandboxConfig
+	infra  container.Container
+	sb     *Sandbox
```
sohankunkerkar, line 132: nit: Can we rename this to `sandboxRef` or `sandbox`?

internal/lib/sandbox/builder.go
```diff
@@ -0,0 +1,393 @@
...
+	// SetPodLinuxResources sets the PodLinuxResources.
+	SetPodLinuxResources(*types.LinuxContainerResources)
+
+	// SetHostnamePath sets the hostname path.
+	SetHostnamePath(string)
+
+	// SetNamespaceOptions sets the namespace options.
+	SetNamespaceOptions(*types.NamespaceOption)
+
+	// SetSeccompProfilePath sets the selinux comp profile path.
```
sohankunkerkar, line 118: `SetSeccompProfilePath sets the seccomp profile path.`

internal/lib/sandbox/builder.go
```diff
@@ -0,0 +1,393 @@
...
+	// Must be called after InitInfraContainer.
+	Spec() *generate.Generator
+
+	// ResolvPath returns the sandbox's resolvPath.
+	ResolvPath() string
+
+	// SetDNSConfig sets the dns configs.
+	SetDNSConfig(*types.DNSConfig)
+
+	// SetCRISandbox sets and creates CriSandbox.
```
sohankunkerkar, line 52: `s/CriSandbox./CRI sandbox with the given parameters./g`

internal/lib/sandbox/builder.go
```diff
@@ -0,0 +1,393 @@
...
+
+// ID returns the id of the pod sandbox.
+func (b *builder) ID() string {
+	if b.sb == nil || b.sb.criSandbox == nil {
+		return ""
+	}
+	return b.sb.criSandbox.Id
+}
+
+// Name returns the id of the pod sandbox.
```
sohankunkerkar, line 236: `Name returns the name of the pod sandbox.`

internal/lib/sandbox/builder.go
```diff
@@ -0,0 +1,393 @@
...
+	// SetContainers sets the containers.
+	SetContainers(memorystore.Storer[*oci.Container])
+
+	// SetProcessLabel sets the processLabel.
+	SetProcessLabel(string)
+
+	// SetMountLabel sets the mountLabel.
+	SetMountLabel(string)
+
+	// SetShmPath sets the shim path.
```
sohankunkerkar, line 79: `SetShmPath sets the shared memory path`

internal/runtimehandlerhooks/high_performance_hooks_test.go
```diff
@@ -717,14 +717,19 @@ var _ = Describe("high_performance_hooks", func() {
...
+		err = sbox.SetCRISandbox(sbox.ID(), make(map[string]string), map[string]string{
+			crioannotations.CPUSharedAnnotation + "/" + c.CRIContainer().GetMetadata().GetName(): annotationEnable,
+		}, &types.PodSandboxMetadata{})
+		Expect(err).ToNot(HaveOccurred())
+		sbox.SetPrivileged(false)
+		sbox.SetHostNetwork(false)
+		sbox.SetCreatedAt(createdAt)
+		sb, err := sbox.GetSandbox()
 		Expect(err).ToNot(HaveOccurred())
+
```
sohankunkerkar, line 732: stray line.

sohankunkerkar (commented):
Few nits but overall LGTM

internal/runtimehandlerhooks/high_performance_hooks_test.go
xw19: Can we keep this it adds to better readability.

internal/lib/sandbox/builder.go
xw19: Changed to sandboxRef

internal/runtimehandlerhooks/high_performance_hooks_test.go
sohankunkerkar: Sure!

sohankunkerkar (approved):
/lgtm

sohankunkerkar:
@xw19 Thanks for your patience!

sohankunkerkar:
@cri-o/cri-o-maintainers I think we may need to override `e2e-gcp-ovn`.

PR #8617, opened 2024-09-18
Don't rely on vendored tools

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now use `go install` to get go-md2man, ginkgo as well as the uber maintained version of gomock.

None

None

```release-note
None
```


saschagrunert:
@cri-o/cri-o-maintainers PTAL

haircommander:
/lgtm
nice cleanup!

hack/go-install.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -uo pipefail
+
+DEST_DIR=${1:-}
+CMD=${2:-}
+GO_INSTALL_LOC=${3:-}
+
+if [[ ("$DEST_DIR" == "" || "$CMD" == "" || "$GO_INSTALL_LOC" == "") ]]; then
```
kwilczynski, line 8: Curious, as I am not sure, do the brackets make any difference here?

hack/go-install.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "Usage: $0 DEST_DIR CMD GO_INSTALL_LOC"
+    exit 1
+fi
+
+set +e
+CMD_PATH=$(command -v "$CMD")
+set -e
```
kwilczynski, line 15: What if we make it more explicit, for example:

```suggestion
CMD_PATH=$(command -v "$CMD" || true)
```

hack/go-install.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ "$CMD_PATH" == "" ]; then
```
kwilczynski, line 17: For `[`, perhaps `-n` or `-z` would be better? For `[[` (the built-in), `==` is probably fine.

hack/go-install.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -uo pipefail
```
kwilczynski, line 2:
```suggestion
#!/usr/bin/env bash

set -uo pipefail
```

hack/go-install.sh
saschagrunert: Yep adjusted as suggested.

hack/go-install.sh
saschagrunert: I think we can remove them.

hack/go-install.sh
saschagrunert: Using `-z` now :+1:

hack/go-install.sh
saschagrunert: Fixed

saschagrunert:


@cri-o/cri-o-maintainers PTAL

PR #8399, opened 2024-07-18
ci: fix reload config for empty pinned image test

littlejawa:
#### What type of PR is this?

/kind failing-test

#### What this PR does / why we need it:
Modify the "wait_for_log" function in helpers.bash to allow it to catch multiple occurrences of the same message.
Use that in the failing test to synchronize and wait for config reload to be finished before checking the content of the configuration.

#### Which issue(s) this PR fixes:
Fixes #8324

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #7887, opened 2024-03-14
Add dynamic image pull timeout

saschagrunert:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
We now support a pull context which can actually be cancelled. This way we can combine the pull progress together with a dynamic timeout, which cancels the context if no progress is being made on pull.

#### Which issue(s) this PR fixes:

Refers to https://github.com/cri-o/cri-o/issues/7885
#### Special notes for your reviewer:
cc @mrunalp 
#### Does this PR introduce a user-facing change?

```release-note
Added dynamic image pull timeout if the network availability goes down during the operation.
```

PR #8619, opened 2024-09-18
[WIP] [release-1.30] Enable composefs support in CRI-O

kwilczynski:
#### What type of PR is this?

/kind dependency-change
/kind feature
/kind other

> [!WARNING]  
> This is an experiment. Please keep this in mind when offering reviews.

TBA

Related:

- [containers/composefs](https://github.com/containers/composefs): The reliability of disk images, the flexibility of files
- https://github.com/openshift/machine-config-operator/pull/4631

None

None

```release-note
None
```


kwilczynski:
@littlejawa, I cherry-picked https://github.com/cri-o/cri-o/pull/8399 here, but it seems to still have issues.

Anything I am doing wrong?

littlejawa:
@kwilczynski - the first "wait_for_log" is failing to find the "Configuration reload completed" message. And according to the logs, it is effectively not there.

I just pulled your PR locally, and I don't see the line in the code that writes this log :
https://github.com/cri-o/cri-o/blob/64b00ec2ecfec69d8dcce2f053eac4194cb08f4b/server/server.go#L585

It's been introduced with this commit: https://github.com/cri-o/cri-o/commit/c3bfcd3b4cbe8b364a9c69c149e966d61cb732b6

Apparently it's not on the 1.30 branch? From what I can see, the test was backported without the code that writes the log - it works only because of the surrounding code that looks for all the other log messages.

Anyway, you'll need to add that if you want the new "wait_for_log()" to work.

kwilczynski:
[...]
> I just pulled your PR locally, and I don't see the line in the code that writes this log :
> 
> https://github.com/cri-o/cri-o/blob/64b00ec2ecfec69d8dcce2f053eac4194cb08f4b/server/server.go#L585
> 
> It's been introduced with this commit: [c3bfcd3](https://github.com/cri-o/cri-o/commit/c3bfcd3b4cbe8b364a9c69c149e966d61cb732b6)
> 
> Apparently it's not on the 1.30 branch? From what I can see, the test was backported without the code that writes the log - it works only because of the surrounding code that looks for all the other log messages.
[...]

@littlejawa, oh nice find!

Bummer. I did not look that far. Thank you for looking! For some reason, I assumed that the test fix would fix the test that has always been failing on the 1.30 release branch. I didn't think that the code was not there... 

... now we need to check what is up with this, as either we are missing something or, as you say, the test shouldn't be here.

kwilczynski:
> [...]
> 
> > I just pulled your PR locally, and I don't see the line in the code that writes this log :
> > https://github.com/cri-o/cri-o/blob/64b00ec2ecfec69d8dcce2f053eac4194cb08f4b/server/server.go#L585
> > 
> > It's been introduced with this commit: [c3bfcd3](https://github.com/cri-o/cri-o/commit/c3bfcd3b4cbe8b364a9c69c149e966d61cb732b6)
> > Apparently it's not on the 1.30 branch? From what I can see, the test was backported without the code that writes the log - it works only because of the surrounding code that looks for all the other log messages.
> > [...]
> 
> @littlejawa, oh nice find!
> 
> Bummer. I did not look that far. Thank you for looking! For some reason, I assumed that the test fix would fix the test that has always been failing on the 1.30 release branch. I didn't think that the code was not there...
> 
> ... now we need to check what is up with this, as either we are missing something or, as you say, the test shouldn't be here.

@littlejawa, you are a star!

The following added the changes to the image pinning code, plus added the test:

- https://github.com/cri-o/cri-o/pull/8323

Whereas the following added the log line, and hence fixed the issue the other change brought but was never backported:

- https://github.com/cri-o/cri-o/pull/8124

server/image_pull.go
```diff
@@ -273,7 +273,7 @@ func (s *Server) pullImageCandidate(ctx context.Context, sourceCtx *imageTypes.S
 func consumeImagePullProgress(ctx context.Context, cancel context.CancelFunc, progress <-chan imageTypes.ProgressProperties, remoteCandidateName storage.RegistryImageReference) {
 	// The progress interval is 1s, but we give it a bit more time just in case
 	// that the connection revives.
-	const timeout = 10 * time.Second
+	const timeout = 60 * time.Second
```
mtrmac, line 276: Hum, I didn’t realize this exists.

This is currently a mismatch with the c/image implementation; it only reports progress on “layer copies”, but a pull is not just layer copies.

In particular, in the relevant case of pulling to c/storage, the part that reports progress is just a network download, storing to a local file (and decompressing to obtain a digest); and afterwards a “commit” phase happens which actually extracts the local files into individual files for the overlay filesystem; _and the commit phase makes no progress updates at all_.

Now, the code _tries_ to interleave commit operations and starts doing them as soon as the layers (in order) finish downloading, but that’s not guaranteed to happen in general, and even if it _did_ happen perfectly, if the last layer is large, that is going to take a long time without any progress updates.

I’m not saying that it’s _good_ that it works this way; just that it works that way now, and imposing a timeout on operations which don’t report progress will inevitably cause pull errors on large enough images with slow enough storage.

kwilczynski: Would there be a better way of doing this?

Related:

- https://github.com/cri-o/cri-o/pull/7887

@saschagrunert was aiming to address an upstream Kubernetes issue/feature here.

kwilczynski: Also, this is the gist of why I increased this timeout to 60 seconds to get things working with composefs:

_(to quote from @mtrmac's comment above)_

> (...) I’m not saying that it’s good that it works this way; just that it works that way now, and imposing a timeout on operations which don’t report progress will inevitably cause pull errors on large enough images with slow enough storage.

mtrmac: Right now c/image just doesn’t have a “watchdog keep-alive” feature.

kwilczynski: As an interim solution, I will turn this into a configuration option, so that people can adjust this and set as needed, should it be causing issues.

kwilczynski:
Closing for now. Will re-open in the future, if needed.

PR #8677, opened 2024-10-15
[release-1.28] OCPBUGS-42912: Cherry-pick changes from containers/storage/pull#2134

kwilczynski:
#### What type of PR is this?

/kind bug
/assign kwilczynski

Manually cherry-pick changes from https://github.com/containers/storage/pull#2134 as these changes contain a fix that needs to be backported to CRI-O release 1.28, part of OpenShift 4.15 release.

Related:

- https://github.com/containers/storage/pull/2134

None

None

```release-note
None
```


vendor/github.com/containers/storage/userns.go
```diff
@@ -309,3 +331,19 @@ func getAutoUserNSIDMappings(
...
+	}
+
+	flags := unix.O_PATH | unix.O_CLOEXEC | unix.O_RDONLY
+	fileHandle, err := os.OpenFile(filePath, flags, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	return fileHandle, nil
+}
```
kwilczynski, line 349: @mheon, would this be a sufficient replacement for the newer API?

mheon: Yes, that LGTM.

kwilczynski:
@cri-o/cri-o-maintainers, please have a look. Thank you!

vendor/github.com/containers/storage/userns.go
cyphar: This is not sufficient if there is a risk of a racing process changing the filesystem from underneath you. You should instead do something like (untested):

```go
func somewhatSafeOpen(root, unsafePath string, mode int) (_ *os.File, Err error) {
	if mode&(os.O_CREAT|os.O_TRUNC) != 0 {
		return nil, fmt.Errorf("O_CREAT|O_TRUNC cannot be used safely")
	}
	path, err := securejoin.SecureJoin(root, unsafePath)
	if err != nil {
		return nil, err
	}
	handle, err := os.OpenFile(path, mode)
	if err != nil {
		return nil, err
	}
	defer func() {
		if Err != nil {
			_ = handle.Close()
		}
	}()
	// At this point, we aren't sure that the handle we grabbed is the same
	// path we expected. So double-check it using procfs.
	actualPath, err := os.Readlink(fmt.Sprintf("/proc/self/fd/%d", handle.Fd()))
	if err != nil {
		return nil, fmt.Errorf("failed to check handle path for %q: %w", path, err)
	}
	if path != actualPath {
		return nil, fmt.Errorf("container breakout detected: tried to open %q but instead opened %q", path, actualPath)
	}
	return handle, nil
}
```

kwilczynski: @cyphar what about the version that uses the newer API from `securejoin` package.  Anything to improve there?

kwilczynski: @mheon @Luap99 @giuseppe  please have a look when you have a moment.

If we agree this is the way to go, then I will open a PR to incorporate the improvements into the relevant branches under the containers/storage.

Luap99: My understanding from looking at the storage code is that we create a temporary layer that is mounted where we lookup /etc/passed and /etc/group and no container process should ever be running on that fs.
So I don't think we have to worry about race conditions from malicious processes.

Of course a host process could try to trick us but we cannot protect us against a process that can write to the storage anyway so this is not a concern here I think.

cyphar: @kwilczynski The new `OpenInRoot` API in v0.3.x of filepath-securejoin is safe against races where an attacker can modify the underlying filesystem. There are some minor improvements to these protections in `libpathrs` but all of the core attacks are protected against in filepath-securejoin.

If it is indeed the case that an attacker cannot modify the filesystem (as @Luap99 says), then that's fine. But I would advise some caution in dismissing the issue without thoroughly checking -- in runc we've had plenty of attacks where a new container being set up could still have an attacker modify the rootfs (the most common case is a Kubernetes pod with a shared volume in say `/etc` so the other running container can modify the filesystem while we are still operating on the rootfs).

kwilczynski: @cyphar @Luap99 thank you both!

Luap99: @cyphar We are using the new API on main it was just not backported (I guess because a go version constraint? Or trying to keep the size small?)

The code I am referring to is https://github.com/containers/storage/blob/664fe5d9b95004e1be3eee004d56a1715c8ca790/userns.go#L198-L235
so I think it is impossible that we end up using that temporary layer anywhere for a container.

giuseppe: the assumption in https://github.com/containers/storage/blob/664fe5d9b95004e1be3eee004d56a1715c8ca790/userns.go#L336 is that the underlying files can't change since there are no external mounts at this point and there is only one consumer for the directory, so if anyone gets direct access to the storage directory we already have big problems.  If you are constrained by the go version available, I agree with @cyphar that adding a check using proc should be enough for now (we already have a version of the check in https://github.com/containers/storage/blob/main/pkg/chunked/filesystem_linux.go#L314-L324).

kwilczynski: @Luap99 @giuseppe you both have different view on whether this is an issue or not.

However, if we reach a consensus, then I will be happy to add the procfs check to harden the older implementation.

cyphar: @kwilczynski Let me try to lower the Go requirement and add a CI runner to check that it works with older Go versions. Sorry, I should've been more careful about bumping the version...

Unfortunately I'm not familiar enough with `containers/storage` to be able to say whether or not the way the layer is mounted is okay -- if the mount is from a read-only source (i.e. nothing on the entire system can write to it) then it would be fine, but if you're using a directory that can be written to by someone else as a lowerdir then I'm not so sure.

giuseppe: > @Luap99 @giuseppe you both have different view on whether this is an issue or not.
> 
> However, if we reach a consensus, then I will be happy to add the procfs check to harden the older implementation.

it seems we agree?  What @Luap99 said is correct, there is no other process accessing the same directory so race conditions can't happen, and adding a check using procfs would catch them anyway.  Of course a privileged process can do that, but a privileged process can already write to the host anyway so there is nothing to protect from.

Of course for us it would be better to use securejoin as we are already using that version in the upstream c/storage version, the only disadvantage is the additional burden on @cyphar since we could just workaround this problem for now.

cyphar: https://github.com/cyphar/filepath-securejoin/pull/37 is a PoC of downgrading the minimum Go version to 1.18 (the lowest we can practically go). I'm still not entirely sure if I'm happy about how many stdlib implementations we have to copy, though it's not _that_ bad...

cyphar: https://github.com/cyphar/filepath-securejoin/releases/tag/v0.3.6 contains the compatibility code.

PR #8684, opened 2024-10-16
Port hostport code from iptables to nftables

danwinship:
#### What type of PR is this?
/kind cleanup
/kind other

Ports hostport code from iptables to nftables, which will be needed when RHEL 10 drops support for iptables.

None (but it's part of https://issues.redhat.com/browse/OCPSTRAT-940)

I'm not sure exactly what sort of compatibility constraints cri-o has but note that:
1. As currently written, this PR entirely drops iptables support, requiring all users to use nftables. This is not a problem for OCP (all RHEL9-based releases have working nftables) but it is a new/changed requirement overall that other users may need to consider on upgrade.
2. `nft` releases before 1.0.1 are problematic (see discussion [here](https://github.com/kubernetes/kubernetes/issues/122743#issuecomment-1889932301)) so the knftables package requires that version or newer. This means using a mid-2021-or-later distro, meaning not RHEL 7, Debian buster or bullseye, or Ubuntu 20.04 (which are all _allegedly_ currently supported according to `install.md`). 
3. As currently written, this PR does not try to do any cleanup/migration of existing iptables hostport rules, meaning that if you upgraded cri-o on a running node (ie, without draining/rebooting), you could end up with "stranded" iptables hostport rules that would not get cleaned up when their corresponding pods were deleted.

Also, the arm64 integration tests are currently failing because the image used for them doesn't include `nft`. I tried updating `.github/workflows/integration.yml` to do that, but the PR seems to be ignoring it. Maybe it has to be updated separately? (Or I did something wrong...)

```release-note
Hostports are now implemented using nftables rather than iptables.
Action required: cri-o now depends on the `nft` binary being available (and
does not depend on the iptables binaries being available).
```


danwinship:
> 1. As currently written, this PR entirely drops iptables support

(meant to imply: I _could_ rewrite it to support both iptables and nftables (though that would obviously be more complicated). It just didn't occur to me until I was filing the PR that we might want that.)

kwilczynski:
@danwinship, we are making an assumption here that all users of CRI-O are going to be using nftables, isn't it?

danwinship:
@kwilczynski it assumes that all users of CRI-O will be on distros with new-enough nftables (see the "Special notes for your reviewer" section of the initial comment). It doesn't assume they'll be using nftables for kube-proxy or firewalling or anything else. The hostport rules are independent.

(And I can rewrite it to support either iptables or nftables if necessary. It will just be more complicated...)

danwinship:
So there's only one test of HostPorts in the e2e suite (`[It] [sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]`), and it's passing in both `ci/prow/e2e-aws-ovn` and `ci/prow/ci-e2e` (which, if I understand correctly, runs against a vanilla upstream k8s environment, not OCP, right?)

haircommander:
> if I understand correctly, runs against a vanilla upstream k8s environment, not OCP, right

correct!

Can we optimistically use nftables and if that fails fallback to iptables? I'm fine dropping iptables in our docs and recommending nftables but I Think there should be some transition time so we don't pull the rug out from underneath users who use older OSs (I know of some using debian, not sure which version)

danwinship:
@haircommander done, though probably this needs some hand-testing to make sure it really is supporting both modes (and the linter doesn't like what I did with the unit tests...)

Does it need to do both-iptables-and-nftables-cleanup on Remove? (eg, does it need to support the possibility of upgrading and restarting cri-o without rebooting).

haircommander:
> Does it need to do both-iptables-and-nftables-cleanup on Remove? (eg, does it need to support the possibility of upgrading and restarting cri-o without rebooting).

Unfortunately, yes

haircommander:
@danwinship do we need this for 1.32 or is it possible to defer to 1.33?

danwinship:
It can be deferred to 1.33

saschagrunert:
Moved it to the next milestone, thank you!

haircommander:
also: we're approaching 1.33 deadline, is this needed for 1.33?

danwinship:
It would be best to have it for OCP 4.20. We should really really have it by 4.22.

internal/hostport/hostport_nftables.go
```diff
@@ -0,0 +1,308 @@
...
+	// twice. We can't _exactly_ match only the traffic that was definitely DNATted by
+	// our rule as opposed to someone else's, but if the traffic has been DNATted and
+	// has src=dst=podIP then _someone_ needs to masquerade it.
+	tx.Add(&knftables.Set{
+		Name: hairpinSet,
+		Type: knftables.Concat(
+			ipaddr, ".", ipaddr,
+		),
+		Comment: knftables.PtrTo("hostport hairpin connections"),
+	})
```
aojea, line 296: I never remember, but IIRC sets has different semantics and is easy to wipe them out when just trying to add a single element, can't say this is correct or wrong, just you should double check adding  a new  hostport does not delete all the previous set entries

internal/hostport/hostport_nftables.go
```diff
@@ -0,0 +1,308 @@
...
+	// Note that this rule runs after any "dnat" rules in "hostports", so "ip daddr"
+	// here is the DNATted destination address (the pod IP), not the original packet's
+	// destination address.
+	tx.Add(&knftables.Rule{
+		Chain: "masquerading",
+		Rule: knftables.Concat(
+			"ct", "status", "&", "dnat|snat", "==", "dnat",
```
aojea, line 303: this is really cool

internal/hostport/hostport_nftables.go
```diff
@@ -0,0 +1,308 @@
...
+				Value: []string{
+					podIP, containerPort,
+				},
+				Comment: &comment,
+			})
+		}
+	}
+
+	// If we added any hostport mappings, add a hairpinning mapping.
+	if tx.NumOperations() != 0 {
```
aojea, line 104: doesn't this also take into account the operations in `ensureHostPortsTable(tx, hm.family)` ?

is not that you always need it if  len(hostportMappings) > 0?

internal/hostport/hostport_nftables.go
```diff
@@ -0,0 +1,308 @@
...
+	"sigs.k8s.io/knftables"
+)
+
+const (
+	// our nftables table.
+	hostPortsTable string = "crio-hostports"
+
+	// maps and sets referred to from HostportManager.
+	hostPortsMap   string = "hostports"
+	hostIPPortsMap string = "hostipports"
```
aojea, line 37: just FYI I use the all zeros address and only use one map in kindnet https://github.com/aojea/kindnet/blob/main/cmd/cni-kindnet/portmap.go

```
     map hostport-map-v4 {
                        type ipv4_addr . inet_proto . inet_service : ipv4_addr . inet_service
                        flags interval
                }

                map hostport-map-v6 {
                        type ipv6_addr . inet_proto . inet_service : ipv6_addr . inet_service
                        flags interval
                        elements = { ::/0 . tcp . 18090 : 2001:db8::7051:163c:51d2:3aa0 . 8080 }
                }

                chain prerouting {
                        type nat hook prerouting priority dstnat; policy accept;
                        fib daddr type local dnat ip to ip daddr . ip protocol . th dport map @hostport-map-v4
                        fib daddr type local dnat ip6 to ip6 daddr . meta l4proto . th dport map @hostport-map-v6
                }

                chain output {
                        type nat hook output priority dstnat; policy accept;
                        fib daddr type local dnat ip to ip daddr . ip protocol . th dport map @hostport-map-v4
                        fib daddr type local dnat ip6 to ip6 daddr . meta l4proto . th dport map @hostport-map-v6
                }
```

internal/hostport/hostport_nftables.go
```diff
@@ -0,0 +1,308 @@
...
+
+	// Create the "hostports" chain with the map lookup rules, and then create
+	// "prerouting" and "output" chains that call the "hostports" chain for
+	// locally-destined packets.
+	tx.Add(&knftables.Chain{
+		Name: "hostports",
+	})
+	tx.Flush(&knftables.Chain{
+		Name: "hostports",
+	})
```
aojea, line 229: that is more a genuine question, is there any value on creating chains and jumping to them in nftables? 
I feel more comfortable adding all rules to the same chain , specially when tracing for debugging

internal/hostport/hostport_nftables.go
aojea: you need to set the interval flag to the map though

aojea:
> @aojea @npinaeva could one of you look over the nftables rules ([this commit](https://github.com/cri-o/cri-o/pull/8684/commits/267dfe39da8a7d9038382810b2bf22710f2cf94f))?

technically LGTM, module the question about the Sets https://github.com/cri-o/cri-o/pull/8684#discussion_r2071664026 that they always gave me surprises since the semantics are different to the tables, and sometimes you think they are deleted and they are not or viceversa .... other comments are more philosophical 😄

internal/hostport/hostport_nftables.go
danwinship: "Add" never erases anything. It does one of:
1. create the object if it doesn't exist.
2. no-op if the object already exists and matches what you tried to create.
3. error if the object already exists and is different from what you tried to create.

As noted in a comment above, for the chains I do Add+Flush, but for the sets just Add.

internal/hostport/hostport_nftables.go
danwinship: hm... you're right, this is wrong.

I think in an earlier stage of the refactoring, `hostportMappings` still had both IP families in it, so `len(hostportMappings) > 0` wouldn't work, but you're right, at this point that's what I need to be checking.

danwinship: (actually, we don't even need to check, because the MetaHostportManager wouldn't have called us if there were no mappings)

internal/hostport/hostport_nftables.go
danwinship: ah... yeah, that would work. I'll leave it the way I have it though since I tested with that...

internal/hostport/hostport_nftables.go
danwinship: We need to run the same rules from `output` and `prerouting`, and I like having only a single copy of them.

haircommander:
/approve
/retest
content wise I'm no expert but the code is well structured and the commits clear in intent. I'm trusting you and @aojea here :) 

@cri-o/cri-o-maintainers PTAL

haircommander:
@danwinship can you rebase please?

danwinship:
done

saschagrunert (approved):

PR #3300, opened 2020-02-18
Add live configuration reload to seccomp profile

saschagrunert:
The seccomp profile is now processed in a dedicated sub-package of the
`config` package and supports the live config reload feature. Necessary
test have been added as well.

PR #8686, opened 2024-10-16
Re-allow building without seccomp installed

### LINKED ISSUE #218: `libseccomp` is required for compilation
If don't have `libseccomp` package locally, will have a compile error.
Maybe need a update to README to add hint of install `libseccomp` package since we have supported.

michalsieron:
#### What type of PR is this?

/kind other

I think #3300 regressed #218 by unconditionally compiling with seccomp enabled. #6488 fixed this for non-linux and non-cgo build platforms, but the issue still exists when one uses a cgo-enabled Linux build host but still wants to build without seccomp. In such case build ends quickly with pkg-config error: `Package 'libseccomp' not found`.

This patch makes `internal/config/seccomp` respect the seccomp buildtag.

Fixes #218 regression

```release-note
Fixed building CRI-O without libseccomp.
```

None?

### DIFF
--- a/internal/config/seccomp/notifier.go
+++ b/internal/config/seccomp/notifier.go
@@ -1,5 +1,5 @@
-//go:build linux && cgo
-// +build linux,cgo
+//go:build seccomp && linux && cgo
+// +build seccomp,linux,cgo
 
 package seccomp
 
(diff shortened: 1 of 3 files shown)


michalsieron:
@kwilczynski I think I fixed lint issues, but I don't really understand the test failures :confused:

PR #8696, opened 2024-10-21
Refactor memory stores to use generics

saschagrunert:
#### What type of PR is this?

/kind cleanup

This allows to remove one of the two double implementations and provide a generic memory store.

None
None

```release-note
None
```


saschagrunert:
@cri-o/cri-o-maintainers PTAL

kwilczynski:
Perhaps we should move to `sync.Map` here at the same time.

saschagrunert:
> Perhaps we should move to `sync.Map` here at the same time.

Good point, doing that.

PR #8697, opened 2024-10-22
crio status: add `goroutines` subcommand

saschagrunert:
#### What type of PR is this?

/kind feature

The subcommand can be used to query the CRI-O goroutines in the same way as writing them to disk using `SIGUSR1`. This also allows to query the goroutines using the HTTP endpoint:

```shell
sudo curl --unix-socket /var/run/crio/crio.sock http://crio/debug/goroutines
```

None
None

```release-note
Added crio status `goroutines` subcommand and `/debug/goroutines` HTTP endpoint.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,13 +251,14 @@ The following API entry points are currently supported:
 
 <!-- markdownlint-disable MD013 -->
 
-| Path              | Content-Type       | Description                                                                        |
-| ----------------- | ------------------ | ---------------------------------------------------------------------------------- |
-| `/info`           | `application/json` | General information about the runtime, like `storage_driver` and `storage_root`.   |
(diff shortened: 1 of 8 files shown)


kwilczynski:
We can use this to also solve the "heap dump" decable we've been having for a while now. See the following:

- https://github.com/cri-o/cri-o/issues/7307

If we move this to a "debug" namespace, such as we have:

- /debug/heap
- /debug/goroutines

Then `crio status` could serve both nicely. Same for a curl against the endpoint.

That said, goroutines and heap dumps are available when the profiling support is enabled, why add more?

saschagrunert:
> That said, goroutines and heap dumps are available when the profiling support is enabled, why add more?

Do you suggest to not add the feature at all or should we move it just to the `/debug/…` endpoint?

saschagrunert:
Updated the endpoint and added an integration test.

PR #8702, opened 2024-10-23
Add `crio status heap` and `/debug/heap` endpoint

### LINKED ISSUE #7307: Add an endpoint for heap dumps

saschagrunert:
#### What type of PR is this?

/kind feature

The subcommand can be used to create an CRI-O heap dump. This also allows to query the same the HTTP endpoint:

sudo curl --unix-socket /var/run/crio/crio.sock http://crio/debug/heap

The dump will be placed in a temp file while the output points the user towards that.

Fixes #7307 
None

```release-note

```


server/inspect.go
```diff
@@ -254,6 +257,22 @@ func (s *Server) GetExtendInterfaceMux(enableProfile bool) *chi.Mux {
...
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		debug.WriteHeapDump(f.Fd())
+
+		if _, err := fmt.Fprintf(w, "Wrote heap dump to: %s\n", f.Name()); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	}))
```
kwilczynski, line 274: Now that this is more user-facing, so to speak, having a slightly different approach could be more prudent.

A future CLI could use one of the following approaches:

- `crio status heap` - without any argument, it will write the heap dump to local/current directory rather than the `/tmp` (so not quite like what sending the signal currently does)
- `crio status heap` - without argument simply writes to the standard output (but this might be too messy)
- `crio status heap -f <FILE NAME>` - the `-f` or `--file` (or perhaps even `--output`) will be a required argument denoting where the heap dump should be written
- `crio status heap -f -` will write the heap dump to the standard output (since this will be done on request, the user has to handle the output)

server/inspect.go
```diff
@@ -254,6 +257,29 @@ func (s *Server) GetExtendInterfaceMux(enableProfile bool) *chi.Mux {
...
+		content, err := os.ReadFile(f.Name())
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		if _, err := w.Write(content); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
```
saschagrunert, line 280: That's a bit unfortunate and It would be cool if we could write the heap dump directly to the `http.Responsewriter` :thinking:

server/inspect.go
kwilczynski: You mean "heap.out". :smile: 

Looking at...

https://github.com/cri-o/cri-o/blob/7a9778cadc77302b17de7d637b28d1c8089313a3/cmd/crio/main.go#L43-L52

If we extract the date bit into a small helper (to reuse the compressed timestamp), it could then help us to render a file name automatically following the pattern we already have for the SIGUSR1 handling. So, we pick one:

- crio-heap-%s.out
- crio-heap-dump-%s.out
- crio-heap-%s.dump

Add the compressed timestamp, and et voila.

saschagrunert: Okay, changed as suggested :+1:

internal/criocli/status.go
```diff
@@ -159,3 +175,38 @@ func goroutines(c *cli.Context) error {
...
+	}
+
+	data, err := crioClient.HeapInfo(c.Context)
+	if err != nil {
+		return err
+	}
+
+	outputPath := c.String("file")
+	if outputPath == "" {
+		outputPath = "heap.out"
```
kwilczynski, line 192:
```suggestion
		outputPath = fmt.Sprintf("crio-heap-%s.out", Timestamp())
```

internal/criocli/status.go
```diff
@@ -53,6 +55,20 @@ var StatusCommand = &cli.Command{
...
+	}, {
+		Action:  heap,
+		Aliases: []string{"hp"},
+		Name:    "heap",
+		Usage:   "Write the heap dump to a temp file and print its location on disk.",
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:      "file",
+				Aliases:   []string{"f"},
+				Value:     fmt.Sprintf("crio-heap-%s.out", Timestamp()),
```
kwilczynski, line 67:
```suggestion
```

internal/criocli/status.go
```diff
@@ -159,3 +175,38 @@ func goroutines(c *cli.Context) error {
...
+	outputPath := c.String("file")
+	if outputPath == "" {
+		outputPath = "heap.out"
+	}
+
+	var file *os.File
+
+	if outputPath == "-" {
+		file = os.Stdout
+	} else {
```
kwilczynski, line 199: I wonder if this `outputPath` setting could be a select? We repeat some of the if-else statements checks.

internal/criocli/status.go
```diff
@@ -159,3 +175,38 @@ func goroutines(c *cli.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("create output file %s: %w", outputPath, err)
+		}
+	}
+
+	if _, err := file.Write(data); err != nil {
+		return fmt.Errorf("write heap dump: %w", err)
+	}
+
+	logrus.Infof("Wrote heap dump to: %s", outputPath)
```
kwilczynski, line 210: Here, when we have "-" as the desired output destination, we need to make sure we don't print anything else, otherwise the heap dump format might be potentially invalid due to the extra output.

saschagrunert: Sounds good, changed as suggested.

internal/criocli/status.go
saschagrunert: Yes, cleaned that up.

kwilczynski:
Very nice!

PR #8707, opened 2024-10-23
pass cliContext instead of creating a new one

tariq1890:


saschagrunert (approved):
The CLI context should be similar to the background, so I guess we're fine here.

Thank you for the PR!

Issue #8666, opened 2024-10-10
Policy Signature Verification Fails


MaryJulian17:
### What happened?

Using a policy define like this, when crio starts a container, the signature verification fails with:
```
Oct 09 15:39:26 HOST crio[2094]: time="2024-10-09 15:39:26.904125667Z" level=debug msg="Response: &RunPodSandboxResponse{PodSandboxId:327f5ecea3068153344a14a569c254875c05e64daf45769fced1503f9a99066f,}" file="otel-collector/interceptors.go:75" id=a403abc5-f746-4ace-a453-9d6244b2329c name=/runtime.v1.RuntimeService/RunPodSandbox
Oct 09 15:39:26 HOST crio[2094]: time="2024-10-09 15:39:26.905114611Z" level=debug msg="Response error: checking signature of \"registry.k8s.io/kube-controller-manager:v1.31.1\": verifying signatures: Invalid crypto engine" file="otel-collector/interceptors.go:72" id=86731c11-5ad3-442b-bca6-cb2ab551cfa5 name=/runtime.v1.RuntimeService/CreateContainer
Oct 09 15:39:26 HOST crio[2094]: time="2024-10-09 15:39:26.907250611Z" level=debug msg="Request: &PodSandboxStatusRequest{PodSandboxId:327f5ecea3068153344a14a569c254875c05e64daf45769fced1503f9a99066f,Verbose:false,}" file="otel-collector/interceptors.go:63" id=0d2927fa-4a43-4607-b164-9576e13e21f6 name=/runtime.v1.RuntimeService/PodSandboxStatus
[...]

PR #8708, opened 2024-10-24
Gpgme fixed path

saschagrunert:
#### What type of PR is this?

/kind bug

Allow to search local paths for gpgme by disabling the `--enable-fixed-path` configure flag.

Fixes https://github.com/cri-o/cri-o/issues/8666
/hold for verification

```release-note
Fixed gpgme/gnupg search path in static build binaries.
```


nix/overlay.nix
```diff
@@ -3,7 +3,11 @@ let
 in
 self: super:
 {
-  gpgme = (static super.gpgme);
+  gpgme = (static super.gpgme).overrideAttrs (x: {
+    # Drop the --enable-fixed-path:
+    # https://github.com/nixos/nixpkgs/blob/9a79bc99/pkgs/development/libraries/gpgme/default.nix#L94
+    configureFlags = self.lib.lists.drop 1 x.configureFlags;
```
kwilczynski, line 9: This looks fine, but a bit brittle. Can the order of the included flag change? Maybe.

saschagrunert: Yeah it can, we probably have to make it a bit more robust if the manual test succeeds.

saschagrunert: Now it should be better :)

kwilczynski: Now, that looks solid!

PR #8731, opened 2024-11-01
expose Pod assigned IPs to NRI plugins

aojea:
/kind feature

Follow up on https://github.com/containerd/nri/pull/119

Expose the assigned IPs to the Pod sandbox to the NRI plugins

No

```release-note
nri plugins can obtain access to the assigned Pod IPs on the PodSandbox hooks
```


internal/nri/sandbox.go
```diff
@@ -39,6 +40,7 @@ func commonPodSandboxToNRI(pod PodSandbox) *nri.PodSandbox {
 		Annotations:    pod.GetAnnotations(),
 		RuntimeHandler: pod.GetRuntimeHandler(),
 		Pid:            pod.GetPid(),
+		Ips:            pod.IPs(),
```
aojea, line 43: interestingly , crio already had a method to obtain this information

aojea: let me know if you want to duplicate the method to be `GetIPs` for consistency with the other methods in the interface

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/containerd/containerd v1.7.21
 	github.com/containerd/containerd/api v1.7.19
 	github.com/containerd/fifo v1.1.0
-	github.com/containerd/nri v0.7.0
+	github.com/containerd/nri v0.7.1-0.20241031201336-6d486acd3f71
```
saschagrunert, line 17: Do we have a release to pin to or can we cut one?

saschagrunert (commented):
Thanks! Just a nit about the release.

go.mod
aojea: Samuel says is going to cut one https://github.com/containerd/containerd/pull/10921#issuecomment-2451072751 , I was just sending the PR to check the CI and get the functionality on both runtimes at the same time

aojea: We can wait for the tag, I just leave this open in the meantime if you don't mind

saschagrunert: Sounds great, thank you!

aojea:
@saschagrunert updated the PR to use the latest tag

haircommander:
/approve

LGTM, thanks!

internal/nri/sandbox.go
kwilczynski: @aojea, would this be a lot of work? If not, then GetIPs() would be nice, indeed.

aojea: It is not work at all, that was my preference in the beginning, but it duplicates an existing methods so I didn't know what what the preference of the maintainers

aojea: done, I just wrapped the existin IPs() method instead of returning the field directly to ensure there can no be any drift in the future

saschagrunert (approved):
Thanks!

PR #8738, opened 2024-11-03
Update default cni network config in readme

networkandcode:
Is this line correct?
```
By default, we install the dual stack version: 10-crio-bridge.conflist
```

To reproduce on new ubunty system(I used mutlipass on mac):
```
sudo -i

# Ubuntu 24.04.1 LTS
lsb_release -a
# should be empty
ls /etc | grep -ie 'cni\|crio'

CRIO_VERSION=v1.31

curl -fsSL https://pkgs.k8s.io/addons:/cri-o:/stable:/$CRIO_VERSION/deb/Release.key |
    gpg --dearmor -o /etc/apt/keyrings/cri-o-apt-keyring.gpg

echo "deb [signed-by=/etc/apt/keyrings/cri-o-apt-keyring.gpg] https://pkgs.k8s.io/addons:/cri-o:/stable:/$CRIO_VERSION/deb/ /" |
    tee /etc/apt/sources.list.d/cri-o.list

apt-get update -y -qq
apt-get install -y -qq cri-o

systemctl start crio.service

# should show cni and crio
ls /etc | grep -ie 'cni\|crio'
```

What I see is 11-crio-ipv4-bridge.conflist by default:
```
$ ls /etc/cni/net.d/
11-crio-ipv4-bridge.conflist
```

```release-note

```

### DIFF
--- a/contrib/cni/README.md
+++ b/contrib/cni/README.md
@@ -28,7 +28,7 @@ Copy the respective files to the configuration directory like so:
 sudo cp 10-crio-bridge.conflist /etc/cni/net.d
 ```
 
-By default, we install the dual stack version: [10-crio-bridge.conflist][dual-stack]
+By default, we install the IPv4 only version: [11-crio-ipv4-bridge.conflist][ipv4-only]
 
 However, if you are installing on a node with ipv6 disabled
 (`sysctl net.ipv6.conf.default.disable_ipv6` and


contrib/cni/README.md
```diff
@@ -28,7 +28,7 @@ Copy the respective files to the configuration directory like so:
 sudo cp 10-crio-bridge.conflist /etc/cni/net.d
 ```
 
-By default, we install the dual stack version: [10-crio-bridge.conflist][dual-stack]
+By default, we install the IPv4 only version: [11-crio-ipv4-bridge.conflist][ipv4-only]
```
saschagrunert, line 31: @cri-o/cri-o-maintainers would it be better to install the dual stack config per default in the packages?

haircommander: yeah I think installing the dual stack one in the packaging would be better

saschagrunert: > yeah I think installing the dual stack one in the packaging would be better

Then let's close this PR.

saschagrunert:
Thank you for the PR @networkandcode! :pray: We decided to change the default CNI configuration to use the dual stack, so I assume we can close this PR.

/close

PR #8754, opened 2024-11-12
RuntimeHandler inheritance

### LINKED ISSUE #8753: Allow CRIO runtime handler to inherit default runtime binary and path from MCO configuration
Currently, when creating a CRIO runtime handler in the [crio.runtime.runtimes.${HANDLER_NAME}] section, users must specify the runtime_path for the desired runtime binary (e.g., `runc` or `crun`). If `runtime_path` is not provided, CRIO attempts to retrieve the binary location from $PATH, but it does not inherit the configured default runtime binary and path as set by the Machine Config Operator (MCO).

To enhance flexibility and alignment with MCO configurations, we propose allowing CRIO to inherit both the binary name and path of the default runtime when runtime_path is omitted. Alternatively, a new option such as` inherit = default` could be introduced, explicitly instructing CRIO to use the default runtime configuration.

MarSik:
#### What type of PR is this?

/kind feature

This allows custom RuntimeHandlers to always inherit the default configured runtime paths. This helps with adding extra annotations or other "minor" changes while following what rest of the system does.

Fixes #8753

This is a minimal proof of concept so I could test whether the configuration flow would make sense. I actually do not know if it covers all code paths.

```release-note
A runtime handler definition in the configuration file can use a new option `use_default_runtime`. Setting it to true causes the values for runtime path, runtime type and runtime root to be inherited from the currently configured default runtime.
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -264,6 +264,10 @@ type RuntimeHandler struct {
 	// Output of the "features" subcommand.
 	// This is populated dynamically and not read from config.
 	features runtimeHandlerFeatures
+
+	// Inheritance request
+	// Fill in the Runtime information (paths and type) from the default runtime
+	InheritDefaultRuntime bool `toml:"inherit_default_runtime,omitempty"`
 }
 
 // Multiple runtime Handlers in a map.
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -1231,6 +1231,7 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
 # runtime_path = "/path/to/the/executable"
 # runtime_type = "oci"
 # runtime_root = "/path/to/the/root"
+# inherit_default_runtime = false
 # monitor_path = "/path/to/container/monitor"
 # monitor_cgroup = "/cgroup/path"
 # monitor_exec_cgroup = "/cgroup/path"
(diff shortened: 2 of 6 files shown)


saschagrunert (changes requested):
We need to update `docs/crio.conf.5.md` and `pkg/config/template.go`. Please also fill the `release-notes` block in the body of this PR. Is it still WIP? :)

MarSik:
@saschagrunert Like this?

pkg/config/config.go
```diff
@@ -264,6 +264,10 @@ type RuntimeHandler struct {
 	// Output of the "features" subcommand.
 	// This is populated dynamically and not read from config.
 	features runtimeHandlerFeatures
+
+	// Inheritance request
+	// Fill in the Runtime information (paths and type) from the default runtime
+	UseDefaultRuntime bool `toml:"use_default_runtime,omitempty"`
```
saschagrunert, line 270: I'm just a bit concerned about the naming, how about `inherit_default_options` or similar? Maybe other @cri-o/cri-o-maintainers have other thoughts on that.

saschagrunert (commented):
Just a nit about the naming, otherwise LGTM

pkg/config/template.go
```diff
@@ -1321,6 +1321,7 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
 {{ $.Comment }}runtime_path = "{{ $runtime_handler.RuntimePath }}"
 {{ $.Comment }}runtime_type = "{{ $runtime_handler.RuntimeType }}"
 {{ $.Comment }}runtime_root = "{{ $runtime_handler.RuntimeRoot }}"
+{{ $.Comment }}use_default_runtime = {{ $runtime_handler.UseDefaultRuntime }}
```
saschagrunert, line 1324: Can we add a section above to explain the feature, like which values are inherited from the default runtime?

MarSik: Done.

pkg/config/config.go
MarSik: Is inherit_default_runtime OK? Because it only replaces the runtime_* variables.

saschagrunert (approved):

PR #8755, opened 2024-11-12
Remove validation for TLSCA

bitoku:
#### What type of PR is this?

/kind bug

https://github.com/cri-o/cri-o/pull/8744 introduced new validations for TLS options of the stream server, but TLSCA is not a required option and it's allowed to be empty. (sorry I should've noticed it)

https://github.com/cri-o/cri-o/blob/0470ab6ed5d1e0f08565b34feac10bfefc9b8225/server/server.go#L140-L153

None 

```release-note
None
```


saschagrunert (approved):

saschagrunert:
Makes sense, thank you for the fix!

PR #8757, opened 2024-11-13
Improve hostport unit tests

danwinship:
#### What type of PR is this?
/kind cleanup

Another spinoff of #8684. This fixes up the hostport unit tests in preparation for adding nftables. (In particular, in my first attempt at #8684, the linter complained about the fact that I had just copied the test cases from the iptables unit tests to the nftables unit tests, so this splits them out so we can share them directly.)

#8684 currently has my complete branch if you want to see how the nftables stuff builds on top of this, but I thought it would be easier to review if I split this part out separately...

None

```release-note
None
```

### DIFF
--- a/internal/hostport/fake_iptables.go
+++ b/internal/hostport/fake_iptables.go
@@ -20,11 +20,11 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"net"
 	"strings"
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
/ok-to-test

Thank you!

PR #8768, opened 2024-11-19
RuntimeHandler inheritance bug-fix

MarSik:
#### What type of PR is this?

/kind bug

The original inheritance PR missed few important bits

- the reload code path was not the only one that loads configs
- an empty RuntimePath has a special meaning and inheriting it does not inherit the parent runtime
- golang does a copy during interation, not a reference and the updated values were forgotten
- the unit test ignores some of the checks during CI when not on a bare metal

This patch fixes all of these and adds better logging and also tests that actually catch the above mistakes.

Fixes https://github.com/cri-o/cri-o/issues/8753

I apologize for missing the issue during the first submission, but it all looked like working. It was just an accident though..

This fix was manually tested including the negative cases both by running unit tests with the code commented out (which I did before too...) and by manually editing crio configs on a real bare metal node.

```release-note
None
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1307,6 +1307,30 @@ func defaultRuntimeHandler() *RuntimeHandler {
 func (c *RuntimeConfig) ValidateRuntimes() error {
 	var failedValidation []string
 
+	// Update the default runtime paths in all runtimes that are asking for inheritance
+	for name := range c.Runtimes {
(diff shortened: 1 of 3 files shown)


saschagrunert (approved):

saschagrunert:
/lgtm cancel 

please fix the lint CI

PR #8791, opened 2024-11-28
Add systemd watchdog support

saschagrunert:
#### What type of PR is this?

/kind feature

Adding support for the systemd watchdog. Right now it verifies that the CRI socket is reachable and the runtime reports ready status.

None
None

```release-note
Adding support for the systemd watchdog. For now it verifies that the CRI socket is reachable and the runtime reports ready status.
```


saschagrunert:
Not exactly sure why the CI fails here.

cmd/crio/daemon_linux.go
```diff
@@ -1,12 +1,14 @@
...
-	systemdDaemon "github.com/coreos/go-systemd/v22/daemon"
+	"github.com/coreos/go-systemd/v22/daemon"
 	"github.com/sirupsen/logrus"
+
+	"github.com/cri-o/cri-o/internal/watchdog"
 )
 
 func sdNotify() {
-	if _, err := systemdDaemon.SdNotify(true, "READY=1"); err != nil {
+	if _, err := watchdog.NewSystemd().Notify(daemon.SdNotifyReady); err != nil {
```
saschagrunert, line 11: This change may break something.

server/container_start.go
```diff
@@ -13,12 +13,17 @@ import (
...
+	"github.com/cri-o/cri-o/internal/watchdog"
 )
 
 // StartContainer starts the container.
 func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRequest) (res *types.StartContainerResponse, retErr error) {
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
+
+	watchdog.DefaultSystemd().UnsetNotifySocket(ctx)
+	defer watchdog.DefaultSystemd().RestoreNotifySocket(ctx)
```
saschagrunert, line 25: @cri-o/cri-o-maintainers we may need to find a better way to handle this.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

saschagrunert:
> What kinds of failures is this intended on addressing? a deadlock is one, but I'm not sure we'd get stuck when reporting status

Right now we just check everything around the CRI-O socket: being able to connect, transfer and analyze the data. CRI-O now assumes that it doesn't work correctly if the socket or the transferred data is misbehaving. I would say we could also think about more use cases to follow-up, but I don't have anything specific in mind.

We can also make it configurable if y'all want! :)

saschagrunert:
> We can also make it configurable if y'all want! :)

I mean, it is already configurable by using or omitting `WatchdogSec=60s` in the service spec.

I rebased the PR on top of the latest fixes, PTAL again @cri-o/cri-o-maintainers

PR #8794, opened 2024-11-29
Only ctx timeout errors detected in the final stage can be ignored in createcontainer or runsandbox operation

### LINKED ISSUE #8790: When the nri.CreateContainer interface call encounters a ctx timeout, the container creation fails and the status record is not cleared, resulting in subsequent creation requests failing
### What happened?

When the nri.CreateContainer interface call encounters a ctx timeout, the container creation fails and the status record is not cleared, resulting in subsequent creation requests failing;
image

After the nri interface call timeout, subsequent creation requests initiated by kubelet should be able to execute normally, rather than being treated as duplicate requests.

Construct an nri plugin to timeout the call to the CreateContainer interface

_No response_

<details>

```console
$ crio --version
# paste output here
```
1.24.0
```console
# paste output here
```
1.24.1
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```
centos 7.9
</details>

<details>
physical
</details>

lidajia:
#### What type of PR is this?

/kind bug

Fixes #8790 

```release-note
None
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -521,7 +521,7 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 			log.Errorf(ctx, "CreateCtr: failed to save progress of container %s: %v", newContainer.ID(), err)
 		}
 		log.Infof(ctx, "CreateCtr: context was either canceled or the deadline was exceeded: %v", ctx.Err())
-		// no need to cleanup
+		// should not cleanup
 		storeResource = true
 		return nil, ctx.Err()
 	}
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -433,10 +433,11 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 	}
 
 	resourceCleaner := resourcestore.NewResourceCleaner()
-	noNeedToCleanup := false
+	// in some cases, it is still necessary to reserve container resources when an error occurs (such as just a request context timeout error)
+	storeResource := false
 	defer func() {
-		// no error, no need to cleanup
-		if retErr == nil || noNeedToCleanup {
+		// No errors or resource need to be stored, no need to cleanup
+		if retErr == nil || storeResource {
 			return
 		}
 		if err := resourceCleaner.Cleanup(); err != nil {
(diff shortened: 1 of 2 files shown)


saschagrunert (commented):
Thank you! Do you see any way to test this?

lidajia:
> Thank you! Do you see any way to test this?
congifg kubelet runtimeRequestTimeout=2m;
Construct an nri plugin make CreateContainer interface execution time exceeds 2 minutes（just sleep 2m)；

server/sandbox_run_linux.go
```diff
@@ -365,9 +365,10 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 	}
 
 	resourceCleaner := resourcestore.NewResourceCleaner()
+	noNeedToCleanup := false
```
haircommander, line 368: can this be `storeResource` or something, and can there be a comment explaining why we're doing this?

haircommander:
Thanks for the quick fix! there are NRI tests in `test/nri.bats` possibly we could extend the integration tests to make sure we don't regress? if that doesn't sound interesting to you @lidajia  a cri-o maintainer can do it

server/sandbox_run_linux.go
lidajia: Thank you for your suggestion. I have resubmitted it

server/container_create.go
```diff
@@ -519,6 +521,8 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 			log.Errorf(ctx, "CreateCtr: failed to save progress of container %s: %v", newContainer.ID(), err)
 		}
 		log.Infof(ctx, "CreateCtr: context was either canceled or the deadline was exceeded: %v", ctx.Err())
+		// no need to cleanup
```
haircommander, line 524: sorry this is not `no need to cleanup` but rather `should not cleanup` as we've explicitly stored it to return to later

lidajia: processed

saschagrunert (approved):

PR #8803, opened 2024-12-02
[release-1.26] OCPBUGS-45908: Only restore container if all bind mounts are defined

kwilczynski:
This is a manual cherry-pick of commit 429ef7c36

```mail
commit 429ef7c366406280cfa3daa15812d14b22953a7a
Author: Adrian Reber <areber@redhat.com>
Date:   Mon Sep 30 23:37:13 2024

    Only restore container if all bind mounts are defined
    
    To avoid the situation where a container that is restored via a registry
    mounts unexpected host paths into the container, this changes the
    restore behaviour of CRI-O.
    
    Previously all bind mounted paths in the original container which were
    defined for example like this:
    
        volumeMounts:
        - mountPath: /data
          name: data-volume
      volumes:
      - name: data-volume
        hostPath:
          path: /srv/container/data
    
    Were automatically mounted in the restored container without and
    definition necessary. This lead to the situation that the user does not
    know which path will be mounted if starting a restored container.
    
    Now CRI-O will refuse to restore a container if not all bind mounts are
    defined via the CRI CreateContainer RPC in the CreateContainerRequest
    message.
    
    CRI-O will now return an error that will look something likes this:
    
    Error: the container to restore (7f...be) expects following bind mounts defined (/data,/data2)
    
    Now the user has to explicitly add those bind mounts in the same way as
    it was done during initial container creation.
    
    Signed-off-by: Adrian Reber <areber@redhat.com>
```

/assign kwilczynski

> [!NOTE] 
> This cherry-pickl brings the following Pull Requests as a dependency:
> - https://github.com/cri-o/cri-o/pull/8150
> - https://github.com/cri-o/cri-o/pull/8786

```release-note
- Only restore container if all bind mounts are defined.
- Ensure that the "image" attribute has been provided.
```

### DIFF
--- a/server/container_restore.go
+++ b/server/container_restore.go
@@ -69,7 +69,7 @@ func (s *Server) CRImportCheckpoint(
 		return "", errors.New(`attribute "image" missing from container definition`)
 	}
 
-	if createConfig.Metadata == nil && createConfig.Metadata.Name == "" {
+	if createConfig.Metadata == nil || createConfig.Metadata.Name == "" {
 		return "", errors.New(`attribute "metadata" missing from container definition`)
 	}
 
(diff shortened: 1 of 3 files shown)


haircommander:
personally I think this can be closed. It's out of upstream support and not used downstream

kwilczynski:
> personally I think this can be closed. It's out of upstream support and not used downstream

@haircommander makes sense! Thank you!

kwilczynski:
Closing. This backport is no longer needed.

PR #8808, opened 2024-12-02
Use `monitor_env` for calling the OCI runtime

saschagrunert:
#### What type of PR is this?

/kind bug

We do not unset the env for calling the runtime directly, but we do when calling conmon(-rs). This behavioral difference should be aligned.

None
None

```release-note
Fixed bug to always inherit `monitor_env` when calling the OCI runtime.
```


sohankunkerkar (approved):
/lgtm

PR #8824, opened 2024-12-06
Fix issues when running on Linux with seccomp disabled

michalsieron:
#### What type of PR is this?

/kind bug

While I re-allowed building cri-o for Linux without seccomp in #8686 I later found some issues with it when actually running it. So this PR fixes some issues I found when running cri-o on a Linux system where both cri-o and runtime were compiled without seccomp support.

None

None

```release-note
None
```


saschagrunert (approved):
Thanks!

PR #8834, opened 2024-12-10
OCPBUGS-45112: Prevent systemd from blocking irqbalance restarts

MarSik:
#### What type of PR is this?
/kind bug

cri-o high performance runtime handler reconfigures and restarts irqbalance after every container with pinned cpus and moved interrupts is created. Systemd does not like this and will block further restarts when the StartLimitBurst (default 5) is crossed within the default timeout (10s). That is too strict as during system startup many such containers might be launched.

Install the provided file to /etc/systemd/system/irqbalance.service.d/ to prevent systemd from blocking the necessary restarts.

Fixes https://issues.redhat.com/browse/OCPBUGS-45112

This requires additional change to whatever build process is used to place the provided file into the proper location on the container node.

```release-note
cri-o high performance runtime handler reconfigures and restarts irqbalance after every container with pinned cpus and moved interrupts is created. Systemd does not like this and will block further restarts when the StartLimitBurst (default 5) is crossed within the default timeout (10s). That is too strict as during system startup many such containers might be launched.

Install the provided file to /etc/systemd/system/irqbalance.service.d/ to prevent systemd from blocking the necessary restarts.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,6 +296,7 @@ install.completions: ## Install the completions.
 install.systemd: ## Install the systemd unit files.
 	install ${SELINUXOPT} -D -m 644 contrib/systemd/crio.service $(PREFIX)/lib/systemd/system/crio.service
 	install ${SELINUXOPT} -D -m 644 contrib/systemd/crio-wipe.service $(PREFIX)/lib/systemd/system/crio-wipe.service
+	install ${SELINUXOPT} -D -m 644 contrib/systemd/irqbalance.conf $(PREFIX)/lib/systemd/system/irqbalance.service.d/restart-limits.conf
 
 .PHONY: uninstall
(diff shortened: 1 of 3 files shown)


haircommander:
/ok-to-test
/approve

LGTM thanks!

MarSik:
/retest-required

Even thought the one contrib file cannot break the tests in any way.

MarSik:
@haircommander Peter, I am a bit confused about how to fix the spec files in CI. The builds fail, because there is an unpackaged file present in the commit. However I cannot fix that here, because the spec file is pulled from Fedora. And fixing this in Fedora cannot happen without merging the PR first, because otherwise there will be a missing file.

How is a file addition supposed to be handled process wise?

haircommander:
/close

let's do this downstream instead

PR #8867, opened 2024-12-20
server: fix panic when default annotations are specified

haircommander:

```release-note
Fix a panic when default_annotations are used
```


kannon92 (commented):
Can we add a unit test to replicate the panic and make sure it is fixed?

haircommander:
I have yet to reproduce the panic in integration tests unfortunately. I feel like it's worth getting a change out as saving the map in the sandbox and then editing it later would cause this, but I suspect it'll be tricky to hit the exact race we do in prod locally

kannon92:
> I have yet to reproduce the panic in integration tests unfortunately. I feel like it's worth getting a change out as saving the map in the sandbox and then editing it later would cause this, but I suspect it'll be tricky to hit the exact race we do in prod locally

I still think a unit test would be sufficient to catch this. We have mocks for a reason.

haircommander:
> I still think a unit test would be sufficient to catch this. We have mocks for a reason.

the mocks don't get to this point--we're not fully unit testing the sandbox creation process. That's part of the motivation of https://github.com/cri-o/cri-o/issues/8289

That said, I did reproduce with an integration test, and verify my fix seems to fix. I am trying to find a reasonable number of containers to create. It always seems to reproduce at 1000 and I think I caught it once at 100, but even that feels like a bit much so I'm inclined to not include a test

kannon92:
Given the revert it would be nice to have some kind of test (unit / integration / e2e) that would cover this. I guess a e2e test with that many containers may be out of the realm for origin?

haircommander:
okay I figured out a test. it doesn't catch the panic, but does catch the context for the panic (that we didn't deep copy a map). 

@cri-o/cri-o-maintainers this should be ready PTAL

haircommander:
/retest
quay flakes

sohankunkerkar (approved):
/lgtm

sohankunkerkar:
In fact, `pull progress timeout should not timeout when set to 0` is a bit flaky too. We have a tracking issue for that: https://github.com/cri-o/cri-o/issues/8872

PR #8883, opened 2025-01-02
Switch to our log module for logging in iptables module

bitoku:
#### What type of PR is this?

/kind cleanup

Replaced klog with logrus to align with other logging in the codebase.
It isn't a big change for users since there's already the klog shim.
The replacement aligns with the way of the klog shim.

https://github.com/cri-o/cri-o/blob/f552e82b023c8e8034a1ecd85bbbecbaa7d61afc/internal/log/klog.go#L12-L38

The only change in behaviour is that the error logs are no longer shown from `iptables.go` just like other error logs when we run unit tests.

None

```release-note
Change log levels and formats for some iptables logs and klog related logs. 
```

### DIFF
--- a/internal/iptables/iptables.go
+++ b/internal/iptables/iptables.go
@@ -33,9 +33,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	utilversion "k8s.io/apimachinery/pkg/util/version"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/klog/v2"
 	utilexec "k8s.io/utils/exec"
 	utiltrace "k8s.io/utils/trace"
+
+	"github.com/cri-o/cri-o/internal/log"
 )
 
 // RulePosition holds the -I/-A flags for iptable.
@@ -218,10 +219,10 @@ type runner struct {
 
 // newInternal returns a new Interface which will exec iptables, and allows the
 // caller to change the iptables-restore lockfile path.
-func newInternal(exec utilexec.Interface, protocol Protocol, lockfilePath14x, lockfilePath16x string) Interface {
+func newInternal(ctx context.Context, exec utilexec.Interface, protocol Protocol, lockfilePath14x, lockfilePath16x string) Interface {
 	version, err := getIPTablesVersion(exec, protocol)
 	if err != nil {
-		klog.InfoS("Error checking iptables version, assuming version at least", "version", MinCheckVersion, "err", err)
+		log.Warnf(ctx, "Error checking iptables version, assuming version at least: %s (version=%q)", err, MinCheckVersion)
 		version = MinCheckVersion
 	}
 
@@ -246,8 +247,8 @@ func newInternal(exec utilexec.Interface, protocol Protocol, lockfilePath14x, lo
 }
 
 // New returns a new Interface which will exec iptables.
-func New(exec utilexec.Interface, protocol Protocol) Interface {
-	return newInternal(exec, protocol, "", "")
+func New(ctx context.Context, exec utilexec.Interface, protocol Protocol) Interface {
+	return newInternal(ctx, exec, protocol, "", "")
 }
 
 // EnsureChain is part of Interface.
@@ -359,7 +360,7 @@ func (runner *runner) SaveInto(table Table, buffer *bytes.Buffer) error {
 	// run and return
 	iptablesSaveCmd := iptablesSaveCommand(runner.protocol)
 	args := []string{"-t", string(table)}
-	klog.V(4).InfoS("Running", "command", iptablesSaveCmd, "arguments", args)
+	log.Debugf(context.Background(), "Running (command=%q arguments=%+v)", iptablesSaveCmd, args)
 	cmd := runner.exec.Command(iptablesSaveCmd, args...)
 	cmd.SetStdout(buffer)
 	stderrBuffer := bytes.NewBuffer(nil)
@@ -417,7 +418,7 @@ func (runner *runner) restoreInternal(args []string, data []byte, flush FlushFla
 		trace.Step("Locks grabbed")
 		defer func(locker iptablesLocker) {
 			if err := locker.Close(); err != nil {
-				klog.ErrorS(err, "Failed to close iptables locks")
+				log.Errorf(context.Background(), "Failed to close iptables locks: %s", err)
 			}
 		}(locker)
 	}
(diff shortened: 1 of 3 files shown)


haircommander:
WDYT about using our internal log package instead?

bitoku:
I decided to choose logrus because:
1. There are some occurrences of logrus in the codebase.
2. There is already klog shim, which uses logrus.
3. The arguments of `logrus.*` is the same as `klog.*`, and `log.*f` has different types of arguments.
4. `log.*f` takes ctx, but it's not used now in this module.

I don't have a strong preference though.
If we want to replace all log modules to ours, then I'd fix, but otherwise I think it's ok to use logrus.

haircommander:
we can use context.Backgroun() or context.TODO() for the ctx in log where appropriate. I am not feeling very stubborn but generally i think it's better to use our shim as it adds some additional pieces of metadata which have been useful in the past

internal/iptables/iptables.go
```diff
@@ -221,7 +221,7 @@ type runner struct {
 func newInternal(exec utilexec.Interface, protocol Protocol, lockfilePath14x, lockfilePath16x string) Interface {
 	version, err := getIPTablesVersion(exec, protocol)
 	if err != nil {
-		klog.InfoS("Error checking iptables version, assuming version at least", "version", MinCheckVersion, "err", err)
+		logrus.Debug("Error checking iptables version, assuming version at least", "version", MinCheckVersion, "err", err)
```
kwilczynski, line 224: Why not a warning here? Or keep it at the same log level. This seems potentially useful for the users.

internal/iptables/iptables.go
```diff
@@ -535,7 +535,7 @@ func (runner *runner) checkRuleWithoutCheck(table Table, chain Chain, args ...st
 		if sets.NewString(fields...).IsSuperset(argset) {
 			return true, nil
 		}
-		klog.V(5).InfoS("DBG: fields is not a superset of args", "fields", fields, "arguments", args)
+		logrus.Debug("DBG: fields is not a superset of args", "fields", fields, "arguments", args)
```
kwilczynski, line 538: I think we can drop the prefix "DBG:" now as we are making this into a proper debug log level.

internal/iptables/iptables.go
```diff
@@ -697,11 +697,11 @@ func getIPTablesRestoreWaitFlag(version *utilversion.Version, exec utilexec.Inte
...
-		klog.V(3).InfoS("Couldn't get iptables-restore version; assuming it doesn't support --wait")
+		logrus.Debug("Couldn't get iptables-restore version; assuming it doesn't support --wait")
 		return nil
 	}
 	if _, err := utilversion.ParseGeneric(vstring); err != nil {
-		klog.V(3).InfoS("Couldn't parse iptables-restore version; assuming it doesn't support --wait")
+		logrus.Debug("Couldn't parse iptables-restore version; assuming it doesn't support --wait")
 		return nil
 	}
 	return []string{WaitString}
```
kwilczynski, line 707: Here also, the log level would probably be better served as a warning to the user.

internal/iptables/iptables.go
bitoku: This is the current way to handle klog.InfoS .
Please see https://github.com/cri-o/cri-o/blob/f552e82b023c8e8034a1ecd85bbbecbaa7d61afc/internal/log/klog.go#L12-L38

bitoku: I just wanted to keep the current way, but it makes sense. thanks!

bitoku:
I updated the PR to address the comments given.

As a reference, these logs were like those using klogshim:
```go 
klog.InfoS("Error checking iptables version, assuming version at least", "version", MinCheckVersion, "err", err)
klog.InfoS("Running", "command", "iptables", "arguments", args)
```

```
time="2025-01-08T13:22:46Z" level=debug msg="Error checking iptables version, assuming version at least (version=\"1.4.11\" err=\"executable file not found in $PATH\")"
time="2025-01-08T13:22:46Z" level=debug msg="Running (command=\"iptables\" arguments=[-t aaa]"
```

With this PR, they output the same except for the back parenthesis ")" and log level.

internal/iptables/iptables.go
```diff
@@ -496,7 +497,7 @@ func trimhex(s string) string {
 // of hack and half-measures.  We should nix this ASAP.
 func (runner *runner) checkRuleWithoutCheck(table Table, chain Chain, args ...string) (bool, error) {
 	iptablesSaveCmd := iptablesSaveCommand(runner.protocol)
-	klog.V(1).InfoS("Running", "command", iptablesSaveCmd, "table", string(table))
+	log.Debugf(context.Background(), "Running (command=%q table=%q)", iptablesSaveCmd, string(table))
```
sohankunkerkar, line 500: Just a question: Do we need this only when the log level is set to debug?

sohankunkerkar: Perfect!

sohankunkerkar (commented):
/approve
LGTM

Thanks!

internal/log/klog.go
```diff
@@ -67,7 +67,7 @@ func writeKeysAndValues(b *strings.Builder, keysAndValues ...any) {
 			}
 		}
 
-		if i == len(keysAndValues) {
+		if i+2 == len(keysAndValues) {
```
haircommander, line 70: hm it's not immediately clear to me what this piece is doing

kwilczynski: Looks like a bracket index matching fix. But... Should probably be a separate Pull Request indeed.

bitoku: The last index `i` of this loop should be `i = len(keysAndValues) - 2` because `i += 2` in each loop, and `i` never equals to `len(keysAndValues)`.
Since this is a trivial change and related to klog, I guess it's ok to have it in this PR.

bitoku: actually never mind, I didn't take into account when `keysAndValues` is odd number, which shouldn't happen though.
I'll remove this fix for now.

bitoku:
@cri-o/cri-o-maintainers Can you PTAL?

internal/iptables/iptables.go
```diff
@@ -221,7 +222,7 @@ type runner struct {
 func newInternal(exec utilexec.Interface, protocol Protocol, lockfilePath14x, lockfilePath16x string) Interface {
 	version, err := getIPTablesVersion(exec, protocol)
 	if err != nil {
-		klog.InfoS("Error checking iptables version, assuming version at least", "version", MinCheckVersion, "err", err)
+		log.Warnf(context.Background(), "Error checking iptables version, assuming version at least: %s (version=%q)", err, MinCheckVersion)
```
saschagrunert, line 225: We should probably pass a context here.

bitoku: Do you mean we should pass down the context from here?
https://github.com/cri-o/cri-o/blob/a5c51b8b006a6452cc596cd58bb7af30a2e78cc2/server/server.go#L395-L401

saschagrunert: Yes, it's the main context where the server runs.

PR #8911, opened 2025-01-13
[OCPBUGS-48334] watchdog: Decouple CNI plugin initialization from CRI-O health checks

sohankunkerkar:
Fixes https://issues.redhat.com/browse/OCPBUGS-48334
The idea is decouple CRI-O's readiness from the CNI plugin readiness.

#### What type of PR is this?

/kind bug

```release-note
Decouple CNI plugin initialization from CRI-O health checks.
```

### DIFF
--- a/server/health.go
+++ b/server/health.go
@@ -4,9 +4,18 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	cri "k8s.io/cri-client/pkg"
+
+	"github.com/cri-o/cri-o/internal/log"
+)
+
+var (
+	cniPluginInitialized atomic.Bool
+	cniInitOnce          sync.Once
 )
 
 func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) error {
@@ -31,7 +40,15 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
 		return errors.New("runtime conditions are nil")
 	}
 
+	s.cniPluginReadinessCheck(ctx)
+
 	for _, c := range response.GetStatus().GetConditions() {
+		if c.GetType() == "NetworkReady" {
+			if !cniPluginInitialized.Load() {
+				log.Warnf(ctx, "CNI plugin not yet initialized. Ignoring NetworkReady status: %v, message: %s, reason: %s", c.GetStatus(), c.GetMessage(), c.GetReason())
+				continue
+			}
+		}
 		if !c.GetStatus() {
 			return fmt.Errorf(
 				"runtime status %q is invalid: %s (reason: %s)",
@@ -41,4 +58,17 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
 	}
 
 	return nil
+}
+
+func (s *Server) cniPluginReadinessCheck(ctx context.Context) {
+	cniInitOnce.Do(func() {
+		go func() {
+			if err := s.waitForCNIPlugin(ctx, ""); err != nil {
+				log.Errorf(ctx, "CNI plugin not ready: %v", err)
+			} else {
+				log.Infof(ctx, "CNI plugin is ready")
+				cniPluginInitialized.Store(true)
+			}
+		}()
+	})
 }
--- a/internal/watchdog/watchdog.go
+++ b/internal/watchdog/watchdog.go
@@ -59,7 +59,7 @@ func (w *Watchdog) Start(ctx context.Context) error {
 
 	log.Infof(ctx, "Starting systemd watchdog using interval: %v", interval)
 
-	go wait.Forever(func() {
+	go wait.Until(func() {
 		if err := w.runHealthCheckers(ctx, interval); err != nil {
 			log.Errorf(ctx, "Will not notify watchdog because CRI-O is unhealthy: %v", err)
 			return
@@ -81,7 +81,7 @@ func (w *Watchdog) Start(ctx context.Context) error {
 		}); err != nil {
 			log.Errorf(ctx, "Failed to notify watchdog: %v", err)
 		}
-	}, interval)
+	}, interval, ctx.Done())
 
 	return nil
 }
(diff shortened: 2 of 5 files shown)


server/utils.go
```diff
@@ -234,3 +234,16 @@ func stopTimeoutFromContext(ctx context.Context) int64 {
 	log.Debugf(ctx, "Using stop timeout: %v", timeout)
 	return timeout
 }
+
+// WaitForCNIPlugin waits for the CNI plugin to be ready before proceeding.
+func (s *Server) WaitForCNIPlugin(ctx context.Context, sboxName string) error {
```
haircommander, line 239: nit: how about server/sandbox_network.go?

server/health.go
```diff
@@ -31,7 +33,28 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
...
 	}
 
+	cniReadyCh := make(chan bool, 1)
+
+	// Trigger the CNI readiness check asynchronously.
+	if err := s.cniPluginReadinessCheck(ctx, cniReadyCh); err != nil {
+		log.Errorf(ctx, "Waiting for CNI readiness: %v", err)
+	}
+
+	cniReady := <-cniReadyCh
```
haircommander, line 43: this still looks like it will block to me

server/health.go
```diff
@@ -42,3 +64,18 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
...
+func (s *Server) cniPluginReadinessCheck(ctx context.Context) {
+	cniInitOnce.Do(func() {
+		go func() {
+			watcher := s.config.CNIPluginAddWatcher()
+			log.Infof(ctx, "Waiting for CNI plugin readiness.")
+			if ready := <-watcher; !ready {
+				log.Errorf(ctx, "CNI plugin not ready after waiting.")
+			} else {
+				log.Infof(ctx, "CNI plugin is ready.")
+				cniPluginInitialized = true
```
haircommander, line 77: I think this needs to either be an atomic.Bool or be protected by a lock

sohankunkerkar: I was thinking about it but `cniPluginInitialized` will not be access by multiple goroutines so I thought of keeping it as-is.

haircommander: it will be written by a separate one than it's read by right?

sohankunkerkar: ok, I see your point.

server/health.go
```diff
@@ -42,3 +64,18 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
 
 	return nil
 }
+
+func (s *Server) cniPluginReadinessCheck(ctx context.Context) {
+	cniInitOnce.Do(func() {
+		go func() {
+			watcher := s.config.CNIPluginAddWatcher()
```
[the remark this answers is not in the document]
sohankunkerkar: Yeah, fixed

server/health.go
```diff
@@ -31,7 +40,21 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
 		return errors.New("runtime conditions are nil")
 	}
 
+	s.cniPluginReadinessCheck(ctx)
+
 	for _, c := range response.GetStatus().GetConditions() {
+		if c.GetType() == "NetworkReady" {
+			if !cniPluginInitialized.Load() {
+				log.Warnf(ctx, "CNI plugin not yet initialized. Ignoring NetworkReady status: %v, message: %s, reason: %s", c.GetStatus(), c.GetMessage(), c.GetReason())
+				continue
```
bitoku, line 49: We don't have to check this when the `NetworkReady` condition is true, I guess?

server/health.go
```diff
@@ -31,7 +40,21 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
...
+			}
+			if !c.GetStatus() {
+				return fmt.Errorf(
+					"runtime status %q is invalid: %s (reason: %s)",
+					c.GetType(), c.GetMessage(), c.GetReason(),
+				)
+			}
```
bitoku, line 56: I think this branch is not needed because it's covered by later code.

server/health.go
sohankunkerkar: Actually, we need to check this when `NetworkReady` is true. If the CNI plugin isn't initialized, CRI-O may fail to send heartbeats, causing systemd to send a SIGABRT. This check ensures CRI-O isn't marked unhealthy prematurely and helps prevent failures by confirming the network plugin is fully ready.

sohankunkerkar:
@saschagrunert Could you take a look at it?

server/health.go
```diff
@@ -42,3 +59,16 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
...
 	return nil
 }
+
+func (s *Server) cniPluginReadinessCheck(ctx context.Context) {
+	cniInitOnce.Do(func() {
+		go func() {
+			if err := s.waitForCNIPlugin(ctx, ""); err != nil {
+				log.Errorf(ctx, "CNI plugin not ready: %v", err)
+			} else {
+				log.Infof(ctx, "CNI plugin is ready.")
```
saschagrunert, line 69:
```suggestion
				log.Infof(ctx, "CNI plugin is ready")
```

server/health.go
```diff
@@ -42,3 +59,16 @@ func (s *Server) checkCRIHealth(ctx context.Context, timeout time.Duration) erro
 
 	return nil
 }
+
+func (s *Server) cniPluginReadinessCheck(ctx context.Context) {
+	cniInitOnce.Do(func() {
+		go func() {
```
saschagrunert, line 65: Would this work avoid the additional `go func`?
```suggestion
	go cniInitOnce.Do(func() {
```

saschagrunert (commented):
Just two nits, otherwise LGTM. An integration test would be beneficial but we could also follow-up on that.

server/health.go
sohankunkerkar: The idea is to start the readiness check immediately but run the actual plugin check asynchronously.

saschagrunert (approved):

Issue #8927, opened 2025-01-21
oci: Add WithDomainname


aojea:
A `domainname` field was added to the OCI spec in v1.1.0:

- https://github.com/opencontainers/runtime-spec/pull/1156
- https://github.com/opencontainers/runtime-spec/issues/592

Prior to this users would need to set it with a sysctl, but now runtimes should be able to setdomainname(2)

Containerd already implemented it in https://github.com/containerd/containerd/pull/7869

This feature will be very useful for Kubernetes so it can provide the cluster domain to the Pods avoiding existing hacks https://github.com/kubernetes/enhancements/pull/4972

PR #8929, opened 2025-01-21
container: add support for domainname

sohankunkerkar:
#### What type of PR is this?

/kind feature

Fixes # https://github.com/cri-o/cri-o/issues/8927

```release-note
Add support for domainname
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -112,6 +112,9 @@ type Container interface {
 	// SpecAddDevices adds devices from the server config, and container CRI config
 	SpecAddDevices([]device.Device, []device.Device, bool, bool) error
 
+	// SpecAddDomainName adds a domain name to the container's spec.
+	SpecAddDomainName(string)
+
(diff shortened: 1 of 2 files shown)


haircommander:
hm it doesn't seem like we're calling it anywhere yet right?

sohankunkerkar:
> hm it doesn't seem like we're calling it anywhere yet right?

Yeah, I think this is more of a 'ready to consume for future' thing

haircommander:
I think we can integrate it, I don't personally see a point in adding the function if we don't use it

aojea:
> I think we can integrate it, I don't personally see a point in adding the function if we don't use it

not strong arguments but let me try to see if I can convince you ;) 

1. correctness,  as it is part of the OCI spec https://github.com/opencontainers/runtime-spec/pull/1156
2. possible future use in kubernetes https://github.com/kubernetes/enhancements/pull/4972#issuecomment-2604737533 , I don't want to get a feature that only work in one runtime

haircommander:
you can have my word we'll merge it if there's a CRI field we can pull the domainname from :)

PR #8940, opened 2025-01-24
[WIP] Add debug option to runtime command, and make it visible from crio log.

bitoku:
#### What type of PR is this?

/kiind other

This PR enables us to retrieve debug logs from crun/runc runtimes.
I believe this will help us troubleshoot around runtime.

Here are crun / runc options.

https://github.com/containers/crun/blob/main/crun.1.md
> --debug Produce verbose output.

https://github.com/opencontainers/runc/blob/8702864454c4001d04aa3eaf49c2f8a8310438f5/man/runc.8.md
> --debug : Enable debug logging.

None

```release-note
Enabled debug logs for low-level container runtime when --log-level=debug.
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -316,7 +316,7 @@ func (r *runtimeOCI) CreateContainer(ctx context.Context, c *Container, cgroupPa
 
 // StartContainer starts a container.
 func (r *runtimeOCI) StartContainer(ctx context.Context, c *Container) error {
-	_, span := log.StartSpan(ctx)
+	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 	c.opLock.Lock()
 	defer c.opLock.Unlock()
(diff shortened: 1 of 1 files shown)


internal/oci/runtime_oci.go
```diff
@@ -1440,6 +1443,9 @@ func (r *runtimeOCI) defaultRuntimeArgs() []string {
 	if r.config.CgroupManager().IsSystemd() {
 		args = append(args, "--systemd-cgroup")
 	}
+	if r.config.LogLevel == "debug" {
```
haircommander, line 1446: I almost wonder if this should be trace level as debug is already pretty noisy, but then again I think this would be useful. I'm not sure, any thoughts @cri-o/cri-o-maintainers ?

bitoku: > I almost wonder if this should be trace level as debug is already pretty noisy

I think so too after looking at the log of failed test logs... 
Also crun log-level is `error` by default, but there's warning logs. And runc doesn't have `log-level` option.
It would be enough to have only `error` logs maybe?

PR #8975, opened 2025-02-03
Switch to golang native error joining and errgroup

saschagrunert:
#### What type of PR is this?

/kind cleanup

Using the sync/errgroup package is slightly more lean compared to the `k8s.io/apimachinery/pkg/util/errors` package. We can also avoid using the package in other places like the hostport manager and rely on golang native error joining.

None
None

```release-note
None
```


sohankunkerkar (approved):
/lgtm

PR #8996, opened 2025-02-11
Add OCI artifact support using libimage

saschagrunert:
#### What type of PR is this?

/kind feature

Goal of this PR is to PoC a possible split-path to pull artifacts before actually pulling the container image.

None
None

```release-note
Added OCI artifact store support. This means that CRI-O now stores artifacts per default in
`/var/lib/containers/storage/artifacts`, which are handled as pinned images. The artifacts
are pullable using `crictl pull` and will also show-up in `crictl images`. It is also possible to
`crictl inspecti` them as well as deleting them using `crictl rmi`.
```


saschagrunert:
@cri-o/cri-o-maintainers PTAL

This is the foundation work which moves the existing seccomp OCI artifact feature on top of the new store-based methods of libimage. The integration tests show that it works as intended and we can now manage those artifacts using the CRI/crictl. Artifacts are handled as pinned images per default.

As a next step we can make those artifacts mountable, having full OCI artifact support for the image volumes feature.

saschagrunert:
Had to rebase, PTAL @cri-o/cri-o-maintainers

internal/storage/image.go
```diff
@@ -871,15 +873,25 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+				DestinationCtx:   options.DestinationCtx,
+				OciDecryptConfig: options.OciDecryptConfig,
+				ProgressInterval: options.ProgressInterval,
+				Progress:         options.Progress,
+			})
+			if err != nil {
+				return RegistryImageReference{}, err
+			}
+		} else {
+			return RegistryImageReference{}, fmt.Errorf("unable to try puling possible OCI artifact: %w", err)
```
bitoku, line 893:
```suggestion
			return RegistryImageReference{}, fmt.Errorf("unable to try pulling possible OCI artifact: %w", err)
```

server/image_remove.go
```diff
@@ -69,7 +69,7 @@ func (s *Server) removeImage(ctx context.Context, imageRef string) (untagErr err
 
 		status, statusErr = s.ContainerServer.StorageImageServer().ImageStatusByName(s.config.SystemContext, name)
 		if statusErr != nil {
-			log.Errorf(ctx, "Error getting image status %s: %v", name, statusErr)
+			log.Debugf(ctx, "Error getting image status %s: %v", name, statusErr)
```
bitoku, line 72: Why did you change this debug?

server/image_remove.go
```diff
@@ -107,6 +107,10 @@ func (s *Server) removeImage(ctx context.Context, imageRef string) (untagErr err
 		return untagErr
 	}
 
+	if err := s.ArtifactStore().Remove(ctx, imageRef); err != nil {
+		log.Infof(ctx, "Unable to remove artifact: %v", err)
```
bitoku, line 111:
```suggestion
		log.Errorf(ctx, "Unable to remove artifact: %v", err)
```

server/image_list.go
```diff
@@ -26,7 +28,14 @@ func (s *Server) ListImages(ctx context.Context, req *types.ListImagesRequest) (
...
 			resp := &types.ListImagesResponse{}
-			if status != nil {
+
+			if status == nil {
+				if artifact, err := s.ArtifactStore().Status(ctx, filterImage.Image); err == nil {
+					resp.Images = append(resp.Images, artifact.CRIImage())
+				} else if !errors.Is(err, ociartifact.ErrNotFound) {
+					log.Errorf(ctx, "Unable to get filtered artifact: %v", err)
+				}
+			} else {
```
bitoku, line 38: nit: It can early-return when it found status.

server/image_status.go
```diff
@@ -34,7 +36,15 @@ func (s *Server) ImageStatus(ctx context.Context, req *types.ImageStatusRequest)
...
-		log.Infof(ctx, "Image %s not found", img.Image)
+		if artifact, err := s.ArtifactStore().Status(ctx, img.Image); err == nil {
+			return &types.ImageStatusResponse{
+				Image: artifact.CRIImage(),
+			}, nil
+		} else if !errors.Is(err, ociartifact.ErrNotFound) {
+			log.Errorf(ctx, "Unable to get artifact: %v", err)
+		}
+
+		log.Infof(ctx, "Neither image nor artfiact %s found", img.Image)
```
bitoku, line 47: It doesn't need else because it returns in the previous if block.
It logs twice when the error is not ErrNotFound.

```suggestion
		}
		if errors.Is(err, ociartifact.ErrNotFound) {
			log.Infof(ctx, "Neither image nor artfiact %s found", img.Image)
		} else {
			log.Errorf(ctx, "Unable to get artifact: %v", err)
		}
```

internal/storage/image.go
bitoku, line 893: nit: It can early-return.
```suggestion
		if !errors.Is(err, ociartifact.ErrIsAnImage) {
			return RegistryImageReference{}, fmt.Errorf("unable to try puling possible OCI artifact: %w", err)
		}
		manifestBytes, err = copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
			SourceCtx:        &srcSystemContext,
			DestinationCtx:   options.DestinationCtx,
			OciDecryptConfig: options.OciDecryptConfig,
			ProgressInterval: options.ProgressInterval,
			Progress:         options.Progress,
		})
		if err != nil {
			return RegistryImageReference{}, err
		}
```

server/image_remove.go
saschagrunert: We need to fall through this case in case of an artifact, means we would indicate an error in the logs which can still succeed if the object is an artifact.

bitoku: I'm concerned that `statusErr` is something else than NotFound and not logged.
How about using `Warnf` or having a branch based on the error?

saschagrunert: Sounds good, let's use `Warnf`.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

saschagrunert:
@cri-o/cri-o-maintainers PTAL

We need a mount feature to make this able to work with image volumes. 

See https://github.com/containers/podman/pull/25397

test/oci_artifacts.bats
```diff
@@ -0,0 +1,46 @@
...
+
+function setup() {
+	setup_test
+}
+
+function teardown() {
+	cleanup_test
+}
+
+ARTIFACT_IMAGE=quay.io/crio/seccomp:v2
```
haircommander, line 14: maybe should be added to IMAGES() in test/helpers.bash so we cache the image before running?

saschagrunert: Actually, the copyimg binary does not work with artifacts. Should we follow-up on that or is it fine as-is?

sohankunkerkar: Let's get this in. We can do this as a follow-up. This PR is already big.

sohankunkerkar (approved):
/lgtm

internal/ociartifact/store.go
```diff
@@ -0,0 +1,432 @@
...
+		return nil, fmt.Errorf("build image source: %w", err)
+	}
+
+	defer func() {
+		if err := s.impl.CloseImageSource(src); err != nil {
+			log.Warnf(ctx, "Unable to close image source: %v", err)
+		}
+	}()
+
+	manifestBytes, mimeType, err := s.impl.GetManifest(ctx, src, nil)
```
sohankunkerkar, line 128: I was wondering if you could accommodate https://github.com/containers/skopeo/pull/2527/files#diff-6cac070631a328cbecde69cd12534e6b77254bcf5d366c4210afaa76e70bfe67R109 in this commit.
This might fixes https://github.com/cri-o/cri-o/issues/9028

test/oci_artifacts.bats
saschagrunert: Sounds good, created https://github.com/cri-o/cri-o/issues/9041 for that.

internal/ociartifact/store.go
saschagrunert: That's already in: https://github.com/cri-o/cri-o/pull/8996/files#diff-214891d73adc1da245487a6f80db7c6891c760310da9179d4c912b46047d6848R68

Issue #8822, opened 2024-12-04
Add validations before returning the Sandbox


xw19:
PR #8610 introduces a builder for initializing the pod sandbox. This means we build the pod sandbox step by step. In the end, we will request the builder to get the sandbox we have built.
When we request the builder to get the pod sandbox, the sandbox builder should validate that the required fields are present before returning the sandbox.

### Required fields:

    Id
    name
    createdAt
    config

saschagrunert:
@xw19 thank you for opening the issue, do you consider to work on this?

xw19:
Yes @saschagrunert

xw19:
@haircommander @saschagrunert @kwilczynski @sohankunkerkar I believe the config and name are not required only, id, createdAt and criSandbox are required. Is the assumption correct?

haircommander:
Eevery field we unconditionally set previously should be required IMO. I also wonder if there should be a function called counter to make sure we call all of the pieces of the construction. the source of anxiety motivating this is we miss some subtle field in the sandbox that formerly was passed to the sandbox constructor which would make it easier to catch

xw19:
See this is an example of how we are using sbox.New earlier

```
mySandbox, err = sandbox.New(sandboxID, "", "", "", "",
make(map[string]string), make(map[string]string), "", "",
&types.PodSandboxMetadata{}, "", "", false, "", "", "",
[]*hostport.PortMapping{}, false, time.Now(), "", nil, nil)
```
This is how we are using now
```
// Setup test vars
createdAt := time.Now()
sbox := sandbox.NewBuilder()
sbox.SetID("sandboxID")
sbox.SetLogDir("test")
sbox.SetCreatedAt(createdAt)
err = sbox.SetCRISandbox(sbox.ID(), make(map[string]string), make(map[string]string), &types.PodSandboxMetadata{})
if err != nil {
panic(err)
}
sbox.SetPrivileged(false)
sbox.SetPortMappings([]*hostport.PortMapping{})
sbox.SetHostNetwork(false)
sbox.SetContainers(memorystore.New[*oci.Container]())
mySandbox, err = sbox.GetSandbox()
```

Most of these values are default type values. Not sure how we validate. Please explain a bit more

xw19:
@haircommander can you check is this what you want for example if id is not given the Getsandbox should fail? https://github.com/cri-o/cri-o/pull/8908.

I feel the sandbox object currently is designed in a way that it will not fail even if fields like id is not there.

[...]

PR #9008, opened 2025-02-15
Validate before creating Sandbox

xw19:
#### What type of PR is this?
/kind cleanup

Ref #8822 

This is a series of PR for validating sandbox. More PRs to follow 

No
```release-note
No
```


xw19:
PTAL @cri-o/cri-o-maintainers

internal/factory/container/container_test.go
```diff
@@ -104,11 +104,27 @@ var _ = t.Describe("Container", func() {
...
+			sbox.SetPortMappings([]*hostport.PortMapping{})
+			sbox.SetHostNetwork(false)
+			sbox.SetUsernsMode("")
+			sbox.SetPodLinuxOverhead(nil)
+			sbox.SetPodLinuxResources(nil)
 			sbox.SetCreatedAt(time.Now())
 			err = sbox.SetCRISandbox(sbox.ID(), make(map[string]string), make(map[string]string), &types.PodSandboxMetadata{})
 			Expect(err).ToNot(HaveOccurred())
-			sbox.SetPrivileged(false)
+			// sbox.SetPrivileged(false)
```
haircommander, line 127: hm why is this commented?

haircommander:
@cri-o/cri-o-maintainers how do others feel about this? it adds some bits of overhead per sandbox , which is ephemerally kept so it shouldn't have a huge effect. i like how it prevents coding errors but I am worried the overhead isn't worth it. 

Also: a follow up here could be to use a bitmask and constantly stored field locations within that bitmask instead of having a struct of bools which could optimize alignment within the struct?

xw19:
@haircommander With this PR I found one bug where previously I missed adding few method calls. I think this will help in long run ( I am biased because I have created this patch). I also rewritten parts of it  use bit mask.

internal/lib/sandbox/builder.go
```diff
@@ -123,37 +123,112 @@ type Builder interface {
...
 	sandboxRef := b.sandboxRef
 	b.config = nil
 	b.sandboxRef = nil
+	b.validations = 0
 
 	return sandboxRef, nil
 }
 
+// Validate validates the sandbox.
+func (b *sandboxBuilder) Validate() error {
```
haircommander, line 199: one more nit: we could short circut this by storing a const of all of the validations and then directly comparing that value to the validations . Then we could loop through the map to report what wasn't done

haircommander:
> @haircommander With this PR I found one bug where previously I missed adding few method calls.  

this is reason enough to go forward with this IMO :)

internal/lib/sandbox/builder.go
xw19: done

internal/lib/sandbox/builder.go
```diff
@@ -123,37 +123,140 @@ type Builder interface {
...
+		"sandbox setPrivileged":        validationSetPrivileged,
+		"sandbox setRuntimeHandler":    validationSetRuntimeHandler,
+		"sandbox setReslovPath":        validationSetResolvPath,
+		"sandbox setHostname":          validationSetHostname,
+		"sandbox setPortMappings":      validationSetPortMappings,
+		"sandbox setHostNetwork":       validationSetHostNetwork,
+		"sandbox setUsernsMode":        validationSetUsernsMode,
+		"sandbox setPodLinuxOverhead":  validationSetPodLinuxOverhead,
+		"sandbox setPodLinuxResources": validationSetPodLinuxResources,
+	}
```
bitoku, line 249: We can make these keys consistent, for example to use field name after `validationSet*`.

And we can use package level var instead of function level to avoid allocating this map in every call (or is it optimised by compiler?).

saschagrunert (approved):

internal/lib/sandbox/builder.go
```diff
@@ -123,37 +123,140 @@ type Builder interface {
...
+	validationSetPodLinuxResources
+)
+
+// setValidation sets the given validation flag.
+func (v *sandboxValidations) setValidation(flag sandboxValidations) {
+	*v |= flag
+}
+
+// isSet returns true if the given validation flag is set.
+func (v *sandboxValidations) isSet(flag sandboxValidations) bool {
```
kwilczynski, line 162:
```suggestion
func (v *sandboxValidations) isValidationSet(flag sandboxValidations) bool {
```

saschagrunert (approved):

Issue #9004, opened 2025-02-13
GetContainerEvents panics due to race condition


ruiwen-zhao:
### What happened?

We recently see panics from https://github.com/cri-o/cri-o/blob/main/server/container_events.go happening a lot:

```
February 10th 2025, 20:20:41.695	goroutine 1 [select, 212 minutes]:	 - 	 - 
February 10th 2025, 20:20:41.695	created by github.com/cri-o/cri-o/server.(*Server).GetContainerEvents.func1	 - 	 - 
February 10th 2025, 20:20:41.695	#011github.com/cri-o/cri-o/server/container_events.go:42 +0x2a fp=0xc0041f1c90 sp=0xc0041f1c70 pc=0x24cebaa	 - 	 - 
February 10th 2025, 20:20:41.695	#011runtime/panic.go:796 +0x88 fp=0xc0041f1ca0 sp=0xc0041f1c90 pc=0xea03e8	 - 	 - 
February 10th 2025, 20:20:41.695	runtime.deferCallSave(0xc0041f1d60, 0xc0041f1f50?)	 - 	 - 
February 10th 2025, 20:20:41.695	#011sync/map.go:354 +0x2b8 fp=0xc0041f1f08 sp=0xc0041f1e38 pc=0xee8538	 - 	 - 
February 10th 2025, 20:20:41.695	panic: sync: negative WaitGroup counter	 - 	 - 
February 10th 2025, 20:20:41.695	#011runtime/panic.go:987 +0x3ba fp=0xc0041f1b48 sp=0xc0041f1a88 pc=0xea07fa	 - 	 - 
February 10th 2025, 20:20:41.695	sync.(*WaitGroup).Add(0xc002c68360?, 0xc00360c000?)	 - 	 - 
February 10th 2025, 20:20:41.695	#011runtime/panic.go:769 +0x1a5 fp=0xc0041f1ce8 sp=0xc0041f1ca0 pc=0xea0205	 - 	 - 
February 10th 2025, 20:20:41.695	panic({0x2786880, 0x2ed0680})	 - 	 - 
February 10th 2025, 20:20:41.695	main.main.func2(0xc000967040)	 - 	 - 
February 10th 2025, 20:20:41.695	sync.(*Map).Range(0xc0005be700?, 0xc0041f1f40)	 - 	 - 
February 10th 2025, 20:20:41.695	#011sync/waitgroup.go:108	 - 	 - 
[...]

PR #9014, opened 2025-02-18
server: fix races in GetContainerEvents

spruett345:
#### What type of PR is this?

/kind bug

This patch addresses the race conditions which cause at least one of the crashes in https://github.com/cri-o/cri-o/issues/9004.

The code in GetContainerEvents has two possible race conditions.

 1. The deferred broadcastEvents may start and complete in between adding the connection to the store and calling `wg.Add()` which causes an invalid state on the WaitGroup, causing a panic.
 2. Multiple messages may be processed by the broadcast before a client goroutine wakes up from the first Done() call and removes itself, leading to multiple Done() calls for a single Add(), which is invalid and causes a panic.

Fixes https://github.com/cri-o/cri-o/issues/9004

```release-note
None
```

### DIFF
--- a/server/container_events.go
+++ b/server/container_events.go
@@ -9,8 +9,20 @@ import (
 )
 
 type containerEventConn struct {
-	wg  sync.WaitGroup
-	err error
+	ch   chan struct{}
+	once sync.Once
+	err  error
+}
+
+func (c *containerEventConn) done() {
+	// only ever close the channel once, even if multiple messages are sent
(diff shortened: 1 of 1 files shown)


bitoku:
/lgtm

It's nit but we can use channel instead of waitgroup. It's make it more clear that the conn must be closed only once.

sohankunkerkar (approved):
/lgtm

PR #9016, opened 2025-02-22
New UpdatePodSandboxResources CRI API handler

### LINKED ISSUE #8990: Implement new UpdatePodSandboxResources CRI method
Pod sandbox cgroups are managed by the Kubelet, but the pod sandbox resources are still passed through to the `RunPodSandbox` request. With the new InPlacePodVerticalScaling (pod resize) feature, these resources are now mutable. To support NRI plugins that need to know the pod sandbox resources, we're adding a new `UpdatePodSandboxResources` informational CRI call for the purpose of passing the information through to NRI.

Kubernetes KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#cri-changes
Kubernetes CRI PR: https://github.com/kubernetes/kubernetes/pull/128123

It would be great to get a draft implementation, but the actual changes are blocked on https://github.com/kubernetes/kubernetes/pull/128123

felipeagger:
This PR Introduces the handler for the UpdatePodSandboxResources CRI API method which is needed for vertical pod autoscaling. 

Depends on https://github.com/kubernetes/kubernetes/pull/128123 and https://github.com/containerd/nri/pull/141.

#### What type of PR is this?

/kind feature

Fixes #8990

```release-note
None
```


internal/nri/nri.go
```diff
@@ -169,6 +172,25 @@ func (l *local) RunPodSandbox(ctx context.Context, pod PodSandbox) error {
...
+		return nil
+	}
+
+	l.Lock()
+	defer l.Unlock()
+
+	request := &nri.UpdatePodSandboxRequest{
+		Pod:                    podSandboxToNRI(pod),
+		OverheadLinuxResources: &nri.LinuxResources{},
+		LinuxResources:         &nri.LinuxResources{},
```
felipeagger, line 186: How can populate &nri.LinuxResources?

server/sandbox_update.go
```diff
@@ -0,0 +1,29 @@
...
+func (s *Server) UpdatePodSandboxResources(ctx context.Context, req *types.UpdatePodSandboxResourcesRequest) (*types.UpdatePodSandboxResourcesResponse, error) {
+	ctx, span := log.StartSpan(ctx)
+	defer span.End()
+
+	sb, err := s.getPodSandboxFromRequest(ctx, req.PodSandboxId)
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "could not find pod %q: %v", req.PodSandboxId, err)
+	}
+
+	err = s.nri.updatePodSandbox(ctx, sb)
```
felipeagger, line 23: Do we need to have something else here? besides the NRI call? something separate by O.S?

go.mod
```diff
@@ -236,8 +236,14 @@ require (
...
 	sigs.k8s.io/json v0.0.0-20241010143419-9aa6b5e7a4b3 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.2 // indirect
 	tags.cncf.io/container-device-interface/specs-go v0.8.0 // indirect
 )
+
+// TODO: Replace once version is cut. - https://github.com/kubernetes/kubernetes/pull/128123
+replace (
+	k8s.io/cri-api => github.com/felipeagger/kubernetes/staging/src/k8s.io/cri-api v0.0.0-20250223122702-3656c969cb13
+	k8s.io/cri-client => github.com/felipeagger/kubernetes/staging/src/k8s.io/cri-client v0.0.0-20250223122702-3656c969cb13
+)
```
saschagrunert, line 249: 1.33.0-alpha.3 should be released on Tuesday 4th March 2025

internal/nri/nri.go
haircommander: you can add a function to sandbox_linux.go that populates a created UPdatePodSandboxRequest object with the {Overhead,}LinuxResources objects.

felipeagger: nice... like [this](https://github.com/cri-o/cri-o/pull/9016/files#diff-b106070a06d88855804587b5d63197c9ebe5f8fdd5dff62abd1e697b46893db6R24)?

server/sandbox_update.go
klihub, line 23: If we do it like this, then my above comment (about the pod already having the updated resources) is probable incorrect. To me it looks like that we use the original non-updated pod here, which is correct. But then, after having called out to NRI, I think we should also update the pod data with the updated resources, so that a query over CRI will return the pod with the update resources and overhead instead of the old ones.

internal/nri/sandbox_other.go
```diff
@@ -9,3 +9,11 @@ import (
 func podSandboxToNRI(pod PodSandbox) *nri.PodSandbox {
 	return commonPodSandboxToNRI(pod)
 }
+
+func createUpdatePodSandboxRequest(pod PodSandbox) *nri.UpdatePodSandboxRequest {
+	podNri := commonPodSandboxToNRI(pod)
+
+	return &nri.UpdatePodSandboxRequest{
+		Pod: podNri,
+	}
```
felipeagger, line 18: **To reviewers**:  How should we treat other OS (e.g: freebsd) ?

klihub: At the moment, the NRI implementation for all other OS's is an effective no-op. Just allows compilation without any real functionality.

felipeagger:
go: k8s.io/cri-api in vendor/modules.txt requires go >= 1.24.0 (running go 1.23.4; GOTOOLCHAIN=local)

we can update cri-o golang version to 1.24 ?

go.mod
```diff
@@ -1,5 +1,7 @@
 go 1.23.4
 
+toolchain go1.24.1
+
```
saschagrunert, line 4: Let's get https://github.com/cri-o/cri-o/pull/9051 in before this PR

felipeagger: Nice, tks.

felipeagger:
the builds are failing because are using go1.23...
/retest

haircommander:
@felipeagger are you able to rebase? I believe we're using latest go now cc @saschagrunert

felipeagger:
Yep, i'll rebase soon.

server/sandbox_update.go
bitoku: I'm reopening the comment since it doesn't look resolved to me. Or am I missing something?

felipeagger: I didn't understand—can you briefly explain what needs to be adjusted here?

bitoku: `req.Overhead` and `req.Resources` are not used, but the pod is supposed to be updated based on the values, isn't it?

haircommander:
@klihub can you PTAL again :pray: ?

internal/nri/nri.go
```diff
@@ -169,6 +172,26 @@ func (l *local) RunPodSandbox(ctx context.Context, pod PodSandbox) error {
...
+	request := &nri.UpdatePodSandboxRequest{
+		Pod:                    podNri,
+		OverheadLinuxResources: podNri.Linux.GetPodOverhead(),
+		LinuxResources:         podNri.Linux.GetPodResources(),
+	}
+
+	_, err := l.nri.UpdatePodSandbox(ctx, request)
+
+	return err
+}
```
klihub, line 193: I think we should change this, adding extra arguments for the updated `overhead` and `resources`, then pass those on to NRI in the request. Otherwise NRI gets is the old pod with `nil` updates. So something like this:

```
func (l *local) UpdatePodSandbox(ctx context.Context, pod PodSandbox, overhead, resources *nri.LinuxResources) error {
	if !l.IsEnabled() {
		return nil
	}

	l.Lock()
	defer l.Unlock()

	podNri := podSandboxToNRI(pod)
	request := &nri.UpdatePodSandboxRequest{
		Pod:                    podNri,
		OverheadLinuxResources: overhead,
		LinuxResources:         resources,
	}

	_, err := l.nri.UpdatePodSandbox(ctx, request)

	return err
}
```

internal/nri/sandbox_other.go
klihub, line 18: I would remove this. AFAICT, it is unused now.

server/nri-api.go
```diff
@@ -69,6 +69,16 @@ func (a *nriAPI) runPodSandbox(ctx context.Context, criPod *sandbox.Sandbox) err
...
+func (a *nriAPI) updatePodSandbox(ctx context.Context, criPod *sandbox.Sandbox) error {
+	if !a.isEnabled() {
+		return nil
+	}
+
+	pod := nriPodSandbox(ctx, criPod)
+
+	return a.nri.UpdatePodSandbox(ctx, pod)
+}
+
```
klihub, line 81: I think we need to update this to get extra arguments for the updated pod `overhead` and `resources` and then pass them on to NRI. Something like this:

```
func (a *nriAPI) updatePodSandbox(ctx context.Context, criPod *sandbox.Sandbox, overhead, resources *cri.LinuxContainerResources) error {
	if !a.isEnabled() {
		return nil
	}

	pod := nriPodSandbox(ctx, criPod)

	return a.nri.UpdatePodSandbox(ctx, pod, api.FromCRILinuxResources(overhead), api.FromCRILinuxResources(resources))
}
```

server/sandbox_update.go
klihub, line 23: And this would then need to be

```
...
       err = s.nri.updatePodSandbox(ctx, sb, req.GetOverhead(), req.GetResources())
...
```

klihub (changes requested):
I think we need to update this to dig out the updated new pod overhead and resources from the CRI request and pass them on, along with the (unupdated) pod, to NRI.

Additionally, I think it would be good to clean up the commit history a it before merging. Maybe with a split like this:

- update deps + re-vendor
- add necessary NRI plumbing (new UpdatePodSandox and related functions in internal/nri and server/nri-api.go)
- add new UpdatePodSandboxResources CRI API handler (server/sandbox_update.go)

Also, the added test case should be modified to really update the pod overhead and resources (now they are nil in the test), using different values, and then verify in the test case that we get them correctly on the NRI side.

klihub:
> I think we need to update this to dig out the updated new pod overhead and resources from the CRI request and pass them on, along with the (unupdated) pod, to NRI.
> 
> Additionally, I think it would be good to clean up the commit history a bit before merging. Maybe with a split like this:
> ...

Or then just squash the whole thing to a single commit. Either way, the current commit content split is a bit weird....

test/nri/plugin.go
```diff
@@ -209,6 +209,26 @@ func (p *plugin) RunPodSandbox(_ context.Context, pod *api.PodSandbox) error {
...
 
+func (p *plugin) UpdatePodSandbox(_ context.Context, pod *api.PodSandbox) error {
+	if !p.inNamespace(pod.GetNamespace()) {
+		return nil
+	}
+
+	p.Lock()
+	defer p.Unlock()
+
+	p.pods[pod.GetId()] = pod
```
klihub, line 220: nit: this should not be done here... we should already have a pod (since it has been started and has not been stopped and removed yet). What we instead could do here is look it up and patch in the updated overhead and resources...

test/nri/plugin.go
```diff
@@ -209,6 +209,26 @@ func (p *plugin) RunPodSandbox(_ context.Context, pod *api.PodSandbox) error {
 	return nil
 }
 
+func (p *plugin) UpdatePodSandbox(_ context.Context, pod *api.PodSandbox) error {
```
klihub, line 212: Eventually we should update this to something like below, so we can have more meaningful tests, but it should be perfectly fine to do that later with a subsequent PR.

```
func (p *plugin) UpdatePodSandbox(_ context.Context, pod *api.PodSandbox, overhead, resources *api.LinuxContainerResources) error {
    p.Lock()
    defer p.Unlock()

    // maybe look up pod and patch in the updates 

    // extra new overhead and resources also need to be added to the event struct type...
    p.emitEvent(
        &event{
            kind: "UpdatePodSandbox"
            pod: pod,
            overhead: overhead,
            resources: resources,
        },
    )

    return nil
}
```

test/nri/plugin.go
```diff
@@ -617,6 +637,10 @@ func RunPodEvent(pod string) *event {
 	return PodEvent("RunPodSandbox", pod)
 }
 
+func UpdatePodEvent(pod string) *event {
+	return PodEvent("UpdatePodSandbox", pod)
+}
+
```
klihub, line 643: Eventually we should update this to also record the updated overhead and resources fields, so we can have more meaningful tests.

klihub (approved):
The newly added NRI test bits will need a bit more work to allow for meaningful tests, but I'm fine with addressing this in a subsequent PR.

klihub (approved):

server/sandbox_update_resources.go
```diff
@@ -1,15 +0,0 @@
-package server
```
haircommander, line 1: hm I actually would prefer we didn't do this change so it continues to match the container variant

server/sandbox_update_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+			// When
+			_, err := sut.UpdatePodSandboxResources(context.Background(),
+				&types.UpdatePodSandboxResourcesRequest{PodSandboxId: testSandbox.ID()})
+
+			// Then
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should error with inavailable sandbox", func() {
```
haircommander, line 34: nit: unavailable*

haircommander:
sorry @felipeagger I rebased for you (hoping to merge right after) and then found some nits. I can take care of them but don't want to step on your toes (more than I already have). would you like me to take care of them for you?

felipeagger:
> sorry @felipeagger I rebased for you (hoping to merge right after) and then found some nits. I can take care of them but don't want to step on your toes (more than I already have). would you like me to take care of them for you?

Yes, no problem, go ahead :)

PR #9045, opened 2025-03-06
Improve artifact error logs

saschagrunert:
#### What type of PR is this?

/kind cleanup

We should not need to log an error if the error is `nil` or if an artifact got not found.

None
None

```release-note
None
```


saschagrunert:
@cri-o/cri-o-maintainers PTAL

PR #9050, opened 2025-03-10
Add image volume subpath support

saschagrunert:
#### What type of PR is this?

/kind feature

Add sub path support to image volumes and extend integration tests for that scenario.

None

Refers to https://github.com/kubernetes/kubernetes/pull/130135
Needs https://github.com/kubernetes-sigs/cri-tools/pull/1793

```release-note
Added image volume subpath support for Kubernetes v1.33 (https://github.com/kubernetes/enhancements/issues/4639).
```


server/container_create_linux.go
```diff
@@ -411,6 +414,30 @@ func (s *Server) mountImage(ctx context.Context, specgen *generate.Generator, im
...
+	if err != nil {
+		return nil, fmt.Errorf("secure join final mount path: %w", err)
+	}
+
+	var st unix.Stat_t
+	if err := unix.Stat(mountPoint, &st); err != nil {
+		// This is a user-facing error message from a kubelet event,
+		// means we should make it as meaningful as possible.
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("sub path %q does not exist in image volume %q", m.ImageSubPath, m.Image.Image)
```
giuseppe, line 433: this is a potential security issue.

There is a TOCTOU here, since the source can change until the OCI runtime uses it.

e.g. if the same volume is used by two different containers, a container already having access to the volume can replace the directory with a symlink.

I ended up doing this in Podman: https://github.com/containers/podman/pull/17992

saschagrunert: Hm, I guess we could also use `securejoin.OpenInRoot` for that: https://github.com/cyphar/filepath-securejoin/blob/f277210982c593f9015a66100c67b20c957c24fe/open_linux.go#L43-L50

saschagrunert:
@cri-o/cri-o-maintainers PTAL

server/container_create_linux.go
```diff
@@ -411,6 +414,27 @@ func (s *Server) mountImage(ctx context.Context, specgen *generate.Generator, im
...
 	log.Infof(ctx, "Image mounted to: %s", mountPoint)
 
+	if m.ImageSubPath != "" {
+		// Kubernetes already verifies that ImageSubPath:
+		// - is a relative path
+		// - does not contain any '..'
+		//
+		// We cannot create non-existing paths within the image volume because
+		// they're mounted read-only.
+		mountPointFile, err := securejoin.OpenInRoot(mountPoint, m.ImageSubPath)
```
giuseppe, line 424: when will it be closed?

saschagrunert: Ah, I forgot that good catch. Defer closing right after opening should be fine from my point of view.

server/container_create_linux.go
```diff
@@ -411,6 +414,28 @@ func (s *Server) mountImage(ctx context.Context, specgen *generate.Generator, im
...
+			// This is a user-facing error message from a kubelet event, means we should make it as meaningful as possible.
+			if errors.Is(err, unix.ENOENT) {
+				return nil, fmt.Errorf("sub path %q does not exist in image volume %q", m.ImageSubPath, m.Image.Image)
+			}
+
+			return nil, fmt.Errorf("join final mount path: %w", err)
+		}
+		defer mountPointFile.Close()
+
+		mountPoint = mountPointFile.Name()
```
giuseppe, line 435: this returns the full path to the original file, right?  It is not safe to use, we need the `/proc/self/fd/$FD` path

giuseppe: and the problem is that the fd must be kept open until it is referenced

saschagrunert: @giuseppe can we make the `safeMountSubPath` a helper in c/common or similar?

giuseppe:
@saschagrunert I've simplified the version in Podman to use OpenInRoot: https://github.com/containers/podman/pull/25617

saschagrunert:
> @saschagrunert I've simplified the version in Podman to use OpenInRoot: [containers/podman#25617](https://github.com/containers/podman/pull/25617)

Do we want to make this a helper in c/common or do you prefer having a split-up implementation?

server/container_create.go
```diff
@@ -818,11 +819,31 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr container.Conta
...
+		for _, s := range cleanupSafeMounts {
+			s.Close()
+		}
+	}
+
+	defer func() {
+		if err != nil {
+			cleanupFunc()
+		}
+	}()
```
saschagrunert, line 838: Hm, we may have to do this in scope of `CreateContainer()` and not `createSandboxContainer()`. :thinking:

saschagrunert:
@giuseppe PTAL again.

giuseppe (approved):
LGTM

PR #9051, opened 2025-03-10
Require go 1.24 for build

saschagrunert:
#### What type of PR is this?

/kind cleanup

Kubernetes v1.33 will require that version to be built and we have to align because of the future vendored sources.

None
None

```release-note
Require go 1.24 for build.
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -16,9 +16,8 @@ dependencies:
         match: GO_VERSION
       - path: nix/derivation.nix
         match: buildGo124Module
-      # TODO: move go.mod to 1.24 once we have full prow CI support
-      # - path: go.mod
-      #   match: go
+      - path: go.mod
+        match: go
 
   # For a new minor (1.x.0) release, move the "development version"
   # below to the "supported versions" section (by omitting the patch version)
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-go 1.23.4
+go 1.24
 
 module github.com/cri-o/cri-o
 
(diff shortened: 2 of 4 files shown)


dependencies.yaml
```diff
@@ -16,9 +16,8 @@ dependencies:
         match: GO_VERSION
       - path: nix/derivation.nix
         match: buildGo124Module
-      # TODO: move go.mod to 1.24 once we have full prow CI support
-      # - path: go.mod
-      #   match: go
+      - path: go.mod
```
sohankunkerkar, line 19: Did we confirm if Prow supports Go 1.24?

saschagrunert: Tests succeed, I assume it has 1.24 now.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

dependencies.yaml
kannon92: Openshift did not seem to have go 1.24 work defined yet. It may be worth following up with golang team to see when go 1.24 will be available for openshift/prow.

saschagrunert: We need go 1.24 support in any case because k/k already switched over, which is then also a requirement for https://github.com/cri-o/cri-o/pull/9050. Not exactly sure why we should block this PR on downstream, though.

saschagrunert:
@cri-o/cri-o-maintainers PTAL, this is required for development on k/k features.

sohankunkerkar (approved):
/lgtm

PR #9061, opened 2025-03-14
Cleanup: ensure image volume path

saschagrunert:
#### What type of PR is this?

/kind cleanup

Use the slices package instead of the additional `noop` variable to ensure the image volume path.

None
None

```release-note
None
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 
@@ -438,17 +439,13 @@ func (s *Server) mountImage(ctx context.Context, specgen *generate.Generator, im
 
 func (s *Server) ensureImageVolumesPath(ctx context.Context, mounts []*types.Mount) (string, error) {
 	// Check if we need to anything at all
-	noop := true
-
-	for _, m := range mounts {
-		if m.Image != nil && m.Image.Image != "" {
-			noop = false
-
-			break
+	if !slices.ContainsFunc(mounts, func(m *types.Mount) bool {
+		if m.GetImage() != nil && m.GetImage().GetImage() != "" {
+			return true
 		}
-	}
 
-	if noop {
+		return false
+	}) {
 		return "", nil
 	}
 


server/container_create_linux.go
```diff
@@ -438,17 +439,13 @@ func (s *Server) mountImage(ctx context.Context, specgen *generate.Generator, im
 
 func (s *Server) ensureImageVolumesPath(ctx context.Context, mounts []*types.Mount) (string, error) {
 	// Check if we need to anything at all
-	noop := true
-
-	for _, m := range mounts {
+	if !slices.ContainsFunc(mounts, func(m *types.Mount) bool {
 		if m.Image != nil && m.Image.Image != "" {
```
bitoku, line 443: This is not a part of changes but we can use a getter function instead.
https://protobuf.dev/reference/go/go-generated-opaque/#singular-message

```suggestion
		if m.GetImage().GetImage() != "" {
```

saschagrunert: Sure, but we still have to call `m.GetImage() != nil && m.GetImage().GetImage() != ""` otherwise we'll still `nil` deref.

bitoku: I don't think we have to.
https://play.golang.com/p/3N7gQlUDvT3

https://github.com/cri-o/cri-o/blob/1005e0e3273b42b65eeebe9702bdfc04b7416029/vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go#L756
https://github.com/cri-o/cri-o/blob/1005e0e3273b42b65eeebe9702bdfc04b7416029/vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go#L3609

bitoku: > The GetHeadliner() accessor method is safe to call even if m is nil. This makes it possible to chain get calls without intermediate nil checks:
```
var m *Concert // defaults to nil
log.Infof("GetFoundingYear() = %d (no panic!)", m.GetHeadliner().GetFoundingYear())
```
> If the field is unset, the getter will return the default value of the field. For messages, the default value is a nil pointer.

sohankunkerkar (approved):
/lgtm

Issue #9068, opened 2025-03-20
Mask Linux thermal interrupt info in /proc and /sys


giuseppe:
### What happened?

Thermal interrupt info in /proc and /sys are not masked

### What did you expect to happen?

Thermal interrupt info in /proc and /sys are masked

### How can we reproduce it (as minimally and precisely as possible)?

cat `/proc/interrupts /sys/devices/system/cpu/*/thermal_throttle` is not empty

### Anything else we need to know?

we need the equivalent of https://github.com/moby/moby/pull/49560 for Moby.

The main difficulty is that /sys/devices/system/cpu/<cpuX>/thermal_throttle is not a static value but it depends on the number of cpus on the current system, so we need to make sure it doesn't break the current unmasking mechanism we have in place.

### CRI-O and Kubernetes version

all of them

### OS version

Linux

### Additional environment details (AWS, VirtualBox, physical, etc.)

All of them

PR #9069, opened 2025-03-20
Set default masked paths

saschagrunert:
#### What type of PR is this?

/kind cleanup

 Adding default masked paths to non privileged containers.

Fixes https://github.com/cri-o/cri-o/issues/9068
None

```release-note
Changed runtime default masked paths for non privileged containers to:
- /proc/acpi
- /proc/asound
- /proc/interrupts
- /proc/kcore
- /proc/keys
- /proc/latency_stats
- /proc/sched_debug
- /proc/scsi
- /proc/timer_list
- /proc/timer_stats
- /sys/devices/system/cpu/cpu<ID>/thermal_throttle
- /sys/devices/virtual/powercap
- /sys/firmware
- /sys/fs/selinux
```

### DIFF
--- a/server/masked_paths.go
+++ b/server/masked_paths.go
@@ -1,0 +1,88 @@
+package server
+
+import (
+	"fmt"
+	"os"
+	"runtime"
+	"slices"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/containers/common/pkg/config"
+)
+
+// appendDefaultMaskedPaths is retrieving the default masked paths and appends
+// the existing ones to it.
+func appendDefaultMaskedPaths(additionalPaths []string) []string {
+	paths := slices.Concat(defaultLinuxMaskedPaths(), additionalPaths)
+	slices.Sort(paths)
+
+	return slices.Compact(paths)
+}
+
+// defaultLinuxMaskedPaths will be used to evaluate the default masked paths once.
+var defaultLinuxMaskedPaths = sync.OnceValue(func() []string {
+	maskedPaths := slices.Concat(
+		config.DefaultMaskedPaths,
+		[]string{"/proc/asound", "/proc/interrupts"},
+	)
+
+	for _, cpu := range possibleCPUs() {
+		path := fmt.Sprintf("/sys/devices/system/cpu/cpu%d/thermal_throttle", cpu)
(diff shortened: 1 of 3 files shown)


giuseppe:
could it use the containers/common list for masked paths so we share a bit more code?

saschagrunert:
@cri-o/cri-o-maintainers @giuseppe PTAL

bitoku:
Do we have integration tests for nested container?
I think it breaks `procMount: unmasked` because the proc paths are given by kubelet.
https://github.com/kubernetes/kubernetes/blob/37afe38abf4edf76b0a3fb820e4e963263dda049/pkg/securitycontext/util.go#L217
https://github.com/cri-o/cri-o/pull/6025/commits/4102586132214263c5d0ae93ec257432653ab82b#r911975412

cc @haircommander

bitoku:
We should open a PR for kubelet. https://github.com/kubernetes/kubernetes/blob/37afe38abf4edf76b0a3fb820e4e963263dda049/pkg/securitycontext/util.go#L190-L213

server/masked_paths.go
```diff
@@ -0,0 +1,88 @@
...
+	slices.Sort(paths)
+
+	return slices.Compact(paths)
+}
+
+// defaultLinuxMaskedPaths will be used to evaluate the default masked paths once.
+var defaultLinuxMaskedPaths = sync.OnceValue(func() []string {
+	maskedPaths := slices.Concat(
+		config.DefaultMaskedPaths,
+		[]string{"/proc/asound", "/proc/interrupts"},
```
giuseppe, line 28: can we add these to `"github.com/containers/common/pkg/config"` so we don't need to maintain a separate list?

We can move the entire logic there, even the `possibleCPUs` function if you prefer it to using Glob

giuseppe (approved):
I am not going to block it for the comment I've left, so:

LGTM

saschagrunert:
> We should open a PR for kubelet, maybe? https://github.com/kubernetes/kubernetes/blob/37afe38abf4edf76b0a3fb820e4e963263dda049/pkg/securitycontext/util.go#L190-L213

Good point: https://github.com/kubernetes/kubernetes/pull/131018

/retest

PR #9108, opened 2025-04-07
OCPBUGS-54658: Increase pull-progress-timeout to `30s`

saschagrunert:
#### What type of PR is this?

/kind cleanup

There are use cases where a higher timeout is required because zscaler scans all transfers. To fulfill those we now use a slightly higher default value.

None

Needs to be backported down to 1.29.

```release-note
Increased `pull-progress-timeout` to default to `30s`.
```

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -383,7 +383,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--profile-port**="": Port for the pprof profiler. (default: 6060)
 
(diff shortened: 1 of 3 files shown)


haircommander:
how did we come to this value? I'm unsure we want to do this for this one use case. 10s worked for years before we dropped it then made it configurable, why change now? It feels like the point of having it configurable is these bespoke solutions can set it themselves

saschagrunert:
> how did we come to this value?

I have two examples where they stated that a higher timeout would be beneficial. We could also disable it by default. I thought having a custom configuration is not doable for managed cloud services downstream.

saschagrunert:
@haircommander https://github.com/cri-o/cri-o/pull/9130

PR #9116, opened 2025-04-08
Refactor metrics descriptors

bitoku:
#### What type of PR is this?

/kind cleanup

This PR cleans up the metrics code and removes unavailable metrics (cpuLoad, disk, diskio, misc, and process) which is found in https://kubernetes.slack.com/archives/CAZH62UR1/p1743786431644619?thread_ts=1743403344.323189&cid=CAZH62UR1.

If https://github.com/kubernetes-sigs/cri-tools/pull/1815 is merge, I can write an e2e test for this.

```release-note
None
```

### DIFF
--- a/internal/lib/stats/cpu_metrics.go
+++ b/internal/lib/stats/cpu_metrics.go
@@ -13,11 +13,7 @@ import (
 func generateSandboxCPUMetrics(sb *sandbox.Sandbox, cpu *cgmgr.CPUStats) []*types.Metric {
 	cpuMetrics := []*containerMetric{
 		{
-			desc: &types.MetricDescriptor{
-				Name:      "container_cpu_user_seconds_total",
-				Help:      "Cumulative user CPU time consumed in seconds.",
-				LabelKeys: baseLabelKeys,
-			},
+			desc: containerCpuUserSecondsTotal,
 			valueFunc: func() metricValues {
 				return metricValues{{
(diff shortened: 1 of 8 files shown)


haircommander:
is it impossible to collect these metrics, or we just currently don't? Ideally, I don't want any gap in metrics if we made the switch to CRI, so they should be collected somewhere. I think the ideal would be we collect them here, rather than dropping them. The one case I can see us not doing so is for disk usage which is already being calculated by cadvisor but for the rest it seems we should do it in CRI, no?

bitoku:
@haircommander 
we just currently don't. I'm not sure if we can collect other metrics in cri-o though.
I didn't mean to drop them, but wanted to correct the wrong answer because cri-o doesn't collect disk usages but it answers as if it does.
I think this can be a starting point to implement other metrics without confusion.

I created a tracker issue https://github.com/cri-o/cri-o/issues/9125 

> The one case I can see us not doing so is for disk usage which is already being calculated by cadvisor

We are going to implement disk usage for https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/2371-cri-pod-container-stats/README.md , aren't we?

haircommander:
thanks for creating the tracking issue. we can do this now but we should work on expanding the metrics we collect

/approve

bitoku:
@cri-o/cri-o-maintainers Can you PTAL?

PR #9123, opened 2025-04-10
info: improve AppArmor detection using runc's helper

fmuyassarov:
#### What type of PR is this?
/kind bug
/kind cleanup

When printing the CRI-O version, there is `AppArmorEnabled` that indicates whether AppArmor is enabled or not. However, in case of running it as non-root it reports AppArmor as not enabled even though it is actually enabled on the host. For example, `apparmor_status`  binary confirms that the default CRI-O profile is loaded. 

```
crio version
INFO[2025-03-31T13:00:26.638381533Z] Updating config from single file: /etc/crio/crio.conf
INFO[2025-03-31T13:00:26.638417334Z] Updating config from drop-in file: /etc/crio/crio.conf
INFO[2025-03-31T13:00:26.638433367Z] Skipping not-existing config file "/etc/crio/crio.conf"
INFO[2025-03-31T13:00:26.638455635Z] Updating config from path: /etc/crio/crio.conf.d
INFO[2025-03-31T13:00:26.638506983Z] Updating config from drop-in file: /etc/crio/crio.conf.d/10-crio.conf
Version:        1.32.2
GitCommit:      318db72eb0b3d18c22c995aa7614a13142287296
GitCommitDate:  2025-03-02T18:05:31Z
GitTreeState:   dirty
BuildDate:      1970-01-01T00:00:00Z
GoVersion:      go1.23.3
Compiler:       gc
Platform:       linux/amd64
Linkmode:       static
BuildTags:
  static
  netgo
  osusergo
  exclude_graphdriver_btrfs
  seccomp
  apparmor
  selinux
LDFlags:          unknown
SeccompEnabled:   true
AppArmorEnabled:  false
```
```
sudo apparmor_status | grep crio
   crio-default
```
Currently, CRI-O uses the IsSupported() method to determine whether AppArmor is enabled. However, IsSupported() checks more than just the existence of the AppArmor parser — it also considers whether the environment is rootless and whether runc is enabled. This leads to incorrect reporting in scenarios where AppArmor is enabled on the host but the environment is rootless, causing AppArmor to be shown as disabled.

This patch updates the Info struct to utilize [runc's isEnabled()](https://github.com/opencontainers/runc/blob/159c67f8e2233b6cb84a203dc47ed182381b666c/libcontainer/apparmor/apparmor_linux.go#L18) helper for determining whether AppArmor is enabled. Unlike previous checks, this method directly reads from `/sys/module/apparmor/parameters/enabled`, providing a more reliable indication of the AppArmor status on the host.

```release-note
NONE
```

### DIFF
--- a/internal/config/apparmor/apparmor_linux.go
+++ b/internal/config/apparmor/apparmor_linux.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/containers/common/pkg/apparmor"
+	libctraa "github.com/opencontainers/runc/libcontainer/apparmor"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
@@ -23,7 +24,7 @@ type Config struct {
 // New creates a new default AppArmor configuration instance.
 func New() *Config {
 	return &Config{
-		enabled:        apparmor.IsEnabled(),
+		enabled:        libctraa.IsEnabled(),
 		defaultProfile: DefaultProfile,
 	}
 }
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -13,10 +13,10 @@ import (
 	"text/tabwriter"
 
 	"github.com/blang/semver/v4"
-	"github.com/containers/common/pkg/apparmor"
 	"github.com/containers/common/pkg/seccomp"
 	"github.com/google/renameio"
 	json "github.com/json-iterator/go"
+	libctraa "github.com/opencontainers/runc/libcontainer/apparmor"
 	"github.com/sirupsen/logrus"
 )
 
@@ -206,7 +206,7 @@ func Get(verbose bool) (*Info, error) {
 		BuildTags:       buildTags,
 		LDFlags:         ldFlags,
 		SeccompEnabled:  seccomp.IsEnabled(),
-		AppArmorEnabled: apparmor.IsEnabled(),
+		AppArmorEnabled: libctraa.IsEnabled(),
 		Dependencies:    dependencies,
 	}, nil
 }
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -47,6 +47,8 @@ var _ = t.Describe("Config", func() {
 		return os.Geteuid() != 0
 	}
 
+	expectedError := `unable to load AppArmor profile: installing default AppArmor profile "crio-default" failed`
+
 	t.Describe("ValidateConfig", func() {
 		It("should succeed with default config", func() {
 			// Given
@@ -254,7 +256,9 @@ var _ = t.Describe("Config", func() {
 			err := sut.RuntimeConfig.Validate(nil, true)
 
 			// Then
-			Expect(err).ToNot(HaveOccurred())
+			if err != nil && err.Error() != expectedError {
+				Expect(err).ToNot(HaveOccurred())
+			}
 		})
 
 		It("should succeed with additional devices", func() {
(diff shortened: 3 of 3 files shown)


fmuyassarov:
@saschagrunert @haircommander Can you please take a look at this another attempt?

internal/config/apparmor/apparmor_linux.go
```diff
@@ -23,7 +24,7 @@ type Config struct {
 // New creates a new default AppArmor configuration instance.
 func New() *Config {
 	return &Config{
-		enabled:        apparmor.IsEnabled(),
+		enabled:        runcaa.IsEnabled(),
```
haircommander, line 27: typically we call this libctr, so maybe libctraa?

haircommander:
one nit, otherwise LGTM

internal/config/apparmor/apparmor_linux.go
fmuyassarov: sure, fixed.

haircommander:
/approve

LGTM, thanks!

fmuyassarov:
Hi @saschagrunert , would you mind to give lgtm if looks good to you?

saschagrunert (approved):
/ok-to-test

@fmuyassarov do you mind another rebase to fix the GitHub actions CI?

fmuyassarov:
> /ok-to-test
> 
> @fmuyassarov do you mind another rebase to fix the GitHub actions CI?

Yup, done.

fmuyassarov:
@saschagrunert I wonder if centos based build failure is unrelated to my patch....

saschagrunert (approved):

bitoku:
/hold
You may need to change the unit test, I guess.
https://github.com/cri-o/cri-o/actions/runs/14490212728/job/41147086597?pr=9123

bitoku:
@fmuyassarov Do you want to fix the unit test?

fmuyassarov:
> @fmuyassarov Do you want to fix the unit test?

Hi, yes sure. I was just on vacation. I will fix it today.

bitoku:
@fmuyassarov thanks!

fmuyassarov:
@bitoku can you please take a look now?

bitoku:
Sorry it's a bit late to bring this up, but I think we should be more careful about this change.
In the past, c/common tried to use rootless apparmor, but it's reverted apparently because of the lack of usecases and security concern changing to unconfined.

- https://github.com/containers/common/pull/1572
- https://github.com/containers/common/pull/1620

According to the comment in the issue, https://github.com/containers/common/pull/1572#issuecomment-1644809125 , it can't run apparmor_parser without root privilege.
The cause of the test failure in the rootless unit test may be this.
If so, I guess rootless crio should always fail in apparmor enabled environment.

https://github.com/cri-o/cri-o/blob/c491001e9c9f290359081b6cacb8ad3a51b3f5e4/vendor/github.com/containers/common/pkg/apparmor/apparmor_linux.go#L83-L129

I incline to the way aligning with c/common and not merging this.
To get it merged, we should know why the unit test fails in rootless environment at least.

I'm not an expert of apparmor, so I might be wrong though.

@cri-o/cri-o-maintainers @fmuyassarov wdyt?

fmuyassarov:
> I'm not an expert of apparmor, so I might be wrong though.
> 
> @cri-o/cri-o-maintainers @fmuyassarov wdyt?

I tried to understand the same issue but couldn't get much further beyond the `InstallDefault()` call in the common. I had the same question - why isn't it failing on the main branch?
Since I couldn't pinpoint the root cause, I ended up implementing a temporary workaround by catching the specific error and ignoring it. It's a bit of a hack, but it unblocked me for now. But I can't disagree with your comment @bitoku about being careful when merging this patch, as I'm have an open question. And I can't say much to be honest how to move forward. If you think it is safer to drop this PR rather then creating further problems, I'm all in for that as well.

PR #9140, opened 2025-04-17
Add container_spec_memory_limit_bytes metric

bingikarthik:
#### What type of PR is this?
/kind feature

This PR adds the container_spec_memory_limit_bytes metric to the CRI-O metrics suite, which reports the memory limit for containers in bytes. This metric is valuable for monitoring and alerting on memory configuration versus usage. It allows users to easily calculate memory usage percentages and visualize resource allocation versus consumption in monitoring systems like Prometheus and Grafana.

The implementation follows the existing pattern for metrics in CRI-O. The new metric has been added to the "memory" metrics section and implemented in the generateSandboxMemoryMetrics function to collect memory limit data from cgroup stats.

None

```release-note
Added new metric `container_spec_memory_limit_bytes` to display the memory limit of containers in bytes.
```

### DIFF
--- a/internal/lib/stats/memory_metrics.go
+++ b/internal/lib/stats/memory_metrics.go
@@ -1,6 +1,7 @@
 package statsserver
 
 import (
+	"math"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 
 	"github.com/cri-o/cri-o/internal/config/cgmgr"
@@ -38,6 +39,17 @@ func generateSandboxMemoryMetrics(sb *sandbox.Sandbox, mem *cgmgr.MemoryStats) [
 			desc: containerMemorySwap,
 			valueFunc: func() metricValues {
 				return metricValues{{value: mem.SwapUsage, metricType: types.MetricType_GAUGE}}
+			},
+		},
+		{
+			desc: containerSpecMemoryLimitBytes,
+			valueFunc: func() metricValues {
+				limit := mem.Limit
+				if limit == math.MaxUint64 {
+					// For unlimited memory, use the system's memory limit
+					limit = cgmgr.MemLimitGivenSystem(limit)
+				}
+				return metricValues{{value: limit, metricType: types.MetricType_GAUGE}}
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


bitoku:
@bingikarthik Can you rebase and align it with the new way of writing?

bingikarthik:
> @bingikarthik Can you rebase and align it with the new way of writing?

The PR has been updated now. Could you please review the changes? Thanks..

internal/lib/stats/memory_metrics.go
```diff
@@ -40,6 +41,17 @@ func generateSandboxMemoryMetrics(sb *sandbox.Sandbox, mem *cgmgr.MemoryStats) [
...
 			},
 		},
+		{
+			desc: containerSpecMemoryLimitBytes,
+			valueFunc: func() metricValues {
+				limit := mem.Limit
+				if limit == math.MaxUint64 {
+					// For unlimited memory, use the system's memory limit
+					limit = cgmgr.MemLimitGivenSystem(limit)
+				}
```
bitoku, line 51: For consistency, I think we should follow the way in cAdvisor.
Also in this way, users can't know whether the memory is limited by the system memory size or unlimited (I'm not sure if this difference helps something though).
wdyt?

https://github.com/google/cadvisor/blob/f6e31a3cff918285fd74cb1f20d0af32c3554a68/metrics/prometheus.go#L2000-L2009

bitoku (commented):
Can you run `make lint` to fix the lint errors? I think the errors can be fixed just by running it.

internal/lib/stats/memory_metrics.go
bingikarthik: Agree, we need to align with cAdvisor approach!
So that experience would be pretty much similar..
I have updated the code accordingly, Thanks for the reference link!

bitoku:
Thanks! lgtm if the tests pass

bingikarthik:
> Thanks! lgtm if the tests pass

Could you please review, seems tests passed now except `ci/prow/ci-e2e-evented-pleg` which seems unrelated failure!

bitoku:
/lgtm

@cri-o/cri-o-maintainers PTAL

PR #9147, opened 2025-04-22
Add the option to disable/enable OCI Artifact mount

bitoku:
#### What type of PR is this?

/kind feature

Add toggle option to enable/disable OCI Artifact.

Fixes https://issues.redhat.com/browse/OCPNODE-3156

```release-note
Add option to enable/disable OCI Artifact mount.
```


sohankunkerkar:
It looks like a downstream change for OCP. Is there a reason we aren't adding that knob in the MCO instead?

bitoku:
@sohankunkerkar I'm going to disable OCI Artifact by specifying this option as `false` in MCO.
It'd be better if we can disable the artifact mount just with MCO change, but I can't think of any idea.

sohankunkerkar:
> @sohankunkerkar I'm going to disable OCI Artifact by specifying this option as `false` in MCO. It'd be better if we can disable the artifact mount just with MCO change, but I can't think of any idea.

Ah, you need a knob in CRI-O to do that, so this makes sense.

docs/crio.conf.5.md
```diff
@@ -497,6 +497,10 @@ If true, CRI-O will automatically reload the mirror registry when there is an up
 **pull_progress_timeout**="0s"
 The timeout for an image pull to make progress until the pull operation gets canceled. This value will be also used for calculating the pull progress interval to pull_progress_timeout / 10. Can be set to 0 to disable the timeout as well as the progress output.
 
+**oci_artifact_mount_support**=true
```
haircommander, line 500: instead of a bool toggle, can we see a world where we have a supported mediatype or something? i.e: a more fine grained artifact support where we support some but not others?

bitoku: I was thinking this option is something temporary, and will be deprecated soon when the downstream GA the OCI Artifact.

The idea sounds good, but there are some concerns. 
 - which value should be used for restriction. mediaType in manifest or layer or use something else?
 - How to handle references https://oras.land/docs/concepts/reftypes/
 - maybe something else I don't notice.

I think we have two options. I'd like option 1, but option 2 is also good enough.

1. Keep it as it is, and add a new option for the fine-grained control when it's needed.
2. Make the option value string to make room to extend this option like:
  "": Don't support all artifacts
  "*": Support all artifacts
  and support only those two options for now.
  Once we have enough feedback/information about what kind of control is needed, extend this option.

I'd like to hear opinions from others.
wdyt? @saschagrunert @sohankunkerkar

sohankunkerkar: I agree with @saschagrunert 
IMO, If the long-term goal is to deprecate this option once the OCI artifact feature is GA in OCP, then let’s keep it as a simple boolean to either fully enable it or disable it completely.

bitoku:
@haircommander PTAL again

haircommander:
/approve

thanks!

bitoku:
@saschagrunert @sohankunkerkar PTAL for LGTM

saschagrunert (approved):

Issue #9152, opened 2025-04-24
CRI-O should treat OCI artifacts as artifacts if `artifactType` is set.


saschagrunert:
Right now we publish CRI-O bundles as OCI artifacts, for example:

[ghcr.io/cri-o/bundle@sha256:7970952e8dbacfaa9fce00afba65bc3e3217bffb4b4eb2cc4313816f91d28456](https://explore.ggcr.dev/?image=ghcr.io/cri-o/bundle@sha256:7970952e8dbacfaa9fce00afba65bc3e3217bffb4b4eb2cc4313816f91d28456&mt=application%2Fvnd.oci.image.manifest.v1%2Bjson&size=9788) 

CRI-O treats the artifact as image, because we have the manifest mediaType `application/vnd.oci.image.manifest.v1+json` and the config mediaType `application/vnd.oci.image.config.v1+json`. But we also set

```
"artifactType": "application/vnd.cncf.cri-o.bundle.v1"
```

in the manifest, which should get evaluated as artifact from my point of view. 

@cri-o/cri-o-maintainers do you agree?

bitoku:
I agree. artifactType is used for an artifact according to the spec.

https://github.com/opencontainers/image-spec/blob/c05acf7eb327dae4704a4efe01253a0e60af6b34/image-index.md
> This OPTIONAL property contains the type of an artifact when the manifest is used for an artifact. If defined, the value MUST comply with [RFC 6838](https://tools.ietf.org/html/rfc6838), including the [naming requirements in its section 4.2](https://tools.ietf.org/html/rfc6838#section-4.2), and MAY be registered with [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml).

sohankunkerkar:
Yeah, I second this.

saschagrunert:
Working on a fix in [#9154](https://github.com/cri-o/cri-o/pull/9154)

PR #9154, opened 2025-04-24
Support `artifactType` OCI artifacts

saschagrunert:
#### What type of PR is this?

/kind bug

If the `artifactType` field is set, then we treat the image as OCI artifact. This allows to specify artifacts which still use `application/vnd.oci.image.manifest.v1+json` as manifest media type and/or `application/vnd.oci.image.config.v1+json` config media type.

Fixes https://github.com/cri-o/cri-o/issues/9152#issuecomment-2827357097
None

```release-note
None
```

### DIFF
--- a/internal/ociartifact/store.go
+++ b/internal/ociartifact/store.go
@@ -169,7 +169,18 @@ func (s *Store) PullManifest(
 	}
 
 	if slices.Contains(imageMimeTypes, mimeType) && slices.Contains(configMediaTypes, mediaType) {
-		return nil, ErrIsAnImage
+		ociManifest, err := manifest.OCI1FromManifest(manifestBytes)
+		// Unable to parse an OCI manifest, assume an image
+		if err != nil {
+			return nil, ErrIsAnImage
+		}
+
+		// No artifact type set, assume an image
+		if ociManifest.ArtifactType == "" {
+			return nil, ErrIsAnImage
+		}
+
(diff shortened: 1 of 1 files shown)


internal/ociartifact/store.go
```diff
@@ -168,7 +168,13 @@ func (s *Store) PullManifest(
 		manifest.DockerV2Schema2ConfigMediaType,
 	}
 
-	if slices.Contains(imageMimeTypes, mimeType) && slices.Contains(configMediaTypes, mediaType) {
+	artifactType := ""
+	if ociManifest, err := manifest.OCI1FromManifest(manifestBytes); err == nil {
+		artifactType = ociManifest.ArtifactType
+		log.Debugf(ctx, "Found artifact type: %s", artifactType)
+	}
```
bitoku, line 175: nit: we can check the artifact type only when `slices.Contains(imageMimeTypes, mimeType) && slices.Contains(configMediaTypes, mediaType)` is met, then we can reduce unnecessary manifest parsing.
It will make the additional nest though...

bitoku (commented):
lgtm

internal/ociartifact/store.go
saschagrunert: Yep, fixed as intended.

bitoku:
@saschagrunert  Sorry there was a lint error. Please fix it.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

PR #9172, opened 2025-05-01
[WIP] Add mechanism to monitor conmon

bitoku:
#### What type of PR is this?

/kind feature

This PR adds mechanism to monitor conmon processes, and logs errors when they exited before the container stops.

It monitors conmon by polling with signal 0 periodically.

There's a concern about its performance impact. A goroutine has a minimum stack size 8KB + other usage. If the node runs 1000 containers, cri-o uses more memory usage by 10MB - 100MB order I guess.

```release-note
Log errors when conmon stops unexpectedly
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -339,7 +339,34 @@ func (r *runtimeOCI) CreateContainer(ctx context.Context, c *Container, cgroupPa
 		return err
 	}
 
+	c.state.ContainerMonitorProcess, err = r.getConmonProcess(c)
+	if err != nil {
+		return err
+	}
+
+	if err := r.StartWatchContainerMonitor(ctx, c); err != nil {
+		return err
+	}
+
 	return nil
+}
+
+// getConmonProcess returns the pid and the start time of conmon.
+func (r *runtimeOCI) getConmonProcess(c *Container) (*ContainerMonitorProcess, error) {
+	conmonPid, err := ReadConmonPidFile(c)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read conmon pid: %w", err)
+	}
+
+	startTime, err := getPidStartTime(conmonPid)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get conmon pid start time: %w", err)
+	}
+
+	return &ContainerMonitorProcess{
+		Pid:       conmonPid,
+		StartTime: startTime,
+	}, nil
 }
 
 // StartContainer starts a container.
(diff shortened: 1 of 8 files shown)


internal/oci/runtime_oci.go
```diff
@@ -1725,3 +1752,82 @@ func (r *runtimeOCI) checkpointRestoreSupported(runtimePath string) error {
...
+			log.Errorf(ctx, "Skipping start watch container monitor for container %q: conmon process is already watched", c.ID())
+		}
+
+		c.watched = true
+		defer func() {
+			c.watched = false
+		}()
+
+		// Wait until the conmon process exits
+		state, err := conmonProcess.Wait()
```
haircommander, line 1792: we may want to do some perf scale testing to make sure this scales up with the number of containers. It may be more efficient to open the process and store it and periodically check it rather than have a goroutine open for each conmon

haircommander: I also think if we catch this situation happening we should emit a metric so we can track over time

bitoku: Thanks, don't we have a performance test CI or something, do we?

haircommander: no but we really should...

haircommander: I had hoped we could use pidfds for this behavior as non parent processes can open then and then the kernel will close them when the process dies, so we could asynchronously check whether the fd was closed (I believe this is how it works, it's possible I've got details fuzzy there)

bitoku: Thanks. I was trying to achieve it only with high-level libraries, but it should be possible with epoll or some way to mointor fds. I'll try implementing and compare the efficiency.

bitoku:
https://github.com/cri-o/cri-o/pull/9185
I wrote epoll + pidfd version in another PR.

I tested the performance with 5 containers * 100 pods = 500 containers. The script I used is https://github.com/cri-o/cri-o/pull/9185/files#diff-b99687ee53b96d05bae8378f7364c7999d379d47a742edbfab6d2af3227da9db .


In short, I got these results with ps command

no change: rss 155,695 kb,  cpu 3.99 %
signal polling (this PR): rss 157,030 kb, cpu 4.7 %
epoll (#9185): rss 208,881 kb, cpu 8.79 %

 [details](https://docs.google.com/spreadsheets/d/1s6GZjpxa5BlEfbSeTmGsYabrpQwmoAqx3KohZOH-bN8/edit?usp=sharing) (it's only accessible with redhat account)

The increase in epoll cpu was mainly EpollWait syscall. I'm not sure the cause of the increase in epoll memory, it shows LoadContainer used some memories, which was not shown in the signal polling. ([epoll pprof](https://pprof.me/1cfb60518704a518a056e9ec6bea79b1/?profileType=profile%253Ainuse_space%253Abytes%253Aspace%253Abytes&color_by=filename), [polling pprof](https://pprof.me/85a18884f4323cb9d70cae093b18573c/?profileType=profile%253Ainuse_space%253Abytes%253Aspace%253Abytes&color_by=filename) ) 

IMO the signal polling way is simple and cheap enough, but let me know your thought about this change. Do we want to monitor conmon at the expense of these consumption?
@cri-o/cri-o-maintainers

bitoku:
I did some additional tests with a single goroutine way.

| branch | rss (kb) | cpu (%) |
|--------|--------|--------|
| main | 155,695 | 3.99 |
| goroutine per container + send signal polling (10s) (this PR) | 157,030 | 4.7 |
| epoll + pidfd (#9185) | 208,881  | 8.79 |
| single goroutine + send signal polling (10s) (#9205) | 151,089 | 4.51 | 

[details](https://docs.google.com/spreadsheets/d/1s6GZjpxa5BlEfbSeTmGsYabrpQwmoAqx3KohZOH-bN8/edit?usp=sharing) (it's only accessible with redhat account)

- The memory usage for a goroutine might be negligible compared to the container struct itself. because the main branch memory usage is higher than the single goroutine test.

- I'm not sure if this cpu usage is reliable enough because I tested with the main branch again (though there are some new commits than the binary used in the last test), the average cpu usage was 5.1 %. epoll is high enough, but I'd say less than 5% is within the allowable margin of error.

I'd like to go for "single goroutine + send signal polling" though we can make the interval longer.
Please let me know your thoughts or tests I should do if any.

Issue #9174, opened 2025-05-02
OCI artifacts are not mounted read-only


saschagrunert:
OCI images are mounted read only right now, where artifacts are not. I think this is fine, because we can not remove the actual files, because we get the error:

```
rm: cannot remove '3': Device or resource busy
```

But we can create temporary files within the volume, which are ephemeral until the pod gets evicted. I think we could establish something similar for image volumes as outlined in: https://github.com/kubernetes/kubernetes/issues/131557

saschagrunert:
Working on this as part of https://github.com/cri-o/cri-o/pull/9183

PR #9183, opened 2025-05-06
Make image volumes writable

saschagrunert:
#### What type of PR is this?

/kind feature

We now maintain an additional `upperdir`/`workdir` part of the image volume mount to be able to write to it. The files will be removed on container deletion and will be therefore not persistent.

Fixes https://github.com/cri-o/cri-o/issues/9174

Refers to https://github.com/kubernetes/kubernetes/issues/131557

```release-note
Make image volumes writable: The created files within volumes will be available for each container until its deletion.
```

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -229,7 +229,7 @@ func (s *Server) addOCIBindMounts(ctx context.Context, ctr ctrfactory.Container,
 				log.Debugf(ctx, "Skipping artifact mount because OCI artifact mount support is disabled")
 			}
 
-			volume, safeMount, err := s.mountImage(ctx, specgen, imageVolumesPath, m, runDir)
+			volume, safeMount, err := s.mountImage(ctx, ctr.ID(), specgen, imageVolumesPath, m, runDir)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


server/container_create_linux.go
```diff
@@ -556,19 +556,31 @@ func (s *Server) mountImage(ctx context.Context, specgen *generate.Generator, im
...
+	for _, dir := range []string{upperDirPath, workDirPath} {
+		if err := os.MkdirAll(dir, 0o700); err != nil {
+			return nil, nil, fmt.Errorf("create image volume path: %w", err)
+		}
+	}
+
+	mountOpt := fmt.Sprintf(
+		"lowerdir=%s:%s,upperdir=%s,workdir=%s",
+		mountPoint, imageVolumesPath.mounts, upperDirPath, workDirPath,
+	)
```
saschagrunert, line 571: @giuseppe PTAL from a security standpoint.

giuseppe: thanks for the tag, I don't see anything wrong in this from a security PoV

saschagrunert: Thank you! :pray:

haircommander:
hmm I thought we explictlly decided to do RO image volumes. The solution to me seems like we should mount artifacts RO, not make images RW

PR #9190, opened 2025-05-09
Add option to allow seccomp profiles for privileged containers

saschagrunert:
#### What type of PR is this?

/kind feature

This enables users to apply seccomp profiles to privileged containers.

None
None

```release-note
Added option `--privileged-seccomp-profile` / `privileged_seccomp_profile` to allow specifying a seccomp profile for privileged containers.
```


server/container_create.go
```diff
@@ -1041,7 +1041,7 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr container.Conta
 		specgen.Config.Linux.Seccomp = nil
 	}
 
-	if !ctr.Privileged() {
+	if !ctr.Privileged() || s.config.Seccomp().IsNonBlockingSeccompProfile(ctx, securityContext.Seccomp) {
```
ngopalak-redhat, line 1044: Can the user make the choice using some annotation instead of this code determining if the sec comp profile needs to be applied?
Something like:
```
if !ctr.Privileged() || allowedByannotation("someannotation") {
```

haircommander: I would actually prefer it be on the cri-o config level rather than workload opting in. In the case an admin wants to audit every container's behavior, it shouldn't be opt-out-able by a workload by omitting an annotation. an mutating admission webhook could solve that but that seems more compilcated than just having the cri-o config be the policy decider

saschagrunert: Good, I updated the PR to make it an option.

bitoku:
Does "non blocking seccomp profiles" mean `SCMP_ACT_TRACE`, `SCMP_ACT_LOG` and `SCMP_ACT_NOTIFY` ?
Looking at the e2e test, even blocking ones seems to be applied.

saschagrunert:
> Does "non blocking seccomp profiles" mean `SCMP_ACT_TRACE`, `SCMP_ACT_LOG` and `SCMP_ACT_NOTIFY` ? Looking at the e2e test, even blocking ones seems to be applied.

Yeah the scope of the PR changed. I think this simpler approach by using an option is easier to explain.

bitoku:
@saschagrunert
I have one concern accepting blocking seccomp profiles.
If a user uses both this option and [`--seccomp-default` option](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) in kubelet, I think a privileged container uses RuntimeDefault seccomp and the container might not work properly because some syscalls are blocked.

I don't know if this is a big deal, but to my mind, IMHO accepting only non-blocking seccomp would be more convenient.

<details><summary>CreateContainerRequest payload with `--seccomp-default` option</summary>
<p>

```
cat <<EOF | k apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: test
  labels:
    app: test
spec:
  containers:
  - name: test
    image: quay.io/libpod/busybox
    command: ["sleep", "inf"]
    securityContext:
      privileged: true
EOF
```


DEBU[2025-05-12T11:13:12.781833195Z] Request: &CreateContainerRequest{PodSandboxId:04733f36a9b80865c4f5ff839b8a6e47f8e8458c8079c79af32d9d7283d3828c,Config:&ContainerConfig{Metadata:&ContainerMetadata{Name:test,Attempt:0,},Image:&ImageSpec{Image:quay.io/libpod/busybox@sha256:c9249fdf56138f0d929e2080ae98ee9cb2946f71498fc1484288e6a935b5e5bc,Annotations:map[string]string{},UserSpecifiedImage:quay.io/libpod/busybox,RuntimeHandler:,},Command:[sleep inf],Args:[],WorkingDir:,Envs:[]*KeyValue{&KeyValue{Key:KUBERNETES_SERVICE_PORT,Value:443,},&KeyValue{Key:KUBERNETES_SERVICE_PORT_HTTPS,Value:443,},&KeyValue{Key:KUBERNETES_PORT,Value:tcp://10.0.0.1:443,},&KeyValue{Key:KUBERNETES_PORT_443_TCP,Value:tcp://10.0.0.1:443,},&KeyValue{Key:KUBERNETES_PORT_443_TCP_PROTO,Value:tcp,},&KeyValue{Key:KUBERNETES_PORT_443_TCP_PORT,Value:443,},&KeyValue{Key:KUBERNETES_PORT_443_TCP_ADDR,Value:10.0.0.1,},&KeyValue{Key:KUBERNETES_SERVICE_HOST,Value:10.0.0.1,},},Mounts:[]*Mount{&Mount{ContainerPath:/var/run/secrets/kubernetes.io/serviceaccount,HostPath:/var/lib/kubelet/pods/b10534a0-f9ee-43fe-a26a-6e3cafdfe4a9/volumes/kubernetes.io~projected/kube-api-access-5md4s,Readonly:true,SelinuxRelabel:false,Propagation:PROPAGATION_PRIVATE,UidMappings:[]*IDMapping{},GidMappings:[]*IDMapping{},RecursiveReadOnly:false,Image:nil,ImageSubPath:,},&Mount{ContainerPath:/etc/hosts,HostPath:/var/lib/kubelet/pods/b10534a0-f9ee-43fe-a26a-6e3cafdfe4a9/etc-hosts,Readonly:false,SelinuxRelabel:false,Propagation:PROPAGATION_PRIVATE,UidMappings:[]*IDMapping{},GidMappings:[]*IDMapping{},RecursiveReadOnly:false,Image:nil,ImageSubPath:,},&Mount{ContainerPath:/dev/termination-log,HostPath:/var/lib/kubelet/pods/b10534a0-f9ee-43fe-a26a-6e3cafdfe4a9/containers/test/f5ec487e,Readonly:false,SelinuxRelabel:false,Propagation:PROPAGATION_PRIVATE,UidMappings:[]*IDMapping{},GidMappings:[]*IDMapping{},RecursiveReadOnly:false,Image:nil,ImageSubPath:,},},Devices:[]*Device{},Labels:map[string]string{io.kubernetes.container.name: test,io.kubernetes.pod.name: test,io.kubernetes.pod.namespace: default,io.kubernetes.pod.uid: b10534a0-f9ee-43fe-a26a-6e3cafdfe4a9,},Annotations:map[string]string{io.kubernetes.container.hash: e318b552,io.kubernetes.container.restartCount: 0,io.kubernetes.container.terminationMessagePath: /dev/termination-log,io.kubernetes.container.terminationMessagePolicy: File,io.kubernetes.pod.terminationGracePeriod: 30,},LogPath:test/0.log,Stdin:false,StdinOnce:false,Tty:false,Linux:&LinuxContainerConfig{Resources:&LinuxContainerResources{CpuPeriod:100000,CpuQuota:0,CpuShares:2,MemoryLimitInBytes:0,OomScoreAdj:1000,CpusetCpus:,CpusetMems:,HugepageLimits:[]*HugepageLimit{&HugepageLimit{PageSize:2MB,Limit:0,},&HugepageLimit{PageSize:1GB,Limit:0,},},Unified:map[string]string{memory.oom.group: 1,memory.swap.max: 0,},MemorySwapLimitInBytes:0,},SecurityContext:&LinuxContainerSecurityContext{Capabilities:nil,Privileged:true,NamespaceOptions:&NamespaceOption{Network:POD,Pid:CONTAINER,Ipc:POD,TargetId:,UsernsOptions:&UserNamespace{Mode:NODE,Uids:[]*IDMapping{},Gids:[]*IDMapping{},},},SelinuxOptions:nil,RunAsUser:&Int64Value{Value:0,},RunAsUsername:,ReadonlyRootfs:false,SupplementalGroups:[],ApparmorProfile:,SeccompProfilePath:,NoNewPrivs:false,RunAsGroup:nil,MaskedPaths:[/proc/asound /proc/acpi /proc/kcore /proc/keys /proc/latency_stats /proc/timer_list /proc/timer_stats /proc/sched_debug /proc/scsi /sys/firmware /sys/devices/virtual/powercap],ReadonlyPaths:[/proc/bus /proc/fs /proc/irq /proc/sys /proc/sysrq-trigger],**Seccomp:&SecurityProfile{ProfileType:RuntimeDefault,LocalhostRef:,}**,Apparmor:nil,SupplementalGroupsPolicy:Merge,},},Windows:nil,CDIDevices:[]*CDIDevice{},StopSignal:RUNTIME_DEFAULT,},SandboxConfig:&PodSandboxConfig{Metadata:&PodSandboxMetadata{Name:test,Uid:b10534a0-f9ee-43fe-a26a-6e3cafdfe4a9,Namespace:default,Attempt:0,},Hostname:test,LogDirectory:/var/log/pods/default_test_b10534a0-f9ee-43fe-a26a-6e3cafdfe4a9,DnsConfig:&DNSConfig{Servers:[10.0.0.10],Searches:[default.svc.cluster.local svc.cluster.local cluster.local local europe-west2-c.c.openshift-gce-devel.internal c.openshift-gce-devel.internal google.internal],Options:[ndots:5],},PortMappings:[]*PortMapping{},Labels:map[string]string{app: test,io.kubernetes.pod.name: test,io.kubernetes.pod.namespace: default,io.kubernetes.pod.uid: b10534a0-f9ee-43fe-a26a-6e3cafdfe4a9,},Annotations:map[string]string{kubectl.kubernetes.io/last-applied-configuration: {"apiVersion":"v1","kind":"Pod","metadata":{"annotations":{},"labels":{"app":"test"},"name":"test","namespace":"default"},"spec":{"containers":[{"command":["sleep","inf"],"image":"quay.io/libpod/busybox","name":"test","securityContext":{"privileged":true}}]}}

</p>
</details>

haircommander:
maybe instead we have the flag be `--privileged-seccomp-profile` and point to a path with a different runtime default for prvileged containers?

saschagrunert:
`oc debug` containers also have the `debug.openshift.io/source-container` and `debug.openshift.io/source-resource` annotations set.

haircommander:
> How would we ship that profile? Per default within the packages?

default should be "", up to downstream users to set one. SPO could create one if auditlogging was setup with a drop-in.

pkg/config/template.go
```diff
@@ -937,6 +942,12 @@ const templateStringCrioRuntimeSeccompProfile = `# Path to the seccomp.json prof
 
 `
 
+const templateStringCrioRuntimePrivilegedSeccompProfile = `# Enable seccomp profiles for privileged containers.
+# This option supports live configuration reload.
+{{ $.Comment }}privileged_seccomp_profile = "{{ .PrivilegedSeccompProfile }}"
```
bitoku, line 947: we may want to mention that the value is considered as LocalhostRef.

...No one wants to set RuntimeDefault for privileged containers, do they?

saschagrunert: I don't think it's actually a required use case to use runtime default for privileged containers. But we could still add it later on if we want. Having it only a local path makes it much clearer from my point of view compared to parsing additional pseudo values like `RuntimeDefault` or such. 

I updated the docs to mention that it takes a file path.

bitoku:
/lgtm
Thanks!

PR #9216, opened 2025-05-21
Avoid accessing mirrored registry

bitoku:
#### What type of PR is this?

/kind bug

This PR fixes the issue that it tries to access registries which are mirrored in registries.conf.

Fixes https://issues.redhat.com/browse/OCPBUGS-56561

```release-note
Fixed the bug that cri-o tried to access to registries which are mirrored in registries.conf[.d]
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -21,6 +21,7 @@ import (
 	cimage "github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/shortnames"
+	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/signature"
 	istorage "github.com/containers/image/v5/storage"
 	"github.com/containers/image/v5/transports"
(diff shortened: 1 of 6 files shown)


bitoku:
@cri-o/cri-o-maintainers PTAL

mtrmac:
I don’t know what to recommend here.

Because (AFAIK) this is _completely impossible_ to implement reliably for IP addresses, and very inconvenient to do at this level for host names, the best I can suggest is to make the whole `isSecureIndex` logic deprecated.
- Deprecate IP address matching _without replacement_.
- For DNS name matching, `registries.conf` allows configuring that; it correctly works on individual mirrors; _and_ in OpenShift it is already user-controlled via `machine-config-operator`.
- Deprecate the CRI-O CLI and config options
- For the hard-coded default of `127.0.0.0/8`, make that a config file option; that would let the _current_ customer to disable the lookups. And announce that the config file option will flip to not making this matching in the future.

bitoku:
I found you suggested the same 6 years ago https://github.com/containers/image/pull/539#issuecomment-490261472 .

I'm +1 to your idea, but it can cause disruption for the current users using insecureRegistries, so we need to discreet about it though I can't think of any idea to detour this check without disruption.

mtrmac:
If we knew this was a good idea 6 years ago, we could have deprecated it 6 years ago and it could have been removed by now. So, shall we do the next best thing and deprecate it now, so that we don’t have the same dilemma in 2031?

PR #9223, opened 2025-05-26
Make irq-load-balancing.crio.io configurable per CPU

andreaskaris:
For more details, see https://issues.redhat.com/browse/RFE-7465 | https://issues.redhat.com/browse/TELCOSTRAT-318

#### What type of PR is this?

/kind feature

This commit introduces new value `housekeeping` for annotation `irq-load-balancing.crio.io`.
When `housekeeping` is set:
* The housekeeping CPU set is injected into the container's environment variables as OPENSHIFT_HOUSEKEEPING_CPUS.
* IRQ SMP affinity bits are not disabled on the housekeeping CPUs when adding a new container.
The housekeeping CPUs are chosen as the first CPU inside each container plus its thread siblings.

Reason for this change:
*   **Customer requirements**: Customers have requested the ability to manage IRQs more precisely on the CPUs allocated to their guaranteed pods, particularly in high-performance and latency-sensitive environments.
*   **Improved resource utilization**: As pod density increases, the need for dedicated CPUs for IRQ handling as well as for guaranteed pods without IRQs increases the overall resource footprint of the entire solution. This enhancement helps mitigate the footprint gap compared to competing container orchestration platforms by allowing IRQs on a small subset (e.g., 1–2 CPUs) within a pinned container.
*   **Greater application control**: By allowing targeted IRQ management within a container's CPU set, customers gain more control over  performance tuning and workload characteristics.

This enhancement introduces the ability to configure the `irq-load-balancing.crio.io` annotation to use a new "housekeeping" mode  within CRI-O. When set to "housekeeping", it preserves IRQs on the first CPU and its thread siblings in side each container.

It is acknowledged that CPUs where IRQs are not disabled will handle IRQs for the entire system. Customers have reviewed this behavior and confirmed it is acceptable. The added flexibility and improved efficiency are seen as a worthwhile trade-off.

```release-note
This commit introduces a new `housekeeping` value for the `irq-load-balancing.crio.io` annotation.

When `housekeeping` is set:
* The housekeeping CPU set is injected into the container's environment variables as `OPENSHIFT_HOUSEKEEPING_CPUS`
* IRQ SMP affinity bits are not disabled on the housekeeping CPUs when adding a new container
* The housekeeping CPUs are chosen as the first CPU within each container plus its thread siblings
```

### Smoketest and smoke test environment (single thread per core)

smoketest.sh
```
#!/bin/bash

set -eux

cleanup() {
    set +e
    set +u
    echo "Running cleanup..."
    kubectl delete pod qos-demo
    if [ "$crio_pid" != "" ]; then
    	kill $crio_pid
    fi
    sleep 10
    killall crio || true
    sleep 5
    systemctl restart crio
    wait_cluster_ready

    set +x
    echo "======================"
    echo "For logs, see $tmp_log"
    echo "======================"
}
trap cleanup EXIT

# Adjust this to your cluster, in my case, I need 7 pods with READY 1/1 to be up for the cluster
# to be o.k.
wait_cluster_ready() {
    for i in {0..9}; do
         echo "Waiting for cluster ready, iteration $i"
         sleep 15
	 out=$(timeout 10 kubectl get pods -A | grep "1/1" | wc -l | tr -d '\n')
         if [ "${out}" == "7" ]; then
             return
         else
             echo "Cluster not ready: $out"
	 fi
    done
    exit 1
}

generate_pod_yaml() {
    i=$1
    f="$(mktemp)"
    envsubst < ${DIR}/${pods_yaml[$i]} > $f
    echo $f
}

get_events() {
    i=$1
    sleep 2
    for c in {0..9}; do
        if ! kubectl get events | grep -q "${expected_event[$i]}"; then
            echo "Couldn't find expected event, expected: ${expected_event[$i]}"
        else
            kubectl get events | grep "${expected_event[$i]}"
            return
        fi
        sleep 30
    done
    echo "Couldn't find events after $((i + 1)) tries"
    exit 1
}


DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )

declare -A annotations
annotations[0]='annotations:
   irq-load-balancing.crio.io: "housekeeping"'

annotations[1]='annotations:
    irq-load-balancing.crio.io: "disable"'

annotations[2]='annotations:
    irq-load-balancing.crio.io: "true"'

annotations[3]='annotations:
    irq-load-balancing.crio.io: "garbage"'

annotations[4]='annotations:
    irq-load-balancing.crio.io: ""'

declare -A expected_smp_affinity
expected_smp_affinity[0]="3d1f"
expected_smp_affinity[1]="3c0f"
expected_smp_affinity[2]="3c0f"
expected_smp_affinity[3]="3fff"
expected_smp_affinity[4]="3fff"

declare -A expected_event
expected_event[0]=""
expected_event[1]=""
expected_event[2]=""
expected_event[3]=""
expected_event[4]=""

declare -A pods_yaml
pods_yaml[0]="pods.yaml.j2"
pods_yaml[1]="pods.yaml.j2"
pods_yaml[2]="pods.yaml.j2"
pods_yaml[3]="pods.yaml.j2"
pods_yaml[4]="pods.yaml.j2"

CRIO="/home/akaris/development/cri-o/bin/crio"
tmp_log="$(mktemp)"

systemctl stop crio
sleep 5
killall crio || true
sleep 5
# Start from clean sheet for SMP affinity ...
# Start custom crio ...
${CRIO} 2>&1 | tee "${tmp_log}" | grep smp &
crio_pid=$!
echo "CRIO PID: ${crio_pid}"

wait_cluster_ready

mask=$(cat ${affinity_file} | tr -d '\n')
echo "Starting SMP affinity mask: $mask"
set +x
echo ""
echo ""
echo ""
set -x
for j in {0..1}; do
for i in "${!annotations[@]}"; do
        kubectl delete events --all --timeout=10s
	set +x
	echo "======================="
	echo "Running smoke test ${j}|${i}"
	echo "======================="
	set -x
	export annotation="${annotations[$i]}"
        echo -e "$annotation"
	timeout 15 kubectl apply -f $(generate_pod_yaml $i)
	if [ "${expected_event[$i]}" == "" ]; then
	    kubectl wait --for=condition=Ready pod/qos-demo --timeout=120s
	else
	    kubectl wait --for=condition=Ready pod/qos-demo --timeout=120s || true
	fi
	mask=$(cat ${affinity_file} | tr -d '\n')
	expected_mask="${expected_smp_affinity[$i]}"
	echo "Got mask: $mask, expected mask: $expected_mask"
	if [ "${expected_event[$i]}" == "" ]; then
	    set +x
	    echo "=== HOUSEKEEPING ENV VARS IN CONTAINERS: ==="
	    for cname in qos-demo-ctr qos-demo-ctr-2 qos-demo-ctr-3; do
	       echo "Container $cname"
	       kubectl exec -c $cname qos-demo -- env | grep -i house || true
            done
	    set -x
	    if [ "${mask}" != "${expected_mask}" ]; then
	        exit 1
	    fi
	else
	    get_events $i
	fi
	kubectl delete pod qos-demo
	mask=$(cat ${affinity_file} | tr -d '\n')
	echo "After reset --- Got mask: $mask, expected mask: $expected_reset_affinity"
	if [ "${mask}" != "${expected_reset_affinity}" ]; then
	    echo "WARN - NOT RESET CORRECTLY BUT MAY BE DUE TO RACE"
	    # exit 1
	fi
	set +x
	echo ""
	echo ""
	echo ""
	echo ""
	set -x
done
done
```

pods.yaml.j2
```
apiVersion: v1
kind: Pod
metadata:
  name: qos-demo
  $annotation
spec:
  hostNetwork: true
  containers:
  - name: qos-demo-ctr
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "1Gi"
        cpu: "4"
      requests:
        memory: "1Gi"
        cpu: "4"
  - name: qos-demo-ctr-2
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "1Gi"
        cpu: "2"
      requests:
        memory: "1Gi"
        cpu: "2"
  - name: qos-demo-ctr-3
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "1Gi"
        cpu: "1200m"
      requests:
        memory: "1Gi"
        cpu: "1200m"
```

The virtual system that I'm running the smoke test on has:
* 14 vCPUs (14 cores, no siblings)

Kubelet config:
```
[root@centos9 pod]# grep -i full /var/lib/kubelet/config.yaml -C4
syncFrequency: 0s
volumeStatsAggPeriod: 0s
cpuManagerPolicy: static
cpuManagerPolicyOptions:
  full-pcpus-only: "true"
cpuManagerReconcilePeriod: 5s
reservedSystemCPUs: 0-3
```

crio config:
```
# cat /etc/crio/crio.conf.d/99-runtimes.conf 
[crio.runtime]
infra_ctr_cpuset = "0-3"




# The CRI-O will check the allowed_annotations under the runtime handler and apply high-performance hooks when one of
# high-performance annotations presents under it.
# We should provide the runtime_path because we need to inform that we want to re-use runc binary and we
# do not have high-performance binary under the $PATH that will point to it.
[crio.runtime.runtimes.high-performance]
inherit_default_runtime = true
allowed_annotations = ["cpu-load-balancing.crio.io", "cpu-quota.crio.io", "irq-load-balancing.crio.io", "cpu-c-states.crio.io", "cpu-freq-governor.crio.io"]
```

### Smoketest and smoke test environment (2 thread siblings)

smoketest.sh
```
expected_smp_affinity[0]="3f3f"
```

Everything else the same, 7 cores, 2 thread siblings (0-1, 2-3, and so on) on my test VM.

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -1270,6 +1270,10 @@ func isCgroupParentBestEffort(s *sandbox.Sandbox) bool {
 }
 
 func isContainerRequestWholeCPU(cSpec *specs.Spec) bool {
+	if isContainerCPUEmpty(cSpec) || cSpec.Linux.Resources.CPU.Shares == nil {
+		return false
+	}
+
 	return *(cSpec.Linux.Resources.CPU.Shares)%1024 == 0
 }
 


internal/runtimehandlerhooks/utils_linux.go
```diff
@@ -103,14 +104,37 @@ func isAllBitSet(in []byte) bool {
...
+	if ignoreCPUs != nil {
+		// ignoreList is sorted. The last, highest element must be smaller than len(podcpuset).
+		ignoreList := ignoreCPUs.List()
+		podcpuList := podcpuset.List()
+		ignoreListMax := ignoreList[len(ignoreList)-1]
+		if ignoreListMax >= len(podcpuList) {
+			return cpus, "", fmt.Errorf("highest irq-load-balancing CPU (%d) is larger than highest permissible "+
+				"relative container CPU (%d)",
+				ignoreListMax, len(podcpuList)-1)
+		}
```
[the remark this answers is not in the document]
andreaskaris: unfortunately, not. The ignore list is a relative list  (e.g. 0, 1, 2, 3) whereas podcpuList is an absolute list (e.g. 10, 11, 12, 13, 14, 15), hence the first isn't a subset of the other. It's just strictly smaller than the other. Because the relative list is a filtered list, its highest entry cannot be higher than the length of the absolute list

internal/runtimehandlerhooks/utils_linux.go
```diff
@@ -103,14 +104,37 @@ func isAllBitSet(in []byte) bool {
...
+		var newCPUSet []int
+		for i, cpu := range podcpuList {
+			if !slices.Contains(ignoreList, i) {
+				newCPUSet = append(newCPUSet, cpu)
+			}
+		}
```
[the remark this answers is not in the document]
andreaskaris: Same thing here - the absolute list could be e.g. `10, 11, 12, 13, 14, 15` whereas the relative list would be `0, 1, 2, 3`. Building the difference between the 2 sets would hence make no sense.
The relative list actually represents the indices that shall not be changed.

Hence, one has to:
10, 11, 12, 13, 14, 15  -- transform to -->  0, 1, 2, 3, 4, 5
diff ( [0, 1, 2, 3], [0, 1, 2, 3, 4, 5]) = 4, 5  -- tranform back to -->  14, 15

Which is conceptually what the loop's doing

pkg/annotations/annotations.go
```diff
@@ -29,7 +29,11 @@ const (
 	// CPUQuotaAnnotation indicates that CPU quota should be disabled for CPUs used by the container.
 	CPUQuotaAnnotation = "cpu-quota.crio.io"
 
-	// IRQLoadBalancingAnnotation indicates that IRQ load balancing should be disabled for CPUs used by the container.
+	// IRQLoadBalancingAnnotation indicates IRQ load balancing behavior for CPUs used by the containers.
+	// "disable" disables IRQ load balancing on all CPUs of all containers.
+	// A valid JSON string in the format {<container name>: <cpuset>, …} can be specified. This triggers an allow-list
+	// behavior where IRQ load balancing will be enabled for all matching containers and CPUs but disabled for all other
+	// container CPUs of the containers which have dedicated CPUs.
```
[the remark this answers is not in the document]
andreaskaris: > It lacks the explanation that the given "cpuset" is indices of actual cpuset.

Done, I think it's better now, I hope that it explains everything.

> Aside from that, technically cpuset represents absolute cpu numbers, not relative cpu numbers, so the value here shouldn't be handled with cpuset library (I might be just too particular about this).

I'm using the cpuset.CPUSet library only to parse the indices from a human readable, standardized list format into a list of integer. I moved the parsing logic now into a single location where it's pretty contained and should avoid confusion:
https://github.com/cri-o/cri-o/pull/9223/commits/674c9a6cb2161014a16b0f4ba21b72a198046d7f#diff-d2465500f759904efacd09ab0709903d43712b46422324942f3f8ed092382217R333

> I don't understand the usecase but why do we choose to use index? Is it that convenient than specifying an actual cpuset?

Because the actual CPUset is decided at runtime. An administrator only knows that the container has x amount of static CPUs. So they can use CPU indices for granular control, and these indices must then be mapped to absolute CPU numbers once the container is up.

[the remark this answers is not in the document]
bitoku: > An administrator only knows that the container has x amount of static CPUs. So they can use CPU indices for granular control, and these indices must then be mapped to absolute CPU numbers once the container is up.

If they don't know the absolute cpu numbers, why can they choose correct relative cpu indices?
Let's use one of the testcases as example. In the testcase, the container is assigned `4,5,10,11,12,13,14` and the user specify `0,2-4`, but how do they know they want to specify `0,2-4`? Wouldn't specifying `0-3` (which is in total same 4 cpus) suffice the usecase?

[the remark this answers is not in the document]
andreaskaris: > the container is assigned 4,5,10,11,12,13,14 and the user specify 0,2-4, but how do they know they want to specify 0,2-4 though they don't know which cpus are assigned?

0,2-4 are indices.

You know that the container will have 7 CPUs, so what you want to achieve (looking at indices only):

```
IRQs        0,  2,3,4,
non-IRQs      1,      5,6
```

At runtime, this maps to enabling IRQs on:
```
indices       0, 1, 2, 3, 4, 5, 6
actual CPUs   4, 5,10,11,12,13,14
---------------------------------
IRQ indices   0,    2, 3, 4
IRQ CPUs      4    10,11,12
```

> Wouldn't specifying 0-3 (which is in total same 4 cpus) suffice the usecase?
Yes, but what if the application developer need to handle house-keeping on the last CPUs of the container, and not on the first ones? And what if a year from now, another customer has a more isoteric use case where they need it for some reason in the middle of the container?

MarSik:
Andreas proposes

irq-load-balancing.crio.io: "{<container name>: <relative CPU indices>, ...}"

But it might be better to implement it using something like this I think

housekeeping-cpus.crio.io/[container-name]: 0,1
irq-load-balancing.crio.io: housekeeping

cri-o would have to automatically determine the housekeeping cpus' siblings too
and inject an environment variable to the containers with the list translated
to machine specific CPU IDs.

This would allow easier reuse of the cpu list with other features like "oc exec" cpus.

andreaskaris:
I'll change this PR according to Martin's suggestion

andreaskaris:
@MarSik @bitoku I implemented Martin's suggestions +/-. Let me know what you think. The PR description on the top of this page is up to date with explanations and my smoke test.

bitoku:
Why don't we implement just housekeeping (not housekeeping-full-pcpus-only) and keep this PR simple?
We can add housekeeping-full-pcpus-only later if it's really needed. wdyt?

I'm not sure about the usecase of housekeeping-full-pcpus-only.
In what case do users want to use housekeeping-full-pcpus-only and siblings env var?
If they want to avoid siblings, can't they specify them in housekeeping cpuset?

MarSik:
I need to find time to review the current state of the patch, but in general:

> I'm not sure about the usecase of housekeeping-full-pcpus-only. In what case do users want to use housekeeping-full-pcpus-only and siblings env var? If they want to avoid siblings, can't they specify them in housekeeping cpuset?

They can't, because the pod specification does not know which node will host the Pod and whether it has hyperthreading enabled or not.

Remember the patch introduces a concept of logical Pod cpus that are translated to hardware ids on the destination node only.

andreaskaris:
The unit tests conflict now with: 
```
akaris@workstation cri-o ((17c08c84f...))]$ git show 4dd7fb914
commit 4dd7fb914584619a69324acff0c2ee4145a69ba2
Author: Ayato Tokubi <atokubi@redhat.com>
Date:   Wed Jun 25 10:26:54 2025 +0000

    Set ExecCPUAffinity
```

So I moved the unit tests into their own standalone commits. 

@MarSik any chance to review this patch (in principle) before I rewrite/rebase the unit tests?

andreaskaris:
Let's work in this order:
* review and merge https://github.com/cri-o/cri-o/pull/9364
* then rebase

Then consider adding another commit on top of:
* https://github.com/cri-o/cri-o/pull/9286/commits/4dd7fb914584619a69324acff0c2ee4145a69ba2
To have "exec_cpu_affinity: housekeeping" but it must work even when the housekeeping annotation is missing

/hold

andreaskaris:
@MarSik This is ready for review

internal/runtimehandlerhooks/utils_test.go
```diff
@@ -123,6 +124,15 @@ func writeTempFile(content string) (string, error) {
 	return f.Name(), nil
 }
 
+func cpuSetOrDie(cpus string) cpuset.CPUSet {
+	set, err := cpuset.Parse(cpus)
+	if err != nil {
+		panic(err)
```
MarSik, line 130: This is dangerous. We do not want to kill the whole cri-o in case something is read incorrectly.

MarSik: In tests it probably works fine though.

MarSik:
I like it.

internal/runtimehandlerhooks/utils_test.go
andreaskaris: Yeah, it's just to use with tests. It's really just to make it easier to use in line 30 as an in-place replacement for c.input.cpus; and in line 1061 of the high perf test file

If it's a strong objection, I can rewrite it to use ginkgo and/or to return the error and handle it outside

andreaskaris:
@bitoku ready for review :-)

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1341,11 +1384,14 @@ func setSharedCPUs(c *oci.Container, containerManagers []cgroups.Manager, shared
...
+func isContainerCPUEmpty(spec *specs.Spec) bool {
+	return spec == nil || spec.Linux == nil ||
 		spec.Linux.Resources == nil ||
-		spec.Linux.Resources.CPU == nil ||
-		spec.Linux.Resources.CPU.Cpus == ""
+		spec.Linux.Resources.CPU == nil
+}
+
+func isContainerCPUsSpecEmpty(spec *specs.Spec) bool {
+	return isContainerCPUEmpty(spec) || spec.Linux.Resources.CPU.Cpus == ""
```
bitoku: non-blocking: I would create a function to get `spec.Linux.Resources.CPU.Cpus` and compare it in the caller function.

```go
func getContainerCPU(spec *specs.Spec) string {
	if spec == nil || spec.Linux == nil || spec.Linux.Resources == nil || spec.Linux.Resources.CPU == nil {
		return ""
	}
	return spec.Linux.Resources.CPU.Cpus
}
```

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1472,3 +1518,97 @@ func injectCpusetEnv(specgen *generate.Generator, isolated, shared *cpuset.CPUSe
...
+	cpus := containerSpec.Linux.Resources.CPU.Cpus
+
+	set, err := cpuset.Parse(cpus)
+	if err != nil {
+		return cpuset.CPUSet{}, err
+	}
+
+	if set.IsEmpty() {
+		return cpuset.CPUSet{}, fmt.Errorf("invalid list of cpus provided, set is empty %q", cpus)
+	}
```
bitoku: I wonder, when `Parse` didn't return any error and the cpuset is empty, is it invalid?
It may be just an empty cpuset.

I think we can move `if isContainerCPUsSpecEmpty(containerSpec)` branch after `Parse` and remove this `if set.IsEmpty()` branch.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1472,3 +1518,97 @@ func injectCpusetEnv(specgen *generate.Generator, isolated, shared *cpuset.CPUSe
...
+// This returns the CPUs that should be isolated from interrupt handling.
+func excludeHousekeepingCPUs(cpus string, housekeepingSiblings cpuset.CPUSet) (cpuset.CPUSet, error) {
+	set, err := cpuset.Parse(cpus)
+	if err != nil {
+		return cpuset.CPUSet{}, err
+	}
+
+	if set.IsEmpty() {
+		return cpuset.CPUSet{}, fmt.Errorf("invalid list of cpus provided, set is empty %q", cpus)
+	}
```
bitoku: same as https://github.com/cri-o/cri-o/pull/9223/files#r2398245822

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1472,3 +1518,97 @@ func injectCpusetEnv(specgen *generate.Generator, isolated, shared *cpuset.CPUSe
...
+	if err != nil {
+		// If the file doesn't exist, this CPU has no siblings (or is not hyperthreaded).
+		return siblings, nil
+	}
+
+	siblingsStr := strings.TrimSpace(string(content))
+	if siblingsStr == "" {
+		// If the file is empty, this CPU has no siblings (or is not hyperthreaded).
+		return siblings, nil
+	}
```
bitoku: This should be covered by cpuset.Parse.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1472,3 +1518,97 @@ func injectCpusetEnv(specgen *generate.Generator, isolated, shared *cpuset.CPUSe
...
+	if specgen == nil || specgen.Config == nil || specgen.Config.Process == nil {
+		return fmt.Errorf("specgen or one of its required elements is nil, specgen: %v", specgen)
+	}
+
+	specgen.Config.Process.Env = append(specgen.Config.Process.Env,
+		fmt.Sprintf("%s=%s", HousekeepingCPUsEnvVar, housekeeping.String()),
+	)
+
+	return nil
+}
```
bitoku: We can replace this function to https://pkg.go.dev/github.com/opencontainers/runtime-tools/generate#Generator.AddProcessEnv

andreaskaris: When I add:
```
	specgen.AddProcessEnv(HousekeepingCPUsEnvVar, housekeeping.String())
```

That yields:
```
$ go test -v . -ginkgo.v -ginkgo.focus='.*setIRQLoadBalancing with housekeeping CPUs.*'
(...)
  [PANICKED] Test Panicked
  In [It] at: /usr/local/go/src/internal/runtime/maps/runtime_faststr_swiss.go:265 @ 10/03/25 14:04:38.558

  assignment to entry in nil map

  Full Stack Trace
    github.com/opencontainers/runtime-tools/generate.(*Generator).addEnv(...)
        /home/akaris/development/cri-o/vendor/github.com/opencontainers/runtime-tools/generate/generate.go:543
    github.com/opencontainers/runtime-tools/generate.(*Generator).AddProcessEnv(0xc0000117a0, {0x2b88dbd, 0x11}, {0xc0003c6014, 0x3})
        /home/akaris/development/cri-o/vendor/github.com/opencontainers/runtime-tools/generate/generate.go:519 +0x2f2
    github.com/cri-o/cri-o/internal/runtimehandlerhooks.injectHousekeepingEnv(0xc0000117a0, {0x2f070b8?})
        /home/akaris/development/cri-o/internal/runtimehandlerhooks/high_performance_hooks_linux.go:1576 +0x1a9
    github.com/cri-o/cri-o/internal/runtimehandlerhooks.init.func1.6.4(0xc0002edd40, 0xc0001d0e00, 0x0, {0x2b88d35, 0x11}, {0x2b88d46, 0x11}, {0x2b6d115, 0x3}, 0x0)
        /home/akaris/development/cri-o/internal/runtimehandlerhooks/high_performance_hooks_test.go:392 +0x5ed
    github.com/cri-o/cri-o/internal/runtimehandlerhooks.init.func1.6.6.5()
        /home/akaris/development/cri-o/internal/runtimehandlerhooks/high_performance_hooks_test.go:448 +0xe5
```

Because g.envMap is nil:
```
func (g *Generator) addEnv(env, key string) {
	if idx, ok := g.envMap[key]; ok {
		// The ENV exists in the cache, so change its value in g.Config.Process.Env
		g.Config.Process.Env[idx] = env
	} else {
		// else the env doesn't exist, so add it and add it's index to g.envMap
		g.Config.Process.Env = append(g.Config.Process.Env, env)
		g.envMap[key] = len(g.Config.Process.Env) - 1
	}
}
```

In the same file, we already have this function here which I based injectHouseKeepingEnv on:
```
func injectCpusetEnv(specgen *generate.Generator, isolated, shared *cpuset.CPUSet) {
	spec := specgen.Config
	spec.Process.Env = append(spec.Process.Env,
		fmt.Sprintf("%s=%s", IsolatedCPUsEnvVar, isolated.String()),
		fmt.Sprintf("%s=%s", SharedCPUsEnvVar, shared.String()))
}
```

Given the failure in CI (may be a false positive and I might have to somehow initialize the envmap, but dunno), I'd prefer to not change it

bitoku: Other fields are normally initialized, so I expected Env also to be initialized. It sounds like a bug in runtime-tools, and definitely not a blocker of this PR.

bitoku:
/lgtm
if tests pass

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1472,3 +1520,91 @@ func injectCpusetEnv(specgen *generate.Generator, isolated, shared *cpuset.CPUSe
...
+
+	return set.Difference(housekeepingSiblings), nil
+}
+
+// injectHousekeepingEnv adds the HOUSEKEEPING_CPUS environment variable to the container.
+// This allows the container to be aware of which CPUs are designated for housekeeping tasks.
+func injectHousekeepingEnv(specgen *generate.Generator, housekeeping cpuset.CPUSet) error {
+	if specgen == nil || specgen.Config == nil || specgen.Config.Process == nil {
+		return fmt.Errorf("specgen or one of its required elements is nil, specgen: %v", specgen)
+	}
```
bitoku: sorry one thing I missed.

specgen's fields can be nil, in that case we should create its instance.
See https://github.com/opencontainers/runtime-tools/blob/master/generate/config.go#L13 for example.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: After a second look ...

In theory, in order to make this work, I have to simply change the initialization in the unit test:
```
			specgen := generate.NewFromSpec(
				&specs.Spec{
					Process: &specs.Process{},
				})
```
However, my problem with this is that there are locations in the code where cri-o calls generate.Generator{} directly, instead of using the New... functions:
```
[akaris@workstation cri-o (numeric-irq-load-balancing)]$ grep generate.Generator{ -RI
internal/lib/restore_test.go:   g := generate.Generator{Config: &spec}
internal/oci/oci_linux.go:      g := &generate.Generator{
internal/runtimehandlerhooks/high_performance_hooks_test.go:                    return &generate.Generator{
pkg/config/workloads_test.go:                           g := &generate.Generator{
pkg/config/workloads_test.go:                           g := &generate.Generator{
[akaris@workstation cri-o (numeric-irq-load-balancing)]$ grep generate.New -RI
internal/lib/restore.go:        ctrSpec, err := generate.NewFromFile(filepath.Join(ctr.Dir(), "config.json"))
internal/lib/checkpoint.go:     specgen, err := generate.NewFromFile(configFile)
internal/lib/checkpoint.go:     g, err := generate.NewFromFile(jsonPath)
internal/config/seccomp/seccomp_test.go:                        generator, err := generate.New("linux")
internal/config/seccomp/seccomp_test.go:                        generator, err := generate.New("linux")
internal/config/seccomp/seccomp_test.go:                        generator, err := generate.New("linux")
internal/factory/container/container.go:        spec, err := generate.New(runtime.GOOS)
internal/runtimehandlerhooks/high_performance_hooks_test.go:                    specgen := generate.NewFromSpec(
```

So there's a legit risk that the map may not be initialized.

So this here might work:
```
$ git diff
diff --git a/internal/runtimehandlerhooks/high_performance_hooks_linux.go b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
index c1ba7301f6..014758a484 100644
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -1569,13 +1569,11 @@ func excludeHousekeepingCPUs(cpus string, housekeepingSiblings cpuset.CPUSet) (c
 // injectHousekeepingEnv adds the HOUSEKEEPING_CPUS environment variable to the container.
 // This allows the container to be aware of which CPUs are designated for housekeeping tasks.
 func injectHousekeepingEnv(specgen *generate.Generator, housekeeping cpuset.CPUSet) error {
-       if specgen == nil || specgen.Config == nil || specgen.Config.Process == nil {
-               return fmt.Errorf("specgen or one of its required elements is nil, specgen: %v", specgen)
+       if specgen == nil {
+               return fmt.Errorf("specgen is nil, specgen")
        }
 
-       specgen.Config.Process.Env = append(specgen.Config.Process.Env,
-               fmt.Sprintf("%s=%s", HousekeepingCPUsEnvVar, housekeeping.String()),
-       )
+       specgen.AddProcessEnv(HousekeepingCPUsEnvVar, housekeeping.String())
 
        return nil
 }
diff --git a/internal/runtimehandlerhooks/high_performance_hooks_test.go b/internal/runtimehandlerhooks/high_performance_hooks_test.go
index e297e7c43b..d46c32c68a 100644
--- a/internal/runtimehandlerhooks/high_performance_hooks_test.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_test.go
@@ -383,11 +383,10 @@ var _ = Describe("high_performance_hooks", func() {
                        Expect(bannedCPUs).To(Equal(expectedBan))
 
                        // Also test that injection via injectHousekeepingEnv works correctly.
-                       specgen := generate.Generator{
-                               Config: &specs.Spec{
+                       specgen := generate.NewFromSpec(
+                               &specs.Spec{
                                        Process: &specs.Process{},
-                               },
-                       }
+                               })
                        if !housekeepingSiblings.IsEmpty() {
                                err := injectHousekeepingEnv(&specgen, housekeepingSiblings)
```

But I feel it's risky. I'll create a helper to initialize the missing spec and stuff manually instead of relying on `specgen.AddProcessEnv(`

bitoku: https://pkg.go.dev/github.com/opencontainers/runtime-tools/generate#NewFromSpec
NewFromSpec is deprecated.

andreaskaris:
@bitoku So, I did the following, now - add this commit:
```
$ git show HEAD^^
commit 85088529f2dd6178c0659250d4d48fbbae5c3c9f
Author: Andreas Karis <ak.karis@gmail.com>
Date:   Fri Oct 3 17:15:36 2025 +0200

    Fix Generator initialization to properly initialize envMap
    
    The Generator struct contains an internal envMap field that caches
    environment variable positions for performance optimization when adding
    environment variables. This map is critical for the proper operation of
    the AddProcessEnv() and AddMultipleProcessEnv() methods, which rely on
    it to efficiently detect and update duplicate environment variables.
    
    The previous code directly instantiated a Generator struct literal,
    which leaves the envMap field as nil (the zero value for maps in Go).
    When methods like AddProcessEnv() are subsequently called, they attempt
    to access g.envMap[key], which will work on a nil map for reads but
    creates a bug: the code path at line 543 tries to assign to g.envMap,
    which will panic with "assignment to entry in nil map".
    
    While the current code path in createContainerPlatform() doesn't
    immediately call methods that would trigger this panic, the bug
    represents a latent defect that could cause runtime panics if future
    changes call environment variable manipulation methods on this Generator
    instance.
    
    Signed-off-by: Andreas Karis <ak.karis@gmail.com>

diff --git a/internal/oci/oci_linux.go b/internal/oci/oci_linux.go
index 2a0ea717f1..4b3c25bde4 100644
--- a/internal/oci/oci_linux.go
+++ b/internal/oci/oci_linux.go
@@ -15,13 +15,13 @@ import (
 const InfraContainerName = "POD"
 
 func (r *runtimeOCI) createContainerPlatform(c *Container, cgroupParent string, pid int) error {
-       g := &generate.Generator{
-               Config: &rspec.Spec{
+       g := generate.NewFromSpec(
+               &rspec.Spec{
                        Linux: &rspec.Linux{
                                Resources: &rspec.LinuxResources{},
                        },
                },
-       }
+       )
 
        // First, set the cpuset as the one for the infra container.
        // This should be overridden if specified in a workload.
@@ -34,7 +34,7 @@ func (r *runtimeOCI) createContainerPlatform(c *Container, cgroupParent string,
        }
 
        // Mutate our newly created spec to find the customizations that are needed for conmon
-       if err := r.config.Workloads.MutateSpecGivenAnnotations(InfraContainerName, g, c.Annotations()); err != nil {
+       if err := r.config.Workloads.MutateSpecGivenAnnotations(InfraContainerName, &g, c.Annotations()); err != nil {
                return err
        }
 ```

With that, I think that all the production code is actually using the generator.New...() methods; and it's just the _test stuff that doesn't:
```
$ grep generate.Generator{ -RI
internal/lib/restore_test.go:   g := generate.Generator{Config: &spec}
internal/runtimehandlerhooks/high_performance_hooks_test.go:                    return &generate.Generator{
pkg/config/workloads_test.go:                           g := &generate.Generator{
pkg/config/workloads_test.go:                           g := &generate.Generator{
```

I added another commit for the injectCPUsetEnv:
```
$ git show
commit 80311951a05bb023a93e2b46c3605b3bb95ecee3 (HEAD -> numeric-irq-load-balancing, andreaskaris/numeric-irq-load-balancing)
Author: Andreas Karis <ak.karis@gmail.com>
Date:   Fri Oct 3 17:24:08 2025 +0200

    HighPerformanceHooks: use specgen.AddProcessEnv in injectCPUsetEnv
    
    Signed-off-by: Andreas Karis <ak.karis@gmail.com>

diff --git a/internal/runtimehandlerhooks/high_performance_hooks_linux.go b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
index dcb54c28e6..aac8f34020 100644
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -1515,10 +1515,8 @@ func getPodQuotaV2(mng cgroups.Manager) (string, error) {
 }
 
 func injectCpusetEnv(specgen *generate.Generator, isolated, shared *cpuset.CPUSet) {
-       spec := specgen.Config
-       spec.Process.Env = append(spec.Process.Env,
-               fmt.Sprintf("%s=%s", IsolatedCPUsEnvVar, isolated.String()),
-               fmt.Sprintf("%s=%s", SharedCPUsEnvVar, shared.String()))
+       specgen.AddProcessEnv(IsolatedCPUsEnvVar, isolated.String())
+       specgen.AddProcessEnv(SharedCPUsEnvVar, shared.String())
 }
```

And in the actual commit, I did:
```
+// injectHousekeepingEnv adds the HOUSEKEEPING_CPUS environment variable to the container.
+// This allows the container to be aware of which CPUs are designated for housekeeping tasks.
+func injectHousekeepingEnv(specgen *generate.Generator, housekeeping cpuset.CPUSet) error {
+       if specgen == nil {
+               return errors.New("specgen is nil, specgen")
+       }
+
+       specgen.AddProcessEnv(HousekeepingCPUsEnvVar, housekeeping.String())
+
+       return nil
+}
```

And in the unit test:
```
			specgen := generate.NewFromSpec(&specs.Spec{Process: &specs.Process{}})
```

File: internal/runtimehandlerhooks/high_performance_hooks_linux.go (hunk 9 above)
Comment (bitoku) [reply internal/runtimehandlerhooks/high_performance_hooks_linux.go]: my bad, I was looking at the old version. The deprecation was cancelled.
https://github.com/opencontainers/runtime-tools/pull/747

Comment (andreaskaris) [reply internal/runtimehandlerhooks/high_performance_hooks_linux.go]: ah yeah https://github.com/opencontainers/runtime-tools/blob/0ea5ed0382a279b30530acccafaf070fefeddafd/generate/generate.go#L267

Precisely because of `The recommended replacement does not work because other code paths expect
an initialized envCacheMap.`. I kind of wished runtime-tools initialized that cache on the fly if it wasn't initialized ...

Comment (andreaskaris) [reply internal/runtimehandlerhooks/high_performance_hooks_linux.go]: Anyway, I reported the issue here:
* https://github.com/opencontainers/runtime-tools/issues/790
And submitted a fix here:
*https://github.com/opencontainers/runtime-tools/pull/791

File: internal/runtimehandlerhooks/high_performance_hooks_linux.go [hunk 11]
```diff
@@ -39,26 +39,28 @@ const (
...
-	IsolatedCPUsEnvVar   = "OPENSHIFT_ISOLATED_CPUS"
-	SharedCPUsEnvVar     = "OPENSHIFT_SHARED_CPUS"
+	cgroupSubTreeControl   = "cgroup.subtree_control"
+	cgroupV1QuotaFile      = "cpu.cfs_quota_us"
+	cgroupV2QuotaFile      = "cpu.max"
+	cpusetCpus             = "cpuset.cpus"
+	cpusetCpusExclusive    = "cpuset.cpus.exclusive"
+	IsolatedCPUsEnvVar     = "OPENSHIFT_ISOLATED_CPUS"
+	SharedCPUsEnvVar       = "OPENSHIFT_SHARED_CPUS"
+	HousekeepingCPUsEnvVar = "HOUSEKEEPING_CPUS"
```
bartwensley: Just curious - the other injected env vars are prefixed with "OPENSHIFT_" - why is this one not the same?

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -686,10 +716,15 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
 		return fmt.Errorf("find container %s CPUs", c.ID())
 	}
 
+	set, err := excludeHousekeepingCPUs(lspec.Resources.CPU.Cpus, housekeepingSiblings)
```
bartwensley: nit: Would be nice to use a more descriptive variable name than "set"

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -1233,6 +1270,10 @@ func isCgroupParentBestEffort(s *sandbox.Sandbox) bool {
 }
 
 func isContainerRequestWholeCPU(cSpec *specs.Spec) bool {
+	if isContainerCPUEmpty(cSpec) || cSpec.Linux.Resources.CPU.Shares == nil {
+		return false
+	}
+
```
bartwensley, line 1276: This doesn't seem related to your change - please explain why this is needed?

internal/runtimehandlerhooks/high_performance_hooks_test.go
```diff
@@ -249,6 +278,190 @@ var _ = Describe("high_performance_hooks", func() {
...
+			err = updateIrqBalanceConfigFile(irqBalanceConfigFile, bannedCPUFlags)
+			Expect(err).ToNot(HaveOccurred())
+			bannedCPUs, err := retrieveIrqBannedCPUMasks(irqBalanceConfigFile)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(bannedCPUs).To(Equal(bannedCPUFlags))
+			err = os.WriteFile(irqSmpAffinityFile, []byte(flags), 0o644)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		Context("with enabled equals to false", func() {
```
bartwensley, line 434: Just for completeness, would it be good to add a couple tests with "enabled" equal to true?

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: Because since I introduced my changes, there was duplication in the code in various locations of:

		spec.Linux.Resources == nil ||
		spec.Linux.Resources.CPU == nil ||
		spec.Linux.Resources.CPU.Cpus == ""

I'll branch it off into a different, small commit instead

andreaskaris: I'll move it to a different commit with a decent message

andreaskaris: I moved it into its own commit. I really don't remember why I made this change, it has been a few months since the initial version of the PR. I'm happy to drop it, though I think that the nil pointer check is safer

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: changed to cpuSet

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: I changed it, good catch, thanks!

internal/runtimehandlerhooks/high_performance_hooks_test.go
andreaskaris: I think it's covered in other tests already, but I threw one in for safe measure

andreaskaris: Ideally, we'd have some decent e2e tests for this, but IIRC that's not possible with upstream crio and has to be in the downstream ci lanes

internal/runtimehandlerhooks/high_performance_hooks_linux.go
bartwensley: OK - in this function it appears we are expecting cSpec.Linux.Resources.CPU.Shares to be valid - would it be a good idea to insert a "log.Error" here before returning false?

bitoku: Generally spec fields can be nil. Logging error here doesn't sound right to me.
If we want to verify that, we could do it in the caller of this function.

andreaskaris:
Thanks for the reviews so far. I'm just wondering if I addressed all of your concerns @bitoku @bartwensley , otherwise please lmk and I'm happy to make the required changes

internal/runtimehandlerhooks/high_performance_hooks_linux.go
bartwensley: OK - my assumption was that since the original code would have caused a panic if cSpec.Linux.Resources.CPU.Shares was not valid that this was a very unexpected error scenario. Could be a bad assumption.

bitoku: It may be because our code path always put the value there by coincidence.

bitoku:
@cri-o/cri-o-maintainers PTAL for approval.

haircommander:
/approve

thank you!

PR #9224, opened 2025-05-27
OCPBUGS-54938: [1.32] sandbox: use created/stopped instead of infra container for readiness

haircommander:
As we hit cases of deadlock when a sandbox is stuck stopping

/kind bug
```release-note
Fix a potential deadlock when an infra container is taking a long time to exit and the sandbox's readiness is blocked on the infra container's opLock
```


sohankunkerkar (approved):
/lgtm
 We might need to fix the githubaction tests but failures don't seem to be related this change.

Issue #9227, opened 2025-05-28
Race condition when changing IRQ settings with multiple containers


andreaskaris:
### What happened?

There's a fairly obvious race condition when SMP SRQ affintity is configured and multiple containers are in play. Both the /proc/irq/default_smp_affinity and the IRQ balance config files are read and written without any mutex in place.

### What did you expect to happen?

no race occurs

### How can we reproduce it (as minimally and precisely as possible)?

```
#!/bin/bash

set -x

affinity_file="/proc/irq/default_smp_affinity"
expected_reset_affinity="3fff"
expected_mask="3e0f"

echo $expected_reset_affinity > $affinity_file
cat $affinity_file

for i in {0..20}; do
	echo "========"
	echo "Run ${i}"
	echo "========"
	kubectl apply -f pod.yaml
	kubectl wait --for=condition=Ready pod/qos-demo --timeout=180s
	mask=$(cat ${affinity_file} | tr -d '\n')
        echo "Got mask: $mask, expected mask: $expected_mask"
        if [ "${mask}" != "${expected_mask}" ]; then
            exit 1
        fi
        kubectl delete pod qos-demo
        kubectl wait --for=delete pod/qos-demo --timeout=180s
	mask=$(cat ${affinity_file} | tr -d '\n')
	echo "After reset --- Got mask: $mask, expected mask: $expected_reset_affinity"
	if [ "${mask}" != "${expected_reset_affinity}" ]; then
	    exit 1
	fi
done
```

```
apiVersion: v1
kind: Pod
metadata:
  name: qos-demo
  annotations:
    irq-load-balancing.crio.io: "disable"
spec:
  hostNetwork: true
  runtimeClassName: performance-performance
  containers:
  - name: qos-demo-ctr-1
    image: quay.io/akaris/nice-test
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-2
    image: quay.io/akaris/nice-test
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-3
    image: quay.io/akaris/nice-test
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
[...]

PR #9228, opened 2025-05-28
HighPerformanceHooks: Fix IRQ SMP affinity race conditions

andreaskaris:
Add mutex lock to /proc/irq/default_smp_affinity and to IRQ service configuration file while reading and writing.

Reported-at: https://github.com/cri-o/cri-o/issues/9227

#### What type of PR is this?

/kind bug

I ran my smoke test on a VM with 14 CPUs where I first installed the entire stack with kubeadm.

Dependencies:
```
# rpm -qa | grep kubelet
kubelet-1.33.1-150500.1.1.x86_64
# rpm -qa | grep kubeadm
kubeadm-1.33.1-150500.1.1.x86_64
# rpm -qa | grep crun
crun-1.21-1.el9.x86_64
# rpm -qa | grep conmon
conmon-2.1.13-1.el9.x86_64
```

/etc/crio/crio.conf.d/99-runtimes.conf
```
[crio.runtime]
infra_ctr_cpuset = "0-3"




# The CRI-O will check the allowed_annotations under the runtime handler and apply high-performance hooks when one of
# high-performance annotations presents under it.
# We should provide the runtime_path because we need to inform that we want to re-use runc binary and we
# do not have high-performance binary under the $PATH that will point to it.
[crio.runtime.runtimes.high-performance]
inherit_default_runtime = true
allowed_annotations = ["cpu-load-balancing.crio.io", "cpu-quota.crio.io", "irq-load-balancing.crio.io", "cpu-c-states.crio.io", "cpu-freq-governor.crio.io"]
```

```
# tail -n 5 /var/lib/kubelet/config.yaml
cpuManagerPolicy: static
cpuManagerPolicyOptions:
  full-pcpus-only: "true"
cpuManagerReconcilePeriod: 5s
reservedSystemCPUs: 0-3
```

I then stopped the crio service, started the compiled crio in foreground `make binaries && bin/crio`, and ran this smoke test:

smoke.sh
```
#!/bin/bash

set -x

expected_mask="3e0f"

cat $affinity_file

for i in {0..20}; do
	set +x
	echo "========"
	echo "Run ${i}"
	echo "========"
	set -x
	kubectl apply -f pod.yaml
	mask=$(cat ${affinity_file} | tr -d '\n')
        echo "Got mask: $mask, expected mask: $expected_mask"
        if [ "${mask}" != "${expected_mask}" ]; then
            exit 1
        fi
        kubectl delete pod qos-demo
	mask=$(cat ${affinity_file} | tr -d '\n')
	echo "After reset --- Got mask: $mask, expected mask: $expected_reset_affinity"
	if [ "${mask}" != "${expected_reset_affinity}" ]; then
	    exit 1
	fi
done
```

pod.yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: qos-demo
  annotations:
spec:
  hostNetwork: true
  containers:
  - name: qos-demo-ctr-1
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-2
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-3
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-4
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-5
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
```

```release-note
None
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_test.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_test.go
@@ -767,64 +767,18 @@ var _ = Describe("high_performance_hooks", func() {
 			Expect(env).To(ContainElements("OPENSHIFT_ISOLATED_CPUS=1-2", "OPENSHIFT_SHARED_CPUS=3-4"))
 		})
 	})
-	Describe("Concurrently run PreStart Hook to set IRQ load balancing", func() {
+	Describe("Make sure that correct runtime handler hooks are set", func() {
+		var runtimeName string
+		var sandboxAnnotations map[string]string
+		var sb *sandbox.Sandbox
+		var cfg *config.Config
+		var hooksRetriever *HooksRetriever
+
 		irqSmpAffinityFile := filepath.Join(fixturesDir, "irq_smp_affinity")
 		irqBalanceConfigFile := filepath.Join(fixturesDir, "irqbalance")
+		flags = "0000,0000ffff"
+
 		ctx := context.Background()
-
-		cfg := config.Config{
-			RuntimeConfig: config.RuntimeConfig{
-				Runtimes: config.Runtimes{
-					"high-performance": {
-						AllowedAnnotations: []string{
-							crioannotations.IRQLoadBalancingAnnotation,
-						},
-					},
-				},
-			},
-		}
-		runtimeHooks := NewMap(ctx, &cfg)
-		for _, h := range runtimeHooks {
-			if hph, ok := h.(*HighPerformanceHooks); ok {
-				hph.irqBalanceConfigFile = irqBalanceConfigFile
-				hph.irqSMPAffinityFile = irqSmpAffinityFile
-			}
-		}
-
-		sbox := sandbox.NewBuilder()
-		createdAt := time.Now()
-		sbox.SetCreatedAt(createdAt)
-		sbox.SetID("sandboxID")
-		sbox.SetName("sandboxName")
-		sbox.SetLogDir("test")
-		sbox.SetShmPath("test")
-		sbox.SetNamespace("")
-		sbox.SetKubeName("")
-		sbox.SetMountLabel("test")
-		sbox.SetProcessLabel("test")
-		sbox.SetCgroupParent("")
-		sbox.SetRuntimeHandler("high-performance")
-		sbox.SetResolvPath("")
-		sbox.SetHostname("")
-		sbox.SetPortMappings([]*hostport.PortMapping{})
-		sbox.SetHostNetwork(false)
-		sbox.SetUsernsMode("")
-		sbox.SetPodLinuxOverhead(nil)
-		sbox.SetPodLinuxResources(nil)
-		err = sbox.SetCRISandbox(
-			sbox.ID(),
-			map[string]string{},
-			map[string]string{
-				crioannotations.IRQLoadBalancingAnnotation: "disable",
-			},
-			&types.PodSandboxMetadata{},
-		)
-		Expect(err).ToNot(HaveOccurred())
-		sbox.SetPrivileged(false)
-		sbox.SetHostNetwork(false)
-		sbox.SetCreatedAt(createdAt)
-		sb, err := sbox.GetSandbox()
-		Expect(err).ToNot(HaveOccurred())
 
 		verifySetIRQLoadBalancing := func(expected string) {
 			content, err := os.ReadFile(irqSmpAffinityFile)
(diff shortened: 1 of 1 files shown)


haircommander:
/ok-to-test

I don't love singletons but i think it makes sense here

/approve

andreaskaris:
@haircommander I'm open for other suggestions - by the way, can you give me a hint for how to write an e2e or something similar for this? Also, a lot of the tests here seem to be broken (at least looking at other PRs), which can / should I ignore, which are important?

Thanks!

server/server.go
```diff
@@ -92,6 +92,8 @@ type Server struct {
 
 	// NRI runtime interface
 	nri *nriAPI
+	// highPerformanceHooks holds a single instance for the entire server.
+	highPerformanceHooks *runtimehandlerhooks.HighPerformanceHooks
```
bitoku, line 96: I'd make a map I can get hooks based on runtimehandler(string).

```go
runtimeHandlerHooks map[string]RuntimeHandlerHooks
```
then you can get hooks like this.
```go
hooks := s.runtimeHandlerHooks[sb.RuntimeHandler()]
```
This runtimeHandlerHooks should be initialized at the start up time.
We can use noop hooks for non-highperformance hooks.
With this change, calling its methods will be simpler than now.

The only concern of this way is that it could cause the race condition when there're multiple hooks manipulating irq files. I'm not sure if there's any use case like this.
If we do consider this case, we may want split out irq operations into a new package and use a singleton pattern. This would be still singleton, but its responsibility is well restricted compared to the singleton of the initial change, I think.

internal/runtimehandlerhooks/runtime_handler_hooks_linux.go
```diff
@@ -10,32 +10,62 @@ import (
 	libconfig "github.com/cri-o/cri-o/pkg/config"
 )
 
-// GetRuntimeHandlerHooks returns RuntimeHandlerHooks implementation by the runtime handler name.
-func GetRuntimeHandlerHooks(ctx context.Context, config *libconfig.Config, handler string, annotations map[string]string) (RuntimeHandlerHooks, error) {
+// Map holds a list of RuntimeHandlerHooks for each registered runtime handler.
+type Map map[string]RuntimeHandlerHooks
```
haircommander: I'd prefer a more precisely named type

haircommander: I also think these need to be either moved to runtime_handler_hooks.go or spoofed in _bsd.go

andreaskaris: > I also think these need to be either moved to runtime_handler_hooks.go or spoofed in _bsd.go

Oh yes, sorry about that, I moved it to runtime_handler_hooks.go now. I'll sqash the commits with the changes and repush.

> I'd prefer a more precisely named type

I like following this part of the effective Go guide (though I have no strong preference for it, either): https://go.dev/doc/effective_go
```
The importer of a package will use the name to refer to its contents, so exported names in the package can use that fact to avoid repetition. (Don't use the import . notation, which can simplify tests that must run outside the package they are testing, but should otherwise be avoided.) For instance, the buffered reader type in the bufio package is called Reader, not BufReader, because users see it as bufio.Reader, which is a clear, concise name. 
```

Meaning when the Map is used, it is referred to as:
```
	runtimehandlerhooks.Map
```

I can rename it, but other than RuntimeHandlerHookMap, nothing comes to mind, and then it reads `runtimehandlerhooks.RuntimeHandlerHookMap`.

andreaskaris:
> @andreaskaris Thanks! The changes look good to me. btw do you know if there's any test to check this feature? this race condition is hard to reproduce, so just checking a normal scenario is fine to confirm this change doesn't break the feature.

@bitoku I added my smoke test procedure to the description right now. I'd love to add an e2e test but don't know were (if you could point me to it, I'd appreciate it). Otherwise, I'll quickly check if I can add a good unit test somewhere

bitoku:
Although I couldn't find any tests about `irq-load-balancing.crio.io: "disable"` in this repo (I found one in openshift private upstream), some irqbalance tests are defined in https://github.com/cri-o/cri-o/blob/main/test/irqbalance.bats, and apparently `ci/prow/ci-cgroupv2-integration` and  `ci/prow/ci-fedora-integration` have a capability of running it.

Since openshift private upstream is not run in this CI, we want the integration test for the irq annotation in this repo and/or o/origin.

andreaskaris:
> Although I couldn't find any tests about `irq-load-balancing.crio.io: "disable"` in this repo (I found one in openshift private upstream), some irqbalance tests are defined in https://github.com/cri-o/cri-o/blob/main/test/irqbalance.bats, and apparently `ci/prow/ci-cgroupv2-integration` and `ci/prow/ci-fedora-integration` have a capability of running it.
> 
> Since openshift private upstream is not run in this CI, we want the integration test for the irq annotation in this repo and/or o/origin.

I added a unit test now which IMO covers what needs to be covered

andreaskaris:
@bitoku is the pipeline failing again, or is it my PR's fault?

bitoku:
@andreaskaris 
It had a problem, but I'm rebuilding the test image. It's going to be fixed soon (hopefully).

bitoku:
@andreaskaris btw lint is complaining, and also freebsd build. could you fix it?

andreaskaris:
@bitoku Check now :) Once everything is o.k., I can squash and push again

internal/runtimehandlerhooks/runtime_handler_hooks_linux.go
```diff
@@ -10,32 +10,52 @@ import (
...
-	if strings.Contains(handler, HighPerformance) {
-		log.Warnf(ctx, "The usage of the handler %q without adding high-performance feature annotations under allowed_annotations will be deprecated under 1.21", HighPerformance)
+	var hphInstance *HighPerformanceHooks
 
-		return &HighPerformanceHooks{irqBalanceConfigFile: config.IrqBalanceConfigFile, cpusetLock: sync.Mutex{}, sharedCPUs: config.SharedCPUSet}, nil
-	}
+	rhh := make(Map)
 
-	if highPerformanceAnnotationsSpecified(annotations) {
-		log.Warnf(ctx, "The usage of the handler %q without adding high-performance feature annotations under allowed_annotations will be deprecated under 1.21", HighPerformance)
```
bitoku, line 25: Because of the change, the check is no longer done.
maybe we can check in each [Pre|Post][Start|Stop] function?

andreaskaris: > Because of the change, the check is no longer done.

Actually, I did this on purpose. The check is still done, but it's done when crio starts and prints this exactly once per handler that's using the high-performance feature.
Do we really want to check and print this every time that a container starts? It looked to me like it was spamming the logs, because it was printed for every container whenever one was started or stopped

andreaskaris: I also changed when the check is printed: which is only when the runtime name contains "high-performance" and when the annotation is not specified. Before, the Warning was printed always when the runtime was named ".*high-performance.*", even when the annotations were present (even more log spamming IMO)

bitoku: > The check is still done, but it's done when crio starts and prints this exactly once per handler that's using the high-performance feature.

The code I highlighted checks the sandbox's annotation, not allowed_annotation. The sandbox's annotation can't be checked at cri-o startup time.

However, the older code, only checks the allowed_annotation https://github.com/cri-o/cri-o/commit/106e662ad70fbdc8c03d1ea02e5dc885fbfbaec7#diff-d9b3c6e0b55a2c65987888288a47e86232d3edea55062138ec2a7fef6128050d .
and considering the fact that the warning didn't work properly, we may be able to remove this check.

@haircommander do you know if we should check the sandbox annotation?

andreaskaris: be it as it may, though ... the `func (m Map) Get(name string) RuntimeHandlerHooks` method replaces `GetRuntimeHandlerHooks` where it was used before. I could just add the `annotations` as a parameter to `func (m Map) Get(name string) RuntimeHandlerHooks` and move the check for the sb annotation in there, that way it's in the same location as before and it's for the actual sandbox annotation.
Then, either return the value in `m[name]` if a highperformance annotation is present on the sandbox, or nil otherwise --> should then behave as before. 
I could still keep the `Warnf ` in its current location (or move it back to Get, I don't care). The rest of the code would remain exactly the same.

The question is if that's necessary

andreaskaris: ok, I added a unit test that should capture the cases that were discussed.

From my understanding, the following cases must be covered for Linux:
- runtime name contains "high-performance" OR sandbox annotation has high performance --> `highPerformanceHooks`
- cpuLoadBalancingAllowed anywhere --> DefaultCPULoadBalanceHooks
- nil otherwise

On other architectures: always DefaultCPULoadBalanceHooks

So having a precreated `map[string]RuntimeHandlerHooks` doesn't make any sense; therefore I went back to something similar to an earlier iteration without this map.

bitoku:
/lgtm

Thank you for fixing it!

PR #9229, opened 2025-05-28
UpdateContainerStatus: fix error logging

kolyshkin:
#### What type of PR is this?

/kind bug

Commit c2255b718 introduced error logging in case low-level runtime "state" command fails. It had a special case for exec.ExitError, which was correct at the time.

Commit f09c1d31b later switched this code to use runtimeCmd wrapper. That wrapper do not return stdout in case of an error, and also sets cmd.Stderr to a buffer, meaning exec.ExitError won't have Stderr filled in.

This explains why both printed stdout and stderr are empty, but the issue was not uncovered until commit c69b09563, which replaced error typecasting with error unwrapping. So, now return of a wrapped `exec.ExitError` from runtimeCmd resulted in printing empty stdout and stderr.

The fix is trivial -- always print an error as is.

While at it, fix a comment placement (misplaced since commit c2255b7).

Fixes https://github.com/opencontainers/runc/issues/4773#issuecomment-2917029451

```release-note
NONE
```

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1106,19 +1106,13 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 	stateCmd := func() (*ContainerState, bool, error) {
 		out, err := r.runtimeCmd("state", c.ID())
 		if err != nil {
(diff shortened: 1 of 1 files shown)


haircommander:
/lgtm

CI is borked but @bitoku is dutifully on it!

PR #9246, opened 2025-06-04
[OCPBUGS-4556]: internal/oci: set initial terminal size while attaching container

sohankunkerkar:
The code relied on the resize channel to provide the initial terminal size, but there's no guarantee that a resize event will be sent immediately upon container attachment. This created a race condition where container starts with default/incorrect terminal size. User's actual terminal size was never communicated to the container. Only subsequent resize events (if any) would correct the size. CRI-O should explicitly communicate terminal dimensions.

Screenshot From 2025-06-04 12-34-02

Before this fix:
```sh
 $ kubectl debug node/192.168.1.16 -i -t --image=ubuntu -- sh
Creating debugging pod node-debugger-192.168.1.16-nqfmj with container debugger on node 192.168.1.16.
Warning: metadata.name: this is used in the Pod's hostname, which can result in surprising behavior; a DNS label is recommended: [must not contain dots]
If you don't see a command prompt, try pressing enter.

# tput cols ; tput lines
191
47
# exit

$ tput cols ; tput lines
190
46
```
After the fix:
```sh
$ kubectl debug node/192.168.1.16 -i -t --image=ubuntu -- sh
Creating debugging pod node-debugger-192.168.1.16-nqfmj with container debugger on node 192.168.1.16.
Warning: metadata.name: this is used in the Pod's hostname, which can result in surprising behavior; a DNS label is recommended: [must not contain dots]
If you don't see a command prompt, try pressing enter.

# tput cols ; tput lines
190
46
# exit

$ tput cols ; tput lines
190
46

```

#### What type of PR is this?

/kind bug

```release-note
Fix initial terminal size while attaching a container
```


internal/oci/runtime_oci.go
```diff
@@ -1294,6 +1294,19 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
...
+	if tty {
+		// Get current terminal size from stdin/stdout.
+		if terminalSize, err := getTerminalSize(); err == nil {
+			log.Debugf(ctx, "Setting initial terminal size: %dx%d", terminalSize.Height, terminalSize.Width)
+
+			if _, err := fmt.Fprintf(controlFile, "%d %d %d\n", 1, terminalSize.Height, terminalSize.Width); err != nil {
+				log.Debugf(ctx, "Failed to write initial terminal size to control file: %v", err)
+			}
+		} else {
+			log.Debugf(ctx, "Failed to get initial terminal size: %v", err)
```
bitoku: Don't we want to return error in this case?

sohankunkerkar: I think we need to decide whether we really need to fail in this case. If it's just setting an initial terminal size and the process can still function fine without it, I’d lean toward simply logging the error and proceeding.

bitoku: > If it's just setting an initial terminal size and the process can still function fine without it, I’d lean toward simply logging the error and proceeding.

+1
But unless it happens very frequently, we can make it `Warn` or `Error`

internal/oci/runtime_oci.go
```diff
@@ -1294,6 +1294,19 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
 
 	defer controlFile.Close()
 
+	if tty {
+		// Get current terminal size from stdin/stdout.
+		if terminalSize, err := getTerminalSize(); err == nil {
```
haircommander: I'd prefer without the else personally:
```
terminalSize, err := getTerminalSize()
if err != nil {...}
log.Debug...
if _, err := ....
}
```

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

internal/oci/runtime_oci.go
```diff
@@ -1294,6 +1294,20 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
 
 	defer controlFile.Close()
 
+	if tty {
+		// Get current terminal size from stdin/stdout.
+		terminalSize, err := getTerminalSize()
```
haircommander: wait the other way makes way more sense, terminalSize won't be populated if err != nil :upside_down_face: sorry can you revert the change

kolyshkin (changes requested):
1. You can use golang.org/x/term functions IsTerminal and GetSize to make the code simpler.

2. I don't think querying stdin/stdout/stderr of the current process for terminal size is a good idea, since the current process is a cri-o daemon which is not running on the same terminal as whoever is trying to do exec. 

Probably the correct solution belongs elsewhere -- whoever is calling Attach should send a resize event.

sohankunkerkar:
> 1. You can use golang.org/x/term functions IsTerminal and GetSize to make the code simpler.
> 2. I don't think querying stdin/stdout/stderr of the current process for terminal size is a good idea, since the current process is a cri-o daemon which is not running on the same terminal as whoever is trying to do exec.
> 
> Probably the correct solution belongs elsewhere -- whoever is calling Attach should send a resize event.

Thanks @kolyshkin 
I think it makes complete sense. However, this bug has already been triaged by the workloads team. I think resizing the terminal size during commands such as `oc exec/attach` is managed through bidirectional communication between the local machine and the pod. In this case, oc attach command sends 2 resizing events and cri-o first gets the increased size and then re-renders upon receiving the original size again. I checked the logs in the debug mode and it seems that second event is ignored by CRI-O somehow due to the race condition and monitor size is stuck at +1 version of size. That's where it falls under the purview of CRI-O where we need to evaluate why it's failing to show the correct terminal size.

sohankunkerkar:
>Have you confirmed this fixes the issue?

Yeah, the current fix worked for me locally.

internal/oci/runtime_oci.go
```diff
@@ -1294,13 +1295,32 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
 
 	defer controlFile.Close()
 
+	var (
+		lastSize         remotecommand.TerminalSize
+		controlFileLock  sync.Mutex
+		controlFileValid = true
```
haircommander, line 1301: it doesn't seem like this ever gets set back to true. if we hit this issue then we'll not get anoter resize event, which doesn't seem right.

It seems the investigation is pointing to the fact that the buffer is getting overloaded. if we see this at container startup with two events happening quikcly, maybe we just double the buffer size?

I don't think this is the right approach...

sohankunkerkar: Yeah, you are right. I'm going to revert this change. The actual fix should be in conmon. https://github.com/containers/conmon/pull/566

Issue #9242, opened 2025-06-02
Proposal: Add `never_wipe_images` Configuration Option to Prevent Image Storage Wiping


sonnysasaka:
I am proposing the addition of a new configuration option named `never_wipe_images` in CRI-O to address the needs of users operating in disconnected environments. This option should ensure that image storage is never wiped, regardless of the circumstances, and provide users with greater control over storage management.

### Background

In disconnected environments, where we operate OpenShift without connectivity to an external container registry, it is often crucial to preserve image storage data, even if it is detected to be incomplete or corrupted. The current `internal_wipe` option allows users to prevent storage wiping at startup by setting it to false. However, there are two main concerns:

* [Deprecation](https://github.com/cri-o/cri-o/blob/9e4d86d82370ad44c14649f28ebdd8f94aaa28ca/docs/crio.conf.5.md?plain=1#L61) of `internal_wipe`: There are plans to deprecate this option, which would result in storage being wiped at every startup after an upgrade without the ability to override this behavior.
* Wiping in Other Scenarios: Wiping is also performed in scenarios such as [failed repair attempts](https://github.com/cri-o/cri-o/blob/9e4d86d82370ad44c14649f28ebdd8f94aaa28ca/internal/lib/container_server.go#L122), which are not controlled by the `internal_wipe` flag.

### Proposal

I propose the introduction of a new configuration option, `never_wipe_images`, which can be set to true or false. When set to true, this option should:

* Prevent image storage from being wiped under any circumstances, including startup and failed repair attempts.
* Override any other flags or conditions that would otherwise trigger a wipe of images.
* Not affect wiping of containers, which still makes sense to be controlled separately.

### Benefits

* Flexibility and Control: Users in disconnected environments can maintain image storage integrity without the risk of data loss due to automatic wiping.

### Offer to Contribute

I am capable and willing to contribute to the development of this feature by submitting Pull Requests if the proposal is accepted. I believe this addition will greatly benefit users who require strict control over image storage management in CRI-O.

Thank you for considering this proposal. I look forward to your feedback and am eager to collaborate on implementing this feature.

haircommander:
hm interesting idea. The idea of internal_repair is to make targeted fixes, and only if that fails wipe all the images. In the disconnected case, not doing the wipe means there will need to be manual intervention to solve (wipe the corrupted images and restore to a good state). Isn't that the same as CRI-O doing the wiping but then someone needing to intervene and restore the node to a correct state?

[...]

PR #9264, opened 2025-06-18
Implement never_wipe_images flag

sonnysasaka:
Implement https://github.com/cri-o/cri-o/issues/9242.

#### What type of PR is this?

/kind feature

This adds a new configuration option `never_wipe_images`, which ensures that CRI-O never wipes images under any circumstances. This option makes sense when the deployment is in a disconnected environment and it's very important to preserve local images.

Fixes https://github.com/cri-o/cri-o/issues/9242.

```release-note
Add `never_wipe_images` configuration option.
```


saschagrunert:
I think this change would be fine, do you mind undrafting the PR so that we can run CI for it?

PR #9267, opened 2025-06-20
temporarily enable debug symbols

rphillips:
turn on debug symbols temporarily for kubernetes CI.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ GO_FILES := $(shell find . -type f -name '*.go' -not -name '*_test.go')
 GINKGO_SKIP_PACKAGES = test/nri
 
 # Set DEBUG=1 to enable debug symbols in binaries
-DEBUG ?= 0
+DEBUG ?= 1
 ifeq ($(DEBUG),0)
 SHRINKFLAGS = -s -w
 else


haircommander:
opened https://github.com/cri-o/cri-o/issues/9268 to remind us to revert before we release

/approve
/lgtm

haircommander:
@kolyshkin I'm seeing some runc failures with `runc update` https://github.com/cri-o/cri-o/actions/runs/15781437991/job/44488021983?pr=9267 that I wonder if https://github.com/opencontainers/runc/pull/4785/ caused

haircommander:
testing in https://github.com/cri-o/cri-o/pull/9269

kolyshkin:
> @kolyshkin I'm seeing some runc failures with `runc update` https://github.com/cri-o/cri-o/actions/runs/15781437991/job/44488021983?pr=9267 that I wonder if [opencontainers/runc#4785](https://github.com/opencontainers/runc/pull/4785) caused

Thanks!

I see that it fails at the `[[ "$output" == *"20"* ]]` line.

We've changed the CPU shares to CPU weight formula (see https://github.com/kubernetes/kubernetes/issues/131216, https://github.com/opencontainers/runc/issues/4772, https://github.com/opencontainers/cgroups/pull/20) and the test needs to be amended.

crun will have the same change BTW (https://github.com/containers/crun/pull/1767).

The problem, of course, is now we need to test both old and new runc and crun using the same test. Oh my.

haircommander:
should there be an OCI runtime feature so we can toggle the expectation?

kolyshkin:
Something like this should fix this.

```diff
diff --git a/test/ctr.bats b/test/ctr.bats
index 0375cc9cd..6177f7d4a 100644
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -869,8 +869,10 @@ function assert_log_linking() {
                [[ "$output" == *"20000 10000"* ]]
 
                output=$(crictl exec --sync "$ctr_id" sh -c "cat /sys/fs/cgroup/cpu.weight")
-               # 512 shares are converted to cpu.weight 20
-               [[ "$output" == *"20"* ]]
+               # CPU shares of 512 is converted to cpu.weight of either 20 or 59,
+               # depending on crun/runc version (see https://github.com/kubernetes/kubernetes/issues/131216).
+               echo "got cpu.weight $output, want 20 or 59"
+               [ "$output" = "20" ] || [ "$output" = "59" ]
        else
                output=$(crictl exec --sync "$ctr_id" sh -c "cat /sys/fs/cgroup/cpu/cpu.shares")
                [[ "$output" == *"512"* ]]
```

kolyshkin:
> should there be an OCI runtime feature so we can toggle the expectation?

I might be low on coffee but it seems we don't really have a mechanism for that.

This stems from the fact that we're using obsoleted cgroup v1 units for cgroup v2. The ultimate solution is to switch to [unified settings](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#unified) (i.e. set `cpu.weight` directly so there's no conversion).

kolyshkin:
> Something like this should fix this.

Opened https://github.com/cri-o/cri-o/pull/9271

PR #9271, opened 2025-06-20
test/ctr.bats: fix wrt new CPU units to weight conversion

kolyshkin:
#### What type of PR is this?

/kind ci

It's a long story (see [1], [2], [3], [4]) but both runc and crun is changing the formula to convert cgroup v1 CPU shares to cgroup v2 CPU weight, and it causes a failure in "ctr update resources" test, because it relies on the old conversion formula.

Let's modify it so it works either the old or the new conversion.

(Ultimately, with cgroup v2 we should switch to setting unified.cpu.weight directly).

[1]: https://github.com/kubernetes/kubernetes/issues/131216
[2]: https://github.com/opencontainers/runc/issues/4772
[3]: https://github.com/opencontainers/cgroups/pull/20
[4]: https://github.com/containers/crun/pull/1767

None

```release-note
NONE
```

### DIFF
--- a/test/ctr.bats
+++ b/test/ctr.bats
@@ -869,8 +869,10 @@ function assert_log_linking() {
 		[[ "$output" == *"20000 10000"* ]]
 
 		output=$(crictl exec --sync "$ctr_id" sh -c "cat /sys/fs/cgroup/cpu.weight")
-		# 512 shares are converted to cpu.weight 20
-		[[ "$output" == *"20"* ]]
+		# CPU shares of 512 is converted to cpu.weight of either 20 or 59,
+		# depending on crun/runc version (see https://github.com/kubernetes/kubernetes/issues/131216).
+		echo "got cpu.weight $output, want 20 or 59"
+		[ "$output" = "20" ] || [ "$output" = "59" ]
 	else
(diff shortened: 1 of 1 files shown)


haircommander:
/lgtm

thanks!

sohankunkerkar:
/hold
The test is still failing

kolyshkin:
> The test is still failing

What was the exact error message? Ah, never mind, I found it:

```
not ok 118 ctr update resources
# (in test file ./ctr.bats, line 903)
#   `[[ "$output" == *"10"* ]]' failed
```

As you can see
 * this place was not patched as I overlooked it;
 * the previous failure is no more, i.e. the fix works.

I have updated the PR to patch the second place as well.

If you will see more failures like this, maybe it makes sense to patch it the same way I did in runc (see e.g. https://github.com/opencontainers/runc/pull/4785/commits/da90947848788dd65e8a766723bca85ec452662a#diff-abfdfe7d8e6e32e067593760321e317483bd3425076b195eaee621d4660d99a0).

sohankunkerkar:
@kolyshkin Thanks for the fix. One last CI failure: https://github.com/cri-o/cri-o/actions/runs/15788261889/job/44509215327?pr=9271 (formatting error) and this PR is good to merge.

Issue #9225, opened 2025-05-27
Deprecate `insecure_registries` config


bitoku:
This issue is a tracking issue to deprecate `insecure_registries` option.

Since it's impossible to reliably check if the image registry's IP address is in the specified insecure CIDRs, and it causes another problem that cri-o tries to resolve the primary registry hostname instead of its mirror registries, which resulted in the long-time image pull, we should get rid of this option and should delegate it to registries.conf.

The proposal is:

- Stop CIDR matching
- The functionality of insecure hostname and insecure IP address (only when the registry host is in IP address format) can be achieved by registries.conf insecure option.
- Remove this option completely.

ref:
 - https://github.com/cri-o/cri-o/pull/9216#pullrequestreview-2868711869
 - https://github.com/containers/image/pull/539#issuecomment-490261472

PR #9278, opened 2025-06-24
Deprecate insecure_registries config

bitoku:
#### What type of PR is this?

/kind deprecation

This PR deprecates insecure-registries, which causes unnecessary DNS resolution.

part of #9225 

```release-note
Deprecate `insecure_registries` config.
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -1082,6 +1082,10 @@ func GetImageService(ctx context.Context, store storage.Store, storageTransport 
 		regexForPinnedImages: CompileRegexpsForPinnedImages(serverConfig.PinnedImages),
 	}
 
+	if len(serverConfig.InsecureRegistries) > 0 {
+		log.Warnf(ctx, "Insecure registries option is deprecated and will not have any effect in a future release")
+	}
+
 	serverConfig.InsecureRegistries = append(serverConfig.InsecureRegistries, "127.0.0.0/8")
 	// Split --insecure-registry into CIDR and registry-specific settings.
 	for _, r := range serverConfig.InsecureRegistries {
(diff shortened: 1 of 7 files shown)


internal/storage/image.go
```diff
@@ -1082,6 +1082,9 @@ func GetImageService(ctx context.Context, store storage.Store, storageTransport
 		regexForPinnedImages: CompileRegexpsForPinnedImages(serverConfig.PinnedImages),
 	}
 
+	if len(serverConfig.InsecureRegistries) > 0 {
+		log.Warnf(ctx, "Insecure registries option is deprecated and will be no effect")
+	}
 	serverConfig.InsecureRegistries = append(serverConfig.InsecureRegistries, "127.0.0.0/8")
```
bitoku, line 1089: This 127.0.0.0/8 will be removed when we completely remove the option.
Do we want to have a new option like `insecureLoopback` and deprecate it separately to reduce the impact as much as possible? or is it too much?

saschagrunert: I can't recall who requires this, but I tend to say _no_, let's deprecate without having a replacement for this for now.

bitoku:
@cri-o/cri-o-maintainers PTAL

internal/storage/image.go
```diff
@@ -1082,6 +1082,10 @@ func GetImageService(ctx context.Context, store storage.Store, storageTransport
 		regexForPinnedImages: CompileRegexpsForPinnedImages(serverConfig.PinnedImages),
 	}
 
+	if len(serverConfig.InsecureRegistries) > 0 {
+		log.Warnf(ctx, "Insecure registries option is deprecated and will be no effect")
```
haircommander: nit: `and will not have any affect in a future release` ?

bitoku:
/skip
@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):

PR #9280, opened 2025-06-24
Issue #9125 - Memory swap limit bytes metrics for the containers

venugopal83k:
#### What type of PR is this?
/kind feature

#9125 - This PR adds the container_spec_memory_swap_limit_bytes metric to the CRI-O metrics suite, which reports the SWAP memory limit for containers in bytes. This metric is valuable for monitoring and alerting on SWAP memory configuration versus usage. It allows users to easily calculate SWAP memory usage percentages and visualize resource allocation versus consumption in monitoring systems like Prometheus and Grafana.

```release-note
Added `container_spec_memory_swap_limit_bytes` metric
```

### DIFF
--- a/internal/lib/stats/descriptors.go
+++ b/internal/lib/stats/descriptors.go
@@ -68,6 +68,11 @@ var (
 		Help:      "Memory limit for the container in bytes.",
 		LabelKeys: baseLabelKeys,
 	}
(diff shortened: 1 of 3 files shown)


haircommander:
/ok-to-test

would love a test in `test/cri-metrics.bats` if possible

PR #9286, opened 2025-06-26
OCPNODE-2864: Set ExecCPUAffinity

bitoku:
#### What type of PR is this?

/kind feature

This implements https://issues.redhat.com/browse/OCPNODE-2864

Fixes https://issues.redhat.com/browse/OCPNODE-2864

```release-note
Added `exec_cpu_affinity` type which can specify cpu where exec command runs.
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -27,6 +27,7 @@ import (
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
 	crioannotations "github.com/cri-o/cri-o/pkg/annotations"
+	"github.com/cri-o/cri-o/pkg/config"
 	"github.com/cri-o/cri-o/utils/cmdrunner"
 )
 
@@ -67,6 +68,7 @@ type HighPerformanceHooks struct {
 	irqSMPAffinityFileLock   sync.Mutex
 	irqBalanceConfigFileLock sync.Mutex
 	sharedCPUs               string
+	execCPUAffinity          config.ExecCPUAffinityType
 	irqSMPAffinityFile       string
 }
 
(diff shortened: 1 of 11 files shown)


bitoku:
@haircommander @MarSik @bartwensley PTAL

haircommander:
/approve

looks good!

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -77,35 +79,78 @@ func (h *HighPerformanceHooks) PreCreate(ctx context.Context, specgen *generate.
...
 		// by the low-level runtime and the environment variables are already finalized.
-		injectCpusetEnv(specgen, &exclusiveCPUs, &sharedCPUSet)
+		injectCpusetEnv(specgen, &exclusiveCPUSet, &sharedCPUSet)
 	}
 
 	return nil
 }
 
+// setExecCPUAffinity sets ExecCPUAffinity in the container spec.
+func (h *HighPerformanceHooks) setExecCPUAffinity(ctx context.Context, specgen *generate.Generator, exclusiveCPUSet, sharedCPUSet *cpuset.CPUSet) cpuset.CPUSet {
```
bartwensley: The logic looks good to me. But why didn't you add a unit test for setExecCPUAffinity and also extend the unit test for PreCreate to cover the new functionality you added?

bitoku: I usually rely on e2e tests because of the external dependencies, but this will suit well unit test, it seems. Thanks!

MarSik:
Just a note as I am on vacation and only noticed the change. It might be a good idea to sync this with https://github.com/cri-o/cri-o/pull/9223 where I proposed we allow configuring the housekeeping cpus using an annotation like `housekeeping-cpus.crio.io/[container-name]: 0,1`

I know the first cpu is traditional, but this would give us more flexibility. We can always default to the first cpu (+ hyperthreading siblings?) when the annotation is missing of course.

bitoku:
Thanks @MarSik .
Yeah it may be a good idea to have that option.
I made the option extendable so we can add options later.
I'll keep this PR as it is for now.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
bitoku: I added unit tests.

internal/runtimehandlerhooks/high_performance_hooks_test.go
```diff
@@ -766,6 +767,81 @@ var _ = Describe("high_performance_hooks", func() {
...
+				sbox.SetPodLinuxOverhead(nil)
+				sbox.SetPodLinuxResources(nil)
+				err = sbox.SetCRISandbox(sbox.ID(), make(map[string]string), map[string]string{}, &types.PodSandboxMetadata{})
+				Expect(err).ToNot(HaveOccurred())
+				sbox.SetPrivileged(false)
+				sbox.SetHostNetwork(false)
+				sbox.SetCreatedAt(createdAt)
+				sbNoShared, err = sbox.GetSandbox()
+				Expect(err).ToNot(HaveOccurred())
+			})
```
bartwensley: Thanks for adding the unit tests. Instead of nesting the Describe and duplicating the BeforeEach, it would be cleaner to restructure the original "Precreate hook" test to use the "Context"/"BeforeEach" pattern (see other tests in this file) and pass in the crioannotations for each test, which seems to be the only thing that differs here.

haircommander:
@MarSik I thought @browsell mentioned users don't want another annotation. Maybe we have an additional option to have the range be configurable and introduce both the configurable range, the First, and the default?

browsell:
> @MarSik I thought @browsell mentioned users don't want another annotation. Maybe we have an additional option to have the range be configurable and introduce both the configurable range, the First, and the default?

We do not want to add any additional annotations.

internal/runtimehandlerhooks/high_performance_hooks_test.go
bitoku: updated. let me know how it looks.

test/exec_cpu_affinity.bats
```diff
@@ -0,0 +1,79 @@
...
+
+}
+
+function teardown() {
+	cleanup_test
+}
+
+@test "should not specify the exec cpu affinity" {
+	skip_if_vm_runtime
+	crun_path=$(which crun)
```
haircommander: you can use RUNTIME_BINARY_PATH here

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -294,6 +347,7 @@ func annotationValueDeprecationWarning(annotation string) string {
 func requestedSharedCPUs(annotations fields.Set, cName string) bool {
 	key := crioannotations.CPUSharedAnnotation + "/" + cName
 	v, ok := annotations[key]
+	logrus.Debug(annotations)
```
haircommander: did you mean to leave this?

bitoku: no sorry, I'll remove this, thanks

test/exec_cpu_affinity.bats
bitoku: thanks! fixed

internal/runtimehandlerhooks/high_performance_hooks_test.go
bartwensley: Looks great - thanks for updating/extending the unit tests.

bitoku:
@cri-o/cri-o-maintainers PTAL

haircommander:
you've got a merge conflict

bitoku:
@cri-o/cri-o-maintainers PTAL

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -77,35 +79,85 @@ func (h *HighPerformanceHooks) PreCreate(ctx context.Context, specgen *generate.
...
+		case exclusiveCPUSet != nil && !exclusiveCPUSet.IsEmpty():
+			execCPUSet = cpuset.New(exclusiveCPUSet.List()[0])
+		default:
+			log.Errorf(ctx, "ExecCPUAffinityType %s is set, but no CPUSet is available. Falling back to default.", h.execCPUAffinity)
+		}
+	case config.ExecCPUAffinityTypeDefault:
+		// Don't set ExecCPUAffinity, which means using runtime default.
+	default:
+		// This shouldn't happen because there's config validation.
+		log.Errorf(ctx, "Unknown ExecCPUAffinityType %s is used. Falling back to default.", h.execCPUAffinity)
```
haircommander: if this shouldn't happen we should panic here because it's a coding error that should be caught by testing

haircommander:

still LGTM, one more nit if you have time but not a blocker

do any other @cri-o/cri-o-maintainers want to TAL?

internal/runtimehandlerhooks/high_performance_hooks_linux.go
bitoku: It makes sense, but I wonder if we can find it in test because cri-o runs as a service, and get restarted easily.
I'd rather returning error instead of panic, which don't break entire system, but still have a chance to catch coding errors I guess.

bitoku:
@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):

Issue #7826, opened 2024-02-28
[FEAT] Enhance exec operation to survive cri-o restarts


hasan4791:
### What happened?

Currently cri-o holds the streaming server which manages the exec connection to/from the running containers. This is causing the exec to break when the user has open exec session to pod or debuging node (using kubectl/oc debug command). If possible, we can offload exec operation to intermediate application like conmon or something similar, it would ideally survive the cri-o restarts. Open to discussion on this topic.

### What did you expect to happen?

NA

### How can we reproduce it (as minimally and precisely as possible)?

NA

### Anything else we need to know?

NA

### CRI-O and Kubernetes version

NA

### OS version

NA

### Additional environment details (AWS, VirtualBox, physical, etc.)

NA

saschagrunert:
This is a feature we can elaborate on, maybe it makes sense to keep the connection open in conmon-rs.

kwilczynski:
/retitle Enhance exec operation to survive CRI-O restarts

kwilczynski:
@hasan4791, do you have some things that require to be run for longer?

Why would a momentary interruption be a problem? While CRI-O restarts, which should be momentary, hopefully. What would be the use case you have in mind?

I am asking as there is also a push to enable and ensure a timeout for any streaming operations from the container back to the user.

haircommander:
I think this is an all or nothing thing. If CRI-O restarts, it looses the connection to the kubelet which will tear down the endpoint assuming the container stopped . I think delegating to conmon-rs is a good idea

hasan4791:
@kwilczynski No in my case, but i was fiddling with registry config in the cluster nodes and had to restart the runtime which disconnected the session. I thought it would be a good to have thing to keep the session open unless user disconnects or times out on ideal.

hasan4791:
@haircommander Is it possible to extend the same for log streaming as well?

haircommander:
Yeah the mechanism could be similar

saschagrunert:
So, exec and attach are pretty much the same, portforward is slightly different but should work as well. Logs is tricky, because the log reader in the kubelet (used by crictl as well) is doing a container status to ensure that the container is running: https://github.com/kubernetes/kubernetes/blob/9c50b2503b7190dbbf7e745103bc58574b3da590/pkg/kubelet/kuberuntime/logs/logs.go#L418-L444

[...]

PR #9289, opened 2025-06-27
Add conmon-rs streaming server support for `Exec` and `Attach`

saschagrunert:
#### What type of PR is this?

/kind feature

Add support for websocket exec and attach using conmon-rs. This means that those sessions will survive a CRI-O restart. Drawback is that we cannot know if spdy or a websocket connection is being used on setup. Therefore we cannot support spdy through conmon-rs anymore. The whole feature is gated through the `stream_websockets` runtime handler configuration.

Refers to #7826, https://github.com/kubernetes/enhancements/issues/4006

Requires https://github.com/cri-o/cri-o/pull/9290

```release-note
Added support for conmon-rs streaming server on `Exec` and `Attach`. To enable it, set
`stream_websockets = true` as part of the runtime handler configuration.
```

### DIFF
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -83,6 +83,8 @@ type RuntimeImpl interface {
 	IsContainerAlive(*Container) bool
 	// ProbeMonitor is used to check the liveness of the container monitor process.
 	ProbeMonitor(context.Context, *Container) error
+	ServeExecContainer(context.Context, *Container, []string, bool, bool, bool, bool) (string, error)
+	ServeAttachContainer(context.Context, *Container, bool, bool, bool) (string, error)
 }
 
 // New creates a new Runtime with options provided.
@@ -240,6 +242,16 @@ func (r *Runtime) RuntimeDefaultAnnotations(runtimeHandler string) (map[string]s
 	}
 
 	return rh.RuntimeDefaultAnnotations(), nil
+}
+
+// RuntimeStreamWebsockets returns the configured websocket streaming option for this handler.
+func (r *Runtime) RuntimeStreamWebsockets(runtimeHandler string) (bool, error) {
+	rh, err := r.getRuntimeHandler(runtimeHandler)
+	if err != nil {
+		return false, err
+	}
+
+	return rh.RuntimeStreamWebsockets(), nil
 }
 
 func (r *Runtime) newRuntimeImpl(c *Container) (RuntimeImpl, error) {
@@ -544,4 +556,22 @@ func (r *Runtime) ProbeMonitor(ctx context.Context, c *Container) error {
 	}
 
 	return impl.ProbeMonitor(ctx, c)
+}
+
+func (r *Runtime) ServeExecContainer(ctx context.Context, c *Container, cmd []string, tty, stdin, stdout, stderr bool) (string, error) {
+	impl, err := r.RuntimeImpl(c)
+	if err != nil {
+		return "", err
+	}
+
+	return impl.ServeExecContainer(ctx, c, cmd, tty, stdin, stdout, stderr)
+}
+
+func (r *Runtime) ServeAttachContainer(ctx context.Context, c *Container, stdin, stdout, stderr bool) (string, error) {
+	impl, err := r.RuntimeImpl(c)
+	if err != nil {
+		return "", err
+	}
+
+	return impl.ServeAttachContainer(ctx, c, stdin, stdout, stderr)
 }
--- a/internal/oci/runtime_pod.go
+++ b/internal/oci/runtime_pod.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	"github.com/containers/common/pkg/resize"
@@ -315,4 +316,48 @@ func (r *runtimePod) IsContainerAlive(c *Container) bool {
 func (r *runtimePod) ProbeMonitor(ctx context.Context, c *Container) error {
 	// Not implemented
 	return nil
+}
+
+func (r *runtimePod) ServeExecContainer(ctx context.Context, c *Container, cmd []string, tty, stdin, stdout, stderr bool) (string, error) {
+	res, err := r.client.ServeExecContainer(ctx, &conmonClient.ServeExecContainerConfig{
+		ID:      c.ID(),
+		Command: cmd,
+		Tty:     tty,
+		Stdin:   stdin,
+		Stdout:  stdout,
+		Stderr:  stderr,
+	})
+	if err != nil {
+		if isUnimplementedRPCErr(err) {
+			return "", nil
+		}
+
+		return "", fmt.Errorf("call ServeExecContainer RPC: %w", err)
+	}
+
+	return res.URL, nil
+}
+
+func (r *runtimePod) ServeAttachContainer(ctx context.Context, c *Container, stdin, stdout, stderr bool) (string, error) {
+	res, err := r.client.ServeAttachContainer(ctx, &conmonClient.ServeAttachContainerConfig{
+		ID:     c.ID(),
+		Stdin:  stdin,
+		Stdout: stdout,
+		Stderr: stderr,
+	})
+	if err != nil {
+		if isUnimplementedRPCErr(err) {
+			return "", nil
+		}
+
+		return "", fmt.Errorf("call ServeAttachContainer RPC: %w", err)
+	}
+
+	return res.URL, nil
+}
+
+func isUnimplementedRPCErr(err error) bool {
+	const errUnimplementedString = "Method not implemented."
+
+	return strings.Contains(err.Error(), errUnimplementedString)
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -271,6 +271,13 @@ type RuntimeHandler struct {
 	// Default annotations specified for runtime handler if they're not overridden by
 	// the pod spec.
 	DefaultAnnotations map[string]string `toml:"default_annotations,omitempty"`
+
+	// StreamWebsockets can be used to enable the WebSocket protocol for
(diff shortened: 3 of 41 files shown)


server/container_exec.go
```diff
@@ -17,6 +17,22 @@ import (
...
+		return nil, fmt.Errorf("could not find container %q: %w", req.ContainerId, err)
+	}
+
+	url, err := s.Runtime().ServeExecContainer(ctx, c, req.Cmd, req.Tty, req.Stdin, req.Stdout, req.Stderr)
+	if err != nil {
+		return nil, fmt.Errorf("could not serve exec for container %q: %w", req.ContainerId, err)
+	}
+
+	if url != "" {
+		log.Infof(ctx, "Using exec URL from conmon-rs: %s", url)
```
haircommander: theoretically it could be supported by another process, maybe saying "Using exec URL from container montior" or something?

internal/oci/runtime_pod.go
```diff
@@ -311,3 +312,47 @@ func (r *runtimePod) ReopenContainerLog(ctx context.Context, c *Container) error
 func (r *runtimePod) IsContainerAlive(c *Container) bool {
 	return c.Living() == nil
 }
+
+func isUnimplementedRPCErr(err error) bool {
```
haircommander: nit: can you define below where it's called?

haircommander:
I wonder if there's a way we can plumb the knowledge of websocket v spdy into the request so we can toggle behavior, but that's longer term (or maybe we just move away from spdy...)

internal/oci/runtime_pod.go
saschagrunert: Yes, done :+1:

server/container_exec.go
saschagrunert: Changed the message as requested. :+1:

saschagrunert:
> I wonder if there's a way we can plumb the knowledge of websocket v spdy into the request so we can toggle behavior, but that's longer term (or maybe we just move away from spdy...)

I was thinking about the same. Considering that the transition from SPDY to WS is already ongoing in https://github.com/kubernetes/enhancements/issues/4006, I'd say we don't keep too much backwards compatibility (we also don't have websocket support right now).

The conversation could be a good topic for the v1.35 cycle.

saschagrunert:
Requires https://github.com/kubernetes-sigs/cri-tools/pull/1862 now

saschagrunert:
@cri-o/cri-o-maintainers PTAL

internal/oci/oci.go
```diff
@@ -242,6 +244,16 @@ func (r *Runtime) RuntimeDefaultAnnotations(runtimeHandler string) (map[string]s
 	return rh.RuntimeDefaultAnnotations(), nil
 }
 
+// RuntimeStreamWebsockets returns the default annotations for this runtime handler.
```
haircommander: comment seems off

server/container_attach.go
```diff
@@ -16,6 +17,29 @@ import (
...
+			return nil, fmt.Errorf("could not serve attach for container %q: %w", req.ContainerId, err)
+		}
+
+		if url == "" {
+			log.Infof(ctx, "Runtime %q does not support websocket streaming, falling back to SPDY", runtimeHandler)
+		} else {
+			log.Infof(ctx, "Using attach URL from container monitor: %s", url)
+
+			return &types.AttachResponse{Url: url}, nil
+		}
```
bitoku:
```suggestion
		if url != "" {
			log.Infof(ctx, "Using attach URL from container monitor: %s", url)
			return &types.AttachResponse{Url: url}, nil
		}
		log.Infof(ctx, "Runtime %q does not support websocket streaming, falling back to SPDY", runtimeHandler)
```

pkg/config/config.go
```diff
@@ -271,6 +271,13 @@ type RuntimeHandler struct {
 	// Default annotations specified for runtime handler if they're not overridden by
 	// the pod spec.
 	DefaultAnnotations map[string]string `toml:"default_annotations,omitempty"`
+
+	// StreamWebsockets can be used to enable the WebSocket protocol for
+	// container exec, attach and port forward.
+	//
+	// conmon-rs (runtime_type = "pod") supports this configuration for exec
+	// and attach. Forwarding ports will be supported in future releases.
+	StreamWebsockets bool `toml:"stream_websockets,omitempty"`
```
bitoku, line 280: I think we should validate it when launching cri-o to find the misconfiguration earlier.

nit: If it fallbacks to `StreamWebsockets = false` when it loads the config, we may not have to fallback in Attach and Exec implementation. I don't know if it's intuitive behaviour for users though.

internal/oci/oci.go
saschagrunert: Fixed

server/container_attach.go
saschagrunert: Fixed, also for exec.

pkg/config/config.go
saschagrunert: I was thinking about this as well, but decided to go for a live validation, because we need a recent enough conmon-rs version to support the new protocol. That's why I added an additional check for "not implemented" in: https://github.com/cri-o/cri-o/pull/9289/files#diff-31e979d56d7fd58363546aef5808721eb8d3a2da1aba43e03b5bebcf1d138455R349-R351

The main problem is that we don't check the conmon-rs version during the runtime configuration, where I'm also not sure we should do that. In a long term we will move away from SPDY and switch over to websockets completely.

bitoku:
I think we need to rebuild the test image to update crun. https://prow.ci.openshift.org/?repo=cri-o%2Fcri-o&job=*setup*

bitoku:
We can trigger them manually by opening a PR in o/release and run pj-rehearse. https://github.com/cri-o/cri-o/tree/main/contrib/test

saschagrunert:
@cri-o/cri-o-maintainers PTAL

pkg/config/config.go
saschagrunert: I follow-up on that.

server/container_execsync_test.go
```diff
@@ -18,7 +18,7 @@ var _ = t.Describe("ContainerStart", func() {
 
 	AfterEach(afterEach)
 
-	t.Describe("ContainerStart", func() {
+	t.Describe("ContainerExecSynct", func() {
```
bitoku: typo?
```suggestion
	t.Describe("ContainerExecSync", func() {
```

internal/oci/runtime_pod.go
```diff
@@ -316,3 +317,47 @@ func (r *runtimePod) ProbeMonitor(ctx context.Context, c *Container) error {
...
+	res, err := r.client.ServeExecContainer(ctx, &conmonClient.ServeExecContainerConfig{
+		ID:      c.ID(),
+		Command: cmd,
+		Tty:     tty,
+		Stdin:   stdin,
+		Stdout:  stdout,
+		Stderr:  stderr,
+	})
+	if err != nil {
+		if isUnimplementedRPCErr(err) {
```
bitoku, line 331: https://pkg.go.dev/capnproto.org/go/capnp/v3#IsUnimplemented
Can we use this?

server/container_attach.go
```diff
@@ -16,6 +17,29 @@ import (
...
+	if streamWebsocket, err := s.Runtime().RuntimeStreamWebsockets(runtimeHandler); err == nil && streamWebsocket {
+		log.Debugf(ctx, "Runtime handler %q is configured to use websockets", runtimeHandler)
+
+		url, err := s.Runtime().ServeAttachContainer(ctx, c, req.Stdin, req.Stdout, req.Stderr)
+		if err != nil {
+			return nil, fmt.Errorf("could not serve attach for container %q: %w", req.ContainerId, err)
+		}
+
+		if url != "" {
+			log.Infof(ctx, "Using attach URL from container monitor: %s", url)
```
bitoku: I wonder if we should expose URL in the log?
Is it ok to see people other than the requester can see the url?

internal/oci/runtime_pod.go
saschagrunert: Unfortunately not, it's using type assertions which get lost: https://github.com/capnproto/go-capnp/blob/5f984c42aa2e273c87c7568c32938af740677e50/error.go#L18

server/container_attach.go
saschagrunert: I don't think it's harmful but we can omit the URL :+1:

test/conmonrs.bats
```diff
@@ -11,14 +15,33 @@ function teardown() {
...
+	unset CONTAINER_DEFAULT_RUNTIME
+	unset CONTAINER_RUNTIMES
+
+	start_crio_no_setup
+
+	CTR=$(crictl run "$TESTDATA"/container_redis.json "$TESTDATA"/sandbox_config.json)
+
+	OUTPUT=$(crictl exec -r websocket "$CTR" echo test)
+	[ "$OUTPUT" == "test" ]
+	grep -q "Using exec URL from container monitor " "$CRIO_LOG"
```
bitoku: The test failure looks come from here.
```suggestion
	grep -q "Using exec URL from container monitor" "$CRIO_LOG"
```

test/conmonrs.bats
```diff
@@ -11,14 +15,33 @@ function teardown() {
...
-
 	start_crio
 
 	crictl run "$TESTDATA"/container_redis.json "$TESTDATA"/sandbox_config.json
 
 	# Validate that we actually used conmonrs
 	grep -q "Using conmonrs version:" "$CRIO_LOG"
 }
+
+@test "conmonrs streaming server for exec" {
```
bitoku: Let's add a test for attach.

saschagrunert: Yep, done :+1:

saschagrunert:
PTAL @bitoku

PR #9295, opened 2025-06-27
[release-1.33] server,factory/container: delay CDI device injection later.

klihub:
#### What type of PR is this?

/kind bug

Currently CDI device injection is performed right after injecting other devices into the container. This is problematic because CDI device injection might alter, among other things, the environment. However setting up the final environment happens only later during container creation and it involves setting environment variables from the image and the Pod Spec. If the same environment variable is injected both from an image or a container, and from a CDI Spec, now the former take precedence of the latter. This is unintentional and wrong.

This patch moves CDI device injection much later during container creation, between OCI Hook injection and *oci.Container creation.

None

None

```release-note
server: delay CDI device injection, to ensure that CDI Spec edits take precedence over image defaults and the Pod Spec.
```


klihub:
@haircommander This is a cherry-pick/backport of #9292 to the 1.33 release branch. If we don't suspect those CI failures to be related to the changes this PR brings in, could we get this in ?

dependencies.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   # tag the .0 release if it does not already exist. If the .0 release is done,
   # increase the development version to the next minor (1.x.0).
   - name: development version
-    version: 1.33.0
+    version: 1.33.2
```
haircommander: hm do we need this for a CI test? is this typical @saschagrunert ?

haircommander:
/approve

@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):

sohankunkerkar:
gcp has some infra issues.

PR #9296, opened 2025-06-27
[release-1.32] server,factory/container: delay CDI device injection later.

klihub:
#### What type of PR is this?

/kind bug

Currently CDI device injection is performed right after injecting other devices into the container. This is problematic because CDI device injection might alter, among other things, the environment. However setting up the final environment happens only later during container creation and it involves setting environment variables from the image and the Pod Spec. If the same environment variable is injected both from an image or a container, and from a CDI Spec, now the former take precedence of the latter. This is unintentional and wrong.

This patch moves CDI device injection much later during container creation.

None

```release-note
server: delay CDI device injection, to ensure that CDI Spec edits take precedence over image defaults and the Pod Spec.
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -112,6 +112,9 @@ type Container interface {
 	// SpecAddDevices adds devices from the server config, and container CRI config
 	SpecAddDevices([]device.Device, []device.Device, bool, bool) error
 
+	// SpecInjectCDIDevices injects any requested CDI devices to the container's Spec.
+	SpecInjectCDIDevices() error
+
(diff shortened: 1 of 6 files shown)


elezar:
thanks @klihub

klihub:
@haircommander This is a cherry-pick/backport of #9292 to the 1.32 release branch. If we don't suspect those CI failures to be related to the changes this PR brings in, could we get this in ?

elezar:
lgtm

haircommander:
/approve

@cri-o/cri-o-maintainers PTAL

PR #9301, opened 2025-06-30
OCPBUGS-58229: server: handle missing network namespace gracefully during networkStop

sohankunkerkar:
After host reboot, network namespaces are destroyed but CRI-O attempts to clean them up during pod sandbox destruction, causing CNI plugin failures and preventing pods from restarting properly. The fix ensures pods can restart normally after host reboots.

#### What type of PR is this?

/kind bug

[Link to journal logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-microshift-release-4.20-periodics-e2e-aws-tests-bootc-nightly/1939504337394864128/artifacts/e2e-aws-tests-bootc-nightly/openshift-microshift-e2e-metal-tests/artifacts/scenario-info/el96-src@ai-model-serving-offline/vms/host1/sos/journal_2025-06-30_03:38:43.log)

```release-note
Handle missing network namespace gracefully during networkStop
```

### DIFF
--- a/server/sandbox_network.go
+++ b/server/sandbox_network.go
@@ -181,23 +181,42 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
 
 	podNetwork, err := s.newPodNetwork(ctx, sb)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create pod network for sandbox %s(%s): %w", sb.Name(), sb.ID(), err)
+	}
+
+	// Check if the network namespace file exists and is valid before attempting CNI teardown.
+	// If the file doesn't exist or is invalid, skip CNI teardown and mark network as stopped.
+	if podNetwork.NetNS != "" {
+		if _, statErr := os.Stat(podNetwork.NetNS); statErr != nil {
+			// Network namespace file doesn't exist, mark network as stopped and return success
+			log.Debugf(ctx, "Network namespace file %s does not exist for pod sandbox %s(%s), skipping CNI teardown",
+				podNetwork.NetNS, sb.Name(), sb.ID())
+
+			return sb.SetNetworkStopped(ctx, true)
+		}
+
+		if validateErr := s.validateNetworkNamespace(podNetwork.NetNS); validateErr != nil {
+			// Network namespace file exists but is invalid (e.g., corrupted or fake file)
+			log.Warnf(ctx, "Network namespace file %s is invalid for pod sandbox %s(%s): %v, removing and skipping CNI teardown",
+				podNetwork.NetNS, sb.Name(), sb.ID(), validateErr)
+			s.cleanupNetns(ctx, podNetwork.NetNS, sb)
+
+			return sb.SetNetworkStopped(ctx, true)
+		}
 	}
 
 	if err := s.config.CNIPlugin().TearDownPodWithContext(stopCtx, podNetwork); err != nil {
-		retErr := fmt.Errorf("failed to destroy network for pod sandbox %s(%s): %w", sb.Name(), sb.ID(), err)
+		log.Warnf(ctx, "Failed to destroy network for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
 
-		if _, statErr := os.Stat(podNetwork.NetNS); statErr != nil {
-			return fmt.Errorf("%w: stat netns path %q: %w", retErr, podNetwork.NetNS, statErr)
+		// If the network namespace exists but CNI teardown failed, try to clean it up.
+		if podNetwork.NetNS != "" {
+			if _, statErr := os.Stat(podNetwork.NetNS); statErr == nil {
+				// Clean up the netns file since CNI teardown failed.
+				s.cleanupNetns(ctx, podNetwork.NetNS, sb)
+			}
 		}
 
-		// The netns file may still exists, which means that it's likely
-		// corrupted. Remove it to allow cleanup of the network namespace:
-		if rmErr := os.RemoveAll(podNetwork.NetNS); rmErr != nil {
-			return fmt.Errorf("%w: failed to remove netns path: %w", retErr, rmErr)
-		}
-
-		log.Warnf(ctx, "Removed invalid netns path %s from pod sandbox %s(%s)", podNetwork.NetNS, sb.Name(), sb.ID())
+		return fmt.Errorf("network teardown failed for pod sandbox %s(%s): %w", sb.Name(), sb.ID(), err)
 	}
 
 	return sb.SetNetworkStopped(ctx, true)
--- a/server/sandbox_network_linux.go
+++ b/server/sandbox_network_linux.go
@@ -1,0 +1,36 @@
+//go:build linux
+// +build linux
+
+package server
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	"github.com/cri-o/cri-o/internal/log"
+)
+
+// validateNetworkNamespace checks if the given path is a valid network namespace.
+func (s *Server) validateNetworkNamespace(netnsPath string) error {
+	netns, err := ns.GetNS(netnsPath)
+	if err != nil {
+		return fmt.Errorf("invalid network namespace: %w", err)
+	}
+
+	defer netns.Close()
+
+	return nil
+}
+
+// cleanupNetns removes a network namespace file and logs the action.
+func (s *Server) cleanupNetns(ctx context.Context, netnsPath string, sb *sandbox.Sandbox) {
+	if rmErr := os.RemoveAll(netnsPath); rmErr != nil {
+		log.Warnf(ctx, "Failed to remove netns path %s: %v", netnsPath, rmErr)
+	} else {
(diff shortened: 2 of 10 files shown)


danwinship:
After a reboot there would be no iptables/nftables hostport rules anyway, so I don't think this is the problem. (And I don't see anything about hostports in the linked bug, though I didn't pull the full logs from the Google drive link...)

server/sandbox_network.go
```diff
@@ -173,17 +173,38 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
+		if err := s.hostportManager.Remove(sb.ID(), sb.PortMappings()); err != nil {
+			log.Warnf(ctx, "Failed to remove hostport for pod sandbox %s(%s): %v",
+				sb.Name(), sb.ID(), err)
+		}
+
+		return sb.SetNetworkStopped(ctx, true)
+	}
+
+	// Check if the network namespace file exists
+	if _, statErr := os.Stat(podNetwork.NetNS); statErr != nil {
```
haircommander: maybe this hsould be pulled in the err handling above? I'm wondering if there are cases we fail to create the pod network but want to keep the hostport rules in place.. If we use this Stat as the way to handle error then we know the creation will continue to fail because we never recreate the pod namespaces.

sohankunkerkar: >maybe this hsould be pulled in the err handling above?

If `newPodNetwork` throws an error then we can't assure that `podNetwork` will non-nil and that might create a panic when we try to access `podNetwork.NetNS`.

>I'm wondering if there are cases we fail to create the pod network but want to keep the hostport rules in place.. If we use this Stat as the way to handle error then we know the creation will continue to fail because we never recreate the pod namespaces.

Yeah, that's a valid concern. How about we check for `NetNS` value and based on the perform the deletion? I pushed the changes. Let me know how do you feel.

haircommander:
/approve

it's possible you'd be able to make a integration test by killing a pod and unmounting its netns, and restarting crio. may be annoying and not worth the specific case, but worth a shot I think

sohankunkerkar:
e2e tests are flaking due to https://github.com/kubernetes/kubernetes/issues/131545

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

bitoku:
@sohankunkerkar ci-fedora-kata failed at the new test.

haircommander:
/approve

LGTM

@cri-o/cri-o-maintainers PTAL

server/sandbox_network.go
```diff
@@ -181,23 +181,47 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
+		if podNetwork.NetNS != "" {
+			// if staterr != nil then netns file does not exist.
+			if _, statErr := os.Stat(podNetwork.NetNS); statErr == nil {
+				// The netns file exists, which means it might be corrupted
+				s.removeInvalidNetns(ctx, podNetwork.NetNS, sb)
+			}
 		}
 
-		log.Warnf(ctx, "Removed invalid netns path %s from pod sandbox %s(%s)", podNetwork.NetNS, sb.Name(), sb.ID())
+		// Return the error so the retry logic can handle it.
```
bitoku: Seems redundant.
And it's not necessarily called by ResourceCleaner, nor retried.

server/sandbox_network.go
```diff
@@ -181,23 +181,47 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
+		// Log the error and return it so the retry logic can handle it
+		// This allows the ResourceCleaner to retry when CNI plugin becomes available.
+		log.Warnf(ctx, "Failed to destroy network for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
+
+		// If the network namespace exists but CNI teardown failed, try to clean it up.
+		if podNetwork.NetNS != "" {
+			// if staterr != nil then netns file does not exist.
+			if _, statErr := os.Stat(podNetwork.NetNS); statErr == nil {
+				// The netns file exists, which means it might be corrupted
+				s.removeInvalidNetns(ctx, podNetwork.NetNS, sb)
```
bitoku: I think in this case the netns is not necessarily invalid. Is it?

server/sandbox_network.go
```diff
@@ -181,23 +181,47 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
+	}
+
+	if err := s.config.CNIPlugin().TearDownPodWithContext(stopCtx, podNetwork); err != nil {
+		// Log the error and return it so the retry logic can handle it
+		// This allows the ResourceCleaner to retry when CNI plugin becomes available.
+		log.Warnf(ctx, "Failed to destroy network for pod sandbox %s(%s): %v", sb.Name(), sb.ID(), err)
+
+		// If the network namespace exists but CNI teardown failed, try to clean it up.
+		if podNetwork.NetNS != "" {
+			// if staterr != nil then netns file does not exist.
```
bitoku: Seems redundant.

server/sandbox_network.go
```diff
@@ -181,23 +181,47 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
+				podNetwork.NetNS, sb.Name(), sb.ID(), validateErr)
+			s.removeInvalidNetns(ctx, podNetwork.NetNS, sb)
+
+			return sb.SetNetworkStopped(ctx, true)
+		}
+	}
+
+	if err := s.config.CNIPlugin().TearDownPodWithContext(stopCtx, podNetwork); err != nil {
+		// Log the error and return it so the retry logic can handle it
+		// This allows the ResourceCleaner to retry when CNI plugin becomes available.
```
bitoku: It's not necessarily called by ResourceCleaner, so this comment is confusing.

server/sandbox_network.go
```diff
@@ -181,23 +181,47 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
+				podNetwork.NetNS, sb.Name(), sb.ID())
+
+			return sb.SetNetworkStopped(ctx, true)
 		}
 
-		// The netns file may still exists, which means that it's likely
-		// corrupted. Remove it to allow cleanup of the network namespace:
-		if rmErr := os.RemoveAll(podNetwork.NetNS); rmErr != nil {
-			return fmt.Errorf("%w: failed to remove netns path: %w", retErr, rmErr)
+		// Validate that the network namespace file is actually a valid network namespace
```
bitoku: Seems redundant.

server/sandbox_network.go
sohankunkerkar: Yeah, it's more for tracking things but I can remove that if it doesn't serve purpose.

server/sandbox_network.go
sohankunkerkar: sure

server/sandbox_network.go
sohankunkerkar: Yeah, right. Actually, it's not always called by `ResourceCleaner`. The retry logic is handled at a higher level (ResourceCleaner wrapper). I'm happy to remove that comment if this is confusing.

server/sandbox_network.go
sohankunkerkar: I think this was the assumption from my side but it may not be the case. It could be one of the reasons but not the only reason.

bitoku: If `removeInvalidNetns` is called for valid ns too, let's make this name more general like `removeNetns` and change the logs accordingly.

sohankunkerkar: I'm using `cleanupNetns` more generic one.

server/sandbox_network_linux.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// validateNetworkNamespace checks if the given path is a valid network namespace.
+func (s *Server) validateNetworkNamespace(netnsPath string) error {
+	netns, err := ns.GetNS(netnsPath)
+	if err != nil {
+		return fmt.Errorf("invalid network namespace: %w", err)
+	}
+
+	defer netns.Close()
```
saschagrunert, line 24: I'd remove the `defer` here.

server/sandbox_network.go
```diff
@@ -181,23 +181,42 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create pod network for sandbox %s(%s): %w", sb.Name(), sb.ID(), err)
 	}
 
-	if err := s.config.CNIPlugin().TearDownPodWithContext(stopCtx, podNetwork); err != nil {
-		retErr := fmt.Errorf("failed to destroy network for pod sandbox %s(%s): %w", sb.Name(), sb.ID(), err)
-
+	// Check if the network namespace file exists and is valid before attempting CNI teardown.
+	// If the file doesn't exist or is invalid, skip CNI teardown and mark network as stopped.
```
aojea, line 188: We had IP leaks on containerd because of a similar change that ignored calling the network CNI plugin on some cases, causing the CNI to no release the IP address https://github.com/containerd/containerd/pull/12132

I can not see this is similar or not, but @danwinship you should ensure that in this scenario the plugin gets the CNI DEL to release any resource associated to the Pod

sohankunkerkar: @aojea Thanks for calling that out. I can see the potential issue there. Let me go ahead and fix the issue.

PR #9329, opened 2025-07-09
Added container_file_descriptors metrics

sreeram-venkitesh:
#### What type of PR is this?

/kind feature

This PR adds the `container_file_descriptors` metric to CRI-O which reports the number of open file descriptors for each container. 

Part of https://github.com/cri-o/cri-o/issues/9125.

I've created `internal/lib/stats/process_metrics.go` and added a new metric descriptor for process metrics. I was able to test this locally. I still need work on the tests, but if this is in the right direction, I'd like to work on the rest of the process metrics as well.

```release-note
Added new metric container_file_descriptors to expose the number of open file descriptors for each container from CRI-O metrics
```

### DIFF
--- a/internal/lib/stats/descriptors.go
+++ b/internal/lib/stats/descriptors.go
@@ -149,3 +149,12 @@ var (
 		LabelKeys: baseLabelKeys,
 	}
 )
+
+// Process metrics
+var (
+	containerFileDescriptors = &types.MetricDescriptor{
+		Name:      "container_file_descriptors",
+		Help:      "Number of open file descriptors for the container",
+		LabelKeys: baseLabelKeys,
+	}
+)
--- a/internal/lib/stats/stats_server_linux.go
+++ b/internal/lib/stats/stats_server_linux.go
@@ -263,6 +263,16 @@ func (ss *StatsServer) containerMetricsFromCgStats(sb *sandbox.Sandbox, c *oci.C
 			metrics = append(metrics, oomMetrics...)
 		case NetworkMetrics:
 			continue // Network metrics are collected at the pod level only.
+		case ProcessMetrics:
+			pid, err := c.Pid()
+			if err != nil {
+				log.Errorf(ss.ctx, "Unable to fetch process ID for container %s: %v", c.ID(), err)
+
+				continue
+			}
+			if processMetrics := generateSandboxProcessMetrics(sb, pid); processMetrics != nil {
+				metrics = append(metrics, processMetrics...)
+			}
 		default:
 			log.Warnf(ss.ctx, "Unknown metric: %s", m)
 		}
--- a/internal/lib/stats/metrics.go
+++ b/internal/lib/stats/metrics.go
@@ -15,6 +15,7 @@ const (
 	MemoryMetrics  = "memory"
 	NetworkMetrics = "network"
 	OOMMetrics     = "oom"
+	ProcessMetrics = "process"
 )
 
 type metricValue struct {
@@ -84,6 +85,9 @@ func (ss *StatsServer) PopulateMetricDescriptors(includedKeys []string) map[stri
 		},
 		OOMMetrics: {
 			containerOomEventsTotal,
+		},
+		ProcessMetrics: {
+			containerFileDescriptors,
 		},
 	}
 
--- a/internal/lib/stats/process_metrics.go
+++ b/internal/lib/stats/process_metrics.go
@@ -1,0 +1,32 @@
+package statsserver
+
+import (
+	"os"
+	"fmt"
+
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+)
+
+func generateSandboxProcessMetrics(sb *sandbox.Sandbox, pid int) []*types.Metric {
+	processMetrics := []*containerMetric{
+		{
+			desc: containerFileDescriptors,
+			valueFunc: func() metricValues {
+				fdDir := fmt.Sprintf("/proc/%d/fd", pid)
+				entries, err := os.ReadDir(fdDir)
+				if err != nil {
+					return metricValues{}
+				}
+
+				return metricValues{{
+					value: uint64(len(entries)),
+					metricType: types.MetricType_GAUGE,
+				}}
+			},
+		},
+	}
+
+	return computeSandboxMetrics(sb, processMetrics, "process")
+}


bitoku:
@sreeram-venkitesh Thank you! Can you add e2e test in `test/cri-metrics.bats`?

sreeram-venkitesh:
@bitoku The `ci/prow/e2e-aws-ovn` failure doesn't seem related to the change? Please correct me if otherwise. I tried retesting but it failed again. I believe [this is the log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/9329/pull-ci-cri-o-cri-o-main-e2e-aws-ovn/1944116427808051200/build-log.txt) for the job.

bitoku:
@sreeram-venkitesh yeah, the infra of prow is flaky right now.

internal/lib/stats/stats_server_linux.go
```diff
@@ -263,6 +263,17 @@ func (ss *StatsServer) containerMetricsFromCgStats(sb *sandbox.Sandbox, c *oci.C
...
+			pid, err := c.Pid()
+			if err != nil {
+				log.Errorf(ss.ctx, "Unable to fetch process ID for container %s: %v", c.ID(), err)
+
+				continue
+			}
+			
+			if processMetrics := generateSandboxProcessMetrics(sb, pid); processMetrics != nil {
+				metrics = append(metrics, processMetrics...)
+			}
```
bitoku, line 284: I'd want to extend `ContainerStats` in each runtime to return stats based on `/proc` in addition to cgroup stats.
This is how cgroup stats do, and we don't have to check container liveness here.

https://github.com/cri-o/cri-o/blob/740ea78286cbd4206dd67164ec140082adf50813/internal/oci/runtime_oci.go#L1274-L1283

bitoku: For example, runtimeVM needs extra care about stats.
You can see how runtimeVM does here.
https://github.com/cri-o/cri-o/blob/740ea78286cbd4206dd67164ec140082adf50813/internal/oci/runtime_vm.go#L1024-L1062

I don't think we always need to adapt stats for runtimeVM when we add new stats, but we should keep it extendable.
The container's initPid is vm's pid on host OS, so it's not the same as the actual container's fds.

sreeram-venkitesh: @bitoku Please take a look at https://github.com/cri-o/cri-o/pull/9329/commits/e3e8cdbfa567f67b1fba2c27dbcf9f4619058f68. I've only implemented runtimeOCI since I wanted to confirm if I'm on the right path before implementing the file descriptor logic for other runtimes. I realize that these changes conflicts with https://github.com/cri-o/cri-o/pull/9366. In #9366, @gavinkflam has used the existing `PidsStats` to create ProcessMetrics. 

I'm okay with adding file descriptors and the rest of the process metrics under `PidsStats` if it makes more sense.

sreeram-venkitesh: Mainly I want to know if adding the logic inside `ContainerStats` in `internal/oci/runtime_oci.go` is correct.

https://github.com/sreeram-venkitesh/cri-o/blob/e3e8cdbfa567f67b1fba2c27dbcf9f4619058f68/internal/oci/runtime_oci.go#L1265-L1287

gavinkflam: I think these fd metrics are best implemented in the upstream libraries - https://github.com/opencontainers/cgroups and https://github.com/containerd/cgroups.

We are already relying on upstream for the other metrics. Having them in upstream can reduce the maintenance surface of CRI-O and benefit other projects too.

Would you be open to submitting PRs to the upstream repositories and integrating them into CRI-O once they’re merged?

sreeram-venkitesh: Hmm I had thought of this when looking at how cgmgr was using `*cgroups.Stats` to get MemoryStats and CpuStats and so on.

If we go with this approach we would be able to call `pids.FileDescriptors` directly without having to implement the logic in `ContainerStats` in the runtimes. This is how [you're doing](https://github.com/cri-o/cri-o/pull/9366/files#diff-543c3362e33e4488db8b637eae32a82d7b9dd1a980f4109469a4494f7c6d2e8cR16) it for the container_processes metric in your PR right?

@bitoku what do you think of this? I'm okay with implementing this in cgroups if that's the more suitable place.

bitoku: I don't think we should implement it in upstream cgroups libraries, because fd metric can't be calculated from cgroups, but /proc.
They use /proc a bit, but it's for supplementary usage.
https://github.com/opencontainers/cgroups/blob/a3e2ecd1f756a19cee15f85b96337a59c3b5337b/fs2/memory.go#L105-L113

@gavinkflam 
Is there any other case they use /proc data to calculate the stats?

gavinkflam: That's a good point. The only /proc data usage of upstreams metrics is the root cgroup memory metrics you mentioned.

internal/oci/runtime_oci_linux.go
```diff
@@ -1,10 +1,13 @@
+//go:build linux
+
```
bitoku: I don't think it's needed. because it has a file suffix.

sreeram-venkitesh: I've removed it. Should I remove it from `runtime_oci_unsupported.go` too? My editor was throwing an `method runtimeOCI.getContainerFileDescriptors already declared in runtime_oci_linux.go` error when I don't have the tag in `runtime_oci_unsupported.go`.

bitoku: no you shouldn't, because `unsupported` isn't a builtin build tag, and should be defined explicitly.

sreeram-venkitesh:
@bitoku Looking at the logs of the CI failures they all seem to be related to exceeding timeouts. Would rerunning the jobs be a good option?

sreeram-venkitesh:
@bitoku `make shellcheck` was failing earlier with the following error and was cancelling other tests.
```
In test/cri-metrics.bats line 260:
	file_descriptors=$(ls "/proc/$container_pid/fd" | wc -l)
                           ^--------------------------^ SC2012 (info): Use find instead of ls to better handle non-alphanumeric filenames.
```
I've fixed it in the latest commit.

sreeram-venkitesh:
I've been trying to debug this today. There was a potential deadlock happening with using `c.Pid()` instead of `c.pid()`, which I had updated in https://github.com/cri-o/cri-o/pull/9329/commits/f3c2701baa13e09992fc573a57c0ea72c2e6a128. Trying to run the test locally now..

sreeram-venkitesh:
The tests which were failing earlier seems to be passing now, but after talking with @haircommander last day, I'll be refactoring this PR to fetch all PIDs from cgroup.procs and also to add all the process related metrics we need for the KEP.

haircommander:
FYI I recommend you rebase on https://github.com/cri-o/cri-o/pull/9535/commits/f69d4871ffd1efa32c0cea6c4d5a2c8352c53fdc and incorporate it also

internal/config/cgmgr/stats_linux.go
```diff
@@ -119,17 +130,90 @@ func libctrManager(cgroup, parent string, systemd bool) (cgroups.Manager, error)
...
+					if len(fields) >= 1 {
+						if softLimit, err := strconv.ParseUint(fields[0], 10, 64); err == nil {
+							ulimitsSoft += softLimit
+						}
+					}
+				}
+			}
+		}
+	}
+
```
haircommander: this may feel cleaner if it was factored into a helper function, similar to cgroupMemStats or cgroupCpuStats.

haircommander: also noting the inspiration from cadvisor code would be good I think

test/cri-metrics.bats
```diff
@@ -256,6 +256,14 @@ EOF
 	# assert container_processes == 3
 	metrics_container_processes=$(echo "$metrics" | jq 'select(.name == "container_processes") | .value.value | tonumber')
 	[[ $metrics_container_processes == "3" ]]
+
+	container_pid=$(crictl inspect "$CONTAINER_ID" | jq -r '.info.pid')
+	file_descriptors=$(find "/proc/$container_pid/fd" | wc -l)
```
haircommander: missing some additional tests for other metrics added here

internal/lib/stats/descriptors.go
```diff
@@ -161,11 +161,36 @@ var (
 
 // Process metrics.
 var (
+	containerFileDescriptors = &types.MetricDescriptor{
+		Name:      "container_file_descriptors",
+		Help:      "Number of open file descriptors for the container",
```
haircommander: cadvisor is unfortunately inconsistent here, but in the case of this metric, adds a period at the end. let's match the help text for all of these

internal/lib/stats/descriptors.go
```diff
@@ -161,11 +161,36 @@ var (
...
+	}
+	containerThreadsMax = &types.MetricDescriptor{
+		Name:      "container_threads_max",
+		Help:      "Maximum number of threads allowed inside the container",
+		LabelKeys: baseLabelKeys,
+	}
+	containerUlimitsSoft = &types.MetricDescriptor{
+		Name:      "container_ulimits_soft",
+		Help:      "Soft ulimit values for the container root process. Unlimited if -1, except priority and nice",
+		LabelKeys: baseLabelKeys,
```
haircommander: I think this should have the ulimit name as an extra label

haircommander:
thanks for this! I like the direction, I have some remaining notes. would you also squash your commits together please :)

haircommander:
@cri-o/cri-o-maintainers this should be ready for review, I have taken over getting the tests to pass and fixed some small nits I had in the meantime

haircommander:
@cri-o/cri-o-maintainers PTAL

internal/config/cgmgr/stats_linux.go
```diff
@@ -238,3 +257,89 @@ func isMemoryUnlimited(v uint64) bool {
...
+	return &PidsStats{
+		Current:         stats.PidsStats.Current,
+		Limit:           stats.PidsStats.Limit,
+		Pids:            pids,
+		FileDescriptors: fdCount,
+		Sockets:         socketCount,
+		Threads:         stats.PidsStats.Current,
+		ThreadsMax:      stats.PidsStats.Limit,
```
saschagrunert, line 277: We removed the original logic to obtain the threads in cbea275. Is that correct? Now we count the number of PIDs, not the threads.

sreeram-venkitesh: After discussing with @haircommander, we decided to copy cadvisor behaviour, which was the reasoning behind this. @haircommander What do you think?

haircommander: I think they're functionally considered the same here, and specifically I wanted to go bug for bug compatible with https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L909-L910

PR #9333, opened 2025-07-10
Do nothing when the conmon is already confirmed dead.

bitoku:
#### What type of PR is this?

/kind bug

Don't log errors or emit metrics when cri-o already confirmed conmon is dead.

```release-note
Fix the bug where it continues to emit a metric after it confirms the conmon is stopped once
```


internal/oci/runtime_oci.go
```diff
@@ -1755,13 +1755,20 @@ func (r *runtimeOCI) ProbeMonitor(ctx context.Context, c *Container) error {
 		return nil
 	}
 
+	if c.monitorProcessDead {
```
haircommander: should we also just unset monitorProcess and then the block above will run?

bitoku: Thanks, I changed to unset it.

saschagrunert (approved):

PR #9335, opened 2025-07-11
Update golangci-lint and config

saschagrunert:
#### What type of PR is this?

/kind cleanup

Update linter and its configuration.

None
None

```release-note
None
```


.golangci.yml
```diff
@@ -14,6 +14,7 @@ run:
 linters:
   default: none
   enable:
+    - arangolint
```
bitoku, line 17: Not sure if we should disable it, but I don't think we need this.
https://github.com/Crocmagnon/arangolint

bitoku (commented):
/lgtm
/hold
Can we wait for a while not to make conflicts? I have some things I want to merge.

.golangci.yml
saschagrunert: It reports no issues so I'm wondering why we should disable it now. If we ever have any code path in CRI-O triggering the linter then we would catch issues early-on, even if it's unlikely.

saschagrunert:
/override ci/prow/e2e-gcp-ovn

Unrelated.

PR #9344, opened 2025-07-15
feat: Add Disk Metrics

R3hankhan123:
#### What type of PR is this?
/kind feature

This PR implements the previously missing disk metrics:
- `container_fs_inodes_free`
- `container_fs_inodes_total`
- `container_fs_limit_bytes`
- `container_fs_usage_bytes`
contributes towards #9125

None
No

```release-note
Added Disk metrics
```


internal/lib/stats/disk_metrics.go
```diff
@@ -0,0 +1,50 @@
...
+	"syscall"
+
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	"github.com/cri-o/cri-o/internal/oci"
+)
+
+// generateSandboxDiskMetrics computes filesystem disk metrics for a container sandbox.
+func generateSandboxDiskMetrics(sb *sandbox.Sandbox, ctr *oci.Container) []*types.Metric {
```
haircommander: where are we calling this? I think it could use an integration test

R3hankhan123: Had forgotten to add that file during git add😅, now have added the function call

bitoku:
Thanks @R3hankhan123 . Can you add e2e tests in test/cri-metrics.bats?

R3hankhan123:
> Thanks @R3hankhan123 . Can you add e2e tests in test/cri-metrics.bats?

Done @bitoku

R3hankhan123:
@bitoku can you help me debug the prow tests, i see that container is being exited but can't find the reason why exactly its exiting

bitoku:
@R3hankhan123 
See the path of the URL to know where the log is.
For integration tests
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/9344/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1945473469558820864/artifacts/fedora-integration/cri-o-gather/artifacts/testout.txt

I think other e2e tests come from infra issues, not because of this change.

R3hankhan123:
@bitoku , looks like port 9090 is being used when trying to run disk metrics but its already in use, should i kill the process that is using the port (ie memory metrics crio) wdyt

internal/lib/stats/disk_metrics.go
```diff
@@ -0,0 +1,51 @@
...
+	mountpoint := ctr.MountPoint()
+
+	var fsStats syscall.Statfs_t
+	if err := syscall.Statfs(mountpoint, &fsStats); err != nil {
+		return nil
+	}
+	// Calculate usage as total blocks minus free blocks
+	usageBytes := (uint64(fsStats.Blocks) - uint64(fsStats.Bfree)) * uint64(fsStats.Bsize)
```
bitoku: I think this should be calculated in each runtime, not here.
Same as https://github.com/cri-o/cri-o/pull/9329/files#r2209874133

bitoku:
@R3hankhan123 You can use `free_port`.
https://github.com/cri-o/cri-o/blob/7c127d30ac7c71f9d9385d7786e62640f1d8f3ea/test/conmon.bats#L17-L20

internal/lib/stats/disk_metrics.go
R3hankhan123: @bitoku even usageBytes?

R3hankhan123: Done

R3hankhan123:
Hi @bitoku, is the test flaky or its failing as a result of the tests I have implemented as I am unable to pin point the error

R3hankhan123:
Hi @haircommander @bitoku 
Please take a look when you're free. If all looks good, could you approve the pending workflows too? Thanks!

internal/lib/stats/disk_metrics.go
bitoku: I mean we should implement it in runtime_oci.go (for oci and pod runtime type) though you may need to extend the return type because diskstats are not cgroup stats.
https://github.com/cri-o/cri-o/blob/7ae2aa50402be6b2bfaa055385ee601c6577b048/internal/oci/runtime_oci.go#L1274-L1283
cgroup stats are calculated differently in oci/pod and vm runtime type. see https://github.com/cri-o/cri-o/pull/9329/files#r2210628276. 
I think other stats should be calculated differently among runtime types too, or at least keep them separately so that we can extend it later.

R3hankhan123: @bitoku I have some doubts, can you clarify them Do i have to extend for `runtimeVM` also as you have mentioned in another PR

R3hankhan123: @bitoku something like this needs to be done?
```
// ContainerStats provides statistics of a container.
func (r *runtimeOCI) ContainerStats(ctx context.Context, c *Container, cgroup string) (*cgmgr.CgroupStats, error) {
	_, span := log.StartSpan(ctx)
	defer span.End()

	c.opLock.Lock()
	defer c.opLock.Unlock()

	stats, err := r.config.CgroupManager().ContainerCgroupStats(cgroup, c.ID())
	if err != nil {
		return nil, err
	}

	mountpoint := c.MountPoint()
	if mountpoint != "" {
		var fsStats syscall.Statfs_t
		if err := syscall.Statfs(mountpoint, &fsStats); err == nil {
			usageBytes := (fsStats.Blocks - fsStats.Bfree) * uint64(fsStats.Bsize)
			limitBytes := fsStats.Blocks * uint64(fsStats.Bsize)

			stats.Disk = &cgmgr.DiskStats{
				InodesFree:  fsStats.Ffree,
				InodesTotal: fsStats.Files,
				LimitBytes:  limitBytes,
				UsageBytes:  usageBytes,
			}
		} else {
			log.Warnf(ctx, "failed to collect disk stats for container %s: %v", c.ID(), err)
		}
	}

	return stats, nil
}
```
Provided I have added the following 
```
type CgroupStats struct {
	Memory     *MemoryStats
	CPU        *CPUStats
	Hugetlb    map[string]HugetlbStats
	Pid        *PidsStats
	SystemNano int64
	Disk       *DiskStats
}

type DiskStats struct {
	InodesFree  uint64
	InodesTotal uint64
	LimitBytes  uint64
	UsageBytes  uint64
}
```

bitoku: > Do i have to extend for runtimeVM also as you have mentioned in another PR

Better to have it, but we can skip it for now. I think kata container needs changing.

> something like this needs to be done?

yes, but the disk stats is not cgroup stats. we shouldn't include them in CgroupStats.

R3hankhan123: @bitoku can please take a look when you are free, i have extended ContainerStats and waiting for your review, please and thank you

R3hankhan123:
@bitoku can please take a look when you are free, i have extended ContainerStats and waiting for your review, please and thank you

R3hankhan123:
Hi @bitoku @haircommander can you review the code, I had to make some changes as linter was failing😅. Please and thank you

bitoku:
/release-note-edit
```release-note
Added disk metrics (`container_fs_inodes_free`, `container_fs_inodes_total`, `container_fs_limit_bytes`, `container_fs_usage_bytes`)
```

R3hankhan123:
> Thanks! The changes look good to me. can you please fix the build failure for freebsd?

@bitoku for freebsd, some values in `Statfs_t` are int64 which are causing problems, initially I had made it so that all `Statfs_t` which are different in freebsd and linux are typecasted to `uint64` but as a result linter is failing. Do have any suggestions for this?

bitoku:
@R3hankhan123 You can use `*_linux.go`, `*_freebsd.go`, `*_unsupported.go` files and `//go:build` directive to separate code depending on OS or architecture. See other files in `oci` directory

R3hankhan123:
> @R3hankhan123 You can use `*_linux.go`, `*_freebsd.go`, `*_unsupported.go` files and `//go:build` directive to separate code depending on OS or architecture. See other files in `oci` directory

@bitoku just for clarification `runtime_oci_freebsd.go` has to be created with typecast specific to freebsd?

bitoku:
You may need to change runtime_oci_linux.go

R3hankhan123:
@bitoku I have made the changes, can you review it. Please and thank you

R3hankhan123:
@bitoku can you please take a look and approve any pending workflows 
Please and thank you

internal/lib/stats/stats_server_linux.go
```diff
@@ -347,3 +358,11 @@ func criProcessStats(pStats *cgmgr.PidsStats, systemNano int64) *types.ProcessUs
 		ProcessCount: &types.UInt64Value{Value: pStats.Current},
 	}
 }
+
+func criFilesystemStats(diskStats *cgmgr.DiskMetrics, systemNano int64) *types.FilesystemUsage {
+	return &types.FilesystemUsage{
+		Timestamp: systemNano,
+		FsId:      &types.FilesystemIdentifier{Mountpoint: "/"}, // Default mountpoint
```
bitoku: Can we always return "/"? It sounds wrong to me.

R3hankhan123:
@bitoku sorry for the inconvenience one minor typo had slipped in😅. Can you re trigger the workflows. Thank you

internal/oci/runtime_oci_linux.go
```diff
@@ -107,3 +111,33 @@ func (r *runtimeOCI) PortForwardContainer(ctx context.Context, c *Container, net
...
+	inodesTotal := stat.Files
+	inodesFree := stat.Ffree
+
+	return &cgmgr.DiskMetrics{
+		UsageBytes:  usageBytes,
+		LimitBytes:  limitBytes,
+		InodesTotal: inodesTotal,
+		InodesFree:  inodesFree,
+	}, nil
+}
```
bitoku: I think we should consider other filesystems, for example https://github.com/google/cadvisor/blob/21337093481db755f642fb5596edcde4222de0a6/fs/fs.go has several ways of calculation for each filesystem.
Could we import and resuse the fs.go?

R3hankhan123: Ok I will look into it

internal/lib/stats/stats_server_linux.go
R3hankhan123: So i can get mountpoint from `ctr.MountPoint()` and if its empty then return "/". That sounds better?

R3hankhan123:
Hi @bitoku , I have extended disk metrics to other filesystems by using `fs.go`, can you please take a look at it. Thank you

R3hankhan123:
Hi @bitoku i have fixed the lint and vendor issue🙂

internal/lib/stats/stats_server_linux.go
bitoku: I'm wondering in which case it lacks mountPoint. and I don't think falling back is a good idea... 
Did you refer to any other implementation?

R3hankhan123: @bitoku initially i thought it was better to fall back, but now I have made it so if it lacks mountPoint then nil is being returned

R3hankhan123:
Hi @bitoku , please let me know if you have any further concerns I can address.

internal/oci/runtime_oci_linux.go
```diff
@@ -107,3 +111,48 @@ func (r *runtimeOCI) PortForwardContainer(ctx context.Context, c *Container, net
...
+
+// getContainerDiskStats collects disk metrics for a container on Linux.
+func (r *runtimeOCI) getContainerDiskStats(c *Container) (*cgmgr.DiskMetrics, error) {
+	mountPoint := c.MountPoint()
+	if mountPoint == "" {
+		return nil, fmt.Errorf("container %s has no mount point", c.ID())
+	}
+
+	// Build a fs context:
+	fsContext := fs.Context{Crio: fs.CrioContext{}}
```
bitoku: Do we need to initialize this?

test/cri-metrics.bats
```diff
@@ -229,6 +232,65 @@ EOF
...
+	local new_fs_usage=0
+	for ((i = 0; i < timeout; i++)); do
+		new_fs_usage=$(crictl metricsp | jq '.podMetrics[0].containerMetrics[0].metrics[] | select(.name == "container_fs_usage_bytes") | .value.value | tonumber')
+		if [[ "$new_fs_usage" -gt "$fs_usage" ]]; then
+			break
+		fi
+		sleep 1
+	done
+
+	stop_crio
```
bitoku, line 350: This looks it won't fail even if there's no increase.
And I want the similar one for `container_fs_inodes_total`.

test/cri-metrics.bats
```diff
@@ -229,6 +232,65 @@ EOF
...
+	fs_inodes_free=$(echo "$metrics" | jq '.podMetrics[0].containerMetrics[0].metrics[] | select(.name == "container_fs_inodes_free") | .value.value | tonumber')
+	[[ "$fs_inodes_free" -gt 0 ]]
+
+	# assert container_fs_inodes_total is present
+	fs_inodes_total=$(echo "$metrics" | jq '.podMetrics[0].containerMetrics[0].metrics[] | select(.name == "container_fs_inodes_total") | .value.value | tonumber')
+	[[ "$fs_inodes_total" -gt 0 ]]
+
+	# Generate disk usage and validate increase
+	crictl exec --sync "$CONTAINER_ID" mkdir -p /var/lib/mydisktest
+	crictl exec --sync "$CONTAINER_ID" dd if=/dev/zero of=/var/lib/mydisktest/bloatfile bs=50M count=1
```
bitoku: nit: I think 50M is too much. just a few KB is fine.

internal/oci/runtime_oci_linux.go
R3hankhan123: its required to initialise `Fs.Context` struct implemented, otherwise it wont be able to fetch the metrics

bitoku: I mean what if we do `fsInfo, err := fs.NewFsInfo(fs.Context{})` and remove this line?

R3hankhan123: It will be the same but not to initialize for CRIO?

test/cri-metrics.bats
R3hankhan123: Maybe I add a variable to if block and if that variable set then it will be counted as successful?

R3hankhan123: For `container_fs_inodes_total` its just how many total inodes are there, not if there is an increase or decrease of inodes 🤔

test/cri-metrics.bats
R3hankhan123: Noted

test/cri-metrics.bats
R3hankhan123: Are you expecting something like  getting initial inode count, then creating some small files that consume inodes and see if there was a decrease or not?

bitoku: > For container_fs_inodes_total its just how many total inodes are there, not if there is an increase or decrease of inodes 🤔

If it doesn't work, you can use the other inode metric. I want to make sure the inode metrics also looking at the correct data.

internal/oci/runtime_oci_linux.go
bitoku: If we really need to initialize it, we should fill in the fields of CrioContext.

R3hankhan123:
```
type CrioContext struct {
	Root       string
	ImageStore string
	Driver     string
}
```

R3hankhan123:
```
Root: mountPoint,
ImageStore:  r.config.ImageStore,
Driver:     r.config.Storage,
```

R3hankhan123: Should i set this as these values

test/cri-metrics.bats
R3hankhan123: Noted

internal/oci/runtime_oci_linux.go
bitoku: I don't know tbh. I think we don't have to initialize it since those values are only used for labeling.
We should read the cadvisor code. 
If you think we don't need it, you should write `fsInfo, err := fs.NewFsInfo(fs.Context{})`.

test/cri-metrics.bats
R3hankhan123: Resolved!

test/cri-metrics.bats
R3hankhan123: Can you review the new function for `container_fs_inodes_total`

internal/oci/runtime_oci_linux.go
R3hankhan123: Hi @bitoku , here are my findings, the function [`NewFsInfo`](https://github.com/google/cadvisor/blob/85d18c6139a3f1fd48120c49ae44eac5aa352d86/fs/fs.go#L99C6-L99C15) calls this function [`addCrioImagesLabel`](https://github.com/google/cadvisor/blob/85d18c6139a3f1fd48120c49ae44eac5aa352d86/fs/fs.go#L300C22-L300C40) which  identifies and labels filesystem mount points that are used by CRI-O for storing container images and container data

R3hankhan123:
@bitoku i have address the review comments and added some comments also, can you please take a look, Thank you

internal/oci/runtime_oci_linux.go
```diff
@@ -107,3 +111,42 @@ func (r *runtimeOCI) PortForwardContainer(ctx context.Context, c *Container, net
...
 }
+
+// getContainerDiskStats collects disk metrics for a container on Linux.
+func (r *runtimeOCI) getContainerDiskStats(c *Container) (*cgmgr.DiskMetrics, error) {
+	mountPoint := c.MountPoint()
+	if mountPoint == "" {
+		return nil, fmt.Errorf("container %s has no mount point", c.ID())
+	}
+
+	fsInfo, err := fs.NewFsInfo(fs.Context{Crio: fs.CrioContext{Root: mountPoint, ImageStore: r.config.ImageStore, Driver: r.config.Storage}})
```
bitoku: Root is a storage root. 
https://github.com/google/cadvisor/blob/db33cc7562d90b4ae46d844f6b7c478a460d557d/container/crio/plugin.go#L43

test/cri-metrics.bats
```diff
@@ -229,6 +232,109 @@ EOF
...
+	for ((i = 0; i < timeout; i++)); do
+		new_fs_usage=$(crictl metricsp | jq '.podMetrics[0].containerMetrics[0].metrics[] | select(.name == "container_fs_usage_bytes") | .value.value | tonumber')
+		if [[ "$new_fs_usage" -gt "$fs_usage" ]]; then
+			found_increase=true
+			break
+		fi
+		sleep 1
+	done
+
+	if [[ "$found_increase" != true ]]; then
```
bitoku: it looks redundant. I think you can simplify this condition.

test/cri-metrics.bats
```diff
@@ -39,6 +39,7 @@ collection_period = 0
 included_pod_metrics = [
     "network",
     "cpu",
+	"disk",
```
bitoku: I think we remove these already. why did you add again?

bitoku (commented):
Just out of curiosity. Do you happen to use AI to generate a code?

R3hankhan123:
> Just out of curiosity. Do you happen to generate the code with AI?

Only for the test cases(not good at writing bat files and shell scripts 🥲😅, also its a new thing for me)

internal/oci/runtime_oci_linux.go
R3hankhan123: what value should i put?

test/cri-metrics.bats
R3hankhan123: I have added the disk part and during rebasing, only `hugetlb` metrics were added with the test part

R3hankhan123: I guess something must have went wrong during rebasing

internal/oci/runtime_oci_linux.go
bitoku: `config.Root`
https://github.com/cri-o/cri-o/blob/ab3f85e5fc00b55ca036cb0b61252310db5d9b0f/server/inspect.go#L51-L59

test/cri-metrics.bats
R3hankhan123: @bitoku now its fine?

R3hankhan123:
Hi @bitoku can you retrigger `[integration / critest / conmon-rs / runc / amd64 (pull_request)]`, please and thank you

R3hankhan123:
Hi @bitoku any more suggestions you have that I need to address, please and thank you

internal/lib/stats/descriptors.go
```diff
@@ -36,6 +36,30 @@ var (
 	}
 )
 
+// Disk metrics.
+var (
+	containerFsInodesFree = &types.MetricDescriptor{
+		Name:      "container_fs_inodes_free",
+		Help:      "Number of free inodes.",
+		LabelKeys: baseLabelKeys,
+	}
```
bitoku, line 46: we could add extra labels, but I found general discrepancies in labels between crio and cadvisor. we can ignore it for now.
https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/metrics/prometheus.go#L537

bitoku:
@cri-o/cri-o-maintainers PTAL

R3hankhan123:
Thanks @bitoku for the review, once the PR is approved I will start implementing the DISK IO metrics also☺️

R3hankhan123:
Hi @cri-o/cri-o-maintainers PTAL when you get a chance 🙏. Thanks!

R3hankhan123:
@bitoku rebasing removed the lgtm tag😅

bitoku:
@haircommander 
Since https://kep.k8s.io/2371 is going to be beta, we should enable the metrics listed in the [conformance test](https://github.com/kubernetes/kubernetes/blob/ee1ff4866e30ac3685da3e007979b0e9ab7651a6/test/e2e_node/container_metrics_test.go#L87-L101) in 1.35, shouldn't we?

haircommander:
> we should enable the metrics listed in the [conformance test](https://github.com/kubernetes/kubernetes/blob/ee1ff4866e30ac3685da3e007979b0e9ab7651a6/test/e2e_node/container_metrics_test.go#L87-L101) in 1.35, shouldn't we?

yeah I think so, but more importantly we should probably create a job that runs this with the KEP enabled for cri-o, as that test is currently only running alongside cadvisor AFAIU

internal/oci/runtime_oci_linux.go
```diff
@@ -107,3 +111,45 @@ func (r *runtimeOCI) PortForwardContainer(ctx context.Context, c *Container, net
...
 }
+
+// getContainerDiskStats collects disk metrics for a container on Linux.
+func (r *runtimeOCI) getContainerDiskStats(c *Container) (*cgmgr.DiskMetrics, error) {
+	mountPoint := c.MountPoint()
+	if mountPoint == "" {
+		return nil, fmt.Errorf("container %s has no mount point", c.ID())
+	}
+
+	fsInfo, err := fs.NewFsInfo(fs.Context{Crio: fs.CrioContext{Root: r.config.Root, ImageStore: r.config.ImageStore, Driver: r.config.Storage}})
```
haircommander: it's not totally clear to me: is this reading from the cadvisor that's embedded in the kubelet or is this recalculating on the fly?

R3hankhan123: Afaik it calculates on the FS info on the fly

haircommander:
OK sorry to be a pain and thanks for your patience here. I do have a number of problems with this structurally. Since you've been patient and hard working, I'm willing to merge this and have a cri-o maintainer take it on, or you can have the opportunity. Since we're approaching code freeze for upstream kube, we'll need to get this sooner rather than later, so I'll wanna hear ASAP if you wanna take this one.

First, I think the disk collection code shouldnt' go in the cgmgr. cgmgr should really stick to just managing cgroups, but this is collecting info on disks. I think there should be a separate interface we use for collecting disk stats.

I think using the cadvisor disk collection code is fine. I worry a bit about introducing the dependency if we're not benefiting from the cached data, but I'm actually pretty sure the cadvisor embedded in the kubelet isn't caching disk info anyway, but collecting it on the fly. 

Also: there are a number of stats we need on top of this. I think our complete list is described in https://github.com/kubernetes-sigs/cri-tools/pull/1931:
```
container_fs_inodes_free
container_fs_inodes_total
container_fs_io_current
container_fs_io_time_seconds_total
container_fs_io_time_weighted_seconds_total
container_fs_limit_bytes
container_fs_read_seconds_total
container_fs_reads_merged_total
container_fs_reads_total
container_fs_sector_reads_total
container_fs_sector_writes_total
container_fs_usage_bytes
container_fs_write_seconds_total
container_fs_writes_merged_total
container_fs_writes_total
```
So the advantage of using cadvisor is it's already collecting all of this info anyway. I think having an interface will allow us to plug in a different provider in the future potentially...

@R3hankhan123 let me know which of these you'd like to take on. Happy to let you take it, or take it over myself. Sorry this is coming late and now is kinda urgent, that's my bad

R3hankhan123:
Hi @haircommander when is the code freeze i am happy to work on all of them but if its near its better to divide and do

haircommander:
k8s code freeze is Nov 7 and I think these should be done before that

R3hankhan123:
Oh time is there, i am happy to do all of them, should i do it in the same PR or a different pr so that reviewing is easier

haircommander:
maybe do the first point first, and then add the additional metrics in follow-up?

haircommander:
thank you!

R3hankhan123:
For first point will try to make the changes by tomorrow

R3hankhan123:
@haircommander two fedora builds are failing becasue of go version mismatch 🥲

R3hankhan123:
Hi @haircommander can you review the changes, Please and thank you

internal/config/diskmgr/diskmgr_linux.go
```diff
@@ -0,0 +1,50 @@
...
+// New returns a new DiskManager instance
+func New() DiskManager {
+	return &diskManagerImpl{}
+}
+
+// Name returns the name of this disk manager
+func (d *diskManagerImpl) Name() string {
+	return CrioPrefix
+}
+
```
haircommander: I'm unsure we should add this until it's needed. also odd it's called crio when really we're using cadvisor code under the hood

haircommander:
the current state of the impl is confusing to me. We have statsmgr that calls into oci.Runtime which calls into the diskmgr. But oci.Runtime has cadvisor coded in, but then diskmgr is using cadvsior? I think this could use some simplification.

@fidencio @zvonkok is the container's rootfs for a kata container present on the parent system? I know we go through c/storage to initialize the rootfs, so could we just use the disk usage of the host path rather than calling into the shim? that would simplify things.

zvonkok:
> @fidencio @zvonkok is the container's rootfs for a kata container present on the parent system? I know we go through c/storage to initialize the rootfs, so could we just use the disk usage of the host path rather than calling into the shim? that would simplify things.

This will not work since the rootfs may be backed by memory, disk or other mediums that we cannot inspect on the host. Even if we could it would break the Confidential Containers use-case where the rootfs may not be available on the host in any form.

haircommander:
FYI I recommend you rebase on https://github.com/cri-o/cri-o/pull/9535/commits/f69d4871ffd1efa32c0cea6c4d5a2c8352c53fdc also

internal/lib/stats/stats_server_linux.go
```diff
@@ -68,14 +67,20 @@ func (ss *StatsServer) updateSandbox(sb *sandbox.Sandbox) *types.PodSandboxStats
...
 
 			continue
 		}
-		// Convert cgroups stats to CRI stats.
-		cStats := containerCRIStats(cgstats, c, cgstats.SystemNano)
+		diskStats, err := ss.Runtime().DiskStats(ss.ctx, c, sb.CgroupParent())
+		if err != nil {
+			log.Errorf(ss.ctx, "Error getting disk stats %s: %v", c.ID(), err)
+
+			continue
```
haircommander: I think we should drop the continues in either case and only continue if both diskSTats and ctrStats are nil

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/go-chi/chi/v5 v5.2.3
 	github.com/go-logr/logr v1.4.3
 	github.com/godbus/dbus/v5 v5.1.1-0.20230522191255-76236955d466
+	github.com/google/cadvisor v0.53.0
```
haircommander: are we still using this?

R3hankhan123: Nope, so this can be removed?

haircommander: yeah!

test/cri-metrics.bats
```diff
@@ -230,6 +233,106 @@ EOF
...
+collection_period = 0
+included_pod_metrics = [
+    "network",
+	"disk",
+    "memory",
+]
+EOF
+	start_crio_no_setup
+	check_images
+	sleep 10
```
haircommander: hm why do we need this?

R3hankhan123: The sleep? Without sleep the test was flaky

internal/oci/runtime_vm.go
```diff
@@ -1058,17 +1058,22 @@ func (r *runtimeVM) ContainerStats(ctx context.Context, c *Container, _ string)
...
 			return metricsV2ToCgroupStats(ctx, m), nil
+		} else {
+			return nil, fmt.Errorf("unknown stats type %T", stats)
 		}
 	}
+}
 
-	return nil, fmt.Errorf("unknown stats type %T", stats)
+func (r *runtimeVM) DiskStats(ctx context.Context, c *Container, _ string) (*DiskMetrics, error) {
+	return nil, errors.New("DiskStats not implemented for runtimeVM")
```
haircommander: let's return empty for now, it's failing a stats test
```
# time="2025-10-31T16:57:06.311922954Z" level=debug msg="RuntimeVM.ContainerStats() end" file="oci/runtime_vm.go:1068"
# time="2025-10-31T16:57:06.311971459Z" level=error msg="Error getting disk stats 090463bdd4766801bc675a1ce3e7c526fbeac7269bc11b9ec964b82bb1ec6ac5: DiskStats not implemented for runtimeVM" file="stats/stats_server_linux.go:76"
# panic: runtime error: invalid memory address or nil pointer dereference
# [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2619625]
#
# goroutine 168 [running]:
# github.com/cri-o/cri-o/internal/lib/stats.(*StatsServer).updateSandbox(0xc000349080, 0xc0009a4400)
# 	github.com/cri-o/cri-o/internal/lib/stats/stats_server_linux.go:89 +0xdc5
# github.com/cri-o/cri-o/internal/lib/stats.(*StatsServer).statsForSandbox(0xc000d80540?, 0xffd285?)
# 	github.com/cri-o/cri-o/internal/lib/stats/stats_server.go:176 +0x75
# github.com/cri-o/cri-o/internal/lib/stats.(*StatsServer).StatsForSandboxes(0xc000349080, {0xc000b8a068, 0x1, 0xc0009756f0?})
# 	github.com/cri-o/cri-o/internal/lib/stats/stats_server.go:164 +0x125
# github.com/cri-o/cri-o/server.(*Server).ListPodSandboxStats(0xc000581808, {0x3298ef8, 0xc000603b30}, 0xc000602b10)
# 	github.com/cri-o/cri-o/server/sandbox_stats_list.go:22 +0x13c
# k8s.io/cri-api/pkg/apis/runtime/v1._RuntimeService_ListPodSandboxStats_Handler.func1({0x3298ef8?, 0xc000603b30?}, {0x2cbad20?, 0xc000602b10?})
# 	k8s.io/cri-api@v0.35.0-alpha.2/pkg/apis/runtime/v1/api_grpc.pb.go:1134 +0xce
# main.main.func3.UnaryInterceptor.6({0x3298ef8, 0xc000602ae0}, {0x2cbad20, 0xc000602b10}, 0xc000b4a000, 0xc0000133f8)
# 	github.com/cri-o/cri-o/internal/log/interceptors/interceptors.go:66 +0x192
# k8s.io/cri-api/pkg/apis/runtime/v1._RuntimeService_ListPodSandboxStats_Handler({0x2ea55a0, 0xc000581808}, {0x3298ef8, 0xc000602ae0}, 0xc000a0af00, 0x2fdb160)
# 	k8s.io/cri-api@v0.35.0-alpha.2/pkg/apis/runtime/v1/api_grpc.pb.go:1136 +0x143
# google.golang.org/grpc.(*Server).processUnaryRPC(0xc00062b400, {0x3298ef8, 0xc000602900}, 0xc000d800c0, 0xc0003518c0, 0x443f678, 0x0)
# 	google.golang.org/grpc@v1.76.0/server.go:1431 +0xfe8
# google.golang.org/grpc.(*Server).handleStream(0xc00062b400, {0x3299898, 0xc000717520}, 0xc000d800c0)
# 	google.golang.org/grpc@v1.76.0/server.go:1842 +0xb95
# google.golang.org/grpc.(*Server).serveStreams.func2.1()
# 	google.golang.org/grpc@v1.76.0/server.go:1061 +0x7f
# created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 348
# 	google.golang.org/grpc@v1.76.0/server.go:1072 +0x11d
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/9344/pull-ci-cri-o-cri-o-main-ci-fedora-kata/1984296674733854720/artifacts/fedora-kata/cri-o-gather/artifacts/testout.txt

and I think it's better than skipping

internal/oci/container.go
```diff
@@ -62,6 +62,8 @@ type Container struct {
 	state              *ContainerState
 	opLock             sync.RWMutex // For handling the container OCI runtime state transitions (i.e: anything that needs to call OCI runtime)
 	metaLock           sync.RWMutex // For handling the internal oci.Container metadata. Currently only used for Spec and Resources fields
+	opLock             sync.RWMutex // For handling the container OCI runtime state transitions (i.e: anything that needs to call OCI runtime)
```
haircommander: this commit should be dropped, we're redefining these fields

R3hankhan123: Can you share the  commit sha?

R3hankhan123: Hey I guess you are looking at an older commit, i dont see this line in my changes

internal/lib/stats/descriptors.go
haircommander: @R3hankhan123 can you drop a TODO comment for this?

internal/oci/container.go
haircommander: d8b7965a7c5294da3f3f88287faae403c0f3bc2f is still in https://github.com/cri-o/cri-o/pull/9344/commits

R3hankhan123: Done

internal/lib/stats/descriptors.go
R3hankhan123: Done

haircommander:
/approve

@cri-o/cri-o-maintainers PTAL

saschagrunert (commented):
Code LGTM, but please fix the linting and space-at-eol CI.

R3hankhan123:
Done @saschagrunert , can you re trigger those pipelines, thank you

test/cri-metrics.bats
saschagrunert: Do you have a link to that flake? Other tests also do `metrics_setup` and don't require the sleep.

R3hankhan123: Here are the sets of passing/failing links in prow
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/cri-o_cri-o/9344/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1949442989914853376
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/cri-o_cri-o/9344/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1950081866799255552
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/cri-o_cri-o/9344/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1950120031215947776
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/cri-o_cri-o/9344/pull-ci-cri-o-cri-o-main-ci-fedora-integration/1950204897815171072

test/cri-metrics.bats
```diff
@@ -232,6 +233,106 @@ EOF
...
+
+	metrics_setup
+	set_container_pod_cgroup_root "" "$CONTAINER_ID"
+
+	sleep 5 # Allow disk stats collection
```
saschagrunert: We should create a function to query the stats and succeed once they're available. A simple sleep seems too risky to flakes.

test/cri-metrics.bats
saschagrunert: The failures seem to be down to line 284. Can we remove that sleep and retry the integration tests?

R3hankhan123: this line is a sync command that flushes filesystem buffers to disk after creating test files.

R3hankhan123: Better will be to create the query function that you have suggested

test/cri-metrics.bats
```diff
@@ -232,6 +233,109 @@ EOF
...
+    "memory",
+]
+EOF
+	start_crio_no_setup
+	check_images
+
+	metrics_setup
+	set_container_pod_cgroup_root "" "$CONTAINER_ID"
+
+	if ! wait_for_metric 60 "container_fs_usage_bytes"; then
```
R3hankhan123: @saschagrunert something like this are you expecting?

saschagrunert: Yeah, I guess it could be efen simpler: 
```suggestion
	wait_for_metric container_fs_usage_bytes
```
And do the rest (logging, exiting, default timeout) directly in the function/

R3hankhan123:
Hi @saschagrunert i see that you have wrote but i cant find it, so writing it here, we cannot re-use wait_for_metric.Then call `crictl metricsp | jq '.podMetrics[0].containerMetrics[0].metrics[] | select(.name == "container_fs_inodes_free") | .value.value | tonumber' ` afterwards as that loop check for change in value of metrics

PR #8931, opened 2025-01-22
Move to IRQBALANCE_BANNED_CPULIST 

Tal-or:
#### What type of PR is this?
/kind dependency-change

#### What this PR does / why we need it:
The IRQBALANCE_BANNED_CPUS variable is deprecated and will not be used anymore in the near future.

The newer IRQBALANCE_BANNED_CPULIST variable is already supported by irqbalanced (https://github.com/Irqbalance/irqbalance/blob/v1.9.0/cputree.c#L135) and should be used.

Implementing this requires new code for byte/hex formatted cpu mask to CPUSet formatting code.

#### Which issue(s) this PR fixes:
None

#### Does this PR introduce a user-facing change?

```release-note
None
```

## Summary by CodeRabbit

* **Refactor**
  * Internal rewrite to use CPU-list (cpuset) semantics across IRQ load‑balancing and CPU affinity flows; improved saving/restoring, error handling and logging; config now supports a CPULIST form for banned CPUs while public interfaces remain unchanged.
* **Tests**
  * Expanded unit and integration tests and helpers (including CPULIST conversion utilities and shell helpers) to validate CPULIST behavior, conversions, and save/restore/restart scenarios.

PR #9364, opened 2025-07-28
OCPBUGS-5_9_4_0_3: HighPerformanceHooks: Implement idempotent IRQ SMP affinity management with atomic locking

andreaskaris:
Addresses race conditions issues IRQ SMP affinity management for high-performance hooks.

1. Race conditions from dual locking: The prior patch used separate locks for IRQ SMP affinity and irqbalance configuration files, creating a window where concurrent operations could corrupt system state. This fix consolidates to a single atomic lock protecting the entire read-calculate-write cycle for both /proc/irq/default_smp_affinity and /etc/sysconfig/irqbalance.
2. Kubelet container replacement timing: Addresses scenarios where kubelet starts replacement containers before deleting old ones, leading to incorrect IRQ affinity calculations. The new idempotent approach recalculates complete state from all running containers rather than incremental updates, ensuring consistent results regardless of container lifecycle timing. In order to achieve this, it keeps a list of containers which passed PreStop, as crio doesn't do that.
3. systemctl restart limitations: While adding reference to PR8834's systemd StartLimitBurst workaround, the implementation recognizes that rapid systemctl restarts are inherently problematic (concurrent restarts are ignored while service is  restarting).

#### What type of PR is this?

/kind bug

Fix race condition in IRQ balance configuration

This is a follow-up to https://github.com/cri-o/cri-o/pull/9228 which did not fix all race conditions

##### One big fat lock 

The current approach assumes that we cannot rely on optimistic concurrency.

The section under mutex lock is pretty large and a bit too complex for my taste. No matter how much I think about it, I cannot remove the containerServer.ListContainers() operations from within the lock.

```
	// Moving ListContainers outside of the lock instead of inside calcIRQSMPAffinityCPUs under the lock.
	// a) On add operations, we don't list at all --> it's o.k. for add
	// b) On delete operations, the following potential race scenario should be ok thanks to the prestopping list.
	// 	  The potential race scenario without putting the list operation under a lock. Assuming X and Y share same
	//    static set:
	//    1. Thread A: Lock -> Container X enters prestopping -> Unlock
	//    2. Thread A: ListContainers() outside lock → sees container Y
	//    3. Thread B: Lock -> Container Y enters prestop, adds itself to in-prestopping -> Unlock
	//    4. Thread B: Lock -> updates IRQ files (without X, without Y) -> Unlock
	//    5. Thread B: container Y is deleted from cri-o
	//    6. Thread A: Lock -> calculates based on stale list including Y; however, prestopping contains Y,
	//                 updates IRQ files (without X, without Y) -> Unlock
        //   c) But on delete, the following races:
	// 	  The race scenario:
	//    1. Thread A: Lock -> Container X enters prestopping -> Unlock
	//    2. Thread A: ListContainers() outside lock → container Y not yet started
	//    3. Thread B: Container Y starts, locks, updates IRQ files, sets IRQs correctly, unlocks
	//    6. Thread A: Lock -> calculates based on stale list without Y;
	//                 updates IRQ files (without X, without Y) -> Unlock
	//    ---> RACE
```

The alternative is having crio manage all IRQ SMP CPUs and having it recalculate the entire set based on a timer. I.e., running a go func() when the highperformancehook is first initialized that checks and corrects want vs have affinity every second (?). That way, we could use optimistic concurrency on container prestart / prestop. But that would mean that crio manages all IRQ SMP CPUs and resets them to what it wants, using infra_ctr_cpuset to mark the unmanaged CPU set.

##### how to reproduce / smoke test

Here's a deployment which reproduces the race conditions:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: qos-demo
  name: qos-demo
spec:
  replicas: 3
  selector:
    matchLabels:
      app: qos-demo
  strategy: {}
  template:
    metadata:
      name: qos-demo
      labels:
        app: qos-demo
      annotations:
    spec:
      hostNetwork: true
      runtimeClassName: performance-sno-pp
      containers:
      - name: qos-demo-ctr-1
        command:
        - "/bin/sleep"
        - "infinity"
        resources:
          limits:
            memory: "100Mi"
            cpu: "2"
          requests:
            memory: "100Mi"
            cpu: "2"
        livenessProbe:
          exec:
            command:
            - cat
            - /tmp/healthy
          initialDelaySeconds: 5
          periodSeconds: 5
      - name: qos-demo-ctr-2
        command:
        - "/bin/sleep"
        - "infinity"
        resources:
          limits:
            memory: "100Mi"
            cpu: "2"
          requests:
            memory: "100Mi"
            cpu: "2"
        livenessProbe:
          exec:
            command:
            - cat
            - /tmp/healthy
          initialDelaySeconds: 1
          periodSeconds: 1
      - name: qos-demo-ctr-3
        command:
        - "/bin/sleep"
        - "infinity"
        resources:
          limits:
            memory: "100Mi"
            cpu: "2"
          requests:
            memory: "100Mi"
            cpu: "2"
      - name: qos-demo-ctr-4
        command:
        - "/bin/sleep"
        - "infinity"
        resources:
          limits:
            memory: "100Mi"
            cpu: "2"
          requests:
            memory: "100Mi"
            cpu: "2"
      - name: qos-demo-ctr-5
        command:
        - "/bin/sleep"
        - "infinity"
        resources:
          limits:
            memory: "100Mi"
            cpu: "2"
          requests:
            memory: "100Mi"
            cpu: "2"
```

Smoke test:
```
mkdir -p /etc/systemd/system/irqbalance.service.d
cat <<'EOF' > /etc/systemd/system/irqbalance.service.d/restart-limits.conf
# cri-o high performance runtime handler reconfigures and restarts irqbalance
# after every container with pinned cpus and moved interrupts is created.
# Systemd does not like this and will block further restarts when the
# StartLimitBurst (default 5) is crossed within the default timeout (10s).
# That is too strict as during system startup many such containers might
# be launched.
# Install this to /etc/systemd/system/irqbalance.service.d/restart-limits.conf
# to prevent systemd from blocking the necessary restarts.
[Unit]
StartLimitBurst=100
EOF
systemctl daemon-reload
```

```
mv /usr/sbin/irqbalance{,.back}
cat <<'EOF' > /usr/sbin/irqbalance
# cat /usr/sbin/irqbalance
#!/bin/bash

log_file=/tmp/irqbalance.mock.log

echo "$(env | tr '\n' ' ') $@" >> $log_file
/usr/sbin/irqbalance.back "$@"
EOF
chmod +x /usr/sbin/irqbalance.back
```

```
cat <<'EOF' > smoke.sh
#!/bin/bash

set -x

banned_file="/etc/sysconfig/irqbalance"
expected_affinity="3e0f"
expected_reset_banned="IRQBALANCE_BANNED_CPUS=\"0000c000\""
expected_banned="IRQBALANCE_BANNED_CPUS=\"0000c1f0\""
expected_irqbalance_oneshots=10
# The delete runs systemctl restart in such quick succession that it shows only once (or more).
expected_irqbalance_restarts=6

cat $affinity_file

check() {
	expected_a="$1"
	expected_b="$2"

	affinity=$(cat ${affinity_file} | tr -d '\n')
        echo "Got affinity: $affinity, expected affinity: $expected_a"
        if [ "${affinity}" != "${expected_a}" ]; then
            exit 1
        fi
	banned=$(grep -E "^IRQBALANCE_BANNED_CPUS=" ${banned_file} | tr -d '\n')
        echo "Got banned: '$banned', expected banned: '$expected_b'"
        if [ "${banned}" != "${expected_b}" ]; then
            exit 1
        fi
}

# irqbalance --oneshot
systemctl disable --now irqbalance
for i in {0..5}; do
	set +x
	echo "========"
	echo "Run ${i}"
	echo "========"
	set -x
	irqbalance_restarts_start=$(grep "oneshot" /tmp/irqbalance.mock.log | wc -l)

	kubectl apply -f pod.yaml
	echo "After set:"
	check "$expected_affinity" "$expected_banned"
        kubectl delete pod qos-demo
	echo "After reset:"
	check "$expected_reset_affinity" "$expected_reset_banned"

	irqbalance_restarts_end=$(grep "oneshot" /tmp/irqbalance.mock.log | wc -l)
	irqbalance_restarts=$((irqbalance_restarts_end - irqbalance_restarts_start))
	echo "Got $irqbalance_restarts irqbalance --oneshot runs"
	if [ "$irqbalance_restarts" != "$expected_irqbalance_oneshots" ]; then
	    exit 1
	fi
done

# systemctl restart irqbalance
systemctl enable --now irqbalance
for i in {0..5}; do
	set +x
	echo "========"
	echo "Run ${i}"
	echo "========"
	set -x
	irqbalance_restarts_start=$(journalctl -u irqbalance --boot | grep "Started irqbalance daemon" | wc -l)

	kubectl apply -f pod.yaml
	echo "After set:"
	check "$expected_affinity" "$expected_banned"
        kubectl delete pod qos-demo
	echo "After reset:"
	check "$expected_reset_affinity" "$expected_reset_banned"

	irqbalance_restarts_end=$(journalctl -u irqbalance --boot | grep "Started irqbalance daemon" | wc -l)
	irqbalance_restarts=$((irqbalance_restarts_end - irqbalance_restarts_start))
	echo "Got $irqbalance_restarts irqbalance restarts"
	if [ "$irqbalance_restarts" -lt "$expected_irqbalance_restarts" ]; then
	    exit 1
	fi
done
EOF
chmod +x smoke.sh
```

```
cat <<'EOF' > pod.yaml
apiVersion: v1
kind: Pod
metadata:
  name: qos-demo
  annotations:
spec:
  hostNetwork: true
  containers:
  - name: qos-demo-ctr-1
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-2
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-3
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-4
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-5
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
EOF
```

```
smoke.sh
```

```release-note
None
```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/cri-o/cri-o/internal/config/cgmgr"
 	"github.com/cri-o/cri-o/internal/config/node"
+	"github.com/cri-o/cri-o/internal/lib"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/oci"
@@ -50,6 +51,7 @@ const (
 	sysCPUDir            = "/sys/devices/system/cpu"
 	sysCPUSaveDir        = "/var/run/crio/cpu"
 	milliCPUToCPU        = 1000
+	preStoppingMaxLen    = 100
 )
 
 const (
@@ -113,6 +115,7 @@ type HighPerformanceHooks struct {
 	sharedCPUs               string
 	execCPUAffinity          config.ExecCPUAffinityType
 	irqSMPAffinityFile       string
+	preStoppingContainers    map[string]bool
 }
 
 func (h *HighPerformanceHooks) PreCreate(ctx context.Context, specgen *generate.Generator, s *sandbox.Sandbox, c *oci.Container) error {
@@ -199,7 +202,7 @@ func (h *HighPerformanceHooks) setExecCPUAffinity(ctx context.Context, specgen *
 	return nil
 }
 
-func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
+func (h *HighPerformanceHooks) PreStart(ctx context.Context, containerServer *lib.ContainerServer, c *oci.Container, s *sandbox.Sandbox) error {
 	log.Infof(ctx, "Run %q runtime handler pre-start hook for the container %q", HighPerformance, c.ID())
 
 	cSpec := c.Spec()
@@ -237,7 +240,7 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 	if shouldIRQLoadBalancingBeDisabled(ctx, s.Annotations()) {
 		log.Infof(ctx, "Disable irq smp balancing for container %q", c.ID())
 
-		if err := h.setIRQLoadBalancing(ctx, c, false); err != nil {
+		if err := h.setIRQLoadBalancing(ctx, containerServer, c, false); err != nil {
 			return fmt.Errorf("set IRQ load balancing: %w", err)
 		}
 	}
@@ -279,11 +282,14 @@ func (h *HighPerformanceHooks) PreStart(ctx context.Context, c *oci.Container, s
 	return nil
 }
 
-func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s *sandbox.Sandbox) error {
+func (h *HighPerformanceHooks) PreStop(ctx context.Context, containerServer *lib.ContainerServer, c *oci.Container, s *sandbox.Sandbox) error {
 	ctx, span := log.StartSpan(ctx)
 	defer span.End()
 
 	log.Infof(ctx, "Run %q runtime handler pre-stop hook for the container %q", HighPerformance, c.ID())
+
+	// Update prestopping list with current container and cleanup stale entries
+	h.updatePreStopping(ctx, containerServer, c)
 
 	cSpec := c.Spec()
 	if !shouldRunHooks(ctx, c.ID(), &cSpec, s) {
@@ -292,7 +298,7 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s 
 
 	// enable the IRQ smp balancing for the container CPUs
 	if shouldIRQLoadBalancingBeDisabled(ctx, s.Annotations()) {
-		if err := h.setIRQLoadBalancing(ctx, c, true); err != nil {
+		if err := h.setIRQLoadBalancing(ctx, containerServer, c, true); err != nil {
 			return fmt.Errorf("set IRQ load balancing: %w", err)
 		}
 	}
@@ -330,6 +336,50 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s 
 	}
 
 	return nil
+}
+
+// updatePreStopping adds the current container to the prestopping list and cleans up
+// stale entries by removing containers that are no longer present in the container list.
+// Uses the IRQ affinity lock to ensure atomic access with IRQ calculations.
+func (h *HighPerformanceHooks) updatePreStopping(ctx context.Context, containerServer *lib.ContainerServer, c *oci.Container) {
+	h.updateIRQSMPAffinityLock.Lock()
+	defer h.updateIRQSMPAffinityLock.Unlock()
+
+	// Initialize map if needed
+	if h.preStoppingContainers == nil {
+		h.preStoppingContainers = make(map[string]bool)
+	}
+
+	// Add current container to prestopping list
+	h.preStoppingContainers[c.ID()] = true
+
+	// Do not clean up if we haven't passed a threshold, yet.
+	if len(h.preStoppingContainers) < preStoppingMaxLen {
+		return
+	}
+
+	log.Debugf(ctx, "Running prestopping cleanup since %d above threshold (%d)", len(h.preStoppingContainers), preStoppingMaxLen)
+
+	// Clean up stale entries by getting current container list
+	containers, err := containerServer.ListContainers()
+	if err != nil {
+		log.Warnf(ctx, "Failed to list containers for prestopping cleanup: %v", err)
+
+		return
+	}
+
+	// Create a set of current container IDs for efficient lookup
+	currentContainerIDs := make(map[string]bool)
+	for _, container := range containers {
+		currentContainerIDs[container.ID()] = true
+	}
+
+	// Remove prestopping entries for containers that no longer exist
+	for containerID := range h.preStoppingContainers {
+		if !currentContainerIDs[containerID] {
+			delete(h.preStoppingContainers, containerID)
+		}
+	}
 }
 
 // If CPU load balancing is enabled, then *all* containers must run this PostStop hook.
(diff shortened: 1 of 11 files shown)


andreaskaris:
@haircommander @bitoku while backporting https://github.com/cri-o/cri-o/pull/9228 for  https://issues.redhat.com/browse/OCPBUGS-59321 (customer bug) I found that the race was fixed for /proc/irq/default_smp_affinity but not for /etc/sysconfig/irqbalance. Hence this follow up which should correct it. Jira: https://issues.redhat.com/browse/OCPBUGS-59403

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -301,7 +301,7 @@ func requestedSharedCPUs(annotations fields.Set, cName string) bool {
 
 // setCPULoadBalancing relies on the cpuset cgroup to disable load balancing for containers.
 // The requisite condition to allow this is `cpuset.sched_load_balance` field must be set to 0 for all cgroups
-// that intersect with `cpuset.cpus` of the container that desires load balancing.
+// that intersect with `cpuset.cpus` of the container that desires load balancing.// that intersect with `cpuset.cpus` of the container that desires load balancing.
```
bitoku: You seem to have forgotten a newline.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -581,69 +581,79 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
 	currentIRQSMPSetting := strings.TrimSpace(string(content))
 
 	newIRQSMPSetting, newIRQBalanceSetting, err := calcIRQSMPAffinityMask(cpus, currentIRQSMPSetting, enable)
 	if err != nil {
-		return "", err
+		return err
 	}
 
+	log.Debugf(ctx, "Container %q set %q: %q; %q: %q", cName,
+		h.irqBalanceConfigFile, newIRQSMPSetting,
```
bitoku: This is wrong probably?
```suggestion
		h.irqSMPAffinityFile, newIRQSMPSetting,
```

internal/runtimehandlerhooks/high_performance_hooks_test.go
```diff
@@ -820,6 +830,7 @@ var _ = Describe("high_performance_hooks", func() {
 
 			// create tests affinity file
 			err = os.WriteFile(irqSmpAffinityFile, []byte(flags), 0o644)
+			err = os.WriteFile(irqBalanceConfigFile, []byte(formatIRQBalanceBannedCPUs(bannedCPUFlags)), 0o644)
```
bitoku, line 1264: It's better to have `Expect(err).ToNot(HaveOccurred())` before this line.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: I fat fingered something, thanks!

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: indeed, thanks

internal/runtimehandlerhooks/high_performance_hooks_test.go
andreaskaris: thanks for pointing that out, I shouldn't have missed the first "Expect" line O_o

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -581,69 +580,83 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
-	return newIRQBalanceSetting, nil
-}
+	// Nothing else to do if irq balance config file does not exist.
+	if !fileExists(h.irqBalanceConfigFile) {
+		return nil
+	}
+
+	if err := updateIrqBalanceConfigFile(h.irqBalanceConfigFile, newIRQBalanceSetting); err != nil {
+		// Rollback IRQ SMP affinity file to maintain consistency
+		if rollbackErr := os.WriteFile(h.irqSMPAffinityFile, []byte(originalIRQSMPSetting), 0o644); rollbackErr != nil {
```
andreaskaris: added this rollback here

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -64,8 +64,7 @@ const (
 type HighPerformanceHooks struct {
 	irqBalanceConfigFile     string
 	cpusetLock               sync.Mutex
-	irqSMPAffinityFileLock   sync.Mutex
-	irqBalanceConfigFileLock sync.Mutex
+	updateIRQSMPAffinityLock sync.Mutex
```
andreaskaris, line 114: removed the superfluous second lock

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -581,69 +581,119 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
 	}
 
-	newIRQBalanceSetting, err := h.updateNewIRQSMPAffinityMask(lspec.Resources.CPU.Cpus, enable)
-	if err != nil {
+	if err := h.updateNewIRQSMPAffinityMask(ctx, c.Name(), lspec.Resources.CPU.Cpus, enable); err != nil {
 		return err
 	}
+	// Outside of the lock section, we can restart the irqbalance service or run irqbalance --oneshot command.
+	// handleIRQBalanceRestart will log errors but will not return them, as it is not critical for the pod to start.
+	h.handleIRQBalanceRestart(ctx)
```
andreaskaris, line 731: moved both irqbalance --oneshot and service restart outside of the lock section

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -61,6 +61,49 @@ const (
 	SharedCPUsEnvVar     = "OPENSHIFT_SHARED_CPUS"
 )
 
+// ServiceManager interface for managing system services.
```
andreaskaris, line 67: Added new interfaces so that we can mock service state & restarts and command looksups and runs.

andreaskaris:
@MarSik for review

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -581,69 +624,133 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
-	if isIrqConfigExists {
-		if err := h.updateIrqBalanceConfigFile(newIRQBalanceSetting); err != nil {
-			return err
+// handleIRQBalanceRestart handles - outside of the lock section - the restart of the irqbalance service or runs
+// irqbalance --oneshot command if the service is not enabled. The environment variable for irqbalance oneshot
+// is read from h.irqBalanceConfigFile (/etc/sysconfig/irqbalance) which is guaranteed to be in a consistent state
+// after the lock section in updateNewIRQSMPAffinityMask.
+func (h *HighPerformanceHooks) handleIRQBalanceRestart(ctx context.Context, cName string) {
+	// Nothing else to do if irq balance config file does not exist.
+	if !fileExists(h.irqBalanceConfigFile) {
```
MarSik, line 836: Is this potentially a race condition too? This place is not protected by the lock. On the other hand, if the file was created earlier, this is executed after and should return true. Nobody deletes the file. So very probably ok.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -581,69 +624,133 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if !strings.HasPrefix(scanner.Text(), irqBalanceBannedCpus+"=") {
+			continue
+		}
+
+		text := strings.TrimPrefix(scanner.Text(), irqBalanceBannedCpus+"=")
```
MarSik, line 879: We will support two different variables soon https://github.com/cri-o/cri-o/pull/8931. The old mask and the new cpu list. We need to make sure to adapt the second PR to be merged.

MarSik:
@andreaskaris You should probably reference a bug and we will have to sync the other irqbalance PR. Other than that this looks fine to me I think.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: Well if there are x containers, this will be called x times. And for each x_i, this will be called after the lock section in updateNewIRQSMPAffinityMask. Because the locked sections cannot run at the same time, we're guaranteed that handleIRQBalanceRestart will run at least once after takes_longest_to_finish(updateNewIRQSMPAffnityMask). However, now I'm a bit concerned about systemd. In my smoketests, I can see that systemd actually forgoes service restarts if they happen too often, and the reason seems to be:
```
# systemctl show irqbalance | grep StartLimit
StartLimitIntervalUSec=10s
StartLimitBurst=5
StartLimitAction=none
```
So yes, that's a problem. But this problem exists regardless if we put handleIRQBalanceRestart into the mutex lock or not

The reason why I kept it (and the oneshot) out of the locked section is that I don't feel comfortable having command exec calls within a mutex to avoid deadlocks

TL;DR: now I'm scratching my head for how we can make sure that irqbalance really restarts given the systemd configuration ..

andreaskaris: I moved the service restart into its own go function processIRQBalanceRestarts which is responsible for restarting and for irqbalance --oneshot, now. I also added a 100ms debouncer to avoid that restarts happen in very quick succession. In addition, I'm resetting the service counters to make sure that any systemd restart request actually restarts the service.

MarSik: Even though the debouncing and service counters are going in the proper direction, we actually solved that differently some time ago. This PR (applied downstream) configured systemd - https://github.com/cri-o/cri-o/pull/8834

The issue is real, I am just trying to find the cleanest way that does not overcomplicate the code.

andreaskaris: Ah sorry, I didn't know about that commit. I'll shelve the last commit, then, and go with the initial 2 commits to keep things simpler

andreaskaris: Hmm .. I just tried that ... (even running systemd reset-failed irqbalance in a loop in the background). Instead of the 10 expected restarts for 5 containers (create  / delete) I get between 6 and 7.

```
(ded118f3-e6ed-4a3f-94e1-dedd35d3bc53) [root@centos9 ~]# journalctl -f -u irqbalance | grep Started
Aug 04 06:12:21 centos9 systemd[1]: Started irqbalance daemon.
Aug 04 06:12:24 centos9 systemd[1]: Started irqbalance daemon.
Aug 04 06:12:26 centos9 systemd[1]: Started irqbalance daemon.
Aug 04 06:12:28 centos9 systemd[1]: Started irqbalance daemon.
Aug 04 06:12:31 centos9 systemd[1]: Started irqbalance daemon.
Aug 04 06:12:33 centos9 systemd[1]: Started irqbalance daemon.
```

That said, it might be just the fact that on container shutdown, the last restart is still in progress. I get something similar when I run this:
```
(ded118f3-e6ed-4a3f-94e1-dedd35d3bc53) [root@centos9 ~]# for i in {1..3}; do sleep 0.01; systemctl restart irqbalance & done
[1] 8406
[2] 8411
[3] 8415
```

```
# journalctl -f -u irqbalance | grep Started
Aug 04 06:20:49 centos9 systemd[1]: Started irqbalance daemon.
Aug 04 06:20:49 centos9 systemd[1]: Started irqbalance daemon.
```

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +631,129 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
 	}
 
-	newIRQBalanceSetting, err := h.updateNewIRQSMPAffinityMask(lspec.Resources.CPU.Cpus, enable)
-	if err != nil {
+	if err := h.updateNewIRQSMPAffinityMask(ctx, c.Name(), lspec.Resources.CPU.Cpus, enable); err != nil {
 		return err
 	}
+	// Outside of the lock section, we can restart the irqbalance service or run irqbalance --oneshot command.
+	// handleIRQBalanceRestart will log errors but will not return them, as it is not critical for the pod to start.
+	h.handleIRQBalanceRestart(ctx, c.Name())
```
andreaskaris: I think this here needs some more work / thinking about in a follow-up. We need to make sure that the restarts are properly handled
a) they must happen after the irq update.
b) they should be debounced in some way
Just for the record, here's my earlier, complex attempt at fixing it: https://github.com/cri-o/cri-o/commit/be853e1d400f5bd8ff53e57e80dbbb989f8e1f4d (removed from this PR in order to keep the code simple)

andreaskaris:
/hold This issue isn't only limited to smp irq affinity and irqbalance files not being written to in an atomic fashion. In addition, as discussed with Martin already, there might also be a problem with systemd service restarts. Though I don't want to address that part here. However, while looking at the customer logs, I actually identified that kubelet will not always send container create / delete in the correct order - on a restart, crio might spawn a new container before tearing down the old one (it's a matter of ms, but here it's significant). That's why I believe that this must be fixed by making the logic level driven. I'm currently working on it

andreaskaris:
I updated the PR description to lay out my reasoning for this patch.

I see either this non-optimistic concurrency approach, meaning that we need a pretty big mutex lock but no background go func() and we can set/unset CPU bits for the containers that go up/down.

Or I see an optimistic concurrency approach, without the big lock, where a single go func() runs in the background and recalculates and manages all IRQ SMP affinity CPUs, whenever a timer fires or whenever prestop / prestart add an event to its channel. We'd not need any explicit locks here and instead rely on channels (well, we'd still lock implicitly ;-) ).

For the second option, we'd by the way then be able to debounce and manage systemd restarts.

andreaskaris:
integration test is failing on:
```
# time="2025-08-08T14:13:41.681284601Z" level=debug msg="Accessing \"quay.io/crio/pause:go\" failed: reading manifest go in quay.io/crio/pause: received unexpected HTTP status: 502 Bad Gateway" file="docker/docker_image_src.go:104"
# time="2025-08-08T14:13:41.681406056Z" level=debug msg="Error pulling image quay.io/crio/pause:go: initializing source docker://quay.io/crio/pause:go: reading manifest go in quay.io/crio/pause: received unexpected HTTP status: 502 Bad Gateway" file="server/image_pull.go:249" id=a0fbebbb-f8bf-40bc-93b0-eb9c8a31f759 name=/runtime.v1.ImageService/PullImage
```

unless I'm wrong, that's infra and has nothing to do with the actual pr

haircommander:
> unless I'm wrong, that's infra and has nothing to do with the actual pr

correct!

andreaskaris:
The customer came back to us, and simply moving stuff to posthook does not seem to be a reliable solution:
Instead, the idempotent approach of this PR here fixes the issue.

1) The testing using https://github.com/cri-o/cri-o/pull/9364 was positive. The issue of disappearing IRQ cores was not observed.
2) The testing using https://github.com/cri-o/cri-o/pull/9405 was not positive. The issue was observed with this rpm installed.

PR #9372, opened 2025-07-31
server: ensure CNI teardown prevents IP leaks with missing netns

sohankunkerkar:
This fix leverages CNI cache files to retrieve network attachments and call CNI DEL when network namespace files are missing, preventing IP address leaks.

#### What type of PR is this?

/kind bug

This will fix the issue mentioned in https://github.com/cri-o/cri-o/pull/9301#discussion_r2244988147

```release-note
server: ensure CNI teardown prevents IP leaks with missing netns
```

### DIFF
--- a/server/sandbox_network.go
+++ b/server/sandbox_network.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
@@ -16,6 +18,10 @@ import (
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/server/metrics"
+)
+
+const (
+	cacheDir = "/var/lib/cni/results"
 )
 
 // networkStart sets up the sandbox's network and returns the pod IP on success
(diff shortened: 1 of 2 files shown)


sohankunkerkar:
@danwinship could you PTAL?

sohankunkerkar:
https://github.com/cri-o/cri-o/actions/runs/16651679437/job/47126005482?pr=9372
This fix ensures CNI teardown is always called to prevent IP leaks, even when network namespaces are invalid/missing. However, this causes the above test to fail when the CNI plugin is temporarily unavailable (it looks like we need to fix the test?)

server/sandbox_network.go
```diff
@@ -185,43 +187,77 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
+	if err != nil {
+		log.Warnf(ctx, "Failed to read CNI cache directory %s: %v", cacheDir, err)
+
+		return
+	}
+
+	for _, entry := range entries {
+		if !entry.IsDir() && strings.Contains(entry.Name(), containerID) {
+			filePath := filepath.Join(cacheDir, entry.Name())
+			if err := os.Remove(filePath); err != nil {
```
lance5890, line 254: as this works as cleanup func, should we use os.RemoveAll to rm the dir if it contains something

sohankunkerkar: I  think `os.RemoveAll` would delete CNI result files for ALL containers on the system, potentially breaking networking for other running pods. The current implementation correctly targets only the specific container's result files by checking if the filename contains the containerID.

sohankunkerkar:
@cri-o/cri-o-maintainers PTAL

server/sandbox_network.go
```diff
@@ -185,43 +187,77 @@ func (s *Server) networkStop(ctx context.Context, sb *sandbox.Sandbox) error {
...
 		return fmt.Errorf("network teardown failed for pod sandbox %s(%s): %w", sb.Name(), sb.ID(), err)
 	}
 
 	return sb.SetNetworkStopped(ctx, true)
 }
 
+// cleanupCNIResultFiles removes CNI result files for a given container ID.
+// This is called when CNI teardown fails to prevent stale result files from accumulating.
+func (s *Server) cleanupCNIResultFiles(ctx context.Context, containerID string) {
+	cacheDir := "/var/lib/cni/results"
```
saschagrunert: Please make this a const or use it from the CNI library, which may not be possible, though.

PR #9385, opened 2025-08-06
server: add real-time memory validation for limit updates

sohankunkerkar:
This change validates memory limits against current usage from cgroup files instead of stale cAdvisor cache.

#### What type of PR is this?

/kind bug

```release-note
server: add real-time memory validation for limit updates
```

### DIFF
--- a/server/container_update_resources.go
+++ b/server/container_update_resources.go
@@ -44,6 +44,11 @@ func (s *Server) UpdateContainerResources(ctx context.Context, req *types.Update
 			updated = req.GetLinux()
 		}
 
+		// Validate memory update before applying it.
+		if err := s.validateMemoryUpdate(ctx, c, updated.GetMemoryLimitInBytes()); err != nil {
+			return nil, fmt.Errorf("memory validation failed: %w", err)
+		}
+
 		resources := toOCIResources(updated)
 		if err := s.ContainerServer.Runtime().UpdateContainer(ctx, c, resources); err != nil {
 			return nil, err
--- a/server/container_update_resources_linux.go
+++ b/server/container_update_resources_linux.go
@@ -1,0 +1,59 @@
+//go:build linux
+
+package server
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cri-o/cri-o/internal/log"
+	"github.com/cri-o/cri-o/internal/oci"
+)
+
+// validateMemoryUpdate checks if the new memory limit is safe to apply by getting current usage from CRI-O's stats server.
+// This ensures real-time accuracy and prevents decreasing memory limits below current usage.
+func (s *Server) validateMemoryUpdate(ctx context.Context, c *oci.Container, newMemoryLimit int64) error {
+	// Negative memory limits are invalid
+	if newMemoryLimit < 0 {
+		return fmt.Errorf("invalid memory limit: %d bytes (cannot be negative)", newMemoryLimit)
+	}
+
+	// Zero means unlimited memory, no validation needed.
+	if newMemoryLimit == 0 {
+		return nil
+	}
+
+	sb := s.GetSandbox(c.Sandbox())
+	if sb == nil {
+		log.Warnf(ctx, "Could not get sandbox %s for container %s", c.Sandbox(), c.ID())
+
+		return nil
+	}
+
+	containerStats := s.StatsForContainer(c, sb)
+	if containerStats == nil {
+		log.Warnf(ctx, "No memory stats available for container %s", c.ID())
+
+		return nil
+	}
+
+	usageBytes := containerStats.GetMemory().GetUsageBytes()
+	if usageBytes == nil {
+		log.Warnf(ctx, "Memory usage not available for container %s", c.ID())
+
+		return nil
+	}
+
+	currentUsage := int64(usageBytes.GetValue())
+
+	// Check if new limit is below current usage.
+	if newMemoryLimit < currentUsage {
+		return fmt.Errorf("cannot decrease memory limit to %d bytes: current usage is %d bytes",
+			newMemoryLimit, currentUsage)
+	}
+
+	log.Debugf(ctx, "Memory validation passed: new limit %d >= current usage %d",
+		newMemoryLimit, currentUsage)
+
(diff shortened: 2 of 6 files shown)


server/container_update_resources.go
```diff
@@ -15,6 +17,47 @@ import (
...
+	}
+
+	sb := s.GetSandbox(c.Sandbox())
+	if sb == nil {
+		log.Warnf(ctx, "Could not get sandbox %s for container %s", c.Sandbox(), c.ID())
+
+		return nil
+	}
+
+	cgstats, err := s.Runtime().ContainerStats(ctx, c, sb.CgroupParent())
```
haircommander: I think we should go through stats server with "StatsForContainer" instead. Now, it'll do the same (gather live) but in the future it will use the cached value which will be faster

server/container_update_resources.go
```diff
@@ -15,6 +17,47 @@ import (
 	"github.com/cri-o/cri-o/internal/oci"
 )
 
+// validateMemoryUpdate checks if the new memory limit is safe to apply by getting current usage from CRI-O's stats server.
+// This ensures real-time accuracy and prevents decreasing memory limits below current usage.
+func (s *Server) validateMemoryUpdate(ctx context.Context, c *oci.Container, newMemoryLimit int64) error {
+	if newMemoryLimit <= 0 {
+		return nil
+	}
```
bitoku: what's the intent of this? Shouldn't this be error?

server/container_update_resources_unsupported.go
bitoku, line 1: Can we create `container_update_resources_linux.go` and `container_update_resources_unsupported.go` and define just `validateMemoryUpdate` in them?

server/container_update_resources.go
```diff
@@ -15,6 +17,47 @@ import (
...
+
+	if containerStats.GetMemory().GetUsageBytes() == nil {
+		log.Warnf(ctx, "Memory usage not available for container %s", c.ID())
+
+		return nil
+	}
+
+	currentUsage := int64(containerStats.GetMemory().GetUsageBytes().GetValue())
```
bitoku: nil check is not required for auto-generated proto struct. Get*() chain is smart enough and if one of them is nil, the whole chain returns nil without panic.
We just need to check whether the final value is nil or not.
https://github.com/cri-o/cri-o/blob/ab3f85e5fc00b55ca036cb0b61252310db5d9b0f/vendor/k8s.io/cri-api/pkg/apis/runtime/v1/api.pb.go#L9353-L9358

test/cri-metrics.bats
```diff
@@ -256,3 +256,36 @@ EOF
...
+
+	# Verify the container is still running with original memory limit.
+	run crictl inspect "$ctr_id"
+	[[ "$status" -eq 0 ]]
+	[[ "$output" == *"CONTAINER_RUNNING"* ]]
+
+	# Test that memory limit increase still works.
+	run crictl update --memory 268435456 "$ctr_id"
+	echo "Increase attempt output: $output"
+	[[ "$status" -eq 0 ]]
```
bitoku: nit:
```suggestion
	crictl update --memory 268435456 "$ctr_id"
	echo "Increase attempt output: $output"
```

test/cri-metrics.bats
```diff
@@ -256,3 +256,36 @@ EOF
...
+	# Attempt to update memory limit to 32MB
+	run crictl update --memory 33554432 "$ctr_id"
+	echo "Update attempt output: $output"
+	[[ "$status" -ne 0 ]]
+	[[ "$output" =~ "cannot decrease memory limit" ]]
+
+	# Verify the container is still running with original memory limit.
+	run crictl inspect "$ctr_id"
+	[[ "$status" -eq 0 ]]
+	[[ "$output" == *"CONTAINER_RUNNING"* ]]
```
bitoku: It doesn't check the current memory limit.

server/container_update_resources_unsupported.go
sohankunkerkar: Done!

test/cri-metrics.bats
sohankunkerkar: I think `validateMemoryUpdate` function reads current usage from stats. The test shouldn't duplicate the implementation's logic. IMO, We should be testing the behavior here, not implementation details.

sohankunkerkar: Alright, let me add that logic as well.

test/cri-metrics.bats
```diff
@@ -256,3 +256,49 @@ EOF
 	metrics_container_processes=$(echo "$metrics" | jq 'select(.name == "container_processes") | .value.value | tonumber')
 	[[ $metrics_container_processes == "3" ]]
 }
+
+@test "memory limit decrease below usage should be blocked" {
```
bitoku: I should've noticed it first, but should this test be put somewhere else? It's a test file for metrics.

sohankunkerkar: I placed it in `ctr.bats`

server/container_update_resources_linux.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil
+	}
+
+	currentUsage := int64(usageBytes.GetValue())
+
+	// Check if new limit is below current usage.
+	if newMemoryLimit < currentUsage {
+		return fmt.Errorf("cannot decrease memory limit to %d bytes: current usage is %d bytes",
+			newMemoryLimit, currentUsage)
+	}
```
bitoku, line 53: How should we handle if `currentUsage` is 0 (= unlimited?)?

sohankunkerkar: IIRC, `currentUsage = 0` doesn't mean unlimited, it means the container hasn't allocated any memory yet (or is using negligible memory). This is different from `newMemoryLimit = 0` which means unlimited. When `currentUsage = 0`, it either means the container just started and hasn't allocated significant memory or the container is idle. I think we should allow setting any reasonable memory limit when the container isn't using memory yet.

bitoku: ah sorry I got the actual usage and the current limit mixed up. thanks for clarification.

PR #9401, opened 2025-08-12
image pull: enforce shortnames

haircommander:
I personally thought "disable" here meant "don't allow shortnames to be resolved" actually, it means "allow any shortname to be resolved if one is found in the unqualified search registry"

I think CRI-O's official position should be to discourage the use of 'unqualified-search-registries', but if there are some present, we should block if there's ambiguity in which image to pull from which, so we can prevent squatting issues

```release-note
Enable shortname enforcement, requiring shortnames to be unambiguous if there are multiple entries that could be pulled from the list of 'unqualified-search-registries'. Generally, users should not enable 'unqualified-search-registries' but if they must be used, they should be unambiguous
```


haircommander:
/retest
I would not be surprised if a bunch of stuff breaks

internal/storage/image.go
```diff
@@ -1017,8 +1017,8 @@ func (svc *imageService) CandidatesForPotentiallyShortImageName(systemContext *t
 		sc = *systemContext // A shallow copy
 	}
 
-	disabled := types.ShortNameModeDisabled
-	sc.ShortNameMode = &disabled
+	enforcing := types.ShortNameModeEnforcing
+	sc.ShortNameMode = &enforcing
```
bitoku: nit:
I'd suggest ensuring the length == 1 instead of using `types.ShortNameModeEnforcing` because the error message is not user-friendly.
https://github.com/cri-o/cri-o/blob/c36f2b5bddfe321fa35976dfe5d3fe19eb3a1c59/vendor/github.com/containers/image/v5/pkg/shortnames/shortnames.go#L351-L353

bitoku (commented):
It can be a follow up, but we can change the name `CandidatesForPotentiallyShortImageName` and return value type because it will always return a single image.

internal/storage/image.go
haircommander: fixed!

haircommander:
okay I think this is good now (unless unit tests fail, I'm letting CI test that)

@cri-o/cri-o-maintainers PTAL

internal/ociartifact/impl.go
```diff
@@ -148,8 +148,8 @@ func (d *defaultImpl) CandidatesForPotentiallyShortImageName(systemContext *type
 		sc = *systemContext // A shallow copy
 	}
 
-	disabled := types.ShortNameModeDisabled
-	sc.ShortNameMode = &disabled
+	enforcing := types.ShortNameModeEnforcing
+	sc.ShortNameMode = &enforcing
 
 	return shortnames.ResolveLocally(&sc, imageName)
 }
```
bitoku: It wants the same logic as the image. also it should use the given systemContext to respect the config.

test/image.bats
```diff
@@ -397,3 +397,19 @@ EOF
 
 	crictl pull "$IMAGE_LIST_TAG"
 }
+
+@test "short name mode enabled should fail to pull ambiguous image" {
+	start_crio
+
+	# There should be many busyboxes
+	run crictl pull nginx
```
bitoku: nit: busybox or nginx?

test/image.bats
```diff
@@ -397,3 +397,19 @@ EOF
...
+	[[ "$output" == *"short name mode is enforcing, but image name nginx returns ambiguous list"* ]]
+	[[ "$status" -ne 0 ]]
+}
+
+@test "short name mode disabled should succeed to pull ambiguous image" {
+	CONTAINER_SHORT_NAME_MODE="disabled" start_crio
+
+	# There should be many busyboxes
+	crictl pull nginx
```
bitoku: nit: busybox or nginx?

haircommander:
@cri-o/cri-o-maintainers PTAL

bitoku (commented):
/lgtm

kevinastone:
Is there a mechanism to override this?  It's causing image pull failures when I upgrade to `v1.34`.

haircommander:
yup! a drop-in cri-o configuration would do it
```
[crio.image]
short_name_mode = "disabled
```
anywhere in `/etc/crio/crio.conf.d` :)

PR #9402, opened 2025-08-13
Add support for conmon-rs log driver and heaptrack config

saschagrunert:
#### What type of PR is this?

/kind feature

Reuse `monitor_env` to pass down the options to the pod runtime (conmon-rs).

None

None

```release-note
Added support for conmon-rs log driver and heaptrack config by using the `monitor_env` runtime configuration.
```

### DIFF
--- a/internal/oci/runtime_pod.go
+++ b/internal/oci/runtime_pod.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	"github.com/containers/common/pkg/resize"
@@ -59,10 +60,38 @@ func newRuntimePod(r *Runtime, handler *config.RuntimeHandler, c *Container) (Ru
 		cgroupManager = conmonClient.CgroupManagerCgroupfs
 	}
 
+	heaptrack := &conmonClient.Heaptrack{}
+	logDriver := conmonClient.LogDriverNone
+
+	for _, env := range handler.MonitorEnv {
+		keyVal := strings.SplitN(env, "=", 2)
+		if len(keyVal) != 2 {
+			logrus.Warnf("Skipping monitor env %q because it is not in key=value format", env)
+
+			continue
+		}
+
+		switch keyVal[0] {
+		case "LOG_DRIVER":
+			logDriver = conmonClient.LogDriver(keyVal[1])
+
+		case "HEAPTRACK_OUTPUT_PATH":
+			heaptrack.Enabled = true
+			heaptrack.OutputPath = filepath.Join(keyVal[1], "cri-o.conmon-rs."+c.ID())
+
+		case "HEAPTRACK_BINARY_PATH":
+			heaptrack.Enabled = true
+			heaptrack.BinaryPath = keyVal[1]
+
+		default:
+			logrus.Warnf("Unknown monitor env option %q", env)
+		}
+	}
+
 	client, err := conmonClient.New(&conmonClient.ConmonServerConfig{
 		ConmonServerPath: handler.MonitorPath,
 		LogLevel:         conmonClient.FromLogrusLevel(logrus.GetLevel()),
-		LogDriver:        conmonClient.LogDriverSystemd,
+		LogDriver:        logDriver,
 		Runtime:          handler.RuntimePath,
 		ServerRunDir:     c.dir,
 		RuntimeRoot:      runRoot,
@@ -72,6 +101,7 @@ func newRuntimePod(r *Runtime, handler *config.RuntimeHandler, c *Container) (Ru
 			Enabled:  r.config.EnableTracing,
 			Endpoint: "http://" + r.config.TracingEndpoint,
 		},
+		Heaptrack: heaptrack,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -243,6 +243,16 @@ type RuntimeHandler struct {
 	MonitorCgroup string   `toml:"monitor_cgroup,omitempty"`
 	MonitorEnv    []string `toml:"monitor_env,omitempty"`
 
+	// MonitorLogDriver is the log driver used for the container monitor.
+	// Can be empty, "systemd" or "stdout". An empty or incorrect value
+	// defaults to disabled logging. Works only for the "pod" runtime_type.
+	MonitorLogDriver string `toml:"monitor_log_driver,omitempty"`
```
bitoku: I think we want validation when something other than "stdout", "systemd", (and "none" or ""?) is set.

pkg/config/config.go
```diff
@@ -243,6 +243,16 @@ type RuntimeHandler struct {
...
+
+	// MonitorHeaptrackOutputPath is the path to output heaptrack profiles
+	// generated by the container monitor. Works only for the "pod"
+	// runtime_type and creates profiles per pod.
+	MonitorHeaptrackOutputPath string `toml:"monitor_heaptrack_output_path,omitempty"`
```
haircommander: hmm any ways we could make this more generic?

haircommander: MonitorExtraBinaries or something? and then the runtime interprets? could be key->value.. i dunno

saschagrunert: Some generic `monitor_options`? Then we could enable heaptrack via `monitor_options=heaptrack_output=/path/to,heaptrack_binary=/foo/bar`. We could also just use the existing `monitor_env` then, which is unused for conmon-rs right now.

Edit: Switched over to reuse `monitor_env`.

pkg/config/config.go
saschagrunert: Switched over to reuse `monitor_env`.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

saschagrunert:
@bitoku @haircommander PTAL

pkg/config/config.go
bitoku: IMHO we don't necessarily have the common options for conmon and conmon-rs.
conmon-rs is already quite different from conmon, and if we continue to add new options, I'm concerned more about the maintainability like parsing each option, similar options but different format between them etc.
I'd want separate configs for conmon and conmon-rs. It also helps users understand what feature is supported by the container monitors.

It's not blocking the change though.

internal/oci/runtime_pod.go
```diff
@@ -59,10 +60,38 @@ func newRuntimePod(r *Runtime, handler *config.RuntimeHandler, c *Container) (Ru
...
+			logDriver = conmonClient.LogDriver(keyVal[1])
+
+		case "HEAPTRACK_OUTPUT":
+			heaptrack.Enabled = true
+			heaptrack.OutputPath = filepath.Join(keyVal[1], "cri-o.conmon-rs."+c.ID())
+
+		case "HEAPTRACK_BINARY_PATH":
+			heaptrack.Enabled = true
+			heaptrack.BinaryPath = keyVal[1]
+
```
bitoku: what if one of them is specified? Do we need the validation or delegate it to conmonClient?

internal/oci/runtime_pod.go
```diff
@@ -59,10 +60,38 @@ func newRuntimePod(r *Runtime, handler *config.RuntimeHandler, c *Container) (Ru
...
+			logrus.Warnf("Skipping monitor env %q because it is not in key=value format", env)
+
+			continue
+		}
+
+		switch keyVal[0] {
+		case "LOG_DRIVER":
+			logDriver = conmonClient.LogDriver(keyVal[1])
+
+		case "HEAPTRACK_OUTPUT":
```
bitoku: nit: `HEAPTRACK_OUTPUT_PATH` or `HEAPTRACK_OUTPUT_DIR` is clearer maybe?

saschagrunert: Changed as suggested.

internal/oci/runtime_pod.go
saschagrunert: I'd prefer to let conmon-rs client do that validation and just pass down the variables.

pkg/config/template.go
```diff
@@ -1290,6 +1290,11 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
 #   should be moved to the container's cgroup
 # - monitor_env (optional, array of strings): Environment variables to pass to the monitor.
 #   Replaces deprecated option "conmon_env".
+#   When using the pod runtime and conmon-rs, then the monitor_env can be used to further configure
+#   conmon-rs by using:
+#     - LOG_DRIVER=[none,systemd,stdout] - Enable logging to the configured target, defaults to none.
+#     - HEAPTRACK_OUTPUT=/path/to/dir - Enable heaptrack profiling and save the files to the set directory.
```
bitoku: This needs change too.

saschagrunert: Good catch, thank you!

PR #9405, opened 2025-08-13
OCPBUGS-59403: HighPerformanceHooks: Atomic locking and fix container replacement race conditions

andreaskaris:
Supersedes https://github.com/cri-o/cri-o/pull/9364

Addresses race conditions issues IRQ SMP affinity management for high-performance hooks.

1. Race conditions from dual locking: The prior patch used separate locks for IRQ SMP affinity and irqbalance configuration files, creating a window where concurrent operations could corrupt system state. This fix consolidates to a single atomic lock protecting the entire read-calculate-write cycle for both /proc/irq/default_smp_affinity and /etc/sysconfig/irqbalance.
2. Kubelet container replacement timing: Addresses scenarios where kubelet starts replacement containers before deleting old ones, leading to incorrect IRQ affinity calculations. The new approach moves the SMP IRQ affinity logic into PostStop hooks where the order of events should be correct even on replacement.
3. systemctl restart limitations: While adding reference to PR8834's systemd StartLimitBurst workaround, the implementation recognizes that rapid systemctl restarts are inherently problematic (concurrent restarts are ignored while service is  restarting).

#### What type of PR is this?

/kind bug

Fix race condition in IRQ balance configuration and during container replacement.

This is a follow-up to https://github.com/cri-o/cri-o/pull/9228 which did not fix all race conditions

> **NOTE**: The move to PostStop causes the logic to be triggered several times per container stop. The context carries a name field (e.g. `name=/runtime.v1.RuntimeService/StopContainer`) and we might be able to check that to reduce repetitive calls, but I decided against that for the time being in the interest of simplicity.

##### Reproduce / test for container replacement

I ported the fix to 4.18 and tested it on an OpenShift cluster.

Here's a deployment which reproduces the contrainer replacement race condition without the move of code into PostStop:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: qos-demo
  name: qos-demo
spec:
  replicas: 10
  selector:
    matchLabels:
      app: qos-demo
  strategy: {}
  template:
    metadata:
      name: qos-demo
      labels:
        app: qos-demo
      annotations:
    spec:
      hostNetwork: true
      runtimeClassName: performance-sno-pp
      containers:
      - name: qos-demo-ctr-1
        image: quay.io/akaris/nice-test@sha256:95b176d4f54257c89614453f93a919559b377ba416401a2ef76bff4936effbd8
        command:
        - "/bin/bash"
        - "-c"
        - |
          exit 1
        resources:
          limits:
            memory: "100Mi"
            cpu: "2"
          requests:
            memory: "100Mi"
            cpu: "2"
        livenessProbe:
          exec:
            command:
            - cat
            - /tmp/healthy
          initialDelaySeconds: 1
          periodSeconds: 1
          failureThreshold: 1
          terminationGracePeriodSeconds: 1
      - name: qos-demo-ctr-2
        image: quay.io/akaris/nice-test@sha256:95b176d4f54257c89614453f93a919559b377ba416401a2ef76bff4936effbd8
        command:
        - "/bin/bash"
        - "-c"
        - |
          exit 1
        resources:
          limits:
            memory: "100Mi"
            cpu: "2"
          requests:
            memory: "100Mi"
            cpu: "2"
        livenessProbe:
          exec:
            command:
            - cat
            - /tmp/healthy
          initialDelaySeconds: 1
          periodSeconds: 1
          failureThreshold: 1
          terminationGracePeriodSeconds: 1
```

With debug logging enabled in cri-o, you can run:
```
for i in {1..3}; do ssh core@192.168.18.21 cat /proc/irq/default_smp_affinity; oc apply -f deployment.yaml ; sleep $(($RANDOM % 120)); oc delete -f deployment.yaml;  kubectl wait --for delete pod --selector=app=qos-demo --timeout=120s;  done
```
> This requires the change to Debugf in one of the commits.

And then grep for this in the journal:
```
journalctl --boot  -u crio | grep -E 'Container .* (.*) enable' > /tmp/journal.txt
```

Get the starting mask:
```
# cat /proc/irq/default_smp_affinity
ffffffff,ffffffff
```

Then, put this into a file and run the following script, where the mask is the starting mask (should be all fffff...., otherwise the script will run into issues):
```
./analyze.py journal.txt ffffffff,ffffffff
```

Here's the parser script:
```
#!/usr/bin/env python

import re
import sys
import codecs


def parse_msg(msg: str):
    r = re.compile(
        r'Container "(.*)" \("(.*)"\) enable ([a-z]+) cpus \"(.*)\" set \"/proc/irq/default_smp_affinity\": \"(.*)\" -> \"(.*)\";.*'
    )
    items = r.findall(msg)
    if len(items) != 1 or len(items[0]) != 6:
        raise Exception(f"issue in parse_msg with {msg}")
    container_id, container_name, enable, cpus, from_irq, to_irq = items[0]
    return {
        "container_id": container_id,
        "container_name": container_name,
        "enable": enable,
        "cpus": cpus,
        "from_irq": from_irq,
        "to_irq": to_irq,
    }


def parse_line(line: str):
    r = re.compile(r'([a-z]+)=("(?:\\\"|.)*?"|[^"].+?\s)')
    items = r.findall(line)
    parsed_items = {}
    for field, value in items:
        value = value.strip()
        value = str(codecs.escape_decode(bytes(value, "utf-8"))[0])
        value = value.strip('"')
        parsed_items[field] = value
    return parsed_items


def line_iterator(file_name: str):
    with open(file_name) as f:
        for line in f:
            items = parse_line(line)
            if "msg" not in items:
                raise Exception("msg not in items, items: {items}")
            event = parse_msg(items["msg"])
            yield event


def hex_to_cpu_lists(hex_affinity: str):
    hex_affinity = hex_affinity.replace(",", "").strip()

    x = int(hex_affinity, 16)
    i = 0
    enabled_cpus = []
    disabled_cpus = []

    for i in range(len(hex_affinity) * 4):
        shifted = 1 << i
        if x & shifted != 0:
            enabled_cpus.append(i)
        else:
            disabled_cpus.append(i)

    return enabled_cpus, disabled_cpus


def parse_cpu_list(cpus: str):
    cpu_set = set()
    cpu_list = cpus.split(",")
    for sublist in cpu_list:
        r = sublist.split("-")
        if len(r) == 1:
            r = set([int(r[0])])
        elif len(r) == 2:
            r = set(range(int(r[0]), int(r[1]) + 1))
        else:
            raise Exception(f"invalid list {cpus}")
        cpu_set.update(r)
    return cpu_set


def strip(mask: str, lim: int) -> str:
    """
    Strip leading zeros from a CPU mask beyond the CPU count.

    Args:
        mask: CPU mask in hex format with commas (e.g., "0000ffff,ffffffff,ffffffff")
        lim: Maximum number of CPUs allowed

    Returns:
        Stripped CPU mask

    Raises:
        Exception: If any bit is set beyond the CPU count limit
    """
    # Remove commas and convert to integer
    mask_clean = mask.replace(",", "").strip()
    mask_int = int(mask_clean, 16)

    # Check if any bits are set beyond the CPU limit
    # Create a mask for bits beyond lim
    if lim < len(mask_clean) * 4:
        # Check if any bit >= lim is set
        for bit_pos in range(lim, len(mask_clean) * 4):
            if mask_int & (1 << bit_pos):
                raise Exception(
                    f"CPU mask has bit {bit_pos} set, but CPU limit is {lim}"
                )

    # Create a mask with only the bits we care about (0 to lim-1)
    valid_bits_mask = (1 << lim) - 1
    stripped_int = mask_int & valid_bits_mask

    # Convert back to hex string
    # Calculate minimum hex digits needed for lim bits
    min_hex_digits = (lim + 3) // 4  # Round up
    hex_str = format(stripped_int, f"0{min_hex_digits}x")

    # Add commas every 8 hex digits from the right
    result_parts = []
    for i in range(len(hex_str), 0, -8):
        start = max(0, i - 8)
        result_parts.append(hex_str[start:i])

    return ",".join(reversed(result_parts))


class CPU:
    def __init__(self):
        self.containers = {}

    def add_container(self, name: str):
        self.containers[name] = True

    def remove_container(self, name: str):
        try:
            del self.containers[name]
        except:
            raise Exception(
                f"Trying to remove container {name} from CPU but it's not on it"
            )

    def is_empty(self):
        return len(self.containers) == 0

    def is_enabled(self):
        return len(self.containers) == 0


class SystemState:
    def __init__(self, start_state: str):
        self.events = []
        self.cpus = {}

        # Add CPUs if they must be added to the tracking list.
        enabled_cpus, disabled_cpus = hex_to_cpu_lists(start_state)
        if len(disabled_cpus) != 0:
            raise Exception("invalid start state, all CPUs must be enabled")
        for cpu_id in enabled_cpus:
            if cpu_id not in self.cpus:
                self.cpus[cpu_id] = CPU()

    def add_event(self, event: dict):
        # Sanity check - Make sure that we have a chronological list of changes.
        if len(self.events) > 0:
            last_event = self.events[len(self.events) - 1]
            if strip(last_event["to_irq"], len(self.cpus)) != strip(
                event["from_irq"], len(self.cpus)
            ):
                raise Exception(
                    f"Consecutive events do not match, t-1: {last_event}, t: {event}"
                )

        # Sanity check for invalid CPUs.
        from_enabled_cpus, from_disabled_cpus = hex_to_cpu_lists(
            strip(event["from_irq"], len(self.cpus))
        )
        to_enabled_cpus, to_disabled_cpus = hex_to_cpu_lists(
            strip(event["to_irq"], len(self.cpus))
        )
        all_cpus = set(
            from_enabled_cpus + from_disabled_cpus + to_enabled_cpus + to_disabled_cpus
        )
        for cpu_id in all_cpus:
            if cpu_id not in self.cpus:
                s = f"found invalid CPU in event, valid CPUs are {self.cpus}, got: {all_cpus}"
                raise Exception(s)

        # Sanity check.
        # Get CPUs that were enabled / disabled and make sure that there's never an
        # event that does both at the same time.
        # We can have noop events, though.
        enabled_cpus = set(to_enabled_cpus).difference(set(from_enabled_cpus))
        disabled_cpus = set(to_disabled_cpus).difference(set(from_disabled_cpus))
        if len(enabled_cpus) != 0 and len(disabled_cpus) != 0:
            raise Exception(
                "a container enabled and disabled CPUs at the same time; "
                f"enabled: {enabled_cpus}, "
                f"disabled: {disabled_cpus}, "
                f"event: {event}"
            )
        elif len(enabled_cpus) != 0 and event["enable"] != "true":
            raise Exception(
                "a container enabled CPUs but the enable is not true, "
                f"enabled: {enabled_cpus}, "
                f"enable: {event['enable']}, "
                f"event: {event}"
            )
        elif len(disabled_cpus) != 0 and event["enable"] != "false":
            raise Exception(
                "a container disabled CPUs but the enable is not false, "
                f"disabled: {disabled_cpus}, "
                f"enable: {event['enable']}, "
                f"event: {event}"
            )

        # Now, add /remove containers -> this will update the internal expected state of CPU.
        try:
            cpu_list = parse_cpu_list(event["cpus"])
        except Exception as e:
            raise Exception(f"{e}, event: {event}")

        if event["enable"] == "false":
            for cpu_id in cpu_list:
                self.cpus[cpu_id].add_container(event["container_name"])
        elif event["enable"] == "true":
            for cpu_id in cpu_list:
                try:
                    self.cpus[cpu_id].remove_container(event["container_name"])
                except Exception as e:
                    print("\tINFO", e, cpu_id)
        else:
            raise Exception(f"Invalid enable {event["enable"]} in {event}")

        # Now, check that is and expected states are consistent.
        for cpu_id in to_disabled_cpus:
            if self.cpus[cpu_id].is_enabled():
                raise Exception(
                    f"invalid CPU state for CPU {cpu_id}; expected: enabled according, got: disabled from log, "
                    f"containers on CPU: {self.cpus[cpu_id].containers}, "
                    f"event: {event}"
                )
        for cpu_id in to_enabled_cpus:
            if not self.cpus[cpu_id].is_enabled():
                raise Exception(
                    f"invalid CPU state for CPU {cpu_id}; expected: disabled according to script, got: enabled from log, "
                    f"containers on CPU: {self.cpus[cpu_id].containers}, "
                    f"event: {event}"
                )
        self.events.append(event)


def main():
    if len(sys.argv) != 3:
        print("Must provide file name of log file, and expected init state for IRQs")
        sys.exit(1)
    file_name = sys.argv[1]
    init_state = sys.argv[2]

    s = SystemState(init_state)

    for event in line_iterator(file_name):
        print("EVENT", event)
        s.add_event(event)


if __name__ == "__main__":
    main()
```

For PreStop logic, that yields:
```
Exception: invalid CPU state for CPU 7; expected: disabled according to script, got: enabled from log, containers on CPU: {'k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_1': True}, event: {'container_id': 'ac33610833dbddb7295bc94096a367973059bff373eb80c19895057f1c13c4d3', 'container_name': 'k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_0', 'action': 'stop', 'cpus': '7,39', 'from_irq': 'fc00003f,fc00003f', 'to_irq': 'fc0000bf,fc0000bf'}
```

And when looking for the 2 containers in question, one can clearly see that the state was inconsistent for CPUs 7,39 for 17 seconds:
```
$ grep -E "k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_1|k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_0" journal.prestop.txt 
Aug 13 21:59:06 sno10-1 crio[4148]: time="2025-08-13 21:59:06.402571811Z" level=debug msg="Container \"ac33610833dbddb7295bc94096a367973059bff373eb80c19895057f1c13c4d3\" (\"k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_0\") action start cpus \"7,39\" set \"/proc/irq/default_smp_affinity\": \"febbbabf,febbbabf\" -> \"febbba3f,febbba3f\"; \"/etc/sysconfig/irqbalance\": \"014445c0,014445c0\"" file="runtimehandlerhooks/high_performance_hooks_linux.go:707" id=d1e3fb5d-a965-42b5-9e5b-c7a175b8b20c name=/runtime.v1.RuntimeService/StartContainer
Aug 13 21:59:08 sno10-1 crio[4148]: time="2025-08-13 21:59:08.463707597Z" level=debug msg="Container \"1ed1264f1d37758973ccf87aea1189828cbbefa439d6b1eb9332dcd917205791\" (\"k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_1\") action start cpus \"7,39\" set \"/proc/irq/default_smp_affinity\": \"fc00003f,fc00003f\" -> \"fc00003f,fc00003f\"; \"/etc/sysconfig/irqbalance\": \"03ffffc0,03ffffc0\"" file="runtimehandlerhooks/high_performance_hooks_linux.go:707" id=8d211390-41e8-4309-b620-4346f86ab120 name=/runtime.v1.RuntimeService/StartContainer
Aug 13 21:59:08 sno10-1 crio[4148]: time="2025-08-13 21:59:08.638505598Z" level=debug msg="Container \"ac33610833dbddb7295bc94096a367973059bff373eb80c19895057f1c13c4d3\" (\"k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_0\") action stop cpus \"7,39\" set \"/proc/irq/default_smp_affinity\": \"fc00003f,fc00003f\" -> \"fc0000bf,fc0000bf\"; \"/etc/sysconfig/irqbalance\": \"03ffff40,03ffff40\"" file="runtimehandlerhooks/high_performance_hooks_linux.go:707" id=e8b53e38-c5ae-48f8-a41d-5029326d1517 name=/runtime.v1.RuntimeService/RemoveContainer
Aug 13 21:59:25 sno10-1 crio[4148]: time="2025-08-13 21:59:25.356663358Z" level=debug msg="Container \"1ed1264f1d37758973ccf87aea1189828cbbefa439d6b1eb9332dcd917205791\" (\"k8s_qos-demo-ctr-2_qos-demo-56cb556bb8-lbbqh_default_a26497bb-18e6-4099-a959-14f588e361ac_1\") action stop cpus \"7,39\" set \"/proc/irq/default_smp_affinity\": \"ffffff3f,ffffff3f\" -> \"ffffffbf,ffffffbf\"; \"/etc/sysconfig/irqbalance\": \"00000040,00000040\"" file="runtimehandlerhooks/high_performance_hooks_linux.go:707" id=dd2b3cd2-3f66-4bd8-9aa9-6c5973e327ba name=/runtime.v1.RuntimeService/RemoveContainer
```

I tested this several times for the PostStop fix and could not reproduce the issue. If it's still there, it's significantly less likely to happen.

##### Smoke test for atomic locking

This is the smoke test for atomic locking (however, not for the container replacement issue).

Smoke test on a system with the upstream cri-o build.
```
mkdir -p /etc/systemd/system/irqbalance.service.d
cat <<'EOF' > /etc/systemd/system/irqbalance.service.d/restart-limits.conf
# cri-o high performance runtime handler reconfigures and restarts irqbalance
# after every container with pinned cpus and moved interrupts is created.
# Systemd does not like this and will block further restarts when the
# StartLimitBurst (default 5) is crossed within the default timeout (10s).
# That is too strict as during system startup many such containers might
# be launched.
# Install this to /etc/systemd/system/irqbalance.service.d/restart-limits.conf
# to prevent systemd from blocking the necessary restarts.
[Unit]
StartLimitBurst=100
EOF
systemctl daemon-reload
```

```
mv /usr/sbin/irqbalance{,.back}
cat <<'EOF' > /usr/sbin/irqbalance
# cat /usr/sbin/irqbalance
#!/bin/bash

log_file=/tmp/irqbalance.mock.log

echo "$(env | tr '\n' ' ') $@" >> $log_file
/usr/sbin/irqbalance.back "$@"
EOF
chmod +x /usr/sbin/irqbalance.back
```

```
cat <<'EOF' > smoke.sh
#!/bin/bash

set -x

banned_file="/etc/sysconfig/irqbalance"
expected_affinity="3e0f"
expected_reset_banned="IRQBALANCE_BANNED_CPUS=\"0000c000\""
expected_banned="IRQBALANCE_BANNED_CPUS=\"0000c1f0\""
expected_irqbalance_oneshots=10
# The delete runs systemctl restart in such quick succession that it shows only once (or more).
expected_irqbalance_restarts=6

cat $affinity_file

check() {
	expected_a="$1"
	expected_b="$2"

	affinity=$(cat ${affinity_file} | tr -d '\n')
        echo "Got affinity: $affinity, expected affinity: $expected_a"
        if [ "${affinity}" != "${expected_a}" ]; then
            exit 1
        fi
	banned=$(grep -E "^IRQBALANCE_BANNED_CPUS=" ${banned_file} | tr -d '\n')
        echo "Got banned: '$banned', expected banned: '$expected_b'"
        if [ "${banned}" != "${expected_b}" ]; then
            exit 1
        fi
}

# irqbalance --oneshot
systemctl disable --now irqbalance
for i in {0..5}; do
	set +x
	echo "========"
	echo "Run ${i}"
	echo "========"
	set -x
	irqbalance_restarts_start=$(grep "oneshot" /tmp/irqbalance.mock.log | wc -l)

	kubectl apply -f pod.yaml
	echo "After set:"
	check "$expected_affinity" "$expected_banned"
        kubectl delete pod qos-demo
	echo "After reset:"
	check "$expected_reset_affinity" "$expected_reset_banned"

	irqbalance_restarts_end=$(grep "oneshot" /tmp/irqbalance.mock.log | wc -l)
	irqbalance_restarts=$((irqbalance_restarts_end - irqbalance_restarts_start))
	echo "-------------"
	echo "==> Got $irqbalance_restarts irqbalance --oneshot runs"
	echo "-------------"
	if [ "$irqbalance_restarts" -lt "$expected_irqbalance_oneshots" ]; then
	    exit 1
	fi
done

# systemctl restart irqbalance
systemctl enable --now irqbalance
for i in {0..5}; do
	set +x
	echo "========"
	echo "Run ${i}"
	echo "========"
	set -x
	irqbalance_restarts_start=$(journalctl -u irqbalance --boot | grep "Started irqbalance daemon" | wc -l)

	kubectl apply -f pod.yaml
	echo "After set:"
	check "$expected_affinity" "$expected_banned"
        kubectl delete pod qos-demo
	echo "After reset:"
	check "$expected_reset_affinity" "$expected_reset_banned"

	irqbalance_restarts_end=$(journalctl -u irqbalance --boot | grep "Started irqbalance daemon" | wc -l)
	irqbalance_restarts=$((irqbalance_restarts_end - irqbalance_restarts_start))
	echo "-------------"
	echo "==> Got $irqbalance_restarts irqbalance restarts"
	echo "-------------"
	if [ "$irqbalance_restarts" -lt "$expected_irqbalance_restarts" ]; then
	    exit 1
	fi
done
EOF
chmod +x smoke.sh
```

```
cat <<'EOF' > pod.yaml
apiVersion: v1
kind: Pod
metadata:
  name: qos-demo
  annotations:
spec:
  hostNetwork: true
  containers:
  - name: qos-demo-ctr-1
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-2
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-3
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-4
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
  - name: qos-demo-ctr-5
    command:
    - "/bin/sleep"
    - "infinity"
    resources:
      limits:
        memory: "100Mi"
        cpu: "1"
      requests:
        memory: "100Mi"
        cpu: "1"
EOF
```

```
smoke.sh
```

```release-note
None
```

```release-note

```

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -748,7 +748,7 @@ func (h *HighPerformanceHooks) handleIRQBalanceOneShot(ctx context.Context, cNam
 }
 
 // updateNewIRQSMPAffinityMask updates SMP IRQ affinity and IRQ balance configuration files.
-func (h *HighPerformanceHooks) updateNewIRQSMPAffinityMask(ctx context.Context, cID, cName, cpus string, enable bool) (string, error) {
+func (h *HighPerformanceHooks) updateNewIRQSMPAffinityMask(ctx context.Context, cID, cName, cpus string, enable bool) (newIRQBalanceSetting string, err error) {
 	content, err := os.ReadFile(h.irqSMPAffinityFile)
 	if err != nil {
 		return "", err
@@ -771,17 +771,22 @@ func (h *HighPerformanceHooks) updateNewIRQSMPAffinityMask(ctx context.Context, 
 		return "", err
 	}
 
+	// Rollback IRQ SMP affinity file to maintain consistency if something goes wrong.
+	defer func() {
+		if err != nil {
+			if rollbackErr := os.WriteFile(h.irqSMPAffinityFile, []byte(originalIRQSMPSetting), 0o644); rollbackErr != nil {
+				log.Errorf(ctx, "Failed to rollback IRQ SMP affinity file after config update failure: err: %q, rollback err: %q",
+					err, rollbackErr)
+			}
+		}
+	}()
+
 	// Nothing else to do if irq balance config file does not exist.
 	if !fileExists(h.irqBalanceConfigFile) {
 		return newIRQBalanceSetting, nil
 	}
 
 	if err := updateIrqBalanceConfigFile(h.irqBalanceConfigFile, newIRQBalanceSetting); err != nil {
-		// Rollback IRQ SMP affinity file to maintain consistency
-		if rollbackErr := os.WriteFile(h.irqSMPAffinityFile, []byte(originalIRQSMPSetting), 0o644); rollbackErr != nil {
-			log.Errorf(ctx, "Failed to rollback IRQ SMP affinity file after config update failure: %v", rollbackErr)
-		}
-
 		return "", err
 	}
 
--- a/internal/runtimehandlerhooks/high_performance_hooks_test.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_test.go
@@ -813,6 +813,60 @@ var _ = Describe("high_performance_hooks", func() {
 		)
 	})
 
+	Describe("updateNewIRQSMPAffinityMask rollback", func() {
+		irqBalanceConfigFile := filepath.Join(fixturesDir, "irqbalance")
+		irqSMPAffinityFile := filepath.Join(fixturesDir, "irqsmpaffinity")
+
+		h := &HighPerformanceHooks{
+			irqSMPAffinityFile:   irqSMPAffinityFile,
+			irqBalanceConfigFile: irqBalanceConfigFile,
+		}
+
+		type parameters struct {
+			irqBalanceFileRO           bool
+			originalIRQSMPAffinityMask string
+			expectedIRQSMPAffinityMask string
+		}
+
+		DescribeTable("test rollback",
+			func(p parameters) {
+				err := os.WriteFile(irqSMPAffinityFile, []byte(p.originalIRQSMPAffinityMask), 0o644)
+				Expect(err).ToNot(HaveOccurred())
+
+				if p.irqBalanceFileRO {
+					err = os.Symlink("/proc/version", irqBalanceConfigFile)
+					Expect(err).ToNot(HaveOccurred())
+				} else {
+					err = os.WriteFile(irqBalanceConfigFile, []byte(""), 0o644)
+					Expect(err).ToNot(HaveOccurred())
+				}
+
+				_, err = h.updateNewIRQSMPAffinityMask(context.TODO(), "cID", "CName", "2-3", false)
+				if p.irqBalanceFileRO {
+					Expect(err).To(HaveOccurred())
+				} else {
+					Expect(err).NotTo(HaveOccurred())
+				}
+
+				writtenMask, err := os.ReadFile(irqSMPAffinityFile)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(string(writtenMask)).To(Equal(p.expectedIRQSMPAffinityMask))
+			},
+			Entry("writing IRQ balance file fails",
+				parameters{
+					irqBalanceFileRO:           true,
+					originalIRQSMPAffinityMask: "ffffffff",
+					expectedIRQSMPAffinityMask: "ffffffff",
+				}),
+			Entry("writing IRQ balance file succeeds",
+				parameters{
+					irqBalanceFileRO:           false,
+					originalIRQSMPAffinityMask: "ffffffff",
+					expectedIRQSMPAffinityMask: "fffffff3",
+				}),
+		)
+	})
+
 	Describe("convertAnnotationToLatency", func() {
 		verifyConvertAnnotationToLatency := func(annotation string, expected string, expect_error bool) {
 			latency, err := convertAnnotationToLatency(annotation)


andreaskaris:
Failing CI tests are due to image pull issues (issues with the CI itself).

@MarSik I'll give it another look myself tomorrow, but I think this is ready for review now

andreaskaris:
The customer came back to us, and simply moving stuff to posthook does not seem to be a reliable solution:
Instead, the idempotent approach of my earlier PR fixes the issue.

1) The testing using  https://github.com/cri-o/cri-o/pull/9364 was positive. The issue of disappearing IRQ cores was not observed.
2) The testing using https://github.com/cri-o/cri-o/pull/9405 was not positive. The issue was observed with this rpm installed.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +640,103 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
 		return fmt.Errorf("find container %s CPUs", c.ID())
 	}
 
-	newIRQBalanceSetting, err := h.updateNewIRQSMPAffinityMask(lspec.Resources.CPU.Cpus, enable)
```
andreaskaris, line 634: Problem:
- If I move the lock here then irqbalance --oneshot and systemctl restart irqbalance are inside the lock; but no race condition.
- If I move the lock into updateNewIRQSMPAffinityMask, the locks guarantee that the SMP affinity file and irqbalance configuration are consistent. But there's a potential race (thread B overtaking thread A after the locked section, thread B running irqbalance --oneshot with new config, thread A running irqbalance --oneshot with the old config
I initially though I could avoid this by writing state to the file, then reading from the file with irqbalance --oneshot. But a) that changed the logic and required the irqbalance configutation file to be present. And b) it wouldn't get rid of the race

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +640,103 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
+	h.updateIRQSMPAffinityLock.Lock()
```
andreaskaris, line 689: Problem:

    If I move the lock here then irqbalance --oneshot and systemctl restart irqbalance are inside the lock; but no race condition.
    If I move the lock into updateNewIRQSMPAffinityMask, the locks guarantee that the SMP affinity file and irqbalance configuration are consistent. But there's a potential race (thread B overtaking thread A after the locked section, thread B running irqbalance --oneshot with new config, thread A running irqbalance --oneshot with the old config
    I initially though I could avoid this by writing state to the file, then reading from the file with irqbalance --oneshot. But a) that changed the logic and required the irqbalance configutation file to be present. And b) it wouldn't get rid of the race

I'm not sure if it's better to live with a potential race for irqbalance, or to move service restart / irqbalance --oneshot inside the lock

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +640,103 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
-
-	isIrqConfigExists := fileExists(h.irqBalanceConfigFile)
-
-	if isIrqConfigExists {
-		if err := h.updateIrqBalanceConfigFile(newIRQBalanceSetting); err != nil {
-			return err
-		}
+	// Now, restart the irqbalance service or run irqbalance --oneshot command.
+	// On failure, this will log errors but will not return them, as it is not critical for the pod to start.
+	if !h.handleIRQBalanceRestart(ctx, c.Name()) {
```
andreaskaris, line 698: I introduced a small logic change here: if systemctl restart irqbalance fails, the logic now falls back to irqbalance --oneshot

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +640,103 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
+	if !isServiceEnabled(irqBalancedName) || !fileExists(h.irqBalanceConfigFile) {
+		return false
 	}
 
+	log.Debugf(ctx, "Container %q restarting irqbalance service", cName)
+
 	if err := restartIrqBalanceService(); err != nil {
 		log.Warnf(ctx, "Irqbalance service restart failed: %v", err)
+
+		return false
```
andreaskaris, line 721: I introduced a small logic change here: if systemctl restart irqbalance fails, the logic now falls back to irqbalance --oneshot

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +640,103 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
+func (h *HighPerformanceHooks) handleIRQBalanceOneShot(ctx context.Context, cName, newIRQBalanceSetting string) {
+	irqBalanceFullPath, err := exec.LookPath(irqBalancedName)
+	if err != nil {
+		// irqbalance is not installed, skip the rest; pod should still start, so return nil instead.
+		log.Warnf(ctx, "Irqbalance binary not found: %v", err)
+
+		return
+	}
 
+	cmd := cmdrunner.Command(irqBalanceFullPath, "--oneshot")
```
andreaskaris: small change here, using the full path

andreaskaris:
@MarSik ready for review

internal/runtimehandlerhooks/high_performance_hooks_linux.go
MarSik: I wonder what the performance impact will be, but the race condition is the primary reason for this PR.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
MarSik: +1

andreaskaris:
@haircommander

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -61,12 +61,54 @@ const (
...
+	return isServiceEnabled(serviceName)
+}
+
+func (d *defaultServiceManager) RestartService(serviceName string) error {
+	return restartService(serviceName)
+}
+
+type defaultCommandRunner struct{}
+
+func (d *defaultCommandRunner) LookPath(file string) (string, error) {
```
haircommander, line 89: I'm open to these living in cmdrunner also, but this works too

internal/runtimehandlerhooks/high_performance_hooks_linux.go
haircommander: yeah inside the lock seems best to prevent races at this point

haircommander:
/ok-to-test
/approve

LGTM, one nit but needn't be done

andreaskaris:
@haircommander  "ci/kata-jenkinsExpected — Waiting for status to be reported"  is this stuck somehow?

haircommander:
nope just needs a lgtm label

@cri-o/cri-o-maintainers any takers?

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +686,106 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
 		return err
 	}
+	// Now, restart the irqbalance service or run irqbalance --oneshot command.
+	// On failure, this will log errors but will not return them, as it is not critical for the pod to start.
+	if !h.handleIRQBalanceRestart(ctx, c.Name()) {
+		h.handleIRQBalanceOneShot(ctx, c.Name(), newIRQBalanceSetting)
+	}
 
-	isIrqConfigExists := fileExists(h.irqBalanceConfigFile)
+	return nil
```
bitoku, line 702: Do we want to rollback irqSMPAffinityMask if both operations fail?

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +686,106 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
+	}
 
-func (h *HighPerformanceHooks) updateIrqBalanceConfigFile(newIRQBalanceSetting string) error {
-	h.irqBalanceConfigFileLock.Lock()
-	defer h.irqBalanceConfigFileLock.Unlock()
+	if err := updateIrqBalanceConfigFile(h.irqBalanceConfigFile, newIRQBalanceSetting); err != nil {
+		// Rollback IRQ SMP affinity file to maintain consistency
+		if rollbackErr := os.WriteFile(h.irqSMPAffinityFile, []byte(originalIRQSMPSetting), 0o644); rollbackErr != nil {
+			log.Errorf(ctx, "Failed to rollback IRQ SMP affinity file after config update failure: %v", rollbackErr)
+		}
```
bitoku: nit: I'd use defer instead of if in case we want to add some logics after here later.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: I've been trying now for a few hours at least to reorganize the code to include a rollback, and it always leads to a big change, and it always looks ugly. It also changes the behavior in my opinion ..
For one reason or another, originally it was decided that a restart should just log a warning and continue as is.
I'd argue that in case of a failed restart of irqbalance service / oneshot:
* it should either fail and rollback (would have been the better behavior IMO)
* or continue and not rollback (established behavior)

But continuing and rolling back will leave the configuration files in a bad state, on top of not restarting the service.

I'd say that failing and rolling back would have been the right thing to do, but that now changes the established behavior pretty significantly, I'd say let's not do it in this PR if you're o.k. with it

internal/runtimehandlerhooks/high_performance_hooks_linux.go
andreaskaris: done

andreaskaris:
Pretty sure the CI failure isn't because of my new changes. @bitoku  I addressed your nit, PTAL

internal/runtimehandlerhooks/high_performance_hooks_linux.go
bitoku: I don't have a strong opinion about that. I'm just curious why we don't.
If we have a sensible reason not to do, I'm ok.

internal/runtimehandlerhooks/high_performance_hooks_linux.go
```diff
@@ -631,69 +686,110 @@ func (h *HighPerformanceHooks) setIRQLoadBalancing(ctx context.Context, c *oci.C
...
-	return newIRQBalanceSetting, nil
-}
+	// Rollback IRQ SMP affinity file to maintain consistency if something goes wrong.
+	defer func() {
+		if err != nil {
+			if rollbackErr := os.WriteFile(h.irqSMPAffinityFile, []byte(originalIRQSMPSetting), 0o644); rollbackErr != nil {
+				log.Errorf(ctx, "Failed to rollback IRQ SMP affinity file after config update failure: %v", rollbackErr)
+			}
+		}
+	}()
```
bitoku: Let's log `err`, not only `rollbackErr`.

andreaskaris: err is being returned via the return statement (named `err error`); it's passed up the stack and will be logged in  high_performance_hooks_linux.go, e.g. in:
```
			log.Errorf(ctx, "Failed to run post-stop hook for container %s: %v", ctr.ID(), err)
```

I changed now though to log both errors, as well:
```
	defer func() {
		if err != nil {
			if rollbackErr := os.WriteFile(h.irqSMPAffinityFile, []byte(originalIRQSMPSetting), 0o644); rollbackErr != nil {
				log.Errorf(ctx, "Failed to rollback IRQ SMP affinity file after config update failure: err: %q, rollback err: %q",
					err, rollbackErr)
			}
		}
	}()
```

bitoku: aha you're right, my bad.

bitoku:
The test failures should be related to https://github.com/containers/crun/issues/1879. 
But I want to wait for a while until we fix the test if it doesn't take time to fix.

andreaskaris:
I suppose we're still waiting for https://github.com/containers/crun/issues/1879 ?

bitoku:
sorry I forgot this PR. we internally rollback the crun version, so it should pass now.
/retest

andreaskaris:
@haircommander PTAL and thank you!

sohankunkerkar:
@andreaskaris This PR has all the necessary labels. I think CI should merge this change shortly. Thanks for your patience!

PR #9407, opened 2025-08-15
Fix potential panic when crio fails to parse artifact's ref name.

bitoku:
#### What type of PR is this?
/kind bug

This PR fixes panic when the ref name is invalid although this shouldn't happen in a normal scenario, because the ref is specified based on docker-reference (`<registry>/<image-name>:<tag>`) by cri-o.

```release-note
None
```


saschagrunert:
@bitoku do we have to cherry-pick this one?

bitoku:
@saschagrunert thanks!
I don't think it's necessary, though no reason not to cherry-pick it.

PR #9410, opened 2025-08-19
[OCPBUGS-58229]: server: Fix network cleanup failures when NetNS path is empty

sohankunkerkar:
Fixes an issue where empty network namespace paths cause CNI teardown failures, preventing pod deletion and creating stuck pods. This happens when infra containers die, leaving NetNsPath() returning empty strings that CNI plugins cannot handle.

The failure cascade: dead infra container → empty NetNS → CNI failure → StopPodSandbox failure → stuck pod → systemd cgroup conflicts → new pod creation failures. Observed in MicroShift production environments.

#### What type of PR is this?

/kind bug

#### Special notes for your reviewer:

The network cleanup failure is blocking pod deletion, which is causing the systemd transaction conflicts.
```sh
kubelet E0819 16:05:51.010476 pod_workers.go:1301] "Error syncing pod, skipping" err="failed to \"KillPodSandbox\" [...] network teardown failed [...] failed to Statfs \"\": no such file or directory"
```

#### Does this PR introduce a user-facing change?

```release-note
server: Fix network cleanup failures when NetNS path is empty
```

PR #9411, opened 2025-08-19
inspect: add hostnetwork information

haircommander:
#### What type of PR is this?

/kind api-change
which could allow cadvisor to choose not to report network metrics

```release-note
Add HostNetwork field to the `/info` endpoint to inform cadvisor whether a container is in the host network or not.
```

### DIFF
--- a/server/inspect.go
+++ b/server/inspect.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-chi/chi/v5"
 	json "github.com/json-iterator/go"
 	"github.com/sirupsen/logrus"
+	"k8s.io/utils/ptr"
 
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/log"
@@ -130,6 +131,7 @@ func (s *Server) getContainerInfo(ctx context.Context, id string, getContainerFu
 		LogPath:         ctr.LogPath(),
 		Sandbox:         ctr.Sandbox(),
 		IPs:             sb.IPs(),
+		HostNetwork:     ptr.To(sb.HostNetwork()),
 	}, nil
 }
 
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -18,6 +18,7 @@ type ContainerInfo struct {
 	Root            string            `json:"root"`
 	Sandbox         string            `json:"sandbox"`
 	IPs             []string          `json:"ip_addresses"`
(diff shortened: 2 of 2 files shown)


haircommander:
inspired by https://issues.redhat.com/browse/OCPSTRAT-2346

server/inspect.go
```diff
@@ -130,6 +131,7 @@ func (s *Server) getContainerInfo(ctx context.Context, id string, getContainerFu
 		LogPath:         ctr.LogPath(),
 		Sandbox:         ctr.Sandbox(),
 		IPs:             sb.IPs(),
+		HostNetwork:     ptr.To(sb.HostNetwork()),
```
kannon92, line 134: I think this could potentially seg fault.

Should you use deref?

haircommander: how would it segfault? HostNetwork() returns a bool, and ptr.To() just creates a pointer out of that returned bool

haircommander:
critest failures reproduced here https://github.com/cri-o/cri-o/pull/9410 so unrelated

server/inspect.go
kannon92: HostNetwork is a pointer to a bool though.

couldn't that be nil?

haircommander: [sb.HostNetwork()](https://github.com/cri-o/cri-o/blob/main/internal/lib/sandbox/sandbox.go#L237) returns a bool, the HostNetwork we nest in the ContainerInfo field is a pointer. They're different

kannon92: Why the use of a bool pointer here?

In kube we would use it for marshaling non default fields but I don't see a reason to have this here.

kannon92: > [sb.HostNetwork()](https://github.com/cri-o/cri-o/blob/main/internal/lib/sandbox/sandbox.go#L237) returns a bool, the HostNetwork we nest in the ContainerInfo field is a pointer. They're different

I see. Yea I got confused.

sohankunkerkar: There's already a nil check for sb at lines 88-93

haircommander: > In kube we would use it for marshaling non default fields but I don't see a reason to have this here.

exactly this reason. Eventually cadvisor will consume this field, and we need to be able to differentiate between cri-o not supporting the field and the field being false.

sohankunkerkar:
> critest failures reproduced here #9410 so unrelated

Looks like the cri-tools CI is also broken with the same failure

PR #9351, opened 2025-07-16
[1.31][4.18] OCPBUGS-59416: HighPerformanceHooks: Fix IRQ SMP affinity race conditions

andreaskaris:
cherry-pick of https://github.com/cri-o/cri-o/pull/9350

#### Special notes for your reviewer:

The same as for the change to `main`, I ran a smoke test for sanity checking:

I ran my smoke test on a VM with 14 CPUs where I first installed the entire stack with kubeadm.

Dependencies:
```
# rpm -qa | grep kubelet
kubelet-1.33.1-150500.1.1.x86_64
# rpm -qa | grep kubeadm
kubeadm-1.33.1-150500.1.1.x86_64
# rpm -qa | grep crun
crun-1.21-1.el9.x86_64
# rpm -qa | grep conmon
conmon-2.1.13-1.el9.x86_64
```

/etc/crio/crio.conf.d/99-runtimes.conf
```
[crio.runtime]
infra_ctr_cpuset = "0-3"




# The CRI-O will check the allowed_annotations under the runtime handler and apply high-performance hooks when one of
# high-performance annotations presents under it.
# We should provide the runtime_path because we need to inform that we want to re-use runc binary and we
# do not have high-performance binary under the $PATH that will point to it.
[crio.runtime.runtimes.high-performance]
inherit_default_runtime = true
allowed_annotations = ["cpu-load-balancing.crio.io", "cpu-quota.crio.io", "irq-load-balancing.crio.io", "cpu-c-states.crio.io", "cpu-freq-governor.crio.io"]
```

```
# tail -n 5 /var/lib/kubelet/config.yaml
cpuManagerPolicy: static
cpuManagerPolicyOptions:
  full-pcpus-only: "true"
cpuManagerReconcilePeriod: 5s
reservedSystemCPUs: 0-3
```

I then stopped the crio service, started the compiled crio in foreground `make binaries && bin/crio`, and ran this smoke test:

smoke.sh
```
#!/bin/bash

set -x

affinity_file="/proc/irq/default_smp_affinity"
expected_reset_affinity="3fff"
expected_mask="3e0f"

echo $expected_reset_affinity > $affinity_file
cat $affinity_file

for i in {0..20}; do
	set +x
	echo "========"
	echo "Run ${i}"
	echo "========"
	set -x
	kubectl apply -f pod.yaml
	kubectl wait --for=condition=Ready pod/qos-demo --timeout=180s
	mask=$(cat ${affinity_file} | tr -d '\n')
[...]

PR #9422, opened 2025-08-25
server: run post stop hooks before updating container status

haircommander:
/kind bug
ContainerStateToDisk will call UpdateContainerStatus which will update the internal representation of the container state to be stopped. While correct, it opens a race where the kubelet may take action on a container that seems stopped before post stop hooks are run.

```release-note
Fix a bug where CRI-O reports a container is stopped before post stop hooks are run 
```

### DIFF
--- a/server/container_remove_test.go
+++ b/server/container_remove_test.go
@@ -31,8 +31,6 @@ var _ = t.Describe("ContainerRemove", func() {
 				State: specs.State{Status: oci.ContainerStateStopped},
 			})
 			gomock.InOrder(
-				runtimeServerMock.EXPECT().StopContainer(gomock.Any(), gomock.Any()).
-					Return(nil),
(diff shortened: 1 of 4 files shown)


haircommander:
cc @andreaskaris @MarSik

andreaskaris:
@haircommander  https://github.com/cri-o/cri-o/pull/9405 merged now, so I guess it's time to revisit this one, here :-)

andreaskaris:
So I applied this on top of OCPBUGS-59416 (https://github.com/cri-o/cri-o/pull/9351):
https://github.com/andreaskaris/cri-o/tree/OCPBUGS-59416-pr9422

It works fine, but I still have no way of testing the original customer issue, as PR9351 by itself fixes my own test case, too (meaning that my reproducer doesn't catch 100% of possible issues)

haircommander:
/cloes

carried in https://github.com/cri-o/cri-o/pull/9613

PR #9424, opened 2025-08-27
Add runtime handler seccomp profile

### LINKED ISSUE #5715: Configure seccomp `RuntimeDefault` profile based on `RuntimeClass`
As mentioned in the Kubernetes seccomp graduation KEP, it would be helpful to separate the default seccomp profile based on the use runtime class: https://github.com/kubernetes/enhancements/issues/2413#issuecomment-1034358366

We could think about adding such a feature if we decide to not modify the RuntimeClass object in Kubernetes.

gavinkflam:
#### What type of PR is this?

/kind feature

Allow configuration of a default seccomp profile per runtime.

Fixes #5715

```release-note
Allow configuration of a default seccomp profile per runtime.
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -288,13 +288,23 @@ type RuntimeHandler struct {
 	//   When it has only exclusive cpuset, use the first CPU in the exclusive cpuset.
 	//   When it has both shared and exclusive cpusets, use first CPU in the shared cpuset.
 	ExecCPUAffinity ExecCPUAffinityType `toml:"exec_cpu_affinity,omitempty"`
+
+	// SeccompProfile is the absolute path of the seccomp.json profile which is used as the
+	// default for the runtime. This configuration takes precedence over runtime config seccomp_profile.
+	// If set to "", the runtime config seccomp_profile will be used.
+	// If that is also set to "", the internal default seccomp profile will be applied.
+	SeccompProfile string `toml:"seccomp_profile,omitempty"`
+
+	// seccompConfig is the seccomp configuration for the handler.
+	seccompConfig *seccomp.Config
 }
 
 type ExecCPUAffinityType string
 
 const (
-	ExecCPUAffinityTypeDefault ExecCPUAffinityType = ""
-	ExecCPUAffinityTypeFirst   ExecCPUAffinityType = "first"
+	ExecCPUAffinityTypeDefault   ExecCPUAffinityType = ""
+	ExecCPUAffinityTypeFirst     ExecCPUAffinityType = "first"
+	runtimeSeccompProfileDefault string              = ""
 )
 
 // Multiple runtime Handlers in a map.
@@ -379,6 +389,7 @@ type RuntimeConfig struct {
 
 	// SeccompProfile is the seccomp.json profile path which is used as the
 	// default for the runtime.
+	// If set to "" or not found, the internal default seccomp profile will be used.
 	SeccompProfile string `toml:"seccomp_profile"`
 
 	// PrivilegedSeccompProfile can be set to enable a seccomp profile for
@@ -1064,6 +1075,14 @@ func (c *Config) Validate(onExecution bool) error {
 		filepath.Join(filepath.Dir(c.Listen), "seccomp"),
 	)
 
+	for name := range c.Runtimes {
+		if c.Runtimes[name].seccompConfig != nil {
+			c.Runtimes[name].seccompConfig.SetNotifierPath(
+				filepath.Join(filepath.Dir(c.Listen), "seccomp"),
+			)
+		}
+	}
+
 	if err := c.ImageConfig.Validate(onExecution); err != nil {
 		return fmt.Errorf("validating image config: %w", err)
 	}
@@ -1310,15 +1329,20 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 			logrus.Infof("Checkpoint/restore support disabled via configuration")
 		}
 
-		if err := c.seccompConfig.LoadProfile(c.SeccompProfile); err != nil {
+		if c.SeccompProfile == "" {
+			if err := c.seccompConfig.LoadDefaultProfile(); err != nil {
+				return fmt.Errorf("unable to load default seccomp profile: %w", err)
+			}
+		} else if err := c.seccompConfig.LoadProfile(c.SeccompProfile); err != nil {
 			if !errors.Is(err, os.ErrNotExist) {
 				return fmt.Errorf("unable to load seccomp profile: %w", err)
 			}
 
-			logrus.Info("Specified profile does not exist on disk")
+			// Fallback to the internal default in order not to break upgrade paths.
+			logrus.Info("Seccomp profile does not exist on disk, fallback to internal default profile")
 
 			if err := c.seccompConfig.LoadDefaultProfile(); err != nil {
-				return fmt.Errorf("load default seccomp profile: %w", err)
+				return fmt.Errorf("unable to load default seccomp profile: %w", err)
 			}
 		}
 
@@ -1394,6 +1418,7 @@ func defaultRuntimeHandler(isSystemd bool) *RuntimeHandler {
 		ContainerMinMemory: units.BytesSize(defaultContainerMinMemoryCrun),
 		MonitorCgroup:      getDefaultMonitorGroup(isSystemd),
 		ExecCPUAffinity:    ExecCPUAffinityTypeDefault,
+		SeccompProfile:     runtimeSeccompProfileDefault,
 	}
 }
 
@@ -1765,6 +1790,10 @@ func (r *RuntimeHandler) Validate(name string) error {
 		return err
 	}
 
+	if err := r.validateRuntimeSeccompProfile(); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -1989,6 +2018,11 @@ func (r *RuntimeHandler) RuntimeStreamWebsockets() bool {
 	return r.StreamWebsockets
 }
 
+// RuntimeSeccomp returns the configuration of the loaded seccomp profile for this handler.
+func (r *RuntimeHandler) RuntimeSeccomp() *seccomp.Config {
+	return r.seccompConfig
+}
+
 // validateRuntimeExecCPUAffinity checks if the RuntimeHandler enforces proper CPU affinity settings.
 func (r *RuntimeHandler) validateRuntimeExecCPUAffinity() error {
 	switch r.ExecCPUAffinity {
@@ -1997,6 +2031,22 @@ func (r *RuntimeHandler) validateRuntimeExecCPUAffinity() error {
 	}
 
 	return fmt.Errorf("invalid exec_cpu_affinity %q", r.ExecCPUAffinity)
+}
+
+// validateRuntimeSeccompProfile tries to load the RuntimeHandler seccomp profile.
+func (r *RuntimeHandler) validateRuntimeSeccompProfile() error {
+	if r.SeccompProfile == "" {
+		r.seccompConfig = nil
+
+		return nil
+	}
+
+	r.seccompConfig = seccomp.New()
+	if err := r.seccompConfig.LoadProfile(r.SeccompProfile); err != nil {
+		return fmt.Errorf("unable to load runtime handler seccomp profile: %w", err)
+	}
+
+	return nil
 }
 
 func validateAllowedAndGenerateDisallowedAnnotations(allowed []string) (disallowed []string, _ error) {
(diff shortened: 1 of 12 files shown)


gavinkflam:
/cc saschagrunert haircommander

Please take a look if you get a chance.

pkg/config/config.go
```diff
@@ -288,13 +288,18 @@ type RuntimeHandler struct {
 	//   When it has only exclusive cpuset, use the first CPU in the exclusive cpuset.
 	//   When it has both shared and exclusive cpusets, use first CPU in the shared cpuset.
 	ExecCPUAffinity ExecCPUAffinityType `toml:"exec_cpu_affinity,omitempty"`
+
+	// SeccompProfile is the seccomp.json profile path which is used as the
+	// default for the runtime.
+	SeccompProfile string `toml:"seccomp_profile,omitempty"`
```
saschagrunert, line 296: Should we also allow `RuntimeDefault` as fixed value here or only absolute paths?

pkg/config/config.go
```diff
@@ -1999,6 +2009,19 @@ func (r *RuntimeHandler) validateRuntimeExecCPUAffinity() error {
 	return fmt.Errorf("invalid exec_cpu_affinity %q", r.ExecCPUAffinity)
 }
 
+// ValidateRuntimeSeccompProfile checks if the RuntimeHandler seccomp profile exists.
+func (r *RuntimeHandler) ValidateRuntimeSeccompProfile(name string) error {
```
saschagrunert: Please verify that it's an absolute path.

pkg/config/template.go
```diff
@@ -1300,6 +1301,8 @@ const templateStringCrioRuntimeRuntimesRuntimeHandler = `# The "crio.runtime.run
 #   when a machine crash happens.
 # - default_annotations (optional, map): Default annotations if not overridden by the pod spec.
 # - stream_websockets (optional, bool): Enable the WebSocket protocol for container exec, attach and port forward.
+# - seccomp_profile (optional, string): The path of the seccomp.json profile which is used as the default seccomp profile
```
saschagrunert:
```suggestion
# - seccomp_profile (optional, string): The absolute path of the seccomp.json profile which is used as the default seccomp profile
```

test/seccomp_runtime_handler.bats
```diff
@@ -0,0 +1,49 @@
...
+	start_crio_no_setup
+
+	CTR=$(crictl run "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json)
+
+	# Assert
+	grep -q "Applied runtime handler seccomp profile" "$CRIO_LOG"
+	crictl inspect "$CTR" | jq -e .info.runtimeSpec.linux.seccomp | grep -q $TEST_SYSCALL
+}
+
+@test "seccomp runtime handler profile file is invalid" {
```
saschagrunert: let's split this test case into
- not existing path
- relative path

saschagrunert (commented):
Thanks! I just have a few suggestions, otherwise LGTM

pkg/config/config.go
gavinkflam: Is "RuntimeDefault" different from ""? I.e. fallback to RuntimeConfig seccomp_profile or internal default.

test/seccomp_runtime_handler.bats
gavinkflam: Added two more test cases.

pkg/config/template.go
gavinkflam: Done.

pkg/config/config.go
gavinkflam: Added absolute path validation and a unit test.

pkg/config/config.go
saschagrunert: What do other @cri-o/cri-o-maintainers think?

bitoku: It looks ok to me, but want it to say that this supersedes SeccompProfile in RuntimeConfig, and fall back to it if it's empty.
I'm wondering if we want the one for the privileged pod, which is in RuntimeConfig.

gavinkflam: I’ve revised multiple comments to better explain the precedence.

`PrivilegedSeccompProfile` - I'd be glad to submit a follow-up pull request if that is what we want.

gavinkflam:
@saschagrunert @bitoku Please take another look. I realized that I should only apply the profile when the profile type is `RuntimeDefault`.

I've corrected that and added two tests to verify the profile type behavior and fallback logic.

pkg/config/config.go
```diff
@@ -379,6 +386,7 @@ type RuntimeConfig struct {
 
 	// SeccompProfile is the seccomp.json profile path which is used as the
 	// default for the runtime.
+	// If set to "", the internal default seccomp profile will be used.
```
bitoku: thanks!👍

pkg/config/config.go
```diff
@@ -1999,6 +2012,23 @@ func (r *RuntimeHandler) validateRuntimeExecCPUAffinity() error {
...
+	if !filepath.IsAbs(r.SeccompProfile) {
+		return fmt.Errorf("seccomp_profile for runtime '%s' is not absolute: %s", name, r.SeccompProfile)
+	}
+
+	if _, err := os.Stat(r.SeccompProfile); err != nil && os.IsNotExist(err) {
+		return fmt.Errorf("invalid seccomp_profile for runtime '%s': %w", name, err)
+	}
+
+	return nil
+}
```
bitoku: I think this validation is also wanted. we can replace os.Stat check with this.
https://github.com/cri-o/cri-o/blob/51ce3dd659becb9d31c2752cf11c4411e3b18c22/pkg/config/config.go#L1313-L1318

pkg/config/config.go
bitoku: `PrivilegedSeccompProfile` is required for https://github.com/kubernetes-sigs/security-profiles-operator to catch syscalls. Although I brought it up, I don't think we need it. I mentioned it just in case. we can leave it for now.

pkg/config/config.go
gavinkflam: I can work on aligning `RuntimeHandler.SeccompProfile` with this.

`LoadProfile` does more than just validate - it checks whether seccomp is enabled on the host and unmarshalls the file into a `Seccomp` struct. However, it also handles the fallback to the internal default profile when the path is `""`, which needs to be moved into `config.go` for reuse.

I believe loading the profile upfront has clear benefits. It allows us to catch invalid profiles when configurations are loaded rather than at container creation. Moreover, it ensures consistent and predictable behavior as its `RuntimeConfig` counterpart and simplifies maintenance.

gavinkflam: This is done.

gavinkflam:
I have aligned the implementation with `RuntimeConfig.SeccompProfile`. This ensures consistent and predictable behavior as the `RuntimeConfig` counterpart. Moreover, loading the profile upfront catches invalid profiles during configuration loading rather than at container creation.

A notable behavior change is that if the file does not exist, it now falls back to the internal default instead of returning an error. This is consistent with `RuntimeConfig`, where the fallback logic was introduced in #6029.

@saschagrunert @bitoku Please take another look when you get a chance.

pkg/config/config.go
```diff
@@ -1310,15 +1321,20 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
...
+				return fmt.Errorf("unable to load default seccomp profile: %w", err)
+			}
+		} else if err := c.seccompConfig.LoadProfile(c.SeccompProfile); err != nil {
 			if !errors.Is(err, os.ErrNotExist) {
 				return fmt.Errorf("unable to load seccomp profile: %w", err)
 			}
 
-			logrus.Info("Specified profile does not exist on disk")
+			// Fallback to the internal default in order not to not break upgrade paths.
+			logrus.Info("Seccomp profile does not exist on disk, fallback to internal default profile")
```
bitoku: Is this typo? The comment has duplicate "not". same for other places having this comment.

btw, @saschagrunert do you know why does it fall back to the default?
It seems introduced in this PR. https://github.com/cri-o/cri-o/pull/6029
I think it should raise an error, otherwise there's a risk of having a different seccomp profile without noticing it.

pkg/config/config.go
```diff
@@ -1999,6 +2025,30 @@ func (r *RuntimeHandler) validateRuntimeExecCPUAffinity() error {
 	return fmt.Errorf("invalid exec_cpu_affinity %q", r.ExecCPUAffinity)
 }
 
+// validateRuntimeSeccompProfile tries to load the RuntimeHandler seccomp profile.
+func (r *RuntimeHandler) validateRuntimeSeccompProfile() error {
+	if r.SeccompProfile == "" {
+		r.seccompConfig = nil
+	} else {
```
bitoku: nit: it can early-return and can reduce a nesting level.
```suggestion
	if r.SeccompProfile == "" {
		r.seccompConfig = nil
		return nil
	}
```

gavinkflam: Done. However, our linters insist adding an empty line before return statement which isn't ideal in this case.

pkg/config/config.go
gavinkflam: Good catch. I've fixed the typo.

I'm fine with either returning an error or falling back to internal default. @saschagrunert WDYT?

saschagrunert: That was intentional on upgrade, because we stopped shipping the default profile in the CRI-O rpm. We have to clarify if that's still required (I don't think so), then we can change the code.

We can also open an issue, clarify downstream (I can help with that) and change the logic as follow-up.

bitoku: Thanks! that makes sense.
Yeah I'll open an issue to track it.

bitoku (commented):
seccomp `Setup()` should be closed to the receiver and shouldn't accept other `Config`.
Instead, how about returning the right seccomp in `Runtime.Seccomp()`?
I think this will make the code simpler.

Also this is nit, but it seems we can create a new function to validate seccomp. They look almost same.

pkg/config/config.go
gavinkflam: Thank you. I have changed `runtimeHandler.SeccompProfile` to return an error instead. Since this is a new configuration item and not constrained by the upgrade compatibility issue.

gavinkflam:
@bitoku To use `RuntimeHandler.seccompConfig` directly, we’d need to set up its `notifierPath` correctly. That requires `APIConfig.Listen`, which would mean passing the root `Config` struct deep into `validateRuntimeSeccompProfile()` - not ideal.

Instead, I’ve updated the config and `Setup()` to work with the `seccomp.Seccomp` struct directly. This makes the implementation simpler and more natural.

Let me know what do you think.

bitoku (commented):
> To use RuntimeHandler.seccompConfig directly, we’d need to set up its notifierPath correctly. That requires APIConfig.Listen, which would mean passing the root Config struct deep into validateRuntimeSeccompProfile() - not ideal.

I don't think that using rootConfig is not ideal.
Cascading the value deep into `validateRuntimeSeccompProfile()` may look messy, but we can do just like this instead.
https://github.com/cri-o/cri-o/blob/51ce3dd659becb9d31c2752cf11c4411e3b18c22/pkg/config/config.go#L1063-L1065
Ideally, now that runtimeHandler has seccomp profile, everything should rely on runtimeHandler seccomp profile.

But this is more style suggestion, and if there's a reason not to do that, it's not a blocker of this PR.
We could also follow up if we find a better way.

/lgtm

server/sandbox_run_linux.go
```diff
@@ -866,6 +866,7 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 			"",
 			nil,
 			nil,
+			nil,
```
bitoku: Can we set it nil?

bitoku (commented):
/lgtm cancel
sorry there's one thing I missed.

server/sandbox_run_linux.go
gavinkflam: Good catch. I've implemented this for pods and added a test.

gavinkflam:
@bitoku Please take another look.

I've added pod support, and implemented your suggestion of having `Runtime.Seccomp()` returning the right `seccomp.Config`.

pkg/config/config.go
```diff
@@ -288,13 +288,23 @@ type RuntimeHandler struct {
 	//   When it has only exclusive cpuset, use the first CPU in the exclusive cpuset.
 	//   When it has both shared and exclusive cpusets, use first CPU in the shared cpuset.
 	ExecCPUAffinity ExecCPUAffinityType `toml:"exec_cpu_affinity,omitempty"`
+
+	// SeccompProfile is the absolute path of the seccomp.json profile which is used as the
+	// default for the runtime. This configuration takes precedence over runtime config seccomp_profile.
+	// If set to "", the runtime config seccomp_profile will be used.
+	// If that is also set to "", the internal default seccomp profile will be applied.
```
bitoku, line 295: 👍 When the seccomp profile is not found, it should fail. I like this behaviour.
We don't have to consider the backward compatibility which RuntimeConfig.SeccompProfile has, because it's a new config.

bitoku:
@saschagrunert Can you PTAL?

saschagrunert (approved):

PR #9433, opened 2025-08-28
runtime_vm: Implement the ReopenContainerLog function

littlejawa:
#### What type of PR is this?

/kind bug

This commit implements the ReopenContainerLog function. This fixes an issue where kata container logs could not be rotated.

This required duplicating part of the createContainerIO function, so I am moving this in its own subfunction for reuse.

None

```release-note
Fix log rotation not working for containers running with the kata-containers runtime
```


bitoku:
Thank you @littlejawa !
Can we have an integration test for it?

littlejawa:
> Can we have an integration test for it?

Hey @bitoku - 
I have added a test for log rotation. 
Without the patch in this commit, this test passes for regular containers and fails for kata.
With the patch applied, the test passes for both runtimes.

haircommander:
/approve

LGTM, thanks!

littlejawa:
Thanks @haircommander @bitoku!

Can this be backported to previous versions?
We'd need to get it on 1.31 and later

haircommander:
yup!

PR #9456, opened 2025-09-05
[WIP] OCPBUGS-61220: internal/cgmgr: fix CPU weight setting for cgroup v2

sohankunkerkar:
It resolves Pod InPlace Resize Container test failure where cpu.weight file was empty for Guaranteed QoS pods with restartable init containers. Sometimes, systemd fails to convert CPUShares → CPUWeight for this specific scenario, likely due to how it handles complex pod hierarchies with concurrent containers, but the exact internal mechanism is unclear. This acts more as a safeguard in case systemd doesn’t set the value.

#### What type of PR is this?

/kind bug
/kind flake

Fixes https://issues.redhat.com/browse/OCPBUGS-61220

```release-note
Use systemd property as fallback for cpu weight
```

### DIFF
--- a/internal/config/cgmgr/systemd_linux.go
+++ b/internal/config/cgmgr/systemd_linux.go
@@ -192,10 +192,23 @@ func (m *SystemdManager) MoveConmonToCgroup(cid, cgroupParent, conmonCgroup stri
 		}
 
 		if resources.CPU.Shares != nil {
-			props = append(props, systemdDbus.Property{
-				Name:  "CPUShares",
-				Value: dbus.MakeVariant(resources.CPU.Shares),
-			})
+			// For cgroup v2, systemd should automatically convert CPUShares to CPUWeight,
+			// but there are cases where this conversion doesn't work properly.
+			// To ensure compatibility, we explicitly set CPUWeight for cgroup v2 systems.
+			if node.CgroupIsV2() {
+				cpuWeight := cgroups.ConvertCPUSharesToCgroupV2Value(*resources.CPU.Shares)
+				if cpuWeight > 0 {
+					props = append(props, systemdDbus.Property{
+						Name:  "CPUWeight",
+						Value: dbus.MakeVariant(cpuWeight),
+					})
+				}
+			} else {
+				props = append(props, systemdDbus.Property{
+					Name:  "CPUShares",
+					Value: dbus.MakeVariant(resources.CPU.Shares),
+				})
+			}
 		}
 
 		if resources.CPU.Quota != nil {


internal/config/cgmgr/systemd_linux.go
```diff
@@ -25,6 +25,22 @@ import (
 
 const defaultSystemdParent = "system.slice"
 
+// convertCPUSharesToSystemdWeight converts CPU shares to systemd's CPUWeight
+// using systemd's linear conversion formula: weight = shares * 100 / 1024
+func convertCPUSharesToSystemdWeight(shares uint64) uint64 {
```
haircommander: what about using https://github.com/opencontainers/cgroups/blob/6a793b6ea778415b3c0f7a3d2efdb70c786fad92/utils.go#L425?

internal/config/cgmgr/systemd_linux.go
```diff
@@ -192,10 +208,20 @@ func (m *SystemdManager) MoveConmonToCgroup(cid, cgroupParent, conmonCgroup stri
 		}
 
 		if resources.CPU.Shares != nil {
-			props = append(props, systemdDbus.Property{
-				Name:  "CPUShares",
-				Value: dbus.MakeVariant(resources.CPU.Shares),
-			})
+			if node.CgroupIsV2() {
```
haircommander: I'm curious why the IPPR test fails when conmon cgroup doesn't have cpu weight set. it's checking the pod cgroup weight and that's the culmination of its child cgroups?

internal/config/cgmgr/systemd_linux.go
sohankunkerkar: I initially tried that function, but it failed the workload.bats test, specifically, "test workload pod gets configured to defaults". The test expects CPUWeight=19 (from systemd’s linear formula: 200 * 100 / 1024 = 19), but the OCI formula produced CPUWeight=29 (from the logarithmic conversion).

internal/config/cgmgr/systemd_linux.go
sohankunkerkar: I think the pod-level cgroup aggregates CPU weights from all its children, including conmon. If conmon’s cgroup has no cpu.weight, it defaults to 100, which can make the pod-level weight inconsistent. This will avoid that issue.

internal/config/cgmgr/systemd_linux.go
haircommander: that's okay I think we should change the test and use the new function. there are other cases we hardcoded values in the tests but as time is moving on we're converting to this function more. see https://github.com/kubernetes/kubernetes/issues/131216 for more

internal/config/cgmgr/systemd_linux.go
bitoku: Looking at the log, it checked `/sys/fs/cgroup/cpu.weight` in the target container (c1-init) and it was empty.
Is it a pod level cgroup? It looks like container-level.

haircommander: yeah sounds like container level. @sohankunkerkar this fix did fix the test for you though?

internal/factory/container/container.go
```diff
@@ -849,7 +850,20 @@ func (c *container) SpecSetLinuxContainerResources(resources *types.LinuxContain
 	specgen := c.Spec()
 	specgen.SetLinuxResourcesCPUPeriod(uint64(resources.GetCpuPeriod()))
 	specgen.SetLinuxResourcesCPUQuota(resources.GetCpuQuota())
-	specgen.SetLinuxResourcesCPUShares(uint64(resources.GetCpuShares()))
+
+	shares := uint64(resources.GetCpuShares())
+	if node.CgroupIsV2() {
```
haircommander: nit: you could reverse these conditions

```
if !node.CgroupIsV2() {
} else if shares > 0 { 
}
```

bitoku:
closing this issue since the jira is closed.

PR #9463, opened 2025-09-11
Consume additional pull auth if available

saschagrunert:
#### What type of PR is this?

/kind feature

CRI-O will override the provided auth if a file for a namespace is available, which got written by a third party [credential provider](https://github.com/saschagrunert/credential-provider).

None
None

```release-note
Added support for the [namespaced pull secret credential provider](https://github.com/cri-o/credential-provider).
```


server/image_pull.go
```diff
@@ -136,11 +136,23 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
...
+		defer func() {
+			if err := os.Remove(sourceCtx.AuthFilePath); err != nil {
+				log.Warnf(ctx, "Unable to remove short-lived auth file: %v", err)
+
+				return
+			}
+
+			log.Debugf(ctx, "Removed short-lived auth file: %s", sourceCtx.AuthFilePath)
+		}()
+	}
```
saschagrunert: @QiWang19 we should probably create an auth file per image to avoid bad races within a namespace.

QiWang19: We can create a deterministic filename based on the image, how about `<sha256(req.Image)>-auth.json`, hash the req.Image value from CredentialProviderRequest?

saschagrunert: Yeah we still need the namespace as part of the filename, I decided to go with `<NAMESPACED_AUTH_DIR>/<NAMESPACE>-<IMAGE_NAME_SHA256>.json` for now.

saschagrunert:
Ready for review. PTAL @cri-o/cri-o-maintainers

pkg/config/config.go
```diff
@@ -568,6 +568,15 @@ type ImageConfig struct {
 	// containing credentials necessary for pulling images from secure
 	// registries.
 	GlobalAuthFile string `toml:"global_auth_file"`
+	// NamespacedAuthDir is the root path for pod namespace-separated
+	// auth files. The final auth file will be <NAMESPACED_AUTH_DIR>/<NAMESPACE>-<IMAGE_NAME_SHA256.json.
```
bitoku: `<NAMESPACED_AUTH_DIR>/<NAMESPACE>/<IMAGE_NAME_SHA256>.json` may be better in case users want to have directory based access control, or am I paranoid?

pkg/config/config.go
```diff
@@ -568,6 +568,15 @@ type ImageConfig struct {
...
+	// The final auth values will be used instead of any configured GlobalAuthFile.
+	// The auth file will be automatically removed by CRI-O after its usage.
```
bitoku: Why do we need to delete them each time?
Is it because this option is supposed to be used with the credential-provider?

pkg/config/config.go
saschagrunert: I was thinking about the same, but don't see much benefit compared to the whole directory creation and cleanup handling. The files in `<NAMESPACED_AUTH_DIR>` have to be `0600` for in any case. The critical point is leaking credentials from one Kubernetes namespace to the other, which we avoid by the `<NAMESPACE>-` prefix if a user specifies the same image from namespace `A` and `B`. The `IMAGE_NAME_SHA256` is just for avoiding cleanup races, because the kubelet will already take care not not pulling the same image in parallel on a node.

pkg/config/config.go
saschagrunert: Exactly, the credential provider will not be called again after pull, so CRI-O has to take care of the file cleanup.

internal/criocli/criocli.go
```diff
@@ -738,6 +742,14 @@ func getCrioFlags(defConf *libconfig.Config) []cli.Flag {
 			EnvVars:   []string{"CONTAINER_GLOBAL_AUTH_FILE"},
 			TakesFile: true,
 		},
+		&cli.StringFlag{
+			Name:      "namespaced-auth-dir",
+			Usage:     "Path to the root directory for namespaced auth files. Must be an absolute path.",
+			Value:     defConf.NamespacedAuthDir,
+			EnvVars:   []string{"CONTAINER_NAMESPACED_AUTH_DIR"},
+			TakesFile: true,
+			Hidden:    true,
```
saschagrunert, line 751: I think it's important to hide this option but allow setting it for testing purposes.

pkg/config/config.go
bitoku: I wonder if this feature will be used without credential providers... like manually putting files.

How is it supposed to work if the image is removed somehow, or if the pods have pullPolicy always?

pkg/config/config.go
```diff
@@ -568,6 +568,15 @@ type ImageConfig struct {
...
+	// If no pod namespace is being provided on image pull (via the sandbox
+	// config), or the concatenated path is non existent, then the
+	// system wide auth file will be used as fallback.
```
bitoku: Is this feature for extending the way of authentication, or for restricting access in specific namespaces?
I'm curious whether users want it to fall back to global auth or just fail, when there's no auth file for the image though there are some auth files for other images in the same namespace.

pkg/config/config.go
saschagrunert: A pull policy of always will re-trigger the credential provider. Every initiated pull from the kubelet should re-trigger the credential provider (for the matched images), because it does not actually return anything useful for the kubelet but write the file.

pkg/config/config.go
saschagrunert: We pre-merge everything with the global auth file in the credential provider, which is simpler than doing it in CRI-O. We could also extend the c/image API for that, but I think for the initial iteration of that feature that's too much.

pkg/config/config.go
bitoku: Thank you. That sounds like this race condition could happen.
Let's say there are two pods using the same image in the same namespace on the same node, and they have pull policy = always.
```
(pod1) CP: create credentials for the image
(pod1) cri-o: pull the image
(pod2) CP: create credentials for the image (no op because there's already credentials)
(pod1) cri-o: finish pulling the image and delete credentials
(pod2) cri-o: pull the image <-- credentials don't exist anymore, what happens in this case?
```

saschagrunert: The kubelet should take care not pulling the same image twice in parallel on a single node. And if a third party comes into that game, then the kubelet should re-initiate the pull cycle in the next backoff phase.

bitoku: ah sorry you explained it in a different thread.

- race condition will be taken care of by kubelet
- this options is supposed to be used with credential provider.

They are worth mentioning in the codebase as well.

internal/criocli/criocli.go
bitoku: I think we want to remove it from completions as well, and we not want to write a release note about it, maybe?

pkg/config/config.go
bitoku:
```suggestion
	// auth files. The final auth file will be <NAMESPACED_AUTH_DIR>/<NAMESPACE>-<IMAGE_NAME_SHA256>.json.
```

bitoku (commented):
Thank you for helping me understand the feature.
Overall lgtm, but a few non-design suggestions.

internal/criocli/criocli.go
saschagrunert: We also have other hidden options as part of the completions like `--registries-conf`, so the CI will complain when I manually remove it. :) I'd say we can keep it to understand what is possible via the completions.

pkg/config/config.go
saschagrunert: Alright, I updated the comment and linked the docs. :+1:

pkg/config/config.go
saschagrunert: Fixed as suggested :+1:

saschagrunert:
I updated the release note to just state that we now support the credential provider. More docs will come as part of the credential-provider project.

bitoku (commented):
/lgtm
/hold 
I want to keep it open for a few days to make room for others to review.
@cri-o/cri-o-maintainers

server/image_pull.go
```diff
@@ -206,19 +220,34 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
...
+			return sysCtx, fmt.Errorf("read policy path %s: %w", policyPath, err)
+		}
+
+		// Normalize the image ref to use the same format as the credential provider
+		image, err := reference.ParseDockerRef(imageRef)
+		if err != nil {
+			return sysCtx, fmt.Errorf("parse image name: %w", err)
+		}
+		// Follow the strict format of <NAMESPACE>-<IMAGE_NAME_SHA256>.json to resolve possible auth files.
+		hash := sha256.Sum256([]byte(image.String()))
```
QiWang19: I did some testing and found that the hash created by the provider and crio are different if the image in the pod spec is a non-latest tag, like `image: docker.io/nginx:1.29.1`
I think we need to trim the tag or digest of the `image.String()` before generating the hash. 
Or fix in the provider to get the correct tag or digest from the pod: https://github.com/cri-o/credential-provider/blob/1470511e9376cd14113cbe193ca3564dee5dd32f/internal/pkg/app/app.go#L52. the original `CredentialProviderRequest.Image` only contains the repo, without tag or digest, the `reference.ParseDockerRef` would always add the tag `latest`

> The kubelet should take care not pulling the same image twice in parallel on a single node.

We need to include the exact tag or digest in the auth file path to take advantage of this kubelet feature?

saschagrunert: @QiWang19 we use `ParseDockerRef` here and in the credential provider, so I assumed we closed that gap already. Is there anything I'm missing?

pkg/config/config.go
```diff
@@ -568,6 +568,23 @@ type ImageConfig struct {
...
+	// NamespacedAuthDir is the root path for pod namespace-separated
+	// auth files, which is intended to be used together with CRI-O's credential provider:
+	// https://github.com/cri-o/credential-provider
+	// The final auth file will be <NAMESPACED_AUTH_DIR>/<NAMESPACE>-<IMAGE_NAME_SHA256>.json.
+	// This temporary auth file will be used instead of any configured
+	// GlobalAuthFile. The file will be automatically removed by CRI-O
+	// after its usage on image pull. This means that there are theoretically
+	// race conditions possible within a single namespace, if the same image
+	// get's pulled at the same time. The kubelet will avoid those races
+	// by managing which images should be pulled in parallel.
```
saschagrunert: @QiWang19 @bitoku I did a bit more research and saw that we have an outstanding issue when the kubelet parallel image pull feature is being used: https://github.com/kubernetes/kubernetes/issues/125164

We concluded in my approach to fix that within the kubelet (https://github.com/kubernetes/kubernetes/issues/125164) that the runtime should take care of handling parallel pulls accordingly.

I think CRI-O already handles it correctly: If multiple pull RPCs are coming in at the same time, c/image will block each request until the overall pull is done. The file will be removed once the initial pull is done, so we should be good to go. I updated the comment to reflect that.

server/image_pull.go
QiWang19: I think the results of ParseDockerRef are different, 
- `CredentialProviderRequest.Image` is the image repo, doesn't have the tag or digest, `ParseDockerRef` adds the `latest` tag by default
- crio image.string() keeps the original tag or digest reference from the pod.spec

saschagrunert: I don’t understand that, do you mind providing me an example? I changed the implementation so that both (CRI-O and the credential provider) call `ParseDockerRef`, so I thought we’re already good.

QiWang19: Use `docker.io/nginx:1.29.1` instead of `latest` tag in the pod.spec.
The provider created `default-86cee2c10898a48ff47a0626f8202a2408b5893aff5df58ae10d83166e6d7b0e.json`
crio checks `default-64f77a9720499f32babcfd7b98ded57f1d4f1233d552193bc3bd9c78e32be796.json`

```
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: default
spec:
  containers:
    - name: test-container
      image: docker.io/nginx:1.29.1
      imagePullPolicy: Always
```

provider(https://github.com/cri-o/credential-provider/commit/60dff6bbfc9cd6a3cde78549a477cdc2c678ecca) log
```
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:26: Running credential provider
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:31: Reading from stdin
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:38: Got stdin, parsing JSON as CredentialProviderRequest
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:45: Parsed credential provider request for image "docker.io/library/nginx"
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:53: Normalized provided image name from "docker.io/library/nginx" to "docker.io/library/nginx:latest"
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:56: Parsing namespace from request
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:67: Getting secrets from namespace: default
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:74: Got 1 secret(s)
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:76: Matching mirrors for registry config: /etc/containers/registries.conf
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:83: Got mirror(s) for "docker.io/library/nginx:latest": "quay.io/libpod/busybox, localhost:5000/v2/nginx"
Sep 18 21:53:27 fedora credential-prov[653621]: auth.go:74: Parsing secret: my-secret
Sep 18 21:53:27 fedora credential-prov[653621]: auth.go:84: Found docker config JSON auth in secret "my-secret" for "http://localhost:5000"
Sep 18 21:53:27 fedora credential-prov[653621]: auth.go:95: Checking if mirror "quay.io/libpod/busybox" matches registry "localhost:5000"
Sep 18 21:53:27 fedora credential-prov[653621]: auth.go:95: Checking if mirror "localhost:5000/v2/nginx" matches registry "localhost:5000"
Sep 18 21:53:27 fedora credential-prov[653621]: auth.go:98: Using mirror auth "localhost:5000/v2/nginx" for registry from secret "localhost:5000"
Sep 18 21:53:27 fedora credential-prov[653621]: auth.go:43: Wrote auth file to /etc/crio/auth/default-86cee2c10898a48ff47a0626f8202a2408b5893aff5df58ae10d83166e6d7b0e.json with 5 number of entries
Sep 18 21:53:27 fedora credential-prov[653621]: app.go:90: Auth file path: /etc/crio/auth/default-86cee2c10898a48ff47a0626f8202a2408b5893aff5df58ae10d83166e6d7b0e.json

```

crio log(I added a line to print out the authfile path: https://github.com/cri-o/cri-o/compare/0f6de02bf5b10730f032acdd5ddfeeecd7d049fa...QiWang19:crio-provider-auth-2?expand=1): 
```
Sep 18 21:53:27 fedora crio[650542]: time="2025-09-18T21:53:27.488051468-04:00" level=info msg="Pulling image: docker.io/nginx:1.29.1" file="server/image_pull.go:43" id=ab92d721-b984-43ac-8719-1a41553ae2c4 name=/runtime.v1.ImageService/PullImage
Sep 18 21:53:27 fedora crio[650542]: time="2025-09-18T21:53:27.488285973-04:00" level=info msg="Using auth file for namespace default: /etc/crio/auth/default-64f77a9720499f32babcfd7b98ded57f1d4f1233d552193bc3bd9c78e32be796.json, docker.io/library/nginx:1.29.1" file="server/image_pull.go:243" id=ab92d721-b984-43ac-8719-1a41553ae2c4 name=/runtime.v1.ImageService/PullImage
```

saschagrunert: Alright, so the problem is that the credential provider does not have the information about the tag at all. It's being called in the kubelet via the `repoToPull`:

https://github.com/kubernetes/kubernetes/blob/6070f5a92918020b8c792ca1219a5fbbfa53a556/pkg/kubelet/images/image_manager.go#L218

Which is retrieved via:

```go
repoToPull, _, _, err := parsers.ParseImageName(spec.Image)
```

```go
func ParseImageName(image string) (string, string, string, error) {
	named, err := dockerref.ParseNormalizedNamed(image)
	if err != nil {
		return "", "", "", fmt.Errorf("couldn't parse image name %q: %v", image, err)
	}

	repoToPull := named.Name()

	// …

	return repoToPull, tag, digest, nil
}
```

This method basically calls `ParseNormalizedNamed()` which we also have in c/image. I changed the code path and adjusted the credential provider to remove the `ParseDockerRef`.

Ref: https://github.com/cri-o/credential-provider/pull/19

The problem is now that parallel calls to the same repo (like `docker.io/library/nginx`) for multiple tags in parallel can cause a race condition. I removed the whole cleanup logic for now.

saschagrunert:
@cri-o/cri-o-maintainers I'll follow-up on the cleanup process, we'll have to find a different solution for that.

saschagrunert:
@cri-o/cri-o-maintainers PTAL again

Issue #9358, opened 2025-07-22
CRI-O does not pass device fileMode to OCI spec, resulting in default 0666 permissions


Barakmor1:
### What happened?

Hi,

I’m using Kubernetes with CRI-O as the container runtime interface and passing USB and other devices to containers using the devices field.

On the host, the device files have correct permissions (e.g., 0660) with specific owners and groups. However, inside the container (using crun as the OCI runtime), the uid and gid are preserved, but the device permissions are always 0666.

I opened an[ issue in crun](https://github.com/containers/crun/issues/1826), and the response was that the OCI config.json passed to the runtime does not include a fileMode field for the devices, so crun falls back to its default (0666). Here is an example of the generated config:

```
"devices": [
  {
    "path": "/dev/bus/usb/002/002",
    "type": "c",
    "major": 189,
    "minor": 129,
    "uid": 107,
    "gid": 107
  }
]
```

As per the [OCI runtime spec](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#device), the fileMode field is optional, but when not provided, the runtime will use its own default (often 0666). This behavior causes all containers to have full read/write access to the device regardless of host file permissions.

### What did you expect to happen?

From what I understand, the CRI implementation (in this case CRI-O) should be responsible for populating the fileMode in the OCI spec if it wants to preserve the host permissions.

Could CRI-O be updated to propagate the device file permissions (mode) into the OCI config? Otherwise, this could lead to unexpected security and access issues inside containers.

### How can we reproduce it (as minimally and precisely as possible)?

On the host node, set a device file (e.g. /dev/bus/usb/002/002) with specific permissions:

```
sudo chown 107:107 /dev/bus/usb/002/002
sudo chmod 0660 /dev/bus/usb/002/002
```

Deploy a pod that requests the device using the devices field in the PodSpec (via Kubernetes Device Plugin or static configuration if supported).

Inside the running container, check the device file permissions:

```
ls -l /dev/bus/usb/002/002
```

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
[vagrant@node02 ~]$ crio --version
crio version 1.33.2
   GitCommit:      0da1b7c95fa3ec7cb83306a39724bf39e35033f4
   GitCommitDate:  2025-07-01T08:46:54Z
   GitTreeState:   dirty
   BuildDate:      1970-01-01T00:00:00Z
   GoVersion:      go1.24.2
   Compiler:       gc
[...]

PR #9466, opened 2025-09-16
Implement getDeviceFileMode function to retrieve file mode for devices and update device specifications to include FileMode.

prasad89:
#### What type of PR is this?

/kind bug

Populate the `fileMode` field for devices in the generated OCI spec so that container device permissions match host device permissions. This prevents containers from getting overly permissive (0666) access to devices when using runtimes like crun, improving security and respecting host configuration.

Resolves https://github.com/cri-o/cri-o/issues/9358

Minimal change: retrieves device file permissions via `os.Stat` and sets `fileMode` in `spec.LinuxDevice`. Falls back to current behavior if stat fails.

Device permissions inside containers now reflect the host device permissions (instead of defaulting to 0666) when using CRI-O with runtimes that support `fileMode` (e.g., crun).

```release-note
Device permissions inside containers now reflect the host device permissions (instead of defaulting to 0666) when using CRI-O with runtimes that support `fileMode` (e.g., crun).
```

### DIFF
--- a/internal/factory/container/device_linux.go
+++ b/internal/factory/container/device_linux.go
@@ -24,7 +24,9 @@ func getDeviceFileMode(devicePath string) *os.FileMode {
 	if err != nil {
 		return nil
 	}
+
 	mode := stat.Mode()
+
 	return &mode
 }
 


internal/factory/container/device_linux.go
```diff
@@ -64,12 +75,13 @@ func (c *container) specAddHostDevicesIfPrivileged(privilegedWithoutHostDevices
...
-			Type:  string(hostDevice.Type),
-			Major: hostDevice.Major,
-			Minor: hostDevice.Minor,
-			UID:   &hostDevice.Uid,
-			GID:   &hostDevice.Gid,
+			Path:     hostDevice.Path,
+			Type:     string(hostDevice.Type),
+			Major:    hostDevice.Major,
+			Minor:    hostDevice.Minor,
+			FileMode: getDeviceFileMode(hostDevice.Path),
```
bitoku, line 82: I think you can use [hostDevice.FileMode](https://pkg.go.dev/github.com/opencontainers/cgroups/devices/config#Device) instead

PR #9470, opened 2025-09-19
fix: create container log directories to prevent conmon failures

T0MASD:
/kind feature

CRI-O constructs log paths but doesn't create container subdirectories,
causing conmon to fail with "No such file or directory" errors.

Add os.MkdirAll() in LogPath() to ensure directories exist before
container creation. Uses single syscall, succeeds silently if directories
already exist.

Fixes container creation failures in Kubernetes environments.

CRI-O fails to create container-specific subdirectories within the configured `log_dir`, causing `conmon` to fail when attempting to create log files. This results in container creation failures with the error:

```
[conmon:e]: Failed to open log file No such file or directory
```

this may be handled by crio user, which was not the case when trying to get `kind` to work in crio

Eliminates errors

```release-note
fix: create container log directories to prevent conmon failures
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


haircommander:
weird the kubelet should be creating this path, what version of k8s are you using?
/ok-to-test

T0MASD:
I have discovered this behaviour while developing https://github.com/T0MASD/kinc using AI. somehow I got to the state where the kubelet was not creating the mentioned dir, then I asked AI to find the issue and here we are... I agree this should be handled by user of crio, but exiting crio because when it's not with cryptic error is sub-optimal

saschagrunert:
Shouldn't conmon create the directory if it does not exists? :thinking:

PR #9475, opened 2025-09-23
Remove temporary auth files if used

saschagrunert:
#### What type of PR is this?

/kind feature

We now move the namespaced auth files to a unique location for singular usage. This means that CRI-O can also remove the file afterwards, while other parallel pulls can re-use the previous file location for parallel pulls.

Follow-up on https://github.com/cri-o/cri-o/pull/9463

cc @QiWang19 

```release-note
None
```

### DIFF
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -16,6 +16,7 @@ import (
 	imageTypes "github.com/containers/image/v5/types"
 	encconfig "github.com/containers/ocicrypt/config"
 	"github.com/docker/distribution/registry/api/errcode"
+	"github.com/google/uuid"
 	"github.com/opencontainers/go-digest"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 	crierrors "k8s.io/cri-api/pkg/errors"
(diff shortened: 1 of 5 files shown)


saschagrunert:
@cri-o/cri-o-maintainers PTAL

server/image_pull.go
```diff
@@ -218,28 +225,58 @@ func (s *Server) contextForNamespace(ctx context.Context, imageRef, namespace st
...
+
+	log.Infof(ctx, "Using auth file for namespace %s: %s", namespace, authFilePath)
+
+	inUseAuthDirPath := filepath.Join(s.config.NamespacedAuthDir, "in-use")
+	if err := os.MkdirAll(inUseAuthDirPath, 0o700); err != nil {
+		return cleanup, fmt.Errorf("unable to ensure in-use auth dir: %w", err)
+	}
+
+	tempAuthFilePath := filepath.Join(inUseAuthDirPath, fmt.Sprintf("%s-%x-%s.json", namespace, hash, uuid.New()))
+	if err := os.Rename(authFilePath, tempAuthFilePath); err != nil {
```
QiWang19, line 266: would os.Rename remove the `authFilePath`? the parallel pull will not be able to reuse this file

saschagrunert: The credential provider should be invoked again and therefore we should get a new auth file. I know we cannot really make it 100% race free but that should be close for realistic scenarios.

PR #9499, opened 2025-10-05
Make ContainerCreateTimeout configurable at runtime handler level

### LINKED ISSUE #9151: Make ContainerCreateTimeout configurable
In some use cases, such as with [Confidential Containers or Peer-Pods](https://github.com/confidential-containers), a longer [ContainerCreateTimeout](https://github.com/cri-o/cri-o/blob/5991aec3baf922445791cb6dbb9ed4522e251c16/internal/oci/oci.go#L35) is necessary. For instance, when deploying a Peer-Pod using a [cloud instance with a long startup time](https://learn.microsoft.com/en-us/azure/virtual-machines/ncads-h100-v5), such as DGX-like systems, the current hardcoded timeout of 240 seconds may be exceeded before the container’s hosting VM completes its initialization.

To support such scenarios, it would be helpful to either make ContainerCreateTimeout configurable via the CRI-O configuration file or consider increasing the default value.

snir911:
This allows different runtime handlers to have different container creation timeouts,
useful for VM-based runtimes that may need longer timeouts than OCI runtimes.

#### What type of PR is this?

/kind feature

Fixes #9151 and additional minor fixes

Fixes #9151

```release-note
Added `container_create_timeout` option to control timeout duration of container creation
```

### DIFF
--- a/test/ctr_timeout.bats
+++ b/test/ctr_timeout.bats
@@ -1,0 +1,140 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
+	setup_test
+}
+
+function teardown() {
+	cleanup_test
+}
+
+@test "container create timeout with default value" {
+	# Test that CRI-O uses default timeout (240s) when not specified
+	setup_crio
+	start_crio_no_setup
+	info_json=$(crictl info -o json)
+
+	# Verify the default timeout is set
+	default_runtime=$(jq -r '.config.crio.DefaultRuntime' <<< "$info_json")
+	timeout=$(jq -r ".config.crio.Runtimes.[\"$default_runtime\"].ContainerCreateTimeout" <<< "$info_json")
+	[[ "$timeout" == "240" ]]
+}
+
+@test "container create timeout with custom value" {
+	# Test that CRI-O starts with valid container_create_timeout configuration
+	setup_crio
+	cat > "$CRIO_CONFIG_DIR"/01-timeout.conf << EOF
+[crio.runtime.runtimes.testruntime]
+runtime_path = "$RUNTIME_BINARY_PATH"
+runtime_root = "$RUNTIME_ROOT"
+runtime_type = "$RUNTIME_TYPE"
+container_create_timeout = 300
+EOF
+
+	start_crio_no_setup
+	info_json=$(crictl info -o json)
+
+	# Verify the timeout is set in the runtime handler configuration
+	timeout=$(jq -r '.config.crio.Runtimes.testruntime.ContainerCreateTimeout' <<< "$info_json")
+	[[ "$timeout" == "300" ]]
+}
+
+@test "container create timeout with minimum value enforced" {
+	# Test that CRI-O enforces minimum timeout value (30s)
+	setup_crio
+	cat > "$CRIO_CONFIG_DIR"/01-timeout.conf << EOF
+[crio.runtime.runtimes.testruntime]
+runtime_path = "$RUNTIME_BINARY_PATH"
+runtime_root = "$RUNTIME_ROOT"
+runtime_type = "$RUNTIME_TYPE"
+container_create_timeout = 15
+EOF
+
+	start_crio_no_setup
+	info_json=$(crictl info -o json)
+
+	# Verify the timeout is set to minimum value
+	timeout=$(jq -r '.config.crio.Runtimes.testruntime.ContainerCreateTimeout' <<< "$info_json")
+	[[ "$timeout" == "30" ]]
+}
+
+@test "container create timeout with different runtime handlers" {
+	# Test that different runtime handlers can have different timeouts
+	setup_crio
+	cat > "$CRIO_CONFIG_DIR"/01-timeout.conf << EOF
+[crio.runtime]
+default_runtime = "testruntime"
+
+[crio.runtime.runtimes.testruntime]
+runtime_path = "$RUNTIME_BINARY_PATH"
+runtime_root = "$RUNTIME_ROOT"
+runtime_type = "$RUNTIME_TYPE"
+container_create_timeout = 300
+
+[crio.runtime.runtimes.testruntime2]
+runtime_path = "$RUNTIME_BINARY_PATH"
+runtime_root = "$RUNTIME_ROOT"
+runtime_type = "$RUNTIME_TYPE"
+container_create_timeout = 600
+EOF
+
+	unset CONTAINER_DEFAULT_RUNTIME
+	unset CONTAINER_RUNTIMES
+
+	start_crio_no_setup
+	info_json=$(crictl info -o json)
+
+	# Verify different timeouts for different runtimes
+	default_timeout=$(jq -r '.config.crio.Runtimes.testruntime.ContainerCreateTimeout' <<< "$info_json")
+	second_timeout=$(jq -r '.config.crio.Runtimes.testruntime2.ContainerCreateTimeout' <<< "$info_json")
+
+	[[ "$default_timeout" == "300" ]]
+	[[ "$second_timeout" == "600" ]]
+}
+
+@test "container create timeout actual enforcement with container" {
+	if [[ "${RUNTIME_TYPE}" == "pod" ]]; then
+		skip "not yet supported by conmonrs"
+	elif [[ "${RUNTIME_TYPE}" == "vm" ]]; then
+		# With Containerd's runtime shim v2, CreateContainer happens over ttrpc, hence, the call can't be
+		# intercepted by a simple wrapper like in the "oci" RUNTIME_TYPE case
+		skip "test doesn't support shim v2 API yet"
+	else
+		export RUNTIME_WRAPPER_PATH="${TESTDIR}/runtime-wrapper.sh"
+	fi
+
+	setup_crio
+
+	cat > "$RUNTIME_WRAPPER_PATH" << EOF
+#!/bin/bash
+
+[[ " \$* " =~ " create " ]] && sleep 40
+exec $RUNTIME_BINARY_PATH "\$@"
+EOF
+
+	chmod +x "$RUNTIME_WRAPPER_PATH"
+
+	cat > "$CRIO_CONFIG_DIR"/01-timeout.conf << EOF
+[crio.runtime]
+default_runtime = "testruntime"
+
+[crio.runtime.runtimes.testruntime]
+runtime_path = "$RUNTIME_WRAPPER_PATH"
+runtime_root = "$RUNTIME_ROOT"
+runtime_type = "$RUNTIME_TYPE"
+container_create_timeout = 30
+EOF
+
+	start_crio_no_setup
+
+	# Create pod first - infra container creation may also timeout
+	pod_id=$(crictl runp --runtime testruntime "$TESTDATA"/sandbox_config.json)
+
+	# Attempt to create container - should fail due to timeout
+	run ! crictl create -T 60s "$pod_id" "$TESTDATA"/container_sleep.json "$TESTDATA"/sandbox_config.json
+
+	# Verify the container creation timeout message appears in CRI-O logs
+	grep -q "Container creation timeout (30s)" "$CRIO_LOG"
+}


fidencio:
@snir911, what happens when the ContainerCreateTimeout is set to, let's say, 600 (10 minutes), but kubelet's runtimeRequestTimeout is lower than that?

What I think that would happen is that kubelet's runtimeRequestTimeout would be the value respected in this case, which makes me think it's at least worth it adding to the documentation that the value set here depends also on the kubelet's value.

snir911:
That's right, I'll mention it, thanks @fidencio !

docs/crio.conf.5.md
```diff
@@ -392,6 +392,9 @@ conmon-rs (`runtime_type = "pod"`) supports this configuration for exec and atta
 Path to the seccomp.json profile which is used as the default seccomp profile for the runtime. If not specified, then the `crio.runtime` seccomp profile will be used.
 If that is also not specified, then the internal default seccomp profile will be used.
 
+**container_create_timeout**=240
+The timeout for container creation operations in seconds. If not set, defaults to 240 seconds. If set to a value less than 30 seconds, it will be automatically adjusted to 30 seconds (the minimum allowed value). This allows different runtime handlers to have different container creation timeouts, which is useful for VM-based runtimes that may need longer timeouts than OCI runtimes.
```
bitoku, line 396: I want it to mention about runtimeRequestTimeout in kubelet, and its document link if any.

internal/oci/runtime_vm.go
```diff
@@ -175,7 +172,7 @@ func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupPar
 		return err
 	}
 
-	containerIO, err := r.createContainerIO(ctx, c, cio.WithNewFIFOs(r.fifoDir, c.terminal, c.stdin))
+	containerIO, err := r.createContainerIO(ctx, c, cio.WithNewFIFOs(filepath.Join(r.handler.RuntimeRoot, "crio", "fifo"), c.terminal, c.stdin))
```
bitoku: Let's make a function to create this fifo path.

bitoku:
/release-note-edit
```release-note
Added `container_create_timeout` option to control timeout duration of container creation
```

internal/oci/runtime_vm.go
```diff
@@ -214,9 +216,13 @@ func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupPar
...
 	createdCh := make(chan error)
 
+	// Create a context with timeout for the task creation
+	taskCtx, taskCancel := context.WithTimeout(ctx, timeout)
+	defer taskCancel()
+
 	go func() {
 		// Create the container
-		if resp, err := r.task.Create(r.ctx, request); err != nil {
+		if resp, err := r.task.Create(taskCtx, request); err != nil {
```
bitoku, line 235: This changes the previous behavior. Is it intentional?

Apparently the context was introduced in this PR.
https://github.com/cri-o/cri-o/pull/4803/files

If we do this, we can use `taskCtx.Done()` instead of `time.After(timeout)` later, but I'd want to keep the previous behavior.

docs/crio.conf.5.md
```diff
@@ -392,6 +392,11 @@ conmon-rs (`runtime_type = "pod"`) supports this configuration for exec and atta
...
+
+Note: This timeout is subject to kubelet's `--runtime-request-timeout` flag (default: 2 minutes). If the container creation exceeds the kubelet timeout, the request will be canceled by kubelet before this timeout is reached. For more information about kubelet's runtime request timeout, see the [Kubelet documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
+
```
bitoku: I like this document. thanks!

internal/oci/runtime_vm.go
snir911: yes, this is intentional, if context has no timeout the goroutine may stay running after the createContainer timeout has been reached

bitoku: I prefer PRs focus on one topic, in this case, adding configuration.

However the change here looks totally legit, so I think we can keep this change for now.
In case https://github.com/cri-o/cri-o/issues/4798 this kind of regression was found, we should revert this commit https://github.com/cri-o/cri-o/pull/9499/commits/f48d19b736cbd0439fb30d7af1180830e94dac57.

bitoku (commented):
sorry one more thing. we want integration tests for this config.
See `test/*.bats` for other integration tests.

test/ctr_timeout.bats
```diff
@@ -0,0 +1,85 @@
...
+
+	# Verify the default timeout is set
+	default_runtime=$(jq -r '.config.crio.DefaultRuntime' <<< "$info_json")
+	timeout=$(jq -r ".config.crio.Runtimes.[\"$default_runtime\"].ContainerCreateTimeout" <<< "$info_json")
+	[[ "$timeout" == "240" ]]
+}
+
+@test "container create timeout with custom value" {
+	# Test that CRI-O starts with valid container_create_timeout configuration
+	cat > "$CRIO_CONFIG_DIR"/01-timeout.conf << EOF
```
littlejawa, line 28: I think you need to do:
```
setup_crio
cat > [timeout.conf]...
start_crio_no_setup
```

test/ctr_timeout.bats
bitoku, line 1: I think we want at least one test case that ensures it times out in the seconds we set. only shortest (30s) test is enough.
You can use `--cancel-timeout` or `-T` option to set a client (crictl) timeout.

snir911: I'm not sure how to do that, i need to mimic a long creation time, no? otherwise container will just run and cancelled after specified time (with -T)

bitoku: We could do by using the binary which sleeps for a while as a container runtime binary, but if it's hard to implement, we can add the test later as a follow up, I guess.

bitoku: I mean it's not a blocker.

snir911: @bitoku thanks for you help, got something working..

snir911: @bitoku it didn't work eventually, for some reason it caused only the conmon-rs+runc suit to fail, I'll try to figure this out later

bitoku: It may be because runtime_pod doesn't have a mechanism to set timeout when creating a container. @saschagrunert may know better.

snir911:
@bitoku @haircommander @fidencio @littlejawa I’ve incorporated the requested changes, PTAL when you have a moment. Thanks 🙏

test/ctr_timeout.bats
bitoku: We shouldn't remove the test at least. Whether we can ignore runtime_pod is open to discussion though.

pkg/config/config.go
```diff
@@ -1961,6 +1974,22 @@ func (r *RuntimeHandler) ValidateContainerMinMemory(name string) error {
...
+		logrus.Infof("Runtime handler %q container create timeout not set, using default: %d seconds", name, r.ContainerCreateTimeout)
+	case r.ContainerCreateTimeout < minimumContainerCreateTimeout:
+		logrus.Warnf("Runtime handler %q container create timeout (%d seconds) is less than minimum (%d seconds), setting to minimum: %d seconds", name, r.ContainerCreateTimeout, minimumContainerCreateTimeout, minimumContainerCreateTimeout)
+		r.ContainerCreateTimeout = minimumContainerCreateTimeout
+	default:
+		logrus.Infof("Runtime handler %q container create timeout set to: %d seconds", name, r.ContainerCreateTimeout)
+	}
+
+	return nil
+}
```
saschagrunert, line 1991: This method always returns `nil`.

snir911: because it sets to default value if invalid, would be better to return error and avoid setting defaults?

saschagrunert: I guess returning nothing should be fine in that case, right?

internal/oci/runtime_vm.go
```diff
@@ -230,14 +236,14 @@ func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupPar
 		if err != nil {
 			return fmt.Errorf("CreateContainer failed: %w", err)
 		}
-	case <-time.After(ContainerCreateTimeout):
+	case <-time.After(timeout):
 		if err := r.remove(c.ID(), ""); err != nil {
-			return err
+			return fmt.Errorf("failed to cleanup container after creation timeout has reached: %w", err)
```
saschagrunert:
```suggestion
			return fmt.Errorf("failed to cleanup container after creation timeout (%v): %w", timeout, err)
```

internal/oci/runtime_vm.go
```diff
@@ -214,9 +216,13 @@ func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupPar
 
 	createdCh := make(chan error)
 
+	// Create a context with timeout for the task creation
+	taskCtx, taskCancel := context.WithTimeout(ctx, timeout)
+	defer taskCancel()
```
saschagrunert, line 231: You're using both `context.WithTimeout` and `time.After` with the same timeout value. This is redundant and could cause subtle race conditions where one fires slightly before the other.

I's say we use the context timeout OR the select timeout, not both:

```
taskCtx, taskCancel := context.WithTimeout(ctx, timeout)
defer taskCancel()

// ... goroutine ...

select {
case err = <-createdCh:
    // ...
case <-taskCtx.Done():
    if err := r.remove(c.ID(), ""); err != nil {
        return fmt.Errorf("failed to cleanup container after creation timeout: %w", err)
    }
    <-createdCh  // wait for goroutine
    return fmt.Errorf("CreateContainer timeout (%v)", timeout)
}
```

docs/crio.conf.5.md
```diff
@@ -392,6 +392,11 @@ conmon-rs (`runtime_type = "pod"`) supports this configuration for exec and atta
 Path to the seccomp.json profile which is used as the default seccomp profile for the runtime. If not specified, then the `crio.runtime` seccomp profile will be used.
 If that is also not specified, then the internal default seccomp profile will be used.
 
+**container_create_timeout**=240
+The timeout for container creation operations in seconds. If not set, defaults to 240 seconds. If set to a value less than 30 seconds, it will be automatically adjusted to 30 seconds (the minimum allowed value). This allows different runtime handlers to have different container creation timeouts, which is useful for VM-based runtimes that may need longer timeouts than OCI runtimes.
+
+Note: This timeout is subject to kubelet's `--runtime-request-timeout` flag (default: 2 minutes). If the container creation exceeds the kubelet timeout, the request will be canceled by kubelet before this timeout is reached. For more information about kubelet's runtime request timeout, see the [Kubelet documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
```
saschagrunert: I'd extend that a bit, like:

```suggestion
Note: The effective timeout is the **minimum** of this value and kubelet's `--runtime-request-timeout` (default: 2 minutes). If you set `container_create_timeout = 600` (10 minutes) but kubelet has the default 2-minute timeout, the operation will be canceled after 2 minutes. Configure both values consistently for VM-based runtimes. For more information about kubelet's runtime request timeout, see the [Kubelet documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
```

snir911:
@saschagrunert @bitoku any hints on why the conmon-rs + runc combination is failing the [enforcement test](https://github.com/cri-o/cri-o/pull/9499/commits/95f9e7c83e174d02cc406d5d94e17f47466d6ada#diff-0ae49255f3171f8a6fe33bf3fb4c621ed2245371bc0f36f0a7d3e97c7cd08d94R97) ?

saschagrunert:
> @saschagrunert @bitoku any hints on why the conmon-rs + runc combination is failing the [enforcement test](https://github.com/cri-o/cri-o/pull/9499/commits/95f9e7c83e174d02cc406d5d94e17f47466d6ada#diff-0ae49255f3171f8a6fe33bf3fb4c621ed2245371bc0f36f0a7d3e97c7cd08d94R97) ?

Could be a flake, let's check.

internal/oci/runtime_vm.go
saschagrunert: I don't think that's resolved, we probably need something like this:

```diff
diff --git a/internal/oci/runtime_vm.go b/internal/oci/runtime_vm.go
index ebfd4e987c5..9d4e5c3a2e1 100644
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -233,15 +233,24 @@ func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupPar
 
 	select {
 	case err = <-createdCh:
 		if err != nil {
 			return fmt.Errorf("CreateContainer failed: %w", err)
 		}
 	case <-taskCtx.Done():
+		// Timeout occurred - attempt cleanup
 		if err := r.remove(c.ID(), ""); err != nil {
-			return fmt.Errorf("failed to cleanup container after creation timeout (%v): %w", timeout, err)
+			log.Warnf(ctx, "Failed to cleanup container after creation timeout (%v): %v", timeout, err)
 		}
 
-		<-createdCh
+		// Drain the channel asynchronously to avoid goroutine leak
+		// If the task.Create is still blocked, we don't want to wait for it
+		select {
+		case err = <-createdCh:
+			if err != nil {
+				log.Debugf(ctx, "Container creation eventually failed: %v", err)
+			}
+		default:
+			log.Warnf(ctx, "Container creation goroutine may still be running after timeout")
+		}
 
 		return fmt.Errorf("CreateContainer timeout (%v)", timeout)
 	}
```

bitoku:
I don't think it's flake. There's no code to respect the timeout value in `runtime_pod.go`.
We need something similar timeout mechanism to runtime_oci, or just skip it for runtime_pod.go for now and follow up later.

internal/oci/runtime_vm.go
snir911: @saschagrunert, i did it a bit different, does it make sense now?

snir911:
@bitoku, see the last commit i tried to add the implementation to runtime_pod.go, however, it doesn't seem to help. should i remove it (& the test itelf) so we can get progress for now?

bitoku:
We shouldn't remove the test. we can skip the test based on `$RUNTIME_TYPE` instead.

docs/crio.conf.5.md
```diff
@@ -392,6 +392,11 @@ conmon-rs (`runtime_type = "pod"`) supports this configuration for exec and atta
 Path to the seccomp.json profile which is used as the default seccomp profile for the runtime. If not specified, then the `crio.runtime` seccomp profile will be used.
 If that is also not specified, then the internal default seccomp profile will be used.
 
+**container_create_timeout**=240
+The timeout for container creation operations in seconds. If not set, defaults to 240 seconds. If set to a value less than 30 seconds, it will be automatically adjusted to 30 seconds (the minimum allowed value). This allows different runtime handlers to have different container creation timeouts, which is useful for VM-based runtimes that may need longer timeouts than OCI runtimes.
+
+Note: The effective timeout is the **minimum** of this value and kubelet's `--runtime-request-timeout` (default: 2 minutes). If you set `container_create_timeout = 600` (10 minutes) but kubelet has the default 2-minute timeout, the operation will be canceled after 2 minutes. Configure both values consistently for VM-based runtimes. For more information about kubelet's runtime request timeout, see the [Kubelet documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
+
```
bitoku, line 400: I want to mention that pod runtimetype is not supported.

snir911:
@saschagrunert @bitoku recent changes includes:
1. skipping the execution test for runtime_vm as i cannot intercept the create calls for the shim
2. modify log text to be persistent
3. rebase
(current error seems unrelated )

Another look will be appreciated, thanks

bitoku:
@snir911 
> skipping the execution test for runtime_vm as i cannot intercept the create calls for the shim

I thought this is needed especially for runtime_vm. It doesn't make sense to disable the test for vm.
Is it just a test infra issue, or code?

snir911:
> It doesn't make sense to disable the test for vm. Is it just a test infra issue, or code?

@bitoku You’re correct. However, at the moment I have no idea how to simulate the hanging CreateContainer call in this test infra. When containerd-shim-v2 is used, the CreateContainer call is performed over ttrpc, so I can’t intercept it with a simple wrapper like we do in the runtime_oci case.

If it is possible, it will probably require a more complex test scenario, so it might make sense to implement it separately.

bitoku:
@snir911 
Alright, can you add more explanation why we can't test it in the test code?
Then I'm good to merge this one.

snir911:
> Alright, can you add more explanation why we can't test it in the test code?
Then I'm good to merge this one.

@bitoku done, BTW how the two pending checks can be triggered?

bitoku:
They seem running. let me wait for the tests to be green.

bitoku:
/lgtm

@snir911 Thank you for your contribution and patience!

PR #9502, opened 2025-10-07
Re-use public credential provider API

saschagrunert:
#### What type of PR is this?

/kind cleanup

Re-using the public auth and config API from the credential provider to keep both projects in sync.

None
None

```release-note
None
```


server/image_pull.go
```diff
@@ -262,7 +267,10 @@ func (s *Server) prepareTempAuthFile(ctx context.Context, sysCtx *imageTypes.Sys
 		return cleanup, fmt.Errorf("unable to ensure in-use auth dir: %w", err)
 	}
 
-	tempAuthFilePath := filepath.Join(inUseAuthDirPath, fmt.Sprintf("%s-%x-%s.json", namespace, hash, uuid.New()))
+	ext := filepath.Ext(authFilePath)
+	newAuthFileName := fmt.Sprintf("%s-%s%s", strings.TrimSuffix(filepath.Base(authFilePath), ext), uuid.New(), ext)
+
+	tempAuthFilePath := filepath.Join(inUseAuthDirPath, newAuthFileName)
```
bitoku, line 291: https://github.com/cri-o/cri-o/pull/9463/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5

This description may be outdated? The final auth file name will be `<NAMESPACE>-<IMAGE_NAME_SHA256>-<UUID>.json`.

server/image_pull.go
```diff
@@ -262,7 +267,10 @@ func (s *Server) prepareTempAuthFile(ctx context.Context, sysCtx *imageTypes.Sys
...
 	if err := os.Rename(authFilePath, tempAuthFilePath); err != nil {
 		return cleanup, fmt.Errorf("unable to move auth file path to temporary location: %w", err)
 	}
```
bitoku, line 296: https://github.com/cri-o/cri-o/pull/9475#discussion_r2372477902
We should mention potential race scenarios (same name with different tag images pulled at once) and the reason why it's acceptable for posterity.

server/image_pull.go
```diff
@@ -262,7 +267,10 @@ func (s *Server) prepareTempAuthFile(ctx context.Context, sysCtx *imageTypes.Sys
 		return cleanup, fmt.Errorf("unable to ensure in-use auth dir: %w", err)
```
bitoku, line 285: I'm wondering if we want to cleanup `authFilePath` when it fails in the middle. Does it make this more race-prone?

server/image_pull.go
saschagrunert: Updated the doc within the function to explain the possible race.

server/image_pull.go
saschagrunert: Good catch, updated the docs.

server/image_pull.go
saschagrunert: Makes sense, I added another cleanup function if that method errors.

bitoku:
/lgtm
/hold
I hold it in case @QiWang19 wants to look, otherwise unhold anytime.

PR #9509, opened 2025-10-12
Clean up duplicate SignaturePolicyPath logic in image_pull.go

### LINKED ISSUE #9217: Redundant policy logic, incorrect logging
https://github.com/cri-o/cri-o/blob/e738af12221029190523129c2705defeb889dfce/server/image_pull.go#L138-L159 (via the called `contextForNamespace`) contains two duplicate copies of the logic that sets `SignaturePolicyPath`, _and_ two duplicate debug logs of the chosen value (one of which is irrelevant).

Please clean that up.

gouthamhusky:
This PR addresses the duplicate logic and debug logs for SignaturePolicyPath in the pullImage function.
fixes #9217 

## Changes Made

- Removed duplicate SignaturePolicyPath setting: The pullImage function was setting the signature policy path twice - once via contextForNamespace() and again with duplicate logic directly in the function.

- Eliminated redundant debug logs: There were two debug logs for the signature policy path, one of which was irrelevant after the first one. Now there is only one relevant debug log.

```release-note
Cleaned up duplicate signature policy path logic in server image pull
```

### DIFF
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -143,6 +143,7 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
 	if err != nil {
 		return storage.RegistryImageReference{}, fmt.Errorf("get context for namespace: %w", err)
 	}
+
 	log.Debugf(ctx, "Using pull policy path for image %s: %q", pullArgs.image, sourceCtx.SignaturePolicyPath)
 
 	if pullArgs.namespace != "" {


gouthamhusky:
@bitoku mind reviewing this one when you get the chance to?

server/image_pull.go
```diff
@@ -152,23 +152,12 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
...
-		policyPath := filepath.Join(s.config.SignaturePolicyDir, pullArgs.namespace+".json")
-		if _, err := os.Stat(policyPath); err == nil {
-			sourceCtx.SignaturePolicyPath = policyPath
-		} else if !os.IsNotExist(err) {
-			return storage.RegistryImageReference{}, fmt.Errorf("read policy path %s: %w", policyPath, err)
-		}
-	}
-
-	log.Debugf(ctx, "Using pull policy path for image %s: %s", pullArgs.image, sourceCtx.SignaturePolicyPath)
+	log.Debugf(ctx, "Using pull policy path for image %s: %q", pullArgs.image, sourceCtx.SignaturePolicyPath)
```
mtrmac: I think this makes more sense to log immediately after `contextForNamespace`, where that value is decided.

bitoku:
@gouthamhusky Thank you for taking it! Can you add release-note?
/ok-to-test

server/image_pull.go
gouthamhusky: makes sense!

gouthamhusky: @mtrmac made the change

saschagrunert:
/lgtm cancel

Do you mind fixing the linter CI?
```
  	log.Debugf(ctx, "Using pull policy path for image %s: %q", pullArgs.image, sourceCtx.SignaturePolicyPath)
  	^
```

mtrmac:
@gouthamhusky Thanks for taking care of this!

PR #9511, opened 2025-10-13
Remove support for `InsecureRegistries` in favor of `registries.conf`

bitoku:
#### What type of PR is this?

/kind deprecation

This PR makes `insecureRegistries` no effective.

part of #9278 

I left the option for now.
Let me know if you think it's too early to make it ineffective, or if you think we can remove it completely right now.

```release-note
Deprecated --insecure-registries option, and made it ineffective.
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net"
 	"os"
 	"regexp"
 	"sort"
@@ -71,11 +70,6 @@ type ImageResult struct {
 	MountPoint          string
 }
 
-type indexInfo struct {
-	name   string
-	secure bool
-}
-
 // A set of information that we prefer to cache about images, so that we can
 // avoid having to reread them every time we need to return information about
 // images.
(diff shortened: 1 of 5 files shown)


internal/storage/image.go
```diff
@@ -1083,24 +1041,7 @@ func GetImageService(ctx context.Context, store storage.Store, storageTransport
...
-			// Valid CIDR.
-			is.lookup.InsecureRegistryCIDRs = append(is.lookup.InsecureRegistryCIDRs, ipnet)
-		} else {
-			// Assume `host:port` if not CIDR.
-			is.lookup.IndexConfigs[r] = &indexInfo{
-				name:   r,
-				secure: false,
-			}
-		}
+		log.Warnf(ctx, "Insecure registries option is deprecated and no longer effective. Please use `insecure` in `registries.conf` instead.")
```
saschagrunert: I think we should error here: 
```suggestion
		log.Errorf(ctx, "Insecure registries option is deprecated and no longer effective. Please use `insecure` in `registries.conf` instead.")
```

saschagrunert:
Ah, linter CI is not happy:
```
  Error: internal/storage/image.go:74:2: field name is unused (unused)
  	name   string
  	^
  Error: internal/storage/image.go:75:2: field secure is unused (unused)
  	secure bool
  	^
```
/lgtm cancel

internal/storage/image.go
bitoku: Thanks! done

PR #9531, opened 2025-10-24
metrics: add container_last_seen and spec metrics

haircommander:
#### What type of PR is this?

/kind feature
contributes to https://github.com/cri-o/cri-o/issues/9125

```release-note
add container_spec* and container_last_seen metrics

**Action required**: `container_spec_memory_limit_bytes` has moved from the `memory`
metrics category to the new `spec` category. Update your CRI-O configuration to include
`spec` in `included_pod_metrics` if you rely on this metric.

```

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -59,7 +59,8 @@ type Container struct {
 	conmonCgroupfsPath string
 	crioAnnotations    fields.Set
 	state              *ContainerState
-	opLock             sync.RWMutex
+	opLock             sync.RWMutex // For handling the container OCI runtime state transitions (i.e: anything that needs to call OCI runtime)
+	metaLock           sync.RWMutex // For handling the internal oci.Container metadata. Currently only used for Spec and Resources fields
 	spec               *specs.Spec
 	idMappings         *idtools.IDMappings
 	terminal           bool
@@ -251,6 +252,9 @@ func (c *Container) CRIContainer() *types.Container {
 
 // SetSpec loads the OCI spec in the container struct.
 func (c *Container) SetSpec(s *specs.Spec) {
+	c.metaLock.Lock()
+	defer c.metaLock.Unlock()
+
 	c.spec = s
 	c.SetResources(s)
 	c.SetRuntimeUser(s)
@@ -258,6 +262,9 @@ func (c *Container) SetSpec(s *specs.Spec) {
 
 // Spec returns a copy of the spec for the container.
 func (c *Container) Spec() specs.Spec {
+	c.metaLock.RLock()
+	defer c.metaLock.RUnlock()
+
 	if c.spec != nil {
 		return *c.spec
 	}
@@ -835,6 +842,9 @@ func (c *Container) SetResources(s *specs.Spec) {
 
 // GetResources returns a copy of the Linux resources from Container.
 func (c *Container) GetResources() *types.ContainerResources {
+	c.metaLock.RLock()
+	defer c.metaLock.RUnlock()
+
 	return c.resources
 }
 


internal/lib/stats/spec_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+
+	"github.com/cri-o/cri-o/internal/lib/sandbox"
+	"github.com/cri-o/cri-o/internal/oci"
+)
+
+func generateSandboxSpecMetrics(sb *sandbox.Sandbox, c *oci.Container) []*types.Metric {
+	// TODO FIXME read lock
+	resources := c.GetResources()
```
saschagrunert: Would this already work?

```go
c.StateLock().RLock()
defer c.StateLock().RUnlock()
resources := c.GetResources()
```

internal/lib/stats/stats_server_linux.go
```diff
@@ -232,7 +233,17 @@ func (ss *StatsServer) GenerateSandboxContainerMetrics(sb *sandbox.Sandbox, c *o
...
+	metrics := computeSandboxMetrics(sb, []*containerMetric{
+		{
+			desc: containerLastSeen,
+			valueFunc: func() metricValues {
+				return metricValues{{
+					value:      uint64(time.Now().Unix()),
+					metricType: types.MetricType_GAUGE,
+				}}
+			},
+		},
```
saschagrunert, line 245: The `container_last_seen` metric is computed for all containers even when not explicitly requested in the configuration. This differs from other metrics.

Is this intentional? Should it check if "" is in `ss.Config().IncludedPodMetrics`?

internal/lib/stats/descriptors.go
```diff
@@ -77,11 +77,6 @@ var (
 		Help:      "Container swap usage in bytes.",
 		LabelKeys: baseLabelKeys,
 	}
-	containerSpecMemoryLimitBytes = &types.MetricDescriptor{
-		Name:      "container_spec_memory_limit_bytes",
-		Help:      "Memory limit for the container in bytes.",
-		LabelKeys: baseLabelKeys,
-	}
```
saschagrunert, line 84: We need to document that breaking change in the notes, like:

```
**Action required**: `container_spec_memory_limit_bytes` has moved from the `memory`
metrics category to the new `spec` category. Update your CRI-O configuration to include
`spec` in `included_pod_metrics` if you rely on this metric.
```

internal/lib/stats/spec_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+	if resources.GetLinux() == nil {
```
saschagrunert: We probably should check if `resources != nil`, too.

internal/lib/stats/spec_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+			desc: containerSpecCpuQuota,
+			valueFunc: func() metricValues {
+				return metricValues{{value: uint64(resources.GetLinux().GetCpuQuota()), metricType: types.MetricType_GAUGE}}
+			},
+		})
+	}
+
+	if m, ok := resources.GetLinux().GetUnified()["memory.min"]; ok {
+		u, err := strconv.ParseUint(m, 10, 64)
+		if err == nil {
```
saschagrunert: Should we log a warning in case of an error?

internal/lib/stats/spec_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+				}
+
+				return metricValues{{value: uint64(limit), metricType: types.MetricType_GAUGE}}
+			},
+		},
+		{
+			desc: containerSpecMemorySwapLimitBytes,
+			valueFunc: func() metricValues {
+				return metricValues{{value: uint64(resources.GetLinux().GetMemorySwapLimitInBytes()), metricType: types.MetricType_GAUGE}}
+			},
```
saschagrunert, line 31: MemorySwapLimitInBytes is reported directly without the same "unlimited" handling that memory limit has. Should swap also check for values > maxMemorySize?

internal/lib/stats/spec_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+		{
+			desc: containerSpecMemoryLimitBytes,
+			valueFunc: func() metricValues {
+				// For consistency with cAdvisor and Kubernetes, consider memory to be "unlimited"
+				// when above a certain threshold (2^62) and report it as 0 in the metrics.
+				// This approach is more useful for monitoring tools than reporting the physical limit.
+				limit := resources.GetLinux().GetMemoryLimitInBytes()
+				if limit > int64(maxMemorySize) {
+					return metricValues{{value: 0, metricType: types.MetricType_GAUGE}}
+				}
```
saschagrunert: Comparing int64 to uint64(maxMemorySize) could have sign issues. If limit is -1 (unlimited), this comparison may not work as expected. How about:

```go
if limit < 0 || uint64(limit) > maxMemorySize {
      return metricValues{{value: 0, metricType: types.MetricType_GAUGE}}
}
```

internal/lib/stats/stats_server_linux.go
haircommander: yeah it's intentional to copy cadvisor behavior https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L126-L140 I'm open to putting it in a bucket though

internal/oci/container.go
```diff
@@ -835,6 +842,9 @@ func (c *Container) SetResources(s *specs.Spec) {
 
 // GetResources returns a copy of the Linux resources from Container.
 func (c *Container) GetResources() *types.ContainerResources {
+	c.metaLock.Lock()
+	defer c.metaLock.Unlock()
```
bitoku: https://github.com/cri-o/cri-o/pull/9531#discussion_r2465153989
Can't this be just read lock?

haircommander: :facepalm: good catch, meant to do that

PR #9535, opened 2025-10-28
metrics: correct container metric metadata

haircommander:
#### What type of PR is this?

/kind bug
building on https://github.com/cri-o/cri-o/pull/9531

```release-note
Fix a bug where CRI metrics had the incorrect metadata. Now, instead of the metrics being populated with the sandbox metadata, they are populated with the container metadata.
```


haircommander:
@cri-o/cri-o-maintainers PTAL

internal/lib/stats/stats_server_linux.go
bitoku, line 236: don't we need to change here too?

haircommander: yup good catch

saschagrunert (approved):

PR #9563, opened 2025-11-03
OCPBUGS-62150: server: ignore /etc/passwd mount

adrianreber:
#### What type of PR is this?

/kind bug

Ignore /etc/passwd when checking for mounts. Always rely on the mount as defined from Kubernetes. This follows existing practice like for /etc/hosts or /etc/resolv.conf

OCPBUGS-62150

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed container checkpoint and restore operations to properly preserve system authentication files and user identity settings, ensuring user credentials and access permissions remain consistent when containers are restored from checkpoints.

### DIFF
--- a/server/container_restore.go
+++ b/server/container_restore.go
@@ -272,6 +272,8 @@ func (s *Server) CRImportCheckpoint(
 		"/dev/shm":           true,
 		"/etc/resolv.conf":   true,
 		"/etc/hostname":      true,
+		"/etc/passwd":        true,
+		"/etc/group":         true,
 		"/run/secrets":       true,
 		"/run/.containerenv": true,
 	}
--- a/test/checkpoint.bats
+++ b/test/checkpoint.bats
@@ -174,3 +174,44 @@ function teardown() {
 	[[ "$container_name" == "restored-sleep-container" ]]
 	[[ "$pod_name" == "restoresandbox2" ]]
 }
+
+@test "checkpoint and restore: /etc/passwd uses Kubernetes run_as_user on restore" {
+	CONTAINER_ENABLE_CRIU_SUPPORT=true start_crio
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	# Create container with run_as_user=1001
+	START_JSON=$(mktemp)
+	jq '.linux.security_context.run_as_user.value = 1001
+		| .command=["/bin/sh"]
+		| .args=["-c","sleep inf"]' \
+		"$TESTDATA"/container_sleep.json > "$START_JSON"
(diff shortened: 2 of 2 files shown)


assafWeaversoft:
Details of this bug :
On OpenShift, SCC (Security Context Constraints) applies uid_masking for all namespaces.
Because of this, the current handling of /etc/passwd during container restore causes unexpected issues.
CRIO Reference: commit 429ef7c
To align with SCC behavior and prevent conflicts, CRI-O should skip /etc/passwd when performing restore operations.
Impact:
Containers are failing to restore because the expected mount for /etc/passwd is missing.
This prevents containers from being properly recovered after a restart.

adrianreber:
Not sure what to do about the Jira version complaint comment, but all tests are looking happy.

bitoku:
Can we add a test case for this change?

adrianreber:
> Can we add a test case for this change?

Sure. Here it is. This test fails on CRI-O before the patch. From what I understand SCC is using exactly the `run_as_user` mechanism to inject a UID in the container.

server/container_restore.go
```diff
@@ -272,6 +272,7 @@ func (s *Server) CRImportCheckpoint(
 		"/dev/shm":           true,
 		"/etc/resolv.conf":   true,
 		"/etc/hostname":      true,
+		"/etc/passwd":        true,
```
bitoku, line 275: should `/etc/group` also be in this list?

adrianreber: Good point. Probably yes.

adrianreber:
> I'm not familiar with checkpoint/restore, sorry if this is a silly question.
> 
> What happens if `/etc/passwd` is not auto-generated, but the container config requested via CRI has a bind mount to `/etc/passwd`? Should it be still ignored?

I don't know. It is up to us. In a Kubernetes context it seems pretty unlikely. We already block a couple of file the same way as now with `/etc/passwd`. As nobody has done this before it is hard to tell. I took the initial list of mounts to block from my Podman implementation. At this point I would not worry about it. It seems unlikely from my point of view and we can still fix it if somebody complains. But I don't really know.

bitoku:
/lgtm
@cri-o/cri-o-maintainers PTAL

server/container_restore.go
adrianreber: Added.

bitoku:
/lgtm
@cri-o/ocicni-maintainers PTAL

haircommander:
/approve

thanks!

bitoku:
Hi @adrianreber !
Do we want to cherry-pick this to 1.31 branch?

adrianreber:
> Hi @adrianreber ! Do we want to cherry-pick this to 1.31 branch?

I can do that on Monday.

PR #9567, opened 2025-11-03
metrics: add container_start_time_seconds

haircommander:
#### What type of PR is this?

/kind feature

```release-note
add `container_start_time_seconds` metric, nested under the `spec` metrics family
```

### DIFF
--- a/internal/lib/stats/spec_metrics.go
+++ b/internal/lib/stats/spec_metrics.go
@@ -57,18 +57,21 @@ func generateContainerSpecMetrics(ctr *oci.Container) []*types.Metric {
 		})
 	}
 
-	if m, ok := resources.GetLinux().GetUnified()["memory.min"]; ok {
-		u, err := strconv.ParseInt(m, 10, 64)
-		if err != nil {
-			log.Errorf(context.Background(), "Failed to parse memory.min value %s: %v", m, err)
-		} else {
-			specMetrics = append(specMetrics, &containerMetric{
-				desc: containerSpecMemoryReservationLimitBytes,
-				valueFunc: func() metricValues {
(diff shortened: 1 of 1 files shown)


internal/lib/stats/spec_metrics.go
```diff
@@ -42,6 +42,12 @@ func generateSandboxSpecMetrics(sb *sandbox.Sandbox, c *oci.Container) []*types.
 				return metricValues{{value: uint64(resources.GetLinux().GetCpuPeriod()), metricType: types.MetricType_GAUGE}}
 			},
 		},
+		{
+			desc: containerStartTimeSeconds,
+			valueFunc: func() metricValues {
+				return metricValues{{value: uint64(c.CreatedAt().Unix()), metricType: types.MetricType_GAUGE}}
```
sohankunkerkar: I don't think your branch has the latest code.

haircommander: oops, updated!

internal/lib/stats/descriptors.go
```diff
@@ -234,4 +234,9 @@ var (
 		Help:      "Memory swap limit for the container.",
 		LabelKeys: baseLabelKeys,
 	}
+	containerStartTimeSeconds = &types.MetricDescriptor{
```
saschagrunert, line 262: Do we have to add this metric to:

https://github.com/cri-o/cri-o/blob/8e164e8b78ce4e44302259e946acce816d287ae0/internal/lib/stats/metrics.go#L108-L115

haircommander: great catch! updated.

PR #9568, opened 2025-11-04
WIP: Add OCI artifact support to copyimg binary

### LINKED ISSUE #9041: Add artifact support to `copyimg` binary
The [`copyimg` binary](https://github.com/cri-o/cri-o/blob/main/test/copyimg/copyimg.go) is used in the tests and does not support OCI artifacts from https://github.com/cri-o/cri-o/pull/8996. We should add artifact support, and add `quay.io/crio/seccomp:v2` to the `$IMAGES` variable in https://github.com/cri-o/cri-o/blob/086f182d7d883326159b165d9b5958ae2ff53e14/test/common.sh#L110-L114

Then use the `${IMAGES[3]}` value from the referencing tests.

Follow-up on https://github.com/cri-o/cri-o/pull/8996#discussion_r1975911045

saschagrunert:
#### What type of PR is this?

/kind feature

This PR adds OCI artifact support to the `copyimg` binary used in tests. The main changes are:

- **Migrated copyimg to use libimage**: Replaced the low-level `containers/image/v5/copy` package with `containers/common/libimage`, which properly handles OCI artifacts
- **Removed unused policyContext code**: The `libimage.Copier` handles signature policy internally via `SystemContext`, so the explicit policy and policyContext creation/teardown was removed
- **Added artifact to IMAGES array**: Added `quay.io/crio/seccomp:v2` to the `IMAGES` array in `test/common.sh` to ensure the artifact is cached before running tests
- **Updated test references**: Changed `test/seccomp_oci_artifacts.bats` to reference `${IMAGES[3]}` instead of hardcoding the artifact image
- **Fixed shellcheck warnings**: Updated jq commands to use `--arg` for proper variable handling instead of string interpolation

OCI artifacts differ from container images in that they may use `application/vnd.oci.empty.v1+json` as their config media type and rely on the `artifactType` field. The `libimage.Copier` handles these cases properly, whereas the lower-level `copy.Image` function does not.

Fixes #9041

None

```release-note
None
```

### DIFF
--- a/test/common.sh
+++ b/test/common.sh
@@ -111,6 +111,7 @@ IMAGES=(
     registry.k8s.io/pause:3.10.1
     quay.io/crio/fedora-crio-ci:latest
     quay.io/crio/hello-wasm:latest
+    quay.io/crio/seccomp:v2
 )
 
 function img2dir() {
--- a/test/copyimg/copyimg.go
+++ b/test/copyimg/copyimg.go
@@ -3,8 +3,7 @@ package main
 import (
 	"os"
 
-	"github.com/containers/image/v5/copy"
-	"github.com/containers/image/v5/signature"
+	"github.com/containers/common/libimage"
 	"github.com/containers/image/v5/storage"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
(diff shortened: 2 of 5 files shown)


saschagrunert:
I think we should wait for the libartifact transition for this, otherwise we risk too many workarounds in that binary.

PR #9614, opened 2025-11-21
feat: allow exec to containers during graceful termination

### LINKED ISSUE #7160: Container cannot be accessed via kubectl exec when pod is in Terminating state
### What happened?

Containers associated to pods cannot be accessed via `kubectl exec` command (or `crictl exec` in the node where the container is running) when they are in Terminating state. 

We expect to be able to access containers even if they are in Terminating state. We have verified Docker works under the same conditions

1. Deploy a pod that ignores SIGTERM signals if pod is deleted. Set the spec parameter `terminationGracePeriodSeconds` parameter to a really high value to avoid the kubelet sending a SIGKILL to the container associated with the pod

```console
apiVersion: v1
kind: Pod
metadata:
  name: nginx-example
spec:
  containers:
  - name: nginx-example
    image: nginx
    imagePullPolicy: IfNotPresent
    command: [ "/bin/bash", "-c", "--" ]
    args: [ "while true; do sleep 30; done;" ]
  terminationGracePeriodSeconds: 3600
```

2. Wait for the pod to be running and then try to delete the pod (in the example, you should execute the command `kubectl delete pod nginx-example`). The delete command is synchronous but the SIGTERM is sent immediately, meaning the user will need to manually cancel the action with Ctrl+C but after that the pod will be left in `Terminating` state
```console
# kubectl  get pods  nginx-example
NAME            READY   STATUS        RESTARTS   AGE
nginx-example   1/1     Terminating   0          14m
```

3. Try to access the container associated with the pod executing the command `kubectl exec -it nginx-example -- bash`. The command will hung and as in the `kubectl delete` command, with the difference that Ctrl + C won't work. The same behaviour is shown if `crictl exec` command is executed in the container associated with the pod in the host where the container is running

_No response_

<details>

```console
crio version 1.25.2
Version:        1.25.2
GitCommit:      unknown
GitCommitDate:  unknown
GitTreeState:   clean
BuildDate:      2023-03-06T07:45:59Z
GoVersion:      go1.19
Compiler:       gc
Platform:       linux/amd64
Linkmode:       dynamic
BuildTags:
  rpm_crashtraceback
  exclude_graphdriver_btrfs
  btrfs_noversion
  libdm_no_deferred_remove
  seccomp
  containers_image_openpgp
LDFlags:          -X github.com/cri-o/cri-o/internal/pkg/criocli.DefaultsPath= -X  github.com/cri-o/cri-o/internal/version.buildDate=2023-03-06T07:45:59Z -X  github.com/cri-o/cri-o/internal/version.gitCommit=1d7407e62446d25ca4fa77c9f6853143ec994d15 -X  github.com/cri-o/cri-o/internal/version.version=1.25.2 -X  github.com/cri-o/cri-o/internal/version.gitTreeState=clean  -B 0x4b9fcda34660fd3501556a3366899982947c7308 -extldflags '-Wl,-z,relro  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld ' -compressdwarf=false
SeccompEnabled:   true
AppArmorEnabled:  false
Dependencies:
```

```console
~ k8s(yul1) kubectl version --output=json
{
  "clientVersion": {
    "major": "1",
    "minor": "25",
    "gitVersion": "v1.25.2",
    "gitCommit": "5835544ca568b757a8ecae5c153f317e5736700e",
    "gitTreeState": "clean",
    "buildDate": "2022-09-21T14:33:49Z",
    "goVersion": "go1.19.1",
    "compiler": "gc",
    "platform": "darwin/amd64"
  },
  "kustomizeVersion": "v4.5.7",
  "serverVersion": {
    "major": "1",
    "minor": "25",
    "gitVersion": "v1.25.7",
    "gitCommit": "723bcdb232300aaf5e147ff19b4df7ec8a20278d",
    "gitTreeState": "clean",
    "buildDate": "2023-02-22T13:58:23Z",
    "goVersion": "go1.19.6",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

</details>

### OS version

<details>

```console
# cat /etc/os-release
NAME="Oracle Linux Server"
VERSION="8.8"
ID="ol"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="8.8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Oracle Linux Server 8.8"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:oracle:linux:8:8:server"
HOME_URL="https://linux.oracle.com/"
BUG_REPORT_URL="https://github.com/oracle/oracle-linux"

ORACLE_BUGZILLA_PRODUCT="Oracle Linux 8"
ORACLE_BUGZILLA_PRODUCT_VERSION=8.8
ORACLE_SUPPORT_PRODUCT="Oracle Linux"
ORACLE_SUPPORT_PRODUCT_VERSION=8.8

# uname -a
Linux yul1-r13-u17 4.18.0-477.13.1.el8_8.x86_64 #1 SMP Tue May 30 16:09:32 PDT 2023 x86_64 x86_64 x86_64 GNU/Linux
```

</details>

<details>
Physical
</details>

willianpaixao:
**What type of PR is this?**

/kind feature

**What this PR does / Why we need it**:

This PR enables `kubectl exec` and `crictl exec` to access containers during their graceful termination period. Previously, these commands would hang indefinitely when attempting to access containers in the "Terminating" state.

The root cause was that the container stop loop held an exclusive operation lock (`opLock`) for the entire graceful shutdown duration, which could be minutes with high `terminationGracePeriodSeconds` values. The exec code path required this same lock to check container status, creating a deadlock scenario.

The fix implements granular locking in the stop loop, only holding locks around individual runtime operations rather than for the entire stop duration. This allows exec operations to proceed during the graceful termination window while maintaining thread safety.

**Which issue(s) this PR fixes**:

Fixes #7160

**Special notes for your reviewer**:

Key changes:
1. **server/container_exec.go**: Replaced `UpdateContainerStatus() + State()` check with `c.Living()` to avoid opLock dependency (matches ExecSync pattern)
2. **internal/oci/runtime_oci.go**: Removed global opLock from `StopLoopForContainer()` and added granular locking around individual runtime calls (pause/resume, kill, status checks)

The change allows execs during graceful termination (between SIGTERM and kill loop) but correctly blocks them once the kill loop begins.

**Does this PR introduce a user-facing change?**:

```release-note
Fixed kubectl exec and crictl exec commands hanging when accessing containers in the Terminating state. These commands now work correctly throughout the container shutdown period.
```

**Testing**:

Manual testing logs showing successful exec during termination:

<details>
<summary>CRI-O logs</summary>

```bash
INFO[2025-11-21T20:28:51.173942561+01:00] Pulling image: quay.io/quay/busybox:latest    id=23bdf144-dd71-43bd-9613-3712cc680b65 name=/runtime.v1.ImageService/PullImage
INFO[2025-11-21T20:28:51.17440329+01:00] Trying to access "quay.io/quay/busybox:latest" 
INFO[2025-11-21T20:28:52.949670864+01:00] Pulled image: quay.io/quay/busybox@sha256:92f3298bf80a1ba949140d77987f5de081f010337880cd771f7e7fc928f8c74d  id=23bdf144-dd71-43bd-9613-3712cc680b65 name=/runtime.v1.ImageService/PullImage
INFO[2025-11-21T20:28:52.962772532+01:00] Running pod sandbox: default/busybox-pod/POD  id=7525cb69-928b-4dee-b9fa-8e1f3fa52941 name=/runtime.v1.RuntimeService/RunPodSandbox
INFO[2025-11-21T20:28:52.962806243+01:00] Allowed annotations are specified for workload [io.containers.trace-syscall io.kubernetes.cri-o.Devices] 
INFO[2025-11-21T20:28:52.999253033+01:00] Ran pod sandbox e211f799afdc43b3b7455fed448476f818c5c857cfe3983be5a272f7bba5308f with infra container: default/busybox-pod/POD  id=7525cb69-928b-4dee-b9fa-8e1f3fa52941 name=/runtime.v1.RuntimeService/RunPodSandbox
INFO[2025-11-21T20:28:53.012440737+01:00] Creating container: //                        id=ba521307-4396-4ac3-9236-0bcf82dce774 name=/runtime.v1.RuntimeService/CreateContainer
INFO[2025-11-21T20:28:53.012501751+01:00] Allowed annotations are specified for workload [io.containers.trace-syscall io.kubernetes.cri-o.Devices] 
INFO[2025-11-21T20:28:53.014505914+01:00] Allowed annotations are specified for workload [io.containers.trace-syscall io.kubernetes.cri-o.Devices] 
INFO[2025-11-21T20:28:53.014998016+01:00] Allowed annotations are specified for workload [io.containers.trace-syscall io.kubernetes.cri-o.Devices] 
INFO[2025-11-21T20:28:53.042561101+01:00] Created container b7d01f7e2df0fe168b2bb853b3cf9e56a222ea006f00066ae721b4982df03ee5: //  id=ba521307-4396-4ac3-9236-0bcf82dce774 name=/runtime.v1.RuntimeService/CreateContainer
INFO[2025-11-21T20:28:53.053949058+01:00] Starting container: b7d01f7e2df0fe168b2bb853b3cf9e56a222ea006f00066ae721b4982df03ee5  id=4926d6b5-603a-4eec-b91c-0a5045c1832c name=/runtime.v1.RuntimeService/StartContainer
INFO[2025-11-21T20:28:53.055139909+01:00] Started container                             PID=114359 containerID=b7d01f7e2df0fe168b2bb853b3cf9e56a222ea006f00066ae721b4982df03ee5 description=// id=4926d6b5-603a-4eec-b91c-0a5045c1832c name=/runtime.v1.RuntimeService/StartContainer sandboxID=e211f799afdc43b3b7455fed448476f818c5c857cfe3983be5a272f7bba5308f
INFO[2025-11-21T20:28:55.100217021+01:00] Stopping container: b7d01f7e2df0fe168b2bb853b3cf9e56a222ea006f00066ae721b4982df03ee5 (timeout: 3600s)  id=bd041593-11a0-4524-82b6-309110da0761 name=/runtime.v1.RuntimeService/StopContainer
```
</details>

<details>
<summary>crictl exec output</summary>

```bash
$ crictl exec -it $CONTAINER_ID /bin/sh
DEBU[0000] Get runtime connection                       
DEBU[0000] Using runtime connection timeout: 5m0s       
DEBU[0000] Get image connection                         
DEBU[0000] ExecRequest: container_id:"8cba5e5cdf45b3b3f4322b84ab40856e81aa6f63ac27b03c30a416521738365b"  cmd:"/bin/sh"  tty:true  stdin:true  stdout:true 
DEBU[0000] ExecResponse: url:"http://127.0.0.1:38111/exec/OtbCYE8x" 
DEBU[0000] Exec URL: http://127.0.0.1:38111/exec/OtbCYE8x 
DEBU[0000] StreamOptions: {0xc000130028 0xc000130030 0xc000130038 true <nil>}
/ #
```
</details>

Unit tests added in `internal/oci/container_test.go` verify:
- `AddExecPID` accepts exec processes during graceful termination
- `AddExecPID` rejects exec processes once kill loop begins
- `DeleteExecPID` safely removes tracked processes
- `KillExecPIDs` terminates processes with appropriate signals
- `SetAsDoneStopping` correctly cleans up resources

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable exec registration and PID tracking during container lifecycle; execs can be started during the graceful-shutdown window but are rejected once forced-kill begins.
  * Improved handling of exec attempts when containers are not alive.

* **New Features**
  * PTY-backed exec startup now integrates with the improved atomic exec registration flow.

* **Tests**
  * Large new test suite covering exec behavior during termination and expanded PID/lifecycle tests.

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -866,28 +866,6 @@ func (c *Container) RuntimePathForPlatform(r *runtimeOCI) string {
 	return c.runtimePath
 }
 
-// AddExecPID registers a PID associated with an exec session.
-// It is tracked so exec sessions can be cancelled when the container is being stopped.
-// If the PID is conmon, shouldKill should be false, as we should not call SIGKILL on conmon.
-// If it is an exec session, shouldKill should be true, as we can't guarantee the exec process
-// will have a SIGINT handler.
-func (c *Container) AddExecPID(pid int, shouldKill bool) error {
-	c.stopLock.Lock()
-	defer c.stopLock.Unlock()
-
-	logrus.Debugf("Starting to track exec PID %d for container %s (should kill = %t) ...", pid, c.ID(), shouldKill)
-
-	// Allow exec during graceful termination (stopping=true but kill loop not begun)
-	// Only block exec once the kill loop has started (force kill with SIGKILL)
-	if c.stopKillLoopBegun {
-		return errors.New("cannot register an exec PID: container is being killed")
-	}
-
-	c.execPIDs[pid] = shouldKill
-
-	return nil
-}
-
 // StartExecCmd atomically starts an exec command and registers its PID.
 func (c *Container) StartExecCmd(cmd ExecStarter, shouldKill bool) (int, error) {
 	c.stopLock.Lock()
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -555,8 +555,6 @@ func (r *runtimeOCI) ExecContainer(ctx context.Context, c *Container, cmd []stri
 			execCmd.Stderr = stderr
 		}
 
-		// Atomically start the exec and register its PID to prevent race conditions
-		// where the exec starts but isn't registered before the kill loop begins.
 		pid, err := c.StartExecCmd(&execCmdWrapper{cmd: execCmd}, true)
 		if err != nil {
 			return err
@@ -721,8 +719,6 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 	cmd.ExtraFiles = append(cmd.ExtraFiles, childPipe, childStartPipe)
 	r.prepareEnv(cmd, true)
 
-	// Atomically start the command and register its PID to prevent race conditions
-	// where the exec starts but isn't registered before the kill loop begins.
 	pid, err := c.StartExecCmd(&execCmdWrapper{cmd: cmd}, false)
 	if err != nil {
 		childPipe.Close()
@@ -762,8 +758,6 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
 				c.DeleteExecPID(pid)
 			}
 		}()
-
-		// The exec PID was already registered atomically by StartExecCmd above
 
 		if r.handler.MonitorExecCgroup == config.MonitorExecCgroupContainer && r.config.InfraCtrCPUSet != "" {
 			// Update the exec's cgroup
@@ -1025,7 +1019,6 @@ func (r *runtimeOCI) StopLoopForContainer(ctx context.Context, c *Container, bm 
 				// Periodically check if the container is still running.
 				// This avoids busy-waiting and reduces resource usage while
 				// ensuring timely detection of container termination.
-				//
 				if err := c.Living(); err != nil {
 					// The initial container process either doesn't exist, or isn't ours.
 					if !errors.Is(err, ErrNotFound) {
--- a/internal/oci/container_test.go
+++ b/internal/oci/container_test.go
@@ -677,72 +677,19 @@ var _ = t.Describe("Container", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
-	t.Describe("AddExecPID", func() {
-		It("should succeed to add exec PID when container is not stopping", func() {
-			// Given
-			testPID := 12345
-
-			// When
-			err := sut.AddExecPID(testPID, true)
-
-			// Then
-			Expect(err).ToNot(HaveOccurred())
-		})
-
-		It("should succeed to add exec PID during graceful termination (before kill loop)", func() {
-			// Given
-			testPID := 12345
-			sut.SetAsStopping()
-
-			// When - Container is stopping but kill loop hasn't begun
-			err := sut.AddExecPID(testPID, true)
-
-			// Then - Should succeed because stopKillLoopBegun is false
-			Expect(err).ToNot(HaveOccurred())
-		})
-
-		It("should fail to add exec PID when kill loop has begun", func() {
-			// Given
-			testPID := 12345
-			sut.SetAsStopping()
-			sut.SetStopKillLoopBegun()
-
-			// When
-			err := sut.AddExecPID(testPID, false)
-
-			// Then - Should fail because stopKillLoopBegun is true
-			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("container is being killed"))
-		})
-
-		It("should succeed to add multiple exec PIDs before kill loop", func() {
-			// Given
-			pid1 := 12345
-			pid2 := 12346
-			sut.SetAsStopping()
-
-			// When
-			err1 := sut.AddExecPID(pid1, true)
-			err2 := sut.AddExecPID(pid2, false)
-
-			// Then - Both should succeed during graceful termination
-			Expect(err1).ToNot(HaveOccurred())
-			Expect(err2).ToNot(HaveOccurred())
-		})
-	})
 
 	t.Describe("DeleteExecPID", func() {
 		It("should succeed to delete existing exec PID", func() {
 			// Given
 			testPID := 12345
-			Expect(sut.AddExecPID(testPID, true)).To(Succeed())
+			Expect(addTestExecPID(sut, testPID, true)).To(Succeed())
 
 			// When
 			sut.DeleteExecPID(testPID)
 
 			// Then - Should not panic and PID should be removed
 			// We can verify by adding the same PID again (which would be tracked separately)
-			Expect(sut.AddExecPID(testPID, false)).To(Succeed())
+			Expect(addTestExecPID(sut, testPID, false)).To(Succeed())
 		})
 
 		It("should handle deleting non-existent PID gracefully", func() {
@@ -760,9 +707,9 @@ var _ = t.Describe("Container", func() {
 			pid1 := 12345
 			pid2 := 12346
 			pid3 := 12347
-			Expect(sut.AddExecPID(pid1, true)).To(Succeed())
-			Expect(sut.AddExecPID(pid2, false)).To(Succeed())
-			Expect(sut.AddExecPID(pid3, true)).To(Succeed())
+			Expect(addTestExecPID(sut,pid1, true)).To(Succeed())
+			Expect(addTestExecPID(sut,pid2, false)).To(Succeed())
+			Expect(addTestExecPID(sut,pid3, true)).To(Succeed())
 
 			// When
 			sut.DeleteExecPID(pid2)
@@ -779,7 +726,7 @@ var _ = t.Describe("Container", func() {
 		It("should be safe to delete same PID multiple times", func() {
 			// Given
 			testPID := 12345
-			Expect(sut.AddExecPID(testPID, true)).To(Succeed())
+			Expect(addTestExecPID(sut,testPID, true)).To(Succeed())
 
 			// When/Then - Multiple deletes should not panic
 			Expect(func() {
@@ -818,7 +765,7 @@ var _ = t.Describe("Container", func() {
 			// Given - Add a PID that doesn't exist
 			// KillExecPIDs will try to kill this PID, get ESRCH error, and should handle it
 			nonExistentPID := neverRunningPid
-			Expect(sut.AddExecPID(nonExistentPID, true)).To(Succeed())
+			Expect(addTestExecPID(sut,nonExistentPID, true)).To(Succeed())
 
 			// When/Then - Should handle ESRCH error and complete without panic
 			Expect(func() {
@@ -831,9 +778,9 @@ var _ = t.Describe("Container", func() {
 			pid1 := neverRunningPid
 			pid2 := neverRunningPid - 1
 			pid3 := neverRunningPid - 2
-			Expect(sut.AddExecPID(pid1, true)).To(Succeed())
-			Expect(sut.AddExecPID(pid2, false)).To(Succeed())
-			Expect(sut.AddExecPID(pid3, true)).To(Succeed())
+			Expect(addTestExecPID(sut,pid1, true)).To(Succeed())
+			Expect(addTestExecPID(sut,pid2, false)).To(Succeed())
+			Expect(addTestExecPID(sut,pid3, true)).To(Succeed())
 
 			// When/Then - Should attempt to kill all PIDs
 			Expect(func() {
@@ -847,8 +794,8 @@ var _ = t.Describe("Container", func() {
 			// but we can verify the function completes without error
 			pid1 := neverRunningPid
 			pid2 := neverRunningPid - 1
-			Expect(sut.AddExecPID(pid1, true)).To(Succeed())  // Should use SIGKILL
-			Expect(sut.AddExecPID(pid2, false)).To(Succeed()) // Should use SIGINT
+			Expect(addTestExecPID(sut,pid1, true)).To(Succeed())  // Should use SIGKILL
+			Expect(addTestExecPID(sut,pid2, false)).To(Succeed()) // Should use SIGINT
 
 			// When/Then - Should complete successfully
 			Expect(func() {
@@ -859,13 +806,13 @@ var _ = t.Describe("Container", func() {
 		It("should clear exec PIDs map after killing", func() {
 			// Given
 			pid1 := neverRunningPid
-			Expect(sut.AddExecPID(pid1, true)).To(Succeed())
+			Expect(addTestExecPID(sut,pid1, true)).To(Succeed())
 
 			// When
 			sut.KillExecPIDs()
 
 			// Then - Should be able to add the same PID again (map was cleared)
-			Expect(sut.AddExecPID(pid1, false)).To(Succeed())
+			Expect(addTestExecPID(sut,pid1, false)).To(Succeed())
 		})
 	})
 
@@ -1106,4 +1053,14 @@ func (m *mockExecStarter) Start() error {
 
 func (m *mockExecStarter) GetPid() int {
 	return m.pid
+}
+
+// addTestExecPID is a test helper that uses StartExecCmd to add an exec PID.
+func addTestExecPID(c *oci.Container, pid int, shouldKill bool) error {
+	mock := &mockExecStarter{
+		startFunc: func() error { return nil },
+		pid:       pid,
+	}
+	_, err := c.StartExecCmd(mock, shouldKill)
+	return err
 }
--- a/internal/oci/oci_unix.go
+++ b/internal/oci/oci_unix.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 
 	"github.com/creack/pty"
@@ -16,6 +17,27 @@ import (
 
 	"github.com/cri-o/cri-o/utils"
 )
+
+// ptyStarter wraps pty.Start to implement the ExecStarter interface.
+// It stores the pty file descriptor for later use after Start() is called.
+type ptyStarter struct {
+	cmd *exec.Cmd
+	pty *os.File
+}
+
+func (p *ptyStarter) Start() error {
+	var err error
+	p.pty, err = pty.Start(p.cmd)
+	return err
+}
+
+func (p *ptyStarter) GetPid() int {
+	return p.cmd.Process.Pid
+}
+
+func (p *ptyStarter) Pty() *os.File {
+	return p.pty
+}
 
 func Kill(pid int) error {
 	err := unix.Kill(pid, unix.SIGKILL)
@@ -33,20 +55,17 @@ func setSize(fd uintptr, size remotecommand.TerminalSize) error {
 }
 
 func ttyCmd(execCmd *exec.Cmd, stdin io.Reader, stdout io.WriteCloser, resizeChan <-chan remotecommand.TerminalSize, c *Container) error {
-	p, err := pty.Start(execCmd)
+	starter := &ptyStarter{cmd: execCmd}
+	pid, err := c.StartExecCmd(starter, true)
 	if err != nil {
 		return err
 	}
+	defer c.DeleteExecPID(pid)
+
+	p := starter.Pty()
 	defer p.Close()
 	// make sure to close the stdout stream
 	defer stdout.Close()
-
-	pid := execCmd.Process.Pid
-	if err := c.AddExecPID(pid, true); err != nil {
-		return err
-	}
-
-	defer c.DeleteExecPID(pid)
 
 	utils.HandleResizing(resizeChan, func(size remotecommand.TerminalSize) {
 		if err := setSize(p.Fd(), size); err != nil {
--- a/server/container_exec_test.go
+++ b/server/container_exec_test.go
@@ -12,6 +12,19 @@ import (
 	"github.com/cri-o/cri-o/internal/oci"
 )
 
+// mockExecStarter is a mock implementation of oci.ExecStarter for testing.
+type mockExecStarter struct {
+	pid int
+}
+
+func (m *mockExecStarter) Start() error {
+	return nil
+}
+
+func (m *mockExecStarter) GetPid() int {
+	return m.pid
+}
+
 // The actual test suite.
 var _ = t.Describe("ContainerExec", func() {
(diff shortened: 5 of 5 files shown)


sohankunkerkar:
@willianpaixao Thanks for working on this! The feature looks useful, but I have some concerns about potential race conditions that I think we should verify with integration tests before merging. Would you be okay adding one? You can refer to https://github.com/cri-o/cri-o/tree/main/test for more information.

server/container_exec.go
```diff
@@ -54,13 +52,8 @@ func (s *StreamService) Exec(ctx context.Context, containerID string, cmd []stri
...
 	}
 
-	if err := s.runtimeServer.ContainerServer.Runtime().UpdateContainerStatus(s.ctx, c); err != nil {
-		return err
-	}
-
-	cState := c.State()
-	if cState.Status != oci.ContainerStateRunning && cState.Status != oci.ContainerStateCreated {
-		return errors.New("container is not created or running")
+	if err := c.Living(); err != nil {
```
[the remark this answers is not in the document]
willianpaixao: @sohankunkerkar you mean this race condition here, right? Do you agree with Copilots suggestion?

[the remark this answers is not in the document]
sohankunkerkar: I don't think so, but it's always good to test this kind of behavior with integration tests. I can think of the following tests cases:
```sh 
1. exec during graceful termination
2. multiple execs during graceful
3. exec fails after stopped
4. long exec killed when kill loop starts
5. execsync/tty exec during graceful

```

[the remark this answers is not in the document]
willianpaixao: Thank you for the clarification. Test case 3 is already implemented, but the remaining were now added.

willianpaixao:
@sohankunkerkar integration tests added!

internal/oci/runtime_oci.go
```diff
@@ -939,34 +939,47 @@ func (r *runtimeOCI) StopLoopForContainer(ctx context.Context, c *Container, bm
...
 		// Ref: https://github.com/kubernetes/kubernetes/issues/124743
+		c.opLock.Lock()
 		c.KillExecPIDs()
 		c.state.Finished = time.Now()
 		c.opLock.Unlock()
 		c.SetAsDoneStopping()
 	}()
 
-	if c.state.Status == ContainerStatePaused {
+	c.opLock.Lock()
```
bitoku: The stop process is supposed to be atomic. We really need to be careful to change this.
Why do we need this change?

sohankunkerkar: IIUC, the granular locking still protects each runtime command and state modification. The only "non-atomic" window is the wait period where no state is being modified anyway. While other operations could theoretically interleave, the stop loop handles this gracefully by checking `Living()` periodically.

bitoku: > IIUC, the granular locking still protects each runtime command and state modification. The only "non-atomic" window is the wait period where no state is being modified anyway. 

For example, the state could change A -> Z -> B even though it's supposed to change A -> B. It's just a thought experiment and I don't know whether it'll happen or whether it has impact, but it could happen without lock.

> While other operations could theoretically interleave, the stop loop handles this gracefully by checking Living() periodically.

I'm not sure if this is true. Even if the stop operation handles it gracefully, other operations which interleave could have impact.

From readability point of view, opLock sounds it locks during operation, not each variable change.

Personally I don't think it has a problem, but not very confident. First I want to know why this change is needed, and if there's other options.

willianpaixao: > I'm not sure if this is true. Even if the stop operation handles it gracefully, other operations which interleave could have impact.

No state is modified during wait periods therefore the lock doesn't need to be held

>  I don't know whether it'll happen or whether it has impact, but it could happen without lock.

The Living() check (which the stop loop uses) is atomic and doesn't depend on intermediate states. Operations that could interleave (like exec) are intentionally allowed but dangerous operations (multiple stops) are still prevented.

bitoku: > Operations that could interleave (like exec) are intentionally allowed but dangerous operations (multiple stops) are still prevented.

I'm not sure what operations are dangerous or not.
Some operations supposed to be safe could have impact as well, because they are implemented based on the assumption that the stop operation is atomic.

Which part exactly prevents exec from running ?

sohankunkerkar: >Some operations supposed to be safe could have impact as well, because they are implemented based on the assumption that the stop operation is atomic.

While we're here, I think it's a good opportunity to make this part of the code a bit stronger. We've historically had a few locking-related issues in this part of the code. I think you raised a valid concern but, I'm also thinking about the cases where we might run into similar problems again.

These are the scenarios I was trying to guard against:
```sh
 1.  pause during stop should not deadlock or hang
 2.  external SIGKILL during stop should complete cleanly
 3.  delete during stop should not cause errors
 4.  exec at kill loop boundary should fail cleanly or be killed
 5.  update container resources during stop
 6.  multiple concurrent stops should not race
 7.  unpause during stop should not cause issues
 8.  container stats during stop should not deadlock
 9.  long running exec started before stop should be terminated
 10. execsync during stop should work in graceful period
 11. stop with zero timeout should kill immediately
 12. SIGTERM signal during stop should not conflict
 13. inspect during stop should return valid state
```
All of these are passing with the current changes. Can we think of any other edge cases where this code might break?

willianpaixao: Based on the granular locking changes in `StopLoopForContainer`, here are additional edge cases to consider:

#### State consistency edge cases:

1. **UpdateContainerStatus during graceful termination** - Runtime status updates racing with the stop loop's periodic `Living()` checks
2. **Checkpoint during stop** - Attempting to checkpoint the container during graceful termination period
3. **Attach during graceful termination** - Attaching to container stdio while it's being gracefully terminated

#### Timing/race conditions:

4. **Race between SetStopKillLoopBegun and AddExecPID** - An exec request arriving exactly when transitioning from graceful to kill loop (this is critical)
5. **Multiple execs racing to register** - Several concurrent exec operations trying to call `AddExecPID` during the grace -> kill transition
6. **Container removal during graceful termination** - `RemoveContainer` called while stop loop is running
7. **Start during stop** - Attempting to start a container that's mid-stop (should be blocked but worth verifying)

#### Process state edge cases:

8. **Process becomes zombie during unlocked window** - Container process transitions to Z/X state between lock release and next `Living()` check
9. **PID reuse during stop** - Container exits and PID is reused between periodic `Living()` checks (though `InitStartTime` verification should catch this)

#### Resource/metadata operations:

10. **UpdateResources during stop** - Specifically: what if resources are updated during the unlocked wait period between `Living()` checks?
11. **Pause/unpause racing with exec** - Container gets paused while exec is being registered or starting (critical for the new behavior)
12. **Log operations during stop** - Heavy log reading while container is stopping

#### Channel/synchronization:

13. **stopTimeoutChan overflow** - More than 10 rapid timeout updates (channel buffer = 10, though there's a default case)
14. **Watcher registration during SetAsDoneStopping** - Race between goroutine trying to register a watcher and cleanup closing all watchers

#### Network/volume cleanup:

15. **Network teardown during graceful period** - CNI operations racing with exec during graceful termination
16. **Volume unmount during graceful period** - Attempting to unmount volumes while exec might still be accessing them

I can write some tests for the above cases, but I'd like to do that in a follow up issue/PR.

internal/oci/runtime_oci.go
```diff
@@ -939,34 +939,47 @@ func (r *runtimeOCI) StopLoopForContainer(ctx context.Context, c *Container, bm
...
+	isPaused := c.state.Status == ContainerStatePaused
+	c.opLock.Unlock()
+
+	if isPaused {
+		c.opLock.Lock()
```
bitoku: isPaused could be changed when UnpauseContainer is called at this point.

And what happens if PauseContainer is called during termination? I'm not sure that operation is supported but it can happen.

internal/oci/runtime_oci.go
bitoku: > While we're here, I think it's a good opportunity to make this part of the code a bit stronger.

While I agree with the motivation to strengthen the code, I believe we need to proceed with extreme caution.
Given the edge cases @willianpaixao highlighted and the race condition I identified, I am hesitant to move forward with this change.
My concern is that there is a high risk of overlooking other potential race conditions.

Could we revisit the questions I raised earlier?
**Why is this granular locking necessary here?**
**Is it possible to implement exec during termination by modifying only the exec code?**

internal/oci/runtime_oci.go
willianpaixao: You're correct that `isPaused` could change if `UnpauseContainer` is called between the check and the action. However, the consequence is benign:
- If container was unpaused externally: the `resume` command will fail or no-op (already running)
- The stop loop continues normally in either case

That said, this could be made more robust by keeping the lock through the resume operation:

```go
c.opLock.Lock()
if c.state.Status == ContainerStatePaused {
    if _, err := r.runtimeCmd("resume", c.ID()); err != nil {
        log.Errorf(ctx, "Failed to unpause container %s: %v", c.Name(), err)
    }
}
c.opLock.Unlock()
```

I can update the PR if you'd like.

internal/oci/runtime_oci.go
willianpaixao: #### Why is this granular locking necessary here?

The original implementation holds `opLock` for the **entire duration** of `StopLoopForContainer`, which can be many seconds (up to the graceful termination timeout, typically 30s).

During this entire wait period, any operation that needs `opLock` is blocked - including exec. This means `kubectl exec` fails during graceful termination even though the container process is still running and can accept exec sessions.

The granular locking releases the lock **during wait periods** where no state is being modified:
- Lock held during: state reads, state writes, runtime commands (kill/resume)
- Lock released during: sleeping/waiting for container to exit

This allows exec to acquire the lock and proceed while the stop loop is waiting.

While I believe you both know that, I take you posed the question to maybe have the reasoning on writing?

#### Is it possible to implement exec during termination by modifying only the exec code?

**No**, modifying only the exec code is not sufficient.

The `ExecContainer` function at `internal/oci/runtime_oci.go:464` doesn't explicitly acquire `opLock`, but look at the call chain:

```
server/container_exec.go:55 -> c.Living() -> accesses c.state.InitPid, c.state.Pid
```

The `Living()` method at `internal/oci/container.go:558` accesses `c.state` fields that are protected by `opLock`. As noted in Copilot's review, calling `Living()` without the lock is a race condition.

Even if we made exec completely lock-free (which would be unsafe), the fundamental blocker remains: **the stop loop owns `opLock` for the entire duration**. Any safe access to container state during exec would block waiting for the stop loop to finish.

bitoku: InitPid and Pid basically are initialized at the startup time, and unchanged after that.
I don't see any issue not acquiring lock.

willianpaixao: Ok, the discussion is split in several places and also divided between my attempt to improve the behavious and the formalization of how it should work.

Can we define and restrict the scope of this PR? I can raise different issues and subsequent PRs.

So @bitoku, do you have request to changes in this PR? Or can we move forward to approval?

bitoku: As it stands, I don't want to move it forward.
At least, stop operation has too granular lock, which is error-prone IMO.

We could free lock during waiting for the termination, as we do in UpdateContainerStatus.
My suggestion is to not acquiring lock during exec, but this one also, I'm not 100% sure is safe.

Either way, I want more time to review. sorry for keeping you wait.

bitoku: @willianpaixao Thank you for the patience.

I found another race condition with `StopContainer` and `UpdateContainerStatus`
 
1. call `StopContainer`
2. while the lock is free, call `UpdateContainerStatus`
3. the container terminates for some reasons
4. `UpdateContainerStatus` updates `c.state`
5. `StopContainer` finishes because the container is stopped
6. in the end, `StopContainer` overwrites `c.state.Finished`, which causes race condition

We can prevent operations other than safe operations by checking `c.stopping`, but I don't think it's worth doing because basically the safe operations shouldn't need lock. If it needs lock, it means it can access to the variables which could vary and it's not safe.

As I said in the thread, Living() is safe to run without RLock.
If no other concerns, we should remove RLock from the exec, instead of removing Lock from the stop.

cc @sohankunkerkar

internal/oci/container.go
```diff
@@ -873,8 +873,10 @@ func (c *Container) AddExecPID(pid int, shouldKill bool) error {
 
 	logrus.Debugf("Starting to track exec PID %d for container %s (should kill = %t) ...", pid, c.ID(), shouldKill)
 
-	if c.stopping {
-		return errors.New("cannot register an exec PID: container is stopping")
+	// Allow exec during graceful termination (stopping=true but kill loop not begun)
+	// Only block exec once the kill loop has started (force kill with SIGKILL)
+	if c.stopKillLoopBegun {
+		return errors.New("cannot register an exec PID: container is being killed")
```
bitoku: We should change the AddExecPID function entirely.
It can run exec without adding the pid to the execPIDs list. It cause cri-o to fail to kill the exec process when the container stops.
Running exec and adding its pid should be done atomically.

willianpaixao: I've rewritten that part, you can see me latest commit.

internal/oci/runtime_oci.go
willianpaixao: I was able to reproduce that. Wrote a test file but won't commit.
I'd applied your instructions and the race test is not detected anymore.
You can check the latest commit.

sohankunkerkar: Nice catch @bitoku!

willianpaixao:
@coderabbitai generate docstrings

internal/oci/runtime_oci.go
```diff
@@ -984,6 +992,7 @@ func (r *runtimeOCI) StopLoopForContainer(ctx context.Context, c *Container, bm
 				// Periodically check if the container is still running.
 				// This avoids busy-waiting and reduces resource usage while
 				// ensuring timely detection of container termination.
+				//
```
haircommander: nit: drop this

server/container_exec.go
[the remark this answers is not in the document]
haircommander: Living() was designed to not race against opLock, so if it does, that's a problem outside of the scope of this PR

internal/oci/container.go
```diff
@@ -873,15 +873,46 @@ func (c *Container) AddExecPID(pid int, shouldKill bool) error {
...
 	}
 
 	c.execPIDs[pid] = shouldKill
 
 	return nil
 }
 
+// StartExecCmd atomically starts an exec command and registers its PID.
+func (c *Container) StartExecCmd(cmd ExecStarter, shouldKill bool) (int, error) {
```
haircommander, line 870: I think this should be internal `startExecCmd` as it shouldn't be used outside of this package

internal/oci/runtime_oci.go
```diff
@@ -712,14 +725,12 @@ func (r *runtimeOCI) ExecSyncContainer(ctx context.Context, c *Container, comman
...
+				c.DeleteExecPID(pid)
 			}
 		}()
 
-		// A neat trick we can do is register the exec PID before we send info down the start pipe.
-		// Doing so guarantees we can short circuit the exec process if the container is stopping already.
-		if err := c.AddExecPID(cmd.Process.Pid, false); err != nil {
-			return err
-		}
+		// The exec PID was already registered atomically by StartExecCmd above
```
haircommander: I think this can be dropped personally

internal/oci/container.go
haircommander: nvm, the tests are external :upside_down_face:

haircommander:
couple of nits, great work here

internal/oci/container.go
bitoku: Is this still needed? I think we can remove this.

internal/oci/runtime_oci.go
```diff
@@ -101,6 +101,19 @@ type exitCodeInfo struct {
...
+}
+
+func (w *execCmdWrapper) Start() error {
+	return w.cmd.Start()
+}
+
+func (w *execCmdWrapper) GetPid() int {
+	return w.cmd.Process.Pid
+}
+
```
bitoku, line 116: I guess this is for testing. Can we extend this interface? It should be cleaner.
https://github.com/cri-o/cri-o/blob/d19a9d641b7f5ce2b86f57ead5e910d257f34fd1/utils/cmdrunner/cmdrunner.go#L13-L19

haircommander: +1

willianpaixao: Adding `Start()` and `GetPid()` to `CommandRunner` would conflate command creation with command execution, violating single responsibility. Additionally, `CommandRunner` is a global singleton used throughout the codebase for command creation, changing its contract would have wider implications.

The `ExecStarter` interface is minimal (2 methods), focused, and only used within `internal/oci/` for the specific purpose of atomic exec PID registration during container termination. I believe keeping it separate is the cleaner design.

Happy to discuss alternative approaches if you have a different design in mind.

willianpaixao:
Hey guys, thanks for the help. I've addressed everything I could find. The comments are scattered and it's getting messy at this point. So I'm sorry if I forgot to change something. Just comment again and I will tackle it. Wish you all a Merry Christmas.

internal/oci/runtime_oci.go
bitoku: Fair enough. We can follow up if necessary.

PR #9615, opened 2025-11-24
Preserve knownRepoDigests order in ImageStatus

saschagrunert:
#### What type of PR is this?

/kind bug

For multi-arch images, RepoDigests are currently sorted alphabetically, which can place the manifest list digest before the platform-specific digest that was actually pulled.

This causes issues for consumers like Kubernetes that store the digest from PullImage and expect to find it in RepoDigests (ideally first).

**This PR fixes the issue in two ways:**
1. During image pull: reorder the image's stored names to place the actual pulled digest first
2. During ImageStatus: preserve the order of knownRepoDigests (from image's stored names) and only sort additional computed digests

This ensures the pulled digest appears first while maintaining deterministic ordering for consistency.

The alphabetical sort was originally added in commit 061e6d52d for test consistency. This PR maintains that consistency while ensuring meaningful ordering.

I found this issue during the serial test debugging as part of https://github.com/kubernetes/kubernetes/pull/135369. Vendoring that change into test-infra should make the `KubeletEnsureSecretPulledImages` serial tests green.

```release-note
None
```

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -324,12 +324,43 @@ func (svc *imageService) buildImageResult(image *storage.Image, cacheItem imageC
 
 	sort.Strings(repoTagStrings)
 
+	// Build repo digest strings with stable ordering:
+	// - Keep knownRepoDigests (from image's stored names) in original order
+	// - Sort additional computed digests alphabetically
+	// This ensures the PullImage digest appears first, allowing Kubernetes to rely on
+	// the first RepoDigest for credential tracking (kubernetes/kubernetes#135369).
+
+	// Convert known digests to strings (preserve order)
+	// `digests` contains canonical references from the image's stored names - these are
+	// the "known" digests that should appear first in RepoDigests in their original order.
+	// `repoDigests` (returned from makeRepoDigests earlier) contains the union of:
+	// - known digests from the image's stored names
+	// - computed digests from combining the image digest with repository names from tags
 	repoDigestStrings := make([]string, 0, len(repoDigests))
-	for _, d := range repoDigests {
+	for _, d := range digests {
 		repoDigestStrings = append(repoDigestStrings, d.String())
 	}
 
-	sort.Strings(repoDigestStrings)
+	// Collect additional (computed) digests
+	knownSet := make(map[string]bool, len(digests))
+	for _, d := range repoDigestStrings {
+		knownSet[d] = true
+	}
+
+	additionalDigests := make([]string, 0)
+
+	for _, d := range repoDigests {
+		digestString := d.String()
+		if !knownSet[digestString] {
+			additionalDigests = append(additionalDigests, digestString)
+		}
+	}
+
+	// Sort only the additional digests
+	sort.Strings(additionalDigests)
+
+	// Combine: known (in original order) + additional (sorted)
+	repoDigestStrings = append(repoDigestStrings, additionalDigests...)
 
 	previousName := ""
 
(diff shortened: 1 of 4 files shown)


saschagrunert:
@cri-o/cri-o-maintainers PTAL

internal/storage/image.go
```diff
@@ -891,6 +925,33 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+	// The manifestDigest may differ from the requested reference for multi-arch images
+	// (platform-specific vs manifest list digest) or pull-by-tag (tag -> digest form).
+	// Ensure the pulled digest appears first in image names so Kubernetes can find it
+	// in ImageStatus RepoDigests for credential tracking (kubernetes/kubernetes#135369).
+	// Note: This only applies to regular container images, not OCI artifacts.
+	canonicalRefString := canonicalRef.String()
+	if !isOCIArtifact && canonicalRefString != imageName.Raw().String() {
+		// The pulled digest differs from the requested reference
+		// Look up the image to reorder its names (reuse destRef from above)
+		_, img, err := istorage.ResolveReference(destRef)
```
mtrmac, line 937: ⚠️ 

If the original pull was by tag, this resolves the image by tag, and races against a concurrent pull from another namespace (with a different signature policy) and a registry moving the tag. This might refer to a completely different image which does not satisfy the current puller’s policy.

internal/storage/image.go
```diff
@@ -891,6 +925,33 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+
+		// Reorder names to put the platform-specific digest first
+		reorderedNames := []string{canonicalRefString}
+		for _, name := range img.Names {
+			if name != canonicalRefString {
+				reorderedNames = append(reorderedNames, name)
+			}
+		}
+
+		if err := store.SetNames(img.ID, reorderedNames); err != nil {
```
mtrmac, line 950: “Read image names; (drop storage locks); `SetNames`” is racy against concurrent updates to `Image.Names`; this might e.g. lose tags.

mtrmac, line 950: Whatever this was supposed to fix, this can’t be the fix: there can be only one order within the store, so if there are two deduplicated images that need this treatment, they can’t both get it.

internal/storage/image.go
```diff
@@ -891,6 +925,33 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+	if !isOCIArtifact && canonicalRefString != imageName.Raw().String() {
+		// The pulled digest differs from the requested reference
+		// Look up the image to reorder its names (reuse destRef from above)
+		_, img, err := istorage.ResolveReference(destRef)
+		if err != nil {
+			return RegistryImageReference{}, fmt.Errorf("looking up pulled image: %w", err)
+		}
+
+		// Reorder names to put the platform-specific digest first
+		reorderedNames := []string{canonicalRefString}
```
mtrmac, line 943: Also, for pulls by tag (when `destRef` has a tag), `canonicalRefString` might not be in the c/storage `.Names` array at all.

So, on net, this PR might add a repo@digest entry to an image which never had that digest, and then … we are in a completely uncharted territory.

internal/storage/image.go
saschagrunert: Makes sense, we probably can revert this one in favor of https://github.com/cri-o/cri-o/pull/9728 (not exactly sure, though).

Ref https://github.com/cri-o/cri-o/pull/9730

PR #9616, opened 2025-11-24
[release-1.27] OCPBUGS-55964: Pause container during checkpointing

adrianreber:
#### What type of PR is this?

/kind bug

The initial implementation of checkpointing in CRI-O was based on Podman and initially the default workflow would have been that the container is stopped after checkpointing. CRIU will just kill all processes in the container.

As the initial Kubernetes checkpointing feature is based on the Forensic Container Checkpointing KEP, the container keeps on running after checkpointing.

If the container keeps on running after checkpointing it can happen that the files in the container are changed once we put the into the checkpoint archive. This means that the files are different during restore than they were during checkpointing.

CRIU aborts restoring if the file size of any open file has changed as restoring will put the FD pointer at the same location as during checkpointing. If the file, however, is different this can lead to data loss or crashes.

To solve this, this commit pauses the container before checkpointing and unpauses it after the checkpoint archive has been written to disk.

As checkpointing only works with the OCI runtimes currently, we do not have to handle the pause during restore.

The OCI runtime (runc/crun) uses the cgroup freezer to pause the process. CRIU also uses the cgroup freezer to pause all processes in the container. CRIU does not change the state of the cgroup freezer if the cgroup is already frozen. So the cgroup was already always frozen during checkpointing. With this change the frozen time is now controlled by CRI-O and not CRIU, but we still do not have to handle it during restore.

(cherry picked from commit 3d450274e8ab959436147edeb72d1e08b89d4edf)

Fixes: OCPBUGS-55964

```release-note
None
```

### DIFF
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -42,6 +42,35 @@ func (c *ContainerServer) ContainerCheckpoint(
 		return "", fmt.Errorf("container %s is not running", ctr.ID())
 	}
 
+	// At this point the container needs to be paused. As we first checkpoint
+	// the processes in the container and the container will continue to run
+	// after checkpointing, there is a chance that the changed files we include
+	// in the checkpoint archive might change by the now again running processes
+	// in the container.
+	// Assuming this uses runc/crun (the OCI runtime is currently the only one
+	// supporting checkpointing), PauseContainer() will use the cgroup freezer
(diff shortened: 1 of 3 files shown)


bitoku:
/lgtm
The prow test failures are long-standing ones

@cri-o/cri-o-maintainers PTAL
It needs manual merge since the prow tests can't pass.

PR #9628, opened 2025-12-01
Add image content cache to retain compressed layers for P2P distribution

sohankunkerkar:
This change helps enabling p2p image distribution support in CRI-O.

#### What type of PR is this?

/kind feature

```release-note
Add image content cache to retain compressed layers for P2P distribution
```

## Summary by CodeRabbit

* **New Features**
  * Image content cache for compressed image layers to speed pulls and enable P2P sharing.
  * In-process caching during pulls with automatic garbage collection to remove unreferenced blobs.

* **Documentation**
  * New CLI flag and config option to enable/configure the image content cache; updated manpage, config templates, and shell completions (bash, fish, zsh).
  * New GC metrics exposed: totals, durations, blobs removed, and bytes freed.

* **Tests**
  * Extensive unit and integration tests covering caching, persistence, deduplication, concurrency, and GC.


internal/storage/blobcache_wrapper.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	select {
+	case cacheErr := <-errChan:
+		if cacheErr != nil {
+			log.Warnf(ctx, "Blob caching failed but pull succeeded: %v", cacheErr)
+		}
+	case <-ctx.Done():
+		log.Warnf(ctx, "Context cancelled while waiting for blob cache")
+	}
```
saschagrunert, line 130: The goroutine that writes to the blob cache could finish after PutBlob returns. The code uses channels to wait, but the `select` statement has a `case <-ctx.Done()` that could exit early if context is cancelled.

We could close `pw` before waiting, and add a timeout instead of ctx.Done() to prevent goroutine leaks:

```go 
pw.Close()

// Wait with timeout instead of context cancellation
select {
case cacheErr := <-errChan:
    // ...
case <-time.After(30 * time.Second):
    log.Warnf(ctx, "Timeout waiting for blob cache")
}
```

internal/storage/image.go
```diff
@@ -992,7 +1038,64 @@ func (svc *imageService) DeleteImage(systemContext *types.SystemContext, id Stor
...
+				} `json:"layers"`
+			}
+
+			if json.Unmarshal(data, &m) == nil {
+				for _, l := range m.Layers {
+					referenced[l.Digest] = true
+				}
+			}
+		}
+	}
```
saschagrunert, line 1170: The GC implementation iterates all images and parses all their "big data" blobs to extract layer digests. This could be expensive with many images.

We could:
1. Filter to manifest-related big data names (e.g., manifest, manifest-*)
2. Add metrics/logging for GC duration and blobs removed
3. Consider async GC to avoid blocking image operations

saschagrunert (commented):
I think that's a good start! We definately need docs on how to use the feature, though. 

I think it would be nice to have metrics for the GC operations as well as an async GC.

sohankunkerkar:
Here's the demo for spegel + cri-o integration: https://asciinema.org/a/dxa6WA2jYS5plirMVnNPbeBDO

docs/crio.8.md
```diff
@@ -421,6 +425,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--read-only**: Setup all unprivileged containers to run as read-only. Automatically mounts the containers' tmpfs on '/run', '/tmp' and '/var/tmp'.
 
+**--retain-image-content**: Retain the original compressed image content for distribution (default: false).
```
haircommander: why not just have this implicit if image-content-cache-dir is empty then it's false, otherwise true

internal/blobcache/blobcache.go
```diff
@@ -0,0 +1,331 @@
...
+
+	info.Sources = append(info.Sources, Source{Registry: registry, Repository: repository})
+	bc.metadata.Blobs[digestStr] = info
+
+	log.Debugf(ctx, "Added source %s/%s to blob %s", registry, repository, dgst)
+
+	return bc.saveMetadataLocked()
+}
+
+func (bc *BlobCache) saveMetadataLocked() error {
```
haircommander: I'm not crazy about calling this *Locked if there is no unlocked/public alternative. could just use a comment saying it assumes the lock is taken already

internal/blobcache/blobcache.go
```diff
@@ -0,0 +1,331 @@
...
+	tmpPath := tmpFile.Name()
+	defer os.Remove(tmpPath)
+
+	if _, err := tmpFile.Write(data); err != nil {
+		tmpFile.Close()
+
+		return fmt.Errorf("writing metadata: %w", err)
+	}
+
+	if err := tmpFile.Sync(); err != nil {
```
haircommander: I'm not positive we need this

internal/blobcache/blobcache_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"path/filepath"
+	"sync"
+	"testing"
+
+	digest "github.com/opencontainers/go-digest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNew(t *testing.T) {
```
haircommander: can we translate these to ginkgo?

sohankunkerkar:
@coderabbitai review

docs/crio.8.md
saschagrunert: I was thinking the same, but then considered having strong defaults for the directory value.

sohankunkerkar: @haircommander @saschagrunert I kept only image-content-cache-dir. Let me know if the UX doesn't look right.

haircommander: we could make a recommendation in the docs to achieve that @saschagrunert ?

saschagrunert: Yeah, sounds fair.

sohankunkerkar: Done!

internal/blobcache/blobcache.go
sohankunkerkar: Fixed

docs/crio.conf.5.md
```diff
@@ -525,6 +525,11 @@ The valid values are "enforcing" and "disabled", and the default is "enforcing".
 If "enforcing", an image pull will fail if a short name is used, but the results are ambiguous.
 If "disabled", the first result will be chosen.
 
+**image_content_cache_dir**=""
+The directory where compressed layer blobs are cached for P2P image distribution.
+When set, enables caching of compressed image layers that can be served to peer nodes.
+If empty, CRI-O will not retain the image content cache.
```
bitoku, line 531: I think it's safe to say it's experimental.
and we should call out this will more than double the disk consumption.

internal/blobcache/blobcache.go
```diff
@@ -0,0 +1,359 @@
...
+
+	realPath, err := filepath.EvalSymlinks(rootDir)
+	if err != nil {
+		return nil, fmt.Errorf("resolving cache path: %w", err)
+	}
+
+	rootDir = realPath
+
+	// Create blobs base directory (algorithm subdirs created on-demand).
+	blobsDir := filepath.Join(rootDir, "blobs")
```
bitoku, line 83: I wonder if we can use OCI layout.
Does it make difficult to manage?
https://github.com/opencontainers/image-spec/blob/main/image-layout.md

internal/blobcache/blobcache.go
```diff
@@ -0,0 +1,359 @@
...
+
+// GCStats contains statistics from a garbage collection operation.
+type GCStats struct {
+	BlobsRemoved int
+	BytesFreed   int64
+}
+
+// GarbageCollect removes blobs that are no longer referenced by any image.
+// It returns statistics about the GC operation.
+func (bc *BlobCache) GarbageCollect(ctx context.Context, referencedDigests map[string]bool) (GCStats, error) {
```
bitoku, line 293: Do you know in which case there's orphaned image cache?
It seems the cache is tied to the image lifecycle. I wonder if we really need this Garbage collection mechanism instead of just a delete function.

bitoku (commented):
Do we have a plan to discuss it with podman team?
It should be beneficial for them too.
If they make something in containers/container-libs, we can avoid double work.

sohankunkerkar:
> Do we have a plan to discuss it with podman team? It should be beneficial for them too. If they make something in containers/container-libs, we can avoid double work.

I’d prefer to keep this implementation inside CRI-O for now. We previously discussed pushing this down into containers/storage or containers/image, but it didn’t seem like there was interest in supporting the feature at that level. So I think it’s reasonable for us to try it here first, get feedback, and then revisit whether it makes sense to collaborate with the podman team later depending on how things evolve.

docs/crio.conf.5.md
sohankunkerkar: >I think it's safe to say it's experimental.

Yeah, that's a good call but I wouldn't call this as an experimental though. Ideally, we would like to support this API. We might refactor things depending on how things go.

> we should call out this will more than double the disk consumption.

Yeah, will update the docs.

internal/blobcache/blobcache.go
sohankunkerkar: Yeah, I will do that as a follow-up item.

docs/crio.conf.5.md
bitoku: experimental doesn't mean we're not sure if we support this.
It means we could change the behavior (for example we could discuss the way of GC). It's just a kind of graduation process.

sohankunkerkar: hmm... I liked your idea but I don't think we ever followed this norm (for the GA process) of saying a field `[EXEPERIMENTAL]` except for the one place.

bitoku: It doesn't have to be "experimental".
I just want to let users know that the behavior could change without notice. otherwise we should follow the deprecation process.
For example changing to oci layout may be a breaking change.

sohankunkerkar: >For example changing to oci layout may be a breaking change.

I'm still confused. Can you elaborate more on the behavior side of this feature? We're just storing the compressed layer tarballs by digest, following the standard OCI format. If that changes, it would break image distribution broadly, not just this image cache. I don’t think we need a special disclaimer for this feature.

bitoku: You're right. I didn't understand the underlying structure.
It may need some additional files, but they should be trivial.

I still think adding experimental doesn't hurt and give us time to improve the feature.
However, if we won't change the behavior (and we'll have the deprecation period if we do) and we won't add special treatment to keep backward consistency, it's not a requirement.

internal/blobcache/blobcache.go
sohankunkerkar: Orphaned blobs can occur from: shared layers between images, interrupted pulls, CRI-O restarts during pulls, or metadata corruption.
IIUC, direct deletion is hard because you’d need per-layer reference counting. Multiple images can share the same layers. GC is simpler: it scans all stored images, finds every layer they reference, and removes only the blobs that are no longer used.

bitoku:
@sohankunkerkar Do we want to include this in 1.35 release? or can we defer it?

internal/blobcache/blobcache.go
bitoku: thanks, c/storage oci layout implementation should have the feature. We can revisit it if we use oci layout for it.

sohankunkerkar:
> @sohankunkerkar Do we want to include this in 1.35 release? or can we defer it?

Yeah, we can defer it for later but I wanted to get this in before the shutdown (if feasible).

xw19:
Hi @sohankunkerkar 👋 This is a great feature and I'd love to see it land. It looks like it's gone quiet over the last few months, are you still planning to push it forward? If you're stretched thin, I'm happy to help out in whatever way is most useful. Just let me know what works for you.

PR #9634, opened 2025-12-03
Allow containers to use both host network and user namespace

HirazawaUi:
#### What type of PR is this?

/kind feature

This is the implementation of [KEP-5607](https://github.com/kubernetes/enhancements/issues/5607) in cri-o.

In KEP-5607, we allow containers to use both user namespace and host network simultaneously. This requires us to change the sys mount to a bind mount when the user employs both user namespace and host network, because mounting sys in the standard way is a privileged operation, which would prevent the container from using the standard /sys mount.

```release-note
Allow containers to use both host network and user namespace.
```

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * System filesystem mounts inside containers now respect user-namespace settings, improving isolation and compatibility when host networking is used.
  * Sandbox ID mappings are consulted during container filesystem setup to ensure consistent ownership and access behavior across environments.
  * Container startup now fails early on mapping retrieval errors, adding clearer error handling for more reliable cross-platform initialization.

### DIFF
--- a/server/container_create_linux.go
+++ b/server/container_create_linux.go
@@ -867,22 +867,35 @@ func (s *Server) specSetDevices(ctr ctrfactory.Container, sb *sandbox.Sandbox) e
 	return ctr.SpecAddDevices(configuredDevices, annotationDevices, privilegedWithoutHostDevices, s.config.DeviceOwnershipFromSecurityContext)
 }
 
-func addSysfsMounts(ctr ctrfactory.Container, containerConfig *types.ContainerConfig, hostNet bool) {
+func addSysfsMounts(ctr ctrfactory.Container, containerConfig *types.ContainerConfig, hostNet bool, sb *sandbox.Sandbox, containerIDMappings *idtools.IDMappings) {
+	usernsEnabled := containerIDMappings != nil
+
 	// If the sandbox is configured to run in the host network, do not create a new network namespace
 	if hostNet {
 		if !isInCRIMounts("/sys", containerConfig.GetMounts()) {
-			ctr.SpecAddMount(rspec.Mount{
-				Destination: "/sys",
-				Type:        "sysfs",
-				Source:      "sysfs",
-				Options:     []string{"nosuid", "noexec", "nodev", "ro"},
-			})
-			ctr.SpecAddMount(rspec.Mount{
-				Destination: cgroupSysFsPath,
-				Type:        "cgroup",
-				Source:      "cgroup",
-				Options:     []string{"nosuid", "noexec", "nodev", "relatime", "ro"},
-			})
+			// When using both host network and user namespace, use bind mount for /sys
+			// to avoid sysfs mounting failures in this configuration
+			if usernsEnabled {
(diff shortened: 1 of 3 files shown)


saschagrunert:
Thank youf for the PR @HirazawaUi, please sign-off your commit. 

/ok-to-test

server/container_create_linux.go
```diff
@@ -867,22 +867,29 @@ func (s *Server) specSetDevices(ctr ctrfactory.Container, sb *sandbox.Sandbox) e
...
+					Options:     []string{"nosuid", "noexec", "nodev", "ro", "rbind"},
+				})
+			} else {
+				ctr.SpecAddMount(rspec.Mount{
+					Destination: "/sys",
+					Type:        "sysfs",
+					Source:      "sysfs",
+					Options:     []string{"nosuid", "noexec", "nodev", "ro"},
+				})
+			}
```
giuseppe, line 898: you still need the mount for cgroup here

HirazawaUi: Sorry, I'm still not very familiar with this area. My understanding is that `/sys/fs/cgroup` has already been bind-mounted as part of `/sys`, and mounting the cgroup type again would cause a "value too large for defined data type" error, preventing the container from functioning properly. If I'm mistaken, please correct me.

giuseppe: but in this case it is not a bind mount, you've deleted the previous code to mount a cgroup file system.  We still need it when you are mounting a fresh sysfs

HirazawaUi: Oh, you're right. It's fixed now.

saschagrunert:
/lgtm cancel

Please fix the linter: https://github.com/cri-o/cri-o/actions/runs/19898717159/job/57037105155?pr=9634

HirazawaUi:
> Please fix the linter: https://github.com/cri-o/cri-o/actions/runs/19898717159/job/57037105155?pr=9634

Thanks for the heads up. The lint error has been fixed.

server/container_create_linux.go
```diff
@@ -867,22 +867,35 @@ func (s *Server) specSetDevices(ctr ctrfactory.Container, sb *sandbox.Sandbox) e
 	return ctr.SpecAddDevices(configuredDevices, annotationDevices, privilegedWithoutHostDevices, s.config.DeviceOwnershipFromSecurityContext)
 }
 
-func addSysfsMounts(ctr ctrfactory.Container, containerConfig *types.ContainerConfig, hostNet bool) {
+func addSysfsMounts(ctr ctrfactory.Container, containerConfig *types.ContainerConfig, hostNet bool, sb *sandbox.Sandbox, containerIDMappings *idtools.IDMappings) {
```
haircommander, line 870: I think this would be better as `hostUser` and do the `usernsEnabled` check in the caller. WDYT?

HirazawaUi: Of course. This code was modified in PR #9660, and some additional cleanup was performed.

PR #9636, opened 2025-12-04
server: skip processing early if we get a nil adjustment from NRI.

klihub:
#### What type of PR is this?

/kind other

Skip processing early if we get a nil adjustment from NRI instead of going through the motions. Currently NRI never returns a nil adjustment, using instead a fully instantiated one without any content. There is a pending PR to change that. This PR prepares us for that.

None

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during container creation to prevent processing when certain conditions aren't met, enhancing stability.

### DIFF
--- a/server/nri-api.go
+++ b/server/nri-api.go
@@ -118,6 +118,10 @@ func (a *nriAPI) createContainer(ctx context.Context, specgen *generate.Generato
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


saschagrunert:
/ok-to-test

Thanks!

PR #9643, opened 2025-12-08
Fix pod stop timeout allocation for containers and infra

saschagrunert:
#### What type of PR is this?

/kind failing-test

When stopping a pod sandbox with a deadline set via context, the stopTimeoutFromContext() function was called twice:
1. Once for regular containers
2. Once for the infra container

Both calls would attempt to use all remaining time from the deadline. This caused issues when the first phase consumed most/all of the timeout, leaving 0 seconds for the infra container stop.

Fix this by calculating the timeout once and allocating it to regular containers after reserving a small amount (1 second) for the infra container. The infra container stop then uses the full totalTimeout, allowing it to benefit from any time saved when regular containers stop earlier than their allocation. For very small timeouts (less than 2 seconds), regular containers get half the timeout.

None

This was particularly problematic on slower systems (e.g., arm64 CI runners) where operations took longer. For example, with a 5s deadline:
- CNI teardown: ~127ms
- Container stop: 4s (used all remaining time)
- Infra stop + cleanup: needed ~300ms but got 0s
- Result: DeadlineExceeded error

Fixes the flaky "pod remove with timeout from context" test on arm64.

```release-note
Fixed pod sandbox stop timeout allocation to properly distribute deadline between container and infra container stops, preventing timeout failures on slower systems.
```

### DIFF
--- a/server/sandbox_stop_freebsd.go
+++ b/server/sandbox_stop_freebsd.go
@@ -6,11 +6,12 @@ import (
 	"fmt"
 
 	"go.podman.io/storage"
+	"golang.org/x/sync/errgroup"
+	types "k8s.io/cri-api/pkg/apis/runtime/v1"
+
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/log"
 	oci "github.com/cri-o/cri-o/internal/oci"
-	"golang.org/x/sync/errgroup"
-	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
 func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error {
@@ -30,6 +31,24 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error 
 		return nil
 	}
 
+	// Calculate the timeout once. Regular containers get most of the timeout,
+	// reserving a small amount for infra container shutdown. The infra container
+	// will use the full totalTimeout, allowing it to use any time saved from
+	// containers stopping earlier than their allocated timeout.
+	totalTimeout := stopTimeoutFromContext(ctx)
+
+	const infraReservedTimeout int64 = 1 // 1 second reserved for infra container
+
+	var containerTimeout int64
+
+	if totalTimeout < infraReservedTimeout*2 {
+		// If total timeout is too small, split evenly
+		containerTimeout = totalTimeout / 2
+	} else {
+		// Reserve fixed time for infra, give rest to containers
+		containerTimeout = totalTimeout - infraReservedTimeout
+	}
+
 	errorGroup := &errgroup.Group{}
 
 	for _, ctr := range sb.Containers().List() {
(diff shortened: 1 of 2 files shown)


server/sandbox_stop_linux.go
```diff
@@ -62,7 +70,7 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error
 	}
 
 	podInfraContainer := sb.InfraContainer()
-	if err := s.stopContainer(ctx, podInfraContainer, stopTimeoutFromContext(ctx)); err != nil && !errors.Is(err, storage.ErrContainerUnknown) {
+	if err := s.stopContainer(ctx, podInfraContainer, infraTimeout); err != nil && !errors.Is(err, storage.ErrContainerUnknown) {
 		return fmt.Errorf("failed to stop infra container for pod sandbox %s: %w", sb.ID(), err)
 	}
```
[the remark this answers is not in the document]
saschagrunert: Interesting :thinking: let's fix that right here.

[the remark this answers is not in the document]
saschagrunert: Done :+1:

saschagrunert:
@coderabbitai review

saschagrunert:
@cri-o/cri-o-maintainers PTAL

server/sandbox_stop_freebsd.go
```diff
@@ -30,24 +31,37 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error
 		return nil
 	}
 
+	// Calculate the timeout once and split it between regular containers and infra container.
+	// This ensures both phases get a fair share of the deadline and prevents the second
+	// stopTimeoutFromContext() call from getting 0 seconds when the first phase consumes
+	// all remaining time.
+	totalTimeout := stopTimeoutFromContext(ctx)
+	containerTimeout := totalTimeout / 2
```
bitoku: I don't think dividing by two is fair because normal workload would take more time.
I think defining some seconds for infra shutdown (maybe 1 second?), and detract it from stopTimeoutFromContext would be better. we can divide by two if the stopTimeoutFromContext is less than the predefined value.

saschagrunert: Sounds good, I updated the implementation as suggested.

server/sandbox_stop_linux.go
```diff
@@ -62,7 +86,7 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error
...
 
 	podInfraContainer := sb.InfraContainer()
-	if err := s.stopContainer(ctx, podInfraContainer, stopTimeoutFromContext(ctx)); err != nil && !errors.Is(err, storage.ErrContainerUnknown) {
+	if err := s.stopContainer(ctx, podInfraContainer, infraTimeout); err != nil && !errors.Is(err, storage.ErrContainerUnknown) && !errors.Is(err, oci.ErrContainerStopped) {
```
bitoku: I think we can use `totalTimeout` here.

If the main containers stop earlier than containerTimeout, we can use the rest for infraContainer stop.
I saw some cases where infraTimeout took long. It may happen.

saschagrunert: Updated the implementation as suggested, let's see what the CI thinks about this.

bitoku:
/lgtm
thanks!

PR #9661, opened 2025-12-15
[1.32] container: filter thermal_throttle masked paths

saschagrunert:
#### What type of PR is this?

/kind cleanup

Filter thermal_throttle paths from kubelet to fix performance regression on high-CPU bare metal systems during mass pod deployments.

None
Refers to https://issues.redhat.com/browse/OCPBUGS-66238

```release-note
None
```

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -798,7 +798,8 @@ func (c *container) SpecSetPrivileges(ctx context.Context, securityContext *type
 
 	if !c.Privileged() {
 		if securityContext.MaskedPaths != nil {
-			for _, path := range securityContext.MaskedPaths {
+			filteredPaths := filterThermalThrottlePaths(securityContext.MaskedPaths)
+			for _, path := range filteredPaths {
 				specgen.AddLinuxMaskedPaths(path)
 			}
 		}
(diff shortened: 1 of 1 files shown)


oarribas:
This seems related to #9425, which is removing the `thermal_throttle` masked paths in 1.34.0 and newer.

saschagrunert:
/hold 

I have to catch-up with the current state.

saschagrunert:
/unhold

QE verified that this fix works.

@cri-o/cri-o-maintainers PTAL

bitoku:
@sohankunkerkar we fixed it by changing the way of tmpfs mount in crun, didn't we?

bitoku:
https://github.com/containers/crun/pull/1859
https://issues.redhat.com/browse/OCPBUGS-60605

We avoided removing the masked paths at that time because it's added for a security reason https://github.com/containers/common/commit/4c30da054cca77cc5f2f38b74391f2195a803892 .

saschagrunert:
> [containers/crun#1859](https://github.com/containers/crun/pull/1859) https://issues.redhat.com/browse/OCPBUGS-60605
> 
> We avoided removing the masked paths at that time because it's added for a security reason [containers/common@4c30da0](https://github.com/containers/common/commit/4c30da054cca77cc5f2f38b74391f2195a803892) .

Alright, do we already have an OpenShift patch release available containing this fix?

bitoku:
https://issues.redhat.com/browse/OCPBUGS-60605?focusedId=28105584&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28105584

> crun and container-selinux packages from the dependent bugs have been picked up in 4.19-9.6, 4.20-9.6, and 4.21-9.6 builds

saschagrunert:
@bitoku we would require this change when working with runc instead of crun.

bitoku:
@saschagrunert Thank you for working on this.

We need to discuss how we fix this issue.
The masking patch was a security measurement. We need to assess the risk first.
Also we were able to mitigate the issue with the crun patch. Can't we do in runc as well?
containerd should be affected by this change, how did they fix it?

Even if we went this way, we should patch the main branch first.

cc: @haircommander @sohankunkerkar @giuseppe @kolyshkin

PR #9663, opened 2025-12-15
Fix the bug where the ContainersStatuses.Image fields returned by the GetContainerEvents is nil

HirazawaUi:
#### What type of PR is this?

/kind bug

If `s.ContainerStatus()` returns a **NotFound** error, the element at the corresponding index in `containerStatuses` will be nil, because we initially initialized the list with all elements set to nil.

When a list containing nil pointers is serialized and sent to the client via grpc, the protocol handling mechanism typically does not allow _null_ values in the list.

To preserve the list structure, the serialization layer usually treats this nil pointer as a _default empty message_.

After the receiving end deserializes the data, this empty message will be converted into a non-nil `ContainerStatus` struct, but all fields of this struct will be in a zero value state.

*  `ContainerStatus` is a valid struct pointer.

*  In the `ContainerStatus` definition, `Image` is a pointer type. The zero value of a pointer is nil.

*  Ultimately, this causes kubelet to panic when directly accessing the value of `ContainerStatus.Image` due to a nil pointer.

Code path in kubelet that directly uses this field:
https://github.com/kubernetes/kubernetes/blob/c180d6762d7ac5059d9b50457cafb0d7f4cf74a9/pkg/kubelet/kuberuntime/kuberuntime_container.go#L715-L724

Test code:
```
func TestNilSerialization(t *testing.T) {
	statuses := []*ContainerStatus{
		{Image: &ImageSpec{Id: "image1"}},
		nil, // This is the nil pointer
		{Image: &ImageSpec{Id: "image3"}},
	}
	resp := &Response{Statuses: statuses}

	data, err := proto.Marshal(resp)
	if err != nil {
		t.Fatalf("Marshal failed: %v", err)
	}

	var received Response
	if err := proto.Unmarshal(data, &received); err != nil {
		t.Fatalf("Unmarshal failed: %v", err)
	}

	middle := received.Statuses[1]
	if middle == nil {
		t.Error("Expected middle status to be non-nil, but it is nil")
	} else {
		fmt.Printf("Middle status is non-nil: %+v\n", middle)
		if middle.Image != nil {
			t.Errorf("Expected Image to be nil, but got %+v", middle.Image)
		} else {
			fmt.Println("Image field is nil, as expected")
		}
	}
}
```
Output:
```
Middle status is non-nil: 
Image field is nil, as expected
```

It can be demonstrated that during serialization and deserialization, protobuf will convert nil pointer elements in a list into an empty/default value.

```release-note
Fix the bug where the `ContainersStatuses.Image` returned by the `GetContainerEvents` is nil.
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -933,9 +933,9 @@ func (s *Server) getContainerStatuses(ctx context.Context, sandboxUID string) ([
 		return []*types.ContainerStatus{}, err
 	}
 
-	containerStatuses := make([]*types.ContainerStatus, len(containers.GetContainers()))
+	containerStatuses := make([]*types.ContainerStatus, 0, len(containers.GetContainers()))
 
-	for i, cc := range containers.GetContainers() {
+	for _, cc := range containers.GetContainers() {
 		containerStatusRequest := &types.ContainerStatusRequest{ContainerId: cc.GetId()}
 
 		resp, err := s.ContainerStatus(ctx, containerStatusRequest)
@@ -947,7 +947,7 @@ func (s *Server) getContainerStatuses(ctx context.Context, sandboxUID string) ([
 			return []*types.ContainerStatus{}, err
 		}
 
-		containerStatuses[i] = resp.GetStatus()
+		containerStatuses = append(containerStatuses, resp.GetStatus())
 	}
 
 	return containerStatuses, nil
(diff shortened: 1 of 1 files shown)


haircommander:
/ok-to-test
/approve

bitoku:
Thanks!
/lgtm

HirazawaUi:
These tests appear to be quite fragile. They might have been corrupted for some reason and are unable to pass, which is unrelated to the changes in this PR.

haircommander:
/retest
should be fixed

PR #9680, opened 2025-12-22
Remove unused memorystore.Size()

ankit98040:
#### What type of PR is this?

/kind cleanup

This PR optimizes the `Size()` method in `internal/memorystore` from O(N) to O(1) complexity by maintaining an atomic counter instead of iterating over the entire `sync.Map` on every call.

**Problem**: The previous implementation iterated over all entries in the map to count them, which becomes inefficient as the number of containers/sandboxes grows.

**Solution**: 
- Added an `atomic.Int64` counter to track the number of elements in the store
- Updated `Add()` to use `Swap()` and increment the counter only for new keys
- Updated `Delete()` to use `LoadAndDelete()` and decrement the counter only when a key is actually removed
- Changed `Size()` to simply return the atomic counter value

This improves performance for operations that frequently check the store size, especially in environments with many containers.

None

The changes are thread-safe as they use atomic operations and the existing `sync.Map` guarantees. The logic ensures:
- Updating an existing key (via `Add`) does not increment the counter
- Deleting a non-existent key does not decrement the counter
- All operations remain concurrent-safe

I verified the correctness with a standalone test that validates Add, Update, and Delete operations maintain accurate counts.

```release-note
None
```

## Summary by CodeRabbit

* **Refactor**
  * Removed an internal store-size query from the public interface; internal storage behavior otherwise unchanged.
* **Tests**
  * Updated unit tests to stop relying on the removed size query and to assert expected list results directly.

### DIFF
--- a/internal/memorystore/memory_store.go
+++ b/internal/memorystore/memory_store.go
@@ -21,8 +21,6 @@ type Storer[T AnyCreated[T]] interface {
 	Delete(string)
 	// List returns a list of values from the store.
 	List() []T
-	// Size returns the number of values in the store.
-	Size() int
 	// First returns the first value found in the store by a given filter.
 	First(StoreFilter[T]) T
 	// ApplyAll calls the reducer function with every value in the store.
@@ -75,15 +73,6 @@ func (c *memoryStore[T]) List() []T {
 	values.sort()
 
 	return values
-}
-
-// Size returns the number of values in the store.
-func (c *memoryStore[T]) Size() (l int) {
-	for range c.s.Range {
-		l++
-	}
-
-	return l
 }
 
 // First returns the first value found in the store by a given filter.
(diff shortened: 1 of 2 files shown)


bitoku:
Thank you for the PR @ankit98040 !

In fact, Size() is only used in this test here. (correct me if I'm wrong)
https://github.com/cri-o/cri-o/blob/fbbc7bfd5aa4b2a02312144dbe366a7bbd073199/internal/memorystore/memory_store_test.go#L51-L66

So this optimization should degrade the performance unfortunately because of unnecessary size management though It should be negligible.
I'd rather remove `Size()` instead, from the test too, not to use this inefficient Size() by mistake. wdyt?

bitoku:
/lgtm
Thanks!

PR #9683, opened 2025-12-22
Fix inconsistent imageRef when there are two images that are identical but have different names

bitoku:
#### What type of PR is this?

/kind bug

Currently when image status is requested, imageRef is just a first element of repoDigests returned from images package.
When there are some images that are identical but have different names, the name of imageRef sometimes doesn't match the name of the userRequested image.
This PR prioritizes the repoDigest that has the same name as userRequested, so that imageRef has a consistent name with userRequestedImage.

Partially addressed https://issues.redhat.com/browse/OCPBUGS-69675

```release-note
Fix inconsistent imageRef when there are two images that are identical but have different names
```

## Summary by CodeRabbit

* **New Features**
  * Enhanced image digest selection with repository matching logic for improved container creation.
  * Added support for mirrored image repositories in CI/CD pipeline.

* **Tests**
  * Added comprehensive test coverage for image digest selection and repository matching scenarios.
  * Added new tests validating correct image repository reference when using mirrored images.

* **Chores**
  * Updated build orchestration with new image artifact push targets.

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
-	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -21,6 +20,7 @@ import (
 	"github.com/opencontainers/runtime-tools/generate"
 	"go.podman.io/common/pkg/subscriptions"
 	"go.podman.io/common/pkg/timezone"
+	"go.podman.io/image/v5/docker/reference"
 	cstorage "go.podman.io/storage"
 	"go.podman.io/storage/pkg/idtools"
 	"go.podman.io/storage/pkg/mount"
@@ -1305,6 +1305,10 @@ func (s *Server) resolveAndVerifyContainerImage(ctx context.Context, ctr contain
 		}
 	}
 
+	if imgResult == nil {
+		return nil, fmt.Errorf("failed to find image %q", userRequestedImage)
+	}
+
 	if userRequestedImage == "" {
 		return nil, errors.New("internal error: successfully found an image, but userRequestedImage is empty")
 	}
@@ -1312,15 +1316,7 @@ func (s *Server) resolveAndVerifyContainerImage(ctx context.Context, ctr contain
 	someNameOfTheImage := imgResult.SomeNameOfThisImage
 	imageID := imgResult.ID
 
-	// For ImageRef, prefer user-requested image if it's a digest in RepoDigests.
-	// This ensures manifest list digests appear in ImageRef (not platform-specific ones).
-	someRepoDigest := ""
-	if len(imgResult.RepoDigests) > 0 {
-		someRepoDigest = imgResult.RepoDigests[0]
-		if slices.Contains(imgResult.RepoDigests, userRequestedImage) {
-			someRepoDigest = userRequestedImage
-		}
-	}
+	someRepoDigest := FindRepoDigestForImage(imgResult.RepoDigests, userRequestedImage)
 
 	if err := s.verifyImageSignature(ctx, sb.Metadata().GetNamespace(), ctr.Config().GetImage().GetUserSpecifiedImage(), imgResult); err != nil {
 		return nil, err
@@ -1333,6 +1329,43 @@ func (s *Server) resolveAndVerifyContainerImage(ctx context.Context, ctr contain
 		imageID:            imageID,
 		someRepoDigest:     someRepoDigest,
 	}, nil
+}
+
+// FindRepoDigestForImage finds an appropriate repo digest to use for imageRef.
+// Priority:
+// 1. If userRequestedImage exactly matches a repo digest, return it
+// 2. Otherwise, find a digest with matching repository name
+// 3. If no match is found, returns the first repo digest
+// Returns empty string if repoDigests is empty.
+func FindRepoDigestForImage(repoDigests []reference.Canonical, userRequestedImage string) string {
+	if len(repoDigests) == 0 {
+		return ""
+	}
+
+	// Default to the first repo digest
+	result := repoDigests[0].String()
+
+	// First, check for exact match (e.g., user requested a manifest list digest)
+	for _, d := range repoDigests {
+		if d.String() == userRequestedImage {
+			return userRequestedImage
+		}
+	}
+
+	// Try to find a digest with matching repository name
+	userRef, err := reference.ParseNormalizedNamed(userRequestedImage)
+	if err != nil {
+		return result
+	}
+
+	userRepo := reference.TrimNamed(userRef).String()
+	for _, d := range repoDigests {
+		if reference.TrimNamed(d).String() == userRepo {
+			return d.String()
+		}
+	}
+
+	return result
 }
 
 func setupWorkingDirectory(rootfs, mountLabel, containerCwd string) error {
--- a/Makefile
+++ b/Makefile
@@ -597,3 +597,12 @@ upload-artifacts: ## Upload the built artifacts.
 .PHONY: push-oci-artifacts
 push-oci-artifacts: ## Push OCI Artifacts to quay.io/crio
 	./test/testdata/artifacts/push-oci-artifacts
+
+.PHONY: push-mirror-image
+push-mirror-image: ## Push mirror image of fedora-crio-ci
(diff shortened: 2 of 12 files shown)


akalenyu (commented):
looks good, has this always existed? just a heads up in case somebody was relying on this behavior as API (though that would be surprising)

did we just accidentally start making deterministic images in kubevirt and thus started hitting it?
xref https://github.com/kubevirt/kubevirt/pull/16336

bitoku:
@akalenyu 
> has this always existed?

Yes, I don't know since when, but it should have had this behavior for some time.

> just a heads up in case somebody was relying on this behavior as API (though that would be surprising)

This is an old API and I'm not sure what's the exact usecase.
I guess because it's not well-defined, some implementation have emerged, and the initial object became obfuscated.
There's upstream discussion but not have been settled yet.

- https://github.com/kubernetes/kubernetes/issues/115199

This will break the current behavior, but it's not guaranteed to have consistent value, so changing the value shouldn't be a big deal, and I think it's beneficial to have a predictable value to avoid confusion in the future though I'm not sure if we should backport it to old versions. (maybe we shouldn't)

bitoku:
@cri-o/cri-o-maintainers PTAL

internal/oci/container.go
```diff
@@ -421,6 +421,10 @@ func (c *Container) CrioAnnotations() map[string]string {
 // UserRequestedImage returns the users' input originally used to find imageID; it might evaluate to a different image
 // (or to a different kind of reference!) at any future time.
 func (c *Container) UserRequestedImage() string {
+	if userSpecifiedImage := c.criContainer.GetImage().GetUserSpecifiedImage(); userSpecifiedImage != "" {
```
mtrmac: Looking at `resolveAndVerifyContainerImage`, this causes requests to create a container by ID to be handled by _ignoring_ that ID and creating it by doing a repeated lookup by `UserSpecifiedImage`, potentially finding a _different _image than that ID.

I think ignoring an explicit user input and doing something else is completely undesirable.

Similarly, internally, taking the `UserRequestedImage` API with a specific documented semantics, and just returning a different value than documented, also seems like a very bad idea for longer-term maintenance of the project. We can always add one more method, at ~trivial cost; making an existing method not do what is documented is not better.

---

Same for the `internal/factory` function.

server/container_create.go
```diff
@@ -1301,22 +1305,18 @@ func (s *Server) resolveAndVerifyContainerImage(ctx context.Context, ctr contain
 		}
 	}
 
+	if imgResult == nil {
```
mtrmac, line 1308: Why should this ever trigger? The lookups of `imgResult` should have returned a specific error.

mtrmac (changes requested):
I don’t have _much_ of an opinion on changing the image properties returned in `ContainerStatus` (_if_ that’s the API this PR wants to actually change — the PR description does not explicitly say).

But looking at how this affects the (pull/don’t pull +) create container code path, I think that’s a clearly undesirable change, AFAICS really unrelated, and _I guess_ unnecessary — if the goal is to record a value to move it from `ContainerCreate` to `ContainerStatus`, that should always be possible by adding a new variable/field/method with a clear semantics (and then perhaps removing any no-longer-used fields / method), without touching the core input→ID lookup  logic.
---

(As a reviewer, it would be easier for me if refactoring and behavior changes were split into separate commits: I _think_ the move of `BuildRepoDigests` into a separate function does not change behavior but it’s not clear what the intent was.)

internal/oci/container.go
bitoku: > Similarly, internally, taking the UserRequestedImage API with a specific documented semantics, and just returning a different value than documented, also seems like a very bad idea for longer-term maintenance of the project. We can always add one more method, at ~trivial cost; making an existing method not do what is documented is not better.

I intended to fix the behavior as documented.
iiuc spec.Image is a resolved image ID, and spec.UserSpecifiedImage is what the user originally input, isn't it?
UserRequestedImage function returning imageID, not UserSpecifiedImage is more confusing to me.

I understand the behavioral concern though.

server/container_create.go
bitoku: I added because my IDE's static check complained about it.
If potentialMatches is empty, it could be nil.
potentialMatches may not be nil unless err is nil, though it's not that obvious and I think it's good to have nil guard.

internal/oci/container.go
mtrmac: > iiuc spec.Image is a resolved image ID, and spec.UserSpecifiedImage is what the user originally input, isn't it?

`ImageStatus(~userInput)` -> image present with `imageID`: `CreateContainer(userSpecifiedImage: ~userInput, Image: imageID)`

`ImageStatus(userInput)` -> image not present: `PullImage(~userInput) -> repo@digest, CreateContainer(userSpecifiedImage: ~userInput, Image: repo@digest)`

(I’m being rather vague about the precise values because also short name lookups inside CRI-O, and Kubelet’s normalization of short names to `docker.io`, _at least_ for credential lookups, are involved, and I don’t remember the exact details.)

For the purposes of this `resolveAndVerifyContainerImage` operation during `ContainerCreate`, we’d really prefer to always have a repo@theRightDigest (because signatures are relevant for specific digests, not for all manifests that might be deduplicated into a single image ID), but that’s not what Kubelet provides; and Kubelet’s image removal code needs IDs to match, IIRC.

I also have ~no idea about what the use cases / assumptions are in non-Kubelet callers, e.g. consumers of Pod status.

server/container_create.go
mtrmac: `CandidatesForPotentiallyShortImageName` never succeeds with a zero-length result.

I’ll let actual CRI-O maintainers decide on the preferred style here.

internal/oci/container.go
bitoku: yeah I understand the behavioral change is undesirable.

I just wanted to confirm the intention of your comment.

> taking the UserRequestedImage API with a specific documented semantics, and just returning a different value than documented, also seems like a very bad idea for longer-term maintenance of the project. 

According to this doc string, it's natural to return UserSpecifiedImage, not ImageID.

```
// UserRequestedImage returns the users' input originally used to find imageID; it might evaluate to a different image
// (or to a different kind of reference!) at any future time.
```

mtrmac: No, not in the case of “image present with `imageID`”. In that case “the input user specified” is `userInput`, but “the input used to find `imageID`” is `imageID`.

I have fairly week opinions on what data should be recorded for individual containers at all, and I know fairly little about how the container/pod data is used within Kubernetes.

But I feel very strongly that if Kubelet says `createContainer{Image: imageID}`, the lookup _must_ be by `imageID`. (If Kubelet wanted to do a lookup by `userInput`, it can always call `createContainer{Image: userInput}`.)

mtrmac: … there’s certainly an opportunity to document things better. Naming concepts is hard.

bitoku: Thank you for the insight anyway.
I removed the changes regarding UserRequestedImage.

server/container_create.go
```diff
@@ -1331,6 +1331,43 @@ func (s *Server) resolveAndVerifyContainerImage(ctx context.Context, ctr contain
...
+	result := repoDigests[0].String()
+
+	// First, check for exact match (e.g., user requested a manifest list digest)
+	for _, d := range repoDigests {
+		if d.String() == userRequestedImage {
+			return userRequestedImage
+		}
+	}
+
+	// Try to find a digest with matching repository name
```
mtrmac, line 1355: This is something, but it’s not _reliable_ for the “image already found, create container by `imageID` case”: it is possible for various representations of the same image to be uploaded to a single repo; if two such representations are pulled, they will be deduplicated and both show up in RepoDigests, and there’s no way to tell which digest is “correct” at this point.

(That also affects signature validation, because signatures are attached to digests, and it‘s pre-existing in principle, but my point is that after this PR, the callers can _still_ not rely on the repo@digest in container status for exact matching, except for specially constrained situations.)

bitoku: hmm fair enough.
It's the problem of the nature of ImageID and the lack of the document for it, but I think it still reduces some confusions regarding it.

I'd like to hear what others think. @cri-o/cri-o-maintainers

bitoku:
I couldn't get much traction to it, and also there's no reliable way. I'm closing this.

PR #9692, opened 2026-01-02
checkpoint: clean up checkpoint dir on error

rst0git:
#### What type of PR is this?

/kind bug

CRIU might fail when options such as "--tcp-established" have not been specified and the container has established TCP connections. In such cases, the "checkpoint" directory is leftover with CRIU images that were generated before the error was detected. These CRIU images are then included in subsequent checkpoints and may cause unexpected errors or incorrect results during forensic analysis. To address this problem, we need to clean up the checkpoint directory in case of an error. 

None

This change is similar to https://github.com/cri-o/cri-o/pull/7355.

No

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved checkpoint error handling: when a checkpoint attempt fails, the system now automatically tries to remove leftover checkpoint images and logs a warning if cleanup fails; the original checkpoint error is still returned.

### DIFF
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -93,6 +93,11 @@ func (c *ContainerServer) ContainerCheckpoint(
 	}
 
 	if err := c.runtime.CheckpointContainer(ctx, ctr, specgen.Config, opts.KeepRunning); err != nil {
+		// in the case of an error, clean up any leftover CRIU images
+		if err := os.RemoveAll(ctr.CheckpointPath()); err != nil {
+			log.Warnf(ctx, "Unable to remove checkpoint directory %s: %v", ctr.CheckpointPath(), err)
+		}
+
 		return "", fmt.Errorf("failed to checkpoint container %s: %w", ctr.ID(), err)
 	}
 


adrianreber (approved):

haircommander:
/approve

If the checkpoint succeeds, does the path stay? do we need to clean up in any other scnarios/ failures?

rst0git:
>If the checkpoint succeeds, does the path stay? do we need to clean up in any other scnarios/ failures?

We addressed this in https://github.com/cri-o/cri-o/pull/7355 - the folder with CRIU images is removed after `exportCheckpoint()`, otherwise the tar archive would contain files from previous checkpoints and increase in size after each checkpoint.

internal/lib/checkpoint.go
```diff
@@ -93,6 +93,11 @@ func (c *ContainerServer) ContainerCheckpoint(
 	}
 
 	if err := c.runtime.CheckpointContainer(ctx, ctr, specgen.Config, opts.KeepRunning); err != nil {
+		// in the case of an error, clean up any leftover CRIU images
+		if err := os.RemoveAll(ctr.CheckpointPath()); err != nil {
```
bitoku, line 97: non blocking: This removal should reside in `CheckpointContainer` because it should be always cleaned up when `CheckpointContainer` fails.
There's no other this function call though.

Issue #9583, opened 2025-11-11
Speed up AI model-loading by having a pre-download mechanism for OCI volumes


saschagrunert:
## Overview

Add a mechanism to pre-pull OCI artifacts (especially AI/ML models) before they are needed by workloads, enabling faster pod startup for latency-sensitive inference services.

## Problem Statement

**Current behavior**: When a pod requests an OCI artifact volume, CRI-O downloads it on-demand. For large AI models (10GB+), this causes significant cold-start delays during autoscaling events.

**User impact**:
- LLM/GenAI workloads are latency-sensitive with spiky traffic patterns
- Autoscalers (KEDA, HPA) cannot respond quickly due to model download delays
- Pod startup times of 3-5 minutes are unacceptable for inference services
- Time between scaling decisions and ability to serve requests is critical for maintaining SLAs

### 1. KEDA-based AI Inference Autoscaling
**Scenario**: LLM inference services autoscale based on queue depth. When traffic spikes, new pods must come online quickly.

**Problem**: New pods wait 3-5 minutes downloading 10GB models. By the time pods are ready, the spike may have passed or users have experienced degraded latency.

**Desired outcome**: Pods start in <30 seconds with pre-cached models, enabling system to respond quickly to load changes.

### 2. Spot/Preemptible Instance Preparation
**Scenario**: Cloud clusters use spot instances for cost savings. New nodes need to be ready immediately to handle workloads.

**Problem**: New spot instance joins, pod is scheduled, downloads model for 5 minutes. If instance is preempted during download, work is lost.

**Desired outcome**: Node initialization pre-pulls critical artifacts before marking node ready. Pods start with predictable latency.

### 3. Multi-Model Serving Platforms
**Scenario**: Model serving platforms support hundreds of models, but only a subset are "hot" at any time.

**Desired outcome**: Pre-pull top 20 most-requested models on all nodes. 80% of requests served with zero download latency, graceful degradation for rare models.

### 4. Air-gapped/Edge Deployments
**Scenario**: Edge locations or air-gapped environments cannot pull artifacts on-demand due to network constraints.

**Desired outcome**: Pre-pull all required artifacts during node provisioning or via offline bundle/USB drive. Nodes operate without registry access.

## Related

- #9570: Configurable artifact storage locations (complementary feature)
- **Combined use case**: Pre-pull models into shared storage, configure as read-only artifact store on all nodes

aviralgarg05:
I would like to solve this issue

[...]

PR #9691, opened 2026-01-02
OCPNODE-3945: Adapt ociartifact to new interfaces of libartifact

bitoku:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This PR fixes cri-o code to adapt it to new interfaces of libartifact store.
https://github.com/containers/container-libs/commit/539435a8c6cc6b36320ca2461d683fcc61e4a73a

#### Does this PR introduce a user-facing change?

```release-note
None
```

## Summary by CodeRabbit

- New Features
  - Added an OCI artifact data store with data-pulling capabilities, per-layer size limits, digest verification, and improved lookup by name or digest.
- Refactor
  - Simplified artifact store integration and streamlined artifact handling and references across the codebase.
- Chores
  - Upgraded numerous dependencies, including CRIU to v8, runc, SELinux, and various Go modules for improved compatibility and security.
- Tests
  - Added new datastore tests; updated mocks to align with the refactored datastore and dependency changes; removed obsolete tests.

PR #9702, opened 2026-01-08
Add support for additional read-only artifact stores

### LINKED ISSUE #9570: Default artifact location
## Overview

Add configuration options to specify custom storage locations for OCI artifacts in CRI-O, similar to how `additionalimagestores` works in containers/storage.

## Current Behavior

- **Default path**: `<graphroot>/artifacts/` (typically `/var/lib/containers/storage/artifacts/`)
- **Implementation**: `internal/ociartifact/store.go:47` - hardcoded as `filepath.Join(rootPath, "artifacts")`
- **Limitation**: No way to use separate storage or pre-populated read-only artifact stores

### 1. Separate Storage Devices
Store large ML models on dedicated high-capacity storage separate from OS/container images.
- Prevents artifacts from filling root filesystem
- Allows specialized storage (HDD for artifacts, SSD for images)

### 2. Pre-populated Artifact Cache
Pre-populate shared read-only artifacts across cluster nodes (via NFS/distributed filesystem).
- Reduces network bandwidth and registry load
- Faster pod startup (no download required)
- Ideal for air-gapped environments

### 3. Multi-tier Storage
Fast local storage for hot artifacts, slower network storage for archives.
- Performance optimization for frequently accessed artifacts
- Cost-effective storage tiering

### 4. Immutable Infrastructure
Nodes use baked-in or verified artifact sources, preventing dynamic downloads.
- Enhanced security (no external registry access)
- Compliance with regulatory requirements

### 5. Edge/IoT Deployments
Limited local storage with read-only artifacts on removable media.
- Minimizes local storage requirements
- Offline artifact delivery via USB/SD cards

## CRI-O Configuration

Based on the OpenShift enhancement (openshift/enhancements#1934) and API (openshift/api#2681), the CRI-O configuration field is:

### Config field

```toml
[crio.runtime]
additional_artifact_stores = [
  "/mnt/ssd-artifacts",
  "/mnt/nfs-artifacts"
]
```

- **Field name**: `additional_artifact_stores`
- **Section**: `[crio.runtime]`
- **Type**: `[]string` (list of absolute paths)
- **Semantics**: Read-only stores, checked in order before the default location (`<graphroot>/artifacts/`)
- **Default**: empty (only `<graphroot>/artifacts/` is used)

### OpenShift Integration

On OpenShift, this is configured via `ContainerRuntimeConfig`:

```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: ContainerRuntimeConfig
metadata:
  name: set-artifact-stores
spec:
  machineConfigPoolSelector:
    matchLabels:
      pools.operator.machineconfiguration.openshift.io/worker: ""
  containerRuntimeConfig:
    additionalArtifactStores:
      - path: /mnt/ssd-artifacts
      - path: /mnt/nfs-artifacts
```

The MCO generates a CRI-O drop-in file at `/etc/crio/crio.conf.d/01-ctrcfg-additionalArtifactStores`.

Feature-gated behind `AdditionalStorageConfig` (DevPreviewNoUpgrade/TechPreviewNoUpgrade).

### Implementation notes

- Paths must exist on the node before config is applied
- Stores are read-only from CRI-O's perspective
- Lookup order: additional stores (in list order) → default store
- Graceful fallback: if an additional store path doesn't exist or is inaccessible, CRI-O should log a warning and continue with remaining stores

## Related

- #9583: Pre-pull mechanism for OCI artifacts (complementary feature)
- #9702: Implementation PR
- Existing pattern: `additionalimagestores` in containers/storage
- https://issues.redhat.com/browse/OCPSTRAT-2623
- openshift/enhancements#1934: Enhancement proposal
- openshift/api#2681: API changes
- openshift/machine-config-operator#5666: MCO implementation

pauloappbr:
#### What type of PR is this?

/kind feature

Adds support for additional read-only OCI artifact stores via the `additional_artifact_stores` configuration option and `--additional-artifact-stores` CLI flag. This allows pre-populated artifacts (e.g. large AI/ML models on NFS) to be used without re-downloading them into the main store.

Artifacts in additional stores take priority over the main store. Pulls are skipped when an artifact already exists in any store. Tag re-pointing is not supported for read-only stores.

Fixes #9570
Related to #9583

```release-note
Added support for configuring additional read-only artifact stores via the `additional_artifact_stores` configuration option.
```

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -131,6 +131,9 @@ If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
 **decryption_keys_path**="/etc/crio/keys/"
 Path where the keys required for image decryption are located
 
+**additional_artifact_stores**=[]
+A list of additional read-only OCI artifact store paths (experimental, subject to change). CRI-O expects an "artifacts/" subdirectory within each configured path. All entries must be absolute paths. Artifacts in these stores take priority over the main store. Because these stores are read-only, CRI-O cannot remove artifacts from them. If a tag is re-pointed on the registry, the stale local copy in a read-only store will continue to be used; the artifact must be removed from the read-only store directly on the filesystem to pick up the new version.
+
 **conmon**=""
 Path to the conmon binary, used for monitoring the OCI runtime. Will be searched for using $PATH if empty.
 This option is currently deprecated, and will be replaced with RuntimeHandler.MonitorPath.
--- a/internal/ociartifact/datastore/store.go
+++ b/internal/ociartifact/datastore/store.go
@@ -42,11 +42,11 @@ type Store struct {
 }
 
 // New creates a new OCI artifact data store.
-// Note: The datastore currently only handles artifacts pulled into the main store.
-// Additional read-only stores are not threaded through here (we pass nil for
-// additionalPaths) since the datastore is primarily used for in-memory artifact
-// data managed by the main CRI-O lifecycle.
 func New(rootPath string, systemContext *types.SystemContext) (*Store, error) {
+	// The datastore only handles artifacts pulled into the main store.
+	// Additional read-only stores are not threaded through here (we pass
+	// nil for additionalPaths) since the datastore is used for in-memory
+	// artifact data managed by the main CRI-O lifecycle.
 	ociStore, err := ociartifact.NewStore(rootPath, nil, systemContext)
 	if err != nil {
 		return nil, fmt.Errorf("create OCI artifact store: %w", err)
--- a/internal/ociartifact/store.go
+++ b/internal/ociartifact/store.go
@@ -44,6 +44,12 @@ var configMediaTypes = []string{
 	manifest.DockerV2Schema2ConfigMediaType,
 }
 
+// additionalStore pairs a read-only artifact store with its path.
+type additionalStore struct {
+	path  string
+	store LibartifactStore
+}
+
 // Store is the main structure to build an artifact storage.
 type Store struct {
 	libartifactStore LibartifactStore
@@ -51,8 +57,7 @@ type Store struct {
 
 	// rootPath is required for BlobMountPaths.
 	rootPath         string
-	additionalPaths  []string
-	additionalStores []LibartifactStore
+	additionalStores []additionalStore
 }
 
 // NewStore creates a new OCI artifact store.
@@ -65,8 +70,7 @@ func NewStore(rootPath string, additionalPaths []string, systemContext *types.Sy
 	}
 
 	// Configure additional stores (RO)
-	var additionalStores []LibartifactStore
-	var validAdditionalPaths []string
+	var additional []additionalStore
 
 	for _, path := range additionalPaths {
 		addPath := filepath.Join(path, "artifacts")
@@ -74,18 +78,21 @@ func NewStore(rootPath string, additionalPaths []string, systemContext *types.Sy
 		addStore, err := libart.NewArtifactStore(addPath, systemContext)
 		if err != nil {
 			log.Warnf(context.Background(), "Skipping additional artifact store %q: %v", addPath, err)
+
 			continue
 		}
-		additionalStores = append(additionalStores, &artifactStore{addStore})
-		validAdditionalPaths = append(validAdditionalPaths, addPath)
+
+		additional = append(additional, additionalStore{
+			path:  addPath,
+			store: &artifactStore{addStore},
+		})
 	}
 
 	return &Store{
 		libartifactStore: &artifactStore{store},
 		impl:             &defaultImpl{},
 		rootPath:         storePath,
-		additionalPaths:  validAdditionalPaths,
-		additionalStores: additionalStores,
+		additionalStores: additional,
 	}, nil
 }
 
@@ -104,14 +111,22 @@ func (s *Store) Pull(
 
 	strRef := ref.DockerReference().String()
 
-	// Skip pulling if the artifact already exists in a read-only additional store.
-	// Note: Because we short-circuit here, tag re-pointing for artifacts present
-	// in additional stores is not supported. If a user needs to force-pull a new
-	// digest for a tag, the artifact must not exist in the read-only store.
-	if art, err := s.Status(ctx, strRef); err == nil && art.RootPath() != s.rootPath {
-		log.Infof(ctx, "Artifact %s found locally in additional read-only store, skipping pull", strRef)
-		dgst := art.Digest()
-		return &dgst, nil
+	// Skip pulling if the artifact already exists in a read-only additional
+	// store. This avoids duplicating artifacts that are already available.
+	// We intentionally do NOT skip for artifacts in the main store, because
+	// the remote tag may have been re-pointed to a different digest.
+	if len(s.additionalStores) > 0 {
+		artRef, err := libart.NewArtifactStorageReference(strRef)
+		if err == nil {
+			for _, add := range s.additionalStores {
+				if art, err := add.store.Inspect(ctx, artRef); err == nil {
+					log.Infof(ctx, "Artifact %s already exists in additional store %s, skipping pull", strRef, add.path)
+					dgst := NewArtifact(art, add.path).Digest()
+
+					return &dgst, nil
+				}
+			}
+		}
 	}
 
 	log.Infof(ctx, "Pulling OCI artifact %s", strRef)
@@ -205,14 +220,16 @@ func (s *Store) List(ctx context.Context) (res []*Artifact, err error) {
 	// We warn and continue on errors here because additional stores may
 	// reside on unreliable media (like NFS) and shouldn't block listing
 	// artifacts from the main store.
-	for i, addStore := range s.additionalStores {
-		addArts, err := addStore.List(ctx)
+	for _, add := range s.additionalStores {
+		addArts, err := add.store.List(ctx)
 		if err != nil {
-			log.Warnf(ctx, "Failed to list artifacts from additional store %q: %v", s.additionalPaths[i], err)
+			log.Warnf(ctx, "Failed to list artifacts from additional store %q: %v", add.path, err)
+
 			continue
 		}
+
 		for _, art := range addArts {
-			arts = append(arts, NewArtifact(art, s.additionalPaths[i]))
+			arts = append(arts, NewArtifact(art, add.path))
 		}
 	}
 
@@ -221,20 +238,24 @@ func (s *Store) List(ctx context.Context) (res []*Artifact, err error) {
 	if err != nil {
 		return nil, fmt.Errorf("list artifacts from main store: %w", err)
 	}
+
 	for _, art := range mainArts {
 		arts = append(arts, NewArtifact(art, s.rootPath))
 	}
 
-	// Deduplicate preserving priority (uses both Reference and Digest)
-	// We include the reference in the key so that the same blob with
-	// different tags (e.g., :v1 and :latest) will correctly appear twice.
+	// Deduplicate by reference, preserving priority (additional stores
+	// are listed first, so they win). This ensures that a tag which
+	// exists in both an additional and the main store only appears once.
 	seen := make(map[string]struct{})
+
 	for _, artifact := range arts {
-		key := fmt.Sprintf("%s@%s", artifact.Reference(), artifact.Digest().Encoded())
+		key := artifact.Reference()
 		if _, ok := seen[key]; ok {
 			continue
 		}
+
 		seen[key] = struct{}{}
+
 		res = append(res, artifact)
 	}
 
@@ -250,11 +271,16 @@ func (s *Store) Status(ctx context.Context, nameOrDigest string) (*Artifact, err
 	}
 
 	// Check additional stores first (Prioritized)
-	for i, store := range s.additionalStores {
-		if artifact, err := store.Inspect(ctx, artRef); err == nil {
-			return NewArtifact(artifact, s.additionalPaths[i]), nil
+	for _, add := range s.additionalStores {
+		artifact, err := add.store.Inspect(ctx, artRef)
+		if err == nil {
+			return NewArtifact(artifact, add.path), nil
+		}
+
+		if errors.Is(err, ErrNotFound) {
+			log.Debugf(ctx, "Artifact not found in additional store %q", add.path)
 		} else {
-			log.Debugf(ctx, "Inspect in additional store %q failed: %v", s.additionalPaths[i], err)
+			log.Warnf(ctx, "Failed to inspect artifact in additional store %q: %v", add.path, err)
 		}
 	}
 
@@ -311,6 +337,7 @@ func (s *Store) BlobMountPaths(ctx context.Context, artifact *Artifact, sys *typ
 		name := artifactName(l.Annotations)
 		if name == "" {
 			log.Warnf(ctx, "Unable to find name for artifact layer which makes it not mountable")
+
 			continue
 		}
 
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -120,6 +120,10 @@ type ImageCopyOptions struct {
 	ProgressInterval time.Duration
 	Progress         chan types.ProgressProperties `json:"-"`
 	CgroupPull       CgroupPullConfiguration
+
+	// AdditionalArtifactStores is a list of paths to additional read-only
+	// artifact stores. Used in the OCI artifact fallback pull path.
+	AdditionalArtifactStores []string
 }
 
 // ImageServer wraps up various CRI-related activities into a reusable
@@ -869,10 +873,7 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 	if shouldTryArtifact(err) {
 		log.Infof(ctx, "Falling back to pull %s as an OCI artifact: %v", imageName, err)
 
-		// Note: We pass nil for additionalPaths here. Since this fallback path
-		// is only hit when a regular container image pull fails, we do not currently
-		// thread the additional read-only artifact stores through the image service.
-		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), nil, &srcSystemContext)
+		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), options.AdditionalArtifactStores, &srcSystemContext)
 		if artifactErr != nil {
 			return RegistryImageReference{}, fmt.Errorf("unable to pull image or OCI artifact: create store err: %w", artifactErr)
 		}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -428,7 +428,7 @@ type RuntimeConfig struct {
 	// Note that CRI-O expects an "artifacts/" subdirectory within each configured
 	// path (mirroring the main store convention). For example, if configured with
 	// "/mnt/nfs", the artifacts should be placed in "/mnt/nfs/artifacts/".
-	AdditionalArtifactStores []string `toml:"additional_artifact_stores,omitempty"`
+	AdditionalArtifactStores []string `toml:"additional_artifact_stores"`
 
 	// Conmon is the path to conmon binary, used for managing the runtime.
 	// This option is currently deprecated, and will be replaced with RuntimeHandler.MonitorConfig.Path.
@@ -1339,12 +1339,6 @@ func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution
 	for _, p := range c.AdditionalArtifactStores {
 		if !filepath.IsAbs(p) {
 			return fmt.Errorf("additional_artifact_stores entry must be absolute: %q", p)
-		}
-		if len(p) > 256 {
-			return fmt.Errorf("additional_artifact_stores entry cannot exceed 256 characters: %q", p)
-		}
-		if strings.Contains(p, "//") {
-			return fmt.Errorf("additional_artifact_stores entry cannot contain consecutive slashes: %q", p)
 		}
 	}
 
--- a/test/artifact_additional_stores.bats
+++ b/test/artifact_additional_stores.bats
@@ -10,40 +10,148 @@ function teardown() {
 	cleanup_test
 }
 
+ARTIFACT_REPO=quay.io/crio/artifact
+ARTIFACT_IMAGE="$ARTIFACT_REPO:singlefile"
+
+# Helper: populate an additional store by pulling an artifact into the main
+# store, copying the OCI layout, and then removing it from the main store.
+# After this, the artifact only exists in the additional store.
+function populate_additional_store() {
+	local additional_store="$1"
+
+	# Pull artifact into the main store
+	crictl pull "$ARTIFACT_IMAGE"
+
+	# Copy the OCI artifact layout to the additional store
+	local main_store="$TESTDIR/crio/artifacts"
+	cp -a "$main_store" "$additional_store/"
+
+	# Remove from main store so it only exists in the additional store
+	crictl rmi "$ARTIFACT_IMAGE"
+}
+
 @test "should support valid additional_artifact_stores" {
-	# Create a mock additional store directory
 	ADDITIONAL_STORE="$TESTDIR/additional-store"
 	mkdir -p "$ADDITIONAL_STORE/artifacts"
 
-	# Configure CRI-O to use the additional store using a drop-in config
-	cat <<EOF > "$CRIO_CONFIG_DIR/99-artifact.conf"
+	cat << EOF > "$CRIO_CONFIG_DIR/99-artifact.conf"
 [crio.runtime]
 additional_artifact_stores = [
     "$ADDITIONAL_STORE"
 ]
 EOF
 
-	# Start CRI-O
 	start_crio
 
-	# Verify CRI-O started successfully and parsed the config
-	run crio-status config
-	assert_success
-	assert_output --partial 'additional_artifact_stores = ['
-	assert_output --partial "$ADDITIONAL_STORE"
+	run -0 "${CRIO_BINARY_PATH}" status --socket="${CRIO_SOCKET}" config
+	[[ "$output" == *'additional_artifact_stores = ['* ]]
+	[[ "$output" == *"$ADDITIONAL_STORE"* ]]
 }
 
 @test "should fail if additional_artifact_stores path is not absolute" {
-	# Configure CRI-O with a relative path
-	cat <<EOF > "$CRIO_CONFIG_DIR/99-artifact-invalid.conf"
+	cat << EOF > "$CRIO_CONFIG_DIR/99-artifact-invalid.conf"
 [crio.runtime]
 additional_artifact_stores = [
     "./relative/path/store"
 ]
 EOF
 
-	# Try to start CRI-O, it should fail due to our config validation
-	run "$CRIO_BINARY_PATH" --config-dir "$CRIO_CONFIG_DIR" config
-	assert_failure
-	assert_output --partial "additional_artifact_stores entry must be absolute: \"./relative/path/store\""
+	run -1 "$CRIO_BINARY_PATH" --config-dir "$CRIO_CONFIG_DIR" config
+	[[ "$output" == *'additional_artifact_stores entry must be absolute'* ]]
+	[[ "$output" == *'./relative/path/store'* ]]
+}
+
+@test "should list artifact from additional store" {
+	ADDITIONAL_STORE="$TESTDIR/additional-store"
+	mkdir -p "$ADDITIONAL_STORE"
+
+	cat << EOF > "$CRIO_CONFIG_DIR/99-artifact.conf"
+[crio.runtime]
+additional_artifact_stores = [
+    "$ADDITIONAL_STORE"
+]
+EOF
+
+	start_crio
+	populate_additional_store "$ADDITIONAL_STORE"
+
+	# The artifact should be visible from the additional store
+	crictl images | grep -qE "$ARTIFACT_REPO.*singlefile"
+}
+
+@test "should inspect artifact from additional store" {
+	ADDITIONAL_STORE="$TESTDIR/additional-store"
+	mkdir -p "$ADDITIONAL_STORE"
+
+	cat << EOF > "$CRIO_CONFIG_DIR/99-artifact.conf"
+[crio.runtime]
+additional_artifact_stores = [
+    "$ADDITIONAL_STORE"
+]
+EOF
+
+	start_crio
+	populate_additional_store "$ADDITIONAL_STORE"
+
+	crictl inspecti "$ARTIFACT_IMAGE" |
+		jq -e '
+		(.status.pinned == true) and
+		(.status.repoDigests | length == 1) and
+		(.status.repoTags | length == 1) and
+		(.status.size != "0")'
+}
+
+@test "should skip pull when artifact exists in additional store" {
+	ADDITIONAL_STORE="$TESTDIR/additional-store"
+	mkdir -p "$ADDITIONAL_STORE"
+
+	cat << EOF > "$CRIO_CONFIG_DIR/99-artifact.conf"
+[crio.runtime]
+additional_artifact_stores = [
+    "$ADDITIONAL_STORE"
+]
+EOF
+
+	start_crio
+	populate_additional_store "$ADDITIONAL_STORE"
+
+	# Pull again; should be skipped because it exists in the additional store
+	crictl pull "$ARTIFACT_IMAGE"
+
+	# Verify the artifact is NOT in the main store (pull was skipped)
+	local main_index="$TESTDIR/crio/artifacts/index.json"
+	run jq -r '.manifests[].annotations["org.opencontainers.image.ref.name"] // empty' "$main_index"
+	[[ "$output" != *"$ARTIFACT_IMAGE"* ]]
+
+	# Verify the log shows the skip message
+	grep -q "already exists in additional store" "$CRIO_LOG"
+}
+
+@test "should mount artifact from additional store" {
+	ADDITIONAL_STORE="$TESTDIR/additional-store"
+	mkdir -p "$ADDITIONAL_STORE"
+
+	cat << EOF > "$CRIO_CONFIG_DIR/99-artifact.conf"
+[crio.runtime]
+additional_artifact_stores = [
+    "$ADDITIONAL_STORE"
+]
+EOF
+
+	start_crio
+	populate_additional_store "$ADDITIONAL_STORE"
+
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	jq --arg ARTIFACT_IMAGE "$ARTIFACT_IMAGE" \
+		'.mounts = [ {
+      container_path: "/root/artifact",
+      image: { image: $ARTIFACT_IMAGE },
+    } ] |
+    .command = ["sleep", "3600"]' \
+		"$TESTDATA"/container_config.json > "$TESTDIR/container_config.json"
+	ctr_id=$(crictl create "$pod_id" "$TESTDIR/container_config.json" "$TESTDATA/sandbox_config.json")
+	crictl start "$ctr_id"
+
+	run crictl exec --sync "$ctr_id" cat /root/artifact/artifact.txt
+	[[ "$output" == "hello artifact" ]]
 }
--- a/completions/bash/crio
+++ b/completions/bash/crio
@@ -16,6 +16,7 @@ help
 h
 --absent-mount-sources-to-reject
 --add-inheritable-capabilities
+--additional-artifact-stores
 --additional-devices
 --address
 --allowed-devices
--- a/completions/fish/crio.fish
+++ b/completions/fish/crio.fish
@@ -11,6 +11,7 @@ end
 
 complete -c crio -n '__fish_crio_no_subcommand' -f -l absent-mount-sources-to-reject -r -d 'A list of paths that, when absent from the host, will cause a container creation to fail (as opposed to the current behavior of creating a directory).'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l add-inheritable-capabilities -d 'Add capabilities to the inheritable set, as well as the default group of permitted, bounding and effective.'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l additional-artifact-stores -r -d 'Additional read-only OCI artifact store paths.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l additional-devices -r -d 'Devices to add to the containers.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l allowed-devices -r -d 'Devices a user is allowed to specify with the "io.kubernetes.cri-o.Devices" allowed annotation.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l apparmor-profile -r -d 'Name of the apparmor profile to be used as the runtime\'s default. This only takes effect if the user does not specify a profile via the Kubernetes Pod\'s metadata annotation.'
--- a/completions/zsh/_crio
+++ b/completions/zsh/_crio
@@ -41,6 +41,7 @@ it later with **--config**. Global options will modify the output.'
   opts=(
         '--absent-mount-sources-to-reject'
         '--add-inheritable-capabilities'
+        '--additional-artifact-stores'
         '--additional-devices'
         '--address'
         '--allowed-devices'
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -13,6 +13,7 @@ crio
 ```
 [--absent-mount-sources-to-reject]=[value]
 [--add-inheritable-capabilities]
+[--additional-artifact-stores]=[value]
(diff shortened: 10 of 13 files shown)


pkg/config/config.go
```diff
@@ -422,6 +422,10 @@ type RuntimeConfig struct {
 	// DecryptionKeysPath is the path where keys for image decryption are stored.
 	DecryptionKeysPath string `toml:"decryption_keys_path"`
 
+	// AdditionalArtifactStores is a list of additional read-only artifact stores.
+	// Each path behaves like the main artifact store but is read-only.
+	AdditionalArtifactStores []string `toml:"additional_artifact_stores,omitempty"`
```
bitoku: I would prefer ReadOnlyArtifactStores or something like that.

bitoku (commented):
Thank you for the PR!

We're in the middle of migration from our implementation to libartifact in c/container-libs.
I think we should wait for a while until it's done to avoid rework.

pkg/config/config.go
pauloappbr: Hi @bitoku, thank you for the feedback!

I understand the situation regarding the migration to `libartifact` in `c/container-libs`. I will convert this PR to a Draft for now to avoid conflicts and rework. Once the migration is complete, please let me know, and I will rebase/update my implementation accordingly.

Regarding the naming suggestion: I agree that `ReadOnlyArtifactStores` is much clearer than `AdditionalArtifactStores`. I will apply this rename in the next iteration when we resume this work.

saschagrunert:
Ref: https://github.com/cri-o/cri-o/pull/9691

internal/ociartifact/store.go
```diff
@@ -23,37 +23,60 @@ var ErrNotFound = libartTypes.ErrArtifactNotExist
...
+		addPath := filepath.Join(path, "artifacts")
+
+		addStore, err := libart.NewArtifactStore(addPath, systemContext)
+		if err != nil {
+			// Just ignore invalid stores to prevent breaking CRI-O startup
+			continue
+		}
+		additionalStores = append(additionalStores, addStore)
+		validAdditionalPaths = append(validAdditionalPaths, addPath)
+	}
```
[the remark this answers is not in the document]
bitoku: @pauloappbr please check this.

internal/ociartifact/store.go
```diff
@@ -89,12 +120,20 @@ func (s *Store) Status(ctx context.Context, nameOrDigest string) (*Artifact, err
...
 
-	return NewArtifact(artifact), nil
+	// Check additional stores
+	for _, store := range s.additionalStores {
+		if artifact, err := store.Inspect(ctx, artRef); err == nil {
+			return NewArtifact(artifact), nil
+		}
+	}
+
+	return nil, fmt.Errorf("inspect artifact: %w", err)
```
bitoku, line 293: I think additionalStore should be prioritized.

internal/ociartifact/store.go
```diff
@@ -105,15 +144,42 @@ func (s *Store) Remove(ctx context.Context, nameOrDigest string) error {
...
+	// Only remove from the main writeable store
 	_, err = s.libartifactStore.Remove(ctx, artRef)
 
 	return err
 }
 
+// resolveRootPath finds which directory (main or additional) the artifact resides in.
+func (s *Store) resolveRootPath(ctx context.Context, nameOrDigest string) (string, error) {
+	artRef, err := libart.NewArtifactStorageReference(nameOrDigest)
+	if err != nil {
```
bitoku: I wonder if we could add rootPath field in Artifact struct. This would require additional reference which is redundant.

pkg/config/config.go
```diff
@@ -2289,3 +2293,15 @@ func (c *APIConfig) GetTLSMinVersion() uint16 {
...
+// Used for testing only.
+func (c *Config) SetCgroupManager(manager cgmgr.CgroupManager) {
+	c.cgroupManager = manager
+}
+
+// SetCheckpointRestore sets the checkpoint/restore support.
+// Used for testing only.
+func (c *Config) SetCheckpointRestore(enabled bool) {
+	c.EnableCriuSupport = enabled
+}
```
bitoku: Let's avoid adding test functions in the real code. You can find in *_test_inject.go how we handle this.

internal/ociartifact/store.go
```diff
@@ -23,37 +23,60 @@ var ErrNotFound = libartTypes.ErrArtifactNotExist
...
 ) (manifestDigest *digest.Digest, err error) {
 	strRef := ref.DockerReference().String()
 
+	// Check if it already exists in any store (including RO) to avoid pulling
+	if art, err := s.Status(ctx, strRef); err == nil {
+		log.Infof(ctx, "Artifact %s found locally, skipping pull", strRef)
+		dgst := art.Digest()
+		return &dgst, nil
+	}
+
```
bitoku: I'm just curious, do we already handle the duplicate artifact case in `libartifactStore.Pull` ?

internal/ociartifact/store.go
pauloappbr: Makes complete sense! Good catch.
I've updated the logic in Status(), resolveRootPath(), and List() to prioritize the additionalStores over the main store. The latest push includes this fix. Let me know if it looks good now!"

internal/ociartifact/store.go
pauloappbr: You are totally right! Calling Inspect again in resolveRootPath was redundant. I've added the rootPath to the Artifact struct so it inherently knows where it came from, and removed resolveRootPath entirely. The latest push includes this fix. Thanks for the suggestion!S

pkg/config/config.go
pauloappbr: Ah, good catch! I missed moving these two into the _test_inject.go file during the rebase conflict resolution. I've moved them now. Thanks!

internal/ociartifact/store.go
pauloappbr: ou are right! libartifactStore.Pull handles duplicates, but only for the main writeable store. It has no awareness of our new additionalStores.

If we don't do this manual check using s.Status() beforehand, it would re-download the artifact into the main store even if it already exists in one of the read-only stores, consuming unnecessary disk space and network bandwidth. This check ensures we truly skip the pull if the artifact is available anywhere.

internal/ociartifact/datastore/store.go
saschagrunert: This uses `s.RootPath() ` to create a layout reference for reading artifact data. When an artifact is found in an additional store (via `getByNameOrDigest` -> `List`), the artifact's `rootPath` will point to the additional store, but the layout reference still uses the main store's path. This will cause failures when trying to read data from artifacts in additional stores.

We should use `artifact.RootPath() `instead of `s.RootPath()` here.

pkg/config/config.go
```diff
@@ -424,6 +424,10 @@ type RuntimeConfig struct {
 	// DecryptionKeysPath is the path where keys for image decryption are stored.
 	DecryptionKeysPath string `toml:"decryption_keys_path"`
 
+	// ReadOnlyArtifactStores is a list of additional read-only artifact stores.
+	// Each path behaves like the main artifact store but is read-only.
+	ReadOnlyArtifactStores []string `toml:"read_only_artifact_stores,omitempty"`
```
saschagrunert: We should stick to one name here: `additional_artifact_stores`.

saschagrunert: We need a config validation for this field that every path should be absolute, like:

```go
  for _, p := range c.ReadOnlyArtifactStores {
      if !filepath.IsAbs(p) {
          return fmt.Errorf("read_only_artifact_stores entry must be absolute: %q", p)
      }
  }
```

internal/ociartifact/store.go
```diff
@@ -105,15 +164,18 @@ func (s *Store) Remove(ctx context.Context, nameOrDigest string) error {
 		return fmt.Errorf("invalid nameOrDigest: %w", err)
 	}
 
+	// Only remove from the main writeable store
 	_, err = s.libartifactStore.Remove(ctx, artRef)
```
saschagrunert: Let's wrap this error: 
```suggestion
	if _, err := s.libartifactStore.Remove(ctx, artRef); err != nil { return fmt.Errorf(…) }
```

internal/factory/container/device_test.go
```diff
@@ -296,16 +296,18 @@ containerEdits:
 				},
```
saschagrunert, line 296: This change is unrelated, right?

pkg/config/config_test_inject.go
```diff
@@ -38,3 +38,15 @@ func (c *RuntimeConfig) SetCheckpointRestore(cr bool) {
...
+// Used for testing only.
+func (c *Config) SetCgroupManager(manager cgmgr.CgroupManager) {
+	c.cgroupManager = manager
+}
+
+// SetCheckpointRestore sets the checkpoint/restore support.
+// Used for testing only.
+func (c *Config) SetCheckpointRestore(enabled bool) {
+	c.EnableCriuSupport = enabled
+}
```
saschagrunert: I guess those are unnecessary, right?

internal/ociartifact/store.go
```diff
@@ -23,37 +24,63 @@ var ErrNotFound = libartTypes.ErrArtifactNotExist
...
-//
-// copyOptions will be passed down to libimage.
 func (s *Store) Pull(
 	ctx context.Context,
 	ref types.ImageReference,
 	opts *libimage.CopyOptions,
 ) (manifestDigest *digest.Digest, err error) {
 	strRef := ref.DockerReference().String()
 
+	// Check if it already exists in any store (including RO) to avoid pulling
```
saschagrunert: I'd like to extend this comment to underline how it's working:


```suggestion
	// Skip pulling if the artifact already exists in any store (including
	// read-only additional stores). This avoids duplicating artifacts that
	// are already available and prevents unnecessary network traffic.
	// To force a re-pull (e.g., after a tag re-point), the artifact must
	// be removed first.
```

This should be covered by an integration test.

saschagrunert:
@pauloappbr do you mind giving this a rebase and addressing the review comments? :)

internal/factory/container/device_test.go
pauloappbr: Yes, it was unrelated. I've reverted it back to the upstream version in the latest push.

pkg/config/config_test_inject.go
pauloappbr: Yes, unnecessary. Reverted back to upstream as well!

pauloappbr:
Hi @saschagrunert, I've addressed all the feedback!

Fixed the layout reference to use artifact.RootPath() in the datastore.

Renamed the configuration back to additional_artifact_stores.

Added absolute path validation in config.go.

Wrapped the Remove error as requested.

Extended the comment in Pull().

Reverted the unrelated test changes.

Added an integration test in test/artifact_additional_stores.bats.

Let me know if anything else is needed!

saschagrunert:
@pauloappbr thanks! You need to rebase on top of the latest `main` branch.

pauloappbr:
Hi @saschagrunert, rebased successfully on top of the latest main!
Also, regarding device_test.go: I actually had to keep that fix. Recent changes in the CDI upstream make it default to 0660 (432 decimal) instead of nil, so the test was failing locally and in the CI without it. I've restored the 0o660 fix to keep the suite green. Everything is ready to go!

internal/storage/image.go
```diff
@@ -870,7 +870,7 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 	if shouldTryArtifact(err) {
 		log.Infof(ctx, "Falling back to pull %s as an OCI artifact: %v", imageName, err)
 
-		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), &srcSystemContext)
+		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), nil, &srcSystemContext)
```
saschagrunert: This creates a new store with `nil` for `additionalPaths`. When the regular image pull fails and CRI-O falls back to pulling as an OCI artifact, the skip-if-already-present optimization won't work. Artifacts in read-only stores will still be re-downloaded. The config needs to be threaded through here.

internal/ociartifact/store.go
```diff
@@ -86,6 +108,17 @@ func (s *Store) Pull(
...
 
+	// Skip pulling if the artifact already exists in any store (including
+	// read-only additional stores). This avoids duplicating artifacts that
+	// are already available and prevents unnecessary network traffic.
+	// To force a re-pull (e.g., after a tag re-point), the artifact must
+	// be removed first.
+	if art, err := s.Status(ctx, strRef); err == nil {
+		log.Infof(ctx, "Artifact %s found locally, skipping pull", strRef)
+		dgst := art.Digest()
+		return &dgst, nil
```
saschagrunert: This short-circuits the pull for **all** artifacts, not just those in read-only stores. If a tag is re-pointed on the registry, `Pull` will silently return the stale local copy. Worse: if the stale copy is in a read-only store, the user **cannot** remove it since `Remove` only operates on the main store. This should either only skip for read-only additional stores, or be a clearly documented intentional behavioral change.

internal/ociartifact/store.go
```diff
@@ -171,13 +204,37 @@ func (s *Store) EnsureNotContainerImage(ctx context.Context, ref types.ImageRefe
...
-	arts, err := s.libartifactStore.List(ctx)
+	var arts []*Artifact
+
+	// Get from additional stores first (Prioritized)
+	for i, addStore := range s.additionalStores {
+		addArts, err := addStore.List(ctx)
+		if err != nil {
+			continue
+		}
+		for _, art := range addArts {
```
saschagrunert, line 231: Errors from additional stores are silently swallowed while errors from the main store are hard failures. At minimum emit a warning log here so operators can diagnose issues with misconfigured additional stores.

internal/ociartifact/store.go
```diff
@@ -171,13 +204,37 @@ func (s *Store) EnsureNotContainerImage(ctx context.Context, ref types.ImageRefe
...
 	}
 
-	for _, art := range arts {
-		res = append(res, NewArtifact(art))
+	// Deduplicate preserving priority
+	seen := make(map[string]struct{})
+	for _, artifact := range arts {
+		encoded := artifact.Digest().Encoded()
+		if _, ok := seen[encoded]; ok {
+			continue
```
saschagrunert, line 254: Deduplication uses only digest. Two artifacts with the same digest but different tags (e.g. `model:v1` and `model:latest`) will be collapsed to whichever is seen first, making some tags disappear from listings.

internal/ociartifact/store.go
```diff
@@ -50,29 +51,50 @@ type Store struct {
...
+		if err != nil {
+			logrus.WithFields(map[string]interface{}{
+				"path": addPath,
+				"err":  err,
+			}).Warn("skipping additional artifact store")
+			continue
+		}
+		additionalStores = append(additionalStores, &artifactStore{addStore})
+		validAdditionalPaths = append(validAdditionalPaths, addPath)
+	}
```
saschagrunert, line 89: `logrus` is used directly here while the rest of the codebase uses `internal/log`. Consider either adding a `context.Context` parameter to `NewStore` or using `log.Warnf(context.Background(), ...)` for consistency.

test/artifact_additional_stores.bats
```diff
@@ -0,0 +1,49 @@
...
+additional_artifact_stores = [
+    "./relative/path/store"
+]
+EOF
+
+	# Try to start CRI-O, it should fail due to our config validation
+	run "$CRIO_BINARY_PATH" --config-dir "$CRIO_CONFIG_DIR" config
+	assert_failure
+	assert_output --partial "additional_artifact_stores entry must be absolute: \"./relative/path/store\""
+}
```
saschagrunert: These tests only verify config parsing (start with valid config, reject relative paths). The actual feature — pulling an artifact that already exists in a read-only store and verifying the pull is skipped, listing across stores — is not tested. As noted in the earlier review, the skip-logic in `Pull` should be covered by an integration test. Also: missing newline at end of file.

internal/factory/container/device_test.go
```diff
@@ -296,16 +296,18 @@ containerEdits:
...
 					},
 					{
-						Path:  "/dev/loop9",
-						Type:  "b",
-						Major: 7,
-						Minor: 9,
+						Path:     "/dev/loop9",
+						Type:     "b",
+						Major:    7,
+						Minor:    9,
```
saschagrunert: These `FileMode` changes are unrelated to this feature and should be dropped from this PR.

internal/storage/image.go
pauloappbr: You're absolutely right, the config needs to be threaded here so we don't lose the additional paths. Since pullImageImplementation and the ImageServer interface don't currently have access to config.AdditionalArtifactStores, what would be your preferred way to thread it? Should I add it to ImageCopyOptions, modify the ImageServer struct/interface, or is there a cleaner path you'd recommend

internal/ociartifact/store.go
pauloappbr: Great catch! I've updated the logic in the latest push to only short-circuit if the artifact is found inside the additionalStores loop. The main store will now always fall through to libartifactStore.Pull.

internal/ociartifact/store.go
pauloappbr: Fixed in the latest push. It now emits a log.Warnf if listing an additional store fails.

internal/ociartifact/store.go
pauloappbr: Good point regarding the tags. I've updated the deduplication logic to use a compound key (artifact.Name() + "@" + artifact.Digest().Encoded()) so we don't collapse different tags with the same digest.

internal/ociartifact/store.go
pauloappbr: Changed to log.Warnf(context.Background(), ...) for consistency. Thanks!

internal/factory/container/device_test.go
pauloappbr: Dropped from the PR completely. I'll let the CI handle the FileMode expectations.

test/artifact_additional_stores.bats
pauloappbr: I added the missing newline! Regarding the integration test for the Pull skip-logic: Since this requires pulling an artifact that exists locally, could you give me a quick pointer on how CRI-O's BATS framework usually mocks the registry for artifact tests? I'll gladly add the test once I know the preferred mocking pattern.

internal/ociartifact/store.go
```diff
@@ -86,6 +104,14 @@ func (s *Store) Pull(
 
 	strRef := ref.DockerReference().String()
 
+	// Skip pulling if the artifact already exists in a read-only additional store.
+	// If it is in the main writeable store, we proceed with pull to allow tag re-pointing.
+	if art, err := s.Status(ctx, strRef); err == nil && art.RootPath() != s.rootPath {
```
saschagrunert: `Status()` checks additional stores first, so if the artifact exists in both an additional store and the main store, it will always match the additional store and skip the pull. This means tag re-pointing in the main store is only possible when the artifact does **not** exist in any additional store. Is that the intended behavior? If a user wants to update the main store's tag, they'd have to remove the artifact from the additional store first.

Also consider logging at `Debug` level when `Status()` returns an error here, to aid troubleshooting when an artifact unexpectedly isn't found in additional stores.

internal/ociartifact/store.go
```diff
@@ -191,12 +242,20 @@ func (s *Store) Status(ctx context.Context, nameOrDigest string) (*Artifact, err
 		return nil, fmt.Errorf("invalid nameOrDigest: %w", err)
 	}
 
+	// Check additional stores first (Prioritized)
+	for i, store := range s.additionalStores {
+		if artifact, err := store.Inspect(ctx, artRef); err == nil {
```
saschagrunert: Errors from `Inspect()` on additional stores are silently discarded. This will make it very hard to debug issues like permission denied, corrupt OCI layout, or NFS timeouts. Please add at least `log.Debugf(ctx, "Inspect in additional store %q failed: %v", s.additionalPaths[i], err)` so operators have a chance to diagnose why an artifact isn't being found.

internal/ociartifact/store.go
```diff
@@ -171,13 +197,38 @@ func (s *Store) EnsureNotContainerImage(ctx context.Context, ref types.ImageRefe
...
+	for _, art := range mainArts {
+		arts = append(arts, NewArtifact(art, s.rootPath))
 	}
 
-	for _, art := range arts {
-		res = append(res, NewArtifact(art))
+	// Deduplicate preserving priority (uses both Name and Digest)
+	seen := make(map[string]struct{})
+	for _, artifact := range arts {
+		key := fmt.Sprintf("%s@%s", artifact.Reference(), artifact.Digest().Encoded())
```
saschagrunert: This deduplicates by reference+digest, meaning the same blob with different tags (e.g., `:v1` in additional store, `:latest` in main store) will appear twice. Is that intentional? If so, a brief comment explaining the dedup semantics would help future readers. If not, deduplicating by digest alone would collapse these.

internal/ociartifact/store.go
```diff
@@ -171,13 +197,38 @@ func (s *Store) EnsureNotContainerImage(ctx context.Context, ref types.ImageRefe
 
 // List creates a slice of all available artifacts.
 func (s *Store) List(ctx context.Context) (res []*Artifact, err error) {
-	arts, err := s.libartifactStore.List(ctx)
+	var arts []*Artifact
+
+	// Get from additional stores first (Prioritized)
+	for i, addStore := range s.additionalStores {
```
saschagrunert: The asymmetry between additional stores (warn and continue) and the main store (return error) is correct, but worth a short comment explaining *why*. For example: additional stores may reside on unreliable media like NFS and shouldn't block the entire listing.

internal/ociartifact/datastore/store.go
```diff
@@ -43,7 +43,7 @@ type Store struct {
 
 // New creates a new OCI artifact data store.
 func New(rootPath string, systemContext *types.SystemContext) (*Store, error) {
```
saschagrunert, line 45: The datastore is used for in-memory artifact data (seccomp profiles, etc.). Passing `nil` means artifacts in additional stores won't be discoverable through this code path. If this is intentional (datastore only handles pulled-into-main-store artifacts), please add a comment explaining why. If not, the additional paths should be threaded through.

internal/storage/image.go
saschagrunert: Same as the datastore: this fallback path creates a fresh store without additional paths, so artifacts pre-populated in additional stores won't be found during the image-pull fallback. Since this path is hit when a regular image pull fails and CRI-O retries as an artifact, it should arguably also check additional stores. At minimum, document why it's excluded.

pkg/config/config.go
```diff
@@ -1331,6 +1335,12 @@ func (c *RootConfig) CleanShutdownSupportedFileName() string {
 // execution checks. It returns an `error` on validation failure, otherwise
 // `nil`.
 func (c *RuntimeConfig) Validate(systemContext *types.SystemContext, onExecution bool) error {
+	for _, p := range c.AdditionalArtifactStores {
```
saschagrunert, line 1339: The openshift/api `StorePath` type enforces stricter validation: only `[a-zA-Z0-9/._-]`, no consecutive `//`, max 256 chars. Consider matching these rules here so standalone CRI-O catches the same invalid configs that OpenShift would reject. This prevents confusing failures when migrating configs between environments.

pkg/config/config.go
```diff
@@ -425,6 +425,10 @@ type RuntimeConfig struct {
 	// DecryptionKeysPath is the path where keys for image decryption are stored.
 	DecryptionKeysPath string `toml:"decryption_keys_path"`
 
+	// AdditionalArtifactStores is a list of additional read-only artifact stores.
```
saschagrunert, line 427: Please document that CRI-O expects an `artifacts/` subdirectory within each configured path (mirroring the main store convention). Without this, users will create paths like `/mnt/nfs-artifacts` and wonder why their artifacts aren't found when they should be at `/mnt/nfs-artifacts/artifacts/`.

internal/ociartifact/artifact.go
```diff
@@ -26,13 +26,15 @@ func (u unknownRef) Name() string {
 type Artifact struct {
 	*libartifact.Artifact
 
+	rootPath string
```
saschagrunert, line 29: Nit: consider whether this belongs on the `Artifact` type or should be returned alongside it (e.g., a result struct). Embedding storage location into a domain object couples the artifact identity with where it's stored. Not blocking, but worth considering for future maintainability.

test/artifact_additional_stores.bats
saschagrunert, line 157: These tests verify config parsing and validation, but don't test the actual runtime behavior: pulling an artifact that exists in an additional store, listing from multiple stores, or blob mount resolution from an additional store. Consider adding at least one end-to-end test that pre-populates an artifact in an additional store directory and verifies CRI-O resolves it without pulling.

pkg/config/config.go
pauloappbr: I've added the 256 character limit and consecutive slashes (//) checks. Note: I had initially added a strict regex for the character restrictions as you suggested, but it broke the e2e CI tests because the test suite generates temporary paths that didn't match it. I've relaxed it to only check for absolute paths, length, and slashes so the CI can pass while still keeping it safe.

pkg/config/config.go
pauloappbr: Done! Added the documentation explicitly mentioning the artifacts/ subdirectory expectation in the struct.

internal/ociartifact/store.go
pauloappbr: You are absolutely right. Since Status() checks the additional stores first, tag re-pointing for an artifact that already exists in a read-only store is intentionally not supported. If a user needs to force-pull a new digest for that specific tag, the artifact must be removed from the read-only store first. I've added a comment block right above this if statement to explicitly document this behavior constraint.

I've also added the log.Debugf directly inside the s.Status() method as you suggested, so we can now trace exactly when an artifact isn't found or errors out in the additional stores.

internal/ociartifact/store.go
pauloappbr: Yes, it is strictly intentional! If we deduplicated by digest alone, we would collapse and hide different tags (like :v1 and :latest) that happen to point to the exact same blob, which could confuse users trying to list their artifacts. As suggested, I've added a brief comment right above the deduplication block explaining this exact semantic to help future readers.

internal/ociartifact/store.go
pauloappbr: Done! I've added the exact log.Debugf you suggested to the else block here to ensure operators can properly diagnose underlying storage or permission issues.

internal/ociartifact/store.go
pauloappbr: Great point! I've added a comment right above this loop explaining exactly that: we warn and continue to prevent unreliable media (like NFS) from blocking the main store's listing.

internal/ociartifact/datastore/store.go
pauloappbr: It is intentional for now! Threading the configuration through here (and the image service fallback) would require a larger refactor of the CRI-O interfaces. As you suggested, I've added a clear comment block right above func New explicitly documenting why additionalPaths are excluded and passed as nil here.

internal/storage/image.go
pauloappbr: Exactly the same situation as the datastore! Threading the configuration through the CRI image service interfaces just for this fallback path would require a fairly large refactor. Following your suggestion, I've added a clear comment block right above this NewStore call to explicitly document why additionalPaths is passed as nil here for now.

internal/ociartifact/artifact.go
pauloappbr: That is a very fair architectural point! I embedded it here to avoid a cascading refactor of the return types across the List and Status interfaces, which would significantly increase the scope of this PR. Since it's not blocking, I'll leave it as is for now to keep things contained, but I completely agree it's a great candidate for a future cleanup PR to decouple the identity from the storage location.

test/artifact_additional_stores.bats
pauloappbr: You are absolutely right, an end-to-end test is needed here. However, since manually generating a fully valid OCI layout (with the correct blobs, manifest, and sha256 hashes) purely via BATS without relying on external tools like skopeo is quite complex, would you be open to me handling the E2E artifact BATS test in a follow-up PR? This way I can make sure I use the proper CRI-O testing mocks and tools for it without holding up this feature.

pauloappbr:
Hi @saschagrunert, I've addressed all the points! I also relaxed the config validation (removed the strict regex but kept the 256-char and consecutive slashes checks) because it was breaking the e2e test suite with temporary paths. Everything should be clean now. Retesting!

saschagrunert:
@pauloappbr thank you, please fix the CI as well

internal/ociartifact/store.go
```diff
@@ -191,12 +249,22 @@ func (s *Store) Status(ctx context.Context, nameOrDigest string) (*Artifact, err
 		return nil, fmt.Errorf("invalid nameOrDigest: %w", err)
 	}
 
+	// Check additional stores first (Prioritized)
+	for i, store := range s.additionalStores {
+		if artifact, err := store.Inspect(ctx, artRef); err == nil {
+			return NewArtifact(artifact, s.additionalPaths[i]), nil
+		} else {
+			log.Debugf(ctx, "Inspect in additional store %q failed: %v", s.additionalPaths[i], err)
```
saschagrunert: `Debug` level here means permission errors, corrupt OCI layouts, or NFS timeouts on additional stores will be invisible at default log levels. Consider splitting: `log.Debugf` when the error wraps `ErrNotFound`, `log.Warnf` otherwise. That way operators can diagnose misconfigurations without enabling debug logging.

internal/ociartifact/store.go
```diff
@@ -86,6 +104,16 @@ func (s *Store) Pull(
 
 	strRef := ref.DockerReference().String()
 
+	// Skip pulling if the artifact already exists in a read-only additional store.
+	// Note: Because we short-circuit here, tag re-pointing for artifacts present
+	// in additional stores is not supported. If a user needs to force-pull a new
+	// digest for a tag, the artifact must not exist in the read-only store.
+	if art, err := s.Status(ctx, strRef); err == nil && art.RootPath() != s.rootPath {
```
saschagrunert: The constraint that tag re-pointing is blocked when the artifact exists in a read-only store (and can't be removed from there via `Remove`) should be documented somewhere user-facing, not just in this code comment. Users who hit this have no CRI-O-native way to resolve it; they'd need manual filesystem access to the read-only store.

internal/storage/image.go
```diff
@@ -870,7 +870,10 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 	if shouldTryArtifact(err) {
 		log.Infof(ctx, "Falling back to pull %s as an OCI artifact: %v", imageName, err)
 
-		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), &srcSystemContext)
+		// Note: We pass nil for additionalPaths here. Since this fallback path
+		// is only hit when a regular container image pull fails, we do not currently
+		// thread the additional read-only artifact stores through the image service.
+		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), nil, &srcSystemContext)
```
saschagrunert: This is a primary pull path (hit whenever a regular image pull fails), not an edge case. Passing `nil` means artifacts pre-populated in additional stores won't be found during this fallback. Could you file a follow-up issue and reference it here so this doesn't get lost?

saschagrunert:
@bitoku PTAL as well.

internal/ociartifact/store.go
```diff
@@ -50,29 +50,47 @@ type Store struct {
 	impl             Impl
 
 	// rootPath is required for BlobMountPaths.
-	rootPath string
+	rootPath         string
+	additionalPaths  []string
+	additionalStores []LibartifactStore
```
bitoku: I want them to be a list of structs so that we won't make unintended inconsistency.

internal/ociartifact/artifact.go
bitoku, line 86: @saschagrunert do you remember why it's pinned true?
I wonder if we can make artifacts from r/o store pinned and otherwise pinnned false.

internal/ociartifact/store.go
```diff
@@ -191,12 +249,22 @@ func (s *Store) Status(ctx context.Context, nameOrDigest string) (*Artifact, err
...
+		} else {
+			log.Debugf(ctx, "Inspect in additional store %q failed: %v", s.additionalPaths[i], err)
+		}
+	}
+
+	// Check main store
 	artifact, err := s.libartifactStore.Inspect(ctx, artRef)
-	if err != nil {
-		return nil, fmt.Errorf("inspect artifact: %w", err)
+	if err == nil {
```
bitoku, line 289: What made you decide to change this 'if' statement?

internal/ociartifact/artifact.go
saschagrunert: I think this was related to garbage collection avoidance.

saschagrunert:
@pauloappbr do you still work on this? :)

saschagrunert:
> What if additional store has registry.io/model:v1 pointing to sha256:aaa, and the main store has registry.io/model:v1 pointing to sha256:bbb? How do we resolve that issue in image store? I think we need more discussion about dedup strategy (validation before running? or doing something smart in runtime?)

The current implementation follows the same precedence model as `additionalimagestores` in containers/storage: the read-only store wins. `Status()` and `Pull()` check additional stores first, so if `registry.io/model:v1` exists in the RO store, that version is used and the main store's copy is shadowed.

I think that's the right default for the target use case (pre-populated models on shared/read-only media should not be overwritten), but we should at minimum emit a warning when we detect a tag pointing to different digests across stores. That way operators can spot conflicts without enabling debug logging.

For a first iteration I'd keep it simple (RO wins, warn on conflict) and defer more complex strategies (like startup validation or user-configurable precedence) until we have real-world feedback. Agreed that the dedup semantics should be documented clearly in the user-facing config docs.

internal/ociartifact/store.go
saschagrunert: The original code was:

```go
artifact, err := s.libartifactStore.Inspect(ctx, artRef)
if err != nil {
    return nil, fmt.Errorf("inspect artifact: %w", err)
}
return NewArtifact(artifact), nil
```

It was changed to `if err == nil` to match the control flow of the additional stores loop above it, which also uses `if err == nil` for the success path. Both branches follow the same pattern: try inspect, return on success, fall through on failure. The error is only returned at the end if no store found the artifact. It's a stylistic change to keep the two code paths consistent, not a behavioral change.

internal/ociartifact/artifact.go
saschagrunert: `Pinned: true` was set in the original code before this PR (not a change from this PR). It means the kubelet's garbage collector won't remove the artifact. This makes sense because OCI artifacts (models, configs) are explicitly pulled and should not be garbage collected like unused container images. Making artifacts from read-only stores pinned and others not would be risky, since non-pinned artifacts in the main store could get garbage collected unexpectedly.

internal/ociartifact/store.go
saschagrunert: Makes sense. Right now they're kept in sync by index (`additionalStores[i]` corresponds to `additionalPaths[i]`), which is fragile. A struct like `type additionalStore struct { path string; store LibartifactStore }` would be cleaner. We can address this in a follow-up.

internal/ociartifact/store.go
saschagrunert: Fixed in 5ff8cc3d7. The inspect loop now uses `log.Debugf` for `ErrNotFound` and `log.Warnf` for all other errors:

```go
if errors.Is(err, ErrNotFound) {
    log.Debugf(ctx, "Artifact not found in additional store %q", add.path)
} else {
    log.Warnf(ctx, "Failed to inspect artifact in additional store %q: %v", add.path, err)
}
```

internal/ociartifact/store.go
saschagrunert: Fixed in 5ff8cc3d7. The tag re-pointing limitation is now documented in both the man page (`crio.conf.5.md`) and the config template output.

internal/storage/image.go
saschagrunert: Fixed in 5ff8cc3d7. `AdditionalArtifactStores` is now threaded through `ImageCopyOptions` and passed to `ociartifact.NewStore()` in the fallback path, so no follow-up issue needed.

saschagrunert:
I pushed a commit on top of this branch to address the review comments. PTAL again @cri-o/cri-o-maintainers 

/cherry-pick release-1.35

internal/ociartifact/store.go
```diff
@@ -86,6 +111,18 @@ func (s *Store) Pull(
...
+	// are already available and prevents unnecessary network traffic.
+	// To force a re-pull (e.g., after a tag re-point), the artifact must
+	// be removed first.
+	if art, err := s.Status(ctx, strRef); err == nil {
+		log.Infof(ctx, "Artifact %s already exists locally in %s, skipping pull", strRef, art.RootPath())
+		dgst := art.Digest()
+
+		return &dgst, nil
+	}
+
```
bitoku: This will change the current behavior even without additional store.

`registry.example.com/models/llm:v2` which resolves to digest `sha256:aaa` won't be updated once it's in the main store even if someone pushes a new artifact to `registry.example.com/models/llm:v2`, same tag with different digest `sha256:bbb`.

internal/ociartifact/datastore/store.go
```diff
@@ -42,8 +42,12 @@ type Store struct {
 }
 
 // New creates a new OCI artifact data store.
+// Note: The datastore currently only handles artifacts pulled into the main store.
+// Additional read-only stores are not threaded through here (we pass nil for
+// additionalPaths) since the datastore is primarily used for in-memory artifact
+// data managed by the main CRI-O lifecycle.
```
bitoku: non blocking: this is implementation detail and should be in the function code before L50, instead of the function document.

internal/ociartifact/artifact.go
bitoku: It's expected that artifacts are garbage collected, isn't it? There should be a check in GC to check whether the artifact/image is used.

docs/crio.conf.5.md
```diff
@@ -131,6 +131,9 @@ If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
 **decryption_keys_path**="/etc/crio/keys/"
 Path where the keys required for image decryption are located
 
+**additional_artifact_stores**=[]
+A list of additional read-only OCI artifact store paths. CRI-O expects an "artifacts/" subdirectory within each configured path. All entries must be absolute paths. Artifacts in these stores take priority over the main store. Because these stores are read-only, CRI-O cannot remove artifacts from them. If a tag is re-pointed on the registry, the stale local copy in a read-only store will continue to be used; the artifact must be removed from the read-only store directly on the filesystem to pick up the new version.
```
bitoku: I think it'd be better to say "experimental" or "subject to change", if we may add some validation when registering the additional stores in the future.

internal/ociartifact/store.go
```diff
@@ -171,13 +208,49 @@ func (s *Store) EnsureNotContainerImage(ctx context.Context, ref types.ImageRefe
...
+	for _, artifact := range arts {
+		key := fmt.Sprintf("%s@%s", artifact.Reference(), artifact.Digest().Encoded())
+		if _, ok := seen[key]; ok {
+			continue
+		}
+
+		seen[key] = struct{}{}
+
+		res = append(res, artifact)
```
bitoku: I wonder how podman with additional image store handle this case.
Could it show the duplicate tags?

internal/ociartifact/store.go
saschagrunert: Fixed. The skip now only applies to artifacts found in read-only additional stores. Artifacts in the main store are always re-pulled, so tag re-pointing works as expected.

internal/ociartifact/datastore/store.go
saschagrunert: Good catch, moved the note inside the function body.

docs/crio.conf.5.md
saschagrunert: Done. Added "experimental, subject to change" to the docs and config template.

internal/ociartifact/store.go
saschagrunert: Changed the dedup key to reference only (instead of reference+digest). Since additional stores are listed first, they win when the same tag exists in both stores. This matches how podman's additional image stores handle priority.

internal/ociartifact/artifact.go
saschagrunert: Artifacts are not garbage collected today. They are pinned and only removed via explicit crictl rmi. GC awareness for artifacts is out of scope for this PR, but worth tracking separately.

Ref https://github.com/cri-o/cri-o/issues/9827

saschagrunert:
Integration tests fail all over the place because critest requires go 1.26 now. We may wanna push https://github.com/cri-o/cri-o/pull/9774 for that.

saschagrunert:
Rebased

PR #9706, opened 2026-01-12
server,nri: pass extended container status to NRI.

klihub:
#### What type of PR is this?

/kind other

This PR implements passing more complete container status information to NRI, including init PID, exit code, and timestamps for container creation, start, and exit events.

None

```release-note
- nri: pass more complete container status to NRI, including PID, exit code, and timestamps fro container creation, start, and exit events
```

## Summary by CodeRabbit

* **Improvements**
  * Richer container status reporting: lifecycle state plus PID, creation/start/finish timestamps, and exit code.
  * More detailed exit reasons and messages for clearer diagnostics (OOM, seccomp, completed vs. error).
  * Status retrieval now returns a structured status object for more complete state information.

* **Bug Fixes**
  * Stop/cleanup flow adjusted so container state can be updated before invoking the runtime stop, reducing risk of lost state.

### DIFF
--- a/server/container_stop.go
+++ b/server/container_stop.go
@@ -79,9 +79,10 @@ func (s *Server) postStopCleanup(ctx context.Context, ctr *oci.Container, sb *sa
 		}
 	}
 
-	if err := s.nri.stopContainer(ctx, sb, ctr); err != nil {
+	if err := s.nri.stopContainer(ctx, sb, ctr, true); err != nil {
 		log.Warnf(ctx, "NRI stop container request of %s failed: %v", ctr.ID(), err)
 	}
+
 	// persist container state at the end, so there's no window where CRI-O reports the container
 	// as stopped, but hasn't run post stop hooks.
 	if err := s.ContainerStateToDisk(ctx, ctr); err != nil {
--- a/server/container_start.go
+++ b/server/container_start.go
@@ -89,7 +89,7 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
 				}
 			}
 
-			if err := s.nri.stopContainer(ctx, sandbox, c); err != nil {
+			if err := s.nri.stopContainer(ctx, sandbox, c, false); err != nil {
 				log.Warnf(ctx, "NRI stop failed for container %q: %v", c.ID(), err)
 			}
 
--- a/server/nri-api.go
+++ b/server/nri-api.go
@@ -293,9 +293,15 @@ func (a *nriAPI) postUpdateContainer(ctx context.Context, criCtr *oci.Container)
 	return a.nri.PostUpdateContainer(ctx, pod, ctr)
 }
 
-func (a *nriAPI) stopContainer(ctx context.Context, criPod *sandbox.Sandbox, criCtr *oci.Container) error {
+func (a *nriAPI) stopContainer(ctx context.Context, criPod *sandbox.Sandbox, criCtr *oci.Container, updateState bool) error {
 	if !a.isEnabled() {
 		return nil
+	}
+
+	if updateState {
+		if err := a.cri.Runtime().UpdateContainerStatus(ctx, criCtr); err != nil {
+			log.Warnf(ctx, "Error updating the container status  %q: %v", criCtr.ID(), err)
+		}
 	}
 
 	ctr := &criContainer{


server/container_stop.go
```diff
@@ -79,12 +79,13 @@ func (s *Server) postStopCleanup(ctx context.Context, ctr *oci.Container, sb *sa
...
-	}
 	// persist container state at the end, so there's no window where CRI-O reports the container
 	// as stopped, but hasn't run post stop hooks.
 	if err := s.ContainerStateToDisk(ctx, ctr); err != nil {
 		log.Warnf(ctx, "Unable to write containers %s state to disk: %v", ctr.ID(), err)
 	}
+
+	if err := s.nri.stopContainer(ctx, sb, ctr); err != nil {
+		log.Warnf(ctx, "NRI stop container request of %s failed: %v", ctr.ID(), err)
+	}
```
bitoku: Any reason to move this to the end?
Considering the comment above `ContainerStateToDisk`, the `ContainerStateToDisk` should be the end.

klihub: @bitoku @haircommander If we don't do it this way around, then we end up passing to NRI plugins a wrong state (running), no ExitCode and no FinishedAt timestamp.

Also BTW, if we don't do it in this order and we try to do a `server.ContainerServer.Runtime().UpdateContainerStatus(context.Background(), ctr)` (which now we don't) prior to digging out state, timestamps and ExitCode, extremely bad things happen when (processing) containers that exit on their own accord (or crash).

bitoku: The reason why we put it at the end is to avoid race conditions.
There's a risk that while the postStop hook is running, something may do something believing the container is cleaned up. Nothing can know that the hook is running.

We may be able to extract a function to fetch container status from each runtime implementation.

klihub: I'm fine with any alternative, as long as we get the NRI-related semantics correct (the NRI StopContainer event/request is implicitly post-StopContainer by semantics):
- we get the container state correctly as stopped
- we get a correct (non-zero) FinishedAt timestamp
- we get the container's ExitCode

I tried to get to this by keeping the call to s.ContainerStateToDisk() at the end and then trying to infer the correct container state from within the newly added criContainer.GetStatus() in nri-api, by doing there a s.Runtime().UpdateContainerStatus() prior to a call to ctr.State() to acquire the state, but as I stated in my earlier message that resulted in catastrophic side-effects (which I did not understand the reasons of).

klihub: @bitoku Okay, I revisited that idea with a less dumb implementation, where we now only try to do an extra container state update on the StopContainer code path. I need to test it more carefully, but if we get with this the state correctly, then from my point of view we could go with this implementation.

klihub: @bitoku PTAL at the latest version, if you have cycles to spare.

bitoku: Hi @klihub Thank you for working on this.

At first, I meant we should separate these fetch parts into independent methods for `Runtime` and use them, because even just `UpdateContainerStatus` will change the status returned by `ListContainers` or some other CRI requests.

https://github.com/cri-o/cri-o/blob/bcc6495a9bf55bcbb16ff7d40776c096ca4bb7f7/internal/oci/runtime_oci.go#L1180-L1198

https://github.com/cri-o/cri-o/blob/29340189a38c2439a155ad835cdb4815b7bbbe6b/internal/oci/runtime_vm.go#L865-L882

but I'm starting to feel we may need to fix the design of cri-o, rather than just fix this.
I reconsidered the risk of race conditions, now I think we are ok with this for now.
I don't know what other maintainers think though.

bitoku:
/lgtm

I lgtm'd though I have a concern about a potential race condition.
https://github.com/cri-o/cri-o/pull/9706#discussion_r2699213367

@cri-o/cri-o-maintainers wdyt?

server/container_stop.go
klihub: Ok, let's hear what the verdict is by others, then decide how to go forward depending on that.

klihub:
> /lgtm
> 
> I lgtm'd though I have a concern about a potential race condition. [#9706 (comment)](https://github.com/cri-o/cri-o/pull/9706#discussion_r2699213367)
> 
> @cri-o/cri-o-maintainers wdyt?

@bitoku Just to better understand, is the race condition of your concern triggered by a `Runtime().UpdateContainerStatus()` being triggered for a container which has already exited and for which `postStopCleanup()` is being executed, but for which the final ContainerStateToDisk() there has not been run yet ?

bitoku:
> Just to better understand, is the race condition of your concern triggered by a Runtime().UpdateContainerStatus() being triggered for a container which has already exited and for which postStopCleanup() is being executed, but for which the final ContainerStateToDisk() there has not been run yet ?

Yes. Containers are stored as a pointer. When listing the container, this container pointers are returned.

https://github.com/cri-o/cri-o/blob/d2d7d1f2c49dc05a0e4b6299bb23e5fe3d113d89/internal/lib/container_server.go#L726

All container operations use this pointer. In `UpdateContainerStatus` it updates the container status directly via this pointer. If the list operation is called after `UpdateContainerStatus`, the updated container status will be returned. 

https://github.com/cri-o/cri-o/blob/d2d7d1f2c49dc05a0e4b6299bb23e5fe3d113d89/internal/lib/container_server.go#L752-L755

klihub:
> > Just to better understand, is the race condition of your concern triggered by a Runtime().UpdateContainerStatus() being triggered for a container which has already exited and for which postStopCleanup() is being executed, but for which the final ContainerStateToDisk() there has not been run yet ?
> 
> Yes. Containers are stored as a pointer. When listing the container, this container pointers are returned.
> 
> https://github.com/cri-o/cri-o/blob/d2d7d1f2c49dc05a0e4b6299bb23e5fe3d113d89/internal/lib/container_server.go#L726
> 
> All container operations use this pointer. In `UpdateContainerStatus` it updates the container status directly via this pointer. If the list operation is called after `UpdateContainerStatus`, the updated container status will be returned.
> 
> https://github.com/cri-o/cri-o/blob/d2d7d1f2c49dc05a0e4b6299bb23e5fe3d113d89/internal/lib/container_server.go#L752-L755

Okay, so doesn't that race condition already exist prior to this PR, and can be triggered, at least in  theory, if someone manages to hit CRI-O with a very unfortunately timed `crictl inspect $CONTAINER_ID` ?

haircommander:
yeah the case is really narrow, and only possible to see if there's some state that is being tracked between stops/starts like IRQ balance disablement. 

/approve

btw I don't love adding another UpdateContainerStatus to the stop flow but I can def see how funky things could happen. that was part of the rational of https://github.com/cri-o/cri-o/pull/9422/ but it didn't work in the way I wanted

Issue #9705, opened 2026-01-10
Upgrade of CRI-O from 1.34 to 1.35 breaks pod running systemd with `hostUsers: false`


adelton:
### What happened?

In https://github.com/freeipa/freeipa-container we run systemd in container. We test on numerous setups and runtimes, one of them being CRI-O. When I tried to update the test to use CRI-O 1.35 instead of 1.34, the tests started to fail.

I tried to minimize the reproducer of the issue on a fresh Ubuntu 24.04 VM, and upgraded to latest versions of all packages, and found the problem deterministic with CRI-O 1.35 and Kubernetes 1.34 and 1.35, both a fresh installation of CRI-O 1.35, and upgrade of CRI-O from 1.34 to 1.35. I saw the problem both with Kubernetes installed using `kubeadm init` which is what we will show bellow, and with RKE2 and K0s.

### What did you expect to happen?

I expected my pod
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-systemd
  labels:
    app: test-systemd
spec:
  restartPolicy: Always
  hostUsers: false
  containers:
  - name: test-systemd
    image: registry.access.redhat.com/ubi10/ubi-init
    imagePullPolicy: IfNotPresent
    securityContext:
      readOnlyRootFilesystem: true
    tty: true
```
to run on a Kubernetes node with CRI-O 1.35 just as well as it did with 1.34.

### How can we reproduce it (as minimally and precisely as possible)?

Have a fresh Ubuntu 24.04 VM with packages upgraded to latest versions:
```
sudo apt update
sudo apt upgrade -y
sudo reboot
```

We will install CRI-O 1.34:
```
export CRIO_VERSION=v1.34
curl -fsSL https://download.opensuse.org/repositories/isv:/cri-o:/stable:/$CRIO_VERSION/deb/Release.key \
        | gpg --dearmor | sudo tee /etc/apt/keyrings/cri-o-apt-keyring.gpg > /dev/null
echo "deb [signed-by=/etc/apt/keyrings/cri-o-apt-keyring.gpg] https://download.opensuse.org/repositories/isv:/cri-o:/stable:/$CRIO_VERSION/deb/ /" \
        | sudo tee /etc/apt/sources.list.d/cri-o.list

sudo apt update
sudo apt install -y cri-o

sudo cp /etc/cni/net.d/10-crio-bridge.conflist.disabled /etc/cni/net.d/10-crio-bridge.conflist
[...]

PR #9712, opened 2026-01-14
server: Always include UID/GID mappings for user namespace containers

saschagrunert:
#### What type of PR is this?

/kind bug

When containers join a user namespace via path, UID/GID mappings must still be included in the OCI spec. While the namespace path indicates which namespace to join, the mappings are essential for the runtime to properly handle file ownership operations, particularly cgroup delegation.

Without these mappings, runtimes cannot chown the cgroup directory to the mapped UID 0, causing systemd containers to fail with "Permission denied" when attempting to create cgroups like /init.scope.

This restores the behavior from v1.34.4 where both namespace path and mappings coexisted successfully.

Fixes https://github.com/cri-o/cri-o/issues/9705
None

```release-note
Fixed a regression in v1.35.0 where systemd containers with `hostUsers: false` (user namespaces enabled) would fail with "Permission denied" errors when systemd attempted to create cgroups.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure UID/GID mappings are consistently applied to containers using user namespaces.

* **Tests**
  * Added a cgroup integration test that verifies user namespace containers include UID/GID mappings and can access cgroup information (duplicate test case added).

### DIFF
--- a/test/cgroups.bats
+++ b/test/cgroups.bats
@@ -249,3 +249,68 @@ EOF
 	output=$(crictl inspect "$ctr_id" | jq -r '.status.state')
 	[[ "$output" == "CONTAINER_RUNNING" ]]
 }
+
+@test "user namespace containers include UID/GID mappings for cgroup delegation" {
+	if test -n "$CONTAINER_UID_MAPPINGS"; then
+		skip "userNS already enabled globally"
+	fi
+	if ! is_cgroup_v2; then
+		skip "test requires cgroup v2"
+	fi
+
+	start_crio
+
+	# Create a pod with user namespace enabled (hostUsers: false)
+	jq '	.linux.security_context.namespace_options.userns_options = {
+			"mode": 0,
+			"uids": [{
+				"host_id": 100000,
+				"container_id": 0,
+				"length": 65536
+			}],
+			"gids": [{
+				"host_id": 100000,
+				"container_id": 0,
+				"length": 65536
+			}]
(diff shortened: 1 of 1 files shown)


bitoku:
Reverting this change https://github.com/cri-o/cri-o/commit/c2ebfbac1e57716355cfe0eb4b886acb04cd3c01#diff-1ec6bb9cf833fd4ded46ce1feb6440f1d7532294e0445623f2f2e0bb2726ea63L1451-L1456

We want the test cases for systemd container. I tried once and unfortunately the exisiting images in quay.io/cri-o can't be used.
Probably https://hub.docker.com/r/redhat/ubi10-init can be used.

saschagrunert:
I added a testcase in 0f68aa8d072a1f0c5252fa1d4579ee1d2e842e66

adelton:
@bitoku While I don't work with CRI-O directly, just through various Kubernetes distributions, I've been running systemd-based containers and GitHub Actions in https://github.com/freeipa/freeipa-container for years. Let me know if I can help getting that high-level use case (for example with `registry.access.redhat.com/ubi10/ubi-init` as we've shown in https://github.com/cri-o/cri-o/issues/9705#issue-3800314093) added.

bitoku:
@adelton 
Thank you for the cooperation!
For now I'm thinking just adding test to launch a systemd container as a starting step, but we may reach out to you in the future.

bitoku:
/lgtm

The fix was confirmed though we still need the continuous way to test it.
https://github.com/cri-o/cri-o/issues/9705#issuecomment-3749173088

PR #9714, opened 2026-01-15
cgroup v2: automatically set memory.high to prevent OOM on Fedora CoreOS

saschagrunert:
#### What type of PR is this?

/kind failing-test

The Kubernetes e2e test "should report Memory pressure in PSI metrics" fails consistently on Fedora CoreOS with CRI-O since September 2024, while passing on Container-Optimized OS with containerd and Ubuntu with CRI-O.

Test reference:
https://github.com/kubernetes/kubernetes/blob/0ba578f91f5de11776152b55bac37491d9848ef3/test/e2e_node/summary_test.go#L407-L444

The test creates containers with a 200MB memory limit and runs a workload that writes 50MB files in a loop, expecting the kernel to reclaim file cache and generate sustained PSI (Pressure Stall Information) metrics.

Analysis of test logs shows identical behavior on both runtimes:
- Both use the same OCI spec: memory.max=200000000, no memory.high
- Containerd on COS 125: container runs for 31 seconds, PSI metrics work
- CRI-O on Fedora CoreOS 43: container OOM killed in 0.5 seconds

This is not a CRI-O vs containerd difference. Both runtimes use identical OCI specs. The issue is that different kernels have different page cache reclaim policies. Without memory.high set, containers writing files can hit memory.max immediately on Fedora CoreOS before the kernel reclaims page cache, causing instant OOM kills.

In cgroup v2, memory.max is a hard limit that triggers OOM kills, while memory.high is a soft limit that triggers aggressive memory reclaim before reaching memory.max. Setting memory.high enables the kernel to reclaim page cache and generate PSI pressure metrics instead of OOM killing.

This change automatically sets memory.high to 95% of memory.max when:
- cgroup v2 is in use
- memory_limit_in_bytes is set
- memory.high is not already configured via the Unified field

This ensures kernel memory reclaim happens before hitting the hard limit, containers survive on Fedora CoreOS, and PSI metrics are generated. User configurations are always respected.

Added unit tests to verify automatic memory.high setting, user-provided value preservation, zero-limit handling, and swap limit interaction.

Refers to: https://github.com/kubernetes/kubernetes/issues/134141
Related: https://github.com/kubernetes/kubernetes/pull/136191
Related: https://github.com/kubernetes/test-infra/pull/36234

None

```release-note
CRI-O now automatically sets memory.high to 95% of memory.max for all containers with memory limits on cgroup v2. This enables kernel memory reclaim before hitting the hard limit, preventing OOM kills on systems with less aggressive page cache reclamation (like Fedora CoreOS) and enabling PSI pressure metrics. User-configured memory.high values are always respected.
```

## Summary by CodeRabbit

* **New Features**
  * Automatic memory.high configuration for cgroup v2: When a memory limit is set, memory.high is automatically configured to 95% of the limit if not explicitly provided.

* **Tests**
  * Added test cases for automatic memory.high handling, covering scenarios with memory limits, swap limits, and user-provided configurations.

### DIFF
--- a/internal/factory/container/container.go
+++ b/internal/factory/container/container.go
@@ -869,6 +869,10 @@ func (c *container) SpecSetLinuxContainerResources(resources *types.LinuxContain
 
 		specgen.SetLinuxResourcesMemoryLimit(memoryLimit)
 
+		// Save the original memory limit for memory.high calculation before it gets
+		// potentially overwritten by swap limit below
+		memoryMax := memoryLimit
+
 		if resources.GetMemorySwapLimitInBytes() != 0 {
 			if resources.GetMemorySwapLimitInBytes() > 0 && resources.GetMemorySwapLimitInBytes() < resources.GetMemoryLimitInBytes() {
 				return fmt.Errorf(
@@ -885,6 +889,27 @@ func (c *container) SpecSetLinuxContainerResources(resources *types.LinuxContain
 		// otherwise the container creation fails.
 		if node.CgroupHasMemorySwap() {
 			specgen.SetLinuxResourcesMemorySwap(memoryLimit)
+		}
+
+		// For cgroup v2, automatically set memory.high to enable memory reclaim
+		// and PSI pressure metrics generation before hitting the hard limit (memory.max).
+		// This mitigates OOM kills on systems like Fedora CoreOS where file cache
+		// reclaim is less aggressive, allowing containers to generate sustained
+		// memory pressure metrics instead of immediate OOM.
+		// Only set if not explicitly configured via Unified field.
+		if node.CgroupIsV2() {
+			// Check if memory.high is already set in unified cgroup settings
+			_, alreadySet := resources.GetUnified()["memory.high"]
+			if !alreadySet {
+				// Set memory.high to 95% of memory.max to trigger reclaim before OOM
+				memoryHigh := int64(float64(memoryMax) * 0.95)
+
+				if specgen.Config.Linux.Resources.Unified == nil {
+					specgen.Config.Linux.Resources.Unified = make(map[string]string)
+				}
+
+				specgen.Config.Linux.Resources.Unified["memory.high"] = strconv.FormatInt(memoryHigh, 10)
+			}
 		}
 	}
 
--- a/internal/factory/container/container_test.go
+++ b/internal/factory/container/container_test.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/cri-o/cri-o/internal/annotations"
 	"github.com/cri-o/cri-o/internal/config/capabilities"
+	"github.com/cri-o/cri-o/internal/config/node"
 	"github.com/cri-o/cri-o/internal/hostport"
 	"github.com/cri-o/cri-o/internal/lib/constants"
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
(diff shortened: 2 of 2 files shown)


bitoku:
It may have a risk of livelock according to this.
https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2570-memory-qos#latest-update-stalled

Also it doesn't sound right to me to add a new logic just to pass a test.

Can't we change the page cache policy on OS side by setting https://www.kernel.org/doc/html/v6.17/admin-guide/sysctl/vm.html#vfs-cache-pressure or something else?

saschagrunert:
Thanks for raising these concerns.

> It may have a risk of livelock according to this. https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2570-memory-qos#latest-update-stalled
> 
> Also it doesn't sound right to me to add a new logic just to pass a test.
> 
> Can't we change the page cache policy on OS side by setting https://www.kernel.org/doc/html/v6.17/admin-guide/sysctl/vm.html#vfs-cache-pressure or something else?

On the OS configuration approach: `vfs_cache_pressure` and similar sysctls are global settings affecting all containers, but this problem happens at the per-cgroup level. When a container approaches `memory.max` without `memory.high` set, the next allocation hits the hard limit and triggers OOM before the kernel can reclaim page cache. There's no global kernel parameter that triggers aggressive reclaim when individual cgroups approach their limits. This is exactly what `memory.high` was designed for in cgroup v2, giving the kernel advance notice to start reclaiming before hitting the hard limit.

Different kernels have different page cache reclaim aggressiveness (why the same workload OOMs on Fedora CoreOS but not COS), but there's no way to tune this at the OS level to solve per-container limits.

On the livelock risk: yes, the Memory QoS KEP documents this for processes that allocate memory faster than the kernel can reclaim it. But the current state without `memory.high` is strictly worse (instant OOM). The 95% threshold provides 5% headroom for allocation spikes during reclaim. Complete livelock only happens if a process consumes memory faster than the kernel can reclaim it AND faster than it can allocate through that buffer. Most real workloads doing file I/O don't exhibit this pathological behavior.

This isn't just about passing a test. Any container that writes files and relies on page cache on kernels with less aggressive reclaim (like Fedora CoreOS) will hit instant OOM when it could survive with proper memory reclaim. The test exposed a real kernel behavior difference that impacts production workloads.

If livelock becomes a practical problem, we can make this configurable or adjust the threshold. But instant OOM for file I/O workloads is definitively broken.

bitoku:
Thank you for the explanation.

> This isn't just about passing a test. Any container that writes files and relies on page cache on kernels with less aggressive reclaim (like Fedora CoreOS) will hit instant OOM when it could survive with proper memory reclaim. The test exposed a real kernel behavior difference that impacts production workloads.

If we think this is really a problematic behavior, we should reach out to fedora coreos team, especially because coreos is container-centric OS. 

And what OS users use is up to themselves. They may want less aggressive reclaim behavior. We shouldn't tinker around.
If they don't, they can choose other OS that has more aggressive reclaim behavior.

bitoku:
I think we haven't figured out "why the same workload OOMs on Fedora CoreOS but not COS", have we?
Figuring it out is the path forward I'd say.

saschagrunert:
> If we think this is really a problematic behavior, we should reach out to fedora coreos team, especially because coreos is container-centric OS.
>
> And what OS users use is up to themselves. They may want less aggressive reclaim behavior. We shouldn't tinker around.

You're right that we should reach out to the Fedora CoreOS team. I checked their tracker and there's no issue about this yet, so I'll file one.

_Edit: https://github.com/coreos/fedora-coreos-tracker/issues/2094_

I want to clarify this isn't really "tinkering" though. In cgroup v2, **memory.high** triggers aggressive reclaim and throttling when exceeded, while **memory.max** is the hard limit that triggers OOM kills. Without memory.high, the kernel only reclaims when approaching memory.max, which on Fedora CoreOS isn't fast enough for file-heavy workloads to generate sustained PSI metrics. Kubernetes has the [MemoryQoS feature](https://kubernetes.io/blog/2023/05/05/qos-memory-resources/) that does exactly this (sets memory.high to 90% of memory.max), but it's been stuck in alpha for 2+ years.

We do respect user configuration. If someone explicitly sets memory.high via the `io.kubernetes.cri-o.UnifiedCgroup` annotation, we don't override it. That said, we should make this configurable with an opt-out flag and maybe only apply it for Burstable/BestEffort QoS classes like Kubernetes does.

---

> I think we haven't figured out "why the same workload OOMs on Fedora CoreOS but not COS", have we?
> Figuring it out is the path forward I'd say.

The test passes on COS with containerd but fails on Fedora CoreOS with CRI-O. Neither runtime sets memory.high automatically, so it's something about the OS/kernel.

COS made [fixes for workloads with heavy page cache usage](https://cloud.google.com/container-optimized-os/docs/release-notes/m89) around June 2021, and they use different kernel versions and configs than Fedora. The exact difference is hard to pin down without digging through kernel sources, but it's likely related to how aggressively they reclaim page cache under memory pressure.

The underlying issue is that [with cgroups, the kernel doesn't have good reclaim logic for clean cache](https://storpool.com/blog/the-weird-interactions-of-cgroups-and-linux-page-cache-in-hypervisor-environments), so it prefers OOM kills. This is a known limitation.

Setting memory.high to 95% of memory.max gives the kernel a trigger point to start aggressive reclaim before hitting the hard limit. This works consistently across both systems.

haircommander:
I know I've pushed back on turning on memoryQOS for this bug in upstream kube, but maybe we need to push forward memoryQOS to beta. now that we have PSI metrics, the kubelet could use consistent memory pressure on the cgroup to trigger evictions, rather than letting containers sit at memory.high forever. I have heard from kernel devs that they want us to use memory.high generally, and I think we may have the tools now to prevent the livelock

mariafromano-25:
Yes, PSI is currently in beta and I am working on promoting it to GA for 1.36. 
Also, thank you @saschagrunert for your thorough explanations, I'm learning a lot!

saschagrunert:
> I know I've pushed back on turning on memoryQOS for this bug in upstream kube, but maybe we need to push forward memoryQOS to beta. now that we have PSI metrics, the kubelet could use consistent memory pressure on the cgroup to trigger evictions, rather than letting containers sit at memory.high forever. I have heard from kernel devs that they want us to use memory.high generally, and I think we may have the tools now to prevent the livelock

/hold

All of this makes sense. The question is what do we want to do to make the serial tests green sooner than later: https://testgrid.k8s.io/sig-node-cri-o#node-kubelet-serial-crio

I'm not hesitant to any follow-ups, (mid-term) graduations or even reverts when it does not turn out to be correct. Right now it looks like that the _right path_ leads into a long term investment which has a high risk of becoming stale. 

---

I'm working since months over various k/k, CRI-O and test-infra PRs to get this dashboard green and we're closer than ever. If we wait longer, we risk to get more failing tests or flakes without even knowing it. 

Should we disable the upstream test case for now? See: https://github.com/kubernetes/kubernetes/pull/136191

bitoku:
> Should we disable the upstream test case for now?

+1 if the community approves it.

saschagrunert:
Closing in favor of the CoreOS tracker.

PR #9719, opened 2026-01-20
Fix the bug where cri-o doesn't emit any metrics when all is set.

bitoku:
#### What type of PR is this?

/kind bug

This PR fixes a bug where CRI-O didn't return all metrics when "all" is set.

```release-note
Fixed a bug where CRI-O didn't return all metrics when "all" is set.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable handling of the special "all" metric setting to prevent unexpected gaps.
  * Fixed inconsistent selection so pod- and container-level network/metric collection align.

* **Changes**
  * Metric selection is unified across the system via a stable enabled-metrics accessor.
  * A canonical list of available metrics was added; descriptor output and collected metrics may change.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -92,8 +92,9 @@ const (
 	PressureMetrics = "pressure"
 )
 
+// AvailableMetrics is a list of all available metrics that can be included in stats.
+// It excludes the "all" metric, which is a special value that includes all other metrics.
 var AvailableMetrics = []string{
-	AllMetrics,
 	CPUMetrics,
 	DiskMetrics,
 	DiskIOMetrics,
@@ -804,7 +805,14 @@ type StatsConfig struct {
 
 	// IncludedPodMetrics specifies the list of metrics to include when collecting pod metrics.
 	// If "all" is specified, all metrics are included. In that case, "all" should be the only element.
+	//
+	// Deprecated: Use this field only when the user input config is needed because it's not formalized.
+	// Use EnabledPodMetrics() instead.
 	IncludedPodMetrics []string `toml:"included_pod_metrics"`
+
+	// includedPodMetrics is an internal representation of IncludedPodMetrics.
+	// It doesn't contain "all".
+	includedPodMetrics []string
 }
 
 // tomlConfig is another way of looking at a Config, which is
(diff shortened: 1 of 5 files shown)


bitoku:
@cri-o/cri-o-maintainers PTAL

pkg/config/config.go
```diff
@@ -2213,8 +2218,14 @@ func (c *Config) SetSingleConfigPath(singleConfigPath string) {
 }
 
 func (c *StatsConfig) Validate() error {
+	if len(c.IncludedPodMetrics) == 1 && c.IncludedPodMetrics[0] == AllMetrics {
+		c.includedPodMetrics = AvailableMetrics
```
saschagrunert, line 2272: Should we filter out `AllMetrics` here?

bitoku: thanks, good catch!

bitoku: I removed "all" from AvailableMetrics. It's only used for this `Validate` function anyway.

pkg/config/config.go
```diff
@@ -788,7 +789,12 @@ type StatsConfig struct {
 
 	// IncludedPodMetrics specifies the list of metrics to include when collecting pod metrics.
 	// If "all" is specified, all metrics are included. In that case, "all" should be the only element.
+	// Do NOT use this field directly. Use Metrics() instead.
 	IncludedPodMetrics []string `toml:"included_pod_metrics"`
```
saschagrunert: I think we should depreciate it to make linters aware.

saschagrunert (approved):
Just one non-blocking nit, otherwise LGTM

/hold

Feel free to unhold if you think we should not deprecate the config field.

pkg/config/config.go
```diff
@@ -2223,5 +2235,11 @@ func (c *StatsConfig) Validate() error {
 		}
 	}
 
+	c.includedPodMetrics = c.IncludedPodMetrics
+
 	return nil
 }
+
+func (c *StatsConfig) Metrics() []string {
```
haircommander: Maybe EnabledPodMetrics or something? I wonder if Metrics is too generic, there is a whole metrics table in the config

bitoku: I was looking for a better naming. Thanks!

pkg/config/config.go
```diff
@@ -2223,5 +2237,11 @@ func (c *StatsConfig) Validate() error {
 		}
 	}
 
+	c.includedPodMetrics = c.IncludedPodMetrics
+
 	return nil
 }
+
+func (c *StatsConfig) EnabledMetrics() []string {
```
haircommander: I still feel including pod is good here, cri-o has server metrics it exposes which are different than these

bitoku: It slipped my eyes. Thanks.

bitoku:
/unhold

@cri-o/cri-o-maintainers PTAL

bitoku:
@cri-o/cri-o-maintainers PTAL

bitoku:
@cri-o/cri-o-maintainers PTAL

bitoku:
sorry rebase removed lgtm, PTAL again @cri-o/cri-o-maintainers

PR #9723, opened 2026-01-22
Feature Request: Make TLS minimum version and cipher suites configura…

### LINKED ISSUE #9721: Feature Request: Make TLS minimum version and cipher suites configurable for CRI-O server
**Summary**

CRI-O currently hardcodes the TLS minimum version to TLS 1.2 in its server implementation. This prevents operators from enforcing TLS 1.3 as the minimum version and from controlling allowed cipher suites.

This issue proposes adding configuration options to CRI-O to make the following TLS settings configurable:

- Minimum TLS version (minTLSVersion)
- Allowed TLS cipher suites (cipherSuites)
Providing these controls would allow CRI-O to align with modern security requirements and with other Kubernetes node components.

**Why This Is Needed**
1. Post-Quantum Cryptography (PQC) Readiness

TLS 1.3 is a foundational requirement for introducing post-quantum or hybrid key exchange mechanisms in the future. While CRI-O is not expected to implement PQC algorithms today, hardcoding TLS 1.2 prevents environments from being PQC-ready when needed.

2. Security & Compliance Requirements

Several security standards and compliance frameworks are moving toward or already recommend TLS 1.3 as the minimum supported version, including:

- Modern FIPS guidance
- CNSA 2.0
- FedRAMP and similar government profiles
Allowing operators to configure the minimum TLS version is necessary to meet these requirements.

3. Parity with Kubelet and Kubernetes Components

The kubelet already supports configurable TLS settings such as:

- tlsMinVersion
- tlsCipherSuites

CRI-O currently lacks equivalent configurability, making it difficult for cluster operators to apply consistent TLS policies across node components.

4. Container Platform Requirements

Container platforms (for example, OpenShift) rely on being able to centrally configure TLS policies across all node-level components. The lack of TLS configurability in CRI-O becomes a blocker for enforcing platform-wide security profiles.

**Current Behavior**

TLS minimum version is hardcoded to TLS 1.2 in CRI-O server code
Cipher suites are not configurable
Users cannot enforce TLS 1.3 for:
           CRI-O streaming server
           Metrics and other TLS endpoints

**Proposed Behavior**

Introduce CRI-O configuration options to control TLS behavior:
[crio.tls]
min_tls_version = "1.3"   # Allowed values: "1.2", "1.3"
cipher_suites = []        # Optional list of allowed cipher suites

**Defaults**

min_tls_version: "1.2" (maintains backward compatibility)
cipher_suites: empty (use Go’s secure defaults)

This ensures no behavior change for existing deployments unless explicitly configured.

**Scope**

- Applies to CRI-O TLS servers (e.g., streaming, metrics)
- Does not introduce new cryptographic algorithms
- Does not change default TLS behavior
- Focuses solely on configurability

**Benefits**

- Enables TLS 1.3 enforcement where required
- Aligns CRI-O with Kubernetes security configurability
- Unblocks platform-level TLS policy enforcement
- Prepares CRI-O for future PQC and hybrid crypto adoption

**I’m willing to work on this change, including:**

- Adding configuration fields
- Updating TLS server initialization
- Providing tests and documentation updates

asahay19:
#### What type of PR is this?

/kind feature

This PR adds configurable TLS minimum version settings for CRI-O's streaming server and metrics server.

**Background:**
Currently, CRI-O hardcodes `tls.VersionTLS12` in both the streaming server (`server/server.go`) and the metrics server (`server/metrics/metrics.go`). This prevents users from configuring TLS 1.3, which is a prerequisite for Post-Quantum Cryptography (PQC) readiness.

**Changes:**
- Add `tls_min_version` and `tls_cipher_suites` to `[crio.api]` config
- Use `k8s.io/component-base/cli/flag` for TLS parsing (consistent with Kubernetes)
- Validation done during `config.Validate()`, getters return pre-validated values
- Both streaming and metrics servers use the same TLS configuration
- Default TLS version is `VersionTLS12`
- Added integration tests in test/tls_config.bats

Fixes #9721

- The default TLS version remains 1.2, so this change is backward compatible
- TLS 1.3 cipher suites are managed automatically by Go, so `tls_cipher_suites` only applies to TLS 1.2
- This is part of the OpenShift PQC (Post-Quantum Cryptography) readiness initiative

**Files changed:**
1. `pkg/config/config.go` - Added config fields, validation, and helper functions
2. `pkg/config/template.go` - Added template strings for crio.conf documentation
3. `server/server.go` - Use config value instead of hardcoded `tls.VersionTLS12`
4. `server/metrics/metrics.go` - Use config value instead of hardcoded `tls.VersionTLS12`
5. `pkg/config/config_test.go` -  Added unit tests for TLS configuration 
           - TLSMinVersion parsing and validation tests
           - TLSCipherSuites parsing and validation tests  
           - Combined TLSMinVersion + CipherSuites behavior tests
6. `test/tls_config.bats` - Added 6 Integration test cases.

**Output** 
I tested all the Unit and Integration tests manually . All got passed.
Pasting Output for Integration Tests:
```
asahay@asahay-ubuntu:~/crio-test/cri-o$ sudo -E ./test/test_runner.sh tls_config.bats
warning: preserving the entire environment is not supported, `-E` is ignored
+ TEST_USERNS=
+ TEST_KEEP_ON_FAILURE=
+ '[' -n '' ']'
+++ readlink -f /home/asahay/crio-test/cri-o/test/test_runner.sh
++ dirname /home/asahay/crio-test/cri-o/test/test_runner.sh
+ cd /home/asahay/crio-test/cri-o/test
+ [[ '' == \1 ]]
...........................................
++ bats --jobs 2 --tap tls_config.bats --filter-tags !crio:serial
+ bats --jobs 2 --tap tls_config.bats --filter-tags '!crio:serial'
1..6
ok 1 tls_min_version: TLS 1.3 server should accept TLS 1.3 connections
ok 2 tls_min_version: TLS 1.3 server should reject TLS 1.2 connections
ok 3 tls_min_version: TLS 1.2 server should accept TLS 1.3 connections
ok 4 tls_min_version: TLS 1.2 server should accept TLS 1.2 connections
ok 5 tls_min_version: default (empty) should behave as TLS 1.2
ok 6 tls_min_version: invalid version should fail validation

real	1m3.437s
user	1m24.987s
sys	0m15.983s
+ execute bats --tap tls_config.bats --filter-tags crio:serial
+ echo ++ bats --tap tls_config.bats --filter-tags crio:serial
++ bats --tap tls_config.bats --filter-tags crio:serial
+ bats --tap tls_config.bats --filter-tags crio:serial
1..0

real	0m0.262s
user	0m0.142s
sys	0m0.086s
```

Added `tls_min_version` and `tls_cipher_suites` options to the `[crio.api]` section in crio.conf. These settings apply to both the streaming server and the metrics server. This allows users to configure TLS 1.3 for enhanced security and PQC readiness. The default remains TLS 1.2 for backward compatibility.

```release-note
Added `tls_min_version` and `tls_cipher_suites` configuration options to `[crio.api]` for configuring TLS settings on streaming and metrics servers. Supports TLS 1.2 (default) and TLS 1.3.
```

## Summary by CodeRabbit

* **New Features**
  * Configurable API TLS minimum version (default TLS1.2) and optional cipher-suite list; streaming and metrics endpoints now honor these settings. Configuration templates expose TLS options. Vendored CLI adds richer shell completions (Bash/Fish/PowerShell/Zsh), active-help, flag-grouping and enhanced flag/command helpers.

* **Bug Fixes / Validation**
  * TLS version and cipher-suite inputs validated at startup; invalid values cause startup failure. Cipher-suite settings ignored for TLS1.3 with a warning.

* **Tests**
  * Added unit and integration tests for TLS parsing, defaults, and cipher-suite behavior.


pkg/config/config.go
```diff
@@ -1206,6 +1224,20 @@ func (c *APIConfig) Validate(onExecution bool) error {
 		}
 	}
 
+	// Validate TLS version if set
+	if c.TLSMinVersion != "" {
```
ngopalak-redhat: @asahay19  Please add the unit tests for all the code changes

pkg/config/config.go
```diff
@@ -735,6 +736,15 @@ type APIConfig struct {
 
 	// StreamIdleTimeout is how long to leave idle connections open for
 	StreamIdleTimeout string `toml:"stream_idle_timeout"`
+
+	// TLSMinVersion is the minimum TLS version for the streaming and other TLS servers.
+	// Valid values are: "1.2" and "1.3".
+	TLSMinVersion string `toml:"tls_min_version"`
```
ngopalak-redhat, line 744: I prefer keeping the values in sync with what Kubernetes uses here: https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/. Something like VersionTLS12. However, I am open to suggestions from other reviewers in CRI-O.

pkg/config/config.go
asahay19: Added unit test cases for all the code changes

bitoku:
/ok-to-test

@asahay19 Can you modify the PR description to follow the template?

pkg/config/config.go
```diff
@@ -735,6 +736,16 @@ type APIConfig struct {
 
 	// StreamIdleTimeout is how long to leave idle connections open for
 	StreamIdleTimeout string `toml:"stream_idle_timeout"`
+
+	// TLSMinVersion is the minimum TLS version for the streaming and other TLS servers.
+	// Valid values are: "VersionTLS12" and "VersionTLS13" (matching Kubernetes conventions).
+	// Default is "VersionTLS12".
+	TLSMinVersion string `toml:"tls_min_version"`
```
bitoku, line 744:
```suggestion
	StreamTLSMinVersion string `toml:"stream_tls_min_version"`
```

pkg/config/config.go
```diff
@@ -735,6 +736,16 @@ type APIConfig struct {
...
+
+	// TLSCipherSuites is the list of cipher suites to use for TLS 1.2.
+	// If empty, the Go default cipher suites are used.
+	// This has no effect on TLS 1.3 as Go manages cipher suites automatically for TLS 1.3.
+	TLSCipherSuites []string `toml:"tls_cipher_suites"`
```
bitoku, line 749:
```suggestion
	StreamTLSCipherSuites []string `toml:"stream_tls_cipher_suites"`
```

pkg/config/config.go
```diff
@@ -2225,3 +2259,86 @@ func (c *StatsConfig) Validate() error {
...
+
+// GetTLSMinVersion returns the parsed TLS minimum version for the API config.
+// It returns the default TLS 1.2 if not configured.
+func (c *APIConfig) GetTLSMinVersion() (uint16, error) {
+	return ParseTLSVersion(c.TLSMinVersion)
+}
+
+// GetTLSCipherSuites returns the TLS cipher suites for the API config.
+// Returns nil if no cipher suites are configured (uses Go defaults).
+func (c *APIConfig) GetTLSCipherSuites() ([]uint16, error) {
```
bitoku: Probably we can leverage k8s.io/component-base for this and other functions too.

https://github.com/kubernetes/component-base/blob/master/cli/flag/ciphersuites_flag.go

server/metrics/metrics.go
```diff
@@ -513,9 +513,14 @@ func (m *Metrics) startEndpoint(
...
 
+			tlsMinVersion, err := m.config.GetMetricsTLSMinVersion()
+			if err != nil {
+				log.Fatalf(ctx, "Getting metrics TLS min version: %v", err)
+			}
+
 			srv.TLSConfig = &tls.Config{
 				GetConfigForClient: cc.GetConfigForClient,
-				MinVersion:         tls.VersionTLS12,
+				MinVersion:         tlsMinVersion,
```
bitoku: It lacks ciphersuites?

server/server.go
```diff
@@ -545,13 +545,33 @@ func New(
...
+
+		tlsCipherSuites, err := config.GetTLSCipherSuites()
+		if err != nil {
+			return nil, fmt.Errorf("get TLS cipher suites: %w", err)
+		}
+
+		// Warn if cipher suites are configured with TLS 1.3 (they have no effect)
+		if tlsMinVersion == tls.VersionTLS13 && len(config.TLSCipherSuites) > 0 {
+			log.Warnf(ctx, "tls_cipher_suites configuration is ignored when tls_min_version is 1.3 (Go manages TLS 1.3 cipher suites automatically)")
+		}
```
bitoku: Ideally, we want to get the validations done beforehand, and can just use the validated values here.

pkg/config/config.go
```diff
@@ -760,6 +771,11 @@ type MetricsConfig struct {
 
 	// MetricsKey is the certificate key for the secure metrics server.
 	MetricsKey string `toml:"metrics_key"`
+
+	// MetricsTLSMinVersion is the minimum TLS version for the metrics server.
+	// Valid values are: "VersionTLS12" and "VersionTLS13" (matching Kubernetes conventions).
+	// Default is "VersionTLS12".
+	MetricsTLSMinVersion string `toml:"metrics_tls_min_version"`
```
haircommander: are there any reasons to have a different min version for metrics vs stream? It seems like we could embed TLSMinVersion in the API config and then just use it here, rather than having duplicate fields that users have to remember to set both for

haircommander: if we do have both (which I don't think we should) we should probably have both cipher suites as well

pkg/config/config.go
```diff
@@ -2225,3 +2259,86 @@ func (c *StatsConfig) Validate() error {
...
+			return fmt.Errorf("validating metrics_tls_min_version: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// ValidTLSVersions contains the valid TLS version strings.
+// These match Kubernetes naming conventions (e.g., kubelet tlsMinVersion).
+var ValidTLSVersions = []string{"VersionTLS12", "VersionTLS13"}
```
haircommander: might be good to define these as consts
```
const (
 versionTLS12 = "VersionTLS12"
 versionTLS13 = "VersionTLS13"
)
```
and then use those variables for the strings instead

server/server.go
```diff
@@ -545,13 +545,33 @@ func New(
...
 			GetConfigForClient: certConf.GetConfigForClient,
 			Certificates:       []tls.Certificate{certificate},
-			MinVersion:         tls.VersionTLS12,
+			MinVersion:         tlsMinVersion,
+			CipherSuites:       tlsCipherSuites,
 		}
 
-		log.Debugf(ctx, "Applying stream server TLS configuration")
+		tlsVersionStr := config.TLSMinVersion
+		if tlsVersionStr == "" {
```
haircommander: I don't think we should get here with empty string. the config pkg should make sure it's populated

pkg/config/config.go
asahay19: Yeah, even I think it is better that we go with one config field that applies to both servers.

pkg/config/config.go
asahay19: as per @haircommander's feedback,I consolidated to a single TLSMinVersion that applies to both servers, so the Stream prefix is not needed now.

pkg/config/config.go
asahay19: same as above

pkg/config/config.go
asahay19: This I addressed since I now use k8s.io/component-base/cli/flag.TLSVersion() which handles validation internally. This is according to Ayato's comment .

vendor/github.com/inconshreveable/mousetrap/README.md
```diff
@@ -0,0 +1,23 @@
+# mousetrap
+
+mousetrap is a tiny library that answers a single question.
+
+On a Windows machine, was the process invoked by someone double clicking on
+the executable file while browsing in explorer?
```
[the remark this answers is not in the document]
asahay19: This is a vendor third-party file (github.com/inconshreveable/mousetrap), not part of this PR's changes. It was automatically added by go mod vendor. We shouldn't modify vendor files

pkg/config/config.go
```diff
@@ -2225,3 +2267,24 @@ func (c *StatsConfig) Validate() error {
 
 	return nil
 }
+
+// Validate is the main entry point for metrics configuration validation.
+func (c *MetricsConfig) Validate() error {
```
bitoku: we can remove this

pkg/config/config.go
```diff
@@ -1206,6 +1227,27 @@ func (c *APIConfig) Validate(onExecution bool) error {
...
+	// Parse and validate TLS version using Kubernetes component-base
+	tlsVersion, err := cliflag.TLSVersion(c.TLSMinVersion)
+	if err != nil {
+		return fmt.Errorf("validating tls_min_version: %w", err)
+	}
+	c.tlsMinVersionParsed = tlsVersion
+
+	// Parse and validate TLS cipher suites using Kubernetes component-base
+	// Note: TLS 1.3 cipher suites are managed by Go automatically
+	if tlsVersion < tls.VersionTLS13 && len(c.TLSCipherSuites) > 0 {
```
bitoku: we should error when other than tls1.2 or tls1.3 is used.

pkg/config/template.go
```diff
@@ -878,6 +888,22 @@ const templateStringCrioAPIStreamTLSCa = `# Path to the x509 CA(s) file used to
...
+# Valid values are: "VersionTLS12" and "VersionTLS13" (matching Kubernetes conventions).
+# Default is "VersionTLS12".
+{{ $.Comment }}tls_min_version = "{{ .TLSMinVersion }}"
+
+`
+
+const templateStringCrioAPITLSCipherSuites = `# List of cipher suites to use for TLS 1.2.
+# If empty, the Go default cipher suites are used.
+# This has no effect on TLS 1.3 as Go manages cipher suites automatically.
+# Example cipher suites: TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
```
bitoku: nit: The document that looks like https://github.com/kubernetes/kubernetes/blob/7002ccc892a0c54a0739d297c77e9a4a0ce471d1/cmd/kubelet/app/options/options.go#L412-L417 may be better, but non blocking.

server/server.go
```diff
@@ -545,13 +545,19 @@ func New(
...
 		streamServerConfig.TLSConfig = &tls.Config{
 			GetConfigForClient: certConf.GetConfigForClient,
 			Certificates:       []tls.Certificate{certificate},
-			MinVersion:         tls.VersionTLS12,
+			MinVersion:         config.GetTLSMinVersion(),
+			CipherSuites:       config.GetTLSCipherSuites(),
 		}
 
-		log.Debugf(ctx, "Applying stream server TLS configuration")
+		log.Debugf(ctx, "Applying stream server TLS configuration with min version %s", config.TLSMinVersion)
```
bitoku: nit: This may be confusing and redundant.
If it works properly, then users can check min version by checking tls_min_version in the config (it's logged when debug log enabled).
If there's a bug when converting `TLSMinVersion(string)` to `tlsMinVersionParsed(uint16)`, the log says wrong and not reliable.

bitoku (commented):
lgtm overall. some small minor comments.
I want integration tests for it.

bitoku (changes requested):
It needs change in cert.go as well. 
In internal/cert/cert.go, the GetConfigForClient() function returns a tls.Config without MinVersion set. This overrides the MinVersion we set in server.go.

https://github.com/cri-o/cri-o/blob/740ea78286cbd4206dd67164ec140082adf50813/internal/cert/cert.go#L29-L34

server/server.go
```diff
@@ -545,10 +545,16 @@ func New(
 			return nil, fmt.Errorf("load stream server x509 key pair: %w", err)
 		}
 
+		// Warn if cipher suites are configured with TLS 1.3 (they have no effect)
+		if config.GetTLSMinVersion() == tls.VersionTLS13 && len(config.TLSCipherSuites) > 0 {
```
haircommander: this should only be emitted once at config validation IMO, instead of also twice more with server and metrics

pkg/config/config_test.go
```diff
@@ -1944,4 +1944,93 @@ var _ = t.Describe("Config", func() {
...
+
+		It("should fail validation for deprecated TLS 1.1", func() {
+			sut.TLSMinVersion = "VersionTLS11"
+			Expect(sut.APIConfig.Validate(false)).ToNot(Succeed())
+		})
+
+		It("should return correct version from GetTLSMinVersion after Validate", func() {
+			sut.TLSMinVersion = "VersionTLS13"
+			Expect(sut.APIConfig.Validate(false)).To(Succeed())
+			Expect(sut.APIConfig.GetTLSMinVersion()).To(Equal(uint16(0x0304)))
```
bitoku: nit:
This can be combined to the first unit test.
```suggestion
			Expect(sut.APIConfig.GetTLSMinVersion()).To(Equal(tls.VersionTLS13)))
```

pkg/config/config_test.go
```diff
@@ -1944,4 +1944,93 @@ var _ = t.Describe("Config", func() {
...
+		})
+
+		It("should default to TLS 1.2 when empty", func() {
+			sut.TLSMinVersion = ""
+			Expect(sut.APIConfig.Validate(false)).To(Succeed())
+			Expect(sut.APIConfig.GetTLSMinVersion()).To(Equal(uint16(0x0303)))
```
bitoku: This can be combined to the first unit test.
```suggestion
			Expect(sut.APIConfig.GetTLSMinVersion()).To(Equal(tls.VersionTLS12))
```

pkg/config/config_test.go
```diff
@@ -1944,4 +1944,93 @@ var _ = t.Describe("Config", func() {
...
+			Expect(sut.APIConfig.Validate(false)).To(Succeed())
+			// GetTLSCipherSuites returns nil for TLS 1.3
+			Expect(sut.APIConfig.GetTLSCipherSuites()).To(BeNil())
+		})
+
+		It("should validate config with TLS 1.3 even if cipher suites specified", func() {
+			sut.TLSMinVersion = "VersionTLS13"
+			sut.TLSCipherSuites = []string{"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"}
+			Expect(sut.APIConfig.Validate(false)).To(Succeed())
+		})
```
bitoku, line 2033: This test is a subset of the previous test. We don't need it.

pkg/config/config_test.go
bitoku:
```suggestion
			Expect(sut.APIConfig.GetTLSMinVersion()).To(Equal(tls.VersionTLS13))
```

pkg/config/config_test.go
bitoku:
```suggestion
			Expect(sut.APIConfig.GetTLSMinVersion()).To(Equal(tls.VersionTLS12))
```

pkg/config/config_test.go
```diff
@@ -1944,4 +1944,93 @@ var _ = t.Describe("Config", func() {
...
+			sut.TLSMinVersion = "VersionTLS10"
+			Expect(sut.APIConfig.Validate(false)).ToNot(Succeed())
+		})
+
+		It("should fail validation for deprecated TLS 1.1", func() {
+			sut.TLSMinVersion = "VersionTLS11"
+			Expect(sut.APIConfig.Validate(false)).ToNot(Succeed())
+		})
+
+		It("should return correct version from GetTLSMinVersion after Validate", func() {
```
bitoku: This can be combined in the second unit test.

test/tls_config.bats
```diff
@@ -0,0 +1,118 @@
...
+@test "tls_min_version: TLS 1.3 server should accept TLS 1.3 connections" {
+	generate_test_certs
+	PORT=$(free_port)
+
+	start_crio_with_tls "VersionTLS13" "$PORT"
+	run openssl s_client -connect "127.0.0.1:$PORT" -tls1_3
+	[[ "$output" == *"TLSv1.3"* ]]
+}
+
+@test "tls_min_version: TLS 1.3 server should reject TLS 1.2 connections" {
```
bitoku: This can be included in the previous TLS1.3 test.

test/tls_config.bats
```diff
@@ -0,0 +1,118 @@
...
+@test "tls_min_version: TLS 1.2 server should accept TLS 1.3 connections" {
+	generate_test_certs
+	PORT=$(free_port)
+
+	start_crio_with_tls "VersionTLS12" "$PORT"
+	run openssl s_client -connect "127.0.0.1:$PORT" -tls1_3
+	[[ "$output" == *"TLSv1.3"* ]]
+}
+
+@test "tls_min_version: TLS 1.2 server should accept TLS 1.2 connections" {
```
bitoku: This can be included in the previous TLS1.2 test.

docs/crio.8.md
```diff
@@ -421,9 +421,9 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--read-only**: Setup all unprivileged containers to run as read-only. Automatically mounts the containers' tmpfs on '/run', '/tmp' and '/var/tmp'.
 
-**--root, -r**="": The CRI-O root directory. (default: "/var/lib/containers/storage")
+**--root, -r**="": The CRI-O root directory. (default: "/home/asahay/.local/share/containers/storage")
 
-**--runroot**="": The CRI-O state directory. (default: "/run/containers/storage")
+**--runroot**="": The CRI-O state directory. (default: "/run/user/1002/containers")
```
haircommander: when generating completions, use sudo so the paths don't change (bad UX, i know :\)

haircommander: this change can just be unstaged

asahay19: yeah done. Thank you !

server/metrics/metrics.go
```diff
@@ -508,14 +514,17 @@ func (m *Metrics) startEndpoint(
...
 				log.Fatalf(ctx, "Creating key pair reloader: %v", err)
 			}
 
+			// disable "G402 (CWE-295): TLS MinVersion too low. (Confidence: HIGH, Severity: HIGH)"
+			// #nosec G402
 			srv.TLSConfig = &tls.Config{
 				GetConfigForClient: cc.GetConfigForClient,
-				MinVersion:         tls.VersionTLS12,
+				MinVersion:         m.apiConfig.GetTLSMinVersion(),
+				CipherSuites:       m.apiConfig.GetTLSCipherSuites(),
```
bitoku: It lgtm but please double check.

server/server.go
```diff
@@ -545,10 +545,12 @@ func New(
 			return nil, fmt.Errorf("load stream server x509 key pair: %w", err)
 		}
 
+		// #nosec G402 -- GetTLSMinVersion returns the validated TLS version. Any version older than TLS 1.2 will be rejected in config validation.
```
bitoku, line 548: For reviewers, please double check if it's ok.

Issue #9717, opened 2026-01-15
Concurrent `RemoveImage` calls show inconsistent ImageStatus results


saschagrunert:
## Description

The CRI-O implementation fails the Kubernetes CRI conformance test `should not fail on simultaneous RemoveImage calls [Conformance]` introduced in [kubernetes-sigs/cri-tools#1959](https://github.com/kubernetes-sigs/cri-tools/pull/1959).

The issue manifests as a race condition where `ImageStatus` can return image data even after `RemoveImage` has completed successfully, violating the CRI specification's idempotency requirement.

## How to Reproduce

**Important**: This test is not yet in the master branch. You need to build critest from the PR branch:

```bash
# Clone and build critest from the PR
git clone https://github.com/kubernetes-sigs/cri-tools.git
cd cri-tools
git fetch origin pull/1959/head:pr-1959
git checkout pr-1959
make
cd build/bin

# Run the test
./critest --runtime-endpoint=unix:///var/run/crio/crio.sock \
  --ginkgo.focus="should not fail on simultaneous RemoveImage calls"
```

The test will fail with errors indicating that `ImageStatus` found an image that was already removed.

## Expected Behavior

According to the [CRI API specification](https://github.com/kubernetes/cri-api/blob/c20fa40/pkg/apis/runtime/v1/api.proto#L156-L157):

> RemoveImage removes the image. This call is idempotent, and must not return an error if the image has already been removed.

Once `RemoveImage` returns successfully (or with success due to idempotency), a subsequent `ImageStatus` call should immediately return empty/nil, indicating the image does not exist.

## Actual Behavior

When multiple goroutines concurrently call:
1. `RemoveImage(imageID)` - completes successfully
2. `ImageStatus(imageID)` - immediately after RemoveImage completes

There is a timing window where `ImageStatus` still returns the image information instead of returning empty.

### Root Cause (Suspected)

The issue appears to be in one or more of these layers:

1. **containers/storage (c/storage)** - Timing issues between deletion and query operations at the storage layer
2. **go.podman.io/image/v5/storage (istorage)** - Caching or stale data in the image transport layer
3. **CRI-O's imageCache** - Internal caching that may serve stale data after deletion

### Attempted Workarounds

Several workarounds were tested:

1. **Synchronization** (`removeOperationsInProgress` map) - Prevents concurrent RemoveImage calls from racing with each other, but does NOT fix the `ImageStatus` consistency issue
[...]

PR #9726, opened 2026-01-26
Fix ImageStatus race after RemoveImage

saschagrunert:
#### What type of PR is this?

/kind bug

Fixes a race where ImageStatus can return image data even after RemoveImage completes successfully, violating the CRI specification.
    
The race occurs because ResolveReference can still find images after removal in read-only additional stores or when an image has multiple names and only one is removed.
    
Fix by tracking removed references: UntagImage/DeleteImage mark refs as removed, ImageStatus checks and returns not found, PullImage clears on successful pull, ListImages cleans up when refs reappear.

Fixes https://github.com/cri-o/cri-o/issues/9717
None

```release-note
Fixed a race condition where ImageStatus could return image data immediately after RemoveImage completed successfully, violating the CRI specification. This could occur when images existed in read-only additional stores or when removing one reference from an image with multiple names.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tracking of recently removed image references to prevent stale "not found" results after untag/delete or pull.
  * Cache refresh now re-recognizes re-added images and clears stale removal records.
  * Better handling of races between untag/remove, delete and pull to avoid false ImageStatus errors.

* **Tests**
  * Updated tests to account for additional storage reference handling and timing in image operations.

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -90,11 +90,25 @@ type imageLookupService struct {
 }
 
 type imageService struct {
-	lookup               *imageLookupService
-	store                storage.Store
-	storageTransport     StorageTransport
-	imageCache           imageCache
-	imageCacheLock       sync.Mutex
+	lookup           *imageLookupService
+	store            storage.Store
+	storageTransport StorageTransport
+	imageCache       imageCache
+	imageCacheLock   sync.Mutex
+	// removedReferences tracks recently removed image references.
+	// This is necessary because:
+	// 1. containers/storage supports read-only additional image stores that are never
+	//    updated when images are deleted from the read-write store
+	// 2. When an image has multiple names and one is removed, ResolveReference can
+	//    still find the image by digest even though that specific name was removed
+	// Without tracking, ImageStatus can find a deleted/untagged image even after
+	// RemoveImage completes successfully.
+	// Protected by imageCacheLock.
+	// Entries are cleaned up when ListImages discovers the reference exists again
+	// (e.g., after re-pulling). In long-running instances with many unique ephemeral
+	// images, this map may grow over time. Future enhancement: add TTL-based cleanup.
+	// See: https://github.com/cri-o/cri-o/issues/9717
+	removedReferences    map[string]struct{}
 	ctx                  context.Context
 	config               *config.Config
 	regexForPinnedImages []*regexp.Regexp
@@ -430,21 +444,21 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext) ([]Image
 	}
 
 	results := make([]ImageResult, 0, len(images))
-	newImageCache := make(imageCache, len(images))
 
+	newImageCache := make(imageCache, len(images))
 	for i := range images {
 		image := &images[i]
+
+		svc.imageCacheLock.Lock()
+		cacheItem, cached := svc.imageCache[image.ID]
+		svc.imageCacheLock.Unlock()
 
 		ref, err := istorage.Transport.NewStoreReference(svc.store, nil, image.ID)
 		if err != nil {
 			return nil, err
 		}
 
-		svc.imageCacheLock.Lock()
-		cacheItem, ok := svc.imageCache[image.ID]
-		svc.imageCacheLock.Unlock()
-
-		if !ok {
+		if !cached {
 			cacheItem, err = svc.buildImageCacheItem(systemContext, ref)
 			if err != nil {
 				if os.IsNotExist(err) && imageIsBeingPulled(image) { // skip reporting errors if the images haven't finished pulling
@@ -466,8 +480,38 @@ func (svc *imageService) ListImages(systemContext *types.SystemContext) ([]Image
 	}
 	// replace image cache with cache we just built
 	// this invalidates all stale entries in cache
+	// Build a map of all current image names for cleanup
+	// Must convert to transport format to match what's stored in removedReferences
+	existingReferences := make(map[string]struct{})
+
+	for i := range images {
+		for _, name := range images[i].Names {
+			// Convert name to the same format used in removedReferences (with transport prefix)
+			// Parse the name string into a reference.Named
+			namedRef, err := reference.ParseNamed(name)
+			if err != nil {
+				continue // Skip invalid references
+			}
+
+			ref, err := istorage.Transport.NewStoreReference(svc.store, namedRef, "")
+			if err != nil {
+				continue // Skip invalid references
+			}
+
+			refString := ref.StringWithinTransport()
+			existingReferences[refString] = struct{}{}
+		}
+	}
+
 	svc.imageCacheLock.Lock()
 	svc.imageCache = newImageCache
+	// Clean up removedReferences - remove entries for references that exist again
+	for ref := range svc.removedReferences {
+		if _, exists := existingReferences[ref]; exists {
+			delete(svc.removedReferences, ref)
+		}
+	}
+
 	svc.imageCacheLock.Unlock()
 
 	return results, nil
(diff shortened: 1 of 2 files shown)


bitoku:
> The race occurs because ResolveReference can still find images after removal in read-only additional stores or when an image has multiple names and only one is removed.

It sounds to me that we should fix in c/storage. Any reason it's difficult to fix?

saschagrunert:
> > The race occurs because ResolveReference can still find images after removal in read-only additional stores or when an image has multiple names and only one is removed.
> 
> It sounds to me that we should fix in c/storage. Any reason it's difficult to fix?

Yes I was approaching multiple fixes in c/storage or c/image but from my point of view it would require architectural changes to both to finally fix that up which possibly breaks podman down the road.

The actual bug is:

1. Image pulled: `Names = ["repo/image@sha256:abc", "repo/image:tag"]`
2. RemoveImage(@sha256:abc): Removes digest name, leaving `Names = ["repo/image:tag"]`
3. ImageStatus(@sha256:abc): Called immediately after removal
4. Bug: ResolveReference succeeds and returns the image

c/image finds the image through its digest fallback mechanism in [storage_reference.go:resolveImage()](https://github.com/cri-o/cri-o/blob/163f2fe6c1523370d65ebbc96358aaa0e26672c6/vendor/go.podman.io/image/v5/storage/storage_reference.go#L112-L189):

```go
// Line 116: Try exact name lookup
image := store.Image("repo/image@sha256:abc")  // FAILS (good)

// Line 122-151: Digest fallback for digested references
images := store.ImagesByDigest(sha256:abc)  // Finds the image
for _, img := range images {
    if imageMatchesRepo(img, "repo/image") {  // Checks if ANY name matches repo
        // img.Names = ["repo/image:tag"]
        // Repo matches! Returns image (bad)
    }
}
```

The [imageMatchesRepo()](https://github.com/cri-o/cri-o/blob/163f2fe6c1523370d65ebbc96358aaa0e26672c6/vendor/go.podman.io/image/v5/storage/storage_reference.go#L54-L62) function checks if any name shares the same repository. So `repo/image:tag` matches a lookup for `repo/image@sha256:abc`.

This digest fallback is intentional and correct for multi-platform images:

```
Pull nginx:latest
  Names: ["nginx:latest", "nginx@sha256:manifest", "nginx@sha256:amd64"]
Remove nginx:latest
  Names: ["nginx@sha256:manifest", "nginx@sha256:amd64"]
Lookup nginx@sha256:amd64
  Should succeed (digest refs must work)
```

And for short name resolution:

```
Pull nginx@sha256:abc
  Stores: "docker.io/library/nginx@sha256:abc"
Lookup nginx@sha256:abc
  Exact match fails, digest fallback needed
```

In c/image, removing digest fallback breaks multi-platform images and short names. Making `imageMatchesRepo` stricter can't distinguish short/long names. Both changes break Podman, Buildah, and Skopeo.

In c/storage, tracking "removed references" state raises questions: where to persist? what's the lifetime? This violates the CRUD abstraction, and c/storage doesn't understand "reference strings", only image IDs and names.

The fundamental issue is a semantic mismatch. The CRI specification requires reference-specific semantics: RemoveImage(ref) succeeds, then ImageStatus(ref) returns not found. But c/image implements content-addressed semantics: ResolveReference(digest) returns image if digest matches AND any name matches repo.

The digest fallback in c/image is correct and intentional for content-addressed storage. Changing it breaks legitimate workflows in other tools.

Only CRI-O knows which specific reference string was removed and that CRI requires reference-specific (not content-addressed) semantics. The `removedReferences` tracking implements CRI-specific semantics on top of storage primitives, which looks like the correct architectural layer. :eyes:

bitoku:
Thanks @saschagrunert for the clear explanation.
So it's not race, but rather a behavior by design.

Personally, it feels stranger that users can't specify an image by digest even though the image is present if `RemoveImage(digest)` is called, because the digest is something all images inherently have.

I'm inclined to change the critest.

saschagrunert:
> Thanks @saschagrunert for the clear explanation. So it's not race, but rather a behavior by design.
> 
> Personally, it feels stranger that users can't specify an image by digest even though the image is present if `RemoveImage(digest)` is called, because the digest is something all images inherently have.
> 
> I'm inclined to change the critest.

Tricky, we still have to consider:

- CRI specification compliance: The test validates what appears to be expected CRI behavior, so an image should not be accessible after removal
- It's counterintuitive that RemoveImage(digest) succeeds but the image is still accessible by that digest
- Other runtimes: If containerd and other CRI implementations pass this test, changing it would lower the bar

I'm still open to different approaches, though.

bitoku:
Or I'm wondering if we should remove (not untag) the image more aggressively. RemoveImage is called only when GC is required, isn't it?

saschagrunert:
> Or I'm wondering if we should remove (not untag) the image more aggressively. RemoveImage is called only when GC is required, isn't it?

Yes, RemoveImage is called by kubelet's image GC when disk usage exceeds thresholds. But we actually cannot remove images more aggressively because:

- We physically cannot delete images from read-only storage locations. They remain discoverable by ResolveReference after deletion from the read-write store.
- When removing one tag from an image with multiple names, we should only remove that specific tag, not the entire image.

bitoku:
I suggested change in https://github.com/kubernetes-sigs/cri-tools/pull/1959#discussion_r2732264125

> We physically cannot delete images from read-only storage locations. They remain discoverable by ResolveReference after deletion from the read-write store.

We can just ignore this scenario, I guess.

> When removing one tag from an image with multiple names, we should only remove that specific tag, not the entire image.

Yeah, that's true.

saschagrunert:
/hold
for now

SergeyKanzhelev:
> > When removing one tag from an image with multiple names, we should only remove that specific tag, not the entire image.
> 
> Yeah, that's true.

this is not an expectation as I understand and I am adding a test to lock it. See `removing image by one tag should remove all tags [Conformance]` in https://github.com/kubernetes-sigs/cri-tools/pull/1959/

It will be very confusing if runtimes will behave differently there

SergeyKanzhelev:
> Or I'm wondering if we should remove (not untag) the image more aggressively. RemoveImage is called only when GC is required, isn't it?

so if CRI-O will keep the knowledge that image was there for a while, while not having it as a "real" image, what will happen if a new Pod is scheduled that is asking for the same image? It was scheduled fore removal, while the reference will be given back to kubelet and kubelet will think it is still exists. This will potentially break the Pod

bitoku:
> > > When removing one tag from an image with multiple names, we should only remove that specific tag, not the entire image.
> > 
> > Yeah, that's true.

> this is not an expectation as I understand and I am adding a test to lock it. See `removing image by one tag should remove all tags [Conformance]` in https://github.com/kubernetes-sigs/cri-tools/pull/1959/

I assumed that CRI runtime behaves in the same way as docker or podman. It may cause confusion when they try to delete the image by `crictl`. I'm not opposing to the idea, but is that behavior something we already agreed on?

> so if CRI-O will keep the knowledge that image was there for a while, while not having it as a "real" image, what will happen if a new Pod is scheduled that is asking for the same image? It was scheduled fore removal, while the reference will be given back to kubelet and kubelet will think it is still exists. This will potentially break the Pod

I don't know what "not having real image" means, but iiuc kubelet only removes images by imageID (not `name@digest` or `name:tag` in CRI-O) returned by the CRI runtime, so as long as it's deleted by kubelet without any errors, the image is really deleted. The image won't stay there as we see when deleting by `name@digest` or `name:tag`.

SergeyKanzhelev:
> > this is not an expectation as I understand and I am adding a test to lock it. See `removing image by one tag should remove all tags [Conformance]` in [kubernetes-sigs/cri-tools#1959](https://github.com/kubernetes-sigs/cri-tools/pull/1959)
> 
> I assumed that CRI runtime behaves in the same way as docker or podman. It may cause confusion when they try to delete the image by `crictl`. I'm not opposing to the idea, but is that behavior something we already agreed on?

This comes from the discussion: https://github.com/kubernetes-sigs/cri-tools/issues/1911 I do not think this behavior makes difference for kubelet. But if you feel strongly about it, feel free to offer the clean up in docs and tests. I am open for discussion

PR #9728, opened 2026-01-27
Return image ID from PullImage instead of repo digest

saschagrunert:
#### What type of PR is this?

/kind bug

This PR changes `PullImage` to return the storage image ID instead of the repo digest, fixing a compatibility issue with Kubernetes credential verification which expects `PullImage` and `GetImageRef` to return compatible values for pull record lookups.

The ref-to-ID resolution happens in the server layer (`server/image_pull.go`) via a new `resolveImageRefToID` method, keeping the storage layer's `PullImage` contract unchanged. For OCI artifacts (not stored in container storage), it falls back to the artifact store.

None

None

```release-note
PullImage now returns the image ID directly, ensuring compatibility with Kubernetes credential verification for image pulls.
```

### DIFF
--- a/server/image_pull.go
+++ b/server/image_pull.go
@@ -123,17 +123,18 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
 		return nil, storage.WrapSignatureCRIErrorIfNeeded(pullOp.err)
 	}
 
-	log.Infof(ctx, "Pulled image: %v", pullOp.imageRef)
+	log.Infof(ctx, "Pulled image: %s", pullOp.imageRef)
 
 	return &types.PullImageResponse{
-		ImageRef: pullOp.imageRef.StringForOutOfProcessConsumptionOnly(),
+		ImageRef: pullOp.imageRef,
 	}, nil
 }
 
 // pullImage performs the actual pull operation of PullImage. Used to separate
 // the pull implementation from the pullCache logic in PullImage and improve
 // readability and maintainability.
-func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storage.RegistryImageReference, error) {
+// It returns the image ID string suitable for PullImageResponse.ImageRef.
+func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (string, error) {
 	var err error
 
 	ctx, span := log.StartSpan(ctx)
@@ -141,7 +142,7 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
 
 	sourceCtx, err := s.contextForNamespace(pullArgs.namespace)
 	if err != nil {
-		return storage.RegistryImageReference{}, fmt.Errorf("get context for namespace: %w", err)
+		return "", fmt.Errorf("get context for namespace: %w", err)
 	}
 
 	log.Debugf(ctx, "Using pull policy path for image %s: %q", pullArgs.image, sourceCtx.SignaturePolicyPath)
@@ -149,7 +150,7 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
 	if pullArgs.namespace != "" {
 		authCleanup, err := s.prepareTempAuthFile(ctx, &sourceCtx, pullArgs.image, pullArgs.namespace)
 		if err != nil {
-			return storage.RegistryImageReference{}, fmt.Errorf("prepare temp auth file: %w", err)
+			return "", fmt.Errorf("prepare temp auth file: %w", err)
 		}
 		defer authCleanup()
 	}
@@ -161,14 +162,14 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
 
 	decryptConfig, err := getDecryptionKeys(s.config.DecryptionKeysPath)
 	if err != nil {
-		return storage.RegistryImageReference{}, err
+		return "", err
 	}
 
 	cgroup := ""
 
 	if s.config.SeparatePullCgroup != "" {
 		if !s.config.CgroupManager().IsSystemd() {
-			return storage.RegistryImageReference{}, errors.New("--separate-pull-cgroup is supported only with systemd")
+			return "", errors.New("--separate-pull-cgroup is supported only with systemd")
 		}
 
 		if s.config.SeparatePullCgroup == utils.PodCgroupName {
@@ -176,32 +177,32 @@ func (s *Server) pullImage(ctx context.Context, pullArgs *pullArguments) (storag
 		} else {
 			cgroup = s.config.SeparatePullCgroup
 			if !strings.Contains(cgroup, ".slice") {
-				return storage.RegistryImageReference{}, fmt.Errorf("invalid systemd cgroup %q", cgroup)
+				return "", fmt.Errorf("invalid systemd cgroup %q", cgroup)
 			}
 		}
 	}
 
 	remoteCandidates, err := s.ContainerServer.StorageImageServer().CandidatesForPotentiallyShortImageName(s.config.SystemContext, pullArgs.image)
 	if err != nil {
-		return storage.RegistryImageReference{}, err
+		return "", err
 	}
 	// CandidatesForPotentiallyShortImageName is defined never to return an empty slice on success, so if the loop considers all candidates
 	// and they all fail, this error value should be overwritten by a real failure.
 	lastErr := errors.New("internal error: pullImage failed but reported no error reason")
 
 	for _, remoteCandidateName := range remoteCandidates {
-		repoDigest, err := s.pullImageCandidate(ctx, &sourceCtx, remoteCandidateName, decryptConfig, cgroup)
+		imageRef, err := s.pullImageCandidate(ctx, &sourceCtx, remoteCandidateName, decryptConfig, cgroup)
 		if err == nil {
 			// Update metric for successful image pulls
 			metrics.Instance().MetricImagePullsSuccessesInc(remoteCandidateName)
 
-			return repoDigest, nil
+			return s.resolveImageRefToID(ctx, imageRef)
 		}
 
 		lastErr = err
 	}
 
-	return storage.RegistryImageReference{}, lastErr
+	return "", lastErr
 }
 
 // contextForNamespace takes the provided namespace and returns a modifiable
@@ -328,6 +329,27 @@ func (s *Server) pullImageCandidate(ctx context.Context, sourceCtx *imageTypes.S
 	}
 
 	return repoDigest, nil
+}
+
+// resolveImageRefToID converts a pulled image reference (repo@digest) to a
+// storage image ID suitable for PullImageResponse.ImageRef. For regular
+// container images the ID is looked up via ImageStatusByName. For OCI artifacts
+// (which are not stored in container storage) it falls back to the artifact
+// store.
+func (s *Server) resolveImageRefToID(ctx context.Context, imageRef storage.RegistryImageReference) (string, error) {
+	// Try resolving as a regular container image first.
+	imageResult, err := s.ContainerServer.StorageImageServer().ImageStatusByName(s.config.SystemContext, imageRef)
+	if err == nil {
+		return imageResult.ID.IDStringForOutOfProcessConsumptionOnly(), nil
+	}
+
+	// Fall back to the artifact store for OCI artifacts.
+	artifact, artifactErr := s.ArtifactStore().Status(ctx, imageRef.StringForOutOfProcessConsumptionOnly())
+	if artifactErr != nil {
+		return "", fmt.Errorf("resolve pulled image %s: not found in container storage (%w) or artifact store (%w)", imageRef, err, artifactErr)
+	}
+
+	return artifact.CRIImage().GetId(), nil
 }
 
 // consumeImagePullProgress consumes progress and turns it into metrics updates.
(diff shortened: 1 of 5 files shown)


test/image.bats
```diff
@@ -469,3 +469,33 @@ EOF
...
+
+	# Get the image status for the same image
+	imageid=$(crictl images --quiet "$IMAGE")
+	[ "$imageid" != "" ]
+
+	# The pulled reference should match the image ID from ImageStatus
+	# Both PullImage and GetImageRef (via ImageStatus) should return the same value
+	# to ensure Kubernetes credential verification works correctly
+	[ "$pulled_id" = "$imageid" ]
+
```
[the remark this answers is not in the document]
saschagrunert: This should be fixed now.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

saschagrunert:
/hold 

Integration tests are not happy.

bitoku:
iiuc ID field is not well-defined. I'd want to define the format first though I'm not opposed to the change.

saschagrunert:
> iiuc ID field is not well-defined. I'd want to define the format first though I'm not opposed to the change.

We can also follow-up on CRI docs for that, yep.

server/image_pull_test.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/opencontainers/go-digest"
```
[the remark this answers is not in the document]
saschagrunert: Should be fixed now.

saschagrunert:
@cri-o/cri-o-maintainers PTAL

server/image_pull.go
```diff
@@ -125,9 +125,27 @@ func (s *Server) PullImage(ctx context.Context, req *types.PullImageRequest) (*t
...
+		}, nil
+	}
+
+	// If not found in regular image storage, check if it's an OCI artifact
+	artifact, artifactErr := s.ArtifactStore().Status(ctx, pullOp.imageRef.StringForOutOfProcessConsumptionOnly())
+	if artifactErr == nil {
+		return &types.PullImageResponse{
+			ImageRef: artifact.CRIImage().GetId(),
+		}, nil
+	}
```
bitoku: I wonder if we can change the signature of PullImage to return ID. It may require more changes though...

mtrmac:
To add a bit more nuance: The very strict NAK is for promising that image IDs == config digests. That **will** break or at the very least incompatibly change syntax: https://github.com/containers/podman/issues/27923 .

As for `PullImage` returning a storage-provided ID, as this PR does, without assuming any specific format, that should work. I agree that modifying `internal/storage` to return that value would be rather easier.

The consequence of that is that this will make it _harder_ to deal with using the correct set of signatures for deduplicated images (compare the comment in  `imageService.checkSignature`), but that already happens in the “image already exists locally, no pull issued” code path.

saschagrunert:
> To add a bit more nuance: The very strict NAK is for promising that image IDs == config digests. That **will** break or at the very least incompatibly change syntax: [containers/podman#27923](https://github.com/containers/podman/issues/27923) .
> 
> As for `PullImage` returning a storage-provided ID, as this PR does, without assuming any specific format, that should work. I agree that modifying `internal/storage` to return that value would be rather easier.
> 
> The consequence of that is that this will make it _harder_ to deal with using the correct set of signatures for deduplicated images (compare the comment in `imageService.checkSignature`), but that already happens in the “image already exists locally, no pull issued” code path.

Right now it's broken in a way that we have different values returned by containerd and CRI-O, which is something we have to fix because an upstream feature already depends on it (https://github.com/kubernetes/enhancements/issues/2535) and others may come.

If it's the config hash or not does not matter, it only matters that it's a unique identifier of an image.

mtrmac:
Either way please deal with https://github.com/cri-o/cri-o/pull/9615#pullrequestreview-3723026527 first.

server/image_pull.go
saschagrunert: By modifying/extending the CRI?

saschagrunert:
@cri-o/cri-o-maintainers @mtrmac PTAL again, we reverted the previous change of https://github.com/cri-o/cri-o/pull/9615

test/policy.bats
```diff
@@ -138,13 +138,14 @@ SANDBOX_CONFIG="$TESTDATA/sandbox_config.json"
 
 @test "allow signed image with restrictive policy on container creation1 (fresh pull)" {
 	start_crio
-	IMAGE_DIGEST=$(crictl pull "$SIGNED_IMAGE" | cut -d' ' -f7)
+	crictl pull "$SIGNED_IMAGE"
+	IMAGE_ID=$(crictl images -q "$SIGNED_IMAGE")
```
mtrmac: Doesn’t `pull` now return the image ID? If so, why the extra `images -q`?

This is not just aesthetics, I think testing the _exact_ data flow for `PullImage` + `CreateContainer` is valuable.

(Two times. I didn’t look into what `crictl pull` outputs _at all_.)

server/image_pull.go
bitoku: I mean this https://github.com/cri-o/cri-o/blob/8e70e64ef474b0cbd430e1432af6e19fb8fd58c5/internal/storage/image.go#L134-L146

test/image.bats
```diff
@@ -413,3 +413,33 @@ EOF
...
+	# Pull an image and capture the returned image reference
+	pulled_ref=$(crictl pull "$IMAGE")
+
+	# Extract the image ID from crictl output
+	# crictl may output "Image is up to date for <id>" or just "<id>"
+	# We want just the ID part (the last word)
+	pulled_id=$(echo "$pulled_ref" | awk '{print $NF}')
+
+	# The pulled ID should be an image ID (64 hex characters)
+	[[ "$pulled_id" =~ ^[a-f0-9]{64}$ ]]
```
mtrmac: It‘s unclear that this will hold long-term (see earlier NAKs) but it’s fine for now.

server/image_pull.go
mtrmac: I think this was talking about `s.ContainerServer.StorageImageServer().PullImage`, and I agree that moving the lookup there would be better. No need to have the repo@digest value in the `StorageImageServer` API, and just the lookup of a repo@digest to ID (which still needs to happen) can be done cheaper than the whole `ImageStatusByName`.

test/policy.bats
saschagrunert: I've updated the tests to use the image ID from pull output directly :+1:

server/image_pull.go
saschagrunert: The PullImage interface now returns both the image ID and the canonical reference.

internal/storage/image.go
```diff
@@ -680,13 +682,21 @@ func pullImageChild() {
...
 		fmt.Fprintf(os.Stderr, "%v", err)
 		os.Exit(1)
 	}
 
-	output <- pullImageOutputItem{Result: canonicalRef.StringForOutOfProcessConsumptionOnly()}
+	outputItem := pullImageOutputItem{
+		Result: canonicalRef.StringForOutOfProcessConsumptionOnly(),
+	}
+	// Only serialize ImageID if non-empty (OCI artifacts return empty StorageImageID)
+	if imageID != (StorageImageID{}) {
```
mtrmac: From the documentation of `StorageImageID`:
> // This is intended to be a value type; if a value exists, it is a correctly-formatted ID.

This is violating that principle. [It’s up to the CRI-O maintainers whether they want to change the design, but I think  the documentation and implementation must match the use.]

Use `*StorageImageID` if you want to express optionality, compare `HeuristicallyTryResolvingStringAsIDPrefix`.

internal/storage/image.go
```diff
@@ -753,26 +763,31 @@ func (svc *imageService) pullImageParent(ctx context.Context, imageName Registry
...
-		return RegistryImageReference{}, fmt.Errorf("json encode to pipe failed: %w", err)
+		return StorageImageID{}, RegistryImageReference{}, fmt.Errorf("json encode to pipe failed: %w", err)
 	}
 
 	stdin.Close()
 
-	resultChan := make(chan string)
+	type pullResult struct {
+		imageID      string
+		canonicalRef string
```
mtrmac: Do we even need to return the `canonicalRef`? AFAICS the other consumer also ultimately wants an image ID.

(It is, admittedly, potentially _very_ valuable for logging and debugging. So maybe we do.)

internal/storage/image.go
```diff
@@ -877,28 +904,41 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+		return StorageImageID{}, RegistryImageReference{}, fmt.Errorf("creating storage reference: %w", err)
+	}
+
+	_, image, err := storageTransport.ResolveReference(unstableRef)
+	if err != nil {
+		return StorageImageID{}, RegistryImageReference{}, fmt.Errorf("resolving image reference: %w", err)
 	}
 
-	return references.RegistryImageReferenceFromRaw(canonicalRef), nil
+	return storageImageIDFromImage(image), references.RegistryImageReferenceFromRaw(canonicalRef), nil
```
mtrmac: ACK

internal/storage/image_test.go
```diff
@@ -580,7 +580,7 @@ var _ = t.Describe("Image", func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			// When
-			res, err := sut.PullImage(context.Background(), imageRef, &storage.ImageCopyOptions{
+			_, res, err := sut.PullImage(context.Background(), imageRef, &storage.ImageCopyOptions{
```
mtrmac: Consider testing more, throughout.

internal/storage/image.go
saschagrunert: I removed the `canonicalRef` in favor of the `imageID`.

internal/storage/image.go
saschagrunert: Switched over to use `*StorageImageID`. :+1:

internal/storage/image_test.go
saschagrunert: Yeah I added some value assertions at least.

test/image.bats
saschagrunert: Addressed this by checking for the ID not being empty for now.

internal/storage/image.go
```diff
@@ -637,7 +636,7 @@ type pullImageArgs struct {
 
 type pullImageOutputItem struct {
 	Progress *types.ProgressProperties `json:",omitempty"`
-	Result   string                    `json:",omitempty"` // If not "", in the format of RegistryImageReference.StringForOutOfProcessConsumptionOnly(), and always contains a digest.
+	ImageID  string                    `json:",omitempty"` // If not "", the storage image ID in the format of StorageImageID.IDStringForOutOfProcessConsumptionOnly()
```
bitoku: nit: 

```suggestion
	// ImageID is in the format of StorageImageID.IDStringForOutOfProcessConsumptionOnly(), or "" if it's OCI artifacts
	ImageID  string                    `json:",omitempty"`
```

internal/storage/image.go
```diff
@@ -868,37 +876,46 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
-
-		canonicalRef, err := reference.WithDigest(reference.TrimNamed(imageName.Raw()), *artifactManifestDigest)
-		if err != nil {
-			return RegistryImageReference{}, fmt.Errorf("create canonical reference: %w", err)
+			return nil, fmt.Errorf("unable to pull image or OCI artifact: pull image err: %w; artifact err: %w", err, artifactErr)
 		}
 
-		return references.RegistryImageReferenceFromRaw(canonicalRef), nil
+		// For OCI artifacts, we don't have a traditional image ID, so we return nil
+		return nil, nil
```
bitoku: I guess, we should return something else? I want more clarification why it's acceptable.

internal/storage/image.go
saschagrunert: Fixed

internal/storage/image.go
saschagrunert: Added a comment to clarify it.

internal/storage/image.go
```diff
@@ -868,37 +877,48 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
-		if err != nil {
-			return RegistryImageReference{}, fmt.Errorf("create canonical reference: %w", err)
+			return nil, fmt.Errorf("unable to pull image or OCI artifact: pull image err: %w; artifact err: %w", err, artifactErr)
 		}
 
-		return references.RegistryImageReferenceFromRaw(canonicalRef), nil
+		// OCI artifacts are not stored in container storage and therefore have
+		// no storage image ID. Returning nil signals callers to fall back to
+		// the artifact store for status lookups (see PullImage in server/image_pull.go).
+		return nil, nil
```
bitoku: Probably we shouldn't have implemented artifact pull in this storage, but returning nil here doesn't sound right to me.
If we think the OCI artifacts are managed by this package, artifacts should have storageID, and otherwise we should move pull implementation to elsewhere.

For now, I think keeping it return reference, and return ID in image_pull is better. Sorry for backing down my initial comment.

saschagrunert: I've updated the PR to keep `pullImageImplementation` returning `RegistryImageReference` (unchanged storage layer) and moved the ref to ID resolution to `server/image_pull.go` via a new `resolveImageRefToID` method. It tries `ImageStatusByName` first, then falls back to the artifact store for OCI artifacts. No more `nil` return path.

bitoku:
lgtm, though I want to make sure there's enough time for others to review it.

internal/storage/image.go
mtrmac: I really have no idea why doing that outside of `pullImageImplementation` should be easier (what if an object with the same name@digest exists in both??) but it’s not my codebase to maintain.

PR #9778, opened 2026-02-24
Skip OCI artifact fallback on retryable errors

bitoku:
#### What type of PR is this?

/kind cleanup

Avoid falling back to an OCI artifact pull when the initial image pull fails due to a network error, since the artifact pull would fail the same way. This prevents blocking pod operations for an extra timeout cycle on unreachable registries. Also disable retries on the artifact pull path to further reduce latency on failures.

```release-note
Skip the OCI artifact pull fallback when the initial image pull fails due to a retryable error
```

## Summary by CodeRabbit

* **New Features**
  * Added an OCI-artifact fallback for image pulls when primary pulls fail, with debug logging and final image reference construction from artifact digest; artifact pulls use zero retries to avoid blocking.

* **Bug Fixes**
  * Returns clearer combined errors when primary and artifact pull attempts fail.
  * Detects network-related failures and avoids artifact fallback to prevent delays.

* **Tests**
  * Added a test ensuring network errors do not trigger artifact fallback and cleaned up related test scaffolding.

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"regexp"
 	"sort"
@@ -34,6 +35,7 @@ import (
 	"go.podman.io/storage"
 	"go.podman.io/storage/pkg/reexec"
 	crierrors "k8s.io/cri-api/pkg/errors"
+	"k8s.io/utils/ptr"
 
 	"github.com/cri-o/cri-o/internal/log"
 	"github.com/cri-o/cri-o/internal/ociartifact"
@@ -865,7 +867,9 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 		ProgressInterval: options.ProgressInterval,
 		Progress:         options.Progress,
 	})
-	if err != nil {
+	if shouldTryArtifact(err) {
+		log.Infof(ctx, "Falling back to pull %s as an OCI artifact: %v", imageName, err)
+
 		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), &srcSystemContext)
 		if artifactErr != nil {
 			return RegistryImageReference{}, fmt.Errorf("unable to pull image or OCI artifact: create store err: %w", artifactErr)
@@ -874,6 +878,10 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 		artifactManifestDigest, artifactErr := artifactStore.Pull(ctx, srcRef, &libimage.CopyOptions{
 			OciDecryptConfig: options.OciDecryptConfig,
 			Progress:         options.Progress,
+			// Disable retries to avoid blocking pod operations for
+			// (timeout * MaxRetries) on network failures. Rely on
+			// Kubelet retries instead.
+			MaxRetries:       ptr.To(uint(0)),
 			RemoveSignatures: true, // signature is not supported for OCI layout dest
 		})
 		if artifactErr != nil {
@@ -888,6 +896,10 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 		return references.RegistryImageReferenceFromRaw(canonicalRef), nil
 	}
 
+	if err != nil {
+		return RegistryImageReference{}, fmt.Errorf("unable to pull image: %w", err)
+	}
+
 	manifestDigest, err := manifest.Digest(manifestBytes)
 	if err != nil {
 		return RegistryImageReference{}, fmt.Errorf("digesting image: %w", err)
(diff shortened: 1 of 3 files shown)


internal/storage/image.go
```diff
@@ -828,6 +830,22 @@ func (svc *imageService) PullImage(ctx context.Context, imageName RegistryImageR
 	}
 }
 
+// shouldTryArtifact determines whether a failed image pull should fall back to
+// an OCI artifact pull.
+func shouldTryArtifact(err error) bool {
```
haircommander: nit: can this be defined below where it's called?

bitoku: yup 👍

internal/storage/image.go
```diff
@@ -901,6 +913,22 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+	switch {
+	case err == nil:
+		return false
+	case errors.As(err, &netErr):
+		// Network errors indicate the registry is unreachable, so an artifact
+		// pull would fail the same way. Rely on Kubelet retries instead.
+		return false
+	default:
+		return true
+	}
```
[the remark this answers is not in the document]
bitoku: https://cs.opensource.google/go/go/+/refs/tags/go1.26.0:src/context/context.go;l=165-167
Does context.Canceled implements Timeout()?

internal/storage/image.go
```diff
@@ -901,6 +913,24 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 	return references.RegistryImageReferenceFromRaw(canonicalRef), nil
 }
 
+// shouldTryArtifact determines whether a failed image pull should fall back to
+// an OCI artifact pull.
+func shouldTryArtifact(err error) bool {
+	switch {
+	case err == nil:
+		return false
+	case retry.IsErrorRetryable(err):
```
bitoku: https://github.com/containers/common/blob/v0.64.2/pkg/retry/retry.go#L68

haircommander:
/retest
/approve

LGTM

bitoku:
@saschagrunert Can you PTAL?

internal/storage/image.go
```diff
@@ -901,6 +913,24 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+		// Retryable errors should be retried as an image pull rather than
+		// falling back to an artifact pull.
+		return false
+	case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
+		// If a caller explicitly cancels the context or times out, attempting an artifact pull is wasted work
+		return false
```
saschagrunert, line 926: This is redundant due to https://github.com/containers/common/blob/82146f3af4750038f936ec9cd804d12a837e5cde/pkg/retry/retry.go#L72-L73

internal/storage/image.go
saschagrunert: This function is documented as unstable. The function explicitly says "That heuristic is NOT STABLE and it CAN CHANGE AT ANY TIME." Using it as a gate for the artifact fallback means the artifact fallback behavior is coupled to an unstable heuristic.

bitoku: right, I fixed it only skip fall back when there's network timeout, and other context errors.

internal/storage/image.go
```diff
@@ -901,6 +913,27 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+	case err == nil:
+		return false
+	case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
+		// Caller-initiated cancellation or timeout; no point retrying as artifact.
+		return false
+	case errors.As(err, &netError):
+		if netError.Timeout() {
+			// Network errors are transient; retry as image pull instead of falling back.
+			return false
+		}
```
saschagrunert: Non-timeout `net.Error` (e.g. `ECONNREFUSED`, `ENETUNREACH`, `EHOSTUNREACH`) falls through to `return true`, so the artifact fallback still runs against the same unreachable registry. Consider returning `false` for all `net.Error` matches, they are equally transient and the artifact pull would fail the same way.

test/image.bats
```diff
@@ -413,3 +413,13 @@ EOF
 	# There should be many nginx images
 	crictl pull nginx
 }
+
+@test "image pull should not fall back to OCI artifact on network error" {
+	start_crio
+
+	# 192.0.2.1 is TEST-NET-1 (RFC 5737), guaranteed unreachable
```
saschagrunert, line 450: This assumes `192.0.2.1` produces a `net.Error` with `Timeout() == true` (silently dropped packets). Depending on firewall/routing rules, it may return `EHOSTUNREACH` or `ENETUNREACH` instead (`Timeout() == false`), causing `shouldTryArtifact` to return `true` and the fallback to run.

internal/storage/image.go
```diff
@@ -888,6 +896,10 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 		return references.RegistryImageReferenceFromRaw(canonicalRef), nil
 	}
 
+	if err != nil {
```
saschagrunert, line 899: Nit: other error paths in this function use `"unable to pull image..."` this one says `"unable to copy image"`.

internal/storage/image.go
bitoku: True. I fixed it.

test/image.bats
bitoku: yeah my initial thought was that it can be imperfect because it just ends up with another pull.
And with https://github.com/cri-o/cri-o/pull/9782 CRI-O won't pull images as artifacts accidentally.
But I added the check, that if the error is net.Error, fail the whole pull.

internal/storage/image.go
```diff
@@ -901,6 +913,25 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
...
+// an OCI artifact pull. It optimistically falls back for non-transient errors,
+// since the image reference may actually be an artifact.
+func shouldTryArtifact(err error) bool {
+	var netError net.Error
+	switch {
+	case err == nil:
+		return false
+	case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
+		// Caller-initiated cancellation or timeout; no point retrying as artifact.
+		return false
```
saschagrunert, line 926: `errors.Is(err, context.DeadlineExceeded)` is technically redundant with `errors.As(err, &netError)` since Go's `deadlineExceededError` satisfies `net.Error`. But keeping it explicit alongside `context.Canceled` is fine for readability and intent.

bitoku:
@saschagrunert
Rebase removed lgtm, Can you tag it again?

PR #9780, opened 2026-02-25
Revert artifact dependency change.

bitoku:
#### What type of PR is this?

/kind bug

Reverting https://github.com/cri-o/cri-o/pull/9770

```release-note
Fixed the issue where the image pull failed by some temporary issue, pulled as an artifact instead.
```


bitoku:
/hold
maybe we don't have to revert. we can cherry-pick the fix commit instead.

PR #9784, opened 2026-02-26
Fix metric label cardinality mismatches in CRI stats

dgrisonnet:
#### What type of PR is this?

/kind bug

When `PodAndContainerStatsFromCRI` is enabled, the kubelet rejects all CRI metrics from CRI-O with "inconsistent label cardinality" errors because the number of label values doesn't match the number of label keys declared in the metric descriptors.

While testing I ran into two separate label inconsistency issues:

1. The container_ulimits_soft and container_network_* metrics were missing one label when they were generated
```
E0226 13:44:29.365916 cri_metrics.go:118] "Error getting CRI prometheus metric" err="inconsistent label cardinality: expected 4 label values but got 3 in []string{"06b2e3a8...", "k8s_coredns_coredns-...", "registry.k8s.io/coredns/coredns:v1.14.1"}" descriptor="Desc{fqName: "container_ulimits_soft", variableLabels: {id,name,image,ulimit}}"
```

2. For all metrics, cri-o was generating a metric name (cpu, memory, process, ...) label which wasn't defined in the metric descriptor
```
got 4 in []string{"da989c2aa...", "k8s_coredns...", "registry.k8s.io/coredns/coredns:v1.14.1", "process"}
```

In this PR, I added the missing labels and removed the metric name as a label as I don't think adding it is intended.

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Network metrics now include per-network name labels for clearer identification.
  * Process metrics include a "max_open_files" label for ulimits.
  * Metric label handling simplified and made consistent across CPU, disk, memory, hugetlb, spec, pressure, and others to prevent inconsistent label outputs.

* **Tests**
  * Added unit tests validating metric label cardinality to ensure reliable Prometheus metrics.

### DIFF
--- a/internal/lib/statsserver/cpu_metrics_linux.go
+++ b/internal/lib/statsserver/cpu_metrics_linux.go
@@ -83,5 +83,5 @@ func generateContainerCPUMetrics(ctr *oci.Container, cpu *cgroups.CpuStats) []*t
 		},
 	}
 
-	return computeContainerMetrics(ctr, cpuMetrics, "cpu")
+	return computeContainerMetrics(ctr, cpuMetrics)
 }
--- a/internal/lib/statsserver/disk_metrics_linux.go
+++ b/internal/lib/statsserver/disk_metrics_linux.go
@@ -47,7 +47,7 @@ func generateContainerDiskMetrics(ctr *oci.Container, diskStats *stats.Filesyste
 		},
 	}
 
-	return computeContainerMetrics(ctr, diskMetrics, "disk")
+	return computeContainerMetrics(ctr, diskMetrics)
 }
 
 // generateContainerDiskIOMetrics computes filesystem disk metrics from DiskStats for a container sandbox.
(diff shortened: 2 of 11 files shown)


bitoku:
Is that discrepancy caused by CRI-O's metric descriptors and CRI-O's actual metrics?

I feel we want integration tests to verify that in test/.

haircommander:
> Is that discrepancy caused by CRI-O's metric descriptors and CRI-O's actual metrics?

yeah I think this may actually be a bug in what we return for metrics descriptors, not metrics

dgrisonnet:
The descriptors are good, the problem is that CRI-O generates metrics that have additional or missing labels depending on the case

dgrisonnet:
The test failure is caused by https://github.com/cri-o/cri-o/issues/9717.

Re-running the workflow.

haircommander:
for some reason I was convinced there was the metric family embedded in a label in the metric, but I can't find it! 

/approve

bitoku:
We want integration tests to prevent this happens again, or we can implement it in critest?

haircommander:
either, though I bet containerd could benefit and critest would be good for that

dgrisonnet:
I'll pick it manually

PR #9793, opened 2026-03-01
checkpoint: allow callers to specify work and image paths

adrianreber:
#### What type of PR is this?

/kind cleanup

Add workPath and imagePath parameters to the CheckpointContainer interface and all implementations (runtimeOCI, runtimePod, runtimeVM).

Previously these paths were derived internally from the container object.  Making them explicit enables callers to direct CRIU output to a location of their choice.

The main motivation for this change is pod-level checkpointing. When creating a pod checkpoint every container checkpoint needs to be included in a single checkpoint archive (or image).  If the checkpoint data is always written to the container's default directory first, it has to be copied into the archive afterwards. For large checkpoints this copying step can be very expensive. By letting the caller specify the target paths directly, the checkpoint data can be written straight into the archive's directory structure, avoiding the extra copy entirely.

The existing single-container checkpoint path continues to pass `ctr.Dir()` and `ctr.CheckpointPath()`, so its behaviour is unchanged.

Generated with Claude Code (https://claude.ai/code)

None

No functional changes. Just additional parameters. Trying to add as a simple change early to reduce review size

```release-note
None
```

## Summary by CodeRabbit

* **Refactor**
  * Improved container checkpoint handling to support explicit working and checkpoint path usage for more reliable checkpoint operations.

* **Bug Fixes**
  * Better error detection and cleanup when checkpoint paths are invalid or inaccessible, reducing leftover artifacts on failure.

* **Tests**
  * Expanded tests to cover new path-related failure modes and ensure robust checkpoint behavior.

### DIFF
--- a/internal/lib/checkpoint.go
+++ b/internal/lib/checkpoint.go
@@ -92,7 +92,7 @@ func (c *ContainerServer) ContainerCheckpoint(
 		}
 	}
 
-	if err := c.runtime.CheckpointContainer(ctx, ctr, specgen.Config, opts.KeepRunning); err != nil {
+	if err := c.runtime.CheckpointContainer(ctx, ctr, specgen.Config, opts.KeepRunning, ctr.Dir(), ctr.CheckpointPath()); err != nil {
 		// in the case of an error, clean up any leftover CRIU images
 		if err := os.RemoveAll(ctr.CheckpointPath()); err != nil {
 			log.Warnf(ctx, "Unable to remove checkpoint directory %s: %v", ctr.CheckpointPath(), err)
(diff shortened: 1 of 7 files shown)


adrianreber:
@haircommander This is related to the discussion in the working group about pod level checkpointing. This is just some additional parameters with no functional changes so far. I thought this could be merged earlier as it doesn't break anything and would reduce the size of the changes later. It will be important once pod level checkpointing is introduced to not require copying checkpoint data around. Let me know if you prefer to take all related changes rather in one PR and then I can close this again.

rst0git:
> This is related to the discussion in the working group about pod level checkpointing. This is just some additional parameters with no functional changes so far. I thought this could be merged earlier as it doesn't break anything and would reduce the size of the changes later.

@adrianreber I agree that the copying step can be expensive, and we want to avoid it, but it is not very clear how we are going to use this later. It might be better to introduce it when this is needed for functional changes.

PR #9799, opened 2026-03-05
OCPBUGS-76451: fix: prevent panic on closed stopTimeoutChan in StopContainer

sabujmaity:
/kind bug

There is a race condition in the container stop path. When a second `StopContainer` call
comes in after the first one has already finished (i.e. `SetAsDoneStopping` has
run), `WaitOnStopTimeout` still gets called. At that point `stopTimeoutChan`
is already closed, so we panic.

The sequence is:

1. First `StopContainer` -> starts the stop loop, waits via `WaitOnStopTimeout`
2. Stop loop finishes -> `SetAsDoneStopping` closes `stopTimeoutChan` and watchers
3. Second `StopContainer` -> `SetAsStopping` returns false (already stopping),
   falls through to `WaitOnStopTimeout` which tries to use the closed channel

This adds a `stopDone` bool (guarded by the existing `stopLock`) that gets set
in `SetAsDoneStopping`. `WaitOnStopTimeout` checks it and returns early if
the stop lifecycle is already done.

Addresses: [OCPBUGS-76451](https://issues.redhat.com/browse/OCPBUGS-76451)

The change is small : one new bool field, one assignment, one extra condition
check. All under the existing `stopLock` so no new synchronization concerns.

Added a unit test that exercises the exact crash sequence:
`SetAsStopping -> SetAsDoneStopping -> WaitOnStopTimeout` and confirms it
doesn't panic.

```release-note
Fixed a panic when concurrent StopContainer calls race against the stop lifecycle completing.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented rare panics during container shutdown by improving coordination between stop and timeout handling, making shutdown sequences more reliable and resilient.

* **Tests**
  * Added regression tests to ensure stop/wait operations do not panic when invoked after shutdown completion, reducing regressions and increasing stability.

### DIFF
--- a/internal/oci/container_test.go
+++ b/internal/oci/container_test.go
@@ -1008,6 +1008,23 @@ var _ = t.Describe("Container", func() {
 			}
 		})
 
+		// Regression test for a race between concurrent StopContainer calls.
+		// When a second StopContainer arrives after the first has already
+		// completed (SetAsDoneStopping closed stopTimeoutChan),
+		// WaitOnStopTimeout used to panic on the closed channel.
+		// The stopDone guard ensures it returns early instead.
+		It("should not panic when WaitOnStopTimeout is called after SetAsDoneStopping", func() {
+			ctx := context.Background()
+
+			sut.SetAsStopping()
+
+			sut.SetAsDoneStopping()
+
+			Expect(func() {
+				sut.WaitOnStopTimeout(ctx, 1000)
+			}).ToNot(Panic())
+		})
+
 		It("should clear the watchers slice", func() {
 			// Given
 			ctx := context.Background()
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -70,6 +70,7 @@ type Container struct {
 	created            bool
 	spoofed            bool
 	stopping           bool
+	stopDone           bool
 	stopLock           sync.Mutex
 	// stopTimeoutChan is used to update the stop timeout.
 	// After the container goes into the kill loop, the channel must not be used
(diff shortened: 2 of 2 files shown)


internal/oci/container_test.go
```diff
@@ -1008,6 +1008,20 @@ var _ = t.Describe("Container", func() {
...
+			sut.SetAsStopping()
+
+			// Simulation of the stop loop finishing.
+			sut.SetAsDoneStopping()
+
+			Expect(func() {
+				sut.WaitOnStopTimeout(ctx, 1000)
+			}).ToNot(Panic())
+		})
+
```
bitoku, line 1027: I want more explanation here, about why we want to check this, otherwise developers in the future may have trouble understanding why we check this.

bitoku: @sabujmaity

sabujmaity: @bitoku I have put up the explanation !

bitoku: How about this
```
// Regression test for a race between concurrent StopContainer calls.
// When a second StopContainer arrives after the first has already
// completed (SetAsDoneStopping closed stopTimeoutChan),
// WaitOnStopTimeout used to panic on the closed channel.
// The stopDone guard ensures it returns early instead.
```

sabujmaity: Sure thing. This is a better explanation ! Thanks for the suggestions.

bitoku:
@cri-o/cri-o-maintainers PTAL

PR #9800, opened 2026-03-05
Fixes the locking using a new state

ngopalak-redhat:
#### What type of PR is this?
https://kubernetes.slack.com/archives/CAZH62UR1/p1772497739779279

```release-note

```

### DIFF
--- a/internal/nri/nri.go
+++ b/internal/nri/nri.go
@@ -83,12 +83,11 @@ const (
 )
 
 type local struct {
-	sync.Mutex
-
-	cfg *config.Config
-	nri *nri.Adaptation
-
-	state map[string]State
+	nriMu   sync.Mutex // protects nri field
+	cfg     *config.Config
+	nri     *nri.Adaptation
+	stateMu sync.Mutex // protects state map
+	state   map[string]State
 }
 
 var _ API = &local{}
@@ -149,8 +148,8 @@ func (l *local) Stop() {
 		return
 	}
 
-	l.Lock()
-	defer l.Unlock()
+	l.nriMu.Lock()
+	defer l.nriMu.Unlock()
 
 	l.nri.Stop()
 	l.nri = nil
@@ -161,14 +160,15 @@ func (l *local) RunPodSandbox(ctx context.Context, pod PodSandbox) error {
 		return nil
 	}
 
-	l.Lock()
-	defer l.Unlock()
+	l.nriMu.Lock()
+	nriInstance := l.nri
+	l.nriMu.Unlock()
 
 	request := &nri.RunPodSandboxRequest{
 		Pod: podSandboxToNRI(pod),
 	}
 
-	err := l.nri.RunPodSandbox(ctx, request)
+	err := nriInstance.RunPodSandbox(ctx, request)
 
 	return err
 }
@@ -178,8 +178,9 @@ func (l *local) UpdatePodSandbox(ctx context.Context, pod PodSandbox, overhead, 
 		return nil
 	}
 
-	l.Lock()
-	defer l.Unlock()
+	l.nriMu.Lock()
+	nriInstance := l.nri
+	l.nriMu.Unlock()
 
 	podNri := podSandboxToNRI(pod)
 	request := &nri.UpdatePodSandboxRequest{
@@ -188,7 +189,7 @@ func (l *local) UpdatePodSandbox(ctx context.Context, pod PodSandbox, overhead, 
 		LinuxResources:         resources,
 	}
 
-	_, err := l.nri.UpdatePodSandbox(ctx, request)
+	_, err := nriInstance.UpdatePodSandbox(ctx, request)
 
 	return err
 }
@@ -198,14 +199,15 @@ func (l *local) StopPodSandbox(ctx context.Context, pod PodSandbox) error {
 		return nil
 	}
 
-	l.Lock()
-	defer l.Unlock()
+	l.nriMu.Lock()
+	nriInstance := l.nri
+	l.nriMu.Unlock()
 
 	request := &nri.StopPodSandboxRequest{
 		Pod: podSandboxToNRI(pod),
 	}
 
-	err := l.nri.StopPodSandbox(ctx, request)
+	err := nriInstance.StopPodSandbox(ctx, request)
 
 	return err
 }
@@ -215,14 +217,15 @@ func (l *local) RemovePodSandbox(ctx context.Context, pod PodSandbox) error {
 		return nil
 	}
 
-	l.Lock()
-	defer l.Unlock()
+	l.nriMu.Lock()
+	nriInstance := l.nri
+	l.nriMu.Unlock()
 
 	request := &nri.RemovePodSandboxRequest{
 		Pod: podSandboxToNRI(pod),
 	}
 
-	err := l.nri.RemovePodSandbox(ctx, request)
+	err := nriInstance.RemovePodSandbox(ctx, request)
 
 	return err
 }
@@ -232,16 +235,21 @@ func (l *local) CreateContainer(ctx context.Context, pod PodSandbox, ctr Contain
 		return nil, nil
 	}
 
-	l.Lock()
-	defer l.Unlock()
+	l.nriMu.Lock()
+	nriInstance := l.nri
+	l.nriMu.Unlock()
 
 	request := &nri.CreateContainerRequest{
 		Pod:       podSandboxToNRI(pod),
 		Container: containerToNRI(ctr),
 	}
 
-	response, err := l.nri.CreateContainer(ctx, request)
-	l.setState(request.GetContainer().GetId(), Created)
+	response, err := nriInstance.CreateContainer(ctx, request)
+
+	l.stateMu.Lock()
+	defer l.stateMu.Unlock()
+
+	l.setStateNoLock(request.GetContainer().GetId(), Created)
 
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


klihub:
@ngopalak-redhat That does not look right to me. For instance, nothing protects now against two parallel CreateContainers being executed so that plugins process them in one order but the related applyUpdates are done in the reverse order. If those then touch the same containers, the end result is wrong.

ngopalak-redhat:
@klihub  
So if 2 containers are created in parallel:
  - Thread A calls plugin
  - Thread B calls plugin
  - Say, Thread B finishes first, runs setState() and applyUpdates() for container B
  - Thread A finishes second, runs setState() and applyUpdates() for container A
the updates get applied in the wrong order (B then A)

**The fix:**
- Hold `stateMu` during the state update and `applyUpdates`
Something like this:
```
	l.stateMu.Lock()
	defer l.stateMu.Unlock()

	l.setStateNoLock(request.GetContainer().GetId(), Created)

	if err != nil {
		return nil, err
	}

	if _, err := l.applyUpdates(ctx, response.GetUpdate()); err != nil {
		return nil, err
	}
```

I will try this tomorrow. In the meantime, feel free to suggest a fix if you find one, and I will be happy to test it.

klihub:
> @klihub So if 2 containers are created in parallel:
> 
> * Thread A calls plugin
> * Thread B calls plugin
> * Say, Thread B finishes first, runs setState() and applyUpdates() for container B
> * Thread A finishes second, runs setState() and applyUpdates() for container A
>   the updates get applied in the wrong order (B then A)
> 
> **The fix:**
> 
> * Hold `stateMu` during the state update and `applyUpdates`
>   Something like this:
> 
> ```
> 	l.stateMu.Lock()
> 	defer l.stateMu.Unlock()
> 
> 	l.setStateNoLock(request.GetContainer().GetId(), Created)
> 
> 	if err != nil {
> 		return nil, err
> 	}
> 
> 	if _, err := l.applyUpdates(ctx, response.GetUpdate()); err != nil {
> 		return nil, err
> 	}
> ```
> 
> I will try this tomorrow. In the meantime, feel free to suggest a fix if you find one, and I will be happy to test it.

I'd like to understand what goes wrong now with the attempted fix in place. I was not able to figure out from the stack trace where we get stuck. AFAICT, the above suggested code is basically what we have now in place (except the state altering call is named setState and there is no locking version of it).

ngopalak-redhat:
> > @klihub So if 2 containers are created in parallel:
> > 
> > * Thread A calls plugin
> > * Thread B calls plugin
> > * Say, Thread B finishes first, runs setState() and applyUpdates() for container B
> > * Thread A finishes second, runs setState() and applyUpdates() for container A
> >   the updates get applied in the wrong order (B then A)
> > 
> > **The fix:**
> > 
> > * Hold `stateMu` during the state update and `applyUpdates`
> >   Something like this:
> > 
> > ```
> > 	l.stateMu.Lock()
> > 	defer l.stateMu.Unlock()
> > 
> > 	l.setStateNoLock(request.GetContainer().GetId(), Created)
> > 
> > 	if err != nil {
> > 		return nil, err
> > 	}
> > 
> > 	if _, err := l.applyUpdates(ctx, response.GetUpdate()); err != nil {
> > 		return nil, err
> > 	}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I will try this tomorrow. In the meantime, feel free to suggest a fix if you find one, and I will be happy to test it.
> 
> I'd like to understand what goes wrong now with the attempted fix in place. I was not able to figure out from the stack trace where we get stuck.

@klihub With the fix in the branch https://github.com/klihub/cri-o/tree/fixes/1.34.5/nri-plugin-sync and before that also I could see one behavior: The cri-o process was hung, even `systemctl stop crio` didn't work. I had to kill the crio process to stop it.

With the fix in the branch https://github.com/klihub/cri-o/tree/fixes/1.34.5/nri-plugin-sync there was one improvement, I didn't see pods in the terminating state with zero containers running. Also it took longer for me to get to a state where cri-o hangs.
The thread dumps are shared in the slack thread.

klihub:
> > I'd like to understand what goes wrong now with the attempted fix in place. I was not able to figure out from the stack trace where we get stuck.
> 
> @klihub With the fix in the branch https://github.com/klihub/cri-o/tree/fixes/1.34.5/nri-plugin-sync and before that also I could see one behavior: The cri-o process was hung, even `systemctl stop crio` didn't work. I had to kill the crio process to stop it.
> 
> With the fix in the branch https://github.com/klihub/cri-o/tree/fixes/1.34.5/nri-plugin-sync there was one improvement, I didn't see pods in the terminating state with zero containers running. Also it took longer for me to get to a state where cri-o hangs. The thread dumps are shared in the slack thread.

@ngopalak-redhat Yes, and a big thanks for that stack trace ! But that latter one is already the stack trace I was talking about. So far I haven't been able to understand/spot from it how/why/where we get stuck. If it's a deadlock we should be able to tell the goroutines involved/responsible there... I think. One thing that I found weird is that oci.Container.State() on the StopContainer code path being stuck waiting for a read-lock on its mutex.

ngopalak-redhat:
/payload-job periodic-ci-openshift-release-main-ci-4.22-e2e-aws-ovn-techpreview

klihub:
> > @klihub So if 2 containers are created in parallel:
> > 
> > * Thread A calls plugin
> > * Thread B calls plugin
> > * Say, Thread B finishes first, runs setState() and applyUpdates() for container B
> > * Thread A finishes second, runs setState() and applyUpdates() for container A
> >   the updates get applied in the wrong order (B then A)
> > 
> > **The fix:**
> > 
> > * Hold `stateMu` during the state update and `applyUpdates`
> >   Something like this:
> > 
> > ```
> > 	l.stateMu.Lock()
> > 	defer l.stateMu.Unlock()
> > 
> > 	l.setStateNoLock(request.GetContainer().GetId(), Created)
> > 
> > 	if err != nil {
> > 		return nil, err
> > 	}
> > 
> > 	if _, err := l.applyUpdates(ctx, response.GetUpdate()); err != nil {
> > 		return nil, err
> > 	}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I will try this tomorrow. In the meantime, feel free to suggest a fix if you find one, and I will be happy to test it.
> 
> I'd like to understand what goes wrong now with the attempted fix in place. I was not able to figure out from the stack trace where we get stuck. AFAICT, the above suggested code is basically what we have now in place (except the state altering call is named setState and there is no locking version of it).

Okay, so @askervin figured out the root cause. Basically

- CRI StopContainer being processed
- runtimeOCI.StopContainer starts a goroutine running StopLoopForContainer then returns
- StopLoopForContainer takes and holds opLock, only unlocking when returns via a deferred function call
- container process is apparently stuck in an unkillable state => StopLoopForContainer never returns => opLock is held forever
- we do a oci.Container.State() to acquire the container's state for the NRI notification
- oci.Container.State() tries to take opLock and is

I'll try to roll a workaround for this local to NRI. But I think we should also try to fix the root cause in CRI-O as well.

ngopalak-redhat:
> > > @klihub So if 2 containers are created in parallel:
> > > 
> > > * Thread A calls plugin
> > > * Thread B calls plugin
> > > * Say, Thread B finishes first, runs setState() and applyUpdates() for container B
> > > * Thread A finishes second, runs setState() and applyUpdates() for container A
> > >   the updates get applied in the wrong order (B then A)
> > > 
> > > **The fix:**
> > > 
> > > * Hold `stateMu` during the state update and `applyUpdates`
> > >   Something like this:
> > > 
> > > ```
> > > 	l.stateMu.Lock()
> > > 	defer l.stateMu.Unlock()
> > > 
> > > 	l.setStateNoLock(request.GetContainer().GetId(), Created)
> > > 
> > > 	if err != nil {
> > > 		return nil, err
> > > 	}
> > > 
> > > 	if _, err := l.applyUpdates(ctx, response.GetUpdate()); err != nil {
> > > 		return nil, err
> > > 	}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I will try this tomorrow. In the meantime, feel free to suggest a fix if you find one, and I will be happy to test it.
> > 
> > 
> > I'd like to understand what goes wrong now with the attempted fix in place. I was not able to figure out from the stack trace where we get stuck. AFAICT, the above suggested code is basically what we have now in place (except the state altering call is named setState and there is no locking version of it).
> 
> Okay, so @askervin figured out the root cause. Basically
> 
> * CRI StopContainer being processed
> * runtimeOCI.StopContainer starts a goroutine running StopLoopForContainer then returns
> * StopLoopForContainer takes and holds opLock, only unlocking when returns via a deferred function call
> * container process is apparently stuck in an unkillable state => StopLoopForContainer never returns => opLock is held forever
> * we do a oci.Container.State() to acquire the container's state for the NRI notification
> * oci.Container.State() tries to take opLock and is
> 
> I'll try to roll a workaround for this local to NRI. But I think we should also try to fix the root cause in CRI-O as well.

@klihub I just completed trying the fix in this PR:  https://github.com/cri-o/cri-o/pull/9800/changes/219acf64341c0db3a75b183e5694e4271b772749. 
It seems to solve as I had one successful run of the conformance test.

ngopalak-redhat:
Closed in favor of https://github.com/cri-o/cri-o/pull/9804

PR #9803, opened 2026-03-06
handle ErrNotAnImage in RemoveImage for concurrent deletion idempotency

### LINKED ISSUE #9796: RemoveImage race condition: concurrent calls fail with "identifier is not an image"
### What happened?

The CRI conformance test `[k8s.io] Image Consistency - should not fail on simultaneous RemoveImage calls [Conformance]` fails consistently on CRI-O v1.36.0.

The test pulls `gcr.io/k8s-staging-cri-tools/test-image-tag:test`, then fires 5 concurrent `RemoveImage` RPC calls for the same image. One call succeeds, but the remaining concurrent calls fail with:

  rpc error: code = Unknown desc = delete image: identifier is not an image

The image digest (`00a99e91400bb2195e38a1459084a882f228ab1e0929c7bc80c64c0496dd4c69`) no longer resolves to a valid image by the time the other goroutines attempt deletion.

The test retried 3 times and failed each time -- this is not a flake but a consistent race condition.

Failure location: `sigs.k8s.io/cri-tools/pkg/validate/image_consistency.go:172`

Revealed by conmon test CI: https://github.com/containers/conmon/actions/runs/22673875556/job/65724981026

Concurrent `RemoveImage` calls should be idempotent per the CRI API spec, which states that `RemoveImage` "is idempotent, and must not return an error if the image has already been removed."

If the image was already removed by a concurrent call, subsequent calls should succeed or return gracefully rather than returning an error.

The idempotency improvement shipped in v1.31.0 appears to handle the sequential case (calling `RemoveImage` after the image is already gone) but not the concurrent race where multiple goroutines resolve the image identifier simultaneously and then race on the actual deletion.

Environment:
- CRI-O version: v1.36.0 (commit bcb999d2f21f38e7475b5a77e67c07885e78c7e4)
- critest version: 1.35.0-64-gda706098
- OS: Ubuntu (GitHub Actions CI)
- 112 of 113 CRI conformance tests passed; this is the only failure.

Run the CRI conformance test suite against CRI-O v1.36.0:                                                                                                                  
                                                                
```bash
critest --runtime-endpoint unix:///var/run/crio/crio.sock \
      --ginkgo.focus "should not fail on simultaneous RemoveImage calls"
```
Alternatively, run the full suite -- this is the only failing test out of 113:

```
  critest --runtime-endpoint unix:///var/run/crio/crio.sock
```
  The failure is consistent and reproduces on every attempt (3/3 retries failed in our CI).

_No response_

<details>

- CRI-O version: v1.36.0 (commit bcb999d2f21f38e7475b5a77e67c07885e78c7e4)
- critest version: 1.35.0-64-gda706098
- 112 of 113 CRI conformance tests passed; this is the only failure.

</details>

### OS version

<details>

OS: Ubuntu (GitHub Actions CI)

</details>

<details>
conmon CI: https://github.com/containers/conmon/actions/runs/22673875556/job/65724981026
</details>

jnovy:
When multiple concurrent RemoveImage calls race on the same image, the first call succeeds but subsequent calls fail with "identifier is not an image" (ErrNotAnImage). This happens because the image is already gone from storage by the time the other goroutines attempt deletion.

The existing idempotency check only handled ErrImageUnknown but not ErrNotAnImage, which is the error returned by containers/storage when the image ID no longer resolves. Add ErrNotAnImage to the idempotency checks in both the ID-based and name-based deletion paths.

Fixes #9796

#### What type of PR is this?                                                                                                                                              
 
/kind bug                                                                                                                                                                  
                                                                
When multiple concurrent `RemoveImage` calls race on the same image, the first call succeeds but subsequent calls fail with "identifier is not an image" (`ErrNotAnImage`). This happens because the image is already gone from storage by the time the other goroutines attempt deletion.

The existing idempotency check only handled `ErrImageUnknown` but not `ErrNotAnImage`, which is the error returned by `containers/storage` when the image ID no longer resolves. This PR adds `ErrNotAnImage` to the idempotency checks in both the ID-based and name-based deletion paths.

Fixes #9796

None

```release-note
Fix concurrent RemoveImage race condition by handling ErrNotAnImage as an idempotent deletion result.
```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

## Release Notes

* **Bug Fixes**
  * Image removal operations are now idempotent. Attempting to remove an image that is not found no longer returns an error, allowing safe repeated removal attempts.
  * Improved handling of concurrent deletion scenarios in both direct and indirect image removal paths for more robust operations.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/server/image_remove.go
+++ b/server/image_remove.go
@@ -46,7 +46,7 @@ func (s *Server) removeImage(ctx context.Context, imageRef string) (untagErr err
 		}
 
 		if err := s.ContainerServer.StorageImageServer().DeleteImage(s.config.SystemContext, *id); err != nil {
-			if errors.Is(err, storagetypes.ErrImageUnknown) {
+			if errors.Is(err, storagetypes.ErrImageUnknown) || errors.Is(err, storagetypes.ErrNotAnImage) {
 				// The RemoveImage RPC is idempotent, and must not return an
 				// error if the image has already been removed. Ref:
 				// https://github.com/kubernetes/cri-api/blob/c20fa40/pkg/apis/runtime/v1/api.proto#L156-L157
@@ -96,6 +96,10 @@ func (s *Server) removeImage(ctx context.Context, imageRef string) (untagErr err
 	}
 
 	if !deleted && untagErr != nil {
+		if errors.Is(untagErr, storagetypes.ErrImageUnknown) || errors.Is(untagErr, storagetypes.ErrNotAnImage) {
+			return nil
+		}
+
 		return untagErr
 	}
 
--- a/server/image_remove_test.go
+++ b/server/image_remove_test.go
@@ -140,5 +140,47 @@ var _ = t.Describe("ImageRemove", func() {
 			// Then
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+		// https://github.com/cri-o/cri-o/issues/9796
+		It("should succeed if image was concurrently deleted (ErrNotAnImage)", func() {
+			// Given
+			parsedTestSHA256, err := storage.ParseStorageImageIDFromOutOfProcessData(testSHA256)
+			Expect(err).ToNot(HaveOccurred())
+			gomock.InOrder(
+				imageServerMock.EXPECT().HeuristicallyTryResolvingStringAsIDPrefix(testSHA256).
+					Return(&parsedTestSHA256),
+				imageServerMock.EXPECT().DeleteImage(
+					gomock.Any(), parsedTestSHA256).
+					Return(fmt.Errorf("delete image: %w", storagetypes.ErrNotAnImage)),
+			)
+
+			// When
+			_, err = sut.RemoveImage(context.Background(),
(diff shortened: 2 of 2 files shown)


server/image_remove.go
```diff
@@ -96,6 +96,10 @@ func (s *Server) removeImage(ctx context.Context, imageRef string) (untagErr err
 	}
 
 	if !deleted && untagErr != nil {
+		if errors.Is(untagErr, storagetypes.ErrImageUnknown) || errors.Is(untagErr, storagetypes.ErrNotAnImage) {
```
saschagrunert, line 99: Good catch adding `ErrImageUnknown` here as well. The existing code only had `ErrImageUnknown` handling in the ID-based path (line 49) and `ErrNotAnImage` handling via `statusErr` at line 119. This closes the gap where `UntagImage` could also return `ErrImageUnknown` during a concurrent deletion race.

PR #9804, opened 2026-03-06
nri: fix a deadlock and a livelock, block plugin sync during other event processing.

klihub:
#### What type of PR is this?

/kind bug

Recent long-running conformance tests have uncovered a deadlock in NRI event processing. Fix a deadlock by lock inversion on 2 different code paths (and eliminate an unnecessary lock). Acquire state of container with a timeout to avoid an indefinite livelock if a container (shutdown) is stuck (probably with the container's init process being in an unkillable state). Also, block plugin synchronization during other event processing.

None

None

```release-note
Fix NRI event processing deadlock.
```

### DIFF
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -1330,15 +1330,15 @@ func (s *Server) createSandboxContainer(ctx context.Context, ctr container.Conta
 
 	hooks := s.hooksRetriever.Get(ctx, sb.RuntimeHandler(), sb.Annotations())
 
-	if err := s.nri.createContainer(ctx, specgen, sb, ociContainer); err != nil {
-		return nil, err
-	}
-
 	defer func() {
 		if retErr != nil {
 			s.nri.undoCreateContainer(ctx, specgen, sb, ociContainer)
 		}
 	}()
+
+	if err := s.nri.createContainer(ctx, specgen, sb, ociContainer); err != nil {
+		return nil, err
+	}
 
 	if hooks != nil {
 		if err := hooks.PreCreate(ctx, specgen, sb, ociContainer); err != nil {
--- a/server/container_create.go
+++ b/server/container_create.go
@@ -466,6 +466,8 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 		return nil, fmt.Errorf("CreateContainer failed as the sandbox was stopped: %s", sb.ID())
 	}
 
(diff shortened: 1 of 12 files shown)


server/nri-api.go
```diff
@@ -683,6 +696,37 @@ func (c *criContainer) GetState() api.ContainerState {
 	return api.ContainerState_CONTAINER_UNKNOWN
 }
 
+func (c *criContainer) asyncGetState() *oci.ContainerState {
```
haircommander: should we still check if c is nil here?

klihub: I think we were checking earlier if c.ctr != nil, not if c != nil, and this is still checked but now in `asyncGetState()`. Anyway I had to rework this a bit more, to help address two more code paths where we could still eventually get blocked on the same oci.Container.opLock if the container is stuck in an unkillable state.

server/container_create.go
```diff
@@ -466,6 +466,8 @@ func (s *Server) CreateContainer(ctx context.Context, req *types.CreateContainer
 		return nil, fmt.Errorf("CreateContainer failed as the sandbox was stopped: %s", sb.ID())
 	}
 
+	defer s.nri.BlockPluginSync().Unblock()
```
bitoku, line 469: I don't understand what we do here. Does it block and unblock at the same time?

bitoku (commented):
Would you mind creating a PR in main ? I can do backport work.

klihub:
> Would you mind creating a PR in main ? I can do backport work.

I have a branch ready for the same changes against main. I'll file a PR from that, too.

server/container_create.go
klihub: No, it blocks plugin synchronization (that's the s.nri.BlockPluginSync() which is immediately evaluated) and does a deferred Unblock() once the function returns. It is equivalent to writing

```golang
b := s.nri.BlockPluginSync()
defer b.Unblock()
```

but without using an explicit and otherwise unused temporary variable.

bitoku: ah good to know!
I didn't know that only the last function call is deferred in golang.
make sense then

bitoku (commented):
/lgtm

ozhuraki (approved):
LGTM

bitoku:
closing this in favor of https://github.com/cri-o/cri-o/pull/9835

PR #9810, opened 2026-03-12
Test Case: Ensure container stopping and creating can happen in parallel

ngopalak-redhat:
#### What type of PR is this?

/kind ci

An attempt to recreate the locking issue discussed here: https://kubernetes.slack.com/archives/CAZH62UR1/p1772497739779279

Basically I made an attempt to create containers while there are other containers stopping in parallel.

@klihub This didn't recreate the problem but I thought its worth having this test case. I have a few more test cases which I can add on the same line which tests the code around this with more load.

```release-note
None
```

## Summary by CodeRabbit

* **Documentation**
  * Added a "CI Test Logs" section documenting where CI test output is stored.

* **Tests**
  * Added a new integration test validating that stopping a sandbox obtains an exclusive lock that blocks concurrent container creation, with timing and cleanup checks to ensure correct lock behavior.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -111,3 +111,11 @@ function teardown() {
   crictl runtimeversion
 }
 ```
+
+## CI Test Logs
+
+For the `ci/prow/ci-cgroupv2-integration` CI run, BATS test output can be found at:
+
+```text
+artifacts/cgroupv2-integration/cri-o-gather/artifacts/testout.txt
+```
--- a/test/sandbox_stop_locking.bats
+++ b/test/sandbox_stop_locking.bats
@@ -1,0 +1,67 @@
+#!/usr/bin/env bats
+# vim: set syntax=sh:
+
+load helpers
+
+function setup() {
+	setup_test
+}
+
+function teardown() {
+	cleanup_test
+}
+
+# Helper function to create container config that ignores SIGTERM
+# This simulates a container that won't terminate gracefully, forcing
+# CRI-O to wait through the full timeout period before sending SIGKILL
+function create_nonterminating_container_config() {
+	local output_path="$1"
+	# This forces CRI-O to wait through full timeout before SIGKILL
+	jq '.command = ["sh", "-c", "trap '"'"''"'"' TERM; while true; do sleep 1; done"]' \
+		"$TESTDATA"/container_sleep.json > "$output_path"
+}
+
+@test "locking CreateContainer blocks when StopPodSandbox holds exclusive lock" {
+	start_crio
+
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	create_nonterminating_container_config "$TESTDIR/container_noterminate.json"
+	ctr_id=$(crictl create "$pod_id" "$TESTDIR/container_noterminate.json" "$TESTDATA"/sandbox_config.json)
+	crictl start "$ctr_id"
+
+	# Container stopping in background
+	crictl stopp "$pod_id" &
+	stop_pid=$!
+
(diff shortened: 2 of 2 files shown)


klihub:
> This didn't recreate the problem but I thought its worth having this test case. I have a few more test cases which I can add on the same line which tests the code around this with more load.

@ngopalak-redhat I suspect that we won't be able to reproduce it with a bats test. I've tried to do that in my test setup, using the same CRI-O tree and identical crun version, a criminally braindead podspec+workload which deliberately ignores SIGTERM and does not specify a terminationGracePeriodSeconds, and running some NRI plugins, and some racy creation and stopping of multiple workloads, and I have not been able to. The workloads sometimes lingers for long-ish times in a Terminating state but always recovers being able to remove the pods (transitioning via a short Error state).

ngopalak-redhat:
> > This didn't recreate the problem but I thought its worth having this test case. I have a few more test cases which I can add on the same line which tests the code around this with more load.
> 
> @ngopalak-redhat I suspect that we won't be able to reproduce it with a bats test. I've tried to do that in my test setup, using the same CRI-O tree and identical crun version, a criminally braindead podspec+workload which deliberately ignores SIGTERM and does not specify a terminationGracePeriodSeconds, and running some NRI plugins, and some racy creation and stopping of multiple workloads, and I have not been able to. The workloads sometimes lingers for long-ish times in a Terminating state but always recovers being able to remove the pods (transitioning via a short Error state).

Agreed, its best to check-in whatever good test we have written during the process of solving this issue. Few other test cases I wrote were to create a large number of containers randomly and stopping them. I need to polish them to make sure it consistently passes. Other idea I have is to try https://kube-burner.github.io/kube-burner/latest/ 

Can we add this test to begin with?

ngopalak-redhat:
> > > This didn't recreate the problem but I thought its worth having this test case. I have a few more test cases which I can add on the same line which tests the code around this with more load.
> > 
> > 
> > @ngopalak-redhat I suspect that we won't be able to reproduce it with a bats test. I've tried to do that in my test setup, using the same CRI-O tree and identical crun version, a criminally braindead podspec+workload which deliberately ignores SIGTERM and does not specify a terminationGracePeriodSeconds, and running some NRI plugins, and some racy creation and stopping of multiple workloads, and I have not been able to. The workloads sometimes lingers for long-ish times in a Terminating state but always recovers being able to remove the pods (transitioning via a short Error state).
> 
> Agreed, its best to check-in whatever good test we have written during the process of solving this issue. Few other test cases I wrote were to create a large number of containers randomly and stopping them. I need to polish them to make sure it consistently passes. Other idea I have is to try https://kube-burner.github.io/kube-burner/latest/
> 
> Can we add this test to begin with?

@klihub @bitoku Can you review this? I feel its good to have some test cases that test the parallelism in e2e.

klihub:
@ngopalak-redhat If you want to recreate the exact low-level runtime misbehavior induced problems we were running into, it is fairly easy now that we figured out what was going wrong. The key is to add a level of indirection between what cri-o thinks is the low-level runtime and the real low-level runtime, so that you can switch from one (crun) to another (runc) without any involvement from cri-o itself.

Here is, for instance, one way that should emulate the bug artificially in a test case:

1. make sure both crun and runc are installed.
2. add a symlink /usr/local/bin/xrunx pointing initially to the crun binary
3. configure crio to use /usr/local/bin/xrunx as the default runtime handler
4. create a pod and a container in it
5. update the symlink to point to the runc binary
6. try to stop the container

It should fail with an error from the low-level runtime saying the container does not exist. This is now an equivalent state where we ended up with the nvct bug.

PR #9822, opened 2026-03-16
fix: ensure CRI container event ordering

### LINKED ISSUE #9738: CRI Container Events delivered out of order
### What happened?

Observed while testing CRI events with Go client on Ubuntu 22.04 crio v 1.35
Events for container start and container stop are delivered out of order for short lived container
Eg a test container that only does "echo HELLO" and exits

Would expect events for short lived container for container Start and Stop to be delivered in order via CRI events , same as for any other container 

On Ubunti22 machine with crio v 1.35 installed (non k8s environment)
Use simple Go client to stream CRI events (sample code below). 
Use crictl to start pod with a single container that executes command echo HELLO and exits
Replace the above container with a container that executes sleep 1 to observe the difference

Sample code:

```
import (
	"context"
	"encoding/json"
	"fmt"
	"io"
	"log"
	"net"
	"os"
	"strings"
	"sync"
	"time"

	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials/insecure"
	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
)

type CRIEventsMonitor struct {
	runtimeSockets map[string]string
	clients        map[string]pb.RuntimeServiceClient
	connections    map[string]*grpc.ClientConn
	socketPath     string
	mu             sync.RWMutex
}

type ContainerEvent struct {
	Timestamp int64                 `json:"timestamp"`
	Plugin    string                `json:"plugin"`
	EventType string                `json:"event_type"`
	Runtime   string                `json:"runtime"`
	Container *pb.ContainerStatus   `json:"container,omitempty"`
	Metadata  map[string]interface{} `json:"metadata"`
}

func NewCRIEventsMonitor(socketPath string) *CRIEventsMonitor {
	return &CRIEventsMonitor{
		runtimeSockets: map[string]string{
			"crio":       "unix:///var/run/crio/crio.sock"
		   
		},
		clients:     make(map[string]pb.RuntimeServiceClient),
		connections: make(map[string]*grpc.ClientConn),
		socketPath:  socketPath,
	}
}

func (m *CRIEventsMonitor) isRuntimeAvailable(socket string) bool {
	if _, err := os.Stat(strings.TrimPrefix(socket, "unix://")); err != nil {
		return false
	}

	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
	defer cancel()

	conn, err := grpc.DialContext(ctx, socket,
		grpc.WithTransportCredentials(insecure.NewCredentials()),
		grpc.WithBlock(),
	)
	if err != nil {
		return false
	}
	conn.Close()
	return true
}

func (m *CRIEventsMonitor) connectToRuntime(name, socket string) error {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	conn, err := grpc.DialContext(ctx, socket,
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
	if err != nil {
		return fmt.Errorf("failed to connect to %s at %s: %w", name, socket, err)
	}

	client := pb.NewRuntimeServiceClient(conn)

	// Test the connection
	_, err = client.Version(context.Background(), &pb.VersionRequest{})
	if err != nil {
		conn.Close()
		return fmt.Errorf("failed to get version from %s: %w", name, err)
	}

	m.mu.Lock()
	m.clients[name] = client
	m.connections[name] = conn
	m.mu.Unlock()

	log.Printf("✅ Connected to CRI runtime: %s (%s)", name, socket)
	return nil
}

func (m *CRIEventsMonitor) Start() error {
	log.Printf("🚀 Starting CRI Events Monitor...")
	log.Printf("📡 Events will be sent to: %s", m.socketPath)

	// Connect to available runtimes
	connectedCount := 0
	for name, socket := range m.runtimeSockets {
		if m.isRuntimeAvailable(socket) {
			if err := m.connectToRuntime(name, socket); err != nil {
				log.Printf("⚠️  %v", err)
			} else {
				connectedCount++
			}
		} else {
			log.Printf("❌ CRI runtime not available: %s (%s)", name, socket)
		}
	}

	if connectedCount == 0 {
		return fmt.Errorf("no CRI runtimes available")
	}

	// Start event streaming for each runtime
	var wg sync.WaitGroup
	for name := range m.clients {
		wg.Add(1)
		go func(runtimeName string) {
			defer wg.Done()
			m.streamEvents(runtimeName)
		}(name)
	}

	log.Printf("✅ Streaming events from %d CRI runtime(s)", connectedCount)
	wg.Wait()
	return nil
}

func (m *CRIEventsMonitor) streamEvents(runtimeName string) {
	log.Printf("🔊 Starting event stream for runtime: %s", runtimeName)

	m.mu.RLock()
	client, exists := m.clients[runtimeName]
	m.mu.RUnlock()

	if !exists {
		log.Printf("❌ Runtime client not found: %s", runtimeName)
		return
	}

	for {
		// Create streaming request
		req := &pb.GetEventsRequest{}

		ctx := context.Background()
		stream, err := client.GetContainerEvents(ctx, req)
		if err != nil {
			log.Printf("❌ Failed to start event stream for %s: %v", runtimeName, err)
			time.Sleep(5 * time.Second)
			continue
		}

		log.Printf("📡 Event stream active for %s", runtimeName)

		// Read events from stream
		for {
			event, err := stream.Recv()
			if err == io.EOF {
				log.Printf("📡 Event stream ended for %s", runtimeName)
				break
			}
			if err != nil {
				log.Printf("❌ Event stream error for %s: %v", runtimeName, err)
				break
			}

		}

	}
}

```
```

I'd like to understand if this is by design - ie events are kept async, maybe processed by different threads internally, or this is a bug that should be fixed. 

<details>

```console
$ crio --version
# paste output here
```

```console
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

<details>

</details>

sabujmaity:
/kind bug

There is a race condition in the CRI container event delivery path. When a
short-lived container exits before the `StartContainer` gRPC handler finishes
generating the `CONTAINER_STARTED_EVENT`, the exit-file watcher (`handleExit`)
can emit `CONTAINER_STOPPED_EVENT` first because it runs in an independent
goroutine with no ordering guarantee.

The sequence is:
1. `StartContainer` -> `Runtime().StartContainer()` starts the process, **sets `state.Started`**, releases `opLock`
2. Container exits immediately, conmon writes exit file
3. `handleExit` goroutine spawns -> calls `generateCRIEvent(STOPPED)`
4. `StartContainer` handler calls `generateCRIEvent(STARTED)` concurrently
5. `STOPPED` can reach `ContainerEventsChan` before `STARTED`

This adds a per-container `startedEvtDone` channel barrier. `handleExit` waits
on it before emitting `CONTAINER_STOPPED_EVENT`, and `StartContainer` closes
it after `CONTAINER_STARTED_EVENT` is sent.

For containers restored after a CRI-O restart, the barrier is signaled during
`LoadContainer` and `LoadSandbox` so that `handleExit` does not block
indefinitely. The same applies to the checkpoint/restore path.

Additionally, this fixes a panic in `NewSpoofedContainer` (used when
`drop_infra_ctr = true`) where the `startedEvtDone` channel was manually
closed at construction time without consuming the `sync.Once`. When
`finalizeSandboxCreation` later called `SignalStartedEventDone()`,
`sync.Once.Do` fired and tried to close an already-closed channel, causing
`panic: close of closed channel`. The fix uses `SignalStartedEventDone()`
in the constructor instead, which atomically consumes the `Once` and closes
the channel.

A defensive `SignalStartedEventDone()` is also added on the `StartContainer`
error path so `handleExit` can never block indefinitely if the start fails
after the OCI runtime has launched the process.

Fixes #9738

#### Special notes for reviewer:

Same pattern as the existing `stopWatchers`/`stopDone`/`SetAsDoneStopping`
in the `Container` struct -- just a channel + `sync.Once` barrier.

An earlier revision used `state.Started.IsZero()` to skip the wait on
restart, but `runtimeOCI.StartContainer()` sets `state.Started` before
returning to the gRPC handler, so `handleExit` always bypassed the barrier.
Removed that and signal from `LoadContainer`/`LoadSandbox` instead.

The `NewSpoofedContainer` panic (`close of closed channel`) was caught
while running `ctr_events.bats` locally -- the channel was closed manually
at construction without going through `sync.Once`.

```release-note
Fixed out-of-order CRI container event delivery for short-lived containers where STOPPED could arrive before STARTED.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure container "started" signaling is always completed across normal, restore, and failure paths; the server now waits for that completion before emitting CONTAINER_STOPPED events to prevent out-of-order start/stop events.

* **Tests**
  * Added unit and integration tests validating start-event synchronization and CRI event ordering for short-lived, multiple, and immediate-exit containers.

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -94,6 +94,11 @@ type Container struct {
 	// When set, the exec process will spawn on this cgroup.
 	// If this is used, InfraCtrCPUSet will be ignored for the exec operation.
 	execCgroupPath string
+	// eventMu serializes CRI event emission so that handleExit cannot
+	// emit CONTAINER_STOPPED_EVENT before CONTAINER_STARTED_EVENT.
+	// Held by StartContainer across the runtime start + STARTED event,
+	// and by handleExit around the STOPPED event.
+	eventMu sync.Mutex
 }
 
 func (c *Container) CRIAttributes() *types.ContainerAttributes {
@@ -211,6 +216,7 @@ func NewSpoofedContainer(id, name string, labels map[string]string, sandbox stri
 	state := &ContainerState{}
 	state.Created = created
 	state.Started = created
+
 	c := &Container{
 		criContainer: &types.Container{
 			Id:           id,
@@ -729,6 +735,18 @@ func (c *Container) SetAsDoneStopping() {
 	c.stopWatchers = make([]chan struct{}, 0)
 	close(c.stopTimeoutChan)
 	c.stopLock.Unlock()
+}
+
+// LockEventOrder acquires the event-ordering mutex.
+// It is held by StartContainer across the runtime start + STARTED event
+// emission, ensuring handleExit cannot emit STOPPED before STARTED.
+func (c *Container) LockEventOrder() {
+	c.eventMu.Lock()
+}
+
+// UnlockEventOrder releases the event-ordering mutex.
+func (c *Container) UnlockEventOrder() {
+	c.eventMu.Unlock()
 }
 
 func (c *Container) AddManagedPIDNamespace(ns nsmgr.Namespace) {
--- a/internal/oci/container_test.go
+++ b/internal/oci/container_test.go
@@ -1055,6 +1055,38 @@ var _ = t.Describe("Container", func() {
 			}).To(Panic())
 		})
 	})
+
+	t.Describe("EventOrderMutex", func() {
+		It("should serialize LockEventOrder callers", func() {
+			done := make(chan struct{})
+
+			sut.LockEventOrder()
+
+			go func() {
+				sut.LockEventOrder()
+				sut.UnlockEventOrder()
+				close(done)
+			}()
+
+			Consistently(done, 100*time.Millisecond).ShouldNot(BeClosed())
+
+			sut.UnlockEventOrder()
+
+			Eventually(done, 100*time.Millisecond).Should(BeClosed())
+		})
+
+		It("should not block when mutex is not held", func() {
+			done := make(chan struct{})
+
+			go func() {
+				sut.LockEventOrder()
+				sut.UnlockEventOrder()
+				close(done)
+			}()
+
+			Eventually(done, 100*time.Millisecond).Should(BeClosed())
+		})
+	})
 })
 
 var _ = t.Describe("SpoofedContainer", func() {
--- a/server/container_events_test.go
+++ b/server/container_events_test.go
@@ -1,6 +1,7 @@
 package server_test
 
 import (
+	"sync"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -63,7 +64,13 @@ var _ = t.Describe("ContainerEvents", func() {
 				client2.EXPECT().Send(&events[i]).Return(nil)
 			}
 
+			var wg sync.WaitGroup
+			wg.Add(2)
+
 			recv := func(ces types.RuntimeService_GetContainerEventsServer) {
+				defer GinkgoRecover()
+				defer wg.Done()
+
 				err := sut.GetContainerEvents(nil, ces)
 				Expect(err).ToNot(HaveOccurred())
 			}
@@ -79,6 +86,8 @@ var _ = t.Describe("ContainerEvents", func() {
 			for _, event := range events {
 				sut.ContainerEventsChan <- event
 			}
+
+			wg.Wait()
 		})
 	})
 })
--- a/server/container_start.go
+++ b/server/container_start.go
@@ -27,6 +27,9 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
 	}
 
 	if c.Restore() {
+		c.LockEventOrder()
+		defer c.UnlockEventOrder()
+
 		// If the create command found a checkpoint image, the container
 		// has the restore flag set to true. At this point we need to jump
 		// into the restore code.
@@ -74,6 +77,9 @@ func (s *Server) StartContainer(ctx context.Context, req *types.StartContainerRe
 	if err := s.nri.startContainer(ctx, sandbox, c); err != nil {
 		log.Warnf(ctx, "NRI start failed for container %q: %v", c.ID(), err)
 	}
+
+	c.LockEventOrder()
+	defer c.UnlockEventOrder()
 
 	defer func() {
 		// if the call to StartContainer fails below we still want to fill
--- a/server/sandbox_run_freebsd.go
+++ b/server/sandbox_run_freebsd.go
@@ -539,6 +539,8 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 	}
 
 	s.generateCRIEvent(ctx, sb.InfraContainer(), types.ContainerEventType_CONTAINER_CREATED_EVENT)
+	container.LockEventOrder()
+	defer container.UnlockEventOrder()
 	if err := s.ContainerServer.Runtime().StartContainer(ctx, container); err != nil {
 		return nil, err
 	}
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -744,6 +744,9 @@ func (s *Server) runPodSandbox(ctx context.Context, req *types.RunPodSandboxRequ
 		return nil, err
 	}
 
+	container.LockEventOrder()
+	defer container.UnlockEventOrder()
+
 	if err := s.createAndStartInfraContainer(ctx, sb, container, sandboxIDMappings, sboxName, resourceCleaner); err != nil {
 		return nil, err
 	}
--- a/server/server.go
+++ b/server/server.go
@@ -897,7 +897,9 @@ func (s *Server) handleExit(ctx context.Context, event fsnotify.Event) {
 
 	s.postStopCleanup(ctx, c, sb, s.hooksRetriever.Get(ctx, sb.RuntimeHandler(), sb.Annotations()))
 
+	c.LockEventOrder()
 	s.generateCRIEvent(ctx, c, types.ContainerEventType_CONTAINER_STOPPED_EVENT)
+	c.UnlockEventOrder()
 
 	if err := os.Remove(event.Name); err != nil {
 		log.Warnf(ctx, "Failed to remove exit file: %v", err)
--- a/test/ctr_events.bats
+++ b/test/ctr_events.bats
@@ -1,0 +1,99 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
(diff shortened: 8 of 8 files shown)


sabujmaity:
### Update: Fixed panic in `drop_infra_ctr` path + added integration tests

**Bug found during local integration testing:**

`NewSpoofedContainer` (used when `drop_infra_ctr = true`) manually closed the
`startedEvtDone` channel at construction time without consuming the `sync.Once`.
When `finalizeSandboxCreation` later called `SignalStartedEventDone()`,
`sync.Once.Do` fired and tried to close an already-closed channel — **panic:
close of closed channel**.

**Fix:** Replace the manual `close()` in `NewSpoofedContainer` with a call to
`c.SignalStartedEventDone()`, which atomically consumes the `Once` and closes
the channel.

**Additional hardening:** Added a defensive `c.SignalStartedEventDone()` on the
`StartContainer` error path to ensure `handleExit` can never block indefinitely
if the start fails after the OCI runtime has launched the process.

**New integration tests** (`test/ctr_events.bats`):
- Short-lived container (`echo HELLO`)
- Multiple short-lived containers (5 iterations)
- Immediate-exit container (`true`)

All verify `CONTAINER_STARTED_EVENT` log line appears before
`CONTAINER_STOPPED_EVENT` for the same container.

**Local test results:**
| Suite | Result |
|---|---|
| Unit tests (`internal/oci`) | 90/90 pass |
| `ctr_events.bats` | 3/3 pass |
| `ctr.bats` | 73/74 pass (1 pre-existing) |
| `pod.bats` | 30/30 pass |

sairameshv (commented):
/ok-to-test

bitoku (commented):
I don't love the idea of having new `startedEvtDone`/`startedEvtOnce` just for this fix.
While the container operation is ongoing, the container acquires `opLock`, and this prevents the container state race condition. I think we can do the same for the event as well.

sabujmaity:
> I don't love the idea of having new `startedEvtDone`/`startedEvtOnce` just for this fix. While the container operation is ongoing, the container acquires `opLock`, and this prevents the container state race condition. I think we can do the same for the event as well.

I did look into using `opLock` for this, but the main issue is that `opLock` gets released inside `runtimeOCI.StartContainer()` before we emit the STARTED event in the server handler. So there's still a window for the race. Also `handleExit` doesn't acquire `opLock` today, and adding it there risks contention with `StopLoopForContainer` which holds it for the whole kill loop.
The channel approach follows the same pattern as `stopWatchers/stopDone/SetAsDoneStopping` that's already in Container. It felt like the safest fit here.

But can rethink if you have a specific idea for how to bridge the `runtime/server` lock scope though!

bitoku:
I'm not that confident about my suggestion tbh, but the CRI event should be generated when the state is changed, and all state change should happen in opLock guard.

> I did look into using opLock for this, but the main issue is that opLock gets released inside runtimeOCI.StartContainer() before we emit the STARTED event in the server handler.

Can't we fix that too?

> Also handleExit doesn't acquire opLock today, and adding it there risks contention with StopLoopForContainer which holds it for the whole kill loop.

In handleExit, we call StopContainer in postStopCleanup, we can emit the event there.

> The channel approach follows the same pattern as stopWatchers/stopDone/SetAsDoneStopping that's already in Container. It felt like the safest fit here.

I don't want to have another complexity anymore in here if possible :)

sabujmaity:
> I'm not that confident about my suggestion tbh, but the CRI event should be generated when the state is changed, and all state change should happen in opLock guard.
> 
> > I did look into using opLock for this, but the main issue is that opLock gets released inside runtimeOCI.StartContainer() before we emit the STARTED event in the server handler.
> 
> Can't we fix that too?
> 
> > Also handleExit doesn't acquire opLock today, and adding it there risks contention with StopLoopForContainer which holds it for the whole kill loop.
> 
> In handleExit, we call StopContainer, we can emit the event there.
> 
> > The channel approach follows the same pattern as stopWatchers/stopDone/SetAsDoneStopping that's already in Container. It felt like the safest fit here.
> 
> I don't want to have another complexity anymore in here if possible :)

Will try to workaround it. Thanks for the suggestions.

sabujmaity:
> I'm not that confident about my suggestion tbh, but the CRI event should be generated when the state is changed, and all state change should happen in opLock guard.
> 
Agree in principle, but `generateCRIEvent ` calls `Runtime().UpdateContainerStatus()` on the same container as its first step which acquires` opLock `. Go mutexes aren't reentrant so holding opLock while emitting an event deadlocks. Replaced the channel/Once with a dedicated `eventMu sync.Mutex` that serializes only the event emission `StartContainer` holds it across runtime start + STARTED event, blocking `handleExit` from emitting STOPPED before that completes.
> 
> Can't we fix that too?
> 
That would mean splitting `RuntimeImpl.StartContainer()` or having it return without releasing `opLock` changing the contract for `runtimeOCI`, `runtimeVM`, and `runtimePod`, all of which manage their own `opLock` internally today. And even with that refactor, the deadlock from point 1 still applies since `generateCRIEvent` re-acquires `opLock` via `UpdateContainerStatus`. The `eventMu` gives the same ordering guarantee without touching the runtime abstraction.
> 
> In handleExit, we call StopContainer in postStopCleanup, we can emit the event there.
> 
That StopContainer is `StorageRuntimeServer().StopContainer()`, it's the storage-layer unmount, not the OCI runtime stop, so it doesn't hold opLock either. Also `postStopCleanup` is called from both `handleExit` and the explicit `stopContainer` gRPC path , so emitting there would double-emit for explicitly stopped containers.
> 
> I don't want to have another complexity anymore in here if possible :)
>
Dropped the channel/Once entirely. Replaced 2 fields (`chan struct{}` + `sync.Once`) with 1 (`sync.Mutex`). No channels, no `select`, no close-of-closed-channel risk. Load/restore paths need zero changes since a zero-value mutex is already unlocked 2 fewer call sites than before.
>

bitoku:
> That StopContainer is StorageRuntimeServer().StopContainer(), it's the storage-layer unmount, not the OCI runtime stop, so it doesn't hold opLock either. Also postStopCleanup is called from both handleExit and the explicit stopContainer gRPC path , so emitting there would double-emit for explicitly stopped containers.

Right. sorry about that.

Still I think we can do without a new lock.

The reason why we can't just put the `generateCRIEvent` in `opLock` is that it acquires `opLock` inside (`UpdateContainerStatus` calls in `generateCRIEvent` and `server.ContainerStatus`).
So the next question I had was why we can't remove them.
I haven't checked throughly, but it looks doable though the diff would be larger than the current one.

PR #9824, opened 2026-03-17
WIP: Add storage deduplication support using reflinks

harche:
/kind feature

#### What type of PR is this?

/kind feature

Adds storage deduplication support to CRI-O using filesystem-level reflinks (copy-on-write clones). Identical files across container image layers are deduplicated at the filesystem level, reducing disk usage. Since reflinks are copy-on-write, modifying one copy does not affect others.

This is useful for storage-constrained environments (SNO, edge, small clusters) where multiple images share common base layers or files.

**Two ways to trigger dedup:**

| Trigger | When | Details |
|---------|------|---------|
| Startup | `enable_storage_dedup = true` in config | Runs in background goroutine during server startup |
| On-demand | `crio dedup` CLI command | Run while CRI-O is stopped; admin controls timing |

#### Changes

- `pkg/config/config.go`: `EnableStorageDedup` field on `ImageConfig`
- `pkg/config/template.go`: Config template for `crio config --default`
- `internal/criocli/criocli.go`: `--enable-storage-dedup` CLI flag
- `internal/criocli/dedup.go`: `crio dedup` subcommand
- `server/dedup.go`: `RunDedup()` — calls `store.Dedup()` with SHA256 hashing
- `server/server.go`: Startup dedup goroutine when enabled
- `cmd/crio/main.go`: Register `DedupCommand`
- `docs/crio.conf.5.md`: Man page documentation

<details>
<summary><b>E2E test: 26% storage reduction on XFS with reflinks</b></summary>

**Setup:** 3 container images built with podman on an XFS filesystem (`reflink=1`). Each image has a **different** layer (different digest) containing the **same** 10 MB file (`shared.dat`), plus a unique file to force distinct layers.

```
Images in storage:
  localhost/image-a:latest  10.5 MB
  localhost/image-b:latest  10.5 MB
  localhost/image-c:latest  10.5 MB

Copies of shared.dat across layers: 3
```

**Before dedup** — 3 separate copies of `shared.dat`, each with its own inode:

```
XFS block usage: 76 MB

Copies of shared.dat (each 10 MB, different inodes):
  layer 0862f10c3d23  10M  inode=1048726
  layer cd87f17f7901  10M  inode=3145870
  layer ee6668357256  10M  inode=2117776
```

**Running `crio dedup`:**

```
Starting storage deduplication
Storage deduplication complete: 20MiB saved
```

**After dedup:**

```
XFS block usage: 56 MB
```

**Results:**

```
  Disk usage before dedup:       76 MB
  Disk usage after dedup:        56 MB
  Space saved:                   20 MB (20484 KB)
  Reduction:                     26%
```

The 3 copies of `shared.dat` still appear as separate files (separate inodes), but at the filesystem level they now share physical blocks via copy-on-write. Writing to any copy triggers a transparent clone — no data corruption risk.

</details>

<details>
<summary><b>Unit tests (Ginkgo)</b></summary>

`server/dedup_test.go` covers:
- Successful dedup with savings
- Successful dedup with no savings (empty storage)
- `ENOTSUP` error → returns error indicating unsupported filesystem
- `EOPNOTSUPP` error → same behavior
- Other errors → returns wrapped error
- Server creation with `EnableStorageDedup = true`

</details>

<details>
<summary><b>Integration tests (BATS)</b></summary>

`test/dedup.bats` covers:
- Startup dedup runs when `enable_storage_dedup = true`
- No dedup when `enable_storage_dedup = false`
- `crio dedup` CLI command succeeds
- Config option appears in `crio config --default`

</details>

None

- Requires filesystem support for reflinks (XFS with `reflink=1` or Btrfs). On unsupported filesystems, dedup returns an error and logs a warning.
- The `crio dedup` command should be run while CRI-O is stopped to avoid lock contention — `store.Dedup()` acquires an exclusive write lock on the layer store.
- Startup dedup runs before kubelet connects, so no pull contention in practice.

```release-note
Add storage deduplication support using filesystem reflinks. Enable via
`enable_storage_dedup = true` in crio.conf for automatic dedup on startup,
or run `crio dedup` manually. Reduces disk usage by deduplicating identical
files across image layers using copy-on-write clones. Requires XFS with
reflink=1 or Btrfs.
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -501,6 +501,14 @@ func New(
 	deletedImages := s.restore(ctx)
 	s.wipeIfAppropriate(ctx, deletedImages)
 
+	if config.EnableStorageDedup {
+		go func() {
+			if err := RunDedup(ctx, s.Store()); err != nil {
+				log.Warnf(ctx, "%v", err)
+			}
+		}()
+	}
+
 	var bindAddressStr string
 
 	bindAddress := net.ParseIP(config.StreamAddress)
--- a/cmd/crio/main.go
+++ b/cmd/crio/main.go
@@ -161,6 +161,7 @@ func main() {
 	app.Commands = append(app.Commands,
 		criocli.CheckCommand,
 		criocli.ConfigCommand,
+		criocli.DedupCommand,
 		criocli.PublishCommand,
 		criocli.StatusCommand,
 		criocli.VersionCommand,
--- a/completions/bash/crio
+++ b/completions/bash/crio
@@ -6,6 +6,7 @@ _cli_bash_autocomplete() {
 complete
 completion
 config
+dedup
 man
 markdown
 md
@@ -56,6 +57,7 @@ h
 --enable-nri
 --enable-pod-events
 --enable-profile-unix-socket
+--enable-storage-dedup
 --enable-tracing
 --gid-mappings
 --global-auth-file
--- a/completions/fish/crio.fish
+++ b/completions/fish/crio.fish
@@ -2,7 +2,7 @@
 
 function __fish_crio_no_subcommand --description 'Test if there has been any subcommand yet'
     for i in (commandline -opc)
-        if contains -- $i check complete completion help h config man markdown md status config c containers container cs s info i goroutines g heap hp version wipe help h
+        if contains -- $i check complete completion help h config dedup man markdown md status config c containers container cs s info i goroutines g heap hp version wipe help h
             return 1
         end
     end
@@ -59,6 +59,7 @@ complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-metrics -d 'Enable 
 complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-nri -d 'Enable NRI (Node Resource Interface) support.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-pod-events -d 'If true, CRI-O starts sending the container events to the kubelet'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-profile-unix-socket -d 'Enable pprof profiler on crio unix domain socket.'
+complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-storage-dedup -d 'Enable background storage deduplication using reflinks on startup.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l enable-tracing -d 'Enable OpenTelemetry trace data exporting.'
 complete -c crio -n '__fish_crio_no_subcommand' -f -l gid-mappings -r -d 'Specify the GID mappings to use for the user namespace. This option is deprecated, and will be replaced with Kubernetes user namespace (KEP-127) support in the future.'
 complete -c crio -n '__fish_crio_no_subcommand' -l global-auth-file -r -d 'Path to a file like /var/lib/kubelet/config.json holding credentials necessary for pulling images from secure registries.'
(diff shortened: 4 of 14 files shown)


server/server.go
```diff
@@ -501,6 +501,14 @@ func New(
...
 	s.wipeIfAppropriate(ctx, deletedImages)
 
+	if config.EnableStorageDedup {
+		go func() {
+			if err := RunDedup(ctx, s.ContainerServer.Store()); err != nil {
+				log.Warnf(ctx, "%v", err)
+			}
+		}()
+	}
+
```
giuseppe, line 511: would it make sense to have a goroutine that periodically runs the dedup?  You could configure the interval.

On the other hand, it could just be a cron job that runs `crio dedup` as it doesn't affect running containers nor you need to shutdown any other daemon

harche: I considered a periodic goroutine but decided against it because of lock contention with image pulls.

[`store.Dedup()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L4056) grabs an exclusive write lock on the layer store through [`writeToLayerStore()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L1285) and holds it for the entire pass (layer graph walk, SHA256 hashing every file, reflink ioctl calls). On a large store this could take a while.

While that lock is held, image pulls through [`PutLayer()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L1554) are partially affected. The overlay driver's [staged extraction can proceed without the lock](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L1579-L1607) so the heavy I/O part is fine, but the final layer commit [blocks on `startWriting()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L1615) until dedup releases it. A periodic timer firing during active pulls would cause unpredictable stalls at the commit step.

Also worth noting that [`DedupDirs()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/internal/dedup/dedup.go#L73) doesn't accept a context, so there's no way to cancel a dedup once it starts.

Given all that, the current approach is:
- **Startup dedup** (`enable_storage_dedup = true`): runs before kubelet connects, so no pull contention
- **`crio dedup` CLI**: meant to be used as a cron/systemd timer during maintenance windows, which is basically what you're suggesting too

haircommander: are we able to dedup on image pull? I feel `enable-storage-dedup` won't do very much except for existing images on restart, but at steady state it won't dedup a lot. maybe the flag enables dedup on image pull, which will increase image pull latency. Then we can have the `crio dedup` option if users don't want to incur that cost at runtime

harche: Good point about steady state. I did consider post-pull dedup but moved away from it because of lock contention.

[`store.Dedup()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L4056) grabs an exclusive write lock on the layer store through [`writeToLayerStore()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L1285) and holds it for the entire pass. If dedup gets triggered after a pull and another pull comes in while it's running, the second pull's layer commit [blocks on `startWriting()`](https://github.com/containers/container-libs/blob/b0f86df5a665/storage/store.go#L1615) until dedup finishes. On a busy node (kubelet pulling multiple images after scheduling), that could cause pull timeouts.

The startup case is safe because kubelet hasn't connected yet. For steady state, the idea is that admins run `crio dedup` during maintenance windows (cron/systemd timer) when pull activity is low.

Even with just startup dedup, I saw a 26% reduction in the e2e test (76 MB → 56 MB). For nodes that have been running a while and accumulated images across upgrades, startup dedup after a reboot or upgrade should still give meaningful savings. It's not perfect for steady state, but better than nothing and doesn't risk impacting pulls.

Longer term, if the containers/storage `Dedup()` API gets support for per-layer dedup or context-based cancellation, post-pull dedup could be revisited without the contention risk.

bitoku: I'm a bit concerned about how long it takes to dedup even if it runs only at startup time.
Once we had some issue about long integrity check during cri-o wipe/check in large scale clusters iirc.

PR #9835, opened 2026-03-19
nri: fix a deadlock and a livelock, block plugin sync during other event processing

klihub:
#### What type of PR is this?

/kind bug

Recent long-running conformance tests have uncovered a deadlock in NRI event processing. Fix a deadlock by lock inversion on 2 different code paths (and eliminate an unnecessary lock). Acquire state of container with a timeout to avoid an indefinite livelock if a container (shutdown) is stuck (probably with the container's init process being in an unkillable state). Also, block plugin synchronization during other event processing.

None

None

```release-note
Get rid of an unnecessary lock in NRI integration code. Tolerate better long stuck container processes during container stop. Block plugin registration during event processing.
```

## Summary by CodeRabbit

* **Refactor**
  * Improved NRI plugin synchronization across container and pod sandbox lifecycle (create, start, stop, remove, update) to ensure unblock actions run reliably on all exit paths.
  * Added timeout-tolerant container state retrieval to reduce delays and improve responsiveness during status and update operations.

### DIFF
--- a/server/nri-api.go
+++ b/server/nri-api.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"maps"
 	"slices"
+	"time"
 
 	"github.com/containerd/nri/pkg/api"
 	nrigen "github.com/containerd/nri/pkg/runtime-tools/generate"
@@ -413,8 +414,10 @@ func (a *nriAPI) ListContainers() []nri.Container {
 	}
 
 	for _, ctr := range ctrList {
-		switch ctr.State().Status {
-		case oci.ContainerStateCreated, oci.ContainerStateRunning, oci.ContainerStatePaused:
+		if s := asyncGetState(ctr); s != nil &&
+			(s.Status == oci.ContainerStateCreated ||
+				s.Status == oci.ContainerStateRunning ||
+				s.Status == oci.ContainerStatePaused) {
 			containers = append(containers, &criContainer{
 				api: a,
 				ctr: ctr,
@@ -460,7 +463,17 @@ func (a *nriAPI) UpdateContainer(ctx context.Context, u *api.ContainerUpdate) er
 		return nil
 	}
 
-	if s := ctr.State().Status; s != oci.ContainerStateRunning && s != oci.ContainerStateCreated {
+	s := asyncGetState(ctr)
+	if s == nil {
+		log.Errorf(ctx, "Failed to get container state %q (timeout)",
+			u.GetContainerId())
+
+		return nil
+	}
+
+	if s.Status != oci.ContainerStateCreated &&
+		s.Status != oci.ContainerStateRunning &&
+		s.Status != oci.ContainerStatePaused {
 		return nil
 	}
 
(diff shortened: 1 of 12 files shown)


bitoku:
https://github.com/cri-o/cri-o/pull/9835/changes#r2961220417
Regarding this comment, I know we have a risk of goroutine leak even without this PR when there's a uninterruptible process in the container.

server/nri-api.go
```diff
@@ -698,49 +730,90 @@ func (c *criContainer) GetStatus() *nri.ContainerStatus {
...
-		status.Pid = uint32(cState.InitPid)
+		if cState.InitPid > 0 {
+			status.Pid = uint32(cState.InitPid)
+		}
 	}
 
 	return status
 }
 
+func asyncGetState(ctr *oci.Container) *oci.ContainerState {
```
bitoku, line 778: What do you think about always using StateNoLock for ListContainers and other operations that get affected by locks on other containers?

If a container fails to delete, it shouldn't block operations for any other containers. However, it should stay locked to block further operations on that specific container.

I know we eventually need to fix the actual locking mechanism, but I think this is an acceptable workaround for now.

cc: @ngopalak-redhat @haircommander @saschagrunert

klihub: @bitoku The thing that worries me about always blindly using `oci.Container.StateNoLock()` is the comment there about 'has been known to cause segfaults', but it might be that it is mostly since I lack the context to know under which conditions those segfaults had occurred and can't tell if it would be safe to call it on these code paths.

But if you tell me that it should be safe in these contexts, then I guess I'm, too, willing to trade a lockup for a potentially out of date state instead of a leaked goroutine.

bitoku: Actually that was reported before I joined apparently, so I don't have any information about it like how often it happened and whether it's mitigated now.
I hope @haircommander or @saschagrunert give some ideas.

I'm planning to add a new lock for state, but not sure if we can backport it to older versions since the change will be big.

saschagrunert: I think it would be better to use a dedicated state lock, similar to metaLock from a3c41c49. Until then, the `asyncGetState` timeout approach here seems like a reasonable workaround.

klihub: > I think it would be better to use a dedicated state lock, similar to metaLock from a3c41c49. Until then, the `asyncGetState` timeout approach here seems like a reasonable workaround.

@saschagrunert Yes, but after having talked yesterday to @haircommander about the specifics of the 'segfault-warning comment' for oci.Container.StateNoLock() on the timeout/fallback path in the latest incarnation of asyncGetState(), I think we need to roll back the very last change which, instead of returning a nil state, uses StateNoLock to acquire the state and returns (a copy of) that.

Since the warning is actually related to the risk of panics due to concurrent read/write access to the state, I now think that the fallback to a lockless path is not a good idea, especially with such a short timeout. Now we know that someone is holding the lock (that's why we time out), then we goe ahead and access it anyway. If the lock holder is not a StopLoopForContainer trying to kill the (stuck) container process, but instead some other state updating operation which happens to take some time, then doing what we do now is increasing the probability of triggering an r/w access panic.

What is your opinion on this ?

haircommander: Btw i think @bitoku is putting together a stateLock implementation

haircommander: https://github.com/cri-o/cri-o/pull/9837 he's too fast

klihub: Okay, I'll wait with this for bitoku's fix PR to get in, then rebase and only leave in the updated plugin registration blocking bits.

PR #9836, opened 2026-03-20
Make artifact pinned status respect pinned_images configuration

bitoku:
#### What type of PR is this?

/kind bug

Previously, all OCI artifacts were hardcoded as Pinned: true in CRIImage(). This change makes artifacts respect the same pinned_images configuration used by regular container images, by having the artifact store check artifact names against the compiled pinned image regexps.

```release-note
Respect the same pinned_images configuration used by regular container images
```

## Summary by CodeRabbit

* **New Features**
  * Image pinning via configuration to mark artifacts as pinned (prevents GC); supports exact refs, globs, and digest/name patterns.
  * Pinned state exposed in image/artifact status and propagated to CRI image output; pinning can be updated at runtime without restart.

* **Tests**
  * Added tests validating pin matching, status visibility, canonical/digest matching, and dynamic updates across config reloads.

### DIFF
--- a/internal/ociartifact/artifact.go
+++ b/internal/ociartifact/artifact.go
@@ -28,10 +28,14 @@ type Artifact struct {
 
 	rootPath string
 	namedRef reference.Named
+
+	// pinned indicates whether this artifact should be excluded from
+	// garbage collection, based on the pinned_images configuration.
+	pinned bool
 }
 
-// NewArtifact creates a new Artifact from a libartifact.Artifact.
-func NewArtifact(art *libartifact.Artifact, rootPath string) *Artifact {
+// newArtifact creates a new Artifact from a libartifact.Artifact.
+func (s *Store) newArtifact(art *libartifact.Artifact, rootPath string, pinned bool) *Artifact {
 	artifact := &Artifact{
 		Artifact: art,
 		rootPath: rootPath,
@@ -48,6 +52,8 @@ func NewArtifact(art *libartifact.Artifact, rootPath string) *Artifact {
 
 		artifact.namedRef = namedRef
 	}
+
+	artifact.pinned = pinned || s.isArtifactPinned(artifact)
 
 	return artifact
 }
@@ -83,6 +89,6 @@ func (a *Artifact) CRIImage() *critypes.Image {
 		Size:        uint64(a.TotalSizeBytes()),
 		RepoTags:    repoTags,
 		RepoDigests: []string{a.CanonicalName()},
-		Pinned:      true,
+		Pinned:      a.pinned,
 	}
 }
--- a/internal/ociartifact/store.go
+++ b/internal/ociartifact/store.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"regexp"
 	"slices"
+	"sync/atomic"
 
 	modelSpec "github.com/modelpack/model-spec/specs-go/v1"
 	"github.com/opencontainers/go-digest"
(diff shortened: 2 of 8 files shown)


internal/storage/image.go
```diff
@@ -873,7 +876,9 @@ func pullImageImplementation(ctx context.Context, lookup *imageLookupService, st
 	if shouldTryArtifact(err) {
 		log.Infof(ctx, "Falling back to pull %s as an OCI artifact: %v", imageName, err)
 
-		artifactStore, artifactErr := ociartifact.NewStore(store.GraphRoot(), options.AdditionalArtifactStores, &srcSystemContext)
+		// TODO: pinnedImageRegexps is nil because we can get it here. The result of the pull image is not used for
+		//   image GC. but still should be configured at store level.
```
saschagrunert: This reads like the opposite of what's intended. How about:


```suggestion
		// TODO: pinnedImageRegexps is nil here because the image lookup service
		// does not have access to the compiled regexps. The pull result is not
		// used for image GC, but ideally the store should be configured at a
		// higher level where the regexps are available.
```

internal/ociartifact/store.go
```diff
@@ -229,7 +239,7 @@ func (s *Store) List(ctx context.Context) (res []*Artifact, err error) {
 		}
 
 		for _, art := range addArts {
-			arts = append(arts, NewArtifact(art, add.path))
+			arts = append(arts, s.newArtifact(art, add.path, true))
```
saschagrunert, line 244: Nit: a short comment explaining why additional-store artifacts are force-pinned would help (read-only, not subject to GC). Same for the other `true` call sites in `Pull()` and `Status()`.

internal/ociartifact/artifact.go
```diff
@@ -28,10 +28,14 @@ type Artifact struct {
 
 	rootPath string
 	namedRef reference.Named
+
+	// Pinned indicates whether this artifact should be excluded from
+	// garbage collection, based on the pinned_images configuration.
+	Pinned bool
```
saschagrunert: Should this be unexported with a getter, like `rootPath` and `namedRef`?

server/server.go
```diff
@@ -634,6 +634,7 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 			// ImageServer compiles the list with regex for both
 			// pinned and sandbox/pause images, we need to update them
 			s.ContainerServer.StorageImageServer().UpdatePinnedImagesList(append(s.config.PinnedImages, s.config.PauseImage))
+			s.artifactStore.SetPinnedImageRegexps(s.ContainerServer.StorageImageServer().PinnedImageRegexps())
```
saschagrunert, line 637: Pre-existing issue: `regexForPinnedImages` in `imageService` has no synchronization, unlike the artifact store's `atomic.Value`. Worth a follow-up fix.

internal/storage/image.go
saschagrunert:
```suggestion
		// TODO: pinnedImageRegexps is nil here because the image lookup service
		// does not have access to the compiled regexps. The pull result is not
		// used for image GC, but the store should ideally be configured at a
		// higher level where the regexps are available.
```

internal/ociartifact/store.go
```diff
@@ -362,6 +372,34 @@ func artifactName(annotations map[string]string) string {
...
+	}
+
+	for _, re := range regexps {
+		if re.MatchString(artifact.Reference()) || re.MatchString(artifact.CanonicalName()) {
+			return true
+		}
+	}
+
+	return false
+}
```
saschagrunert, line 404: Consider reusing `storage.FilterPinnedImage` to avoid duplicating the matching logic. Something like:
```go
return storage.FilterPinnedImage(artifact.Reference(), regexps) ||
    storage.FilterPinnedImage(artifact.CanonicalName(), regexps)
```

bitoku: That completely makes sense and I wanted to do, but I couldn't due to cyclic dependency.
https://github.com/cri-o/cri-o/blob/f8be11c7fd7402a04683d04edd16ead064212dec/internal/storage/image.go#L40

I'm thinking changing how storage package calls artifact so that we can reuse the function and also artifacts can get storageID, but for now, I'd like to keep it as is, to narrower the scope of this PR to only artifacts change. Added a TODO comment.

server/server.go
bitoku: yup I'll do it in another follow-up PR.

bitoku:
sorry. In the last time I changed to use atomic.Value to atomic.Pointer which doesn't require type assertion.

PR #9837, opened 2026-03-22
Add stateLock to protect container state from concurrent access

bitoku:
#### What type of PR is this?

/kind bug

Introduce stateLock RWMutex and ModifyState/GetStatus methods to protect concurrent access to container state.Status. ModifyState acquires the write lock and is used exclusively for runtime state transitions (created, running, stopped, paused).

All state modification across runtime_oci, runtime_pod, and runtime_vm are now wrapped with ModifyState. 

🤖 Generated with [Claude Code](https://claude.com/claude-code)

This fixes the NRI livelock when one container is stuck due to interruptible sleep.

```release-note
Fixed the NRI livelock when one container is stuck due to interruptible sleep.
```

## Summary by CodeRabbit

* **New Features**
  * Added safe container state APIs: snapshot reads, status accessors, controlled mutation helpers, and guarded init-PID initialization.

* **Refactor**
  * Centralized state locking and routed container lifecycle state updates through the new locked APIs to improve concurrency safety.

* **Tests**
  * Added tests verifying state snapshot immutability, mutation semantics and error propagation, and init-PID initialization behavior.

### DIFF
--- a/internal/oci/container.go
+++ b/internal/oci/container.go
@@ -60,18 +60,27 @@ type Container struct {
 	conmonCgroupfsPath string
 	crioAnnotations    fields.Set
 	state              *ContainerState
-	opLock             sync.RWMutex // For handling the container OCI runtime state transitions (i.e: anything that needs to call OCI runtime)
-	metaLock           sync.RWMutex // For handling the internal oci.Container metadata. Currently only used for Spec and Resources fields
-	spec               *specs.Spec
-	idMappings         *idtools.IDMappings
-	terminal           bool
-	stdin              bool
-	stdinOnce          bool
-	created            bool
-	spoofed            bool
-	stopping           bool
-	stopDone           bool
-	stopLock           sync.Mutex
+	// stateLock protects the state pointer for memory-safe reads and writes.
+	// Use opLock to prevent data inconsistency.
+	// Do not acquire opLock while stateLock is being acquired.
+	stateLock sync.RWMutex
+	// opLock serializes OCI runtime operations on this container (create, start, stop,
+	// delete, update, pause, unpause, checkpoint, status queries, and stats collection).
+	// It is held across external runtime calls (e.g. runc/crun invocations) to prevent
+	// concurrent runtime operations from racing against each other for the same container.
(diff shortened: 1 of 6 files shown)


bitoku:
@cri-o/cri-o-maintainers PTAL

bitoku:
@cri-o/cri-o-maintainers PTAL

internal/oci/container.go
```diff
@@ -481,6 +514,43 @@ func (c *Container) StateNoLock() *ContainerState {
...
 }
 
+// GetState returns a shallow copy of the container's state, safe for reading outside of locks.
+// Callers must not mutate pointer fields (ExitCode, ContainerMonitorProcess, Annotations).
+func (c *Container) GetState() ContainerState {
+	c.stateLock.RLock()
+	defer c.stateLock.RUnlock()
+
+	return *c.state
+}
```
saschagrunert: The shallow copy shares the pointer fields with the live state. A concurrent ModifyState writing to ExitCode while a caller reads the copy's ExitCode is still a data race.

internal/oci/runtime_oci.go
```diff
@@ -1143,7 +1157,7 @@ func updateContainerStatusFromExitFile(c *Container) error {
 		return fmt.Errorf("failed to find container exit file for %s: %w", c.ID(), err)
 	}
 
-	c.state.Finished, err = getFinishedTime(fi)
+	finished, err := getFinishedTime(fi)
```
saschagrunert, line 1160: Previously c.state.Finished was set here immediately, before exit code parsing. Now Finished is only set inside the ModifyState block at the end. If strconv.ParseInt fails on line 1170, the function returns an error and Finished is never set. The old code would have set it.

bitoku: good catch!

I'm still wondering if we should write `state.Finished` separately, but It's always better to keep the old behavior if we are unsure about it.

internal/oci/container.go
bitoku: Right, I'll check the current GetState() usages if any of these using ExitCode or pointer fields.

PR #9846, opened 2026-03-27
OCPBUGS-81547: oci: wait for exit file before defaulting to exit code 255

### LINKED ISSUE #9840: CRI-O reports exit code 255 for fast-exiting init containers in production (conmon race condition)
### What happened?

CRI-O intermittently reports exit code 255 for containers that exit very quickly (<10ms), even though the container process completed successfully.

CRI-O logs show:
```
level=error msg="Failed to update container state for a0b136050c270...: stdout: , stderr: "
```

This is a re-report of #8980 which was closed without a fix. Unlike #8980 which only affected e2e tests, this issue is a **production blocker** — multiple production environments are experiencing random container failures with no error logs and no user-actionable information.

Containers that exit with code 0 should have exit code 0 reported by CRI-O, regardless of how quickly they exit.

**Minimal reproducer (no Tekton required):**

```yaml
apiVersion: v1
kind: Pod
metadata:
  generateName: fast-exit-test-
spec:
  restartPolicy: Never
  initContainers:
  - name: fast-init
    image: busybox
    command: ["true"]  # exits immediately with code 0
  containers:
  - name: main
    image: busybox
    command: ["echo", "done"]
```

Run many of these concurrently:
```bash
for i in $(seq 1 100); do kubectl create -f fast-exit-pod.yaml & done
wait
kubectl get pods -o json | jq '.items[] | select(.status.initContainerStatuses[]?.state.terminated.exitCode == 255) | .metadata.name'
```

The issue is intermittent. Check CRI-O logs on affected nodes:
```bash
journalctl -u crio --no-pager | grep "Failed to update container state"
```

**Production context:** This was discovered via Tekton Pipelines, where init containers (`prepare`, `place-scripts`) perform fast filesystem operations (binary copy, symlink creation) and exit in <1ms. These init containers run on every single Tekton TaskRun, making the race condition a frequent production issue.

- Related: #8980 (same root cause, closed without fix, only affected e2e tests)
- The `"Failed to update container state"` error in CRI-O logs appears to be the point where the exit code is lost
- Node conditions at time of failure show no memory/disk pressure — this is purely a timing issue
- Multiple independent reports from different environments and cloud providers
- Not environment-specific — affects AWS, bare metal, and other infrastructure
- Downstream tracking: https://github.com/tektoncd/pipeline/issues/9654

<details>

```console
$ crio --version
# CRI-O 1.33.x (shipped with OpenShift 4.20.15)
# Also observed on earlier CRI-O versions
```

```console
# Kubernetes 1.33.x (OpenShift 4.20.15)
```

</details>

### OS version

<details>

```console
# Red Hat Enterprise Linux CoreOS (RHCOS)
# Shipped with OpenShift 4.20.15
```

</details>

<details>

Observed on AWS and other cloud environments. Not environment-specific — the issue is a timing race in container state tracking for fast-exiting processes.

</details>

bitoku:
#### What type of PR is this?

/kind bug

When a container exits very quickly, the OCI runtime may clean
up its state before CRI-O queries it. In this case, runtimeCmd("state")
fails and CRI-O falls back to reading the exit file. However, conmon
may not have written the exit file yet, causing CRI-O to immediately
default to exit code 255 with no retry.

Extract the existing exponential backoff wait for the exit file into a
shared waitForExitFile() function and use it in both code paths:
- Path A: when runtimeCmd("state") fails and the exit file is missing
- Path B: when runtimeCmd("state") reports the container as stopped

This ensures fast-exiting containers get the correct exit code from
conmon's exit file instead of a spurious 255.

Fixes #9840 

```release-note
Fixed the race condition where cri-o reports exitCode 255 when the container exits fast.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability by waiting for container exit markers when runtime state queries fail, and clarifying related logging.
  * Ensures exit time and exit code fallback to 255 when exit status cannot be determined after retries.

* **Tests**
  * Added tests validating successful detection from delayed exit markers and fallback-to-255 when none appear.

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1192,8 +1192,18 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 			// We always populate the fields below so kube can restart/reschedule
 			// containers failing.
 			c.state.Status = ContainerStateStopped
-			if err := updateContainerStatusFromExitFile(c); err != nil {
-				log.Errorf(ctx, "Failed to update container status from exit file for %s: %v", c.ID(), err)
+
+			// The exit file may not exist yet if conmon hasn't written it.
+			// Wait for it to appear before defaulting to exit code 255.
+			// This prevents a race where fast-exiting containers
+			// get a spurious exit code 255.
+			waitErr := waitForExitFile(c)
+			if waitErr != nil {
+				log.Errorf(ctx, "Failed to update container status from exit file for %s: %v", c.ID(), waitErr)
+				c.state.Finished = time.Now()
+				c.state.ExitCode = new(int32(255))
+			} else if err := updateContainerStatusFromExitFile(c); err != nil {
+				log.Errorf(ctx, "Failed to read exit file for %s after wait: %v", c.ID(), err)
 				c.state.Finished = time.Now()
 				c.state.ExitCode = new(int32(255))
 			}
@@ -1224,26 +1234,11 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 
 		return nil
 	}
-	// release the lock before waiting
+
 	c.opLock.Unlock()
-	exitFilePath := c.exitFilePath()
-	err = kwait.ExponentialBackoff(
-		kwait.Backoff{
-			Duration: 500 * time.Millisecond,
-			Factor:   1.2,
-			Steps:    6,
-		},
-		func() (bool, error) {
-			_, err := os.Stat(exitFilePath)
-			if err != nil {
-				// wait longer
-				return false, nil
-			}
-
-			return true, nil
-		})
-
+	err = waitForExitFile(c)
 	c.opLock.Lock()
+
 	// run command again
 	state, _, err2 := stateCmd()
 	if err2 != nil {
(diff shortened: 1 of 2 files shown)


internal/oci/runtime_oci.go
```diff
@@ -1163,6 +1163,28 @@ func updateContainerStatusFromExitFile(c *Container) error {
 	return nil
 }
 
+// waitForExitFile waits for the container's exit file to appear using
+// exponential backoff. Returns nil if the file appeared, or an error if it did
+// not appear within the timeout (~5s total: 500ms initial, 1.2 factor, 6 steps).
+func waitForExitFile(c *Container) error {
```
haircommander: nit: define below called

internal/oci/runtime_oci.go
```diff
@@ -1192,10 +1214,20 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
...
 			c.state.Status = ContainerStateStopped
-			if err := updateContainerStatusFromExitFile(c); err != nil {
+
+			// The exit file may not exist yet if conmon hasn't written it.
+			// Wait for it to appear before defaulting to exit code 255.
+			// This prevents a race where fast-exiting containers
+			// get a spurious exit code 255.
+			waitErr := waitForExitFile(c)
+			if waitErr != nil {
 				log.Errorf(ctx, "Failed to update container status from exit file for %s: %v", c.ID(), err)
```
haircommander: did you mean to use waitErr here?

internal/oci/runtime_oci_test.go
```diff
@@ -232,6 +235,118 @@ var _ = t.Describe("Oci", func() {
...
+
+		defer cmdrunner.ResetPrependedCmd()
+
+		runner.EXPECT().Command(gomock.Any(), gomock.Any()).Return(
+			exec.Command("/bin/false"),
+		).AnyTimes()
+
+		// Write the exit file after a short delay, simulating conmon
+		// writing it after the first read attempt fails.
+		go func() {
```
haircommander: why not spawn this closer to UpdateContainerStatus? we probably won't go through 800 ms before it's started but if we did that could be a false positive. unlikely, but just in case?

internal/oci/runtime_oci.go
bitoku: good catch!

bitoku: good catch!

internal/oci/runtime_oci_test.go
bitoku: that sounds better. tbh I didn't think about that case.

bitoku:
@cri-o/cri-o-maintainers PTAL

PR #9855, opened 2026-03-28
cnimgr: continuously poll CNI STATUS to detect runtime health changes

tsorya:
#### What type of PR is this?

/kind feature

CRI-O polls CNI Status() at startup to wait for readiness, but stops
after the first success. With the CNI STATUS verb (spec v1.1.0), plugins
can now report unhealthy at runtime, but CRI-O doesn't react to this
verb after initial readiness. If a plugin becomes unhealthy (e.g. daemon
restart, config regeneration, or network failure), CRI-O keeps reporting
NetworkReady=true and pods fail at network setup instead of the node
being marked not-ready.

This is analogous to containerd, which calls Status() on every CRI
Status request (see containerd/containerd#11135).

This PR changes the polling to two stages:
- **Init poll** (500ms): fast poll until first ready, triggers GC and
  notifies watchers. Same as before.
- **Monitor poll** (5s): continuous health check. Sets not-ready on
  failure, self-heals on recovery.

Also replaces the shutdown bool with context cancellation for clean
goroutine lifecycle, and handles edge cases in AddWatcher for shutdown
and already-ready states.

Fixes https://redhat.atlassian.net/browse/OCPBUGS-81474

The `doGC` call is still made under the mutex (inherited from the
original code). This could be improved in a follow-up by splitting the
lock around the I/O-heavy GC operation.

```release-note
CRI-O now continuously monitors CNI plugin health using the STATUS
verb. If a plugin becomes unhealthy after initial readiness, the node
is reported as NetworkReady=false, preventing pod scheduling on
affected nodes. The node self-heals when the plugin recovers.
```

### DIFF
--- a/internal/config/cnimgr/cnimgr.go
+++ b/internal/config/cnimgr/cnimgr.go
@@ -14,13 +14,17 @@ import (
 
 type PodNetworkLister func() ([]*ocicni.PodNetwork, error)
 
+var errShutdown = errors.New("CNI manager shut down")
+
 type CNIManager struct {
 	// cniPlugin is the internal OCI CNI plugin
-	plugin    ocicni.CNIPlugin
-	lastError error
-	watchers  []chan bool
-	shutdown  bool
-	mutex     sync.RWMutex
+	plugin              ocicni.CNIPlugin
+	lastError           error
+	watchers            []chan bool
+	mutex               sync.RWMutex
+	cancel              context.CancelFunc
+	initPollInterval    time.Duration
+	monitorPollInterval time.Duration
 
 	validPodList PodNetworkLister
 }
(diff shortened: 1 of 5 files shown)


internal/config/cnimgr/cnimgr.go
```diff
@@ -34,48 +38,69 @@ func New(defaultNetwork, networkDir string, pluginDirs ...string) (*CNIManager,
...
-	// valid and stable because new pods can't be created until the plugin is
-	// announced as ready
-	if err := c.doGC(context.Background()); err != nil {
-		logrus.Warnf("Garbage collect stale network resources during plugin startup failed: %v", err)
-	}
 
-	c.lastError = nil
-	for _, watcher := range c.watchers {
-		watcher <- true
+	if c.lastError != nil {
```
haircommander, line 86: hm I'm not sure we still want to do the GC if the plugin went down and came back up, but rather just once on startup. I think this should only happen on stopOnReady (or, maybe better, we should pull this section out and add it after we've done phase 1 but before we start phase 2, assuming they're done synchronously?)

haircommander: oh wait no we need to do gc before watchers, ignore the second option

tsorya: sure, i ll change a name for stopOnReady, will not match in this case

tsorya: Done

haircommander:
/approve

bitoku:
@tsorya can you update the PR description to include release note? (see PR template)

internal/config/cnimgr/cnimgr.go
bitoku: Can you update the PR description as well for posterity?

tsorya:
@bitoku updated PR description, does it look ok?

tsorya:
go: go.mod requires go >= 1.26.2 (running go 1.26.0; GOTOOLCHAIN=local)
make: *** [Makefile:101: /tmp/tmp.N7UFu2Rj1s/build/bin/linux/amd64/critest] Error 1

seems not relevant to the change

bitoku:
@tsorya
linter failed. `make lint` will fix them.

PR #9860, opened 2026-03-31
Add min_injected_gomaxprocs via runtime handler precreate hook

harche:
## Summary

Implements GOMAXPROCS injection as a composable `PreCreate` runtime handler hook.

- `GomaxprocsHooks.PreCreate()` reads CPU shares from the OCI spec, calculates GOMAXPROCS with a configurable floor (`min_injected_gomaxprocs`), and injects the env var for burstable/best-effort pods
- `CompositeHooks` chains multiple `RuntimeHandlerHooks` (e.g. GOMAXPROCS + high-performance hooks)
- Works for all OCI runtimes
- Skips guaranteed pods, workload-partitioned pods, pods with CPU limits, and pods that already set GOMAXPROCS
- `skip-gomaxprocs.crio.io` pod annotation to opt out

This is an alternative to #9844 based on reviewer feedback to use the precreate hook mechanism.

```release-note
add min_injected_gomaxprocs option, which allows a user to specify GOMAXPROCS in every container CRI-O creates. The config field itself is an integer that represents the floor of GOMAXPROCS. CRI-O will inject max(floor, cpu.request), if the pod is not a guaranteed pod or is part of a partitioned workload
```


internal/runtimehandlerhooks/runtime_handler_hooks_linux.go
```diff
@@ -64,16 +64,29 @@ func (hr *HooksRetriever) Get(ctx context.Context, runtimeName string, sandboxAn
...
+		hooks = append(hooks, &DefaultCPULoadBalanceHooks{
+			CgroupManager: hr.config.CgroupManager(),
+		})
 	}
 
-	if cpuLoadBalancingAllowed(hr.config) {
-		return &DefaultCPULoadBalanceHooks{
-			CgroupManager: hr.config.CgroupManager(),
-		}
+	if hr.config.InjectGOMAXPROCS > 0 {
```
haircommander: can we drop the config field and instead introduce a new config field (in crio.runtime table) that's `default_allowed_annotations` injects allowed annotations into each runtime handler? then, we can use the presence of the annotation in the runtime handler to toggle this behavior

internal/runtimehandlerhooks/runtime_handler_hooks_linux.go
```diff
@@ -64,16 +64,29 @@ func (hr *HooksRetriever) Get(ctx context.Context, runtimeName string, sandboxAn
...
 	}
 
-	return nil
+	switch len(hooks) {
+	case 0:
+		return nil
+	case 1:
+		return hooks[0]
+	default:
+		return &CompositeHooks{hooks: hooks}
```
haircommander, line 87: I think we can do without composite hooks if we just call the gomaxprocs hook in https://github.com/cri-o/cri-o/blob/main/internal/runtimehandlerhooks/high_performance_hooks_linux.go#L134 (assuming it's enabled in the runtime)

internal/runtimehandlerhooks/runtime_handler_hooks_linux.go
harche: The current `inject_gomaxprocs = 4` carries both the enable toggle and the floor value. With the annotation approach, `default_allowed_annotations` handles the toggle, but where would the floor value (the `4`) come from? Would we use `default_annotations = {"gomaxprocs.crio.io": "4"}` on the runtime handler for that?

harche: Spoke with @haircommander offline, we will keep the existing arg.

pkg/annotations/v2/annotations.go
```diff
@@ -279,6 +283,7 @@ var AllAnnotations = []string{
 	PodLinuxResources,
 	SeccompNotifierAction,
 	SeccompProfile,
+	SkipGoMaxProcs,
```
haircommander: this shouldn't be an allowed annotation, remove this line

harche: @haircommander If remove it then pods won't be able skip setting `GOMAXPROCS` by setting annotation? We want pods to be able to skip.

internal/runtimehandlerhooks/runtime_handler_hooks_linux.go
harche: but it is called only from high performance hook, https://github.com/search?q=repo%3Acri-o/cri-o%20shouldRunHooks&type=code

pkg/annotations/v2/annotations.go
haircommander: that's not true, inclusion in this list means that the annotation will be scrubbed from the container if it's not in a runtime handler that has the allowed annotation. pods are allowed to have annotations that aren't in this list and cri-o can take any action on any annotation it wants. the reason to include here is if we feel there's a risk to the annotation and we want an admin to explicitly opt into its enablement

internal/runtimehandlerhooks/runtime_handler_hooks_linux.go
haircommander: hooks retriever should run for every runtime handler, regardless of whether it's high performance

haircommander: though I take this comment back I think composite hooks is a better design

pkg/annotations/v2/annotations.go
harche: Ok, thanks, looking into it.

harche: Just to be sure, I will test this again on AWS.

internal/runtimehandlerhooks/gomaxprocs_hooks_linux.go
```diff
@@ -0,0 +1,135 @@
...
+func (*GomaxprocsHooks) PostStop(context.Context, *oci.Container, *sandbox.Sandbox) error {
+	return nil
+}
+
+// calculateGOMAXPROCS derives the GOMAXPROCS value from CPU shares and a floor.
+// Kubelet sets shares = cpu_request_in_millicores * 1024 / 1000.
+// We reverse that with ceil(shares / 1024) to get the CPU count.
+// The floor is used when the calculated value is lower.
+func calculateGOMAXPROCS(shares, fallbackMaxProcs int64) int64 {
+	if shares > 2 {
```
haircommander: why are we checking this? seems like we could just return

`max( max((shares + 1023)/1024, 1), fallbackMaxProcs)`

internal/runtimehandlerhooks/gomaxprocs_hooks_linux.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	return fallbackMaxProcs
+}
+
+// injectGOMAXPROCS sets the GOMAXPROCS environment variable to the given value.
+// Injection is skipped if GOMAXPROCS is already set in defaultEnv or the OCI spec's process env.
+func injectGOMAXPROCS(specgen *generate.Generator, defaultEnv []string, maxProcs int64) {
+	for _, env := range defaultEnv {
```
haircommander: can you check server/container_create.go to see if we've already injected defaultEnv into the Env struct? this loop may be redundant

haircommander:
/retest
still WIP @harche ?

harche:
> /retest still WIP @harche ?

Thanks, removed it.

haircommander:
/hold
sorry to be annoying ,actually can the config field be `min_injected_gomaxprocs`? better reflects it's an integer and not a bool

PR #9875, opened 2026-04-09
seccomp: keep logging blocked syscalls in notifier log mode

### LINKED ISSUE #9872: SecComp Notifier function stops loging after first block syscall detected
### Description:
#5720 introduced #6120 PR which implements the seccomp notifier feature with a specific condition that the container will be stopped once a blocked syscall is being made.

It introduces a deadlock scenario where if one application makes uses of glibc as per
https://github.com/moby/moby/pull/42681 and would like to use SecComp, the application unintentionally uses clone3 syscall, the "seccompNotifierAction: stop" annotation will stop the container almost immediately.  The application will in crash loop as the application always try to use clone3 in the startup sequence. 

The workaround for such an application is to use "seccompNotifierAction: log" to preserve the seccomp notifier observability. 
However, [it](https://github.com/cri-o/cri-o/blob/2ebd656141658cee4fab9166586fb3d626088558/internal/config/seccomp/notifier.go#L271) has a side effect that when the notifer annotation is set to "seccompNotifierAction: log", only the first offending syscall is logged, any subsequent offending syscall will be blocked but not logged. 

### CRI-O Configuration
CRI-O drop-in config:

```
[crio.runtime]
default_runtime = "runc"

[crio.runtime.runtimes.runc]
runtime_path = "/usr/bin/runc"
allowed_annotations = ["io.kubernetes.cri-o.seccompNotifierAction"]
```
**Execution flow for our test:**
| Step | What happens | Code location |
|------|-------------|---------------|
| 1 | `handler()` goroutine starts, blocks on `NotifReceive(fd)` | Line [233 ](https://github.com/cri-o/cri-o/blob/2ebd656141658cee4fab9166586fb3d626088558/internal/config/seccomp/notifier.go#L233)|
| 2 | Container calls `swapon()` → kernel pauses thread, sends notification | Kernel seccomp |
| 3 | `NotifReceive` returns with `req.Data.Syscall = swapon` | Line 233 |
| 4 | Syscall name sent to `msgChan` → CRI-O logs it, updates metric | Line 250 |
| 5 | Handler responds with `ENOSYS` → kernel unpauses thread, app gets errno=38 | Line 252-265 |
| 6 | **`break`** → exits the `for` loop | **[Line 271](https://github.com/cri-o/cri-o/blob/2ebd656141658cee4fab9166586fb3d626088558/internal/config/seccomp/notifier.go#L271)** |
| 7 | `defer unix.Close(int(fd))` → seccomp notification fd is closed | Line 231 |
| 8 | Container calls `kexec_load()` → kernel has no listener → returns ENOSYS | Kernel default |
| 9 | Same for userfaultfd, bpf, perf_event_open — all get ENOSYS, none logged | — |

### Question
Is this a design behavior that only the first syscall will be logged and all subsequent syscall will be dropped silently despite using "seccompNotifierAction: log" annotation ?

officialasishkumar:
#### What type of PR is this?
/kind bug

Keeps the seccomp notifier handler polling its notify file descriptor when the pod uses notifier log mode. That preserves logging for later blocked syscalls instead of dropping everything after the first event. The change also adds unit coverage for the stop-mode and log-mode handler behavior.

Fixes #9872

None

```release-note
CRI-O now continues logging blocked syscalls when seccomp notifier log mode is enabled instead of only reporting the first event.
```

## Summary by CodeRabbit

* **Refactor**
  * Improved seccomp notification handling to support both single-notification (stop) and continuous-polling (log) modes for more predictable runtime behavior.

* **Bug Fixes**
  * Handler now treats specific terminal errors as immediate exits and avoids unnecessary retries or logging.

* **Tests**
  * Added tests validating notification sequencing and correct stop, continuous, and terminal-error behaviors.


bitoku (commented):
/ok-to-test

robot0728:
Thanks for the fix.

One concern before merging: goroutine leak in log mode on every container exit.

In log mode (stopAfterFirstNotification=false), the handler goroutine stays alive for the container's full lifetime. When the container exits, the kernel closes the seccomp fd, causing notifReceive(fd) to return an error (EBADF or ECANCELED). The current error path:
```
if err != nil {
    log.Errorf(ctx, "Unable to receive notification: %v", err)
    continue  // ← loops back to notifReceive on a closed fd
}
```
..puts the goroutine into a tight infinite loop — spamming error logs and never returning. The defer unix.Close(int(fd)) never runs, leaking the fd. The ctx passed to handler() is never cancelled on container exit, so there is no other exit path.

This affects every log mode container exit — normal, OOM, kubectl delete, all of them. It does not affect stop mode because break fires before the container is killed.

Suggested fix:
```
if err != nil {
    if errors.Is(err, unix.EBADF) || errors.Is(err, unix.ECANCELED) {
        log.Infof(ctx, "Seccomp notifier fd closed for container %s, exiting handler", containerID)
        return
    }
    log.Errorf(ctx, "Unable to receive notification: %v", err)
    continue
}
```

adilGhaffarDev:
> Thanks for the fix.
> 
> One concern before merging: goroutine leak in log mode on every container exit.
> 
> In log mode (stopAfterFirstNotification=false), the handler goroutine stays alive for the container's full lifetime. When the container exits, the kernel closes the seccomp fd, causing notifReceive(fd) to return an error (EBADF or ECANCELED). The current error path:
> 
> ```
> if err != nil {
>     log.Errorf(ctx, "Unable to receive notification: %v", err)
>     continue  // ← loops back to notifReceive on a closed fd
> }
> ```
> 
> ..puts the goroutine into a tight infinite loop — spamming error logs and never returning. The defer unix.Close(int(fd)) never runs, leaking the fd. The ctx passed to handler() is never cancelled on container exit, so there is no other exit path.
> 
> This affects every log mode container exit — normal, OOM, kubectl delete, all of them. It does not affect stop mode because break fires before the container is killed.
> 
> Suggested fix:
> 
> ```
> if err != nil {
>     if errors.Is(err, unix.EBADF) || errors.Is(err, unix.ECANCELED) {
>         log.Infof(ctx, "Seccomp notifier fd closed for container %s, exiting handler", containerID)
>         return
>     }
>     log.Errorf(ctx, "Unable to receive notification: %v", err)
>     continue
> }
> ```

@officialasishkumar can you address these comments?

officialasishkumar:
Addressed in 74fb0e1: log-mode handlers now return when the seccomp fd is closed, instead of looping on receive errors.

ZiyueChen0728:
Since this PR keeps the handler loop running in log mode, the journal could now have many notifier entries per container. Worth improving the log message while you're here?

The log is in server/server_linux.go:

go
`log.Infof(ctx, "Got seccomp notifier message for container ID: %s (syscall = %s)", id, syscall)`


s.GetContainer(ctx, id) is already called a few lines below for the Prometheus metric. Moving it up gives access to ctr.Labels() and oci.LabelsToDescription():

go
```
ctr := s.GetContainer(ctx, id)
if ctr != nil {
    log.Warnf(ctx, "Seccomp blocked syscall '%s' in container %s (%s)",
        syscall, id, oci.LabelsToDescription(ctr.Labels()))
} else {
    log.Warnf(ctx, "Seccomp blocked syscall '%s' in container %s",
        syscall, id)
}

```

Note: GetContainer can return nil — the exit monitor in server.go checks for this, but the existing seccomp consumer doesn't. The nil guard above avoids that.

The word "blocked" is safe — injectNotifier() only rewrites SCMP_ACT_ERRNO and SCMP_ACT_KILL* to SCMP_ACT_NOTIFY. SCMP_ACT_ALLOW is never rewritten, and defaultAction can't be (kernel limitation). So every notification the handler receives is guaranteed to be a blocked syscall.

For PID, the Notification struct would need a new field since req.Pid is available in handler() but not passed through msgChan today. That could be a follow-up.

Journal entry comparison:

Before:
`Apr 23 09:29:41 worker-node-03 crio[8421]: time="2026-04-23T07:29:41.832Z" level=info msg="Got seccomp notifier message for container ID: a4f29e71b083 (syscall = swapon)" id=e7a31f02-5c8b-4d19-a6f3-92b1c4d8e5f7 name=/runtime.v1.RuntimeService/CreateContainer
`

After:
`Apr 23 09:29:41 worker-node-03 crio[8421]: time="2026-04-23T07:29:41.832Z" level=warning msg="Seccomp blocked syscall 'swapon' in container a4f29e71b083 (my-namespace/seccomp-test-stop/test)" id=e7a31f02-5c8b-4d19-a6f3-92b1c4d8e5f7 name=/runtime.v1.RuntimeService/CreateContainer`

internal/config/seccomp/notifier.go
```diff
@@ -227,11 +235,16 @@ func handler(
...
-		req, err := libseccomp.NotifReceive(fd)
+		req, err := notifReceive(fd)
 		if err != nil {
+			if errors.Is(err, unix.EBADF) || errors.Is(err, unix.ECANCELED) {
+				log.Infof(ctx, "Stopping notifier for container %s", containerID)
+				return
+			}
 			log.Errorf(ctx, "Unable to receive notification: %v", err)
 			continue
 		}
```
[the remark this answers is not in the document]
officialasishkumar: Handled in 404829983. ENOENT now stops the notifier handler, with coverage in the existing terminal-error test table.

officialasishkumar:
Addressed in 5cf87edf6e0b97552b8136f799ced1fa3f68c1e4. The seccomp notifier log now includes blocked syscall context and the container description when available.

adilGhaffarDev:
@bitoku can you take a look all.

internal/config/seccomp/notifier_test.go
bitoku, line 1: We prefer ginkgo style tests though we still have go standard unit tests.
Can you use ginkgo instead of standard unit tests?

If you want to inject or reference private values, you can create some functions in files only built during test.
Look for "test" tag and "_inject" files for examples.

internal/config/seccomp/notifier.go
```diff
@@ -257,18 +270,22 @@ func handler(
...
 
-		// We only catch the first syscall
-		break
+		if stopAfterFirstNotification {
+			// Stop-mode only cares about the first blocked syscall because the
+			// container is terminated immediately afterwards. Log-mode must keep
+			// polling so later blocked syscalls are still reported.
+			break
+		}
```
bitoku: I wonder if we need this. Instead can we just keep this running?
Because when the container is stopped, seccompFd is closed right? We can safely return this function at `notifReceive`, I guess.

server/server_linux.go
```diff
@@ -89,8 +90,14 @@ func (s *Server) startSeccompNotifierWatcher(ctx context.Context) error {
...
 			syscall := msg.Syscall()
+			ctr := s.GetContainer(ctx, id)
 
-			log.Infof(ctx, "Got seccomp notifier message for container ID: %s (syscall = %s)", id, syscall)
+			if ctr != nil {
+				log.Warnf(ctx, "Seccomp blocked syscall '%s' in container %s (%s)",
+					syscall, id, oci.LabelsToDescription(ctr.Labels()))
+			} else {
+				log.Warnf(ctx, "Seccomp blocked syscall '%s' in container %s", syscall, id)
+			}
```
bitoku, line 101: If ctr == nil, it shouldn't continue. Otherwise the subsequent usage will fail.

Also I'd like to keep `Info` rather than `Warn` because it's the problem of the workload (it's ok when ctr == nil). It's just a preference and non-blocking though.

bitoku (commented):
Thank you for the fix!

We'd like integration test to verify this. Can you add it?

officialasishkumar:
Updated in `40959bc`: converted the notifier tests to Ginkgo style, kept the notifier handler polling until the seccomp fd close/error path, fixed the `ctr == nil` path to log at `Info` and continue before later `ctr` usage, and updated the BATS integration expectations/coverage.

Validation:
- `git diff --check`: passed
- `go test -tags "test containers_image_ostree_stub exclude_graphdriver_btrfs btrfs_noversion" ./internal/config/seccomp`: passed

I also tried the seccomp-tag package test and broader server compile, but this runner is missing `gpgme.pc` and `libseccomp.pc`; the broader compile also hit local disk pressure. I updated the integration BATS test but could not run it in this environment.

server/server_linux.go
```diff
@@ -89,8 +90,16 @@ func (s *Server) startSeccompNotifierWatcher(ctx context.Context) error {
 			ctx := msg.Ctx()
 			id := msg.ContainerID()
 			syscall := msg.Syscall()
+			ctr := s.GetContainer(ctx, id)
+
+			if ctr == nil {
+				log.Infof(ctx, "Seccomp blocked syscall '%s' in container %s", syscall, id)
```
bitoku:
```suggestion
				log.Warnf(ctx, "Seccomp blocked syscall '%s' in container %s", syscall, id)
				// It can't get the container name, but still should emit the metric.
				metrics.Instance().MetricContainersSeccompNotifierCountTotalInc(id, syscall)
```

internal/config/seccomp/notifier_test.go
bitoku, line 1: `notifier_test.go` rather than `notifier_internal_test.go` may be better following the convention.

internal/config/seccomp/notifier.go
```diff
@@ -170,6 +170,11 @@ func NewNotifier(
 		return nil, fmt.Errorf("listen for seccomp socket: %w", err)
 	}
 
+	action, ok := v2.GetAnnotationValue(annotationMap, v2.SeccompNotifierAction)
+	if !ok {
+		return nil, fmt.Errorf("%s annotation not set on container", v2.SeccompNotifierAction)
+	}
```
bitoku, line 180: It's preexisting, but we should close the listener here

adilGhaffarDev:
@officialasishkumar can you address the comments? if you dont have time, I can take this.
This is something that we need in internally and we are blocked because of this PR.

server/server_linux.go
officialasishkumar: Done in 75209ffb6: the nil-container path now emits the seccomp notifier metric using the container ID before continuing.

internal/config/seccomp/notifier_test.go
officialasishkumar: Done in 75209ffb6: renamed the test file to internal/config/seccomp/notifier_test.go.

internal/config/seccomp/notifier.go
officialasishkumar: Done in 75209ffb6: NewNotifier now closes the listener before returning when the notifier action annotation is missing.

adilGhaffarDev:
we can close this pr, this pr https://github.com/cri-o/cri-o/pull/10120 supersedes this one.

PR #9887, opened 2026-04-14
[1.33] Add inject_gomaxprocs via runtime handler precreate hook

haircommander:
This is an semi automated cherry-pick of #9860

assisted by claude

```release-note
add min_injected_gomaxprocs option, which allows a user to specify GOMAXPROCS in every container CRI-O creates. The config field itself is an integer that represents the floor of GOMAXPROCS. CRI-O will inject max(floor, cpu.request), if the pod is not a guaranteed pod or is part of a partitioned workload
```


harche (commented):
/lgtm

PR #9898, opened 2026-04-17
oci: clone execPIDs map to avoid segfault

haircommander:
#### What type of PR is this?

/kind bug

```release-note
none
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread-safety in container operation handling by fixing a concurrency issue that could potentially cause race conditions during parallel execution operations.


bitoku (commented):
good catch!
/lgtm

PR #9903, opened 2026-04-23
OCPBUGS-81474: [1.35] cnimgr: continuously poll CNI STATUS to detect runtime health changes

haircommander:
#### What type of PR is this?

/kind feature

CRI-O polls CNI Status() at startup to wait for readiness, but stops
after the first success. With the CNI STATUS verb (spec v1.1.0), plugins
can now report unhealthy at runtime, but CRI-O doesn't react to this
verb after initial readiness. If a plugin becomes unhealthy (e.g. daemon
restart, config regeneration, or network failure), CRI-O keeps reporting
NetworkReady=true and pods fail at network setup instead of the node
being marked not-ready.

This is analogous to containerd, which calls Status() on every CRI
Status request (see containerd/containerd#11135).

This PR changes the polling to two stages:
- **Init poll** (500ms): fast poll until first ready, triggers GC and
  notifies watchers. Same as before.
- **Monitor poll** (5s): continuous health check. Sets not-ready on
  failure, self-heals on recovery.

Also replaces the shutdown bool with context cancellation for clean
goroutine lifecycle, and handles edge cases in AddWatcher for shutdown
and already-ready states.

Fixes https://redhat.atlassian.net/browse/OCPBUGS-81474

The `doGC` call is still made under the mutex (inherited from the
original code). This could be improved in a follow-up by splitting the
lock around the I/O-heavy GC operation.

cherry-pick of https://github.com/cri-o/cri-o/pull/9855 

```release-note
CRI-O now continuously monitors CNI plugin health using the STATUS
verb. If a plugin becomes unhealthy after initial readiness, the node
is reported as NetworkReady=false, preventing pod scheduling on
affected nodes. The node self-heals when the plugin recovers.
```

### DIFF
--- a/internal/config/cnimgr/cnimgr.go
+++ b/internal/config/cnimgr/cnimgr.go
@@ -14,13 +14,17 @@ import (
 
 type PodNetworkLister func() ([]*ocicni.PodNetwork, error)
 
+var errShutdown = errors.New("CNI manager shut down")
+
 type CNIManager struct {
 	// cniPlugin is the internal OCI CNI plugin
-	plugin    ocicni.CNIPlugin
-	lastError error
-	watchers  []chan bool
-	shutdown  bool
-	mutex     sync.RWMutex
+	plugin              ocicni.CNIPlugin
+	lastError           error
+	watchers            []chan bool
+	mutex               sync.RWMutex
+	cancel              context.CancelFunc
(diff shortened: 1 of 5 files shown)


haircommander:
cherry-pick assisted by claude

PR #9904, opened 2026-04-23
oci: trim leading CR/LF from exec args before OCI process spec

### LINKED ISSUE #9885: exec fails when shell argument starts with a newline on s390x
### What happened?

`kubectl exec` into a container fails when the shell script argument to `/bin/sh -euc` starts with a leading newline character on s390x. The exec'd command does not run. Instead, the output contains what appears to be CRI-O/conmon user context metadata `({"user":{"uid":...,"gid":...}})` leaked into the stream, which the container's main process then fails to parse.

The same invocation works on x86_64 with the same CRI-O version line.
```bash
# works
kubectl exec <pod> -n <ns> -- /bin/sh -euc 'echo hello'
# output: hello

# fails — only difference is a newline after the opening quote
kubectl exec <pod> -n <ns> -- /bin/sh -euc '
echo hello'
# output:
# cannot parse the data: `lexical error: invalid character inside string.
#                                         {"user":{"uid":1001,"gid":0,"a
#                      (right here) ------^
# `
# command terminated with exit code 1
```
Tested on multiple pods with different images and UIDs — the leaked UID always matches the container's running user. Multi-line scripts work fine as long as the argument does not start with \n.

Both commands should print `hello`. 
A leading newline in a shell script is valid syntax and should be ignored by the shell. The exec should behave identically regardless of whether the script argument begins with a newline.

On an s390x cluster running CRI-O 1.35.0:

```bash
# pick any running pod
POD=$(kubectl get pods -A -o jsonpath='{.items[0].metadata.name}')
NS=$(kubectl get pods -A -o jsonpath='{.items[0].metadata.namespace}')
# this works
kubectl exec "$POD" -n "$NS" -- /bin/sh -euc 'echo hello'
# this fails
kubectl exec "$POD" -n "$NS" -- /bin/sh -euc '
echo hello'
```

- Confirmed on every pod tested on the cluster (different namespaces, images, UIDs)
- The {"user":{"uid":...,"gid":...}} in the error output matches the container's id output exactly
- x86_64 clusters with the same Kubernetes version are not affected
- Workaround: ensure the shell script argument does not start with a newline

<details>

```console
$ crio --version
crio version 1.35.0-2.rhaos4.22.git7deb12c.el9
  GoVersion:      go1.25.7 (Red Hat 1.25.7-1.el9)
```

```console
# Client: v1.34.2, Server: v1.34.2
# Platform: linux/s390x
```

</details>

### OS version

<details>

```console
$ cat /etc/os-release
PRETTY_NAME="Red Hat Enterprise Linux CoreOS 9.8.20260221-0 (Plow)"
OPENSHIFT_VERSION="4.22"
VARIANT=CoreOS
VARIANT_ID=coreos

$ uname -a
Linux ....com 5.14.0-686.el9.s390x #1 SMP s390x GNU/Linux
```

</details>

<details>
Bare-metal s390x , OpenShift 4.22.0-ec.3, 3 master + 3 worker nodes.
</details>

isumitsolanki:
#### What type of PR is this?

/kind bug

`kubectl exec` can pass a shell `-c` script whose first character is a newline (e.g. `'/bin/sh' '-euc' '\necho hello'`). That is valid for the shell, but could break the exec I/O path and surface garbage such as OCI process `user` JSON in the stream (see issue). This change normalizes CRI exec command arguments by stripping leading `\r` and `\n` from each string before building the OCI process spec for `runc`/`crun` exec and for the VM runtime path. Semantics match the shell: leading blank lines in a `-c` script are ignored. Adds `NormalizeExecCmdArgs` in `internal/oci/exec_args.go` and unit tests.

Fixes #9885

None

```release-note
NONE```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of exec command arguments so leading newline and carriage return characters are removed before launch.
  * Runtime configuration updates now better validate Pod CIDR values, support multiple CIDR ranges, and clear stored ranges when the setting is emptied.

* **Tests**
  * Added coverage for command argument normalization and runtime configuration update behavior.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


bitoku:
Waiting for the answer https://github.com/cri-o/cri-o/issues/9885#issuecomment-4379166510

isumitsolanki:
Hi @bitoku @vikash2310 , Can you have look here ?

bitoku:
This should be fixed by https://github.com/cri-o/cri-o/pull/10073

isumitsolanki:
Dependent PR  #10073  got merged, hence closing this PR

PR #9918, opened 2026-05-01
TRT-2646: gate continuous CNI STATUS monitoring behind config flag and provide a tolerance window

tsorya:
/kind bug
/kind feature

    **cnimgr: gate continuous CNI STATUS monitoring behind config flag**

    Add enable_cni_status_monitoring configuration option (default false)
    to control whether CRI-O continuously polls the CNI STATUS verb after
    initial startup readiness.

    When disabled (default), behavior is identical to pre-monitoring:
    Phase 1 polls until the plugin is ready, then stops. The node stays
    NetworkReady=true after startup with no further health checks.

    When enabled, Phase 2 background polling detects plugin health changes
    at runtime and can mark the node not-ready (subject to
    cni_status_grace_period).

    This allows safe rollout of continuous monitoring without changing
    default behavior.

    Signed-off-by: Igal Tsoiref <itsoiref@redhat.com>
    Made-with: Cursor

 **cnimgr: add configurable grace period for CNI status monitoring**

    During CNI plugin upgrades (e.g. OVN-K daemonset rollout), the plugin
    becomes briefly unavailable (~10-15s). Without a grace period, CRI-O
    immediately reports NetworkReady=false to kubelet, which marks the node
    NotReady and disrupts pod scheduling.

    Add a configurable grace period (default 60s) that tolerates brief CNI
    disruptions. When the plugin was previously healthy and a status check
    fails, CRI-O waits for the grace period before reporting unhealthy. If
    the plugin recovers within the grace window, no disruption is visible
    to kubelet or the scheduler.

    The grace period only applies to Phase 2 (runtime monitoring). Startup
    polling (Phase 1) is unaffected -- failures are reported immediately
    until the plugin is first ready.

    Configuration:
    - crio.conf: cni_status_grace_period = "60s" under [crio.network]
    - CLI: --cni-status-grace-period=60s
    - Env: CNI_STATUS_GRACE_PERIOD=60s
    - Set to 0s for immediate reporting (no grace)

```release-note
Add `enable_cni_status_monitoring` config option (default false) and
`cni_status_grace_period` (default 60s) to gate continuous CNI STATUS
monitoring and tolerate brief plugin disruptions during upgrades.
```

## Summary by CodeRabbit

* **New Features**
  * Added a new cni_status_grace_period configuration option under [crio.network] that controls continuous CNI plugin health monitoring. When set to a positive duration (e.g. "1m"), CRI-O continuously polls the CNI plugin status every 5 seconds after startup and waits for the configured grace period before reporting NetworkReady=false to kubelet. This tolerates brief CNI disruptions during plugin upgrades (e.g. OVN-Kubernetes daemonset rollout, where the plugin is unavailable for ~10-15s) without unnecessarily marking the node as NotReady. When set to "0s" (the default), continuous monitoring is disabled and plugin health is only determined at startup, preserving the previous behavior. The option can be configured via crio.conf, the --cni-status-grace-period CLI flag, or the CNI_STATUS_GRACE_PERIOD environment variable.

* **Configuration**
  * Exposed both options in CLI flags, TOML config keys, and config template.

* **Documentation**
  * Updated manpage and config docs with behavior, defaults, and grace-period semantics.

* **Shell Completions**
  * Enhanced Bash, Fish, and Zsh completions for the new flags.


bitoku:
Hi @tsorya ! Thanks for the PR.
Can you add a release note?
Also you need to run `make completions-generation` and `make docs-generation`.

tsorya:
> Hi @tsorya ! Thanks for the PR. Can you add a release note? Also you need to run `make completions-generation` and `make docs-generation`.

done

completions/bash/crio
```diff
@@ -52,6 +53,7 @@ h
 --device-ownership-from-security-context
 --disable-hostport-mapping
 --drop-infra-ctr
+--enable-cni-status-monitoring
```
haircommander: do we really think setting period to 0 is a case people will do? I kind of want 0 to disable the feature and to drop this boolean from the config

tsorya: Sounds totally logical. Was thinking the same but though maybe boolean will be more clear. We change the pr

tsorya:
@haircommander done

pkg/config/template.go
```diff
@@ -1644,6 +1649,17 @@ const templateStringCrioNetworkPluginDirs = `# Paths to directories where CNI pl
...
 `
 
+const templateStringCrioNetworkCNIStatusGracePeriod = `# Enable continuous CNI STATUS monitoring with the given grace period.
+# When set to "0s" (default), monitoring is disabled and plugin health is
+# only determined at startup; runtime failures will not be detected.
+# When set to a positive duration (e.g. "1m"), a background goroutine polls
+# the plugin every 5 seconds and waits for this grace period before marking
+# the node not-ready, tolerating brief CNI disruptions during plugin
+# upgrades (e.g. OVN-K daemonset rollout).
+{{ $.Comment }}cni_status_grace_period = "{{ .CNIStatusGracePeriod }}"
```
haircommander, line 1659: hm there's nothing in toml making sure this is a duration string. what happens when you specify something bogus?

tsorya: Added a unitest for the case you pointed, field will be validated in go rather than in toml.
Should have same behavior as pull_progress_timeout

haircommander:
/lgtm cancel
/approve

@cri-o/cri-o-maintainers PTAL

internal/config/cnimgr/cnimgr.go
```diff
@@ -78,12 +93,35 @@ func (c *CNIManager) statusPollFunc(ctx context.Context, isStartup bool) (bool,
...
 		c.lastError = err
 
 		return false, nil
 	}
 
+	// Plugin is healthy -- reset grace period timer
+	c.firstFailureTime = time.Time{}
+
 	if c.lastError != nil {
+		logrus.Infof("CNI plugin is now healthy (was: %v)", c.lastError)
```
bitoku, line 120: I'm wondering how often we see this. If it's too often, it should be debug.

internal/config/cnimgr/cnimgr.go
```diff
@@ -78,12 +93,35 @@ func (c *CNIManager) statusPollFunc(ctx context.Context, isStartup bool) (bool,
...
+				c.firstFailureTime = time.Now()
+				logrus.Warnf("CNI plugin status check failed, will report unhealthy after grace period (%v): %v", c.gracePeriod, err)
+			}
+
+			if time.Since(c.firstFailureTime) < c.gracePeriod {
+				return false, nil
+			}
+
+			logrus.Errorf("CNI plugin unhealthy beyond grace period (%v): %v", c.gracePeriod, err)
+		} else if c.lastError == nil {
```
bitoku: This is unreachable.

internal/config/cnimgr/cnimgr_test.go
```diff
@@ -79,6 +79,7 @@ func newTestManager(plugin *fakeCNIPlugin) *CNIManager {
 		cancel:              cancel,
 		initPollInterval:    testPollInterval,
 		monitorPollInterval: testPollInterval,
+		gracePeriod:         time.Millisecond,
```
bitoku: why time.Millisecond? It's not default value

internal/config/cnimgr/cnimgr_test.go
bitoku, line 73: We can reuse newTestManagerWithGrace here.

docs/crio.8.md
```diff
@@ -207,6 +208,8 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--cni-plugin-dir**="": CNI plugin binaries directory.
 
+**--cni-status-grace-period**="": Enable continuous CNI STATUS monitoring with the given grace period. When set to 0 (default), monitoring is disabled and plugin health is only determined at startup. When set to a positive duration (e.g. 1m), a background goroutine polls the plugin every 5s and waits for this grace period before marking the node not-ready. (default: 0s)
```
bitoku, line 211: It seems it's breaking the current default behavior.
Before this PR, it polled continuously, but now not.
grace_period == 0 should mean it reports the status immediately, not stop polling.

bitoku (commented):
I want integration tests for this feature.

docs/crio.8.md
tsorya: @haircommander WDYT? We need someway to have disabled by default i believe, the question is what to set here?

internal/config/cnimgr/cnimgr.go
tsorya: It should not happened a lot but I think it should be helpful to see that CNI is back to normal no?

docs/crio.8.md
haircommander: I think we want to break the previous default behavior because the previous default behavior made a bunch of events happen that weren't desired

bitoku: Then I guess, should we have decent grace period by default?

bitoku: it's still breaking the current behavior though. but it makes sense if it's a desired behavior.

tsorya: In my case it is a desired behavior, previous change was too optimistic

tsorya:
> I want integration tests for this feature.

@bitoku added integration tests, can you take a look please?

docs/crio.8.md
bitoku: ok the explanation is descriptive and the default behavior is the same as before the monitoring was introduced, probably good.

internal/config/cnimgr/cnimgr.go
```diff
@@ -46,22 +48,35 @@ func New(defaultNetwork, networkDir string, pluginDirs ...string) (*CNIManager,
...
 		func(ctx context.Context) (bool, error) {
 			return c.statusPollFunc(ctx, true)
 		})
+}
+
+func (c *CNIManager) pollContinuously(ctx context.Context) {
+	c.pollUntilReady(ctx)
+
+	if c.gracePeriod <= 0 {
+		logrus.Info("Continuous CNI STATUS monitoring is disabled (cni_status_grace_period is 0)")
```
bitoku: nit: it can be minus.

```suggestion
		logrus.Info("Continuous CNI STATUS monitoring is disabled (cni_status_grace_period is %s)", c.gracePeriod)
```

internal/config/cnimgr/cnimgr.go
```diff
@@ -78,12 +93,37 @@ func (c *CNIManager) statusPollFunc(ctx context.Context, isStartup bool) (bool,
...
+		}
+
 		c.lastError = err
 
 		return false, nil
 	}
 
+	// Plugin is healthy -- reset grace period timer
+	if !c.firstFailureTime.IsZero() {
+		logrus.Debugf("CNI plugin recovered within grace period (after %v)", time.Since(c.firstFailureTime))
```
bitoku: firstFailureTime is never reset unless it succeeds.
That means it can recovers after grace period, but it logs `within grace period`. That contradicts the actual behavior.

tsorya: Totally right, ill just remove it

tsorya:
@bitoku Fixed the comments but seems like CI is broken and all the tests fails with 
{  release "release-latest" failed: could not watch pod: the pod ci-op-9gnnndpl/release-latest failed after 11m13s (failed containers: release): ContainerFailed one or more containers exited

PR #9921, opened 2026-05-04
runtimehandlerhooks: document irqbalance restart coalescing behavior

oblau:
Expand the comment in handleIRQBalanceRestart to explain that systemd coalesces restart requests arriving close in time, making the actual restart count lower than the number of triggering pods.

#### What type of PR is this?

/kind documentation

Adds documentation for a subtle systemd coalescing behavior that is not
widely known.
Without this context, it is hard to understand why irqbalance restart
counts caused by pod activity do not always match the number of pods,
and why start-limit-hit may trigger in some runs but not others with
the exact same setup.

None

Context came from investigating OCPBUGS-45112, where the coalescing behavior
made the bug hard to reproduce and a functional test unreliable.
Relates to https://github.com/cri-o/cri-o/pull/8834

```release-note
None
```

## Release Notes

* **Documentation**
  * Updated internal code documentation for improved clarity.


MarSik (approved):
/lgtm

oblau:
CC @haircommander

PR #9896, opened 2026-04-17
for release 1.33: netns file cleanup after CRI-O restart with invalid namespace

ngopalak-redhat:
#### Does this PR introduce a user-facing change?

```release-note

```

PR #9923, opened 2026-05-05
netns file cleanup after CRI-O restart with invalid namespace

ngopalak-redhat:
#### What type of PR is this?

/kind bug

This fix ensures that invalid namespace files are properly cleaned up even when the namespace handle is nil.

This was initially raised for 1.33 in https://github.com/cri-o/cri-o/pull/9896 , I'm raising this to the main branch. This should merge first

```release-note
None
```

Test output:
```
1..13
ok 1 ensure correct hostname
ok 2 ensure correct hostname for hostnetwork:true
ok 3 Check for valid pod netns CIDR
ok 4 Ensure correct CNI plugin namespace/name/container-id arguments
ok 5 Connect to pod hostport from the host # skip node configured with cgroupv2 flakes this test sometimes
ok 6 Clean up network if pod sandbox fails
ok 7 Clean up network if pod sandbox fails after plugin success
ok 8 Clean up network if pod sandbox gets killed
ok 9 Clean up network if pod netns gets destroyed
ok 10 Network recovery after reboot with destroyed netns
ok 11 CNI teardown called even with missing or invalid netns to prevent IP leaks
ok 12 pod deletion succeeds when NetNS path is missing
ok 13 netns file cleanup after CRI-O restart with invalid namespace
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Network namespace cleanup now reliably removes stale or orphaned namespace files even when namespace handles are missing or already closed, preventing leaked namespace files after restarts.

* **Tests**
  * Added an end-to-end test that validates removal of invalid/stale network namespace files across service restart and verifies new pods reach ready state.

### DIFF
--- a/test/network.bats
+++ b/test/network.bats
@@ -281,3 +281,76 @@ function check_networking() {
 
 	run ! crictl inspectp "$pod_id"
 }
+
+@test "netns file cleanup after CRI-O restart with invalid namespace" {
+	# This test reproduces the bug where invalid netns files are not cleaned up
+	# after CRI-O restart, preventing pods from restarting.
+	start_crio
+
+	# Create and run a pod
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+
+	# Get the network namespace path
+	NETNS_PATH=/var/run/netns/
+	NS=$(crictl inspectp "$pod_id" |
+		jq -er '.info.runtimeSpec.linux.namespaces[] | select(.type == "network").path | sub("'$NETNS_PATH'"; "")')
+
+	echo "Pod $pod_id created with netns: $NETNS_PATH$NS"
+
+	# Verify the netns file exists before stopping
+	[[ -f "$NETNS_PATH$NS" ]] || [[ -f "/run/netns/$NS" ]]
+
+	# Stop the pod (this marks it as stopped in CRI-O state)
+	crictl stopp "$pod_id"
+
+	# After stopping, the namespace might already be cleaned up by CRI-O
+	# We need to recreate the invalid file scenario that happens after reboot
+	# where the kernel namespace is gone but the file remains
+
+	# Try to delete the kernel namespace if it still exists (may fail if already gone)
+	ip netns del "$NS" 2> /dev/null || true
+
+	# Create a fake/invalid netns file (kernel namespace is gone but file exists)
+	touch "$NETNS_PATH$NS" || touch "/run/netns/$NS"
+
+	# Use whichever path actually exists
+	if [[ -f "/run/netns/$NS" ]]; then
+		NETNS_PATH=/run/netns/
+	fi
+
+	# Verify the invalid file exists
+	[[ -f "$NETNS_PATH$NS" ]]
+	echo "Invalid netns file created at: $NETNS_PATH$NS"
+	ls -la "$NETNS_PATH$NS" || true
+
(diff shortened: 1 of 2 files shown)


test/network.bats
```diff
@@ -281,3 +281,83 @@ function check_networking() {
...
+	echo "After CRI-O restart, checking if invalid file still exists..."
+	if [[ -f "$NETNS_PATH$NS" ]]; then
+		echo "BUG CONFIRMED: Invalid netns file still exists: $NETNS_PATH$NS"
+		ls -la "$NETNS_PATH$NS"
+	else
+		echo "File was cleaned up (bug may be fixed)"
+	fi
+
+	# Try to remove the pod - this should cleanup the invalid file
+	crictl rmp -f "$pod_id" 2>&1 || true
```
saschagrunert: Swallowing the exit status means a regression in pod removal won't be caught. Consider asserting success directly or using `run` to capture the status.

test/network.bats
```diff
@@ -281,3 +281,83 @@ function check_networking() {
...
+		echo "TEST FAILED: Bug reproduced - netns file leaked"
+		return 1
+	else
+		echo "SUCCESS: Invalid netns file was properly cleaned up"
+	fi
+
+	# Verify we can create a new pod now (would fail with leaked file)
+	new_pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	output=$(crictl inspectp "$new_pod_id" | jq -r '.status.state')
+	[[ "$output" == "SANDBOX_READY" ]]
```
saschagrunert, line 355: Same note as on #9896: these lines are unreachable. On the success path the file is already verified removed. On the failure path, `return 1` on line 351 exits first. The `cleanup_test` teardown handles leftover state anyway, so these can just be dropped.

test/network.bats
ngopalak-redhat: Fixed

test/network.bats
ngopalak-redhat: Fixed

saschagrunert (changes requested):
The Go fix looks good. One remaining nit from the previous review: the `rm -f` cleanup lines (346-347) are unnecessary since `cleanup_test` teardown handles leftover state. Please drop them.

test/network.bats
```diff
@@ -281,3 +281,79 @@ function check_networking() {
...
+	# Try to remove the pod - this should cleanup the invalid file
+	crictl rmp -f "$pod_id"
+
+	echo "After pod removal, verifying file cleanup..."
+	if [[ -f "$NETNS_PATH$NS" ]]; then
+		echo "LEAKED: Invalid netns file was not cleaned up: $NETNS_PATH$NS"
+		ls -la "$NETNS_PATH$NS" || true
+		# Cleanup for test
+		rm -f "$NETNS_PATH$NS"
+		rm -f "/run/netns/$NS" 2> /dev/null || true
```
saschagrunert: These can be dropped, `cleanup_test` teardown handles leftover state.

saschagrunert (changes requested):

test/network.bats
ngopalak-redhat: Thanks, I have addressed this

saschagrunert (approved):

bitoku:
/cherry-pick release-1.36 release-1.35 release-1.34 release-1.33

PR #9927, opened 2026-05-06
Validate that /run/netns is pointing to nsfs file

ngopalak-redhat:
#### What type of PR is this?
/kind bug

### Problem
  When a directory-level overmount (e.g., `mount --bind /run/netns /run/netns`) is created after CRI-O creates sandbox bind mounts, earlier bind mounts become shadowed - visible in mountinfo but unreachable via
  path lookup. This causes `setns EINVAL` errors when containers try to use the namespace.

  Customer scenario: ovnkube-node creates overmount at /run/netns after CRI-O restart, shadowing stopped sandbox bind mounts. CRI-O fails to detect the issue because errors for stopped sandboxes were swallowed.

  ### Solution
  - Add `ShadowedMountError` type to identify mount shadowing
  - Update `GetNamespace()` to wrap `NSPathNotNSErr` as `ShadowedMountError`
  - Never swallow shadowed mount errors, even for stopped sandboxes
  - Auto-delete broken sandboxes during CRI-O restart

  ### Testing
  Added E2E test reproducing the exact customer scenario with mount overmount, filesystem type validation, and auto-deletion verification.

The bug is based on customer analysis of a failure. The analysis was done by AI and the customer shared the results with us. The key parts of the analysis are these:

**/run/netns mount hierarchy (inside container-mount-namespace)**
Mount 273 — tmpfs at /run
Mount 12967 — ← tmpfs[/netns] at /run/netns
Mount 1077 — nsfs at /run/netns/3eb3af07-… (parent 12967 ✓)
Mount 2174 — nsfs at /run/netns/e3dd4dac-… (parent 12967 ✓)
… all other working sandboxes (parent 12967 ✓)
PATH LOOKUP STOPS HERE — files below are unreachable
Mount 7437 — nsfs[net:4026539247] at /run/netns/cc30bf0a-… SHADOWED
Mount 14629 — nsfs[net:4026540775] at /run/netns/921b97ff-… SHADOWED
Mount 16795 — nsfs[net:4026543886] at /run/netns/5b51a85c-… SHADOWED

**Startup Race — Corrected Timeline, May 5 2026**
07:08:14 | Node boots: /run tmpfs created fresh. All previous namespaces gone.
07:08:37 | container-mount-namespace.service starts: Creates private mount NS mnt:[4026537777] via unshare --mount --propagation slave.
07:08:51 | CRI-O and kubelet start: Both enter the container-mount-namespace. No overmount exists yet.
07:09:47 | Race window begins: 3 broken sandboxes created by pinns (parent 273). All 3 have parent=273 because mount 12967 does not yet exist.
~07:09:50 | Race window ends: ovnkube-node creates /run/netns self-bind-mount (mount 12967), shadowing the 3 pinns bind mounts.
07:21:12 | MCO restarts CRI-O: CRI-O restores records from JSON but does not re-run pinns. The shadowed mounts remain shadowed.
07:56 onward | First setns EINVAL: Containers crash. crun opens the path but gets the empty tmpfs file from the overmount. setns() returns EINVAL.

In short it validates that "/run/netns" is pointing to nsfs file and not created by other mounts.

**Local testing**
I couldn't fully recreate the issue, but when I investigated the `/run/netns` folder 
```
sh-5.1# findmnt -o TARGET,PARENT,FSTYPE -T /run/netns/8e4e68d0-92a0-4a6a-9809-9d52cbe555c3
TARGET                                          PARENT FSTYPE
/run/netns/8e4e68d0-92a0-4a6a-9809-9d52cbe555c3   9049 nsfs
grep "/run/netns" /proc/self/mountinfo
9057 9049 0:4 net:[4026531840] /run/netns/90606e92-cea2-4f11-af31-473a13d64231 rw master:235 - nsfs nsfs rw
9061 9049 0:4 net:[4026531840] /run/netns/bbaf7a84-532a-4138-ad4a-ad7a43357a4f rw master:438 - nsfs nsfs rw
...
```
Normally the parent is nsfs. In the customer case its changed.

Also the magic number for nsfs used in the code can be confirmed from a node:
```
stat -f -c %t /run/netns/*
6e736673
6e736673
```

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Detect and treat invalid or shadowed network-namespace mounts as broken; affected stopped sandboxes are now detected, logged, and cleaned up on startup instead of remaining silently orphaned.
  * Namespace-join failures caused by invalid mounts are surfaced even for stopped sandboxes to ensure proper cleanup.

* **Tests**
  * Added integration tests validating detection and cleanup of shadowed/invalid netns bind mounts.

### DIFF
--- a/internal/config/nsmgr/types_linux.go
+++ b/internal/config/nsmgr/types_linux.go
@@ -20,27 +20,25 @@ func supportedNamespacesForPinning() []NSType {
 	return []NSType{NETNS, IPCNS, UTSNS, USERNS, PIDNS}
 }
 
-// ShadowedMountError indicates a namespace bind mount is shadowed by a directory overmount.
-// This error should never be ignored, even for stopped sandboxes, as it will cause
-// setns EINVAL when containers try to use the namespace.
-type ShadowedMountError struct {
+// InvalidNamespaceMountError indicates a namespace path does not point to a valid namespace filesystem.
+type InvalidNamespaceMountError struct {
 	Path string
 	Err  error
 }
 
-func (e *ShadowedMountError) Error() string {
-	return fmt.Sprintf("namespace bind mount at %s is shadowed by directory overmount: %v", e.Path, e.Err)
+func (e *InvalidNamespaceMountError) Error() string {
+	return fmt.Sprintf("namespace path %s is not a valid namespace filesystem: %v", e.Path, e.Err)
 }
 
-func (e *ShadowedMountError) Unwrap() error {
+func (e *InvalidNamespaceMountError) Unwrap() error {
 	return e.Err
 }
 
-// IsShadowedMountError returns true if the error indicates a shadowed mount.
-func IsShadowedMountError(err error) bool {
-	var shadowedErr *ShadowedMountError
+// IsInvalidNamespaceMountError returns true if the error indicates an invalid namespace mount.
+func IsInvalidNamespaceMountError(err error) bool {
+	var invalidErr *InvalidNamespaceMountError
 
-	return errors.As(err, &shadowedErr)
+	return errors.As(err, &invalidErr)
 }
 
 type PodNamespacesConfig struct {
@@ -126,10 +124,9 @@ func (n *namespace) Remove() error {
 func GetNamespace(nsPath string, nsType NSType) (Namespace, error) {
 	ns, err := nspkg.GetNS(nsPath)
 	if err != nil {
-		// Wrap NSPathNotNSErr to indicate potential mount shadowing
 		var nsPathNotNSErr nspkg.NSPathNotNSErr
 		if errors.As(err, &nsPathNotNSErr) {
-			return &namespace{nsType: nsType, nsPath: nsPath, closed: true}, &ShadowedMountError{Path: nsPath, Err: err}
+			return &namespace{nsType: nsType, nsPath: nsPath, closed: true}, &InvalidNamespaceMountError{Path: nsPath, Err: err}
 		}
 		// Failed to GetNS. It's possible this is expected (pod is stopped).
 		return &namespace{nsType: nsType, nsPath: nsPath, closed: true}, err
--- a/internal/config/nsmgr/types_linux.go
+++ b/internal/config/nsmgr/types_linux.go
@@ -20,6 +20,29 @@ func supportedNamespacesForPinning() []NSType {
 	return []NSType{NETNS, IPCNS, UTSNS, USERNS, PIDNS}
 }
 
+// ShadowedMountError indicates a namespace bind mount is shadowed by a directory overmount.
+// This error should never be ignored, even for stopped sandboxes, as it will cause
+// setns EINVAL when containers try to use the namespace.
+type ShadowedMountError struct {
+	Path string
+	Err  error
+}
+
+func (e *ShadowedMountError) Error() string {
+	return fmt.Sprintf("namespace bind mount at %s is shadowed by directory overmount: %v", e.Path, e.Err)
+}
+
+func (e *ShadowedMountError) Unwrap() error {
+	return e.Err
+}
+
+// IsShadowedMountError returns true if the error indicates a shadowed mount.
+func IsShadowedMountError(err error) bool {
+	var shadowedErr *ShadowedMountError
+
+	return errors.As(err, &shadowedErr)
+}
+
 type PodNamespacesConfig struct {
 	Namespaces []*PodNamespaceConfig
 	IDMappings *idtools.IDMappings
@@ -103,6 +126,11 @@ func (n *namespace) Remove() error {
 func GetNamespace(nsPath string, nsType NSType) (Namespace, error) {
 	ns, err := nspkg.GetNS(nsPath)
 	if err != nil {
+		// Wrap NSPathNotNSErr to indicate potential mount shadowing
+		var nsPathNotNSErr nspkg.NSPathNotNSErr
+		if errors.As(err, &nsPathNotNSErr) {
+			return &namespace{nsType: nsType, nsPath: nsPath, closed: true}, &ShadowedMountError{Path: nsPath, Err: err}
+		}
 		// Failed to GetNS. It's possible this is expected (pod is stopped).
 		return &namespace{nsType: nsType, nsPath: nsPath, closed: true}, err
 	}
--- a/test/network.bats
+++ b/test/network.bats
@@ -324,26 +324,26 @@ function check_networking() {
 	ls -la "$NETNS_PATH$NS" || true
 
 	# Restart CRI-O - this triggers LoadSandbox which calls NetNsJoin
-	# NetNsJoin should fail to GetNS but will store partial namespace anyway
+	# With the shadowing detection fix, CRI-O should detect the invalid namespace
+	# and auto-delete the broken sandbox during startup
 	restart_crio
 
-	echo "After CRI-O restart, checking if invalid file still exists..."
-	if [[ -f "$NETNS_PATH$NS" ]]; then
-		echo "BUG CONFIRMED: Invalid netns file still exists: $NETNS_PATH$NS"
-		ls -la "$NETNS_PATH$NS"
-	else
-		echo "File was cleaned up (bug may be fixed)"
-	fi
+	# CRI-O should have auto-deleted the broken sandbox and cleaned up the file
+	echo "After CRI-O restart, verifying sandbox was auto-deleted..."
+	run ! crictl inspectp "$pod_id"
 
-	# Try to remove the pod - this should cleanup the invalid file
-	crictl rmp -f "$pod_id"
+	# Verify CRI-O logged the shadowed mount detection and cleanup
+	grep -q "is shadowed by directory overmount" "$CRIO_LOG" || grep -q "unknown FS magic" "$CRIO_LOG"
+	grep -q "Could not restore sandbox.*$pod_id" "$CRIO_LOG"
+	grep -q "Deleting all containers under sandbox $pod_id" "$CRIO_LOG"
+	grep -q "Successfully cleaned up network for pod $pod_id" "$CRIO_LOG"
 
-	echo "After pod removal, verifying file cleanup..."
+	# Verify the invalid file was cleaned up during auto-deletion
+	echo "Verifying invalid file was cleaned up..."
 	if [[ -f "$NETNS_PATH$NS" ]]; then
 		echo "LEAKED: Invalid netns file was not cleaned up: $NETNS_PATH$NS"
 		ls -la "$NETNS_PATH$NS" || true
-		# Fail the test - this demonstrates the bug
-		echo "TEST FAILED: Bug reproduced - netns file leaked"
+		echo "TEST FAILED: File leaked despite auto-deletion"
 		return 1
 	else
 		echo "SUCCESS: Invalid netns file was properly cleaned up"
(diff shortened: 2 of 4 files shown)


internal/config/nsmgr/types_linux.go
```diff
@@ -101,6 +102,18 @@ func (n *namespace) Remove() error {
...
+	// by a directory-level overmount (e.g., tmpfs at /run/netns created after
+	// the bind mount). This can happen when ovnkube-node or other components
+	// create a self-bind-mount at /run/netns after CRI-O has already created
+	// some sandbox bind mounts, causing the kernel VFS path lookup to bypass
+	// the earlier bind mounts.
+	if err := validateNsfsPath(nsPath); err != nil {
+		logrus.Warnf("Namespace bind mount at %s appears to be shadowed: %v", nsPath, err)
+		// Return GetNS failed
+		return &namespace{nsType: nsType, nsPath: nsPath, closed: true}, err
+	}
```
saschagrunert: The vendored `nspkg.GetNS` already performs this check. `GetNS` calls `IsNSorErr` (`vendor/github.com/containernetworking/plugins/pkg/ns/ns_linux.go:125-142`), which does `syscall.Statfs` and rejects anything that isn't `NSFS_MAGIC` or `PROCFS_MAGIC`. A shadowed path resolves through the overmount's tmpfs, so `Statfs` returns tmpfs magic and `GetNS` fails with `NSPathNotNSErr`.

That error should already propagate: for a running sandbox on CRI-O restart, `RestoreStopped()` won't set `stopped=true` (no "stopped" file on disk), so `NetNsJoin` returns the error, `LoadSandbox` fails, and `server.go` cleans up the sandbox.

Could you verify whether the existing `GetNS` error is actually swallowed in the customer's scenario? If it is, the fix should target that specific swallowing rather than adding a redundant pre-check.

internal/config/nsmgr/types_linux.go
```diff
@@ -109,3 +122,29 @@ func GetNamespace(nsPath string, nsType NSType) (Namespace, error) {
 
 	return &namespace{ns: ns, nsType: nsType, nsPath: nsPath}, nil
 }
+
+// validateNsfsPath checks if the file at nsPath is actually an nsfs file
+// Returns an error if the path does not point to an nsfs filesystem.
+func validateNsfsPath(nsPath string) error {
+	// NSFS_MAGIC is the filesystem magic number for nsfs (namespace filesystem)
+	// I got this from the kernel source code (Searched the web)
+	const nsfsMagic = 0x6e736673
```
saschagrunert: `unix.NSFS_MAGIC` is already available in the vendored `golang.org/x/sys/unix` package (same value `0x6e736673`). Use that instead of defining a local constant. Also the comment `Searched the web` is an AI artifact that should be removed.

```suggestion
	const nsfsMagic = unix.NSFS_MAGIC
```

internal/config/nsmgr/types_linux.go
```diff
@@ -101,6 +102,18 @@ func (n *namespace) Remove() error {
 // GetNamespace takes a path and type, checks if it is a namespace, and if so
 // returns an instance of the Namespace interface.
 func GetNamespace(nsPath string, nsType NSType) (Namespace, error) {
```
saschagrunert, line 124: This 6-line comment block is excessive for what boils down to "check the path is actually nsfs". If the function is kept, the name `validateNsfsPath` is self-explanatory. One short line at most.

Also, the `logrus` import was added solely for this warning log. Utility functions like `GetNamespace` shouldn't log directly; the caller should handle the error. The error message from `validateNsfsPath` already contains all the relevant info (path, expected vs actual type). Removing the log also removes the `logrus` import.

test/network.bats
```diff
@@ -281,3 +281,132 @@ function check_networking() {
...
+
+	# Create the bind mount
+	mount --bind "$NETNS_PATH" "$NETNS_PATH" || {
+		echo "Failed to create bind mount at $NETNS_PATH" >&2
+		exit 1
+	}
+	mount --make-shared "$NETNS_PATH" || {
+		echo "Failed to make mount shared at $NETNS_PATH" >&2
+		exit 1
+	}
```
saschagrunert, line 399: The `mount --bind` at `/run/netns` is never torn down. This could leak into subsequent test runs and the host. Add cleanup in a teardown or trap, e.g.:

```bash
# In a cleanup/teardown section:
umount "$NETNS_PATH" 2>/dev/null || true
```

test/network.bats
```diff
@@ -281,3 +281,132 @@ function check_networking() {
...
+	output=$(crictl inspectp "$new_pod_id" | jq -r '.status.state')
+	[[ "$output" == "SANDBOX_READY" ]]
+
+	# Verify the new bind mount is NOT shadowed (accessible via path lookup)
+	NEW_NS=$(crictl inspectp "$new_pod_id" |
+		jq -er '.info.runtimeSpec.linux.namespaces[] | select(.type == "network").path | sub("'$NETNS_PATH'"; "")')
+
+	# This bind mount should have the overmount as parent, so it's accessible
+	new_fstype=$(stat -f -c %t "$NETNS_PATH$NEW_NS" 2> /dev/null)
+	echo "New pod filesystem type: 0x$new_fstype"
```
saschagrunert, line 451: `new_fstype` is captured but never asserted. Add the validation:

```bash
[[ "$new_fstype" == "6e736673" ]]
```

test/network.bats
```diff
@@ -281,3 +281,132 @@ function check_networking() {
 
 	run ! crictl inspectp "$pod_id"
 }
+
```
saschagrunert, line 284: The test is very comment-heavy (many multi-line explanations). The bats framework and the test name `detect and handle shadowed netns bind mount after overmount` provide context. Consider trimming inline comments to single lines or removing the obvious ones (e.g., `# Verify the bind mount exists and is nsfs` before `stat -f -c %t`).

saschagrunert (changes requested):
Thanks for working on this, Neeraj. The race condition analysis from the customer is solid, and detecting shadowed mounts during restore is the right approach.

However, the core validation in `validateNsfsPath` duplicates a check that `nspkg.GetNS` already performs. The vendored `containernetworking/plugins/pkg/ns/ns_linux.go` calls `IsNSorErr()` which does `syscall.Statfs` and checks for `NSFS_MAGIC` before opening the file. A shadowed path would return tmpfs magic, causing `GetNS` to fail with `NSPathNotNSErr`.

The error from `GetNS` should already propagate through `NetNsJoin` (since `!s.stopped` for running sandboxes on restart) and cause `LoadSandbox` to fail, triggering the cleanup in `server.go:226-257`.

Could you verify whether the existing `GetNS` error is actually insufficient for the customer's scenario? If there's a specific code path where the error gets swallowed, fixing that path directly would be more targeted. If the fix is still wanted as defense-in-depth, the implementation needs some cleanup (see inline comments).

internal/config/nsmgr/types_linux.go
ngopalak-redhat: Thanks, you are right, its already checked.  I think this line:
https://github.com/cri-o/cri-o/blob/main/internal/lib/sandbox/namespaces.go#L114
```
func (s *Sandbox) NetNsJoin(nspath string) error {
	ns, err := nsJoin(nspath, nsmgr.NETNS, s.netns)
	// Regardless of error, set the namespace
	s.netns = ns
	// Only error if the sandbox is not stopped
	if err != nil && !s.stopped { // <---- This could be an issue
		return err
	}

	return nil
}
```
is swallowing the error as the sandbox is stopped.

ngopalak-redhat:
Waiting for https://github.com/cri-o/cri-o/pull/9923 to merge

ngopalak-redhat:
@bitoku @saschagrunert I'm getting closer to the fix. I plan to wait tests to succeed before asking for another review.

internal/config/nsmgr/types_linux.go
ngopalak-redhat: Made the comment short and removed validateNsfsPath

test/network.bats
ngopalak-redhat: Removed the multi-line comment

internal/config/nsmgr/types_linux.go
ngopalak-redhat: I have fixed the error handling in the caller now

test/network.bats
ngopalak-redhat: Done

test/network.bats
ngopalak-redhat: Done

ngopalak-redhat:
@bitoku @saschagrunert I have improved the error handling for this. Can you review this approach?

internal/config/nsmgr/types_linux.go
```diff
@@ -20,6 +20,29 @@ func supportedNamespacesForPinning() []NSType {
...
+// ShadowedMountError indicates a namespace bind mount is shadowed by a directory overmount.
+// This error should never be ignored, even for stopped sandboxes, as it will cause
+// setns EINVAL when containers try to use the namespace.
+type ShadowedMountError struct {
+	Path string
+	Err  error
+}
+
+func (e *ShadowedMountError) Error() string {
+	return fmt.Sprintf("namespace bind mount at %s is shadowed by directory overmount: %v", e.Path, e.Err)
```
saschagrunert: `NSPathNotNSErr` fires whenever `statfs` returns anything other than `NSFS_MAGIC` or `PROCFS_MAGIC`, not just for directory overmounts. The first test demonstrates this: it `touch`es a plain file at the namespace path (no overmount involved), yet the error says "shadowed by directory overmount." This will be misleading in production postmortems where an operator investigates mount hierarchies when the issue is actually a stale file.

Consider using a more general name that covers all non-namespace filesystem cases. The comment block is also excessive for a self-explanatory type name.

```suggestion
// InvalidNamespaceMountError indicates a namespace path does not point to a valid namespace filesystem.
type InvalidNamespaceMountError struct {
	Path string
	Err  error
}

func (e *InvalidNamespaceMountError) Error() string {
	return fmt.Sprintf("namespace path %s is not a valid namespace filesystem: %v", e.Path, e.Err)
```

This would also require renaming `IsShadowedMountError` to `IsInvalidNamespaceMountError` (and all call sites in `namespaces.go` + `types_default.go`).

test/network.bats
```diff
@@ -324,26 +324,26 @@ function check_networking() {
...
-		echo "File was cleaned up (bug may be fixed)"
-	fi
+	# CRI-O should have auto-deleted the broken sandbox and cleaned up the file
+	echo "After CRI-O restart, verifying sandbox was auto-deleted..."
+	run ! crictl inspectp "$pod_id"
 
-	# Try to remove the pod - this should cleanup the invalid file
-	crictl rmp -f "$pod_id"
+	# Verify CRI-O logged the shadowed mount detection and cleanup
+	grep -q "is shadowed by directory overmount" "$CRIO_LOG" || grep -q "unknown FS magic" "$CRIO_LOG"
```
saschagrunert: Since `ShadowedMountError.Error()` wraps `NSPathNotNSErr.Error()`, the log line contains both strings. The `||` fallback will never be reached.

```suggestion
	grep -q "is shadowed by directory overmount" "$CRIO_LOG"
```

If the error type is renamed per the other comment, update this pattern to match the new message.

test/network.bats
```diff
@@ -354,3 +354,100 @@ function check_networking() {
 	output=$(crictl inspectp "$new_pod_id" | jq -r '.status.state')
 	[[ "$output" == "SANDBOX_READY" ]]
 }
+
+@test "detect and handle shadowed netns bind mount after overmount" {
+	# Reproduces mount shadowing where directory-level overmount at /run/netns
+	# shadows earlier sandbox bind mounts, making them unreachable via path lookup.
+
+	NETNS_PATH=/var/run/netns/
+	trap 'umount "$NETNS_PATH" 2>/dev/null || true' EXIT INT TERM
```
saschagrunert: By the time the trap fires, child mounts may exist under `$NETNS_PATH` (the new pod's nsfs bind mount). Plain `umount` will fail with EBUSY and silently leave the overmount on the host, polluting subsequent tests.

```suggestion
	trap 'umount -l "$NETNS_PATH" 2>/dev/null || true' EXIT INT TERM
```

saschagrunert (changes requested):
The core fix is correct: `NSPathNotNSErr` errors should not be swallowed for stopped sandboxes during restore. A namespace file that exists but isn't on nsfs is broken regardless of sandbox state, and `LoadSandbox` should fail so cleanup runs.

Three inline comments on naming/diagnostics and test robustness.

internal/config/nsmgr/types_linux.go
ngopalak-redhat: I have adopted this change

test/network.bats
ngopalak-redhat: Thanks, made this change

bitoku:
cross-posting here too.

It seems like the issue happened in this scenario.
```
1. cri-o started the sandbox and the container
--- after some time ---
2. cri-o restarted
3. cri-o loaded the sandbox and the container (still running)
(GetNS verified netns availability. netns was still accessible)
4. overmount happened
--- after some time ---
5. the container crashed
6. cri-o tried to restart the container but failed
(netns was not accessible after 4)
```
The PR makes `NetNSJoin` fail when it's overmounted.
However `NetNSJoin` is called in `LoadSandbox` , which mean it's called at 3.
At that point the overmount doesn't happen, so the PR won't fix the issue, I guess.

My question is whether overmounting the directory after CRI-O starts is legit behavior. I don't think so.

bitoku:
@ngopalak-redhat can we close this?

bitoku:
closing. feel free to reopen if needed.

PR #9937, opened 2026-05-08
Fold all the Accept Headers.

### LINKED ISSUE #9902: CRI-O sends multiple `Accept` headers when pulling images
### What happened?

We observed CRI-O setting multiple `Accept` headers when requesting a manifest:

```
10.0.0.1:65535: GET https://harbor.example/v2/foo/bar/manifests/v0.0.1
    user-agent: cri-o/1.33.9 os/linux arch/amd64
    accept: application/vnd.oci.image.manifest.v1+json
    accept: application/vnd.docker.distribution.manifest.v2+json
    accept: application/vnd.docker.distribution.manifest.v1+prettyjws
    accept: application/vnd.docker.distribution.manifest.v1+json
    accept: application/vnd.docker.distribution.manifest.list.v2+json
    accept: application/vnd.oci.image.index.v1+json
    authorization: Bearer [...]
    docker-distribution-api-version: registry/2.0
    [...]
```

A single `Accept` header with multiple values separated by commas:

```
10.0.0.1:65535: GET https://harbor.example/v2/foo/bar/manifests/v0.0.1
    user-agent: cri-o/1.33.9 os/linux arch/amd64
    accept: application/vnd.oci.image.manifest.v1+json, application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.v1+prettyjws, application/vnd.docker.distribution.manifest.v1+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.index.v1+json
```

Instruct CRI-O to pull an image from a host where you can observe the raw inbound request headers.

This is _technically_ valid per RFC 9110 (Sections [5.2](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-lines-and-combined-fi), [5.3](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-order), and [12.5.1](https://www.rfc-editor.org/rfc/rfc9110.html#name-accept)).

However, some proxies like Amazon CloudFront drop all but the first `Accept` header in a request. This prevents container registries behind the proxy from seeing the full set of MIME types that CRI-O supports. In our case, pulls would fail with Harbor logging the error message: `OCI index found, but accept header does not support OCI indexes.`

CRI-O 1.33.9 on OpenShift 4.20, running in AWS

xw19:
I've been looking further into [RFC 9110, Section 5.3](https://www.rfc-editor.org/rfc/rfc9110#section-5.3), and it seems that `Set-Cookie` is the primary exception we need to account for. The specification is explicit about how recipients should handle multiple field lines:
 
> "This means that, aside from the well-known exception noted below, a sender MUST NOT generate multiple field lines with the same name in a message (whether in the headers or trailers) or append a field line when a field line of the same name already exists in the message, unless that field's definition allows multiple field line values to be recombined as a comma-separated list (i.e., at least one alternative of the field's definition allows a comma-separated list, such as an ABNF rule of #(values) defined in Section 5.6.1)."
 
> "Note: In practice, the `Set-Cookie` header field ([COOKIE]) often appears in a response message across multiple field lines and does not use the list syntax, violating the above requirements on multiple field lines with the same field name. Since it cannot be combined into a single field value, recipients ought to handle `Set-Cookie` as a special case while processing fields. (See Appendix A.2.3 of [Kri2001] for details.)"
 
From this it safe to assume that if we fold the accept headers it should be generally safe.

Fixes: https://github.com/cri-o/cri-o/issues/9902

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image manifest handling compatibility with proxies that have header line limitations.

### DIFF
--- a/internal/storage/image.go
+++ b/internal/storage/image.go
@@ -49,6 +49,14 @@ const (
 //nolint:gochecknoinits // Init function is intentional here.
 func init() {
 	reexec.Register("crio-pull-image", pullImageChild)
+	// Collapse Accept MIME types into one comma-separated value so proxies like
+	// CloudFront (which drop all but the first header line) see the full list.
+	// See https://github.com/cri-o/cri-o/issues/9902.
+	if len(manifest.DefaultRequestedManifestMIMETypes) > 0 {
+		manifest.DefaultRequestedManifestMIMETypes = []string{
+			strings.Join(manifest.DefaultRequestedManifestMIMETypes, ", "),
+		}
+	}
 }
 
 // ImageResult wraps a subset of information about an image: its ID, its names,


xw19:
@cri-o/cri-o-maintainers PTAL.

bitoku:
Hi thank you @xw19 .

I read it in this way.
> a sender MUST NOT generate multiple field lines with the same name in a message
> unless that field's definition allows multiple field line values to be recombined as a comma-separated list (i.e., at least one alternative of the field's definition allows a comma-separated list, such as an ABNF rule of #(values) defined in Section 5.6.1).

and ACCEPT accepts comma-separated list
https://www.rfc-editor.org/rfc/rfc9110#name-accept

So I think multiple ACCEPT is acceptable

xw19:
Agree but the CDNs (in this case cloudfront) may cause this issue .Another instance is documented here 

https://support.gitlab.com/hc/en-us/articles/20331100094364-Failed-to-pull-OCI-Image-from-GitLab-Container-Registry

I think there is some value here. But I would also be worried if this change breaks something else.

internal/storage/image.go
```diff
@@ -49,6 +49,14 @@ const (
 //nolint:gochecknoinits // Init function is intentional here.
 func init() {
 	reexec.Register("crio-pull-image", pullImageChild)
+	// Collapse Accept MIME types into one comma-separated value so proxies like
+	// CloudFront (which drop all but the first header line) see the full list.
+	// See https://github.com/cri-o/cri-o/issues/9902.
+	if len(manifest.DefaultRequestedManifestMIMETypes) > 0 {
```
mtrmac, line 55: For the record, `DefaultRequestedManifestMIMETypes`
- is not intended to be mutable; it is a `var` only because Go does not provide constants of that type
- Defines a _list_ of MIME types: putting a comma-separated string there is completely violating the documentation, with nominally-unpredictable consequences.

PR #9938, opened 2026-05-08
Retain container runtime information

### LINKED ISSUE #9521: Change Request: crio should not block pod termination when the runtime class of the pod no longer exists
Hi CRI-O Team,

We deploy the [nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit) as a daemonset which overrides the crio config toml and adds a new runtime handler(this is defined as the `nvidia` runtime). Once the nvidia-container-toolkit is deployed, we spin up other components that leverage the newly declared runtime classes.

Where our problems arise is during the tearing down/undeployment of these components. We have logic in our `nvidia-container-toolkit` where all of previously applied crio config overrides (including runtime handler declarations) are reverted as part of its termination routine. We have found that this introduces a bad race condition where the other components that leverage the `nvidia` runtime are stuck indefinitely in the Terminating state due to `KillPodSandboxErrors`. These `KillPodSandboxErrors` happen as the `nvidia` runtime that is referenced no longer exists. 

I've attached the screenshot of this error for your reference. Please let me know you if need more details or if anything in this issue description is unclear. Thank you!

Image

xw19:
Caching ConfigRuntimes

Fixes: https://github.com/cri-o/cri-o/issues/9521

```release-note
Retain per-container cached runtime information
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored containers now pre-populate and retain per-container runtime-handler cache so stop/delete/recreate succeed even if the configured runtime is later removed or changed.
  * Improved mirror-registries config watcher debounce to better coalesce rapid file events and trigger reliable reloads.

* **Tests**
  * Added tests covering runtime-handler caching, fallback behavior, idempotence, container isolation, and config-reload regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cri-o/cri-o/pull/9938)

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -405,6 +405,9 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
 		return sb, err
 	}
 
+	// Pre-populate the runtime handler config cache while the handler is still available in the live config.
+	c.Runtime().EnsureRuntimeHandlerConfig(scontainer)
+
 	// We should restore the infraContainer to the container state store
 	c.AddInfraContainer(ctx, scontainer)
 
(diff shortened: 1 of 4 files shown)


xw19:
@cri-o/cri-o-maintainers PTAL

xw19:
@coderabbitai review

bitoku:
Thank you @xw19 .
Though I suggested caching, we need more discussion about it.

How common to change the runtime while CRI-O is running?
What if the config changes during operation? 
When we should use cache, what if the part of it is changed?

First I'd like to explore any other way.

PR #9939, opened 2026-05-09
Feat: Add pinned_artifacts support with crio artifact pull

### LINKED ISSUE #9583: Speed up AI model-loading by having a pre-download mechanism for OCI volumes

xw19:
Fixes: https://github.com/cri-o/cri-o/issues/9583

/kind feature

```release-note
Fetch pinned Artifacts from CLI and Fetch during startup.
```

## Summary by CodeRabbit

* **New Features**
  * Added an `artifact pull` CLI command and a `--pinned-artifacts` option; server pre-pulls listed OCI artifacts asynchronously at startup and on SIGHUP, skipping invalid refs and logging failures without blocking.

* **Documentation**
  * Updated manpage and config docs and included the new config template entry for `pinned_artifacts`.

* **Shell Completions**
  * Updated Bash, Fish, and Zsh completions to include the new command and option.

* **Tests**
  * Added unit, integration, and E2E tests covering CLI, pre-pull behavior, and reload semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cri-o/cri-o/pull/9939)

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -10,11 +10,15 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
+	"go.podman.io/common/libimage"
+	"go.podman.io/image/v5/docker"
+	"go.podman.io/image/v5/docker/reference"
 	imageTypes "go.podman.io/image/v5/types"
 	"go.podman.io/storage/pkg/idtools"
 	storageTypes "go.podman.io/storage/types"
@@ -101,6 +105,10 @@ type Server struct {
 	hooksRetriever *runtimehandlerhooks.HooksRetriever
 
 	artifactStore *ociartifact.Store
+	// pinnedArtifactsPullMu serializes concurrent pullPinnedArtifacts goroutines
+	// (triggered at startup and on each SIGHUP) so they don't race on the shared
+	// store or pull the same artifact concurrently.
+	pinnedArtifactsPullMu sync.Mutex
 }
 
 // pullArguments are used to identify a pullOperation via an input image name and
@@ -478,6 +486,8 @@ func New(
 		artifactStore:            artifactStore,
 	}
 
+	go s.pullPinnedArtifacts(ctx, append([]string(nil), s.config.PinnedArtifacts...))
+
 	if s.config.EnablePodEvents {
 		// creating a container events channel only if the evented pleg is enabled
 		s.ContainerEventsChan = make(chan types.ContainerEventResponse, 1000)
@@ -627,6 +637,10 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 			// Block until the signal is received
 			<-ch
 
+			// Snapshot the artifact list before reload so we can detect whether
+			// it actually changed (avoids network I/O on unrelated reloads).
+			prevArtifacts := append([]string(nil), s.config.PinnedArtifacts...)
+
 			if err := s.config.Reload(ctx); err != nil {
 				log.Errorf(ctx, "Unable to reload configuration: %v", err)
 
@@ -639,6 +653,18 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 			// pinned and sandbox/pause images, we need to update them
 			s.ContainerServer.StorageImageServer().UpdatePinnedImagesList(append(s.config.PinnedImages, s.config.PauseImage))
 			s.artifactStore.SetPinnedImageRegexps(s.ContainerServer.StorageImageServer().PinnedImageRegexps())
+
+			// Only re-pull when the list changed; use sorted comparison to
+			// match ReloadPinnedArtifacts' order-insensitive equality check.
+			slices.Sort(prevArtifacts)
+
+			currArtifacts := append([]string(nil), s.config.PinnedArtifacts...)
+			slices.Sort(currArtifacts)
+
+			if !slices.Equal(prevArtifacts, currArtifacts) {
+				go s.pullPinnedArtifacts(ctx, append([]string(nil), s.config.PinnedArtifacts...))
+			}
+
 			log.Infof(ctx, "Configuration reload completed")
 			// Print the current configuration.
 			tomlConfig, err := s.config.ToString()
@@ -653,12 +679,63 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 	log.Infof(ctx, "Registered SIGHUP reload watcher")
 }
 
+// pullPinnedArtifacts pulls every artifact in refs.
+// Errors are logged and skipped so a single bad reference does not block the
+// others. Intended to run in a background goroutine on startup and reload.
+// refs must be a snapshot taken before launching the goroutine so the caller
+// and this function do not race on s.config.PinnedArtifacts.
+func (s *Server) pullPinnedArtifacts(ctx context.Context, refs []string) {
+	if len(refs) == 0 {
+		return
+	}
+
+	s.pinnedArtifactsPullMu.Lock()
+	defer s.pinnedArtifactsPullMu.Unlock()
+
+	log.Infof(ctx, "Pre-pulling %d pinned artifact(s)", len(refs))
+
+	seen := make(map[string]struct{}, len(refs))
+
+	for _, refStr := range refs {
+		if _, ok := seen[refStr]; ok {
+			continue
+		}
+
+		seen[refStr] = struct{}{}
+
+		named, err := reference.ParseNormalizedNamed(refStr)
+		if err != nil {
+			log.Errorf(ctx, "Failed to parse pinned artifact reference %q: %v", refStr, err)
+
+			continue
+		}
+
+		imageRef, err := docker.NewReference(named)
+		if err != nil {
+			log.Errorf(ctx, "Failed to create image reference for pinned artifact %q: %v", refStr, err)
+
+			continue
+		}
+
+		if _, err := s.artifactStore.Pull(ctx, imageRef, &libimage.CopyOptions{
+			RemoveSignatures: true, // OCI layout destination does not support signature storage
+		}); err != nil {
+			log.Errorf(ctx, "Failed to pull pinned artifact %q: %v", refStr, err)
+
+			continue
+		}
+
+		log.Infof(ctx, "Pinned artifact pre-pulled: %s", refStr)
+	}
+}
+
 func useDefaultUmask(ctx context.Context) {
 	const defaultUmask = 0o022
 
 	oldUmask := unix.Umask(defaultUmask)
 	if oldUmask != defaultUmask {
-		log.Infof(ctx,
+		log.Infof(
+			ctx,
 			"Using default umask 0o%#o instead of 0o%#o",
 			defaultUmask, oldUmask,
 		)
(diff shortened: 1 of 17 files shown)


xw19:
@cri-o/cri-o-maintainers PTAL

xw19:
@saschagrunert \ @bitoku Can you please look at this PR

server/server.go
```diff
@@ -639,6 +648,9 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
 			// pinned and sandbox/pause images, we need to update them
 			s.ContainerServer.StorageImageServer().UpdatePinnedImagesList(append(s.config.PinnedImages, s.config.PauseImage))
 			s.artifactStore.SetPinnedImageRegexps(s.ContainerServer.StorageImageServer().PinnedImageRegexps())
+
+			go s.pullPinnedArtifacts(ctx, append([]string(nil), s.config.PinnedArtifacts...))
```
saschagrunert: This fires on every SIGHUP regardless of whether the pinned list actually changed. `ReloadPinnedArtifacts` already detects no-op reloads (the `cmp.Equal` early return), but that signal isn't fed back here. Since `Store.Pull` intentionally re-checks the registry for the main store ("the remote tag may have been re-pointed to a different digest"), every unrelated config reload (e.g. changing log level) triggers network requests + `EnsureNotContainerImage` for all pinned artifacts.

For the target use case of large AI/ML model artifacts, this could be significant unnecessary I/O.

Consider having `ReloadPinnedArtifacts` return a `bool`:
```go
if changed := c.ReloadPinnedArtifacts(newConfig); changed {
    go s.pullPinnedArtifacts(ctx, append([]string(nil), s.config.PinnedArtifacts...))
}
```

saschagrunert (commented):
Overall this is a well-structured PR that follows existing CRI-O patterns closely. The config/reload/CLI/template/completions/docs all mirror the established `pinned_images` conventions. Race condition handling is solid (slice snapshot + mutex). Tests are comprehensive across unit, server, and integration layers.

A few items worth addressing before merge, noted inline.

server/server.go
```diff
@@ -653,6 +665,48 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
...
+
+		imageRef, err := docker.NewReference(named)
+		if err != nil {
+			log.Errorf(ctx, "Failed to create image reference for pinned artifact %q: %v", refStr, err)
+
+			continue
+		}
+
+		if _, err := s.artifactStore.Pull(ctx, imageRef, &libimage.CopyOptions{
+			RemoveSignatures: true,
```
saschagrunert: Why `RemoveSignatures: true`? If the artifact store doesn't support storing signatures, a brief comment here would help explain the intent. If it's not strictly required, dropping it would let container policy signature verification apply to pinned artifacts too.

Same applies to `internal/criocli/artifact.go:64`.

server/server.go
```diff
@@ -653,6 +665,48 @@ func (s *Server) startReloadWatcher(ctx context.Context) {
...
+	if len(refs) == 0 {
+		return
+	}
+
+	s.pinnedArtifactsPullMu.Lock()
+	defer s.pinnedArtifactsPullMu.Unlock()
+
+	log.Infof(ctx, "Pre-pulling %d pinned artifact(s)", len(refs))
+
+	for _, refStr := range refs {
```
saschagrunert, line 699: No deduplication: if a user lists the same reference twice in `pinned_artifacts`, it gets pulled twice (each going through `EnsureNotContainerImage` + the full pull). Consider deduplicating `refs` before iterating:
```go
seen := make(map[string]struct{}, len(refs))
for _, refStr := range refs {
    if _, ok := seen[refStr]; ok {
        continue
    }
    seen[refStr] = struct{}{}
    // ...
}
```

server/container_events_test.go
```diff
@@ -64,6 +64,8 @@ var _ = t.Describe("ContainerEvents", func() {
 			}
 
 			recv := func(ces types.RuntimeService_GetContainerEventsServer) {
+				defer GinkgoRecover()
```
saschagrunert: Correct fix, but unrelated to pinned artifacts. Consider splitting this into a separate commit to keep the diff focused.

test/pinned_artifacts.bats
```diff
@@ -0,0 +1,243 @@
...
+@test "pinned_artifacts pre-pulls artifact at startup and logs success" {
+	cat << EOF > "$CRIO_CONFIG_DIR/99-pinned.conf"
+[crio.image]
+pinned_artifacts = ["$ARTIFACT_IMAGE"]
+EOF
+	start_crio
+	wait_for_log "Pinned artifact pre-pulled: $ARTIFACT_IMAGE"
+
+	# Artifact must be queryable via crictl after pre-pull.
+	crictl images | grep -q "crio/artifact"
```
saschagrunert: Does `crictl images` (CRI `ListImages`) actually surface OCI artifacts from the artifact store? Artifacts and container images live in separate stores. If this works, it might be worth a brief comment explaining why artifacts appear in `crictl images` output. If it doesn't work reliably, the `wait_for_log` assertion above already covers the success case.

test/pinned_artifacts.bats
```diff
@@ -0,0 +1,243 @@
...
+	printf '[crio.image]\npinned_artifacts = ["%s"]\n' "$ARTIFACT_IMAGE" \
+		> "$CRIO_CONFIG_DIR/99-pinned.conf"
+	start_crio_no_setup
+	wait_for_log "Pre-pulling 1 pinned artifact"
+
+	printf '[crio.image]\npinned_artifacts = []\n' \
+		> "$CRIO_CONFIG_DIR/99-pinned.conf"
+	reload_crio
+
+	wait_for_log '"set config pinned_artifacts to \\"\[\]\\""'
```
saschagrunert: This nested escape pattern is fragile. If the log format changes slightly (quoting style, spacing), this breaks silently. Consider matching a simpler substring, or add a comment documenting the expected log format so future maintainers know what to update.

bitoku (changes requested):
Thanks for the PR!

However, with this current design, CRI-O's config would need to be updated whenever the pre-pulled image becomes stale, which isn't a great user experience. I'm also unsure what specific gap this feature fills; I'd argue that relying on the default artifact location is sufficient.

A couple of additional thoughts:

1. Since images and artifacts can already be downloaded using crictl, there's no need for a dedicated command within the CRI-O binary.

2. The name pinned-artifacts is a bit confusing, as it behaves differently from pinned-images.

server/server.go
xw19: okay

xw19:
> Thanks for the PR!
> 
> However, with this current design, CRI-O's config would need to be updated whenever the pre-pulled image becomes stale, which isn't a great user experience. I'm also unsure what specific gap this feature fills; I'd argue that relying on the default artifact location is sufficient.
> 
> A couple of additional thoughts:
> 
>     1. Since images and artifacts can already be downloaded using crictl, there's no need for a dedicated command within the CRI-O binary.
> 
>     2. The name pinned-artifacts is a bit confusing, as it behaves differently from pinned-images.

@bitoku this is as per the design mentioned in the issue. I am happy to address any changes. @saschagrunert can you weigh in?

xw19:
@saschagrunert can you please again take a look?

saschagrunert:
Thanks for the feedback @bitoku.

The core value here is declarative, config-driven artifact pre-pulling at startup and on SIGHUP. `crictl` can't do this today since it would need CRI changes to support artifact pulls. For large AI/ML models on autoscaling nodes, having this in CRI-O's lifecycle is the gap.

That said, I agree on two points:

1. **Naming**: `pinned_artifacts` implies GC protection (like `pinned_images`), but this does pre-pulling. Let's rename to `prepull_artifacts`.
2. **CLI subcommand**: Let's drop `crio artifact pull` to keep the scope focused on the config-driven pre-pull.

Staleness follows the same pattern as `pinned_images` and SIGHUP reload handles updates without restarts.

@bitoku would you be fine with this approach?

bitoku:
Thank you @saschagrunert 

> For large AI/ML models on autoscaling nodes, having this in CRI-O's lifecycle is the gap.

The question I have is why it should be config-driven and pre-pull strategy at startup time.
Why don't users bake pre-pulled image into the OS image? Pulling them consumes network bandwidth more.

bitoku:
I mean it doesn't have to bake. they can attach external storage or something.

saschagrunert:
Thanks for the work on this, @xw19!

After digging deeper into the discussion with @bitoku and reviewing the existing code, we've concluded that this PR isn't the right approach for now:

1. **GC protection is already covered**: `pinned_images` already applies to OCI artifacts (not just container images), so a separate `pinned_artifacts` config for pinning is redundant. We'll update the docs to make this clearer.

2. **Pre-pulling needs more design work**: The pre-pull at startup approach has trade-offs (network bandwidth on every node boot, staleness management) that need broader agreement. For the autoscaling use case, there may be better paths forward, such as leveraging additional artifact stores with pre-populated storage, or solving this at the orchestration layer.

Closing this PR and the tracking issue for now. If we revisit pre-pulling in the future, we can open a fresh proposal with a narrower scope focused purely on that.

Thanks again for the contribution!

PR #9940, opened 2026-05-09
utils: prefix fully numeric names in generated passwd/group entries

### LINKED ISSUE #9432: Stop using fully numeric usernames
### What happened?

Many systems, including those derived from Fedora, [do not support fully numeric usernames](https://access.redhat.com/solutions/3103631). An example of one of these limitations can be seen in the problem at https://github.com/shadow-maint/shadow/issues/1339, but I fear it will not be the only one.
As @tolusha points out, CRI-O automatically adds these users, so I strongly recommend changing this logic.

As a suggestion, you can add a suffix to the number to avoid any problems.

NA

_No response_

NA

### OS version

NA

NA

kimjune01:
/kind bug

`GeneratePasswd` and `GenerateGroup` write fully numeric usernames and group names (e.g. `1234`) into `/etc/passwd` and `/etc/group` when a container runs with a numeric `runAsUser` whose UID doesn't exist in the image. shadow-utils on Fedora/RHEL rejects these names, causing container creation failures.

This adds `safeAccountName()` which detects all-digit names and prefixes them with `user` (e.g. `1234` becomes `user1234`). Applied to both `GeneratePasswd` (username field) and `GenerateGroup` (group name field).

Fixes #9432

A previous attempt at this fix (#9446) was auto-closed for inactivity. This takes the same approach (prefix with "user") but also covers the group name path in `GenerateGroup`, which the earlier PR missed.

```
ginkgo run --focus "numeric username|numeric group|non-numeric|mixed alphanumeric" -v utils/
```

```release-note
Fully numeric usernames and group names generated by CRI-O are now prefixed with "user" (e.g. "1234" becomes "user1234") to avoid rejection by shadow-utils on Fedora/RHEL systems.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of numeric usernames and group names by prefixing them appropriately for valid formatting.

* **Tests**
  * Added test cases to validate numeric account name behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cri-o/cri-o/pull/9940)

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -203,6 +203,35 @@ func GetUserInfo(rootfs, userName string) (uid, gid uint32, additionalGids []uin
 	return uid, gid, additionalGids, nil
 }
 
+// isAllDigits reports whether s is non-empty and consists entirely of
+// ASCII digits. This is the same class of names that shadow-utils on
+// Fedora/RHEL rejects as usernames and group names.
+func isAllDigits(s string) bool {
+	if s == "" {
+		return false
+	}
+
+	for _, r := range s {
+		if r < '0' || r > '9' {
+			return false
+		}
+	}
+
+	return true
+}
+
+// safeAccountName returns a name safe for /etc/passwd or /etc/group.
+// Fully numeric names (e.g. "1234") are rejected by shadow-utils on
+// Fedora/RHEL, so we add a "user" prefix (e.g. "user1234").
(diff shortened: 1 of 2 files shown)


ikerexxe:
This looks like a sensible solution to the fully numeric usernames problem. Please do give it a look

PR #9943, opened 2026-05-10
Invoke NRI StopPodSandbox hook before tearing down network namespace

### LINKED ISSUE #9905: NRI StopSandbox hook and network namesapces
Since network functionality start to consume more and more the NRI plugin API, we start to find some inconsistencies on the hooks implementation https://github.com/kubernetes-sigs/dranet/issues/137

In this case specifically, the network nri plugin expects the network namespace to be available during the `StopPodSandbox` hook https://github.com/kubernetes-sigs/dranet/issues/137#issuecomment-4213326829

In order to use the NRI repo as the source of truth , we documented the expected contracts so plugin developers and implementations offers the same functionality https://github.com/containerd/nri/pull/286

This issue is just to request CRIO to move the NRI hook for StopSandbox before the network namespaces are deleted

https://github.com/cri-o/cri-o/blob/ff94ddfaf9f1d28cb36d370509f87ba0ae1250f7/server/sandbox_stop_linux.go#L100-L102

amritansh1502:
NRI StopPodSandbox hook was being called after the network namespace was already torn down. NRI network plugins (e.g. dranet) need the netns to be alive during this hook to perform cleanup. This moves the hook to run before networkStop.

Fixes #9905

```release-note
Fixed NRI StopPodSandbox hook being called after network namespace deletion.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted sandbox shutdown to always perform network cleanup even for already-stopped sandboxes.
  * Reordered shutdown steps so external runtime teardown runs before network namespace cleanup for consistent shutdown behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cri-o/cri-o/pull/9943)

### DIFF
--- a/server/sandbox_stop_linux.go
+++ b/server/sandbox_stop_linux.go
@@ -34,15 +34,26 @@ func (s *Server) stopPodSandbox(ctx context.Context, sb *sandbox.Sandbox) error 
 		}
 	}
 
-	// Clean up sandbox networking and close its network namespace.
-	if err := s.networkStop(ctx, sb); err != nil {
-		return err
-	}
-
 	if sb.Stopped() {
+		// Network cleanup may still be needed even for already-stopped sandboxes.
+		if err := s.networkStop(ctx, sb); err != nil {
+			return fmt.Errorf("failed to stop network for pod sandbox %s: %w", sb.ID(), err)
+		}
+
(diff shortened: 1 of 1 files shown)


bitoku:
/hold
I'd rather wait for the implementation until the https://github.com/containerd/nri/pull/286 is merged if there's no reason to rush, just in case.

PR #9954, opened 2026-05-12
fix(oci): fix goroutine leak in AttachContainer

Reamer:
Please note that this fix was implemented with the help of AI. The AI helped me understand the problem (by analyzing the goroutines in the running program) and write clean comments.

#### What type of PR is this?

/kind bug

Fixes a goroutine leak in `AttachContainer` where goroutines accumulated
indefinitely when containers were stopped with SIGKILL after a SIGTERM
timeout.

Two issues caused the leak:

1. Unbuffered channels blocked forever when early-return paths
   (`leave-stdin-open`, `DetachError`) exited without draining
   `receiveStdout`.

2. `defer conn.Close()` alone did not unblock a pending `Read()` in the
   `receiveStdout` goroutine since conmon keeps the attach socket open
   as long as the container runs.

Fix uses buffered channels, `context.WithCancel` and explicitly closes
`conn` on `ctx.Done()` to unblock any pending `Read()`.

The leak was introduced between v1.29 and v1.30 when the `outputStream`
guard was removed to fix conmon blocking on large stdin data. This PR
fixes both issues simultaneously.

Verified in production with 200+ GitLab CI jobs on OpenShift/CRI-O 1.34.4.

None

`conn.Close()` is called via both `defer` and the `ctx.Done()` goroutine.
`net.Conn.Close()` is idempotent so this is safe.

Backports to release-1.30 through release-1.37 are recommended.

No

```release-note
Fixed a goroutine leak in AttachContainer caused by SIGKILL after
SIGTERM timeout. Over time this could lead to increased memory usage
and resource exhaustion.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container attach reliability by adding cancellable handling to attachment I/O, ensuring prompt cleanup when operations are cancelled.
  * Prevented potential hangs by unblocking pending reads and making stdin/stdout termination cooperative with cancellation, improving overall responsiveness during detach/timeout scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cri-o/cri-o/pull/9954?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1382,6 +1382,10 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
 		return nil
 	}
 
+	// cancel ensures all goroutines are cleaned up when the function returns
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	controlPath := filepath.Join(c.BundlePath(), "ctl")
 
 	controlFile, err := os.OpenFile(controlPath, os.O_WRONLY, 0)
@@ -1409,15 +1413,31 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
 
 	defer conn.Close()
 
-	receiveStdout := make(chan error)
-
+	// Close conn when ctx is done to unblock any pending Read()
+	// in redirectResponseToOutputStreams(). This is necessary because
+	// conmon keeps the other side of the socket open as long as the
+	// container is running, so conn.Close() via defer alone is not
+	// sufficient to unblock the Read() in the goroutine below.
 	go func() {
-		receiveStdout <- redirectResponseToOutputStreams(outputStream, errorStream, conn)
-
-		close(receiveStdout)
+		<-ctx.Done()
+		conn.Close()
 	}()
 
-	stdinDone := make(chan error)
+	// Buffered channels of size 1 prevent goroutine leaks when the receiver
+	// exits early. Context cancellation provides an additional safety net.
+	receiveStdout := make(chan error, 1)
+
+	go func() {
+
+		streamErr := redirectResponseToOutputStreams(outputStream, errorStream, conn)
+		select {
+		case receiveStdout <- streamErr:
+		case <-ctx.Done():
+			log.Debugf(ctx, "receiveStdout goroutine cancelled: %v", ctx.Err())
+		}
+	}()
+
+	stdinDone := make(chan error, 1)
 
 	go func() {
 		var err, closeErr error
@@ -1426,17 +1446,21 @@ func (r *runtimeOCI) AttachContainer(ctx context.Context, c *Container, inputStr
 			closeErr = conn.CloseWrite()
 		}
 
+		var result error
 		switch {
 		case err != nil:
-			stdinDone <- err
+			result = err
 		case closeErr != nil:
-			stdinDone <- closeErr
+			result = closeErr
 		default:
 			// neither CopyDetachable nor CloseWrite returned error
-			stdinDone <- nil
 		}
 
-		close(stdinDone)
+		select {
+		case stdinDone <- result:
+		case <-ctx.Done():
+			log.Debugf(ctx, "stdinDone goroutine cancelled: %v", ctx.Err())
+		}
 	}()
 
 	select {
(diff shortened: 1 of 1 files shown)


Reamer:
Is there anything I can do to get feedback?

bitoku:
Sorry @Reamer.
I didn't have bandwidth. I'll review it this week.

harche:
Thanks for this fix, and sorry it sat unreviewed for so long. I hit the same leak independently while running the Go 1.26 goroutine leak profiler across the codebase (#10314), and I'd like this PR to be the one that lands for attach since you found it first and have production validation behind it. I'll hold the attach commit in #10314 and rebase onto this once it lands.

A few suggestions after comparing the two, so this stays minimal and backports cleanly to release-1.30 through 1.37:

1. **The buffered channels are the whole fix.** On `main`, both result channels are unbuffered ([`receiveStdout`](https://github.com/cri-o/cri-o/blob/39d524aec4d15760d8ca912dc43d36901d1d7663/internal/oci/runtime_oci.go#L1410), [`stdinDone`](https://github.com/cri-o/cri-o/blob/39d524aec4d15760d8ca912dc43d36901d1d7663/internal/oci/runtime_oci.go#L1418)), and the function can return from the `stdinDone` arm without ever draining `receiveStdout` ([L1449-L1457](https://github.com/cri-o/cri-o/blob/39d524aec4d15760d8ca912dc43d36901d1d7663/internal/oci/runtime_oci.go#L1449-L1457)). When it does, the deferred `conn.Close()` wakes the stdout goroutine's `Read`, and that goroutine then blocks forever on the [send](https://github.com/cri-o/cri-o/blob/39d524aec4d15760d8ca912dc43d36901d1d7663/internal/oci/runtime_oci.go#L1413). Your two `make(chan error, 1)` lines ([L1428](https://github.com/cri-o/cri-o/blob/1cb7698c65bf1e260573e625545f9ccdbd642bfe/internal/oci/runtime_oci.go#L1428), [L1440](https://github.com/cri-o/cri-o/blob/1cb7698c65bf1e260573e625545f9ccdbd642bfe/internal/oci/runtime_oci.go#L1440)) resolve that on their own.

2. **The goroutine that calls `conn.Close()` on `ctx.Done()` ([L1416-L1424](https://github.com/cri-o/cri-o/blob/1cb7698c65bf1e260573e625545f9ccdbd642bfe/internal/oci/runtime_oci.go#L1416-L1424)) can go.** The comment there says the deferred close alone is not sufficient to unblock the pending `Read`, but unblocking blocked reads is part of the `net.Conn` contract: ["Any blocked Read or Write operations will be unblocked and return errors."](https://github.com/golang/go/blob/go1.27.0/src/net/net.go#L135-L136) The existing [`defer conn.Close()`](https://github.com/cri-o/cri-o/blob/1cb7698c65bf1e260573e625545f9ccdbd642bfe/internal/oci/runtime_oci.go#L1414) already does this. Worth rewording the PR description's second bullet too, so backport reviewers get the right mental model.

3. **The `ctx.Done()` arms inside the two worker `select`s ([L1433-L1437](https://github.com/cri-o/cri-o/blob/1cb7698c65bf1e260573e625545f9ccdbd642bfe/internal/oci/runtime_oci.go#L1433-L1437), [L1459-L1463](https://github.com/cri-o/cri-o/blob/1cb7698c65bf1e260573e625545f9ccdbd642bfe/internal/oci/runtime_oci.go#L1459-L1463)) are unreachable.** With a capacity-1 buffer and a single send per goroutine, the send case is always ready, so the `ctx.Done()` case never fires. Dropping them keeps the diff to the two `make` lines.

4. **The `ctx.Done()` returns in the outer `select` ([L1487-L1491](https://github.com/cri-o/cri-o/blob/1cb7698c65bf1e260573e625545f9ccdbd642bfe/internal/oci/runtime_oci.go#L1487-L1491)) are optional.** The server passes its lifetime context here, not a per-request one ([`server/container_attach.go#L66`](https://github.com/cri-o/cri-o/blob/39d524aec4d15760d8ca912dc43d36901d1d7663/server/container_attach.go#L66)), so those arms only fire on daemon shutdown. Fine to keep if you want that behavior, but it's independent of the leak and I'd leave it out of a backport candidate.

5. **Regression test.** I have a Linux Ginkgo test that fails on the unfixed code for both completion orders (stdin first, stdout first) against a real `unixpacket` server, and passes with only the buffered channels applied. Feel free to pull it in as-is: [`internal/oci/runtime_oci_attach_leak_linux_test.go`](https://github.com/harche/cri-o/blob/attach-leak-test/internal/oci/runtime_oci_attach_leak_linux_test.go) (commit [`e44d1aec9`](https://github.com/harche/cri-o/commit/e44d1aec9e2f0caca11ca8b2a4e23b67fa01e369) on the `attach-leak-test` branch of my fork). That would also address the Codecov patch-coverage warning.

Happy to review as soon as you've had a chance to update.

/ok-to-test

PR #9955, opened 2026-05-13
bump oc/selinux to v1.15.1, use SetProcessKind and ReserveLabelV2

kolyshkin:
~~This is a draft until oc/selinux v1.15.0 is released; mostly to test https://github.com/opencontainers/selinux/pull/271.~~

#### What type of PR is this?

/kind dependency-change

1. Change ReserveLabel to ReserveLabelV2. The former is deprecated; the latter can return ErrMCSAlreadyReserved for a duplicate label. Looking at the code, the label should be unique so I'm not ignoring the error, but **it may result in a regression** so needs a check from someone who understands this code.

2. Change KVMLabel/InitLabel to `SetProcessKind` which is much more straightforward.

None

 - not sure why server/sandbox_run_freebsd.go contains selinux stuff; selinux is Linux specific;
 - server/selinux.go adds go:fix directives which are not (yet) recognized by the gocheckcompilerdirectives linter, so add a temporary exception).

```release-note
NONE
```

## Summary by CodeRabbit

* **Dependencies**
  * Bumped the SELinux library to v1.15.1.

* **Refactor**
  * Updated SELinux process-label generation to use the standardized `SetProcessKind` APIs for init/systemd and kernel-separated KVM paths.
  * Deprecated legacy label helper behavior in favor of `SetProcessKind`.

* **Bug Fixes**
  * Improved SELinux label reservation by switching to the newer reservation flow with clearer error reporting.

* **Chores**
  * Updated linting rules to recognize `//go:fix` directives and enable the related compiler-directive check.


kolyshkin:
I see some tests are failing with
>  unable to reserve selinux process label "system_u:system_r:container_runtime_t:s0": MCS label already exists

This is not exactly what I warned about. This label does not even have MCS labels (that look like `s0:c1,c2`). Maybe we need to fix this in `selinux` (either ignore the request if there's no MCS, or return a different error (missing MCS), but this needs to be fixed here as well (if there's no MCS, there's no need to reserve).

Cc @rhatdan

kolyshkin:
> I see some tests are failing with
> 
> > unable to reserve selinux process label "system_u:system_r:container_runtime_t:s0": MCS label already exists
> 
> This is not exactly what I warned about. This label does not even have MCS labels (that look like `s0:c1,c2`). Maybe we need to fix this in `selinux` (either ignore the request if there's no MCS, or return a different error (missing MCS), but this needs to be fixed here as well (if there's no MCS, there's no need to reserve).
> 
> Cc @rhatdan

Opened https://github.com/opencontainers/selinux/pull/272; testing it here.

kolyshkin:
> Opened [opencontainers/selinux#272](https://github.com/opencontainers/selinux/pull/272); testing it here.

v1.15.1 released; this PR is ready for review/merge.

kolyshkin:
Frankly I can't figure out if CI failures are legit or not.

internal/lib/container_server.go
```diff
@@ -446,7 +446,9 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
 		sb.SetStopped(ctx, true)
 	}
 
-	selinux.ReserveLabel(processLabel)
+	if err := selinux.ReserveLabelV2(processLabel); err != nil { // FIXME: may need to ignore selinux.ErrMCSAlreadyExists.
```
bitoku: Can we remove this FIXME since it's already addressed?

kolyshkin: I am not 100% sure about it, thus the FIXME. But I'm like 90% sure so let's remove it.

DONE

bitoku:
/lgtm

I'm fixing the CI ...

kolyshkin:
rebased

kolyshkin:
> New changes are detected. LGTM label has been removed.

I haven't changed anything, merely did a rebase :(

@bitoku ptal 🙏🏻

bitoku:
`not ok 93 cri-o continues to monitor when cri-o restarts`
This test failure looks consistent, we need to look into it.

bitoku:
@cri-o/cri-o-maintainers PTAL

internal/lib/container_server.go
```diff
@@ -446,7 +446,9 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
 		sb.SetStopped(ctx, true)
 	}
 
-	selinux.ReserveLabel(processLabel)
+	if err := selinux.ReserveLabelV2(processLabel); err != nil {
```
saschagrunert, line 457: `LoadSandbox` is called during CRI-O restart to reload existing sandboxes from disk. `ReserveLabelV2` returns `ErrMCSAlreadyExists` when the same MCS level is reserved twice, and this code makes that a hard error.

The `ReserveLabelV2` godoc itself says callers "may safely ignore `ErrMCSAlreadyExists`", and the vendored `label.InitLabels` does exactly that (with a comment explaining the rationale). Making it fatal here means an edge case (e.g. manual label manipulation, or a past bug that assigned duplicate MCS) could cause CRI-O to mark a previously healthy sandbox for deletion on restart.

Consider:
```suggestion
	if err := selinux.ReserveLabelV2(processLabel); err != nil && !errors.Is(err, selinux.ErrMCSAlreadyExists) {
```
This would also need adding `"errors"` to the imports.

server/container_create.go
```diff
@@ -1213,7 +1214,7 @@ func (s *Server) configureSELinuxLabels(ctr container.Container, sb *sandbox.San
 	// the container and the process label is unset, the init selinux label is required
 	// to run the container.
 	if ctr.WillRunSystemd() && processLabel == "" {
-		processLabel, err = InitLabel(processLabel)
+		processLabel, err = selinux.SetProcessKind(processLabel, selinux.ProcessKindInit)
```
saschagrunert, line 1217: Pre-existing issue, not from this PR: this block is a no-op. The condition on line 1216 requires `processLabel == ""`, but `SetProcessKind("", ...)` returns `("", nil)` (same as the old `InitLabel("")`). If the intent is to assign an init label when none exists, this would need to allocate one (e.g. via `selinux.InitContainerLabel()`). As-is, the comment on lines 1214-1215 is misleading.

.golangci.yml
```diff
@@ -152,6 +152,12 @@ linters:
     presets:
       - comments
       - std-error-handling
+    rules:
+      # TODO: remove once https://github.com/leighmcculloch/gocheckcompilerdirectives/issues/7 is fixed.
```
saschagrunert: The upstream issue was fixed in gocheckcompilerdirectives v1.4.0 (closed 2026-05-13). golangci-lint v2.10.1 (and v2.12.2) still bundles v1.3.0, so the exclusion is still needed. Just noting the fix exists upstream and will land once golangci-lint picks up the new version.

saschagrunert (commented):
Overall the migration looks correct. The `SetProcessKind` replacement is cleaner than the old allocate-and-release dance, and the vendor bump is solid. Three inline notes below, the first one being actionable.

.golangci.yml
kolyshkin: Right; the bump was merged in https://github.com/golangci/golangci-lint/pull/6584

internal/lib/container_server.go
kolyshkin: Fixed (I added a warning in this case, so we know if this ever happens).

server/container_create.go
kolyshkin: Indeed, InitLabel returns an empty string if called with an empty string as an argument.
It looks like `selinux.InitContainerLabel` should be used here instead.

This code was initially added by commit 4e2f7dbed and then moved to this place by commit 922d3edc8.

@bitoku can you please take a look?

kolyshkin:
Rebased; addressed review comments from @saschagrunert; PTAL

server/selinux.go
```diff
@@ -5,43 +5,19 @@ import (
 )
 
 // KVMLabel returns labels for running kvm isolated containers.
+//
+// Deprecated: use [selinux.SetProcessKind].
+//
+//go:fix inline
 func KVMLabel(cLabel string) (string, error) {
```
saschagrunert: All three call sites (`container_create.go`, `sandbox_run_linux.go`, `sandbox_run_freebsd.go`) now call `selinux.SetProcessKind` directly. `KVMLabel` and `InitLabel` have zero remaining callers. Since the `server` package isn't a public API, these could be deleted outright instead of deprecated. The `//go:fix inline` directives won't help anyone if nobody imports these functions.

internal/lib/container_server.go
```diff
@@ -454,7 +454,12 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
...
+		if !errors.Is(err, selinux.ErrMCSAlreadyExists) {
+			return sb, fmt.Errorf("unable to reserve selinux process label %q: %w", processLabel, err)
```
saschagrunert, line 459: Nit: the message reads like the reservation is happening, not that it's a duplicate. Since this line is only reached when `ErrMCSAlreadyExists`, something like `"SELinux label %s already reserved, ignoring: %v"` would be clearer when reading restart logs.

saschagrunert (commented):
Two inline comments.

internal/lib/container_server.go
kolyshkin: fixed

server/selinux.go
kolyshkin: Sure, removed (my idea was since it's not an internal package it may have users outside of this repo, and although I haven't found any, I didn't want to break anything).

kolyshkin:
@saschagrunert addressed your review comments; PTAL.

saschagrunert (approved):

saschagrunert:
/lgtm cancel

lint CI is not happy:

```
  Error: internal/lib/container_server.go:461:3: missing whitespace above this line (invalid statement above expr) (wsl_v5)
  		log.Warnf(ctx, "SELinux label %s already reserved, ignoring: %v", processLabel, err)
  		^
  1 issues:
  * wsl_v5: 1
  *
```

internal/lib/container_server.go
```diff
@@ -454,7 +454,12 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
...
+		}
+		log.Warnf(ctx, "SELinux label %s already reserved, ignoring: %v", processLabel, err)
```
saschagrunert: nit: 
```suggestion
		log.Warnf(ctx, "SELinux label %q already reserved, ignoring: %v", processLabel, err)
```

kolyshkin: fixed

kolyshkin:
review nit addressed; rebased; PTAL @saschagrunert @bitoku

saschagrunert (approved):

bitoku:
/override ci/prow/ci-cgroupv2-e2e-crun ci/prow/ci-crun-e2e ci/prow/ci-e2e ci/prow/ci-e2e-conmonrs ci/prow/ci-rhel-e2e

They are not related to the change.

PR #9962, opened 2026-05-18
bump containerd/api, conmon-rs, podman common and storage deps

### LINKED ISSUE #9934: Bump held-back dependencies from #9929
PR #9929 bumped 30 Go module dependencies but had to hold back several due to breaking API changes. These need dedicated follow-up work:

**containerd/containerd/api v1.10.0 -> v1.11.0**
v1.11.0 renames all ttrpc-generated service types (`TaskService` -> `TTRPCTaskService`, `NewTaskClient` -> `NewTTRPCTaskClient`, `EventsService` -> `TTRPCEventsService`, etc.). Affects `internal/oci/runtime_vm.go` and vendored `containerd/containerd/pkg/ttrpcutil/client.go`. containerd v1.7.x only requires API v1.8.0; the v1.11.0 types come from the 2.x line.

**go.podman.io/common v0.66.2 -> v0.67.0**
Removes `ArtifactStore`, `ArtifactReference`, `ArtifactStoreReference` types and their constructors from `pkg/libartifact`. Removes `Artifact.Digest` field (replaced with `GetDigest()` method). Requires rewriting `internal/ociartifact/` (store.go, artifact.go, libartifact_store.go).

**go.podman.io/storage v1.61.1 -> v1.62.0**
Uses `securejoin.OpenInRoot` and `securejoin.Reopen` at the top-level import, but these were removed in `filepath-securejoin` v0.6.0 (moved to `pathrs-lite` sub-package). Other deps (runc, conmon-rs) force securejoin to v0.6.1.

**containers/conmon-rs v0.7.3 -> v0.8.0**
Transitively requires `go.podman.io/common` v0.67.0, forcing the breaking version.

Both `containers/common` and `containers/storage` have been moved to a monorepo, so the securejoin and libartifact fixes may land there.

vishu-25:
Fixes #9934

This PR upgrades the four dependencies that were held back from #9929 due to breaking API changes.

### `github.com/containerd/containerd/api` v1.10.0 → v1.11.0

The generated ttrpc service types were renamed with a `TTRPC` prefix:
- `TaskService` → `TTRPCTaskService`
- `NewTaskClient` → `NewTTRPCTaskClient`
- `EventsService` → `TTRPCEventsService`
- `NewEventsClient` → `NewTTRPCEventsClient`

Updated `internal/oci/runtime_vm.go` to use the new names. The vendored `containerd/containerd` v1.7.31 (a separate module) also references these types in `pkg/ttrpcutil/client.go` and `runtime/v2/shim/shim.go` — those files are patched in-vendor since `go mod vendor` only updates the `containerd/api` module.

### `github.com/containers/conmon-rs` v0.7.3 → v0.8.0

No source changes required; vendor regeneration is sufficient.

### `go.podman.io/common` v0.66.x → v0.67.0

The `ArtifactStore` was moved to the `pkg/libartifact/store` sub-package, and the `ArtifactReference`/`ArtifactStoreReference` wrapper types were removed — all store methods now accept plain strings. Updated `internal/ociartifact/` and the generated mock in `test/mocks/ociartifact/`.

The `Artifact.Digest` field was also removed; the replacement is a `GetDigest() (*digest.Digest, error)` method that computes the digest from `json.Marshal(manifest)`. Updated `CanonicalName()` and the new `Digest()` helper accordingly.

### `go.podman.io/storage` v1.61.x → v1.62.0

`storage_linux.go` and `userns.go` call `securejoin.OpenInRoot`, `securejoin.OpenatInRoot`, and `securejoin.Reopen`, which were moved to the `pathrs-lite` sub-package in `cyphar/filepath-securejoin` v0.6+. CRI-O's other deps force securejoin to v0.6.1, so the two vendored files are patched to import from `github.com/cyphar/filepath-securejoin/pathrs-lite`.

## Testing

- `go build ./internal/oci/...` (cross-compiled for linux/amd64) ✓
- `go build ./internal/ociartifact/...` ✓
- `go test ./internal/ociartifact/...` — 23/23 tests pass ✓

## RHEL 10.2 compatibility

Verified that podman v5.8.0 (shipping in RHEL 10.2 Beta) uses exactly `go.podman.io/common v0.67.0` and `go.podman.io/storage v1.62.0`, so there is no storage backend incompatibility risk.

## Summary by CodeRabbit

* **Chores**
  * Updated dependencies: containerd API to v1.11.0, conmon-rs to v0.8.0, and Podman-related libraries to newer versions.

* **Refactor**
  * Simplified OCI artifact storage operations by streamlining internal interfaces and removing intermediate object construction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cri-o/cri-o/pull/9962?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


internal/ociartifact/artifact.go
```diff
@@ -64,12 +64,24 @@ func (a *Artifact) Reference() string {
...
 func (a *Artifact) Digest() digest.Digest {
-	return a.Artifact.Digest
+	dgst, err := a.Artifact.GetDigest()
+	if err != nil {
+		log.Warnf(context.Background(), "Failed to get artifact digest: %v", err)
+		return ""
+	}
+
+	return *dgst
 }
```
saschagrunert, line 85: `Digest()` and `CanonicalName()` now silently swallow `GetDigest()` errors and return degraded values (empty digest / `<unknown>`). The callers assume these always succeed:

- `CRIImage()` (line 100) uses `a.Digest().Encoded()` as the CRI image `Id`. An empty digest gives kubelet an empty image ID.
- `store.go:133` takes a pointer to the returned (empty) digest and hands it back as a successful `Pull` result.
- `datastore/store.go:233` uses `a.Digest().Encoded()` in `strings.HasPrefix` for digest-based lookup, which silently breaks for empty digests.
- `CanonicalName()` feeds into `CRIImage().RepoDigests` and is used for artifact resolution.

`GetDigest()` only fails if the manifest is nil or cannot be marshaled, so this should not happen for artifacts that were successfully pulled/inspected. But if it does happen, returning a bogus value is worse than returning an error.

Consider propagating the error, e.g. compute the digest once in `newArtifact()` (right after the artifact is created from a known-good source) and store it as a field, or change the signatures to return errors.

internal/ociartifact/store.go
```diff
@@ -126,28 +126,20 @@ func (s *Store) Pull(
...
-					// Force-pinned: additional stores are read-only and not subject to GC.
-					dgst := s.newArtifact(art, add.path, true).Digest()
-
-					return &dgst, nil
-				}
+		for _, add := range s.additionalStores {
+			if art, err := add.store.Inspect(ctx, strRef); err == nil {
+				log.Infof(ctx, "Artifact %s already exists in additional store %s, skipping pull", strRef, add.path)
+				// Force-pinned: additional stores are read-only and not subject to GC.
+				dgst := s.newArtifact(art, add.path, true).Digest()
```
saschagrunert, line 133: If `Digest()` hits the error path and returns `""`, this line takes a pointer to that empty digest and returns it as a successful pull result. The caller has no way to distinguish this from a real digest.

This could be avoided by computing the digest here with `GetDigest()` directly so the error can be checked and propagated.

internal/ociartifact/store.go
```diff
@@ -126,28 +126,20 @@ func (s *Store) Pull(
...
 
 	log.Infof(ctx, "Pulling OCI artifact %s", strRef)
 
-	artRef, err := libart.NewArtifactReference(strRef)
-	if err != nil {
-		return nil, fmt.Errorf("invalid reference: %w", err)
-	}
-
-	dgst, err := s.libartifactStore.Pull(ctx, artRef, *opts)
+	dgst, err := s.libartifactStore.Pull(ctx, strRef, *opts)
```
saschagrunert, line 142: Nit: the old code validated the reference via `NewArtifactReference(strRef)` before calling `Pull`. Now validation is deferred entirely to the store. Worth confirming that the store does reject invalid references with a reasonable error, otherwise bad input that used to fail early could now produce a less clear error or unexpected behavior.

adilGhaffarDev:
@saschagrunert can we bump go.podman.io/common v0.66.2 -> v0.68.0

adilGhaffarDev:
> @saschagrunert can we bump go.podman.io/common v0.66.2 -> v0.68.0

@vishu-25

saschagrunert:
@vishu-25 do you mind to rebase to fix the CI?

adilGhaffarDev:
> Updated internal/oci/runtime_vm.go to use the new names. The vendored containerd/containerd v1.7.31 (a separate module) also references these types in pkg/ttrpcutil/client.go and runtime/v2/shim/shim.go — those files are patched in-vendor since go mod vendor only updates the containerd/api module.

I dont think we should edit vendors manually. We should work on migrating to `containerd v2` to support `github.com/containerd/containerd/api v1.11.0`. and this pr is not doing that.

adilGhaffarDev:
> I dont think we should edit vendors manually. We should work on migrating to `containerd v2` to support `github.com/containerd/containerd/api v1.11.0`. and this pr is not doing that.

verify vendor test is also failing because of this.

PR #9968, opened 2026-05-20
OCPBUGS-84718: server: fix unnecessary storage wipe after unclean shutdown

Chandan9112:
### Summary

After an unclean shutdown, stale overlay mounts cause `store.Check()` to report false layer errors. `store.Repair()` then fails with "device or resource busy" and CRI-O wipes the entire primary image store at `/var/lib/containers/storage`. This is destructive and unnecessary — the storage is not actually corrupted.

### Changes

1. Clean stale overlay mounts (via `cmount.RecursiveUnmount`) before `store.Check()` runs, preventing false layer errors.
2. If `Repair()` still fails with only transient "device or resource busy" errors, skip the wipe instead of nuking the store.
3. Add integration tests for stale mount and real corruption scenarios.

### Root Cause Evidence

Debug logging after unclean shutdown on CRI-O v1.31:
- `grep -c "device or resource busy"` → **19** (all errors are stale mounts)
- `grep -c "ted layer"` → **0** (no actual corruption)
- `grep -c "Wiping storage"` → **1** (unnecessary wipe triggered)

### Testing

- Reproduced on RHEL 9.4 + MicroShift 4.18 + CRI-O v1.31.13
- Verified fix prevents the wipe (0 wipes, 0 busy errors after stale mount cleanup)

```
level=warning msg="Checking storage directory /var/lib/containers/storage for errors because of unclean shutdown"
level=warning msg="Attempting to repair storage directory /var/lib/containers/storage because of unclean shutdown"

$ grep -c "Wiping" /tmp/fix.log
0
$ grep -c "device or resource busy" /tmp/fix.log
0

```

Fixes: https://issues.redhat.com/browse/OCPBUGS-84718

```release-note
Fix unnecessary storage wipe after unclean shutdown when only stale overlay mounts are present.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unclean shutdown handling to clean up stale overlay mounts before attempting storage repair.
  * Enhanced storage repair logic to intelligently preserve storage when encountering transient errors, avoiding unnecessary data loss.

* **Tests**
  * Added test cases validating storage behavior during unclean shutdowns with stale mounts and corruption scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cri-o/cri-o/pull/9968?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"syscall"
 	"time"
 
 	json "github.com/goccy/go-json"
@@ -119,6 +120,13 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
 
 	if config.InternalRepair && ShutdownWasUnclean(config) {
 		graphRoot := store.GraphRoot()
+
+		// Clean stale overlay mounts from the unclean shutdown so that
+		// store.Check() does not report false layer errors.
+		if err := cmount.RecursiveUnmount(filepath.Join(graphRoot, "overlay")); err != nil {
+			log.Warnf(ctx, "Failed to clean stale overlay mounts under %s: %v", graphRoot, err)
+		}
+
 		log.Warnf(ctx, "Checking storage directory %s for errors because of unclean shutdown", graphRoot)
 
 		wipeStorage := false
@@ -128,7 +136,14 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
 			log.Warnf(ctx, "Attempting to repair storage directory %s because of unclean shutdown", graphRoot)
 
 			if errs := store.Repair(report, cstorage.RepairEverything()); len(errs) > 0 {
-				wipeStorage = true
+				for _, repairErr := range errs {
+					log.Warnf(ctx, "Storage repair error for %s: %v", graphRoot, repairErr)
+				}
+				if allRepairErrorsAreTransient(errs) {
+					log.Warnf(ctx, "Storage repair for %s encountered only transient errors (e.g. device busy), skipping wipe", graphRoot)
+				} else {
+					wipeStorage = true
+				}
 			}
 		} else if err != nil {
 			// Storage check has failed with irrecoverable errors.
@@ -1070,4 +1085,15 @@ func (c *ContainerServer) probeMonitorProcesses(ctx context.Context) {
 
 		timer.Reset(probeInterval + time.Duration(rand.Int63n(probeJitter.Nanoseconds())))
 	}
+}
+
+// allRepairErrorsAreTransient returns true if every repair error is a transient
+// "device or resource busy" issue rather than actual storage corruption.
+func allRepairErrorsAreTransient(errs []error) bool {
+	for _, err := range errs {
+		if !errors.Is(err, syscall.EBUSY) {
+			return false
+		}
+	}
+	return len(errs) > 0
 }
--- a/test/crio-wipe.bats
+++ b/test/crio-wipe.bats
@@ -386,6 +386,62 @@ function start_crio_with_stopped_pod() {
 	fi
 }
 
+@test "skip wipe on unclean shutdown with only stale overlay mounts" {
+	setup_crio
+	touch "$CONTAINER_CLEAN_SHUTDOWN_FILE".supported
+
+	start_crio_no_setup
+
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	ctr_id=$(crictl create "$pod_id" "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json)
+	crictl start "$ctr_id"
+	crictl stopp "$pod_id"
+
+	stop_crio_no_clean
+
+	# Simulate stale overlay mounts left behind after kill -9.
+	for dir in "$TESTDIR"/crio/overlay/*/merged; do
+		[ -d "$dir" ] && mount --bind "$dir" "$dir"
+	done
+
+	rm -Rf "$CONTAINER_CLEAN_SHUTDOWN_FILE"
+
+	CONTAINER_INTERNAL_REPAIR=true start_crio_no_setup
+
+	# Storage should not have been wiped.
+	[[ $(crictl images) == *"quay.io/crio"* ]]
+}
+
+@test "wipe storage on unclean shutdown with real corruption" {
+	setup_crio
+	touch "$CONTAINER_CLEAN_SHUTDOWN_FILE".supported
+
+	start_crio_no_setup
+
+	pod_id=$(crictl runp "$TESTDATA"/sandbox_config.json)
+	ctr_id=$(crictl create "$pod_id" "$TESTDATA"/container_config.json "$TESTDATA"/sandbox_config.json)
+	crictl start "$ctr_id"
+
+	# This will corrupt the storage directory.
+	cp -r "$TESTDIR"/crio/overlay{,.old}
+	umount -R -l -f "$TESTDIR"/crio/overlay
+	rm -Rf "$TESTDIR"/crio/overlay
+	cp -r "$TESTDIR"/crio/overlay{.old,}
+
+	stop_crio_no_clean
+
+	rm -Rf "$CONTAINER_CLEAN_SHUTDOWN_FILE"
+
+	CONTAINER_INTERNAL_REPAIR=true start_crio_no_setup
+
+	# Storage directory wipe should leave only the metadata behind.
+	size=$(du -sb "$TESTDIR"/crio | cut -f 1)
+	if ((size > 1024 * 128)); then
+		echo "The storage directory wipe did not work" >&3
+		return 1
+	fi
+}
+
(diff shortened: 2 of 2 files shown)


internal/lib/container_server.go
```diff
@@ -896,3 +908,14 @@ func CheckReportHasErrors(report cstorage.CheckReport) bool {
 		len(report.Images) > 0 || len(report.ROImages) > 0 ||
 		len(report.Containers) > 0
 }
+
+// allRepairErrorsAreTransient returns true if every repair error is a transient
+// "device or resource busy" issue rather than actual storage corruption.
+func allRepairErrorsAreTransient(errs []error) bool {
+	for _, err := range errs {
+		if !strings.Contains(err.Error(), "device or resource busy") {
```
haircommander: are we not able to use errors.Is instead?

Chandan9112: used in latest commit.

Chandan9112:
@coderabbitai review

test/crio-wipe.bats
```diff
@@ -386,6 +386,62 @@ function start_crio_with_stopped_pod() {
...
+		[ -d "$dir" ] && mount --bind "$dir" "$dir"
+	done
+
+	rm -Rf "$CONTAINER_CLEAN_SHUTDOWN_FILE"
+
+	CONTAINER_INTERNAL_REPAIR=true start_crio_no_setup
+
+	# Storage should not have been wiped.
+	[[ $(crictl images) == *"quay.io/crio"* ]]
+}
```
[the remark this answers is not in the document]
Chandan9112: Bind-mount simulation is intentional -- it produces the exact EBUSY condition seen in production. Existing tests in this file also use synthetic simulation rather than real crashes.

internal/lib/container_server.go
```diff
@@ -107,14 +108,25 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
 
 	if config.InternalRepair && ShutdownWasUnclean(config) {
 		graphRoot := store.GraphRoot()
+
+		// Clean stale overlay mounts from the unclean shutdown so that
+		// store.Check() does not report false layer errors.
+		if err := cmount.RecursiveUnmount(filepath.Join(graphRoot, "overlay")); err != nil {
+			log.Warnf(ctx, "Failed to clean stale overlay mounts under %s: %v", graphRoot, err)
```
ngopalak-redhat, line 127: Should we abort here? Or at least set a flag that we're not confident?

internal/lib/container_server.go
```diff
@@ -107,14 +108,25 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
...
+
 		log.Warnf(ctx, "Checking storage directory %s for errors because of unclean shutdown", graphRoot)
 
 		wipeStorage := false
 		report, err := store.Check(checkQuick())
 		if err == nil && CheckReportHasErrors(report) {
 			log.Warnf(ctx, "Attempting to repair storage directory %s because of unclean shutdown", graphRoot)
 			if errs := store.Repair(report, cstorage.RepairEverything()); len(errs) > 0 {
-				wipeStorage = true
+				if allRepairErrorsAreTransient(errs) {
```
ngopalak-redhat, line 142: Logging/understanding what layers failed and why. If you have mixed errors, you want to know WHY

ngopalak-redhat (changes requested):

internal/lib/container_server.go
Chandan9112: I think RecursiveUnmount failure is good effort as if it can't clean stale mounts, the subsequent store.Check/Repair path will still detect and handle any real corruption, so aborting early would be unnecessarily aggressive.

internal/lib/container_server.go
Chandan9112: modified in the latest commit

ngopalak-redhat:
I understand the change to some extent. Both `RecursiveUnmount` and `Repair` will fail when EBUSY error is encountered. But there is no way to confirm that the layer which got EBUSY is not corrupt. This PR only makes sure that an attempt to wipe the layer which got EBUSY is avoided.

internal/lib/container_server.go
```diff
@@ -107,14 +108,28 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
 
 	if config.InternalRepair && ShutdownWasUnclean(config) {
 		graphRoot := store.GraphRoot()
+
+		// Clean stale overlay mounts from the unclean shutdown so that
+		// store.Check() does not report false layer errors.
+		if err := cmount.RecursiveUnmount(filepath.Join(graphRoot, "overlay")); err != nil {
```
bitoku, line 126: Does it always unmount overlay mounts when cri-o shutdown was unclean?
If containers are running successfully during CRI-O is dead, we don't have to unmount, do we?

internal/lib/container_server.go
```diff
@@ -107,14 +108,28 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
...
 		wipeStorage := false
 		report, err := store.Check(checkQuick())
 		if err == nil && CheckReportHasErrors(report) {
 			log.Warnf(ctx, "Attempting to repair storage directory %s because of unclean shutdown", graphRoot)
 			if errs := store.Repair(report, cstorage.RepairEverything()); len(errs) > 0 {
-				wipeStorage = true
+				for _, repairErr := range errs {
+					log.Warnf(ctx, "Storage repair error for %s: %v", graphRoot, repairErr)
+				}
+				if allRepairErrorsAreTransient(errs) {
```
bitoku, line 142: I'm skeptical that we can treat EBUSY as a transient error.
I think some process could keep holding the file.

bitoku:
and is it applicable to the latest CRI-O? We should make a PR to main first and then propagate it to the older versions.

internal/lib/container_server.go
ngopalak-redhat: That's the path I was investigating too... May the there is no way to confirm that the layer which got EBUSY is not corrupt. The fix is better than failing for EBUSY.

internal/lib/container_server.go
bitoku, line 126: Other storage drivers (e.g., vfs, zfs, btrfs) could also leave stale mounts.

ngopalak-redhat: So if containers are running and cri-o is dead, this method would get EBUSY error in my analysis. Hence unmount won't happen.

Chandan9112: Not sure, for now the fix targets overlay since that's where this bug was observed.

bitoku: How did you test that? Or maybe our definition of unclean is different.

I tested:
1. run cri-o
2. run container
3. SIGKILL cri-o
4. run cri-o

and it didn't get any EBUSY error in the log, and the container is still running.

bitoku: I got `Checking storage directory /var/lib/containers/storage for errors because of unclean shutdown` so it should've reached L130.

internal/lib/container_server.go
bitoku: > The fix is better than failing for EBUSY.

It depends. It may break auto recovery that worked before. some clusters might be lenient with pulling images, but strict with manual intervention.

internal/lib/container_server.go
bitoku: Do we have this test scenario? @Chandan9112 

> run cri-o
run container
SIGKILL cri-o
run cri-o

Chandan9112: @bitoku., Not this exact scenario. Our test simulates the stale mounts via bind-mounts after a graceful stop rather than an actual SIGKILL, I think SIGKILL alone doesn't reliably produce EBUSY without additionalImageStores configured.

bitoku: > Not this exact scenario.

I think we should test this scenario to ensure the change won't break the current behavior.

> Our test simulates the stale mounts via bind-mounts after a graceful stop rather than an actual SIGKILL

Graceful shutdown of CRI-O is not unclean shutdown.
I wonder if the current test cases are testing the scenarios properly.

I think this fix is required when CRI-O shutdown uncleanly, is that correct?
If so, I want the test cases to check "Checking storage directory %s for errors because of unclean shutdown" to ensure that the shutdown is unclean.

> I think SIGKILL alone doesn't reliably produce EBUSY without additionalImageStores configured.

I think so too.

internal/lib/container_server.go
```diff
@@ -119,6 +120,13 @@ func New(ctx context.Context, configIface libconfig.Iface) (*ContainerServer, er
...
 	if config.InternalRepair && ShutdownWasUnclean(config) {
 		graphRoot := store.GraphRoot()
+
+		// Clean stale overlay mounts from the unclean shutdown so that
+		// store.Check() does not report false layer errors.
+		if err := cmount.RecursiveUnmount(filepath.Join(graphRoot, "overlay")); err != nil {
+			log.Warnf(ctx, "Failed to clean stale overlay mounts under %s: %v", graphRoot, err)
+		}
+
 		log.Warnf(ctx, "Checking storage directory %s for errors because of unclean shutdown", graphRoot)
```
bitoku, line 130: This wants to move before unmount.

PR #9974, opened 2026-05-24
feat: Add support for containerd-shim-runsc-v1

### LINKED ISSUE #10313: cri-o support
### Description

I have been tinkering to get runsc to work with crio and I think I have it mostly working. Except for the fact that initContainers don't seem to be supported properly.

Reading the multi container support code it seems like gVisor expects the first container to be the root container but this assumptions breaks down when the first container is an init container which runs to completion.

I am not sure what is the right way to handle this as the CRI spec doesn't seem to indicate if a container is an init container.

Opening it here for discussion.

### Is this feature related to a specific bug?

_No response_

### Do you have a specific solution in mind?

_No response_

xw19:
#### What type of PR is this?

/kind feature

To support gVisor in CRI-O, this update extends the `RuntimeTypeVMBinaryPattern` to accept the `runsc-v1` shim naming convention and refactors `ParseShimAddress()` to handle gVisor’s JSON-formatted `BootstrapParams` alongside standard address strings. Additionally, the configuration now requires `runtime_type` to be set to `vm` to ensure CRI-O correctly creates the pause container via the shim path, a necessary step for gVisor sandbox initialization.

Fixes https://github.com/google/gvisor/issues/10313

```release-note
Support gVisor containerd-shim-runsc-v1
```

## Summary by CodeRabbit

* **New Features**
  * Added support for additional VM shim naming conventions (e.g., v1-style shims) alongside existing v2 support.

* **Bug Fixes**
  * Improved shim address extraction with stricter JSON handling and clearer parsing errors.

* **Documentation**
  * Expanded runtime type docs to list valid values (oci, vm, pod) and updated runtime selection examples to include gVisor.

* **Tests**
  * Added tests for shim address parsing and runtime binary validation.

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -357,7 +357,11 @@ Path to the OCI compatible runtime used for this runtime handler.
 Root directory used to store runtime data
 
 **runtime_type**="oci"
-Type of the runtime used for this runtime handler. "oci", "vm"
+Type of the runtime used for this runtime handler. Valid values are:
+
+- `"oci"` (default): Standard OCI runtime (e.g. runc, crun). The `runtime_path` should point to the OCI runtime binary.
+- `"vm"`: VM-isolation shim using the containerd shimv2/ttrpc protocol. The `runtime_path` must be a `containerd-shim-*` binary. This runtime type is exercised in CI with [Kata Containers](https://github.com/kata-containers/kata-containers) (`containerd-shim-kata-v2`). Other shimv2 shims implement the same protocol and may work — for example [gVisor](https://github.com/google/gvisor)'s `containerd-shim-runsc-v1` — but are not currently covered by CI.
+- `"pod"`: Pod-level runtime using [conmon-rs](https://github.com/containers/conmon-rs) instead of conmon. conmon-rs operates at pod granularity.
 
 **inherit_default_runtime**=false
 Override the runtime path, runtime config path, runtime root and runtime type from the default runtime on load.
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -254,6 +254,28 @@ func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupPar
 	return nil
 }
 
+// ParseShimAddress extracts the shim's ttrpc address from its "start" stdout
+// output. It mirrors containerd's own parseStartResponse:
+// https://github.com/containerd/containerd/blob/v1.7.31/runtime/v2/shim.go#L211
+//
+// Modern shims (the containerd v2 shim API, including gVisor's
+// containerd-shim-runsc-v1) emit a JSON BootstrapParams object; legacy shims
+// emit just the address string. If the output does not unmarshal into a
+// versioned (>= 2) BootstrapParams object, it is treated as a raw address.
+func ParseShimAddress(out []byte) (string, error) {
+	var params client.BootstrapParams
+	if err := json.Unmarshal(out, &params); err != nil || params.Version < 2 {
+		// Legacy shim: the raw output is the ttrpc address.
+		return strings.TrimSpace(string(out)), nil
+	}
+
+	if params.Version > 2 {
+		return "", fmt.Errorf("unsupported shim version (%d): %w", params.Version, errdefs.ErrNotImplemented)
+	}
+
+	return params.Address, nil
+}
+
 func (r *runtimeVM) startRuntimeDaemon(ctx context.Context, c *Container) error {
 	log.Debugf(ctx, "RuntimeVM.startRuntimeDaemon() start")
 	defer log.Debugf(ctx, "RuntimeVM.startRuntimeDaemon() end")
@@ -301,14 +323,22 @@ func (r *runtimeVM) startRuntimeDaemon(ctx context.Context, c *Container) error 
 		}
 	}()
 
-	// Start the server
-	out, err := cmd.CombinedOutput()
+	// Capture stderr separately so that anything the shim writes
+	// there cannot corrupt the BootstrapParams JSON (or raw address)
+	// it emits on stdout, which we can parse below.
+	var stderr bytes.Buffer
+
+	cmd.Stderr = &stderr
+
+	out, err := cmd.Output()
 	if err != nil {
-		return fmt.Errorf("%s: %w", string(out), err)
+		return fmt.Errorf("%s: %w", stderr.String(), err)
 	}
 
-	// Retrieve the address from the output
-	address := strings.TrimSpace(string(out))
+	address, err := ParseShimAddress(out)
+	if err != nil {
+		return fmt.Errorf("parsing shim address: %w", err)
+	}
 
 	// Now the RPC server is running, let's connect to it
 	conn, err := client.Connect(address, client.AnonDialer)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -70,7 +70,7 @@ const (
 	RuntimeTypePod                = "pod"
 	defaultCtrStopTimeout         = 30 // seconds
 	defaultNamespacesDir          = "/var/run"
-	RuntimeTypeVMBinaryPattern    = "containerd-shim-([a-zA-Z0-9\\-\\+])+-v2"
+	RuntimeTypeVMBinaryPattern    = "^containerd-shim-[a-zA-Z0-9\\-\\+]+$"
 	tasksetBinary                 = "taskset"
 	MonitorExecCgroupDefault      = ""
 	MonitorExecCgroupContainer    = "container"
--- a/internal/oci/runtime_vm_test.go
+++ b/internal/oci/runtime_vm_test.go
@@ -1,0 +1,86 @@
+package oci_test
+
+import (
+	"testing"
+
+	"github.com/cri-o/cri-o/internal/oci"
+)
+
+func TestParseShimAddress(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		input   string
+		want    string
+		wantErr bool
+	}{
+		{
+			name:  "plain address string",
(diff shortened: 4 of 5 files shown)


xw19:
@cri-o/cri-o-maintainers PTAL

pkg/config/config.go
```diff
@@ -70,10 +70,11 @@ const (
 	RuntimeTypePod                = "pod"
 	defaultCtrStopTimeout         = 30 // seconds
 	defaultNamespacesDir          = "/var/run"
-	RuntimeTypeVMBinaryPattern    = "containerd-shim-([a-zA-Z0-9\\-\\+])+-v2"
-	tasksetBinary                 = "taskset"
-	MonitorExecCgroupDefault      = ""
-	MonitorExecCgroupContainer    = "container"
+	// runsc (gVisor) uses the v1 shim; all other VM runtimes are expected to use v2.
```
bitoku: containerd-shim-runsc-v1 is actually a containerd v2 shim.

https://github.com/google/gvisor/blob/9b9807a3c82d075307c55da1b8ea0e52c972680c/shim/main.go#L15

pkg/config/config.go
```diff
@@ -70,10 +70,11 @@ const (
...
+	RuntimeTypeVMBinaryPattern = "containerd-shim-(runsc-v1|([a-zA-Z0-9\\-\\+])+-v2)"
```
bitoku: I'd rather cri-o allow broader shims (containerd-shim-*) and explicitly say the supported shims in the document (or having test matrix what's tested).

test/gvisor.bats
```diff
@@ -0,0 +1,126 @@
...
+
+# Integration tests for the CRI-O + gVisor (runsc) integration.
+#
+# STATUS: STUB — these tests require a patched containerd-shim-runsc-v1 that
+# accepts CRI-O's Options proto (no TypeUrl) and recognises the CRI-O sandbox
+# annotation for shim grouping. That fix is tracked in the companion gVisor PR.
+# Until that PR merges and a patched shim binary is available in CI, all tests
+# in this file skip automatically.
+#
+# Covers: https://github.com/cri-o/cri-o/issues/10313
```
bitoku: typo or hallucination

test/gvisor.bats
bitoku, line 1: I'm opposed to have the code (especially test code) that's not tested at all in CI. 
Also if we support this shim, it should pass most of the existing tests.
These don't look gvisor specific, so I don't see benefit to have this test file.

test/gvisor.bats
```diff
@@ -0,0 +1,126 @@
...
+# Companion gVisor PR: https://github.com/google/gvisor/pull/XXXX
```
bitoku: It should be prepared.

test/gvisor.bats
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env bats
+# vim: set syntax=sh:
+
+# Integration tests for the CRI-O + gVisor (runsc) integration.
+#
+# STATUS: STUB — these tests require a patched containerd-shim-runsc-v1 that
+# accepts CRI-O's Options proto (no TypeUrl) and recognises the CRI-O sandbox
+# annotation for shim grouping. That fix is tracked in the companion gVisor PR.
+# Until that PR merges and a patched shim binary is available in CI, all tests
+# in this file skip automatically.
```
bitoku: Can you elaborate?

pkg/config/config.go
bitoku: https://github.com/containerd/containerd/discussions/7407#discussioncomment-3689436

test/gvisor.bats
xw19: Let me clear this file. I just want a placeholder test file for now if we our changes merged in gvisor then we put some real tests using that.

xw19:
@littlejawa  hi can you please have a look at the PR?

pkg/config/config.go
littlejawa: I agree. This is pretty restrictive, and I'm not sure it helps.

littlejawa:
> @littlejawa hi can you please have a look at the PR?

@xw19 - I don't think that your changes will cause issues to the kata shim.
As long as the kata integration tests are passing, I think it's ok.

I'm a bit curious about your shim's support though.
The runtime_vm implementation has always been done with kata in mind. It implements the shimv2 protocol, so in theory it should be usable with other shims, but in practice it's never been exercised with something else than kata...

Did you try to run the integration tests with your shim?

xw19:
> > @littlejawa hi can you please have a look at the PR?
> 
> @xw19 - I don't think that your changes will cause issues to the kata shim. As long as the kata integration tests are passing, I think it's ok.
> 
> I'm a bit curious about your shim's support though. The runtime_vm implementation has always been done with kata in mind. It implements the shimv2 protocol, so in theory it should be usable with other shims, but in practice it's never been exercised with something else than kata...
> 
> Did you try to run the integration tests with your shim?

@littlejawa  Thank you for the review.
I think most tests are passing I still need to check on few failures.
The runsc shim is v2 in disguise of v1
runsc works kind of similar to vm isolation.
Yes most of the tests are passing there are few unrelated failure I guess which I am trying to fix.

docs/crio.conf.5.md
```diff
@@ -357,7 +357,13 @@ Path to the OCI compatible runtime used for this runtime handler.
 Root directory used to store runtime data
 
 **runtime_type**="oci"
-Type of the runtime used for this runtime handler. "oci", "vm"
+Type of the runtime used for this runtime handler. Valid values are:
+
+- `"oci"` (default): Standard OCI runtime (e.g. runc, crun). The `runtime_path` should point to the OCI runtime binary.
+- `"vm"`: VM-isolation shim using the containerd shimv2/ttrpc protocol. The `runtime_path` must be a `containerd-shim-*` binary. Known supported shims:
+  - [Kata Containers](https://github.com/kata-containers/kata-containers): `containerd-shim-kata-v2`, `containerd-shim-kata-qemu-v2`
```
bitoku: @littlejawa Can you double check if it's true?

docs/crio.conf.5.md
```diff
@@ -357,7 +357,13 @@ Path to the OCI compatible runtime used for this runtime handler.
...
+  - [gVisor](https://github.com/google/gvisor): `containerd-shim-runsc-v1`
```
bitoku: We don't have a test job yet, so we don't want to say it's "supported".
I'd rather this PR to be something that enables more runtime options, but not call them "supported".

test/README.md
```diff
@@ -86,6 +86,12 @@ For example, to use
 make CONTAINER_DEFAULT_RUNTIME=containerd-shim-kata-v2 RUNTIME_TYPE=vm localintegration
 ```
 
+Or to use [gVisor](https://github.com/google/gvisor):
+
+```shell
+make CONTAINER_DEFAULT_RUNTIME=containerd-shim-runsc-v1 RUNTIME_TYPE=vm localintegration
+```
```
bitoku: Have you checked all tests passed?
If so, why don't we add the test in GA?
Otherwise we shouldn't write this in documents.

pkg/config/config.go
```diff
@@ -70,10 +70,12 @@ const (
 	RuntimeTypePod                = "pod"
 	defaultCtrStopTimeout         = 30 // seconds
 	defaultNamespacesDir          = "/var/run"
-	RuntimeTypeVMBinaryPattern    = "containerd-shim-([a-zA-Z0-9\\-\\+])+-v2"
-	tasksetBinary                 = "taskset"
-	MonitorExecCgroupDefault      = ""
-	MonitorExecCgroupContainer    = "container"
+	// The suffix (v1, v2, etc.) is the plugin API version, not the containerd shim protocol version.
+	// All current VM shims speak shimv2/ttrpc regardless of their suffix.
```
bitoku: It doesn't have to be a VM shim.
I think we can remove this comment.

bitoku (commented):
I'm open to enable CRI-O to run other shims, but don't want to explicitly say it's "supported".

xw19:
@bitoku can you please re-review I have addressed all the comments.

internal/oci/runtime_vm.go
```diff
@@ -254,6 +254,24 @@ func (r *runtimeVM) CreateContainer(ctx context.Context, c *Container, cgroupPar
...
+		var params client.BootstrapParams
+		if err := json.Unmarshal(out, &params); err != nil {
+			return "", fmt.Errorf("parse shim bootstrap params %q: %w", address, err)
+		}
+
+		address = params.Address
+	}
+
+	return address, nil
+}
```
bitoku: Do we have a document about it? I don't know how valid this check is.

bitoku (commented):
/lgtm

Thanks!
To follow up, we may want to run integration tests with gvisor.

bitoku:
@cri-o/cri-o-maintainers PTAL

internal/oci/runtime_vm.go
sohankunkerkar: CombinedOutput() mixes stderr into the output, so if the gVisor shim writes anything to stderr during start, the JSON parse fails and the fallback returns garbage as the address, should this be cmd.Output() instead?

internal/oci/runtime_vm.go
```diff
@@ -307,8 +329,10 @@ func (r *runtimeVM) startRuntimeDaemon(ctx context.Context, c *Container) error
 		return fmt.Errorf("%s: %w", string(out), err)
 	}
 
-	// Retrieve the address from the output
-	address := strings.TrimSpace(string(out))
+	address, err := ParseShimAddress(out)
+	if err != nil {
+		return err
```
sohankunkerkar: Can you log an error here?
```suggestion
		return return fmt.Errorf("parsing shim address: %w", err)
```

xw19:
@sohankunkerkar  can you please re-review?

sohankunkerkar (commented):
/lgtm

PR #9991, opened 2026-06-01
Fix patch release workflow spec file version mismatch

saschagrunert:
/kind bug

#### What type of PR is this?

/kind bug

The [patch release workflow failed](https://github.com/cri-o/cri-o/actions/runs/26729881084/job/78771719011) because the release script required an exact version match in `contrib/test/ci/cri-o.spec`, but release branches created before commit 0305311d2 still have a stale version (`1.32.0`) in the spec file. This caused the workflow to fail for `release-1.36`.
This change matches any `Version:` line in the spec file instead of requiring the exact old version, since the script sets the correct new version regardless.

None

The `release-1.33`, `release-1.34`, `release-1.35`, and `release-1.36` branches all have `Version: 1.32.0` in the spec file. With this fix, the release script will handle that gracefully.

```release-note
None
```

## Summary by CodeRabbit

* **Chores**
  * Simplified the internal release script version update process for improved efficiency.

### DIFF
--- a/scripts/release/release.go
+++ b/scripts/release/release.go
@@ -118,7 +118,7 @@ func updateVersionAndCreatePR(
 
 	logrus.Infof("Updating version in spec file %s", utils.SpecFile)
 
-	if err := updateSpecFile(oldVersion, newVersion); err != nil {
+	if err := updateSpecFile(newVersion); err != nil {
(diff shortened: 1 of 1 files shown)


saschagrunert:
@cri-o/cri-o-maintainers PTAL

PR #9776, opened 2026-02-19
oci: prevent nil pointer panic in runtimeVM.kill after restart

tobiasworkstech:
/kind bug

## What this PR does / why we need it

When CRI-O restarts, `runtimeVM` is reconstructed with `r.task == nil`.
The task client is only assigned later — either when `CreateContainer` connects
to a newly spawned shim, or when `updateContainerStatus` reconnects to an
existing one via the `address` file in the container bundle.

If kubelet immediately drives a VM container to termination after the restart
(before `ContainerStatus` has been called to trigger reconnection), the call chain

```
StopContainer → kill → r.task.Kill(...)   ← panic: nil pointer dereference
```

crashes CRI-O, as reported in #9772. The goroutine dump from that report shows
a container that had been in the `StopContainer` path for 837 minutes.

## Which issue(s) this PR fixes

Fixes #9772

## How the fix works

**`connectTask()` helper** — extracts the reconnect-to-shim logic already present
in `updateContainerStatus` into a reusable method. It is a no-op when `r.task` is
already set. If the shim `address` file is missing and the container is already in
stopped state, it returns `nil` with `r.task` still `nil` (caller should bail out).
Otherwise it reads the address, connects, and sets `r.task`/`r.client`.

**`StopContainer`** — calls `connectTask()` before the first `kill()` call. If
reconnection fails the error is returned (no panic). If the shim is gone and the
container is already stopped, returns `nil` cleanly.

**`kill()` nil guard** — defence-in-depth: returns `errdefs.ErrNotFound` instead
of panicking when `r.task` is nil. Protects `IsContainerAlive` and any other
caller.

`updateContainerStatus` is refactored to use `connectTask()` while preserving its
original error handling and log messages.

## Release note

```release-note
Fixed a nil pointer panic in runtimeVM.kill (Kata Containers / VM runtime) that
could crash CRI-O when kubelet restarts and immediately terminates VM containers
before their shim connection is re-established.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes by guarding container operations when no runtime connection exists.
  * Kill/stop now return clear "not found" behavior for stopped containers and avoid nil-pointer errors.
  * Improved handling and error messages for containers whose runtime/shim is missing, and removed duplicated connection logic.
* **Tests**
  * Added unit tests covering reconnect behavior, missing shim/address cases, and nil-task guards.

Issue #9772, opened 2026-02-17
Repeated panic while shutting down VM runtime


c3d:
### What happened?

On a node with OpenShift Sandboxed Containers running CRI-O 1.33.6 on OpenShift 4.20 / RHCOS 9.6, I observed repeated CRI-O panics in `runtimeVM.kill` (`runtime_vm.go:1295`) for pods with `runtimeClass: kata` with the following stack trace:


```
Feb 16 15:41:23 kata420-worker-2.kata420.com systemd-coredump[340278]: Process 339888 (crio) of user 0 dumped core.                                                                   Stack trace of thread 339894:
                                                                   #0  0x000056405c451b61 runtime.raise.abi0 (crio + 0x140bb61)
                                                                   #1  0x000056405c428368 runtime.sigfwdgo (crio + 0x13e2368)
                                                                   #2  0x000056405c426ba5 runtime.sigtrampgo (crio + 0x13e0ba5)
                                                                   #3  0x000056405c451e69 runtime.sigtramp.abi0 (crio + 0x140be69)
                                                                   #4  0x00007f24e48c9bf0 __restore_rt (libc.so.6 + 0x3ebf0)
                                                                   #5  0x000056405c451b61 runtime.raise.abi0 (crio + 0x140bb61)
                                                                   #6  0x000056405c40daf3 runtime.fatalpanic (crio + 0x13c7af3)
                                                                   #7  0x000056405c447728 runtime.gopanic (crio + 0x1401728)
                                                                   #8  0x000056405c449e39 runtime.sigpanic (crio + 0x1403e39)
                                                                   #9  0x000056405db32a67 github.com/cri-o/cri-o/internal/oci.(*runtimeVM).kill (crio + 0x2aeca67)
                                                                   #10 0x000056405db2dda5 github.com/cri-o/cri-o/internal/oci.(*runtimeVM).StopContainer (crio + 0x2ae7da5)
                                                                   ELF object binary architecture: AMD x86-64Feb 16 15:41:23 kata420-worker-2.kata420.com systemd[1]: crio.service: Main process exited, code=dumped, status=6/ABRT
```

There are other issues in the shutdown path, including #8030, #8031, #8300, #8392 and #8393, but I have not found any with this specific panic.


### What did you expect to happen?

CRI-O should do a clean shutdown even for a VM-based runtime, even if there is a misconfiguration

### How can we reproduce it (as minimally and precisely as possible)?

This happened after:
1. Installing a cluster using `kcli`
[...]

PR #9994, opened 2026-06-01
[wip] oci: fix kata pod stuck in Terminating when VM shim exits before StopPodSandbox

thejasn:
#### What type of PR is this?

/kind bug

When a kata pod is deleted, the VM shim can exit before CRI-O calls
`StopPodSandbox`. This causes two issues in `runtimeVM`:

1. `kill()` receives `ttrpc.ErrClosed` from the dead shim and maps it
    to `ErrUnknown`, causing `StopPodSandbox` to fail and kubelet to
    retry indefinitely — leaving the pod stuck in `Terminating`.
    
2. `StopContainer` updates `c.state.Finished` on success but does not
    set `c.state.Status`, leaving the container in a non-stopped
    internal state after termination.

Fix (1) adds a `ttrpc.ErrClosed` guard to `kill()`, consistent with
the same guard already present in `wait()` and `remove()`. 
Fix (2) explicitly sets `ContainerStateStopped` on both exit paths
of `StopContainer`.

Fixes: rhjira#KATA-4869

The `kill()` fix is the critical one — without it the pod never
terminates. The `StopContainer` status fix is a correctness improvement
in the same code path.

Tested on OCP 4.21 SNO with kata-qemu (kata 3.25.0) terminations with
this patch applied.

```release-note
TBD
```

Testing: WIP

## Summary by CodeRabbit

* **Bug Fixes**
  * Container state is now correctly marked as stopped after successful graceful or forced termination.
  * Container termination is more resilient when the underlying task service is already closed, while other errors continue to be reported.
* **Tests**
  * Added coverage for successful container termination, closed task services, and other termination errors.

### DIFF
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -686,6 +686,7 @@ func (r *runtimeVM) StopContainer(ctx context.Context, c *Container, timeout int
 		err := r.waitCtrTerminate(sig, stopCh, timeoutDuration)
 		if err == nil {
 			c.state.Finished = time.Now()
+			c.state.Status = ContainerStateStopped
 
 			return nil
 		}
@@ -706,6 +707,7 @@ func (r *runtimeVM) StopContainer(ctx context.Context, c *Container, timeout int
 	}
 
 	c.state.Finished = time.Now()
+	c.state.Status = ContainerStateStopped
 
 	return nil
 }
--- a/internal/oci/runtime_vm.go
+++ b/internal/oci/runtime_vm.go
@@ -1151,7 +1151,7 @@ func (r *runtimeVM) kill(ctrID, execID string, signal syscall.Signal) error {
 		ExecID: execID,
 		Signal: uint32(signal),
 		All:    false,
-	}); err != nil {
+	}); err != nil && !errors.Is(err, ttrpc.ErrClosed) {
 		return errdefs.FromGRPC(err)
 	}
 
--- a/internal/oci/runtime_vm_test.go
+++ b/internal/oci/runtime_vm_test.go
@@ -1,0 +1,64 @@
+package oci_test
+
+import (
+	"context"
+	"syscall"
+
+	task "github.com/containerd/containerd/api/runtime/task/v2"
+	"github.com/containerd/ttrpc"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/cri-o/cri-o/internal/oci"
+)
+
+// fakeTaskService is a minimal task.TaskService stub for testing kill() behavior.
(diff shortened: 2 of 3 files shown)


fidencio:
@thejasn, I think we have more to dig into.

For instance, there are a few questions that I'd love to get an answer before reviewing this PR.
1. Has this always been happening?
2. If yes, how this was not found by the CI @ldoktor has been maintaining against Kata?
3. If not, which version of CRI-O was the first one affected?
4. If not, which version of Kata Containers is the first one causing this?

thejasn:
Hi @fidencio, 

> 1. Has this always been happening?

No

> 4. If not, which version of Kata Containers is the first one causing this?

Regression Commit: [956f43c6c](https://github.com/kata-containers/kata-containers/commit/956f43c6c605fe8c3102f4a6ec71c34ddaedce88) — "runtime: skip MoveTo for systemd cgroups" (kata 3.25.0). 

The issues seems to only affect multi process workloads (like `wordpress:latest` as the bug was filed on this). Even though the CI runs conformance tests on latest kata + crio, my guess is I don't think they have any multi-process workloads in their tests. 

I am able to reproduce the issue always on the below environment,
```
  ┌─────────────────┬───────────────────────────────────┐
  │    Component    │              Version              │
  ├─────────────────┼───────────────────────────────────┤
  │ OCP             │ 4.21 SNO                          │
  ├─────────────────┼───────────────────────────────────┤
  │ kata-containers │ 3.25.0-5.rhaos4.21.el9            │
  ├─────────────────┼───────────────────────────────────┤
  │ CRI-O           │ 1.34.8-3.rhaos4.21.git91b4b0b.el9 │
  ├─────────────────┼───────────────────────────────────┤
  │ RuntimeClass    │ kata (QEMU backend)               │
  └─────────────────┴───────────────────────────────────┘
``` 
Test workload,
```
apiVersion: v1
kind: Pod
metadata:
  name: kata-termination-reproducer
spec:
  runtimeClassName: kata
  terminationGracePeriodSeconds: 30
  containers:
  - name: workload
    image: busybox
    command:
    - sh
    - -c
    - |
      for i in 1 2 3 4 5; do
        sh -c 'while true; do sleep 1; done' &
      done
      trap 'sleep 1; kill 0; wait; exit 0' TERM
      wait
```

fidencio:
> The issues seems to only affect multi process workloads (like wordpress:latest as the bug was filed on this). Even though the CI runs conformance tests on latest kata + crio, my guess is I don't think they have any multi-process workloads in their tests.

And are you sure this is a CRI-O specific issue? Have you tried the same workload using, let's say, containerd to ensure this is something that should be fixed at the CRI-O level and not at the Kata Containers level?

c3d:
Hi @thejasn, the fix looks good to me. I was going to mention #9772 which I believe is related, but the bot beat me to it. Maybe you want to take a look at #9776, which is a fix to it.

PR #10014, opened 2026-06-08
User namespace - Handle UID/GID mappings from the infra container

ngopalak-redhat:
#### What type of PR is this?

/kind bug

When using user namespaces with ID-mapped mounts, containers fail with "permission denied" errors after:
  1. CRI-O is restarted
  2. The real container process is killed (kill 1)

Here's the error the user has been facing:
```
2026-02-16T22:15:29.874393249Z I0216 22:15:29.874325       1 cmd.go:253] Using service-serving-cert provided certificates
2026-02-16T22:15:29.874393249Z I0216 22:15:29.874375       1 leaderelection.go:121] The leader election gives 4 retries and allows for 30s of clock skew. The kube-apiserver downtime tolerance is 78s. Worst non-graceful lease acquisition is 2m43s. Worst graceful lease acquisition is {26s}.
2026-02-16T22:15:29.874528883Z F0216 22:15:29.874507       1 cmd.go:182] open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied
```
### Note on AI assistance
I tried initially using AI but it suspected the kernel and not cri-o. But I was pretty sure that cri-o restart was causing the issue. Then I added logs for container creation and capture the config.json during the creation and crash looping. That helped me find the issue.
AI was helpful in find some in-built functions and test case while I was writing the fix.

### Current Flow
Here's the current flow of code (AI didn't help)
- `UserNsMode()` - Deprecated - This is set only when there is an annotation
- `UserNsPath()` - This is the real path where the user namespace is created. If host namespace is used this will return empty string
- `DefaultIDMappings` - Applied to all the pods if set on the crio.config but will be overridden by the pod mappings
- `sandbox_run_linux.go:getSandboxIDMappings()` - This is lazy loading the ID mappings
- `drop_infra_ctr` set to true - Spoofed Containers - Pause container is not created to keep the linux namespace alive. cri-o uses pinns and creates bind mounts

### Reason why cri-o restart affects
- The namespace mappings are not loaded on restart for running containers. Lazy load doesn't work in this case

Some Observations:
1. From OpenShift 4.20 user namespaces are used by etcd operator pods. The issue does not occur in OCP 4.19 where user namespaces are disabled
2. The cri-o restart is a required step in the recreate process. The pod should be running before the cri-o is restarted

During the crashloop backoff I have been able to capture the cat /run/containers/storage/overlay-containers/$CONTAINER_ID/userdata/config.json and I observed this missing section (after linux):
```
	"linux": {
		"uidMappings": [
			{
				"containerID": 0,
				"hostID": 3417243648,
				"size": 65536
			}
		],
		"gidMappings": [
			{
				"containerID": 0,
				"hostID": 3417243648,
				"size": 65536
			}
		],
```

The Projected volumes is setup but above missing part indicated that something is wrong. Also confirmed that a new pod has all these sections

I initially thought this should fix: https://github.com/cri-o/cri-o/pull/9712. But it didn't help

@Chandan9112 If we are backporting this, we need to include https://github.com/cri-o/cri-o/pull/9712

```release-note
Fix user namespace mappings persistence across CRI-O restarts
```

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust handling of Linux user-namespace UID/GID mappings for sandboxes and infra containers, with clearer error context and ensured application of mappings when available to improve container startup consistency.
* **Tests**
  * Added integration tests validating UID/GID mapping persistence across service restarts and across container delete/recreate flows, plus a new userns-configured sandbox test fixture.

### DIFF
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -381,35 +381,28 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
 		return nil, errors.New("infra container not found")
 	}
 
-	icPid, err := ic.Pid()
+	if !ic.Spoofed() && s.defaultIDMappings == nil {
+		return nil, nil
+	}
 
-	var (
-		uids, gids []spec.LinuxIDMapping
-	)
+	var uids, gids []spec.LinuxIDMapping
 
-	// Treat PID errors the same as PID 0 - both indicate we should read from config.json
-	if err != nil || icPid == 0 {
-		// read the UID/GID mappings from the infra container's saved config.json
-		configPath := filepath.Join(ic.Dir(), "config.json")
+	// If infra container is spoofed (drop_infra_ctr=true), read mappings from namespace options
+	if ic.Spoofed() {
+		usernsOpts := sb.NamespaceOptions().GetUsernsOptions()
 
-		configData, readErr := os.ReadFile(configPath)
-		if readErr != nil {
-			return nil, fmt.Errorf("cannot read infra container config.json: %w", readErr)
-		}
+		uids = getOCIMappings(usernsOpts.GetUids())
+		gids = getOCIMappings(usernsOpts.GetGids())
 
-		var containerConfig spec.Spec
-		if parseErr := json.Unmarshal(configData, &containerConfig); parseErr != nil {
-			return nil, fmt.Errorf("cannot parse infra container config.json: %w", parseErr)
-		}
-
-		if containerConfig.Linux == nil ||
-			(len(containerConfig.Linux.UIDMappings) == 0 && len(containerConfig.Linux.GIDMappings) == 0) {
+		if len(uids) == 0 || len(gids) == 0 {
 			return nil, nil
 		}
-
-		uids = containerConfig.Linux.UIDMappings
-		gids = containerConfig.Linux.GIDMappings
 	} else {
+		icPid, err := ic.Pid()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get infra container PID: %w", err)
+		}
+
 		uids, gids, err = unshare.GetHostIDMappings(strconv.Itoa(icPid))
 		if err != nil {
 			return nil, fmt.Errorf("failed to get ID mappings from infra container PID %d: %w", icPid, err)
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -372,8 +372,7 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
 		}
 	}
 
-	// UsernsMode() is empty but UserNsPath() is set when the cri-o restarts and container is killed
-	if sb.UsernsMode() == "" && sb.UserNsPath() == "" && s.defaultIDMappings == nil {
+	if sb.UsernsMode() == "" && s.defaultIDMappings == nil {
 		return nil, nil
 	}
 
@@ -381,32 +380,9 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
 		return nil, errors.New("infra container not found")
 	}
 
-	if !ic.Spoofed() && s.defaultIDMappings == nil {
-		return nil, nil
-	}
-
-	var uids, gids []spec.LinuxIDMapping
-
-	// If infra container is spoofed (drop_infra_ctr=true), read mappings from namespace options
-	if ic.Spoofed() {
-		usernsOpts := sb.NamespaceOptions().GetUsernsOptions()
-
-		uids = getOCIMappings(usernsOpts.GetUids())
-		gids = getOCIMappings(usernsOpts.GetGids())
-
-		if len(uids) == 0 || len(gids) == 0 {
-			return nil, nil
-		}
-	} else {
-		icPid, err := ic.Pid()
-		if err != nil {
-			return nil, fmt.Errorf("failed to get infra container PID: %w", err)
-		}
-
-		uids, gids, err = unshare.GetHostIDMappings(strconv.Itoa(icPid))
-		if err != nil {
-			return nil, fmt.Errorf("failed to get ID mappings from infra container PID %d: %w", icPid, err)
-		}
+	uids, gids, err := unshare.GetHostIDMappings(strconv.Itoa(ic.State().Pid))
+	if err != nil {
+		return nil, err
 	}
 
 	mappings := convertToStorageIDMappings(uids, gids)
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -372,7 +372,8 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
 		}
 	}
 
-	if sb.UsernsMode() == "" && s.defaultIDMappings == nil {
+	// UsernsMode() is empty but UserNsPath() is set when the cri-o restarts and container is killed
+	if sb.UsernsMode() == "" && sb.UserNsPath() == "" && s.defaultIDMappings == nil {
 		return nil, nil
 	}
 
@@ -380,9 +381,39 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
 		return nil, errors.New("infra container not found")
 	}
 
-	uids, gids, err := unshare.GetHostIDMappings(strconv.Itoa(ic.State().Pid))
-	if err != nil {
-		return nil, err
+	icPid, err := ic.Pid()
+
+	var (
+		uids, gids []spec.LinuxIDMapping
+	)
+
+	// Treat PID errors the same as PID 0 - both indicate we should read from config.json
+	if err != nil || icPid == 0 {
+		// read the UID/GID mappings from the infra container's saved config.json
+		configPath := filepath.Join(ic.Dir(), "config.json")
+
+		configData, readErr := os.ReadFile(configPath)
+		if readErr != nil {
+			return nil, fmt.Errorf("cannot read infra container config.json: %w", readErr)
+		}
+
+		var containerConfig spec.Spec
+		if parseErr := json.Unmarshal(configData, &containerConfig); parseErr != nil {
+			return nil, fmt.Errorf("cannot parse infra container config.json: %w", parseErr)
+		}
+
+		if containerConfig.Linux == nil ||
+			(len(containerConfig.Linux.UIDMappings) == 0 && len(containerConfig.Linux.GIDMappings) == 0) {
+			return nil, nil
+		}
+
+		uids = containerConfig.Linux.UIDMappings
+		gids = containerConfig.Linux.GIDMappings
+	} else {
+		uids, gids, err = unshare.GetHostIDMappings(strconv.Itoa(icPid))
+		if err != nil {
+			return nil, fmt.Errorf("failed to get ID mappings from infra container PID %d: %w", icPid, err)
+		}
 	}
 
 	mappings := convertToStorageIDMappings(uids, gids)
--- a/server/sandbox_run_linux.go
+++ b/server/sandbox_run_linux.go
@@ -28,6 +28,7 @@ import (
 	"github.com/cri-o/cri-o/internal/annotations"
 	"github.com/cri-o/cri-o/internal/config/nsmgr"
 	ctrfactory "github.com/cri-o/cri-o/internal/factory/container"
+	"github.com/cri-o/cri-o/internal/lib"
 	"github.com/cri-o/cri-o/internal/lib/constants"
 	libsandbox "github.com/cri-o/cri-o/internal/lib/sandbox"
 	"github.com/cri-o/cri-o/internal/linklogs"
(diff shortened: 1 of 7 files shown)


ngopalak-redhat:
@coderabbitai review

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,46 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
 
-	uids, gids, err := unshare.GetHostIDMappings(strconv.Itoa(ic.State().Pid))
-	if err != nil {
-		return nil, err
+	icPid := ic.State().Pid
+
+	var uids, gids []spec.LinuxIDMapping
+
+	var err error
+
```
saschagrunert: nit: combine into a single `var` block

```suggestion
	var (
		uids, gids []spec.LinuxIDMapping
		err        error
	)
```

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,46 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
+			return nil, fmt.Errorf("cannot read infra container config.json: %w", readErr)
+		}
+
+		var containerConfig spec.Spec
+		if parseErr := json.Unmarshal(configData, &containerConfig); parseErr != nil {
+			return nil, fmt.Errorf("cannot parse infra container config.json: %w", parseErr)
+		}
+
+		if containerConfig.Linux == nil || len(containerConfig.Linux.UIDMappings) == 0 {
+			return nil, nil
```
saschagrunert: This only checks `UIDMappings`. CRI-O always sets both together, so it won't cause issues today, but for consistency you could guard on both:

```suggestion
		if containerConfig.Linux == nil ||
			(len(containerConfig.Linux.UIDMappings) == 0 && len(containerConfig.Linux.GIDMappings) == 0) {
```

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,46 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
-	if err != nil {
-		return nil, err
+	icPid := ic.State().Pid
+
+	var uids, gids []spec.LinuxIDMapping
+
+	var err error
+
+	if icPid == 0 {
+		// read the UID/GID mappings from the infra container's saved config.json
```
saschagrunert: This fallback path needs an integration test: create a user namespace pod, restart CRI-O (or kill the infra container), then verify that containers in the pod still get correct ID mappings. The `integration / userns` suite would be the natural place. PR #9712 added a test for its change; this one should too.

saschagrunert (commented):
Core fix is sound, both parts address a real bug. Main gap is test coverage for the CRI-O restart scenario. A few inline suggestions.

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,46 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
 	}
 
 	if ic == nil {
 		return nil, errors.New("infra container not found")
 	}
 
-	uids, gids, err := unshare.GetHostIDMappings(strconv.Itoa(ic.State().Pid))
-	if err != nil {
-		return nil, err
+	icPid := ic.State().Pid
```
haircommander: you can use `ic.Pid()` here

server/sandbox_run_linux.go
haircommander: userns suite actually runs cri-o in a userns so that wouldn't be the right place. the normal integration tests should be able to cover it though

haircommander:
so the infra container has to be killed and cri-o needs to restart? I'd expect cri-o to cleanup the infra container and the pod in that case, and the kubelet to start from a clean state.

ngopalak-redhat:
> so the infra container has to be killed and cri-o needs to restart? I'd expect cri-o to cleanup the infra container and the pod in that case, and the kubelet to start from a clean state.

@haircommander I had some mis-understanding about the pause container's process Id. I was of the assumption that PID=1 is pause container. After looking into drop infra container enhancement I see that Pid=1 is the real container. When I kill (or crash) the process, its the real process (in case of the customer its the etcd operator). I have fixed the PR description also.

The kubelet does detect and restart (num restarts):
```
  oc get po -n openshift-etcd-operator
  NAME                             READY   STATUS    RESTARTS     AGE
  etcd-operator-64bc54d7dc-lt7wm   1/1     Running   1 (9s ago)   2m19s
```

I was of the assumption that there is nothing cached in the cri-o with respect to user namespaces and cri-o will act as a pass through layer which will send the user namespaces options directly to runc/crun. But `userns-mode.crio.io` annotation is not stored in the config.json, hence sb.UsernsMode() returns "". But sb.UserNsPath() IS set, which I found while debugging.

Hence logic of the fix still holds good.

server/sandbox_run_linux.go
ngopalak-redhat: I have added the tests

server/sandbox_run_linux.go
ngopalak-redhat: I added the checks for both UID and GID

test/ctr_userns.bats
```diff
@@ -27,3 +27,36 @@ function teardown() {
 	[[ "$out" == *"100000"* ]]
 	[[ "$out" == *"200000"* ]]
 }
+
+# This test follows a similar pattern to the one above but adds
+# an adds extra step of crio restart.
```
bitoku:
```suggestion
# This test follows a similar pattern to the one above but adds
# an extra step of crio restart.
```

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,48 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
 		}
 	}
 
-	if sb.UsernsMode() == "" && s.defaultIDMappings == nil {
+	// UsernsMode() is empty but UserNsPath() is set when the cri-o restarts and container is killed
+	if sb.UsernsMode() == "" && sb.UserNsPath() == "" && s.defaultIDMappings == nil {
```
bitoku: Isn't it possible to fix UsernsMode handling so that it keeps the value after cri-o restarts?

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,48 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
-	if err != nil {
-		return nil, err
+	icPid, err := ic.Pid()
+
+	var (
+		uids, gids []spec.LinuxIDMapping
+	)
+
+	// Treat PID errors the same as PID 0 - both indicate we should read from config.json
+	if err != nil || icPid == 0 {
```
bitoku: icPid == 0 means spoofed container (`drop_infra_ctr=true`)

I'd rather check whether it's spoofed or not.

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,48 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
+		}
+
+		var containerConfig spec.Spec
+		if parseErr := json.Unmarshal(configData, &containerConfig); parseErr != nil {
+			return nil, fmt.Errorf("cannot parse infra container config.json: %w", parseErr)
+		}
+
+		if containerConfig.Linux == nil ||
+			(len(containerConfig.Linux.UIDMappings) == 0 && len(containerConfig.Linux.GIDMappings) == 0) {
+			return nil, nil
```
bitoku: I'm wondering if we should load this in `ic.idMappings`.

ngopalak-redhat:
@bitoku Thanks for the comment: https://github.com/cri-o/cri-o/pull/10014#discussion_r3381253345

I have a simpler fix now as I understood more about how usernamespace paths are handled by cri-o.

Basically what's missing is calling `sb.UserNsJoin(userNsPath)` when the path exists while cri-o starts. I'll keep the current fix in the older commit and create a new commit for this fix.

I'll hold this until I test it fully.

/hold

server/sandbox_run_linux.go
ngopalak-redhat: I understood how this populated. I tried another fix also:
https://github.com/cri-o/cri-o/compare/main...ngopalak-redhat:cri-o:ngopalak/fix-userns-mode-persistence
But that didn't help. If I try to fix during the cri-o restart, I can see that UsernsMode is correctly set. The real problem occurs when invoking `unshare.GetHostIDMappings(strconv.Itoa(icPid))` . After cri-o restart this doesn't work. I have changed the logic to get it from the NameSpaceOptions of the sandbox container.

server/sandbox_run_linux.go
ngopalak-redhat: Fixed by using the spoof method

server/sandbox_run_linux.go
ngopalak-redhat: Fixed by loading from NameSpaceOptions in Sandbox

ngopalak-redhat:
Fixed by loading the UID and GID mapping from sandbox
/unhold

test/testdata/sandbox_config_userns.json
```diff
@@ -0,0 +1,65 @@
+{
```
haircommander: we did a lot of reducing the number of files in this dir and just use jq to make new ones as needed. is it sufficiently annoying to do so here to warrant a new file?

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,60 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
+		}
+
+		gids = make([]spec.LinuxIDMapping, len(criGids))
+		for i, m := range criGids {
+			gids[i] = spec.LinuxIDMapping{
+				ContainerID: m.GetContainerId(),
+				HostID:      m.GetHostId(),
+				Size:        m.GetLength(),
+			}
+		}
```
bitoku: You can reuse getOCIMappings at container_create_linux.go:629

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,60 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
-	if err != nil {
-		return nil, err
+	var uids, gids []spec.LinuxIDMapping
+
+	// If infra container is spoofed (drop_infra_ctr=true), read mappings from namespace options
+	if ic.Spoofed() {
+		nsOpts := sb.NamespaceOptions()
+		if nsOpts == nil || nsOpts.GetUsernsOptions() == nil {
+			return nil, nil
+		}
```
bitoku: I think this can be removed because grpc Get* methods are nil safe.

test/ctr_userns.bats
bitoku: I think we can remove this check? The tests should pass without it, correct?

server/sandbox_run_linux.go
ngopalak-redhat: Fixed

server/sandbox_run_linux.go
ngopalak-redhat: Thanks, I used getOCIMappings

test/testdata/sandbox_config_userns.json
ngopalak-redhat: Thanks. I didn't know about this pattern. I have removed this new file

test/ctr_userns.bats
ngopalak-redhat: Yes, I have removed this check

ngopalak-redhat:
Thanks @bitoku and @haircommander . I have addressed the above review comments

ngopalak-redhat:
@bitoku As this merged https://github.com/cri-o/cri-o/pull/10025 , can you help trigger the integration tests on this PR

ngopalak-redhat:
Fixed the failure from integration tests

server/sandbox_run_linux.go
```diff
@@ -372,17 +372,41 @@ func (s *Server) getSandboxIDMappings(ctx context.Context, sb *libsandbox.Sandbo
...
 	if ic == nil {
 		return nil, errors.New("infra container not found")
 	}
 
-	uids, gids, err := unshare.GetHostIDMappings(strconv.Itoa(ic.State().Pid))
-	if err != nil {
-		return nil, err
+	if !ic.Spoofed() && s.defaultIDMappings == nil {
+		return nil, nil
+	}
```
bitoku: What is this for? This won't work when the pod has its own mappings , but no default mappings.

bitoku (changes requested):

server/sandbox_run_linux.go
ngopalak-redhat: Without defaultIDMappings checked there were test case failures:
"systemd cgroup manager uses system dbus when running as root with rootless env"
when its run with `TEST_USERNS=0`

Also the e2e test was failing:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/cri-o_cri-o/10014/pull-ci-cri-o-cri-o-main-ci-e2e/2065248237823987712

After some investigation I found when no explicit uid_mappings/gid_mappings are configured in CRI-O config (i.e., s.defaultIDMappings == nil), we shouldn't try to use mappings from /proc, even if they exist. Those are just rootless defaults, not intentional user namespace configuration.

bitoku: Can you elaborate on "we shouldn't try to use mappings from /proc, even if they exist. Those are just rootless defaults, not intentional user namespace configuration." ?

What if infra_container has mappings and no defaultIDMappings? Isn't the mappings in infra_container "intentional user namespace configuration"?

Also I don't see any relations between this and the test failures.

I think we need e2e tests for drop_infra_ctr=false/true .

ngopalak-redhat: The error from the cri-o test log:
`Error: container create failed: setns to target user namespace `/proc/169279/ns/user`: Invalid argument`

Previously `s.defaultIDMappings == nil` check was present along with the mode check. defaultIDMappings is a global configuration for user namespace ID mappings. My understanding is that it comes from the CRI-O config file (/etc/crio/crio.conf):

  [crio.runtime]
  uid_mappings = "0:100000:65536"
  gid_mappings = "0:200000:65536"

  Or via environment variables:
  CONTAINER_UID_MAPPINGS="0:100000:65536"
  CONTAINER_GID_MAPPINGS="0:200000:65536"

This being nil means no userns configuration - user namespaces were NOT configured in CRI-O config.

But I will write e2e tests for drop_infra_ctr=false/true and update this.

haircommander: note server level uidmapping is to run cri-o rootless, whereas the correct way to make a pod user namespaced in the general case is to use hostUsers. in the vast majority of cri-o deployments (and every openshfit one) we don't use server level user namespace (cri-o runs rootful)

ngopalak-redhat:
/hold
Working on the comments

ngopalak-redhat:
@bitoku I'll fix the linter error. Can you review the new approach I have taken? The fix simply loads the id mappings at the time of starting from the spec. So now I don't need to change anything while container is created.

bitoku (commented):
Why did you remove integration tests? I think they were useful.

ngopalak-redhat:
The tests are working when I pinned the cri-tools to 1.36 version. I have to wait for the cri-tools issue to resolve 
@bitoku I have added the integration tests and also added tests for drop_infra_ctr=true/false.

I'll keep this on hold until this is merged: https://github.com/cri-o/cri-o/pull/10042

ngopalak-redhat:
@haircommander @bitoku @saschagrunert Now that the test suite is stable, this one is good for review. Can you please review?

haircommander:
/retest
/approve

good research and work here

PR #10024, opened 2026-06-09
Fix OCPBUGS-81765:  CVE-2026-34986 Go JOSE: Denial of Service

germanparente:
rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

#### What type of PR is this?

/kind dependency-change
Fix OCPBUGS-81765:  CVE-2026-34986 Go JOSE: Denial of Service

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -191,7 +191,7 @@ require (
 	github.com/go-git/gcfg v1.5.0 // indirect
 	github.com/go-git/go-billy/v5 v5.4.1 // indirect
 	github.com/go-git/go-git/v5 v5.6.1 // indirect
-	github.com/go-jose/go-jose/v3 v3.0.3 // indirect
+	github.com/go-jose/go-jose/v3 v3.0.5 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-ole/go-ole v1.2.6 // indirect
 	github.com/go-openapi/analysis v0.21.4 // indirect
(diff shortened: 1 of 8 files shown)


germanparente:
All three CI failures (`e2e-agnostic`, `e2e-gcp`, `images`) share the same root cause — a pre-existing infrastructure issue on the `release-1.26` branch:

```
sed: can't read /etc/yum.repos.d/CentOS-Sources.repo: No such file or directory
[ERROR] hack/build-rpms.sh:41: `sed -i 's|//vault.centos.org/centos/|//archive.kernel.org/centos-vault/centos/|g' /etc/yum.repos.d/CentOS-Sources.repo` exited with status 2.
```

The builder image no longer contains `/etc/yum.repos.d/CentOS-Sources.repo`, so `hack/build-rpms.sh` fails before any code from this PR is exercised. This is unrelated to the CVE fix.

For reference, all recently merged PRs on `release-1.26` have the same three failures (#10157, #9605) and were merged successfully.

germanparente:
There's a particular case here. 

To bump to 3.0.5, the module is using byte.Clone() that is not present in 1.19. For this particular case, I will modify the vendored source file.

bitoku:
I feel we should bump the go version rather than modifying the vendored file.
The reason is that if someone who pretends to be a person from redhat tries to add malicious code in the vendor, there's no automatic check to verify the vendor. Right now, we have [verify / vendor (pull_request)](https://github.com/cri-o/cri-o/actions/runs/32352596862/job/96374810517?pr=10024), but it won't work anymore if we have that modified vendor.
Or we should have a package without bytes.Clone in `openshift-sustaining` and replace it just like https://github.com/cri-o/cri-o/blob/release-1.31/go.mod#L240

@saschagrunert @haircommander any thought?

PR #10042, opened 2026-06-17
test: base64-encode env var values for cri-api v0.36.2 compatibility

bitoku:
The cri-api v0.36.2 changed KeyValue.Value from string to bytes in the protobuf definition. Go's encoding/json expects base64-encoded strings when unmarshaling []byte fields, so crictl now requires env var values in JSON config files to be base64-encoded.

Update all test container configs and BATS scripts to provide base64-encoded env var values.

Assisted-by: Claude Code <https://claude.com/claude-code>

#### What type of PR is this?

/kind failing-test

- https://github.com/kubernetes-sigs/cri-tools/issues/2127

```release-note
None
```

### DIFF
--- a/test/testdata/container_config.json
+++ b/test/testdata/container_config.json
@@ -13,23 +13,23 @@
   "envs": [
     {
       "key": "PATH",
-      "value": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+      "value": "L3Vzci9sb2NhbC9zYmluOi91c3IvbG9jYWwvYmluOi91c3Ivc2JpbjovdXNyL2Jpbjovc2JpbjovYmlu"
     },
     {
       "key": "TERM",
-      "value": "xterm"
+      "value": "eHRlcm0="
     },
     {
       "key": "GLIBC_TUNABLES",
-      "value": "glibc.pthread.rseq=0"
+      "value": "Z2xpYmMucHRocmVhZC5yc2VxPTA="
     },
     {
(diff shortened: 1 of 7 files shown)


test/testdata/container_config.json
```diff
@@ -13,23 +13,23 @@
   "envs": [
     {
       "key": "PATH",
-      "value": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
bartwensley, line 16: It would be great if this could be fixed without having to make the value unreadable. One option is to do the conversion in the jq command that is processing these json files - like this:
```
function create_container_config() {
	local name="$1"
	local outfile="$2"
	local cpuset="$3"
	jq --arg name "$name" --arg cpuset "$cpuset" \
		'.metadata.name = $name
		| .linux.resources.cpu_shares = 1024
		| .linux.resources.cpuset_cpus = $cpuset
		| .envs |= map(.value = (.value | @base64))' \
		"$TESTDATA"/container_sleep.json >"$outfile"
}
```

ngopalak-redhat:
@bitoku This is required to resolve the build issues.
Thinking if it would be better to pin cri-tools to the last released version (1.36). That will prevent build breaks while working on other PRs.

bitoku:
@ngopalak-redhat 
yeah, just pinning is fine for now, I guess.
Could you do it?

bitoku:
opened a new one https://github.com/cri-o/cri-o/pull/10047 for temporary fix. Will discuss how we fix this later.

PR #10043, opened 2026-06-17
config: remove experimental markers from additional_artifact_stores

saschagrunert:
#### What type of PR is this?

/kind cleanup

Removes the "(experimental, subject to change)" markers from the `additional_artifact_stores` configuration option in the config template and man page for GA readiness.

None

```release-note
Remove experimental markers from additional_artifact_stores configuration option.
```

## Summary by CodeRabbit

* **Documentation**
  * Updated `crio.runtime.additional_artifact_stores` documentation to remove the “experimental, subject to change” wording.

* **Bug Fixes**
  * Runtime config validation for `additional_artifact_stores` now also verifies each configured absolute path exists on the filesystem (invalid/non-existent paths are rejected with clear errors).

* **Tests**
  * Expanded validation tests for empty, absolute-only success, and relative/mixed/non-existent failure cases.
  * Added unit and end-to-end coverage for additional-store precedence, fallback behavior, removal-from-main-only, pull skip/deduplication when artifacts exist in additional stores.

### DIFF
--- a/internal/ociartifact/store_test.go
+++ b/internal/ociartifact/store_test.go
@@ -12,8 +12,11 @@ import (
 	"github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
+	"go.podman.io/common/libimage"
 	"go.podman.io/common/pkg/libartifact"
+	"go.podman.io/image/v5/docker/reference"
 	"go.podman.io/image/v5/manifest"
+	"go.podman.io/image/v5/types"
 	"go.uber.org/mock/gomock"
 
 	"github.com/cri-o/cri-o/internal/ociartifact"
@@ -23,6 +26,37 @@ import (
 var errTest = errors.New("test")
 
 const testDigest = "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+
+type fakeImageRef struct {
+	ref reference.Named
+}
+
+func newFakeImageRef() *fakeImageRef {
+	ref, err := reference.ParseNormalizedNamed("docker.io/library/nginx:latest")
+	if err != nil {
+		panic(err)
+	}
+
+	return &fakeImageRef{ref: reference.TagNameOnly(ref)}
+}
+
+func (f *fakeImageRef) Transport() types.ImageTransport         { return nil }
+func (f *fakeImageRef) StringWithinTransport() string           { return f.ref.String() }
+func (f *fakeImageRef) DockerReference() reference.Named        { return f.ref }
+func (f *fakeImageRef) PolicyConfigurationIdentity() string     { return "" }
+func (f *fakeImageRef) PolicyConfigurationNamespaces() []string { return nil }
+func (f *fakeImageRef) NewImage(_ context.Context, _ *types.SystemContext) (types.ImageCloser, error) {
+	return nil, nil
+}
+
+func (f *fakeImageRef) NewImageSource(_ context.Context, _ *types.SystemContext) (types.ImageSource, error) {
+	return nil, nil
+}
+
+func (f *fakeImageRef) NewImageDestination(_ context.Context, _ *types.SystemContext) (types.ImageDestination, error) {
+	return nil, nil
+}
+func (f *fakeImageRef) DeleteImage(_ context.Context, _ *types.SystemContext) error { return nil }
 
 func makeOCIManifest(configMediaType, artifactType string) []byte {
 	m := map[string]any{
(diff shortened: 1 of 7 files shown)


saschagrunert:
@cri-o/cri-o-maintainers PTAL

bitoku:
I'd say it lacks some test cases.
This is AI generated, and some aren't necessary, but it's better to take a look.


  NewStore()

  - [ ] Creates store with valid additional paths
  - [ ] Skips unreachable additional path with warning, still returns valid store
  - [ ] Skips additional path with missing artifacts/ subdirectory
  - [ ] Handles mix of valid and invalid additional paths (good ones kept)
  - [ ] Works with nil additional paths (no-op, existing behavior preserved)
  - [ ] Works with empty []string{} additional paths

  Pull()

  - [ ] Skips pull when artifact exists in additional store, returns correct digest
  - [ ] Falls through to normal pull when NewArtifactStorageReference fails
  - [ ] Falls through to normal pull when artifact not in any additional store
  - [ ] Checks additional stores in order (first match wins)
  - [ ] Does not skip pull when artifact exists only in main store (re-pull allowed for tag re-pointing)
  - [ ] Pull with tagless reference vs additional store that stored with :latest (reference normalization mismatch)

  List()

  - [ ] Returns artifacts from additional store when main store is empty
  - [ ] Returns artifacts from main store when no additional stores configured
  - [ ] Merges artifacts from additional and main stores
  - [ ] Deduplicates by reference — additional store wins over main store
  - [ ] Continues listing main store when additional store returns error
  - [ ] Logs warning when additional store listing fails
  - [ ] Returns correct RootPath() on each artifact (additional vs main)
  - [ ] Artifacts from additional stores are marked as pinned

  Status()

  - [ ] Finds artifact in additional store (checked first)
  - [ ] Falls back to main store when not in additional stores
  - [ ] Returns ErrNotFound-wrapped error when not in any store
  - [ ] Logs debug for ErrNotFound in additional store, warns on other errors
  - [ ] Returns correct RootPath() for artifact from additional store
  - [ ] Artifact from additional store is marked as pinned

  Remove()

  - [ ] Removes artifact from main store successfully
  - [ ] Returns error when artifact exists only in additional store (not in main)
  - [ ] Does not attempt to remove from additional stores

  BlobMountPaths()

  - [ ] Uses artifact.RootPath() for artifact from main store
  - [ ] Uses artifact.RootPath() for artifact from additional store
  - [ ] Returns correct blob paths when artifact is in additional store

  ---
  Unit Tests — pkg/config/config_test.go

  - [ ] Validates additional_artifact_stores with empty list []
  - [ ] Validates with valid absolute paths
  - [ ] Rejects relative path (returns error quoting the path)
  - [ ] Rejects mix of absolute and relative paths (fails on first invalid)

  ---
  Unit Tests — internal/ociartifact/datastore/store_test.go

  - [ ] New() passes nil for additional paths (no additional store behavior leaks into datastore)
  - [ ] artifactData() uses artifact.RootPath() for layout reference (not store root)

  ---
  Integration Tests — test/artifact_additional_stores.bats

  Error handling

  - [ ] Should warn and start when additional store path does not exist
  - [ ] Should warn and start when additional store has no artifacts/ subdirectory
  - [ ] Should return meaningful error when removing artifact that exists only in additional store

  Multiple stores

  - [ ] Should support multiple additional stores in config
  - [ ] Should prioritize first additional store when artifact exists in multiple stores
  - [ ] Should list artifacts from all additional stores

  Deduplication and priority

  - [ ] Should deduplicate when same artifact exists in additional and main store
  - [ ] Should serve additional store version when same tag exists in both stores

  Lifecycle

  - [ ] Should serve artifacts from additional store after CRI-O restart
  - [ ] Should pick up new artifact after it is added to additional store filesystem (no restart needed, or document restart required)
  - [ ] Should stop serving artifact after it is removed from additional store filesystem

  Edge cases

  - [ ] Should handle empty additional store (no artifacts in it)
  - [ ] Should handle additional store becoming unavailable after startup (e.g., NFS unmount)

saschagrunert:
Thanks for the thorough list! I went through it and added test coverage where it made sense. Here's the breakdown:

**Covered by new unit tests (internal/ociartifact/store_test.go):**

- List: additional store when main empty, no additional stores configured, merge, dedup (additional wins), error resilience, multiple stores, pinning, RootPath
- Status: priority lookup, fallback, ErrNotFound wrapping, non-ErrNotFound warning/fallback, pinning, RootPath
- Remove: main-store-only removal, error propagation (ErrNotFound), no calls to additional stores (gomock strict)
- Pull: skip when in additional store (with digest assertion), fallthrough, first-match-wins ordering, no-skip when only in main store

**Covered by new config tests (pkg/config/config_test.go):**

- Empty list, valid absolute paths, relative path rejection, mixed paths rejection

**Covered by new BATS tests (test/artifact_additional_stores.bats):**

- Non-existent store path (warn and start), missing artifacts/ subdirectory, multiple stores in config, deduplication across stores

**Intentionally skipped:**

- BlobMountPaths: requires real OCI layout on disk, already covered by existing BATS mount test
- Datastore: explicitly passes nil for additional paths, nothing additional-store-specific to test
- Pull with NewArtifactStorageReference failure / tagless reference normalization: libart internals, not additional-stores logic
- Lifecycle tests (restart, hot-reload, NFS unmount): hot-reload is not implemented, restart/NFS scenarios are too infrastructure-heavy for this scope
- "Should return meaningful error when removing artifact only in additional store": Remove operates on the main store only, so this is the same as "artifact not found in main store" (covered)

internal/ociartifact/store_test.go
```diff
@@ -532,4 +566,420 @@ var _ = t.Describe("Store", func() {
...
+
+			store, err = ociartifact.NewStore(t.MustTempDir("artifact"), nil, nil, nil)
+			Expect(err).NotTo(HaveOccurred())
+			store.SetFakeStore(&ociartifact.FakeLibartifactStore{mainStoreMock})
+			store.SetFakeImpl(implMock)
+		})
+
+		AfterEach(func() {
+			mockCtrl.Finish()
+		})
```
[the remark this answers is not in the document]
saschagrunert: Should be fixed.

test/artifact_additional_stores.bats
```diff
@@ -127,6 +127,80 @@ EOF
...
+	crictl pull "$ARTIFACT_IMAGE"
+
+	# Copy to additional store (artifact now exists in both)
+	local main_store="$TESTDIR/crio/artifacts"
+	cp -a "$main_store" "$ADDITIONAL_STORE/"
+
+	# Listing should show the artifact exactly once
+	count=$(crictl images 2>/dev/null | grep -cE "$ARTIFACT_REPO.*singlefile" || true)
+	[[ "$count" -eq 1 ]]
+}
```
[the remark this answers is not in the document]
saschagrunert: Should be fixed.

saschagrunert:
@bitoku PTAL again

saschagrunert:
Rebased

pkg/config/config_test.go
```diff
@@ -248,6 +248,53 @@ var _ = t.Describe("Config", func() {
...
+			// When
+			err := sut.RuntimeConfig.Validate(nil, false)
+
+			// Then
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should succeed with valid absolute paths in additional_artifact_stores", func() {
+			// Given
+			sut.AdditionalArtifactStores = []string{"/mnt/nfs/store1", "/opt/artifacts"}
```
bitoku: What if the path doesn't exist?
I think we should reject it so that the user can know the misconfiguration.

internal/ociartifact/store_test.go
```diff
@@ -532,4 +566,446 @@ var _ = t.Describe("Store", func() {
...
+				arts, err := store.List(context.Background())
+
+				// Then
+				Expect(err).NotTo(HaveOccurred())
+				Expect(arts).To(HaveLen(1))
+				Expect(arts[0].CRIImage().GetPinned()).To(BeTrue())
+				Expect(arts[0].RootPath()).To(Equal(addStorePath1))
+			})
+
+			It("should continue listing main store when additional store returns error", func() {
```
bitoku, line 701: Is it safe to fallback to main when it gets any error?
What if the additional store is corrupted? When can users know that?

internal/ociartifact/store_test.go
```diff
@@ -532,4 +566,446 @@ var _ = t.Describe("Store", func() {
...
+				// Then
+				Expect(err).To(HaveOccurred())
+				Expect(errors.Is(err, ociartifact.ErrNotFound)).To(BeTrue())
+			})
+		})
+
+		t.Describe("Pull", func() {
+			artifactManifest := makeOCIManifest(specs.MediaTypeImageConfig, "application/vnd.test.artifact")
+
+			It("should skip pull when artifact exists in additional store and return correct digest", func() {
```
bitoku, line 885: Should it always pull instead of skip because kubelet already handles `IfNotPresent`?
How does `additionalimagestores` work?

pkg/config/config_test.go
saschagrunert: Good catch! Added an `os.Stat` check during config validation so non-existent paths are rejected early with a clear error.

internal/ociartifact/store_test.go
saschagrunert: The fallback is intentional here. Additional stores may reside on unreliable media like NFS, so we warn and continue to avoid blocking the main store. The warning log surfaces the issue for operators. This matches the same pattern used in `List()` where we also warn-and-continue on additional store errors.

internal/ociartifact/store_test.go
saschagrunert: The skip is intentional for additional stores only. They are read-only and externally managed, so re-pulling would just duplicate data into the main store. For the main store we always pull because the remote tag may have been re-pointed to a different digest. This is similar to how Podman's `additionalImageStores` work: read-only overlay stores (typically on shared/NFS storage) that provide content without pulling. Kubelet's `IfNotPresent` governs whether CRI-O gets a pull request at all. Once we do get one, this logic decides whether the artifact already exists locally in a read-only store.

saschagrunert:
@bitoku PTAL again

pkg/config/config_test.go
bitoku: This validation contradicts https://github.com/cri-o/cri-o/pull/10043#discussion_r3452392244

internal/ociartifact/store_test.go
bitoku: So what if the additional store is temporarily unavailable when pulling?
Could it pull the artifact and not use the one in the additional store?

pkg/config/config_test.go
saschagrunert: Removed the os.Stat check. Config validation now only rejects relative paths. Availability is handled at runtime with warn-and-continue, consistent with the NFS rationale from the other thread.

internal/ociartifact/store_test.go
saschagrunert: Yes, if the additional store is temporarily unavailable during pull, the Inspect fails and CRI-O pulls the artifact from the registry into the main store. When the additional store comes back, it takes priority again on subsequent Status/List calls (the main store copy is effectively shadowed). I added a warning log for this case so operators can see when a pull bypasses a failing additional store.

bitoku:
@saschagrunert Thank you for the clarification!
[verify / prettier (pull_request)](https://github.com/cri-o/cri-o/actions/runs/28374522491/job/84060558375?pr=10043) this needs to be addressed, but other than that lgtm.

bitoku:
prettier is being fixed in a different pr
/lgtm

saschagrunert:
Same here, label plugin does not work.

PR #10060, opened 2026-06-21
Set default unprivileged port start to 0 for non-host-network pods

### LINKED ISSUE #10050: Set net.ipv4.ip_unprivileged_port_start=0 by default
Containerd 2.0 now sets `net.ipv4.ip_unprivileged_port_start=0` by default for pods with new network namespaces ([containerd/containerd#11502](https://github.com/containerd/containerd/pull/11502)), allowing containers to bind to privileged ports without `CAP_NET_BIND_SERVICE`. CRI-O does not currently do this — users must manually configure it via `default_sysctls` in the CRI-O config.

**Why this matters:**
- Removes a common reason pods run with unnecessary capabilities (e.g. `NET_BIND_SERVICE`) or as root
- `net.ipv4.ip_unprivileged_port_start` has been a safe sysctl in Kubernetes since v1.22 ([kubernetes/kubernetes#103298](https://github.com/kubernetes/kubernetes/issues/103298))
- Docker has set this by default since 20.10.0 ([moby/moby#41030](https://github.com/moby/moby/pull/41030))
- Containerd 2.0 now sets this by default
- CRI-O is the remaining major runtime that does not

**Proposed behavior:**
Set `net.ipv4.ip_unprivileged_port_start=0` by default for containers with new network namespaces (skip for host network), matching containerd/Docker behavior.

**Requirements:**
- Kernel 4.11+ (the sysctl was introduced there)

**References:**
- Upstream Kubernetes issue: https://github.com/kubernetes/kubernetes/issues/102612
- Previous CRI-O discussion: https://github.com/cri-o/cri-o/issues/6706
- Containerd implementation: https://github.com/containerd/containerd/pull/11502
- Docker implementation: https://github.com/moby/moby/pull/41030

amritansh1502:
#### What type of PR is this?
/kind feature

Defaults `net.ipv4.ip_unprivileged_port_start` to `0` for pods that have their own network namespace, allowing non-root processes inside containers to bind to any port without manual sysctl configuration.

Fixes #10050

Added unit test cases also for this updation.

```release-note
Non-root containers can now bind to any port by default without additional sysctl configuration.

```

## Summary by CodeRabbit

* **New Features**
  * Automatically defaults `net.ipv4.ip_unprivileged_port_start` to `0` for pods using isolated network namespaces, without changing values set via sysctls or affecting host-networked pods.

* **Tests**
  * Added unit tests covering defaulting vs user-provided sysctls across host networking scenarios.
  * Expanded integration (Bats) tests to verify in-container port binding behavior for defaulted, overridden, and host-network cases.

* **Documentation**
  * Documented the default sysctl behavior and how to override it per-pod or via `default_sysctls`.


ngopalak-redhat:
@amritansh1502  Can you try adding an e2e test?

amritansh1502:
Added three e2e test cases 
1.Default net.ipv4.ip_unprivileged_port_start=0 is applied for pods with their own network namespace.
2. net.ipv4.ip_unprivileged_port_start is not touched when pod uses host network (hostNetwork=true).
3. User-supplied value for net.ipv4.ip_unprivileged_port_start is not overridden by CRI-O's default.

test/pod.bats
```diff
@@ -292,6 +292,57 @@ EOF
 	[[ "$output" != *"net.ipv4.ip_forward = 0"* ]]
 }
 
+@test "default net.ipv4.ip_unprivileged_port_start to 0 for pod with own network namespace" {
```
ngopalak-redhat: A e2e test needs to use this feature net.ipv4.ip_unprivileged_port_start.
A completely unprivileged user should be able to bind to low ports like 80, 443, or 53 without requiring root privileges or the CAP_NET_BIND_SERVICE capability.

May be create a process and bind to this port before and after this change would be a good test

Also test this with user namespaces enabled.

ngopalak-redhat (changes requested):

ngopalak-redhat:
This is a user facing change and needs doc change. containerd has docs here: https://github.com/containerd/containerd/pull/11502/changes

test/pod.bats
amritansh1502: Added 3 functional e2e tests 
1. A non-root process  successfully binds to port 80 with the default ip_unprivileged_port_start=0  no CAP_NET_BIND_SERVICE required. (Improved 1st test case)
2. The same non-root process fails to bind to port 80 when ip_unprivileged_port_start=1024 is explicitly set  showing the "before" behavior.
3. A non-root process  inside a pod with user namespaces enabled also successfully binds to port 80.

ngopalak-redhat: Can you manually test to make sure that these tests fail without the changes in the PR?

amritansh1502: Without the changes i have tested manually:- 

1.  uid 65534 cannot bind to port 80 , Permission denied as expected.

Screenshot From 2026-06-24 19-59-47
2. With ip_unprivileged_port_start=1024 explicitly set: uid 65534 cannot bind to port 80  confirms 1024 is the blocker.

Screenshot From 2026-06-24 20-03-25
3.User-supplied ip_unprivileged_port_start=512 is visible inside the container  CRI-O does not override it.

Screenshot From 2026-06-24 20-04-00

4.Pod with hostNetwork=true shows the host's sysctl value  CRI-O did not modify it.
Screenshot From 2026-06-24 20-00-01

5.Pod with userNS enabled shows sysctl=1024 without PR changes  confirms feature also needs to work in userNS.

Screenshot From 2026-06-24 20-05-33

test/pod.bats
```diff
@@ -292,6 +292,126 @@ EOF
...
+
+	jq '.linux.security_context.run_as_user = {"value": 65534}' \
+		"$TESTDATA"/container_sleep.json > "$TESTDIR"/container.json
+
+	pod_id=$(crictl runp "$TESTDIR"/sandbox.json)
+	ctr_id=$(crictl create "$pod_id" "$TESTDIR"/container.json "$TESTDIR"/sandbox.json)
+	crictl start "$ctr_id"
+
+	# Binding to port 80 must fail because non-root cannot bind to ports below 1024
+	run ! crictl exec --sync "$ctr_id" python3 -c "
```
ngopalak-redhat: Can you explore some ways to avoid depending on python installation?

amritansh1502: @ngopalak-redhat , can we use `bash -c "echo > /dev/tcp/0.0.0.0/80"`  instead of python script ?, it is built in feature .

amritansh1502: I have added this `bash -c "echo > /dev/tcp/0.0.0.0/80"`  it is not depending on any installation

server/sandbox_run_linux.go
```diff
@@ -892,6 +906,22 @@ func (s *Server) configureGeneratorForSysctls(ctx context.Context, g *generate.G
...
 	}
 
+	// Default net.ipv4.ip_unprivileged_port_start to 0 so non-root processes
+	// inside the container can bind to any port. Only applied when the pod has
+	// its own network namespace and the user has not explicitly set the value.
+	const (
+		unprivilegedPortStartKey     = "net.ipv4.ip_unprivileged_port_start"
+		unprivilegedPortStartDefault = "0"
+	)
+	if !hostNetwork {
```
ngopalak-redhat: Small nit:
```
_, alreadySet := sysctlsToReturn[unprivilegedPortStartKey]
_, userSet := sysctls[unprivilegedPortStartKey]

if !hostNetwork && !alreadySet && !userSet {
```

ngopalak-redhat:
@bitoku @haircommander Please review

bitoku:
@amritansh1502 can you update the release-note in the PR description?

bitoku:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/10060/pull-ci-cri-o-cri-o-main-ci-cgroupv2-integration/2070071701495877632/artifacts/cgroupv2-integration/cri-o-gather/artifacts/testout.txt
```
not ok 398 non-root process can bind to privileged port with default ip_unprivileged_port_start
# (from function `crictl' in file ./helpers.bash, line 87,
#  in test file ./pod.bats, line 311)
#   `crictl exec --sync "$ctr_id" bash -c "echo > /dev/tcp/0.0.0.0/80"' failed
# time="2026-06-25T09:33:36.346913584Z" level=info msg="Updating config from single file: /etc/crio/crio.conf"
# time="2026-06-25T09:33:36.346936342Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf"
# time="2026-06-25T09:33:36.348115071Z" level=info msg="Updating config from path: /etc/crio/crio.conf.d"
# time="2026-06-25T09:33:36.348165611Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/01-ns-lifecycle.conf"
# time="2026-06-25T09:33:36.348265067Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/01-overlay.conf"
# time="2026-06-25T09:33:36.348306756Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/99-log-level.conf"
# time="2026-06-25T09:33:36.398965909Z" level=info msg="Updating config from path: "
# time="2026-06-25T09:33:36Z" level=fatal msg="validate service connection: validate CRI v1 runtime API for endpoint \"unix:///tmp/tmp.UDqWyFR5Wj/crio.sock\": rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /tmp/tmp.UDqWyFR5Wj/crio.sock: connect: no such file or directory\""
# fcfaf8560a6f4716189bf582affaf7c39055896a64027ec1b98c9217f8a424da
# time="2026-06-25T09:33:37Z" level=fatal msg="execing command in container fcfaf8560a6f4716189bf582affaf7c39055896a64027ec1b98c9217f8a424da synchronously: command 'bash -c echo > /dev/tcp/0.0.0.0/80' exited with 1: bash: connect: Connection refused\nbash: line 1: /dev/tcp/0.0.0.0/80: Connection refused\n"
```

server/sandbox_run_linux.go
amritansh1502: fixed, thanks

amritansh1502:
> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/cri-o_cri-o/10060/pull-ci-cri-o-cri-o-main-ci-cgroupv2-integration/2070071701495877632/artifacts/cgroupv2-integration/cri-o-gather/artifacts/testout.txt
> 
> ```
> not ok 398 non-root process can bind to privileged port with default ip_unprivileged_port_start
> # (from function `crictl' in file ./helpers.bash, line 87,
> #  in test file ./pod.bats, line 311)
> #   `crictl exec --sync "$ctr_id" bash -c "echo > /dev/tcp/0.0.0.0/80"' failed
> # time="2026-06-25T09:33:36.346913584Z" level=info msg="Updating config from single file: /etc/crio/crio.conf"
> # time="2026-06-25T09:33:36.346936342Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf"
> # time="2026-06-25T09:33:36.348115071Z" level=info msg="Updating config from path: /etc/crio/crio.conf.d"
> # time="2026-06-25T09:33:36.348165611Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/01-ns-lifecycle.conf"
> # time="2026-06-25T09:33:36.348265067Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/01-overlay.conf"
> # time="2026-06-25T09:33:36.348306756Z" level=info msg="Updating config from drop-in file: /etc/crio/crio.conf.d/99-log-level.conf"
> # level=info msg="Using default capabilities: CAP_CHOWN, CAP_DAC_OVERRIDE, CAP_FSETID, CAP_FOWNER, CAP_SETGID, CAP_SETUID, CAP_SETPCAP, CAP_NET_BIND_SERVICE, CAP_KILL"
> # time="2026-06-25T09:33:36.398965909Z" level=info msg="Updating config from path: "
> # level=info msg="Using default capabilities: CAP_CHOWN, CAP_DAC_OVERRIDE, CAP_FSETID, CAP_FOWNER, CAP_SETGID, CAP_SETUID, CAP_SETPCAP, CAP_NET_BIND_SERVICE, CAP_KILL"
> # time="2026-06-25T09:33:36Z" level=fatal msg="validate service connection: validate CRI v1 runtime API for endpoint \"unix:///tmp/tmp.UDqWyFR5Wj/crio.sock\": rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /tmp/tmp.UDqWyFR5Wj/crio.sock: connect: no such file or directory\""
> # fcfaf8560a6f4716189bf582affaf7c39055896a64027ec1b98c9217f8a424da
> # time="2026-06-25T09:33:37Z" level=fatal msg="execing command in container fcfaf8560a6f4716189bf582affaf7c39055896a64027ec1b98c9217f8a424da synchronously: command 'bash -c echo > /dev/tcp/0.0.0.0/80' exited with 1: bash: connect: Connection refused\nbash: line 1: /dev/tcp/0.0.0.0/80: Connection refused\n"
> ```
taken reference from existing test and fixed this by changing this `/dev/tcp/0.0.0.0/80` to  `output=$(crictl exec --sync "$ctr_id" sysctl net.ipv4.ip_unprivileged_port_start)
[[ "$output" == *"= 0"* ]]`

bitoku:
@amritansh1502 
> taken reference from existing test and fixed this by changing this /dev/tcp/0.0.0.0/80 to output=$(crictl exec --sync "$ctr_id" sysctl net.ipv4.ip_unprivileged_port_start) [[ "$output" == *"= 0"* ]]

I'd want to know why it failed rather than how to fix.
The check itself may be correct, but if we can check using the reserved ports is allowed, that'd be better.

amritansh1502:
> @amritansh1502
> 
> > taken reference from existing test and fixed this by changing this /dev/tcp/0.0.0.0/80 to output=$(crictl exec --sync "$ctr_id" sysctl net.ipv4.ip_unprivileged_port_start) [[ "$output" == _"= 0"_ ]]
> 
> I'd want to know why it failed rather than how to fix. The check itself may be correct, but if we can check using the reserved ports is allowed, that'd be better.

 @bitoku  `/dev/tcp`  failed because it performs a TCP connection not a bind , so when it tries to connect to port 80 as a client it gets connection refused regardless of the sysctl value.

  `output=$(crictl exec --sync "$ctr_id" sysctl net.ipv4.ip_unprivileged_port_start)
[[ "$output" == *"= 0"* ]]` i think this just tests that it is applied or not . 

For actually binding and testing end to end which package i should use(which is available in CI image)?

bitoku:
@amritansh1502 

AI suggested ncat. It lgtm, but we need to check if it works.

```
  @amritansh1502 Good analysis on the /dev/tcp failure — you're right that it does a TCP connect(), not a bind(), so it gets "Connection refused" since nothing is listening on port 80.

  The CI image (quay.io/crio/fedora-crio-ci) has nmap-ncat installed (see test/testdata/Dockerfile line 9), so you can use ncat to do an actual bind:

  # This binds (listens) on port 80 — tests that a non-root user can bind to a privileged port
  crictl exec --sync "$ctr_id" ncat -l 80 --idle-timeout 1 --recv-only

  ncat -l 80 creates a listening socket on port 80 (i.e. calls bind() + listen()). If ip_unprivileged_port_start=0 is in effect, the bind succeeds and ncat exits cleanly (with the idle timeout so it doesn't hang forever). If the sysctl is not set (value is 1024), the bind fails with "Permission denied".

  For the negative test case (ip_unprivileged_port_start=1024), run the same command and assert failure:

  # Should fail — non-root cannot bind to port 80 when unprivileged_port_start=1024
  run crictl exec --sync "$ctr_id" ncat -l 80 --idle-timeout 1 --recv-only
  [ "$status" -ne 0 ]

  This gives you a true end-to-end bind test using a tool already in the CI image.
```

amritansh1502:
> @amritansh1502
> 
> AI suggested ncat. It lgtm, but we need to check if it works.
> 
> ```
>   @amritansh1502 Good analysis on the /dev/tcp failure — you're right that it does a TCP connect(), not a bind(), so it gets "Connection refused" since nothing is listening on port 80.
> 
>   The CI image (quay.io/crio/fedora-crio-ci) has nmap-ncat installed (see test/testdata/Dockerfile line 9), so you can use ncat to do an actual bind:
> 
>   # This binds (listens) on port 80 — tests that a non-root user can bind to a privileged port
>   crictl exec --sync "$ctr_id" ncat -l 80 --idle-timeout 1 --recv-only
> 
>   ncat -l 80 creates a listening socket on port 80 (i.e. calls bind() + listen()). If ip_unprivileged_port_start=0 is in effect, the bind succeeds and ncat exits cleanly (with the idle timeout so it doesn't hang forever). If the sysctl is not set (value is 1024), the bind fails with "Permission denied".
> 
>   For the negative test case (ip_unprivileged_port_start=1024), run the same command and assert failure:
> 
>   # Should fail — non-root cannot bind to port 80 when unprivileged_port_start=1024
>   run crictl exec --sync "$ctr_id" ncat -l 80 --idle-timeout 1 --recv-only
>   [ "$status" -ne 0 ]
> 
>   This gives you a true end-to-end bind test using a tool already in the CI image.
> ```

@bitoku  i have  updated the tests to use `ncat -l 80 --idle-timeout 1 --recv-only ` to actually bind to a privileged port instead of just checking the sysctl value. relevant CI jobs are passing now

bitoku (commented):
/approve

amritansh1502:
Pushed a follow-up commit to address CI feedback: the original implementation set `net.ipv4.ip_unprivileged_port_start=0` unconditionally for every non-host-network pod. That broke the sig-arch conformance test (sysctl whitelists `net.ipv4.ip_unprivileged_port_start`), which expects this sysctl to stay at the kernel default (1024) unless a pod explicitly requests otherwise, and it also silently changed a security-relevant default (any non-root process could bind privileged ports) for every existing workload with no opt-in.

Made the behavior opt-in instead:
- New `default_unprivileged_port_start `option in crio.conf (CLI: `--default-unprivileged-port-start, env: CONTAINER_DEFAULT_UNPRIVILEGED_PORT_START)`, defaults to false.

- When enabled, `net.ipv4.ip_unprivileged_port_start=0 i`s applied to non-host-network pods that haven't already set the value themselves (via securityContext.sysctls or default_sysctls), which always take precedence.
- When disabled (default), behavior is unchanged from stock CRI-O,  conformant with the 1024 default.
- Updated docs (cri.md, crio.conf.5.md, crio.8.md), shell completions, the unit test, and test/pod.bats to cover both the opted-in and default-off paths.

bitoku:
@amritansh1502 

Thank you for exploring the way.

I’m sorry to change direction, but after reading your latest update, I’m leaning toward not implementing this feature.

> New default_unprivileged_port_start option in crio.conf (CLI: --default-unprivileged-port-start, env: CONTAINER_DEFAULT_UNPRIVILEGED_PORT_START), defaults to false.

Because this can already be handled using the default_sysctls option, a new flag isn't really needed.
cc @haircommander

PR #10062, opened 2026-06-22
Fix RemoveSandbox error and added tests for create/delete race condition using NRI plugin

ngopalak-redhat:
#### What type of PR is this?

/kind ci

One of the customers asked:
```
Is RemovePodSandbox expected to handle the case where RunPodSandbox hasn't fully completed yet?
```
I investigated that cri-o does handle this case already and I didn't find any issues. 

#### Why did I use NRI plugins?
To create a delay in the RunPodSandbox I used NRI plugin. The NRI plugin generally has a low timeout. So I had to increase it for the test to work. I added a small NRI plugin to consistently recreate the problem.

At the end of it, I didn't find any issues with cri-o. So I thought its worth keeping the test cases.

```release-note
None
```

## Summary

* **New Features**
  * Extended the NRI delay test plugin to optionally write per-pod execution timing details to a caller-specified file.
* **Tests**
  * Added Bats coverage for sandbox-creation delay behavior and concurrent pod removal, validating expected timing and recovery.
* **Chores**
  * Updated CI integration to upload the additional delay-plugin test binary, along with build/cleanup and test setup to recognize it.


test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,132 @@
...
+	cp "$NRI_DELAY_PLUGIN_BINARY" "$NRI_PLUGIN_DIR/90-delay-plugin"
+	chmod +x "$NRI_PLUGIN_DIR/90-delay-plugin"
+
+	# Configure NRI with extended timeout and plugin directory
+	cat << EOF > "$CRIO_CONFIG_DIR/20-nri-timeout.conf"
+[crio.nri]
+enable_nri = true
+nri_plugin_dir = "$NRI_PLUGIN_DIR"
+nri_plugin_request_timeout = "30s"
+nri_plugin_registration_timeout = "10s"
```
haircommander, line 20: what consumes this?

ngopalak-redhat: Default is 5 seconds from
https://github.com/containerd/nri/blob/main/pkg/api/timeouts.go#L25

Cri-o invokes: nri.SetPluginRegistrationTimeout(c.PluginRegistrationTimeout)

https://github.com/cri-o/cri-o/blob/main/internal/config/nri/nri.go#L156

Its set here: https://github.com/cri-o/cri-o/blob/main/internal/config/nri/nri.go#L19

Initially I didn't set this and the test was failing.

Also I have made this an annotation now based on @rphillips comment

saschagrunert (commented):
Thanks for looking into this and adding test coverage for the create/delete race! A few things came up during review.

The main issue: `TestRunPodSandboxRemoveRaceCondition` doesn't actually test a race. `createPod()` calls `crio.CreatePod()`, which calls `r.runtime.RunPodSandbox()` — a blocking gRPC call that doesn't return until all NRI hooks (including the 10s delay) complete. So by the time `createPod()` returns and the test reaches `RemovePod`, the pod is already fully created. The BATS test gets this right by running `crictl runp` in the background with `&`. The Go test would need to run `createPod` in a goroutine to actually race.

See inline comments for details.

test/nri/race_condition_test.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestRunPodSandboxRemoveRaceCondition tests the race condition where
+// RemovePodSandbox is called while RunPodSandbox is still in progress.
+func TestRunPodSandboxRemoveRaceCondition(t *testing.T) {
```
saschagrunert: Both new tests are missing the `skipTestForCondition` guard that every other test in this package uses. If `crio` is nil, `test.Setup(t)` will panic via `purgePodsAndContainers()` → `crio.ListContainers()`.

test/nri/race_condition_test.go
```diff
@@ -0,0 +1,89 @@
...
+				WithRunPodSandboxDelay(10 * time.Second),
+			},
+		},
+	}
+
+	test.Setup(t)
+	test.StartPlugins(WaitForPluginSync)
+
+	pod := test.createPod()
+	require.NotEmpty(t, pod, "create pod should return pod ID")
```
saschagrunert: `createPod()` is a blocking call. It calls `crio.CreatePod()` which calls `r.runtime.RunPodSandbox()` (a unary gRPC call that blocks until all NRI hooks complete, including the 10s delay). So by the time `createPod()` returns, the pod is already fully created. The `RemovePod` below is not racing with creation.

To actually test the race, `createPod` would need to run in a goroutine (similar to how the BATS test runs `crictl runp` in the background with `&` on line 84 of `sandbox_race_condition.bats`).

test/nri/race_condition_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	time.Sleep(2 * time.Second)
+
+	err := crio.RemovePod(pod)
+	if err != nil {
+		t.Logf("RemovePod during creation returned error (expected): %v", err)
+	} else {
+		t.Logf("RemovePod during creation succeeded (may indicate race handling)")
+	}
+
```
saschagrunert: Neither branch asserts anything, both just log. The test passes regardless of what `RemovePod` returns. If this should verify a specific behavior (e.g. removal is rejected during creation), it should assert on the error.

test/nri/runtime.go
```diff
@@ -39,6 +38,10 @@ type imageRefs struct {
 }
 
 var (
+	// requestTimeout can be overridden for tests that need longer timeouts
+	// (e.g., tests with NRI plugin delays).
+	requestTimeout = 10 * time.Second
```
saschagrunert: This changes `requestTimeout` from a `const` to a mutable package-level `var`. Consider adding a timeout field to `nriTest` or `runtime` instead, so tests can override it without mutating shared state.

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,136 @@
...
+function teardown() {
+	cleanup_test
+}
+
+function check_nri_activity() {
+	# Verify NRI plugin was invoked by checking that RunPodSandbox took >= 9 seconds
+	# The log for NRI plugin is not part of cri-o logs
+	local start_time end_time
+	start_time=$(grep "Running pod sandbox" "$CRIO_LOG" | tail -1 | awk '{print $1}' | tr -d 'time="' | tr -d 'Z"')
+	end_time=$(grep "Ran pod sandbox.*with infra container" "$CRIO_LOG" | tail -1 | awk '{print $1}' | tr -d 'time="' | tr -d 'Z"')
```
saschagrunert: This relies on specific CRI-O log message strings and the logrus text format (`time="..."` as first field). If either changes, the function breaks silently. The macOS `date -j` fallback on lines 41-42 is also dead code since CI is Linux-only.

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,136 @@
...
+	jq '.annotations += {"nri-delay-plugin/delay": "10s"}' "$pod_config" > "$pod_config.tmp"
+	mv "$pod_config.tmp" "$pod_config"
+
+	# Start pod creation in background - NRI plugin will delay based on annotation
+	crictl runp "$pod_config" &
+	RUNP_PID=$!
+
+	# Wait for storage creation to complete
+	sleep 2
+
```
saschagrunert: This assumes the most recently modified directory is the in-flight pod. Leftover dirs or timestamp granularity could return the wrong ID.

Nit: the comments on lines 92-94 about what didn't work are helpful for review context but will go stale in the codebase. Consider moving them to the commit message.

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,136 @@
...
+	# Find the pod ID by looking at the most recently created storage directory
+	# This works because storage is created before the NRI delay
+	# These methods didn't work - Noting down to save someone else's time :)
+	# - "sudo crictl pods --state NotReady" - Didn't give the pod id
+	# - Tail on the cri-o logs worked but very flaky
+	storage_dir="$TESTDIR/crio/overlay-containers"
+	pod_id=$(find "$storage_dir" -mindepth 1 -maxdepth 1 -type d -printf '%T@ %f\n' 2> /dev/null | sort -rn | head -1 | awk '{print $2}')
+
+	if [ -z "$pod_id" ]; then
+		echo "# ERROR: No pod storage found in $storage_dir"
```
saschagrunert: Error message strings aren't part of the API contract. If someone rewords the message, this breaks even though the behavior is correct. Matching on the gRPC status code would be more robust.

ngopalak-redhat: Fixed by using gRPC codes

test/sandbox_race_condition.bats
ngopalak-redhat: Fixed this by using POD_ID from the NRI plugin. I made sure to flush the log so that its available to the test immediately.

test/sandbox_race_condition.bats
ngopalak-redhat: Removed it and added an annotation for the log file. Now using the logs from the test directory

test/nri/runtime.go
ngopalak-redhat: Removed go lang test

test/nri/race_condition_test.go
ngopalak-redhat: Removed go lang test

test/nri/race_condition_test.go
ngopalak-redhat: Removed golang test

test/nri/race_condition_test.go
ngopalak-redhat: Removed the golang test

ngopalak-redhat:
> Thanks for looking into this and adding test coverage for the create/delete race! A few things came up during review.
> 
> The main issue: `TestRunPodSandboxRemoveRaceCondition` doesn't actually test a race. `createPod()` calls `crio.CreatePod()`, which calls `r.runtime.RunPodSandbox()` — a blocking gRPC call that doesn't return until all NRI hooks (including the 10s delay) complete. So by the time `createPod()` returns and the test reaches `RemovePod`, the pod is already fully created. The BATS test gets this right by running `crictl runp` in the background with `&`. The Go test would need to run `createPod` in a goroutine to actually race.
> 
> See inline comments for details.

Thanks for reviewing. I have fixed the bats test and removed the golang test. I have addressed the comments

bitoku:
@ngopalak-redhat can you please update the release-note?

/ok-to-test

ngopalak-redhat:
> Thanks for looking into this and adding test coverage for the create/delete race! A few things came up during review.
> 
> The main issue: `TestRunPodSandboxRemoveRaceCondition` doesn't actually test a race. `createPod()` calls `crio.CreatePod()`, which calls `r.runtime.RunPodSandbox()` — a blocking gRPC call that doesn't return until all NRI hooks (including the 10s delay) complete. So by the time `createPod()` returns and the test reaches `RemovePod`, the pod is already fully created. The BATS test gets this right by running `crictl runp` in the background with `&`. The Go test would need to run `createPod` in a goroutine to actually race.
> 
> See inline comments for details.

@saschagrunert Can you review again? This only has BATS test now.

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,148 @@
...
+		return 1
+	fi
+
+	echo "# Found pod ID from NRI plugin: $pod_id"
+
+	# RemovePodSandbox called while RunPodSandbox is in NRI hook (before SetCreated())
+	# Expected: gRPC NotFound error (pod exists in storage but not yet marked as created)
+	output=$(crictl rmp "$pod_id" 2>&1 || true)
+
+	if [[ "$output" != *"code = NotFound"* ]]; then
```
saschagrunert: `RemovePodSandbox` in `server/sandbox_remove.go:29` handles the "not yet created" case with `fmt.Errorf(...)` (plain error), not a gRPC status error. gRPC maps plain errors to `codes.Unknown`, not `codes.NotFound`. There is no `codes.NotFound` usage anywhere in `sandbox_remove.go`. So `crictl rmp` output will contain `code = Unknown` and this assertion will always fail.

Either match on `code = Unknown` here, or (if NotFound is the intended semantics) fix `sandbox_remove.go:29` to use `status.Errorf(codes.NotFound, ...)`. This likely explains the CI failure.

test/nri/plugin.go
```diff
@@ -82,6 +85,12 @@ func WithUpdateHandler(fn func(*plugin, *api.PodSandbox, *api.Container) ([]*api
 	}
 }
 
+func WithRunPodSandboxDelay(delay time.Duration) PluginOption {
```
saschagrunert: The Go test that used this was removed, but the delay infrastructure remains: `runPodSandboxDelay` field (line 38), this function, and the delay logic in `RunPodSandbox` (lines 207-218). Nothing references any of it. Please remove.

test/nri/runtime.go
saschagrunert: Flagged in the previous review. Nothing assigns to `requestTimeout` anymore, so this should be reverted back to a `const`.

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,148 @@
...
+		echo "# ERROR: Expected gRPC NotFound error during race condition"
+		echo "# Got: $output"
+		return 1
+	fi
+
+	echo "# Race condition correctly triggered: NotFound error received"
+
+	# Wait for RunPodSandbox to complete
+	wait $RUNP_PID || true
```
saschagrunert: If `crictl runp` fails for a reason unrelated to the race, `|| true` silently masks it. Consider checking the exit status.

Also, the conditional at line 135 should be a hard assertion: after `RunPodSandbox` completes, the pod must be present since the failed `rmp` returned an error without touching storage. If the pod is missing here, something unexpected happened and the test should fail rather than skip cleanup.

saschagrunert (changes requested):
Thanks for iterating on this! The BATS test structure is solid, nice approach using the NRI plugin log to get the pod ID. A few things to clean up, and one that likely explains the CI failure.

test/sandbox_race_condition.bats
ngopalak-redhat: Fixed this. Now I have added one more test. Since pod sandbox status is first checked while doing `crictl rmp` I have used GRPC call for RemoveSandbox to hit that condition. Also fixed the RemoveSandbox code return NotFound.

test/sandbox_race_condition.bats
ngopalak-redhat: Fixed

test/nri/runtime.go
ngopalak-redhat: Fixed

test/nri/plugin.go
ngopalak-redhat: Fixed

ngopalak-redhat:
> Thanks for iterating on this! The BATS test structure is solid, nice approach using the NRI plugin log to get the pod ID. A few things to clean up, and one that likely explains the CI failure.

@saschagrunert Thanks for the review, I have fixed RemoveSandbox error that is returned. Hence edited the title also. But that condition is never hit as pod status is always checked before RemoveSandbox is invoked. Hence I have now enhancement one of the test modules to send GRPC requests directly to cri-o. This can be used in future tests also.

ngopalak-redhat:
@saschagrunert Can you review this?

saschagrunert (commented):
Thanks for the investigation and test coverage. The gRPC status code fix is correct, and the NRI delay plugin approach for testing the race condition is solid.

A couple of issues that need fixing before merge (see inline comments), plus a few nits.

.github/workflows/integration.yml
```diff
@@ -221,7 +222,7 @@ jobs:
             test/checkcriu/checkcriu \
             test/checkseccomp/checkseccomp \
             test/copyimg/copyimg \
-            test/updateunified/updateunified \
+            test/criogrpccaller/criogrpccaller \
             test/nri/nri.test
```
saschagrunert, line 227: `nri-delay-plugin` is uploaded as an artifact (line 56) but not included in the chmod block here. The `sandbox_race_condition.bats` tests will fail in CI because `$NRI_DELAY_PLUGIN_BINARY` won't be executable (or present) after artifact download.

Add `test/nri-delay-plugin/nri-delay-plugin \` before `test/nri/nri.test`.

.gitignore
```diff
@@ -14,6 +14,7 @@
 /test/checkcriu/checkcriu
 /test/copyimg/copyimg
 /test/updateunified/updateunified
+/test/nri-delay-plugin/nri-delay-plugin
```
saschagrunert, line 17: `/test/criogrpccaller/criogrpccaller` is missing from `.gitignore`. The `updateunified` entry on line 16 was renamed to `criogrpccaller` everywhere else (Makefile, CI, common.sh) but not here.

Replace `/test/updateunified/updateunified` on line 16 with `/test/criogrpccaller/criogrpccaller`.

server/sandbox_remove.go
```diff
@@ -26,7 +28,7 @@ func (s *Server) RemovePodSandbox(ctx context.Context, req *types.RemovePodSandb
 		}
 
 		if errors.Is(err, errSandboxNotCreated) {
-			return nil, fmt.Errorf("sandbox %s is not yet created", req.GetPodSandboxId())
+			return nil, status.Errorf(codes.NotFound, "sandbox %s is not yet created", req.GetPodSandboxId())
```
saschagrunert, line 31: `StopPodSandbox` handles the same `errSandboxNotCreated` with a bare `fmt.Errorf` (`sandbox_stop.go:29`), which maps to `codes.Unknown` over gRPC. If the goal is proper gRPC status codes for "sandbox not ready" errors, `StopPodSandbox` should get the same treatment. Otherwise callers see different gRPC codes for the same underlying condition depending on which RPC they call.

test/nri/plugin.go
```diff
@@ -192,7 +192,7 @@ func (p *plugin) Synchronize(_ context.Context, pods []*api.PodSandbox, ctrs []*
 	return nil, nil
 }
 
-func (p *plugin) RunPodSandbox(_ context.Context, pod *api.PodSandbox) error {
+func (p *plugin) RunPodSandbox(ctx context.Context, pod *api.PodSandbox) error {
```
saschagrunert: Nit: `ctx` is named but unused in the function body (lines 196-212). The original `_ context.Context` was correct.

test/nri/runtime.go
```diff
@@ -38,6 +37,10 @@ type imageRefs struct {
 	busyboxName string
 }
 
+const (
```
saschagrunert: Nit: `requestTimeout` was moved from the existing const block (line 17) into its own standalone block. There is no functional reason for the separation. Keep it with the other timeout constants.

ngopalak-redhat: Fixed

test/nri/plugin.go
ngopalak-redhat: This file is not changed

server/sandbox_remove.go
ngopalak-redhat: Both use "status.Errorf(codes.NotFound, "

.gitignore
ngopalak-redhat: added

.github/workflows/integration.yml
ngopalak-redhat: fixed

ngopalak-redhat:
> Thanks for the investigation and test coverage. The gRPC status code fix is correct, and the NRI delay plugin approach for testing the race condition is solid.
> 
> A couple of issues that need fixing before merge (see inline comments), plus a few nits.

@saschagrunert Can you review? Probably a good time to get this in as most of the integration test issues are resolved

server/sandbox_remove.go
saschagrunert, line 31: `PodSandboxStatus` already returns `codes.NotFound` for `errSandboxNotCreated` (sandbox_status.go:25), so making `RemovePodSandbox` consistent is reasonable. However, for Remove/Stop the semantics differ: `NotFound` can be interpreted as "sandbox does not exist, nothing to do" rather than "sandbox is mid-creation, try again later." `codes.FailedPrecondition` would be more precise for these two methods.

server/sandbox_stop.go
```diff
@@ -26,7 +27,7 @@ func (s *Server) StopPodSandbox(ctx context.Context, req *types.StopPodSandboxRe
 		}
 
 		if errors.Is(err, errSandboxNotCreated) {
-			return nil, fmt.Errorf("StopPodSandbox failed as the sandbox is not created: %s", req.GetPodSandboxId())
+			return nil, status.Errorf(codes.NotFound, "StopPodSandbox failed as the sandbox is not created: %s", req.GetPodSandboxId())
```
saschagrunert: Inconsistent message style with `sandbox_remove.go:31` which uses `"sandbox %s is not yet created"`. This one embeds the function name in the error message (the caller already knows which function it called). Consider using the same pattern in both places.

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,194 @@
...
+	fi
+}
+
+function check_delay_plugin() {
+	[ -f "$CRIO_LOG" ] && grep -q "delay-plugin\|90-delay-plugin" "$CRIO_LOG"
+}
+
+@test "verify NRI delay plugin is loaded and invoked" {
+	start_crio
+	sleep 2
```
saschagrunert: Fixed `sleep 2` could be insufficient in slow CI. Polling the CRI-O log for the plugin registration message (e.g. grepping for `"delay-plugin"` in `$CRIO_LOG` with a retry loop) would be more robust.

server/sandbox_remove.go
ngopalak-redhat: PodSandboxStatus already returns NotFound for errSandboxNotCreated (sandbox_status.go:25), and crictl rmp calls Status before Remove. Returning FailedPrecondition from Remove/Stop while Status returns NotFound for the same condition would be inconsistent. Keeping NotFound across all three methods means the caller gets a uniform signal that the sandbox isn't available yet.

server/sandbox_stop.go
ngopalak-redhat: Fixed

test/sandbox_race_condition.bats
ngopalak-redhat: Fixed

ngopalak-redhat:
@saschagrunert I have addressed the comments.

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,200 @@
...
+		echo "# ERROR: Pod ID not found in NRI log file: $nri_log"
+		[ -f "$nri_log" ] && echo "# NRI log contents:" && cat "$nri_log" || echo "# NRI log does not exist"
+		wait $RUNP_PID || true
+		return 1
+	fi
+
+	echo "# Found pod ID: $pod_id"
+
+	output=$(crictl rmp "$pod_id" 2>&1 || true)
+	if [[ "$output" != *"code = NotFound"* ]] || [[ "$output" != *"sandbox not created"* ]]; then
```
saschagrunert, line 120: Still matching on error message strings (flagged in my first review). During this race window, the only path that returns `NotFound` from `PodSandboxStatus` is `errSandboxNotCreated`: the sandbox is in the map (`addSandbox` at `sandbox_run_linux.go:660` runs before NRI hooks at line 743), but `Created()` is false (`SetCreated` at line 779 runs after hooks return). The "specified pod sandbox not found" path can't fire because `getSandbox` finds the sandbox. The gRPC code check is sufficient, drop the message string assertion:

```suggestion
	if [[ "$output" != *"code = NotFound"* ]]; then
```

test/sandbox_race_condition.bats
```diff
@@ -0,0 +1,200 @@
...
+		echo "# ERROR: Pod ID not found in NRI log file: $nri_log"
+		[ -f "$nri_log" ] && echo "# NRI log contents:" && cat "$nri_log" || echo "# NRI log does not exist"
+		wait $RUNP_PID || true
+		return 1
+	fi
+
+	echo "# Found pod ID: $pod_id"
+
+	output=$("$CRIOGRPCCALLER_BINARY" remove-pod-sandbox "$CRIO_SOCKET" "$pod_id" 2>&1 || true)
+	if [[ "$output" != *"NotFound"* ]] || [[ "$output" != *"not yet created"* ]]; then
```
saschagrunert, line 179: Same issue. `RemovePodSandbox` returns `NotFound` only from `errSandboxNotCreated` during this race (the "not found at all" path returns an empty response with no error, and `ErrIDEmpty` returns a bare error, not `NotFound`). The gRPC code check is sufficient:

```suggestion
	if [[ "$output" != *"NotFound"* ]]; then
```

server/sandbox_stop.go
```diff
@@ -26,7 +27,7 @@ func (s *Server) StopPodSandbox(ctx context.Context, req *types.StopPodSandboxRe
 		}
 
 		if errors.Is(err, errSandboxNotCreated) {
-			return nil, fmt.Errorf("StopPodSandbox failed as the sandbox is not created: %s", req.GetPodSandboxId())
+			return nil, status.Errorf(codes.NotFound, "sandbox %s is not yet created", req.GetPodSandboxId())
```
saschagrunert, line 30: This change is correct, but there is no unit test for it. `sandbox_remove_test.go` adds a test asserting `codes.NotFound` for the same `errSandboxNotCreated` path in `RemovePodSandbox`. `sandbox_stop_test.go` should get an equivalent test (add sandbox without marking it created, call `StopPodSandbox`, assert `codes.NotFound`).

saschagrunert (changes requested):
Thanks for iterating on this. Three items left.

PR #10066, opened 2026-06-23
Add local registry for integration test images

### LINKED ISSUE #9859: Use local registry for robust integration tests
Since we’re seeing so many 502 and connection errors when pulling from Quay, I suggest we use a local registry instead. If we pull the required images in a preparation step of GitHub Action, we can avoid external dependencies and make our test suite much more robust.

ngopalak-redhat:
Fixes #9859

```release-note

```


ngopalak-redhat:
@bitoku I tried using the local registry. It eliminates the copyimg also. There are around 18 tests failing. Still working on it

bitoku:
Thank you @ngopalak-redhat !
I believe this will stabilize the CI more.

For the follow up, we may be able to separate setup steps and execution steps for each runner.
Then it will reduce access to quay.io more, and can get more clear signal about what's wrong.

xw19:
@bitoku @ngopalak-redhat When are we planning to merge this PR?

Issue #10031, opened 2026-06-10
s390x: malformed state.json — first " not escaped in annotation values, breaking cAdvisor ContainerStatus (regression in v1.35.0 PR #9566 via goccy/go-json SWAR bug)


ashokpariya0:
### What happened?

On s390x architecture, CRI-O 1.35.x writes malformed JSON to
/var/lib/containers/storage/overlay-containers/<id>/userdata/state.json

for every container. The first `"` character inside any annotation value
that contains a nested JSON object or array is written without the
required  \  escape. All subsequent  " characters in the same value are
correctly escaped as `\"`.

This affects 100% of containers on s390x. The result is:
- crictl inspect <any-container> fails on all containers
- oc adm top pods / Kubernetes Metrics API return no data
- cAdvisor cannot unmarshal ContainerStatus → no container metrics collected

## Regression introduced by 
PR- https://github.com/cri-o/cri-o/pull/9566 introduced this bug

The specific trigger: internal/oci/container.go imports
`json "github.com/goccy/go-json"` (a third-party JSON library) which
uses a SWAR (SIMD Within A Register) byte-scanning optimisation
inherited from segmentio/encoding. That optimisation uses
bits.TrailingZeros64 on a uint64 loaded via unsafe.Pointer. This
is correct on little-endian architectures but produces wrong byte
positions on big-endian s390x, causing the first `"` inside certain
string values to be flushed without escaping.

Attaching more details
```
sh-5.1# crictl ps -q | head -1
48e3977b590fa4e73784aeda0518d6cd7f0cb466cc04723c75670b13ac15eed2
```

sh-5.1# 
```
sh-5.1# crictl inspect 48e3977b590fa4e73784aeda0518d6cd7f0cb466cc04723c75670b13ac15eed2
FATA[0000] get the status of containers: unmarshal status info JSON: invalid character 'i' after object key:value pair 
sh-5.1# 
```

```
sh-5.1# cat /var/lib/containers/storage/overlay-containers/48e3977b590fa4e73784aeda0518d6cd7f0cb466cc04723c75670b13ac15eed2/userdata/state.json 
[...]

Issue #9419, opened 2025-08-22
crio panics and writes strange container started at time


olyazavr:
### What happened?

There was a node with several pods that appeared to be unready, despite looking completely healthy. Upon further investigation and digging into kubelet logs, we noticed that kubelet was not running probes against these pods. Manual run of the probe resulted in success. Restarting kubelet did not change anything, and other very similar pods had probes running on them just fine.

Further investigation with crictl inspect showed `"startedAt": "1754-08-30T22:43:41.128654848Z",` for one container, and a similar date in 8/30/1754 17:47:39 for the other. Unaffected containers showed normal startedAt dates: `2025-08-21T21:24:16Z`

crio logs showed `Aug 21 21:23:49 ip-172-27-15-58 crio[12172]: fatal error: concurrent map iteration and map write`

### What did you expect to happen?

I expected a normal looking `startedAt` and no concurrent map iteration/write panics

### How can we reproduce it (as minimally and precisely as possible)?

Hard to repro, we have only encountered this a few times in the wild

### CRI-O and Kubernetes version

<details>

```console
$ crio --version
crio version 1.31.2
   GitCommit:      920af25f2b6fd4f52bced0e73f696e48464da543
   GitCommitDate:  2024-11-01T10:41:59Z
   GitTreeState:   dirty
   BuildDate:      1970-01-01T00:00:00Z
   GoVersion:      go1.22.5
   Compiler:       gc
   Platform:       linux/arm64
   Linkmode:       static
   BuildTags:
     static
     netgo
     osusergo
     exclude_graphdriver_btrfs
     seccomp
     apparmor
     selinux
     exclude_graphdriver_devicemapper
   LDFlags:          unknown
   SeccompEnabled:   true
   AppArmorEnabled:  false

```

```console
$ kubectl version --output=json
{
  "clientVersion": {
    "major": "1",
    "minor": "31",
    "gitVersion": "v1.31.7",
    "gitCommit": "da53587841b4960dc3bd2af1ec6101b57c79aff4",
    "gitTreeState": "clean",
    "buildDate": "2025-03-11T20:03:56Z",
    "goVersion": "go1.23.6",
    "compiler": "gc",
    "platform": "darwin/arm64"
  },
  "kustomizeVersion": "v5.4.2",
  "serverVersion": {
    "major": "1",
    "minor": "31",
    "gitVersion": "v1.31.7",
    "gitCommit": "da53587841b4960dc3bd2af1ec6101b57c79aff4",
[...]

PR #10071, opened 2026-06-24
oci: fix data races in container state handling

dgrisonnet:
#### What type of PR is this?

/kind bug

This is a draft PR to highlight two data races in container state handling that cause `concurrent map iteration and map write` panics and corrupted container state.

These races are the root cause of #9419 -- the bug is in CRI-O's state management, not in the JSON library. This is step 1 of the plan outlined in https://github.com/cri-o/cri-o/issues/10031#issuecomment-4682334443.

**Race 1: shared Annotations map**

[`UpdateContainerStatus`](https://github.com/cri-o/cri-o/blob/17125b27da7c9035aac4310912563274810813a7/internal/oci/runtime_oci.go#L1214) does a shallow copy of `c.state` (`state := *c.state`), which shares the `Annotations` map pointer. [`ContainerStateToDisk`](https://github.com/cri-o/cri-o/blob/17125b27da7c9035aac4310912563274810813a7/internal/lib/container_server.go#L657) calls `json.Encode(ctr.State())` which iterates the `Annotations` map. Concurrently, `json.Decode(&state)` in `UpdateContainerStatus` writes new keys into the same map through the shallow copy. Without the race detector, the runtime panics on concurrent map read+write.

**Fix:** clone the Annotations map after the shallow copy in `UpdateContainerStatus`, so `json.Decode` writes to its own map.

**Race 2: non-atomic struct copy**

[`State()`](https://github.com/cri-o/cri-o/blob/17125b27da7c9035aac4310912563274810813a7/internal/oci/container.go#L471) returns a pointer to the shared `c.state` and releases `opLock.RLock`. Callers then read fields without any lock held, racing with [`*c.state = *state`](https://github.com/cri-o/cri-o/blob/17125b27da7c9035aac4310912563274810813a7/internal/oci/runtime_oci.go#L1232) in `UpdateContainerStatus`. The struct copy is not atomic -- it writes each field one by one. `time.Time` is a multi-word struct (wall, ext, loc). A reader calling `State()` gets the raw pointer and can observe a partially-written `time.Time` -- some words from the old state, some from the new. The result is a nonsensical timestamp like year 1754, which then gets persisted to `state.json` and causes kubelet to skip probes for that container.

**Fix:** `State()` now returns a struct copy (snapshot) made under `RLock`. All production callers are read-only, so this is a safe change.

Ref #9419
Ref #10031

This is a draft to highlight the data races and propose fixes. Both races are exercised by new tests:

- `TestContainerStateToDiskRace` -- reproduces the shared Annotations map race
- `TestContainerStateStructCopyRace` -- reproduces the non-atomic struct copy race

```
go test -race -tags test -run 'TestContainerState.*Race' ./internal/oci/
```

```release-note
None
```

This PR was written in part with the assistance of Claude.

### DIFF
--- a/internal/oci/container_state_race_test.go
+++ b/internal/oci/container_state_race_test.go
@@ -104,11 +104,10 @@ func TestContainerStateToDiskRace(t *testing.T) {
 	wg.Wait()
 }
 
-// TestContainerStateStructCopyRace demonstrates that the struct
-// assignment *c.state = *state in UpdateContainerStatus
-// (runtime_oci.go:1232, 1252) races with concurrent readers.
-// The struct copy is not atomic — a reader can observe a
-// partially-written struct with fields from both old and new states.
+// TestContainerStateStructCopyRace verifies that State() returns an
+// independent snapshot, so the struct assignment *c.state = *state
+// in UpdateContainerStatus (runtime_oci.go:1232, 1252) does not
+// race with concurrent readers.
 //
 // Run with: go test -race -tags test -run TestContainerStateStructCopyRace ./internal/oci/
 func TestContainerStateStructCopyRace(t *testing.T) {
@@ -128,7 +127,8 @@ func TestContainerStateStructCopyRace(t *testing.T) {
 
 	var wg sync.WaitGroup
 
-	// Reader: ContainerStateToDisk encoding the state
+	// Reader: ContainerStateToDisk encoding the state.
+	// State() returns a copy, so encoding is safe.
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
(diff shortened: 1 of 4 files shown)


internal/oci/runtime_oci.go
```diff
@@ -1212,6 +1213,7 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 		}
 
 		state := *c.state
```
haircommander, line 1215: seems safer if we had a deepcopy method for state and used that here, rather than requiring this function to know to clone annotations map as well

internal/oci/container_state_race_test.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	}()
+
+	// Goroutine B: UpdateContainerStatus's shallow-copy + decode step
+	// (runtime_oci.go:1214-1215)
+	//
+	//   state := *c.state
+	//   json.NewDecoder(strings.NewReader(out)).Decode(&state)
+	//
+	// The shallow copy shares the Annotations map with c.state.
```
olyazavr, line 91: this does `stateCopy.Annotations = maps.Clone(stateCopy.Annotations)` so it doesn't test the existing race, rather the fix, right?

olyazavr:
fyi we have brought in this patch to our clusters and we haven't seen this same error since (nor any strange/incorrect behavior)

dgrisonnet:
Thank you for checking @olyazavr, I'll revisit this PR so it can be brought into cri-o

internal/oci/container_state_race_test.go
dgrisonnet: Correct, I fixed that

internal/oci/runtime_oci.go
dgrisonnet: true, I switched to a deepcopy method

PR #10082, opened 2026-06-30
Add BATS_TEST_TIMEOUT for integration tests

### LINKED ISSUE #10079: Add BATS_TEST_TIMEOUT for integration tests
## Problem

Integration tests can hang indefinitely when a test gets stuck. This was observed in the `ci-fedora-kata` Prow job for PR #10054:

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/cri-o_cri-o/10054/pull-ci-cri-o-cri-o-main-ci-fedora-kata/2070422031282737152

`test/test_runner.sh` sets `BATS_TEST_RETRIES=1` but does not set `BATS_TEST_TIMEOUT`. If a single test hangs, the entire CI job blocks until the platform kills it.

## Proposal

Add `BATS_TEST_TIMEOUT=600` (10 minutes) in `test/test_runner.sh` to enforce a per-test time limit:

```bash
export BATS_TEST_TIMEOUT=600
```

This would abort any individual test that exceeds 10 minutes, making it easy to identify which test hung instead of waiting for the entire job to time out.

Tests that intentionally sleep (e.g. `test/timeout.bats`) may need a per-file override.

## References

- [bats-core BATS_TEST_TIMEOUT docs](https://bats-core.readthedocs.io/en/stable/writing-tests.html)
- [bats-core/bats-core#1020](https://github.com/bats-core/bats-core/issues/1020) — known issue where `run` subprocesses may not be killed on timeout

MilindRastogi24:
#### What type of PR is this?

/kind ci

Integration tests can hang indefinitely when a single BATS test gets stuck, blocking the entire CI job until the platform kills it. test/test_runner.sh already sets BATS_TEST_RETRIES=1 but did not set a per-test time limit.

This PR adds export BATS_TEST_TIMEOUT=600 (10 minutes) so any individual test that exceeds that limit is aborted and reported as timed out. That makes hung tests fail fast and easier to identify instead of waiting for the whole job to time out.

Fixes #10079 

```release-note
None
```

## Summary by CodeRabbit

* **Chores**
  * Updated the test runner to stop any single test after 10 minutes, helping prevent hung tests from blocking validation.

### DIFF
--- a/test/test_runner.sh
+++ b/test/test_runner.sh
@@ -47,6 +47,8 @@ export JOBS=${JOBS:-$(nproc --all)}
 # The maximum number of additional attempts that will be made on a failed test before it is finally considered failed.
 # https://bats-core.readthedocs.io/en/stable/writing-tests.html#special-variables
 export BATS_TEST_RETRIES=1
+# Abort any individual test that exceeds 10 minutes so hung tests fail fast instead of blocking the entire CI job.
+export BATS_TEST_TIMEOUT=600
 
 bats --version
 


bitoku:
/ok-to-test

Thanks!
Can you try creating a test case that takes more than 10 mins and put the result here?

MilindRastogi24:
> /ok-to-test
> 
> Thanks! Can you try creating a test case that takes more than 10 mins and put the result here?

Sure

ngopalak-redhat:
@bitoku Adding `BATS_TEST_TIMEOUT` prevents hung tests from blocking CI. However, I think 120 seconds (2 minutes) would be a better default than 600 seconds (10 minutes). A 2-minute timeout means hung tests fail in 2 minutes instead of 10, giving faster CI feedback. Tests that legitimately need >2 minutes should document why by explicitly overriding the timeout and add change it just before the test starts.

What do you think?

My choice of 2 minutes may not be accurate. @MilindRastogi24 can do some tests to find the right value. I'm thinking 80% if the tests will be under 2 mins.

bitoku:
@ngopalak-redhat 
I'm open to 2mins or any value that won't make the tests flaky and that is not too long.
I just thought 10 mins is long enough and all tests should finish within it, so it will be easier. No other specific reason for it.

bitoku:
@MilindRastogi24 Do you want to continue to work on this?

Issue #10085, opened 2026-06-30
Flaky test: "clear podman containers when shutdown file not found"


bitoku:
## Summary

The test "clear podman containers when shutdown file not found" in `test/crio-wipe.bats` is flaky. It fails intermittently across unrelated PRs, architectures, and job variants.

## Failure frequency (last 2 weeks, Jun 16–30 2026)

Out of 25 failed integration jobs that ran this test, **6 (24%)** had this specific test fail — while the remaining 19 ran the test and it passed (those jobs failed for other reasons).

| Date | Job variant | Branch | Run |
|------|------------|--------|-----|
| Jun 30 | `integration / userns / crun / amd64` | `issue/gvisor/10313` | [28426712122](https://github.com/cri-o/cri-o/actions/runs/28426712122) |
| Jun 30 | `integration / conmon / crun / amd64` | `ngopalak/retry_delay` | [28419101281](https://github.com/cri-o/cri-o/actions/runs/28419101281) |
| Jun 29 | `integration / conmon / crun / amd64` | `config/additional-artifact-stores-ga` | [28374522668](https://github.com/cri-o/cri-o/actions/runs/28374522668) |
| Jun 26 | `integration / conmon / crun / arm64` | `ngopalak/delay_nri` | [28218825722](https://github.com/cri-o/cri-o/actions/runs/28218825722) |
| Jun 25 | `integration / userns / crun / amd64` | `increase-kata-integration-timeout` | [28160015041](https://github.com/cri-o/cri-o/actions/runs/28160015041) |
| Jun 25 | `integration / conmon / crun / arm64` | `fix/default-unprivileged-ports` | [28159446258](https://github.com/cri-o/cri-o/actions/runs/28159446258) |

## Observations

- Fails across **6 different, unrelated PRs** — not tied to any specific code change
- Fails on both **amd64** and **arm64**
- Fails in both **conmon** and **userns** job variants
- Roughly **~3 failures per week**

ngopalak-redhat:
@bitoku The failure in this test is due to:
```
# time="2026-06-30T04:10:02.3422537Z" level=info msg="Checking whether cri-o should wipe containers: open /tmp/tmp.LA5w28e2AH/version.tmp: no such file or directory"
[...]

PR #10088, opened 2026-06-30
test: deflake default runtime metric test

### LINKED ISSUE #10086: Flaky integration test: "default runtime metric" (~24% failure rate)
## Summary

The `default runtime metric` test in `test/metrics.bats` is the most frequently flaky integration test, failing in approximately **24% of CI runs** over the last two weeks (Jun 15–29, 2026). It fails across all job variants (conmon/conmon-rs, amd64/arm64).

## Failure Details

The test fails at line 107 of `test/metrics.bats`. After calling `reload_crio` and `wait_for_log '"updating runtime configuration"'`, the curl to fetch the updated default runtime metric fails:

```
#   `METRIC=$(curl -sf "http://localhost:$PORT/metrics" | grep "^container_runtime_crio_default_runtime{runtime=\"new\"}")' failed
# time="..." level=fatal msg="validate service connection: ... rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix ...crio.sock: connect: no such file or directory\""
```

Race condition: `wait_for_log` sees the log line, but the metrics endpoint is not yet updated or CRI-O is briefly unavailable during reload.

## Reproduction

Intermittent. Failed in ~24 of 99 integration CI runs over the last 2 weeks. Not branch-specific — fails on `main`, PRs, and release branches.

## Suggested fix

Poll the metrics endpoint for the new runtime value with a timeout, rather than querying once immediately after `wait_for_log`.

mateenali66:
/kind flake

The "default runtime metric" test waited for the `"updating runtime configuration"` log line, which is emitted when the reload *starts*, then queried the metrics endpoint before the reload finished, so it flaked (~24% of integration runs). Wait for the `"Configuration reload completed"` line instead, which is logged once the reload is done and the new default runtime metric is in place.

Fixes #10086

`.bats` files are excluded from shfmt/shellcheck.

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Improved the reliability of the runtime metrics test after configuration reload by waiting for the correct CRI-O reload completion message before running post-reload assertions.
  * This reduces flakiness caused by timing differences between reload completion and metric availability.

### DIFF
--- a/test/metrics.bats
+++ b/test/metrics.bats
@@ -101,7 +101,7 @@ runtime_path = "$RUNTIME_BINARY_PATH"
 EOF
 
 	reload_crio
-	wait_for_log '"updating runtime configuration"'
+	wait_for_log '"Configuration reload completed"'
 
 	# Check that the new default runtime metric is present
 	METRIC=$(curl -sf "http://localhost:$PORT/metrics" | grep "^container_runtime_crio_default_runtime{runtime=\"new\"}")


bitoku:
thanks!

/ok-to-test

xw19:
The changes looks good to me.

test/metrics.bats
```diff
@@ -103,9 +108,9 @@ EOF
 	reload_crio
 	wait_for_log '"updating runtime configuration"'
 
-	# Check that the new default runtime metric is present
-	METRIC=$(curl -sf "http://localhost:$PORT/metrics" | grep "^container_runtime_crio_default_runtime{runtime=\"new\"}")
-	[[ "$METRIC" == "container_runtime_crio_default_runtime{runtime=\"new\"} 1" ]]
+	# the metrics endpoint can lag the reload log line, and crio may be briefly
```
ngopalak-redhat: @mateenali66  I don't think retry would be best solution here. The config reload won't create a new socket. Can you try this fix?
```
wait_for_log "Configuration reload completed"
```
This will make sure that the config reload completes. If it still happens after this fix we need to investigate why there is a socket issue.

ngopalak-redhat (changes requested):

mateenali66:
Switched to `wait_for_log '"Configuration reload completed"'` and dropped the retry/helper, so the metric is only checked after the reload finishes. If it still flakes we can look at the socket side you mentioned.

ngopalak-redhat:
/lgtm

Based on my analysis https://github.com/cri-o/cri-o/pull/10088#discussion_r3503667482 and the success of the integration tests the fix is good. If the error re-appears please re-open the issue.

xw19:
@cri-o/cri-o-maintainers PTAL

mateenali66:
The remaining red here is not this change. `default runtime metric` (the test this PR deflakes) passes on both amd64 and arm64 in the latest run. The failing case is `clear podman containers when shutdown file not found`, which is a separate flake fixed by #10089 and is not in this branch. `e2e-gcp-ovn` and `images` are prow infra failures.

/retest keeps hitting the #10089 flake until that lands, so the two are effectively gated on each other.

test/metrics.bats
mateenali66: Done in c2a55b15f. Swapped the wait to `wait_for_log "Configuration reload completed"` (server logs it once the reload finishes) and dropped the retry loop, so the single curl reads the new runtime value deterministically. Latest integration run passed it (`ok 324 default runtime metric`). The remaining red on this PR is the crio-wipe flake from #10085, which #10089 fixes.

bitoku:
/lgtm
@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):

PR #10105, opened 2026-07-02
Use fsNotify for fetching container exit

ngopalak-redhat:
#### What type of PR is this?

/kind cleanup

Use a faster way to detect presence of container exit file

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exit-file detection by switching from periodic polling to real-time filesystem event monitoring, reducing delays.
  * Added an immediate initial check for the exit file before starting event monitoring.
  * Made the wait operation context-aware, so cancellation returns promptly with the correct error.
  * Kept waiting behavior bounded with a hard timeout and a final fallback check to prevent indefinite waits.

### DIFF
--- a/internal/oci/runtime_oci.go
+++ b/internal/oci/runtime_oci.go
@@ -1197,11 +1197,7 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 			// Wait for it to appear before defaulting to exit code 255.
 			// This prevents a race where fast-exiting containers
 			// get a spurious exit code 255.
-			// Release opLock before waiting to avoid blocking other operations.
-			c.opLock.Unlock()
-			waitErr := waitForExitFile(c)
-			c.opLock.Lock()
-
+			waitErr := waitForExitFile(ctx, c)
 			if waitErr != nil {
 				log.Errorf(ctx, "Failed to update container status from exit file for %s: %v", c.ID(), waitErr)
 				c.state.Finished = time.Now()
@@ -1240,7 +1236,7 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 	}
 
 	c.opLock.Unlock()
-	err = waitForExitFile(c)
+	err = waitForExitFile(ctx, c)
 	c.opLock.Lock()
 
 	// run command again
@@ -1285,9 +1281,9 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
 }
 
 // waitForExitFile waits for the container's exit file to appear using
-// exponential backoff. Returns nil if the file appeared, or an error if it did
-// not appear within the timeout (~5s total: 500ms initial, 1.2 factor, 6 steps).
-func waitForExitFile(c *Container) error {
+// fsnotify. Returns nil if the file appeared, or an error if the context
+// is cancelled or the file did not appear within the 5s timeout.
+func waitForExitFile(ctx context.Context, c *Container) error {
 	exitFilePath := c.exitFilePath()
 
 	if _, err := os.Stat(exitFilePath); err == nil {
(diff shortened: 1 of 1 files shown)


ngopalak-redhat:
@bitoku I came across this while reviewing the cri-o wipe PR yesterday. This is unrelated to that but saw the usage of "ExponentialBackoff". Here I see 2 issues hence thinking of this fix:
1. The lock is held while its waiting for the exit file. This can be avoided. This is small improvement only as we are anyway in the error state. Its only making sure that we don't call wait while holding a lock.
2. fsNotify can replace the ExponentialBackOff as we are just waiting for the file to the created. This can help as it brings down the wait time from 500ms to single digit ms (Theoretically). 

Would like your review as its an improvement on top of https://github.com/cri-o/cri-o/pull/9846.

Will fix the linter errors later.

bitoku:
/assign

The direction looks reasonable, though I have a concern wether it's safe to remove the lock.

internal/oci/runtime_oci.go
```diff
@@ -1286,20 +1290,44 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
...
-			}
+	if _, err := os.Stat(exitFilePath); err == nil {
+		return nil
+	}
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+	defer watcher.Close()
```
bitoku, line 1296: There could be a race condition, I guess.
If exitFilePath is created after `os.Stat` and `fsnotify.NewWatcher`, what will happen?
Does it wait till the timeout?

ngopalak-redhat: Yes, right I have fixed it now

ngopalak-redhat:
> /assign
> 
> The direction looks reasonable, though I have a concern wether it's safe to remove the lock.

@bitoku Thanks for the review, I reverted the lock change and retained only the fsNotify().

internal/oci/runtime_oci.go
```diff
@@ -1281,25 +1281,59 @@ func (r *runtimeOCI) UpdateContainerStatus(ctx context.Context, c *Container) er
...
-			Factor:   1.2,
-			Steps:    6,
-		},
-		func() (bool, error) {
-			_, err := os.Stat(exitFilePath)
-			if err != nil {
-				return false, nil
+	if _, err := os.Stat(exitFilePath); err == nil {
+		return nil
+	}
```
bitoku, line 1290: There's already check after watcher. Can we remove this, or did you intentionally keep it for performance reason?
If it's intentional, it's worth commenting about why we keep it.

ngopalak-redhat: Both checks are intentional. The first os.Stat is a fast path to skip watcher setup when the exit file already exists (avoids allocating an inotify fd + goroutine). The second os.Stat after watcher.Add closes the race where the file appears between the first check and the watcher registration. I have added a comment.

ngopalak-redhat:
@bitoku Can you please review?

bitoku (commented):
Sorry having read it again, I started to think this will cause regression.

inotify is limited resource, limited by `fs.inotify.max_user_instances`. This will spawn as many inotify instances as the containers. When users drain node, there will be more instances than the limit, and the watcher will fail.

ngopalak-redhat:
> Sorry having read it again, I started to think this will cause regression.
> 
> inotify is limited resource, limited by `fs.inotify.max_user_instances`. This will spawn as many inotify instances as the containers. When users drain node, there will be more instances than the limit, and the watcher will fail.

@bitoku thanks, I didn't think from the node drain angle. I'm thinking of calling NewWatcher() once and adding watchers to it when containers are created. The value /proc/sys/fs/inotify/max_user_instances is 128 in many systems but max_user_watches is a very large number (min 8192). 

Will take another day digging into this and will propose a fix if it works.

bitoku:
@ngopalak-redhat 
What was the original motivation of this fix?
Actually we have a watcher already for exitFiles (it watches exitFiles directory)
https://github.com/cri-o/cri-o/blob/1c0b7145fe1f2694a2d070e0cee9337653479e3b/server/server.go#L851-L865

I think we can optimize it in different way though I don't have any concrete idea yet.

ngopalak-redhat:
> @ngopalak-redhat What was the original motivation of this fix? Actually we have a watcher already for exitFiles (it watches exitFiles directory)
> 
> https://github.com/cri-o/cri-o/blob/1c0b7145fe1f2694a2d070e0cee9337653479e3b/server/server.go#L851-L865
> 
> I think we can optimize it in different way though I don't have any concrete idea yet.

@bitoku While working on that area of code I observed that we are waiting on a file getting created. So notify made more sense than Exponential backoff. Hence thought of fixing it. I was also looking at the same monitorExits() piece of code. Working on a idea to re-use it.

PR #10109, opened 2026-07-06
utils: improve GetDiskUsageStats performance and sparse file accuracy

ngopalak-redhat:
#### What type of PR is this?

/kind cleanup

While working on Reflinks/Deduping I came across this improvement. This does not have any impact on the functionality but is a good performance improvement when trying to get stats for image FS. I tested Reflinked/dedupped filesystem also and saw no impact on the functionality. Its only a pref improvement.

Replace filepath.Walk with ReadDir and manual recursion for better performance and accuracy on Linux:

- 6-8% faster across all workload sizes
- Uses ReadDir(-1) without sorting as in file.Walk
- Reports actual allocated blocks (stat.Blocks * 512) instead of logical file size (fi.Size())

Platform-specific implementation:
- Linux: New ReadDir-based implementation in filesystem_linux.go
- FreeBSD/Other: Keep original filepath.Walk in platform-specific files

I have done following test on Linux, 4-core Intel Xeon @ 2.80GHz (Ubuntu on GCP VM):
- Small (10 files):      17,687 ns → 16,648 ns  (5.9% faster)
- Medium (1,000 files):  1,377,015 ns → 1,283,460 ns  (6.8% faster)
- Large (10,000 files):  16,624,500 ns → 15,353,141 ns  (7.6% faster)

I could see a larger impact when invoking this in a loop from a test as it had overall improved the performance.

AI Usage:
- I used AI for this function getAllocatedSize() - Reviewed it fully to make sure that its correct

- Currently used recursion. I don't see an issue with the usage in this context

```release-note
Improve disk usage stats performance on linux
```

## Summary by CodeRabbit

* **New Features**
  * Added disk usage statistics for files and directories.
  * Reports total storage consumption and inode counts.
  * Supports Linux-specific allocated-block accounting and compatible behavior on other platforms.
  * Handles both individual files and recursively traversed directories.

### DIFF
--- a/utils/filesystem.go
+++ b/utils/filesystem.go
@@ -2,29 +2,8 @@ package utils
 
 import (
 	"os"
-	"path/filepath"
 	"syscall"
 )
-
-// GetDiskUsageStats accepts a path to a directory or file
-// and returns the number of bytes and inodes used by the path.
-func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, _ error) {
-	if err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
-		// Walk does not follow symbolic links
-		if err != nil {
-			return err
-		}
-
-		dirSize += uint64(info.Size())
-		inodeCount++
-
-		return nil
-	}); err != nil {
-		return 0, 0, err
-	}
-
-	return dirSize, inodeCount, nil
-}
 
 // IsDirectory tests whether the given path exists and is a directory. It
 // follows symlinks.
--- a/utils/filesystem_linux.go
+++ b/utils/filesystem_linux.go
@@ -1,0 +1,75 @@
+//go:build linux
+
+package utils
+
+import (
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+// GetDiskUsageStats accepts a path to a directory or file
+// and returns the actual allocated disk blocks and inodes used by the path.
+func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, err error) {
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	inodeCount++
+	dirSize += getAllocatedSize(fi)
+
+	if !fi.IsDir() {
+		return dirSize, inodeCount, nil
+	}
+
+	s, i, err := calculateDir(path)
+
+	return dirSize + s, inodeCount + i, err
+}
+
+func calculateDir(dirPath string) (dirSize, inodeCount uint64, err error) {
+	f, err := os.Open(dirPath)
+	if err != nil {
+		return 0, 0, err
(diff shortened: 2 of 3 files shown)


ngopalak-redhat:
cc: @asahay19 @haircommander

utils/filesystem_unsupported.go
```diff
@@ -0,0 +1,28 @@
...
+package utils
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// GetDiskUsageStats accepts a path to a directory or file
+// and returns the number of bytes and inodes used by the path.
+func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, _ error) {
```
haircommander, line 12: this looks the same as freebsd one, do we need to have both?

ngopalak-redhat: Addressed

ngopalak-redhat:
@haircommander I have addressed the comments. Can you review?

utils/filesystem_linux.go
```diff
@@ -0,0 +1,75 @@
...
+	f, err := os.Open(dirPath)
+	if err != nil {
+		return 0, 0, err
+	}
+	defer f.Close()
+
+	entries, err := f.ReadDir(-1)
+	if err != nil {
+		return 0, 0, err
+	}
```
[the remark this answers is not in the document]
ngopalak-redhat: The current os.Open + f.ReadDir(-1) pattern and the proposed os.ReadDir(dirPath) are functionally equivalent

ngopalak-redhat:
@haircommander Can you please review?

utils/filesystem_linux.go
```diff
@@ -0,0 +1,75 @@
...
+
+	for _, entry := range entries {
+		inodeCount++
+
+		info, err := entry.Info()
+		if err != nil {
+			return 0, 0, err
+		}
+
+		dirSize += getAllocatedSize(info)
```
haircommander, line 51: is there any increased risk vs Walk of stack overflow or OOM if there's a really deep directory

utils/filesystem.go
```diff
@@ -2,30 +2,9 @@ package utils
...
 import (
 	"os"
-	"path/filepath"
 	"syscall"
 )
 
-// GetDiskUsageStats accepts a path to a directory or file
-// and returns the number of bytes and inodes used by the path.
-func GetDiskUsageStats(path string) (dirSize, inodeCount uint64, _ error) {
-	if err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
```
haircommander, line 12: there's also now [filepath.WalkDir](https://pkg.go.dev/path/filepath#WalkDir)  which purports to be more efficient, I wonder if we get the same gains if we compare this change against that?

PR #10117, opened 2026-07-07
Populate container name and pod sandbox ID in NRI events

### LINKED ISSUE #10011: NRI CreateContainer must have ContainerName
### What happened?

It looks like that Container Name is not populated in NRI CreateContainer event. See https://github.com/kubernetes-sigs/cri-tools/pull/2118/

Test https://github.com/kubernetes-sigs/cri-tools/pull/2118/ passes on CI with CRI-O.

https://github.com/kubernetes-sigs/cri-tools/pull/2118/

_No response_

<details>

```console
$ crio --version
# paste output here
```

```console
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

<details>

</details>

nispriha:
#### What type of PR is this?

/kind bug

The `GetName()` method on `criContainer` in `server/nri-api.go` reads the container name solely from the OCI spec annotation `io.kubernetes.container.name`. This annotation originates as a CRI label set by the kubelet, so when a CRI client (like critest) doesn't set this label, the NRI `CreateContainer` event has an empty `ContainerName`.

This PR adds a fallback to `c.ctr.Metadata().GetName()`, which reads the container name from the CRI `ContainerConfig.Metadata.Name` field which is a required field that every CRI client populates.

Additionally, `GetPodSandboxID()` had the same annotation-only pattern and would return empty for CRI clients that don't set the annotation. This PR applies the same fallback, using `c.ctr.Sandbox()` when the annotation is missing.

Fixes #10011

The cri-tools NRI conformance test (https://github.com/kubernetes-sigs/cri-tools/pull/2118) currently skips on CRI-O.

**Before fix:**
```
[SKIPPED] spec discrepancy: runtime does not populate container name in NRI CreateContainer event metadata
Ran 0 of 157 Specs — 0 Passed | 157 Skipped
```

**After fix:**
```
STEP: verifying CreateContainer event has correct metadata
• [2.683 seconds]
Ran 2 of 157 Specs — 2 Passed | 155 Skipped
```

```release-note
NRI container lifecycle events now correctly include the container name and pod sandbox ID.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container name and sandbox ID resolution so the app now uses fallback values when annotations are missing or empty.
  * This helps prevent blank or incorrect container metadata from appearing in related views and logs.

### DIFF
--- a/server/nri-api.go
+++ b/server/nri-api.go
@@ -13,6 +13,7 @@ import (
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/runtime-tools/generate"
 	cri "k8s.io/cri-api/pkg/apis/runtime/v1"
+	kubeletTypes "k8s.io/kubelet/pkg/types"
 	"tags.cncf.io/container-device-interface/pkg/cdi"
 
 	"github.com/cri-o/cri-o/internal/annotations"
@@ -668,11 +669,27 @@ func (c *criContainer) GetID() string {
 }
 
 func (c *criContainer) GetPodSandboxID() string {
-	return c.GetSpec().Annotations[annotations.SandboxID]
+	if id := c.GetSpec().Annotations[annotations.SandboxID]; id != "" {
+		return id
+	}
+
+	if c.ctr != nil {
+		return c.ctr.Sandbox()
+	}
+
+	return ""
 }
 
 func (c *criContainer) GetName() string {
-	return c.GetSpec().Annotations["io.kubernetes.container.name"]
+	if name := c.GetSpec().Annotations[kubeletTypes.KubernetesContainerNameLabel]; name != "" {
+		return name
+	}
+
+	if c.ctr != nil {
+		return c.ctr.Metadata().GetName()
+	}
+
+	return ""
 }
 
 func (c *criContainer) GetStatus() *nri.ContainerStatus {


server/nri-api.go
```diff
@@ -672,7 +672,13 @@ func (c *criContainer) GetPodSandboxID() string {
 }
 
 func (c *criContainer) GetName() string {
-	return c.GetSpec().Annotations["io.kubernetes.container.name"]
+	if name := c.GetSpec().Annotations["io.kubernetes.container.name"]; name != "" {
+		return name
+	}
+	if c.ctr != nil {
```
saschagrunert, line 688: The `lint` CI is failing on this line and line 681:
```
server/nri-api.go:678:2: missing whitespace above this line (invalid statement above if) (wsl_v5)
server/nri-api.go:681:2: return with no blank line before (nlreturn)
```
Needs a blank line before this `if` block and before the final `return ""`. Something like:
```go
func (c *criContainer) GetName() string {
	if name := c.GetSpec().Annotations["io.kubernetes.container.name"]; name != "" {
		return name
	}

	if c.ctr != nil {
		return c.ctr.Metadata().GetName()
	}

	return ""
}
```

server/nri-api.go
```diff
@@ -672,7 +672,13 @@ func (c *criContainer) GetPodSandboxID() string {
 }
 
 func (c *criContainer) GetName() string {
-	return c.GetSpec().Annotations["io.kubernetes.container.name"]
+	if name := c.GetSpec().Annotations["io.kubernetes.container.name"]; name != "" {
+		return name
```
saschagrunert, line 685: nit: `"io.kubernetes.container.name"` is a raw string while the neighboring methods use `annotations.ContainerID` and `annotations.SandboxID`. Could you add a constant for this in the annotations package and use it here?

server/nri-api.go
```diff
@@ -672,7 +672,13 @@ func (c *criContainer) GetPodSandboxID() string {
 }
```
saschagrunert, line 681: Not introduced by this PR, but `GetPodSandboxID` has the same annotation-only pattern and would return empty for CRI clients that don't set the annotation. Worth applying the same fallback here (or tracking it as a follow-up)?

saschagrunert (changes requested):
Thanks for the fix, the approach looks right. A few things to address inline.

server/nri-api.go
nispriha: Thanks. Fixed this.

server/nri-api.go
nispriha: Added the constant and updated `GetName()` to use it.

server/nri-api.go
nispriha: Applied the same fallback for `GetPodSandboxID()`.

server/nri-api.go
```diff
@@ -668,11 +668,27 @@ func (c *criContainer) GetID() string {
...
+	return ""
 }
 
 func (c *criContainer) GetName() string {
-	return c.GetSpec().Annotations["io.kubernetes.container.name"]
+	if name := c.GetSpec().Annotations[annotations.KubernetesContainerName]; name != "" {
+		return name
+	}
+
+	if c.ctr != nil {
```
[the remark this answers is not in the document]
nispriha: The cri-tools NRI conformance test already validates this, adding assertions to the internal tests could be a follow-up.

internal/annotations/annotations.go
```diff
@@ -10,6 +10,9 @@ const (
 	// ContainerName is the container name annotation.
 	ContainerName = "io.kubernetes.cri-o.ContainerName"
 
+	// KubernetesContainerName is the Kubernetes container name annotation.
+	KubernetesContainerName = "io.kubernetes.container.name"
```
xw19: This duplicates an existing k8s.io/kubelet/pkg/types  

`KubernetesContainerNameLabel = "io.kubernetes.container.name"`

Maybe just use that.

nispriha: I have made the change to use that.

bitoku:
Did you confirm it passes https://github.com/kubernetes-sigs/cri-tools/pull/2118/changes#diff-516ed7b88775ea3ad52ebadd047a68fc5bbeab4bab0d499613adf8b1e369a9cbR326-R328 ?

Can you post the result here? It would be better if we can check from the CI, but a manual test result is also fine.

nispriha:
Verified locally with critest built from master (includes NRI tests from kubernetes-sigs/cri-tools#2118):
```
sudo ~/go/bin/critest \
  --runtime-endpoint unix:///var/run/crio/crio.sock \
  --image-endpoint unix:///var/run/crio/crio.sock \
  --ginkgo.focus="NRI.*container lifecycle" \
  -nri-socket /var/run/nri/nri.sock \
  --ginkgo.v
```
Screenshot From 2026-07-07 12-23-12

Screenshot From 2026-07-07 12-22-53


Screenshot From 2026-07-07 12-22-24

bitoku:
Thanks!
/approve
/retest
@cri-o/cri-o-maintainers PTAL

xw19 (commented):
/lgtm

Thanks!

PR #10120, opened 2026-07-07
seccomp: keep logging blocked syscalls in not stop notifier mode

### LINKED ISSUE #9872: SecComp Notifier function stops loging after first block syscall detected

adilGhaffarDev:
#### What type of PR is this?
/kind bug

The seccomp notifier annotation (`seccomp-notifier-action.crio.io`) has two
behaviors: `stop` terminates the container after a blocked syscall, and any
other value keeps the container running while CRI-O logs the syscall and
increments `container_runtime_crio_containers_seccomp_notifier_count_total`.

The notifier handler loop broke out after handling the first syscall
(`// We only catch the first syscall`). In `stop` mode that's harmless since
the container is terminated anyway, but in the non-stop path the handler
goroutine exited after one syscall ,so the container silently stopped
logging further blocked syscalls and the metric could never increase past 1
for the rest of its lifetime.

This removes the early `break` so every blocked syscall keeps being logged
and counted. The loop now only exits on a terminal fd error
(EBADF/ECANCELED/ENOENT) or when the seccomp fd is closed. A short backoff
is added after transient receive errors to avoid a hot spin, and the log
line now includes the container's labels for easier attribution.

Also adds unit tests for the handler loop (injectable libseccomp calls) and
updates the integration test to assert repeated logging.

Fixes https://github.com/cri-o/cri-o/issues/9872

This PR is taking over https://github.com/cri-o/cri-o/pull/9875 because of unresponsiveness from the contributor. 
I am adding previous contributor as a co-author.

```release-note
CRI-O now keeps logging blocked syscalls detected by the seccomp notifier instead of only reporting the first event when the notifier action is not "stop".
```
Co-authored-by: Asish Kumar <officialasishkumar@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Seccomp notifier logs now report the blocked syscall using a more readable, container-focused format.
* **Bug Fixes**
  * Seccomp-notifier annotation is validated immediately, providing clearer startup failures when required settings are missing.
  * Improved notifier watcher resilience: transient receive errors back off, the loop continues through validation/respond failures, and stops on terminal receive conditions.
  * If a container can’t be resolved for a notification, metrics/logging now fall back to an “unknown” container label.
* **Tests**
  * Expanded notifier test coverage and updated log assertions to match the new behavior and wording.

### DIFF
--- a/internal/config/seccomp/notifier.go
+++ b/internal/config/seccomp/notifier.go
@@ -87,17 +87,17 @@ type Notification struct {
 }
 
 // Ctx returns the context of the notification.
-func (n *Notification) Ctx() context.Context {
+func (n Notification) Ctx() context.Context {
 	return n.ctx
 }
 
 // ContainerID returns the container identifier for the notification.
-func (n *Notification) ContainerID() string {
+func (n Notification) ContainerID() string {
 	return n.containerID
 }
 
 // Syscall returns the syscall name for the notification.
-func (n *Notification) Syscall() string {
+func (n Notification) Syscall() string {
 	return n.syscall
 }
 
@@ -170,6 +170,15 @@ func NewNotifier(
 		return nil, fmt.Errorf("listen for seccomp socket: %w", err)
 	}
 
+	action, ok := v2.GetAnnotationValue(annotationMap, v2.SeccompNotifierAction)
+	if !ok {
+		if err := listener.Close(); err != nil {
+			log.Errorf(ctx, "Unable to close seccomp listener: %v", err)
+		}
+
+		return nil, fmt.Errorf("%s annotation not set on container", v2.SeccompNotifierAction)
+	}
+
 	go func() {
 		for {
 			conn, err := listener.Accept()
@@ -208,11 +217,6 @@ func NewNotifier(
 		}
 	}()
 
-	action, ok := v2.GetAnnotationValue(annotationMap, v2.SeccompNotifierAction)
-	if !ok {
-		return nil, fmt.Errorf("%s annotation not set on container", v2.SeccompNotifierAction)
-	}
-
 	return &Notifier{
 		listener:       listener,
 		syscalls:       sync.Map{},
@@ -222,7 +226,30 @@ func NewNotifier(
 	}, nil
 }
 
+// receiveErrorBackoff throttles the notifier polling loop after a transient
+// receive error to avoid a hot spin on repeated failures.
+const receiveErrorBackoff = 100 * time.Millisecond
+
+type notifierHandler struct {
+	notifReceive func(libseccomp.ScmpFd) (*libseccomp.ScmpNotifReq, error)
+	notifIDValid func(libseccomp.ScmpFd, uint64) error
+	notifRespond func(libseccomp.ScmpFd, *libseccomp.ScmpNotifResp) error
+}
+
 func handler(
+	ctx context.Context,
+	containerID string,
+	msgChan chan Notification,
+	fd libseccomp.ScmpFd,
+) {
+	notifierHandler{
+		notifReceive: libseccomp.NotifReceive,
+		notifIDValid: libseccomp.NotifIDValid,
+		notifRespond: libseccomp.NotifRespond,
+	}.handle(ctx, containerID, msgChan, fd)
+}
+
+func (h notifierHandler) handle(
 	ctx context.Context,
 	containerID string,
 	msgChan chan Notification,
@@ -230,9 +257,15 @@ func handler(
 ) {
 	defer unix.Close(int(fd))
 	for {
-		req, err := libseccomp.NotifReceive(fd)
+		req, err := h.notifReceive(fd)
 		if err != nil {
+			if errors.Is(err, unix.EBADF) || errors.Is(err, unix.ECANCELED) || errors.Is(err, unix.ENOENT) {
+				log.Infof(ctx, "Stopping notifier for container %s", containerID)
+				return
+			}
 			log.Errorf(ctx, "Unable to receive notification: %v", err)
+			time.Sleep(receiveErrorBackoff)
+
 			continue
 		}
 
@@ -257,18 +290,15 @@ func handler(
 		}
 
 		// TOCTOU check
-		if err := libseccomp.NotifIDValid(fd, req.ID); err != nil {
+		if err := h.notifIDValid(fd, req.ID); err != nil {
 			log.Errorf(ctx, "TOCTOU check failed: req.ID is no longer valid: %v", err)
 			continue
 		}
 
-		if err = libseccomp.NotifRespond(fd, resp); err != nil {
+		if err = h.notifRespond(fd, resp); err != nil {
 			log.Errorf(ctx, "Unable to send notification response: %v", err)
 			continue
 		}
-
-		// We only catch the first syscall
-		break
 	}
 }
 
--- a/internal/config/seccomp/notifier_test.go
+++ b/internal/config/seccomp/notifier_test.go
@@ -1,0 +1,207 @@
+//go:build seccomp && linux && cgo && test
+
+package seccomp_test
+
+import (
+	"context"
+	"errors"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	libseccomp "github.com/seccomp/libseccomp-golang"
+	"golang.org/x/sys/unix"
+
+	"github.com/cri-o/cri-o/internal/config/seccomp"
+)
+
+var errReceiveAgain = errors.New("receive again")
+
+var _ = t.Describe("Notifier", func() {
+	t.Describe("handler", func() {
+		It("should keep polling notifications until the seccomp fd closes", func() {
+			// Given
+			msgChan := make(chan seccomp.Notification, 2)
+			notifReceive := stubNotifReceive("getpid", "getppid", unix.ENOENT)
+
+			// When
+			seccomp.RunHandlerForTest(
+				context.Background(),
+				"ctr",
+				msgChan,
+				libseccomp.ScmpFd(-1),
+				notifReceive,
+				stubNotifIDValid,
+				stubNotifRespond,
+			)
+
+			// Then
+			Expect(msgChan).To(HaveLen(2))
+			Expect((<-msgChan).Syscall()).To(Equal("getpid"))
+			Expect((<-msgChan).Syscall()).To(Equal("getppid"))
+		})
+
+		It("should continue polling after a transient receive error", func() {
+			// Given
+			msgChan := make(chan seccomp.Notification, 1)
+			notifReceive := stubNotifReceive(errReceiveAgain, "getpid", unix.ENOENT)
+
+			// When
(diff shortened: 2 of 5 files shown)


adilGhaffarDev:
@bitoku @xw19 please check this one too.

adilGhaffarDev:
looks like flake is related to infra
/test ci-e2e-conmonrs

adilGhaffarDev:
test coverage is improved now. Please check

bitoku:
The original PR author addressed my comment.
https://github.com/cri-o/cri-o/pull/9875

adilGhaffarDev:
> The original PR author addressed my comment.
> 
> https://github.com/cri-o/cri-o/pull/9875

I can close mine if that is ready to merge.

bitoku:
@adilGhaffarDev

I might have misunderstood the feature.
First of all, there's no "log" mode for seccompNotifierAction, so it's not a bug, but a new feature.
Do you mind telling me why you need this feature, even though seccomp already has a way to audit?

adilGhaffarDev:
> @adilGhaffarDev
> 
> I might have misunderstood the feature. First of all, there's no "log" mode for seccompNotifierAction, so it's not a bug, but a new feature. Do you mind telling me why you need this feature, even though seccomp already has a way to audit?



You're right, there's no explicit "log" action. `SeccompNotifierActionStop = "stop"` is the only defined value, and `NewNotifier` just sets `stopContainers: action == SeccompNotifierActionStop`. So it's really "stop" vs. "not stop"(which this pr is calling log mode). I'll drop that wording in the title and description.

The PR isn't a new mode, it's a fix to the non-stop path. The handler loop used to `break` after the first syscall (`// We only catch the first syscall`). In stop mode that's fine (the container dies anyway), but in non-stop mode the goroutine exits after one syscall, so the container silently stops logging and the metric can never go past 1. Removing the `break` keeps every blocked syscall logged and counted.

And regarding seccomp audit logs, those are logged to the host kernel audit log, not container-attributed and not in CRI-O's metrics. The notifier gives per-container, label-attributed counts via the existing Prometheus metric, which is only meaningful if we don't stop at the first syscall.

I have updated pr title and description.

bitoku:
Thank you @adilGhaffarDev for the response.

> The PR isn't a new mode, it's a fix to the non-stop path.

It's a new mode. There hasn't been a non-stop path with `seccompNotifierAction`. 
The non-stop path would be not using `seccompNotifierAction`.
`seccompNotifierAction: stop` basically replaces ActErrno/ActKill/ActKillProcess/ActKillThread to ActNotify and stops the container in reaction to the notification.

> And regarding seccomp audit logs, those are logged to the host kernel audit log, not container-attributed and not in CRI-O's metrics. The notifier gives per-container, label-attributed counts via the existing Prometheus metric, which is only meaningful if we don't stop at the first syscall.

You want this change for the metric, correct? because not using `seccompNotifierAction` won't kill the container.
I'd say the feature is not for the metric, but for killing the container if the forbidden syscalls are used. The metric is auxiliary.
At least, this changes ActKill/ActKillProcess/ActKillThread to less restrictive ActErrno essentially, which I don't think is secure.

Also I'm concerned about the performance. 
The current implementation of `seccompNotifierAction` is not designed for non-stop path.
This will add latency to every blocked syscalls.

I think `seccompNotifierAction: log` can't be used for production at scale, and for debugging, the existing kernel audit log would be sufficient.

adilGhaffarDev:
> Thank you @adilGhaffarDev for the response.
> 
> > The PR isn't a new mode, it's a fix to the non-stop path.
> 
> It's a new mode. There hasn't been a non-stop path with `seccompNotifierAction`. The non-stop path would be not using `seccompNotifierAction`. `seccompNotifierAction: stop` basically replaces ActErrno/ActKill/ActKillProcess/ActKillThread to ActNotify and stops the container in reaction to the notification.
> 
> > And regarding seccomp audit logs, those are logged to the host kernel audit log, not container-attributed and not in CRI-O's metrics. The notifier gives per-container, label-attributed counts via the existing Prometheus metric, which is only meaningful if we don't stop at the first syscall.
> 
> You want this change for the metric, correct? because not using `seccompNotifierAction` won't kill the container. I'd say the feature is not for the metric, but for killing the container if the forbidden syscalls are used. The metric is auxiliary. At least, this changes ActKill/ActKillProcess/ActKillThread to less restrictive ActErrno essentially, which I don't think is secure.
> 
> Also I'm concerned about the performance. The current implementation of `seccompNotifierAction` is not designed for non-stop path. This will add latency to every blocked syscalls.
> 
> I think `seccompNotifierAction: log` can't be used for production at scale, and for debugging, the existing kernel audit log would be sufficient.


I dug into the history here and I think the "there's no log mode" framing isn't quite accurate, a `non-stop` logging/metrics behavior was intentional from day one.

The original PR that introduced this feature (#6120) documents it explicitly in its release note:

> Added seccomp notifier feature, which can be enabled by setting the annotation `io.kubernetes.cri-o.seccompNotifierAction` either to `stop` (for terminating the workload) **or anything else to just create metrics or logs**.

So the design has always been two modes:
- `stop` → terminate the workload after the timeout
- any other value → just emit logs/metrics for the blocked syscalls (no termination)

That non-stop path is also reflected in the current code: the annotation's *presence* gates the profile injection, and only `== "stop"` sets `stopContainers = true`.

This PR isn't adding a new mode, it's fixing the non-stop logging path, which currently stops emitting after the first blocked syscall. That regression is exactly what's reported in #9872 ("SecComp Notifier function stops logging after first block syscall detected"). Removing the `break` restores the documented "keep logging" behavior.

Happy to adjust the wording/docs if we want to make the two modes more explicit in `crio.conf.5`, but I wanted to establish that the log mode(non-stop) is a documented, pre-existing feature rather than something new.

bitoku:
Thanks @adilGhaffarDev, Looking at the old PR, they initially think only "stop".
But whether it's a bug or not is moot point. Sorry about that.

Any specific reason you want this feature? This is notable drawback of this feature.

> The current implementation of seccompNotifierAction is not designed for non-stop path.
This will add latency to every blocked syscalls.

Also if we move forward, this should be fixed.

> At least, this changes ActKill/ActKillProcess/ActKillThread to less restrictive ActErrno essentially, which I don't think is secure.

adilGhaffarDev:
> Thanks @adilGhaffarDev, Looking at the old PR, they initially think only "stop".
> But whether it's a bug or not is moot point. Sorry about that.

I do not believe the old PR was only initially for stop, it was clearly stated in the changelog from @saschagrunert  that stop and other actions like log and audit, probably @saschagrunert can also comment on this
 

> Any specific reason you want this feature? This is notable drawback of this feature.

this is not a new feature, it is a bug fix, the loop was closed resulting to only the first offending syscall is logged, any subsequent offending syscall will be blocked but not logged.
We explained detailed here and why we needed it - [SecComp Notifier function stops loging after first block syscall detected · Issue #9872 · cri-o/cri…](https://github.com/cri-o/cri-o/issues/9872)




> Also if we move forward, this should be fixed.
> 
> > At least, this changes ActKill/ActKillProcess/ActKillThread to less restrictive ActErrno essentially, which I don't think is secure.

I will fix these and push the changes.

adilGhaffarDev:
After checking code on main, I can see that this override isn't introduced by this PR, it already exists on `main`. The `isActionToOverride` helper in `injectNotifier` has rewritten `ActErrno`/`ActKill`/`ActKillProcess`/`ActKillThread` → `ActNotify` since the feature was first added in #6120:

https://github.com/cri-o/cri-o/blob/844603c4ba1b9dfcdefac37ee0efc5bc25a2b86f/internal/config/seccomp/notifier.go#L122

That rewrite happens once at container creation and is completely independent of the change in this PR. All this PR does is remove the `break` in the handler loop so the notifier keeps recording syscalls past the first one (fixing #9872), instead of stopping after the first.

So the `ActKill*`→notify behavior and the fact that it turns a kill into a notify-then-`ENOSYS`, is pre-existing and orthogonal to this fix, I can fix that in separate PR.

adilGhaffarDev:
@saschagrunert please check this one since this is related to https://github.com/cri-o/cri-o/pull/6120

saschagrunert (approved):
LGTM, pending Bitoku's suggestion to use `id` instead of `"unknown"` for the metric label in the `ctr == nil` fallback path.

adilGhaffarDev:
> LGTM, pending Bitoku's suggestion to use `id` instead of `"unknown"` for the metric label in the `ctr == nil` fallback path.

done

saschagrunert:
RPM failures seem not to be related.

saschagrunert (approved):

PR #10127, opened 2026-07-09
WIP: Dedup on image pull incrementally

ngopalak-redhat:

```release-note

```


server/dedup_scheduler.go
```diff
@@ -0,0 +1,176 @@
+package server
```
haircommander: probably best as a separate package, maybe under internal/storage?

Issue #7703, opened 2024-01-20
Set `LimitNOFILE=1024:524288` for `crio.service`


polarathene:
### What happened?

I was recently made aware of this configuration line ([contributed Oct 2016](https://github.com/cri-o/cri-o/pull/135)):

https://github.com/cri-o/cri-o/blob/91816d7e8a38df2c307811f80d2e34fb53f01a21/contrib/systemd/crio.service#L20

Quite a bit has changed since then, notably with systemd v240 release in 2018Q4. Both Docker and Containerd projects have recently removed the line from their configs to rely on the `1024:524288` default systemd v240 provides (_unless the system has been configured explicitly to some other value, which the system administrator may do so when they know they need higher limits_).

You can find insights related to those PRs, along with a third link to the Envoy project (_as an example of a popular software that presently does not raise it's soft limit or document that requirement, but has depended upon this implicit config in the environment_) where the linked comment details why the soft limit should be `1024` to avoid software incompatibility:

- https://github.com/moby/moby/pull/45534
- https://github.com/containerd/containerd/pull/8924
- https://github.com/envoyproxy/envoy/issues/31502#issuecomment-1868149343
- https://github.com/awslabs/amazon-eks-ami/pull/1535#issuecomment-1868076423

---

This issue is raised to suggest consider applying the same change.

**Either:**
- Remove the line like Docker and containerd have done
- Include `LimitNOFILE=1024:524288` with contextual comment.
  - Although it may be better to remove the line, instead relying implicitly on the system default.
  - Admins / users could also use a [drop-in service override unit](https://github.com/moby/moby/issues/45436#issuecomment-1534520189).

### What did you expect to happen?

For `LimitNOFILE` to have a soft limit of `1024`, so that software running in a container operates with the same environment defaults of the host system.

Raising the default soft limit should be done explicitly by the admin, or via the process that needs it implicitly (_see Python reproduction below for an example of this_).

### Commands

I am not familiar with `cri-o`, but the equivalent Docker commands demonstrate the difference (_which for `LimitNOFILE=1048576` can be more subtle, for example `postsrsd` would be <500ms vs 8 minutes_):

```console
# Demonstrating the impact on a python process with `LimitNOFILE=1048576`:
$ docker run --rm -it --ulimit "nofile=1048576" --volume './python_close_individual.py:/tmp/test.py' python:3.12-alpine3.19 ash -c 'time python3 /tmp/test.py 1048570 100'
115.63215670500358
[...]

PR #10135, opened 2026-07-14
Set default nofile ulimit to avoid EMFILE errors at scale

Chandan9112:
## Summary

CRI-O ships no `default_ulimits` for `nofile`. Containers inherit the systemd soft limit of 1024, which is insufficient for workloads like Ceph OSDs that maintain thousands of peer connections at scale (e.g., 222 OSDs × 3-5 connections each ≈ 1500+ FDs per OSD process).

This partially reverts the default introduced in #7703, which aligned with systemd's soft=1024 but proved too restrictive for real-world workloads at scale.

This PR sets `default_ulimits = ["nofile=65536:524288"]` so containers get an adequate soft limit by default:
- **Soft limit (65536):** Provides headroom for high-connection workloads while remaining low enough to avoid FD-iteration overhead in programs that close all descriptors before exec.
- **Hard limit (524288):** Matches the systemd v240+ default. Unchanged from what containers already inherit.

Admins can override via config or set `default_ulimits = []` to inherit from the CRI-O daemon process.

Fixes: https://issues.redhat.com/browse/OCPBUGS-98458

```release-note
Set default container nofile ulimit to 65536:524288 to prevent EMFILE errors for workloads with high file descriptor usage
```

## Test Evidence

**Cluster:** OCP 4.22.0-0.nightly-2026-07-05-180412 (3 masters + 3 workers)

### Before fix (soft nofile = 1024):
```
=== Current ulimits ===
Soft nofile: 1024
Hard nofile: 524288

=== Attempting to open 1500 file descriptors ===
FAILED at FD #1021: [Errno 24] Too many open files
Total FDs opened: 1021
```

### After fix (soft nofile = 65536):

Applied via drop-in config (`/etc/crio/crio.conf.d/99-nofile-fix.conf`), restarted CRI-O:
```
=== Current ulimits ===
Soft nofile: 65536
Hard nofile: 524288

=== Attempting to open 1500 file descriptors ===
SUCCESS: Opened all 1500 sockets
Total FDs opened: 1500
```

Container inspection confirms:
```
$ ulimit -n
65536
$ ulimit -Hn
524288
$ cat /proc/self/limits | grep "open files"
Max open files            65536                524288               files
```

## Changes

| File | Change |
|------|--------|
| `pkg/config/config.go` | Add `DefaultNofileUlimit` constant, set it in `DefaultRuntimeConfig()` |
| `pkg/config/config_test.go` | Test that default config validates and produces correct RLIMIT_NOFILE |
| `pkg/config/template.go` | Update config template example and opt-out documentation |
| `docs/crio.conf.5.md` | Update manpage with new default and opt-out info |

## Summary by CodeRabbit

* **New Features**
  * CRI-O now sets a default `nofile` (open-file) ulimit to `65,536` soft and `524,288` hard when no `default_ulimits` are configured.
* **Documentation**
  * Updated `crio.conf` and `crio` option docs to show the new `--default-ulimits` default value and clarify behavior when `default_ulimits` is an empty list.
* **Tests**
  * Added validation coverage to ensure the default `RLIMIT_NOFILE` ulimit is present and matches the expected soft/hard values.

### DIFF
--- a/docs/crio.8.md
+++ b/docs/crio.8.md
@@ -251,7 +251,7 @@ crio [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
 
 **--default-transport**="": A prefix to prepend to image names that cannot be pulled as-is. (default: "docker://")
 
-**--default-ulimits**="": Ulimits to apply to containers by default (name=soft:hard).
+**--default-ulimits**="": Ulimits to apply to containers by default (name=soft:hard). (default: "nofile=65536:524288")
 
 **--device-ownership-from-security-context**: Set devices' uid/gid ownership from runAsUser/runAsGroup.
 
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -122,8 +122,8 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 **default_runtime**="crun"
 The _name_ of the OCI runtime to be used as the default. This option supports live configuration reload.
 
-**default_ulimits**=[]
-A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example:"nofile=1024:2048". If nothing is set here, settings will be inherited from the CRI-O daemon.
+**default_ulimits**=["nofile=65536:524288"]
+A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example: "nofile=65536:524288". If set to an empty list, settings will be inherited from the CRI-O daemon.
 
 **no_pivot**=false
 If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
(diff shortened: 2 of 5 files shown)


saschagrunert (approved):

saschagrunert:
/lgtm cancel

CI fails. Other than that LGTM. I was a bit concerned about the upgrade behavior but that should be non-blocking, right?

Chandan9112:
> /lgtm cancel
> 
> CI fails. Other than that LGTM. I was a bit concerned about the upgrade behavior but that should be non-blocking, right?

Yes, non-blocking. On upgrade, only clusters without an explicit default_ulimits in crio.conf get the new default. We're raising the soft limit (1024 → 65536), so no workload breaks — nothing depends on a low nofile. Explicit configs take precedence and are unaffected.

saschagrunert (approved):

bitoku:
We don't know what workloads are running on the cluster, and it's not appropriate to configure a default based on a specific use case.

PR #10137, opened 2026-07-14
Fix flaky duplicate-events registry reload test

### LINKED ISSUE #10132: Flaky: "system handles duplicate events for the same file" in reload_system.bats
Test "system handles duplicate events for the same file" in `test/reload_system.bats` asserts that rapid `sed -i` calls on the same config file produce exactly 1 registry reload. This fails intermittently because the debounce timer in `server/server.go` can expire between `sed -i` invocations on a loaded CI runner, causing more than one reload.

Example failure: https://github.com/cri-o/cri-o/actions/runs/29101404529/job/86391708085?pr=10066

The adjacent test "system handles burst of configuration changes without excessive reloads" already covers debounce behavior. This test should be removed or its assertion relaxed.

xw19:
Fix Flaky tests

#### What type of PR is this?

/kind ci

Fixes #10132 

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * Improved validation of configuration reloads during bursty file changes by verifying reload timestamps and enforcing debounce spacing (minimum spacing with small timing slack).
  * Updated relevant test flows to wait for successful configuration application before checking reload behavior.
  * Enhanced coverage to ensure duplicate file events don’t cause excessive or overly rapid successive reloads.

### DIFF
--- a/test/reload_system.bats
+++ b/test/reload_system.bats
@@ -37,13 +37,31 @@ function expect_log_failure() {
 	wait_for_log "system registries reload failed"
 }
 
-function count_log_entries() {
-	local search_pattern="$1"
-	local count
+function assert_debounced_reloads() {
+	# A burst normally coalesces into one reload, but a slow machine may fire
+	# the debounce timer more than once; consecutive reloads must still be at
+	# least the debounce interval apart (50ms slack for timestamp lag). We use
+	# the "reloading" log line (timer firing) rather than "Applied" (reload
+	# done) to measure timer cadence; success is checked by expect_log_success.
+	local -a reload_times
+	readarray -t reload_times < <(grep "reloading registries configuration" "$CRIO_LOG" |
+		grep -oE 'time="[^"]*"' | cut -d'"' -f2)
 
-	count=$(grep -c "$search_pattern" "$CRIO_LOG")
+	[ "${#reload_times[@]}" -ge 1 ]
 
-	echo "$count"
+	if [ "${#reload_times[@]}" -gt 1 ]; then
+		local timestamp prev_ns=0 curr_ns
+		for timestamp in "${reload_times[@]}"; do
+			if ! curr_ns=$(date -d "$timestamp" +%s%N); then
+				echo "could not parse reload timestamp: $timestamp" >&2
+				return 1
+			fi
+			if [ "$prev_ns" -gt 0 ]; then
+				[ $((curr_ns - prev_ns)) -ge 150000000 ]
+			fi
+			prev_ns=$curr_ns
+		done
+	fi
 }
 
 @test "reload system registries should succeed" {
@@ -113,6 +131,7 @@ function count_log_entries() {
 	drop_in_for_auto_reload_registries
 	start_crio
 
+	# when
 	sed -i 's;true;false;g' "$CONTAINER_REGISTRIES_CONF"
 	sed -i 's;true;false;g' "$CONTAINER_REGISTRIES_CONF_DIR/01-registry.conf"
 	cat << EOF >> "$CONTAINER_REGISTRIES_CONF_DIR/02-registry.conf"
@@ -120,10 +139,11 @@ function count_log_entries() {
 location = 'registry.k8s.io/pause'
 blocked = false
 EOF
-	sleep 1
+
 	# then
-	log_count=$(count_log_entries "Applied new registry configuration")
-	[ "$log_count" -eq 1 ]
+	expect_log_success
+	sleep 1
+	assert_debounced_reloads
 }
 
 @test "system handles duplicate events for the same file" {
@@ -135,8 +155,9 @@ EOF
 	for _ in {1..5}; do
 		sed -i 's;true;false;g' "$CONTAINER_REGISTRIES_CONF_DIR/01-registry.conf"
 	done
-	sleep 1
+
 	# then
-	log_count=$(count_log_entries "Applied new registry configuration")
-	[ "$log_count" -eq 1 ]
+	expect_log_success
+	sleep 1
+	assert_debounced_reloads
 }


ngopalak-redhat:
@xw19  How about getting the log_count before and after the test and making sure its 1?

xw19:
@ngopalak-redhat Thanks for looking into it!

But, before/after would protect the count from unrelated reloads earlier in the log, but it turns out it doesn't fix the underlying flake: the watcher coalesces events with a 200ms debounce timer in  `server/server.go`, and this test generates events by spawning `sed -i` five times. 

So I feel if the ci server is under pressure one >200ms gap between two sed invocations means the timer fires twice so even the difference can legitimately may be more than 1.

So, I need to circle back.

ngopalak-redhat:
@xw19 The test itself seems wrong:
The first sed -i 's;true;false;g' replaces all true with false. The remaining 4 iterations find no true left to replace, so they're no-ops that don't actually modify the file.

Try:
```
  for i in {1..5}; do
      if ((i % 2)); then
          sed -i 's;true;false;g' "$CONTAINER_REGISTRIES_CONF_DIR/01-registry.conf"
      else
          sed -i 's;false;true;g' "$CONTAINER_REGISTRIES_CONF_DIR/01-registry.conf"
      fi
  done
```

xw19:
> @xw19 The test itself seems wrong: The first sed -i 's;true;false;g' replaces all true with false. The remaining 4 iterations find no true left to replace, so they're no-ops that don't actually modify the file.
>

Yes and no. The sed -i doesn't care about matching contents it will create a temp file and write it out so regardless of content we will get 5 inode replacement and 5 inotify events.

ngopalak-redhat:
> > @xw19 The test itself seems wrong: The first sed -i 's;true;false;g' replaces all true with false. The remaining 4 iterations find no true left to replace, so they're no-ops that don't actually modify the file.
> 
> Yes and no. The sed -i doesn't care the about matching contents it will create a temp file and write it out so regardless of content we will get 5 inode replacement and 5 inotify events.

It would be easier to create the file 5 times instead of "sed" right?

xw19:
> It would be easier to create the file 5 times instead of "sed" right?

I think that sed part is doing its job its not a problem. For our readability and easier understanding we can do that

How about we put a log statement in debounce something like

```
	if event.Op&(fsnotify.Write|fsnotify.Create|fsnotify.Remove|fsnotify.Chmod) != 0 {
				// Reset timer if exists, else create a new one.
				if timer != nil {
					log.Debugf(ctx, "Debouncing registry reload, coalescing event for %q", event.Name)
					timer.Reset(debounceDuration)
				} else {
					timer = time.AfterFunc(debounceDuration, func() {
						reloadChannel <- event.Name
					})
				}
			}

```

And check for this log?

xw19:
@bitoku @ngopalak-redhat Please review my PR.

The debounce will ensure that no reload happens if all events arrive within 200ms but because of system load it doesn't arrive instead of failing we will check the timestamps and see if those logs are outside debounce window.

test/reload_system.bats
```diff
@@ -37,13 +37,28 @@ function expect_log_failure() {
...
+	if [ "${#reload_times[@]}" -gt 1 ]; then
+		local timestamp prev_ns=0 curr_ns
+		for timestamp in "${reload_times[@]}"; do
+			curr_ns=$(date -d "$timestamp" +%s%N)
+			if [ "$prev_ns" -gt 0 ]; then
+				[ $((curr_ns - prev_ns)) -ge 180000000 ]
+			fi
+			prev_ns=$curr_ns
+		done
+	fi
```
bitoku, line 64: I would want explanation about what each line does. Is it only me?

bitoku (commented):
I have one comment though not blocking.
/approve

test/reload_system.bats
xw19: I get it what it looks sophisticated.

xw19:
@cri-o/cri-o-maintainers PTAL

test/reload_system.bats
```diff
@@ -37,13 +37,28 @@ function expect_log_failure() {
...
+		grep -oE 'time="[^"]*"' | cut -d'"' -f2)
+
+	[ "${#reload_times[@]}" -ge 1 ]
+
+	if [ "${#reload_times[@]}" -gt 1 ]; then
+		local timestamp prev_ns=0 curr_ns
+		for timestamp in "${reload_times[@]}"; do
+			curr_ns=$(date -d "$timestamp" +%s%N)
+			if [ "$prev_ns" -gt 0 ]; then
+				[ $((curr_ns - prev_ns)) -ge 180000000 ]
```
saschagrunert: 20ms slack might be tight on a heavily loaded CI runner. The jitter comes from scheduling delay between the Go timer firing and logrus recording the timestamp. If the first log entry is delayed while the second is recorded promptly, the apparent gap compresses. Consider widening to 150000000 (50ms slack) to make this more resilient; the test still validates that debounce exists, just with more margin for timestamp noise.

test/reload_system.bats
```diff
@@ -37,13 +37,28 @@ function expect_log_failure() {
...
+	local -a reload_times
+	readarray -t reload_times < <(grep "reloading registries configuration" "$CRIO_LOG" |
+		grep -oE 'time="[^"]*"' | cut -d'"' -f2)
+
+	[ "${#reload_times[@]}" -ge 1 ]
+
+	if [ "${#reload_times[@]}" -gt 1 ]; then
+		local timestamp prev_ns=0 curr_ns
+		for timestamp in "${reload_times[@]}"; do
+			curr_ns=$(date -d "$timestamp" +%s%N)
```
saschagrunert: If the timestamp extraction on line 47-48 yields an empty string (e.g., log format changes), `date -d ""` errors and `curr_ns` is empty. Bash arithmetic treats empty as 0, so the assertion on line 57 silently fails with a confusing message. A guard would help debuggability:
```suggestion
			curr_ns=\$(date -d "\$timestamp" +%s%N) || fail "could not parse timestamp: \$timestamp"
```

test/reload_system.bats
```diff
@@ -37,13 +37,28 @@ function expect_log_failure() {
...
-	echo "$count"
+function assert_debounced_reloads() {
+	# A burst of events should normally coalesce into a single reload. On a
+	# slow machine a >200ms gap between two events legitimately fires the
+	# debounce timer again - in that case consecutive reloads must still
+	# respect the debounce interval. The threshold keeps 20ms slack since
+	# log timestamps can lag the timer firings slightly.
+	local -a reload_times
+	readarray -t reload_times < <(grep "reloading registries configuration" "$CRIO_LOG" |
+		grep -oE 'time="[^"]*"' | cut -d'"' -f2)
```
saschagrunert, line 48: Worth noting (maybe a one-line comment) that this intentionally greps for the "reloading" message (`server.go:1080`, when the debounce timer fires) rather than the "Applied" message (`reload.go:217`, after reload completes) that the old test counted. The distinction matters: this measures timer firing cadence, which is what debounce validation needs. The "Applied" check is handled separately by `expect_log_success`.

xw19: @saschagrunert  Added please check.

PR #10142, opened 2026-07-14
Check free port more consistently in bats test

### LINKED ISSUE #10134: Flaky: "cri-o doesn't log the error when conmon doesn't exit"
`test/conmon_track.bats:31` fails when the metrics port returned by `free_port()` is already in use by the time CRI-O tries to bind it.

```
level=fatal msg="create metrics endpoint on 127.0.0.1:40047: creating listener: listen tcp 127.0.0.1:40047: bind: address already in use"
```

[Example failure](https://github.com/cri-o/cri-o/actions/runs/29098642233/job/86420711835) (integration / userns / crun / amd64)

ngopalak-redhat:
#### What type of PR is this?

/kind bug

The test case was flaky. Python was required to find a free port.

Fixes #10134

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic network port selection for test and development workflows by attempting multiple candidate ports and checking availability.
  * Added clearer failure handling when no available port can be found, including returning an error status and emitting a message.

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -247,7 +247,16 @@ function check_journald() {
 
 # get a random available port
 function free_port() {
-    python -c 'import socket; s=socket.socket(); s.bind(("", 0)); print(s.getsockname()[1]); s.close()'
+    local port
+    for _ in {1..5}; do
+        port=$((RANDOM % 22768 + 10000))
+        if ! host_and_port_listens ".*" "$port"; then
+            echo "$port"
+            return 0
+        fi
+    done
+    echo "ERROR: free_port: could not find a free port after 5 attempts" >&2
+    return 1
 }
 
 # Check whether a port is listening


ngopalak-redhat:
@coderabbitai review

ngopalak-redhat:
@coderabbitai review

ngopalak-redhat:
@bitoku I'm waiting for another run of the integration test, you can review.

ngopalak-redhat:
@bitoku Please review

bitoku (commented):
I think the problem is TOCTOU. I don't get how it fixes the flakiness.

xw19:
@bitoku  10000–32767 sits entirely below the default range. So the kernel will never allocates these ports to an outgoing connection. Therefore its very less likely to get into a collision. I think we should give this a shot.

@ngopalak-redhat correct me if I am wrong.

bitoku:
@xw19 aha, thanks
/approve

ngopalak-redhat:
> @bitoku 10000–32767 sits entirely below the default range. So the kernel will never allocates these ports to an outgoing connection. Therefore its very less likely to get into a collision. I think we should give this a shot.
> 
> @ngopalak-redhat correct me if I am wrong.

Thanks @xw19 @bitoku

PR #10143, opened 2026-07-14
statsserver: add active_anon, inactive_anon and AnonHugePages memory metrics

cragr:
#### What type of PR is this?

/kind feature

Adds five per-container memory metrics, read from the `memory.stat` keys `active_anon`, `inactive_anon`, `anon_thp`, `shmem_thp` and `file_thp`:

- `container_memory_active_anon_bytes`
- `container_memory_inactive_anon_bytes`
- `container_memory_anon_thp_bytes`
- `container_memory_shmem_thp_bytes`
- `container_memory_file_thp_bytes`

These let consumers distinguish hot (active) from cold (inactive) anonymous memory and see how much of a container's memory is backed by transparent hugepages. Relevant for KubeVirt/OpenShift Virtualization, where guest memory is anonymous: active memory is an early warning signal for swap pressure before PSI reflects real impact, and the THP metrics show how much is hugepage-backed.

All three THP flavors are exposed because `anon_thp` alone does not cover every workload. KubeVirt normally backs guest memory with plain anonymous memory, but with QEMU memfd backing the guest memory is shmem and its hugepages are counted towards `shmem_thp` instead (gated by `/sys/kernel/mm/transparent_hugepage/shmem_enabled`); PyTorch has the same pattern. `file_thp` covers file-backed workloads such as Elasticsearch.

These are transparent hugepages, not hugetlb — hugetlb is already reported by the existing `container_hugetlb_*` metrics. The metric names use `thp` to mirror the kernel's own naming, and each help string says so explicitly.

They are gated behind a new `memoryExtra` value for `included_pod_metrics`, separate from `memory`. `memory` keeps the metrics that have a cAdvisor equivalent, so the two sets can be enabled and disabled independently; `memoryExtra` is in `AvailableMetrics`, so `all` still includes them. The metric names themselves are unchanged and keep following the existing `container_memory_*` convention — the split is only in the configuration.

They follow the existing `container_memory_*` pattern in `internal/lib/statsserver` (descriptor + `availableMetricDescriptors` registration + `generateContainerMemoryExtraMetrics`). The cgroup version is handled explicitly via helpers, rather than relying on a missing map key returning zero:

- `computeAnonMemory` reads `active_anon`/`inactive_anon` on cgroup v2 and the `total_`-prefixed hierarchical equivalents on cgroup v1, so the active and inactive anonymous metrics report real values on both versions.
- `computeTransparentHugepages` returns zero on cgroup v1, which has no THP accounting at all.

Both helpers take the cgroup version as a parameter, with `generateContainerMemoryExtraMetrics` reading `node.CgroupIsV2()` once and passing it down, so both branches are unit-testable on any host.

No CRI proto or vendor change is required — the values already exist in the libcontainer `MemoryStats.Stats` map.

This originated as google/cadvisor#3915; per the [cAdvisor roadmap](https://github.com/google/cadvisor/blob/master/docs/roadmap.md) and [KEP-2371](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2371-cri-pod-container-stats), new container-stats metrics belong in the runtime rather than cAdvisor, so it is proposed here instead.

None

Thanks @michalskrivanek for the review — the cgroup v1 handling and the additional `shmem_thp`/`file_thp` metrics both come from your feedback.

Thanks @dgrisonnet for the naming feedback. Dropped the `total_` prefix in favour of the cgroup v2 naming, and renamed the hugepage metrics to `thp` so they mirror the kernel and are not confused with hugetlb; the help text now says that outright and points at the `container_hugetlb_*` metrics. `computeAnonMemory` and `computeTransparentHugepages` now take the cgroup version as a parameter, so the cgroup v1 key names and the zeroed THP values are exercised by unit tests on any host rather than only on a v1 node — the gap I flagged in an earlier revision of this description is closed.

Verified on a cgroup v2 host: build, `go vet` and the `statsserver` unit tests all pass, and the metrics emit the expected values from the fake `memory.stat`. `TestComputeAnonMemory` and `TestComputeTransparentHugepages` table-test both cgroup versions directly, so the v1 paths now run regardless of the host's cgroup version.

```release-note
Added container memory metrics `container_memory_active_anon_bytes` and `container_memory_inactive_anon_bytes` (cgroup v1 and v2), and `container_memory_anon_thp_bytes`, `container_memory_shmem_thp_bytes` and `container_memory_file_thp_bytes` (cgroup v2 only). These are gated behind a new `memoryExtra` value for `included_pod_metrics`, separate from `memory`.
```

## Summary by CodeRabbit

* **New Features**
  * Added new container memory gauge metrics for **active anonymous bytes**, **inactive anonymous bytes**, **anonymous transparent huge pages bytes**, **shared transparent huge pages bytes**, and **file transparent huge pages bytes**.
  * These metrics are included in the existing memory metric set and appear when requesting memory metrics.
* **Tests**
  * Expanded unit tests to confirm the new metric names and byte values from the memory statistics fixture, including cgroup-version-specific transparent-hugepage behavior.


cragr:
@mrunalp @littlejawa @QiWang19 CI is green now (aside from unrelated non-blocking rawhide rpm-build failures). ready for review whenever you get a chance.

cragr:
@bitoku @mrunalp @littlejawa @QiWang19  ready for review whenever you get a chance.

cragr:
@bitoku @mrunalp @littlejawa @QiWang19 ready for review whenever you get a chance.

michalskrivanek:
Hi @cragr, thank you so much for posting this! This would be very beneficial for kubevirt where it uses anonymous memory for the guest memory backing.

I think the `active/inactive_anon` is good and works for both cgroups v1 and v2. But for THP it's more complicated. I think there's no equivalent in v1 for these, so probably makes sense to use a helper (similar to e.g. `computeMemoryMetricValues`) to handle that?

In addition, depending on the actual backing kubevirt uses "normal" anonymous memory (`anon_thp`) but in some configurations where shared mem is needed it uses shmem (via QEMU's memfd memory backing) - in that case the THP considerations are the same, THPs can be enabled by `/sys/kernel/mm/transparent_hugepage/shmem_enabled` and generally work the same as normal memory, but are counted towards `shmem_thp`. There are other users as well, e.g. PyTorch.
Similarly for file backed memory there's `file_thp` that (kubevirt doesn't use it, but it is be valid for elasticsearch and such).
I wonder if its worth adding all three then.

cragr:
Thanks @michalskrivanek, good catches. Pushed [619fb92c2](https://github.com/cri-o/cri-o/pull/10143/commits/619fb92c2).

**cgroup v1:** Worse than just THP, as it turns out. v1 uses `total_active_anon`/`total_inactive_anon` for the subtree totals, and I was reading the v2 names on both versions, so both anon metrics were quietly reporting 0 on v1. Now behind two helpers, following the `computeFileMapped` pattern you pointed at: `computeAnonMemory` picks the right key names per version, and `computeTransparentHugepages` returns zero on v1, which has no THP accounting at all.

**shmem_thp and file_thp:** Added both, v2 only:

- `container_memory_shmem_hugepages_bytes`
- `container_memory_file_hugepages_bytes`

Your memfd point sold me. A dashboard watching only `anon_thp` would show zero THP for a VM that's actually fully hugepage-backed. `file_thp` was a small addition alongside.

On `shmem_enabled`, did you have something in mind, or was that background for why `shmem_thp` matters?

One caveat: only tested on a v2 host. The test asserts the v1 key names and zeroed THP values, but that branch only runs on a v1 host, so the v1 path is reviewed, not run.

michalskrivanek:
@cragr that was quick, overall it looks good to me!
I don't know if anyone still uses cgroups v1 at all, considering it's being completely removed by https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/5573-remove-cgroup-v1/README.md, so defaulting to 0 sounds good to me.
Regarding file and shmem, I meant that unlike THP for anonymous pages it's just less common to be enabled on nodes. Anonymous THP are in use on CoreOS, RHEL, SUSE, Ubuntu by default, shmem is more recent (kernel 6.8+) and needs to be turned on on common distros.

haircommander:
@dgrisonnet PTAL

internal/lib/statsserver/descriptors.go
```diff
@@ -151,6 +151,31 @@ var (
 		Help:      "Current working set in bytes.",
 		LabelKeys: baseLabelKeys,
 	}
+	containerMemoryTotalActiveAnon = &types.MetricDescriptor{
+		Name:      "container_memory_total_active_anon_bytes",
```
dgrisonnet: I think we should favorise the cgroups v2 naming

```suggestion
		Name:      "container_memory_active_anon_bytes",
```

internal/lib/statsserver/descriptors.go
```diff
@@ -151,6 +151,31 @@ var (
...
+		Help:      "Current total active anonymous in bytes.",
+		LabelKeys: baseLabelKeys,
+	}
+	containerMemoryTotalInactiveAnon = &types.MetricDescriptor{
+		Name:      "container_memory_total_inactive_anon_bytes",
+		Help:      "Current total inactive anonymous in bytes.",
+		LabelKeys: baseLabelKeys,
+	}
+	containerMemoryAnonHugepages = &types.MetricDescriptor{
+		Name:      "container_memory_anon_hugepages_bytes",
```
dgrisonnet: I am worried that `hugepages` might create a confusion for users between hugetlb and thp. The HELP text is making it clearer, but maybe calling the metric `container_memory_anon_thp_bytes` would be better. This would mirror what the kernel does.

internal/lib/statsserver/descriptors.go
cragr: Agreed, done in c63a76610.  Applied to both:

- `container_memory_total_active_anon_bytes` → `container_memory_active_anon_bytes`
- `container_memory_total_inactive_anon_bytes` → `container_memory_inactive_anon_bytes`

The cgroup v1 code path still reads the `total_`-prefixed keys, since that is what v1 actually reports; only the metric names changed.

internal/lib/statsserver/descriptors.go
cragr: Good point.  I can see where the hugetlb/THP could caause confusion. Renamed all three to mirror the kernel::

- `container_memory_anon_hugepages_bytes` → `container_memory_anon_thp_bytes`
- `container_memory_shmem_hugepages_bytes` → `container_memory_shmem_thp_bytes`
- `container_memory_file_hugepages_bytes` → `container_memory_file_thp_bytes`

Thanks for the help @dgrisonnet

internal/lib/statsserver/descriptors.go
```diff
@@ -151,6 +151,31 @@ var (
...
+		LabelKeys: baseLabelKeys,
+	}
+	containerMemoryInactiveAnonBytes = &types.MetricDescriptor{
+		Name:      "container_memory_inactive_anon_bytes",
+		Help:      "Current inactive anonymous memory in bytes.",
+		LabelKeys: baseLabelKeys,
+	}
+	containerMemoryAnonTHPBytes = &types.MetricDescriptor{
+		Name:      "container_memory_anon_thp_bytes",
+		Help:      "Amount of anonymous memory backed by transparent hugepages (THP) in bytes. This is not hugetlb memory, which is reported by the container_hugetlb_* metrics. Requires anonymous THP to be enabled, which is the default on most distributions. Always zero on cgroup v1.",
```
dgrisonnet: Now that the name of the metric clarify that it is about THP, I think we can omit "This is not hugetlb memory, which is reported by the container_hugetlb_* metrics."

cragr: Agreed.  Snippit omitted in [c5874ed](https://github.com/cri-o/cri-o/pull/10143/changes/c5874ed9353ce82df4d84b701eab2aa67366f560)

dgrisonnet (approved):
/lgtm

dgrisonnet:
@cri-o/cri-o-maintainers for approval

fabiand:
@saschagrunert @haircommander @giuseppe @fidencio (!) any of you who could help here?

giuseppe (approved):
LGTM

internal/lib/statsserver/descriptors.go
```diff
@@ -151,6 +151,31 @@ var (
 		Help:      "Current working set in bytes.",
 		LabelKeys: baseLabelKeys,
 	}
+	containerMemoryActiveAnonBytes = &types.MetricDescriptor{
```
haircommander, line 154: one nit: I find myself wanting a separate category for these since they're not emulating existing cadvisor metrics. could be named container_memory_extra or something, not sure. WDYT @dgrisonnet 

/hold

dgrisonnet: I feel like having a separate category in the name will be confusing to the users.
I was also debating what we should do about the new CRI-O specific container metrics, but I think the better UX is to stick to the existing conventions that users are already familiar with and that we are inheriting from cadvisor.

One problem I could see is that users migrating between runtimes might see discrepancy in the available metrics.
Maybe we could add the categorization in the HELP texts? CRI metrics could have the [CRI] tag and the other [Extra]?

haircommander: I guess I mostly want a way to toggle off these metrics but allow the memory metrics to be on. so a separate field in `IncludedPodMetrics` would work for me

cragr:
Pushed [01b52b708](https://github.com/cri-o/cri-o/pull/10143/commits/01b52b708) with a separate included_pod_metrics value, memoryExtra, carrying the five new metrics.

@haircommander that should give you the toggle you wanted.  Memory keeps the cAdvisor-equivalent set, so you can run with memory on and these off. memoryExtra is in AvailableMetrics, so all still picks them up.

@dgrisonnet metric names are unchanged and stay on the existing convention; the split is only in the config, so nothing new shows up in the metric names themselves. On the [CRI]/[Extra] HELP tags, I left those out here, since to mean anything they'd need to go on all 61 descriptors, and classifying some of them (the container_pressure_* ones especially) is really a call about CRI-O's metrics rather than something to settle in this PR. Let me know if the split is more clear-cut than I'm assuming.

PR #10150, opened 2026-07-16
docs: clarify default_ulimits inheritance under systemd cgroup manager

### LINKED ISSUE #10006: conmon spawned by CRI-O does not inherit rlimits from the daemon
### What happened?

Related: containers/libkrun#704

The [documentation](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md) claims that 

> default_ulimits ... If nothing is set here, settings will be inherited from the CRI-O daemon.

However, with a setup that does not have `default_ulimits` set, it seems that conmon, and by extension crun and the container processes, are started with a smaller `nofiles` soft limit:

```console
$ ps aux | grep /usr/bin/crio
root     1116296  1.3  0.1 4067972 154112 ?      Ssl  21:37   0:04 /usr/bin/crio
hkj      1120908  0.0  0.0   7212  2220 pts/0    S+   21:43   0:00 grep /usr/bin/crio
$ cat /proc/1116296/limits
Limit                     Soft Limit           Hard Limit           Units
Max cpu time              unlimited            unlimited            seconds
Max file size             unlimited            unlimited            bytes
Max data size             unlimited            unlimited            bytes
Max stack size            8388608              unlimited            bytes
Max core file size        unlimited            unlimited            bytes
Max resident set          unlimited            unlimited            bytes
Max processes             1048576              1048576              processes
Max open files            524287               524288               files
Max locked memory         8388608              8388608              bytes
Max address space         unlimited            unlimited            bytes
Max file locks            unlimited            unlimited            locks
Max pending signals       509507               509507               signals
Max msgqueue size         819200               819200               bytes
Max nice priority         0                    0
Max realtime priority     0                    0
Max realtime timeout      unlimited            unlimited            us
$ ps aux | grep conmon | head -n1 # just a random one spawned by cri-o, this node is only running cri-o and thus all conmon processes are from cri-o
root        3798  0.0  0.0   7728  2180 ?        Ss   Jun02   0:00 /usr/bin/conmon -b /run/containers/storage/overlay-containers/b96a6fe911c3250913b15e6fe00e46f95d718e2a8a44504c5f41bb01d943600e/userdata -c b96a6fe911c3250913b15e6fe00e46f95d718e2a8a44504c5f41bb01d943600e --exit-dir /run/crio/exits -l /var/log/pods/metallb-system_metallb-speaker-jr4wj_37590200-2710-4695-b023-6788657ef549/b96a6fe911c3250913b15e6fe00e46f95d718e2a8a44504c5f41bb01d943600e.log --log-level info -n k8s_POD_metallb-speaker-jr4wj_metallb-system_37590200-2710-4695-b023-6788657ef549_0 -P /run/containers/storage/overlay-containers/b96a6fe911c3250913b15e6fe00e46f95d718e2a8a44504c5f41bb01d943600e/userdata/conmon-pidfile -p /run/containers/storage/overlay-containers/b96a6fe911c3250913b15e6fe00e46f95d718e2a8a44504c5f41bb01d943600e/userdata/pidfile --persist-dir /var/lib/containers/storage/overlay-containers/b96a6fe911c3250913b15e6fe00e46f95d718e2a8a44504c5f41bb01d943600e/userdata -r /usr/bin/crun --runtime-arg --root=/run/crun --socket-dir-path /run/crio --syslog -u b96a6fe911c3250913b15e6fe00e46f95d718e2a8a44504c5f41bb01d943600e -s
$ cat /proc/3798/limits
Limit                     Soft Limit           Hard Limit           Units
Max cpu time              unlimited            unlimited            seconds
Max file size             unlimited            unlimited            bytes
Max data size             unlimited            unlimited            bytes
Max stack size            8388608              unlimited            bytes
Max core file size        unlimited            unlimited            bytes
Max resident set          unlimited            unlimited            bytes
Max processes             1048576              1048576              processes
Max open files            1024                 524288               files
Max locked memory         8388608              8388608              bytes
Max address space         unlimited            unlimited            bytes
Max file locks            unlimited            unlimited            locks
Max pending signals       509507               509507               signals
Max msgqueue size         819200               819200               bytes
Max nice priority         0                    0
Max realtime priority     0                    0
Max realtime timeout      unlimited            unlimited            us
```

This does not seem to be a problem with normal container processes, as the hard limit are quite enough and most libraires/applications will try to raise the limit itself should the soft limit is reached.

However as noted in the linked related issue, this is particularly a problem with `libkrun`. When `crun` is running with `krun` mode, the entire microVM is contained within the crun process and the process makes all the fs accesses for the guest. The guest does not have access to the host kernel's rlimit values nor does it have ability to raise the soft value. Upon hitting the soft limit, `libkrun` would just relay the syscall error back to the guest. This effectively makes the soft limit a hard limit, and the 1024 default value are **very** limiting and almost no apps can work with it.

Note that this behavior can also be caused by specific OS distribution/ Systemd builds or configs. Running env info is attached at the bottom.

Manually setting `default_ulimits` in crio config can workaround this, but this is obviously an inconsistency with the documentation.

`conmon` showing the same set of `limits` as the `crio` daemon process.

Run within the attach described system any CRI-O load. And execute the commands as described in the first section of this report.

_No response_

<details>

```console
$ crio --version
crio version 1.36.0
   GitCommit:      unknown
   GitCommitDate:  unknown
   GitTreeState:   clean
   BuildDate:      2026-05-09T11:22:05Z
   GoVersion:      go1.26.3-X:nodwarf5
   Compiler:       gc
   Platform:       linux/amd64
   Linkmode:       dynamic
   BuildTags:
     containers_image_ostree_stub
     apparmor
     seccomp
   LDFlags:           -X github.com/cri-o/cri-o/internal/version.buildDate=2026-05-09T11:22:05Z -X github.com/cri-o/cri-o/internal/version.buildCommit= -compressdwarf=false -linkmode external
   SeccompEnabled:   true
   AppArmorEnabled:  false
```

```console
{
  "clientVersion": {
    "major": "1",
    "minor": "36",
    "gitVersion": "v1.36.1",
    "gitCommit": "756939600b9a7180fc2df6550a4585b638875e67",
    "gitTreeState": "archive",
    "buildDate": "1980-01-01T00:00:00Z",
    "goVersion": "go1.26.3",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "kustomizeVersion": "v5.8.1",
  "serverVersion": {
    "major": "1",
    "minor": "35",
    "emulationMajor": "1",
    "emulationMinor": "35",
    "minCompatibilityMajor": "1",
    "minCompatibilityMinor": "34",
    "gitVersion": "v1.35.3+k0s",
    "gitCommit": "6c1cd99aef09161ddb07b8ade6c9564e9b9bfe27",
    "gitTreeState": "clean",
    "buildDate": "2026-03-27T10:02:32Z",
    "goVersion": "go1.25.8",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
VERSION_ID=20260601.0.539401
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
$ uname -a
Linux hatsune 7.0.10-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 23 May 2026 14:21:20 +0000 x86_64 GNU/Linux
$ systemctl --version
systemd 260 (260.2-2-arch)
+PAM +AUDIT -SELINUX +APPARMOR -IMA +IPE +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 +KMOD +LIBCRYPTSETUP +LIBCRYPTSETUP_PLUGINS +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK +BTF +XKBCOMMON +UTMP +LIBARCHIVE
```

</details>

N.A.

pujitha24:
/kind documentation

What this PR does / why we need it:

Motivation:
The docs for `default_ulimits` state that, when unset, limits are
"inherited from the CRI-O daemon". This is only true when
`cgroup_manager` is `cgroupfs`. When `cgroup_manager` is `systemd`,
conmon is started by CRI-O (inheriting the daemon's rlimits via
fork/exec) but is then moved into a transient systemd scope
(MoveConmonToCgroup -> RunUnderSystemdScope). CRI-O does not set any
`Limit*` property on that transient unit, so systemd applies its own
defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see
systemd-system.conf(5)) to the scope, silently overriding the limit
conmon originally inherited from the daemon. This can surprise
operators who set a high `nofile` limit on the CRI-O daemon expecting
containers to inherit it.

Approach:
This is a documentation-only fix, as agreed on the issue by a
maintainer. A prior attempt to fix this by changing CRI-O's default
ulimit values in code (#10135) was rejected, since a default nofile
value cannot be chosen correctly for all workloads. Instead, this
updates the `default_ulimits` description in docs/crio.conf.5.md and
the matching comment in pkg/config/template.go (used to generate the
default crio.conf) to explain the systemd-scope behavior.

Validation:
Ran `go build ./pkg/config/...` (passes) and
`go run ./test/docs-validation/...` to confirm no new doc/template
tag mismatches are introduced by this change (pre-existing,
unrelated failures for `default_runtime`/`cgroup_manager` are present
identically on main, due to missing runtime binaries in this
environment, and are not affected by this change).

Which issue(s) this PR fixes:

Fixes #10006

Special notes for your reviewer:
None.

Does this PR introduce a user-facing change?

```release-note
NONE
```

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

## Summary by CodeRabbit

- **Documentation**
  - Clarified that containers inherit the CRI-O daemon’s resource limits when no default ulimits are configured, regardless of the selected cgroup manager.
  - Recommended explicitly setting `default_ulimits` when deterministic limits are required.
  - Updated the embedded configuration template comments to match the documented inheritance behavior.
  - Retained guidance on the expected default ulimits entry format, such as `nofile=1024:2048`.

### DIFF
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -123,7 +123,7 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 The _name_ of the OCI runtime to be used as the default. This option supports live configuration reload.
 
 **default_ulimits**=[]
-A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example: "nofile=1024:2048". If nothing is set here and `cgroup_manager` is `cgroupfs`, settings will be inherited from the CRI-O daemon. If `cgroup_manager` is `systemd`, conmon is moved into a transient systemd scope after being started, so any unset limits are instead inherited from systemd's own defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see systemd-system.conf(5)) rather than from the CRI-O daemon.
+A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example: "nofile=1024:2048". If nothing is set here and `cgroup_manager` is `cgroupfs`, settings will be inherited from the CRI-O daemon. If `cgroup_manager` is `systemd`, unset limits are instead inherited from systemd's own defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see systemd-system.conf(5)) rather than from the CRI-O daemon.
 
 **no_pivot**=false
 If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -123,7 +123,7 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 The _name_ of the OCI runtime to be used as the default. This option supports live configuration reload.
 
 **default_ulimits**=[]
-A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example:"nofile=1024:2048". If nothing is set here, settings will be inherited from the CRI-O daemon.
+A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example:"nofile=1024:2048". If nothing is set here and `cgroup_manager` is `cgroupfs`, settings will be inherited from the CRI-O daemon. If `cgroup_manager` is `systemd`, conmon is moved into a transient systemd scope after being started, so any unset limits are instead inherited from systemd's own defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see systemd-system.conf(5)) rather than from the CRI-O daemon.
 
 **no_pivot**=false
 If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -123,7 +123,7 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 The _name_ of the OCI runtime to be used as the default. This option supports live configuration reload.
 
 **default_ulimits**=[]
-A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example:"nofile=1024:2048". If nothing is set here and `cgroup_manager` is `cgroupfs`, settings will be inherited from the CRI-O daemon. If `cgroup_manager` is `systemd`, conmon is moved into a transient systemd scope after being started, so any unset limits are instead inherited from systemd's own defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see systemd-system.conf(5)) rather than from the CRI-O daemon.
+A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example: "nofile=1024:2048". If nothing is set here and `cgroup_manager` is `cgroupfs`, settings will be inherited from the CRI-O daemon. If `cgroup_manager` is `systemd`, conmon is moved into a transient systemd scope after being started, so any unset limits are instead inherited from systemd's own defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see systemd-system.conf(5)) rather than from the CRI-O daemon.
 
 **no_pivot**=false
 If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
--- a/docs/crio.conf.5.md
+++ b/docs/crio.conf.5.md
@@ -123,7 +123,7 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 The _name_ of the OCI runtime to be used as the default. This option supports live configuration reload.
 
 **default_ulimits**=[]
-A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example: "nofile=1024:2048". If nothing is set here and `cgroup_manager` is `cgroupfs`, settings will be inherited from the CRI-O daemon. If `cgroup_manager` is `systemd`, unset limits are instead inherited from systemd's own defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see systemd-system.conf(5)) rather than from the CRI-O daemon.
+A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example: "nofile=1024:2048". If nothing is set here, containers inherit whatever rlimits the CRI-O daemon itself was started with (for example, via its systemd unit's `LimitNOFILE=`), regardless of the configured `cgroup_manager`. Set `default_ulimits` explicitly if a deterministic limit is required.
 
 **no_pivot**=false
 If true, the runtime will not use `pivot_root`, but instead use `MS_MOVE`.
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -944,7 +944,10 @@ const templateStringCrioRuntime = `# The crio.runtime table contains settings pe
 const templateStringCrioRuntimeDefaultUlimits = `# A list of ulimits to be set in containers by default, specified as
 # "<ulimit name>=<soft limit>:<hard limit>", for example:
 # "nofile=1024:2048"
-# If nothing is set here, settings will be inherited from the CRI-O daemon
+# If nothing is set here and cgroup_manager is cgroupfs, settings will be
+# inherited from the CRI-O daemon. If cgroup_manager is systemd, unset limits
+# are instead inherited from systemd's own defaults (e.g.
+# DefaultLimitNOFILE=1024:524288), not from the CRI-O daemon
 {{ $.Comment }}default_ulimits = [
 {{ range $ulimit := .DefaultUlimits }}{{ $.Comment }}{{ printf "\t%q,\n" $ulimit }}{{ end }}{{ $.Comment }}]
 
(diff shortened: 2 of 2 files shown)


pujitha24:
Thanks for the catch, @Copilot! Fixed the missing space after the colon in the `default_ulimits` example (`for example:"nofile=1024:2048"` → `for example: "nofile=1024:2048"`) in `docs/crio.conf.5.md` to match the convention used elsewhere in the doc (e.g. the `additional-devices` example).

bitoku:
Thanks!
/ok-to-test

docs/crio.conf.5.md
```diff
@@ -123,7 +123,7 @@ The `crio.runtime` table contains settings pertaining to the OCI runtime used an
 The _name_ of the OCI runtime to be used as the default. This option supports live configuration reload.
 
 **default_ulimits**=[]
-A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example:"nofile=1024:2048". If nothing is set here, settings will be inherited from the CRI-O daemon.
+A list of ulimits to be set in containers by default, specified as "<ulimit name>=<soft limit>:<hard limit>", for example: "nofile=1024:2048". If nothing is set here and `cgroup_manager` is `cgroupfs`, settings will be inherited from the CRI-O daemon. If `cgroup_manager` is `systemd`, conmon is moved into a transient systemd scope after being started, so any unset limits are instead inherited from systemd's own defaults (e.g. `DefaultLimitNOFILE=1024:524288`, see systemd-system.conf(5)) rather than from the CRI-O daemon.
```
bitoku: > conmon is moved into a transient systemd scope after being started,

This is a bit too detailed. This can be commented on the code, but I want the doc to be simpler like the change in template.go.
Also they may use conmon-rs, and other runtime.

pujitha24:
@bitoku Good catch. Removed the "conmon is moved into a transient systemd scope after being started" detail from the `default_ulimits` doc in `docs/crio.conf.5.md` — that mechanism isn't conmon-specific anyway (it's a generic transient-unit start, so it applies regardless of whether conmon, conmon-rs, or another monitor is used). The line now reads simpler, matching the wording already in `pkg/config/template.go`: "unset limits are instead inherited from systemd's own defaults ... rather than from the CRI-O daemon."

The Copilot bot's missing-space nit was already fixed in a prior commit on this branch.

The currently-failing CI jobs (e2e, integration, rpm-build) don't look related to this docs-only change — I didn't touch them.

bitoku:
/approved

@cri-o/cri-o-maintainers PTAL

bitoku:
/approve

@cri-o/cri-o-maintainers PTAL

xw19:
Hi @pujitha24,

Thank you so much for the PR and for investigating this issue. This is a particularly tricky edge case to debug, and I really appreciate the effort you’ve put into it.

I’ve been looking into this, and I think the root cause may be slightly different. Let’s first clarify the distinction between `cgroupfs` and the `systemd` cgroup manager:

* With `cgroupfs`, the process that spawns the workload has direct control over the cgroup.
* With `systemd`, CRI-O asks systemd to create transient scopes for the relevant processes.

However, placing a process into a systemd transient scope does not change its file-descriptor limits. Scope placement organizes the process within the cgroup hierarchy, but it does not modify the process’s existing `RLIMIT_NOFILE` values.

The relevant behavior is in how CRI-O and Go handle these limits. When CRI-O starts, it inherits an initial soft limit `x` and hard limit `y`. Starting with Go 1.19, Go internally raises the soft limit to `y-1` so that the process can operate efficiently. As a result, CRI-O itself may appear to run with limits of `(y-1):y`.

However, when Go starts a child process, such as `conmon`, it restores and passes the original `(x:y)` limits to the child. If `conmon` is subsequently placed into a systemd scope, systemd does not modify those limits.

Therefore, the limits inherited by the container do not actually depend on the selected `cgroup_manager`; they depend on the limits with which the CRI-O daemon was initially started.

Because of this, documenting that systemd overrides the file limits may be misleading. Would you be open to updating the PR to strongly recommend configuring explicit container limits through `default_ulimits` when operators need deterministic values?

Thanks again for the PR and for taking the time to investigate this.

pujitha24:
You're right — the systemd-scope claim doesn't hold up, and the actual doc text now reflects that. I dropped the cgroupfs-vs-systemd distinction entirely; it just says containers inherit whatever rlimits the daemon itself was started with, regardless of `cgroup_manager`, and recommends setting `default_ulimits` explicitly when a deterministic value is needed. That's in the current commit (c919d2cb).

The failing `ci-fedora-integration` check isn't related to this change: the actual integration test step passed, the failure is in the unrelated post-test `gcp-crio-deprovision-vpc` cleanup step, which hit a DNS resolution failure reaching `oauth2.googleapis.com`.

PR #10154, opened 2026-07-20
test: tolerate memory metrics sampling races

### LINKED ISSUE #10055: Flaky test: container memory metrics uses exact comparison subject to race
The container memory metrics test in `test/cri-metrics.bats` is flaky. It asserts that `container_memory_working_set_bytes` from CRI-O's `/metrics` endpoint exactly equals `memory.current - inactive_file` read from the cgroup filesystem. Since these two sources are read at different times, the values can diverge and the test fails spuriously.

**Proposed fix:** Replace exact `==` comparisons with tolerance-based checks (e.g., within 16 KiB).

**CI failure:** https://github.com/cri-o/cri-o/actions/runs/27818313454/job/82325223527?pr=10049

clcc2019:
#### What type of PR is this?

/kind flake

The container memory metrics test compares values sampled from CRI-O and the
cgroup filesystem at different times. Requiring exact equality makes the test
fail when memory changes between those reads.

This change compares usage and working-set metrics with a 16 KiB tolerance and
prints the actual difference when an assertion fails. The tolerance absorbs
the small sampling race while still detecting meaningful metric regressions.

Fixes #10055

Validated with the repository-wide `make shellcheck` and `make lint` targets.

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * Improved container memory metrics validation by allowing small measurement differences.
  * Added clearer failure messages when reported values fall outside the accepted tolerance.

### DIFF
--- a/test/cri-metrics.bats
+++ b/test/cri-metrics.bats
@@ -29,6 +29,20 @@ function wait_for_metric() {
 	return 1
 }
 
+function assert_within_tolerance() {
+	local actual="$1"
+	local expected="$2"
+	local tolerance="$3"
+	local difference
+
+	difference=$((actual - expected))
+	((difference < 0)) && difference=$((-difference))
+	if ((difference > tolerance)); then
+		echo "Expected $actual to be within $tolerance of $expected (difference: $difference)" >&2
+		return 1
+	fi
+}
+
 function metrics_setup() {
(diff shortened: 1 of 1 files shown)


bitoku:
closing because it's already addressed.

PR #10167, opened 2026-07-20
Rewrite datastore to use podman libartifact directly

bitoku:
Replace the ociartifact.Store embedding in datastore.Store with a direct dependency on go.podman.io/common/pkg/libartifact.ArtifactStore. This removes the intermediary ociartifact wrapper layer, simplifying the code from 9 Impl interface methods down to 2, and replacing manual OCI layout blob reading (artifactData/readBlob/verifyDigest) with ArtifactStore.BlobMountPaths for file-based blob access.

Assisted-by: Claude Code <https://claude.com/claude-code>

```release-note
None
```

## Summary by CodeRabbit

* **Improvements**
  * OCI artifact retrieval now uses a streamlined artifact-store based workflow (pull → mount blobs → read mounted content).
  * Artifact data is returned per mounted blob, with maximum configured size enforced across all retrieved content.
  * Pull options are sanitized and forwarded consistently during artifact retrieval.

* **Bug Fixes**
  * Improved error propagation from reference creation, pulling, blob mounting, and file reading; oversized artifacts are rejected reliably.
  * Seccomp OCI artifact pull no longer enforces a required config media type during retrieval.


bitoku:
@cri-o/cri-o-maintainers PTAL

internal/ociartifact/datastore/store.go
```diff
@@ -76,199 +65,64 @@ type PullOptions struct {
...
-	dockerRef, err := s.getImageReference(ref)
+	artRef, err := s.impl.NewArtifactReference(ref)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get image reference: %w", err)
+		return nil, fmt.Errorf("create artifact reference: %w", err)
 	}
 
-	manifestDigest, err := s.Pull(ctx, dockerRef, opts.CopyOptions)
-	if err != nil {
+	if _, err := s.store.Pull(ctx, artRef, *opts.CopyOptions); err != nil {
```
saschagrunert, line 74: The old `PullData` went through `ociartifact.Store.Pull`, which called `EnsureNotContainerImage` before pulling (`internal/ociartifact/store.go:118`). That check inspects the manifest to reject regular container images from entering the artifact store.

The new code calls `libartifact.ArtifactStore.Pull` directly, which is a plain copy with no such check. If someone puts a container image reference in a seccomp annotation (e.g., `docker.io/library/nginx:latest`), the old code failed fast with a clear error. The new code would pull the full image into the artifact store (wasting bandwidth/disk), then fail later with a confusing error when the layers can't be parsed as seccomp profiles.

Is the intent to add this check elsewhere, or is it considered unnecessary for the datastore's use case?

saschagrunert (commented):
Minor pre-existing observation: `PullOptions.EnforceConfigMediaType` is set by `seccompociartifact.TryPull` but never consumed by `PullData` or anything downstream (neither before nor after this PR). Might be worth removing while you're cleaning up this area.

bitoku:
@saschagrunert PTAL?
The test failures seem unrelated.

/retest

saschagrunert:
> @saschagrunert PTAL? The test failures seem unrelated.

Have you checked out https://github.com/cri-o/cri-o/pull/10167#discussion_r3629815971 ?

internal/ociartifact/datastore/store.go
bitoku: Sorry, I thought I'd already replied!

I'm looking to use this for other purposes (like CoCo), so I'd prefer not to store EnsureNotContainerImage in the datastore.

I was initially going to add an `EnforceConfigMediaType` check instead, but doing so creates a race condition due to the time gap between checking and pulling. We could check it after pulling, but that would mean incurring an unnecessary pull.

Because of that, I decided not to check `EnforceConfigMediaType` here as well. Ultimately, this validation belongs in container-libs, not CRI-O.

internal/config/seccomp/seccompociartifact/seccompociartifact.go
```diff
@@ -76,7 +73,7 @@ func (s *SeccompOCIArtifact) TryPull(
 		return nil, nil
 	}
 
-	artifactData, err := s.impl.PullData(ctx, profileRef, &datastore.PullOptions{EnforceConfigMediaType: requiredConfigMediaType})
+	artifactData, err := s.impl.PullData(ctx, profileRef, nil)
```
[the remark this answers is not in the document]
bitoku: EnforceConfigMediaType was no-op. It doesn't change the behavior

bitoku:
@saschagrunert
Sorry I forgot to leave the comment. PTAL again?
https://github.com/cri-o/cri-o/pull/10167#discussion_r3639478909

saschagrunert (approved):

PR #10173, opened 2026-07-21
statsserver: add Kubernetes labels to CRI metrics

### LINKED ISSUE #140778: [cri-o] `/metrics/cadvisor` missing `namespace`, `pod`, `container` labels when using CRI stats
### What happened?

CRI runtimes provide labels like `id`, `name`, and `image`, but not the Kubernetes-level `namespace`, `pod`, and `container` labels. The cadvisor path adds these via `containerPrometheusLabelsFunc` in `pkg/kubelet/server/server.go`, but the CRI path has no equivalent.

This means any monitoring setup that queries by `namespace`, `pod`, or `container` on cadvisor metrics (which is most of them) breaks when the feature gate is turned on.

`/metrics/cadvisor` should produce the same label set regardless of whether the cadvisor or CRI stats path is active.

1. Enable `PodAndContainerStatsFromCRI` feature gate on the kubelet
2. Run a workload
3. Scrape `/metrics/cadvisor`
4. Observe that metrics like `container_cpu_usage_seconds_total` are missing `namespace`, `pod`, and `container` labels

Two approaches:

**Option 1: Kubelet resolves labels at scrape time**

Call `ListPodSandbox` and `ListContainers` in the metrics collector to build sandbox/container ID → metadata maps, then append `namespace`, `pod`, `container` to each metric. This follows the same pattern `criStatsProvider.getPodAndContainerMaps` uses for `/stats/summary`. The two RPCs are local gRPC calls and the scrape interval is typically 30s, so the cost is low.

**Option 2: Extend the CRI API**

`ListPodSandboxMetrics` currently returns metrics keyed by sandbox/container ID, but without any pod or container metadata. The runtime already has this metadata, it receives `PodSandboxMetadata` (name, namespace, uid) during `RunPodSandbox` and `ContainerMetadata` (name) during `CreateContainer`, and stores both for the lifetime of the sandbox/container.

Adding these existing metadata fields to `PodSandboxMetrics` and `ContainerMetrics` would let the kubelet read `namespace`, `pod`, and `container` directly from the metrics response without any additional RPCs or cache lookups.

A lighter variant: only add the pod UID to `PodSandboxMetrics`. The kubelet can then look up name and namespace from its internal pod cache (keyed by UID), and resolve container names by matching container IDs against the cached pod status.

/sig node
/kind bug

dgrisonnet:
#### What type of PR is this?

/kind bug

Include namespace, pod, and container labels in all CRI metrics to match the labels produced by the cadvisor metrics path. The descriptors declare these as base label keys, and the values are populated from sandbox and container metadata.

Fixes https://github.com/kubernetes/kubernetes/issues/140778

The Kubernetes labels are filled in after metric generation to avoid passing the sandbox through every generate function.

```release-note
Added `namespace`, `pod`, and `container` labels to CRI metrics
```

## Summary by CodeRabbit

* **New Features**
  * Metrics now include additional runtime and Kubernetes context, including namespace, pod name, and container name.
  * Expanded metric label sets provide more granular visibility into sandbox and container activity.

* **Bug Fixes**
  * Updated generated network and container metrics so label values consistently match the expanded label definitions.

* **Tests**
  * Adjusted network metrics test expectations to reflect the expanded label values.

### DIFF
--- a/internal/lib/statsserver/metrics.go
+++ b/internal/lib/statsserver/metrics.go
@@ -10,7 +10,11 @@ import (
 	"github.com/cri-o/cri-o/pkg/config"
 )
 
-var baseLabelKeys = []string{"id", "name", "image"}
+var (
+	runtimeLabelKeys    = []string{"id", "name", "image"}
+	kubernetesLabelKeys = []string{"namespace", "pod", "container"}
+	baseLabelKeys       = append(append([]string{}, runtimeLabelKeys...), kubernetesLabelKeys...)
+)
 
 type metricValue struct {
 	value      uint64
@@ -145,8 +149,13 @@ func computeSandboxMetrics(sb *sandbox.Sandbox, metrics []*containerMetric) []*t
 }
 
 func sandboxBaseLabelValues(sb *sandbox.Sandbox) []string {
+	namespace, podName := "", ""
+	if md := sb.Metadata(); md != nil {
+		namespace = md.GetNamespace()
+		podName = md.GetName()
+	}
 	// TODO FIXME: image?
-	return []string{sb.ID(), "POD", ""}
+	return []string{sb.ID(), "POD", "", namespace, podName, "POD"}
 }
 
 // computeContainerMetrics computes the metrics for container.
@@ -160,7 +169,7 @@ func containerBaseLabelValues(ctr *oci.Container) []string {
 		image = someNameOfTheImage.StringForOutOfProcessConsumptionOnly()
 	}
 
-	return []string{ctr.ID(), ctr.Name(), image}
+	return []string{ctr.ID(), ctr.Name(), image, "", "", ""}
 }
 
 func computeMetrics(baseLabels []string, metrics []*containerMetric) []*types.Metric {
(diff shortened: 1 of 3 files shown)


internal/lib/statsserver/metrics.go
```diff
@@ -10,7 +10,11 @@ import (
 	"github.com/cri-o/cri-o/pkg/config"
 )
 
-var baseLabelKeys = []string{"id", "name", "image"}
+var (
+	runtimeLabelKeys    = []string{"id", "name", "image"}
+	kubernetesLabelKeys = []string{"namespace", "pod", "container"}
```
dgrisonnet, line 15: containerd already exposed the kubernetes labels, but it doesn't have `name`: https://github.com/containerd/containerd/blob/main/internal/cri/server/list_metric_descriptors.go#L35. To me we should also remove it in cri-o since for pod-level metrics, we set it to `POD` and for container-level metrics we set it to the container name like the `container` label so it is redundant.

haircommander:
/ok-to-test
/approve

/cherry-pick release-1.36

internal/lib/statsserver/metrics.go
bitoku: I think so too, but we should do in a separate PR, since we're going to backport this PR to release-1.36.

bitoku (commented):
/lgtm

internal/lib/statsserver/metrics.go
dgrisonnet: I agree

dgrisonnet:
unrelated failures

/retest-required

PR #10081, opened 2026-06-30
Retry in copyimg when using remote registries

ngopalak-redhat:
#### What type of PR is this?

/kind ci

#### What this PR does / why we need it:
This is a smaller attempt to reduce the failures in integration tests. Using a local registry is much larger change and required careful review of multiple test cases. Hence taking an smaller step by adding retry to the copyimg binary for now.

It will address one of the common errors:
```
time="2026-06-29T19:51:56Z" level=fatal msg="Error copying docker://quay.io/crio/hello-wasm:latest to dir:/home/runner/work/cri-o/cri-o/.artifacts/hello_wasm-image: reading blob sha256:eb7836fc120a3ace0ea10c2752ad9c3894643f7146ec7d05f8c56d8794c9f1c3: fetching blob: received unexpected HTTP status: 503 Service Unavailable"
```

#### Which issue(s) this PR fixes:
Partially fixes/ https://github.com/cri-o/cri-o/issues/9859

#### Special notes for your reviewer:
@bitoku The local registry approach fails for some test cases. Hence this smaller change
#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a retry option for image copy operations, helping remote image pulls recover from temporary failures.

* **Bug Fixes**
  * Improved the image copy flow so remote source handling is more reliable.
  * Corrected command argument handling in the shared test script to ensure options are passed properly during copy operations.

PR #9995, opened 2026-06-02
ci: bump kata version to 3.31.0 and update tarball extraction

thejasn:
#### What type of PR is this?
/kind ci

#### What this PR does / why we need it:

Bumps the kata-containers version used in CI from 3.6.0 to 3.31.0 to
verify whether existing integration tests catch the pod termination
regression introduced in kata 3.25.0 (commit 956f43c6c — "runtime:
skip MoveTo for systemd cgroups").

Also updates `build/kata.yml` to handle the `.tar.zst` tarball format
used by newer kata-containers releases.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

## Summary by CodeRabbit

* **Chores**
  * Updated CI runtime/tooling versions: Kata to 3.31.0 (with refreshed checksum), CRI-Tools to v1.36.0, runc to v1.5.0, and crun to 1.28, plus increased integration test timeout.
  * Improved Kata installation/build handling by switching to zstd-compressed archive downloading, checksum verification, and direct extraction (with cleanup of temporary archives).

PR #10184, opened 2026-07-25
Retry network fetches in the CI setup script

xw19:
#### What type of PR is this?

/kind ci

Every job in the integration matrix runs `scripts/github-actions-setup` before
any test does, so one transient network failure there costs a whole job before
a single test has run. Surveying recent pull requests, two jobs died exactly
this way — at 47s and 2m40s, short enough to be mistaken for a build break:

- [#9875](https://github.com/cri-o/cri-o/actions/runs/28502733790/job/84492545038): `curl: (22) The requested URL returned error: 403`
- [#9995](https://github.com/cri-o/cri-o/actions/runs/26818600353/job/79083434998): toolchain failure building runc from source

Two holes:

1. **`--retry` does not cover a 403.** `curl_retry` already passes `--retry 5`,
   but that only retries transient errors — timeouts, 429, 5xx. The #9875
   failure happened *inside* a `curl_retry` call which then declined to retry
   it. `--retry-all-errors` closes that.
2. **The three `git clone`s had no retry at all** —
   `install_from_github_org_repo`, crun and buildah.

Wrapping the clones needs a little care: a failed clone can leave a partial
directory behind, and git refuses to clone into a non-empty one, so a naive
retry loop fails every subsequent attempt with `already exists` instead of
retrying the fetch. `git_clone_retry` clears the target before each attempt.

This is the same reasoning as
[#10081](https://github.com/cri-o/cri-o/pull/10081) (retry remote fetches used
by the tests), applied one layer further out to the setup that every job
depends on.

None.

The cni-plugins download is switched from `wget` to `curl_retry` so that every
download in the script goes through one helper. Note this one is *consistency,
not a fix* — GNU wget already defaults to `--tries=20`, so it retried transient
errors; like curl it simply does not retry a fatal 4xx. Happy to drop that hunk
if you would rather keep the diff to the two real holes.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation reliability by adding stronger retry behavior for network downloads and repository checkouts during transient failures.
  * Prevented incomplete or pre-existing target directories from blocking retry attempts.
  * Enhanced setup of CNI plugins, crun, and Buildah with more resilient download/clone operations.

### DIFF
--- a/scripts/github-actions-setup
+++ b/scripts/github-actions-setup
@@ -24,7 +24,31 @@ main() {
 }
 
 curl_retry() {
-    sudo curl -sSfL --retry 5 --retry-delay 3 "$@"
+    # --retry alone only covers transient errors, which does not include the
+    # 403 a release download occasionally returns.
+    sudo curl -sSfL --retry 5 --retry-delay 3 --retry-all-errors "$@"
+}
+
+# Clone with retries. Every job in the matrix runs this script before any test
+# does, so one transient failure here costs a whole job.
+#
+# A failed clone can leave a partial directory behind and git refuses to clone
+# into a non-empty one, so clear the target before each attempt - otherwise
+# every retry fails with "already exists" instead of retrying the fetch.
+git_clone_retry() {
+    local url=$1 dir=${2:-}
+    [[ -n "$dir" ]] || dir=$(basename "$url" .git)
+
+    local i
+    for ((i = 1; i <= 5; i++)); do
+        sudo rm -rf "$dir"
+        if git clone "$url" "$dir"; then
+            return 0
+        fi
+        echo "Clone attempt $i of 5 failed: $url" >&2
+        sleep 3
+    done
+    return 1
 }
 
 prepare_system() {
@@ -57,7 +81,7 @@ install_from_github_org_repo() {
     URL=https://github.com/$1/$2
     DIR=$(mktemp -ud)
 
-    git clone "$URL" "$DIR"
+    git_clone_retry "$URL" "$DIR"
     pushd "$DIR"
     git checkout "$3"
     "${@:4}"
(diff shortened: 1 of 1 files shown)


xw19:
Adding the raw evidence for the two failures referenced above, since setup
failures are easy to misread as build breaks.

**[#9875 — `integration / userns / crun / amd64`](https://github.com/cri-o/cri-o/actions/runs/28502733790/job/84492545038)**, failed after 47s at step 9 *Run scripts/github-actions-setup*:

```
Setting up libgpg-error-dev (1.47-3build2.1) ...
curl: (22) The requested URL returned error: 403
##[error]Process completed with exit code 22.
```

This is the one that motivates `--retry-all-errors`: it happened inside a
`curl_retry` call, which already passes `--retry 5` and declined to retry
because a 403 is not a transient error to curl.

**[#9995 — `integration / conmon-rs / crun / amd64`](https://github.com/cri-o/cri-o/actions/runs/26818600353/job/79083434998)**, failed after 2m40s at the same step:

```
[pkg-config] generating pathrs pkg-config
##[error]/opt/hostedtoolcache/go/1.26.3/x64/src/debug/elf/file.go:21:2:
  package debug/dwarf is not in std (...)
make: *** [Makefile:85: runc-bin] Error 1
```

That second one is a toolchain mismatch rather than a network error, so this
patch does **not** fix it. I include it because it makes the same structural
point: the setup step builds several tools from source and any of them can take
a whole job down before testing starts. Worth considering separately whether
more of these should come from release binaries, as runc and conmon already do.

Both jobs are `conclusion=failure` at step 9 with zero tests executed.

bitoku:
I like the idea to build and cache it once, and use the binaries from the cache, not always build it.

PR #10192, opened 2026-07-27
Skip normalizing bare tag artifact names from OCI layouts

### LINKED ISSUE #10151: Using skopeo pre-pulled images with `additional_artifact_stores`
Hi, 

I was just testing the new `additional_artifact_stores` feature, expecting I could pre-pull images with `skopeo` and then have `crio` use them.

I can "somehow" get it to work; but for some reason, `crio` wrongly recognized the image name and tag. For instance, I pulled `docker://docker.io/coredns/coredns:1.14.4` and moved the contents to the given directory. However, when I list my images using `crictl`, the image has the tag as name set?

```sh
# check images with crictl
$ sudo crictl images
IMAGE                      TAG                 IMAGE ID            SIZE
docker.io/library/1.14.4   <none>              59ac1d5ecb6f4       23.7MB
```

I would expect the image name to be `docker.io/library/coredns` and the tag to be `1.14.4`.

I think (?), `crio` is using the `index.json` file to determine the image name and tag. But for some reason, it seems to be using the tag as the image name. But! I am unsure if this is a `skopeo` or `crio` issue.

```sh
$ cat /tmp/coredns/artifacts/index.json | yq -oj
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:59ac1d5ecb6f40224d5f1df16771f5c12418ee69d4a2e7fc76319d236c0fdd4f",
      "size": 2265,
      "annotations": {
        "org.opencontainers.image.ref.name": "1.14.4"
      }
    }
  ]
}
```

**Reproduction:**

```sh
# pull the image as oci archive to some location
skopeo copy docker://docker.io/coredns/coredns:1.14.4 oci:/tmp/coredns:1.14.4

# needed to make crio find the image
mkdir -p /tmp/coredns/artifacts/
mv /tmp/coredns/blobs/ /tmp/coredns/artifacts
mv /tmp/coredns/index.json /tmp/coredns/artifacts
mv /tmp/coredns/oci-layout /tmp/coredns/artifacts

# start crio in the background with the additional artifact stores
sudo crio --additional-artifact-stores /tmp/coredns/ &
```

Afterwards, inspect the images with `crictl`:

```sh
# check images with crictl
$ sudo crictl images
IMAGE                      TAG                 IMAGE ID            SIZE
docker.io/library/1.14.4   <none>              59ac1d5ecb6f4       23.7MB
```

I am using a crio build from the main branch

```sh
$  sudo crio --version
crio version 1.37.0
   GitCommit:      7e79671255660e45e39c19a9421d65fc8bcbf738
   GitCommitDate:  2026-07-16T20:18:30Z
   GitTreeState:   clean
   BuildDate:      2026-07-16T20:18:30Z
   GoVersion:      go1.26.4
   Compiler:       gc
   Platform:       linux/amd64
   Linkmode:       dynamic
   BuildTags:
     containers_image_openpgp
     btrfs_noversion
   LDFlags:          unknown
   SeccompEnabled:   false
   AppArmorEnabled:  false
```

Besides the issue; I have two more questions:

1) Forcing the `artifacts/` subpath is a bit weird. What's the reason behind this? Using skopeo, I have to manually move the `blobs/`, `index.json` and `oci-layout` files into the `artifacts/` subpath. Is there a way to avoid this?
   
2) What if I have multiple pre-pulled images? Currently it seems, I have to add an extra `--additional-artifact-stores` entry for each image. Is there a way to recursively add all subdirectories of a given path as additional artifact stores?

saschagrunert:
/kind bug

External tools like skopeo write only the tag portion (e.g. `1.14.4`) into the `org.opencontainers.image.ref.name` annotation in `index.json` when creating OCI layouts. `ParseNormalizedNamed` turns this bare tag into the nonsensical `docker.io/library/1.14.4`, producing a wrong image name and no tag in `crictl images` output. This PR detects names without a `/` as bare tags and leaves the reference as unknown instead of producing a wrong one.

Fixes #10151

The upstream fix is podman-container-tools/container-libs#1027. This CRI-O-side guard prevents the wrong name from surfacing regardless.

```release-note
Fix artifact names from externally created OCI layouts (e.g. via skopeo) being incorrectly normalized to wrong docker references.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved artifact name handling for bare tags, which are now treated as unknown references instead of being incorrectly normalized.
  * Preserved fully qualified artifact references, including registry, repository, and tag information.

* **Tests**
  * Added coverage for displaying bare tags as unknown and retaining fully qualified references in listings and inspection output.

### DIFF
--- a/internal/ociartifact/artifact.go
+++ b/internal/ociartifact/artifact.go
@@ -3,6 +3,7 @@ package ociartifact
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/opencontainers/go-digest"
 	"go.podman.io/common/pkg/libartifact"
@@ -35,7 +36,7 @@ type Artifact struct {
 }
 
 // newArtifact creates a new Artifact from a libartifact.Artifact.
-func (s *Store) newArtifact(art *libartifact.Artifact, rootPath string, pinned bool) *Artifact {
+func (s *Store) newArtifact(ctx context.Context, art *libartifact.Artifact, rootPath string, pinned bool) *Artifact {
 	artifact := &Artifact{
 		Artifact: art,
 		rootPath: rootPath,
@@ -43,14 +44,24 @@ func (s *Store) newArtifact(art *libartifact.Artifact, rootPath string, pinned b
 	}
 
 	if art.Name != "" {
-		namedRef, err := reference.ParseNormalizedNamed(art.Name)
-		if err != nil {
-			log.Warnf(context.Background(), "Failed to parse artifact name %s with the error %s", art.Name, err)
+		// Guard against bare tags from OCI layout annotations written by
+		// external tools (e.g. skopeo writes "1.14.4" instead of the full
+		// "docker.io/coredns/coredns:1.14.4"). ParseNormalizedNamed would
+		// turn these into nonsensical refs like "docker.io/library/1.14.4".
+		// The upstream fix lives in libartifact (container-libs #1027).
+		if !strings.Contains(art.Name, "/") {
+			log.Warnf(ctx,
+				"Artifact name %q looks like a bare tag, not a fully qualified reference; skipping normalization", art.Name)
+		} else {
+			namedRef, err := reference.ParseNormalizedNamed(art.Name)
+			if err != nil {
+				log.Warnf(ctx, "Failed to parse artifact name %s: %v", art.Name, err)
 
-			namedRef = unknownRef{}
+				namedRef = unknownRef{}
+			}
+
+			artifact.namedRef = namedRef
 		}
-
-		artifact.namedRef = namedRef
 	}
 
 	artifact.pinned = pinned || s.isArtifactPinned(artifact)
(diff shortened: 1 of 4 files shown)


saschagrunert:
/retest

@cri-o/cri-o-maintainers PTAL

bitoku (commented):
I'd like e2e tests for this change, and want to see how it's handled.

and I'm wondering if we can enforce the artifact name to be fully qualified. not sure which is better yet though.

saschagrunert:
> I'd like e2e tests for this change, and want to see how it's handled.

Done, added 2 integration tests.

> and I'm wondering if we can enforce the artifact name to be fully qualified. not sure which is better yet though.

Enforcing fully qualified names would mean rejecting bare tags at the OCI layout level. The challenge is that CRI-O doesn't control what external tools write into index.json. If we reject bare tags, artifacts from skopeo-created OCI layouts would become invisible rather than showing as "unknown". The current approach (degrade gracefully to "unknown") seems safer since it still lets users see and use the artifact by digest.

PR #10204, opened 2026-07-29
stats: fix network metrics using host netns instead of pod's

dgrisonnet:
#### What type of PR is this?

/kind bug

Network metrics were previously collected using `netlink.LinkList()`, which
operates on whatever network namespace the calling thread is in. For
non-hostNetwork pods, this returned the host's interfaces instead of the pod's.

This replaces the netlink approach with reading `/proc/<pid>/net/dev` via the
`prometheus/procfs` library. Linux's procfs is namespace-aware: reading from the
infra container's PID gives the pod's network stats without needing to enter the
network namespace. This is the same approach cadvisor uses for container network
metrics collection.

```release-note
Fix network metrics collection to use the pod's network namespace.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Network metrics now reflect the exact set of interfaces available in each pod’s network namespace, including isolated (non-host) pods.
  * Host-network pods now report metrics based on host interfaces.
  * When interface/link data can’t be retrieved (or no links are available), network metrics stay unpopulated instead of being partially filled.
* **Tests**
  * Added a Bats test validating metric interface label values against `/sys/class/net` for isolated pods, and confirming expanded interface coverage for host-network pods.

### DIFF
--- a/internal/lib/statsserver/network_metrics_linux.go
+++ b/internal/lib/statsserver/network_metrics_linux.go
@@ -1,7 +1,9 @@
 package statsserver
 
 import (
-	"github.com/vishvananda/netlink"
+	"fmt"
+
+	"github.com/prometheus/procfs"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 
 	"github.com/cri-o/cri-o/internal/lib/sandbox"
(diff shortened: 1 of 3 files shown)


kannon92 (commented):
/ok-to-test

internal/lib/statsserver/network_metrics_linux.go
```diff
@@ -11,24 +12,41 @@ import (
...
-	for i := range links {
-		if attrs := links[i].Attrs(); attrs != nil {
-			networkMetrics := generateSandboxNetworkMetrics(sb, attrs)
-			metrics = append(metrics, networkMetrics...)
-		}
+	var err error
+	if sb.HostNetwork() {
+		err = gather()
+	} else {
+		err = ns.WithNetNSPath(sb.NetNsPath(), func(_ ns.NetNS) error {
```
haircommander: we should not use withnetnspath it locks the os thread and can get messy in multithreaded apps. claude helped me find this

```
you can use netlink.NewHandleAt() with a netns file descriptor instead of entering the namespace. The approach:

  1. Open the netns path as a file descriptor via netns.GetFromPath(sb.NetNsPath())
  2. Create a netlink handle bound to that namespace via netlink.NewHandleAt(nsHandle)
  3. Call handle.LinkList() on the handle
  4. Close both the handle and the ns fd

  This sends netlink messages to the target namespace's socket directly — no thread-locking, no runtime.LockOSThread, no goroutine scheduling hazards. Something like:

  nsHandle, err := netns.GetFromPath(sb.NetNsPath())
  if err != nil {
      return nil
  }
  defer nsHandle.Close()

  nlHandle, err := netlink.NewHandleAt(nsHandle)
  if err != nil {
      return nil
  }
  defer nlHandle.Close()

  links, err := nlHandle.LinkList()

```

haircommander: hm that still does lock os thread briefly... digging deeper

haircommander:
```
❯ is there anyway to avoid lockosthread entirely                                                                                                                                                                    

  Thought for 13s, searched for 1 pattern, read 1 file (ctrl+o to expand)

● Yes. All you actually need is interface names and rx/tx counters. That's exactly what /proc/<pid>/net/dev provides — and reading a file requires no namespace switching at all.

  The kernel exposes per-namespace network stats at /proc/<pid>/net/dev for any process in that namespace. You just need a PID from the sandbox's infra container. The format is:

  Inter-|   Receive                                                |  Transmit
   face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo frame compressed multicast
      lo:  1000      10    0    0    0     0          0         0     1000      10    0    0    0     0       0          0
    eth0: 50000     500    0    0    0     0          0         0    30000     300    0    0    0     0       0          0

  Every field the code uses — RxBytes, RxPackets, RxDropped, RxErrors, TxBytes, TxPackets, TxDropped, TxErrors, and interface name — maps directly to columns in that file. Pure file I/O, no netlink, no
  LockOSThread, no setns.

  Searched for 2 patterns (ctrl+o to expand)

● The sandbox has InfraContainer() and the container state has InitPid. So the PID is available via sb.InfraContainer().State().InitPid.

  You'd parse /proc/<pid>/net/dev — skip the 2 header lines, then each line gives you the interface name and all the counters in fixed column order:

  iface: RxBytes RxPackets RxErrors RxDropped _ _ _ _ TxBytes TxPackets TxErrors TxDropped ...

  This drops the netlink dependency from this code path entirely. No goroutine pinning, no namespace switching, just os.Open and a scanner. It's also what tools like cadvisor use for the same reason.


```

internal/lib/statsserver/network_metrics_linux.go
```diff
@@ -1,114 +1,138 @@
 package statsserver
 
 import (
-	"github.com/vishvananda/netlink"
+	"fmt"
+
+	"github.com/prometheus/procfs"
```
dgrisonnet, line 6: By using prometheus/procfs, we don't have to maintain the procfs net dev parser ourselves like cadvisor does: https://github.com/google/cadvisor/blob/6a0c4f2539a8b1c11215804dee404c06fe3e0218/lib/container/libcontainer/handler.go#L381-L466

The main difference with cadvisor though is that procfs parses sligthly more data than what we need for the metrics: https://github.com/prometheus/procfs/blob/f57ba67d28cea29fe197eb007aa6ebf656c12250/net_dev.go#L26-L44 as opposed to https://github.com/google/cadvisor/blob/master/lib/model/container.go#L516-L535.

internal/lib/statsserver/network_metrics_linux.go
dgrisonnet: Good catch, I hadn't thought about kernel lock.

I switched to parsing the procfs like cadvisor does

haircommander:
/approve

@cri-o/cri-o-maintainers PTAL

PR #10207, opened 2026-07-30
 stats: filter pod-owned interfaces from hostNetwork pod metrics

dgrisonnet:
#### What type of PR is this?

/kind bug

#### Summary

CRI network metrics (`container_network_*_total`) include an `interface` label. On a node with N pods, a hostNetwork pod sees every host interface including all N veth pairs created by the CNI plugin for other pods. Each veth has a unique name (e.g. `veth12ab34cd`), so the metric cardinality for a single hostNetwork pod scales as `N × 8 network metrics`. With multiple hostNetwork pods (node-exporter, kube-proxy, SDN pods, etc.) this becomes `hostNetwork_pods × N × 8`, a cardinality explosion that puts pressure on Prometheus TSDB, especially on large nodes.

This is one of two issues with CRI network metrics. The other is that `GenerateNetworkMetrics` currently reads links from the host namespace for all pods, not just hostNetwork ones. Non-hostNetwork pods should read from their own netns, and that fix is tracked separately in https://github.com/cri-o/cri-o/pull/10204.

This PR addresses the hostNetwork side by filtering pod-owned interfaces out of hostNetwork pod metrics. It tracks which host-side interfaces the CNI plugin creates and excludes them from the metrics snapshot.

#### Design

```
┌─────────────────────────────────────────────────────────┐
│                    Pod creation                         │
│                                                         │
│  RunPodSandbox ──► CNI plugin returns Result            │
│                     │                                   │
│                     ▼                                   │
│              HostInterfacesFromCNIResult()               │
│              Extracts iface names where                  │
│              Sandbox == "" (host-side)                   │
│                     │                                   │
│                     ▼                                   │
│              sandbox.SetHostInterfaces(["vethXXX"])      │
│                                                         │
│  LoadSandbox ──► Restores from CNI annotation            │
│                  (survives restarts)                     │
└─────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────┐
│               Metrics collection cycle                  │
│                                                         │
│  snapshotHostNetworkLinks(sandboxes)                     │
│  ┌────────────────────────────────────────────────────┐  │
│  │ 1. Build exclusion set from all non-hostNetwork    │  │
│  │    sandboxes' HostInterfaces                       │  │
│  │                                                    │  │
│  │    podIfaces = {"veth1", "veth2", "cni0", ...}     │  │
│  │                                                    │  │
│  │ 2. netlink.LinkList() -- single snapshot           │  │
│  │                                                    │  │
│  │ 3. Filter: keep links NOT in podIfaces             │  │
│  │                                                    │  │
│  │ 4. Store as ss.hostNetworkLinks                    │  │
│  └────────────────────────────────────────────────────┘  │
│                         │                               │
│                         ▼                               │
│  GenerateNetworkMetrics(sb)                              │
│    ├─ hostNetwork pod  → use ss.hostNetworkLinks         │
│    └─ regular pod      → netlink.LinkList() directly     │
└─────────────────────────────────────────────────────────┘
```

The snapshot is taken once per collection cycle (or once per on-demand request when `collection_period = 0`), so all hostNetwork sandboxes see a consistent set of interfaces without redundant netlink calls or races from concurrent pod creation.

#### Backward compatibility

Fully backward compatible. If the CNI result annotation is missing (older sandbox) or the CNI plugin doesn't report host-side interfaces, nothing enters the exclusion set and the hostNetwork sandbox reports all interfaces, same as before this change.

#### Commits

fb4d412b2620b683ce3cb6b613bb23778f890a9f **Add host interfaces field to sandbox**: `hostInterfaces []string` field with getter/setter on the `Sandbox` struct.
80cab997d57c23017a0832c90184aceb1885a81a **Populate host interfaces from CNI result**: extract host-side interface names (where `iface.Sandbox == ""`) from the CNI result at pod creation time (Linux and FreeBSD paths).
cdf6aeaae9de0c3bdbef66713577dcf1182aab49 **Restore host interfaces from CNI annotation**: on `LoadSandbox`, re-parse the persisted CNI result annotation to restore `hostInterfaces`, so filtering survives CRI-O restarts.
14dac55cb1e1ed851058b2c074ad0e79457609dd **Filter pod-owned interfaces from host network metrics**: pre-compute a filtered snapshot of host network links and use it for hostNetwork pod metrics. Includes unit tests for the filtering logic.
dd8d72917d9edc1187fc3475c5c43f89d5520eca **E2e test**: verify that a hostNetwork pod's CRI metrics do not include the veth peer of a non-hostNetwork pod.

```release-note
Filter pod-owned interfaces (veth pairs) from hostNetwork pod CRI network metrics to prevent cardinality explosion in container_network_*.
```

## New Features
- Automatically identifies host-side network interfaces during container network setup.
- Improves network metrics by using interface data visible to each sandbox and cached host-network information.

## Bug Fixes
- Excludes pod-owned interfaces from host-network metrics.
- Handles unavailable or empty network-interface data more reliably.

## Tests
- Added coverage for filtering host and pod interfaces.
- Added integration coverage confirming accurate sandbox interfaces and host-network metric exclusions.

### DIFF
--- a/internal/lib/sandbox/sandbox.go
+++ b/internal/lib/sandbox/sandbox.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	current "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/proto"
 	"k8s.io/apimachinery/pkg/fields"
@@ -519,4 +520,25 @@ func (s *Sandbox) Ready() bool {
 	defer s.stateMutex.RUnlock()
 
 	return s.created && !s.stopped
+}
+
+// HostInterfacesFromCNIResult returns the host-side interface names from a CNI result.
+// Per the CNI spec, an interface's Sandbox field holds the path to the network
+// namespace that contains it. An empty Sandbox means the interface is on the
+// host side (e.g. the host-side veth peer), while a non-empty value identifies
+// a container-side interface inside the pod's network namespace.
+func HostInterfacesFromCNIResult(result *current.Result) []string {
+	var ifaces []string
+
+	for _, iface := range result.Interfaces {
+		if iface == nil {
+			continue
+		}
+
+		if iface.Sandbox == "" {
+			ifaces = append(ifaces, iface.Name)
+		}
+	}
+
+	return ifaces
 }
--- a/internal/lib/sandbox/sandbox.go
+++ b/internal/lib/sandbox/sandbox.go
@@ -55,6 +55,7 @@ type Sandbox struct {
 	hostname       string
 	// ipv4 or ipv6 cache
 	ips                []string
+	hostInterfaces     []string
 	seccompProfilePath string
 	infraContainer     *oci.Container
 	nsOpts             *types.NamespaceOption
@@ -147,6 +148,16 @@ func (s *Sandbox) IPs() []string {
 // Wrap the IPs() method to match the NRI interface.
 func (s *Sandbox) GetIPs() []string {
 	return s.IPs()
+}
+
+// SetHostInterfaces stores the host network interfaces in the sandbox.
+func (s *Sandbox) SetHostInterfaces(ifaces []string) {
+	s.hostInterfaces = ifaces
+}
+
+// HostInterfaces returns the host network interfaces of the sandbox.
+func (s *Sandbox) HostInterfaces() []string {
+	return s.hostInterfaces
 }
 
 // ID returns the id of the sandbox.
--- a/internal/lib/statsserver/network_metrics_linux.go
+++ b/internal/lib/statsserver/network_metrics_linux.go
@@ -11,9 +11,34 @@ import (
 )
 
 func (ss *StatsServer) GenerateNetworkMetrics(sb *sandbox.Sandbox) []*types.Metric {
-	netDev, err := ss.readNetDev(sb)
+	if sb.HostNetwork() {
+		return ss.generateHostNetworkMetrics(sb)
+	}
+
+	return ss.generatePodNetworkMetrics(sb)
+}
+
+// generateHostNetworkMetrics returns metrics from the pre-computed host
+// network dev snapshot, which already has pod-owned interfaces filtered out.
+func (ss *StatsServer) generateHostNetworkMetrics(sb *sandbox.Sandbox) []*types.Metric {
+	var metrics []*types.Metric
+
+	for name := range ss.hostNetDev {
+		iface := ss.hostNetDev[name]
+		networkMetrics := generateSandboxNetworkMetrics(sb, &iface)
+		metrics = append(metrics, networkMetrics...)
+	}
+
+	return metrics
+}
+
+func (ss *StatsServer) generatePodNetworkMetrics(sb *sandbox.Sandbox) []*types.Metric {
+	netDev, err := readPodNetDev(sb)
 	if err != nil {
-		log.Errorf(ss.ctx, "Unable to retrieve network stats %s: %v", sb.ID(), err)
+		log.WithFields(ss.ctx, map[string]any{
+			"sandboxID": sb.ID(),
+			"error":     err,
+		}).Error("Unable to retrieve network stats")
 
 		return nil
 	}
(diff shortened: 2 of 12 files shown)


internal/lib/sandbox/sandbox.go
```diff
@@ -149,6 +151,16 @@ func (s *Sandbox) GetIPs() []string {
...
 
+// SetHostInterfaces stores the host network interfaces in the sandbox.
+func (s *Sandbox) SetHostInterfaces(ifaces []string) {
+	s.hostInterfaces = ifaces
+}
+
+// HostInterfaces returns the host network interfaces of the sandbox.
+func (s *Sandbox) HostInterfaces() []string {
+	return s.hostInterfaces
+}
```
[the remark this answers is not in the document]
dgrisonnet: The sandbox setter/getter pattern here is consistent with every other field on the Sandbox struct, none of which use mutex synchronization.

internal/lib/statsserver/network_metrics_linux.go
```diff
@@ -9,11 +9,38 @@ import (
...
+		networkMetrics := generateSandboxNetworkMetrics(sb, attrs)
+		metrics = append(metrics, networkMetrics...)
+	}
+
+	return metrics
+}
+
+// generatePodNetworkMetrics returns metrics by reading the links from the
+// host namespace. Each non-hostNetwork pod will get its own netns handling
+// once the netns fix lands.
```
haircommander: I don't think we should have a comment about this

internal/lib/statsserver/stats_server_linux.go
```diff
@@ -483,3 +488,44 @@ func computeSwapUsage(memStats *cgroups.MemoryStats) uint64 {
 func isMemoryUnlimited(v uint64) bool {
 	return v == math.MaxUint64
 }
+
+// snapshotHostNetworkLinks builds a snapshot of the host namespace's network
+// links with pod-owned interfaces (e.g. veth peers created by the CNI plugin)
+// filtered out. The resulting snapshot is used by all hostNetwork sandboxes so
+// their metrics only report genuine host interfaces, not pod-owned veth pairs.
+func (ss *StatsServer) snapshotHostNetworkLinks(sandboxes []*sandbox.Sandbox) {
+	ss.hostNetworkLinks = nil
```
haircommander: do we need a lock or is the lock held rn

dgrisonnet: The callers hold the lock:
  - update https://github.com/cri-o/cri-o/blob/dd8d72917d9edc1187fc3475c5c43f89d5520eca/internal/lib/statsserver/stats_server.go#L92-L96
  - MetricsForPodSandbox https://github.com/cri-o/cri-o/blob/dd8d72917d9edc1187fc3475c5c43f89d5520eca/internal/lib/statsserver/stats_server.go#L274-L279
  - MetricsForPodSandboxList https://github.com/cri-o/cri-o/blob/dd8d72917d9edc1187fc3475c5c43f89d5520eca/internal/lib/statsserver/stats_server.go#L286-L291

dgrisonnet:
Holding to accommodate for the data source changes from https://github.com/cri-o/cri-o/pull/10204

/hold

internal/lib/statsserver/stats_server.go
```diff
@@ -291,3 +309,19 @@ func (ss *StatsServer) RemoveMetricsForPodSandbox(sb *sandbox.Sandbox) {
 
 	delete(ss.sboxMetrics, sb.ID())
 }
+
+// buildPodIfacesExclusionSet returns the set of host-side interface names
+// owned by non-hostNetwork sandboxes.
+func buildPodIfacesExclusionSet(sandboxes []*sandbox.Sandbox) map[string]struct{} {
+	podIfaces := make(map[string]struct{})
```
haircommander, line 316: a limitation of this approach is we still overreport host-side interfaces. it's better because we're excluding likely the majority of interfaces, but we still have a cardinality dillema. I wonder if anything would break if we just reported hostnetwork metrics once under a special name or something, and then reported no network metrics for hostnetwork pods...

haircommander: my approach would be more of a breaking change though which may require a config field, and yours probably isn't actually a breaking change

dgrisonnet: Something we could also do is have an exception list on top of that, like the one cadvisor has: https://github.com/google/cadvisor/blob/6a0c4f2539a8b1c11215804dee404c06fe3e0218/lib/utils/sysinfo/sysinfo.go#L118
But as you correctly pointed, there could still be cardinality issues regardless.

Having only one entry for hostnetwork metrics instead of one per pods is the most logical approach, but as you said, it could break stuff. The first thing that comes to mind would be any pod-level network dashboards.

dgrisonnet: How about doing both?

1. Add a non-breaking change that would limit the number of interfaces exposed by:
.1  Scoping non host-network enabled pods to their network namespace
.2 Removing pod-specific interfaces from host-network pod metrics
2. Add a breaking change behind a config flag that would either:
.1 Merge all host network pods into one single entry
.2 Skip network metrics for host network pods and instead expose two set of metrics. A host-specific version of the existing container-level network metrics and another one exposing whether a pod has host-network enabled or not. The goal is that users will be able to continue matching host network pods with their network metrics, but at a lower cost.

PR #10208, opened 2026-07-31
highperfhooks: move cpuset cleanup to post stop

haircommander:
#### What type of PR is this?

/kind bug
when a container gracefully shuts down, prestop hooks don't run (it's already stopped) that's generally fine. However, in the case where an init container runs and has an overlapping range with a container that will run after it, we will fail to run that container.

Move the cpuset cleanup to poststop so it runs on graceful exit

```release-note
fix a bug in high performance hooks where an init container would continue to hold a cpuset that could be used by a regular container in the same pod. 
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CPU load-balancing restoration after containers stop naturally.
  * Ensured shared CPU assignments are restored during post-stop cleanup.
  * Prevented cleanup failures when system-managed cgroups have already been removed.
  * Resolved stale exclusive CPU assignments remaining on pod cgroups after container termination.

* **Tests**
  * Added regression coverage for post-stop CPU assignment cleanup, including immediately exiting containers.

### DIFF
--- a/internal/runtimehandlerhooks/high_performance_hooks_linux.go
+++ b/internal/runtimehandlerhooks/high_performance_hooks_linux.go
@@ -388,37 +388,6 @@ func (h *HighPerformanceHooks) PreStop(ctx context.Context, c *oci.Container, s 
 		return nil
 	}
 
-	containerName := c.CRIContainer().GetMetadata().GetName()
-
-	// disable the CPU load balancing for the container CPUs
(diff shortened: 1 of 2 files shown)


haircommander:
/retest
@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):

PR #10213, opened 2026-08-02
fix: make runtime handlers added by SIGHUP reload available to RunPodSandbox

### LINKED ISSUE #10202: Runtime handler added by SIGHUP appears in crio status config but remains unavailable to RunPodSandbox
### What happened?

I added a valid runtime-handler drop-in to a running CRI-O instance and reloaded CRI-O using SIGHUP through:
```
$ systemctl reload crio
```
The reload command exited successfully. Before the reload, `crio status config` listed only the original `crun` and `runc` handlers. Immediately after the reload, it also listed the newly added `kata-qemu` handler.

Despite the handler appearing in `crio status config`, both an existing pod and a newly created pod using that handler continued to fail during `RunPodSandbox`:
```
rpc error: code = Unknown desc = failed to find runtime handler kata-qemu
from runtime list map[crun:0xdf5a0256000 runc:0xdf5a0256540]
```
The runtime list in the error was unchanged from before the reload. It did not contain `kata-qemu`.

This looks like two live copies of the CRI-O configuration: the inspection endpoint sees the reloaded runtime map, while the sandbox runtime continues using the original map.

After the valid runtime drop-in was added and CRI-O successfully processed SIGHUP, I expected the new handler to become available for new RunPodSandbox requests without restarting CRI-O.

This appears to be the intended behavior documented for the live-reloadable [crio.runtime.runtimes table](https://github.com/cri-o/cri-o/blob/v1.36.0/docs/crio.conf.5.md#L350-L369) and originally introduced by [#6056](https://github.com/cri-o/cri-o/pull/6056).

1. Start CRI-O with only the normal runc and crun handlers
2. Install a valid additional runtime and its configuration. In this reproduction, Kata Containers 3.28.0 was installed under `/opt/kata`.
3. Create /etc/crio/crio.conf.d/99-kata-deploy
```
[crio.runtime.runtimes.kata-qemu]
runtime_path = "/opt/kata/bin/containerd-shim-kata-v2"
runtime_type = "vm"
runtime_root = "/run/vc"
runtime_config_path = "/opt/kata/share/defaults/kata-containers/runtimes/qemu/configuration-qemu.toml"
privileged_without_host_devices = true
```
4. Create the RuntimeClass
```
apiVersion: node.k8s.io/v1
kind: RuntimeClass
metadata:
  name: kata-qemu
handler: kata-qemu
```
5. Reload crio `systemctl reload crio`
6. Verify that it now contains kata-qemu `crio status config`
7. Create a second fresh pod after the reload using the RuntimeClass
8. Observe that the fresh pod fails with the same stale runtime list

This behavior appears consistent with CRI-O retaining two relevant configuration objects:

1. `Server.config` is stored as a value rather than a pointer:  
   [`server/server.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/server/server.go#L68-L76)

2. `Server.New` passes the original configuration pointer to `lib.New`, but stores a value copy in `Server.config`:  
   [`server/server.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/server/server.go#L396-L480)

3. `lib.New` passes the original configuration pointer to `oci.New`:  
   [`internal/lib/container_server.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/internal/lib/container_server.go#L103-L158)

4. `oci.Runtime` retains the original pointer, and `ValidateRuntimeHandler` reads `r.config.Runtimes`:  
   [`internal/oci/oci.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/internal/oci/oci.go#L45-L123)

5. The SIGHUP watcher reloads `s.config`, which is the value copy:  
   [`server/server.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/server/server.go#L619-L653)

6. `ReloadRuntimes` replaces the runtime map on that copy:

   ```go
   c.Runtimes = newConfig.Runtimes
   ```

   [`pkg/config/reload.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/pkg/config/reload.go#L309-L347)

7. The `/config` inspection endpoint serializes `s.config`. This explains why `crio status config` displays the newly reloaded handler:  
   [`server/inspect.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/server/inspect.go#L144-L161)

8. Sandbox creation validates the requested handler through `ContainerServer.Runtime()`, which retains the original configuration pointer:  
   [`server/sandbox_run.go`](https://github.com/cri-o/cri-o/blob/v1.36.0/server/sandbox_run.go#L50-L64)

The original runtime-addition implementation in [#6056](https://github.com/cri-o/cri-o/pull/6056) mutated the existing runtime map in place. [#6248](https://github.com/cri-o/cri-o/pull/6248) subsequently changed this behavior to replace the entire runtime map so that complete runtime-table updates could be supported.

Because the original and copied configurations initially share the same map, replacing the map only on `Server.config` may leave `oci.Runtime.config.Runtimes` referring to the old map. This is a source-level hypothesis that is consistent with the observed behavior, but it is not yet a confirmed root cause.

<details>

```console
$ crio --version
crio version 1.36.0-16.81af8589483.el9
   GitCommit:      unknown
   GitCommitDate:  unknown
   GitTreeState:   clean
   BuildDate:      2019-05-08T00:00:00Z
   GoVersion:      go1.26.2 140-3
   Compiler:       gc
   Platform:       linux/amd64
   Linkmode:       dynamic
   BuildTags:
     selinux
     seccomp
     exclude_graphdriver_btrfs
     containers_image_ostree_stub
     containers_image_openpgp
   LDFlags:          unknown
   SeccompEnabled:   true
   AppArmorEnabled:  false```

```console
{
  "clientVersion": {
    "major": "1",
    "minor": "35",
    "gitVersion": "v1.35.4",
    "gitCommit": "7b8c6cf0edd376b3d7c2f255142977c7f93db258",
    "gitTreeState": "clean",
    "buildDate": "2026-04-15T18:01:22Z",
    "goVersion": "go1.26.2",
    "compiler": "gc",
    "platform": "darwin/arm64"
  },
  "kustomizeVersion": "v5.7.1",
  "serverVersion": {
    "major": "1",
    "minor": "36",
    "emulationMajor": "1",
    "emulationMinor": "36",
    "minCompatibilityMajor": "1",
    "minCompatibilityMinor": "35",
    "gitVersion": "v1.36.1",
    "gitCommit": "5fb3e8384cb746dc3c7aac6c7a433db1cb16649b",
    "gitTreeState": "clean",
    "buildDate": "2026-05-19T10:46:32Z",
    "goVersion": "go1.26.3 140-3",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
NAME="Oracle Linux Server"
VERSION="9.7"
ID="ol"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="9.7"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Oracle Linux Server 9.7"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:oracle:linux:9:7:server"
HOME_URL="https://linux.oracle.com/"
BUG_REPORT_URL="https://github.com/oracle/oracle-linux"

ORACLE_BUGZILLA_PRODUCT="Oracle Linux 9"
ORACLE_BUGZILLA_PRODUCT_VERSION=9.7
ORACLE_SUPPORT_PRODUCT="Oracle Linux"
ORACLE_SUPPORT_PRODUCT_VERSION=9.7
$ uname -a
Linux <hostname-redacted> 6.12.0-203.76.7.1.el9uek.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Jun 1 14:29:53 PDT 2026 x86_64 x86_64 x86_64 GNU/Linux
```

</details>

<details>
Using Oracle Cloud environment
</details>

1991santhu:
## What type of PR is this?

/kind bug

## What this PR does / why we need it

A runtime handler added through a drop-in config and picked up by SIGHUP shows up in `crio status config`, but `RunPodSandbox` keeps rejecting it:

```
failed to find runtime handler kata-qemu from runtime list map[crun:0x... runc:0x...]
```

`Server.config` is a value copy (`server/server.go:469` stores `*config`), while `lib.New` / `oci.New` retain the original `*config.Config`. The SIGHUP handler calls `s.config.Reload()`, so only the copy is updated, and `oci.Runtime` goes on resolving handlers against the map it was constructed with.

The reload path already pushes reloaded values into subsystems that keep their own state — `UpdatePinnedImagesList`, `SetPinnedImageRegexps`. The runtime handler map was just never wired in. This follows that same pattern:

- `oci.Runtime.UpdateRuntimeConfig()`, guarded by a new `RWMutex`. The existing `runtimeImplMapMutex` covers `runtimeImplMap` only, and the handler lookups read `r.config.Runtimes` with no synchronisation at all, so the new mutex covers the map now that it can be written after startup.
- `ValidateRuntimeHandler` takes the read lock and snapshots the map it reports in its error message.
- `startReloadWatcher` calls it after a successful reload, alongside the existing pinned-image push.

I also tried simply sharing the `*config.Config` pointer with `oci.Runtime` instead. It's smaller and it compiles, but reload mutates the config in place while gRPC handlers read it concurrently, and nothing in `pkg/config` is synchronised — that trades a stale-read bug for a data race, so I didn't go that way. Happy to revisit if maintainers prefer a different shape.

## Which issue(s) this PR fixes

Fixes #10202

## Special notes for your reviewer

The regression test covers a handler absent at construction time, resolving after `UpdateRuntimeConfig`, and pre-existing handlers still resolving.

Full `internal/oci` suite run on Linux: 84 passed, 0 failed, 8 skipped.

Happy to add a server-level test exercising the SIGHUP path end to end if that would be preferred.

## Does this PR introduce a user-facing change?

```release-note
Fixed runtime handlers added by a SIGHUP configuration reload not becoming available to RunPodSandbox until CRI-O was restarted.
```

## Summary by CodeRabbit

* **New Features**
  * Runtime configuration changes now take effect safely when server configuration is reloaded.
  * Newly configured runtime handlers become available immediately, while existing handlers continue to work.
  * The configured default runtime is updated alongside other runtime settings.

* **Bug Fixes**
  * Improved runtime configuration consistency during concurrent access and reloads.
  * Runtime handler validation now reflects the latest available configuration.

### DIFF
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -46,8 +46,26 @@ const (
 // information about the runtime.
 type Runtime struct {
 	config              *config.Config
+	configMutex         sync.RWMutex
 	runtimeImplMap      map[string]RuntimeImpl
 	runtimeImplMapMutex sync.RWMutex
+}
+
+// UpdateRuntimeConfig replaces the runtime handler configuration this Runtime
+// resolves handlers against. It is called after a SIGHUP configuration reload
+// so that newly added runtime handlers become available to RunPodSandbox
+// without restarting CRI-O.
+//
+// Server.config is a distinct value from the *config.Config this Runtime was
+// constructed with, so reloading the former does not update the latter. The
+// reload path must push the new values in explicitly, in the same way it
+// already does for the pinned image list.
+func (r *Runtime) UpdateRuntimeConfig(runtimes config.Runtimes, defaultRuntime string) {
+	r.configMutex.Lock()
+	defer r.configMutex.Unlock()
+
+	r.config.Runtimes = runtimes
+	r.config.DefaultRuntime = defaultRuntime
 }
 
 // RuntimeImpl is an interface used by the caller to interact with the
@@ -110,10 +128,14 @@ func (r *Runtime) ValidateRuntimeHandler(handler string) (*config.RuntimeHandler
 		return nil, errors.New("empty runtime handler")
 	}
 
+	r.configMutex.RLock()
 	runtimeHandler, ok := r.config.Runtimes[handler]
+	runtimes := r.config.Runtimes
+	r.configMutex.RUnlock()
+
 	if !ok {
 		return nil, fmt.Errorf("failed to find runtime handler %s from runtime list %v",
-			handler, r.config.Runtimes)
+			handler, runtimes)
 	}
 
 	if runtimeHandler.RuntimePath == "" {
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -118,6 +118,9 @@ func New(c *config.Config) (*Runtime, error) {
 
 // Runtimes returns the map of OCI runtimes.
 func (r *Runtime) Runtimes() config.Runtimes {
+	r.configMutex.RLock()
+	defer r.configMutex.RUnlock()
+
 	return r.config.Runtimes
 }
 
@@ -147,7 +150,13 @@ func (r *Runtime) ValidateRuntimeHandler(handler string) (*config.RuntimeHandler
 
 func (r *Runtime) getRuntimeHandler(handler string) (*config.RuntimeHandler, error) {
 	// Define the current runtime handler as the default runtime handler.
+	// Read both fields under a single lock so the default handler cannot be
+	// resolved against a runtime map from a different reload. The lock is
+	// released before ValidateRuntimeHandler below, which takes it itself:
+	// a blocked writer stops new readers, so a nested RLock could deadlock.
+	r.configMutex.RLock()
 	rh := r.config.Runtimes[r.config.DefaultRuntime]
+	r.configMutex.RUnlock()
 
 	// Override the current runtime handler with the runtime handler
 	// corresponding to the runtime handler key provided with this
--- a/internal/oci/oci.go
+++ b/internal/oci/oci.go
@@ -45,7 +45,16 @@ const (
 // Runtime is the generic structure holding both global and specific
 // information about the runtime.
 type Runtime struct {
-	config              *config.Config
+	config *config.Config
+	// runtimes and defaultRuntime are owned by this Runtime rather than read
+	// through config. The *config.Config that config points at is the one
+	// Config.GetData returns, which is the same pointer held by ContainerServer
+	// and by HooksRetriever, and HooksRetriever reads Runtimes without any
+	// synchronisation from gRPC handler goroutines. Writing the reloaded values
+	// back through config would therefore race with those readers, so the
+	// reload updates these fields instead and leaves the shared config alone.
+	runtimes            config.Runtimes
+	defaultRuntime      string
 	configMutex         sync.RWMutex
 	runtimeImplMap      map[string]RuntimeImpl
 	runtimeImplMapMutex sync.RWMutex
@@ -60,12 +69,15 @@ type Runtime struct {
 // constructed with, so reloading the former does not update the latter. The
 // reload path must push the new values in explicitly, in the same way it
 // already does for the pinned image list.
+//
+// The values are stored on the Runtime rather than written back through the
+// shared *config.Config; see the field comments above.
 func (r *Runtime) UpdateRuntimeConfig(runtimes config.Runtimes, defaultRuntime string) {
 	r.configMutex.Lock()
 	defer r.configMutex.Unlock()
 
-	r.config.Runtimes = runtimes
-	r.config.DefaultRuntime = defaultRuntime
+	r.runtimes = runtimes
+	r.defaultRuntime = defaultRuntime
 }
 
 // RuntimeImpl is an interface used by the caller to interact with the
@@ -112,6 +124,8 @@ func New(c *config.Config) (*Runtime, error) {
 
 	return &Runtime{
 		config:         c,
+		runtimes:       c.Runtimes,
+		defaultRuntime: c.DefaultRuntime,
 		runtimeImplMap: make(map[string]RuntimeImpl),
 	}, nil
 }
@@ -121,7 +135,7 @@ func (r *Runtime) Runtimes() config.Runtimes {
 	r.configMutex.RLock()
 	defer r.configMutex.RUnlock()
 
-	return r.config.Runtimes
+	return r.runtimes
 }
 
 // ValidateRuntimeHandler returns an error if the runtime handler string
@@ -132,8 +146,8 @@ func (r *Runtime) ValidateRuntimeHandler(handler string) (*config.RuntimeHandler
 	}
 
 	r.configMutex.RLock()
-	runtimeHandler, ok := r.config.Runtimes[handler]
-	runtimes := r.config.Runtimes
+	runtimeHandler, ok := r.runtimes[handler]
+	runtimes := r.runtimes
 	r.configMutex.RUnlock()
 
 	if !ok {
@@ -155,7 +169,7 @@ func (r *Runtime) getRuntimeHandler(handler string) (*config.RuntimeHandler, err
 	// released before ValidateRuntimeHandler below, which takes it itself:
 	// a blocked writer stops new readers, so a nested RLock could deadlock.
 	r.configMutex.RLock()
-	rh := r.config.Runtimes[r.config.DefaultRuntime]
+	rh := r.runtimes[r.defaultRuntime]
 	r.configMutex.RUnlock()
 
 	// Override the current runtime handler with the runtime handler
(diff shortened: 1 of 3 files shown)


1991santhu:
Good catch, thanks, that was a real gap. `UpdateRuntimeConfig` writes both fields but I'd only guarded `ValidateRuntimeHandler`.

`getRuntimeHandler` was the worse of the two. It read `Runtimes` and `DefaultRuntime` in one expression, so a concurrent reload could resolve the default handler against a runtime map from a different configuration. Both are now read under a single `RLock`.

One thing worth noting for reviewers: I release the lock before the `ValidateRuntimeHandler` call rather than holding it across, since that method takes the same `RLock` and a blocked writer stops new readers, so nesting would risk a deadlock.

Verified with `go test -race ./internal/oci/`, passes clean.

1991santhu:
I have to retract something I wrote in the PR description.

I said I'd deliberately avoided sharing the `*config.Config` pointer with `oci.Runtime`, because doing so would trade a stale read for a data race. The reasoning was fine but the premise was wrong. `Config.GetData()` returns the receiver, so the pointer was already shared. `server.go:405`, `container_server.go:115` (which feeds `oci.New`) and `server.go:477` (`NewHooksRetriever`) all hold the same object, and `runtime_handler_hooks_linux.go:45` reads `Runtimes` with no lock from five request paths.

So writing `r.config.Runtimes` under the new mutex only covered the readers inside `internal/oci`, and raced with the rest. The tests can't catch it because they build a `Runtime` on its own with no second holder, which is why `-race` stayed green.

The fix is to stop writing through the shared config. `Runtime` keeps its own `runtimes` and `defaultRuntime`, seeded in `New`, and nothing mutates the shared `*config.Config` any more.

One consequence worth flagging: `HooksRetriever` still won't see handlers added by a reload. That's what it did before this PR too, so nothing regresses, but if it should see them that wants its own synchronisation rather than falling out of shared pointers. Happy to file that separately.

I also fixed the comment on the reload path, which claimed ownership that isn't there.

`go test -race ./internal/oci/` on Linux: ok, 36.6s.

internal/oci/oci.go
```diff
@@ -45,11 +45,41 @@ const (
...
+	config *config.Config
+	// runtimes and defaultRuntime are owned by this Runtime rather than read
+	// through config. The *config.Config that config points at is the one
+	// Config.GetData returns, which is the same pointer held by ContainerServer
+	// and by HooksRetriever, and HooksRetriever reads Runtimes without any
+	// synchronisation from gRPC handler goroutines. Writing the reloaded values
+	// back through config would therefore race with those readers, so the
+	// reload updates these fields instead and leaves the shared config alone.
+	runtimes            config.Runtimes
+	defaultRuntime      string
```
bitoku, line 57: Is it a reason why we have `runtimes` and `defaultRuntime`?
I think the fact that there's other component that read the same `*config.Config` won't be the reason, because it won't write configs anyway.

1991santhu: You're right that no other component writes the config, but that's the reason rather than the counter-argument.

`HooksRetriever.Get` reads the map directly and holds no lock while doing it:

https://github.com/cri-o/cri-o/blob/main/internal/runtimehandlerhooks/runtime_handler_hooks_linux.go#L45

```go
runtimeConfig, ok := hr.config.Runtimes[runtimeName]
```

The only mutexes in that file are `cpusetLock` and `updateIRQSMPAffinityLock`, and both belong to the hook structs it builds, not to this read. It's reached from `container_create.go`, `container_start.go`, `container_stop.go`, `sandbox_run_linux.go` and `server.go`, so it runs concurrently on CRI request goroutines.

That's exactly why it's unprotected: a component that only ever reads has no reason to take a lock. So if the SIGHUP reload wrote the new handlers back through the shared `*config.Config`, that write would be the only writer in the system, racing against unsynchronised concurrent map reads. In Go that isn't just a race-detector finding — concurrent map read and map write is a fatal runtime error, so a reload landing while a pod is being created could take the node down.

Keeping `runtimes` and `defaultRuntime` on the `Runtime` means the reload mutates only state that `Runtime` owns and guards, and the shared config stays read-only for everyone who reads it without a lock.

The alternative I considered was routing `HooksRetriever` through `Runtime.Runtimes()` so there's one guarded owner of the handler map, and dropping these fields. That's the cleaner end state, but it changes a constructor signature and pulls `runtimehandlerhooks` into a dependency on `oci`, which felt like more than this fix should carry. Happy to do it that way if you'd prefer it as the shape, either here or as a follow-up.

PR #10231, opened 2026-08-13
[release-1.34] User namespace - Handle UID/GID mappings from the infra container

Chandan9112:
/kind bug

#### What type of PR is this?

This is a manual cherry-pick of #10014 to `release-1.34` (OCP 4.21).

When using user namespaces with ID-mapped mounts, containers fail with "permission denied" errors after:
1. CRI-O is restarted
2. The real container process is killed (kill 1)

This happens because the namespace mappings are not loaded on restart for running containers.

The fix loads UID and GID mappings from the sandbox's saved `config.json` during CRI-O startup, ensuring containers retain proper user namespace configuration across restarts.

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-76990

- Cherry-pick of #10014 (merged to `main`)
- See #10075 for the equivalent cherry-pick to `release-1.35` (OCP 4.22).

```release-note
User namespace UID/GID mappings are now correctly restored after CRI-O restart, fixing "permission denied" errors for containers using user namespaces with ID-mapped mounts.
```

### DIFF
--- a/internal/lib/container_server.go
+++ b/internal/lib/container_server.go
@@ -367,6 +367,13 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
 	scontainer.SetSpec(&m)
 	scontainer.SetMountPoint(m.Annotations[annotations.MountPoint])
 
+	// Restore ID mappings from the OCI spec if user namespace is in use
+	if m.Linux != nil && len(m.Linux.UIDMappings) > 0 && len(m.Linux.GIDMappings) > 0 {
+		mappings := ConvertOCIToStorageIDMappings(m.Linux.UIDMappings, m.Linux.GIDMappings)
+		scontainer.SetIDMappings(mappings)
+		log.Debugf(ctx, "Restored ID mappings for sandbox %s from OCI spec", id)
+	}
+
 	if err := restoreVolumes(&m, scontainer); err != nil {
 		return sb, fmt.Errorf("restore volumes: %w", err)
 	}
--- a/test/ctr_userns.bats
+++ b/test/ctr_userns.bats
@@ -3,9 +3,6 @@
 load helpers
 
 function setup() {
-	if [ -z "$CONTAINER_UID_MAPPINGS" ]; then
-		skip "userns testing not enabled"
-	fi
 	setup_test
 	start_crio
 }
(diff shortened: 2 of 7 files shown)


saschagrunert (approved):
/ok-to-test

test/ctr_userns.bats
```diff
@@ -14,16 +11,94 @@ function teardown() {
 	cleanup_test
 }
 
+function run_userns_container_test() {
+	# Create sandbox config with userns options using jq
+	jq '.linux.security_context.namespace_options.userns_options = {
+		"mode": 0,
+		"uids": [{"container_id": 0, "host_id": 100000, "length": 100000}],
+		"gids": [{"container_id": 0, "host_id": 200000, "length": 100000}]
+	}' "$TESTDATA"/sandbox_config.json > "$TESTDIR"/sandbox_userns.json
```
saschagrunert, line 19: nit: This 5-line jq block is duplicated in all three helper functions (`run_userns_container_test`, `run_userns_restart_test`, `run_userns_kill_restart_test`). Consider extracting it into a shared helper:

```bash
function create_userns_sandbox_config() {
	jq '.linux.security_context.namespace_options.userns_options = {
		"mode": 0,
		"uids": [{"container_id": 0, "host_id": 100000, "length": 100000}],
		"gids": [{"container_id": 0, "host_id": 200000, "length": 100000}]
	}' "$TESTDATA"/sandbox_config.json > "$TESTDIR"/sandbox_userns.json
}
```

internal/lib/container_server.go
```diff
@@ -367,6 +367,14 @@ func (c *ContainerServer) LoadSandbox(ctx context.Context, id string) (sb *sandb
 	scontainer.SetSpec(&m)
 	scontainer.SetMountPoint(m.Annotations[annotations.MountPoint])
 
+	// Restore ID mappings from the OCI spec if user namespace is in use
+	if m.Linux != nil && len(m.Linux.UIDMappings) > 0 && len(m.Linux.GIDMappings) > 0 {
+		if mappings := ConvertOCIToStorageIDMappings(m.Linux.UIDMappings, m.Linux.GIDMappings); mappings != nil {
+			scontainer.SetIDMappings(mappings)
+			log.Debugf(ctx, "Restored ID mappings for sandbox %s from OCI spec", id)
+		}
```
saschagrunert: nit: The outer guard already ensures both slices are non-empty, and `ConvertOCIToStorageIDMappings` only returns nil when either input is empty (line 11 of `idmappings_linux.go`). So the inner `mappings != nil` check is always true on this code path. Could simplify to:

```go
if m.Linux != nil && len(m.Linux.UIDMappings) > 0 && len(m.Linux.GIDMappings) > 0 {
	mappings := ConvertOCIToStorageIDMappings(m.Linux.UIDMappings, m.Linux.GIDMappings)
	scontainer.SetIDMappings(mappings)
	log.Debugf(ctx, "Restored ID mappings for sandbox %s from OCI spec", id)
}
```

saschagrunert (approved):

PR #10235, opened 2026-08-13
hack: add containerized build and unit test environment

bartwensley:
#### What type of PR is this?
/kind other

Add hack/Containerfile.dev, a UBI9-based image with all build dependencies, so CRI-O can be built and unit tested without installing toolchain and C dependencies on the host. Integration tests still require a full system and are out of scope for the container.

The Go toolchain is installed for the target architecture (via TARGETARCH) and the download is verified against a pinned checksum.

Two unit tests assumed capabilities that are not guaranteed in every environment, so guard them accordingly:

- SpecSetLinuxContainerResources swap tests dereferenced the OCI spec's memory swap field, which is only set when the node has cgroup memory swap support. Skip them when node.CgroupHasMemorySwap() is false to avoid a nil pointer dereference.

- TestAddOCIBindsRROMountsError's bidirectional-propagation case reaches ensureShared() before the assertion under test, which requires the host path to be a shared mount. Skip that case when it is not.

Document the workflow in install.md.

None

None

```release-note
None
```

## Summary by CodeRabbit

- **Documentation**
  - Added instructions for setting up a container-based development environment with Podman.
  - Documented how to build the development image and run build, lint, and unit-test checks inside the container.
  - Clarified that integration tests are not supported in containerized development environments.
  - Documented a consistent, verified Go toolchain for container-based development.

- **Tests**
  - Improved test reliability by detecting unavailable host capabilities and skipping incompatible memory, swap, and mount-propagation scenarios.

### DIFF
--- a/hack/Containerfile.dev
+++ b/hack/Containerfile.dev
@@ -1,0 +1,36 @@
+FROM registry.access.redhat.com/ubi9/ubi:9.8
+
+RUN dnf install -y \
+    autoconf \
+    automake \
+    gcc \
+    git \
+    glib2-devel \
+    glibc-static \
+    gpgme-devel \
+    libassuan-devel \
+    libcap-devel \
+    libseccomp-devel \
+    libtool \
+    make \
+    pkgconfig \
+    && dnf clean all
+
+# Install Go. Keep GOVERSION in sync with go.mod, and update the pinned
+# checksums below when bumping it (see https://go.dev/dl/?mode=json&include=all).
+ARG TARGETARCH
+ENV GOVERSION=1.26.4
+RUN case "${TARGETARCH}" in \
+        amd64)   GO_SHA256=1153d3d50e0ac764b447adfe05c2bcf08e889d42a02e0fe0259bd47f6733ad7f ;; \
+        arm64)   GO_SHA256=ef758ae7c6cf9267c9c0ef080b8965f453d89ab2d25d9eb22de4405925238768 ;; \
+        ppc64le) GO_SHA256=53f49b8c7eace2d30389327b4a516b13321f90377fdf5929a6b63174609bc22e ;; \
+        s390x)   GO_SHA256=1e12a2318f3dd4c57027c865f6cb51f5d1e36b02d10f3e6316ce7b61a27b3ca1 ;; \
+        *) echo "unsupported TARGETARCH: ${TARGETARCH:-unset}" >&2; exit 1 ;; \
+    esac && \
+    curl -fsSL "https://go.dev/dl/go${GOVERSION}.linux-${TARGETARCH}.tar.gz" -o /tmp/go.tar.gz && \
+    echo "${GO_SHA256}  /tmp/go.tar.gz" | sha256sum -c - && \
+    tar -C /usr/local -xzf /tmp/go.tar.gz && \
+    rm -f /tmp/go.tar.gz
+ENV PATH="/usr/local/go/bin:${PATH}"
+
+WORKDIR /src
--- a/install.md
+++ b/install.md
@@ -27,6 +27,7 @@ It is assumed you are running a Linux machine.
   - [Build](#build)
     - [Install with Ansible](#install-with-ansible)
     - [Build Tags](#build-tags)
+    - [Container-based build environment](#container-based-build-environment)
   - [Static builds](#static-builds)
   - [Download conmon](#download-conmon)
 - [Setup CNI networking](#setup-cni-networking)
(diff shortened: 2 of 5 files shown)


bitoku:
Thank you! I think we should have a unit test CI with this devcontainer to prevent it from breaking silently in the future.

install.md
```diff
@@ -350,6 +351,30 @@ which uses the following buildtags.
 
 <!-- markdownlint-enable MD013 -->
 
+#### Container-based build environment
+
+As an alternative to installing build dependencies on the host, a
+`Containerfile` is provided to create a container image with all
+required build dependencies. This can be used to build CRI-O and run
+unit tests without modifying the host system.
```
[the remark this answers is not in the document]
bartwensley: This workflow is intended to be used with rootless Podman (the default for a regular user), not rootful. With rootless Podman the container's root is mapped through a user namespace to the invoking user's UID, so generated files in the mounted tree are owned by that user, not root. Running the container under sudo/rootful Podman isn't the intended usage, so I'd prefer not to document --user/--userns workarounds for a mode I'm not recommending.

The read-write bind mount is intentional: make all needs to write build outputs (bin/, build/) back into the checkout, which is the point of the mount. The :Z suffix is the standard Podman option for SELinux relabeling and is a no-op on systems without SELinux; I've kept it so the example works out of the box on SELinux-enforcing hosts.

hack/Containerfile.dev
```diff
@@ -0,0 +1,34 @@
...
+ENV GOVERSION=1.26.4
+RUN case "${TARGETARCH}" in \
+        amd64) GO_SHA256=1153d3d50e0ac764b447adfe05c2bcf08e889d42a02e0fe0259bd47f6733ad7f ;; \
+        arm64) GO_SHA256=ef758ae7c6cf9267c9c0ef080b8965f453d89ab2d25d9eb22de4405925238768 ;; \
+        *) echo "unsupported TARGETARCH: ${TARGETARCH:-unset}" >&2; exit 1 ;; \
+    esac && \
+    curl -fsSL "https://go.dev/dl/go${GOVERSION}.linux-${TARGETARCH}.tar.gz" -o /tmp/go.tar.gz && \
+    echo "${GO_SHA256}  /tmp/go.tar.gz" | sha256sum -c - && \
+    tar -C /usr/local -xzf /tmp/go.tar.gz && \
+    rm -f /tmp/go.tar.gz
```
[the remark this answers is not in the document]
bartwensley: OK - I made the change.

saschagrunert (approved):

hack/Containerfile.dev
```diff
@@ -0,0 +1,36 @@
...
+    libseccomp-devel \
+    libtool \
+    make \
+    pkgconfig \
+    && dnf clean all
+
+# Install Go. Keep GOVERSION in sync with go.mod, and update the pinned
+# checksums below when bumping it (see https://go.dev/dl/?mode=json&include=all).
+ARG TARGETARCH
+ENV GOVERSION=1.26.4
```
bitoku, line 22: This should be pinned in dependencies.yaml

bartwensley: @bitoku - what are you asking me to do? Should I update the comment to say `dependencies.yaml` instead of `go.mod`?

bitoku: We should add a new refPath for Containerfile.dev in dependencies.yaml so that we won't forget to update this.

https://github.com/cri-o/cri-o/blob/1c29c66fb9c38140fac37e4c4f757eba39da96e8/dependencies.yaml#L3-L10

bartwensley: Thanks - I made the change.

bartwensley:
@saschagrunert and @bitoku - I have made the requested changes. The unit tests are passing in the CI. Since I only changed unit test code I don't see how my changes could possibly be contributing to any of the other failures in the CI tests. Are you guys willing to merge this as is or should I be running /retest over and over and over again?

bitoku:
They are caused by openshift base image failure and github runners image change.

You seem not to have addressed this. Are you going to follow it up?
> I think we should have a unit test CI with this devcontainer to prevent it from breaking silently in the future.

bartwensley:
> You seem not to have addressed this. Are you going to follow it up?
> > I think we should have a unit test CI with this devcontainer to prevent it from breaking silently in the future.

I do not plan to do that. I think my contribution is valuable as it stands. I am not at all familiar with the cri-o CI and I don't have time to learn it at this point, especially given how chronically broken the CI is. If a new CI is necessary for the dev container, then I would request that an issue be raised and someone else who knows the CI can take care of it.

bartwensley:
Rebased on main to pick up recent changes and resolve conflict in dependencies.yaml

bartwensley:
@saschagrunert @bitoku - I think all the "required" tests are passing now. Could you merge this?

bartwensley:
Looks like more CI environment errors...

/retest

bartwensley:
@saschagrunert - the failed test has nothing to do with my change. Any chance you could merge this?

PR #10242, opened 2026-08-16
ci: remove pre-installed runc and crun

kolyshkin:
#### What type of PR is this?

/kind ci

#### What this PR does / why we need it:

**1. `ci: remove pre-installed runc and crun`**

The GitHub runner image ships a static podman bundle in `/usr/local/bin`
(actions/runner-images#14412), which includes its own runc and crun. As
`/usr/local/bin` comes first in `PATH`, those shadow the runtimes installed by
`scripts/github-actions-setup`: the tests were run with runc 1.4.3 rather than
the pinned version and, worse, with a crun built without systemd support:

```
crun version 1.28
+SELINUX +APPARMOR +CAP +SECCOMP +EBPF +JSON_C
```

The test suite uses the systemd cgroup manager, so every container creation
failed with `container create failed: systemd not supported: Not supported`,
failing all `integration` and `critest` jobs on main and on every open PR.

This removes all the pre-installed runc and crun binaries (in addition to the
bundle, docker brings its own `/usr/bin/runc`), and checks that none is left
before installing ours, so that a future image change fails loudly rather than
silently testing something else.

Same fix as containers/conmon#672.

**2. `ci: bump crun to 1.29.1 and runc to v1.5.1`**

Update the pinned OCI runtimes to their latest releases. The runc checksums are
taken from the published `runc.sha256sum` of the v1.5.1 release.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

`runc` is still installed to `/usr/sbin/runc` (unlike the conmon fix, which
moved it to `/usr/bin`): nothing in the GitHub Actions path hardcodes the
location and `/usr/sbin` is in the runner `PATH`, so this keeps the change
minimal.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime setup by removing preinstalled runtime binaries before installation.
  * Added verification to ensure runtimes are no longer discoverable before proceeding.
  * Helps prevent conflicts caused by previously installed runtime versions.

* **Updates**
  * Updated the included `runc` and `crun` runtimes to newer versions.
  * Refreshed integrity checks for supported architectures.

PR #9907, opened 2026-04-28
Confidential Container: do not pull images on the host

littlejawa:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

With #7471 we introduced a way to support Confidential Container's feature of pulling the image in the guest VM.
When this happens, cri-o is still pulling the image on the host, because kubernetes keeps sending the "PullImageRequest", and cri-o has no reason to not process it.

This has two drawbacks:
- it is a waste of time and resources, as the image that crio will pull and prepare will not be used anyway
- it can block the actual execution of the container in the Confidential Container use case: if the image is encrypted, and the key is not shared with crio, then crio will fail to prepare the image because it can't read it.
  This failure will block the container creation, as kubernetes will not proceed with CreateContainer if PullImage failed.

This PR is meant to make crio skip the pull image phase when the runtime is configured with the "runtime_pull_image" flag introduced in #7471

#### Which issue(s) this PR fixes:

Fixes: #8261

#### Special notes for your reviewer:

This PR is a follow-up to https://github.com/cri-o/cri-o/pull/8008
It tries to address some of the problems raised for it, while rebasing on the latest code.

#### Creating a separate workflow for Confidential Containers

A lot of the work done on this PR compared to #8008 is to add a separate implementation of ImageServer and RuntimeServer to handle the specific behavior of Confidential Containers.
In this respect, I have done the following:
- introduce a "manager" object for each service, allowing to keep two instances of ImageServer and RuntimeServer : one for containers running with a regular workflow (using the existing implementation), and the other for the Confidential Container use case, where the runtime class is configured to delegate the image management to the runtime.
- implemented runtimePulledImageService and runtimePulledRuntimeService
- modify the callers to provide the sandbox object as parameter whenever possible when accessing the ImageServer or RuntimeServer. The sandbox object is used to check which ImageServer/RuntimeServer must be used, and where to store the needed metadata (see below).

#### Reduce the amount of I/O done to retrieve the needed information

To be able to build internal IDs and data structures, we need to retrieve some information from the repository.
In #8008, this was done every time we needed to access this information.
To reduce the amount of I/Os, this PR retrieves the manifests and config during the PullImage request, and stores this information in memory, for later use by the modified services.
This cache is rebuilt from the locally stored metadata when CRI-O is restarted.

#### Store location and data lifecycle

To avoid mixing incomplete image data with the actual image store, we use a separate location for the store.
[...]

PR #10243, opened 2026-08-16
server: fix ContainerEvents unit test flake

kolyshkin:
#### What type of PR is this?

/kind flake

The `ContainerEvents` specs fail intermittently (seen on #10242, and reproducible
locally) with a panic raised from inside `broadcastEvents`. As that is production
code it has no `GinkgoRecover`, so a single bad send takes down the whole unit
test suite rather than failing one spec:

```
Unexpected call to MockRuntimeService_GetContainerEventsServer.Send()
server/container_events.go:78
```

There are three separate problems behind it.

**1. Protobuf messages compared with `reflect.DeepEqual`.** The expectations
passed the message itself, so gomock matched with `gomock.Eq`, i.e.
`reflect.DeepEqual`. That is not a valid way to compare protobuf messages: it
also compares their internal state, which is populated lazily. The server hands
the *very same* message to every client, so as soon as anything looks at it
through the protobuf reflection (gomock does, when formatting a failure),
`DeepEqual` stops matching for all the remaining clients. Instrumenting the
delivery shows it plainly — `c1` matches, `c2` matches nothing:

```
SEND c1 -> "2" deepEqualToEvents=[false true  false]
SEND c2 -> "2" deepEqualToEvents=[false false false]
```

Match with `proto.Equal` instead, against a private copy of the expected message
so that comparing it does not race with the specs sending their own copy.

**2. A `BeforeEach` goroutine closing the next spec's channel.** It closed
`sut.ContainerEventsChan` two seconds later, reading the package level `sut` at
that point rather than when it was spawned. By then `sut` usually belonged to a
later spec, so a spec could have its channel closed by its predecessor while it
was still sending to it. `-race` catches it directly — `container_events_test.go:34`
is the `close()`, `:50` is a *later* spec's send:

```
WARNING: DATA RACE
Write at 0x00c000447430 by goroutine 75:
  runtime.closechan()
  server_test.init.func12.1.1()  container_events_test.go:34
Previous read at 0x00c000447430 by goroutine 76:
  runtime.chansend()
  server_test.init.func12.2.1()  container_events_test.go:50
```

The specs now drive the channel themselves, closing it once everything has been
delivered.

**3. The spec ending with sends still in flight.** The two client spec returned
as soon as it had queued the events, so `mockCtrl.Finish()` in `AfterEach` raced
the in-flight sends. It now waits for the events to be delivered and for both
`GetContainerEvents` calls to return.

Additionally, `GetContainerEvents` now registers the client *before* starting the
broadcaster. The other way around, the broadcaster could deliver already buffered
events while the very client that started it was not in the map yet.

None

Reproducer for the race, which fails on `main` within ~15 iterations and passes
30/30 with this change:

```
ginkgo run --race --repeat=29 --focus=ContainerEvents \
    --tags "test $BUILDTAGS" --mod=vendor ./server
```

The specs no longer sleep, which takes them from about three seconds down to
fifty milliseconds. Full `./server` suite (178 specs) passes, `make lint` is
clean.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container event delivery reliability by ensuring connected clients are registered before events are broadcast.
  * Prevented early buffered events from being missed by clients.

* **Tests**
  * Strengthened container event coverage with synchronization and lifecycle checks.
  * Added validation that clients are ready before test events are sent.

### DIFF
--- a/server/container_events.go
+++ b/server/container_events.go
@@ -31,16 +31,18 @@ func (s *Server) GetContainerEvents(_ *types.GetEventsRequest, ces types.Runtime
 		return nil
 	}
 
-	s.containerEventStreamBroadcaster.Do(func() {
-		// note that this function will run indefinitely until ContainerEventsChan is closed
-		go s.broadcastEvents()
-	})
-
 	conn := &containerEventConn{
 		ch:   make(chan struct{}),
 		once: sync.Once{},
 	}
+	// Register before starting the broadcaster so the client that triggers
+	// it does not miss already buffered events.
 	s.containerEventClients.Store(ces, conn)
+
+	s.containerEventStreamBroadcaster.Do(func() {
+		// note that this function will run indefinitely until ContainerEventsChan is closed
+		go s.broadcastEvents()
+	})
 
 	// wait here until we don't want to send events to this client anymore
 	conn.wait()
--- a/server/container_events_test.go
+++ b/server/container_events_test.go
@@ -1,15 +1,24 @@
 package server_test
 
 import (
+	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/protobuf/proto"
 	types "k8s.io/cri-api/pkg/apis/runtime/v1"
 
 	containereventservermock "github.com/cri-o/cri-o/test/mocks/containereventserver"
 )
+
+// eventTimeout is how long to wait for the events to be delivered. Everything
+// here is in-memory and takes way less than that, but a CI machine running the
+// race detector under load can be slow, and this is not a benchmark.
+const eventTimeout = 30 * time.Second
 
 var events = []types.ContainerEventResponse{
 	{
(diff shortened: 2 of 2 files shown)


server/container_events_test.go
```diff
@@ -22,65 +25,159 @@ var events = []types.ContainerEventResponse{
...
+// it through the protobuf reflection, DeepEqual stops matching for all the
+// remaining clients.
+//
+// The message is cloned because comparing it populates that internal state as
+// well, which would race with the specs sending their own copy of it.
+func sendOf(want *types.ContainerEventResponse) gomock.Matcher {
+	clone := &types.ContainerEventResponse{}
+	proto.Merge(clone, want)
+
+	return gomock.Cond(func(got *types.ContainerEventResponse) bool {
```
saschagrunert: nit: `gomock.Cond` prints expectations as `Cond: server_test.sendOf.func1`, so when a send does not match you see the actual value but not which event was expected. A custom `gomock.Matcher` with a `String()` method would make failures self-diagnosing:

```go
type protoMatcher struct {
	want *types.ContainerEventResponse
}

func (m *protoMatcher) Matches(x any) bool {
	got, ok := x.(*types.ContainerEventResponse)
	return ok && proto.Equal(m.want, got)
}

func (m *protoMatcher) String() string {
	return fmt.Sprintf("proto.Equal(%v)", m.want)
}
```

Not blocking, just a usability thought for future debugging.

server/container_events_test.go
```diff
@@ -22,65 +25,159 @@ var events = []types.ContainerEventResponse{
...
+	done := make(chan struct{})
+
+	go func() {
+		defer GinkgoRecover()
+
+		wg.Wait()
+		close(done)
+	}()
+
+	Eventually(done).Should(BeClosed())
```
saschagrunert: nit: This relies on Gomega's default 1s timeout. Fine for in-memory mocks, but making it explicit (e.g. `.WithTimeout(5 * time.Second)`) would document the intent and avoid surprises if the project ever changes Gomega defaults or this runs under heavy CI load with `-race`.

server/container_events.go
```diff
@@ -31,17 +31,19 @@ func (s *Server) GetContainerEvents(_ *types.GetEventsRequest, ces types.Runtime
...
-		// note that this function will run indefinitely until ContainerEventsChan is closed
-		go s.broadcastEvents()
-	})
-
 	conn := &containerEventConn{
 		ch:   make(chan struct{}),
 		once: sync.Once{},
 	}
+	// Register before starting the broadcaster, otherwise it could deliver
+	// already buffered events while this client is not yet in the map.
```
saschagrunert: nit: This is only actionable for the first client, the one whose call triggers `sync.Once`. For subsequent clients `Do` is a no-op and the broadcaster is already running regardless of ordering. Consider making that scope explicit, e.g.:

```suggestion
	// Register before starting the broadcaster so the client that triggers
	// it does not miss already buffered events.
```

saschagrunert (commented):
LGTM, three minor nits below.

server/container_events_test.go
kolyshkin: Done. One tweak: the matcher has to keep the clone and print that, since formatting a proto goes through ProtoReflect() and would race the spec sending its own copy of the event.

server/container_events_test.go
kolyshkin: Done, added an explicit 30s timeout, applied to both Eventually sites.

server/container_events.go
kolyshkin: Right, took your wording.

saschagrunert (approved):

kolyshkin:
So, the same issue was fixed slightly differently by PR #9907, thus the conflict.

I still feel my fix is more robust, so rebased; @haircommander @bitoku can this please be merged? )

haircommander:
can you fix the lint @kolyshkin ?

haircommander:
jk it's not related
/lgtm

bitoku:
/override ci/prow/ci-cgroupv2-e2e ci/prow/ci-cgroupv2-e2e-crun ci/prow/ci-fedora-critest

PR #10244, opened 2026-08-17
test: add serial-tagged placeholder to prevent empty suite error

jnovy:
When conmon CI runs with `TESTS=critest.bats`, the serial pass
(`--filter-tags 'crio:serial'`) finds zero matching tests because
`critest.bats` has no tagged tests. With bats >= v1.14.0 this causes
a hard error on empty test suites.

Instead of bumping bats and adding `--allow-empty-suite` (which breaks
backward compatibility with older bats versions), add a minimal
serial-tagged placeholder test so the filtered pass always finds at
least one test.

This approach was suggested by @martin-schulze-vireso (bats-core maintainer)
in the [review discussion](https://github.com/cri-o/cri-o/pull/10244#issuecomment-3360710370).

Ref: https://github.com/containers/conmon/pull/672
Ref: https://github.com/bats-core/bats-core/issues/1239

## Summary by CodeRabbit

* **Tests**
  * Added a tagged placeholder test to ensure the serial test suite is detected when filtered.

### DIFF
--- a/test/critest.bats
+++ b/test/critest.bats
@@ -14,6 +14,15 @@ function teardown() {
 	cleanup_test
 }
 
+# Dummy test to ensure --filter-tags 'crio:serial' never produces an empty
+# test suite, which would cause bats to error out. This is needed because
+# critest.bats is the only test file run in conmon CI (via TESTS=critest.bats)
+# and the main test below has no tags.
+# bats test_tags=crio:serial
+@test "placeholder for serial tag filtering" {
+	true
+}
+
 @test "run the critest suite" {
 	WEBSOCKET_ARGS=()
 	if [[ $RUNTIME_TYPE == pod ]]; then


kolyshkin:
@jnovy the problem here is once you bump bats to v1.14 and use `--allow-empty-suite`, you are no longer compatible with older bats versions (as they don't support the new CLI flag). I wish they would rather use an environment variable for this in https://github.com/bats-core/bats-core/pull/1211  (something like `BATS_ALLOW_EMPTY_SUITE=1`) -- this way it would be backwards-compatible. Cc @martin-schulze-vireso.

As bats v1.14 was only released last month, pretty much no distro has it. I'd rather not upgrade just yet.

Is there any other way to fix this to stay backward compatible? I was thinking of using `bats -c` to count the number of tests and skip the execution, but oh well it errors out for me:

```console
[kir@kir-tp1 bats-core]$ bats -c test -f run
209
rm: cannot remove '/home/kir/git/bats-core/test/.bats/run-logs/2026-08-17 22:34:44 UTC.log': No such file or directory
[kir@kir-tp1 bats-core]$ echo $?
1
```

kolyshkin:
> Is there any other way to fix this to stay backward compatible? I was thinking of using `bats -c` to count the number of tests and skip the execution, but oh well it errors out for me:

Hmm, on cri-o it does not:

```console
[kir@kir-tp1 cri-o]$ bats -c --filter-tags '!crio:serial' test/
548
[kir@kir-tp1 cri-o]$ bats -c --filter-tags 'crio:serial' test/
30
[kir@kir-tp1 cri-o]$ bats --version
Bats 1.13.0
```

So it's possible but I'd rather have `BATS_ALLOW_EMPTY_SUITE`.

kolyshkin:
> So it's possible but I'd rather have `BATS_ALLOW_EMPTY_SUITE`.

Opened an issue (https://github.com/bats-core/bats-core/issues/1239) and a fix (https://github.com/bats-core/bats-core/pull/1240).

martin-schulze-vireso:
While I concur that we need BATS_ALLOW_EMPTY_SUITE dir Backwards compatibility, I think you should fix this in another way. 

The intention of this change in Bats is to catch accidental empty suites, which you forgo with this setting. 

Instead, you could add a dummy test with the right tags.

kolyshkin:
> Instead, you could add a dummy test with the right tags.

@jnovy PTAL ^^^

jnovy:
Done. Dropped the bats version bump and --allow-empty-suite in favor of adding a serial-tagged placeholder test to critest.bats, as suggested by @martin-schulze-vireso. This way the --filter-tags crio:serial pass always finds at least one test without requiring any bats version change. Rebased on current main as well.

PR #10248, opened 2026-08-17
config: add new Checkpoint/Restore config block, and deprecate old field

haircommander:
#### What type of PR is this?

/kind feature
we need more granularity on the checkpoint/restore field for containers. partially because pod level checkpointing is on the horizon, and partially because we will be deprecating support for container restore (and possibly container checkpoint).

```release-note
Add a new config field to allow node admins to only allow checkpointing, and disallowing restore.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable checkpoint and restore levels: disabled, checkpoint-only, or full checkpoint/restore.
  * Added the `--checkpoint-restore-container-level-enabled` command-line option, environment support, and shell completions.
  * Added configuration templates and documentation for checkpoint/restore settings.

* **Deprecation**
  * Deprecated `enable_criu_support`; use the checkpoint/restore level setting instead.

* **Bug Fixes**
  * Checkpoint and restore operations now honor their respective configuration levels.
  * Invalid settings and unavailable CRIU requirements are detected during startup.


docs/crio.conf.5.md
```diff
@@ -479,6 +480,17 @@ Specifies the number of CPU shares this pod has access to.
...
+
+The `crio.checkpoint_restore` table contains settings pertaining to the checkpoint and restore (CRIU) support for containers.
+
+**container_level_enabled**="checkpoint_restore"
+Configures the level of container checkpoint and restore (CRIU) support. It accepts one of the following values:
+
+- "none": checkpoint and restore support is disabled.
+- "checkpoint_only": only checkpointing containers is enabled.
+- "checkpoint_restore": both checkpointing and restoring containers is enabled.
+
```
bitoku, line 493: When we have "pod level checkpoint/restore", are we going to add more modes or options?
I'm wondering whether we want to make this option more generic so that it accepts pod-level config too, or should it be configured separately?

docs/crio.conf.5.md
```diff
@@ -479,6 +480,17 @@ Specifies the number of CPU shares this pod has access to.
...
+## CRIO.CHECKPOINT_RESTORE TABLE
+
+The `crio.checkpoint_restore` table contains settings pertaining to the checkpoint and restore (CRIU) support for containers.
+
+**container_level_enabled**="checkpoint_restore"
+Configures the level of container checkpoint and restore (CRIU) support. It accepts one of the following values:
+
+- "none": checkpoint and restore support is disabled.
+- "checkpoint_only": only checkpointing containers is enabled.
+- "checkpoint_restore": both checkpointing and restoring containers is enabled.
```
bitoku, line 492: Since we're going to deprecate `checkpoint_restore` anyway, why don't we remove `checkpoint_restore` option and leave it to `enable_criu_support`?

docs/crio.conf.5.md
haircommander: yeah my intention is to have another one for pod checkpointing, possibly with different options

docs/crio.conf.5.md
haircommander: I think we have some users who will want to continue to use checkpointing but not restore, so this is a band-aid until we can actually drop support because pod level can replace

haircommander:
@cri-o/cri-o-maintainers PTAL

saschagrunert (approved):

